PR #15824, opened 2018-11-24
refactor: eliminate duplicate code exposing process APIs

miniak:
#### Description of Change
- Eliminate duplicate code exposing process APIs (sandboxed vs non-sandboxed renderer).
- Expose missing `process.getCreationTime()` in sandboxed renderers.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Missing `process.getCreationTime()` exposed in sandboxed renderers.

### DIFF
--- a/atom/common/api/atom_bindings.cc
+++ b/atom/common/api/atom_bindings.cc
@@ -61,35 +61,44 @@ AtomBindings::~AtomBindings() {
   uv_close(reinterpret_cast<uv_handle_t*>(&call_next_tick_async_), nullptr);
 }
 
+// static
+void AtomBindings::BindProcess(v8::Isolate* isolate,
+                               mate::Dictionary* process,
+                               base::ProcessMetrics* metrics) {
+  // These bindings are shared between sandboxed & unsandboxed renderers
+  process->SetMethod("crash", &Crash);
+  process->SetMethod("hang", &Hang);
+  process->SetMethod("log", &Log);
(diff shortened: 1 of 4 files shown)


atom/common/api/atom_bindings.cc
```diff
@@ -51,35 +51,43 @@ AtomBindings::~AtomBindings() {
   uv_close(reinterpret_cast<uv_handle_t*>(&call_next_tick_async_), nullptr);
 }
 
+// static
+void AtomBindings::BindProcess(v8::Isolate* isolate,
```
nornagon: Add a comment here that these bindings are shared between sandboxed & unsandboxed renderers

atom/common/api/atom_bindings.cc
```diff
@@ -51,35 +51,44 @@ AtomBindings::~AtomBindings() {
...
+                               mate::Dictionary* process,
+                               base::ProcessMetrics* metrics) {
+  // These bindings are shared between sandboxed & unsandboxed renderers
+  process->SetMethod("crash", &Crash);
+  process->SetMethod("hang", &Hang);
+  process->SetMethod("log", &Log);
```
ppontes: "log" was not available in sandbox renderers originally. Is it intentional to add it and if so should it also be added to process.md docs?

miniak: Log is not documented even in non-sandboxed renderer

PR #15836, opened 2018-11-26
fix: backport patch that ensures that cookie store is always created

### LINKED ISSUE #15365: webview cookies are not persisted
* Output of `node_modules/.bin/electron --version`: 4.0.0-beta.5
* Operating System (Platform and Version): macOS 10.14
* Output of `node_modules/.bin/electron --version` on last known working Electron version (if applicable): There is no issue with Electron v3.

**Expected Behavior**
Cookies should be persisted in a webview that uses partition.

**Actual behavior**
Cookies are lost after relaunching app.

**To Reproduce**
1. Create webview with partition: `<webview partition="persist:webview"/>`
2. Sign in to GitHub to create a cookie.
3. Relaunch app.
4. When webview opens again, the cookies are lost.

brenca:
#### Description of Change

Fixes #15365, but the issue actually affected all cookies. This patch ensures that the cookie store is created properly.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed a bug that made all cookies non persistent.

### DIFF
--- a/atom/browser/net/url_request_context_getter.cc
+++ b/atom/browser/net/url_request_context_getter.cc
@@ -188,6 +188,9 @@ URLRequestContextGetter::Handle::CreateNetworkContextParams() {
         base_path.Append(chrome::kChannelIDFilename);
     network_context_params->restore_old_session_cookies = false;
     network_context_params->persist_session_cookies = false;
+    // TODO(deepak1556): Matches the existing behavior https://git.io/fxHMl,
+    // enable encryption as a followup.
+    network_context_params->enable_encrypted_cookies = false;
   }
 
   // TODO(deepak1556): Decide the stand on chrome ct policy and
--- a/atom/browser/net/url_request_context_getter.cc
+++ b/atom/browser/net/url_request_context_getter.cc
@@ -262,6 +262,11 @@ void URLRequestContextGetter::NotifyContextShuttingDown(
     std::unique_ptr<ResourceContext> resource_context) {
   DCHECK_CURRENTLY_ON(BrowserThread::IO);
 
+  // todo(brenca): remove once C70 lands
+  if (url_request_context_ && url_request_context_->cookie_store()) {
+    url_request_context_->cookie_store()->FlushStore(base::NullCallback());
+  }
+
   context_shutting_down_ = true;
   resource_context.reset();
   net::URLRequestContextGetter::NotifyContextShuttingDown();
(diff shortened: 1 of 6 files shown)


brenca:
@deepak1556 added a test for this

patches/common/chromium/ensure_cookie_store.patch
```diff
@@ -0,0 +1,376 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Matt Menke <mmenke@chromium.org>
+Date: Fri, 27 Jul 2018 19:05:47 +0000
+Subject: Backports crrev.com/c/1152083
+ https://chromium-review.googlesource.com/c/chromium/src/+/1152083
```
nornagon:
```suggestion
```

atom/browser/net/url_request_context_getter.cc
```diff
@@ -259,6 +262,11 @@ void URLRequestContextGetter::NotifyContextShuttingDown(
     std::unique_ptr<ResourceContext> resource_context) {
   DCHECK_CURRENTLY_ON(BrowserThread::IO);
 
+  // todo(brenca): remove once C70 lands
+  if (url_request_context_ && url_request_context_->cookie_store()) {
+    url_request_context_->cookie_store()->FlushStore(base::NullCallback());
```
deepak1556: Is this necessary ? I was able to confirm that cookies are persisted without any external flushing. My test scenario was, from a clean app state.

* visit https://github.com and sign-in
* exit app and restart
* account remains signed-in

brenca: Alright, not sure what's going on here - the test that I added passes without the explicit flush, but the test app I used doesn't keep the cookies. Can you check if this works on your machine? https://gist.github.com/brenca/205ea1876c844b4a6eec58ee735f6537

deepak1556: Hmm if the app is quit immediately the cookie is not persisted, must be the persistent db write period https://cs.chromium.org/chromium/src/net/extras/sqlite/sqlite_persistent_cookie_store.cc?l=1277, seems good to have an external flush.

codebytere: @deepak1556 @brenca is this ready to merge or do more changes need to be made?

deepak1556: I am good with the current patch, needs to be rebased to fix conflicts before merge.

ffflorian:
Looking forward to this fix! Will this trigger a new beta release? :slightly_smiling_face:

ckerr:
We're already working on a new beta release, actually. There have been other tooling issues that are holding that up.

javan:
Looks like this change didn't make it into [v4.0.0-beta.8](https://github.com/electron/electron/releases/tag/v4.0.0-beta.8), correct?

brenca:
Yepp, should be in the next beta.

PR #15841, opened 2018-11-26
chore: add promise api runtime flag

codebytere:
#### Description of Change

Refs: https://github.com/electron/maintainers/issues/159

Add and document runtime flag which will be used to determine whether or not deprecation warnings are shown when formerly callback-based functions since converted to Promises trigger deprecation warnings.

/cc @ckerr @MarshallOfSound 
 
#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


lib/common/api/deprecate.js
```diff
@@ -73,7 +73,7 @@ const deprecate = {
     const warn = warnOnce(oldName, newName)
 
     if (typeof cb !== 'function') return promise
-    warn()
+    if (process.enablePromiseAPIs) warn()
```
MarshallOfSound: This should actually prevent a promise being returned at all.  If this flag is set to true, we should be returning undefined

codebytere: Resolving based on conversation in Slack ☝️

Issue #15095, opened 2018-10-11
flaky test: webContents module takeHeapSnapshot() fails with invalid file path


brenca:
* Output of `node_modules/.bin/electron --version`: 4.0.0
* Operating System (Platform and Version): OSX
* Output of `node_modules/.bin/electron --version` on last known working Electron version (if applicable):

**Expected Behavior**
Test should behave consistently.

**Actual behavior**
Test seems to be flaky.

**Additional Information**
Here is the flaked test run:
https://github.com/electron/electron/pull/15023/checks?check_run_id=23183436
And the successful re-run:
https://github.com/electron/electron/pull/15023/checks?check_run_id=23228533

nornagon:
Still flaky :< https://github.visualstudio.com/electron/_build/results?buildId=23199&view=logs

I'm going to disable this test.

PR #15855, opened 2018-11-27
feat: webContents.loadURL returns a promise

nornagon:
#### Description of Change
Many tests use the following pattern to load a page in a WebContents and wait for the load to finish:

```js
async () => {
  webContents.loadFile('path/to/page.html')
  await emittedOnce(webContents, 'did-finish-load')
  // ... do something with the page
}
```

This is an aesthetic improvement over the callback method of `webContents.on('did-finish-load', ...)`, but is unfortunately flaky: it is sometimes possible that the `'did-finish-load'` event is emitted before the listener is registered (see #15853 for example, which fixes the flaky test described in #15095). This is to do with the fact that the tests use the `BrowserWindow` module via the `remote` API, so there is a short window of time after loading the URL but before the event listener is registered. If the event is fired during that window, it will be lost, and the test times out.

It's possible to eliminate the additional `emittedOnce` helper call, clean up the code, and remove the possibility of flakiness by having `loadURL` return a promise. The above could would be written as follows with this patch:

```js
async () => {
  await webContents.loadFile('path/to/page.html')
  // ... do something with the page
}
```

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: webContents.loadURL and loadFile now return a promise

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -1236,7 +1236,11 @@ Captures a snapshot of the page within `rect`. Omitting `rect` will capture the 
   * `postData` ([UploadRawData[]](structures/upload-raw-data.md) | [UploadFile[]](structures/upload-file.md) | [UploadBlob[]](structures/upload-blob.md)) (optional)
   * `baseURLForDataURL` String (optional) - Base url (with trailing path separator) for files to be loaded by the data url. This is needed only if the specified `url` is a data url and needs to load other files.
 
-Same as `webContents.loadURL(url[, options])`.
+Returns `Promise` - the promise will resolve when the page has finished loading
+(see [`did-finish-load`](web-contents.md#event-did-finish-load)), and rejects
+if the page fails to load (see [`did-fail-load`](web-contents.md#event-did-fail-load)).
+
+Same as [`webContents.loadURL(url[, options])`](web-contents.md#contentsloadurlurl-options).
 
 The `url` can be a remote address (e.g. `http://`) or a path to a local
 HTML file using the `file://` protocol.
@@ -1275,6 +1279,10 @@ win.loadURL('http://localhost:8000/post', {
   * `query` Object (optional) - Passed to `url.format()`.
   * `search` String (optional) - Passed to `url.format()`.
   * `hash` String (optional) - Passed to `url.format()`.
+
+Returns `Promise` - the promise will resolve when the page has finished loading
+(see [`did-finish-load`](web-contents.md#event-did-finish-load)), and rejects
+if the page fails to load (see [`did-fail-load`](web-contents.md#event-did-fail-load)).
 
 Same as `webContents.loadFile`, `filePath` should be a path to an HTML
 file relative to the root of your application.  See the `webContents` docs
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -1236,7 +1236,7 @@ Captures a snapshot of the page within `rect`. Omitting `rect` will capture the 
   * `postData` ([UploadRawData[]](structures/upload-raw-data.md) | [UploadFile[]](structures/upload-file.md) | [UploadBlob[]](structures/upload-blob.md)) (optional)
   * `baseURLForDataURL` String (optional) - Base url (with trailing path separator) for files to be loaded by the data url. This is needed only if the specified `url` is a data url and needs to load other files.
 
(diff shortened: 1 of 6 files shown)


lib/browser/api/navigation-controller.js
```diff
@@ -55,9 +55,35 @@ const NavigationController = (function () {
...
+        Object.assign(err, {
+          errno: errorCode,
+          code: errorDescription,
+          validatedURL,
+          frameProcessId,
+          frameRoutingId
+        })
+        reject(err)
+      }
+      this.webContents.once('did-finish-load', finishListener)
```
MarshallOfSound: I know this is getting picky but if we're exposing this as API we should think about it.

Is it possible for this `did-finish-load` event be for a previous load?  E.g. if two calls are sequentially made to `loadURL` will the first promise resolve or reject?  If a `loadURL` call is made a few ticks before a different navigation finishes would this promise be resolved eroneously?

Unless we validate a navigation request ID or something I don't think we can 100% sure this is the same navigation

nornagon: Yeah... it's also possible for the promise to _never resolve_, which I noticed when trying to test load failures with a bad scheme (like `bad-scheme://foo`). The navigation gets cancelled and you're left hanging.

I think these are fundamental issues with the API as it stands (the promise is just exposing the existing flaws) and we should probably see if we can fix them.

lib/browser/api/navigation-controller.js
```diff
@@ -55,9 +55,66 @@ const NavigationController = (function () {
...
+      }
+
+      let navigationStarted = false
+      const navigationListener = (event, url, isSameDocument, isMainFrame, frameProcessId, frameRoutingId, navigationId) => {
+        if (isMainFrame) {
+          if (navigationStarted) {
+            // the webcontents has started another unrelated navigation in the
+            // main frame (probably from the app calling `loadURL` again); reject
+            // the promise
+            return rejectAndCleanup(-3, 'ERR_ABORTED', url)
```
MarshallOfSound: Can we explicitly test this case?

nornagon: done

lib/browser/api/navigation-controller.js
nornagon: fixed 'em

lib/browser/api/navigation-controller.js
```diff
@@ -55,9 +55,66 @@ const NavigationController = (function () {
...
+          }
+          navigationStarted = true
+        }
+      }
+      const stopLoadingListener = () => {
```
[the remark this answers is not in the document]
nornagon: I figured this would be a reasonable "catch-all", since I think `did-stop-loading` will be emitted regardless of what happens?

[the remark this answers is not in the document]
nornagon: but `CanRequestURL` might be a good extra check to give a more informative error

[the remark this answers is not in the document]
deepak1556: > since I think did-stop-loading will be emitted regardless of what happens?

yup that is correct, its emitted whenever a document stops its resource loading be it failure or success.  Having this check can simplify existing url validity check https://github.com/electron/electron/blob/master/atom/browser/api/atom_api_web_contents.cc#L1157

docs/api/browser-window.md
```diff
@@ -1236,7 +1236,11 @@ Captures a snapshot of the page within `rect`. Omitting `rect` will capture the
   * `postData` ([UploadRawData[]](structures/upload-raw-data.md) | [UploadFile[]](structures/upload-file.md) | [UploadBlob[]](structures/upload-blob.md)) (optional)
   * `baseURLForDataURL` String (optional) - Base url (with trailing path separator) for files to be loaded by the data url. This is needed only if the specified `url` is a data url and needs to load other files.
 
-Same as `webContents.loadURL(url[, options])`.
+Returns `Promise` - the promise will resolve when the page has finished loading
```
MarshallOfSound: Can we be super specific what the `Promise` type is here; if you don't declare the type it defaults to `any`.  I believe here this should be `Promise<void>` indicating the promise resolves with nothing.

(Check the archeologist for the different in types generation)

(Same for all other new promise docs in this PR)

nornagon: done

PR #15859, opened 2018-11-28
security: only handle related IPCs when <webview> tag is enabled

miniak:
#### Description of Change
- Move `<webview>` related IPC handling from `rpc-server.js` to `guest-view-manager.js`, where all other related IPCs are being handled.
- Only handle related IPCs when the `<webview>` tag is enabled. This should increase security as it lowers the attack surface of the IPC channel.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/guest-view-manager.js
+++ b/lib/browser/guest-view-manager.js
@@ -3,6 +3,7 @@
 const { webContents } = require('electron')
 const ipcMain = require('@electron/internal/browser/ipc-main-internal')
 const parseFeaturesString = require('@electron/internal/common/parse-features-string')
+const errorUtils = require('@electron/internal/common/error-utils')
 
 // Doesn't exist in early initialization.
 let webViewManager = null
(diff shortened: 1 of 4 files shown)


lib/browser/guest-view-manager.js
```diff
@@ -319,29 +320,71 @@ const watchEmbedder = function (embedder) {
   })
 }
 
-ipcMain.on('ELECTRON_GUEST_VIEW_MANAGER_CREATE_GUEST', function (event, params, requestId) {
+const handleMessage = function (channel, handler) {
+  ipcMain.on(channel, (event, ...args) => {
+    if (event.sender._isWebViewTagEnabled()) {
```
MarshallOfSound: If this is firing for a lot of events (E.g. a lot of async calls) should we memoize `_isWebViewTagEnabled`?

On a related not we don't need to expose a new API for this, we can use `event.sender.getLastWebPreferences().webviewTag`

miniak: reimplemented as suggested

PR #15821, opened 2018-11-23
fix: use appropriate site instance for cross-site nav's

ppontes:
#### Description of Change
When a window is opened with window.open to navigate to a cross-site destination, the new site instance should belong to the browsing instance of the opener window in order not to break scripting relationships between windows.

fixes #8100.

This change turns Chromium's site isolation on and depends on it being on in order to work properly. Turning on site isolation revealed the following latent problems that are / will be fixed also in this PR:
- [X] If a cross-site redirection happens during a navigation, the renderer process that will handle the final URL will be locked to the pre-redirection URL with Chromium's child process security. If the page tries to access local resources, Chromium will refuse and kill the renderer process.
- [X] Chromium now kills renderer processes that attempt to access resources for which they have no permissions. This means that the "custom non standard schemes cannot access localStorage" UT had to be adapted.

The fix also requires the following to be addressed:

- [x] If a cross-site redirection happens during a navigation, Chromium discards the speculative RFH that was created for the pre-redirection URL. This is harmless in Chromium but Electron reacts by asynchronously closing the associated BrowserWindow. User sees the new window navigating to the final URL and then being closed.

#### Checklist
- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

#### Release Notes

Notes: Fixed a bug where `window.opener` of a window created with window.open from a sandboxed renderer was null.

PR #15870, opened 2018-11-28
feat: support mixed-sandbox mode on linux

nornagon:
#### Description of Change
This adds a patch to Chromium to allow mixing `--no-zygote` renderers with regular zygote-launched renderers.

This currently only works when launching the app with `--enable-mixed-sandbox` on the command line—when using the `app.enableMixedSandbox` API, the zygote process is launched before the JS initialization code is run, and thus fails to pick up the mixed-sandbox flag and launches with `--no-sandbox`, resulting in a weird mix of zygote-launched and non-zygote-launched renderers all of which are without sandbox. My preferred fix for that problem is to remove the `--enable-mixed-sandbox` flag and have it be the default (overridden by `--no-sandbox` or `--enable-sandbox`).

I'm not exactly sure how this will interact with site-per-process and navigation. In particular, will this behave correctly if you load `a.com` in a sandboxed renderer and then navigate to `b.com` with site-per-process enabled? (Further, what's the status of site-per-process? cc @ppontes)

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: mixed-sandbox mode works on Linux

### DIFF
--- a/atom/browser/web_contents_preferences.cc
+++ b/atom/browser/web_contents_preferences.cc
@@ -245,10 +245,12 @@ void WebContentsPreferences::AppendCommandLineSwitches(
   // If the `sandbox` option was passed to the BrowserWindow's webPreferences,
   // pass `--enable-sandbox` to the renderer so it won't have any node.js
   // integration.
-  if (IsEnabled(options::kSandbox))
+  if (IsEnabled(options::kSandbox)) {
     command_line->AppendSwitch(switches::kEnableSandbox);
-  else if (!command_line->HasSwitch(switches::kEnableSandbox))
+  } else if (!command_line->HasSwitch(switches::kEnableSandbox)) {
     command_line->AppendSwitch(service_manager::switches::kNoSandbox);
+    command_line->AppendSwitch(::switches::kNoZygote);
+  }
 
   // Check if nativeWindowOpen is enabled.
   if (IsEnabled(options::kNativeWindowOpen))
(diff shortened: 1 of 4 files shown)


ppontes:
@nornagon, site-per-process is also WIP: [15821](https://github.com/electron/electron/pull/15821). We're addressing one last issue before we can attempt to merge it.

miniak:
@nornagon would it make sense to keep two zygotes? one sandboxed and one non-sandboxed?

nornagon:
possibly, but it would be a rather more involved change and it's not clear to me that the zygote is providing substantial value other than optimizing sandbox setup (see e.g. [rsesek's comment here](https://chromium-review.googlesource.com/c/chromium/src/+/1316356/1#message-5dec694040cec1ff1e06188a0558883af176612e)). It may indeed be a worthwhile improvement, but without data to back it up I don't think it's worth going through the effort of supporting a 2nd zygote.

nornagon:
ping @zcbenz @deepak1556, this is ready for review :)

deepak1556:
Can we have a tracking issue for https://github.com/electron/electron/pull/15870#issuecomment-443896793 ? Could lead to a potential enhancement.

nornagon:
I don't think it makes sense to have a tracking issue at this time, given we haven't even evaluated whether it would be an improvement or not.

deepak1556:
>  given we haven't even evaluated whether it would be an improvement or not.

The tracking issue in my mind was for this specifically, don't want it to get lost in PR comments :)

nornagon:
> don't want it to get lost in PR comments

want it to get lost in the issue tracker instead? :D

nornagon:
Merging even though I still have some concerns about the way this might interact with site-per-process, we can revert / fix forward if we discover issues.

PR #15871, opened 2018-11-28
test: fix a bunch of flaky tests related to emittedOnce

nornagon:
#### Description of Change

It seems that a common source of test flake is events firing before listeners are registered, in particular with `loadURL` and `did-finish-load`. It's not clear to me how those events could be getting fired early, but they are, and this fixes it, hopefully.

Relatedly: #15855

#### Checklist

#### Release Notes

Notes: no-notes


alexeykuzmin:
@nornagon Can you please create a backport it to `4-0-x`?

PR #15873, opened 2018-11-29
chore: DCHECK that events are only emitted on the UI thread

nornagon:
#### Description of Change
Emitting events on non-UI threads is a Very Bad Idea

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/api/event_emitter.h
+++ b/atom/browser/api/event_emitter.h
@@ -8,6 +8,7 @@
 #include <vector>
 
 #include "atom/common/api/event_emitter_caller.h"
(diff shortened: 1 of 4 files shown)


deepak1556:
Just to get a bit more context on what triggered this, was there an improper usage of this api ?

nornagon:
@MarshallOfSound mentioned that there have been improper usages in the past, and I was worried that `Emit` might have been getting called from an IPC thread or IO thread in some cases and that might have been a cause of the test flakiness fixed by #15871. It wasn't, but if this was here it would have been easy to be sure about it :)

PR #15555, opened 2018-11-02
fix: ensure that file descriptors 0/1/2 are opened at startup

anszom:
#### Description of Change
Libuv has a nasty mis-feature (https://github.com/libuv/libuv/issues/2062), where it marks the standard input/output/error tty streams with the close-on-exec flag. In effect, when the chromium gpu process is spawned, it starts with the fd's 0,1,2 closed. These are reused for other purposes, and if any message is output on stdout/stderr, it ends up in an unrelated socket, pipe or file which generally messes things up. This is the root cause of at least two issues: https://github.com/electron/electron/issues/14246 and https://github.com/electron/electron/issues/12937, and possibly these (duplicates) as well: https://github.com/electron/electron/issues/13415, https://github.com/electron/electron/issues/14954, https://github.com/electron/electron/issues/14697

The proper place to fix this would be in libuv, but the upstream developers insist that the current behaviour is right. My change introduces a workaround which checks the standard fds at startup, and reopens them as /dev/null if missing. Feel free to point me to a better solution.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

In my environment, npm test on the 3-0-x branch fails on 5 tests.
My commit doesn't introduce any more failures.

#### Release Notes

Notes: Fixed a freeze when using the GPU in some configurations.

PR #15884, opened 2018-11-29
fix: ignore unused freopen result

codebytere:
#### Description of Change

The `freopen`s introduced in https://github.com/electron/electron/pull/15555 were erroring on release builds a result of their return values being unused. This wraps them in a macro to mitigate that error.

/cc @deepak1556 @ckerr 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/app/atom_main.cc
+++ b/atom/app/atom_main.cc
@@ -69,11 +69,11 @@ void FixStdioStreams() {
   // For details see https://github.com/libuv/libuv/issues/2062
   struct stat st;
   if (fstat(STDIN_FILENO, &st) < 0 && errno == EBADF)
-    freopen("/dev/null", "r", stdin);
+    ignore_result(freopen("/dev/null", "r", stdin));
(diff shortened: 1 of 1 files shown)


atom/app/atom_main.cc
```diff
@@ -69,11 +69,11 @@ void FixStdioStreams() {
   // For details see https://github.com/libuv/libuv/issues/2062
   struct stat st;
   if (fstat(STDIN_FILENO, &st) < 0 && errno == EBADF)
-    freopen("/dev/null", "r", stdin);
+    ignore_result(freopen("/dev/null", "r", stdin));
```
nornagon: I think the better fix here would be `CHECK(freopen(...))`. On success, the return value is just `stdin`, but on failure the return value is `null`.

nornagon: I see this is already merged and backported, but we should at least switch this to `CHECK` on master.

PR #15893, opened 2018-11-29
fix: restore ability to disable color correct rendering

poiru:
In Electron 2.0, `--disable-features=ColorCorrectRendering` could be
used to make the app use the display color space (e.g. P3 on Macs)
instead of color correcting to sRGB. Because color correct rendering is
always enabled on Chromium 62 and later and because
`--force-color-profile` has no effect on macOS, apps that need e.g. P3
colors are currently stuck on Electron 2.0.

This restores the functionality removed in
https://chromium-review.googlesource.com/698347 in the form of the
`--disable-color-correct-rendering` switch.

This can be removed once web content (including WebGL) learn how
to deal with color spaces. That is being tracked at
https://crbug.com/634542 and https://crbug.com/711107.

As an example of a widely used app using
`--disable-features=ColorCorrectRendering`, see VSCode:
https://github.com/Microsoft/vscode/blob/3f33ef2593d3efe6eca5230f3d34d3406fb73498/src/main.js#L138-L139

Notes: Add `--disable-color-correct-rendering` switch

### DIFF
--- a/patches/common/chromium/.patches
+++ b/patches/common/chromium/.patches
@@ -77,3 +77,4 @@ content_allow_embedder_to_prevent_locking_scheme_registry.patch
 fix_trackpad_scrolling.patch
 mac_fix_form_control_rendering_on_10_14_mojave.patch
 ensure_cookie_store.patch
+disable_color_correct_rendering.patch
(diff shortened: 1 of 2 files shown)


poiru:
I was working with an older Electron checkout using Chromium 68 and didn't fully test this after upgrading to Chromium 69. Something in the 68 => 69 transition broke this patch for `<canvas>` although it continues to work for CSS and image colors. I'll investigate and open a new PR after I figure it out.

poiru:
Here's the new PR: #15898

PR #15894, opened 2018-11-29
feat: enable mixed-sandbox mode by default

nornagon:
#### Description of Change
Currently, in order to launch some renderers sandboxed, an app must enable mixed-sandbox mode either with the command-line switch `--enable-mixed-sandbox` or the `app.enableMixedSandbox()` API. This PR will deprecate that switch, and make mixed-sandbox mode the default. Renderers launched with `sandbox: true` will now be actually sandboxed, where previously they would only be sandboxed if mixed-sandbox mode was also enabled.

Depends on #15870 for mixed-sandbox mode to be meaningful on Linux.

#### Checklist

#### Release Notes

Notes: Mixed-sandbox mode is now enabled by default. `enableMixedSandbox` and the `--enable-mixed-sandbox` command-line switch still exist for compatibility, but are deprecated and have no effect.

### DIFF
--- a/atom/app/atom_main_delegate.cc
+++ b/atom/app/atom_main_delegate.cc
@@ -204,12 +204,6 @@ void AtomMainDelegate::PreSandboxStartup() {
   // linux (namespace sandbox is available on most distros).
   command_line->AppendSwitch(service_manager::switches::kDisableSetuidSandbox);
 
-  if (!command_line->HasSwitch(switches::kEnableMixedSandbox) &&
-      !command_line->HasSwitch(switches::kEnableSandbox)) {
-    // Disable renderer sandbox for most of node's functions.
-    command_line->AppendSwitch(service_manager::switches::kNoSandbox);
-  }
-
   // Allow file:// URIs to read other file:// URIs by default.
   command_line->AppendSwitch(::switches::kAllowFileAccessFromFiles);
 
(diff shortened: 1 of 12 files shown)


miniak:
Be aware that
- GPU is going to be sandboxed by default, this should not have any negative effect on apps
- pepper plugin host processes are going to be sandboxed, this might be a breaking change if apps depend on so-called host-native pepper plugins

nornagon:
Additionally, devtools processes will be sandboxed.

MarshallOfSound:
@nornagon will that break devtron?

nornagon:
Pushed a change which unsandboxes devtools processes.

zcbenz:
@nornagon The Linux CI are still failing for devtools tests, it seems that sandbox are still enabled for devtools process because of zygote process?

nornagon:
As mentioned in the PR description, this depends on #15870

nornagon:
hm, actually it looks like that PR is already merged in the base of this PR. I'll investigate.

nornagon:
@zcbenz linux tests are now passing; i think the ARM failures are unrelated

nornagon:
Talked with @miniak and agreed that we can follow up on the ppapi sandboxing stuff during the beta cycle.

PR #15900, opened 2018-11-30
chore: improve granular release cleanup

codebytere:
#### Description of Change

Adapts the release cleanup script to account for the case where a commit is pushed to the release branch but a tag and draft is not created by removing the interdependence of these functions. Also ensures that tags are not deleted unless drafts are deleted, so that a tag is never accidentally removed from a published draft.

/cc @MarshallOfSound @ckerr @BinaryMuse

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


codebytere:
☝️ backports are blocked on https://github.com/electron/electron/pull/15904, i'll open then (should be clean) once it's merged

PR #15921, opened 2018-12-02
chore: move getCurrentBranch to a util file

codebytere:
#### Description of Change

`getCurrentBranch` is used in the same manner across several release files; to DRY it out this PR moves it to a utils file.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/lib/utils.js
+++ b/script/lib/utils.js
@@ -1,6 +1,11 @@
 const OUT_DIR = process.env.ELECTRON_OUT_DIR || 'Debug'
 
+const { GitProcess } = require('dugite')
 const path = require('path')
+
+require('colors')
+const pass = '\u2713'.green
+const fail = '\u2717'.red
 
 function getElectronExec () {
   switch (process.platform) {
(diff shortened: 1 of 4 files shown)


script/lib/utils.js
```diff
@@ -19,7 +24,22 @@ function getAbsoluteElectronExec () {
...
+  const gitArgs = ['rev-parse', '--abbrev-ref', 'HEAD']
+  const branchDetails = await GitProcess.exec(gitArgs, gitDir)
+  if (branchDetails.exitCode === 0) {
+    const currentBranch = branchDetails.stdout.trim()
+    console.log(`${pass} current git branch is: ${currentBranch}`)
+    return currentBranch
+  } else {
+    const error = GitProcess.parseError(branchDetails.stderr)
+    console.log(`${fail} couldn't get details current branch: `, error)
+    process.exit(1)
```
BinaryMuse: I wonder if this should throw instead of just exiting? 🤔 That said, everywhere it's used exiting is the right behavior. Your call

PR #15927, opened 2018-12-03
fix: support async child process methods without callback in asar

alexgreenland:
#### Description of Change

In Node land, async child process methods do not require a callback to be provided. In Electron land in the packaged asar environment, there's a bug which means if you don't provide a callback the method does not run — it runs synchronously and blocks execution when a long running child process is launched. I saw this with `execFile`, I had to add a callback to make the child process run.

In development, original Node is used, but in production, Electron's Node (with changes) is used. This means there is a discrepancy between development and production environments and will catch users out, as it did to me. As it is a silent failure, it's quite difficult to spot.

This change no longer branches to call the synchronous method but continues to use the async method as requested and makes this flow work. I added a new asar test and all asar tests pass.

#### Checklist

- [x] `npm test` passes
- [x] tests added

#### Release Notes

Notes: support async child process methods without callback in asar

### DIFF
--- a/lib/common/asar.js
+++ b/lib/common/asar.js
@@ -171,9 +171,6 @@
       if (!isAsar) return old.apply(this, arguments)
 
       const callback = arguments[arguments.length - 1]
-      if (typeof callback !== 'function') {
-        return overrideAPISync(module, name, pathArgumentIndex)
-      }
 
       const archive = getOrCreateArchive(asarPath)
       if (!archive) {
--- a/spec/asar-spec.js
+++ b/spec/asar-spec.js
@@ -912,6 +912,18 @@ describe('asar package', function () {
         })
       })
 
+      it('executes binaries without callback', function (done) {
+        const process = execFile(echo, ['test'])
+        process.on('close', function (code) {
+          assert.strictEqual(code, 0)
+          done()
+        })
+        process.on('error', function () {
+          assert.fail()
+          done()
+        })
+      })
+
       it('execFileSync executes binaries', function () {
         const output = execFileSync(echo, ['test'])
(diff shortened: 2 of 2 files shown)


alexgreenland:
Looking closer, I don't think this has ever worked, the flow from async to sync (`overrideAPI` -> `overrideAPISync`). 

If you look at the original code: 

https://github.com/electron/electron/blob/90d1c0b2b6307ca64fff6576e91c2e003889a65b/lib/common/asar.js#L165-L176 

When there's no callback `overrideAPI`  invokes `overrideAPISync` but `overrideAPISync` merely overrides the API with `module[name] = function ()`, it crucially doesn't run, which it needs to because we're already in our own invocation `module[name] = function ()` from `overrideAPI`. It also has the side effect of overriding the async method with the sync method for future invocations.

https://github.com/electron/electron/blob/90d1c0b2b6307ca64fff6576e91c2e003889a65b/lib/common/asar.js#L146-L149

I've introduced a more comprehensive change that passes a flag `fromAsync` from `overrideAPI` into `overrideAPISync` when there's no callback. When we detect that we're coming from an async flow, we return the overridden function and `overrideAPI` invokes it with the original arguments.

My change also means that we don't force the sync method on all further async calls, so if you didn't include a callback when calling once, it will allow you to use a callback in future invocations — there's an at-call check so it doesn't set `module[name]` in `overrideAPISync` in this flow.

The new code:

https://github.com/electron/electron/blob/c58517f6723f57816a9e86a45ef814c1279f0af7/lib/common/asar.js#L146-L180

All original and introduced asar tests pass.

alexgreenland:
@zcbenz @MarshallOfSound ^^

PR #15929, opened 2018-12-03
security: block chrome.tabs.executeScript() for non chrome-extension: URLs

miniak:
#### Description of Change
Don't handle IPCs implementing Chrome Extension APIs if the sender is not an extension (URL needs to be `chrome-extension:` scheme). Similar to https://github.com/electron/electron/pull/15859.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/chrome-extension.js
+++ b/lib/browser/chrome-extension.js
@@ -201,7 +201,18 @@ ipcMain.on('CHROME_TABS_SEND_MESSAGE', function (event, tabId, extensionId, isBa
   resultID++
 })
 
+const isChromeExtension = function (pageURL) {
+  const { protocol } = url.parse(pageURL)
+  return protocol === 'chrome-extension:'
+}
+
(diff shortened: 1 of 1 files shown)


lib/browser/chrome-extension.js
```diff
@@ -201,7 +201,17 @@ ipcMain.on('CHROME_TABS_SEND_MESSAGE', function (event, tabId, extensionId, isBa
   resultID++
 })
 
+const isChromeExtension = function (contents) {
```
alexeykuzmin: Can you pass a URL here instead of WebContents?

lib/browser/chrome-extension.js
```diff
@@ -201,7 +201,17 @@ ipcMain.on('CHROME_TABS_SEND_MESSAGE', function (event, tabId, extensionId, isBa
...
+  const { protocol } = url.parse(contents._getURL())
```
alexeykuzmin: Why `_getURL` here and `getURL` below?

Issue #15912, opened 2018-12-01
The `windowMenu` MenuItem role is incorrect on macOS


sindresorhus:
* Output of `node_modules/.bin/electron --version`: v3.0.10
* Operating System (Platform and Version): macOS 10.14.1

**Expected Behavior**

I expected the `windowMenu` role to make the menu look like a native macOS app.

Here's how it should look like (From a new Xcode macOS project):

<img width="170" alt="fresh xcode project" src="https://user-images.githubusercontent.com/170270/49296966-23824800-f4ec-11e8-8b12-738311cee241.png">

**Actual behavior**

Here's how it actually looks like (In Electron):

<img width="170" alt="windowmenu" src="https://user-images.githubusercontent.com/170270/49296869-f59d0380-f4eb-11e8-9ff1-2eee8b49d558.png">

Notice how it's missing the `Zoom` menu item and the list of windows. It should also not have a `Close Window` menu item, that should be in the `File` menu.

PR #15930, opened 2018-12-03
fix:  correct windowMenu role on MacOS

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/15912.

Updates `windowMenu` MenuItem role to more accurately reflect the default window menu on MacOS. Also updates the default app to use this role.

/cc @sindresorhus @BinaryMuse

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: correct windowMenu MenuItem role on MacOS

### DIFF
--- a/default_app/menu.js
+++ b/default_app/menu.js
@@ -69,15 +69,7 @@ const setDefaultApplicationMenu = () => {
       ]
     },
     {
-      role: 'window',
-      submenu: [
-        {
-          role: 'minimize'
-        },
-        {
-          role: 'close'
-        }
-      ]
+      role: 'windowMenu'
     },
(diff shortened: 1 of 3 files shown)


MarshallOfSound:
Is this technically `semver/major`?

codebytere:
Maybe? i'd say fix is arguable since it's incorrect for MacOS as it currently stands

BinaryMuse:
Hm, I could see the argument each way. Even though the behavior is technically incorrect, if fixing it could potentially cause apps or behavior to change in a non-backward-compat way, I'd argue that it's probably better a `semver/major` change too.

PR #15938, opened 2018-12-03
chore: publish nightlies to electron-nightly

MarshallOfSound:
cc @BinaryMuse @nornagon 

Notes: no-notes

### DIFF
--- a/script/publish-to-npm.js
+++ b/script/publish-to-npm.js
@@ -114,10 +114,20 @@ new Promise((resolve, reject) => {
 
     if (release.tag_name.indexOf('nightly') > 0) {
       if (currentBranch === 'master') {
-        npmTag = 'nightly'
+        // Nightlies get published to their own module, so master nightlies should be tagged as latest
+        npmTag = 'latest'
       } else {
         npmTag = `nightly-${currentBranch}`
       }
+
+      const currentJson = JSON.stringify(fs.readFileSync(path.join(tempDir, 'package.json'), 'utf8'))
+      currentJson.name = 'electron-nightly'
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
Please note releases won't work until appropriate changes have been made to `electron-download`

script/publish-to-npm.js
```diff
@@ -118,6 +118,15 @@ new Promise((resolve, reject) => {
...
       }
+
+      const currentJson = JSON.stringify(fs.readFileSync(path.join(tempDir, 'package.json'), 'utf8'))
+      currentJson.name = 'electron-nightly'
+      rootPackageJson.name = 'electron-nightly'
+
+      fs.writeFileSync(
+        path.join(tempDir, 'package.json'),
+        JSON.stringify(currentJson, null, 2)
+      )
```
nornagon: i don't suppose we could arrange to have this information the _first_ time we write to `tempDir/package.json`?

MarshallOfSound: Unfortunately not, it's a chicken --> egg problem.  We need the temporary package.json to get the release, and we need to the release to get whether it's nightly or not.

In an ideal world this would all be refactored, but with no tests for this file I'd like to keep the changes to a minimum to reduce the chance this breaks at the end of a 2 hour release 😆

MarshallOfSound:
Hold on this, there's a big bad bug

nornagon:
I'm still not totally convinced that `electron-nightly` is the right way to go, but this change does indeed appear to publish to `electron-nightly` appropriately.

PR #15950, opened 2018-12-04
perf: optimize invocation of promisifed functions / methods via the remote module

miniak:
#### Description of Change
Instead of 2 sync IPCs only 1 async IPC is needed to invoke functions / methods returning promises. The trick is to convert a promisifed function to callback based in the main process (`rpc-server.js`) when invoking over IPC and then transparently wrap it back to Promise on the renderer side (`remote.js`).

before:
- `ipcRenderer.sendSync` to invoke method -> returns proxy `Promise`
- `ipcRenderer.sendSync` to invoke `then` on the `Promise` instance

after:
- `ipcRenderer.send` (async) to invoke method

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Optimized invocation of promisifed functions / methods via the `remote` module

### DIFF
--- a/lib/browser/rpc-server.js
+++ b/lib/browser/rpc-server.js
@@ -21,6 +21,7 @@ const bufferUtils = require('@electron/internal/common/buffer-utils')
 const errorUtils = require('@electron/internal/common/error-utils')
 const clipboardUtils = require('@electron/internal/common/clipboard-utils')
 const { isParentDir } = require('@electron/internal/common/path-utils')
+const { isPromisified } = require('@electron/internal/common/promise-utils')
 
 const hasProp = {}.hasOwnProperty
 
@@ -49,6 +50,9 @@ const getObjectMembers = function (object) {
     const member = { name, enumerable: descriptor.enumerable, writable: false }
     if (descriptor.get === undefined && typeof object[name] === 'function') {
       member.type = 'method'
+      if (isPromisified(object[name])) {
+        member.promisified = true
+      }
     } else {
       if (descriptor.set || descriptor.writable) member.writable = true
       member.type = 'get'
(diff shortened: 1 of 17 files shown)


lib/browser/rpc-server.js
```diff
@@ -236,9 +243,34 @@ const unwrapArgs = function (sender, contextId, args) {
   return args.map(metaToValue)
 }
 
+const callbackify = function (func) {
```
codebytere: We can probably use node's [`util.callbackify()`](https://nodejs.org/dist/latest-v10.x/docs/api/util.html#util_util_callbackify_original) here

miniak:
@MarshallOfSound we need to know in advance that we are calling a promisified method / function. In this case we do `ipcRenderer.send` instead of `ipcRenderer.sendSync`

miniak:
@MarshallOfSound also those V8 hidden values are just hints. If the annotation is missing, the API still works just fine, it's just without the performance optimization.

MarshallOfSound:
>  If the annotation is missing, the API still works just fine, it's just without the performance optimization.

It's just my opinion is that we shouldn't be using v8 hiddens to optimize these.  Wrapping all our prototype methods like this is IMO not that great.  If we can't find a way to do it generically without wrapping all our prototypes JS side I personally wouldn't want this to land.

miniak:
@MarshallOfSound the only other option seems to be doing that in the native code. something like
`.SetMethod("getFileIcon", &App::GetFileIcon, true)` or `.SetPromiseMethod("getFileIcon", &App::GetFileIcon)`

miniak:
@nornagon, @deepak1556 any ideas?

PR #15952, opened 2018-12-04
chore: simplify promisify helper

codebytere:
#### Description of Change

Simplifies callbackification of methods that have been converted to promises so as to avoid excessive proliferation of:

```
const nativefn = this.fn
this.fn = function (arg, cb) {
  return deprecate.promisify(nativeFn.call(this, arg), cb)
}
```

/cc @miniak 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/lib/browser/api/app.js
+++ b/lib/browser/api/app.js
@@ -40,10 +40,7 @@ Object.assign(app, {
   }
 })
 
-const nativeGetFileIcon = app.getFileIcon
-app.getFileIcon = (path, options = {}, cb) => {
-  return deprecate.promisify(nativeGetFileIcon.call(app, path, options), cb)
-}
+app.getFileIcon = deprecate.promisify(app.getFileIcon)
 
 const nativeAppMetrics = app.getAppMetrics
 app.getAppMetrics = () => {
--- a/lib/browser/api/app.js
+++ b/lib/browser/api/app.js
@@ -40,7 +40,7 @@ Object.assign(app, {
   }
 })
 
-app.getFileIcon = deprecate.promisify(app.getFileIcon)
+app.getFileIcon = deprecate.promisify(app.getFileIcon, 2)
 
 const nativeAppMetrics = app.getAppMetrics
 app.getAppMetrics = () => {
--- a/lib/common/api/deprecate.js
+++ b/lib/common/api/deprecate.js
@@ -69,23 +69,28 @@ const deprecate = {
     })
   },
 
-  promisify: (promise, cb) => {
-    const oldName = `function with callbacks`
-    const newName = `function with Promises`
-    const warn = warnOnce(oldName, newName)
+  promisify: (fn, numArgs) => {
+    return function (...args) {
+      const promise = fn.apply(this, args.splice(0, numArgs))
+      const cb = args[numArgs]
 
-    if (typeof cb !== 'function') return promise
-    if (process.enablePromiseAPIs) warn()
-    return promise
-      .then(res => {
-        process.nextTick(() => {
-          cb(null, res)
+      const oldName = `function with callbacks`
+      const newName = `function with Promises`
+      const warn = warnOnce(oldName, newName)
+
+      if (typeof cb !== 'function') return promise
+      if (process.enablePromiseAPIs) warn()
+      return promise
+        .then(res => {
+          process.nextTick(() => {
+            cb(null, res)
+          })
+        }, err => {
+          process.nextTick(() => {
+            cb(err)
+          })
         })
-      }, err => {
-        process.nextTick(() => {
-          cb(err)
-        })
-      })
+    }
   },
 
   renameProperty: (o, oldName, newName) => {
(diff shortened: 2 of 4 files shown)


lib/browser/api/app.js
```diff
@@ -40,10 +40,7 @@ Object.assign(app, {
   }
 })
 
-const nativeGetFileIcon = app.getFileIcon
-app.getFileIcon = (path, options = {}, cb) => {
-  return deprecate.promisify(nativeGetFileIcon.call(app, path, options), cb)
-}
+app.getFileIcon = deprecate.promisify(app.getFileIcon)
```
miniak: you need to pass the number of arguments

lib/common/api/deprecate.js
```diff
@@ -69,23 +69,28 @@ const deprecate = {
...
-    return promise
-      .then(res => {
-        process.nextTick(() => {
-          cb(null, res)
-        })
-      }, err => {
-        process.nextTick(() => {
-          cb(err)
+      const oldName = `function with callbacks`
+      const newName = `function with Promises`
```
ckerr:
```js
`${fn.name} with callbacks`
`${fn.name} with Promises`
```

maybe?

lib/common/api/deprecate.js
```diff
@@ -69,23 +69,28 @@ const deprecate = {
...
+      const warn = warnOnce(oldName, newName)
```
ckerr: _(bug)_ This is creating a new `warn ` each time this return function is called -- so wouldn't a warning be given every time our wrapper is function is called, rather than just the first time?

To give the warning just once, I think we'd want a `warn` to be created each time `deprecate.promisify()` is called, *not* each time the wrapper function is called, e.g.:

```js
promisify: (fn, numArgs) => {
  const oldName = ...
  const newName= ...
  const warn = warnOnce(oldName, newName)

  return function (...args) {
    const promise= fn.apply(this, args.splice(0, numArgs))
    ...
  }
}
```

lib/common/api/deprecate.js
```diff
@@ -69,23 +69,28 @@ const deprecate = {
...
   },
 
-  promisify: (promise, cb) => {
-    const oldName = `function with callbacks`
-    const newName = `function with Promises`
-    const warn = warnOnce(oldName, newName)
+  promisify: (fn, numArgs) => {
+    return function (...args) {
+      const promise = fn.apply(this, args.splice(0, numArgs))
+      const cb = args[numArgs]
```
ckerr: _(bug)_ wouldn't this be `args[0]` after the previous line's splice?

ckerr: _(tweak)_ Also, since we've coded this to the assumption that `cb` always the final parameter... why not just `.pop()` it off of args and drop the `numArgs` parameter, e.g.

```
promisify: fn => {
  return function (...args) {
    const cb = args.pop();
    const promise = fn.apply(this, args);
```

miniak: it's not fully equivalent to the original callback based function, for example if you have `contents.printToPDF(options, callback)` and call it with `contents.printToPDF({}, () => console.log('done'), () =>  console.log('should not be called'));`, then the extra arguments are not ignored as they were before.

ckerr: OK, that's fair.

Is that something that happens in practice? Should we warn when people pass too many arguments?

lib/common/api/deprecate.js
MarshallOfSound: @ckerr `fn.name` isn't set for anonymous function, should defs default to `function` if we do want to use `fn.name`

lib/common/api/deprecate.js
MarshallOfSound: @ckerr Passing too many arguments is not considered an error / warning.  It can happen if you're doing things like blindly spreading args `...args`.  We do already error out in native land if you provide invalid arguments.

lib/common/api/deprecate.js
codebytere: ☝️ we don't have access to `fn.name`, which is why i did it in the above manner :)

lib/common/api/deprecate.js
```diff
@@ -69,23 +69,28 @@ const deprecate = {
...
-      .then(res => {
-        process.nextTick(() => {
-          cb(null, res)
-        })
-      }, err => {
-        process.nextTick(() => {
-          cb(err)
+    return function (...args) {
+      const promise = fn.apply(this, args.splice(0, numArgs))
+      const cb = args[numArgs]
```
ckerr: _(bug)_ wouldn't this be `args[0]` after the previous line's splice?

And if that's correct, that `cb` and `args[numArgs]` don't point to a callback after the splice -- how did this PR pass CI, as the callback would never be invoked? Do we test `deprecate.promisify()` to confirm that the callback is called?

lib/common/api/deprecate.js
ckerr: We appear to have it sometimes, e.g. we have it in the spec tests.

I've changed it to `fn.name || 'function'`

PR #13028, opened 2018-05-21
feat: security: add an option to disable the remote module

miniak:
##### Description of Change

The `webPreferences.sandbox` option is not sufficient to prevent fully exploitation if the remote module still allows access to `fs` and other modules via the main process. This proposal adds a new option `webPreferences.enableRemoteModule`, which blocks access to the remote module on the main process side.

- On the renderer process side, it removes access to `electron.remote`, which allows code to check whether the app can use `remote` beforehand
- On the main process side, the `rpc-server.js` ignores all IPC messages implementing the `remote` module

##### Checklist
- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

##### Release Notes
Notes: Added `webPreferences.enableRemoteModule` option allowing to disable the remote module to increase sandbox security.

PR #15960, opened 2018-12-05
refactor: use helpers when using the remote module in sandboxed renderers

miniak:
#### Description of Change
Use helpers for any remaining uses of modules, which are using `remote.remote` internally or the `remote` module itself, as it can be disabled.

Follow up to https://github.com/electron/electron/pull/13028.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/renderer/extensions/i18n.js
+++ b/lib/renderer/extensions/i18n.js
@@ -6,9 +6,10 @@
 // Does not implement predefined messages:
 // https://developer.chrome.com/extensions/i18n#overview-predefined
 
+const { nativeRequire } = require('@electron/internal/renderer/remote')
 const ipcRenderer = require('@electron/internal/renderer/ipc-renderer-internal')
-const fs = require('fs')
-const path = require('path')
+const fs = nativeRequire('fs')
+const path = nativeRequire('path')
 
 let metadata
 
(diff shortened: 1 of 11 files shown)


lib/renderer/extensions/i18n.js
```diff
@@ -6,9 +6,10 @@
 // Does not implement predefined messages:
 // https://developer.chrome.com/extensions/i18n#overview-predefined
 
+const { nativeRequire } = require('@electron/internal/renderer/remote')
```
nornagon: why is this called `nativeRequire`...? it isn't native require.

miniak: @miniak I agree, it’s a bad name. Any suggestions? I can rename it in another PR.

nornagon: Perhaps `potentiallyRemoteRequire`? `requirePrivileged`? `unsandboxedRequire`?

miniak: `potentiallyRemoteRequire` it is, sounds scary as it should, to deter people from using it

miniak: https://github.com/electron/electron/pull/15971

PR #15966, opened 2018-12-06
feat: Upgrade to Chromium 71.0.3578.98

deepak1556:
#### Description of Change

Refs https://github.com/orgs/electron/projects/26

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Upgrade to Chromium 71.0.3578.98


patches/common/chromium/enable_widevine.patch
```diff
@@ -8,15 +8,18 @@ Electron needs that flag to be enabled on those paltforms,
...
+@@ -6,9 +6,8 @@ import("//build/config/chrome_build.gni")
+ import("//media/media_options.gni")
  
  declare_args() {
-   # Allow Widevine key system support in Chromium.
--  enable_widevine = false
+-  # Enables Widevine key system support. Enabled by default in Google Chrome or
+-  # on Android. Can be optionally enabled in Chromium.
+-  enable_widevine = is_chrome_branded || is_android
++  # Allow Widevine key system support in Chromium.
```
nornagon: nit: we don't need to patch this comment

zcbenz:
I'm rebasing this branch to fix conflict with webview fixes.

patches/common/chromium/xf86drm_include_sys_sysmacros_h.patch
```diff
@@ -0,0 +1,20 @@
...
+diff --git a/third_party/libdrm/BUILD.gn b/third_party/libdrm/BUILD.gn
+index 97b6505b20c70ca9c1023b28cc3f47d4e6708e76..bfc193c6d0e6da93c58d27c0530892b334dfb4d0 100644
+--- a/third_party/libdrm/BUILD.gn
++++ b/third_party/libdrm/BUILD.gn
+@@ -40,6 +40,8 @@ if (!use_system_libdrm) {
+   }
+ 
+   static_library("libdrm") {
++    # Include <sys/sysmacros.h> for major, minor and mkdev.
++    defines = [ "MAJOR_IN_SYSMACROS" ]
```
nornagon: I think this should be unnecessary with the proper sysroot (https://github.com/electron/debian-sysroot-image-creator/pull/9)

deepak1556: Yeah it should be the case, since upstream builds fine without it. Was just testing in hope that it might be the only build error :) Feel free to remove it when updating the sysroots. Thanks!

BUILD.gn
```diff
@@ -54,55 +54,40 @@ config("branding") {
...
-    "-r",
-    "./lib/sandboxed_renderer/api/exports/electron.js:electron",
-    "-t",
-    "aliasify",
+  outputs = [
+    "$target_gen_dir/js2c/preload_bundle.js",
   ]
 
-  inputs = [
+  args = [
```
nornagon: why'd `inputs` disappear here? it's necessary to tell GN when it needs to rebuild.

BUILD.gn
```diff
@@ -54,55 +54,40 @@ config("branding") {
...
   ]
 
-  script = "browserify"
-  args = inputs + [
-           "-t",
-           "aliasify",
-           "-o",
-           rebase_path(outputs[0]),
-         ]
+  args = [
```
nornagon: here too, `inputs` is a special GN name and not just fluff. without it GN won't know when it needs to rebuild this target.

atom/browser/api/atom_api_web_contents.cc
```diff
@@ -1599,15 +1600,20 @@ void WebContents::ReplaceMisspelling(const base::string16& word) {
 uint32_t WebContents::FindInPage(mate::Arguments* args) {
   uint32_t request_id = GetNextRequestId();
   base::string16 search_text;
-  blink::WebFindOptions options;
+  mate::Dictionary dict;
```
nornagon: move this down to just before where it's used?

atom/renderer/atom_sandboxed_renderer_client.cc
```diff
@@ -182,30 +183,21 @@ void AtomSandboxedRendererClient::DidCreateScriptContext(
       !IsDevToolsExtension(render_frame))
     return;
 
+  // Wrap the bundle into a function that receives the binding object as
+  // argument.
```
nornagon: is this comment still accurate?

atom/renderer/atom_renderer_client.cc
```diff
@@ -189,28 +188,20 @@ void AtomRendererClient::SetupMainWorldOverrides(
     v8::Handle<v8::Context> context,
     content::RenderFrame* render_frame) {
   // Setup window overrides in the main world context
-  v8::Isolate* isolate = context->GetIsolate();
-  v8::HandleScope handle_scope(isolate);
-  v8::Context::Scope context_scope(context);
-
   // Wrap the bundle into a function that receives the isolatedWorld as
   // an argument.
```
nornagon: is this comment still accurate?

spec/api-session-spec.js
```diff
@@ -587,7 +587,7 @@ describe('session module', () => {
       server.listen(0, '127.0.0.1', () => {
         const config = { pacScript: `http://127.0.0.1:${server.address().port}` }
         customSession.setProxy(config, () => {
-          customSession.resolveProxy('http://localhost', (proxy) => {
+          customSession.resolveProxy('https://google.com', (proxy) => {
```
nornagon: ⚠️ our tests shouldn't rely on external network connections

spec/fixtures/native-addon/echo/binding.cc
```diff
@@ -0,0 +1,38 @@
+#include <js_native_api.h>
+#include <node_api.h>
+
+namespace {
+
+napi_value Print(napi_env env, napi_callback_info info) {
```
nornagon: Why is this called `Print`? Perhaps it should be called `Echo`?

spec/fixtures/pages/storage/local_storage.html
```diff
@@ -1,11 +1,8 @@
...
-  const {ipcRenderer} = require('electron')
-
-  let message = 'Ok'
   try {
     window.localStorage
   } catch (e) {
-    message = e.message
+    const {ipcRenderer} = require('electron')
+    ipcRenderer.send('local-storage-response', e.message)
   }
```
nornagon: This looks like it won't send a message back if it doesn't throw, which will just make the test time out. Let's make tests fail fast & with useful error messages rather than dealing with yet another "test timed out" failure.

spec/api-session-spec.js
deepak1556: This doesn't hit the network, only verifies the result from proxy resolver. Also `localhost` cannot be used with PAC tests because of https://chromium-review.googlesource.com/c/chromium/src/+/1324880

PR #15969, opened 2018-12-06
fix: use appropriate site instance for cross-site nav's (backport: 4-0-x)

### LINKED ISSUE #8100: In sandbox mode, window.opener doesn't work when children are from different domain
* Electron version: 1.4.10
* Operating system: Darwin x64

### Expected behavior
When a popup pointed to a different domain from the parent is opened. window.opener should not be null. It is required for many OAuth authentication where the parent and the child (popup) is from different domains.

### Actual behavior

When a popup is pointed towards the same domain window.opener is present but when its pointed towards a different domain window.opener is null

### How to reproduce

**main.js**
`app.on('ready', () => { console.log("ready"); const win = new BrowserWindow({webPreferences: { sandbox: true }}) win.loadURL('http://localhost:8888/index.html'); })`

**index.html**
`<a href="javascript:window.open('https://app.asana.com')">Click here</a>`

window.opener is null at app.asana.com because the domain is different from the parent

ppontes:
#### Description of Change
This is a backport of the fix merged to master: https://github.com/electron/electron/commit/d5d1fa8290015d6f0ddf25a00f27f6edec631657.

When a window is opened with window.open to navigate to a cross-site destination, the new site instance should belong to the browsing instance of the opener window in order not to break scripting relationships between windows.

fixes #8100.

This change turns Chromium's site isolation on and depends on it being on in order to work properly. Turning on site isolation revealed the following latent problems that are / will be fixed also in this PR:
- [X] If a cross-site redirection happens during a navigation, the renderer process that will handle the final URL will be locked to the pre-redirection URL with Chromium's child process security. If the page tries to access local resources, Chromium will refuse and kill the renderer process.
- [X] Chromium now kills renderer processes that attempt to access resources for which they have no permissions. This means that the "custom non standard schemes cannot access localStorage" UT had to be adapted.

The fix also requires the following to be addressed:

- [x] If a cross-site redirection happens during a navigation, Chromium discards the speculative RFH that was created for the pre-redirection URL. This is harmless in Chromium but Electron reacts by asynchronously closing the associated BrowserWindow. User sees the new window navigating to the final URL and then being closed.

#### Checklist
- [x] `npm test` passes

#### Release Notes

Notes: Fixed a bug where `window.opener` of a window created with window.open from a sandboxed renderer was null.

### DIFF
--- a/atom/browser/api/atom_api_web_contents.cc
+++ b/atom/browser/api/atom_api_web_contents.cc
@@ -788,8 +788,29 @@ void WebContents::RenderViewCreated(content::RenderViewHost* render_view_host) {
     impl->disable_hidden_ = !background_throttling_;
 }
 
+void WebContents::RenderViewHostChanged(content::RenderViewHost* old_host,
+                                        content::RenderViewHost* new_host) {
+  currently_committed_process_id_ = new_host->GetProcess()->GetID();
+}
+
 void WebContents::RenderViewDeleted(content::RenderViewHost* render_view_host) {
+  // This event is necessary for tracking any states with respect to
+  // intermediate render view hosts aka speculative render view hosts. Currently
+  // used by object-registry.js to ref count remote objects.
   Emit("render-view-deleted", render_view_host->GetProcess()->GetID());
+
+  if (-1 == currently_committed_process_id_ ||
+      render_view_host->GetProcess()->GetID() ==
+          currently_committed_process_id_) {
+    currently_committed_process_id_ = -1;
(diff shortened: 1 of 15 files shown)


ppontes:
@ckerr @MarshallOfSound , can you please check and merge this PR if you approve it? We'd like to have as much beta exposure to this fix as possible. Thanks!

codebytere:
REVERT: https://github.com/electron/electron/pull/15995

Issue #14319, opened 2018-08-26
"idle" electron periodically pegs the cpu on a single core vm


bughit:
* Output of `node_modules/.bin/electron --version`: v3.0.0-beta.7 x86
* Operating System (Platform and Version): win7 sp1 (x86 & x64) single core

**Expected Behavior**
no cpu pegging when electron is idle

**Actual behavior**
cpu is is periodically pegged
![image](https://user-images.githubusercontent.com/895588/44633561-1426b800-a95b-11e8-9f65-35f98f4c00e0.png)

**To Reproduce**
This is triggered by simply creating winston.transports.File.  app.disableHardwareAcceleration contributes as well.  Single core is required to repro.  I first discovered it on a single core x86 win7 vm, could not repro on a 4 core x64 vm, but lowering the cores to 1, reproduced.

AFAICT from some profiling, no js is running, electron should be idle.  Does not happen in node.

run he following app (npm install first)

##### package.json
```json
{
  "dependencies": {
    "winston": "^3.0.0"
  }
}
```

##### index.js
```js
'use strict';

const winston = require('winston'),
      {app} = require('electron');

app.disableHardwareAcceleration();

const winstonTransports = [
  new winston.transports.File({
    filename: 'test.log',
    maxsize: 10 * 2**20,
    maxFiles: 2,
    tailable: true,
  }),
];
```

sometimes the initial spike is brief and you have to wait for the next

bughit:
repro can be simplified, winston can be replaced by simply doing async file io

```js
'use strict';

require('electron').app.disableHardwareAcceleration();

const fs = require('fs'),
      fd = fs.openSync('test.log', 'a');


log();
setInterval(log, 250);

function log() {
  fs.appendFile(fd, 'log line\n', error => {});
}
```

on one single core vm this pegs the cpu continuously, on another periodically.  I thought disableHardwareAcceleration makes this worse (easier to repro), not sure, leaving it in, just in case.

@zcbenz Node is ok, so it seems like a serious electron bug.

bughit:
@zcbenz Could you please give some guidance on this. Our code would be running on customer vms and some of them are single core, this is a show stopper for us.  I also found that sync or async stdin reading triggers this ( `fs.readFileSync(0, 'utf8')`)

[...]

Issue #12026, opened 2018-02-23
Windows single core 100% CPU lockup (possible fix)


gshively11:
* Electron version: All?
    * problem reproduced on v1.4.12, v1.8.2, and v2.0.0-beta.1
    * fix tested on v1.4.12
* Operating system: Windows Server 2012 (single core)

I know this has been reported multiple times, but I wanted to start a new thread to discuss a potential fix.

I'm running a simple script to reproduce the problem, inspired by another issue describing the problem (see below).

### Expected behavior

I would expect the node/electron processes to make these http requests as fast as possible, alternating log messages `END FIRST` and `END SECOND` on the console.

### Actual behavior

Intermittently, sometimes right away, sometimes after a few seconds, the node.exe process drops to 0% CPU and the main electron.exe process hits 95%-100% cpu for a significant amount of time. The log messages stop, meaning the http requests are no longer firing. Eventually, the electron.exe process starts freeing up CPU, and the node.exe picks back up firing requests.

### How to reproduce
Run this with `electron.exe`:

```js
const http = require("http");

const url = "http://127.0.0.1:8081"; // running the http-server module

function callMe() {
  http.request({
      host: '127.0.0.1',
      port: '8081',
      timeout: 999999999,
    }, function() {
      console.log("END FIRST");
      http.request({
          host: '127.0.0.1',
          port: '8081',
          timeout: 999999999,
        }, function() {
          console.log("END SECOND");
          callMe();
        })
        .on('error', function(err) { console.log(err); })
        .end();
    })
    .on('error', function(err) { console.log(err); })
    .end();
}

callMe();
```

Now on to the fun part.  I don't pretend to fully understand the inner workings of electron, node, and chromium, but after digging in for a few days I stumbled upon a change that appears to fix the problem.  I changed [this line](https://github.com/electron/node/blob/811cfe3fcd360179d3dd436e3d80e1b045adf633/deps/uv/src/win/core.c#L133) to this:
```
loop->iocp = CreateIoCompletionPort(INVALID_HANDLE_VALUE, NULL, 0, 2); // last arg changed from 1 to 2
```
According to the [documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363862(v=vs.85).aspx), that last arg sets
> The maximum number of threads that the operating system can allow to concurrently process I/O completion packets for the I/O completion port. This parameter is ignored if the ExistingCompletionPort parameter is not NULL.

[...]

Issue #11699, opened 2018-01-23
Electron hangs and leads to 100% CPU usage after calling callback via uv_async_send on a single-core machine


alexey-ambrozevitch:
* Electron version: 1.7.8
* Operating system: Windows 7 x64 Enterprise

### Expected behavior
Electron processes callback dispatched by uv_async_send function in the normal way.

### Actual behavior
Electron hangs and leads to 100% CPU usage after calling callback via uv_async_send.

### How to reproduce
The problem can be reproduced only on a single-core machine.

zcbenz:
Can you make a minimal reproducible example? `uv_async_send` is heavily used in native modules and Electron itself, the problem is probably on other things.

alexey-ambrozevitch:
We implement our own node.js module. It has to initiate simple JavaScript function call. So first of all we initialize loop by uv_default_loop(), then initialize uv_async_t(), make call uv_async_init with initialized parameters and then make uv_async_send. After that Electron root process begins to consume 100% CPU and our JavaScript procedure does not implement (callback, passed to uv_async_init as the third parameter does not begin at all). Then if click on UI or make some actions with it, CPU slows down  and our procedure begins to implement.

alexey-ambrozevitch:
After investigation of the source code of Electron, it was turned out that the problem was caused by atom::NodeBindingsWin::PollEvents method. The call of ::GetQueuedCompletionStatus on thread that differs from libuv poll leaded to errors inside uv_run API. After replacing the pair of IOCP functions to polling on uv_loop_alive, the problem had been sorted out.

alexey-ambrozevitch:
Also, could you please provide any information whether the fix with uv_loop_alive will not cause any side effects.

gshively11:
@alexey-ambrozevitch I'm tracking down a similar issue, can you paste the changes you made to get it working, please?

bughit:
[minimal reproducible example](https://github.com/electron/electron/issues/14319#issuecomment-417138296) that @zcbenz asked for

zcbenz:
Fixed by https://github.com/electron/electron/pull/15975.

PR #15975, opened 2018-12-07
fix: allow 2 threads for CreateIoCompletionPort on single-core to prevent busy looping

bughit:
#### Description of Change

Electron on single-core windows has a problem with cpu pegging when it should be idle, triggered by (among other things) async io.

- electron/electron#12026
- electron/electron#11699
- electron/electron#14319

There are related issues in downstream projects like vscode.

This fix replaces the I/O completion port created by libuv with one with 2 concurrent threads, on single core.

Original idea came from #12026.
@refack recommended to do the iocp override in electron code.
CCing @zcbenz for review

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixes cpu pegging on single-core windows.

### DIFF
--- a/atom/common/node_bindings_win.cc
+++ b/atom/common/node_bindings_win.cc
@@ -7,6 +7,7 @@
 #include <windows.h>
 
 #include "base/logging.h"
+#include "base/sys_info.h"
 
 namespace atom {
 
@@ -14,8 +15,7 @@ NodeBindingsWin::NodeBindingsWin(BrowserEnvironment browser_env)
     : NodeBindings(browser_env) {
   // on single-core the io comp port NumberOfConcurrentThreads needs to be 2
   // to avoid cpu pegging likely caused by a busy loop in PollEvents
-  char* numProcessors = getenv("NUMBER_OF_PROCESSORS");
-  if (numProcessors && strcmp(numProcessors, "1") == 0) {
+  if (base::SysInfo::NumberOfProcessors() == 1) {
     if (uv_loop_->iocp && uv_loop_->iocp != INVALID_HANDLE_VALUE)
       CloseHandle(uv_loop_->iocp);
     uv_loop_->iocp = CreateIoCompletionPort(INVALID_HANDLE_VALUE, NULL, 0, 2);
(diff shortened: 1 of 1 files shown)


atom/common/node_bindings_win.cc
```diff
@@ -11,7 +11,16 @@
 namespace atom {
 
 NodeBindingsWin::NodeBindingsWin(BrowserEnvironment browser_env)
-    : NodeBindings(browser_env) {}
+    : NodeBindings(browser_env) {
+  // on single-core the io comp port NumberOfConcurrentThreads needs to be 2
+  // to avoid cpu pegging likely caused by a busy loop in PollEvents
+  char* numProcessors = getenv("NUMBER_OF_PROCESSORS");
```
nornagon: This isn't the right way to test whether we're single-core. Use `base::SysInfo::NumberOfProcessors()`, declared in `base/sys_info.h`.

refack: But it makes for easy testing ;)

bughit:
updated with base::SysInfo::NumberOfProcessors and CHECKS

bughit:
@refack thanks for the help with this

refack:
> masking an underlying issue with some resource being accessed on multiple threads that should only be accessed on a single thread

I'm still investigation that... But I'm not too concerned since I saw the `NodeBindingsWin::PollEvents()` implementation.

bughit:
@zcbenz

- Can this be merged? Is there something I need to do? I think ci failures are not related to this change.
- Since this is a bug fix, is backporting to 4-0-x allowed?

zcbenz:
If no opposition, I would like to merge this PR because:
1. it only affects single-core machines;
2. even if something wrong happens, it won't make things worse since Electron is already very broken on single-core machines with the bug.

brenca:
Just out of curiosity, would this bug repro in node? Could this be eventually upstreamed?

bughit:
> Just out of curiosity, would this bug repro in node?

https://github.com/electron/electron/issues/14319#issuecomment-445079579

> Could this be eventually upstreamed?

https://github.com/electron/electron/issues/14319#issuecomment-445233411
https://github.com/libuv/libuv/pull/2106#issuecomment-445565889

bughit:
3-0-x?

bughit:
@zcbenz Why not 3-0-x?  You are still doing 3.0 releases, doesn't make sense that it would be fixed in 2.0 and not 3.0.

BinaryMuse:
@bughit I believe this was an oversight; I'll get this into the next 3-0-x release.

PR #15987, opened 2018-12-10
test: skip test for not supported setLoginItemSettings on mas platform

neo291:
#### Description of Change
The `Browser::SetLoginItemSettings` is not supported on mas and tests that are executed on this feature always fails.
To avoid this I've disabled the `Browser::SetLoginItemSettings` related tests on mas platform.
This is the `Browser::SetLoginItemSettings` function source code inside `electron/atom/browser/browser_mac.mm` on line 279:
```
void Browser::SetLoginItemSettings(LoginItemSettings settings) {
#if defined(MAS_BUILD)
  if (!platform_util::SetLoginItemEnabled(settings.open_at_login)) {
    LOG(ERROR) << "Unable to set login item enabled on sandboxed app.";
  }
#else
  if (settings.open_at_login)
    base::mac::AddToLoginItems(settings.open_as_hidden);
  else {
    RemoveFromLoginItems();
  }
#endif
}
```

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/spec/api-app-spec.js
+++ b/spec/api-app-spec.js
@@ -450,7 +450,7 @@ describe('app module', () => {
     ]
 
     before(function () {
-      if (process.platform === 'linux') {
+      if (process.platform === 'linux' || process.mas) {
         this.skip()
       }
     })


codebytere:
@neo291 yes, `Browser::SetLoginItemSettings` is supported on MAS. It uses a different approach than does non-sandboxed normal MacOS build. It takes longer to apply the setting changes using the Service Management framework, but it works.

ckerr:
@neo291 is right that this test fails a whole lot though. @codebytere any thoughts on how to make the test give a false error less often?

neo291:
@codebytere and @ckerr, thanks for the explanation, now is more clear the reason because it often fails.
In this case I think that I can proceed to close this pull request? Or you prefer to take it open to find a possible way to improve this false positive?
Let me now know if I've something to do and thanks again. 😄

codebytere:
@neo291 i'm going to close this for now but if you have some ideas about how to reduce the flakiness of this PR we'd love to chat and would welcome a PR!

PR #15994, opened 2018-12-10
chore: disable bump-version on arm32

codebytere:
#### Description of Change

Dugite doesn't work on arm32, and we don't run bump-version on arm32 anyway so we can disable the associated tests on that arch.

/cc @jkleinsc 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/spec/version-bump-spec.js
+++ b/spec/version-bump-spec.js
@@ -6,7 +6,7 @@ describe('bump-version script', () => {
   const betaPattern = /[0-9.]*(-beta[0-9.]*)/g
 
   before(function () {
-    if (process.platform === 'linux' && process.arch === 'arm32') {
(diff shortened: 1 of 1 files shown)


spec/version-bump-spec.js
```diff
@@ -5,6 +5,12 @@ describe('bump-version script', () => {
   const nightlyPattern = /[0-9.]*(-nightly.(\d{4})(\d{2})(\d{2}))$/g
   const betaPattern = /[0-9.]*(-beta[0-9.]*)/g
 
+  before(function () {
+    if (process.platform === 'linux' && process.arch === 'arm32') {
```
jkleinsc: Should be process.arch === 'arm': https://nodejs.org/api/process.html#process_process_arch

Issue #15911, opened 2018-12-01
BrowserWindow.getNativeWindowHandle() returns same value in different BrowserWindow


KochiyaOcean:
* Output of `node_modules/.bin/electron --version`: 4.0.0-beta.8
* Operating System (Platform and Version): Windows 10 1803
* Output of `node_modules/.bin/electron --version` on last known working Electron version (if applicable): 4.0.0-beta.7

**Expected Behavior**
`BrowserWindow.getNativeWindowHandle()` should return different value for different window

**Actual behavior**
`BrowserWindow.getNativeWindowHandle()` returns same value for different window, which is not true nativeWindowHandle for any of them.

**To Reproduce**
Open two `BrowserWindow`, and run `getNativeWindowHandle` on both `BrowserWindow`
```js
const { remote } = require('electron')
w1 = new remote.BrowserWindow()
w2 = new remote.BrowserWindow()
console.log(w1.getNativeWindowHandle(), w2.getNativeWindowHandle())
```
![image](https://user-images.githubusercontent.com/8194131/49328509-7411a800-f5b5-11e8-9a24-827aa589d937.png)

estoykov:
Probably it is caused by these fixes:
https://github.com/electron/electron/pull/15521
https://github.com/electron/electron/pull/15644
which implementation is wrong for both Windows and MacOS.
- for Windows it is wrong because the returned pointer to the HWND is somewhere in the stack, i.e. will point to some garbage at the time when it will be used
- for MacOS returned pointer is the direct pointer to the NSView, and the first 8 bytes from NSView internal structure will copied, not the pointer to the NSView

estoykov:
This is still broken in 3.1.0, but fixed in 4.x.x. Can we backport the fix in 3.1.0?

nikolovp:
I am also interested in cherry picking this fix into 3.*

alexeykuzmin:
@estoykov @nikolovp 
The fix will be included in the next 3.1.x release.

PR #16014, opened 2018-12-11
fix: return pointer instead of pointer's content

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/15911.

I have written a module to validate that the result of `BrowserWindow.getNativeWindowHandle` is a valid window: https://github.com/electron/node-is-valid-window.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix `BrowserWindow.getNativeWindowHandle()` returning invalid handle.


alexeykuzmin:
@zcbenz 
The are reports that #15911 also affects the `3.1.x` line.
Could you please backport the fix to the `3-1-x` branch?

PR #16028, opened 2018-12-12
fix: fix an original-fs regression introduced in 4.0 betas

### LINKED ISSUE #15543: `createWriteStream` in `original-fs` is broken
* Output of `node_modules/.bin/electron --version`: v4.0.0-beta.6
* Operating System (Platform and Version): Windows 10 x64 Build 17134
* Output of `node_modules/.bin/electron --version` on last known working Electron version (if applicable): v3.0.7

**Expected Behavior**
`require('original-fs').createWriteStream('some/asar/file.asar')` should not error.

**Actual behavior**
The write stream fails with error: `Error: Invalid package some/asar/file.asar`

**To Reproduce**
I've forked `electron-quick-start` with the most trivial repro: https://github.com/adill/electron-quick-start/tree/demo-create-write-stream

ckerr:
#### Description of Change

Fixes a 4.0.0-beta.X regression in `original-fs`. Node's `fs` and `internal/fs/streams` have a lazy-loaded circular dependency, so to expose the unmodified Node `fs` functionality, we have to copy both `fs` *and* `internal/fs/streams` and modify the copies to depend on each other instead of on our asarified 'fs' code.

Fixes #15543.

Discussion: an earlier draft of this code added a new `original_fs` module into atom/common/node-bindings so that multiple Node code that calls `process.binding('fs')` could also be swapped out with a `process.binding('fs_original')`. This more completely separates the asarified and original versions of `fs`, but it also requires more cloning and inter-dependencies, and is not required for #15543. This PR tries to KISS by replacing only what's clearly needed.

CC @codebytere 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fix 4-0-x regression that broke original-fs

### DIFF
--- a/lib/common/asar_init.js
+++ b/lib/common/asar_init.js
@@ -8,10 +8,6 @@
     // dependency. So to expose the unmodified Node 'fs' functionality here,
     // we have to copy both 'fs' *and* 'internal/fs/streams' and modify the
     // copies to depend on each other instead of on our asarified 'fs' code.
-    // source['original_fs'] = source.fs.replace(/'internal\/fs\/streams'/g, "'original_fs/internal/streams'")
-    // source['original_fs/internal/streams'] = source['internal/fs/streams'].replace(/'fs'/g, "'original_fs'")
-    // source['original-fs'] = source['original_fs']
-
     source['original-fs'] = source.fs.replace("require('internal/fs/streams')", "require('original-fs/streams')")
     source['original-fs/streams'] = source['internal/fs/streams'].replace("require('fs')", "require('original-fs')")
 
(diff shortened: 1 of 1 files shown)


ckerr:
Hmm, adding on a CC to @MarshallOfSound since he appears to be online sometimes despite it being his holiday. :smile: 

This PR works, and I believe avoids the [previous PR's](https://github.com/electron/node/pull/79) incorrect dependency chain, but I'd like a second opinion on whether or not this is enough or if we should pull `process.binding('fs')` replacements into this patch as well?

ckerr:
Sounds good. Glad to have a sanity check on the priority on the `process.binding()` usage. Thanks :D

ckerr:
The linter bot seems to be hiccuping here, but the change was already validate by our precommit hook, so merging despite the hiccup.

jkleinsc:
Lint passed: https://github.visualstudio.com/electron/_build/results?buildId=25393

PR #16040, opened 2018-12-12
chore: add deprecation warning for contextIsolation default change

nornagon:
#### Description of Change
In 5.0, the default for `contextIsolation` will be `true`. This only affects you if you're using the `preload` option to interact directly with globals on the `window` or `document` objects.

#### Checklist

#### Release Notes

Notes: no-notes


nornagon:
it'd be great to be able to link to a place that explains what `contextIsolation` does, why it does it, and how to work with it enabled.

codebytere:
@nornagon agree, i think that calls to a slightly larger issue of holes in our documentation around concepts like that

nornagon:
CI failure is unrelated, merging

PR #16042, opened 2018-12-13
fix: do not allow the window to grab focus when tabbing / shift+tabbing

### LINKED ISSUE #12919: Browser windows have an extra tab focus which can not be disabled and is not visible
* Electron Version: 2.0.0
* Last known working Electron version:  N/A

**Expected Behavior**
If the last focusable HTML element is focused and I press 'tab', then the next thing that is focused should either be the first focusable HTML element, or if not, then whatever is focused should have a visible focus.

**Actual behavior**
Something is being focused (I think it is the entire HTML document?) and there is no visual indication as to what it is. It is fine that this other thing is being focused, what is not fine is that I/a user does not know where the keyboard focus has gone.

**To Reproduce**
- Checkout the [electron-quick-start](https://github.com/electron/electron-quick-start)- Put an `<input type="text"/>` before the `renderer.js` script
- Launch quick start
- Set focus in the input
- Press 'tab'

ppontes:
#### Description of Change
Keep the focus restrained to the set of focusable elements when the user tabs or SHIFT + tabs through the focusable elements in a page. Do not allow it to move to the browser window as it is unexpected in a desktop app and no visual indication is given that the browser window is focused.

Fixes #12919.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a problem where the focus would move to the browser window after the user (SHIFT +) tabbed through all the elements in the page (#12919).

### DIFF
--- a/spec/chromium-spec.js
+++ b/spec/chromium-spec.js
@@ -1411,6 +1411,111 @@ describe('chromium feature', () => {
       await new Promise((resolve) => { utter.onend = resolve })
     })
   })
+
+  describe('focus handling', () => {
+    let webviewContents = null
+
+    beforeEach(async () => {
+      w = new BrowserWindow({
+        show: true
+      })
+
+      const webviewReady = emittedOnce(w.webContents, 'did-attach-webview')
+      await w.loadFile(path.join(fixtures, 'pages', 'tab-focus-loop-elements.html'))
+      const [, wvContents] = await webviewReady
+      webviewContents = wvContents
+      await emittedOnce(webviewContents, 'did-finish-load')
+      w.focus()
+    })
+
+    afterEach(() => {
+      webviewContents = null
+    })
+
+    const expectFocusChange = async () => {
+      const [, focusedElementId] = await emittedOnce(ipcMain, 'focus-changed')
+      return focusedElementId
+    }
+
+    describe('a TAB press', () => {
+      const tabPressEvent = {
+        type: 'keyDown',
+        keyCode: 'Tab'
+      }
+
+      it('moves focus to the next focusable item', async () => {
+        let focusChange = expectFocusChange()
+        w.webContents.sendInputEvent(tabPressEvent)
+        let focusedElementId = await focusChange
+        assert.strictEqual(focusedElementId, 'BUTTON-element-1', `should start focused in element-1, it's instead in ${focusedElementId}`)
+
+        focusChange = expectFocusChange()
+        w.webContents.sendInputEvent(tabPressEvent)
+        focusedElementId = await focusChange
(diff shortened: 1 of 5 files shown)


spec/chromium-spec.js
```diff
@@ -1411,6 +1411,83 @@ describe('chromium feature', () => {
...
+      const [, focusedElementId, focusedElementTag] = await focusedElementIdResult
+      return [focusedElementId, focusedElementTag]
+    }
+
+    const getFinalFocusedElementId = async () => {
+      const [focusedElementId, focusedElementTag] = await getCurrentFocusedElement()
+      if (focusedElementTag === 'BODY') {
+        // Wait for Electron to react and move the focus back
+        // to the first / last element if necessary.
+        const waitTimeMs = 10
```
ppontes: Hate this magic constant, but the test has to wait a bit until Electron reacts to Chromium's notification that the focus has gone through all the focusable elements and sets it back to the first or last element. Any idea to implement this in a better way?

spec/chromium-spec.js
```diff
@@ -1411,6 +1411,83 @@ describe('chromium feature', () => {
       await new Promise((resolve) => { utter.onend = resolve })
     })
   })
+
+  describe('focus handling', () => {
```
ppontes: Is chromium-spec.js the best place for this test?

bpasero:
Very cool, looking forward to have this in VSCode 👍

spec/chromium-spec.js
deepak1556: Yup it looks good since we aren't testing any webContent specific apis.

spec/chromium-spec.js
deepak1556: Rather than requesting the currently focused element, the loaded page can signal when an element receives focus and the test can validate with it.

spec/fixtures/pages/tab-focus-loop-elements.html
```diff
@@ -0,0 +1,22 @@
...
+    })
+  </script>
+  <body>
+    <div>
+      <button id="element-1">Button 1</button>
+      <button id="element-2">Button 2</button>
+      <button id="element-3">Button 3</button>
+    </div>
+  </body>
+</html>
```
deepak1556: nit: new line missing at EOF

spec/chromium-spec.js
ppontes: Excellent idea, changed accordingly.

spec/fixtures/pages/tab-focus-loop-elements.html
ppontes: Fixed, thanks.

ppontes:
@deepak1556 Thanks for the review and great suggestion to include a webview in the tests, I had tested it manually but we should ensure it doesn't regress. It has been included in the tests..

PR #16043, opened 2018-12-13
docs: contextIsolation is no longer experimental

BinaryMuse:
#### Description of Change

This removes the note from the `BrowserWindow` docs that says that `contextIsolation` is experimental in preparation for the 4.0 release.

Should we also reword the explanation of the default since it's changing for 5.0?

/cc @nornagon @electron/docs 

#### Checklist

#### Release Notes

Notes: no-notes


jkleinsc:
Merging as arm64 failure has nothing to do with doc change.

Issue #15789, opened 2018-11-21
Notifications posted by systemPreferences.postNotification are not delivered immediately


turbobabr:
* Output of `node_modules/.bin/electron --version`: Affects all Electron versions
* Operating System (Platform and Version): macOS 10.13.6
* Output of `node_modules/.bin/electron --version` on last known working Electron version (if applicable): Affects all Electron versions

**Expected Behavior**

Notifications posted by [systemPreferences.postNotification](https://electronjs.org/docs/api/system-preferences#systempreferencespostnotificationevent-userinfo-macos) should be delivered to subscribers immediately no matter whether subscriber app is active or not.

**Actual behavior**

Notifications are suspended until the receiving app gets activated.

**To Reproduce**

Fiddle:
- https://gist.github.com/turbobabr/e4021ab39126e3afc8ea1cc01e99b5d0

(can be opened by [Electron Fiddle](https://github.com/electron/fiddle))

**Screenshots**

A tiny screencast explaining the problem (using the same electron app as a notification subscriber and emulating issue by deactivating the app and posting notification using global shortcut) :
![screen recording 2018-11-21 12 38 23 copy](https://user-images.githubusercontent.com/6356526/48819835-f4056a00-ed8c-11e8-922f-d9679e7c4296.gif)

**Additional Information**

[systemPreferences.postNotification](https://electronjs.org/docs/api/system-preferences#systempreferencespostnotificationevent-userinfo-macos) method uses [NSDistributedNotificationCenter](https://developer.apple.com/documentation/foundation/nsdistributednotificationcenter#1651293?language=objc) behind the scene, but it uses a wrong selector to post notifications to it.

Instead of using this selector (✅):
- [-NSDistributedNotificationCenter.postNotificationName:object:userInfo:deliverImmediately:](https://developer.apple.com/documentation/foundation/nsdistributednotificationcenter/1418360-postnotificationname?language=objc), where `deliverImmediately` is set to `TRUE`

It uses an inherited selected from a parent class (🚫):
- [-NSNotificationCenter.postNotificationName:object:userInfo:](https://developer.apple.com/documentation/foundation/nsnotificationcenter/1410608-postnotificationname?language=objc)

All this makes the powerful tool called `NSDistributedNotificationCenter`, widely used for inter-process communications - completely useless, since all the posted notifications are suspended by default and can reach external applications only in case they are active. 

**Possible solution**

I suggest that this method should use a correct `NSDistributedNotificationCenter`'s selected by default:
[...]

PR #16060, opened 2018-12-13
feat: allow immediate MacOS notifications

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/15789.

Refactors notification posting on MacOS to allow for some notifications to deliver immediately regardless of whether or not the subscribing app is active or not. 

Only `NSDistributedNotificationCenter` has access to the `deliverImmediate` selector, so I split out the `DoPostNotification` helper into the calling functions as it now make more sense for each to just call the methods themselves rather than call out to a helper. `deliverImmediate` was also added as an optional new param so as to maintain backwards compat.

/cc @ckerr

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: allow for MacOS notifications to be immediately delivered

### DIFF
--- a/atom/browser/api/atom_api_system_preferences_mac.mm
+++ b/atom/browser/api/atom_api_system_preferences_mac.mm
@@ -109,9 +109,8 @@ std::string ConvertAuthorizationStatus(AVAuthorizationStatusMac status) {
 void SystemPreferences::PostNotification(const std::string& name,
                                          const base::DictionaryValue& user_info,
                                          mate::Arguments* args) {
-  bool immediate;
-  if (!args->GetNext(&immediate))
-    immediate = false;
+  bool immediate = false;
+  args->GetNext(&immediate);
 
   NSDistributedNotificationCenter* center =
       [NSDistributedNotificationCenter defaultCenter];
--- a/atom/browser/api/atom_api_system_preferences_mac.mm
+++ b/atom/browser/api/atom_api_system_preferences_mac.mm
@@ -106,10 +106,19 @@ std::string ConvertAuthorizationStatus(AVAuthorizationStatusMac status) {
 
 }  // namespace
 
-void SystemPreferences::PostNotification(
-    const std::string& name,
-    const base::DictionaryValue& user_info) {
-  DoPostNotification(name, user_info, kNSDistributedNotificationCenter);
+void SystemPreferences::PostNotification(const std::string& name,
+                                         const base::DictionaryValue& user_info,
+                                         mate::Arguments* args) {
+  bool immediate;
+  if (!args->GetNext(&immediate))
+    immediate = false;
+
+  NSDistributedNotificationCenter* center =
+      [NSDistributedNotificationCenter defaultCenter];
+  [center postNotificationName:base::SysUTF8ToNSString(name)
+                        object:nil
+                      userInfo:DictionaryValueToNSDictionary(user_info)
+            deliverImmediately:immediate];
 }
 
 int SystemPreferences::SubscribeNotification(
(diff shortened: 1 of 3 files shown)


docs/api/system-preferences.md
```diff
@@ -59,10 +59,11 @@ Returns `Boolean` - Whether the system is in Dark Mode.
 
 Returns `Boolean` - Whether the Swipe between pages setting is on.
 
-### `systemPreferences.postNotification(event, userInfo)` _macOS_
+### `systemPreferences.postNotification(event, userInfo[, deliverImmediately])` _macOS_
 
 * `event` String
 * `userInfo` Object
+* `deliverImmediately` Boolean (optional) - `true` to post notifications immediately regardless of whether or not the subscribing app is active or not.
```
nornagon:
```suggestion
* `deliverImmediately` Boolean (optional) - `true` to post notifications immediately regardless of whether or not the subscribing app is active or not. Defaults to `false`. See [NSDistributedNotificationCenter documentation](https://developer.apple.com/documentation/foundation/nsdistributednotificationcenter/1418360-postnotificationname) for details.
```

atom/browser/api/atom_api_system_preferences_mac.mm
```diff
@@ -106,10 +106,18 @@ AVMediaType ParseMediaType(const std::string& media_type) {
 
 }  // namespace
 
-void SystemPreferences::PostNotification(
-    const std::string& name,
-    const base::DictionaryValue& user_info) {
-  DoPostNotification(name, user_info, kNSDistributedNotificationCenter);
```
nornagon: can we delete the `NotificationCenterKind` enum now?

codebytere: yes! will do

atom/browser/api/atom_api_system_preferences_mac.mm
```diff
@@ -156,29 +171,6 @@ AVMediaType ParseMediaType(const std::string& media_type) {
...
-      center = [[NSWorkspace sharedWorkspace] notificationCenter];
-      break;
-    default:
-      break;
-  }
-  [center postNotificationName:base::SysUTF8ToNSString(name)
-                        object:nil
-                      userInfo:DictionaryValueToNSDictionary(user_info)];
-}
-
```
ckerr: This function declaration should be removed from `atom_api_system_preferences.h` too

docs/api/system-preferences.md
```diff
@@ -59,10 +59,11 @@ Returns `Boolean` - Whether the system is in Dark Mode.
 
 Returns `Boolean` - Whether the Swipe between pages setting is on.
 
-### `systemPreferences.postNotification(event, userInfo)` _macOS_
+### `systemPreferences.postNotification(event, userInfo[, deliverImmediately])` _macOS_
 
 * `event` String
 * `userInfo` Object
+* `deliverImmediately` Boolean (optional) - `true` to post notifications immediately regardless of whether or not the subscribing app is active or not. Defaults to `false`. See [NSDistributedNotificationCenter documentation](https://developer.apple.com/documentation/foundation/nsdistributednotificationcenter/1418360-postnotificationname) for details.
```
ckerr: _(nit)_  "regardless of", "whether or not", and "active or not" all say the same thing.

```diff
- immediately regardless of whether or not the subscribing app is active or not
+ immediately even when the subscribing app is inactive
```

nornagon:
Hey @codebytere, the release notes here are confusing because "notification" sounds like it means [Notification](https://electronjs.org/docs/api/notification). Can you edit @release-clerk's notes and clarify?

Issue #10622, opened 2017-09-27
The fate of BrowserView


zeke:
I heard some murmurs at our recent Tokyo mini-summit that an upcoming version of Chrome will address whatever problems [Electron's BrowserView API](https://github.com/electron/electron/blob/18589a498cad9e8d8f31cb9b1b36c6c00d255ef1/docs/api/browser-view.md) was created to solve. Is it Chromium 61? Does that mean that we should deprecate BrowserView when Chromium 61 lands in Electron?

cc @electron/browserview @poiru @felixrieseberg

poiru:
I don't think so. `<webview>` still has a [number of bugs](https://github.com/electron/electron/issues?q=is%3Aopen+is%3Aissue+label%3Acomponent%2Fwebview). There are also architectural bugs like slow auto-resizing that are unlikely to be fixed in Chromium in the near future.

Out of curiosity, do you recall any details from the murmurs?

deepak1556:
Chromium 61 lands OOPIF (actually its already available behind a flag), its an architectural change in how webview works. We need to implement it in Electron and its quite some work. I don't think we would want to deprecate BrowserView for it.

zeke:
> do you recall any details

No I don't have any details; just heard people talking about it in passing. Apparently Slack is using BrowserView but they think they'll be able to switch back to webviews in Chromium 61.

CharlieHess:
I'm going to have to disagree with any plans (on Slack's end) to switch back to webview (at least anytime soon). Admittedly we haven't shipped our release with `BrowserView` yet, but all early signs point to it being much more reliable than `webview`. Unless we find a _new set_ of problems with `BrowserView`, there'd really be no good reason to rearrange all our code a second time.

TimNZ:
For the rest of us mortals is there a summary (bullet point is fine) of why/when BrowserView should be used over BrowserWindow? Top 3-5 list would do the job, a when and when not to use..

I am building an app which uses a lot of webview and am finding issues that can be worked around or just tolerated for the time being, and have no idea if BrowserView would address them.

Thanks

TimNZ:
https://www.chromium.org/developers/design-documents/oop-iframes

"It is worth noting that the <webview> implementation is being migrated to work on top of the new OOPIF infrastructure."

felixrieseberg:
[...]

PR #9166, opened 2017-04-11
Implement initial, experimental BrowserView API

poiru:
Right now, `<webview>` is the only way to embed additional content in a
`BrowserWindow`. Unfortunately `<webview>` suffers from a [number of
problems](https://github.com/electron/electron/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20label%3Awebview%20). To make matters worse, many of these are upstream Chromium bugs instead of Electron-specific bugs.

For us at [Figma](https://www.figma.com), the main issue is very slow performance.

Despite the upstream improvements to `<webview>` through the OOPIF work, it is probable that there will continue to be `<webview>`-specific bugs in the future.

Therefore, this introduces a `<webview>` alternative to called `BrowserView`, which...

- is a thin wrapper around `api::WebContents` (so bugs in `BrowserView` will
  likely also be bugs in `BrowserWindow` web contents)

- is instantiated in the main process like `BrowserWindow` (and unlike
  `<webview>`, which lives in the DOM of a `BrowserWindow` web contents)

- needs to be added to a `BrowserWindow` to display something on the screen

This implements the most basic API. The API is expected to evolve and change in the near future and has consequently been marked as experimental. Please do not use this API in production unless you are prepared to deal with breaking changes.

In the future, we will want to change the API to support multiple `BrowserView`s per window. We will also want to consider z-ordering, auto-resizing, and possibly even nested views.

PR #16064, opened 2018-12-14
feat: multi browser views support for BrowserWindow

summeroff:
#### Description of Change
new feature: multi browser views support for BrowserWindow 

Add functions addBrowserView, removeBroserView, getBrowserViews to BrowserWindow class. 

Existing API setBrowserView and getBrowserView changed to use new API inside.

This functionality was mentioned in the initial BrowserView feature PR https://github.com/electron/electron/pull/9166 . And asked few times since then. Recently in a discussion https://github.com/electron/electron/issues/10622 .

#### Checklist

- [x] `npm test` passes for `npm run test -match=BrowserView`, `all-test` run have few fails even on unchanged `2-0-x` branch.  
- [x] tests are [added](https://github.com/electron/electron/blob/master/docs/development/testing.md)

#### Release Notes

Notes: multi browser views support for BrowserWindow

### DIFF
--- a/atom/browser/api/atom_api_window.cc
+++ b/atom/browser/api/atom_api_window.cc
@@ -203,7 +203,6 @@ void Window::OnWindowClosed() {
 
   RemoveFromParentChildWindows();
 
-  ResetBrowserView();
   ResetBrowserViews();
 
   // Destroy the native class when window is closed.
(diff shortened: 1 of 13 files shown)


atom/browser/api/atom_api_window.cc
```diff
@@ -914,6 +915,65 @@ void Window::ResetBrowserView() {
   browser_view_.Reset();
 }
 
+void Window::AddBrowserView(v8::Local<v8::Value> value) {
+  mate::Handle<BrowserView> browser_view;
+  if (value->IsNull() || value->IsUndefined()) {
```
zcbenz: There is no need to explicitly check `null`, type check can be down with:

```c++
if (mate::ConvertFromV8(isolate(), value, &browser_view)) {
} else {
  // Throw type error.
}
```

zcbenz: There is no need to check `null`, type check can be done with:

```c++
if (mate::ConvertFromV8(isolate(), value, &browser_view)) {
} else {
  // Throw type error.
}
```

`Window::RemoveBrowserView` can also be modified in the same way.

atom/browser/api/atom_api_window.cc
```diff
@@ -914,6 +915,65 @@ void Window::ResetBrowserView() {
...
+
+      (*get_that_view).second.Reset(isolate(), value);
+      browser_views_.erase(get_that_view);
+    }
+    window_->UpdateDraggableRegionViews();
+  }
+}
+
+void Window::ResetBrowserViews() {
+  for (auto views_iter = browser_views_.begin();
```
zcbenz: Can use range-based for loop here:

```c++
for (auto item : browser_views_) {
  item.second.Reset()
}
```

Other for loops can also be modified as such.

atom/browser/native_window.h
```diff
@@ -156,6 +156,8 @@ class NativeWindow : public base::SupportsUserData,
   virtual void SetMenu(AtomMenuModel* menu);
   virtual void SetParentWindow(NativeWindow* parent);
   virtual void SetBrowserView(NativeBrowserView* browser_view) = 0;
```
zcbenz: We can remove the `NativeWindow::SetBrowserView` implementation, and add a simple fallback for it by reusing the new API:

```js
BrowserWindow.prototype.setBrowserView = (view) => {
  this.resetBrowserViews()
  this.addBrowserView(view)
}
```

docs/api/browser-window.md
```diff
@@ -1478,7 +1478,21 @@ removed in future Electron releases.
...
 #### `win.getBrowserView()` _Experimental_
 
-Returns `BrowserView | null` - an attached BrowserView. Returns `null` if none is attached.
+Returns `BrowserView | null` - an BrowserView attached by setBrowserView. Returns `null` if none is attached.
+
+#### `win.addBrowserView(browserView)` _Experimental_
+
+* `browserView` [BrowserView](browser-view.md)
+
+Replacement API for setBrowserView supporting work with multi browser views.
```
zcbenz: I think we should have a clear definition on the behavior when using `setBrowserView` and `addBrowserView` together, and have it documented. I suggest:

1. Make `setBrowserView` an alias for `removeAllBrowserViews() && addBrowserView(view)`.
2. Make `getBrowserView` throw if there are multiple browser views, and return `getBrowserViews()[0]` as fallback.

atom/browser/api/atom_api_window.cc
summeroff: done.

atom/browser/native_window.h
summeroff: I have tried to keep APIs separate to simplify usage in old code and new code written with new API. Even thinking about more explicit way what code can request new functionality. 
Your suggestion is good as it is let add new API without keeping legacy code and will not affect 
 existing user code. 
I have changed code in accordance with your suggestion.

atom/browser/api/atom_api_window.cc
summeroff: Work of ConvertFromV8 is done by template implementation for corresponding type. For most types this indeed have check for null. But mate::Handle<> FromV8 converter does something opposite. 
```
 if (val->IsNull() || val->IsUndefined()) {
       *out = mate::Handle<T>();
       return true;
     }
```
And that lead to crash if null is passed. So I have to add check back. 
But change it to value->IsObject() 
Is it OK?

Do I have to add "Throw type error"? Prev API does not do it and as AddBrowserView works now as fall back for SetBrowserView. Then SetBrowserView start to throwing error too. But sending null to setBrowserView is required way to remove BrowserView what was set before.

docs/api/browser-window.md
summeroff: documentation updated

zcbenz:
Sorry I just noticed that this PR is targeted for 2.0.x. According to our versioning policy, we will not add new features to 2.0.x: https://www.electronjs.org/docs/tutorial/electron-versioning.

Do you mind rebasing your changes on `master` branch?

summeroff:
Will it goes only to electron 4 from master or it can be backported to 3-0-x too?

zcbenz:
New features will be included in 5.0, currently we don't have plans to publish minor updates for 3.x/4.x for new features.

Issue #16030, opened 2018-12-12
Isolated Security Origin for App


dbkr:
**Is your feature request related to a problem? Please describe.**
Our app currently loads its page from a file:// URI. The problem with this is that pages from file:// URIs have access to any other file in the file system. Our app doesn't need this so I'd like to disable this so that if there were an XSS vulnerability, malicious code would not be able to access the local file system.

If we loaded our resources from an HTTPS URI, the page would be restricted to the origin it was loaded from and not have file system access, so this would achieve this, but for obvious reasons we don't want to hit the network to load the apps resources.

Is there already a recommended way of achieving this?

**Describe the solution you'd like**
Load a page in a WebView with a specific origin such that:
 * The page can load other resources from the same directory or subdirectories
 * The page does not have access to the local filesystem via file:// URIs
 * Resources come from disk, installed when the electron app is installed, rather than having to load resources over the network.

A potential solution would be something like setting up a fake, internal URL scheme that will serve the contents of an asar bundle, so the page can be loaded from this.

**Describe alternatives you've considered**
I've also considered somehow setting the CSP of the page to 'sandbox' to give it a unique origin, but this means the app can't load other files it needs (that would otherwise be loaded from the same origin).

**Additional context**
https://github.com/electron/electron/issues/14350 is a similar problem, although more specifically asking just to execute javascript.

dbkr:
It looks like registering a custom protocol (https://electronjs.org/docs/api/protocol) might do the trick, except that I still can't set a CSP for the page other than as a meta tag in the HTML.

MarshallOfSound:
> except that I still can't set a CSP for the page other than as a meta tag in the HTML.

You can set it through a header with the `webRequest` module

MarshallOfSound:
Using a custom protocol is the solution you're looking for here

dbkr:
Thanks, although I was trying this and was not getting callbacks for the custom protocol handler using `webRequest` (I was successfully getting callbacks for http/https requests the page was making). Should requests to custom protocols appear as callbacks in `webRequest`?

dbkr:
Actually, I get `onBeforeRequest` and `onCompleted` for requests to the custom protocol but I would need `onHeadersReceived` to inject custom headers?

dbkr:
[...]

PR #16098, opened 2018-12-17
feat: Support response headers in File protocol handler

dbkr:
#### Description of Change
This adds support for response headers to `protocol.registerFileProtocol` to match `protocol.registerStreamProtocol`.

My specific use case for this is to set a CSP for pages served from local content rather than HTTPS: setting the header by intercepting the request with the `webRequest` module doesn't work because `URLRequestAsarJob` doesn't call the network_delegate methods (see https://github.com/electron/electron/issues/16030). In any case, this seems like a more direct way of setting the header.

I've also changed how this works a little to make it more consistent (hopefully) with the other handlers (ie. convert the dict to a `base::DictionaryValue` and use `base::DictionaryValue::GetFoo()`).

Note that I've documented this in the text as the callback param in the list has `filePath` as a string. This could be changed so the `callback` taking an object is in the main docs: I'm happy to make this change.

Stakeholders: @zcbenz & @paulcbetts appear to have last made major changes here. cc @MarshallOfSound as you kindly helped on the issue I filed.

Test failures don't appear to be related to my changes: one of them is:
```
AssertionError: exit signal should be null, if you see this please tag @MarshallOfSound: expected 'SIGSEGV' to equal null
```
(so, cc @MarshallOfSound)

#### Checklist
- [x] `npm test` passes: failures appear unrelated to my changes

#### Release Notes
Notes: Add response header support to `protocol.registerFileProtocol` to match `protocol.registerStreamProtocol`.

### DIFF
--- a/atom/browser/net/url_request_async_asar_job.cc
+++ b/atom/browser/net/url_request_async_asar_job.cc
@@ -73,11 +73,15 @@ void URLRequestAsyncAsarJob::StartAsync(std::unique_ptr<base::Value> options,
   std::string file_path;
   response_headers_ = new net::HttpResponseHeaders("HTTP/1.1 200 OK");
   if (options->is_dict()) {
-    base::DictionaryValue* headersValue = nullptr;
     base::DictionaryValue* dict =
         static_cast<base::DictionaryValue*>(options.get());
-    dict->GetString("path", &file_path);
-    dict->GetDictionary("headers", &headersValue);
+    base::Value* pathValue =
+        dict->FindKeyOfType("path", base::Value::Type::STRING);
+    if (pathValue) {
+      file_path = pathValue->GetString();
+    }
+    base::Value* headersValue =
+        dict->FindKeyOfType("headers", base::Value::Type::DICTIONARY);
     if (headersValue) {
       for (const auto& iter : headersValue->DictItems()) {
         response_headers_->AddHeader(iter.first + ": " +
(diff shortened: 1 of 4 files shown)


atom/browser/net/url_request_async_asar_job.cc
```diff
@@ -71,14 +71,23 @@ void URLRequestAsyncAsarJob::StartAsync(std::unique_ptr<base::Value> options,
...
+  response_headers_ = new net::HttpResponseHeaders("HTTP/1.1 200 OK");
   if (options->is_dict()) {
-    auto* path_value =
-        options->FindKeyOfType("path", base::Value::Type::STRING);
-    if (path_value)
-      file_path = path_value->GetString();
+    base::DictionaryValue* headersValue = nullptr;
+    base::DictionaryValue* dict =
+        static_cast<base::DictionaryValue*>(options.get());
+    dict->GetString("path", &file_path);
```
ckerr: This introduces a minor regression of going back to deprecated API:

```c++
// DEPRECATED, use Value::FindPath(path) and Value::GetString() instead.
bool GetString(StringPiece path, std::string* out_value) const;
```

Could you keep the previous `FindKeyOfType` implementation?

dbkr: Ah, hadn't noticed that was deprecated. Fixed!

PR #16111, opened 2018-12-17
docs: move `require` related breaking changes to Electron 6.0 (4-0-x) (#16109)

BinaryMuse:
#### Description of Change

Backport of https://github.com/electron/electron/pull/16109

#### Release Notes

Notes: no-notes


BinaryMuse:
Actually this never made it into `4-0-x` in the first place, so this is not necessary.

Issue #16090, opened 2018-12-16
macOS: Expose APIs to check an Accessibility consent in "Privacy & Security" 


4ndv:
**Is your feature request related to a problem? Please describe.**
It's mostly related to #14837. Without "Accessibility" consent in "Privacy and Security" since Mojave electron will crash on unregisterAll and return incorrect result in register (true even if it actually wasn't registered) for media shortcuts.

It also will be good to be able to ask for this consent manually or check it's presence.

**Describe the solution you'd like**
Expose [AXIsProcessTrustedWithOptions](https://developer.apple.com/documentation/applicationservices/1459186-axisprocesstrustedwithoptions) with both popup showing and not and add these checks to register/unregister/unregisterAll for Mojave.

**Describe alternatives you've considered**
I've created [a small node.js package](https://github.com/4ndv/node-mac-accessibility-features-check) to perform this checks on my app side, but it's still an incorrect behavior to crash on unregisterAll

deed02392:
I was experiencing this crash on an Electron app, where `unregisterAll()` calls presented in the crash stack trace. However after removing such calls completely, I still got a crash:

`Process:               Electron [714]
Path:                  /Users/USER/Documents/*/Electron.app/Contents/MacOS/Electron
Identifier:            com.dev.google-play-music-desktop-player
Version:               2.0.8 (2.0.8)
Code Type:             X86-64 (Native)
Parent Process:        node [713]
Responsible:           iTerm2 [1594]
User ID:               501

Date/Time:             2018-12-16 23:00:09.103 +0000
OS Version:            Mac OS X 10.14.1 (18B75)
Report Version:        12
Bridge OS Version:     3.0 (14Y667)
Anonymous UUID:        B84FB050-3E4E-38F1-06EC-1CD3AE3EFADD

Sleep/Wake UUID:       9ED647BC-EA31-4A6C-BC4A-0C3457A2E7E9

Time Awake Since Boot: 200000 seconds
Time Since Wake:       4800 seconds

System Integrity Protection: enabled

Crashed Thread:        0  CrBrowserMain  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes:       0x0000000000000001, 0x0000000000000000

Application Specific Information:
Crashing on exception: -[__NSCFString isMenuOpen]: unrecognized selector sent to instance 0x7fbe1eb0a230

Application Specific Backtrace 1:
[...]

PR #16119, opened 2018-12-18
feat: expose app accessibility status api

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/16090.

Exposes an api through System Preferences allowing an app to determine whether or not it's been designated a [trusted accessibility client](https://developer.apple.com/library/archive/documentation/Accessibility/Conceptual/AccessibilityMacOSX/), where that means an app that is allowed to modify the way users interact with their computer.

This API takes a prompt argument; if `false` is passed then the result will be returned silently, and if `true` is passed then a dialog (like the one below) will be shown.

screen shot 2018-12-17 at 8 32 33 pm

/cc @miniak @4ndv

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Exposes an API to allow apps to determine their status as a trusted accessibility client.

### DIFF
--- a/docs/api/system-preferences.md
+++ b/docs/api/system-preferences.md
@@ -327,6 +327,12 @@ You can use the `setAppLevelAppearance` API to set this value.
 Sets the appearance setting for your application, this should override the
 system default and override the value of `getEffectiveAppearance`.
 
+### `systemPreferences.isTrustedAccessibilityClient(prompt)` _macOS_
+
+* `prompt` Boolean - whether or not the user will be informed via prompt if the current process is untrusted.
+
+Returns `Boolean` - `true` if the current process is a trusted accessibility client and `false` if it is not.
+
 ### `systemPreferences.getMediaAccessStatus(mediaType)` _macOS_
 
 * `mediaType` String - `microphone` or `camera`.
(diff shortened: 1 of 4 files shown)


miniak:
@codebytere you had a small bug in the code, passing `@YES` both when `prompt` was `true` and `false`. Fixed that + eliminated duplicate code by using conditional operator for the value.

codebytere:
ah ok, i wasn't sure if objc bool literal could be coerced from c++ bools

docs/api/system-preferences.md
```diff
@@ -327,6 +327,12 @@ You can use the `setAppLevelAppearance` API to set this value.
 Sets the appearance setting for your application, this should override the
 system default and override the value of `getEffectiveAppearance`.
 
+### `systemPreferences.isTrustedAccessibilityClient(prompt)` _macOS_
+
+* `prompt` Boolean - whether or not the user will be informed via prompt if the current process is untrusted.
+
+Returns `Boolean` - `true` if the current process is a trusted accessibility client and `false` if it is not.
```
BinaryMuse: Does this return before or after the user is prompted? E.g. if the user opts to accept the prompt and add the app, will this return `true`?

codebytere: the prompt can't be accepted per se, it just links to the system prefpane; so no :)

fttx:
Will it be included in electron v4?

PR #16121, opened 2018-12-18
chore: remove unused catch bindings

ZYSzys:
#### Description of Change

Remove unused catch bindings in `lib`

Refs: https://github.com/nodejs/node/pull/24079

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/lib/browser/api/auto-updater/squirrel-update-win.js
+++ b/lib/browser/api/auto-updater/squirrel-update-win.js
@@ -94,7 +94,7 @@ exports.checkForUpdate = function (updateURL, callback) {
       // Last line of output is the JSON details about the releases
       const json = stdout.trim().split('\n').pop()
       update = (ref = JSON.parse(json)) != null ? (ref1 = ref.releasesToApply) != null ? typeof ref1.pop === 'function' ? ref1.pop() : void 0 : void 0 : void 0
-    } catch (jsonError) {
+    } catch {
       // Disabled for backwards compatibility:
       // eslint-disable-next-line standard/no-callback-literal
       return callback(`Invalid result:\n${stdout}`)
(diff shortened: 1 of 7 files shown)


ZYSzys:
Election built error in CI......😅

It seems that we don't support `try {} catch {}` syntax at present right ?

codebytere:
@ZYSzys ping, if you could address the above change request we can get this PR moving again!

ZYSzys:
OK，give me some time and I’ll try.

ZYSzys:
There were errors said `SyntaxError: Unexpected token` when parsing files.

Do we need to update `acorn` or something...how should I fix it ?

miniak:
@ZYSzys I am not sure about the release notes, this is not a user facing change. Also the notes should be written in past tense.

codebytere:
@ZYSzys can you please fix the conflicts in this PR?

lib/.eslintrc
```diff
@@ -1,8 +1,13 @@
 {
   "parserOptions": {
+    "ecmaVersion": 2019,
```
nornagon: This seems fine to me, but I'm not sure what the default is here? If it defaults to latest, we should probably leave it unspecified.

nornagon:
(Also, looks like you got bit by another conflict; would be great if you could rebase again!)

lib/.eslintrc
ZYSzys: I reverted here since it seems that we supported `try {} catch {}` syntax now.
(Maybe duo to [.eslintrc.json](https://github.com/electron/electron/blob/master/.eslintrc.json) in the root dir)

PR #16125, opened 2018-12-18
fix: don't register some shortcuts without accessibility 

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/14837. 

/cc @ckerr 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed crash on macOS when using `globalShortcut` for media keys when accessibility access is not granted.

### DIFF
--- a/atom/browser/api/atom_api_global_shortcut.cc
+++ b/atom/browser/api/atom_api_global_shortcut.cc
@@ -19,6 +19,29 @@
 
 using extensions::GlobalShortcutListener;
 
+namespace {
+
+#if defined(OS_MACOSX)
+bool RegisteringMediaKeyForUntrustedClient(const ui::Accelerator& accelerator) {
+  if (base::mac::IsAtLeastOS10_14()) {
+    constexpr ui::KeyboardCode mediaKeys[] = {ui::VKEY_MEDIA_PLAY_PAUSE,
+                                              ui::VKEY_MEDIA_NEXT_TRACK,
+                                              ui::VKEY_MEDIA_PREV_TRACK};
+
+    if (std::find(std::begin(mediaKeys), std::end(mediaKeys),
+                  accelerator.key_code()) != std::end(mediaKeys)) {
+      bool trusted =
+          atom::api::SystemPreferences::IsTrustedAccessibilityClient(false);
+      if (!trusted)
+        return true;
+    }
+  }
+  return false;
+}
+#endif
+
+}  // namespace
+
 namespace atom {
 
 namespace api {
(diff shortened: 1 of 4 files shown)


codebytere:
@4ndv the crash occurs on unregister because the shortcuts are falsely registered from what i could tell, meaning that the simplest way to prevent that would be to ensure that no globalShortcuts are registered with incorrect accessibility permissions

alirezavalizade:
@codebytere I could not any changes on new version of electron, Do you know that when it will be fix?

alirezavalizade:
ping @codebytere

atom/browser/api/atom_api_global_shortcut.cc
```diff
@@ -43,14 +45,26 @@ bool GlobalShortcut::RegisterAll(
...
+  std::vector<std::string> mediaKeys = {"Media Play/Pause", "Media Next Track",
+                                        "Media Previous Track"};
+
   for (auto& accelerator : accelerators) {
+#if defined(OS_MACOSX)
+    std::string shortcutText = base::UTF16ToUTF8(accelerator.GetShortcutText());
+    if (std::find(mediaKeys.begin(), mediaKeys.end(), shortcutText) !=
+        mediaKeys.end()) {
+      SystemPreferences* sys = nullptr;
+      bool trusted = sys->IsTrustedAccessibilityClient(false);
```
miniak: we should make `IsTrustedAccessibilityClient` static instead

atom/browser/api/atom_api_global_shortcut.cc
```diff
@@ -59,6 +73,20 @@ bool GlobalShortcut::RegisterAll(
 
 bool GlobalShortcut::Register(const ui::Accelerator& accelerator,
                               const base::Closure& callback) {
+#if defined(OS_MACOSX)
+  std::vector<std::string> mediaKeys = {"Media Play/Pause", "Media Next Track",
```
miniak: can we avoid repeating this code?

atom/browser/api/atom_api_global_shortcut.cc
```diff
@@ -43,14 +45,26 @@ bool GlobalShortcut::RegisterAll(
     const std::vector<ui::Accelerator>& accelerators,
     const base::Closure& callback) {
   std::vector<ui::Accelerator> registered;
+  std::vector<std::string> mediaKeys = {"Media Play/Pause", "Media Next Track",
+                                        "Media Previous Track"};
+
   for (auto& accelerator : accelerators) {
+#if defined(OS_MACOSX)
+    std::string shortcutText = base::UTF16ToUTF8(accelerator.GetShortcutText());
+    if (std::find(mediaKeys.begin(), mediaKeys.end(), shortcutText) !=
```
miniak: can we check the actual keys instead of the text?

codebytere: the only accessor chromium gives us is `GetShortcutText()`; i wanted to check the keys originally but i'm not sure it's possible 🤔

docs/api/global-shortcut.md
```diff
@@ -54,6 +54,14 @@ When the accelerator is already taken by other applications, this call will
 silently fail. This behavior is intended by operating systems, since they don't
 want applications to fight for global shortcuts.
 
+The following accelerators will not be registered successfully on macOS 10.14 Mojave unless
+the app has been authorizes as a trusted accessibility client:
```
BinaryMuse:
```suggestion
the app has been authorized as a trusted accessibility client:
```

docs/api/global-shortcut.md
```diff
@@ -65,6 +73,14 @@ When a given accelerator is already taken by other applications, this call will
...
```
BinaryMuse:
```suggestion
the app has been authorized as a trusted accessibility client:
```

PR #16153, opened 2018-12-20
fix: add missing require in chromium-spec

zcbenz:
Fix the failing CI on master branch.

Notes: no-notes.


nornagon:
Looks like clerk is a bit sensitive to the `.` at the end of `no-notes`. cc @codebytere and @MarshallOfSound

PR #16176, opened 2018-12-21
feat: split openExternal into sync and async

codebytere:
#### Description of Change

This API can't be promisified in the same way as previous apis, as its callback is optional. 

This therefore follows Node's conventions and splits `shell.openExternal(url[, options, callback])` into `shell.openExternalSync(url[, options])` and `shell.openExternal(url[, options])`. 
`shell.openExternal(url[, options])` now returns a promise.

This PR also adds specs for previously-untested `shell.openExternal()` and also for `shell.openExternalSync()`.

/cc @miniak @ckerr 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: promisify `shell.openExternal()` by splitting it into a sync and async method

### DIFF
--- a/atom/common/api/atom_api_shell.cc
+++ b/atom/common/api/atom_api_shell.cc
@@ -10,6 +10,7 @@
 #include "atom/common/native_mate_converters/string16_converter.h"
 #include "atom/common/node_includes.h"
 #include "atom/common/platform_util.h"
+#include "atom/common/promise_util.h"
 #include "native_mate/dictionary.h"
 
 #if defined(OS_WIN)
@@ -43,17 +44,16 @@ struct Converter<base::win::ShortcutOperation> {
 
 namespace {
 
-void OnOpenExternalFinished(
-    v8::Isolate* isolate,
-    const base::Callback<void(v8::Local<v8::Value>)>& callback,
-    const std::string& error) {
+void OnOpenExternalFinished(v8::Isolate* isolate,
+                            scoped_refptr<atom::util::Promise> promise,
+                            const std::string& error) {
   if (error.empty())
-    callback.Run(v8::Null(isolate));
+    promise->Resolve();
   else
-    callback.Run(v8::String::NewFromUtf8(isolate, error.c_str()));
+    promise->RejectWithErrorMessage(error.c_str());
 }
 
-bool OpenExternal(
+bool OpenExternalSync(
 #if defined(OS_WIN)
     const base::string16& url,
 #else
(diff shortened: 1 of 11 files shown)


docs/api/shell.md
```diff
@@ -34,21 +34,29 @@ Returns `Boolean` - Whether the item was successfully opened.
...
+
+### `shell.openExternal(url[, options])`
+
+* `url` String - Max 2081 characters on windows, or the function returns false.
+* `options` Object (optional)
+  * `activate` Boolean (optional) - `true` to bring the opened application to the
+    foreground. The default is `true`. _macOS_
+  * `workingDirectory` String (optional) - The working directory. _Windows_
+
+Returns `Promise<Any>`
```
sindresorhus: The promise doesn't resolve to a value, so `Promise<void>` would be more correct.

docs/api/shell.md
```diff
@@ -34,21 +34,29 @@ Returns `Boolean` - Whether the item was successfully opened.
...
 Returns `Boolean` - Whether an application was available to open the URL.
-If callback is specified, always returns true.
 
-Open the given external protocol URL in the desktop's default manner. (For
-example, mailto: URLs in the user's default mail agent).
+Open the given external protocol URL in the desktop's default manner. (For example, mailto: URLs in the user's default mail agent).
+
+### `shell.openExternal(url[, options])`
+
+* `url` String - Max 2081 characters on windows, or the function returns false.
```
sindresorhus: I realize this is an existing typo, but `windows` => `Windows`.

docs/api/shell.md
codebytere: @sindresorhus in the old api, when a callback was used, it always returned `true`; since i don't think that signal is valuable here i chose to just resolve with nothing

sindresorhus: Then you need to update the `url` option text, since it currently says:

> Max 2081 characters on windows, **or the function returns false.**

---

It should then be `Promise<void>` to indicate it returns nothing.

atom/common/api/atom_api_shell.cc
```diff
@@ -43,17 +44,24 @@ struct Converter<base::win::ShortcutOperation> {
...
 namespace {
 
-void OnOpenExternalFinished(
-    v8::Isolate* isolate,
-    const base::Callback<void(v8::Local<v8::Value>)>& callback,
-    const std::string& error) {
+void OnOpenExternalFinished(const v8::Global<v8::Context>& context,
+                            scoped_refptr<atom::util::Promise> promise,
+                            const std::string& error) {
+  v8::Isolate* isolate = promise->isolate();
```
MarshallOfSound: Should this be done in the promise helper to reduce this boilerplate being duplicated for all promises

zcbenz: I agree moving the code to promise helper. But I would prefer doing it in a separate PR since it is a refactoring that modifies multiple files.

PR #16180, opened 2018-12-22
chore: update deprecated LSGetApplicationForURL

codebytere:
#### Description of Change

` LSGetApplicationForURL` is deprecated [here](https://developer.apple.com/documentation/coreservices/1445210-lsgetapplicationforurl?language=objc). 

It's replaced with `LSCopyDefaultApplicationURLForURL`, which returns `NULL` if an application can't be found, and populates `outError` with `kLSApplicationNotFoundErr`. Since there is only the possibility of one error being returned here, we no longer have a need for `MessageForOSStatus` and it is thus removed.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


miniak:
@codebytere should we maybe use `CFErrorCopyDescription` to get the text from the error instead of the hardcoded string?

PR #16203, opened 2018-12-27
build: extract external binaries config

alexeykuzmin:
#### Description of Change

I think it's better to have an explicit list of all "external binaries" we download.
/cc @jkleinsc 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no notes

### DIFF
--- a/script/update-external-binaries.py
+++ b/script/update-external-binaries.py
@@ -2,7 +2,7 @@
 
 import argparse
 import errno
-import sys
+import json
 import os
 
 from lib.config import PLATFORM, get_target_arch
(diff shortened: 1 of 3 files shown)


script/update-external-binaries.py
```diff
@@ -64,21 +73,37 @@ def is_updated(version_file, version):
...
 
-def download_and_unzip(url_prefix, framework):
-  zip_path = download_framework(url_prefix, framework)
-  if zip_path:
-    extract_zip(zip_path, 'external_binaries')
+def binary_should_be_downloaded(binary):
+  if 'platform' in binary and binary['platform'] != PLATFORM:
+    return False
 
+  if 'arch' in binary and binary['arch'] != get_target_arch():
```
nornagon: i think we need a distinction between host and target arch here, since some of the binaries are to be linked with the target binary, and others are to be run on the host machine.

jkleinsc: Maybe rename the json property to `targetArch`?
```suggestion
  if 'targetArch' in binary and binary['targetArch'] != get_target_arch():
```

nornagon: we'll also need a `hostArch` property, because e.g. `sccache` is only run on the host.

nornagon: it's possible to build Electron ARM with an x86 machine though.

nornagon: ah i see, you're just saying that there's no binary currently that depends on the host architecture but not the host OS. yes, i agree.

PR #16238, opened 2019-01-02
refactor: pass isWebViewTagEnabled via ELECTRON_BROWSER_SANDBOX_LOAD

miniak:
#### Description of Change
Passing options via command-line to a sandboxed renderer process is not safe as the process can be shared between multiple webContents.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/guest-view-manager.js
+++ b/lib/browser/guest-view-manager.js
@@ -328,8 +328,8 @@ const isWebViewTagEnabledCache = new WeakMap()
 
 const isWebViewTagEnabled = function (contents) {
   if (!isWebViewTagEnabledCache.has(contents)) {
-    const value = contents.getLastWebPreferences().webviewTag
-    isWebViewTagEnabledCache.set(contents, value)
+    const webPreferences = contents.getLastWebPreferences() || {}
+    isWebViewTagEnabledCache.set(contents, !!webPreferences.webviewTag)
   }
 
   return isWebViewTagEnabledCache.get(contents)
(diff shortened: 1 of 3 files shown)


nornagon:
In what situation would two WebContents which differ in this option share a process? affinity?

miniak:
@nornagon, for example. Also a child popup window can unless site isolation kicks in

miniak:
@nornagon bad example then, but affinity is still relevant.

miniak:
@nornagon can you approve?

lib/browser/guest-view-manager.js
```diff
@@ -328,8 +328,8 @@ const isWebViewTagEnabledCache = new WeakMap()
 
 const isWebViewTagEnabled = function (contents) {
   if (!isWebViewTagEnabledCache.has(contents)) {
-    const value = contents.getLastWebPreferences().webviewTag
-    isWebViewTagEnabledCache.set(contents, value)
+    const webPreferences = contents.getLastWebPreferences() || {}
+    isWebViewTagEnabledCache.set(contents, webPreferences.webviewTag)
```
alexeykuzmin: Why does it make sense to return `undefined` in some cases?
The function is supposed to return a boolean value.

miniak: fixed

alexeykuzmin:
The **electron-arm-testing** failed is not related to the change.

PR #15624, opened 2018-11-07
feat: add media access apis for mojave

codebytere:
#### Description of Change

This PR adds new APIs to control new media privacy and consent rules introduced in MacOS Mojave. Namely:

```
systemPreferences.getMediaAccessStatus(mediaType)
systemPreferences.askForMediaAccess(mediaType)
```

Apple Docs:
- [[AVCaptureDevice authorizationStatusForMediaType:]](https://developer.apple.com/documentation/avfoundation/avcapturedevice/1624613-authorizationstatusformediatype?language=objc)
- [[AVCaptureDevice requestAccessForMediaType:completionHandler:]](https://developer.apple.com/documentation/avfoundation/avcapturedevice/1624584-requestaccessformediatype?language=objc)
- [AVAuthorizationStatus](https://developer.apple.com/documentation/avfoundation/avauthorizationstatus?language=objc)

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

#### Release Notes

Notes: Added media access APIs for macOS Mojave

PR #16240, opened 2019-01-02
fix: add default media usage strings to info.plist

codebytere:
#### Description of Change

This PR adds default keys to Info.plist for `NSMicrophoneUsageDescription` and `NSCameraUsageDescription`. This will simplify usage of the [media access apis](https://github.com/electron/electron/pull/15624) added for MacOS Mojave.

/cc @nornagon @miniak 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added default `NSMicrophoneUsageDescription` and `NSCameraUsageDescription` keys to Info.plist.

### DIFF
--- a/atom/browser/resources/mac/Info.plist
+++ b/atom/browser/resources/mac/Info.plist
@@ -32,5 +32,9 @@
   <true/>
   <key>NSHighResolutionCapable</key>
   <true/>
+  <key>NSMicrophoneUsageDescription</key>
+	<string></string>
+  <key>NSCameraUsageDescription</key>
+	<string></string>
 </dict>
 </plist>
--- a/atom/browser/resources/mac/Info.plist
+++ b/atom/browser/resources/mac/Info.plist
@@ -33,8 +33,8 @@
   <key>NSHighResolutionCapable</key>
   <true/>
   <key>NSMicrophoneUsageDescription</key>
-  <string></string>
+  <string>This app needs access to the microphone</string>
   <key>NSCameraUsageDescription</key>
-  <string></string>
+  <string>This app needs access to the camera</string>
 </dict>
 </plist>
--- a/atom/browser/resources/mac/Info.plist
+++ b/atom/browser/resources/mac/Info.plist
@@ -33,8 +33,8 @@
   <key>NSHighResolutionCapable</key>
   <true/>
   <key>NSMicrophoneUsageDescription</key>
-	<string></string>
+  <string></string>
   <key>NSCameraUsageDescription</key>
-	<string></string>
+  <string></string>
 </dict>
 </plist>


atom/browser/resources/mac/Info.plist
```diff
@@ -32,5 +32,9 @@
   <true/>
   <key>NSHighResolutionCapable</key>
   <true/>
+  <key>NSMicrophoneUsageDescription</key>
+	<string></string>
```
nornagon: - indentation is weird
- put a dummy description here? I'm not quite sure what the dialog looks like so I'm not sure what would make sense

atom/browser/resources/mac/Info.plist
```diff
@@ -33,8 +33,8 @@
   <key>NSHighResolutionCapable</key>
   <true/>
   <key>NSMicrophoneUsageDescription</key>
-  <string></string>
+  <string>This app needs access to the microphone</string>
   <key>NSCameraUsageDescription</key>
-  <string></string>
+  <string>This app needs access to the camera</string>
```
nornagon: Can you share a screenshot of what the dialog looks like with this text?

codebytere:
@nornagon it'd look just like this:

img_5b6a2cfee4b79

nornagon:
Then where does the string you just added go?

nornagon:
The wording doesn't even seem to show up in the dialog, so i'm confused why this is needed at all :/ perhaps we can get some clarification on the original issue as to what the problem is?

BinaryMuse:
@nornagon I'm not 100% sure but I'm fairly certain the text shows up where the existing text is in the dialog @codebytere shared.

nornagon:
screen shot 2019-01-02 at 4 40 33 pm

@codebytere shared this screenshot with me, which she took with this patch applied.

nornagon:
though... wait a sec... why does that say `/Users/miniak`

BinaryMuse:
Okay, it looks like this string shows up as additional context:

BinaryMuse:
See https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture/requesting_authorization_for_media_capture_on_macos?language=objc for all the fun details

nornagon:
I don't 100% trust that the mac docs line up with what we're doing here. I don't have mojave, so can't test this right now, but someone really just needs to actually apply this patch and run it on their own machine to test. I thought @codebytere did that but now i'm confused by the `miniak` in the screenshot.

codebytere:
@nornagon i had chrome open in the background and took the screenshot while i was scrolled down to https://github.com/electron/electron/pull/15624#issuecomment-444048720.

codebytere:
Do we think we should try to merge this?

PR #16243, opened 2019-01-03
feat: export BoringSSL symbols

### LINKED ISSUE #13176: Addons cannot use the bundled OpenSSL
I maintain https://github.com/nodegit/nodegit and would like to start building NodeGit against the OpenSSL library that is bundled with Node. This would significantly improve our build process.

Node exports OpenSSL symbols for Windows https://github.com/nodejs/node/pull/6274. It would be very cool if Electron would export the openssl symbols as well, so that native module creators could target openssl that ships with Electron's copy of node.

**Describe alternatives you've considered**
We have been maintaining a copy of OpenSSL from node that we try to update when NodeJS updates their copy. This has been a bit of a nightmare due to the gyp files not perfectly matching our use case in NodeGit.

**Additional context**
It looks like there was a decision made to not export OpenSSL symbols in the past, but no reason explicitly mentioned here: https://github.com/electron/electron/pull/6585

But it does look like for regular nodejs, exporting those symbols has been working fine for the last couple years https://github.com/nodejs/node/pull/6274

nornagon:
#### Description of Change
Allows third-party native modules to link against the BoringSSL that's built with Electron.

Closes #13176.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Exported BoringSSL symbols, allowing third-party native modules to link against the included BoringSSL.

### DIFF
--- a/patches/common/chromium/boringssl_build_gn.patch
+++ b/patches/common/chromium/boringssl_build_gn.patch
@@ -1,15 +1,28 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Jeremy Apthorp <nornagon@nornagon.net>
 Date: Thu, 20 Sep 2018 17:48:05 -0700
-Subject: boringssl_build_gn.patch
+Subject: BoringSSL BUILD.gn
 
-Build BoringSSL with some extra functions that nodejs needs. Only affects
-the GN build; with the GYP build, nodejs is still built with OpenSSL.
+Build BoringSSL with some extra functions that nodejs needs.
+
+Also, export BoringSSL symbols from the main binary. This allows
+third-party native modules to link against the BoringSSL that's compiled
+into Electron. There's no way to flip this `#define` other than through
+a patch to BUILD.gn.
 
 diff --git a/third_party/boringssl/BUILD.gn b/third_party/boringssl/BUILD.gn
-index d31a9f29fa9c12e753708b2a1e75c33b70924300..dea5a6403f4c32f94bb58198c467bc7cc87a8217 100644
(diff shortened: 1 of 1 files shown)


BinaryMuse:
@nornagon Mind working on more [user-facing/formatted notes](https://github.com/electron/clerk/blob/master/README.md) before this lands? Thank you!

nornagon:
@BinaryMuse i've updated the notes, wdyt?

nornagon:
adding wip because the windows failures are real

codebytere:
@nornagon do you plan to come back to this PR?

nornagon:
Not at this time. Will revisit if there's renewed interest.

alejandroclaro:
My two cents here against the request:

The fact NodeJS expose OpenSSL symbols is a big issue. There are some issues in NodeJS repository taking about this. This is one in particular:

https://github.com/nodejs/node/issues/24964

Absorbing the OpenSSL dependency and re-exposing it is not a good idea. The `crypto` and `tls` modules do not expose all the functionalities that OpenSSL or any of its forks provide. Also, NodeJS does no exposed all the symbols, only a subset, so either way, any third party add-on will have to deal with symbols clashing issues, runtime incompatibility with the library version if the can link only with the symbols provided in node, and also issue with other dependencies that depend on dynamic versions of OpenSSL, or BoringSSL, or LibreSSL.

I think it would be better if NodeJS and Electron does not export the OpenSSL/BoringSSL symbols when it's statically linked, an allows that way third-party addons and their dependencies to use the system installed version of OpenSSL, and/or deal with multiple versions or forks in independent way (responsibility of the node addon developer).

PR #16252, opened 2019-01-03
fix: security: improve IPC validation in guest-view-manager

miniak:
#### Description of Change
- fix sender validation for `ELECTRON_GUEST_VIEW_MANAGER_FOCUS_CHANGE` (regression from https://github.com/electron/electron/pull/15859)
- add sender validation for `ELECTRON_GUEST_VIEW_MANAGER_DESTROY_GUEST`, `ELECTRON_GUEST_VIEW_MANAGER_ATTACH_GUEST` and `ELECTRON_BROWSER_GUEST_WEB_CONTENTS`

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/guest-view-manager.js
+++ b/lib/browser/guest-view-manager.js
@@ -184,9 +184,12 @@ const attachGuest = function (event, embedderFrameId, elementInstanceId, guestIn
   const guestInstance = guestInstances[guestInstanceId]
   // If this isn't a valid guest instance then do nothing.
   if (!guestInstance) {
-    return
+    throw new Error(`Invalid guestInstanceId: ${guestInstanceId}`)
   }
   const { guest } = guestInstance
+  if (guest.hostWebContents !== event.sender) {
+    throw new Error(`Access denied to guestInstanceId: ${guestInstanceId}`)
+  }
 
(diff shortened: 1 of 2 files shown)


miniak:
@zcbenz can you please review?

lib/browser/guest-view-manager.js
```diff
@@ -342,6 +345,17 @@ const handleMessage = function (channel, handler) {
   })
 }
 
+const getGuestForWebContents = function (guestInstanceId, contents) {
```
nornagon: perhaps fold `getGuest` into this function? no particular need to have it separate any more, right?

miniak: @nornagon original `getGuest` is still when handling `ELECTRON_GUEST_VIEW_MANAGER_FOCUS_CHANGE` as in this case it's the `<webview>` itself sending the IPC and the `guestInstanceId` validation is different.

nornagon: Ah, ok. Well in that case, at least put this function next to `getGuest` :)

miniak: moved

PR #16254, opened 2019-01-03
docs: Documentation changes - Alphabetize BrowserWindow Options, Events, Instance Properties, and Instance Methods

wsteidley:
#### Description of Change
Alphabetize browser window Options, Events, Instance Properties, and Instance Methods to improve readability and ease of use

#### Checklist

#### Release Notes

Notes:  Alphabetize BrowserWindow Options, Events, Instance Properties, and Instance Methods

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -139,81 +139,85 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
 ### `new BrowserWindow([options])`
 
 * `options` Object (optional)
-  * `width` Integer (optional) - Window's width in pixels. Default is `800`.
-  * `height` Integer (optional) - Window's height in pixels. Default is `600`.
-  * `x` Integer (optional) (**required** if y is used) - Window's left offset from screen.
-    Default is to center the window.
-  * `y` Integer (optional) (**required** if x is used) - Window's top offset from screen.
-    Default is to center the window.
-  * `useContentSize` Boolean (optional) - The `width` and `height` would be used as web
-    page's size, which means the actual window's size will include window
-    frame's size and be slightly larger. Default is `false`.
+  * `acceptFirstMouse` Boolean (optional) - Whether the web view accepts a single
+    mouse-down event that simultaneously activates the window. Default is
+    `false`.
+  * `alwaysOnTop` Boolean (optional) - Whether the window should always stay on top of
+    other windows. Default is `false`.
+  * `autoHideMenuBar` Boolean (optional) - Auto hide the menu bar unless the `Alt`
+    key is pressed. Default is `false`.
+  * `backgroundColor` String (optional) - Window's background color as a hexadecimal value,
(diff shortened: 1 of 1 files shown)


BinaryMuse:
Thanks for the PR, @wsteidley.

Some thoughts as I was considering this PR, interested in hearing 💭s from others:

* This is a pretty dense change set and could be difficult to review thoroughly, although a good GUI diff tool might help here.
* I think I like the idea of alphabetizing the docs *in general*; some pros and cons:
  * 👍 It makes it easier to find the docs for a specific event or option without using browser search
  * 👎 It makes it more difficult to "explore" and discover options related to each other
  * 👎 Since these files are all edited by hand, there's nothing stopping them from getting out-of-order again other than people being careful
  * 👎 I don't love that we'd have this one API document alphabetized but not the others, but I guess one must start somewhere
  * 👎 This would likely break any open PR that touches the document, but since this PR is only one doc maybe that wouldn't be so bad
  * ⚠️ I'm not sure how this would affect our translations — need to do more investigating here

I would really love to be in a place where our entire API documentation could be auto-generated from the code, but I'm not sure how practical that is. In such a world, we could do really nice and fancy things with the docs to make them sortable, filterable, etc.

wsteidley:
@BinaryMuse 

>     * This is a pretty dense change set and could be difficult to review thoroughly, although a good GUI diff tool might help here.

Fair point. It could be broken up if necessary. The number of changes is a product of this doc growing with events/options/methods being placed in no obvious order

>     * I think I like the idea of alphabetizing the docs _in general_; some pros and cons:
>       
>       * +1 It makes it easier to find the docs for a specific event or option without using browser search
>       * -1 It makes it more difficult to "explore" and discover options related to each other
I'd argue it does the opposite because by alphabetizing all the methods things like gets, sets, and checks (isSomethingOrOther) methods are grouped. 

In regards to Options I submitted this PR for the exact reason that I found it difficult to read/explore the document and parse the various options. Moving from the doc to code and then back meant it was difficult to find my place and parse new options I hadn't previously read through.

Maybe there is some happy medium using subsections for things like OS specific items or experimental items?

>       * -1 Since these files are all edited by hand, there's nothing stopping them from getting out-of-order again other than people being careful
100% true. Certainly this has lead to this document (and others) growing until arriving at the current state

>       * -1 I don't love that we'd have this one API document alphabetized but not the others, but I guess one must start somewhere
This just happens to be a document that I use regularly and having everything out of order has made things difficult so I decided to start here

>       * -1 This would likely break any open PR that touches the document, but since this PR is only one doc maybe that wouldn't be so bad
>       * warning I'm not sure how this would affect our translations — need to do more investigating here
Good point. I hadn't looked into those

miniak:
@MarshallOfSound we should close this PR and create a follow-up task to address this.

codebytere:
@wsteidley in agreement with the above, i'm going to go ahead and close this. We appreciate your contribution even though this one was not merged!

PR #15057, opened 2018-10-10
[WIP] ci: enable logging that was disable during 69 upgrade

deepak1556:
##### Description of Change

Lets identify the root cause and fix it.

##### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [ ] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

##### Release Notes

Notes: no-notes

PR #16257, opened 2019-01-04
ci: re-enable logging on CI

nornagon:
This reverts commit a5c3091c34efdadd388e09d003e57e8e88d6c6a9.

Hoping that the underlying issue was fixed by #15555

Notes: no-notes


deepak1556:
We should run linux release tests in this branch where the issue was present. See Ci results from https://github.com/electron/electron/pull/15057

codebytere:
(pending verification above)

Issue #16256, opened 2019-01-03
app quits immediately if setAccessibilitySupportEnabled is called


marisademeglio:
* Output of `node_modules/.bin/electron --version`: v4.0.0
* Operating System (Platform and Version): macOS 10.14.2

* Output of `node_modules/.bin/electron --version` on last known working Electron version (if applicable): 2.0.7

### Expected behavior

I should be able to call `app.setAccessibilitySupportEnabled(true)` during initialization, and the application should start and stay open.

### Actual behavior

The application starts and exits immediately if `app.setAccessibilitySupportEnabled(true)` is called during initialization. Same for passing in `false`.

### How to reproduce

Here is a simple project that shows this error. The application behaves as expected (does not exit immediately) when this line is commented out:
https://github.com/marisademeglio/electron4-setA11yTreeEnabled/blob/master/src/main/main.js#L27

  $ git clone https://github.com/marisademeglio/electron4-setA11yTreeEnabled.git
  $ yarn install
  $ yarn start

codebytere:
i just ran this in electron fiddle on `v4.0.0` with no issues:
```
const {app, BrowserWindow, systemPreferences} = require('electron')

let mainWindow

function createWindow () {
  mainWindow = new BrowserWindow({width: 800, height: 600})

  mainWindow.loadURL('https://github.com')

  app.setAccessibilitySupportEnabled(true)

  mainWindow.on('closed', () => mainWindow = null)
}

app.on('ready', createWindow)

// Quit when all windows are closed.
app.on('window-all-closed', () => {
  if (process.platform !== 'darwin') app.quit()
})

app.on('activate', () => {
  if (mainWindow === null) createWindow()
})
```
do you have a more minimal example that shows your issue?

marisademeglio:
Here, it seems to depend on the order: 
https://gist.github.com/969873022f427dfa3746360e399ccfe9

If I put it in createWindow() as you did, it works. But if I put it before the window is created, it exits immediately.

PR #16261, opened 2019-01-04
fix: error if a11y support changed before ready

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/16256.

Starting with `v4.0.0`, `app.setAccessibilitySupportEnabled()` can only be called after the `ready` event is emitted. This clarifies that in documentation and ensures that an error is thrown so the method does not crash.

/cc @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Throw an error if `app.setAccessibilitySupportEnabled()` is called before the `ready` event is emitted.


jkleinsc:
Merging as arm failure is unrelated to change (and broken in master).

PR #16276, opened 2019-01-04
fix: don't register some shortcuts without accessibility

codebytere:
#### Description of Change

Backports https://github.com/electron/electron/pull/16125 to `3-1-x`. Also partially backports https://github.com/electron/electron/pull/16119, as it's required for the fix impl.

See those PRs for more details.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed crash on macOS when using `globalShortcut` for media keys when accessibility access is not granted.

### DIFF
--- a/atom/common/platform_util_mac.mm
+++ b/atom/common/platform_util_mac.mm
@@ -8,6 +8,8 @@
 #import <Cocoa/Cocoa.h>
 #import <ServiceManagement/ServiceManagement.h>
 
+#include "atom/common/platform_util.h"
+
 #include "base/callback.h"
 #include "base/files/file_path.h"
 #include "base/files/file_util.h"
@@ -18,6 +20,7 @@
 #include "base/strings/stringprintf.h"
 #include "base/strings/sys_string_conversions.h"
 #include "net/base/mac/url_conversions.h"
+#include "third_party/WebKit/Source/platform/mac/VersionUtilMac.h"
 #include "url/gurl.h"
 
 namespace {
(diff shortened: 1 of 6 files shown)


miniak:
```
../../atom/browser/api/atom_api_global_shortcut.cc:28:18: error: no member named 'IsAtLeastOS10_14' in namespace 'base::mac'
  if (base::mac::IsAtLeastOS10_14()) {
```

codebytere:
yeah @miniak i think that version of chrome doesn't have that method

nornagon:
The PR description says "backports to 4-0-x" but it looks like the target is 3-1-x?

atom/common/platform_util_mac.mm
```diff
@@ -18,6 +20,7 @@
 #include "base/strings/stringprintf.h"
 #include "base/strings/sys_string_conversions.h"
 #include "net/base/mac/url_conversions.h"
+#include "third_party/blink/renderer/platform/mac/version_util_mac.h"
```
miniak: replace with `#include "third_party/WebKit/Source/platform/mac/VersionUtilMac.h"`

atom/common/platform_util_mac.mm
```diff
@@ -166,6 +169,10 @@ void OpenExternal(const GURL& url,
                  });
 }
 
+bool IsAtLeastOS10_14() {
+  return MacOSXMinorVersion() >= 14;
```
miniak: `blink::internal::MacOSXMinorVersion()`

atom/common/platform_util_mac.mm
```diff
@@ -170,7 +170,7 @@ void OpenExternal(const GURL& url,
 }
 
 bool IsAtLeastOS10_14() {
-  return MacOSXMinorVersion() >= 14;
+  return blink::internal::MacOSXMinorVersion()
```
miniak: `return blink::internal::MacOSXMinorVersion() >= 14;`

miniak: `return blink::internal::MacOSXMinorVersion() >= 14;`

PR #16282, opened 2019-01-04
feat: add app.commandLine.hasSwitch() / app.commandLine.getSwitchValue()

miniak:
#### Description of Change
Expose the following APIs, which we are already using internally in [lib/renderer/init.js](https://github.com/electron/electron/blob/9cc3fbabf73cdb2d59fd492cb1fb39a9b24f1bb3/lib/renderer/init.js#L34)
- `app.commandLine.hasSwitch()`
- `app.commandLine.getSwitchValue()`

These methods are useful for parsing command-line options in robust way without having to use external libraries.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `app.commandLine.hasSwitch()` / `app.commandLine.getSwitchValue()`

### DIFF
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -1159,6 +1159,20 @@ correctly.
 
 **Note:** This will not affect `process.argv`.
 
+### `app.commandLine.hasSwitch(switch)`
+
+* `switch` String - A command-line switch
+
+Returns `Boolean` - Whether the command-line switch is present.
+
+### `app.commandLine.getSwitchValue(switch)`
+
+* `switch` String - A command-line switch
+
+Returns `String` - The command-line switch value.
+
+**Note:** When the switch is not present, it returns empty string.
+
 ### `app.enableSandbox()` _Experimental_ _macOS_ _Windows_
 
 Enables full sandbox mode on the app.
--- a/lib/browser/api/app.js
+++ b/lib/browser/api/app.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const bindings = process.atomBinding('app')
+const commandLine = process.atomBinding('command_line')
 const path = require('path')
 const { app, App } = bindings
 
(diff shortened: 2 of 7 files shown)


lib/browser/api/app.js
```diff
@@ -25,6 +25,7 @@ Object.assign(app, {
     return Menu.getApplicationMenu()
   },
   commandLine: {
+    ...process.atomBinding('command_line'),
     appendSwitch (...args) {
```
deepak1556: You could move these methods implemented via `app` module into `command_line` module in a followup PR.

miniak: done in this PR

docs/api/app.md
```diff
@@ -1159,6 +1159,20 @@ correctly.
...
+
+Returns `Boolean` - Whether the command-line switch is present.
+
+### `app.commandLine.getSwitchValue(switch)`
+
+* `switch` String - A command-line switch
+
+Returns `String` - The command-line switch value.
+
+**Note:** When the switch is not present, it returns empty string.
```
alexeykuzmin: This is sad.

lib/browser/api/app.js
```diff
@@ -17,6 +18,12 @@ let dockMenu = null
 Object.setPrototypeOf(App.prototype, EventEmitter.prototype)
 EventEmitter.call(app)
 
+const castArgs = function (args) {
+  return args.map((arg) => {
+    return typeof arg !== 'string' ? `${arg}` : arg
```
alexeykuzmin: Why not use a `String()` call to convert to string? Calling `String` without `new` will produce a primitive not a `String` instance.

miniak: done

docs/api/app.md
miniak: that's how the chromium API behaves

PR #15014, opened 2018-10-08
feat: add remote.require() / remote.getGlobal() filtering

miniak:
##### Description of Change
Allows filtering of `remote.require()` / `remote.getGlobal()` requests by handling `remote-require` / `remote-get-global` events.

This is an alternative to the nuclear option of disabling the remote module completely (https://github.com/electron/electron/pull/13028).

Fixes https://github.com/electron/electron/issues/12932.

##### Checklist
- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

##### Release Notes
Notes: Allowed filtering of `remote.require()` / `remote.getGlobal()` requests by handling `remote-require` / `remote-get-global` events.

PR #16293, opened 2019-01-07
feat: add remote.getBuiltin() / remote.getCurrentWindow() / remote.getCurrentWebContents() filtering

miniak:
#### Description of Change
Add additional remote filtering:
- `remote.getBuiltin()` by handling `remote-get-builtin`
- `remote.getCurrentWindow()` by handling `remote-get-current-window`
- `remote.getCurrentWebContents` - by handling `remote-get-current-web-contents`

Follow up to https://github.com/electron/electron/pull/15014

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added filtering for `remote.getBuiltin()`, `remote.getCurrentWindow()` and `remote.getCurrentWebContents`.


miniak:
@MarshallOfSound, @nornagon can you please review?

miniak:
@nornagon some tests were broken after `nodeIntegration` was disabled by default. fixed those

lib/browser/rpc-server.js
```diff
@@ -309,8 +309,19 @@ handleRemoteCommand('ELECTRON_BROWSER_REQUIRE', function (event, contextId, modu
   return valueToMeta(event.sender, contextId, customEvent.returnValue)
 })
 
-handleRemoteCommand('ELECTRON_BROWSER_GET_BUILTIN', function (event, contextId, module) {
-  return valueToMeta(event.sender, contextId, electron[module])
+handleRemoteCommand('ELECTRON_BROWSER_GET_BUILTIN', function (event, contextId, moduleName) {
+  const customEvent = eventBinding.createWithSender(event.sender)
+  event.sender.emit('remote-get-builtin', customEvent, moduleName)
+
+  if (customEvent.defaultPrevented) {
```
nornagon: it's surprising to me that you have to `preventDefault()` _as well as_ setting `returnValue` to return a custom value. My expectation from reading the docs is that this would work:

```js
webContents.on('remote-get-builtin', (e, mod) => {
  if (mod === 'something') {
    e.returnValue = someWrapper()
  } else {
    e.preventDefault()
  }
})
```

I can certainly imagine it taking some head-scratching to figure out that you have to call `preventDefault` in order for your custom return value to be respected.

lib/browser/rpc-server.js
```diff
@@ -309,8 +309,19 @@ handleRemoteCommand('ELECTRON_BROWSER_REQUIRE', function (event, contextId, modu
...
+    if (typeof customEvent.returnValue === 'undefined') {
+      throw new Error(`Invalid electron module: ${moduleName}`)
```
nornagon: Perhaps this error message should say something about `Blocked` like the other ones?

spec/api-app-spec.js
```diff
@@ -425,6 +425,52 @@ describe('app module', () => {
       w.loadURL('about:blank')
       w.webContents.executeJavaScript(`require('electron').remote.getGlobal('test')`)
     })
+
+    it('should emit remote-get-builtin event when remote.getBuiltin() is invoked', (done) => {
+      app.once('remote-get-builtin', (event, webContents, moduleName) => {
+        expect(webContents).to.equal(w.webContents)
+        expect(moduleName).to.equal('app')
```
nornagon: `expect()` in a callback will cause the test to hang and timeout if it fails, i think? either a) refactor this test to use `async` and `emittedOnce` and so on, or b) wrap this in `try { ...; done() } catch (e) { done(e) }`

spec/api-app-spec.js
```diff
@@ -425,6 +425,52 @@ describe('app module', () => {
...
+        done()
+      })
+      w = new BrowserWindow({
+        show: false,
+        webPreferences: {
+          nodeIntegration: true
+        }
+      })
+      w.loadURL('about:blank')
+      w.webContents.executeJavaScript(`require('electron').remote.app`)
```
nornagon: do you need to wait for the `loadURL` to complete?

lib/browser/rpc-server.js
miniak: The other events that we already have in master and 4.0 behave like that. I can change it in a separate PR for all events consistently.

nornagon: yeah :( sad that it didn't get caught earlier.

is there a web API that behaves like the current way that you were following? perhaps i'm missing context

spec/api-app-spec.js
miniak: added

spec/api-app-spec.js
miniak: made async

lib/browser/rpc-server.js
miniak: made consistent

PR #16308, opened 2019-01-07
fix: enable autofill popups on mac

### LINKED ISSUE #360: Datalist elements do not work
I tried to use the datalist element for auto completion of an input field but I can't make it work.
I stripped down my HTML file to be sure nothing interferes. 

```
<!DOCTYPE html>
<html>
<head>
</head>
<body>
  <input type="text" list="modules" />
  <datalist id="modules">
    <select>
      <option value="IN4NWK"></option>
      <option value="IN4CDT"></option>
      <option value="IN4INS"></option>
    </select>
   </datalist>
</body>
</html>
```

My main Javascript file does nothing but open a browser window and loading this HTML. Now when you start typing an "I" or press the down key in the input field, nothing happens.

Did I overlook something? The WebKit version used in atom-shell should support it.

Regards

brenca:
#### Description of Change

Pretty much what the label says, it looks like shuffling some code around was enough to enable autofill popups on Mac. This PR also fixes the positioning (the popup appeared below the autofilled element before, now it matches chrome (if there is no space, it appears above the element)).

Resolves #360

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:  Enabled autofill popups on Mac.

### DIFF
--- a/atom/browser/ui/autofill_popup.cc
+++ b/atom/browser/ui/autofill_popup.cc
@@ -10,6 +10,7 @@
 #include "atom/browser/ui/autofill_popup.h"
 #include "atom/common/api/api_messages.h"
 #include "base/i18n/rtl.h"
+#include "chrome/browser/ui/autofill/popup_view_common.h"
 #include "electron/buildflags/buildflags.h"
 #include "ui/display/display.h"
 #include "ui/display/screen.h"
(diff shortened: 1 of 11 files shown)


atom/browser/ui/autofill_popup.cc
```diff
@@ -27,80 +28,97 @@ namespace atom {
...
-  int popup_width =
-      std::min(popup_required_width, std::max(right_available, left_available));
-
-  std::pair<int, int> grow_right(right_growth_start, popup_width);
-  std::pair<int, int> grow_left(left_growth_end - popup_width, popup_width);
-
-  // Prefer to grow towards the end (right for LTR, left for RTL). But if there
-  // is not enough space available in the desired direction and more space in
-  // the other direction, reverse it.
+struct WidthCalculationResults {
```
nornagon: Did this code come from somewhere else (e.g. //chrome?)

brenca: Yupp, that comes from https://cs.chromium.org/chromium/src/chrome/browser/ui/autofill/popup_view_common.cc?q=WidthCalculationResults&sq=package:chromium&g=0&l=31. Shall I add a comment about that?

nornagon: is there any chance we could reuse the chrome code directly rather than copying it?

brenca: Yupp, but I had to define 2 methods that the original code uses. Is it fine if those will live in the popups's file and not under `chromium_src`?

patches/common/chromium/autofill_size_calculation.patch
```diff
@@ -0,0 +1,40 @@
...
+   return container_view->GetWindowAndroid()->bounds();
+ #else
+-  views::Widget* widget =
+-      views::Widget::GetTopLevelWidgetForNativeView(container_view);
+-  if (widget)
+-    return widget->GetWindowBoundsInScreen();
++  // views::Widget* widget =
++  //     views::Widget::GetTopLevelWidgetForNativeView(container_view);
++  // if (widget)
++  //   return widget->GetWindowBoundsInScreen();
```
nornagon: It's surprising to me that this should be commented out? It doesn't look like it's calling into Chrome at all.

Also, prefer commenting in patches with `#if 0`, as it's more resilient to upstream changes.

atom/browser/ui/cocoa/views_delegate_mac.mm
```diff
@@ -16,7 +16,18 @@
...
+  if (params->native_widget)
+    return;
+
+  if (!native_widget_factory().is_null()) {
+    params->native_widget = native_widget_factory().Run(*params, delegate);
+    if (params->native_widget)
+      return;
+  }
+
+  params->native_widget = nullptr;
```
nornagon: I don't understand this, why would it have been `DCHECK`ed to be present before but it's OK for it to be `nullptr` now?

atom/browser/ui/cocoa/views_delegate_mac.mm
```diff
@@ -16,7 +16,19 @@
...
+  if (params->native_widget)
+    return;
+
+  if (!native_widget_factory().is_null()) {
+    params->native_widget = native_widget_factory().Run(*params, delegate);
+    if (params->native_widget)
+      return;
+  }
+
+  // By returning null Widget creates the default NativeWidget implementation.
```
nornagon:
```suggestion
  // Setting null here causes Widget to create the default NativeWidget implementation.
```

Issue #16303, opened 2019-01-07
v4: MenuItems with roles default registerAccelerator to false, breaking all default shortcuts


pimterry:
* Output of `node_modules/.bin/electron --version`: v4.0.1
* Operating System (Platform and Version): Ubuntu 18.04

**Expected Behavior**
The default keyboard shortcuts for menu item roles should be registered (except in special cases where they're handled elsewhere - cut/copy/etc).

**Actual behavior**
As of v4, the keyboard shortcuts for any menu item with a role are not registered by default.

**To Reproduce**
In a v4 electron app, adding the below menu item will show a Ctrl/Cmd+R accelerator in the menu, but won't actually register it, so it won't work:

```
{
  role: 'reload'
}
```

this menu item does work:

```
{
  role: 'reload',
  registerAccelerator: true
}
```

I think this is caused by https://github.com/electron/electron/pull/15723, specifically https://github.com/electron/electron/blob/a1a431eb8735ed677b2ba980ef77549230c6363b/lib/browser/api/menu-item-roles.js#L256-L258

I.e. `registerAccelerator` defaults to `true` only if there's no role for the menu item. If there is a role set, it instead uses the `registerAccelerator` setting for that role, but for almost all roles this is `undefined`.

codebytere:
Ah, thanks for pointing this out! i'll see what can be done about it 😀

codebytere:
@pimterry a fix is in flight, it should go out with the next `4-0-x` patch release i would guess

PR #16309, opened 2019-01-07
fix: register accelerator if role has no registerAccelerator

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/16303.

Previously, `registerAccelerator` defaulted to `true` only if there was no role for the menu item. If there is a role set, it instead used the `registerAccelerator` property for that role; however many of the roles had an `undefined` `registerAccelerator`. 

This therefore changes behavior such that a `MenuItem` will only be set to `roles[role].registerAccelerator` if it has a role AND that role has a defined `registerAccelerator` property. 

/cc @brenca

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue whereby `registerAccelerator` was being set to the `registerAccelerator` property of its role even when `registerAccelerator` was `undefined`.

### DIFF
--- a/lib/browser/api/menu-item-roles.js
+++ b/lib/browser/api/menu-item-roles.js
@@ -254,7 +254,8 @@ exports.getDefaultAccelerator = (role) => {
 }
 
 exports.shouldRegisterAccelerator = (role) => {
-  return roles.hasOwnProperty(role) ? roles[role].registerAccelerator : true
+  const hasRoleRegister = roles.hasOwnProperty(role) && roles[role].registerAccelerator
+  return hasRoleRegister ? roles[role].registerAccelerator : true
 }
 
 exports.getDefaultSubmenu = (role) => {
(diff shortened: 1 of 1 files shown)


lib/browser/api/menu-item-roles.js
```diff
@@ -254,7 +254,8 @@ exports.getDefaultAccelerator = (role) => {
 }
 
 exports.shouldRegisterAccelerator = (role) => {
-  return roles.hasOwnProperty(role) ? roles[role].registerAccelerator : true
+  const hasRoleRegister = roles.hasOwnProperty(role) && roles[role].registerAccelerator
```
pimterry: This will be `false` if `roles[role].registerAccelerator` is not defined (correct), but also if it's been explicitly set to `false`.

That's exactly how this is actually used unfortunately - to disable registering accelerators on the cut/copy/paste/etc roles - so this will cause some problems.

Needs to be something like `roles[role].hasOwnProperty('registerAccelerator')` instead I think.

codebytere: oh oops i meant to check explicitly that it's `undefined` - fixing

PR #16310, opened 2019-01-07
feat: use default-app behavior in packaged apps

miniak:
#### Description of Change
unify the behavior between default app and packaged apps:
- create default application menu unless the app has one
- default `window-all-closed` handling unless the app handles the event

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Unified behavior between the default app and packaged apps (application menu / `window-all-closed` handling)


lib/browser/init.js
```diff
@@ -184,5 +184,20 @@ if (currentPlatformSupportsAppIndicator()) {
...
+// Quit when all windows are closed and no other one is listening to this.
+app.on('window-all-closed', () => {
+  if (app.listenerCount('window-all-closed') === 1) {
+    app.quit()
+  }
+})
+
+// Create default menu.
+app.once('ready', () => {
+  if (!Menu.getApplicationMenu()) {
```
MarshallOfSound: Your point in the call this morning gave me some thoughts on this, maybe there should be an explicit way to opt-out of this default without setting an empty menu.

E.g. Setting `process.noDefaultMenu` to `true` (similar to our `process.noAsar`)

miniak: @nornagon, @codebytere what do you think?

lib/browser/default-menu.js
```diff
@@ -0,0 +1,174 @@
...
+  const fileMenu = {
+    label: 'File',
+    submenu: [{
+      role: isMac ? 'close' : 'quit'
+    }]
+  }
+
+  const editMenu = {
+    label: 'Edit',
+    submenu: [
```
sindresorhus: Why not use the `editMenu` role here?

> editMenu - Whole default "Edit" menu (Undo, Copy, etc.).

miniak: good point! used that + added roles for several other default menus (app, view)

lib/browser/init.js
miniak: done

docs/api/menu.md
```diff
@@ -30,6 +30,9 @@ effect on macOS.
 
 **Note:** This API has to be called after the `ready` event of `app` module.
 
+**Note:** Default menu is created automatically if the app does not set one.
+You can also set `process.noDefaultMenu` to `true` to disable this behavior.
```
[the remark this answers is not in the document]
miniak: cc @MarshallOfSound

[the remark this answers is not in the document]
sindresorhus: I agree. “No”-prefixed options names are an anti-pattern. Also, wouldn’t this fit better on the `app` namespace?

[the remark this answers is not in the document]
miniak: I've removed it completely for now. The default menu can be suppressed by calling `Menu.setApplicationMenu(null)`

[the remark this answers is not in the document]
miniak: also the comment
> **Note:** This API has to be called after the `ready` event of `app` module.
does not seem accurate. `Menu.getApplicationMenu` can definitely be called before `ready`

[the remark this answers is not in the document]
miniak: in that case `Menu.setApplicationMenu(null)` would be enough for suppressing the default menu. Otherwise something like `app.useDefaultMenu` would be necessary as the default menu would be created first and then immediately overwritten by the app menu, which is not efficient.

lib/browser/init.js
```diff
@@ -184,5 +184,19 @@ if (currentPlatformSupportsAppIndicator()) {
...
+app.on('window-all-closed', () => {
+  if (app.listenerCount('window-all-closed') === 1) {
+    app.quit()
+  }
+})
+
+const { setDefaultApplicationMenu } = require('@electron/internal/browser/default-menu')
+
+// Create default menu.
+app.once('ready', () => {
```
nornagon:
```suggestion
app.once('ready', setDefaultApplicationMenu)
```

docs/api/menu.md
```diff
@@ -26,9 +26,9 @@ opens the associated menu. The indicated character in the button label gets an
 underline. The `&` character is not displayed on the button label.
 
 Passing `null` will remove the menu bar on Windows and Linux but has no
-effect on macOS.
+effect on macOS (it will suppress the default menu, however).
 
-**Note:** This API has to be called after the `ready` event of `app` module.
+**Note:** Default menu is created automatically if the app does not set one.
```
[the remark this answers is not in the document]
nornagon: Also it'd be great to get some more details on what the default menu actually is.

default_app/main.js
```diff
@@ -142,6 +128,9 @@ function startRepl () {
     process.exit(1)
   }
 
+  // prevent quitting
+  app.on('window-all-closed', () => {})
```
nornagon: wait... this doesn't seem right? I don't think we want to prevent quitting in the default app?

spec/fixtures/api/default-menu/main.js
```diff
@@ -0,0 +1,20 @@
...
+}
+
+app.on('ready', () => {
+  setImmediate(() => {
+    process.stdout.write(JSON.stringify(Menu.getApplicationMenu() === expectedMenu))
+    process.stdout.end()
+
+    app.quit()
+  })
+})
```
nornagon: please wrap this whole thing in a try/catch, incl. inside the `setImmediate`, so that the test will fail rather than timeout if something in here `throw`s. Makes things much easier to debug.

default_app/main.js
miniak: This is REPL mode, it's supposed to keep running after all windows are closed.

miniak: It was handled by the `option.interactive` check before

lib/browser/init.js
miniak: done

spec/fixtures/api/default-menu/main.js
miniak: done

docs/api/menu.md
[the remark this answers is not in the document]
miniak: added something. is it good or should I add even more details?

docs/api/menu.md
```diff
@@ -25,10 +25,11 @@ indicate which letter should get a generated accelerator. For example, using
 opens the associated menu. The indicated character in the button label gets an
 underline. The `&` character is not displayed on the button label.
 
-Passing `null` will remove the menu bar on Windows and Linux but has no
-effect on macOS.
+Passing `null` will remove the menu bar on Windows and Linux. On all platforms
+it will suppress the default menu (including macOS).
```
nornagon: This is a bit confusing, how about:

> Passing `null` will suppress the default menu. On Windows and Linux, this has the additional effect of removing the menu bar from the window.

PR #16329, opened 2019-01-08
fix: properly determine if WebContents is offscreen in WebContentsDelegate

### LINKED ISSUE #16300: macOS html file picker dialog behaviour confusing
* Output of `node_modules/.bin/electron --version`: 4.0.1
* Operating System (Platform and Version): macOS Mojave
* Output of `node_modules/.bin/electron --version` on last known working Electron version (if applicable): 2.0.16

**Expected Behavior**
When we open the file picker using the html input file tag, we expect to see a modal or a dialog that is always on top unless closed (like in Chrome).

**Actual behavior**
In the 3.x & 4.x versions of Electron, the dialog that appears for file picking using html input file tag is a dialog that behaves more like a regular window. Problem with this is that if the window is put to the background and we focus on our main app window, then try file selection again, the existing window file picker window doesn't come to the foreground either or a new dialog is created. This is very confusing for a user who assumes that the file picking html button is not working.

Please see video below to understand better.

**To Reproduce**

```sh
$ git clone $https://github.com/VishwasShashidhar/electron-quick-start -b file-picker
$ npm install
$ npm start || electron .
```
**Screenshots**
screenrecording20190107at1

**Additional Information**
This seems to be a behavioural change from https://github.com/electron/electron/commit/c50a460ce68d4b1e938e898c1b6a4d0db31164b9#diff-68a7264c657f5cfda99d3cdd12649277

I've also tested by reversing the above change and things work as expected.

BinaryMuse:
#### Description of Change

`WebContentsDelegate::InitWithWebContents` previously used the following code to set `offscreen_`:

```c++
  // Determien whether the WebContents is offscreen.
  auto* web_preferences = WebContentsPreferences::From(web_contents);
  offscreen_ =
      !web_preferences || web_preferences->IsEnabled(options::kOffscreen);
```

However, `options::kOffscreen` defaults to `false` when its not set, so when `web_preferences` was `null`, `offscreen_` would be set to `true`. The most immediate result of this is that `RunModalDialog` would have `settings.force_detached` set in the following code, causing the dialog to be shown as a regular modal instead of a sheet:

```objectivec
// Run modal dialog with parent window and return user's choice.
int RunModalDialog(NSSavePanel* dialog, const DialogSettings& settings) {
  __block int chosen = NSFileHandlingPanelCancelButton;
  if (!settings.parent_window || !settings.parent_window->GetNativeWindow() ||
      settings.force_detached) {
    chosen = [dialog runModal];
  } else {
    NSWindow* window = settings.parent_window->GetNativeWindow();

    [dialog beginSheetModalForWindow:window
                   completionHandler:^(NSInteger c) {
                     chosen = c;
                     [NSApp stopModal];
                   }];
    [NSApp runModalForWindow:window];
  }

  return chosen;
}
```

This change defaults `offscreen_` to false when `web_preferences` is `null`.

/cc @VishwasShashidhar

Fixes #16300 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Changed the file dialog presented by inputs with the type `file` to use sheets on macOS instead of detached modals.

### DIFF
--- a/atom/browser/common_web_contents_delegate.cc
+++ b/atom/browser/common_web_contents_delegate.cc
@@ -187,7 +187,7 @@ void CommonWebContentsDelegate::InitWithWebContents(
   // Determien whether the WebContents is offscreen.
   auto* web_preferences = WebContentsPreferences::From(web_contents);
   offscreen_ =
-      !web_preferences || web_preferences->IsEnabled(options::kOffscreen);
+      web_preferences && web_preferences->IsEnabled(options::kOffscreen);
 
   // Create InspectableWebContents.
   web_contents_.reset(InspectableWebContents::Create(


BinaryMuse:
Interestingly, I'm seeing `web_preferences` being null even though I set `webPreferences` when I created the `BrowserWindow`. This seems unexpected to me, but I may be missing something.

VishwasShashidhar:
Confirmed that it is fixed on master and we see a sheet. Thanks @BinaryMuse @codebytere

PR #16330, opened 2019-01-08
fix: emit IPC event in correct context if isolation and sandbox enabled

### LINKED ISSUE #11922: Some ipc messages are not sent when contextIsolation and sandbox are set to true
* Electron version: 1.8.2
* Operating system: MacOS, Windows10

### Expected behavior

ipc messages from the main process should be sent to the renderer process when both `sandbox` and `contextIsolation` are set.

### Actual behavior

ipc messages sent in listeners to webview events are not visible in the renderer process.

### How to reproduce

```
$ git clone https://gist.github.com/4f6c4450cff194838e746ce23e60116d.git
$ cd 4f6c4450cff194838e746ce23e60116d/
$ electron main.js 
```

The devtools will open and the console will log one message.  But the second message on the ipc `ping` channel is not observed.

Toggling either `sandbox` or `contextIsolation` to false allows the `ping` to travel to the renderer process.  In the 1.7.X branch setting both values to true still allows the message to pass through successfully.

aegarbutt:
#### Description of Change

IPC events were not being delivered to renderer processes when both
`contextIsolation` and `sandbox` were enabled. This is because the
`AtomSandboxedRenderFrameObserver` class was incorrectly using the
`MainWorldScriptContext`, rather than conditionally selecting the
context based on if isolation was enabled.

Fixes #11922

cc @tarruda 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed delivery of IPC events to renderer processes when both `contextIsolation` and `sandbox` are enabled.

### DIFF
--- a/spec/api-ipc-renderer-spec.js
+++ b/spec/api-ipc-renderer-spec.js
@@ -157,7 +157,7 @@ describe('ipc renderer module', () => {
       contents.loadFile(path.join(fixtures, 'pages', 'ping-pong.html'))
     })
 
-    it('sends message to WebContents (sanboxed renderer)', done => {
+    it('sends message to WebContents (sandboxed renderer)', done => {
       contents = webContents.create({
         preload: path.join(fixtures, 'module', 'preload-inject-ipc.js'),
         sandbox: true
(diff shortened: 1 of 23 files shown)


aegarbutt:
I think those CI checks are false positives or otherwise unrelated to this pull request.

aegarbutt:
Aside: This would be a great bug-fix to backport to the 2.x and 3.x lines.

spec/api-ipc-renderer-spec.js
```diff
@@ -177,6 +177,47 @@ describe('ipc renderer module', () => {
       contents.loadFile(path.join(fixtures, 'pages', 'ping-pong.html'))
     })
 
+    it('sends message to WebContents (context isolation)', done => {
```
miniak: please use the `generateSpecs` pattern as here [spec/api-browser-window-spec.js](https://github.com/electron/electron/blob/744f737cba67cd6f59ea598aa00795fa4375a929/spec/api-browser-window-spec.js#L1426) to avoid duplicate code

miniak: also modify the existing test to use the new preload script

miniak: another example in [spec/api-crash-reporter-spec.js](https://github.com/electron/electron/blob/744f737cba67cd6f59ea598aa00795fa4375a929/spec/api-crash-reporter-spec.js#L191)

miniak: I've modified the tests for you: https://github.com/electron/electron/commit/8c9589bd9ed3f9e979194670e378579ab158bb86

aegarbutt: Thanks so much!

aegarbutt: Sorry, I'm not a frequent GItHub user (Phabricator at work), do I need to update the pull request with your commit?

miniak: Yes. The other option would be to abandon this PR and create a new one from my branch.

miniak: Or merge this as is and refactor the tests in a follow up PR.

aegarbutt: I seem to have mucked things up, so I'm just going to abandon this and create a new pull request from your branch.

PR #16333, opened 2019-01-09
chore: pass dryRun arg correctly

codebytere:
#### Description of Change

We were passing `--dry-run` instead of `--dryRun`

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


MarshallOfSound:
Are we sure that nothing else is reading this output (spawning and reading) elsewhere in our scripts, maybe in an old python one 🤔 ?

codebytere:
No, the only thing that does since we moved to sudowoodo is `sudowoodo/src/utils/get-bump-version.ts`

PR #16342, opened 2019-01-09
fix: return after handling Esc for fullscreen in WebContentsDelegate

nornagon:
#### Description of Change
It doesn't seem to me like we should be continuing to handle the key event after exiting fullscreen. I'm not sure what behaviour that might result in, though, so here's a PR to discuss it!

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes


jkleinsc:
Merging as arm CI failure is known issue unrelated to this change.

PR #16343, opened 2019-01-10
fix: improve release notes

ckerr:
#### Description of Change

Addresses some of the release-notes script issues that we found during the 4.0.0 rollout:

* if we're making a stable release, omit notes that have already been released in older branches
* sniff semantic commit types from PR subjects, not just from commit messages
* don't honor unrecognized semantic commit types
* add an option to specify the version name to be shown in the notes

also, some housekeeping:

* Don't hardcode the release-notes' placeholder comment
* Update prepare-release.js to pass in the target version number

@codebytere @BinaryMuse 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/release-notes/notes.js
+++ b/script/release-notes/notes.js
@@ -63,6 +63,18 @@ const setPullRequest = (commit, owner, repo, number) => {
   }
 }
 
+// copied from https://github.com/electron/clerk/blob/master/src/index.ts#L4-L13
+const OMIT_FROM_RELEASE_NOTES_KEYS = [
+  'no-notes',
+  'no notes',
+  'no_notes',
+  'none',
+  'no',
+  'nothing',
+  'empty',
+  'blank'
+]
+
 const getNoteFromBody = body => {
   if (!body) {
     return null
@@ -83,13 +95,8 @@ const getNoteFromBody = body => {
       .trim()
   }
 
-  if (note) {
-    if (note.match(/^[Nn]o[ _-][Nn]otes\.?$/)) {
-      return NO_NOTES
-    }
-    if (note.match(/^[Nn]one\.?$/)) {
-      return NO_NOTES
-    }
+  if (note && OMIT_FROM_RELEASE_NOTES_KEYS.includes(note.toLowerCase())) {
+    return NO_NOTES
   }
 
   return note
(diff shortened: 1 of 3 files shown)


script/release-notes/notes.js
```diff
@@ -437,37 +438,81 @@ const getNotes = async (fromRef, toRef) => {
...
+
+    // if we still don't have a note, it's because someone missed a 'Notes:
+    // comment in a PR somewhere... use the PR subject as a fallback.
+    commit.note = commit.note || prSubject
   }
 
-  // remove uninteresting commits
+  // remove procedural commits
   pool.commits = pool.commits
     .filter(commit => commit.note !== NO_NOTES)
```
MarshallOfSound: Just noticed this here but there are a multitude of things that correspond to `No Notes`.

There is a complete list here --> https://github.com/electron/clerk/blob/master/src/index.ts#L4-L13

script/release-notes/notes.js
```diff
@@ -437,37 +438,81 @@ const getNotes = async (fromRef, toRef) => {
...
+      pr = prParsed.pr && (prParsed.pr.number !== pr.number) ? prParsed.pr : null
     }
+
+    // if we still don't have a note, it's because someone missed a 'Notes:
+    // comment in a PR somewhere... use the PR subject as a fallback.
+    commit.note = commit.note || prSubject
   }
 
-  // remove uninteresting commits
+  // remove procedural commits
```
codebytere: nit: procedural is a bit vague: "non-user-facing" maybe?

script/release-notes/notes.js
```diff
@@ -437,37 +438,81 @@ const getNotes = async (fromRef, toRef) => {
...
+          .reduce(reducer, null)
+        if (!semver.valid(earliestRelease)) {
+          return true
+        }
+        return semver.diff(earliestRelease, newVersion).includes('patch')
+      })
+  }
+
   const notes = {
     breaks: [],
```
codebytere: s/breaks/breaking?

ckerr: 7efd5f594c7d65fc6d32e28fd6af2ff9c4ea28b2

script/release-notes/notes.js
ckerr: 4c6510811dea1f75507ab86d348a1f1b3d829fee

script/release-notes/notes.js
ckerr: e5ee608d57b10fa91d695ea014973b5800ca2f2e

PR #16352, opened 2019-01-10
fix: emit IPC event in correct context if isolation and sandbox enabled

### LINKED ISSUE #11922: Some ipc messages are not sent when contextIsolation and sandbox are set to true

aegarbutt:
#### Description of Change

IPC events were not being delivered to renderer processes when both
`contextIsolation` and `sandbox` were enabled. This is because the
`AtomSandboxedRenderFrameObserver` class was incorrectly using the
`MainWorldScriptContext`, rather than conditionally selecting the
context based on if isolation was enabled.

Fixes #11922

cc @miniak  

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed delivery of IPC events to renderer processes when both `contextIsolation` and `sandbox` are enabled.


aegarbutt:
I made a mess of #16330, so I'm trying again here. I included your recommended test changes.

Thank you for your patience with me.

miniak:
@nornagon, @codebytere can you please review? This is fixing quite a bad bug.

PR #16369, opened 2019-01-11
build: fail a build if "npm install" fails

alexeykuzmin:
#### Description of Change

It can fail you know.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -107,7 +107,7 @@ hooks = [
     'action': [
       'python',
       '-c',
-      'import os; os.chdir("src"); os.chdir("electron"); os.system("npm install")',
+      'import os, subprocess; os.chdir(os.path.join("src", "electron")); subprocess.check_call(["python", "script/lib/npm.py", "install"]);',
     ],
   },
   {
@@ -117,7 +117,7 @@ hooks = [
     'action': [
       'python',
       '-c',
-      'import os; os.chdir("src"); os.chdir("electron"); os.chdir("vendor"); os.chdir("boto"); os.system("python setup.py build");',
+      'import os, subprocess; os.chdir(os.path.join("src", "electron", "vendor", "boto")); subprocess.check_call(["python", "setup.py", "build"]);',
     ],
   },
   {
@@ -127,9 +127,9 @@ hooks = [
     'action': [
       'python',
       '-c',
-      'import os; os.chdir("src"); os.chdir("electron"); os.chdir("vendor"); os.chdir("requests"); os.system("python setup.py build");',
+      'import os, subprocess; os.chdir(os.path.join("src", "electron", "vendor", "requests")); subprocess.check_call(["python", "setup.py", "build"]);',
     ],
-  }
+  },
 ]
 
 recursedeps = [
--- a/script/lib/npm.py
+++ b/script/lib/npm.py
@@ -1,0 +1,18 @@
+import subprocess
+import sys
+
+
+def npm(*npm_args):
+    call_args = [__get_executable_name()] + list(npm_args)
+    subprocess.check_call(call_args)
+
+
+def __get_executable_name():
+    executable = 'npm'
+    if sys.platform == 'win32':
+        executable += '.cmd'
+    return executable
+
+
+if __name__ == '__main__':
+    npm(*sys.argv[1:])


DEPS
```diff
@@ -107,7 +107,7 @@ hooks = [
     'action': [
       'python',
       '-c',
-      'import os; os.chdir("src"); os.chdir("electron"); os.system("npm install")',
+      'import os; os.chdir("src"); os.chdir("electron"); exit_code = os.system("npm instal"); import sys; sys.exit(exit_code)',
```
alexeykuzmin: FIXME `instal`

alexeykuzmin:
@nornagon 
The typo has been made intentionally to make the `npm` command fail and exit with code `1`.
The problem is that it didn't work
https://circleci.com/gh/electron/electron/115735
```
________ running '/usr/bin/python -c import os; os.chdir("src"); os.chdir("electron"); exit_code = os.system("npm instal"); import sys; sys.exit(exit_code)' in '/home/builduser/project'

Usage: npm <command>

<...>

Did you mean one of these?
    install
    unstar
    uninstall

<...>

Running hooks: 100% (79/79), done.
```

I'll replace all `os.system()` calls with `subprocess.*` calls.

alexeykuzmin:
@nornagon 
The typo has been made intentionally to make the command fail and exit with code `1`.
The problem is that it didn't work
https://circleci.com/gh/electron/electron/115735
```
________ running '/usr/bin/python -c import os; os.chdir("src"); os.chdir("electron"); exit_code = os.system("npm instal"); import sys; sys.exit(exit_code)' in '/home/builduser/project'

Usage: npm <command>

<...>

Did you mean one of these?
    install
    unstar
    uninstall

<...>

Running hooks: 100% (79/79), done.
```

I'll replace all `os.system()` calls with `subprocess.`

alexeykuzmin:
> I'll replace all os.system() calls with subprocess.* calls.

Done. @nornagon, please re-review.

alexeykuzmin:
> I'll replace all os.system() calls with subprocess.* calls.

Done.
Now it works: https://circleci.com/gh/electron/electron/116496?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

DEPS
```diff
@@ -107,7 +107,7 @@ hooks = [
     'action': [
       'python',
       '-c',
-      'import os; os.chdir("src"); os.chdir("electron"); os.system("npm install")',
+      'import os, subprocess, sys; os.chdir(os.path.join("src", "electron")); npm = "npm.cmd" if sys.platform == "win32" else "npm"; subprocess.check_call([npm, "install"]);',
```
nornagon: how come `npm.cmd` is now needed on windows? afaik just `npm` was working fine previously.

If this is truly needed, then this has become too complicated for a one-liner & should be moved to a script.

alexeykuzmin: > `npm` was working fine previously

Only when you use `os.system()`.

`subprocess.call()` needs different executable names on different platforms, see here for example https://github.com/electron/electron/blob/52fe92d02ef94712fc7b19b3f03d3363cfa71a4d/script/lib/util.py#L28-L30

alexeykuzmin: I can move the `npm` call to a Python script.
You think it would be better to move the whole thing to a new script?

nornagon: IMO the `if` statement is enough to push it over the edge of "shouldn't be inlined in a string in a config file". The only reason these are in Python instead of just being direct commands is because there's no way to `cd` in a cross-platform way, and npm doesn't have a `-C` argument like `make` and `ninja`.

I'd be fine with having a `script/cd-npm.py` that took a directory as its first argument and passed the rest of its arguments directly to `npm` (or `npm.cmd`).

alexeykuzmin: @nornagon Any thoughts?

DEPS
```diff
@@ -107,7 +107,7 @@ hooks = [
     'action': [
       'python',
       '-c',
-      'import os; os.chdir("src"); os.chdir("electron"); os.system("npm install")',
+      'import os, subprocess; os.chdir(os.path.join("src", "electron")); subprocess.check_call(["python", "script/lib/npm.py", "install"]);',
```
nornagon: it's weird to invoke python to cd and then invoke python :/

Would [the --prefix option](https://docs.npmjs.com/misc/config#prefix) work? i.e. `npm install --prefix src/electron`

alexeykuzmin: Works on [Linux](https://circleci.com/gh/electron/electron/118669?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), but doesn't work on [Windows](https://windows-ci.electronjs.org/project/AppVeyor/electron-i8wjp/build/1.0.4736).

PR #16374, opened 2019-01-12
fix: crash when calling setProgressBar on macOS

zcbenz:
#### Description of Change

Depending on the version of macOS and macOS SDK, the system may install a default contentView for the dock, resulting in crashes like https://github.com/electron/electron/issues/14008.

This PR makes the code correctly check if a progress bar has been installed in dock.

I'm not sure the exact condition to trigger the crash, and it is weird that there isn't much people seeing it. But I'm seeing it on my machine, and I can verify this PR fixes it on my machine.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix crash when calling setProgressBar on macOS

### DIFF
--- a/atom/browser/native_window_mac.mm
+++ b/atom/browser/native_window_mac.mm
@@ -1127,15 +1127,20 @@ void NativeWindowMac::SetProgressBar(double progress,
                                      const NativeWindow::ProgressState state) {
   NSDockTile* dock_tile = [NSApp dockTile];
 
+  // Sometimes macOS would install a default contentView for dock, we must
+  // verify whether NSProgressIndicator has been installed.
+  bool first_time = !dock_tile.contentView ||
+                    [[dock_tile.contentView subviews] count] == 0 ||
+                    ![[[dock_tile.contentView subviews] lastObject]
+                        isKindOfClass:[NSProgressIndicator class]];
+
   // For the first time API invoked, we need to create a ContentView in
   // DockTile.
-  if (dock_tile.contentView == nullptr) {
+  if (first_time) {
     NSImageView* image_view = [[NSImageView alloc] init];
     [image_view setImage:[NSApp applicationIconImage]];
     [dock_tile setContentView:image_view];
-  }
 
-  if ([[dock_tile.contentView subviews] count] == 0) {
     NSProgressIndicator* progress_indicator = [[AtomProgressBar alloc]
         initWithFrame:NSMakeRect(0.0f, 0.0f, dock_tile.size.width, 15.0)];
     [progress_indicator setStyle:NSProgressIndicatorBarStyle];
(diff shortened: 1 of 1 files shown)


atom/browser/native_window_mac.mm
```diff
@@ -1127,15 +1127,20 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
                                      const NativeWindow::ProgressState state) {
   NSDockTile* dock_tile = [NSApp dockTile];
 
+  // Sometimes macOS would install a default contentView for dock, we must
+  // verify whether NSProgressIndicator has been installed.
+  bool first_time = !dock_tile.contentView ||
+                    [[dock_tile.contentView subviews] count] == 0 ||
+                    ![[[dock_tile.contentView subviews] lastObject]
+                        isKindOfClass:[NSProgressIndicator class]];
```
nornagon: Can we record this explicitly rather than using the state of the dock tile as an implicit state variable? e.g. `has_installed_indicator_` or something? & perhaps store a reference to the `NSProgressIndicator`?

zcbenz: The `dockTile.contentView` is not guaranteed to keep unchanged once after setting, in my case simply refreshing the tests window would reset `dockTile.contentView` to some kind of default view.

So we could not effectively track whether the progress bar is installed in `dockTile.contentView`, as the change was made by macOS or Chromium or something else.

This only happens on very few machines, I doubt it is some bug of macOS.

atom/browser/native_window_mac.mm
```diff
@@ -1127,28 +1127,34 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
...
+
   // For the first time API invoked, we need to create a ContentView in
   // DockTile.
-  if (dock_tile.contentView == nullptr) {
-    NSImageView* image_view = [[NSImageView alloc] init];
+  if (first_time) {
+    base::scoped_nsobject<NSImageView> image_view([[NSImageView alloc] init]);
```
nornagon: I don't think there's much point in using `scoped_nsobject` if you're going to immediately autorelease it. Just:
```suggestion
    NSImageView* image_view = [[[NSImageView alloc] init] autorelease];
```

atom/browser/native_window_mac.mm
```diff
@@ -1127,28 +1127,34 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
...
     [image_view setImage:[NSApp applicationIconImage]];
-    [dock_tile setContentView:image_view];
-  }
+    [dock_tile setContentView:image_view.autorelease()];
 
-  if ([[dock_tile.contentView subviews] count] == 0) {
-    NSProgressIndicator* progress_indicator = [[AtomProgressBar alloc]
-        initWithFrame:NSMakeRect(0.0f, 0.0f, dock_tile.size.width, 15.0)];
+    base::scoped_nsobject<NSProgressIndicator> progress_indicator(
```
nornagon: same here.

PR #16387, opened 2019-01-14
fix: windowmenu role did not show the window list on macOS

miniak:
#### Description of Change
Fix https://github.com/electron/electron/issues/15912 properly.
The default window menu regressed in https://github.com/electron/electron/pull/15930.
The whole menu needs to have `window` role set, as the OS appends the window list at the end of the submenu instead of replacing a single menu item. In case of `windowmenu` role, we cannot change the role, so we need to check the other role name in the menu controller.

cc @codebytere

#### Checklist
- [ ] `npm test` passes

#### Release Notes
Notes: Fixed `windowmenu` role not showing the window list on macOS


miniak:
@codebytere should we backport some version of the `windowmenu` fix into release branches?

codebytere:
@miniak it just needs to go into `4-0-x`

PR #16396, opened 2019-01-14
feat: allow setting autoplayPolicy in webPreferences

nornagon:
#### Description of Change
Chrome 71 shipped with the autoplay policy set to require user gesture by default. I don't think this default makes sense for electron, so I've made the default in electron be kNoUserGestureRequired, but I can imagine app devs wanting to set a different option. This exposes that API as `{autoplayPolicy: "user-gesture-required"}` in webPreferences.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Allowed configuring of media autoplay policy via webPreferences.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -371,6 +371,18 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       English and not localized.
     * `navigateOnDragDrop` Boolean (optional) - Whether dragging and dropping a
       file or link onto the page causes a navigation. Default is `false`.
+    * `autoplayPolicy` String (optional) - Autoplay policy to apply to
+      content in the window. Defaults to `'no-user-gesture-required'`. Valid
+      values are:
+      * `'no-user-gesture-required'` - no user gesture is required before
+        playing media on a page. This is the default behavior in Electron.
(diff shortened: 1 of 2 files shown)


nornagon:
wip because no docs yet.

docs/api/browser-window.md
```diff
@@ -371,6 +371,18 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       English and not localized.
     * `navigateOnDragDrop` Boolean (optional) - Whether dragging and dropping a
       file or link onto the page causes a navigation. Default is `false`.
+    * `autoplayPolicy` String (optional) - Autoplay policy to apply to
```
miniak: can you make it an enum?
such as `options.mode` in `contents.openDevTools`?
```
  * `mode` String - Opens the devtools with specified dock state, can be
    `right`, `bottom`, `undocked`, `detach`. Defaults to last used dock state.
    In `undocked` mode it's possible to dock back. In `detach` mode it's not.
```

MarshallOfSound: Your current syntax should work @nornagon but would lose info about each value in the definitions anyway.  

I'll look into why this isn't being picked up as an enum

PR #16420, opened 2019-01-16
feat: add contents.ipc (EventEmitter), which allows handling messages for given webContents

miniak:
#### Description of Change
- The motivation is to simplify application code, when messages only from a particular `webContents` instance are being handled. 
- This is especially handy when using `once`, where doing the `event.sender === someWebContents` check can't be done in the listener, as it would be unregistered even if we don't handle the message.
- We cannot just emit the IPC events on the `webContents` directly as it could clash with existing events.
- It allows to use `webContents.ipc.removeAllListeners()`, which is safe(r) unlike `ipcMain.removeAllListeners()`

example
```js
const win = new BrowserWindow()
...

// this does not work properly
ipcMain.once('ready', (event) => {
  if (event.sender === win.webContents) {
    console.log('ready')
  }
})

// this is annoying
ipcMain.on('ready', function handler (event) => {
  if (event.sender === win.webContents) {
    console.log('ready')
    ipcMain.removeListener('ready', handler)
  }
})

// this is much better
win.webContents.ipc.once('ready', () => {
  console.log('ready')
})
```

#### Checklist
- [ ] `npm test` passes

#### Release Notes
Notes: Added `contents.ipc` (`EventEmitter`), which allows handling messages for given `webContents`

### DIFF
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -1617,6 +1617,11 @@ when the page becomes backgrounded. This also affects the Page Visibility API.
 
 A `Integer` representing the unique ID of this WebContents.
 
+#### `contents.ipc`
+
+An [`EventEmitter`](https://nodejs.org/api/events.html#events_class_events_eventemitter),
+which allows handling IPC messages for this `contents` only.
+
 #### `contents.session`
 
 A [`Session`](session.md) used by this webContents.
--- a/lib/browser/api/web-contents.js
+++ b/lib/browser/api/web-contents.js
@@ -337,7 +337,13 @@ WebContents.prototype._init = function () {
   this.capturePage = deprecate.promisify(this.capturePage, 1)
 
   // Dispatch IPC messages to the ipc module.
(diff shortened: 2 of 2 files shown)


miniak:
@MarshallOfSound `ipc-message` for webContents is not documented, it's an implementation detail basically. And it's not as convenient to use as you cannot directly lister for individual messages.

miniak:
@MarshallOfSound also you would have to handle `ipc-message-sync` for sync messages separately. In this case we are also adding the `returnValue` property on the `event` in the internal code before doing `ipcMain.emit`

nornagon:
I agree with @MarshallOfSound that this seems like what the `ipc-message` event is for. I'd be interested to see a PR to clean up & refactor & document that event. I don't think it's a problem that `ipc-message-sync` is a separate event—synchronous messages are substantially different to asynchronous ones and you should know which kind you're dealing with.

miniak:
Replaced by https://github.com/electron/electron/pull/16468

PR #16425, opened 2019-01-17
feat: preloads and nodeIntegration in iframes

MarshallOfSound:
#### Description of Change

This adds support for running preload scripts and nodeIntegration in sub-frames (iframes)

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added support for running preload scripts and nodeIntegration in iframes

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -371,6 +371,14 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       English and not localized.
     * `navigateOnDragDrop` Boolean (optional) - Whether dragging and dropping a
       file or link onto the page causes a navigation. Default is `false`.
+    * `nodeSupportInSubFrames` Boolean (optional) - Experimental option for
+      enabling NodeJS support in sub-frames such as iframes.  This may have
+      an increased chance of memory leaks.  All your preloads will load for
+      every iframe, you can use `process.isMainFrame` to determine if you are
+      in the main frame or not.  If you have `nodeIntegration=false` and
+      `nodeSupportInSubFrames=true` your preloads will still run in iframes
+      with node enabled but nodeIntegration will be disabled for all iframe
+      content just like your main frame.
 
 When setting minimum or maximum window size with `minWidth`/`maxWidth`/
 `minHeight`/`maxHeight`, it only constrains the users. It won't prevent you from
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -255,6 +255,10 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
     * `nodeIntegrationInWorker` Boolean (optional) - Whether node integration is
       enabled in web workers. Default is `false`. More about this can be found
       in [Multithreading](../tutorial/multithreading.md).
+    * `nodeIntegrationInSubFrames` Boolean (optional) - Experimental option for
+      enabling NodeJS support in sub-frames such as iframes. All your preloads will load for
+      every iframe, you can use `process.isMainFrame` to determine if you are
+      in the main frame or not.
     * `preload` String (optional) - Specifies a script that will be loaded before other
       scripts run in the page. This script will always have access to node APIs
       no matter whether node integration is turned on or off. The value should
@@ -371,14 +375,6 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       English and not localized.
     * `navigateOnDragDrop` Boolean (optional) - Whether dragging and dropping a
       file or link onto the page causes a navigation. Default is `false`.
-    * `nodeSupportInSubFrames` Boolean (optional) - Experimental option for
-      enabling NodeJS support in sub-frames such as iframes.  This may have
-      an increased chance of memory leaks.  All your preloads will load for
-      every iframe, you can use `process.isMainFrame` to determine if you are
-      in the main frame or not.  If you have `nodeIntegration=false` and
-      `nodeSupportInSubFrames=true` your preloads will still run in iframes
-      with node enabled but nodeIntegration will be disabled for all iframe
-      content just like your main frame.
 
 When setting minimum or maximum window size with `minWidth`/`maxWidth`/
 `minHeight`/`maxHeight`, it only constrains the users. It won't prevent you from
--- a/docs/api/process.md
+++ b/docs/api/process.md
@@ -59,6 +59,11 @@ process.once('loaded', () => {
 A `Boolean`. When app is started by being passed as parameter to the default app, this
 property is `true` in the main process, otherwise it is `undefined`.
 
+### `process.isMainFrame`
+
+A `Boolean`.  Set to `true` when the current renderer context is the "main" renderer
+frame.  If you want the ID of the current frame you should use `webFrame.routingId`.
+
 ### `process.mas`
 
 A `Boolean`. For Mac App Store build, this property is `true`, for other builds it is
--- a/docs/api/process.md
+++ b/docs/api/process.md
@@ -61,7 +61,7 @@ property is `true` in the main process, otherwise it is `undefined`.
 
 ### `process.isMainFrame`
 
-A `Boolean`.  Set to `true` when the current renderer context is the "main" renderer
+A `Boolean`, `true` when the current renderer context is the "main" renderer
 frame.  If you want the ID of the current frame you should use `webFrame.routingId`.
 
 ### `process.mas`
--- a/docs/api/process.md
+++ b/docs/api/process.md
@@ -62,7 +62,7 @@ property is `true` in the main process, otherwise it is `undefined`.
 ### `process.isMainFrame`
 
 A `Boolean`, `true` when the current renderer context is the "main" renderer
-frame.  If you want the ID of the current frame you should use `webFrame.routingId`.
+frame. If you want the ID of the current frame you should use `webFrame.routingId`.
 
 ### `process.mas`
 
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -1427,8 +1427,8 @@ app.on('ready', () => {
 * `...args` any[]
 
 Send an asynchronous message to a specific frame in a renderer process via
-`channel`, you can also send arbitrary arguments. Arguments will be serialized
-in JSON internally and hence no functions or prototype chain will be included.
+`channel`. Arguments will be serialized
+as JSON internally and as such no functions or prototype chains will be included.
 
 The renderer process can handle the message by listening to `channel` with the
 [`ipcRenderer`](ipc-renderer.md) module.
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -1420,6 +1420,36 @@ app.on('ready', () => {
 </html>
 ```
 
+#### `contents.sendToFrame(frameId, channel[, arg1][, arg2][, ...])`
+
+* `frameId` Integer
+* `channel` String
+* `...args` any[]
+
+Send an asynchronous message to a specific frame in a renderer process via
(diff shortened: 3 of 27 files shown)


docs/api/process.md
```diff
@@ -59,6 +59,11 @@ process.once('loaded', () => {
 A `Boolean`. When app is started by being passed as parameter to the default app, this
 property is `true` in the main process, otherwise it is `undefined`.
 
+### `process.isMainFrame`
+
+A `Boolean`.  Set to `true` when the current renderer context is the "main" renderer
```
nornagon:
```suggestion
A `Boolean`, `true` when the current renderer context is the "main" renderer
```

docs/api/web-contents.md
```diff
@@ -1398,6 +1398,36 @@ app.on('ready', () => {
...
 ```
 
+#### `contents.sendToFrame(frameId, channel[, arg1][, arg2][, ...])`
+
+* `frameId` Integer
+* `channel` String
+* `...args` any[]
+
+Send an asynchronous message to a specific frame in a renderer process via
+`channel`, you can also send arbitrary arguments. Arguments will be serialized
```
nornagon:
```suggestion
`channel`. Arguments will be serialized
```

docs/api/web-contents.md
```diff
@@ -1398,6 +1398,36 @@ app.on('ready', () => {
...
+in JSON internally and hence no functions or prototype chain will be included.
```
nornagon:
```suggestion
as JSON internally and as such no functions or prototype chains will be included.
```

lib/browser/api/web-contents.js
```diff
@@ -325,6 +337,22 @@ WebContents.prototype.getZoomFactor = function (callback) {
   })
 }
 
+const wrapEventWithReply = (event, internal) => {
```
nornagon: split this into two functions, `wrapEventWithReply` and `wrapEventWithReplyInternal`.

... also, this doesn't actually "wrap" the event, so maybe `addReplyToEvent` might make more sense?

lib/browser/chrome-extension.js
```diff
@@ -196,7 +196,7 @@ ipcMain.on('CHROME_TABS_SEND_MESSAGE', function (event, tabId, extensionId, isBa
 
   contents._sendInternalToAll(`CHROME_RUNTIME_ONMESSAGE_${extensionId}`, senderTabId, message, resultID)
   ipcMain.once(`CHROME_RUNTIME_ONMESSAGE_RESULT_${resultID}`, (event, result) => {
-    event.sender._sendInternal(`CHROME_TABS_SEND_MESSAGE_RESULT_${originResultID}`, result)
+    event._replyInternal(`CHROME_TABS_SEND_MESSAGE_RESULT_${originResultID}`, result)
   })
   resultID++
 })
```
nornagon: does `CHROME_TABS_EXECUTESCRIPT` below also need to use `_reply`?

lib/browser/desktop-capturer.js
```diff
@@ -40,14 +40,14 @@ ipcMain.on(electronSources, (event, captureWindow, captureScreen, thumbnailSize,
...
-    request.webContents = null
+    request.event = null
   })
 })
 
 desktopCapturer.emit = (event, name, sources, fetchWindowIcons) => {
   // Receiving sources result from main process, now send them back to renderer.
   const handledRequest = requestsQueue.shift()
-  const handledWebContents = handledRequest.webContents
+  const handledWebContents = handledRequest.event ? handledRequest.event.sender : null
```
nornagon: since this is only actually used to switch on whether `event` is non-null, perhaps rather:

```js
const {event} = handledRequest
// ...
if (event) {
  event._replyInternal(...)
}
```

lib/browser/desktop-capturer.js
```diff
@@ -61,15 +61,15 @@ desktopCapturer.emit = (event, name, sources, fetchWindowIcons) => {
...
 
   if (handledWebContents) {
-    handledWebContents._sendInternal(capturerResult(handledRequest.id), result)
+    handledRequest.event._replyInternal(capturerResult(handledRequest.id), result)
   }
 
   // Check the queue to see whether there is another identical request & handle
   requestsQueue.forEach(request => {
-    const webContents = request.webContents
+    const event = request.event
```
nornagon: not really necessary to break this out as a variable

lib/browser/guest-view-manager.js
```diff
@@ -242,7 +242,8 @@ const attachGuest = function (event, embedderFrameId, elementInstanceId, guestIn
     ['nativeWindowOpen', true],
     ['nodeIntegration', false],
     ['enableRemoteModule', false],
-    ['sandbox', true]
+    ['sandbox', true],
+    ['kNodeSupportInSubFrames', false]
```
nornagon: `s/kN/n`

lib/browser/guest-window-manager.js
```diff
@@ -16,7 +16,8 @@ const inheritedWebPreferences = new Map([
   ['nodeIntegration', false],
   ['enableRemoteModule', false],
   ['sandbox', true],
-  ['webviewTag', false]
+  ['webviewTag', false],
+  ['kNodeSupportInSubFrames', false]
```
nornagon: `s/kN/n`

lib/renderer/init.js
```diff
@@ -70,11 +70,13 @@ if (window.location.protocol === 'chrome-devtools:') {
...
-  // Inject content scripts.
-  require('@electron/internal/renderer/content-scripts-injector')
+  // Inject content scripts in the main frame only
+  if (process.isMainFrame) {
+    require('@electron/internal/renderer/content-scripts-injector')
+  }
 
   // Load webview tag implementation.
-  if (webviewTag && guestInstanceId == null) {
+  if (webviewTag && guestInstanceId == null && process.isMainFrame) {
```
nornagon: it's intentional that `<webview>`s aren't allowed in subframes?

lib/renderer/window-setup.js
```diff
@@ -26,7 +26,7 @@
 const { defineProperty, defineProperties } = Object
 
 // Helper function to resolve relative url.
-const a = window.top.document.createElement('a')
+const a = window.document.createElement('a')
 const resolveURL = function (url) {
   a.href = url
   return a.href
```
nornagon: this function can be replaced with the [URL API](https://developer.mozilla.org/en-US/docs/Web/API/URL)

```js
const resolveURL = function(url) {
  return new URL(url, location.href).href
}
```

spec/api-subframe-spec.js
```diff
@@ -0,0 +1,88 @@
...
+
+const { emittedNTimes, emittedOnce } = require('./events-helpers')
+const { closeWindow } = require('./window-helpers')
+
+const { BrowserWindow } = remote
+
+describe('renderer nodeSupportInSubFrames', () => {
+  const generateTests = (sandboxEnabled) => {
+    describe(`with sandbox ${sandboxEnabled ? 'enabled' : 'disabled'}`, () => {
+      let w
```
nornagon: instead of this shared global, perhaps we could do something like

```js
const withWindow = async (f) => {
  const w = new BrowserWindow(...)
  await f(w)
  await closeWindow(w)
}

// later

it('something', async () => {
  await withWindow(async (w) => {
    // ...
  })
})
```

Or perhaps there's a nice way to do something like 

```js
it('something', withWindow(async () => {
  // ...
}))
```

mainly I don't like that the window is closed in the test _after_. That could result in test failures relating to teardown being misattributed.

spec/api-subframe-spec.js
```diff
@@ -0,0 +1,88 @@
...
+        const [event1, event2] = await detailsPromise
+        expect(event1[0].frameId).to.not.equal(event2[0].frameId)
+        expect(event1[0].frameId).to.equal(event1[2])
+        expect(event2[0].frameId).to.equal(event2[2])
+      })
+
+      it('should load preload scripts in nested iframes', async () => {
+        const detailsPromise = emittedNTimes(remote.ipcMain, 'preload-ran', 3)
+        w.loadFile(path.resolve(__dirname, 'fixtures/sub-frames/frame-with-frame-container.html'))
+        const [event1, event2, event3] = await detailsPromise
```
nornagon: Would it make sense to refactor these tests to fail rather than timeout if the preload doesn't get loaded?

lib/renderer/init.js
MarshallOfSound: Yes

lib/renderer/window-setup.js
MarshallOfSound: I'll follow up with this and other clean up referenced above.  I want to keep this down to just the new code

spec/api-subframe-spec.js
MarshallOfSound: Yes, but that's hard to do without complicated IPC.  A timeout in this case seemed the simplest (unfortunately) unless you have a good idea here

docs/api/browser-window.md
```diff
@@ -371,6 +371,14 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       English and not localized.
     * `navigateOnDragDrop` Boolean (optional) - Whether dragging and dropping a
       file or link onto the page causes a navigation. Default is `false`.
+    * `nodeSupportInSubFrames` Boolean (optional) - Experimental option for
+      enabling NodeJS support in sub-frames such as iframes.  This may have
+      an increased chance of memory leaks.  All your preloads will load for
```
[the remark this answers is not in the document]
MarshallOfSound: You can have multiple preloads, and we don't know when they occur, this was mostly a disclaimer "we may be leaking stuff that we aren't aware of yet".  Happy to remove 👍

spec/api-subframe-spec.js
MarshallOfSound: This is our current pattern, I like the idea of `withWindow` but I'd rather do that as a sweeping change rather than file-by-file for consitency 👍

lib/browser/chrome-extension.js
MarshallOfSound: No, `CHROME_TABS_EXECUTESCRIPT` always targets the main frame (even in chrome)

lib/browser/desktop-capturer.js
MarshallOfSound: Purely minimizing diff

lib/browser/api/web-contents.js
```diff
@@ -330,6 +342,22 @@ WebContents.prototype.loadFile = function (filePath, options = {}) {
   }))
 }
 
+const wrapEventWithReply = (event) => {
```
nornagon: these still don't "wrap" a thing?

MarshallOfSound: Ah right, will rename

miniak:
> Is there a chance this will make it into v4.1.x?
@CVertex highly unlikely

vladimiry:
Could the `even.reply` call potentially end up with `Object has been destroyed: sender`-like error? If so will the `event.sender.isDestroyed()` be a proper check?

vladimiry:
I answer the question myself.

Here is how the `reply` is defined:
https://github.com/electron/electron/blob/e78b902a7dfcf90d10f6b97b17c2a227b7ad50a1/lib/browser/api/web-contents.js#L287-L289

So yes `event.sender.isDestroyed()` check is needed since `reply` internally accesses the `event.sender`.

Here is how the error looks if there is no `event.sender.isDestroyed()` check:
```
Error: Object has been destroyed
    at WebContents.sendToFrame (/opt/ElectronMail/resources/electron.asar/browser/api/web-contents.js:138:17)
    at Object.event.reply (/opt/ElectronMail/resources/electron.asar/browser/api/web-contents.js:301:22)
```

PR #16441, opened 2019-01-18
build: use typescript for internal Electron JS code

MarshallOfSound:
This makes it possible for us to write Typescript for Electron's internals, it hooks up `tsc` inside GN before we generate `electron.asar` and `default_app.asar` and wires up generating Electron's type definitions into GN as well.

I haven't converted everything, just a few files from each ASAR to prove it works 👍 

Notes: no-notes

cc @BinaryMuse @codebytere @ckerr 

Note for reviewers: This is now based on top of #16718


build/tsc.gni
```diff
@@ -0,0 +1,40 @@
...
+                             "public_deps",
+                             "outputs",
+                           ])
+    script = "tsc"
+
+    sources = invoker.sources
+
+    args = [
+      "-p",
+      "tsconfig." + invoker.tsconfig + ".json",
```
nornagon: - the tsconfig needs to be listed in the inputs
- don't construct the name of the file by string concatenation, just specify the name of the file in the variable. i.e. `tsconfig = "tsconfig.electron.json"`. resolve it here with `rebase_path`.

build/tsc.gni
```diff
@@ -0,0 +1,40 @@
...
+      "--outDir",
+      rebase_path("$target_gen_dir/tsc/electron/" + invoker.electron_dir),
```
nornagon: `electron_dir` is vague. This looks like where the outputs go, so perhaps call it `output_dir` and have the invoker specify `"$target_gen_dir/tsc/electron/whatever"`? No need to enforce a particular directory structure on all users of this template.

build/tsc.gni
```diff
@@ -0,0 +1,40 @@
...
+      "--typeRoots",
+      "node_modules/@types,typings," +
+          rebase_path("$target_gen_dir/tsc/electron/typings"),
```
nornagon: what generates `tsc/electron/typings`?

spec/fixtures/no-proprietary-codecs.js
```diff
@@ -10,22 +10,22 @@ const path = require('path')
...
   })
 
-  window.webContents.on('crashed', (event, killed) => {
+  mainWindow.webContents.on('crashed', (event, killed) => {
     console.log(`WebContents crashed (killed=${killed})`)
     app.exit(1)
   })
 
-  window.loadFile(path.resolve(__dirname, 'asar', 'video.asar', 'index.html'))
+  mainWindow.loadFile(path.resolve(__dirname, 'asar', 'video.asar', 'index.html'))
```
nornagon: unrelated change..?

spec/static/main.js
```diff
@@ -277,7 +277,7 @@ ipcMain.on('set-client-certificate-option', function (event, skip) {
       ipcMain.on('client-certificate-response', function (event, certificate) {
         callback(certificate)
       })
-      window.webContents.send('select-client-certificate', webContents.id, list)
+      mainWindow.webContents.send('select-client-certificate', webContents.id, list)
```
nornagon: here too

tools/cp-defs.js
```diff
@@ -0,0 +1,9 @@
+const fs = require('fs')
+const path = require('path')
+
+const src = path.resolve(__dirname, '../electron.d.ts')
+const target = process.argv[2]
+
+if (!target) return
+
+fs.writeFileSync(target, fs.readFileSync(src))
```
nornagon: it seems like you could just do this with GN's `copy` action, or just pass an extra `--out` to `npm run create-typescript-definitions`

tools/js2asar.py
```diff
@@ -27,7 +27,11 @@ def copy_files(source_files, output_dir, folder_name):
     # Files that aren't in the default_app folder need to be put inside
     # the temp one we are making so they end up in the ASAR
     if not os.path.normpath(source_file).startswith(folder_name + os.sep):
-      output_path = os.path.join(output_dir, folder_name, source_file)
+      inner = os.sep + folder_name + os.sep
+      if inner in source_file:
```
nornagon: yikes, what is this even doing and why?

tsconfig.default-app.json
```diff
@@ -0,0 +1,13 @@
...
+		"rootDir": "default_app",
+	},
+	"exclude": [
+		"script",
+		"spec",
+		"tools",
+		"lib",
+		"npm"
+	]
+}
```
nornagon: nit: 2 spaces pls (here and in the other new json files)

default_app/main.ts
```diff
@@ -1,20 +1,31 @@
...
+
+type DefaultAppOptions = {
+  file: null | string;
+  noHelp: boolean;
+  version: boolean;
+  webdriver: boolean;
+  interactive: boolean;
+  abi: boolean;
+  modules: string[];
+}
```
sindresorhus: All these needs to be marked as optional otherwise they'll all required even if you just want to specify one option.

```typescript
type DefaultAppOptions = {
  file?: string;
  noHelp?: boolean;
  version?: boolean;
  webdriver?: boolean;
  interactive?: boolean;
  abi?: boolean;
  modules?: string[];
}
```

tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+	"compilerOptions": {
+		"module": "commonjs",
+		"target": "es2017",
+		"lib": [
+			"es6",
+			"es7",
```
sindresorhus: For consistency with `target`, these should be:

```json
"es2017"
```

You don't need to specify older versions, as `es2017` imports those.

typings/internal-electron.d.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * This file augments the Electron TS namespace with the internal APIs
+ * that are not documented but are used by Electron internally
+ */
+
+declare namespace Electron {
+  interface App {
+    setVersion(version: string): void;
+    setDesktopName(name: string): void;
+    setAppPath(path: string | null): void;
```
sindresorhus:
```suggestion
    setAppPath(path?: string): void;
```

MarshallOfSound: This method does explicitly need `null`, see the implementation of the file path converter --> https://github.com/electron/electron/blob/master/atom/common/native_mate_converters/file_path_converter.h#L21-L34

typings/internal-ambient.d.ts
```diff
@@ -0,0 +1,7 @@
+declare namespace NodeJS {
+  interface Process {
+    // TODO: Strictly type this based on the input maybe?
+    atomBinding: (name: string) => any;
```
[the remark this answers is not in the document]
MarshallOfSound: If we make this `unknown` the entire world will blow up.  I'm planning on making it

`atomBinding: <T>(name: string) => T;`

So we can at least get strong typing, this method is basically a bridge between native land and JS so it's gonna be a pain to strictly type it by hand.

typings/internal-electron.d.ts
sindresorhus: Using `null` to mean no argument is an anti-pattern today as new features like default parameters only work with `undefined`. It's also surprising that it *only* works with `null` and not also `undefined`. I guess I should open a new issue about changing these to only accept `undefined`? Or at minimum `undefined` too.

https://github.com/electron/electron/search?p=2&q=isnull&unscoped_q=isnull

MarshallOfSound: I agree setting null isn't exactly the JS way of doing things, but I wouldn't want it to work with `undefined` that would look super weird.  Having `setMenu()` remove the menu isn't as explicit as `setMenu(null)`.

cc @codebytere we discussed this at some point, having `removeX` instead of `setX(null)`

sindresorhus: Oh, sorry, I missed this was for removing the menu. I would definitely prefer an explicit method for this, or even better, use getters/setters whenever possible.

codebytere: @sindresorhus i agree, we're looking towards updating these types of things to make them more in line with modern practices

typings/internal-ambient.d.ts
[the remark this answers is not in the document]
samuelmaddock: You could type this as
```ts
interface Process {
  atomBinding<T = any>(name: string): T
}
```
That way you could optionally provide a type like `process.atomBinding<number>('example')` until the codebase developed further.

tools/js2asar.py
MarshallOfSound: Dealt with this entire fire of a file in #16718

.eslintrc.json
```diff
@@ -1,12 +1,21 @@
...
+    "typescript/no-unused-vars": "error",
     "prefer-const": ["error", {
       "destructuring": "all"
     }]
-  }
+  },
+  "parserOptions": {
+    "ecmaVersion": 6,
+    "sourceType": "module"
+}
```
nornagon:
```suggestion
  }
```

BUILD.gn
```diff
@@ -53,6 +55,16 @@ config("branding") {
   ]
 }
 
+npm_action("build_electron_definitions") {
```
nornagon: Please add a comment here about why `electron.d.ts` is needed in two places

build/tsc.gni
```diff
@@ -0,0 +1,47 @@
+import("npm.gni")
+
+template("typescript_build") {
+  assert(defined(invoker.tsconfig),
+         "Need tsconfig name to run (must exist in electron root)")
```
nornagon: what does `must exist in electron root` mean here? it looks like the path gets rebased so it'll be the absolute path, why does it matter where the tsconfig is?

build/tsc.gni
nornagon: This implies a dependency on `tsc/electron/typings`. Either state that dependency here, or expose the `typeRoots` functionality in such a way that this template can function without that implicit dependency.

script/gen-filenames.js
```diff
@@ -0,0 +1,17 @@
...
+const gniPath = path.resolve(__dirname, '../filenames.auto.gni')
+
+const allDocs = fs.readdirSync(path.resolve(__dirname, '../docs/api'))
+
+fs.writeFileSync(
+  gniPath,
+  `# THIS FILE IS AUTO-GENERATED, PLEASE DO NOT EDIT BY HAND
+auto_filenames = {
+  api_docs = [
+${allDocs.map(doc => `    "docs/api/${doc}",`).join('\n')}
```
nornagon: this should include `docs/api/structures`, right?

tsconfig.electron.json
```diff
@@ -0,0 +1,14 @@
...
+	},
+	"exclude": [
+		"script",
+		"spec",
+		"tools",
+		"default_app",
+		"npm",
+		"electron.d.ts"
+	]
+}
```
nornagon:
```suggestion
}
```

tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+		"allowJs": true,
+		"outDir": "ts-gen",
+		"paths": {
+			"@electron-internal/*": ["lib"]
+		}
+	},
+	"exclude": [
+		"electron.d.ts"
+	]
+}
```
nornagon:
```suggestion
}
```

nornagon: also 2 spaces, on this and the other tsconfigs

build/tsc.gni
```diff
@@ -0,0 +1,50 @@
...
+                           [
+                             "deps",
+                             "public_deps",
+                             "outputs",
+                           ])
+    script = "tsc"
+    if (!defined(deps)) {
+      deps = []
+    }
+    deps += [ ":build_electron_definitions" ]
```
nornagon: eek this isn't great. this means that this `template` will _only work in `//electron/BUILD.gn`_.

If that's your intention, just move this template into `//electron/BUILD.gn`.

lib/browser/init.ts
```diff
@@ -149,9 +150,6 @@ app.setPath('userData', path.join(app.getPath('appData'), app.getName()))
 app.setPath('userCache', path.join(app.getPath('cache'), app.getName()))
 app.setAppPath(packagePath)
 
-// Load the chrome devtools support.
```
miniak: @MarshallOfSound seems like this was removed by mistake.

MarshallOfSound: Looks like a dodgy git rebase, I'll add it back

dhewya:
https://github.com/RangeNetworks/dev/commit/68a2915ae45a03a34ff7dc72c04c67ce5b306af1#r32335525

Issue #15459, opened 2018-10-29
Memory leak with use sandbox true and webFrame.setSpellCheckProvider


vrbsm:
* Output of `node_modules/.bin/electron --version`: v3.0.5
* Operating System (Platform and Version): macOS 10.14

**Expected Behavior**
When webPreferences.sandbox is set to true and webFrame is using a provider for spell checking, the memory must be released after page reload.

**Actual behavior**
The memory isn't released and caused memory leak

**To Reproduce**

- Clone https://github.com/vrbsm/example-memory-leak-electron
- Run npm run start
Memory usage of electron process will be increasing in task manager after reloading the page several times

**Screenshots**
![test-memory-leak-01](https://user-images.githubusercontent.com/9887288/47675838-c2d4b800-db99-11e8-97e8-b5b27bcab929.gif)

richardlopes-daitangroup:
Same here

MarshallOfSound:
@ckerr Master has the new async spellchecker API, that impl might have fixed some things by accident? 😆

vrbsm:
> @ckerr Master has the new async spellchecker API, that impl might have fixed some things by accident? 😆

I tested the master, but still saw the problem of memory leak :-(
Do you have any updates if this could be released soon?

vrbsm:
@nitsakh I tested **v4.0.3** and the memory leak still persists.
basically: 
clone: **https://github.com/vrbsm/example-memory-leak-electron**
run: npm run start
Memory usage will increase in task manager after reloading the page several times

PR #16448, opened 2019-01-18
fix: correctly destroy spellcheck client

nitsakh:
#### Description of Change
This change addresses https://github.com/electron/electron/issues/15459.
The spellcheck client wasn't destroyed when the renderer goes away.
This change adds an observer which takes care of the destruction.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed `webFrame.setSpellCheckProvider` memory leak

### DIFF
--- a/atom/renderer/api/atom_api_web_frame.cc
+++ b/atom/renderer/api/atom_api_web_frame.cc
@@ -149,7 +149,11 @@ class AtomWebFrameObserver : public content::RenderFrameObserver {
   ~AtomWebFrameObserver() final {}
 
   // RenderFrameObserver implementation.
-  void OnDestruct() final { spell_check_client_.reset(nullptr); }
+  void OnDestruct() final {
+    spell_check_client_.reset();
+    // Frame observers should delete themselves
+    delete this;
+  }
 
  private:
   std::unique_ptr<SpellCheckClient> spell_check_client_;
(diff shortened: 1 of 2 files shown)


nitsakh:
We should backport this to 3-0-x also, right?

atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -231,11 +249,13 @@ void WebFrame::SetSpellCheckProvider(mate::Arguments* args,
...
   // Set spellchecker for all live frames in the same process or
   // in the sandbox mode for all live sub frames to this WebFrame.
-  FrameSpellChecker spell_checker(
-      client.get(), content::RenderFrame::FromWebFrame(web_frame_));
+  auto* render_frame = content::RenderFrame::FromWebFrame(web_frame_);
+  FrameSpellChecker spell_checker(client.get(), render_frame);
   content::RenderFrame::ForEach(&spell_checker);
   spell_check_client_.swap(client);
   web_frame_->SetSpellCheckPanelHostClient(spell_check_client_.get());
+  web_frame_observer_.reset(
```
deepak1556: This can be just `new AtomWebFrameObserver(render_frame, std::move(spell_check_client_)` instead of storing it in a member variable, frame observers are meant to be self destructing classes since the observed frame can go away at any time signaled by `OnDestruct`. There is no point in holding onto the observer after that.

atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -137,6 +139,22 @@ class FrameSpellChecker : public content::RenderFrameVisitor {
...
+ public:
+  explicit AtomWebFrameObserver(
+      content::RenderFrame* render_frame,
+      std::unique_ptr<SpellCheckClient> spell_check_client)
+      : content::RenderFrameObserver(render_frame),
+        spell_check_client_(std::move(spell_check_client)) {}
+  ~AtomWebFrameObserver() final {}
+
+  // RenderFrameObserver implementation.
+  void OnDestruct() final { spell_check_client_.reset(nullptr); }
```
deepak1556: Should add `delete this;` corresponding to the changes described below. Also `spell_check_client_.reset()` should be enough.

Issue #16195, opened 2018-12-26
Electron 4.0.0  'unknown cipher' appears  in Crypto module


mazhorse:
* Output of `node_modules/.bin/electron --version`:  4.0.0
* Operating System (Platform and Version):  Windows 10
* Output of `node_modules/.bin/electron --version` on last known working Electron version (if applicable):  v3.x.x

**Expected Behavior**
Generally , invoke ``crypto.createCipheriv ()`` that should return a Cipheriv Object.

**Actual behavior**
When I call  ``crypto.createCipheriv()`` , it will  response me a error that is **unknown cipher** .
But   If I run this program(**test code** ) alone by node cmd. (ex. node test.js ) that it work.
So I don't think it is a node10 problem.

**To Reproduce**
**sample code** 
in main.js
```
const crypto = require('crypto')
const methodName = 'aes-256-cfb'
const key =  'nMvTdb7VXMPudFWH'
const iv = crypto.randomBytes(16)
const cipher = crypto.createCipheriv(methodName, key, iv)
console.log(cipher)
```
**Response:**
in Electron 4  or  more newer
```
unknown cipher
```

in Electron 3.x , more older or  run alone by node cmd
```
 Cipheriv {
  _handle: {},
  _decoder: null,
  _options: undefined,
  writable: true,
  readable: true }  
```

shiftkey:
@dessertboy which version of Node 10 are you testing on locally that works with your sample code? I ask because Electron 3 is based on Node 10.2 and Electron 4 is based on Node 10.11, so having more data points here about what works and what doesn't will help us troubleshoot the issue better...

dessertboy:
> @dessertboy which version of Node 10 are you testing on locally that works with your sample code? I ask because Electron 3 is based on Node 10.2 and Electron 4 is based on Node 10.11, so having more data points here about what works and what doesn't will help us troubleshoot the issue better...

@shiftkey  I have tried v10.2.0 and v10.11.0 on locally with sample code in test.js ,  and execute the sample code  via node cmd.  As a result both versions can succeed
 If  I  run sample code in main.js via **Electron 4** that will fail.  but **Electron 3** is  ok.
 so  I don't think that is  Node10  version problem.

shiftkey:
For reference, I checked the similar `crypto.getCiphers()` function on macOS (not Windows like the original reporter) in Electron `3.0.13` (135 listed):

[...]

PR #16454, opened 2019-01-18
fix: expose ripemd160 hash from boringssl

nornagon:
Ref #16195

ripemd160 is available in openssl by default, but boringssl does not build it by default. This adds a patch to expose it from boringssl.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Restored support for RIPEMD160 digest, which was lost when switching from OpenSSL to BoringSSL.

### DIFF
--- a/patches/common/boringssl/.patches
+++ b/patches/common/boringssl/.patches
@@ -1,2 +1,3 @@
 add_ec_group_order_bits_for_openssl_compatibility.patch
 add_ec_key_key2buf_for_openssl_compatibility.patch
+expose_ripemd160.patch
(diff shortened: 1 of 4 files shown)


nornagon:
Yep, I agree that ripemd160 is not a wise choice of digest, but since it's right there I figured this was easier than arguing about it. If BoringSSL removes it upstream I'm happy to drop support for it in Electron.

ryzokuken:
@nornagon sounds fair.

nornagon:
CI failures are unrelated, so I'm going to go ahead and merge.

sirasistant:
This is why some cryptocurrency-related libraries require this hash: 
https://en.bitcoin.it/wiki/Technical_background_of_version_1_Bitcoin_addresses

ryzokuken:
@sirasistant I suppose to trim the hash (which is otherwise 32 bytes because SHA256)?

peterremote1980:
So, how to fix this problem? thanks

PR #16467, opened 2019-01-20
perf: and refactor: Code Scope + Minor Performance Improvement.

shahzadlone:
Made two vectors scope narrower and reserved them because we know the size in advance. This helps save on allocation costs. 

"It's better to delay initialization until it is actually used".

notes: no-notes

### DIFF
--- a/atom/browser/api/atom_api_app.cc
+++ b/atom/browser/api/atom_api_app.cc
@@ -1154,8 +1154,9 @@ v8::Local<v8::Promise> App::GetFileIcon(const base::FilePath& path,
 
 std::vector<mate::Dictionary> App::GetAppMetrics(v8::Isolate* isolate) {
   std::vector<mate::Dictionary> result;
+  result.reserve(app_metrics_.size());
   int processor_count = base::SysInfo::NumberOfProcessors();
-
+  
   for (const auto& process_metric : app_metrics_) {
     mate::Dictionary pid_dict = mate::Dictionary::CreateEmpty(isolate);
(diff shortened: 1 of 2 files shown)


atom/browser/api/atom_api_app.cc
```diff
@@ -1154,6 +1154,7 @@ v8::Local<v8::Promise> App::GetFileIcon(const base::FilePath& path,
 
 std::vector<mate::Dictionary> App::GetAppMetrics(v8::Isolate* isolate) {
   std::vector<mate::Dictionary> result;
+  result.reserve(app_metrics_.size());
```
nornagon: I think this can be accomplished with the `vector( size_type count );` constructor, i.e.

```cpp
std::vector<mate::Dictionary> result(app_metrics_.size())
```

(in this and the other two locations)

shahzadlone: That is bad, because then you aren't just reserving enough space but also default-initializing those many values, the when instead reserve effects only the cap (not the size) so enough memory is reserved, but does not initialize them.

nornagon: Ah, yes, you're right.

shahzadlone:
@nornagon Thanks for the approve. What are the next steps to finish for the merge?

codebytere:
@shahzadlone we're just waiting to ensure that CI either passes or fails only on known flakes, and then we can get this merged

jkleinsc:
Merging as CI failures are known flakes

PR #16494, opened 2019-01-22
feat: upgrade to Chromium 73.0.3683.27

deepak1556:
#### Description of Change

Refs https://github.com/electron/electron/projects/17

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: upgrade to Chromium 73.0.3683.27


deepak1556:
Will rebase once beta branch is cut in upstream, its real close. Lets merge once we have updated to beta release.

deepak1556:
@MarshallOfSound the `nightly-mac-release-test` chromedriver workflow failed with

```
/bin/sh: /Users/.../external_binaries/sccache: cannot execute binary file
```

looks like its missing `step-fix-sync-on-mac` step ?

deepak1556:
Ignore lint errors, they should be fixed by https://github.com/electron/electron/pull/16906

atom/browser/media/media_capture_devices_dispatcher.cc
```diff
@@ -49,75 +42,75 @@ MediaCaptureDevicesDispatcher::MediaCaptureDevicesDispatcher()
 
 MediaCaptureDevicesDispatcher::~MediaCaptureDevicesDispatcher() {}
 
-const MediaStreamDevices&
+const blink::MediaStreamDevices&
 MediaCaptureDevicesDispatcher::GetAudioCaptureDevices() {
   DCHECK(BrowserThread::CurrentlyOn(BrowserThread::UI));
   if (is_device_enumeration_disabled_)
-    return EmptyDevices();
+    return test_audio_devices_;
```
nornagon: What's the intent behind returning an internal list instead of making a new empty one? This seems like a potential issue if this somehow became non-empty

atom/browser/media/media_stream_devices_controller.cc
```diff
@@ -133,40 +133,45 @@ void MediaStreamDevicesController::Accept() {
...
       }
-      case content::MEDIA_DEVICE_ACCESS:
+      case blink::MEDIA_DEVICE_ACCESS: {
         // Get the default devices for the request.
         MediaCaptureDevicesDispatcher::GetInstance()->GetDefaultDevices(
             microphone_requested_, webcam_requested_, &devices);
         break;
+      }
+      case blink::MEDIA_DEVICE_UPDATE: {
+        NOTREACHED();
```
nornagon: should we do something else with this event?

atom/common/application_info.cc
```diff
@@ -4,7 +4,29 @@
...
 #include "atom/common/application_info.h"
 
+#include "atom/browser/browser.h"
+#include "atom/common/atom_version.h"
 #include "base/no_destructor.h"
+#include "base/strings/stringprintf.h"
+#include "chrome/common/chrome_version.h"
+#include "content/public/common/user_agent.h"
+
+std::string GetUserAgent() {
```
nornagon: this should perhaps go in the `atom` namespace?

atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -530,12 +522,7 @@ void Initialize(v8::Local<v8::Object> exports,
   dict.SetMethod("executeJavaScript", &ExecuteJavaScript);
   dict.SetMethod("executeJavaScriptInIsolatedWorld",
                  &ExecuteJavaScriptInIsolatedWorld);
-  dict.SetMethod("setIsolatedWorldSecurityOrigin",
-                 &SetIsolatedWorldSecurityOrigin);
-  dict.SetMethod("setIsolatedWorldContentSecurityPolicy",
-                 &SetIsolatedWorldContentSecurityPolicy);
-  dict.SetMethod("setIsolatedWorldHumanReadableName",
-                 &SetIsolatedWorldHumanReadableName);
+  dict.SetMethod("setIsolatedWorldInfo", &SetIsolatedWorldInfo);
```
nornagon: Did we update the docs for this? We should probably continue to support the old API as deprecated for 5-0-x

patches/common/chromium/can_create_window.patch
```diff
@@ -5,114 +5,29 @@ Subject: can_create_window.patch
...
-+          NOTREACHED();
-+          break;
-+      }
-+      elements.push_back(std::move(element));
-+    }
-+    body->swap_elements(&elements);
-+    body->set_identifier(params->body->identifier);
-+    body->set_contains_sensitive_info(params->body->contains_sensitive_info);
-+  }
-+
```
nornagon: can we really get rid of all this stuff? that's great!

atom/browser/ui/message_box_mac.mm
```diff
@@ -157,10 +157,13 @@ int ShowMessageBox(NativeWindow* parent_window,
           callEndModal:true];
 
   NSWindow* window = parent_window->GetNativeWindow().GetNativeNSWindow();
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
   [alert beginSheetModalForWindow:window
                     modalDelegate:delegate
                    didEndSelector:@selector(alertDidEnd:returnCode:contextInfo:)
                       contextInfo:nil];
```
nornagon: can we instead switch to using [beginSheetModalForWindow:completionHandler:](https://developer.apple.com/documentation/appkit/nsalert/1524296-beginsheetmodalforwindow?language=objc)? here and below

build/args/all.gn
```diff
@@ -13,3 +13,7 @@ ffmpeg_branding = "Chrome"
 enable_basic_printing = true
 
 is_cfi = false
+
+# FIXME(deepak1556): workaround for https://crbug.com/924225
+# remove this when clang roll 352138 lands.
+enable_precompiled_headers = false
```
nornagon: can we remove this now?

patches/common/v8/disallow_implicit_constructors_for_macroassembler.patch
```diff
@@ -0,0 +1,102 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: deepak1556 <hop2deep@gmail.com>
+Date: Tue, 5 Feb 2019 13:32:33 +0530
+Subject: DISALLOW_IMPLICIT_CONSTRUCTORS for MacroAssembler
```
nornagon: please note that this is a backport

nornagon: oh you did :)

nornagon:
Looking good but a few questions before I give it the ✅

atom/renderer/api/atom_api_web_frame.cc
MarshallOfSound: @nornagon There is a separate PR to deprecate the current API in `5-0-x`

MarshallOfSound: https://github.com/electron/electron/pull/16801

patches/common/chromium/can_create_window.patch
deepak1556: `network::ResourceRequestBody` got typemapped in https://chromium-review.googlesource.com/c/chromium/src/+/1395977 , which helped us get rid off these manual conversions. Side note: mojom typemaps are really cool feature :)

build/args/all.gn
deepak1556: The change got landed in 74 https://chromium-review.googlesource.com/c/chromium/src/+/1436036

atom/browser/ui/message_box_mac.mm
deepak1556: Yup that seems like the alternative, can we change this as a follow up ?

atom/browser/media/media_stream_devices_controller.cc
deepak1556: It was added in https://chromium-review.googlesource.com/c/chromium/src/+/1354442. The feature seems specifically designed for chrome's screen sharing UI, can consider it for improvement in our desktop capture model once the issue goes public and has a bit more context.

atom/browser/media/media_capture_devices_dispatcher.cc
deepak1556: This is already behind a noop code path until `DisableDeviceEnumerationForTesting` is called. I am inclined to remove this condition all together, the goal of rewriting this was to see if the upstream `MediaCaptureDevicesDispatcher` class can be reused with less patch.

atom/common/application_info.cc
deepak1556: Fixed.

nornagon:
I can't speak for the OSR changes but everything else LGTM.

deepak1556:
@brenca can you review the OSR changes, thanks!

PR #16334, opened 2019-01-09
feat: upgrade to Chromium 72.0.3626.52

nornagon:
#### Description of Change
Ref https://github.com/electron/electron/projects/16

#### Checklist

- [ ] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [ ] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [ ] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: upgrade to Chromium 72.0.3626.52

PR #16501, opened 2019-01-23
fix: use frame tree api to remove <webview> frame

deepak1556:
#### Description of Change

Replaces the usage of testing api added in https://github.com/electron/electron/pull/16334/commits/03d499bf34d664866f30b280f55b72290b538653

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/api/atom_api_web_contents_impl.cc
+++ b/atom/browser/api/atom_api_web_contents_impl.cc
@@ -4,7 +4,10 @@
 
 #include "atom/browser/api/atom_api_web_contents.h"
 
+#include "content/browser/frame_host/frame_tree.h"
+#include "content/browser/frame_host/frame_tree_node.h"
 #include "content/browser/web_contents/web_contents_impl.h"
+#include "content/public/browser/guest_mode.h"
 
 #if BUILDFLAG(ENABLE_OSR)
 #include "atom/browser/osr/osr_render_widget_host_view.h"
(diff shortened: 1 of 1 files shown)


atom/browser/api/atom_api_web_contents_impl.cc
```diff
@@ -21,8 +24,15 @@ namespace api {
 
 void WebContents::DetachFromOuterFrame() {
   // See detach_webview_frame.patch on how to detach.
-  auto* impl = static_cast<content::WebContentsImpl*>(web_contents());
-  impl->GetRenderManagerForTesting()->RemoveOuterDelegateFrame();
+  DCHECK(content::GuestMode::IsCrossProcessFrameGuest(web_contents()));
+  int frame_tree_node_id =
+      static_cast<content::WebContentsImpl*>(web_contents())
+          ->GetOuterDelegateFrameTreeNodeId();
+  if (frame_tree_node_id != content::FrameTreeNode::kFrameTreeNodeInvalidId) {
```
nornagon: in what situation would `frame_tree_node_id` be the invalid ID? should we `DCHECK` or `CHECK` it?

deepak1556: This specifically tries to obtain the node id of an outer webcontents, so it would return an invalid id when there is no outer webcontents associated for this content.

PR #16518, opened 2019-01-23
build: ensure index.json is actually valid JSON before uploading

MarshallOfSound:
I managed to recover the damage from yesterday but this will ensure that we never push bad JSON again 👍 

Notes: no-notes

### DIFF
--- a/script/upload-index-json.py
+++ b/script/upload-index-json.py
@@ -18,8 +18,8 @@ authToken = os.getenv('META_DUMPER_AUTH_HEADER')
 
 def is_json(myjson):
   try:
-    json_object = json.loads(myjson)
-  except ValueError, e:
+    json.loads(myjson)
+  except ValueError:
     return False
   return True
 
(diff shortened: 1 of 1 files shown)


script/upload-index-json.py
```diff
@@ -15,16 +16,28 @@
...
 
-    return urllib2.urlopen(
+    proposed_content = urllib2.urlopen(
       request
     ).read()
+
+    if is_json(proposed_content):
+      return proposed_content
+    print("bad attempt")
+    raise Exception("Failed to fetch valid JSON from the metadumper service")
```
ckerr: Would it be useful to append the json.loads error message to this for postmortems?

MarshallOfSound: @ckerr I'm uncertain as to what that error message may contain (E.g. Secret keys from the metadumper) The error would be logged on the metadumper anyway so we should be 🆗 here without logging

PR #16544, opened 2019-01-25
chore: fix promisify helper

codebytere:
#### Description of Change

Fixes an issue with our promisification helper whereby we were mimicking node-style callbacks and most of our callback-based APIs didn't conform to this format, and so it broke a lot of backwards compatibility that was, in turn, previously not caught owing to miss-passed `done`s.

cc @ckerr 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/spec/api-web-contents-spec.js
+++ b/spec/api-web-contents-spec.js
@@ -597,19 +597,6 @@ describe('webContents module', () => {
       }
     })
 
-    it('can set the correct zoom level (callback)', async () => {
-      try {
-        await w.loadURL('about:blank')
-        const zoomLevel = await new Promise(resolve => w.webContents.getZoomLevel(resolve))
-        expect(zoomLevel).to.eql(0.0)
-        w.webContents.setZoomLevel(0.5)
-        const newZoomLevel = await new Promise(resolve => w.webContents.getZoomLevel(resolve))
-        expect(newZoomLevel).to.eql(0.5)
-      } finally {
-        w.webContents.setZoomLevel(0)
-      }
-    })
-
     it('can persist zoom level across navigation', (done) => {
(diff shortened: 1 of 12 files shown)


spec/api-web-contents-spec.js
```diff
@@ -597,19 +597,6 @@ describe('webContents module', () => {
...
-        await w.loadURL('about:blank')
-        const zoomLevel = await new Promise(resolve => w.webContents.getZoomLevel(resolve))
-        expect(zoomLevel).to.eql(0.0)
-        w.webContents.setZoomLevel(0.5)
-        const newZoomLevel = await new Promise(resolve => w.webContents.getZoomLevel(resolve))
-        expect(newZoomLevel).to.eql(0.5)
-      } finally {
-        w.webContents.setZoomLevel(0)
-      }
-    })
```
ckerr: DYK what happens if one of these calls throws an exception and there's not a `catch` block here? I expect that mocha/chai would catch it itself and fail the test -- or the entire thing would just die, which would still show up in CI as a failed test -- but do we know it for sure?

docs/api/desktop-capturer.md
```diff
@@ -87,11 +87,10 @@ The `desktopCapturer` module has the following methods:
     value is false. When false the appIcon property of the sources return null. Same if a source has
     the type screen.
 * `callback` Function
-  * `error` Error
   * `sources` [DesktopCapturerSource[]](structures/desktop-capturer-source.md)
 
 Starts gathering information about all available desktop media sources,
-and calls `callback(error, sources)` when finished.
```
ckerr: Doesn't this count as a semver breaking change?

PR #16571, opened 2019-01-28
fix: reject with error when url not loaded

codebytere:
#### Description of Change

When a user tried to call `webContents.hasServiceWorker()`on a `webContents` that hadn't yet loaded a valid URL, a DCHECK would get triggered. This prevents unexplained behavior for the end user by clearly rejecting with an error when `webContents.hasServiceWorker()` is called and there is no valid file or url in which to check for the presence of a ServiceWorker.

cc @ckerr 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Ensured that `webContents.hasServiceWorker()` rejects with an error for invalid URLs.

### DIFF
--- a/atom/browser/api/atom_api_web_contents.cc
+++ b/atom/browser/api/atom_api_web_contents.cc
@@ -1444,12 +1444,17 @@ v8::Local<v8::Promise> WebContents::HasServiceWorker() {
   auto* context = GetServiceWorkerContext(web_contents());
   if (!context) {
     promise->RejectWithErrorMessage("Unable to get ServiceWorker context.");
+    return promise->GetHandle();
+  }
+
+  GURL url = web_contents()->GetLastCommittedURL();
+  if (!url.is_valid()) {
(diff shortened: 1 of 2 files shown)


atom/browser/api/atom_api_web_contents.cc
```diff
@@ -1444,12 +1444,17 @@ v8::Local<v8::Promise> WebContents::HasServiceWorker() {
   auto* context = GetServiceWorkerContext(web_contents());
   if (!context) {
     promise->RejectWithErrorMessage("Unable to get ServiceWorker context.");
+    return promise->GetHandle();
+  }
+
+  GURL url = web_contents()->GetLastCommittedURL();
+  if (!url.is_valid()) {
```
deepak1556: When would a committed url be invalid ? Its a virtual url generated by the navigation controller. The DCHECK you hit is probably the other url that should be encompassed isn't recognized, in which case try using visible URL. Also using the result of committed URL to test page loading state is not a good idea.

PR #16600, opened 2019-01-29
feat: promisify contentTracing.getTraceBufferUsage()

codebytere:
#### Description of Change

Promisify `contentTracing.getTraceBufferUsage()`.

**NB**: this PR required backwards-incompatible changes and thus cannot be backported to `5-0-x`.

cc @ckerr @deepak1556

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Converted `contentTracing.getTraceBufferUsage()` to return a Promise instead of taking a callback.

### DIFF
--- a/atom/browser/api/atom_api_content_tracing.cc
+++ b/atom/browser/api/atom_api_content_tracing.cc
@@ -115,11 +115,10 @@ v8::Local<v8::Promise> StartTracing(
   return promise->GetHandle();
 }
 
-void OnTraceBufferUsageAvailable(v8::Isolate* isolate,
-                                 scoped_refptr<atom::util::Promise> promise,
+void OnTraceBufferUsageAvailable(scoped_refptr<atom::util::Promise> promise,
                                  float percent_full,
                                  size_t approximate_count) {
-  mate::Dictionary dict = mate::Dictionary::CreateEmpty(isolate);
+  mate::Dictionary dict = mate::Dictionary::CreateEmpty(promise->isolate());
   dict.Set("percentage", percent_full);
   dict.Set("value", approximate_count);
 
@@ -129,12 +128,10 @@ void OnTraceBufferUsageAvailable(v8::Isolate* isolate,
 v8::Local<v8::Promise> GetTraceBufferUsage(v8::Isolate* isolate) {
   scoped_refptr<atom::util::Promise> promise = new atom::util::Promise(isolate);
   bool success = TracingController::GetInstance()->GetTraceBufferUsage(
-      base::BindOnce(&OnTraceBufferUsageAvailable, isolate, promise));
+      base::BindOnce(&OnTraceBufferUsageAvailable, promise));
 
-  if (!success) {
+  if (!success)
     promise->RejectWithErrorMessage("Could not get trace buffer usage.");
-    return promise->GetHandle();
-  }
 
   return promise->GetHandle();
 }
--- a/atom/browser/api/atom_api_content_tracing.cc
+++ b/atom/browser/api/atom_api_content_tracing.cc
@@ -115,10 +115,28 @@ v8::Local<v8::Promise> StartTracing(
   return promise->GetHandle();
 }
 
-bool GetTraceBufferUsage(
-    const base::RepeatingCallback<void(float, size_t)>& callback) {
-  return TracingController::GetInstance()->GetTraceBufferUsage(
-      base::BindOnce(callback));
+void OnTraceBufferUsageAvailable(v8::Isolate* isolate,
+                                 scoped_refptr<atom::util::Promise> promise,
+                                 float percent_full,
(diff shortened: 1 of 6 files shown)


atom/browser/api/atom_api_content_tracing.cc
```diff
@@ -84,10 +85,28 @@ bool StartTracing(const base::trace_event::TraceConfig& trace_config,
...
 
-bool GetTraceBufferUsage(
-    const base::RepeatingCallback<void(float, size_t)>& callback) {
-  return TracingController::GetInstance()->GetTraceBufferUsage(
-      base::BindOnce(callback));
+void OnTraceBufferUsageAvailable(v8::Isolate* isolate,
+                                 scoped_refptr<atom::util::Promise> promise,
+                                 float percent_full,
+                                 size_t approximate_count) {
+  mate::Dictionary dict = mate::Dictionary::CreateEmpty(isolate);
```
miniak: you can use `promise->isolate()`

atom/browser/api/atom_api_content_tracing.cc
```diff
@@ -84,10 +85,28 @@ bool StartTracing(const base::trace_event::TraceConfig& trace_config,
...
+}
+
+v8::Local<v8::Promise> GetTraceBufferUsage(v8::Isolate* isolate) {
+  scoped_refptr<atom::util::Promise> promise = new atom::util::Promise(isolate);
+  bool success = TracingController::GetInstance()->GetTraceBufferUsage(
+      base::BindOnce(&OnTraceBufferUsageAvailable, isolate, promise));
+
+  if (!success) {
+    promise->RejectWithErrorMessage("Could not get trace buffer usage.");
+    return promise->GetHandle();
```
miniak: no need for the duplicate return

docs/api/content-tracing.md
```diff
@@ -122,13 +122,22 @@ all child processes to flush any pending trace data.
...
 
 Get the maximum usage across processes of trace buffer as a percentage of the
 full state. When the TraceBufferUsage value is determined the `callback` is
 called.
+
+**[Deprecated Soon](promisification.md)**
+
+### `contentTracing.getTraceBufferUsage()`
+
+Returns `Promise<Object>` - Resolves with an object containing the `value` and `percentage` of trace buffer maximum usage
```
miniak: I would wait for the `electron.d.ts` generation to be fixed in master to be able to check if it's going to be generated properly.

docs/api/content-tracing.md
```diff
@@ -84,9 +84,19 @@ temporary file. The actual file path will be passed to `callback` if it's not
 ### `contentTracing.getTraceBufferUsage(callback)`
 
 * `callback` Function
-  * `value` Number
-  * `percentage` Number
+  * Object
+    * `value` Number
+    * `percentage` Number
```
ckerr: This is a breaking change, unfortunately, and goes against the project's policy of publicly announcing this at least one major version ahead of time.

We should either add *another* `deprecate()` helper to explicitly address this, or to get a breaking change announcement into the 5.0 betas so that this breaking change can land in 6.0.

atom/browser/api/atom_api_content_tracing.cc
```diff
@@ -115,10 +115,25 @@ v8::Local<v8::Promise> StartTracing(
...
+  mate::Dictionary dict = mate::Dictionary::CreateEmpty(promise->isolate());
+  dict.Set("percentage", percent_full);
+  dict.Set("value", approximate_count);
+
+  promise->Resolve(dict.GetHandle());
+}
+
+v8::Local<v8::Promise> GetTraceBufferUsage(v8::Isolate* isolate) {
+  scoped_refptr<atom::util::Promise> promise = new atom::util::Promise(isolate);
+  bool success = TracingController::GetInstance()->GetTraceBufferUsage(
```
shahzadlone: Please consider making this const as it isn't modified.
```diff
- bool success = ...
+ const bool success = ...
```

miniak: we are not that strict with const for values in the C++ code (except for making const references to avoid copies)

lib/common/api/deprecate.js
```diff
@@ -94,6 +94,32 @@ const deprecate = {
...
+          process.nextTick(() => {
+            // eslint-disable-next-line standard/no-callback-literal
+            cb.length > 2 ? cb(null, ...res) : cb(...res)
+          })
+        }, err => {
+          process.nextTick(() => cb(err))
+        })
+    }
+  },
+
```
ckerr: :tada:

codebytere:
sadness

lib/common/api/deprecate.js
Gam3356: Thank you guys!

PR #16607, opened 2019-01-30
refactor: implement inspector APIs without the remote module

miniak:
#### Description of Change
Eliminates usage of the `remote` module in devtools (`inspector.js`). New IPCs messages are only accepted from devtools for security reasons.

The end goal is to completely eliminate the usage of the `remote` module and then re-enable sandbox for devtools renderers.
https://github.com/electron/electron/blob/9ed89ace976682809f97e165430d65cde493cb4a/atom/browser/atom_browser_client.cc#L497

Helpers taken from https://github.com/electron/electron/pull/16619

/cc @nornagon

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes


lib/browser/chrome-devtools.js
```diff
@@ -0,0 +1,121 @@
...
+  menu.popup({ window })
+})
+
+ipcMainUtils.handle('ELECTRON_INSPECTOR_SELECT_FILE', function (event) {
+  return new Promise((resolve, reject) => {
+    assertChromeDevTools(event.sender, 'window.UI.createFileSelectorElement()')
+
+    dialog.showOpenDialog({}, function (files) {
+      if (files) {
+        const [path] = files
```
alexeykuzmin: `const path = files[0]`?

not a big deal of course

lib/browser/chrome-devtools.js
```diff
@@ -0,0 +1,121 @@
...
+      label: item.label,
+      enabled: item.enabled,
+      checked: item.checked
+    } : {
+      type: 'normal',
+      label: item.label,
+      enabled: item.enabled
+    }
+
+    if (item.id != null) {
```
alexeykuzmin: `!=` and `==` should be avoided, they are confusing.

Is it even possible that `item.id === null` or it's of the `undefined` type?
If you want to check if `item` has `id` property then `'id' in item` is the way.

miniak: I don't like that either, but I just kept the original code, which was doing that

alexeykuzmin: This is sad, but ok.

nornagon:
Why are we re-implementing this stuff? Can we reuse Blink's devtools code instead?

lib/browser/chrome-devtools.js
```diff
@@ -0,0 +1,121 @@
...
+
+ipcMainUtils.handle('ELECTRON_INSPECTOR_CONTEXT_MENU', function (event, items, isEditMenu) {
+  assertChromeDevTools(event.sender, 'window.InspectorFrontendHost.showContextMenuAtPoint()')
+
+  const template = isEditMenu ? getEditMenuItems() : convertToMenuTemplate(event, items)
+  const menu = Menu.buildFromTemplate(template)
+  const window = event.sender.getOwnerBrowserWindow()
+
+  menu.once('menu-will-close', () => {
+    setTimeout(() => {
```
nornagon: Why does this need a `setTimeout`? Can we reply immediately on close?

miniak: @nornagon the close event comes before the click handler, it needs to be delayed

nornagon: does `setTimeout` guarantee that ordering, or just make it more likely?

what happens if they're misordered?

if this really is the right way to do it, please add a comment about why, otherwise this will be lost when this code next gets refactored.

miniak: @nornagon I will add the comment. I don’t like the timeout, ideally we would have an event like `dismissed` indicating that no item was selected. The original version of the code aso had a bug as it didn’t let the devtools know when the menu was closed. So if you would right click several times, dismiss the menu and last time click on an item, it would trigger the selected action multiple times

PR #16608, opened 2019-01-30
refactor: merge event._replyInternal into event.reply

miniak:
#### Description of Change
Have a single `event.reply()`, which calls either `sendToFrame()` or `_sendToFrameInternal()` accordingly. The is also only a single `event.returnValue` property.

/cc @MarshallOfSound

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/api/web-contents.js
+++ b/lib/browser/api/web-contents.js
@@ -342,20 +342,14 @@ WebContents.prototype.loadFile = function (filePath, options = {}) {
   }))
 }
 
-const addReplyToEvent = (event) => {
+const addReplyToEvent = (event, internal) => {
   event.reply = (...args) => {
-    event.sender.sendToFrame(event.frameId, ...args)
-  }
-}
-
-const addReplyInternalToEvent = (event) => {
-  Object.defineProperty(event, '_replyInternal', {
-    configurable: false,
-    enumerable: false,
-    value: (...args) => {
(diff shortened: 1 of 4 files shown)


lib/browser/api/web-contents.js
```diff
@@ -379,23 +373,21 @@ WebContents.prototype._init = function () {
...
     } else {
-      addReplyToEvent(event)
       this.emit('ipc-message', event, channel, ...args)
       ipcMain.emit(channel, event, ...args)
     }
   })
 
   this.on('-ipc-message-sync', function (event, internal, channel, args) {
     addReturnValueToEvent(event)
+    addReplyToEvent(event, internal)
```
codebytere: it looks like this adds an internal reply no matter what? is this an intentional change>

miniak: it always adds `event.reply()`, which `_sendToFrameInternal` for internal messages or `sendToFrame` for public messages.

codebytere:
☝️  I didn't consider this originally, but given that explanation i'm now more inclined to leave this as it was unless there is significant additional value that this refactor confers

miniak:
@MarshallOfSound I've asked about that a week ago in chat, but didn't get a reply. I guess there is no point in arguing about this.

Issue #14606, opened 2018-09-13
Save dialog does not expand/collapse properly on macOS


BONDbATIF:
**Electron Version:**"electron": "^1.8.4"
Operating System (Platform and Version):MAC .10.12 and 10.13

**Expected Behaviour**
After Click "^" on 'Save As'  in showSaveDialog, it should display expanded save as dialog window in MAC.
**_Note:_** working in Windows 

**Actual behavior**
In showSaveDialog , UI shows some abnormal buttons after Click "^" on 'Save As' dialog box.
Note: save as window not Expanded & New folder not responding

**Recreation Steps:**
 this.dialog = this._electronService.remote.dialog;
 this.saveDialog = this.dialog.showSaveDialog({
            title: 'select File Folder',
            defaultPath: '123 copy 1.cxf',
            filters: FilterValues,
        },
     function (fileNames) {
     });
![image](https://user-images.githubusercontent.com/5045450/45485271-96451800-b774-11e8-9087-0f4b1d8d7b88.png)
If i double click the Right side arrow to expand, Most of the time i'm getting the below issue, Some rare cases i'm also getting issue in single click
![image](https://user-images.githubusercontent.com/5045450/45485301-b543aa00-b774-11e8-99a7-68b61473a6d7.png)

1) Please give the solution for the above issue 
2) and also In angular i  programmatically want to set expanded save as dialog by default- as per below snap  (without arrow click on save as )

![image](https://user-images.githubusercontent.com/5045450/45486176-d0fc7f80-b777-11e8-931a-c4ceea5123c7.png)

Romace:
I'm experiencing the similar issue on expanding/collapsing Save Dialog in Electron v.2.0.9. I've made a screen record of Postman as an example.
I've verified that the issue is not reproducible in Electron v.1.6.11.

![postman_savedialog_issue](https://user-images.githubusercontent.com/17388835/45683359-04f6ec80-bb4c-11e8-8c4c-1fcccad7b81a.gif)

sindresorhus:
I can reproduce this on Electron 3.0.2 too.

Anrock:
Seeing this on 1.7. 
In my case file dialog is attached as a sheet to a BrowserWindow and glitchy part is over BrowserView, i suspect this is somehow related.

cwillisf:
As far as I can tell, this issue affects version 1.7.0 and newer. The issue is not present in 1.6.18.

[...]

PR #16612, opened 2019-01-30
fix: Use async save dialog for anchor download attribute

poiru:
On macOS, the synchronous save dialog does not work properly and breaks
in weird ways when the dialog is expanded/collapsed (see #14606). This
works around the problem for the dialog shown for `<a download=...>` by
switching to the asynchronous API.

To test this:

main.js:
```
const { app, BrowserWindow, session } = require('electron')
const path = require('path')

function createWindow() {
  win = new BrowserWindow({
    width: 600,
    height: 600,
    webPreferences: {
      nodeIntegration: false,
    },
  })

  session.defaultSession.on('will-download', (event, item) => {
    // Test with and without this commented out.
    //item.setSavePath(path.join(app.getPath('desktop'), 'test-download.txt'))
  })

  win.loadURL(`file://${__dirname}/test.html`)
}

app.on('ready', createWindow)
```

test.html:
```
<html>
<head>
  <meta http-equiv="Content-Security-Policy" content="default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'; img-src 'self' data:">
</head>
<body>
<a href="data:application/xml;charset=utf-8,foo" download="download.txt">Save</a>
</body>
```

Notes: Fix broken save dialog on macOS for `<a>` downloads

### DIFF
--- a/atom/browser/atom_download_manager_delegate.cc
+++ b/atom/browser/atom_download_manager_delegate.cc
@@ -100,23 +100,58 @@ void AtomDownloadManagerDelegate::OnDownloadPathGenerated(
   if (relay)
     window = relay->GetNativeWindow();
 
-  auto* web_preferences = WebContentsPreferences::From(web_contents);
-  bool offscreen =
-      !web_preferences || web_preferences->IsEnabled(options::kOffscreen);
-
+  // Show save dialog if save path was not set already on item
   base::FilePath path;
   GetItemSavePath(item, &path);
-  // Show save dialog if save path was not set already on item
-  file_dialog::DialogSettings settings;
-  GetItemSaveDialogOptions(item, &settings);
-  if (!settings.parent_window)
-    settings.parent_window = window;
-  settings.force_detached = offscreen;
-  if (settings.title.size() == 0)
-    settings.title = item->GetURL().spec();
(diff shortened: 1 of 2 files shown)


atom/browser/atom_download_manager_delegate.cc
```diff
@@ -100,23 +100,58 @@ void AtomDownloadManagerDelegate::OnDownloadPathGenerated(
...
+      settings.default_path = default_path;
+
+    auto* web_preferences = WebContentsPreferences::From(web_contents);
+    const bool offscreen =
+        !web_preferences || web_preferences->IsEnabled(options::kOffscreen);
+    settings.force_detached = offscreen;
+
+    auto dialog_callback =
+        base::Bind(&AtomDownloadManagerDelegate::OnDownloadSaveDialogDone,
+                   base::Unretained(this), download_id, callback);
```
poiru: I'm using `base::Unretained(this)` here, but I think it should be fine since AtomDownloadManagerDelegate should exist until shutdown. @zcbenz, is this safe?

zcbenz: Yeah it should be fine in this case.

PR #16619, opened 2019-01-30
refactor: use ipcRendererUtils.invoke / ipcMainUtils.handle for desktopCapturer.getSources()

miniak:
#### Description of Change
- Use `ipcRendererUtils.invoke` / `ipcMainUtils.handle` to eliminate duplicate code.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/desktop-capturer.js
+++ b/lib/browser/desktop-capturer.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { ipcMainInternal } = require('@electron/internal/browser/ipc-main-internal')
+const ipcMainUtils = require('@electron/internal/browser/ipc-main-internal-utils')
 
 const { desktopCapturer } = process.atomBinding('desktop_capturer')
 const eventBinding = process.atomBinding('event')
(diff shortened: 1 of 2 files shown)


poiru:
Why not expose this publicly? We have similar code in our to wrap the renderer <-> main IPC API with promises. It would be great to have this built-in.

miniak:
@MarshallOfSound can you please review the current version?

lib/renderer/api/desktop-capturer.js
```diff
@@ -31,8 +23,8 @@ const getSources = (options) => {
   return new Promise((resolve, reject) => {
     if (!isValid(options)) throw new Error('Invalid options')
 
-    const captureWindow = includes.call(options.types, 'window')
-    const captureScreen = includes.call(options.types, 'screen')
+    const captureWindow = options.types.includes('window')
```
MarshallOfSound: These are not equivalent. 

`[].includes.call({}, 'a') === false`
`{}.includes('a') === Error`

lib/browser/desktop-capturer.js
```diff
@@ -10,27 +10,33 @@ const deepEqual = (a, b) => JSON.stringify(a) === JSON.stringify(b)
 // A queue for holding all requests from renderer process.
 let requestsQueue = []
 
-const electronSources = 'ELECTRON_BROWSER_DESKTOP_CAPTURER_GET_SOURCES'
-const capturerResult = (id) => `ELECTRON_RENDERER_DESKTOP_CAPTURER_RESULT_${id}`
+function defer () {
```
MarshallOfSound: I'm not a huge fan of this "deferred" pattern, is there any alternative?

lib/renderer/api/desktop-capturer.js
miniak: isValid enforces that types is an array

lib/browser/desktop-capturer.js
miniak: Me neither

lib/browser/desktop-capturer.js
```diff
@@ -10,27 +10,33 @@ const deepEqual = (a, b) => JSON.stringify(a) === JSON.stringify(b)
...
   const customEvent = eventBinding.createWithSender(event.sender)
   event.sender.emit('desktop-capturer-get-sources', customEvent)
 
   if (customEvent.defaultPrevented) {
-    event._replyInternal(capturerResult(id), [])
-    return
+    return []
   }
 
+  const deferred = defer()
```
MarshallOfSound: What if we did

```js
return new Promise((resolve, reject) => {
  ...
  const request = {
    ...
    resolve,
    reject,
    ....
  }
  ...
})
```

so we could remove this deferred pattern 🤔

lib/browser/desktop-capturer.js
miniak: @MarshallOfSound I've removed the `deferred`, but I still need to store the `resolve` function to keep it working without having to refactor the C++ code.

lib/browser/desktop-capturer.js
miniak: done

lib/browser/desktop-capturer.js
```diff
@@ -10,37 +10,34 @@ const deepEqual = (a, b) => JSON.stringify(a) === JSON.stringify(b)
...
+    }
 
-  // If the WebContents is destroyed before receiving result, just remove the
-  // reference from requestsQueue to make the module not send the result to it.
-  event.sender.once('destroyed', () => {
-    request.event = null
+    // If the WebContents is destroyed before receiving result, just remove the
+    // reference from requestsQueue to make the module not send the result to it.
+    event.sender.once('destroyed', () => {
+      request.resolve = null
```
alexeykuzmin: Should `request` be removed from the `requestsQueue`?
Removing can be done easier if the `requestsQueue` were a `Set` instead of `Array`.

miniak: let's leave it for now. queue kind of needs to be an array, not set

miniak:
@MarshallOfSound can you please check the latest version?

PR #16659, opened 2019-01-31
fix: show proper clerk notes in release notes script

ckerr:
#### Description of Change

Further release note script changes to fix issues reported by @nornagon:

 - [x] Improved the logic that decides when to omit changes that also exist in older branches
 - [x] Fixed linefeed-matching regex bug that caused `Notes` to not be matched in PRs
 - [x] Fixed similar bug that caused the string `BREAKING CHANGE` to not be matched in PR bodies

cc @BinaryMuse @codebytere @nornagon 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/release-notes/notes.js
+++ b/script/release-notes/notes.js
@@ -187,9 +187,9 @@ const parseCommitMessage = (commitMessage, owner, repo, commit = {}) => {
 
   // https://www.conventionalcommits.org/en
   if (commitMessage
-    .split(/\r?\n\r?\n/) // split into paragraphs
-    .map(paragraph => paragraph.trim())
-    .some(paragraph => paragraph.startsWith('BREAKING CHANGE'))) {
+    .split(/\r?\n/) // split into lines
+    .map(line => line.trim())
+    .some(line => line.startsWith('BREAKING CHANGE'))) {
     commit.type = 'breaking-change'
   }
 
--- a/script/release-notes/notes.js
+++ b/script/release-notes/notes.js
@@ -81,10 +81,12 @@ const getNoteFromBody = body => {
   }
 
   const NOTE_PREFIX = 'Notes: '
+  const NOTE_HEADER = '#### Release Notes'
 
   let note = body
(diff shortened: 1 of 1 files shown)


script/release-notes/notes.js
```diff
@@ -81,10 +81,12 @@ const getNoteFromBody = body => {
   }
 
   const NOTE_PREFIX = 'Notes: '
+  const NOTE_HEADER = '#### Release Notes'
 
   let note = body
     .split(/\r?\n\r?\n/) // split into paragraphs
     .map(paragraph => paragraph.trim())
+    .map(paragraph => paragraph.startsWith(NOTE_HEADER) ? paragraph.slice(NOTE_HEADER.length).trim() : paragraph)
     .find(paragraph => paragraph.startsWith(NOTE_PREFIX))
```
nornagon: I'm new to this code so forgive me if I've misunderstood, but shouldn't we be taking the release notes from Clerk's comment, rather than from the PR body?

codebytere: No, we take them from the PR body at present!

ckerr: Fixed in 4e12b2c824e5a9a4ea9180cdffc02bf79fc36e12

ckerr: I was going to do it in a followup since it's a little more intrusive than the other fixes, but I guess it doesn't hurt to do it here.

ckerr:
As a side note, pulling from Clerk surfaces this clerk issue: https://github.com/electron/clerk/issues/20

So if you re-run `./script/release-notes/index.js v5.0.0-beta.1` with this PR you'll see a new regression:

```
## Other changes
 * <!-- One-line Change Summary Here--> improved tray icon context menu and menu bar accessibility. #15302
```

codebytere:
Shall we try to fix this at the clerk level?

ckerr:
IMO yes. It's slightly better for the `e/e` script to act as more of a pass-through

PR #16663, opened 2019-01-31
build: hack around GitHub upload API failure / flake

MarshallOfSound:
This is a super hacky workaround to the GitHub API failure.  Assuming a case where we want to upload file "X" instead we upload that file to "fake-X-fake-DATE" and then on success, rename that file.  This way if a file we upload dies like we have been seeing the "DATE" stamped file name will be broken, not the actual file name we want to end up using.

Notes: no-notes


BinaryMuse:
@MarshallOfSound What's the plan for dealing with the stale assets that get left in the release?

codebytere:
@BinaryMuse if assets are broken, they won't appear, and if they do appear this handles that!

Issue #16632, opened 2019-01-30
renable failing tests on Windows 32 bit


jkleinsc:
As part of #16610 3766bf6369e42149ffe29f54103090f2673c0050 disabled tests for 32 bit Windows because they are holding up CI on master.  These tests should be fixed and renabled.

PR #16671, opened 2019-02-01
fix: shutdown after message loop is ready

zcbenz:
#### Description of Change

When `app.quit()` is called while the message loop is not fully started, do not shutdown until message loop is ready.

Quitting too early would break Chromium's shutdown code, and leave a few defunct processes behind.

Close https://github.com/electron/electron/issues/16632.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix defunct processes after quitting.


deepak1556:
@zcbenz should this be backported to 4-0-x as well, where the changes originated from https://github.com/electron/electron/commit/f5eaa97d9d7f8a23f077e41a91623d57c32d5bdc#diff-735da1d7bca6fa4542f4eea75dc6595d

zcbenz:
@deepak1556 4-0-x is working fine so I think we can leave it be. It is required in 5-0-x because CI is broken there.

PR #16687, opened 2019-02-01
fix: use a new Electron dedicated ABI number for Electron 4.0

MarshallOfSound:
This is the outcome of https://github.com/nodejs/TSC/issues/651 and should fix everyone having issues building / using native modules on Electron 4 with strange ABI mis-match issues.

Brings in: https://github.com/electron/node/commit/8bc5d171a0873c0ba49f9433798bc8b67399788c

Notes: Change the `NODE_MODULE_VERSION` to 69

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -12,7 +12,7 @@ vars = {
   'chromium_version':
     '69.0.3497.106',
   'node_version':
-    '4d44266b78256449dd6ae86e419e3ec07257b569',
+    '8bc5d171a0873c0ba49f9433798bc8b67399788c',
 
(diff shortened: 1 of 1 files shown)


trusktr:
Is this problem resurrected? No matter how I try to rebuild, I always end up with:

```
was compiled against a different Node.js version using
NODE_MODULE_VERSION 72. This version of Node.js requires
NODE_MODULE_VERSION 73. Please try re-compiling or re-installing
the module
```

and if you look on the [Node ABI releases](https://nodejs.org/en/download/releases/), it only goes up to `72`. There's no `73`.

`node -v`: 12.12.0
'npm -v: 6.11.3

Any ideas?

PR #16701, opened 2019-02-02
[wip] chore: add release wg codeownership of script

codebytere:
#### Description of Change

Update primary codeownership of the `script/` folder to `@wg-releases`. This is a broad stroke, as not every single file in there is release-specific.  However, I would hedge that 95% of changes to that folder are in fact release-specific, so this seems to me to be more efficient than marking specific files under the new WG.

cc @electron/wg-releases 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


.github/CODEOWNERS
```diff
@@ -18,4 +18,7 @@
 /default_app/ @electron/docs
 /docs/ @electron/docs
 /docs-translations/ @electron/i18n
-/npm/ @electron/electrocats
\ No newline at end of file
+/npm/ @electron/electrocats
+
+# WG CODEOWNER Patterns
+/script/ @electron/wg-releases
```
nornagon: this seems overly broad. Can we move all release-related scripts to e.g. `script/release` and protect that instead?

codebytere: my concern is that it'd break all our backports and i'd love to not do that i think

codebytere: i can try though :)

nornagon:
well, i'm 👎 on this idea at all then.

codebytere:
Please hold on merging this until i've had time to think it through a little more, since it would cascade into changes with sudowoodo I'd potentially like to avoid.

codebytere:
This needs some work and may be best done in a slightly different way in the near future.

Issue #16699, opened 2019-02-02
Touch ID support for macOS


idoodler:
**Is your feature request related to a problem? Please describe.**
I am building an Application that requires frequent password input. It would be awesome, to have native Touch ID support on macOS.

**Describe the solution you'd like**
An easy solution to show the Touch ID authentication dialog on macOS.

**Describe alternatives you've considered**
Writing my own Node.js add-on, but I couldn't wrap my head around it

**Additional context**
<img width="1085" alt="touch bar bild 2019-02-02 um 09 58 04" src="https://user-images.githubusercontent.com/6825153/52162266-71285600-26d1-11e9-80a7-5f85cc50bf90.png">

codebytere:
Related: https://developer.apple.com/documentation/localauthentication?language=objc

codebytere:
@idoodler I took a stab at this today, I think i can get it out in the next beta or two of 5.0 potentially!

nornagon:
Is this functionality perhaps better suited to a 3rd-party native module?

codebytere:
I considered that initially, but I think that we have a significant amount of macOS-specific code in `systemPreferences` as it stands, and that this feature would be more difficult to wrangle as a native node module subject to compilation difficulties against Node. It would also spread out maintenance responsibilities, where here we can support it more easily and better guarantee its functionality.

emilbayes:
I made a module a while back that does that for my electron app (no real security, just checks that you have a valid identity as a boolean): https://www.npmjs.com/package/macos-touchid

appinteractive:
You can use now the native electron feature: https://www.electronjs.org/docs/api/system-preferences#systempreferencesprompttouchidreason-macos

PR #16707, opened 2019-02-03
feat: add Touch ID authentication support for macOS

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/16699.

This PR adds Touch ID authentication support for macOS with two new `SystemPreferences` methods. 

1. `systemPreferences.promptForTouchID()` returns a Promise that resolves with `true` if successful and rejects with an error message if authentication could not be completed.
2. `systemPreferences.isTouchIDAvailable()` returns a Boolean that's `true` if this device is a Mac running a supported OS that has the necessary hardware for Touch ID and `false` otherwise.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added native Touch ID authentication support for macOS.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -689,6 +689,7 @@ if (is_mac) {
 
     libs = [
       "AVFoundation.framework",
+      "LocalAuthentication.framework",
       "Carbon.framework",
       "QuartzCore.framework",
       "Quartz.framework",
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -689,8 +689,8 @@ if (is_mac) {
 
     libs = [
       "AVFoundation.framework",
-      "LocalAuthentication.framework",
       "Carbon.framework",
+      "LocalAuthentication.framework",
       "QuartzCore.framework",
       "Quartz.framework",
       "Security.framework",
--- a/atom/browser/api/atom_api_system_preferences.cc
+++ b/atom/browser/api/atom_api_system_preferences.cc
@@ -93,6 +93,7 @@ void SystemPreferences::BuildPrototype(
       .SetMethod("setAppLevelAppearance",
                  &SystemPreferences::SetAppLevelAppearance)
       .SetMethod("getSystemColor", &SystemPreferences::GetSystemColor)
+      .SetMethod("isTouchIDAvailable", &SystemPreferences::IsTouchIDAvailable)
       .SetMethod("promptTouchID", &SystemPreferences::PromptTouchID)
       .SetMethod("isTrustedAccessibilityClient",
                  &SystemPreferences::IsTrustedAccessibilityClient)
--- a/atom/browser/api/atom_api_system_preferences.cc
+++ b/atom/browser/api/atom_api_system_preferences.cc
@@ -93,6 +93,7 @@ void SystemPreferences::BuildPrototype(
       .SetMethod("setAppLevelAppearance",
                  &SystemPreferences::SetAppLevelAppearance)
       .SetMethod("getSystemColor", &SystemPreferences::GetSystemColor)
+      .SetMethod("promptTouchID", &SystemPreferences::PromptTouchID)
       .SetMethod("isTrustedAccessibilityClient",
                  &SystemPreferences::IsTrustedAccessibilityClient)
       .SetMethod("getMediaAccessStatus",
--- a/atom/browser/api/atom_api_system_preferences.cc
+++ b/atom/browser/api/atom_api_system_preferences.cc
@@ -93,7 +93,7 @@ void SystemPreferences::BuildPrototype(
       .SetMethod("setAppLevelAppearance",
                  &SystemPreferences::SetAppLevelAppearance)
       .SetMethod("getSystemColor", &SystemPreferences::GetSystemColor)
-      .SetMethod("isTouchIDAvailable", &SystemPreferences::IsTouchIDAvailable)
+      .SetMethod("canPromptTouchID", &SystemPreferences::CanPromptTouchID)
       .SetMethod("promptTouchID", &SystemPreferences::PromptTouchID)
       .SetMethod("isTrustedAccessibilityClient",
                  &SystemPreferences::IsTrustedAccessibilityClient)
--- a/atom/browser/api/atom_api_system_preferences_mac.mm
+++ b/atom/browser/api/atom_api_system_preferences_mac.mm
@@ -443,6 +443,16 @@ std::string SystemPreferences::GetSystemColor(const std::string& color,
   }
 }
 
+bool SystemPreferences::IsTouchIDAvailable() {
+  if (@available(macOS 10.12.2, *)) {
+    base::scoped_nsobject<LAContext> context([[LAContext alloc] init]);
+    return [context
+        canEvaluatePolicy:LAPolicyDeviceOwnerAuthenticationWithBiometrics
+                    error:nil];
+  }
+  return false;
+}
+
 void OnTouchIDCompleted(scoped_refptr<util::Promise> promise, bool success) {
   promise->Resolve(success);
 }
@@ -451,7 +461,7 @@ v8::Local<v8::Promise> SystemPreferences::PromptTouchID(
     v8::Isolate* isolate,
     const std::string& reason) {
   scoped_refptr<util::Promise> promise = new util::Promise(isolate);
-  if (@available(macOS 10.12.1, *)) {
+  if (@available(macOS 10.12.2, *)) {
     base::scoped_nsobject<LAContext> context([[LAContext alloc] init]);
     base::ScopedCFTypeRef<SecAccessControlRef> access_control =
         base::ScopedCFTypeRef<SecAccessControlRef>(
@@ -481,7 +491,7 @@ v8::Local<v8::Promise> SystemPreferences::PromptTouchID(
                         }];
   } else {
     promise->RejectWithErrorMessage(
-        "This API is not available on macOS versions older than 10.12.1");
+        "This API is not available on macOS versions older than 10.12.2");
   }
   return promise->GetHandle();
 }
--- a/atom/browser/api/atom_api_system_preferences_mac.mm
+++ b/atom/browser/api/atom_api_system_preferences_mac.mm
@@ -470,13 +470,18 @@ v8::Local<v8::Promise> SystemPreferences::PromptTouchID(
                     operation:LAAccessControlOperationUseKeySign
               localizedReason:[NSString stringWithUTF8String:reason.c_str()]
                         reply:^(BOOL success, NSError* error) {
-                          runner->PostTask(FROM_HERE,
-                                           base::BindOnce(&OnTouchIDCompleted,
+                          if (!success) {
+                            promise->RejectWithErrorMessage(std::string(
+                                [error.localizedDescription UTF8String]));
+                          } else {
+                            runner->PostTask(
+                                FROM_HERE, base::BindOnce(&OnTouchIDCompleted,
                                                           promise, !!success));
+                          }
                         }];
   } else {
     promise->RejectWithErrorMessage(
-        "This API is not available on macOS older than 10.12");
+        "This API is not available on macOS versions older than 10.12.1");
   }
   return promise->GetHandle();
 }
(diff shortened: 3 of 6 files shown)


docs/api/system-preferences.md
```diff
@@ -380,6 +380,20 @@ You can use the `setAppLevelAppearance` API to set this value.
 Sets the appearance setting for your application, this should override the
 system default and override the value of `getEffectiveAppearance`.
 
+### `systemPreferences.promptTouchID(reason)` _macOS_
```
MarshallOfSound: My concern here is that this method **by itself** doesn't actually provide any true security for user-data / app-data.  Apps would have to implement some kind of locking mechanism based on a `true` / `false` value here through something like `keytar`.  (Store an encryption key in `keytar` and only fetch it if `promptTouchID` resolves with `true`).

Not exactly sure how I would phrase it, but I feel as though we need to do due diligence here and warn people people they have to implement real security / encryption based around this method.  E.g. The way native apps would do this is by setting [`kSecAccessControlUserPresence `](https://developer.apple.com/documentation/security/secaccesscontrolcreateflags/ksecaccesscontroluserpresence?language=objc) on their keychain entry so that reading it would auto-prompt for biometrics.

docs/api/system-preferences.md
```diff
@@ -380,6 +380,20 @@ You can use the `setAppLevelAppearance` API to set this value.
...
+
+* `reason` String - The reason you are asking for Touch ID authentication
+
+Returns `Promise<Boolean>` - whether ot not the user has successfully authenticated with Touch ID.
+
+```javascript
+const { systemPreferences } = require('electron')
+
+systemPreferences.promptTouchID('To get consent for In-App Purchase').then(success => {
```
MarshallOfSound: Can we use an example string here that doesn't mention in-app purchases, don't want people to think that this method is required / can be used for IAP's

atom/browser/api/atom_api_system_preferences_mac.mm
```diff
@@ -438,6 +443,49 @@ AVMediaType ParseMediaType(const std::string& media_type) {
...
+                            promise->RejectWithErrorMessage(std::string(
+                                [error.localizedDescription UTF8String]));
+                          } else {
+                            runner->PostTask(
+                                FROM_HERE, base::BindOnce(&OnTouchIDCompleted,
+                                                          promise, !!success));
+                          }
+                        }];
+  } else {
+    promise->RejectWithErrorMessage(
```
MarshallOfSound: Can we have a secondary method or property somewhere `isTouchIDAvailable`

BUILD.gn
```diff
@@ -591,6 +591,7 @@ if (is_mac) {
 
     libs = [
       "AVFoundation.framework",
+      "LocalAuthentication.framework",
```
ckerr: _(nit)_ is the order important, or can this new lib be added in alphabetical order like the others?

atom/browser/api/atom_api_system_preferences.cc
```diff
@@ -93,6 +93,8 @@ void SystemPreferences::BuildPrototype(
       .SetMethod("setAppLevelAppearance",
                  &SystemPreferences::SetAppLevelAppearance)
       .SetMethod("getSystemColor", &SystemPreferences::GetSystemColor)
+      .SetMethod("isTouchIDAvailable", &SystemPreferences::IsTouchIDAvailable)
+      .SetMethod("promptTouchID", &SystemPreferences::PromptTouchID)
```
ckerr: Is SystemPreferences the right place for this functionality? It doesn't seem preference-related.

atom/browser/api/atom_api_system_preferences_mac.mm
```diff
@@ -438,6 +443,59 @@ AVMediaType ParseMediaType(const std::string& media_type) {
...
+        base::SequencedTaskRunnerHandle::Get();
+
+    [context
+        evaluateAccessControl:access_control
+                    operation:LAAccessControlOperationUseKeySign
+              localizedReason:[NSString stringWithUTF8String:reason.c_str()]
+                        reply:^(BOOL success, NSError* error) {
+                          if (!success) {
+                            promise->RejectWithErrorMessage(std::string(
+                                [error.localizedDescription UTF8String]));
```
ckerr: Does it matter what thread the promise is rejected in? This code block "is evaluated on a private queue internal to the framework in an unspecified threading context" [(source)](https://developer.apple.com/documentation/localauthentication/lacontext/1514174-evaluateaccesscontrol?language=objc)

atom/browser/api/atom_api_system_preferences_mac.mm
```diff
@@ -438,6 +443,59 @@ AVMediaType ParseMediaType(const std::string& media_type) {
   }
 }
 
+bool SystemPreferences::IsTouchIDAvailable() {
+  if (@available(macOS 10.12.2, *)) {
+    base::scoped_nsobject<LAContext> context([[LAContext alloc] init]);
+    return [context
+        canEvaluatePolicy:LAPolicyDeviceOwnerAuthenticationWithBiometrics
+                    error:nil];
```
ckerr: I _think_ this is just returning that _some_ biometric is available, but that it could be face or touch.

https://stackoverflow.com/a/48901055 indicates we might also want to check context's biometryType for LABiometryTypeTouchID

atom/browser/api/atom_api_system_preferences_mac.mm
```diff
@@ -438,6 +443,59 @@ AVMediaType ParseMediaType(const std::string& media_type) {
...
+
+void OnTouchIDCompleted(scoped_refptr<util::Promise> promise, bool success) {
+  promise->Resolve(success);
+}
+
+v8::Local<v8::Promise> SystemPreferences::PromptTouchID(
+    v8::Isolate* isolate,
+    const std::string& reason) {
+  scoped_refptr<util::Promise> promise = new util::Promise(isolate);
+  if (@available(macOS 10.12.2, *)) {
```
ckerr: Would it make sense to use `IsTouchIDAvailable()` here instead of just checking for 10.12.2? That way the other biometrics tests in that function will can be reused here and let us error out gracefully in this func's promise->RejectWithErrorMessage()`

docs/api/system-preferences.md
```diff
@@ -380,6 +380,30 @@ You can use the `setAppLevelAppearance` API to set this value.
 Sets the appearance setting for your application, this should override the
 system default and override the value of `getEffectiveAppearance`.
 
+### `systemPreferences.isTouchIDAvailable()` _macOS_
+
+Returns `Boolean` - whether or not this device has the ability to use Touch ID.
+
+**NOTE:** This API is only available on macOS Sierra 10.12.2 or newer.
```
ckerr: _(pedantic)_ This API appears to be available on all versions of macOS; it just returns `false` on older versions. The current phrasing implies that app devs need to test their OS version before calling this function.

How about:

> **NOTE:** Touch ID is only available on macOS Sierra 10.12.2 or newer.

docs/api/system-preferences.md
```diff
@@ -380,6 +380,30 @@ You can use the `setAppLevelAppearance` API to set this value.
...
+
+### `systemPreferences.promptTouchID(reason)` _macOS_
+
+* `reason` String - The reason you are asking for Touch ID authentication
+
+Returns `Promise<Boolean>` - whether ot not the user has successfully authenticated with Touch ID.
```
ckerr: _(nit)_ s/ot/or/

_(non-nit)_ This doc implies that the app dev should test the boolean to see if they succeeded, but the boolean is always True. IMO this should be a `Promise<void>`

docs/api/system-preferences.md
```diff
@@ -380,6 +380,30 @@ You can use the `setAppLevelAppearance` API to set this value.
...
+
+```javascript
+const { systemPreferences } = require('electron')
+
+systemPreferences.promptTouchID('To get consent for a Security-Gated Thing').then(success => {
+  console.log('You have successfully authenticated with Touch ID!')
+})
```
ckerr: _(optional)_ would be good to document handling the error case too, really just to demonstrate that the LAError message that we send back is suitable for being presented to the end user

docs/api/system-preferences.md
```diff
@@ -380,6 +380,30 @@ You can use the `setAppLevelAppearance` API to set this value.
...
+```
+
+This API itself will not protect your user data; rather, it is a mechanism to allow you to do so. Native apps will need to set [Access Control Constants](https://developer.apple.com/documentation/security/secaccesscontrolcreateflags?language=objc) like [`kSecAccessControlUserPresence`](https://developer.apple.com/documentation/security/secaccesscontrolcreateflags/ksecaccesscontroluserpresence?language=objc) on the their keychain entry so that reading it would auto-prompt for Touch ID biometric consent. This could be done with [`node-keytar`](https://github.com/atom/node-keytar), such that one would store an encryption key with `node-keytar` and only fetch it if `promptTouchID()` resolves with `true`.
```
ckerr: :100:

ckerr: (note, if you take my suggestion to make prompTouchID() a Promise<Void>, the last part of this sentence should be updated)

atom/browser/api/atom_api_system_preferences.cc
codebytere: i thought about this too, but wasn't sure there was any more appropriate place 🤔

atom/browser/api/atom_api_system_preferences_mac.mm
codebytere: I thought about this, but the challenge is that `LABiometryType` is an enum that wasn't added until `10.13.2`. So we could, but _many_ users are still on 10.12 so we'd be locking out a whole upgrade's worth of users by doing that...what do you think?

atom/browser/api/atom_api_system_preferences_mac.mm
codebytere: I think in this case it'd be fine? Chrome did [post it back in the original thread](https://cs.chromium.org/chromium/src/device/fido/mac/touch_id_context.mm?sq=package:chromium&g=0&l=112) but i don't think we can correspondingly do that with promises

docs/api/system-preferences.md
codebytere: The problem with this is that `TouchID` is available on versions prior to `10.12.2`. The API to interact with it, however, is not, which causes this issue.

atom/browser/api/atom_api_system_preferences_mac.mm
ckerr: @codebytere points out this doesn't make sense because @available is like an `#ifdef` in that we need this at compile time to ensure use of the api tokens introduced in 10.12.2 :+1:

atom/browser/api/atom_api_system_preferences_mac.mm
ckerr: This is resolved by posting to `OnTouchIDFailed()`

codebytere:
I gotta forward-declare the instance method `biometryType` and the enum `LABiometryType` 👇

alchaplinsky:
Was this eventually removed from 5.0x?  It doesn't look like systemPreferences module has these methods anymore. And they are also missing in the [documentation](https://electronjs.org/docs/api/system-preferences)

codebytere:
@alchaplinsky this was never backported to 5; it's only present in 6 onward.

alchaplinsky:
@codebytere Ok, got it. Thank you for response. Will try out v6.

sunknudsen:
> This API itself will not protect your user data; rather, it is a mechanism to allow you to do so. Native apps will need to set [Access Control Constants](https://developer.apple.com/documentation/security/secaccesscontrolcreateflags?language=objc) like [kSecAccessControlUserPresence](https://developer.apple.com/documentation/security/secaccesscontrolcreateflags/ksecaccesscontroluserpresence?language=objc) on their keychain entry so that reading it would auto-prompt for Touch ID biometric consent. This could be done with [node-keytar](https://github.com/atom/node-keytar), such that one would store an encryption key with node-keytar and only fetch it if promptTouchID() resolves.

How does one “set [Access Control Constants](https://developer.apple.com/documentation/security/secaccesscontrolcreateflags?language=objc) like [kSecAccessControlUserPresence](https://developer.apple.com/documentation/security/secaccesscontrolcreateflags/ksecaccesscontroluserpresence?language=objc) on their keychain entry”?

PR #16718, opened 2019-02-04
refactor: remove js2asar.py and port logic to JS in more readable / GN-style way

MarshallOfSound:
This ASAR refactor is working towards the Typescript PR #16441 

We now use `gn` copy steps to copy all JS to a `gen` directory and call a `gn-asar.js` file with those files which then uses the `asar` JS API instead of spawning the asar CLI.

In the TS future the "copy to gen folder" step will actually be "run tsc with the gen folder as the output dir"

Notes: no-notes


BUILD.gn
```diff
@@ -156,18 +159,49 @@ asar("js2asar") {
...
   sources = filenames.default_app_sources
+  outputs = [
+    "$target_gen_default_app_js/{{source}}",
+  ]
+}
+
+copy("copy_default_app_octicon_deps") {
+  sources = filenames.default_app_octicon_sources
+  outputs = [
+    "$target_gen_default_app_js/electron/default_app/octicon/{{source_file_part}}",
```
MarshallOfSound: Instead of some magical `node_modules` redirect / copy here we put it in a reasonable folder "octicon" and load relative to `__dirname` in code

tools/build-asar.py
```diff
@@ -0,0 +1,14 @@
+import os
```
nornagon: IMO it would be preferable to reuse `npm_action` for this and put the `gn-asar.js` invocation in an npm script.

script/gn-asar.js
```diff
@@ -0,0 +1,58 @@
...
+
+  // Create the ASAR archive
+  await new Promise((resolve, reject) => {
+    asar.createPackageWithOptions(tmpPath, out[0], {}, (err) => {
+      if (err) return reject(err)
+      resolve()
+    })
+  })
+})().then(() => fs.remove(tmpPath)).catch((err) => {
+  console.error('Unexpected error while generating ASAR', err)
```
nornagon: use try/finally inside the `async`

BUILD.gn
```diff
@@ -156,18 +159,49 @@ asar("js2asar") {
...
+    "$target_gen_electron_js/{{source}}",
+  ]
+}
+
+asar("electron.asar") {
+  deps = [
+    ":copy_lib_js",
+  ]
+
+  root = "$target_gen_electron_js/electron/lib"
```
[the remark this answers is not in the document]
MarshallOfSound: @nornagon It seems quite complex to strip out those `electron`'s with `rebase_path` so I just owned them 😆 The `tsc` stuff will strip a few of them anyway so that's a win 👍

tools/build-asar.py
MarshallOfSound: @nornagon Thought about this, how about a new `node_action` that runs an arbitrary `.js` script.  Not too keen on polluting the `package.json.scripts` section just to make GN commands work

BUILD.gn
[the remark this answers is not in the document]
MarshallOfSound: `gen/electron` is defined by` target_gen_dir`
the "electron" in `js/electron` is required because of the `copy` using `{{source}}` (which starts with `electron`.  It's accept the weird path or do a bunch of path extraction and `rebase_path` calls.

I'd personally rather have the annoying but explainable paths than have to strip apart a deep chain of `rebase_path` and `get_path_info`

dhewya:
https://github.com/electron/electron/pull/16441#issuecomment-464079883

PR #16721, opened 2019-02-04
chore: update issue template, comment out instructions

nornagon:
Hopefully this will make it easier to read issues that come in, & make it less likely that people leave the template text in.

Notes: no-notes


nornagon:
I'd also like @sofianguy and @MarshallOfSound to weigh in please!

sofianguy:
Looks good @nornagon! Thanks for updating

PR #16739, opened 2019-02-05
refactor: implement chrome.i18n.getMessage() without the remote module

miniak:
#### Description of Change
- Follow up to [refactor: implement inspector APIs without the remote module](https://github.com/electron/electron/pull/16607).
- Eliminates usage of the `remote` module in `chrome.i18n.getMessage()` implementation (`i18n.js`).

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/chrome-extension.js
+++ b/lib/browser/chrome-extension.js
@@ -4,9 +4,11 @@ const { app, webContents, BrowserWindow } = require('electron')
 const { getAllWebContents } = process.atomBinding('web_contents')
 const renderProcessPreferences = process.atomBinding('render_process_preferences').forAllWebContents()
 const { ipcMainInternal } = require('@electron/internal/browser/ipc-main-internal')
+const ipcMainUtils = require('@electron/internal/browser/ipc-main-internal-utils')
 
 const { Buffer } = require('buffer')
 const fs = require('fs')
+const os = require('os')
 const path = require('path')
 const url = require('url')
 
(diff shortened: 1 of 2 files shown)


lib/browser/chrome-extension.js
```diff
@@ -201,6 +198,34 @@ ipcMainInternal.on('CHROME_TABS_SEND_MESSAGE', function (event, tabId, extension
   resultID++
 })
 
+const getLanguage = () => {
+  return app.getLocale().replace(/-.*$/, '').toLowerCase()
```
MarshallOfSound: `app.getLocale()` and `navigator.language` aren't always the same thing right?

miniak: @MarshallOfSound I am not sure. I wanted to avoid passing the language via IPC as it would have to be validated.

miniak:
@MarshallOfSound the PR is longer changing behavior, I kept passing `navigator.language` from the renderer.

nornagon:
Perhaps we could do some research on how `navigator.language` is filled? there should be no need to get it from the renderer.

miniak:
@nornagon I can check the implementation of `navigator.language`

miniak:
[NavigatorLanguage::language()](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/frame/navigator_language.cc)

deepak1556:
The value of `navigator.language` is based on the `accept_languages` renderer preferences filled for a `webContents` https://github.com/electron/electron/blob/master/atom/browser/api/atom_api_web_contents.cc#L408 , we try to keep it the same as the `app.getLocale`

miniak:
@nornagon, @deepak1556, @MarshallOfSound what should I do then? Should I use `app.getLocale()` in the main process instead of passing `navigator.language` from the renderer. Or should I keep sending the value over IPC and validate it? Can we agree on a solution to unblock this PR?

deepak1556:
IMO it is fine to use `app.getLocale` as its the primary source of truth for `navigator.languages`, the value of `navigator.language` is not something a user changes in runtime. Also we have tests to guarantee this https://github.com/electron/electron/blob/master/spec/chromium-spec.js#L202.

MarshallOfSound:
Based on @deepak1556 's description of how `navigator.languages` is filled using `app.getLocale()` should be 🆗

miniak:
@nornagon, @deepak1556, @MarshallOfSound changed back to use `app.getLocale()` as discussed. Can you please approve?

PR #16760, opened 2019-02-05
build: make mac CI 15 minutes faster by using notion for nodejs

MarshallOfSound:
Before: Install Node.JS takes 6 minutes * 3
After: Install Node.JS takes 40 seconds * 3

This change should make macOS CI faster by ~15 minutes overall.  Couldn't make `nvm` work but shoutout to @felixrieseberg for pointing me in the direction of `notion` which works _flawlessly_

Notes: no-notes

cc @jkleinsc

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,9 +173,13 @@ step-install-nodejs-on-mac: &step-install-nodejs-on-mac
     name: Install Node.js 10 on MacOS
     command: |
       if [ "`uname`" == "Darwin" ]; then
-        brew update
-        brew install node@10
-        echo 'export PATH="/usr/local/opt/node@10/bin:$PATH"' >> $BASH_ENV
+        curl -sSLf https://get.notionjs.com | bash
+        export NOTION_HOME="$HOME/.notion"
+        source "$NOTION_HOME/load.sh"
+        notion install node 10
+        echo 'export NOTION_HOME="$HOME/.notion"' >> $BASH_ENV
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -172,19 +172,22 @@ step-install-nodejs-on-mac: &step-install-nodejs-on-mac
...
+        fi
 
 step-install-gnutar-on-mac: &step-install-gnutar-on-mac
   run:
     name: Install gnu-tar on macos
     command: |
       if [ "`uname`" == "Darwin" ]; then
-        brew update
-        brew install gnu-tar
+        HOMEBREW_NO_AUTO_UPDATE=1 brew install gnu-tar
```
MarshallOfSound: This also makes `brew install` faster (we don't need to update to install this, it's already in the brew local)

.circleci/config.yml
```diff
@@ -172,19 +172,22 @@ step-install-nodejs-on-mac: &step-install-nodejs-on-mac
   run:
     name: Install Node.js 10 on MacOS
     command: |
-      if [ "`uname`" == "Darwin" ]; then
-        brew update
-        brew install node@10
-        echo 'export PATH="/usr/local/opt/node@10/bin:$PATH"' >> $BASH_ENV
-      fi
+        if [ "`uname`" == "Darwin" ]; then
```
codebytere: looks like an extra tab snuck in here

MarshallOfSound:
Closing out for now as `notion` does not support `npm` yet (did not realize)

PR #16765, opened 2019-02-06
fix:webframe spellcheck client leak for sandbox mode

nitsakh:
#### Description of Change
RenderFrameObserver OnDestruct method is only called in non sandbox mode. So spellcheck client was destroyed only in non sandbox mode.
This fixes that and moves the spellcheck client destruction to `WillReleaseScriptContext` which will be called in both sandbox and non-sandbox mode.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed webframe spellcheck provider memory leak for sandbox mode


zcbenz:
I have created https://github.com/electron/electron/pull/16770 based on the changes here, which should hopefully fix most leak sources.

nitsakh:
Thanks! I'll close this one then.

PR #16766, opened 2019-02-06
ci: cache brew update result

MarshallOfSound:
This is a secondary attempt to make macOS CI faster by targeting the low-hanging-fruit of "why does it take 6 minutes to install node 10".

Turns out it's `brew update`, if we cache the result of `brew update` we drop from 6 minutes --> 30 seconds + 10 second cache restore.

This change should save us ~15 minutes for an end-to-end macOS CI run

Notes: no-notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -458,6 +458,11 @@ step-maybe-generate-typescript-defs: &step-maybe-generate-typescript-defs
 # Lists of steps.
 steps-checkout: &steps-checkout
   steps:
+    - restore_cache:
+        paths:
+          - /usr/local/Homebrew
+        keys:
+          - v1-brew-cache-{{ arch }}
     - *step-checkout-electron
     - *step-depot-tools-get
     - *step-depot-tools-add-to-path
@@ -481,6 +486,10 @@ steps-checkout: &steps-checkout
         paths:
           - ~/.gclient-cache
         key: v1-gclient-cache-{{ arch }}-{{ checksum "src/electron/DEPS" }}
+    - save_cache:
+        paths:
+          - /usr/local/Homebrew
+        key: v1-brew-cache-{{ arch }}
 
     - run:
         name: Remove some unused data to avoid storing it in the workspace


MarshallOfSound:
Build times vary a lot on macOS due to cache restore time, but here is a quick before vs after

#### Before

screen shot 2019-02-05 at 8 59 47 pm

#### After

screen shot 2019-02-05 at 8 58 18 pm

.circleci/config.yml
```diff
@@ -755,9 +800,10 @@ jobs:
     <<: *steps-checkout
 
   mac-checkout:
-    <<: *machine-mac-large
+    <<: *machine-linux-2xlarge
     environment:
-      <<: *env-mac-large
+      <<: *env-linux-2xlarge
+      GCLIENT_EXTRA_ARGS: '--custom-var=checkout_mac=True --custom-var=host_os=mac'
```
deepak1556: I think it should be `target_os` and `target_os_only` instead https://cs.chromium.org/chromium/tools/depot_tools/gclient.py?l=64-81. `checkout_mac`will get the right value based on it.

deepak1556: If this doesn't fix getting the right binaries as a part of gclient hooks, then I am nervous about this change. I would like to avoid re-downloading the right binaries in a separate step, its not worth the benefit of a checkout time saving.

MarshallOfSound: I tried `target_os` and it doesn't do what we need.  We need it think it's running on macOS not building for macOS.

I would argue that 15+ minutes less build time is 100% worth re-downloading the clang binaries

deepak1556: Its not just clang binary, are other binaries and buildtools chromium downloads during the hooks correspond to the right os ? If thats unaffected then I am good with this change.

deepak1556: Cool, thanks for confirming!

nornagon: wrong clang binary when setting `host_os` explicitly seems like a bug upstream. can we fix it there?

MarshallOfSound: @nornagon It's something we can follow up on, I looked into supporting it but the file that downloads clang uses `sys.platform` *everywhere* 😄 Patchable but not required to make this work 👍

PR #16767, opened 2019-02-06
docs: added webContents.getType() method

emer7:
Notes: no notes


docs/api/web-contents.md
```diff
@@ -34,6 +34,10 @@ console.log(webContents)
 Returns `WebContents[]` - An array of all `WebContents` instances. This will contain web contents
 for all windows, webviews, opened devtools, and devtools extension background pages.
 
+### `webContents.getType()`
+
+Returns `String` - the type of the webContent.
```
MarshallOfSound: This can be enumerated so the type is correct.

```suggestion
Returns `String` - the type of the webContent.  Can be `X`, `Y` or `Z`
```

PR #16770, opened 2019-02-06
Fix memory leak when using webFrame and spell checker

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/15459.

When using spell checker in pages with sandbox, there are two sources of memory leak:
1. Every time when page is reloaded, the code would create a new `api::WebFrame`, and the old one would be leaked since the renderer process is not restarted.
2. The spell check client is only destroyed when web frame is destroyed, but reloading a frame does not destroy the corresponding `blink::WebFrame`.

This PR fixes the leak with 2 methods:
1. Refactor the webFrame module so it does not create `api::WebFrame` instances, the native methods become simple getters without a global state. So no memory will be leaked on reload, and the code now is cleaner and uses less memory.
2. The spell check client is now destroyed when the page context is destroyed, so it won't hold a reference to code in the old context, which prevents garbage collection of V8.

Note that this PR should be rebased instead of squashed as the commits have clear responsibilities.

#### Checklist

- [x] `npm test` passes

#### Release Notes


lib/renderer/api/web-frame.js
```diff
@@ -1,13 +1,67 @@
...
+for (const name in binding) {
+  if (!name.startsWith('_')) { // some methods are manully populated above
+    WebFrame.prototype[name] = function (...args) {
+      return binding[name](this.context, ...args)
+    }
+  }
+}
+
+// Helper to return WebFrame or null depending on context.
+// TODO(zcbenz): Consider returning same WebFrame for the same context.
```
deepak1556: Maybe using `gin::PerContextData` would help here, which is similar to `WebContentsUserData` in behavior.

zcbenz: After thinking more about it, I realize the `blink::WebFrame` can change its v8 context when reloading/navigating, so we can not really map a v8 context to `WebFrame`.

To return same `WebFrame` for same frame, we probably need to use web elements as keys.

But no one has ever complained about it for so long, I should probably just remove this TODO.

PR #16771, opened 2019-02-06
Fix memory leak when using webFrame and spell checker (3-1-x)

zcbenz:
Backport https://github.com/electron/electron/pull/16770  to 3-1-x.


zcbenz:
The failing test is not related to this PR as it also happens on the target branch.

PR #16772, opened 2019-02-06
Fix memory leak when using webFrame and spell checker (4-0-x)

zcbenz:
Backport https://github.com/electron/electron/pull/16770 to 4-0-x.

### DIFF
--- a/atom/renderer/api/atom_api_web_frame.cc
+++ b/atom/renderer/api/atom_api_web_frame.cc
@@ -112,15 +112,15 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
   DISALLOW_COPY_AND_ASSIGN(ScriptExecutionCallback);
 };
 
-class FrameSpellChecker : public content::RenderFrameVisitor {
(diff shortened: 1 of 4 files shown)


atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -132,93 +133,131 @@ class FrameSpellChecker : public content::RenderFrameVisitor {
...
-class AtomWebFrameObserver : public content::RenderFrameObserver {
+class SpellCheckerHolder : public content::RenderFrameObserver {
  public:
-  explicit AtomWebFrameObserver(
-      content::RenderFrame* render_frame,
-      std::unique_ptr<SpellCheckClient> spell_check_client)
+  // Find existing holder for the |render_frame|.
+  static SpellCheckerHolder* FromRenderFrame(
+      content::RenderFrame* render_frame) {
+    for (auto* holder : instances_) {
```
shahzadlone: perhaps const if not modified?

zcbenz:
The failing test is not related to this PR as it also happens on the target branch.

PR #16775, opened 2019-02-06
fix: security doc now provides solid url check

### LINKED ISSUE #16774: Security doc contains a weak URL check
The security doc (https://github.com/electron/electron/blob/master/docs/tutorial/security.md#4-handle-session-permission-requests-from-remote-content) contains a weak check of URL:
```js
const { session } = require('electron')

session
  .fromPartition('some-partition')
  .setPermissionRequestHandler((webContents, permission, callback) => {
    const url = webContents.getURL()

    if (permission === 'notifications') {
      // Approves the permissions request
      callback(true)
    }
    
    if (!url.startsWith('https://my-website.com')) { // can be bypassed with 'https://my-website.com.attacker.com' 
      // Denies the permissions request
      return callback(false)
    }
  })
```

As noted in point 11 (https://github.com/electron/electron/blob/master/docs/tutorial/security.md#how-9) and point 12 (https://github.com/electron/electron/blob/master/docs/tutorial/security.md#how-10) such check can be bypassed.

Like any other security documentation this should be avoided in order to provide only solid code. I will provide a pull-request in a few minutes to add a '/' at the end of the url.

polict:
#### Description of Change
The security doc now provides a solid check, as in point 11 and 12.

#### Release Notes
Notes: fixes issue #16774

### DIFF
--- a/docs/tutorial/security.md
+++ b/docs/tutorial/security.md
@@ -278,7 +278,8 @@ session
       callback(true)
     }
 
-    if (!url.startsWith('https://my-website.com')) {
+    // Verify URL
+    if (!url.startsWith('https://my-website.com/')) {
       // Denies the permissions request
       return callback(false)
     }


docs/tutorial/security.md
```diff
@@ -278,7 +278,8 @@ session
       callback(true)
     }
 
-    if (!url.startsWith('https://my-website.com')) {
+    // Verify URL
+    if (!url.startsWith('https://my-website.com/')) {
```
shahzadlone: how does that extra ```/``` benefit?

polict: As I said in #16774, the current example (if followed by the reader) can be bypassed with an URL which starts with the required string, but may be a subdomain (or part of the domain) like `https://my-website.com.attacker.com`, which passes the check. 
Also, this follows the example provided in point 11 (https://github.com/electron/electron/blob/master/docs/tutorial/security.md#how-9). Another option would be to parse the URL and strictly check the origin like in point 12 (https://github.com/electron/electron/blob/master/docs/tutorial/security.md#how-10).

PR #16784, opened 2019-02-06
refactor: use invoke/invokeSync helpers for <webview> implementation

miniak:
#### Description of Change
- Use `ipcRendererUtils.invoke(Sync)` / `ipcMainUtils.handle` to eliminate duplicate code.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/guest-view-manager.js
+++ b/lib/browser/guest-view-manager.js
@@ -2,8 +2,8 @@
 
 const { webContents } = require('electron')
 const { ipcMainInternal } = require('@electron/internal/browser/ipc-main-internal')
+const ipcMainUtils = require('@electron/internal/browser/ipc-main-internal-utils')
 const parseFeaturesString = require('@electron/internal/common/parse-features-string')
-const errorUtils = require('@electron/internal/common/error-utils')
 const {
   syncMethods,
   asyncCallbackMethods,
@@ -341,21 +341,18 @@ const isWebViewTagEnabled = function (contents) {
 }
 
 const handleMessage = function (channel, handler) {
-  ipcMainInternal.on(channel, (event, ...args) => {
+  ipcMainUtils.handle(channel, (event, ...args) => {
     if (isWebViewTagEnabled(event.sender)) {
-      handler(event, ...args)
+      return handler(event, ...args)
     } else {
-      event.returnValue = null
+      console.error(`<webview> IPC message ${channel} sent by WebContents with <webview> disabled (${event.sender.id})`)
+      throw new Error('<webview> disabled')
     }
   })
 }
 
-handleMessage('ELECTRON_GUEST_VIEW_MANAGER_CREATE_GUEST', function (event, params, requestId) {
-  event._replyInternal(`ELECTRON_RESPONSE_${requestId}`, createGuest(event.sender, params))
-})
-
-handleMessage('ELECTRON_GUEST_VIEW_MANAGER_CREATE_GUEST_SYNC', function (event, params) {
-  event.returnValue = createGuest(event.sender, params)
+handleMessage('ELECTRON_GUEST_VIEW_MANAGER_CREATE_GUEST', function (event, params) {
+  return createGuest(event.sender, params)
 })
 
 handleMessage('ELECTRON_GUEST_VIEW_MANAGER_DESTROY_GUEST', function (event, guestInstanceId) {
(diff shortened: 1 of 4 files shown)


miniak:
@nornagon, @codebytere can you please review?

lib/browser/guest-view-manager.js
```diff
@@ -340,35 +340,35 @@ const isWebViewTagEnabled = function (contents) {
...
 
-const handleMessage = function (channel, handler) {
-  ipcMainInternal.on(channel, (event, ...args) => {
-    if (isWebViewTagEnabled(event.sender)) {
-      handler(event, ...args)
-    } else {
-      event.returnValue = null
-    }
-  })
+const assertWebViewTagEnabled = function (event) {
```
alexeykuzmin: Passing some generic `event` here looks somewhat confusing.
Can you please pass `event.sender` when you call this function?, and call the function argument, let's say, `webContents`?

lib/browser/guest-view-manager.js
```diff
@@ -340,35 +340,35 @@ const isWebViewTagEnabled = function (contents) {
...
-handleMessage('ELECTRON_GUEST_VIEW_MANAGER_CREATE_GUEST', function (event, params, requestId) {
-  event._replyInternal(`ELECTRON_RESPONSE_${requestId}`, createGuest(event.sender, params))
+ipcMainUtils.handle('ELECTRON_GUEST_VIEW_MANAGER_CREATE_GUEST', function (event, params) {
+  assertWebViewTagEnabled(event)
+  return createGuest(event.sender, params)
 })
 
-handleMessage('ELECTRON_GUEST_VIEW_MANAGER_CREATE_GUEST_SYNC', function (event, params) {
-  event.returnValue = createGuest(event.sender, params)
+ipcMainUtils.handle('ELECTRON_GUEST_VIEW_MANAGER_CREATE_GUEST_SYNC', function (event, params) {
```
alexeykuzmin:
```js
function({sender: webContents}, params) {
  <...>
}
```

It will shadow the "global" webContents we have in this file, but maybe it's not a big deal.
If it is, a different name can be used =)

lib/renderer/web-view/web-view-impl.js
```diff
@@ -265,30 +253,13 @@ const setupMethods = (WebViewElement) => {
...
+      return ipcRendererUtils.invoke('ELECTRON_GUEST_VIEW_MANAGER_ASYNC_CALL', getGuestInstanceId(this), method, args)
     }
   }
 
   for (const method of asyncPromiseMethods) {
-    WebViewElement.prototype[method] = createPromiseHandler(method)
+    WebViewElement.prototype[method] = deprecate.promisify(createPromiseHandler(method))
   }
-
-  WebViewElement.prototype.printToPDF = deprecate.promisify(WebViewElement.prototype.printToPDF)
```
alexeykuzmin: Where did it go?

lib/browser/guest-view-manager.js
miniak: I've used `contents`

lib/renderer/web-view/web-view-impl.js
miniak: it's handled by
```
  for (const method of asyncPromiseMethods) {
    WebViewElement.prototype[method] = deprecate.promisify(createPromiseHandler(method))
  }
```

miniak:
@MarshallOfSound, @felixrieseberg I would like to get this in before the affected files are converted to TypeScript to avoid conflicts. Also, the conversion will have to deal with less code :)

lib/browser/guest-view-manager.js
```diff
@@ -340,35 +340,35 @@ const isWebViewTagEnabled = function (contents) {
...
-  ipcMainInternal.on(channel, (event, ...args) => {
-    if (isWebViewTagEnabled(event.sender)) {
-      handler(event, ...args)
-    } else {
-      event.returnValue = null
-    }
-  })
+const assertWebViewTagEnabled = function (contents) {
+  if (!isWebViewTagEnabled(contents)) {
+    throw new Error('<webview> disabled')
```
nornagon: Hmm, is this code only reached when we receive a webview message from a process that shouldn't be sending any webview messages? If that's the case, we should make this message clearer, e.g.
```suggestion
    throw new Error(`<webview> IPC method ${ipc_method_name} called on WebContents with webview disabled (${webcontents_id})`)
```

Additionally, we should probably unceremoniously tear down that process because it's doing something it shouldn't be.

lib/browser/guest-view-manager.js
```diff
@@ -340,35 +340,35 @@ const isWebViewTagEnabled = function (contents) {
...
 
-handleMessage('ELECTRON_GUEST_VIEW_MANAGER_CREATE_GUEST_SYNC', function (event, params) {
-  event.returnValue = createGuest(event.sender, params)
+ipcMainUtils.handle('ELECTRON_GUEST_VIEW_MANAGER_CREATE_GUEST_SYNC', function (event, params) {
+  assertWebViewTagEnabled(event.sender)
+  return createGuest(event.sender, params)
 })
 
-handleMessage('ELECTRON_GUEST_VIEW_MANAGER_DESTROY_GUEST', function (event, guestInstanceId) {
+ipcMainInternal.on('ELECTRON_GUEST_VIEW_MANAGER_DESTROY_GUEST', function (event, guestInstanceId) {
```
nornagon: maybe a silly question, why are some of these `ipcMainUtils.handle` and others are `ipcMainInternal.on`?

lib/renderer/web-view/web-view-impl.js
```diff
@@ -246,16 +240,10 @@ const setupMethods = (WebViewElement) => {
...
-        if (error == null) {
-          if (callback) callback(result)
-        } else {
-          throw errorUtils.deserialize(error)
-        }
-      })
-      ipcRendererInternal.send('ELECTRON_GUEST_VIEW_MANAGER_ASYNC_CALL', requestId, getGuestInstanceId(this), method, args, callback != null)
+      ipcRendererUtils.invoke('ELECTRON_GUEST_VIEW_MANAGER_CALL', getGuestInstanceId(this), method, args)
+        .catch(error => {
+          throw error
```
nornagon: What's the intended behaviour on error here? because I think catch/throw is the same as not catching at all. the exception will be thrown in a subsequent microtask.

miniak: hmm, does not make sense. removed.

lib/browser/guest-view-manager.js
miniak: the exception is propagated back to the renderer. I've added a `console.error` with a more detailed message similar to the suggested one.

miniak: > Additionally, we should probably unceremoniously tear down that process because it's doing something it shouldn't be.
That I would do in a follow up PR, I just wanted to adopt the IPC helpers in this one

lib/browser/guest-view-manager.js
miniak: I've restored the `handleMessage` helper which uses `ipcMainInternal.handle` consistently

PR #16794, opened 2019-02-06
chore: add menu-will-close comment explaining setTimeout

miniak:
#### Description of Change
As discussed in https://github.com/electron/electron/pull/16607.

/cc @nornagon

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/chrome-devtools.js
+++ b/lib/browser/chrome-devtools.js
@@ -71,6 +71,9 @@ ipcMainUtils.handle('ELECTRON_INSPECTOR_CONTEXT_MENU', function (event, items, i
   const window = event.sender.getOwnerBrowserWindow()
 
   menu.once('menu-will-close', () => {
+    // menu-will-close is emitted before the click handler, but needs to be sent after.
+    // otherwise, DevToolsAPI.contextMenuCleared() would be called before
+    // DevToolsAPI.contextMenuItemSelected(id) and the menu will not work properly.
     setTimeout(() => {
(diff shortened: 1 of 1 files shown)


lib/browser/chrome-devtools.js
```diff
@@ -71,6 +71,7 @@ ipcMainUtils.handle('ELECTRON_INSPECTOR_CONTEXT_MENU', function (event, items, i
   const window = event.sender.getOwnerBrowserWindow()
 
   menu.once('menu-will-close', () => {
+    // menu-will-close is emitted before the click handler, but needs to be sent after
```
nornagon: why does it need to be sent after? what happens if it isn't?

miniak: added more details

lib/browser/chrome-devtools.js
```diff
@@ -71,6 +71,9 @@ ipcMainUtils.handle('ELECTRON_INSPECTOR_CONTEXT_MENU', function (event, items, i
   const window = event.sender.getOwnerBrowserWindow()
 
   menu.once('menu-will-close', () => {
+    // menu-will-close is emitted before the click handler, but needs to be sent after.
+    // otherwise, DevToolsAPI.contextMenuCleared() would be called before
+    // DevToolsAPI.contextMenuItemSelected(id) and the menu will not work properly.
```
nornagon: "will not work properly" is pretty vague but it'll do i guess :\

PR #16798, opened 2019-02-06
test: add test for second-instance event parameter

nornagon:
Notes: no-notes


spec/api-app-spec.js
```diff
@@ -226,6 +226,27 @@ describe('app module', () => {
...
+      const data1 = await emittedOnce(first.stdout, 'data')
+      const data2Promise = emittedOnce(first.stdout, 'data')
+
+      const secondInstanceArgs = [remote.process.execPath, appPath, '--some-switch', 'some-arg']
+      const second = ChildProcess.spawn(secondInstanceArgs[0], secondInstanceArgs.slice(1))
+      const [code2] = await emittedOnce(second, 'exit')
+      expect(code2).to.equal(1)
+      const [code1] = await firstExited
+      expect(code1).to.equal(0)
+      const data2 = (await data2Promise).toString('ascii')
```
MarshallOfSound: `ascii` vs `utf8`?  Not sure which is "correct" here?

nornagon: they're identical because the paths contain only ascii characters /shrug

spec/api-app-spec.js
```diff
@@ -226,6 +226,27 @@ describe('app module', () => {
...
       })
     })
+
+    it('passes arguments to the second-instance event', async () => {
+      const appPath = path.join(__dirname, 'fixtures', 'api', 'singleton')
+      const first = ChildProcess.spawn(remote.process.execPath, [appPath])
+      const firstExited = emittedOnce(first, 'exit')
+
+      // Wait for the first app to boot.
+      const data1 = await emittedOnce(first.stdout, 'data')
```
MarshallOfSound: This doesn't work on Windows because there is no guaruntee that "data" will contain the entire line, in fact it almost certainly won't 😢 

You'll need to buffer all `data` events and split / read the string somehow to determine when you can launch the second app and what JSON you need to parse 👍 

(Check out the windows failures)

nornagon: Ah, I was just copying this from the test above which does the same thing. I'll robustify this.

spec/package.json
```diff
@@ -24,6 +24,7 @@
     "multiparty": "^4.2.1",
     "q": "^1.5.1",
     "send": "^0.16.2",
+    "split": "^1.0.1",
```
nornagon: we already depended on this transitively, so no package-lock.json change needed

codebytere:
some conflict and this win failure:
```
not ok 909 webContents module devtools window can receive and handle menu events
  AssertionError: value should eventually become the pasted value: expected '' to equal 'test value'
      at Context.<anonymous> (C:\projects\src\electron\spec\api-web-contents-spec.js:422:22)
```

nornagon:
the win failure is unrelated (& that test was disabled with the 74 merge). i'll rebase though.

PR #16800, opened 2019-02-06
chore: dont log 100000000 chars of URL

MarshallOfSound:
Once of our tests was causing a VERRRYYY long URL to be logged to console, this made CI logs very hard to read (you had to download them and scroll past several pages of "A" before you could see the failure).

This changes that log to only log the first 2048 chars of a URL (should be plenty for 99% of use cases and most URLs are capped at this number anyway)

Notes: no-notes

### DIFF
--- a/lib/browser/navigation-controller.js
+++ b/lib/browser/navigation-controller.js
@@ -69,7 +69,7 @@ const NavigationController = (function () {
         resolve()
       }
       const rejectAndCleanup = (errorCode, errorDescription, url) => {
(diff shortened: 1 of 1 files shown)


codebytere:
cc @BinaryMuse

MarshallOfSound:
I think this will also fix the flake I've been seeing here -->

image

nornagon:
chore__dont_log_100000000_chars_of_url_by_marshallofsound_ _pull_request__16800_ _electron_electron

PR #16801, opened 2019-02-06
chore: Add new webFrame IsolatedWorldInfo API and deprecate

nitsakh:
#### Description of Change
Deprecate and add new webFrame API to support chrome 73 update.
[Chromium upgrade commit](https://github.com/electron/electron/commit/a2a88bade06dc4e18026e2a509b177f6089d0e55)

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `webFrame.setIsolatedWorldInfo` API and deprecated related APIs


docs/api/web-frame.md
```diff
@@ -148,6 +148,16 @@ Set the name of the isolated world. Useful in devtools.
 
 Set the security origin of the isolated world.
 
+### `webFrame.setIsolatedWorldInfo(worldId, info)`
```
MarshallOfSound: Please flag the old APIs as deprecated above here

codebytere:
@nitsakh this doesn't seem like it'd be a chore, would it?

atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -344,6 +344,40 @@ void WebFrame::SetIsolatedWorldHumanReadableName(int world_id,
...
 }
 
+void WebFrame::SetIsolatedWorldInfo(int world_id, mate::Arguments* args) {
+  if (args->Length() < 1) {
+    args->ThrowError("Invalid args");
+    return;
+  }
+
+  v8::Local<v8::Value> val;
+  args->GetNext(&val);
```
deepak1556: No need to extract to a local handle explicitly here, attempting something like:

```
mate::Dictionary options;
if (!args->GetNext(&options))
```

should handle the conversion.

nitsakh:
> @nitsakh this doesn't seem like it'd be a `chore`?

@codebytere `feat`?

atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -344,6 +344,40 @@ void WebFrame::SetIsolatedWorldHumanReadableName(int world_id,
...
+  options.Get("csp", &csp);
+  options.Get("name", &name);
+
+  if (!csp.empty() && origin.empty()) {
+    args->ThrowError(
+        "If csp is spicified, securityOrigin should also be specified");
+    return;
+  }
+
+  if (!origin.empty())
```
deepak1556: Not necessary to check the emptiness of the types for the following here, the blink apis will handle it.

atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -344,6 +344,37 @@ void WebFrame::SetIsolatedWorldHumanReadableName(int world_id,
...
 }
 
+void WebFrame::SetIsolatedWorldInfo(int world_id, mate::Arguments* args) {
+  if (args->Length() < 1) {
+    args->ThrowError("Invalid args");
+    return;
+  }
+
+  v8::Local<v8::Value> val;
+  args->GetNext(&val);
```
deepak1556: Can you remove this unused val. Thanks!

atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -344,6 +344,34 @@ void WebFrame::SetIsolatedWorldHumanReadableName(int world_id,
       world_id, blink::WebString::FromUTF8(name));
 }
 
+void WebFrame::SetIsolatedWorldInfo(int world_id, mate::Arguments* args) {
+  if (args->Length() < 1) {
```
MarshallOfSound: I'm pretty sure calling `GetNext` will automatically throw an error if their is no "next" arg.  Can we remove this check so that insufficient arg error messages are consistent

atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -344,6 +344,34 @@ void WebFrame::SetIsolatedWorldHumanReadableName(int world_id,
...
+    args->ThrowError("Invalid args");
+    return;
+  }
+
+  mate::Dictionary options;
+  if (!args->GetNext(&options)) {
```
MarshallOfSound: Why do we use `args` to fetch this if it is a required parameter.  See above suggestion for change to function signature

atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -344,6 +344,34 @@ void WebFrame::SetIsolatedWorldHumanReadableName(int world_id,
       world_id, blink::WebString::FromUTF8(name));
 }
 
+void WebFrame::SetIsolatedWorldInfo(int world_id, mate::Arguments* args) {
```
MarshallOfSound:
```suggestion
void WebFrame::SetIsolatedWorldInfo(int world_id, mate::Dictionary options, mate::Arguments* args) {
```

atom/renderer/api/atom_api_web_frame.cc
MarshallOfSound: We can definitely remove this if you do the suggested change above

nitsakh: Done! Thanks! 🙇

PR #16820, opened 2019-02-07
refactor: convert more files to typescript

MarshallOfSound:
I'm doing this work in small chunks to make review easier.  This is another chunk of TS conversion

Notes: no-notes

This PR also adds the `tsify` dependency to allow TS transformations for the `browserify`'ed code for sandbox and isolated


lib/browser/init.ts
```diff
@@ -49,10 +49,12 @@ process.on('uncaughtException', function (error) {
   }
 
   // Show error in GUI.
-  const dialog = require('electron').dialog
-  const stack = error.stack ? error.stack : `${error.name}: ${error.message}`
-  const message = 'Uncaught Exception:\n' + stack
-  dialog.showErrorBox('A JavaScript error occurred in the main process', message)
+  import('electron')
```
nornagon: why not a static import?

lib/browser/init.ts
```diff
@@ -192,10 +194,13 @@ app.on('window-all-closed', () => {
   }
 })
 
-const { setDefaultApplicationMenu } = require('@electron/internal/browser/default-menu')
-
-// Create default menu.
-app.once('ready', setDefaultApplicationMenu)
+Promise.all([
+  import('@electron/internal/browser/default-menu'),
```
nornagon: here too

lib/common/init.ts
```diff
@@ -1,29 +1,35 @@
...
       return func.apply(this, arguments)
-    }
+    } as any
   })
 }
 
-function wrap (func, wrapper) {
+/**
+ * Casts to any below for func are due to Typescript not supporting symbols
+ * in index signitures
```
nornagon:
```suggestion
 * in index signatures
```

tsconfig.json
```diff
@@ -1,23 +1,23 @@
...
+    "experimentalDecorators": false,
+    "strict": true,
+    "baseUrl": ".",
+    "allowJs": true,
+    "outDir": "ts-gen",
+    "paths": {
+      "@electron/internal/*": ["lib/*"]
+    }
+  },
+  "exclude": [
```
nornagon: can this be `include` instead of `exclude`?

lib/browser/init.ts
MarshallOfSound: You can't `import 'electron'` half way down a file, static imports have to be at the top of a file

lib/browser/init.ts
MarshallOfSound: Same as above, and the async nature here actually works well with `app.whenReady`

lib/browser/init.ts
nornagon: is there any particular reason not to just import this at the top of the file then?

tsconfig.json
MarshallOfSound: Not easily, we have to explicitly `exclude` this file and this file only

MarshallOfSound: Otherwise TS get's multiple-type-declaration errors with the TS defs GN passes in

lib/browser/init.ts
MarshallOfSound: We can't 😢 The `electron` alias is configured / setup half-way down this file in `reset-search-paths`

nornagon: ah, okay. Let's comment here about that then!

lib/common/init.ts
```diff
@@ -1,29 +1,35 @@
...
 // called.
-const wrapWithActivateUvLoop = function (func) {
+const wrapWithActivateUvLoop = function <T extends Function> (func: T): T {
   return wrap(func, function (func) {
-    return function () {
+    return function (this: any) {
       process.activateUvLoop()
       return func.apply(this, arguments)
-    }
+    } as any
```
felixrieseberg: Can we do better than `any`? It seems like it'll at least be `(...args: Array<any>) => any`.

CvX:
This PR (or more specifically https://github.com/electron/electron/pull/16820/files#diff-27aca9003e7bfe3e2c635e67d04fc917L17) should probably be back-ported to 5-0-x. See: https://github.com/electron/electron/issues/17647

Issue #16812, opened 2019-02-07
Expose cipher DES-EDE-CBC in Electron 4 crypto module


lonk:
* **Electron Version** (output of `node_modules/.bin/electron --version`):
  * v4.0.0
* **Operating System** (Platform and Version):
  * archlinux
* **Last known working Electron version** (if applicable):
  * v3.1.3

Hi !

I just upgraded Electron from version 3 to version 4 and I noticed that the cipher DES-EDE-CBC is no longer exposed in the crypto module, which is breaking my app.

Is it possible to expose it again, as mentioned by @nornagon in #16195 ?

Thank you !

nornagon:
It looks to me like `DES_EDE_CBC` is already exposed, see e.g. https://gist.github.com/e6db0acbf6b736c49c8bab4d86fff0dc which reports the following list in `crypto.getCiphers()` as of Electron 4.0.4:

```["aes-128-cbc","aes-128-cfb","aes-128-ctr","aes-128-ecb","aes-128-ofb","aes-256-cbc","aes-256-cfb","aes-256-ctr","aes-256-ecb","aes-256-ofb","aes-256-xts","des-cbc","des-ecb","des-ede","des-ede-cbc","des-ede3-cbc","rc2-cbc","rc4"]```

nornagon:
Ah, actually it looks like it's listed in `getCiphers` but doesn't work when you try to actually use it. Reopening.

patwhite:
Are these changes going to be available in the 4 branch, or just 5? I'm on 4.0.8 and still don't see RipeMD-160.

PR #16822, opened 2019-02-07
fix: backport patch to sync exposed crypto

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/16812.

Backports https://boringssl-review.googlesource.com/c/boringssl/+/33984 to our patch set in order to correctly enumerate and instantiate ciphers. Also adds a test to ensure that `DES-EDE-CBC` actually works.

cc @nornagon 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added a patch to fix incorrect enumeration and instantiation of Node.js ciphers in the Crypto module.

### DIFF
--- a/spec/node-spec.js
+++ b/spec/node-spec.js
@@ -462,6 +462,14 @@ describe('node feature', () => {
     it('should be able to create an aes-256-cfb cipher', () => {
       require('crypto').createCipheriv('aes-256-cfb', '0123456789abcdef0123456789abcdef', '0123456789abcdef')
     })
+
+    it('should list des-ede-cbc in getCiphers', () => {
+      expect(require('crypto').getCiphers()).to.include('des-ede-cbc')
+    })
+
+    it('should be able to create an des-ede-cbc cipher', () => {
+      require('crypto').createCipheriv('des-ede-cbc', '0123456789abcdeff1e0d3c2b5a49786', 'fedcba9876543210')
+    })
   })
 
   it('includes the electron version in process.versions', () => {
--- a/spec/node-spec.js
+++ b/spec/node-spec.js
@@ -468,7 +468,9 @@ describe('node feature', () => {
     })
 
     it('should be able to create an des-ede-cbc cipher', () => {
-      require('crypto').createCipheriv('des-ede-cbc', '0123456789abcdeff1e0d3c2b5a49786', 'fedcba9876543210')
+      const key = Buffer.from('0123456789abcdeff1e0d3c2b5a49786', 'hex')
+      const iv = Buffer.from('fedcba9876543210', 'hex')
+      require('crypto').createCipheriv('des-ede-cbc', key, iv)
     })
   })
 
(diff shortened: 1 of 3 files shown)


nornagon:
FYI, that patch may change a little still before it gets merged upstream so we may need to update this.

spec/node-spec.js
```diff
@@ -462,6 +462,14 @@ describe('node feature', () => {
     it('should be able to create an aes-256-cfb cipher', () => {
       require('crypto').createCipheriv('aes-256-cfb', '0123456789abcdef0123456789abcdef', '0123456789abcdef')
     })
+
+    it('should list des-ede-cbc in getCiphers', () => {
+      expect(require('crypto').getCiphers()).to.include('des-ede-cbc')
+    })
+
+    it('should be able to create an des-ede-cbc cipher', () => {
+      require('crypto').createCipheriv('des-ede-cbc', '0123456789abcdeff1e0d3c2b5a49786', 'fedcba9876543210')
```
nornagon: I think the IV needs to be a different length here. try 8 or 32?
```suggestion
      require('crypto').createCipheriv('des-ede-cbc', '0123456789abcdeff1e0d3c2b5a49786', 'fedcba98')
```

codebytere: hmm: i pulled that key and IV from Chromium's own tests [here](https://cs.chromium.org/chromium/src/third_party/boringssl/src/crypto/cipher_extra/test/cipher_tests.txt?q=des-ede-cbc&sq=package:chromium&dr=C&l=94)

nornagon: huh, that is odd then. it'd be worth tracking down what exactly is going on & why node things this is the wrong IV length!

codebytere: @ryzokuken any ideas?

nornagon: I'm not really sure what Chromium's tests are doing there, but I'm looking at [the definition of `EVP_des_ede_cbc`](https://github.com/google/boringssl/blob/15ba2d11a9873cb40b2a45bba18242fdb16ca4c6/crypto/fipsmodule/cipher/e_des.c#L183-L193) and it looks like the IV and keys should be half the length this test sets them to:

```cpp
DEFINE_METHOD_FUNCTION(EVP_CIPHER, EVP_des_ede_cbc) {
  // ...
  out->key_len = 16;
  out->iv_len = 8;
  // ...
}
```

nornagon: oh, you know what it is? Chrome's tests encode the bytes as hex, which are 2 chars per byte.

nornagon:
```suggestion
      require('crypto').createCipheriv('des-ede-cbc', Buffer.from('0123456789abcdeff1e0d3c2b5a49786', 'hex'), Buffer.from('fedcba9876543210', 'hex'))
```

codebytere:
@nornagon @ryzokuken all set, ptal when you can 🙇‍♀️

codebytere:
Flake on linux tests that is unrelated! Merging.

PR #16832, opened 2019-02-08
fix: display empty menu item for non-visible submenus

codebytere:
#### Description of Change

Previously, when a context menu was created with a submenu that had no visible `MenuItems`, the menu item would display and not show anything on hover, which i don't believe to be the most user-friendly behavior:

screen shot 2019-02-08 at 9 42 46 am

This PR improves the UX of that scenario by first checking if a submenu has any visible `MenuItems`, and in the case that it doesn't, it creates a localized `MenuItem` to indicate that the submenu does not currently have any items with which the user may interact:

screen shot 2019-02-08 at 9 46 37 am

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed issue whereby a user was not well informed when interacting with a menu submenu that did not have any visible `MenuItems`.

### DIFF
--- a/atom/browser/ui/cocoa/atom_menu_controller.mm
+++ b/atom/browser/ui/cocoa/atom_menu_controller.mm
@@ -58,6 +58,30 @@ Role kRolesMap[] = {
     {@selector(clearRecentDocuments:), "clearrecentdocuments"},
 };
 
+// Called when adding a submenu to the menu and checks if the submenu, via its
+// |model|, has visible child items.
+bool MenuHasVisibleItems(const atom::AtomMenuModel* model) {
+  int count = model->GetItemCount();
+  for (int index = 0; index < count; index++) {
+    if (model->IsVisibleAt(index))
+      return true;
+  }
+  return false;
+}
+
+// Called when an empty submenu is created. This inserts a menu item labeled
+// "(empty)" into the submenu. Matches Windows behavior.
+NSMenu* MakeEmptySubmenu() {
+  base::scoped_nsobject<NSMenu> submenu([[NSMenu alloc] initWithTitle:@""]);
+  base::string16 empty_menu_title = base::ASCIIToUTF16("(empty)");
+  NSString* empty_menu_label =
+      l10n_util::FixUpWindowsStyleLabel(empty_menu_title);
+
+  [submenu addItemWithTitle:empty_menu_label action:NULL keyEquivalent:@""];
+  [[submenu itemAtIndex:0] setEnabled:NO];
+  return submenu.autorelease();
+}
+
 }  // namespace
 
 // Menu item is located for ease of removing it from the parent owner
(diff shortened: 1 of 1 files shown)


atom/browser/ui/cocoa/atom_menu_controller.mm
```diff
@@ -58,6 +58,30 @@
...
+      return true;
+  }
+  return false;
+}
+
+// Called when an empty submenu is created. This inserts a menu item labeled
+// "(empty)" into the submenu. Matches Windows behavior.
+NSMenu* MakeEmptySubmenu() {
+  base::scoped_nsobject<NSMenu> submenu([[NSMenu alloc] initWithTitle:@""]);
+  base::string16 empty_menu_title = base::ASCIIToUTF16("(empty)");
```
MarshallOfSound: In the past we've steered clear of hard-coded English strings.  Is there a translation in Chrome somewhere for this string we can leverage?

codebytere: @MarshallOfSound yes, native Windows/Linux apps use the same "empty" UX for submenus with no visible Menu Items!

And potentially, Chrome itself uses an int we can't leverage so i copied our approach to the Services submenu localization hardcode for this case

MarshallOfSound: Would be cool if we could use `IDS_APP_MENU_EMPTY_SUBMENU`, not sure if we get access to that in our build although we do use other `IDS_` strings in our code 🤔

codebytere: success!

sindresorhus:
I think it would be better to disable the menu if its submenu has no visible items. Even if you disagree, the menu label should at least be `(empty)` => `(Empty)`.

codebytere:
@sindresorhus this mirrors behavior on macOS itself, and was directly lifted from Chrome, capitalization included: 

screen shot 2019-02-20 at 11 38 35 am

sindresorhus:
> this mirrors behavior on macOS itself

I have never seen this behavior outside of Chrome. Example?

> and was directly lifted from Chrome, capitalization included:

Chrome doesn't have a good track record at following native UI conventions.

nornagon:
This is not the native behaviour of macOS, e.g:

screenshot of iTerm 2 with empty submenu, showing no (empty) item

I'm somewhat ambivalent as to whether we should show the _(empty)_ item anyway. On the one hand, it does seem like a better UX. On the other hand, it makes it impossible to match native behaviour.

sindresorhus:
I wouldn't use iTerm as an example of good macOS UX.

A better example:
Go to Finder, open the "Go" menu, then "Recent Folders", click "Clear Menu", then open "Go" again, and you'll see that "Recent Folders" is disabled. So Apple is doing what I argued for in https://github.com/electron/electron/pull/16832#issuecomment-465504770.

screen shot 2019-02-27 at 16 34 09

MarshallOfSound:
Even Apple aren't consistent with their UX for this (and a lot of other things).  For instance the Safari favorites menu.

image

Without a strong reference from Apple Human Interface Guidelines this is just going to be an opinionated choice, especially if Apple themselves aren't consistent here 😄

sindresorhus:
@MarshallOfSound That seems like a bug. Notice how the title is also not titleized and the title should have been something like "No Favorites". I'll report it later. Most likely, no one at Apple noticed, as it's very unlikely not to have a single Favorite.

Another example, from Safari (History menu), of the menu being disabled when there are no items:

screen shot 2019-02-28 at 02 11 07

sindresorhus:
Although, users should prefer adding a descriptive menu item instead:

screen shot 2019-02-28 at 02 26 23
screen shot 2019-02-28 at 02 26 31

Which, IMHO, the Electron docs should recommend.

I think the best default behavior is to disable the owner of the empty submenu.

PR #16839, opened 2019-02-08
chore: disable get/setLoginItemSettings specs on MAS

MarshallOfSound:
These tests have all kinds of asynchronous issues on MAS CI.  We disabled them on master here --> https://github.com/electron/electron/pull/16552/files#diff-6428a7ecb6a3a2831eae23aee7efeac5R539

To reduce CI "flakes" this test should be disabled all the way back to 2-0-x 👍 

Notes: no-notes

Refs: #16838

### DIFF
--- a/spec/api-app-spec.js
+++ b/spec/api-app-spec.js
@@ -440,7 +440,7 @@ describe('app module', () => {
     ]
 
     before(function () {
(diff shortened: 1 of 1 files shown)


codebytere:
(pending above change request)

Issue #16796, opened 2019-02-06
[win] flaky tests on Windows, potential crash-on-exit


nornagon:
```text
not ok 49 app module app launch through uri launches successfully for multiple URIs following --
  AssertionError: expected 3221226505 to equal 123
      at ChildProcess.first.once.code (C:\projects\src\electron\spec\api-app-spec.js:822:25)
      at Object.onceWrapper (events.js:276:13)
      at ChildProcess.emit (events.js:188:13)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:254:12)
```

https://windows-ci.electronjs.org/project/AppVeyor/electron-hew04/build/1.0.4231?fullLog=true

MarshallOfSound:
`3221226505` normally means that the app is crashing, we've seen this caused before by a crash-on-exit

MarshallOfSound:
Dropping this flake on this issue as I believe it's the same root cause: [AppVeyor Build](https://windows-ci.electronjs.org/project/AppVeyor/electron-67iac/build/1.0.5662?fullLog=true)

```
not ok 14 app module app.requestSingleInstanceLock prevents the second launch of app
  AssertionError: expected 3221225477 to equal 1
      at ChildProcess.second.once.code (C:\projects\src\electron\spec\api-app-spec.js:224:27)
      at Object.onceWrapper (events.js:276:13)
      at ChildProcess.emit (events.js:188:13)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:254:12)
```

I believe we have a crash-on-exit in Windows

MarshallOfSound:
Another example: https://windows-ci.electronjs.org/project/AppVeyor/electron-67iac/build/1.0.5730?fullLog=true

Gvozd:
Same error
1. update electron to 5.0.0-beta.2 in electronquick-start
2. run
3. try open devtools - crash

MarshallOfSound:
@Gvozd I'd be surprised if you saw the same error as this issue, as these logs come from our test suite (that electron-quick-start doesn't run).  If you are experiencing a reproducible crash on Electron 5.0.0-beta.2 can you please raise a new issue following the issue template and provide a way to reproduce the crash 👍

Gvozd:
@MarshallOfSound , sorry I made a little mistake
In my case exit code is 3221225477, not 3221226505
Also js not worked in renderer process
#16883 #16885

PR #16840, opened 2019-02-08
chore: make crash-reporter specs not use URL module

MarshallOfSound:
These specs fail 100% locally (no idea how they sometimes pass on CI) because the `url` module is not available to sandboxed renderers.

This updates the specs to use standard web APIs to do URl parsing instead of `url`.

These tests are responsible for almost 30% of the flakes in the last 30 days 👍 

Notes: no-notes

### DIFF
--- a/spec/fixtures/api/crash-restart.html
+++ b/spec/fixtures/api/crash-restart.html
@@ -2,7 +2,7 @@
 <body>
 <script type="text/javascript" charset="utf-8">
 
-const {port} = require('url').parse(window.location.href, true).query
+const port = (new URLSearchParams(location.search)).get('port')
 const {crashReporter, ipcRenderer} = require('electron')
 
 crashReporter.start({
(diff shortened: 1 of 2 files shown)


MarshallOfSound:
Windows failure is flake, reported in #16796

miniak:
@MarshallOfSound the url module from browserify is actually available in sandboxed renderers 
https://github.com/electron/electron/blob/1898f9162073910c05958295c612deec6121a892/lib/sandboxed_renderer/init.js#L44. But of course it’s better to use standard web api instead.

MarshallOfSound:
@miniak it wasn't available though, that was causing the errors (unable to find module "url").  Revert this change and try the tests locally :+1:

miniak:
@MarshallOfSound I remember the tests failing locally. I definitely appreciate your fix.

Issue #16747, opened 2019-02-05
Setting a MenuItem to visible: false also disables it


aaronraimist:
* **Electron Version** (output of `node_modules/.bin/electron --version`):
  * 4.0.4
* **Operating System** (Platform and Version):
  * macOS 10.14.3
* **Last known working Electron version** (if applicable):
  * unknown

### Expected Behavior
Setting a MenuItem to `visible: false` will hide it from the menu but **not** disable it.

See documentation https://electronjs.org/docs/api/menu-item

### Actual behavior
MenuItem is hidden AND disabled

### To Reproduce
https://github.com/aaronraimist/electron-quick-start/tree/visible-false

There should be one menu item visible (zoomin) and zoom out should also work but it doesn't.
(my issue is with zoomin but since that is kind of broken I did the demo with zoomout)

Menu code looks like
```
const template = [
  {
    label: 'View',
    submenu: [
	  { role: 'zoomin' },
      { role: 'zoomout', visible: false },
    ],
  }
]
```

```sh
# Clone this repository
git clone https://github.com/electron/electron-quick-start
# Go into the repository
cd electron-quick-start
# Checkout branch
git checkout visible-false
# Install dependencies
npm install
# Run the app
npm start
```

### Additional Information

I'm trying to work around https://github.com/electron/electron/issues/15496 by creating two menu items, one visible and one invisible but it doesn't work. See https://github.com/vector-im/riot-web/pull/8381

codebytere:
@aaronraimist this is intended behavior, if a MenuItem is invisible it currently will not be functional. I can potentially look into changing this behavior, however, if you'd prefer!

aaronraimist:
@codebytere Thanks for your answer. I can't speak to other platforms but at least in native macOS apps I would expect to be able to use a hidden menu item, or at least have the option to use hidden menu items.

Xcode gives you the option to allow the keyboard shortcut to continue working when the menu item is hidden
![screen shot 2019-02-08 at 12 43 40 pm](https://user-images.githubusercontent.com/5855073/52498723-2ad66980-2b9f-11e9-94c1-15e8d64e77e1.png)

codebytere:
@aaronraimist i'll see what i can do!

PR #16853, opened 2019-02-08
feat: allow MenuItems to work optionally when hidden

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/16747.

This PR enables MenuItems on macOS to work optionally when `visible: false`. using a new `MenuItem` option `acceleratorWorksWhenHidden`. The default of this new options is true, to match behavior on Windows and Linux, but since native macOS development allows for this to be turned off, this option allows that to be done.

Example:
```js
  const menu = Menu.buildFromTemplate([
    {
      label: 'View',
      submenu: [
        { role: 'zoomin' },
        { 
          role: 'zoomout', 
          visible: false,
          acceleratorWorksWhenHidden: true
        },
      ],
    }
  ])
```

cc @MarshallOfSound @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added an option to enable `MenuItems` on macOS to work optionally when `visible: false`.

### DIFF
--- a/atom/browser/api/atom_api_menu.cc
+++ b/atom/browser/api/atom_api_menu.cc
@@ -40,6 +40,7 @@ void Menu::AfterInit(v8::Isolate* isolate) {
   delegate.Get("isCommandIdChecked", &is_checked_);
   delegate.Get("isCommandIdEnabled", &is_enabled_);
   delegate.Get("isCommandIdVisible", &is_visible_);
+  delegate.Get("shouldCommandIdWorkWhenHidden", &works_when_hidden_);
   delegate.Get("getAcceleratorForCommandId", &get_accelerator_);
   delegate.Get("shouldRegisterAcceleratorForCommandId",
                &should_register_accelerator_);
@@ -63,6 +64,12 @@ bool Menu::IsCommandIdVisible(int command_id) const {
   v8::Locker locker(isolate());
   v8::HandleScope handle_scope(isolate());
   return is_visible_.Run(GetWrapper(), command_id);
+}
+
+bool Menu::ShouldCommandIdWorkWhenHidden(int command_id) const {
+  v8::Locker locker(isolate());
+  v8::HandleScope handle_scope(isolate());
+  return works_when_hidden_.Run(GetWrapper(), command_id);
 }
 
 bool Menu::GetAcceleratorForCommandIdWithParams(
@@ -181,6 +188,10 @@ bool Menu::IsVisibleAt(int index) const {
   return model_->IsVisibleAt(index);
 }
 
+bool Menu::WorksWhenHiddenAt(int index) const {
+  return model_->WorksWhenHiddenAt(index);
+}
+
 void Menu::OnMenuWillClose() {
   Emit("menu-will-close");
 }
@@ -212,6 +223,7 @@ void Menu::BuildPrototype(v8::Isolate* isolate,
       .SetMethod("getAcceleratorTextAt", &Menu::GetAcceleratorTextAt)
       .SetMethod("isItemCheckedAt", &Menu::IsItemCheckedAt)
       .SetMethod("isEnabledAt", &Menu::IsEnabledAt)
+      .SetMethod("worksWhenHiddenAt", &Menu::WorksWhenHiddenAt)
       .SetMethod("isVisibleAt", &Menu::IsVisibleAt)
       .SetMethod("popupAt", &Menu::PopupAt)
       .SetMethod("closePopupAt", &Menu::ClosePopupAt);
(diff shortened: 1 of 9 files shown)


atom/browser/ui/cocoa/atom_menu_controller.mm
```diff
@@ -302,8 +307,7 @@ - (void)addItemToMenu:(NSMenu*)menu
 }
 
 // Called before the menu is to be displayed to update the state (enabled,
-// radio, etc) of each item in the menu. Also will update the title if
```
codebytere: comment removed as it's a remnant of dead code previously removed in https://github.com/electron/electron/pull/14939

atom/browser/api/atom_api_menu.cc
```diff
@@ -40,6 +40,7 @@ void Menu::AfterInit(v8::Isolate* isolate) {
   delegate.Get("isCommandIdChecked", &is_checked_);
   delegate.Get("isCommandIdEnabled", &is_enabled_);
   delegate.Get("isCommandIdVisible", &is_visible_);
+  delegate.Get("isCommandIdWorksWhenHidden", &is_works_when_hidden_);
```
MarshallOfSound: /nit This is like english semantics but can this be `doesCommandIdWorkWhenHidden` so that it reads as a question like the other methods above

MarshallOfSound: Same throughout the PR if possible, not too set on it just an IMO thing

codebytere: done

atom/browser/api/atom_api_menu.cc
```diff
@@ -181,6 +188,10 @@ bool Menu::IsVisibleAt(int index) const {
   return model_->IsVisibleAt(index);
 }
 
+bool Menu::IsWorksWhenHiddenAt(int index) const {
```
ckerr: _(minor)_ `IsWorks...` Is an odd name. Would `worksWhen...` be better?

docs/api/menu-item.md
```diff
@@ -48,6 +49,8 @@ See [`Menu`](menu.md) for examples.
     the placement of their containing group after the containing group of the item
     with the specified label.
 
+**Note:** `acceleratorWorksWhenHidden` is specified as being macOS-only becuase the behavior it enables when `true` is the default on Windows and Linux.
```
ckerr: Tyop: becuase

More importantly I don't understand the explanation. If it's useful for app devs to be able to choose to enable or disable this, the platform default doesn't matter?

Or, if the description is saying this is always useful on Win and Linux... would it always be useful on Mac too -- and should always be 'on' and not need a user-visible Mac setting?

codebytere: This is something that is _always_ on for windows and linux; and on Mac it's something that, in native development, is always a choice by the developer. Thus, we now default it to true, to match behavior on other platforms, but allow developers to turn it off for Mac if they wish, since they would be able to do this when developing native Mac apps

(i'll clean up the property description, bc as I typed this i realized my original was not at all clear 😆)

ckerr: So there is no native mechanism for making this choice on Windows?

I'm not understanding why we want to let app devs choose the setting value on Platform A but not let them choose on Platforms B and C.

codebytere: @ckerr primarily, to most closely match what they could do if they were developing a native application for macOS.

docs/api/menu-item.md
```diff
@@ -24,6 +24,7 @@ See [`Menu`](menu.md) for examples.
   * `icon` ([NativeImage](native-image.md) | String) (optional)
   * `enabled` Boolean (optional) - If false, the menu item will be greyed out and
     unclickable.
+  * `acceleratorWorksWhenHidden` Boolean (optional) - default is `true`, and when `false` will prevent the accelerator from triggering the item if the item is not visible`. _macOS_
```
zcbenz: Should probably note it only works after 10.13.

MarshallOfSound:
Looks like you're gonna need one of those lovely Forward Declarations 😢 

```
../../electron/atom/browser/ui/cocoa/atom_menu_controller.mm:296:11: error: instance method '-setAllowsKeyEquivalentWhenHidden:' not found (return type defaults to 'id') [-Werror,-Wobjc-method-access]
          setAllowsKeyEquivalentWhenHidden:(model->WorksWhenHiddenAt(index))];
```

codebytere:
@MarshallOfSound oops yeah i was planning on fixing this yesterday but then gclient died on me; fixing soon :)

PR #16866, opened 2019-02-11
chore: disable the remote module in devtools / chrome extension background scripts

miniak:
#### Description of Change
- Disables the remote module in devtools / chrome extension background scripts
- Reverts https://github.com/electron/electron/pull/16565 as it's no longer necessary (remote related IPC are blocked by the `isRemoteModuleEnabled` check)

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/atom/browser/api/atom_api_web_contents.cc
+++ b/atom/browser/api/atom_api_web_contents.cc
@@ -1959,6 +1959,9 @@ v8::Local<v8::Value> WebContents::GetLastWebPreferences(
 }
 
 bool WebContents::IsRemoteModuleEnabled() const {
+  if (web_contents()->GetVisibleURL().SchemeIs("chrome-devtools")) {
+    return false;
+  }
   if (auto* web_preferences = WebContentsPreferences::From(web_contents())) {
     return web_preferences->IsRemoteModuleEnabled();
   }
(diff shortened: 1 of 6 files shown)


miniak:
@MarshallOfSound, @nornagon can you please review?

atom/browser/api/atom_api_web_contents.cc
```diff
@@ -1959,6 +1959,9 @@ v8::Local<v8::Value> WebContents::GetLastWebPreferences(
 }
 
 bool WebContents::IsRemoteModuleEnabled() const {
+  if (web_contents()->GetVisibleURL().SchemeIs("chrome-devtools")) {
```
MarshallOfSound: This disables it for `chrome-devtools` but the original code marked `chrome-extension:` as a safe protocol as well.  Should that be referenced here somewhere?

atom/browser/atom_browser_client.cc
```diff
@@ -505,6 +505,7 @@ void AtomBrowserClient::AppendExtraCommandLineSwitches(
     if (web_contents->GetVisibleURL().SchemeIs("chrome-devtools")) {
```
MarshallOfSound: Same here

atom/browser/api/atom_api_web_contents.cc
miniak: I am disabling the remote module for the background script host process here
https://github.com/electron/electron/pull/16866/files#diff-8d85c651cb0fb8feef2c4694d52da2e3R95
that one is loading the `chrome-extensions:` URL

miniak: we only have handle devtools explicitly, as those don't have webPreferences

atom/browser/atom_browser_client.cc
miniak: the same explanation as above

atom/browser/api/atom_api_web_contents.cc
MarshallOfSound: 👍

PR #16888, opened 2019-02-12
chore: suppress output of npm_action unless it fails

MarshallOfSound:
Notes: no-notes

### DIFF
--- a/build/npm-run.py
+++ b/build/npm-run.py
@@ -1,10 +1,18 @@
 #!/usr/bin/env python
 import os
+import subprocess
 import sys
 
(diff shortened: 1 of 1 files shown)


build/npm-run.py
```diff
@@ -1,10 +1,18 @@
 #!/usr/bin/env python
 import os
+import subprocess
 import sys
 
 SOURCE_ROOT = os.path.dirname(os.path.dirname(__file__))
-args = ["npm", "run",
+cmd = "npm"
+if sys.platform == "win32":
+    executable += ".cmd"
```
jkleinsc: `executable` is undefined.  Did you mean:
```suggestion
    cmd += ".cmd"
```

PR #16894, opened 2019-02-12
refactor: Port window-setup to TS

felixrieseberg:
Another chunk of TypeScript conversion, this time around for `window-setup`.

Notes: no-notes


lib/renderer/window-setup.ts
```diff
@@ -0,0 +1,281 @@
...
+  public postMessage (message: any, targetOrigin: any) {
+    this.ipcRenderer.send('ELECTRON_GUEST_WINDOW_MANAGER_WINDOW_POSTMESSAGE', this.guestId, message, toString(targetOrigin), window.location.origin)
+  }
+
+  public eval (...args: any[]) {
+    this.ipcRenderer.send('ELECTRON_GUEST_WINDOW_MANAGER_WEB_CONTENTS_METHOD', this.guestId, 'executeJavaScript', ...args)
+  }
+}
+
+const windowSetup = (
```
MarshallOfSound: Can we `export const` instead of `export default` further down the file.  It's an IMO thing but I much prefer enforcing naming conventions and when we still have `require` the `default` export is gonna look really weird

lib/renderer/window-setup.ts
```diff
@@ -0,0 +1,281 @@
...
+  if (guestInstanceId == null) {
+    // Override default window.close.
+    window.close = function () {
+      ipcRenderer.sendSync('ELECTRON_BROWSER_WINDOW_CLOSE')
+    }
+  }
+
+  if (!usesNativeWindowOpen) {
+    // Make the browser window or guest view emit "new-window" event.
+    (window as any).open = function (url?: string, frameName?: string, features?: string) {
```
MarshallOfSound: Is this TS not letting us override a `readonly` or us not matching the `window.open` function signature?

lib/renderer/window-setup.ts
```diff
@@ -0,0 +1,281 @@
...
+    }
+  }
+
+  // But we do not support prompt().
+  window.prompt = function () {
+    throw new Error('prompt() is and will not be supported.')
+  }
+
+  ipcRenderer.on('ELECTRON_GUEST_WINDOW_POSTMESSAGE', function (
+    _event: Event, sourceId: number, message: any, sourceOrigin: string
```
MarshallOfSound: `Electron.Event`

lib/renderer/window-setup.ts
```diff
@@ -0,0 +1,281 @@
...
+    // the actual <webview> element visibility) for backwards compatibility.
+    // See discussion in #9178.
+    //
+    // Note that this results in duplicate visibilitychange events (since
+    // Chromium also fires them) and potentially incorrect visibility change.
+    // We should reconsider this decision for Electron 2.0.
+    let cachedVisibilityState = isHiddenPage ? 'hidden' : 'visible'
+
+    // Subscribe to visibilityState changes.
+    ipcRenderer.on('ELECTRON_GUEST_INSTANCE_VISIBILITY_CHANGE', function (_event: Event, visibilityState: VisibilityState) {
```
MarshallOfSound: `Electron.Event`

lib/renderer/window-setup.ts
```diff
@@ -0,0 +1,281 @@
...
+    // set event.source.
+    const messageEvent = new MessageEvent('message', {
+      data: message,
+      origin: sourceOrigin,
+      source: getOrCreateProxy(ipcRenderer, sourceId) as any as Window,
+      bubbles: false,
+      cancelable: false
+    })
+
+    window.dispatchEvent(messageEvent as MessageEvent)
```
MarshallOfSound: This `as` cast seems strange, it's `new MessageEvent`, shouldn't it know that it is indeed a `MessageEvent`

lib/renderer/window-setup.ts
felixrieseberg: I agree with that, I'll change it.

lib/renderer/window-setup.ts
felixrieseberg: The first!

MarshallOfSound:
Tests are failing with

> `TypeError: Failed to construct 'MessageEvent': member source is not of type EventTarget.`

PR #16911, opened 2019-02-12
chore: 'breaking change' in PR body overrides commit type in relnotes

nornagon:
#### Description of Change
This allows the line 'BREAKING CHANGE' in the PR body to override the commit type inferred from the semantic commit message. The reason for this is to allow post-editing the PR body to add 'BREAKING CHANGE' into commits that didn't have that tag to begin with.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/release-notes/notes.js
+++ b/script/release-notes/notes.js
@@ -522,9 +522,10 @@ const getNotes = async (fromRef, toRef, newVersion) => {
 
       // try to pull a release note from the pull comment
       const prParsed = parseCommitMessage(`${prData.data.title}\n\n${prData.data.body}`, pr.owner, pr.repo)
-      commit.note = commit.note || prParsed.note
-      commit.type = commit.type || prParsed.type
-      if (prParsed.type === 'breaking-change') {
+      if (!commit.note) {
+        commit.note = prParsed.note
+      }
+      if (!commit.type || prParsed.type === 'breaking-change') {
         commit.type = prParsed.type
       }
       prSubject = prSubject || prParsed.subject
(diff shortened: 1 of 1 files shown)


script/release-notes/notes.js
```diff
@@ -521,10 +521,12 @@ const getNotes = async (fromRef, toRef, newVersion) => {
...
 
       // try to pull a release note from the pull comment
-      const prParsed = {}
-      parseCommitMessage(`${prData.data.title}\n\n${prData.data.body}`, pr.owner, pr.repo, prParsed)
+      const prParsed = parseCommitMessage(`${prData.data.title}\n\n${prData.data.body}`, pr.owner, pr.repo)
       commit.note = commit.note || prParsed.note
       commit.type = commit.type || prParsed.type
+      if (prParsed.type === 'breaking-change') {
+        commit.type = prParsed.type
+      }
```
ckerr: I the flow of "when does `commit.type` get set" is a little hard to read with this and 526 side-by-side.

What we joined them this way:

```js
if (!commit.type || prParsed.type === 'breaking-change') {
  commit.type= prParsed.type
} 
```

script/release-notes/notes.js
```diff
@@ -521,10 +521,12 @@ const getNotes = async (fromRef, toRef, newVersion) => {
       }
 
       // try to pull a release note from the pull comment
-      const prParsed = {}
-      parseCommitMessage(`${prData.data.title}\n\n${prData.data.body}`, pr.owner, pr.repo, prParsed)
+      const prParsed = parseCommitMessage(`${prData.data.title}\n\n${prData.data.body}`, pr.owner, pr.repo)
```
ckerr: good catch :+1:

script/release-notes/notes.js
nornagon: yeah, makes sense. Did this & changed the `note` logic to match.

PR #16917, opened 2019-02-12
build: use async remove method to handle errors better

MarshallOfSound:
On windows removeSync randomly seems to fail with DIRNOTEMPTY. By using
the async version fs-extra will do some back-off-retry logic to
hopefully get this dir deleted

Notes: no-notes

### DIFF
--- a/script/gn-asar.js
+++ b/script/gn-asar.js
@@ -46,14 +46,18 @@ try {
   }
 } catch (err) {
   console.error('Unexpected error while generating ASAR', err)
-  fs.removeSync(tmpPath)
-  process.exit(1)
+  fs.remove(tmpPath)
+    .then(() => process.exit(1))
+    .catch(() => process.exit(1))
+  return
 }
 
 // Create the ASAR archive
 asar.createPackageWithOptions(tmpPath, out[0], {})
   .catch(err => {
-    fs.removeSync(tmpPath)
-    console.error('Unexpected error while generating ASAR', err)
-    process.exit(1)
+    const exit = () => {
+      console.error('Unexpected error while generating ASAR', err)
+      process.exit(1)
+    }
+    fs.remove(tmpPath).then(exit).catch(exit)
   }).then(() => fs.remove(tmpPath))


MarshallOfSound:
Windows failure is the crash-on-exit #16796

script/gn-asar.js
```diff
@@ -46,13 +46,15 @@ try {
...
+    .catch(() => process.exit(1))
+  return
 }
 
 // Create the ASAR archive
 asar.createPackageWithOptions(tmpPath, out[0], {}, (err) => {
-  fs.removeSync(tmpPath)
+  fs.remove(tmpPath)
 
   if (err) {
```
jkleinsc: Since fs.remove is async, it seems like it would be a good idea to:
a. check for error first
b. wait for fs.remove to finish before checking for error.

MarshallOfSound: Good point 👍

script/gn-asar.js
```diff
@@ -46,16 +46,20 @@ try {
...
-    console.error('Unexpected error while generating ASAR', err)
-    process.exit(1)
+  const checkErr = () => {
+    if (err) {
+      console.error('Unexpected error while generating ASAR', err)
+      process.exit(1)
+    }
   }
+
+  fs.remove(tmpPath).then(checkErr).catch(checkErr)
```
codebytere: @MarshallOfSound it's dying here bc the version of fs-extra we get for free in `package-lock` is too old: only `fs-extra` > v3 has inbuilt support for promises. It's probably too much effort to elevate `fs-extra` just to get promise support so maybe just use a callback here?

script/gn-asar.js
```diff
@@ -46,16 +46,20 @@ try {
   }
 } catch (err) {
   console.error('Unexpected error while generating ASAR', err)
-  fs.removeSync(tmpPath)
-  process.exit(1)
+  fs.remove(tmpPath)
```
codebytere: Same issue here as commented below

codebytere:
@MarshallOfSound actually this looks to conflict with https://github.com/electron/electron/pull/17061 which also accomplishes this goal: do you want to keep this open or close in favor of that one? It adds `fs-extra` as a `devDep` which solves the Promise issue anyways.

MarshallOfSound:
@codebytere That PR is missing one change in here (it still has a `removeSync`) gonna land that and rebase this on top of it 👍

Issue #16919, opened 2019-02-13
MAS build crashes with "unable to get application ASN from launchservicesd ..."


miniak:
* **Electron Version** (output of `node_modules/.bin/electron --version`):
  * 5.0.0-beta.2
* **Operating System** (Platform and Version):
  * macOS 10.13.6
* **Last known working Electron version** (if applicable):
  * 4.0.4

### Expected Behavior
Adding `<input type="checkbox" />` to the DOM in a sandboxed renderer process should work.
**Note:** `app.enableMixedSandboxed()` needs to be called in Electron 4.0 and older. 
### Actual behavior
The renderer process crashes.

### To Reproduce
- configure Fiddle to run [electron-v5.0.0-beta.2-mas-x64.zip](https://github.com/electron/electron/releases/download/v5.0.0-beta.2/electron-v5.0.0-beta.2-mas-x64.zip)
- add `app.enableMixedSandbox()` to Main Process code
- add `<input type="checkbox" />` to HTML code
- run the app

### Screenshots
Electron 5.0
<img width="912" alt="screen shot 2019-02-13 at 1 03 08 am" src="https://user-images.githubusercontent.com/1281234/52676703-52289000-2f2b-11e9-8bd7-41cf502a6ce5.png">
Electron 4.0
<img width="912" alt="screen shot 2019-02-13 at 1 03 47 am" src="https://user-images.githubusercontent.com/1281234/52676710-5654ad80-2f2b-11e9-931f-f592ff1e1dcb.png">

### Additional Information
```
Application Specific Information:
abort() called
rdar://problem/28724618 Couldn't create connection object
_RegisterApplication(), unable to get application ASN from launchservicesd, and this application requires an ASN, so aborting.  error=#-1.
 

Thread 0 Crashed:: CrRendererMain  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff5efb4b66 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff5f17f080 pthread_kill + 333
2   libsystem_c.dylib             	0x00007fff5ef1024d __abort + 144
3   libsystem_c.dylib             	0x00007fff5ef101bd abort + 142
4   com.apple.HIServices          	0x00007fff3578f60a _RegisterApplication + 7911
5   com.apple.HIServices          	0x00007fff3578d64c GetCurrentProcess + 24
[...]

PR #16920, opened 2019-02-13
fix: disable Chromium sandbox v2 in MAS builds

miniak:
#### Description of Change
Fixes https://github.com/electron/electron/issues/16919

/cc @nornagon

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Fixed Chromium sandbox v2 related crashes in MAS builds.

### DIFF
--- a/atom/app/atom_main.cc
+++ b/atom/app/atom_main.cc
@@ -43,9 +43,9 @@
 #include "base/i18n/icu_util.h"
 #include "electron/buildflags/buildflags.h"
 
-#if defined(HELPER_EXECUTABLE)
+#if defined(HELPER_EXECUTABLE) && !defined(MAS_BUILD)
 #include "sandbox/mac/seatbelt_exec.h"  // nogncheck
-#endif                                  // defined(HELPER_EXECUTABLE)
+#endif
 
 namespace {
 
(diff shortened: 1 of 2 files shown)


deepak1556:
Would it also be better to guard `"//sandbox/mac:seatbelt", "//sandbox/mac:seatbelt_extension",` 
 targets in upstream code behind `!is_mas_build` and simplify the patch in `mas_no_private_api.patch`. That way we are guaranteed to remove v2 sandbox code that may appear in future.

miniak:
@deepak1556 should we do that here or in a follow up PR?

deepak1556:
It can be a follow up, this PR addresses the immediate issue 👍

miniak:
> It can be a follow up, this PR addresses the immediate issue 👍

@deepak1556 can you please approve in that case?

miniak:
@nornagon are you ok with the fix?

atom/app/atom_main.cc
```diff
@@ -43,9 +43,9 @@
 #include "base/i18n/icu_util.h"
 #include "electron/buildflags/buildflags.h"
 
-#if defined(HELPER_EXECUTABLE)
+#if defined(ENABLE_SEATBELT)
```
nornagon: I'd rather the `HELPER_EXECUTABLE` code stay as-is and we change these `#ifs` to `#if defined(HELPER_EXECUTABLE) && !defined(MAS_BUILD)`. `ENABLE_SEATBELT` is a little misleading here because it will be undefined in the main executable even when seatbelt is enabled.

miniak: I actually did it like that initially :) will revert to that version

PR #16921, opened 2019-02-13
chore: refactor browser IPC into TS and app API into TS

MarshallOfSound:
* chore: refactor browser IPC into typescript
* chore: refactor app.ts into Typescript
  * Refactors app.dock into cpp
  * Removes app.launcher which has not existed for 3 years
  * Removes 2 deprecated APIs (that have been deprecated for more than one
major)
  * Refactors deprecate.ts as well

Notes: no-notes


lib/common/api/deprecate.ts
```diff
@@ -137,7 +135,7 @@ const deprecate = {
     // inject it and warn about it
     if ((oldName in o) && !(newName in o)) {
       warn()
-      o[newName] = o[oldName]
+      o[newName] = (o as any)[oldName]
```
felixrieseberg: This bothers me – and while trying to fix it, I noticed that we don't use `renameProperty` anywhere. Should we cut it?

lib/sandboxed_renderer/api/exports/electron.js
```diff
@@ -2,6 +2,12 @@
 
 const moduleList = require('@electron/internal/sandboxed_renderer/api/module-list')
 
+const handleESModule = (m) => {
```
felixrieseberg: Should we turn this into a mini-utility and add it to all the ESM imports?

lib/common/api/deprecate.ts
MarshallOfSound: We don't use it at the moment, but we may in the future and it sucks having to add back in code that we removed.  Better to have a battle tested un-used bit of code than write a new one each time IMO

codebytere: @felixrieseberg if you'd prefer to note lack of use, we can comment it out but i'd rather it remain  there in some form as opposed to removed i think!

lib/browser/ipc-main-internal-utils.ts
```diff
@@ -1,9 +1,9 @@
-'use strict'
+import { ipcMainInternal } from '@electron/internal/browser/ipc-main-internal'
+import * as errorUtils from '@electron/internal/common/error-utils'
 
-const { ipcMainInternal } = require('@electron/internal/browser/ipc-main-internal')
-const errorUtils = require('@electron/internal/common/error-utils')
+type IPCHandler = (...args: any[]) => any
```
nornagon: could this potentially be more tightly typed with type parameters? idk if typescript has variadic type parameters

lib/common/api/deprecate.ts
```diff
@@ -77,57 +75,57 @@ const deprecate = {
     })
   },
 
-  promisify: (fn) => {
+  promisify: <T extends (...args: any) => any>(fn: T): T => {
```
nornagon:
```suggestion
  promisify: <T extends (...args: any[]) => any>(fn: T): T => {
```
right?

MarshallOfSound: Yeah, interestingly `any[]` is a subset of `any`, updating 👍

lib/browser/ipc-main-internal-utils.ts
MarshallOfSound: The idea is the caller will provide the types.  Hence `handle` and `handleSync` have a generic

```js
<T extends IPCHandler> 
```

atom/browser/api/atom_api_app.cc
```diff
@@ -1259,6 +1259,46 @@ bool App::MoveToApplicationsFolder(mate::Arguments* args) {
...
+    dock_obj.SetMethod("hide", base::Bind(&Browser::DockHide, browser));
+    dock_obj.SetMethod("show", base::Bind(&Browser::DockShow, browser));
+    dock_obj.SetMethod("isVisible",
+                       base::Bind(&Browser::DockIsVisible, browser));
+    dock_obj.SetMethod("setMenu", &DockSetMenu);
+    dock_obj.SetMethod("setIcon", base::Bind(&Browser::DockSetIcon, browser));
+
+    dock_.Reset(isolate, dock_obj.GetHandle());
+  }
+  return v8::Local<v8::Value>::New(isolate, dock_);
```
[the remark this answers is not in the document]
MarshallOfSound: The template builder doesn't like that, hence the lazy-getter (also a tiny perf bonus, we don't bind / construct those things till you need them)

PR #16928, opened 2019-02-13
refactor: remove LaunchServices from LoginItem api

codebytere:
#### Description of Change

Previously, we had two different means of setting a login item on Mac: Launch Services and ServiceManagement. The Launch Services APIs are almost [entirely deprecated](https://developer.apple.com/documentation/coreservices/launch_services?language=objc), and often failed as what remained of them was rather buggy. This removes all traces of the LaunchServices APIs in favor of the ServiceManagement APIs. This means that macOS users will no longer be able to set login items as hidden, but this brings the API more in line with the Windows.

Requires https://github.com/electron/electron-typescript-definitions/pull/128.

cc @deepak1556 @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Removed deprecated LaunchServices from LoginItem APIs.

### DIFF
--- a/atom/browser/api/atom_api_app.cc
+++ b/atom/browser/api/atom_api_app.cc
@@ -321,7 +321,6 @@ struct Converter<Browser::LoginItemSettings> {
       return false;
 
     dict.Get("openAtLogin", &(out->open_at_login));
-    dict.Get("openAsHidden", &(out->open_as_hidden));
     dict.Get("path", &(out->path));
     dict.Get("args", &(out->args));
     return true;
@@ -331,10 +330,6 @@ struct Converter<Browser::LoginItemSettings> {
                                    Browser::LoginItemSettings val) {
     mate::Dictionary dict = mate::Dictionary::CreateEmpty(isolate);
     dict.Set("openAtLogin", val.open_at_login);
-    dict.Set("openAsHidden", val.open_as_hidden);
-    dict.Set("restoreState", val.restore_state);
-    dict.Set("wasOpenedAtLogin", val.opened_at_login);
-    dict.Set("wasOpenedAsHidden", val.opened_as_hidden);
     return dict.GetHandle();
   }
 };
--- a/spec/api-app-spec.js
+++ b/spec/api-app-spec.js
@@ -536,7 +536,7 @@ describe('app module', () => {
     ]
 
     before(function () {
-      if (process.platform === 'linux' || process.mas) this.skip()
+      if (process.platform !== 'win32') this.skip()
     })
 
     beforeEach(() => {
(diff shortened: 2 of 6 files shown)


spec/api-app-spec.js
```diff
@@ -536,7 +536,7 @@ describe('app module', () => {
     ]
 
     before(function () {
-      if (process.platform === 'linux' || process.mas) this.skip()
+      if (process.platform !== 'win32') this.skip()
```
nornagon: this is surprising, shouldn't these tests run on mac?

codebytere: @nornagon the SMLogin APIs are the ones that failed on MAS circle like 75% of the time; now those will also be present on non-MAS builds so we don't want to run them

nornagon: Wait, so the _new_ APIs are the failing ones? I'd like to find a way that we can reliably test these APIs.

atom/browser/api/atom_api_app.cc
```diff
@@ -331,10 +330,6 @@ struct Converter<Browser::LoginItemSettings> {
                                    Browser::LoginItemSettings val) {
     mate::Dictionary dict = mate::Dictionary::CreateEmpty(isolate);
     dict.Set("openAtLogin", val.open_at_login);
-    dict.Set("openAsHidden", val.open_as_hidden);
-    dict.Set("restoreState", val.restore_state);
-    dict.Set("wasOpenedAtLogin", val.opened_at_login);
-    dict.Set("wasOpenedAsHidden", val.opened_as_hidden);
```
sindresorhus: There seems to be a workaround for this. Should be documented: https://stackoverflow.com/questions/33181336/os-x-how-to-detect-when-app-got-launched-in-hidden-state-at-login?noredirect=1&lq=1

atom/browser/api/atom_api_app.cc
```diff
@@ -331,10 +330,6 @@ struct Converter<Browser::LoginItemSettings> {
                                    Browser::LoginItemSettings val) {
     mate::Dictionary dict = mate::Dictionary::CreateEmpty(isolate);
     dict.Set("openAtLogin", val.open_at_login);
-    dict.Set("openAsHidden", val.open_as_hidden);
-    dict.Set("restoreState", val.restore_state);
-    dict.Set("wasOpenedAtLogin", val.opened_at_login);
```
sindresorhus: Removing `wasOpenedAtLogin` is going to affect a bunch of apps: https://github.com/search?q=wasOpenedAtLogin&type=Code

In [Kap](https://github.com/wulkano/kap), we don't show the main window at system login, but instead just put Kap in the menu bar, but when launched normally, we show the main window. Removing this will hurt our UX...

spec/api-app-spec.js
codebytere: @nornagon yes, unfortunately :(

atom/browser/api/atom_api_app.cc
greenimpala: +1 @sindresorhus - this seems quite a regression. Do we still have no way of checking (on mas builds) whether the app was launched from login?

PR #16934, opened 2019-02-13
refactor: Port web-frame-init to TypeScript

felixrieseberg:
Another chunk of TypeScript conversion, this time around for `web-frame-init`.

Notes: no-notes


lib/renderer/web-frame-init.ts
```diff
@@ -0,0 +1,38 @@
...
+    new Promise(resolve =>
+      // The TypeScript compiler cannot handle the sheer number of
+      // call signatures here and simply gives up. Incorrect invocations
+      // will be caught by "keyof WebFrameMethod" though.
+      (webFrame[method] as any)(...args, resolve)
+    ).then(result => {
+      return [null, result]
+    }, error => {
+      return [errorUtils.serialize(error)]
+    }).then(responseArgs => {
```
MarshallOfSound: Doesn't this complain about implicitAny (and the `error` above)?

felixrieseberg: ✨ T y p e S c r i p t ✨ It's not an implicit `any`, the compiler thinks we're dealing with `{}`. We could make the `any` explicit but it wouldn't buy us anything here, so I'd rather leave it and wait for TS to improve.

MarshallOfSound:
@felixrieseberg Conflicting 😢

PR #16937, opened 2019-02-13
refactor: Port security warnings to TypeScript

felixrieseberg:
Another chunk of TypeScript conversion, this time around for `security-warnings`.

Notes: no-notes


lib/renderer/security-warnings.ts
```diff
@@ -1,6 +1,14 @@
-'use strict'
+import { webFrame } from 'electron'
+import { invokeSync } from '@electron/internal/renderer/ipc-renderer-internal-utils'
 
-let shouldLog = null
+declare global {
```
MarshallOfSound: Can this be put in our `ambient-*.d.ts` files so we don't have type mutations in individual files?

felixrieseberg: You can't mutate the global namespace from non-module files, so sadly not.

lib/renderer/security-warnings.ts
```diff
@@ -73,7 +79,7 @@ const isUnsafeEvalEnabled = function () {
         return false
       }
       return true
-    }).toString()})()`, resolve)
+    }).toString()})()`, undefined, resolve)
```
MarshallOfSound: Is this a docs issue, or an issue in the code that typescript caught?

lib/renderer/security-warnings.ts
```diff
@@ -264,17 +272,15 @@ const logSecurityWarnings = function (webPreferences, nodeIntegration) {
...
 const getWebPreferences = function () {
-  const ipcRendererUtils = require('@electron/internal/renderer/ipc-renderer-internal-utils')
-
   try {
-    return ipcRendererUtils.invokeSync('ELECTRON_BROWSER_GET_LAST_WEB_PREFERENCES')
+    return invokeSync('ELECTRON_BROWSER_GET_LAST_WEB_PREFERENCES')
   } catch (error) {
     console.warn(`getLastWebPreferences() failed: ${error}`)
-    return null
+    return undefined
```
MarshallOfSound: just `return` should work here right?

lib/renderer/security-warnings.ts
MarshallOfSound: @felixrieseberg You can mutate the top level `Window` interface though

```
declare interface Window {
  hideLoader: () => void;
}
```

(copy pasted from another project)

lib/renderer/security-warnings.ts
felixrieseberg: The latter 🙈

MarshallOfSound:
@felixrieseberg Conflicting 😢

PR #16938, opened 2019-02-13
refactor: Port getRemote to TypeScript

felixrieseberg:
Another chunk of TypeScript conversion, this time around for `web-frame-init`.

Notes: no-notes

### DIFF
--- a/lib/renderer/remote.ts
+++ b/lib/renderer/remote.ts
@@ -1,6 +1,6 @@
-import { remote, Remote } from 'electron'
+import { remote } from 'electron'
 
-export function getRemote (name: keyof Remote) {
+export function getRemote (name: keyof Electron.Remote) {
   if (!remote) {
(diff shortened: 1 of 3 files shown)


lib/renderer/remote.ts
```diff
@@ -1,8 +1,6 @@
-'use strict'
+import { remote, Remote } from 'electron'
 
-const { remote } = require('electron')
-
-exports.getRemote = function (name) {
+export function getRemote (name: keyof Remote) {
```
MarshallOfSound: Just as a point of standardization I've been using `Electron.Remote` instead of importing the `Remote` interface elsewhere in code.

Is there a strong case for one over the other?  (I'm not set on either)

felixrieseberg: Same end result, so it doesn't really matter. Let me go with your preferred version!

PR #16939, opened 2019-02-13
refactor: Port renderer-internal to TypeScript

felixrieseberg:
Another chunk of TypeScript conversion, this time around for `renderer-internal`.

Notes: no-notes


lib/renderer/ipc-renderer-internal.ts
```diff
@@ -1,10 +1,8 @@
-'use strict'
-
 const binding = process.atomBinding('ipc')
 const v8Util = process.atomBinding('v8_util')
 
 // Created by init.js.
-const ipcRenderer = v8Util.getHiddenValue(global, 'ipc-internal')
+export const ipcRenderer: Electron.IpcRenderer = v8Util.getHiddenValue(global, 'ipc-internal')
```
MarshallOfSound: Can we copy the main process here and `export const ipcRendererInternal: ...` to enforce a naming convention in our code 👍 (So we don't get confused between `ipcRenderer` and `ipcRendererInternal` 👍 )

felixrieseberg: Good callout! I've updated the whole PR.

MarshallOfSound:
Few TSC errors now 👍

felixrieseberg:
`**build-mac **` passed, it just failed in `mksnapshot` for `mas` again, which is unrelated 😬

PR #16942, opened 2019-02-13
 refactor: Port renderer-internal-utils to TypeScript 

felixrieseberg:
Another chunk of TypeScript conversion, this time around for `renderer-internal-utils`. We're importing things as `*` that are becoming named exports in other PRs, but opening these PRs in small chunks is probably still the best way to ensure that I'm not just opening one 10k line change monster 😅 

Notes: no-notes

### DIFF
--- a/lib/renderer/ipc-renderer-internal-utils.ts
+++ b/lib/renderer/ipc-renderer-internal-utils.ts
@@ -1,14 +1,14 @@
-'use strict'
-
-const ipcRenderer = require('@electron/internal/renderer/ipc-renderer-internal')
-const errorUtils = require('@electron/internal/common/error-utils')
+import * as ipcRenderer from '@electron/internal/renderer/ipc-renderer-internal'
+import * as errorUtils from '@electron/internal/common/error-utils'
 
 let nextId = 0
 
-exports.invoke = function (command, ...args) {
+export function invoke (command: string, ...args: any[]) {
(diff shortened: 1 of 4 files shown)


lib/renderer/ipc-renderer-internal-utils.ts
```diff
@@ -1,14 +1,14 @@
...
-const ipcRenderer = require('@electron/internal/renderer/ipc-renderer-internal')
-const errorUtils = require('@electron/internal/common/error-utils')
+import * as ipcRenderer from '@electron/internal/renderer/ipc-renderer-internal'
+import * as errorUtils from '@electron/internal/common/error-utils'
 
 let nextId = 0
 
-exports.invoke = function (command, ...args) {
+export function invoke (command: string, ...args: any[]) {
   return new Promise((resolve, reject) => {
```
MarshallOfSound: Can this be `T`?


```js
export function invoke<T=any>(command: string, ...args: any[]) {
  return new Promise<T>((resolve, reject) => {
```

The default resolve for `Promise` is `{}` which isn't very helpful

felixrieseberg:
The `build-mac` tests actually passed, we had a fluke along the way in `mksnapshot` that's unrelated:

```
ok v8_context_snapshot_generator successfully created v8_context_snapshot.bin
Too long with no output (exceeded 10m0s)
```

PR #16944, opened 2019-02-13
 fix: enable inputpane virtual keyboard by default

nornagon:
#### Description of Change
First commit here is just running `git-import-patches` followed by `git-export-patches`. The second commit backports https://chromium-review.googlesource.com/c/chromium/src/+/946928 to enable the inputpane virtual keyboard by default, which should fix some scenarios in Windows 10 RS4 and later in which the virtual keyboard wasn't showing when it should have.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed an issue on Windows 10 RS4 and later that was causing the virtual keyboard not to appear when focusing some input fields.

### DIFF
--- a/patches/common/chromium/.patches
+++ b/patches/common/chromium/.patches
@@ -90,3 +90,4 @@ sqlite_upgrade_from_3_24_to_3_26.patch
 sqlite_update_api_3_26.patch
 tts.patch
 do_not_allow_impl_side_invalidations_until_frame_sink_is_fully_active.patch
+enable_inputpane_virtual_keyboard_functionality_by_default.patch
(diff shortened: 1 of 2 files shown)


nornagon:
I prefer the backport because:
1. it'll disappear automatically when rebasing (though i suppose in this case, since it's already gone on master, this isn't really an advantage)
2. it's nice to have a direct backport of code that i know is in chrome
3. this allows disabling the feature with `--disable-features` rather than forcing it
4. i don't know where the "pass extra things to `--enable-features`" code in Electron is 😅

deepak1556:
1 and 2 are compelling reasons to have this as a patch :)

Regarding 4 , they are in [AtomBrowserMainParts::InitializeFeatureList](https://github.com/electron/electron/blob/master/atom/browser/atom_browser_main_parts.cc#L201-L215)

PR #16962, opened 2019-02-14
fix: Patch download architecture for ARM (fixes ARM broken install)

headmelted:
Electron no longer produces an 'arm' build, and now labels it 'armv7l'.  As NPM still uses 'arm', we need to patch the variable name here or the download fails with a 404 error like so:

 Error: GET https://github.com/electron/electron/releases/download/v3.1.3/electron-v3.1.3-linux-arm.zip returned 404

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:

### DIFF
--- a/npm/install.js
+++ b/npm/install.js
@@ -28,7 +28,7 @@ download({
   cache: process.env.electron_config_cache,
   version: version,
   platform: process.env.npm_config_platform,
-  arch: process.env.npm_config_arch,
+  arch: ( process.env.npm_config_arch == 'arm' ? 'armv7l' : process.env.npm_config_arch ),
   strictSSL: process.env.npm_config_strict_ssl === 'true',
   force: process.env.force_no_cache === 'true',
   quiet: process.env.npm_config_loglevel === 'silent' || process.env.CI


headmelted:
The linked issue isn't exactly the same problem - we'll still be getting passed "arm" but will need to re-direct to "armv7l" in that instance.

In the linked version we would fall back to the current architecture and version - but this would mean the options are either "be on a native armhf system" or "pass armv7l" - some scenarios can do neither.

In any case what we think of as "arm" is the "armv7l" in all cases, as ARM64 is handled seperately anyway - so this should be hard-coded on the way _**in**_ to electron-download, unless I've missed something?

(i.e. the electron download patch is approaching the native case - changing it here approaches the "passing arch in" case).

MarshallOfSound:
That PR should (in it's ideal state) handle people passing in `arm` and automatically download `armv7l` instead.  (I think that's what it does at the moment, just waiting for some tests)

zcbenz:
I'm closing this PR since there has been no update from the op for 5 months.

PR #16742, opened 2019-02-05
feat: promisify webContents.savePage()

codebytere:
#### Description of Change

This PR promisifies `webContents.savePage()`. There was, unfortunately, no way to perfectly maintain backwards compatibility that I was able to see; previously, this method returned a Boolean and called the callback with an error but now it calls the callback with both the Boolean and the error and returns nothing.

cc @deepak1556 @ckerr 

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Converted `webContents.savePage()` to return a Promise instead of taking a callback.

PR #16967, opened 2019-02-14
feat: promisify webContents.savePage()

codebytere:
#### Description of Change

Backport of https://github.com/electron/electron/pull/16742.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Converted `webContents.savePage()` to return a Promise instead of taking a callback.

### DIFF
--- a/spec/api-browser-window-spec.js
+++ b/spec/api-browser-window-spec.js
@@ -2532,12 +2532,8 @@ describe('BrowserWindow module', () => {
 
     it('should save page to disk', async () => {
       await w.loadFile(path.join(fixtures, 'pages', 'save_page', 'index.html'))
-      const error = await new Promise(resolve => {
-        w.webContents.savePage(savePageHtmlPath, 'HTMLComplete', function (error) {
-          resolve(error)
-        })
-      })
-      expect(error).to.be.null()
+      await w.webContents.savePage(savePageHtmlPath, 'HTMLComplete')
+
       assert(fs.existsSync(savePageHtmlPath))
       assert(fs.existsSync(savePageJsPath))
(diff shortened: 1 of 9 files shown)


spec/api-browser-window-spec.js
```diff
@@ -2532,12 +2532,8 @@ describe('BrowserWindow module', () => {
 
     it('should save page to disk', async () => {
       await w.loadFile(path.join(fixtures, 'pages', 'save_page', 'index.html'))
-      const error = await new Promise(resolve => {
-        w.webContents.savePage(savePageHtmlPath, 'HTMLComplete', function (error) {
-          resolve(error)
-        })
-      })
-      expect(error).to.be.null()
```
jkleinsc: Shouldn't we keep the assert that error is null, eg change the await to contain a try/catch or use .then().catch()

codebytere:
@jkleinsc this one's a WIP for now bc it's pending some linting failures (lint weirdly didn't run here?) resultant of changes that would affect our typescript definitions' smoke tests

spec/api-browser-window-spec.js
codebytere: In promise-land the only place to assert this error expectation would be in something like `.catch(err)`, but a catch function would only ever be triggered if there _is_ an error..so i'm fairly certain that's it'd be a catch-22 codepath in that it would never be asserted 99% of the time and the other 1% of the time it would fail by necessity of the conditions necessary to reach that assertion in the first place 🤔

jkleinsc: ok I think I'm fine with that as there are other asserts to verify savePage worked.

PR #16973, opened 2019-02-14
feat: promisify dialog.showOpenDialog()

codebytere:
#### Description of Change

Promisifies `dialog.showOpenDialog()`.

cc @ckerr @MarshallOfSound @deepak1556 

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Split `dialog.showOpenDialog()` into a synchronous version and a version that returns a Promise

PR #16978, opened 2019-02-15
chore: add Then helper for native promises

codebytere:
#### Description of Change

Adds a native `Then` helper for our promise wrapper utility to prevent truly terrible workarounds [like this](https://github.com/electron/electron/pull/16973/files#diff-1723fc06542f55eca6f1abc2a3988822R49).

cc @MarshallOfSound 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/atom/common/promise_util.h
+++ b/atom/common/promise_util.h
@@ -8,6 +8,7 @@
 #include <string>
 
 #include "atom/common/api/locker.h"
+#include "atom/common/native_mate_converters/callback.h"
 #include "content/public/browser/browser_thread.h"
 #include "native_mate/converter.h"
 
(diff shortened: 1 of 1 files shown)


atom/common/promise_util.h
```diff
@@ -46,6 +46,16 @@ class Promise : public base::RefCounted<Promise> {
     return GetInner()->Reject(GetContext(), v8::Undefined(isolate()));
   }
 
+  v8::MaybeLocal<v8::Promise> Then() {
```
MarshallOfSound: This would `CHECK` crash right? `v8::Undefined` can't be cast to `v8::Function`.

I don't think a zero-param `Then` is needed here

atom/common/promise_util.h
```diff
@@ -72,6 +82,16 @@ class Promise : public base::RefCounted<Promise> {
                               mate::ConvertToV8(isolate(), value));
   }
 
+  template <typename T>
+  v8::MaybeLocal<v8::Promise> Then(const T& value) {
```
MarshallOfSound: IMO this should not be a `template`'ed function and instead explicitly take a `base::Callback` and do the `mate::ConvertToV8` call that we need to do 🤔

PR #16994, opened 2019-02-15
chore: replace usage of deprecated beginSheetModalForWindow API

### LINKED ISSUE #16984: Replace deprecated beginSheetModalForWindow api
Remove clang pragma workaround https://github.com/electron/electron/commit/7743d51dbf4e66cd34a789ba2af9a35ff9dc2a2b

Replace `beginSheetModalForWindow:modalDelegate:didEndSelector:contextInfo:` for `beginSheetModalForWindow:completionHandler:`

MarshallOfSound:
#### Description of Change
Replaces usages of deprecated APIs that required some hax in the C73 upgrade

Fixes #16984 

This also updates the docs for a subtlety I noted in the `dialog.showMessageBox` API (validated against v3/v4 to ensure it wasn't a regression rather "always behavior")

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/ui/message_box_mac.mm
+++ b/atom/browser/ui/message_box_mac.mm
@@ -13,43 +13,6 @@
 #include "skia/ext/skia_utils_mac.h"
 #include "ui/gfx/image/image_skia.h"
 
-@interface ModalDelegate : NSObject {
- @private
-  atom::MessageBoxCallback callback_;
-  NSAlert* alert_;
-  bool callEndModal_;
-}
-- (id)initWithCallback:(const atom::MessageBoxCallback&)callback
-              andAlert:(NSAlert*)alert
-          callEndModal:(bool)flag;
-@end
-
-@implementation ModalDelegate
-
-- (id)initWithCallback:(const atom::MessageBoxCallback&)callback
-              andAlert:(NSAlert*)alert
-          callEndModal:(bool)flag {
-  if ((self = [super init])) {
-    callback_ = callback;
-    alert_ = alert;
-    callEndModal_ = flag;
-  }
-  return self;
-}
-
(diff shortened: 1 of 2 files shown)


atom/browser/ui/message_box_mac.mm
```diff
@@ -192,21 +145,17 @@ void ShowMessageBox(NativeWindow* parent_window,
...
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-    [alert
-        beginSheetModalForWindow:window
-                   modalDelegate:delegate
-                  didEndSelector:@selector(alertDidEnd:returnCode:contextInfo:)
-                     contextInfo:nil];
-#pragma clang diagnostic pop
+    // Duplicate the callback object here since c is a reference and gcd would
+    // only store the pointer, by duplication we can force gcd to store a copy.
```
nornagon: I don't suppose this is something that we could do with the standard c++ syntax for block binding? e.g.

```mm
[alert beginSheetModalForWindow:window 
              completionHandler:^[&, callback](NSModalResponse response) {
                // ...
              }]
```

MarshallOfSound: @nornagon I copied this from another `completionHandler` usage in our code.  Seem to remember this being the best workaround back when we wrote that.  I tried binding like your comment and it crashes 😢

deepak1556: The binding is possible, the variable should be captured by value rather than by reference

```
completionHandler: [=](NSModalResponse response) {}
```

MarshallOfSound: @deepak1556 If this is the case I'll update these _everywhere_ in our code in a follow up PR

PR #17012, opened 2019-02-17
refactor: Port content-scripts-injector to TypeScript

felixrieseberg:
Another chunk of TypeScript conversion, this time around for `content-scripts-injector`.

Notes: no-notes


typings/internal-electron.d.ts
```diff
@@ -27,6 +27,25 @@ declare namespace Electron {
     __ELECTRON_SERIALIZED_ERROR__: true
   }
 
+  interface InjectionBase {
```
MarshallOfSound: Can these be shifted down into the `ElectronInternal` namespace?

PR #17014, opened 2019-02-17
refactor: Port chrome-api to TypeScript

felixrieseberg:
Another chunk of TypeScript conversion, this time around for `chrome-api`.

Notes: no-notes

### DIFF
--- a/lib/renderer/chrome-api.ts
+++ b/lib/renderer/chrome-api.ts
@@ -1,18 +1,24 @@
-'use strict'
+import { ipcRendererInternal } from '@electron/internal/renderer/ipc-renderer-internal'
+import * as ipcRendererUtils from '@electron/internal/renderer/ipc-renderer-internal-utils'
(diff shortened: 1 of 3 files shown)


lib/renderer/chrome-api.ts
```diff
@@ -92,12 +107,14 @@ exports.injectTo = function (extensionId, isBackgroundPage, context) {
       })
     },
 
+    // https://developer.chrome.com/extensions/runtime#method-getManifest
```
samuelmaddock: TypeScript will associate these comments with the function if using the Docs style comments (e.g. `/** comment */`).
https://typedoc.org/guides/doccomments/

felixrieseberg: That's actually on purpose – one shouldn't mistake the link for the _actual_ implementation, but it's a useful guide for developers coming in and working with this code.

PR #17027, opened 2019-02-18
refactor: Port renderer/init to TypeScript

felixrieseberg:
Another chunk of TypeScript conversion, this time around for `renderer/init`.

Notes: no-notes


lib/renderer/init.ts
```diff
@@ -34,7 +33,9 @@ webFrameInit()
 // Process command line arguments.
 const { hasSwitch, getSwitchValue } = process.atomBinding('command_line')
 
-const parseOption = function (name, defaultValue, converter = value => value) {
+const parseOption = function (
+  name: string, defaultValue: any, converter: (value: any) => any = value => value
```
MarshallOfSound: The `converter` here can be typed as `string` and `T`.

```suggestion
  name: string, defaultValue: T, converter: (value: string) => T = value => value
```

Then `parseOption` is strongly typed

typings/internal-ambient.d.ts
```diff
@@ -36,7 +36,16 @@ declare namespace NodeJS {
   }
 }
 
+declare module NodeJS  {
```
MarshallOfSound: This concerns me, why are we having to redeclare node globals, shouldn't they be coming through from `@types/node`

felixrieseberg: Ah, I can answer that! `@types/node` has those defined as _actual_ globals.

Not defined:
```
global.__dirname
```

Defined:
```
__dirname
```

lib/renderer/init.ts
felixrieseberg: Upon closer investigation, that doesn't work – and it makes sense, this here doesn't work: `(value: string) => T = value => value`

felixrieseberg:
The two failures are unrelated :v:

PR #17028, opened 2019-02-18
refactor: convert AutoUpdater + CrashReporter to Typescript

MarshallOfSound:
This PR converts the AutoUpdater and CrashReporter APIs to Typescript, please take **a lot** of attention to the AutoUpdater refactor, it's a very sensitive part of the codebase 👍 

Notes: no-notes


lib/browser/api/auto-updater/auto-updater-win.ts
```diff
@@ -66,8 +67,12 @@ class AutoUpdater extends EventEmitter {
 
   // Private: Emit both error object and message, this is to keep compatibility
   // with Old APIs.
-  emitError (message) {
-    this.emit('error', new Error(message), message)
+  emitError (message: Error | string) {
+    if (message instanceof Error) {
```
felixrieseberg: Nice!

lib/browser/api/auto-updater/squirrel-update-win.ts
```diff
@@ -79,20 +77,25 @@ const spawnUpdate = function (args, detached, callback) {
 }
 
 // Start an instance of the installed app.
-exports.processStart = function () {
+export const processStart = () => {
   return spawnUpdate(['--processStartAndWait', exeName], true, function () {})
 }
 
+type SquirrelUpdateInfo = {
```
felixrieseberg: We could move that to `Electron.*`, but I think it's fine here, too

MarshallOfSound: If it was re-used for the macOS impl or something I totally would extract it higher, but this purely lives for this single fn 🤔

MarshallOfSound:
@jkleinsc At the beginning they were intrinsically linked (I can't remember how) but they appear to be slightly less linked now as other PRs doing the same work have landed.  Splitting it out would still result in two PRs that would instantly conflict however so in this case I'd rather we keep it unified.

I can enable the crashReporter specs on all platforms and see how that plays out.  I have manually validated that these files appear to do what they used to do though 😄

zcbenz:
It seems that the crashReporter tests are passing on macOS but not Linux.

@jkleinsc Do you think should we only convert AutoUpdater + CrashReporter files after the tests get fixed in CI?

MarshallOfSound:
So out of date it's not funny, I'll raise a new PR at some point 👍

PR #17038, opened 2019-02-19
refactor: Split 'Event' docs/types into more specific Event types

MarshallOfSound:
* Event = Base event type (with preventDefault)
* IpcMainEvent = Event that ipcMain emits (with sender, reply, etc.)
* IpcRendererEvent = Event that ipcRenderer emits (with sender,
senderId, etc.)
* KeyboardEvent = Event that we emit with keyboard flags (ctrlKey,
altKey, etc.)

This will dramatically improve peoples TS experience with IPC events

For full context on the changes please review the [Archeologist Check](https://github.com/electron/electron/pull/17038/checks?check_run_id=65246647)

This also allows us to more strictly type the different between `ipcMain` and `ipcMainInternal` (the `reply` vs `_replyInternal`)

Notes: no-notes


MarshallOfSound:
Further note, this is required for strong typing and further TSification of the `browser` code in particular things like the `navigation-controller`

PR #17042, opened 2019-02-19
Fix docs for application architecture

rousan:
I just fixed the docs for `remote` module availability on `renderer` process.

### DIFF
--- a/docs/tutorial/application-architecture.md
+++ b/docs/tutorial/application-architecture.md
@@ -73,7 +73,7 @@ const win = new BrowserWindow()
 Since communication between the processes is possible, a renderer process
 can call upon the main process to perform tasks. Electron comes with a
 module called `remote` that exposes APIs usually only available on the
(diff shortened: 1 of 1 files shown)


felixrieseberg:
Hey! Thanks for the PR, but the original text is correct - the `remote` exposes APIs usually only available on the `main` process.

rousan:
@felixrieseberg 

```javascript
const {app, BrowserWindow, remote} = require('electron')
console.log(remote);
```
it gives me `undefined` in `main` process.

codebytere:
@rousan what @felixrieseberg  said is correct:
> the `remote` exposes APIs usually only available on the `main` process.

That is, the `remote` module is _used_ in the _renderer_ process, not the main process.

rousan:
Hey @codebytere,
Your comment makes me confused.
if `remote` module is _only_ available on the `main` process, then why is it _used_ only on the renderer process, not the main process?

PR #17053, opened 2019-02-19
refactor: Port api/web-frame to TypeScript

felixrieseberg:
Another chunk of TypeScript conversion, this time around for `api/web-frame`.

Notes: no-notes

### DIFF
--- a/lib/renderer/api/web-frame.ts
+++ b/lib/renderer/api/web-frame.ts
@@ -1,27 +1,25 @@
-'use strict'
+import { EventEmitter } from 'events'
+import { deprecate } from 'electron'
 
-const { EventEmitter } = require('events')
 const binding = process.atomBinding('web_frame')
-const { deprecate } = require('electron')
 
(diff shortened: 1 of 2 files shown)


lib/renderer/api/web-frame.ts
```diff
@@ -51,38 +49,42 @@ class WebFrame extends EventEmitter {
...
 
 // Helper to return WebFrame or null depending on context.
 // TODO(zcbenz): Consider returning same WebFrame for the same frame.
-function getWebFrame (context) {
+function getWebFrame (context: Window) {
   return context ? new WebFrame(context) : null
 }
 
-module.exports = new WebFrame(window)
+const _webFrame = new WebFrame(window)
```
MarshallOfSound: We can't `export default new WebFrame(window)` here?

felixrieseberg: We can, but this way it's named if anything goes wrong!

PR #17057, opened 2019-02-19
docs: fix missing escape

LongTengDao:
#### Description of Change

`String[]` should be escaped as `String\[]` in markdown, which means array of string, not string.

#### Checklist

#### Release Notes

Notes: no-notes


MarshallOfSound:
I believe what this is trying to fix is the mis-renderering of this doc on our website, this is caused by the markdown renderer we use and @BinaryMuse is all over fixing it 🎉 --> https://github.com/electron/electronjs.org/pull/2160 (using a better markdown renderer).

Although surprisingly this doesn't cause a `.d.ts` file change it also doesn't change the rendered markdown when run through GitHub's markdown renderer.

Because of the noop nature of this change I'm going to close this out, let me know in a follow up if I'm missing something here 😄

PR #17061, opened 2019-02-20
build: upgrade to asar@^1.0.0

malept:
#### Description of Change

* The `asar` upgrade is to remove a transitive security vulnerability. The API also changed from callbacks to Promises, so `script/gn-asar.js` needed to be changed.
* `klaw` is added to `devDependencies`. It was a transitive dependency that was removed when `asar` was upgraded. Meanwhile, it is used directly in `script/lint.js`, which made it very difficult to commit :smile:

#### Checklist

- [x] `npm test` passes
- ~~tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)~~ not relevant, changed build script
- ~~relevant documentation is changed or added~~ not necessary, did not change build functionality

#### Release Notes

Notes: no-notes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6731,6 +6731,15 @@
         "is-buffer": "^1.1.5"
       }
     },
+    "klaw": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/klaw/-/klaw-3.0.0.tgz",
+      "integrity": "sha512-0Fo5oir+O9jnXu5EefYbVK+mHMBeEVEy2cmctR1O1NECcCkPRreJKrS6Qt/j3KC2C148Dfo9i3pCmCMsdqGr0g==",
+      "dev": true,
+      "requires": {
+        "graceful-fs": "^4.1.9"
+      }
+    },
     "labeled-stream-splicer": {
       "version": "2.0.1",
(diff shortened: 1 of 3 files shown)


malept:
Apparently `fs-extra` was also an implicit devDependency :cry:

malept:
I'm unsure about the mac build failures. I don't think my changes should affect that?

GitSquared:
@MarshallOfSound Could it be possible to backport this to the [4-0-x](https://github.com/electron/electron/tree/4-0-x) branch?

The 1.0 release of `asar` fixed a blocking bug on electron apps packaged for Windows with both `asar`-packed resources and native dependencies that caused a pretty nasty `callStackSizeExceeded` error (see electron/asar#139). I'm wondering when we can expect the fix to land in electron stable.

malept:
@GitSquared I'm a bit confused - according to the PR you linked, it was fixed in `asar@0.14.1`, which was released on 2018-01-30.

GitSquared:
@malept Wait, you're right. `asar` version in electron 4.0-x is set to `^0.11.0`, which should include `asar@0.14.1`... Sorry, this might be another bug.

MarshallOfSound:
@GitSquared The dep is also only a devDep used for Electron's internal build process, it has no impact on anyone using Electron

GitSquared:
@MarshallOfSound Looks like I got confused on this one. Thanks for the info.

PR #17062, opened 2019-02-20
refactor: merge ipcMainUtils.handle / handleSync

miniak:
#### Description of Change
There is no need for a separate `handleSync` method, we can tell whether the request is sync or async by the presence of the `requestId`.

/cc @MarshallOfSound

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/chrome-extension.js
+++ b/lib/browser/chrome-extension.js
@@ -170,7 +170,7 @@ ipcMainInternal.on('CHROME_RUNTIME_CONNECT', function (event, extensionId, conne
   page.webContents._sendInternalToAll(`CHROME_RUNTIME_ONCONNECT_${extensionId}`, event.sender.id, portId, connectInfo)
 })
 
-ipcMainUtils.handleSync('CHROME_EXTENSION_MANIFEST', function (event, extensionId) {
+ipcMainUtils.handle('CHROME_EXTENSION_MANIFEST', function (event, extensionId) {
   const manifest = manifestMap[extensionId]
   if (!manifest) {
     throw new Error(`Invalid extensionId: ${extensionId}`)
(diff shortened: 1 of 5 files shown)


lib/browser/ipc-main-internal-utils.ts
```diff
@@ -15,15 +15,11 @@ const callHandler = async function (handler: IPCHandler, event: ElectronInternal
...
-      event._replyInternal(`${channel}_RESPONSE_${requestId}`, ...responseArgs)
-    })
-  })
-}
-
-export const handleSync = function <T extends IPCHandler> (channel: string, handler: T) {
-  ipcMainInternal.on(channel, (event, ...args) => {
-    callHandler(handler, event, args, responseArgs => {
-      event.returnValue = responseArgs
+      if (requestId) {
```
alexeykuzmin: How do check if it's really the `requestId` and not one of the `args`?

miniak: why would it be one of the args? `invokeSync` sends a `null` there

miniak:
@MarshallOfSound can you please have a look?

lib/browser/chrome-extension.js
```diff
@@ -232,7 +232,7 @@ const getMessagesPath = (extensionId) => {
   }
 }
 
-ipcMainUtils.handleSync('CHROME_GET_MESSAGES', function (event, extensionId) {
+ipcMainUtils.handle('CHROME_GET_MESSAGES', function (event, extensionId) {
   const messagesPath = getMessagesPath(extensionId)
   return fs.readFileSync(messagesPath)
```
MarshallOfSound: Not required in this PR, but worth noting this can become `async`

miniak: I've actually noticed that and have a PR https://github.com/electron/electron/pull/17148 :)

Issue #16799, opened 2019-02-06
Make util::Promise a move-only type


zcbenz:
Passing `scoped_refptr` across threads is very dangerous as it would be very easy to accidentally increase/decrease on the wrong thread, currently we already have a couple issues caused by that:
https://github.com/electron/electron/pull/16788, https://github.com/electron/electron/pull/16738.

We should change the design of `util::Promise` to prevent this from happening from the first place, and the best way should be making `util::Promise` a move-only type.

deepak1556:
Took a stab at it with a slight different approach https://github.com/electron/electron/commit/6450f3de4d6ffc223de6f146286c86fff5a9c75d that mimics task runner's traits implementation but its not ideal solution for developers to interact with promises natively.

But on a larger scale v8::promises are not meant to be dealt directly with chromium types, but this is a problem even chromium is trying to solve by introducing [base::Promise](https://docs.google.com/document/d/1l12PAJgEtlrqTXKiw6mk2cR2jP7FAfCCDr-DGIdiC9w/edit#).

PR #17071, opened 2019-02-20
chore: make util::Promise a move-only type

### LINKED ISSUE #17037: CHECK failure cookies API
In the Electron 5 branch and master we have a CHECK failure in the cookies API

> `[401:0218/145432.207065:FATAL:ref_counted.h(63)] Check failed: CalledOnValidSequence(). `

Example: https://circleci.com/gh/electron/electron/138174

It doesn't happen 100% of the time, but shouldn't be too hard to repro on a Linux machine.  (Unsure if it fails on other OS's, haven't seen it personally but may have happened)

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/16799, close #16788, close #17037.

By making `util::Promise` a move-only type, we can:
* Eliminate the `CalledOnValidSequence` error when passing promise between threads as explained in #16799;
* Make the code easier to understand since Promise is no longer ref-counted;
* Have slight performance gain.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/api/atom_api_cookies.cc
+++ b/atom/browser/api/atom_api_cookies.cc
@@ -136,29 +136,29 @@ inline net::CookieStore* GetCookieStore(
   return getter->GetURLRequestContext()->cookie_store();
 }
 
-void ResolvePromiseWithCookies(scoped_refptr<util::Promise> promise,
-                               net::CookieList cookieList) {
-  promise->Resolve(cookieList);
+void ResolvePromiseWithCookies(util::Promise promise,
+                               const net::CookieList& cookie_list) {
+  promise.Resolve(cookie_list);
 }
 
-void ResolvePromise(scoped_refptr<util::Promise> promise) {
-  promise->Resolve();
+void ResolvePromise(util::Promise promise) {
+  promise.Resolve();
 }
 
 // Resolve |promise| in UI thread.
-void ResolvePromiseInUI(scoped_refptr<util::Promise> promise) {
+void ResolvePromiseInUI(util::Promise promise) {
   base::PostTaskWithTraits(FROM_HERE, {BrowserThread::UI},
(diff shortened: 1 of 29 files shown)


atom/browser/api/atom_api_cookies.cc
```diff
@@ -174,58 +174,58 @@ void FilterCookies(std::unique_ptr<base::DictionaryValue> filter,
...
 void GetCookiesOnIO(scoped_refptr<net::URLRequestContextGetter> getter,
                     std::unique_ptr<base::DictionaryValue> filter,
-                    scoped_refptr<util::Promise> promise) {
+                    util::Promise promise) {
   std::string url;
   filter->GetString("url", &url);
 
   auto filtered_callback =
-      base::Bind(FilterCookies, base::Passed(&filter), std::move(promise));
+      base::BindOnce(FilterCookies, base::Passed(&filter), std::move(promise));
```
deepak1556: Can replace `base::Passed` with `std::move` under BindOnce

atom/common/promise_util.h
```diff
@@ -16,16 +16,30 @@ namespace atom {
 
 namespace util {
 
-class Promise : public base::RefCounted<Promise> {
+// A wrapper around the v8::Promise.
+//
+// This is a move-only type that should always be `std::move`d when passed to
+// callbacks, and it should be destroyed on the same thread of creation.
```
deepak1556: Can you add a [SequenceChecker](https://cs.chromium.org/chromium/src/base/sequence_checker.h) macro to assert this criteria ?

zcbenz: I just tried it and unfortunately it does not support moving, I'll see if I can upstream a change to add move constructor for SequenceChecker.

PR #17084, opened 2019-02-21
refactor: let mate converters do their job, don't call GetHandle manually

MarshallOfSound:
We have mate converters for `util::Promise`, there is no need to convert to `v8::Local<T>` manually, in fact this will allow us to more easily change how `util::Promise` works in the future.

Notes: no-notes

### DIFF
--- a/atom/browser/api/atom_api_app.cc
+++ b/atom/browser/api/atom_api_app.cc
@@ -1123,10 +1123,9 @@ JumpListResult App::SetJumpList(v8::Local<v8::Value> val,
 }
 #endif  // defined(OS_WIN)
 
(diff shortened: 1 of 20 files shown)


deepak1556:
I am down with having this merged once the sequence checker macro has been added to `util::Promise`.

zcbenz:
The `promise` has been `std::move`d before being returned, so the code is actually returning an empty handle. That's why the tests are failing with crashes.

I'm actually surprised that the compiler did not complain about this, since this belongs to use-after-move errors.

codebytere:
@MarshallOfSound should we close this one out given the above comments??

MarshallOfSound:
Closing out as per above comments

PR #17097, opened 2019-02-22
fix: Removed $ in order to allow simple copy & paste into console

ALIENQuake:
#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Removed $ in order to allow simple copy & paste into console

### DIFF
--- a/docs/tutorial/first-app.md
+++ b/docs/tutorial/first-app.md
@@ -208,13 +208,13 @@ Clone and run the code in this tutorial by using the
 
 ```sh
 # Clone the repository
-$ git clone https://github.com/electron/electron-quick-start
+git clone https://github.com/electron/electron-quick-start
 # Go into the repository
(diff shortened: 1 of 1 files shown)


ALIENQuake:
> @ALIENQuake we already handle this [on the website](https://electronjs.org/docs/tutorial/first-app#trying-this-example): if you copy the snippet with the `COPY` button you'll see that the `$` instances are not copied.

@codebytere But it doesn't have anything to do when the user manually select text and use copy & paste, right? Anyway, this change is very minor, no big deal if it won't be merged.

MarshallOfSound:
@ALIENQuake We can make those `$` symbols not selectable in the code box 🤔 This should be moved over to the website repo.

I've raised https://github.com/electron/electronjs.org/pull/2219 which should do what you want 👍

PR #17100, opened 2019-02-23
build: enable gn check for //electron:electron_lib

deepak1556:
#### Description of Change

- [x] Fix gn check errors for `//electron`
- [x] Add --gn-check lint option
- [x] Add check to ci

The headers included from some private targets have been suffixed with `nogncheck`, these targets are known to be built as a transitive dep, so lets avoid a visibility patch in chromium. Also this gives a good idea of private headers we depend on, lets try to minimize them. 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/api/atom_api_app.cc
+++ b/atom/browser/api/atom_api_app.cc
@@ -34,8 +34,8 @@
 #include "chrome/browser/browser_process.h"
 #include "chrome/browser/icon_manager.h"
 #include "chrome/common/chrome_paths.h"
-#include "content/browser/gpu/compositor_util.h"
-#include "content/browser/gpu/gpu_data_manager_impl.h"
+#include "content/browser/gpu/compositor_util.h"        // nogncheck
+#include "content/browser/gpu/gpu_data_manager_impl.h"  // nogncheck
 #include "content/public/browser/browser_accessibility_state.h"
 #include "content/public/browser/browser_child_process_host.h"
 #include "content/public/browser/child_process_data.h"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -237,6 +237,7 @@ step-gn-check: &step-gn-check
     command: |
       cd src
       gn check out/Default //electron:electron_lib
+      gn check out/Default //electron:electron_app
 
 step-electron-build: &step-electron-build
   run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -579,6 +579,7 @@ steps-electron-gn-check: &steps-electron-gn-check
   steps:
     - attach_workspace:
         at: .
+    - *step-depot-tools-add-to-path
     - *step-setup-env-for-build
     - *step-gn-gen-default
     - *step-gn-check
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,6 +231,13 @@ step-gn-gen-default: &step-gn-gen-default
       cd src
       gn gen out/Default --args='import("'$GN_CONFIG'") cc_wrapper="'"$SCCACHE_PATH"'"'" $GN_EXTRA_ARGS"
 
+step-gn-check: &step-gn-check
+  run:
+    name: GN check
+    command: |
+      cd src
+      gn check out/Default //electron:electron_lib
+
 step-electron-build: &step-electron-build
   run:
     name: Electron build
@@ -568,6 +575,14 @@ steps-checkout: &steps-checkout
           - depot_tools
           - src
 
+steps-electron-gn-check: &steps-electron-gn-check
+  steps:
+    - attach_workspace:
+        at: .
+    - *step-setup-env-for-build
+    - *step-gn-gen-default
+    - *step-gn-check
+
 steps-electron-build: &steps-electron-build
   steps:
     - attach_workspace:
@@ -857,6 +872,13 @@ jobs:
       <<: *env-enable-sccache
     <<: *steps-electron-build
 
+  linux-x64-debug-gn-check:
+    <<: *machine-linux-medium
+    environment:
+      <<: *env-linux-medium
+      <<: *env-debug-build
+    <<: *steps-electron-gn-check
+
   linux-x64-testing:
     <<: *machine-linux-2xlarge
     environment:
@@ -864,6 +886,13 @@ jobs:
       <<: *env-testing-build
       <<: *env-enable-sccache
     <<: *steps-electron-build-for-tests
+
+  linux-x64-testing-gn-check:
+    <<: *machine-linux-medium
+    environment:
+      <<: *env-linux-medium
+      <<: *env-testing-build
+    <<: *steps-electron-gn-check
 
   linux-x64-chromedriver:
     <<: *machine-linux-medium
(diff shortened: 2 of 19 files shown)


atom/browser/api/atom_api_app.cc
```diff
@@ -34,8 +34,8 @@
 #include "chrome/browser/browser_process.h"
 #include "chrome/browser/icon_manager.h"
 #include "chrome/common/chrome_paths.h"
-#include "content/browser/gpu/compositor_util.h"
-#include "content/browser/gpu/gpu_data_manager_impl.h"
+#include "content/browser/gpu/compositor_util.h"        // nogncheck
+#include "content/browser/gpu/gpu_data_manager_impl.h"  // nogncheck
```
nornagon: can we have additional comments for nogncheck'd files re: why we can't just have the correct dependencies?

nornagon: er nvm, you commented on this in the PR description.

deepak1556:
@nornagon can you review this again for precommit hook changes.

script/lint.js
```diff
@@ -200,17 +235,21 @@ async function main () {
...
 
   const linters = LINTERS.filter(x => opts[x.key])
 
   for (const linter of linters) {
-    const filenames = await findFiles(opts, linter)
-    if (filenames.length) {
-      if (opts.verbose) { console.log(`linting ${filenames.length} ${linter.key} ${filenames.length === 1 ? 'file' : 'files'}`) }
-      linter.run(opts, filenames)
+    if (linter.key === 'gn-check') {
+      linter.run(opts)
```
nornagon: we should only run `gn check` if a *.{gn,gni,h,c,cc,m,mm} file changes.

package.json
```diff
@@ -102,7 +102,8 @@
     ],
     "*.{gn,gni}": [
       "python script/run-gn-format.py",
-      "git add"
+      "git add",
+      "node script/lint.js --gn-check --"
```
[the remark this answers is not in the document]
deepak1556: > can `gn check` change a file?

Nope it doesn't, the errors needs to be fixed manually. Thats why I added it after `git add`

script/lint.js
deepak1556: Adding this to `lint-staged` hook should take care of it.

deepak1556:
gn gen takes considerable time that might be intrusive for a pre-commit hook even if its run on a cached dir. Moreover we need to test both testing and debug config, maybe we should just have this as a pre-push hook instead ?

nornagon:
Could we run it on a bot?

deepak1556:
Yeah that would make things a lot easier, if I add them as a step in the testing and build CI runs, since this needs a complete source checkout and should be tested across all platforms. /cc @jkleinsc thoughts on this ?

MarshallOfSound:
@deepak1556 How long would this take to run, and would the output / result change on different platforms?

jkleinsc:
@deepak1556 is there any reason to run it against multiple architectures or could we just run it on one?

deepak1556:
> How long would this take to run

```
Header dependency check OK

real	0m10.266s
user	0m25.371s
sys	0m5.329s
```

> would the output / result change on different platforms?

Yup only when you have platform specific includes and deps.

MarshallOfSound:
Hm yeah that's probably a bit too long for pre-commit 🤔 I think letting the `lint` stage catch this on CircleCI would be enough

deepak1556:
> I think letting the `lint` stage catch this on CircleCI would be enough

I don't think the lint ci has the complete src/ checkout, this check needs it. Any concern in adding this as a step in the build CI runs ? Just need the following variations, the cpu arch doesn't matter since we don't include deps based on that condition.

```
win-testing
win-debug
linux-testing
linux-debug
mac-testing
mac-debug
```

MarshallOfSound:
@deepak1556 I'm 👍 on adding it to those jobs you mentioned but **only** if it is the first step in those jobs.  i.e. the check runs before we try to build Electron.  Otherwise we might have to wait an hour just to find out about a lint failure 👍

nornagon:
Perhaps on circleci it would make sense to run it in parallel with the build? that way if lint fails, you still get to see the result of the tests.

deepak1556:
I could use some help with fixing gn check errors on linux, thanks! will push fix for errors on windows shortly.

script/lint.js
```diff
@@ -114,12 +114,47 @@ const LINTERS = [ {
       process.exit(1)
     }
   }
+}, {
+  key: 'gn-check',
+  run: (opts) => {
+    const env = Object.assign({
```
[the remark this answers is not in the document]
nornagon: still think spread > Object.assign :)

script/lint.js
```diff
@@ -199,18 +239,23 @@ async function findFiles (args, linter) {
 async function main () {
   const opts = parseCommandLine()
 
-  // no mode specified? run 'em all
-  if (!opts['c++'] && !opts.javascript && !opts.python && !opts.gn) {
-    opts['c++'] = opts.javascript = opts.python = opts.gn = true
+  // no mode specified? run 'em all excpet gn-check,
```
nornagon:
```suggestion
  // no mode specified? run 'em all except gn-check,
```

script/lint.js
nornagon: since this is running in CI now, this is moot.

script/lint.js
```diff
@@ -41,6 +41,32 @@ function spawnAndCheckExitCode (cmd, args, opts) {
   if (status) process.exit(status)
 }
 
+function gnCheckConfig (config, env) {
```
nornagon: i'm inclined to drop this from the script now that it's just a simple `gn check` in CI.

deepak1556: Yup I would be in favor of that, wasn't sure if users wanted to a simple one liner to replicate the failure they saw in ci.

nornagon: i think this one-liner will probably cause more problems than it solves. `gn check out/Debug //electron:electron_lib` is already a one-liner :)

deepak1556:
@jkleinsc @MarshallOfSound can you review the ci config changes. Thanks!

jkleinsc:
@nornagon since you suggested running gn check in parallel, I'm wondering if you still stand by that given how quickly gn check runs?  We are incurring an extra 5 minutes of compute time to run a 15 second check.

nornagon:
the reason i suggested it being a parallel build is so that if the `gn check` fails you also get build/test results. it sucks when lint prevents your build from happening, especially when it takes 15+ minutes for that build to even start.

my desired solutions, from most to least desirable:
1. have the `check` step run on the same checkout as the build/tests, but have it show up as a separate check from the build/test results in the GitHub UI
2. have the `check` step run on the same checkout as the build/tests, but let the build keep going and only report failure at the end
3. have the `check` step block the build

if we end up having to go with (3) it wouldn't be the worst thing, just a bit of a bummer.

jkleinsc:
I did not realize we were trying to build even if gn check failed.  If that is the case, I'm fine with the PR as is.

PR #17121, opened 2019-02-25
refactor: make shell.ShowItemInFolder asynchronous

codebytere:
#### Description of Change

BREAKING CHANGE: change return type of `ShowItemInFolder` to `void` from `bool` and offload task to new thread.

cc @nornagon @deepak1556

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Made `ShowItemInFolder` asynchronous with no return value.

### DIFF
--- a/atom/common/platform_util_win.cc
+++ b/atom/common/platform_util_win.cc
@@ -232,28 +232,29 @@ HRESULT DeleteFileProgressSink::ResumeTimer() {
 
 namespace platform_util {
 
-bool ShowItemInFolder(const base::FilePath& full_path) {
+void ShowItemInFolderOnWorkerThread(const base::FilePath& full_path) {
   base::ScopedBlockingCall scoped_blocking_call(base::BlockingType::MAY_BLOCK);
   base::win::ScopedCOMInitializer com_initializer;
   if (!com_initializer.Succeeded())
-    return false;
+    return;
 
   base::FilePath dir = full_path.DirName().AsEndingWithSeparator();
   // ParseDisplayName will fail if the directory is "C:", it must be "C:\\".
   if (dir.empty())
-    return false;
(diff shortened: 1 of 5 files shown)


deepak1556:
Oops missed that, yeah the main thread disallows blocking calls. We can't add [base::ScopedAllowBlockng](https://cs.chromium.org/chromium/src/base/threading/thread_restrictions.h?l=311) without patch to upstream. Best way forward is to move these tasks to a different thread, then again its not possible with synchronous api versions.

deepak1556:
> also whats the benefit of these two apis returning synchronous boolean results ? If it can be deprecated for async version, we could offload the task to a different thread.

@zcbenz what are your thoughts on this ?

zcbenz:
Deprecating sync versions looks good to me.

codebytere:
@zcbenz @deepak1556 sounds good to me! I'll start working on that :)

atom/common/platform_util_win.cc
```diff
@@ -231,63 +232,65 @@ HRESULT DeleteFileProgressSink::ResumeTimer() {
 
 namespace platform_util {
 
-bool ShowItemInFolder(const base::FilePath& full_path) {
+void ShowItemInFolderOnWorkerThread(const base::FilePath& full_path) {
```
deepak1556: Move this helper function to an anonymous namespace

nornagon:
NB. you can also write the line "BREAKING CHANGE" in the PR description to have it listed in that section of the release notes.

atom/common/platform_util_win.cc
Gam3356: Thank you guys! Also want to thank God for this financial blessing. Willie Parish

PR #17129, opened 2019-02-25
test: add missing checks to "validates process APIs access in sandboxed renderer"

miniak:
#### Description of Change
Add missing checks to "validates process APIs access in sandboxed renderer"

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/spec/api-browser-window-spec.js
+++ b/spec/api-browser-window-spec.js
@@ -10,7 +10,6 @@ const qs = require('querystring')
 const http = require('http')
 const { closeWindow } = require('./window-helpers')
 const { emittedOnce } = require('./events-helpers')
-const { resolveGetters } = require('./assert-helpers')
 const { ipcRenderer, remote } = require('electron')
 const { app, ipcMain, BrowserWindow, BrowserView, protocol, session, screen, webContents } = remote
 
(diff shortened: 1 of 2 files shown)


spec/fixtures/module/preload-sandbox.js
```diff
@@ -9,6 +9,15 @@
     window.process = process
     if (process.env.sandboxmain) {
       window.test = {
+        crash: typeof process.crash,
+        hang: typeof process.hang,
+        getCreationTime: typeof process.getCreationTime,
```
nornagon: getting the types of these doesn't validate that they actually work, which is IMO important for testing sandbox functions, since we've previously observed crashes that only occur when you actually call a function from the sandbox (b/c that's when the sandbox policy gets checked).

For `crash` and `hang` I'm willing to let it slide, but these other functions are easily testable.

miniak: done

miniak: @nornagon should I also check all the properties in the objects (as we do in `spec/api-process-spec.js`)?

nornagon: I'm fine with just checking that the function returned an appropriate object. as long as we call it!

spec/api-browser-window-spec.js
```diff
@@ -1923,14 +1922,23 @@ describe('BrowserWindow module', () => {
...
         ipcMain.once('answer', function (event, test) {
-          assert.strictEqual(test.arch, remote.process.arch)
-          assert.strictEqual(test.platform, remote.process.platform)
-          assert.deepStrictEqual(...resolveGetters(test.env, remote.process.env))
-          assert.strictEqual(test.execPath, remote.process.helperExecPath)
-          assert.strictEqual(test.sandboxed, true)
-          assert.strictEqual(test.type, 'renderer')
-          assert.strictEqual(test.version, remote.process.version)
-          assert.deepStrictEqual(test.versions, remote.process.versions)
+          expect(test.crash).to.equal('function')
```
MarshallOfSound: Looks like this is not exposed on linux?

miniak: should be, there is no ifdef in the native code. will try to figure out what's wrong

miniak: `process.getSystemMemoryInfo()` was throwing on Linux, which caused the whole `window.test` to be unavailable.

miniak:
@jkleinsc can you please check why the lint check is stuck?

MarshallOfSound:
@miniak CircleCI is experiencing issues at the moment with github hook processing, give it some time.

miniak:
@MarshallOfSound can you please review the fixed version?

spec/api-browser-window-spec.js
```diff
@@ -1923,14 +1922,30 @@ describe('BrowserWindow module', () => {
...
+          expect(test.arch).to.equal(remote.process.arch)
+          expect(test.platform).to.equal(remote.process.platform)
+          expect(test.env).to.deep.equal(remote.process.env)
+          expect(test.execPath).to.equal(remote.process.helperExecPath)
+          expect(test.sandboxed).to.be.true()
+          expect(test.type).to.equal('renderer')
+          expect(test.version).to.equal(remote.process.version)
+          expect(test.versions).to.deep.equal(remote.process.versions)
+
+          if (process.platform === 'linux' && test.osSandbox) {
```
MarshallOfSound: This behavior does not appear to be documented here:

https://github.com/electron/electron/blob/master/docs/api/process.md#processgetsystemmemoryinfo

I wouldn't expect it to fail in sandbox?

miniak: I can fix the docs in a follow up PR

PR #17135, opened 2019-02-26
refactor: make shell.OpenExternal async

codebytere:
#### Description of Change

Offload `OpenExternal` to new thread on Windows.

cc @deepak1556 @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


atom/common/platform_util_win.cc
```diff
@@ -295,24 +296,29 @@ bool OpenItem(const base::FilePath& full_path) {
...
-    // bug 1136923.
-    return false;
+    // On failure, it may be good to display a message to the user.
+    // https://crbug.com/727913
+    return;
   }
-  return true;
+}
+
+void OpenExternal(const GURL& url) {
```
deepak1556: Use already available `void OpenExternal(const base::string16&, const OpenExternalOptions&, OpenExternalCallback)` method, a new signature is unnecessary.

atom/common/platform_util_win.cc
```diff
@@ -295,24 +296,29 @@ bool OpenItem(const base::FilePath& full_path) {
     return ui::win::OpenFileViaShell(full_path);
 }
 
-bool OpenExternal(const base::string16& url,
-                  const OpenExternalOptions& options) {
+void OpenExternalOnWorkerThread(const base::string16& url,
```
deepak1556: Move this helper function to an anonymous namespace

deepak1556:
Do we want to support both `OpenExternalSync` and `OpenExternal` ? Its not possible to support sync version on windows with this change.

zcbenz:
We can either deprecate the sync API or just leave a fake sync API. I'm good with either, the async API was fake anyway.

atom/browser/atom_browser_client.cc
```diff
@@ -791,15 +791,16 @@ void AtomBrowserClient::RenderProcessExited(
   }
 }
 
+void OnOpenExternalFinished(const std::string& msg) {
+  LOG(INFO) << msg;
```
nornagon:
```suggestion
```

atom/browser/atom_browser_client.cc
```diff
@@ -791,15 +791,16 @@ void AtomBrowserClient::RenderProcessExited(
...
-#if defined(OS_WIN)
-        base::UTF8ToUTF16(escaped_url.spec()),
-#else
-        escaped_url,
-#endif
-        platform_util::OpenExternalOptions());
+  if (allowed) {
+    platform_util::OpenExternal(escaped_url,
+                                platform_util::OpenExternalOptions(),
+                                base::BindOnce(&OnOpenExternalFinished));
```
deepak1556: `base::NullCallback` maybe ?

atom/browser/atom_browser_client.cc
codebytere: should it just be empty then do you think?

atom/common/platform_util_win.cc
```diff
@@ -301,31 +320,16 @@ bool OpenItem(const base::FilePath& full_path) {
...
-}
-
-void OpenExternal(const base::string16& url,
+void OpenExternal(const GURL& url,
                   const OpenExternalOptions& options,
                   OpenExternalCallback callback) {
   // TODO(gabriel): Implement async open if callback is specified
-  std::move(callback).Run(OpenExternal(url, options) ? "" : "Failed to open");
+  std::move(callback).Run(
+      base::CreateCOMSTATaskRunnerWithTraits(
```
deepak1556: `callback` should be executed after the `OpenExternalOnWorkerThread` task is complete. Also use the return values from that task to depict errors in the callback accordingly.

atom/browser/atom_browser_client.cc
nornagon: if it does nothing, why have it?

deepak1556: I think it was intended to adjust to the openExternal api that needed a callback as argument. @codebytere you can use `base::NullCallback()` in such cases.

atom/browser/atom_browser_client.cc
codebytere: til that's a thing! ty :D

atom/common/platform_util_win.cc
codebytere: done!

atom/browser/atom_browser_client.cc
codebytere: yep! thanks :D

atom/common/platform_util_win.cc
```diff
@@ -230,6 +230,25 @@ HRESULT DeleteFileProgressSink::ResumeTimer() {
...
+  // parameters unexpected by the external program. This url should already
+  // have been escaped.
+  std::string escaped_url = url.spec();
+  auto working_dir = options.working_dir.value();
+
+  if (reinterpret_cast<ULONG_PTR>(
+          ShellExecuteW(nullptr, "open", escaped_url.c_str(), nullptr,
+                        working_dir.empty() ? nullptr : working_dir.c_str(),
+                        SW_SHOWNORMAL)) <= 32) {
+    std::move(callback).Run("Failed to open");
```
deepak1556: `callback` should be invoked on the main thread, this is a different sequence.

atom/common/platform_util_win.cc
```diff
@@ -301,31 +320,14 @@ bool OpenItem(const base::FilePath& full_path) {
...
-
-void OpenExternal(const base::string16& url,
+void OpenExternal(const GURL& url,
                   const OpenExternalOptions& options,
                   OpenExternalCallback callback) {
-  // TODO(gabriel): Implement async open if callback is specified
-  std::move(callback).Run(OpenExternal(url, options) ? "" : "Failed to open");
+  base::CreateCOMSTATaskRunnerWithTraits(
+      {base::MayBlock(), base::TaskPriority::USER_BLOCKING})
+      ->PostTaskReplyWithResult(
```
deepak1556: This should be

```
base::CreateCOMSTATaskRunnerWithTraits(
      {base::MayBlock(), base::TaskPriority::USER_BLOCKING})
      ->PostTaskAndReplyWithResult(FROM_HERE, base::BindOnce(&OpenExternalOnWorkerThread, url, options),
                                   std::move(callback)));
```

Also make the signature of `OpenExternalOnWorkerThread` to be `std::string OpenExternalOnWorkerThread(...)`

atom/common/platform_util_win.cc
```diff
@@ -301,31 +319,15 @@ bool OpenItem(const base::FilePath& full_path) {
...
                   const OpenExternalOptions& options,
                   OpenExternalCallback callback) {
-  // TODO(gabriel): Implement async open if callback is specified
-  std::move(callback).Run(OpenExternal(url, options) ? "" : "Failed to open");
+  std::string msg = base::CreateCOMSTATaskRunnerWithTraits(
+      {base::MayBlock(), base::TaskPriority::USER_BLOCKING})
+      ->PostTaskAndReplyWithResult(FROM_HERE, base::BindOnce(&OpenExternalOnWorkerThread, url, options),
+                                   std::move(callback)));
+
+  std::move(callback).Run(msg);
```
deepak1556: This line is not required. `callback` is now task that will be called with the result from `OpenExternalOnWorkerThread` in the previous call.

atom/common/platform_util_win.cc
```diff
@@ -301,31 +319,15 @@ bool OpenItem(const base::FilePath& full_path) {
...
-  return true;
-}
-
-void OpenExternal(const base::string16& url,
+void OpenExternal(const GURL& url,
                   const OpenExternalOptions& options,
                   OpenExternalCallback callback) {
-  // TODO(gabriel): Implement async open if callback is specified
-  std::move(callback).Run(OpenExternal(url, options) ? "" : "Failed to open");
+  std::string msg = base::CreateCOMSTATaskRunnerWithTraits(
```
deepak1556: There are no synchronous responses here.

codebytere:
@deepak1556 this should finally be ready for final review!

PR #17143, opened 2019-02-26
chore: update CODEOWNERS to reflect governance

nornagon:
#### Description of Change
The component-specific groups have been adding more friction than utility.

#### Release Notes

Notes: no-notes

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,6 +6,8 @@
 # Everything that falls through the cracks:
 * @electron/reviewers
 
+/patches/ @electron/wg-upgrades
+
 # directories
 /.github/ @electron/wg-admin
 /default_app/ @electron/wg-docs-tools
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -6,16 +6,11 @@
...
+/patches/ @electron/wg-upgrades
 
 # directories
-/.github/ @electron/electrocats
-/default_app/ @electron/docs
-/docs/ @electron/docs
+/.github/ @electron/wg-admin
+/default_app/ @electron/wg-docs-tools
+/docs/ @electron/wg-docs-tools
 /docs-translations/ @electron/i18n
```
MarshallOfSound: This can be removed, translations don't live in this repo any more

.github/CODEOWNERS
```diff
@@ -6,16 +6,11 @@
 # Everything that falls through the cracks:
 * @electron/reviewers
```
MarshallOfSound: This is a pet-peeve of mine, can we remove this catch-all, it means that every PR we end up "participating" in which makes it hard to filter notifications.

People should be able to review PRs without being mentioned like this

PR #17172, opened 2019-02-28
refactor: remove worker_context_will_destroy.patch

MarshallOfSound:
Chromium has a hook that is called at the same time as this patched in
API.

Notes: no-notes


MarshallOfSound:
Ah @deepak1556 thanks for catching that, I thought it was too easy 😆 Will work on upstreaming instead 👍

Issue #10363, opened 2017-08-26
Node macro `DISALLOW_COPY_AND_ASSIGN ` is incompatible with the Chromium code


alexeykuzmin:
`DISALLOW_COPY_AND_ASSIGN` is defined in both Node and Chromium code:

**Chromium** [base/macros.h](https://chromium.googlesource.com/chromium/src.git/+/61.0.3163.59/base/macros.h#23):
```cpp
// Put this in the declarations for a class to be uncopyable.
#define DISALLOW_COPY(TypeName) \
  TypeName(const TypeName&) = delete
  
// Put this in the declarations for a class to be unassignable.
#define DISALLOW_ASSIGN(TypeName) TypeName& operator=(const TypeName&) = delete

#define DISALLOW_COPY_AND_ASSIGN(TypeName) \
  DISALLOW_COPY(TypeName);                 \
  DISALLOW_ASSIGN(TypeName)
```
**Node** [src/util.h](https://github.com/nodejs/node/blob/v8.2.1/src/util.h#L90)
```cpp
#define DISALLOW_COPY_AND_ASSIGN(TypeName)                                    \
  void operator=(const TypeName&) = delete;                                   \
  void operator=(TypeName&&) = delete;                                        \
  TypeName(const TypeName&) = delete;                                         \
  TypeName(TypeName&&) = delete
```

In the Electron the Chromium macro is undefined in the [atom/common/node_includes.h](https://github.com/electron/electron/blob/v1.7.6/atom/common/node_includes.h#L23).

The problem is that the Chromium code, obviously, expects this macro to be the Chromium one, and usage of the Node definition leads to `functions that differ only in their return type cannot be overloaded` and `constructor cannot be redeclared` compilation errors if a class defines something similar to the Node macro contents.
Check [base/threading/sequence_local_storage_map.h](https://chromium.googlesource.com/chromium/src.git/+/61.0.3163.59/base/threading/sequence_local_storage_map.h#36) for example:
```cpp
ValueDestructorPair(ValueDestructorPair&& value_destructor_pair);

ValueDestructorPair& operator=(ValueDestructorPair&& value_destructor_pair);
```

@zcbenz, any thoughts how to avoid those errors?

/cc @tonyganch

zcbenz:
@alexeykuzmin My workaround is to put `#include "atom/common/node_includes.h"` after all other include directives, see `atom_renderer_client.cc` for example.

alexeykuzmin:
[...]

PR #17178, opened 2019-02-28
refactor: prevent node macros from overriding base

nornagon:
#### Description of Change
These redefinitions meant that any Electron file that had `#include "node_includes.h"` would be unable to use `CHECK`, `DCHECK`, etc. from `base/`. This change makes it so that Electron code will use Chromium's logging macros, instead of Node's, but keep Node using Node's logging macros.

`push_macro` and `pop_macro` are non-standard C++, but [are used by Chromium](https://chromium.googlesource.com/chromium/src/+/72.0.3626.84/ui/gl/init/gl_initializer_android.cc#29) occasionally.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/common/node_includes.h
+++ b/atom/common/node_includes.h
@@ -17,18 +17,32 @@
 // again. But we still need the tracing functions, so declaring them.
 #define SRC_TRACING_TRACE_EVENT_H_
 
+#pragma push_macro("ASSERT")
+#pragma push_macro("CHECK")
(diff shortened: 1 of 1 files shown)


deepak1556:
This will unblock fixing https://github.com/electron/electron/issues/10363 , the ordering of `atom/common/node_includes.h` won't be a problem now.

PR #17181, opened 2019-03-01
feat: promisify dialog.showCertificateTrustDialog()

codebytere:
#### Description of Change

Promisify `dialog.showCertificateTrustDialog()`.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Converted `dialog.showCertificateTrustDialog()` to return a Promise instead of taking a callback.

### DIFF
--- a/atom/browser/ui/certificate_trust_mac.mm
+++ b/atom/browser/ui/certificate_trust_mac.mm
@@ -15,7 +15,7 @@
 
 @interface TrustDelegate : NSObject {
  @private
-  certificate_trust::ShowTrustCallback callback_;
+  std::unique_ptr<atom::util::Promise> promise_;
   SFCertificateTrustPanel* panel_;
   scoped_refptr<net::X509Certificate> cert_;
   SecTrustRef trust_;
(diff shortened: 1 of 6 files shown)


atom/browser/ui/certificate_trust_win.cc
```diff
@@ -79,7 +80,8 @@ void ShowCertificateTrust(atom::NativeWindow* parent_window,
     CertFreeCertificateChain(chain_context);
   }
 
-  callback.Run();
+  promise.Resolve();
```
MarshallOfSound: Is this method is synchronous / blocking on windows?  Might be worth documenting / possibly making async somehow 🤔

atom/browser/ui/certificate_trust_mac.mm
```diff
@@ -15,20 +15,20 @@
...
   SecPolicyRef sec_policy_;
 }
 
-- (id)initWithCallback:(const certificate_trust::ShowTrustCallback&)callback
-                 panel:(SFCertificateTrustPanel*)panel
-                  cert:(const scoped_refptr<net::X509Certificate>&)cert
-                 trust:(SecTrustRef)trust
-             certChain:(CFArrayRef)certChain
-             secPolicy:(SecPolicyRef)secPolicy;
+- (id)initWithPromise:(const atom::util::CopyablePromise&)promise
```
zcbenz: I think it should be possible to pass a move-only `Promise` here?

codebytere: done!

atom/browser/ui/certificate_trust_mac.mm
```diff
@@ -47,14 +47,14 @@ - (void)dealloc {
...
-             secPolicy:(SecPolicyRef)secPolicy {
+- (id)initWithPromise:(atom::util::Promise)promise
+                panel:(SFCertificateTrustPanel*)panel
+                 cert:(const scoped_refptr<net::X509Certificate>&)cert
+                trust:(SecTrustRef)trust
+            certChain:(CFArrayRef)certChain
+            secPolicy:(SecPolicyRef)secPolicy {
   if ((self = [super init])) {
-    callback_ = callback;
+    promise_.reset(new atom::util::Promise(std::move(promise)));
```
zcbenz: There is no need keeping `promise_` in a `std::unique_ptr`. It is only needed when the stored promise may be empty, but in this class since the promise is guaranteed to be always available, it can just be stored as normal value.


```objective-c++
@interface TrustDelegate : NSObject {
 @private
  atom::util::Promise promise_;
}

promise_ = std::move(promise);
```

zcbenz: I just realized there is no constructor in objective-c classes so `std::unique_ptr` is the only way to go. This PR should be ready to go!

PR #17196, opened 2019-03-02
docs: add upstream doc links to default window

### LINKED ISSUE #14728: add upstream doc links to default window
when we run `electron` a default window opens that has chromium and nodejs versions at the top. would be really helpful if these would be links to the upstream docs, just like the links at the bottom.

old nodejs docs can be found here https://nodejs.org/docs/ but not sure where to find chromium html, css and javascript docs for a specific version.

dinaelhanan:
#### Description of Change
Added the link to the Node docs and Chromium docs to the default window (not sure if I linked the chromium docs correctly, please let me know/feel free to suggest changes if you know the correct way to do so)
Fixes #14728.

#### Checklist
(I'm new, so I apologize, but I have no clue what the first item means)
- [x] `npm test` passes

#### Release Notes
notes: no-notes

### DIFF
--- a/default_app/renderer.ts
+++ b/default_app/renderer.ts
@@ -27,7 +27,7 @@ function initialize () {
   }
 
   document.querySelector<HTMLAnchorElement>('.electron-version')!.innerText = `Electron v${process.versions.electron}`
-  document.querySelector<HTMLAnchorElement>('.chrome-version')!.innerText =`<a href="https://www.chromium.org/v${process.versions.chrome}">Chromium v${process.versions.chrome}</a>``
+  document.querySelector<HTMLAnchorElement>('.chrome-version')!.innerText =`<a href="https://www.chromium.org/v${process.versions.chrome}">Chromium v${process.versions.chrome}</a>`
   document.querySelector<HTMLAnchorElement>('.node-version')!.innerText = `<a href="https://nodejs.org/docs/v${process.versions.node}">Node v${process.versions.node}</a>`
   document.querySelector<HTMLAnchorElement>('.v8-version')!.innerText = `v8 v${process.versions.v8}`
   document.querySelector<HTMLAnchorElement>('.command-example')!.innerText = `${electronPath} path-to-app`
--- a/default_app/renderer.ts
+++ b/default_app/renderer.ts
@@ -27,7 +27,7 @@ function initialize () {
   }
 
   document.querySelector<HTMLAnchorElement>('.electron-version')!.innerText = `Electron v${process.versions.electron}`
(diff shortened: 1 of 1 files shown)


default_app/renderer.ts
```diff
@@ -27,8 +27,8 @@ function initialize () {
   }
 
   document.querySelector<HTMLAnchorElement>('.electron-version')!.innerText = `Electron v${process.versions.electron}`
-  document.querySelector<HTMLAnchorElement>('.chrome-version')!.innerText = `Chromium v${process.versions.chrome}`
-  document.querySelector<HTMLAnchorElement>('.node-version')!.innerText = `Node v${process.versions.node}`
+  document.querySelector<HTMLAnchorElement>('.chrome-version')!.innerText = `<a href="https://www.chromium.org/v${process.versions.chrome}">Chromium v${process.versions.chrome}</a>`
```
felixrieseberg: You probably want `innerHTML` here, `innerText` would encode/decode the "text" you give it:

```suggestion
  document.querySelector<HTMLAnchorElement>('.chrome-version')!.innerHTML = `<a href="https://www.chromium.org/v${process.versions.chrome}">Chromium v${process.versions.chrome}</a>`
```

default_app/renderer.ts
```diff
@@ -27,8 +27,8 @@ function initialize () {
...
+  document.querySelector<HTMLAnchorElement>('.node-version')!.innerText = `<a href="https://nodejs.org/docs/v${process.versions.node}">Node v${process.versions.node}</a>`
```
felixrieseberg: Same thing here: `innerHTML` is required. Also, I'm noticing that a direct link constructed this way leads to [a folder listing](https://nodejs.org/docs/v10.0.0/), so we probably want to add [an `/api` at the end](https://nodejs.org/docs/v10.0.0/api).

```suggestion
  document.querySelector<HTMLAnchorElement>('.node-version')!.innerHTML = `<a href="https://nodejs.org/docs/v${process.versions.node}/api">Node v${process.versions.node}</a>`
```

dinaelhanan: Sounds good 👍

default_app/renderer.ts
dinaelhanan: You're right, the link doesn't work. I think that in cases like this we should just link to the general docs URL. Especially since I don't believe chrome has version-specific documentation.

dinaelhanan:
@felixrieseberg thanks for the feedback! See my comment below on what I think should be done about the chrome docs:
https://github.com/electron/electron/pull/17196#discussion_r262279405

dinaelhanan:
@miniak ok then, I've closed it :)

Issue #15112, opened 2018-10-12
Make clipboard module work on Linux with disabled remote module


miniak:
**Is your feature request related to a problem? Please describe.**
The clipboard module depends on the remote module on Linux.

**Describe the solution you'd like**
It should be implemented using dedicated IPCs that are properly validated to avoid security issues.

**Describe alternatives you've considered**
The app can call the clipboard APIs in the main process in response to app-level IPCs from the renderer.

**Additional context**
Follow-up from https://github.com/electron/electron/pull/13028

PR #17200, opened 2019-03-02
refactor: implement clipboard APIs without the remote module on Linux

miniak:
#### Description of Change
Fixes https://github.com/electron/electron/issues/15112

After this change, the only remaining internal use of the `remote` module in Electron is `WebViewElement.prototype.getWebContents`, which is addressed in https://github.com/electron/electron/pull/17201.

This finally makes it possible to consider disabling the `remote` module by default to improve security.

#### Checklist
- [x] `npm test` passes

/cc @nornagon, @MarshallOfSound 

#### Release Notes
Notes: The `clipboard` module in renderers no longer depends on the `remote` module on Linux.


lib/common/value-utils.js
```diff
@@ -0,0 +1,51 @@
...
+
+const { nativeImage, NativeImage } = process.atomBinding('native_image')
+
+const objectMap = function (value, mapper) {
+  const object = {}
+  for (const key of Object.keys(value)) {
+    object[key] = mapper(value[key])
+  }
+  return object
+}
```
ckerr: The in-loop object lookups can be avoided with `Object.entries()`

```js
const objectMap = function (source, mapper) {
  const target = {}
  for (const [key, val] of Object.entries(source)) {
    target[key] = mapper(val)
  }
  return target
}
```

miniak: or even
```
const objectMap = function (object, mapper) {
  return Object.fromEntries(Object.entries(object).map(([ key, value ]) => [ key, mapper(value) ]))
}
```

lib/common/value-utils.js
```diff
@@ -0,0 +1,44 @@
+'use strict'
+
+const { nativeImage, NativeImage } = process.atomBinding('native_image')
+
+const objectMap = function (object, mapper) {
+  return Object.fromEntries(Object.entries(object).map(([ key, value ]) => [ key, mapper(value) ]))
```
MarshallOfSound: I know this was discussed in a thread below but I really don't like this, we should be playing [Code Golf](https://en.wikipedia.org/wiki/Code_golf) with our internal code.

I should be able to look at this function and **instantly** see what it does.  The sample that @ckerr wrote IMO is much more readable --> https://github.com/electron/electron/pull/17200/files#r261829857

ckerr: If `Object.fromEntries()` is desirable -- e.g., if it has some low-level v8 code that makes object-building faster than the code I suggested -- then we can dial the knob a little and have something performant but still readable:

```js
const objectMap = (mapper, source) => {
  const sourceEntries = Object.entries(source)
  const targetEntries = sourceEntries.map(([key, val]) => [key, mapper(val)])
  return Object.fromEntries(targetEntries)
}
```

ckerr: ...so I went to benchmark this and see if `Object.fromEntries()` contains any intrinsic advantage, and I'm seeing this:

```js
console.log({ 'process.version': process.version })

const mapObjectFromEntries = (mapper, source) => {
  const sourceEntries = Object.entries(source)
  const targetEntries = sourceEntries.map(([key, val]) => [key, mapper(val)])
  return Object.fromEntries(targetEntries)
}
```

yields:

```txt
{ 'process.version': 'v12.0.0-nightly20190302453ed05343' }
/tmp/foo.js:17
  return Object.fromEntries(targetEntries)
                ^

TypeError: Object.fromEntries is not a function
    at mapObjectFromEntries (/tmp/foo.js:17:17)
```

It appears that `Object.fromEntries()` is still bleeding-edge. Are we certain that this code actually works in Electron? And if not, how are all the tests passing -- are these paths covered in the specs?

miniak: It does work.

miniak: I can defunitel use the code from @ckerr if its more readable

alexeykuzmin: > It appears that Object.fromEntries() is still bleeding-edge. Are we certain that this code actually works in Electron?

@ckerr It was introduced in Chromium 73 behind a flag (enabled by default) https://www.chromestatus.com/feature/5747878282657792

miniak:
@nornagon, @MarshallOfSound can you please review?

lib/common/api/clipboard.js
```diff
@@ -1,20 +1,23 @@
...
-  // Read/write to find pasteboard over IPC since only main process is notified
-  // of changes
-  if (process.platform === 'darwin' && process.type === 'renderer') {
-    const ipcRendererUtils = require('@electron/internal/renderer/ipc-renderer-internal-utils')
-
-    clipboard.readFindText = (...args) => ipcRendererUtils.invokeSync('ELECTRON_BROWSER_CLIPBOARD_READ_FIND_TEXT', ...args)
-    clipboard.writeFindText = (...args) => ipcRendererUtils.invokeSync('ELECTRON_BROWSER_CLIPBOARD_WRITE_FIND_TEXT', ...args)
+if (process.type === 'renderer') {
+  const ipcRendererUtils = require('@electron/internal/renderer/ipc-renderer-internal-utils')
+  const makeMethod = function (method) {
```
nornagon:
```suggestion
  const makeRemoteMethod = function (method) {
```

lib/browser/ipc-main-internal-utils.ts
```diff
@@ -1,12 +1,13 @@
...
+import * as valueUtils from '@electron/internal/common/value-utils'
 
 type IPCHandler = (...args: any[]) => any
 
 const callHandler = async function (handler: IPCHandler, event: ElectronInternal.IpcMainInternalEvent, args: any[], reply: (args: any[]) => void) {
   try {
-    const result = await handler(event, ...args)
-    reply([null, result])
+    const result = await handler(event, ...valueUtils.deserialize(args))
+    reply([null, valueUtils.serialize(result)])
```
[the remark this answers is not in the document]
miniak: I will refactor it

miniak:
@nornagon comments addressed

atom/common/api/atom_api_native_image.cc
```diff
@@ -607,18 +607,26 @@ bool Converter<mate::Handle<atom::api::NativeImage>>::FromV8(
 
 namespace {
 
+using atom::api::NativeImage;
```
codebytere: slight preference towards minimizing diff by making this change in a later PR, as it's not necessary for the change this PR is purportedly making.

miniak:
@nornagon I've added a check for the buffer size. To create a huge image you have to pass a huge buffer to the main process, which will probably OOM the process on the IPC layer already.

atom/common/api/atom_api_native_image.cc
miniak: good point, but in this case I had to re-arrange the code anyway to expose both the class and the API, so it would not make the change much smaller

miniak:
@nornagon, @MarshallOfSound can you please have another look?

PR #17201, opened 2019-03-02
[wip] feat: make remote.getGuestWebContents() public

miniak:
#### Description of Change
- makes `remote.getGuestWebContents()` public to make the usage of the `remote` module explicit
- adds `<webview>.getGuestInstanceId()` to get the `guestInstanceId`
- deprecates the original `<webview>.getWebContents()`

#### Checklist
- [x] `npm test` passes

/cc @nornagon, @MarshallOfSound, @deepak1556 

#### Release Notes

Notes: Added `remote.getGuestWebContents()` / `<webview>.getGuestInstanceId()` to replace deprecated `<webview>.getWebContents()`.

### DIFF
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -487,7 +487,7 @@ Returns:
 * `webContents` [WebContents](web-contents.md)
 * `guestWebContents` [WebContents](web-contents.md)
 
-Emitted when `<webview>.getWebContents()` is called in the renderer process of `webContents`.
+Emitted when `remote.getGuestWebContents()` is called in the renderer process of `webContents`.
 Calling `event.preventDefault()` will prevent the object from being returned.
 Custom value can be returned by setting `event.returnValue`.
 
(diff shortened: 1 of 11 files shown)


miniak:
@nornagon the only other option is to just abandon this PR if we decide that the new API is too complicated to use. I can just make sure that the original API documents properly that it needs the remote module

nornagon:
I think we should definitely push forward with removing usage of the remote module. But I'd like some more input from the API design side of things.

Currently on the `remote` module:
- `remote.require(module)`
- `remote.getCurrentWindow()`
- `remote.getCurrentWebContents()`
- `remote.getGlobal(name)`

Since the main process is capable of getting any WebContents, would it make sense to have a function like `remote.getWebContentsForId(...)`? Is there an ID space that includes both guest and regular webcontents?

miniak:
@nornagon We already have `webContents.fromId(id)`, we just need `<webview>.getWebContentsId()`

miniak:
@nornagon seems like the code can be modified to actually use `webContents.id` as `clientInstanceId`. also the reason why we have `remote.getGuestWebContents()` is that it only allows getting `WebContents` for `<webview>` embedded in the given host, not `WebContents` for any other window, etc.

PR #17203, opened 2019-03-03
feat: add 'disableHtmlFullscreenWindowResize' option to webPreferences 

samuelmaddock:
#### Description of Change
Adds an option to `webPreferences` named `disableHtmlFullscreenWindowResize`. If set to `true`, the window will retain its size when using the HTML5 Fullscreen API. The requested HTML element will still fullscreen within the viewport though.

This is a fairly specialized use case for my application, [Metastream](https://github.com/samuelmaddock/metastream). I'd like to be able to trigger HTML5 fullscreen without causing the window to resize. An example can be seen in the following GIF.

html5-fullscreen

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `disableHtmlFullscreenWindowResize` option to `webPreferences` to allow preventing window from resizing when entering HTML5 fullscreen.

### DIFF
--- a/atom/browser/common_web_contents_delegate.cc
+++ b/atom/browser/common_web_contents_delegate.cc
@@ -616,7 +616,17 @@ void CommonWebContentsDelegate::SetHtmlApiFullscreen(bool enter_fullscreen) {
     return;
   }
 
-  owner_window_->SetFullScreen(enter_fullscreen);
+  // Set fullscreen on window if allowed.
+  auto* web_preferences = WebContentsPreferences::From(GetWebContents());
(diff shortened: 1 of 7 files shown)


samuelmaddock:
> Calling `preventDefault` on `will-enter-html-full-screen` doesn't make a whole log of sense to not cancel the full screen, rather just change it's behavior.

Agreed.

> I'd rather add a flag in `webPreferences` to change the behavior of this for the entire `webContents` (which would appear to fit your use case.

@MarshallOfSound I've implemented these changes, how'd I do?

PR #17218, opened 2019-03-04
fix: correct chrome version in chrome_version.h

nornagon:
In #15275 we switched to using Chrome's auto-generated version headers instead
of hard-coding the Chrome version into this header, but that change wasn't
backported to 4-0-x and we've since rolled our Chrome dependency. This fixes
the discrepancy. We don't expect any further Chrome rolls on this branch since
M69 is out of Chrome's support window.

- [ ] `npm test` passes

Notes: Fixed `process.versions.chrome` reporting the wrong patch version.


nornagon:
cc @bpasero re #16636

bpasero:
Thanks!

PR #17221, opened 2019-03-04
fix: make StreamSubscriber ref counted

### LINKED ISSUE #16881: Electron 4.x freezing issue (possible webview/custom-stream-protocol connection)
* **Electron Version**: 4.0.4
* **Operating System** (Platform and Version): MacOS 10.14.3, but it's been reported by my users on Windows and Linux
* **Last known working Electron version** (if applicable): 3.x

Related issue: https://github.com/beakerbrowser/beaker/issues/1333

We started to experience renderer-process freeze-crashes once Beaker was upgraded to v4. When the freeze occurs, the program becomes unresponsive and has to be force-quit.

The freeze only seems to occur when visiting custom stream-protocol resources, and has some randomness to it. The best way I've found to reproduce the issue is to repeatedly refresh a webview which is pointed at a custom stream-protocol resource, which is what I've provided below.

### To Reproduce

https://github.com/pfrazee/electron-bug-webview-freeze

```bash
# Clone this repository
git clone https://github.com/pfrazee/electron-bug-webview-freeze
# Go into the repository
cd electron-bug-webview-freeze
# Install dependencies
npm install
# Run the app
npm start
```

The renderer has a webview embedded with a red border. It is viewing a custom stream-protocol resource (`custom://foo/index.html`).

Once a second, the webview is refreshed. A text-line will inform you of the last refresh time.

Eventually, the renderer process will freeze. It's not always consistent in the amount of time, so you may need to leave it running for a while.

deepak1556:
#### Description of Change

It is owned by URLRequestStreamJob on the IO thread once request starts, but if the ownership was abondoned while transfering it to IO thread which is possible when a request is aborted, then we need to make sure its destroyed on the right thread.

Fixes #16881 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fix app freeze when using custom stream protocol

### DIFF
--- a/atom/browser/api/stream_subscriber.cc
+++ b/atom/browser/api/stream_subscriber.cc
@@ -19,12 +19,15 @@ namespace mate {
 StreamSubscriber::StreamSubscriber(
     v8::Isolate* isolate,
     v8::Local<v8::Object> emitter,
-    base::WeakPtr<atom::URLRequestStreamJob> url_job)
-    : isolate_(isolate),
+    base::WeakPtr<atom::URLRequestStreamJob> url_job,
+    scoped_refptr<base::SequencedTaskRunner> ui_task_runner)
+    : base::RefCountedDeleteOnSequence<StreamSubscriber>(ui_task_runner),
+      isolate_(isolate),
       emitter_(isolate, emitter),
       url_job_(url_job),
       weak_factory_(this) {
-  DCHECK_CURRENTLY_ON(content::BrowserThread::UI);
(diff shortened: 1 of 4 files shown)


pfrazee:
Will test this shortly, I'm just fighting with the electron build process

pfrazee:
Just tested this against https://github.com/pfrazee/electron-bug-webview-freeze and confirmed the fix works 👍

deepak1556:
Manually backported to 4-0-x in https://github.com/electron/electron/pull/17267

PR #17224, opened 2019-03-04
fix: remove usage of private APIs in the MAS build

### LINKED ISSUE #17197: Mac App Store rejects the build with 5.0.0-beta*
* **Electron Version:** 
  * 5.0.0-beta.4

### Rejection Email
Guideline 2.5.1 - Performance - Software Requirements

Your app uses or references the following non-public APIs:

Found private symbol usage.
Symbol: __LSSetApplicationLaunchServicesServerConnectionStatus
From framework: /System/Library/Frameworks/CoreServices.framework/Versions/A/CoreServices
Symbol: _SetApplicationIsDaemon
From framework: /System/Library/Frameworks/ApplicationServices.framework/Versions/A/ApplicationServices
Symbols: _pthread_chdir_np, _pthread_fchdir_np, _abort_report_np
From framework: /usr/lib/libSystem.B.dylib
In binary: Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework

The use of non-public APIs is not permitted on the App Store, because it can lead to a poor user experience should these APIs change.

We are constantly reevaluating and identifying non-public APIs that you may have been using for an extended period of time. You should always use public APIs and frameworks and ensure they are up-to-date to prevent this issue in the future

### Additional Information
Version 4 has no problems.

MarshallOfSound:
#### Description of Change
This PR introduces patches to remove / replace usages of private APIs used in the MAS build.

Please note this PR includes 3 separate patches each solving their own problem, it would probably make sense to rebase and merge this rather than squash and merge in order to retain history.

Fixes #17197 

#### Release Notes

Notes: Fixed some MAS rejection errors by removing usages of private APIs


patches/common/chromium/fix_disable_usage_of_pthread_fchdir_np_and_pthread_chdir_np_in_mas.patch
```diff
@@ -0,0 +1,54 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <sattard@slack-corp.com>
+Date: Mon, 4 Mar 2019 14:46:48 -0800
+Subject: fix: disable usage of pthread_fchdir_np and pthread_chdir_np in MAS
+ builds
```
nornagon: hrrrrm is this what `export-patches` produced? the 2-line subject line is probably gonna cause issues

patches/common/chromium/fix_disable_usage_of_setapplicationisdaemon_and.patch
```diff
@@ -0,0 +1,52 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <sattard@slack-corp.com>
+Date: Mon, 4 Mar 2019 14:51:45 -0800
+Subject: fix: disable usage of SetApplicationIsDaemon and
+ _LSSetApplicationLaunchServicesServerConnectionStatus in MAS builds
```
nornagon: here too

patches/common/chromium/fix_disable_usage_of_pthread_fchdir_np_and_pthread_chdir_np_in_mas.patch
MarshallOfSound: Yup, this is exactly what the script produced.  It looks like the builds are OK and handle it though?

nornagon: hmmm... does it survive `import-patches && export-patches`? if not, might have to kick the `export-patches` script

deepak1556:
@nornagon we already have disabled seatbelt for MAS builds, wouldn't it better if we add dummy build groups for targets in `//sandbox/mac` under `is_mas_build` which will also achieve the same purpose but simplify the patch set we have.

nornagon:
@deepak1556 would we have to provide stubs for functions/classes that other modules depend on? If so, that seems like it might not actually be easier, since we'd have to track that API.

codebytere:
@MarshallOfSound 
```
/home/builduser/project/src/electron/lib/renderer/web-view/web-view-attributes.ts
  58:29  error  '_args' is defined but never used  @typescript-eslint/no-unused-vars

/home/builduser/project/src/electron/lib/renderer/web-view/web-view-element.ts
  12:10  error  'WebViewImpl' is defined but never used  @typescript-eslint/no-unused-vars
```

MarshallOfSound:
@codebytere Rebased on master to fix those issues 👍

PR #17239, opened 2019-03-06
build: copy generated electron.d.ts file in a cross-platform way

MadfishDT:
#### Description of Change
 in window os, electron.d.ts copy method is not working. because script in package.json use 'cp' command. so i add new script "gn-typescript-definitions-win": "npm run create-typescript-definitions && copy electron.d.ts" and BUILD.gn modify for window build

#### Checklist

- [x ] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

#### Release Notes

Notes: no-notes

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -59,24 +59,14 @@ config("branding") {
 # The one in $target_gen_dir is used for the actual TSC build later one
 # and the one in //electron/electron.d.ts is used by your IDE (vscode)
 # for typescript prompting
-if (is_win) {
-  npm_action("build_electron_definitions") {
-    script = "gn-typescript-definitions-win"
-    args = [ rebase_path("$target_gen_dir/tsc/typings/electron.d.ts") ]
-    inputs = auto_filenames.api_docs + [ "package-lock.json" ]
-    outputs = [
-      "$target_gen_dir/tsc/typings/electron.d.ts",
-    ]
-  }
-} else {
-  npm_action("build_electron_definitions") {
-    script = "gn-typescript-definitions"
-    args = [ rebase_path("$target_gen_dir/tsc/typings/electron.d.ts") ]
-    inputs = auto_filenames.api_docs + [ "package-lock.json" ]
-    outputs = [
-      "$target_gen_dir/tsc/typings/electron.d.ts",
-    ]
-  }
+npm_action("build_electron_definitions") {
+  script = "gn-typescript-definitions"
+  args = [ rebase_path("$target_gen_dir/tsc/typings/electron.d.ts") ]
+  inputs = auto_filenames.api_docs + [ "package-lock.json" ]
+
+  outputs = [
+    "$target_gen_dir/tsc/typings/electron.d.ts",
+  ]
 }
 
 npm_action("atom_browserify_sandbox") {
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -4622,8 +4622,7 @@
           "version": "2.1.1",
           "resolved": false,
           "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
-          "dev": true,
-          "optional": true
+          "dev": true
```
MarshallOfSound: The package lock changes should not be committed for this change

BUILD.gn
```diff
@@ -59,14 +59,24 @@ config("branding") {
...
-  script = "gn-typescript-definitions"
-  args = [ rebase_path("$target_gen_dir/tsc/typings/electron.d.ts") ]
-  inputs = auto_filenames.api_docs + [ "package-lock.json" ]
-
-  outputs = [
-    "$target_gen_dir/tsc/typings/electron.d.ts",
-  ]
+if (is_win) {
+  npm_action("build_electron_definitions") {
+    script = "gn-typescript-definitions-win"
```
MarshallOfSound: Only the `script = ""` needs to be wrapped in the `if (is_win)` (It's the only thing that needs to change, no need to duplicate the entire block)

package.json
```diff
@@ -61,6 +61,7 @@
     "lint:js-in-markdown": "standard-markdown docs",
     "create-api-json": "electron-docs-linter docs --outfile=electron-api.json",
     "create-typescript-definitions": "npm run create-api-json && electron-typescript-definitions --in=electron-api.json --out=electron.d.ts && node spec/ts-smoke/runner.js",
+    "gn-typescript-definitions-win": "npm run create-typescript-definitions && copy electron.d.ts",
```
MarshallOfSound: Instead of duplicating this I'd rather we use something like [`shx`](https://www.npmjs.com/package/shx) as a dev dep to provide `cp` on both platforms

MadfishDT: thanks your review I got it. I agree your opinion, 
I think, your solution should add in build instructions doc

deepak1556: @MarshallOfSound instead of doing the copy here, why not add a gn copy dep https://gn.neocities.org/#copy . It will work across platforms.

MadfishDT: shx not work in window~ I tried install npm -g shx. after that, I run build electron, but build make same error

MarshallOfSound: @deepak1556 this is a strange hack that is required for the TS build.  We went through this when it was initially added. A gn action output must the be in out directory but the TS definitions need to be in the electron directory.  So we gen them in the electron directory and copy them out to trick gn into thinking an actual output item was made

MarshallOfSound: @MadfishDT check out the usage instructions for `shx`.  You don't use `cp` you use `shx cp`

deepak1556: Is it not possible to do the reverse ? Use the action output in the gen directory and copy them to electron directory by using `copy() {}`

MarshallOfSound: It's also against the rules to use GN copy to copy something to a somewhere not in the out directory.  GN doesn't support this use case so we had to hack around it

deepak1556: Thanks for the clarification!

MadfishDT: thanks~ all

nornagon:
It looks like this PR is now a no-op? @MadfishDT do you want to update this to use `shx`?

PR #17247, opened 2019-03-06
refactor: load electron builtin modules with process._linkedBinding

deepak1556:
#### Description of Change

`NODE_BUILTIN_MODULE_CONTEXT_AWARE` and `process.binding` are removed in nodejs/node#25829. This change adopts the alternative available without any functionality change.

Relevant electron/node PR: https://github.com/electron/node/pull/95

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -12,7 +12,7 @@ vars = {
   'chromium_version':
     '73.0.3683.68',
   'node_version':
-    '70a78f07b1c4d53f3da462b08cef42a4ff8f949f',
+    '5e32b02e3c180c9997d60fe85042d335b6d9a588',
 
   'boto_version': 'f7574aa6cc2c819430c1f05e9a1a1a666ef8169b',
   'pyyaml_version': '3.12',
--- a/atom/browser/api/atom_api_app.cc
+++ b/atom/browser/api/atom_api_app.cc
@@ -1425,4 +1425,4 @@ void Initialize(v8::Local<v8::Object> exports,
 
 }  // namespace
 
-NODE_BUILTIN_MODULE_CONTEXT_AWARE(atom_browser_app, Initialize)
+NODE_LINKED_MODULE_CONTEXT_AWARE(atom_browser_app, Initialize)
--- a/atom/browser/api/atom_api_auto_updater.cc
+++ b/atom/browser/api/atom_api_auto_updater.cc
@@ -154,4 +154,4 @@ void Initialize(v8::Local<v8::Object> exports,
 
 }  // namespace
 
-NODE_BUILTIN_MODULE_CONTEXT_AWARE(atom_browser_auto_updater, Initialize)
+NODE_LINKED_MODULE_CONTEXT_AWARE(atom_browser_auto_updater, Initialize)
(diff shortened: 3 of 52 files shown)


deepak1556:
Preferred explicit module registration over `__attribute__((constructor))` macro, since the latter doesn't work well with static library config. Also having an explicit registration is better in combination with `BUILDFLAG` for some internal modules.

deepak1556:
> although it might still be nice if we could ultimately end up using public Node.js APIs for this

The public api ends up defining the module registrators with `__attribute__((constructor))` which doesn't work with the static library that is built in electron https://github.com/electron/electron/blob/master/BUILD.gn#L321. If referencing the constructors in the executable is the solution, I would prefer the manual registration over it. Similar to https://github.com/nodejs/node/pull/16565

addaleax:
@deepak1556 I think it might just be an issue of copying/inlinine `NODE_MODULE_CONTEXT_AWARE_CPP()` into the Electron source rather than relying on the macro from `node_binding.h`, and maybe switching `NM_F_LINKED` to `node::ModuleFlags::kLinked` once https://github.com/nodejs/node/pull/26457 makes its way into your Node.js fork?

The layout of the `node_module` struct itself is guaranteed to remain stable during a major Node.js release, so I think that should be okay.

deepak1556:
Ah okay I misunderstood the public api switch then!

>  I think it might just be an issue of copying/inlinine NODE_MODULE_CONTEXT_AWARE_CPP() into the Electron source rather than relying on the macro from node_binding.h, and maybe switching NM_F_LINKED to node::ModuleFlags::kLinked
> The layout of the node_module struct itself is guaranteed to remain stable during a major Node.js release

Yeah this should be good thing to do, given the guarantee. Thanks!

MarshallOfSound:
The linux tests failed here with

> [401:0307/234309.568071:FATAL:zygote_host_impl_linux.cc(116)] No usable sandbox! Update your kernel or see https://chromium.googlesource.com/chromium/src/+/master/docs/linux_suid_sandbox_development.md for more information on developing with the SUID sandbox. If you want to live dangerously and need an immediate workaround, you can try using --no-sandbox.

Which is confusing 🤔 

cc @nornagon

nornagon:
afaict that's a flake. I don't know what's causing it, perhaps some change in circleci's internal architecture? I've tried rerunning the test, let's see how it goes.

MarshallOfSound:
@nornagon As part of the setuid PR will the setuid sandbox become usable by the tests?  Would be good to see that flake disappear as a result of that work if it is actually a flake.

nornagon:
i'm a little worried about why it only happens sometimes, it shouldn't be flaky.

MarshallOfSound:
@nornagon If we're happy the flake isn't somehow _caused_ by this PR I think we're good to merge it now

nornagon:
i'm pretty sure i've seen it in other PRs and i can't imagine this would cause such a flake so i'm pretty sure we're good to go.

PR #17249, opened 2019-03-06
feat: promisify session.clearStorageData()

codebytere:
#### Description of Change

This PR promisifies `session.clearStorageData()`. I removed the synchronous version, since i think it's not altogether necessary and simplifies the overall changes that need to be made.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Converted `session.clearStorageData()` such that it returns a Promise instead of taking a callback.


atom/browser/api/atom_api_session.cc
```diff
@@ -429,12 +424,17 @@ void Session::DoCacheAction(const net::CompletionCallback& callback) {
                      action, callback));
 }
 
-void Session::ClearStorageData(mate::Arguments* args) {
-  // clearStorageData([options, callback])
+void ResolvePromise(const util::CopyablePromise& promise) {
```
miniak: use the new helpers?

atom/browser/api/atom_api_session.cc
```diff
@@ -443,9 +443,12 @@ void Session::ClearStorageData(mate::Arguments* args) {
...
     MediaDeviceIDSalt::Reset(browser_context()->prefs());
   }
-  storage_partition->ClearData(options.storage_types, options.quota_types,
-                               options.origin, base::Time(), base::Time::Max(),
-                               base::Bind(&OnClearStorageDataDone, callback));
+
+  storage_partition->ClearData(
+      options.storage_types, options.quota_types, options.origin, base::Time(),
+      base::Time::Max(),
+      base::Bind(&ResolvePromise, util::CopyablePromise(promise)));
```
miniak: why not move the promise here?

codebytere: @miniak `ClearData` only accepts a `RepeatingCallback` with `base::Bind`, and Promises are move-only so we can't move it and thus need to use CopyablePromises

atom/browser/api/atom_api_session.cc
```diff
@@ -429,12 +424,13 @@ void Session::DoCacheAction(const net::CompletionCallback& callback) {
                      action, callback));
 }
 
-void Session::ClearStorageData(mate::Arguments* args) {
-  // clearStorageData([options, callback])
+v8::Local<v8::Promise> Session::ClearStorageData(mate::Arguments* args) {
+  v8::Isolate* isolate = v8::Isolate::GetCurrent();
+  util::Promise promise(isolate);
```
deepak1556: Prefer getting the isolate from simple getters when available, for instance here it can be from `args->isolate()` rather than using `v8::Isolate::GetCurrent()` which tries to obtain it from a TLS.

atom/browser/api/atom_api_session.cc
miniak: gotcha

PR #17258, opened 2019-03-07
feat: add support for content scripts 'all_frames' option

samuelmaddock:
#### Description of Change

Adds support for Chrome extension content script option [all_frames](https://developer.chrome.com/extensions/content_scripts#frames).

> The "all_frames" field allows the extension to specify if JavaScript and CSS files should be injected into all frames matching the specified URL requirements or only into the topmost frame in a tab.

> Optional. Defaults to false, meaning that only the top frame is matched. 
> 
> If specified true, it will inject into all frames, even if the frame is not the topmost frame in the tab. Each frame is checked independently for URL requirements, it will not inject into child frames if the URL requirements are not met.

These changes will only work with the non-sandboxed renderer until #16218 is merged.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `all_frames` support for Chrome extension content scripts.


erez12:
@samuelmaddock on Electron 6.0.0-nightly.20190311 i cant get my extension to run on all frames. 
I have the following setup: 

Extension manifest:
```
{
	"name": "content-script-test",
	"version": "1.0",
	"content_scripts": [{
		"matches": ["<all_urls>"],
		"js": [
			"test_working_content_document_start.js"
		],
		"run_at": "document_start",
		"all_frames": true
	}],
	"manifest_version": 2
}
```
test_working_content_document_start.js:
`console.log(window.location.href);`

Electron main:
`const { BrowserWindow } = require('electron');
BrowserWindow.addExtension('../my_extension_folder');

let win = new BrowserWindow({
     width: 1600,
     height: 1000,
        
});

win.loadURL('https://www.w3schools.com/html/html_iframe.asp')`

On chrome im getting multiple logs of the iframe urls. On electron im only getting printout of the top frame. 

had a look at nightly#6.0.0-nightly.20190311 source (https://github.com/electron/electron/tree/b43e4b8e0e93eedd466a2794e02e0452f27407d9) and it seems that both this #17258 and #16218 are there. 

What am i missing ?

samuelmaddock:
@erez12: try adding the [`nodeIntegrationInSubFrames`](https://github.com/electron/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions) property to your `webPreferences`. It's required to enable script injection in subframes. The work in #17440 will eventually deprecate the need for this.

erez12:
@samuelmaddock Great! That solved the issue but raised more issues...
Now when i send a message from content script to background i dont get the frameId:

content.js:
`chrome.runtime.sendMessage({topic: "page_load"})`

background.js:
```
chrome.runtime.onMessage.addListener(function(request, sender, sendResponse) {
    if ( request.topic === "page_load") {
        console.log(sender.frameId) // This is undefined
    }
});
```

Since there is a known issue in Electron with sendResponse not working, i wanted to use the tabs API to respond to the sender but i can't send the message to a specific frame...

Seems there is more stuff to do for full extensions support. Is there any time frame for #17440 ?

PR #17269, opened 2019-03-07
feat: enable setuid sandbox on linux

### LINKED ISSUE #16631: electron@5.0.0-beta.1 fails with error: No usable sandbox!
electron: 5.0.0-beta.1
platform: linux x64

**Actual behavior**

https://github.com/electron/electron-quick-start app fails to start with the following error:
```
[3937:0131/003107.790938:FATAL:zygote_host_impl_linux.cc(116)] No usable sandbox! Update your kernel or see https://chromium.googlesource.com/chromium/src/+/master/docs/linux_suid_sandbox_development.md for more information on developing with the SUID sandbox. If you want to live dangerously and need an immediate workaround, you can try using --no-sandbox.
```

**To Reproduce**

- `git clone https://github.com/electron/electron-quick-start.git`
- `cd electron-quick-start`
- `npm install`
- `npm add --only=dev electron@5.0.0-beta.1`
- any of the following commands fail:
  - `npm start`
  - `node_modules/.bin/electron --version`

**Additional Information**

Worth to mention that event just `node_modules/.bin/electron --version` execution fails with the same error.

CC @nornagon

nornagon:
#### Description of Change
This change enables the [setuid sandbox](https://chromium.googlesource.com/chromium/src/+/73.0.3683.65/docs/linux_suid_sandbox.md) on Linux. This allows Electron to run sandboxed in environments that disable CLONE_NEWUSER for unprivileged users (e.g. docker without `CAP_SYS_ADMIN`, and Arch Linux).

Closes #16631.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Enabled the setuid sandbox on Linux, allowing Electron to launch sandboxed processes in environments that disable CLONE_NEWUSER for unprivileged users.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -1026,6 +1026,9 @@ dist_zip("electron_dist_zip") {
     ":licenses",
     ":electron_version",
   ]
+  if (is_linux) {
+    data_deps += [ "//sandbox/linux:chrome_sandbox" ]
+  }
   outputs = [
     "$root_build_dir/dist.zip",
   ]
--- a/atom/app/atom_main_delegate.cc
+++ b/atom/app/atom_main_delegate.cc
@@ -237,10 +237,6 @@ void AtomMainDelegate::PreSandboxStartup() {
   if (!IsBrowserProcess(command_line))
     return;
 
-  // Disable setuid sandbox since it is not longer required on
-  // linux (namespace sandbox is available on most distros).
-  command_line->AppendSwitch(service_manager::switches::kDisableSetuidSandbox);
-
   // Allow file:// URIs to read other file:// URIs by default.
   command_line->AppendSwitch(::switches::kAllowFileAccessFromFiles);
 
--- a/build/zip.py
+++ b/build/zip.py
@@ -61,7 +61,10 @@ def main(argv):
             for file in files:
               z.write(os.path.join(root, file))
         else:
-          z.write(dep)
+          basename = os.path.basename(dep)
+          dirname = os.path.dirname(dep)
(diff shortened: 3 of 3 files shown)


nornagon:
This is a draft PR because it doesn't yet address the distribution side of this change. Specifically, in order to work correctly, the `chrome_sandbox` executable must be:
- named `chrome-sandbox`, in the same directory as the `electron` executable,
- owned by `root`, and
- have permission mask `4755` set.

If the above conditions are not met, the sandbox will not function. Chrome solves this problem during installation: the `deb` file distribution has those permissions encoded. We'll need to do something similar for both `npm install`ed versions of Electron and versions that are packaged for distribution.

nornagon:
Ref. https://github.com/electron-userland/electron-installer-debian/pull/184

nornagon:
Also Ref. https://github.com/electron-userland/electron-installer-redhat/pull/112

nornagon:
Requesting review from @electron/wg-security and @zcbenz

deepak1556:
> Is it possible to support renaming of this binary to something static like electron_sandbox

Its hardcoded in upstream `SetuidSandboxHost::GetSandboxBinaryPath()`, will be a patch.

nornagon:
it would be relatively easy but questionably valuable to rename the binary to something like `electron-sandbox`. it would be an unreasonable amount of work to make it dynamic.

MarshallOfSound:
If it requires a patch to make it `electron` and too much work + patches to make it dynamic then I'm 👍 on this change.  Worth asking the question 😄

malept:
Can we get this backported to 5-0-x?

PR #17274, opened 2019-03-07
fix: gn check errors in release build

deepak1556:
#### Description of Change

```
ERROR at //electron/atom/browser/atom_browser_client.cc:74:11: Can't include this header from here.
#include "services/device/public/cpp/geolocation/location_provider.h"
          ^---------------------------------------------------------
The target:
  //electron:electron_lib
  //services/device/public/cpp/geolocation:geolocation
to include headers.
  //electron:electron_lib -->
  //content/public/browser:browser -->
  //content/public/browser:browser_sources --[private]-->
  //services/device/public/cpp/geolocation:geolocation

```

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -358,6 +358,7 @@ static_library("electron_lib") {
     "//ppapi/shared_impl",
     "//printing/buildflags",
     "//services/audio/public/mojom:constants",
+    "//services/device/public/cpp/geolocation",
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
Bypassing `new-pr` to unblock nightly release

PR #17291, opened 2019-03-08
fix: improve focused menu bar item visibility

brenca:
#### Description of Change

It looks like during the C73 upgrade we lost the highlight on the focused menu bar items (there was a small dotted border around the text of the focused item). This is bad because you can now navigate the menu with just the keyboard, and you can't tell which top level item you are currently on.

This PR enables the background highlighting of the `InkDrop` for the focused item, giving some visual feedback to the user when navigating with the keyboard. Also, I made the highlight a tiny bit darker to make it easier to see.

This is what we had before the C73 upgrade:
electron-acc-4
After the upgrade:
electron-acc-2
With this PR:
electron-acc-3

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Improved the visibility of the focused item on the menu bar.

### DIFF
--- a/atom/browser/ui/views/submenu_button.cc
+++ b/atom/browser/ui/views/submenu_button.cc
@@ -37,7 +37,7 @@ SubmenuButton::SubmenuButton(const base::string16& title,
 
   SetInkDropMode(InkDropMode::ON);
   set_ink_drop_base_color(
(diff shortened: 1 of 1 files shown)


codebytere:
@brenca would you mind including a before and after screenshot?

brenca:
@codebytere Created some GIFs to illustrate the problem

PR #17298, opened 2019-03-08
feat: promisify dialog.showMessageBox() 

codebytere:
#### Description of Change

Promisify `dialog.showMessageBox()` by splitting it into a sync and async method.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Split `dialog.showMessageBox()` into a synchronous version and a version that returns a Promise.

### DIFF
--- a/atom/browser/api/atom_api_dialog.cc
+++ b/atom/browser/api/atom_api_dialog.cc
@@ -40,16 +40,15 @@ int ShowMessageBoxSync(int type,
       cancel_id, options, title, message, detail, icon);
 }
 
-void ResolvePromiseObject(const atom::util::CopyablePromise& promise,
+void ResolvePromiseObject(atom::util::Promise promise,
                           int result,
                           bool checkbox_checked) {
-  mate::Dictionary dict =
-      mate::Dictionary::CreateEmpty(promise.GetPromise().isolate());
+  mate::Dictionary dict = mate::Dictionary::CreateEmpty(promise.isolate());
 
   dict.Set("response", result);
   dict.Set("checkboxChecked", checkbox_checked);
 
-  promise.GetPromise().Resolve(dict.GetHandle());
+  promise.Resolve(dict.GetHandle());
 }
 
 v8::Local<v8::Promise> ShowMessageBox(int type,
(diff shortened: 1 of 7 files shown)


docs/api/dialog.md
```diff
@@ -247,21 +247,61 @@ expanding and collapsing the dialog.
...
 It returns the index of the clicked button.
 
 The `browserWindow` argument allows the dialog to attach itself to a parent window, making it modal.
 
-If the `callback` and `browserWindow` arguments are passed, the dialog will not block the process. The API call
-will be asynchronous and the result will be passed via `callback(response)`.
+### `dialog.showMessageBox([browserWindow, ]options)`
+
+* `browserWindow` [BrowserWindow](browser-window.md) (optional)
+* `options` Object
```
MarshallOfSound: As a future task we should find a way to de-dupe this documentation, having the same object documented twice is going to cause divergence issues in the future

atom/browser/api/atom_api_dialog.cc
```diff
@@ -23,33 +23,59 @@
...
+                       const std::string& checkbox_label,
+                       bool checkbox_checked,
+                       const gfx::ImageSkia& icon,
+                       atom::NativeWindow* window) {
+  return atom::ShowMessageBoxSync(
+      window, static_cast<atom::MessageBoxType>(type), buttons, default_id,
+      cancel_id, options, title, message, detail, icon);
+}
+
+void ResolvePromiseObject(const atom::util::CopyablePromise& promise,
```
zcbenz: `CopyablePromise` should only be used when there is no other choice, such as passing it to some Chromium APIs.

For `ShowMessageBox`, we should change its callback type to `base::OnceCallback`, and then use `base::BindOnce` and move-only `Promise` for callbacks.

atom/browser/ui/message_box_win.cc
```diff
@@ -220,28 +219,29 @@ void RunMessageBoxInNewThread(base::Thread* thread,
...
-                              const MessageBoxCallback& callback) {
+                              MessageBoxCallback callback) {
   int result = ShowTaskDialogUTF8(parent, type, buttons, default_id, cancel_id,
                                   options, title, message, detail,
                                   checkbox_label, &checkbox_checked, icon);
-  base::PostTaskWithTraits(FROM_HERE, {content::BrowserThread::UI},
-                           base::Bind(callback, result, checkbox_checked));
+  base::PostTaskWithTraits(
+      FROM_HERE, {content::BrowserThread::UI},
+      base::Bind(std::move(callback), result, checkbox_checked));
```
zcbenz: Should be `BindOnce`.

audionerd:
This should probably be listed as a "breaking change" in the release notes.

Any existing call to `dialog.showMessageBox` will now return a Promise instead of a value. So users will need to change to `dialog.showMessageBoxSync` to return to the prior behavior, or modify their code to accept a Promise instead.

mhenr18:
I fully support @audionerd 's comment about updating the docs - I only caught this because the typings for electron updated to reflect the breaking API change and our typescript codebase failed to compile.

PR #17303, opened 2019-03-09
feat: promisify session.getBlobData()

codebytere:
#### Description of Change

Promisifies `session.getBlobData()`.

cc @deepak1556 @miniak @zcbenz 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Converted `session.getBlobData()` to return a Promise instead of taking a callback.


atom/browser/atom_blob_reader.cc
```diff
@@ -28,22 +28,22 @@ void FreeNodeBufferData(char* data, void* hint) {
   delete[] data;
 }
 
-void RunCallbackInUI(const AtomBlobReader::CompletionCallback& callback,
-                     char* blob_data,
-                     int size) {
+void RunPromiseInUI(const atom::util::CopyablePromise& promise,
```
MarshallOfSound: Can't we remove this now that we have `.Resolve()` automatically sending things to the right thread?

codebytere: (resolved via slack)

atom/browser/atom_blob_reader.cc
```diff
@@ -53,22 +53,21 @@ AtomBlobReader::AtomBlobReader(content::ChromeBlobStorageContext* blob_context)
...
+                                       "Could not get blob data handle");
     return;
   }
 
   auto blob_reader = blob_data_handle->CreateReader();
-  BlobReadHelper* blob_read_helper =
-      new BlobReadHelper(std::move(blob_reader), callback);
+  BlobReadHelper* blob_read_helper = new BlobReadHelper(
+      std::move(blob_reader),
+      base::Bind(&RunPromiseInUI, atom::util::CopyablePromise(promise)));
```
zcbenz: Creating `CopyablePromise` involves V8 APIs internally, so it can not be created on IO thread.

For this PR, I think the best solution is to change `BlobReadHelper::CompletionCallback` to use `base::OnceCallback`, and then use move-only `Promise` in all places.

atom/browser/api/atom_api_session.cc
```diff
@@ -659,16 +659,17 @@ std::string Session::GetUserAgent() {
   return browser_context_->GetUserAgent();
 }
 
-void Session::GetBlobData(const std::string& uuid,
-                          const AtomBlobReader::CompletionCallback& callback) {
-  if (callback.is_null())
-    return;
+v8::Local<v8::Promise> Session::GetBlobData(const std::string& uuid) {
+  v8::Isolate* isolate = v8::Isolate::GetCurrent();
```
zcbenz: While using `v8::Isolate::GetCurrent` is fine, the usual way to get isolate in a method is to put isolate in parameters.

```c++
v8::Local<v8::Promise> Session::GetBlobData(v8::Isolate* isolate, const std::string& uuid)
```

atom/browser/atom_blob_reader.cc
```diff
@@ -53,22 +53,21 @@ AtomBlobReader::AtomBlobReader(content::ChromeBlobStorageContext* blob_context)
...
+void AtomBlobReader::StartReading(const std::string& uuid,
+                                  atom::util::Promise promise) {
   DCHECK_CURRENTLY_ON(BrowserThread::IO);
 
   auto blob_data_handle = blob_context_->context()->GetBlobDataFromUUID(uuid);
-  auto callback = base::Bind(&RunCallbackInUI, completion_callback);
   if (!blob_data_handle) {
-    base::PostTaskWithTraits(FROM_HERE, {BrowserThread::UI},
-                             base::BindOnce(callback, nullptr, 0));
+    atom::util::Promise::RejectPromise(std::move(promise),
```
zcbenz: The code is already under `atom` namespace, so the `atom::` prefix is not needed.

PR #17306, opened 2019-03-09
build: remove strip_debug_info gn arg declaration

MarshallOfSound:
It is [android only](https://cs.chromium.org/chromium/src/build/config/compiler/BUILD.gn) and causing release build failures on windows

This is currently causing release build failures on Windows with C74

Notes: no-notes


codebytere:
Can we run a release build to test this?

Issue #12159, opened 2018-03-07
webFrame.executeJavaScript(InIsolatedWorld) APIs don't return promises


bughit:
* Electron version: recent master (dfa1dc43df17037df68b5f39ca7c3c2bb424e9c1)
* Operating system: win7 sp1 x64

### Expected behavior

[Docs](https://github.com/electron/electron/blob/master/docs/api/web-frame.md) say they return  promises.

However since they don't and execution is synchronous, this is an opportunity to change the behavior without breaking anything.  They should return the result. 

### Actual behavior

they return undefined

### How to reproduce

call the api

felixrieseberg:
Do you mean [this API](https://github.com/electron/electron/blob/master/docs/api/web-frame.md#webframeexecutejavascriptinisolatedworldworldid-scripts-usergesture-callback)? Because it's documented to use a callback instead of a Promise.

bughit:
Where does it say "instead"?

For webFrame.executeJavaScript it says:

> Returns Promise - A promise that resolves with the result of the executed code or is rejected if the result of the code is a rejected promise.

For webFrame.executeJavaScriptInIsolatedWorld

> Work like executeJavaScript

However since it's synchronous and unless there are plans to support async mode, it might as well return the result.

MarshallOfSound:
This is actually just incorrect docs, `webFrame.executeJavaScript` does not return a promise.  `webContents.executeJavaScript` **does**.

Just need to correct the docs for `webFrame`

FibreFoX:
This was mentioned in an older isse before:
https://github.com/electron/electron/issues/9804

Maybe this should be done equal?! So it makes less confusion?!

bughit:
@MarshallOfSound 

There three issues here:

1. docs: documenting that these are synchronous and that they don't return promises
2. bug: (#12160) they swallow errors, which was auto-closed, I'll check it with 6.0
3. enhancement: since they are synchronous they can return the eval result

bughit:
no longer applies, see #19161

PR #17312, opened 2019-03-09
feat: promisify executeJavaScript

miniak:
#### Description of Change
- `contents.executeJavaScript()`
- `webFrame.executeJavaScript()`
- `webFrame.executeJavaScriptInIsolatedWorld()`
- `webviewTag.executeJavaScript()`

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Converted `contents.executeJavaScript()`, `webFrame.executeJavaScript()`, `webFrame.executeJavaScriptInIsolatedWorld()` and `webviewTag.executeJavaScript()` to return Promises instead of taking callbacks.

### DIFF
--- a/atom/renderer/api/atom_api_web_frame.cc
+++ b/atom/renderer/api/atom_api_web_frame.cc
@@ -14,6 +14,7 @@
 #include "atom/common/native_mate_converters/gfx_converter.h"
 #include "atom/common/native_mate_converters/string16_converter.h"
 #include "atom/common/node_includes.h"
+#include "atom/common/promise_util.h"
 #include "atom/renderer/api/atom_api_spell_check_client.h"
 #include "base/memory/memory_pressure_listener.h"
 #include "content/public/renderer/render_frame.h"
@@ -92,23 +93,20 @@ class RenderFrameStatus : public content::RenderFrameObserver {
 
 class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
  public:
-  using CompletionCallback =
-      base::Callback<void(const v8::Local<v8::Value>& result)>;
-
-  explicit ScriptExecutionCallback(const CompletionCallback& callback)
-      : callback_(callback) {}
+  explicit ScriptExecutionCallback(atom::util::Promise promise)
+      : promise_(std::move(promise)) {}
   ~ScriptExecutionCallback() override {}
 
   void Completed(
       const blink::WebVector<v8::Local<v8::Value>>& result) override {
-    if (!callback_.is_null() && !result.empty() && !result[0].IsEmpty())
+    if (!result.empty() && !result[0].IsEmpty())
       // Right now only single results per frame is supported.
-      callback_.Run(result[0]);
+      promise_.Resolve(result[0]);
     delete this;
   }
 
  private:
-  CompletionCallback callback_;
+  atom::util::Promise promise_;
 
   DISALLOW_COPY_AND_ASSIGN(ScriptExecutionCallback);
 };
@@ -322,25 +320,32 @@ void InsertCSS(v8::Local<v8::Value> window, const std::string& css) {
   }
 }
 
-void ExecuteJavaScript(mate::Arguments* args,
-                       v8::Local<v8::Value> window,
-                       const base::string16& code) {
+v8::Local<v8::Promise> ExecuteJavaScript(mate::Arguments* args,
+                                         v8::Local<v8::Value> window,
+                                         const base::string16& code) {
+  v8::Isolate* isolate = args->isolate();
+  util::Promise promise(isolate);
+  v8::Local<v8::Promise> handle = promise.GetHandle();
+
   bool has_user_gesture = false;
   args->GetNext(&has_user_gesture);
-  ScriptExecutionCallback::CompletionCallback completion_callback;
-  args->GetNext(&completion_callback);
-  std::unique_ptr<blink::WebScriptExecutionCallback> callback(
-      new ScriptExecutionCallback(completion_callback));
+
   GetRenderFrame(window)->GetWebFrame()->RequestExecuteScriptAndReturnValue(
       blink::WebScriptSource(blink::WebString::FromUTF16(code)),
-      has_user_gesture, callback.release());
+      has_user_gesture, new ScriptExecutionCallback(std::move(promise)));
+
+  return handle;
 }
 
-void ExecuteJavaScriptInIsolatedWorld(
+v8::Local<v8::Promise> ExecuteJavaScriptInIsolatedWorld(
     mate::Arguments* args,
     v8::Local<v8::Value> window,
     int world_id,
     const std::vector<mate::Dictionary>& scripts) {
+  v8::Isolate* isolate = args->isolate();
+  util::Promise promise(isolate);
+  v8::Local<v8::Promise> handle = promise.GetHandle();
+
   std::vector<blink::WebScriptSource> sources;
 
   for (const auto& script : scripts) {
@@ -351,8 +356,8 @@ void ExecuteJavaScriptInIsolatedWorld(
     script.Get("startLine", &start_line);
 
     if (!script.Get("code", &code)) {
-      args->ThrowError("Invalid 'code'");
-      return;
+      promise.RejectWithErrorMessage("Invalid 'code'");
+      return handle;
     }
 
     sources.emplace_back(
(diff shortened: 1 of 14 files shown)


atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -351,8 +358,8 @@ void ExecuteJavaScriptInIsolatedWorld(
     script.Get("startLine", &start_line);
 
     if (!script.Get("code", &code)) {
-      args->ThrowError("Invalid 'code'");
-      return;
+      promise.Reject("Invalid 'code'");
```
codebytere: `promise.RejectWithErrorMessage("Invalid 'code'")`

atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -322,25 +320,34 @@ void InsertCSS(v8::Local<v8::Value> window, const std::string& css) {
...
+  v8::Local<v8::Promise> handle = promise.GetHandle();
+
   bool has_user_gesture = false;
   args->GetNext(&has_user_gesture);
-  ScriptExecutionCallback::CompletionCallback completion_callback;
-  args->GetNext(&completion_callback);
-  std::unique_ptr<blink::WebScriptExecutionCallback> callback(
-      new ScriptExecutionCallback(completion_callback));
+
+  auto callback = std::make_unique<ScriptExecutionCallback>(std::move(promise));
```
deepak1556: The `callback` variable created here is released in the next line, better to inline it at the call site.

atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -367,14 +374,13 @@ void ExecuteJavaScriptInIsolatedWorld(
       blink::WebLocalFrame::kSynchronous;
   args->GetNext(&scriptExecutionType);
 
-  ScriptExecutionCallback::CompletionCallback completion_callback;
-  args->GetNext(&completion_callback);
-  std::unique_ptr<blink::WebScriptExecutionCallback> callback(
-      new ScriptExecutionCallback(completion_callback));
+  auto callback = std::make_unique<ScriptExecutionCallback>(std::move(promise));
```
deepak1556: Same here

codebytere:
@miniak just resolve the markdown file conflict and we can get this across the finish line 🏁

miniak:
@codebytere resolved, can you please merge after the builds pass again?

bughit:
Though I initially filed a bug (#12159) about the docs discrepancy, I very much agree with @MarshallOfSound assessment that this was primarily a [docs](12159#issuecomment-372904750) issue.

These api have always been synchronous (which makes perfect sense, this is in-process js evaluation), and under the hood still are, despite now returning promises. The callback used to deliver the result synchronously before the method returned.

There is nothing gained and only loss when you make actually sync api, async. It's trivial for the user to go from sync to async (`Promise.resolve`) but going from async to sync is impossible.

So this change is:

1. not an improvement
2. a breaking change for 6.0 (I have code that depends on the synchronous nature of these)

Please revert it. An actual improvement would be to replace the sync callback with a sync return.

bughit:
If you believe that promise versions of the webFrame execute APIs should exist out the box (though it's not clear to me why they are needed), that can be done without breaking changes or at least loss of valuable sync functionality, by adding new wrapper methods.

bughit:
#19161

bughit:
@codebytere 

Just to be clear, I am just trying to have a conversation about the impact of the webFrame.execute change, to reach a consensus on how to handle it better, without breaking code or at least losing current capabilities. 6.0 goes into quiet period on the 23rd.

To summarize, the webFrame.execute calls are synchronous. Wrapping them in promises is not an enhancement, it takes away the flexibility of using them synchronously, breaking existing code (example in #19161).

codebytere:
cc @miniak ☝️

miniak:
@codebytere I unfortunately didn't notice that the methods were synchronous before (although invoking the callback instead of returning the value as result). We can make them just return the value directly?

bughit:
@miniak Your comment does not appear here, but I'll respond because I'd like to for this conversation to proceed. My suggestion from #19161:

> ... is to keep the callback in the webFrame.execute for compat, but deprecate it, which I believe is the standard procedure for removing functionality. Additionally return the eval result, which is a compatible change (before it was undefined)

However if it's simpler and quicker to drop the callback, that would be fine too.

bughit:
@codebytere @miniak's 

```c++
class SyncScriptExecutionCallback : public blink::WebScriptExecutionCallback {
 public:
  const char* error_message;
  v8::Local<v8::Value> result;

  explicit SyncScriptExecutionCallback() : error_message(nullptr) {}

  void Completed(
      const blink::WebVector<v8::Local<v8::Value>>& result) override {
    if (!result.empty()) {
      if (!result[0].IsEmpty()) {
        // Right now only single results per frame is supported.
        this->result = result[0];
      } else {
        this->error_message =
            "Script failed to execute, this normally means an error "
            "was thrown. Check the renderer console for the error.";
      }
    } else {
      this->error_message =
          "WebFrame was removed before script could run. This normally means "
          "the underlying frame was destroyed";
    }
  }
};

v8::Local<v8::Value> ExecuteJavaScript(mate::Arguments* args,
                                         v8::Local<v8::Value> window,
                                         const base::string16& code) {
  bool has_user_gesture = false;
  args->GetNext(&has_user_gesture);

  SyncScriptExecutionCallback sync_callback;

  GetRenderFrame(window)->GetWebFrame()->RequestExecuteScriptAndReturnValue(
      blink::WebScriptSource(blink::WebString::FromUTF16(code)),
      has_user_gesture, &sync_callback);

  if (sync_callback.error_message) {
    args->ThrowError(sync_callback.error_message);
    return v8::Null(args->isolate());
  }
  
  return sync_callback.result;
}
```

Is this acceptable?

codebytere:
@bughit while it may have sometimes _worked_ synchronously, this has always been quite clearly documented as an async callback. A given function performing synchronously doesn’t mean the API contract is synchronous, and as such it is, in our opinion, not a breaking change in that manner. As such, we plan to keep this as it is.

bughit:
@codebytere 

Not sometimes, in practice, which matters much more than incorrect documentation, it has always been synchronous, so it does break code, but more significantly, it can't be worked around.

A value can easily be turned to a promise but a fulfilled promise can not synchronously be turned into a value. So the cost of this change is very high, but the benefit is 0.

When I initially started using this api, I saw the documentation was incorrect and I reported it (#12159), so then as a fallback, experimentally and by examining the implementation I confirmed that the call is synchronous. But now you are saying that in the absence of valid docs, it was wrong of me to use the api as it was actually implemented (synchronously) and so I only have myself to blame my code is suddenly broken in 6.0?

If you don't want to revert existing methods,  can you at least accept sync varieties (webFrame.ExecuteJavaScriptSync)? There is value in those and no harm.

bughit:
> it may have sometimes worked synchronously

`ExecuteJavaScript` calls [`RequestExecuteScriptAndReturnValue`](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/frame/web_local_frame_impl.cc?q=RequestExecuteScriptAndReturnValue&sq=package:chromium&g=0&l=794)

which creates a `PausableScriptExecutor` and calls run on it `executor->Run();`

`executor->Run()` is also what [`RequestExecuteScriptInIsolatedWorld`](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/frame/web_local_frame_impl.cc?q=RequestExecuteScriptInIsolatedWorld&sq=package:chromium&g=0&l=820) does when passed `kSynchronous`. So `executor->Run()` == `kSynchronous`

```c++
  switch (option) {
    case kAsynchronousBlockingOnload:
      executor->RunAsync(PausableScriptExecutor::kOnloadBlocking);
      break;
    case kAsynchronous:
      executor->RunAsync(PausableScriptExecutor::kNonBlocking);
      break;
    case kSynchronous:
      executor->Run();
      break;
  }
```

[`kSynchronous`](https://cs.chromium.org/chromium/src/third_party/blink/public/web/web_local_frame.h?sq=package:chromium&q=ScriptExecutionType&g=0&l=378) is defined as "Execute script synchronously, unless the page is suspended."

```c++
  enum ScriptExecutionType {
    // Execute script synchronously, unless the page is suspended.
    kSynchronous,
    // Execute script asynchronously.
    kAsynchronous,
    // Execute script asynchronously, blocking the window.onload event.
    kAsynchronousBlockingOnload
  };
```

I am not a 100% sure but I think suspended means unloaded to save memory, which does not apply to electron, which means `WebFrame.executeJavaScript` is always synchronous which is borne out by my experience with it.

PR #17313, opened 2019-03-09
refactor: add ipcMainUtils.invokeInWebContents / ipcRendererUtils.handle helpers

miniak:
#### Description of Change
Allows invoking handlers easily in renderers from main via IPC, handling errors and promises.
Simplifies https://github.com/electron/electron/pull/17312 (feat: promisify executeJavaScript)

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

/cc @nornagon, @MarshallOfSound

### DIFF
--- a/lib/browser/api/web-contents.js
+++ b/lib/browser/api/web-contents.js
@@ -9,6 +9,7 @@ const { app, ipcMain, session, deprecate } = electron
 
 const NavigationController = require('@electron/internal/browser/navigation-controller')
 const { ipcMainInternal } = require('@electron/internal/browser/ipc-main-internal')
+const ipcMainUtils = require('@electron/internal/browser/ipc-main-internal-utils')
 const errorUtils = require('@electron/internal/common/error-utils')
 
 // session is not used here, the purpose is to make sure session is initalized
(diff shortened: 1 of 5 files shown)


lib/browser/api/web-contents.js
```diff
@@ -193,7 +194,7 @@ const asyncWebFrameMethods = function (requestId, method, callback, ...args) {
 
 for (const method of webFrameMethods) {
   WebContents.prototype[method] = function (...args) {
-    this._sendInternal('ELECTRON_INTERNAL_RENDERER_WEB_FRAME_METHOD', method, args)
+    ipcMainUtils.invoke(this, 'ELECTRON_INTERNAL_RENDERER_WEB_FRAME_METHOD', method, ...args)
```
MarshallOfSound: To avoid confusion can we call this `invokeInRenderer` or `invokeInWebContents` or something to separate it from `ipcRendererUtils`?

miniak: renamed to `invokeInWebContents`

lib/renderer/ipc-renderer-internal-utils.ts
```diff
@@ -1,6 +1,21 @@
 import { ipcRendererInternal } from '@electron/internal/renderer/ipc-renderer-internal'
 import * as errorUtils from '@electron/internal/common/error-utils'
 
+type IPCHandler = (event: Electron.IpcRendererEvent, ...args: any[]) => any
+
+export const handle = function <T extends IPCHandler> (channel: string, handler: T) {
+  ipcRendererInternal.on(channel, (event, requestId, ...args) => {
+    new Promise(resolve => resolve(handler(event, ...args))
```
alexeykuzmin:
```js
Promise.resolve(handler(...))
```

miniak: this would not return a rejected promise if handler throws

PR #17325, opened 2019-03-10
spec: add new test runner for main-process based tests

MarshallOfSound:
## What is this?

This is the initial chunk of work required for us to move some of our tests into a new spec runner that runs in the **main** process of Electron.

## Why?

All of our specs at the moment run in the renderer process of Electron which makes it convenient to write some tests **but**.

a) Some of our tests are nasty, doing sync IPC calls to the main process just to set up hooks that the tests use.  Following this logic across processes is hard and gives people the wrong idea when they look at our specs that this is a _good_ thing to do
b) Our excessive use of `remote` in our renderer tests constantly causes race conditions for us because of the inherent race condition that `remote` brings (being cross process and only synchronous in one direction)

This move should dramatically reduce the flakiness of our tests and avoid the amount of weird patterns we have to introduce just to make our tests do what they should be doing out of the box.

Notes: no-notes


script/spec-runner.js
```diff
@@ -13,6 +13,15 @@ const NPM_CMD = process.platform === 'win32' ? 'npm.cmd' : 'npm'
 
 const specHashPath = path.resolve(__dirname, '../spec/.hash')
 
+let only = null
+const onlyArg = process.argv.find(arg => arg.startsWith('--only='))
```
MarshallOfSound: This allows you to do `npm run test -- --only=remote` locally or `npm run test -- --only=main`

nornagon: can we please use yargs :|

do we foresee having more runners than {main, remote}? if not, i'd argue this should be `--only-remote` / `--only-main`. nitpicky tho.

script/spec-runner.js
```diff
@@ -38,8 +62,41 @@ function saveSpecHash ([newSpecHash, newSpecInstallHash]) {
...
+      console.error(err[1])
+    }
+    throw new Error('Electron test runners have failed')
+  }
+}
+
+async function runRemoteBasedElectronTests () {
   let exe = path.resolve(BASE, utils.getElectronExec())
-  const args = process.argv.slice(2)
+  const args = ['electron/spec', ...process.argv.slice(2).filter(arg => !arg.startsWith('--only='))]
```
nornagon: let's maybe be a little less gung ho about `argv`?

spec-main/events-helpers.ts
```diff
@@ -0,0 +1,40 @@
+import { EventEmitter } from "electron";
```
nornagon: i don't love that this is copy/pasted in two places. it'll get out of sync quick. let's find a way to deduplicate this.

spec-main/index.js
```diff
@@ -0,0 +1,90 @@
...
+  if (argv.grep) mocha.grep(argv.grep)
+  if (argv.invert) mocha.invert()
+
+  // Read all test files.
+  const walker = require('walkdir').walk(__dirname, {
+    no_recurse: true
+  })
+
+  // This allows you to run specific modules only:
+  // npm run test -match=menu
```
nornagon: TIL

spec/static/index.html
```diff
@@ -17,6 +17,12 @@
   const electron = require('electron')
   const { remote, ipcRenderer } = electron
 
+  // Set up chai-as-promised here first to avoid conflicts
+  // It must be loaded first or really strange things happen inside
+  // chai that cause test failures
+  // DO NOT MOVE, REMOVE OR EDIT THIS LINE
```
nornagon: it just needs to be set up _before_ dirty-chai right?

can we just `.use()` them both in this file and remove all the `.use()`s from other places? that'd let us have a less scary comment here

script/spec-runner.js
MarshallOfSound: > do we foresee having more runners than {main, remote}?

In theory, yes, UI regression tests, app-like tests, etc.

I built it out with the intention of supporting N runners

script/spec-runner.js
MarshallOfSound: We want to pass everything through here though so that spec runners get the arguments passed in

spec/static/index.html
MarshallOfSound: @nornagon Yes, mocha actually has a built in way to do this with a "setup" file.  I'll add this in after this PR in a follow up 👍

spec-main/events-helpers.ts
MarshallOfSound: I de-duped it 🎉

script/spec-runner.js
```diff
@@ -5,6 +5,12 @@ const crypto = require('crypto')
 const fs = require('fs')
 const { hashElement } = require('folder-hash')
 const path = require('path')
+const unknownArgs = []
+
+const args = require('minimist')(process.argv, {
+  string: ['only'],
```
nornagon: Can we name this `runners` or something? "only" is a weirdly general flag name

spec-main/events-helpers.ts
nornagon: This doesn't seem to be deduped?

spec-main/api-app-spec.ts
```diff
@@ -28,33 +24,16 @@ describe('electron module', () => {
...
-
-    afterEach(() => {
-      return closeWindow(window).then(() => { window = null })
-    })
-
-    it('always returns the internal electron module', (done) => {
-      ipcMain.once('answer', () => done())
-      window.loadFile(path.join(__dirname, 'fixtures', 'api', 'electron-module-app', 'index.html'))
+    it('always returns the internal electron module', () => {
+      require('electron')
```
nornagon: this doesn't assert anything...?

spec-main/api-app-spec.ts
```diff
@@ -140,7 +119,7 @@ describe('app module', () => {
 
   describe('app.isPackaged', () => {
     it('should be false durings tests', () => {
-      expect(app.isPackaged).to.be.false()
+      expect(app.isPackaged).equal(false)
```
nornagon:
```suggestion
      expect(app.isPackaged).to.equal(false)
```

also why not `to.be.false`?

spec-main/api-app-spec.ts
```diff
@@ -195,32 +174,32 @@ describe('app module', () => {
...
-      const appPath = path.join(__dirname, 'fixtures', 'api', 'singleton')
-      appProcess = ChildProcess.spawn(electronPath, [appPath])
+      const electronPath = process.execPath
+      const appPath = path.join(fixturesPath, 'api', 'singleton')
+      appProcess = cp.spawn(electronPath, [appPath])
 
       // Singleton will send us greeting data to let us know it's running.
       // After that, ask it to exit gracefully and confirm that it does.
-      appProcess.stdout.on('data', data => appProcess.kill())
+      appProcess.stdout.on('data', data => appProcess && appProcess.kill())
```
nornagon: how would `appProcess` be null here? should this be
```suggestion
      appProcess.stdout.on('data', data => appProcess!.kill())
```

spec-main/api-app-spec.ts
```diff
@@ -1242,31 +1227,31 @@ describe('app module', () => {
     })
 
     it('becomes fulfilled if the app is already ready', () => {
-      expect(app.isReady()).to.be.true()
-      expect(app.whenReady()).to.be.eventually.fulfilled()
+      expect(app.isReady()).to.equal(true)
+      expect(app.whenReady()).to.be.eventually.fulfilled.equal(undefined)
```
nornagon:
```suggestion
      await expect(app.whenReady()).to.be.eventually.fulfilled.equal(undefined)
```

spec-main/api-app-spec.ts
MarshallOfSound: Yah the old test didn't either, it just checks that requiring Electron doesn't throw an error

spec-main/events-helpers.ts
MarshallOfSound: As discussed not possible rn

spec-main/api-app-spec.ts
MarshallOfSound: @nornagon Properties are super bad for assertions.  I.e. `to.be.abc` is a valid property.  `dirty-chai` isn't typed correctly either so that doesn't work well in this case

nornagon:
lgtm but
```
Unhandled exception in main spec runner: electron/spec-main/api-app-spec.ts(231,113): error TS7006: Parameter 'x' implicitly has an 'any' type.```

PR #17337, opened 2019-03-11
feat: add safer nativeImage.createFromBitmap(), which does not decode PNG/JPEG

miniak:
#### Description of Change
Required to safely serialize/deserialize `NativeImage` instances to be sent over IPC in https://github.com/electron/electron/pull/17200

/cc @nornagon

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added safer `nativeImage.createFromBitmap()`, which does not decode PNG/JPEG.

### DIFF
--- a/atom/common/api/atom_api_native_image.cc
+++ b/atom/common/api/atom_api_native_image.cc
@@ -505,6 +505,54 @@ mate::Handle<NativeImage> NativeImage::CreateFromPath(
 }
 
 // static
+mate::Handle<NativeImage> NativeImage::CreateFromBitmap(
+    mate::Arguments* args,
+    v8::Local<v8::Value> buffer,
+    const mate::Dictionary& options) {
+  if (!node::Buffer::HasInstance(buffer)) {
+    args->ThrowError("buffer must be a node Buffer");
+    return mate::Handle<NativeImage>();
+  }
+
+  unsigned int width = 0;
+  unsigned int height = 0;
+  double scale_factor = 1.;
+
+  if (!options.Get("width", &width)) {
+    args->ThrowError("width is required");
+    return mate::Handle<NativeImage>();
+  }
+
+  if (!options.Get("height", &height)) {
+    args->ThrowError("height is required");
+    return mate::Handle<NativeImage>();
+  }
+
+  auto info = SkImageInfo::MakeN32(width, height, kPremul_SkAlphaType);
+  auto size_bytes = info.computeMinByteSize();
+
+  if (size_bytes != node::Buffer::Length(buffer)) {
+    args->ThrowError("invalid buffer size");
+    return mate::Handle<NativeImage>();
+  }
+
+  options.Get("scaleFactor", &scale_factor);
+
+  if (width == 0 || height == 0) {
+    return CreateEmpty(args->isolate());
+  }
+
+  SkBitmap bitmap;
+  bitmap.allocN32Pixels(width, height, false);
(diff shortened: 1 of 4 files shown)


docs/api/native-image.md
```diff
@@ -137,6 +137,18 @@ let image = nativeImage.createFromPath('/Users/somebody/images/icon.png')
...
+
+* `buffer` [Buffer][buffer]
+* `options` Object
+  * `width` Integer
+  * `height` Integer
+  * `scaleFactor` Double (optional) - Defaults to 1.0.
+
+Returns `NativeImage`
+
+Creates a new `NativeImage` instance from `buffer`.
```
nornagon: This should describe the expected pixel format. Is it RGBA8888? Or some other form?

docs/api/native-image.md
```diff
@@ -147,7 +159,7 @@ console.log(image)
 
 Returns `NativeImage`
 
-Creates a new `NativeImage` instance from `buffer`.
+Creates a new `NativeImage` instance from `buffer`. Tries to decode as PNG first.
```
nornagon: It also tries to decode as JPEG:

https://github.com/electron/electron/blob/master/atom/common/api/atom_api_native_image.cc#L92-L95

docs/api/native-image.md
miniak: seems to be platform specific (BGRA8888 or RGBA8888). IMO saying it's the raw bitmap pixel data as returned by `toBitmap()` should be enough. Otherwise we need to document what `toBitmap()` returns as well.

docs/api/native-image.md
miniak: added

miniak:
@nornagon I was just thinking about doing that. I agree that having separate `createFromPNG` and `createFromJPEG` is cleaner. I will do that in a follow up PR. What about `addRepresentation`? That one has the same issues.

MarshallOfSound:
I'm thinking we should probably have some kind of "guess what this is" method.  I can see it being handy when you think a file is an image but don't know what format it is.

nornagon:
I think that falls outside the scope of Electron. See e.g. http://www.darwinsys.com/file/

MarshallOfSound:
@nornagon At least pointing people to that / an npm module for file type detection would be enough.  I just don't want to say "figure it out have fun" 👍

atom/common/api/atom_api_native_image.cc
```diff
@@ -505,6 +505,49 @@ mate::Handle<NativeImage> NativeImage::CreateFromPath(
...
+    args->ThrowError("width is required");
+    return mate::Handle<NativeImage>();
+  }
+
+  if (!options.Get("height", &height)) {
+    args->ThrowError("height is required");
+    return mate::Handle<NativeImage>();
+  }
+
+  auto size_bytes = width * height * 4;
```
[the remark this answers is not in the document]
miniak: I think only the byte order is platform dependent. We are calling `allocN32Pixels`, which should guarantee 4 bytes per pixel. I will try to find a Skia API to return this value though.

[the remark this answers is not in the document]
miniak: done

sindresorhus:
> At least pointing people to that / an npm module for file type detection would be enough. I just don't want to say "figure it out have fun" 👍

My [`file-type`](https://github.com/sindresorhus/file-type) package is the most popular one.

PR #17344, opened 2019-03-11
fix: don't crash when nativeImage.createFromBuffer() is called with invalid buffer

miniak:
#### Description of Change
Adds missing `node::Buffer::HasInstance(buffer)` check before using the `buffer`.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Fixed crash when `nativeImage.createFromBuffer()` is called with invalid `buffer`.

### DIFF
--- a/spec/api-native-image-spec.js
+++ b/spec/api-native-image-spec.js
@@ -127,7 +127,7 @@ describe('nativeImage module', () => {
     })
   })
 
-  describe('createFromBuffer(buffer, scaleFactor)', () => {
+  describe('createFromBuffer(buffer, options)', () => {
     it('returns an empty image when the buffer is empty', () => {
       expect(nativeImage.createFromBuffer(Buffer.from([])).isEmpty())
     })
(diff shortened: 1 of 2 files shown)


spec/api-native-image-spec.js
```diff
@@ -165,6 +165,10 @@ describe('nativeImage module', () => {
         { width: 538, height: 190, scaleFactor: 2.0 })
       expect(imageI.getSize()).to.deep.equal({ width: 269, height: 95 })
     })
+
+    it('throws on invalid arguments', () => {
+      expect(() => nativeImage.createFromBuffer(null)).to.throw('buffer must be a node Buffer')
```
MarshallOfSound: Can we add a test for something that conceivably be used to trick this API, like a normal array full of garbage data.  `[12, 14, 124, 12]`?.

miniak:
@nornagon I've seen this in other APIs to validate if an instance is a Buffer
```
void Clipboard::WriteBuffer(const std::string& format,
                            const v8::Local<v8::Value> buffer,
                            mate::Arguments* args) {
  if (!node::Buffer::HasInstance(buffer)) {
    args->ThrowError("buffer must be a node Buffer");
    return;
  }
...
}
```

spec/api-native-image-spec.js
miniak: added

miniak:
@nornagon `gsl::span<gsl::byte>` would be a perfect abstraction for the `Buffer` data/size that can be parse by `mate` converters. see https://github.com/Microsoft/GSL

PR #17347, opened 2019-03-12
build: use npm ci instead of npm install to ensure no lock file changes

MarshallOfSound:
Every damm time you sync or run the tests the `package-lock` changes, because, Reasons ™️ 

Well...

nomore

The `package-lock` shouldn't change now 🎉 

Notes: no-notes

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -107,7 +107,7 @@ hooks = [
     'action': [
       'python',
       '-c',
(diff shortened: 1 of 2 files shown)


MarshallOfSound:
So `npm ci` fails on linux for Reasons ™️ and `--no-package-lock` completely ignores it as well.

Feeling like I should just propose we use `yarn` instead that seems to have (at least in my experience) less of these random and uncontrollable lock file changes

MarshallOfSound:
Superseded by #17955

PR #17363, opened 2019-03-13
feat: make badgeCount a property on app

codebytere:
#### Description of Change

This PR converts `badgeCount` to a proper JS property on `app`, so that users can use get and set on properties instead of using `app.getBadgeCount()` and `app.setbadgeCount(count)`.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Converted `badgeCount` to an actual property on the `app` module.


lib/browser/api/app.ts
```diff
@@ -33,6 +33,14 @@ Object.assign(app, {
   } as Electron.CommandLine,
   enableMixedSandbox () {
     deprecate.log(`'enableMixedSandbox' is deprecated. Mixed-sandbox mode is now enabled by default. You can safely remove the call to enableMixedSandbox().`)
+  },
+  badgeCount: {
```
MarshallOfSound: Can we do this in native code instead with `SetProperty(getter, setter)`?

PR #17377, opened 2019-03-13
chore: add deprecation helper for fnToProperty

codebytere:
#### Description of Change

Add a helper for deprecating a getter/setter function in favor of a bespoke property as part of the modernization initiative.

cc @MarshallOfSound @ckerr @miniak 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/lib/common/api/deprecate.ts
+++ b/lib/common/api/deprecate.ts
@@ -56,14 +56,24 @@ const deprecate: ElectronInternal.DeprecationUtil = {
   },
 
   // deprecate a getter/setter function in favor of a property
-  fnToProperty: (fn, propName) => {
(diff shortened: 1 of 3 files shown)


lib/common/api/deprecate.ts
```diff
@@ -53,6 +55,18 @@ const deprecate: ElectronInternal.DeprecationUtil = {
     })
   },
 
+  // deprecate a getter/setter function in favor of a property
+  fnToProperty: (fn, propName) => {
```
MarshallOfSound: Because we're going to be doing these in pairs, would be make sense for `fn` to either be an array of functions or just take in two functions, a getter and a setter

PR #17378, opened 2019-03-13
refactor: make ELECTRON_INSPECTOR_CONFIRM handler async

miniak:
#### Description of Change
Follow up to "feat: promisify dialog.showMessageBox()" https://github.com/electron/electron/pull/17298

#### Checklist
- [x] `npm test` passes

/cc @codebytere

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/chrome-devtools.js
+++ b/lib/browser/chrome-devtools.js
@@ -97,22 +97,16 @@ ipcMainUtils.handle('ELECTRON_INSPECTOR_SELECT_FILE', async function (event) {
   return [path, data]
 })
 
-ipcMainUtils.handle('ELECTRON_INSPECTOR_CONFIRM', function (event, message, title) {
-  return new Promise((resolve, reject) => {
-    assertChromeDevTools(event.sender, 'window.confirm()')
+ipcMainUtils.handle('ELECTRON_INSPECTOR_CONFIRM', async function (event, message = '', title = '') {
+  assertChromeDevTools(event.sender, 'window.confirm()')
 
-    if (message == null) message = ''
-    if (title == null) title = ''
-
-    const options = {
-      message: `${message}`,
-      title: `${title}`,
-      buttons: ['OK', 'Cancel'],
-      cancelId: 1
-    }
-    const window = event.sender.getOwnerBrowserWindow()
-    dialog.showMessageBox(window, options, (response) => {
-      resolve(response === 0)
-    })
-  })
+  const options = {
+    message: String(message),
+    title: String(title),
+    buttons: ['OK', 'Cancel'],
+    cancelId: 1
+  }
+  const window = event.sender.getOwnerBrowserWindow()
+  const { response } = await dialog.showMessageBox(window, options)
+  return response === 0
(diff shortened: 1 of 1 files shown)


lib/browser/chrome-devtools.js
```diff
@@ -97,22 +97,18 @@ ipcMainUtils.handle('ELECTRON_INSPECTOR_SELECT_FILE', async function (event) {
...
-    }
-    const window = event.sender.getOwnerBrowserWindow()
-    dialog.showMessageBox(window, options, (response) => {
-      resolve(response === 0)
-    })
-  })
+ipcMainUtils.handle('ELECTRON_INSPECTOR_CONFIRM', async function (event, message, title) {
+  assertChromeDevTools(event.sender, 'window.confirm()')
+
+  if (message == null) message = ''
```
codebytere: nit: we can set these as default args
```
async function (event, message = '', title = '') {
```

lib/browser/chrome-devtools.js
```diff
@@ -97,22 +97,18 @@ ipcMainUtils.handle('ELECTRON_INSPECTOR_SELECT_FILE', async function (event) {
...
+  if (title == null) title = ''
+
+  const options = {
+    message: `${message}`,
+    title: `${title}`,
```
alexeykuzmin: weird

lib/browser/chrome-devtools.js
```diff
@@ -97,22 +97,18 @@ ipcMainUtils.handle('ELECTRON_INSPECTOR_SELECT_FILE', async function (event) {
...
+    buttons: ['OK', 'Cancel'],
+    cancelId: 1
+  }
+  const window = event.sender.getOwnerBrowserWindow()
+  return await dialog.showMessageBox(window, options) === 0
```
codebytere: we shouldn't `return await` here since this isn't in a `try/catch` block

lib/browser/chrome-devtools.js
codebytere: should prob just be 
```
    message, title
```

lib/browser/chrome-devtools.js
miniak: of course, fixing now

lib/browser/chrome-devtools.js
miniak: true

lib/browser/chrome-devtools.js
miniak: throwing an exception rejects the promise returned by the async function

miniak: which will make `window.confirm()` throw on the caller side

lib/browser/chrome-devtools.js
alexeykuzmin: If we assume that message and title are already string. Which is probably fine.

miniak: browser's `window.confirm()` stringifies the message

nornagon:
+wg-security for ipc change

PR #17382, opened 2019-03-14
feat: add systemPreferences.getAnimationSettings()

miniak:
#### Description of Change
Expose system animation settings provided by https://cs.chromium.org/chromium/src/ui/gfx/animation/animation.h

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `systemPreferences.getAnimationSettings()` returning system animation settings provided by Chromium.

### DIFF
--- a/docs/api/system-preferences.md
+++ b/docs/api/system-preferences.md
@@ -429,3 +429,13 @@ Returns `Promise<Boolean>` - A promise that resolves with `true` if consent was 
 **Important:** In order to properly leverage this API, you [must set](https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture/requesting_authorization_for_media_capture_on_macos?language=objc) the `NSMicrophoneUsageDescription` and `NSCameraUsageDescription` strings in your app's `Info.plist` file. The values for these keys will be used to populate the permission dialogs so that the user will be properly informed as to the purpose of the permission request. See [Electron Application Distribution](https://electronjs.org/docs/tutorial/application-distribution#macos) for more information about how to set these in the context of Electron.
 
 This user consent was not required until macOS 10.14 Mojave, so this method will always return `true` if your system is running 10.13 High Sierra or lower.
+
+### `systemPreferences.getAnimationSettings()`
+
+Returns `Object`:
+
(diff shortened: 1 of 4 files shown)


codebytere:
@miniak what was the motivation for this change? I'd potentially like to avoid adding methods to modules with no clear community requests or identified use cases 🤔

miniak:
@codebytere we would like to use this API in Teams to adjust our animations

docs/api/system-preferences.md
```diff
@@ -429,3 +429,13 @@ Returns `Promise<Boolean>` - A promise that resolves with `true` if consent was
 **Important:** In order to properly leverage this API, you [must set](https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture/requesting_authorization_for_media_capture_on_macos?language=objc) the `NSMicrophoneUsageDescription` and `NSCameraUsageDescription` strings in your app's `Info.plist` file. The values for these keys will be used to populate the permission dialogs so that the user will be properly informed as to the purpose of the permission request. See [Electron Application Distribution](https://electronjs.org/docs/tutorial/application-distribution#macos) for more information about how to set these in the context of Electron.
 
 This user consent was not required until macOS 10.14 Mojave, so this method will always return `true` if your system is running 10.13 High Sierra or lower.
+
+### `systemPreferences.getAnimationSettings()`
+
+Returns `Object`:
+
+* `shouldRenderRichAnimation` Boolean - Returns true if rich animations should be rendered. Looks at session type (e.g. remote desktop) and accessibility settings to give guidance for heavy animations.
```
MarshallOfSound: For end users this should be `willRenderRichAnimation` right?  Chrome uses it as "should we", this API says "Chromium will"

ppontes: ```gfx::Animation::ShouldRenderRichAnimation()``` seems to be providing guidance for browser clients to do heavy animations or not, not saying that Chromium will do them. Seems to me ```shouldRenderRichAnimation``` is the right property name.

MarshallOfSound: Fair 👍

PR #17385, opened 2019-03-14
feat: add tray.getTitle()

codebytere:
#### Description of Change

This PR adds a getter for the tray title on macOS. This allows a user to programmatically ensure that `setTitle` took effect as well as acting as a precursor for making `title` a property on the Tray module, which would necessitate both a getter and setter where only a setter existed before.

Also cleans up methods within defines so that the title setter and getter don't have to be implemented cross-platform emptily.

cc @nornagon @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `tray.getTitle()` to allows users to see the title set next to the status bar icon on macOS.


codebytere:
@MarshallOfSound done!

Issue #17225, opened 2019-03-05
Calling menu.insert with invalid position causes crash


wjhir:
* **Electron Version:** `3.0.15`, `4.0.6`
* **Operating System:** Windows 10

### Expected Behavior
Calling `menu.insert(pos, item)` should not cause electron to crash, even if the passed values are invalid.

### Actual Behavior
Calling `menu.insert(pos, item)` with an integer outside the range `[0, menu.items.length]` (for `insert*` methods) or `[0, menu.items.length)` (for `get*At` and `is*At` methods) causes electron to crash. Non-integer values throw as expected. Large integers `>= 2**31` also throw, rather than crashing.

### To Reproduce
```javascript
const electron = require('electron')
const { Menu, MenuItem } = electron.remote || electron
const menu = new Menu()
const item = new MenuItem({})
const pos = -1 // or any pos != 0
menu.insert(pos, item)
// The following undocumented methods also cause a crash:
menu.insertCheckItem(pos, item.commandId, item.label)
menu.insertItem(pos, item.commandId, item.label)
menu.insertRadioItem(pos, item.commandId, item.label, item.groupId)
menu.insertSeparator(pos)
menu.insertSubMenu(pos, item.commandId, item.label, item.submenu)
menu.getAcceleratorTextAt(pos)
menu.getCommandIdAt(pos)
menu.getLabelAt(pos)
menu.getSublabelAt(pos)
menu.isEnabledAt(pos)
menu.isItemCheckedAt(pos)
menu.isVisibleAt(pos)
```

PR #17401, opened 2019-03-15
fix: throw error when inserting menu items out-of-range

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/17225.

Adds a check for invalid menu index and throws an appropriate `RangeError`.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added a check for invalid menu index to prevent out-of-range crashes on insertion.

### DIFF
--- a/lib/browser/api/menu.js
+++ b/lib/browser/api/menu.js
@@ -110,7 +110,9 @@ Menu.prototype.insert = function (pos, item) {
     throw new TypeError('Invalid item')
   }
 
-  if (pos > this.getItemCount()) {
+  if (pos < 0) {
+    throw new RangeError(`Position ${pos} cannot be less than 0`)
+  } else if (pos > this.getItemCount()) {
     throw new RangeError(`Position ${pos} cannot be greater than the total MenuItem count`)
   }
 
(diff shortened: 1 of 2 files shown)


lib/browser/api/menu.js
```diff
@@ -110,6 +110,10 @@ Menu.prototype.insert = function (pos, item) {
     throw new TypeError('Invalid item')
   }
 
+  if (pos > this.getItemCount()) {
+    throw new RangeError('Position cannot be greater than the total MenuItem count')
```
MarshallOfSound:
```suggestion
    throw new RangeError(`Position ${pos} cannot be greater than the total MenuItem count ${this.getItemCount()}`)
```

lib/browser/api/menu.js
```diff
@@ -110,6 +110,10 @@ Menu.prototype.insert = function (pos, item) {
     throw new TypeError('Invalid item')
   }
 
+  if (pos > this.getItemCount()) {
```
MarshallOfSound: Purely because some devs might be doing strange math and getting bad indexes.  We should probably also check for `pos < 0` 🤔

codebytere: lol
good point tho, on it

Issue #17193, opened 2019-03-01
[Mac] Top level windows are not excluded from the dock menu by default


CapOM:
* **Electron Version:**
  4.0.2
* **Operating System:**
  macOS 10.14.3
* **Last Known Working Electron version:**:
  1.7

### Expected Behavior
When creating a BrowserWindow it should not appear in the dock menu by defaults.

### Actual Behavior
It appears in the dock menu (example: `raven` in https://electronjs.org/docs/tutorial/macos-dock)

### To Reproduce
Just create new BrowserWindow, then right click on the app icon in the dock. (see https://electronjs.org/docs/tutorial/macos-dock)

At least it should be a way to control that, this can be achieve for top level window with:
`[window_ setExcludedFromWindowsMenu:YES];`, indeed [window isExcludedFromWindowsMenu] returns false by default, see

https://developer.apple.com/documentation/appkit/nsapplication/1428625-removewindowsitem
`Use the isExcludedFromWindowsMenu method of NSWindow if you want the item to remain excluded from the Window menu.`
https://developer.apple.com/documentation/appkit/nswindow/1419175-isexcludedfromwindowsmenu

CapOM:
Actually it is not a regression, just setting the title as empty an it works. But maybe we can change this bug entry to a feature request to add a `BrowserWindow:setExcludedFromWindowsMenu`

codebytere:
@CapOM i'll see what i can do!

PR #17404, opened 2019-03-15
feat: allow windows to be excluded from the windows menu

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/17193.

Allows for windows to be excluded from the windows menu so that they don't appear in either the dock or the primary application menu.

Before:
Screen Shot 2019-03-15 at 2 13 11 PM

After:
Screen Shot 2019-03-15 at 2 17 28 PM

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `excludedFromShownWindowsMenu` property to allow for windows to be excluded from the windows menu.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -1648,3 +1648,24 @@ removed in future Electron releases.
 [vibrancy-docs]: https://developer.apple.com/documentation/appkit/nsvisualeffectview?preferredLanguage=objc
 [window-levels]: https://developer.apple.com/documentation/appkit/nswindow/level
 [chrome-content-scripts]: https://developer.chrome.com/extensions/content_scripts#execution-environment
+
+### Properties
+
+#### `win.excludedFromWindowsMenu` _macOS_
+
+A `Boolean` property that determines whether the window is excluded from the application’s Windows menu. `false` by default.
+
+```js
+const win = new BrowserWindow({ height: 600, width: 600 })
+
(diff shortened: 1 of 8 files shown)


docs/api/browser-window.md
```diff
@@ -1648,3 +1648,24 @@ removed in future Electron releases.
 [vibrancy-docs]: https://developer.apple.com/documentation/appkit/nsvisualeffectview?preferredLanguage=objc
 [window-levels]: https://developer.apple.com/documentation/appkit/nswindow/level
 [chrome-content-scripts]: https://developer.chrome.com/extensions/content_scripts#execution-environment
+
+### Properties
+
+#### `win.excludedFromWindowsMenu` _macOS_
```
MarshallOfSound: > `excludedFromWindowsMenu`

The naming off this could be confusing considering "Windows" as the OS as well.  No suggestions right now, but I think we need to come up with an alternate here 🤔

Kilian: suggestions:

* `excludedFromListOfWindows`
* `excludedFromListOfWindowsMenu`
* `excludedFromShownWindows`
* `excludedFromShownWindowsMenu`
* `excludedFromShownWindowsList`
* `excludedFromWindowsList`

codebytere: updated with `excludedFromShownWindowsMenu` ☝️

Issue #12045, opened 2018-02-26
Preload script executed twice on Electron 2.0.x


MarshallOfSound:
* Electron version: `2.0.0-beta.1`
* Operating system: macOS

### Expected behavior

Preload script should execute once

### Actual behavior

Preload script (and in fact `node_bootstrap.js`) is executed twice in the same process but in a different context.

### How to reproduce

Struggling to isolate this from the application that is experiencing it but I have narrowed it down to this being _kind of_ expected behavior but with a newly introduced bug in `2.0.x`.  Basically when you call `loadURL` a renderer process is initially launched pointing at `about:blank`.  In this process `renderer/init.js` is called and then the window navigates to the URL you requested.

For an as yet unknown reason in `2.0.x` the `--preload` argument is present on the renderer process when the `about:blank` page is loaded causing user code to execute twice in the same process.  This is only a problem because native modules can only be loaded once in a given process.

The fix for this would be to revert to original behavior where the `--preload` (amongst others) arguments are not sent to the renderer until the users page is loaded.  (Don't send them for `about:blank`).

I'm currently working on a minimal sample

nornagon:
Did you manage to find a minimal sample? It'd be great to have a test for this.

PR #17406, opened 2019-03-15
refactor: use mojo for electron internal IPC

nornagon:
#### Description of Change
This refactors Electron's internal IPC to use [Mojo](https://chromium.googlesource.com/chromium/src/+/74.0.3724.8/mojo/), instead of legacy IPC.

This PR is a straightforward port that swaps out `AtomFrameMsg_Message` and `AtomFrameHostMsg_Message` and their siblings to use Mojo as their underlying transport. In future I intend to further expand our use of Mojo and describe the various IPC interfaces we expose as fleshed-out Mojo IDLs, using the [Mojo JS binding](https://chromium.googlesource.com/chromium/src/+/74.0.3724.8/mojo/public/js/README.md).

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


atom/app/manifests.cc
```diff
@@ -25,6 +28,25 @@ const service_manager::Manifest& GetElectronContentBrowserOverlayManifest() {
...
+  static base::NoDestructor<service_manager::Manifest> manifest{
+      service_manager::ManifestBuilder()
+          .WithDisplayName("Electron (renderer process)")
+          .ExposeCapability("electron_api",
+                            std::set<const char*>{
+                                "electron_api.mojom.Electron",
+                            })
+          .ExposeInterfaceFilterCapability_Deprecated(
+              "navigation:frame", "browser",
+              std::set<const char*>{
```
deepak1556: I think its preferred to use `service_manager::Manifest::InterfaceList<T>()` variation for this ?

atom/browser/api/atom_api_web_contents.h
```diff
@@ -462,6 +464,41 @@ class WebContents : public mate::TrackableObject<WebContents>,
 #endif
 
  private:
+  // Handles IPC messages from a specific RenderFrame. Created lazily.
+  // TODO(nornagon): this probably shouldn't be tied to the api::WebContents.
+  // Message/MessageSync are api::WebContents-specific but it's also totally
+  // plausible that there would be IPC messages that could be handled by the
+  // browser process that don't require an api::WebContents to exist.
+  class IPCHandler : public electron_api::mojom::ElectronBrowser,
```
deepak1556: yeah this seems to be a combination of `OnMessageReceived` and `FrameDispatchHelper`. Maybe we can have interfaces specific to `api::WebContents` declared separately from the generic ones, that way we can override them directly without having an `IPCHandler` class ?

atom/common/api/BUILD.gn
```diff
@@ -0,0 +1,11 @@
+import("//mojo/public/tools/bindings/mojom.gni")
+
+mojom("mojo") {
```
deepak1556: when we had `atom/common/api/api_messages.h` it was based on the directory structure following upstream ipc codebase. WDYT about defining browser interfaces under `atom/browser` and renderer interfaces under `atom/renderer`, since in mojo world these are two distinct services.

atom/app/manifests.cc
nornagon: Huh, TIL. Interestingly, this code seems to work fine now when I remove this stuff from the manifest. Not sure what changed.

atom/browser/api/atom_api_web_contents.h
nornagon: Not quite sure what you're saying; that `api::WebContents` should derive from `mojom::ElectronBrowser`?

nornagon: I tried folding the `mojom::ElectronBrowser` implementation up into `api::WebContents`, but ran into a snag, which is that `content::WebContentsFrameBindingSet` must be initialized in the constructor, but not all constructors of `api::WebContents` have a relevant `content::WebContents` pointer we could pass in.

I think this is a weakness in the design of `api::WebContents`, but I don't plan to fix that here.

atom/common/api/BUILD.gn
nornagon: Hmm, looking at `//chrome`, the vast majority of mojom files are under `//chrome/common`. `//chrome/renderer` contains _no_ mojom files. Most of the mojom files in `//chrome/browser` are webui-related.

I think these should be in common because the interfaces are _shared_ between the browser and renderer (one implementing, one using).

nornagon:
Here are some things I've learned about mojo along the path to this PR. This comment intended as both informative & also as a ping, please review this :)

### Associated Interfaces

[Associated Interfaces](https://chromium.googlesource.com/chromium/src/+/master/mojo/public/cpp/bindings/README.md#associated-interfaces) are a concept in Mojo that allows interfaces to be associated with an existing message pipe. Normally, every interface in Mojo gets its own message pipe (& so can receive messages out-of-order compared with other interfaces), but using an associated interface means you'll receive messages in order with the associated pipe.

The linked document above talks about an `associated` mojom keyword, which we aren't using. We're using associated interfaces so we can share the message pipe with the legacy IPC system, meaning that message order won't be affected by this change (messages over the new Mojo IPCs will remain in-order with legacy IPCs). I don't know if this is important for the IPCs that I've converted (renderer->browser: `Message`, `MessageSync` and browser->renderer: `Message`, `TakeHeapSnapshot`), but it seemed prudent to stick with the same order. In future we can experiment with switching to regular, non-associated Mojo interfaces.

### RenderFrames, RenderFrameHosts, and InterfaceProviders

Mojo pipes aren't inherently linked to frames or renderers or anything, they're totally generic. So there needs to be plumbing hooked up if you want to keep track of which frame belongs to which Mojo pipe. Fortunately, Chromium already has some infrastructure for this, in the form of `{RenderFrame,RenderFrameHost}::GetRemote{,Associated}Interfaces()`. These functions return `service_manager::InterfaceProvider`s which are hooked up to track the `RenderFrame` / `RenderFrameHost`, and provide interfaces to the other end, like so:

```cpp
    //  we're in the browser process here.
    // this will become a pointer to an interface provided by the renderer.
    mojom::ElectronRendererAssociatedPtr electron_ptr;
    // request the interface via the InterfaceProvider
    render_frame_host->GetRemoteAssociatedInterfaces()->GetInterface(
        mojo::MakeRequest(&electron_ptr));
    // call a method on the interface.
    electron_ptr->Message(internal, channel, args.Clone(), sender_id);
```

Note that in the above code, there are no round-trips over the IPC pipe! Everything is pipelined. If the interface fails to bind on the remote end, the method will never be called and an error will be raised on the `ElectronRendererAssociatedPtr` (interceptible by `set_connection_error_handler`).

### Interface lifetimes

When you request an interface, you get a handle to it (something like `mojom::ElectronRendererAssociatedPtr`). A handle is also created on the remote end. If either end of the pipe is closed, the other end will finish processing whatever messages it has left in its queue, then close its own end of the pipe.

The above code sample requests an interface, sends a message over it, then closes its end of the pipe (implicitly, when the `ElectronRenderAssociatedPtr` goes out of scope). The other end of the pipe will receive the request, bind a new interface (in this case, a new instance of `ElectronApiServiceImpl` will be created), call the method on the interface, then receive a notification that the other end of the pipe has been closed. `ElectronApiServiceImpl` listens for the remote pipe being closed and cleans itself up.

It would also be a reasonable design to have a single handle to a remote interface through which all communication would be routed. I'm not sure yet which is better, but this design was simpler for now because there are a fair few places around the code that send IPCs and I wasn't sure where to store such a single handle per frame. This way ends up with a little more traffic on the IPC pipe (requesting the interface anew & closing it for each message), as well as a few more function calls and allocations on the receiving end, but the cost is small and the tradeoff is that the code can be simpler on the calling side.

lib/renderer/init.ts
```diff
@@ -22,8 +22,20 @@ globalPaths.push(path.join(__dirname, 'api', 'exports'))
...
+const ipcEmitter = new EventEmitter()
+const ipcInternalEmitter = new EventEmitter()
+v8Util.setHiddenValue(global, 'ipc', ipcEmitter)
+v8Util.setHiddenValue(global, 'ipc-internal', ipcInternalEmitter)
+
+v8Util.setHiddenValue(global, 'ipcNative', {
+  onMessage (internal: boolean, channel: string, args: any[], senderId: number) {
+    (internal ? ipcInternalEmitter : ipcEmitter).emit(
+      channel,
+      { sender: ipcInternalEmitter, senderId },
```
deepak1556: shouldn't sender also be decided based on `internal ? ipcInternalEmitter : ipcEmitter`

lib/sandboxed_renderer/init.js
```diff
@@ -45,30 +45,30 @@ const loadedModules = new Map([
...
-    ipcRendererInternal.emit(channel, { sender: ipcRendererInternal, senderId }, ...args)
-  } else {
-    electron.ipcRenderer.emit(channel, { sender: electron.ipcRenderer, senderId }, ...args)
+// ElectronApiServiceImpl will look for the "ipcNative" hidden object when
+// invoking the 'onMessage' callback.
+v8Util.setHiddenValue(global, 'ipcNative', {
+  onMessage (internal, channel, args, senderId) {
+    (internal ? ipcRendererInternal : electron.ipcRenderer).emit(
+      channel,
+      { sender: ipcRendererInternal, senderId },
```
deepak1556: same here

atom/renderer/electron_api_service_impl.cc
```diff
@@ -0,0 +1,151 @@
...
+namespace atom {
+
+namespace {
+
+const char kIpcKey[] = "ipcNative";
+
+void InvokeIpcCallback(v8::Handle<v8::Context> context,
+                       const std::string& callback_name,
+                       std::vector<v8::Handle<v8::Value>> args) {
+  TRACE_EVENT0("devtools.timeline", "FunctionCall");
```
deepak1556: Seems like the common elements in `InvokeIpcCallback` and `InvokeHiddenCallback` can be extracted into some v8 util inside native mate

atom/renderer/electron_api_service_impl.cc
```diff
@@ -0,0 +1,151 @@
...
+void ElectronApiServiceImpl::OnDestruct() {
+  delete this;
+}
+
+void ElectronApiServiceImpl::Message(bool internal,
+                                     bool send_to_all,
+                                     const std::string& channel,
+                                     base::Value arguments,
+                                     int32_t sender_id) {
+  blink::WebLocalFrame* frame = render_frame_->GetWebFrame();
```
deepak1556: Is the `document_created_` check no a longer a concern with mojo for preload scripts https://github.com/electron/electron/blob/master/atom/renderer/atom_render_frame_observer.cc#L201-L205 ? Although the service creation is only after render frame is created but that doesn't guarantee document creation.

/cc @MarshallOfSound , since you worked on it https://github.com/electron/electron/issues/12045

atom/browser/api/atom_api_web_contents.cc
```diff
@@ -425,6 +421,8 @@ void WebContents::InitWithSessionAndOptions(
   // Initialize zoom controller.
   InitZoomController(web_contents(), options);
 
+  IPCHandler::CreateForWebContentsWithApiWebContents(web_contents(), this);
```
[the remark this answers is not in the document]
nornagon: Just following the pattern of other things that do this in Chrome, e.g. https://chromium.googlesource.com/chromium/src/+/73.0.3683.58/chrome/browser/ui/tab_helpers.cc#196

atom/browser/api/atom_api_web_contents.cc
```diff
@@ -886,6 +884,41 @@ bool WebContents::EmitNavigationEvent(
...
 
+void WebContents::IPCHandler::CreateForWebContentsWithApiWebContents(
+    content::WebContents* web_contents,
+    WebContents* api_web_contents) {
+  if (FromWebContents(web_contents))
+    return;
+
+  web_contents->SetUserData(
+      UserDataKey(),
+      base::WrapUnique(new IPCHandler(web_contents, api_web_contents)));
```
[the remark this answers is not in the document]
nornagon: Done; oddly, this required that I make the constructor of `IPCHandler` public.

atom/renderer/electron_api_service_impl.cc
```diff
@@ -0,0 +1,151 @@
...
+
+ElectronApiServiceImpl::ElectronApiServiceImpl(
+    content::RenderFrame* render_frame,
+    mojom::ElectronRendererAssociatedRequest request)
+    : content::RenderFrameObserver(render_frame),
+      binding_(this),
+      render_frame_(render_frame),
+      weak_ptr_factory_(this) {
+  (void)render_frame_;
+  weak_this_ = weak_ptr_factory_.GetWeakPtr();
```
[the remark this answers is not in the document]
nornagon: I think I used it in a previous iteration of this but yeah, it no longer seems necessary. Removed.

atom/renderer/electron_api_service_impl.cc
```diff
@@ -0,0 +1,151 @@
...
+                  const std::string& channel,
+                  const std::vector<base::Value>& args,
+                  int32_t sender_id) {
+  if (!frame)
+    return;
+
+  v8::Isolate* isolate = blink::MainThreadIsolate();
+  v8::HandleScope handle_scope(isolate);
+
+  v8::Local<v8::Context> context =
```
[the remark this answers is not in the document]
nornagon: We have no access to `RenderClientBase` here, I don't think. Perhaps we could break it out into a non-instance-method though? It doesn't actually depend on anything in `RenderClientBase` except the bool for `isolated_world_`, which would be easy to break out.

atom/renderer/electron_api_service_impl.cc
```diff
@@ -0,0 +1,151 @@
...
+ElectronApiServiceImpl::~ElectronApiServiceImpl() = default;
+
+ElectronApiServiceImpl::ElectronApiServiceImpl(
+    content::RenderFrame* render_frame,
+    mojom::ElectronRendererAssociatedRequest request)
+    : content::RenderFrameObserver(render_frame),
+      binding_(this),
+      render_frame_(render_frame),
+      weak_ptr_factory_(this) {
+  (void)render_frame_;
```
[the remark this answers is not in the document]
nornagon: I think I was testing something, I don't remember what 😅 removed.

lib/renderer/init.ts
nornagon: oops, yes. done.

atom/renderer/electron_api_service_impl.cc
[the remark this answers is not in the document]
nornagon: On second thoughts, maybe we should just pass in a reference to the renderer client. I also don't love `isolated_world_` being a global variable.

atom/renderer/electron_api_service_impl.cc
nornagon: I'd prefer to avoid modifications to native_mate, as I'm expecting it to eventually go away and be replaced by gin. I agree that these functions are largely similar though!

nornagon:
I spent some time digging up the root cause of these test failures, writing it down here because it's 5pm on Friday and I don't yet have a good idea of how to fix them.

I think it comes down to a subtle difference in the way synchronous messages are handled in Mojo vs. how we were previously handling them.

In Mojo, if a process is currently waiting for a reply to a synchronous call, _that process will still handle incoming synchronous calls_. i.e. if process A calls `SomeMethodSync()` on process B, then process C sends process A a _synchronous_ message while process A is still waiting for the reply to `SomeMethodSync()`, process A will handle the incoming message from process C while it waits. However, asynchronous messages will be delayed until the synchronous call is finished.

Mojo_C___Bindings_API


These tests are flaking out—they're not _deterministically_ failing—because sometimes, when a child window closes and the asynchronous `ELECTRON_GUEST_WINDOW_MANAGER_WINDOW_CLOSE` message gets sent to the browser, the message arrives when the browser process is in the middle of making a synchronous call. So instead of processing the message immediately, it puts it in a queue for later. Then, a synchronous call for the `remote` module arrives (i.e. `ELECTRON_BROWSER_GLOBAL`). This call is processed immediately, despite the pending synchronous call. Once the synchronous call finally returns, the `WINDOW_CLOSE` message is processed.

This results in the messages appearing to arrive out of order:

Sender:
1. `WINDOW_CLOSE` (async)
2. `BROWSER_GLOBAL` (sync)

Receiver sees:
1. `WINDOW_CLOSE` (async, not handled immediately but instead placed on a queue)
1. `BROWSER_GLOBAL` (sync, handled immediately)
1. `WINDOW_CLOSE` actually handled

chrome___tracing


So Mojo seems to be capable of reordering messages when they're synchronous. The [mojo docs](https://chromium.googlesource.com/chromium/src/+/master/mojo/public/cpp/bindings/README.md#synchronous-calls) call this out:
> ### Think carefully before you decide to use sync calls
>
> Although sync calls are convenient, you should avoid them whenever they are not absolutely necessary:
> *   Sync calls hurt parallelism and therefore hurt performance.
> *   Re-entrancy changes message order and produces call stacks that you probably never think about while you are coding. It has always been a huge pain.
> *   Sync calls may lead to deadlocks.

I'm not sure if it's possible to emulate the legacy sync IPC behaviour exactly with Mojo. This will require a bit more investigation.

nornagon:
Further, it's unclear to me whether the legacy IPC system actually doesn't have this issue, or if we just got lucky in not seeing it.

nornagon:
A little research into the implementation of legacy IPC over Mojo lead me to this:

https://chromium.googlesource.com/chromium/src/+/74.0.3729.5/ipc/ipc.mojom#27

```
interface Channel {
  // Transmits a classical Chrome IPC message.
  Receive(Message message);

  // ...
};
```

i.e. all legacy IPC messages are transmitted as asynchronous Mojo messages, regardless of their sync-ness in legacy.

nornagon:
Adding a note here of a failed exploration. I looked into patching mojo to allow a new message flag that would cause synchronous messages to be processed in order with asynchronous messages on the receiving side (i.e. disable the re-entrancy described [here](https://github.com/electron/electron/pull/17406#issuecomment-475824538)).

To test if this would work before going full-on into the mojo code generator, instead of editing the template files, I just directly edited the `api.mojom.cc` file to add `(1 << 3)` to the message's header flags, e.g.

```cpp
  const uint32_t kFlags =
      ((kExpectsResponse) ? mojo::Message::kFlagExpectsResponse : 0) |
      ((kIsSync) ? mojo::Message::kFlagIsSync : 0) |
      (1 << 3);
```

in `ElectronBrowserProxy::MessageSync`.

Then I edited `ipc/ipc_mojo_bootstrap.cc`, which is what dispatches incoming messages on the legacy ipc channel (which our interface is tied to because it's an `AssociatedInterface`). Specifically I made these changes:

```
diff --git a/ipc/ipc_mojo_bootstrap.cc b/ipc/ipc_mojo_bootstrap.cc
index 5eb74655287b..2a8e8957fd1d 100644
--- a/ipc/ipc_mojo_bootstrap.cc
+++ b/ipc/ipc_mojo_bootstrap.cc
@@ -828,7 +828,7 @@ class ChannelAssociatedGroupController
       // runs or else it's programmer error.
       DCHECK(proxy_task_runner_);

-      if (message->has_flag(mojo::Message::kFlagIsSync)) {
+      if (message->has_flag(mojo::Message::kFlagIsSync) && !message->has_flag(1 << 3)) {
         MessageWrapper message_wrapper(this, std::move(*message));
         // Sync messages may need to be handled by the endpoint if it's blocking
         // on a sync reply. We pass ownership of the message to the endpoint's
@@ -878,7 +878,7 @@ class ChannelAssociatedGroupController
     DCHECK(endpoint->task_runner()->RunsTasksInCurrentSequence());

     // Sync messages should never make their way to this method.
-    DCHECK(!message.has_flag(mojo::Message::kFlagIsSync));
+    DCHECK(!message.has_flag(mojo::Message::kFlagIsSync) || message.has_flag(1 << 3));

     bool result = false;
     {
```

This fixed the ordering issue.

I took this idea to #mojo in the Chromium Slack, suggesting that Mojo could add a `[NoInterrupt]` flag to a message that would prevent the message from interrupting a synchronous call and jumping the queue. There wasn't much enthusiasm for the idea though, and it was instead suggested that we send a regular async message and block on its response in a worker thread (which is apparently how synchronous XHR works).

atom/app/manifests.cc
nornagon: Restored part of this due to changing how I'm binding the `ElectronBrowser` service and used `Manifest::InterfaceList`. Thanks for the pointer.

nornagon:
The latest commit makes a couple of significant changes:

1. `MessageSync` is no longer a mojo `[Sync]` call; instead, we call it asynchronously in a background thread and manually block on the response. This fixes the out-of-order issue.
2. We no longer use an "associated" interface, meaning it's possible for messages on the Electron IPC interface to be transmitted out of order with respect to legacy IPC messages on the frame channel.

(2) is required for (1) to work, because otherwise blocking the main thread prevents any messages from being processed on the frame channel. I don't believe we rely on consistent ordering between that frame channel and the `Message`/`MessageSync` API—if we do, we should refactor until we don't, otherwise we'll have to take yet another different approach. Additionally, channel-associated interfaces are a very delicate feature and are [quite possibly going away soon](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/mojo_ipc_conversion.md#Channel_Associated-Interfaces), so switching away from an associated interface is probably the more future-proof decision anyway.

atom/renderer/api/atom_api_renderer_ipc.cc
```diff
@@ -27,40 +33,88 @@ RenderFrame* GetCurrentRenderFrame() {
...
+        base::CreateSingleThreadTaskRunnerWithTraits({});
+
+    base::WaitableEvent response_received_event;
+
+    auto interface_info = electron_browser_ptr_.PassInterface();
+    atom::mojom::ElectronBrowserPtrInfo returned_interface_info;
+    task_runner->PostTask(
+        FROM_HERE,
+        base::BindOnce(
+            [](atom::mojom::ElectronBrowserPtrInfo interface_info,
```
deepak1556: Not sure why capturing lambdas aren't allowed via `base::Bind` (wild guess it might be the **cons** described in https://google.github.io/styleguide/cppguide.html#Lambda_expressions), but they seem to available for tests with help of [base::BindLambadForTesting](https://cs.chromium.org/chromium/src/base/test/bind_test_util.h?l=29-36), using a similar utility can simplify this block. In this case the execution is known to be blocked, so there shouldn't any worry of dangling pointers.

```c++
    task_runner->PostTask(
        FROM_HERE,
        base::BindOnce(
            BindLambda(
                [&](atom::mojom::ElectronBrowserPtrInfo interface_info) {
                  atom::mojom::ElectronBrowserPtr browser_ptr(
                      std::move(interface_info));
                  browser_ptr->MessageSync(
                      internal, channel, arguments.Clone(),
                      base::BindOnce(
                          BindLambda(
                              [&](atom::mojom::ElectronBrowserPtr ptr,
                                  base::Value response) {
                                result = std::move(response);
                                returned_interface_info = ptr.PassInterface();
                                response_received_event.Signal();
                              }),
                          std::move(browser_ptr)));
                }),
            std::move(interface_info)));
```

its just that nested lambdas doesn't look good at first pass :) even binding them as helper functions would be good.

nornagon: I'm hesitant to use a utility that's specifically marked for testing only in a non-test code path. I'll split these out into helper functions to make this a little clearer.

atom/renderer/electron_api_service_impl.cc
nornagon: ping @MarshallOfSound, would love to get your input on this

spec/chromium-spec.js
```diff
@@ -371,7 +371,7 @@ describe('chromium feature', () => {
 
     it('disables node integration when it is disabled on the parent window', (done) => {
       let b = null
-      listener = (event) => {
+      listener = async (event) => {
```
MarshallOfSound: Why did this become `async`?

nornagon: oops, i changed it because i had a wait loop in here, but then i removed the wait loop. fixed.

nornagon:
> Could we potentially add a test that just sends a few thousand IPC messages back and forth and asserts they arrive in the right processes in the correct order?

Yes, but I think this would essentially just be testing a property that's already guaranteed by Mojo?

MarshallOfSound:
> Yes, but I think this would essentially just be testing a property that's already guaranteed by Mojo?

I was referring more to testing the work you did in making our sync messages work.

> MessageSync is no longer a mojo [Sync] call; instead, we call it asynchronously in a background thread and manually block on the response. This fixes the out-of-order issue.

☝️

nornagon:
It would be difficult to replicate these circumstances exactly, because the only time that messages would get reordered is if the main process is waiting on a response to synchronous call, and the main process almost never actually does that (synchronous calls are actually banned in the browser process, you have to specifically allow them and only a [small set of classes](https://chromium.googlesource.com/chromium/src/+/74.0.3729.5/mojo/public/cpp/bindings/sync_call_restrictions.h#75) are permitted to do so). One of the very few synchronous calls that the browser makes is called during window close (it has to do with GL contexts); outside of that situation, the `[Sync]` method works fine & delivers messages in order.

So IMO, to test this effectively we would have to:
1. Set up the browser to call a synchronous method, probably by adding our own test-only synchronous message,
2. Call `Message` in a renderer,
3. Call `MessageSync` in a renderer,
4. Complete the synchronous call in the browser,
5. Ensure that the `Message` arrived before the `MessageSync`.

Ultimately I don't think such a test is worth the effort it would take to create it, because it's testing a guarantee that Mojo already provides: that async messages are always delivered in order.

atom/renderer/api/atom_api_renderer_ipc.cc
```diff
@@ -27,40 +33,91 @@ RenderFrame* GetCurrentRenderFrame() {
...
-
-base::ListValue SendSync(mate::Arguments* args,
-                         bool internal,
-                         const std::string& channel,
-                         const base::ListValue& arguments) {
-  base::ListValue result;
-
-  RenderFrame* render_frame = GetCurrentRenderFrame();
-  if (render_frame == nullptr)
+class IPCRenderer : public mate::Wrappable<IPCRenderer> {
```
nornagon: @zcbenz it looks like you moved away from this pattern in d16b5811409445724c1db9d5f95b5aa9ad236be8, do you have thoughts on whether it's OK here or will this leak also?

atom/renderer/electron_api_service_impl.cc
```diff
@@ -0,0 +1,166 @@
...
+
+  v8::Isolate* isolate = blink::MainThreadIsolate();
+  v8::HandleScope handle_scope(isolate);
+
+  v8::Local<v8::Context> context = renderer_client_->GetContext(frame, isolate);
+
+  EmitIPCEvent(context, internal, channel, arguments.GetList(), sender_id);
+
+  // Also send the message to all sub-frames.
+  if (send_to_all) {
```
MarshallOfSound: `send_to_all` is moving to the main process (initial work is here --> https://github.com/electron/electron/pull/17614)

atom/renderer/api/atom_api_renderer_ipc.cc
zcbenz: This will leak too since instances may be created in subframes and currently we don't cleanup Node environments there.

But this leak should be fine as it only stores a pointer, unlike `SpellChecker` which takes a large chuck of memory.

MarshallOfSound: > currently we don't cleanup Node environments there.

FWIW in Electron 6 we may be able to do this, we already do clean them up when you enable `nodeIntegrationSupportInSubFrames`.  Enabling this for all frames should be a thing we can _try_ on master

atom/renderer/electron_api_service_impl.cc
nornagon: Cool, I'll resolve this conflict then. LGTY otherwise? Specifically re: https://github.com/electron/electron/pull/17406#discussion_r270522576

atom/renderer/api/atom_api_renderer_ipc.cc
nornagon: Perhaps this should implement `RenderFrameObserver` and delete itself when the render frame is deleted?

MarshallOfSound: @nornagon The issue isn't that it outlives the frame, rather it comes into existence *before* the document has been created inside the frame.

YurySolovyov:
Any idea if/how this might affect performance or responsiveness of the renderer process?

nornagon:
> Any idea if/how this might affect performance or responsiveness of the renderer process?

I don't expect this change to significantly affect performance, as legacy IPC has been using mojo under the hood for several years. I intend to check throughput + latency before/after this change in order to confirm this thesis.

nornagon:
Perf test for sync IPC round-trip time:

<details>
<summary>renderer.js</summary>

```js
const {ipcRenderer} = require('electron')

const n = 10000
let total = 0
for (let i = 0; i < n; i++) {
    const begin = performance.now()
    ipcRenderer.sendSync("hello")
    const end = performance.now()
    total += end - begin
}
console.log(`average: ${total / n}`)
```

</details>

<details>
<summary>main.js</summary>

```js
const {ipcMain} = require('electron')

ipcMain.on('hello', (e) => {
  e.returnValue = 'hi'
})
```

</details>

Results (3 runs):
legacy: 0.457ms, 0.452ms, 0.472ms (0.461ms avg)
mojo: 0.421ms, 0.418ms, 0.435ms (0.425ms avg)

mojo 8.5% faster.

Testing async call performance is a little trickier since `performance.now()` isn't available in the main process, and `process.hrtime()` doesn't seem to be comparable between processes.

yohan1234:
@nornagon

Using electron 11.1.0.

I'm seeing random crashes related to this PR. We frequently create and destroy BrowserViews using BrowserView.webcontents.destroy() as discussed in [23578](https://github.com/electron/electron/pull/23578).

In msvc I'm seeing this call stack:
image

Error in question: _The thread tried to read from or write to a virtual address for which it does not have the appropriate access._

Seems that some kind of callback is not set? (InvokeProcessErrorCallback)

Within each BrowserView I'm using MessagePort to the main process. 

Any ideas?

nornagon:
I'm not sure how you concluded that this PR was related. Would you mind opening a new issue? If possible, please include a [Fiddle](/electron/fiddle) that reproduces the issue.

yohan1234:
@nornagon Just from digging through the commits of when this was merged. I will try and create a repro over the upcoming weeks using fiddle thanks.

yohan1234:
@nornagon sorry to have jumped the gun on this one. I tracked it down to being a Chromium bug:
https://bugs.chromium.org/p/chromium/issues/detail?id=1160307

Johan

PR #17416, opened 2019-03-16
fix: validate sender for replies in ipcMainInternalUtils.invokeInWebContents()

miniak:
#### Description of Change
This prevents malicious webContents from replying to IPCs addressed to another webContents.
Follow up to [refactor: add ipcMainUtils.invokeInWebContents / ipcRendererUtils.handle helpers](https://github.com/electron/electron/pull/17313)

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/ipc-main-internal-utils.ts
+++ b/lib/browser/ipc-main-internal-utils.ts
@@ -29,9 +29,17 @@ let nextId = 0
 export function invokeInWebContents<T> (sender: Electron.WebContentsInternal, command: string, ...args: any[]) {
   return new Promise<T>((resolve, reject) => {
     const requestId = ++nextId
-    ipcMainInternal.once(`${command}_RESPONSE_${requestId}`, (
-      _event, error: Electron.SerializedError, result: any
-    ) => {
+    const channel = `${command}_RESPONSE_${requestId}`
+    ipcMainInternal.on(channel, function handler (
+      event, error: Electron.SerializedError, result: any
+    ) {
+      if (event.sender !== sender) {
+        console.error(`Reply to ${command} sent by unexpected WebContents (${event.sender.id})`)
+        return
+      }
+
+      ipcMainInternal.removeListener(channel, handler)
+
       if (error) {
         reject(errorUtils.deserialize(error))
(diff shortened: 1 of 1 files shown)


lib/browser/ipc-main-internal-utils.ts
```diff
@@ -29,9 +29,17 @@ let nextId = 0
...
   return new Promise<T>((resolve, reject) => {
     const requestId = ++nextId
-    ipcMainInternal.once(`${command}_RESPONSE_${requestId}`, (
-      _event, error: Electron.SerializedError, result: any
-    ) => {
+    const channel = `${command}_RESPONSE_${requestId}`
+    ipcMainInternal.on(channel, function handler (
+      event, error: Electron.SerializedError, result: any
+    ) {
+      if (event.sender !== sender) {
```
MarshallOfSound: Can we make this check more robust to potentially having multiple instances of a single `webContents` in JS land.

I.e. Check that neither `event.sender` or `sender` are destroyed and then check the `id` properties for equality 👍

miniak: @deepak1556, @zcbenz is this situation possible?

zcbenz: The `api::WebContents` (and its JS object) is bound to `content::WebContent`, they have a 1-to-1 relationship because it is commonly needed to get the JS object from a WebContents, so by design it is impossible to have multiple JS instances for one C++ WebContents .

miniak: so the simple comparison is enough then, right?

miniak: @MarshallOfSound I would suggest keeping the simple comparison `event.sender != sender` to keep the code simple

miniak:
@zcbenz can you please also leave a review?

lib/browser/ipc-main-internal-utils.ts
zcbenz: Yeah simple comparison is enough.

Issue #15791, opened 2018-11-21
Enabling network service


deepak1556:
This is a meta bug to track all the work required to use the network service from chromium, which is being experimented in canary and stable channels at the moment. To use the service, pass `--enable-features=NetworkService`

## What is network service ?

As a part of chrome modularization project, network code is being moved from browser process into a separate process.

High level Design doc: https://docs.google.com/document/d/1wAHLw9h7gGuqJNCgG1mP1BmLtCGfZ2pys-PdZQ1vg7M/edit?pref=2&pli=1#heading=h.w406cmqiibno

Upstream meta bug: https://crbug.com/598073

## Changes required in Electron

As we move to consuming the service, we will start to remove all usages from `//net` and use interfaces from `//services/network`. So existing terminologies like `URLRequest`, `URLRequestContext`, `URLRequestJob`, `URLFetcher` will now be `URLLoaderRequest`, `NetworkContext`, `URLLoaderJob`, `URLLoader`

* [x] Link network service code
  - Build `//services/network`
* [x] Create legacy `URLRequestContext` with network service glue code
  - Use `URLRequestContextBuilderMojo` to populate request context params
  - Upstream related changes: https://crbug.com/825242, https://crbug.com/715695
  - Changes made in PR: https://github.com/electron/electron/pull/14656
* [x] Use Cookie Manager to track cookie changes - **P1**
  - Upstream related changes: https://crbug.com/721395
  - Update `atom::api::Cookies` to use cookie store from network context
  - Changes made in PR: **TODO**  
* [x] Update proxy configuration with network service
  - Upstream related changes: https://crbug.com/715697, https://crbug.com/717253, https://crbug.com/754007
  - Changes made in PR: https://github.com/electron/electron/pull/14656
* [x] Update legacy `net::URLRequestContext` consumers in `atom::api::Session` module - **P2**
  - Upstream related changes: https://crbug.com/837333
  - Changes made in PR: **TODO**
* [x] Update `atom::api::WebRequest` module - **P2**
  - Upstream related changes: https://crbug.com/721414
  - Changes made in PR: **TODO**
* [x] Update `atom::api::net` module - **P3**
  - Convert `URLRequest` variations into `URLLoaderRequest`
  - Upstream example changes: https://crbug.com/715632
  - Remove patch: [expose_net_observer_api.patch](https://github.com/electron/electron/blob/master/patches/common/chromium/expose-net-observer-api.patch)
  - Changes made in PR: **TODO** 
[...]

PR #17431, opened 2019-03-18
refactor: Enable network service (Part 1)

deepak1556:
#### Description of Change

This is first in the series of network refactors to enable the network service as default. The PR allows creation of network context from the service rather than using the legacy apis. There are no functionality changes in this PR, the existing code path are guarded behind network service feature flag so that we can remove the old code path incrementally.

Refs https://github.com/electron/electron/issues/15791

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


deepak1556:
@zcbenz once this is merged, lets spilt up the tasks in https://github.com/electron/electron/issues/15791#issuecomment-474922182

PR #17442, opened 2019-03-19
spec: add tests for the autoUpdater on macOS that actually test if it works

MarshallOfSound:
#### Description of Change
This is part of some ongoing work to increase confidence in our CI runs / releases.  Previously our `autoUpdater` specs only tested the methods didn't throw errors, or correctly returned the previously set `feedUrl`.

These specs actually go E2E in the update process to ensure that an app can actually successfully update if they are using the built in updater API.

These tests will only run if:
* You are running the test son macOS
* There is a valid code signing cert on your machine
* You are running in a non-component build (testing/release).  This means `debug` builds will not run these tests

If there is not a valid code signing cert on your machine the tests will be skipped **unless** running in CI mode in which case the tests will be marked as failed so that the fact their isn't a valid code signing cert on the machine can be addressed.

On CI I have added a self-signed code signing certificate to our build, this certificate is useless to anyone (as it's self-signed) and the password for the cert is securely stored in 1Password / CI secrets.

#### Release Notes

Notes: no-notes


spec-main/api-autoupdater-darwin-spec.ts
```diff
@@ -0,0 +1,263 @@
...
+import * as os from 'os'
+import * as path from 'path'
+import { AddressInfo } from 'net';
+
+const features = process.atomBinding('features')
+
+const fixturesPath = path.resolve(__dirname, '../spec/fixtures')
+
+// We can only test the auto updater on darwin non-component builds
+const describeFn = (process.platform === 'darwin' && !process.mas && !features.isComponentBuild() ? describe : describe.skip)
```
codebytere: i see the value in this style but i think it makes things a bit harder to read potentially 🤔

script/codesign/get-trusted-identity.sh
```diff
@@ -0,0 +1,13 @@
...
+set -e
+
+valid_certs=$(security find-identity -p codesigning -v)
+if [[ $valid_certs == *"1)"* ]]; then
+  first_valid_cert=$(echo $valid_certs | sed 's/ \".*//' | sed 's/.* //')
+  echo $first_valid_cert
+  exit 0
+else
+  # No Certificate
+  exit 0
```
nornagon: maybe `exit 1`?

script/codesign/import-testing-cert-ci.sh
```diff
@@ -0,0 +1,25 @@
...
+security import $(dirname $0)/signing.p12 -k $KEY_CHAIN -P $SPEC_KEY_PASSWORD -A /usr/bin/codesign
+
+echo "Add keychain to keychain-list"
+security list-keychains -s mac-build.keychain
+
+echo "Settting key partition list"
+security set-key-partition-list -S apple-tool:,apple: -s -k $KEYCHAIN_PASSWORD $KEY_CHAIN
+
+echo "Trusting self-signed certificate"
+sudo security trust-settings-import -d $(dirname $0)/trust-settings.plist
```
nornagon: i hope this file is right because i have no idea what any of these commands do

nornagon: maybe wrap your substitutions in quotes tho? e.g.
```suggestion
sudo security trust-settings-import -d "$(dirname $0)"/trust-settings.plist
```
& for the other vars

script/codesign/import-testing-cert-ci.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Add certificates to keychain and allow codesign to access them
+security import $(dirname $0)/signing.cer -k $KEY_CHAIN -A /usr/bin/codesign
+security import $(dirname $0)/signing.pem -k $KEY_CHAIN -A /usr/bin/codesign
+security import $(dirname $0)/signing.p12 -k $KEY_CHAIN -P $SPEC_KEY_PASSWORD -A /usr/bin/codesign
+
+echo "Add keychain to keychain-list"
+security list-keychains -s mac-build.keychain
+
+echo "Settting key partition list"
```
nornagon:
```suggestion
echo "Setting key partition list"
```

script/codesign/import-testing-cert-ci.sh
```diff
@@ -0,0 +1,25 @@
...
+
+KEY_CHAIN=mac-build.keychain
+KEYCHAIN_PASSWORD=unsafe_keychain_pass
+security create-keychain -p $KEYCHAIN_PASSWORD $KEY_CHAIN
+# Make the keychain the default so identities are found
+security default-keychain -s $KEY_CHAIN
+# Unlock the keychain
+security unlock-keychain -p $KEYCHAIN_PASSWORD $KEY_CHAIN
+# Set keychain locking timeout to 3600 seconds
+security set-keychain-settings -t 3600 -u $KEY_CHAIN
```
nornagon: why 3600?

spec-main/api-autoupdater-darwin-spec.ts
```diff
@@ -0,0 +1,263 @@
...
+
+describeFn('autoUpdater behavior', function () {
+  this.timeout(120000)
+
+  let identity = ''
+
+  beforeEach(function () {
+    const result = cp.spawnSync('bash', [path.resolve(__dirname, '../script/codesign/get-trusted-identity.sh')])
```
nornagon: i don't believe you need `bash` here if the .sh is `+x`

spec-main/api-autoupdater-darwin-spec.ts
```diff
@@ -0,0 +1,263 @@
...
+
+// We can only test the auto updater on darwin non-component builds
+const describeFn = (process.platform === 'darwin' && !process.mas && !features.isComponentBuild() ? describe : describe.skip)
+
+describeFn('autoUpdater behavior', function () {
+  this.timeout(120000)
+
+  let identity = ''
+
+  beforeEach(function () {
```
nornagon: does this need to be rerun before every test or can we just do it once at the beginning?

spec-main/api-autoupdater-darwin-spec.ts
```diff
@@ -0,0 +1,263 @@
...
+  })
+
+  it('should have a valid code signing identity', () => {
+    expect(identity).to.be.a('string').with.lengthOf.at.least(1)
+  })
+
+  const copyApp = async (newDir: string, fixture = 'initial') => {
+    const appBundlePath = path.resolve(process.execPath, '../../..')
+    const newPath = path.resolve(newDir, 'Electron.app')
+    cp.execSync(`cp -R "${appBundlePath}" "${path.dirname(newPath)}"`)
```
nornagon:
```suggestion
    cp.execSync("cp", ["-R", appBundlePath, path.dirname(newPath)])
```

also add a comment about why not use `fs.copy`

spec-main/api-autoupdater-darwin-spec.ts
```diff
@@ -0,0 +1,263 @@
...
+    const appDir = path.resolve(newPath, 'Contents/Resources/app')
+    await fs.mkdirp(appDir)
+    await fs.copy(path.resolve(fixturesPath, 'auto-update', fixture), appDir)
+    const plistPath = path.resolve(newPath, 'Contents', 'Info.plist')
+    await fs.writeFile(
+      plistPath,
+      (await fs.readFile(plistPath, 'utf8')).replace('<key>BuildMachineOSBuild</key>', `<key>NSAppTransportSecurity</key>
```
nornagon: gross but ok

spec-main/api-autoupdater-darwin-spec.ts
```diff
@@ -0,0 +1,263 @@
...
+    const child = cp.spawn(cmd, args, opts)
+    child.stdout.on('data', (chunk: Buffer) => {
+      out += chunk.toString()
+    })
+    child.stderr.on('data', (chunk: Buffer) => {
+      out += chunk.toString()
+    })
+    return await new Promise<{ code: number, out: string }>((resolve) => {
+      child.on('exit', (code, signal) => {
+        expect(signal).to.equal(null)
```
nornagon: does chai catch this thrown error, or should this be thrown through `reject`?

spec-main/api-autoupdater-darwin-spec.ts
```diff
@@ -0,0 +1,263 @@
...
+  const spawn = async (cmd: string, args: string[], opts: any = {}) => {
+    let out = ''
+    const child = cp.spawn(cmd, args, opts)
+    child.stdout.on('data', (chunk: Buffer) => {
+      out += chunk.toString()
+    })
+    child.stderr.on('data', (chunk: Buffer) => {
+      out += chunk.toString()
+    })
+    return await new Promise<{ code: number, out: string }>((resolve) => {
```
nornagon: no need to `return await new Promise`, just drop the `async` keyword from the function definition and `return new Promise`. also might as well wrap the whole function in `new Promise`.

spec-main/api-autoupdater-darwin-spec.ts
```diff
@@ -0,0 +1,263 @@
...
+        })
+      })
+    })
+  }
+
+  const signApp = (appPath: string) => {
+    return spawn('codesign', ['-s', identity, '--deep', '--force', appPath])
+  }
+
+  const launchApp = async (appPath: string, args: string[] = []) => {
```
nornagon:
```suggestion
  const launchApp = (appPath: string, args: string[] = []) => {
```

spec-main/api-autoupdater-darwin-spec.ts
```diff
@@ -0,0 +1,263 @@
...
+    return spawn(path.resolve(appPath, 'Contents/MacOS/Electron'), args)
+  }
+
+  const withTempDirectory = async (fn: (dir: string) => Promise<void>) => {
+    const dir = await fs.mkdtemp(path.resolve(os.tmpdir(), 'electron-update-spec-'))
+    try {
+      await fn(dir)
+    } catch (err) {
+      cp.execSync(`rm -r "${dir}"`)
```
nornagon:
```suggestion
      cp.execSync("rm", ["-rf", dir])
```

spec-main/api-autoupdater-darwin-spec.ts
```diff
@@ -0,0 +1,263 @@
...
+      throw err
+    }
+    cp.execSync(`rm -r "${dir}"`)
```
nornagon: use `finally` instead of `catch` to deduplicate this call.

script/codesign/get-trusted-identity.sh
MarshallOfSound: `exit 1` causes things like `execSync` to throw instead of returning.  I exit with code `0` here to avoid having to try catch and instead inspect the output (like we would have to anyway)

script/codesign/import-testing-cert-ci.sh
MarshallOfSound: @nornagon Local devs should never have to (or should) run this script, it should only be run on CI 👍 

I'll check that quotes thing

script/codesign/import-testing-cert-ci.sh
MarshallOfSound: It's a substantially long enough period of time (an hour) that there's no way the keychain will relock before the tests run on the CI box.

spec-main/api-autoupdater-darwin-spec.ts
MarshallOfSound: @nornagon You'd think so, but it worked on CI without `bash` and blew up locally.  This seemed like a reasonable option to ensure it worked everywhere.

spec-main/api-autoupdater-darwin-spec.ts
MarshallOfSound: I did it in a `beforeEach` so that `throw new Error('No valid signing identity available to run autoUpdater specs')` runs on every test and marks them all as failed (6 failed tests) instead of a `before` which would cause (1 failed test) and be less obvious as a legit failure.

The time taken to run that script is a few ms, no noticeable performance impact when the tests themselves like a number of seconds to run each time

spec-main/api-autoupdater-darwin-spec.ts
MarshallOfSound: @nornagon `exec` does not take in an array of arguments, it only takes in an entire command string to run

spec-main/api-autoupdater-darwin-spec.ts
MarshallOfSound: `chai` catches this 👍

spec-main/api-autoupdater-darwin-spec.ts
MarshallOfSound: Same as above, that's not how `exec` works 👍

MarshallOfSound:
> Is there any particular reason to have a passphrase on the private key?

Why not when we can 😄 Hypotheticals come to mind around our CI machine and it trusting things signed with that cert and so on 🤔 Better safe than sorry 👍

spec-main/api-autoupdater-darwin-spec.ts
nornagon: blew up how? that seems relevant

spec-main/api-autoupdater-darwin-spec.ts
nornagon: can we instead use a function that does take an array of arguments then? `spawn`?

spec-main/api-autoupdater-darwin-spec.ts
MarshallOfSound: `ENOENT` iirc

script/codesign/import-testing-cert-ci.sh
nornagon: Yep, I'm sure our tests will never ever take an hour to run :laughcry:

spec-main/api-autoupdater-darwin-spec.ts
nornagon: IMO this is of dubious value and rather seems to me like a bug in mocha that failing a `before` only shows up as a single failure rather than N failures.

PR #17443, opened 2019-03-19
fix: check the result when calling js function

zcbenz:
#### Description of Change

When a JS function returns `undefined`, the returned `v8::MaybeLocal` in C++ would be empty, and calling `ToLocalChecked()` without checking would crash immediately.

This PR does 2 things:
1. Check whether `MaybeLocal` is empty before using it;
2. Replace `ToLocalChecked` with `IsEmpty` where the only purpose is to suppress compiler warning.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix crash when passing callbacks which return `undefined` to some APIs.


codebytere:
@zcbenz this potentially has broken:
```js
webContents module getAllWebContents() API returns an array of web contents - returns an array of web contents
```

zcbenz:
> @zcbenz this potentially has broken:

It seems to be a flaky one, the CI is green after restarting.

PR #17444, opened 2019-03-19
fix: FrameSubscriber should not scale frame down

zcbenz:
#### Description of Change

The `beginFrameSubscription` currently passes the DIP size to Chromium, which would then scale the frame down and return blurred image on HiDPI displays.

This PR changes the code to pass pixel size, so the API works correctly on HiDPI displays.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix `beginFrameSubscription` returning blurred images on HiDPI displays.


MarshallOfSound:
@zcbenz This is now in conflict 😢

zcbenz:
I have rebased the branch.

Issue #17399, opened 2019-03-15
getLastCrashReport doesn't return if uploadToServer false


issacgerges:
* **Electron Version:**
  * 4.0.0
* **Operating System:**
  * MacOS
* **Last Known Working Electron version:**:
  * n/a

### Expected Behavior
I would expect `getLastCrashReport` to return the latest crashReport, even if un-uploaded

### Actual Behavior
`getLastCrashReport` returns `null`, even though a crash was triggered with `process.crash()` and crash reports are present in the `crashDirectory`

### To Reproduce
The following can be loaded in electron fiddle to demonstrate the issue

https://gist.github.com/4f69a92b0d061a65eae07774fba73d68

### Additional Information
There is a possibility this is by design, if so we may want to rename the api method to make it clear, ie: `getLastUploadedCrashReport`

issacgerges:
It looks like this may be as-designed, see https://github.com/electron/electron/blob/6d674eebb12123aa79c2e52701980568770743c1/lib/common/crash-reporter.js#L60

codebytere:
hmm, per docs we should be doing:
```
Returns the date and ID of the last crash report. If no crash reports have been
sent or the crash reporter has not been started, `null` is returned.
```
i'll update this to return null, as i think we should be conforming to what's expected in docs.

PR #17458, opened 2019-03-19
docs: clarify documentation for getLastCrashReport

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/17399.

At present, it's not clear that `getLastCrashReport` _only_ returns crash reports that have been uploaded, and so this PR add clarifying information to documentation.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Clarified usage of `crashReporter.getLastCrashReport()` in documentation.

### DIFF
--- a/docs/api/crash-reporter.md
+++ b/docs/api/crash-reporter.md
@@ -96,7 +96,7 @@ them will get reported without `companyName`, `productName` or any of the `extra
 Returns [`CrashReport`](structures/crash-report.md):
 
 Returns the date and ID of the last crash report. If no crash reports have been
-sent or the crash reporter has not been started, `null` is returned.
+sent (i.e a crash was triggered with `process.crash()` and crash reports are present in the `crashDirectory`, but have not been explicitly uploaded) or the crash reporter has not been started, `null` is returned.
 
 ### `crashReporter.getUploadedReports()`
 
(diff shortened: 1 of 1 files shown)


codebytere:
hmm, i agree; if we think that just being really explicit in the documentation would address the same root problem then i'm happy with that tbqh

what do you think @MarshallOfSound?

MarshallOfSound:
@codebytere Adding explicit docs would be way better for both newcomers to the API and reducing the churn on existing users :+1:

codebytere:
@MarshallOfSound updated!

docs/api/crash-reporter.md
```diff
@@ -96,7 +96,7 @@ them will get reported without `companyName`, `productName` or any of the `extra
 Returns [`CrashReport`](structures/crash-report.md):
 
 Returns the date and ID of the last crash report. If no crash reports have been
-sent or the crash reporter has not been started, `null` is returned.
+sent (i.e a crash was triggered with `process.crash()` and crash reports are present in the `crashDirectory`, but have not been explicitly uploaded) or the crash reporter has not been started, `null` is returned.
```
MarshallOfSound: I think mentioning `process.crash()` here might throw people a bit off the trail, maybe just saying something along the lines of "only crash reports that have been uploaded will be returned, even if a crash report is present on disk it will not be returned until it is uploaded", just slightly better than that 😆

PR #17459, opened 2019-03-19
fix: crashReporter.getParameters() takes no params

codebytere:
#### Description of Change

Fixes an issue in `crashReporter` whereby `getParameters` took `key` and `val` as parameters and should take none. Also cleans up `crashReporter.start(options)` with default parameters.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/lib/common/crash-reporter.js
+++ b/lib/common/crash-reporter.js
@@ -15,29 +15,17 @@ class CrashReporter {
   start (options) {
     if (options == null) options = {}
 
-    let {
+    const {
       productName,
       companyName,
-      extra,
-      ignoreSystemCrashHandler,
+      extra = {},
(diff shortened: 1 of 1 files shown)


ckerr:
@MarshallOfSound have those new crash reporter specs landed yet? If not, could you link to them in this PR so it's easy to see when this can land?

miniak:
@codebytere this does not seem to be a user facing change, unless you check `crashReporter. getParameters.length`

MarshallOfSound:
@ckerr @codebytere I'm going to unblock this, I've got a lot on my plate at the moment and it doesn't look like I'll be getting back to crash reporter specs for a bit.  :shipit: at will

PR #4869, opened 2016-03-21
Run callback of setDestructor immediately when GC happens

zcbenz:
@mitchhentges Can you try if this fixes #4733 for you? I can't reproduce the original issue so I'm not able to try it myself.

Issue #4733, opened 2016-03-09
Exception in objects-registry when doing remote work


CharlieHess:
* Electron version: 0.36.9
* Operating system: Mac

We're still trying to nail down repro steps in Slack, but it seems to happen on 0.36.9 when calling `remote` methods. Specifically, we build context menus using `remote` (very similar to the [`Menu`](https://github.com/atom/electron/blob/master/docs/api/menu.md) sample) and occasionally they stop working with this exception:
```
Uncaught Error: Cannot read property 'object' of undefined
TypeError: Cannot read property 'object' of undefined
  at ObjectsRegistry.get (C:\Users\paulb\AppData\Local\slack\app-2.0.2\resources\atom.asar\browser\lib\objects-registry.js:44:28)
  at metaToValue (C:\Users\paulb\AppData\Local\slack\app-2.0.2\resources\atom.asar\browser\lib\rpc-server.js:155:32)
  at Array.map (native)
  at unwrapArgs (C:\Users\paulb\AppData\Local\slack\app-2.0.2\resources\atom.asar\browser\lib\rpc-server.js:214:15)
  at EventEmitter.<anonymous> (C:\Users\paulb\AppData\Local\slack\app-2.0.2\resources\atom.asar\browser\lib\rpc-server.js:317:12)
  at emitMany (events.js:108:13)
  at EventEmitter.emit (events.js:182:7)
  at EventEmitter.<anonymous> (C:\Users\paulb\AppData\Local\slack\app-2.0.2\resources\atom.asar\browser\api\lib\web-contents.js:133:25)
  at emitTwo (events.js:87:13)
  at EventEmitter.emit (events.js:172:7)
```

zcbenz:
It seems that the renderer process is referencing an object in the main process which has already been removed. It should be related to https://github.com/atom/electron/pull/4570, probably there is something wrong with my bookkeeping logic.

vulture:
We're also getting this (probably same) issue, while calling win.setOverlayIcon(). Only happened after updating our Electron to latest master yesterday (previous version from a month or so ago didn't have this problem I think).

Uncaught Error: Cannot read property 'object' of undefined
TypeError: Cannot read property 'object' of undefined
    at ObjectsRegistry.get (Electron\resources\atom.asar\browser\lib\objects-registry.js:44:28)
    at EventEmitter.<anonymous> (Electron\resources\atom.asar\browser\lib\rpc-server.js:318:31)
    at emitMany (events.js:108:13)
[...]

PR #17464, opened 2019-03-19
fix: maintain a ref count for objects sent over remote

### LINKED ISSUE #17039: remote API is flaking on 5.0.x
The current `5-0-x` branch and `master` are experiencing random `remote` API failures again.  We should look into this and fix before 5.0 stable.

Example: https://circleci.com/gh/electron/electron/138325

MarshallOfSound:
#### Remote Fix
See https://github.com/electron/electron/pull/17464#issuecomment-483506489 for explanation

Fixes #17039

Refs: https://chromium-review.googlesource.com/c/chromium/src/+/1514516

This also includes the test change as that is the "repro" for this issue 😕 

#### Testing Changes
Previously when this assertion failed all tests that ran after the failed assertion also failed.  This ensure that the assertion fails for the test that actually caused the issue but cleans up the left-over windows so that future tests do not fail.

#### Release Notes

Notes: Fixed race condition where the `remote` module would sometimes fail to fetch properties of a remote object

### DIFF
--- a/atom/common/api/remote_object_freer.cc
+++ b/atom/common/api/remote_object_freer.cc
@@ -36,6 +36,14 @@ void RemoteObjectFreer::BindTo(v8::Isolate* isolate,
   new RemoteObjectFreer(isolate, target, context_id, object_id);
 }
 
+// static
+void RemoteObjectFreer::AddRef(const std::string& context_id, int object_id) {
+  ref_mapper_[context_id][object_id]++;
+}
+
+// static
+std::map<std::string, std::map<int, int>> RemoteObjectFreer::ref_mapper_;
+
 RemoteObjectFreer::RemoteObjectFreer(v8::Isolate* isolate,
                                      v8::Local<v8::Object> target,
                                      const std::string& context_id,
@@ -62,6 +70,12 @@ void RemoteObjectFreer::RunDestructor() {
   base::ListValue args;
   args.AppendString(context_id_);
   args.AppendInteger(object_id_);
+  args.AppendInteger(ref_mapper_[context_id_][object_id_]);
+  // Reset our local ref count in case we are in a GC race condition and will
+  // get more references in an inbound IPC message
+  ref_mapper_[context_id_].erase(object_id_);
+  if (ref_mapper_[context_id_].empty())
+    ref_mapper_.erase(context_id_);
 
   mojom::ElectronBrowserAssociatedPtr electron_ptr;
   render_frame->GetRemoteAssociatedInterfaces()->GetInterface(
--- a/atom/common/api/remote_object_freer.h
+++ b/atom/common/api/remote_object_freer.h
@@ -5,6 +5,7 @@
 #ifndef ATOM_COMMON_API_REMOTE_OBJECT_FREER_H_
 #define ATOM_COMMON_API_REMOTE_OBJECT_FREER_H_
 
+#include <map>
 #include <string>
 
 #include "atom/common/api/object_life_monitor.h"
@@ -17,6 +18,7 @@ class RemoteObjectFreer : public ObjectLifeMonitor {
                      v8::Local<v8::Object> target,
                      const std::string& context_id,
                      int object_id);
+  static void AddRef(const std::string& context_id, int object_id);
 
  protected:
   RemoteObjectFreer(v8::Isolate* isolate,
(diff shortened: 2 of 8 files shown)


zcbenz:
This would trigger another regression. Consider following situation:

1. `remote.getCurrentWindow()` to create remote object A for current window
2. Garbage collection
3. `OnObjectGC` called for remote object A, `RunDestructor` schedule for next tick
4. `remote.getCurrentWindow()` to create remote object B for current window
5. `RunDestructor` gets called, and the `objectRegistry` in the main process clears the entry for current window
6. Any operations on remote object B now ends up with `missing remote object` error

See https://github.com/electron/electron/issues/4733#issuecomment-198197090 for more history on the original regression, and #4869 for the fix.

I think this PR fixes the CI failure because it deferred the `RunDestructor` so the actual problem is now hidden.

MarshallOfSound:
@zcbenz I have updated the PR with a (IMO) better approach to ensuring that we only ever remove an object from the cache if all sent references have been dereffed.

MarshallOfSound:
IMG_20190415_190716

This is the exact race condition we were hitting.  In a rare case when a remote object was fetched a second time and a GC occured at the right time (between the main process sending the remote object metadata and the renderer process pulling the proxy out of a weak map) the renderer would tell the main process that the object had been GC'ed but would then on the next tick receive a remote message telling it to use the object it just free'ed.

I've included the commit message below for more info:

> Previously there was a race condition where a GC could occur in the
renderer process between the main process sending a meta.id and the
renderer pulling the proxy out its weakmap to stop it being GC'ed.

> This fixes that race condition by maintaining a "sent" ref count in the
object registry and a "received" ref count in the object cache on the
renderer side.  The deref request now sends the number of refs the
renderer thinks it owns, if the number does not match the value in the
object registry it is assumed that there is an IPC message containing a
new reference in flight and this race condition was hit.

> The browser side ref count is then reduced and we wait for the new deref
message.  This guaruntees that an object will only be removed from the
registry if every reference we sent has been guarunteed to be unreffed.

atom/common/api/object_life_monitor.cc
```diff
@@ -7,6 +7,7 @@
 
 #include "base/bind.h"
 #include "base/message_loop/message_loop.h"
+#include "third_party/blink/public/platform/task_type.h"
```
zcbenz: Changes to this file can be removed .

atom/common/api/remote_object_freer.cc
```diff
@@ -62,6 +70,10 @@ void RemoteObjectFreer::RunDestructor() {
   base::ListValue args;
   args.AppendString(context_id_);
   args.AppendInteger(object_id_);
+  args.AppendInteger(ref_mapper_[context_id_][object_id_]);
+  // Reset our local ref count in case we are in a GC race condition and will
+  // get more references in an inbound IPC message
+  ref_mapper_[context_id_][object_id_] = 0;
```
zcbenz: The key should be erased otherwise memory would not be freed.

```c++
ref_mapper_[context_id_].erase(object_id_);
if (ref_mapper_[context_id_].empty())
  ref_mapper_.erase(context_id_);
```

atom/common/api/remote_object_freer.h
```diff
@@ -27,6 +29,8 @@ class RemoteObjectFreer : public ObjectLifeMonitor {
 
   void RunDestructor() override;
 
+  static std::map<std::string, std::map<int, int>> ref_mapper_;
```
zcbenz: Can you add a comment describing what the key and value are?

lib/browser/objects-registry.js
```diff
@@ -49,14 +51,28 @@ class ObjectsRegistry {
   // Dereference an object according to its ID.
   // Note that an object may be double-freed (cleared when page is reloaded, and
   // then garbage collected in old page).
-  remove (webContents, contextId, id) {
+  // rendererSideRefCount is the ref count that the renderer process reported
+  // at time of GC if this is different to the number of references we sent to
+  // the given owner then a GC occurred between a ref being sent and the value
+  // being pulled out of the weak map.
+  // In this case we decrement out ref count and do not delete the stored
+  // object
```
zcbenz: Can you add the link of this PR in the comment as it describes why we introduced `rendererSideRefCount`?

```md
For more details on why we introduced rendererSideRefCount, please see https://...
```

Otherwise some other developer may want to optimize this code out a few years later.

MarshallOfSound:
Note to self: test failures are because I didn't update https://github.com/electron/electron/blob/master/lib/browser/objects-registry.js#L69

Secondary note to self: typescript would have caught this 😆

zcbenz:
This test is failing:

```
ipc main module remote objects registry does not dereference until the render view is deleted (regression) - does not dereference until the render view is deleted (regression)
```

I guess it is because the code of `clear` method has not been updated for the new type of `this.owners`:

```js
for (const id of owner) this.dereference(id)
```

PR #17468, opened 2019-03-19
docs: update sandbox-option.md

miniak:
#### Description of Change
- update list of modules available in sandbox
- replace `--enable-sandbox` command-line argument with `app.enableSandbox`

Follow up to:
- [chore: remove deprecated modules internally using remote.require in sandboxed renderer context](https://github.com/electron/electron/pull/15957)
- [feat: enable mixed-sandbox mode by default](https://github.com/electron/electron/pull/15894)

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

/cc @electron/wg-security

### DIFF
--- a/docs/api/sandbox-option.md
+++ b/docs/api/sandbox-option.md
@@ -1,6 +1,6 @@
 # `sandbox` Option
 
-> Create a browser window with a renderer that can run inside Chromium OS sandbox. With this
+> Create a browser window with a sandboxed renderer. With this
 option enabled, the renderer must communicate via IPC to the main process in order to access node APIs.
 
 One of the key security features of Chromium is that all blink rendering/JavaScript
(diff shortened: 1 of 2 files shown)


docs/api/sandbox-option.md
```diff
@@ -56,35 +54,20 @@ only via IPC. The use of this option stops Electron from creating a Node.js runt
...
-```
-
-It is not possible to have the OS sandbox active only for some renderers, if
-`--enable-sandbox` is enabled, normal Electron windows cannot be created.
-
-If you need to mix sandboxed and non-sandboxed renderers in one application,
-omit the `--enable-sandbox` argument. Without this argument, windows
-created with `sandbox: true` will still have Node.js disabled and communicate
-only via IPC, which by itself is already a gain from security POV.
+**Note:** It is not possible to have the OS sandbox active only for some renderers, if
```
ppontes: Found this note confusingly unnecessary as it seems to be saying "if you flip the switch to B, then you won't get A". If app.enableSandbox() is not called, then we do have the possibility of enabling the OS sandbox _only_ for the renderers of windows that you start with `sandbox: true`. It is the purpose of `enableSandbox()` to force that all renderers are OS sandboxed, so I don't see the need for this note and would remove it.

miniak: removed

PR #17479, opened 2019-03-20
docs: fully document MenuItem instance properties

### LINKED ISSUE #15000: Typescript bindings for MenuItem lacks 'id' and other
Electron 3.0.2.

Typescript bindings for MenuItem lacks 'id' and other attributes... 

Casting it to 'any' works and printing the object shows all the fields of a proper MenuItem:

```
[ MenuItem {
    id: 'view',
    label: 'View',
    submenu:
     Menu {
       commandsMap: [Object],
       groupsMap: {},
       items: [Array],
       delegate: [Object] },
    type: 'submenu',
    role: null,
    accelerator: null,
    icon: null,
    sublabel: '',
    enabled: true,
    visible: true,
    checked: false,
    commandId: 51,
    click: [Function],
    menu:
     Menu {
       commandsMap: [Object],
       groupsMap: {},
       items: [Array],
       delegate: [Object] } } ]


```

... but the typescript declaration is just:

```

  class MenuItem {

    // Docs: http://electron.atom.io/docs/api/menu-item

    constructor(options: MenuItemConstructorOptions);
    checked: boolean;
    click: Function;
    enabled: boolean;
    label: string;
    visible: boolean;
  }
```

nixxquality:
#### Description of Change

Closes #15000.

I've documented the instance properties of MenuItem so that typescript typings will work as intended.

#### Checklist

#### Release Notes
Notes: no-notes

However, I think I need to iterate on it. I am not sure if properties have the correct `| null` and `readonly` typings right now.
I'm requesting the assistance of someone familiar with the automatic typing system to tell me how to specify that correctly.

### DIFF
--- a/docs/api/menu-item.md
+++ b/docs/api/menu-item.md
@@ -122,6 +122,46 @@ Lowercase `role`, e.g. `toggledevtools`, is still supported.
 
 The following properties are available on instances of `MenuItem`:
 
+#### `menuItem.id`
+
+A `String` indicating the item's unique id, this property can be
+dynamically changed.
+
+#### `menuItem.label`
+
+A `String` indicating the item's visible label, this property can be
+dynamically changed.
+
+#### `menuItem.click`
+
+A `Function` that is fired when the MenuItem receives a click event.
+
+#### `menuItem.submenu`
+
+A `MenuItemConstructorOptions[] | Menu` (optional) containing the menu
+item's submenu, if present.
+
+#### `menuItem.type`
+
+A `String` indicating the type of the item.
+
+#### `menuItem.role`
+
+A `String` (optional) indicating the item's role, if set.
+
+#### `menuItem.accelerator`
+
+A `String` (optional) indicating the item's accelerator, if set.
+
+#### `menuItem.icon`
+
+A `[NativeImage](native-image.md) | String` (optional) indicating the
+item's icon, if set.
+
+#### `menuItem.sublabel`
+
+A `String` indicating the item's sublabel, this property can be dynamically changed.
+
 #### `menuItem.enabled`
 
 A `Boolean` indicating whether the item is enabled, this property can be
@@ -145,10 +185,15 @@ will turn off that property for all adjacent items in the same menu.
 
 You can add a `click` function for additional behavior.
 
-#### `menuItem.label`
+#### `menuItem.registerAccelerator`
 
-A `String` representing the menu items visible label.
+A `Boolean` indicating if the accelerator should be registered with the
+system or just displayed, this property can be dynamically changed.
 
-#### `menuItem.click`
+#### `menuItem.commandId`
 
-A `Function` that is fired when the MenuItem receives a click event.
+A `Number` indicating an item's sequential unique id.
+
+#### `menuItem.menu`
+
+A `[Menu](#menu.md))` that the item is a part of.
--- a/docs/api/menu-item.md
+++ b/docs/api/menu-item.md
@@ -135,6 +135,10 @@ dynamically changed.
 #### `menuItem.click`
 
 A `Function` that is fired when the MenuItem receives a click event.
+It can be called with `menuItem.click(event, focusedWindow, focusedWebContents)`.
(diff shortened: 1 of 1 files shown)


nixxquality:
https://github.com/electron/electron/pull/17479/checks?check_run_id=81497065
I can see the issues here, but I'm not sure how to fix them.
- Properties that should be marked readonly, aren't.
- Properties that should be marked as optional (`| null`), aren't.
- Click should be MenuTemplate["Click"], if possible.
- `icon` and `menu` are both `any`.
- `submenu` is only `MenuItemConstructorOptions[]`

codebytere:
hey thanks @nixxquality, this is super helpful! If you clone down electron, you can actually see if things were generated correctly without needing to build electron or anything. Just clone, add your changes, and run `npm create-typescript-definitions`. You'll see the resulting changes in `electron.d.ts` and be able to verify their correctness that way.

nixxquality:
That's useful, but I'm still in the dark how to get the definitions correct.
I tried looking at the style guide earlier but my questions were mostly unanswered.

docs/api/menu-item.md
```diff
@@ -122,6 +122,46 @@ Lowercase `role`, e.g. `toggledevtools`, is still supported.
...
+
+A `String` (optional) indicating the item's role, if set.
+
+#### `menuItem.accelerator`
+
+A `String` (optional) indicating the item's accelerator, if set.
+
+#### `menuItem.icon`
+
+A `[NativeImage](native-image.md) | String` (optional) indicating the
```
MarshallOfSound: We do not support linking `native-image` in this part of the docs-parser.  I'd suggest changing this to `NativeImage | String`

nixxquality: I changed it to `NativeImage | String` locally but it only comes out as `NativeImage` in the definitions.

MarshallOfSound: @nixxquality I guess the docs parser hasn't been made to handle the multi-type case for properties then 🤔

ckerr: If the docs are correct with `NativeImage | String`, maybe let's open a separate issue for the parser, but not block on it here?

MarshallOfSound: Sounds good to me, let's ship correct docs and fix the generator async

nixxquality:
Any suggestions on the other concerns in https://github.com/electron/electron/pull/17479#issuecomment-474855105, or should those be separate issues as well?

docs/api/menu-item.md
```diff
@@ -145,10 +185,15 @@ will turn off that property for all adjacent items in the same menu.
 
 You can add a `click` function for additional behavior.
 
-#### `menuItem.label`
```
MarshallOfSound: Why are we removing the `label` docs?

I agree that we should remove the docs for `click`.  The `click` property isn't actually the `click` function you pass to the constructor but `label` appears to both exist and be correct 👍

MarshallOfSound:
> Properties that should be marked readonly, aren't.

Can you raise an issue on `electron-docs-linter`?

> Properties that should be marked as optional (| null), aren't.

Also an issue on `electron-docs-linter`

> Click should be MenuTemplate["Click"], if possible.

I don't think this is 100% correct, the `click` property on `MenuItem` is not exactly the function you passed in to the `MenuItem` constructor

nixxquality:
I've seen optional parameters on other data structures before.
Just confirming there is no way to mark an instance property as optional or readonly right now?

MarshallOfSound:
@nixxquality You can mark a parameter as optional, there is no support for detecting "readonly"

nixxquality:
I haven't figured out how to mark an *instance property* as optional if so.
Neither
```
A `String` (optional) ...
# nor
A `String (optional)` ...
```
works.

MarshallOfSound:
@nixxquality Hm, I thought we just scanned the entire description for the `(optional)` tag but I guess not. Raise an issue for it on `docs-linter` I guess

nixxquality:
Fair enough.

Issue #15, opened 2013-05-29
Replace underscores with dashes in file names


probablycorey:
In Atom we use `-` instead of `_` to separate words. Should we keep the same convention with file names in atom-shell. For example, should we rename `browser_window.coffee` to `browser-window.coffee`?

PR #17480, opened 2019-03-20
 docs: add remote module to docs/tutorial/security.md

miniak:
#### Description of Change
With @ppontes help, we finally put together a few words on the `remote` module in the security guide. Follow up to: https://github.com/electron/electron/pull/13028, https://github.com/electron/electron/pull/15014 and https://github.com/electron/electron/pull/16293

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

/cc @electron/wg-security

### DIFF
--- a/docs/tutorial/security.md
+++ b/docs/tutorial/security.md
@@ -782,7 +782,10 @@ provided by the main process modules may be dangerous in the hands of
 malicious code running in a compromised renderer process. By limiting
 the set of accessible modules to the minimum that your app needs and
 filtering out the others, you reduce the toolset that malicious code
-can use to attack the system.
+can use to attack the system. Note that you'll still be vulnerable to
+"prototype pollution" attacks on the remote module internal mechanism,
+so [disabling it fully](#15-disable-the-remote-module) is always the
+safest option.
 
 ### How?
 
--- a/docs/tutorial/security.md
+++ b/docs/tutorial/security.md
@@ -782,10 +782,13 @@ provided by the main process modules may be dangerous in the hands of
 malicious code running in a compromised renderer process. By limiting
 the set of accessible modules to the minimum that your app needs and
 filtering out the others, you reduce the toolset that malicious code
-can use to attack the system. Note that you'll still be vulnerable to
-"prototype pollution" attacks on the remote module internal mechanism,
-so [disabling it fully](#15-disable-the-remote-module) is always the
-safest option.
+can use to attack the system.
+
+Note that the safest option is to
+[fully disable the remote module](#15-disable-the-remote-module). If
+you choose to filter access rather than completely disable the module,
+you must be very careful to ensure that no escalation of privilege is
+possible through the modules you allow past the filter.
 
 ### How?
 
--- a/docs/tutorial/security.md
+++ b/docs/tutorial/security.md
@@ -109,6 +109,8 @@ You should at least follow these steps to improve the security of your applicati
 12. [Disable or limit navigation](#12-disable-or-limit-navigation)
 13. [Disable or limit creation of new windows](#13-disable-or-limit-creation-of-new-windows)
 14. [Do not use `openExternal` with untrusted content](#14-do-not-use-openexternal-with-untrusted-content)
+15. [Disable the `remote` module](#15-disable-the-remote-module)
+16. [Filter the `remote` module](#16-filter-the-remote-module)
 
 To automate the detection of misconfigurations and insecure patterns, it is
 possible to use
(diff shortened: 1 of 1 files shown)


docs/tutorial/security.md
```diff
@@ -709,6 +709,97 @@ const { shell } = require('electron')
 shell.openExternal('https://example.com/index.html')
 ```
 
+## 15) Disable the `remote` module
+
+The `remote` module provides a simple way to do inter-process communication (IPC) between the renderer process (web page) and the main process. Using it, a renderer can invoke methods of a main process object without explicitly sending inter-process messages. If your desktop application does not run untrusted content, this is a very elegant way to have your renderer processes access and work with modules that are only available to the main process, such as GUI-related modules (dialog, menu etc.).
```
miniak: provides a simple way for the renderer processes to access APIs normally only available in the main process

miniak: I would also omit web page, it can just be the preload script calling those APIs

docs/tutorial/security.md
```diff
@@ -709,6 +709,97 @@ const { shell } = require('electron')
...
+
+However, if your app can run untrusted content and you sandbox your renderer processes accordingly, the `remote` module makes it easy for malicious code to escape the sandbox and have access to system resources via the higher privileges of the main process. Therefore, it should be disabled in such circumstances.
```
miniak: **even if** you sandbox your renderer processes

docs/tutorial/security.md
```diff
@@ -709,6 +709,97 @@ const { shell } = require('electron')
...
+<!-- Bad -->
+<webview src="page.html"></webview>
+
+<!-- Good -->
+<webview enableremotemodule="false" src="page.html"></webview>
+```
+
+## 16) Filter the `remote` module
+
+If you cannot disable the `remote` module, you should filter the `remote` modules that your application does not require. This can be done by blocking certain modules entirely and by replacing others with proxies that expose only the functionality that your app needs.
```
miniak: you should filter the globals, node and electron modules (so-called built-ins) accessible via `remote`

docs/tutorial/security.md
```diff
@@ -48,21 +48,21 @@ the result of the overall security of the framework foundation
 your code. As such, it is your responsibility to follow a few important best
 practices:
 
-* **Keep your application up-to-date with the latest Electron framework release.**
```
miniak: even though I don't like trailing whitespace, we should not include this change

felixrieseberg:
This is good stuff and I like it. Do we also want a warning to go with it? Happy to write one.

docs/tutorial/security.md
ppontes: Restored.

docs/tutorial/security.md
ppontes: Copied this from the `remote` documentation 😁 . But I agree with your changes, amended.

docs/tutorial/security.md
ppontes: Excellent point, fixed.

docs/tutorial/security.md
ppontes: 👍🏻

docs/tutorial/security.md
```diff
@@ -709,6 +711,124 @@ const { shell } = require('electron')
 shell.openExternal('https://example.com/index.html')
 ```
 
+## 15) Disable the `remote` module
+
+The `remote` module provides a simple way for the renderer processes to
+access APIsnormally only available in the main process. Using it, a
```
ckerr:
```suggestion
access APIs normally only available in the main process. Using it, a
```

docs/tutorial/security.md
```diff
@@ -709,6 +711,124 @@ const { shell } = require('electron')
...
+renderer can invoke methods of a main process object without explicitly sending
+inter-process messages. If your desktop application does not run untrusted
+content, this is a very elegant way to have your renderer processes access and
+work with modules that are only available to the main process, such as
+GUI-related modules (dialog, menu etc.).
```
ckerr:
```suggestion
GUI-related modules (dialogs, menus, etc.).
```

docs/tutorial/security.md
```diff
@@ -709,6 +711,124 @@ const { shell } = require('electron')
...
+Disabling the `remote` module eliminates these attack vectors. Enabling
+context isolation also prevents the "prototype pollution" attacks from
+succeeding.
+
+### How?
+
+Sandboxed renderers
+
+```js
+// Bad
```
ckerr: The bad/good examples make sense here & in the rest of the docs, but I do suspect a lot of people will ignore the text and skip straight down to this.

For that reason, I might modify the comment a little:

```suggestion
// Bad if the renderer can run untrusted content
```

docs/tutorial/security.md
```diff
@@ -709,6 +711,124 @@ const { shell } = require('electron')
...
+// Good
+const mainWindow = new BrowserWindow({
+  webPreferences: {
+    enableRemoteModule: false
+  }
+})
+```
+
+```html
+<!-- Bad -->
```
ckerr:
```suggestion
<!-- Bad if the renderer can run untrusted content  -->
```

docs/tutorial/security.md
```diff
@@ -709,6 +711,124 @@ const { shell } = require('electron')
...
+<webview src="page.html"></webview>
+
+<!-- Good -->
+<webview enableremotemodule="false" src="page.html"></webview>
+```
+
+## 16) Filter the `remote` module
+
+If you cannot disable the `remote` module, you should filter the globals,
+node and electron modules (so-called built-ins) accessible via `remote`
```
ckerr:
```suggestion
Node, and Electron modules (so-called built-ins) accessible via `remote`
```

ckerr:
> This is good stuff and I like it. Do we also want a warning to go with it? Happy to write one.

@felixrieseberg is there a reliable way of knowing whether the app can load untrusted content? If so, then yes.

miniak:
@felixrieseberg I would appreciate if you could do that!

docs/tutorial/security.md
```diff
@@ -709,6 +711,124 @@ const { shell } = require('electron')
...
+sandboxed renderer. Leaking `remote` arms malicious code with a multitude
+of main process modules with which to perform an attack.
+
+Disabling the `remote` module eliminates these attack vectors. Enabling
+context isolation also prevents the "prototype pollution" attacks from
+succeeding.
+
+### How?
+
+Sandboxed renderers
```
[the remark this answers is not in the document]
miniak: removed

docs/tutorial/security.md
```diff
@@ -782,7 +782,10 @@ provided by the main process modules may be dangerous in the hands of
 malicious code running in a compromised renderer process. By limiting
 the set of accessible modules to the minimum that your app needs and
 filtering out the others, you reduce the toolset that malicious code
-can use to attack the system.
+can use to attack the system. Note that you'll still be vulnerable to
+"prototype pollution" attacks on the remote module internal mechanism,
```
nornagon: Hm, to me this implies that there are known prototype pollution bugs in the remote module mechanism, which isn't the case. Perhaps:

> Note that the safest option is to [fully disable the remote module](#15-disable-the-remote-module). If you choose to filter access rather than completely disable the module, you must be very careful to ensure that no escalation of privilege is possible through the modules you allow past the filter.

ppontes: Agreed, using now your suggestion.

miniak:
@zcbenz, @deepak1556 can you also please have a look?

PR #16468, opened 2019-01-20
feat: make ipc-message and ipc-message-sync events public

miniak:
#### Description of Change
- Allows handling IPC messages for particular `webContents`.
- Both `ipc-message` and `ipc-message-sync` on `webContents` are refactored to make the arguments easier to handle.

/cc @MarshallOfSound, @nornagon

#### Checklist
- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes
Notes: Added `ipc-message` and `ipc-message-sync` events to `webContents`.

PR #17489, opened 2019-03-20
feat: add a getScopedToSender API on ipcMain

MarshallOfSound:
#### Description of Change

This solves the usecase where folks for either security or use-case reasons only want to listen for message X coming from webContents Y.

Instead of checking the `sender` or `sender.id` each time and hoping you don't miss a listener or writing your own abstraction around IPC (let's be real a lot of us have done that before) this provides a low-overhead first-party API for solving this problem.

This use-case has actually come up in internal core code as well so we can benefit from this API internally.  E.g. https://github.com/electron/electron/pull/17416

I'll add tests and such if we're happy with this approach.

#### Release Notes

Notes: Added a new `ipcMain.getScopedToSender` API to create `ipcMain` instances that only emit IPC messages from a specific `webContents`

### DIFF
--- a/lib/browser/ipc-main-creator.ts
+++ b/lib/browser/ipc-main-creator.ts
@@ -8,7 +8,7 @@ enum IpcMainHeirachy {
 }
 
 export const createIndependentIpcMain = () => {
-  const senderScoped = new Map<number, IpcMain>()
+  const senderScoped = new WeakMap<Electron.WebContents, IpcMain>()
 
   class IpcMain extends EventEmitter implements Electron.IpcMain {
     constructor (private heirachy: IpcMainHeirachy) {
@@ -18,12 +18,11 @@ export const createIndependentIpcMain = () => {
     }
 
     getScopedToSender (webContents: Electron.WebContents) {
-      const senderId = webContents.id
-      let scoped = senderScoped.get(senderId)
+      let scoped = senderScoped.get(webContents)
       if (!scoped) {
         // Store based on senderId here and don't persist the webContents to avoid it leaking
         scoped = new IpcMain(IpcMainHeirachy.SCOPED)
-        senderScoped.set(senderId, scoped)
+        senderScoped.set(webContents, scoped)
       }
       return scoped
     }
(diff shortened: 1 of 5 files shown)


zcbenz:
The code looks good to me, but I don't have experiences with securing IPC messages so I'm not sure whether the API design is reasonable.

I think we should have actual users commenting whether they would use this API.

MarshallOfSound:
@nornagon The high level use-case is the opposite end of targeted sending of messages.

Currently we can say: Send this message to exactly this webContents
But there is no easy way to say: Only notify be about messages from this webContents

Example use cases for caring about that are per-webcontents features or security scoping of IPC messages (you only expect a particular webContents to respond to a certain message).

Either you get messages from all webConents with `ipcMain.on('X')` or you have to manually filter IPC messages with `webContents.on('ipc-message')`.

The intention of this API is to provide a minimal overhead way to subscribe to a given message from a given webContents.

miniak:
@MarshallOfSound any progress on this?

lib/browser/ipc-main-creator.ts
```diff
@@ -0,0 +1,47 @@
...
+        senderScoped.set(webContents, scoped)
+      }
+      return scoped
+    }
+
+    emit (channel: string, event: Electron.IpcMainEvent, ...args: any[]): boolean {
+      let handled = false
+      // If this is the root instance of IpcMain and we have an event with a valid sender
+      // we should try emit the event on the scoped IpcMain instance
+      if (this.heirachy === IpcMainHeirachy.ROOT && event && event.sender && event.sender.id) {
```
miniak: no need to check `event.sender.id` anymore

codebytere:
Is this ready to be merged @MarshallOfSound?

miniak:
@MarshallOfSound can you please rebase and resolve the conflicts?

miniak:
@nornagon I was proposing `webContents.ipc`

miniak:
@nornagon #16420

zcbenz:
According to the discussions at https://github.com/electron/electron/pull/16420 and https://github.com/electron/electron/pull/16468, it seems that this PR is no longer needed since `webContents.ipc` was proposed to replace this and then `webContents.ipc` was replaced by the `ipc-message` event.

So I'm closing this PR. Please reopen this if anyone still wants to continue this direction.

Issue #16168, opened 2018-12-21
require screen module before app ready causes error


timfish:
Affects all supported Electron versions and platforms and has been reported and closed before [here](https://github.com/electron/electron/issues/5897) in 2016.

**Expected Behavior**
This should not have any side effects:
```typescript
const { screen } = require('electron');
```
**Actual behavior**
However it gives an error:
`Error: Cannot require "screen" module before app is ready`

As well as the undesired side effects, the error is pretty poor as we're not even requiring the screen module, we're simply accessing the property.

This also means the docs ask developers to do unnecessary complicated things like:
```
const electron = require('electron')
const { app, BrowserWindow } = require('electron')

app.on('ready', () => {
  let displays = electron.screen.getAllDisplays()
```

**Additional Information**
The session API cannot be used before app ready, but can be required without side effects:
```
// This doesn't result in an error
const { session } = require('electron');
```

The APIs that are unavailable before app ready should at least behave consistently.

nornagon:
Seems reasonable to me. Looks like the right approach from the code side would be to lazily create an instance of `Screen` when it's first needed, rather than when the property is accessed. PR welcome!

sandeep1995:
Using `electron-window-state` package throws the same error as above

PR #17496, opened 2019-03-21
refactor: allow requiring modules with no side effects

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/16168.

This creates and returns a proxy object when modules that can't be used before `app` `ready`  are required such an instance of is no longer created at require-time and only when a method is called. 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue whereby requiring some modules before the app ready event had side effects.


lib/browser/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export function createLazyInstance <T, B extends boolean> (
+  creator: (() => T),
+  holder: new () => T,
+  isEventEmitter: B
+): Module<T, B> {
+  let lazyModule: T
+  const module: any = {}
+  for (const method in holder.prototype as (keyof T)[]) {
+    module[method] = (...args: any) => {
```
MarshallOfSound: This is making the assumption that all keys on the prototype are methods, I believe this is true currently, but to avoid a foot gun here can we either implement by checking `Object.getOwnPropertyDescriptor` and ensuring that all keys on the prototype are indeed methods and throw an error at runtime if they're not.

That way if people try to use this helper in the future with a property they'll have an error to fix instead of wondering where on earth their property is 👍

codebytere: should be set!

lib/browser/api/power-monitor.ts
```diff
@@ -1,53 +1,47 @@
...
+  if (typeof callback !== 'function') throw new Error('Must pass callback as a function argument')
 
   const idleState = this.getSystemIdleState(threshold)
-
   process.nextTick(() => callback(idleState))
 }
 
 // TODO(nitsakh): Remove in 7.0
-powerMonitor.querySystemIdleTime = function (callback) {
+powerMonitor.querySystemIdleTime = function (callback: Function) {
```
MarshallOfSound:
```suggestion
powerMonitor.querySystemIdleTime = function (callback: (idleTime: number) => void) {
```

lib/browser/api/power-monitor.ts
```diff
@@ -1,53 +1,47 @@
...
+  powerMonitor.on('removeListener', (event: string) => {
     if (event === 'shutdown' && powerMonitor.listenerCount('shutdown') === 0) {
       powerMonitor.unblockShutdown()
     }
   })
 }
 
 // TODO(nitsakh): Remove in 7.0
-powerMonitor.querySystemIdleState = function (threshold, callback) {
+powerMonitor.querySystemIdleState = function (threshold: number, callback: Function) {
```
MarshallOfSound:
```suggestion
powerMonitor.querySystemIdleState = function (threshold: number, callback: (idleTime: number) => void) {
```

lib/browser/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+  const module: any = {}
+  for (const method in (holder as any).prototype) {
+    module[method] = (...args: any) => {
+      // create new instance of module at runtime if none exists
+      if (!lazyModule) {
+        lazyModule = creator()
+        if (isEventEmitter) EventEmitter.call(lazyModule as any)
+      }
+      // ensure that all keys on the prototype are methods
+      if (Object.getOwnPropertyDescriptor(lazyModule, method)) {
```
MarshallOfSound: I think this was a mis-step on my part, we should just be checking `typeof lazyModule[method]` like we had to in the net log fix

lib/browser/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+* app 'ready' event by returning a proxy object to mitigate side effects
+* on 'require'
+*
+* @param {Function} creator - Function that creates a new module instance
+* @param {Object} holder - the object holding the module prototype
+* @param {Boolean} isEventEmitter - whether or not the module is an EventEmitter
+* @returns {Object} - a proxy object for the
+*/
+
+export function createLazyInstance (
```
miniak: why not use an ES6 Proxy?

codebytere: probably not unfortunately, the intricacies of doing that correctly would introduce far more space for potential issues

zcbenz: The `remote` module does not work very well with ES6 Proxy objects.

PR #17515, opened 2019-03-22
test: un-disable 60 chromium-spec tests

nornagon:
the skip() was unexpectedly applying to all following tests

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/spec/chromium-spec.js
+++ b/spec/chromium-spec.js
@@ -275,7 +275,7 @@ describe('chromium feature', () => {
   })
 
   describe('navigator.geolocation', () => {
-    before(function () {
+    beforeEach(function () {
       if (!features.isFakeLocationProviderEnabled()) {
         return this.skip()
       }


spec/chromium-spec.js
```diff
@@ -275,7 +275,7 @@ describe('chromium feature', () => {
   })
 
   describe('navigator.geolocation', () => {
-    before(function () {
+    beforeEach(function () {
```
deepak1556: why would this skip all tests ? wouldn't `before` be only applied to its `describe` block ?

nornagon: That's a great question, but without this change when I run `npm test -- --ci -g 'chromium feature'` i see 21 tests run. With this change, I see 82 tests run.

nornagon: ... hm, i'm not seeing the same difference when running on CI. https://circleci.com/gh/electron/electron/162245 and https://circleci.com/gh/electron/electron/161864 show the same number of tests. So perhaps this is a weird interaction with `--grep` or something...?

nornagon:
hrm, not sure what was going on here. can't repro now.

Issue #8117, opened 2016-12-01
Disparity between net module headers and Node IncommingMessage headers


niik:
As I understand things the net module is supposed to follow the Node API as closely as possible, from [net.md](https://github.com/electron/electron/blob/master/docs/api/net.md)

> The net module API has been specifically designed to mimic, as closely as possible, the familiar Node.js API. The API components including classes, methods, properties and event names are similar to those commonly used in Node.js.

The net module version of `IncommingMessage`, however, [returns a different structure](https://github.com/electron/electron/blob/master/lib/browser/api/net.js#L42) for its header collection than its Node counterpart. Where Node returns a simple object like `{ "Content-Type": "foo", ... }` the net module returns an object where each value is an array containing any duplicate headers: `{ "Content-Type": [ "foo" ], ...}`

<img width="356" alt="screen shot 2016-12-01 at 14 55 49" src="https://cloud.githubusercontent.com/assets/634063/20796414/490139ba-b7d6-11e6-920f-d24f8e7cc036.png">

So the net module returns duplicate headers as an array whereas Node will concat duplicate headers into a single string value (comma separated list) if the header is defined as a list in the HTTP specification.

From [the Node docs](https://nodejs.org/api/http.html#http_message_headers)

> Duplicates in raw headers are handled in the following ways, depending on the header name:
>
> Duplicates of age, authorization, content-length, content-type, etag, expires, from, host, if-modified-since, if-unmodified-since, last-modified, location, max-forwards, proxy-authorization, referer, retry-after, or user-agent are discarded.
>
> set-cookie is always an array. Duplicates are added to the array.
>
> For all other headers, the values are joined together with ', '.

### Expected behavior

Duplicate HTTP response headers should be joined together if allowed by the HTTP specification or else discarded. 

### Actual behavior

Headers are returned in an array regardless of duplicates.

### Proposed solution

Node has already solved this problem once, see [_http_incoming.js#L134-L180](https://github.com/nodejs/node/blob/8264a2225628cbc32116f83a079f6024786988ab/lib/_http_incoming.js#L134-L180). Although the method is exposed on the `IncommingMessage` prototype it's an undocumented method which is subject to change so using it directly is probably #dangerzone. Since the licenses of electron and node seem compatible I propose borrowing that code and adapting it for use in electron.

cc @kevinsawicki @joshaber @shiftkey

kevinsawicki:
> Proposed solution

Sounds good to me, thanks for bringing this up 👍

[...]

PR #17517, opened 2019-03-22
fix: match net module headers & http.IncomingMessage headers

codebytere:
#### Description of Change

BREAKING CHANGE

Fixes https://github.com/electron/electron/issues/8117.

Filters net module incoming headers to more accurately match those returned in [`http.IncomingMessage`](https://nodejs.org/api/http.html#http_message_headers).

cc @nornagon @MarshallOfSound 

More specifically, after this change: 
* Duplicates of `age`, `authorization`, `content-length`, `content-type`, `etag`, `expires`, `from`, `host`, `if-modified-since`, `if-unmodified-since`, `last-modified`, `location`, `max-forwards`, `proxy-authorization`, `referer`, `retry-after`, or `user-agent` are discarded.
* `set-cookie` is always an array. Duplicates are added to the array.
* For duplicate cookie headers, the values are joined together with '; '.
* For all other headers, the values are joined together with ', '.

Tested with:
```js
const { app } = require('electron')

app.on('ready', () => {
  const { net } = require('electron')
  const request = net.request('https://github.com')
  request.on('response', response => {
    console.log('HEADERS: ', response.headers)
  })
  request.end()
})
```
to ensure that headers conform to the new rules appropriately

#### Checklist

- [x] `npm test` passes
- [x] tests are added or changed

#### Release Notes

Notes: Fixed disparity between `net` module headers and Node.js'  `http.IncomingMessage` headers

### DIFF
--- a/lib/browser/api/net.js
+++ b/lib/browser/api/net.js
@@ -75,7 +75,7 @@ class IncomingMessage extends Readable {
           // see https://nodejs.org/api/http.html#http_message_headers
           filteredHeaders[header] = rawHeaders[header]
         } else {
-          // for all other headers, the values are joined together with ', '
+          // for non-cookie headers, the values are joined together with ', '
           filteredHeaders[header] = rawHeaders[header].join(', ')
         }
       }
--- a/lib/browser/api/net.js
+++ b/lib/browser/api/net.js
@@ -41,7 +41,46 @@ class IncomingMessage extends Readable {
   }
 
   get headers () {
-    return this.urlRequest.rawResponseHeaders
+    // list of headers that Node.js discards duplicates for
+    // see https://nodejs.org/api/http.html#http_message_headers
+    const discardableDuplicateHeaders = [
+      'content-type',
+      'content-length',
+      'user-agent',
+      'referer',
+      'host',
+      'authorization',
+      'proxy-authorization',
+      'if-modified-since',
+      'if-unmodified-since',
+      'from',
+      'location',
+      'max-forwards',
+      'retry-after',
+      'etag',
+      'last-modified',
+      'server',
+      'age',
+      'expires'
+    ]
+
+    const filteredHeaders = {}
+    const rawHeaders = this.urlRequest.rawResponseHeaders
+    Object.keys(rawHeaders).forEach(header => {
+      if (header in filteredHeaders && discardableDuplicateHeaders.includes(header)) {
+        // do nothing with discardable duplicate headers
+      } else {
+        if (header === 'set-cookie') {
+          // keep set-cookie as an array per Node.js rules
+          // see https://nodejs.org/api/http.html#http_message_headers
+          filteredHeaders[header] = rawHeaders[header]
+        } else {
+          // a header is present iff it has a value, and there will only be 1 value
+          filteredHeaders[header] = rawHeaders[header][0]
+        }
+      }
+    })
+    return filteredHeaders
   }
 
   get httpVersion () {
--- a/lib/browser/api/net.js
+++ b/lib/browser/api/net.js
@@ -16,6 +16,29 @@ Object.setPrototypeOf(URLRequest.prototype, EventEmitter.prototype)
 
 const kSupportedProtocols = new Set(['http:', 'https:'])
 
+// set of headers that Node.js discards duplicates for
+// see https://nodejs.org/api/http.html#http_message_headers
+const discardableDuplicateHeaders = new Set([
+  'content-type',
+  'content-length',
+  'user-agent',
+  'referer',
+  'host',
+  'authorization',
+  'proxy-authorization',
+  'if-modified-since',
+  'if-unmodified-since',
+  'from',
+  'location',
+  'max-forwards',
+  'retry-after',
+  'etag',
+  'last-modified',
+  'server',
+  'age',
+  'expires'
+])
+
 class IncomingMessage extends Readable {
   constructor (urlRequest) {
     super()
@@ -41,33 +64,10 @@ class IncomingMessage extends Readable {
   }
 
   get headers () {
-    // list of headers that Node.js discards duplicates for
-    // see https://nodejs.org/api/http.html#http_message_headers
-    const discardableDuplicateHeaders = [
-      'content-type',
-      'content-length',
-      'user-agent',
-      'referer',
-      'host',
-      'authorization',
-      'proxy-authorization',
-      'if-modified-since',
-      'if-unmodified-since',
-      'from',
-      'location',
-      'max-forwards',
-      'retry-after',
-      'etag',
-      'last-modified',
-      'server',
-      'age',
-      'expires'
-    ]
-
     const filteredHeaders = {}
     const rawHeaders = this.urlRequest.rawResponseHeaders
     Object.keys(rawHeaders).forEach(header => {
-      if (header in filteredHeaders && discardableDuplicateHeaders.includes(header)) {
+      if (header in filteredHeaders && discardableDuplicateHeaders.has(header)) {
         // do nothing with discardable duplicate headers
       } else {
         if (header === 'set-cookie') {
@@ -75,8 +75,8 @@ class IncomingMessage extends Readable {
           // see https://nodejs.org/api/http.html#http_message_headers
           filteredHeaders[header] = rawHeaders[header]
         } else {
-          // a header is present iff it has a value, and there will only be 1 value
-          filteredHeaders[header] = rawHeaders[header][0]
+          // for all other headers, the values are joined together with ', '
+          filteredHeaders[header] = rawHeaders[header].join(', ')
         }
       }
     })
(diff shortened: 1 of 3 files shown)


spec/api-net-spec.js
```diff
@@ -1388,8 +1388,7 @@ describe('net module', () => {
         assert.strictEqual(statusMessage, 'OK')
         const headers = response.headers
         assert(typeof headers === 'object')
-        assert.deepStrictEqual(headers[customHeaderName.toLowerCase()],
-          [customHeaderValue])
+        assert.deepStrictEqual(headers[customHeaderName.toLowerCase()], customHeaderValue)
```
nornagon: Does this mean this is a breaking change to a publicly-accessible API?

spec/api-net-spec.js
```diff
@@ -1400,11 +1399,8 @@ describe('net module', () => {
...
-        response.on('data', (chunk) => {
-        })
-        response.on('end', () => {
-          done()
-        })
+        response.on('data', (chunk) => {})
+        response.on('end', () => { done() })
         response.resume()
       })
       urlRequest.end()
```
nornagon: Would like a specific test for the new deduping behaviour

lib/browser/api/net.js
```diff
@@ -41,7 +41,46 @@ class IncomingMessage extends Readable {
...
+      'last-modified',
+      'server',
+      'age',
+      'expires'
+    ]
+
+    const filteredHeaders = {}
+    const rawHeaders = this.urlRequest.rawResponseHeaders
+    Object.keys(rawHeaders).forEach(header => {
+      if (header in filteredHeaders && discardableDuplicateHeaders.includes(header)) {
```
nornagon: do we need to `toLowerCase` here?

lib/browser/api/net.js
```diff
@@ -41,7 +41,46 @@ class IncomingMessage extends Readable {
   }
 
   get headers () {
-    return this.urlRequest.rawResponseHeaders
+    // list of headers that Node.js discards duplicates for
+    // see https://nodejs.org/api/http.html#http_message_headers
+    const discardableDuplicateHeaders = [
```
nornagon: Break this out as a top-level `Set` (for faster `includes` check)

lib/browser/api/net.js
```diff
@@ -41,7 +41,46 @@ class IncomingMessage extends Readable {
...
+    Object.keys(rawHeaders).forEach(header => {
+      if (header in filteredHeaders && discardableDuplicateHeaders.includes(header)) {
+        // do nothing with discardable duplicate headers
+      } else {
+        if (header === 'set-cookie') {
+          // keep set-cookie as an array per Node.js rules
+          // see https://nodejs.org/api/http.html#http_message_headers
+          filteredHeaders[header] = rawHeaders[header]
+        } else {
+          // a header is present iff it has a value, and there will only be 1 value
```
nornagon: This doesn't seem right, the node.js docs say:
> For all other headers, the values are joined together with ', '.

spec/api-net-spec.js
codebytere: yes, which lends itself to a next question: should this be classified as a fix or a refactor? i can see the argument for both

lib/browser/api/net.js
codebytere: yes, the headers aren't always consistent so it's simplest to lowercase them all

lib/browser/api/net.js
codebytere: (resolved offline)

spec/api-net-spec.js
nornagon: we should also put `BREAKING CHANGE` in the PR description then.

it's not a refactor, because it changes the API.

spec/api-net-spec.js
codebytere: done!

lib/browser/api/net.js
```diff
@@ -62,12 +62,12 @@ class IncomingMessage extends Readable {
       'server',
       'age',
       'expires'
-    ]
+    ])
```
nornagon: Extract this `Set` to the global level so that it's not re-created every time this function is called

spec/api-net-spec.js
```diff
@@ -1375,36 +1376,90 @@ describe('net module', () => {
...
+
+      const urlRequest = net.request({
+        method: 'GET',
+        url: `${server.url}${requestUrl}`
+      })
+
+      urlRequest.on('response', (response) => {
+        const statusCode = response.statusCode
+        assert(typeof statusCode === 'number')
+        assert.strictEqual(statusCode, 200)
```
nornagon:
```suggestion
        expect(statusCode).to.be(200)
```

No need to check that `typeof x === 'number'` if you're going to check that it's `strictEqual` to a number.

spec/api-net-spec.js
```diff
@@ -1375,36 +1376,90 @@ describe('net module', () => {
...
+        const statusMessage = response.statusMessage
+        assert(typeof statusMessage === 'string')
+        assert.strictEqual(statusMessage, 'OK')
+
+        const headers = response.headers
+        assert(typeof headers === 'object')
+
+        assert.strictEqual(includedHeader in headers, true)
+        assert.strictEqual(discardableHeader in headers, false)
+
```
nornagon: Can we also test the "Joined by `", "`" behaviour?

spec/api-net-spec.js
codebytere: i don't think we have expect in this file so i was matching previous style but i can add it if that seems like a better way forward :)

nornagon: oh weird, yeah we shouldn't be adding new code in the old style IMO.

spec/api-net-spec.js
```diff
@@ -1375,31 +1377,103 @@ describe('net module', () => {
...
+        response.on('end', () => { done() })
+        response.resume()
+      })
+      urlRequest.end()
+    })
+
+    it('should join repeated non-discardable value with ,', (done) => {
+      const urlRequest = net.request({
+        method: 'GET',
+        url: 'https://github.com'
```
nornagon: Never contact a remote host in a test; our tests shouldn't fail if GitHub goes down (or more likely, if any of the several internet relays between CircleCI and GitHub are slightly flaky).

spec/api-net-spec.js
```diff
@@ -1375,31 +1377,103 @@ describe('net module', () => {
...
+        expect(response.statusCode).to.equal(200)
+        expect(response.statusMessage).to.equal('OK')
+
+        const headers = response.headers
+        expect(headers).to.be.an('object')
+
+        expect(headers).to.have.a.property('referrer-policy')
+        const referrerPolicyValues = headers['referrer-policy']
+        const values = referrerPolicyValues.split(',')
+        expect(values).to.be.an('array').with.lengthOf.at.least(1)
```
nornagon: This doesn't actually check that there exists any `,` in the input value, since `"foo".split(',')` is an array with length of at least one.

When you switch to using the local server instead of github.com, it should be easier to know what to expect here, and we should expect the exact string (not the result of `split`ting it).

spec/api-net-spec.js
```diff
@@ -1375,36 +1377,116 @@ describe('net module', () => {
...
+
+      urlRequest.on('response', response => {
+        expect(response.statusCode).to.equal(200)
+        expect(response.statusMessage).to.equal('OK')
+
+        const headers = response.headers
+        expect(headers).to.be.an('object')
+
+        expect(headers).to.have.property(includedHeader)
+        expect(headers).to.not.have.property(discardableHeader)
```
nornagon:
```suggestion
        expect(headers).to.not.have.property(discardableHeader)
        expect(headers[includedHeader]).to.equal(includedHeaderValue)
```

codebytere:
Failure is:
```
not ok 139 BrowserWindow module "useContentSize" option make window created with content size when used
  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
  
  124 !== 400
  
      at Context.<anonymous> (C:\projects\src\electron\spec\api-browser-window-spec.js:1229:14)
      at callFn (C:\projects\src\electron\spec\node_modules\mocha\lib\runnable.js:372:21)
      at Test.Runnable.run (C:\projects\src\electron\spec\node_modules\mocha\lib\runnable.js:364:7)
      at Runner.runTest (C:\projects\src\electron\spec\node_modules\mocha\lib\runner.js:455:10)
      at C:\projects\src\electron\spec\node_modules\mocha\lib\runner.js:573:12
      at next (C:\projects\src\electron\spec\node_modules\mocha\lib\runner.js:369:14)
      at C:\projects\src\electron\spec\node_modules\mocha\lib\runner.js:379:7
      at next (C:\projects\src\electron\spec\node_modules\mocha\lib\runner.js:303:14)
      at Immediate._onImmediate (C:\projects\src\electron\spec\node_modules\mocha\lib\runner.js:347:5)
      at processImmediate (timers.js:632:19)
```

which is assuredly not related, so merging regardless.

PR #17518, opened 2019-03-22
fix: backport ICU changes for Japanese Era placeholder name

jefgen:
#### Description of Change

This change backports changes from the upstream ICU 62.1 maintenance branch into the Chromium copy of ICU, in order to enable the "placeholder era" for the upcoming new Japanese Era.

Upstream change: https://github.com/unicode-org/icu/pull/514
Upstream ticket: https://unicode-org.atlassian.net/browse/ICU-20482

With this change you can set/define an Environment Variable named `ICU_ENABLE_TENTATIVE_ERA` to `true` which will enable the placeholder era name.

Notes on the changes:
- This change required re-generating the checked-in Chromium ICU data files (ex: `icudtl.dat`) in order to update them with the new data for the placeholder era name. (This is why the .patch file is so large and contains binary diffs.)
- The Chromium ICU folder under `third_party` is a separate git repo, so this required added a new folder named `icu` to the `electron\patches\common` folder (since there are currently no existing patches for ICU)
- Also included in this change are some minor changes to the scripts used by Chromium for building the data files so that they actually work on Ubuntu.

CC: @alexeykuzmin (also FYI for @aandrejs).

Here's a simple example showing the results of the change:

Before:
before

After:
after
(Note the use of "ＱＱ" instead of the "平成" for the date format for ja-JP).

Sample code used for testing:
```html
<p>Outputs:</p>
<form>
<textarea id="jaEraOutputs" rows="9" cols="60"></textarea>
</form>

<script type="text/javascript">
    // Heisi Era will end on April 30, 2019 so pick a date before.
    var dateHeisi = new Date(Date.UTC(2019, 01, 15, 3, 0, 0));
    // New Era starts on May 1, 2019.
    var dateNew = new Date(Date.UTC(2019, 05, 02, 3, 0, 0));
    
    var out = document.getElementById("jaEraOutputs");
    out.value = "";
    out.value += "Heisi: \n";
    out.value += "\t ja-JP:  " + dateHeisi.toLocaleString('ja-JP-u-ca-japanese', {era: 'long'}) + "\n";
    out.value += "\t en-US:  " + dateHeisi.toLocaleString('en-US-u-ca-japanese', {era: 'long'}) + "\n";
    out.value += "\n";
    out.value += "New Era: \n";
    out.value += "\t ja-JP:  " + dateNew.toLocaleString('ja-JP-u-ca-japanese', {era: 'long'}) + "\n";
    out.value += "\t en-US:  " + dateNew.toLocaleString('en-US-u-ca-japanese', {era: 'long'}) + "\n";
</script>
```

#### Checklist

- [x] `npm test` passes
 - (I can't seem to get the `npm test` command working on Windows, even without my changes.)

#### Release Notes

Notes: Added support for testing the new Japanese Era with a placeholder era name. Set an environment variable named `ICU_ENABLE_TENTATIVE_ERA` to true to enable the placeholder era.


deepak1556:
Just to clarify, where the ICU data files generated following the instructions in https://cs.chromium.org/chromium/src/third_party/icu/README.chromium ?

jefgen:
> Just to clarify, where the ICU data files generated following the instructions in https://cs.chromium.org/chromium/src/third_party/icu/README.chromium ?

Yes of course. :)
I regenerated *all* of the checked-in data files (for Android, iOS, ChromeCast, etc.) even though I think Electron only uses the "common" one.

deepak1556:
Thanks! Just to get some understanding, what would steps C1-4 be for in the above document

> C. Chromium-specific data build files and converters

Is that something we should care about when updating icu with patches ?

jefgen:
> Thanks! Just to get some understanding, what would steps C1-4 be for in the above document
> 
> > C. Chromium-specific data build files and converters
> 
> Is that something we should care about when updating icu with patches ?

I'm not sure I understand the question? Perhaps you could rephrase it?.

However, regarding ICU and Chromium changes:
Yes, Chromium makes a number of "Chromium-specific" changes to both the ICU code and the ICU data files.

If you are just *patching* things in ICU (by pulling things from upstream ICU) then you generally/hopefully will not have any issues with conflicting Chromium-specific changes.

However, if you try to independent bump the ICU version in Electron, then yes -- you will also need to carry forward all of the Chromium-specific changes, which might need to be changed between ICU versions.

(Hopefully this helps explain things a bit?)

jefgen:
Oh other other important point:
This is just the *placeholder* era name. Once the real/official era name is released, I'll need to regenerate the data files again with the real name. (We can likely replace this patch with a newer patch, so there would only be one patch file instead of two.)

deepak1556:
> I'm not sure I understand the question? Perhaps you could rephrase it?.

Sorry, I was referring to the steps here https://cs.chromium.org/chromium/src/third_party/icu/README.chromium?l=116-166

> If you are just patching things in ICU (by pulling things from upstream ICU) then you generally/hopefully will not have any issues with conflicting Chromium-specific changes.

You have clarified it, thanks!

PR #17546, opened 2019-03-26
fix: use file path as key for safeDialogs on filesystem

### LINKED ISSUE #17543: safeDialogs option doesn't work
### Preflight Checklist

### Issue Details

* **Electron Version:** 5.0.0-beta.6
* **Operating System:** macOS 10.13.6
* **Last Known Working Electron version:**: unknown

### To Reproduce

```
git clone https://github.com/PalmerAL/electron-quick-start.git
cd electron-quick-start
git checkout safe-dialogs
electron .
```

When running the example app, the "prevent this app from creating additional dialogs" option will appear after the second popup, but checking it doesn't actually do anything. Instead, dialogs will continue to appear forever until you force-quit the app.

MarshallOfSound:
#### Description of Change
`file://` protocol URLs weren't getting valid keys for the count store.

Notes: The `safeDialogs` webPreferences option now actually works on `file://` URLs

### DIFF
--- a/atom/browser/atom_javascript_dialog_manager.cc
+++ b/atom/browser/atom_javascript_dialog_manager.cc
@@ -41,7 +41,16 @@ void AtomJavaScriptDialogManager::RunJavaScriptDialog(
     DialogClosedCallback callback,
     bool* did_suppress_message) {
   auto origin_url = rfh->GetLastCommittedURL();
-  const std::string& origin = origin_url.GetOrigin().spec();
+
+  std::string origin;
+  // For file:// URLs we do the alert filtering by the
+  // file path currently loaded
+  if (origin_url.SchemeIsFile()) {
+    origin = origin_url.path();
+  } else {
+    origin = origin_url.GetOrigin().spec();
+  }
+
   if (origin_counts_[origin] == kUserWantsNoMoreDialogs) {
     return std::move(callback).Run(false, base::string16());
   }


PalmerAL:
@MarshallOfSound The option doesn't appear to work correctly even if modify the webcontents to use a `http://` URL, so I'm guessing there may be more than one bug preventing this from working.

MarshallOfSound:
@PalmerAL I just checked `5.0.0-beta.6` without this change and it works fine for `http://` URLs.  With this change it also works for `file://` URLs

PalmerAL:
@MarshallOfSound Interesting, I'm testing with the same version of Electron, and it definitely doesn't work for me.

Code:

```
  mainWindow = new BrowserWindow({
    width: 800,
    height: 600,
    webPreferences: {
      nodeIntegration: false,
      safeDialogs: true
    }
  })

  // and load the index.html of the app.
  mainWindow.loadURL('https://electronjs.org/')

  // Open the DevTools.
  mainWindow.webContents.openDevTools()
```

Screen recording: 
[electron-safe-dialogs.zip](https://github.com/electron/electron/files/3006401/electron-safe-dialogs.zip)

codebytere:
@MarshallOfSound should we wait for further investigation of the above before merging?

MarshallOfSound:
@codebytere I want to merge this as the issue is definitely fixed on master on both `file://` and `http://`.  I have removed the `"Fixes #X"` though to leave that issue open for more investigation into 5.x.

PR #17547, opened 2019-03-26
fix: set default_id and cancel_id correctly on confirm dialogs

### LINKED ISSUE #17502: Pressing escape confirms `window.confirm` dialog
* **Electron Version:**
  * v4.0.1
* **Operating System:**
  * macOS 10.13.6

### Expected Behavior
When you press the escape key on a confirm dialog, it should be the same as pressing cancel.

### Actual Behavior
When you press escape, it presses "OK". This has the potential to be pretty dangerous, as it may continue with a destructive action the user had not intended.

### To Reproduce

Open a window.confirm, then press escape.

### Additional Information

This seems to be due to [this line](https://github.com/electron/electron/blob/master/atom/browser/atom_javascript_dialog_manager.cc#L79), where `cancelId`([5th argument](https://github.com/electron/electron/blob/master/atom/browser/ui/message_box_mac.mm#L97)) is hardcoded to `0`.

For a confirm dialog, we [`push_back`](https://github.com/electron/electron/blob/master/atom/browser/atom_javascript_dialog_manager.cc#L57) the `Cancel` button, so this should be `1`.

MarshallOfSound:
#### Description of Change
Sets `cancel_id` to the Cancel button and `default_id` to the OK button in CONFIRM type dialogs. (`window.confirm`)

Fixes #17502

#### Release Notes

Notes: Fixed the handling of the escape key in dialogs created by `window.confirm`. The "Cancel" button is now triggered by the escape key, and the "OK" button is now triggered by the return key.

### DIFF
--- a/atom/browser/atom_javascript_dialog_manager.cc
+++ b/atom/browser/atom_javascript_dialog_manager.cc
@@ -52,9 +52,16 @@ void AtomJavaScriptDialogManager::RunJavaScriptDialog(
     return;
   }
 
+  // No default button
+  int default_id = -1;
+  int cancel_id = 0;
+
   std::vector<std::string> buttons = {"OK"};
   if (dialog_type == JavaScriptDialogType::JAVASCRIPT_DIALOG_TYPE_CONFIRM) {
     buttons.push_back("Cancel");
+    // First button is default, second button is cancel
+    default_id = 0;
+    cancel_id = 1;
   }
 
   origin_counts_[origin]++;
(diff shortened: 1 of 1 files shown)


atom/browser/atom_javascript_dialog_manager.cc
```diff
@@ -52,9 +52,14 @@ void AtomJavaScriptDialogManager::RunJavaScriptDialog(
     return;
   }
 
+  int default_id = -1;
```
codebytere: can you include a comment here for future file visitors about what these id values do?

jsteiner:
@MarshallOfSound would it be possible to backport this to 4.x? Seems worth it given the potential consequences of wrongly confirming something.

MarshallOfSound:
@jsteiner I don't think 4.x has this bug and can't verify rn. Can you repro for 4?

jsteiner:
@MarshallOfSound yeah, we're seeing this issue on 4.0.1. It does seem to have the same hardcoded line of code on the `4-1-x` branch here: https://github.com/electron/electron/blob/4-1-x/atom/browser/atom_javascript_dialog_manager.cc#L79

PR #13301, opened 2018-06-19
revert: frame subscriber deprecation and re-enable tests

brenca:
##### Checklist

- [ ] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [ ] commit messages or PR title follow semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

PR #17548, opened 2019-03-26
fix: the callback of beginFrameSubscription should pass NativeImage instead of Buffer

zcbenz:
#### Description of Change

As discussed in https://github.com/electron/electron/pull/13301#issuecomment-474167710, the callback of `beginFrameSubscription` should pass `NativeImage` instead of `Buffer`.

This PR makes sure `NativeImage` is passed, and fixes the original crash caused by the returned image not copying the bitmap pixels.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix the callback of `beginFrameSubscription` passing `Buffer` instead of `NativeImage`.


codebytere:
@zcbenz could you please backport this to `5-0-x`?

Issue #17551, opened 2019-03-26
crash in atom_api_desktop_capturer.cc


epicylq:
electron code head commit:  2fb9085e5b9b28b7adf0ab7de720c87461b6689d
os: win10 64bit
visual studio 2019 preview

after build electron , I run 
```
electron.exe electron/spec
```
and the program crash, i use vs2019 debug it, see the picture, 
![捕获](https://user-images.githubusercontent.com/43543863/54986357-7331d580-4fed-11e9-9419-0775d533d3fe.PNG)

device_names is empty, and the code does not check it

PR #17557, opened 2019-03-26
fix: crash when failed to get devices in desktopCapturer

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/17551.

`GetDeviceNames` returns a boolean on success, which we previously weren't checking and just assuming would succeed. On failure, the devices list would be empty, which caused the crash experience by a user in the above issue.

This PR changes that `DxgiDuplicatorController` call and checks the boolean so that a descriptive error is now emitted instead of a crash on failure.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash when failed to get devices in desktopCapturer on Windows.

### DIFF
--- a/atom/browser/api/atom_api_desktop_capturer.cc
+++ b/atom/browser/api/atom_api_desktop_capturer.cc
@@ -165,8 +165,14 @@ void DesktopCapturer::UpdateSourcesList(DesktopMediaList* list) {
       std::vector<std::string> device_names;
       // Crucially, this list of device names will be in the same order as
       // |media_list_sources|.
-      webrtc::DxgiDuplicatorController::Instance()->GetDeviceNames(
-          &device_names);
+      bool success =
+          webrtc::DxgiDuplicatorController::Instance()->GetDeviceNames(
+              &device_names);
+
+      if (!success)
+        Emit("finished",
+             "Failed to get device names of all screens on the system.");
+
       int device_name_index = 0;
       for (auto& source : screen_sources) {
         const auto& device_name = device_names[device_name_index++];
--- a/atom/browser/api/atom_api_desktop_capturer.cc
+++ b/atom/browser/api/atom_api_desktop_capturer.cc
@@ -165,13 +165,11 @@ void DesktopCapturer::UpdateSourcesList(DesktopMediaList* list) {
       std::vector<std::string> device_names;
       // Crucially, this list of device names will be in the same order as
       // |media_list_sources|.
-      bool success =
-          webrtc::DxgiDuplicatorController::Instance()->GetDeviceNames(
-              &device_names);
-
-      if (!success)
-        Emit("finished",
-             "Failed to get device names of all screens on the system.");
+      if (!webrtc::DxgiDuplicatorController::Instance()->GetDeviceNames(
+              &device_names)) {
+        Emit("error", "Failed to get sources.");
+        return;
+      }
 
       int device_name_index = 0;
       for (auto& source : screen_sources) {
--- a/lib/browser/desktop-capturer.js
+++ b/lib/browser/desktop-capturer.js
@@ -26,7 +26,8 @@ ipcMainUtils.handle('ELECTRON_BROWSER_DESKTOP_CAPTURER_GET_SOURCES', (event, cap
         thumbnailSize,
         fetchWindowIcons
       },
-      resolve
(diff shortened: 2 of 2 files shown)


atom/browser/api/atom_api_desktop_capturer.cc
```diff
@@ -135,7 +135,10 @@ bool DesktopCapturer::ShouldScheduleNextRefresh(DesktopMediaList* list) {
   return false;
 }
 
-void DesktopCapturer::UpdateSourcesList(DesktopMediaList* list) {
+void DesktopCapturer::UpdateSourcesList(mate::Arguments* args) {
+  DesktopMediaList* list;
+  args->GetNext(&list);
```
MarshallOfSound: This isn't guaranteed to succeed, in order to offload the argument parsing to native mate we can do this instead.

```
void DesktopCapturer::UpdateSourcesList(DesktopMediaList* list, mate::Arguments* args) {
```

atom/browser/api/atom_api_desktop_capturer.cc
```diff
@@ -169,8 +172,14 @@ void DesktopCapturer::UpdateSourcesList(DesktopMediaList* list) {
...
       // Crucially, this list of device names will be in the same order as
       // |media_list_sources|.
-      webrtc::DxgiDuplicatorController::Instance()->GetDeviceNames(
-          &device_names);
+      bool success =
+          webrtc::DxgiDuplicatorController::Instance()->GetDeviceNames(
+              &device_names);
+
+      if (!success)
+        args->ThrowError(
```
MarshallOfSound: Should this actually be an error or can we somehow fail softly and indicate that the `display_id` property won't be available but still return the other information?

codebytere: i don't think so, because all subsequent info surfaced on windows hinges on having those names

atom/browser/api/atom_api_desktop_capturer.cc
```diff
@@ -169,8 +169,14 @@ void DesktopCapturer::UpdateSourcesList(DesktopMediaList* list) {
       std::vector<std::string> device_names;
       // Crucially, this list of device names will be in the same order as
       // |media_list_sources|.
-      webrtc::DxgiDuplicatorController::Instance()->GetDeviceNames(
-          &device_names);
+      bool success =
+          webrtc::DxgiDuplicatorController::Instance()->GetDeviceNames(
```
deepak1556: According to the api definition, it should return false only when DirectX capturing is not supported, in that case `using_directx_capturer_` should have never been true, weird.

atom/browser/api/atom_api_desktop_capturer.cc
```diff
@@ -169,8 +169,14 @@ void DesktopCapturer::UpdateSourcesList(DesktopMediaList* list) {
...
+              &device_names);
+
+      if (!success)
```
deepak1556: Shouldn't continue to iterate over screen captured sources for display id modifications once this has failed.

codebytere: fixed!

sjames1958gm:
I am seeing this error 3221225477 on Windows 10, I tried upgrading to Electron 5.x and still see it. I switch out "window" for "screen" and it works just fine.

CapOM:
Hi @sjames1958gm , can you clarify how you are calling desktopCapturer.getSources ?  Thx

atom/browser/api/atom_api_desktop_capturer.cc
```diff
@@ -165,8 +165,12 @@ void DesktopCapturer::UpdateSourcesList(DesktopMediaList* list) {
       std::vector<std::string> device_names;
       // Crucially, this list of device names will be in the same order as
       // |media_list_sources|.
-      webrtc::DxgiDuplicatorController::Instance()->GetDeviceNames(
-          &device_names);
+      if (!webrtc::DxgiDuplicatorController::Instance()->GetDeviceNames(
+              &device_names)) {
+        Emit("error", "Failed to get sources.");
```
CapOM: I do not think we want to emit an error here, we might still want to return the TYPE_WINDOW sources, even if user is just asking TYPE_SCREEN we should just return an empty list.

PR #17559, opened 2019-03-26
build: speed up windows download of external binaries

MarshallOfSound:
So turns out Windows so _so bad_ at printing stuff on stdout that this script literally makes my computer take 5 minutes to download a 300kb zip file.

This makes it finish in a few seconds 👍 

Notes: no-notes


nornagon:
lol

Issue #12508, opened 2018-04-02
Encourage Touch Bar items reusability OR add a warning in the docs


dimitarnestorov:
If Touch Bar items are used over 10 times a `MaxListenersExceededWarning` is being emitted:
```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 change listeners added. Use emitter.setMaxListeners() to increase limit
```

The problem: every time `BrowserWindow.setTouchBar` is called, every Touch Bar item gets an event listener for the `change` event. However this event is never detached from those items.

Example code for achieving this warning (`mainWindow` is a `BrowserWindow` instance)
```
const button = new TouchBarButton({
	label: 'Test',
	backgroundColor: '#32DB7E'
});
for(let i = 0; i < 11; i++){
	mainWindow.setTouchBar(new TouchBar([
		button
	]));
}
```

My ask is either detach the change listener of the items when the Touch Bar is removed/changed or add a warning somewhere in the docs for people to keep in mind to not reuse Touch Bar items.

Escape hatch (very bad practice): `TouchBarItem.removeAllListeners('change')`

PR #17560, opened 2019-03-26
fix: event leak when reusing TouchBarItems

dimitarnestorov:
#### Description of Change
When reusing a `TouchBarItem` in more than 10 instances of `TouchBar` the following warning will show up:
```
(node:14027) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 change listeners added. Use emitter.setMaxListeners() to increase limit
```

Here's a code example of this issue ([fiddle](https://gist.github.com/e86b3054fe72f9a47c645737f0ea874e)):
```javascript
const {TouchBar} = require('electron')
const {TouchBarSpacer} = TouchBar

const flexSpace = new TouchBarSpacer({ size: 'flexible' })
for(let i = 0; i < 11; i++){
  new TouchBar([
    flexSpace
  ]);
}
```

This warning will still appear if reusing a `TouchBarItem` over 11 different windows with different `TouchBar`s but in that case it won't be a leak.

Related issue: https://github.com/electron/electron/issues/12508 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed event leak when creating many TouchBar instances that reuse TouchBarItems.

### DIFF
--- a/lib/browser/api/touch-bar.js
+++ b/lib/browser/api/touch-bar.js
@@ -46,7 +46,6 @@ class TouchBar extends EventEmitter {
 
     const registerItem = (item) => {
       this.items[item.id] = item
-      item.on('change', this.changeListener)
       if (item.child instanceof TouchBar) {
         item.child.ordereredItems.forEach(registerItem)
       }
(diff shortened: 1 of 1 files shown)


dimitarnestorov:
Yes but what if I never attach those TouchBars to a window? That's exactly what I'm doing in my app, I create two TouchBars and swap them, but when I have to update a popover I have to recreate them, one of them leaks (if it was never attached), and in 10 recreations I get the warning.

MarshallOfSound:
> Yes but what if I never attach those TouchBars to a window?

I would argue that you shouldn't be attaching the `TouchBarItem` instances to a `TouchBar` if you aren't going to use them.  You can have an array of `TouchBarItem`'s laying around doing nothing, the second you put them in a `TouchBar` instance it expects to be used.

dimitarnestorov:
I refactored my code and worked a solution without bugs and without the warning. But at least update the docs to say that one shouldn't create multiple `TouchBar`s without using them.

zcbenz:
I'm closing this PR as this solution would introduce bugs as mentioned by MarshallOfSound:

> This means if you set and unset the same touchbar to a window multiple listeners will be added and this will cause bugs / multiple listeners firing when they shouldn't.

PR #17563, opened 2019-03-27
docs: added fee hint

kapilepatel:
#### Description of Change

Updated documentation to add fee hint.

Notes: none

### DIFF
--- a/docs/tutorial/code-signing.md
+++ b/docs/tutorial/code-signing.md
@@ -50,9 +50,9 @@ Before signing Windows builds, you must do the following:
 
 You can get a code signing certificate from a lot of resellers, popular ones include:
 
-* [digicert $$$](https://www.digicert.com/code-signing/microsoft-authenticode.htm)
-* [Comodo $](https://www.comodo.com/landing/ssl-certificate/authenticode-signature/)
-* [GoDaddy $$](https://au.godaddy.com/web-security/code-signing-certificate)
+* [digicert](https://www.digicert.com/code-signing/microsoft-authenticode.htm) $$$
+* [Comodo](https://www.comodo.com/landing/ssl-certificate/authenticode-signature/) $
+* [GoDaddy](https://au.godaddy.com/web-security/code-signing-certificate) $$
 * Amongst others, please shop around to find one that suits your needs, Google is your friend :)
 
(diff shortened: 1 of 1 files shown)


docs/tutorial/code-signing.md
```diff
@@ -45,14 +45,14 @@ For more info, see the [Mac App Store Submission Guide].
 
 Before signing Windows builds, you must do the following:
 
-1. Get a Windows Authenticode code signing certificate
+1. Get a Windows Authenticode code signing certificate (requires an annual fee)
 2. Install Visual Studio 2015/2017 (to get the signing utility)
 
 You can get a code signing certificate from a lot of resellers, popular ones include:
```
ckerr: I wonder how useful the price rating is here -- all three links show the price prominently on the landing page, and it's possible that these prices could go up or down, making these values wrong.

IMO it would be better to remove the price guide and add a suggestion that people shop around before deciding which to purchase.

```suggestion
You can get a code signing certificate from a lot of resellers. Prices vary, so it may be worth your time to shop around. Popular resellers include:
```

docs/tutorial/code-signing.md
```diff
@@ -45,14 +45,14 @@ For more info, see the [Mac App Store Submission Guide].
...
 
-* [digicert](https://www.digicert.com/code-signing/microsoft-authenticode.htm)
-* [Comodo](https://www.comodo.com/landing/ssl-certificate/authenticode-signature/)
-* [GoDaddy](https://au.godaddy.com/web-security/code-signing-certificate)
+* [digicert](https://www.digicert.com/code-signing/microsoft-authenticode.htm) $$$
+* [Comodo](https://www.comodo.com/landing/ssl-certificate/authenticode-signature/) $
+* [GoDaddy](https://au.godaddy.com/web-security/code-signing-certificate) $$
```
ckerr: I'd leave these lines as they were without the $ price guide

docs/tutorial/code-signing.md
kapilepatel: Hi, I am a bit novice on github, can you please make these suggested changes yourself? I am not able to "Commit suggestion" probably because of the outdated file.

ckerr:
Thanks for the update!

PR #17584, opened 2019-03-27
docs: added info on bookmark return values for securityScopedBookmarks

rr326:
#### Description of Change
Dialog.md does not document the expected return values for securityScopedBookmarks. You have to look through the code or old issues to understand how it is supposed to work.

This fixes that. 

#### Checklist

#### Release Notes

Notes: Dialog.md does not document the expected return values for securityScopedBookmarks. You have to look through the code or old issues to understand how it is supposed to work.

### DIFF
--- a/docs/api/dialog.md
+++ b/docs/api/dialog.md
@@ -120,7 +120,7 @@ Returns `Promise<Object>` - Resolve with an object containing the following:
 
 * `canceled` Boolean - whether or not the dialog was canceled.
 * `filePaths` String[] - An array of file paths chosen by the user. If the dialog is cancelled this will be an empty array.
-* `bookmarks` String[] (optional) _macOS_ _mas_ - An array matching the `filePaths` array of base64 encoded strings which contains security scoped bookmark data. `securityScopedBookmarks` must be enabled for this to be populated.
+* `bookmarks` String[] (optional) _macOS_ _mas_ - An array matching the `filePaths` array of base64 encoded strings which contains security scoped bookmark data. `securityScopedBookmarks` must be enabled for this to be populated. (For return values, see [table here](#bookmarks-array).)
 
 The `browserWindow` argument allows the dialog to attach itself to a parent window, making it modal.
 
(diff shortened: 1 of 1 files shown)


rr326:
That sounds good to me, but I was trying to match the existing convention where it seems to repeat the complete information for every API call.  

I'm happy to make the change, but do we need to wait for any other confirmation from others, or do I just do it?

codebytere:
@rr326 you don't need confirmation, please make the change so we can move this forward!

rr326:
@codebytere Ok - I think I did this properly. I rebased off of the current version, made my changes, and then pushed to my branch. Please let me know if I did something incorrectly. (First pr.)

codebytere:
@rr326 looks like it's struggling a bit on the table linting:
```
AssertionError: We only support plain text, links, softbreaks, inline code, strong tags and paragraphs inside joinable tokens: expected 'table_open' to be one of [ Array(20) ]
```
if you've cloned Electron locally you can get a shorter feedback loop on this than waiting for CI by running `npm run lint:docs`

MarshallOfSound:
To fix the above error the table can't be inside a method documentation block.  It should be moved to it's own ## level block

rr326:
Ok - I think I've got it. Note - I noticed an error that one header (showSavedDialog) was incorrect, so I fixed that. It doesn't seem to warrant a whole separate PR process.

codebytere:
Last thing just looks like:
`Trailing whitespace in: /home/builduser/project/src/electron/docs/api/dialog.md`

docs/api/dialog.md
```diff
@@ -154,7 +155,7 @@ dialog.showOpenDialog(mainWindow, {
 })
 ```
 
-### `dialog.showSaveDialog([browserWindow, ]options)`
+### `dialog.showSaveDialogSync([browserWindow, ]options)`
```
codebytere: ah, this just got fixed yesterday actually! You'll see it on rebase

rr326: Sorry about that. Maybe I hadn't properly installed my fork. Now it's linting on its own and this one passes. Thanks for your patience.

codebytere: thank YOU for the contribution 🚀

zcbenz:
I have rebase this branch on master.

PR #17598, opened 2019-03-28
build: ensure we use objc regexps for export-patches

MarshallOfSound:
This will minimize the churn when folks with different versions of git export/import patches 👍 

Notes: no-notes

### DIFF
--- a/script/.electron.attributes
+++ b/script/.electron.attributes
@@ -1,1 +1,2 @@
 *.mm diff=objc
+*.m diff=objc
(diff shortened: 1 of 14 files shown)


script/.electron.attributes
```diff
@@ -0,0 +1 @@
+*.mm diff=objc
```
nornagon:
```suggestion
*.mm diff=objc
*.m diff=objc
```

per the `gitattributes` file in https://opensource.apple.com/tarballs/Git/Git-101.1.tar.gz

PR #17611, opened 2019-03-29
refactor: move //electron/manifests:manifests target into main BUILD.gn

nornagon:
#### Description of Change
Per @brenca:
> I think there were a couple of JSON manifests there alongside that BUILD.gn file, and when we removed those (https://github.com/electron/electron/commit/d7557cb66ab94c66b0f3bf14644bd728e3594148#diff-74295569778564ed979e8dd33e356257) it was left behind there. I don't know of any reason why it couldn't be moved to the main .gn file.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes


deepak1556:
My reason to keep it in a separate directory was to minimize the scope and size of main `BUILD.gn`. Any reason to maintain a single file for all the configs ?

nornagon:
Manifests don't seem to go in a separate folder in upstream Chromium, so IMO we should follow their lead. e.g.

```
../content/public/app/BUILD.gn
277:    "content_renderer_manifest.cc",
../chrome/browser/extensions/BUILD.gn
248:    "api/messaging/native_messaging_host_manifest.cc",
../services/file/public/cpp/BUILD.gn
7:    "manifest.cc",
../components/payments/content/BUILD.gn
54:    "web_app_manifest.cc",
```

I don't necessarily think that everything should be in a top-level `BUILD.gn` but having a `manifests/` directory just for manifests doesn't seem right to me either.

deepak1556:
> Manifests don't seem to go in a separate folder in upstream Chromium, so IMO we should follow their lead

Yeah agree with that :) , but subtle difference there is that browser, render, app have their respective build configs, which makes it easier for manifests to be localized accordingly. Maybe if we consider such a config separation that would lessen the impact on main build config file.

deepak1556:
> but having a manifests/ directory just for manifests doesn't seem right to me either.

Reminds me, another unwanted build config folder in the source tree is `components/pepper_flash/BUILD.gn`, we should move that config into `chromium_src/BUILD.gn`. Originally I created that folder to move some of the chrome features like flash, cdm, printing etc following the footsteps of upstream `//components`. But we were able to maintain them in `chromium_src/BUILD.gn`. Lets delete this one too :)

PR #17613, opened 2019-03-30
refactor: mojofy MessageTo and MessageHost

nornagon:
#### Description of Change
Based on #17406

This ports the `MessageTo` and `MessageHost` IPC calls to Mojo.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/api/atom_api_web_contents.cc
+++ b/atom/browser/api/atom_api_web_contents.cc
@@ -934,6 +934,13 @@ void WebContents::MessageTo(bool internal,
   }
 }
 
+void WebContents::MessageHost(const std::string& channel,
+                              base::Value arguments) {
+  // webContents.emit('ipc-message-host', new Event(), channel, args);
+  EmitWithSender("ipc-message-host", bindings_.dispatch_context(),
+                 base::nullopt, channel, std::move(arguments));
+}
+
 void WebContents::RenderFrameDeleted(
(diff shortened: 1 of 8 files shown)


atom/browser/api/atom_api_web_contents.cc
```diff
@@ -900,6 +904,12 @@ void WebContents::BindElectronBrowser(
   frame_to_bindings_map_[render_frame_host].push_back(id);
 }
 
+void WebContents::OnElectronBrowserConnectionError() {
```
MarshallOfSound: Do we know what could trigger this?  Is it something that could feasibly happen to users and should we somehow notify them that it happened?

nornagon: yes, this is triggered when the remote end closes the connection. It happens eg. if the renderer does something like:

```cpp
{
  mojom::ElectronBrowserPtr ptr;
  render_frame->GetRemoteInterfaces()->GetInterface(
      mojo::MakeRequest(&ptr));
  ptr->Message(/* ... */);
}  // ptr goes out of scope and is deleted; connection is closed.
```

Given the above sequence in the renderer, the browser will observe:

1. an interface is requested. the browser will bind the interface.
2. a method is called on the interface. the browser will invoke the call.
3. the connection is closed. the browser will clean up the connection and invoke the connection error callback.

PR #17440, opened 2019-03-18
feat: preliminary support for //extensions

samuelmaddock:
## Overview

These are the initial, exploratory changes for adding support for Chrome extensions using the [`//extensions`](https://cs.chromium.org/chromium/src/extensions/shell/) dependency from Chromium's source code. This dependency is meant to allow embedders of Chromium the ability to add extension support for their project.

Within the Electron community, there are [several projects](https://gist.github.com/samuelmaddock/bf4b862567e233da2f729f2ac0d04d7a) with their own implementation of Chrome extensions. No such implementation is complete and ends up duplicating the work of other implementations.

The goal of these changes are to introduce the first steps towards a more complete, official implementation of Chrome extensions for Electron-based projects.

#### What's included
- Loading of unpackaged Chrome extension
- Content script injection
- World isolation (aka context isolation) of content scripts
- Content script list in DevTools
- Extension manifest parsing and validation

#### What's _not_ included (yet)
- Background scripts
- DevTools extensions
- Extension badges
- Popup windows for extension badges

#### What _won't_ be included
- [Chrome platform apps](https://developer.chrome.com/apps/about_apps)
- [.CRX archive](https://developer.chrome.com/extensions/hosting) loader

### Usage

Currently an extension can be loaded by providing a `--load-extension=/absolute/path/to/extension` switch. This is planned to be superseded by JavaScript bindings prior to merging.

### Architecture

The file and code additions to this PR are mostly sourced from the [`//src/extensions/shell`](https://cs.chromium.org/chromium/src/extensions/shell/) project.

> app_shell is an experimental project to build a minimal environment like content_shell.
> The goal is to be able to run a v2 app and supply most of the chrome.* extension APIs without running the rest of Chrome.

In other words, the changes in this PR are mostly necessary boilerplate copied from app_shell to embed extensions.

app_shell implements support for Chrome Apps while we want support Chrome Extensions ([differences](https://developer.chrome.com/webstore/choosing)). Luckily they both share a majority of the same core APIs. The main addition we required was to create an instance of [`SharedUserScriptMaster`](https://cs.chromium.org/chromium/src/extensions/browser/shared_user_script_master.h). The history of commits provide a good timeline of APIs added.

Below are some of the notable classes added or changed as part of embedding extensions.

#### Browser

The entry point of changes for the browser process can be found in `atom_browser_main_parts.cc`.

| Class | Description |
|-|-|
[...]

PR #17614, opened 2019-03-30
fix: make devtools extensions load correctly

### LINKED ISSUE #17586: React Devtools can't connect in v5.0.0
### Preflight Checklist

### Issue Details

* **Electron Version: 5.0.0-beta.6**
* **Operating System: macOS Mojave**
* **Last Known Working Electron version: 4.0.0**

### Expected Behavior

React Devtools 3.6.0 work well in Electron v5.0.0
### To Reproduce

Open BrowserWindow with url point to a react page ( in my case, react page has been created by create-react-app)

### Screenshots

Screen Shot 2019-03-28 at 6 43 35 AM

### Additional Information
Terminal output
```
Attempted to load extension "React Developer Tools" that has already been loaded.
2019-03-28 06:39:19.922 electron[26697:1164642] *** WARNING: Textured window <EventDispatchingWindow: 0x7fe65dca97d0> is getting an implicitly transparent titlebar. This will break when linking against newer SDKs. Use NSWindow's -titlebarAppearsTransparent=YES instead.
[26697:0328/063944.495011:ERROR:CONSOLE(19)] "Uncaught TypeError: chrome.tabs.create is not a function", source: chrome-extension://react-developer-tools/build/panel.js (19)
[26697:0328/064024.449538:ERROR:CONSOLE(28)] "failed to connect", source: chrome-extension://react-developer-tools/build/panel.js (28)
```

```
Uncaught TypeError: chrome.tabs.create is not a function
```
I guest have bug when support chrome api
Thanks.

MarshallOfSound:
This is a multi-part fix as each fix revealed a new bug, so, yay I guess 😄 

Fixes #17586

#### `send_to_all` now works better'er

First off, we should remove this API, it's private and bad and we can remove it as soon as extension support lands for real in #17440 

Previously `send_to_all` was handled on the render_frame side  and messages were dispatched to the main frame which then proxied the event to all `WebLocalFrame`'s.  This missed out on all `WebRemoteFrame`'s which only became a problem recently.

This PR changes our `send_to_all` logic where it matters to send to all frames from the browser process.  There is a TODO to completely move our `send_to_all` logic and then remove it 👍 

#### Legacy DevToolsAPI

We were still using the legacy `DevToolsAPI` attach point, updated to use the `InspectorFrontendAPI` API

#### `chrome.browserAction` stubs

Newer versions of React Dev Tools require those APIs to exist or it throws 👍 

#### Fixing `chrome.runtime.connect`

The ipc utils refactor made these APIs stop working so now we use `invoke` on the renderer side correctly

Notes: DevTools Extensions now load correctly

### DIFF
--- a/atom/browser/api/atom_api_web_contents.cc
+++ b/atom/browser/api/atom_api_web_contents.cc
@@ -1649,13 +1649,24 @@ bool WebContents::SendIPCMessageWithSender(bool internal,
                                            const std::string& channel,
                                            const base::ListValue& args,
                                            int32_t sender_id) {
-  auto* frame_host = web_contents()->GetMainFrame();
-  if (frame_host) {
-    return frame_host->Send(new AtomFrameMsg_Message(frame_host->GetRoutingID(),
-                                                     internal, send_to_all,
-                                                     channel, args, sender_id));
+  std::vector<content::RenderFrameHost*> target_hosts;
+  if (!send_to_all) {
+    auto* frame_host = web_contents()->GetMainFrame();
+    if (frame_host) {
+      target_hosts.push_back(frame_host);
+    }
+  } else {
+    target_hosts = web_contents()->GetAllFrames();
   }
-  return false;
+
+  bool handled = false;
+  for (auto* frame_host : target_hosts) {
+    handled = frame_host->Send(
+                  new AtomFrameMsg_Message(frame_host->GetRoutingID(), internal,
+                                           false, channel, args, sender_id)) ||
+              handled;
+  }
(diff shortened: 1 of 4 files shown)


dsabanin:
Could you make a new 5 beta build with that change? That'd be very handy as we weren't able to give any betas a fair try with all our devtools being broken.

MarshallOfSound:
@dsabanin The `5-0-x` backport of this fix can be monitored here --> #17616

ErickTamayo:
Will this get to 6 and 7?

PR #17615, opened 2019-03-30
build: ensure the sccache server is running before we build

MarshallOfSound:
This fixes a flake seen locally a lot and on CI here --> https://circleci.com/gh/electron/electron/166604 where the sccache build times out because there server did not start in time.  `--start-server` does not time out 👍

Notes: no-notes


MarshallOfSound:
Closing out as no bandwidth to deal with

PR #17625, opened 2019-03-30
build: optionally exclude some parts of patches from being applied

alexeykuzmin:
#### Description of Change
The change targets Electron consumers who build Electron themselves.
Sometimes not all patch contents are needed, and it's not possible to apply it,
because some parts of the dependencies are missing.

The change is not supposed to affect Electron itself.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/lib/git.py
+++ b/script/lib/git.py
@@ -40,13 +40,17 @@ def get_repo_root(path):
   return get_repo_root(parent_path)
 
 
-def am(repo, patch_data, threeway=False, directory=None,
+def am(repo, patch_data, threeway=False, directory=None, exclude=None,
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
What's the use case for this? i.e. which patches do you want to conditionally apply

alexeykuzmin:
@MarshallOfSound I updated the description.

alexeykuzmin:
> if someone sweeps for dead code they might remove it

I would prefer this to happen, because I cannot guarantee that this code will actually be used by us _all the time_ in the future.

PR #17626, opened 2019-03-31
build: auto-generate the browserify step GN inputs

MarshallOfSound:
Previously when editing files that ended up in the browserify bundles you had to manually delete the bundles or `touch` the bundle entry point for ninja to realize it had to rebuild the bundles.

This updates our `filenames.auto.gni` generation script to auto-generate the file list for these bundles as well which will make life much easier.

Notes: no-notes


script/gen-filenames.js
```diff
@@ -10,13 +13,92 @@ const allDocs = fs.readdirSync(path.resolve(__dirname, '../docs/api'))
...
+    },
+    {
+      name: 'isolated',
+      entry: 'lib/isolated_renderer/init.js'
+    },
+    {
+      name: 'context_script',
+      entry: 'lib/content_script/init.js'
+    }
+  ]
```
nornagon: i'd love it if `context_script_browserify_deps` as mentioned in `filenames.gni` was greppable in this file. can we put the entire variable name here rather than just the bit before `_browserify_deps`?

MarshallOfSound:
@nornagon Done, should be greppable now 👍

PR #17642, opened 2019-04-01
fix: Make --explicitly-allowed-ports work with NetworkService.

alexeykuzmin:
https://chromium-review.googlesource.com/c/1400042
Landed in 73.0.3666.0, no need to backport to any other version.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed the --explicitly-allowed-ports switch.

### DIFF
--- a/patches/common/chromium/.patches
+++ b/patches/common/chromium/.patches
@@ -96,3 +96,4 @@ fix_system_tray_icons_being_cropped_under_kde.patch
 set_proper_permissions_for_package_s_framework_directory.patch
 intersection-observer.patch
 keyboard-lock-service-impl.patch
+make_--explicitly-allowed-ports_work_with_networkservice.patch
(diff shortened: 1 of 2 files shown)


alexeykuzmin:
@deepak1556 

> This switch never worked in electron

That's basically the reason, I wanted to use it and found out it doesn't work.

I need the flag to explicitly allow some ports on Windows CI machines I use internally at MS.
Some Electron tests start a Node server on a random port and Windows sometimes assigns port 2049 which is in the list of "restricted" ports in Chromium.
https://chromium.googlesource.com/chromium/src/+/master/net/base/port_util.cc#22

> do we want to fix this for older versions too ?

I seems like nobody else wants the flag to work, and I don't need it in Electron 3 =)
So I would say there's no need to backport the change anywhere else.

PR #17650, opened 2019-04-02
fix: make ServiceWorkers respect the session-wide User Agent override

### LINKED ISSUE #16196: Can't set userAgent in serviceWorker
* Output of `node_modules/.bin/electron --version`: v4.0.0
* Output of `node_modules/.bin/electron --version`: v5.0.0-beta.5
* Operating System (Platform and Version): MacOS 10.14.2

**Expected Behavior**

The userAgent of a serviceworker context should be settable.
None of the following methods work:

* `session.setUserAgent('foo')`
* `<webview useragent='foo' />`
* `win.loadURL(..,{ userAgent: 'foo'})`
* `webRequest.onBeforeSendHeaders([filter, ]listener)`

**Actual behavior**

Currently the `userAgent` of serviceworkers used in an electron app ist not settable.

**To Reproduce**

Check out the following example:
https://github.com/HaNdTriX/service-worker-user-agent-electron-bug-

If you provide a URL, please list the commands required to clone/setup/run your repo e.g.
```sh
$ git clone git@github.com:HaNdTriX/service-worker-user-agent-electron-bug-.git
$ npm install
$ npm start

Then reload the renderer once to see the value of navigator.userAgent
```

MarshallOfSound:
#### Description of Change

This adds a patch that exposes our `GetUserAgent` API on `content::BrowserContext` and then safely tries to fetch it during service worker initialization.  If no override has been set it falls back to the original technique.

Fixes #16196 

#### Release Notes

Notes: Fixed issue where `session.setUserAgent` did not override the user agent for service workers.


MarshallOfSound:
Following discussions with @deepak1556 I've updated the patch to be much more upstreamable.  It is now slightly larger but we should be able to upstream and remove 👍

patches/common/chromium/refactor_pass_the_browsercontext_through_to_getuseragent.patch
```diff
@@ -0,0 +1,274 @@
...
+diff --git a/content/browser/service_worker/service_worker_version.cc b/content/browser/service_worker/service_worker_version.cc
+index c6d61cc8795e1eb12f19b5eafec7e80446403afa..0bec70805949f37581eaf8a5e378b557c9aa071d 100644
+--- a/content/browser/service_worker/service_worker_version.cc
++++ b/content/browser/service_worker/service_worker_version.cc
+@@ -1634,7 +1634,7 @@ void ServiceWorkerVersion::StartWorkerInternal() {
+   params->scope = scope_;
+   params->script_url = script_url_;
+   params->script_type = script_type_;
+-  params->user_agent = GetContentClient()->browser()->GetUserAgent();
++  params->user_agent = context_->process_manager()->GetUserAgent(GetContentClient()->browser());
```
deepak1556: `GetContentClient()->browser()` can be accessed directly in `ServiceWorkerProcessManager`, since its a per process variable. Pass it as an argument isn't the usual style :)

MarshallOfSound: @deepak1556 That's what I thought but the thing failed to build when I tried to use it directly.  I'll try again when I try to upstream

MarshallOfSound:
Upstream CL: https://chromium-review.googlesource.com/c/chromium/src/+/1552100

deepak1556:
For the code path that sets useragent on network context params, its sufficient to pass `nullptr` as argument, since in chromium we create a default app level network context and also per context level network context. But lets see what the upstream team suggests :)

HaNdTriX:
@MarshallOfSound just wanted to say thank you so much for taking a shot at this 🙏!
Is there anything I can help you with to get this merged?

MarshallOfSound:
Closing due to the new `app.userAgentOverride` feature

PR #17658, opened 2019-04-02
fix: re-entracy problem with InvalidateFrameSinkId().

miniak:
#### Description of Change
Backport fix for https://bugs.chromium.org/p/chromium/issues/detail?id=907211
https://chromium-review.googlesource.com/c/chromium/src/+/1352525

> Fix re-entracy problem with InvalidateFrameSinkId().
> 
> HostFrameSinkManager::InvalidateFrameSinkId() can make a synchronous IPC
> to ensure whatever is drawing the platform window is destroyed before
> the platform window gets destroyed. However, while waiting for the
> synchronous IPC response other synchronous IPCs continue to get
> processed. |frame_sink_data_map_| can get mutated in response to a
> different synchronous IPC and |data| may no longer point to a valid
> memory location when DestroyCompositorFrameSink() returns.
> 
> Change the order so that all modifications to |data| happen before the
> synchronous IPC. Also add a comment to clarify that FrameSinkIds
> shouldn't be reused after they are invalidated. We don't do this today
> and it would cause more re-entrancy problems.
> 
> Also switch |frame_sink_data_map_| to use std::unordered_map instead of
> base::flat_map. This map can get large (tens of elements per open tab)
> so std::unordered_map is probably a better choice.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Fixed re-entracy problem with InvalidateFrameSinkId().

/cc @electron/wg-security

### DIFF
--- a/patches/common/chromium/.patches
+++ b/patches/common/chromium/.patches
@@ -98,3 +98,4 @@ intersection-observer.patch
 keyboard-lock-service-impl.patch
 make_--explicitly-allowed-ports_work_with_networkservice.patch
 fix_crashes_in_renderframeimpl_onselectpopupmenuitem_s.patch
+fix_re-entracy_problem_with_invalidateframesinkid.patch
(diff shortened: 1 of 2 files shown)


codebytere:
@miniak can you please fix conflicts so we can get this merged?

alexeykuzmin:
@codebytere done

PR #17664, opened 2019-04-02
build: ensure runner is defined inside the spec callback

### LINKED ISSUE #17648: Spec runner fails with confusing exception if 0 tests are run
Spec runner throws a confusing exception if no tests are selected by `--grep` (as at 75442b7, current master at time of writing).

To repro:
```
$ npm test -- -g "atseouhnaeohtsduidhtaieuthdeuain"
1..0
# tests 0
# pass 0
# fail 0
Unhandled exception in main spec runner: ReferenceError: Cannot access 'runner' before initialization
    at /Users/nornagon/work/electron-bootstrap/src/electron/spec-main/index.js:87:20
```

cc @MarshallOfSound

MarshallOfSound:
Fixes #17648 

Turns out if there are no tests the callback in `run` gets called synchronously before `runner` is defined 👍 

Notes: no-notes

### DIFF
--- a/spec-main/index.js
+++ b/spec-main/index.js
@@ -75,16 +75,19 @@ app.whenReady().then(() => {
 
   walker.on('end', () => {
     const runner = mocha.run(() => {
-      if (isCi && runner.hasOnly) {
-        try {
-          throw new Error('A spec contains a call to it.only or describe.only and should be reverted.')
-        } catch (error) {
-          console.error(error.stack || error)
+      // Ensure the callback is called after runner is defined
+      process.nextTick(() => {
+        if (isCi && runner.hasOnly) {
+          try {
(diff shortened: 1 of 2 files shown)


spec-main/index.js
```diff
@@ -75,16 +75,18 @@ app.whenReady().then(() => {
 
   walker.on('end', () => {
     const runner = mocha.run(() => {
-      if (isCi && runner.hasOnly) {
-        try {
-          throw new Error('A spec contains a call to it.only or describe.only and should be reverted.')
-        } catch (error) {
-          console.error(error.stack || error)
+      process.nextTick(() => {
```
nornagon: drop a comment here about why we're calling this in a later tick?


also it's a bummer that these two runners are copy/pasted from each other

PR #15782, opened 2018-11-20
chore: clean up FIXME and TODO comments in our tests

MarshallOfSound:
Clean up the `FIXME` and `TODO` comments in our tests

This enables a few tests and establishes a better pattern for platform-dependently skipping tests.

Notes: no-notes

PR #17681, opened 2019-04-03
[wip] test: workaround for nested describe skips

codebytere:
#### Description of Change

Corrects some tests that we were incorrectly marking as passed when they should be explicitly skipped instead.

cc @alexeykuzmin 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/spec/api-auto-updater-spec.js
+++ b/spec/api-auto-updater-spec.js
@@ -3,21 +3,15 @@ const { ipcRenderer } = require('electron')
 const { expect } = require('chai')
 
 describe('autoUpdater module', function () {
-  // XXX(alexeykuzmin): Calling `.skip()` in a 'before' hook
-  // doesn't affect nested 'describe's
   beforeEach(function () {
-    // Skip autoUpdater tests in MAS build.
-    if (process.mas) {
-      this.skip()
-    }
+    if (process.mas) this.skip()
   })
 
   describe('checkForUpdates', function () {
     it('emits an error on Windows when called the feed URL is not set', function (done) {
       if (process.platform !== 'win32') {
-        // FIXME(alexeykuzmin): Skip the test.
-        // this.skip()
-        return done()
+        this.test.parent.pending = true
+        this.skip()
       }
 
       ipcRenderer.once('auto-updater-error', (event, message) => {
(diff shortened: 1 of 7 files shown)


spec/api-auto-updater-spec.js
```diff
@@ -3,21 +3,15 @@ const { ipcRenderer } = require('electron')
...
+    if (process.mas) this.skip()
   })
 
   describe('checkForUpdates', function () {
     it('emits an error on Windows when called the feed URL is not set', function (done) {
       if (process.platform !== 'win32') {
-        // FIXME(alexeykuzmin): Skip the test.
-        // this.skip()
-        return done()
+        this.test.parent.pending = true
```
MarshallOfSound: I'm not a fan of hitting this internal mocha API.

Might be slightly biased but I prefer the approach I started taking in #15782 with declarative `platformDescribe` and `platformIt` methods.

spec/api-auto-updater-spec.js
```diff
@@ -3,21 +3,15 @@ const { ipcRenderer } = require('electron')
...
+        this.skip()
```
alexeykuzmin: `done` still has to be called after `.skip()`.

spec/api-auto-updater-spec.js
```diff
@@ -55,9 +48,7 @@ describe('autoUpdater module', function () {
       }
 
       before(function () {
-        if (process.platform !== 'win32' && process.platform !== 'darwin') {
-          this.skip()
-        }
+        if (process.platform === 'linux') this.skip()
```
alexeykuzmin: 👍

spec/api-auto-updater-spec.js
```diff
@@ -125,9 +114,8 @@ describe('autoUpdater module', function () {
   describe('quitAndInstall', () => {
     it('emits an error on Windows when no update is available', function (done) {
       if (process.platform !== 'win32') {
-        // FIXME(alexeykuzmin): Skip the test.
-        // this.skip()
-        return done()
+        this.test.parent.pending = true
+        this.skip()
```
alexeykuzmin: Call `done()` and `return`

alexeykuzmin:
I like half of the changes, but agree to Sam that usage of the internal API isn't great.
It can easily break when we update to a newer version of Mocha.

codebytere:
Yeah, that's fair. Let's shelve this one for now and see if there's a better approach we can potentially take!

Issue #11582, opened 2018-01-05
Picture in Picture Support macOS


fishcharlie:
Electron should add support for video Picture in Picture Support on macOS.

MarshallOfSound:
This Picture-in-Picture API on macOS is private and undocumented.  It's not something we can support at this time

tobiasbueschel:
Since Chrome version 70, Picture-in-Picture is [officially supported on macOS](https://www.zdnet.com/article/google-chrome-picture-in-picture-support-is-now-live-for-desktop-users/) 🎉

PR #17686, opened 2019-04-03
feat: enable picture-in-picture mode for video tags

### LINKED ISSUE #17369: PiP Window is not created - v5.0.X
* **Electron Version:**
  5.0.5-beta.5
* **Operating System:**
  MacOS Mojave 10.14.3
 Never worked I guess...

### Expected Behavior
PiP was introduced in Chromium at version 70, and should create a new window detached to the current window, with the video playing on it.

### Actual Behavior
The video is detached to be played in a PiP, and a window is created with dimensions of 0x0px.
You can play it, and the audio will work, but the video is not presented to the user.

### To Reproduce
Simply load this URL on any electron v5.0.X

https://googlechrome.github.io/samples/picture-in-picture/

It will show it is supported and the output in the page shows that a window is created.

Captura de Tela 2019-03-13 às 11 26 22

brenca:
#### Description of Change

This PR enables the picture-in-picture mode already available in `chrome`. This is partly a `feat:`, partly a `fix:` - without these changes, the `DOM` reports that PiP mode is available, but no actual window is created for it, and the `DOM` APIs return a 0x0 phantom PiP window. With these changes if PiP is disabled with the build flag, the `DOM` APIs correctly report that support for this feature is unavailable.

The added `//chrome` dependencies are self contained, but there are a couple of extra icons baked into the code when this feature is enabled. I'm happy to explore ways to only include the necessary icons, but we might we might end up with two folders with a small `BUILD.gn` in each that way, and I thought the easier maintainability/readability vs small size increase was worth it.

Resolves #17369 and #11582.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added support for picture-in-picture mode for video tags.

### DIFF
--- a/patches/common/chromium/picture-in-picture.patch
+++ b/patches/common/chromium/picture-in-picture.patch
@@ -1,0 +1,110 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Heilig Benedek <benecene@gmail.com>
+Date: Wed, 3 Apr 2019 16:11:00 +0200
+Subject: feat: enable picture in picture mode for video players
+
+
+diff --git a/chrome/browser/ui/views/overlay/back_to_tab_image_button.cc b/chrome/browser/ui/views/overlay/back_to_tab_image_button.cc
+index fae1519ec6c1d39b70716265385b1736d9905df8..fff8b3bca0403739d014cb242b271279574f97bf 100644
+--- a/chrome/browser/ui/views/overlay/back_to_tab_image_button.cc
++++ b/chrome/browser/ui/views/overlay/back_to_tab_image_button.cc
+@@ -5,7 +5,7 @@
+ #include "chrome/browser/ui/views/overlay/back_to_tab_image_button.h"
+ 
+ #include "chrome/app/vector_icons/vector_icons.h"
+-#include "chrome/grit/generated_resources.h"
++#include "electron/grit/electron_resources.h"
+ #include "third_party/skia/include/core/SkColor.h"
+ #include "ui/base/l10n/l10n_util.h"
+ #include "ui/gfx/color_palette.h"
+diff --git a/chrome/browser/ui/views/overlay/close_image_button.cc b/chrome/browser/ui/views/overlay/close_image_button.cc
+index 65e0eb972044cab6d687ef8c0ba21e4a6135f409..31bd5b086b47025432fcfeee36b44502573b8bf8 100644
+--- a/chrome/browser/ui/views/overlay/close_image_button.cc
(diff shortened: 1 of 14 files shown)


patches/common/chromium/picture-in-picture.patch
```diff
@@ -0,0 +1,110 @@
...
+ #include "ui/base/l10n/l10n_util.h"
+diff --git a/chrome/browser/ui/views/overlay/skip_ad_label_button.cc b/chrome/browser/ui/views/overlay/skip_ad_label_button.cc
+index bd1d9952d9fe055ba8ae5edfb0777a783d5fab48..50a0ee5e917a23c6901e4f10da8b62057567a451 100644
+--- a/chrome/browser/ui/views/overlay/skip_ad_label_button.cc
++++ b/chrome/browser/ui/views/overlay/skip_ad_label_button.cc
+@@ -4,7 +4,7 @@
+ 
+ #include "chrome/browser/ui/views/overlay/skip_ad_label_button.h"
+ 
+-#include "chrome/grit/generated_resources.h"
```
MarshallOfSound: Is there anyway we can pull some compiler/ninja voodoo to make this work without a patch

cc @nornagon @deepak1556

nornagon: Perhaps we could consider refactoring upstream to make these resources more self-contained?

deepak1556:
> requesting a 6-0-x backport as without it requesting a PiP view crashes

For `6-0-x` I would suggest disabling PIP web apis through the embedder hooks https://groups.google.com/a/chromium.org/d/msg/embedder-dev/A4zqjsmibXc/Osy_LLZ0DwAJ. I don't see the reason to ship this feature.

MarshallOfSound:
@deepak1556 I want to do that for `5-0-x` (disable it) but I'd like to make it possible to use this in `6-0-x`

deepak1556:
LGTM, Can you fix the conflicts.

ckerr:
@brenca ping, can you fix the conflicts

mohamedghan:
guys, i'm having a bad time trying to add pip in my project. i'm using electron 6.x.x beta15

alirezavalizade:
Any update on this pull request?

Ajeey:
Guys, any updates on this? Would love to see this ported to Electron!

alirezavalizade:
I updated to electron v7, but it doesn't work correctly. Actually it has the feature on video tag controls but when you click on it, it doesn't show anything, I only see something like this.

image

@brenca @MarshallOfSound @zcbenz

alectrocute:
Same issue here a few months after @alirezavalizade's comment above. This would be a really nice feature to have.

PR #17687, opened 2019-04-03
refactor: turn OnOffscreen message into a command-line flag

nornagon:
#### Description of Change
This turns the `AtomViewMsg_Offscreen` message, which only called the global setter `blink::WebView::SetUseExternalPopupMenus(false)` to get Chrome to render popup menus in the renderer instead of externally on macOS, into a command-line renderer flag `--offscreen` which does the same thing, except at render thread startup time, which is where Chromium sets the flag: https://chromium.googlesource.com/chromium/src/+/refs/tags/75.0.3755.3/content/renderer/render_thread_impl.cc#728.

This was the last usage of `RenderViewObserver` in our codebase, so this PR also removes that class.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes


atom/browser/web_contents_preferences.cc
```diff
@@ -309,6 +309,11 @@ void WebContentsPreferences::AppendCommandLineSwitches(
     command_line->AppendSwitchASCII(switches::kBackgroundColor, "#fff");
   }
 
+  // --offscreen
```
brenca: This should apply to `<webview>`s inside OSR windows as well. Those are not marked as `offscreen` in the `webPreferences` automatically if the parent is `offscreen`.

nornagon: Can you suggest how I might write a test for this? I was playing around in Fiddle trying to trigger a context menu to appear in an offscreen window and I couldn't figure out how to get it working.

PR #17688, opened 2019-04-03
[wip] feat: allow optional activation of app windows on show

codebytere:
#### Description of Change

Previously, our implementation of `app.show()` used [`unhide`](https://developer.apple.com/documentation/appkit/nsapplication/1428761-unhide?language=objc) unconditionally, which meant that hidden windows would be restored to the screen and automatically activated. There are many scenarios where a developer may not want that specific show behavior, and may instead want to show windows without automatically activating them.

This PR this uses [`unhideWithoutActivation`](https://developer.apple.com/documentation/appkit/nsapplication/1428566-unhidewithoutactivation?language=objc) to allow users to pass a boolean to `app.show()` for more granular control over this behavior.

cc @MarshallOfSound @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Allow optional activation of app windows on `app.show()`.

### DIFF
--- a/atom/browser/browser_mac.mm
+++ b/atom/browser/browser_mac.mm
@@ -36,8 +36,15 @@ void Browser::Hide() {
   [[AtomApplication sharedApplication] hide:nil];
 }
 
-void Browser::Show() {
-  [[AtomApplication sharedApplication] unhide:nil];
+void Browser::Show(mate::Arguments* args) {
+  bool should_activate;
+  args->GetNext(&should_activate);
+
+  if (should_activate) {
+    [[AtomApplication sharedApplication] unhide:nil];
+  } else {
+    [[AtomApplication sharedApplication] unhideWithoutActivation:nil];
+  }
 }
 
 void Browser::AddRecentDocument(const base::FilePath& path) {
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -565,9 +565,12 @@ the active app. On Windows, focuses on the application's first window.
 
 Hides all application windows without minimizing them.
 
-### `app.show()` _macOS_
+### `app.show(options)` _macOS_
 
-Shows application windows after they were hidden. Does not automatically focus
+* `options` Object
+  * `shouldActivate` String - whether or not app should activate windows after restoring them.
+
+Shows application windows after they were hidden. If `noActivate` is `true`, the app does not automatically focus
 them.
 
 ### `app.getAppPath()`
--- a/atom/browser/browser.h
+++ b/atom/browser/browser.h
@@ -118,7 +118,7 @@ class Browser : public WindowListObserver {
   void Hide();
 
   // Show the application.
-  void Show();
+  void Show(mate::Arguments* args);
 
   // Creates an activity and sets it as the one currently in use.
   void SetUserActivity(const std::string& type,


docs/api/app.md
```diff
@@ -565,9 +565,11 @@ the active app. On Windows, focuses on the application's first window.
 
 Hides all application windows without minimizing them.
 
-### `app.show()` _macOS_
+### `app.show(noActivate)` _macOS_
 
-Shows application windows after they were hidden. Does not automatically focus
+* `noActivate` String - whether or not app should activate windows after restoring them.
+
+Shows application windows after they were hidden. If `noActivate` is `true`, the app does not automatically focus
```
codebytere: this documentation was previously incorrect; our implementation did automatically focus the restored windows.

docs/api/app.md
```diff
@@ -565,9 +565,11 @@ the active app. On Windows, focuses on the application's first window.
 
 Hides all application windows without minimizing them.
 
-### `app.show()` _macOS_
+### `app.show(noActivate)` _macOS_
```
MarshallOfSound: Can we make this an options object with a `shouldActivate` key.

`app.show(false)` gives your average reader no idea as to what is about to happen or what the `false` means

codebytere:
it's always been a mac-only api but i can look into it!

docs/api/app.md
```diff
@@ -565,9 +565,12 @@ the active app. On Windows, focuses on the application's first window.
...
 Hides all application windows without minimizing them.
 
-### `app.show()` _macOS_
+### `app.show(options)` _macOS_
 
-Shows application windows after they were hidden. Does not automatically focus
+* `options` Object
+  * `shouldActivate` String - whether or not app should activate windows after restoring them.
+
+Shows application windows after they were hidden. If `noActivate` is `true`, the app does not automatically focus
```
nornagon:
```suggestion
Shows application windows after they were hidden. If `activate` is `false`, the app does not automatically focus
```

docs/api/app.md
```diff
@@ -565,9 +565,12 @@ the active app. On Windows, focuses on the application's first window.
 
 Hides all application windows without minimizing them.
 
-### `app.show()` _macOS_
+### `app.show(options)` _macOS_
 
-Shows application windows after they were hidden. Does not automatically focus
+* `options` Object
+  * `shouldActivate` String - whether or not app should activate windows after restoring them.
```
nornagon:
```suggestion
  * `activate` Boolean - activate windows after restoring them. Defaults to true.
```

atom/browser/browser_mac.mm
```diff
@@ -36,8 +36,15 @@
   [[AtomApplication sharedApplication] hide:nil];
 }
 
-void Browser::Show() {
-  [[AtomApplication sharedApplication] unhide:nil];
+void Browser::Show(mate::Arguments* args) {
+  bool should_activate;
+  args->GetNext(&should_activate);
```
nornagon: this doesn't match the docs, which say the argument should be an object

codebytere: oops yeah sorry this should say wip :) still working out some things

PR #17694, opened 2019-04-04
build: disable Vulkan support by default

alexeykuzmin:
_Maybe it shouldn't be done in the `master` but only in v4 and 5._

#### Description of Change

It was enabled in
https://chromium.googlesource.com/chromium/src.git/+/327326656f2c8c45ecb0ff4675e5c577e7c275ef which landed in 69.0.3460.0 (Electron 4).

Vulkan support would require a command line parameter "--enable-vulkan"
to be passes to a binary to be enabled anyway,
so this change doesn't actually alter the current behaviour of Electron.

Reasons to disable the flag back:
1. Electron has never supported it prior to version 4.
3. We build stuff we don't need.
4. We ship stuff we don't use, and Vulkan's DLLs size is > 0.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Disabled unintentionally enabled Vulkan support.

### DIFF
--- a/build/args/all.gn
+++ b/build/args/all.gn
@@ -7,6 +7,7 @@ v8_typed_array_max_size_in_heap = 0
 v8_embedder_string = "-electron.0"
 
 enable_cdm_host_verification = false
+enable_vulkan = false
 proprietary_codecs = true
 ffmpeg_branding = "Chrome"
 


nornagon:
I wouldn't say this was _unintentionally_ enabled. We inherit changes in Chrome intentionally.

Vulkan will eventually be required in Chrome (to support WebGPU, and additionally because OpenGL is deprecated on macOS), so we can't disable this forever. Disabling it in 4 and 5 might be reasonable if it's not usable/useful there anyway. What's the difference in binary size? If it's less than ~1MiB I don't think we should bother, since it will be required eventually anyway.

miniak:
@nornagon on 64-bit Windows builds, the Vulkan binaries are 12 MB uncompressed

miniak:
@nornagon event the latest version of Chrome on Windows does not include these Vulkan DLLs.

alexeykuzmin:
> I wouldn't say this was _unintentionally_ enabled. We inherit changes in Chrome intentionally.

The thing is enabled by default in Chromium, but I'm not sure about Chrome.

alexeykuzmin:
We are going to leave it as is in the `master` and only make the change in `4-1-x` and `5-1-x`.

PR #17695, opened 2019-04-04
docs: update guide on MAS entitlements and export compliance

samuelmeuli:
#### Description of Change

I updated the following information in the Mac App Store Submission Guide:

* Fixed the value of the app groups entitlement in the `parent.plist` snippet – needs to be a string array instead of a string (see [Apple's docs](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_application-groups))
* Updated the information on encryption export compliance: An ERN is apparently no longer required (see [App Store Connect Help](https://help.apple.com/app-store-connect/#/devc3f64248f)) and therefore the provided blog link is no longer relevant

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/docs/tutorial/mac-app-store-submission-guide.md
+++ b/docs/tutorial/mac-app-store-submission-guide.md
@@ -72,7 +72,9 @@ Then, you need to prepare three entitlements files.
     <key>com.apple.security.app-sandbox</key>
     <true/>
     <key>com.apple.security.application-groups</key>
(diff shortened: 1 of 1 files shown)


sethlu:
@MarshallOfSound That the security groups are specified as an array aligns with the documentation and should work fine 😸 Also in fact, `electron-osx-sign`'s been automatically injecting security group as an array since 2 years ago.

PR #17712, opened 2019-04-05
fix: ensure the sandboxed preloads globals do not leak

MarshallOfSound:
#### Description of Change
This PR re-adds the preload wrapping we used to have for sandboxed renderers that was accidentally removed as part of C71.  It takes a different approach partially due to API changes in node and partially for performance reasons.  Instead of wrapping the script at runtime, we wrap the scripts at build time.

This PR enables the `--standalone` flag for our browserify builds and then wraps the scripts with a fake` module: { exports }`.  Standalone will ensure nothing leaks and the fake exports will ensure that nothing gets injected onto the `window` object 👍 

#### Release Notes

Notes: Fixed issue where sandboxed renderers could sometimes leak globals outside of the preload script

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -10,6 +10,7 @@ import("//tools/grit/repack.gni")
 import("//tools/v8_context_snapshot/v8_context_snapshot.gni")
 import("//v8/snapshot_toolchain.gni")
 import("build/asar.gni")
+import("build/js_wrap.gni")
 import("build/npm.gni")
 import("build/tsc.gni")
 import("buildflags/buildflags.gni")
@@ -70,7 +71,7 @@ npm_action("build_electron_definitions") {
   ]
 }
 
-npm_action("atom_browserify_sandbox") {
+npm_action("atom_browserify_sandbox_unwrapped") {
   script = "browserify"
   deps = [
     ":build_electron_definitions",
@@ -79,7 +80,7 @@ npm_action("atom_browserify_sandbox") {
   inputs = auto_filenames.sandbox_browserify_deps
 
   outputs = [
-    "$target_gen_dir/js2c/preload_bundle.js",
+    "$target_gen_dir/js2c/preload_bundle_unwrapped.js",
   ]
 
   args = [
@@ -94,12 +95,14 @@ npm_action("atom_browserify_sandbox") {
     "-p",
     "tsconfig.electron.json",
     "]",
+    "--standalone",
+    "sandboxed_preload",
     "-o",
     rebase_path(outputs[0]),
   ]
 }
 
-npm_action("atom_browserify_isolated") {
+npm_action("atom_browserify_isolated_unwrapped") {
   script = "browserify"
   deps = [
     ":build_electron_definitions",
@@ -108,7 +111,7 @@ npm_action("atom_browserify_isolated") {
   inputs = auto_filenames.isolated_browserify_deps
 
   outputs = [
-    "$target_gen_dir/js2c/isolated_bundle.js",
+    "$target_gen_dir/js2c/isolated_bundle_unwrapped.js",
   ]
 
   args = [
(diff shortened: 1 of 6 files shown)


spec/api-browser-window-spec.js
```diff
@@ -1357,6 +1357,49 @@ describe('BrowserWindow module', () => {
     afterEach(() => { ipcMain.removeAllListeners('answer') })
 
     describe('"preload" option', () => {
+      const doesNotLeakSpec = (name, webPrefs) => {
+        it(name, async function () {
+          this.timeout(120000)
```
nornagon: does this really need an increased timeout? it's not doing anything that should take _2 minutes_.

spec/fixtures/api/no-leak.html
```diff
@@ -0,0 +1,17 @@
...
+<html lang="en">
+<head>
+  <title>Document</title>
+</head>
+<body>
+  <script>
+  window.postMessage({
+    require: typeof require,
+    exports: typeof exports,
+    windowRequire: typeof window.require,
```
nornagon: it's unclear to me in what circumstances `window.require` would refer to a different object than `require` here; what's this testing for? (same for `window.exports`)

spec/fixtures/module/empty.js
```diff
@@ -0,0 +1,6 @@
+const { ipcRenderer } = require('electron')
+console.log('doing nothing')
```
nornagon: is this needed?

BUILD.gn
```diff
@@ -94,11 +95,33 @@ npm_action("atom_browserify_sandbox") {
...
     "-o",
     rebase_path(outputs[0]),
   ]
 }
 
+js_wrap("atom_browserify_wrap") {
+  deps = [
+    ":atom_browserify_content_script",
+    ":atom_browserify_isolated",
+    ":atom_browserify_sandbox",
```
nornagon: consider breaking this into 3 separate targets with 1 input/output each. there's no particular reason all these need to be built together.

spec/api-browser-window-spec.js
MarshallOfSound: Oh no, my bad, I left this in from debugging

spec/fixtures/api/no-leak.html
MarshallOfSound: Unclear to me as well, but hey, better to test implict and explicit globals rather than let it break when something else funky happens

BUILD.gn
```diff
@@ -165,9 +214,9 @@ copy("atom_js2c_copy") {
 
 action("atom_js2c") {
   deps = [
-    ":atom_browserify_content_script",
-    ":atom_browserify_isolated",
-    ":atom_browserify_sandbox",
+    ":atom_browserify_content_script_wrap",
+    ":atom_browserify_isolated_wrap",
+    ":atom_browserify_sandbox_wrap",
```
nornagon: i'd prefer we label the unwrapped targets as `_unwrapped` and have the ones we actually depend on be unsuffixed, but it's not a big deal

build/js_wrap.py
```diff
@@ -0,0 +1,19 @@
...
+
+if len(in_bundles) is not len(out_bundles):
+  print("--out and --in must provide the same number of arguments")
+  sys.exit(1)
+
+for i in range(len(in_bundles)):
+  in_bundle = in_bundles[i]
+  out_path = out_bundles[i]
+  with open(in_bundle, 'r') as f:
+    lines = ["(function(){var exports={},module={exports};"] + f.readlines() + ["})()"]
```
nornagon: these are never expected to be `cat`ed with anything, right? if so then we'd probably also want a `;` at the end there

spec/fixtures/api/no-leak.html
nornagon: I disagree; don't test something unless you actually understand what you're testing for :)

At the very least, be consistent with `sandboxed_preload`.

build/js_wrap.py
MarshallOfSound: They're never supposed to be, but hey, let's through a `;` in for good measure 👍

codebytere:
@MarshallOfSound this is gonna need manual bp to `5-0-x`

PR #17728, opened 2019-04-08
feat: support ELECTRON_DEFAULT_ERROR_MODE in the GPU process

miniak:
#### Description of Change
It should be possible to enable Windows Error Reporting for the GPU process as well.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added missing support for `ELECTRON_DEFAULT_ERROR_MODE` in the GPU process

### DIFF
--- a/patches/common/chromium/add_contentgpuclient_precreatemessageloop_callback.patch
+++ b/patches/common/chromium/add_contentgpuclient_precreatemessageloop_callback.patch
@@ -1,0 +1,48 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Milan Burda <milan.burda@gmail.com>
+Date: Thu, 11 Apr 2019 14:49:20 +0200
+Subject: Add ContentGpuClient::PreCreateMessageLoop() callback
+
+Invoke in GpuMain after SetErrorMode, before starting the message loop.
+Allows Electron to restore WER when ELECTRON_DEFAULT_ERROR_MODE is set.
+
+This should be upstreamed
+
+diff --git a/content/gpu/gpu_main.cc b/content/gpu/gpu_main.cc
+index 39e87967360a9c7a2856ce4df9c182c782efb2a5..ade3e7905d01b25128f477c68c10edd0f96626c3 100644
+--- a/content/gpu/gpu_main.cc
++++ b/content/gpu/gpu_main.cc
+@@ -234,6 +234,10 @@ int GpuMain(const MainFunctionParams& parameters) {
+ 
+   logging::SetLogMessageHandler(GpuProcessLogMessageHandler);
+ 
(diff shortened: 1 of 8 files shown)


patches/common/chromium/support_electron_default_error_mode_in_the_gpu_process.patch
```diff
@@ -0,0 +1,29 @@
...
+ #include "base/lazy_instance.h"
+ #include "base/message_loop/message_loop.h"
+ #include "base/metrics/histogram_macros.h"
+@@ -225,6 +226,10 @@ int GpuMain(const MainFunctionParams& parameters) {
+       SEM_NOGPFAULTERRORBOX |
+       SEM_NOOPENFILEERRORBOX);
+
++  auto env = base::Environment::Create();
++  if (env->HasVar("ELECTRON_DEFAULT_ERROR_MODE"))
++    SetErrorMode(GetErrorMode() & ~SEM_NOGPFAULTERRORBOX);
```
nornagon: I don't love this; it's not upstreamable. Can we perhaps do this through a hook that's available somewhere in the GPU process startup path? If not, perhaps we could add such a hook?

miniak: I will investigate that option

miniak: Added a hook as suggested, can you please have another look?

miniak:
@deepak1556, @zcbenz can you have a look?

miniak:
```
ERROR at //electron/atom/browser/atom_gpu_client.h:8:11: Can't include this header from here.
#include "content/public/gpu/content_gpu_client.h"
          ^--------------------------------------
The target:
  //electron:electron_lib
  //content/public/gpu:gpu_sources

to include headers.

  //electron:electron_lib -->
  //content/public/app:both --[private]-->
  //content/public/gpu:gpu_sources
```

patches/common/chromium/add_contentgpuclient_precreatemessageloop_callback.patch
```diff
@@ -0,0 +1,44 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Milan Burda <milan.burda@gmail.com>
+Date: Thu, 11 Apr 2019 14:49:20 +0200
+Subject: Add ContentGpuClient::PreCreateMessageLoop() callback
+
```
nornagon: Please add a description to the patch, preferably w/ a link to the Chromium CL once you upload it.

miniak: Added description. I haven't tried upstreaming Chromium patches yet. I will update the patch with a CL link in case I do it in a follow up PR, is that ok?

PR #17730, opened 2019-04-09
docs: Add missing parameter from ipc-main.md

dustingraham:
#### Description of Change

Reference: https://electronjs.org/docs/api/ipc-main missing `sender` parameter in `event.sender.reply`

#### Checklist

#### Release Notes

Notes: `no-notes`

### DIFF
--- a/docs/api/ipc-main.md
+++ b/docs/api/ipc-main.md
@@ -30,7 +30,7 @@ processes:
 const { ipcMain } = require('electron')
 ipcMain.on('asynchronous-message', (event, arg) => {
   console.log(arg) // prints "ping"
-  event.sender.reply('asynchronous-reply', 'pong')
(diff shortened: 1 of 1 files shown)


dustingraham:
Happy to help. The build pipeline is pretty awesome. But, this is a bit overkill for a couple documentation updates I've noticed.

MarshallOfSound:
Hey @dustingraham

Thanks for the PR, however the current example in master is correct for the APIs available in master.

`evenr.reply` is a new API added in `5-0-x` and therefore isn't on the website yet.

Refs: https://github.com/electron/electron/blob/master/docs/api/structures/ipc-main-event.md

dustingraham:
Understood, thank you.

https://github.com/electron/electron/blob/5-0-x/docs/api/ipc-main.md Might not be fixed there.

PR #17742, opened 2019-04-10
feat: allow pages to override window.history.length

magne4000:
#### Description of Change
If a page tries to override `window.history.length`, an error is thrown.
image

Example: Login to [Wunderlist](https://www.wunderlist.com/webapp) inside a webview or BrowserWindow, the above error will be thrown.

#### Checklist
- [x] `npm test` passes

#### Release Notes
notes: Fixed Wunderlist app not loading

### DIFF
--- a/lib/renderer/window-setup.ts
+++ b/lib/renderer/window-setup.ts
@@ -233,7 +233,8 @@ export const windowSetup = (
   defineProperty(window.history, 'length', {
     get: function () {
       return ipcRendererInternal.sendSync('ELECTRON_NAVIGATION_CONTROLLER_LENGTH')
-    }
+    },
+    configurable: true
   })
 
   if (guestInstanceId != null) {
--- a/lib/renderer/window-setup.ts
+++ b/lib/renderer/window-setup.ts
@@ -234,7 +234,7 @@ export const windowSetup = (
     get: function () {
       return ipcRendererInternal.sendSync('ELECTRON_NAVIGATION_CONTROLLER_LENGTH')
     },
-    set() {}
+    set () {}
   })
 
   if (guestInstanceId != null) {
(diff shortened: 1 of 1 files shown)


magne4000:
Current bypass for those interested (in a preload):
```js
  // History object without `length` prop.
  const { length, ...histo } = window.history;
  // Define the length prop as configurable
  defineProperty(histo, 'length', {
    get() {
      return ipcRenderer.sendSync('ELECTRON_NAVIGATION_CONTROLLER_LENGTH');
    },
    configurable: true,
  });
  // Affect new object to window.history
  defineProperty(window, 'history', {
    value: histo,
    configurable: true,
    enumerable: true,
  });
```

lib/renderer/window-setup.ts
```diff
@@ -233,7 +233,8 @@ export const windowSetup = (
   defineProperty(window.history, 'length', {
     get: function () {
       return ipcRendererInternal.sendSync('ELECTRON_NAVIGATION_CONTROLLER_LENGTH')
-    }
+    },
+    configurable: true
```
MarshallOfSound: The behavior in browsers isn't that this is configurable, if you do `window.history.length = 100` in Chrome it doesn't throw but `window.history.length` doesn't change.

We should instead make a noop-setter here.
```suggestion
    set() {}
```

magne4000:
@MarshallOfSound do you think we should do the same for the 2 `defineProperty ` at the bottom of the file?
https://github.com/electron/electron/blob/d597a0e8b0121b54375412af3d8a4a9db60ba21a/lib/renderer/window-setup.ts#L264

codebytere:
@magne4000 lint error needs addressing:

```
/home/builduser/project/src/electron/lib/renderer/window-setup.ts
  237:8  error  Missing space before function parentheses  space-before-function-paren

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

and i think for the purposes of this PR we can scope just to `window.history.length`!

miniak:
@magne4000 you cannot access `ipcRendererInternal` in the preload script

magne4000:
I removed my comment with a workaround because it was not working properly

zcbenz:
@MarshallOfSound Can you take another look at this?

PR #17747, opened 2019-04-10
feat: add webContents 'zoom-changed' event

miniak:
#### Description of Change
Emit `zoom-changed` event when the user is trying to zoom in / out using the mouse wheel on Windows. This is the only way the Electron app will be able to handle these events when the focus is inside a (x-origin) iframe.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `zoom-changed` event to `webContents`, which is emitted when the user is trying to zoom in / out using the mouse wheel on Windows, even if the focus is inside an (x-origin) iframe.

### DIFF
--- a/atom/browser/api/atom_api_web_contents.cc
+++ b/atom/browser/api/atom_api_web_contents.cc
@@ -235,6 +235,27 @@ struct Converter<atom::api::WebContents::Type> {
   }
 };
 
+template <>
+struct Converter<atom::WebContentsZoomController::ZoomDirection> {
+  static v8::Local<v8::Value> ToV8(
+      v8::Isolate* isolate,
+      atom::WebContentsZoomController::ZoomDirection val) {
+    using ZoomDirection = atom::WebContentsZoomController::ZoomDirection;
+    std::string direction = "";
+    switch (val) {
+      case ZoomDirection::kIn:
+        direction = "in";
+        break;
+      case ZoomDirection::kOut:
+        direction = "out";
+        break;
+      default:
+        break;
+    }
+    return mate::ConvertToV8(isolate, direction);
+  }
+};
+
 }  // namespace mate
 
 namespace atom {
@@ -634,7 +655,11 @@ content::KeyboardEventProcessingResult WebContents::PreHandleKeyboardEvent(
 }
 
 void WebContents::ContentsZoomChange(bool zoom_in) {
-  Emit("zoom-changed", zoom_in);
+  const atom::WebContentsZoomController::ZoomDirection zoom_direction =
+      zoom_in ? atom::WebContentsZoomController::ZoomDirection::kIn
+              : atom::WebContentsZoomController::ZoomDirection::kOut;
+
+  Emit("zoom-changed", zoom_direction);
 }
 
 void WebContents::EnterFullscreenModeForTab(
--- a/atom/browser/api/atom_api_web_contents.cc
+++ b/atom/browser/api/atom_api_web_contents.cc
@@ -235,27 +235,6 @@ struct Converter<atom::api::WebContents::Type> {
   }
 };
 
-template <>
-struct Converter<atom::WebContentsZoomController::ZoomDirection> {
-  static v8::Local<v8::Value> ToV8(
-      v8::Isolate* isolate,
-      atom::WebContentsZoomController::ZoomDirection val) {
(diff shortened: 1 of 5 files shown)


spec/api-web-contents-spec.js
```diff
@@ -347,6 +347,54 @@ describe('webContents module', () => {
...
+  describe('zoom-changed', () => {
+    before(function () {
+      // On Mac, zooming isn't done with the mouse wheel.
+      if (process.platform === 'darwin') {
+        this.skip()
+      }
+    })
+
+    it('is emitted with the correct zooming info', (done) => {
+      w.loadURL(`file://${path.join(__dirname, 'fixtures', 'pages', 'base-page.html')}`)
```
miniak: @ppontes can you please use `loadFile` instead?

ppontes: 👍🏻

codebytere:
@miniak this needs documentation :)

docs/api/web-contents.md
```diff
@@ -369,6 +369,14 @@ Emitted when the window enters a full-screen state triggered by HTML API.
 
 Emitted when the window leaves a full-screen state triggered by HTML API.
 
+#### Event: 'zoom-changed'
+
+Returns:
+* `event` Event
+* `zoomedIn` Boolean - true if the user is trying to zoom in, false if the user is trying to zoom out
```
MarshallOfSound: Is this really the Chromium API?

We must be able to do better than this.  How about making this `wheelDeltaY` similar to the JS side scroll event with `+1` and `-1` as possible values.  I won't block this PR but a boolean indicating zooming in / our seems like a bit of a rough API here

miniak: @ppontes It makes sense

miniak: @ppontes what do you think?

ppontes: @MarshallOfSound, I find it odd if an event called `zoom-changed` will have a `wheelDeltaY` parameter - they don't seem semantically related. `zoom-changed` represents a complete zoom action. Users will likely be interested in knowing if the zoom level is to increase or decrease, not in the particular movement of the mouse wheel (even though they're correlated). That's how I understood the Chromium API and kept it. But I don't oppose changing the parameter as long as we also rename the event so that its name makes it clear this is a CTRL + mouse wheel event. Something like `mouse-wheel-zoom-changed`. What do you think?

MarshallOfSound: Hm this is a good point, I guess what I'm trying to convey here is that using a Boolean to represent a direction feels wrong to me.

Maybe an enum `ZoomDirection.OUT`?

atom/browser/api/atom_api_web_contents.cc
```diff
@@ -630,6 +651,14 @@ content::KeyboardEventProcessingResult WebContents::PreHandleKeyboardEvent(
   return content::KeyboardEventProcessingResult::NOT_HANDLED;
 }
 
+void WebContents::ContentsZoomChange(bool zoom_in) {
+  const atom::WebContentsZoomController::ZoomDirection zoom_direction =
+      zoom_in ? atom::WebContentsZoomController::ZoomDirection::kIn
+              : atom::WebContentsZoomController::ZoomDirection::kOut;
+
+  Emit("zoom-changed", zoom_direction);
```
miniak: @MarshallOfSound what about just `zoom_in ? "in" : "out"`?

spec/api-web-contents-spec.js
```diff
@@ -345,6 +345,54 @@ describe('webContents module', () => {
...
+    before(function () {
+      // On Mac, zooming isn't done with the mouse wheel.
+      if (process.platform === 'darwin') {
+        this.skip()
+      }
+    })
+
+    it('is emitted with the correct zooming info', (done) => {
+      w.loadFile(path.join(fixtures, 'pages', 'base-page.html'))
+      w.webContents.once('did-finish-load', () => {
```
alexeykuzmin: Can you use `async/await` here please?

miniak: done

miniak:
@MarshallOfSound are you ok with the current state of the PR?

vanowm:
What about zoom changed by [menuItem](https://www.electronjs.org/docs/v14-x-y/api/menu-item#menuitemrole) with role `zoomReset`, `zoomIn` and `zoomOut`, is there events for these?

PR #17757, opened 2019-04-11
fix: populate phase of WebMouseWheelEvents in webContents.sendInputEvent

ppontes:
When wheel scroll latching and async wheel events are enabled, phase information gets added to `WebMouseWheelEvents` when they are generated from native events. Chromium expects this information to be available and asserts that expectation with a `DCHECK`. Phase information must be synthesized and set to a `WebMouseWheelEvent` that is generated with `webContents.sendInputEvent` or a `DCHECK` will fail in Chromium.

The very same problem was found and fixed in Chromium for the simulation of mouse wheel events from the devtools. See [ticket](https://bugs.chromium.org/p/chromium/issues/detail?id=756524) and [fix commit](https://chromium.googlesource.com/chromium/src.git/+/2abac3742f431d44ebad70c8c052d55432bfbd9d). This is basically a port of that fix to `webContents.sendInputEvent`.

This makes it possible to run the test in [this PR](https://github.com/electron/electron/pull/17747).

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Using `webContents.sendInputEvent` to send a `WebMouseWheelEvent` now has the expected effect.

### DIFF
--- a/atom/browser/api/atom_api_web_contents.cc
+++ b/atom/browser/api/atom_api_web_contents.cc
@@ -1804,6 +1804,19 @@ void WebContents::SendInputEvent(v8::Isolate* isolate,
             mouse_wheel_event);
 #endif
       } else {
+        // Chromium expects phase info in wheel events (and applies a
+        // DCHECK to verify it). See: https://crbug.com/756524.
+        mouse_wheel_event.phase = blink::WebMouseWheelEvent::kPhaseBegan;
+        mouse_wheel_event.dispatch_type = blink::WebInputEvent::kBlocking;
+        rwh->ForwardWheelEvent(mouse_wheel_event);
+
+        // Send a synthetic wheel event with phaseEnded to finish scrolling.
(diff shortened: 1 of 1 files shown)


atom/browser/api/atom_api_web_contents.cc
```diff
@@ -1804,6 +1804,19 @@ void WebContents::SendInputEvent(v8::Isolate* isolate,
             mouse_wheel_event);
 #endif
       } else {
+        // Populate phase info for the wheel event, otherwise DCHECK
+        // will fail in Chromium. See: CR issue 756524
```
alexeykuzmin: Would be better to add a link here https://crbug.com/756524. Not a big deal though.

ppontes: Added.

ppontes:
@alexeykuzmin updated the release note to be more relevant to users.

PR #17759, opened 2019-04-11
docs: explanation on using protocols with partitions

Kilian:
#### Description of Change
I just spent the past two days debugging and figuring out why a custom protocol didn't work. Turns out it was the partition on my browserwindow's webPreferences. That's not documented anywhere, so I added it.

#### Release Notes

Notes: no-notes

### DIFF
--- a/docs/api/protocol.md
+++ b/docs/api/protocol.md
@@ -35,16 +35,16 @@ const { session, app, protocol } = require('electron')
 const path = require('path')
 
 app.on('ready', () => {
-  const partition = "persist:example"
+  const partition = 'persist:example'
   const ses = session.fromPartition(partition)
-  
+
   ses.protocol.registerFileProtocol('atom', (request, callback) => {
     const url = request.url.substr(7)
     callback({ path: path.normalize(`${__dirname}/${url}`) })
   }, (error) => {
     if (error) console.error('Failed to register protocol')
-  });
-  
+  })
+
   mainWindow = new BrowserWindow({
     width: 800,
     height: 600,
--- a/docs/api/protocol.md
+++ b/docs/api/protocol.md
@@ -24,7 +24,7 @@ app.on('ready', () => {
 **Note:** All methods unless specified can only be used after the `ready` event
 of the `app` module gets emitted.
 
-## Using `protocol` with a `partition` in `webPreferences`
+## Using `protocol` with a custom `partition` or `session`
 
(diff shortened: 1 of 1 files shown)


docs/api/protocol.md
```diff
@@ -24,6 +24,38 @@ app.on('ready', () => {
 **Note:** All methods unless specified can only be used after the `ready` event
 of the `app` module gets emitted.
 
+## Using `protocol` with a `partition` in `webPreferences`
```
MarshallOfSound:
```suggestion
## Using `protocol` with a custom `partition` or `session`
```

docs/api/protocol.md
```diff
@@ -24,6 +24,38 @@ app.on('ready', () => {
...
+
+A protocol is registered to the default Electron session. If you define a `partition` on your `browserWindow`'s `webPreferences`, then that partition will use a different session and your custom protocol will not work.
```
MarshallOfSound:
```suggestion
A protocol is registered to a specific Electron [`session`](./session.md) object. If you define a `partition` or `session` on your `browserWindow`'s `webPreferences`, then that window will use a different session and your custom protocol will not work if you just use `electron.protocol.XXX`.
```

docs/api/protocol.md
```diff
@@ -24,6 +24,38 @@ app.on('ready', () => {
...
+
+To have your custom protocol work in combination with a partition, you need to register it to the session of that partition:
```
MarshallOfSound:
```suggestion
To have your custom protocol work in combination with a custom session, you need to register it to that session explicitly.
```

docs/api/protocol.md
```diff
@@ -24,6 +24,38 @@ app.on('ready', () => {
...
+  mainWindow = new BrowserWindow({
+    width: 800,
+    height: 600,
+    webPreferences: {
+      partition: partition
+    }
+  })
+})
+```
+Using `protocol.registerStandardSchemes` without the session will still register your custom protocol as a standard scheme.
```
MarshallOfSound: Although not something that needs to be fixed here, this is confusing / unexplainable.  We should separate those APIs if some are per-session and others aren't.

Kilian: There's also the (currently undocumented) getStandardSchemes, which behaves the same.

codebytere:
@Kilian would you mind popping open manuals to 4 and 5 for this pr?

PR #17762, opened 2019-04-11
feat: add process.getBlinkMemoryInfo()

miniak:
#### Description of Change
Expose blink memory information provided by https://cs.chromium.org/chromium/src/third_party/blink/renderer/platform/heap/process_heap.h

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `process.getBlinkMemoryInfo()`.

### DIFF
--- a/atom/common/api/electron_bindings.cc
+++ b/atom/common/api/electron_bindings.cc
@@ -29,6 +29,7 @@
 #include "native_mate/dictionary.h"
 #include "services/resource_coordinator/public/cpp/memory_instrumentation/global_memory_dump.h"
 #include "services/resource_coordinator/public/cpp/memory_instrumentation/memory_instrumentation.h"
+#include "third_party/blink/renderer/platform/heap/process_heap.h"  // nogncheck
 
 namespace atom {
 
@@ -68,6 +69,7 @@ void ElectronBindings::BindProcess(v8::Isolate* isolate,
   process->SetMethod("log", &Log);
   process->SetMethod("getCreationTime", &GetCreationTime);
   process->SetMethod("getHeapStatistics", &GetHeapStatistics);
+  process->SetMethod("getBlinkMemoryInfo", &GetBlinkMemoryInfo);
   process->SetMethod("getProcessMemoryInfo", &GetProcessMemoryInfo);
   process->SetMethod("getSystemMemoryInfo", &GetSystemMemoryInfo);
   process->SetMethod("getSystemVersion",
@@ -250,6 +252,21 @@ v8::Local<v8::Promise> ElectronBindings::GetProcessMemoryInfo(
           base::BindOnce(&ElectronBindings::DidReceiveMemoryDump,
                          std::move(context), std::move(promise)));
   return handle;
+}
+
+// static
+v8::Local<v8::Value> ElectronBindings::GetBlinkMemoryInfo(
+    v8::Isolate* isolate) {
+  auto allocated = blink::ProcessHeap::TotalAllocatedObjectSize();
+  auto marked = blink::ProcessHeap::TotalMarkedObjectSize();
+  auto total = blink::ProcessHeap::TotalAllocatedSpace();
+
+  mate::Dictionary dict = mate::Dictionary::CreateEmpty(isolate);
+  dict.SetHidden("simple", true);
+  dict.Set("allocated", static_cast<double>(allocated >> 10));
+  dict.Set("marked", static_cast<double>(marked >> 10));
+  dict.Set("total", static_cast<double>(total >> 10));
+  return dict.GetHandle();
 }
 
 // static
--- a/docs/api/process.md
+++ b/docs/api/process.md
@@ -15,6 +15,7 @@ In sandboxed renderers the `process` object contains only a subset of the APIs:
 - `hang()`
 - `getCreationTime()`
 - `getHeapStatistics()`
+- `getBlinkMemoryInfo()`
 - `getProcessMemoryInfo()`
 - `getSystemMemoryInfo()`
 - `getSystemVersion()`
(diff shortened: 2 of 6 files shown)


MarshallOfSound:
Few questions before my plane takes off.

How does this line up with the existing process.getMemoryInfo API?

Can we / should we combine them?

How does the unified heap (shipped in 73 iirc) impact these APIs?

And then can we strongly document the answers to the above questions so when users look at these APIs they know what the values mean / represent

miniak:
@MarshallOfSound actually that blink API I've used seems to be private. I am negotiating with our performance team what they actually need. Closing for now.

miniak:
hmm, `/ nogncheck` supresses the error

docs/api/process.md
```diff
@@ -170,6 +171,16 @@ Returns `Object`:
...
 
+### `process.getBlinkMemoryInfo()`
+
+Returns `Object`:
+
+* `allocated` Integer
+* `marked` Integer
+* `total` Integer
+
+Returns an object with blink memory information. Note that all values are reported in Kilobytes.
```
alexeykuzmin: "Blink"

MarshallOfSound:
@miniak Any update on the above questions?

ahujasusheel:
@MarshallOfSound 
> Few questions before my plane takes off.
> 
> How does this line up with the existing process.getMemoryInfo API?
> I assume the question is on getProcessMemoryInfo. Since getProcessMemoryInfo gives back the memory info in terms of resident set, private and shared bytes, and since these are used to measure how any process is behaving and known measures of process with respect to OS is slightly different from getBlinkMemoryInfo, which gives an idea of how much memory is taken by blink_gc/oilpan which is specific to chromium and electron. 
> Can we / should we combine them?
> We should not combine them because these are used for debugging different issue. Blink memory info will be used mostly to debug rendering/dom related memory issues.
> How does the unified heap (shipped in 73 iirc) impact these APIs?
> We are investigating this.
> And then can we strongly document the answers to the above questions so when users look at these APIs they know what the values mean / represent
> Sure, we should do this.

miniak:
@MarshallOfSound can you please check @ahujasusheel's comment. I was not formatted properly and therefore appeared as spam. I've fixed it.

MarshallOfSound:
@miniak Yeah I'm just waiting on the output of

> We are investigating this.

and

> And then can we strongly document the answers to the above questions so when users look at these

> Sure, we should do this.

Docs + Unified Heap impact

miniak:
@MarshallOfSound will talk to my colleagues about those and hopefully update the PR soon

miniak:
@MarshallOfSound I've updated the docs. Is it sufficient or do we need even more details? @ahujasusheel updated his reply, it should answer all the questions now (Unified Heap impact + not more We are investigating this)

docs/api/process.md
```diff
@@ -170,6 +171,18 @@ Returns `Object`:
 
 Returns an object with V8 heap statistics. Note that all statistics are reported in Kilobytes.
 
+### `process.getBlinkMemoryInfo()`
+
+Returns `Object`:
+
+* `allocated` Integer - Total allocated object size in Kilobytes.
+* `marked` Integer - Total marked object size in Kilobytes.
+* `total` Integer - Total allocated space in Kilobytes.
```
MarshallOfSound: This probably shouldn't use the word "allocated" because now `allocated` and `total` both say "Total allocated"

docs/api/process.md
```diff
@@ -170,6 +171,18 @@ Returns `Object`:
 
 Returns an object with V8 heap statistics. Note that all statistics are reported in Kilobytes.
 
+### `process.getBlinkMemoryInfo()`
```
MarshallOfSound: Can you confirm this API works in the main process?  This API is living in `common/` in our code but the `TotalAllocatedObjectSize` APIs are `blink/renderer`

docs/api/process.md
MarshallOfSound: Or it should just be clarified the difference between this and `allocated`.

docs/api/process.md
```diff
@@ -170,6 +171,18 @@ Returns `Object`:
...
+
+Returns `Object`:
+
+* `allocated` Integer - Total allocated object size in Kilobytes.
```
MarshallOfSound:
```suggestion
* `allocated` Integer -Size of all allocated objects Kilobytes.
```

docs/api/process.md
```diff
@@ -170,6 +171,18 @@ Returns `Object`:
...
+* `marked` Integer - Total marked object size in Kilobytes.
```
MarshallOfSound:
```suggestion
* `marked` Integer - Size of all marked objects in Kilobytes.
```

docs/api/process.md
miniak: it does work, it just returns zeros

miniak:
@MarshallOfSound are you ok with the current state of the PR?

atom/common/api/electron_bindings.cc
```diff
@@ -29,6 +29,7 @@
 #include "native_mate/dictionary.h"
 #include "services/resource_coordinator/public/cpp/memory_instrumentation/global_memory_dump.h"
 #include "services/resource_coordinator/public/cpp/memory_instrumentation/memory_instrumentation.h"
+#include "third_party/blink/renderer/platform/heap/process_heap.h"  // nogncheck
```
alexeykuzmin: @nornagon 
Is it ok to use `nogncheck` to suppress gn error `Can't include this header from here.`?

miniak: we do it here as well https://github.com/electron/electron/blob/master/atom/renderer/renderer_client_base.cc#L38

miniak:
@alexeykuzmin are we going to merge this one or not?

PR #17789, opened 2019-04-15
build: disable Vulkan support by default (4-1-x)

alexeykuzmin:
#### Description of Change
Is was enabled in
https://chromium.googlesource.com/chromium/src.git/+/327326656f2c8c45ecb0ff4675e5c577e7c275ef
which landed in 69.0.3460.0

Vulkan support would require a command line parameter "--enable-vulkan"
to be passes to a binary to be enabled anyway,
so this change doesn't actual alter the current behaviour of Electron.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Removed unused Vulkan libraries from bundle.

### DIFF
--- a/build/args/all.gn
+++ b/build/args/all.gn
@@ -8,6 +8,7 @@ v8_embedder_string = "-electron.0"
 
 enable_cdm_host_verification = false
 enable_extensions = false
+enable_vulkan = false
 proprietary_codecs = true
 ffmpeg_branding = "Chrome"
 


nornagon:
Suggest release notes change:

> Remove unused Vulkan libraries from bundle.


Since we don't actually _support_ vulkan even when shipping these binaries (you mentioned it doesn't work even if you pass the flag), I wouldn't say we're disabling anything here?

alexeykuzmin:
@nornagon You just  could have updated the release notes yourself =) Done.
Any other suggestions?

PR #17802, opened 2019-04-16
fix: launch child processes in a job to ensure that they don't outlive the browser (4-1-x)

miniak:
#### Description of Change
Backport https://chromium-review.googlesource.com/c/chromium/src/+/1350977
> Fix network service not working when running as a remote app before Windows 8.
> 
> r562058 made the network process always run in a job object to try to avoid it outliving the browser process. This was added to make the network process quit sooner if there was a lot of pending I/O, which happened with layout tests. Windows before version 8 doesn't support nested job objects though, so we need to stop this mitigation for those old versions.

Backport https://chromium-review.googlesource.com/c/chromium/src/+/1548839
> Launch child unsandboxed processes in "Kill on Job Close" job.
> 
> When the --no-sandbox flag is used, hanging child processes can outlive
> a browser process that gets killed. This results in leftover processes
> on bots when the process launcher sequence is made CONTINUE_ON_SHUTDOWN
> (see https://crbug.com/830954#c18).
> 
> With this CL, all child processes are part of a "Kill on Job Close" job
> and are guaranteed not to outlive the browser process.
> 
> This is a prerequisite to make the process launcher sequence
> CONTINUE_ON_SHUTDOWN, which will fix ~12% of shutdown hangs on Windows.

#### Checklist
- [ ] `npm test` passes

#### Release Notes
Notes: Fixed non-sandboxed renderer processes on Windows not terminating in all cases (JS thread stuck in infinite loop for example) when the main process is killed.


deepak1556:
Is `fix_network_service_not_working_when_running_as_a_remote_app_before.patch` needed ? Reason I don't think its needed, the second patch is a generic version that applies to all child processes and most importantly there is no network service process in `4-1-x`

miniak:
Replacing by #18147

Issue #16091, opened 2018-12-16
[macOS]NSWindowZoomButton is disabled in frameless window since 3.0.0


Gizeta:
* Output of `node_modules/.bin/electron --version`: 3.0.11 (since 3.0.0-beta.1)
* Operating System (Platform and Version): macOS Mojave 10.14.2
* Output of `node_modules/.bin/electron --version` on last known working Electron version (if applicable): 2.0.14

**Actual behavior**
I create a frameless window and use `Accessibility Inspector` in `Xcode` to display the information properties of window's `ZoomButton`. I see that `ZoomButton` is disabled since `3.0.0-beta.1`. That breaks some window manager, such as [Magnet](http://magnet.crowdcafe.com/) which may use `Accessibility API`. If I downgrade to `2.0.14`, `Magnet` works well because `ZoomButton` is enabled.

**To Reproduce**
just create a frameless window

**Screenshots**
<img width="1680" alt="2.x" src="https://user-images.githubusercontent.com/1572561/50050527-c962c300-0138-11e9-942b-ce0f6379f0d0.png">
<img width="1680" alt="3.x" src="https://user-images.githubusercontent.com/1572561/50050797-5a896800-0140-11e9-823a-f5abbece2a37.png">

with Magnet's menu
<img width="1280" alt="2018-12-16 2 38 43" src="https://user-images.githubusercontent.com/1572561/50050802-6412d000-0140-11e9-85dd-b071638ca715.png">

**Additional Information**
`ZoomButton` will be enabled if I make the window full screen and restore it. And `Magnet` will works correctly.

codebytere:
This can't currently be worked around, as macOS itself disallows NSViews with `NSWindowStyleMaskFullSizeContentView` from entering full screen. Should something change, we'll return to this and address it.

PR #17813, opened 2019-04-16
fix: remove disable NSWindowZoomButton in frameless window

zp1996:
#### Description of Change
Fixes [16091](https://github.com/electron/electron/issues/16091)

electron `1-3-x ~ 2-1-x` [achieve](https://github.com/electron/electron/blob/2-0-x/atom/browser/native_window_mac.mm#L1047):
```objective-c
// see in NativeWindowMac constructor
{
  // ...
  InstallView();

  SetMaximizable(maximizable);
}
// InstallView set like this 
[[window_ standardWindowButton:NSWindowZoomButton] setEnabled:NO];
// then quickly SetMaximizable reset
[[window_ standardWindowButton:NSWindowZoomButton] setEnabled:maximizable];
```

Can remove disable zoom-button? If remove will support some tools like [Magnet](https://magnet.crowdcafe.com/)
#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Remove disable NSWindowZoomButton in frameless window

### DIFF
--- a/atom/browser/native_window_mac.mm
+++ b/atom/browser/native_window_mac.mm
@@ -1431,11 +1431,6 @@ void NativeWindowMac::AddContentViewLayers() {
       [[window_ standardWindowButton:NSWindowMiniaturizeButton] setHidden:YES];
       [[window_ standardWindowButton:NSWindowCloseButton] setHidden:YES];
     }
-
-    // Some third-party macOS utilities check the zoom button's enabled state to
-    // determine whether to show custom UI on hover, so we disable it here to
-    // prevent them from doing so in a frameless app window.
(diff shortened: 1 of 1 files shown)


zp1996:
> @zp1996 this line has utility for reasons stated in the comment above it, and so i don't feel comfortable removing it wholesale. We would, however, potentially consider an option to make it optionally enforcable.

yeah, but i found that it always be enabled before, should I can add option for whether disabled it (default value is disable)?

codebytere:
Unfortunately, this code is important as it relates to the fabric of our frameless window code on the whole, and so at this time this isn't something we think is in our best interest to change. We'll be sure to let you know if something changes though!

PR #17823, opened 2019-04-16
fix: skip zygote for unsandboxed ppapi processes

### LINKED ISSUE #17034: [Bug] Linux Flash Plugin not loading
* **Electron Version**: 
  * v5.0.x
* **Operating System** 
  * Linux (Ubuntu 18.10 x64)
* **Last known working Electron version**: 
  * v4.0.x

### Expected Behavior
Flash Plugin should load.

### Actual behavior
Flash Plugin does not load.

### Additional info
The Flash Player under Linux never worked in a mixed-sandbox or (full-)sandbox environment. As of v5.0.x, mixed-sandbox is forced. This prevents the plugin from loading.

### To Reproduce
[app.zip](https://github.com/electron/electron/files/2877526/app.zip)

nornagon:
#### Description of Change
Fixes #17034.

The sequence of events here was:

1. The main process starts
2. Early in the initialization sequence, before `main.js` is run, content spawns a zygote and initializes the sandbox
3. `main.js` runs, and adds `--no-sandbox` to the command-line flags
4. The renderer boots and requests a ppapi process
5. The ppapi process is forked from the zygote, which already has the sandbox pre-initialized, since it was warmed up before the `--no-sandbox` flag was added.

With this patch, we intervene before (5) and re-check the command-line flags. If `--no-sandbox` is present, we do not launch from the zygote, and instead spawn a fresh process without the sandbox.

This patch is similar to [the one that supports mixed-sandbox mode](https://github.com/electron/electron/blob/a34c47b54234237696a6630718ff74609ab7cd00/patches/common/chromium/support_mixed_sandbox_with_zygote.patch#L66).

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed an issue preventing the Flash plugin from loading on Linux.

### DIFF
--- a/patches/common/chromium/.patches
+++ b/patches/common/chromium/.patches
@@ -73,3 +73,4 @@ fix_disable_usage_of_setapplicationisdaemon_and.patch
 disable_custom_libcxx_on_windows.patch
 fix_retain_compatibility_with_msvc.patch
 disable_network_services_by_default.patch
+unsandboxed_ppapi_processes_skip_zygote.patch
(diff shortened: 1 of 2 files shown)


nornagon:
linux failure is flake; merging.

PR #17824, opened 2019-04-16
build: save HEAD when git-import-patches runs

nornagon:
#### Description of Change
This saves what HEAD is when `git-import-patches` runs so that when we later run `git-export-patches` we can remember what base to use. Should be faster & more reliable, and will work for repos that don't always have a tag/branch at HEAD when we clone them (notably: boringssl, ffmpeg, and lkgr chromium itself).

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/git-export-patches
+++ b/script/git-export-patches
@@ -9,14 +9,35 @@ import sys
 
 def guess_base_commit(repo):
   """Guess which commit the patches might be based on"""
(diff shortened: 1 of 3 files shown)


nornagon:
CI failures are flakes.

Issue #17785, opened 2019-04-15
nativeImage.createFromNamedImage returns nothing


erikjalevik:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 4.1.4
* **Operating System:**
  * MacOS 10.14.4
* **Last Known Working Electron version:**:
  * n/a

### Expected Behavior
`createFromNamedImage` to return the image name passed as a string, in this case "NSImageNameFolder".

### Actual Behavior
`createFromNamedImage` returns no image data.

### To Reproduce
Here is a repo showing the issue: 

https://github.com/erikjalevik/electron-retina-nativeimage-bugs

Instructions:
```
git clone git@github.com:erikjalevik/electron-retina-nativeimage-bugs.git
npm install
npm run start
```

The third section of the page shows the non-existent image under "nativeImage.createFromNamedImage".

### Screenshots

<img width="475" alt="Screenshot 2019-04-15 at 13 53 52" src="https://user-images.githubusercontent.com/8722592/56130703-ef349300-5f85-11e9-8c97-89e082cf1294.png">

codebytere:
Strangely, this is localized: it doesn't repro if i use, for example, `NSActionTemplate`.

PR #17826, opened 2019-04-16
docs: document occasional mismatch between NSImageName and string

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/17785.

In some cases, the `NSImageName` doesn't match its string representation; one example of this is `NSFolderImageName`, whose string representation would actually be `NSFolder`. Therefore, you'll need to determine the correct string representation for your image before passing it in. This documents that need.

cc @nornagon 

#### Checklist

#### Release Notes

Notes: Documented an issue with `nativeImage.createFromNamedImage` where the string representations of NSImageName might not match the `NSImageName`.

### DIFF
--- a/docs/api/native-image.md
+++ b/docs/api/native-image.md
@@ -198,6 +198,12 @@ The `hslShift` is applied to the image with the following rules
 This means that `[-1, 0, 1]` will make the image completely white and
 `[-1, 1, 0]` will make the image completely black.
 
+In some cases, the `NSImageName` doesn't match its string representation; one example of this is `NSFolderImageName`, whose string representation would actually be `NSFolder`. Therefore, you'll need to determine the correct string representation for your image before passing it in. This can be done with the following:
+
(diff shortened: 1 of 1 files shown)


docs/api/native-image.md
```diff
@@ -198,6 +198,12 @@ The `hslShift` is applied to the image with the following rules
 This means that `[-1, 0, 1]` will make the image completely white and
 `[-1, 1, 0]` will make the image completely black.
 
+In some cases, the `NSImageName` doesn't match its string representation; one example of this is `NSFolderImageName`, whose string representation would actually be `NSFolder`. Therefore, you'll need to determine the correct string representation for your image before passing it in. This can be done with the following:
+
+`echo -e '#import <Cocoa/Cocoa.h>\nint main() { NSLog(@"%@", SYSTEM_IMAGE_NAME); }' | clang -otest -x objective-c -framework Cocoa - && ./test`
```
MarshallOfSound: That's one heck of a one-liner 😆

nornagon: maybe also note that you should replace `SYSTEM_IMAGE_NAME` by the standard name you want to look up.

PR #17828, opened 2019-04-16
build: override the NMV for Electron 5

MarshallOfSound:
This corrects the NMV for Electron 5 now that we are updated to node 12

Notes: no-notes

### DIFF
--- a/build/args/all.gn
+++ b/build/args/all.gn
@@ -2,6 +2,7 @@ is_electron_build = true
 use_jumbo_build = true
 root_extra_deps = [ "//electron" ]
 
+# Registry of NMVs --> https://github.com/nodejs/node/pull/24114/files
 node_module_version = 70
 
 v8_promise_internal_field_count = 1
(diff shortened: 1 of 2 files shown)


build/args/all.gn
```diff
@@ -2,6 +2,8 @@ is_electron_build = true
 use_jumbo_build = true
 root_extra_deps = [ "//electron" ]
 
+node_module_version = 70
```
nornagon: please link to the source of truth there

nornagon:
see also comments on the node repo

build/args/all.gn
```diff
@@ -2,6 +2,9 @@ is_electron_build = true
 use_jumbo_build = true
 root_extra_deps = [ "//electron" ]
 
+# Registry of NMVs --> https://github.com/nodejs/node/pull/24114/files
```
nornagon:
```suggestion
# Registry of NMVs --> https://github.com/nodejs/node/blob/master/doc/abi_version_registry.json
```

PR #17838, opened 2019-04-17
feat: upgrade to Node 12 (master)

zcbenz:
#### Description of Change

Upgrade to Node 12 on master branch.

Cherry-picks changes from https://github.com/electron/electron/pull/17752 and https://github.com/electron/electron/pull/17828, node module version uses 71.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Upgrade to Node 12


build/args/all.gn
```diff
@@ -2,6 +2,9 @@ is_electron_build = true
 use_jumbo_build = true
 root_extra_deps = [ "//electron" ]
 
+# Registry of NMVs --> https://github.com/nodejs/node/pull/24114/files
+node_module_version = 71
```
deepak1556: we are on v8 7.5 in master, but this number doesn't matter in master. Maybe we should have a gclient hook that pulls the right number from that json registry if one is available, which might be useful sanity check when we cut branches,

MarshallOfSound: This number shouldn't be `71`, it is in direct conflict with node.  We should request another number in that JSON registry and set this to that number.

nornagon: Requested here: https://github.com/nodejs/node/pull/27288

Issue #17732, opened 2019-04-09
Ability to disable creation of logs directory


agjs:
Hey guys. We are keeping our logs in a different path (not using the electron's one) and we are wondering if there is some flag to disable creation of this folder.

Thanks in advance :)

agjs:
Anyone? :-)

codebytere:
@agjs yeah i can try to refactor this slightly so that users have more granular control over it.

agjs:
Would really appreciate it and also honestly, it makes sense to make this optional. Thank you

Issue #15877, opened 2018-11-29
`app.getPath('logs')` throws error on Ubuntu


jarrodek:
* Output of `node_modules/.bin/electron --version`: v3.0.10
* Operating System (Platform and Version): Ubuntu 18.04.1 LTS

**Expected Behavior**

Executing `app.getPath('logs')` returns path to the log folder of the application.

**Actual behavior**

Throws an error:

```
A JavaScript error occurred in the main process
Uncaught Exception:
Error: Failed to get 'logs' path
    at Object.<anonymous> (/home/pawel/workspace/electron-quick-start/main.js:4:21)
    at Object.<anonymous> (/home/pawel/workspace/electron-quick-start/main.js:55:3)
    at Module._compile (internal/modules/cjs/loader.js:711:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:722:10)
    at Module.load (internal/modules/cjs/loader.js:620:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:559:12)
    at Function.Module._load (internal/modules/cjs/loader.js:551:3)
    at loadApplicationPackage (/home/pawel/workspace/electron-quick-start/node_modules/electron/dist/resources/default_app.asar/main.js:287:12)
    at Object.<anonymous> (/home/pawel/workspace/electron-quick-start/node_modules/electron/dist/resources/default_app.asar/main.js:328:5)
    at Object.<anonymous> (/home/pawel/workspace/electron-quick-start/node_modules/electron/dist/resources/default_app.asar/main.js:365:3)
```

**To Reproduce**

Repo: https://github.com/jarrodek/electron-quick-start/tree/get-path-log

If you provide a URL, please list the commands required to clone/setup/run your repo e.g.
```sh
$ git clone https://github.com/jarrodek/electron-quick-start.git -b get-path-log
$ npm install
$ npm start || electron .
```

**Additional Information**

It works on macOS. I haven't tested it on windows or other Linux systems.

Issue #14470, opened 2018-09-05
Inconsistent "appData" and "logs" path names


primalanomaly:
Electron v2.0.8
Mac and Windows

As package.json does not allow spaces in the name field, my app is named "MyApp". In my build settings my productName is set to "My App" so that the user sees the name with spaces.

Using app.getPath('appData') the app data path clearly uses the "name" and saves all data in a folder named "MyApp" in my library.
Using app.getPath('logs') the log path uses the "productName" field and saves all logs in a folder named "My App". This is inconsistent, and results in multiple folders being created instead of just 1.

Using third-party log apps such as "electron-store" saves logs to the folder "MyApp" along with the appData, so it seems it is just Electron's logs path that is wrong.

```
{
  "name": "MyApp",
  "version": "0.0.1",
  "main": "./js/app.js",
  "scripts": {
    "start": "electron .",
    "build": "electron-builder --dir -mw --ia32 --x64",
    "ship": "electron-builder -mw --ia32 --x64 -p always",
    "postinstall": "electron-builder install-app-deps"
  },
  "build": {
    "appId": "com.my.app",
    "productName": "My App"
```

bpasero:
https://github.com/electron/electron/issues/14276 ?

joernroeder:
same behavior with electron 3.0.13

Issue #11858, opened 2018-02-08
getPath("logs") use the wrong "userData" path


KEVINdEsBuRgErS:
Hi, getPath("logs") return the wrong path.
getPath("logs") return : %APPDATA% + ProductName from the executable of Electron
when it should return : "userData" (when edited) + productName from package.json

The consequence is that Electron creates an empty folder "logs" in the wrong place.

It seems to me that the problem has occurred since this fix : #10911

* Electron version: 1.8.2 (since 1.8.2-beta.2)
* Operating system: Windows 7 64 bit

MarshallOfSound:
@KEVINdEsBuRgErS How are you launching your app?

KEVINdEsBuRgErS:
@MarshallOfSound My application is located in the folder /resources/app and I launch the app through Electron.exe

kspearrin:
I'm also seeing something similar. We change the log path with `app.setPath('logs', path.join(process.env.PORTABLE_EXECUTABLE_DIR, 'logs'))` as the first line in our `main.js`, yet we are still seeing an empty `logs` folder created at `%APPDATA% + ProductName.

Mic75:
Also, our ProductName contains a space in it, so now we're ending with:
`%APPDATA%\Product Name\logs` and `%APPDATA%\ProductName` (no space).

Note that the second path is what is returned by `getPath('userData')`

codebytere:
i'll resolve this shortly

codebytere:
As a quick update, diagnosing this problem more thoroughly told me some light refactoring may be in order so this is absolutely still on my active radar but it may take slightly more time :)

KEVINdEsBuRgErS:
@codebytere Take your time, I just didn't want this issue to be forgotten forever !

KEVINdEsBuRgErS:
Electron v3 is coming and no news for this issue :(

pronebird:
Sorry if I repeat the issue, but just to give a little info for debugging, tested Electron 2.0.2 and `logs` option does not respect `productName`, returns `Electron` instead of `Product Name` when running in development mode.

```
userData:  /Users/pronebird/Library/Application Support/Product Name
logs:  /Users/pronebird/Library/Logs/Electron
```

tbenyon:
@codebytere any news on this fix? I'm having the same problem as @kspearrin: 
[...]

PR #17841, opened 2019-04-17
refactor: make app logs dir creation opt-in

codebytere:
#### Description of Change

BREAKING CHANGE

Resolves https://github.com/electron/electron/issues/17732.
Resolves https://github.com/electron/electron/issues/11858.
Resolves https://github.com/electron/electron/issues/15877.
Resolves https://github.com/electron/electron/issues/14470.

Previously, we were creating the app logs folder at a predefined location during initial electron startup, which meant that it had to be manually removed and prevented clean app portability. This refactors that implementation such that it's now an opt-in feature and developers must call `app.setAppLogsPath()` in order to set this directory.

Alternatively, this could be done such that app log dir creation is still enabled by default, but can be _disabled_ at the developers discretion although this would be notably more challenging. I'd love some feedback on which approach seems optimal!

cc @zcbenz @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Made app log directory creation opt-in with a new function `app.setAppLogsPath`.

### DIFF
--- a/atom/browser/api/atom_api_app.cc
+++ b/atom/browser/api/atom_api_app.cc
@@ -16,6 +16,7 @@
 #include "atom/browser/atom_paths.h"
 #include "atom/browser/login_handler.h"
 #include "atom/browser/relauncher.h"
+#include "atom/common/application_info.h"
 #include "atom/common/atom_command_line.h"
 #include "atom/common/native_mate_converters/callback.h"
 #include "atom/common/native_mate_converters/file_path_converter.h"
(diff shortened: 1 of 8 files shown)


atom/browser/api/atom_api_app.cc
```diff
@@ -830,6 +831,17 @@ void App::SetAppPath(const base::FilePath& app_path) {
   app_path_ = app_path;
 }
 
+#if !defined(OS_MACOSX)
+void App::SetAppLogsPath() {
```
deepak1556: Wouldn't it be better to allow users provide a custom logs path as an argument rather than expecting them to change it with `app.setPath` ?

codebytere: @deepak1556 what about if we let this be the default and let them pass an argument optionally?

deepak1556: Yup I am good with that,

codebytere: done!

atom/browser/api/atom_api_app.cc
```diff
@@ -830,6 +831,26 @@ void App::SetAppPath(const base::FilePath& app_path) {
   app_path_ = app_path;
 }
 
+#if !defined(OS_MACOSX)
+void App::SetAppLogsPath(mate::Arguments* args) {
+  std::string custom_path;
```
deepak1556: `base::FilePath` converter should work here.

vladimiry:
Is it going to land in v5 any time soon? v5.0.0 didn't get this change.

MarshallOfSound:
@vladimiry This PR is tagged as `semver/major` and therefore won't go out until a major release happens (the next one being 6.0)

atom/browser/api/atom_api_app.cc
Gam3356: Thank you guys!

PR #17847, opened 2019-04-17
[WIP] chore: add missing capabilities to run cypress tests in Electron

MarshallOfSound:
#### What?

This does two things

1. Adds `ELECTRON_UNSAFELY_ALLOW_NODE_IN_CHILD_WINDOWS` as an undocumented environment variable to override the memory leak protection when opening child windows with `node` enabled.  For testing purposes we **need** to be able to do that.  It is deliberately left undocumented as using it causes memory leaks.
2. Adds support for `chrome.tabs.query` as the Cypress helper extension requires this API to function correctly.

#### Why?

We are currently working towards running Cypress powered browser tests in a wrapping Electron application.  These two changes allow us to do so with minimal architectural changes to either the app or Cypress.

Notes: no-notes

### DIFF
--- a/atom/renderer/atom_renderer_client.cc
+++ b/atom/renderer/atom_renderer_client.cc
@@ -70,19 +70,28 @@ void AtomRendererClient::DidCreateScriptContext(
     content::RenderFrame* render_frame) {
   RendererClientBase::DidCreateScriptContext(context, render_frame);
 
+  auto env = base::Environment::Create();
+  // NOTE: This should never be used in production, it is designed to be used
+  // for testing purposes
+  bool unsafely_allow_node_in_child_windows =
+      env->HasVar("ELECTRON_UNSAFELY_ALLOW_NODE_IN_CHILD_WINDOWS");
+
   // TODO(zcbenz): Do not create Node environment if node integration is not
   // enabled.
 
   // Do not load node if we're aren't a main frame or a devtools extension
   // unless node support has been explicitly enabled for sub frames
-  bool is_main_frame =
-      render_frame->IsMainFrame() && !render_frame->GetWebFrame()->Opener();
+  bool is_child_window = render_frame->GetWebFrame()->Opener();
+  bool is_main_frame = render_frame->IsMainFrame();
+  bool should_load_node_into_frame =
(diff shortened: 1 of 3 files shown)


atom/renderer/atom_renderer_client.cc
```diff
@@ -70,19 +70,28 @@ void AtomRendererClient::DidCreateScriptContext(
     content::RenderFrame* render_frame) {
   RendererClientBase::DidCreateScriptContext(context, render_frame);
 
+  auto env = base::Environment::Create();
+  // NOTE: This should never be used in production, it is designed to be used
+  // for testing purposes
```
nornagon: Note that this is intentionally undocumented, or someone will come along and document it.

atom/renderer/atom_renderer_client.cc
```diff
@@ -70,19 +70,28 @@ void AtomRendererClient::DidCreateScriptContext(
...
+
   // TODO(zcbenz): Do not create Node environment if node integration is not
   // enabled.
 
   // Do not load node if we're aren't a main frame or a devtools extension
   // unless node support has been explicitly enabled for sub frames
-  bool is_main_frame =
-      render_frame->IsMainFrame() && !render_frame->GetWebFrame()->Opener();
+  bool is_child_window = render_frame->GetWebFrame()->Opener();
+  bool is_main_frame = render_frame->IsMainFrame();
```
nornagon: this doesn't really need its own variable, since the function is named the same thing as the variable & is only used once

atom/renderer/atom_renderer_client.cc
```diff
@@ -70,19 +70,28 @@ void AtomRendererClient::DidCreateScriptContext(
...
+  bool should_load_node_into_frame =
```
nornagon: this is a confusing variable name, maybe instead:

```cpp
bool is_root_window = !render_frame->GetWebFrame()->Opener();
bool allow_node_in_window =
    render_frame->IsMainFrame() &&
    (is_root_window || unsafely_allow_node_in_child_windows);
```

codebytere:
Closing with consent from @MarshallOfSound

PR #17865, opened 2019-04-18
docs: move organizational info docs to new directory

codebytere:
#### Description of Change

This moves our important ecosystem documentation for consumers into a bespoke directory. Please feel free to comment if you think more documents should be moved here or it should be organized differently. This will also require changes to the website, so i'd love to hear ya'lls thoughts!

/cc @electron/wg-docs-tools @electron/wg-website 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Moved some administrative documentation to a clearer location.


codebytere:
@malept agree, i spent 15 minutes just struggling to find a new name 😅 

totally open to suggestions!

sofianguy:
@codebytere @malept 🤔what do you think of these?
Informational
Internal
Maintainers

So far I think organizational is a good one to begin with if we cannot think of a better term.

malept:
I think leaving them "top level" is fine for now.

Issue #16214, opened 2018-12-29
createCipheriv crash renderer window


githoniel:
### Version Info
electron: 4.0.0
OS: Windows 10   x64 10.0.17134.472

### Expected Behavior 
```js
require('crypto').createCipheriv('rc4', 'asdasd', null)
require('crypto').createCipheriv('rc4', 'asdasd', '')
```
 work well as  node v10.15.0 

### Actual behavior
```js
require('crypto').createCipheriv('rc4', 'asdasd', null)
```
cause renderer process crash 
```js
require('crypto').createCipheriv('rc4', 'asdasd', '')
```
throw 
```
Invalid IV length
    at Cipheriv.createCipherBase (internal/crypto/cipher.js:79)
    at Cipheriv.createCipherWithIV (internal/crypto/cipher.js:119)
    at new Cipheriv (internal/crypto/cipher.js:225)
    at Object.createCipheriv (crypto.js:105)
    at <anonymous>:1:19
createCipherBase @ internal/crypto/cipher.js:79
createCipherWithIV @ internal/crypto/cipher.js:119
Cipheriv @ internal/crypto/cipher.js:225
createCipheriv @ crypto.js:105
```
### To Reproduce

exec code above in devTools

shiftkey:
Likely the same root cause as #16195

orangeChu:
Before it is repaired, you may be able to use the method of the main process.:)

nornagon:
Duplicate of #16195

PR #17873, opened 2019-04-19
fix: add a compatibility EVP_CIPH_OCB_MODE value (#16214).

TaikiAkita:
Backport of [google/boringssl@4b9683](https://boringssl.googlesource.com/boringssl/+/4b968339e3ced2d498f4182cd725243bb6cca81b). This patch replaces the no-op modes with negative numbers rather than zero. Stream ciphers like RC4 report a "mode" of zero, so code comparing the mode to a dummy value will get confused.

This patch fixed issue #16214.

#### Release Notes

Notes: Fixed the "rc4" cipher (#16214) (4.1.x).


patches/common/boringssl/.patches
```diff
@@ -1,3 +1,4 @@
+compatibility_evp_ciph_ocb_mode.patch
```
MarshallOfSound: Was this patch added manually?  I'm not sure how this could have been added to the top of the file if the `export-patches` script was used

TaikiAkita: This patch was added manually. Shall I re-add it with the `export-patches` command?

TaikiAkita: > Was this patch added manually? I'm not sure how this could have been added to the top of the file if the `export-patches` script was used

This patch was added manually. Shall I re-add it with the export-patches command?

MarshallOfSound: @TaikiAkita Yep, the flow for making these patches is like this

* Run `gclient sync --your --flags --here`
* Add your commit on top of the checked out `boringssl`
* Run `git-export-patches -o electron/patches/common/boringssl`

👍

TaikiAkita: @MarshallOfSound 

I regenerated the patches with `git-export-patches`, please have a look.

TaikiAkita:
@MarshallOfSound

I regenerated the patches with git-export-patches, please have a look.

TaikiAkita:
> This patch is currently flagged as an invalid backport because it is targeting 4.1.
> 
> Can you confirm that this fix is not needed in current 5.0 or master branches?

This patch is exactly backported from 5.x. The upstream BoringSSL of 5.x branch has this patch already. So this patch is *NOT* needed in 5.x.

Test with following code:
```
var cipher = require("crypto").createCipher("rc4", "hello");
```

In 4.x, an error will occur.
In 5.x, run successfully.

TaikiAkita:
@codebytere 
Now it is ready to be merged. Please help to merge this PR.

TaikiAkita:
@codebytere 
Please merge this PR.

shiftkey:
@TaikiAkita please address the failing commit status: https://github.com/electron/electron/pull/17873/checks?check_run_id=105806788

TaikiAkita:
> > This patch is currently flagged as an invalid backport because it is targeting 4.1.
> > Can you confirm that this fix is not needed in current 5.0 or master branches?
> 
> @MarshallOfSound
> 
> This patch is exactly backported from 5.x. The upstream BoringSSL of 5.x branch has this patch already. So this patch is _NOT_ needed in 5.x.
> 
> Test with following code:
> 
> ```
> var cipher = require("crypto").createCipher("rc4", "hello");
> ```
> In 4.x, an error will occur.
> In 5.x, run successfully.
> 
> BTW. This patch should also be backported to 4.0.x branch.

@shiftkey 
Described above. This patch is derived from upstream, NOT from other PR.

Issue #17845, opened 2019-04-17
getGPUInfo failing on 32bit Windows since Node 12 upgrade


MarshallOfSound:
```
not ok 61 app module getGPUInfo() API succeeds with complete GPUInfo
      at getGPUInfo (C:\projects\src\electron\spec\api-app-spec.js:969:31)
      at processTicksAndRejections (internal/process/task_queues.js:86:5)
      at async Context.<anonymous> (C:\projects\src\electron\spec\api-app-spec.js:991:28)
```

Appears to be a regression caused by https://github.com/electron/electron/commit/ffd2224bbe818a12596a9db07bace2b364dcbd41

MarshallOfSound:
Fixed by #17877

PR #17877, opened 2019-04-19
fix: drain tasks before shutting down isolate (5-0-x)

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/17845.

According to the crash trace, it happened because there are pending tasks in queue when shutting down, so draining tasks before shutdown can fix it.

I'm not sure why it only happened for x86 Windows, and why it does not happen in https://github.com/electron/electron/pull/17838 on master branch which also has Node 12.

Anyway this fixes the problem without bad side effect as I see, and we can also put it in master branch if needed.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


zcbenz:
> I'd prefer if this landed in master and then was backported. Just because the issue doesn't seem to appear on master doesn't mean it won't be triggered by some obscure condition so landing this in master makes sense IMO 👍

Sounds good to me, I have created https://github.com/electron/electron/pull/17879.

PR #17894, opened 2019-04-21
chore: remove deprecated app.enableMixedSandbox()

miniak:
#### Description of Change
Follow up to https://github.com/electron/electron/pull/15894

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Removed deprecated `app.enableMixedSandbox()` API.

/cc @nornagon

### DIFF
--- a/docs/api/breaking-changes.md
+++ b/docs/api/breaking-changes.md
@@ -81,6 +81,15 @@ powerMonitor.querySystemIdleTime(callback)
 const idleTime = getSystemIdleTime()
 ```
 
+## `app.enableMixedSandbox`
+
+```js
+// Deprecated
+app.enableMixedSandbox()
+```
+
+Mixed-sandbox mode is now enabled by default.
+
 ## Preload scripts outside of app path are not allowed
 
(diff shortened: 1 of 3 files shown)


docs/api/breaking-changes.md
```diff
@@ -81,6 +81,15 @@ powerMonitor.querySystemIdleTime(callback)
...
 ```
 
+## `app.enableMixedSandbox`
+
+```js
+// Deprecated
+app.enableMixedSandbox()
+```
+
+Mixed-sandbox mode is now enabled by default.
```
alexeykuzmin: Is there a way to disable it? Should it be mentioned here?

miniak: it can be disabled by running Electron with `--no-sandbox`, @nornagon should we mention it? Also should this option be listed in `docs/api/chrome-command-line-switches.md`?

miniak: We also have the environment variable added in https://github.com/electron/electron/pull/16576, which seems to be deliberately undocumented

PR #17308, opened 2019-03-09
feat: only allow bundled preload scripts

miniak:
#### Description of Change
For security reasons, preload scripts can only be loaded from the app path (`app.getAppPath()`).

#### Checklist
- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes
Notes: For security reasons, preload scripts can now only be loaded from a subpath of app path (`app.getAppPath()`).

/cc @electron/wg-security

PR #17896, opened 2019-04-21
feat: add support for relative preload script paths (resolved against app.getAppPath())

miniak:
#### Description of Change
Allow relative preload script paths, resolved against `app.getAppPath()` similar to `WebContents.prototype.loadFile()`.

Previously, only the main preload script was enforced to be an absolute path, while the session preload scripts could be relative (with potentially inconsistent behavior between sandboxed and non-sandboxed renderers).

Follow up to #17308.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added support for relative preload script paths (resolved against `app.getAppPath()`).

### DIFF
--- a/atom/browser/web_contents_preferences.cc
+++ b/atom/browser/web_contents_preferences.cc
@@ -218,12 +218,8 @@ bool WebContentsPreferences::GetPreloadPath(
   DCHECK(path);
   base::FilePath::StringType preload;
   if (GetAsString(&preference_, options::kPreloadScript, &preload)) {
-    if (base::FilePath(preload).IsAbsolute()) {
-      *path = std::move(preload);
-      return true;
-    } else {
-      LOG(ERROR) << "preload script must have absolute path.";
-    }
+    *path = std::move(preload);
+    return true;
   } else if (GetAsString(&preference_, options::kPreloadURL, &preload)) {
     // Translate to file path if there is "preload-url" option.
     base::FilePath preload_path;
--- a/docs/api/sandbox-option.md
+++ b/docs/api/sandbox-option.md
@@ -77,7 +77,7 @@ app.on('ready', () => {
   win = new BrowserWindow({
     webPreferences: {
       sandbox: true,
-      preload: path.join(app.getAppPath(), 'preload.js')
+      preload: 'preload.js'
     }
   })
   win.loadURL('http://google.com')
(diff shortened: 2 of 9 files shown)


miniak:
@MarshallOfSound, @nornagon can you please have a look?

zcbenz:
One problem with relative preload script path is, it is hard to determine how the path is resolved by looking at the code. It is natural to think the path is relative to current file (like how `require` resolves paths), and the `preload` attribute of `<webview>` tag resolves paths to current page.

MarshallOfSound:
I agree with @zcbenz and @codebytere above, relative paths are inherently confusing in this case where the relative path is provided in one file / process and resolved in another file / process.  The absolute path is good for user readability and for our API being easily understandably / explainable.

miniak:
I surrender then

miniak:
@zcbenz but doesn't `webContents.loadFile()` have the same issue?
IMHO being able to use `preload: 'renderer/preload.js'` instead of `preload: path.join(app.getAppPath(), 'renderer/preload.js')` is way more convenient. The application has a root directory, so that paths should be relative to that.

miniak:
I think clearly specifying this in the documentation would help avoiding any confusion, as we already do for `loadFile`

```
Loads the given file in the window, `filePath` should be a path to
an HTML file relative to the root of your application.  For instance
an app structure like this:

```sh
| root
| - package.json
| - src
|   - main.js
|   - index.html

win.loadFile('src/index.html')
```

miniak:
```
| root
| - package.json
| - src
|   - main.js
|   - index.html
|   - preload.js

preload: 'src/preload.js'
```

miniak:
@MarshallOfSound the fact that the relative path is provided in one file / process and resolved in another file / process is just an implementation detail. if the application root is the base path, then there is no confusion IMO.

zcbenz:
> @zcbenz but doesn't `webContents.loadFile()` have the same issue?

I'm neutral on adding this API, I'm good with it as long as the behavior is documented.
(I don't really like the `loadFile` API.)

alitaheri:
This is a painful addition for development. I get the benefit for production though :+1:

we should at least disable this check during development. Right now  I have to copy the preload to where the mainProcess is in order to be able to run in. since you're using `realpath` symbolic links won't work, hard links won't do as the reload is not modified but rather re-created. so please at least add a mechanism to disable this check for dev mode (or just check some environment variable)

Thank you

PR #17902, opened 2019-04-22
feat: add app.enableSecureMode() API

miniak:
#### Description of Change
This is an opt-in locked down mode designed for apps loading remote content.

Check [docs/api/secure-mode.md](https://github.com/electron/electron/blob/miniak/secure-mode/docs/api/secure-mode.md) for details.

#### Checklist
- [ ] `npm test` passes

#### Release Notes
Notes: Added `app.enableSecureMode()` API to opt-in into more secure mode designed for apps running remote content.

/cc @electron/wg-security

### DIFF
--- a/atom/browser/atom_permission_manager.cc
+++ b/atom/browser/atom_permission_manager.cc
@@ -9,6 +9,7 @@
 
 #include "atom/browser/atom_browser_client.h"
 #include "atom/browser/atom_browser_main_parts.h"
+#include "atom/browser/browser.h"
 #include "atom/browser/web_contents_preferences.h"
 #include "content/public/browser/child_process_security_policy.h"
 #include "content/public/browser/permission_controller.h"
@@ -155,9 +156,15 @@ int AtomPermissionManager::RequestPermissionsWithDetails(
     return content::PermissionController::kNoPendingOperation;
   }
 
+  bool secure_mode_enabled = Browser::Get()->secure_mode().has_value();
+
   if (request_handler_.is_null()) {
     std::vector<blink::mojom::PermissionStatus> statuses;
     for (auto permission : permissions) {
+      if (secure_mode_enabled) {
+        statuses.push_back(blink::mojom::PermissionStatus::DENIED);
+        continue;
+      }
       if (permission == content::PermissionType::MIDI_SYSEX) {
         content::ChildProcessSecurityPolicy::GetInstance()
             ->GrantSendMidiSysExMessage(
@@ -237,6 +244,9 @@ bool AtomPermissionManager::CheckPermissionWithDetails(
     const GURL& requesting_origin,
     const base::DictionaryValue* details) const {
   if (check_handler_.is_null()) {
+    if (Browser::Get()->secure_mode()) {
+      return false;
+    }
     return true;
   }
   auto* web_contents =
--- a/atom/browser/browser.cc
+++ b/atom/browser/browser.cc
@@ -249,4 +249,8 @@ void Browser::NewWindowForTab() {
 }
 #endif
 
+void Browser::EnableSecureMode(const SecureModeOptions& options) {
+  secure_mode_ = options;
+}
+
 }  // namespace atom
--- a/atom/browser/browser.h
+++ b/atom/browser/browser.h
@@ -15,6 +15,7 @@
 #include "base/compiler_specific.h"
 #include "base/macros.h"
 #include "base/observer_list.h"
+#include "base/optional.h"
 #include "base/strings/string16.h"
 #include "base/values.h"
 #include "native_mate/arguments.h"
@@ -262,6 +263,16 @@ class Browser : public WindowListObserver {
   bool is_ready() const { return is_ready_; }
   const util::Promise& WhenReady(v8::Isolate* isolate);
 
+  struct SecureModeOptions {
+    bool configurable_sandbox = false;
+    bool configurable_context_isolation = false;
+    bool configurable_native_window_open = false;
+    bool configurable_remote_module = false;
+  };
+
+  base::Optional<SecureModeOptions> secure_mode() const { return secure_mode_; }
+  void EnableSecureMode(const SecureModeOptions& options);
+
  protected:
   // Returns the version of application bundle or executable file.
   std::string GetExecutableFileVersion() const;
@@ -293,6 +304,9 @@ class Browser : public WindowListObserver {
 
   // The browser is being shutdown.
   bool is_shutdown_ = false;
+
+  // Secure mode options.
+  base::Optional<SecureModeOptions> secure_mode_;
 
   // Null until/unless the default main message loop is running.
   base::OnceClosure quit_main_message_loop_;
--- a/atom/browser/web_contents_preferences.cc
+++ b/atom/browser/web_contents_preferences.cc
@@ -9,6 +9,7 @@
 #include <utility>
 #include <vector>
 
+#include "atom/browser/browser.h"
 #include "atom/browser/native_window.h"
 #include "atom/browser/web_view_manager.h"
 #include "atom/common/native_mate_converters/value_converter.h"
@@ -116,6 +117,8 @@ WebContentsPreferences::WebContentsPreferences(
 
   instances_.push_back(this);
 
+  const auto& secure_mode = Browser::Get()->secure_mode();
+
   // Set WebPreferences defaults onto the JS object
   SetDefaultBoolIfUndefined(options::kPlugins, false);
   SetDefaultBoolIfUndefined(options::kExperimentalFeatures, false);
@@ -124,10 +127,12 @@ WebContentsPreferences::WebContentsPreferences(
   SetDefaultBoolIfUndefined(options::kNodeIntegrationInWorker, false);
   SetDefaultBoolIfUndefined(options::kDisableHtmlFullscreenWindowResize, false);
   SetDefaultBoolIfUndefined(options::kWebviewTag, false);
-  SetDefaultBoolIfUndefined(options::kSandbox, false);
-  SetDefaultBoolIfUndefined(options::kNativeWindowOpen, false);
-  SetDefaultBoolIfUndefined(options::kEnableRemoteModule, true);
-  SetDefaultBoolIfUndefined(options::kContextIsolation, false);
+  if (!secure_mode) {
+    SetDefaultBoolIfUndefined(options::kSandbox, false);
+    SetDefaultBoolIfUndefined(options::kNativeWindowOpen, false);
+    SetDefaultBoolIfUndefined(options::kEnableRemoteModule, true);
+    SetDefaultBoolIfUndefined(options::kContextIsolation, false);
+  }
   SetDefaultBoolIfUndefined(options::kJavaScript, true);
   SetDefaultBoolIfUndefined(options::kImages, true);
   SetDefaultBoolIfUndefined(options::kTextAreasAreResizable, true);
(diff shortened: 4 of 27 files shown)


docs/api/app.md
```diff
@@ -1252,6 +1252,26 @@ Returns `String` - The command-line switch value.
...
+- `nodeIntegration`
+- `nodeIntegrationInWorker`
+
+When the following events are unhandled, `event.preventDefault()` is called implicitly:
+- `new-window`
+- `will-attach-webview`
+
+When no handler is set using `session.setPermissionRequestHandler()`, it behaves as if `callback(false)` was called.
+
+This method can only be called before app is ready.
```
MarshallOfSound: This will involve a race condition, is there any technical reason why this is the case.  I would imagine it doesn't matter when this is called?

typings/internal-electron.d.ts
```diff
@@ -13,6 +13,7 @@ declare namespace Electron {
   }
 
   interface App {
+    _isSecureModeEnabled(): boolean;
```
MarshallOfSound: I think this should be exposed as real API to allow third party Electron module to know if secure mode is enabled

docs/api/app.md
```diff
@@ -1252,6 +1252,26 @@ Returns `String` - The command-line switch value.
 
 **Note:** When the switch is not present or has no value, it returns empty string.
 
+### `app.enableSecureMode()`
+
+Changes the default values of following `webPreferences`:
```
MarshallOfSound: In anticipation of this API doing more things in the future this should probably be moved to `api/secure-mode.md` or similar so we can add detail / more things to it without making the app documentation page huge.

atom/browser/web_contents_preferences.cc
```diff
@@ -124,10 +127,10 @@ WebContentsPreferences::WebContentsPreferences(
   SetDefaultBoolIfUndefined(options::kNodeIntegrationInWorker, false);
   SetDefaultBoolIfUndefined(options::kDisableHtmlFullscreenWindowResize, false);
   SetDefaultBoolIfUndefined(options::kWebviewTag, false);
-  SetDefaultBoolIfUndefined(options::kSandbox, false);
-  SetDefaultBoolIfUndefined(options::kNativeWindowOpen, false);
-  SetDefaultBoolIfUndefined(options::kEnableRemoteModule, true);
-  SetDefaultBoolIfUndefined(options::kContextIsolation, false);
+  SetDefaultBoolIfUndefined(options::kSandbox, secure_mode_enabled);
```
MarshallOfSound: I think these defaults should remain the same and we should add a new block on line 152 like below

```cpp
if (secure_mode_enabled) {
  SetBool(options:kSandbox, true);
}
```

To make the current defaults explicit and the behavior of secure mode explicit

atom/browser/browser.h
```diff
@@ -293,6 +296,9 @@ class Browser : public WindowListObserver {
   // The browser is being shutdown.
   bool is_shutdown_ = false;
 
+  // TODO
```
MarshallOfSound: TODO?

atom/browser/browser.h
```diff
@@ -261,6 +261,9 @@ class Browser : public WindowListObserver {
   bool is_ready() const { return is_ready_; }
   const util::Promise& WhenReady(v8::Isolate* isolate);
 
+  bool secure_mode_enabled() const { return secure_mode_enabled_; }
+  void EnableSecureMode(mate::Arguments* args);
```
MarshallOfSound: `atom::Browser` shouldn't receive `args` IMO.  The arg parsing should be done in `atom::api::App` and this should take no arguments

atom/browser/browser.cc
```diff
@@ -249,4 +249,15 @@ void Browser::NewWindowForTab() {
 }
 #endif
 
+void Browser::EnableSecureMode(mate::Arguments* args) {
+  if (is_ready()) {
+    args->ThrowError(
```
MarshallOfSound: As mentioned below this check can and should be done in `atom::api::App` IMO

atom/browser/atom_permission_manager.cc
```diff
@@ -155,9 +156,15 @@ int AtomPermissionManager::RequestPermissionsWithDetails(
     return content::PermissionController::kNoPendingOperation;
   }
 
+  bool secure_mode_enabled = Browser::Get()->secure_mode_enabled();
+
   if (request_handler_.is_null()) {
     std::vector<blink::mojom::PermissionStatus> statuses;
     for (auto permission : permissions) {
+      if (secure_mode_enabled) {
+        statuses.push_back(blink::mojom::PermissionStatus::DENIED);
```
MarshallOfSound: There are also permission checks (not requests) that we should `return false` for when secure mode is enabled to be consistent here

atom/browser/browser.h
miniak: I wasn't sure what to put there, `Whether secure mode is enabled.` is not that useful

docs/api/app.md
miniak: I was thinking about that, makes sense. I will do it.

typings/internal-electron.d.ts
miniak: should I make it a readonly property?

miniak: is it possible to make a property `readonly` in `electron.d.ts`?

docs/api/app.md
miniak: it probably does not matter, but it seems more future-proof to make this limitation? it should be enabled before any windows are created.

miniak:
@MarshallOfSound should I mark it `_Experimental_`?

miniak:
@MarshallOfSound I've added `remote` and `desktopCapturer` events as well.

docs/api/app.md
MarshallOfSound: Let's make that the restriction then instead of before app ready.  Enforce that it must be called before any webcontents are made.

miniak: my concern is that people are used to the limitation that APIs must / cannot be called before / after `app.isReady`. do we have any API which can only be used before webcontents are created?

atom/browser/atom_permission_manager.cc
```diff
@@ -237,6 +244,9 @@ bool AtomPermissionManager::CheckPermissionWithDetails(
     const GURL& requesting_origin,
     const base::DictionaryValue* details) const {
   if (check_handler_.is_null()) {
+    if (Browser::Get()->secure_mode_enabled()) {
+      return false;
+    }
```
ckerr: _(minor)_ 247-250 could be rephrased as `return Browser::Get()->secure_mode_enabled();`

docs/api/secure-mode.md
```diff
@@ -0,0 +1,72 @@
...
+Request handlers must be used to explicitly grant the permissions (potentially asking the user for consent):
+```js
+session.setPermissionCheckHandler((webContents, permission) => {
+  return true
+})
+
+session.setPermissionRequestHandler((webContents, permission, callback) => {
+  callback(true)
+})
+```
```
ckerr: These doc changes are very helpful! Thank you for writing this so clearly.

docs/api/app.md
ckerr: Dumb question, would it make sense to make this a `webPreferences` argument instead? That would prevent the race condition and also resolve the question of when the API call was allowed.

miniak: btw where is the race condition?

atom/browser/atom_permission_manager.cc
miniak: @MarshallOfSound what do you think? I wanted to make an explicit fork in the code for the secure mode to make the behavior clearly visible.

docs/api/app.md
MarshallOfSound: @miniak It is very easy to _delay_ a task (run something after an event)
There is no easy way to guarantee to do something before an event.

If there is not a technical limitation here, I would suggest we tie this to the actual thing we care about not some other arbitrary event.

miniak: Ok, I will change the condition. Is checking if there are 0 webcontents instances alive at the time the API is called good enough check?

MarshallOfSound: `webContents().getAll().length === 0` sounds like it will suffice for the actual condition you are trying to enforce here.  (That this API is called before any windows / web contents are made)

atom/browser/web_contents_preferences.cc
```diff
@@ -147,6 +144,18 @@ WebContentsPreferences::WebContentsPreferences(
 #endif
   SetDefaultBoolIfUndefined(options::kOffscreen, false);
 
+  if (Browser::Get()->secure_mode_enabled()) {
```
MarshallOfSound: This might have gotten lost in the original comment but I meant that we should always do the `SetDefaultBoolIfUndefined` calls for the normal defaults.  And if `secure_mode_enabled()` then override the values regardless of whether they are undefined.  E.g.

```cpp
SetBoolIfUndefined(optionA, true);
if (is_secure_mode) {
  SetBool(optionA, false);
}
```

Then we can remove the `if (!secure_mode_enabled && IsEnabled(options::kNodeIntegration))` style code below (we won't need to check `secure_mode_enabled`)

atom/browser/atom_permission_manager.cc
MarshallOfSound: As I alluded to below in a different comment for this particular API I would greatly appreciate the explicitness this code fork brings.  It makes it really clear what is going to happen when secure mode is enabled.

docs/api/app.md
miniak: @MarshallOfSound I've implemented the check in native code, can you check that? if you don't like the approach, I can do it in JS instead

atom/browser/web_contents_preferences.cc
miniak: some options need to stay overridable. I only want to kill `nodeIntegration` and `nodeIntegrationInWorker` as those are not necessary to keep functionality as you can use preload script instead.

miniak: or maybe it should be configurable, which options are going to be disabled completely.

miniak: something like?
```
interface OptionsToDisable {
  sandbox?: boolean;
  contextIsolation?: boolean;
  nativeWindowOpen?: boolean;
  enableRemoteModule?: boolean;
  nodeIntegration?: boolean;
  nodeIntegrationInWorker?: boolean;
}

app.enableSecureMode(optionsToDisable: OptionsToDisable): void;
```

miniak: options will always default to the more secure value and if specified in this list, they will also not be overridable.

miniak: e.g.
```ts
app.enableSecureMode({
    contextIsolation: true // allow disabling contextIsolation
});
```

miniak: or some other variant of this, to make the API as least confusing as possible

MarshallOfSound: > some options need to stay overridable.

I disagree, if the options can be overridable then any of the previous vulns we've had where people have been able to set `nodeIntegration: true` (somehow) would still work making this option not very helpful.

I think this option should give no options and violently enforce the security guidelines (disable node, enable isolation, reject permission requests if not already handled, etc..)

That's just IMO though, would like some weigh in from @electron/wg-security

miniak: Well, `contextIsolation` for example is not an option for us for now (we'll need to do a lot of refactoring to get there). We need it to be optionally overridable. The default can be that nothing is overridable and you have to opt-in explicitly into being able to override some options.

nornagon:
Closing this due to discussion in @electron/wg-security meeting. This should be implemented as granular, descriptive APIs, not as one "secure mode" flag.

PR #17907, opened 2019-04-23
chore: remove promisification deprecation callbacks

codebytere:
#### Description of Change

Remove promisification deprecations and tests verifying callback versions. I think i got everything but please let me know if i missed any of them!

cc @miniak @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Removed deprecations for asynchronous functions to force use of promises.

### DIFF
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -118,7 +118,7 @@ Returns `Promise<void>` - resolves when the storage data has been cleared.
 
 #### `ses.flushStorageData()`
 
-Writes any unwritten DOMStorage data to disk.s
+Writes any unwritten DOMStorage data to disk.
 
 #### `ses.setProxy(config)`
 
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -93,50 +93,15 @@ session.defaultSession.on('will-download', (event, item, webContents) => {
 
 The following methods are available on instances of `Session`:
 
-#### `ses.getCacheSize(callback)`
-
-* `callback` Function
-  * `size` Integer - Cache size used in bytes.
-  * `error` Integer - The error code corresponding to the failure.
-
-Callback is invoked with the session's current cache size.
-
-**[Deprecated Soon](modernization/promisification.md)**
-
 #### `ses.getCacheSize()`
 
 Returns `Promise<Integer>` - the session's current cache size, in bytes.
-
-#### `ses.clearCache(callback)`
-
-* `callback` Function - Called when operation is done.
-  * `error` Integer - The error code corresponding to the failure.
-
-Clears the session’s HTTP cache.
-
-**[Deprecated Soon](modernization/promisification.md)**
 
 #### `ses.clearCache()`
 
 Returns `Promise<void>` - resolves when the cache clear operation is complete.
 
 Clears the session’s HTTP cache.
-
-#### `ses.clearStorageData([options,] callback)`
-
-* `options` Object (optional)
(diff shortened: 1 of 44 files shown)


miniak:
this is missing:
```
diff --git a/lib/browser/api/app.ts b/lib/browser/api/app.ts
index 0d6110e4c..625581e1b 100644
--- a/lib/browser/api/app.ts
+++ b/lib/browser/api/app.ts
@@ -89,9 +89,6 @@ for (const name of events) {
   })
 }
 
-// Function Deprecations
-app.getFileIcon = deprecate.promisify(app.getFileIcon)
-
 // Property Deprecations
 deprecate.fnToProperty(app, 'accessibilitySupportEnabled', '_isAccessibilitySupportEnabled', '_setAccessibilitySupportEnabled')
 
diff --git a/lib/renderer/api/web-frame.ts b/lib/renderer/api/web-frame.ts
index ca640c30a..4d3b06489 100644
--- a/lib/renderer/api/web-frame.ts
+++ b/lib/renderer/api/web-frame.ts
@@ -85,15 +85,6 @@ function getWebFrame (context: Window) {
   return context ? new WebFrame(context) : null
 }
 
-const promisifiedMethods = new Set<string>([
-  'executeJavaScript',
-  'executeJavaScriptInIsolatedWorld'
-])
-
-for (const method of promisifiedMethods) {
-  (WebFrame as any).prototype[method] = deprecate.promisify((WebFrame as any).prototype[method])
-}
-
 const _webFrame = new WebFrame(window)
 
 export default _webFrame
diff --git a/lib/renderer/web-view/web-view-impl.ts b/lib/renderer/web-view/web-view-impl.ts
index 6849f3061..34545167b 100644
--- a/lib/renderer/web-view/web-view-impl.ts
+++ b/lib/renderer/web-view/web-view-impl.ts
@@ -273,7 +273,7 @@ export const setupMethods = (WebViewElement: typeof ElectronInternal.WebViewElem
   }
 
   for (const method of asyncPromiseMethods) {
-    (WebViewElement.prototype as Record<string, any>)[method] = deprecate.promisify(createPromiseHandler(method))
+    (WebViewElement.prototype as Record<string, any>)[method] = createPromiseHandler(method)
   }
 }
```

miniak:
@codebytere shouldn't we also remove the documentation for the callback versions?

codebytere:
I need to fix the smoke tests here 👇

miniak:
> I need to fix the smoke tests here 👇
yeah, when I tried locally what happens when the docs are updated, I could see a lot of those blowing up :)

codebytere:
Fixed now pending a issue with `electron-ts-defs`; we need to update such that Promise inner types are correctly parsed as arrays instead of raw types unconditionally.

cc @MarshallOfSound

docs/api/session.md
```diff
@@ -153,85 +118,7 @@ Returns `Promise<void>` - resolves when the storage data has been cleared.
...
-   Examples:
-     "192.168.1.1/16", "fefe:13::abc/33".
-
-* `<local>`
-
-   Match local addresses. The meaning of `<local>` is whether the
-   host matches one of: "127.0.0.1", "::1", "localhost".
-
-**[Deprecated Soon](modernization/promisification.md)**
+Writes any unwritten DOMStorage data to disk.s
```
miniak: typo disk.s

PR #17921, opened 2019-04-23
chore: upgrade chromium to 76

MarshallOfSound:
As per @electron/wg-upgrades 

Notes: Upgraded Chromium to 76

### DIFF
--- a/atom/browser/ui/views/autofill_popup_view.cc
+++ b/atom/browser/ui/views/autofill_popup_view.cc
@@ -150,7 +150,7 @@ void AutofillPopupView::OnSelectedRowChanged(
 
   if (current_row_selection) {
     int selected = current_row_selection.value_or(-1);
-    if (selected == -1 || selected >= children().size())
+    if (selected == -1 || static_cast<size_t>(selected) >= children().size())
       return;
     child_at(selected)->NotifyAccessibilityEvent(ax::mojom::Event::kSelection,
                                                  true);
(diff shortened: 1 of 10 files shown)


atom/browser/ui/views/autofill_popup_view.cc
```diff
@@ -150,7 +150,7 @@ void AutofillPopupView::OnSelectedRowChanged(
 
   if (current_row_selection) {
     int selected = current_row_selection.value_or(-1);
-    if (selected == -1 || selected >= child_count())
+    if (selected == -1 || static_cast<uint64_t>(selected) >= children().size())
```
nornagon:
```suggestion
    if (selected == -1 || static_cast<size_t>(selected) >= children().size())
```

atom/common/platform_util_win.cc
```diff
@@ -341,7 +341,7 @@ bool MoveItemToTrash(const base::FilePath& path) {
   // Elevation prompt enabled for UAC protected files.  This overrides the
   // SILENT, NO_UI and NOERRORUI flags.
 
-  if (base::win::GetVersion() >= base::win::VERSION_WIN8) {
+  if (base::win::GetVersion() >= base::win::Versionn::WIN8) {
```
deepak1556:
```suggestion
  if (base::win::GetVersion() >= base::win::Version::WIN8) {
```

deepak1556:
`76.0.3780.0` seems to be the latest canary, do we want to update here and merge ? Pretty sure there would be api and patch conflicts, I don't want electron-bot to handle it :)

nornagon:
I think we should merge this first and then roll forward. I'd rather be on _any_ 76 version in 6-0-x ASAP.

deepak1556:
cool, sounds fine.

PR #17923, opened 2019-04-23
docs: update 6.0.0 release dates per Upgrades WG

sofianguy:
#### Description of Change
Updating 6.0.0 release schedule according to @electron/wg-upgrades decision to upgrade to M76 in electron v6. Electron v6 and M76 will both release stable on July 30.

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/docs/electron-timelines.md
+++ b/docs/electron-timelines.md
@@ -1,15 +1,12 @@
 # Electron Release Timelines
+### Notes:
+* The `-beta.1` and `stable` dates are our *concrete* release dates.
+* We strive for weekly beta releases, however we often release more betas than scheduled.
+* All dates are our goals but there may be reasons for adjusting the stable deadline, such as security bugs.
 
 ## 5.0.0 Release Schedule
+*Includes: Chromium M73 and Node v12.0*
 
-Take a look at 5.0.0 Timeline [blog post](https://electronjs.org/blog/electron-5-0-timeline) for info about publicizing our release dates.
-
-### Notes
-
-- These dates are our goals but there may be reasons for adjusting the stable deadline, such as security bugs.
-- These are our scheduled beta releases, however we often release more betas than scheduled.
-
-## 5.0.0 Release Schedule
 Take a look at 5.0.0 Timeline [blog post](https://electronjs.org/blog/electron-5-0-timeline) for info about publicizing our release dates.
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@sofianguy it's probably too late for v5 to do this 😁, but are there plans to keep the beta versions up-to-date on the site with what builds were eventually published?

sofianguy:
@shiftkey Do you mean something like this where we link to each beta release and the date it was released? 

| Date     | Release      | Comments       |
| ---------------  | ------------ | -------------- |
| 2019-Jan-22 | [5.0.0-beta.1](https://electronjs.org/releases/beta?page=2#5.0.0-beta.1) |🔥 |
| 2019-Feb-03 | [5.0.0-beta.2](https://electronjs.org/releases/beta?page=2#5.0.0-beta.2) ||
| 2019-Feb-14 | [5.0.0-beta.3](https://electronjs.org/releases/beta?page=2#5.0.0-beta.3) ||

...and so on...

shiftkey:
@sofianguy yeah, that'd be great!

sofianguy:
@shiftkey all the dates of each beta release are listed on the site https://electronjs.org/releases/beta. Is your suggestion to solve readability and easy access to release info? I'm trying to pinpoint what problem we're trying to solve here.

shiftkey:
@sofianguy I guess my issue here is understanding the value of the release timeline if wasn't updated as things progressed. I wanted to link to this schedule in some "add/test Electron 5 support" issues for downstream projects, but given it wasn't updated for any of the `5.0.0-beta.x` builds it might not reflect that `5.0.0` is actually coming out as scheduled. Does that make sense?

sofianguy:
@shiftkey Ah, I see. Right now we want to strive for weekly beta releases so those dates are estimates at the moment. However the -beta.1 and the stable release dates are our set-in-stone dates. Most apps want to see the stable release date, but I see your reasoning for updating this for beta releases also for following progression.

shiftkey:
>  However the -beta.1 and the stable release dates are our set-in-stone dates.

Makes sense, thanks for the extra insight. Maybe this could eventually live under `electronjs.org/releases` so you can avoid the manual work of updating betas, and everyone else can go there to see this information as well as the betas shipped (to save duplicating work)?

PR #17925, opened 2019-04-23
chore: update base::Bind in app module

codebytere:
#### Description of Change

Update `base::Bind` to `base::BindRepeating` and `base::BindOnce` where applicable.

cc @deepak1556 @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/atom/browser/api/atom_api_app.cc
+++ b/atom/browser/api/atom_api_app.cc
@@ -1354,20 +1354,26 @@ void App::BuildPrototype(v8::Isolate* isolate,
                  base::BindRepeating(&Browser::ClearRecentDocuments, browser))
       .SetMethod("setAppUserModelId",
                  base::Bind(&Browser::SetAppUserModelID, browser))
-      .SetMethod("isDefaultProtocolClient",
-                 base::BindRepeating(&Browser::IsDefaultProtocolClient, browser))
-      .SetMethod("setAsDefaultProtocolClient",
-                 base::BindRepeating(&Browser::SetAsDefaultProtocolClient, browser))
-      .SetMethod("removeAsDefaultProtocolClient",
-                 base::BindRepeating(&Browser::RemoveAsDefaultProtocolClient, browser))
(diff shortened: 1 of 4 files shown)


codebytere:
Depends on https://github.com/electron/electron/pull/17941

atom/browser/api/atom_api_app.cc
```diff
@@ -1063,8 +1065,9 @@ void App::ImportCertificate(const base::DictionaryValue& options,
         base::Value::ToUniquePtrValue(options.Clone()));
     CertificateManagerModel::Create(
         browser_context.get(),
-        base::Bind(&App::OnCertificateManagerModelCreated,
-                   base::Unretained(this), base::Passed(&copy), callback));
+        base::BindRepeating(&App::OnCertificateManagerModelCreated,
+                            base::Unretained(this), base::Passed(&copy),
```
deepak1556: can you try `std::move` instead of `base::Passed` here.

atom/browser/api/event_emitter.h
```diff
@@ -51,25 +51,25 @@ class EventEmitter : public Wrappable<T> {
   template <typename... Args>
   bool EmitCustomEvent(const base::StringPiece& name,
                        v8::Local<v8::Object> event,
-                       const Args&... args) {
+                       Args&&... args) {
```
deepak1556: Does this mean that emit arguments should always be movable ?

atom/browser/api/atom_api_app.cc
deepak1556: Looks like the change needs to be reverted, cannot forward in repeating variation.

atom/browser/api/event_emitter.h
zcbenz: For template args, this means _universal reference_ (or _forward reference_):
https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers

PR #17941, opened 2019-04-24
chore: support converting OnceCallback to V8

zcbenz:
#### Description of Change

This PR adds a converter to convert `base::OnceCallback` to V8 function. This is needed for NeworkService migration of protocol module, and I'm separating it for review.

Note that the converter is implemented by reusing the code of `base::Callback`'s converter, this is because the callback converter relies on hundreds lines of C++ template code, and making it support both `base::Callback` and `base::OnceCallback` is twisting my mind and will make the code much harder to understand.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


nornagon:
Hm, is there no way we could reuse `base::AdaptCallbackForRepeating`?

zcbenz:
> Hm, is there no way we could reuse `base::AdaptCallbackForRepeating`?

I didn't know it exists, looks like a better solution.

PR #17948, opened 2019-04-24
refactor: use IPC helpers in window-setup

miniak:
#### Description of Change
Use `ipcRendererUtils.invoke(Sync)` / `ipcMainUtils.handle` to eliminate duplicate code and properly propagate errors to the caller side.

We need to handle `loadURL` similar to #18990

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/guest-window-manager.js
+++ b/lib/browser/guest-window-manager.js
@@ -3,6 +3,7 @@
 const { BrowserWindow, webContents } = require('electron')
 const { isSameOrigin } = process.electronBinding('v8_util')
 const { ipcMainInternal } = require('@electron/internal/browser/ipc-main-internal')
+const ipcMainUtils = require('@electron/internal/browser/ipc-main-internal-utils')
 const parseFeaturesString = require('@electron/internal/common/parse-features-string')
 
 const hasProp = {}.hasOwnProperty
@@ -153,6 +154,9 @@ const getGuestWindow = function (guestContents) {
       guestWindow = BrowserWindow.fromWebContents(hostContents)
     }
   }
+  if (!guestWindow) {
+    throw new Error('getGuestWindow failed')
+  }
   return guestWindow
 }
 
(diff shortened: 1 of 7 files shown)


lib/browser/guest-window-manager.js
```diff
@@ -267,53 +271,42 @@ ipcMainInternal.on('ELECTRON_GUEST_WINDOW_MANAGER_INTERNAL_WINDOW_OPEN', functio
...
+    console.error(`Blocked ${event.sender.getURL()} from calling ${method}`)
+    throw new Error(`Access denied to guestId: ${guestContents.id}`)
   }
 
-  if (!canAccessWindow(event.sender, guestContents) || !windowMethods.has(method)) {
-    console.error(`Blocked ${event.sender.getURL()} from calling ${method} on its opener.`)
-    event.returnValue = null
-    return
+  if (!windowMethods.has(method)) {
+    console.error(`Blocked ${event.sender.getURL()} from calling ${method}`)
```
alexeykuzmin: Should this be different from the message up there?

spec/chromium-spec.js
```diff
@@ -666,10 +666,10 @@ describe('chromium feature', () => {
       if (webview != null) webview.remove()
     })
 
-    it('does nothing when origin of webview src URL does not match opener', (done) => {
+    it('fails when origin of webview src URL does not match opener', (done) => {
       webview = new WebView()
       webview.addEventListener('console-message', (e) => {
-        expect(e.message).to.equal('')
+        expect(e.message).to.equal('null')
```
alexeykuzmin: The `'null'` string looks weird.

miniak: it does, but that's how `console.log(null)` ends up in `'console-message'`

lib/browser/guest-window-manager.js
miniak: what do you mean exactly?

alexeykuzmin: I mean that you throw exceptions with two different errors, but log two exactly the same messages to the console.
Maybe the message should be different as well?

PR #17955, opened 2019-04-24
build: ensure consistent lock files across multiple machines

MarshallOfSound:
Uses `npx yarn` instead of `npm` to fix inconsistent lock files

Notes: no-notes


nornagon:
tests failing? with 
```
../src/main.cc:1:10: fatal error: js_native_api.h: No such file or directory
```

MarshallOfSound:
This is the commit that does the `npm->yarn` conversion.

https://github.com/electron/electron/pull/17955/commits/f00066f207ef7956928c386977908b66e88d7c69

MarshallOfSound:
As an added bonus `--frozen-lockfile` will cause CI to fail if the `package.json` doesn't match the current lock file.

PR #17957, opened 2019-04-24
build: allow pulling in a later version of Clang

richard-townsend-arm:
Due to an MSVC interoperability defect in the version of Clang used for
Chromium 75, it's necessary to build with a later version of Clang. This
change allows the correct version to be specified by setting
`LLVM_FORCE_REVISION=version` in the environment during a `gclient sync`.
We will provide a documentation update when a fixed version of Clang is
available. This PR is only intended to be a temporary workaround and
will be removed when Electron's Chromium updates to a new version.

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/patches/common/chromium/.patches
+++ b/patches/common/chromium/.patches
@@ -7,6 +7,7 @@ blink_local_frame.patch
 blink_world_context.patch
 browser_compositor_mac.patch
 can_create_window.patch
+clang_fetch_env.patch
 disable_hidden.patch
 dom_storage_limits.patch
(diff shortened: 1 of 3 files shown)


jkleinsc:
@richard-townsend-arm is this needed for Chromium 76?  We are looking to use Chromium 76 for our next release.

richard-townsend-arm:
Ah, OK, I was thinking that it’d be 75-based. Hopefully the compiler will be fixed in the 76 timeframe, so closing for now.

PR #17960, opened 2019-04-24
refactor: add emitCustomEvent() helper

miniak:
#### Description of Change
Simplifies #17902 by moving duplicate code emitting filtering events into a helper function.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/rpc-server.js
+++ b/lib/browser/rpc-server.js
@@ -8,6 +8,7 @@ const path = require('path')
 const v8Util = process.electronBinding('v8_util')
 const eventBinding = process.electronBinding('event')
 const clipboard = process.electronBinding('clipboard')
+const features = process.electronBinding('features')
 
 const { isPromise } = electron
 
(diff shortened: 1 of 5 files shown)


chauser123:
Pull request not working

lib/browser/rpc-server.js
```diff
@@ -274,6 +275,15 @@ const handleRemoteCommand = function (channel, handler) {
   })
 }
 
+const emitCustomEvent = function (contents, eventName, ...args) {
+  const event = eventBinding.createWithSender(contents)
+
+  electron.app.emit(eventName, event, contents, ...args)
```
alexeykuzmin: What for?

ppontes: It's applying here automatically the event forwarding step that was removed from `web-contents.js`. Will we always want to forward _any_ custom events to `app`? Not sure if the purpose of keeping the `forwardedEvents` array was to keep a list of events to forward, while others wouldn't be forwarded.

miniak: @alexeykuzmin 
- check the other PR that depends on this, I need to be able to check if there are no listeners, which is much easier to do with this change
- it's cleaner like this and also there is internal listener used, which could be screwed up by the app code theoretically

miniak: if we need to change the behavior in the future, we will just change the code to do whatever might be needed at the time.

PR #17965, opened 2019-04-25
feat: migrate protocol module to NetworkService (Part 2)

zcbenz:
#### Description of Change

This is part of the changes to reimplement the protocol module with NetworkService API, the new implementation lives in parallel with current implementation and only gets used when `--enable-features=NetworkService` is passed.

This PR implements the `protocol.registerFileProtocol/registerBufferProtocol/registerStringProtocol` with NetworkService APIs.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/net/atom_url_loader_factory.cc
+++ b/atom/browser/net/atom_url_loader_factory.cc
@@ -7,9 +7,12 @@
 #include <string>
 #include <utility>
 
+#include "atom/common/native_mate_converters/file_path_converter.h"
 #include "atom/common/native_mate_converters/net_converter.h"
 #include "content/public/browser/browser_thread.h"
+#include "content/public/browser/file_url_loader.h"
 #include "gin/dictionary.h"
+#include "net/base/filename_util.h"
 #include "services/network/public/cpp/url_loader_completion_status.h"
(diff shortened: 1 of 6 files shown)


atom/browser/net/atom_url_loader_factory.cc
```diff
@@ -7,15 +7,24 @@
...
 #include "services/network/public/mojom/url_loader.mojom.h"
 
+#include "atom/common/node_includes.h"
+
 using content::BrowserThread;
 
 namespace atom {
 
-AtomURLLoaderFactory::AtomURLLoaderFactory() {}
+AtomURLLoaderFactory::AtomURLLoaderFactory(ProtocolType type,
```
deepak1556: since the loader factory is unique to every handler. Can we instead have variations of the factory like `FileURLLoaderFactory`, `BufferURLLoaderFactory`, `StringURLLoaderFactory` ?

zcbenz: I'm thinking of providing a universal `registerProtocol` API that can return arbitrary response in one handler, but I'm not sure how the design of the class should look like. I'll refactor the class after implementing the `Http` and `Stream` handlers.

PR #17970, opened 2019-04-25
refactor: session.clearAuthCache nws13n

nornagon:
#### Description of Change
Refactor session.clearAuthCache to use the network service. This results in losing the ability to filter what is cleared; we should deprecate this in 6-0-x.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: session.clearAuthCache no longer allows filtering the cleared cache entries.

### DIFF
--- a/atom/browser/api/atom_api_session.cc
+++ b/atom/browser/api/atom_api_session.cc
@@ -74,15 +74,6 @@ struct ClearStorageDataOptions {
   uint32_t quota_types = StoragePartition::QUOTA_MANAGED_STORAGE_MASK_ALL;
 };
 
-struct ClearAuthCacheOptions {
-  std::string type;
-  GURL origin;
-  std::string realm;
-  base::string16 username;
-  base::string16 password;
-  net::HttpAuth::Scheme auth_scheme;
-};
-
 uint32_t GetStorageMask(const std::vector<std::string>& storage_types) {
   uint32_t storage_mask = 0;
   for (const auto& it : storage_types) {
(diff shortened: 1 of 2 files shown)


nornagon:
Do not merge yet, I want to understand if it's actually OK to remove these APIs. (& land the deprecation first.)

nornagon:
Landing this because we're going to have to figure out _some_ way of moving this function to the network service before 7-0-x lands, so if we need to do more to support pre-existing use cases, we can build on top of this change :)

HeikoH:
@nornagon - is there are a 'new' way to clear the cache for a specific `origin` after this API change? i.e. we're still one 5.x and using something like `sessing.clearAuthCache({ type: 'clientCertificate', origin: url })`, which doesn't work anymore. And the new behavior where entire auth cache is cleared doesn't really work for us.

nornagon:
@HeikoH Sadly, the [underlying API](https://source.chromium.org/chromium/chromium/src/+/master:services/network/public/mojom/network_context.mojom;l=975-977;drc=e2e073d4a6a5418a55ca5ade66a252145eb7fafb) does not support filtering by origin.

Can you say more about your use case?

HeikoH:
We're connecting to various back-ends that need to be authenticated with using PKI certificates, potentially different ones. Currently we also allow users to *un-authenticate* with individual endpoints, which currently uses `session.clearAuthCache({ type: 'clientCertificate', origin: url })`. By changing this to the argument-less `session.clearAuthCache` the entire auth cache is cleared.

PR #17978, opened 2019-04-25
chore: Add patch to partially revert chromium crashpad change

nitsakh:
#### Description of Change
This adds a patch to support functionality that we were using but chromium changed it. Electron uses breakpad on windows, chromium uses crashpad (which is newer). So this patch is needed until we update electron to use crashpad for windows.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


codebytere:
@nitsakh can you update the patch conflict?

nitsakh:
@nornagon I'm working on https://github.com/electron/electron/issues/17993, which should be the path going forward.

PR #17985, opened 2019-04-26
build: disable Vulkan validation layers

miniak:
#### Description of Change
Removes the following DLLs from the build, which are only meant to be used for Chromium development:
- VkLayer_core_validation.dll
- VkLayer_object_tracker.dll
- VkLayer_parameter_validation.dll
- VkLayer_threading.dll
- VkLayer_unique_objects.dll

[angle.gni](https://cs.chromium.org/chromium/src/third_party/angle/gni/angle.gni?type=cs&q=angle_enable_vulkan_validation_layers&g=0&l=78) for more details

Follow up to https://github.com/electron/electron/pull/17694

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Removed Vulkan validation layers DLLs from electron.zip, which are only meant to be used for Chromium development.

### DIFF
--- a/build/args/all.gn
+++ b/build/args/all.gn
@@ -14,5 +14,6 @@ proprietary_codecs = true
 ffmpeg_branding = "Chrome"
 
 enable_basic_printing = true
+angle_enable_vulkan_validation_layers = false
 
 is_cfi = false


deepak1556:
> leave enable_vulkan at its default as it doesn't cost us much to leave it enabled.

whats the benefit of this ?

nornagon:
It keeps us closer to Chromium's configuration. I'd rather ask, what's the benefit of disabling it? :)

deepak1556:
Cool, I am not super tied to my argument, but just that what we are doing is just disabling a subset argument to achieve the main purpose, while the main gateway for vulkan in the codebase is `enable_vulkan`. But if we are going to be using vulkan when chrome enables it, then this change makes sense.

nornagon:
If we're going to backport this everywhere that the `enable_vulkan` patch went, we should revert that in this PR as well so we only have to deal with one set of backports.

miniak:
@nornagon, we only did that in 4-1-x and 5-0-x

miniak:
#17694 was abandoned

PR #18005, opened 2019-04-27
fix: skip media-started-playing media-paused events test when media is not supported

miniak:
#### Description of Change
Skip a test requiring the `<audio>` element to support playing WAV file when the format is not supported (internal VSCode Electron build has special FFMpeg configuration without codecs, where this test would fail with "Failed to load because no supported source was found.")

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/spec/webview-spec.js
+++ b/spec/webview-spec.js
@@ -1061,6 +1061,12 @@ describe('<webview> tag', function () {
   })
 
   describe('media-started-playing media-paused events', () => {
+    before(function () {
+      if (!document.createElement('audio').canPlayType('audio/wav')) {
+        this.skip()
+      }
+    })
+
     it('emits when audio starts and stops playing', async () => {
       await loadWebView(webview, { src: `file://${fixtures}/pages/base-page.html` })
 


spec/webview-spec.js
```diff
@@ -1061,6 +1061,12 @@ describe('<webview> tag', function () {
   })
 
   describe('media-started-playing media-paused events', () => {
+    before(function () {
+      if (!document.createElement('audio').canPlayType('audio/wav')) {
```
MarshallOfSound: If electrons ffmpeg build ever breaks for some reason then this test will not capture the issue anymore.

I'd prefer if you used the --grep and --invert flags to not run this test in your internal CI

miniak: makes sense, we'll work around it in our CI then

miniak: @MarshallOfSound hmm, actually wouldn't it be better to have a separate test, which checks if ffmpeg / list of supported formats works as expected? this test is whether a particular webContents event works, which cannot be verified if the format used to test it is not supported

MarshallOfSound: Having a separate test for that specifically probably makes sense.  You'll still have the same issue of having to disable the test yourself in your ci

PR #18016, opened 2019-04-28
 feat: allow setting of global fallback user agent

MarshallOfSound:
Currently we support letting apps set user agents on `webContents` and `session` but not "for everything".  Well this PR allows you to do just that.  In particular it allows apps to do a few things

* Set the user agent in **one** place for everything regardless of sessions / webContents used
* Set the user agent for parts of Chromium that we previously couldn't override due to incomplete wiring inside Chromium.  E.g. Tracing controllers, accessibility, service workers, etc.

Notes: Added new `app.get/setUserAgentFallback` APIs to allow setting the apps useragent globally

### DIFF
--- a/atom/browser/api/atom_api_app.cc
+++ b/atom/browser/api/atom_api_app.cc
@@ -1456,8 +1456,8 @@ void App::BuildPrototype(v8::Isolate* isolate,
 #if defined(OS_MACOSX)
       .SetProperty("dock", &App::GetDockAPI)
 #endif
-      .SetMethod("setUserAgentFallback", &App::SetUserAgentFallback)
-      .SetMethod("getUserAgentFallback", &App::GetUserAgentFallback)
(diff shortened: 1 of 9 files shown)


MarshallOfSound:
Put this on the @electron/wg-releases agenda to discuss `target/6-0-x`

codebytere:
```

--- a/atom/browser/api/atom_api_app.cc
+++ b/atom/browser/api/atom_api_app.cc
@@ -1456,7 +1456,8 @@
 #if defined(OS_MACOSX)
       .SetProperty("dock", &App::GetDockAPI)
 #endif
-      .SetProperty("userAgentFallback", &App::GetUserAgentFallback, &App::SetUserAgentFallback)
+      .SetProperty("userAgentFallback", &App::GetUserAgentFallback,
+                   &App::SetUserAgentFallback)
       .SetMethod("enableSandbox", &App::EnableSandbox);
 }

```

PR #18022, opened 2019-04-28
docs: correct doc for registerFileProtocol

pathim:
#### Description of Change

In the registerFileProtocol docs the "headers" argument of the callback was described as being a list.
In fact is has to be an object mapping header-entries to values. This can be seen in Line 326 of [`/spec/api-protocol-spec.js`](https://github.com/electron/electron/blob/fe618631f1e6858077874226682fa299d1700b36/spec/api-protocol-spec.js#L326).
#### Checklist

#### Release Notes

Notes: no-notes


codebytere:
Thanks!

PR #18029, opened 2019-04-29
refactor: rewire the desktop capturer API to remove race conditions

MarshallOfSound:
We now create a new instance of atom::api::DesktopCapturer for every request instead of weirdly re-using the same instance and queuing requests.  This means there is now a 1:1 relationship between request and DesktopCapturer so there isn't a race condition between the observer for one request calling back before the observer of another.  This is an issue ever since the backing APIs moved to worker threads.

This also does a few things to ensure memory management
* Only ever listen to one event per-request, after that we wipe the emit function to ignore all future events
* Ensures we clean up the `window_capturer_`, `screen_capturer_` and `captured_sources_` in native land once the request is over.

This _in theory_ fixes a flake we've been seeing on CI where we try to resolve the promise for a request that no longr exists.

Notes: Fixed race condition in the `desktopCapturer` module where some requests for sources would never be resolved or unhandled exceptions would be thrown in the main process.

### DIFF
--- a/atom/browser/api/atom_api_desktop_capturer.cc
+++ b/atom/browser/api/atom_api_desktop_capturer.cc
@@ -226,7 +226,7 @@ void Initialize(v8::Local<v8::Object> exports,
                 void* priv) {
   v8::Isolate* isolate = context->GetIsolate();
   mate::Dictionary dict(isolate, exports);
-  dict.Set("desktopCapturer", atom::api::DesktopCapturer::Create(isolate));
+  dict.SetMethod("createDesktopCapturer", &atom::api::DesktopCapturer::Create);
 }
 
 }  // namespace
--- a/lib/browser/desktop-capturer.js
+++ b/lib/browser/desktop-capturer.js
@@ -2,13 +2,12 @@
 
 const ipcMainUtils = require('@electron/internal/browser/ipc-main-internal-utils')
 
-const { desktopCapturer } = process.electronBinding('desktop_capturer')
+const { createDesktopCapturer } = process.electronBinding('desktop_capturer')
 const eventBinding = process.electronBinding('event')
 
 const deepEqual = (a, b) => JSON.stringify(a) === JSON.stringify(b)
 
-// A queue for holding all requests from renderer process.
-let requestsQueue = []
+let currentlyRunning = []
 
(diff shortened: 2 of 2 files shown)


atom/browser/api/atom_api_desktop_capturer.cc
```diff
@@ -62,7 +62,11 @@ DesktopCapturer::DesktopCapturer(v8::Isolate* isolate) {
   Init(isolate);
 }
 
-DesktopCapturer::~DesktopCapturer() {}
+DesktopCapturer::~DesktopCapturer() {
+  window_capturer_.reset();
```
nornagon: Are these resets necessary? Won't this happen automatically?

atom/browser/api/atom_api_desktop_capturer.cc
```diff
@@ -220,13 +224,18 @@ void DesktopCapturer::BuildPrototype(
...
+}
+
 void Initialize(v8::Local<v8::Object> exports,
                 v8::Local<v8::Value> unused,
                 v8::Local<v8::Context> context,
                 void* priv) {
   v8::Isolate* isolate = context->GetIsolate();
   mate::Dictionary dict(isolate, exports);
-  dict.Set("desktopCapturer", atom::api::DesktopCapturer::Create(isolate));
+  dict.SetMethod("createDesktopCapturer", &CreateDesktopCapturer);
```
nornagon:
```suggestion
  dict.SetMethod("createDesktopCapturer", &atom::api::DesktopCapturer::Create);
```

no need for a helper function

atom/browser/api/atom_api_desktop_capturer.cc
deepak1556: yup they will be cleaned up automatically.

MarshallOfSound: This was me puzzling over some memory leaks before I dove into the refactor, I'll remove

MarshallOfSound:
@nornagon Yeah the JS side of this is Not Too Great ™️ but this was the shortest refactor that fixed the issues.  I'm planning on following up at some point with a proper rewrite with promises in native land and such 🤔 But that's a bigger task

PR #18035, opened 2019-04-29
chore: convert some apis incl. View APIs from base::Bind

codebytere:
#### Description of Change

Continuation of `base::Bind` migration for the Views API & a few other classes. These instances are called in the class `Initialize` methods, so they should all need to be `RepeatingCallbacks` and thus created with `base::BindRepeating`.

cc @deepak1556 @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


codebytere:
@deepak1556 the pr description makes it clearer since it does _include_ all the view APIs, but i wasn't sure how to indicate `view apis & more` in the title 🤔

PR #18038, opened 2019-04-29
chore: convert session base::Bind instances

codebytere:
#### Description of Change

Convert instances of `base::Bind` to `base::BindOnce` and `base::BindRepeating` as applicable in the session module as well as in Autofill popups.

cc @deepak1556 @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/atom/browser/api/atom_api_session.cc
+++ b/atom/browser/api/atom_api_session.cc
@@ -350,9 +350,9 @@ v8::Local<v8::Promise> Session::ResolveProxy(mate::Arguments* args) {
   args->GetNext(&url);
 
   browser_context_->GetResolveProxyHelper()->ResolveProxy(
-      url,
-      base::Bind(util::CopyablePromise::ResolveCopyablePromise<std::string>,
-                 util::CopyablePromise(promise)));
+      url, base::BindRepeating(
+               util::CopyablePromise::ResolveCopyablePromise<std::string>,
+               util::CopyablePromise(promise)));
 
   return handle;
 }
@@ -413,8 +413,7 @@ v8::Local<v8::Promise> Session::ClearStorageData(mate::Arguments* args) {
   storage_partition->ClearData(
       options.storage_types, options.quota_types, options.origin, base::Time(),
       base::Time::Max(),
-      base::Bind(util::CopyablePromise::ResolveEmptyCopyablePromise,
-                 util::CopyablePromise(promise)));
+      base::BindOnce(util::Promise::ResolveEmptyPromise, std::move(promise)));
   return handle;
 }
 
@@ -496,17 +495,18 @@ void Session::DisableNetworkEmulation() {
       network_emulation_token_, network::mojom::NetworkConditions::New());
 }
 
-void WrapVerifyProc(base::Callback<void(const VerifyRequestParams& request,
-                                        base::Callback<void(int)>)> proc,
-                    const VerifyRequestParams& request,
-                    base::OnceCallback<void(int)> cb) {
-  proc.Run(request, base::AdaptCallbackForRepeating(std::move(cb)));
+void WrapVerifyProc(
+    base::OnceCallback<void(const VerifyRequestParams& request,
+                            base::RepeatingCallback<void(int)>)> proc,
+    const VerifyRequestParams& request,
+    base::OnceCallback<void(int)> cb) {
+  std::move(proc).Run(request, base::AdaptCallbackForRepeating(std::move(cb)));
 }
 
 void Session::SetCertVerifyProc(v8::Local<v8::Value> val,
                                 mate::Arguments* args) {
-  base::Callback<void(const VerifyRequestParams& request,
-                      base::Callback<void(int)>)>
+  base::RepeatingCallback<void(const VerifyRequestParams& request,
+                               base::RepeatingCallback<void(int)>)>
       proc;
   if (!(val->IsNull() || mate::ConvertFromV8(args->isolate(), val, &proc))) {
     args->ThrowError("Must pass null or function");
@@ -517,7 +517,7 @@ void Session::SetCertVerifyProc(v8::Local<v8::Value> val,
       FROM_HERE, {BrowserThread::IO},
       base::BindOnce(&SetCertVerifyProcInIO,
                      WrapRefCounted(browser_context_->GetRequestContext()),
-                     base::Bind(&WrapVerifyProc, proc)));
+                     base::BindRepeating(&WrapVerifyProc, proc)));
 }
 
 void Session::SetPermissionRequestHandler(v8::Local<v8::Value> val,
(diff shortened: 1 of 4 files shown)


atom/browser/api/atom_api_session.cc
```diff
@@ -271,12 +271,12 @@ void DoCacheActionInIO(
   // Call GetBackend and make the backend's ptr accessable in OnGetBackend.
   using BackendPtr = disk_cache::Backend*;
   auto** backend_ptr = new BackendPtr(nullptr);
-  net::CompletionRepeatingCallback on_get_backend =
-      base::Bind(&OnGetBackend, base::Owned(backend_ptr), action,
-                 util::CopyablePromise(promise));
-  int rv = http_cache->GetBackend(backend_ptr, on_get_backend);
+  net::CompletionOnceCallback on_get_backend =
+      base::BindOnce(&OnGetBackend, base::Owned(backend_ptr), action,
+                     util::CopyablePromise(promise));
```
nornagon: does this still need to be `CopyablePromise` now that we're a `OnceCallback`?

atom/browser/api/atom_api_session.cc
```diff
@@ -419,9 +419,9 @@ v8::Local<v8::Promise> Session::ResolveProxy(mate::Arguments* args) {
   args->GetNext(&url);
 
   browser_context_->GetResolveProxyHelper()->ResolveProxy(
-      url,
-      base::Bind(util::CopyablePromise::ResolveCopyablePromise<std::string>,
-                 util::CopyablePromise(promise)));
+      url, base::BindRepeating(
```
nornagon: should `ResolveProxyCallback` be `ResolveProxyOnceCallback`?

atom/browser/api/atom_api_session.cc
```diff
@@ -692,7 +692,7 @@ void Session::CreateInterruptedDownload(const mate::Dictionary& options) {
   }
   auto* download_manager =
       content::BrowserContext::GetDownloadManager(browser_context());
-  download_manager->GetDelegate()->GetNextId(base::Bind(
+  download_manager->GetDelegate()->GetNextId(base::BindRepeating(
```
nornagon:
```suggestion
  download_manager->GetDelegate()->GetNextId(base::BindOnce(
```

https://cs.chromium.org/chromium/src/content/public/browser/download_manager.h?l=231&rcl=8d3660a6b51f5ca40d1ce56981ec107645344ca0

atom/browser/api/atom_api_session.cc
codebytere: nope! i'll update associated code, thought i got this one!

atom/browser/api/atom_api_session.cc
codebytere: @nornagon we don't enforce this with any degree of standardization across the codebase, but i'm happy to change it if you think it'd be clearer :)

atom/browser/api/atom_api_session.cc
codebytere: @nornagon i originally thought this too, but that's not in fact the callback being called: it's `GetNextID` on the _delegate_ which is a `RepeatingCallback` https://cs.chromium.org/chromium/src/content/public/browser/download_manager_delegate.h?sq=package:chromium&g=0&l=73

atom/browser/api/atom_api_session.cc
nornagon: Well, given that it looks like the resolution only happens once, I think `BindOnce` is warranted.

codebytere: done!

atom/browser/net/resolve_proxy_helper.cc
```diff
@@ -75,8 +75,8 @@ void ResolveProxyHelper::OnProxyLookupComplete(
   if (proxy_info)
     proxy = proxy_info->ToPacString();
 
-  if (!completed_request.callback.is_null())
-    completed_request.callback.Run(proxy);
+  if (!std::move(completed_request.callback).is_null())
+    std::move(completed_request.callback).Run(proxy);
```
nornagon: hm, this doesn't make sense, you can't move it twice.

codebytere: oh derp yeah: this may need to be a RepeatingCallback then?

atom/browser/api/atom_api_session.cc
```diff
@@ -543,17 +538,18 @@ void Session::DisableNetworkEmulation() {
       network_emulation_token_, network::mojom::NetworkConditions::New());
 }
 
-void WrapVerifyProc(base::Callback<void(const VerifyRequestParams& request,
-                                        base::Callback<void(int)>)> proc,
-                    const VerifyRequestParams& request,
-                    base::OnceCallback<void(int)> cb) {
-  proc.Run(request, base::AdaptCallbackForRepeating(std::move(cb)));
+void WrapVerifyProc(
+    base::OnceCallback<void(const VerifyRequestParams& request,
```
deepak1556: Why is this `base::OnceCallback` while the original declaration is `base::RepeatingCallback` for `proc` ?

atom/browser/api/atom_api_session.cc
```diff
@@ -224,29 +224,27 @@ void ResolveOrRejectPromiseInUI(util::Promise promise, int net_error) {
...
+    util::Promise::RejectPromise(std::move(promise), std::move(err_msg));
   } else if (backend_ptr && *backend_ptr) {
     if (action == Session::CacheAction::CLEAR) {
-      auto success =
-          (*backend_ptr)
-              ->DoomAllEntries(base::BindOnce(&ResolveOrRejectPromiseInUI,
-                                              promise.GetPromise()));
+      auto success = (*backend_ptr)
+                         ->DoomAllEntries(base::BindOnce(
+                             &ResolveOrRejectPromiseInUI, std::move(promise)));
```
deepak1556: Wouldn't the storage class `BindState` of `base::Bind` move the parameters into their members even before the callback is called ? If thats the case, trying to access it below will crash. /cc @zcbenz

atom/browser/api/atom_api_session.cc
```diff
@@ -271,12 +269,11 @@ void DoCacheActionInIO(
   // Call GetBackend and make the backend's ptr accessable in OnGetBackend.
   using BackendPtr = disk_cache::Backend*;
   auto** backend_ptr = new BackendPtr(nullptr);
-  net::CompletionRepeatingCallback on_get_backend =
-      base::Bind(&OnGetBackend, base::Owned(backend_ptr), action,
-                 util::CopyablePromise(promise));
-  int rv = http_cache->GetBackend(backend_ptr, on_get_backend);
+  net::CompletionOnceCallback on_get_backend = base::BindOnce(
+      &OnGetBackend, base::Owned(backend_ptr), action, std::move(promise));
+  int rv = http_cache->GetBackend(backend_ptr, std::move(on_get_backend));
```
deepak1556: same here for `on_get_backend`

atom/browser/api/atom_api_session.cc
zcbenz: Yeah we can't `std::move(promise)` twice.

According to the source code of `DoomAllEntries`, the return value is always `ERR_IO_PENDING`, so we can just ignore the return value. The return value is only kept to make old code compile.
https://cs.chromium.org/chromium/src/net/disk_cache/blockfile/backend_impl.cc?sq=package:chromium&g=0&l=1284

PR #18045, opened 2019-04-29
docs: added 'nodeIntegration: true' to the first-app.md tutorial

FidelVe:
#### Description of Change
Added "webPreferences: {nodeIntegration: true}" to the code in the example at first-app.md. Without this change when an user runs the code in the example the app doesnt shows the node, chrome and electron versions. Also the edit complies with the code that appears in the electron-quick-start repo.

#### Checklist

#### Release Notes

Notes: no notes

### DIFF
--- a/docs/tutorial/first-app.md
+++ b/docs/tutorial/first-app.md
@@ -109,7 +109,13 @@ const { app, BrowserWindow } = require('electron')
 
 function createWindow () {
   // Create the browser window.
-  let win = new BrowserWindow({ width: 800, height: 600 })
+  let win = new BrowserWindow({ 
+    width: 800,
+    height: 600,
+    webPreferences: {
+      nodeIntegration: true
+    }
+   })
 
   // and load the index.html of the app.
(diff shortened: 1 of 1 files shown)


malept:
I may have tried that months ago (along with getting rid of `document.write`) in a local branch of `electron-quick-start`. I'll look later on today whether I actually saved it somewhere.

FidelVe:
> Instead of recommending users turn on `nodeIntegration: true` I'd prefer if we updated this "first app" so that it didn't require `node` in the renderer.
> 
> cc @electron/wg-docs-tools

that would be better, but in that case the example must be changed and the electron-quick-start repo should also be modified to reflect that.

malept:
> in that case the example must be changed and the electron-quick-start repo should also be modified to reflect that.

Yep, I decided to start on the other end first: https://github.com/electron/electron-quick-start/pull/279

codebytere:
Given the actions above, this PR is no longer needed. However, we appreciate your contribution and welcome more of them!

Issue #18034, opened 2019-04-29
BrowserWindow.show() always doesn't focus.


TheGoddessInari:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  *  4.1.4
* **Operating System:**
  *  Windows 10 (1903)
* **Last Known Working Electron version:**:
  *  N/A

### Expected Behavior
BrowserWindow.show(), as documented, shows **and** focuses the window.

### Actual Behavior
BrowserWindow.show() doesn't get focus, as noted in https://github.com/TheGoddessInari/rambox/issues/92.
I'm now having to explicitly call .focus() after .show(), even though the documentation explicitly states that .show() also focuses.

### Additional Information

I checked the code, and I don't see NativeWindow call anything like Focus. Maybe I'm looking in the wrong part of the code, but NativeWindowViews::Show() doesn't look to directly or indirectly call NativeWindowViews::Focus(), which is what you would expect from the documentation.

This is possibly specific to Linux? As ShowWindow on Windows **should** activate the window same as Focus does.

Many other applications use .show() and then .focus() on browserwindow, even though the focus is specifically documented as redundant in that case. But either way, the documentation doesn't match the behavior.

It's hard for me to search the code online since GitHub search doesn't seem to support literal searches within "quotes", so I might've missed something.

Issue #2867, opened 2015-09-22
Windows 10: Window is not put to front on focus()


bpasero:
I am hitting an issue where a window (in Windows 10) is not being put to the front when I call focus() on it. The same code brings the window to the front on Windows 8 and below, so something must have changed in Windows 10 with regards to window management.

What I see happening is that the task bar is flashing in orange, so there is some indication. Just the window does not get focussed properly.

OlsonDev:
I find it interesting you've had this working at all in the past. I've always found it to be relatively inconsistent due to `ForegroundFlashCount` and `ForegroundLockTimeout`; see http://www.thewindowsclub.com/disable-flashing-taskbar-buttons-windows if you're unfamiliar.

zcbenz:
Under the hood `focus()` just called [SetForegroundWindow](https://msdn.microsoft.com/en-us/library/windows/desktop/ms633539(v=vs.85).aspx) API, from the description it only brings window to foreground if the calling process is a foreground process.

From #1954 it seems that Windows 10 starts to treat the renderer process of Chromium as foreground process, and thus the browser process becomes a background process, resulting in `SetForegroundWindow` not working.

I can probably fix this by calling `AllowSetForegroundWindow` in the renderer process to allow browser process to set foreground window, but I need to install Windows 10 first.

P.S. @bpasero Is it possible for you to reach the Windows team to verify my assumption? I know Microsoft is HUGE and it is hard to find the right team, but it can help understand the mysterious behaviors.

bpasero:
@zcbenz I will try to find a contact, meanwhile if you have a change in mind, I can build electron on my windows 10 to verify the fix. Is it possible that this also causes the issue that I cannot set user tasks anymore to the task bar menu of the window?

paulcbetts:
@bpasero Start at `win32prg` and go from there

bpasero:
@paulcbetts ???

paulcbetts:
@bpasero Open up Outlook and look in the GAL for it :)

bpasero:
@paulcbetts ah you are talking about an email alias, I was searching electron source code for "win32prg" to fix this myself ;). Yeah I might have some people to ask, might not get back to me until Redmond wakes up though.

paulcbetts:
Fixing this Properly actually might be not particularly trivial if you have to do it The Hard Way. Basically:

1. Every time a renderer process gets focus (either pragmatically, or via normal means), it needs to signal the browser process with its window ID.
[...]

PR #18046, opened 2019-04-29
fix: explicitly focus window on win.show()

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/18034.

Our documentation for [`win.show`](https://github.com/electron/electron/blob/f901170a4feba092d20d12d85261059af74d5f24/docs/api/browser-window.md#L789) says that this explicitly gives the window focus, and on macOS we implement `Show` with [`unhide`](https://developer.apple.com/documentation/appkit/nsapplication/1428761-unhide?language=objc) and so this rings true.

However, this is not true on Windows, which is implemented in [`NativeWindowViews`](https://github.com/electron/electron/blob/141fb9e7582572d7c5f61b59a9bd2684d86159ee/atom/browser/native_window_views.cc#L376). This PR fixes that issue by explicitly focusing the window with `widget()->Activate()`.

cc @MarshallOfSound @deepak1556 @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue on Windows where calling `.show()` on a BrowserWindow did not focus the window.

### DIFF
--- a/atom/browser/native_window_views.cc
+++ b/atom/browser/native_window_views.cc
@@ -380,6 +380,9 @@ void NativeWindowViews::Show() {
 
   widget()->native_widget_private()->Show(GetRestoredState(), gfx::Rect());
 
+  // explicitly focus the window
+  widget()->Activate();
+
   NotifyWindowShow();
 
 #if defined(USE_X11)


jwheare:
Does this fix #2867?

jwheare:
I've now tested, and this doesn't fix #2867, see comments there for my workaround.

softremake:
> I've now tested, and this doesn't fix #2867, see comments there for my workaround.

Thanks @jwheare 
However I've tried remote.app.emit('activate'); in electron 5 and it didn't help... 
I'm using nodemon and I want to see the main window after changing source code, but on Windows 10 the only thing I could find working is setting alwaysOnTop = true on window creation and then use setAlwaysOnTop(false) after page is rendered (through IPC)...

jwheare:
@softremake ah that’s just a custom event I listen to in the main process to show and focus the main window https://github.com/irccloud/irccloud-desktop/blob/694923a20efd42bcc09943c8167c8fcffb3f0a01/app/main.js#L211

liuyike98:
still have the problem in electron 16.0.2, when i hide a window, then from main process call show(), still loss focus.tried many methods, still doesn't work

PR #18074, opened 2019-04-30
build: fix master build by updating the yarn.lock after race merge

MarshallOfSound:
There was a small race here between my yarn PR landing and another dependency being bumped so when they both got merged the lock file was out of date.  This is my bad, should have rebased before merging.  This will fix it up for good now though 👍 

Notes: no-notes


MarshallOfSound:
As soon as a checkout passes I'm gonna merge this is green-ify master 👍

PR #18084, opened 2019-05-01
feat: migrate protocol module to NetworkService (Part 4)

zcbenz:
#### Description of Change

This is part of the changes to reimplement the protocol module with NetworkService API, the new implementation lives in parallel with current implementation and only gets used when `--enable-features=NetworkService` is passed.

This PR implements the `protocol.registerStreamProtocol` with NetworkService APIs, and also refactors the code of `AtomURLLoaderFactory`.

Due to the migration of NetworkService APIs, it is now possible to return different types of responses in one protocol handler, I have added an experimental `protocol.registerProtocol` API to demonstrate the possibility:

```js
protocol.registerProtocol('test', (request, callback) => {
  if (/* stream */) {
    callback('stream', {
      statusCode: 200,
      headers: {
        'content-type': 'text/html'
      },
      data: createStream('<h5>Response</h5>')
    })
  } else if (/* file */) {
    callback('file', '/path/to/file')
  }
})
```

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/net/atom_url_loader_factory.cc
+++ b/atom/browser/net/atom_url_loader_factory.cc
@@ -186,7 +186,7 @@ void AtomURLLoaderFactory::StartLoading(
                        request, std::move(client), traffic_annotation, dict);
       break;
     case ProtocolType::kStream:
-      StartLoadingStream(std::move(client), dict);
+      StartLoadingStream(std::move(loader), std::move(client), dict);
       break;
     case ProtocolType::kFree:
       ProtocolType type;
@@ -306,6 +306,7 @@ void AtomURLLoaderFactory::StartLoadingHttp(
 
 // static
 void AtomURLLoaderFactory::StartLoadingStream(
+    network::mojom::URLLoaderRequest loader,
     network::mojom::URLLoaderClientPtr client,
     const mate::Dictionary& dict) {
   network::ResourceResponseHead head = ToResponseHead(dict);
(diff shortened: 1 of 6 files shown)


atom/browser/net/atom_url_loader_factory.cc
```diff
@@ -243,25 +304,42 @@ void AtomURLLoaderFactory::SendResponseHttp(
...
+
+  mate::Dictionary data = ToDict(dict.isolate(), stream);
+  v8::Local<v8::Value> method;
+  if (!data.Get("on", &method) || !method->IsFunction() ||
+      !data.Get("removeListener", &method) || !method->IsFunction()) {
+    client->OnComplete(network::URLLoaderCompletionStatus(net::ERR_FAILED));
+    return;
+  }
+
+  new NodeStreamLoader(std::move(head), std::move(client), data.isolate(),
```
deepak1556: Is it okay for this class to manage its own lifetime ?

zcbenz: This seems to be the only choice, I couldn't find a way to have Chromium manage its lifetime. `content::FileURLLoader` is also managing its own lifetime:

https://cs.chromium.org/chromium/src/content/browser/file_url_loader_factory.cc?type=cs&g=0&l=341

deepak1556: Oh cool, in that case should we check if `producer_` and `client_` handles are bound before accessing the methods when we receive data from JS land ?

zcbenz: `content::FileURLLoader` does not check bound for every call but it deletes itself when connection error happens, I have added the same for `NodeStreamLoader`.

PR #18085, opened 2019-05-01
chore: retry octokit calls several times on fail

codebytere:
#### Description of Change

Retry any of the octokit requests in the notes generator 3 times before actually failing the script.

cc @MarshallOfSound 

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/script/release-notes/notes.js
+++ b/script/release-notes/notes.js
@@ -9,6 +9,9 @@ const { GitProcess } = require('dugite')
 const octokit = require('@octokit/rest')()
 const semver = require('semver')
 
+const MAX_FAIL_COUNT = 3
+const CHECK_INTERVAL = 5000
+
 const CACHE_DIR = path.resolve(__dirname, '.cache')
 const NO_NOTES = 'No notes'
 const FOLLOW_REPOS = [ 'electron/electron', 'electron/libchromiumcontent', 'electron/node' ]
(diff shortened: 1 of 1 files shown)


script/release-notes/notes.js
```diff
@@ -300,33 +303,50 @@ const checkCache = async (name, operation) => {
...
-    try {
-      return await octokit.pulls.get({ number, owner, repo })
-    } catch (error) {
-      // Silently eat 404s.
-      // We can get a bad pull number if someone manually lists
-      // an issue number in PR number notation, e.g. 'fix: foo (#123)'
-      if (error.code !== 404) {
-        throw error
-      }
+    let failCount = 0
```
jkleinsc: @codebytere I could be missing something here, but I don't think this retry is going to work the way you are expecting because there is no await happening in the async function.

Also, setTimeout is not async friendly unless you wrap it, eg one of the solutions here: https://stackoverflow.com/questions/33289726/combination-of-async-function-await-settimeout

script/release-notes/notes.js
```diff
@@ -300,33 +303,50 @@ const checkCache = async (name, operation) => {
...
+    const tryGetComments = () => {
+      octokit.issues.listComments({ number, owner, repo, per_page: 100 })
```
jkleinsc: Same comment above about not having an await in play.  This video here explains a way to do retries on an async function: https://www.youtube.com/watch?v=7N8wbrCYqeM

script/release-notes/notes.js
```diff
@@ -9,6 +9,9 @@ const { GitProcess } = require('dugite')
 const octokit = require('@octokit/rest')()
 const semver = require('semver')
 
+const MAX_FAIL_COUNT = 3
+const CHECK_INTERVAL = 30000
```
MarshallOfSound: 30 seconds is a decent pause.  Can we do 5 seconds?

script/release-notes/notes.js
```diff
@@ -298,35 +301,40 @@ const checkCache = async (name, operation) => {
...
+      lastError = error
     }
+  }
+  throw lastError
+}
+
+const getPullRequest = async (number, owner, repo) => {
+  const name = `${owner}-${repo}-pull-${number}`
+  return checkCache(name, async () => {
+    await runRetryable(octokit.pulls.get({
```
MarshallOfSound: We've lost the "Silently eat 404" code

script/release-notes/notes.js
```diff
@@ -298,35 +301,40 @@ const checkCache = async (name, operation) => {
...
-      return await octokit.issues.listComments({ number, owner, repo, per_page: 100 })
-    } catch (error) {
-      // Silently eat 404s.
-      // We can get a bad pull number if someone manually lists
-      // an issue number in PR number notation, e.g. 'fix: foo (#123)'
-      if (error.code !== 404) {
-        throw error
-      }
-    }
+    await runRetryable(octokit.issues.listComments({
```
MarshallOfSound: Same here, we need to ignore / absorb 404 responses

script/release-notes/notes.js
```diff
@@ -298,35 +301,41 @@ const checkCache = async (name, operation) => {
...
     }
+  }
+  // Silently eat 404s.
+  if (lastError.status !== 404) throw lastError
+}
+
+const getPullRequest = async (number, owner, repo) => {
+  const name = `${owner}-${repo}-pull-${number}`
+  return checkCache(name, async () => {
+    await runRetryable(octokit.pulls.get({
```
MarshallOfSound: This needs to be `return`'ed
```suggestion
    return await runRetryable(octokit.pulls.get({
```

script/release-notes/notes.js
```diff
@@ -298,35 +301,41 @@ const checkCache = async (name, operation) => {
...
-      return await octokit.issues.listComments({ number, owner, repo, per_page: 100 })
-    } catch (error) {
-      // Silently eat 404s.
-      // We can get a bad pull number if someone manually lists
-      // an issue number in PR number notation, e.g. 'fix: foo (#123)'
-      if (error.code !== 404) {
-        throw error
-      }
-    }
+    await runRetryable(octokit.issues.listComments({
```
MarshallOfSound: Here too
```suggestion
    return await runRetryable(octokit.issues.listComments({
```

script/release-notes/notes.js
```diff
@@ -298,35 +301,41 @@ const checkCache = async (name, operation) => {
...
-      return await octokit.issues.listComments({ number, owner, repo, per_page: 100 })
-    } catch (error) {
-      // Silently eat 404s.
-      // We can get a bad pull number if someone manually lists
-      // an issue number in PR number notation, e.g. 'fix: foo (#123)'
-      if (error.code !== 404) {
-        throw error
-      }
-    }
+    return await runRetryable(octokit.issues.listComments({
```
jkleinsc: Same as above
```suggestion
    return runRetryable(octokit.issues.listComments({
```

PR #18086, opened 2019-05-01
fix: correctly parse default_app path on windows

MarshallOfSound:
Noticed this on my Windows machine while verifying the `6.0.0-beta.1` release.  On windows, `pathname` looks like `/c:/a/b/c` (that preceding `/` breaks things).  No notes as it's a `default_app` UI change, app devs don't need to know.

This is the bug -->
image

Notes: no-notes

### DIFF
--- a/default_app/default_app.ts
+++ b/default_app/default_app.ts
@@ -30,7 +30,11 @@ function isTrustedSender (webContents: Electron.WebContents) {
   }
 
   const parsedUrl = new URL(webContents.getURL())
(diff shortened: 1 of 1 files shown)


miniak:
@MarshallOfSound can it be related to #18091?

MarshallOfSound:
@miniak The issue here isn't asar or realpath.  We can't call `realpath` on the result of `URL().pathname` as `URL` does not return a valid Windows path.  It is very much UNIX style.

PR #18091, opened 2019-05-01
revert: only allow bundled preload scripts

miniak:
#### Description of Change
Revert #17308. There are issues with `electron-mocha`.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Reverted change to only allow preload scripts to be loaded from application root due to issues with `electron-mocha`.

/cc @electron/wg-security


miniak:
The actual issue was #18092

zcbenz:
Would like to hear more voices from @electron/wg-security before merging.

miniak:
@zcbenz I would like to re-land this change later after I resolve all the issues.

alitaheri:
Will this be backported to 6.x-beta?

Also @miniak please keep in mind that in the development stage this is a real pain. Allow disabling this behavior somehow, like an environment variable or a function on the `app` module.

I understand that this is an important security feature. but it's not needed during development. Since you're using `realpath` even symbolic links can't be used as a workaround. please allow some safe way of disabling this, at least during development. thanks 🙏

MarshallOfSound:
@alitaheri What is your development use case for loading a preload script outside of your app path?  In development the app path should be the directory containing your apps `package.json`.  If that doesn't include your preload script I'm not sure what your set up is?

alitaheri:
@MarshallOfSound Our app is pretty complex, we use typescript and have many outputs. but a single preload to be used for all the processes. for production we copy everything under the same folder, but in development the main processes are in different folders than the preload itself. it was all working fine. but this change broke it. right now we copy the preload output to the mainProcess folders too but would like to be able to disable this during development.

alitaheri:
I have worked around this by always copying preload after bundling to where the main processes are. this isn't ideal but gets the job done. though I would still prefer an option to disable this check. thank you

miniak:
@alitaheri in the new version, I will probably only apply this for bundled apps, which is basically production only. Would that work for you?

alitaheri:
@miniak Yes absolutely, thank you 👍

PR #18088, opened 2019-05-01
chore: clang-format

miniak:
#### Description of Change
Makes sure that all sources are properly formatted by clang-format.

#### Checklist
- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

#### Release Notes
Notes: no-notes

PR #18100, opened 2019-05-01
chore: add native_mate to lint:cpp

miniak:
#### Description of Change
Adds `native_mate` to `lint:cpp` and fixes all issues found. Follow up to #18088.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/script/lint.js
+++ b/script/lint.js
@@ -43,7 +43,7 @@ function spawnAndCheckExitCode (cmd, args, opts) {
 
 const LINTERS = [ {
   key: 'c++',
-  roots: ['atom'],
+  roots: ['atom', 'native_mate'],
   test: filename => filename.endsWith('.cc') || filename.endsWith('.h'),
   run: (opts, filenames) => {
     if (opts.fix) {
--- a/native_mate/native_mate/arguments.h
+++ b/native_mate/native_mate/arguments.h
@@ -2,8 +2,11 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE.chromium file.
 
-#ifndef NATIVE_MATE_ARGUMENTS_H_
-#define NATIVE_MATE_ARGUMENTS_H_
+#ifndef NATIVE_MATE_NATIVE_MATE_ARGUMENTS_H_
+#define NATIVE_MATE_NATIVE_MATE_ARGUMENTS_H_
+
+#include <string>
+#include <vector>
 
 #include "base/macros.h"
 #include "native_mate/converter.h"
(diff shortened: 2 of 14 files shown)


script/lint.js
```diff
@@ -43,7 +43,7 @@ function spawnAndCheckExitCode (cmd, args, opts) {
 
 const LINTERS = [ {
   key: 'c++',
-  roots: ['atom'],
+  roots: ['atom', 'native_mate'],
   test: filename => filename.endsWith('.cc') || filename.endsWith('.h'),
```
deepak1556: `clang-format` can handle `.mm` files, lets try adding it.

miniak: however `cpplint.py` does not like it
```
Ignoring atom/app/atom_library_main.mm; not a valid file name (cc, h, cpp, cu, cuh)
etc.
```

miniak: added run clang-format on Objective-C files, but not cpplint.py

miniak: actually, I will do that in another PR. I can enable limited `cpplint.py`

atom/browser/api/atom_api_content_tracing.cc
```diff
@@ -116,7 +116,7 @@ void OnTraceBufferUsageAvailable(atom::util::Promise promise,
                                  size_t approximate_count) {
   mate::Dictionary dict = mate::Dictionary::CreateEmpty(promise.isolate());
   dict.Set("percentage", percent_full);
-  dict.Set("value", approximate_count);
+  dict.Set("value", static_cast<uint64_t>(approximate_count));
```
nornagon: I'm curious what lint rule triggered this, because the old version seems fine to me here.

codebytere:
@nornagon the subdirectorying was originally done to preserve history; it may be removable now but i'm not sure! cc @MarshallOfSound

atom/browser/api/atom_api_content_tracing.cc
miniak: there was a converter from `unsigned long`, which was used for `size_t`

miniak: I wasn't sure what to do here. I can keep the converter and just silence the warnings about `unsigned long`

script/lint.js
miniak: #18104

atom/browser/api/atom_api_content_tracing.cc
miniak: I've reverted this change and kept the `unsigned long` converter as there were some build errors on Windows. Added `// NOLINT` there for now

PR #18103, opened 2019-05-01
build: move to the new docs parser

MarshallOfSound:
Switched from:
* `electron-docs-linter` --> `@electron/docs-parser`
* `electron-typescript-definitions` --> `@electron/typescript-definitions`

This makes our definition file **way** better, we don't lose any information in the docs.  Noticeably this means markdown, links and code blocks are persisted in the TS definitions.  It is also _much_ better under the hood so future improvements / fixes will be much easier.  E.g. `readonly` support is in a branch somewhere :)

I'm planning on writing a blog post at some point detailing how `@electron/docs-parser` now works.  But from a thousand foot view.  `doc-linter` used to parse the markdown in HTML and then parse the docs as a DOM structure with `jquery` and selector querying.  This had a few issues the biggest being it was very hard to semantically now what was "inside" a heading when the generated HTML just looks like this:

```html
<h3>Method()</h3>
<p>stuff</p>
<p>more stuff</p>
<p>stuff</p>
<h3>Next Method()</h3>
```

The new `docs-parser` module parses the markdown file into a markdown AST which **does** know where things start and end so it makes parsing things _way_ easier and it means we can losslessly maintain the markdown in each block without worrying about HTML entities or escaping or anything like that.

This PR also fixes a bunch of issues with a current documentation, highlights below.
* `app.commandLine.*` and `app.dock.*` are now in separate "module" files.  Lumping them in with `app.md` over-complicated the parser and the read experience.
* Fixed a bunch of `Event` documentation not having `Returns:` (the magic keyword to document the event properties)
* Added a bunch of `- ` and `: ` and `\n` so that documentation for properties in deep objects are correctly parsed.

Notes: Switched to a new Typescript Definitions generator.  This means that some interface names may have changed, if your Typescript build is failing this is the cause.


MarshallOfSound:
So the diff in the `electron.d.ts` file is literally Too Big for the `Artifact Comparison` check to deal with.  So here is the `diff`:

https://gist.github.com/MarshallOfSound/14fdce4022c3453eeb24c7520f0c8356

And here is the actual new file:

https://gist.github.com/MarshallOfSound/7773cc6501a303fa6dfb62792d477ba1

docs/api/app.md
```diff
@@ -1342,7 +1242,7 @@ This API must be called after the `ready` event is emitted.
 
 ### `app.applicationMenu`
 
-A `Menu` property that return [`Menu`](menu.md) if one has been set and `null` otherwise.
+A `Menu | null` property that return [`Menu`](menu.md) if one has been set and `null` otherwise.
```
malept:
```suggestion
A `Menu | null` property that returns [`Menu`](menu.md) if one has been set and `null` otherwise.
```

docs/api/app.md
```diff
@@ -1354,14 +1254,15 @@ On macOS, setting this with any nonzero integer shows on the dock icon. On Linux
 **Note:** Unity launcher requires the existence of a `.desktop` file to work,
 for more information please read [Desktop Environment Integration][unity-requirement].
 
-### `app.userAgentFallback`
+### `app.commandLine`
 
-A `String` which is the user agent string Electron will use as a global fallback.
+A [`CommandLine`](./command-line.md) object that allows you to read and manipulate the
+command line arguments that chromium uses.
```
malept:
```suggestion
command line arguments that Chromium uses.
```

docs/api/app.md
```diff
@@ -1354,14 +1254,15 @@ On macOS, setting this with any nonzero integer shows on the dock icon. On Linux
...
 
-This is the user agent that will be used when no user agent is set at the
-`webContents` or `session` level.  Useful for ensuring your entire
-app has the same user agent.  Set to a custom value as early as possible
-in your apps initialization to ensure that your overridden value is used.
+### `app.dock` _macOS_
+
+A [`Dock`](./dock.md) object that allows you to perform actions on your app icon in the users
```
malept:
```suggestion
A [`Dock`](./dock.md) object that allows you to perform actions on your app icon in the user's
```

docs/api/app.md
```diff
@@ -1388,3 +1289,12 @@ Usually the `name` field of `package.json` is a short lowercased name, according
 to the npm modules spec. You should usually also specify a `productName`
 field, which is your application's full capitalized name, and which will be
 preferred over `name` by Electron.
+
+### `app.userAgentFallback`
+
+A `String` which is the user agent string Electron will use as a global fallback.
+
+This is the user agent that will be used when no user agent is set at the
+`webContents` or `session` level.  Useful for ensuring your entire
```
malept:
```suggestion
`webContents` or `session` level.  It is useful for ensuring that your entire
```

docs/api/app.md
```diff
@@ -1388,3 +1289,12 @@ Usually the `name` field of `package.json` is a short lowercased name, according
...
+app has the same user agent.  Set to a custom value as early as possible
+in your apps initialization to ensure that your overridden value is used.
```
malept:
```suggestion
in your app's initialization to ensure that your overridden value is used.
```

docs/api/command-line.md
```diff
@@ -0,0 +1,50 @@
+## Class: CommandLine
+
+> Manipulate the command line arguments for your app that Chromium reads
+
+Process: [Main](../glossary.md#main-process)
```
malept: Also link to https://electronjs.org/docs/api/chrome-command-line-switches somewhere in this block?

docs/api/debugger.md
```diff
@@ -32,6 +32,32 @@ win.webContents.debugger.on('message', (event, method, params) => {
...
+### Instance Events
+
+#### Event: 'detach'
+
+Returns:
+
+* `event` Event
+* `reason` String - Reason for detaching debugger.
+
+Emitted when debugging session is terminated. This happens either when
```
malept:
```suggestion
Emitted when the debugging session is terminated. This happens either when
```

docs/api/debugger.md
```diff
@@ -32,6 +32,32 @@ win.webContents.debugger.on('message', (event, method, params) => {
...
+#### Event: 'message'
+
+Returns:
+
+* `event` Event
+* `method` String - Method name.
+* `params` Object - Event parameters defined by the 'parameters'
+   attribute in the remote debugging protocol.
+
+Emitted whenever debugging target issues instrumentation event.
```
malept:
```suggestion
Emitted whenever the debugging target issues an instrumentation event.
```

docs/api/dialog.md
```diff
@@ -198,7 +198,7 @@ The `filters` specifies an array of file types that can be displayed, see
 
 Returns `Promise<Object>` - Resolve with an object containing the following:
   * `canceled` Boolean - whether or not the dialog was canceled.
-  * `filePath` String (optional) If the dialog is canceled this will be `undefined`.
+  * `filePath` String (optional) - If the dialog is canceled this will be `undefined`.
```
malept:
```suggestion
  * `filePath` String (optional) - If the dialog is canceled, this will be `undefined`.
```

docs/api/menu-item.md
```diff
@@ -147,15 +147,15 @@ item's submenu, if present.
 
 #### `menuItem.type`
 
-A `String` indicating the type of the item.
+A `String` indicating the type of the item. Can be `normal`, `separator`, `submenu`, `checkbox` or `radio`.
```
malept: Oxford comma?
```suggestion
A `String` indicating the type of the item. Can be `normal`, `separator`, `submenu`, `checkbox`, or `radio`.
```

docs/api/process.md
```diff
@@ -206,12 +206,6 @@ Returns `String` - The version of the host operating system.
 
 Examples:
```
malept: ✂️ too?

docs/api/structures/trace-categories-and-options.md
```diff
@@ -1,10 +1,6 @@
 # TraceCategoriesAndOptions Object
 
-* `categoryFilter` String – is a filter to control what category groups
-  should be traced. A filter can have an optional `-` prefix to exclude
-  category groups that contain a matching category. Having both included
-  and excluded category patterns in the same list is not supported. Examples:
-  `test_MyTest*`, `test_MyTest*,test_OtherStuff`, `-excluded_category1,-excluded_category2`.
+* `categoryFilter` String
```
malept: Not knowing anything about this, did its docs need to be completely rewritten?

docs/api/web-contents.md
```diff
@@ -1525,7 +1525,7 @@ End subscribing for frame presentation events.
 #### `contents.startDrag(item)`
 
 * `item` Object
-  * `file` String or `files` Array - The path(s) to the file(s) being dragged.
+  * `file` String[] | String - The path(s) to the file(s) being dragged.
```
malept: Parentheses not needed for `String[] | String`? (As opposed to `(String | null)` above.)

docs/api/webview-tag.md
```diff
@@ -227,7 +227,7 @@ windows. Popups are disabled by default.
 <webview src="https://github.com" webpreferences="allowRunningInsecureContent, javascript=no"></webview>
 ```
 
-A list of strings which specifies the web preferences to be set on the webview, separated by `,`.
+A `String` which is a comma seperated list of strings which specifies the web preferences to be set on the webview.
```
malept:
```suggestion
A `String` which is a comma separated list of strings which specifies the web preferences to be set on the webview.
```

docs/api/menu-item.md
MarshallOfSound: I don't think this will play nicely with the parser

docs/api/structures/trace-categories-and-options.md
MarshallOfSound: No, good catch I must have deleted this by accident at some point

docs/api/process.md
MarshallOfSound: Woops, not sure why I deleted that

docs/api/web-contents.md
MarshallOfSound: Not required, the new parser doesn't need them 👍

docs/api/menu-item.md
malept: That's unfortunate.

docs/api/command-line.md
```diff
@@ -0,0 +1,54 @@
...
+
+The following example shows how to check if the `--disable-gpu` flag is set.
+
+```javascript
+const { app } = require('electron')
+app.commandLine.hasSwitch('disable-gpu')
+```
+
+For more information on what kinds of flags and switches you can use check
```
malept: Minor nit
```suggestion
For more information on what kinds of flags and switches you can use, check
```

PR #18107, opened 2019-05-01
docs: clarify clipboard type options

codebytere:
#### Description of Change

Clarify clipboard type options and that the `selection` type is linux-only.

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/clipboard.md
+++ b/docs/api/clipboard.md
@@ -11,7 +11,7 @@ const { clipboard } = require('electron')
 clipboard.writeText('Example String')
 ```
 
-On X Window systems, there is also a selection clipboard. To manipulate it
+On Linux, there is also a `selection` clipboard. To manipulate it
 you need to pass `selection` to each method:
 
 ```javascript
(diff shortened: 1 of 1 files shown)


docs/api/clipboard.md
```diff
@@ -28,37 +28,37 @@ The `clipboard` module has the following methods:
 
 ### `clipboard.readText([type])`
 
-* `type` String (optional)
+* `type` String (optional) - Can be 'selection' (only available on Linux) or 'clipboard'
```
MarshallOfSound: In order for TS defs to pick this up:

```suggestion
* `type` String (optional) - Can be 'selection' or 'clipboard'.  'selection' is only available on Linux.
```

MarshallOfSound: Same all the way down

codebytere: wfm, will update

jkleinsc:
Merging as CI failure unrelated to PR.

PR #18110, opened 2019-05-02
feat: allow setting of global fallback user agent

MarshallOfSound:
Backport of #18016 

See that PR for details.  Approved for `4.2.0` backport by @electron/wg-releases on 01/05/19

Notes: Added new `app.userAgentFallback` property to allow setting the apps useragent globally

### DIFF
--- a/atom/app/atom_content_client.cc
+++ b/atom/app/atom_content_client.cc
@@ -7,6 +7,7 @@
 #include <string>
 #include <vector>
 
+#include "atom/browser/browser.h"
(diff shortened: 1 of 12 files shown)


MarshallOfSound:
This is slightly different to the `master` PR as we have to send the `user-agent` over the process arguments in order to support the override for service workers.  In `4-0-x` the service worker UA is initialized in the renderer.  In `master` it is initialized in the browser process so we don't have this work around.

Issue #13218, opened 2018-06-11
`base::GetProcId` is deprecated


alexeykuzmin:
As of Chromium `63.0.3239.150`

[base/process/process_handle.h#84](https://chromium.googlesource.com/chromium/src/+/63.0.3239.150/base/process/process_handle.h#84):
> New code should be using `Process::Pid()` instead.

miniak:
`Process::Pid()` is not a direct replacement

alexeykuzmin:
@miniak That's what the code says. I wish it were a search-and-replace task, but yeah, it would require slightly more effort.

PR #18111, opened 2019-05-02
chore: remove use of base::GetProcId

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/13218.

Removes use of `base::GetProcId` in favor of `Process::Pid()`.

cc @deepak1556 @miniak @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


codebytere:
@deepak1556 i think they're going to stop exposing it an only expose it _though_ `Process.Pid()`

deepak1556:
> They're going to stop exposing it

Is that stated somewhere, I am not able to find it in that deprecation note. If the api is going to be replaced by something else, I would like to see the alternative implementation before switching to it, at the moment its the same api thats being called.

codebytere:
Closing until there's an active replacement.

PR #18124, opened 2019-05-02
refactor: make templateImage a property on nativeImage

codebytere:
#### Description of Change

Update `nativeImage.templateImage` to be a property on the `nativeImage` prototype.
 
#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Updates `nativeImage.templateImage` to be a property on the `nativeImage` prototype.


docs/api/native-image.md
```diff
@@ -324,3 +328,9 @@ to explicitly add different scale factor representations to an image. This
 can be called on empty images.
 
 [buffer]: https://nodejs.org/api/buffer.html#buffer_class_buffer
+
+## Properties
+
+### `nativeImage.templateImage`
```
MarshallOfSound:
```suggestion
### `nativeImage.isTemplateImage`
```

docs/api/native-image.md
```diff
@@ -324,3 +328,9 @@ to explicitly add different scale factor representations to an image. This
...
+
+A `Boolean` property that determines whether the image is a template image.
```
MarshallOfSound:
```suggestion
A `Boolean` property that determines whether the image is considered a [template image](https://developer.apple.com/documentation/appkit/nsimage/1520017-template).

Please note that this property only has an effect on macOS.
```

spec/api-native-image-spec.js
```diff
@@ -105,6 +105,14 @@ describe('nativeImage module', () => {
     return matchingImage
   }
 
+  describe('templateImage property', () => {
+    it('returns whether the image is a template image', () => {
```
MarshallOfSound: Can we add tests for the behavior here.  i.e. if you do `nativeImage.createFromPath('MyImage_Template.png`)` then it should be `true`.

If you do `nativeImage.createFromPath('MyImage.png')` it should be `false`.

Only on macOs though

MarshallOfSound: This tests the setter, but not the getter returns the correct value on the first try.  

```js
nativeImage.createFromPath('MyImage_Template.png') === true
```

Is a test we want here as well

atom/common/api/atom_api_native_image.cc
```diff
@@ -638,8 +638,10 @@ void NativeImage::BuildPrototype(v8::Isolate* isolate,
       .SetMethod("toDataURL", &NativeImage::ToDataURL)
       .SetMethod("isEmpty", &NativeImage::IsEmpty)
       .SetMethod("getSize", &NativeImage::GetSize)
-      .SetMethod("setTemplateImage", &NativeImage::SetTemplateImage)
-      .SetMethod("isTemplateImage", &NativeImage::IsTemplateImage)
+      .SetMethod("_setTemplateImage", &NativeImage::SetTemplateImage)
+      .SetMethod("_isTemplateImage", &NativeImage::IsTemplateImage)
+      .SetProperty("templateImage", &NativeImage::IsTemplateImage,
```
MarshallOfSound: I'm still not a huge fan of this property name, I get we can't call it `isTemplateImage` for deprecation reasons.  But I still feel `nativeImage.templateImage` as a boolean doesn't quite fit.

Maybe `nativeImage.isMacTemplateImage`?

codebytere: hmm yes! this is an unexpectedly annoying naming clash but i definitely agree that `templateImage` might lead the user to make incorrect assumptions about what exactly the property should be set to.  

let's go with `nativeImage.isMacTemplateImage` for now 🤔

docs/api/native-image.md
```diff
@@ -324,3 +327,11 @@ to explicitly add different scale factor representations to an image. This
 can be called on empty images.
 
 [buffer]: https://nodejs.org/api/buffer.html#buffer_class_buffer
+
+## Properties
+
+### `nativeImage.isMacTemplateImage`
```
MarshallOfSound:
```suggestion
### `nativeImage.isMacTemplateImage` _macOS_
```

☝️ is how we do platform support

PR #18129, opened 2019-05-02
refactor: nws13n: session.allowNTLMCredentialsForDomains

nornagon:
#### Description of Change
Refactors `session.allowNTLMCredentialsForDomains` to use the network service.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/api/atom_api_session.cc
+++ b/atom/browser/api/atom_api_session.cc
@@ -22,6 +22,7 @@
 #include "atom/browser/browser.h"
 #include "atom/browser/media/media_device_id_salt.h"
 #include "atom/browser/net/atom_cert_verifier.h"
+#include "atom/browser/net/system_network_context_manager.h"
 #include "atom/browser/session_preferences.h"
 #include "atom/common/native_mate_converters/callback.h"
 #include "atom/common/native_mate_converters/content_converter.h"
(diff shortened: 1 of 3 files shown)


MarshallOfSound:
Question, does this have to work for both network service and non-network service? This looks like it removes the old code path with the network service still disabled

deepak1556:
It will work with both versions, in non network service mode, everything will stub to the network layer in the browser process.

atom/browser/api/atom_api_session.cc
```diff
@@ -503,11 +492,13 @@ v8::Local<v8::Promise> Session::ClearAuthCache() {
...
-      FROM_HERE, {BrowserThread::IO},
-      base::BindOnce(&AllowNTLMCredentialsForDomainsInIO,
-                     WrapRefCounted(browser_context_->GetRequestContext()),
-                     domains));
+  auto* command_line = base::CommandLine::ForCurrentProcess();
+
+  auto auth_params = ::network::mojom::HttpAuthDynamicParams::New();
+  auth_params->server_whitelist = domains;
+  auth_params->delegate_whitelist = command_line->GetSwitchValueASCII(
+      atom::switches::kAuthNegotiateDelegateWhitelist);
```
deepak1556: Do we also want to preserve the value of `enable_negotiate_port` here ? Since the default value from the service is not what we want.

nornagon: great catch, yes.

PR #18141, opened 2019-05-03
feat: add a tabbingIdentifier property on windows

### LINKED ISSUE #18138: Allow to retrieve window tabbingIdentifier at runtime
Electron supports macOS sierra native tabs when creating windows, however in order to properly restore an arrangement of multiple groups of tabs, we would need to get this information at runtime before closing the application.

MarshallOfSound:
Closes #18138

Notes: Added a new `tabbingIdentifier` property on `BrowserWindow`'s to allow reading the current identifier at runtime.

### DIFF
--- a/atom/browser/api/atom_api_top_level_window.cc
+++ b/atom/browser/api/atom_api_top_level_window.cc
@@ -1022,6 +1022,16 @@ void TopLevelWindow::RemoveFromParentChildWindows() {
   parent->child_windows_.Remove(weak_map_id());
 }
 
+v8::Local<v8::Value> TopLevelWindow::GetTabbingIdentifier() {
+#if defined(OS_MACOSX)
+  std::string identifier = window_->GetTabbingIdentifier();
+  if (!identifier.empty()) {
+    return mate::ConvertToV8(isolate(), identifier);
+  }
+#endif
+  return v8::Undefined(isolate());
+}
+
 // static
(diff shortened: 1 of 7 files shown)


atom/browser/api/atom_api_top_level_window.cc
```diff
@@ -1174,7 +1184,8 @@ void TopLevelWindow::BuildPrototype(v8::Isolate* isolate,
       .SetMethod("setThumbnailToolTip", &TopLevelWindow::SetThumbnailToolTip)
       .SetMethod("setAppDetails", &TopLevelWindow::SetAppDetails)
 #endif
-      .SetProperty("id", &TopLevelWindow::GetID);
+      .SetProperty("id", &TopLevelWindow::GetID)
+      .SetProperty("tabbingIdentifier", &TopLevelWindow::GetTabbingIdentifier);
```
codebytere: i think there's a mismatch here in platform defines: it's guarded in the header but not in the impl. I think i'd just recommend exposing it for mac-only as opposed exposing it across platforms and having it no-op where not applicable

miniak:
@MarshallOfSound can you please rebase?

zcbenz:
@MarshallOfSound Do you still have plan to work on this?

bpasero:
Why was this closed actually?

PR #18170, opened 2019-05-06
feat: migrate protocol module to NetworkService (Part 5)

zcbenz:
#### Description of Change

This PR fixes some failing protocol module tests when NetworkService is enabled.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/net/atom_url_loader_factory.cc
+++ b/atom/browser/net/atom_url_loader_factory.cc
@@ -19,6 +19,7 @@
 #include "content/public/browser/browser_thread.h"
 #include "content/public/browser/file_url_loader.h"
 #include "content/public/browser/storage_partition.h"
+#include "mojo/public/cpp/system/string_data_pipe_producer.h"
 #include "net/base/filename_util.h"
 #include "net/http/http_status_code.h"
 #include "services/network/public/cpp/url_loader_completion_status.h"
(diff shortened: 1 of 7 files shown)


atom/browser/net/atom_url_loader_factory.cc
```diff
@@ -215,8 +237,9 @@ void AtomURLLoaderFactory::StartLoadingBuffer(
     return;
   }
 
-  SendContents(std::move(client), ToResponseHead(dict),
-               node::Buffer::Data(buffer), node::Buffer::Length(buffer));
+  SendContents(
+      std::move(client), ToResponseHead(dict),
+      std::string(node::Buffer::Data(buffer), node::Buffer::Length(buffer)));
```
deepak1556: Is it okay to convert this into `std::string` and transmit ? Can we reuse `WriteDataToProducerHandle` and provide the buffer data to mojo pipe ?

atom/browser/net/atom_url_loader_factory.cc
```diff
@@ -341,21 +363,32 @@ void AtomURLLoaderFactory::StartLoadingStream(
...
     return;
   }
 
-  head.headers->AddHeader(kCORSHeader);
-  client->OnReceiveResponse(head);
-  client->OnStartLoadingResponseBody(std::move(pipe.consumer_handle));
-  client->OnComplete(network::URLLoaderCompletionStatus(net::OK));
+  client->OnStartLoadingResponseBody(std::move(consumer));
+
+  auto write_data = std::make_unique<WriteData>();
```
deepak1556: Can we simplify the self ownership of this struct `WriteData` ? I am thinking maintaining a map of `client` and `WriteData` in `AtomURLLoaderFactory` to manage the lifetime. Thoughts ?

atom/browser/net/atom_url_loader_factory.cc
zcbenz: Using `std::string` to store binary data is a safe pattern and commonly used in Chromium, its network classes like `DataURLLoaderFactory` are also using `std::string` to store binary data.

The `StringDataPipeProducer` writes the data asynchronously and needs the original data to be available until the write is done, copying the buffer to a string is the easiest implementation. If we use `WriteDataToProducerHandle` directly we have to handle the case when one write can not write all data. 

Our old `URLRequestBufferJob` implementation copies the data too so there is no performance loss, it is possible to save the copy completely with a few hacks, and we can do it in future.

atom/browser/net/atom_url_loader_factory.cc
zcbenz: This pattern is giving the ownership of `WriteData` to the callback of `producer->Write`, and it destructs automatically when the write is done, so we don't need to do manual memory management.

The code here is copied from `DataURLLoaderFactory`.

PR #18156, opened 2019-05-04
docs: precise nodeIntegrationInSubFrames and node integration for nativeWindowOpen

alexstrat:
#### Description of Change
In the documentation, precise that node integration is enabled in natively opened windows if `nodeIntegrationInSubFrames` is true

#### Checklist

- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: no-notes

PR #17318, opened 2019-03-10
chore: increase security of default_app

miniak:
#### Description of Change
- enable `sandbox`
- disable `remote` module
- disable creation of new windows, open in default browser instead
- apply `session.setPermissionRequestHandler()`
- load SVGs via `fetch()` instead of `fs.readFileSync()`

/cc @electron/wg-security

#### Checklist
- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes
Notes: Enabled `sandbox` and disabled `remote` module in default_app.

PR #18173, opened 2019-05-06
fix: lost window.opener after cross-origin navigation

### LINKED ISSUE #18032: window.opener opened from webview is null after cross site navigation
### Preflight Checklist

### Issue Details

* **Electron Version:**
`5.0.0`
* **Operating System:**
` macOS 10.14.4`
`4.0.4`

### Expected Behavior

Open a Window with domain A via `window.open` et navigate to domain B in the opened window should keep the `window.opener` reference.

At the business logic level, sign-in with Google Auth from a web frame doesn't work. The callback on final application domain can't post a message to the window opener. 

### Actual Behavior

The `window.opener` is null after cross-site navigation.

### To Reproduce

```js
// main.js

function createWindow() {
        webPreferences: {
          nodeIntegration: true,
          webviewTag: true,
          contextIsolation: false,
        }
  });

  mainWindow.loadFile('index.html')
}

app.on('ready', createWindow)
app.on('window-all-closed', app.quit);
```

```html
  <webview src="https://mail.google.com" allowpopups></webview>
```

- Call `window.open('https://google.com')` from the web view dev tools
- The `window.opener` is not null
- Go to Github with `window.location  = 'https://github.com'`; the `window.opener` is null

### Additional Information

Similar issues have been reported: https://github.com/electron/electron/issues/17869,
https://github.com/electron/electron/pull/15821, https://github.com/electron/electron/issues/8100

Also preloads doesn't load in the same origin context for the opened window.
Seems linked to https://github.com/electron/electron/issues/17989:

- Web view (`myapp.com`) open `auth.google.com`, `window.opener` is null and preloads are loaded
- Web view (`myapp.com`) open `auth.myapp.com`, `window.opener` is set but preloads are not loaded

More details [here](https://github.com/getstation/electron-chrome-extension/pull/47#issuecomment-487856527)
History: https://github.com/electron/electron/issues/11471#issuecomment-352627083

alexstrat:
Fixes #18032

#### Description of Change

With #15821, on cross-origin navigation, we use `SiteInstance::CreateForURL` to [ _get the appropriate `SiteInstance` but **in a new BrowsingInstance**_](https://cs.chromium.org/chromium/src/content/public/browser/site_instance.h?type=cs&q=CreateForURL&sq=package:chromium&g=0&l=160-166). It means we lose the relationship (by browsing instance) between the current and the new `SiteInstance`, and thus the `window.opener` is _null_ in the opened window.

~In this PR, I suggest using [`GetRelatedSiteInstance`](https://cs.chromium.org/chromium/src/content/public/browser/site_instance.h?type=cs&q=CreateForURL&sq=package:chromium&g=0&l=119-124) to keep the relationship.~

In this PR, I introduce and use a `SiteInstance::CreateRelatedSiteInstance` that keeps the browsing instance relationship.

cc @ppontes @deepak1556 

#### Checklist

- [x] `npm test` passes
- [x] tests are [~changed or~ added](https://github.com/electron/electron/blob/master/docs/development/testing.md)

#### Release Notes

Notes: Fixed `window.opener` null after cross-origin navigation

### DIFF
--- a/patches/common/chromium/frame_host_manager.patch
+++ b/patches/common/chromium/frame_host_manager.patch
@@ -7,8 +7,42 @@ Allows embedder to intercept site instances chosen by chromium
 and respond with custom instance. Also allows for us to at-runtime
 enable or disable this patch.
 
+diff --git a/content/browser/browsing_instance.cc b/content/browser/browsing_instance.cc
+index 12e1c5cff95aa6d0a907a249208e23371cf29785..3bc26b7870ff3bf6a69cb1e123fb372f763442ee 100644
+--- a/content/browser/browsing_instance.cc
++++ b/content/browser/browsing_instance.cc
+@@ -79,6 +79,13 @@ scoped_refptr<SiteInstanceImpl> BrowsingInstance::GetSiteInstanceForURL(
+   return instance;
+ }
+ 
++scoped_refptr<SiteInstanceImpl> BrowsingInstance::CreateSiteInstanceForURL(
++    const GURL& url) {
++  scoped_refptr<SiteInstanceImpl> instance = new SiteInstanceImpl(this);
++  instance->SetSite(url);
++  return instance;
++}
++
+ void BrowsingInstance::GetSiteAndLockForURL(const GURL& url,
+                                             bool allow_default_instance,
+                                             GURL* site_url,
+diff --git a/content/browser/browsing_instance.h b/content/browser/browsing_instance.h
+index 775b64a8d20f89845812852a2904a1e6875c2b4a..5235b57bbf44fc7b30ca6943c43a290f07f003bf 100644
+--- a/content/browser/browsing_instance.h
++++ b/content/browser/browsing_instance.h
+@@ -136,6 +136,11 @@ class CONTENT_EXPORT BrowsingInstance final
+       const GURL& url,
+       bool allow_default_instance);
+ 
++  // Create a new SiteInstance for the given URL bound the current
++  // BrowsingInstance.
++  scoped_refptr<SiteInstanceImpl> CreateSiteInstanceForURL(
++      const GURL& url);
++
+   // Adds the given SiteInstance to our map, to ensure that we do not create
+   // another SiteInstance for the same site.
+   void RegisterSiteInstance(SiteInstanceImpl* site_instance);
 diff --git a/content/browser/frame_host/render_frame_host_manager.cc b/content/browser/frame_host/render_frame_host_manager.cc
-index b5301d164138f21ca8ae01abfb153efde51ec324..886b6d3beb3e2d7b13a15af830bea6fb5a567cba 100644
+index b5301d164138f21ca8ae01abfb153efde51ec324..55f87cc788c7eed13494ebe6ea6eb18027a04996 100644
 --- a/content/browser/frame_host/render_frame_host_manager.cc
 +++ b/content/browser/frame_host/render_frame_host_manager.cc
 @@ -2127,6 +2127,20 @@ bool RenderFrameHostManager::InitRenderView(
@@ -56,7 +90,7 @@ index b5301d164138f21ca8ae01abfb153efde51ec324..886b6d3beb3e2d7b13a15af830bea6fb
 +          overriden_site_instance =
 +              candidate_site_instance
 +                  ? candidate_site_instance
-+                  : SiteInstance::CreateForURL(browser_context,
++                  : current_site_instance->CreateRelatedSiteInstance(
 +                                              request.common_params().url);
 +          break;
 +        case ContentBrowserClient::SiteInstanceForNavigationType::FORCE_CURRENT:
@@ -117,6 +151,33 @@ index b5301d164138f21ca8ae01abfb153efde51ec324..886b6d3beb3e2d7b13a15af830bea6fb
    return dest_site_instance;
  }
  
+diff --git a/content/browser/site_instance_impl.cc b/content/browser/site_instance_impl.cc
+index fd184108a7993094c29be3f7ebde658e259ede2c..75aa4a6b7d58a1bebe34efc923953c69348428a9 100644
+--- a/content/browser/site_instance_impl.cc
++++ b/content/browser/site_instance_impl.cc
+@@ -342,6 +342,10 @@ bool SiteInstanceImpl::HasRelatedSiteInstance(const GURL& url) {
+   return browsing_instance_->HasSiteInstance(url);
+ }
+ 
++scoped_refptr<SiteInstance> SiteInstanceImpl::CreateRelatedSiteInstance(const GURL& url) {
++  return browsing_instance_->CreateSiteInstanceForURL(url);
++}
++
+ scoped_refptr<SiteInstance> SiteInstanceImpl::GetRelatedSiteInstance(
+     const GURL& url) {
+   return browsing_instance_->GetSiteInstanceForURL(
+diff --git a/content/browser/site_instance_impl.h b/content/browser/site_instance_impl.h
+index a46901055bdf17b6b0dab14edf753b234dc04a12..29c201b0c95eb0c7a35f47d6f3ab5b48c73dbf15 100644
+--- a/content/browser/site_instance_impl.h
++++ b/content/browser/site_instance_impl.h
+@@ -83,6 +83,7 @@ class CONTENT_EXPORT SiteInstanceImpl final : public SiteInstance,
+   BrowserContext* GetBrowserContext() override;
+   const GURL& GetSiteURL() override;
+   scoped_refptr<SiteInstance> GetRelatedSiteInstance(const GURL& url) override;
++	scoped_refptr<SiteInstance> CreateRelatedSiteInstance(const GURL& url) override;
+   bool IsRelatedSiteInstance(const SiteInstance* instance) override;
+   size_t GetRelatedActiveContentsCount() override;
+   bool RequiresDedicatedProcess() override;
 diff --git a/content/public/browser/content_browser_client.cc b/content/public/browser/content_browser_client.cc
 index 787cd81b26508d3e04956721f0e7cec2f457aa60..8e62a5dd26595411757e03078ed0e44646c47a52 100644
 --- a/content/public/browser/content_browser_client.cc
@@ -187,3 +248,19 @@ index bf9b3a601fc16d5070e4467a258a047f47b059f3..3c35eddc2498157c2b98bab55991d8aa
    // Allows the embedder to set any number of custom BrowserMainParts
    // implementations for the browser startup code. See comments in
    // browser_main_parts.h.
+diff --git a/content/public/browser/site_instance.h b/content/public/browser/site_instance.h
+index a3e880e20e51d988175f0e8e2c42e7f5c1740104..61bbf88265e717934533117efbc2330661e32b11 100644
+--- a/content/public/browser/site_instance.h
++++ b/content/public/browser/site_instance.h
+@@ -121,6 +121,11 @@ class CONTENT_EXPORT SiteInstance : public base::RefCounted<SiteInstance> {
+   //   corresponds to a site URL with the host "example.com".
+   virtual const GURL& GetSiteURL() = 0;
+ 
++	// Create a SiteInstance for the given URL that shares the current
++	// BrowsingInstance.
++	virtual scoped_refptr<SiteInstance> CreateRelatedSiteInstance(
++	    const GURL& url) = 0;
++
+   // Gets a SiteInstance for the given URL that shares the current
+   // BrowsingInstance, creating a new SiteInstance if necessary.  This ensures
+   // that a BrowsingInstance only has one SiteInstance per site, so that pages
--- a/spec/api-browser-window-spec.js
+++ b/spec/api-browser-window-spec.js
@@ -2101,6 +2101,27 @@ describe('BrowserWindow module', () => {
         expect(typeofProcess).to.eql('undefined')
       })
 
+      it('window.opener is not null when window.location is changed to a new origin', async () => {
+        await serveFileFromProtocol('foo', path.join(fixtures, 'api', 'window-open-location-change.html'))
+        await serveFileFromProtocol('bar', path.join(fixtures, 'api', 'window-open-location-final.html'))
+
+        w.destroy()
+        w = new BrowserWindow({
+          show: true,
+          webPreferences: {
+            nativeWindowOpen: true,
+            // test relies on preloads in opened window
+            nodeIntegrationInSubFrames: true
+          }
+        })
+
+        ipcRenderer.send('set-web-preferences-on-next-new-window', w.webContents.id, 'preload', path.join(fixtures, 'api', 'window-open-preload.js'))
+        const p = emittedOnce(ipcMain, 'answer')
+        w.loadFile(path.join(fixtures, 'api', 'window-open-location-open.html'))
+        const [, , , windowOpenerIsNull] = await p
+        expect(windowOpenerIsNull).to.be.false('window.opener is null')
+      })
+
       it('should have nodeIntegration disabled in child windows', async () => {
         w.destroy()
         w = new BrowserWindow({
--- a/spec/api-browser-window-spec.js
+++ b/spec/api-browser-window-spec.js
@@ -9,6 +9,7 @@ const qs = require('querystring')
 const http = require('http')
 const { closeWindow } = require('./window-helpers')
 const { emittedOnce } = require('./events-helpers')
+const { createNetworkSandbox } = require('./network-helper')
 const { ipcRenderer, remote } = require('electron')
 const { app, ipcMain, BrowserWindow, BrowserView, protocol, session, screen, webContents } = remote
 
@@ -1987,9 +1988,12 @@ describe('BrowserWindow module', () => {
     })
 
     describe('nativeWindowOpen option', () => {
+      const networkSandbox = createNetworkSandbox(protocol)
+
       beforeEach(async () => {
-        await serveFileFromProtocol('foo', path.join(fixtures, 'api', 'window-open-location-change.html'))
-        await serveFileFromProtocol('bar', path.join(fixtures, 'api', 'window-open-location-final.html'))
+        // used to create cross-origin navigation situations
+        await networkSandbox.serveFileFromProtocol('foo', path.join(fixtures, 'api', 'window-open-location-change.html'))
+        await networkSandbox.serveFileFromProtocol('bar', path.join(fixtures, 'api', 'window-open-location-final.html'))
 
         w.destroy()
         w = new BrowserWindow({
@@ -2004,8 +2008,7 @@ describe('BrowserWindow module', () => {
       })
 
       afterEach(async () => {
-        await unregisterProtocol('foo')
-        await unregisterProtocol('bar')
+        await networkSandbox.reset()
       })
 
       it('opens window of about:blank with cross-scripting enabled', (done) => {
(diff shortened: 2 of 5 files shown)


codebytere:
```
not ok 842 webContents module zoom api cannot persist zoom level after navigation with webFrame
  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
  
  0.6 !== 0
```
failed on all platforms, so i believe it's a regression from this pr

alexstrat:
Oddly, it looks like it makes the [test-case about temporary zoom](https://github.com/alexstrat/electron/blob/d87b3ead760ae2d20f2401a8dac4ce548f8cd5f5/spec/api-web-contents-spec.js#L706-L729) fail :/

My feeling is that `GetRelatedSiteInstance` is less restrictive than `IsSameWebSite` for some particular cases: it prefers keeping the same `SiteInstance`.
For instance, my guess is that from `file://a` to `file://b`, `GetRelatedSiteInstance` will keep the same `SiteInstance` while we were creating a new one previously.
If some tests (like the test-case about temporary zoom) were relying on `file://a-> file://b` to simulate cross-origin situations, they now are abusively failing.

Looking into `GetRelatedSiteInstance` implementation and will try a version of the temporary zoom test that uses other cross-origin simulations.

alexstrat:
Given `SiteInstance` [unittests](https://cs.chromium.org/chromium/src/content/browser/site_instance_impl_unittest.cc?type=cs&sq=package:chromium&g=0&l=330), it seems that by using `GetRelatedSiteInstance`:
- `file://a` and `file://b` will be in the same site instance
- `http://localhost:123` and `http://localhost:456` will be in the same site instance
- (obviously, `http://host1.com` and `http://host2.com` will be in different site instances)

So the tests relying on `file://a -> `file://b` to simulate cross-site navigation should be updated 😬

codebytere:
I would say that the second option would probably be optimal for the purposes of this PR: to update tests to reflect the behavior your PR is designed to create

alexstrat:
@codebytere I'm not sure to understand your answer: you seem to recommend option 2 with the arguments of option 1 :)
- option 1: update tests that uses `file://a -> file://b` to simulate cross-origin navigation (for instance by using `http://hotst1.com -> http://hotst2.com` and `interceptStringProtocol('http',..)`
- option 2: change the patch to make sure that, with `file://` protocol, we change the`SiteInstance`, any time while, with other protocols, we trust `GetRelatedSiteInstance`

After some thought, I lean towards option 1 because:
- the original intentions of the cross-origin tests were to test cross-origin navigation situations (`http://host1.com` to `http://host2.com). The modifications of option 1 will make them closer to reality.
- option 2 will introduce a special behavior (more code) that differs from Chromium behavior (risks of inconstencies and confusion) whereas, AFAIK, there is no demand for such behavior (treating `file://a` and `file://b` as different SiteInstance)

But I'd like to get the opinion of @deepak1556 and @ppontes beforehand. 🙏

codebytere:
Whoops poor communication on my part: I intended option 1

i'll also defer to @deepak1556 and @ppontes though 🙇‍♀

deepak1556:
Yeah the file scheme behavior difference we have from chrome is for legacy reasons, `file` doesn't have a origin concept so it doesn't make sense for `file://a` and `file://b` to have different site instances. IIRC the reason we went against this policy was because users were loading native modules using this scheme and we wanted to restart the renderer process for those navigations as well.

For fixing the test, yes option 1 is the right route. But I also want to decide on the `file` scheme behavior, whether we want to maintain this difference or align with browser behavior. There are obvious advantages this. /cc @zcbenz  thoughts ?

zcbenz:
> But I also want to decide on the `file` scheme behavior, whether we want to maintain this difference or align with browser behavior. There are obvious advantages this. /cc @zcbenz thoughts ?

As long as the renderer process is still restarted for navigations, I'm good with either way, otherwise it would be a breaking change.

alexstrat:
OK, I updated the failing test to test (more) real cross-origin navigations. However:
- I now have a doubt on the original intention of this test. I originally thought it was to test cross-origin navigation and not just navigation on the `file://` protocol but I stumbled upon [this test](https://github.com/electron/electron/blob/0a7537f2eb7f183ddec16637e8a2e92a0d600321/spec/api-web-contents-spec.js#L576) that seems to also test zoom persistence for cross-origin navigation.
- In reference to @zcbenz remark, I don't know how to verify that the renderer process is still restarted for navigations 🤷‍♂️
- I'm hoping this fix to land in 5.x release. So if there is anything is this PR that is too "breaking change", I'd prefer to rework it to remove breaking changes (or make 2 PRs with and without breaking changes).

MarshallOfSound:
>  in Electron window.opener is implemented with JS in lib/renderer/window-setup.ts so it should be possible to fix it JS.

FWIW this doesn't work with `contextIsolation` and a medium-long term goal for me is to remove this.  We should be relying on `nativeWindowOpen` and find a way to fix this without JS tricks.  If we fundamentally can't (which is the conclusion I'm coming to) then we should start seriously looking into how we can follow Chromiums process model and safely destruct/cleanup node environments.

patches/common/chromium/frame_host_manager.patch
```diff
@@ -50,7 +50,7 @@ index 2314c6d1d29ebc61d5156644617e9eb088df5d43..9c75367a173b836027b3b7628b49b363
 +        overriden_site_instance =
 +            candidate_site_instance
 +                ? candidate_site_instance
-+                : SiteInstance::CreateForURL(browser_context,
++                : current_site_instance->GetRelatedSiteInstance(
```
ppontes: I'm afraid that the fix to this problem is not this simple. As @zcbenz has already mentioned, we need to force the renderer process to restart upon navigating, to properly restart Node. This is the reason Electron is forcing an independent site instance in this case. #15821 fixed the window.opener problem for sandboxed renderers only, where Node is disabled and therefore there is no need to restart the process. I agree with @MarshallOfSound in that we should find a way to move towards the Chromium process model while properly restarting Node without forcing full process restarts.

alexstrat:
@zcbenz oh ok I see it now: indeed it does break the purpose of `frame_host_manager.patch` because `SiteInstance:: GetRelatedSiteInstance` will re-use an existing `SiteInstance` if one matches while we were expecting to create one every-time.

Actually, I was originally looking an equivalent of `SiteInstance::CreateForURL` that’d keep the `BrowsingInstance` but could not find it, so I had to use `GetRelatedSiteInstance`.
I will try to implement a `SiteInstance::CreateRelatedSiteInstance` as a patch.

I +1 @MarshallOfSound opinion for the JS based implementation of `window.opener`, even more, because it will produce an incomplete implementation (what happens to the result of `window.open` upon the navigation of the opener?).

For anyone (maybe me) willing to dig the subject of safely destruct/cleanup node environments, what are the relevant issues/code bytes one can look into? I had the feeling that @MarshallOfSound made some good progress with 
#16425, didn’t you?

ppontes:
@alexstrat, thanks for looking into this!

> I will try to implement a SiteInstance::CreateRelatedSiteInstance as a patch.

With this approach we can end up with multiple site instances for the same site as part of the same browsing instance. Please ensure that this doesn't fail Chromium's expectations.

alexstrat:
I implemented the `SiteInstance::CreateRelatedSiteInstance` and it works as it was expected in https://github.com/electron/electron/pull/18173#issuecomment-490900468. (I also reverted my changes on the zoom persistence tests and verified it still passes).

Regarding @ppontes 's [remark](https://github.com/electron/electron/pull/18173#issuecomment-490903899), I'm still unsure of the consequences of having multiple site instances for the same site as part of the same browsing instance — except that it distances us from the 1 process-per-site model, from which we are already far anyway.

alexstrat:
@zcbenz @ppontes I made the necessary changes. Do you think you could have a second look soon? 🙏

ppontes:
@alexstrat, please reset the changes introduced in the last commit to unrelated patch files.
The code seems logical but while testing your branch I bumped into the fact that preload scripts are no longer executed when opening x-site popups from an non-sandboxed renderer with nativeWindowOpen === true. Can you please check this on your side?

alexstrat:
Thanks for your time!

Regarding the changes to unrelated patch files, I [have had a discussion on the Electron's Slack](https://electronhq.slack.com/archives/CB6CG54DB/p1557441164132700?thread_ts=1557440484.129900&cid=CB6CG54DB) with @nornagon about this who advised me to keep them 🤷‍♂️:
> that’s not unusual. it happens when roller-bot updates the DEPS version but doesn’t update the patches.
> you can include the changes in your PR

Happy to remove it anyway!


Regarding preload scripts non-execution in child windows with `nativeWindowOpen`, it is an intended behavior since #15213.
With my changes, because we now keep the opener reference, it's enforced even after cross-site navigation of the child window.
So, technically, it's intended behavior, and my changes fix cases that were not handled before.

However, the intention of #15213 was to make sure we don't have several node integrations in a single render process. Given the fact that cross-site navigation creates a new render process, we could implement this intention differently.
I think, though, it's outside the scope of this PR but LMK if you disagree, I'll make the necessary.

Note you can make sure preload-scripts are executed in child-windows with `nodeIntegrationInSubFrames` (I suggested to precise the documentation about it in #18156).

zcbenz:
I have tried the new change but it still breaks the policy that navigations should restart the renderer process. It can be tested by open a process monitor and the reload the default app, and the PID should change for every reload.

This would be a breaking change if the policy is changed. Using Atom and VS Code as example, currently you can reload the window and everything would still work, with this change the app would crash after reloading, because they are using native modules.

We should of course find a better way than always reloading the renderer process, but before we get a better solution we have to stick to current way to ensure apps can work correctly.

alexstrat:
@zcbenz I had understood these requirements and their implications and I was sure I made the necessary, so I was surprised and sorry that it does not work on your side. 😔

However, I tested with `v6.0.0-beta.2` that the PID changes for every reloads of the default app and, on my side, I was [**not** able to verify it](http://recordit.co/eWwaefME2r).
My guess is that the [addition of the `sandbox` flag to the default app](https://github.com/electron/electron/pull/17318/files#diff-64c62ccb41326824946e3d14793c19e9R55) changed the behavior you were used to see. 

You can verify that my changes have no impact when `sandbox` flag is off, the render process is properly restarted.

ppontes:
@MarshallOfSound, sorry if my knowledge is incomplete, but I believe Chromium's permission model won't allow anything to be done with window.opener except post messages to it.
```
window.opener.location.href
Uncaught DOMException: Blocked a frame with origin "..." from accessing a cross-origin frame.
```
The facts that you listed already occur since Electron 5 for windows popped up from sandboxed renderers with nativeWindowOpen === true. This PR aligns the non-sandboxed behaviour with the sandboxed one.

alexstrat:
@MarshallOfSound my knowledge may be incomplete, so I'm not sure to fully understand your remark. But I'm motivated to fully dig into it with a bit of help 💪:

Do you have a concrete example of a deadlock situation in Electron with `ipc` module if bidirectional synchronous communication were enabled?

In the context of this PR, do you think of a concrete example that'd lead to deadlocks caused by cross-process bi-directional synchronous communication?

AFAIK my changes activate Chromium's implementation of cross-site scripting, that, internally, makes use of cross-process communication. We can assume Chromium knows how to avoid deadlocks in this machinery, right?  What makes you think it wouldn't be the case?

Thank for your time 🙏

alexstrat:
@MarshallOfSound @deepak1556, as last reviewers, do you think you can have a (second) look at this PR/discussion so that we can move forward? Thanks 🙏

patches/common/chromium/frame_host_manager.patch
```diff
@@ -107,8 +141,36 @@ index 2314c6d1d29ebc61d5156644617e9eb088df5d43..9c75367a173b836027b3b7628b49b363
...
+diff --git a/content/browser/site_instance_impl.cc b/content/browser/site_instance_impl.cc
+index 703409b11a43f94896dc8b74eb87e3bfc7313710..ed5e84980f12cb86634b24a6d73d343a3308ee2a 100644
+--- a/content/browser/site_instance_impl.cc
++++ b/content/browser/site_instance_impl.cc
+@@ -297,6 +297,11 @@ scoped_refptr<SiteInstance> SiteInstanceImpl::GetRelatedSiteInstance(
+       url, /* allow_default_instance */ true);
+ }
+ 
++scoped_refptr<SiteInstance> SiteInstanceImpl::CreateRelatedSiteInstance(const GURL& url) {
++  return browsing_instance_->CreateSiteInstanceForURL(url);
```
deepak1556: Why can't this use `GetSiteInstanceForURL` in the existing browsing instance ? `browsing_instance_->GetSiteInstanceForURL(url, false /* allow_default_instance */);`

with the current implementation in the PR, any call to create the site instance for same url will go untracked by the site instance map maintained by browsing instance. I don't see an issue with reusing the site instance from the browsing instance, since we are maintaining a 1:1 relation.

alexstrat: > Why can't this use `GetSiteInstanceForURL` in the existing browsing instance ? `browsing_instance_->GetSiteInstanceForURL(url, false /* allow_default_instance */);`

`browsing_instance_->GetSiteInstanceForURL` [is equivalent](https://cs.chromium.org/chromium/src/content/browser/site_instance_impl.cc?type=cs&sq=package:chromium&g=0&l=336-340) to the [first suggestion](https://github.com/electron/electron/pull/18173/commits/50aae0b0bf8612dfdec8c2c75493dc09818d3e3f) of this PR.
As [discussed earlier in the thread](https://github.com/electron/electron/pull/18173#pullrequestreview-235362945), it can't work because, for the same URL (during a reload for instance), we'll end up reusing the same `SiteInstance`, and, as a result, the render process won't be restarted — in opposition to what is expected.

> with the current implementation in the PR, any call to create the site instance for the same url will go untracked by the site instance map maintained by browsing instance

I'm not sure because in `CreateSiteInstanceForURL`, I call `instance->SetSite` that calls `RegisterSiteInstance` which adds the `SiteInstance` to the ` site_instance_map_` of browsing instance.

>  I don't see an issue with reusing the site instance from the browsing instance since we are maintaining a 1:1 relation.

As I understood with @zcbenz , reusing the site instance will not trigger render process restart which is an issue. Maybe I have not understood this last sentence @deepak1556 though

deepak1556: Yes we need to restart process. Regarding the site instance map statement, when `GetSiteInstanceForURL` it will either create a new instance for the url, add it to the map and return it, otherwise return an existing one from the map. With this PR implementation, a site instance is created and add to the map but when a new request is made for the same url, it will create the instance but will not add it to the map, check the condition in `BrowsingInstance::RegisterSiteInstance`. I was concerned about this.

But since we are using the same browser instance makes sense that `GetSiteInstanceForURL` can't be used to force restart the process.

deepak1556: > Even if we did not register it, I don't see any issue with that so far, since we want to use new SiteInstance every time.

Browsing Instance has a reference to the site instance when its created, if its not registered we will be managing the lifetime.

alexstrat:
@MarshallOfSound Thanks for your thoughts and suggestions! 

To set up the context again, the goal of this PR is to fix #18032 which is a regression in electron 5: indeed, in <5, `window.opener` was present after cross-site navigation in popups, even if it was only the proxy version. This regression, for instance, prevents OAuth popups from working properly.

So I'd be fine with considering #18396 as a fix to #18032, but, as of now, there are several drawbacks that prevent it 😕:
- it won't land in electron 5
- it breaks/prevents the imports of (some?) native module in renderers (breaking change)
- it forces its user to walk into a (so far) uncharted territory
- it still leaves the implementations of `window.opener` with `nativeWindowOpen` misleadingly incomplete

I understand this patch getting even more complex is worrying, but obviously (#15821, #9961, #10204) the problem is a tough one*, which, in your eyes, could justify the iterations, challenging the solutions and, in the end, the complexity of its solution.
Note also, that the discussions in this PR would help anyone willing to understand and tackle the complexity :)

Also, shouldn't the (still experimental) `sandbox` option be the norm in few versions? In this case, we could remove the site instance patch when `sandbox=true` will be the only option.

\* Thanks 🙏 for the sweat of everyone who gave it a shot!

MarshallOfSound:
@alexstrat Maybe I'm missing something here

> To set up the context again, the goal of this PR is to fix #18032 which is a regression in electron 5

This is fair. I was under the impression that this didn't really work before (the result was unpredictable and broke native modules in the child window).

> it still leaves the implementations of window.opener with nativeWindowOpen misleadingly incomplete

I'm not sure I understand this, `nativeWindowOpen` relies entirely on Chromiums implementation.  `window.opener` works exactly as expected.

> I understand this patch getting even more complex is worrying, but obviously (#15821, #9961, #10204) the problem is a tough one*, which, in your eyes, could justify the iterations, challenging the solutions and, in the end, the complexity of its solution.

Something to make clear here, my goal isn't to figure out how to make this patch work perfectly, because IMO there is no way we can figure that out and keep it up to date with Chromium's changes.  Even now the patch has issues where preload scripts run twice in the same process.  My goal here is to **remove** this patch.

> I understand we would like to have a plan to remove the patch. Isn't it the plan that the (still experimental) sandbox option be the norm in few versions?

`sandbox=true` will never (IMO) be the only option, we'll always have to support running a real node environment in the renderer.  The path to removing this patch is outlined in #18397 and is something we'll be working towards for a while.

I see the _potential_ that this bug is something we should fix in 5, my concern is as Chromium's process model has gotten more complex we've had to make this patch more and more complex.  At some point we need to say, no more, and point people at the future alternative.  I'm not going to block this patch, but it seems counter-productive to have an ongoing initiative to remove a patch and simultaneously make that patch more complex at the same time.

alexstrat:
>>it still leaves the implementations of window.opener with nativeWindowOpen misleadingly incomplete

>I'm not sure I understand this, `nativeWindowOpen` relies entirely on Chromiums implementation. `window.opener` works exactly as expected.

🤔I'm not sure it does — or depends on the scope of expectations 😀. To be precise, with using `nativeWindowOpen`, `window.opener` ends up being `null` after cross-site navigation: it's the [reproduction case](https://github.com/alexstrat/electron-bug-repro-cross-origin-opener) I based on for this PR.

---

>I see the potential that this bug is something we should fix in 5, my concern is as Chromium's process model has gotten more complex we've had to make this patch more and more complex. At some point we need to say, no more, and point people at the future alternative. I'm not going to block this patch, but it seems counter-productive to have an ongoing initiative to remove a patch and simultaneously make that patch more complex at the same time.

👍Of course: I do understand, I do imagine the puzzle of juggling between the past and the future!*

And as a user, I'm willing to be pointed to try future alternatives. (we are on our way to test `sandbox` that, in our case, is an alternative).

Though, when a regression is blocking and that the alternatives are still future or experimental, it's difficult to accommodate with "no more". :/

Hence this PR that is a fix to the regression #18032 — but I'm open to exploring alternatives that work around it in Electron 5.

\* and, once again, kudos to the team 🙏

deepak1556:
I am +1 on landing this PR, since its a required fix for 5.x. The complexity of this patch is trivial IMO.

Also agree with @MarshallOfSound that users requiring chromium process model with node Integration should start moving to the model described in https://github.com/electron/electron/issues/18397, but that should be a path forward. Not something to be considered as a fix for apps in 5.x and lower.

MarshallOfSound:
> To be precise, with using nativeWindowOpen, window.opener ends up being null after cross-site navigation: it's the reproduction case I based on for this PR.

I don't think we're understanding each other, I'm saying that with the changes in #18397 `nativeWindowOpen` will operate identically to Chromium.  In its current state, it does not.  That is actually one of the benefits outlined in that discussion issue.

> Not something to be considered as a fix for apps in 5.x and lower.

This is a fair point, I'll unblock this PR 👍

alexstrat:
> I don't think we're understanding each other, I'm saying that with the changes in #18397 nativeWindowOpen will operate identically to Chromium. In its current state, it does not. That is actually one of the benefits outlined in that discussion issue.

Ok yes indeed. My remark was original about #18396 , not #18397.

MarshallOfSound:
> Ok yes indeed. My original remark was about #18396 , not #18397.

They're the same thing 😄

alexstrat:
Sorry, it looks like I (force-)pushed a wrong rebase. Will fix that ASAP.

alexstrat:
Rebase ✅
Ready to merge, as far as I can tell 👐

alexstrat:
@MarshallOfSound @deepak1556 can one of you merge this PR, please?

alexstrat:
Backports for 5-0-x and 6-0-x are ready to review and merge.

alexstrat:
@MarshallOfSound @deepak1556 could you approve the backport for electron 5 #18614 so that the fix gets released soon, please?

Issue #17622, opened 2019-03-30
API  tray.destroy is not working 


rhythm-sharma:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  *  4.0.4
* **Operating System:** 
  *  Ubuntu 18.10 x64

### Expected Behavior

 Observed in Zulip desktop app
 1. ```tray.destroy()``` api should destroy tray icon in panel
 2.  ```show app tray icon``` option in general setting, it should destroy and create tray icon on toggle

### Actual Behavior

Previous tray icon are not destroy on toggle and new icon created i.e ```tray.destroy``` API is not working 

![Screenshot from 2019-03-30 18-57-58](https://user-images.githubusercontent.com/36422396/55276803-eacd7080-531d-11e9-842b-8d15bcfc09dd.png)

### To Reproduce

```
$ git clone https://github.com/zulip/zulip-electron
$ cd zulip-electron
$ npm install
$ npm start
```
you can find code in ``` zulip-electron/app/renderer/js/tray.js ```

```
function toggleTray() {
	let state;
	if (window.tray) {
		state = false;
		window.tray.destroy();
		if (window.tray.isDestroyed()) {
			window.tray = null;
		}
		ConfigUtil.setConfigItem('trayIcon', false);
	} else {
		state = true;
		createTray();
		if (process.platform === 'linux' || process.platform === 'win32') {
			renderNativeImage(unread).then(image => {
				window.tray.setImage(image);
				window.tray.setToolTip(unread + ' unread messages');
			});
		}
		ConfigUtil.setConfigItem('trayIcon', true);
	}
	const selector = 'webview:not([class*=disabled])';
	const webview = document.querySelector(selector);
	const webContents = webview.getWebContents();
	webContents.send('toggletray', state);
}
```

### Screenshots

![Image 2019-03-17 at 12 45 47 PM](https://user-images.githubusercontent.com/36422396/55276412-7395dd80-5319-11e9-9032-8da0050cd16c.jpeg)

rrhythmsharma:
@MarshallOfSound  can you please review this ?

[...]

PR #18196, opened 2019-05-07
fix: destroy tray on current tick

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/17622.

This code was originally added in https://github.com/electron/electron/pull/6448 to handle an edge case crash in 10.9, and we no longer support 10.9 and therefore no longer need to account for this case.

It addressed the crash, but also created a race condition whereby when a new tray is created the old tray's destroy wouldn't have been fully completed and therefore a new one would be spawned. This fixes that by destroying the tray on the current tick once more. 

cc @zcbenz @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where tray.destroy was not working properly on some linux distros.

### DIFF
--- a/atom/browser/api/atom_api_tray.cc
+++ b/atom/browser/api/atom_api_tray.cc
@@ -60,7 +60,7 @@ Tray::Tray(v8::Isolate* isolate,
   InitWith(isolate, wrapper);
 }
 
-Tray::~Tray() {}
+Tray::~Tray() = default;
 
 // static
 mate::WrappableBase* Tray::New(mate::Handle<NativeImage> image,
--- a/atom/browser/api/atom_api_tray.cc
+++ b/atom/browser/api/atom_api_tray.cc
@@ -60,11 +60,7 @@ Tray::Tray(v8::Isolate* isolate,
   InitWith(isolate, wrapper);
 }
 
-Tray::~Tray() {
-  // Destroy the native tray in next tick.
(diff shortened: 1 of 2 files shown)


codebytere:
@deepak1556 i added some specs but if there are more specifics that i should test please let me know!

MarshallOfSound:
Can we add a test for the original issue?  (Calling `app.quit()` inside menu even handler).  Not sure if this is possible but would be good if it is 👍

atom/browser/api/atom_api_tray.cc
```diff
@@ -60,11 +60,7 @@ Tray::Tray(v8::Isolate* isolate,
   InitWith(isolate, wrapper);
 }
 
-Tray::~Tray() {
-  // Destroy the native tray in next tick.
-  base::ThreadTaskRunnerHandle::Get()->DeleteSoon(FROM_HERE,
-                                                  tray_icon_.release());
-}
+Tray::~Tray() {}
```
miniak: you can use fancy C++11 `Tray::~Tray() = default;` :)

codebytere:
@MarshallOfSound i don't think that's possible sadly :( 
The original issue didn't call `app.quit()` i don't believe?

codebytere:
Mac failure is a flake: merging!

PR #18202, opened 2019-05-07
chore: run protocol tests in separate WebContents

zcbenz:
#### Description of Change

Refs #15791.

After migrating protocol module to NetworkService, there is a behavior change that, after registering a new protocol, it is impossible to access the new-registered protocol via ajax requests immediately in the opened WebContents, all WebContents have to do navigations to be able to access the new-registered protocol.

However our tests for protocol module rely on this behavior, so in order to make the tests pass with NetworkService enabled, I have to refactor the tests to do ajax requests in a separate WebContents instead of current WebContents.

I have also simplified the tests with async/await, and if in future we found a way to work around this limitation of NetworkService, it is very easy to revert back to the old way by changing a few lines.

I'll document the behavior changes of protocol module in other PRs.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/spec/api-protocol-spec.js
+++ b/spec/api-protocol-spec.js
@@ -2,6 +2,8 @@ const assert = require('assert')
 const http = require('http')
 const path = require('path')
 const qs = require('querystring')
+const { promisify } = require('util')
+const { emittedOnce } = require('./events-helpers')
 const { closeWindow } = require('./window-helpers')
 const { remote } = require('electron')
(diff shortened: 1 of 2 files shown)


deepak1556:
If the registration of `URLLoader` factories was handled by `NetworkContext` we could rever this, but its handled by `NavigationURLLoader`, so requires a navigation for the registration to kick in.

PR #18221, opened 2019-05-09
refactor: nws13n: setCertificateVerifyProc

nornagon:
#### Description of Change
Converts setCertificateVerifyProc to use the network service. This requires a patch to Chromium, as the network context mojo interface does not expose this functionality currently.

This is a WIP because:
1. The current patch has not been submitted to Chromium for feedback, and I don't think it would be wise to float a patch of this size on top of volatile new code such as the network service, and
2. The current patch incurs a process-hop overhead for every certificate verification regardless of whether a certificate verification procedure has been set for the session.
3. The tests don't pass.
4. I'm not yet sure how to do this in a way that will work in both network service and non-network service conditions.

Also, the API in network_context.mojom is not final; I don't yet fully understand the CertVerifyResult structure and whether it needs to be communicated from the remote verifier back to the network service.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/patches/chromium/network_service_allow_remote_certificate_verification_logic.patch
+++ b/patches/chromium/network_service_allow_remote_certificate_verification_logic.patch
@@ -5,10 +5,18 @@ Subject: network service: allow remote certificate verification logic
 
 
 diff --git a/services/network/network_context.cc b/services/network/network_context.cc
-index 5d871fc08218b71c93141963db66465f775b1bc2..9e04684bede95aaa3a32c4b3aa280ff99d914a4c 100644
+index 5d871fc08218b71c93141963db66465f775b1bc2..0784ff23c10563f25af1a520705f3264b8cfe9dd 100644
 --- a/services/network/network_context.cc
 +++ b/services/network/network_context.cc
-@@ -310,6 +310,75 @@ std::string HashesToBase64String(const net::HashValueVector& hashes) {
+@@ -43,6 +43,7 @@
+ #include "net/base/registry_controlled_domains/registry_controlled_domain.h"
+ #include "net/cert/cert_verifier.h"
+ #include "net/cert/ct_verify_result.h"
++#include "net/cert/multi_threaded_cert_verifier.h"
+ #include "net/cert_net/cert_net_fetcher_impl.h"
+ #include "net/cookies/cookie_monster.h"
+ #include "net/dns/host_cache.h"
+@@ -310,6 +311,75 @@ std::string HashesToBase64String(const net::HashValueVector& hashes) {
  
  }  // namespace
  
@@ -84,7 +92,7 @@ index 5d871fc08218b71c93141963db66465f775b1bc2..9e04684bede95aaa3a32c4b3aa280ff9
  constexpr uint32_t NetworkContext::kMaxOutstandingRequestsPerProcess;
  constexpr bool NetworkContext::enable_resource_scheduler_;
  
-@@ -668,6 +737,12 @@ void NetworkContext::SetClient(mojom::NetworkContextClientPtr client) {
+@@ -668,6 +738,12 @@ void NetworkContext::SetClient(mojom::NetworkContextClientPtr client) {
    client_ = std::move(client);
  }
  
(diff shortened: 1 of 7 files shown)


nornagon:
Undrafting this because (2), (3) and (4) from the initial PR description have been addressed. I'm still not sure how to approach upstreaming this yet.

atom/browser/atom_browser_context.h
```diff
@@ -11,6 +11,7 @@
 #include <vector>
 
 #include "atom/browser/media/media_device_id_salt.h"
+#include "atom/browser/net/atom_cert_verifier.h"
```
nornagon:
```suggestion
```

patches/chromium/network_service_allow_remote_certificate_verification_logic.patch
```diff
@@ -0,0 +1,176 @@
...
++             const net::NetLogWithSource& net_log) override {
++    out_req->reset();
++
++    net::CompletionOnceCallback callback2 = base::BindOnce(
++        &RemoteCertVerifier::OnRequestFinished, base::Unretained(this),
++        params, std::move(callback), verify_result);
++    int result = upstream_->Verify(params, verify_result,
++                                   std::move(callback2), out_req, net_log);
++    if (result != net::ERR_IO_PENDING) {
++      // Synchronous completion
```
deepak1556: what happens with synchronous results ?

atom/browser/api/atom_api_session.cc
```diff
@@ -425,19 +461,32 @@ void WrapVerifyProc(
...
-                     WrapRefCounted(browser_context_->GetRequestContext()),
-                     base::BindRepeating(&WrapVerifyProc, proc)));
+  if (base::FeatureList::IsEnabled(network::features::kNetworkService)) {
+    network::mojom::CertVerifierClientPtr cert_verifier_client;
+    if (proc) {
+      mojo::MakeStrongBinding(
+          std::make_unique<ElectronCertVerifierClient>(proc),
+          mojo::MakeRequest(&cert_verifier_client));
+    }
+    content::BrowserContext::GetDefaultStoragePartition(browser_context_.get())
```
deepak1556: This could also be part of `if (proc)`

patches/chromium/network_service_allow_remote_certificate_verification_logic.patch
```diff
@@ -0,0 +1,176 @@
...
+ void NetworkContext::CreateURLLoaderFactory(
+     mojom::URLLoaderFactoryRequest request,
+     mojom::URLLoaderFactoryParamsPtr params) {
+@@ -2134,8 +2209,13 @@ URLRequestContextOwner NetworkContext::MakeURLRequestContext() {
+       cert_verifier = net::CertVerifier::CreateDefault(cert_net_fetcher_);
+   }
+ 
+-  builder.SetCertVerifier(IgnoreErrorsCertVerifier::MaybeWrapCertVerifier(
+-      *command_line, nullptr, std::move(cert_verifier)));
++  cert_verifier = IgnoreErrorsCertVerifier::MaybeWrapCertVerifier(
```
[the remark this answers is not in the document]
nornagon: I think it's better for rebaseability to insert lines rather than edit lines.

patches/chromium/network_service_allow_remote_certificate_verification_logic.patch
```diff
@@ -0,0 +1,176 @@
...
++
+ constexpr uint32_t NetworkContext::kMaxOutstandingRequestsPerProcess;
+ constexpr bool NetworkContext::enable_resource_scheduler_;
+ 
+@@ -668,6 +737,12 @@ void NetworkContext::SetClient(mojom::NetworkContextClientPtr client) {
+   client_ = std::move(client);
+ }
+ 
++void NetworkContext::SetCertVerifierClient(mojom::CertVerifierClientPtr client) {
++  if (remote_cert_verifier_) {
```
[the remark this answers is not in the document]
nornagon: hm... i'm not sure how this could get called if `url_request_context_` is null?

patches/chromium/network_service_allow_remote_certificate_verification_logic.patch
nornagon: synchronous results means the result was served from the cache. `upstream_` is a `CachingCertVerifier`, which will serve up a result from its cache synchronously and delete its callback if there's a cache entry present. It means your handler will only get called every so often rather than on every request, which I think is reasonable.

deepak1556: Does the cache use result provided previously by the user or is it based on the results of the upstream verifier ?

patches/chromium/network_service_allow_remote_certificate_verification_logic.patch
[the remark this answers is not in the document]
deepak1556: Yeah it will not be null, I was just writing it with the old network code path in mind.

patches/chromium/network_service_allow_remote_certificate_verification_logic.patch
nornagon: That's a great question. I think it'll use the result provided, but I'll write a test for it.

nornagon: bleh, it only caches the results of the upstream verifier. i'll have to ponder for a bit to figure out what to do here. it's a bit tricky because this code `std::move()`s the reference to the callback into the upstream `Verify` call, which means it's no longer accessible when in the sync branch. I could do some nonsense with `AdaptCallbackForRepeating` but it would be a bit of a hack...

deepak1556: Since `RemoteCertVerifier` is the one finally calling into the response callback in `OnRemoteResponse`, we could store it as a class member rather than binding it to different functions. And in the synchronous result, we can directly call `OnRequestFinished`

nornagon: I don't think so; there's no guarantee that only one request is outstanding at a time.

deepak1556: Ah forgot that the `CertVerifier` has a request queue, you are right in that case this won't work.

nornagon: hmm, maybe i can try putting the `RemoteCertVerifier` higher in the verifier chain than the caching verifier.

Issue #18236, opened 2019-05-09
showMessageBox BrowserWindow is not optional


adill:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 6.0.0-beta.2
* **Operating System:**
  * Windows 10 (1809)
* **Last Known Working Electron version:**:
  * 5.0.0

### Expected Behavior
`dialog.showMessageBox({options...})` should display a message box matching the options passed in

### Actual Behavior
A message box complaining about parsing argument at index 11 displays

### To Reproduce
```sh
$ git clone https://github.com/adill/electron-quick-start -b show-message-box
$ npm install
$ npm start || electron .
```

codebytere:
Fairly certain i know what's behind this, i'll try to get a fix out imminently!

PR #18240, opened 2019-05-09
fix: set window to null when no window is passed

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/18236.

Our options parsing for dialogs correctly set `options` if no `window` was passed, but then did not correctly set the `window` back to null and it was then left as an object equal to options. This caused a crash when passed to native code as there is no valid circumstance where the `window` would be an object.

This same issue would have affected `showSaveDialog` and `showOpenDialog` and as such it is also fixed in those methods.

cc @adill @sofianguy @MarshallOfSound @miniak 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a dialog crash when no BrowserWindow was passed.

### DIFF
--- a/lib/browser/api/dialog.js
+++ b/lib/browser/api/dialog.js
@@ -47,7 +47,11 @@ const checkAppInitialized = function () {
 const saveDialog = (sync, window, options) => {
   checkAppInitialized()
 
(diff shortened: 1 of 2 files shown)


codebytere:
no :( not until we can effectively implement gui testing

Issue #18242, opened 2019-05-09
app.setPath() does not create directories


inukshuk:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.1
* **Operating System:**
  * macOS 10.14

### Expected Behavior
According to the [API documentation](https://github.com/electron/electron/blob/master/docs/api/app.md#appsetpathname-path) `app.setPath()` should create the directory passed, if it does not exist.

### Actual Behavior
It does not.

### To Reproduce
In particular, I was calling `app.setPath('logs', path)` and `app.setPath('userCache', path)` with non-existent directories; they were not created.

### Additional Information
The creation [seems to be delegated to `base::PathService::OverrideAndCreateIfNeeded`](https://github.com/electron/electron/blob/master/atom/browser/api/atom_api_app.cc#L879), so I think this may just be a simple bug (the last parameter should be `true` not `false`).

Issue #2668, opened 2015-08-31
default userData directory is created in the default location when using app.setPath()


etiktin:
I'm using `app.setPath` to change the location where `userData` and `cache` are stored and it works (data is saved to the new location). The problem is that even due I set the paths before the `ready` event, the default location (`%APPDATA%/productName`) is still created.

To test this, I deleted the default path and added to the top of my main.js the following lines:
```javascript
var fs = require('fs');
console.log('Exists?', fs.existsSync('C:/Users/eran.t/AppData/Roaming/foo'));
```
The output was `Exists? true`, which means the directory was created by some initialization code (it looks like it comes from [here](https://github.com/atom/electron/blob/c441dd143690aab71a925f0b941a6d9760768fa5/atom/browser/lib/init.coffee#L88-L89)).

Is it possible to move the creation of the directories to when we actually need them (lazy creation), which will give the developer time to change them?

To circumvent the issue, one can simply delete the old path using `fs.rmdir` after changing the path to the new location, but it's not ideal.

leshow:
Also seeing this bug on windows/osx

PR #18244, opened 2019-05-09
fix: explicitly don't create missing directories with app.setPath

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/18242.

Our [documentation](https://github.com/electron/electron/blob/master/docs/api/app.md#appsetpathname-path) specifies that nonexistent directories should be created when `app.setPath` is called if they don't already exist. We do this with [`base::PathService::OverrideAndCreateIfNeeded`](https://cs.chromium.org/chromium/src/base/path_service.h?sq=package:chromium&pv=1&g=0&l=58), which uses the last parameter to specify whether the path should be created.

This has potential for side effects should it be enabled, and so this PR corrects documentation to reflect implementation reality.

cc @nornagon @inukshuk @deepak1556 @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -580,7 +580,7 @@ them.
 
 * `path` String (optional) - A custom path for your logs. Must be absolute.
 
-Sets or creates a directory your app's logs which can then be manipulated with `app.getPath()` or `app.setPath(newPath)`.
+Sets or creates a directory your app's logs which can then be manipulated with `app.getPath()` or `app.setPath(pathName, newPath)`.
 
 On _macOS_, this directory will be set by deafault to `/Library/Logs/YourAppName`, and on _Linux_ and _Windows_ it will be placed inside your `userData` directory.
 
@@ -641,9 +641,9 @@ On _Linux_ and _macOS_, icons depend on the application associated with file mim
 * `name` String
 * `path` String
 
-Overrides the `path` to a special directory or file associated with `name`. If
-the path specifies a directory that does not exist, the directory will be
-created by this method. On failure an `Error` is thrown.
+Overrides the `path` to a special directory or file associated with `name`.
+If the path specifies a directory that does not exist, an `Error` is thrown.
+In that case, the directory should be created with `fs.mkdirSync` or similar.
 
 You can only override paths of a `name` defined in `app.getPath`.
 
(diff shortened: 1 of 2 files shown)


deepak1556:
Oh wow, has this always been like this ?

* Creating directory is a blocking task, I would have liked it to be posted as a task to different sequence but given the user facing api needs synchronous results, there is nothing we can do here except marking it as a blocking call site.

* I don't want this api to be abused, if users are changing paths would it be better if we make sure the previous path is deleted and new path is created, that way we don't leave a bunch of stale directories in the user system.

deepak1556:
@zcbenz wouldn't implementing the following , fix the above issue too ?

```
if users are changing paths would it be better if we make sure the previous path is deleted and new path is created, that way we don't leave a bunch of stale directories in the user system.
```

zcbenz:
@deepak1556 Having the API automatically deleting files is a bit frightening, especially when users are passing user data paths to it, a bug in app would then end up erasing users' files.

For example:

```js
let appName = getAppName() // which somehow returns empty string
app.setPath(USER_DATA_DIR + '/' + appName)
doSomething()

if (someFlag) {
  app.setPath(SOME_WHERE_ELSE)
  doSomethingElse()
}

app.setPath(USER_DATA_DIR + '/' + appName)
// USER_DATA_DIR is gone forever
```

A safer solution is probably to have Electron set the default data path without side effect, while make the API create the directory when it is called by users. (This can be done by adding a parameter.)

But current behavior has been in Electron for 3 years, I'm not sure whether changing the behavior would cause troubles for existing apps, it would be practically a breaking change.

codebytere:
In that case, do we think for now our best bet might be to change documentation and then later circle back to changing the behavior itself?

zcbenz:
There are 2 things that we want to achieve:
1. Electron should not create directory on default data path (#2668);
2. Directory should be created when user calls `setPath` explicitly.

I think we can achieve both by adding a new parameter to `setPath` which specifies whether to create the directory.

The new parameter can:

1. default to `true` if we want to strictly follow the documentation;
2. default to `false` if we want to keep current behavior to avoid possible breaks.

I'm good with either as long as #2668 is not broken.

And yeah we should fix the documentation first as it is currently misleading users.

deepak1556:
+1 on @zcbenz suggestion for new api.

codebytere:
I decided to leave out directory creation and just leave that to users @deepak1556 @zcbenz.

PR #18255, opened 2019-05-11
feat: enable reporting api

### LINKED ISSUE #18252: Network Error Logging is disabled by default
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 4, 5, 6 
* **Operating System:**
  * not related

  * doesn't work at all

### Expected Behavior

Electron's network request honor NEL header and make NEL request.

### Actual Behavior

Does not happens

### To Reproduce

Navigate known NEL-enabled endpoint (i.e https://dcreager.net/nel/intro/), collect netlog and check `Reporting` section.

### Additional Information

NEL (https://www.w3.org/TR/network-error-logging/) is enabled by default from M69 (https://www.chromestatus.com/features/5391249376804864), but in our desktop client based on v4 (and v5, 6 for verification purpose) doesn't honor NEL header so we are not able to collect users using desktop client.

In chromium visiting NEL configured endpoint will make proper NEL request to specified endpoint, and client side behavior can be verified via netlog like below:

image

In Electron, some version simply doesn't have `Reporting` section at all, or if exists it says it's disabled:

image

nornagon:
#### Description of Change
Fixes #18252.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Enabled the W3C [Reporting API](https://w3c.github.io/reporting/).

### DIFF
--- a/atom/browser/net/atom_network_delegate.cc
+++ b/atom/browser/net/atom_network_delegate.cc
@@ -435,7 +435,7 @@ bool AtomNetworkDelegate::OnCancelURLRequestWithPolicyViolatingReferrerHeader(
 // https://crbug.com/704259
 bool AtomNetworkDelegate::OnCanQueueReportingReport(
     const url::Origin& origin) const {
-  return false;
+  return true;
 }
 
 void AtomNetworkDelegate::OnCanSendReportingReports(
--- a/atom/browser/net/atom_network_delegate.cc
+++ b/atom/browser/net/atom_network_delegate.cc
@@ -431,8 +431,6 @@ bool AtomNetworkDelegate::OnCancelURLRequestWithPolicyViolatingReferrerHeader(
   return false;
 }
 
-// TODO(deepak1556) : Enable after hooking into the reporting service
-// https://crbug.com/704259
 bool AtomNetworkDelegate::OnCanQueueReportingReport(
     const url::Origin& origin) const {
   return true;
@@ -440,16 +438,18 @@ bool AtomNetworkDelegate::OnCanQueueReportingReport(
 
 void AtomNetworkDelegate::OnCanSendReportingReports(
     std::set<url::Origin> origins,
-    base::OnceCallback<void(std::set<url::Origin>)> result_callback) const {}
+    base::OnceCallback<void(std::set<url::Origin>)> result_callback) const {
+  std::move(result_callback).Run(std::move(origins));
+}
 
 bool AtomNetworkDelegate::OnCanSetReportingClient(const url::Origin& origin,
                                                   const GURL& endpoint) const {
-  return false;
+  return true;
 }
 
 bool AtomNetworkDelegate::OnCanUseReportingClient(const url::Origin& origin,
                                                   const GURL& endpoint) const {
-  return false;
+  return true;
 }
 
 void AtomNetworkDelegate::OnErrorOccurred(net::URLRequest* request,
--- a/atom/browser/net/url_request_context_getter.cc
+++ b/atom/browser/net/url_request_context_getter.cc
@@ -47,6 +47,7 @@
 #include "net/url_request/url_request_job_factory_impl.h"
 #include "services/network/ignore_errors_cert_verifier.h"
 #include "services/network/network_service.h"
+#include "services/network/public/cpp/features.h"
 #include "services/network/public/cpp/network_switches.h"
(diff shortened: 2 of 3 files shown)


atom/browser/net/atom_network_delegate.cc
```diff
@@ -435,7 +435,7 @@ bool AtomNetworkDelegate::OnCancelURLRequestWithPolicyViolatingReferrerHeader(
 // https://crbug.com/704259
 bool AtomNetworkDelegate::OnCanQueueReportingReport(
     const url::Origin& origin) const {
-  return false;
+  return true;
```
MarshallOfSound: Is this something we might want to allow users to override.  E.g. `return Emit("will-queue-report")`

Idk what the surface of this is, is it just a web feature we should have enabled a while back or is it more 🤔

nornagon: > is it just a web feature we should have enabled a while back

yup. https://w3c.github.io/reporting

atom/browser/net/atom_network_delegate.cc
```diff
@@ -435,7 +435,7 @@ bool AtomNetworkDelegate::OnCancelURLRequestWithPolicyViolatingReferrerHeader(
 // https://crbug.com/704259
```
deepak1556: You can remove this `TODO`

atom/browser/net/atom_network_delegate.cc
nornagon: I'm not sure it makes a ton of sense to expose this as something users might want to override. Maybe? But this is a synchronous API on the IO thread and it's unclear how to trigger an event for that effectively and get a response.

I think it makes sense to defer adding a feature for this until such a time as it is needed. Otherwise we'll likely build the wrong API.

nornagon: On further consideration, it probably does make sense to expose some of this stuff.

atom/browser/net/atom_network_delegate.cc
```diff
@@ -431,25 +431,25 @@ bool AtomNetworkDelegate::OnCancelURLRequestWithPolicyViolatingReferrerHeader(
...
     const url::Origin& origin) const {
-  return false;
+  return true;
 }
 
 void AtomNetworkDelegate::OnCanSendReportingReports(
     std::set<url::Origin> origins,
-    base::OnceCallback<void(std::set<url::Origin>)> result_callback) const {}
+    base::OnceCallback<void(std::set<url::Origin>)> result_callback) const {
+  std::move(result_callback).Run(std::move(origins));
```
deepak1556: Will we be exposing this to users to implement with some permission model ?

nornagon: hm... yes, i'd like to, but i'm not sure where to expose the method. WebRequest doesn't seem like the right place. Session doesn't seem to have access to the network delegate currently. Thoughts?

deepak1556: I think its better to wait for the webrequest and network delegate refactor with the network service to expose this properly. In the meantime , can you leave a `TODO`

codebytere:
macos is a flake, passed on non-mas.

codebytere:
@nornagon would you like to make a few more changes to this or is it ok to merge as-is?

nornagon:
I haven't fully thought through the implications of enabling the reporting API by default with no option to disable it in apps. At first blush I don't think there's an issue (if you're loading a website with headers you don't control, then you're probably going to be running JS you don't control also...) but I haven't thought through all the potential ways it could be dangerous.

Adding @electron/wg-security for their thoughts.

deepak1556:
> I haven't fully thought through the implications of enabling the reporting API by default with no option to disable it in apps

Isn't it still behind a feature flag ? Can't apps disable it via `--disable-features`

>  but I haven't thought through all the potential ways it could be dangerous.

Not sure either, but the reporting types clearly seems harmless https://w3c.github.io/reporting/#report-types . If apps are loading remote content not they control, then they should be running in a sandboxed environment, also they still have the ability to control headers via the `webRequest` api

nornagon:
Ah yeah, I hadn't considered the `disable-features` route. That should work.

In that case, I think I'll go ahead and merge this. We can add configuration options and permissions requests / filtering handlers in a followup. Until then, apps can disable the feature if they want.

Issue #18210, opened 2019-05-08
Document webContents page-title-updated event


pushkin-:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

(Actually a duplicate of [this closed issue](https://github.com/electron/electron/issues/10040))

### Issue Details

* **Electron Version:** 4.0.4
* **Operating System:** Windows 10
* **Last Known Working Electron version:**: n/a

`WebContents` apparently has a `page-title-updated` event that isn't documented. We were latching the handler onto the [`webview`](https://electronjs.org/docs/api/webview-tag#event-page-title-updated), but we're in the process of transitioning from a `webview` to a `BrowserView` which doesn't have that event.

I haven't tested this, but I was told that we can latch onto the event on the webcontents ([source](https://github.com/electron/electron/blob/0755857a0c1c76453391659c6d5ac6218dc9286f/atom/browser/api/atom_api_web_contents.cc#L1049)), but it's not documented.

Can it be?

miniak:
Fixed in master

miniak:
Pending backports to release branches

PR #18270, opened 2019-05-12
fix: 'page-title-updated' event forwarding + documentation

miniak:
#### Description of Change
Address https://github.com/electron/electron/issues/18210.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added missing `'page-title-updated'` event on `webContents` to documentation. Also fixed forwarding of the `explicitSet` argument when emitted on `BrowserWindow`.

### DIFF
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -101,6 +101,17 @@ Returns:
 
 Emitted when the document in the given frame is loaded.
 
+#### Event: 'page-title-updated'
+
+Returns:
+
+* `event` Event
+* `title` String
+* `explicitSet` Boolean
+
+Fired when page title is set during navigation. `explicitSet` is false when
+title is synthesized from file url.
+
 #### Event: 'page-favicon-updated'
 
 Returns:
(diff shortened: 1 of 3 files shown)


miniak:
cc @pushkin-

docs/api/web-contents.md
```diff
@@ -101,6 +101,16 @@ Returns:
...
 Emitted when the document in the given frame is loaded.
 
+#### Event: 'page-title-updated'
+
+Returns:
+
+* `event` Event
+* `title` String
+
+Fired when page title is set during navigation. `explicitSet` is false when
```
MarshallOfSound: `explicitSet` is not documented here?

miniak: fixed

pushkin-:
@miniak  Just to clarify, `explicitSet` is false when navigating to a page that doesn't explicitly set the title via the `<title>` tag or through Javascript, and it would only ever be false the first time you navigate to a title-less page. All future `page-title-updated` calls on that page would have `explicitSet` set to `true` - is that correct?

Issue #17132, opened 2019-02-26
Clicking link with target=_blank does not fire 'new-window' event with Sandbox Enabled


chrismohr:
* **Electron Version** (output of `node_modules/.bin/electron --version`):
  * 3.0.6
* **Operating System** (Platform and Version):
  * macOS 10.14

### Expected Behavior
a `new-window` event fires

### Actual behavior
a `did-fail-load` fires with errorCode -3

### To Reproduce
```sh
$ git clone git@github.com:chrismohr/no-new-window.git -b master
$ cd no-new-window/
$ npm i && npm start
```

chrismohr:
Repro code:
https://github.com/chrismohr/no-new-window/blob/master/main.js

PR #18273, opened 2019-05-12
fix: set nativeWindowOpen when sandboxed

miniak:
#### Description of Change
Fixes https://github.com/electron/electron/issues/17132.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes:
* Fixed issues with popups created from sandboxed `<webview>`: 
* Clicking link with `target="_blank"` not emitting `'new-window'` event.
* `window.open()` not returning `null` when `allowpopups` is not set.

### DIFF
--- a/atom/browser/atom_browser_client.cc
+++ b/atom/browser/atom_browser_client.cc
@@ -609,11 +609,6 @@ bool AtomBrowserClient::CanCreateWindow(
 
   int opener_render_process_id = opener->GetProcess()->GetID();
 
-  if (IsRendererSandboxed(opener_render_process_id)) {
-    *no_javascript_access = false;
-    return true;
-  }
-
   if (RendererUsesNativeWindowOpen(opener_render_process_id)) {
     if (RendererDisablesPopups(opener_render_process_id)) {
       // <webview> without allowpopups attribute should return
--- a/atom/browser/web_contents_preferences.cc
+++ b/atom/browser/web_contents_preferences.cc
@@ -147,6 +147,8 @@ WebContentsPreferences::WebContentsPreferences(
 #endif
   SetDefaultBoolIfUndefined(options::kOffscreen, false);
 
+  SetDefaults();
+
   // If this is a <webview> tag, and the embedder is offscreen-rendered, then
   // this WebContents is also offscreen-rendered.
   int guest_instance_id = 0;
@@ -172,6 +174,12 @@ WebContentsPreferences::~WebContentsPreferences() {
                    instances_.end());
 }
 
+void WebContentsPreferences::SetDefaults() {
+  if (IsEnabled(options::kSandbox)) {
+    SetBool(options::kNativeWindowOpen, true);
+  }
+}
+
 bool WebContentsPreferences::SetDefaultBoolIfUndefined(
     const base::StringPiece& key,
     bool val) {
@@ -183,6 +191,10 @@ bool WebContentsPreferences::SetDefaultBoolIfUndefined(
     preference_.SetKey(key, base::Value(val));
     return val;
   }
+}
+
+void WebContentsPreferences::SetBool(const base::StringPiece& key, bool value) {
+  preference_.SetKey(key, base::Value(value));
 }
 
 bool WebContentsPreferences::IsEnabled(const base::StringPiece& name,
@@ -197,6 +209,8 @@ bool WebContentsPreferences::IsEnabled(const base::StringPiece& name,
 void WebContentsPreferences::Merge(const base::DictionaryValue& extend) {
   if (preference_.is_dict())
     static_cast<base::DictionaryValue*>(&preference_)->MergeDictionary(&extend);
+
+  SetDefaults();
 }
 
 void WebContentsPreferences::Clear() {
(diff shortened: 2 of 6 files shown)


miniak:
cc @chrismohr

lib/browser/guest-view-manager.js
```diff
@@ -149,7 +149,8 @@ const createGuest = function (embedder, params) {
   // Forward internal web contents event to embedder to handle
   // native window.open setup
   guest.on('-add-new-contents', (...args) => {
-    if (guest.getLastWebPreferences().nativeWindowOpen === true) {
+    const webPreferences = guest.getLastWebPreferences()
+    if (webPreferences.nativeWindowOpen === true || webPreferences.sandbox === true) {
```
MarshallOfSound: Does `sandbox` truly imply `nativeWindowOpen`?  I think we have checks for `nativeWindowOpen` in a few places around the code base.  We should probably adjust the web prefs impl to set `nativeWindowOpen` to `true` if `sandbox` is `true`

miniak: We don't execute `window-setup` in sandbox. The other places are: https://github.com/electron/electron/commit/aa241771a381560e586456850119245cbdca2c9e

miniak: we have comments like
> "window.open", used in sandbox and nativeWindowOpen mode.

atom/browser/atom_browser_client.cc
```diff
@@ -608,11 +608,6 @@ bool AtomBrowserClient::CanCreateWindow(
 
   int opener_render_process_id = opener->GetProcess()->GetID();
 
-  if (IsRendererSandboxed(opener_render_process_id)) {
```
MarshallOfSound: Why did this go away?

miniak: @MarshallOfSound it's now covered by `RendererUsesNativeWindowOpen`

miniak: `sandbox: true` also sets `nativeWindowOpen: true` in `webPreferences`

otherwise, this would have to be
```
  if (IsRendererSandboxed(opener_render_process_id) || RendererUsesNativeWindowOpen(opener_render_process_id)) {
    if (RendererDisablesPopups(opener_render_process_id)) {
      // <webview> without allowpopups attribute should return
      // null from window.open calls
      return false;
    } else {
      *no_javascript_access = false;
      return true;
    }
  }
```

MarshallOfSound: Sweet thanks, didn't trace that, good to have the context 👍

atom/browser/web_contents_preferences.cc
```diff
@@ -147,6 +147,8 @@ WebContentsPreferences::WebContentsPreferences(
 #endif
   SetDefaultBoolIfUndefined(options::kOffscreen, false);
 
+  Merge(base::DictionaryValue());
```
zcbenz: What does this `Merge` do?

miniak: merge calls this:
```
  if (IsEnabled(options::kSandbox)) {
    SetBool(options::kNativeWindowOpen, true);
  }
```

miniak: should I move that into a separate method instead and call it from both the ctor and Merge?

miniak:
@MarshallOfSound removed backport labels for 4 and 5

atom/browser/web_contents_preferences.cc
zcbenz: Can you add a comment for this line? It looks like a no-op at first glance.

miniak: Can you check this one? https://github.com/electron/electron/pull/18369
Seems like it would be cleaner to just move the whole section setting the defaults into a separate method and call it both from the ctor and merge.

miniak: I've moved it to a separate `SetDefaults` method for better clarity, similar to the other PR

miniak:
@MarshallOfSound are you ok with the PR? It's only targeting 6.0 for backporting.

miniak:
@nornagon can you please have a look?

nornagon:
Should this PR wait for #18369 to be merged first?

miniak:
@nornagon I would merge this one first and we would like to backport it to 6.0 to avoid conflicts. And I also need to address your concern in the other PR and potentially reimplement it differently taking into account your suggestion.

nornagon:
Yeah, refactoring the implementation was the thing I wanted this rebased on top of, because as it stands I find this change very difficult to reason about and as such I'm unsure if it's correct.

miniak:
@nornagon ideally yes, but as I've mentioned, I would like to backport this fix. Therefore I don't think it should depend on the refactoring. I agree that it is a bit complicated to reason about it. The tests that I've added are covering the change and confirming that the fix works.

nornagon:
@alexeykuzmin I updated the release notes in the release-clerk bot message. When using multi-line release notes, each note should stand on its own I think. Let me know if I wrote anything inaccurate (or feel free to update it yourself).

miniak:
@nornagon, @zcbenz do you agree with @MarshallOfSound that backporting to 4/5 is risky, even though it's a fix?

zcbenz:
I'm neutral on this, backporting it doesn't seem very risky to me though.

Issue #14617, opened 2018-09-13
Provide binary releases without bundled MSVC++ and UCRT Runtimes


Xaymar-Streamlabs:
**Is your feature request related to a problem? Please describe.**
In 2.x and 3.x releases, electron currently ships with a bundled MSVC++ and UCRT Runtime, which cause native C++ modules that were built with Microsoft Visual Studio 2015 or 2017 to behave differently than expected. Even if the user installs the latest version on their system, the local copy next to the electron.exe is loaded first, resulting in unexpected behavior.

For our case ([Streamlabs OBS](https://github.com/stream-labs/streamlabs-obs) with native module [OBS Studio for Node.JS](https://github.com/stream-labs/obs-studio-node)) this results in C++11 (and newer) features being locked away, as they can cause instability. We have observed std::stringstream crashing the application with a heap corruption due to this rather old runtime being shipped with electron.

**Describe the solution you'd like**
A binary release without bundled MSVC and UCRT Runtimes as an alternative to those with them bundled. This will allow us to properly use C++11 (and newer) features in native modules without being restricted by a very old, possibly buggy and unsafe, MSVC and UCRT library. 

**Describe alternatives you've considered**
We've attempted to use 3.0.0-beta which ships with 2017 binaries, however early tests however did not work out due to a bug that prevented our code from working in 3.0.0-beta. In the short term, I'll try to modify electron-builder to forcefully remove those files as they at times make debugging impossible (electron 2.0.8 has a vcruntime140.dll that doesn't exist in the Microsoft Symbol Server or Electron Symbol Server - Visual Studio 2017 can't find it in either).

**Steps to reproduce the C++11 issue on electron 2.0.x**

1. Create a native module
2. Put the following code into a function callable from JavaScript:
```
std::stringstream helloworld;
helloworld << "hello";
helloworld << " ";
helloworld << "world";
std::string mystring = helloworld.str();
```
3. Run the function from JavaScript
4. Crash as soon as Windows notices that the heap memory got corrupted.

nornagon:
Electron 4.0.x will be linked with a statically-linked CRT (i.e. built with `/MT`), so the CRT dlls won't be included. That seems like it might cause a different set of compatibility problems, though. I can't imagine that mixing CRTs in the same process is going to end well...

computerquip-streamlabs:
[...]

PR #18281, opened 2019-05-13
fix: don't export private V8 symbols that can cause native node modules to crash

torycl:
#### Description of Change
This change stops private V8 symbols and internal CRT methods being exported. It fixes an issue where native node modules can import incorrect CRT methods and crash on Windows.

- Windows specific 
- Directly disables exporting all symbols decorated with `V8_EXPORT_PRIVATE`
- Removes `std::basic_ostream` from `node.lib` which causes crashes if node modules use `std::ostream`
- Removes `v8::internal` from node.lib
- Makes electron binary smaller
- Native node module can be now linked with dynamic CRT on Windows

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Removed incorrectly published internal V8 symbols and CRT methods from node.lib, causing heap corruptions with node modules using the dynamic CRT on Windows.


patches/common/v8/do_not_export_private_v8_symbols_on_windows.patch
```diff
@@ -0,0 +1,22 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Tomas Rycl <torycl@microsoft.com>
+Date: Mon, 13 May 2019 15:48:48 +0200
+Subject: Do not export private V8 symbols on Windows
+
+
```
nornagon: NB. we require that every patch contains an explanation of why it is necessary and can't be implemented without needing a patch. Something like:

> Since Electron forces BUILDING_V8_SHARED, but doesn't need to export V8's private symbols (since we don't run V8's tests in component mode), we prevent private symbols from being exported here. This reduces the number of symbols that electron.lib exports and consequently [DESCRIBE VALUE OF REDUCING SYMBOL COUNT]

Also maybe consider merging this with the other patch for BUILDING_V8_SHARED?

torycl: Thanks for comment. I am now in the middle of testing. I still don't know if this is the best approach. I definitely need to reduce number of symbols and remove stl ones. But I don't know if this is the right way to do this. That is the reason this is just a draft. I needed to see what will happen if I try to disable private V8 symbols.

miniak:
Related to https://github.com/electron/electron/issues/14617#issuecomment-480313029

nornagon:
@miniak that comment is about STL symbols, which this PR doesn't affect.

patches/common/v8/do_not_export_private_v8_symbols_on_windows.patch
nornagon: Does reducing the number of symbols have any performance effect?

ckerr: Performance isn't the only potential reason for this PR, but @torycl I agree with @nornagon that it would be great if you'd update the PR description to discuss the problem you're trying to solve -- e.g. does exporting stl symbols cause conflicts somewhere?

torycl:
> @miniak that comment is about STL symbols, which this PR doesn't affect.

Actually this PR is just about STL symbols. std::basic_ostream is in description of this PR from the beginning. I am trying to fix issue @miniak reported. That is the main goal of my work.
Removing V8 private symbols will remove these STL symbols.

patches/common/v8/do_not_export_private_v8_symbols_on_windows.patch
torycl: Yes. It causes crashes of node modules which just use std::basic_ostream and dll crt on Windows. I am trying to fix this by filtering number of symbols which will make it to final node.lib.

torycl:
> Oh huh. How does exporting the `V8_EXPORT_PRIVATE` symbols cause those STL symbols to be exported..? Perhaps I'm missing something about how dllexport works 🤔

Exactly! I don't know. I have just tested it fixes our issues and it makes Electron 4 in Teams Desktop App much better. This is the reason I am quite hesitant to publish this change to full PR. Fist, I would like to know if this will be useful for somebody else. Or, in the meantime, I can find some better solution.

nornagon:
Looks like the build is broken because master is broken. See #18443

patches/common/v8/do_not_export_private_v8_symbols_on_windows.patch
```diff
@@ -0,0 +1,58 @@
...
++#else //!(BUILDING_V8_SHARED || USING_V8_SHARED)
+ #define V8_EXPORT_PRIVATE
+-#endif  // BUILDING_V8_SHARED
++#endif
++#endif
+ 
+ #else  // V8_OS_WIN
+ 
+-- 
+2.21.0
```
nornagon: If you're using `git-export-patches`, the version number shouldn't be present in the patch.

patches/common/v8/do_not_export_private_v8_symbols_on_windows.patch
```diff
@@ -0,0 +1,58 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Tomas Rycl <torycl@microsoft.com>
+Date: Mon, 13 May 2019 15:48:48 +0200
+Subject: Do not export private V8 symbols on Windows
+
+This change stops private V8 symbols and internal crt methods being exported.
+It fixes an issue where native node modules can import
+incorrect CRT methods and crash on Windows.
+It also reduces size of node.lib by 75%.
```
nornagon:
```suggestion
It also reduces size of node.lib by 75%.

This patch can safely be removed if, when it is removed, `node.lib` does not contain any standard C++ library exports (e.g. `std::ostringstream`).
```

patches/common/v8/do_not_export_private_v8_symbols_on_windows.patch
torycl: It took me some time to find out how to use this script properly. It should be fixed now.

Issue #12745, opened 2018-04-29
APIs should use a property instead of separate getter/setter methods


sindresorhus:
Continued from https://github.com/electron/electron/pull/12496/files#r181941682

---

* Electron version: 2.0.0-beta.8
* Operating system: macOS 10.13

### Expected behavior

JavaScript supports getters and setters, so it's quite awkward that most of the Electron APIs instead use separate methods for getting and setting. This feels more like Java than JavaScript.

I propose that new APIs start using getters and setters (for example https://github.com/electron/electron/pull/12496), when it makes sense, and that eventually old APIs moves to it too.

For example, the `app.getName()`/`app.setName()` methods could simply be:

```js
const {app} = require('electron');

console.log(app.name);
//=> 'Electron'

app.name = 'Unicorn';

console.log(app.name);
//=> 'Unicorn'
```

### Actual behavior

This is how it currently works:

```js
const {app} = require('electron');

console.log(app.getName());
//=> 'Electron'

app.setName('Unicorn');

console.log(app.getName());
//=> 'Unicorn'
```

dmckenna24:
Agreed, why waste up more lines of code when JavaScript does the work for you? Anyone tackling this issue?

codebytere:
@sindresorhus at the moment, we don't do that because we have some instances (like `getBounds()`) where the property would be an object, so we couldn't handle deep modifications like `mainWindow.bounds.x`. 

I agree that it's definitely more idiomatic, so once we find a clean way to account for aforementioned cases i think this is something we could definitely do.

LinusU:
@codebytere should be easily fixed with a `Proxy`? 🙌 

Something like

```js
class Window {
  get bounds () {
    return this._bounds
  }

  set bounds (newValue) {
    this._bounds = new Proxy(newValue, {
      set: (obj, prop, value) => {
        Reflect.set(obj, prop, value)
        this._boundsChanged()
      }
    })

    this._boundsChanged()
  }
}
```

MarshallOfSound:
@LinusU 

> easily fixed with a `Proxy`

Not really in a scalable and generic way, sure we could wrap everything in a Proxy in JS land but then you have this weird bridge between a JS proxy and native land.  You would also need to wrap things in JS for **all** the API's that do this.  Even if we do `v8::Proxy` stuff in native land it's going to be messy code

We're definitely going to look into this, but it isn't as simple as "let's put a proxy everywhere" unfortunately.

[...]

PR #18282, opened 2019-05-13
refactor: change autoUpdater to use feedURL property

jkleinsc:
#### Description of Change

This PR is part of the[ API modernization initiative](https://github.com/electron/electron/blob/master/docs/api/modernization/property-updates.md).  This PR changes the `autoUpdater` API to use a `feedURL` property instead of `getFeedURL` and `setFeedURL`.  Additionally, what was the `setFeedURL` function has been refactored as the `initialize` function because it was an overloaded function that was no longer properly named.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: changed autoUpdater to use feedURL property instead of getFeedURL/setFeedURL.

### DIFF
--- a/atom/browser/api/atom_api_auto_updater.cc
+++ b/atom/browser/api/atom_api_auto_updater.cc
@@ -105,8 +105,12 @@ void AutoUpdater::OnWindowAllClosed() {
   QuitAndInstall();
 }
 
-void AutoUpdater::SetFeedURL(mate::Arguments* args) {
-  auto_updater::AutoUpdater::SetFeedURL(args);
+void AutoUpdater::SetFeedURL(const std::string& feedURL) {
+  auto_updater::AutoUpdater::SetFeedURL(feedURL);
+}
+
+void AutoUpdater::Initialize(mate::Arguments* args) {
+  auto_updater::AutoUpdater::Initialize(args);
 }
 
 void AutoUpdater::QuitAndInstall() {
@@ -134,8 +138,11 @@ void AutoUpdater::BuildPrototype(v8::Isolate* isolate,
   prototype->SetClassName(mate::StringToV8(isolate, "AutoUpdater"));
   mate::ObjectTemplateBuilder(isolate, prototype->PrototypeTemplate())
       .SetMethod("checkForUpdates", &auto_updater::AutoUpdater::CheckForUpdates)
-      .SetMethod("getFeedURL", &auto_updater::AutoUpdater::GetFeedURL)
-      .SetMethod("setFeedURL", &AutoUpdater::SetFeedURL)
+      .SetMethod("initialize", &AutoUpdater::Initialize)
+      .SetProperty("feedURL", &auto_updater::AutoUpdater::GetFeedURL,
+                   &AutoUpdater::SetFeedURL)
+      .SetMethod("_getFeedURL", &auto_updater::AutoUpdater::GetFeedURL)
+      .SetMethod("_setFeedURL", &AutoUpdater::SetFeedURL)
       .SetMethod("quitAndInstall", &AutoUpdater::QuitAndInstall);
 }
 
--- a/atom/browser/auto_updater_mac.mm
+++ b/atom/browser/auto_updater_mac.mm
@@ -31,6 +31,8 @@ SQRLUpdater* g_updater = nil;
 namespace {
 
 bool g_update_available = false;
+AutoUpdater::HeaderMap headers_;
+std::string serverType_ = "default";
 std::string update_url_ = "";  // NOLINT(runtime/string)
 
 }  // namespace
@@ -39,8 +41,13 @@ std::string AutoUpdater::GetFeedURL() {
   return update_url_;
 }
 
+void AutoUpdater::SetFeedURL(std::string feedURL) {
+  update_url_ = feedURL;
+  AutoUpdater::SetFeedOptions(update_url_, serverType_, headers_);
+}
+
 // static
-void AutoUpdater::SetFeedURL(mate::Arguments* args) {
+void AutoUpdater::Initialize(mate::Arguments* args) {
   mate::Dictionary opts;
   std::string feed;
   HeaderMap requestHeaders;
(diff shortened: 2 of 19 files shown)


jkleinsc:
@MarshallOfSound I'm happy to defer this to a longer deprecation period.  I was wondering if we should put the deprecation in 6 versus 7, but there's no hurry needed here.  I'll update the specs with the deprecated calls and update the relevant docs as well.

docs/api/auto-updater.md
```diff
@@ -97,23 +97,33 @@ When this API is called, the `before-quit` event is not emitted before all windo
...
   * `url` String
   * `headers` Object (optional) _macOS_ - HTTP request headers.
   * `serverType` String (optional) _macOS_ - Either `json` or `default`, see the [Squirrel.Mac][squirrel-mac]
     README for more information.
 
-Sets the `url` and initialize the auto updater.
+Sets the `url` and initializes the auto updater.
 
-### `autoUpdater.getFeedURL()`
+### `autoUpdater.setFeedURL(options)`  _(Deprecated)_
```
codebytere: Can you update this `Deprecated` styling to match style previously used for other property deprecations?

atom/browser/api/atom_api_auto_updater.cc
```diff
@@ -105,8 +105,12 @@ void AutoUpdater::OnWindowAllClosed() {
   QuitAndInstall();
 }
 
-void AutoUpdater::SetFeedURL(mate::Arguments* args) {
-  auto_updater::AutoUpdater::SetFeedURL(args);
+void AutoUpdater::SetFeedURL(std::string feedURL) {
```
MarshallOfSound: This can be a const reference

docs/api/auto-updater.md
```diff
@@ -97,6 +97,18 @@ When this API is called, the `before-quit` event is not emitted before all windo
 
 The `autoUpdater` object has the following methods:
 
+### `autoUpdater.initialize(options)`
+
+* `options` Object
+  * `url` String
+  * `headers` Object (optional) _macOS_ - HTTP request headers.
```
MarshallOfSound:
```suggestion
  * `headers` Record<string, string> (optional) _macOS_ - HTTP request headers.
```

docs/api/auto-updater.md
```diff
@@ -97,6 +97,18 @@ When this API is called, the `before-quit` event is not emitted before all windo
...
+  * `serverType` String (optional) _macOS_ - Either `json` or `default`, see the [Squirrel.Mac][squirrel-mac]
+    README for more information.
+
+Sets the `url` and initializes the auto updater.
+
+**[Deprecated](modernization/property-updates.md)**
```
MarshallOfSound: `initialize` isn't deprecated right, it's the new API?

docs/api/breaking-changes.md
```diff
@@ -64,6 +64,17 @@ webFrame.setIsolatedWorldInfo(
   })
 ```
 
+### `autoUpdater.setFeedURL`
```
MarshallOfSound: Can these be moved to _at least_ planned to break in Electron 8 🤔 preferably 9 now we're on the 12 week cycle

docs/tutorial/updates.md
```diff
@@ -90,7 +90,7 @@ Next, construct the URL of the update server and tell
 const server = 'https://your-deployment-url.com'
 const feed = `${server}/update/${process.platform}/${app.getVersion()}`
 
-autoUpdater.setFeedURL(feed)
+autoUpdater.initialize({ url: feed })
```
MarshallOfSound:
```suggestion
autoUpdater.feedURL = feed
```

atom/browser/auto_updater_mac.mm
```diff
@@ -39,8 +41,13 @@
...
 }
 
+void AutoUpdater::SetFeedURL(std::string feedURL) {
+  update_url_ = feedURL;
+  AutoUpdater::SetFeedOptions(update_url_, serverType_, headers_);
+}
+
 // static
-void AutoUpdater::SetFeedURL(mate::Arguments* args) {
+void AutoUpdater::Initialize(mate::Arguments* args) {
```
MarshallOfSound: If I'm reading this right this usage is not possible.

```js
autoUpdater.feedUrl = 'my.feed.url'
autoUpdater.checkForUpdates()
```

i.e. the call to initialize is required, that's not immediately obvious and I think we should remove the footgun by making both `SetFeedURL` and `Initialize` set up the `g_updater`

docs/api/breaking-changes.md
codebytere: yeah, i think we actually should just centralize that info in the `property-updates.md` file and not put it in `breaking-changes.md`

atom/browser/auto_updater_mac.mm
jkleinsc: @MarshallOfSound it is possible.  `AutoUpdater::SetFeedOptions` sets up the `g_updater`.  The only difference between `autoUpdater.feedUrl = 'my.feed.url'` and `autoUpdater.initialize(...)` is that you can pass server type and headers to initialize

docs/api/breaking-changes.md
jkleinsc: They should be in a new 8.0 section instead of the 7.0 section

atom/browser/auto_updater_mac.mm
MarshallOfSound: Ah I see, missed that third method 👍

atom/browser/api/atom_api_auto_updater.cc
```diff
@@ -105,8 +105,12 @@ void AutoUpdater::OnWindowAllClosed() {
   QuitAndInstall();
 }
 
-void AutoUpdater::SetFeedURL(mate::Arguments* args) {
-  auto_updater::AutoUpdater::SetFeedURL(args);
+void AutoUpdater::SetFeedURL(const std::string feedURL) {
```
MarshallOfSound:
```suggestion
void AutoUpdater::SetFeedURL(const std::string& feedURL) {
```

Can this be a reference instead of copying?

jkleinsc:
Closing as based on discussion in https://github.com/electron/electron/issues/12745#issuecomment-501376367 we are really only looking to do this for properties that are primitive properties.  `setFeedURL` as it stands in autoUpdater is more that a primitive property.

PR #18239, opened 2019-05-09
chore: update @types/node to match with version embedded in Electron 3.x

shiftkey:
#### Description of Change

Backport of #16174

Electron 3 embeds Node `10.2.0`, but the corresponding `@types/node` declaration is still on `8.something`.

It looks like the original PR #16174 was backported to v4 in #16177 but there wasn't a backport to v3, so here's me trying to do this.

I thought I'd be able to hack around this using yarn's `resolutions` trick:

```
  "resolutions": {
    "**/@types/node": "10.1.4"
  }
```

 But even that is not happy with me:

Maybe it works anyway? 

Not sure what other checks I need to do while I'm in here to catch any potential regressions, but I'll see what CI does first.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- ~~[ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)~~
- ~~[ ] relevant documentation is changed or added~~
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Update @types/node dependency to version compatible with Node 10.2.0

PR #18283, opened 2019-05-13
chore: update @types/node version for 3-1-x

MarshallOfSound:
Backport of #16174

A direct copy of #18239 just running on first-party CI instead of external-contrib CI

Notes: Update @types/node dependency to a version compatible with Node 10.2.0


shiftkey:
@MarshallOfSound is it worth rebasing and re-running this if #18426 seems to be Working Fine™ to see if it can land?

PR #18289, opened 2019-05-14
refactor: netLog directly uses network service

nornagon:
#### Description of Change
This refactors the `netLog` API to directly use the netlog interface exposed by the network service, instead of going through `NetExportFileWriter`. As part of the refactor:

1. This fixes an issue where calling `netLog.startLogging()` immediately in `app.on('ready')` would silently fail to start the netlog.
2. `netLog.startLogging()` now returns a promise, which you can await and be sure that the net log has started recording.
3. `netLog.currentlyRecordingPath` has been deprecated. If you need access to the currently recording path, you'll have to track it yourself.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes:
* Fixed an issue where `netLog.startLogging()` would silently fail when called immediately during `app.on('ready')`.
* `netLog.startLogging()` now returns a promise which resolves when the net log has started recording.
* Deprecated `netLog.currentlyLoggingPath`.


atom/browser/api/atom_api_net_log.cc
```diff
@@ -21,103 +22,145 @@
...
   // TODO(deepak1556): Provide more flexibility to this module
   // by allowing customizations on the capturing options.
-  net_log_writer_->StartNetLog(
-      log_path, net::NetLogCaptureMode::Default(),
-      net_log::NetExportFileWriter::kNoLimit /* file size limit */,
-      base::CommandLine::ForCurrentProcess()->GetCommandLineString(),
-      std::string(), network_context);
-}
+  auto capture_mode = network::mojom::NetLogCaptureMode::DEFAULT;
+  auto max_file_size = network::mojom::NetLogExporter::kUnlimitedFileSize;
```
deepak1556: Can we inline these at the call site, I don't see the variables being much useful. Any reason to have these declared separately ?

atom/browser/api/atom_api_net_log.cc
```diff
@@ -21,103 +22,145 @@
...
+void NetLog::StartNetLogAfterCreateFile(
+    network::mojom::NetLogCaptureMode capture_mode,
+    uint64_t max_file_size,
+    base::Value custom_constants,
+    util::Promise promise,
+    base::File output_file) {
+  if (!net_log_exporter_) {
+    // Theoretically the mojo pipe could have been closed by the time we get
+    // here via the connection error handler.
+    promise.RejectWithErrorMessage("Failed to create net log exporter");
```
deepak1556: Wouldn't it be better to handle this in `NetLog::OnConnectionError` ? The main reason I want to do that is because `NetLog::StartNetLogAfterCreateFile` is a task invoked after a task in the file task runner is done with, but the mojo connection error handler is on the main task runner, and both these tasks aren't sequenced, so there can be a race.

atom/browser/api/atom_api_net_log.cc
nornagon: My reason for breaking these out is that I plan to make them configurable in a follow-up PR.

atom/browser/api/atom_api_net_log.cc
nornagon: Yep, that's the reason to have this check here. Here's the sequence of operations this protects against:
Untitled

`OnConnectionError` doesn't have access to the `Promise` so it can't reject the error.

deepak1556: If `file created` task happens before `!!error!!` task, will it still be resolved to a failure ? Thats the case I was worried about.

> OnConnectionError doesn't have access to the Promise so it can't reject the error.

Yeah, I was imagining the `NetLog` class can manage the promise so that the rejection can happen in this method, but I am not sure if thats a good design.

nornagon: Hm.. if the file is created first, then we have a valid `net_log_exporter_` to send a `Start` message on. If the `error` task happens after that, the promise will never be resolved. I'm not sure what situation would cause that behaviour though... network service crash maybe? network context shutdown?

deepak1556: > network service crash maybe? network context shutdown?

yup those are the possibilities I am imagining, just want to make sure we handle these scenarios gracefully.

nornagon: Perhaps this sequence of events could cause the promise to never be resolved:

1. Successfully create file
2. Send `Start` message to logger
3. Before the logger can report that it's started logging, there's a connection error on the mojo pipe (maybe because of a network service internal error of some kind, or possibly due to other things happening e.g. windows closing or shutdown in-progress)
4. the mojo pipe is closed and the promise is destroyed without being resolved

I guess in order to resolve this we need to stash the pending `startLogging` promise around somewhere and properly reject it in the connection error case.

deepak1556: Yup, i was thinking to stash the promise in the `NetLog` class and use that to resolve. Based on our api design, seems like we have to deal with only one promise at a time so this might be possible ?

atom/browser/api/atom_api_net_log.cc
```diff
@@ -21,103 +22,155 @@
...
+void NetLog::StartNetLogAfterCreateFile(
+    network::mojom::NetLogCaptureMode capture_mode,
+    uint64_t max_file_size,
+    base::Value custom_constants,
+    base::File output_file) {
+  DCHECK(pending_start_promise_);
+  if (!net_log_exporter_) {
+    // Theoretically the mojo pipe could have been closed by the time we get
+    // here via the connection error handler.
+    std::move(*pending_start_promise_)
```
deepak1556: This is no longer required, since `OnConnectionError` now handles it ?

nornagon:
mac failure is flake; merging.

Issue #18310, opened 2019-05-15
Frameless window prevents taskbar unhide


adill:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 6.0.0-beta.3
* **Operating System:**
  * Windows 10 (1809)
* **Last Known Working Electron version:**:
  * 5.0.1

### Expected Behavior
Dragging your mouse to the bottom of the screen with a maximized frameless window and a hidden Windows taskbar should make the taskbar appear.

### Actual Behavior
The taskbar does not appear.

### To Reproduce

* Ensure you have Taskbar hiding enabled
![image](https://user-images.githubusercontent.com/144914/57793291-d12d9000-76f5-11e9-8c63-178e928a5358.png)

```sh
$ git clone https://github.com/adill/electron-quick-start -b no-taskbar-unhide
$ npm install
$ npm start || electron .
```

* Move your mouse to the bottom of the screen

PR #18321, opened 2019-05-16
fix: leave space for autohide taskbar for frameless maximized window

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/18310.

When autohide taskbar is enabled, the window needs to leave some space for it when maximized. Chromium already does the trick for normal window, however it is not doing the trick for frameless window. 

Since Chromium is deprecating Chrome Apps and it does not use frameless windows in other parts of code, I don't think Chromium will fix it for us, and patching Chromium may not be a good idea since all frameless window code may gradually stop working.

So in this PR I followed Chromium's trick on resizing the maximized window to leave space for autohide taskbar when using frameless window. It should be able to work through time no matter what Chromium does with its frameless window code.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix autohide taskbar not showing when frameless window is maximized.

### DIFF
--- a/atom/browser/native_window_views_win.cc
+++ b/atom/browser/native_window_views_win.cc
@@ -2,14 +2,19 @@
 // Use of this source code is governed by the MIT license that can be
 // found in the LICENSE file.
 
+#include <dwmapi.h>
+#include <shellapi.h>
+
 #include "atom/browser/browser.h"
 #include "atom/browser/native_window_views.h"
 #include "atom/browser/ui/views/root_view.h"
 #include "atom/common/atom_constants.h"
 #include "content/public/browser/browser_accessibility_state.h"
 #include "ui/base/win/accessibility_misc_utils.h"
+#include "ui/display/display.h"
 #include "ui/display/win/screen_win.h"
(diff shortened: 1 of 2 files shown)


atom/browser/native_window_views_win.cc
```diff
@@ -138,11 +143,155 @@ bool IsScreenReaderActive() {
...
+void NativeWindowViews::Maximize() {
+  int autohide_edges = 0;
+  if (!has_frame())
+    autohide_edges = GetAppbarAutohideEdges(GetAcceleratedWidget());
+
+  // Only use Maximize() when:
+  // 1. window has WS_THICKFRAME style;
+  // 2. and window is not frameless when there is autohide taskbar.
+  if ((::GetWindowLong(GetAcceleratedWidget(), GWL_STYLE) & WS_THICKFRAME) &&
+      (has_frame() || autohide_edges == 0)) {
```
ckerr: This specific conditional is a little busy but on a re-read, I the usual refactors don't result in much of an improvement. :neutral_face:

PR #18326, opened 2019-05-16
docs: use "Breaking Changes" as the changes doc's title

ckerr:
#### Description of Change

Multiple people have commented that this important page is difficult to find, in part because people search for "Breaking" or "Changes" or "Breaking Changes" rather than "API Contract".

This PR changes the name to "Breaking Changes" so that it'll show up on electronjs.org/docs/.

It also adds a backwards-friendly link so people still searching for "API Contract" will get a redirect. I'm not sure if this is necessary -- the actual URL of https://electronjs.org/docs/api/breaking-changes is
unchanged -- would especially welcome feedback on this part.

CC @BinaryMuse @Kilian @HashimotoYT 

#### Checklist

- [ ] `npm test` passes
- [x] ~tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)~

#### Release Notes

Notes: no-notes

### DIFF
--- a/docs/api/api-contract.md
+++ b/docs/api/api-contract.md
@@ -1,3 +1,0 @@
-# API Contract
-
-Moved to the [Breaking Changes](breaking-changes.md) doc.
--- a/docs/api/api-contract.md
+++ b/docs/api/api-contract.md
@@ -1,0 +1,3 @@
+# API Contract
+
+Moved to the [Breaking Changes](breaking-changes.md) doc.
--- a/docs/api/breaking-changes.md
+++ b/docs/api/breaking-changes.md
@@ -1,4 +1,4 @@
-# API Contract
+# Breaking Changes
 
 Breaking changes will be documented here, and deprecation warnings added to JS code where possible, at least [one major version](../tutorial/electron-versioning.md#semver) before the change is made.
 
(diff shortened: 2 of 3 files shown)


docs/api/api-contract.md
```diff
@@ -0,0 +1,3 @@
+# API Contract
+
+Moved to the [Breaking Changes](breaking-changes.md) doc.
```
MarshallOfSound: I'm pretty sure theirs a way to configure redirects on the website repo so we won't need this placeholder

ckerr: > It also adds a backwards-friendly link so people still searching for "API Contract" will get a redirect. I'm not sure if this is necessary -- the actual URL of https://electronjs.org/docs/api/breaking-changes is unchanged -- would especially welcome feedback on this part.

MarshallOfSound: Oh yeah, I don't think this is required.  I think the URL being the same is enough 👍

ckerr: Sounds good to me. Literally a one-liner now.

ckerr:
(backporting to stable 5-0-x branch to propagate this change to the website)

Issue #18309, opened 2019-05-15
Customize Electron/Chromium cache location


emmkimme:
### Problem Description
I would like to customize the location of the Electron/Chromium cache (for Windows and other OS/s). Currently cache used the same folder than getPath('userData')
2 reasons :
- having cache in the roaming user part impacts the size of the profile. When your deploy your application on system like Citrix, it has a huge cost ($) in term of space disk and cost (time) to synchronize profile.
- we store our own user data under getPath('userData'). But this folder is shared with the Electron cache, so there is a mix of Electron folders/files and of our product folders/file. It is bit confusing.

### Proposed Solution
I see in C++ code an attempt of 'cache'/ 'userCache' path implementation. But not finished. As far as I see, it would be to add a piece of code here (very roughly) :

src\electron\atom\browser\atom_browser_context.cc
`
AtomBrowserContext::AtomBrowserContext
...
>> if (!base::PathService::Get(DIR_CACHE, &path_)) {
    if (!base::PathService::Get(DIR_USER_DATA, &path_)) {
      base::PathService::Get(DIR_APP_DATA, &path_);
      path_ = path_.Append(base::FilePath::FromUTF8Unsafe(GetApplicationName()));
      base::PathService::Override(DIR_USER_DATA, path_);
    }
  }
`

### Alternatives Considered
We customized setPath('userData') at startup of our application to force the cache to another location but later in the code, we can no more use getPath('userData') for storing our data. We expose a new method but it is less friendly and error prone.

### Additional Information
I have not the background /context of the 1st attempt, it may have a good reason to stop. If you can share your concerns and see if it is feasible, will be grateful

I can manage a MR, if you agree.

nornagon:
Duplicate of #1404

PR #18331, opened 2019-05-17
feat: customize Electron cache

emmkimme:
#### Description of Change
This MR covers the following issues:
[Customize Electron/Chromium cache location](https://github.com/electron/electron/issues/18309)
[AppData folder has large cache which affects "roaming profile" users](https://github.com/desktop/desktop/issues/6523)

Notes: Allow to customize Electron/Chromium cache location

#### Checklist
- [x] `npm test` passes
- [x] tests are unchanged 

#### Technical Notes
The notion of cache (DIR_CACHE, DIR_USER_CACHE) is already present in the code but not used.

atom\browser\api\atom_api_app.cc
```
int GetPathConstant(const std::string& name) {
...
  else if (name == "cache")
    return DIR_CACHE;
  else if (name == "userCache")
    return DIR_USER_CACHE;
...
```

atom/browser/atom_paths.h
```
...
  DIR_USER_CACHE,              // Directory where user cache can be written.
#if defined(OS_POSIX)
  DIR_CACHE = base::DIR_CACHE,  // Directory where to put cache data.
#else
  DIR_CACHE = base::DIR_APP_DATA,
#endif
...
```
lib\browser\api\app.ts
```
app._setDefaultAppPaths = (packagePath) => {
...
  app.setPath('userData', path.join(app.getPath('appData'), app.name!))
  app.setPath('userCache', path.join(app.getPath('cache'), app.name!))
```

There is 2 levels of cache : 
- userCache (DIR_USER_CACHE) and cache (DIR_CACHE) 
equivalent to the 2 levels of user data
- userData (DIR_USER_DATA) and appData (DIR_APP_DATA) 

Purpose was to use the userCache value for customizing the cache

atom\browser\atom_browser_context.cc
Before
```
if (!base::PathService::Get(DIR_USER_DATA, &path_)) {
    base::PathService::Get(DIR_APP_DATA, &path_);
    path_ = path_.Append(base::FilePath::FromUTF8Unsafe(GetApplicationName()));
    base::PathService::Override(DIR_USER_DATA, path_);
  }
```

After
```
  if (!base::PathService::Get(DIR_USER_CACHE, &path_)) {
    base::PathService::Get(DIR_CACHE, &path_);
    path_ = path_.Append(base::FilePath::FromUTF8Unsafe(GetApplicationName()));
    base::PathService::Override(DIR_USER_CACHE, path_);
  }
```

#### Usage
```
import { app } from 'electron';

// Must be called before application is ready
app.setPath('userCache', myOwnUserCache);
```

#### Additional notes
For concurrent access violation, the cache folder must not be shared by different users.
For performance purpose, it must be set on an efficient drive.

We can even imagine to got it deep further and have a DIR_CACHE independent as well. So user can either change the cache (and let the system append the application name) or change the userCache


emmkimme:
I restore files ending (I'm always in pain with that).
You are right, it will not work anymore for people using userData for moving the cache., I did not realized. 
But using userData for changing the Electron Cache location was a desperate solution.
To not break this, I can define userCache empty by default and fallback to previous implementation in this case.
Do you prefer ?

atom/browser/atom_browser_context.cc
```diff
@@ -78,10 +78,10 @@ AtomBrowserContext::AtomBrowserContext(const std::string& partition,
   base::StringToInt(command_line->GetSwitchValueASCII(switches::kDiskCacheSize),
                     &max_cache_size_);
 
-  if (!base::PathService::Get(DIR_USER_DATA, &path_)) {
-    base::PathService::Get(DIR_APP_DATA, &path_);
+  if (!base::PathService::Get(DIR_USER_CACHE, &path_)) {
+    base::PathService::Get(DIR_CACHE, &path_);
     path_ = path_.Append(base::FilePath::FromUTF8Unsafe(GetApplicationName()));
-    base::PathService::Override(DIR_USER_DATA, path_);
+    base::PathService::Override(DIR_USER_CACHE, path_);
```
ckerr: The docs indicate that the intent is for `DIR_USER_DATA` and `DIR_USER_CACHE` to _both_ exist side-by-side, but this PR removes the code that sets the fallback value of `DIR_USER_DATA`. Is that intentional?

docs/api/app.md
```diff
@@ -641,9 +643,9 @@ created by this method. On failure an `Error` is thrown.
 
 You can only override paths of a `name` defined in `app.getPath`.
 
-By default, web pages' cookies and caches will be stored under the `userData`
+By default, web pages' cookies and caches will be stored under the `userCache`
```
ckerr: Is this actually true? Does overriding `DIR_USER_CACHE` in the AtomBrowserContext constructor actually change behavior s.t. all the cache data goes there?

atom/browser/atom_browser_context.cc
ckerr: If we're setting `DIR_USER_CACHE` we probably also want to ensure it exists, as we do for `DIR_USER_DATA` in `Browser::DidFinishLaunching()`?

docs/api/app.md
```diff
@@ -598,6 +598,8 @@ You can request the following paths by the name:
   * `~/Library/Application Support` on macOS
 * `userData` The directory for storing your app's configuration files, which by
   default it is the `appData` directory appended with your app's name.
+* `userCache` The directory for storing Electron cache configuration files, which by
```
ckerr: s/cache configuration/cache

atom/browser/atom_browser_context.cc
emmkimme: At startup DIR-CACHE is set by default to DIR_USER_DATA, so the fallback may remain. The problem is when the userData is changed, the DIR_CACHE is not sync and the cache does not follow anymore, It is what @MarshallOfSound want to fix. I agree and try to find the best solution

docs/api/app.md
emmkimme: This notion of DIR_CACHE, DIR_USER_CACHE already exist in the code but are R/O info, changing them have currently no effect.

docs/api/app.md
emmkimme: thanks will update

emmkimme:
I have a solution which should keep the backward compatibility, But unfortunately I have to introduce an additional path 'appCache'.

docs/api/app.md
```diff
@@ -604,6 +604,8 @@ You can request the following paths by the name:
   * `~/Library/Application Support` on macOS
 * `userData` The directory for storing your app's configuration files, which by
   default it is the `appData` directory appended with your app's name.
+* `appCache` The directory for storing your app's Electron cache, which by
+  default it is the `appData` directory appended with your app's name.
```
zcbenz: Document should warn user that setting `appCache` would make app incompatible with old data and they need to migrate data from old dir to new dir themselves.

zcbenz: Using `appCache` as key is quite confusing as `app`-prefix has other meanings in `appData`. Maybe `browserCache` is better.

zcbenz:
I'm good with introducing an opt-in feature to change cache directory, considering the complexity of correctly migrating old data, existing apps might not be willing to use it, but I think it is fine to give new apps an option.

/cc @MarshallOfSound @ckerr as the PR has updated.

miniak:
@emmkimme please rebase

codebytere:
@emmkimme please address comments and rebase, or this PR will be closed in 5 days due to inactivity.

emmkimme:
Hi, I have a better solution in mind which should handle backward. I will delete the branch and redo from scratch. Thks for your comments.

PR #18339, opened 2019-05-17
fix: ensure value of atom::PATH_END is valid

ckerr:
#### Description of Change

The last entries in the `atom::PATH_*` enum copy their values from `base::PATH_*` entries, so the final item in the enum -- `atom::PATH_END` -- accidentally picked up a value somewhere between `base::PATH_START` and `base::PATH_END`.

This a minor correctness patch. `atom::PATH_END` was the only enum bitten by this and it doesn't appear to be used anywhere.

Kind of a random PR. I spotted this while reading for an unrelated issue.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/atom_paths.h
+++ b/atom/browser/atom_paths.h
@@ -28,18 +28,22 @@ enum {
 
 #if defined(OS_LINUX)
   DIR_APP_DATA,  // Application Data directory under the user profile.
-#else
+#endif
+
+  PATH_END,  // End of new paths. Those that follow redirect to base::DIR_*
+
+#if !defined(OS_LINUX)
   DIR_APP_DATA = base::DIR_APP_DATA,
 #endif
 
 #if defined(OS_POSIX)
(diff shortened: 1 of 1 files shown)


atom/browser/atom_paths.h
```diff
@@ -28,19 +28,23 @@ enum {
...
+  DIR_CACHE = base::DIR_CACHE  // Directory where to put cache data.
 #else
-  DIR_CACHE = base::DIR_APP_DATA,
+  DIR_CACHE = base::DIR_APP_DATA
 #endif
-
-  PATH_END
 };
 
+static_assert(PATH_START < PATH_END, "invalid PATH boundaries");
```
codebytere: would we perhaps want to `DCHECK` here?

ckerr: I don't think so, `static_assert` is a compile-time check while `DCHECK` is runtime.

The assertion is arguably overkill; but if we're going to have a check here, it should be at compile time.

PR #18343, opened 2019-05-17
fix: correctly support the webkitdirectory input attr

### LINKED ISSUE #839: full support for multiple webkitdirectory
<input> file . multiple webkitdirectory in chrome can get sub files.but atom-shell can select directory,sub file set only current directory? is bug?

MarshallOfSound:
Fixes #839

The implementation here was loosely inspired by the implementation in
`//chrome` found in their FileSelectHelper.  I.e. That's where the usage
of `net::DirectoryList` comes from.

Refs: https://cs.chromium.org/chromium/src/chrome/browser/file_select_helper.cc

Notes: Fixed support for the `webkitdirectory` attribute on `input[type=file]` elements

### DIFF
--- a/atom/browser/web_dialog_helper.cc
+++ b/atom/browser/web_dialog_helper.cc
@@ -35,7 +35,8 @@ using blink::mojom::FileChooserParams;
 namespace {
 
 class FileSelectHelper : public base::RefCounted<FileSelectHelper>,
-                         public content::WebContentsObserver {
+                         public content::WebContentsObserver,
+                         public atom::DirectoryListerHelperDelegate {
  public:
   FileSelectHelper(content::RenderFrameHost* render_frame_host,
                    std::unique_ptr<content::FileSelectListener> listener,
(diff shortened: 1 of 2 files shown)


codebytere:
Do we also want this to target `5-0-x`? This could be perceived as a fix or a feature, depending on strictness of evaluation 🤔

zcbenz:
If an app is using the `webkitdirectory` attribute then it is currently relying on the old buggy behavior and will likely break with this change, so I think we should at least bump the minor version for it, and it may even belong to the breaking changes.

MarshallOfSound:
> then it is currently relying on the old buggy behavior and will likely break with this change

I'm firmly in the camp of "if you're relying on broken, non-spec behavior then when it's fixed that's on you to deal with".  But will defer to @codebytere / @electron/wg-releases on this one 👍

atom/browser/web_dialog_helper.cc
```diff
@@ -216,6 +258,30 @@ file_dialog::Filters GetFileTypesFromAcceptType(
...
+    const net::DirectoryLister::DirectoryListerData& data) {
+  // We don't want to return directory paths, only file paths
+  if (data.info.IsDirectory())
+    return;
+
+  paths_.push_back(data.path);
+}
+void DirectoryListerHelper::OnListDone(int error) {
+  std::vector<FileChooserFileInfoPtr> file_info;
+  for (auto path : paths_)
```
deepak1556: Why copy ? could have taken references.

atom/browser/web_dialog_helper.cc
```diff
@@ -71,23 +72,61 @@ class FileSelectHelper : public base::RefCounted<FileSelectHelper>,
...
+    Release();
+  }
+
  private:
   friend class base::RefCounted<FileSelectHelper>;
 
   ~FileSelectHelper() override {}
 
+  void EnumerateDirectory(base::FilePath base_dir) {
+    auto* lister = new net::DirectoryLister(
```
deepak1556: Whats the lifetime of the `lister` ? I don't see it being destroyed.

atom/browser/web_dialog_helper.cc
```diff
@@ -71,23 +72,61 @@ class FileSelectHelper : public base::RefCounted<FileSelectHelper>,
...
-          file_info.push_back(FileChooserFileInfo::NewNativeFile(
-              blink::mojom::NativeFileInfo::New(
-                  path, path.BaseName().AsUTF16Unsafe())));
+        // If we are uploading a folder we need to enumerate its contents
+        if (mode_ == FileChooserParams::Mode::kUploadFolder &&
+            paths.size() >= 1) {
+          base_dir = paths[0];
+
+          // Actually enumerate soemwhere off-thread
+          base::SequencedTaskRunnerHandle::Get()->PostTask(
```
deepak1556: This doesn't mean off-thread, it will be a task that runs on the same sequence as where its called but will be guaranteed to run in a defined order. Prefer `base::PostTaskWithTraits`.

jkleinsc:
Per discussion with @electron/wg-releases it was decided to not backport this to 5-0-x

mklnz:
So previously I was able to get the path of the directory selected. However with the "fixed" version I'm only able to get a list of files within that directory, how to get the directory path if there are no files inside said directory since the target.files array is empty?

maayanalbert:
I'm using npm as my package manager and my version of electron doesn't seem to contain this fix (I'm on 6.0.11). How can I access this fix? Thanks!

maayanalbert:
Thanks @MarshallOfSound for the fix! I'm using npm as my package manager and my version of electron doesn't seem to contain these changes (I'm on 6.0.11). Is there I way I can access this via npm? Thanks!

maayanalbert:
Oops just had to do `npm i -D electron@latest`, lols

PR #18361, opened 2019-05-20
feat: migrate protocol module to NetworkService (Part 8)

zcbenz:
#### Description of Change

This PR makes the `file://` protocol support asar files when using NetworkService.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/net/proxying_url_loader_factory.cc
+++ b/atom/browser/net/proxying_url_loader_factory.cc
@@ -4,6 +4,10 @@
 
 #include "atom/browser/net/proxying_url_loader_factory.h"
 
+#include <utility>
+
+#include "atom/browser/net/asar/asar_url_loader.h"
+
 namespace atom {
 
 ProxyingURLLoaderFactory::ProxyingURLLoaderFactory(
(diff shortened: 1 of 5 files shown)


atom/browser/net/proxying_url_loader_factory.cc
```diff
@@ -0,0 +1,62 @@
...
+    network::mojom::URLLoaderRequest loader,
+    int32_t routing_id,
+    int32_t request_id,
+    uint32_t options,
+    const network::ResourceRequest& request,
+    network::mojom::URLLoaderClientPtr client,
+    const net::MutableNetworkTrafficAnnotationTag& traffic_annotation) {
+  // Intercept file:// protocol to support asar archives.
+  if (request.url.SchemeIsFile()) {
+    asar::CreateAsarURLLoader(request, std::move(loader), std::move(client),
```
deepak1556: When managing `URLLoaders` the factory seems to be constructed on the IO thread, is this a pattern we should follow ? check `AwProxyingURLLoaderFactory`, `extensions::WebRequestProxyingURLLoaderFactory`

zcbenz: According to the comments of [`WebRequestProxyingURLLoaderFactory`](https://cs.chromium.org/chromium/src/extensions/browser/api/web_request/web_request_proxying_url_loader_factory.h?type=cs&g=0&l=38-44):

```c++
// This is loosely controlled by the WebRequestAPI on the UI
// thread, but does all its real work on the IO thread. This is only because
// it is tightly coupled to ExtensionsWebRequestEventRouter, and that object
// must stay on the IO thread until we can deprecate the non-Network Service
// path. Once Network Service is the only path, we can move all this stuff to
// the UI thread.
```

It is created on IO thread because it shares code from non-NetworkService path which must live on IO thread, and they do plan to move everything back to UI thread once non-NetworkService path is removed.

PR #18369, opened 2019-05-20
refactor: set defaults after WebContentsPreferences::Merge()

miniak:
#### Description of Change
Eliminates duplicate default values for `webPreferences`. Related to #18273.

The default values are set in `WebContentsPreferences::SetDefaults()` called from the constructor and the `Merge`. Therefore it should not be necessary to specify the default again value when calling `IsEnabled`. Also in this case we call `Clear` before `Merge`
https://github.com/electron/electron/blob/master/atom/browser/api/atom_api_browser_window.cc#L61. In the current version of the code, we loose the defaults after this.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/shell/browser/web_contents_preferences.cc
+++ b/shell/browser/web_contents_preferences.cc
@@ -115,35 +115,7 @@ WebContentsPreferences::WebContentsPreferences(
 
   instances_.push_back(this);
 
-  // Set WebPreferences defaults onto the JS object
-  SetDefaultBoolIfUndefined(options::kPlugins, false);
-  SetDefaultBoolIfUndefined(options::kExperimentalFeatures, false);
-  SetDefaultBoolIfUndefined(options::kNodeIntegration, false);
-  SetDefaultBoolIfUndefined(options::kNodeIntegrationInSubFrames, false);
-  SetDefaultBoolIfUndefined(options::kNodeIntegrationInWorker, false);
-  SetDefaultBoolIfUndefined(options::kDisableHtmlFullscreenWindowResize, false);
-  SetDefaultBoolIfUndefined(options::kWebviewTag, false);
-  SetDefaultBoolIfUndefined(options::kSandbox, false);
-  SetDefaultBoolIfUndefined(options::kNativeWindowOpen, false);
-  SetDefaultBoolIfUndefined(options::kContextIsolation, false);
-  SetDefaultBoolIfUndefined(options::kJavaScript, true);
-  SetDefaultBoolIfUndefined(options::kImages, true);
-  SetDefaultBoolIfUndefined(options::kTextAreasAreResizable, true);
-  SetDefaultBoolIfUndefined(options::kWebGL, true);
-  bool webSecurity = true;
-  SetDefaultBoolIfUndefined(options::kWebSecurity, webSecurity);
-  // If webSecurity was explicity set to false, let's inherit that into
-  // insecureContent
-  if (web_preferences.Get(options::kWebSecurity, &webSecurity) &&
-      !webSecurity) {
-    SetDefaultBoolIfUndefined(options::kAllowRunningInsecureContent, true);
-  } else {
-    SetDefaultBoolIfUndefined(options::kAllowRunningInsecureContent, false);
-  }
-#if defined(OS_MACOSX)
-  SetDefaultBoolIfUndefined(options::kScrollBounce, false);
-#endif
-  SetDefaultBoolIfUndefined(options::kOffscreen, false);
+  SetDefaults();
 
   // If this is a <webview> tag, and the embedder is offscreen-rendered, then
   // this WebContents is also offscreen-rendered.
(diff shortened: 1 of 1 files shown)


atom/browser/web_contents_preferences.cc
```diff
@@ -197,6 +193,8 @@ bool WebContentsPreferences::IsEnabled(const base::StringPiece& name,
 void WebContentsPreferences::Merge(const base::DictionaryValue& extend) {
   if (preference_.is_dict())
     static_cast<base::DictionaryValue*>(&preference_)->MergeDictionary(&extend);
+
+  SetDefaults();
```
zcbenz: The constructor already has defaults set, so I thought calling `SetDefaults` after `Merge` will not have any effect as `MergeDictionary` won't be deleting properties?

miniak: In this case we call `Clear` before `Merge`
https://github.com/electron/electron/blob/master/atom/browser/api/atom_api_browser_window.cc#L61

miniak: also some property's default value can depend on other property's value, which can be modified by the merge operation.

zcbenz: Thanks for clarification.

atom/browser/web_contents_preferences.cc
```diff
@@ -405,30 +399,24 @@ void WebContentsPreferences::AppendCommandLineSwitches(
...
 void WebContentsPreferences::OverrideWebkitPrefs(
     content::WebPreferences* prefs) {
-  prefs->javascript_enabled =
-      IsEnabled(options::kJavaScript, true /* default_value */);
-  prefs->images_enabled = IsEnabled(options::kImages, true /* default_value */);
-  prefs->text_areas_are_resizable =
-      IsEnabled(options::kTextAreasAreResizable, true /* default_value */);
-  prefs->navigate_on_drag_drop =
-      IsEnabled(options::kNavigateOnDragDrop, false /* default_value */);
+  prefs->javascript_enabled = IsEnabled(options::kJavaScript);
```
deepak1556: Aren't the defaults now `false` ?

miniak: the default values are set in `WebContentsPreferences::SetDefaults()` called from the constructor and the merge method. therefore it should not be necessary to specify the default again value when calling `IsEnabled`

deepak1556: ah missed that, in which case the signature of `IsEnabled` can be made stricter, just `bool IsEnabled(const base::StringPiece& name) const;`

miniak: modified `IsEnabled()` as suggested

miniak:
> The mutability here concerns me and is difficult to reason about (e.g. the `Clear` issue that you mentioned in the PR description).
> 
> Perhaps a clearer approach would be something like:
> 
> ```c++
> bool WebContentsPreferences::IsEnabled(const base::StringPiece& name) const {
>   auto* current_value = preference_.FindKeyOfType(name, base::Value::Type::BOOLEAN);
>   if (current_value) {
>     return current_value->GetBool();
>   } else {
>     return IsEnabledByDefault(name);
>   }
> }
> 
> 
> bool WebContentsPreferences::IsEnabledByDefault(const base::StringPiece& name) const {
>   if (name == options::kEnableRemoteModule)
>     return true;
>   if (name == options::kJavaScript)
>     return true;
>   // ...
>   if (name == options::kAllowRunningInsecureContent)
>     return !IsEnabled(options::kWebSecurity);
>   return false;
> }
> ```

this would work for the native code reading the options, however it won't for JS code using `getLastWebPreferences()`

nornagon:
Can we dynamically create a read-only object that reflects the current state when JS calls `getLastWebPreferences`? e.g. with [`gin::DataObjectBuilder`](https://cs.chromium.org/chromium/src/gin/data_object_builder.h)

codebytere:
@miniak do you still plan to work on this? There are some conflicts and unresolved comments.

miniak:
@codebytere yes, I just didn't have time for it

miniak:
@nornagon the PR is now simpler after the #19220 was merged. Is it really such a big deal to mutate the preferences? We are doing that already, this just makes sure the defaults are applied properly after the Merge.

nornagon:
@miniak mutable preferences here creates an opportunity for bugs, as you discovered. Refactoring to a read-only system will eliminate that class of bugs going forward. I understand that this is different to what we currently do; indeed, I believe it would be an improvement!

codebytere:
@miniak do you plan to work on this?

zcbenz:
I'm closing this PR since there has been no activity for a long time, and the direction of this PR did not get an agreement.

nornagon:
NB #30193, which takes a different approach.

miniak:
Obsoleted by #30193

PR #18374, opened 2019-05-21
feat: migrate protocol module to NetworkService (Part 9)

zcbenz:
#### Description of Change

This PR implements the `protocol.interceptXXXProtocol` APIs with NetworkService, also fixes a few related tests.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/atom_browser_client.cc
+++ b/atom/browser/atom_browser_client.cc
@@ -971,13 +971,15 @@ bool AtomBrowserClient::WillCreateURLLoaderFactory(
     bool* bypass_redirect_checks) {
   content::WebContents* web_contents =
       content::WebContents::FromRenderFrameHost(frame_host);
-  if (!web_contents)
+  api::ProtocolNS* protocol = api::ProtocolNS::FromWrappedClass(
+      v8::Isolate::GetCurrent(), web_contents->GetBrowserContext());
+  if (!protocol)
     return false;
 
   auto proxied_request = std::move(*factory_request);
   network::mojom::URLLoaderFactoryPtrInfo target_factory_info;
   *factory_request = mojo::MakeRequest(&target_factory_info);
-  new ProxyingURLLoaderFactory(std::move(proxied_request),
+  new ProxyingURLLoaderFactory(protocol, std::move(proxied_request),
                                std::move(target_factory_info));
   return true;
 }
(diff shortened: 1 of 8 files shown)


atom/browser/net/proxying_url_loader_factory.cc
```diff
@@ -32,6 +34,16 @@ void ProxyingURLLoaderFactory::CreateLoaderAndStart(
     const network::ResourceRequest& request,
     network::mojom::URLLoaderClientPtr client,
     const net::MutableNetworkTrafficAnnotationTag& traffic_annotation) {
+  // Check if user has intercepted this scheme.
+  auto it = protocol_->intercept_handlers().find(request.url.scheme());
+  if (it != protocol_->intercept_handlers().end()) {
+    AtomURLLoaderFactory(it->second.first, it->second.second)
```
deepak1556: Any reason to create this factory on the stack ?

atom/browser/atom_browser_client.cc
```diff
@@ -971,13 +971,15 @@ bool AtomBrowserClient::WillCreateURLLoaderFactory(
...
+  api::ProtocolNS* protocol = api::ProtocolNS::FromWrappedClass(
+      v8::Isolate::GetCurrent(), web_contents->GetBrowserContext());
+  if (!protocol)
     return false;
 
   auto proxied_request = std::move(*factory_request);
   network::mojom::URLLoaderFactoryPtrInfo target_factory_info;
   *factory_request = mojo::MakeRequest(&target_factory_info);
-  new ProxyingURLLoaderFactory(std::move(proxied_request),
+  new ProxyingURLLoaderFactory(protocol, std::move(proxied_request),
```
deepak1556: Why not just pass the `intercept_handlers_` ? would avoid accessing api class out of its boundary.

atom/browser/net/proxying_url_loader_factory.cc
zcbenz: To call `CreateLoaderAndStart` we have to create the factory first, and to create the factory creating on stack is the simplest and cheapest way, calling `make_unique` would allocate memory on heap which is unnecessarily expensive.

atom/browser/atom_browser_client.cc
zcbenz: This sounds a good idea, I have refactored the code to accept the handlers instead of protocol class.

atom/browser/net/proxying_url_loader_factory.cc
deepak1556: In that case, we can maintain one `AtomURLLoaderFactory` per network context and make new connections using `AtomURLLoaderFactory::Clone` which will essentially reference count the class and delete it when all the connection drops. `AppCacheSubresourceURLFactory` is a good example.

zcbenz: Didn't know that trick, will do some learning on it.

On this code specially, I realize we don't even need to create `AtomURLLoaderFactory` at all, I have changed the code to call the helper method directly to avoid any extra overhead.

Issue #16019, opened 2018-12-11
An app crashes when `contentTracing.stopRecording` is called with an empty string as its first argument


ZacWalk:
* Electron 3 - all versions
* Windows 10
* Works in Electron 2 - all versions

**Expected Behavior**
ContentTracing example code from the Electron documentation should not crash.

**Actual behavior**
ContentTracing api crashes after startRecording is called. It looks like some issue with the callback.

```javascript
const { app, contentTracing } = require('electron')

app.on('ready', () => {
  const options = {
    categoryFilter: '*',
    traceOptions: 'record-until-full,enable-sampling'
  }

  contentTracing.startRecording(options, () => {
    console.log('Tracing started')

    setTimeout(() => {
      // *** Following line crashes ***
      contentTracing.stopRecording('', (path) => {
        console.log('Tracing data recorded to ' + path)
      })
    }, 5000)
  })
})
```

ZacWalk:
This is the stack trace from the crash I see:

- electron.exe![thunk]:crash_reporter::CrashReporter::`vcall'{8,{flat}}' }'()	C++
- [Inline Frame] electron.exe!base::RepeatingCallback<bool __cdecl(base::trace_event::TraceConfig const &,base::RepeatingCallback<void __cdecl(void)> const &)>::Run(const base::trace_event::TraceConfig &) Line 124	C++
- [Inline Frame] electron.exe!mate::internal::Invoker<mate::internal::IndicesHolder<0,1>,base::trace_event::TraceConfig const &,base::RepeatingCallback<void __cdecl(void)> const &>::DispatchToCallback(base::RepeatingCallback<bool __cdecl(base::trace_event::TraceConfig const &,base::RepeatingCallback<void __cdecl(void)> const &)> callback) Line 196	C++
- electron.exe!mate::internal::Dispatcher<bool __cdecl(base::trace_event::TraceConfig const & __ptr64,base::RepeatingCallback<void __cdecl(void)> const & __ptr64)>::DispatchToCallback(const v8::FunctionCallbackInfo<v8::Value> & info) Line 239	C++
- node.dll!v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo * handler) Line 30	C++
[...]

PR #18391, opened 2019-05-21
test: add spec for contentTracing.stopRecording with empty path

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/16019.

The crash raised in the issue has been verified to be fixed in `master` locally, so this PR adds a test to prevent future regressions with the `''` argument to `contentTracing.stopRecording`.

cc @miniak @alexeykuzmin @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/spec/api-content-tracing-spec.js
+++ b/spec/api-content-tracing-spec.js
@@ -15,10 +15,6 @@ const timeout = async (milliseconds) => {
   })
 }
 
-const getPathInATempFolder = (filename) => {
-  return path.join(app.getPath('temp'), filename)
-}
-
 describe('contentTracing', () => {
   beforeEach(function () {
     // FIXME: The tests are skipped on arm/arm64.
@@ -38,7 +34,7 @@ describe('contentTracing', () => {
     return resultFilePath
   }
 
-  const outputFilePath = getPathInATempFolder('trace.json')
+  const outputFilePath = path.join(app.getPath('temp'), 'trace.json')
   beforeEach(() => {
     if (fs.existsSync(outputFilePath)) {
       fs.unlinkSync(outputFilePath)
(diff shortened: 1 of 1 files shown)


spec/api-content-tracing-spec.js
```diff
@@ -115,6 +111,21 @@ describe('contentTracing', () => {
...
+      const options = {
+        categoryFilter: '*',
+        traceOptions: 'record-until-full,enable-sampling'
+      }
+
+      expect(() => {
+        contentTracing.startRecording(options).then(() => {
+          setTimeout(() => {
+            contentTracing.stopRecording('').then(path => console.log(path))
+          }, 5000)
```
nornagon: is there any particular need for this test to take 5 seconds? Can we just immediately call `stopRecording` with no timeout?

Can we also check that `path` exists & is a file?

deepak1556: my understanding was that it is to emulate some tracing delay, there should be no reason to not do it immediately if needed.

codebytere: done!

spec/api-content-tracing-spec.js
```diff
@@ -115,6 +111,20 @@ describe('contentTracing', () => {
...
 
+    it('does not crash on empty string', (done) => {
+      const options = {
+        categoryFilter: '*',
+        traceOptions: 'record-until-full,enable-sampling'
+      }
+
+      contentTracing.startRecording(options).then(() => {
+        contentTracing.stopRecording('').then(path => {
+          expect(path).to.be.a('string').that.is.not.empty()
```
nornagon: please check with `fs.statSync` that this exists on the filesystem & is a file.
```suggestion
          expect(path).to.be.a('string').that.is.not.empty()
          expect(fs.statSync(path).isFile()).to.be.true()
```

spec/api-content-tracing-spec.js
```diff
@@ -115,6 +111,21 @@ describe('contentTracing', () => {
...
+          expect(fs.statSync(path).isFile()).to.be.true()
+          done()
+        })
+      })
```
nornagon: minor nit, but you could take advantage of async/await here:

```js
it('does not crash on empty string', async () => {
  const options = {
    categoryFilter: '*',
    traceOptions: 'record-until-full,enable-sampling'
  }
  await contentTracing.startRecording(options)
  const path = await contentTracing.stopRecording('')
  expect(path).to.be.a('string').that.is.not.empty()
  expect(fs.statSync(path).isFile()).to.be.true()
})
```

nornagon:
. o O ( is this already covered by https://github.com/electron/electron/blob/master/spec/api-content-tracing-spec.js#L131 ? )

PR #17668, opened 2019-04-03
build: auto-generate the codesigning cert used for macOS CI testing runs

MarshallOfSound:
Fixes #17649

Instead of using a pre-generated cert we can dynamically make a new random one every time.  Thanks to @nornagon for the optimism that this would be possible 😀 

Notes: no-notes

PR #18395, opened 2019-05-21
fix: use print() function in both Python 2 and Python 3

cclauss:
#### Description of Change

notes: Fix issues that will halt the runtime on modern Python.

Legacy __print__ statements are ___syntax errors___ in Python 3 but __print()__ function works as expected in both Python 2 and Python 3. 

Old style exceptions are ___syntax errors___ in Python 3 but new style exceptions work as expected in both Python 2 and Python 3.

* __basestring__ was removed in Python 3 in favor of __str__.
* __file()__ was removed in Python 3 in favor of __open()__ which works in both Python 2 and Python 3.

225 days until Python 2 end of life.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:


nornagon:
I'm all for switching to Py3 but we very heavily depend on Py2 currently because Chromium's build scripts are all Py2. Are you aware of any efforts in Chromium to upgrade to 3?

script/pump.py
```diff
@@ -61,6 +61,7 @@
            | EMPTY
```
nornagon: I think we don't actually use this script any more. cc @zcbenz who I think added it originally.

tools/linux/rewrite_dirs.py
```diff
@@ -4,6 +4,7 @@
 # found in the LICENSE file.
```
nornagon: I don't think we use this any more either.

cclauss:
@nornagon None of these changes break Python 2 compatibility.  There has been a lot of movement towards Python 3 compatibility in the Chromium codebase in recent months.  Please check it if you get a chance.

cclauss:
Am I correct in assuming that the __build-mac__ failure is unrelated?

MarshallOfSound:
@cclauss Yup, that failure is due to the PR coming in from a fork.  That test always fails for forks at the moment --> #17668

script/pump.py
codebytere: we don't: @cclauss would you please remove it instead?

tools/linux/rewrite_dirs.py
codebytere: this gets used by `pkg-config-wrapper`, which i believe we still use to wrap around `pkg-config` to generate the correct include and  library paths when cross-compiling using a sysroot

nornagon: I couldn't find any reference to this script, can you link me to where it's used?

codebytere: cc @deepak1556 i think you may know more?

deepak1556: This is a pre GN thing, not needed anymore.

codebytere:
@cclauss can you resolve the merge conflicts? once that's done we can get this merged!

Issue #11189, opened 2017-11-20
Notification with tag works only once


manix:
After creating a notification and passing a tag, then subsequent notifications with the same tag don't show up for the entire lifespan of the app.

* Electron version: 1.8.1
* Operating system: Windows 7 x64

### Expected behavior

New notifications should open after the last one has closed.

### Actual behavior

Nothing happens

### How to reproduce

  $ git clone https://github.com/electron/electron-quick-start .
  $ npm install
  $ npm start

Then in the app open dev tools and paste this in the console:

    new Notification("test", { tag: "test" });

Wait for the notification to disappear and paste it again, observe nothingness.

MarshallOfSound:
For clarity this is not a 1.8.x regression, it only occurs on Windows 7 and it's due to how we attempt to de-duplicate notification objects based on there tag in our windows 7 notification presenter

gasi-signal:
We ran into this as well for the Signal Desktop app.

**Repro**


![windows-7-notifications-static-tag](https://user-images.githubusercontent.com/36971170/39647642-c3a7f7e2-4fad-11e8-803c-9ca1e7010585.gif)

**Code**

```javascript
let counter = 0;
let intervalId = setInterval(
  () => new Notification(counter++, { tag: "static-tag" }),
  1000
);
```

sofianguy:
Thank you for taking the time to report this issue and helping to make Electron better.

The version of Electron you reported this on has been superseded by newer releases. 

If you're still experiencing this issue in Electron v2.0.x or later, please add a comment specifying the version you're testing with and any other new information that a maintainer trying to reproduce the issue should know.

I'm setting the `blocked/need-info` label for the above reasons.

Thanks in advance! Your help is appreciated.

sofianguy:
Thank you for your issue!

We haven't gotten a response to our questions in our comment above. With only the information that is currently in the issue, we don't have enough information to take action. I'm going to close this but don't hesitate to reach out if you have or find the answers we need, we'll be happy to reopen the issue.

r0stig:
Hello,

I just want to point out that I still get this bug on electron 4.1.4.

I'm able to reproduce the bug with the the code @gasi-signal posted and a clean electron-quick-start-repo.

r0stig:
Just letting you know the bug is still present in 5.0.0.

PR #18401, opened 2019-05-22
fix: remove lingering data when notifications autodismiss on Windows 7

r0stig:
#### Description of Change

I Windows 7 notifications that gets auto dismissed (notifications that closes after 4 seconds due to no interaction) won't remove the reference to the notification object which means following notifications that uses the same tag won't be shown.

To solve this issue we need to remove the old reference to the notification within the presenter when a new notification with the same tag is created. Also when a notification is auto dismissed the calls for a manual dismiss is called to remove the notification objets.

Fixes: https://github.com/electron/electron/issues/11189

#### Checklist

#### Release Notes

Notes: Fixed bug where notifications with same tag wouldn't show more than once when first notification was auto dismissed in Windows 7.

### DIFF
--- a/atom/browser/notifications/win/win32_desktop_notifications/toast.cc
+++ b/atom/browser/notifications/win/win32_desktop_notifications/toast.cc
@@ -270,10 +270,16 @@ LRESULT DesktopNotificationController::Toast::WndProc(HWND hwnd,
     case WM_MOUSEACTIVATE:
       return MA_NOACTIVATE;
 
-    case WM_TIMER:
+    case WM_TIMER: {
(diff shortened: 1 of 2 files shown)


zcbenz:
@alespergl Can you take a look at this?

r0stig:
Is there anyone else who are able to confirm this pull request if @alespergl isn't present?

zcbenz:
This should be good to go, I don't see any harm on this.

r0stig:
> This should be good to go, I don't see any harm on this.

Thanks, is there a way to merge this to current versions (5.0.x, 6.0.x) or how does this work?

PR #18408, opened 2019-05-22
refactor: remove textured BrowserWindow type option

codebytere:
#### Description of Change

Both [`NSTexturedWindowMask`](https://developer.apple.com/documentation/appkit/nstexturedbackgroundwindowmask?language=objc) and [`NSWindowStyleMaskTexturedBackground`](https://developer.apple.com/documentation/appkit/nswindowstylemask/nswindowstylemasktexturedbackground?language=objc) have been deprecated with no replacements in macOS. This therefore removes that option from the `type` options for new BrowserWindows on macOS.

Since there is no longer a textured window option, there will no longer be a case where `standardWindow` will be false and this we can remove it as well.

A paired deprecation PR will be opened against `6-0-x` once this has been merged.

cc @nornagon @deepak1556 @miniak 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Removed `textured` from the `type` options for new BrowserWindows on macOS.

### DIFF
--- a/atom/browser/native_window_mac.mm
+++ b/atom/browser/native_window_mac.mm
@@ -311,17 +311,9 @@ NativeWindowMac::NativeWindowMac(const mate::Dictionary& options,
   std::string windowType;
   options.Get(options::kType, &windowType);
 
-  bool useStandardWindow = true;
-  // eventually deprecate separate "standardWindow" option in favor of
-  // standard / textured window types
-  options.Get(options::kStandardWindow, &useStandardWindow);
-  if (windowType == "textured") {
-    useStandardWindow = false;
-  }
-
   NSUInteger styleMask = NSWindowStyleMaskTitled;
   if (title_bar_style_ == TitleBarStyle::CUSTOM_BUTTONS_ON_HOVER &&
-      (!useStandardWindow || transparent() || !has_frame())) {
+      (transparent() || !has_frame())) {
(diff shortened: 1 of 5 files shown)


atom/browser/native_window_mac.mm
```diff
@@ -334,9 +326,6 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
     // The window without titlebar is treated the same with frameless window.
     set_has_frame(false);
   }
-  if (!useStandardWindow || transparent() || !has_frame()) {
-    styleMask |= NSTexturedBackgroundWindowMask;
```
MarshallOfSound: This concerns me, deprecated or not we should keep this behavior until the API is completely removed.  Frameless / transparent windows depend on this mask under some conditions iirc from the Mojave White Line incident

codebytere: Do you know which conditions specifically? I'd prefer to try to work around those then keep this if that's a viable pathway i'd say

MarshallOfSound: @codebytere I think High Sierra and below 🤔 I can't recall / guarantee to know all the cases though.  Until it's removed from the SDK I'd prefer we kept it

codebytere: i can close this PR then :)

PR #18411, opened 2019-05-22
feat: [contentTracing] allow calling stopTracing() with no arguments

nornagon:
#### Description of Change
This refactors the contentTracing code somewhat, and adds the ability to call `contentTracing.stopTracing()` with no arguments (with the same behaviour as when calling with an empty string).

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Calling contentTracing.stopTracing() with no arguments is now permitted and has the same behaviour as calling it with an empty string (tracing data will be written to a temporary file).

### DIFF
--- a/atom/browser/api/atom_api_content_tracing.cc
+++ b/atom/browser/api/atom_api_content_tracing.cc
@@ -4,6 +4,7 @@
 
 #include <set>
 #include <string>
+#include <utility>
 
 #include "atom/common/native_mate_converters/callback.h"
 #include "atom/common/native_mate_converters/file_path_converter.h"
@@ -12,6 +13,7 @@
 #include "atom/common/promise_util.h"
 #include "base/bind.h"
 #include "base/files/file_util.h"
+#include "base/optional.h"
 #include "base/threading/thread_restrictions.h"
 #include "content/public/browser/tracing_controller.h"
 #include "native_mate/dictionary.h"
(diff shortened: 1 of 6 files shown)


atom/browser/api/atom_api_content_tracing.cc
```diff
@@ -104,10 +116,15 @@ v8::Local<v8::Promise> StartTracing(
...
-  // Note: This method always succeeds.
-  TracingController::GetInstance()->StartTracing(
-      trace_config, base::BindOnce(atom::util::Promise::ResolveEmptyPromise,
-                                   std::move(promise)));
+  if (!TracingController::GetInstance()->StartTracing(
+          trace_config, base::BindOnce(atom::util::Promise::ResolveEmptyPromise,
+                                       std::move(promise)))) {
+    // If StartTracing returns false, that means it didn't invoke its callback.
+    // Return an already-resolved promise and abandon the previous promise (it
+    // was std::move()d into the StartTracing callback and has been deleted by
```
codebytere: Given the above, might it perhaps just be better to use a RepeatingCallback here?

nornagon: Not really; if I were going to use a repeating callback that would just force me to use `CopyablePromise`. But I could just as well use `CopyablePromise` with a `OnceCallback` to avoid the `std::move`. However, I think this is a better design, because it avoids unnecessary copying in the usual case—mostly, `startTracing` will only be called once, and this code won't be executed.

atom/browser/api/atom_api_content_tracing.cc
```diff
@@ -53,36 +55,46 @@ struct Converter<base::trace_event::TraceConfig> {
...
-  // for now since offloading this to a different sequence would require
-  // changing the api shape
-  base::ThreadRestrictions::ScopedAllowIO allow_io;
-  if (result_file_path.empty() && !base::CreateTemporaryFile(&result_file_path))
-    LOG(ERROR) << "Creating temporary file failed";
+base::Optional<base::FilePath> CreateTemporaryFileOnIO() {
+  base::FilePath temp_file_path;
+  if (base::CreateTemporaryFile(&temp_file_path)) {
+    return base::make_optional(std::move(temp_file_path));
+  } else {
```
codebytere: unnecessary `else` here i would say

nornagon:
fails are flakes, merging.

PR #18413, opened 2019-05-23
chore: update the npm package to use @electron/get

MarshallOfSound:
This updates our npm package to use `@electron/get` the new and shiny version of `electron-download`.  Tested locally and appears to work 👍 

Notes: Updated the `electron` module to use `@electron/get`.  The minimum supported node version is now Node 8.

### DIFF
--- a/npm/install.js
+++ b/npm/install.js
@@ -1,14 +1,14 @@
 #!/usr/bin/env node
 
-var version = require('./package').version
+const version = require('./package').version
 
-var fs = require('fs')
-var os = require('os')
-var path = require('path')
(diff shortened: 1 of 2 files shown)


malept:
> The minimum supported node version is now Node 6.

Pretty sure it's Node 8, due to `got` :smile:

MarshallOfSound:
@malept Updated 👍 Node 6 is EOL as well

malept:
Do we want to backport to v6 or is it too late?

MarshallOfSound:
@malept Too late, and no pressing need IMO

PR #18432, opened 2019-05-23
test: move beforeunload tests to main runner and fix flake

nornagon:
#### Description of Change
This fixes a flaky test where a race in window closing could cause the `'closed'` event to never be emitted when cleaning up in the `afterEach` of the webContents spec.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/spec-main/window-helpers.ts
+++ b/spec-main/window-helpers.ts
@@ -2,26 +2,37 @@ import { expect } from 'chai'
 import { BrowserWindow, WebContents } from 'electron'
 import { emittedOnce } from './events-helpers';
 
+async function ensureWindowIsClosed(window) {
+  if (window && !window.isDestroyed()) {
+    if (window.webContents && !window.webContents.isDestroyed()) {
+      // If a window isn't destroyed already, and it has non-destroyed WebContents,
+      // then calling destroy() won't immediately destroy it, as it may have
+      // <webview> children which need to be destroyed first. In that case, we
+      // await the 'closed' event which signals the complete shutdown of the
+      // window.
+      const isClosed = emittedOnce(window, 'closed')
+      window.destroy()
+      await isClosed()
+    } else {
+      // If there's no WebContents or if the WebContents is already destroyed,
+      // then the 'closed' event has already been emitted so there's nothing to
+      // wait for.
+      window.destroy()
+    }
+  }
+}
+
 export const closeWindow = async (
   window: BrowserWindow | null = null,
   { assertNotWindows } = { assertNotWindows: true }
 ) => {
-  if (window && !window.isDestroyed()) {
-    window.destroy()
-  }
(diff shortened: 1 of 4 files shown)


spec-main/window-helpers.ts
```diff
@@ -7,18 +7,13 @@ export const closeWindow = async (
   { assertNotWindows } = { assertNotWindows: true }
 ) => {
   if (window && !window.isDestroyed()) {
-    const isClosed = emittedOnce(window, 'closed')
-    window.setClosable(true)
-    window.close()
-    await isClosed
+    window.destroy()
```
deepak1556: `window.destroy()` will not be a synchronous signal if the window embeds a `<webivew>`. It is still better to wait for close signal before returning function control.

nornagon: The thing that's causing flake here is that in some cases, especially when the test itself closes the window, waiting for the 'closed' event doesn't work because the window is already closed. the `isDestroyed` check doesn't seem to be a good enough signal to determine whether or not a `closed` event will be emitted.

deepak1556: In that case the tests that are closing windows can use `.destroy()` api , so that `isDestroyed` will become a reliable signal ?

nornagon: I'm worried about future flakes. This behaviour is far from obvious to test writers / reviewers. The fix needs to be resilient to future flakes.

spec-main/window-helpers.ts
```diff
@@ -2,31 +2,37 @@ import { expect } from 'chai'
...
+async function ensureWindowIsClosed(window) {
+  if (window && !window.isDestroyed()) {
+    if (window.webContents && !window.webContents.isDestroyed()) {
+      // If a window isn't destroyed already, and it has non-destroyed WebContents,
+      // then calling destroy() won't immediately destroy it, as it may have
+      // <webview> children which need to be destroyed first. In that case, we
+      // await the 'closed' event which signals the complete shutdown of the
+      // window.
+      const isClosed = emittedOnce(window, 'closed')
+      window.destroy()
```
deepak1556:
```
window.setClosable(true)
window.close()
```

there is nothing wrong with calling `destroy` here but the api users adopt is `close` for `BrowserWindow`, it would be good to test that.

nornagon: This is not a test, this is cleanup. If we want to test `close()`, we should be explicit and write a test for it, I think.

deepak1556: I was thinking, its not possible to test all different scenarios of closing a window, but adding it to a helper class will atleast cover a bunch of cases. Originally `destroy` was exposed to cleanup c++ classes from js land. But considering your point, yeah its better to explicitly test `close`.

PR #18448, opened 2019-05-24
build: add env var to allow easily triggering a 3way patch

MarshallOfSound:
I find myself manually adding `threeway=True` in this script every now and again to deal with patch rebases / conflicts.  This would allow us to do `ELECTRON_USE_THREE_WAY_MERGE_FOR_PATCHES=true e sync` to deal with these conflicts.

Notes: no-notes

### DIFF
--- a/script/apply_all_patches.py
+++ b/script/apply_all_patches.py
@@ -2,6 +2,7 @@
 
 import argparse
 import json
+import os
 import sys
 
 from lib import git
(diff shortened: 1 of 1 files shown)


nornagon:
Can you say a little more about the workflow you're using this in?

In general I find that I can never rely on 3-way merge to be successful, so I always want it to be a manual process. Usually that looks something like:

```
$ gclient sync
...
[failed to apply patches]
$ cd path/to/failed/repo
$ git am --abort
$ ../../electron/script/git-import-patches -3 ../../electron/patches/common/repo
[... fix merge conflicts ...]
```

MarshallOfSound:
> In general I find that I can never rely on 3-way merge to be successful

In my experience its worked like 99% of the time.  This isn't when performing upgrades, this is when an upgrade lands and your changes to patch files conflict  i.e.

```bash
e sync # Fails due to patch conflict
git am --abort # Clean up failed patch
e sync --with-three-way-merge # Works
./electron/script/git-export-patches -o ./electron/patches/common/chromium # Export resulting things
```

zcbenz:
@nornagon Are you fine with merging this?

PR #18449, opened 2019-05-24
feat: add ipcRenderer.invoke()

nornagon:
#### Description of Change
This adds a new IPC primitive, `ipcRenderer.invoke()`. This models a request/response paradigm, similar to `ipcRenderer.sendSync` but asynchronous.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `ipcRenderer.invoke()` and `ipcMain.handle()` for asynchronous request/response-style IPC.


docs/api/ipc-renderer.md
```diff
@@ -57,11 +57,41 @@ Removes all listeners, or those of the specified `channel`.
...
+* `...args` any[]
+
+Returns `Promise<any>` - Resolves with the response from the main process.
+
+Send a message to the main process asynchronously via `channel` and expect an
+asynchronous result. Arguments will be serialized as JSON internally and
+hence no functions or prototype chain will be included.
+
+The main process should listen for `channel` with the [`ipcMain`](ipc-main.md)
+module, and send a result by calling `event.reply()`.
```
MarshallOfSound: This leaves a few unanswered questions:
* What happens when `event.reply()` is called multiple times?
* What happens if multiple listeners for the `channel` are added in the main process

It isn't quite as nice an API as `ipcMainInternal.handle(channel)` which automatically proxies errors thrown in the handler.  I'm imaginging:

```js
// Will throw / warn if a handler for "channel" is already set
ipcMain.setHandler('channel', async () => {
  return 'myvalue';
});
```

atom/renderer/api/atom_api_renderer_ipc.cc
```diff
@@ -62,6 +66,20 @@ class IPCRenderer : public mate::Wrappable<IPCRenderer> {
     electron_browser_ptr_->Message(internal, channel, arguments.Clone());
   }
 
+  v8::Local<v8::Promise> Invoke(mate::Arguments* args,
+                                const std::string& channel,
+                                const base::Value& arguments) {
+    atom::util::Promise p(args->isolate());
+    auto handle = p.GetHandle();
+    electron_browser_ptr_->Invoke(
```
MarshallOfSound: Although probs not as clean as this, IMO we should proxy thrown errors over this as well and reject the promise

docs/api/ipc-renderer.md
nornagon: the 2nd and subsequent `event.reply()` calls will be no-ops (the callback is stored as an `Optional<OnceCallback>` and it gets `std::move`d to run, with a check to make sure it's present).

Maybe they should error instead?

Multiple listeners = all of them get called, whichever one calls `event.reply` first wins. I don't think it's especially useful to have multiple listeners, but I also don't see a particular reason to go through any special pains to disallow it.

atom/renderer/api/atom_api_renderer_ipc.cc
nornagon: Hmmmm maybe, but it's totally possible (likely even) that errors won't happen synchronously in the handler. e.g.

```
ipcMain.on('read-some-file', (e) => {
  fs.readFile('some-file', (err, data) => e.reply(data))
})
```

there's no way we can automatically catch that kind of error. But i guess there should be a way for the caller to send a rejection? and we can at least catch synchronous errors if they happen.

MarshallOfSound: @nornagon That's why I was suggesting a `setHandler` method that provided a channel and an async function.  Then we should be able to catch async errors as well as it's just a promise.

```
ipcMain.setHandler = (channel, fn) => {
  if (ipcMain.handlers(channel).length !== 0) throw err;
  ipcMain.on(channel, async (...args) => {
    try {
      event.reply(await Promise.resolve(fn(...args)))
    } catch (err) {
      event.sendError(err);
    }
  });
}

ipcMain.removeHandler = (channel, fn) => {
  // ...
}
```

docs/api/ipc-renderer.md
MarshallOfSound: > Multiple listeners = all of them get called, whichever one calls event.reply first wins

This sounds like a foot-gun that we should prevent users experiencing

atom/renderer/api/atom_api_renderer_ipc.cc
nornagon: I mean, assuming that the caller uses promises :)

MarshallOfSound: @nornagon That's why I wrapped it in `Promise.resolve(fn(...args))`, whether it returns promises or not that will work 👍

nornagon: You have to use promises all the way down for exceptions to be correctly propagated, though. All I'm saying is that the handler can never be 100% guaranteed to call `reply` or `sendError` because it's under the control of the user.

MarshallOfSound: I mean if the user isn't handling errors / leaving dangling / un-awaited promises that's on them.  We can't catch everything, but we can make it really easy to catch everything if you write a proper async handler

nornagon: If we build a `setHandler` API like you propose, do you think it would be worth separating the channel surface so it wasn't just handled with `ipcMain.on` but rather through a separate event bus?

MarshallOfSound: > but rather through a separate event bus?

I think so yes

nornagon: On further thought, this approach means that it's not possible to write a handler in a promise-less asynchronous style, like

```js
ipcMain.handle('thing', (e, arg) => {
  doAsyncThing(arg, (result) => e.reply(result))
})
```

because the return value of the handler, if not a promise, will be immediately sent as a response. You'd have to write

```js
ipcMain.handle('thing', (e, arg) => {
    doAsyncThing(arg, resolve)
  })
})
```

or

```js
ipcMain.handle('thing', async (e, arg) => {
  await new Promise(resolve => doAsyncThing(arg, resolve))
})
```

or some similar incantation.

I like promises and think we should support them, but I'm a little worried about adding a core API that doesn't support not-Promises.

MarshallOfSound: @nornagon I think it's ok to be promise-only, that's what all our other APIs are moving towards (removing callbacks, using promises).

If folks really want to use promises they can do `new Promise((resolve)=>{})` as you outlined above or they can just promisify their handler.  E.g.

```js
const { promisify } = require('util');

ipcMain.handle('thing', promisify((cb) => {
  doAsyncWork(arg, cb);
});
```

nornagon: Ok, I did it :) PTAL

docs/api/ipc-main.md
```diff
@@ -88,6 +88,54 @@ Removes the specified `listener` from the listener array for the specified
 
 Removes listeners of the specified `channel`.
 
+### `ipcMain.handle(channel, listener)`
+
+* `channel` String
+* `listener` Function
```
MarshallOfSound:
```suggestion
* `listener` Function<Promise>
```

We can declare the return type that this listener needs to have

docs/api/ipc-main.md
```diff
@@ -88,6 +88,54 @@ Removes the specified `listener` from the listener array for the specified
...
+  * `event` IpcMainEvent
```
MarshallOfSound:
```suggestion
  * `event` IpcMainHandleableEvent
```

We should keep `reply` on the `IpcMainEvent` interface, there are use cases for that method (multiple responses, responding with a different channel, etc.)

docs/api/ipc-main.md
```diff
@@ -88,6 +88,54 @@ Removes the specified `listener` from the listener array for the specified
...
+```
+
+The `event` that is passed as the first argument to the handler is the same as
+that passed to a regular event listener. It includes information about which
+WebContents is the source of the invoke request.
+
+### `ipcMain.handleOnce(channel, listener)`
+
+* `channel` String
+* `listener` Function
```
MarshallOfSound: Same as above
```suggestion
* `listener` Function<Promise>
```

docs/api/ipc-main.md
```diff
@@ -88,6 +88,54 @@ Removes the specified `listener` from the listener array for the specified
...
+  * `event` IpcMainEvent
```
MarshallOfSound: Same as above
```suggestion
  * `event` IpcMainHandleableEvent
```

docs/api/ipc-main.md
```diff
@@ -88,6 +88,54 @@ Removes the specified `listener` from the listener array for the specified
...
+  * `...args` any[]
+
+Handles a single `invoke`able IPC message, then removes the listener. See
+`ipcMain.handle(channel, listener)`.
+
+### `ipcMain.removeHandler(channel)`
```
MarshallOfSound: Maybe an `isHandled` API would be nice? 🤷‍♂

docs/api/ipc-renderer.md
```diff
@@ -57,11 +57,41 @@ Removes all listeners, or those of the specified `channel`.
...
+
+For example:
+```javascript
+// Renderer
+ipcRenderer.invoke('some-name', someArgument).then((result) => {
+  // ...
+})
+
+// Main
+ipcMain.on('some-name', (event, someArgument) => {
```
MarshallOfSound: This example needs to be updated

lib/browser/api/ipc-main.ts
```diff
@@ -1,8 +1,44 @@
 import { EventEmitter } from 'events'
+import { IpcMainEvent } from 'electron'
 
-const emitter = new EventEmitter()
+class IpcMain extends EventEmitter {
+  _invokeHandlers: {[channel: string]: (e: IpcMainEvent, ...args: any[]) => void};
```
MarshallOfSound:
```suggestion
  private _invokeHandlers: Record<string, (e: IpcMainEvent, ...args: any[]) => Promise<any>> = {};
```

lib/browser/api/ipc-main.ts
```diff
@@ -1,8 +1,44 @@
...
+  constructor () {
```
MarshallOfSound: This constructor is not required, we can initialize in the property declaration above

lib/browser/api/ipc-main.ts
```diff
@@ -1,8 +1,44 @@
...
+    super()
+    this._invokeHandlers = {}
+  }
+
+  handle (method: string, fn: (e: IpcMainEvent, ...args: any[]) => any) {
+    if (method in this._invokeHandlers) {
```
MarshallOfSound: This will prevent people using methods that are also the name of things on the `Object` prototype.  `_invokeHelpers` should probably become a `Map` then you can do `_invokeHelpers.has(method)`

lib/browser/api/web-contents.js
```diff
@@ -326,6 +326,17 @@ WebContents.prototype._init = function () {
     }
   })
 
+  this.on('-ipc-invoke', function (event, channel, args) {
+    event.reply = (result) => event.sendReply({ result })
+    event.throw = (error) => event.sendReply({ error: error.toString() })
+    this.emit('ipc-invoke', event, channel, ...args)
```
MarshallOfSound: What is the purpose of this `emit`?

lib/browser/api/web-contents.js
```diff
@@ -326,6 +326,17 @@ WebContents.prototype._init = function () {
...
+    if (channel in ipcMain._invokeHandlers) {
```
MarshallOfSound: This will allow people to do _Bad Things_ ™️ with the object prototype.  As outlined above this should be changed to be a `Map` and a `has` check

docs/api/structures/ipc-main-event.md
```diff
@@ -3,6 +3,3 @@
 * `frameId` Integer - The ID of the renderer frame that sent this message
 * `returnValue` any - Set this to the value to be returned in a syncronous message
 * `sender` WebContents - Returns the `webContents` that sent the message
-* `reply` Function - A function that will send an IPC message to the renderer frame that sent the original message that you are currently handling.  You should use this method to "reply" to the sent message in order to guaruntee the reply will go to the correct process and frame.
```
MarshallOfSound: I think we should leave this documented for normal `ipc` events but have a new type `IpcMainHandleableEvent` that does not have a `reply` property.

Not set on that naming btw, should probs me more like

`IpcMainEvent` and `IpcMainReplyableEvent` where the replyable event `extends` the `IpcMainEvent`

lib/browser/api/web-contents.js
```diff
@@ -326,6 +326,17 @@ WebContents.prototype._init = function () {
     }
   })
 
+  this.on('-ipc-invoke', function (event, channel, args) {
+    event.reply = (result) => event.sendReply({ result })
+    event.throw = (error) => event.sendReply({ error: error.toString() })
```
miniak: use `serialize` from `lib/common/error-utils.js`?

lib/renderer/api/ipc-renderer.js
```diff
@@ -27,4 +27,11 @@ ipcRenderer.sendToAll = function (webContentsId, channel, ...args) {
   return ipc.sendTo(internal, true, webContentsId, channel, args)
 }
 
+ipcRenderer.invoke = function (channel, ...args) {
+  return ipc.invoke(channel, args).then(({ error, result }) => {
+    if (error) { throw new Error(`Error invoking remote method '${channel}': ${error}`) }
```
miniak: use `deserialize` from `lib/common/error-utils.js`?

lib/browser/api/ipc-main.ts
nornagon: made it a `Map` instead. These functions don't return `Promise`s, since they wrap `event.reply`/`event.throw`.

lib/browser/api/web-contents.js
nornagon: just following other handlers. it could be useful to blanket catch all invokes? but maybe it should just be removed.

lib/renderer/api/ipc-renderer.js
nornagon: Hmmm, I'm not sure we want to re-throw the same error that was emitted in the main process.

1. When the error is thrown in the renderer process, it's important to know that the error came _from an invoke() call_, which `errorUtils.deserialize` masks. Otherwise it could be very confusing to see a stack trace from the main process in the renderer.
2. JavaScript in the renderer process shouldn't have access to details about the main process by default. Stack traces can contain potentially sensitive information. I'm cautious about the security implications of transmitting stack data from the main process to the renderer.

If this was Java I'd suggest we use [exception causes](https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html) but JS doesn't seem to have an equivalent.

lib/browser/api/web-contents.js
nornagon: discussion below; resolving this comment.

lib/renderer/api/ipc-renderer.js
MarshallOfSound: I agree the stack shouldn't be sent, but it should still be traceable IMO.

E.g. Log the error in the main process with an error ID, send the error ID to the renderer so that is can say "Error 1234 in the main process logs"

lib/browser/api/ipc-main.ts
```diff
@@ -1,8 +1,40 @@
 import { EventEmitter } from 'events'
+import { IpcMainInvokeEvent } from 'electron'
 
-const emitter = new EventEmitter()
+class IpcMain extends EventEmitter {
```
MarshallOfSound: We can set this up so that you don't have to type `handle` and `handleOnce` and such.  Also ensuring type safety / matching of our docs.

```
class IpcMain extends EventEmitter implements Electron.IpcMain {
    handle: Electron.IpcMain['handle'] = (method, fn) => {

    }

    handleOnce: Electron.IpcMain['handleOnce'] = (method, fn) => {

    }
}
```

In that example both `method` and `fn` have types from the docs rather than re-declaring

lib/browser/api/web-contents.js
MarshallOfSound: The others are documented here --> https://electronjs.org/docs/api/web-contents#event-ipc-message and are used to as you say, catch all the IPC.

Not sure that in the case of invokes where we expect a 1:1 relationship between message and handler it makes sense.  Up to you though 👍 Either remove or document as API with the others

nornagon: i'm going to remove it for now in favour of a smaller API surface. if we add it back in future we should consider things like:
- should the catch-all handler be able to e.g. `event.preventDefault` to prevent the registered handler from executing?
- should this be implemented as an event on the webcontents or an event on ipcMain?
- if the latter, should it be an event or should it be something like `setDefaultInvokeHandler`?

etc.

lib/renderer/api/ipc-renderer.js
nornagon: Hm... yeah, I agree that the error stack should be visible _somewhere_. The error ID thing is neat but is kind of a junk developer experience.

Ideally someone would convince me that passing back `e.stack` info to `invoke` callers is O.K. security wise (since e.g. sandbox / contextIsolation should be preventing the data from leaking). And then we could add the stack trace in after the existing stack, something like `e.stack += '\nCaused by: ' + remoteError.stack`

docs/api/ipc-main.md
nornagon: Along the lines of the other conversation about the catch-all `ipc-invoke` event and keeping API surface down, I'm going to pass on this for now.

lib/renderer/api/ipc-renderer.js
nornagon: I settled on a compromise for now: we log the error in the main process without any IDs to match. 95% of the time I think it will be obvious which error in the main process matches with which error in the renderer process (we print the channel and the error message in both main & renderer).

lib/browser/api/web-contents.js
```diff
@@ -326,6 +326,19 @@ WebContents.prototype._init = function () {
     }
   })
 
+  this.on('-ipc-invoke', function (event, channel, args) {
+    event._reply = (result) => event.sendReply({ result })
+    event._throw = (error) => {
+      console.error(`Error occurred in handler for '${channel}':`, error)
+      event.sendReply({ error: error.toString() })
+    }
+    if (ipcMain._invokeHandlers.has(channel)) {
```
MarshallOfSound: This property is private, how does this work?

nornagon: it's `.js`, not `.ts`.

PR #18452, opened 2019-05-25
fix: do not crash when async id check fails

### LINKED ISSUE #12798: Unhandled errors in renderer in node context terminate the renderer
* Electron version: 1.8.6 
Does not occur on 2.0.0
* Operating system: Win 10

Paste the following into the devtools console and the renderer terminates:
```
window.setImmediate(() => {throw new Error('Die!')});
```
Adding a `uncaughtException` listener on process stops the renderer from exiting
```
process.on('uncaughtException', () => {})
```
That difference in behaviour appears to be due to this code in `renderer/init.js`:
```
  // Redirect window.onerror to uncaughtException.
  window.onerror = function (message, filename, lineno, colno, error) {
    if (global.process.listeners('uncaughtException').length > 0) {
      global.process.emit('uncaughtException', error)
      return true
    } else {
      return false
    }
  }
```

zcbenz:
#### Description of Change

Close #12798.

Due to the way node integration works in Electron, the async hooks can not correctly track the execution of async calls under certain cases. We should eventually find out how to make async hooks work correctly in Electron, but for now we just disable the check to avoid hard crashes.

The async hooks will not be able to track those executions which crashed before, but there are no other side effects.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix crash when throwing Error in `setImmediate`

### DIFF
--- a/spec/node-spec.js
+++ b/spec/node-spec.js
@@ -393,6 +393,22 @@ describe('node feature', () => {
           setImmediate(() => setImmediate(done))
         })
       })
+
+      it('does not crash with exception in it', (done) => {
+        setImmediate(() => {
+          expect(() => { throw new Error() }).to.throw(Error)
+          done()
+        })
+      })
+    })
+
+    describe('async hooks', () => {
+      it('does not crash', () => {
+        const asyncHooks = require('async_hooks')
+        const hook = asyncHooks.createHook({ init: function () {} })
+        after(() => hook.disable())
+        hook.enable()
+      })
     })
   })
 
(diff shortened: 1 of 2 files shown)


spec/node-spec.js
```diff
@@ -393,6 +393,21 @@ describe('node feature', () => {
...
+          done()
+        })
+      })
+    })
+
+    describe('async hooks', () => {
+      it('does not crash', () => {
+        const asyncHooks = require('async_hooks')
+        const hook = asyncHooks.createHook({ init: function () {} })
+        hook.enable()
```
nornagon: can we disable this after the test is done?

zcbenz: I have updated the PR, will update the back ports after this gets merged.

nornagon:
I'm concerned that just failing to crash in this situation will cause more subtle and hard-to-detect failures later on as the async hook stack gets corrupted.

I think we need a real fix for this, I'm not convinced that merging this will truly improve the situation.

PR #18455, opened 2019-05-25
fix: do not crash when async id check fails (4-2-x)

zcbenz:
Backport of #18452.

Notes: Fix crash when throwing Error in setImmediate.


zcbenz:
I'm closing this for now until #18452 is approved.

PR #18456, opened 2019-05-25
fix: do not crash when async id check fails (3-1-x)

zcbenz:
Backport of #18452.

Notes: Fix crash when throwing Error in setImmediate.

### DIFF
--- a/atom/common/node_bindings.cc
+++ b/atom/common/node_bindings.cc
@@ -261,6 +261,14 @@ node::Environment* NodeBindings::CreateEnvironment(
       node::CreateIsolateData(context->GetIsolate(), uv_loop_, platform),
       context, args.size(), c_argv.get(), 0, nullptr);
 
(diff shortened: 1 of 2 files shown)


zcbenz:
The Windows CI fails with linking error:

```
electron_lib.lib(electron_lib.node_bindings.obj) : error LNK2019: unresolved external symbol "void __cdecl node::Assert(char const * (*)[4])" (?Assert@node@@YAXPEAY03QEBD@Z) referenced in function "public: unsigned int const __cd
1889ecl node::AliasedBuffer<unsigned int,class v8::Uint32Array>::GetValue(unsigned __int64)const " (?GetValue@?$AliasedBuffer@IVUint32Array@v8@@@node@@QEBA?BI_K@Z)
1890
```

which seems not related to this PR.

zcbenz:
I'm closing this for now until #18452 is approved.

PR #18464, opened 2019-05-27
feat: migrate protocol module to NetworkService (Part 10)

zcbenz:
#### Description of Change

This PR fixes the problem that redirection does not work for intercepted schemes.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/browser/net/atom_url_loader_factory.cc
+++ b/atom/browser/net/atom_url_loader_factory.cc
@@ -25,6 +25,7 @@
 #include "net/http/http_status_code.h"
 #include "services/network/public/cpp/url_loader_completion_status.h"
 #include "services/network/public/mojom/url_loader_factory.mojom.h"
+#include "third_party/blink/renderer/platform/network/network_utils.h"
 
 #include "atom/common/node_includes.h"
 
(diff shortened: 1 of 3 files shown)


atom/browser/net/atom_url_loader_factory.cc
```diff
@@ -216,6 +218,43 @@ void AtomURLLoaderFactory::StartLoading(
...
+    // loader with default factory.
+    //
+    // Note that when handling requests for intercepted scheme, creating loader
+    // with default factory (i.e. calling StartLoadingHttp) would bypass the
+    // ProxyingURLLoaderFactory, we have to explicitly use the proxy factory to
+    // create loader so it is possible to have handlers of intercepted scheme
+    // getting called recursively, which is a behavior expected in protocol
+    // module.
+    //
+    // I'm not sure whether this is an intended behavior in Chromium.
```
deepak1556: I think these are fine with our current implementation of `ProxyingURLLoaderFactory`, but when we redesign to an implementation that uses one `AtomURLLoaderFactory` per network context, we would have to close the current loader connection , so that the redirect request will create a new one and keeps the right reference count of the class.

Issue #18476, opened 2019-05-28
The `services` MenuItem role displays incorrectly in context menus


caesar:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.2
* **Operating System:**
  * macOS 10.14.5
* **Last Known Working Electron version:**
  * none

### Expected Behavior
When used in a context menu (or anywhere other than the main Application menu), a submenu with `role: services` should look like a normal menu, without icons or headings, like in native Mac apps.

Screenshot from Safari  (expected behaviour):
<img width="537" alt="Screenshot 2019-05-26 at 20 44 02" src="https://user-images.githubusercontent.com/1566516/58491915-ca783180-8167-11e9-94be-82c0efa12a3e.png">

### Actual Behavior
A MenuItem with `role: services` always  displays a menu like that used in the main Application menu, with icons and  headings.

Screenshot from Electron (actual behaviour):
<img width="636" alt="Screenshot 2019-05-26 at 21 16 28" src="https://user-images.githubusercontent.com/1566516/58491986-ee3b7780-8167-11e9-9771-612e84753420.png">

### To Reproduce
Create a MenuItem with `role: service`  in a context menu (or anywhere other than in the main menubar).

codebytere:
A cursory look at this tells me we're not distinguishing between [`NSServices`](https://developer.apple.com/documentation/bundleresources/information_property_list/nsservices?language=objc), which are the services a specific app enables, and the bespoke [`servicesMenu`](https://developer.apple.com/documentation/appkit/nsapplication/1428608-servicesmenu?language=objc). For now, the best short term fix might be to only enable the `servicesMenu` on non-context menus, but perhaps eventually we can look into the ability to self-populate `NSServices.`

caesar:
@codebytere Actually I don't think it's related to `NSServices`, which are the services *provided by* the *current* app. This menu should be listing services from *other* apps.

As you say, the problem is that the context menu item is being wrongly designated as a `servicesMenu`. I haven't been able to find what should be used instead, despite some fairly deep digging through Apple's docs.
[...]

PR #18484, opened 2019-05-28
docs: specify use case for the 'services' role

codebytere:
#### Description of Change

Partially addresses https://github.com/electron/electron/issues/18476.

Specifies that the `services` role is not the same as a menu item populated with [NSServices](https://developer.apple.com/documentation/bundleresources/information_property_list/nsservices?language=objc) and should not be used in context menus.

cc @nornagon @miniak @MarshallOfSound

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/menu-item.md
+++ b/docs/api/menu-item.md
@@ -108,7 +108,7 @@ The following additional roles are available on _macOS_:
 * `moveTabToNewWindow` - Map to the `moveTabToNewWindow` action.
 * `window` - The submenu is a "Window" menu.
 * `help` - The submenu is a "Help" menu.
-* `services` - The submenu is a "Services" menu.
+* `services` - The submenu is a ["Services"](https://developer.apple.com/documentation/appkit/nsapplication/1428608-servicesmenu?language=objc) menu; this is not the same as a menu item populated with [NSServices](https://developer.apple.com/documentation/bundleresources/information_property_list/nsservices?language=objc) and should not be used in context menus.
 * `recentDocuments` - The submenu is an "Open Recent" menu.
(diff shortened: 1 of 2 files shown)


docs/api/menu-item.md
```diff
@@ -108,7 +108,7 @@ The following additional roles are available on _macOS_:
 * `moveTabToNewWindow` - Map to the `moveTabToNewWindow` action.
 * `window` - The submenu is a "Window" menu.
 * `help` - The submenu is a "Help" menu.
-* `services` - The submenu is a "Services" menu.
+* `services` - The submenu is a ["Services"](https://developer.apple.com/documentation/appkit/nsapplication/1428608-servicesmenu?language=objc) menu; this is not the same as a menu item populated with [NSServices](https://developer.apple.com/documentation/bundleresources/information_property_list/nsservices?language=objc) and should not be used in context menus.
```
caesar: > a menu item populated with [NSServices]

I'm not sure this is a correct description. The "Services" submenu in a context menu is not populated with our own services (as defined by the `NSServices` plist key), but rather with services from other apps – just like the Application menu's services menu, but with a different layout and less entries.

Assuming the short term solution is indeed just to change the documentation, how about:

```suggestion
* `services` - The submenu is a ["Services"](https://developer.apple.com/documentation/appkit/nsapplication/1428608-servicesmenu?language=objc) menu. This is only intended for use in the Application Menu and is *not* the same as the "Services" submenu menu used in context menus in macOS apps, which is not implemented in Electron.
```

codebytere: thanks ⭐️

PR #18488, opened 2019-05-28
fix: Update gen-filenames.js to generate posix paths

nitsakh:
#### Description of Change
gen-filenames.js on win creates relative paths with `\`.
We want to have posix path separators.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/gen-filenames.js
+++ b/script/gen-filenames.js
@@ -73,7 +73,7 @@ const main = async () => {
       // Ignore empty lines
       .filter(line => line)
       // Get the relative path
-      .map(line => path.relative(rootPath, line))
+      .map(line => path.relative(rootPath, line).replace(/\\/g, '/'))
       // Only care about files in //electron
       .filter(line => !line.startsWith('..'))
       // Only care about our own files


script/gen-filenames.js
```diff
@@ -73,7 +73,7 @@ const main = async () => {
       // Ignore empty lines
       .filter(line => line)
       // Get the relative path
-      .map(line => path.relative(rootPath, line))
+      .map(line => path.relative(rootPath, line).replace(/\\/g, '/'))
```
nornagon:
```suggestion
      .map(line => path.posix.relative(rootPath, line))
```

nornagon: https://nodejs.org/api/path.html#path_windows_vs_posix

nitsakh: I tried that first, it didn't work and returned empty. I thought `path.posix.relative` means "the given path is posix, resolve that and give me the result" and so in this case the result was empty.
I can check the reason again.

nornagon: ah, ok. it seems to me that there ought to be a more rigorous way of handling paths but fine, lgtm for now, i suppose that if we put `\\` or `/` in any file names in our repo then we deserve what we get.

PR #18489, opened 2019-05-28
test: run nan tests to ensure v8 compat has not broken

MarshallOfSound:
As discussed in the Upgrades WG meeting this morning we should probably be running some nan tests to ensure that native modules are gonna be 🆗 on each roll.

This adds some scripts to run `nan`'s test suite on every PR for both x64 and ia32 linux.  Currently depending on a fork until https://github.com/nodejs/nan/pull/857 has been merged upstream

Notes: no-notes

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -13,6 +13,8 @@ vars = {
     'ab588d36191964c4bca8de5c320534d95606c861',
   'node_version':
     'a86a4a160dc520c61a602c949a32a1bc4c0fc633',
+  'nan_version':
+    '960dd6c70fc9eb136efdf37b4bef18fadbc3436f',
 
   'boto_version': 'f7574aa6cc2c819430c1f05e9a1a1a666ef8169b',
   'pyyaml_version': '3.12',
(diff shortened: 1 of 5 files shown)


DEPS
```diff
@@ -21,6 +23,8 @@ vars = {
   'boto_git': 'https://github.com/boto',
   'chromium_git': 'https://chromium.googlesource.com',
   'electron_git': 'https://github.com/electron',
+  # FIXME: Once https://github.com/nodejs/nan/pull/857 lands this should point at nodejs/nan
+  'nodejs_git': 'https://github.com/marshallofsound',
   'requests_git': 'https://github.com/kennethreitz',
   'yaml_git': 'https://github.com/yaml',
```
nornagon: can't comment on the line below but `KEEP IN SYNC WITH` should be updated

script/nan-spec-runner.js
```diff
@@ -0,0 +1,62 @@
...
+    env,
+    cwd: NAN_DIR,
+    stdio: 'inherit'
+  })
+  if (installStatus !== 0) {
+    console.error('Failed to install nan node_modules')
+    return process.exit(installStatus)
+  }
+
+  const TESTS_TO_FIX = ['nannew-test.js']
```
nornagon:
```suggestion
  const DISABLED_TESTS = ['nannew-test.js']
```

MarshallOfSound:
@nornagon it's DCHECK's while opening handles.  I'm gonna build a debug build of node and try run the Nan tests against it and see if they DCHECK then go from there

Issue #13770, opened 2018-07-23
Re-enable mixed-sandbox tests for Windows CI


zcbenz:
As described in https://github.com/electron/electron/pull/13769, we had to disable mixed sandbox related tests for Windows CI. We should re-enable them back after finding out the root reason.

nornagon:
Adding this as a blocker for 3.0.x stable, since the issues causing the sandbox tests to fail were backported & I think it's not OK to call 3.0.x stable if the sandbox tests are consistently failing in CI :)

jkleinsc:
I'm still tracking down the root issue but here is what I have found out so far:
The socket never fires the `data` event expected here:
https://github.com/electron/electron/blob/master/spec/api-app-spec.js#L882

However the socket does fire an `end` event and a `close` event.  So it appears that even though the client sends data via end, the `data` event is not firing for some reason.

Here is the change I made to api-app-spec.js to check what events are firing on the socket:
```js
       server.on('connection', client => {
          console.log('got connection')
          client.once('error', error => {
            console.log('client error', error)
            done(error)
          })
          client.once('close', hadError => {
            console.log('client close', hadError)
          })
          client.once('connect', () => {
            console.log('client connect')
          })
          client.once('drain', () => {
            console.log('client drain')
          })
          client.once('end', () => {
            console.log('client end')
          })
          client.once('lookup', () => {
            console.log('client lookup')
          })
          client.once('ready', () => {
            console.log('client ready')
          })
          client.once('timeout', () => {
            console.log('client timeout')
          })

          client.once('data', data => {
            console.log('got data')
            const argv = JSON.parse(data)
            expect(argv.sandbox).to.include('--enable-sandbox')
            expect(argv.sandbox).to.not.include('--no-sandbox')

            expect(argv.noSandbox).to.not.include('--enable-sandbox')
            expect(argv.noSandbox).to.include('--no-sandbox')

            done()
          })
        })
```

And here is the output when running the test:
```
got connection
client end
client close false
```

cc @zcbenz in case you have any idea why the `data` event wouldn't be firing here.

jkleinsc:
It turns out this issue was caused by the fact that I turned on Chrome's logging via the ELECTRON_ENABLE_LOGGING environment variable on our AppVeyor CI.  This cause a DCHECK to throw an error:
```
[...]

PR #18523, opened 2019-05-30
build: fixes for CI flakes in 3-1-x

jkleinsc:
Currently in 3-1-x both Windows and Linux CI are erroring out with two separate issues:
1. Windows flakes on the `app module mixed sandbox` tests.  This is due to https://github.com/electron/electron/issues/13770#issuecomment-412199748.  This PR fixes that issue by turning off ELECTRON_ENABLE_LOGGING/ELECTRON_ENABLE_STACK_DUMPING on Appveyor builds.

2. Linux builds are timing out but passing tests.  It turns out that dbusmock was hanging  on `dbus_stop`.  This PR changes the docker container to use a newer version of python-dbusmock which adds timeouts to dbus_stop.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


Dockerfile
```diff
@@ -13,12 +12,19 @@ RUN apt-get install -y nodejs
 # Install wget used by crash reporter
 RUN apt-get install -y wget
 
-# Install python-dbusmock
-RUN apt-get install -y python-dbusmock
+# Install python-dbus
+RUN apt-get install -y python-dbus
+RUN pip install python-dbusmock
```
alexeykuzmin: It looks like it can make sense to port this piece to all other release branches.
Just to avoid issues with `python-dbusmock` in case there will be a need to rebuild a Docker container for any of them.

jkleinsc: @alexeykuzmin the newer branches use a newer base docker image which already pulls in a newer version of python-dbusmock, so it is only really needed for 3-1-x

alexeykuzmin: @jkleinsc 

I've recently built a Docker container from the `4-2-x` branch `Dockerfile` to run tests in, and often see exactly the same failures because of dbus timeouts.

So it looks like the current Docker containers used for E4 builds just happen to have a version of `python-dbusmock` not affected by that issue.
So I _think_ that `pip install python-dbusmock` should be used for E4+ builds as well to make sure we always use a newer version of `python-dbusmock`, which doesn't have the timeout issue.

alexeykuzmin: By the way, do you know exactly which version of `python-dbusmock` works fine?
Should be specify this version in the `Dockerfile` explicitly?

PR #18532, opened 2019-05-30
refactor: simplify content script injection

nornagon:
#### Description of Change
This replaces the substantial `RenderProcessPreferences` machinery with a much simpler synchronous IPC to fetch the content scripts that need injecting. Ultimately this should be replaced by the real extensions API (ref #17440), but this works for now and lets us remove a bunch of code.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


atom/common/node_bindings.cc
```diff
@@ -29,43 +29,42 @@
...
-  V(atom_browser_download_item)              \
-  V(atom_browser_event)                      \
-  V(atom_browser_global_shortcut)            \
-  V(atom_browser_in_app_purchase)            \
-  V(atom_browser_menu)                       \
-  V(atom_browser_net)                        \
-  V(atom_browser_power_monitor)              \
-  V(atom_browser_power_save_blocker)         \
-  V(atom_browser_protocol)                   \
-  V(atom_browser_render_process_preferences) \
```
nornagon: this is the line i deleted (cross-check with `?w=1`)

MarshallOfSound:
Do we know the reasoning for the original design, inspired by Chromium, performance, etc?

nornagon:
cc @zcbenz who wrote this initially, I believe

nornagon:
I think this also fixes a potential race condition maybe? It's not clear to me that the `AtomMsg_UpdatePreferences` IPC is guaranteed to arrive before the content script injector runs.

atom/browser/api/atom_api_app.cc
```diff
@@ -807,6 +807,14 @@ void App::BrowserChildProcessKilled(
...
   ChildProcessLaunched(content::PROCESS_TYPE_RENDERER,
                        host->GetProcess().Handle());
+
+  // TODO(jeremy): this isn't really the right place to be creating
+  // `WebContents` instances, but this was implicitly happening before in
+  // `RenderProcessPreferences`, so this is at least more explicit...
+  content::WebContents* web_contents =
+      AtomBrowserClient::Get()->GetWebContentsFromProcessID(host->GetID());
+  if (web_contents)
+    WebContents::FromOrCreate(v8::Isolate::GetCurrent(), web_contents);
```
deepak1556: IIUC the previous version in `IsWebContents` was only there to make sure we update preferences for `BrowserWindow` or `WebView` and not for a devtools `Remote` type. Since the ipc was sent when render process being created without knowing what its contents were. But in this PR, the ipc is made by the renderer process so it chooses the right target `WebContents`, so i don't think this wrapper creation is necessary.

nornagon: If this line is deleted, devtools stops opening.

lib/renderer/content-scripts-injector.ts
```diff
@@ -108,15 +108,17 @@ ipcRendererUtils.handle('CHROME_TABS_EXECUTE_SCRIPT', function (
...
-        for (const script of pref.contentScripts) {
-          injectContentScript(pref.extensionId, script)
-        }
+type ContentScriptEntry = {
+  extensionId: string;
+  contentScripts: Electron.ContentScript[];
+}
+
+module.exports = () => {
+  const entries = ipcRendererUtils.invokeSync<ContentScriptEntry[]>('CHROME_GET_CONTENT_SCRIPTS')
```
MarshallOfSound: Given @zcbenz 's comment I'm wondering how much of an impact this IPC can have, could you do a quick benchmark with `console.time` around this call to see how much slower the sync IPC call is compared to the previous voodoo.  Wouldn't block the PR but an interesting stat we should be aware of

MarshallOfSound: Especially considering this would run on _every_ page load 🤔

zcbenz: It would add about 200ms load time: https://github.com/atom/atom/pull/1503.

zcbenz:
> I think this also fixes a potential race condition maybe? It's not clear to me that the `AtomMsg_UpdatePreferences` IPC is guaranteed to arrive before the content script injector runs.

Sending IPC messages immediately after renderer process gets created should be able to guarantee the message get received before page loading, since page loading is started by sending IPC messages too. This pattern is copied from Chromium.

lib/renderer/content-scripts-injector.ts
MarshallOfSound: @nornagon If the 200ms number above is still accurate (2014 was a while ago) I'd prefer not to land this.  This is a hefty price to pay that we could simply avoid by _not_ doing this 🤔

nornagon: I timed it at about 1.6ms.

nornagon: (I think the PR referenced removes the crashreporter startup, which waits for a new process to start, which on Windows can take 100-200ms. This just returns a string.)

nornagon:
1.6ms is still longer than I'd like, but I think it's within acceptable range considering the reduction in complexity and that we'll eventually be switching to "real" extensions.

atom/browser/api/atom_api_app.cc
deepak1556: @zcbenz whats the intent behind wrapping `api::WebContents` at render process start for devtools ?

zcbenz: Hmm I am not sure why this is related to devtools, I would need to do some debugging to know.

nornagon: Given that this preserves the existing behaviour (i.e. calling `WebContents::FromOrCreate` on every render process at creation time), I'm going to go ahead and merge this, but it does seem odd.

nornagon:
linux failures are flake (see #18630), merging.

PR #18536, opened 2019-05-31
refactor: mojofy draggable regions

nornagon:
#### Description of Change
Convert the `UpdateDraggableRegions` ipc to mojo.

Draft because there's still a little more cleanup I want to do:
- ~the redundant `OnDraggableRegionsUpdated` that just proxies straight to `UpdateDraggableRegions`—that should be collapsed, probably just by renaming `UpdateDraggableRegions` to `OnDraggableRegionsUpdated`~ decided to keep the separate function but dropped the unused `RenderFrameHost` parameter.
- ~it's not clear to me whether anything other than a single `BrowserWindow` is ever the `ExtendedWebContentsObserver`, so maybe that layer should be collapsed too.~ not in this PR.
- ~I'd like to consider moving the `UpdateDraggableRegions` mojo ipc to a separate mojo interface that's only exposed to processes that have a BrowserWindow behind them~ I'll consider this refactor later.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


atom/browser/api/atom_api_browser_window_mac.mm
```diff
@@ -103,7 +101,9 @@ - (NSView*)hitTest:(NSPoint)aPoint {
 
   // Draggable regions is implemented by having the whole web view draggable
   // (mouseDownCanMoveWindow) and overlaying regions that are not draggable.
-  draggable_regions_ = regions;
+  draggable_regions_.clear();
+  for (const auto& r : regions)
+    draggable_regions_.push_back(r.Clone());
```
deepak1556: could use `swap`, to avoid `clear` and `push_back`

nornagon: we have to `Clone` every element though... i'm not sure I understand how `swap` would accomplish that?

deepak1556: Ah vector of mojo ptrs, nvm.

PR #18546, opened 2019-05-31
build: remove Vulkan mock ICD

miniak:
#### Description of Change
Removes from electron.zip:
- VkICD_mock_icd.dll
- libVkICD_mock_icd.so

Chrome is not shipping those files either. Follow up to #17985

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Removed Vulkan mock ICD from electron.zip, which is only meant to be used for Chromium development.

### DIFF
--- a/build/zip.py
+++ b/build/zip.py
@@ -12,8 +12,8 @@ EXTENSIONS_TO_SKIP = [
 
 PATHS_TO_SKIP = [
   'angledata', #Skipping because it is an output of //ui/gl that we don't need
-  './libVkLayer_', #Skipping because these are outputs that we don't need
-  './VkLayerLayer_', #Skipping because these are outputs that we don't need
+  './libVkICD_mock_', #Skipping because these are outputs that we don't need
+  './VkICD_mock_', #Skipping because these are outputs that we don't need
 
   # //chrome/browser:resources depends on this via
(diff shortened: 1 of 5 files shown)


build/zip.py
```diff
@@ -13,7 +13,9 @@
 PATHS_TO_SKIP = [
   'angledata', #Skipping because it is an output of //ui/gl that we don't need
   './libVkLayer_', #Skipping because these are outputs that we don't need
-  './VkLayerLayer_', #Skipping because these are outputs that we don't need
+  './libVkICD_mock_', #Skipping because these are outputs that we don't need
+  './VkLayer_', #Skipping because these are outputs that we don't need
```
nornagon: this is a new skip but i don't see any changes in the zip manifests related to it; is it still needed?

miniak: VkLayerLayer was incorrect, fixing to VkLayer. This was handled by disabling building of the debug layers, however keeping it here just in case. I can remove the the entries for debug layers though.

alexeykuzmin: > keeping it here just in case

If it not used, maybe it can be safely deleted?

miniak: removed

miniak: @nornagon btw is there a reason why we don't have the zip manifests on Windows?

PR #18553, opened 2019-05-31
refactor: bundle the browser and renderer process electron code

MarshallOfSound:
* Bundles browser/init, renderer/init and worker/init
  * Improves load performance of electron main process init by ~85% (90ms -> 10ms)
  * Improves load performance of renderer process init by ~50%
* Prevents users from importing our "requiring" our internal logic such
as `ipc-main-internal`.  This makes those message buses safer as they are
less accessible, there is still some more work to be done though to lock
down those buses completely.
* The electron.asar file now only contains 2 files, as a future
improvement maybe we can use atom_natives to ship these two files
embedded in the binary
* This also removes our dependency on browserify which had some strange
edge cases that caused us to have to hack around require-order and
stopped us using certain ES6/7 features we should have been able to use
(async / await in some files in the sandboxed renderer init script)
* Instead we now use `webpack` to bundle All The Things

TLDR: Things are faster and better :)

Notes: no-notes


lib/browser/api/module-keys.js
```diff
@@ -0,0 +1,48 @@
+'use strict'
+
+// TODO: Figure out a way to not duplicate this information between here and module-list
```
nornagon: it's unclear to me why this duplication is needed, could you write a little here in the comment about why it's necessary?

filenames.auto.gni
```diff
@@ -176,8 +181,138 @@ auto_filenames = {
...
+    "lib/browser/api/power-save-blocker.js",
+    "lib/browser/api/protocol.ts",
+    "lib/browser/api/screen.ts",
+    "lib/browser/api/session.js",
+    "lib/browser/api/system-preferences.js",
+    "lib/browser/api/top-level-window.js",
+    "lib/browser/api/touch-bar.js",
+    "lib/browser/api/tray.js",
+    "lib/browser/api/view.js",
+    "lib/browser/api/views/box-layout.js",
```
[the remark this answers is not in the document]
MarshallOfSound: @nornagon Yes, they are lazy loaded and will not be loaded if the feature is disabled at runtime.  In the future we could use the webpack defines plugin to achieve the same thing but I don't believe it is required 🤔

lib/browser/api/module-keys.js
MarshallOfSound: Yeah I'll add a comment, high level because `module-list` require's all the browser process API files it means that `renderer/api/remote` (which now requires this file) loaded every single browser API into the renderer process bundle which broke some things.

Renderer code shouldn't depend on browser code IMO so need to figure out a way to split these concerns.  Once it's split I plan to enforce it in `filenames.auto.gni` that the renderer bundles can't depend on browser process code and vice-versa

filenames.auto.gni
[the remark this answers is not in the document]
MarshallOfSound: I also have a TODO somewhere now that we're moving to webpack to make things equivilant to `#if BUILDFLAG(ENABLE_*)` work in webpack (you can do this and actually have webpack strip out the un-accessible code at build time.

lib/browser/api/module-keys.js
MarshallOfSound: Added a comment

MarshallOfSound:
FWIW macOS is not a flake, please do not merge 👍

PR #18396, opened 2019-05-21
feat: Add option to conditionally disable site instance patches

MarshallOfSound:
#### Description of Change
This adds a new `app.disableElectronSiteInstanceOverrides` that allows us to dynamically (at runtime) disable the site instance override patches.  The intention is to:
* have this PR land in Electron 6
* have the first deprecation warning to land in Electron 7 for non-context aware native modules
* have a deprecation warning land in Electron 8 for the default value of `app.disableElectronSiteInstanceOverrides` to switch
* switch the default value of `app.disableElectronSiteInstanceOverrides` in Electron 9
* Deprecate the ability to change `app.disableElectronSiteInstanceOverrides` in Electron 10
* Remove the ability to change `app.disableElectronSiteInstanceOverrides` in Electron 11

Yes you're reading that right, 5 major versions 👍 (60~ish weeks)

For more info see this [Discussion Issue](https://github.com/electron/electron/issues/18397)

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Still TODO

- [x] Find a way to prevent loading native modules that are not context aware / NAPI in the renderer process if this new mode is enabled.  **Required for this PR**
- [x] Add a `console.warn` for loading native modules in the renderer process that are not context-aware or NAPI - This warning will ship in Electron 7

#### Release Notes

Notes: Added new `app.disableElectronSiteInstanceOverrides` property that allows apps to disable the site instance overrides Electron has patched into Chromium.  This can be used to prevent the automatic renderer process restarting that Electron currently does.

PR #18554, opened 2019-05-31
feat: Add option to conditionally disable site instance patches

MarshallOfSound:
Backport of #18396

See that PR for details.

Notes: Added new `app.allowRendererProcessReuse` property that allows apps to disable the site instance overrides Electron has patched into Chromium. This can be used to prevent the automatic renderer process restarting that Electron currently does.


MarshallOfSound:
The only conflicts in the cherry pick were in `app.md` and `DEPS`.  Other than that it was a clean cherry pick 👍

atom/renderer/atom_renderer_client.cc
```diff
@@ -103,6 +103,12 @@ void AtomRendererClient::DidCreateScriptContext(
 
   // Setup node environment for each window.
   node::Environment* env = node_bindings_->CreateEnvironment(context);
+  auto* command_line = base::CommandLine::ForCurrentProcess();
+  // If we have disabled the site instance overrides we should prevent loading
+  // any non-context aware native module
+  if (command_line->HasSwitch(switches::kDisableElectronSiteInstanceOverrides))
+    env->ForceOnlyContextAwareNativeModules();
+
```
deepak1556: `env->WarnNonContextAwareNativeModules()` is not needed for 6-0-x ?

MarshallOfSound: @deepak1556 Nope, the warning should only be present on `7-0-x` 👍

PR #18567, opened 2019-06-01
fix: Fixed all OOB accesses in VertexProgram and PixelProgram

miniak:
#### Description of Change
Backport fixes for:
- https://bugs.chromium.org/p/chromium/issues/detail?id=915197
- https://bugs.chromium.org/p/chromium/issues/detail?id=915206
- https://bugs.chromium.org/p/chromium/issues/detail?id=915218

https://swiftshader.googlesource.com/SwiftShader.git/+/0e6a044dff9facf499739be93536bd7f7cc4c6bf
> Fix undefined behavior in OFFSET().
> 
> Accessing members of a null pointer is undefined behavior, even when
> only used to obtain the address again. So use a non-zero value as the
> base pointer address instead. 32 was chosen to provide sufficient
> alignment guarantees.

https://swiftshader.googlesource.com/SwiftShader.git/+/48d47a4912684bfa4d23d3cdaab60046bf3d0468
> Fixed all OOB accesses in VertexProgram and PixelProgram
> 
> A lot of arrays in VertexProgram and PixelProgram have fixed sizes,
> so programs that have more nested loops or ifs or deeper call stacks
> can cause OOB accesses, which causes security issues in Chromium.
> 
> Index clamping was added to prevent any OOB memory accesses here.
> 
> This could eventually be fixed properly by first verifying these sizes
> and giving shader compile errors when these limits are exceeded.

/cc @electron/wg-security

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Backported fix for: https://bugs.chromium.org/p/chromium/issues/detail?id=915197, https://bugs.chromium.org/p/chromium/issues/detail?id=915206 and https://bugs.chromium.org/p/chromium/issues/detail?id=915218.

### DIFF
--- a/patches/common/swiftshader/fixed_all_oob_accesses_in_vertexprogram_and_pixelprogram.patch
+++ b/patches/common/swiftshader/fixed_all_oob_accesses_in_vertexprogram_and_pixelprogram.patch
@@ -1,0 +1,508 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Alexis Hetu <sugoi@google.com>
+Date: Thu, 10 Jan 2019 14:04:26 -0500
+Subject: Fixed all OOB accesses in VertexProgram and PixelProgram
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+A lot of arrays in VertexProgram and PixelProgram have fixed sizes,
+so programs that have more nested loops or ifs or deeper call stacks
+can cause OOB accesses, which causes security issues in Chromium.
+
+Index clamping was added to prevent any OOB memory accesses here.
+
+This could eventually be fixed properly by first verifying these sizes
+and giving shader compile errors when these limits are exceeded.
+
+Bug chromium:915197 chromium:915206 chromium:915218 b/116373662
+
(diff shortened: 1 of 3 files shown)


patches/common/swiftshader/fixed_all_oob_accesses_in_vertexprogram_and_pixelprogram.patch
```diff
@@ -0,0 +1,508 @@
...
+ 	{
+-		int callSiteIndex[2048] = {0};
++		int callSiteIndex[MAX_SHADER_CALL_SITES] = {0};
+ 
+ 		for(auto &inst : instruction)
+ 		{
+ 			if(inst->opcode == OPCODE_CALL || inst->opcode == OPCODE_CALLNZ)
+ 			{
+-				int label = inst->dst.label;
++				int label = sw::min(inst->dst.label, unsigned int(MAX_SHADER_CALL_SITES));
```
zcbenz: This line does not compile:
```
../../third_party/swiftshader/src/Shader/Shader.cpp:1886:51: error: expected '(' for function-style cast or type construction
                                int label = sw::min(inst->dst.label, unsigned int(MAX_SHADER_CALL_SITES));
                                                                     ~~~~~~~~ ^
1 error generated.
```

Replacing `unsigned int` with `unsigned` should be able to fix it.

miniak:
@deepak1556 I've fixed the build error

PR #18571, opened 2019-06-02
perf: lazily create the anchor tag used for URL resolving

MarshallOfSound:
1ms is 1ms 🤷‍♂ No point making this element on the hot load path of the renderer

Notes: no-notes

### DIFF
--- a/lib/renderer/window-setup.ts
+++ b/lib/renderer/window-setup.ts
@@ -20,8 +20,9 @@ import { ipcRendererInternal } from '@electron/internal/renderer/ipc-renderer-in
 const { defineProperty } = Object
 
 // Helper function to resolve relative url.
-const a = window.document.createElement('a')
+let a: HTMLAnchorElement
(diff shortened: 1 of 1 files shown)


lib/renderer/window-setup.ts
```diff
@@ -20,8 +20,9 @@ import { ipcRendererInternal } from '@electron/internal/renderer/ipc-renderer-in
 const { defineProperty } = Object
 
 // Helper function to resolve relative url.
-const a = window.document.createElement('a')
+let a: HTMLAnchorElement
 const resolveURL = function (url: string) {
+  a = a || window.document.createElement('a')
```
nornagon: can we just use the [`URL` API](https://developer.mozilla.org/en-US/docs/Web/API/URL)?

MarshallOfSound: Unknown 🤷‍♂

PR #18572, opened 2019-06-02
fix: [FileSystem] Harden against overflows of OperationID a bit better

miniak:
#### Description of Change
Backport fix for https://bugs.chromium.org/p/chromium/issues/detail?id=925864
https://chromium.googlesource.com/chromium/src.git/+/b38064dbb21aaf32151073dcb7d594b240c68f73

> [FileSystem] Harden against overflows of OperationID a bit better.
> 
> Rather than having a UAF when OperationID overflows instead overwrite
> the old operation with the new one. Can still cause weirdness, but at
> least won't result in UAF. Also update OperationID to uint64_t to
> make sure we don't overflow to begin with.

/cc @electron/wg-security 

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Backported fix for https://bugs.chromium.org/p/chromium/issues/detail?id=925864.


alexeykuzmin:
@miniak Can you please fix conflicts?

miniak:
@alexeykuzmin done

PR #18577, opened 2019-06-02
refactor: remove electron.asar and embed JS in binary

MarshallOfSound:
#### Description of Change
Does what it says on the tin, some more performance gains post-bundling.  This embeds the three `init.js` files in the binary with js2c and includes some changes in node (https://github.com/electron/node/pull/101) that allow us to use the built-in modules as entrypoints.

#### Release Notes

Notes: The `electron.asar` file no longer exists, if you have packaging scripts that depend on it's existence you should update them.

### DIFF
--- a/lib/browser/init.ts
+++ b/lib/browser/init.ts
@@ -106,7 +106,7 @@ if (process.resourcesPath) {
   for (packagePath of searchPaths) {
     try {
       packagePath = path.join(process.resourcesPath, packagePath)
-      packageJson = __non_webpack_require__(path.join(packagePath, 'package.json')) // eslint-disable-line
+      packageJson = Module._load(path.join(packagePath, 'package.json'))
       break
     } catch {
       continue
@@ -197,6 +197,7 @@ Promise.all([
 
 if (packagePath) {
   // Finally load app's main.js and transfer control to C++.
+  process._firstFileName = Module._resolveFilename(path.join(packagePath, mainStartupScript), null, false)
   Module._load(path.join(packagePath, mainStartupScript), Module, true)
 } else {
   console.error('Failed to locate a valid package to load (app, app.asar or default_app.asar)')
--- a/lib/renderer/init.ts
+++ b/lib/renderer/init.ts
@@ -128,8 +128,9 @@ if (contextIsolation) {
 
 if (nodeIntegration) {
   // Export node bindings to global.
-  global.require = __non_webpack_require__ // eslint-disable-line
-  global.module = Module._cache[__filename]
+  const { makeRequireFunction } = __non_webpack_require__('internal/modules/cjs/helpers') // eslint-disable-line
+  global.module = new Module('electron/js2c/renderer_init')
+  global.require = makeRequireFunction(global.module)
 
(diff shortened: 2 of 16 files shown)


miniak:
@MarshallOfSound how does this change affect debugging?

MarshallOfSound:
@miniak Just means that instead of using `--inspect-brk` to debug Electron internals you have to use `--inspect-brk-node`.  I have a future plan of adding `--inspect-brk-electron` but for you can debug things just fine 👍

lib/browser/init.ts
```diff
@@ -106,7 +106,7 @@ if (process.resourcesPath) {
   for (packagePath of searchPaths) {
     try {
       packagePath = path.join(process.resourcesPath, packagePath)
-      packageJson = __non_webpack_require__(path.join(packagePath, 'package.json')) // eslint-disable-line
+      packageJson = Module._load(path.join(packagePath, 'package.json'))
```
nornagon: this seems unrelated to me...?

lib/renderer/init.ts
```diff
@@ -203,7 +205,7 @@ for (const preloadScript of preloadScripts) {
     if (!isParentDir(getAppPath(), fs.realpathSync(preloadScript))) {
       throw new Error('Preload scripts outside of app path are not allowed')
     }
-    __non_webpack_require__(preloadScript)  // eslint-disable-line
+    Module._load(preloadScript)  // eslint-disable-line
```
nornagon: what lint are we disabling here?

lib/worker/init.js
```diff
@@ -27,8 +28,9 @@ if (self.location.protocol === 'file:') {
   global.module.filename = global.__filename
 
   // Also search for module under the html file.
-  global.module.paths = global.module.paths.concat(Module._nodeModulePaths(global.__dirname))
+  global.module.paths = Module._nodeModulePaths(global.__dirname)
 } else {
-  global.__filename = __filename
-  global.__dirname = __dirname
+  // For backwards compatibility we fake these two paths here
```
nornagon: what depends on this?

lib/browser/init.ts
MarshallOfSound: `__non_webpack_require__` maps to the global node provided require function, now we are an install module that require function is `NativeModule.require` which does not load non-internal code.  We need to explicitly load non-internal code so we do what node does internally and directly call `Module._load`

lib/renderer/init.ts
MarshallOfSound: Can't remember, might be a leftover from the `__non*`

lib/worker/init.js
MarshallOfSound: User code, `__dirname` and `__filename` are globals in node land that are expected to exist and be valid.  When the internal `init` doesn't exist on disk node doesn't set these values so here I fake them to their old values.  In all likelihood we will have to keep this logic forever 👍

nornagon: wouldn't those variables be in relation to the user's files though, not the asar?

PR #18579, opened 2019-06-02
refactor: don't use export default

miniak:
#### Description of Change
Use `export = xyz` as this does not need the `handleESModule` hack, which is not reliable as `__esModule` does not necessarily mean that `default` export is used.

/cc @MarshallOfSound

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/api/app.ts
+++ b/lib/browser/api/app.ts
@@ -8,7 +8,7 @@ const commandLine = process.electronBinding('command_line')
 const { app, App } = bindings
 
 // Only one app object permitted.
-export default app
+export = app
 
 let dockMenu: Electron.Menu | null = null
 
--- a/lib/browser/api/exports/electron.js
+++ b/lib/browser/api/exports/electron.js
@@ -10,6 +10,6 @@ common.defineProperties(exports)
 for (const module of moduleList) {
   Object.defineProperty(exports, module.name, {
     enumerable: !module.private,
-    get: common.handleESModule(module.loader)
+    get: module.loader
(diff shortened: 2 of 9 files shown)


alexeykuzmin:
How can we be sure that `export default smth` won't be used by someone in the future?

codebytere:
I think there's value in ensuring compatibility with both approaches:

``
const handleESModule = (m) => {	
  // Handle Typescript modules with an "export default X" statement	
  if (m.__esModule) return m.default	
  return m	
}
```

you can change the `export default ipcMain` to `export = ipcMain` for sure but i'd keep the handler perhaps

nornagon:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export <-- I'm not able to find any reference for the `export =` syntax here, how does it work?

miniak:
@nornagon it's a TypeScript thing https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require. default exports are discouraged

MarshallOfSound:
@miniak I'm not a fan of `default` either but I'm even less of a fan of `export =`, it's effectively just a way to support legacy commonjs modules.  We should either be using `default` or named exports 🤷‍♂ `default` seems to be the easiest convention to stick with.  I would suggest the inverse of this PR, just make all our APIs export on `default`

miniak:
@MarshallOfSound makes sense, I will do the opposite then.

PR #18581, opened 2019-06-02
refactor: use ipcRenderer.invoke / ipcMain.handle in default-app

miniak:
#### Description of Change
Follow up to #18449

/cc @nornagon

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/default_app/preload.ts
+++ b/default_app/preload.ts
@@ -1,7 +1,7 @@
 import { ipcRenderer } from 'electron'
 
-function initialize () {
-  const electronPath = ipcRenderer.sendSync('bootstrap')
+async function initialize () {
+  const electronPath = await ipcRenderer.invoke('bootstrap')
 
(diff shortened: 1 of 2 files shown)


default_app/preload.ts
```diff
@@ -1,7 +1,9 @@
 import { ipcRenderer } from 'electron'
 
-function initialize () {
-  const electronPath = ipcRenderer.sendSync('bootstrap')
+const bootstrap = ipcRenderer.invoke('bootstrap')
```
MarshallOfSound: Why define it out here, seems like a strange pattern compared to just

```js
const electronPath = await ipcRenderer.invoke('bootstrap')
```

miniak: the idea was that the promise may already be resolved by the time `DOMContentLoaded` is fired

PR #18582, opened 2019-06-02
refactor: remove menu-will-close / setTimeout workaround

miniak:
#### Description of Change
Change the `ELECTRON_INSPECTOR_CONTEXT_MENU` handler to not rely on a timeout.
The timeout was added when reimplementing the inspector APIs to not use the remote module. 
This PR changes the implementation to something more straightforward so that a three-line explanation comment is no longer necessary.

We want the callback handler (ie, calling `resolve()`) to happen when the menu closed. The old impl used a timer to do this after `menu-will-close` fired; the new impl uses the closed callback from the [`Menu.popup() options`](https://electronjs.org/docs/api/menu#menupopupoptions) instead.

See discussion at https://github.com/electron/electron/pull/16607#pullrequestreview-200797896 (and followup discussion at https://github.com/electron/electron/pull/16794#pullrequestreview-200834486) for more details.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/devtools.js
+++ b/lib/browser/devtools.js
@@ -69,14 +69,7 @@ ipcMainUtils.handle('ELECTRON_INSPECTOR_CONTEXT_MENU', function (event, items, i
     const menu = Menu.buildFromTemplate(template)
     const window = event.sender.getOwnerBrowserWindow()
 
-    menu.once('menu-will-close', () => {
-      // menu-will-close is emitted before the click handler, but needs to be sent after.
-      // otherwise, DevToolsAPI.contextMenuCleared() would be called before
-      // DevToolsAPI.contextMenuItemSelected(id) and the menu will not work properly.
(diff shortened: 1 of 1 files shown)


codebytere:
@miniak could you please write something more descriptive than `Follow up to #16794`? That makes it hard for me to review this without having to dig through old issues to figure out context and need for this PR.

ckerr:
> What changed to make this workaround no longer necessary?

We want the callback handler (ie, calling `resolve()`) to happen when the menu closed. The old impl used a timer to do this after `will-close` fired; the new impl uses the closed callback from the [`Menu.popup() options`](https://electronjs.org/docs/api/menu#menupopupoptions) instead.

miniak:
@ckerr thanks for the suggestions. I've created the PR in a rush, should have kept it a draft until a proper PR description was added.

PR #18583, opened 2019-06-03
feat: add basic edit (cut/copy/paste) menu to "browser" mode in default app

miniak:
#### Description of Change
Add simple context menu to the default app with basic editing commands (cut / copy / paste) for editable elements + always visible inspect item.

Reason: when you open a webpage with electron via the default-app, there is no context menu at all without this change on `<input type="text">` or `<textarea>`

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added basic edit (cut/copy/paste) menu to "browser" mode in default app.

### DIFF
--- a/default_app/default_app.ts
+++ b/default_app/default_app.ts
@@ -1,4 +1,4 @@
-import { app, dialog, BrowserWindow, shell, ipcMain } from 'electron'
+import { app, dialog, BrowserWindow, Menu, shell, ipcMain } from 'electron'
 import * as path from 'path'
 
 let mainWindow: BrowserWindow | null = null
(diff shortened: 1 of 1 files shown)


miniak:
@ckerr this is not mean to be a demo, it's just adding basic functionality as the default-app is kind of a mini browser when you run `electron https://www.example.com/`

MarshallOfSound:
I don't believe this should live in the default_app.  It just makes the difference between the default app and the packaged app bigger when we've done such a good job recently of making the differences smaller and making the experience consistent

miniak:
@MarshallOfSound this only affects the "browser" mode when you pass a URL to the electron binary, it does not do anything when you run an app via the default_app.

ckerr:
I wouldn't block this PR if a number of other maintainers support it, but I kind of don't like the idea of making Electron more like a full-fledged browser.

miniak:
@MarshallOfSound, @codebytere thoughts?

MarshallOfSound:
@miniak https://github.com/electron/electron/pull/18583#issuecomment-499554647 and I agree with the thoughts of @ckerr that we shouldn't be making the default_app run like a browser.  We should be very clear that that _is not_ the intended use here 👍

miniak:
fair enough, closing

PR #18591, opened 2019-06-03
build: bring in a later compiler for Windows on Arm

richard-townsend-arm:
#### Description of Change

Due to a code-gen defect affecting virtual method calls in the official Clang used in the M76 branch, it's necessary to use a slightly newer compiler for Windows. [The issue has been fixed upstream](https://bugs.llvm.org/show_bug.cgi?id=41997), but for various reasons it was not possible to update the compiler to a version that worked on all devices in M76. This is only intended to be a temporary workaround and will be removed when upstream Chromium updates to a Clang version which fixes the problem.

#### Checklist

#### Release Notes

Notes: Added experimental support for building for Windows on Arm

### DIFF
--- a/docs/development/build-instructions-gn.md
+++ b/docs/development/build-instructions-gn.md
@@ -194,15 +194,38 @@ $ gn gen out/Debug-x86 --args='... target_cpu = "x86"'
 ```
 
 Not all combinations of source and target CPU/OS are supported by Chromium.
-Only cross-compiling Windows 32-bit from Windows 64-bit and Linux 32-bit from
-Linux 64-bit have been tested in Electron. If you test other combinations and
-find them to work, please update this document :)
+
+<table>
+<tr><th>Host</th><th>Target</th><th>Status</th></tr>
+<tr><td>Windows x64</td><td>Windows arm64</td><td>Experimental</td>
+<tr><td>Windows x64</td><td>Windows x86</td><td>Automatically tested</td></tr>
+<tr><td>Linux x64</td><td>Linux x86</td><td>Automatically tested</td></tr>
+</table>
+
+If you test other combinations and find them to work, please update this document :)
 
 See the GN reference for allowable values of [`target_os`][target_os values]
-and [`target_cpu`][target_cpu values]
+and [`target_cpu`][target_cpu values].
 
 [target_os values]: https://gn.googlesource.com/gn/+/master/docs/reference.md#built_in-predefined-variables-target_os_the-desired-operating-system-for-the-build-possible-values
 [target_cpu values]: https://gn.googlesource.com/gn/+/master/docs/reference.md#built_in-predefined-variables-target_cpu_the-desired-cpu-architecture-for-the-build-possible-values
+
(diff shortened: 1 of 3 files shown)


richard-townsend-arm:
@jkleinsc @MarshallOfSound

docs/development/build-instructions-gn.md
```diff
@@ -194,16 +194,39 @@ $ gn gen out/Debug-x86 --args='... target_cpu = "x86"'
...
 
 See the GN reference for allowable values of [`target_os`][target_os values]
-and [`target_cpu`][target_cpu values]
+and [`target_cpu`][target_cpu values].
 
 [target_os values]: https://gn.googlesource.com/gn/+/master/docs/reference.md#built_in-predefined-variables-target_os_the-desired-operating-system-for-the-build-possible-values
 [target_cpu values]: https://gn.googlesource.com/gn/+/master/docs/reference.md#built_in-predefined-variables-target_cpu_the-desired-cpu-architecture-for-the-build-possible-values
 
+#### Windows on Arm (experimental)
+To cross-compile for Windows on Arm, [follow Chromium's guide](https://chromium.googlesource.com/chromium/src/+/refs/heads/master/docs/windows_build_instructions.md#Visual-Studio) to get the necessary dependencies, SDK and libraries, then build with `ELECTRON_BUILDING_WOA=1` in your environment before running `gclient sync`.
```
felixrieseberg: Nit: Newline after title

```suggestion

To cross-compile for Windows on Arm, [follow Chromium's guide](https://chromium.googlesource.com/chromium/src/+/refs/heads/master/docs/windows_build_instructions.md#Visual-Studio) to get the necessary dependencies, SDK and libraries, then build with `ELECTRON_BUILDING_WOA=1` in your environment before running `gclient sync`.
```

jkleinsc:
WOA build with this change here: https://ci.appveyor.com/project/electron-bot/electron-ldhmv/builds/25013835

jkleinsc:
Merging as failing mac test is unrelated to PR and known to not work for fork PRs.

patches/common/chromium/woa_compiler_workaround.patch
```diff
@@ -0,0 +1,37 @@
+From b3414d055399d0a21f6166a536467ea752b2aa8a Mon Sep 17 00:00:00 2001
```
MarshallOfSound: Just FYI @jkleinsc @felixrieseberg @richard-townsend-arm this patch file is invalid.  Although it works it wasn't generated by the `export-patches` script which means that the next time someone runs import/export on their patch directory this file will change.  E.g. The `From` line here should be `000000000...` and the git version `2.19.1.windows.1` shouldn't be present.

`./electron/script/git-export-patches -o ./electron/patches/common/chromium` is the command to export a patch applied on top of chromium 👍

richard-townsend-arm: Sounds like the kind of thing a presubmit hook should check for :D

PR #18607, opened 2019-06-03
build: unify YARN_VERSION variable usage and ensure CI uses yarn not npm

MarshallOfSound:
De-duplicates `yarn_version` so we don't have to maintain two variables.
Ensures that we don't call `npm install` on CI rather `npm run yarn install` so that CI is deterministic

Notes: no-notes


package.json
```diff
@@ -85,6 +85,7 @@
     "start": "node ./script/start.js",
     "test": "node ./script/spec-runner.js",
     "tsc": "tsc",
+    "yarn": "npx yarn@$(node -p 'require(\"./script/lib/utils\").YARN_VERSION')",
```
nornagon: does this work on windows?

MarshallOfSound: Doesn't need to 😄 We only use it on MacOS + Linux.  Appveyor doesn't need to call `yarn` manually anywhere.

Would be nice if it worked on windows too though, that would require a wrapping JS script 🤷‍♂ I can do that if you want?

PR #18610, opened 2019-06-03
spec: move more BrowserWindow specs to main runner

nornagon:
#### Description of Change
Another 500 lines or so.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes


codebytere:
@nornagon looks like
```
not ok 159 BrowserWindow module sizing BrowserWindow.setContentSize(width, height) sets the content size
  AssertionError: expected [ 456, 729 ] to deeply equal [ 456, 789 ]
      at Context.<anonymous> (electron\spec-main\api-browser-window-spec.ts:728:31)
      at processImmediate (internal/timers.js:443:21)
not ok 160 BrowserWindow module sizing BrowserWindow.setContentSize(width, height) works for a frameless window
  AssertionError: expected [ 456, 788 ] to deeply equal [ 456, 789 ]
      at Context.<anonymous> (electron\spec-main\api-browser-window-spec.ts:741:31)
      at processImmediate (internal/timers.js:443:21)
```

is dying across windows platforms

nornagon:
yep i see that. haven't had the chance to debug yet.

nornagon:
Weirdly a bunch more tests failed on my local windows machine than failed on CI, because of off-by-one errors, e.g. setting minimum size to (100, 100) and then reading it back resulted in reading (101, 101). Is this a known issue? This PR wouldn't create that kind of behaviour, I don't think, but I'm not sure why CI isn't seeing that issue.

zcbenz:
@nornagon It is expected if you are using high DPI display, Chromium uses integers for window geometry and off-by-one errors happen very frequently. CI does not use high DPI display so it does not have the problem.

nornagon:
one other weird thing: I don't see the "off-by-60" issue on my windows VM, but I see a different "off-by-1" error (which might be related to what @zcbenz mentioned about DPI)

zcbenz:
> Ok I'm not sure how this test is consistently getting an off-by-60 error, but I'm not sure the old test was correct either. The old test opened the window with `width: 400, height: 400` and then called `setContentSize(400, 400)`, which I'm not sure would actually change the size of the window? the new test keeps the initial size of `400 x 400` but resizes to `456 x 789`, and ends up with a window of size `456 x 729`.

The size passed to window includes the size of frame, while `setContentSize` only sets the size of client area.

Resizing client area to `789` probably makes the window larger than the screen of CI and Windows automatically subtracts sizes of task bar and window frame from it.

nornagon:
Oh wow, the screen is smaller than 800px tall? ok, i'll try a smaller size. thanks Cheng :)

PR #18613, opened 2019-06-04
docs: ensure that optionality matches between documented params and signatures

MarshallOfSound:
This pulls in a new docs parser version that includes some enforcements to ensure that the optionality we define with `(optional)` also matches the optionality in the signature indicated with `(a[, optionalB])`.

Using that enforcement I then fixed a bunch of issues 👍 

Notes: no-notes


docs/api/browser-window.md
```diff
@@ -864,7 +864,7 @@ Returns `Boolean` - Whether the window is in normal state (not maximized, not mi
 
 * `aspectRatio` Float - The aspect ratio to maintain for some portion of the
 content view.
-* `extraSize` [Size](structures/size.md) - The extra size not to be included while
+* `extraSize` [Size](structures/size.md) (Optional) - The extra size not to be included while
```
nornagon: any particular reason not to be consistent re: capitalization between `(Optional)` and `(optional)`?

MarshallOfSound: Fair, enforcing lower case now 👍

codebytere:
@MarshallOfSound does this still need to get backported?

MarshallOfSound:
@codebytere Entirely optional 👍

PR #18615, opened 2019-06-04
build: add basic linting for the patches folder to ensure that .patches match the state on disk

MarshallOfSound:
As raised in Slack, checking that the `.patches` file is consistent with what is actually on disk is a Good Thing

Notes: no-notes

### DIFF
--- a/script/lint.js
+++ b/script/lint.js
@@ -133,12 +133,60 @@ const LINTERS = [ {
       process.exit(1)
     }
   }
+}, {
+  key: 'patches',
+  roots: ['patches'],
+  test: () => true,
+  run: () => {
+    const patchesDir = path.resolve(__dirname, '../patches')
+    for (const patchTarget of fs.readdirSync(patchesDir)) {
+      const targetDir = path.resolve(patchesDir, patchTarget)
+      // If the config does not exist that is OK, we just skip this dir
+      const targetConfig = path.resolve(targetDir, 'config.json')
+      if (!fs.existsSync(targetConfig)) continue
+
+      const config = JSON.parse(fs.readFileSync(targetConfig, 'utf8'))
+      for (const key of Object.keys(config)) {
+        // The directory the config points to should exist
+        const targetPatchesDir = path.resolve(__dirname, '../../..', key)
+        if (!fs.existsSync(targetPatchesDir)) throw new Error(`target patch directory: "${targetPatchesDir}" does not exist`)
+        // We need a .patches file
+        const dotPatchesPath = path.resolve(targetPatchesDir, '.patches')
+        if (!fs.existsSync(dotPatchesPath)) throw new Error(`.patches file: "${dotPatchesPath}" does not exist`)
+
+        // Read the patch list
+        const patchFileList = fs.readFileSync(dotPatchesPath, 'utf8').trim().split('\n')
+        const patchFileSet = new Set(patchFileList)
+        patchFileList.reduce((seen, file) => {
+          if (seen.has(file)) {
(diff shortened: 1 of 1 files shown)


script/lint.js
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+const { expect } = require('chai')
```
nornagon: using `expect` seems like overkill here, just use `if (...) throw`

script/lint.js
```diff
@@ -136,12 +137,54 @@ const LINTERS = [ {
...
+      if (!fs.existsSync(targetConfig)) continue
+
+      const config = JSON.parse(fs.readFileSync(targetConfig, 'utf8'))
+      for (const key of Object.keys(config)) {
+        // The directory the config points to should exist
+        const targetPatchesDir = path.resolve(__dirname, '../../..', key)
+        expect(fs.existsSync(targetPatchesDir)).to.equal(true, `target patch directory: "${targetPatchesDir}" should exist`)
+        // We need a .patches file
+        const dotPatchesPath = path.resolve(targetPatchesDir, '.patches')
+        expect(fs.existsSync(dotPatchesPath)).to.equal(true, `.patches file: "${dotPatchesPath}" should exist`)
```
nornagon: super minor nit: i think these paths will be printed relative to this script? ideally they'd be relative to either `src/` or `src/electron`

script/lint.js
```diff
@@ -136,12 +137,54 @@ const LINTERS = [ {
...
+
+        // Read the patch list
+        const patchFileList = fs.readFileSync(dotPatchesPath, 'utf8').trim().split('\n')
+        const patchFileSet = new Set(patchFileList)
+        expect(patchFileList).to.have.length(patchFileSet.size, 'each patch file should only be in the .patches file once')
```
nornagon: Should list the name that's shown twice, e.g.
```suggestion
        const duplicates = (list) => {
          const count = new Map
          for (const el of list) {
            count.set(el, (count.get(el) || 0) + 1)
          }
          return [...count.entries()].filter(([k, v]) => v > 1).map(([k, v]) => k)
        }
        const duplicatedPatches = duplicates(patchFileList)
        if (duplicatedPatches.length > 0) {
          throw new Error(`Patches appear in the list more than once: ${duplicatedPatches}`)
        }
```

or

```suggestion
        patchFileList.reduce((seen, file) => {
          if (seen.has(file)) {
            throw new Error(`'${file}' is listed in ${dotPatchesPath} more than once`)
          }
          return seen.add(file)
        }, new Set)
```

script/lint.js
```diff
@@ -136,12 +137,54 @@ const LINTERS = [ {
...
+        for (const file of fs.readdirSync(targetPatchesDir)) {
+          // Ignore the .patches file and READMEs
+          if (file === '.patches' || file === 'README.md') continue
```
nornagon: maybe only look at `.patch` files? `git-export-patches` should only ever produce files ending in `.patch`.

script/lint.js
MarshallOfSound: These paths will be printed as absolute paths currently

script/lint.js
MarshallOfSound: Yeah, which is why I wanted to yell at people if a random file ended up in that directory (it should never be there)

nornagon: Hm, idk. Extra files don't hurt; `git-export-patch` only cares about `.patch` files and the `.patches` list. I can imagine it being fine to have other files in there? But I suppose it doesn't really matter. Up to you.

zcbenz:
The branch should be rebased to fix the CI failures.

PR #18628, opened 2019-06-04
chore: remove mips64el patches as they've largely been upstreamed

nornagon:
#### Description of Change
This also collapses the patches/ directory structure to remove the `common` component.

Having multiple separate patch sets per-architecture (or per-anything) is a misfeature because it means the upgrade process has to be carried out once per architecture to ensure that all patches are correctly rebased. Instead, we should aim to have patches that are applied unconditionally, and any architecture-specific logic should be guarded with `#ifdef` or similar.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes


MarshallOfSound:
Weeooowww 
image

PR #18630, opened 2019-06-04
fix: ensure correct ordering of sendSync w.r.t. send

nornagon:
#### Description of Change
#18449 introduced a potential race between `send` and `sendSync` by pushing `sendSync` calls to a separate mojo pipe, which is the cause of the recent test flake in `remote module` tests. This change rearranges the calls so `send` and family all happen on the same pipe, and only `invoke` happens on a different pipe, which it must do because it has a callback with a result (which can't be pending when `sendSync` is called, or else DCHECK). There are no ordering concerns currently w/ `invoke` because we don't really use it, but in future we probably want to guarantee that `send` and `invoke` share an ordering, which will require some further trickery.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/atom/renderer/api/atom_api_renderer_ipc.cc
+++ b/atom/renderer/api/atom_api_renderer_ipc.cc
@@ -43,7 +43,7 @@ class IPCRenderer : public mate::Wrappable<IPCRenderer> {
     render_frame->GetRemoteInterfaces()->GetInterface(
         mojo::MakeRequest(&electron_browser_ptr_));
     render_frame->GetRemoteInterfaces()->GetInterface(
-        mojo::MakeRequest(&electron_browser_sync_ptr_));
+        mojo::MakeRequest(&electron_browser_async_ptr_));
   }
   static void BuildPrototype(v8::Isolate* isolate,
                              v8::Local<v8::FunctionTemplate> prototype) {
@@ -71,7 +71,7 @@ class IPCRenderer : public mate::Wrappable<IPCRenderer> {
                                 const base::Value& arguments) {
     atom::util::Promise p(args->isolate());
     auto handle = p.GetHandle();
-    electron_browser_ptr_->Invoke(
+    electron_browser_async_ptr_->Invoke(
         channel, arguments.Clone(),
         base::BindOnce(
             [](atom::util::Promise p, base::Value value) { p.Resolve(value); },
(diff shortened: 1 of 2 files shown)


atom/renderer/api/atom_api_renderer_ipc.cc
```diff
@@ -55,49 +61,47 @@ class IPCRenderer : public mate::Wrappable<IPCRenderer> {
...
     return mate::CreateHandle(isolate, new IPCRenderer(isolate));
   }
 
-  void Send(mate::Arguments* args,
-            bool internal,
+  void Send(bool internal,
             const std::string& channel,
             const base::ListValue& arguments) {
-    electron_browser_ptr_->Message(internal, channel, arguments.Clone());
+    electron_browser_ptr_->get()->Message(internal, channel, arguments.Clone());
```
deepak1556: Now that the interface ptr is created on the background runner, which is a different sequence to that of the main runner. Is it okay to call methods on this ptr in the main sequence without calling `PassInterface` ? well that would technically create a new pipe which would defeat the purpose. Just to make sure is this a safe operation ?

nornagon: The type of this pointer changed to `ThreadSafeElectronBrowserPtr`, whose _raison d'etre_ is to allow calling methods from multiple threads. This doesn't create any new pipes.

nornagon:
I'm merging this ahead of review because it fixes a very annoying flake on master, but I'm happy to incorporate feedback from @zcbenz or anything further from @deepak1556 :)

atom/renderer/api/atom_api_renderer_ipc.cc
deepak1556: Ah okay it uses `mojo/public/cpp/bindings/thread_safe_interface_ptr.h` , wasn't aware of this helper.

nornagon: I wasn't either until I asked in #mojo on the Chromium Slack :)

zcbenz:
Not sure if it is related, but after rebasing to this change, I'm constantly seeing "Object has been destroyed" error.

Steps:
1. Checkout https://github.com/electron/electron/pull/18483
2. Run `test -g crashReporter`
3. Switch to another window and then switch back

Screen Shot 2019-06-05 at 17 15 28

I have not tested the master branch yet.

miniak:
@nornagon should it be backported to 6-0-x?

PR #18631, opened 2019-06-04
feat: promisify systemPreferences notification methods

codebytere:
#### Description of Change

Converts the following subscription notification methods on `systemPreferences` to return Promises instead of taking callbacks:

- `systemPreferences.subscribeNotification(event, callback)`
- `systemPreferences.subscribeLocalNotification(event, callback)`
- `systemPreferences.subscribeWorkspaceNotification(event, callback)`

I don't believe these were initial candidates during the promisification effort owing to some other mitigating factors which have now been resolved. These changes also require changing the return value of the function as well as the format of the resolves values (two callback values and an integer return value as opposed to a single object containing all three) so as such i'm marking it semver-major and deprecating it in documentation. 

cc @miniak @zcbenz @nornagon 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Converted macOS subscription notification methods on `systemPreferences` to return Promises instead of taking callbacks.


docs/api/system-preferences.md
```diff
@@ -121,6 +143,20 @@ Returns `Number` - The ID of this subscription
...
+**[Deprecated](modernization/promisification.md)**
+
+### `systemPreferences.subscribeLocalNotification(event)` _macOS_
+
+* `event` String
+
+Returns `Promise<Object>` - Resolves with an object containing the following items:
+  * `id` Number - The ID of this subscription
+  * `event` String
+  * `userInfo` Object
```
MarshallOfSound: Can we type this as `unknown` or if we know the type maybe a `Record<A, B>`?

docs/api/system-preferences.md
```diff
@@ -131,21 +167,35 @@ This is necessary for events such as `NSUserDefaultsDidChangeNotification`.
...
+**[Deprecated](modernization/promisification.md)**
+
+### `systemPreferences.subscribeWorkspaceNotification(event)` _macOS_
+
+* `event` String
+
+Returns `Promise<Object>` - Resolves with an object containing the following items:
+  * `id` Number - The ID of this subscription
+  * `event` String
+  * `userInfo` Object
```
MarshallOfSound: Same here

docs/api/system-preferences.md
codebytere: `userInfo`, you mean?

MarshallOfSound: Yeah

miniak:
@codebytere this is wrong, the reason why this was not listed for promisification was that the listener can be invoke repeatedly, see #15951.

miniak:
@codebytere this is also missing
```
deprecate.promisify(SystemPreferences.prototype.unsubscribeNotification)
deprecate.promisify(SystemPreferences.prototype.unsubscribeLocalNotification)
deprecate.promisify(SystemPreferences.prototype.unsubscribeWorkspaceNotification)
```

miniak:
This used to work before properly, every time you toggle between Light and Dark appearance in settings, you get a log line
```
systemPreferences.subscribeNotification('AppleInterfaceThemeChangedNotification', () => {
    console.log('isDarkMode', systemPreferences.isDarkMode())
  }
)
```

zcbenz:
Yeah this should be reverted, I just realized `subscribeNotification` behaves like `addListener`.

codebytere:
Done: https://github.com/electron/electron/pull/18705.

Somehow I also didn't realize this! Thanks for catching it :)

PR #18634, opened 2019-06-04
chore: DirectoryLister memory management improvement

codebytere:
#### Description of Change

Minor memory cleanups in `web_dialog_helper`.

This PR changes a range-based loop to use const refs instead of copies, ensures that `FileSelectHelper` is destroyed when WebContents is destroyed, and removes unneeded `atom::DirectoryListerHelper`.

cc @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/web_dialog_helper.cc
+++ b/shell/browser/web_dialog_helper.cc
@@ -36,7 +36,7 @@ namespace {
 
 class FileSelectHelper : public base::RefCounted<FileSelectHelper>,
                          public content::WebContentsObserver,
-                         public electron::DirectoryListerHelperDelegate {
+                         public net::DirectoryLister::DirectoryListerDelegate {
  public:
   FileSelectHelper(content::RenderFrameHost* render_frame_host,
                    std::unique_ptr<content::FileSelectListener> listener,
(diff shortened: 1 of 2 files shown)


atom/browser/web_dialog_helper.cc
```diff
@@ -182,7 +183,10 @@ class FileSelectHelper : public base::RefCounted<FileSelectHelper>,
   }
 
   // content::WebContentsObserver:
-  void WebContentsDestroyed() override { render_frame_host_ = nullptr; }
+  void WebContentsDestroyed() override {
+    render_frame_host_ = nullptr;
+    Release();
```
MarshallOfSound: If I recall correctly from when I was messing with this calling `Release()` too many times causes a `CHECK` to be fired and I tested it extensively last time to ensure that it was being released down to a ref count of 0 after each open dialog.

What change here requires that we call `Release()` again now?

atom/browser/web_dialog_helper.cc
```diff
@@ -117,9 +117,10 @@ class FileSelectHelper : public base::RefCounted<FileSelectHelper>,
           base_dir = paths[0];
 
           // Actually enumerate soemwhere off-thread
-          base::SequencedTaskRunnerHandle::Get()->PostTask(
-              FROM_HERE, base::BindOnce(&FileSelectHelper::EnumerateDirectory,
-                                        this, base_dir));
+          base::PostTaskWithTraits(
+              FROM_HERE, {content::BrowserThread::UI},
```
deepak1556: The original intent was to post the task to a different sequence, `content::BrowserThread::UI` should be avoided.

atom/browser/web_dialog_helper.cc
codebytere: ah i should have included this in the PR body, but this is just application of the review here: https://github.com/electron/electron/pull/18343#pullrequestreview-240192999

deepak1556: `FileSelectHelper` is now handed off to different sequence for directory listing by `net::DirectoryLister`, we don't want to continue the operation when the webcontents is destroyed, this is just to terminate the background operation.

atom/browser/web_dialog_helper.cc
```diff
@@ -82,13 +82,15 @@ class FileSelectHelper : public base::RefCounted<FileSelectHelper>,
  private:
   friend class base::RefCounted<FileSelectHelper>;
 
+  std::unique_ptr<net::DirectoryLister> lister_;
```
deepak1556: nit: move this declaration alongside other member variables below.

atom/browser/web_dialog_helper.cc
```diff
@@ -82,13 +82,15 @@ class FileSelectHelper : public base::RefCounted<FileSelectHelper>,
...
+
   ~FileSelectHelper() override {}
 
   void EnumerateDirectory(base::FilePath base_dir) {
-    auto* lister = new net::DirectoryLister(
+    lister_.reset(new net::DirectoryLister(
         base_dir, net::DirectoryLister::NO_SORT_RECURSIVE,
-        new atom::DirectoryListerHelper(base_dir, this));
-    lister->Start();
+        new atom::DirectoryListerHelper(base_dir, this)));
```
deepak1556: It would also be good if you can eliminate `atom::DirectoryListerHelper` , its an unneeded redirection for representing a delegate. `FileSelectHelper` can very well represent itself as the delegate.

codebytere:
@deepak1556 we need to keep this as
```
base::SequencedTaskRunnerHandle::Get()->PostTask(
FROM_HERE, base::BindOnce(&FileSelectHelper::EnumerateDirectory,
   this, base_dir));
```
or we hit the following CHECK:
```
Check failed: task_runner_->RunsTasksInCurrentSequence(). 
```

codebytere:
@deepak1556 I saw you requested an extra review from Charles but he and I paired on this  for context :)

deepak1556:
@codebytere yeah because I pushed some style changes, don't want to approve my own commit :)

codebytere:
@ckerr test failures are both flakes :)

PR #18635, opened 2019-06-05
fix: reject the executeJavaScript promise when it fails to execute the script

### LINKED ISSUE #9102: FR: Better error reporting for webContents->executeJavaScript
* Electron version: 1.6.5
* Operating system: OSX

### Expected behavior

**Possible solutions**

- Set a more useful ```filename``` in the throwed error on the window object for identify where it happened (usefull)

- The result/promise is never resolved now. Maybe it should get a callback with error argument and reject the promise exactly in this case?

**Documentation:**

The documentation requires a better overview where/how to catch errors. Additionally some other information to understand the working of the different contexts.

Something like:

| Location | Unhandled errors | Available modules*** | Exposed variables**** |
|--|--|--|--|
| preload.js | ? | ? | ? |
| executeJavaScript()* | ? | ? | ? |
| page scripts _5)_ | window error handler** | ? | ? |

*) webContents.executeJavaScript function
**) ```window.addEventListener('error', function(errorEvent) {})``` or ```window.onerror = function(...) {}```
***) which modules can be imported by using the ```require``` global function
****) In the global context (can be the window object)
_5)_ all scripts which is executed on the page or it resources

**Web preferences which alter things:**

| Web preference | default | How it alter things |
|--|--|--|
| sandbox | false | describe here... |
| contextIsolation | false | describe here... |
| nodeIntegration | true | describe here... |
| javascript | false | if false then: all javascript execution is prevented |
| webSecurity | false | if false then: describe here... |

### Actual behavior

If you use **executeJavaScript** it's difficult to catch the error which occurs in the passed javascript.

#### In renderer
```window.addEventListener('error', ...)``` is catching the error but the it's limited to the following:
```
column:   0
error:    null
filename: 
line:     0
message:  Script error
```
This changes if I set it to ```webSecurity``` to ```false``` then I get:
```
column:   7
error: 
filename: 
line:     1
message:  Uncaught Error: this is my exception
```
So the problem is now, webSecurity is **disabled**, I think this is the best thing to do. And the **filename** is even now empty, I even now have no clue where it happened.

#### Trough the debugger
I attached and registered the debugger with the domains: ```Debugger.enable``` and ```Log.enable```

```Log.entryAdded``` is never received for errors in ```executeJavaScript``` code.

### How to reproduce

use the options above then:
```JS
webContents.executeJavaScript('throw new Error("this is my exception")')
```
Same for syntax error or others...

In all examples the start options are (+ changes are described):
Start options: ```webSecurity: true, nodeIntegration: false, sandbox: false```
Yes, this is almost the default config except nodeIntegration.

This is not related to the new ```sandbox``` function, it's false - in sandbox mode it's broken entirely: #9073

MarshallOfSound:
Closes #9102

Unfortunately we can't get the _actual_ error message, but can at least reject the promise to avoid leaving it dangling.  The best we can do is tell them that their script failed, but not why.

Notes: `.executeJavaScript` will never leave a promise dangling now, scripts that fail to execute will correctly be rejected

### DIFF
--- a/atom/renderer/api/atom_api_web_frame.cc
+++ b/atom/renderer/api/atom_api_web_frame.cc
@@ -107,7 +107,8 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
         promise_.Resolve(result[0]);
       } else {
         promise_.RejectWithErrorMessage(
-            "Script failed to execute, this normally means a syncronous error "
+            "Script failed to execute, this normally means an error "
+            "was thrown. Check the renderer console for the error."
             "was thrown, check the renderer console for the error");
       }
     } else {
--- a/atom/renderer/api/atom_api_web_frame.cc
+++ b/atom/renderer/api/atom_api_web_frame.cc
@@ -114,7 +114,7 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
     } else {
       promise_.RejectWithErrorMessage(
           "WebFrame was removed before script could run. This normally means "
-          "the underlying frame was destroyed")
+          "the underlying frame was destroyed");
     }
     delete this;
   }
--- a/atom/renderer/api/atom_api_web_frame.cc
+++ b/atom/renderer/api/atom_api_web_frame.cc
@@ -101,9 +101,20 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
 
   void Completed(
       const blink::WebVector<v8::Local<v8::Value>>& result) override {
-    if (!result.empty() && !result[0].IsEmpty())
-      // Right now only single results per frame is supported.
-      promise_.Resolve(result[0]);
+    if (!result.empty()) {
+      if (!result[0].IsEmpty()) {
+        // Right now only single results per frame is supported.
+        promise_.Resolve(result[0]);
+      } else {
+        promise_.RejectWithErrorMessage(
+            "Script failed to execute, this normally means a syncronous error "
+            "was thrown, check the renderer console for the error");
+      }
+    } else {
+      promise_.RejectWithErrorMessage(
+          "WebFrame was removed before script could run, this normally means "
+          "the underlying frame was destroyed")
+    }
     delete this;
   }
 
--- a/atom/renderer/api/atom_api_web_frame.cc
+++ b/atom/renderer/api/atom_api_web_frame.cc
@@ -113,7 +113,7 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
       }
     } else {
       promise_.RejectWithErrorMessage(
-          "WebFrame was removed before script could run, this normally means "
+          "WebFrame was removed before script could run. This normally means "
           "the underlying frame was destroyed")
     }
     delete this;


atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -101,9 +101,20 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
-    if (!result.empty() && !result[0].IsEmpty())
-      // Right now only single results per frame is supported.
-      promise_.Resolve(result[0]);
+    if (!result.empty()) {
+      if (!result[0].IsEmpty()) {
+        // Right now only single results per frame is supported.
+        promise_.Resolve(result[0]);
+      } else {
+        promise_.RejectWithErrorMessage(
+            "Script failed to execute, this normally means a syncronous error "
```
nornagon:
```suggestion
            "Script failed to execute, this normally means an error "
            "was thrown. Check the renderer console for the error."
```

atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -101,9 +101,20 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
+            "was thrown, check the renderer console for the error");
+      }
+    } else {
+      promise_.RejectWithErrorMessage(
+          "WebFrame was removed before script could run, this normally means "
```
nornagon:
```suggestion
          "WebFrame was removed before script could run. This normally means "
```

nornagon:
It's a bit silly that we can't get the actual error message, because it's totally possible to catch any error and transmit it back. This seems like a flaw in the API exposed by Chromium. Is there perhaps a different API we should be using for this?

MarshallOfSound:
@nornagon It eventually ends up here --> https://cs.chromium.org/chromium/src/third_party/blink/renderer/bindings/core/v8/script_controller.cc?type=cs&sq=package:chromium&g=0&l=90-145

Where the script is evaluated inside a try catch 🤔

codebytere:
@MarshallOfSound should we backport this to 6 at least?

atom/renderer/api/atom_api_web_frame.cc
```diff
@@ -101,9 +101,21 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
-      promise_.Resolve(result[0]);
+    if (!result.empty()) {
+      if (!result[0].IsEmpty()) {
+        // Right now only single results per frame is supported.
+        promise_.Resolve(result[0]);
+      } else {
+        promise_.RejectWithErrorMessage(
+            "Script failed to execute, this normally means an error "
+            "was thrown. Check the renderer console for the error."
+            "was thrown, check the renderer console for the error");
```
poiru: This is here twice.

MarshallOfSound: How strange 😕 I definitely wouldn't type the same thing twice, the auto-formatter might have done this 🤔

nornagon: i think it's because you directly committed my suggestion in the github UI, which can't handle suggesting changing multiple lines

PR #18650, opened 2019-06-05
feat: sandbox renderer processes for cross-origin frames

miniak:
#### Description of Change
Cross-origin frames, which have a separate renderer process due to site isolation do not have their own `webContents` and therefore use the `webPreferences` of the main frame. When `nodeIntegrationInSubFrames` is not enabled, there is no reason not to sandbox the renderer process as we are not executing any internal JavaScript code or initializing node, which would be impacted by the sandbox.

Follow up to #15821

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Renderer processes hosting cross-origin frames are now sandboxed unless the parent `BrowserWindow` enables `nodeIntegrationInSubFrames`.

### DIFF
--- a/shell/app/atom_main_delegate.cc
+++ b/shell/app/atom_main_delegate.cc
@@ -59,6 +59,11 @@ bool IsBrowserProcess(base::CommandLine* cmd) {
   return process_type.empty();
 }
 
+bool IsSandboxEnabled(base::CommandLine* command_line) {
+  return command_line->HasSwitch(switches::kEnableSandbox) ||
+         !command_line->HasSwitch(service_manager::switches::kNoSandbox);
+}
+
 // Returns true if this subprocess type needs the ResourceBundle initialized
 // and resources loaded.
 bool SubprocessNeedsResourceBundle(const std::string& process_type) {
@@ -281,10 +286,9 @@ content::ContentGpuClient* AtomMainDelegate::CreateContentGpuClient() {
 
 content::ContentRendererClient*
 AtomMainDelegate::CreateContentRendererClient() {
-  if (base::CommandLine::ForCurrentProcess()->HasSwitch(
-          switches::kEnableSandbox) ||
-      !base::CommandLine::ForCurrentProcess()->HasSwitch(
-          service_manager::switches::kNoSandbox)) {
+  auto* command_line = base::CommandLine::ForCurrentProcess();
+
+  if (IsSandboxEnabled(command_line)) {
     renderer_client_.reset(new AtomSandboxedRendererClient);
   } else {
     renderer_client_.reset(new AtomRendererClient);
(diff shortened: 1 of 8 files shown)


miniak:
@MarshallOfSound cross-origin iframes, which have a separate renderer process due to site isolation do not have their own webContents and therefore use the webPreferences of the main frame. When `nodeIntegrationInSubFrames` is not enabled, there is no reason not to sandbox the renderer process as we are not executing any internal JavaScript code or initializing node, which be impacted by the sandbox.

> Why?
improved security, cross-origin iframes are sandboxed even if the main page is not
> What's the potential impact?
there should be no negative effects
> Why do we need to explicitly do this?
because there is no way to explicitly enable sandbox on iframes

atom/browser/atom_browser_client.cc
```diff
@@ -325,6 +325,10 @@ void AtomBrowserClient::ConsiderSiteInstanceForAffinity(
   }
 }
 
+bool AtomBrowserClient::isRendererSubFrame(int process_id) const {
```
MarshallOfSound:
```suggestion
bool AtomBrowserClient::IsRendererSubFrame(int process_id) const {
```

atom/browser/web_contents_preferences.cc
```diff
@@ -293,12 +294,15 @@ void WebContentsPreferences::AppendCommandLineSwitches(
   if (IsEnabled(options::kWebviewTag))
     command_line->AppendSwitch(switches::kWebviewTag);
 
+  bool sandbox =
```
MarshallOfSound: Can we make this more obvious what it means, it doesn't exactly map to `--enable-sandbox`

```suggestion
  bool can_sandbox_frame =
```

atom/browser/web_contents_preferences.cc
```diff
@@ -293,12 +294,15 @@ void WebContentsPreferences::AppendCommandLineSwitches(
...
+      is_subframe && !IsEnabled(options::kNodeIntegrationInSubFrames);
+
   // If the `sandbox` option was passed to the BrowserWindow's webPreferences,
   // pass `--enable-sandbox` to the renderer so it won't have any node.js
   // integration.
-  if (IsEnabled(options::kSandbox)) {
+  if (IsEnabled(options::kSandbox) || sandbox) {
```
MarshallOfSound:
```suggestion
  if (IsEnabled(options::kSandbox) || can_sandbox_frame) {
```

atom/browser/web_contents_preferences.cc
```diff
@@ -293,12 +294,15 @@ void WebContentsPreferences::AppendCommandLineSwitches(
...
     command_line->AppendSwitch(switches::kEnableSandbox);
-  } else if (!command_line->HasSwitch(switches::kEnableSandbox)) {
+  } else if (!command_line->HasSwitch(switches::kEnableSandbox) && !sandbox) {
```
MarshallOfSound: Could you perhaps add some comments around this logic, it took me a few reads to wrap my head around what we were doing here.  Would be great to understand at a glance

miniak: Actually, I think this condition should not be changed.

miniak:
@MarshallOfSound can you please check again?

PR #18653, opened 2019-06-05
build: show ninja stats for testing/debug builds

jkleinsc:
This PR adds ninja stats output for testing and debug builds.  This is primarily informational,  but could be useful in diagnosing where our [builds are slow](https://chromium.googlesource.com/chromium/src/+/master/docs/windows_build_instructions.md#Why-is-my-build-slow).  

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,7 +103,7 @@ env-disable-crash-reporter-tests: &env-disable-crash-reporter-tests
   DISABLE_CRASH_REPORTER_TESTS: true
 
 env-ninja-status: &env-ninja-status
-  NINJA_STATUS: "[%r processes, %f/%t @ %o/s : %es ]"
+  NINJA_STATUS: "[%r processes, %f/%t @ %o/s : %es]"
 
 # Individual (shared) steps.
 step-maybe-notify-slack-failure: &step-maybe-notify-slack-failure
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -542,8 +542,7 @@ step-ninja-summary: &step-ninja-summary
   run:
     name: Print ninja summary
     command: |
-      cd src
-      python third_party/depot_tools/post_build_ninja_summary.py -C out/Default
+      python depot_tools/post_build_ninja_summary.py -C src/out/Default
 
 # Lists of steps.
 steps-lint: &steps-lint
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,6 +102,9 @@ env-mac-large: &env-mac-large
 env-disable-crash-reporter-tests: &env-disable-crash-reporter-tests
   DISABLE_CRASH_REPORTER_TESTS: true
 
+env-ninja-status: &env-ninja-status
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -102,6 +102,9 @@ env-mac-large: &env-mac-large
 env-disable-crash-reporter-tests: &env-disable-crash-reporter-tests
   DISABLE_CRASH_REPORTER_TESTS: true
 
+env-ninja-status: &env-ninja-status
+  NINJA_STATUS: "[%r processes, %f/%t @ %o/s : %es ]"
```
nornagon:
```suggestion
  NINJA_STATUS: "[%r processes, %f/%t @ %o/s : %es]"
```

.circleci/config.yml
```diff
@@ -535,6 +538,13 @@ step-fix-known-hosts-linux: &step-fix-known-hosts-linux
         ./src/electron/.circleci/fix-known-hosts.sh
       fi
 
+step-ninja-summary: &step-ninja-summary
+  run:
+    name: Print ninja summary
+    command: |
+      cd src
+      python third_party/depot_tools/post_build_ninja_summary.py -C out/Default
```
nornagon: any particular reason to use the vendored depot_tools instead of the one that's installed on CI?

jkleinsc: I saw this comment and thought it was always being checked out to third_party: https://github.com/electron/electron/blob/f5339e15b13cab7b217c7c53f58259c333218333/.circleci/config.yml#L555

.circleci/config.yml
```diff
@@ -102,6 +102,9 @@ env-mac-large: &env-mac-large
 env-disable-crash-reporter-tests: &env-disable-crash-reporter-tests
   DISABLE_CRASH_REPORTER_TESTS: true
 
+env-ninja-status: &env-ninja-status
+  NINJA_STATUS: "[%r processes, %f/%t @ %o/s : %es]"
```
nornagon: hm, i see now that this stuff comes from autoninja, which also has that extra space there? I wonder if we should just invoke `autoninja`. either way i was probably wrong about removing the space.
```suggestion
  NINJA_STATUS: "[%r processes, %f/%t @ %o/s : %es ]"
```

jkleinsc: It works either way 😁

appveyor.yml
```diff
@@ -50,6 +50,7 @@ build_script:
       if(($env:APPVEYOR_PULL_REQUEST_HEAD_REPO_NAME -split "/")[0] -eq ($env:APPVEYOR_REPO_NAME -split "/")[0]) {
         Write-warning "Skipping PR build for branch"; Exit-AppveyorBuild
       }
+  - where gclient
```
MarshallOfSound: Is this needed or was it for debugging?

jkleinsc: debugging

MarshallOfSound:
As part of this can we add `out/{OUT_DIR}/.ninja_log` to the artifacts list.  It's really helpful for understanding performance of the build as well

Issue #18293, opened 2019-05-14
shell.openExternal with Firefox as default browser not always getting focus


linonetwo:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

Electron version: v5.0.1
Operating system: MaxOS 10.14.4

### Expected Behavior

As doc https://electronjs.org/docs/api/shell#shellopenexternalurl-options

### Actual Behavior

Link opened in the browser, but the browser is not brought to the foreground.

### To Reproduce

Upgrade electron to v5.0.1 in https://github.com/electron/electron-api-demos

And try "open external links" demo

MarshallOfSound:
@linonetwo In order to activate a the opened application you have to set `activate: true`

```js
shell.openExternal('https://google.com', {
  activate: true,
})
```

linonetwo:
But as doc https://electronjs.org/docs/api/shell#shellopenexternalurl-options
The default of `activate` is true on Mac.

linonetwo:
Actually `shell.openExternal(url, { activate: true })` don't work, please reopen this, might be v5 regression.

linonetwo:
@MarshallOfSound Please reopen this.

MarshallOfSound:
@linonetwo Closed issues take a bit longer to filter to the top of my queue, please be patient and do not raise duplicate issues

jeremyspiegel:
I am seeing the same problem with Safari as well as with Firefox. The problem does not exist when Chrome is the default browser. I am able to reproduce the issue with Electron 5.0.2, but not with 4.2.3. I also reproduced it with v5.0.0-beta.1.

jeremyspiegel:
I think the problem is in this [change](https://github.com/electron/electron/pull/15065/files#diff-3931a5874dc372d4b3db289b762c23c9R151). A reference to an `OpenExternalOptions` structure is being captured by an Objective-C block that outlives the object that's being referenced.

Issue #17649, opened 2019-04-02
macOS autoUpdater spec fails on PRs from forks


malept:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * Code from master
* **Operating System:**
  * macOS on Electron's CircleCI instances
* **Last Known Working Electron version:**:
  * Right before https://github.com/electron/electron/pull/17442 was merged

### Expected Behavior

The testsuite passes on CircleCI's `build-mac` workflow

### Actual Behavior

[1 test failure](https://circleci.com/gh/electron/electron/166983):

```
autoUpdater behavior "before each" hook for "should have a valid code signing identity" - "before each" hook for "should have a valid code signing identity"
/Users/distiller/project/src/electron/spec-main/api-autoupdater-darwin-spec.ts

Error: No valid signing identity available to run autoUpdater specs
    at Context.<anonymous> (electron/spec-main/api-autoupdater-darwin-spec.ts:26:15)
```

### To Reproduce

1. Fork `electron/electron` and make a branch
2. Upload branch to personal fork
3. Make a PR
4. Wait for CircleCI's `build-mac` workflow to run

### Screenshots

![image](https://user-images.githubusercontent.com/11417/55369778-9660f700-54bd-11e9-8041-98cfe87afe2a.png)

### Additional Information

CC: @MarshallOfSound as he told me to :smile:

PR #18657, opened 2019-06-05
fix: properly pass openExternal activate option

jeremyspiegel:
#### Description of Change
A reference to an `OpenExternalOptions` structure was being captured by an Objective-C block that outlived the object that was being referenced (added in https://github.com/electron/electron/pull/15065). This caused the default browser window to not be activated when passing a URL on macOS when the default browser is Safari or Firefox. Fixes https://github.com/electron/electron/issues/18293.

cc @miniak @nornagon @alexeykuzmin @codebytere 

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Fixed issue where `shell.openExternal` would not activate opened window on macOS

### DIFF
--- a/atom/common/platform_util_mac.mm
+++ b/atom/common/platform_util_mac.mm
@@ -104,14 +104,15 @@ void OpenExternal(const GURL& url,
     return;
   }
 
+  bool activate = options.activate;
   __block OpenExternalCallback c = std::move(callback);
-  dispatch_async(
-      dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        __block std::string error = OpenURL(ns_url, options.activate);
-        dispatch_async(dispatch_get_main_queue(), ^{
-          std::move(c).Run(error);
-        });
-      });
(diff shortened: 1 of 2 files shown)


jeremyspiegel:
Looks like I can't rely on a window onblur event in a test on CI (it worked when I ran it via `npm run test`). Not sure how else to test this, so maybe it'd be best to just revert the test change I made.

jeremyspiegel:
Hmm, I see now that with https://github.com/electron/electron/issues/17649, osx-testing-test doesn't pass on PRs from forks. I don't have permission to push a branch to this repo though.

MarshallOfSound:
@jeremyspiegel That's OK, as that's the only failure this PR is effectively green.  Just needs a review 👍

codebytere:
The failing macOS spec is expected for forks; merging.

codebytere:
@jeremyspiegel could you please open manual backports for the targeted release lines?

jeremyspiegel:
@codebytere sure, but the bug is not in 4.2.x, it is new in 5.0.x.

codebytere:
@jeremyspiegel oops, corrected!

Issue #18410, opened 2019-05-22
Main window flickers with maximizing and fullscreen


katiamoklokova:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  *  Electron 4.1.3, 4.2.1, 5.0.0-beta6 and 6.0.0-beta2
* **Operating System:**
  *  macOS 10.14.4
* **Last Known Working Electron version:**
  *  3.1.7

### Expected Behavior

Maximizing/restoring and toggling/exiting fullscreen shouldn't cause the main window to flicker.

### Actual Behavior

Main window flickers when the app is maximized/restored or if fullscreen is toggled/exited.

### To Reproduce

1. Launch an Electron app
2. Maximize and restore the main window
3. Toggle and exit fullscreen

### Screenshots

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/20443967/58202821-6ba94880-7c8d-11e9-8950-fd7491feee57.gif)

codebytere:
@katiamoklokova do you by chance have a more minimal repro? I would hedge at first glance that this is resultant of a specific combination of `titleBar` and `BrowserWindow` options, so it would definitely help if we had a better idea of which ones are in play here!

MarshallOfSound:
@codebytere I can repro on the default app.  The flash happens whenever you transition between fullscreen and normal

MarshallOfSound:
Fixed in #18661

PR #18661, opened 2019-06-05
fix: delay handling occlusion events to avoid flicker on macOS

### LINKED ISSUE #17942: v5.0.0: white flicker during Mission Control transitions and workspace transitions
### Issue Details

* **Electron Version:**
  * all beta versions in 5.0.x
* **Operating System:**
  * macOS 10.13 and 10.14
  * 4.1.4

### Expected Behavior
During Mission Control transitions or workspace transitions, the window should not flicker. See the right window in the screenshot
### Actual Behavior
During Mission Control transitions or workspace transitions, the whole window flickers white. See the left window in the screenshot

### Screenshots
left version is 5.0.0-beta.9, right version is 4.1.4
image

see more in [this video](https://youtu.be/1ERkxTIKk6E)

MarshallOfSound:
Fixes #17942 and #18410

The issue here is caused by a refactor in Chromium to listen to these occlusion events at at this remove view level.  macOS issues two events very quickly during fullscreen transitions.  The first says "occluded" the next says "visible".  When in fact the window is never occluded.

This adds a patch that I hope to upstream to delay handling of the "occluded" event for a few seconds to ensure that a follow up "visible" event isn't incoming.

The second commit here is just patch hash updates.

Notes: Fixed random flicker that occurred on macOS when performing fullscreen or workspace transitions

### DIFF
--- a/patches/common/chromium/chore_add_debounce_on_the_updatewebcontentsvisibility_method_to.patch
+++ b/patches/common/chromium/chore_add_debounce_on_the_updatewebcontentsvisibility_method_to.patch
@@ -1,0 +1,66 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <sattard@slack-corp.com>
+Date: Wed, 5 Jun 2019 15:11:00 -0700
+Subject: chore: add debounce on the updateWebContentsVisibility method to
+ ensure quick changes in occlusion do not result in flickering
+
+
+diff --git a/content/app_shim_remote_cocoa/web_contents_view_cocoa.h b/content/app_shim_remote_cocoa/web_contents_view_cocoa.h
+index 230e259f6017310e556d11dec43973b74015f191..ebd19fc31efef50677da417dfd69f0c8cce6c682 100644
+--- a/content/app_shim_remote_cocoa/web_contents_view_cocoa.h
++++ b/content/app_shim_remote_cocoa/web_contents_view_cocoa.h
+@@ -58,6 +58,8 @@ CONTENT_EXPORT
+                        offset:(NSPoint)offset;
+ - (void)clearViewsHostableView;
+ - (void)updateWebContentsVisibility;
++- (remote_cocoa::mojom::Visibility)currentVisibility;
++- (void)notifyWebContentsVisibilityChanged;
+ - (void)viewDidBecomeFirstResponder:(NSNotification*)notification;
+ @end
+ 
+diff --git a/content/app_shim_remote_cocoa/web_contents_view_cocoa.mm b/content/app_shim_remote_cocoa/web_contents_view_cocoa.mm
+index 615fe671d415747cb84f673327629d3dc771039e..6c31ba5b9149161784f5813598ddcf30e2d8af2a 100644
+--- a/content/app_shim_remote_cocoa/web_contents_view_cocoa.mm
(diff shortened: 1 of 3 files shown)


patches/common/chromium/chore_add_debounce_on_the_updatewebcontentsvisibility_method_to.patch
```diff
@@ -0,0 +1,65 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <sattard@slack-corp.com>
+Date: Wed, 5 Jun 2019 15:11:00 -0700
+Subject: chore: add debounce on the updateWebContentsVisibility method to
+ ensure quick changes in occlusion do not result in flickering
+
```
codebytere: Could you potentially copy the PR body into the extended commit message for posterity?

patches/common/chromium/chore_add_debounce_on_the_updatewebcontentsvisibility_method_to.patch
```diff
@@ -0,0 +1,65 @@
...
+     visibility = Visibility::kHidden;
+@@ -267,7 +272,24 @@ - (void)updateWebContentsVisibility {
+     visibility = Visibility::kVisible;
+   else
+     visibility = Visibility::kOccluded;
+-  host_->OnWindowVisibilityChanged(visibility);
++  return visibility;
++}
++
++- (void)updateWebContentsVisibility {
```
codebytere: Is there a chance this could have unintended side effects for occlusion events we _do_ need to be notified of at the moment they occur?

patches/common/chromium/chore_add_debounce_on_the_updatewebcontentsvisibility_method_to.patch
```diff
@@ -0,0 +1,65 @@
...
++  if (!host_)
```
codebytere: Is there a chance this could have unintended side effects for occlusion events we _do_ need to be notified of at the moment they occur?

patches/common/chromium/chore_add_debounce_on_the_updatewebcontentsvisibility_method_to.patch
```diff
@@ -0,0 +1,65 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <sattard@slack-corp.com>
+Date: Wed, 5 Jun 2019 15:11:00 -0700
+Subject: chore: add debounce on the updateWebContentsVisibility method to
+ ensure quick changes in occlusion do not result in flickering
```
codebytere: Could you potentially copy the PR body into the extended commit message for posterity?

patches/common/chromium/chore_add_debounce_on_the_updatewebcontentsvisibility_method_to.patch
MarshallOfSound: Electron itself has an independent occlusion observer.  My current understanding of the use of this in Chromium is that delaying this call for occlusion has no impact other than rendering the webContents for an extra few seconds

zcbenz:
This needs rebasing.

r0stig:
Is this being backported to 5.0.x? If not, when is 6.0.x planned to be released as stable?

Issue #18664, opened 2019-06-06
Replace StatusItemView implementation with [NSStatusItem button]


MarshallOfSound:
Now that we have a supported platform list of >= 10.10 we can use the `[NSStatusItem button]` instead of the hacks and tricks we use currently.

DeerMichel:
# Tray Migration for macOS Catalina

Docs: https://electronjs.org/docs/api/tray  

`statusItem: NSStatusItem`

## Instance Methods

- `destroy`
  * not tested, but shouldnt be a problem
- `setImage`
  * works via `[[statusItem button] setImage]`
- `setPressedImage`
  * works via `[[statusItem button] setAlternateImage]`
- `setToolTip`
  * works via `[[statusItem button] setToolTip]`
- `setTitle`
  * works via `[[statusItem button] setTitle]`
  * icon & text position needs to be set using `[[statusItem button] setImagePosition]`
- `getTitle`
  * works via `[[statusItem button] title]`
- `setHighlightMode`
  * i have no idea... 
  * there is no custom draw function anymore, maybe drop support for this func? or worth finding a custom solution which might break again in future macOS versions?
  * however there is `[NSButton setHighlighted]` which will be reset on click - maybe a custom click handler can uphold the highlighting state?
  * https://stackoverflow.com/questions/26004684/nsstatusbarbutton-keep-highlighted
- `setIgnoreDoubleClickEvents`
  * no idea about this one... my guess is a custom click handler
- `getIgnoreDoubleClickEvents`
  * correlates with setter
- `popUpContextMenu`
  * maybe?
- `setContextMenu`
  * works somehow via `[[statusItem button] setMenu]`
  * needs further investigation, e.g. radios, buttons not enabled (see #random channel)
  * `[statusItem popUpStatusItemMenu]` but did work.. but is deprecated yay
- `getBounds`
  * works via `[[statusItem button] frame]`
- `isDestroyed`
  * not tested, but shouldnt be a problem

## Events

- `click`
  * defaults to open assigned `[[statusItem button] menu]`
  * not sure if can be overriden
  * custom view needed for that, workaround?
- `right-click`, `double-click`
  * custom view needed for that, workaround?
- `drop`, `drop-files`, `drop-text`
  * custom view needed for that, workaround?
- `drag-enter`, `drag-leave`, `drag-end`
  * custom view needed for that, workaround?
- `mouse-enter`, `mouse-leave`, `mouse-move`
  * detected using a `NSTrackingArea`
[...]

PR #18666, opened 2019-06-06
fix: use effectiveAppearance to check dark mode on mojave and above

### LINKED ISSUE #18663: Tray icon becomes invisible when selecting "Auto" appearance in System Preferences on macOS 10.15 beta1
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 2.0.7
* **Operating System:**
  * macOS 10.15 beta1
  * N/A

### Expected Behavior

The tray icon can display clearly when selecting "Auto" appearance in System Preferences on macOS 10.15 beta1

### Actual Behavior

The tray icon is invisible when selecting "Auto" appearance in System Preferences on macOS 10.15 beta1

### To Reproduce

1. Install WD Discovery or VMware Fusion
2. Switch macOS 10.15 beta1 Appearance setting to "Auto"
3. Check the tray icon in the system menu bar
4. The tray icon is invisible

### Additional Information

Click on the tray icon, if you specify highlight/selected tray image, it displays well

MarshallOfSound:
Fixes #18663

So on Catalina there is a new "auto" mode that doesn't update the user pref but does update `effectiveAppearance`.  The correct thing to do here is read it from `effectiveAppearance` anyway so let's just do that.

In the future I plan on doing #18664 but this is a solution for all our maintained versions 👍 

Notes: Fixed issue on macOS Catalina where the tray icon would be highlighted incorrectly and sometimes invisible

### DIFF
--- a/atom/browser/ui/tray_icon_cocoa.mm
+++ b/atom/browser/ui/tray_icon_cocoa.mm
@@ -10,6 +10,7 @@
 #include "atom/browser/mac/atom_application.h"
 #include "atom/browser/ui/cocoa/NSString+ANSI.h"
 #include "atom/browser/ui/cocoa/atom_menu_controller.h"
+#include "base/mac/sdk_forward_declarations.h"
 #include "base/strings/sys_string_conversions.h"
 #include "ui/display/screen.h"
 #include "ui/events/cocoa/cocoa_event_utils.h"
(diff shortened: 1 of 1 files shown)


deepak1556:
Chromium has unified dark mode detection for macOS and windows, we should try switching to the utilities in `NativeTheme` for this api https://bugs.chromium.org/p/chromium/issues/detail?id=893598

codebytere:
ooh interesting! @MarshallOfSound mind if I look into that?

Issue #16476, opened 2019-01-21
Feature: HTTP preconnect


robatwilliams:
* Output of `node_modules/.bin/electron --version`: 4.01 (originally discovered using 2.0.13)
* Operating System (Platform and Version): Windows 10 64-bit
* Output of `node_modules/.bin/electron --version` on last known working Electron version (if applicable): n/a

**Expected Behavior**
Should behave as standalone Chromium does when loading remote pages:
* When a page provides a `preconnect` resource hint via a response header or link tag, it should preemptively preconnect to the specified domain
* When loading a page that hasn't been seen before, it should preemptively open 2 connections to the origin
* When loading a page that has been seen before, it should use collected data from previous visits to preemptively open however many connections it believes are likely to be necessary

**Actual behavior**
No preconnects are made in any of the above scenarios. This is verified by the absence in the network log of any `HTTP_STREAM_JOB_CONTROLLER` events with `is_preconnect = true`. 

**To Reproduce**
~~I don't have a repo I can share, but...~~ See repo & instructions in immediately-following comments. Starting a window with `--log-net-log=C:\log.json` and causing the scenarios above, will reveal in the log that no preconnects are made. Do the same with a standalone Chromium (same v69), and preconnects are made.

**Screenshots**
n/a

**Additional Information**
⚠️ edited to add: feature not part of the content module? See [comment](https://github.com/electron/electron/issues/16476#issuecomment-456972406)

I've looked through the Chromium source code to find the expected behaviour described above - I believe the relevant parts are [loading_predictor.cc:PrepareForPageLoad](https://chromium.googlesource.com/chromium/src/+/69.0.3497.128/chrome/browser/predictors/loading_predictor.cc#70), [loading_predictor.cc:AddInitialUrlToPreconnectPrediction](https://chromium.googlesource.com/chromium/src/+/69.0.3497.128/chrome/browser/predictors/loading_predictor.cc#27), and [resource_prefetch_predictor.cc:PredictPreconnectOrigins](https://chromium.googlesource.com/chromium/src/+/69.0.3497.128/chrome/browser/predictors/resource_prefetch_predictor.cc#211). While I've been able to debug code found in the electron repo, I haven't been able to debug this because the symbols aren't in the PDB file.

This existing [question](https://stackoverflow.com/questions/50377941/preconnect-resource-hint-not-working-in-electron-but-works-fine-in-chrome) on StackOverflow appears to be having the same problem.

[...]

PR #18671, opened 2019-06-06
feat: HTTP preconnect feature minimal for electronjs

krunt:
#### Description of Change
Added HTTP preconnect hint support.

Using the parameter numSocketsToPreconnect in BrowserWindow you can control how many sockets will be open.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:

### DIFF
--- a/atom/browser/net/preconnect_manager_tab_helper.h
+++ b/atom/browser/net/preconnect_manager_tab_helper.h
@@ -1,0 +1,50 @@
+// Copyright (c) 2019 GitHub, Inc.
+// Use of this source code is governed by the MIT license that can be
+// found in the LICENSE file.
+
+#ifndef ATOM_BROWSER_NET_PRECONNECT_MANAGER_TAB_HELPER_H_
+#define ATOM_BROWSER_NET_PRECONNECT_MANAGER_TAB_HELPER_H_
+
+#include "content/public/browser/navigation_handle.h"
+#include "content/public/browser/web_contents_observer.h"
+#include "content/public/browser/web_contents_user_data.h"
+
+namespace predictors {
+class PreconnectManager;
+}
+
+namespace atom {
+
+class WebContentsPreferences;
+
+class PreconnectManagerTabHelper
+    : public content::WebContentsObserver,
+      public content::WebContentsUserData<PreconnectManagerTabHelper> {
+ public:
+  ~PreconnectManagerTabHelper() override;
+
+  // content::WebContentsObserver implementation
+  void DidStartNavigation(
+      content::NavigationHandle* navigation_handle) override;
+
+  // -1 - in case of number of preconnect sockets option is not set
+  static int GetNumberOfSocketsToPreconnect(WebContentsPreferences* prefs);
+
+  void SetNumberOfSocketsToPreconnect(int numSockets);
+
+ private:
+  explicit PreconnectManagerTabHelper(content::WebContents* web_contents);
+  friend class content::WebContentsUserData<PreconnectManagerTabHelper>;
+
+  predictors::PreconnectManager* preconnect_manager_;
+
+  int number_of_sockets_to_preconnect_;
+
+  WEB_CONTENTS_USER_DATA_KEY_DECL();
+
+  DISALLOW_COPY_AND_ASSIGN(PreconnectManagerTabHelper);
+};
+
+}  // namespace atom
+
+#endif  // ATOM_BROWSER_NET_PRECONNECT_MANAGER_TAB_HELPER_H_
--- a/atom/browser/renderer_host/DEPS
+++ b/atom/browser/renderer_host/DEPS
@@ -1,0 +1,6 @@
+include_rules = [
+  # renderer_host is intended to support the content layer's renderers. No
+  # reference to the tab contents is allowed.
+  "-chrome/browser/tab_contents",
+]
+
--- a/atom/browser/renderer_host/atom_render_message_filter.cc
+++ b/atom/browser/renderer_host/atom_render_message_filter.cc
@@ -1,0 +1,98 @@
+// Copyright (c) 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <stdint.h>
+#include <memory>
+#include <string>
+
+#include "atom/browser/net/preconnect_manager_factory.h"
+#include "atom/browser/renderer_host/atom_render_message_filter.h"
+
+#include "base/bind.h"
+#include "base/bind_helpers.h"
+#include "base/logging.h"
+#include "base/stl_util.h"
+#include "base/task/post_task.h"
+#include "chrome/browser/browser_process.h"
+#include "chrome/browser/chrome_notification_types.h"
+#include "chrome/browser/predictors/preconnect_manager.h"
+#include "chrome/browser/profiles/profile.h"
+#include "components/network_hints/common/network_hints_common.h"
+#include "components/network_hints/common/network_hints_messages.h"
+#include "content/public/browser/browser_task_traits.h"
+#include "content/public/browser/notification_service.h"
+#include "content/public/browser/render_process_host.h"
+#include "content/public/browser/storage_partition.h"
+
+using content::BrowserThread;
+
+namespace {
+
+const uint32_t kRenderFilteredMessageClasses[] = {
+    NetworkHintsMsgStart,
+};
+
+}  // namespace
+
+AtomRenderMessageFilter::AtomRenderMessageFilter(
+    int render_process_id,
+    Profile* profile,
+    int number_of_sockets_to_preconnect)
+    : BrowserMessageFilter(kRenderFilteredMessageClasses,
+                           base::size(kRenderFilteredMessageClasses)),
+      render_process_id_(render_process_id),
+      preconnect_manager_(nullptr),
+      number_of_sockets_to_preconnect_(number_of_sockets_to_preconnect) {
+  preconnect_manager_ = atom::PreconnectManagerFactory::GetForProfile(profile);
+}
+
+AtomRenderMessageFilter::~AtomRenderMessageFilter() {}
+
+bool AtomRenderMessageFilter::OnMessageReceived(const IPC::Message& message) {
+  bool handled = true;
+  IPC_BEGIN_MESSAGE_MAP(AtomRenderMessageFilter, message)
+    IPC_MESSAGE_HANDLER(NetworkHintsMsg_Preconnect, OnPreconnect)
+    IPC_MESSAGE_UNHANDLED(handled = false)
+  IPC_END_MESSAGE_MAP()
+
+  return handled;
+}
+
+void AtomRenderMessageFilter::OnPreconnect(const GURL& url,
+                                           bool allow_credentials,
+                                           int count) {
+  if (count < 1) {
+    LOG(WARNING) << "NetworkHintsMsg_Preconnect IPC with invalid count: "
+                 << count;
+    return;
+  }
+
+  if (!url.is_valid() || !url.has_host() || !url.has_scheme() ||
+      !url.SchemeIsHTTPOrHTTPS()) {
+    return;
+  }
+
+  if (!preconnect_manager_) {
+    return;
+  }
+
+  if (number_of_sockets_to_preconnect_ > 0) {
+    std::vector<predictors::PreconnectRequest> requests;
+    requests.emplace_back(url.GetOrigin(), number_of_sockets_to_preconnect_);
+
+    base::PostTaskWithTraits(
+        FROM_HERE, {BrowserThread::UI},
+        base::BindOnce(&predictors::PreconnectManager::Start,
+                       base::Unretained(preconnect_manager_), url, requests));
+  }
+}
+
+namespace predictors {
+
+PreconnectRequest::PreconnectRequest(const GURL& origin, int num_sockets)
+    : origin(origin), num_sockets(num_sockets) {
+  DCHECK_GE(num_sockets, 0);
+}
+
+}  // namespace predictors
--- a/atom/browser/renderer_host/atom_render_message_filter.h
+++ b/atom/browser/renderer_host/atom_render_message_filter.h
@@ -1,0 +1,66 @@
+// Copyright (c) 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef ATOM_BROWSER_RENDERER_HOST_ATOM_RENDER_MESSAGE_FILTER_H_
+#define ATOM_BROWSER_RENDERER_HOST_ATOM_RENDER_MESSAGE_FILTER_H_
+
+#include <string>
+#include <vector>
+
+#include "base/callback.h"
+#include "base/macros.h"
+#include "base/sequenced_task_runner_helpers.h"
+#include "content/public/browser/browser_message_filter.h"
+#include "content/public/browser/browser_thread.h"
+#include "extensions/buildflags/buildflags.h"
+#include "ppapi/buildflags/buildflags.h"
+
+class GURL;
+class Profile;
+
+namespace predictors {
+class PreconnectManager;
+}
+
+namespace content_settings {
+class CookieSettings;
+}
+
+namespace network_hints {
+struct LookupRequest;
+}
+
+class AtomPreconnectDelegate;
+
+// This class filters out incoming Chrome-specific IPC messages for the renderer
+// process on the IPC thread.
+class AtomRenderMessageFilter : public content::BrowserMessageFilter {
+ public:
+  AtomRenderMessageFilter(int render_process_id,
+                          Profile* profile,
+                          int number_of_sockets_to_preconnect);
+
+  // content::BrowserMessageFilter methods:
+  bool OnMessageReceived(const IPC::Message& message) override;
+
+ private:
+  friend class content::BrowserThread;
+  friend class base::DeleteHelper<AtomRenderMessageFilter>;
+
+  ~AtomRenderMessageFilter() override;
+
+  void OnPreconnect(const GURL& url, bool allow_credentials, int count);
+
+  const int render_process_id_;
+
+  // The PreconnectManager for the associated Profile. This must only be
+  // accessed on the UI thread.
+  predictors::PreconnectManager* preconnect_manager_;
+
+  int number_of_sockets_to_preconnect_;
+
+  DISALLOW_COPY_AND_ASSIGN(AtomRenderMessageFilter);
+};
+
+#endif  // ATOM_BROWSER_RENDERER_HOST_ATOM_RENDER_MESSAGE_FILTER_H_
--- a/atom/browser/net/preconnect_manager_factory.cc
+++ b/atom/browser/net/preconnect_manager_factory.cc
@@ -1,0 +1,73 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+//
+#include <memory>
+
+#include "atom/atom/browser/net/preconnect_manager_factory.h"
+
+#include "chrome/browser/predictors/preconnect_manager.h"
+#include "chrome/browser/profiles/profile.h"
+#include "components/keyed_service/content/browser_context_dependency_manager.h"
+#include "components/keyed_service/content/browser_context_keyed_service_factory.h"
+#include "content/public/browser/browser_context.h"
+#include "content/public/browser/storage_partition.h"
+
+namespace {
+class StubDelegate : public predictors::PreconnectManager::Delegate {
+ public:
+  void PreconnectFinished(
+      std::unique_ptr<predictors::PreconnectStats> stats) override {}
+};
+
+class PreconnectManagerWrapper : public KeyedService {
+ public:
+  PreconnectManagerWrapper(
+      base::WeakPtr<predictors::PreconnectManager::Delegate> delegate,
+      Profile* profile)
+      : ptr_(new predictors::PreconnectManager(delegate, profile)) {
+    ptr_->SetNetworkContextForTesting(
+        content::BrowserContext::GetDefaultStoragePartition(profile)
+            ->GetNetworkContext());
+  }
+  ~PreconnectManagerWrapper() override {
+    delete ptr_;
+    ptr_ = NULL;
+  }
+  predictors::PreconnectManager* GetPtr() { return ptr_; }
+
+ private:
+  predictors::PreconnectManager* ptr_;
+};
+}  // namespace
+
+namespace atom {
+
+// static
+predictors::PreconnectManager* PreconnectManagerFactory::GetForProfile(
+    Profile* profile) {
+  return static_cast<PreconnectManagerWrapper*>(
+             GetInstance()->GetServiceForBrowserContext(profile, true))
+      ->GetPtr();
+}
+
+// static
+PreconnectManagerFactory* PreconnectManagerFactory::GetInstance() {
+  return base::Singleton<PreconnectManagerFactory>::get();
+}
+
+PreconnectManagerFactory::PreconnectManagerFactory()
+    : BrowserContextKeyedServiceFactory(
+          "PreconnectManager",
+          BrowserContextDependencyManager::GetInstance()),
+      weak_factory_(new StubDelegate()) {}
+
+PreconnectManagerFactory::~PreconnectManagerFactory() {}
+
+KeyedService* PreconnectManagerFactory::BuildServiceInstanceFor(
+    content::BrowserContext* context) const {
+  Profile* profile = static_cast<Profile*>(context);
+  return new PreconnectManagerWrapper(weak_factory_.GetWeakPtr(), profile);
+}
+
+}  // namespace atom
--- a/atom/browser/net/preconnect_manager_factory.h
+++ b/atom/browser/net/preconnect_manager_factory.h
@@ -1,0 +1,41 @@
+// Copyright (c) 2019 GitHub, Inc.
+// Use of this source code is governed by the MIT license that can be
+// found in the LICENSE file.
+
+#ifndef ATOM_BROWSER_NET_PRECONNECT_MANAGER_FACTORY_H_
+#define ATOM_BROWSER_NET_PRECONNECT_MANAGER_FACTORY_H_
+
+#include "base/macros.h"
+#include "base/memory/singleton.h"
+#include "base/memory/weak_ptr.h"
+#include "chrome/browser/predictors/preconnect_manager.h"
+#include "components/keyed_service/content/browser_context_keyed_service_factory.h"
+
+class Profile;
+
+namespace atom {
+
+class PreconnectManagerFactory : public BrowserContextKeyedServiceFactory {
+ public:
+  static predictors::PreconnectManager* GetForProfile(Profile* profile);
+  static PreconnectManagerFactory* GetInstance();
+
+ private:
+  friend struct base::DefaultSingletonTraits<PreconnectManagerFactory>;
+
+  PreconnectManagerFactory();
+  ~PreconnectManagerFactory() override;
+
+  // BrowserContextKeyedServiceFactory:
+  KeyedService* BuildServiceInstanceFor(
+      content::BrowserContext* context) const override;
+
+  mutable base::WeakPtrFactory<predictors::PreconnectManager::Delegate>
+      weak_factory_;
+
+  DISALLOW_COPY_AND_ASSIGN(PreconnectManagerFactory);
+};
+
+}  // namespace atom
+
+#endif  // ATOM_BROWSER_NET_PRECONNECT_MANAGER_FACTORY_H_
--- a/atom/browser/net/preconnect_manager_tab_helper.cc
+++ b/atom/browser/net/preconnect_manager_tab_helper.cc
@@ -1,0 +1,97 @@
+// Copyright (c) 2019 GitHub, Inc.
+// Use of this source code is governed by the MIT license that can be
+// found in the LICENSE file.
+
+#include <algorithm>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "atom/browser/net/preconnect_manager_tab_helper.h"
+
+#include "atom/browser/net/preconnect_manager_factory.h"
+#include "atom/browser/web_contents_preferences.h"
+#include "atom/common/options_switches.h"
+
+#include "base/task/post_task.h"
+#include "chrome/browser/profiles/profile.h"
+#include "content/public/browser/browser_context.h"
+#include "content/public/browser/browser_task_traits.h"
+#include "content/public/browser/browser_thread.h"
+#include "content/public/browser/render_process_host.h"
+
+namespace atom {
+
+PreconnectManagerTabHelper::PreconnectManagerTabHelper(
+    content::WebContents* web_contents)
+    : content::WebContentsObserver(web_contents),
+      preconnect_manager_(nullptr),
+      number_of_sockets_to_preconnect_(-1) {
+  preconnect_manager_ = atom::PreconnectManagerFactory::GetForProfile(
+      static_cast<Profile*>(web_contents->GetBrowserContext()));
+}
+
+PreconnectManagerTabHelper::~PreconnectManagerTabHelper() = default;
+
+void PreconnectManagerTabHelper::DidStartNavigation(
+    content::NavigationHandle* navigation_handle) {
+  DCHECK_CURRENTLY_ON(content::BrowserThread::UI);
+  if (!preconnect_manager_)
+    return;
+
+  if (navigation_handle->GetURL().is_empty() ||
+      !navigation_handle->GetURL().SchemeIsHTTPOrHTTPS())
+    return;
+
+  if (number_of_sockets_to_preconnect_ >= 0) {
+    std::vector<predictors::PreconnectRequest> requests;
+    requests.emplace_back(navigation_handle->GetURL().GetOrigin(),
+                          number_of_sockets_to_preconnect_);
+
+    base::PostTaskWithTraits(
+        FROM_HERE, {content::BrowserThread::UI},
+        base::BindOnce(&predictors::PreconnectManager::Start,
+                       base::Unretained(preconnect_manager_),
+                       navigation_handle->GetURL(), requests));
+  }
+}
+
+static bool GetAsInteger(const base::Value* val,
+                         const base::StringPiece& path,
+                         int* out) {
+  if (val) {
+    auto* found = val->FindKey(path);
+    if (found && found->is_int()) {
+      *out = found->GetInt();
+      return true;
+    } else if (found && found->is_string()) {
+      return base::StringToInt(found->GetString(), out);
+    }
+  }
+  return false;
+}
+
+int PreconnectManagerTabHelper::GetNumberOfSocketsToPreconnect(
+    WebContentsPreferences* prefs) {
+  int num_sockets_to_preconnect = -1;
+  if (GetAsInteger(prefs->preference(), options::kNumSocketsToPreconnect,
+                   &num_sockets_to_preconnect)) {
+    const int kMinSocketsToPreconnect = 1;
+    const int kMaxSocketsToPreconnect = 6;
+    num_sockets_to_preconnect =
+        std::max(num_sockets_to_preconnect, kMinSocketsToPreconnect);
+    num_sockets_to_preconnect =
+        std::min(num_sockets_to_preconnect, kMaxSocketsToPreconnect);
+  }
+  return num_sockets_to_preconnect;
+}
+
+void PreconnectManagerTabHelper::SetNumberOfSocketsToPreconnect(
+    int numSockets) {
+  number_of_sockets_to_preconnect_ = numSockets;
+}
+
+WEB_CONTENTS_USER_DATA_KEY_IMPL(PreconnectManagerTabHelper)
+
+}  // namespace atom


coderReview:
Related to issue: https://github.com/electron/electron/issues/16476

krunt:
@MarshallOfSound please look at build-linux failing,
it has AssertionError: exit code should be 123, if you see this please tag @MarshallOfSound: expected 1 to equal 123

Also build-mac fails with
Error: No valid signing identity available to run autoUpdater specs
with I don't understand

coderReview:
@nornagon are you also reviewing this PR? thank you.

atom/browser/renderer_host/DEPS
```diff
@@ -0,0 +1,6 @@
+include_rules = [
+  # renderer_host is intended to support the content layer's renderers. No
+  # reference to the tab contents is allowed.
+  "-chrome/browser/tab_contents",
+]
```
nornagon: Are we checking this? This doesn't seem relevant for Electron.

atom/browser/net/preconnect_manager_tab_helper.h
```diff
@@ -0,0 +1,50 @@
...
+
+namespace predictors {
+class PreconnectManager;
+}
+
+namespace atom {
+
+class WebContentsPreferences;
+
+class PreconnectManagerTabHelper
```
nornagon: While the relevant class in Chrome is called a `TabHelper`, Electron doesn't have the concept of "tabs" so I think this would be better named e.g. `PreconnectManagerHelper`

chromium_src/BUILD.gn
```diff
@@ -55,6 +64,7 @@ static_library("chrome") {
     "//content/public/browser",
   ]
   deps = [
+    "//chrome/browser/:resource_prefetch_predictor_proto",
```
nornagon:
```suggestion
    "//chrome/browser:resource_prefetch_predictor_proto",
```

atom/browser/renderer_host/atom_render_message_filter.cc
```diff
@@ -0,0 +1,98 @@
...
+
+#include "base/bind.h"
+#include "base/bind_helpers.h"
+#include "base/logging.h"
+#include "base/stl_util.h"
+#include "base/task/post_task.h"
+#include "chrome/browser/browser_process.h"
+#include "chrome/browser/chrome_notification_types.h"
+#include "chrome/browser/predictors/preconnect_manager.h"
+#include "chrome/browser/profiles/profile.h"
```
nornagon: We shouldn't be including any `Profile` stuff.

atom/browser/renderer_host/atom_render_message_filter.h
```diff
@@ -0,0 +1,66 @@
...
+#include <string>
+#include <vector>
+
+#include "base/callback.h"
+#include "base/macros.h"
+#include "base/sequenced_task_runner_helpers.h"
+#include "content/public/browser/browser_message_filter.h"
+#include "content/public/browser/browser_thread.h"
+#include "extensions/buildflags/buildflags.h"
+#include "ppapi/buildflags/buildflags.h"
```
nornagon: A lot of these includes appear to be unused.

atom/browser/renderer_host/atom_render_message_filter.h
```diff
@@ -0,0 +1,66 @@
...
+
+ private:
+  friend class content::BrowserThread;
+  friend class base::DeleteHelper<AtomRenderMessageFilter>;
+
+  ~AtomRenderMessageFilter() override;
+
+  void OnPreconnect(const GURL& url, bool allow_credentials, int count);
+
+  const int render_process_id_;
```
nornagon: Is this used anywhere?

atom/browser/renderer_host/atom_render_message_filter.h
```diff
@@ -0,0 +1,66 @@
...
+  AtomRenderMessageFilter(int render_process_id,
+                          Profile* profile,
+                          int number_of_sockets_to_preconnect);
+
+  // content::BrowserMessageFilter methods:
+  bool OnMessageReceived(const IPC::Message& message) override;
+
+ private:
+  friend class content::BrowserThread;
+  friend class base::DeleteHelper<AtomRenderMessageFilter>;
```
nornagon: Is this needed?

atom/browser/atom_browser_client.cc
```diff
@@ -350,7 +354,15 @@ void AtomBrowserClient::RenderProcessWillLaunch(
     prefs.disable_popups = web_preferences->IsEnabled("disablePopups");
     prefs.web_security = web_preferences->IsEnabled(options::kWebSecurity,
                                                     true /* default value */);
+    number_of_sockets_to_preconnect =
+        PreconnectManagerTabHelper::GetNumberOfSocketsToPreconnect(
+            web_preferences);
   }
+
+  Profile* profile = static_cast<Profile*>(host->GetBrowserContext());
```
nornagon: This doesn't seem necessary, as `AtomRenderMessageFilter` doesn't use `Profile`.

nornagon:
Also, this needs tests. Please test that:
1. The feature works (i.e. when specifying the preconnect option, extra sockets are in fact opened when appropriate)
2. The feature isn't enabled unless requested (i.e. when _not_ specifying the preconnect option, no sockets are preconnected)

Additionally, check that things work as expected in iframes and webviews (I think it should work in iframes but maybe not in webviews?)

atom/browser/renderer_host/DEPS
krunt: No, I removed it,

atom/browser/net/preconnect_manager_tab_helper.h
krunt: fixed

atom/browser/atom_browser_client.cc
krunt: I replaced all Profile* with content::BrowserContext*,

atom/browser/renderer_host/atom_render_message_filter.cc
krunt: removed any include of profile.h, replaced with content::BrowserContext

atom/browser/renderer_host/atom_render_message_filter.h
krunt: fixed

atom/browser/renderer_host/atom_render_message_filter.h
krunt: No, they are artifacts from ChromeRenderMessageFilter

atom/browser/renderer_host/atom_render_message_filter.h
krunt: No, I've removed it

chromium_src/BUILD.gn
krunt: fixed

coderReview:
Hello @nornagon thank you for the feedback. We will look at this and get back ASAP.

shell/browser/net/preconnect_manager_factory.cc
```diff
@@ -0,0 +1,75 @@
...
+};
+
+class PreconnectManagerWrapper : public KeyedService {
+ public:
+  PreconnectManagerWrapper(
+      base::WeakPtr<predictors::PreconnectManager::Delegate> delegate,
+      content::BrowserContext* context)
+      : ptr_(new predictors::PreconnectManager(
+            delegate,
+            reinterpret_cast<Profile*>(context))) {
```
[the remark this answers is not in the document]
krunt: preconnect manager uses  AsTestingProfile() for tests in PreconnectManager::GetNetworkContext().
This can be removed by small patch.

chromium_src/BUILD.gn
```diff
@@ -5,6 +5,7 @@
 import("//build/config/ui.gni")
 import("//electron/buildflags/buildflags.gni")
 import("//printing/buildflags/buildflags.gni")
+import("//third_party/protobuf/proto_library.gni")
```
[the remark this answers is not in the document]
krunt: this is actually needed now for compiling preconnect_manager (need to process "//chrome/browser:resource_prefetch_predictor_proto") 
because preconnect_manager.h includes "chrome/browser/predictors/resource_prefetch_predictor.h"
which includes "chrome/browser/predictors/resource_prefetch_predictor_tables.h"
which includes "chrome/browser/predictors/resource_prefetch_predictor.pb.h"

shell/browser/net/preconnect_manager_factory.cc
```diff
@@ -0,0 +1,75 @@
...
+// static
+PreconnectManagerFactory* PreconnectManagerFactory::GetInstance() {
+  return base::Singleton<PreconnectManagerFactory>::get();
+}
+
+PreconnectManagerFactory::PreconnectManagerFactory()
+    : BrowserContextKeyedServiceFactory(
+          "PreconnectManager",
+          BrowserContextDependencyManager::GetInstance()),
+      weak_factory_(new StubDelegate()) {}
```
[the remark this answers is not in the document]
krunt: This stub delegate is only needed for preconnect manager constructor:
PreconnectManager(base::WeakPtr<Delegate> delegate, Profile* profile);

It is not deleted, I will fix that

shell/browser/net/preconnect_manager_helper.cc
```diff
@@ -0,0 +1,93 @@
...
+  return false;
+}
+
+int PreconnectManagerHelper::GetNumberOfSocketsToPreconnect(
+    WebContentsPreferences* prefs) {
+  int num_sockets_to_preconnect = -1;
+  if (GetAsInteger(prefs->preference(), options::kNumSocketsToPreconnect,
+                   &num_sockets_to_preconnect)) {
+    const int kMinSocketsToPreconnect = 1;
+    const int kMaxSocketsToPreconnect = 6;
```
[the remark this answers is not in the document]
krunt: sounds reasonable

shell/browser/net/preconnect_manager_factory.cc
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+  ~PreconnectManagerWrapper() override {
+    delete ptr_;
+    ptr_ = NULL;
+  }
+  predictors::PreconnectManager* GetPtr() { return ptr_; }
+
+ private:
+  predictors::PreconnectManager* ptr_;
```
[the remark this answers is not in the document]
krunt: ok

shell/browser/net/preconnect_manager_factory.cc
```diff
@@ -0,0 +1,75 @@
...
+      ptr_->SetNetworkContextForTesting(
+          content::BrowserContext::GetDefaultStoragePartition(context)
+              ->GetNetworkContext());
+    }
+  }
+  ~PreconnectManagerWrapper() override {
+    delete ptr_;
+    ptr_ = NULL;
+  }
+  predictors::PreconnectManager* GetPtr() { return ptr_; }
```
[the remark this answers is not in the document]
krunt: ok

krunt:
> Additionally, looking through Chrome's code, it looks like it only opens one socket per `Preconnect` message. Can you expand on why you need more than one?

For initial origin there is at least kMinSockets =2 sockets (look at AddInitialUrlToPreconnectPrediction).
We need more than one to speed up loading (to make more sockets available in pool for parallel loading).

krunt:
> I'm not sure that this is the best API we could have for this feature. This gives user per-WebContents control over the number of preconnected sockets, but doesn't allow the user to filter by origin at all (which the underlying API supports).
> 
> Do you think it would make sense to expose some more of the capabilities here & make the API more flexible? e.g. something like
> 
> ```
> webContents.on('preconnect', (url, allow_credentials) => {
>   if (url === 'https://...') {
>     webContents.preconnect(url)
>   }
> })
> ```

I am not sure,
Preconnect can be initiated through:
1) direct navigation on origin via WebContentsObserver
2) through link tag
3) through Link response header

2 and 3 are handled via render message filter which does not know anything 
about web_contents and to which web_contents preconnect message belongs,

Our current interface is not good for electron-style js testing, (it is for c++ unit-tests I guess),
some way is needed to associate electron BrowserWindow with preconnect-request for tests.

chromium_src/BUILD.gn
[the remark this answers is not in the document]
nornagon: I'm not sure how importing the .gni here helps with any of that. Importing the .gni file is only needed if you use the `protoc_convert` or `proto_library` functions in the file, neither of which I see anywhere in this file. Just adding the dependencies should be sufficient.

Does it still work if you remove this line?

shell/browser/net/preconnect_manager_factory.cc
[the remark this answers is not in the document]
nornagon: Ah, right, i got confused by the fact that it's in the "content::BrowserContext implementation" section in https://cs.chromium.org/chromium/src/chrome/browser/profiles/profile.h?q=astestingprofile&sq=package:chromium&g=0&l=151. Yeah, I think it would be better to remove that dependency with a patch.

shell/browser/net/preconnect_manager_factory.cc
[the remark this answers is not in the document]
nornagon: Since we don't actually use the delegate for anything, can we get away with passing nullptr?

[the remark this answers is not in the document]
krunt: year, but maybe we will need delegate for preconnect finished indication for electron tests

chromium_src/BUILD.gn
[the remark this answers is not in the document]
krunt: > Does it still work if you remove this line?
yes, I will fix this

shell/browser/net/preconnect_manager_factory.cc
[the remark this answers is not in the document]
nornagon: For testing, you might be able to get away with detecting the preconnecting sockets on the server end. e.g. something like

```js
const server = http.createServer(...)
const w = new BrowserWindow(...preconnect options...)
let connections = 0
server.on('connection', () => {
  connections++
  if (connections === 4) {
    done()
  }
})
w.loadURL(someUrl)
```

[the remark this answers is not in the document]
krunt: nice idea

shell/browser/net/preconnect_manager_factory.cc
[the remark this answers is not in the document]
krunt: added minimal patch

shell/browser/net/preconnect_manager_factory.cc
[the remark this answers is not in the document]
krunt: added tests

shell/browser/net/preconnect_manager_helper.cc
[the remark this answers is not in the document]
krunt: added checks

shell/browser/renderer_host/atom_render_message_filter.h
```diff
@@ -0,0 +1,56 @@
...
+
+namespace network_hints {
+struct LookupRequest;
+}
+
+class AtomPreconnectDelegate;
+
+// This class filters out incoming Chrome-specific IPC messages for the renderer
+// process on the IPC thread.
+class AtomRenderMessageFilter : public content::BrowserMessageFilter {
```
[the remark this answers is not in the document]
krunt: done

shell/browser/renderer_host/atom_render_message_filter.h
```diff
@@ -0,0 +1,56 @@
...
+
+namespace content_settings {
+class CookieSettings;
+}
+
+namespace network_hints {
+struct LookupRequest;
+}
+
+class AtomPreconnectDelegate;
```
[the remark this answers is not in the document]
krunt: cleaned up a bit

patches/chromium/preconnect_feature.patch
```diff
@@ -0,0 +1,59 @@
+diff --git a/chrome/browser/predictors/preconnect_manager.cc b/chrome/browser/predictors/preconnect_manager.cc
```
nornagon: Please export patches using the `git-export-patches` script. To use this:

1. Make sure your checkout is up to date by running `gclient sync`
2. Edit the files you want to edit in `src/`
3. Commit your changes
4. Run `./electron/script/git-export-patches -o electron/patches/chromium`

This ensures that patches are in a standard format and will be correctly applied by `git-import-patches` when others check out the code.

Further, we require that every patch have a description explaining why the patch is needed, and the conditions under which it could be removed (if that's a plausible future). In this case, a description like this would be appropriate:

> The PreconnectManager in Chrome depends on Profile only for a testing-only feature which we don't use. This patch allows us to avoid an unsafe `reinterpret_cast` and maintain type safety. This could potentially be upstreamed if the testing-only feature were refactored in such a way that Chrome's tests could still run successfully while also removing the dependency on Profile.

spec/api-browser-window-spec.js
```diff
@@ -3027,6 +3027,54 @@ describe('BrowserWindow module', () => {
       })
     })
   })
+
+  describe('preconnect feature', () => {
```
nornagon: This test should go in `spec-main/api-browser-window-spec.js` (which runs in the main process & avoids using the `remote` module, which can result in confusing race conditions)

spec/api-browser-window-spec.js
```diff
@@ -3027,6 +3027,54 @@ describe('BrowserWindow module', () => {
...
+      server.listen(0, '127.0.0.1', () => {
+        url = `http://127.0.0.1:${(server.address() as AddressInfo).port}`
+        done()
+      })
+    })
+    afterEach(() => {
+      server.close()
+    })
+    it(`preconnect number of sockets specified in constructor`, (done) => {
+      connections = 0
```
nornagon: this is done by `beforeEach`, no need to repeat it

spec/api-browser-window-spec.js
```diff
@@ -3027,6 +3027,54 @@ describe('BrowserWindow module', () => {
...
+      const w = new BrowserWindow({ numSocketsToPreconnect: 5 })
```
nornagon: it's important to clean up after the test & close the browser window.

nornagon:
BTW, thanks for your hard work on this—I'm liking where this is going!

coderReview:
Hi @nornagon thank you for your valuable feedback.

spec/api-browser-window-spec.js
krunt: fixed

patches/chromium/preconnect_feature.patch
krunt: done

krunt:
> I'd still like some more thoughts about the API. Given you mentioned (2) and (3) are unrelated to a webcontents, I think the current API is insufficient? Does the current API catch those messages?

current API catch those messages via ElectronRenderMessageFilter

> Perhaps an API like this would make sense?

Yes, this API seems good, 
number of sockets to preconnect can be added as an additional argument maybe?
`   app.preconnect(url, 5)`

coderReview:
@nornagon can you tell us if there is still something missing?
@krunt please fix the small conflict

nornagon:
Yeah, I'm still not happy with the API as implemented. I think the API based on `app.preconnect` is better. I'm OK with preconnecting multiple sockets but given that the implementation would essentially just be a loop in C++, it seems reasonable to just do `for (let i = 0; i < 5; i++) app.preconnect(url)`. The underlying preconnection code doesn't have any particular optimizations for connecting >1 socket (and Chrome itself appears to only ever preconnect a single socket, as far as i can tell).

krunt:
@nornagon, 
I am trying to implement new API, encountered 2 difficulties:
1) when implementing 
`app.on('preconnect', (renderer_id, url, allow_credentials) => {`
how I need to send preconnect event to App from c++, because App is exposed to c++ 
only via delegate in AtomBrowserClient. I consider 
`AtomBrowserClient::Get()->OnPreconnect(renderer_id, url, allow_credentials);`
`void AtomBrowserClient::OnPreconnect(renderer_id, url, allow_credentials)`
`{`
`static_cast<api::App*>(delegate_)->Emit("preconnect", renderer_id, url, allow_credentials);`
`}`
2) when implementing 
`app.preconnect(url)`
we need BrowserContext to get instance of PreconnectManager via electron::PreconnectManagerFactory::GetForContext.
a) we can expand `app.preconnect(url)` to `app.preconnect(url, renderer_id)` 
or
b) put instance of PreconnectManager into AtomBrowserClient
by constructing it with null browser context,
and setting network_context explicitly via PreconnectManager::SetNetworkContextForTesting
can we use AtomBrowserClient::GetSystemNetworkContext() for this?

coderReview:
@nornagon please advise. @krunt please fix conflict. Thanks

nornagon:
@krunt Your code in (1) seems reasonable to me.

(2) implies that preconnections need a BrowserContext to work, so perhaps this api should be on `Session` rather than on `app`?

krunt:
I am not sure about Session,
Is BrowserWindow::webContents::session available before BrowserWindow::loadURL ()?
Preconnections actually do not need BrowserContext to work (after out patch at least), 
but we need to store preconnect_manager instance somewhere, and I used 
PreconnectManagerFactory for this, maybe It's better to create singleton for this,
is it correct practice to use singletons in electron?

nornagon:
The `PreconnectManager` is connected to a particular `BrowserContext`, right? In Electron, each `BrowserContext` (in C++) matches with a specific `Session` (in JS). You can set up a BrowserWindow to belong to a specific Session in the constructor, e.g.

```js
const { session } = require('electron')
const mySession = session.fromPartition('temp-partition')
new BrowserWindow({ webPreferences: { session: mySession } })
```

The default is that new browser windows belong to `session.defaultSession`. So the API could look like:

```js
session.defaultSession.on('preconnect', (event, renderer_id, url, allow_credentials) => {
  session.defaultSession.preconnect({
    origin: ...,
    numSockets: ...,
    allowCredentials: ...,
  })
})
```

Maybe? Does that make sense?

krunt:
> The PreconnectManager is connected to a particular BrowserContext, right?
yes

> Maybe? Does that make sense?
year, this API make sense,

krunt:
@nornagon, do you know how in c++ get an instance to electron::api::Session from content::BrowserContext*? I've not found a way actually.

nornagon:
https://github.com/electron/electron/blob/master/shell/browser/api/atom_api_session.h#L43

docs/api/session.md
```diff
@@ -91,6 +91,15 @@ session.defaultSession.on('will-download', (event, item, webContents) => {
...
 ```
 
+#### Event: 'preconnect'
+
+Returns:
+
+* `event` Event
+* `preconnectUrl` string
+
+Emitted when Electron is about to preconnect `preconnectUrl`.
```
nornagon: This isn't quite accurate: if the JS doesn't do anything, Electron won't preconnect to anything. Perhaps:

> Emitted when a render process requests preconnection to a URL.

docs/api/session.md
```diff
@@ -238,6 +247,14 @@ window.webContents.session.enableNetworkEmulation({
 window.webContents.session.enableNetworkEmulation({ offline: true })
 ```
 
+#### `ses.preconnect(options)`
+
+* `options` Object
+  * `url` String - url for preconnect
```
nornagon:
```suggestion
  * `url` String - URL for preconnect. Only the origin is relevant for opening the socket.
```

docs/api/session.md
```diff
@@ -238,6 +247,14 @@ window.webContents.session.enableNetworkEmulation({
...
+  * `numSockets` Number (optional) - number of sockets to preconnect. Defaults to 1
```
nornagon:
```suggestion
  * `numSockets` Number (optional) - number of sockets to preconnect. Must be between 1 and 6. Defaults to 1.
```

docs/api/session.md
```diff
@@ -238,6 +247,14 @@ window.webContents.session.enableNetworkEmulation({
...
+
+Preconnects number of sockets to url origin
```
nornagon:
```suggestion
Preconnects the given number of sockets to an origin.
```

shell/browser/net/preconnect_manager_helper.cc
```diff
@@ -0,0 +1,49 @@
...
+    content::NavigationHandle* navigation_handle) {
+  DCHECK_CURRENTLY_ON(content::BrowserThread::UI);
+  if (!session_)
+    return;
+
+  if (navigation_handle->GetURL().is_empty() ||
+      !navigation_handle->GetURL().SchemeIsHTTPOrHTTPS())
+    return;
+
+  session_->Emit("preconnect", navigation_handle->GetURL().spec());
```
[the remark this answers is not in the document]
krunt: > this is just the did-start-navigation event on WebContents, right?

year, this is a place in chromium, from which loading predictor initiates preconnection.

> Can we use that event directly from JS?

As I see, It can be done like so:
`contents.session.emit("preconnect", url)`

shell/browser/api/atom_api_session.cc
```diff
@@ -690,6 +693,45 @@ v8::Local<v8::Value> Session::NetLog(v8::Isolate* isolate) {
...
+  options.Get("url", &urlStr);
+  GURL url(urlStr);
+  if (!url.is_valid()) {
+    isolate()->ThrowException(v8::Exception::Error(
+        mate::StringToV8(isolate(), "Must pass non-empty valid url.")));
+    return;
+  }
+  if (options.Get("numSockets", &num_sockets_to_preconnect)) {
+    const int kMinSocketsToPreconnect = 1;
+    const int kMaxSocketsToPreconnect = 6;
```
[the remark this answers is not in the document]
krunt: Actually I've hardcoded it. It does not come from chromium.

shell/browser/renderer_host/electron_render_message_filter.cc
```diff
@@ -0,0 +1,84 @@
...
+    return;
+  }
+
+  if (!url.is_valid() || !url.has_host() || !url.has_scheme() ||
+      !url.SchemeIsHTTPOrHTTPS()) {
+    return;
+  }
+
+  if (session_) {
+    session_->Emit("preconnect", url.spec());
```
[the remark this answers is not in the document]
krunt: > Also, should we pass along the allow_credentials and count values? I'm not sure what the client would do with them but there's presumably a reason they're here, right?

in chromium as I see, allow_credentials is always true when coming from loading predictor,
it can be false when coming from [message](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/html/parser/html_resource_preloader.cc?type=cs&sq=package:chromium&g=0&l=61) 

> The only place where the number of sockets is something other than 0 or 1 is in the loading predictor, which we don't seem to use here?

Right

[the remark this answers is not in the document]
krunt: I made passing allow_credentials

shell/browser/net/preconnect_manager_helper.cc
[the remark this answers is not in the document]
krunt: where in JS can we put it? lib/browser/navigation-controller.js? is the session contents.session at what we should emit?

coderReview:
@krunt we have lot of checks failing. Can you check?
@nornagon can you help checking if now the API is ok? Thanks

chromium_src/BUILD.gn
```diff
@@ -37,6 +38,13 @@ static_library("chrome") {
     "//chrome/browser/net/proxy_config_monitor.h",
     "//chrome/browser/net/proxy_service_factory.cc",
     "//chrome/browser/net/proxy_service_factory.h",
+    "//chrome/browser/predictors/preconnect_manager.cc",
+    "//chrome/browser/predictors/preconnect_manager.h",
+    "//chrome/browser/predictors/proxy_lookup_client_impl.cc",
+    "//chrome/browser/predictors/proxy_lookup_client_impl.h",
+    "//chrome/browser/predictors/resolve_host_client_impl.cc",
+    "//chrome/browser/predictors/resolve_host_client_impl.h",
+    "//chrome/browser/profiles/profile.h",
```
nornagon: I think we don't depend on all of these any more? profile.h definitely shouldn't be included, and I think we can also drop proxy_lookup_client and resolve_host_client.

shell/browser/api/atom_api_session.cc
```diff
@@ -690,6 +693,43 @@ v8::Local<v8::Value> Session::NetLog(v8::Isolate* isolate) {
...
+  predictors::PreconnectManager* preconnect_manager =
+      electron::PreconnectManagerFactory::GetForContext(browser_context());
+  if (preconnect_manager && num_sockets_to_preconnect > 0) {
+    std::vector<predictors::PreconnectRequest> requests = {
+        {url.GetOrigin(), num_sockets_to_preconnect}};
+
+    base::PostTaskWithTraits(
+        FROM_HERE, {content::BrowserThread::UI},
+        base::BindOnce(&predictors::PreconnectManager::Start,
+                       base::Unretained(preconnect_manager), url, requests));
```
nornagon: I don't think `base::Unretained` is correct here; it's possible for the underlying context to be destroyed before this task gets run. Something like this:

```cpp
  base::PostTaskWithTraits(FROM_HERE, {content::BrowserThread::UI},
      base::BindOnce(&StartPreconnectOnUI, browser_context_, url, num_sockets_to_preconnect));
}
// ...

void StartPreconnectOnUI(scoped_refptr<AtomBrowserContext> browser_context, GURL& url, int num_sockets_to_preconnect) {
  std::vector<predictors::PreconnectRequest> requests = {
      {url.GetOrigin(), num_sockets_to_preconnect}}; 
  predictors::PreconnectManager* preconnect_manager =
      electron::PreconnectManagerFactory::GetForContext(browser_context.get());
  preconnect_manager->Start(url, requests);
}
```

shell/browser/net/preconnect_manager_helper.cc
```diff
@@ -0,0 +1,50 @@
...
+    content::NavigationHandle* navigation_handle) {
+  DCHECK_CURRENTLY_ON(content::BrowserThread::UI);
+  if (!session_)
+    return;
+
+  if (navigation_handle->GetURL().is_empty() ||
+      !navigation_handle->GetURL().SchemeIsHTTPOrHTTPS())
+    return;
+
+  session_->Emit("preconnect", navigation_handle->GetURL(), true);
```
nornagon: it's unclear to me that it's useful to emit the `"preconnect"` event whenever we emit `"did-start-navigation"`. should the user be responsible for listening for `"did-start-navigation"` and calling `preconnect()` themselves?

If we do that then we can remove `PreconnectManagerHelper` completely.

shell/browser/net/preconnect_manager_factory.cc
```diff
@@ -0,0 +1,56 @@
...
+#include "content/public/browser/storage_partition.h"
+
+namespace {
+class PreconnectManagerWrapper : public KeyedService {
+ public:
+  explicit PreconnectManagerWrapper(content::BrowserContext* context)
+      : preconnect_manager_(
+            std::make_unique<predictors::PreconnectManager>(nullptr, context)) {
+  }
+  predictors::PreconnectManager* GetPtr() { return preconnect_manager_.get(); }
```
nornagon:
```suggestion
  predictors::PreconnectManager* GetPreconnectManager() { return preconnect_manager_.get(); }
```

shell/browser/api/atom_api_session.cc
```diff
@@ -763,7 +803,8 @@ void Session::BuildPrototype(v8::Isolate* isolate,
       .SetProperty("cookies", &Session::Cookies)
       .SetProperty("netLog", &Session::NetLog)
       .SetProperty("protocol", &Session::Protocol)
-      .SetProperty("webRequest", &Session::WebRequest);
+      .SetProperty("webRequest", &Session::WebRequest)
+      .SetMethod("preconnect", &Session::Preconnect);
```
nornagon: Please put this with the other methods above.

nornagon:
(also, it looks like the tests are legitimately failing!)

chromium_src/BUILD.gn
krunt: I removed profile.h.
preconnect manager needs proxy_lookup_client and resolve_host_client as it resolves host there
before making network requests,

chromium_src/BUILD.gn
```diff
@@ -13,6 +13,7 @@ static_library("chrome") {
   sources = [
     "//chrome/browser/browser_process.cc",
     "//chrome/browser/browser_process.h",
+    "//chrome/browser/chrome_notification_types.h",
```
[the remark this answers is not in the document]
krunt: removed

chromium_src/BUILD.gn
```diff
@@ -55,6 +63,7 @@ static_library("chrome") {
     "//content/public/browser",
   ]
   deps = [
+    "//chrome/browser:resource_prefetch_predictor_proto",
```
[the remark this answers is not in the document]
krunt: yes, compilation is failing if you remove this line with error:
../../chrome/browser/predictors/resource_prefetch_predictor_tables.h:19:10: fatal error: 'chrome/browser/predictors/resource_prefetch_predictor.pb.h' file not found
#include "chrome/browser/predictors/resource_prefetch_predictor.pb.h"

shell/browser/api/atom_api_session.cc
```diff
@@ -690,6 +693,43 @@ v8::Local<v8::Value> Session::NetLog(v8::Isolate* isolate) {
   return v8::Local<v8::Value>::New(isolate, net_log_);
 }
 
+void Session::Preconnect(const mate::Dictionary& options) {
+  GURL url;
+  if (!options.Get("url", &url) || !url.is_valid()) {
+    isolate()->ThrowException(v8::Exception::Error(
+        mate::StringToV8(isolate(), "Must pass non-empty valid url.")));
```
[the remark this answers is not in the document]
krunt: done

shell/browser/api/atom_api_session.cc
```diff
@@ -690,6 +693,43 @@ v8::Local<v8::Value> Session::NetLog(v8::Isolate* isolate) {
...
+  if (options.Get("numSockets", &num_sockets_to_preconnect)) {
+    const int kMinSocketsToPreconnect = 1;
+    const int kMaxSocketsToPreconnect = 6;
+    bool isInsideRange = num_sockets_to_preconnect >= kMinSocketsToPreconnect &&
+                         num_sockets_to_preconnect <= kMaxSocketsToPreconnect;
+    num_sockets_to_preconnect =
+        std::max(num_sockets_to_preconnect, kMinSocketsToPreconnect);
+    num_sockets_to_preconnect =
+        std::min(num_sockets_to_preconnect, kMaxSocketsToPreconnect);
+    if (!isInsideRange) {
```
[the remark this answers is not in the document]
krunt: I agree

shell/browser/api/atom_api_session.cc
krunt: fixed, I used base::RetainedRef also for passing browser_context_ otherwise compilation if failing

shell/browser/api/atom_api_session.cc
krunt: ok

shell/browser/net/preconnect_manager_factory.cc
```diff
@@ -0,0 +1,56 @@
...
+#include <memory>
+
+#include "chrome/browser/predictors/preconnect_manager.h"
+#include "components/keyed_service/content/browser_context_dependency_manager.h"
+#include "components/keyed_service/content/browser_context_keyed_service_factory.h"
+#include "content/public/browser/browser_context.h"
+#include "content/public/browser/storage_partition.h"
+
+namespace {
+class PreconnectManagerWrapper : public KeyedService {
```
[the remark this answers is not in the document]
krunt: Yes, I removed factory

shell/browser/net/preconnect_manager_helper.cc
krunt: I agree, I removed PreconnectManagerHelper completely, 
preconnect event means - explicit preconnect on header/link tag

shell/browser/renderer_host/electron_render_message_filter.cc
```diff
@@ -0,0 +1,88 @@
...
+#include <memory>
+#include <string>
+
+#include "base/bind.h"
+#include "base/bind_helpers.h"
+#include "base/logging.h"
+#include "base/stl_util.h"
+#include "base/task/post_task.h"
+#include "chrome/browser/browser_process.h"
+#include "chrome/browser/chrome_notification_types.h"
```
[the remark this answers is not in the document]
krunt: fixed

coderReview:
@nornagon @MarshallOfSound I think we have covered all required comments made. Also, the tests seems to pass, but fails during cleanup. Can you check again? Once again, thank you for the support on this.

coderReview:
@krunt please double check to see if the tests failure is not caused by any change introduced here.

docs/api/session.md
```diff
@@ -91,6 +91,16 @@ session.defaultSession.on('will-download', (event, item, webContents) => {
 })
 ```
 
+#### Event: 'preconnect'
+
+Returns:
+
+* `event` Event
+* `preconnectUrl` string
```
MarshallOfSound:
```suggestion
* `preconnectUrl` String
```

package.json
```diff
@@ -23,7 +23,7 @@
     "check-for-leaks": "^1.2.1",
     "colors": "^1.1.2",
     "dotenv-safe": "^4.0.4",
-    "dugite": "^1.45.0",
+    "dugite": "^1.87.0",
```
MarshallOfSound: Why bump this?

shell/browser/api/atom_api_session.cc
```diff
@@ -648,6 +649,44 @@ v8::Local<v8::Value> Session::NetLog(v8::Isolate* isolate) {
...
+  std::vector<predictors::PreconnectRequest> requests = {
+      {url.GetOrigin(), num_sockets_to_preconnect}};
+  browser_context->GetPreconnectManager()->Start(url, requests);
+}
+
+void Session::Preconnect(const mate::Dictionary& options,
+                         mate::Arguments* args) {
+  GURL url;
+  if (!options.Get("url", &url) || !url.is_valid()) {
+    args->ThrowError("Must pass non-empty valid url.");
```
MarshallOfSound:
```suggestion
    args->ThrowError("Must pass non-empty valid url to session.preconnect.");
```

package.json
krunt: this was done automatically when I ran tests locally with 
npm run test -- -g=preconnect,
I can revert this if needed

MarshallOfSound: Yeah please revert 👍

krunt:
I've rebased, almost all are passed, 
except linux-ia32-testing-tests which failed by timeout in api-browser-window-spec.ts,  
We don't make any change in api-browser-window-spec.ts in this PR.

package.json
krunt: done

coderReview:
@MarshallOfSound do you know what is missing so we can close this?

shell/browser/api/atom_api_session.cc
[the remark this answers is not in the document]
coderReview: @krunt can you mark this as fixed?

coderReview:
CC @nornagon @MarshallOfSound. We would like to get this to v7.0.0. Is it still possible?

docs/api/session.md
```diff
@@ -91,6 +91,16 @@ session.defaultSession.on('will-download', (event, item, webContents) => {
 })
 ```
 
+#### Event: 'preconnect'
+
+Returns:
+
+* `event` Event
+* `preconnectUrl` String
+* `allow_credentials` Boolean
```
nornagon:
```suggestion
* `allowCredentials` Boolean
```

Also, please document this parameter.

patches/chromium/preconnect_feature.patch
```diff
@@ -0,0 +1,69 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Alexey Kuts <kruntuid@gmail.com>
+Date: Fri, 26 Jul 2019 22:32:54 +0300
+Subject: 	preconnect manager new patch
```
nornagon:
```suggestion
Subject: remove references to Profile from PreconnectManager
```

shell/browser/api/atom_api_session.cc
```diff
@@ -591,6 +592,44 @@ v8::Local<v8::Value> Session::NetLog(v8::Isolate* isolate) {
...
+                         num_sockets_to_preconnect <= kMaxSocketsToPreconnect;
+    if (!isInsideRange) {
+      args->ThrowError(
+          base::StringPrintf("numSocketsToPreconnect is outside range [%d,%d]",
+                             kMinSocketsToPreconnect, kMaxSocketsToPreconnect));
+      return;
+    }
+  }
+
+  if (num_sockets_to_preconnect > 0) {
```
nornagon: This will always be true, right? If so,
```suggestion
  DCHECK(num_sockets_to_preconnect > 0);
```

and un-indent the code below.

shell/browser/renderer_host/electron_render_message_filter.cc
```diff
@@ -0,0 +1,81 @@
...
+    return;
+  }
+
+  if (!url.is_valid() || !url.has_host() || !url.has_scheme() ||
+      !url.SchemeIsHTTPOrHTTPS()) {
+    return;
+  }
+
+  if (session_) {
+    session_->Emit("preconnect", url, allow_credentials);
```
nornagon: is the `count` ever relevant here? Do we want to add it to the call?

nornagon: Actually, it looks like this is always `1` when blink requests it, so we should probably leave it out for now.

nornagon:
This is causing issues in pages with `<link rel="preconnect">`. I've pushed a branch which includes a broken test for `<link rel="preconnect">`: https://github.com/electron/electron/tree/http-preconnect

I'm not yet sure what's causing the hang, but the symptoms are observable by running `npm test -- --runners=main --enable-logging --ci -g 'preconnect'`. When the tests finish (after the link test times out), the app hangs and the GPU helper process is pegged at 100% cpu. Commenting out L371-L375 of atom_browser_client.cc (adding the `ElectronRenderMessageFilter`) causes the issue to go away, though the test still fails.

The cutoff for features merging to v7.0.0 is this week, so if you'd like to get this merged to v7 please take a look soon!

nornagon:
Looks like the cause of the hangs is that `session_->Emit()` is being called on the IO thread (all JS in Electron in the main process must be run on the UI thread).

nornagon:
I've updated the [http-preconnect branch](https://github.com/electron/electron/tree/http-preconnect) with the fix for this: https://github.com/krunt/electron/pull/1

coderReview:
hi @nornagon thank you for the help. We will update the fork @krunt

krunt:
Thanks @nornagon, learned a lot from this PR.

PR #18681, opened 2019-06-07
chore: add ability to log native deprecation warnings

codebytere:
#### Description of Change

This PR seeks to ameliorate an issue by which we're currently incapable of logging bespoke deprecation warnings in native-land, which is something that we've hit in the past and worked around by simply not logging and instead just relying on documentation deprecation to get the job done. In service to users, i'd love to ensure that we're logging warnings in all possible places.

I felt it was overkill to expose this to JS, so it's taken form as a utility that you would pass a deprecation warning to, like so:

```
atom::EmitDeprecationWarning(env, "the thing you are using has been deprecated.", "electron");|
```

which would output:
```
(electron) the thing you are using has been deprecated.
```

cc @MarshallOfSound 

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/atom/common/deprecate_util.cc
+++ b/atom/common/deprecate_util.cc
@@ -4,58 +4,24 @@
 
 #include "atom/common/deprecate_util.h"
 
+#include "atom/common/native_mate_converters/callback.h"
+#include "base/callback.h"
+#include "native_mate/converter.h"
+#include "native_mate/dictionary.h"
+
 namespace atom {
 
-// msg/type/code arguments match Node's process.emitWarning() method. See
-// https://nodejs.org/api/process.html#process_process_emitwarning_warning_type_code_ctor
-v8::Maybe<bool> EmitDeprecationWarning(node::Environment* env,
-                                       std::string warning_msg,
-                                       std::string warning_type = "",
-                                       std::string warning_code = "") {
-  v8::HandleScope handle_scope(env->isolate());
-  v8::Context::Scope context_scope(env->context());
+void EmitDeprecationWarning(node::Environment* env,
+                            const std::string& warning_msg,
+                            const std::string& warning_type) {
+  mate::Dictionary process(env->isolate(), env->process_object());
 
-  const char* warning = warning_msg.empty() ? nullptr : warning_msg.c_str();
-  const char* type = warning_type.empty() ? nullptr : warning_type.c_str();
-  const char* code = warning_code.empty() ? nullptr : warning_code.c_str();
+  base::RepeatingCallback<void(base::StringPiece, base::StringPiece,
+                               base::StringPiece)>
+      emit_warning;
+  process.Get("emitWarning", &emit_warning);
 
-  v8::Local<v8::Object> process = env->process_object();
-  v8::Local<v8::Value> emit_warning;
-  if (!process->Get(env->context(), env->emit_warning_string())
-           .ToLocal(&emit_warning)) {
-    return v8::Nothing<bool>();
-  }
-
-  if (!emit_warning->IsFunction())
-    return v8::Just(false);
-
-  int argc = 0;
-  v8::Local<v8::Value> args[3];  // warning, type, code
-
-  if (!v8::String::NewFromUtf8(env->isolate(), warning,
-                               v8::NewStringType::kNormal)
-           .ToLocal(&args[argc++])) {
-    return v8::Nothing<bool>();
-  }
-  if (type != nullptr) {
-    if (!v8::String::NewFromUtf8(env->isolate(), type,
-                                 v8::NewStringType::kNormal)
-             .ToLocal(&args[argc++])) {
-      return v8::Nothing<bool>();
-    }
-    if (code != nullptr && !v8::String::NewFromUtf8(env->isolate(), code,
-                                                    v8::NewStringType::kNormal)
-                                .ToLocal(&args[argc++])) {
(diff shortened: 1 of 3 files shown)


atom/common/deprecate_util.cc
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2019 GitHub, Inc.
+// Use of this source code is governed by the MIT license that can be
+// found in the LICENSE file.
+
+#include <string>
```
ckerr: _(nit)_ string is part of the public API so it's already included in the header, no need to re-include here

atom/common/deprecate_util.h
```diff
@@ -0,0 +1,21 @@
...
+#ifndef ATOM_COMMON_DEPRECATE_UTIL_H_
+#define ATOM_COMMON_DEPRECATE_UTIL_H_
+
+#include <string>
+
+#include "atom/common/node_includes.h"
+
+namespace atom {
+
+v8::Maybe<bool> EmitDeprecationWarning(node::Environment* env,
```
ckerr:
```suggestion
// msg, type, and code arguments match Node's process.emitWarning() method.
// See https://nodejs.org/api/process.html#process_process_emitwarning_warning_type_code_ctor
v8::Maybe<bool> EmitDeprecationWarning(node::Environment* env,
```

atom/common/deprecate_util.cc
```diff
@@ -0,0 +1,64 @@
...
+  int argc = 0;
+  v8::Local<v8::Value> args[3];  // warning, type, code
+
+  if (!v8::String::NewFromUtf8(env->isolate(), warning,
+                               v8::NewStringType::kNormal)
+           .ToLocal(&args[argc++])) {
+    return v8::Nothing<bool>();
+  }
+  if (type != nullptr) {
+    if (!v8::String::NewFromOneByte(env->isolate(),
```
ckerr: _(minor)_ I think this and below should be `v8::String::NewFromUtf8()` since line 35 reads the warning string is utf-8. It's odd that the first string's encoding would be utf-8 and the other two would not

atom/common/deprecate_util.cc
```diff
@@ -0,0 +1,61 @@
...
+// found in the LICENSE file.
+
+#include "atom/common/deprecate_util.h"
+
+namespace atom {
+
+// msg/type/code arguments match Node's process.emitWarning() method. See
+// https://nodejs.org/api/process.html#process_process_emitwarning_warning_type_code_ctor
+v8::Maybe<bool> EmitDeprecationWarning(node::Environment* env,
+                                       std::string warning_msg,
```
zcbenz: Should pass `const std::string&`.

It seems that this API is mostly called with string literals, if it is the case then passing `base::StringPiece` would be most efficient.

atom/common/deprecate_util.cc
```diff
@@ -0,0 +1,61 @@
...
+                                       std::string warning_type = "",
```
zcbenz: No need for default arguments.

atom/common/deprecate_util.cc
```diff
@@ -0,0 +1,61 @@
...
+
+  int argc = 0;
+  v8::Local<v8::Value> args[3];  // warning, type, code
+
+  if (!v8::String::NewFromUtf8(env->isolate(), warning,
+                               v8::NewStringType::kNormal)
+           .ToLocal(&args[argc++])) {
+    return v8::Nothing<bool>();
+  }
+  if (type != nullptr) {
```
zcbenz: Converting `std::string` to `const char*` is redundant, we can just do `!warning_type.empty()` here.

atom/common/deprecate_util.h
zcbenz: Since it is only used by C++, returning bool would make more sense.

atom/common/deprecate_util.cc
```diff
@@ -0,0 +1,61 @@
...
+  if (!process->Get(env->context(), env->emit_warning_string())
+           .ToLocal(&emit_warning)) {
+    return v8::Nothing<bool>();
+  }
+
+  if (!emit_warning->IsFunction())
+    return v8::Just(false);
+
+  int argc = 0;
+  v8::Local<v8::Value> args[3];  // warning, type, code
```
miniak: I would use `std::vector<v8::Local<v8::Value>> args;`

atom/common/deprecate_util.cc
```diff
@@ -0,0 +1,61 @@
...
+
+  if (!v8::String::NewFromUtf8(env->isolate(), warning,
```
miniak: you can use `mate::StringToV8()` instead

codebytere:
Simplified by using existing process object.

atom/common/deprecate_util.cc
```diff
@@ -0,0 +1,27 @@
...
+#include "atom/common/native_mate_converters/callback.h"
+#include "base/callback.h"
+#include "native_mate/converter.h"
+#include "native_mate/dictionary.h"
+
+namespace atom {
+
+void EmitDeprecationWarning(node::Environment* env,
+                            std::string warning_msg,
+                            std::string warning_type) {
```
zcbenz: Should be `const std::string&` or `base::StringPiece`.

miniak:
@codebytere very nice! it's even less code using all the magic that we have in electron / mate

PR #18682, opened 2019-06-07
chore: convert callbacks-registry to ts

codebytere:
#### Description of Change

Convert callbacks-registry to TypeScript.

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


lib/renderer/callbacks-registry.ts
```diff
@@ -1,16 +1,17 @@
-'use strict'
-
 const v8Util = process.electronBinding('v8_util')
 
-class CallbacksRegistry {
+export class CallbacksRegistry {
+  private nextId: number
+  private callbacks: Record<number, Function>
+
   constructor () {
```
MarshallOfSound: We can remove the constructor now we have members above.  We can initialize them there.

lib/renderer/callbacks-registry.ts
```diff
@@ -1,16 +1,17 @@
-'use strict'
-
 const v8Util = process.electronBinding('v8_util')
 
-class CallbacksRegistry {
+export class CallbacksRegistry {
+  private nextId: number
```
MarshallOfSound: i.e. 
```suggestion
  private nextId = 0
```

lib/renderer/callbacks-registry.ts
```diff
@@ -1,16 +1,17 @@
...
+  private callbacks: Record<number, Function>
```
MarshallOfSound:
```suggestion
  private callbacks: Record<number, Function> = {}
```

lib/renderer/callbacks-registry.ts
```diff
@@ -23,14 +24,14 @@ class CallbacksRegistry {
...
-    while ((match = regexp.exec(stackString)) !== null) {
+    while ((match = regexp.exec(stackString!)) !== null) {
       const location = match[1]
       if (location.includes('(native)')) continue
       if (location.includes('(<anonymous>)')) continue
       if (location.includes('electron/js2c')) continue
 
       const ref = /([^/^)]*)\)?$/gi.exec(location)
-      filenameAndLine = ref[1]
+      filenameAndLine = ref![1]
```
MarshallOfSound: I'd prefer a null check here instead of telling TS we know better, the stack trace here could be Bad somehow (user overrides `global.Error` or something idk

lib/renderer/callbacks-registry.ts
```diff
@@ -23,14 +24,14 @@ class CallbacksRegistry {
     let filenameAndLine
     let match
 
-    while ((match = regexp.exec(stackString)) !== null) {
+    while ((match = regexp.exec(stackString!)) !== null) {
```
MarshallOfSound: Can we do a falsey check above here to ensure that it isn't null, better safe than telling TS we know better.

typings/internal-ambient.d.ts
```diff
@@ -16,6 +16,7 @@ declare namespace NodeJS {
   interface V8UtilBinding {
     getHiddenValue<T>(obj: any, key: string): T;
     setHiddenValue<T>(obj: any, key: string, value: T): void;
+    deleteHiddenValue<T>(obj: any, key: string): T;
```
MarshallOfSound: `deleteHiddenValue` returns `void` --> https://github.com/electron/electron/blob/master/atom/common/api/atom_api_v8_util.cc#L82

```suggestion
    deleteHiddenValue(obj: any, key: string): void;
```

Issue #18685, opened 2019-06-07
win.webContents.destroy() causes a crash


miniak:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 3.1.11, 4.2.4, 5.0.2
* **Operating System:**
  * macOS 10.14.4
* **Last Known Working Electron version:**
  * 2.0.18

### Expected Behavior
Calling `webContents.destroy()` on a `BrowserWindow` instance should not crash Electron.

### Actual Behavior
Calling `webContents.destroy()` on a `BrowserWindow` instance crashes Electron.

### To Reproduce
Append `mainWindow.webContents.destroy()` at the end of the `createWindow` function in Fiddle.

### Additional Information
Seems to be related to https://github.com/electron/electron/commit/a25b49a127d08849835b6781092843f2fc145d87#diff-3d0d8f51a9e7a8d2a462429afa5c5959

```
Thread 0 Crashed:: CrBrowserMain  Dispatch queue: com.apple.main-thread
0   com.github.electron.framework 	0x0000000101e5dc28 base::ObserverList<atom::api::ExtendedWebContentsObserver, false, true>::RemoveObserver(atom::api::ExtendedWebContentsObserver const*) + 648 (algorithm:999)
1   com.github.electron.framework 	0x0000000101e5908f atom::api::WebContents::RemoveObserver(atom::api::ExtendedWebContentsObserver*) + 47 (atom_api_web_contents.h:286)
2   com.github.electron.framework 	0x0000000101e58fbd atom::api::BrowserWindow::~BrowserWindow() + 141 (atom_api_browser_window.cc:101)
3   com.github.electron.framework 	0x0000000101e591d5 atom::api::BrowserWindow::~BrowserWindow() + 21 (atom_api_browser_window.cc:103)
4   com.github.electron.framework 	0x0000000101e59279 atom::api::BrowserWindow::~BrowserWindow() + 25 (atom_api_browser_window.cc:97)
5   com.github.electron.framework 	0x0000000101e7379b mate::TrackableObjectBase::Destroy() + 43 (trackable_object.cc:45)
[...]

Issue #16580, opened 2019-01-29
Segfault in RemoveObserver when destroying BrowserWindow


nornagon:
* Output of `node_modules/.bin/electron --version`: master @ 138ba5351
* Operating System (Platform and Version): macOS 10.13.6

```
Received signal 11 <unknown> 000000000000
0   libbase.dylib                       0x00000001076ee813 base::debug::StackTrace::StackTrace(unsigned long) + 83
1   libbase.dylib                       0x00000001076ee8cd base::debug::StackTrace::StackTrace(unsigned long) + 29
2   libbase.dylib                       0x00000001073c58ca base::debug::StackTrace::StackTrace() + 26
3   libbase.dylib                       0x00000001076ee661 base::debug::(anonymous namespace)::StackDumpSignalHandler(int, __siginfo*, void*) + 1393
4   libsystem_platform.dylib            0x00007fff553f6f5a _sigtramp + 26
5   ???                                 0x0000000000000005 0x0 + 5
6   libbase.dylib                       0x00000001073fb7f3 base::Lock::Acquire() + 35
7   libbase.dylib                       0x00000001073fb7c3 base::AutoLock::AutoLock(base::Lock&) + 35
8   libbase.dylib                       0x00000001073b821d base::AutoLock::AutoLock(base::Lock&) + 29
9   libbase.dylib                       0x0000000107516b4e base::SequenceCheckerImpl::CalledOnValidSequence() const + 46
10  libbase.dylib                       0x000000010746530a base::internal::WeakReference::Flag::IsValid() const + 58
11  libbase.dylib                       0x000000010746573e base::internal::WeakReference::IsValid() const + 62
12  Electron Framework                  0x000000010d83806f base::WeakPtr<base::CheckedObserver>::get() const + 31
13  Electron Framework                  0x000000010d838039 base::internal::CheckedObserverAdapter::IsEqual(base::CheckedObserver const*) const + 25
14  Electron Framework                  0x000000010d6d4e60 auto base::ObserverList<atom::api::ExtendedWebContentsObserver, false, true, base::internal::CheckedObserverAdapter>::RemoveObserver(atom::api::ExtendedWebContentsObserver const*)::'lambda'(atom::api::ExtendedWebContentsObserver const&)::operator()<base::internal::CheckedObserverAdapter>(atom::api::ExtendedWebContentsObserver const&) const + 32
[...]

PR #18686, opened 2019-06-07
fix: crash in BrowserWindow destructor after win.webContents.destroy()

miniak:
#### Description of Change
Fixes https://github.com/electron/electron/issues/18685

Make `api_web_contents_` a weak pointer to avoid accessing a destroyed `api::WebContents` instance.

/cc @zcbenz

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Fixed crash in BrowserWindow destructor after `win.webContents.destroy()`.

### DIFF
--- a/atom/browser/api/atom_api_browser_window.cc
+++ b/atom/browser/api/atom_api_browser_window.cc
@@ -66,7 +66,7 @@ BrowserWindow::BrowserWindow(v8::Isolate* isolate,
   }
 
   web_contents_.Reset(isolate, web_contents.ToV8());
-  api_web_contents_ = web_contents.get();
+  api_web_contents_ = web_contents->GetWeakPtr();
   api_web_contents_->AddObserver(this);
   Observe(api_web_contents_->web_contents());
 
@@ -93,7 +93,9 @@ BrowserWindow::BrowserWindow(v8::Isolate* isolate,
 }
 
 BrowserWindow::~BrowserWindow() {
-  api_web_contents_->RemoveObserver(this);
+  // FIXME This is a hack rather than a proper fix preventing shutdown crashes.
+  if (api_web_contents_)
+    api_web_contents_->RemoveObserver(this);
   // Note that the OnWindowClosed will not be called after the destructor runs,
   // since the window object is managed by the TopLevelWindow class.
   if (web_contents())
--- a/atom/browser/api/atom_api_browser_window.h
+++ b/atom/browser/api/atom_api_browser_window.h
@@ -116,7 +116,7 @@ class BrowserWindow : public TopLevelWindow,
 #endif
 
   v8::Global<v8::Value> web_contents_;
-  api::WebContents* api_web_contents_;
+  base::WeakPtr<api::WebContents> api_web_contents_;
 
   base::WeakPtrFactory<BrowserWindow> weak_factory_;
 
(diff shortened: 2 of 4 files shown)


miniak:
these tests are failing for me locally even without this change
```
not ok 117 BrowserWindow module window.close() should emit unload event
  Error: ENOENT: no such file or directory, open '/Users/miburda/Development/github/electron-gn/src/electron/spec/fixtures/api/close'
      at Object.openSync (fs.js:453:3)
      at Object.func [as openSync] (electron/js2c/asar.js:155:31)
      at Object.readFileSync (fs.js:353:35)
      at Object.fs.readFileSync (electron/js2c/asar.js:557:40)
      at Context.<anonymous> (electron/spec-main/api-browser-window-spec.ts:134:26)
not ok 118 BrowserWindow module window.close() should emit beforeunload event
  Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/miburda/Development/github/electron-gn/src/electron/spec-main/api-browser-window-spec.ts)
      at listOnTimeout (internal/timers.js:535:17)
      at processTimers (internal/timers.js:479:7)
```

zcbenz:
I believe this can also fix https://github.com/electron/electron/issues/16580.

miniak:
@zcbenz can you please review again? I think it's cleaner to make `api_web_contents_` a weak pointer.

atom/browser/api/atom_api_browser_window.cc
```diff
@@ -67,7 +67,7 @@ BrowserWindow::BrowserWindow(v8::Isolate* isolate,
   }
 
   web_contents_.Reset(isolate, web_contents.ToV8());
-  api_web_contents_ = web_contents.get();
+  api_web_contents_ = web_contents->GetWeakPtr();
```
deepak1556: If its a weak ptr, should null check in other places of access too.

miniak:
@deepak1556 This bug seems to manifest not only when the `webContents` is destroyed explicitly. See #16580 mentioned by @zcbenz

deepak1556:
> Converting a raw ptr to weak ptr should not change the relationship, or I am missing something?

weakptr has this implicit notion, that objects accessing it doesn't care about the lifetime and can handle them safely. Most useful when posting tasks.

The raw ptr here , was to explicitly state that `webContents` will outlive the `BrowserWindow` so its safe to handle it. But thats kinda broken now, while I understand the problem this PR is trying to solve. Just want to make sure the lifetime is handled correctly.

@zcbenz has worked recently on refactoring the destruction sequence, would like his opinion before merging.

zcbenz:
The lifetime management of WebContents is a mess, even after I tried to clean it up. For https://github.com/electron/electron/issues/16580, I encountered it occasionally when running small tests, I believe we have code in Electron that incorrectly destroys WebContents before destroying BrowserWindow, but I could not find it.

I would say this PR is a hack rather than a proper fix, but before we find one, I think this can be accepted as a special hack to fix crash when destroying window.

We should probably add a FIXME comment in `~BrowserWindow` explaining it is a hack that should be revisited in future.

deepak1556:
@miniak can you update following cheng's comment, I can approve it after that. Thanks!

miniak:
@deepak1556 FIXME comment added

atom/browser/api/atom_api_browser_window.cc
```diff
@@ -93,7 +93,9 @@ BrowserWindow::BrowserWindow(v8::Isolate* isolate,
 }
 
 BrowserWindow::~BrowserWindow() {
-  api_web_contents_->RemoveObserver(this);
+  // FIXME This is a hack rather than a proper fix preventing shutdown crashes.
```
nornagon: What's the long-term plan for fixing this?

PR #18688, opened 2019-06-07
chore: convert extension apis to TypeScript

codebytere:
#### Description of Change

Converts extensions-related files to TS.

cc @MarshallOfSound 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


lib/renderer/extensions/event.ts
```diff
@@ -1,26 +1,22 @@
...
-  removeListener (callback) {
+  removeListener (callback: Function) {
     const index = this.listeners.indexOf(callback)
     if (index !== -1) {
       this.listeners.splice(index, 1)
     }
   }
 
-  emit (...args) {
+  emit (...args: any) {
```
MarshallOfSound:
```suggestion
  emit (...args: any[]) {
```

lib/renderer/extensions/event.ts
```diff
@@ -1,26 +1,22 @@
...
     for (const listener of this.listeners) {
       listener(...args)
     }
   }
 }
 
-module.exports = Event
+export default Event
```
MarshallOfSound: Can we prefer named exports for things not in `*/api/*`.  E.g. this should be `export class Event`

lib/renderer/extensions/i18n.ts
```diff
@@ -1,59 +1,58 @@
...
+  content: string;
+  example?: string;
+}
 
-const getMessages = (extensionId) => {
+const getMessages = (extensionId: number) => {
   try {
     const data = ipcRendererUtils.invokeSync('CHROME_GET_MESSAGES', extensionId)
-    return JSON.parse(data) || {}
+    return JSON.parse(data as any) || {}
```
MarshallOfSound:
```suggestion
    return JSON.parse(data) || {}
```

lib/renderer/extensions/i18n.ts
```diff
@@ -1,59 +1,58 @@
...
+
+interface Placeholder {
+  content: string;
+  example?: string;
+}
 
-const getMessages = (extensionId) => {
+const getMessages = (extensionId: number) => {
   try {
     const data = ipcRendererUtils.invokeSync('CHROME_GET_MESSAGES', extensionId)
```
MarshallOfSound:
```suggestion
    const data = ipcRendererUtils.invokeSync<string>('CHROME_GET_MESSAGES', extensionId)
```

lib/renderer/extensions/i18n.ts
```diff
@@ -1,59 +1,58 @@
...
+const replacePlaceholders = (message: string, placeholders: Record<string, Placeholder>, substitutions: string[] | string) => {
+  if (typeof substitutions === 'string') substitutions = [substitutions]
+  if (!Array.isArray(substitutions)) substitutions = []
 
   if (placeholders) {
-    Object.keys(placeholders).forEach((name) => {
+    Object.keys(placeholders).forEach((name: string) => {
       let { content } = placeholders[name]
-      content = replaceNumberedSubstitutions(content, substitutions)
+      content = replaceNumberedSubstitutions(content, substitutions as string[])
```
MarshallOfSound: This `as string[]` shouldn't be required, it should be known at this point that `substitutions` is definitely an array.🤔 In order to prove it you may need to make a `const` redef

```js
const substitutionsArray = Array.isArray(substitutions) ? substitutions : []
```

Which is const and of type `string[]`

lib/renderer/extensions/storage.ts
```diff
@@ -1,12 +1,10 @@
-'use strict'
+import * as ipcRendererUtils from '@electron/internal/renderer/ipc-renderer-internal-utils'
 
-const ipcRendererUtils = require('@electron/internal/renderer/ipc-renderer-internal-utils')
-
-const getStorage = (storageType, extensionId, callback) => {
+const getStorage = (storageType: string, extensionId: number, callback: Function) => {
   if (typeof callback !== 'function') throw new TypeError('No callback provided')
 
   ipcRendererUtils.invoke('CHROME_STORAGE_READ', storageType, extensionId)
```
MarshallOfSound:
```suggestion
  ipcRendererUtils.invoke<string>('CHROME_STORAGE_READ', storageType, extensionId)
```

lib/renderer/extensions/storage.ts
```diff
@@ -1,12 +1,10 @@
...
-    .then(data => {
+    .then((data: any) => {
```
MarshallOfSound:
```suggestion
    .then((data) => {
```

lib/renderer/extensions/storage.ts
```diff
@@ -58,37 +56,32 @@ const getStorageManager = (storageType, extensionId) => {
...
     },
 
-    clear (callback) {
+    clear (callback: Function) {
       setStorage(storageType, extensionId, {}, callback)
     }
   }
 }
 
 module.exports = {
```
MarshallOfSound: This should become an `export const setup =`

lib/renderer/extensions/web-navigation.ts
```diff
@@ -1,18 +1,16 @@
...
 class WebNavigation {
-  constructor () {
-    this.onBeforeNavigate = new Event()
-    this.onCompleted = new Event()
+  private onBeforeNavigate = new Event()
+  private onCompleted = new Event()
 
-    ipcRendererInternal.on('CHROME_WEBNAVIGATION_ONBEFORENAVIGATE', (event, details) => {
+  constructor () {
+    ipcRendererInternal.on('CHROME_WEBNAVIGATION_ONBEFORENAVIGATE', (event: Event, details: any) => {
```
MarshallOfSound: This is gonna use the wrong `Event` so I'm surprised TS allowed it.  You shouldn't have to define the types here, TS should infer them from `electron.d.ts`
```suggestion
    ipcRendererInternal.on('CHROME_WEBNAVIGATION_ONBEFORENAVIGATE', (event, details) => {
```

MarshallOfSound: Same below

codebytere: I assumed this too but it blows up with 

```
details has implicit type 'any'
```
etc

lib/renderer/extensions/web-navigation.ts
```diff
@@ -1,23 +1,20 @@
...
+    ipcRendererInternal.on('CHROME_WEBNAVIGATION_ONCOMPLETED', (event: IpcMainEvent, details: any) => {
       this.onCompleted.emit(details)
     })
   }
 }
 
-exports.setup = () => {
-  return new WebNavigation()
-}
+exports.setup = () => new WebNavigation()
```
MarshallOfSound:
```suggestion
export const setup = () => new WebNavigation()
```

I would also question this method, seems a bit useless, why not just export the `class` 😆 (I know it was like this before but hey)

PR #18706, opened 2019-06-10
feat: migrate protocol module to NetworkService (Part 11)

zcbenz:
#### Description of Change

The `registerHttpProtocol` API can make arbitrary requests regardless of the CORS settings. To keep this behavior, we need to explicitly download data from URL and then pipe it to NetworkService, like what we did for other types of protocol types.

This PR implements the API by adding a `URLPipeLoader` class that pipes the data from URL to NetworkService, instead of creating a new loader which was essentially a redirection.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/spec/chromium-spec.js
+++ b/spec/chromium-spec.js
@@ -86,20 +86,6 @@ describe('chromium feature', () => {
     })
   })
 
-  describe('sending request of http protocol urls', () => {
-    it('does not crash', (done) => {
-      const server = http.createServer((req, res) => {
-        res.end()
-        server.close()
-        done()
-      })
-      server.listen(0, '127.0.0.1', () => {
-        const port = server.address().port
-        $.get(`http://127.0.0.1:${port}`)
-      })
-    })
-  })
-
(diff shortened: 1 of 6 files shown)


spec/chromium-spec.js
```diff
@@ -86,20 +86,6 @@ describe('chromium feature', () => {
     })
   })
 
-  describe('sending request of http protocol urls', () => {
```
deepak1556: why delete this test ?

deepak1556:
As a follow up , I am thinking this behavior should be made opt in, the default behavior of custom schemes should be secure, act more like `chrome-extension`. But this will be a different effort that aligns with making `file` scheme less privileged.

spec/chromium-spec.js
zcbenz: This test is essentially sending CORS requests in `file://` and the request always fails, the test passes by accident. I found it when testing http protocol code.

The thing it wants to test (that sending http request should not crash) is already covered by other tests, we don't need to keep this invalid test.

PR #18718, opened 2019-06-10
feat: add creationTime / sandboxed / integrityLevel to app.getAppMetrics()

miniak:
#### Description of Change
This is useful for checking which processes are sandboxed on OS level.

Regarding `creationTime`, since the `pid` can be reused after a process dies, it is useful to use both the `pid` and the `creationTime` to uniquely identify a process.

Required for tests in #18650

/cc @MarshallOfSound, @deepak1556

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `creationTime` / `sandboxed` / `integrityLevel` to `app.getAppMetrics()` output.

### DIFF
--- a/atom/browser/api/atom_api_app.cc
+++ b/atom/browser/api/atom_api_app.cc
@@ -70,6 +70,25 @@ namespace mate {
 
 #if defined(OS_WIN)
 template <>
+struct Converter<atom::ProcessIntegrityLevel> {
+  static v8::Local<v8::Value> ToV8(v8::Isolate* isolate,
+                                   atom::ProcessIntegrityLevel value) {
+    switch (value) {
+      case atom::ProcessIntegrityLevel::Untrusted:
+        return mate::StringToV8(isolate, "untrusted");
+      case atom::ProcessIntegrityLevel::Low:
+        return mate::StringToV8(isolate, "low");
+      case atom::ProcessIntegrityLevel::Medium:
+        return mate::StringToV8(isolate, "medium");
+      case atom::ProcessIntegrityLevel::High:
+        return mate::StringToV8(isolate, "high");
+      default:
+        return mate::StringToV8(isolate, "unknown");
+    }
+  }
+};
+
+template <>
 struct Converter<Browser::UserTask> {
   static bool FromV8(v8::Isolate* isolate,
                      v8::Local<v8::Value> val,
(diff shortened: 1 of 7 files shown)


atom/browser/api/atom_api_app.cc
```diff
@@ -371,17 +400,6 @@ namespace api {
 
 namespace {
 
-class AppIdProcessIterator : public base::ProcessIterator {
```
miniak: this is dead code

atom/browser/api/atom_api_app.cc
```diff
@@ -1215,9 +1211,21 @@ std::vector<mate::Dictionary> App::GetAppMetrics(v8::Isolate* isolate) {
...
+                 process_metric.second->process.CreationTime().ToJsTime());
+
+#if defined(OS_MACOSX)
+    pid_dict.Set("sandboxed", process_metric.second->IsSandboxed());
+#elif defined(OS_WIN)
+    auto integrity_level = process_metric.second->GetIntegrityLevel();
+    auto sandboxed = ProcessMetric::IsSandboxed(integrity_level);
+    pid_dict.Set("integrityLevel", integrity_level);
+    pid_dict.Set("sandboxed", sandboxed);
+#endif
```
MarshallOfSound: On linux this should be set to `unknown`

docs/api/structures/process-metric.md
```diff
@@ -3,3 +3,12 @@
 * `pid` Integer - Process id of the process.
 * `type` String - Process type (Browser or Tab or GPU etc).
 * `cpu` [CPUUsage](cpu-usage.md) - CPU usage of the process.
+* `creationTime` Number - Creation time for this process. Since the `pid` can be reused after a process dies,
```
zcbenz: The document should clarify the time is whether seconds or ms.

deepak1556:
I think this could add support for linux by reading the value of `service_manager::ZygoteHostImpl::GetInstance()->GetRendererSandboxStatus` but its not useful at the moment, having the zygote enabled essentially means some sort of sandbox enabled, maybe helpful in knowing what type of sandbox is enabled similar to https://cs.chromium.org/chromium/src/chrome/browser/ui/webui/sandbox_internals_ui.cc?g=0&l=31-61 

But to make this return useful value in mixed sandbox mode were we bypass zygote, maybe check for `switches::kNoZygote` ?

/cc @nornagon

nornagon:
Hmm, I don't think sandbox status of the zygote tells you whether or not the process is sandboxed, because the way we implement mixed-sandbox mode in linux is that sandboxed processes are launched through the zygote and unsandboxed processes are launched directly.

Issue #18708, opened 2019-06-10
in-app-purchase / product - Miss match between documentation/types to result


oriSomething:
`Product` of (`InAppPurchase`) - Miss match between documentation/types to result of the key `downloadable`.

### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
`v5.0.3`

* **Operating System:**
macOS 10.14.15

* **Last Known Working Electron version:**
N/A

### Expected Behavior

The `Product` structure of returned data from `inAppPurchase.getProducts()` should contain `downloadable` key as in the [docs / TS type definitions](https://github.com/electron/electron/blob/master/docs/api/structures/product.md)

### Actual Behavior

The `Product` contains `isDownloadable` key (which isn't documented).

### Screenshots

![Screen Shot 2019-06-10 at 11 54 30](https://user-images.githubusercontent.com/2685242/59184392-b58f9b00-8b76-11e9-99a0-58fb7eead403.png)

### Additional Information

I think the issue related to [this line](https://github.com/electron/electron/blob/3e5a98b5f43fe70961eb6452dc8837453c6de1a0/atom/browser/api/atom_api_in_app_purchase.cc#L64)

PR #18720, opened 2019-06-10
fix: correct isDownloadable docs to match impl

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/18708.

Corrects product documentation  to match implementation, since the actual instance property for SKProducts is [`isDownloadable`](https://developer.apple.com/documentation/storekit/skproduct/1506161-isdownloadable?language=objc). Also corrects product object nomenclature for consistency and improves the value information a tad.

cc @MarshallOfSound @nornagon @miniak 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/atom/browser/api/atom_api_in_app_purchase.cc
+++ b/atom/browser/api/atom_api_in_app_purchase.cc
@@ -61,7 +61,7 @@ struct Converter<in_app_purchase::Product> {
     dict.Set("formattedPrice", val.formattedPrice);
 
     // Downloadable Content Information
-    dict.Set("isDownloadable", val.downloadable);
+    dict.Set("isDownloadable", val.isDownloadable);
 
     return dict.GetHandle();
   }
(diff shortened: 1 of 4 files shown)


MarshallOfSound:
Fun question.

(a) How long has this been broken.
(b) Should we update the documentation to match what people might have been using or fix the impl to match documentation 🤔

codebytere:
🤔 excellent question...blame says since always.

in that event, perhaps better to correct docs instead?

MarshallOfSound:
@codebytere I'd prefer to update the docs in this case 👍

codebytere:
Linux failures are [flakes](https://circleci.com/gh/electron/electron/224586?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link), they both pertain to `emit` events with no codepath correlation to the changes in this PR.

Issue #12437, opened 2018-03-26
Support accessory view option in `dialog.showMessageBox` and `dialog.showErrorBox`


sindresorhus:
* Electron version: 2.0.0-beta.5
* Operating system: macOS 10.13.3

### Expected behavior

I expected `dialog.showMessageBox` and `dialog.showErrorBox` to have an option to specify an accessory view. In the native world, you can render anything inside it, but I'd be happy to just have a multi-line text field.

My specific use-case is that I need to be able to display an error stack trace in the dialog, for [`electron-unhandled`](https://github.com/sindresorhus/electron-unhandled) which currently looks like:

<img width="532" alt="screen shot 2018-03-26 at 21 45 07" src="https://user-images.githubusercontent.com/170270/37913395-fb0c3546-313e-11e8-83cc-c506cc987754.png">

But I would like something like this:

![image](https://user-images.githubusercontent.com/170270/37913320-c9e0d65c-313e-11e8-86e4-49a8455fb089.png)

MarshallOfSound:
@sindresorhus Is there an equivalent pattern on windows?  This seems like a very specialized use case that would probably benefit more from being a dedicated native module than adding more complexity to our already non-trivial dialog API.  🤔

sindresorhus:
I'm not familiar with Windows, but from what I could find, it looks like it, yes: https://msdn.microsoft.com/en-us/library/windows/desktop/ms644995(v=vs.85).aspx#edit

Although, I'd be fine with having this as a macOS only thing.

---

> This seems like a very specialized use case that would probably benefit more from being a dedicated native module than adding more complexity to our already non-trivial dialog API.

This wouldn't add much complexity. It's just setting a `NSTextView` as the `accessoryView` with the given text: https://developer.apple.com/documentation/appkit/nsalert/1530575-accessoryview?language=objc

MarshallOfSound:
> Although, I'd be fine with having this as a macOS only thing.

Can't speak for anyone else, but personally, I want as much as what Electron does to be 100% cross platform.  I.e. No little OS switches, no platform conditions, just one set of code and it Just Works(tm).  Adding portals to OS-specific features such as this accessory view go in the opposite direction.

> This wouldn't add much complexity

In terms of implementing it stand alone, no, but adding it into options parsing + OS specific interactions it takes a toll and I'm not convinced it's a common enough use case to warrant more code to maintain and test.

IMO this should only really be a thing if we can come up with an easy and consistent way to use it cross-platform.

[...]

PR #18732, opened 2019-06-11
refactor: pass MessageBox params as a struct

codebytere:
#### Description of Change

This work is intended as a precursor to the future implementation of https://github.com/electron/electron/issues/12437.

Historically, we've been passing in all MessageBox parameters individually, which makes augmenting or improving MessageBox functionality challenging because to change or add even one argument requires a huge cascade of argument changes that leaves room for errors.

For other file dialog related APIs, we use a struct (`DialogSettings`), and so this PR takes a similar approach and refactors MessageBox parameters into a struct (`MessageBoxSettings`) which we then use to simplify argument passing and which will enable us to more quickly iterate and improve upon functionality in the future.

cc @nornagon @MarshallOfSound @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/atom/browser/atom_javascript_dialog_manager.cc
+++ b/atom/browser/atom_javascript_dialog_manager.cc
@@ -93,12 +93,9 @@ void AtomJavaScriptDialogManager::RunJavaScriptDialog(
 
   atom::MessageBoxSettings settings;
   settings.parent_window = window;
-  settings.type = 0;  // MessageBoxType::kNone
   settings.buttons = buttons;
   settings.default_id = default_id;
   settings.cancel_id = cancel_id;
-  settings.icon = gfx::ImageSkia();
-  settings.options = atom::MessageBoxOptions::MESSAGE_BOX_NONE;
   settings.message = base::UTF16ToUTF8(message_text);
 
   atom::ShowMessageBox(
--- a/atom/browser/atom_javascript_dialog_manager.cc
+++ b/atom/browser/atom_javascript_dialog_manager.cc
@@ -91,10 +91,18 @@ void AtomJavaScriptDialogManager::RunJavaScriptDialog(
       window = relay->GetNativeWindow();
   }
 
+  atom::MessageBoxSettings settings;
+  settings.parent_window = window;
+  settings.type = 0;  // MessageBoxType::kNone
+  settings.buttons = buttons;
+  settings.default_id = default_id;
+  settings.cancel_id = cancel_id;
+  settings.icon = gfx::ImageSkia();
+  settings.options = atom::MessageBoxOptions::MESSAGE_BOX_NONE;
+  settings.message = base::UTF16ToUTF8(message_text);
+
   atom::ShowMessageBox(
-      window, atom::MessageBoxType::kNone, buttons, default_id, cancel_id,
-      atom::MessageBoxOptions::MESSAGE_BOX_NONE, "",
-      base::UTF16ToUTF8(message_text), "", checkbox, false, gfx::ImageSkia(),
+      settings,
       base::BindOnce(&AtomJavaScriptDialogManager::OnMessageBoxCallback,
                      base::Unretained(this), base::Passed(std::move(callback)),
                      origin));
--- a/atom/browser/ui/message_box.h
+++ b/atom/browser/ui/message_box.h
@@ -35,11 +35,11 @@ struct MessageBoxSettings {
   std::vector<std::string> buttons;
   int default_id;
   int cancel_id;
-  int options;
-  std::string title = "";
-  std::string message = "";
-  std::string detail = "";
-  std::string checkbox_label = "";
+  int options = 0;
+  std::string title;
+  std::string message;
+  std::string detail;
+  std::string checkbox_label;
   bool checkbox_checked = false;
   gfx::ImageSkia icon;
 
--- a/atom/browser/ui/message_box.h
+++ b/atom/browser/ui/message_box.h
@@ -31,11 +31,11 @@ enum MessageBoxOptions {
 
 struct MessageBoxSettings {
   atom::NativeWindow* parent_window = nullptr;
-  int type;
+  MessageBoxType type = atom::MessageBoxType::kNone;
   std::vector<std::string> buttons;
   int default_id;
   int cancel_id;
-  int options = 0;
+  int options = atom::MessageBoxOptions::MESSAGE_BOX_NONE;
   std::string title;
   std::string message;
   std::string detail;
--- a/atom/browser/ui/message_box.h
+++ b/atom/browser/ui/message_box.h
@@ -10,10 +10,7 @@
 
 #include "base/callback_forward.h"
 #include "base/strings/string16.h"
-
-namespace gfx {
-class ImageSkia;
-}
+#include "ui/gfx/image/image_skia.h"
 
 namespace atom {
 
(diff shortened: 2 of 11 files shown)


atom/browser/ui/message_box_gtk.cc
```diff
@@ -200,46 +197,26 @@ void GtkMessageBox::OnCheckboxToggled(GtkWidget* widget) {
...
 
 void ShowErrorBox(const base::string16& title, const base::string16& content) {
   if (Browser::Get()->is_ready()) {
-    GtkMessageBox(nullptr, MessageBoxType::kError, {"OK"}, -1, 0, "Error",
-                  base::UTF16ToUTF8(title).c_str(),
-                  base::UTF16ToUTF8(content).c_str(), "", false,
-                  gfx::ImageSkia())
-        .RunSynchronous();
+    atom::MessageBoxSettings settings;
+    settings.type = 3;  // MessageBoxType::kError
```
ckerr: Why not `settings.type = MessageboxType::kError`?

If that doesn't work and `type` _has_ to be an int, a backup plan would be to cast MessageBoxType::kError to an int here

Either way we should not be hardcoding the number

atom/browser/ui/message_box_gtk.cc
```diff
@@ -200,46 +197,26 @@ void GtkMessageBox::OnCheckboxToggled(GtkWidget* widget) {
...
+    settings.buttons = {"OK"};
+    settings.title = "Error";
+    settings.message = base::UTF16ToUTF8(title).c_str();
+    settings.detail = base::UTF16ToUTF8(content).c_str();
```
ckerr: `base::UTF16ToUTF8(foo)` returns a std::string, so we should be able to take it directly, e.g.

```
settings.message = base::UTF16ToUTF8(title);
settings.detail = base::UTF16ToUTF8(content);
```

atom/browser/ui/message_box_gtk.cc
```diff
@@ -200,46 +197,26 @@ void GtkMessageBox::OnCheckboxToggled(GtkWidget* widget) {
...
+    settings.icon = gfx::ImageSkia();
```
ckerr: `settings.icon` is already an empty `gfx::ImageSkia` so this line can be removed.

atom/browser/ui/message_box_win.cc
```diff
@@ -183,98 +187,57 @@ int ShowTaskDialogUTF16(NativeWindow* parent,
...
   std::vector<base::string16> utf16_buttons;
-  for (const auto& button : buttons)
+  for (const auto& button : settings.buttons)
     utf16_buttons.push_back(base::UTF8ToUTF16(button));
 
+  base::string16 title_16 = base::UTF8ToUTF16(settings.title);
+  base::string16 message_16 = base::UTF8ToUTF16(settings.message);
+  base::string16 detail_16 = base::UTF8ToUTF16(settings.detail);
+  base::string16 checkbox_label_16 = base::UTF8ToUTF16(settings.checkbox_label);
+  bool checkbox_checked16 = settings.checkbox_checked;
```
ckerr: _(nit)_ all the above can be `const`

_(nit)_ the variable name "checkbox_checked16" is fun. Converting a bool to a 16-bit bool is an interesting choice :neutral_face:

atom/common/native_mate_converters/message_box_converter.cc
```diff
@@ -0,0 +1,36 @@
...
+  dict.Get("buttons", &(out->buttons));
+  dict.Get("defaultId", &(out->default_id));
+  dict.Get("cancelId", &(out->cancel_id));
+  dict.Get("options", &(out->options));
+  dict.Get("title", &(out->title));
+  dict.Get("message", &(out->message));
+  dict.Get("detail", &(out->detail));
+  dict.Get("checkboxLabel", &(out->checkbox_label));
+  dict.Get("checkboxChecked", &(out->checkbox_checked));
+  dict.Get("icon", &(out->icon));
```
ckerr: _nit_: The parens are unnecessary here. `&out->icon` is fine

atom/browser/ui/message_box.h
```diff
@@ -32,32 +29,31 @@ enum MessageBoxOptions {
...
+  atom::NativeWindow* parent_window = nullptr;
+  int type;
+  std::vector<std::string> buttons;
+  int default_id;
+  int cancel_id;
+  int options;
+  std::string title;
+  std::string message;
+  std::string detail;
+  std::string checkbox_label = "";
```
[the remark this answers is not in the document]
codebytere: (resolved IRL, not needed!)

atom/browser/ui/message_box.h
```diff
@@ -32,32 +29,31 @@ enum MessageBoxOptions {
...
-                       int options,
-                       const std::string& title,
-                       const std::string& message,
-                       const std::string& detail,
-                       const gfx::ImageSkia& icon);
+struct MessageBoxSettings {
+  atom::NativeWindow* parent_window = nullptr;
+  int type;
+  std::vector<std::string> buttons;
+  int default_id;
```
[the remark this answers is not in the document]
codebytere: We handle this in custom fashion on macOS at least, so i think we may want to leave this as-is 🤔 @MarshallOfSound you may know more?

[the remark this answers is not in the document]
MarshallOfSound: A particular default chosen will be chosen in JS land, this should always receive a value and shouldn't need a default.  However the converter should throw / fail if this value is not provided.

atom/common/native_mate_converters/message_box_converter.cc
```diff
@@ -0,0 +1,36 @@
...
+#include "atom/common/native_mate_converters/message_box_converter.h"
+
+#include "atom/browser/api/atom_api_browser_window.h"
+#include "atom/common/native_mate_converters/file_path_converter.h"
+#include "atom/common/native_mate_converters/image_converter.h"
+#include "native_mate/dictionary.h"
+
+namespace mate {
+
+bool Converter<atom::MessageBoxSettings>::FromV8(
```
MarshallOfSound: Some of the values here are 100% required and should never be undefined or empty.  The converter should fail for instance if `defaultId` and `cancelId` are not set.  Probably things like `title` as well.

atom/browser/ui/message_box.h
```diff
@@ -32,32 +29,31 @@ enum MessageBoxOptions {
...
-                       int default_id,
-                       int cancel_id,
-                       int options,
-                       const std::string& title,
-                       const std::string& message,
-                       const std::string& detail,
-                       const gfx::ImageSkia& icon);
+struct MessageBoxSettings {
+  atom::NativeWindow* parent_window = nullptr;
+  int type;
```
miniak: why `int`?

atom/browser/ui/message_box_gtk.cc
miniak: shouldn't `type` be `MessageboxType`?

codebytere: done

atom/browser/ui/message_box.h
```diff
@@ -32,32 +29,31 @@ enum MessageBoxOptions {
...
-                       int default_id,
-                       int cancel_id,
-                       int options,
-                       const std::string& title,
-                       const std::string& message,
-                       const std::string& detail,
-                       const gfx::ImageSkia& icon);
+struct MessageBoxSettings {
+  atom::NativeWindow* parent_window = nullptr;
+  MessageBoxType type;
```
miniak: `= atom::MessageBoxType::kNone`?

atom/browser/atom_javascript_dialog_manager.cc
```diff
@@ -91,10 +91,18 @@ void AtomJavaScriptDialogManager::RunJavaScriptDialog(
       window = relay->GetNativeWindow();
   }
 
+  atom::MessageBoxSettings settings;
+  settings.parent_window = window;
+  settings.type = atom::MessageBoxType::kNone;
+  settings.buttons = buttons;
+  settings.default_id = default_id;
+  settings.cancel_id = cancel_id;
+  settings.icon = gfx::ImageSkia();
```
miniak: this is not necessary

atom/browser/atom_javascript_dialog_manager.cc
```diff
@@ -91,10 +91,18 @@ void AtomJavaScriptDialogManager::RunJavaScriptDialog(
       window = relay->GetNativeWindow();
   }
 
+  atom::MessageBoxSettings settings;
+  settings.parent_window = window;
+  settings.type = atom::MessageBoxType::kNone;
```
miniak: I would use `atom::MessageBoxType::kNone` as default value in the struct instead

atom/browser/ui/message_box.h
```diff
@@ -32,32 +29,31 @@ enum MessageBoxOptions {
...
+  std::vector<std::string> buttons;
+  int default_id;
+  int cancel_id;
+  int options = 0;
```
miniak: = `atom::MessageBoxOptions::MESSAGE_BOX_NONE`?

atom/browser/atom_javascript_dialog_manager.cc
```diff
@@ -91,10 +91,18 @@ void AtomJavaScriptDialogManager::RunJavaScriptDialog(
...
+  settings.buttons = buttons;
+  settings.default_id = default_id;
+  settings.cancel_id = cancel_id;
+  settings.icon = gfx::ImageSkia();
+  settings.options = atom::MessageBoxOptions::MESSAGE_BOX_NONE;
```
miniak: this is not necessary, we already set it to 0 by default

PR #18744, opened 2019-06-12
docs: reorganize application distribution links in table of contents

malept:
#### Description of Change

* Fix the title/link for "supported platforms", which got lost in a previous TOC reorg.
* Move "Code signing" under "Application Distribution", because it's unclear what the difference between "Packaging" and "Distribution" is, and also this was the only item under "Packaging".

CC: @electron/wg-docs-tools 

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -66,7 +66,7 @@ an issue:
 * Packaging
   * [Code Signing](tutorial/code-signing.md)
 * [Distribution](tutorial/application-distribution.md)
-  * [Support](tutorial/support.md)
+  * [Supported Platforms](tutorial/support.md#supported-platforms)
   * [Mac App Store](tutorial/mac-app-store-submission-guide.md)
   * [Windows Store](tutorial/windows-store-guide.md)
   * [Snapcraft](tutorial/snapcraft.md)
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -63,10 +63,9 @@ an issue:
   * [Testing on Headless CI Systems (Travis, Jenkins)](tutorial/testing-on-headless-ci.md)
   * [DevTools Extension](tutorial/devtools-extension.md)
   * [Automated Testing with a Custom Driver](tutorial/automated-testing-with-a-custom-driver.md)
-* Packaging
-  * [Code Signing](tutorial/code-signing.md)
 * [Distribution](tutorial/application-distribution.md)
-  * [Support](tutorial/support.md)
+  * [Supported Platforms](tutorial/support.md#supported-platforms)
```
BinaryMuse: Hm, this document contains more info than just the supported platforms and versions, it also contains info on how to get actual support. 🤔

BinaryMuse: Ah ever mind, it's listed twice in the TOC.

malept: Yeah... alternatively, we could just remove this link.

ckerr: I'm OK with having it in there twice if the url links have different query sections. It wouldn't surprise me if people on this page were looking for info on supported platforms.

PR #18751, opened 2019-06-12
test: add test for invalid cookie url

DeerMichel:
#### Description of Change

Added a test for invalid cookie urls. Added a notice to docs that an error for invalid urls will be thrown.

cc @codebytere @ckerr 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


spec-main/api-session-spec.js
```diff
@@ -87,15 +87,26 @@ describe('session module', () => {
...
-      await expect((async () => {
-        const { cookies } = session.defaultSession
-        const name = '1'
-        const value = '1'
+    it('yields an error when setting a cookie with missing required fields', () => {
+      const { cookies } = session.defaultSession
+      const name = '1'
+      const value = '1'
+
+      expect((async () => {
```
nornagon: you need to `await` the result of this `expect`, otherwise the test will complete without waiting for the result of the assert. (which means the test function will need to become `async` again, too)
```suggestion
      await expect((async () => {
```

spec-main/api-session-spec.js
```diff
@@ -87,15 +87,26 @@ describe('session module', () => {
...
         await cookies.set({ url: '', name, value })
       })()).to.eventually.be.rejectedWith('Failed to get cookie domain')
     })
 
+    it('yields an error when setting a cookie with an invalid domain', () => {
```
nornagon:
```suggestion
    it('yields an error when setting a cookie with an invalid URL', () => {
```

spec-main/api-session-spec.js
```diff
@@ -87,15 +87,26 @@ describe('session module', () => {
...
+      const { cookies } = session.defaultSession
+      const name = '1'
+      const value = '1'
+
+      expect((async () => {
```
nornagon: `await` here too

docs/api/cookies.md
```diff
@@ -84,7 +84,7 @@ the response.
 #### `cookies.set(details)`
 
 * `details` Object
-  * `url` String - The url to associate the cookie with.
+  * `url` String - The url to associate the cookie with. An error is thrown if url is invalid.
```
nornagon:
```suggestion
  * `url` String - The url to associate the cookie with. An error is thrown if the url is invalid.
```

spec-main/api-session-spec.js
codebytere: @nornagon i'm pretty sure we can still remove an excess async by moving the primary one to the top-level like so?
```
it('yields an error when setting a cookie with missing required fields', async () => {
   const { cookies } = session.defaultSession
   const name = '1'
   const value = '1'
   expect(() => {
        await cookies.set({ url: '', name, value })
    }).to.eventually.be.rejectedWith('Failed to get cookie domain')
})
```

nornagon: Actually you could condense this down a little, because in general, 
```js
(async () => await foo(...))
// is equivalent to
(foo(...))
```
since `foo()` just returns a promise.

so in this case, we can write:

```js
await expect(cookies.set({...})).to.eventually.be.rejectedWith(...)
```

DeerMichel:
@nornagon hmm so I don't really get behind the difference of that - but I followed your suggestion 👍

nornagon:
This might help for understanding async/await: https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await

Essentially, `await` waits for the result of a promise—just like calling `.then()` on it—and `async function` makes a function that returns a promise.

e.g.
```js
$ node
> async function foo() { return 'hello' }
> foo()
Promise { ... }
> foo().then(result => console.log(result))
hello
```

We could fake our own `async` by creating a Promise, e.g. this `foo` would be the same as our async function `foo` above:

```js
function foo() {
    resolve('hello')
  })
}
```

We can also fake our own `await` by using `.then()`, e.g. these two functions do the same thing:
```js
async function usesAwait() {
  console.log("before")
  await someFunctionThatReturnsAPromise()
  console.log("after")
}

function usesThen() {
  console.log("before")
  someFunctionThatReturnsAPromise().then(() => {
    console.log("after")
  })
}
```

so back to the example I gave in the code review comment,

```js
(async () => await foo())
```
is basically
```js
function () {
    foo().then(resolve)
  })
}
```
which is needlessly wrapping `foo()` in another layer of promises. It's the asynchronous equivalent of how `x = (() => someFunction())()` would be the same as `x = someFunction()`.

Sorry, that came out as a lot of text, hope it's at all helpful 😅

nornagon:
I think all those test failures are flake 😅

PR #13869, opened 2018-07-30
fix: use OOPIF for webview tag

zcbenz:
Close https://github.com/electron/electron/issues/13503.
Close #12291.

notes: Migrate `<webview>` to use out-of-process iframe.

PR #13835, opened 2018-07-27
docs: Warn about the use of webviews

felixrieseberg:
In person today, we've discussed expressing our general feelings about the `webview` tag (which I'd describe as 😬) in documentation – Electron "insiders" and maintainers have known for quite some time that the tag isn't as stable as it should be (and that it involves plenty of subtle bugs).

One of the outcomes was that we should share those feelings and difficulties in our docs. This PR takes a first stab at that.

*This PR does not deprecate them (just yet), it just warns about them*

##### Checklist
- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

cc @codebytere

Issue #3675, opened 2015-12-03
What is the right way of calling node.js modules from inside iframe?


ValeryEliseev:
Question:
What is the right way of calling node.js modules from inside iframe?

The core of the problem:
In our big browser-oriented web-application we use "dhtmlx" javascript library, wich draws its own "windows" on the main page using iframe, thus imitating MDI interface. Every such a window is a separate html page, with its own purpose.
We are going to create an Electron-based application having more features than the browser-based one. We need to call node.js modules both from the main page (which is loaded in Electron) and from the pages that reside in several iframes located on the main page.
We can't use webview, because iframes are being created dynamically by a third-party library.
Simple call to "require" from the page loaded into iframe doesn't work.

Let's use a simple "default_app"-based example from Electron distribution to demonstrate the problem.

First we create a two-lines file "node-call.js" in "resources/default_app" folder:

os = require('os');
document.write(os.type());

Next we create another one-line file "iframe-content.html" in the same folder:

&lt;html&gt;&lt;script src='node-call.js'&gt;&lt;/script&gt;&lt;/html&gt;

Then we add 2 lines into "default_app/index.html" file just before the closing &lt;/body&gt; tag:

&lt;script src='node-call.js'&gt;&lt;/script&gt;
&lt;iframe src='iframe-content.html'&gt;&lt;iframe&gt;

Let's start Electron. First output of document.write works, so we see our OS type on the main page. The second output from inside iframe doesn't work. We see an error in console:

Uncaught ReferenceError: require is not defined

So, node.js is not available by default inside iframe. Unfortunately, we didn't find a way to activate it in the manual. If there is such a way, please, tell us.

One can suppose it was designed this way in sake of security. However, if we alter "node-call.js" by adding a "top" to "require"

os = top.require('os');

both calls work fine. Hence, either the reason is not a security concern, or developers didn't take into account that one could call modules through "top".

Anyway it is bad because either the thing that should work does not, or works one that should not.
This means that the possibility of the call through a "top" could be blocked in next releases. Or we do something wrong.
Please help us!

zcbenz:
The node integration in iframe is removed as a part of #1044, you can find the technical reasons there.

PR #18753, opened 2019-06-12
docs: remove out of date <webview> docs warning

bmarcaur:
#### Description of Change
Remove the documentation warnings about the stability and longevity of the `<webview>` tag feature.

Originally added via this PR: https://github.com/electron/electron/pull/13835

My impression reading #13869 and the [Out of Process Iframe Docs](https://www.chromium.org/developers/design-documents/oop-iframes) is that the `<webview>` tag became much more stable and better aligned with the long term direction of Chrome by moving it to OOPIF. Even the `<webview>` docs themselves seem to conflict on the nature of its implementation / stability:

> Electron's webview tag is based on Chromium's webview

vs
> Under the hood webview is implemented with Out-of-Process iframes (OOPIFs)

Based on that understanding, it would make sense to remove the warning from the docs.

**Side question:** Was the [default value change in v5](https://github.com/electron/electron/blob/master/docs/api/webview-tag.md#enabling) in an effort to slowly phase out `<webview>` because of its perceived instability?

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/docs/api/webview-tag.md
+++ b/docs/api/webview-tag.md
@@ -1,13 +1,5 @@
 # `<webview>` Tag
 
-## Warning
-
-Electron's `webview` tag is based on [Chromium's `webview`][chrome-webview], which
-is undergoing dramatic architectural changes. This impacts the stability of `webviews`,
-including rendering, navigation, and event routing. We currently recommend to not
-use the `webview` tag and to consider alternatives, like `iframe`, Electron's `BrowserView`,
-or an architecture that avoids embedded content altogether.
-
 ## Enabling
 
 By default the `webview` tag is disabled in Electron >= 5.  You need to enable the tag by


felixrieseberg:
Hey @bmarcaur! Thanks a ton for your contribution, we really appreciate it when newcomers take a crack at improving anything in Electron, especially when its documentation.

In this case, we've discussed this briefly amongst some maintainers and believe that the warning is still very much warranted - `<webview />` remains fairly unstable, has tons of known and even more unknown issues, and is currently still something we'd like for people to only use if they _know_ they cannot use one of the other options.

Do you think we can make that sentiment more clear in the docs?

bmarcaur:
Thanks for the response @felixrieseberg! I will certainly give that wording a try. If you don't mind, I have a few follow up questions that I think will help improve my docs. 

Is it safe to characterize the feature as here to stay but with a set of tradeoffs that has led the team to turn it off by default? Put another way, I would hate for the docs to make it seem safer to adopt thus making any future changes (or removals) more painful for maintainers / consumers.

Is there a place I can learn more about the tradeoffs of `<webview>`? I would like to accurately document them in my update. My current understanding based on the implementation in #13869 is that `<webview>` is implemented using iframes via the shadow DOM. The one key difference being that it guarantees said iframe has a unique "site instance" thus forcing a separate process. Do iframes from different domains share similar tradeoffs or is the instability based on the additional APIs added to the `<webview>` tag? I guess I am asking, is there a known way to avoid the instability of the tag that could be documented?

zcbenz:
Every time after we upgrade Chromium in Electron, we would receive some reports of broken things in webview:
https://github.com/electron/electron/labels/component%2Fwebview

The biggest problem with webview is, there is no warranty for using it, Chromium is free to change behaviors of webview, and as webview is not very commonly used in Chrome extensions, things break frequently in webview.

So if you use webview extensively, your app is almost certainly to break after 1 or 2 Chromium upgrades. The complaints would come to Electron team, while we don't really have solutions as webview is mostly implemented in Chromium.

miniak:
When #18650 is merged, there will be one less reason to use `<webview>`

miniak:
@zcbenz would it be possible to force a same-origin `<iframe>` to have a separate site-instance and therefore a separate renderer process? By adding a custom HTML attribute perhaps?

zcbenz:
> @zcbenz would it be possible to force a same-origin `<iframe>` to have a separate site-instance and therefore a separate renderer process? By adding a custom HTML attribute perhaps?

I don't know whether there is a solution for it.

bmarcaur:
> The biggest problem with webview is, there is no warranty for using it, Chromium is free to change behaviors of webview, and as webview is not very commonly used in Chrome extensions, things break frequently in webview.

Sorry, I am not following this. @zcbenz Didn't you re-implement `<webview>` using OOPIF in #13869? Doesn't that remove the concerns surrounding Chromium's `<webview>`?

> @zcbenz would it be possible to force a same-origin `<iframe>` to have a separate site-instance and therefore a separate renderer process? By adding a custom HTML attribute perhaps?

Isn't this exactly how `<webview>` is currently implemented?

I am not necessarily arguing that `<webview>` should stabilize as an Electron feature, more that the use case of an in-DOM view with process isolation is really powerful for certain types of applications. Being able to manipulate the "site-instance" for "same-origin" iframes would certainly cover this use case. The OOPIF Chromium devs have even hinted that "site-instance" control for iframes is a desired future feature of the `sandbox` property as described in [this ticket](https://bugs.chromium.org/p/chromium/issues/detail?id=510122). So an iframe based solution in Electron would seem to align with their plans.

zcbenz:
> Sorry, I am not following this. @zcbenz Didn't you re-implement `<webview>` using OOPIF in #13869? Doesn't that remove the concerns surrounding Chromium's `<webview>`?
> 

While `<webview>` is now implemented with OOPIF, Chromium still has lots of special code paths for it and things break a lot.

bmarcaur:
I started authoring a version of the docs the incorporated the feedback from @felixrieseberg and @zcbenz, but struggled to elucidate the sources of instability because they aren't really that clear. In an effort to learn more about the "nature" of the instability I dove into the [webview issues for versions that contain the OOPIF change](https://github.com/electron/electron/issues?utf8=%E2%9C%93&q=label%3Acomponent%2Fwebview+-label%3A2-0-x+-label%3A3-0-x+-label%3A%22enhancement+%3Asparkles%3A%22+created%3A%3E2018-08-01+). As far as I can tell there are only 5 issues and only one of them is labelled as a "regression".

@zcbenz @felixrieseberg Can you help me to understand where your team's perception of instability is originating from? Is there a simple example that I could document to help people better understand the risks?

I am happy to chat on slack, or some other medium, if that would make resolving this easier.

trusktr:
With OOPIF out now, is it possible to use iframes again, with node integration? re: https://github.com/electron/electron/issues/3675

codebytere:
Closing due to inactivity. Please ping if you'd like to work further on this; but at this time we still feel the `webview` tag should be designated unstable.

benfrancis:
Hi @codebytere,

I'm currently using `<webview>` in a [couple](https://github.com/webianproject/shell) of [projects](https://github.com/webianproject/shell-kiosk) because there isn't really an alternative for my use cases, but this warning in the docs makes me nervous.

If the Electron team doesn't feel comfortable committing to this feature being stable and turned on by default (which I understand), would it be possible to create something like a GitHub label to track the known issues with `<webview>`?

That way people can make an informed decision based on known issues, have a place to file new issues and a place to find issues to work on to help improve the feature.

Thanks

ChrisHSandN:
Going to post here rather than pollute the bug reports:

I found this thread after upgrading and getting the following warning being emitted:

> (electron) Security Warning: A WebContents was just created with both webviewTag and contextIsolation enabled.  This combination is fundamentally less secure and effectively bypasses the protections of contextIsolation.  We strongly recommend you move away from webviews to OOPIF or BrowserView in order for your app to be more secure

But on investigation I find `<webview>` **is** now implemented through OOPIF, so I don’t understand how this warning is valid. Surely it is better for security to use a well tested core electron component than have everyone reinvent the wheel everytime they need to display external content in DOM?

There have also been only [2 issues](https://github.com/electron/electron/issues?q=label%3Acomponent%2Fwebview+created%3A%3E2020-01-01+) related to `<webview>` filed since the start of 2020 out of 2,191 total. So the “This impacts the stability of webviews, including rendering, navigation, and event routing.” warning seems even more obsolete now than back when this was opened.

Issue #18220, opened 2019-05-09
Cookie support broken in Electron 5.0.x?


danawoodman:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** `5.0.1`
* **Operating System:** `mac OSX 10.14.4`
* **Last Known Working Electron version:**: `4.x`

### Expected Behavior

I should be able to set/remove a cookie.

### Actual Behavior

I cannot set or remove a cookie

### To Reproduce

```js
import { session } from 'electron'

async function testCookies() {
  if (!session || !session.defaultSession)
    throw new Error('Could not get default user session')

  const cookies = session.defaultSession.cookies
  const name = 'foo'
  const url = '/'

  await cookies.set({
    url,
    name,
    value: 'Hello World!',
    expirationDate: 99999999.0,
  })

  console.log('COOKIES BEFORE:', await cookies.get({}))

  await cookies.remove(url, name)

  console.log('COOKIES AFTER:', await cookies.get({}))
}

//....
app.on('ready', () => {
  testCookies()
})

```

### Additional Information

Running this on a Mac and have tried many permutations of the above but can't get anything to work. Things seemed to work in 4.x

grantcv1:
I seem to be having similar problems with localStorage. It's now cleared between sessions

thesuperhomie:
@danawoodman & @grantcv1 either of you guys want to take this on? If not I'll grab it :)

erickzhao:
Heya @danawoodman,

@DeerMichel and I investigated further into this issue, but couldn't find anything wrong with cookie support in `v5.0.3`

[Here's a link to a gist](https://gist.github.com/1554a78fda2124e34000692291e73521) you can paste into [Electron Fiddle](https://electronjs.org/fiddle). This contains a minimum working example for the Cookies API (also available under `Show Me -> Electron APIs -> Cookies` in the Fiddle menu bar).

Please let us know if the below comments don't solve your issue.

-------------------------

[...]

PR #18756, opened 2019-06-12
fix: throw error on invalid URLs when setting cookie

erickzhao:
#### Description of Change

With this PR, invalid inputs to the `url` parameter will throw an error when using `cookie.set()`. This is done by checking if the URL is parseable using `GURL` rather than checking if the URL string being passed in is empty.

Previously, invalid URLs would be able to be added as a cookie, but you would not be able to filter for them or remove them.

See Electron Fiddle gist: https://gist.github.com/d8ae35040d44d426cb40d97d5d77a47e

_N.B. This behavior was fixed in Electron 6, so we based this PR a lot on the changes in the cookies API from 5.0.3 to 6.0.0-beta._

This issue was discovered when playing around with the example from https://github.com/electron/electron/issues/18220.

cc @DeerMichel @codebytere @ckerr 

#### Checklist

- [x] `npm test` passes(https://github.com/electron/electron/blob/master/docs/development/testing.md)

#### Release Notes

Notes: Added check for invalid URLs upon setting cookies

### DIFF
--- a/atom/browser/api/atom_api_cookies.cc
+++ b/atom/browser/api/atom_api_cookies.cc
@@ -226,13 +226,13 @@ void FlushCookieStoreOnIOThread(
 void SetCookieOnIO(scoped_refptr<net::URLRequestContextGetter> getter,
                    std::unique_ptr<base::DictionaryValue> details,
                    util::Promise promise) {
-  std::string url, name, value, domain, path;
+  std::string url_string, name, value, domain, path;
   bool secure = false;
   bool http_only = false;
   double creation_date;
   double expiration_date;
   double last_access_date;
-  details->GetString("url", &url);
+  details->GetString("url", &url_string);
   details->GetString("name", &name);
   details->GetString("value", &value);
   details->GetString("domain", &domain);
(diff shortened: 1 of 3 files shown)


atom/browser/api/atom_api_cookies.cc
```diff
@@ -261,21 +261,24 @@ void SetCookieOnIO(scoped_refptr<net::URLRequestContextGetter> getter,
...
 
-  std::unique_ptr<net::CanonicalCookie> canonical_cookie(
-      net::CanonicalCookie::CreateSanitizedCookie(
-          GURL(url), name, value, domain, path, creation_time, expiration_time,
-          last_access_time, secure, http_only,
-          net::CookieSameSite::DEFAULT_MODE, net::COOKIE_PRIORITY_DEFAULT));
   auto completion_callback = base::BindOnce(OnSetCookie, std::move(promise));
-  if (!canonical_cookie || !canonical_cookie->IsCanonical()) {
+  GURL url(url_string);
+  if (url.is_empty()) {
```
deepak1556: prefer `is_valid`

```
From GURL::is_empty
// Returns true if the URL is zero-length. Note that empty URLs are also
// invalid, and is_valid() will return false for them. This is provided
// because some users may want to treat the empty case differently.
```

docs/api/cookies.md
```diff
@@ -104,7 +104,7 @@ with `callback(error, cookies)` on complete.
 #### `cookies.set(details)`
 
 * `details` Object
-  * `url` String - The url to associate the cookie with.
+  * `url` String - The url to associate the cookie with. An error is thrown if the url is invalid.
```
MarshallOfSound:
```suggestion
  * `url` String - The url to associate the cookie with. The promise will be rejected if the url is invalid.
```

atom/browser/api/atom_api_cookies.cc
MarshallOfSound:
```suggestion
  if (!url.is_valid()) {
```

docs/api/cookies.md
erickzhao: nit: @DeerMichel do you want to align the wording in the documentation that you put on #18751 with this one? I think confirming that it'll be a rejected promise might make the docs _slightly_ clearer 😅

DeerMichel: sure

codebytere:
@erickzhao trop failed us here; can you open a manual bp to `4-2-x`?

PR #18757, opened 2019-06-12
feat: provide the frame URL with permission requests and checks

### LINKED ISSUE #18755: session.setPermissionRequestHandler for iframes has incorrect content URL
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * Tested with v4.1.4 and above
* **Operating System:**
  * macOS 10.14.4
  * n/a

The **webContents getURL() does not return the contents URL for an iframe when a device permissions request is emitted**.  Compared to a webview when a device permissions request is handled the webContent passed on the event has the correct URL for the content in the webview.  In the case of iframes the URL given is the parent windows and not the url for the content of the iframe.

RUN THE ATTACHED ELECTRON FIDDLE

### Expected Behavior

The url printed in the console should be the same for each call and match the content url of both the iframe and webview.

Both getURL calls should have a value of: https://statics.teams.microsoft.com/evergreen-assets/tabs/v0.4/devicePermissionTab.html

### Actual Behavior

The url printed in the console for the iframe is incorrect.  Its giving us the url of the parent window and not the contents window:  file:///var/folders/1m/mvzrl4cn4c7cqw47g8xk1m3h0000gn/T/tmp-10161DTmlueZz9SLn/index.html

### To Reproduce
* Run the attached test Electron fiddle app.
* In the first window (iframe) click on the button under "notifications" to which will emit a device permissions request on the content.
* In the second window (webview) click the same button under "notifications".  
* Look at the console output

### Additional Information

[iframedeviceperms.zip](https://github.com/electron/electron/files/3283118/iframedeviceperms.zip)

MarshallOfSound:
Fixes #18755

Users need to be able to know where the origin of the permission request is.

Notes: Added `requestingUrl` and `isMainFrame` properties to the permission request and permission check handlers details objects so that apps can check the origin of the requesting frame rather than the root frame.

### DIFF
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -284,15 +284,17 @@ win.webContents.session.setCertificateVerifyProc((request, callback) => {
 #### `ses.setPermissionRequestHandler(handler)`
 
 * `handler` Function | null
-  * `webContents` [WebContents](web-contents.md) - WebContents requesting the permission.
+  * `webContents` [WebContents](web-contents.md) - WebContents requesting the permission.  Please note that if the request comes from a subframe you should use `requestingUrl` to check the request origin.
   * `permission` String - Enum of 'media', 'geolocation', 'notifications', 'midiSysex',
     'pointerLock', 'fullscreen', 'openExternal'.
   * `callback` Function
     * `permissionGranted` Boolean - Allow or deny the permission.
   * `details` Object - Some properties are only available on certain permission types.
-    * `externalURL` String - The url of the `openExternal` request.
-    * `mediaTypes` String[] - The types of media access being requested, elements can be `video`
+    * `externalURL` String (Optional) - The url of the `openExternal` request.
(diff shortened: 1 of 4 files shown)


docs/api/session.md
```diff
@@ -284,15 +284,16 @@ win.webContents.session.setCertificateVerifyProc((request, callback) => {
 #### `ses.setPermissionRequestHandler(handler)`
 
 * `handler` Function | null
-  * `webContents` [WebContents](web-contents.md) - WebContents requesting the permission.
+  * `webContents` [WebContents](web-contents.md) - WebContents requesting the permission.  Please note that if the request comes from a subframe you should use `requestingUrl` to check the request origin.
```
ckerr: @MarshallOfSound @felixrieseberg would it be productive to provide a snippet showing an example of this?

PR #18723, opened 2019-06-10
refactor: mojofy autofill

nornagon:
#### Description of Change
This converts the autofill popup IPC messages to mojo and removes our last usage of legacy IPC.

I considered reusing the existing autofill logic in `//components/autofill`, but it's quite extensive and includes support for credit cards, passwords, and a bunch of other stuff that Electron doesn't currently support. We might want to support that in future, but for now this just converts the existing functionality to mojo.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: no-notes

PR #18760, opened 2019-06-12
refactor: mojofy autofill renderer->browser

brenca:
#### Description of Change

This is a followup PR for #18723, mojofying the `renderer`->`browser` calls similarly to how it's done in `chromium`, as per discussion with @nornagon.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes


atom/browser/atom_autofill_driver_factory.cc
```diff
@@ -0,0 +1,133 @@
...
+    content::WebContents* contents) {
+  if (FromWebContents(contents))
+    return;
+
+  auto new_factory = std::make_unique<AutofillDriverFactory>(contents);
+  const std::vector<content::RenderFrameHost*> frames =
+      contents->GetAllFrames();
+  for (content::RenderFrameHost* frame : frames) {
+    if (frame->IsRenderFrameLive())
+      new_factory->RenderFrameCreated(frame);
```
zcbenz: This can be moved to constructor of `AutofillDriverFactory`. There is no need to implement `AutofillDriverFactory::CreateForWebContents` and `FromWebContents` after making the class inherit `content::WebContentsUserData`.

shell/browser/atom_autofill_driver_factory.cc
```diff
@@ -0,0 +1,113 @@
...
+      AutofillDriverFactory::FromWebContents(web_contents);
+  if (!factory)
+    return;
+
+  AutofillDriver* driver = factory->DriverForFrame(render_frame_host);
+  if (driver)
+    driver->BindRequest(std::move(request));
+}
+
+AutofillDriverFactory::AutofillDriverFactory(content::WebContents* web_contents)
```
deepak1556: Is it necessary to eagerly create drivers for all living frames ? `BindAutofillDriver` seems to be called on demand, would simplify this class if the drivers were lazy loaded.

shell/browser/atom_autofill_driver.cc
```diff
@@ -0,0 +1,65 @@
...
+AutofillDriver::AutofillDriver(content::RenderFrameHost* render_frame_host)
+    : render_frame_host_(render_frame_host), binding_(this) {
+  autofill_popup_.reset(new AutofillPopup());
+}
+
+AutofillDriver::~AutofillDriver() {}
+
+void AutofillDriver::BindRequest(
+    mojom::ElectronAutofillDriverAssociatedRequest request) {
+  binding_.Bind(std::move(request));
```
deepak1556: Based on the lazy load suggestion, its simpler to bind to the request in the constructor rather than a specific method since the lifetime is always connected to one render frame host. If there are multiple requests per frame , then its better to have a separate method like this and also in that case existing connections should be unbound before creating a new one.

PR #18766, opened 2019-06-13
fix: use crashpad on Windows (6-0-x)

zcbenz:
Backport of #18483.

Notes: Migrate crashReporter to use crashpad on Windows.


DEPS
```diff
@@ -12,7 +12,7 @@ vars = {
   'chromium_version':
     '76.0.3809.23',
   'node_version':
-    'dee0db9864a001ffc16440f725f4952a1a417069',
+    '229bd3245b2f54c12ea9ad0abcadbc209f8023dc',
```
miniak: there was a conflict here, which version should be used?

nitsakh: 229bd3245b2f54c12ea9ad0abcadbc209f8023dc is the right commit.

PR #18769, opened 2019-06-13
feat: add contents.getOSProcessIdForFrame()

miniak:
#### Description of Change
Added `webContents.getOSProcessIdForFrame()`. This allows getting the pid of renderers hosting cross-origin frames.

Required for tests in #18650

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `webContents.getOSProcessIdForFrame()`. This allows getting the pid of renderers hosting cross-origin frames.

### DIFF
--- a/spec/api-web-contents-spec.js
+++ b/spec/api-web-contents-spec.js
@@ -559,7 +559,7 @@ describe('webContents module', () => {
   })
 
   describe('getOSProcessId()', () => {
-    it('returns a valid procress id', async () => {
+    it('returns a valid process id', async () => {
       expect(w.webContents.getOSProcessId()).to.equal(0)
 
       await w.loadURL('about:blank')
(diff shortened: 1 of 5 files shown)


spec/api-web-contents-spec.js
```diff
@@ -567,6 +567,55 @@ describe('webContents module', () => {
...
+
+      const promise1 = emittedOnce(ipcMain, 'pid-main')
+      const promise2 = emittedOnce(ipcMain, 'pid-frame')
+
+      await w.loadURL(server.url)
+
+      const [{ frameId: frameIdMain }, pidMain] = await promise1
+      const [{ frameId: frameIdFrame }, pidFrame] = await promise2
+
+      expect(w.webContents.getOSProcessIdForFrame(frameIdMain)).to.equal(pidMain)
```
deepak1556: Can you add a check for verifying the process ids are different and the frame ids are also different.

The reason for frame id check is that its based on routing id, which maybe unique at the moment but if chromium makes them per process identifier then we should switch to something like `GlobalFrameRoutingId`. This spec would help recover from such changes.

miniak: @deepak1556 I am actually thinking about changing the API lookup by `GetFrameName` instead

PR #18770, opened 2019-06-13
fix: use is_valid for cookie url validation

DeerMichel:
#### Description of Change

- Use `is_valid` instead of `is_empty` to validate url on `cookies.set()`. (same behavior as in 5.x)
- Adjust docs to match 5.x docs according to #18756.

cc @codebytere @ckerr 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


DeerMichel:
Huh, that's interesting...

DeerMichel:
@MarshallOfSound should be fine now 😆

PR #18774, opened 2019-06-13
feat: provide the frame URL with permission requests and checks

MarshallOfSound:
Backport of #18757

See that PR for details

Notes: Added `requestingUrl` and `isMainFrame` properties to the permission request and permission check handlers details objects so that apps can check the origin of the requesting frame rather than the root frame.


zcbenz:
@MarshallOfSound Do you still want to get this backport merged?

Issue #18639, opened 2019-06-05
Consistent crash when hardware acceleration is disabled


adill:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 6.0.0-beta.5 (also present in beta.6)
* **Operating System:**
  * macOS 10.14.5
* **Last Known Working Electron version:**
  * 6.0.0-beta.4

### Expected Behavior
Electron should not crash after calling `app.disableHardwareAcceleration()`

### Actual Behavior
Electron crashes (main process) when started with hardware acceleration disabled.

```
[16974:0604/215819.260822:FATAL:gpu_data_manager_impl_private.cc(901)] The display compositor is frequently crashing. Goodbye.
```

### To Reproduce

Reproducible trivially with electron-quick-start

```sh
$ git clone https://github.com/adill/electron-quick-start -b macOS-no-gpu-crash
$ npm install
$ npm start
```

PR #18778, opened 2019-06-13
fix: re-enable disabling hardware acceleration

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/18639.

Refs Chromium CL [1599993](https://chromium-review.googlesource.com/c/chromium/src/+/1599993).

This CL injects the `SharedContextState` into `CommandBufferTaskExecutor`
if one is available and use it in InProcessCommandBuffer, which i believe created the crash because it would be trying to access a memory location which did not yet exist. For now, pass a nullptr instead.

cc @deepak1556 @zcbenz  @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash when enabling `app.disableHardwareAcceleration()`.

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -77,3 +77,4 @@ frame_host_manager.patch
 cross_site_document_resource_handler.patch
 woa_compiler_workaround.patch
 crashpad_pid_check.patch
+fix_enabled_disabling_hw_acceleration.patch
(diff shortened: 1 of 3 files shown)


zcbenz:
This patch would be very hard to maintain after a few more Chromium upgrades, would like to hear opinions from @nornagon.

codebytere:
Figured out a way smaller fix here, updating soon.

nornagon:
Agree with @deepak1556, this seems like something that ought to be reproducible upstream & therefore should be reported there. If it's not reproducible upstream, then I'd like to get more info on why :)

codebytere:
Superseded by https://github.com/electron/electron/pull/18870.

Issue #18805, opened 2019-06-14
`app.moveToApplicationsFolder` fails if an homonymous app is already present and active


fabiospampinato:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** 
  *  5.0.3
* **Operating System:** 
  *  macOS 10.14
* **Last Known Working Electron version:** 
  *  n/a

### Expected Behavior

`app.moveToApplicationsFolder` should throw when it fails, as the docs say.

### Actual Behavior

It doesn't throw, in fact it even returns `true`.

### To Reproduce

- Have Foo v1 located in `/Applications/Foo.app`
- Open it
- Have Foo v2, still named `Foo.app`, located somewhere else
- Open it
- Use `app.moveToApplicationsFolder`.
- Instead of moving the app the currently open Foo v1 window will be focused, and `app.moveToApplicationsFolder` returns `true` instead of throwing.

PR #18811, opened 2019-06-15
fix: return false when name conflict on app move

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/18805.

Previously, if an app was running of the same name in Applications when `app.moveToApplicationsFolder()` was called, we would assume that meant it was the _same_ app, so we would quit this one and focus the other app, while returning true. Given nontrivial potential for naming clashes, this seems to be a potentially dangerous assumption to make as demonstrated by the above error.

This PR thus changes that behavior to throw an error indicating the naming conflict, and return false to indicate no such move occurred.

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where `app.moveToApplicationsFolder()` would incorrectly return true and focus the wrong app if an identically named app was present and running in Applications when called.

### DIFF
--- a/atom/browser/ui/cocoa/atom_bundle_mover.mm
+++ b/atom/browser/ui/cocoa/atom_bundle_mover.mm
@@ -74,21 +74,17 @@ bool AtomBundleMover::Move(mate::Arguments* args) {
     if ([fileManager fileExistsAtPath:destinationPath]) {
       // But first, make sure that it's not running
       if (IsApplicationAtPathRunning(destinationPath)) {
-        // Give the running app focus and terminate myself
-        [[NSTask
-            launchedTaskWithLaunchPath:@"/usr/bin/open"
-                             arguments:[NSArray
-                                           arrayWithObject:destinationPath]]
-            waitUntilExit];
-        atom::Browser::Get()->Quit();
-        return true;
+        // We don't know if the app is another app of the same name,
+        // or the same app, so throw an error and leave it to the user
+        // to determine next steps
+        args->ThrowError("An app of the same name appears to already exist in "
+                         "Applications.");
       } else {
         if (!Trash([applicationsDirectory
-                stringByAppendingPathComponent:bundleName])) {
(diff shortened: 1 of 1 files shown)


atom/browser/ui/cocoa/atom_bundle_mover.mm
```diff
@@ -74,21 +74,17 @@
     if ([fileManager fileExistsAtPath:destinationPath]) {
       // But first, make sure that it's not running
       if (IsApplicationAtPathRunning(destinationPath)) {
-        // Give the running app focus and terminate myself
-        [[NSTask
```
MarshallOfSound: I'd like to retain the ability for folks to do this, I believe it is the sensible default behavior.  Making it overridable / interceptable would make sense but if we remove this apps can't do this without either string matching the error message (which you should never do) and then spawning `open` themselves.

Not sure what the ideal state looks like but retaining this activate-and-quit behavior is important I feel.  i.e. the 99% case is activate-and-quit, we shouldn't remove that to fit the 1% case

codebytere: 🤔 maybe we can just pass an (optional) bool that determines whether to disable this behavior if there's a conflict?

MarshallOfSound: I'd love to give app devs the chance to let users decide at-conflict time rather than in advance.  E.g. an API that looked roughly like this.

```js
app.moveToApplicationsDirectory({
  conflictHandler: (conflictType) => {
    if (conflictType === 'exists') {
      return askUserIfShouldOpenAndQuit()
    }
    return true;
  }
})
```

Disclaimer: super rough sketch

codebytere: sgtm, will work on this at a Work Time next week 😆

codebytere:
Closing in favor of taking new approach outlined above.

Issue #12049, opened 2018-02-26
A confusing "Electron Security Warning" in Electron v1.8.2


AlexSHoffman:
Electron v1.8.2 shows a security warning that makes you think there's a serious security problem in your project and do not provide any details which makes the warning useless and more importantly confusing.

* Electron version: 1.8.2
* Operating system: Win10

### Actual behavior

Some people got confused by the following Electron Warning:

> Electron Security Warning This renderer process has Node.js integration enabled and attempted to load remote content. This exposes users of this app to severe security risks.

You can see it in the console as soon as you create and run a new Electron-Vue project.

Does Electron thinks that Webpack's `localhost` is attempting to download some remote content? What content? What attempt? How does that warning help anybody? 

### How to reproduce

- Create a new [Electron-Vue](https://github.com/SimulatedGREG/electron-vue) project
- `yarn run dev` (or `npm run dev`)
- See the console

qazbnm456:
@AlexSHoffman Well, I think there's a misunderstanding in the original [thread](https://github.com/SimulatedGREG/electron-vue/issues/540). Sorry about that. 😕

These security warnings are actually forcing developers to pay more attention to https://github.com/electron/electron/blob/master/docs/tutorial/security.md, and they will only show up when you're under development mode (since the default binary name will be **electron**). You can find the reference [here](https://github.com/electron/electron/pull/11810).

There's no way to disable this feature in 1.8.2 in development mode though, you can switch them on and off in 2.0 series by setting `ELECTRON_ENABLE_SECURITY_WARNINGS` or `ELECTRON_DISABLE_SECURITY_WARNINGS` on either `process.env` or the `window` object.

AlexSHoffman:
@qazbnm456 thanks for clarification, I edited the issue, the problem is still valid. The Electron Warning says Node is attempting to load remote content, that is not helpful to anybody. It made at least 7 people confused already (based on the feedback related to this issue on stackoverflow and Github)

ckerr:
@felixrieseberg, any thoughts on this?

felixrieseberg:
I do have thoughts! @AlexSHoffman The core issue is that too many developers, especially newcomers to Electron, assume that using `http://` with `nodeIntegration` enabled is okay. It's not - the warning is correct, that's a fairly severe security risk. 

[...]

PR #18814, opened 2019-06-15
fix: disable nodeIntegration & insecure resource warnings for localhost

jerry1100:
#### Description of Change
This PR removes the "node integration with remote content" and "loading insecure content" warning messages when loading from localhost.

A lot of electron projects use webpack-dev-server for development, which involves "remote" content being loaded over an insecure connection from localhost, resulting in at least two warning messages in the console.

The warning messages are annoying and they [confuse people](https://github.com/electron/electron/issues/12049). Worst, they can actually be counterproductive as people are resorting to [disabling the security check](https://stackoverflow.com/questions/48854265/why-do-i-see-an-electron-security-warning-after-updating-my-electron-project-t) just to get rid of the warnings, which means they won't see them when it really matters.

**Note**: I explicitly chose not to test against `127.0.0.1` because although it should be the same as `localhost`, the latter is much more prevalent. Also, that would break a lot of tests.

Other changes:
* added `isLocalhost()` to check if the current window's hostname is localhost
* also ignoring [enabled remote module with remote content](https://github.com/electron/electron/pull/18822) when loading from localhost

#### Checklist
- [ ] `npm test` passes (I'm having issues with setting up the project, I'll leave it up to CI)

#### Release Notes
Notes: "Node integration with remote content" and "loading insecure content" warning messages are suppressed for localhost connections.


jerry1100:
Looks like the tests are failing.

"Node integration with remote content" is failing because it's waiting for a console message that never comes (since it's suppressed). I'm not sure how to test this but here are some ideas:
1. Keep the `console-message` listener and add a `did-finish-load` listener that will pass the test. The idea is that the warning, if it were to come, will come before the `did-finish-load` event. Not sure if that's guaranteed to be the case though.
2. Pass the test after a fixed amount of time. Perhaps 10s? Something way longer than it would take the warning to show.
```
not ok 1124 security warnings should not warn about Node.js integration with remote content from localhost
  Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\projects\src\electron\spec\security-warnings-spec.js)
      at Test.Runnable._timeoutError (C:\projects\src\electron\spec\node_modules\mocha\lib\runnable.js:440:10)
      at C:\projects\src\electron\spec\node_modules\mocha\lib\runnable.js:251:24
```

"Loading insecure resources" is failing because the [resource that's being loaded](https://github.com/electron/electron/blob/master/spec/fixtures/pages/insecure-resources.html) (which is correctly filtered) is loading a stylesheet from `127.0.0.1`, which is not `=== 'localhost'`, and thus the warning message shows. To get around this, we can remove the stylesheet since it's redundant to have two insecure resources for testing.
```
not ok 1132 security warnings without sandbox should not warn about insecure resources from localhost
  AssertionError: expected '%cElectron Security Warning (Insecure Resources)' to not include 'Insecure Resources'
      at C:\projects\src\electron\spec\security-warnings-spec.js:210:34
      at CallbacksRegistry.apply (electron/js2c/renderer_init.js:1752:29)
      at electron/js2c/renderer_init.js:1553:21
      at EventEmitter.<anonymous> (electron/js2c/renderer_init.js:1543:7)
      at EventEmitter.emit (events.js:194:13)
      at Object.onMessage (electron/js2c/renderer_init.js:2343:16)
```

jerry1100:
For the "node integration with remote content" test, I opted in for using `did-finish-load` since I didn't like the idea of using a fixed delay. However, like I mentioned in my above post, this runs the risk of passing early if `did-finish-load` gets fired before `console-message` (though in my testing it does not).

For the "Loading insecure resources" test, I modified the test to check that the resource it's loading from localhost is not included in the warning message.

jerry1100:
An unrelated test failed. Does someone mind restarting it?

MarshallOfSound:
@jerry1100 

> Error: No valid signing identity available to run autoUpdater specs

That spec failure is expected at the moment for fork PRs, we're working on making it work for PRs from forks but for now that failure is expected and ok 👍

spec/security-warnings-spec.js
```diff
@@ -78,6 +78,22 @@ describe('security warnings', () => {
...
+      show: false,
+      webPreferences: {
+        nodeIntegration: true
+      }
+    })
+    w.webContents.once('console-message', (e, level, message) => {
+      expect(message).to.not.include('Node.js Integration with Remote Content')
+      done()
+    })
+    w.webContents.once('did-finish-load', () => done())
```
zcbenz: You can add a `console.log('loaded')` in `base-page-security.html` and then end the test when it is printed.

```js
on('console-message', () => {
  expect(message).to.not.include('Node.js Integration with Remote Content')
  if (message === 'done')
    done()
})
```

(other test loading `base-page-security.html` may be affected so you probably need to create a new file)

jerry1100: Thanks for the suggestion, but I don't think that will work. `console.log('loaded')` will  be executed _before_ the nodeIntegration check has a chance to run, resulting in the test always passing, even if the warning is shown.

We'd need some way to detect when the nodeIntegration check is finished, so we can pass knowing that a message wasn't shown.

zcbenz: Can you have a try of `window.onload = () => console.log('loaded')`?

jerry1100: Nice, that worked.

I assume the event handler in [`security-warning.ts`](https://github.com/electron/electron/blob/master/lib/renderer/security-warnings.ts#L280) gets registered before the one I added, so when the `onload` event fires, the nodeIntegration check is guaranteed to execute first.

We could've actually put the event handler in `base-page-security.html` without breaking existing tests since they only listen for the first console message and "loaded" will come after. However, I think it'd be safer to not pollute the base page, so I added `base-page-security-loaded-message.html`.

zcbenz:
Would like to hear more voices from @electron/wg-security before merging.

jerry1100:
Made some minor updates:
1. Added `isLocalhost()` to reduce code duplication
2. Handled the newly added "enabled remote module with remote content" warning

jerry1100:
Bump. Any input from @electron/wg-security?

felixrieseberg:
Hi, here from the security wg. I'm okay with this, this seems like a solid change that doesn't undermine the spirit of warning people when they're about to shoot themselves in the foot.

PR #18815, opened 2019-06-15
fix: extra space in security warning message causing list misalignment

jerry1100:
#### Description of Change
This PR removes an extra space in the "warn about insecure resources" warning
string so the list of insecure resources are properly aligned.

Update: also removed whitespaces after newlines since they mess up the indentation.

Screen Shot 2019-06-15 at 2 29 52 AM

#### Checklist
- [ ] `npm test` passes (having trouble with setting up the project, I'll leave it up to CI)

#### Release Notes

Notes: no-notes

### DIFF
--- a/lib/renderer/security-warnings.ts
+++ b/lib/renderer/security-warnings.ts
@@ -101,7 +101,7 @@ const warnAboutInsecureResources = function () {
 
   const warning = `This renderer process loads resources using insecure
   protocols. This exposes users of this app to unnecessary security risks.
-  Consider loading the following resources over HTTPS or FTPS. \n ${resources}
+  Consider loading the following resources over HTTPS or FTPS. \n${resources}
   \n ${moreInformation}`
 
   console.warn('%cElectron Security Warning (Insecure Resources)',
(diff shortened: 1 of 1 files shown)


jerry1100:
Could someone with write access restart the tests? It's a non-related failure.

jerry1100:
I think the tests are flaky. Anyone care to have a look?

lib/renderer/security-warnings.ts
```diff
@@ -101,7 +101,7 @@ const warnAboutInsecureResources = function () {
 
   const warning = `This renderer process loads resources using insecure
   protocols.This exposes users of this app to unnecessary security risks.
-  Consider loading the following resources over HTTPS or FTPS. \n ${resources}
+  Consider loading the following resources over HTTPS or FTPS. \n${resources}
   \n ${moreInformation}`
```
nornagon: Should the extra space be removed here too?

jerry1100: Yes it should, I’ll do it later tonight.

Although visually it won’t make a difference since moreInformation starts off with a newline :)

PR #18820, opened 2019-06-15
chore: convert more files to TypeScript

miniak:
#### Description of Change
Convert the following files to TypeScript:
- `lib/common/buffer-utils.js`
- `lib/common/clipboard-utils.js`
- `lib/common/error-utils.js`
- `lib/common/web-view-methods.js`
- `lib/browser/devtools.js`

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes


lib/common/buffer-utils.ts
```diff
@@ -19,16 +14,16 @@ const typedArrays = {
   Float64Array
 }
 
-function getType (value) {
+function getType (value: ArrayBuffer | ArrayBufferView) {
```
nornagon:
```suggestion
function getType (value: BufferSource) {
```

https://heycam.github.io/webidl/#BufferSource

lib/common/buffer-utils.ts
```diff
@@ -38,19 +33,25 @@ function getBuffer (value) {
...
+  data: Buffer;
+  length: number;
+}
+
+export function bufferToMeta (value: ArrayBuffer | ArrayBufferView): BufferMeta {
   return {
     type: getType(value),
     data: getBuffer(value),
-    length: value.length
+    length: (value as Buffer).length
```
nornagon: this doesn't seem right. `ArrayBuffer` doesn't have a `length` property (it has `byteLength`).

lib/common/error-utils.ts
```diff
@@ -10,10 +8,15 @@ const constructors = new Map([
   [URIError.name, URIError]
 ])
 
-exports.deserialize = function (error) {
+interface ElectronError extends Error {
+  from?: string;
+  cause?: ElectronError;
+}
+
+export function deserialize (error: Electron.SerializedError) {
```
nornagon: Add a return type?

lib/common/error-utils.ts
```diff
@@ -22,7 +25,7 @@ exports.deserialize = function (error) {
   return error
 }
 
-exports.serialize = function (error) {
+export function serialize (error: ElectronError) {
```
nornagon: Return type here too please

lib/common/buffer-utils.ts
miniak: we only need `length` when we are serializing `Int8Array`, etc.
when deserializing `Buffer` and `ArrayBuffer`, `length` is not used

lib/common/buffer-utils.ts
```diff
@@ -19,16 +14,16 @@ const typedArrays = {
   Float64Array
 }
 
-function getType (value) {
+function getType (value: BufferSource) {
```
nornagon: hm... actually this should also include `Buffer` which `BufferSource` doesn't.

maybe `type BufferLike = ArrayBuffer | ArrayBufferView | Buffer`?

lib/common/buffer-utils.ts
nornagon: Then `BufferMeta` should have `length: number | undefined`?

lib/common/buffer-utils.ts
miniak: `Buffer` is now actually `Uint8Array`, but it's treated specially to keep it transferred as `Buffer`

miniak:
@nornagon can you please check again?

lib/common/buffer-utils.ts
nornagon: Please also add a comment here, this is confusing & looks like a bug without the appropriate context.

lib/common/buffer-utils.ts
```diff
@@ -38,19 +35,25 @@ function getBuffer (value) {
...
+  data: Buffer;
+  length: number | undefined;
+}
+
+export function bufferToMeta (value: BufferLike): BufferMeta {
   return {
     type: getType(value),
     data: getBuffer(value),
-    length: value.length
+    length: (value as Buffer).length
```
nornagon:
```suggestion
    // NB. We only use length when decoding Int8Array and friends.
    // For other buffer-like types this is expected to be undefined.
    length: (value as Buffer).length
```

lib/common/buffer-utils.ts
nornagon: nvm i went ahead and added it :)

PR #18844, opened 2019-06-17
chore: fix current branch fetch on master

codebytere:
#### Description of Change

When fetching current branch on a nightly release, we need to account for the fact that `master` doesn't fit the release branch formatting. This fixes that error.

cc @MarshallOfSound 

#### Release Notes

Notes: none

### DIFF
--- a/script/lib/utils.js
+++ b/script/lib/utils.js
@@ -28,7 +28,8 @@ function getAbsoluteElectronExec () {
 async function handleGitCall (args, gitDir) {
   const details = await GitProcess.exec(args, gitDir)
   if (details.exitCode === 0) {
-    return details.stdout.trim()
+    const output = details.stdout.replace(/^\*|\s+|\s+$/, '')
+    return output.trim()
   } else {
     const error = GitProcess.parseError(details.stderr)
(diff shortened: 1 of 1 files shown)


script/lib/utils.js
```diff
@@ -41,9 +43,9 @@ async function getCurrentBranch (gitDir) {
   if (!branch.match(/[0-9]+-[0-9]+-x/)) {
```
jkleinsc: You can check for `master` here instead of using git branch to find it.

script/lib/utils.js
```diff
@@ -38,12 +40,12 @@ async function handleGitCall (args, gitDir) {
 
 async function getCurrentBranch (gitDir) {
   let branch = await handleGitCall(['rev-parse', '--abbrev-ref', 'HEAD'], gitDir)
-  if (!branch.match(/[0-9]+-[0-9]+-x/)) {
+  if (branch !== 'master' && !branch.match(/[0-9]+-[0-9]+-x/)) {
```
MarshallOfSound: This should probably be trimmed if we are returning it trimmed below

PR #18845, opened 2019-06-17
fix: use effectiveAppearance to check dark mode on mojave and above

codebytere:
Backport of https://github.com/electron/electron/pull/18666.

See that PR for more details.

Notes: Fixed issue on macOS Catalina where the tray icon would be highlighted incorrectly and sometimes invisible.

### DIFF
--- a/atom/browser/ui/tray_icon_cocoa.mm
+++ b/atom/browser/ui/tray_icon_cocoa.mm
@@ -4,8 +4,10 @@
 
 #include "atom/browser/ui/tray_icon_cocoa.h"
 
+#include "atom/browser/mac/atom_application.h"
 #include "atom/browser/ui/cocoa/NSString+ANSI.h"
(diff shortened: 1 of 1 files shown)


atom/browser/ui/tray_icon_cocoa.mm
```diff
@@ -326,9 +332,9 @@ - (void)mouseUp:(NSEvent*)event {
 - (void)popUpContextMenu:(atom::AtomMenuModel*)menu_model {
   // Show a custom menu.
   if (menu_model) {
-    base::scoped_nsobject<AtomMenuController> menuController([
-        [AtomMenuController alloc] initWithModel:menu_model
-                           useDefaultAccelerator:NO]);
+    base::scoped_nsobject<AtomMenuController> menuController(
```
codebytere: (this is just autoformatting)

Issue #18827, opened 2019-06-16
process.versions should include boringssl


MarshallOfSound:
On a similar vein to https://github.com/electron/electron/issues/18826 the `process.versions` object should include the `boringssl` version to allow people to do version checks 👍

DeerMichel:
See #18826.

jkleinsc:
@DeerMichel this shouldn't be closed until a PR addressing this is merged in.

nornagon:
Closing this as a duplicate of #18826.

Issue #18826, opened 2019-06-16
rename process.versions.{openssl => boringssl}


MarshallOfSound:
We no longer depend on openssl at all, it should not be listed in `process.versions`.

![image](https://user-images.githubusercontent.com/6634592/59560038-b5046380-8fbf-11e9-9207-46874847126e.png)

dulvinw:
Can I get a beginner issue? new to both JS and OSS

nornagon:
I think we should keep `process.versions.openssl`:

1. some things expect `process.versions.openssl` to exist; this would break them.
2. BoringSSL doesn't have a version, so there's nothing meaningful we could put in `process.versions.boringssl`.
3. If people want to version-check boringssl, they can use electron version as a proxy, since it's not possible to load a different boringssl with a given version of electron as it's statically compiled.

nornagon:
Closing this for now, we can reopen if needed.

PR #18846, opened 2019-06-17
chore: use boringssl instead of openssl in process.version

DeerMichel:
#### Description of Change
Addresses #18826 and #18827. Tbh I didnt find out if this PR really does the expected thing.

To which versions should we backport this?

cc @codebytere @erickzhao 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/tools/dump-version-info.js
+++ b/tools/dump-version-info.js
@@ -19,7 +19,7 @@ function getInfoForCurrentVersion () {
   json.version = process.versions.electron
   json.date = getDate()
 
-  var names = ['node', 'v8', 'uv', 'zlib', 'openssl', 'modules', 'chrome']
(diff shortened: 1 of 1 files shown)


DeerMichel:
You're right :). Heading down to c++...

MarshallOfSound:
@DeerMichel FWIW this is where we do `process.versions.chrome` --> https://github.com/electron/electron/blob/master/atom/common/api/electron_bindings.cc#L107-L112

DeerMichel:
So one option would be replacing the dict key right at the file you pointed out @MarshallOfSound. I would consider this as a hacky way and prefer changing it in the embedded node build. See [this PR](https://github.com/electron/node/pull/105).

DeerMichel:
Shelved. See https://github.com/electron/electron/issues/18826#issuecomment-504534264

PR #18847, opened 2019-06-17
refactor: replace atom_version and friends with electron_version

MarshallOfSound:
Slowly but surely, removing atom from the code base 😆 

Notes: no-notes


atom/common/electron_version.h
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2013 GitHub, Inc.
```
MarshallOfSound: Git thinks this is a new file, but it's just renamed and changed a lot

nornagon: can we please generate this file during build?

MarshallOfSound:
> LGTM but I hate that electron_version.h isn't autogenerated.

I'm on that as well 👍 I wanna auto-generate all the things we bump in `bump-version.js` 👍

PR #18849, opened 2019-06-17
refactor: auto generate Info.plist to avoid bumping during releases

MarshallOfSound:
This was already possible (we were doing it for a different plist file).  Now we can remove a step in the bump version script 👍 

Notes: no-notes

### DIFF
--- a/script/bump-version.js
+++ b/script/bump-version.js
@@ -56,7 +56,6 @@ async function main () {
   // update all version-related files
   await Promise.all([
     updateVersion(version),
-    updateInfoPlist(version),
     updatePackageJSON(version),
     updateVersionH(components),
     updateWinRC(components)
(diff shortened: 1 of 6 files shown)


script/bump-version.js
```diff
@@ -119,18 +118,6 @@ async function updatePackageJSON (version) {
   })
 }
 
-// update CFBundle version information and overwrite pre-existing file
-// TODO(codebytere): provide these version fields at GN build time
-async function updateInfoPlist (version) {
-  const filePath = path.resolve(__dirname, '..', 'atom', 'browser', 'resources', 'mac', 'Info.plist')
-  const file = plist.parse(await readFile(filePath, { encoding: 'utf8' }))
```
poiru: @MarshallOfSound You could also remove `require('plist')` now.

PR #18854, opened 2019-06-17
fix: sanitize invalid custom protocol headers

### LINKED ISSUE #18126: Crash when passing an array for header value in registerFileProtocol handler
Here's an example [fiddle](https://github.com/electron/fiddle) that triggers the crash: https://gist.github.com/0cbf622b1b46ea55811f56df9b32df37.

It looks like we're just assuming that this value is a string, and we crash hard if anything other than a string is passed. It would be better for us to throw a JS error in this situation.

```
base::Value::GetString() const (in Electron Framework) (values.cc:319)
atom::URLRequestAsyncAsarJob::StartAsync(std::__1::unique_ptr<base::Value, std::__1::default_delete<base::Value> >, int) (in Electron Framework) (url_request_async_asar_job.cc:88)
base::internal::Invoker<base::internal::BindState<void (atom::URLRequestAsyncAsarJob::*)(std::__1::unique_ptr<base::Value, std::__1::default_delete<base::Value> >, int), base::WeakPtr<atom::URLRequestAsyncAsarJob>, std::__1::unique_ptr<base::Value, std::__1::default_delete<base::Value> >, int>, void ()>::RunOnce(base::internal::BindStateBase*) (in Electron Framework) (bind_internal.h:641)
base::TaskAnnotator::RunTask(char const*, base::PendingTask*) (in Electron Framework) (task_annotator.cc:114)
base::sequence_manager::internal::ThreadControllerWithMessagePumpImpl::DoWorkImpl(base::sequence_manager::LazyNow*, bool*) (in Electron Framework) (heap_profiler.h:93)
base::sequence_manager::internal::ThreadControllerWithMessagePumpImpl::DoSomeWork() (in Electron Framework) (thread_controller_with_message_pump_impl.cc:214)
base::MessagePumpKqueue::Run(base::MessagePump::Delegate*) (in Electron Framework) (message_pump_kqueue.cc:155)
non-virtual thunk to base::sequence_manager::internal::ThreadControllerWithMessagePumpImpl::Run(bool, base::TimeDelta) (in Electron Framework) (thread_controller_with_message_pump_impl.cc:0)
base::RunLoop::RunWithTimeout(base::TimeDelta) (in Electron Framework) (run_loop.cc:167)
content::BrowserProcessSubThread::IOThreadRun(base::RunLoop*) (in Electron Framework) (browser_process_sub_thread.cc:0)
base::Thread::ThreadMain() (in Electron Framework) (thread.cc:315)
base::(anonymous namespace)::ThreadFunc(void*) (in Electron Framework) (platform_thread_posix.cc:83)
```

NB. this is probably only a good first issue if you have some C++ experience :)

DeerMichel:
#### Description of Change
Part of fixes #18126 by throwing a js error instead of crashing.

cc @codebytere @erickzhao 

#### Discussion
- registerStreamProtocol uses stuff from `mate::Converter` to parse the headers... maybe thats a viable option here too as it avoids redundant code to get the headers dict
- are there any other methods of the protocol API that might crash due to invalid custom headers?

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/atom_api_protocol.h
+++ b/shell/browser/api/atom_api_protocol.h
@@ -99,8 +99,23 @@ class Protocol : public mate::TrackableObject<Protocol> {
   // Register the protocol with certain request job.
   template <typename RequestJob>
   void RegisterProtocol(const std::string& scheme,
-                        const Handler& handler,
+                        v8::Local<v8::Value> __handler__todo,
                         mate::Arguments* args) {
+    LOG(INFO) << scheme;
+
+    auto handler = std::make_unique<Handler>();
+    // Handler h1;
+    if (!mate::ConvertFromV8(args->isolate(), __handler__todo, handler.get())) {
+      args->ThrowError("TODO");
+      return;
+    }
+    LOG(INFO) << "worked";
+
+    if (scheme == "test") {
+      args->ThrowError("Aaaaaaa.");
+      return;
+    }
+
     CompletionCallback callback;
     args->GetNext(&callback);
     auto* getter = static_cast<URLRequestContextGetter*>(
@@ -108,7 +123,7 @@ class Protocol : public mate::TrackableObject<Protocol> {
     base::PostTaskWithTraitsAndReplyWithResult(
         FROM_HERE, {content::BrowserThread::IO},
         base::BindOnce(&Protocol::RegisterProtocolInIO<RequestJob>,
-                       base::RetainedRef(getter), isolate(), scheme, handler),
+                       base::RetainedRef(getter), isolate(), scheme, *handler),
         base::BindOnce(&Protocol::OnIOCompleted, GetWeakPtr(), callback));
   }
   template <typename RequestJob>
@@ -117,6 +132,12 @@ class Protocol : public mate::TrackableObject<Protocol> {
       v8::Isolate* isolate,
       const std::string& scheme,
       const Handler& handler) {
+    // if (isolate)
+    // isolate->ThrowException(v8::Exception::Error(mate::StringToV8(isolate,
+    // "my random error")));
+
+    // LOG(INFO) << "Why two times??";
+
     auto* job_factory = request_context_getter->job_factory();
     if (job_factory->IsHandledProtocol(scheme))
       return ProtocolError::REGISTERED;
--- a/shell/browser/api/atom_api_protocol.h
+++ b/shell/browser/api/atom_api_protocol.h
@@ -99,23 +99,8 @@ class Protocol : public mate::TrackableObject<Protocol> {
   // Register the protocol with certain request job.
   template <typename RequestJob>
   void RegisterProtocol(const std::string& scheme,
-                        v8::Local<v8::Value> __handler__todo,
+                        const Handler& handler,
                         mate::Arguments* args) {
-    LOG(INFO) << scheme;
-
-    auto handler = std::make_unique<Handler>();
-    // Handler h1;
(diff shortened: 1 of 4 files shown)


codebytere:
Let's add a test to this to prevent regression in the future: since it's a JS-level error you should now be able to check that a call with incorrect arguments throws and compare error messages :)

DeerMichel:
I actually tried... but didnt got my head around all the async stuff going on. Maybe tomorrow :)

atom/browser/net/url_request_async_asar_job.cc
```diff
@@ -38,6 +38,22 @@ void BeforeStartInUI(base::WeakPtr<URLRequestAsyncAsarJob> job,
     error = net::ERR_NOT_IMPLEMENTED;
   }
 
+  // sanitize custom headers
+  if (request_options && request_options->is_dict()) {
+    base::DictionaryValue* dict =
+        static_cast<base::DictionaryValue*>(request_options.get());
```
MarshallOfSound: We shouldn't have to static_cast the `base::Value` to a `DictionaryValue` here.  Once you've checked `is_dict` you should be able to use the `Value` methods.

Refs: https://cs.chromium.org/chromium/src/base/values.h?type=cs&q=setboolkey&sq=package:chromium&g=0&l=272-325

atom/browser/net/url_request_async_asar_job.cc
```diff
@@ -38,6 +38,22 @@ void BeforeStartInUI(base::WeakPtr<URLRequestAsyncAsarJob> job,
...
+    base::Value* headersValue =
+        dict->FindKeyOfType("headers", base::Value::Type::DICTIONARY);
+    if (headersValue) {
+      for (const auto& iter : headersValue->DictItems()) {
+        if (!iter.second.is_string()) {
+          args->ThrowError("Header value has to be a string");
```
MarshallOfSound: To be ultra helpful to users here we should include both the `key` that wasn't a string

atom/browser/net/url_request_async_asar_job.cc
DeerMichel: Same thing happens in `StartAsync`- should i simplify the code there too?

atom/browser/net/url_request_async_asar_job.cc
DeerMichel: done

atom/browser/net/url_request_async_asar_job.cc
```diff
@@ -38,6 +38,20 @@ void BeforeStartInUI(base::WeakPtr<URLRequestAsyncAsarJob> job,
     error = net::ERR_NOT_IMPLEMENTED;
   }
 
+  // sanitize custom headers
+  if (request_options && request_options->is_dict()) {
+    base::Value* headersDict = request_options->FindDictKey("headers");
```
MarshallOfSound: Can this be `const`?

```suggestion
    const base::Value* headersDict = request_options->FindDictKey("headers");
```

DeerMichel: Yes it can - but does that make a huuuge difference? ^^

codebytere:
@DeerMichel looks like this needs a rebase; once that's done this should be ready for action!

codebytere:
i see
```
<webview> tag <webview>.clearHistory() should clear the navigation history - should clear the navigation history
```

failing and it looks unrelated at first glance, so i think we're good here? would love a secondary review before we merge though!

PR #18865, opened 2019-06-18
fix: correct triggeredByAccelerator Event property behavior

### LINKED ISSUE #18808: bug: triggeredByAccelerator logic should not be tied to modifier key logic
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * v7.0.0-nightly.20190614
* **Operating System:**
  * macOS 10.14.5
  * none

### Expected Behavior
* Using an accelerator without a modifier to activate a menu item (e.g. pressing key 'B') should result in `triggeredByAccelerator: true`.
* Clicking on a menu item with a menu item (e.g. Shift+Click) should result in `triggeredByAccelerator:false`.

### Actual Behavior
* Using an accelerator without a modifier to activate a menu item (e.g. pressing key 'B') results in `triggeredByAccelerator: false`.
* Clicking on a menu item while holding down a modifier key (e.g. Shift+Click) results in `triggeredByAccelerator:true`.

### To Reproduce
Electron Fiddle gist: https://gist.github.com/19163d031cf3e35bb7b163bd2c39305c
```
{
    label: "Blade",
    id: "blade",
    accelerator: "B",
    click(a, b, e) {
        console.log(e);
    }
}
```

### Screenshots
image

### Additional Information

The property is currently derived from some event modifier flags (on Mac: `NSEvent.modifierFlags`). When using an accelerator with a modifier key, these flags are set and the property works as expected.

See:
```cpp
v8::Local<v8::Object> CreateEventFromFlags(v8::Isolate* isolate, int flags) {
  mate::Dictionary obj = mate::Dictionary::CreateEmpty(isolate);
  obj.Set("shiftKey", static_cast<bool>(flags & ui::EF_SHIFT_DOWN));
  obj.Set("ctrlKey", static_cast<bool>(flags & ui::EF_CONTROL_DOWN));
  obj.Set("altKey", static_cast<bool>(flags & ui::EF_ALT_DOWN));
  obj.Set("metaKey", static_cast<bool>(flags & ui::EF_COMMAND_DOWN));
  obj.Set("triggeredByAccelerator", static_cast<bool>(flags));
  return obj.GetHandle();
}
```
https://github.com/electron/electron/blob/master/atom/browser/api/event_emitter.cc#L85 

However, either triggering an accelerator without a modifier key (or clicking an accelerator with a modifier key) will give an invalid result. The `triggeredByAccelerator` logic should be decoupled from the modifier flags.

A better approach covering the edge case of not using a modifier is to use `NSEvent.type` to distinguish between a click or a keydown. Fix should go [here](https://github.com/electron/electron/blob/6f91af93433df4e9c0e7fb592e5b2dcb0b1c7888/atom/browser/ui/cocoa/atom_menu_controller.mm#L307).

cc @codebytere @erickzhao

erickzhao:
#### Description of Change
Fixes #18808 

Previously, the `triggeredByAccelerator` flag would be entirely coupled with whether or not the modifier keys were being used or not.

This PR swaps out the `ui::EventFlagsFromModifiers([event modifierFlags]))` call in the macOS code to `ui::EventFlagsFromNSEventWithModifiers(event, [event modifierFlags]))`. The latter outputs flags that take into account mouse click events on top of modifier flags ([see Chromium documentation](https://chromium.googlesource.com/chromium/src.git/+/75.0.3770.90/ui/events/cocoa/cocoa_event_utils.mm)).

The business logic to detect `triggeredByAccelerator` is then changed to exclude any mouse click flags.

Furthermore, a broken test was changed to properly detect that the accelerator registers properly and that the `triggeredByAccelerator` flag is properly detected.

#### Checklist

- [x] `npm test` passes

cc @DeerMichel @codebytere 

#### Release Notes

Notes: Fixed `triggeredByAccelerator` Event property behavior for `MenuItems`


atom/browser/api/event_emitter.cc
```diff
@@ -77,12 +77,15 @@ v8::Local<v8::Object> CreateCustomEvent(v8::Isolate* isolate,
 }
 
 v8::Local<v8::Object> CreateEventFromFlags(v8::Isolate* isolate, int flags) {
+  int click_flags = (ui::EF_RIGHT_MOUSE_BUTTON | ui::EF_LEFT_MOUSE_BUTTON |
+                     ui::EF_MIDDLE_MOUSE_BUTTON);
+  int mask = flags & click_flags;
```
ckerr: _(nit)_ It's a little confusing that the mask is named 'flags', the remaining flags are named 'mask', and that mask is not very descriptive when used below. I think we could simplify this:

```cc
const int mouseButtonMask = ui::EF_RIGHT_MOUSE_BUTTON | ...;
const bool isMouseClick = static_cast<bool>(flags & mouseButtonMask);
...
obj.set("triggeredByAccelerator", !isMouseClick);
```

To be clear, I think the code will work as-is. This suggestion is about making the code as straightforward as possible.

atom/browser/ui/cocoa/atom_menu_controller.mm
```diff
@@ -345,7 +345,7 @@ - (void)itemSelected:(id)sender {
   if (model) {
     NSEvent* event = [NSApp currentEvent];
     model->ActivatedAt(modelIndex,
-                       ui::EventFlagsFromModifiers([event modifierFlags]));
+                       ui::EventFlagsFromNSEventWithModifiers(event, [event modifierFlags]));
```
ckerr: :+1:

spec/api-menu-spec.js
```diff
@@ -860,11 +860,15 @@ describe('Menu module', () => {
...
-                // is triggered
-                Promise.resolve()
+              accelerator: 'T',
+              click: (a, b, event) => {
+                expect(event).to.deep.equal({
+                  shiftKey: false,
+                  ctrlKey: false,
+                  altKey: false,
+                  metaKey: false,
+                  triggeredByAccelerator: true
```
ckerr: The docs say these object properties are all optional -- but we're testing for them here in the specs. Are they actually optional? If not, should we remove that from the docs?

spec/api-menu-spec.js
```diff
@@ -860,11 +860,15 @@ describe('Menu module', () => {
           submenu: [
             {
               label: 'Test Item',
-              accelerator: 'Ctrl+T',
-              click: () => {
-                // Test will succeed, only when the menu accelerator action
-                // is triggered
-                Promise.resolve()
```
codebytere: Could we also have a test for the inverse, where `triggeredByAccelerator` _is_ true?

erickzhao: We are testing to see if it's true on [line 870](https://github.com/electron/electron/pull/18865/files/a7f946ea8bf681752e36763388178e15bf87d245#diff-0f899d7673156733456b3d561fe6c8feR870) here. Do you mean testing a manual click to see if `triggeredByAccelerator` will show up as false?

spec/api-menu-spec.js
```diff
@@ -860,11 +860,15 @@ describe('Menu module', () => {
...
+              accelerator: 'T',
+              click: (a, b, event) => {
+                expect(event).to.deep.equal({
```
ckerr: I don't think this click handler is actually getting tested?

E.G. If I change line 870 to `triggeredByAccelerator: false`, this test still passes.

spec/api-menu-spec.js
codebytere: My comment was intended to be in line with https://github.com/electron/electron/pull/18865/files#r295039742, but i think i wasn't super clear: i think that we should be checking to ensure that if we change line 870 to `triggeredByAccelerator: false`, this test should fail.

spec/api-menu-spec.js
erickzhao: I'm not sure what optional would mean in this context. :/

From my testing all of these properties are available in the Event object upon triggering the MenuItem, so I figured that I'd test the integrity of the entire object (e.g. all key flags should be false since I registered the accelerator as `T`).

cc @MarshallOfSound since I believe he originally worked on this part of the code in this commit: https://github.com/electron/electron/commit/3b74837020b15a6c7ef74677f78bcab01a7466a9

P.S. Is it normal that https://electronjs.org/docs/api/structures/keyboard-event gives a 404?

spec/api-menu-spec.js
erickzhao: Weird, the test seems to be working as intended on my end! See screen recording below (hope it's readable 😅):
screencast 2019-06-18 15-13-27

ckerr: Ah I see what's happening. I tested in `--ci` mode which for some reason is prevents the click handler from being reached. Interesting, but not this PR's issue.

erickzhao:
It seems like CI is failing the newly added tests on `linux-x64`, ARM, and Windows. A few notes on that after exploring the problem a bit:

* It seems like the accelerator isn't being triggered upon the RobotJS button press. The same test is failing everywhere. We're expecting the [`hasBeenClicked`](https://github.com/electron/electron/pull/18865/files#diff-0f899d7673156733456b3d561fe6c8feR882) flag to be switched to `true`, but that line of code is never triggered.
* The problem doesn't seem to be explicitly tied to the changes made in this PR. A [different branch](https://github.com/electron/electron/commit/136645a95c03d2d8e34d99156be9ae36ff815952) that fixed the test a bit differently also [failed CI in a similar way](https://circleci.com/gh/electron/electron/228978?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link).
* It doesn't seem to be explicitly tied to RobotJS. This PR also adds [a `before()` hook](https://github.com/electron/electron/pull/18865/files#diff-0f899d7673156733456b3d561fe6c8feR854) to the accelerator tests that skips them if the `globalShortcut` API doesn't work with RobotJS.

Will try to debug this on a Linux/Windows machine soon.

cc @DeerMichel @codebytere

erickzhao:
Hey y'all,

The root cause of the failing test was very much related to the problem that @ckerr was referring to here https://github.com/electron/electron/pull/18865#discussion_r295060221. After debugging the test code with @jkleinsc, it seems like the `--ci` flag hiding the app window during tests (see [`src/spec/static/main.js`](https://github.com/electron/electron/blob/master/spec/static/main.js#L130)) is causing weird interactions with RobotJS.

Observed behaviour: RobotJS logs its keystrokes into the console rather than into the test app (observe `tnot ok 10` and `Tok 11` in the screenshot below corresponding to the `T` and `Shift+T` key taps in the spec).
image

Things that have been attempted to fix this:
* Test without this PR's changes with the tests using the binary from `dist.zip` in `master` in the CI build.
* Create a new `BrowserWindow` on the Mocha `before()` hook and clean it up with the `after()` hook.
  * Set the window's options as `show: true` or `show: false`.
  * Calling `w.focus()` on the window `w` before running the tests.
  * Copy over the other instantiation options from `src/spec/static/main.js`.
  * N.B. This gets rid of the RobotJS key taps logging to console, but the tests still fail.

Elsewhere in the code, there's only one other instance of RobotJS, which also incidentally [skips CI on Mac and Windows](https://github.com/electron/electron/blob/master/spec/api-web-contents-spec.js#L386).

**Therefore, I'm advocating to just skip the CI for the accelerator tests for now. Would that be okay?**



cc @DeerMichel @codebytere @ckerr @jkleinsc

codebytere:
@erickzhao this will probably need some manual backports to 6/5/4

PR #18866, opened 2019-06-18
refactor: auto generate electron_version.h from the version file

MarshallOfSound:
Remove another bit of logic from our version bump script, we can auto generate the version header as well.

Notes: no-notes


deepak1556:
Is it possible to reuse `build/util/process_version.gni` and `build/util/version.py` ? No hard rule, just want to see if we can avoid some templates.

MarshallOfSound:
@deepak1556 I looked into `build/util/process_version.gni` but it did what this templating logic does but more complicated and doesn't let us provide values from `semver` and doesn't understand prerelease tags 🤷‍♂ Seemed more straightforward to build this templater in our code

deepak1556:
ah okay if we couldn't customize it, then not useful. Thanks for looking into it.

build/templates/electron_version.tmpl
```diff
@@ -31,4 +31,4 @@
 
 #define ELECTRON_VERSION "v" ELECTRON_VERSION_STRING
```
nornagon: now that this is autogenerated we can get rid of these silly preprocessor hacks and just write

```cpp
#define ELECTRON_VERSION "v$major.$minor.$patch"
```
or so?

or even just add the full semver to the keys and write

```cpp
#define ELECTRON_VERSION "v$semver"
```

build/templated_file.gni
```diff
@@ -0,0 +1,35 @@
...
+      invoker.output,
+    ]
+    script = "//electron/build/generate-template.py"
+    args = [
+      rebase_path(invoker.template),
+      rebase_path(invoker.output),
+      rebase_path(args_path),
+    ]
+
+    if (defined(invoker.extra_args)) {
```
nornagon: rename this to `args_files` or something to make it clear these are supposed to be file paths.

BUILD.gn
```diff
@@ -307,6 +308,34 @@ source_set("manifests") {
...
+  script = "generate-version-json"
+
+  outputs = [
+    "$target_gen_dir/electron_version.args",
+  ]
+
+  args = [ rebase_path(outputs[0]) ]
+
+  inputs = [
+    "script/generate-version-json.js",
```
nornagon: also `ELECTRON_VERSION`?

BUILD.gn
```diff
@@ -307,6 +308,32 @@ source_set("manifests") {
...
 }
 
+npm_action("electron_version_args") {
+  script = "generate-version-json"
+
+  outputs = [
+    "$target_gen_dir/electron_version.args",
+  ]
+
+  args = [ rebase_path(outputs[0]) ]
```
nornagon:
```suggestion
  args = rebase_path(outputs)
```

`rebase_path` can take an array & return an array

PR #18873, opened 2019-06-19
docs: missing instructions for installing electron in first-app.md

ericchase:
#### Description of Change
The 'Electron Development in a Nutshell' section seems to require 
`electron` to be installed globally. I have added instructions to do 
so in the 'Installing Electron' section.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: no-notes

PR #18874, opened 2019-06-19
docs: added instructions under the installing electron section

ericchase:
#### Description of Change
The 'Writing Your First Electron App' section describes the process 
of turning a Node application into an Electron application. The change 
requires Electron to be installed globally in order to work. I have 
added a __note__ under the 'Writing Your First Electron App' section, 
and extra instructions under the 'Installing Electron' section, so 
that new users will not be completely confused when running
`npm start` results in an error.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/docs/tutorial/first-app.md
+++ b/docs/tutorial/first-app.md
@@ -86,6 +86,11 @@ run the following command from your app's directory:
 npm install --save-dev electron
 ```
 
+To install `electron` globally, run the following command:
+```sh
+npm install -g electron
+```
+
 Other means for installing Electron exist. Please consult the
 [installation guide](installation.md) to learn about use with proxies, mirrors,
 and custom caches.
--- a/docs/tutorial/first-app.md
+++ b/docs/tutorial/first-app.md
@@ -73,6 +73,8 @@ merely replace the `node` runtime with the `electron` runtime.
 }
 ```
 
+__Note__: The above change assumes that Electron is installed globally.
+
 ## Installing Electron
 
 At this point, you'll need to install `electron` itself. The recommended way


malept:
Hi @ericchase, as I mentioned in #18873, this tutorial does not require installing Electron globally. When you run `npm install --save-dev electron`, it installs Electron to the Node.js project in the `node_modules` folder, and then when you run `npm start`, NPM will first search `node_modules/.bin` for an `electron` executable, where Electron has installed a script to run the prebuilt binary.

ericchase:
> 
> 
> Hi @ericchase, as I mentioned in #18873, this tutorial does not require installing Electron globally. When you run `npm install --save-dev electron`, it installs Electron to the Node.js project in the `node_modules` folder, and then when you run `npm start`, NPM will first search `node_modules/.bin` for an `electron` executable, where Electron has installed a script to run the prebuilt binary.

This pull request is not concerned with that tutorial. I have made the corrections, and you must have misread this time or decided not to read it. I will open another pull request. Please don't close it again per-maturaley.

ericchase:
> 
> 
> Hi @ericchase, as I mentioned in #18873, this tutorial does not require installing Electron globally. When you run `npm install --save-dev electron`, it installs Electron to the Node.js project in the `node_modules` folder, and then when you run `npm start`, NPM will first search `node_modules/.bin` for an `electron` executable, where Electron has installed a script to run the prebuilt binary.

I think I understand what you're saying, now. However, this is not the case on **Windows** or **Windows Subsystem for Linux with Ubuntu**. `npm start` does NOT search the `node_modules/.bin` folder, but rather throws an error that "`electron` is not recognized as an internal or external command," which is standard for trying to run an executable that does not exist.

My installations of node and npm are up to date and current. Perhaps the outcome is different on a pure linux or mac machine.

ericchase:
It appears that you *must* run `npm install` on a Windows command line in order for `npm start` to have access to the `node_modules/.bin` folder. If `npm install` is run on a WSL bash console, it seems that the wrong file permissions are written.

After running `npm install` from a Windows command console, `npm start` now works as you have described. I have always used a WSL bash console to work with node/npm, and only used the Windows command console to run `npm start` as `electron` cannot run from WSL.

ericchase:
Now that the real issue has been found, I wonder if this should be mentioned in the tutorial at all. Some unfortunate developers like myself may happen to run into this situation, but it is more of a node/npm issue.

malept:
WSL does not work with Electron, since there is no GUI (I believe it will try to install the Linux version... which requires X11).

I think https://electronjs.org/docs/tutorial/development-environment#setting-up-windows should be updated discouraging folks from using WSL.

ericchase:
I agree that WSL should be discouraged for Electron projects.

@malept 
Sorry for the double pull request and misunderstanding of what you said before. I've not had a similar issue with `npm start` before, and didn't even consider there being an issue between Windows and WSL.

Issue #13820, opened 2018-07-26
tray.popUpContextMenu blocks javascript main thread on mac


pjebs:
* Electron Version: latest
* Operating System (Platform and Version): mac
* Last known working Electron version: n/a

**Expected Behavior**
When the tray icon menu is popped up manually (but on a tray click event callback), I expect the menu to open up but not necessarily block the main process.

**Actual behavior**
It blocks the main process until the menu is closed. This is creating issues with other timers etc that I have running on the main process.

pjebs:
An asyn version would be great

pjebs:
I think it's related to https://github.com/electron/electron/issues/1854 but my issue has nothing to do with render process.

pjebs:
would this be hard to implement for tray: https://github.com/electron/electron/pull/8702

pjebs:
@kevinsawicki

pjebs:
Try `popUpContextMenu` on a `tray click event callback`

ckerr:
if one of y'all could make a fork of https://github.com/electron/electron-quick-start that repros this, that would be great. I like @fantasticsid's approach of using a simple timer to display whether or not something's blocked!

pjebs:
```
const {app, Menu, Tray, BrowserWindow } = require('electron')

let tray = null
app.on('ready', () => {
    tray = new Tray("/Users/pj/Developer/eyecumber/desktop-app/app/assets/img/tray/trayTemplate.png")
    const contextMenu = Menu.buildFromTemplate([
        {label: 'Item1', type: 'radio'},
        {label: 'Item2', type: 'radio'},
        {label: 'Item3', type: 'radio', checked: true},
        {label: 'Item4', type: 'radio'}
    ])

     tray.on('click', () => {
      console.log("A")
    tray.popUpContextMenu(contextMenu)

    console.log("B")
  })

    let win = new BrowserWindow();
    win.loadURL('https://www.youtube.com');

    setInterval(() => { console.log('test'); }, 1000);
})
```

`console.log('b')` is not printed out after menu is closed.

pjebs:
For my project, the timers stop running but in the example above the timer works in the main thread. I can't explain that..... but the `popupContextMenu( )` is still synchronous

pjebs:
[...]

PR #18880, opened 2019-06-19
fix: make tray not block main process

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/13820.

When running [this gist](https://gist.github.com/codebytere/f600a0c13a336859251330bd77d924a7), previously `A` would print and then `B` would only print to console after the tray had been closed. This PR fixes that issue by posting a task and allowing events to be pumped while the tray is open. 

After this fix, `A` and `B` are printed successively to console while the tray is open:

Screen Shot 2019-06-19 at 9 10 08 AM

cc @miniak @deepak1556 @MarshallOfSound @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where the Tray icon context menu on macOS blocked the main process while open.

### DIFF
--- a/shell/browser/ui/atom_menu_model.h
+++ b/shell/browser/ui/atom_menu_model.h
@@ -13,7 +13,8 @@
 
 namespace electron {
 
-class AtomMenuModel : public ui::SimpleMenuModel {
+class AtomMenuModel : public ui::SimpleMenuModel,
+                      public base::RefCounted<AtomMenuModel> {
  public:
   class Delegate : public ui::SimpleMenuModel::Delegate {
    public:
@@ -48,7 +49,6 @@ class AtomMenuModel : public ui::SimpleMenuModel {
   };
 
   explicit AtomMenuModel(Delegate* delegate);
-  ~AtomMenuModel() override;
 
   void AddObserver(Observer* obs) { observers_.AddObserver(obs); }
   void RemoveObserver(Observer* obs) { observers_.RemoveObserver(obs); }
@@ -69,6 +69,9 @@ class AtomMenuModel : public ui::SimpleMenuModel {
   AtomMenuModel* GetSubmenuModelAt(int index);
 
  private:
+  friend class base::RefCounted<AtomMenuModel>;
+  ~AtomMenuModel() override;
+
   Delegate* delegate_;  // weak ref.
 
   std::map<int, base::string16> roles_;  // command id -> role
--- a/shell/browser/ui/tray_icon_cocoa.mm
+++ b/shell/browser/ui/tray_icon_cocoa.mm
@@ -8,7 +8,11 @@
 #include <vector>
 
 #include "base/mac/sdk_forward_declarations.h"
+#include "base/message_loop/message_loop.h"
 #include "base/strings/sys_string_conversions.h"
+#include "base/task/post_task.h"
+#include "content/public/browser/browser_task_traits.h"
+#include "content/public/browser/browser_thread.h"
 #include "shell/browser/mac/atom_application.h"
 #include "shell/browser/ui/cocoa/NSString+ANSI.h"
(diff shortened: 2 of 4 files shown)


shell/browser/ui/tray_icon_cocoa.mm
```diff
@@ -489,9 +498,18 @@ - (BOOL)shouldHighlight {
   return [status_item_view_ getIgnoreDoubleClickEvents];
 }
 
+void TrayIconCocoa::PopUpOnUI(scoped_refptr<AtomMenuModel> menu_model) {
+  auto* model = menu_model.get();
+  if (model)
```
miniak: you can do
```
if (auto* model = menu_model.get())
  [status_item_view_ popUpContextMenu:model];
```

miniak: you are not accessing `this`, should you make `PopUpOnUI` static?

shell/browser/ui/atom_menu_model.h
```diff
@@ -13,7 +13,8 @@
 
 namespace electron {
 
-class AtomMenuModel : public ui::SimpleMenuModel {
+class AtomMenuModel : public ui::SimpleMenuModel,
+                      public base::RefCounted<AtomMenuModel> {
```
deepak1556: Not a big fan of reference counting, and in this specific case there doesn't seem to be a necessity. The task is not posted against a different sequence, rather just passes the object around and there is null check in `popUpContextMenu` which should be sufficient.

shell/browser/ui/tray_icon_cocoa.mm
```diff
@@ -349,6 +356,8 @@ - (void)popUpContextMenu:(electron::AtomMenuModel*)menu_model {
   if (menuController_ && ![menuController_ isMenuOpen]) {
     // Redraw the tray icon to show highlight if it is enabled.
     [self setNeedsDisplay:YES];
+
+    base::mac::ScopedSendingEvent sendingEventScoper;
```
deepak1556: which event does this defer ?

shell/browser/ui/tray_icon_cocoa.mm
```diff
@@ -333,13 +337,16 @@ - (void)mouseUp:(NSEvent*)event {
 }
 
 - (void)popUpContextMenu:(electron::AtomMenuModel*)menu_model {
+  base::MessageLoopCurrent::ScopedNestableTaskAllower allow;
   // Show a custom menu.
```
deepak1556: Do we also want to set `ScopedPumpMessagesInPrivateModes` ?

shell/browser/ui/atom_menu_model.h
codebytere: @deepak1556 the reason I chose to refcount it was that a valid pointer can be posted to the sequence and then synchronously destroyed from from JavaScript in situations like this:
```js
  const tray = new Tray(/path/to/icon)

  const contextMenu = Menu.buildFromTemplate([
    { label: 'Item1', type: 'radio' },
    { label: 'Item2', type: 'radio' },
    { label: 'Item3', type: 'radio', checked: true },
    { label: 'Item4', type: 'radio' }
  ])

  tray.popUpContextMenu(contextMenu)
  tray.destroy()
```

which results in a use-after-free crash and which a null check would do nothing to prevent since it'd be a valid pointer and thus pass the check.

deepak1556: @codebytere that shouldn't be a case to keep the class alive rather if the user destroyed it then we should respect it. We should use a `base::WeakPtr` rather than refcounting to have the right behavior.

codebytere: ah ok! i can address that here: https://github.com/electron/electron/pull/18977

PR #18890, opened 2019-06-19
build: add an FYI job for building with RUN_AS_NODE disabled

MarshallOfSound:
Does what it says on the tin, adds a CI job to build electron with `enable_run_as_node=false`.  Will prevent #18887 again.  I don't think we should add jobs for every combination but this seems like a reasonable one to test

Notes: no-notes


deepak1556:
I would suggest adding a windows one if possible :)

MarshallOfSound:
> I would suggest adding a windows one if possible :)

We're currently very resource-limited on windows, we can do linux builds easily, windows not so much (for now)

Issue #17189, opened 2019-03-01
session.getCacheSize dose not work


rocka:
* **Electron Version:**
  * 3.1.4 and 4.0.6
* **Operating System:**
  * Arch Linux 4.20.13-arch1-1-ARCH x86_64
* **Last Known Working Electron version:**
  * 2.0.17

### Expected Behavior

get cache size using

```js
const {BrowserWindow} = require('electron')
const w = new BrowserWindow
w.loadURL('https://electronjs.org')
w.webContents.session.getCacheSize(size => {
  console.log(size) // -> output some random cache size, eg. 669451
})
```

### Actual Behavior

Can't get cache size, the callback never invoked.

### To Reproduce

make a temp test directory and do as follows:

```console
npm init -y
npm i electron@3 # or npm i electron@4
npx electron -i
> const {BrowserWindow} = require('electron')
undefined
> const w = new BrowserWindow
undefined
> w.loadURL('https://electronjs.org')
true
> w.webContents.session.getCacheSize(consloe.log)
undefined
# no cache size output here ...
```

### Additional Information

this problem may exist since 3.0

codebytere:
Did a quick check on this and it looks like we get locked into a pending state which prevents the asynchrous operation from completing.

That is, we're here
https://github.com/electron/electron/blob/aa8b66aae175fa24ec097a9aae9657f9b609873e/atom/browser/api/atom_api_session.cc#L275

and `rv == net::ERR_IO_PENDING` so the callback doesn't run. i'll try to dig in some more soon.

simongregory:
I'm getting this issue on the Mac with v5.0.4 (am noting because it's only labelled as an Linux problem).

codebytere:
This was fixed in https://github.com/electron/electron/pull/17969.

i'll try to backport it soon.

codebytere:
Closing as this has been backported to stable supported lines.

PR #17969, opened 2019-04-25
refactor: session.{clearCache,getCacheSize} nws13n

nornagon:
#### Description of Change
Refactors session.{clearCache,getCacheSize} to use the network service

#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: no-notes

PR #18898, opened 2019-06-20
refactor: session.{clearCache,getCacheSize} nws13n

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/17189.

Backport of https://github.com/electron/electron/pull/17969. We haven't promisified this in 5 so this is a callbackified version of that PR.

cc @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix an issue where `session.getCacheSize` would never call its callback.


codebytere:
Failing spec on linux is:
```
webContents module create() does not crash on exit - does not crash on exit
```
which is not affected by the codepaths changed in this PR and which i would therefore say is a flake.

PR #18903, opened 2019-06-20
[WIP] feat: exit non-silently when running as root on Linux

### LINKED ISSUE #18547: Electron exits silently when running as root on Linux
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 6.0.0-beta.5, 5.0.2
* **Operating System:**
  * Linux - Ubuntu 18.04 x64
  * 4.2.3

### Expected Behavior

Electron should warn root users that `Running as root without --no-sandbox is not supported.`, even if ELECTRON_ENABLE_LOGGING is not set

### Actual Behavior

Running Electron as root fails silently, unless the user sets the environment variable `ELECTRON_ENABLE_LOGGING=true`.  When `ELECTRON_ENABLE_LOGGING=true`, the console displays the message: 
```sh
Running as root without --no-sandbox is not supported.
```
This behaviour is really confusing as causes people to think Electron doesn't work on their platform (eg see #18350) We should provide more context here.

### To Reproduce
As root run:
```sh
$ git clone https://github.com/electron/electron-quick-start.git
$ npm install
$ npm start 
```
No output.

Again as root run: 
```sh
$ export ELECTRON_ENABLE_LOGGING=true
$ npm start
```
Warning message appears

erickzhao:
#### Description of Change
Fixes #18547 

**_N.B. WIP because although I got the CI to pass, I haven't actually tested this in a Linux VM yet._**

Previously, running as `root` on Linux without `--no-sandbox` would fail silently if Electron logging were disabled. The exit is thrown by [`zygote_host_impl_linux.cc`](https://cs.chromium.org/chromium/src/services/service_manager/zygote/host/zygote_host_impl_linux.cc?l=88) at the Chromium level with a `LOG(ERROR)` to provide feedback to the user. If Electron logging were to be disabled, then this would mean that the user would not be provided with the appropriate log.

The goal of the PR is to replicate the business logic done at the Chromium level. This WIP so far adds in a check for if the user is root by using [`getuid`](http://man7.org/linux/man-pages/man2/getuid.2.html) and exiting the program with `LOG(FATAL)` if they don't have `--no-sandbox` enabled. This will ensure that the error will be thrown before the app crashes.

Some notes:
1) I used `getuid` instead of the `sandbox::Credentials::GetRESIds` call in `zygote_host_impl_linux.cc` because we were not drawing `"sandbox/linux/services/` from Chromium. I tried pasting in the relevant function, but it led to a rabbit hole of definitions that were not included in our dependencies.
2) From my testing, `LOG(FATAL)` still logs into Electron Fiddle even when logging is disabled. Would this be enough to solve the issue?

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added error message on Linux when running as root user without `--no-sandbox` flag.


DeerMichel:
Closed in favor of #19216.

PR #18916, opened 2019-06-21
feat: app.moveToApplicationsFolder conflict handling

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/18805.

We want to keep default move conflict handling behavior in that it's still what most users would expect, but there exist edge cases in which users may not want to be forced into that behavior.

This thus introduces an optional conflict handler that allows developers access to more granular move actions. They could now allow the user to choose whether to delete an existing app in favor of the current one being moved, or whether to quit the current app and focus on the existing one should it both exist _and_ be running. I added a fair amount of new documentation outlining this behavior, but if there are things users may benefit from seeing examples of or nuances that should be added please leave feedback!

cc @MarshallOfSound @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added an optional conflict handling callback to `app.moveToApplicationsFolder`.

### DIFF
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -1248,11 +1248,11 @@ For example:
 ```js
 app.moveToApplicationsFolder(conflictType => {
   if (conflictType === 'exists') {
-    dialog.showMessageBox({
-    type: 'question',
-    buttons: ['Halt Move', 'Continue Move'],
-    defaultId: 0,
-    message: 'An app of this name already exists',
+    dialog.showMessageBoxSync({
+      type: 'question',
+      buttons: ['Halt Move', 'Continue Move'],
+      defaultId: 0,
+      message: 'An app of this name already exists'
     }, response => response)
   }
 })
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -1223,7 +1223,10 @@ This method can only be called before app is ready.
 Returns `Boolean` - Whether the application is currently running from the
 systems Application folder. Use in combination with `app.moveToApplicationsFolder()`
 
-### `app.moveToApplicationsFolder()` _macOS_
+### `app.moveToApplicationsFolder([handler])` _macOS_
+
+* `handler` Function (optional) - A handler for potential conflict in move failure.
+  * `conflictType` String - the type of move conflict encountered by the handler; can be `exists` or `existsAndRunning`, where `exists` means that an app of the same name is present in the Applications directory and `existsAndRunning` means both that it exists and that it's presently running.
 
 Returns `Boolean` - Whether the move was successful. Please note that if
 the move is successful, your application will quit and relaunch.
@@ -1236,7 +1239,26 @@ the user to confirm the operation, you may do so using the
 move to fail. For instance if the user cancels the authorization dialog, this
 method returns false. If we fail to perform the copy, then this method will
 throw an error. The message in the error should be informative and tell
-you exactly what went wrong
+you exactly what went wrong.
+
+By default, if an app of the same name as the one being moved exists in the Applications directory and is _not_ running, the existing app will be trashed and the active app moved into its place. If it _is_ running, the pre-existing running app will assume focus and the the previously active app will quit itself. This behavior can be changed by invoking the optional callback handler, where the boolean returned by the handler determines whether or not the move proceeds with its default behavior.
+
+For example:
+
+```js
+app.moveToApplicationsFolder(conflictType => {
+  if (conflictType === 'exists') {
+    dialog.showMessageBox({
+    type: 'question',
+    buttons: ['Halt Move', 'Continue Move'],
+    defaultId: 0,
+    message: 'An app of this name already exists',
+    }, response => response)
+  }
+})
+```
+
+Would mean that if an app already exists in the user directory, if the user chooses to 'Continue Move' then the function would continue with its default behavior and the existing app will be trashed and the active app moved into its place.
 
 ## Properties
 
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -1226,7 +1226,7 @@ systems Application folder. Use in combination with `app.moveToApplicationsFolde
 ### `app.moveToApplicationsFolder([options])` _macOS_
 
 * `options` Object (optional)
-  * `conflictHandler` Function (optional) - A handler for potential conflict in move failure.
+  * `conflictHandler` Function<Boolean> (optional) - A handler for potential conflict in move failure.
     * `conflictType` String - the type of move conflict encountered by the handler; can be `exists` or `existsAndRunning`, where `exists` means that an app of the same name is present in the Applications directory and `existsAndRunning` means both that it exists and that it's presently running.
 
 Returns `Boolean` - Whether the move was successful. Please note that if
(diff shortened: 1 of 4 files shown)


docs/api/app.md
```diff
@@ -1235,7 +1238,26 @@ the user to confirm the operation, you may do so using the
...
+you exactly what went wrong.
+
+By default, if an app of the same name as the one being moved exists in the Applications directory and is _not_ running, the existing app will be trashed and the active app moved into its place. If it _is_ running, the pre-existing running app will assume focus and the the previously active app will quit itself. This behavior can be changed by invoking the optional callback handler, where the boolean returned by the handler determines whether or not the move proceeds with its default behavior.
+
+For example:
+
+```js
+app.moveToApplicationsFolder(conflictType => {
+  if (conflictType === 'exists') {
+    dialog.showMessageBox({
```
MarshallOfSound: This has to be sync right?

docs/api/app.md
```diff
@@ -1222,7 +1222,10 @@ This method can only be called before app is ready.
 Returns `Boolean` - Whether the application is currently running from the
 systems Application folder. Use in combination with `app.moveToApplicationsFolder()`
 
-### `app.moveToApplicationsFolder()` _macOS_
+### `app.moveToApplicationsFolder([handler])` _macOS_
+
+* `handler` Function (optional) - A handler for potential conflict in move failure.
```
MarshallOfSound: To allow cleaner API expansion this should probably be an `options` object with a single property `conflictHandler`

shell/browser/ui/cocoa/atom_bundle_mover.mm
```diff
@@ -70,11 +90,22 @@
...
+    base::OnceCallback<bool(ConflictType)> conflict_cb;
+
     // If a copy already exists in the Applications folder, put it in the Trash
     if ([fileManager fileExistsAtPath:destinationPath]) {
       // But first, make sure that it's not running
       if (IsApplicationAtPathRunning(destinationPath)) {
-        // Give the running app focus and terminate myself
+        // Check for callback handler and get user choice for open/quit
+        if (args->GetNext(&conflict_cb)) {
+          return false;
```
MarshallOfSound: I think this got left here by accident

shell/browser/ui/cocoa/atom_bundle_mover.mm
```diff
@@ -70,11 +90,22 @@
       }
     }
   } else {
+    base::OnceCallback<bool(ConflictType)> conflict_cb;
```
MarshallOfSound: What happens if this method returns something that isn't a `bool` (like a string or something).  I have a terrible feeling it's gonna throw a weird mate error or something

codebytere: yeah, i think this is gonna need to be 
```
base::OnceCallback<v8::MaybeLocal<v8::Boolean>(ConflictType)> conflict_cb;
```

docs/api/app.md
```diff
@@ -1222,7 +1222,11 @@ This method can only be called before app is ready.
 Returns `Boolean` - Whether the application is currently running from the
 systems Application folder. Use in combination with `app.moveToApplicationsFolder()`
 
-### `app.moveToApplicationsFolder()` _macOS_
+### `app.moveToApplicationsFolder([options])` _macOS_
+
+* `options` Object (optional)
+  * `conflictHandler` Function (optional) - A handler for potential conflict in move failure.
```
MarshallOfSound: We can declare the return type 👍 

```suggestion
  * `conflictHandler` Function<Boolean> (optional) - A handler for potential conflict in move failure.
```

docs/api/app.md
```diff
@@ -1235,7 +1239,28 @@ the user to confirm the operation, you may do so using the
 move to fail. For instance if the user cancels the authorization dialog, this
 method returns false. If we fail to perform the copy, then this method will
 throw an error. The message in the error should be informative and tell
-you exactly what went wrong
+you exactly what went wrong.
+
+By default, if an app of the same name as the one being moved exists in the Applications directory and is _not_ running, the existing app will be trashed and the active app moved into its place. If it _is_ running, the pre-existing running app will assume focus and the the previously active app will quit itself. This behavior can be changed by invoking the optional callback handler, where the boolean returned by the handler determines whether or not the move conflict is resolved with default behavior.
```
MarshallOfSound:
```suggestion
By default, if an app of the same name as the one being moved exists in the Applications directory and is _not_ running, the existing app will be trashed and the active app moved into its place. If it _is_ running, the pre-existing running app will assume focus and the the previously active app will quit itself. This behavior can be changed by providing the optional conflict handler, where the boolean returned by the handler determines whether or not the move conflict is resolved with default behavior.  i.e. returning `false` will ensure no further action is taken, returning `true` will result in the default behavior and the method continuing. 
```

docs/api/app.md
```diff
@@ -1235,7 +1239,28 @@ the user to confirm the operation, you may do so using the
...
+
+For example:
+
+```js
+app.moveToApplicationsFolder({
+  conflictHandler: conflictType => {
+    if (conflictType === 'exists') {
+      dialog.showMessageBoxSync({
```
MarshallOfSound: This won't work if I'm reading this right, need to do `return dialog.showMessageBoxSync(...) === 1`, sync doesn't have a callback

shell/browser/ui/cocoa/atom_bundle_mover.h
```diff
@@ -11,6 +11,9 @@
 
 namespace electron {
 
+// Possible bundle movement conflicts
+enum class ConflictType { EXISTS, EXISTS_AND_RUNNING };
```
MarshallOfSound: Just to make this slight more scoped than "an electron conflict"

```suggestion
enum class BundlerMoverConflictType { EXISTS, EXISTS_AND_RUNNING };
```

docs/api/app.md
codebytere: oh derp yes

codebytere:
@MarshallOfSound it throws this error:
for both 1) non-boolean return types and 2) manually thrown Errors 

Screen Shot 2019-07-01 at 9 29 00 PM

codebytere:
@MarshallOfSound this should be all set for final review!

fabiospampinato:
@codebytere I feel this PR completely solves a specific issue: being able to show a dialog in case a conflict arises.

But if an app is already present _and_ running I guess what most people would want to do is quit the old app and then move the new one in, this PR at best would allow me to show the user an error, but what if I just want to quit the old app and proceed? Which would be the most common scenario under this situation IMHO. Is/shoudn't there be an easy way to do that?

And shouldn't an asynchronous `conflictHandler` function be supported too? Say I want to kill the other app via the shell and I want to wait for it to be actually killed before proceeding.

Also if I don't provide a `conflictHandler` is this method still returning `true` in conflicting situations? That would go against the docs that say that an error should be thrown instead: "If we fail to perform the copy, then this method will throw an error."

codebytere:
@fabiospampinato this PR was created to address some use cases; it's unlikely we ever be able to cover all of them.

> I don't provide a conflictHandler is this method still returning true in conflicting situations? 
No, it does not change previous behavior. All new behavior is _optional_ as documented.

fabiospampinato:
> No, it does not change previous behavior. 

The previous behavior is to return `true`, so shouldn't the answer here be "yes" and either the behavior or the docs updated as the two are conflicting? The app is not moved and no error is thrown, which was point of my issue that has been closed because of this PR.

Issue #18922, opened 2019-06-21
Add a `BrowserWindows.moveAbove(mediaSourceId)` method


CapOM:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description
BrowserWindow.{focus,blur,moveTop}() are not enough in some situations. For example when implementing an overlay that follows another window that can lose focus. In that case it is useful to move the overlay above the tracked window.

### Proposed Solution
Implement a new method `BrowserWindows.moveAbove(mediaSourceId)` where mediaSourceId is an id in the format of DesktopCapturerSource's id, see https://electronjs.org/docs/api/structures/desktop-capturer-source#desktopcapturersource-object

### Alternatives Considered
No other alternatives but I considered making a new separate object for mediaSourceId instead of a string but then we would have to change type of DesktopCapturerSource's id. In the end it is just a string so I opted for a string.

### Additional Information
Someone else requested to implement `BrowserWindows.getMediaSourceId()` https://github.com/electron/electron/issues/16460 so this is related in the sense that moveAbove would take a mediaSourceId in parameter whereas  getMediaSourceId would return a mediaSourceId. In any case it seems there are some interest around that.

miniak:
@CapOM please use the checklist (`[ ]` and `[x]`) properly

CapOM:
Sorry can you clarify ? I am sure I am missing something but I do not see it. Thx!

PR #18926, opened 2019-06-21
feat: Implement BrowserWindow.getMediaSourceId() and BrowserWindow.moveAbove(mediaSourceId)

CapOM:
#### Description of Change

BrowserWindow.{focus,blur,moveTop}() are not enough in some
situations. For example when implementing an overlay that
follows another window that can lose focus. In that case
it is useful to move the overlay above the tracked window.

sourceId is a string in the format of DesktopCapturerSource.id,
for example "window:1869:0".

Resolves https://github.com/electron/electron/issues/18922.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added BrowserWindow.getMediaSourceId() and BrowserWindow.moveAbove(mediaSourceId)

### DIFF
--- a/shell/browser/api/atom_api_top_level_window.cc
+++ b/shell/browser/api/atom_api_top_level_window.cc
@@ -730,6 +730,11 @@ void TopLevelWindow::RemoveBrowserView(v8::Local<v8::Value> value) {
     }
   }
 }
+
+std::string TopLevelWindow::GetMediaSourceId() const {
+  return window_->GetDesktopMediaID().ToString();
+}
+
 v8::Local<v8::Value> TopLevelWindow::GetNativeWindowHandle() {
   // TODO(MarshallOfSound): Replace once
   // https://chromium-review.googlesource.com/c/chromium/src/+/1253094/ has
@@ -1135,6 +1140,7 @@ void TopLevelWindow::BuildPrototype(v8::Isolate* isolate,
       .SetMethod("setBrowserView", &TopLevelWindow::SetBrowserView)
       .SetMethod("addBrowserView", &TopLevelWindow::AddBrowserView)
       .SetMethod("removeBrowserView", &TopLevelWindow::RemoveBrowserView)
+      .SetMethod("getMediaSourceId", &TopLevelWindow::GetMediaSourceId)
       .SetMethod("getNativeWindowHandle",
                  &TopLevelWindow::GetNativeWindowHandle)
       .SetMethod("setProgressBar", &TopLevelWindow::SetProgressBar)
--- a/shell/browser/api/atom_api_top_level_window.cc
+++ b/shell/browser/api/atom_api_top_level_window.cc
@@ -552,6 +552,11 @@ std::vector<int> TopLevelWindow::GetPosition() {
   result[1] = pos.y();
   return result;
 }
+void TopLevelWindow::MoveAbove(const std::string& sourceId,
+                               mate::Arguments* args) {
+  if (!window_->MoveAbove(sourceId))
+    args->ThrowError("Invalid media source id");
+}
 
 void TopLevelWindow::MoveTop() {
   window_->MoveTop();
@@ -1081,6 +1086,7 @@ void TopLevelWindow::BuildPrototype(v8::Isolate* isolate,
       .SetMethod("setMaximumSize", &TopLevelWindow::SetMaximumSize)
       .SetMethod("getMaximumSize", &TopLevelWindow::GetMaximumSize)
       .SetMethod("setSheetOffset", &TopLevelWindow::SetSheetOffset)
+      .SetMethod("moveAbove", &TopLevelWindow::MoveAbove)
       .SetMethod("moveTop", &TopLevelWindow::MoveTop)
       .SetMethod("_setResizable", &TopLevelWindow::SetResizable)
       .SetMethod("_isResizable", &TopLevelWindow::IsResizable)
--- a/shell/browser/native_window_views.cc
+++ b/shell/browser/native_window_views.cc
@@ -1117,6 +1117,37 @@ bool NativeWindowViews::IsVisibleOnAllWorkspaces() {
   return false;
 }
 
+content::DesktopMediaID NativeWindowViews::GetDesktopMediaID() const {
+  const gfx::AcceleratedWidget accelerated_widget = GetAcceleratedWidget();
+  content::DesktopMediaID::Id window_handle = content::DesktopMediaID::kNullId;
+  content::DesktopMediaID::Id aura_id = content::DesktopMediaID::kNullId;
+#if defined(OS_WIN)
+  window_handle =
+      reinterpret_cast<content::DesktopMediaID::Id>(accelerated_widget);
+#elif defined(USE_X11)
+  window_handle = accelerated_widget;
+#endif
+  aura::WindowTreeHost* const host =
+      aura::WindowTreeHost::GetForAcceleratedWidget(accelerated_widget);
+  aura::Window* const aura_window = host ? host->window() : nullptr;
+  if (aura_window) {
+    aura_id = content::DesktopMediaID::RegisterNativeWindow(
+                  content::DesktopMediaID::TYPE_WINDOW, aura_window)
+                  .window_id;
+  }
+
+  // No constructor to pass the aura_id. Make sure to not use the other
+  // constructor that has a third parameter, it is for yet another purpose.
+  content::DesktopMediaID result = content::DesktopMediaID(
+      content::DesktopMediaID::TYPE_WINDOW, window_handle);
+
+  // Confusing but this is how content::DesktopMediaID is designed. The id
+  // property is the window handle whereas the window_id property is an id
+  // given by a map containing all aura instances.
+  result.window_id = aura_id;
+  return result;
+}
+
 gfx::AcceleratedWidget NativeWindowViews::GetAcceleratedWidget() const {
   if (GetNativeWindow() && GetNativeWindow()->GetHost())
     return GetNativeWindow()->GetHost()->GetAcceleratedWidget();
--- a/shell/browser/native_window_views.cc
+++ b/shell/browser/native_window_views.cc
@@ -17,6 +17,7 @@
 #include "base/stl_util.h"
 #include "base/strings/utf_string_conversions.h"
 #include "content/public/browser/browser_thread.h"
+#include "content/public/browser/desktop_media_id.h"
 #include "native_mate/dictionary.h"
 #include "shell/browser/api/atom_api_web_contents.h"
 #include "shell/browser/native_browser_view_views.h"
@@ -645,6 +646,31 @@ void NativeWindowViews::SetResizable(bool resizable) {
     FlipWindowStyle(GetAcceleratedWidget(), resizable, WS_THICKFRAME);
 #endif
   resizable_ = resizable;
+}
+
+bool NativeWindowViews::MoveAbove(const std::string& sourceId) {
+  const content::DesktopMediaID id = content::DesktopMediaID::Parse(sourceId);
+  if (id.type != content::DesktopMediaID::TYPE_WINDOW)
+    return false;
+
+#if defined(OS_WIN)
+  const HWND otherWindow = reinterpret_cast<HWND>(id.id);
+  if (!::IsWindow(otherWindow))
+    return false;
+
+  gfx::Point pos = GetPosition();
+  gfx::Size size = GetSize();
+  ::SetWindowPos(GetAcceleratedWidget(), otherWindow, pos.x(), pos.y(),
+                 size.width(), size.height(),
+                 SWP_NOACTIVATE | SWP_NOMOVE | SWP_NOSIZE | SWP_SHOWWINDOW);
+#elif defined(USE_X11)
+  if (!IsWindowValid(id.id))
+    return false;
+
+  electron::MoveWindowAbove(GetAcceleratedWidget(), id.id);
+#endif
+
+  return true;
 }
 
 void NativeWindowViews::MoveTop() {
--- a/spec-main/api-browser-window-spec.ts
+++ b/spec-main/api-browser-window-spec.ts
@@ -653,6 +653,48 @@ describe('BrowserWindow module', () => {
       })
     })
 
+    describe('BrowserWindow.moveAbove(mediaSourceId)', () => {
+      it('should throw an exception if wrong formatting', async () => {
+        const fakeSourceIds = [
+          'none', 'screen:0', 'window:fake', 'window:1234', 'foobar:1:2'
+        ]
+        fakeSourceIds.forEach((sourceId) => {
+          expect(() => {
+            w.moveAbove(sourceId)
+          }).to.throw(/Invalid media source id/)
+        })
+      })
+      it('should throw an exception if wrong type', async () => {
+        const fakeSourceIds = [null as any, 123 as any]
+        fakeSourceIds.forEach((sourceId) => {
+          expect(() => {
+            w.moveAbove(sourceId)
+          }).to.throw(/Error processing argument at index 0 */)
+        })
+      })
+      it('should throw an exception if invalid window', async () => {
+        // It is very unlikely that these window id exist.
+        const fakeSourceIds = ['window:99999999:0', 'window:123456:1',
+          'window:123456:9']
+        fakeSourceIds.forEach((sourceId) => {
+          expect(() => {
+            w.moveAbove(sourceId)
+          }).to.throw(/Invalid media source id/)
+        })
+      })
+      it('should not throw an exception', async () => {
+        const w2 = new BrowserWindow({ show: false, title: 'window2' })
+        const w2Shown = emittedOnce(w2, 'show')
+        w2.show()
+        await w2Shown
+
+        expect(() => {
+          w.moveAbove(w2.getMediaSourceId())
+        }).to.not.throw()
+
+        await closeWindow(w2, { assertNotWindows: false })
+      })
+    })
   })
 
   describe('sizing', () => {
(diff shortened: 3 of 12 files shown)


CapOM:
CC @MarshallOfSound @alexeykuzmin , Hi I am not sure who to add as reviewer, thx.

shell/browser/native_window_views.cc
```diff
@@ -643,6 +643,26 @@ void NativeWindowViews::SetResizable(bool resizable) {
...
 }
 
+void NativeWindowViews::MoveAbove(const std::string& sourceId) {
+  const content::DesktopMediaID id = content::DesktopMediaID::Parse(sourceId);
+  if (id.type != content::DesktopMediaID::TYPE_WINDOW)
+    return;
+
+#if defined(OS_WIN)
+  const HWND otherWindow = reinterpret_cast<HWND>(id.id);
+  if (!IsWindow(otherWindow))
```
codebytere: I think i'd prefer if this threw a descriptive error of some kind: i think that silently failing is generally not intuitive to users and can lead them to think that our APIs are failing rather than there being an issue on their end.

codebytere: I think i'd prefer if this threw a descriptive error of some kind: i think that silently failing is generally not intuitive to users and can lead them to think that our APIs are failing rather than there being an issue on their end.

CapOM: Done it now throw an error

CapOM:
I have added 5 unit tests

CapOM:
I am fighting with the unit tests (flakiness of API I am using in the tests but not the new API itself).
But the code is ready to be re-reviewed. Please take a look, thx!

spec/api-browser-window-spec.js
```diff
@@ -2497,6 +2497,20 @@ describe('BrowserWindow module', () => {
...
+  describe('window.getMediaSourceId()', () => {
+    it('returns valid source id', () => {
+      // Check format 'window:1234:0'.
+      const sourceId = w.getMediaSourceId()
+      const sourceIdSplit = sourceId.split(':')
+      expect(sourceIdSplit.length).to.equal(3)
+      expect(sourceIdSplit[0]).to.equal('window')
+      const windId = Number.parseInt(sourceIdSplit[1])
+      const contentId = Number.parseInt(sourceIdSplit[2])
+      expect(Number.isInteger(windId)).to.be.true()
```
codebytere:
```
expect(windId).to.be.a('number')
expect(contentId).to.be.a('number')
```

spec-main/api-browser-window-spec.ts
```diff
@@ -638,6 +638,110 @@ describe('BrowserWindow module', () => {
       })
     })
 
+    describe('BrowserWindow.moveAbove(mediaSourceId)', () => {
+      it('should throw an exception if wrong formatting', async () => {
+        const fakeSourceIds = ['none', 'screen:0', 'window:fake', 'window:1234']
+        fakeSourceIds.forEach((sourceId) => {
+          try {
+            w.moveAbove(sourceId)
```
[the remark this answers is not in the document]
CapOM: Done

spec-main/api-browser-window-spec.ts
```diff
@@ -638,6 +638,110 @@ describe('BrowserWindow module', () => {
...
+          } catch (err) {
+            expect(err.message).to.equal('Invalid media source id')
+          }
+        })
+      })
+      it('should throw an exception if invalid window', async () => {
+        // It is very unlikely that these window id exist.
+        const fakeSourceIds = ['window:1234:0', 'window:1234:1', 'window:12:9']
+        fakeSourceIds.forEach((sourceId) => {
+          try {
```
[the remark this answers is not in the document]
CapOM: Done, thx

spec-main/api-browser-window-spec.ts
```diff
@@ -638,6 +638,110 @@ describe('BrowserWindow module', () => {
...
+            expect(false).to.equal(true)
+          } catch (err) {
+            expect(err.message).to.equal('Invalid media source id')
+          }
+        })
+      })
+      it('should not throw an exception', async () => {
+        const w2 = new BrowserWindow({ show: false, title: 'window2' })
+        const w2Shown = emittedOnce(w2, 'show')
+        w2.show()
```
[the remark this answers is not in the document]
CapOM: Done, thx

spec/api-browser-window-spec.js
CapOM: Done, thx

CapOM:
I addressed remaining issues. The test failure seem unrelated to this change. Please take a look, thx!

CapOM:
Hi, gentle ping ?

shell/browser/api/atom_api_top_level_window.cc
```diff
@@ -724,6 +729,11 @@ void TopLevelWindow::RemoveBrowserView(v8::Local<v8::Value> value) {
     }
   }
 }
+
+std::string TopLevelWindow::getMediaSourceId() const {
```
nornagon:
```suggestion
std::string TopLevelWindow::GetMediaSourceId() const {
```

spec-main/api-browser-window-spec.ts
```diff
@@ -638,6 +638,38 @@ describe('BrowserWindow module', () => {
...
+        await w2Shown
+
+        expect(() => {
+          w.moveAbove(w2.getMediaSourceId())
+        }).to.not.throw()
+
+        await closeWindow(w2, { assertNotWindows: false })
```
nornagon: Is there some way you can test that the window was actually moved?

CapOM: Not really, or maybe with DesktopCapturer.getSources() as they are returned from top to bottom but I can see the flakiness coming.

What we need is an option to `BrowserWindow.getAllWindows()` to get all windows from top to bottom in term of z-order.

nornagon: if `DesktopCapturer.getSources()` is guaranteed to return in the correct z-order then that seems like a reasonable way to test.

CapOM: Oki will do.

shell/browser/native_window_views.cc
```diff
@@ -643,6 +644,31 @@ void NativeWindowViews::SetResizable(bool resizable) {
...
 }
 
+bool NativeWindowViews::MoveAbove(const std::string& sourceId) {
+  const content::DesktopMediaID id = content::DesktopMediaID::Parse(sourceId);
+  if (id.type != content::DesktopMediaID::TYPE_WINDOW)
+    return false;
+
+#if defined(OS_WIN)
+  const HWND otherWindow = reinterpret_cast<HWND>(id.id);
+  if (!IsWindow(otherWindow))
```
zcbenz: Add `::`prefix to make it clear that this is a win32 API call. 

```suggestion
  if (!::IsWindow(otherWindow))
```

CapOM: Done, thx!

spec-main/api-browser-window-spec.ts
CapOM: Hi @nornagon , I added a such test. You will that if the requirements are not met the test is skipped.

CapOM:
Hi, I have addressed all the remaining remarks, please take a look, thx!

CapOM:
Hi @nornagon , gentle ping ?

shell/browser/native_window_views.cc
```diff
@@ -1094,6 +1120,37 @@ bool NativeWindowViews::IsVisibleOnAllWorkspaces() {
...
+  }
+
+  // No constructor to pass the aura_id. Make sure to not use the other
+  // constructor that has a third parameter, it is for yet another purpose.
+  content::DesktopMediaID result = content::DesktopMediaID(
+      content::DesktopMediaID::TYPE_WINDOW, window_handle);
+
+  // Confusing but this is how content::DesktopMediaID is designed. The id
+  // property is the window handle whereas the window_id property is an id
+  // give by a map containing all aura instances.
```
nornagon: this seems like it's maybe ChromeOS specific..?

spec-main/api-browser-window-spec.ts
```diff
@@ -638,6 +638,38 @@ describe('BrowserWindow module', () => {
       })
     })
 
+    describe('BrowserWindow.moveAbove(mediaSourceId)', () => {
+      it('should throw an exception if wrong formatting', async () => {
+        const fakeSourceIds = ['none', 'screen:0', 'window:fake', 'window:1234']
```
nornagon:
```suggestion
        const fakeSourceIds = ['none', 'screen:0', 'window:fake', 'window:1234', 'foobar:1:2', null as any, 123 as any]
```

spec/api-browser-window-spec.js
```diff
@@ -2525,6 +2525,20 @@ describe('BrowserWindow module', () => {
...
+  describe('window.getMediaSourceId()', () => {
+    it('returns valid source id', () => {
+      // Check format 'window:1234:0'.
+      const sourceId = w.getMediaSourceId()
+      const sourceIdSplit = sourceId.split(':')
+      expect(sourceIdSplit.length).to.equal(3)
+      expect(sourceIdSplit[0]).to.equal('window')
+      const windId = Number.parseInt(sourceIdSplit[1])
+      const contentId = Number.parseInt(sourceIdSplit[2])
+      expect(windId).to.be.a('number')
```
nornagon: Maybe rather:

```js
expect(sourceId).to.match(/^window:\d+:\d+$/)
```

since `parseInt('1aoeu')` => `1`

spec/api-desktop-capturer-spec.js
```diff
@@ -91,12 +91,145 @@ describe('desktopCapturer', () => {
...
+    const sources = await desktopCapturer.getSources({
+      types: ['window'],
+      thumbnailSize: { width: 0, height: 0 }
+    })
+    w.destroy()
+
+    // TODO(julien.isorce): investigate why |sources| is empty on the linux
+    // bots while it is not on my workstation, as expected, with and without
+    // the --ci parameter.
+    if (process.platform === 'linux' && sources.length === 0) { return }
```
nornagon: generally better to skip the test rather than let it pass.

spec/api-desktop-capturer-spec.js
```diff
@@ -91,12 +91,145 @@ describe('desktopCapturer', () => {
...
+    const MAX_WIN = 4
+    const { BrowserWindow } = remote
+    const mainWindow = remote.getCurrentWindow()
+    const wList = [mainWindow]
+    // Add MAX_WIN-1 more window so we have MAX_WIN in total.
+    for (let i = 0; i < MAX_WIN - 1; i++) {
+      const w = new BrowserWindow({ show: true, width: 100, height: 100 })
+      wList.push(w)
+      w.loadURL('about:blank')
+      w.show()
```
nornagon: is the `show()` necessary since the window is constructed with `show: true`?

(also `loadURL('about:blank')` seems unnecessary)

spec/api-desktop-capturer-spec.js
```diff
@@ -91,12 +91,145 @@ describe('desktopCapturer', () => {
...
+
+    // Only keep our windows, they must be in the MAX_WIN first windows.
+    sources.splice(MAX_WIN, sources.length - MAX_WIN)
+    expect(sources.length).to.equal(MAX_WIN)
+    expect(sources.length).to.equal(wList.length)
+
+    // Check that the sources and wList are sorted in the reverse order.
+    let canGofurther = true
+    const wListReversed = wList.slice(0).reverse()
+    sources.forEach((source, index) => {
```
nornagon:
```js
const canGoFurther = sources.every(
    (source, index) => source.id === wListReversed[index].getMediaSourceId())
```

spec/api-desktop-capturer-spec.js
```diff
@@ -91,12 +91,145 @@ describe('desktopCapturer', () => {
...
+        canGofurther = false
+      }
+    })
+    if (!canGofurther) {
+      // Skip remaining checks because either focus or window placement are
+      // not reliable in the running test environment. So there is no point
+      // to go further to test moveAbove as requirements are not met.
+      wList.forEach((w) => {
+        if (w !== mainWindow) {
+          w.destroy()
```
nornagon: put this cleanup in a try/finally to deduplicate it. e.g.

```js
const wList = ...
try {
  // ...
  if (cond) {
    this.skip()
    return
  }
} finally {
  // ... cleanup wList ...
}
```

spec/api-desktop-capturer-spec.js
```diff
@@ -91,12 +91,145 @@ describe('desktopCapturer', () => {
...
+    let canGofurther = true
+    const wListReversed = wList.slice(0).reverse()
+    sources.forEach((source, index) => {
+      if (source.id !== wListReversed[index].getMediaSourceId()) {
+        canGofurther = false
+      }
+    })
+    if (!canGofurther) {
+      // Skip remaining checks because either focus or window placement are
+      // not reliable in the running test environment. So there is no point
```
nornagon: What would cause this on CI? I think it's OK to fail the test rather than skip it on a user's machine if some window pops up during a test, but that shouldn't happen on CI and I don't think it's worth weakening the test for that.

spec/api-desktop-capturer-spec.js
```diff
@@ -91,12 +91,145 @@ describe('desktopCapturer', () => {
...
+    })
+
+    sources = await desktopCapturer.getSources({
+      types: ['window'],
+      thumbnailSize: { width: 0, height: 0 }
+    })
+    // Only keep our windows again.
+    sources.splice(MAX_WIN, sources.length - MAX_WIN)
+    expect(sources.length).to.equal(MAX_WIN)
+    expect(sources.length).to.equal(wList.length)
```
nornagon: perhaps `sources = sources.filter(source => ourWindows.contains(source.id))` would be more robust? the relative ordering of our windows should be consistent even if other windows manage to get in the way.

shell/browser/native_window_views.cc
CapOM: For example if I call `new BrowserWindow` multiple times within the same electron app, it will be window:A:0, window:B:1, window:C:2 and so on.  (A, B, C being their window handles, so different for each). So this not ChromeOS specific (and I am on Ubu 18.04)

spec-main/api-browser-window-spec.ts
CapOM: Done

spec/api-browser-window-spec.js
CapOM: Done, thx!

spec/api-desktop-capturer-spec.js
CapOM: Done, removed

spec/api-desktop-capturer-spec.js
CapOM: Done, added a skip

spec/api-desktop-capturer-spec.js
CapOM: contains does not exist and includes did not work because of the `.id` so I left it like this

spec/api-desktop-capturer-spec.js
CapOM: Don thx

CapOM:
Hi @nornagon I think I addressed all your remarks, let me know if I forgot something, thx!

spec/api-desktop-capturer-spec.js
CapOM: Done

spec/api-browser-window-spec.js
```diff
@@ -2525,6 +2525,15 @@ describe('BrowserWindow module', () => {
     })
   })
 
+  describe('window.getMediaSourceId()', () => {
+    it('returns valid source id', () => {
+      // Check format 'window:1234:0'.
+      const sourceId = w.getMediaSourceId()
+      const sourceIdSplit = sourceId.split(':')
```
nornagon: no need for `sourceIdSplit` any more

CapOM: Done, thx!

codebytere:
```
BrowserWindow module window.getMediaSourceId() returns valid source id - returns valid source id
```
still seems to be failing @CapOM

spec/api-browser-window-spec.js
CapOM: I think the remaining failure is due to this change. Hi @nornagon any idea ?

codebytere:
@CapOM would you still like to move forward with this?

CapOM:
Back from vacations, I just rebased the 2 commits for now.

codebytere:
@CapOM test failure on macOS:
```
BrowserWindow module window.getMediaSourceId() returns valid source id - returns valid source id

AssertionError: expected 'window:-1:0' to match /^window:\d+:\d+$/
```

CapOM:
Thx, I will have a look.

CapOM:
@codebytere I fixed the test. The -1 was because  [NSWindow windowNumber] returns the max value of uint64_t which is `18446744073709551615` (2^64 - 1) until the window is ordered to the screen for the first time. So the test now wait that the window is shown.

(the reason why it is translated to -1 in the string is because content:DesktopMediaID cast it to intptr_t instead of uintptr_t, so the conversion to string through content:DesktopMediaID.ToString selects the `long long` signature of the converting function. Instead of selecting the one that takes `unsigned long long` as parameter)

codebytere:
@CapOM great detective work 🕵 

i'll get this merged as soon as CI is ✅

CapOM:
CI looks green, thx!

bynho:
Has this been removed?  I cant find the API to get the media source id mentioned in this PR

CapOM:
PRs are there:

https://github.com/electron/electron/pull/18926/commits/9ca6bc037a712aa075be2100682da7df0ce6670e
https://github.com/electron/electron/pull/18926/commits/64131e3d1b2ee3a2f2fd416b90a10cb76823a6fa

And it is shipped in electron starting from 8.0

bynho:
Fantastic. Thank you @CapOM

PR #18928, opened 2019-06-21
fix: sanitize invalid custom protocol headers

DeerMichel:
Backport of #18854.

See that PR for more details.

Notes: none


codebytere:
@DeerMichel the build failures look to be fault of CircleCI: i've re-kicked them and will approve once they all go green.

PR #18934, opened 2019-06-21
feat: add app.getTrackableObjectInstances()

miniak:
#### Description of Change
Returns the number of instances per object type, which can be useful for debugging memory leaks.

Example:
```
{ Menu: 8,
  Protocol: 1,
  Session: 1,
  TopLevelWindow: 1,
  WebContents: 1 }
```

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `app.getTrackableObjectInstances()`, which can be useful for debugging memory leaks.

### DIFF
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -1057,6 +1057,10 @@ machineModelVersion: '11.5' }
 
 Using `basic` should be preferred if only basic information like `vendorId` or `driverId` is needed.
 
+### `app.getTrackableObjectInstances()`
+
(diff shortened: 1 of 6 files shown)


codebytere:
@miniak is this information inaccessible to us through other means? if we can easily access it via heap shapshots, what does this give us that that doesn't already?

miniak:
@MarshallOfSound, @codebytere you're right, we can just use the heap snapshot instead. Closing.

Issue #18516, opened 2019-05-30
Sandbox inside ppapi don't inherit sanbox property from BrowserWindow


Feverqwe:
Hello. I run BrowserWindow with sandbox false and when run ppapi plugin inside, it runing inside sandbox.

### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.2
* **Operating System:**
  * Windows 10 (1903)
* **Last Known Working Electron version:**
  * 4.2.2

### Expected Behavior
ppapi process must inheret --no-sandbox switch

### Actual Behavior
ppapi process don't inheret --no-sandbox switch when BrowserWindow have sandbox: false property.

### To Reproduce
Create window and run any ppapi plugin inside
```
new BrowserWindow({
      ...
      webPreferences: {
         sandbox: false
      }
    });
```

### Additional Information
I can fix it with this switch, but is disable sandbox globally.
```
      electron.app.commandLine.appendSwitch('no-sandbox');
```

miniak:
@Feverqwe Why is that a problem?

Feverqwe:
@miniak cause ppapi plugins don't works.

miniak:
@Feverqwe which plugin?

Feverqwe:
I use mpv.js from here https://github.com/Kagami/mpv.js 
if I don't use no-sandbox swith I see black block.

miniak:
@Feverqwe the proper solution IMO in this case is to add a new API to opt-out of sandboxing for a particular plugin. See PR #18939

miniak:
code example would be
```
const {PLUGIN_MIME_TYPE} = require("mpv.js");
app.disablePluginSandbox(PLUGIN_MIME_TYPE);
```

Feverqwe:
Its looks great! Thx!

miniak:
@Feverqwe the only problem is that this is semver minor and we'll be only able to add this to Electron 6, which is still in beta. For Electron 5, we would have to bump 5.0 to 5.1 in order to add a new API there.

Feverqwe:
[...]

PR #18939, opened 2019-06-22
feat: add app.disablePluginSandbox(mimeType)

miniak:
#### Description of Change
Added `app.disablePluginSandbox(mimeType)` that allows to disable sandboxing of the ppapi plugin host process, which is per-plugin and shared by multiple `BrowserWindow` instances.

Address #18516. Follow up to https://github.com/electron/electron/pull/15894#issuecomment-443781871.

/cc @nornagon 

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `app.disablePluginSandbox(mimeType)`, which disables sandbox for the plugin identified by `mimeType`.


miniak:
@nornagon Should we maybe handle #17823 in `AtomBrowserClient::IsPluginSandboxed()` now?

miniak:
cc @Feverqwe

MarshallOfSound:
As a feature being targetted at `6-0-x` this late in the beta cycle it needs approval from @electron/wg-releases.  Please add this PR to the WG agenda with reasoning for it's addition

miniak:
@MarshallOfSound added to the agenda. Regardless of whether it should be backported or not, how does the PR look to you?

patches/chromium/add_contentbrowserclient_ispluginsandboxdisabled_callback.patch
```diff
@@ -0,0 +1,109 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Milan Burda <milan.burda@gmail.com>
+Date: Sat, 22 Jun 2019 12:23:04 +0200
+Subject: Add ContentBrowserClient::IsPluginSandboxDisabled() callback
+
+Allows the embedder to decide whether the plugin should be sandboxed.
```
nornagon: Please upstream this patch & add a link to the chromium codereview :)

shell/browser/atom_browser_client.cc
```diff
@@ -988,6 +989,17 @@ bool AtomBrowserClient::PreSpawnRenderer(sandbox::TargetPolicy* policy) {
 }
 #endif  // defined(OS_WIN)
 
+bool AtomBrowserClient::IsPluginSandboxDisabled(
+    const content::PepperPluginInfo& info) {
+  const auto& unsandboxed_plugins = Browser::Get()->unsandboxed_plugins();
```
nornagon: Would it be reasonable to implement this as a callback into JS? I'm imagining an API like

```js
app.setPluginSandboxEnabledHandler((mimeType) => {
  return mimeType !== 'foo/bar'
})
```

As written, the API doesn't allow for re-enabling the sandbox after it's disabled, or for choosing which webContents are allowed to launch unsandboxed plugins. I'm not sure that it'd be likely to need such features, just asking whether we should consider them.

patches/chromium/add_contentbrowserclient_ispluginsandboxdisabled_callback.patch
miniak: I have never upstreamed stuff to Chromium so far, I need to try it for the first time. Any tips?

shell/browser/atom_browser_client.cc
miniak: I don't think you can easily map the plugin to the webContents. My thinking was that there are plugins that just don't work with sandbox at all, so we just need a way to opt-out. I don't think it's necessary to do it with more granularity.

miniak: the JS handler on the other hand could get more information from `content::PepperPluginInfo`, including the name, path and MIME types

miniak: @Feverqwe what do you think as the person experiencing the issue that we are trying to solve here?

codebytere:
@miniak do you still want to merge this? it's needing rebase

patches/chromium/add_contentbrowserclient_ispluginsandboxdisabled_callback.patch
nornagon: Oops, I meant to respond to this ages ago. The best place to start is https://chromium.googlesource.com/chromium/src/+/master/docs/contributing.md

tldr: get a chromium checkout, make a branch, and run `git cl upload`.

shell/browser/atom_browser_client.cc
Feverqwe: I think that any api will be fine. setPluginSandboxEnabledHandler is more flexible, but and disablePluginSandbox looks good.

codebytere:
Closing for now given no action and no response to the above but we can re-open should there be a path forward :)

fp07:
Can we re-open this?

PR #18944, opened 2019-06-23
build: remove scripts in the tools dir that are unused

MarshallOfSound:
I'm cleaning up the `tools` and `scripts` folder, looks like most of the stuff in `tools` was from the GYP days or sometimes even earlier, no need to keep it around anymore

Part 1 of N

Notes: no-notes


codebytere:
before we ditch these, do we know _where_ they were all used at one point in time so we can be sure we don't need them anymore?

MarshallOfSound:
@codebytere Yeah I traced them all to be sure, I'll put a quick list below.

* `atom_source_root.js` used in GYP bootstrap for < 1.8
* `dump-version-info.js` replaced by metadumper
* `get-endianess` used in ancient GYP bootstrap
* `pkg-config-wrapper` used in GYP sysroot land
* `rewrite_dirs` used in GYP sysroot land
* `sysroot_ld_paths` more GYP sysroot stuff
* `make_locale_paks` used in GYP building days, GN / chromium build scripts does this for us now
* `run-electron` yeah idk, back in GYP days, it points back to `out/D` which is a GYP thing

Issue #18125, opened 2019-05-02
registerStringProtocol doesn't support custom headers


nornagon:
### Issue Details

* **Electron Version:**
  * 6.0.0-beta.1
* **Operating System:**
  * macOS 10.14.4
* **Last Known Working Electron version:**:
  * None

### Expected Behavior
Setting the `headers` key in the response callback for a `registerStringProtocol` handler should cause headers to be set in the response.

### Actual Behavior
The `headers` data is ignored.

### To Reproduce
https://gist.github.com/555971b5d659617d3bfabed2469bef60

### Additional Information
It looks like we just never set this data. We probably should.

DeerMichel:
The docs also doesnt state that setting custom headers is possible for a string protocol. I wonder if theres a use case for them? The desired behavior of the gist can be achieved by using `mimeType`:

![image](https://user-images.githubusercontent.com/8472721/59637746-6b289400-910b-11e9-87be-c4e7502f6ecd.png)

Anyway, custom headers can be injected [here](https://github.com/electron/electron/blob/4dc38d39e98c5703a62790d1ba6ed4d4fb52fb81/atom/browser/net/url_request_string_job.cc#L95).

nornagon:
String protocols are particularly handy during testing, and there are plenty of other headers than `Content-Type` that could be useful to set. We permit setting arbitrary headers for `registerStreamProtocol`, and it seems like an omission to disallow that for string protocols (since a string protocol is just a simpler version of stream protocols).

In fact, I don't see any particular reason we shouldn't implement string protocols on top of stream protocols, entirely in JS, rather than having separate C++ bindings for them. maybe cc @zcbenz for his thoughts?

zcbenz:
The NetworkService-based protocol APIs have custom headers (and other options) implemented for all APIs, thanks to the new architecture, the new API implementations are both simpler and more powerful.

I'll have a PR for documenting all changes and goodies for NetworkService-based protocol API.

electron-triage:
The Electron version reported on this issue is no longer supported. See our [supported versions documentation](https://www.electronjs.org/docs/tutorial/support#supported-versions).

If this is still reproducible on a supported version, please open a new issue with any other new information that a maintainer should know.

Thank you for taking the time to report this issue and helping to make Electron better! Your help is appreciated.

PR #18952, opened 2019-06-24
docs: documentation of NetworkService-based protocol module

zcbenz:
#### Description of Change

Refs https://github.com/electron/electron/issues/18125, https://github.com/electron/electron/issues/15791.

This PR documents the changes of NetworkService-based protocol module, the contents are very likely to change as the NetworkService migration continues.

Since currently we don't have an estimate of when will we enable NetworkService by default in Electron, the docs are put in separate files and are not available to users. They should be moved to `breaking-changes.md` and `protocol.md` after it is determined when to enable NetworkService in Electron.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/docs/api/breaking-changes-ns.md
+++ b/docs/api/breaking-changes-ns.md
@@ -3,12 +3,12 @@
 This document describes changes to Electron APIs after migrating network code
 to NetworkService API.
 
-Currently we don't have an estimate of when will we enable NetworkService by
-default in Electron, but as Chromium is already removing non-NetworkService
-code, the switch may come after a few major version releases.
+We don't currently have an estimate of when we will enable `NetworkService` by
+default in Electron, but as Chromium is already removing non-`NetworkService`
+code, we might switch before Electron 10.
 
-The content of this document should be moved to `breaking-changes.md` after it
-is determined when to enable NetworkService in Electron.
+The content of this document should be moved to `breaking-changes.md` once we have
+determined when to enable `NetworkService` in Electron.
 
 ## Planned Breaking API Changes
 
--- a/docs/api/breaking-changes-ns.md
+++ b/docs/api/breaking-changes-ns.md
@@ -54,7 +54,7 @@ and `protocol.isProtocolIntercepted` instead.
 
 ```javascript
 // Deprecated
-const isHandled = await protocol.isProtocolHandled(scheme)
+protocol.isProtocolHandled(scheme).then(() => { /* ... */ })
 // Replace with
 const isRegistered = protocol.isProtocolRegistered(scheme)
 const isIntercepted = protocol.isProtocolIntercepted(scheme)
--- a/docs/api/breaking-changes-ns.md
+++ b/docs/api/breaking-changes-ns.md
@@ -1,4 +1,4 @@
-# Breaking changes (NetworkService)
+# Breaking changes (NetworkService) (Draft)
 
 This document describes changes to Electron APIs after migrating network code
 to NetworkService API.
--- a/docs/api/breaking-changes-ns.md
+++ b/docs/api/breaking-changes-ns.md
@@ -1,0 +1,61 @@
+# Breaking changes (NetworkService)
+
+This document describes changes to Electron APIs after migrating network code
+to NetworkService API.
+
+Currently we don't have an estimate of when will we enable NetworkService by
+default in Electron, but as Chromium is already removing non-NetworkService
+code, the switch may come after a few major version releases.
+
+The content of this document should be moved to `breaking-changes.md` after it
+is determined when to enable NetworkService in Electron.
+
+## Planned Breaking API Changes
+
+### `protocol.unregisterProtocol`
+### `protocol.uninterceptProtocol`
+
+The APIs are now synchronous and the optional callback is no longer needed.
+
+```javascript
+// Deprecated
+protocol.unregisterProtocol(scheme, () => { /* ... */ })
+// Replace with
+protocol.unregisterProtocol(scheme)
+```
+
+### `protocol.registerFileProtocol`
+### `protocol.registerBufferProtocol`
+### `protocol.registerStringProtocol`
+### `protocol.registerHttpProtocol`
+### `protocol.registerStreamProtocol`
+### `protocol.interceptFileProtocol`
+### `protocol.interceptStringProtocol`
+### `protocol.interceptBufferProtocol`
+### `protocol.interceptHttpProtocol`
+### `protocol.interceptStreamProtocol`
+
+The APIs are now synchronous and the optional callback is no longer needed.
+
+```javascript
+// Deprecated
+protocol.registerFileProtocol(scheme, handler, () => { /* ... */ })
+// Replace with
+protocol.registerFileProtocol(scheme, handler)
+```
+
+The registered or intercepted protocol does not have effect on current page
+until navigation happens.
+
+### `protocol.isProtocolHandled`
+
+This API is deprecated and users should use `protocol.isProtocolRegistered`
+and `protocol.isProtocolIntercepted` instead.
+
+```javascript
+// Deprecated
+const isHandled = await protocol.isProtocolHandled(scheme)
+// Replace with
+const isRegistered = protocol.isProtocolRegistered(scheme)
+const isIntercepted = protocol.isProtocolIntercepted(scheme)
+```
--- a/docs/api/protocol-ns.md
+++ b/docs/api/protocol-ns.md
@@ -2,12 +2,12 @@
 
 This document describes the new protocol APIs based on NetworkService.
 
-Currently we don't have an estimate of when will we enable NetworkService by
-default in Electron, but as Chromium is already removing non-NetworkService
-code, the switch may come after a few major version releases.
+We don't currently have an estimate of when we will enable `NetworkService` by
+default in Electron, but as Chromium is already removing non-`NetworkService`
+code, we might switch before Electron 10.
 
 The content of this document should be moved to `protocol.md` after it is
-determined when to enable NetworkService in Electron.
+determined when to enable `NetworkService` in Electron.
 
 > Register a custom protocol and intercept existing protocol requests.
 
@@ -89,7 +89,7 @@ A standard scheme adheres to what RFC 3986 calls [generic URI
 syntax](https://tools.ietf.org/html/rfc3986#section-3). For example `http` and
 `https` are standard schemes, while `file` is not.
 
-Registering a scheme as standard, will allow relative and absolute resources to
+Registering a scheme as standard allows relative and absolute resources to
 be resolved correctly when served. Otherwise the scheme will behave like the
 `file` protocol, but without the ability to resolve relative URLs.
 
--- a/docs/api/protocol-ns.md
+++ b/docs/api/protocol-ns.md
@@ -126,7 +126,7 @@ an incoming request for the `scheme`.
 
 To handle the `request`, the `callback` should be called with either the file's
 path or an object that has a `path` property, e.g. `callback(filePath)` or
-`callback({ path: filePath })`.
+`callback({ path: filePath })`. The `filePath` must be an absolute path.
 
 By default the `scheme` is treated like `http:`, which is parsed differently
 from protocols that follow the "generic URI syntax" like `file:`.
--- a/docs/api/protocol-ns.md
+++ b/docs/api/protocol-ns.md
@@ -1,0 +1,380 @@
+# protocol (NetworkService)
+
+This document describes the new protocol APIs based on NetworkService.
+
+Currently we don't have an estimate of when will we enable NetworkService by
+default in Electron, but as Chromium is already removing non-NetworkService
+code, the switch may come after a few major version releases.
+
+The content of this document should be moved to `protocol.md` after it is
+determined when to enable NetworkService in Electron.
+
+> Register a custom protocol and intercept existing protocol requests.
+
+Process: [Main](../glossary.md#main-process)
+
+An example of implementing a protocol that has the same effect as the
+`file://` protocol:
+
+```javascript
+const { app, protocol } = require('electron')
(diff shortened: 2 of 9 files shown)


docs/api/breaking-changes-ns.md
```diff
@@ -0,0 +1,61 @@
+# Breaking changes (NetworkService)
+
+This document describes changes to Electron APIs after migrating network code
+to NetworkService API.
+
+Currently we don't have an estimate of when will we enable NetworkService by
```
felixrieseberg:
```suggestion
We don't currently have an estimate of when we will enable `NetworkService` by
```

docs/api/breaking-changes-ns.md
```diff
@@ -0,0 +1,61 @@
...
+default in Electron, but as Chromium is already removing non-NetworkService
```
felixrieseberg:
```suggestion
default in Electron, but as Chromium is already removing non-`NetworkService`
```

docs/api/breaking-changes-ns.md
```diff
@@ -0,0 +1,61 @@
...
+code, the switch may come after a few major version releases.
```
felixrieseberg:
```suggestion
code, we might switch before Electron 10.
```

☝️ I'm not sure if that's true, but it's probably best to give an estimate that's unrelated to what the _current_ version is.

docs/api/breaking-changes-ns.md
```diff
@@ -0,0 +1,61 @@
...
+
+The content of this document should be moved to `breaking-changes.md` after it
```
felixrieseberg:
```suggestion
The content of this document should be moved to `breaking-changes.md` once we have
```

docs/api/breaking-changes-ns.md
```diff
@@ -0,0 +1,61 @@
...
+is determined when to enable NetworkService in Electron.
```
felixrieseberg:
```suggestion
determined when to enable `NetworkService` in Electron.
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
+# protocolNS
```
felixrieseberg:
```suggestion
# `NetworkService` Protocol
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+
+This document describes the new protocol APIs based on NetworkService.
+
+Currently we don't have an estimate of when will we enable NetworkService by
```
felixrieseberg:
```suggestion
We don't currently have an estimate of when we will enable `NetworkService` by
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+default in Electron, but as Chromium is already removing non-NetworkService
```
felixrieseberg:
```suggestion
default in Electron, but as Chromium is already removing non-`NetworkService`
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+code, the switch may come after a few major version releases.
```
felixrieseberg:
```suggestion
code, we might switch before Electron 10.
```

See above: I'm not sure if that's the right number, but we should use something we don't have to update.

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+
+The content of this document should be moved to `protocol.md` after it is
+determined when to enable NetworkService in Electron.
```
felixrieseberg:
```suggestion
determined when to enable `NetworkService` in Electron.
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+protocol.registerSchemesAsPrivileged([
+  { scheme: 'foo', privileges: { bypassCSP: true } }
+])
+```
+
+A standard scheme adheres to what RFC 3986 calls [generic URI
+syntax](https://tools.ietf.org/html/rfc3986#section-3). For example `http` and
+`https` are standard schemes, while `file` is not.
+
+Registering a scheme as standard, will allow relative and absolute resources to
```
felixrieseberg:
```suggestion
Registering a scheme as standard allows relative and absolute resources to
```

This is incorrect in the [original](https://github.com/electron/electron/blob/v6.0.0-beta.10/docs/api/protocol.md), too, but I thought I'd fix it here nevertheless.

docs/api/breaking-changes-ns.md
zcbenz: 10 looks good to me.

docs/api/protocol-ns.md
zcbenz: The documentation linter and generator use this title as object name, spaces and special characters can not be used here.

docs/api/structures/protocol-response.md
```diff
@@ -0,0 +1,36 @@
...
+  other types of responses.
+* `path` String (optional) - Path to the file which would be sent as response
+  body. This is only used for file responses.
+* `url` String (optional) - Download the `url` and pipe the result as response
+  body. This is only used for URL responses.
+* `referrer` String (optional) - The `referrer` URL. This is only used for file
+  and URL responses.
+* `method` String (optional) - The HTTP `method`. This is only used for file
+  and URL responses.
+* `session` Object (optional) - The session used for requesting URL, by default
```
MarshallOfSound:
```suggestion
* `session` Session (optional) - The session used for requesting URL, by default
```

docs/api/structures/protocol-response.md
```diff
@@ -0,0 +1,36 @@
...
+  `error` number . For the available error numbers you can use, please see the
+  [net error list][net-error].
+* `statusCode` Number (optional) - The HTTP response code, default is 200.
+* `charset` String (optional) - The charset of response body, default is
+  `"utf-8"`.
+* `mimeType` String (optional) - The MIME type of response body, default is
+  `"text/html"`. Setting `mimeType` would implicitly set the `content-type`
+  header in response, but if `content-type` is already set in `headers`, the
+  `mimeType` would be ignored.
+* `headers` Object (optional) - An object containing the response headers. The
```
MarshallOfSound:
```suggestion
* `headers` Record<string, string | string[]> (optional) - An object containing the response headers. The
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+
+> Register a custom protocol and intercept existing protocol requests.
+
+Process: [Main](../glossary.md#main-process)
+
+An example of implementing a protocol that has the same effect as the
+`file://` protocol:
+
+```javascript
+const { app, protocol } = require('electron')
```
MarshallOfSound: Is it actually exposed on `electron.protocol`?  If so the documentation generator might need some modifications because currently it is inferring the include path as `electron.protocolNS`

deepak1556: Yup it will be still `electron.protocol` just creates a different class internally based on availability of network service.

docs/api/breaking-changes-ns.md
```diff
@@ -0,0 +1,61 @@
...
+### `protocol.unregisterProtocol`
+### `protocol.uninterceptProtocol`
+
+The APIs are now synchronous and the optional callback is no longer needed.
+
+```javascript
+// Deprecated
+protocol.unregisterProtocol(scheme, () => { /* ... */ })
+// Replace with
+protocol.unregisterProtocol(scheme)
```
deepak1556: Do we want to return the error values for synchronous version when the optional callback support is removed, without that the api might look blind.

docs/api/protocol-ns.md
zcbenz: How about adding a way in documentation generator to ignore the module? For example ignoring titles like `# protocol (Draft)`.

docs/api/breaking-changes-ns.md
zcbenz: Yeah there should be a way to report errors. I wonder if we should throw exceptions for errors, to match the behavior of Node.js sync APIs.

deepak1556: Agreed, that seems to be the pattern thats followed in existing synchronous apis

zcbenz: I'll do it in another PR.

docs/api/protocol-ns.md
zcbenz: @MarshallOfSound I have updated the docs parser to ignore the NetworkService draft docs.

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
+# protocol (NetworkService) (Draft)
+
+This document describes the new protocol APIs based on NetworkService.
+
+We don't currently have an estimate of when we will enable `NetworkService` by
```
MarshallOfSound:
```suggestion
We don't currently have an estimate of when we will enable the `NetworkService` by
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
+# protocol (NetworkService) (Draft)
+
+This document describes the new protocol APIs based on NetworkService.
```
MarshallOfSound:
```suggestion
This document describes the new protocol APIs based on the [NetworkService](https://www.chromium.org/servicification).
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+
+We don't currently have an estimate of when we will enable `NetworkService` by
+default in Electron, but as Chromium is already removing non-`NetworkService`
+code, we might switch before Electron 10.
```
MarshallOfSound:
```suggestion
code, we will probably switch before Electron 10.
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+
+The content of this document should be moved to `protocol.md` after it is
```
MarshallOfSound:
```suggestion
The content of this document should be moved to `protocol.md` after we have
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+determined when to enable `NetworkService` in Electron.
```
MarshallOfSound:
```suggestion
enabled the `NetworkService` by default in Electron.
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+* `customSchemes` [CustomScheme[]](structures/custom-scheme.md)
+
+**Note:** This method can only be used before the `ready` event of the `app`
+module gets emitted and can be called only once.
+
+Registers the `scheme` as standard, secure, bypasses content security policy for
+resources, allows registering ServiceWorker and supports fetch API. Specify a
+privilege with the value of `true` to enable the capability.
+
+An example of registering a privileged scheme, with bypassing Content Security
```
MarshallOfSound:
```suggestion
An example of registering a privileged scheme, that bypasses Content Security
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+
+### `protocol.registerFileProtocol(scheme, handler)`
+
+* `scheme` String
+* `handler` Function
+  * `request` ProtocolRequest
+  * `callback` Function
+    * `response` (String | [ProtocolResponse](structures/protocol-response.md))
+
+Registers a protocol of `scheme` that will send the file as a response. The
```
MarshallOfSound:
```suggestion
Registers a protocol of `scheme` that will send a file as the response. The
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+`handler` will be called with `handler(request, callback)` when a `request` is
```
MarshallOfSound:
```suggestion
`handler` will be called with `request` and `callback` where `request` is
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+going to be created with `scheme`.
```
MarshallOfSound:
```suggestion
an incoming request for the `scheme`.
```

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+
+To handle the `request`, the `callback` should be called with either the file's
```
MarshallOfSound: Does this file path have to be absolute? If so we should document that

docs/api/protocol-ns.md
```diff
@@ -0,0 +1,309 @@
...
+* `scheme` String
+* `handler` Function
+  * `request` ProtocolRequest
+  * `callback` Function
+    * `response` (ReadableStream | [ProtocolResponse](structures/protocol-response.md))
+
+Registers a protocol of `scheme` that will send a stream as a response.
+
+The usage is the same with `registerFileProtocol`, except that the
+`callback` should be called with either a `Readable` object or an object that
```
MarshallOfSound:
```suggestion
`callback` should be called with either a [`ReadableStream`](https://nodejs.org/api/stream.html#stream_class_stream_readable) object or an object that
```

docs/api/structures/protocol-request.md
```diff
@@ -0,0 +1,6 @@
+# ProtocolRequest Object
+
+* `url` String
+* `referrer` String
+* `method` String
+* `uploadData` [UploadData[]](upload-data.md)
```
MarshallOfSound: The request won't always have an `uploadData` property right? Or will it be an empty array when it isn't a POST request?

```suggestion
* `uploadData` [UploadData[]](upload-data.md) (optional)
```

docs/api/protocol-ns.md
zcbenz: It must be absolute, to match old API's behavior. I have added document for it.

Issue #18918, opened 2019-06-21
Improve `app.showAboutPanel()` on Linux


sindresorhus:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description

It's nice that `app.showAboutPanel()` works on Linux now, but how it works could be improved.

#### 1. Add defaults

> On Linux, values must be set in order to be shown; there are no defaults.

This just makes it hard to use, since the macOS version has defaults. I don't see why you couldn't have the defaults and save people time. Electron already has the information:

- `applicationName`: Should default to `productName` in package.json.
- `applicationVersion`: Should default to `version` in package.json.
- `copyright`: Should default to `Copyright © 2019 Author Name`, where `Author Name` is the `author.name` field in package.json, if it exists.
- `website`: Should default to the `homepage` field in package.json, if it exists.
- `iconPath`: Should default to the app's icon.

#### 2. Add `credits` support for Linux

The setting is currently only available on macOS.

https://developer.gnome.org/gtk3/stable/GtkAboutDialog.html#gtk-about-dialog-set-authors
https://developer.gnome.org/gtk3/stable/GtkAboutDialog.html#gtk-about-dialog-add-credit-section

#### 3. Resize the icon to the correct size

The `iconPath` option should accept an image of any resize and resize it to the correct size.

This is what I'm currently seeing:

<img width="1218" alt="Screen Shot 2019-06-21 at 17 13 02" src="https://user-images.githubusercontent.com/170270/59917531-6554ea80-944c-11e9-8a91-32e1c368062d.png">

### Alternatives Considered

Not using the native dialog and make a custom one with HTML and CSS, but I would strongly prefer my apps to look and act as native as possible.

### Additional Information

Electron 5.0.5

---

// @codebytere

codebytere:
I'll see what i can do! For context, the reason I didn't initially add the credits field was primarily because of parameter incompatibilities: our logic for the panel is all native and the Linux option for credits would require either an array or an object and I wasn't sure that the added argument sanitization overhead would be justified.

mahnunchik:
Is there any progress?

PR #18964, opened 2019-06-24
feat: show optional authors in gtk about panel

codebytere:
#### Description of Change

Refs #2 in https://github.com/electron/electron/issues/18918.

This PR adds the ability to set a list of authors in the native Linux about panel.

I chose not to add this to the existing credits options owing to incompatible parameter types. We could potentially do it, but i felt that the added argument sanitization outweighed the benefit of being able to use a single argument across both platforms for this.

cc @ckerr @sindresorhus 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added the ability to set authors in the Linux implementation of the About panel.

### DIFF
--- a/shell/browser/browser_linux.cc
+++ b/shell/browser/browser_linux.cc
@@ -147,30 +147,27 @@ bool Browser::IsEmojiPanelSupported() {
 }
 
 void Browser::ShowAboutPanel() {
-  std::string app_name, version, copyright, icon_path, website;
-  base::ListValue* args;
-
   GtkAboutDialog* dialog = GTK_ABOUT_DIALOG(gtk_about_dialog_new());
 
-  if (about_panel_options_.GetString("applicationName", &app_name))
-    gtk_about_dialog_set_program_name(dialog, app_name.c_str());
-  if (about_panel_options_.GetString("applicationVersion", &version))
-    gtk_about_dialog_set_version(dialog, version.c_str());
-  if (about_panel_options_.GetString("copyright", &copyright))
-    gtk_about_dialog_set_copyright(dialog, copyright.c_str());
-  if (about_panel_options_.GetString("website", &website))
-    gtk_about_dialog_set_website(dialog, website.c_str());
-  if (about_panel_options_.GetList("args", &args)) {
-    std::vector<const char*> cstrs;
-    for (const auto& str : args->GetList()) {
-      cstrs.push_back(str.c_str());
-    }
-    cstrs.push_back(nullptr);
-    gtk_about_dialog_set_authors(dialog, cstrs.data());
+  const auto& opts = about_panel_options_;
+  const std::string* str;
+  const base::Value* val;
+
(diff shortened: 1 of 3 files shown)


shell/browser/browser_linux.cc
```diff
@@ -159,6 +160,13 @@ void Browser::ShowAboutPanel() {
     gtk_about_dialog_set_copyright(dialog, copyright.c_str());
   if (about_panel_options_.GetString("website", &website))
     gtk_about_dialog_set_website(dialog, website.c_str());
+  if (about_panel_options.Get("args", &args)) {
+    std::vector<const char*> cstrs{};
+    for (const auto& str : args) {
+      cstrs.push_back(str.c_str());
+    }
```
ckerr: This is literally just a sequence of pointers, C doesn't know the length. The idiom GTK+ uses here is for the sequence to be NULL-terminated:

```suggestion
    }
    cstrs.push_back(nullptr);
```

Otherwise the code as-is will just keep walking past the end of cstrs.data() and treat whatever random memory is there as char pointers

shell/browser/browser_linux.cc
```diff
@@ -159,6 +160,13 @@ void Browser::ShowAboutPanel() {
     gtk_about_dialog_set_copyright(dialog, copyright.c_str());
   if (about_panel_options_.GetString("website", &website))
     gtk_about_dialog_set_website(dialog, website.c_str());
+  if (about_panel_options.Get("args", &args)) {
+    std::vector<const char*> cstrs{};
```
ckerr: The empty list initializer doesn't buy us anything here.

```suggestion
    std::vector<const char*> cstrs;
```

ckerr:
I just pushed a chonky change to this, so @codebytere a quick explanation:

1. First problem is that `Browser::SetAboutPanelOptions` only accepts key/value string pairs. We _could_ add more sophisticated logic to check for string lists as well, but... really this is reinventing the wheel. base::Value already allows all kinds of types to be embedded and has a `Clone()` function, so why not use that.

2. Running alongside that, you can't call `DictionaryValue dv = value.Clone();` because DictionaryValue is a subclass of Value. There are lots of one-liner optios for this, but since DictionaryValue ["exists only as a compatibility shim that we're in the process of removing"](https://cs.chromium.org/chromium/src/base/values.h?q=dictionaryvalue&sq=package:chromium&l=72), we're better off just making `about_panel_options_` a `base::Value`.

3. Most of the rest is just updating the API usage to `base::Value`, e.g. using `FindStringKey()` and `FindListKey()` instead of `GetString()` and `GetList()`. **However** I've also added some extra safeguards to confirm that the list items are strings and to warn if none are found.

ckerr:
4. `about_panel_options_` is still a DictionaryValue on the mac build because on that platform we're massaging the keys to always have an uppercase first character and updating that code seemed out-of-scope for this PR. Let's get that in a followup.

PR #18969, opened 2019-06-25
fix: make tray not block main process

codebytere:
Backport of https://github.com/electron/electron/pull/18880.

See that PR for more details.

Notes: Fixed an issue where the Tray icon context menu on macOS blocked the main process while open.


atom/browser/ui/tray_icon_cocoa.mm
```diff
@@ -329,14 +333,21 @@ - (void)mouseUp:(NSEvent*)event {
   [self setNeedsDisplay:YES];
 }
 
+<<<<<<< HEAD:atom/browser/ui/tray_icon_cocoa.mm
```
ckerr: Unintended?

codebytere:
@ckerr now includes update from https://github.com/electron/electron/pull/18977.

Issue #17890, opened 2019-04-20
App freeze when iframe is deleted from a webview


pwnsdx:
### Preflight Checklist

* [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [X] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.0-beta.9
* **Operating System:**
  * macOS Mojave
* **Last Known Working Electron version:**:
  * 4.1.4

### Expected Behavior

The iframe is removed without causing any issue

### Actual Behavior

The webview content will freeze (the gif will stop working) once the iframe is removed from it and in some cases, you will be unable to close the app.

### To Reproduce

Run the following Electron Fiddle: https://gist.github.com/ae270063bc98401e6f6cbfeb1fddfa5d

It will remove the iframe after 6 seconds. The gif should then freeze on the latest beta (9) and in some cases, closing the app is impossible unless force exiting.

### Additional Information

After further testing, I noticed that if the iframe is restored then the app start acting normally again.

r0stig:
Able to reproduce in 5.0.1

codebytere:
@r0stig @pwnsdx yeah, i'd imagine there's some issue where the OOPIF is not properly reconnected when the renderer process restarts. I'll try to poke at this soon.

sofianguy:
fixed https://github.com/electron/electron/pull/18976

Issue #14211, opened 2018-08-20
OOPIF webview freezes occasionally on navigation


KochiyaOcean:
* Electron Version: v3.0.0-nightly.20180818
* Operating System (Platform and Version): Windows 10 1803
* Last known working Electron version: v3.0.0-beta.5

**Expected Behavior**
webview will not be freezed

**Actual behavior**
webview freezed occasionlly on navigation

**To Reproduce**
1. clone https://github.com/KochiyaOcean/electron-bugs

2. run `electron .` using electron v3.0.0-nightly.20180818

3. run scripts below in devtools
```javascript
require('electron').remote.webContents.getAllWebContents()
```
![image](https://user-images.githubusercontent.com/8194131/44320386-bf77ce80-a47c-11e8-9371-4e89cec444db.png)
there are 3 webcontents (BrowserWindow, WebView, DevTool)

4. run scripts below in devtools
```javascript
document.querySelector('webview').addEventListener('did-finish-load', () => document.querySelector('webview').loadURL('http://dmm.com'))
document.querySelector('webview').loadURL('http://dmm.com')
```

5. webview will freeze occasionally, and call any method of webview will throw an error
![image](https://user-images.githubusercontent.com/8194131/44320462-0b2a7800-a47d-11e8-9b21-3033df0372c4.png)

6. run `require('electron').remote.webContents.getAllWebContents()` again, and there are only 2 webcontents left (webcontents of webview is removed)
![image](https://user-images.githubusercontent.com/8194131/44320488-3c0aad00-a47d-11e8-99db-634428b1198c.png)

**Additional Information**
This bug occurs easier on complex websites, such as https://amazon.com , https://ebay.com , etc.

sofianguy:
We released `3.0.0-beta.6` today. Can you give [`3.0.0-beta.6`](https://github.com/electron/electron/releases/tag/v3.0.0-beta.6) a try and let us know if you are still running into this issue?

KochiyaOcean:
@sofianguy It still occurs in electron@3.0.0-beta.6
![image](https://user-images.githubusercontent.com/8194131/44376067-bdbc1280-a531-11e8-9c3c-01cd68d99bc4.png)

ckerr:
Judging from that trace, this is possibly a regression caused by the 3.0.0-beta.6 asar changes in https://github.com/electron/electron/pull/14031 https://github.com/electron/electron/pull/14118

[...]

PR #18976, opened 2019-06-25
fix: webview crash on iframe removal

### LINKED ISSUE #18788: Remove iframe node from <webview> will cause webview crash.
### Preflight Checklist

### Issue Details

* **Electron Version:**
  *  v5.0.2
* **Operating System:**
  * macOS 10.13.6
  * Haven't tried on other electron versions.

### Expected Behavior
I have an app which runs in electron `<webview>`. I need to render multiple iframes inside this `<webview>` and dynamic add or remove them. 

### Actual Behavior
When I remove one `<iframe>` from the dom tree, the whole `<webview>` crashes. I didn't use JS to remove iframe node, instead, I just delete one `<iframe>` from the dom inspector, and the `<webview>` crashes.

### To Reproduce
This is an example : https://github.com/cloudsere/electron-quick-start. You can run commands below to reproduce.

```sh
$ git clone https://github.com/cloudsere/electron-quick-start
$ npm install
$ npm start
```

alexstrat:
Fixes #18788 #17890

#### Description of Change
Backports this [Chromium fix](https://chromium.googlesource.com/chromium/src/+/7464a6b4e3cf91dd1eb96d922736f0815b94ed9c%5E%21/#F3) as a patch in 5-0-0 branch.

~As a consequence, remove the previous and temporary patch introduced in https://github.com/electron/libchromiumcontent/pull/676~

cc @codebytere @zcbenz @deepak1556 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fix webview crash on iframe removal


patches/common/chromium/portals_fix_crash_when_detaching_oopif_from_portal.patch
```diff
@@ -0,0 +1,183 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Alexandre=20Lach=C3=A8ze?= <alexandre.lacheze@gmail.com>
+Date: Tue, 25 Jun 2019 17:36:08 +0200
+Subject: Portals: Fix crash when detaching OOPIF from portal.
```
alexstrat: I cherry-pick the Chromium commit as it. Not sure it is the expected formatting in the end.

codebytere:
@alexstrat were you able to determine what specifically _about_ this patch breaks the functionality? Ideally i'd love to pare down the patch as much as possible to just the components that are necessary to revert 🤔

alexstrat:
@codebytere my interpretation is that:

Previously, Chromium was unable to detach a subframe of a _(inner) webcontents_ inside a _(outter) webcontens_ (for instance, in our case, a _webview_ in a _BrowserWindow_). That's basically the statement I read in [this comment](https://github.com/electron/electron/issues/14211#issuecomment-418616406) and that led @zcbenz to this [patch](https://github.com/electron/libchromiumcontent/pull/676) (we just don't detach).

The [Chromium patch](https://chromium.googlesource.com/chromium/src/+/7464a6b4e3cf91dd1eb96d922736f0815b94ed9c%5E%21/#F3) seems to fix **exactly** this case: 
> Prior to this change, any detachment of a proxy would cause the inner WebContents to get detached from the outer WebContents.

That's why I would argue to keep the whole patch as it.

That being said, my current knowledge of Chromium architecture did not allow me to:
- understand/find the root cause of the issue, so I'm only an interpretation for now
- understand correctly how the patch in Chromium actually fixes the issue

Therefore, I think the interpretation and opinion of @zcbenz is necessary here.

deepak1556:
Should also try to revert the manual management https://github.com/electron/electron/pull/14487 and see if the issue persists on reload.

alexstrat:
> Should also try to revert the manual management #14487 and see if the issue persists on reload. The new CL seems to address the concerns raised by Cheng's PR, but would be great to get in some tests.

I should rather do that on master, right?

> @alexstrat can you add in tests for the iframe issues that are fixed by the new patch, seems like there are repros that could be turned into a test. Thanks!

Yes, I’ll try that. Though the difficulty is that the webview does not really crash, it just stops to render. So, I still don’t know how to catch that in a test. I’d appreciate ideas!

deepak1556:
> Though the difficulty is that the webview does not really crash, it just stops to render. So, I still don’t know how to catch that in a test. I’d appreciate ideas!

Does that trigger `webContents` `unresponsive` event ? we could listen for that.

zcbenz:
Can you double check this wouldn't cause regression on https://github.com/electron/electron/issues/14211?

alexstrat:
Regarding testing, neither the `'unresponsive'` nor `'crashed'` events are fired when the webview freezes. I tried to hook into the `'paint'` event as well but it does not seem I'm able to catch the paint events for the `webview` and the paint event for the outer webcontents (BrowserWindow) keeps flowing after iframe removal. 😞

alexstrat:
@zcbenz I tried to reproduce #14211 with the instructions provided in description (I waited for ~20 reloads) and could not reproduce it => this PR does cause a regression on #14211

zcbenz:
> I can suggest a PR that reverts #14211 but I think we should 1/ first backport this CL like suggested here so that `5-0-0` matches `6-0-0` and `master`, and 2/ revert #14211 on `master` and backport the revert in `6-0-0` and `5-0-0`.

The plan sounds good to me.

deepak1556:
Can you revert the deletion of `disable_detach_webview_frame.patch` otherwise it doesn't match 6-0-x and master. And as you suggested, a separate PR targeting master can remove this patch along with the manual management, and probably add some tests.

alexstrat:
> Can you revert the deletion of disable_detach_webview_frame.patch otherwise, it doesn't match 6-0-x and master. 

@deepak1556 good catch, yes!

✅done:
- I made sure the CL is the first patch executed
- I took `isable_detach_webview_frame.patch` as it is [just after bumping chromium to 76.0.3783.1]( https://github.com/electron/electron/pull/18116/files#diff-d9e3362ac2c2c5596c5af5895341578e )

subbiah95:
@deepak1556 , I have a similar crash going on in V4.0.0, I am using webview, the crash happened twice,  3 and 5 seconds after start 
the crash points to..
/content/browser/frame_host/render_frame_host_manager.cc
GetProxyToOuterDelegate()->SetChildRWHView(child_rwhv, nullptr);

Is my crash same as this one ?
I am not attaching the dump file as I have changed few codes in electron module(nothing to do with crash) 
electron_crash

deepak1556:
@subbiah95  Can't say for sure without the dump, but I would suggest to try newer versions of electron where the crash might be fixed, V4 is quite old and unsupported release line.

Issue #18123, opened 2019-05-02
'silent: true' for printing not working in version 5.


PWC-Michael:
Hey there

I have upgraded an app to version 5 as in the list of fixes was the silent printing bug, but the issue is still the same.

My code:
`printWindow.webContents.print({silent: true});`

Results in nothing happening at all. No errors thrown or anything. Changing it back to false, brings up the window to select the printer and then it prints perfectly fine. It's just the silent mode that doesn't work.

I am running this on Windows 10, with Node version 10.15.3. I can not try with a different OS as the printer only has drivers for Windows.

Many thanks,
Michael

Naowut:
Same problem on Ubuntu with Electron 5.0.1
I would really like to see this issue fixed.

flaticols:
Same on macOS Mojave

gclooi:
Hi, I discover something interesting with this bug. I am running electron with window 10.

printWindow.webContents.print({ silent: false, printBackground: false}); -> Print Dialog Appeared

printWindow.webContents.print({ silent: false, printBackground: false, deviceName: 'PRINTER_NAME'}); -> Print Dialog NOT Appear

printWindow.webContents.print({ silent: true, printBackground: false}); ->Print Dialog NOT Appear & No Print Activity

printWindow.webContents.print({ silent: true, printBackground: false, deviceName: 'PRINTER_NAME'}); -> Print Dialog NOT Appear & No Print Activity

@sofianguy, is this issue possible relate to the "deviceName" that is not correct? 

Thanks for help and hope this issue can be fix soon.

StijnVanBergen:
I struggled with this issue for hours before realising it might be Electron, not me. I've noticed the same behaviour. Downgraded to 4.*, still an issue. Upgraded to 6 beta, still an issue. After some more testing I've notices the same as @gclooi , I can only get any print activity when !silent and deviceName is not specified. (Tried on Mojave and Win10)

hiteshambaliya:
I have also tested yesterday, exactly as @gclooi  written test cases, and same result.
My system Linux Debian (buster).
I was using electron 4.0.1 then upgraded to 4.2.2 but no result then upgraded to 5.0.0 & 5.0.1 but no success found.
Is there any way to do it.

StvBB8:
@hiteshambaliya The only way I've found it to work for now is downgrading to Electron 3.

hiteshambaliya:
[...]

PR #18979, opened 2019-06-25
fix: silent printing mode

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/18123.

Previously, printing settings weren't set if printing was silent, and so it would appear to do nothing because subsequent action relied on those being set properly.

cc @brenca 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed silent printing across platforms.


SubhamBuie:
image
on windows 10 ,silent printing is not working 
while "electron": "^26.1.0"

PR #18981, opened 2019-06-25
feat: migrate custom macOS tray view to native one

### LINKED ISSUE #18664: Replace StatusItemView implementation with [NSStatusItem button]

DeerMichel:
#### Description of Change
Under macOS Catalina the current custom Tray implementation stops working since we're using deprecated methods (see #18664). This PR uses native methods of `[NSStatusItem button]` where applicable. Closes #18664.

#### Discussion
I got all previously available methods & events working except for `tray.setHighlightMode(mode)`. Imho we should drop support for this mac-only method because we'll have a hard time finding ways to sneak in Apple's native menu handling. However, we are still able to control highlighting when no menu is set (i guess that's where most use cases for this method are found anyway). Probably should be further discussed in a Releases WG meeting. If someone knows a way around that all, feel free to dm me. --> [Verdict](https://github.com/electron/governance/blob/master/wg-releases/meeting-notes/2019-06-26.md#agenda): Deprecate in 6, remove in 7.

cc @codebytere @MarshallOfSound  @miniak @erickzhao 

#### Checklist
- [x] `npm test` passes

#### Release Notes

Notes:
* Migrated `Tray` from custom to native view to be compatible with Catalina (macOS).
* Removed deprecated `Tray.setHighlightMode` method (macOS).

### DIFF
--- a/shell/browser/ui/tray_icon_cocoa.mm
+++ b/shell/browser/ui/tray_icon_cocoa.mm
@@ -7,7 +7,12 @@
 #include <string>
 #include <vector>
 
+#include "base/message_loop/message_loop.h"
+#include "base/message_loop/message_pump_mac.h"
 #include "base/strings/sys_string_conversions.h"
+#include "base/task/post_task.h"
+#include "content/public/browser/browser_task_traits.h"
+#include "content/public/browser/browser_thread.h"
 #include "shell/browser/ui/cocoa/NSString+ANSI.h"
 #include "shell/browser/ui/cocoa/atom_menu_controller.h"
 #include "ui/events/cocoa/cocoa_event_utils.h"
--- a/shell/browser/ui/tray_icon_cocoa.mm
+++ b/shell/browser/ui/tray_icon_cocoa.mm
@@ -106,7 +106,7 @@
 - (void)setTitle:(NSString*)title {
   if ([title containsANSICodes]) {
     [[statusItem_ button]
-        setAttributedTitle:[[title copy] attributedStringParsingANSICodes]];
+        setAttributedTitle:[title attributedStringParsingANSICodes]];
   } else {
     [[statusItem_ button] setTitle:[title copy]];
   }
@@ -268,12 +268,11 @@ TrayIconCocoa::~TrayIconCocoa() {
 }
 
 void TrayIconCocoa::SetImage(const gfx::Image& image) {
-  [status_item_view_ setImage:image.IsEmpty() ? nil : image.AsNSImage()];
+  [status_item_view_ setImage:image.AsNSImage()];
 }
 
 void TrayIconCocoa::SetPressedImage(const gfx::Image& image) {
-  [status_item_view_
-      setAlternateImage:image.IsEmpty() ? nil : image.AsNSImage()];
+  [status_item_view_ setAlternateImage:image.AsNSImage()];
 }
 
 void TrayIconCocoa::SetToolTip(const std::string& tool_tip) {
--- a/shell/browser/ui/tray_icon_cocoa.mm
+++ b/shell/browser/ui/tray_icon_cocoa.mm
@@ -171,9 +171,6 @@
   // Make sure events can be pumped while the menu is up.
   base::MessageLoopCurrent::ScopedNestableTaskAllower allow;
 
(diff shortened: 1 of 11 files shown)


DeerMichel:
Confirmed to work under macOS Catalina Beta 🎉

DeerMichel:
Should be merged **after** #19202

shell/browser/ui/tray_icon_cocoa.mm
```diff
@@ -237,83 +103,44 @@ - (BOOL)getIgnoreDoubleClickEvents {
 }
 
 - (void)setTitle:(NSString*)title {
+  if ([title containsANSICodes]) {
+    [[statusItem_ button]
+        setAttributedTitle:[[title copy] attributedStringParsingANSICodes]];
```
miniak: no need for the copy here

DeerMichel: Yeah, it returns a new string, in particular a `NSMutableAttributedString`. I think I put the `copy` there just in case.

DeerMichel: Right, i'll change that.

DeerMichel:
@codebytere ready

DeerMichel:
Do we finally want to merge this? 🚀

shell/browser/ui/tray_icon_cocoa.mm
```diff
@@ -465,16 +265,15 @@ - (BOOL)shouldHighlight {
...
 }
 
 void TrayIconCocoa::SetImage(const gfx::Image& image) {
   [status_item_view_ setImage:image.IsEmpty() ? nil : image.AsNSImage()];
 }
 
 void TrayIconCocoa::SetPressedImage(const gfx::Image& image) {
-  [status_item_view_ setAlternateImage:image.AsNSImage()];
+  [status_item_view_
+      setAlternateImage:image.IsEmpty() ? nil : image.AsNSImage()];
```
miniak: just use `image.AsNSImage()`, the methods returns `nil` when `IsEmpty()`

miniak: https://cs.chromium.org/chromium/src/ui/gfx/image/image.cc?sq=package:chromium&g=0&l=578-580

shell/browser/ui/tray_icon_cocoa.mm
```diff
@@ -465,16 +265,15 @@ - (BOOL)shouldHighlight {
 
 TrayIconCocoa::~TrayIconCocoa() {
   [status_item_view_ removeItem];
-  if (menu_model_)
-    menu_model_->RemoveObserver(this);
 }
 
 void TrayIconCocoa::SetImage(const gfx::Image& image) {
   [status_item_view_ setImage:image.IsEmpty() ? nil : image.AsNSImage()];
```
miniak: just use `image.AsNSImage()`, the methods returns `nil` when `IsEmpty()`

miniak: https://cs.chromium.org/chromium/src/ui/gfx/image/image.cc?sq=package:chromium&g=0&l=578-580

DeerMichel:
Ready

codebytere:
@DeerMichel release notes need to be fixed up it seems.

softwheredon:
Hi, @deermichel 

> However, we are still able to control highlighting when no menu is set (i guess that's where most use cases for this method are found anyway)

How can we do this?

sulf:
Hey @deermichel, 

>However, we are still able to control highlighting when no menu is set (i guess that's where most use cases for this method are found anyway)

We've been looking at the same issue, but cannot find any documentation.

donkooijman:
^ This can be achieved by setting an empty context menu on the tray. @sulf 

`const contextMenu = Menu.buildFromTemplate([]);`
`tray.setContextMenu(contextMenu);`

CahierX:
> 

I set empty array click tray is not highlight, is not work

CahierX:
> 

How did you do it， I encountered the same problem

Issue #18933, opened 2019-06-21
'level' param of setAlwaysOnTop is not implemented on Windows


CapOM:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description
Currently it is not possible to make a BrowserWindow always on top but still behind the task bar on Windows. Whereas It is implemented on MacOS with levels "floating", ... ",
"status" (then "pop-up-menu" and "screen-saver" will make it above the dock)

### Proposed Solution
Win32 api allows to do that, I have a local patch that I will submit next week if interested.
In my patch I re-use the same level names as for MacOS to make it easier for the user.

### Alternatives Considered
Choose different level names on Windows.

codebytere:
@CapOM that'd be awesome and appreciated 🙇‍♀

codebytere:
Closed by https://github.com/electron/electron/pull/18982.

PR #18982, opened 2019-06-25
feat: enable BrowserWindow level placement on Windows

CapOM:
#### Description of Change

For now it only adds the ability to place the window below
the task bar while still being always on top.
Previous behaviour was always showing the window above the task
bar when top is true. We keep this default behaviour, i.e. when
the 'level' parameter is omitted.

https://github.com/electron/electron/issues/18933

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Can set a window always on top but behind the taskbar on Windows


CapOM:
CC @codebytere

docs/api/browser-window.md
```diff
@@ -1102,8 +1102,14 @@ On Linux always returns `true`.
 * `flag` Boolean
 * `level` String (optional) _macOS_ - Values include `normal`, `floating`,
   `torn-off-menu`, `modal-panel`, `main-menu`, `status`, `pop-up-menu`,
-  `screen-saver`, and ~~`dock`~~ (Deprecated). The default is `floating`. See the
-  [macOS docs][window-levels] for more details.
+  `screen-saver`, and ~~`dock`~~ (Deprecated). The default is `floating` when
+  `flag` is true. The `level` is reset to `normal` when the flag is false. See
+  the [macOS docs][window-levels] for more details.
+* `level` String (optional) _Windows_ - Values are the same as on macOS to make
```
codebytere: This shouldn't be the a different parameter with the same name; that'll confuse our documentation parser. This should just specify what values of the existing `level` parameter now apply to Windows.

shell/browser/native_window_views.cc
```diff
@@ -757,6 +759,38 @@ void NativeWindowViews::SetAlwaysOnTop(bool top,
     NativeWindow::NotifyWindowAlwaysOnTopChanged();
 
   widget()->SetAlwaysOnTop(top);
+
+#if defined(OS_WIN)
+  // Re-use level names from MacOS to have the same electron api.
+  if (top) {
+    bool above_task_bar = true;
```
codebytere: We can probably make this logic cleaner by considering which `level` values change the default; something like: 
```cpp
if (top) {
    bool above_task_bar = true;
    std::vector<std::string> levels = {"floating", "torn-off-menu", "modal-panel", "main-menu", "status"};
    if (std::find(levels.begin(), levels.end(), value) != levels.end()) {
      above_task_bar = false;
    }
    
    if (!above_task_bar) {
      // Behind the taskbar.
      const HWND task_bar_hwnd = ::FindWindow(kUniqueTaskBarClassName, NULL);
      ::SetWindowPos(GetAcceleratedWidget(), task_bar_hwnd, 0, 0, 0, 0,
                    SWP_NOMOVE | SWP_NOSIZE | SWP_NOACTIVATE);
    }
}
```

We don't need to specifically check for the final three values ("pop-up-menu", "screen-saver", "dock") since they wouldn't change the default value you've already set at the top of the conditional.

CapOM: Done

CapOM: Thx a lot for the suggestion.

docs/api/browser-window.md
CapOM: Done.

CapOM:
The test failures seem unrelated to this change. I addressed the remarks, please take a look, thx!

CapOM:
Hi, thx for the testing and suggestions. One could argue that the user of `setAlwaysOnTop` can still watch for the `focus` event and call this api again. But I am ok to internally ensure the `behind taskbar` property upon `NativeWindowViews::OnWidgetActivationChanged(when active is true)`, i.e. `focus`. I will also do it for `restore`, `resize` and `show`. That should cover most of the cases. Remaining corner cases will have to be handle for the user of this api. How does this plan sound ?

CapOM:
Hi @brenca , I addressed your remarks. Actually `OnWidgetActivationChanged` is internally called upon activate, restore and show so no need to ensure the placement else where.
Please take a look thx!

shell/browser/native_window_views.cc
```diff
@@ -757,6 +759,20 @@ void NativeWindowViews::SetAlwaysOnTop(bool top,
...
+#if defined(OS_WIN)
+  // Reset the placement flag.
+  behind_task_bar_ = false;
+  if (top) {
+    // On macOS the window is placed behind the Dock for the following levels.
+    // Re-use the same names on Windows to make it easier for the user.
+    const std::vector<std::string> levels = {
+        "floating", "torn-off-menu", "modal-panel", "main-menu", "status"};
+    behind_task_bar_ =
+        std::find(levels.begin(), levels.end(), level) != levels.end();
```
miniak: you can use `base::Contains` from `base/stl_util.h`

shell/browser/native_window_views.cc
```diff
@@ -757,6 +759,20 @@ void NativeWindowViews::SetAlwaysOnTop(bool top,
...
 
   widget()->SetAlwaysOnTop(top);
+
+#if defined(OS_WIN)
+  // Reset the placement flag.
+  behind_task_bar_ = false;
+  if (top) {
+    // On macOS the window is placed behind the Dock for the following levels.
+    // Re-use the same names on Windows to make it easier for the user.
+    const std::vector<std::string> levels = {
```
miniak: don't allocate this on the heap unnecessarily
```
static const auto levels[] = {
        "floating", "torn-off-menu", "modal-panel", "main-menu", "status"};
```

shell/browser/native_window_views.cc
```diff
@@ -1346,6 +1364,17 @@ ui::WindowShowState NativeWindowViews::GetRestoredState() {
   return ui::SHOW_STATE_NORMAL;
 }
 
+void NativeWindowViews::MoveBehindTaskBarIfNeeded() {
+#if defined(OS_WIN)
+  if (behind_task_bar_) {
+    const HWND task_bar_hwnd = ::FindWindow(kUniqueTaskBarClassName, NULL);
```
miniak: use `nullptr`

shell/browser/native_window_views.cc
CapOM: Done

shell/browser/native_window_views.cc
CapOM: Done

shell/browser/native_window_views.cc
CapOM: Done, thx

CapOM:
Hi @miniak , thx for the review, I addressed the remarks, please take a look, thx!

CapOM:
Hi @miniak , hi @brenca , gentle ping ?

codebytere:
```
autoUpdater behavior "before each" hook for "should have a valid code signing identity" - "before each" hook for "should have a valid code signing identity"
```
is an expected failure for fork PRs and does not block merge.

Issue #18295, opened 2019-05-14
MenuItem registerAccelerator does not work


Maxime117:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.1
* **Operating System:**
  * macOS 10.14.4

### Expected Behavior
Hello,

For some reasons, I have to use the window.keydown method instead of the MenuItem accelerator, but I still want to display the shortcut to the right of the MenuItem. So I used 'registerAccelerator: false'. But the click() function is always called when I hit "B".

### To Reproduce
```
{
  label: "Blade",
  id: "blade",
  accelerator: "B",
  registerAccelerator: false,
  click(a, b, e) {
    console.log(e);
  }
}
```

codebytere:
@Maxime117 did this work in any previous versions?

Maxime117:
@codebytere It didn't work in v4.x.x

DeerMichel:
So me and @erickzhao played around with this issue and at least on Mac we're kinda stuck because of the limited `NSMenuItem` API. Setting the accelerator also causes the shortcut label to be shown and not setting the accelerator (to prevent it from firing the click handler) does result in not showing the label. There is no "Apple" way to separate those two things.

I fiddled around with Xcode and came to a possible solution for displaying a shortcut label without setting the accelerator using a custom view for `NSMenuItem` (containing 2 labels: menu label aligned left, shortcut label aligned right):
![image](https://user-images.githubusercontent.com/8472721/59538877-11785d80-8eb0-11e9-9d7c-6c34890a46f7.png)
Problem is, this involves reengineering the whole `NSMenuItem` to match style and behavior of the native view. Probably too much effort for an edge case like this.

_Side note_: we didn't investigate how the property behaves on Linux or Windows.

**TL;DR** two options:
- remove the `registerAccelerator` property; though someone needs to check how this affects #15719 because it reverts the fix from #15723
- implement custom view on Mac and adapt other platforms as needed

cc @codebytere

codebytere:
[...]

PR #18985, opened 2019-06-25
docs: menuItem.registerAccelerator not available on macOS

DeerMichel:
#### Description of Change
Follow-up to #18295. Until either we can find a better solution or until Apple allows us greater `NSMenuItem` granularity, the `registerAccelerator` property only works on Windows and Linux. This should be stated in the docs.

cc @codebytere @erickzhao 

#### Checklist
- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/menu-item.md
+++ b/docs/api/menu-item.md
@@ -28,7 +28,7 @@ See [`Menu`](menu.md) for examples.
   * `visible` Boolean (optional) - If false, the menu item will be entirely hidden.
   * `checked` Boolean (optional) - Should only be specified for `checkbox` or `radio` type
     menu items.
-  * `registerAccelerator` Boolean (optional) - If false, the accelerator won't be registered
+  * `registerAccelerator` Boolean (optional) _Linux_ _Windows_ - If false, the accelerator won't be registered
     with the system, but it will still be displayed. Defaults to true.
   * `submenu` (MenuItemConstructorOptions[] | [Menu](menu.md)) (optional) - Should be specified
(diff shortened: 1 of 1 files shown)


docs/api/menu-item.md
```diff
@@ -28,7 +28,7 @@ See [`Menu`](menu.md) for examples.
   * `visible` Boolean (optional) - If false, the menu item will be entirely hidden.
   * `checked` Boolean (optional) - Should only be specified for `checkbox` or `radio` type
     menu items.
-  * `registerAccelerator` Boolean (optional) - If false, the accelerator won't be registered
+  * `registerAccelerator` Boolean (optional) _Linux_ _Windows_ - If false, the accelerator won't be registered
```
codebytere: Platform specificity should go at the end of the description:
```
* `registerAccelerator` Boolean (optional) - If false, the accelerator won't be registered with the system, but it will still be displayed. Defaults to `true`. _Linux_ _Windows_
```

codebytere: Platform specificity should go at the end of the description:
```
* `registerAccelerator` Boolean (optional) - If false, the accelerator won't be registered with the system, but it will still be displayed. Defaults to `true`. _Linux_ _Windows_
```

DeerMichel: I copied the style from [here](https://github.com/electron/electron/blob/master/docs/api/app.md#appsetasdefaultprotocolclientprotocol-path-args)

codebytere: Rip cool it seems we apparently have both

In that case np, carry on 😆

PR #18986, opened 2019-06-25
test: move tray api tests to main process

DeerMichel:
#### Description of Change
- moved tray api tests from remote to main process
- migrated tests to typescript (no definition changes needed)

_nb:_ i might add some tests after #18981 got merged

cc @nornagon @codebytere 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/spec-main/api-tray-spec.ts
+++ b/spec-main/api-tray-spec.ts
@@ -31,7 +31,7 @@ describe('tray module', () => {
       expect(tray.isDestroyed()).to.be.false('tray should not be destroyed')
       tray.destroy()
 
-      expect(tray.isDestroyed()).to.be.true('tray is not destroyed')
+      expect(tray.isDestroyed()).to.be.true('tray should be destroyed')
     })
   })
 
--- a/spec-main/api-tray-spec.ts
+++ b/spec-main/api-tray-spec.ts
@@ -28,7 +28,7 @@ describe('tray module', () => {
 
   describe('tray.destroy()', () => {
     it('destroys a tray', () => {
-      expect(tray.isDestroyed()).to.be.false('tray is destroyed')
+      expect(tray.isDestroyed()).to.be.false('tray should not be destroyed')
       tray.destroy()
 
       expect(tray.isDestroyed()).to.be.true('tray is not destroyed')
--- a/spec-main/api-tray-spec.ts
+++ b/spec-main/api-tray-spec.ts
@@ -1,7 +1,7 @@
 import { expect } from 'chai'
 import { Menu, Tray, nativeImage } from 'electron'
 
-describe.only('tray module', () => {
+describe('tray module', () => {
   let tray: Tray;
 
   beforeEach(() => {
--- a/spec-main/api-tray-spec.ts
+++ b/spec-main/api-tray-spec.ts
@@ -28,10 +28,10 @@ describe('tray module', () => {
 
   describe('tray.destroy()', () => {
     it('destroys a tray', () => {
-      expect(tray.isDestroyed()).to.equal(false)
+      expect(tray.isDestroyed()).to.be.false('tray is destroyed')
       tray.destroy()
 
-      expect(tray.isDestroyed()).to.equal(true)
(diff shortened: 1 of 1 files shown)


spec-main/api-tray-spec.ts
```diff
@@ -1,42 +1,41 @@
-const { remote } = require('electron')
-const { expect } = require('chai')
-const { Menu, Tray, nativeImage } = remote
+import { expect } from 'chai'
+import { Menu, Tray, nativeImage } from 'electron'
 
 describe('tray module', () => {
-  let tray
+  let tray: Tray | null = null
```
MarshallOfSound: I'd prefer to do
```suggestion
  let tray: Tray;
```

Then do a workaround in `afterEach` to set it to null rather than have to use `!` everywhere 🤔

DeerMichel: oki done sth - imo a bit hacky, but works 👀

spec-main/api-tray-spec.ts
```diff
@@ -26,18 +28,16 @@ describe('tray module', () => {
 
   describe('tray.destroy()', () => {
     it('destroys a tray', () => {
-      expect(tray.isDestroyed()).to.be.false()
+      expect(tray.isDestroyed()).to.equal(false)
```
nornagon:
```suggestion
      expect(tray.isDestroyed()).to.be.false('tray is destroyed')
```

this helps a bit when an error is reported, showing what the boolean represents in addition to the info that it's not what was expected.

spec-main/api-tray-spec.ts
```diff
@@ -26,18 +28,16 @@ describe('tray module', () => {
...
       tray.destroy()
 
-      expect(tray.isDestroyed()).to.be.true()
-      tray = null
+      expect(tray.isDestroyed()).to.equal(true)
```
nornagon: same here:
```suggestion
      expect(tray.isDestroyed()).to.be.true('tray is destroyed')
```

spec-main/api-tray-spec.ts
```diff
@@ -1,18 +1,20 @@
...
 describe('tray module', () => {
-  let tray
+  let tray: Tray;
 
   beforeEach(() => {
     tray = new Tray(nativeImage.createEmpty())
   })
 
+  afterEach(() => {
+    tray = null as any
```
nornagon: The pattern I use elsewhere is
```suggestion
    tray = null as unknown as Tray
```

unclear to me if that's any better though? 🤷‍♂

DeerMichel: hmm apart from the fact that it looks more complicated... probably doesn't matter

MarshallOfSound: I think in other places I just used `as any` to force this null resetting 👍

spec-main/api-tray-spec.ts
```diff
@@ -26,18 +28,16 @@ describe('tray module', () => {
 
   describe('tray.destroy()', () => {
     it('destroys a tray', () => {
-      expect(tray.isDestroyed()).to.be.false()
+      expect(tray.isDestroyed()).to.be.false('tray is destroyed')
```
MarshallOfSound: These strings should indicate the _expected_ case IMO not what actually happened.  i.e. following the _should_ syntax

```suggestion
      expect(tray.isDestroyed()).to.be.false('tray should not be destroyed')
```

This is because the error that gets thrown looks like this

```
[AssertionError: message here: expected false to be true]
```

So this would be

```
[AssertionError: tray should not be destroyed: expected false to be true]
```

spec-main/api-tray-spec.ts
```diff
@@ -26,18 +28,16 @@ describe('tray module', () => {
...
       tray.destroy()
 
-      expect(tray.isDestroyed()).to.be.true()
-      tray = null
+      expect(tray.isDestroyed()).to.be.true('tray is not destroyed')
```
MarshallOfSound: Same here

```suggestion
      expect(tray.isDestroyed()).to.be.true('tray should be destroyed')
```

PR #18990, opened 2019-06-26
fix: do not wait on promise returned by remote APIs

### LINKED ISSUE #18638: Setting new src in webview is slow in v6
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 6.0.0-beta.6
* **Operating System:**
  * macOS 10.14.5
  * 5.0.2

### Expected Behavior
Setting new src in previous version of electron would take only 1–2ms.

### Actual Behavior
Setting new src in electron v6 takes ~600–700ms.

### To Reproduce
1. Clone this repo: https://github.com/steverandy/electron-quick-start/
2. Checkout branch `slow-webview`
3. Run `npm install` and `npm start`
4. Open devtool to see the time took to set new src

### Screenshots

Screenshot 2019-06-05 at 10 23 12 AM

zcbenz:
#### Description of Change

Close #18638.

This PR fixes the bug in `ipcMainUtils.handle` that, the helper waits on Promise returned by handler when it shouldn't.
 
For the original issue, `ipcMainUtils.handle` waits on the Promised returned by `WebContents.loadURL`, resulting in setting src being blocked until the URL is loaded. 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix setting `src` on `<webview>` being too slow.

### DIFF
--- a/lib/common/web-view-methods.js
+++ b/lib/common/web-view-methods.js
@@ -3,7 +3,6 @@
 // Public-facing API methods.
 exports.syncMethods = new Set([
   'getURL',
-  'loadURL',
   'getTitle',
   'isLoading',
   'isLoadingMainFrame',
@@ -54,6 +53,7 @@ exports.syncMethods = new Set([
 ])
 
 exports.asyncMethods = new Set([
+  'loadURL',
   'capturePage',
   'executeJavaScript',
   'insertCSS',
--- a/docs/api/webview-tag.md
+++ b/docs/api/webview-tag.md
@@ -277,6 +277,11 @@ webview.addEventListener('dom-ready', () => {
   * `postData` ([UploadRawData[]](structures/upload-raw-data.md) | [UploadFile[]](structures/upload-file.md) | [UploadBlob[]](structures/upload-blob.md)) (optional)
   * `baseURLForDataURL` String (optional) - Base url (with trailing path separator) for files to be loaded by the data url. This is needed only if the specified `url` is a data url and needs to load other files.
 
+Returns `Promise<void>` - The promise will resolve when the page has finished loading
+(see [`did-finish-load`](webview-tag.md#event-did-finish-load)), and rejects
+if the page fails to load (see
+[`did-fail-load`](webview-tag.md#event-did-fail-load)).
+
 Loads the `url` in the webview, the `url` must contain the protocol prefix,
 e.g. the `http://` or `file://`.
 
--- a/lib/renderer/web-view/web-view-attributes.ts
+++ b/lib/renderer/web-view/web-view-attributes.ts
@@ -196,7 +196,7 @@ class SrcAttribute extends WebViewAttribute {
     const method = 'loadURL'
     const args = [this.getValue(), opts]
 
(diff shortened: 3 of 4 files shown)


miniak:
I get the problem we are trying to solve, however I am not sure if the solution is correct. I think about it a bit more

zcbenz:
@miniak Another way to fix it is to split `ipcMainUtils.handle` into 2 parts, one handling normal functions and one handling async functions. Would that feel right to you?

zcbenz:
I have refactored the code to add `ipcMainUtils.handleAsync`, so the behavior of the helper is more clear.

miniak:
@zcbenz We actually had `handle` and `handleSync` before :) I've merged them into a single method, as I wanted the caller to decide whether to use `invoke` or `invokeSync` and have the handler be generic.

miniak:
I would suggest one of these 2 solutions:
- https://github.com/electron/electron/compare/miniak/webview-load-url?expand=1
- https://github.com/electron/electron/compare/miniak/webview-load-url-v2?expand=1

zcbenz:
@miniak The results of `asyncMethods` are still being awaited, so `<webview>.loadURL` still won't work correctly in first version.

I'm good with the second version.

miniak:
@zcbenz it is awaited but it does not block the renderer.

zcbenz:
@miniak Yeah you are right, both versions should work then.

zcbenz:
@miniak Do you want to open a PR for your solution?

lib/common/web-view-methods.js
```diff
@@ -54,6 +53,7 @@ exports.syncMethods = new Set([
 ])
 
 exports.asyncMethods = new Set([
+  'loadURL',
```
jkleinsc: Shouldn't there be a corresponding doc change to note the loadURL is now async?

zcbenz: @jkleinsc The `<webview>.loadURL` was actually asynchronous while it was under put `syncMethods`, the `WebContents.loadURL` API works asynchronously that it only starts the loading and returns immediately without waiting for anything.

So with this PR `<webview>.loadURL` won't change any documented behavior.

The side effect of this PR is making `<webview>.loadURL` return a `Promise`, but since we are backporting it to 6.0, I think it is probably better keeping it undocumented otherwise it would be a `feat`.

deepak1556:
I agree @jkleinsc we should document the api signature change for master, for the backport we can do two things either, take this as a feature and get it through the release-wg process which should be easy based on

https://github.com/electron/governance/tree/master/wg-releases#feature-backport-requests-for-major-versions-in-beta

```
Backport requests after Week 3 will only be considered if there is a very good reason. For backport requests please submit the link to the PR and reason for backport to the Releases WG agenda for consideration.
```

or the backport PR can create a separate handler for `loadURL` that still uses the `ipcRendererUtils.invoke` but maintains the current return value type.

zcbenz:
I have added documentation for the return value change.

For the backport (#19190), I found the code would be rather dirty if we want to keep the return value unchanged, as currently `<webview>.loadURL` returns a serialized version of Promise object. So I kept the backport as it is and it requires release-wg to approve.

PR #18993, opened 2019-06-26
fix: make run-gn-format work properly on Windows

brenca:
#### Description of Change

Pretty much what the title says, `run-gn-format.py` didn't work for me locally in `PowerShell` because of a missing `shell=True`.

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/run-gn-format.py
+++ b/script/run-gn-format.py
@@ -15,6 +15,7 @@ def main():
   for gn_file in sys.argv[1:]:
     subprocess.check_call(
       ['gn', 'format', gn_file],
+      shell=True,
       env=new_env
     )
 


script/run-gn-format.py
```diff
@@ -15,6 +15,7 @@ def main():
   for gn_file in sys.argv[1:]:
     subprocess.check_call(
       ['gn', 'format', gn_file],
+      shell=True,
```
nornagon: This doesn't seem like a good fix to me. `shell=True` has [known security issues](https://docs.python.org/2/library/subprocess.html#frequently-used-arguments), and if you end up on a branch that has a gn file called `"| rm -rf /.gn"` you're gonna have a bad time.

Can you explain how this fixes the issue? Can we find a better fix?

nornagon:
ping @zcbenz @brenca i'm not comfortable with this fix, can we find a different way to do this?

zcbenz:
I'm good reverting this one.

PR #18995, opened 2019-06-26
fix: Correct modal focus behavior on macOS

### LINKED ISSUE #18502: MacOS modal focus
On MacOS (Windows/Linux work fine) I am opening a modal window. The modal window opens up fine but focus behavior is broken. When I click on the main window (outside of modal) it gains the focus (which should not happen) and then the modal window gains focus (which is correct), but right after that the modal gains blur (which is wrong). Now if I try to focus the modal with mouse it does not do anything as the main window is focused. 

Here I have created a demo which prints out to console which window is focused. 
[https://github.com/zarubond/electron-quick-start/tree/modal_focus](https://github.com/zarubond/electron-quick-start/tree/modal_focus)

### Preflight Checklist

### Issue Details

* **Electron Version:**
Tested on Electron 3, 5, 6-beta4
* **Operating System:**
MacOS 10.13 (Windows 10 and Ubuntu 18.04 work fine)
None

### Expected Behavior
Main window should never get focused if the modal window is opened

### Actual Behavior
Main window gains focused event if the modal window is opened.

### To Reproduce
Open up the console to see which windows are getting focused and the open the modal with open button.

[https://github.com/zarubond/electron-quick-start/tree/modal_focus](https://github.com/zarubond/electron-quick-start/tree/modal_focus)

### Screenshots
image
(ignore the last line, I had to switch to screen-short program)

erickzhao:
#### Description of Change
Fixes #18502

This PR changes the `focus` and `blur` events that we emit in Electron to listen to changes in key window rather than main window. It swaps out `windowDidBecomeMain` and `windowDidResignMain` for `windowDidBecomeKey` and `windowDidResignKey`, respectively.

It seems to work as intended now. Also added a new test to test this exact situation. Please let me know if there's a less convoluted way to test this specific case, since it seems like I'm listening to a very specific series of events. Test seems to be working locally on my end, and fails on a downloaded CI build from `master`. ✅ 

##### Context

In our macOS code, the `NativeWindowMac::Show()` function in `native_window_mac.mm` calls this [`beginSheet`](https://developer.apple.com/documentation/appkit/nswindow/1419653-beginsheet) function, which specifies that:
>If the window has no presented sheets, this method displays the specified sheet, **makes it key**, and returns control to the caller. While the sheet remains visible, most events targeted at the receiver are prohibited.

For additional context, see [Apple developer documentation on key and main windows](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/WinPanel/Concepts/ChangingMainKeyWindow.html) here.

Fiddle: https://gist.github.com/c9c19684f6e0d23c102161fb3e86500b

----------------------

cc @codebytere @DeerMichel 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed bug on macOS where the main window could be targeted for a `focus` event when it was disabled behind a modal.

### DIFF
--- a/spec/api-browser-window-spec.js
+++ b/spec/api-browser-window-spec.js
@@ -1639,7 +1639,7 @@ describe('BrowserWindow module', () => {
 
   describe('focus event', () => {
     it('should not emit if focusing on a main window with a modal open', (done) => {
-      let childWindowClosed = false
+      const childWindowClosed = false
       const child = new BrowserWindow({
         parent: w,
         modal: true,
@@ -1652,21 +1652,11 @@ describe('BrowserWindow module', () => {
 
       child.on('show', () => {
         w.once('focus', () => {
-          expect(childWindowClosed).to.equal(true,
-            'Main window should only regain focus once the modal is closed')
+          expect(child.isDestroyed()).to.equal(true)
           done()
         })
         w.focus() // this should not trigger the above listener
         child.close()
-      })
-
-      // trying stuff...
-      child.on('close', () => {
-        childWindowClosed = true
-      })
-
-      child.on('closed', () => {
-        childWindowClosed = true
       })
 
       // act
--- a/spec/api-browser-window-spec.js
+++ b/spec/api-browser-window-spec.js
@@ -1637,6 +1637,44 @@ describe('BrowserWindow module', () => {
     })
   })
 
+  describe('focus event', () => {
+    it('should not emit if focusing on a main window with a modal open', (done) => {
+      let hasMainFocusAfterModalFocus = false
+      const child = new BrowserWindow({
+        parent: w,
+        modal: true,
+        show: false
+      })
+
+      child.once('ready-to-show', () => {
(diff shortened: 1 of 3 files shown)


spec/api-browser-window-spec.js
```diff
@@ -1637,6 +1637,44 @@ describe('BrowserWindow module', () => {
...
+          hasMainFocusAfterModalFocus = true
+        })
+        w.focus()
+        // close child to end the spec
+        child.close()
+      })
+
+      // clean up and assert that main window never
+      // regained focus
+      child.once('closed', () => {
```
jkleinsc: I think it would be a good idea to also verify that you can focus the main window once the modal is closed.  You can do that in this test (as opposed to adding a new test).

erickzhao:
@jkleinsc Seems like calling `child.isDestroyed()` and not relying on `focus` and `closed` emission ordering helped make the spec pass on Linux and Windows!

PR #18999, opened 2019-06-27
fix: crash when tray popup called twice

codebytere:
#### Description of Change

When the Tray popup functionality was made asynchronous, `ScopedPumpMessagesInPrivateModes` was introduced to ensure that the UI can update while the menu is fading out. However, given that we're now asynchronous, there exists a possibility that we can call `popUpContextMenu` more than once and this will cause a crash since we're no longer blocking the main process.

As a result, we only want to call  `ScopedPumpMessagesInPrivateModes` if we're going to open the menu, which means that we first need to ensure that the menu is not already open. This PR thus fixes the aforementioned crash which can be seen in the following code before this PR and is fixed with the code therein.

```js
  tray.setContextMenu(contextMenu)
  tray.popUpContextMenu()
  tray.popUpContextMenu()
```

cc @DeerMichel @miniak 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash on tray popup being called multiple times in a row.

### DIFF
--- a/shell/browser/ui/tray_icon_cocoa.mm
+++ b/shell/browser/ui/tray_icon_cocoa.mm
@@ -340,9 +340,6 @@ const CGFloat kVerticalTitleMargin = 2;
   // Make sure events can be pumped while the menu is up.
   base::MessageLoopCurrent::ScopedNestableTaskAllower allow;
 
-  // Ensure the UI can update while the menu is fading out.
-  base::ScopedPumpMessagesInPrivateModes pump_private;
-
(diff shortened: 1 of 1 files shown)


DeerMichel:
Can you add a test for this or is testing tray stuff difficult on CI?

codebytere:
@DeerMichel the tray tests don't work presently on Mac; for now no but i was planning to dig further into this once your PR to move the tests into the main process gets merged 🚀

Issue #18980, opened 2019-06-25
Better callback value for webContents.print()


popod:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** 5.0.5
* **Operating System:** macOS Mojave 10.14.5 / Windows too ?
* **Last Known Working Electron version:** not known

### Expected Behavior

When using `webContents.print()`, if the user click the **cancel button** on the print dialog (or on the **"PDF" button**) this should: return a callback value of "false" (error) or "true" (success), but with more details that only "true" or "false". Something like an error/success message that tell that the cancel button has been clicked..

This could be:

- `success` when the document is sent to the printer
- `cancelled` when the cancel button is clicked
- `pdf` / `other-action`? when the PDF button is clicked
- `failed` when an error occurs

### Actual Behavior

In my application, I only want to display a success notification to the user when the print process is successfully launched and an error message when they is an error. For now, false-positive notifications are display when :

- the cancel button is clicked on the print dialog (return is `false`), we do not know if it is an `error` or a `cancel` event and when an error occurs (return is `false` too), we do not know if it is an `error` or a `cancel` event.
- the "PDF" button (only on macOS?) is clicked (return is `true`) and we would know if this is a print `success` or something else to display a "successfully print" notification to the user.

Not tested, but Windows should have the same problem !

---

Perhaps linked with #15568 but more specific ?
@codebytere perhaps something that is linked with #17400

codebytere:
@popod what are you testing this with? If I run the original repro you sent in, i see `false` in the console for manual cancellation, like i expect.

However, i think we should potentially be able to callback a string of either `failed`, `success` or `cancelled` depending on the outcome of the print; i'll add that to my backlog.

popod:
@codebytere I do not provide any repo for this issue.. I think you use the repo for the bug where the callback only return `false` (#17265) which use electron `4.x.x` and this is why you get a `false` value. If you want to use this repo, you should update the electron version to `5.0.5`.

> However, i think we should potentially be able to callback a string of either `failed`, `success` or `cancelled` depending on the outcome of the print; i'll add that to my backlog.

[...]

Issue #15568, opened 2018-11-05
Add error information to webContents.print() callback


alexb148:
I've been having some problems getting webContents.print() to work on some machines, but at present there's no way to determine what the error is - the callback simply returns a true or false boolean.

A more detailed error log would be incredibly useful, is it possible to get more info returned if something goes wrong?

The function referenced is in your docs here:

https://github.com/electron/electron/blob/master/docs/api/web-contents.md#contentsprintoptions-callback

codebytere:
Closed by https://github.com/electron/electron/pull/19000.

popod:
@codebytere #19000 is a very useful thing !

But would it be possible to return even more informations about the error ? In the case where `failureReason ` value is `failed` ? Something like the error code and/or message ?

Thanks for reply :)

PR #19000, opened 2019-06-27
feat: improve callback value for webContents.print()

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/18980.

Adds granularity to the optional callback in `webContents.print()` by adding a `failureType` value in addition to the `success` boolean that differentiates between `cancelled` and `failed` print jobs.

Tested with [this repo](https://github.com/popod/electron-quick-start/tree/print-bug-1).

cc @deepak1556 @brenca 

also cc @popod as i know you're tracking printing work :)

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Improved `webContents.print()` callback value to distinguish between success, failure, and cancellation.

### DIFF
--- a/patches/chromium/printing.patch
+++ b/patches/chromium/printing.patch
@@ -46,7 +46,7 @@ index 88a6142eea4c7a219c08fe3463c44711f5c9fada..81db315a0036a123658697aa677e2356
  
  void PrintJobWorker::GetSettingsWithUI(int document_page_count,
 diff --git a/chrome/browser/printing/print_view_manager_base.cc b/chrome/browser/printing/print_view_manager_base.cc
-index 7ba43aada1ac44827cca264d6f37814e4a91f458..d811f5268334e8b75f610f1f7240a335582c67fa 100644
+index 7ba43aada1ac44827cca264d6f37814e4a91f458..1ca9fe408ec9a5129f6cec46daeeee46bab60983 100644
 --- a/chrome/browser/printing/print_view_manager_base.cc
 +++ b/chrome/browser/printing/print_view_manager_base.cc
 @@ -27,10 +27,7 @@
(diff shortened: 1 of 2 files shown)


popod:
I think this is a "little" breaking change and that it should be indicated in the release notes.

- with older electron versions `webContents.print()` return `false` in case of errors
- with newer electron versions `webContents.print()` return `failed` which could be interpreted as `true` (but we attempt to have `false` with older electron logic)

codebytere:
@popod yeah, this wouldn't be backported and the upcoming change would be indicated in notes and code as ability permits!

popod:
Juste for informations, @MarshallOfSound proposal would solve #15568

MarshallOfSound:
> And why not backporting this to v6 which is still in beta for now

@popod We follow semver and this being a feature would be `semver/minor`, they can't be backported to stable release lines.  Admitting a feature to an existing beta release line requires a discussion + good reasoning + a vote in the Releases WG.  With 6 so close to the end of its beta cycle it's unlikely any more features will be admitted to destabilization.

patches/chromium/printing.patch
```diff
@@ -204,19 +214,27 @@ index 7ba43aada1ac44827cca264d6f37814e4a91f458..c41b0c24974147e847baa21b9b489261
...
++#include "v8/include/v8.h"
+ 
+ struct PrintHostMsg_DidPrintDocument_Params;
+ 
+@@ -39,6 +40,8 @@ class PrintJob;
  class PrintQueriesQueue;
  class PrinterQuery;
  
-+using CompletionCallback = base::OnceCallback<void(bool)>;
++using CompletionCallback = base::OnceCallback<void(bool, v8::Local<v8::Value>)>;
```
zcbenz: Should just pass `const std::string&`, V8 conversation would be done automatically.

PR #19004, opened 2019-06-27
refactor: pass preload script paths to renderer via IPC

miniak:
#### Description of Change
Pass the list of preload scripts to be loaded in non-sandboxed renderers via IPC as the command-line is shared when the same renderer process is hosting multiple `BrowserWindows` (scriptable popup with  `nativeWindowOpen: true` for example).

The new `ELECTRON_INIT_RENDERER` IPC replaces `ELECTRON_GET_CONTENT_SCRIPTS` introduced in #18823.

There will be a follow-up PR to pass the remaining options via IPC instead of command-line.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Preload scripts configured for popups when handling the `new-window` event now work properly regardless of whether a new renderer process is hosting the contents (cross-origin) or the same renderer (same-origin).

### DIFF
--- a/lib/browser/rpc-server.js
+++ b/lib/browser/rpc-server.js
@@ -3,6 +3,7 @@
 const electron = require('electron')
 const { EventEmitter } = require('events')
 const fs = require('fs')
+const path = require('path')
 
 const v8Util = process.electronBinding('v8_util')
 const eventBinding = process.electronBinding('event')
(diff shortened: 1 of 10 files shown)


lib/browser/rpc-server.js
```diff
@@ -519,27 +520,45 @@ if (features.isDesktopCapturerEnabled()) {
...
+
+  const preload = contents._getPreloadPath()
+  if (preload) {
+    result.push(preload)
+  }
 
+  return result
+}
+
+ipcMainUtils.handle('ELECTRON_INIT_RENDERER', async function (event) {
```
zcbenz: The `async` is unnecessary here as there is no async operations, adding `async` would make a difference with https://github.com/electron/electron/pull/18990.

I do hope to getting #18990 merged before adding more usages of `ipcMainUtils.handle`.

miniak:
Replaced by #19043

PR #19012, opened 2019-06-27
refactor: improve function deprecation module

DeerMichel:
#### Description of Change
Previously, functions could be marked for renaming or removal using the single method `deprecate.function`. I split that functionality into two separate methods. 

cc @codebytere 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/lib/common/api/deprecate.ts
+++ b/lib/common/api/deprecate.ts
@@ -39,13 +39,11 @@ const deprecate: ElectronInternal.DeprecationUtil = {
     // if the function has already been removed, warn about it
     if (!fn) {
       deprecate.log(`Unable to remove function '${removedName}' from an object that lacks it.`)
-      return function (this: any) {
-        deprecate.log(`'${removedName} function' has been removed. See release notes for further information.`)
-      }
+      throw Error(`'${removedName} function' has been removed. See release notes for further information.`)
     }
 
     // wrap the deprecated function to warn user
-    const warn = warnOnce(`${removedName} function`)
+    const warn = warnOnce(`${fn.name} function`)
     return function (this: any) {
       warn()
       fn.apply(this, arguments)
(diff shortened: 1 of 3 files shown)


lib/common/api/deprecate.ts
```diff
@@ -34,9 +34,27 @@ const deprecate: ElectronInternal.DeprecationUtil = {
     }
   },
 
+  // remove a function with no replacement
+  removeFunction: (fn, removedName) => {
+    // if the function has already been removed, warn about it
+    if (!fn) {
+      deprecate.log(`Unable to remove function '${removedName}' from an object that lacks it.`)
```
codebytere: If the function is gone, its deprecation handling wrapper will also have been removed; we can throw an error here and bail early if the function is somehow invalid but there's no need to return a function i don't think.

DeerMichel: Yes, should only affect us (and as soon as everything is typescript, it won't compile anyway). I'll throw an error...

lib/common/api/deprecate.ts
```diff
@@ -34,9 +34,25 @@ const deprecate: ElectronInternal.DeprecationUtil = {
...
```
codebytere: We can remove this:
```
deprecate.log(`Unable to remove function '${removedName}' from an object that lacks it.`)
```
since it wouldn't apply to this.

lib/common/api/deprecate.ts
```diff
@@ -34,9 +34,25 @@ const deprecate: ElectronInternal.DeprecationUtil = {
...
+      throw Error(`'${removedName} function' has been removed. See release notes for further information.`)
```
codebytere: should be:
```
throw new Error(`${removeName} is invalid or does not exist.`)
```

lib/common/api/deprecate.ts
DeerMichel: But we keep this notice on `removeProperty`?

DeerMichel:
Sry lost track of this one. Should be fine now 💯

Issue #19009, opened 2019-06-27
systemPreferences.isDarkMode always returns false on macOS 10.15 Catalina


alicerunsonfedora:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.6
* **Operating System:**
  * macOS 10.15 Developer Beta 2 (19A487l)
* **Last Known Working Electron version:**
  * 5.0.x (worked before Catalina)

### Expected Behavior
`systemPreferences.isDarkMode` on macOS 10.15 Catalina should report whether it's in dark mode or not, even when set to Auto Mode (based on time of day). Electron 5.0.6 fixed some of the issues from #18913 , but isn't working correctly.

### Actual Behavior
`systemPreferences.isDarkMode` is always reporting to be false.

### To Reproduce
Repository: https://github.com/hyperspacedev/hyperspace

```sh
$ git clone https://github.com/hyperspacedev/hyperspace -b catalina-vibrancy
$ npm install
$ npm run electrify
```

### Screenshots
<img width="1440" alt="Screen Shot 2019-06-27 at 08 23 18" src="https://user-images.githubusercontent.com/13445064/60265918-e7ae3600-98b4-11e9-8519-6a114ff52790.png">
Dark mode reports to be `false` here and is expected because it's in light mode.

<img width="1440" alt="Screen Shot 2019-06-27 at 08 23 33" src="https://user-images.githubusercontent.com/13445064/60265936-f5fc5200-98b4-11e9-98f7-caba828f9e2a.png">
Dark mode reports to be `false` here as well since, at the time of Auto Mode, it shouldn't be in dark mode.

<img width="1440" alt="Screen Shot 2019-06-27 at 08 23 25" src="https://user-images.githubusercontent.com/13445064/60265976-0b717c00-98b5-11e9-8d5a-6a7b7ee50c58.png">
However, dark mode reports to be `false` here, even though that isn't true.

### Additional Information
I speculate that the name of the dark appearance might have changed.

MarshallOfSound:
@codebytere is working on a fix for a related issue which I believe is fundamentally the same cause 👍

[...]

Issue #19006, opened 2019-06-27
Tray icon in macOS dark mode isn’t displayed at full opacity


rdev:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.5
* **Operating System:**
  * macOS Mojave 10.14.5
* **Last Known Working Electron version:**
  * 5.0.4

### Expected Behavior
Tray icon using a template image is white at full opacity in dark mode

### Actual Behavior
Icon is about 50% transparent in dark mode, unless clicked. When active, it’s at full opacity

### To Reproduce
Use any Electron version with #18666 fix, create a tray icon using a template image

### Screenshots

![image (5)](https://user-images.githubusercontent.com/8418866/60260586-1bfd1300-9904-11e9-80b4-3d0c3c25781b.png)

codebytere:
Odd, i can reproduce this in 5.0.6 but not in latest master 🤔

giddens9527:
@codebytere I was on electron 5.0.6 and using the second approach in the comment. The tray icon was still not displayed at full opacity in dark mode.

brunolemos:
Also seeing this at 5.0.6. And I noticed the Title is gray. 

Any patch I can apply locally?

___


#### Template tray image + No title

| State            | Dark Mode | Light Mode |
|-------------|------------|------------|
| Normal        | ❌![image](https://user-images.githubusercontent.com/619186/60543505-d8801b80-9cec-11e9-9fb9-aa1919743972.png) * | ✅ ![image](https://user-images.githubusercontent.com/619186/60543625-21d06b00-9ced-11e9-82c0-8fb4e956d4aa.png) |
| Highlighted | ✅ ![image](https://user-images.githubusercontent.com/619186/60543481-cd2cf000-9cec-11e9-881c-9cfe9f81e48b.png) | ✅ ![image](https://user-images.githubusercontent.com/619186/60543640-29900f80-9ced-11e9-8f21-9301040c9efa.png) |

#### Normal tray image + Title


| State            | Dark Mode | Light Mode |
|-------------|------------|------------|
[...]

PR #19013, opened 2019-06-27
fix: dark mode tray opacity on Mojave

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/19009
Closes https://github.com/electron/electron/issues/19006.

Fixes an issue which is fixed in `master` as a result of this [Chromium CL](https://chromium-review.googlesource.com/c/chromium/src/+/1240172).

Previously Chromium forward-declared `NSAppearanceNameDarkAqua` name as an `NSString*`, so our updated functions in https://github.com/electron/electron/pull/18666 would never believe themselves to be in dark mode.

This fixes that issue by removing Chromium's older forward declarations and adding them in ourselves.

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where the Tray icon would never believe itself to be in dark mode.

### DIFF
--- a/atom/browser/api/atom_api_system_preferences_mac.mm
+++ b/atom/browser/api/atom_api_system_preferences_mac.mm
@@ -625,7 +625,7 @@ void SystemPreferences::RemoveUserDefault(const std::string& name) {
 }
 
 bool SystemPreferences::IsDarkMode() {
-  if (@available(macOS 10.14, *)) {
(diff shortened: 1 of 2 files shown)


MarshallOfSound:
@codebytere I assume you're not running catalina, I'd like to verify this on a catalina machine before we declare it fixed.  I can test this afternoon if no one else has a catalina machine available

PR #19014, opened 2019-06-27
fix: fire close event upon closing modal BrowserWindow in macOS

### LINKED ISSUE #10674: close event not fired on macOS if the window is opened as a modal
* Electron version: 1.6.10
* Operating system: macOS Sierra 10.12.6

### Expected behavior

The `close` event should be fired when the modal window is closed on Mac. It is fired on Windows.

### Actual behavior

The `close` event is not fired.

### How to reproduce

I have created a repro at this [repository](https://github.com/rongduan-zhu/electron-quick-start).

Just run `main.js` using electron.

The code essentially creates another modal window with the parent set to the main browser window. Created a `setTimeout` which closes the modal window after 3 seconds.

The modal window also listens to the `close` event and logs a message when the event is fired.

On Mac, the event listener callback is never called (nothing shows up in terminal). On Windows, the event listener callback is called (message shows up in terminal).

erickzhao:
#### Description of Change
Fixes #10674. _This bug was originally reported for Electron 1.x.x, but still occurs on `master`. See https://github.com/electron/electron/issues/10674#issuecomment-506382985 for screenshot and Fiddle link._

Adds a small function call to `windowShouldClose(_:)` to emit the `close` event upon closing a modal window.

##### Context

For modals on macOS, we use [`NSWindow close()`](https://developer.apple.com/documentation/appkit/nswindow/1419662-close) instead of [`NSWindow performClose(_:)`](https://developer.apple.com/documentation/appkit/nswindow/1419288-performclose) to close the modal window because the latter doesn't get along with the Sheet implementation of modal windows.

According to the docs:
> The close method differs in two important ways from the `performClose(_:)` method:
> 
> * It does not attempt to send a `windowShouldClose(_:)` message to the window or its delegate.
> 
> * It does not simulate the user clicking the close button by momentarily highlighting the button.

The `windowShouldClose(_:)` function in `atom_ns_window_delegate.mm` eventually triggers an `Emit("close")`.

##### Other information
I added the call manually inside the control flow right before `[window_ close_]`. I also tried adding it into the `CloseImmediately()` function first (see first commit in this branch), but that didn't work.

##### Pings
cc @codebytere @DeerMichel (and @zcbenz who worked on the original lines of code)

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed bug where the `close` event would not emit upon closing modal window on macOS.

### DIFF
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -511,6 +511,8 @@ void NativeWindowMac::CloseImmediately() {
     [NSEvent removeMonitor:wheel_event_monitor_];
     wheel_event_monitor_ = nil;
   }
+  // Manually emit close event (not triggered from close fn)
+  [window_delegate_ windowShouldClose:window_];
 
   // Retain the child window before closing it. If the last reference to the
   // NSWindow goes away inside -[NSWindow close], then bad stuff can happen.
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -492,6 +492,8 @@ void NativeWindowMac::Close() {
   // When this is a sheet showing, performClose won't work.
   if (is_modal() && parent() && IsVisible()) {
     [parent()->GetNativeWindow().GetNativeNSWindow() endSheet:window_];
+    // Manually emit close event (not triggered from close fn)
+    [window_delegate_ windowShouldClose:window_];
     CloseImmediately();
     return;
   }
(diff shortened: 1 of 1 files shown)


shell/browser/native_window_mac.mm
```diff
@@ -492,6 +492,8 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
   // When this is a sheet showing, performClose won't work.
   if (is_modal() && parent() && IsVisible()) {
     [parent()->GetNativeWindow().GetNativeNSWindow() endSheet:window_];
+    // Manually emit close event (not triggered from close fn)
+    [window_delegate_ windowShouldClose:window_];
```
zcbenz: Using `NotifyWindowCloseButtonClicked()` directly is probably better, more code may be added in `windowShouldClose` in future and it would then cause side effects here.

PR #19016, opened 2019-06-27
fix: delay emitting screen events by a tick to avoid re-entrancy crash

MarshallOfSound:
If user code created / destroyed windows inside a screen event handler it could cause re-entrancy on the `WndProc` handler for screen related events causing memory corruption. If we delay the events going to user we won't re-entry and everything will be fine.

Notes: Fixed crash when creating certain types of windows in the `display-*` events

### DIFF
--- a/shell/browser/api/atom_api_screen.cc
+++ b/shell/browser/api/atom_api_screen.cc
@@ -51,6 +51,19 @@ std::vector<std::string> MetricsToArray(uint32_t metrics) {
   return array;
 }
 
+void DelayEmit(Screen* screen,
+               const base::StringPiece& name,
+               const display::Display& display) {
+  screen->Emit(name, display);
+}
+
+void DelayEmitWithMetrics(Screen* screen,
+                          const base::StringPiece& name,
+                          const display::Display& display,
+                          const std::vector<std::string>& metrics) {
+  screen->Emit(name, display, metrics);
+}
+
 }  // namespace
 
 Screen::Screen(v8::Isolate* isolate, display::Screen* screen)
@@ -100,16 +113,23 @@ static gfx::Rect DIPToScreenRect(electron::NativeWindow* window,
 #endif
 
 void Screen::OnDisplayAdded(const display::Display& new_display) {
-  Emit("display-added", new_display);
+  base::ThreadTaskRunnerHandle::Get()->PostNonNestableTask(
+      FROM_HERE, base::Bind(&DelayEmit, base::Unretained(this), "display-added",
+                            new_display));
 }
 
 void Screen::OnDisplayRemoved(const display::Display& old_display) {
-  Emit("display-removed", old_display);
+  base::ThreadTaskRunnerHandle::Get()->PostNonNestableTask(
+      FROM_HERE, base::Bind(&DelayEmit, base::Unretained(this),
+                            "display-removed", old_display));
 }
 
 void Screen::OnDisplayMetricsChanged(const display::Display& display,
                                      uint32_t changed_metrics) {
-  Emit("display-metrics-changed", display, MetricsToArray(changed_metrics));
+  base::ThreadTaskRunnerHandle::Get()->PostNonNestableTask(
+      FROM_HERE, base::Bind(&DelayEmitWithMetrics, base::Unretained(this),
+                            "display-metrics-changed", display,
+                            MetricsToArray(changed_metrics)));
 }
 
 // static


patches/chromium/fix_modify_notifydisplayschanged_to_copy_vectors_on_call.patch
```diff
@@ -0,0 +1,45 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <sattard@slack-corp.com>
+Date: Thu, 27 Jun 2019 10:58:39 -0700
+Subject: fix: modify NotifyDisplaysChanged to copy vectors on call
+
+As this method is called directly from a WndProc it is possible for it
+to be re-entried especially for some embedder use cases.  When it is
```
nornagon:
```suggestion
to be re-entered especially for some embedder use cases.  When it is
```

patches/chromium/fix_modify_notifydisplayschanged_to_copy_vectors_on_call.patch
```diff
@@ -0,0 +1,45 @@
...
+re-entried the old reference has been modified which results in memory
+corruption and eventually leads to one of the iterators inside this
+method crashing the application due to an access violation.
+
+By copying the vectors instead of following the reference we can
+guaruntee that the vector will be safe and valid for the lifetime of the
```
nornagon:
```suggestion
guarantee that the vector will be safe and valid for the lifetime of the
```

MarshallOfSound:
@nornagon I tried that locally but it didn't consistently fix it, this did.  I also think we can't guarantee that our code is the only thing that could trigger a re-entry.  We should ensure this method can handle re-entry no matter what causes it,

shell/browser/api/atom_api_screen.cc
```diff
@@ -99,17 +99,38 @@ static gfx::Rect DIPToScreenRect(electron::NativeWindow* window,
...
+                          const base::StringPiece& name,
+                          const display::Display& display,
+                          std::vector<std::string> metrics) {
+  screen->Emit(name, display, metrics);
+}
+
 void Screen::OnDisplayAdded(const display::Display& new_display) {
-  Emit("display-added", new_display);
+  base::PostTaskWithTraits(FROM_HERE, {content::BrowserThread::UI},
+                           base::Bind(&DelayEmit, base::Unretained(this),
```
nornagon: is it ok to use `base::Unretained` here? Is there a guarantee that `Screen` can't be deleted in between posting the task and processing the task?

shell/browser/api/atom_api_screen.cc
```diff
@@ -99,17 +99,38 @@ static gfx::Rect DIPToScreenRect(electron::NativeWindow* window,
...
+void DelayEmit(Screen* screen,
+               const base::StringPiece& name,
+               const display::Display& display) {
+  screen->Emit(name, display);
+}
+
+void DelayEmitWithMetrics(Screen* screen,
+                          const base::StringPiece& name,
+                          const display::Display& display,
+                          std::vector<std::string> metrics) {
```
nornagon: `const std::vector<std::string>&`? it should get copied by `PostTask`, no need to copy it again when the task queue calls the function.

shell/browser/api/atom_api_screen.cc
```diff
@@ -99,17 +99,38 @@ static gfx::Rect DIPToScreenRect(electron::NativeWindow* window,
...
+  screen->Emit(name, display, metrics);
+}
+
 void Screen::OnDisplayAdded(const display::Display& new_display) {
-  Emit("display-added", new_display);
+  base::PostTaskWithTraits(FROM_HERE, {content::BrowserThread::UI},
```
nornagon: you can post to the current thread like
```cpp
base::ThreadTaskRunnerHandle::Get()->PostTask(...)
```

but i guess we're guaranteed to be on the ui thread here anyway so /shrug

shell/browser/api/atom_api_screen.cc
MarshallOfSound: Yes, `Screen` is a singleton that is only destroyed during shutdown.  It won't be ripped out from under itself here afaics

nornagon: inb4 crash on shutdown when kvm switching into a remote desktop or some bs :P

seems fine tho 👍

shell/browser/api/atom_api_screen.cc
```diff
@@ -99,17 +99,38 @@ static gfx::Rect DIPToScreenRect(electron::NativeWindow* window,
...
+  base::ThreadTaskRunnerHandle::Get()->PostTask(
+      FROM_HERE, base::Bind(&DelayEmit, base::Unretained(this),
+                            "display-removed", old_display));
 }
 
 void Screen::OnDisplayMetricsChanged(const display::Display& display,
                                      uint32_t changed_metrics) {
-  Emit("display-metrics-changed", display, MetricsToArray(changed_metrics));
+  base::PostTaskWithTraits(
+      FROM_HERE, {content::BrowserThread::UI},
```
nornagon: make this one the same as the other 2 for consistency?

MarshallOfSound: Huh, oh darn, didn't `git add` the final change :/

shell/browser/api/atom_api_screen.cc
```diff
@@ -99,17 +99,37 @@ static gfx::Rect DIPToScreenRect(electron::NativeWindow* window,
...
 void Screen::OnDisplayAdded(const display::Display& new_display) {
-  Emit("display-added", new_display);
+  base::ThreadTaskRunnerHandle::Get()->PostTask(
+      FROM_HERE, base::Bind(&DelayEmit, base::Unretained(this), "display-added",
+                            new_display));
 }
 
 void Screen::OnDisplayRemoved(const display::Display& old_display) {
-  Emit("display-removed", old_display);
+  base::ThreadTaskRunnerHandle::Get()->PostTask(
```
deepak1556: Can we make this `PostNonNestableTask` so that the ordering is guaranteed wrt `display-added`. check `sequneced_task_runner.h`

PR #19019, opened 2019-06-27
[wip] fix: `getUserMedia` promise resolves before user grants permissions

### LINKED ISSUE #19017: `navigator.mediaDevices.getUserMedia` promise resolves before user grants permissions -- intended usage?
### Issue Details

* **Electron Version:**
  * 5.0.5
* **Operating System:**
  * macOS 10.14.5

### Expected Behavior
`navigator.mediaDevices.getUserMedia` promise doesn't resolve until user allows or denies permission, according to https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia

### Actual Behavior
`navigator.mediaDevices.getUserMedia` resolves before user click "Don't Allow" or "OK" on the system dialog. 

### To Reproduce
- Run `tccutil reset Camera; tccutil reset Microphone` to clear permissions for the macOS terminal.
- Run electron-quick-start (https://github.com/electron/electron-quick-start)
- Open Developer tools
- In the console, run:
```
navigator.mediaDevices.getUserMedia({audio: true, video: true}).then(e => console.log(e));
```
- Observe that the promise resolves with a `MediaStream` before you click "Don't Allow" or "OK" on the dialog that pops up.

### Screenshots
Screen Shot 2019-06-27 at 12 32 03 PM

### Additional Info
If I understand correctly the promise resolves before the dialog closes in order for the dialog to not block the renderer process? However, the media stream returned doesn't do anything until the user actually accepts the permission. How can we know if the user accepted permission, in order to not use the returned media until it has a valid stream?

DeerMichel:
#### Description of Change
Fixes #19017. Or at least tries to fix it. The expected behavior is that the promise won't be resolved until the user grants or declines permission for media access on macOS. I backported [this](https://chromium-review.googlesource.com/c/chromium/src/+/1511401) Chromium fix to v5 but it didn't change anything. The bug seems to be present in v6 and master as well, although Chromium says the [original issue](https://bugs.chromium.org/p/chromium/issues/detail?id=885184) is fixed. Can someone help me out here?

cc @codebytere @erickzhao @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: TODO

### DIFF
--- a/patches/common/chromium/.patches
+++ b/patches/common/chromium/.patches
@@ -77,3 +77,4 @@ viz_osr.patch
 video_capturer_dirty_rect.patch
 unsandboxed_ppapi_processes_skip_zygote.patch
 autofill_size_calculation.patch
+block_getusermedia_on_system_permissions.patch
--- a/patches/common/chromium/block_getusermedia_on_system_permissions.patch
+++ b/patches/common/chromium/block_getusermedia_on_system_permissions.patch
@@ -1,0 +1,431 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Henrik Grunell <grunell@chromium.org>
+Date: Mon, 11 Mar 2019 15:22:16 +0000
+Subject: Block getUserMedia on system permissions.
+
+* New Mac functions for getting and requesting system permissions.
+* Change the request queue to a map with unique IDs for the requests since the order isn't guaranteed anymore.
+* Wait (async) for response from system permission request.
+
+Bug: 885184
+Change-Id: I17b046eaab51c5dd6dd57383fa23d3e7d567f809
+Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/1511401
+Auto-Submit: Henrik Grunell <grunell@chromium.org>
+Commit-Queue: Tommi <tommi@chromium.org>
+Reviewed-by: Tommi <tommi@chromium.org>
+Reviewed-by: Guido Urdaneta <guidou@chromium.org>
+Cr-Commit-Position: refs/heads/master@{#639496}
+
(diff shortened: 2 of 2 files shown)


deepak1556:
@DeerMichel we don't rely on those chromium classes for checking media access permissions , the way forward to fix this would be to tie the media access checks to `WebContentsPermissionHelper` which handles the response for `navigator.getUserMedia` https://github.com/electron/electron/blob/master/shell/browser/web_contents_permission_helper.cc#L99 , the code will end up somewhat similar to the above chromium CL :)

deepak1556:
Also can you check if the following workaround works at the moment. we have `SystemPreferences::AskForMediaAccess` api, and we have `Session.setPermissionRequestHandler` , so ideally users should be able to respond with system indicated value in the permission handler combining both in JS.

DeerMichel:
So I tested `systemPreferences.askForMediaAccess()` and the returned promise behaves just fine. I'll try the `session.setPermissionRequestHandler()` func tomorrow. If they both work as expected, I guess we don't have to change anything, right? He should just use the Electron API.

MarshallOfSound:
>  If they both work as expected, I guess we don't have to change anything, right? He should just use the Electron API.

`navigator.getUserMedia` needs to work not just the electron provided APIs

DeerMichel:
I'm a bit lost here. First, the js workaround from @deepak1556 works. Regarding Chromium, it should be fixed since v74, therefore in Electron v6 & v7. However, it's a strange behavior, since there are two cascaded dialogs shown when you request permissions. The promise resolves after the first one and doesnt wait for the second (actual native) one. Even on Chrome itself it behaves likes that - maybe still an open issue over there?

DeerMichel:
Closing this for now since it doesn't fix the problem and I have no idea how to fix it differently.

Issue #19020, opened 2019-06-27
@types/node outdated


pushkin-:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** 5.0.6

[docs](https://electronjs.org/releases/stable?version=5&page=2#5.0.0) say that Electron 5.x uses Node 12.0.0.

[package.json](https://github.com/electron/electron/blob/master/package.json#L15) uses @types/node 10.12.

PR #19025, opened 2019-06-28
chore: update @types/node and @primer/octicons

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/19020.

Updates our `@types/node` version to match our Node.js version, as well as updating our octicons package for both major version and namespace (`octicons` => `@primer/octicons`).

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


package.json
```diff
@@ -7,19 +7,20 @@
     "@electron/docs-parser": "^0.2.2",
     "@electron/typescript-definitions": "^8.3.1",
     "@octokit/rest": "^16.3.2",
+    "@primer/octicons": "^9.1.1",
```
MarshallOfSound: I feel as though this will break a thing / require a change in the default app?

codebytere: i considered this also, but couldn't find one and nothing seemed broken 🤔

codebytere: yep, pushed!

MarshallOfSound:
Does this still work?

https://github.com/electron/electron/blob/master/default_app/index.ts#L3

I can't remember how I set that up 😆

codebytere:
Screen Shot 2019-06-27 at 8 57 07 PM
Looks shipshape to me!

codebytere:
@nornagon i actually don't think we need to bp this since `@types/node` doesn't exist in 5 :)

pushkin-:
@codebytere It doesn't? My dependency tree for 5.0.6 shows @types/node@10.14.10

codebytere:
i could be mistaken! let me check again :)

pushkin-:
@codebytere Do you know if this will make it into 5.0.7?

codebytere:
@pushkin- no: see [here](https://github.com/electron/electron/blob/5-0-x/package.json)

Issue #19032, opened 2019-06-28
Implement BrowserWindow.setFocusable on Linux


CapOM:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description
Currently BrowserWindow.setFocusable is only implemented on Windows. But not on Mac despite the focusable option being available on all platforms.

### Proposed Solution
I have a patch that implements it on Mac. I can take a look to Linux later.

### Alternatives Considered
None

codebytere:
Hi, while we appreciate this enhancement request it's not something that our team will be able to implement in the near future, and since we use our issues tab as the team's backlog, we prefer to only retain issues that are on our immediate roadmap. If you'd like to make a PR for this request we would welcome that but in the meantime I'm going to go ahead and close this.

PR #19033, opened 2019-06-28
feat: Implement BrowserWindow.setFocusable on macOS

CapOM:
#### Description of Change

It was not implemented on Mac despite being available as a constructor
option. Implementation already exists on Windows. Linux case can be
separately.

https://github.com/electron/electron/issues/19032

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Notes: Implemented BrowserWindow.setFocusable on macOS.

### DIFF
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -1075,6 +1075,13 @@ void NativeWindowMac::SetContentProtection(bool enable) {
       setSharingType:enable ? NSWindowSharingNone : NSWindowSharingReadOnly];
 }
 
+void NativeWindowMac::SetFocusable(bool focusable) {
+  // No known way to unfocus the window if it had the focus. Here we do not
+  // want to call Focus(false) because it moves the window to the back, i.e.
+  // at the bottom in term of z-order.
+  [window_ setDisableKeyOrMainWindow:!focusable];
+}
+
 void NativeWindowMac::AddBrowserView(NativeBrowserView* view) {
   [CATransaction begin];
   [CATransaction setDisableActions:YES];
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -1552,11 +1552,13 @@ Prevents the window contents from being captured by other apps.
 On macOS it sets the NSWindow's sharingType to NSWindowSharingNone.
 On Windows it calls SetWindowDisplayAffinity with `WDA_MONITOR`.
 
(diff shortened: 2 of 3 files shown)


CapOM:
Hi @codebytere , please take a look, thx!

shell/browser/native_window_mac.mm
```diff
@@ -1073,6 +1073,10 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
       setSharingType:enable ? NSWindowSharingNone : NSWindowSharingReadOnly];
 }
 
+void NativeWindowMac::SetFocusable(bool focusable) {
+  [window_ setDisableKeyOrMainWindow:!focusable];
```
erickzhao: Looking at the Windows implementation, maybe we should also immediately remove focus from the window right after calling `w.setFocusable(false)`?

See the `Focus(false)` call below:
https://github.com/electron/electron/blob/1d8e16bc6ecb5a2700e75e8466f17da9e3391628/shell/browser/native_window_views.cc#L918-L925

CapOM:
Hi, all remarks are addressed, please take a look, thx!

shell/browser/native_window_mac.mm
```diff
@@ -1075,6 +1075,12 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
...
+  // Remove focus if the window had it.
+  Focus(false);
```
erickzhao: This should only apply if `focusable` is false.

CapOM: I started to debug and it works if you put the setFocusable(false) in ready-to-show callback, show=false initially, and then call window.show(). 
Note that setFocusable(true) works in all cases. Just above limitation for the `false` case.

I am not sure this is fixable, also I noticed that NativeWindowMac::Focus(false) actually orders the window to the back instead of just removing the focus, so this existing function is already problematic. Any reason for that ? Thx!

erickzhao: I just snooped around the [NSWindow documentation](https://developer.apple.com/documentation/appkit/nswindow/) and didn't find much regarding unfocusing windows. If this isn't possible via the AppKit APIs, should we simply document this OS-specific quirk?

cc @codebytere @zcbenz

zcbenz: I don't think there is a way to unfocus window on macOS.

PR #19037, opened 2019-06-28
fix: silent printing mode

codebytere:
Backport of https://github.com/electron/electron/pull/18979.

See that PR for more details.

Notes: Fixed silent printing across platforms.


ryzokuken:
@codebytere do you think we should prefix backports with a descriptive label like Node? This was super confusing in my notifications since I had three PRs with the same title (original and two backports).

codebytere:
They should all be labeled, which differentiates them in the UI :)

sarathjasrin:
This is not fixed still Version : 6.0.0

PR #19064, opened 2019-07-01
fix: post_task include error

codebytere:
Corrects incorrect `post_task` [changed in rename](https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/CoNmeX488-s) from `task_scheduler` dir to `task` dir in later Chromium versions.

cc @ckerr 

Notes: none

### DIFF
--- a/atom/browser/ui/tray_icon_cocoa.mm
+++ b/atom/browser/ui/tray_icon_cocoa.mm
@@ -10,8 +10,7 @@
 #include "base/mac/sdk_forward_declarations.h"
 #include "base/message_loop/message_loop.h"
 #include "base/strings/sys_string_conversions.h"
-#include "base/task/post_task.h"
-#include "content/public/browser/browser_task_traits.h"
+#include "base/task_scheduler/post_task.h"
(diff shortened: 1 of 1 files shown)


atom/browser/ui/tray_icon_cocoa.mm
```diff
@@ -10,7 +10,7 @@
 #include "base/mac/sdk_forward_declarations.h"
 #include "base/message_loop/message_loop.h"
 #include "base/strings/sys_string_conversions.h"
-#include "base/task/post_task.h"
+#include "base/task_scheduler/post_task.h"
 #include "content/public/browser/browser_task_traits.h"
```
ckerr:
```
../../electron/atom/browser/ui/tray_icon_cocoa.mm:14:10: fatal error: 'content/public/browser/browser_task_traits.h' file not found
#include "content/public/browser/browser_task_traits.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #19073, opened 2019-07-02
feat: expose new vibrancy types

codebytere:
#### Description of Change

In macOS Mojave, Apple added a host of new [`NSVisualEffectMaterials`](https://developer.apple.com/documentation/appkit/nsvisualeffectmaterial?language=objc). This PR enabled developers to use those new types by adding them to options for new `BrowserWindow`s and when invoking the `setVibrancy` API on `BrowserWindow` instances.

cc @MarshallOfSound @nornagon @miniak 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added support for NSVisualEffectMaterials vibrancy types added in macOS Mojave.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -234,9 +234,8 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
     Windows, which adds standard window frame. Setting it to `false` will remove
     window shadow and window animations. Default is `true`.
   * `vibrancy` String (optional) - Add a type of vibrancy effect to the window, only on
-    macOS. Can be `appearance-based`, `light`, `dark`, `titlebar`, `selection`, `menu`,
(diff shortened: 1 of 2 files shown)


codebytere:
@ckerr potentially! and no, this is done so that we don't need to forward-declare everything.

ckerr:
Keeping it working on older macOSes is more important than slightly tidier code.

codebytere:
@ckerr i'm a little unclear on which solution you're advocating for; do you mean you think we _should_ forward declare these enum values?

codebytere:
```
app module app.requestSingleInstanceLock passes arguments to the second-instance event
```
on Windows and 
```
chromium feature command line switches --remote-debugging-port switch should display the discovery page - should display the discovery page
```
on Linux

are unrelated and should not block merge.

Issue #16443, opened 2019-01-18
Reuse //gin in codebase where possible


deepak1556:
This is an effort to refactor `//electron/native_mate`, it was originally created as a fork of `//gin` to add capabilities not supported by chromium https://github.com/electron/electron/pull/231. But moving forward we can extend `//gin` instead with ease, now that the codebase is unified.

- [x] [mate::Arguments](https://github.com/electron/electron/blob/master/native_mate/native_mate/arguments.h) => `gin::Arguments`
  - This has a dependency on the converters which must be transitioned first
  - Missing interfaces that should be extended
     - [ ] `mate::Arguments::GetThis`
- [ ] Remove [native_mate/constructor.h](https://github.com/electron/electron/blob/master/native_mate/native_mate/constructor.h)
   - It only contains dead code
- [x] [mate::Dictionary](https://github.com/electron/electron/blob/master/native_mate/native_mate/dictionary.h) => `gin::Dictionary`
   - This has a dependency on the converters which must be transitioned first
   - Missing interfaces that should be extended
      - [ ] `mate::Dictionary::SetReadOnly`
      - [ ] `mate::Dictionary::SetMethod`
      - [ ] `mate::Dictionary::Delete`
      - [ ] `mate::Dictionary::IsEmpty`
      - [ ] `mate::Dictionary::GetHandle`
- [ ] remove [native_mate/function_template.{cc|h}](https://github.com/electron/electron/blob/master/native_mate/native_mate/function_template.h) for `//gin/function_template.{cc|h}`
   - This has a dependency on `mate::Arguments`
   - Missing interfaces that should be added
      - [ ] `mate::internal::Destroyable` , which is necessary to check for c++ types destroyed by public `destroy()` api
- [x] [mate::Handle](https://github.com/electron/electron/blob/master/native_mate/native_mate/handle.h) => `gin::Handle`
- [ ] [mate::ObjectTemplateBuilder](https://github.com/electron/electron/blob/master/native_mate/native_mate/object_template_builder.h) => `gin::ObjectTemplateBuilder`
   - Missing interfaces that should be extended
      - [ ] `mate::ObjectTemplateBuilder::MakeDestroyable`
- [ ] [mate::WrappableBase](https://github.com/electron/electron/blob/master/native_mate/native_mate/wrappable_base.h) => `gin::WrappableBase`
   - Missing interfaces that should be extended
      - [ ] `mate::WrappableBase::AfterInit`
[...]

PR #19088, opened 2019-07-02
[wip] refactor: migrate WebRequest to //gin

DeerMichel:
#### Description of Change
An adventurous start into #16443 with the _simple_ goal of removing `native_mate` in favor of Chromium's `gin`. This PR covers a pilot migration of `WebRequest` with all involved converters and classes around it. Because I don't want to cram the whole migration into one PR, there are some duplicated files suffixed with `_gin` which will eventually (after dropping native_mate) replace their unsuffixed friends.

For now, I patched missing functionality into gin. Later, we should eliminate these patches by either upstreaming the patches or rewrite some of our usages.

cc @deepak1556 @codebytere @zcbenz

#### Todos
- check all my `TODO(deermichel)` marks for correctness (and clarify mistakes to me 🦊)
- when closing Electron, it prints a decent crash log
- when calling any method of WebRequest, it throws following exception:
```
Uncaught Exception:
TypeError: Illegal invocation: Function must be called on an object of type TODO:PROBLEM:FLAG
    at App.<anonymous> (/private/var/folders/tg/v1p6mpc5389ddclnmpys6gw00000gp/T/tmp-28725mprt94ureVz3/main.js:15:37)
    at App.emit (events.js:199:15)
```

#### Patches
- Arguments: add `GetThis()`
- Dictionary: add `SetHidden()`, `GetHidden()`
- ObjectTemplateBuilder: add constructor with template parameter (prolly not working correctly)
- Wrappable: add `Init()`, `InitWith()`, `AfterInit()`, `GetConstructor()`

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: TODO


zcbenz:
I wonder if we should add a custom class implementing `Wrappable` instead of patching the one in `gin`, I don't think Chromium would accept the `gin_wrappable_init_methods.patch`.

DeerMichel:
Is there a way to access the private `wrapper_` property of `WrappableBase` in a subclass (e.g. for the `InitWith()` method)?

zcbenz:
> Is there a way to access the private `wrapper_` property of `WrappableBase` in a subclass (e.g. for the `InitWith()` method)?

There is no way for that.

I think maintaining our own copies of `Wrappable` and `WrappableBase` (with different class names under `gin` namespace) would be easier for long-term maintenance, as `gin` still gets changes frequently and they may break our code, while the V8 API we use is stable.

DeerMichel:
Closing this as it'll never get merged.

PR #19094, opened 2019-07-03
feat: implement net module with NetworkService

zcbenz:
#### Description of Change

This PR adds NetworkService implementation for `net` module, all tests are passing with NetworkService enabled.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/shell/browser/api/atom_api_url_request_ns.cc
+++ b/shell/browser/api/atom_api_url_request_ns.cc
@@ -11,6 +11,8 @@
 #include "shell/browser/atom_browser_context.h"
 #include "shell/common/native_mate_converters/gurl_converter.h"
 
+#include "shell/common/node_includes.h"
+
 namespace electron {
 
 namespace api {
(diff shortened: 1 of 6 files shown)


shell/browser/api/atom_api_url_request_ns.cc
```diff
@@ -0,0 +1,536 @@
...
+    // data event after request cancel/error/close.
+    if (!(request_state_ & STATE_ERROR) && !(response_state_ & STATE_ERROR)) {
+      response_state_ |= STATE_FINISHED;
+      Emit("end");
+    }
+  } else {  // failed
+    // Only emit error when there is no previous failure.
+    // This is to align with the behavior of non-NetworkService implementation.
+    if (!(request_state_ & STATE_FAILED))
+      EmitRequestError(net::ErrorToString(loader_->NetError()));
```
deepak1556: Should `response_state_` be set to `STATE_FAILED` in the `else` block ? At the moment I see that `response_state_` only toggles between `STATE_STARTED` and `STATE_FINISHED` , there is no indication of `STATE_ERROR` but we do check for this in `Cancel` and other places.

zcbenz: Nice catch. The `SimpleURLLoader` does not distinguish between response errors and request errors, so I have pushed a change to emit error depending on current state.

deepak1556:
Can you rebase on master before merging and fix the compile error that will be triggered because of api change from latest chromium https://chromium-review.googlesource.com/c/chromium/src/+/1660393 .

PR #19100, opened 2019-07-03
test: disable IAP tests that hit Apple servers

nornagon:
#### Description of Change
These tests started failing on an unrelated change, and they pass locally, leading us to believe that Apple started blocking requests to the StoreKit servers from AWS. We shouldn't be hitting remote servers in CI anyway, so I'm disabling these tests for now.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/spec/api-in-app-purchase-spec.js
+++ b/spec/api-in-app-purchase-spec.js
@@ -38,17 +38,22 @@ describe('inAppPurchase module', function () {
     expect(correctUrlEnd).to.be.true()
   })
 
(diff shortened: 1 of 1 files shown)


codebytere:
@nornagon do we want to backport this? i haven't seen it failing on older lines, i don't think 🤔

nornagon:
not really worth backporting, i don't think. it's not about the version on which it fails, it'll fail regardless of the version if there's any error communicating with Apple's servers.

PR #19107, opened 2019-07-04
Fix: tests and api with network service enabled

deepak1556:
#### Description of Change

Noticed some tests for previously converted modules were failing when testing https://github.com/electron/electron/pull/19094 . This PR aims to fix that, with visible ci results.

/cc @zcbenz @nornagon 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes


.circleci/config.yml
```diff
@@ -886,7 +886,7 @@ steps-tests: &steps-tests
         command: |
           cd src
           export ELECTRON_OUT_DIR=Default
-          (cd electron && node script/yarn test -- --ci --enable-logging)
+          (cd electron && node script/yarn test -- --ci --enable-logging --enable-features=NetworkServiceNotSupported)
```
deepak1556: The feature name is weird because the default values are flipped in upstream.

spec-main/api-net-spec.ts
```diff
@@ -628,7 +628,7 @@ describe('net module', () => {
       })
     })
 
-    describe('webRequest', () => {
+    xdescribe('webRequest', () => {
```
zcbenz: I added a `net.isNetworkServiceEnabled` JS flag in https://github.com/electron/electron/pull/19094, which can be used to conditionally skip the tests.

deepak1556: Yup plan to use that once merged 👍

.circleci/config.yml
nornagon: Are we planning to run the tests both with & without the network service? If we're only going to run tests with network service, let's just enable the feature.

deepak1556: Yeah after working a bit on this PR, I am inclined to do that. I will adjust the PR to enable the feature by default.

spec-main/api-protocol-spec.ts
```diff
@@ -223,7 +223,7 @@ describe('protocol module', () => {
       expect(r.headers).to.include('x-great-header: sogreat')
     })
 
-    it('throws an error when custom headers are invalid', (done) => {
+    it.skip('throws an error when custom headers are invalid', (done) => {
```
deepak1556: We are now skipping invalid headers https://github.com/electron/electron/blob/master/shell/browser/net/atom_url_loader_factory.cc#L112-L122, this spec was added recently https://github.com/electron/electron/pull/18854. I am fine with restoring the behavior but I feel that rather than throwing exception , sanitizing should be more of ignoring invalid values.

@zcbenz wdyt ?

zcbenz: For protocol APIs, we usually send network errors when there is invalid input, we didn't throw exceptions under any case.

PR #19110, opened 2019-07-04
feat: add object to subscribe notification callback

### LINKED ISSUE #19097: systemPreferences.subscribeNotification should pass notification object to the callback
### Preflight Checklist

### Problem Description
Today, it's not possible to receive contextual data from notifications originating from sandboxed macOS apps.

This is because Electron's `systemPreferences.subscribeNotification` accepts a callback that receives the `event` and `userInfo` properties from the macOS notification but **not** the `object` property. This is problematic because sandboxed apps cannot use the `userInfo` property, and can only transmit contextual data via the `object` of a notification. [Apple's docs](https://developer.apple.com/documentation/foundation/nsdistributednotificationcenter/1418360-postnotificationname?language=objc) clearly note this restriction.

As such, Electron cannot read contextual data sent from notifications that originate from sandboxed apps. So, for example, a Finder extension can notify an Electron app that *some* file has been right-clicked, but **cannot** indicate *which file* has been right-clicked.

### Proposed Solution
I'm proposing a change similar to #5582, which enhanced `systemPreferences.subscribeNotification` to pass the `userInfo` from the macOS notification to the callback. Here, we need to further enhance `systemPreferences.subscribeNotification` to also pass the `object` from the macOS notification.

To preserve backwards-compatibility, the callback should accept the new `object` argument as its third parameter. As such, the new version of the documentation [here](https://electronjs.org/docs/api/system-preferences#systempreferencessubscribenotificationevent-callback-macos) would read:

```diff
systemPreferences.subscribeNotification(event, callback) [macOS]

* event String
* callback Function
    * event String
    * userInfo Object
+   * object String

Returns Number - The ID of this subscription

Subscribes to native notifications of macOS, callback will be called with `callback(event, userInfo)` when the corresponding event happens. The `userInfo` is an Object that contains the user information dictionary sent along with the notification.
+The `object` is a String that contains additional data sent along with the notification.
```

### Alternatives Considered
I think the only reasonable alternative is to update the documentation [here](https://electronjs.org/docs/api/system-preferences#systempreferencessubscribenotificationevent-callback-macos) to note that sandbox apps cannot send any contextual data to Electron apps.

### Additional Information
There is a precedent for this kind of change in Electron. See #5543 which was fixed under #5582.

For macOS developers running into this issue, a possible workaround is to have your sandboxed app communicate with a non-sandboxed app and then relay notifications on to your Electron app. The sandbox to non-sandbox communication must use `object` to hold contextual data, and the non-sandbox to Electron communication must use `userInfo` to hold contextual data. (However, AFAIK, using a non-sandboxed app prevents publishing your app in the Mac App Store).

louismrose:
#### Description of Change

Closes #19097.

This change exposes the `object` property on NSNotification to the callback passed to Electron's `systemPreferences.subscriptionNotification` methods.

I'm cc-ing @zcbenz on this PR as he worked on a very similar feature under #5582.

#### Checklist
- [x] `npm test` passes

#### Release Notes

Notes: Exposed the value of NSNotification.object to subscribers of notifications in systemPreferences.

### DIFF
--- a/docs/api/system-preferences.md
+++ b/docs/api/system-preferences.md
@@ -90,13 +90,14 @@ that contains the user information dictionary sent along with the notification.
 * `callback` Function
   * `event` String
   * `userInfo` Object
+  * `object` String
 
 Returns `Number` - The ID of this subscription
 
 Subscribes to native notifications of macOS, `callback` will be called with
 `callback(event, userInfo)` when the corresponding `event` happens. The
 `userInfo` is an Object that contains the user information dictionary sent
-along with the notification.
+along with the notification. The `object` is the sender of the notification.
 
 The `id` of the subscriber is returned, which can be used to unsubscribe the
 `event`.
@@ -115,6 +116,7 @@ example values of `event` are:
 * `callback` Function
   * `event` String
   * `userInfo` Object
+  * `object` String
 
 Returns `Number` - The ID of this subscription
 
@@ -127,6 +129,7 @@ This is necessary for events such as `NSUserDefaultsDidChangeNotification`.
 * `callback` Function
   * `event` String
   * `userInfo` Object
+  * `object` String
 
 Same as `subscribeNotification`, but uses `NSWorkspace.sharedWorkspace.notificationCenter`.
 This is necessary for events such as `NSWorkspaceDidActivateApplicationNotification`.
--- a/docs/api/system-preferences.md
+++ b/docs/api/system-preferences.md
@@ -97,7 +97,8 @@ Returns `Number` - The ID of this subscription
 Subscribes to native notifications of macOS, `callback` will be called with
 `callback(event, userInfo)` when the corresponding `event` happens. The
 `userInfo` is an Object that contains the user information dictionary sent
-along with the notification. The `object` is the sender of the notification.
+along with the notification. The `object` is the sender of the notification,
+and only supports `NSString` values for now.
 
 The `id` of the subscriber is returned, which can be used to unsubscribe the
 `event`.
--- a/shell/browser/api/atom_api_system_preferences_mac.mm
+++ b/shell/browser/api/atom_api_system_preferences_mac.mm
@@ -213,7 +213,7 @@ int SystemPreferences::DoSubscribeNotification(
 
                 std::string object = "";
                 if ([notification.object isKindOfClass:[NSString class]]) {
-                  object = base::SysNSStringToUTF8(notification.object));
+                  object = base::SysNSStringToUTF8(notification.object);
                 }
 
                 if (user_info) {
--- a/shell/browser/api/atom_api_system_preferences_mac.mm
+++ b/shell/browser/api/atom_api_system_preferences_mac.mm
@@ -212,11 +212,14 @@ int SystemPreferences::DoSubscribeNotification(
                     NSDictionaryToDictionaryValue(notification.userInfo);
                 if (user_info) {
                   copied_callback.Run(
-                      base::SysNSStringToUTF8(notification.name), *user_info);
+                      base::SysNSStringToUTF8(notification.name),
+                      *user_info,
+                      base::SysNSStringToUTF8(notification.object));
                 } else {
                   copied_callback.Run(
                       base::SysNSStringToUTF8(notification.name),
-                      base::DictionaryValue());
+                      base::DictionaryValue(),
+                      base::SysNSStringToUTF8(notification.object));
                 }
               }];
   return request_id;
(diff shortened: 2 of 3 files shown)


louismrose:
I'm not too sure why these builds are failing -- I don't think it's related to my changes, but I might be missing something obvious! Any tips from a contributor would be greatly appreciated? I'm happy to tackle any further changes needed here.

docs/api/system-preferences.md
```diff
@@ -90,13 +90,14 @@ that contains the user information dictionary sent along with the notification.
...
   * `userInfo` Object
+  * `object` String
 
 Returns `Number` - The ID of this subscription
 
 Subscribes to native notifications of macOS, `callback` will be called with
 `callback(event, userInfo)` when the corresponding `event` happens. The
 `userInfo` is an Object that contains the user information dictionary sent
-along with the notification.
+along with the notification. The `object` is the sender of the notification.
```
zcbenz: It seems that the `object` is only meaningful when it is a `NSString`?

I'm good with only supporting strings for now, but the documentation should explicitly document that "only `NSSString` is supported for now".

louismrose: Ahh, yes, you're right. I hadn't spotted that other types might be used.

I've updated the docs to make this limitation clear.

Thanks for the helpful feedback!

louismrose:
Thanks for the feedback @zcbenz. I've made the documentation change that you suggested.

louismrose:
I'm pretty sure that the Mac build failure is a false positive, and that my changes have not introduced a regression. 

The error from the [CI build](https://circleci.com/gh/electron/electron/255764) is:

```
Error: No valid signing identity available to run autoUpdater specs
    at Context.<anonymous> (electron/spec-main/api-autoupdater-darwin-spec.ts:26:15)
    at processImmediate (internal/timers.js:445:21)
```

And this is for the `autoUpdater behavior "before each" hook for "should have a valid code signing identity"` part of the test suite.

It seems like perhaps this is a CI setup failure, rather than anything to do with the code I've changed (which is in system preferences, and not in auto-updating).

@zcbenz -- do you know if there's anything more I need to do to get a passing build, or for this to be merged? I looked at re-running the build on Circle CI, but I don't have permission to do so. The [Electron docs](https://electronjs.org/docs/development/pull-requests#continuous-integration-testing) suggest that I could ask a releaser to re-run the build, but sadly the link to the list of releasers ((https://github.com/orgs/electron/teams/releasers/members) is a 404 for me.

zcbenz:
I have re-run the test, we can merge even with CI failing since it is obviously a flaky one.

(I want to wait for more reviews before merging this, as it is a new feature.)

louismrose:
Great, thanks for your help @zcbenz!

shell/browser/api/atom_api_system_preferences_mac.mm
```diff
@@ -212,11 +212,14 @@ AVMediaType ParseMediaType(const std::string& media_type) {
                     NSDictionaryToDictionaryValue(notification.userInfo);
                 if (user_info) {
                   copied_callback.Run(
-                      base::SysNSStringToUTF8(notification.name), *user_info);
+                      base::SysNSStringToUTF8(notification.name),
+                      *user_info,
+                      base::SysNSStringToUTF8(notification.object));
```
nornagon: `notification.object` has type `id`, which doesn't necessarily point to an `NSString`. `base::SysNSStringToUTF8` will crash if you pass it an object that isn't an `NSString`. I'm somewhat surprised that this compiles, but in any case we should definitely check that the object is actually a string, e.g.

```objc
if ([notification.object isKindOfClass:[NSString class]]) {
  // ...
}
```

MarshallOfSound:
The macOS codesign failure is due to the PR coming from a fork and is a known issue, if every other tests passes it's fine 👍

shell/browser/api/atom_api_system_preferences_mac.mm
louismrose: Thanks for the spotting this @nornagon, and for including sample code in your feedback. Made this a really easy fix for me!

louismrose:
Thanks for the insightful feedback @nornagon. I've addressed the issue that you spotted.

CI is passing again now, except for the code-signing test which we can ignore. Thanks @MarshallOfSound for the tip about that one!

louismrose:
Thanks very much for your help, everyone!

Dumb question -- how will I know when this change makes it into a release?

nornagon:
You can watch release notes here: https://electronjs.org/releases

Current `master` will eventually become Electron 7, so the first stable version that will have this is 7.0.0.

louismrose:
Perfect -- thanks for the tip!

Is there anything I could do to have this considered for a backport into Electron 6 (or even earlier)? This feature is a blocker for one of my employer's big summer projects, and we'd be happy to dedicate some time to a backport if appropriate.

nornagon:
Unfortunately it's past the feature freeze time for Electron 6, so no new features will be accepted for backport. Once 6 is out though you'll be able to use the 7 beta series. See https://electronjs.org/docs/tutorial/electron-timelines.

louismrose:
Ok, great -- we can likely make a beta series work for us. Thanks @nornagon!

PR #19042, opened 2019-06-28
fix: dump correct breakpad symbols on macOS

nornagon:
#### Description of Change
The breakpad symbols we're currently generating on macOS only include exported symbols, which makes them approximately useless for symbolication. This mimics how [Chromium](https://chromium.googlesource.com/chromium/src/+/74.0.3729.5/chrome/BUILD.gn#1331) generates breakpad symbols.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Fixed macOS breakpad symbol files to include non-public symbols.

PR #19127, opened 2019-07-05
build: fix ffmpeg gn gen

jkleinsc:
#### Description of Change
Somehow #19042 broke non proprietary ffmpeg GN gen.  This PR fixes that.  It also reverts #18993 because that PR introduced a security hole and prevented commits on linux and mac.

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/build/args/ffmpeg.gn
+++ b/build/args/ffmpeg.gn
@@ -4,3 +4,4 @@ is_component_ffmpeg = true
 is_official_build = true
 proprietary_codecs = false
 ffmpeg_branding = "Chromium"
+enable_dsyms = false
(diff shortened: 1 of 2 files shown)


build/args/ffmpeg.gn
```diff
@@ -4,3 +4,4 @@ is_component_ffmpeg = true
 is_official_build = true
 proprietary_codecs = false
 ffmpeg_branding = "Chromium"
+enable_dsyms = false
```
nornagon: hm, ideally we'd still generate symbols for this build, though i'm not sure we currently have a place to publish them.

seems like a good fix for now though 👍

... how come this didn't get caught by CI?

Issue #19124, opened 2019-07-05
Can’t input using macOS “Emoji & Symbols” popover


javan:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 6.0.0-beta.12
* **Operating System:**
  * macOS 10.14.5
* **Last Known Working Electron version:**
  * 6.0.0-beta.11

### Expected Behavior

Selecting a character from the *Emoji & Symbols* popover should enter it to the focused input

### Actual Behavior

The popover closes and no character is input

### To Reproduce

1. `$ git clone https://github.com/javan/electron-quick-start.git -b emoji-and-symbols-bug`
1. `$ npm install`
1.  `$ npm start || electron .`
1. Click the *Edit* → *Emoji & Symbols* menu or press <kbd>⌃</kbd><kbd>⌘</kbd><kbd>Space</kbd> to open the popover
1. Select any character from the popover

### Screenshots

<details><summary>v6.0.0-beta.12 👎</summary>

![electron-6-beta-12](https://user-images.githubusercontent.com/5355/60735287-bc64c000-9f20-11e9-9da8-2dfd6165d693.gif)

</details>

<details><summary>v6.0.0-beta.11 👍</summary>

![electron-6-beta-11](https://user-images.githubusercontent.com/5355/60735353-ff269800-9f20-11e9-9d5c-a883b2da3f5e.gif)

</details>

### Additional Information

https://github.com/electron/electron/pull/18995 may be the culprit since it was introduce in beta.12 (cc: @erickzhao  @codebytere)

javan:
Subtle detail: Note that input loses focus when the popover opens in the beta.12 gif above. In beta.11, focus remains.

PR #19128, opened 2019-07-06
[WIP] fix: avoid blurring window when using Character Viewer pop-up on macOS

erickzhao:
##### Context
In PR #18995, I had submitted a fix such that `focus` and `blur` events were listening for changes in the Key window rather than the Main window.

However, Issue #19124 showed that since a window would lose its Key status when the [macOS Character Viewer](https://support.apple.com/en-us/HT201586) was activated, the `blur` event would make any input on the window lose its focus.

Then, the character selected from the Character Viewer wouldn't be entered into the input.

##### Change

I did not find much documentation regarding the behavior of `NSWindow` with respect to the Character Viewer. After playing around with `windowDidResignKey`, I found that when the Character Viewer appears, the window loses its Key status, but the `[NSApp keyWindow]` variable still referenced the original window.

Thus, I added a conditional check to only apply the `blur` event if the resigning window and the new key window weren't the same.

cc @codebytere @zcbenz @javan 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed bug on macOS that broke input from the Character Viewer popup.

### DIFF
--- a/shell/browser/ui/cocoa/atom_ns_window_delegate.mm
+++ b/shell/browser/ui/cocoa/atom_ns_window_delegate.mm
@@ -89,8 +89,6 @@ using TitleBarStyle = electron::NativeWindowMac::TitleBarStyle;
 }
 
 - (void)windowDidBecomeKey:(NSNotification*)notification {
-  // NSWindow* window = notification.object;
-  // LOG(INFO) << window.childWindows.count;
   shell_->NotifyWindowFocus();
 }
 
--- a/shell/browser/ui/cocoa/atom_ns_window_delegate.mm
+++ b/shell/browser/ui/cocoa/atom_ns_window_delegate.mm
@@ -89,11 +89,31 @@ using TitleBarStyle = electron::NativeWindowMac::TitleBarStyle;
 }
 
 - (void)windowDidBecomeKey:(NSNotification*)notification {
-  shell_->NotifyWindowFocus();
+  NSWindow* window = notification.object;
+  if (window.sheet) {
+    shell_->NotifyWindowFocus();
+  }
 }
 
 - (void)windowDidResignKey:(NSNotification*)notification {
-  shell_->NotifyWindowBlur();
+  NSWindow* window = notification.object;
+  if (window.sheet) {
+    shell_->NotifyWindowBlur();
+  }
+}
+
+- (void)windowDidBecomeMain:(NSNotification*)notification {
+  NSWindow* window = notification.object;
+  if (!window.sheet) {
+    shell_->NotifyWindowFocus();
+  }
+}
+
+- (void)windowDidResignMain:(NSNotification*)notification {
+  NSWindow* window = notification.object;
+  if (!window.sheet) {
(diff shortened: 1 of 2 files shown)


shell/browser/ui/cocoa/atom_ns_window_delegate.mm
```diff
@@ -89,11 +89,21 @@ - (NSRect)windowWillUseStandardFrame:(NSWindow*)window
 }
 
 - (void)windowDidBecomeKey:(NSNotification*)notification {
+  // NSWindow* window = notification.object;
+  // LOG(INFO) << window.childWindows.count;
```
nornagon: these can go now :)

erickzhao:
It seems like removing the `rwhv->SetActive` calls fixes #19124. However, #18502 persists even with those calls removed.

However, both solutions give less than desirable outcomes for firing `focus` and `blur` events. On [browser](https://jsfiddle.net/huy89mw5/) and previous versions of Electron, opening the Character Viewer wouldn't fire either event, since the window's `Main` status wouldn't change.

With the `Key` change:
1) Removing `rwhv->SetActive` fires both `focus` and `blur`.
2) This PR's original change only fires `focus` since the `blur` event is blocked.

I'm not particularly sure what to do in this case. It seems like removing the `rwhv` calls gives a more reasonable behavior, though.

cc @zcbenz @nornagon

zcbenz:
I think we got confused with 2 types of `focus`/`blur` here: one is the focus events of `BrowserWindow`, the other is the focus events of DOM Window.

For `BrowserWindow`, the focus events should be main window events, otherwise things like opening spotlight or status item menu would trigger the focus events. For example with PR, while it hacks `windowDidResignKey` so `blur` is emitted correctly for some cases, opening and closing spotlight would still tigger `focus` on `BrowserWindow`, and I don't think we can cover all edge cases.

For DOM Window, the focus events, which are controlled by `SetActive` or `StoreFocus`/`RestoreFocus`, should be key window events, as investigated by @erickzhao and @nornagon.

So to correctly fix both issues without breaking focus events of `BrowserWindow`, I think we should:
1. Still call `OnWindowBlur/OnWindowBlur`(which emits `focus`/`blur` on `BrowserWindow`) on main window events.
2. Call `SetActive` on key window events, with check in `windowDidResignKey`.
3. Call `StoreFocus`/`RestoreFocus` on main window events. (Chrome calls them on tab change, which seems to map to the main window events, it needs some testing to confirm.)

We would need to add more methods to `NativeWindowObserver` to handle key window events separately instead of keeping everything in `OnWindowBlur/OnWindowBlur`.

erickzhao:
@zcbenz @codebytere @nornagon do you think that we should first move to revert the original change? PR #18995 was backported to 6, 5, and 4, and it seems like this issue is more immediately pressing for macOS users than the modal focus behavior.

zcbenz:
Reverting #18995 first sounds good to me.

erickzhao:
Going to close this and attempt to resolve #18502 in a future PR as per the comments above.

PR #19129, opened 2019-07-06
docs: remind user executeJavaScript will not run immediately.

lvscar:
Remind user the `contents.executeJavaScript()` will not run their code immediately if the web page still in running.  Without the knowledge, user would think their code not function properly and it's hard to debug because different page have different loading time.

According to  [web-contents.js](https://github.com/electron/electron/blob/731edbe2b6a767ce5dab8b5760bc66c687b749cf/lib/browser/api/web-contents.js#L199)

#### Checklist

Notes: none

### DIFF
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -1034,6 +1034,8 @@ In the browser window some HTML APIs like `requestFullScreen` can only be
 invoked by a gesture from the user. Setting `userGesture` to `true` will remove
 this limitation.
 
+Code execution will be suspended until web page stop loading.
+
 ```js
 contents.executeJavaScript('fetch("https://jsonplaceholder.typicode.com/users/1").then(resp => resp.json())', true)
   .then((result) => {


docs/api/web-contents.md
```diff
@@ -1034,6 +1034,8 @@ In the browser window some HTML APIs like `requestFullScreen` can only be
 invoked by a gesture from the user. Setting `userGesture` to `true` will remove
 this limitation.
 
+Code execution will be suspended until web page stop loading.
```
zcbenz: It should "finish loading" instead of "stop loading".

```suggestion
If web page is still loading, code execution will be suspended until loading is finished.
```

lvscar: Thank for your attention, IMHO I still think "stop loading" is better because before page has finished loading , we could use "webContents.stop()" imperatively to stop the loading and run code supplied to "contents.executeJavaScript()" immediately. 

In the meantime I'm happy to accept your final modification.  Any reminder is better than nothing.

zcbenz: Yeah you are right, after checking the code I found it is waiting for "did-stop-loading", which is "stop loading".

PR #15701, opened 2018-11-13
fix: crash in sandbox on linux when getting execPath

nornagon:
calling `base::PathService::Get(base::FILE_EXE, &path)` on linux crashes with `--enable-sandbox`:

```
#0  0x00007faabaa2ae97 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007faabaa2c801 in __GI_abort () at abort.c:79
#2  0x000056153ea42085 in  () at ./../../base/debug/debugger_posix.cc:258
#3  0x000056153e97e5be in ~LogMessage() () at ./../../base/logging.cc:858
#4  0x000056153ea55ac4 in PathProviderPosix() () at ./../../base/base_paths_posix.cc:43
#5  0x000056153e9b11ed in Get() () at ./../../base/path_service.cc:207
#6  0x000056153e90857c in GetExecPath () at ../../electron/atom/renderer/atom_sandboxed_renderer_client.cc:85
#7  0x000056153e90857c in InitializeBindings() () at ../../electron/atom/renderer/atom_sandboxed_renderer_client.cc:160
#8  0x000056153e9093f5 in DidCreateScriptContext() () at ../../electron/atom/renderer/atom_sandboxed_renderer_client.cc:219
#9  0x000056153e906607 in DidCreateScriptContext() () at ../../electron/atom/renderer/atom_render_frame_observer.cc:96
#10 0x0000561542469ae5 in DidCreateScriptContext() () at ./../../content/renderer/render_frame_impl.cc:5150
```

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

#### Release Notes

Notes:
* fix a crash on Linux when starting a sandboxed renderer
* the `resourcesPath` property is no longer available on `process` in sandboxed renderers

PR #19135, opened 2019-07-06
refactor: implement GetResourcesPath using MainApplicationBundlePath on Mac

miniak:
#### Description of Change
Re-use existing logic in `MainApplicationBundlePath` to avoid duplicating similar code.
Also remove leftover `NodeBindings::GetHelperResourcesPath()` after #15701.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/shell/common/mac/main_application_bundle.h
+++ b/shell/common/mac/main_application_bundle.h
@@ -6,7 +6,11 @@
 #ifndef SHELL_COMMON_MAC_MAIN_APPLICATION_BUNDLE_H_
 #define SHELL_COMMON_MAC_MAIN_APPLICATION_BUNDLE_H_
 
+#ifdef __OBJC__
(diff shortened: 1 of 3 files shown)


shell/common/mac/main_application_bundle.h
```diff
@@ -6,7 +6,11 @@
 #ifndef SHELL_COMMON_MAC_MAIN_APPLICATION_BUNDLE_H_
 #define SHELL_COMMON_MAC_MAIN_APPLICATION_BUNDLE_H_
 
+#ifdef __OBJC__
 @class NSBundle;
+#else
+struct NSBundle;
```
nornagon: This is super weird, why do we need to define NSBundle in non-objc code?

PR #14389, opened 2018-08-30
refactor: don't expose CallbacksRegistry as an internal module

miniak:
##### Description of Change
There is no reason to treat `CallbacksRegistry` as a module. The related `ObjectsRegistry` is not treated as a module either. It's leaking implementation details as you can access it in apps via `electron.CallbacksRegistry`.

##### Checklist
- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

##### Release Notes
Notes: no-notes

PR #5689, opened 2016-05-25
Support non-native promises

cdaringe:
#### problem statement

- non-native promises are not supported

#### solution

- support non-native promises

#### discussion

there is a lot of talk on how-to-detect if something is a promise or not.  many people respond to the question with something like "don't try to detect a promise, just `Promise.resolve` it if you need promise behavior."  it's a strange response that doesn't address the intent of question--how to do proper type detection on Promises.  it's tricky because a Promise describes both a formal type as well as a behavioral expectation.  due to this complication, detecting a promise is not so simple. @zcbenz already denoted that a simple `.then` check is out of the question, and right fully so.  hence, the strategy taken should cover ~99.9% of promise implementations, with the downside that it takes many type checks to reach a conclusion.

closes #5633

PR #19139, opened 2019-07-07
refactor: don't expose isPromise as an internal module

miniak:
#### Description of Change
There is no reason to expose `isPromise` as a module. Similar helpers `buffer-utils`, `error-utils`, etc. are not exposed either. Also convert to TypeScript.
Similar to #14389.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes


codebytere:
Do we have context on why it was exposed in the first place?

deepak1556:
@codebytere its an internal implementation from https://github.com/electron/electron/pull/5689 , it ended up being exposed because of modules were declared back then. Now we have separate import directive for internal modules and this PR adopts it.

PR #19141, opened 2019-07-07
build: strip recently added swiftshader/libvulkan.so

miniak:
#### Description of Change
#18648 introduced `swiftshader/libvulkan.so`, which needs to be stripped

[electron-v7.0.0-nightly.20190702-linux-x64.zip](https://github.com/electron/nightlies/releases/download/v7.0.0-nightly.20190702/electron-v7.0.0-nightly.20190702-linux-x64.zip) -> 59.9 MB
[electron-v7.0.0-nightly.20190704-linux-x64.zip](https://github.com/electron/nightlies/releases/download/v7.0.0-nightly.20190704/electron-v7.0.0-nightly.20190704-linux-x64.zip) -> 256 MB

Screen Shot 2019-07-07 at 1 47 22 PM

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/script/strip-binaries.py
+++ b/script/strip-binaries.py
@@ -12,7 +12,8 @@ LINUX_BINARIES_TO_STRIP = [
   'libGLESv2.so',
   'libEGL.so',
   'swiftshader/libGLESv2.so',
-  'swiftshader/libEGL.so'
+  'swiftshader/libEGL.so',
+  'swiftshader/libvulkan.so'
 ]
 
 def strip_binaries(directory, target_cpu):


deepak1556:
LGTM,  but moving forward do you think its a good idea to make `strip-binaries` as a build step, so that these library targets/packaging gn step can add it as a dependency and no need to manually track everytime a file is added.

/cc @nornagon

miniak:
@deepak1556 definitely agree. I am not familiar enough with the build system to do it myself quickly, therefore just a simple fix for now.

PR #19145, opened 2019-07-08
 fix: use gn/clang-format from src

zcbenz:
#### Description of Change

The lint scripts should use the `gn`/`clang-format` binaries inside `src/`, instead of the ones in global environment, so the lint results are deterministic.

This fixes lint errors when switching between different Electron major versions. It also fixes lint scripts on some Windows environments, where python can have trouble searching for `gn`/`clang-format` by name.

This PR also includes a style fix due to the `clang-format` in CI being a different version from the one in Chromium.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/lib/util.py
+++ b/script/lib/util.py
@@ -246,3 +246,15 @@ def get_electron_exec():
 
   raise Exception(
       "get_electron_exec: unexpected platform '{0}'".format(sys.platform))
+
+def get_path_in_buildtools(name):
+  buildtools = os.path.realpath(os.path.join(ELECTRON_DIR, '..', 'buildtools'))
+  chromium_platform = {
+    'darwin': 'mac',
+    'linux2': 'linux64',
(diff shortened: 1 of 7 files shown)


alexeykuzmin:
If a Chromium version of `clang-format` is going to be used, the `clang-format` package is not longer needed: https://github.com/electron/electron/blob/master/package.json#L24

script/lib/util.py
```diff
@@ -246,3 +246,15 @@ def get_electron_exec():
 
   raise Exception(
       "get_electron_exec: unexpected platform '{0}'".format(sys.platform))
+
+def get_path_in_buildtools(name):
```
nornagon:
```suggestion
def get_buildtools_executable(name):
```

since it has `.exe` logic

PR #19147, opened 2019-07-08
build: use glob patterns in strip-binaries.py

miniak:
#### Description of Change
Make stripping more robust by using glob patterns instead of hardcoded list of files.
Follow-up to #19141

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/script/strip-binaries.py
+++ b/script/strip-binaries.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 from __future__ import print_function
 import argparse
+import glob
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


miniak:
@nornagon how do we make sure that we strip all binaries correctly and avoid regression like the one fixed by #19141?

nornagon:
hm... GN knows which so files are related:

```
$ gn desc ../out/Testing '//electron:electron_dist_zip__deps' runtime_deps
[...]
./electron
swiftshader/libGLESv2.so
swiftshader/libEGL.so
./libEGL.so
./libGLESv2.so
./libVkICD_mock_icd.so
./libffmpeg.so
./chrome_sandbox
[...]
```

PR #19157, opened 2019-07-08
fix: dump correct breakpad symbols on macOS

nornagon:
Backport of #19042. See that PR for details.

Notes: Fixed macOS breakpad symbol files to include non-public symbols.


nornagon:
Test failure unrelated, appears to be broken on 5-0-x.

PR #19159, opened 2019-07-08
feat: Use GtkFileChooserNative to support the XDG Desktop Portal specification

### LINKED ISSUE #2911: Use desktop environment-aware file picker
At least when using Atom with Plasma DE, the file dialog is the GTK File Chooser one, which feels severely out-of-place (though one gets used to it from using other non-DE aware GTK apps). It would be nice if it could use QFileDialog as does Chromium.

tristan957:
## Description of Change

With this commit, users on KDE/plasma will finally have support in Electron for their native file chooser dialogs. Fixes #2911.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added support for alternative file choosers on Linux through the usage of the XDG Desktop Portal in `GtkFileChooserNative`.

This PR is backwards compatible to support platforms like Ubuntu 16.04/14.04 whose GTK versions are less than 3.20.

### DIFF
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -4,7 +4,6 @@
 
 #include <gmodule.h>
 
-#include <algorithm>
 #include <cctype>
 #include <memory>
 
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -157,7 +157,7 @@ class FileChooserDialog {
 
     InitGtkFileChooserNativeSupport();
 
-    if (supports_gtk_file_chooser_native.value_or(false)) {
+    if (*supports_gtk_file_chooser_native) {
       dialog_ = GTK_FILE_CHOOSER(
           dl_gtk_file_chooser_native_new(settings.title.c_str(), NULL, action,
                                          confirm_text, gtk_util::kCancelLabel));
@@ -169,7 +169,7 @@ class FileChooserDialog {
 
     if (parent_) {
       parent_->SetEnabled(false);
-      if (supports_gtk_file_chooser_native.value_or(false)) {
+      if (*supports_gtk_file_chooser_native) {
         dl_gtk_native_dialog_set_modal(static_cast<void*>(dialog_), TRUE);
       } else {
         gtk::SetGtkTransientForAura(GTK_WIDGET(dialog_),
@@ -213,7 +213,7 @@ class FileChooserDialog {
   }
 
   ~FileChooserDialog() {
-    if (supports_gtk_file_chooser_native.value_or(false)) {
+    if (*supports_gtk_file_chooser_native) {
       dl_gtk_native_dialog_destroy(static_cast<void*>(dialog_));
     } else {
       gtk_widget_destroy(GTK_WIDGET(dialog_));
@@ -248,7 +248,7 @@ class FileChooserDialog {
   void RunAsynchronous() {
     g_signal_connect(dialog_, "response", G_CALLBACK(OnFileDialogResponseThunk),
                      this);
-    if (supports_gtk_file_chooser_native.value_or(false)) {
+    if (*supports_gtk_file_chooser_native) {
       dl_gtk_native_dialog_show(static_cast<void*>(dialog_));
     } else {
       gtk_widget_show_all(GTK_WIDGET(dialog_));
@@ -328,7 +328,7 @@ class FileChooserDialog {
 };
 
 void FileChooserDialog::OnFileDialogResponse(GtkWidget* widget, int response) {
-  if (supports_gtk_file_chooser_native.value_or(false)) {
+  if (*supports_gtk_file_chooser_native) {
     dl_gtk_native_dialog_hide(static_cast<void*>(dialog_));
   } else {
     gtk_widget_hide(GTK_WIDGET(dialog_));
@@ -409,7 +409,7 @@ void FileChooserDialog::OnUpdatePreview(GtkFileChooser* chooser) {
 }  // namespace
 
 void ShowFileDialog(const FileChooserDialog& dialog) {
-  if (supports_gtk_file_chooser_native.value_or(false)) {
+  if (*supports_gtk_file_chooser_native) {
     dl_gtk_native_dialog_show(static_cast<void*>(dialog.dialog()));
   } else {
     gtk_widget_show_all(GTK_WIDGET(dialog.dialog()));
@@ -418,7 +418,7 @@ void ShowFileDialog(const FileChooserDialog& dialog) {
 
 int RunFileDialog(const FileChooserDialog& dialog) {
   int response = 0;
-  if (supports_gtk_file_chooser_native.value_or(false)) {
+  if (*supports_gtk_file_chooser_native) {
     response = dl_gtk_native_dialog_run(static_cast<void*>(dialog.dialog()));
   } else {
     response = gtk_dialog_run(GTK_DIALOG(dialog.dialog()));
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -206,10 +206,12 @@ class FileChooserDialog {
     // org.freedesktop.portal.FileChooser portal. In the case of running through
     // the org.freedesktop.portal.FileChooser portal, anything having to do with
     // the update-preview signal or the preview widget will just be ignored.
-    preview_ = gtk_image_new();
-    g_signal_connect(dialog_, "update-preview",
-                     G_CALLBACK(OnUpdatePreviewThunk), this);
-    gtk_file_chooser_set_preview_widget(dialog_, preview_);
+    if (!*supports_gtk_file_chooser_native) {
+      preview_ = gtk_image_new();
+      g_signal_connect(dialog_, "update-preview",
+                      G_CALLBACK(OnUpdatePreviewThunk), this);
+      gtk_file_chooser_set_preview_widget(dialog_, preview_);
+    }
   }
 
   ~FileChooserDialog() {
@@ -380,6 +382,7 @@ void FileChooserDialog::AddFilters(const Filters& filters) {
 }
 
 void FileChooserDialog::OnUpdatePreview(GtkFileChooser* chooser) {
+  assert(!*supports_gtk_file_chooser_native);
   gchar* filename = gtk_file_chooser_get_preview_filename(chooser);
   if (!filename) {
     gtk_file_chooser_set_preview_widget_active(chooser, FALSE);
@@ -457,6 +460,7 @@ void ShowOpenDialog(const DialogSettings& settings,
 
 bool ShowSaveDialogSync(const DialogSettings& settings, base::FilePath* path) {
   FileChooserDialog save_dialog(GTK_FILE_CHOOSER_ACTION_SAVE, settings);
+  save_dialog.SetupSaveProperties(settings.properties);
 
   ShowFileDialog(save_dialog);
 
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -357,12 +357,12 @@ void FileChooserDialog::AddFilters(const Filters& filters) {
     for (const auto& extension : filter.second) {
       std::string file_extension{"*." + extension};
       // guarantee a pure lowercase variant
-      std::transform(file_extension.begin(), file_extension.end(),
-                     [](unsigned char c) { return std::tolower(c); });
+      for (auto& c : file_extension)
+        c = std::tolower(c);
       gtk_file_filter_add_pattern(gtk_filter, file_extension.c_str());
       // guarantee a pure uppercase variant
-      std::transform(file_extension.begin(), file_extension.end(),
-                     [](unsigned char c) { return std::toupper(c); });
+      for (auto& c : file_extension)
+        c = std::toupper(c);
       gtk_file_filter_add_pattern(gtk_filter, file_extension.c_str());
     }
 
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -209,7 +209,7 @@ class FileChooserDialog {
     if (!*supports_gtk_file_chooser_native) {
       preview_ = gtk_image_new();
       g_signal_connect(dialog_, "update-preview",
-                      G_CALLBACK(OnUpdatePreviewThunk), this);
+                       G_CALLBACK(OnUpdatePreviewThunk), this);
       gtk_file_chooser_set_preview_widget(dialog_, preview_);
     }
   }
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -4,6 +4,8 @@
 
 #include <gmodule.h>
 
+#include <algorithm>
+#include <cctype>
 #include <memory>
 
 #include "base/callback.h"
@@ -57,21 +59,6 @@ namespace {
 
 static const int kPreviewWidth = 256;
 static const int kPreviewHeight = 512;
-
-// Makes sure that .jpg also shows .JPG.
-gboolean FileFilterCaseInsensitive(const GtkFileFilterInfo* file_info,
-                                   std::string* file_extension) {
-  // Makes .* file extension matches all file types.
-  if (*file_extension == ".*")
-    return true;
-  return base::EndsWith(file_info->filename, *file_extension,
-                        base::CompareCase::INSENSITIVE_ASCII);
-}
-
-// Deletes |data| when gtk_file_filter_add_custom() is done with it.
-void OnFileFilterDataDestroyed(std::string* file_extension) {
-  delete file_extension;
-}
 
 void InitGtkFileChooserNativeSupport() {
   // Return early if we have already setup the native functions or we have tried
@@ -368,12 +355,16 @@ void FileChooserDialog::AddFilters(const Filters& filters) {
     GtkFileFilter* gtk_filter = gtk_file_filter_new();
 
     for (const auto& extension : filter.second) {
-      auto file_extension = std::make_unique<std::string>("." + extension);
-      gtk_file_filter_add_custom(
-          gtk_filter, GTK_FILE_FILTER_FILENAME,
-          reinterpret_cast<GtkFileFilterFunc>(FileFilterCaseInsensitive),
-          file_extension.release(),
-          reinterpret_cast<GDestroyNotify>(OnFileFilterDataDestroyed));
+      std::string file_extension{"*." + extension};
+      std::string lower_file_extension =
+          std::transform(extension.begin(), extension.end(),
+                         [](unsigned char c) { return std::tolower(c); });
+      std::string lower_file_extension =
+          std::transform(extension.begin(), extension.end(),
+                         [](unsigned char c) { return std::toupper(c); });
+
+      gtk_file_filter_add_pattern(gtk_filter, lower_file_extension.c_str());
+      gtk_file_filter_add_pattern(gtk_filter, upper_file_extension.c_str());
     }
 
     gtk_file_filter_set_name(gtk_filter, filter.first.c_str());
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -354,14 +354,11 @@ void FileChooserDialog::AddFilters(const Filters& filters) {
     GtkFileFilter* gtk_filter = gtk_file_filter_new();
 
     for (const auto& extension : filter.second) {
-      std::string file_extension{"*." + extension};
       // guarantee a pure lowercase variant
-      for (auto& c : file_extension)
-        c = std::tolower(c);
+      std::string file_extension = base::ToLowerASCII("*." + extension);
       gtk_file_filter_add_pattern(gtk_filter, file_extension.c_str());
       // guarantee a pure uppercase variant
-      for (auto& c : file_extension)
-        c = std::toupper(c);
+      file_extension = base::ToUpperASCII("*." + extension);
       gtk_file_filter_add_pattern(gtk_filter, file_extension.c_str());
     }
 
@@ -371,7 +368,7 @@ void FileChooserDialog::AddFilters(const Filters& filters) {
 }
 
 void FileChooserDialog::OnUpdatePreview(GtkFileChooser* chooser) {
-  assert(!*supports_gtk_file_chooser_native);
+  CHECK(!*supports_gtk_file_chooser_native);
   gchar* filename = gtk_file_chooser_get_preview_filename(chooser);
   if (!filename) {
     gtk_file_chooser_set_preview_widget_active(chooser, FALSE);
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -356,15 +356,14 @@ void FileChooserDialog::AddFilters(const Filters& filters) {
 
     for (const auto& extension : filter.second) {
       std::string file_extension{"*." + extension};
-      std::string lower_file_extension =
-          std::transform(extension.begin(), extension.end(),
-                         [](unsigned char c) { return std::tolower(c); });
-      std::string lower_file_extension =
-          std::transform(extension.begin(), extension.end(),
-                         [](unsigned char c) { return std::toupper(c); });
-
-      gtk_file_filter_add_pattern(gtk_filter, lower_file_extension.c_str());
-      gtk_file_filter_add_pattern(gtk_filter, upper_file_extension.c_str());
+      // guarantee a pure lowercase variant
+      std::transform(file_extension.begin(), file_extension.end(),
+                     [](unsigned char c) { return std::tolower(c); });
+      gtk_file_filter_add_pattern(gtk_filter, file_extension.c_str());
+      // guarantee a pure uppercase variant
+      std::transform(file_extension.begin(), file_extension.end(),
+                     [](unsigned char c) { return std::toupper(c); });
+      gtk_file_filter_add_pattern(gtk_filter, file_extension.c_str());
     }
 
     gtk_file_filter_set_name(gtk_filter, filter.first.c_str());
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -4,7 +4,6 @@
 
 #include <gmodule.h>
 
-#include <cctype>
 #include <memory>
 
 #include "base/callback.h"
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -89,7 +89,7 @@ class FileChooserDialog {
     else if (action == GTK_FILE_CHOOSER_ACTION_OPEN)
       confirm_text = gtk_util::kOpenLabel;
 
-    ::InitGtkFileChooserNativeSupport();
+    file_dialog::InitGtkFileChooserNativeSupport();
 
     if (supports_gtk_native_dialog.value_or(false)) {
       dialog_ = GTK_FILE_CHOOSER(
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -259,7 +259,7 @@ class FileChooserDialog {
       // to make sure window gets correctly raised and gets focus.
       x11::Time time = ui::X11EventSource::GetInstance()->GetTimestamp();
       gtk_window_present_with_time(GTK_WINDOW(dialog_),
-                                    static_cast<uint32_t>(time));
+                                   static_cast<uint32_t>(time));
     }
 #endif
     }
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -132,7 +132,7 @@ class FileChooserDialog {
     if (GTK_IS_DIALOG(dialog_)) {
       preview_ = gtk_image_new();
       g_signal_connect(dialog_, "update-preview",
-                      G_CALLBACK(OnUpdatePreviewThunk), this);
+                       G_CALLBACK(OnUpdatePreviewThunk), this);
       gtk_file_chooser_set_preview_widget(dialog_, preview_);
     }
   }
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -30,6 +30,23 @@
 
 namespace file_dialog {
 
+static GModule* gtk_module_ = nullptr;
+static bool attempted_to_open_gtk_module_ = false;
+static bool supports_gtk_native_dialog_ = false;
+
+static std::function<void(void*)> dl_gtk_native_dialog_show = nullptr;
+static std::function<void(void*)> dl_gtk_native_dialog_destroy = nullptr;
+static std::function<void(void*, gboolean)> dl_gtk_native_dialog_set_modal =
+    nullptr;
+static std::function<int(void*)> dl_gtk_native_dialog_run = nullptr;
+static std::function<void(void*)> dl_gtk_native_dialog_hide = nullptr;
+static std::function<void*(const char*,
+                           GtkWindow*,
+                           GtkFileChooserAction,
+                           const char*,
+                           const char*)>
+    dl_gtk_file_chooser_native_new = nullptr;
+
 DialogSettings::DialogSettings() = default;
 DialogSettings::DialogSettings(const DialogSettings&) = default;
 DialogSettings::~DialogSettings() = default;
@@ -222,25 +239,7 @@ class FileChooserDialog {
 
   GtkFileChooser* dialog() const { return dialog_; }
 
-  static std::function<void(void*)> dl_gtk_native_dialog_show;
-  static std::function<void(void*)> dl_gtk_native_dialog_destroy;
-  static std::function<void(void*, gboolean)> dl_gtk_native_dialog_set_modal;
-  static std::function<int(void*)> dl_gtk_native_dialog_run;
-  static std::function<void(void*)> dl_gtk_native_dialog_hide;
-  static std::function<void*(const char*,
-                             GtkWindow*,
-                             GtkFileChooserAction,
-                             const char*,
-                             const char*)>
-      dl_gtk_file_chooser_native_new;
-
  private:
-  static GModule* gtk_module_;
-  static bool attempted_to_open_gtk_module_;
-  static bool supports_gtk_native_dialog_;
-
-  static void InitGtkNativeDialogSupport();
-
   void AddFilters(const Filters& filters);
 
   electron::NativeWindowViews* parent_;
@@ -258,89 +257,6 @@ class FileChooserDialog {
 
   DISALLOW_COPY_AND_ASSIGN(FileChooserDialog);
 };
-
-void FileChooserDialog::InitGtkNativeDialogSupport() {
-  // Return early if we have already setup the static members or we have tried
-  // once before and failed. Avoid running expensive dynamic library operations.
-  if (supports_gtk_native_dialog_ || attempted_to_open_gtk_module_) {
-    return;
-  }
-
-  // Mark that we have attempted to initialize support at least once
-  attempted_to_open_gtk_module_ = true;
-
-  if (!g_module_supported()) {
-    return;
-  }
-
-  gtk_module_ = g_module_open("libgtk-3.so", G_MODULE_BIND_LAZY);
-  if (!gtk_module_) {
-    return;
-  }
-
-  // Will never be unloaded
-  g_module_make_resident(gtk_module_);
-
-  void* (*_dl_gtk_file_chooser_native_new)(
-      const char*, GtkWindow*, GtkFileChooserAction, const char*, const char*) =
-      nullptr;
-  bool found = g_module_symbol(
-      gtk_module_, "gtk_file_chooser_native_new",
-      reinterpret_cast<void**>(&_dl_gtk_file_chooser_native_new));
-  if (!found) {
-    return;
-  }
-  dl_gtk_file_chooser_native_new = _dl_gtk_file_chooser_native_new;
-
-  void (*_dl_gtk_native_dialog_set_modal)(void*, gboolean) = nullptr;
-  found = g_module_symbol(
-      gtk_module_, "gtk_native_dialog_set_modal",
-      reinterpret_cast<void**>(&dl_gtk_native_dialog_set_modal));
-  if (!found) {
-    return;
-  }
-  dl_gtk_native_dialog_set_modal = _dl_gtk_native_dialog_set_modal;
-
-  void (*_dl_gtk_native_dialog_destroy)(void*) = nullptr;
-  found =
-      g_module_symbol(gtk_module_, "gtk_native_dialog_destroy",
-                      reinterpret_cast<void**>(&_dl_gtk_native_dialog_destroy));
-  if (!found) {
-    return;
-  }
-  dl_gtk_native_dialog_destroy = _dl_gtk_native_dialog_destroy;
-
-  void (*_dl_gtk_native_dialog_show)(void*) = NULL;
-  found =
-      g_module_symbol(gtk_module_, "gtk_native_dialog_show",
-                      reinterpret_cast<void**>(&_dl_gtk_native_dialog_show));
-  if (!found) {
-    return;
-  }
-  dl_gtk_native_dialog_show = _dl_gtk_native_dialog_show;
-
-  void (*_dl_gtk_native_dialog_hide)(void*) = NULL;
-  found =
-      g_module_symbol(gtk_module_, "gtk_native_dialog_hide",
-                      reinterpret_cast<void**>(&_dl_gtk_native_dialog_hide));
-  if (!found) {
-    return;
-  }
-  dl_gtk_native_dialog_hide = _dl_gtk_native_dialog_hide;
-
-  int (*_dl_gtk_native_dialog_run)(void*) = NULL;
-  found = g_module_symbol(gtk_module_, "gtk_native_dialog_run",
-                          reinterpret_cast<void**>(&_dl_gtk_native_dialog_run));
-  if (!found) {
-    return;
-  }
-  dl_gtk_native_dialog_run = _dl_gtk_native_dialog_run;
-
-  supports_gtk_native_dialog_ =
-      dl_gtk_file_chooser_native_new && dl_gtk_native_dialog_set_modal &&
-      dl_gtk_native_dialog_destroy && dl_gtk_native_dialog_run &&
-      dl_gtk_native_dialog_show && dl_gtk_native_dialog_hide;
-}
 
 void FileChooserDialog::OnFileDialogResponse(GtkWidget* widget, int response) {
   if (supports_gtk_native_dialog_) {
@@ -425,8 +341,7 @@ void FileChooserDialog::OnUpdatePreview(GtkFileChooser* chooser) {
 
 void ShowFileDialog(const FileChooserDialog& dialog) {
   if (dialog.supports_gtk_native_dialog()) {
-    FileChooserDialog::dl_gtk_native_dialog_show(
-        static_cast<void*>(dialog.dialog()));
+    dl_gtk_native_dialog_show(static_cast<void*>(dialog.dialog()));
   } else {
     gtk_widget_show_all(GTK_WIDGET(dialog.dialog()));
   }
@@ -435,8 +350,7 @@ void ShowFileDialog(const FileChooserDialog& dialog) {
 int RunFileDialog(const FileChooserDialog& dialog) {
   int response = 0;
   if (dialog.supports_gtk_native_dialog()) {
-    response = FileChooserDialog::dl_gtk_native_dialog_run(
-        static_cast<void*>(dialog.dialog()));
+    response = dl_gtk_native_dialog_run(static_cast<void*>(dialog.dialog()));
   } else {
     response = gtk_dialog_run(GTK_DIALOG(dialog.dialog()));
   }
@@ -490,6 +404,89 @@ void ShowSaveDialog(const DialogSettings& settings,
   FileChooserDialog* save_dialog =
       new FileChooserDialog(GTK_FILE_CHOOSER_ACTION_SAVE, settings);
   save_dialog->RunSaveAsynchronous(std::move(promise));
+}
+
+void InitGtkNativeDialogSupport() {
+  // Return early if we have already setup the static members or we have tried
+  // once before and failed. Avoid running expensive dynamic library operations.
+  if (supports_gtk_native_dialog_ || attempted_to_open_gtk_module_) {
+    return;
+  }
+
+  // Mark that we have attempted to initialize support at least once
+  attempted_to_open_gtk_module_ = true;
+
+  if (!g_module_supported()) {
+    return;
+  }
+
+  gtk_module_ = g_module_open("libgtk-3.so", G_MODULE_BIND_LAZY);
+  if (!gtk_module_) {
+    return;
+  }
+
+  // Will never be unloaded
+  g_module_make_resident(gtk_module_);
+
+  void* (*_dl_gtk_file_chooser_native_new)(
+      const char*, GtkWindow*, GtkFileChooserAction, const char*, const char*) =
+      nullptr;
+  bool found = g_module_symbol(
+      gtk_module_, "gtk_file_chooser_native_new",
+      reinterpret_cast<void**>(&_dl_gtk_file_chooser_native_new));
+  if (!found) {
+    return;
+  }
+  dl_gtk_file_chooser_native_new = _dl_gtk_file_chooser_native_new;
+
+  void (*_dl_gtk_native_dialog_set_modal)(void*, gboolean) = nullptr;
+  found = g_module_symbol(
+      gtk_module_, "gtk_native_dialog_set_modal",
+      reinterpret_cast<void**>(&dl_gtk_native_dialog_set_modal));
+  if (!found) {
+    return;
+  }
+  dl_gtk_native_dialog_set_modal = _dl_gtk_native_dialog_set_modal;
+
+  void (*_dl_gtk_native_dialog_destroy)(void*) = nullptr;
+  found =
+      g_module_symbol(gtk_module_, "gtk_native_dialog_destroy",
+                      reinterpret_cast<void**>(&_dl_gtk_native_dialog_destroy));
+  if (!found) {
+    return;
+  }
+  dl_gtk_native_dialog_destroy = _dl_gtk_native_dialog_destroy;
+
+  void (*_dl_gtk_native_dialog_show)(void*) = NULL;
+  found =
+      g_module_symbol(gtk_module_, "gtk_native_dialog_show",
+                      reinterpret_cast<void**>(&_dl_gtk_native_dialog_show));
+  if (!found) {
+    return;
+  }
+  dl_gtk_native_dialog_show = _dl_gtk_native_dialog_show;
+
+  void (*_dl_gtk_native_dialog_hide)(void*) = NULL;
+  found =
+      g_module_symbol(gtk_module_, "gtk_native_dialog_hide",
+                      reinterpret_cast<void**>(&_dl_gtk_native_dialog_hide));
+  if (!found) {
+    return;
+  }
+  dl_gtk_native_dialog_hide = _dl_gtk_native_dialog_hide;
+
+  int (*_dl_gtk_native_dialog_run)(void*) = NULL;
+  found = g_module_symbol(gtk_module_, "gtk_native_dialog_run",
+                          reinterpret_cast<void**>(&_dl_gtk_native_dialog_run));
+  if (!found) {
+    return;
+  }
+  dl_gtk_native_dialog_run = _dl_gtk_native_dialog_run;
+
+  supports_gtk_native_dialog_ =
+      dl_gtk_file_chooser_native_new && dl_gtk_native_dialog_set_modal &&
+      dl_gtk_native_dialog_destroy && dl_gtk_native_dialog_run &&
+      dl_gtk_native_dialog_show && dl_gtk_native_dialog_hide;
 }
 
 }  // namespace file_dialog
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -212,7 +212,7 @@ class FileChooserDialog {
   CHROMEG_CALLBACK_1(FileChooserDialog,
                      void,
                      OnFileDialogResponse,
-                     GtkFileChooser*,
+                     GtkWidget*,
                      int);
 
   static bool supports_gtk_native_dialog() {
@@ -234,11 +234,11 @@ class FileChooserDialog {
       dl_gtk_file_chooser_native_new;
 
  private:
-  static GModule* gtk_module_ = nullptr;
-  static bool attempted_to_open_gtk_module_ = false;
-  static bool supports_gtk_native_dialog_ = false;
+  static GModule* gtk_module_;
+  static bool attempted_to_open_gtk_module_;
+  static bool supports_gtk_native_dialog_;
 
-  static InitGtkNativeDialogSupport();
+  static void InitGtkNativeDialogSupport();
 
   void AddFilters(const Filters& filters);
 
@@ -345,7 +345,7 @@ void FileChooserDialog::OnFileDialogResponse(GtkWidget* widget, int response) {
   if (supports_gtk_native_dialog_) {
     dl_gtk_native_dialog_hide(static_cast<void*>(dialog_));
   } else {
-    gtk_widget_hide(dialog_);
+    gtk_widget_hide(GTK_WIDGET(dialog_));
   }
   v8::Isolate* isolate = electron::JavascriptEnvironment::GetIsolate();
   v8::HandleScope scope(isolate);
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -89,7 +89,7 @@ class FileChooserDialog {
     else if (action == GTK_FILE_CHOOSER_ACTION_OPEN)
       confirm_text = gtk_util::kOpenLabel;
 
-    InitGtkFileChooserNativeSupport();
+    ::InitGtkFileChooserNativeSupport();
 
     if (supports_gtk_native_dialog.value_or(false)) {
       dialog_ = GTK_FILE_CHOOSER(
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -2,18 +2,22 @@
 // Use of this source code is governed by the MIT license that can be
 // found in the LICENSE file.
 
-#include <memory>
+#include <glib/gi18n.h>  // _() macro
+#include <gmodule.h>
 
-#include "shell/browser/ui/file_dialog.h"
-#include "shell/browser/ui/gtk_util.h"
+#include <functional>
+#include <memory>
 
 #include "base/callback.h"
 #include "base/files/file_util.h"
 #include "base/strings/string_util.h"
 #include "shell/browser/javascript_environment.h"
 #include "shell/browser/native_window_views.h"
+#include "shell/browser/ui/file_dialog.h"
+#include "shell/browser/ui/gtk_util.h"
 #include "shell/browser/unresponsive_suppressor.h"
 #include "shell/common/gin_converters/file_path_converter.h"
+#include "ui/aura/window.h"
 #include "ui/base/glib/glib_signal.h"
 #include "ui/gtk/gtk_util.h"
 
@@ -66,35 +70,58 @@ class FileChooserDialog {
     else if (action == GTK_FILE_CHOOSER_ACTION_OPEN)
       confirm_text = gtk_util::kOpenLabel;
 
-    dialog_ = gtk_file_chooser_dialog_new(
-        settings.title.c_str(), nullptr, action, gtk_util::kCancelLabel,
-        GTK_RESPONSE_CANCEL, confirm_text, GTK_RESPONSE_ACCEPT, NULL);
+    gtk_module_ = g_module_open("libgtk-3.so", G_MODULE_BIND_LAZY);
+    void* (*dl_gtk_file_chooser_native_new)(const char*, GtkWindow*,
+                                            GtkFileChooserAction, const char*,
+                                            const char*) = NULL;
+    bool found = g_module_symbol(
+        gtk_module_, "gtk_file_chooser_native_new",
+        reinterpret_cast<void**>(&dl_gtk_file_chooser_native_new));
+    if (found && dl_gtk_file_chooser_native_new != NULL) {
+      dialog_ = GTK_FILE_CHOOSER(
+          dl_gtk_file_chooser_native_new(settings.title.c_str(), NULL, action,
+                                         gtk_util::kCancelLabel, confirm_text));
+    } else {
+      dialog_ = GTK_FILE_CHOOSER(gtk_file_chooser_dialog_new(
+          settings.title.c_str(), NULL, action, gtk_util::kCancelLabel,
+          GTK_RESPONSE_CANCEL, confirm_text, GTK_RESPONSE_ACCEPT, NULL));
+    }
+
     if (parent_) {
       parent_->SetEnabled(false);
-      gtk::SetGtkTransientForAura(dialog_, parent_->GetNativeWindow());
-      gtk_window_set_modal(GTK_WINDOW(dialog_), TRUE);
+      if (GTK_IS_DIALOG(dialog_))
+        libgtkui::SetGtkTransientForAura(GTK_WIDGET(dialog_),
+                                         parent_->GetNativeWindow());
+      else
+        SetGtkTransientForAura(parent_->GetNativeWindow());
+      if (GTK_IS_DIALOG(dialog_)) {
+        gtk_window_set_modal(GTK_WINDOW(dialog_), TRUE);
+      } else {
+        void (*dl_gtk_native_dialog_set_modal)(void*, bool) = NULL;
+        g_module_symbol(
+            gtk_module_, "gtk_native_dialog_set_modal",
+            reinterpret_cast<void**>(&dl_gtk_native_dialog_set_modal));
+        dl_gtk_native_dialog_set_modal(static_cast<void*>(dialog_), TRUE);
+      }
     }
 
     if (action == GTK_FILE_CHOOSER_ACTION_SAVE)
-      gtk_file_chooser_set_do_overwrite_confirmation(GTK_FILE_CHOOSER(dialog_),
-                                                     TRUE);
+      gtk_file_chooser_set_do_overwrite_confirmation(dialog_, TRUE);
     if (action != GTK_FILE_CHOOSER_ACTION_OPEN)
-      gtk_file_chooser_set_create_folders(GTK_FILE_CHOOSER(dialog_), TRUE);
+      gtk_file_chooser_set_create_folders(dialog_, TRUE);
 
     if (!settings.default_path.empty()) {
       if (base::DirectoryExists(settings.default_path)) {
         gtk_file_chooser_set_current_folder(
-            GTK_FILE_CHOOSER(dialog_), settings.default_path.value().c_str());
+            dialog_, settings.default_path.value().c_str());
       } else {
         if (settings.default_path.IsAbsolute()) {
           gtk_file_chooser_set_current_folder(
-              GTK_FILE_CHOOSER(dialog_),
-              settings.default_path.DirName().value().c_str());
+              dialog_, settings.default_path.DirName().value().c_str());
         }
 
         gtk_file_chooser_set_current_name(
-            GTK_FILE_CHOOSER(dialog_),
-            settings.default_path.BaseName().value().c_str());
+            dialog_, settings.default_path.BaseName().value().c_str());
       }
     }
 
@@ -104,11 +131,21 @@ class FileChooserDialog {
     preview_ = gtk_image_new();
     g_signal_connect(dialog_, "update-preview",
                      G_CALLBACK(OnUpdatePreviewThunk), this);
-    gtk_file_chooser_set_preview_widget(GTK_FILE_CHOOSER(dialog_), preview_);
+    gtk_file_chooser_set_preview_widget(dialog_, preview_);
   }
 
   ~FileChooserDialog() {
-    gtk_widget_destroy(dialog_);
+    if (GTK_IS_DIALOG(dialog_)) {
+      gtk_widget_destroy(GTK_WIDGET(dialog_));
+    } else {
+      void (*dl_gtk_native_dialog_destroy)(void*) = NULL;
+      g_module_symbol(gtk_module_, "gtk_native_dialog_destroy",
+                      reinterpret_cast<void**>(&dl_gtk_native_dialog_destroy));
+      dl_gtk_native_dialog_destroy(static_cast<void*>(dialog_));
+    }
+
+    g_clear_pointer(&gtk_module_, g_module_close);
+
     if (parent_)
       parent_->SetEnabled(true);
   }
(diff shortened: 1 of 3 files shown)


tristan957:
I would appreciate guidance in getting this PR to completion with regard to CI builds, styling, implementation, and testing.

Off the bat, there are a few issues with this PR:

* I currently am not sure how to cast a ` GtkFileChooserNative` to a `GtkNativeDialog`
* Does the include of `dlfcn.h` need to be guarded in anyway?

tristan957:
Please note that this was not on my dev machine, so I could not compile to.

tristan957:
Lint should be fixed. Apologies.

tristan957:
@TingPing @ckerr since you guys were a part of the last PR, can you provide assistance on this one?

@TingPing specifically if you see anything that can be done about casting a `GtkFileChooserNative` to `GtkNativeDialog` without the macro `GTK_NATIVE_DIALOG` that would be extremely helpful. This is dependent on whether or not GTK stays dynamically loaded or not.

TingPing:
Thanks a lot for picking this up!

> @TingPing specifically if you see anything that can be done about casting a GtkFileChooserNative to GtkNativeDialog without the macro GTK_NATIVE_DIALOG that would be extremely helpful. 

Honestly I think its acceptable to just ignore in this specific case and cast away to `void*`. 

Hardcoding `-ldl` is not ideal as that isn't used on all unix-like systems.

Since this already depends on GLib I suggest using GModule which avoids the linking concerns, header concerns, and any portability concerns.

You'd then use `g_module_open()`, `g_module_symbol()`, etc: https://developer.gnome.org/glib/stable/glib-Dynamic-Loading-of-Modules.html

tristan957:
Thanks for the link @TingPing. I will take a look at `GModule` later today and update the PR accordingly.

tristan957:
Call for action: I do not currently run a KDE Plasma desktop. If someone who does could inevitably test this PR, that would be great.

tristan957:
@TingPing I was more wondering if the `GtkFileChooserNative` object is located in memory at the same spot as the `GtkNativeDialog` or does `GTK_NATIVE_DIALOG` move the pointer. I don't have a complete grasp of the GObject type system just yet. I can write it, but I don't know technicalities like this.

tristan957:
This is going to require a little more effort than previously thought. Still doable I think, but I am under the impression `_dialog` is going to have to change its type to a `GtkFileChooser`. I am also going to have to dip into libgtkui::SetGtkTransientForAura to account for this.

tristan957:
The behavior of `FileDialog::RunAsynchronous` needs to be re-evaluated as "delete-event" does not exist for `GtkNativeDialog` & `GtkFileChooserNative`. Also there is no "present with time" api, so that will need to be reevaluated.

Both `CHROME_CALLBACKS` need to be adjusted.

I am going to need some guidance on `libgtkui::SetGtkTransientForAura`. I cannot find the definition. I just need to overload the function to have a `SetGtkTransientForAura(void* dialog, aura::Window* parent)`. I need use the underlying `GtkNativeDialog` if `GTK_IS_WIDGET(dialog)` fails because that API provides `gtk_native_dialog_set_transient_for`.

tristan957:
I have also learned that preview widgets are ignored in the Native API. What should be the best course of action regarding that? Leaving the preview widget code in regardless of what dialog is actually being used seems best to me instead of making conditional checks around the code.

tristan957:
I have gotten the build where it only fails on the `libgtkui::SetGtkTransientForAura`

tristan957:
I have come up with the following patch

```patch
diff --git a/shell/browser/ui/file_dialog_gtk.cc b/shell/browser/ui/file_dialog_gtk.cc
index ef6d3a545..662a541d9 100644
--- a/shell/browser/ui/file_dialog_gtk.cc
+++ b/shell/browser/ui/file_dialog_gtk.cc
@@ -77,7 +77,7 @@ class FileChooserDialog {
 
     if (parent_) {
       parent_->SetEnabled(false);
-      libgtkui::SetGtkTransientForAura(dialog_, parent_->GetNativeWindow());
+      SetGtkTransientForAura(parent_->GetNativeWindow());
       if (GTK_IS_WINDOW(dialog_)) {
         gtk_window_set_modal(GTK_WINDOW(dialog_), TRUE);
       } else {
@@ -225,6 +225,19 @@ class FileChooserDialog {
   CHROMEG_CALLBACK_0(FileChooserDialog, void, OnUpdatePreview, GtkFileChooser*);
 
   DISALLOW_COPY_AND_ASSIGN(FileChooserDialog);
+
+  // Sets |dialog| as transient for |parent|, which will keep it on top and center
+  // it above |parent|. Do nothing if |parent| is nullptr.
+  void SetGtkTransientForAura(aura::Window* parent) {
+    if (!parent || !parent->GetHost())
+      return;
+
+    void (*dl_gtk_native_dialog_set_transient_for)(void*, GtkWindow*) = nullptr;
+    g_module_symbol(
+        gtk_module_, "gtk_native_dialog_set_transient_for",
+        reinterpret_cast<void**>(&dl_gtk_native_dialog_set_transient_for);
+    dl_gtk_native_dialog_set_transient_for(dialog_, reinterpret_cast<GtkWindow*>(parent));
+  }
 };
 
 void FileChooserDialog::OnFileDialogResponse(GtkFileChooser* widget, int response) {
```

TingPing:
> @TingPing I was more wondering if the GtkFileChooserNative object is located in memory at the same spot as the GtkNativeDialog or does GTK_NATIVE_DIALOG move the pointer. I don't have a complete grasp of the GObject type system just yet. I can write it, but I don't know technicalities like this. If they are at the same spot, I can avoid the cast.

The type casting macros in GLib do two things: They do a type check and then a regular C style cast. They do not change the pointer in any way.

TingPing:
> The behavior of FileDialog::RunAsynchronous needs to be re-evaluated as "delete-event" does not exist for GtkNativeDialog & GtkFileChooserNative

It always should have been watching the `response` signal.

TingPing:
> I have also learned that preview widgets are ignored in the Native API. What should be the best course of action regarding that? 

Nothing can be done about that. Native dialogs cannot be modified.

tristan957:
Company and mclasen were helpful on #gtk yesterday. @TingPing when you close out of a GtkDialog, what is the response code? Is it the same as hitting cancel?

TingPing:
> @TingPing when you close out of a GtkDialog, what is the response code? Is it the same as hitting cancel?

`GTK_RESPONSE_DELETE_EVENT`

tristan957:
I can probably include a workaround for that event then.

Is this even still an issue on KDE systems. I made a post on /r/kde, and people seem to get the plasma file chooser in electron applications.

polarathene:
> Is this even still an issue on KDE systems. I made a post on /r/kde, and people seem to get the plasma file chooser in electron applications.

Which electron application was that? My electron app does not have a native KDE dialog. If you package with Flatpak it will be a native dialog due to portals used there afaik.

tristan957:
Does anyone know why Electron has a separate file chooser implementation than Chromium?

zcbenz:
> Does anyone know why Electron has a separate file chooser implementation than Chromium?

The Chromium implementation of file chooser was limited of features, and could not meet Electron's feature requests.

TingPing:
As I understand it Chromium might be moving in a direction where they always use the ChromiumOS file chooser and not native toolkits also. So Electron probably wants to retain that native integration.

tristan957:
Does anyone know how to pull an electron package out of circle CI so I could use it in a node project, or does the CI not produce those?

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -94,11 +120,21 @@ class FileChooserDialog {
...
+      gtk_widget_destroy(GTK_WIDGET(dialog_));
+    } else {
+      void (*dl_gtk_native_dialog_destroy)(void*) = nullptr;
+      g_module_symbol(gtk_module_, "gtk_native_dialog_destroy",
+                      reinterpret_cast<void**>(&dl_gtk_native_dialog_destroy));
+      dl_gtk_native_dialog_destroy(static_cast<void*>(dialog_));
+    }
+    if (gtk_module_ != nullptr) {
+      g_module_close(gtk_module_);
+    }
```
ckerr: terser / slightly better: `g_clear_pointer(&gtk_module_, g_module_close);`

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -107,24 +143,35 @@ class FileChooserDialog {
...
 
     // We need to call gtk_window_present after making the widgets visible to
     // make sure window gets correctly raised and gets focus.
-    int time = ui::X11EventSource::GetInstance()->GetTimestamp();
-    gtk_window_present_with_time(GTK_WINDOW(dialog_), time);
+    if (GTK_IS_WINDOW(dialog_)) {
+      int time = ui::X11EventSource::GetInstance()->GetTimestamp();
+      gtk_window_present_with_time(GTK_WINDOW(dialog_), time);
+    }
```
ckerr: This block would flow better going in right after `gtk_widget_show_all()`

```c++
if (GTK_IS_NATIVE_DIALOG(...)) {
   ...
} else {
  gtk_wiget_show_all(...)
  // We need to call...
  int time = ...
  gtk_window_present(...)
}
```

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -159,32 +206,56 @@ class FileChooserDialog {
...
 
   // Callback for when we update the preview for the selection.
-  CHROMEG_CALLBACK_0(FileChooserDialog, void, OnUpdatePreview, GtkWidget*);
+  CHROMEG_CALLBACK_0(FileChooserDialog, void, OnUpdatePreview, GtkFileChooser*);
+
+  // Sets |dialog| as transient for |parent|, which will keep it on top and
+  // center it above |parent|. Do nothing if |parent| is nullptr.
+  void SetGtkTransientForAura(aura::Window* parent) {
+    if (!parent || !parent->GetHost())
+      return;
```
ckerr: I don't see the `!parent->GetHost()` call in the earlier version. Has something changed to cause this test to be needed?

Also, if we do need it, should we warn on failure with `G_RETURN_IF_FAIL(!parent->GetHost() != nullptr);` ?

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -159,32 +206,56 @@ class FileChooserDialog {
...
+void FileChooserDialog::OnFileDialogResponse(GtkFileChooser* widget,
+                                             int response) {
+  if (GTK_IS_WIDGET(dialog_)) {
+    gtk_widget_hide(GTK_WIDGET(dialog_));
+  } else {
+    void (*dl_gtk_native_dialog_hide)(void*) = nullptr;
+    g_module_symbol(gtk_module_, "gtk_native_dialog_hide",
+                    reinterpret_cast<void**>(&dl_gtk_native_dialog_hide));
+    dl_gtk_native_dialog_hide(static_cast<void*>(dialog_));
+  }
```
ckerr:
```c++
if (GTK_IS_NATIVE_DIALOG(...)) {
   ...
} else {
   ...
}
```

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -272,8 +339,25 @@ bool ShowOpenDialogSync(const DialogSettings& settings,
...
-  gtk_widget_show_all(open_dialog.dialog());
-  int response = gtk_dialog_run(GTK_DIALOG(open_dialog.dialog()));
+  if (GTK_IS_WIDGET(open_dialog.dialog())) {
+    gtk_widget_show_all(GTK_WIDGET(open_dialog.dialog()));
+  } else {
+    void (*dl_gtk_native_dialog_show)(void*) = nullptr;
+    g_module_symbol(open_dialog.gtk_module(), "gtk_native_dialog_show",
+                    reinterpret_cast<void**>(&dl_gtk_native_dialog_show));
+    dl_gtk_native_dialog_show(static_cast<void*>(open_dialog.dialog()));
+  }
```
ckerr: We have this logic twice now. Maybe a new helper `ShowFileDialog(void*)` similar to the `SetGtkTransientForAura()` helper

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -272,8 +339,25 @@ bool ShowOpenDialogSync(const DialogSettings& settings,
...
+  int response = 0;
+  if (GTK_IS_WIDGET(open_dialog.dialog())) {
+    response = gtk_dialog_run(GTK_DIALOG(open_dialog.dialog()));
+  } else {
+    int (*dl_gtk_native_dialog_run)(void*) = nullptr;
+    g_module_symbol(open_dialog.gtk_module(), "gtk_native_dialog_run",
+                    reinterpret_cast<void**>(&dl_gtk_native_dialog_run));
+    response =
+        dl_gtk_native_dialog_run(static_cast<void*>(open_dialog.dialog()));
+  }
```
ckerr:
```c++
if (GTK_IS_NATIVE_DIALOG(dialog_)) {
   ...
} else {
   ...
}
```

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -293,8 +377,26 @@ void ShowOpenDialog(const DialogSettings& settings,
...
+  int response = 0;
+  if (GTK_IS_WIDGET(save_dialog.dialog())) {
+    response = gtk_dialog_run(GTK_DIALOG(save_dialog.dialog()));
+  } else {
+    int (*dl_gtk_native_dialog_run)(void*) = nullptr;
+    g_module_symbol(save_dialog.gtk_module(), "gtk_native_dialog_run",
+                    reinterpret_cast<void**>(&dl_gtk_native_dialog_run));
+    response =
+        dl_gtk_native_dialog_run(static_cast<void*>(save_dialog.dialog()));
+  }
```
ckerr: DRY: Use `int RunFileDIalog(void*)` here

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -293,8 +377,26 @@ void ShowOpenDialog(const DialogSettings& settings,
...
-  int response = gtk_dialog_run(GTK_DIALOG(save_dialog.dialog()));
+
+  if (GTK_IS_WIDGET(save_dialog.dialog())) {
+    gtk_widget_show_all(GTK_WIDGET(save_dialog.dialog()));
+  } else {
+    void (*dl_gtk_native_dialog_show)(void*) = nullptr;
+    g_module_symbol(save_dialog.gtk_module(), "gtk_native_dialog_show",
+                    reinterpret_cast<void**>(&dl_gtk_native_dialog_show));
+    dl_gtk_native_dialog_show(static_cast<void*>(save_dialog.dialog()));
+  }
```
ckerr: DRY: Use `ShowFileDialog(void*)` here too

shell/browser/ui/file_dialog_gtk.cc
ckerr: Also, this code block exists in two places. We should extract-method and create a helper `int RunFileDialog(void*)`

tristan957:
@ckerr thanks for taking a look at this. I will review and update the PR according to your suggestions hopefully later this weekend.

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -56,35 +59,58 @@ class FileChooserDialog {
     else if (action == GTK_FILE_CHOOSER_ACTION_OPEN)
       confirm_text = _("_Open");
 
-    dialog_ = gtk_file_chooser_dialog_new(
-        settings.title.c_str(), NULL, action, _("_Cancel"), GTK_RESPONSE_CANCEL,
-        confirm_text, GTK_RESPONSE_ACCEPT, NULL);
+    gtk_module_ = g_module_open("libgtk-3.so", G_MODULE_BIND_LAZY);
```
TingPing:
```suggestion
    gtk_module_ = g_module_open("libgtk-3.so.0", G_MODULE_BIND_LAZY);
```

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -56,35 +59,58 @@ class FileChooserDialog {
...
+    void* (*dl_gtk_file_chooser_native_new)(const char*, GtkWindow*,
+                                            GtkFileChooserAction, const char*,
+                                            const char*) = nullptr;
+    bool found = g_module_symbol(
```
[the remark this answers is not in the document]
TingPing: `gtk_module_` should never be NULL, probably fine to handle that ofc but it should be a loud warning when your app using GTK is loading a non-existing so.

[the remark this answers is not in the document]
tristan957: I am thinking we could make gtk_symbol_ a static class var, that way the only time we go searching for the lib is on the first invocation thoughts? Then it could be `(gtk_module_ != NULL) || g_module_symbol...`

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -56,35 +59,58 @@ class FileChooserDialog {
...
+        gtk_module_, "gtk_file_chooser_native_new",
+        reinterpret_cast<void**>(&dl_gtk_file_chooser_native_new));
+    if (gtk_module_ != nullptr && found &&
+        dl_gtk_file_chooser_native_new != nullptr) {
+      dialog_ = GTK_FILE_CHOOSER(dl_gtk_file_chooser_native_new(
+          settings.title.c_str(), NULL, action, _("_Cancel"), confirm_text));
```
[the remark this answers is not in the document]
tristan957: I'll fix. Any reason for the aversion to nullptr over NULL, so that I can learn?

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -56,35 +59,58 @@ class FileChooserDialog {
...
+      dialog_ = GTK_FILE_CHOOSER(gtk_file_chooser_dialog_new(
+          settings.title.c_str(), NULL, action, _("_Cancel"),
+          GTK_RESPONSE_CANCEL, confirm_text, GTK_RESPONSE_ACCEPT, NULL));
+    }
+
     if (parent_) {
       parent_->SetEnabled(false);
-      libgtkui::SetGtkTransientForAura(dialog_, parent_->GetNativeWindow());
-      gtk_window_set_modal(GTK_WINDOW(dialog_), TRUE);
+      if (GTK_IS_WIDGET(dialog_))
```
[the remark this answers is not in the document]
tristan957: Will we have access to `GTK_IS_NATIVE_DIALOG`? I thought you wanted electron to build on Ubuntu 16.04 which doesn't support GtkNativeDialog. GtkNativeDialog is not a GtkWidget as far as I know. @TingPing can you confirm?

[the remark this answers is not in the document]
tristan957: Would a compromise be GTK_IS_WINDOW? GtkNativeDialog is explicitly not a GtkWindow.

> Native dialogs are platform dialogs that don't use GtkDialog or GtkWindow.

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -56,35 +59,58 @@ class FileChooserDialog {
...
+        libgtkui::SetGtkTransientForAura(GTK_WIDGET(dialog_),
+                                         parent_->GetNativeWindow());
+      else
+        SetGtkTransientForAura(parent_->GetNativeWindow());
+      if (GTK_IS_WINDOW(dialog_)) {
```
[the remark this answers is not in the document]
tristan957: Refer to previous follow-up. I could probably be more consistent and use GTK_IS_WINDOW everywhere or GTK_IS_WIDGET everywhere

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -94,11 +120,21 @@ class FileChooserDialog {
...
                      G_CALLBACK(OnUpdatePreviewThunk), this);
-    gtk_file_chooser_set_preview_widget(GTK_FILE_CHOOSER(dialog_), preview_);
+    gtk_file_chooser_set_preview_widget(dialog_, preview_);
   }
 
   ~FileChooserDialog() {
-    gtk_widget_destroy(dialog_);
+    if (GTK_IS_WIDGET(dialog_)) {
+      gtk_widget_destroy(GTK_WIDGET(dialog_));
+    } else {
```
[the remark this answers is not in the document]
tristan957: Refer

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -107,24 +143,35 @@ class FileChooserDialog {
...
                      this);
-    gtk_widget_show_all(dialog_);
+    if (GTK_IS_WIDGET(dialog_)) {
+      gtk_widget_show_all(GTK_WIDGET(dialog_));
+    } else {
+      void (*dl_gtk_native_dialog_show)(void*) = nullptr;
+      g_module_symbol(gtk_module_, "gtk_native_dialog_show",
+                      reinterpret_cast<void**>(&dl_gtk_native_dialog_show));
+      dl_gtk_native_dialog_show(static_cast<void*>(dialog_));
+    }
```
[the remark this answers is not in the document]
tristan957: Refer

shell/browser/ui/file_dialog_gtk.cc
tristan957: Refer

shell/browser/ui/file_dialog_gtk.cc
tristan957: I am referring to the function located in the Chromium code base: https://github.com/chromium/chromium/blob/master/chrome/browser/ui/libgtkui/gtk_util.cc#L136. Is this wrong?

shell/browser/ui/file_dialog_gtk.cc
tristan957: void* over a templated function?

tristan957: Adjusted to take a reference to a FileChooserDialog

shell/browser/ui/file_dialog_gtk.cc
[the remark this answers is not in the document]
TingPing: Using `GTK_IS_WINDOW` would be the simple solution, yeah.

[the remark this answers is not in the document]
TingPing: Well rather, `GTK_IS_DIALOG`, to be slightly more granular.

shell/browser/ui/file_dialog_gtk.cc
ckerr: OK, I was comparing it to the previous version of `shell/browser/ui/file_dialog_gtk.cc`. I don't mind the extra check being there but wanted to know _why_ it was there. Adopting from libgtkui WfM.

shell/browser/ui/file_dialog_gtk.cc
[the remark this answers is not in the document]
ckerr: Also, to be clearer about my `gtk_module_ != nullptr` suggestion: the version I was reviewing had the test later in the code logic:

```c++
if (gtk_module_ != nullptr && found ...
```

My suggestion was to move this test to before the `g_module_symbol()` call.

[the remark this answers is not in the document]
tristan957: @ckerr good point on the release. Static fields for the functions might be a better solution. If none of the fields are NULL, don't do a module search. Let's say I go through the process of dynamically calling all the functions, while saving them as static class functions at the same time, do the dynamically loaded functions reside in memory regardless if we create a GModule object or not? @TingPing

shell/browser/ui/file_dialog_gtk.cc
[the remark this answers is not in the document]
tristan957: Cool. This should have been fixed in the latest patches.

shell/browser/ui/file_dialog_gtk.cc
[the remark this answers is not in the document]
tristan957: Cool. I'll get this fix in place.

shell/browser/ui/file_dialog_gtk.cc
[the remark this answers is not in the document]
TingPing: It doesn't matter if its never released, the module is loaded as long as the process is a live anyway.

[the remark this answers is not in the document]
tristan957: @ckerr I will defer to your opinion on any static class variables (function pointers/module).

tristan957:
GTK_IS_DIALOG has now been implemented. Previously in this thread we talked about the `delete-event` needing to be changed. @ckerr what do you think of @TingPing's opinion in https://github.com/electron/electron/pull/19159#issuecomment-510262880?

gusbemacbe:
Hi @tristan957 !

I have tested your repo, and after `npm start`, it gave an error:

```
> electron@8.0.0-nightly.20190830 start /home/gusbemacbe/GitHub/electron-tristan
> node ./script/start.js

events.js:186
      throw er; // Unhandled 'error' event
      ^

Error: spawn /home/gustavo/GitHub/out/undefined/electron ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:264:19)
    at onErrorNT (internal/child_process.js:456:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
Emitted 'error' event on ChildProcess instance at:
    at Process.ChildProcess._handle.onexit (internal/child_process.js:270:12)
    at onErrorNT (internal/child_process.js:456:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  errno: 'ENOENT',
  code: 'ENOENT',
  syscall: 'spawn /home/gustavo/GitHub/out/undefined/electron',
  path: '/home/gustavo/GitHub/out/undefined/electron',
  spawnargs: []
}
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! electron@8.0.0-nightly.20190830 start: `node ./script/start.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the electron@8.0.0-nightly.20190830 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/gustavo/.npm/_logs/2019-10-14T02_46_32_243Z-debug.log
```

tristan957:
Dang. That message is not very helpful. Was this error thrown on application startup or was it thrown when you tried to open a file chooser dialog?

tristan957:
I will also try to fix the merge conflict tonight as well as give a cursory look at possibly seeing bad application logic on my end. That should give you a new build to test with.

If you test with `electron/master` does the script start fine? That can at least provide a baseline for seeing if it is my code or a bad commit on master.

MarshallOfSound:
@gusbemacbe that doesn't look like you've built electron on your machine.  You cant just clone the repo and run npm start.

Check out our building instructions and follow them, it is a non-trivial process

gusbemacbe:
@tristan957 !

Sorry for my long absence, I was in much hurry with the icons theme repository.

Yes, it was what I did: I gi-cloned your repo and ran the `npm install` and `npm start`. 

> Check out our building instructions and follow them, it is a non-trivial process

Which? Do you mean `npm install electron --save-dev [--save-exact]`?

MarshallOfSound:
https://github.com/electron/electron/blob/master/docs/development/build-instructions-gn.md

gusbemacbe:
I am out of luck. I am on Arch Linux. It does not have gclient installed. The `depot-tools-git`'s AUR is outdated. :-/

tristan957:
@gusbemacbe why not just clone the depot tool repo like the docs say to do?

tristan957:
@salifm Did you mean to approve this PR? There are still some holes in it imo.

tristan957:
PR has been updated to reflect electron/master. Over the weekend I started down the road of hopefully testing the PR. It is certainly not easy setting up a build environment for electron.

gusbemacbe:
I am looking forward to it, @tristan957.

@salifm, please review.

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -2,18 +2,22 @@
 // Use of this source code is governed by the MIT license that can be
 // found in the LICENSE file.
 
-#include <memory>
+#include <glib/gi18n.h>  // _() macro
```
TingPing: Nitpick: This doesn't seem to ever be used.

TingPing:
Looks good at a glance here.

tristan957:
Hitting an error while trying to build on Fedora 31:
```
ninja: Entering directory `out/Testing'
[7/32064] ACTION //chrome/services/file_u...ge_ids(//build/toolchain/linux:clang_x64)
FAILED: gen/chrome/services/file_util/public/mojom/file_util_service.mojom-shared-message-ids.h gen/chrome/services/file_util/public/mojom/safe_archive_analyzer.mojom-shared-message-ids.h 
python ../../mojo/public/tools/bindings/mojom_bindings_generator.py --use_bundled_pylibs generate -d ../../ -I ../../ -o gen --bytecode_path gen/mojo/public/tools/bindings --filelist=__chrome_services_file_util_public_mojom_mojom__generate_message_ids___build_toolchain_linux_clang_x64__rule..rsp --generate_non_variant_code --generate_message_ids -g c++ --scrambled_message_id_salt_path ../../chrome/VERSION
Traceback (most recent call last):
  File "../../mojo/public/tools/bindings/mojom_bindings_generator.py", line 556, in <module>
    sys.exit(main())
  File "../../mojo/public/tools/bindings/mojom_bindings_generator.py", line 551, in main
    return args.func(args, remaining_args)
  File "../../mojo/public/tools/bindings/mojom_bindings_generator.py", line 268, in _Generate
    RelativePath(filename, args.depth), [])
  File "../../mojo/public/tools/bindings/mojom_bindings_generator.py", line 201, in _GenerateModule
    imported_filename_stack + [rel_filename.path])
  File "../../mojo/public/tools/bindings/mojom_bindings_generator.py", line 212, in _GenerateModule
    ScrambleMethodOrdinals(module.interfaces, salt)
  File "../../mojo/public/tools/bindings/mojom_bindings_generator.py", line 131, in ScrambleMethodOrdinals
    sha256.update(interface.mojom_name)
TypeError: Unicode-objects must be encoded before hashing
[14/32064] CXX obj/net/tools/transport_se...ator/transport_security_state_generator.o
ninja: build stopped: subcommand failed.
```

Not sure how to fix

tristan957:
```
ninja: Entering directory `out/Testing'
[1/31631] SOLINK ./libvulkan.so
FAILED: libvulkan.so libvulkan.so.TOC 
python "../../build/toolchain/gcc_solink_wrapper.py" --readelf="readelf" --nm="nm" --sofile="./libvulkan.so" --tocfile="./libvulkan.so.TOC" --output="./libvulkan.so" -- ../../third_party/llvm-build/Release+Asserts/bin/clang++ -shared -Wl,-soname="libvulkan.so" -Wl,--fatal-warnings -fPIC -Wl,-z,noexecstack -Wl,-z,relro -Wl,-z,now -Wl,-z,defs -Wl,--as-needed -fuse-ld=lld -Wl,--icf=all -Wl,--color-diagnostics -m64 -Wl,-O2 -Wl,--gc-sections -rdynamic -nostdlib++ --sysroot=../../build/linux/debian_sid_amd64-sysroot -L../../build/linux/debian_sid_amd64-sysroot/usr/local/lib/x86_64-linux-gnu -L../../build/linux/debian_sid_amd64-sysroot/lib/x86_64-linux-gnu -L../../build/linux/debian_sid_amd64-sysroot/usr/lib/x86_64-linux-gnu -Werror -o "./libvulkan.so" @"./libvulkan.so.rsp"
Traceback (most recent call last):
  File "../../build/toolchain/gcc_solink_wrapper.py", line 151, in <module>
    sys.exit(main())
  File "../../build/toolchain/gcc_solink_wrapper.py", line 134, in main
    result, toc = CollectTOC(args)
  File "../../build/toolchain/gcc_solink_wrapper.py", line 44, in CollectTOC
    result, toc = CollectSONAME(args)
  File "../../build/toolchain/gcc_solink_wrapper.py", line 27, in CollectSONAME
    if 'SONAME' in line:
TypeError: a bytes-like object is required, not 'str'
[8/31631] CC obj/third_party/boringssl/boringssl/pem_lib.o
ninja: build stopped: subcommand failed.```

**Comment by tristan957** [comment]
I don't know how to make this build not stupid so I subbed out $python_path for the absolute path for python2. What is the canonical way to do this through the build system without manually editing files.

Obligatory: python2 in $current_year :/

**Comment by tristan957** [comment]
or Python 2 has nothing to do with it. I do not understand the build system. Guess I will see if it finished in the morning.

**Comment by tristan957** [comment]
I have a hunch the delete-even signal connection does not really do anything.

**Comment by tristan957** [comment]
Ok I am no genius, but if you clone this repo: https://github.com/electron/electron-quick-start

and change main.js to:

```
// Modules to control application life and create native browser window
const {app, BrowserWindow, dialog} = require('electron')
const path = require('path')

function createWindow () {
  // Create the browser window.
    width: 800,
    height: 600,
    webPreferences: {
      preload: path.join(__dirname, 'preload.js')
    }
  })

  // and load the index.html of the app.
  mainWindow.loadFile('index.html')

  // Open the DevTools.
  // mainWindow.webContents.openDevTools()
}

function createDialog() {
	x = dialog.showOpenDialog({ properties: ['openFile', 'multiSelections'] });
	x.then(y => console.log(y));
}

// This method will be called when Electron has finished
// initialization and is ready to create browser windows.
// Some APIs can only be used after this event occurs.
app.on('ready', createDialog)

// Quit when all windows are closed.
app.on('window-all-closed', function () {
  // On macOS it is common for applications and their menu bar
  // to stay active until the user quits explicitly with Cmd + Q
  if (process.platform !== 'darwin') app.quit()
})

app.on('activate', function () {
  // On macOS it's common to re-create a window in the app when the
  // dock icon is clicked and there are no other windows open.
  if (BrowserWindow.getAllWindows().length === 0) createWindow()
})

// In this file you can include the rest of your app's specific main process
// code. You can also put them in separate files and require them here.
```

You can get a file dialog to open up.

If you have a KDE desktop, I would very much appreciate notes on what happens, and if the Qt file chooser opens up. Make sure you have xdg-desktop-portal-kde(maybe that is the package) installed. You may or may not need to uninstall the gtk variant.

**Comment by parkerlreed** [comment]
I'm getting GTK dialog still 

image

**Comment by tristan957** [comment]
@parkerlreed did you compile my fork of electron? Is the electron binary you used the one you compiled yourself, or is it a different electron binary?

**Comment by parkerlreed** [comment]
@tristan957 I was just following the instructions in the linked comment from reddit. If there's extra steps I need to take, please let me know.

**Comment by tristan957** [comment]
@parkerlreed follow these instructions to build electron locally: https://www.electronjs.org/docs/development/build-instructions-gn.

When you get to this part: 
image
substitute my fork instead of the mainline repo.

Note this is an involved and time consuming process.

**Comment by TingPing** [comment]
> I have a hunch the delete-even signal connection does not really do anything.

It probably does something *but* its redundant with `response`, you shouldn't need to handle both.

**Comment by parkerlreed** [comment]
Well I'm just about ready to build...

```
[parker@wolfcola src]$ ninja -C out/Testing electron
ninja: error: loading 'build.ninja': No such file or directory
[parker@wolfcola src]$ ls out/Testing/
args.gn  browser_tests_apprtc.runtime_deps  build.ninja  build.ninja.d  gen  gen.runtime  obj  toolchain.ninja  xr_browser_tests.runtime_deps
```

**Comment by tristan957** [comment]
I wish I knew how to help you :/. All I can say is that I followed the instructions and at some point I got it to work.

**Comment by tristan957** [comment]
@parkerlreed if you have an x86_64 machine you can try this tar file of my out/Testing folder. No idea if it will work for you or not: https://send.firefox.com/download/da64bb8591efb4bd/#r55Em9aIOohNytTTPJx68g

**Comment by medmedin2014** [comment]
I'm not Javascript guy, but on Manjaro KDE and with my installed VSCode-v1.42.1 running on Electron-6.1.7 I get File Chooser similar to this 
Screenshot_20200307_112857

Operating System: Manjaro Linux 
KDE Plasma Version: 5.17.5
KDE Frameworks Version: 5.66.0
Qt Version: 5.14.1
Kernel Version: 5.5.7-1-MANJARO
OS Type: 64-bit
Processors: 2 × Pentium® Dual-Core CPU T4400 @ 2.20GHz
Memory: 5.7 GiB of RAM

**Comment by gusbemacbe** [comment]
@bpasero, check @medmedin2014's comment.

**Comment by tristan957** [comment]
Guys you have to build electron from my master branch. The fixes obviously won't be in previous releases because the code has not been merged.

**Comment by parkerlreed** [comment]
@tristan957 Ok so your archive, with the electron test repo and the replaced main.js. GTK chooser

image

**Comment by tristan957** [comment]
What am I missing here? What distro/version are you running @parkerlreed

**Comment by tristan957** [comment]
Guess I need to do some debugging because even on a GTK system the native file chooser should be chosen which in this case in the GtkFileChooser

**Comment by tristan957** [comment]
New tar file with an output statement for fun https://send.firefox.com/download/33301b06187cb18f/#-wbiYeK08tbx-DNHD3LxMQ

**Comment by zipy124** [comment]
I also appear to get GTK chooser, after building your fork, using the latest version of Kubuntu 19.xx 
screenshot

**Comment by zipy124** [comment]
> New tar file with an output statement for fun https://send.firefox.com/download/33301b06187cb18f/#-wbiYeK08tbx-DNHD3LxMQ

Retried with this tart vs my own compiled version, only difference is the output of "Creating native file chooser"

**Comment by tristan957** [comment]
Whelp there are no DBus messages being sent across the wire, so that would be the reason it does not work. Trying to ascertain why...

**Comment by gusbemacbe** [comment]
I also got like @parkerlreed and @zipy124. I use Arch Linux and at KDE. 

image

**Comment by TingPing** [comment]
Everybody trying to get it to use the portal you need to set the env var `GTK_USE_PORTAL=1`, be on KDE, and have `xdg-desktop-portal-kde` installed.

**Comment by tristan957** [comment]
Well fellow devs, it is done. I did not realize you had to use GTK_USE_PORTAL=1.
it_works

**Comment by tristan957** [comment]
So some things left to do are decide on the shared object file to load because Neon apparently doesn't ship a symlink to libgtk-3.so, so maybe other distributions do the same?

@ckerr what can be done to move this toward the next electron release?

**Comment by TingPing** [comment]
@tristan957 Perhaps don't ignore my comments: https://github.com/electron/electron/pull/19159/files#r312661062

**Comment by tristan957** [comment]
Are you talking about the memory header?

**Comment by TingPing** [comment]
My review was:

```suggestion
    gtk_module_ = g_module_open("libgtk-3.so.0", G_MODULE_BIND_LAZY);
```

The unversioned soname is often not installed and GTK3 is ABI frozen at this point.

TingPing:
You may want to use `gtk_file_chooser_set_current_name()` to avoid the window being called `xdg-desktop-portal-kde`

tristan957:
@TingPing I am a little confused as to why the GtkFileChooserNative constructor is not handling that. The first argument seems to be the title.

TingPing:
Also the reason the Open button is `gtk-open` is because of this: https://github.com/electron/electron/blob/b6246dcf122e584c672566877cb60d33dbc4705e/shell/browser/ui/gtk_util.cc#L19-L40

Stock items are deprecated in GTK3 and should not be used ever, it should be save to delete that entire conditional.

tristan957:
I will fix the so name here in a sec.

tristan957:
Ok I will delete those lines as well.

TingPing:
> @TingPing I am a little confused as to why the GtkFileChooserNative constructor is not handling that. The first argument seems to be the title.

Its always possible its a -kde specific bug if the -gtk portal works as expected. Neon will ofc have outdated versions of everything also.

tristan957:
It doesn't look like a title gets set for GTK at all. I am not sure why that is.

tristan957:
The comment from the GTK_STOCK code doesn't make sense. Is Electron not using gi18n for internationalization? What support are they expecting?

TingPing:
> The comment from the GTK_STOCK code doesn't make sense. Is Electron not using gi18n for internationalization? What support are they expecting?

Well they aren't translating them as-is. You can just wrap them with `_()` to do so.

This isn't a problem though, thats how you use GTK. You translate strings you show to users.

gusbemacbe:
@bpasero, @TingPing, @tristan957 and @parkerlreed 

> Everybody trying to get it to use the portal you need to set the env var `GTK_USE_PORTAL=1`, be on KDE, and have `xdg-desktop-portal-kde` installed.

I have already `xdg-desktop-portal-kde` installed before, but I enved `GTK_USE_PORTAL=1` and it did not open the KDE dialog. **But I understood perfectly what happened...** Electron that opens the GTK dialog is that Electron we installed via npm or natively from the official repository or PPA, and that Electron is still 7, so we had to change the command:

```                                                                                                                                                                                                                                                                                                                        $ ./electron --version
v10.0.0-nightly.20200226
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                $ GTK_USE_PORTAL=1 ./electron main.js
```

It opens the KDE dialog with success. The only solution is to upgrade to Electron 10, with the Tristan's merged PR. 

@bpasero, @roblourens, @sandy081 and @Tyriar 

Is there how to switch from 7 to 10 on VSCode? Your VSCode is still Electron 7.

gusbemacbe:
### Update

@bpasero and @Tyriar 

It seems that Electron 9 and Electron 10 do not work with VSCode OSS. :-/

bpasero:
I would appreciate if people could stop pinging me, I currently do not have the resources to follow this development, sorry.

gusbemacbe:
> I would appreciate if people could stop pinging me, I currently do not have the resources to follow this development, sorry. If you feel there is an issue with VSCode, please report it in our repository.

OK, I will stop. As for the issue, the issue is to build and run Code OSS with Electron 10, so it can not be asked on your repository, then better on Stack Overflow.

Tyriar:
Yes, you have also pinged people who have nothing to do with Electron's integration within VS Code. We have a process in place to adopt new Electron versions over time so we will hit this eventually. New Electron versions aren't meant to just work with VS Code as they almost always require some migration/fixes to go in.

tristan957:
@gusbemacbe slow it down homie. Rome was not built in a day. All I can do is push the PR forward, and eventually we will see it downstream. No rush!

tristan957:
So looks like alternative architecture builds are failing for whatever reason, but x86_64 builds good with tests passing. When this PR started older Linux distros were still supported by Electron. These are platforms where the GTK version of < 3.20. 3.20 is significant because that is the version with initial support for GtkFileChooserNative. We can lose all the GModule calls if we drop support for Linux distributions with out of date GTK versions.

https://www.electronjs.org/docs/tutorial/support#linux

This site says Ubuntu 12.04 is still supported. Ubuntu themselves don't even provide paid support anymore. 14.04 is supported until 2022 on paid support only. Fedora 21 is EOL. Debian 8 hit end of life in 2018. 14.04 has GTK version 3.18.9. If Electron's goal is to continue supporting 14.04, then I think the function pointers for the various GtkFileChooserNative calls should be moved to static members of the class, so that way calls aren't repeated and we can save on performance. Thoughts?

TingPing:
Personally I think the GModule usage is entirely fine and bypasses any worry about platform support. I don't think its worth changing.

polarathene:
Isn't the purpose of major versions to permit introducing breakage? So the only breakage that would occur is on existing apps that are still being updated and need a newer major version of Electron for some reason?

macOS 10.10 is min support listed there which came out late 2014, so probably safe to say the min Linux support could be raised up to that date too. I'd be doubtful that many linux desktop users who are using electron apps with regular updates(most likely not via the distro's core packages, but manual or community repositories, thus safer to rule out locked down enterprise environments) would be on newer releases of their OS.

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -179,34 +220,60 @@ class FileChooserDialog {
...
+  void SetGtkTransientForAura(aura::Window* parent) {
+    if (!parent || !parent->GetHost())
+      return;
+
+    void (*dl_gtk_native_dialog_set_transient_for)(void*, GtkWindow*) = NULL;
+    g_module_symbol(
+        gtk_module_, "gtk_native_dialog_set_transient_for",
+        reinterpret_cast<void**>(&dl_gtk_native_dialog_set_transient_for));
+    dl_gtk_native_dialog_set_transient_for(
+        dialog_, reinterpret_cast<GtkWindow*>(parent));
```
jhenstridge, line 254: This cast looks bogus.  An `aura::Window` is not a `GtkWindow`, and I doubt it's in-memory layout is similar enough for this to even have a chance of working.

polarathene:
Might be worth pointing out via git blame the date that support doc was last updated for Linux was 2 years ago, with some of it being over 5 years old. Note the "Until the release of Electron 2.0", which is a good indication of the maintenance of that section. 

The 3 distros and their [release versions that were cited are from 2015](https://github.com/electron/electron/blame/338a816ffd181116a373855ded8349ab8be7ed28/docs/tutorial/supported-platforms.md), the document also points out these are just verified distros that worked with the pre-built binaries of Electron when that document was written.

I am not sure what is being used now for building on Linux, seems like Dockerfiles existed at some point but have been removed, and CircleCI config only refers to linux in a generic way. Probably not still building off Ubuntu 12.04 though, which was the justification of the versioning in that section. Probably safe to assume Ubuntu 16.04 or greater is acceptable then..

shell/browser/ui/file_dialog_gtk.cc
tristan957: Let me look into this to confirm either way

tristan957:
@polarathene if 16.04 is the minimum supported version of Ubuntu then I can easily adjust some of the code

shell/browser/ui/file_dialog_gtk.cc
jhenstridge: One is a C++ class with a vtable, while the other is a C structure.

In the best case scenario, the `gtk_native_dialog_set_transient_for` call will fail on a `g_return_if_fail` check inside GTK from a runtime GType check (which should be evident from warnings on stderr).  Worst case, you cause a segfault.

tristan957: Weird I never hit that in my testing or compilation. I have an idea to get the underlying GtkWindow that I will confirm in #gtk tomorrow. But yea this is definitely not right at all, and I feel like I just totally missed something when I initially wrote that code a year ago.

```c
GApplication *app = g_application_get_default();
GtkWindow *parent = gtk_application_get_active_window(GTK_APPLICATION(app));
// do stuff with parent like set transient for.
```

Also this `SetGtkTransientForAura` is a hack imo because the regular `gtk::SetGtkTransientForAura` in `chrome/browser/ui/gtk/gtk_util.cc` has no concept of knowing about `GtkFileChooserNative` support. I dropped the X11/GdkWindow specific stuff from that function, and added back the necessary parts. I am a little confused as to why there is X11 code for setting `transient-for` on a `GtkFileChooserDialog` when the code could rely on GTK itself using `gtk_window_set_transient_for`. Does the electron project own that `gtk_util.cc` or does it belong to Chrome/Chromium? I really have no clue what files belong to whom if I am being honest.

jhenstridge: I suspect that there isn't an underlying GtkWindow for the aura::Window.  If there was, then old code would have called `gtk_window_set_transient_for` rather than twiddling X properties behind GTK's back.

tristan957: I had done some digging and found that an aura::Window (which I guess is a gfx:: NativeWindow on Linux) eventually was backed by a GtkWindow member variable called fake_window_, but maybe fake_window_ really is just fake.

Should I just remove the set transient call in the case of a GtkFileChooserNative?

polarathene:
@tristan957 From a look at the CircleCI "build-linux" results in this PR, [the first step provides environment info](https://app.circleci.com/pipelines/github/electron/electron/27199/workflows/edd2ef5d-9bba-4a78-a8a1-036f34bd3ffe/jobs/604984/parallel-runs/0/steps/0-0):

> Docker Engine Version: 18.09.6
> Kernel Version: Linux 6c2f531b4200 4.15.0-1077-aws #81-Ubuntu SMP Wed Jun 24 16:48:15 UTC 2020 x86_64 Linux

It's running off AWS and `81-Ubuntu` appears to [refer to Ubuntu 18.04 LTS](https://www.phoronix.com/forums/forum/software/distributions/1146405-canonical-releases-multipass-1-0-as-a-mini-cloud-on-your-workstation). Thus unless stated otherwise, if builds are being done on Ubuntu 18.04, that's the present baseline for support.

tristan957:
Well sweet. I'm gonna remove the GModule support out of the patch then, and just go for it with a reimplementation using GtkFileChooserNative.

polarathene:
I'm just going off what the platform support docs have said for Linux support, their date since last updated and referencing what the CI system is creating Linux binaries on. Not involved with development here, so I could be totally mistaken. 

Should be pretty safe to go with though, anything older is probably stuck on an old version of electron in constrained environments anyway.

tristan957:
Can always roll back the commit anyways.

tristan957:
Bogus arua::Window to GtkWindow cast is still in at the moment just because I am not sure about how to handle transiency.

ZanderBrown:
https://code.visualstudio.com/Docs/supporting/requirements#_platforms

> - Linux (Debian): Ubuntu Desktop 14.04, Debian 7
> - Linux (Red Hat): Red Hat Enterprise Linux 7, CentOS 7, Fedora 23

It's a bit silly but VSCode at least claims to support 14.04 so 2.5 years early for GtkFileChooserNative

Of course I'm really not convinced anyone is using up-to-date electron apps pre-16.04 or even 18.04

tristan957:
@ZanderBrown VSCode I believe uses Electron 7 with a move to Electron 8 in next month's release maybe, or very soon. I can't find out what Electron version 8's minimum platform requirements are. 14.04 is still supported for another 2 years too, but like I said I would bet people on that platform are using extremely locked down systems with very specific use cases.

ZanderBrown:
(I used vscode as an example because it's a fairly major product that tracks electron fairly closely + atom doesn't list requirements)

tristan957:
Gotcha. It would probably be nice to hear what stakeholders would want from projects like @slackhq, @microsoft, @atom, etc. Not really my place to say either way.

nornagon:
Leaving a note here that this is also relevant for Snapcraft. As far as I understand it, the native file chooser when running in a snap has the ability to poke holes in the snap confinement.

tristan957:
Yea both snap and flatpak support the spec, so this would have equal benefit for containerized electron apps as well :)

soredake:
Any progress on this?

tristan957:
@soredake I come back to the PR and rebase it as necessary. Really I just need someone to finish the review. Technically the code is sound I think especially in its current form. Would be nice to have more testing around the PR.

soredake:
@TingPing @jhenstridge @salifm @ckerr

tristan957:
Merge conflict resolved :). GtkNativeDialog has no notion of present with time, so I just dropped that bit of code.

tristan957:
I'll see if I can reach out through GNOME again to help move this along. People know people. If any other people have connections to electron developers, try to reach out to them in constructive and positive ways :). Funny how time flies.

tristan957:
I also removed the WIP tag. Maybe that will attract attention since the PR really isn't WIP any more.

tristan957:
I also removed just some absolutely blatant UB. Why it compiled/why it ever worked is an exercise best left to no one. At the moment it seems that the file chooser is not transient-for anything at the moment...There are no GtkWindows to be transient off of, not is a GtkNativeDialog backed by a GdkWindow, so dead in the water on that front.

MarshallOfSound:
@tristan957 Can you outline what the actual OS version support we're going to lose here is.

> This PR removes support for platforms that don't support GTK versions >= 3.20.

I assume this translates to certain version of ubuntu / redhat.  Wondering how far back we will still support.  In the same area of discussion, is there _any_ way to support the old way in addition to the new way?

tristan957:
@MarshallOfSound yes. The PR previously had a way to support both. But my research indicated that later versions of electron no longer advertise support for platforms who have GTK <= 3.20. My research could be wrong however.

GTK 3.20 was released on March 21, 2016.

https://repology.org/project/gtk/versions has good information.

3.20 is supported in Debian Stable, CentOS 7,  (which typically means RHEL 7), RHEL 8 (work machine can confirm), Ubuntu >= 16.10. The only questionable distro I notice is Ubuntu 16.04. Does electron v9+ support 16.04?

MarshallOfSound:
> Does electron v8+ support 16.04?

Yes 😄 I believe our current supported platforms are inherited directly from Chromium.

> 64-bit Ubuntu 14.04+, Debian 8+, openSUSE 13.3+, or Fedora Linux 24+

So dropping support for 16.04 and lower is not gonna fly rn.

This PR has been up for literally over a year though so I want to get a crystal clear set of steps for getting it over the line, based on what you've said and the current code I think all we need is for you to add backwards compat support for the "old" way of doing dialogs to support Ubuntu <= 16.04

Once you've done that I'll 👀 and merge.  This probably won't be backported due to risk but we can definitely get this thing landed

tristan957:
@MarshallOfSound thanks for engaging with me. I will add back in the backwards compat part in such a way that the commit could be reverted in the future very easily when support for those cobweb-lined distros is dropped. Look for an update later tonight.

tristan957:
@MarshallOfSound looks like the Linux builds are now working as they were previously.

MurzNN:
This is really needing feature, @ckerr can you please review and approve this PR? Or mention other, that can do this? Thanks!

Natetronn:
Would love to see this as well.

tristan957:
@Natetronn this is the PR that will enable KDE file dialog FYI.

Natetronn:
@tristan957 thanks for working on this! Yeah, I'm hoping it can be merged sometime soon. Having my Dolphin file manager bookmarks back, although in the grand scheme of things is a tiny convenience, day in and day out it's a features that is super helpful.

tristan957:
@MarshallOfSound I think it would be great to try to get this merged before the holidays if possible,

tristan957:
I gave this some thought again, and I think reaching for feature test macros instead of dynamic calls to the shared object library would save a whole lot of fuss and performance. I will iterate on this soon.

tristan957:
So I finished what I talked about in my previous comment just now, but I am keeping it local for now. I think that solution depends on how Electron is built? What if Electron is built on a platform with GTK  >= 3.20 but distributed to platforms with GTK < 3.20?

So many questions...

genodeftest:
> So I finished what I talked about in my previous comment just now, but I am keeping it local for now. I think that solution depends on how Electron is built? What if Electron is built on a platform with GTK >= 3.20 but distributed to platforms with GTK < 3.20?

Is this use case realistic at all? Even the oldstable versions of Ubuntu LTS (18.04) and Debian (9) have Gtk 3.22 ([source](https://distrowatch.com/dwres.php?firstlist=ubuntu&secondlist=debian&firstversions=6&resource=compare-packages&secondversions=3)). You would have to use Ubuntu 16.04 LTS, 14.04 LTS, Debian 7 or 8. This is really really old stuff.

tristan957:
Yes, this has been discussed on this PR already.

ninja-:
so we're using xdg desktop portals to display dialogs, but all we get as a result is the filepath, right?
I was interested in implementing opening files in sandbox using xdg desktop portals so we could finally remove full access to /home in snaps/flatpak.

tristan957:
@ninja- yep file paths are strings. Even the high level Electron APIs seem to only return a string. You then determine what you want to do with the string.

ZanderBrown:
You automatically get perms on files opened via GtkFileChooserNative so it should Just Work:tm:

ninja-:
@ZanderBrown really? sounds smart. I was under impression that it can only pass open file descriptors :thinking:

jebw:
I think inside a sandbox it'll go via the documents portal and return a
custom fuse path providing access to the requested file. The documents
portal handles setting that up transparently and the native file chooser
integrates with that automatically.

Works the same for snaps as well.

https://snapcraft.io/docs/xdg-desktop-portals

tristan957:
@MarshallOfSound anything I can do on my end to make this PR easier for someone who wants to review it? Or is this work being superceded by upstream Chromium at this point?

tristan957:
The JS packaging story sucks, so if that change to the package-lock.json needs to be fixed please let me know. Commits can also be squashed on my end as well.

spec/yarn.lock
```diff
@@ -229,7 +229,7 @@ dashdash@^1.12.0:
     safe-buffer "^5.1.1"
     xml2js "^0.4.17"
   optionalDependencies:
-    abstract-socket "github:saghul/node-abstractsocket#35b1b1491fabc04899bde5be3428abf5cf9cd528"
+    abstract-socket "^2.0.0"
```
ckerr, line 232: I'm not sure what's going on with that the old version of this line, that old version seems strange to me. But unless this change is on-topic to this PR, it should be handled separately

ckerr:
> The JS packaging story sucks, so if that change to the package-lock.json needs to be fixed please let me know. Commits can also be squashed on my end as well.

Agreed wrt reverting the package-lock.json change. That old line looks pretty strange to me but if it's not necessary for native dialogs maybe it should be in a separate PR.

No need for you to squash your branch; that gets done when the branch is merged to master!

tristan957:
@ckerr thanks for your feedback. I will iterate on it tonight. I don't think std::function buys anything except maybe clarity. I will use the syntax you described above. I will look into the base::Optional approach you suggest.

tristan957:
Your feedback was valuable @ckerr. Thanks

tristan957:
Electron just doesn't like building on my Fedora 33 because of Python 2 :/

ssokolow:
@jebw Yeah. It uses a fuse mount:

> However, the file chooser portal works a bit differently than the home interface:
>
>    Files are fuse-mounted to `/run/user/<uid>/doc/<hash>/` in order to give your application access to it. So the path your application sees is different from the path a user chose, even though both are the same file.
>
> -- https://snapcraft.io/docs/xdg-desktop-portals#heading--portal-vs-home

darkdragon-001:
> Electron just doesn't like building on my Fedora 33 because of Python 2 :/, but the PR should be good now for another review

Possible to build in Docker?

tristan957:
Good point about building in a container.

tristan957:
Thinking about switching to KDE Plasma, so I may have some skin in getting this over the hump now.

eclipseo:
Any chance to get this finally reviewed?

Natetronn:
> Thinking about switching to KDE Plasma, so I may have some skin in getting this over the hump now.

Coming from Gnome before it, KDE Plasma has been a breath of fresh air. Of course, to each their own.

sparky8251:
Really hoping to see this merged soon. Got a handful of electron apps that are my last stubborn holdout on using a sane file picker for my desktop. This will thankfully quickly land in most of them since they are open source :)

shell/browser/ui/gtk_util.cc
```diff
@@ -14,30 +14,16 @@
...
 
-// Copied from L40-L55 in
-// https://cs.chromium.org/chromium/src/chrome/browser/ui/libgtkui/select_file_dialog_impl_gtk.cc
-#if GTK_CHECK_VERSION(3, 90, 0)
-// GTK stock items have been deprecated.  The docs say to switch to using the
-// strings "_Open", etc.  However this breaks i18n.  We could supply our own
+// Currently i18n is broken for file dialog labels. We could supply our own
 // internationalized strings, but the "_" in these strings is significant: it's
-// the keyboard shortcut to select these actions.  TODO: Provide
+// the keyboard shortcut to select these actions. TODO: Provide
```
ckerr, line 19: This change seems unrelated to the native dialog patch.

This hack seems to [still exist in upstream Chromium](https://source.chromium.org/chromium/chromium/src/+/master:ui/gtk/select_file_dialog_impl_gtk.cc;l=37) so I've a slight preference to keeping consistency with upstream.

Also -- if I'm reading it right, this change seems to go from "here is an ugly hack to avoid breaking i18n on file dialog button labels on systems using GTK < 3.90" to "file dialog button labels are broken < 3.90". Is that correct?

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -66,30 +155,42 @@ class FileChooserDialog {
     else if (action == GTK_FILE_CHOOSER_ACTION_OPEN)
       confirm_text = gtk_util::kOpenLabel;
 
-    dialog_ = gtk_file_chooser_dialog_new(
-        settings.title.c_str(), nullptr, action, gtk_util::kCancelLabel,
-        GTK_RESPONSE_CANCEL, confirm_text, GTK_RESPONSE_ACCEPT, NULL);
+    InitGtkFileChooserNativeSupport();
+
+    if (supports_gtk_file_chooser_native.value_or(false)) {
```
ckerr, line 160:
```suggestion
    if (*supports_gtk_file_chooser_native) {
```

(here and below) I might be misreading but don't see any code path where the fallback value in `value_or(fallback_value)` will ever be  reached?

ckerr (commented):
I had a couple of questions, but I still like this feature and it's set unreviewed for too long. Let's get this over the hump & merged for 12.0.0

shell/browser/ui/file_dialog_gtk.cc
tristan957: I think you are right. I will get to this change after work.

shell/browser/ui/gtk_util.cc
tristan957: After some discussion in #gtk, I now understand the hack. Since GTK 4.0 has now been released, does it make sense to either update the CHECK_VERSION macro to reflect this, or B drop this ability to compile against 4.0 successfully. It seems a little weird to me, because I am not checking to make sure the new GTK code works against 4.0 unless you think I should do that.

tristan957:
@ckerr what date should I have this ready to go by for Electron 12.0 because I too would very much like to get it in in time for that? Just been away from computers for the past week or so.

tristan957:
Comments are now fixed :). Thanks @ckerr. Please let me know what else I can do to push this along.

soredake:
Please don't stall again :(

AchmadFathoni:
We were on the verge of greatness, we were this close:v:

tristan957:
Hmm it looks like this PR missed the electron 12 release. @ckerr is there anything to do to make it into a point release? If not, is this a candidate for 13, or is this PR just being dropped for potential upstream work in Chromium. I'm a little disappointed this didn't make it into 12. This PR also increases the flexibility of electron applications in sandboxed environments on Linux like Snap and Flatpak, so there are a lot of stakeholders here.

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -101,14 +202,23 @@ class FileChooserDialog {
     if (!settings.filters.empty())
       AddFilters(settings.filters);
```
refi64, line 203: `AddFilters` uses `gtk_file_filter_add_custom`, which is skipped when using the portal. This actually breaks things badly, as GTK will include a filter list when it sees a filter is present, but it ends up empty due to using all the custom filters, resulting in the portal rejecting the call altogether.

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -101,14 +202,23 @@ class FileChooserDialog {
...
 
+    // GtkFileChooserNative does not support preview widgets through the
+    // org.freedesktop.portal.FileChooser portal. In the case of running through
+    // the org.freedesktop.portal.FileChooser portal, anything having to do with
+    // the update-preview signal or the preview widget will just be ignored.
```
refi64, line 208: Is there a particular reason this even bothers to create the image widgets? Seems a bit like a bit of a waste, since it's not doing anything here...

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -325,10 +457,10 @@ void ShowOpenDialog(const DialogSettings& settings,
 
 bool ShowSaveDialogSync(const DialogSettings& settings, base::FilePath* path) {
   FileChooserDialog save_dialog(GTK_FILE_CHOOSER_ACTION_SAVE, settings);
-  save_dialog.SetupSaveProperties(settings.properties);
```
refi64, line 328: Why was this removed but `SetupOpenProperties` is still present?

refi64 (changes requested):
I'm not intimately familiar with Electron's internals, but from some basic experience with Chromium & co, I can definitely say AddFilters needs some changes for it to work!

Side note: I saw earlier that there was a recommendation for GModule to be used instead of the `dl*` family. I'm not sure if it's actually an improvement though; at least off the top of my head, since Electron already links to GTK, you could just do:

```c
void* handle = dlopen(nullptr, RTLD_NOW);

dl_gtk_native_dialog_set_modal = static_cast<dl_gtk_native_dialog_set_modal*>(dlsym(handle, "gtk_native_dialog_set_model"));
// ...
```

which means you could strip out all the `if`s and just leave the `&&` check at the end.

I'm not 100% sure on this, but I want to say Chromium *already* has `-ldl`, since that's how it loads udev, libpulse, etc.

shell/browser/ui/file_dialog_gtk.cc
tristan957: @refi64 it was left in because we could be linking against a version of GTK that doesn't support GtkFileChooserNative.

shell/browser/ui/file_dialog_gtk.cc
tristan957: So you're saying to check "if GtkFileChooserNative, then don't AddFilters"?

shell/browser/ui/file_dialog_gtk.cc
tristan957: This seems in advertant. I will add it back.

shell/browser/ui/file_dialog_gtk.cc
refi64: Ah apologies, I mis-stated. What I meant to say was that it felt weird to still create the image widgets themselves *when a native dialog is being used*, even though it's ignored we still end up with a useless widget in memory.

tristan957: I'll take a look at the logic and rework it. Thanks for your input.

shell/browser/ui/file_dialog_gtk.cc
refi64: Not quite. AddFilters itself needs to be adjusted to use the normal string-based GtkFileFilter functions to add the patterns, instead of using add_custom, when using a native dialog. I mentioned it here since, err, it's the only part of the code related to filters that I could put a comment on 😅 though I definitely should have clarified more.

tristan957: I'll try to review what you have said to the PR. May have to come back to ask more questions :).

zcbenz (commented):
I'm good with this PR, once the reviews raised by tristan957 are solved I think we are good to merge it.

tristan957:
@refi64 thanks for taking the time to look at this PR. I resolved two of your concerns.

Could you link to docs where I can read about `gtk_file_filter_add_custom` not working in the portal? I am currently investigating in IRC. And then for your `dlopen` suggestion, would I need to link against `dl` to support it or does Electron already link against it? Can't find any mention of it in the source code, and I don't really read bazel...

tristan957:
Nevermind found the docs.

tristan957:
And I am an idiot who didn't read your comment fully. Chromium does link against libdl, but wondering if electron should be explicit in this manner.

tristan957:
Failures on Windows and Mac seem unrelated as usual, but a successful compilation was had.

tristan957:
@zcbenz thoughts on merging?

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -250,35 +354,36 @@ void FileChooserDialog::AddFilters(const Filters& filters) {
...
-      auto file_extension = std::make_unique<std::string>("." + extension);
-      gtk_file_filter_add_custom(
-          gtk_filter, GTK_FILE_FILTER_FILENAME,
-          reinterpret_cast<GtkFileFilterFunc>(FileFilterCaseInsensitive),
-          file_extension.release(),
-          reinterpret_cast<GDestroyNotify>(OnFileFilterDataDestroyed));
+      std::string file_extension{"*." + extension};
+      // guarantee a pure lowercase variant
+      for (auto& c : file_extension)
+        c = std::tolower(c);
```
zcbenz, line 360: Can you use `base::ToLowerASCII` to convert the whole string directly?

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -250,35 +354,36 @@ void FileChooserDialog::AddFilters(const Filters& filters) {
...
+      gtk_file_filter_add_pattern(gtk_filter, file_extension.c_str());
+      // guarantee a pure uppercase variant
+      for (auto& c : file_extension)
+        c = std::toupper(c);
```
zcbenz, line 364: And `base::ToUpperASCII` here.

shell/browser/ui/file_dialog_gtk.cc
```diff
@@ -250,35 +354,36 @@ void FileChooserDialog::AddFilters(const Filters& filters) {
...
-    gtk_file_chooser_add_filter(GTK_FILE_CHOOSER(dialog_), gtk_filter);
+    gtk_file_chooser_add_filter(dialog_, gtk_filter);
   }
 }
 
-void FileChooserDialog::OnUpdatePreview(GtkWidget* chooser) {
-  gchar* filename =
-      gtk_file_chooser_get_preview_filename(GTK_FILE_CHOOSER(chooser));
+void FileChooserDialog::OnUpdatePreview(GtkFileChooser* chooser) {
+  assert(!*supports_gtk_file_chooser_native);
```
zcbenz, line 374: We usually use `CHECK` in Electron's codebase.

zcbenz (approved):

zcbenz:
This PR requires at least another API approval from @electron/wg-api.

refi64:
> libdl also isn't cross-platform from previous discussions in this thread.

The GTK file dialog is only used on Linux regardless though, right? So it wouldn't really make much of a difference.

Of course, is mostly just a nit anyway, but I do think it cleans up the loading code a bit.

tristan957:
Is the GTK backend not used on platforms like the BSDs as well?

refi64:
Probably, but they're all lumped into the "Linux" platform support that uses libdl in many other places. In other words, this shouldn't be used on any platform that doesn't *already* use libdl for something else.

tristan957:
@refi64 I think I will poke around for people who know about Chromium's port to BSDs because if I can make the surface area of which they have to apply custom patches, I would like to make life easier for them. Yea the code is ugly, but at least it is portable.

tristan957:
@zcbenz once again thank you for your review

zcbenz:
After discussion on @electron/wg-api, we think this PR does not need API approval, so I'm going to merge it.

On the topic of libdl, Chromium and Electron internally uses dlopen to load dynamic libraries on POSIX systems (see https://source.chromium.org/chromium/chromium/src/+/master:base/native_library_posix.cc), and most code would use the helper functions defined in `base/native_library.h` instead of calling libdl directly.

For this PR I think using either `base/native_library.h` or GModule is fine since code size and performance are basically the same.

tristan957:
Cool. Let me know if it should be changed at some point in the future

ransome1:
Hi @tristan957 . That is great news thanks a lot for being so persistent on this. 

I would love to use the native gtk filer chooser for my Electron app. But quite frankly I have no idea how to setup my app so it starts using it. Is there any documentation how to enable this new file chooser in my JavaScript?

tristan957:
Should be enabled by default in Flatpaks where applications use the native file chooser. All you have to do is upgrade electron when it makes it into a new release.

ransome1:
Ah ok, it was my understanding, that it already has been published with the latest Electron release.

lionirdeadman:
Will this be backported since it's tagged semver/minor? I think it would be good to do so since applications are slow to adopt new electron versions.

Also, it seems that this is not included in the Electron 13 beta builds, is this intentional? It's only been included in 14 nightlies as of now.

ransome1:
Hi @tristan957, do you already have an idea on when the Electron guys want to put your PR into release?

tristan957:
Nope!

ransome1:
Too bad, I'm so excited ;)

lionirdeadman:
cc : @zcbenz 

Are you able to answer the questions above? Since you merged this, I assume you might know if it will be backported. I apologize if it's not appropriate, I'm not familiar with the etiquette here.

MarshallOfSound:
Backports are done on a case by case basis and normally we only backport things that are relatively small, easy to support, or are well tested.

Unfortunately this PR doesn't meet any of those requirements (from my perspective) so if it was put up for backport I'd probably vote against it 🤷 This will be released with Electron 14 currently.

tristan957:
Do you mean 13?

MarshallOfSound:
Nope, the 13-x-y release branch was cut a long time ago.  Things only go there that are backported.  Currently this PR has only landed in master which will become the 14-x-y release branch

tristan957:
Man, that is super unlucky. Oh well.

ransome1:
:/

@MarshallOfSound, from experience, what would you roughly estimate, when the version bump to 14 can be realistic?

MarshallOfSound:
Our release timeline is quite predictable --> https://www.electronjs.org/docs/tutorial/electron-timelines

MatthewForrester:
Thank you to everyone who has worked on this. It will be much appreciated by KDE users.

squalou:
Hi,
I'm wondering : has it landed in 14.0.0 released yesterday ? 
(I've beeing trying to use it without a great success, with portal-kde installed, GTK_USE_PORTAL set to 1. )

vchernin:
No one’s been able to figure out the cause of this feature seemingly not working on Ubuntu (or in Flatpak) so I’ve opened a follow up: https://github.com/electron/electron/issues/31258

If anyone has ideas on what might be the problem please share them.

PR #19179, opened 2019-07-09
refactor: simplify RemoveFromLoginItems

### LINKED ISSUE #19160: crash in 'correctly sets and unsets the LoginItem'
cc @codebytere

```
4   ???                                 0x00007f98dd127038 0x0 + 140294520729656
5   SharedFileList                      0x00007fff36838f62 -[SFLLoginItemList _fetchItems] + 364
6   SharedFileList                      0x00007fff36837942 __49-[SFLLoginItemList handleListChangeNotification:]_block_invoke + 103
```

https://circleci.com/gh/electron/electron/255344

nornagon:
#### Description of Change
Refactors RemoveFromLoginItems to not iterate the SharedFileList twice. I have a feeling this fixes #19160 but not 100% sure as i couldn't reproduce the crash locally.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/shell/browser/browser_mac.mm
+++ b/shell/browser/browser_mac.mm
@@ -222,13 +222,14 @@ Browser::LoginItemSettings Browser::GetLoginItemSettings(
   return settings;
 }
 
-// copied from GetLoginItemForApp in base/mac/mac_util.mm
-LSSharedFileListItemRef GetLoginItemForApp() {
+void RemoveFromLoginItems() {
+  // logic to find the login item copied from GetLoginItemForApp in
+  // base/mac/mac_util.mm
   base::ScopedCFTypeRef<LSSharedFileListRef> login_items(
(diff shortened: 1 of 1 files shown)


shell/browser/browser_mac.mm
```diff
@@ -242,42 +243,8 @@ LSSharedFileListItemRef GetLoginItemForApp() {
     if (!error && item_url_ref) {
```
codebytere: how here are we dealing with the case where a given login item doesn't exist? we previously guarded for that with the now-removed function and it's not clear to me that `LSSharedFileListItemRemove` accounts for that innately.

nornagon:
hm... won't the loop just not hit the `LSSharedFileListItemRemove` call if the item doesn't exist?

codebytere:
If we bail with no login items list i think that should cover it 🤔

Issue #13249, opened 2018-06-14
WebView history is broken if a webpage loads a navigation upon goBack() being called


charlesclu:
* Electron Version: 2.0.2
* Operating System (Platform and Version): macOS 10.13.2
* Last known working Electron version: N/A

**Expected Behavior**
WebView history remains correct after navigating with goBack()

**Actual behavior**
Some webpages, such as `https://twitter.com/`, load a navigation upon `goBack()` being called, (in this case it is a navigation to `https://twitter.com/`), and this navigation triggers `this.webContents.on('navigation-entry-commited')` before the navigation from `goBack()` does, resulting in the wrong blocks code being executed and the WebView history to be messed up.

Here's the relevant code in navigation-controller.js:
```
this.webContents.on('navigation-entry-commited', (event, url, inPage, replaceEntry) => {
      if (this.inPageIndex > -1 && !inPage) {
        // Navigated to a new page, clear in-page mark.
        this.inPageIndex = -1
      } else if (this.inPageIndex === -1 && inPage && !replaceEntry) {
        // Started in-page navigations.
        this.inPageIndex = this.currentIndex
      }
      if (this.pendingIndex >= 0) {
        // Go to index.
        this.currentIndex = this.pendingIndex
        this.pendingIndex = -1
        this.history[this.currentIndex] = url
      } else if (replaceEntry) {
        // Non-user initialized navigation.
        this.history[this.currentIndex] = url
      } else {
        // Normal navigation. Clear history.
        this.history = this.history.slice(0, this.currentIndex + 1)
        this.currentIndex++
        this.history.push(url)
      }
    })
```
```
NavigationController.prototype.goBack = function () {
    if (!this.canGoBack()) {
      return
    }
    this.pendingIndex = this.getActiveIndex() - 1
    if (this.inPageIndex > -1 && this.pendingIndex >= this.inPageIndex) {
      return this.webContents._goBack()
    } else {
      return this.webContents._loadURL(this.history[this.pendingIndex], {})
    }
  }
```
And here's a log of what happens, with some edits to point out where it all goes wrong:
```
this.history: [ 'about:blank', 'https://www.google.com/?gws_rd=ssl', 'https://twitter.com/' ]
goBack() called:
    this.pendingIndex changed to 1
    loadURL(https://www.google.com/?gws_rd=ssl) called
this.webContents.on('navigation-entry-commited' event triggered: <--- this should be from calling loadURL(https://www.google.com/?gws_rd=ssl), but it's not!
[...]

PR #19189, opened 2019-07-10
fix: revert use of IPC helpers for history due to failing test

miniak:
#### Description of Change
I've started seeing this test failing after #17948, which could be caused by some timing issue:
```
<webview> tag <webview>.clearHistory() should clear the navigation history
```
Let's revert the history manipulation part for now.
Could also be related to this issue #13249

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes


nornagon:
I'm pretty sure that test has been flaky for a while. See https://electron-mission-control.herokuapp.com/flakes.html

PR #19191, opened 2019-07-10
doc: correct Windows value for DesktopCapturerSource's id

skatpgusskat:
#### Description of Change
I updated Markdown Documentation about DesktopCapturerSource. DesktopCapturerSource's Id is not just a random generated number, it is a handle number to the window.

#### Checklist

- [x] `npm test` passes

#### Release Notes
Add accurate info of DesktopCapturerSource's Id
Notes: 

You can test it with [`EnumWindows` API](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-enumwindows).
[Here is the source code in C# which can show current visible windows](https://github.com/skatpgusskat/overview/issues/3#issuecomment-509158945)

image

In addition, the id is not just `window:XX`, but it is `window:XX:YY`. I have no idea about YY yet.

### DIFF
--- a/docs/api/structures/desktop-capturer-source.md
+++ b/docs/api/structures/desktop-capturer-source.md
@@ -4,6 +4,7 @@
   `chromeMediaSourceId` constraint when calling
   [`navigator.webkitGetUserMedia`]. The format of the identifier will be
   `window:XX` or `screen:XX`, where `XX` is a random generated number.
+  In Windows, `XX` of `window:XX` is a windows handle to the window.
 * `name` String - A screen source will be named either `Entire Screen` or
   `Screen <index>`, while the name of a window source will match the window
   title.


skatpgusskat:
I see. That's true that we don't know how chromium will be changed later.
One of my suggestion is, "change documentation at that time when chromium change that value".

I mean, for now, it is very true that DesktopCapturerSource's Id is window handle Id (hwnd).
We don't have to close our users choices or ideas. If they know "this is window id" then maybe they can make something from that.

Maybe we can put some comments, "It would be changed later when chromium team change that", isn't it?

skatpgusskat:
> I agree we shouldn't explicitly document it as window ID, we won't be able to fix it if Chromium decides to change the meaning.

Yes but we can change our docs at that time, because we won’t update chromium version; previous version’s user won’t get any affect.

MarshallOfSound:
> We don't have to close our users choices or ideas. If they know "this is window id" then maybe they can make something from that.

The point is we don't want users to make that choice, this to all intents and purposes should be documented as a random number so that we don't introduce a random breaking change when Chromium decides to change it.

Based on my perspective and @zcbenz 's comment above I'm closing this one out, thanks for the PR but at this time I don't think it's appropriate 👍

Issue #19185, opened 2019-07-10
docs parser not generating types for BrowserWindow properties


nornagon:
e.g. `BrowserWindow.minimizable` is not present in `electron.d.ts`. Unsure if this is a docs parser issue or a type generator issue. cc @codebytere @MarshallOfSound

> [...] error TS2551: Property 'minimizable' does not exist on type 'BrowserWindow'. Did you mean 'isMinimizable'?

codebytere:
@nornagon i _think_ this may be a docs parser issue; it seems to be dependent on the indentation of the markdown file headers (it works when i change each property from `####` to `###`) so i'll pop up a PR to address that now but i think a more robust solution might just be to make the docs-parser less brittle about that.

PR #19196, opened 2019-07-10
docs: differentiate static/instance class props

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/19185.

In the latest update of our parser, classes are expected to differentiate between static and instance methods and properties and not all of our documentation had been updated accordingly. This PR performs those updates.

cc @MarshallOfSound @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/styleguide.md
+++ b/docs/styleguide.md
@@ -116,7 +116,7 @@ Using the `Session` and `Cookies` classes as an example:
 
 ### session.fromPartition(partition)
 
-## Properties
+## Static Properties
 
 ### session.defaultSession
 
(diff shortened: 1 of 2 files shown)


codebytere:
Would that be something worth considering adding to the docs parser @MarshallOfSound? I undid it for this PR to unblock it but i would think that'd be a better solution long term perhaps

docs/styleguide.md
```diff
@@ -112,11 +112,11 @@ Using the `Session` and `Cookies` classes as an example:
 ```markdown
 # session
 
-## Methods
+## Static Methods
```
zcbenz: `session.fromPartition` is a method of the `session` module, instead of a static method of `Session` class. I'm not sure if it should be marked as `Static` here.

codebytere: yeah, i think it may be better to use a different module example to illustrate this point

PR #19202, opened 2019-07-10
chore: deprecate `Tray.setHighlightMode`

DeerMichel:
#### Description of Change
According to [this verdict](https://github.com/electron/governance/blob/master/wg-releases/meeting-notes/2019-06-26.md#agenda) we deprecate `Tray.setHighlightMode` in v6 and remove it in v7. This is necessary bc v7 will ship with a new Tray implementation on macOS which doesn't support changing the highlighting behavior anymore (see #18981 for details).

cc @codebytere @miniak 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Deprecated `Tray.setHighlightMode` (will be removed in v7 without replacement).

### DIFF
--- a/docs/api/breaking-changes.md
+++ b/docs/api/breaking-changes.md
@@ -152,6 +152,17 @@ app.enableMixedSandbox()
 
 Mixed-sandbox mode is now enabled by default.
 
+### `Tray`
+
+Under macOS Catalina our former Tray implementation breaks.
+Apple's native substitute doesn't support changing the highlighting behavior.
+
+```js
+// Deprecated
+tray.setHighlightMode(mode)
+// API will be removed in v7.0 without replacement.
+```
+
 ## Planned Breaking API Changes (5.0)
 
 ### `new BrowserWindow({ webPreferences })`
--- a/docs/api/tray.md
+++ b/docs/api/tray.md
@@ -216,6 +216,8 @@ Returns `String` - the title displayed next to the tray icon in the status bar
 
 Sets when the tray's icon background becomes highlighted (in blue).
 
+**[Deprecated](breaking-changes.md#tray)**
+
 **Note:** You can use `highlightMode` with a [`BrowserWindow`](browser-window.md)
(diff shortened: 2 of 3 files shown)


docs/api/tray.md
```diff
@@ -216,6 +216,8 @@ Returns `String` - the title displayed next to the tray icon in the status bar
 
 Sets when the tray's icon background becomes highlighted (in blue).
 
+**[Deprecated](https://github.com/electron/electron/blob/6-0-x/docs/api/breaking-changes.md#removal-of-deprecated-sethighlightmode-method-on-tray): will be removed in v7 without replacement.**
```
codebytere: Let's instead say:
```
**[Deprecated](/link/to/tray/header/in/docs/api/breaking-changes.md)**
```

for cleanliness here and consistency with our other Deprecation notices in docs.

codebytere: otherwise LGTM, i'll ✅ as soon as that's updated.

docs/api/breaking-changes.md
```diff
@@ -152,6 +157,14 @@ app.enableMixedSandbox()
 
 Mixed-sandbox mode is now enabled by default.
 
+### `Tray`
+
```
codebytere:
```
Under macOS Catalina our former Tray implementation breaks.
Apple's native substitute doesn't support changing the highlighting behavior.
```
should go here and the above section should be removed

docs/api/tray.md
DeerMichel: all right, although i've also encountered [this variations](https://github.com/electron/electron/blob/master/docs/api/download-item.md#downloaditemgetsavepath)

DeerMichel:
_nb to the relative heading-links_: let's say we add a new `Tray` heading at top of the document (e.g. for v8.0). I guess then all previous links will break bc they'll change: `tray -> tray-1`, `tray-1 -> tray-2`.

PR #19215, opened 2019-07-11
feat: allow setting capture mode and max file size in netLog API

nornagon:
#### Description of Change
Adds two new options to `netLog.startLogging`: `maxFileSize` and `captureMode`.

NB. the names of the capture modes changed recently: https://chromium-review.googlesource.com/c/chromium/src/+/1685684. This PR uses the new names, and depends on that PR to fix the decoding of the "EVERYTHING" capture mode, which until that PR was decoded on the receiving mojo end as if it were the "INCLUDE_PRIVACY_INFO" capture mode.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added `captureMode` and `maxFileSize` options to the netLog API.

### DIFF
--- a/shell/browser/api/atom_api_net_log.cc
+++ b/shell/browser/api/atom_api_net_log.cc
@@ -92,7 +92,7 @@ v8::Local<v8::Promise> NetLog::StartLogging(base::FilePath log_path,
     return v8::Local<v8::Promise>();
   }
 
-  net::NetLogCaptureMode capture_mode = net::NetLogCaptureMode::Default();
+  net::NetLogCaptureMode capture_mode = net::NetLogCaptureMode::kDefault;
   uint64_t max_file_size = network::mojom::NetLogExporter::kUnlimitedFileSize;
 
   mate::Dictionary dict;
--- a/shell/browser/api/atom_api_net_log.cc
+++ b/shell/browser/api/atom_api_net_log.cc
@@ -11,6 +11,7 @@
 #include "components/net_log/chrome_net_log.h"
 #include "content/public/browser/storage_partition.h"
 #include "electron/electron_version.h"
+#include "native_mate/converter.h"
 #include "native_mate/dictionary.h"
 #include "native_mate/handle.h"
 #include "net/url_request/url_request_context_getter.h"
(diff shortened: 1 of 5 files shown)


shell/browser/api/atom_api_net_log.cc
```diff
@@ -60,13 +85,36 @@ NetLog::NetLog(v8::Isolate* isolate, AtomBrowserContext* browser_context)
...
   }
 
+  net::NetLogCaptureMode capture_mode = net::NetLogCaptureMode::Default();
+  uint64_t max_file_size = network::mojom::NetLogExporter::kUnlimitedFileSize;
+
+  mate::Dictionary dict;
+  if (args->GetNext(&dict)) {
+    v8::Local<v8::Value> capture_mode_v8;
+    if (dict.Get("captureMode", &capture_mode_v8)) {
+      if (!mate::ConvertFromV8(args->isolate(), capture_mode_v8,
```
zcbenz: You can convert to target type directly:

```c++
net::NetLogCaptureMode capture_mode;
if (dict.Get("captureMode", &capture_mode)) {
  // ...
}
```

shell/browser/api/atom_api_net_log.cc
```diff
@@ -60,13 +85,36 @@ NetLog::NetLog(v8::Isolate* isolate, AtomBrowserContext* browser_context)
...
+                               &capture_mode)) {
+        args->ThrowError("Invalid value for captureMode");
+        return v8::Local<v8::Promise>();
+      }
+    }
+    v8::Local<v8::Value> max_file_size_v8;
+    if (dict.Get("maxFileSize", &max_file_size_v8)) {
+      if (!mate::ConvertFromV8(args->isolate(), max_file_size_v8,
```
zcbenz: Same with above.

nornagon:
The test is failing in CI because of the reasoning stated in the PR description; this depends on a Chromium roll to be fixed.

shell/browser/api/atom_api_net_log.cc
nornagon: The reason not to do this is so that we can produce an error if the option was specified but of the incorrect type, as opposed to silently proceeding with the default value.

PR #19216, opened 2019-07-11
fix: exit non-silently when running as root on Linux

### LINKED ISSUE #18547: Electron exits silently when running as root on Linux

DeerMichel:
#### Description of Change
Closes #18547. Follow-up to #18903.

In short, this PR adds an error message on Linux when running as root user without `--no-sandbox` flag. This flag is necessary as described in [a Chromium issue](https://crbug.com/638180). Previously, Electron just failed silently and only warned root users if logging has explicitly been enabled. 

cc @erickzhao @codebytere  

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: none

### DIFF
--- a/shell/app/atom_main_delegate.cc
+++ b/shell/app/atom_main_delegate.cc
@@ -201,6 +201,14 @@ bool AtomMainDelegate::BasicStartupComplete(int* exit_code) {
     base::win::PinUser32();
 #endif
 
+#if defined(OS_LINUX)
+  // Check for --no-sandbox parameter when running as root.
+  if (getuid() == 0 && IsSandboxEnabled(command_line))
+    LOG(FATAL) << "Running as root without --"
+               << service_manager::switches::kNoSandbox
+               << " is not supported. See https://crbug.com/638180.";
+#endif
+
   content_client_ = std::make_unique<AtomContentClient>();
   SetContentClient(content_client_.get());
 


DeerMichel:
I don't know if that's worth the hassle since we had to run these tests as root then. Tbh I can't imagine how this should become a regression bug 🙈

jkleinsc:
@erickzhao @DeerMichel thanks for working on this!  In regards to testing, it would be tricky as we don't usually run CI as root.  The only exception to that is on arm/arm64.  I'm in favor of merging this PR as is without a spec change.

PR #19217, opened 2019-07-12
feat: support `systemPreferences.isDarkMode()` on Windows

### LINKED ISSUE #15316: Windows support for `systemPreferences.isDarkMode()` API
Electron has an API which works on macOS for detecting when the system is in Dark Mode.
[Windows 10 also has a dark mode setting](https://www.digitaltrends.com/computing/how-to-enable-dark-mode-in-windows-10/). It would be useful if the same Electron API also worked on Windows.

1. The `systemPreferences.isDarkMode()` method should work on Windows.
2. Ideally a way of subscribing to change events, equivalent to `systemPreferences.subscribeNotification('AppleInterfaceThemeChangedNotification', callback)`. Or perhaps the `color-changed` event (or another event) would be fired by `systemPreferences`.

**Additional context**
On Windows the state of the dark mode setting [is stored in the](https://stackoverflow.com/a/51336913/365069) `HKCU(HKLM)\Software\Microsoft\Windows\CurrentVersion\Themes\Personalize\AppsUseLightTheme` registry key.

### LINKED ISSUE #18696: Add method for getting Windows 10 Shell Color
### Preflight Checklist

### Problem Description
I need a way of checking if the user is in https://i.imgur.com/yvZYfsG.png
Windows Dark mode or light mode

### Proposed Solution
systemPreferences.getShellColorType()

### Alternatives Considered
n/a

### Additional Information
Because my app icon is white, I want it to invert if using the Windows 10 light mode.
https://i.imgur.com/dzsLQTs.png
You can barely see it in light mode.

DeerMichel:
#### Description of Change
Closes #18696. Closes #15316. Adds support for the `systemPreferences.isDarkMode()` API on Windows using Chromium's `NativeTheme` API. 

#### Outdated (not-actually-used) approach
Whether the system uses dark or light theme is derived from following registry key: 
```
HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Themes\Personalize\AppsUseLightTheme 
  | value 0  == dark
  | value 1+ == light
```

cc @MarshallOfSound @erickzhao @codebytere 

#### Checklist
- [x] `npm test` passes

#### Release Notes

Notes: Added support for `systemPreferences.isDarkMode()` API on Windows.

### DIFF
--- a/docs/api/system-preferences.md
+++ b/docs/api/system-preferences.md
@@ -43,7 +43,7 @@ Returns:
 
 ## Methods
 
-### `systemPreferences.isDarkMode()` _macOS_
+### `systemPreferences.isDarkMode()` _macOS_ _Windows_
 
 Returns `Boolean` - Whether the system is in Dark Mode.
 
--- a/shell/browser/api/atom_api_system_preferences.cc
+++ b/shell/browser/api/atom_api_system_preferences.cc
@@ -10,6 +10,7 @@
 #include "shell/common/node_includes.h"
 #include "ui/gfx/animation/animation.h"
 #include "ui/gfx/color_utils.h"
+#include "ui/native_theme/native_theme.h"
 
 namespace electron {
 
@@ -28,9 +29,9 @@ SystemPreferences::~SystemPreferences() {
 #endif
 }
 
-#if defined(OS_LINUX)
+#if !defined(OS_MACOSX)
 bool SystemPreferences::IsDarkMode() {
-  return false;
+  return ui::NativeTheme::GetInstanceForNativeUi()->SystemDarkModeEnabled();
 }
 #endif
 
--- a/shell/browser/api/atom_api_system_preferences.cc
+++ b/shell/browser/api/atom_api_system_preferences.cc
@@ -28,7 +28,7 @@ SystemPreferences::~SystemPreferences() {
 #endif
 }
 
-#if !defined(OS_MACOSX)
+#if defined(OS_LINUX)
 bool SystemPreferences::IsDarkMode() {
   return false;
 }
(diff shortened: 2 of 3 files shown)


deepak1556:
I want to reiterate on this again https://github.com/electron/electron/pull/18666#issuecomment-502882748, why aren't we reusing the api from `ui::NativeTheme` ?

/cc @codebytere @MarshallOfSound

DeerMichel:
Huh... probs bc that would've been too easy 😄. I'll take a closer look at it tmw - looks promising... and even offers an event emitter for theme changes.

Maybe we can get Linux support as well and upstream it to Chromium?

codebytere:
@DeerMichel @deepak1556 we can’t use `ui::NativeTheme` without modifying Chromium; they only support Mojave and higher for isDarkMode(). They only check `effectiveAppearance` whereas we support legacy user prefs back to `10.12`

MarshallOfSound:
Yeah macOS is the issue there.  We could use that api just for Windows support though?

deepak1556:
@codebytere I don't see the reason to patch even if thats the case we could return `ui::NativeTheme` result for `if (@available(macOS 10.14, *)) {` else return result from our implementation..

codebytere:
oops my comment somehow deleted but in response to the above, yeah it might just be fine to gate it on mac and use for windows in totality

DeerMichel:
**Before** merging, should I also add a `dark-mode-changed` (or sth like that) event or better an additional PR? (And i might migrate the other methods to `NativeTheme` as well)

MarshallOfSound:
@DeerMichel If we do do a first-party `dark-mode-changed` event we should probs do it for macOS as well and that requires some consideration as we support < 10.14.  Let's land `isDarkMode` for windows and think about the event separately

DeerMichel:
You're right, I totally forgot about macOS. Then this PR is ready to get merged 👍

caesar:
Will this make it into Electron 6.0.0?

PR #19226, opened 2019-07-12
feat: enable dark mode support by default on macOS

MarshallOfSound:
This adds the NSRequiresAquaSystemAppearance key to our default Info.plist file which will tell macOS to auto-switch our effectiveAppearance in sync with the OS.  The dark mode documentation has been updated to reflect how to opt *out* of this but it is also noted that certain dark mode APIs will not work on Catalina if you opt out.

#### Release Notes

Notes: Native interfaces on macOS (menus, dialogs, etc.) now automatically match the dark mode setting on the users machine.

### DIFF
--- a/docs/api/system-preferences.md
+++ b/docs/api/system-preferences.md
@@ -47,6 +47,8 @@ Returns:
 
 Returns `Boolean` - Whether the system is in Dark Mode.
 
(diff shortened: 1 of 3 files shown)


caesar:
Since this PR is already merged I have fixed the above issues in #19358

PR #19235, opened 2019-07-12
test: move MenuItem spec to main process

nornagon:
#### Description of Change
A couple mysteries here:

1. The `returns undefined for items without default accelerator` test appears to have been broken for a while, using `for..in` instead of `for..of`. Now that I unbroke it to test the right thing, apparently the test fails. I've skipped it for now but that doesn't seem like the right resolution. cc @codebytere for guidance.
2. The test for `item.commandId` not being overridable seems to have changed behaviour now that it's in the main process? it's saying `Cannot assign to read only property`. Is this an artifact of how `remote` works?

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/lib/browser/api/menu-item-roles.js
+++ b/lib/browser/api/menu-item-roles.js
@@ -251,6 +251,8 @@ const roles = {
   }
 }
 
+exports.roleList = roles
+
 const canExecuteRole = (role) => {
   if (!roles.hasOwnProperty(role)) return false
   if (!isMac) return true
--- a/lib/browser/api/menu-item-roles.js
+++ b/lib/browser/api/menu-item-roles.js
@@ -82,7 +82,7 @@ const roles = {
         default: return 'Quit'
       }
     },
-    accelerator: isWindows ? null : 'CommandOrControl+Q',
+    accelerator: isWindows ? undefined : 'CommandOrControl+Q',
     appMethod: 'quit'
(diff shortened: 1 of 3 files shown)


codebytere:
`Mystery #1` solved :)

zcbenz:
`#1` still seems to be failing

```
[00:37:24] not ok 243 MenuItems MenuItem role returns undefined for items without default accelerator
[00:37:24]   AssertionError: default accelerator: expected null to be undefined
[00:37:24]       at Context.<anonymous> (electron\spec-main\api-menu-item-spec.ts:207:56)
[00:37:24]       at processImmediate (internal/timers.js:445:21)
```

> 2\. The test for `item.commandId` not being overridable seems to have changed behaviour now that it's in the main process? it's saying `Cannot assign to read only property`. Is this an artifact of how `remote` works?

The remote module only works on "own properties", I think `commandId` was very likely ignored.

codebytere:
@nornagon that's failing owing to:
https://github.com/electron/electron/blob/8782d06ed610e346f70862561aa033905c93a86c/lib/browser/api/menu-item-roles.js#L85; imo we should change that to
```js
accelerator: isWindows ? undefined : 'CommandOrControl+Q',
```

for consistency anyways so once that's done the specs should pass

PR #15238, opened 2018-10-18
chore: make spec runner build headers and node.lib

MarshallOfSound:
Annoying that we have to run these manually (I always forget to copy the lib file on windows)

Notes: no-notes

PR #19244, opened 2019-07-13
chore: ensure up-to-date node_headers in spec runner

ckerr:
#### Description of Change

In the spec runner script, re-run `ninja third_party/electron_node:headers` if either the node headers directory doesn't exist or if something's changed.

I'm adding this because I'm switching my [personal scripts](https://github.com/ckerr/dotfiles/blob/master/assets/public/dotfiles/.zshrc.d/electron-gn.zsh) over to use spec-runner and this step seems to be missing. This might be because the runner is just used by CI, which runs that separately. Or if it's used by other maintainers, maybe this step is handled in their personal scripts -- but it needs to be re-run when things change, so handling it in the spec runner is better.

CC @MarshallOfSound, @jkleinsc 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/spec-runner.js
+++ b/script/spec-runner.js
@@ -27,6 +27,7 @@ const { YARN_VERSION } = require('./yarn')
 const BASE = path.resolve(__dirname, '../..')
 const NPM_CMD = process.platform === 'win32' ? 'npm.cmd' : 'npm'
 const NPX_CMD = process.platform === 'win32' ? 'npx.cmd' : 'npx'
+const NODE_HEADERS_DIR = path.resolve(BASE, `out/${utils.OUT_DIR}/gen/node_headers`)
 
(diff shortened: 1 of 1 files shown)


codebytere:
@ckerr this won't work for windows; @MarshallOfSound attempted this PR previously here: https://github.com/electron/electron/pull/15238 and kept hitting issues for the windows lib file copy step

zcbenz:
> I'm adding this because I'm switching my [personal scripts](https://github.com/ckerr/dotfiles/blob/master/assets/public/dotfiles/.zshrc.d/electron-gn.zsh) over to use spec-runner and this step seems to be missing.

I wonder if every Electron developer has their own build scripts 😄 .

PR #19245, opened 2019-07-13
refactor: don't send ipcRenderer.sendSync() reply as array

miniak:
#### Description of Change
`event.sendReply()` was refactored to take `base::Value` instead of `base::ListValue` in #18449. We don't need to pass the return value of `ipcRenderer.sendSync()` as an array via `event.sendReply()` anymore.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes


miniak:
@zcbenz it seems like I cannot be done as `undefined` cannot be transferred as plain `base::Value`, it needs to be done via an empty array

PR #19260, opened 2019-07-15
fix: don't execute preload scripts for internal <iframe> in <webview> when nodeIntegrationInSubFrames is enabled

### LINKED ISSUE #18429: nodeIntegrationInSubFrames webview attribute causes preload script to run multiple times
### Preflight Checklist

### Issue Details

* **Electron Version:**
  5.0.1
* **Operating System:**
  windows
  n/a

### Expected Behavior
A preload script should run only 1 time for each webContents / iframe

### Actual Behavior
The preload script appears to run multiple times

### To Reproduce
https://github.com/Tim91084/electron-node-subframe-issue/tree/node-in-subframe-runs-too-often

```sh
$ git clone https://github.com/Tim91084/electron-node-subframe-issue.git -b node-in-subframe-runs-too-often
$ npm install
$ npm start || electron .
```

### Screenshots
image

miniak:
#### Description of Change
Fixes #18429.

The problem is that before the `webContents` is attached to the internal `<iframe>` inside of the `<webview>`, it's treated as a regular frame and the preload scripts are executed there.

Fix by checking `v8Util.getHiddenValue(window.frameElement, 'internal')` on context initialization, which is set by the `<webview>` implementation:
https://github.com/electron/electron/blob/50f49770483d1579ef47e63eca58498072681f84/lib/renderer/web-view/web-view-impl.ts#L54-L61

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Preload scripts for internal `<iframe>` in `<webview>` are not longer executed when `nodeIntegrationInSubFrames` is enabled.

### DIFF
--- a/shell/renderer/atom_renderer_client.cc
+++ b/shell/renderer/atom_renderer_client.cc
@@ -83,7 +83,8 @@ void AtomRendererClient::DidCreateScriptContext(
       base::CommandLine::ForCurrentProcess()->HasSwitch(
           switches::kNodeIntegrationInSubFrames);
   bool should_load_node =
-      is_main_frame || is_devtools || allow_node_in_subframes;
+      (is_main_frame || is_devtools || allow_node_in_subframes) &&
+      !IsWebViewFrame(renderer_context, render_frame);
   if (!should_load_node) {
     return;
   }
--- a/shell/renderer/atom_sandboxed_renderer_client.cc
+++ b/shell/renderer/atom_sandboxed_renderer_client.cc
@@ -208,7 +208,8 @@ void AtomSandboxedRendererClient::DidCreateScriptContext(
       base::CommandLine::ForCurrentProcess()->HasSwitch(
           switches::kNodeIntegrationInSubFrames);
   bool should_load_preload =
-      is_main_frame || is_devtools || allow_node_in_sub_frames;
+      (is_main_frame || is_devtools || allow_node_in_sub_frames) &&
+      !IsWebViewFrame(context, render_frame);
   if (!should_load_preload)
     return;
 
--- a/shell/renderer/atom_render_frame_observer.cc
+++ b/shell/renderer/atom_render_frame_observer.cc
@@ -81,7 +81,8 @@ void AtomRenderFrameObserver::DidCreateScriptContext(
 
   if (should_create_isolated_context) {
     CreateIsolatedWorldContext();
-    renderer_client_->SetupMainWorldOverrides(context, render_frame_);
+    if (!renderer_client_->IsWebViewFrame(context, render_frame_))
+      renderer_client_->SetupMainWorldOverrides(context, render_frame_);
   }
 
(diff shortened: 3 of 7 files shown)


shell/renderer/atom_renderer_client.cc
```diff
@@ -84,7 +84,7 @@ void AtomRendererClient::DidCreateScriptContext(
           switches::kNodeIntegrationInSubFrames);
   bool should_load_node =
       is_main_frame || is_devtools || allow_node_in_subframes;
-  if (!should_load_node) {
+  if (!should_load_node || IsWebViewFrame(context)) {
```
MarshallOfSound: This condition should be added to `should_load_node`.  i.e. `IsWebViewFrame` implies `!should_load_node`

shell/renderer/atom_renderer_client.cc
```diff
@@ -203,7 +203,8 @@ void AtomRendererClient::SetupMainWorldOverrides(
       node::FIXED_ONE_BYTE_STRING(isolate, "isolatedWorld")};
 
   auto* env = GetEnvironment(render_frame);
-  DCHECK(env);
```
MarshallOfSound: This `DCHECK` was deliberately added, why was it removed?

This will now silently absorb failures, there is no case here where `GetEnvironment` should return `nullptr`.  If there is we shouldn't be calling `SetupMainWorldOverrides` in that case.

shell/renderer/atom_sandboxed_renderer_client.cc
```diff
@@ -209,7 +209,7 @@ void AtomSandboxedRendererClient::DidCreateScriptContext(
           switches::kNodeIntegrationInSubFrames);
   bool should_load_preload =
       is_main_frame || is_devtools || allow_node_in_sub_frames;
-  if (!should_load_preload)
+  if (!should_load_preload || IsWebViewFrame(context))
```
MarshallOfSound: Same comment as above RE extending existing boolean

zcbenz:
The test is failing on Windows CI:

```
not ok 515 renderer nodeIntegrationInSubFrames with context isolation, webview on should correctly reply to the sub-frames with using event.reply
23533  Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\projects\src\electron\spec\api-subframe-spec.js)
23534      at Test.Runnable._timeoutError (C:\projects\src\electron\spec\node_modules\mocha\lib\runnable.js:440:10)
23535      at C:\projects\src\electron\spec\node_modules\mocha\lib\runnable.js:251:24
23536
```

miniak:
@zcbenz I've fixed the tests

KochiyaOcean:
https://github.com/electron/electron/blob/master/docs/api/webview-tag.md#nodeintegrationinsubframes

PR #18924, opened 2019-06-21
chore: update to Node.js v12.4.0

codebytere:
#### Description of Change

Update our Node dependency to v12.4.0.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Updated Node dependency to v12.4.0.

PR #19281, opened 2019-07-16
feat: update to node 12.4 (#19270)

MarshallOfSound:
Backport of #18924

This points our node repo at upstream (nodejs/node) and uses the base node tag as the target ref.  We then use our existing patch system and patch files to apply our changes on top of node.  This unifies how we patch upstream repos and makes our node patches easier to reason, view, understand and most importantly reduce.

This also upgrades Electron 6 to Node 12.4

Notes: Upgraded to Node 12.4


dsanders11:
Does this effectively deprecate the electron/node repo going forward?

MarshallOfSound:
@dsanders11 Once the `4-2-x` branch has been updated with a similar PR for the version of node that `4.2` uses yes, that repo will be no longer used

PR #19292, opened 2019-07-16
build: move our node fork to our patch system

MarshallOfSound:
This points our node repo at upstream (nodejs/node) and uses the base node tag as the target ref.  We then use our existing patch system and patch files to apply our changes on top of node.  This unifies how we patch upstream repos and makes our node patches easier to reason, view, understand and most importantly reduce.

Notes: no-notes


DEPS
```diff
@@ -12,7 +12,7 @@ vars = {
   'chromium_version':
     '73.0.3683.121',
   'node_version':
-    '0a300f60bce0c8f0cb3d846fcb0e1f55f26013ee',
+    '7ccfb6a1de3b8b028e556d6ed2dcee3025e311a6',
```
MarshallOfSound: We should figure out why this commit (what our node fork was based off) and the node `v12.0.0` tag are different 😕

nornagon: agreed, but that shouldn't block this change IMO

PR #19294, opened 2019-07-16
feat: add rotate gesture for macOS

erickzhao:
#### Description of Change
Adds `BrowserWindow` event support for the [rotation trackpad gesture on macOS](https://support.apple.com/en-us/HT204895). This event is emitted constantly between the start and the end of the rotation gesture, and has an attached value corresponding to the degree in angles of the current timespan.

See `NSResponder` documentation: https://developer.apple.com/documentation/appkit/nsresponder/1525572-rotatewithevent?language=objc

cc @codebytere 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added support for rotation multi-touch gestures on `BrowserWindow` for macOS.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -615,7 +615,7 @@ Returns:
 
 Emitted on 3-finger swipe. Possible directions are `up`, `right`, `down`, `left`.
 
-#### Event: 'rotate' _macOS_
+#### Event: 'rotate-gesture' _macOS_
 
 Returns:
 
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -615,6 +615,19 @@ Returns:
 
 Emitted on 3-finger swipe. Possible directions are `up`, `right`, `down`, `left`.
 
+#### Event: 'rotate' _macOS_
+
+Returns:
+
+* `event` Event
+* `rotation` Float
+
+Emitted on trackpad rotation gesture. Continually emitted until rotation gesture is
+ended. The `rotation` value on each emission is the angle in degrees rotated since
+the last emission. The last emitted event upon a rotation gesture will always be of
+value `0`. Counter-clockwise rotation values are positive, while clockwise ones are
+negative.
+
 #### Event: 'sheet-begin' _macOS_
 
 Emitted when the window opens a sheet.
--- a/shell/browser/api/atom_api_top_level_window.cc
+++ b/shell/browser/api/atom_api_top_level_window.cc
@@ -246,8 +246,8 @@ void TopLevelWindow::OnWindowSwipe(const std::string& direction) {
   Emit("swipe", direction);
 }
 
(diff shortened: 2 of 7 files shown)


shell/browser/api/atom_api_top_level_window.cc
```diff
@@ -246,6 +246,10 @@ void TopLevelWindow::OnWindowSwipe(const std::string& direction) {
   Emit("swipe", direction);
 }
 
+void TopLevelWindow::OnWindowRotate(const float rotation) {
+  Emit("rotate", rotation);
```
MarshallOfSound: `rotate` doesn't imply the right thing here, it could be read as the window rotating.  This should probably be called `rotate-gesture` or `swipe-rotate` or something that makes it more clear it's a finger gesture

erickzhao: good point!

erickzhao:
> needs documentation :)

Wrote docs and forgot to push them 🤦‍♂ 

Let me know if the description of the `rotation` variable in the callback is too confusing/verbose!

shell/browser/api/atom_api_top_level_window.cc
erickzhao: switched `rotate` -> `rotate-gesture` both in the exposed API and the underlying native code for clarity's sake!

docs/api/browser-window.md
```diff
@@ -615,6 +615,19 @@ Returns:
...
+#### Event: 'rotate-gesture' _macOS_
+
+Returns:
+
+* `event` Event
+* `rotation` Float
+
+Emitted on trackpad rotation gesture. Continually emitted until rotation gesture is
+ended. The `rotation` value on each emission is the angle in degrees rotated since
+the last emission. The last emitted event upon a rotation gesture will always be of
```
ckerr: Point of clarification -- is this the delta since the last emission, or the delta since the rotation began? These docs seem to be saying the former, but online docs of NSEvent seem to be saying the latter.

I'm not asserting this is correct or incorrect; I haven't tested for myself. But if this is the delta since rotation began, we should say that instead.

erickzhao: @ckerr I found the official docs to be confusing in this regard. From my testing, it seemed like the emitted value on each tick was the Δdegrees since the last tick. I tried doing a 90-degree rotation, and the value from each emitted event added up to more or less 90 degrees.

shell/browser/native_window.h
```diff
@@ -256,6 +256,7 @@ class NativeWindow : public base::SupportsUserData,
   void NotifyWindowScrollTouchBegin();
   void NotifyWindowScrollTouchEnd();
   void NotifyWindowSwipe(const std::string& direction);
+  void NotifyWindowRotateGesture(const float rotation);
```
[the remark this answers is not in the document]
erickzhao: good catch! thanks for the explanation too 🚀 👍

PR #19296, opened 2019-07-17
build: remove unused node patches

MarshallOfSound:
One of these patches we don't use anymore (exposing `NativeModule`)
The other was replaced with another patch later down the patch chain

Notes: no-notes


patches/node/build_modify_js2c_py_to_allow_injection_of_original-fs_and_custom_embedder_js.patch
```diff
@@ -1,7 +1,8 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Cheng Zhao <zcbenz@gmail.com>
 Date: Thu, 11 Apr 2019 17:16:13 +0900
-Subject: build: modify js2c.py to allow injection of original-fs and custom embedder JS
+Subject: build: modify js2c.py to allow injection of original-fs and custom
+ embedder JS
```
nornagon: this is kinda weird that it line wraps now?

MarshallOfSound: Came straight out of `git-export-patches`, idk what to tell you 😆

Issue #19131, opened 2019-07-06
BrowserWindow.on('swipe') broken since Electron 3


bpasero:
### Issue Details

* **Electron Version:**
  * 3.0.x, 4.0.x, 5.0.x, 6.0.x
* **Operating System:**
  * macOS 10.14.5
* **Last Known Working Electron version:**
  * 2.0.x

### Expected Behavior
`BrowserWindow.on('swipe')` fires when doing a 3-finger swipe gesture.

### Actual Behavior
`BrowserWindow.on('swipe')` does not fire.

bpasero:
@codebytere @deepak1556 this seems to have caused a regression in VSCode where our window size is assumed to be larger than it is when macOS native tabs are enabled. See https://github.com/microsoft/vscode/issues/79176

bpasero:
![grid-tw](https://user-images.githubusercontent.com/900690/63094369-76136f80-bf68-11e9-9763-6b078dcfc0f8.gif)

KASOGIT:
Swipe event still isn't triggered with electron 7, news ?

sdnyco:
In addition to 'rotate-gesture'. Still not working in 8.1.1

codebytere:
@sdnyco `rotate-gesture` is working just fine for me, see [this fiddle](https://gist.github.com/4535eedab0fe411b1f20ac184f5099d0).

codebytere:
@KASOGIT it does work: it's your system settings that make it appear not to.

`-swipeWithEvent:` is used to implement 10.6-style trackpad swiping, where the content on the screen doesn't move with the swipe. Most Mac trackpads are not configured to allow this kind of swiping anymore; the "Swipe between pages" preference in Trackpad pref pane has to be set to "swipe with two or three fingers" for it to be available.

<img width="780" alt="Screen Shot 2020-04-14 at 5 40 47 PM" src="https://user-images.githubusercontent.com/2036040/79286977-2b3f5400-7e77-11ea-9a70-4f1df7132a71.png">

sourabhverificient:
i wanted to prevent **three figure swipe gesture** i tried

`win.on("swipe", (e, d) => {
    log.info("Swipe")
    // e.preventDefault()

  })
  win.on("rotate-gesture", (e, r) => {
    log.info("rotate-gesture")
    // e.preventDefault()
  })`
  
but i am not even getting logs. it seems **both event  not working**.
please let me know if anyone have any solution.

seivan:
[...]

PR #19319, opened 2019-07-17
fix: emit swipe event on macOS

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/19131.

Previously, we only overrode the `[window_ contentView]` in frameless windows, which came at cost of preventing users from accessing `views::View` APIs. This PR applies that override to framed windows as well, such that we are now unconditionally removing `BridgedContentView` as content view on BrowserWindows.

cc @nornagon @zcbenz @miniak 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed 'swipe' event emission on macoS>

### DIFF
--- a/shell/browser/api/atom_api_browser_window.cc
+++ b/shell/browser/api/atom_api_browser_window.cc
@@ -84,8 +84,7 @@ BrowserWindow::BrowserWindow(v8::Isolate* isolate,
   InitWith(isolate, wrapper);
 
 #if defined(OS_MACOSX)
-  if (!window()->has_frame())
-    OverrideNSWindowContentView(web_contents->managed_web_contents());
+  OverrideNSWindowContentView(web_contents->managed_web_contents());
 #endif
 
   // Init window after everything has been setup.


deepak1556:
> I'm not 100% confident that there is no side effect, but this should be fine.

@zcbenz it regressed behavior for native tabs https://github.com/microsoft/vscode/issues/79176

zcbenz:
@deepak1556 Hmm I think it should be possible to fix the issue without reverting this, can you make a simple app that I can use to test?

deepak1556:
@zcbenz https://gist.github.com/deepak1556/5edd25929a0a3077a89328eaf60ec0f9 , visiting the inactive tab should show you the clipped view.

deepak1556:
ezgif com-video-to-gif


Demonstrating the clipped view on **4.2.9**

zcbenz:
@deepak1556 Thanks for the example, I can reproduce with it.

zcbenz:
@deepak1556 Do you mind creating a new issue for the regression? I think it would take more time than I expected.

deepak1556:
@zcbenz @codebytere have created https://github.com/electron/electron/issues/19875, in the meantime can we revert this PR change since it makes things worse and the original issue it fixes has been broken for a long time.

zcbenz:
@deepak1556 I'll have a possible fix today, if it does not work then we should probably revert this.

deepak1556:
Sounds good, thanks!

Issue #11371, opened 2017-12-07
Filters don't work in webRequest


pronebird:
* Electron version: 1.7.9
* Operating system: macOS

### Expected behavior

The callback should be used only for URLs matching the filter.

```
const filter = { urls: ['*://www.googleapis.com'] };
session.defaultSession.webRequest.onBeforeRequest(filter, (details, callback) => {
  console.log('onBeforeRequest: ', details);
  callback({ cancel: true });
});
```

### Actual behavior

The callback is called for random URLs.

```
onBeforeRequest:  { id: 1,
  method: 'GET',
  referrer: '',
  resourceType: 'mainFrame',
  timestamp: 1512654154551.9841,
  url: 'file:///Users/pronebird/apple/mail/build/index.html' }
onBeforeRequest:  { id: 4,
  method: 'GET',
  referrer: '',
  resourceType: 'mainFrame',
  timestamp: 1512654154564.2021,
  url: 'chrome-devtools://devtools/bundled/inspector.html?remoteBase=https://chrome-devtools-frontend.appspot.com/serve_file/@691bdb490962d4e6ae7f25c6ab1fdd0faaf19cd0/&can_dock=&toolbarColor=rgba(223,223,223,1)&textColor=rgba(0,0,0,1)&experiments=true' }
```

Pingze-github:
Same problem in v1.8.3

brospars:
Same problem on Linux (Ubuntu 18.04) any news on this ?

tcd156:
Still true in 3.0.x

naturejune:
same problem in 4.0.1

ShadyThGod:
Seems to happen in 4.1.1 as well.
Why has there been no progress on this?

Ketsuekiame:
Still an issue in Electron 5.0.7

OS: Linux Mint 15.1

Issue #19223, opened 2019-07-12
😢The filter of webRequest has no effects.


TongDaDa:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * LATEST
  
* **Operating System:**
  * macOS 10.13.6 

### Expected Behavior

```javascript

const filter = {
     urls: [
          "file://*.sobot.com/*"
     ]
}

browserWindow.webContents.session.webRequest.onBeforeRequest(filter,  (details, callback) => {
            const { id, url, method, resourceType, timestamp, uploadData } = details;
            // url === "file://users/liutong/Desktop/**/projectName/app/***"
           callback({
                cancel: false,
                redirectURL: 'http' + url.slice(4)
            })
        })
```

The callback of `onBeforeRequest` will be called by any request from the file scheme.

### Actual Behavior

For the above codes, I just want to redirect a request that contains the '*.sobot.com' domain,
But I received a URL from the local path of `/users/Oliver/Desktop` in the callback of `onBeforeRequest`.

When I add "file://\*.sobot.com/\*" a rule in the array, for the rule, I actually expected it only to match request that domain is '.sobot.com', but It appears to listen to all requests. e.g 

```
[ "file://users/Oliver/Desktop/**/projectName/app/caching_resources/***" ]
```

### Additional Information

As for my final purpose, you can refer to this [issue](https://github.com/electron/electron/issues/19200). so my fallback is to exclude any request that contains a path of 'caching_resources' as to redirect to the `HTTP` scheme.

TongDaDa:
My filter list is :

```json
["file://www.sobot.com/chat/*", "file://crumb.sobot.com/1.gif*", "file://cdn.console.pipacoding.com/*"],
```

pushkin-:
This is a dupe of [this issue](https://github.com/electron/electron/issues/11371)

TongDaDa:
Yes, I am dupe

codebytere:
Closing as duplicate of https://github.com/electron/electron/issues/11371 in agreement with @pushkin-

TongDaDa:
[...]

Issue #19200, opened 2019-07-10
file:// protocol block resources from "//***.com"


TongDaDa:
😊Oh, I am very embarrassed to ask this question, but I am very cautious after reading the documents I have possible, I came here to ask questions, I hope to get your help!

I used the express to create a local server, and then opened a port using the http protocol to hang my html resources, but the real user environment is full of problems, such as opening a local firewall, port occupancy (although we Can be solved by the program). Because of these problems, I switched to the file protocol, so that we don't have to open the port, and these problems do not exist.

Today, I found another problem. There are some third-party sdk in my application. It loads some scripts and it sounds normal, but the address it loads does not specify an explicit network protocol, such as http or Instead of https, use "//****.com/path/to/resources". This directly led to an error that the file could not be found.

Here, I am a headache, I don't know what to do, I read all the related documents of electron, including using the `register***Protocol` interface, but they do not seem to have a scheme to handle multiple (file, http ) the ability to return data types.

my goal:
I just want to be able to load my resources without opening the port, of course considering the above "//".

sofianguy:
GitHub issues are for feature requests and bug reports, questions about using Electron or code assistance requests should be directed to the [community](https://github.com/electron/electron#community) or to the [Slack Channel](https://atom-slack.herokuapp.com/).

TongDaDa:
There no any reply.

PR #19337, opened 2019-07-19
fix: throw on invalid webRequest filters

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/11371.

Previously, we didn't consider the return value of the `webRequest` `URLPattern` mate converter, which meant that when the pattern wasn't correctly parsed owing to invalid filter specification users would not be made aware of that fact and would just think that the filtering itself had failed. This corrects that error by moving the business logic of url pattern parsing out of the converter and into the function itself so that granular and specific errors can be thrown.

There's also no real reason that i'm aware of not to allow wider breadth of filters by letting users use a wildcard for effective TLD, so I also overrode that (default for the 1-arg `Parse` is not to allow that).

Finally, I added some examples of url filter types for users to reference.

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed error throwing on invalid `webRequest` url pattern filtering in `onBeforeRequest`.

### DIFF
--- a/docs/api/web-request.md
+++ b/docs/api/web-request.md
@@ -68,6 +68,21 @@ The `uploadData` is an array of `UploadData` objects.
 
 The `callback` has to be called with an `response` object.
 
+Some examples of valid `urls`:
+
+```js
+'http://foo:1234/'
+'http://foo.com/'
+'http://foo:1234/bar'
+'*://*/*'
+'*://example.com/*'
+'*://example.com/foo/*'
+'http://*.foo:1234/'
+'file://foo:1234/bar'
+'http://foo:*/'
+'*://www.foo.com/'
+```
+
 #### `webRequest.onBeforeSendHeaders([filter, ]listener)`
 
 * `filter` Object (optional)
--- a/native_mate/native_mate/dictionary.h
+++ b/native_mate/native_mate/dictionary.h
@@ -40,6 +40,16 @@ class Dictionary {
 
   static Dictionary CreateEmpty(v8::Isolate* isolate);
 
+  bool HasKey(const base::StringPiece& key) const {
+    v8::Local<v8::Context> context = isolate_->GetCurrentContext();
+    v8::Local<v8::String> v8_key = StringToV8(isolate_, key);
+    if (!internal::IsTrue(GetHandle()->Has(context, v8_key)))
+      return false;
+
+    v8::Local<v8::Value> val;
+    return GetHandle()->Get(context, v8_key).ToLocal(&val);
+  }
+
   template <typename T>
   bool Get(const base::StringPiece& key, T* out) const {
     // Check for existence before getting, otherwise this method will always
--- a/native_mate/native_mate/dictionary.h
+++ b/native_mate/native_mate/dictionary.h
@@ -40,16 +40,6 @@ class Dictionary {
 
   static Dictionary CreateEmpty(v8::Isolate* isolate);
 
-  bool HasKey(const base::StringPiece& key) const {
-    v8::Local<v8::Context> context = isolate_->GetCurrentContext();
-    v8::Local<v8::String> v8_key = StringToV8(isolate_, key);
-    if (!internal::IsTrue(GetHandle()->Has(context, v8_key)))
-      return false;
-
-    v8::Local<v8::Value> val;
-    return GetHandle()->Get(context, v8_key).ToLocal(&val);
-  }
-
   template <typename T>
   bool Get(const base::StringPiece& key, T* out) const {
     // Check for existence before getting, otherwise this method will always
--- a/shell/browser/api/atom_api_web_request.cc
+++ b/shell/browser/api/atom_api_web_request.cc
@@ -4,6 +4,7 @@
 
 #include "shell/browser/api/atom_api_web_request.h"
 
+#include <set>
 #include <string>
 #include <utility>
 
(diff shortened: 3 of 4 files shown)


TongDaDa:
It just has include-mode match rules using the pattern, What about a feature that can exclude URLs?

Just like host-rules for example:

```
   "--host-rules" : 'MAP file://* http://*, EXCLUDE */catch_resources/*'
```

It reflects all requesting that starting with 'file' protocol to 'HTTP' protocol and excludes requesting URLs from `*/catch_resources/`.

Because I have a [problem](https://discuss.atom.io/t/file-protocol-block-resources-from-com/66786) that needs it to resolve exactly.

spec-main/api-net-spec.ts
```diff
@@ -633,6 +633,24 @@ describe('net module', () => {
...
+
+      it('Should not throw when valid filters are passed', () => {
+        expect(() => {
+          session.defaultSession.webRequest.onBeforeRequest(
+            { urls: ['*://www.googleapis.com/'] },
+            (details, callback) => { callback({ cancel: false }) }
+          )
+        }).to.not.throw()
+      })
+
```
ckerr: It would also be good to see a test with > 1 url in the filter array

shell/browser/api/atom_api_web_request.cc
```diff
@@ -84,7 +68,24 @@ void WebRequest::SetListener(Method method, Event type, mate::Arguments* args) {
   // { urls }.
   URLPatterns patterns;
   mate::Dictionary dict;
-  args->GetNext(&dict) && dict.Get("urls", &patterns);
+  if (args->GetNext(&dict)) {
+    std::set<std::string> filter_patterns;
+    if (dict.HasKey("urls") && dict.Get("urls", &filter_patterns)) {
+      for (const std::string& filter_pattern : filter_patterns) {
```
ckerr: _(minor)_ This block nests the control flow four layers deep -- if-if-for-if -- making the code slightly more complicated without any benefit. If the "get the patterns" and "parse the patterns" steps are separated, the code becomes simpler and loses two nesting layers:

```c++
// get the filter patterns
std::set<std::string> filter_patterns;
mate::Dictionary dict;
args->GetNext(&dict) && dict.Get("urls", &filter_patterns);

// parse the filter patterns
URLPattern pattern (URLPattern::SCHEME_ALL);
for (const std::string& filter_pattern : filter_patterns) {
  const URLPattern::ParseResult result = pattern.Parse(...);
  if (result == URLPattern::ParseResult::kSuccess) ...
}
```

shell/browser/api/atom_api_web_request.cc
```diff
@@ -84,7 +68,24 @@ void WebRequest::SetListener(Method method, Event type, mate::Arguments* args) {
...
+        URLPattern pattern = URLPattern(URLPattern::SCHEME_ALL);
+        URLPattern::ParseResult success = pattern.Parse(
```
ckerr:
```suggestion
        const URLPattern::ParseResult result = pattern.Parse(
```

docs/api/web-request.md
```diff
@@ -68,6 +68,21 @@ The `uploadData` is an array of `UploadData` objects.
...
+'http://foo:1234/bar'
+'*://*/*'
+'*://example.com/*'
+'*://example.com/foo/*'
+'http://*.foo:1234/'
+'file://foo:1234/bar'
+'http://foo:*/'
+'*://www.foo.com/'
+```
+
```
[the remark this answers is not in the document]
codebytere: This is a good question! In writing this, i imagined it more as a failure on our end that we were previously failing silently, in that it lead to false user assumptions and was therefore a fix although I think it could also easily and accurately be positioned as breaking change as well 🤔

shell/browser/api/atom_api_web_request.cc
```diff
@@ -84,7 +68,25 @@ void WebRequest::SetListener(Method method, Event type, mate::Arguments* args) {
   // { urls }.
   URLPatterns patterns;
   mate::Dictionary dict;
-  args->GetNext(&dict) && dict.Get("urls", &patterns);
+  std::set<std::string> filter_patterns;
+
+  if (args->GetNext(&dict) && !dict.Get("urls", &filter_patterns))
+    args->ThrowError(
+        "onBeforeRequest parameter 'filter' must have property 'urls'");
```
ckerr: Do we also want to return here?

```c++
if (filter && !filter.urls) {
  args->ThrowError();
  return;
}
```

ckerr:
```suggestion
        "onBeforeRequest parameter 'filter' must have property 'urls'.");
```

native_mate/native_mate/dictionary.h
```diff
@@ -41,7 +41,7 @@ class Dictionary {
   static Dictionary CreateEmpty(v8::Isolate* isolate);
 
   template <typename T>
-  bool Get(const base::StringPiece& key, T* out) const {
+  bool Get(base::StringPiece key, T* out) const {
```
ckerr: Would you be OK with reverting the `Get()` signature change and letting #19432 handle that? It's more on-topic to 19432 than here in 19337

deepak1556:
@TongDaDa that feature is outside the scope of this PR, although its good to have. Please create a separate issue for it. Thanks!

As a side note, that mapping from file to http is scary.

TongDaDa:
> @TongDaDa that feature is outside the scope of this PR, although its good to have. Please create a separate issue for it. Thanks!
> 
> As a side note, that mapping from file to http is scary. Any reason you are doing that ? Can discuss that in the above issue once you create it.

That feature I have issued, but It has been closed by @codebytere

ckerr:
@TongDaDa you seem to be referring to #19223, which @codebytere closed as a duplicate of #11371 after both another user and *you yourself* agreed that it was a duplicate. Closing duplicate issues is perfectly reasonable and your description of that behavior as "trenchant" is inappropriate. Actually you described the person doing it, which is even more inappropriate. Don't do that.

If you have an issue that is not a duplicate of 11371, please open a separate issue for it and perhaps this time don't describe it as a duplicate.

TongDaDa:
> @TongDaDa you seem to be referring to #19223, which @codebytere closed as a duplicate of #11371 after both another user and _you yourself_ agreed that it was a duplicate. Closing duplicate issues is perfectly reasonable and your description of that behavior as "trenchant" is inappropriate. Actually you described the person doing it, which is even more inappropriate. Don't do that.
> 
> If you have an issue that is not a duplicate of 11371, please open a separate issue for it and perhaps this time don't describe it as a duplicate.

Oh, oh, I'm sorry @codebytere, I'm so sorry.

I was the first to use `trenchant ` the word that I don't know what it means really.

And another sentence that `Please fix it as soon as possible` has imperative mood, please don't mind, of course, that is not your responsibility. (😢My English is not good).

As for # 19223, It is my workaround after I had submitted an [issue](https://github.com/electron/electron/issues/19200) that is mentioned of my comment above and processed by another worker of Electron. and this is my fault, It was I who confused the two men.

PR #19346, opened 2019-07-19
build: use circleci v2.1

MarshallOfSound:
This allows us to have a workflow for release-builds that only runs when we trigger it via API.  This will be used to implement a `/ci` command to allow folks to trigger release-style builds on their PRs.

Tested manually with the API and this does what we want.

Notes: no-notes


codebytere:
Build failure caused by manual API trigger with wrong parameter.

Issue #18019, opened 2019-04-28
Reload nested iframe in webview causes memory leak in electron@5.0.0


KochiyaOcean:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.0
* **Operating System:**
  * Windows 10 1809
* **Last Known Working Electron version:**:
  * 4.x.x

### Expected Behavior

If reload a nested iframe inside webview, the old iframe will be removed from memory.

### Actual Behavior

The old iframe is still existed in memory and still working (e.g, still playing videos)

### To Reproduce
Clone an reproduce example.

```sh
$ git clone https://github.com/KochiyaOcean/electron-bugs.git -b webview-nested-iframe-memory-leak
$ electron .
```

Play youtube video in the webview.

Run `$('webview').executeJavaScript("document.querySelector('iframe').src = document.querySelector('iframe').src")` in devtools.

The iframe is reloaded, but the former iframe is still existed. The video in former iframe is still playing, and the memory of former iframe is not recycled.

Which means:

```
BrowserWindow
  └── Webview
      └── iframe0
          └── iframe1 <= reload this one
              └── iframe2 <= and this one is not destroyed
```

### Screenshots
![image](https://user-images.githubusercontent.com/8194131/56866010-a2ce6600-6a0f-11e9-9390-027b01588e17.png)

### Additional Information
There is another problem seems related: If disable webview's webSecurity option, the CORS policy of nested iframe inside the webview will still keep working, which should be also disabled.

```
BrowserWindow
  └── Webview <= disable webSecurity
      └── iframe0 <= CORS policy disabled
          └── iframe1 <= CORS policy still working
              └── iframe2 <= Cant'be controlled by upper iframe if it doesn't share same origin with upper iframe
```

Nick-lab:
you probably dont want to be disabling webSecurity in you app in the first place what exactly are you trying to accomplish with nested iframes like that anyway? maybe you should look into single page application frameworks such as angular or ionic to manage paging in your app. This is much more secure and you gain alot of control over your app this way.

KochiyaOcean:
[...]

PR #19347, opened 2019-07-19
fix: backport patch for leaking HTML5 video elements

MarshallOfSound:
Backports https://chromium-review.googlesource.com/c/chromium/src/+/1677053 to `4-2-x` to fix a rather noticeable memory leak when removing HTML5 video elements from the DOM

Notes: Fixed memory leak when removing HTML5 video elements from the DOM with Javascript


MarshallOfSound:
@codebytere Those changes are due to the different versions of `git` folks have on their machines (namely me and @nornagon have different git versions) we fixed this in the patch scripts in more recent branches (I think `6-0-x` and above).  The changes are ok 👍

MarshallOfSound:
I just ran git-export-patches.  I'd rather avoid manually editing patch files by hand or picking and choosing.

jkleinsc:
@MarshallOfSound I get that but couldn't you be selective on what files you commit?

MarshallOfSound:
@jkleinsc those hunks generate the sha that my newly added patch is based off of.  It might work, but it might not.  Id rather just take the output of git-export-patches.

MarshallOfSound:
test failures look video related, I'll take a look

MarshallOfSound:
@jkleinsc @codebytere Tests are fixed, should be All Clear now.  Needed to handle a `nullptr` case that the original CL did not as Chromium has removed legacy media inputs in C75, we still have them in C69 👍

codebytere:
Approving given added context above!

deepak1556:
> Needed to handle a nullptr case that the original CL did not as Chromium has removed legacy media inputs in C75, we still have them in C69

Can you share the CL that changed this ? I am concerned that `volume_slider_` is expected to be non-null and while it doesn't. There are other locations in the code where they null protect against `volume_slider_`. 

Did you try running with `--enable-features=UseModernMediaControls` to check if its the legacy controls thats causing this difference ?

peteringram0:
@MarshallOfSound Does this also fix the following bug: https://github.com/electron/electron/issues/18019 ? Thanks

PR #19370, opened 2019-07-22
feat: add getBounds() method for BrowserView

### LINKED ISSUE #18734: Add getBounds for BrowserView like BrowserWindow
### Feature Request

The only current way to get the parameters set by ``setBounds`` of a ``BrowserView`` seems to be to store it as a global variable at creation, and to reference that variable. Ideally, there should be a way to get this information from the created ``BrowserView`` directly with ``getBounds``, like the ``BrowserWindow``. This is particularly important if the bounds of the ``BrowserView`` has been changed by ``setAutoResize`` after creation, because the bounds no longer match a stored global variable.

DeerMichel:
#### Description of Change
Closes #18734. 

This PR extends the functionality of our experimental `BrowserView` API by adding a `getBounds()` method to get the current size & position.

cc @codebytere @erickzhao 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `getBounds()` method for BrowserView.

### DIFF
--- a/docs/api/browser-view.md
+++ b/docs/api/browser-view.md
@@ -94,6 +94,12 @@ Returns `Boolean` - Whether the view is destroyed.
 
 Resizes and moves the view to the supplied bounds relative to the window.
 
+#### `view.getBounds()` _Experimental_
+
+Returns [`Rectangle`](structures/rectangle.md)
+
+The `bounds` of this BrowserView instance as `Object`.
+
 #### `view.setBackgroundColor(color)` _Experimental_
 
 * `color` String - Color in `#aarrggbb` or `#argb` form. The alpha channel is
--- a/shell/browser/api/atom_api_browser_view.cc
+++ b/shell/browser/api/atom_api_browser_view.cc
@@ -127,6 +127,10 @@ void BrowserView::SetBounds(const gfx::Rect& bounds) {
   view_->SetBounds(bounds);
 }
 
+gfx::Rect BrowserView::GetBounds() {
+  return view_->GetBounds();
+}
+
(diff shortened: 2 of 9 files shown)


docs/api/browser-view.md
```diff
@@ -94,6 +94,10 @@ Returns `Boolean` - Whether the view is destroyed.
 
 Resizes and moves the view to the supplied bounds relative to the window.
 
+#### `view.getBounds()` _Experimental_
+
+Returns [`Rectangle`](structures/rectangle.md)
```
codebytere: For extra context, we might say:
```
Returns [`Rectangle`](structures/rectangle.md)

The `bounds` of this BrowserView instance as `Object`.
```

DeerMichel: Either way it should be consistent with [BrowserWindow#getBounds](https://github.com/electron/electron/blob/master/docs/api/browser-window.md#wingetbounds).

codebytere: I would prefer we update it there as well in that case, since it's the case for [Tray bounds](https://github.com/electron/electron/blob/24673501804bc372b73577896cf00cb6282bf667/docs/api/tray.md#L280-L284) and for documentation, more verbosity is generally preferable to less within reason.

DeerMichel: oki, then i'll add it on both docs later

codebytere:
@DeerMichel once the conflicts are resolved this should be good to go :)

rlingineni:
@codebytere - did this ship yet?

deermichel:
It will ship with [Electron v7](https://electronjs.org/docs/tutorial/electron-timelines) 💯

PR #19372, opened 2019-07-22
build: make //electron:electron_lib a source_set

nornagon:
#### Description of Change
We don't need the `libelectron_lib.a` file for anything so we might as well skip creating it. Should make rebuilds a few seconds faster.

From GN [documentation](https://chromium.googlesource.com/chromium/src/tools/gn/+/48062805e19b4697c5fbd926dc649c78b6aaa138/docs/language.md):

> source_set: A lightweight virtual static library (usually preferrable over a real static library since it will build faster).

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -337,7 +337,7 @@ templated_file("electron_version_header") {
   args_files = get_target_outputs(":electron_version_args")
 }
 
-static_library("electron_lib") {
+source_set("electron_lib") {
   configs += [ "//v8:external_startup_data" ]
   configs += [ "//third_party/electron_node:node_internals" ]
 


deepak1556:
We don't need the static library file but we do need the intermediate object library file atleast on windows for v8 symbols https://github.com/electron/electron/blob/master/appveyor.yml#L112

nornagon:
@deepak1556 I'm pretty sure that `.lib` file comes with `electron.exe`, not `electron_lib`.

MarshallOfSound:
Both windows jobs failed on the same test 🤔

nornagon:
Doesn't seem related to this PR

deepak1556:
> source_set: A lightweight virtual static library (usually preferrable over a real static library since it will build faster).

That definition about source_set is misleading, please refer here for better clarity https://gn.googlesource.com/gn/+/master/docs/reference.md#func_source_set

PR #19375, opened 2019-07-23
chore: clean up BrowserView/TopLevelWindow New impl

codebytere:
#### Description of Change

Idiomatically, JavaScript ignores extra arguments passed to a function if they're not used or needed, and in our `BrowserView::New` impl we threw an error if there were too many. To better align with  behavior that JavaScript users would expect, this PR thus defaults the `options` dict to an empty one, and attempts to get an `options` dict param should it be passed.

Also defaults the options dict in `TopLevelWindow::New` even though we didn't throw an error there. 

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/atom_api_browser_view.cc
+++ b/shell/browser/api/atom_api_browser_view.cc
@@ -102,15 +102,8 @@ mate::WrappableBase* BrowserView::New(mate::Arguments* args) {
     return nullptr;
   }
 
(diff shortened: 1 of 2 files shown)


codebytere:
Failure is a known flake.

PR #19392, opened 2019-07-23
fix: "TypeError: fn is not a function" crash in release notes generator

ckerr:
#### Description of Change

 * Fix a bug introduced in https://github.com/electron/electron/pull/18085. In that code, the two functions calling `runRetryable(fn)` seem to have always passed in a Promise, rather than a Promise-generating-function, as the retryable function `fn`. It seems this threw a `TypeError: fn is not a function` on every call? :boom: 

Fixed a pair of secondary issues in the related code:

 * octokit has deprecated use of `number` in the `octokit.issues.listComments()` and `octokit.pulls.get()` API, so this PR also uses `issue_number` and `pull_number` instead.

 * minor fix: the cached file names of issue comments mistakenly used the word `pull` rather than `issue`. I don't think this caused any issues, but it's cleaner to use the right name.

CC @codebytere 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/release/notes/notes.js
+++ b/script/release/notes/notes.js
@@ -319,25 +319,14 @@ async function runRetryable (fn, maxRetries) {
 
 const getPullRequest = async (number, owner, repo) => {
   const name = `${owner}-${repo}-pull-${number}`
-  return checkCache(name, async () => {
-    return runRetryable(octokit.pulls.get({
-      number,
-      owner,
-      repo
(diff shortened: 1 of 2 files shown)


ckerr:
Uh oh, I built this in a dirty sandbox. Closing this PR and will re-open with a clean one.

Issue #19389, opened 2019-07-23
flaky test: net module ClientRequest API should be able to pipe a readable stream into a net request


nornagon:
This test started failing regularly on Windows builds, e.g. https://ci.appveyor.com/project/electron-bot/electron-ljo26/builds/26172206

```
Unhandled exception in main spec runner: AssertionError: Unexpected URL: /200
    at C:\projects\src\electron\spec-main\api-net-spec.ts:45:14
    at Server.<anonymous> (C:\projects\src\electron\spec-main\api-net-spec.ts:27:7)
    at Server.emit (events.js:203:13)
    at parserOnIncoming (_http_server.js:713:12)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:116:17)
```

cc @zcbenz and @deepak1556 who have been doing network service stuff recently.

deepak1556:
Spent some time on this, here are my findings:

1) There were no code changes done in the legacy net module code path , also checked upstream for changes in `//net/url_request` which we use for the module.

2) Unable to repro locally, my only hypothesis at the moment is it might be related to https://github.com/electron/electron/issues/19368 where the responses to previous requests are returned at unexpected times causing this flake.

Lets wait for a few more test runs on latest master to investigate further.

PR #19393, opened 2019-07-23
test: ensure cleanup of net requests in tests

nornagon:
#### Description of Change
Ref #19389. Not sure if this fixes that but it seems like a good idea anyway.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


deepak1556:
Hmm a little unsure, this might hide the fact that why request was being leaked in the first place.

nornagon:
we should maybe test that explicitly? so we get a focused test failure instead of a broad "something broke"

PR #19394, opened 2019-07-23
fix: "TypeError: fn is not a function" crash in release notes generator

ckerr:
#### Description of Change

 * Fix a bug introduced in https://github.com/electron/electron/pull/18085. In that code, the two functions calling `runRetryable(fn)` seem to have always passed in a Promise, rather than a function-which-returns-a-promise, as the retryable function `fn`. It seems this threw a `TypeError: fn is not a function` on every call? :boom: 

Fixed a pair of secondary issues in the related code:

 * octokit has deprecated use of `number` in the `octokit.issues.listComments()` and `octokit.pulls.get()` API, so this PR also uses `issue_number` and `pull_number` instead.

 * minor fix: the cached file names of issue comments mistakenly used the word `pull` rather than `issue`. I don't think this caused any issues, but it's cleaner to use the right name.

CC @codebytere 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/release/notes/notes.js
+++ b/script/release/notes/notes.js
@@ -329,7 +329,7 @@ const getPullRequest = async (number, owner, repo) => {
 }
 
 const getComments = async (number, owner, repo) => {
-  const name = `${owner}-${repo}-pull-${number}-comments`
+  const name = `${owner}-${repo}-issue-${number}-comments`
   return checkCache(name, async () => {
     return runRetryable(octokit.issues.listComments({
       number,
--- a/script/release/notes/notes.js
+++ b/script/release/notes/notes.js
@@ -320,13 +320,13 @@ async function runRetryable (fn, maxRetries) {
 const getPullRequest = async (number, owner, repo) => {
   const name = `${owner}-${repo}-pull-${number}`
   const retryableFunc = () => octokit.pulls.get({ pull_number: number, owner, repo })
-  return checkCache(name, async () => runRetryable(retryableFunc, MAX_FAIL_COUNT))
+  return checkCache(name, () => runRetryable(retryableFunc, MAX_FAIL_COUNT))
 }
 
 const getComments = async (number, owner, repo) => {
   const name = `${owner}-${repo}-issue-${number}-comments`
   const retryableFunc = () => octokit.issues.listComments({ issue_number: number, owner, repo, per_page: 100 })
-  return checkCache(name, async () => runRetryable(retryableFunc, MAX_FAIL_COUNT))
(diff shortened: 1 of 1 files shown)


script/release/notes/notes.js
```diff
@@ -319,25 +319,14 @@ async function runRetryable (fn, maxRetries) {
...
   const name = `${owner}-${repo}-pull-${number}`
-  return checkCache(name, async () => {
-    return runRetryable(octokit.pulls.get({
-      number,
-      owner,
-      repo
-    }), MAX_FAIL_COUNT)
-  })
+  const retryableFunc = () => octokit.pulls.get({ pull_number: number, owner, repo })
+  return checkCache(name, async () => runRetryable(retryableFunc, MAX_FAIL_COUNT))
```
codebytere: Since `octokit.pulls.get` returns a promise, we'd want to `await runRetryable(retryableFunc, MAX_FAIL_COUNT))` here, otherwise the `async` keyword wouldn't be needed

script/release/notes/notes.js
```diff
@@ -319,25 +319,14 @@ async function runRetryable (fn, maxRetries) {
...
-    return runRetryable(octokit.issues.listComments({
-      number,
-      owner,
-      repo,
-      per_page: 100
-    }), MAX_FAIL_COUNT)
-  })
+  const name = `${owner}-${repo}-issue-${number}-comments`
+  const retryableFunc = () => octokit.issues.listComments({ issue_number: number, owner, repo, per_page: 100 })
+  return checkCache(name, async () => runRetryable(retryableFunc, MAX_FAIL_COUNT))
```
codebytere: same as above!

script/release/notes/notes.js
ckerr: To put this at the front: you're right, the second param to `checkCache()` doesn't need to be an async function here. It will still work, and be more readable, if we remove the `async` keyword from the second line here..

That said, as an academic exercise I'm wondering why this works in 67311dd. `async() => runRetryable(retryableFunc, MAX_FAIL_COUNT)` is an async function which returns a promise, so `await operation()` in `checkCache()` seems like it should resolve with the inner promise, rather than the octokit response object. So why does the octokit response object get cached instead of a pending promise? :thinking:

ckerr: answer: because if you `await` a Promise has an inner Promise, the `await` will fulfill with the inner Promise's value.

If x is a promise, adopt its state [3.4]:
If x is pending, promise must remain pending until x is fulfilled or rejected.
If/when x is fulfilled, fulfill promise with the same value.

https://promisesaplus.com/#point-49

PR #19403, opened 2019-07-23
build: run node tests on CI

MarshallOfSound:
This PR does a few things

* Runs the node.js test suite on our CI 🎉 
* Ignores all the node.js tests that are currently failing (around 300 / 3000)
* Fixes an issue where `ELECTRON_RUN_AS_NODE` would not parse CLI flags correctly because the v8 isolate was initialized before SetFlagsFromCommandLine was called inside `node::Init`.  A quick re-ordering fixed that issue 👍 

Notes: `ELECTRON_RUN_AS_NODE` now correctly parses both node options and v8 flags.  E.g. `--expose_gc`


script/node-spec-runner.js
```diff
@@ -0,0 +1,35 @@
...
+
+const BASE = path.resolve(__dirname, '../..')
+const NODE_DIR = path.resolve(BASE, 'third_party', 'electron_node')
+const NPX_CMD = process.platform === 'win32' ? 'npx.cmd' : 'npx'
+
+const utils = require('./lib/utils')
+const { YARN_VERSION } = require('./yarn')
+
+if (!process.mainModule) {
+  throw new Error('Must call the node spec runner directly')
```
nornagon: why check this?

patches/node/chore_handle_default_configuration_not_being_set_in_the_electron_env.patch
```diff
@@ -0,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <sattard@slack-corp.com>
+Date: Tue, 23 Jul 2019 11:36:48 -0700
+Subject: chore: handle default_configuration not being set in the electron env
```
nornagon: should we rather set this up in the env?

script/node-spec-runner.js
MarshallOfSound: I copied from `nan-spec-runner.js`.  Don't want folks randomly requir'ing this file by accident (it'll just run the tests on-require)

patches/node/chore_handle_default_configuration_not_being_set_in_the_electron_env.patch
MarshallOfSound: _mmaayybbeeee_ but our config.gypi doesn't currently support dynamic changes, happy to raise a TODO though to update our GN files to generate the config.gypi 😆

MarshallOfSound:
@jkleinsc I did some tricks and now node failures are reported through Circle

image

PR #19404, opened 2019-07-23
build: unify YARN_VERSION variable usage and ensure CI uses yarn not npm

jkleinsc:
#### Description of Change

Manual backport of #18607. The lint job was failing on CI in 6-0-x because we were using npm instead of yarn.  This PR changes our CI lint job to use yarn.
#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -553,8 +553,8 @@ steps-lint: &steps-lint
           # but then we would lint its contents (at least gn format), and it doesn't pass it.
 
           cd src/electron
-          npm install
-          npm run lint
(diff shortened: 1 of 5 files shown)


MarshallOfSound:
@jkleinsc This file has a bunch of `npm install` in it that aren't on master, we should remove all of them and replace with yarn to ensure we use the lock file correctly

jkleinsc:
@ckerr because we have a yarn.lock file that ensures the right dependencies for linting.  We no longer have a package.json.lock file so therefore it installs a dependency that breaks lint.

PR #19420, opened 2019-07-24
feat: add about panel customization on Windows

### LINKED ISSUE #15591: Cross-platform 'About' window

An 'About' window is something every app needs, yet Electron only has a native About window for macOS (the default one). For Windows and Linux, every Electron app has to reinvent the wheel, usually badly, and not adhering to platform-specific behavior nor style.

GTK actually has a widget specifically for this: https://developer.gnome.org/gtk3/stable/GtkAboutDialog.html

For Windows, from what I can tell, there's no standard, so here it could just be a simple native modal window with the icon and a few labels for the information.

**Describe alternatives you've considered**

The alternative is for every app to create a custom browser window and style it with CSS. The problem is that it's hard to imitate the platform native style on Windows, and near impossible to do it on Linux, as every desktop manager has their own style.

erickzhao:
#### Description of Change
Closes #15591. Adds support for `showAboutPanel()` and `setAboutPanelOptions()` on Windows.

Windows doesn't have a native About dialog. This PR reuses the `electron::ShowMessageBox` function to generate a dialog with the About Panel information as a string message.

Notes:
* Default settings are the `applicationName` and `applicationVersion` provided by [`/src/base/file_version_info.h`](https://cs.chromium.org/chromium/src/base/file_version_info.h).
* The image path accepts any file types that `NativeImage` API does.
* The About panel is synchronous (same as on Linux).

Fields added (all optional):
* `applicationName`: string
* `applicationVersion`: string
* `copyright`: string
* `credits`: string
* `iconPath`: string

cc @codebytere @ckerr @sindresorhus 

#### Example
image

#### Additional Reading

Microsoft UX Guidelines for Dialog Boxes: https://docs.microsoft.com/en-us/windows/win32/uxguide/win-dialog-box

Relevant bits:
> * Use an icon in About Box dialogs for application branding.
> About Dialog Example
> In this example, a bitmap is used in the About Box to identify and brand the application.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added about panel customization on Windows.

### DIFF
--- a/shell/browser/browser_win.cc
+++ b/shell/browser/browser_win.cc
@@ -22,6 +22,7 @@
 #include "base/win/win_util.h"
 #include "base/win/windows_version.h"
 #include "electron/electron_version.h"
+#include "shell/browser/ui/message_box.h"
 #include "shell/browser/ui/win/jump_list.h"
 #include "shell/common/application_info.h"
 #include "shell/common/native_mate_converters/string16_converter.h"
@@ -367,6 +368,20 @@ void Browser::ShowEmojiPanel() {
   input[3].ki.wVk = ui::WindowsKeyCodeForKeyboardCode(ui::VKEY_OEM_PERIOD);
   input[3].ki.dwFlags |= KEYEVENTF_KEYUP;
   ::SendInput(4, input, sizeof(INPUT));
+}
+
+void Browser::ShowAboutPanel() {
+  electron::MessageBoxSettings settings = {};
+  settings.title = "About";
+  settings.message = "Testing 123";
+  settings.detail = "Detail?";
+  settings.checkbox_label = "checkbox label";
+  electron::ShowMessageBoxSync(settings);
+  // const auto& opts = about_panel_options_;
+}
+
+void Browser::SetAboutPanelOptions(const base::DictionaryValue& options) {
+  options.Clone();
 }
 
 }  // namespace electron
--- a/shell/browser/browser_win.cc
+++ b/shell/browser/browser_win.cc
@@ -85,6 +85,16 @@ bool FormatCommandLineString(base::string16* exe,
   return true;
 }
 
+std::unique_ptr<FileVersionInfo> FetchFileVersionInfo() {
+  base::FilePath path;
+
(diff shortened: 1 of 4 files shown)


codebytere:
@erickzhao this is blocked on https://github.com/electron/electron/pull/19460, right?

erickzhao:
> @erickzhao this is blocked on #19460, right?
@codebytere yep! I was waiting for that one to get merged to avoid having to re-implement all the business logic available in `AddImageSkiaRepFromPath`.

shell/browser/browser_win.cc
```diff
@@ -369,4 +371,37 @@ void Browser::ShowEmojiPanel() {
...
+    }
+
+    if ((str = opts.FindStringKey("iconPath"))) {
+      base::FilePath path = base::FilePath::FromUTF8Unsafe(*str);
+      electron::util::PopulateImageSkiaRepsFromPath(&image, path);
+    }
+  }
+
+  electron::MessageBoxSettings settings = {};
+  settings.title = "About";
```
DeerMichel: Hardcoding the title won't let users localize their app - can we get around this problem?

erickzhao: I think the easiest way would be to omit the "About" string entirely. This would be a cross-platform change, since it is omitted on macOS and Linux.

However, I don't have too much context on if that's a correct native Windows pattern.

codebytere: we can also use the l10n utility, potentially!

erickzhao: Looked a bit through the l10n stuff, but didn't seem to find any strings that would match in here: https://cs.chromium.org/chromium/src/ui/strings/translations/ui_strings_en-GB.xtb

🤔 think i might just stick to removing it

codebytere: in that case, that's fine imo!

erickzhao:
Update: added default values for `applicationName` and `applicationVersion` based on existing Chromium APIs. :)

erickzhao:
> Are you aware that the dialog's title will be derived from app.name?
Should be okay this way!

Issue #19424, opened 2019-07-24
js garbage collector destroys active popup menus


CezaryKulakowski:
Javascript's garbage collector destroys popup menus created by menu.popup from api Menu even when they are still active. It can lead to a crash when during destruction phase (when js object is already destroyed but c++ instance of class electron::api::Menu still lives) user makes any interaction with it, like changing focused element.

CezaryKulakowski:
To reproduce the problem it's enough to create simple application which creates context menu created by api Menu (I've used the one from https://www.npmjs.com/package/electron-context-menu). When menu is opened it will be automatically closed after first run of js garbage collector (it's being launched once a minute). Additionaly if we keep using the menu (like pressing arrow down to cycle through menu entries) crash may occur when menu is going to be destroyed (~50% reproducibility rate).

CezaryKulakowski:
I've created pr with working solution: https://github.com/electron/electron/pull/19427

DeerMichel:
Do you hold a global reference to the `Menu` on the JS side? It sounds a bit to me like [here](https://github.com/electron/fiddle/pull/221/files) where the references lives in local scope and naturally goes out of scope after the function returns.

CezaryKulakowski:
No. I agree that this problem is caused by a faulty implementation of popup menu in package electron-context-menu, but bad js shouldn't cause a crash in v8.

CezaryKulakowski:
The other solution would be checking if GetWrapper() isn't null in all calls (for example: in api::Menu::GetAcceleratorForCommandIdWithParams) but I don't know if it's a better solution.

CezaryKulakowski:
To make a full picture: when js object has been destroyed by gc we get information on c++ side. Then c++ object is being destroyed in two phases: in first run we reset wrapper to js object and in second run we destroy the c++ object. If any call to the menu which still exists and is visible happens after step one but before step two we will get a crash in v8. It's quite easy to reproduce.

PR #19427, opened 2019-07-24
chore: keep references to active menus to prevent gc

CezaryKulakowski:
Fixes https://github.com/electron/electron/issues/19424.

### How to reproduce
Launch this simple application:

```
const { app, BrowserWindow, Menu } = require('electron')

function createWindow () {
  let win = new BrowserWindow({
    width: 800,
    height: 600
  })
  let menu = Menu.buildFromTemplate(
    [
      { label: 'one', accelerator: 'Control+1' },
      { label: 'two', accelerator: 'Control+2' }
    ]
  )
  menu.popup({ "x": 50, "y": 50})
}

app.on('ready', createWindow)
```
After it's started press arrow down to cycle through both menu's entries.

### Expected behavior

Nothing bad happens. 

### Actual behavior

After few seconds - when js's garbage collector is started  - menu is being destroyed. Additionaly crash may occur (~50% reproducibility rate).

As object `menu` is not kept as global reference in js garbage collector destroys it. It's a matter of debate if that's correct behaviour (on one side object is "local" on js side but on the other menu is active and displayed). The bigger issue is that js object is being destroyed while c++ object is still alive and active. If user interacts with it (like changing focus of current element) when js object is already destroyed crash in v8 will occur. We get notification on c++ side that js object was destroyed by callback but the problem is that - by design - all instances of class `WrappableBase`, like `Menu`, are destroyed in two steps: first `WrappableBase::FirstWeakCallback` is called where we clear wrapper to js object, and after that `WrappableBase::SecondWeakCallback` is called where we destroy c++ object. Crash occurs when there is any call to Menu (like `Menu::GetAcceleratorTextAt` after focused element was changed) after first callback but before second callback was called.

With proposed fix we keep references to all currently displayed menus so garbage collector doesn't destroy active menus. If we decide that this is expected behaviour that popup can be destroyed by js garbage collector even when it's still displayed the solution would be to check if `GetWrapper()` doesn't return empty object in all calls in class `Menu` (e.g. `Menu::IsCommandIdChecked`, `Menu::GetAcceleratorTextAt`, ` Menu::IsCommandIdEnabled` ...).

notes: Don't destroy active menus created as local objects in javascript

### DIFF
--- a/shell/browser/api/atom_api_menu.cc
+++ b/shell/browser/api/atom_api_menu.cc
@@ -4,6 +4,8 @@
 
 #include "shell/browser/api/atom_api_menu.h"
 
+#include <map>
+
 #include "native_mate/constructor.h"
 #include "native_mate/dictionary.h"
 #include "native_mate/object_template_builder.h"
@@ -13,6 +15,13 @@
 #include "shell/common/native_mate_converters/image_converter.h"
 #include "shell/common/native_mate_converters/string16_converter.h"
 #include "shell/common/node_includes.h"
+
+namespace {
+// We need this map to keep references to currently opened menus.
+// Without this menus would be destroyed by js garbage collector
+// even when they are still displayed.
+std::map<uint32_t, v8::Global<v8::Object>> g_menus;
+}  // unnamed namespace
 
 namespace electron {
 
@@ -200,10 +209,12 @@ bool Menu::WorksWhenHiddenAt(int index) const {
 }
 
 void Menu::OnMenuWillClose() {
+  g_menus.erase(weak_map_id());
   Emit("menu-will-close");
 }
 
 void Menu::OnMenuWillShow() {
+  g_menus[weak_map_id()] = v8::Global<v8::Object>(isolate(), GetWrapper());
   Emit("menu-will-show");
 }
 


codebytere:
@CezaryKulakowski thanks for this! Before we review it, would you please fill out the template properly with a more complete description, a filled template, and Release Notes as described in the template comment? We'll be able to review more effectively once that has been completed.

CezaryKulakowski:
@codebytere please let me know if updated description is sufficient :).

shell/browser/api/atom_api_menu.cc
```diff
@@ -14,6 +16,13 @@
 #include "shell/common/native_mate_converters/string16_converter.h"
 #include "shell/common/node_includes.h"
 
+namespace {
+// We need this map to keep references to currently opened menus.
+// Without this menus would be destroyed by js garbage collector
+// even when they are still displayed.
+std::map<uint32_t, v8::Global<v8::Object>> g_menus;
```
MarshallOfSound: Maybe it's because it's early in the morning, but I'm not 100% convinced this will free the memory.  cc @deepak1556 will this leak the objects?

deepak1556: This would have leaked, but its being manually tracked in menu creation and close. So its fine.

deepak1556:
> it's probably not the best idea to tell users to retain the reference globally

Coming back to the original problem outlined, I would say the current behavior expecting a global reference for menu aligns with what we expect for BrowserWindow. These are objects that are meant to be maintained for the lifetime of the app or however the user decides in the main process, tying it to a particular scope would be difficult. Being explicit about this behavior is fine IMO.


> If we decide that this is expected behaviour that popup can be destroyed by js garbage collector even when it's still displayed the solution would be to check if GetWrapper() doesn't return empty object in all calls in class

@CezaryKulakowski this sounds like a good thing to have.

/cc @zcbenz who wrote the original implementation.

zcbenz:
> If we decide that this is expected behaviour that popup can be destroyed by js garbage collector even when it's still displayed the solution would be to check if `GetWrapper()` doesn't return empty object in all calls in class `Menu`

It is hard to make things still work correctly when `GetWrapper()` is empty, if we return empty results in menu methods then the menu may end up in weird state, if we still execute the delegate methods then JS exceptions may happen as the menu object is gone.

codebytere:
@CezaryKulakowski could you potentially backport this to the above branches? I can try to get to it if not

CezaryKulakowski:
@codebytere I will take care of this today or on Monday.

CezaryKulakowski:
I've checked that this issue doesn't reproduce either on 6-1-x or 7-1-x. @codebytere do you have another test case for this?

codebytere:
@CezaryKulakowski the original issue was opened for 6 though? either way i think we should backport for safety's sake. I'll handle it.

PR #19435, opened 2019-07-24
fix: clearing of backgroundColor property on TouchBar items

miniak:
#### Description of Change
The code parsing some of the properties simply ignores them when the value is not valid (or `null`) instead of setting the native value to `nil`, which is fine when the TouchBar item is being constructed. But if you later want to set the `backgroundColor` to default by setting it to `null`, it does not work.

/cc @codebytere

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Fix clearing of the `backgroundColor` property on `TouchBar` items.

### DIFF
--- a/shell/browser/ui/cocoa/atom_touch_bar.mm
+++ b/shell/browser/ui/cocoa/atom_touch_bar.mm
@@ -357,8 +357,11 @@ static NSString* const ImageScrubberItemIdentifier = @"scrubber.image.item";
   NSButton* button = (NSButton*)item.view;
 
   std::string backgroundColor;
-  if (settings.Get("backgroundColor", &backgroundColor)) {
+  if (settings.Get("backgroundColor", &backgroundColor) &&
+      !backgroundColor.empty()) {
     button.bezelColor = [self colorFromHexColorString:backgroundColor];
+  } else {
+    button.bezelColor = nil;
   }
 
   std::string label;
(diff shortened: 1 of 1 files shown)


shell/browser/ui/cocoa/atom_touch_bar.mm
```diff
@@ -368,16 +370,18 @@ - (void)updateButton:(NSCustomTouchBarItem*)item
...
-      button.imagePosition = NSImageLeft;
-    } else if (iconPosition == "right") {
-      button.imagePosition = NSImageRight;
-    } else {
-      button.imagePosition = NSImageOverlaps;
-    }
+  std::string iconPosition;
+  settings.Get("iconPosition", &iconPosition);
+  if (iconPosition == "left") {
+    button.imagePosition = NSImageLeft;
```
MarshallOfSound: Is it ok to set `imagePosition` when `image` is `nil`?

miniak: I've tried it, it's just ignored when `icon` is `nil`.

miniak:
@MarshallOfSound sorry for the noise with all the changes, I've pushed some changes by mistake and then fixed it later. It's too late...

shell/browser/ui/cocoa/atom_touch_bar.mm
```diff
@@ -499,9 +501,8 @@ - (void)updatePopover:(NSPopoverTouchBarItem*)item
   item.collapsedRepresentationLabel = base::SysUTF8ToNSString(label);
 
   gfx::Image image;
-  if (settings.Get("icon", &image)) {
-    item.collapsedRepresentationImage = image.AsNSImage();
-  }
+  settings.Get("icon", &image);
+  item.collapsedRepresentationImage = image.AsNSImage();
```
miniak:
```cpp
NSImage* Image::AsNSImage() const {
  return IsEmpty() ? nil : ToNSImage();
}
```

shell/browser/ui/cocoa/atom_touch_bar.mm
```diff
@@ -357,27 +357,29 @@ - (void)updateButton:(NSCustomTouchBarItem*)item
...
-    settings.Get("iconPosition", &iconPosition);
-    if (iconPosition == "left") {
-      button.imagePosition = NSImageLeft;
-    } else if (iconPosition == "right") {
-      button.imagePosition = NSImageRight;
-    } else {
-      button.imagePosition = NSImageOverlaps;
-    }
+  settings.Get("icon", &image);
+  button.image = image.AsNSImage();
```
miniak:
```cpp
NSImage* Image::AsNSImage() const {
  return IsEmpty() ? nil : ToNSImage();
}
```

shell/browser/ui/cocoa/atom_touch_bar.mm
```diff
@@ -357,27 +357,29 @@ - (void)updateButton:(NSCustomTouchBarItem*)item
...
+
+  std::string iconPosition;
+  settings.Get("iconPosition", &iconPosition);
```
miniak: setting `imagePosition` when `icon` is `nil` has no effect

miniak:
@MarshallOfSound replaced by new PR #19437 to make it easier to review as I made a complete mess here.

miniak:
@MarshallOfSound I will create another PR with more changes, which might be technically breaking (different `null` vs `undefined` vs invalid value handling), which won't be backported.

Issue #19425, opened 2019-07-24
process.uptime() returns unexpected result


MareikeTaeubner:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  v5.0.8
* **Operating System:**
 Microsoft Windows 10 Pro (Version: 10.0.18362 Build 18362)
* **Last Known Working Electron version:**
  unknown

### Expected Behavior
The same result node.js produces: the time the process is already running.
Output Node.js 12.7.0
```
node foo.js
2.1028718
4.1022395
6.1052112
8.1049451
```
### Actual Behavior
Electron returns some other numbers. 
Output Electron 5.0.8
```
electron foo.js 

 585175.0756366
585177.0780253
585179.0767907
585181.0769859
```

### To Reproduce
foo.js
```
function uptime() {
  console.log(process.uptime());
}

setTimeout(uptime, 2000);
setTimeout(uptime, 4000);
setTimeout(uptime, 6000);
setTimeout(uptime, 8000);
```

DeerMichel:
Broke between v5-beta.8 and v5-beta.9. Most likely the node upgrade https://github.com/electron/electron/pull/17752 - investigating further there...

PR #19436, opened 2019-07-24
fix: make process.uptime() return the correct time

DeerMichel:
#### Description of Change
Fixes https://github.com/electron/electron/issues/19425.

The issue was rooted in an upstream change: https://github.com/nodejs/node/pull/26016 (found its way to electron in https://github.com/electron/electron/pull/17752). Previously, the process-relative uptime was set in a method which got called by both `node::Start` (used by nodejs) and `node::Init` (used by embedders, _deprecated_) init methods. After the change, it is set directly in `node::Start` and thereby not when using `node::Init`. This PR just reverts that.

Since i don't see any difference in either way of setting the variable, we might wanna upstream this patch so other embedders can benefit from it as well.

cc @MarshallOfSound @codebytere 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed `process.uptime()` returning the wrong time.


nornagon:
@DeerMichel before we merge this, do you want to try submitting a PR to nodejs/node to see if anyone from the node maintainer group has thoughts about it?

DeerMichel:
Yep, sounds good to me :)

DeerMichel:
@nornagon my node PR got merged 🎉. So can we keep the patch here until we reached the current upstream version?

codebytere:
@DeerMichel needs rebase ☝️

DeerMichel:
done ✅

patches/node/fix_set_uptime_offset_in_correct_init_method.patch
```diff
@@ -0,0 +1,28 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Micha Hanselmann <deermichel@github.com>
+Date: Wed, 24 Jul 2019 11:44:25 -0700
+Subject: fix: set uptime offset in correct init method
```
MarshallOfSound: Can we add a note to this patch explaining when it can be removed and linking to the upstream nodejs PR.  At some point this patch will disappear and that context will help 👍

codebytere:
@DeerMichel could you please open manual backports for this to 5/6?

DeerMichel:
oh missed this, sure...

PR #19440, opened 2019-07-24
test: move autoUpdater specs to main process

nornagon:
#### Description of Change
- removed the test for using autoUpdater over remote, since that test should be covered by the remote spec
- clarified tests for `serverType`

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


spec-main/api-auto-updater-spec.ts
```diff
@@ -0,0 +1,92 @@
+import { autoUpdater } from 'electron'
+import { expect } from 'chai'
+import { ifit, ifdescribe } from './spec-helpers'
+
+ifdescribe(!process.mas)('autoUpdater module', function () {
+  describe('checkForUpdates', function () {
+    ifit(process.platform === 'win32')('emits an error on Windows when called the feed URL is not set', function (done) {
```
codebytere:
```suggestion
    ifit(process.platform === 'win32')('emits an error on Windows if the feed URL is not set', function (done) {
```

spec-main/api-auto-updater-spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    ifit(process.platform === 'win32')('correctly fetches the previously set FeedURL', function (done) {
+      const updateURL = 'https://fake-update.electron.io'
+      autoUpdater.setFeedURL({url: updateURL})
+      expect(autoUpdater.getFeedURL()).to.equal(updateURL)
+      done()
+    })
+  })
+
+  describe('setFeedURL', function () {
+    ifdescribe(process.platform === 'win32' || process.platform === 'darwin')('on Mac or Windows', () => {
```
codebytere:
```suggestion
    ifdescribe(process.platform !== 'linux')('on Mac or Windows', () => {
```

spec-main/api-auto-updater-spec.ts
```diff
@@ -0,0 +1,92 @@
...
+        autoUpdater.once('error', function (error) {
+          expect(error.message).equal('Could not get code signature for running application')
+          done()
+        })
+        autoUpdater.setFeedURL({url:''})
+      })
+
+      it('does not throw if default is the serverType', () => {
+        // "Could not get code signature..." means the function got far enough to validate that serverType was OK.
+        expect(() => autoUpdater.setFeedURL({ url: '', serverType: 'default' })).to.throw('Could not get code signature for running application')
```
codebytere:
```suggestion
        expect(() => autoUpdater.setFeedURL({ url: '', serverType: 'default' })).to.not.throw('Could not get code signature for running application')
```

The title and old test align in expectation that we _aren't_ throwing but this changes semantics so we're expecting a thrown error; is this intentional? if so, we should update the spec description accordingly.

spec-main/api-auto-updater-spec.ts
```diff
@@ -0,0 +1,92 @@
...
+      })
+
+      it('does not throw if json is the serverType', () => {
+        // "Could not get code signature..." means the function got far enough to validate that serverType was OK.
+        expect(() => autoUpdater.setFeedURL({ url: '', serverType: 'json' })).to.throw('Could not get code signature for running application')
```
codebytere: same as above

spec-main/api-auto-updater-spec.ts
nornagon: Yes, previously the test was testing that we didn't throw a certain error message ("Expected serverType to be..."). However, I think that's not a very good test because:
- it doesn't test that the code even gets to the place where `serverType` is tested (e.g. it could throw `Couldn't read property setFeedURL of undefined` and we'd call that good)
- it's flaky with respect to the specific error message, in that if we changed the error message to be something like "Unexpected value for serverType..." instead of "Expected serverType to be...", and then started throwing that error, this test wouldn't catch it

The new behaviour tests that it at least gets to _trying_ to set the feed url. It's not the best way of doing it (i'd rather have a success case than assume that this error message means we're OK for serverType), but it's the best way short of figuring out how to sign the test app which i don't want to do.

spec-main/api-auto-updater-spec.ts
nornagon: i prefer it to explicitly match the name of the test :)

spec-main/api-auto-updater-spec.ts
codebytere: cool, ty for the context!

PR #19449, opened 2019-07-25
fix: activate the uv_loop on incoming IPC messages

### LINKED ISSUE #19368: Sending IPC messages to renderers interferes with node http module on windows
### Preflight Checklist

### Issue Details

* **Electron Version:**
  *  6.0.0-beta.15
* **Operating System:**
  * Windows 10 1903
  * 5.0.7

### Expected Behavior
Node's built-in should work all the time the app starts.

### Actual Behavior
I noticed when using NeDB in my Electron app in renderer process after the update to 6.0.0-beta.15, the databases just don't load (they load properly very rare). But when reloading my renderer process using `Ctrl + R` everything works fine. Also it's worth to mention that requests using `http` or `https` modules behave in the same way, they just sometimes don't work. The problem doesn't occur in Electron 5.0.7. The problem seems to be very random and for me it's hard to describe what causes it.

### To Reproduce
The best way to reproduce this is to build my [project](https://github.com/wexond/wexond). All instructions are in the README. Then try to add some bookmarks, close the app and restart it and see if they appear under the address bar. Also as I mentioned before, sometimes web requests doesn't work so you may not see the weather card. Then try to reload using `Ctrl + R` and you will see that everything works correctly.

MarshallOfSound:
Fixes #19368

I don't *completely* understand what is happening here but this is what I was seeing.

* `http` called `Connect` on `TCPWrap`, this queued up a callback with an async ID
* Our message came in and with async ID 0
* The callback for TCPWrap got called but never got entered (it was stuck as a pending tick)

For whatever reason in Electron 6 and higher our IPC calls into the renderer in a way that prevents other pending ticks being called.  This fixes it and I'd like to ship it in `6.0.0` next week as this issue is unpredictable and incredibly hard to trace.

I believe this is related to the mojo-refactor but even removing the `NodeCallbackScope` from our ipc server did not fix this issue, with limited time on hand I just threw this fix up 👍 

Notes: Fixed issue where sometimes asynchronous node.js methods such as `http`, `setImmediate` and `fs` wouldn't call their callbacks for an incredibly long amount of time.

### DIFF
--- a/lib/renderer/init.ts
+++ b/lib/renderer/init.ts
@@ -48,6 +48,7 @@ v8Util.setHiddenValue(global, 'ipcNative', {
   onMessage (internal: boolean, channel: string, args: any[], senderId: number) {
     const sender = internal ? ipcInternalEmitter : ipcEmitter
     sender.emit(channel, { sender, senderId }, ...args)
+    process.activateUvLoop()
   }
 })
 


MarshallOfSound:
> Can we make a tracking issue for it and set it as blocking for v7?

Done https://github.com/electron/electron/issues/19466

PR #19458, opened 2019-07-25
docs: fix missing possibleValues

ckerr:
#### Description of Change

Inspired by https://github.com/electron/electron/pull/19448. Tweaks the docs so that proper `possibleValues` will be generated for:

 * contents.insertCSS()'s `options` parameter's `cssOrigin` property
 * image.resize()'s `options` parameter's `quality` property
 * new ClientRequest()'s `options` parameter's `protocol` property
 * new ClientRequest()'s `options` parameter's `redirect` property

CC @MarshallOfSound @codebytere 

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/docs/api/structures/client-request-constructor-options.md
+++ b/docs/api/structures/client-request-constructor-options.md
@@ -4,7 +4,7 @@
 method.
 * `url` String (optional) - The request URL. Must be provided in the absolute
 form with the protocol scheme specified as http or https.
-* `session` Object (optional) - The [`Session`](session.md) instance with
+* `session` Session (optional) - The [`Session`](session.md) instance with
 which the request is associated.
 * `partition` String (optional) - The name of the [`partition`](session.md)
   with which the request is associated. Defaults to the empty string. The
--- a/docs/api/structures/client-request-constructor-options.md
+++ b/docs/api/structures/client-request-constructor-options.md
@@ -1,0 +1,42 @@
+# ClientRequestConstructorOptions Object
+
+* `method` String (optional) - The HTTP request method. Defaults to the GET
+method.
+* `url` String (optional) - The request URL. Must be provided in the absolute
+form with the protocol scheme specified as http or https.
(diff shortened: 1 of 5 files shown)


codebytere:
@ckerr:
```
new ClientRequest()'s options parameter's protocol property
new ClientRequest()'s options parameter's redirect property
```
still aren't showing up in the ts defs file it seems!

filenames.auto.gni
```diff
@@ -1,6 +1,7 @@
 # THIS FILE IS AUTO-GENERATED, PLEASE DO NOT EDIT BY HAND
 auto_filenames = {
   api_docs = [
+    "docs/api/.menu-item.md.swp",
```
codebytere: 🤔 not sure about this extension, was it added on purpose?

docs/api/structures/client-request-constructor-options.md
```diff
@@ -0,0 +1,42 @@
+# ClientRequestConstructorOptions Object
```
MarshallOfSound: We probably don't want to do this? This constructor object should be automatically generated by the docs parser, if it's not then that's a bug in the docs parser we should fix

ckerr: Agreed. I think this is resolved in docs-parser 0.4.2 in that when there's a union (here, an object | string) followed by a key list, docs-parser now only applies the list to the first union type.

ckerr:
Some of this PR overlaps with the changes in https://github.com/electron/electron/pull/19585, and since both PRs rely on docs-parser 0.4.2, this should be updated _after_ 19585 lands

codebytere:
@ckerr the dependent PR is merged, so this should be good to rebase and merge 🚀

zcbenz:
@ckerr Can you rebase this branch?

zcbenz:
Closing since there has been no activity for a long time.

PR #19460, opened 2019-07-25
refactor: move `gfx::ImageSkia` functions to their own util file

erickzhao:
#### Description of Change

Inside `atom_api_native_image.cc`, we have a set of functions that are useful for generating `gfx::ImageSkia` representations.

I wanted to use those outside the context of `NativeWindow` in https://github.com/electron/electron/pull/19420, where I am reusing `electron::ShowMessageBoxSync` function to pop up an About dialog for Windows.

For this About Dialog, I am grabbing the message box icon from a file path, like we do in Linux.

Since `gfx::ImageSkia` is used outside of `NativeImage`, I was thinking that it'd be useful to do extract these utility functions into their own shared file instead of keeping them private within the NativeImage API file.

Would like feedback on the usefulness/viability of this!

cc @codebytere @DeerMichel 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/shell/common/skia_util.cc
+++ b/shell/common/skia_util.cc
@@ -18,10 +18,13 @@
 #include "ui/gfx/codec/jpeg_codec.h"
 #include "ui/gfx/codec/png_codec.h"
 #include "ui/gfx/geometry/size.h"
-#include "ui/gfx/icon_util.h"
 #include "ui/gfx/image/image_skia.h"
 #include "ui/gfx/image/image_skia_operations.h"
 #include "ui/gfx/image/image_util.h"
+
+#if defined(OS_WIN)
+#include "ui/gfx/icon_util.h"
+#endif
 
 namespace electron {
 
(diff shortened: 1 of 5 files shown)


shell/common/skia_util.cc
```diff
@@ -155,6 +147,17 @@ bool PopulateImageSkiaRepsFromPath(gfx::ImageSkia* image,
...
+    return false;
+
+  image->AddRepresentation(gfx::ImageSkiaRep(bitmap, 1.0f));
+  return true;
+#endif
+
 }  // namespace util
 
-}  // namespace electron
+}  // namespace util
```
codebytere: i don't think this should have been changed?

erickzhao: Has already been fixed in a later commit (ab9e458)! Problem was I forgot to paste the closing bracket in the Windows file, and the linter auto-changed these names because the brackets were wonky. :)

PR #19462, opened 2019-07-25
feat: emit always-on-top-changed on macOS

codebytere:
#### Description of Change

The `always-on-top-changed` event  for BrowserWindows showed in documentation as being emitted on `macOS` only, but in fact it was emitted previously on `windows` and `linux` only. This corrects that error so that `always-on-top-changed` is emitted on all platforms.

cc @ckerr @MarshallOfSound @miniak 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `always-on-top-changed` event emission for macOS.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -553,7 +553,7 @@ Emitted when the window enters a full-screen state triggered by HTML API.
 
 Emitted when the window leaves a full-screen state triggered by HTML API.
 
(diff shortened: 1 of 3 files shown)


codebytere:
NB: when https://github.com/electron/electron/pull/19463 is merged, the test added therein should remove its platform scoping so as to ensure correct behavior on all platforms.

Issue #19453, opened 2019-07-25
BrowserWindow event always-on-top-changed called with negated bool parameter (isAlwaysOnTop)


cmeeren:
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** 6.0.0-beta.13
* **Operating System:** Windows 10

### Expected Behavior

When calling `browserWindow.setAlwaysOnTop(true)`, I expect the `always-on-top-changed` event to be called with the boolean `isAlwaysOnTop` parameter set to `true`, and vice versa.

### Actual Behavior

I experience the opposite of above: When calling `browserWindow.setAlwaysOnTop(true)`, my `always-on-top-changed` handler is called with `false`, and vice versa.

### To Reproduce

I don't have any JS code to share and I'm not a JS dev (I use F# and compile to JS with Fable). I think the issue should be easy to reproduce though.

PR #19463, opened 2019-07-25
fix: correctly emit BrowserWindow alwaysOnTop status in change event

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/19453.

Previously, `alwaysOnTop` status was being calculated dynamically and emitted at the top of `NativeWindowViews::SetAlwaysOnTop`, which meant that it had not yet been changed at the operating system level. Consequently, when the event itself called `IsAlwaysOnTop()` it would return the inverse of what it should be.

To fix this, I determined whether the level has been changed at the top of `SetAlwaysOnTop`, and then moved the actual event emission itself to the bottom of the method so that it would pick up the correct new `alwaysOnTop` status. Tests have been added to verify this behavior.

cc @ckerr @MarshallOfSound @DeerMichel  

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed inverted `alwaysOnTop` status returned when `always-on-top-changed` was emitted on Linux and Windows.

### DIFF
--- a/shell/browser/native_window_views.cc
+++ b/shell/browser/native_window_views.cc
@@ -756,9 +756,7 @@ void NativeWindowViews::SetAlwaysOnTop(ui::ZOrderLevel z_order,
                                        const std::string& level,
                                        int relativeLevel,
                                        std::string* error) {
-  if (z_order != widget()->GetZOrderLevel())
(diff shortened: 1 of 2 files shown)


DeerMichel:
Confirmed ✅

Issue #19455, opened 2019-07-25
browser-window-created and web-contents-created called with empty object as first parameter instead of event


cmeeren:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** 6.0.0-beta.13
* **Operating System:** Windows 12

### Expected Behavior

The documentation for the App events `browser-window-created` and `web-contents-created` state that the first parameter is an `Event`. I therefore expect the first parameter to be an `Event`.

### Actual Behavior

The first parameter for both of these events is `undefined`. (The second parameter is a `BrowserWindow` and `WebContents`, respectively, as documented.)

PR #19465, opened 2019-07-25
fix: [wc|bw]-created events do not emit an event object

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/19455.

Previously, an [empty object](https://github.com/electron/electron/blob/d1292833e924fc80043a937a85fcb8f49cf67c30/lib/browser/api/browser-window.js#L85) was emitted in place of an Event during `browser-window-created` as well as when `web-contents-created` was [emitted](https://github.com/electron/electron/blob/8782d06ed610e346f70862561aa033905c93a86c/lib/browser/api/web-contents.js#L416).

This led to false assumptions about the nature of the object for end-users. Instead, we now create an empty but legitimate Event and emit that instead.

cc @zcbenz @deepak1556 @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: `web-contents-created` and `browser-window-created` no longer emit an empty object in place of Event.

### DIFF
--- a/lib/browser/api/browser-window.js
+++ b/lib/browser/api/browser-window.js
@@ -82,7 +82,7 @@ BrowserWindow.prototype._init = function () {
   }
 
   // Notify the creation of the window.
-  const event = process.electronBinding('event').createWithSender(null)
+  const event = process.electronBinding('event').createEmpty()
   app.emit('browser-window-created', event, this)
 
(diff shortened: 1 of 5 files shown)


codebytere:
i'm looking into those failures ☝️

PR #19469, opened 2019-07-25
fix: use GetAuraColor to get theme dependant menu bg color

MarshallOfSound:
Easiest to explain with pictures

Before:
image

After:
image

Notes: Fixed window menu background color when in dark mode on Windows 10


MarshallOfSound:
Unclear, the underline is correctly working atm and we depend on that header transitively.  Just wanted to fix the giant white blob of a menu 😆

PR #19479, opened 2019-07-26
docs: fix type information for contents.sendInputEvent

MarshallOfSound:
The `event` type and keyword is reserved by the typescript generator, in the future we should throw an error from the generator if someone uses a reserved interface like that.

Fixes https://github.com/electron/typescript-definitions/issues/36

Notes: no-notes


MarshallOfSound:
Oh this type is slightly more complicated, gonna need to fix this locally 😆 too complex for github ui editor

MarshallOfSound:
The `extends` feature is new on `master` so this can't be backported

codebytere:
Confirmed [here](https://github.com/electron/electron/pull/19479/checks?check_run_id=178560143)

PR #19482, opened 2019-07-26
refactor: migrate dialog API to //gin

DeerMichel:
#### Description of Change
Get something rollin' 🚀.

This time, my pilot migration from `native_mate` to Chromium's `gin` succeeded - the dialog API works as before. There's not a lot magic happening since it's just binding static methods instead of fiddling around with objects & lifetimes.

My approach doesn't need any patches on `gin` itself, although i am a bit afraid that how i [convert functions](https://github.com/electron/electron/blob/b26290d5bc11e04b57e165040b6895733898c878/shell/common/gin_converters/callback_converter.h#L18) might be prone to memory leaks ([ref](https://github.com/electron/electron/blob/b26290d5bc11e04b57e165040b6895733898c878/shell/common/native_mate_converters/callback.h#L149)). 

Adapters are used for converters that are also used by `native_mate`-driven-APIs. Eventually, i want to flip it around that we have only `gin_converters` and adapt for `native_mate` as needed - but for now, this seems safer 😇.

I appreciate any feedback and further clarification on this (and why we did x/y in `native_mate` instead of using a simpler approach) as i want to dive deeper and get objects working. Feel free to reach out to me via dm 💯.

cc @zcbenz @deepak1556 @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


shell/common/gin_converters/callback_converter.h
```diff
@@ -0,0 +1,29 @@
...
+// Use of this source code is governed by the MIT license that can be
+// found in the LICENSE file.
+
+#ifndef SHELL_COMMON_GIN_CONVERTERS_CALLBACK_CONVERTER_H_
+#define SHELL_COMMON_GIN_CONVERTERS_CALLBACK_CONVERTER_H_
+
+#include "gin/converter.h"
+#include "gin/function_template.h"
+
+// TODO(deermichel): check if this is a good idea?!
```
zcbenz: I think we should either add a utility function with name like `ConvertCallbackToV8Leaked`, or use the non-leak version in native-mate.

DeerMichel: A function named `*Leaked` doesn't make me feel very good 😄. So for clarification, `CreateFunctionTemplate` caches each template and therefore is leaky - that makes it unsuitable for a converter where things might happen without the user noticing.

However, using it here (as part of a utility function) shouldn't hurt anybody since `Initialize` is only called once per Electron instance, right? 

For objects, this is indeed problematic because the leaky method would get called for every object instantiation. I have an idea for a solution which i will push in another PR if that's ok 😬.

zcbenz: It is fine in this case, but not when people start using it for arbitrary callbacks. We must explicitly warn them that this functions leaks.

Having `Leak` in names is fine, Chromium has some commonly used utilities with this style, for example `LeakyLazyInstanceTraits` and `LeakySingletonTraits`, and I think it helps write correct code.

DeerMichel:
Done. Refactored:
- add leaky util function
- remove unused converter adapters

shell/common/gin_converters/file_dialog_converter_gin_adapter.h
```diff
@@ -0,0 +1,33 @@
...
+// -- this adapter forwards all conversions to the existing mate converter --
+// (other direction might be preferred, but this is safer for now :D)
+
+namespace gin {
+
+template <>
+struct Converter<file_dialog::DialogSettings> {
+  static v8::Local<v8::Value> ToV8(v8::Isolate* isolate,
+                                   const file_dialog::DialogSettings& in) {
+    return mate::ConvertToV8(isolate, in);
```
MarshallOfSound: I realize this has been merged now but can a follow up PR be raised to reverse this logic.  The conversion logic should be moved to the gin converter and the mate converter should just call `gin::ConvertToV8`.

The reasoning behind this is that eventually when we have converted everything to gin we will only have to remove a bunch of placeholder converters and not have to do the risky "copy paste" for every converter at the same time, we should be moving the converters slowly in PRs like this.

cc @deepak1556

DeerMichel: Yes i'm actually surprised that this got merged so fast. My plan is to move all the converters to gin (have placeholders for mate) in some PR's.. adding this one to my backlog as well 👍

PR #19490, opened 2019-07-27
Notifications. Error passing values to notifications object.

Josileno:
This way it didn't work for me:

    let myNotification = new Notification ('Title', {
        body: 'Lorem Ipsum Dolor Sit Amet'
    })

It only worked when I decided to try changing the code to:

    let myNotification = new Notification ({
        title: 'Title',
        body: 'Lorem Ipsum Dolor Sit Amet'
    })

I hope this helps someone.


MarshallOfSound:
This document (as mentioned a few lines up) is for renderer process notifications, the api you are changing it to is for the main process.

The docs are currently correct

PR #19512, opened 2019-07-29
chore: disable JS impl when //extensions support is on

nornagon:
#### Description of Change
The code in `lib/browser/chrome-extension.js` conflicts with the `//extensions` code, so disable it when the latter is enabled.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/lib/browser/chrome-extension.js
+++ b/lib/browser/chrome-extension.js
@@ -1,533 +1,540 @@
 'use strict'
 
-const { app, webContents, BrowserWindow } = require('electron')
-const { getAllWebContents } = process.electronBinding('web_contents')
-const ipcMainUtils = require('@electron/internal/browser/ipc-main-internal-utils')
+if (process.electronBinding('features').isExtensionsEnabled()) {
+  exports.getContentScripts = () => {
+    // When //extensions support is enabled, this function shouldn't be called.
+    return []
+  }
+} else {
+  const { app, webContents, BrowserWindow } = require('electron')
+  const { getAllWebContents } = process.electronBinding('web_contents')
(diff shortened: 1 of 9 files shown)


lib/browser/chrome-extension.js
```diff
@@ -1,533 +1,540 @@
...
-
-// Mapping between extensionId(hostname) and manifest.
-const manifestMap = {} // extensionId => manifest
-const manifestNameMap = {} // name => manifest
-const devToolsExtensionNames = new Set()
-
-const generateExtensionIdFromName = function (name) {
-  return name.replace(/[\W_]+/g, '-').toLowerCase()
-}
+if (process.electronBinding('features').isExtensionsEnabled()) {
```
MarshallOfSound: Can't we just not execute this file at all if extension support is enabled instead of wrapping the whole thing in an `if`?

nornagon:
Recommend reviewing with [`?w=1`](https://github.com/electron/electron/pull/19512/files?w=1)

lib/browser/chrome-extension.js
nornagon: Hm... I was avoiding refactoring the things that call this but I suppose that's doable.

lib/browser/init.ts
```diff
@@ -152,7 +152,9 @@ app._setDefaultAppPaths(packagePath)
 require('@electron/internal/browser/devtools')
 
 // Load the chrome extension support.
-require('@electron/internal/browser/chrome-extension')
+if (!process.electronBinding('features').isExtensionsEnabled()) {
```
MarshallOfSound: We should add similar checks around require calls to `chrome-api` in the renderer (the thing that binds `global.chrome`)

PR #18306, opened 2019-05-15
feat: implement win.setAspectRatio() on Windows 

MadfishDT:
#### Description of Change
Description of Change
Implemented win.setAspectRatio() on Windows by handling the WM_SIZING message.

#### Checklist

- [ ] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x ] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

Release Notes
Notes: Implemented win.setAspectRatio() on Windows.

Issue #8036, opened 2016-11-21
Enforce a aspect ratio when resizing for Windows


szymonc:
As in topic. API currently offers fixed ratio only for Mac but not for Windows

Cleod9:
Possible workaround/implementation? (16:9 aspect ratio example)
```js
// 'win' as the BrowserWindow instance
win.on('resize', function () {
  setTimeout(function () {
    var size = win.getSize();
    win.setSize(size[0], parseInt(size[0] * 9 / 16));
  }, 0);
});
```

szymonc:
@Cleod9 did you try this one? We have tried something similar but it causes blinking of window and strange behaviour as it interferes with native resize events. It worked well but just for one edge

Cleod9:
@szymonc Yes, I did try it and it works for my purposes on Windows 7. Doesn't look pretty while resizing but not a deal-breaker for me. Also I had to account for the top/bottom windows borders using `(size[0] + 60) * 9 / 16` (which may not translate to all versions of windows, need to confirm)

darkguy2008:
Almost 1 year and I can't believe this issue hasn't been fixed yet for Windows.

darkguy2008:
After 1 year of nobody watching this... here's my fix. It works and it doesn't flicker, somebody please PR this :)

```

**Comment by bluntspoon** [comment]
Would be nice to see win.setAspectRatio added for Windows? Any plans for this?

**Comment by brett84c** [comment]
```
var resizeTimeout;
mainWindow.on('resize', (e)=>{
    clearTimeout(resizeTimeout);
    resizeTimeout = setTimeout(function(){
        var size = mainWindow.getSize();
        mainWindow.setSize(size[0], parseInt(size[0] * 9 / 16));
    }, 100);
});
```

Jacobboogiebear:
Based upon the upper answers, I made a npm package for this reason!
https://www.npmjs.com/package/electron-aspectratio

HaulinOats:
@Jacobboogiebear Cool, I'll give this a try in the next couple days and report back!

HaulinOats:
@Jacobboogiebear Do you know if there's a way to prevent 'resize' from triggering when just moving the window? I don't want to stop the ability to move the window, but for some reason, Electron triggers a resize on move. It's bizarre.

Jacobboogiebear:
@HaulinOats I don't know how to deal with that but I will put some time in to study it and report back

HaulinOats:
[...]

PR #19516, opened 2019-07-29
feat: implement `win.setAspectRatio()` on Linux

erickzhao:
#### Description of Change

Ref #8036 

Implements `win.setAspectRatio()` on Linux using Chromium widget's [setAspectRatio](https://cs.chromium.org/chromium/src/ui/views/widget/widget.h?dr=CSs&g=0&l=584) function. Works on Ubuntu 19.04.

Although the Chromium API is available on both Windows and Linux, a separate PR (#18306) is already opened with a Windows-specific approach. Moreover, this approach doesn't seem to be working when tested locally, both for me and @CapOM (see https://github.com/electron/electron/pull/18306#issuecomment-511059726).

Might want to hold off on merging this until the #18306 is merged, as there will be conflicts.

#### Limitations to this approach
* `extraSize` option not available
* Cannot unset the forced aspect ratio

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Implemented `win.setAspectRatio()` on Linux.


shell/browser/native_window_views.cc
```diff
@@ -668,6 +668,22 @@ bool NativeWindowViews::IsResizable() {
...
+#if defined(OS_LINUX)
+  NativeWindow::SetAspectRatio(aspect_ratio, extra_size);
+  gfx::SizeF aspect(aspect_ratio, 1.0);
+  // widget SetAspectRatio() won't work with numbers too small
+  aspect.Scale(100);
+
+  if (aspect_ratio > 0) {
+    widget()->SetAspectRatio(aspect);
+  } else {
+    // reset aspect ratio somehow
```
ckerr: There doesn't seem to be any explicit way of doing this in the existing code, at least for Linux. It looks like the low-level implementation is at `DesktopWindowTreeHostX11::SetAspectRatio` where the X11 `PAspect` flag gets turned on for`XSizeHints` ... and there's no other place where `PAspect` is used explicitly again.

If we wanted to add support for unsetting, I think we'd have to add it ourselves. For example in the Linux code we'd need something like this:

```diff
--- a/ui/views/widget/desktop_aura/desktop_window_tree_host_x11.cc
+++ b/ui/views/widget/desktop_aura/desktop_window_tree_host_x11.cc
@@ -1129,9 +1129,13 @@ void DesktopWindowTreeHostX11::SetAspectRatio(const gfx::SizeF& aspect_ratio) {
   long supplied_return;
 
   XGetWMNormalHints(xdisplay_, xwindow_, &size_hints, &supplied_return);
-  size_hints.flags |= PAspect;
-  size_hints.min_aspect.x = size_hints.max_aspect.x = aspect_ratio.width();
-  size_hints.min_aspect.y = size_hints.max_aspect.y = aspect_ratio.height();
+  if (aspect_ratio.IsEmpty()) {
+    size_hints.flags &= ~PAspect;
+  } else {
+    size_hints.flags |= PAspect;
+    size_hints.min_aspect.x = size_hints.max_aspect.x = aspect_ratio.width();
+    size_hints.min_aspect.y = size_hints.max_aspect.y = aspect_ratio.height();
+  }
   XSetWMNormalHints(xdisplay_, xwindow_, &size_hints);
 }
```

Maybe upstream would want a patch like this, but I'm not sure -- they probably don't have a use case for it?

shell/browser/native_window_views.cc
```diff
@@ -668,6 +668,22 @@ bool NativeWindowViews::IsResizable() {
   return CanResize();
 }
 
+void NativeWindowViews::SetAspectRatio(double aspect_ratio,
+                                       const gfx::Size& extra_size) {
+#if defined(OS_LINUX)
+  NativeWindow::SetAspectRatio(aspect_ratio, extra_size);
+  gfx::SizeF aspect(aspect_ratio, 1.0);
+  // widget SetAspectRatio() won't work with numbers too small
+  aspect.Scale(100);
```
ckerr: :+1: on calling `Scale()` here but the comment doesn't explain why it's necessary, only that it is necessary.

Maybe something more like
`Must scale aspect higher because setAspectRatio() truncates aspect.width,height to ints`

shell/browser/native_window_views.cc
```diff
@@ -668,6 +668,22 @@ bool NativeWindowViews::IsResizable() {
   return CanResize();
 }
 
+void NativeWindowViews::SetAspectRatio(double aspect_ratio,
+                                       const gfx::Size& extra_size) {
+#if defined(OS_LINUX)
+  NativeWindow::SetAspectRatio(aspect_ratio, extra_size);
```
ckerr: The superclass line probably should go above the ifdef

shell/browser/native_window_views.cc
erickzhao: I found the relevant bit of code in `x11_window.cc`, but implemented the same logic!

patches/chromium/feat_implement_unset_window_aspect_ratio_on_linux.patch
```diff
@@ -0,0 +1,49 @@
...
+Subject: feat: implement unset window aspect ratio on linux
+
+
+diff --git a/electron b/electron
+new file mode 160000
+index 0000000000000000000000000000000000000000..6ae31bb94fdb0675fb134aa42e15f66b42cdc6e8
+--- /dev/null
++++ b/electron
+@@ -0,0 +1 @@
++Subproject commit 6ae31bb94fdb0675fb134aa42e15f66b42cdc6e8
```
MarshallOfSound: Wut?

patches/chromium/feat_implement_unset_window_aspect_ratio_on_linux.patch
```diff
@@ -0,0 +1,49 @@
...
+index 0000000000000000000000000000000000000000..6d81b828df977ccfa8cab5501f0e96d883cda1f7
+--- /dev/null
++++ b/third_party/electron_node
+@@ -0,0 +1 @@
++Subproject commit 6d81b828df977ccfa8cab5501f0e96d883cda1f7
+diff --git a/third_party/nan b/third_party/nan
+new file mode 160000
+index 0000000000000000000000000000000000000000..2ee313aaca52e2b478965ac50eb5082520380d1b
+--- /dev/null
++++ b/third_party/nan
```
MarshallOfSound: wut?

patches/chromium/feat_implement_unset_window_aspect_ratio_on_linux.patch
```diff
@@ -0,0 +1,49 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
MarshallOfSound: The default should be to try upstream this patch, before we land this patch into `electron` can you submit upstream and see if it lands?

docs/api/browser-window.md
```diff
@@ -928,12 +928,12 @@ Returns `Boolean` - Whether the window is in simple (pre-Lion) fullscreen mode.
 
 Returns `Boolean` - Whether the window is in normal state (not maximized, not minimized, not in fullscreen mode).
 
-#### `win.setAspectRatio(aspectRatio[, extraSize])` _macOS_
+#### `win.setAspectRatio(aspectRatio[, extraSize])` _macOS_ _Linux_
 
 * `aspectRatio` Float - The aspect ratio to maintain for some portion of the
 content view.
 * `extraSize` [Size](structures/size.md) (optional) - The extra size not to be included while
```
MarshallOfSound: The `_macOS_` should be on this line

```suggestion
* `extraSize` [Size](structures/size.md) (optional) _macOS_ - The extra size not to be included while
```

erickzhao:
@ckerr will try to get the upstream ticket up before end of week!

erickzhao:
@MarshallOfSound fixed the conflict in the `.patches` file by adding the incoming patch from master and removing one that was removed on master (due to being upstreamed).

also regarding this patch's upstream: will need to get my local setup ready, but might take a while because internet back home is real slow. 😬

codebytere:
@erickzhao can you rebase this?

erickzhao:
@codebytere Don't have write access to the repo anymore so I can't 😱 

Can't seem to change the branch that this PR was from, either.

codebytere:
ah ok i can do it then :)

jaehayoo:
@codebytere  When will it be available on really important windows?

PR #19518, opened 2019-07-29
test: fix some failing and disabled node specs

MarshallOfSound:
These tests were failing due to mismatched stacktraces as a result of our Module._load hook.  This fixes those by adding a flag to optionally not hook those calls and instead do the asar override eagily.

No user facing code was modified to enable these tests, just re-positioned

Notes: no-notes


lib/common/asar.js
```diff
@@ -726,31 +726,42 @@
...
+      const { exec, execSync } = childProcess
+      childProcess.exec = invokeWithNoAsar(exec)
+      childProcess.exec[util.promisify.custom] = invokeWithNoAsar(exec[util.promisify.custom])
+      childProcess.execSync = invokeWithNoAsar(execSync)
+
+      overrideAPI(childProcess, 'execFile')
+      overrideAPISync(childProcess, 'execFileSync')
+    }
+
     // Lazily override the child_process APIs only when child_process is fetched the first time
```
nornagon:
```suggestion
    // Lazily override the child_process APIs only when child_process is fetched the first time.
```

lib/common/asar.js
```diff
@@ -726,31 +726,42 @@
...
-          // Electron to consider the full command as a single path
-          // to an archive.
-          const { exec, execSync } = childProcess
-          childProcess.exec = invokeWithNoAsar(exec)
-          childProcess.exec[util.promisify.custom] = invokeWithNoAsar(exec[util.promisify.custom])
-          childProcess.execSync = invokeWithNoAsar(execSync)
-
-          overrideAPI(childProcess, 'execFile')
-          overrideAPISync(childProcess, 'execFileSync')
+    // we will eagily override the child_process APIs when this env var is set so that stack traces
```
nornagon:
```suggestion
    // We will eagerly override the child_process APIs when this env var is set so that stack traces
```

script/node-spec-runner.js
```diff
@@ -21,7 +21,8 @@ async function main () {
   const testChild = cp.spawn('python', ['tools/test.py', '--verbose', '-p', 'tap', '--logfile', TAP_FILE_NAME, '--mode=debug', 'default', `--skip-tests=${DISABLED_TESTS.join(',')}`, '--shell', utils.getAbsoluteElectronExec(), '-J'], {
     env: {
       ...process.env,
-      ELECTRON_RUN_AS_NODE: 'true'
+      ELECTRON_RUN_AS_NODE: 'true',
+      ELECTRON_EAGER_ASAR_HOOK: 'true'
```
nornagon: Maybe make it explicit that this is a test-only thing?

```suggestion
      ELECTRON_EAGER_ASAR_HOOK_FOR_TESTING: 'true'
```

Issue #19522, opened 2019-07-30
app.dock.bounce no longer has a default bounce type


nghiepit:
Hi all,

I just upgraded to `6.0` and `TypeError: Insufficient number of arguments.`

MarshallOfSound:
Looks like a refactor caused the `bounce` implementation to lose its old default value --> https://github.com/electron/electron/blob/4ccd6d59002e46cecee0237c4ef58323ad6b4017/lib/browser/api/app.js#L63-L65

We should provide a default value in the native `DockBounce` impl

PR #19523, opened 2019-07-30
fix: set default dock bounce type

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/19522.

Sets `informational` as the default dock bounce type value once more.

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed default dock bounce type on macOS.

### DIFF
--- a/shell/browser/api/atom_api_app.cc
+++ b/shell/browser/api/atom_api_app.cc
@@ -1338,8 +1338,11 @@ bool App::IsInApplicationsFolder() {
   return ui::cocoa::AtomBundleMover::IsCurrentAppInApplicationsFolder();
 }
 
-int DockBounce(const std::string& type) {
+int DockBounce(mate::Arguments* args) {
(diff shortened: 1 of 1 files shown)


lib/browser/api/app.ts
```diff
@@ -65,6 +65,8 @@ if (process.platform === 'darwin') {
     setDockMenu(menu)
   }
   app.dock!.getMenu = () => dockMenu
+  const dockBounce = app.dock!.bounce
+  app.dock!.bounce = (type = 'informational') => dockBounce(type)
```
MarshallOfSound: Can we move this to CPP instead, I want to move away from these JS side hacks as it makes things like context isolated electron harder in the future.  Using either a cpp default (not sure how that works with mate) or using `mate::Arguments` or using a `base::Optional<std::string>`

PR #19545, opened 2019-07-31
fix: throw better error on getPath('logs')

codebytere:
Backport of .https://github.com/electron/electron/pull/19514.

See that PR for more details.

Notes: `getPath('logs')` now throws better error when it fails to find the logs path.

### DIFF
--- a/atom/browser/api/atom_api_app.cc
+++ b/atom/browser/api/atom_api_app.cc
@@ -860,8 +860,15 @@ base::FilePath App::GetPath(mate::Arguments* args, const std::string& name) {
   int key = GetPathConstant(name);
   if (key >= 0)
     succeed = base::PathService::Get(key, &path);
-  if (!succeed)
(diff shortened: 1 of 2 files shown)


ckerr:
I'm of two minds about this:

1. apps that relied on `getPath('log')` to work before will now get an exception, which is not great API.

2. adding side-effects to an accessor, e.g. creating a directory as a side-effect of `getPath('log')`, is not great API.

Maybe we could keep the side-effect for 6 to keep apps from breaking in the short term, but deprecate the behavior and require users of 7 (or 8, if we're being strict on the deprecation protocol) to call `app.setAppLogsPath()` before using `app.getPath('log')`?

MarshallOfSound:
> adding side-effects to an accessor, e.g. creating a directory as a side-effect of getPath('log'), is not great API.

As an opposite take, an API returning a file path that doesn't exist isn't great either 🤷‍♂

PR #19547, opened 2019-07-31
feat: add tray.removeBalloon()

miniak:
#### Description of Change
Add missing method to remove a balloon notification, which is already displayed.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `tray.removeBalloon()`, which removes an already displayed balloon notification.

### DIFF
--- a/shell/browser/ui/win/notify_icon.cc
+++ b/shell/browser/ui/win/notify_icon.cc
@@ -138,6 +138,16 @@ void NotifyIcon::DisplayBalloon(HICON icon,
     LOG(WARNING) << "Unable to create status tray balloon.";
 }
 
+void NotifyIcon::RemoveBalloon() {
+  NOTIFYICONDATA icon_data;
+  InitIconData(&icon_data);
+  icon_data.uFlags |= NIF_INFO;
+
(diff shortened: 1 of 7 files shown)


shell/browser/ui/win/notify_icon.cc
```diff
@@ -138,6 +138,16 @@ void NotifyIcon::DisplayBalloon(HICON icon,
...
 }
 
+void NotifyIcon::RemoveBalloon() {
+  NOTIFYICONDATA icon_data;
+  InitIconData(&icon_data);
+  icon_data.uFlags |= NIF_INFO;
+
+  BOOL result = Shell_NotifyIcon(NIM_MODIFY, &icon_data);
+  if (!result)
+    LOG(WARNING) << "Unable to remove status tray balloon.";
```
codebytere: As this stands, this will fail silently for end users. Do we think that's the best behavior in this case? We may want to throw an error or at least surface one through some means?

miniak: `tray.displayBalloon()` has the same issue. If we want to propagate the error to the caller, it should be done in both APIs IMO

codebytere: i'd be in favor of that personally, what do you think?

PR #19556, opened 2019-07-31
refactor: convert ipc-renderer.ts to TypeScript

miniak:
#### Description of Change
Convert `ipc-renderer.ts` to TypeScript. Also make the `ipc` binding typesafe.

Note: there is an issue with `ipcRenderer.sendToAll`, which is not documented. We can either remove it or document it.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Removed undocumented `ipcRenderer.sendToAll()`.


lib/renderer/api/ipc-renderer.ts
```diff
@@ -23,15 +21,9 @@ ipcRenderer.sendTo = function (webContentsId, channel, ...args) {
   return ipc.sendTo(internal, false, webContentsId, channel, args)
 }
 
-ipcRenderer.sendToAll = function (webContentsId, channel, ...args) {
```
nornagon: I'm pretty sure we use this for our chrome extension stuff.

miniak: we use `ipcRendererInternal.sendToAll()` there

miniak: https://github.com/electron/electron/blob/master/typings/internal-electron.d.ts#L65-L67

miniak: this is about the public API

miniak: @MarshallOfSound do you think the public API should be documented instead being removed as undocumented?

typings/internal-ambient.d.ts
```diff
@@ -14,6 +14,14 @@ declare namespace NodeJS {
     isComponentBuild(): boolean;
   }
 
+  interface IpcBinding {
+    send(internal: boolean, channel: string, args: any[]): void;
+    sendSync(internal: boolean, channel: string, args: any[]): any;
+    sendToHost(channel: string, args: any[]): void;
+    sendTo(internal: boolean, sendToAll: boolean, webContentsId: number, channel: string, args: any[]): void;
+    invoke(channel: string, args: any[]): Promise<{ error: string, result: any }>;
```
MarshallOfSound:
```suggestion
    invoke<T>(channel: string, args: any[]): Promise<{ error: string, result: T }>;
```

ConradIrwin:
@miniak We are currently using ipcRenderer.sendToAll() in our electron app at Superhuman to emulate broadcast channels.

Is there a replacement and/or can we reimplement it ourselves easily?

PR #19574, opened 2019-08-01
refactor: replace ipcRendererUtils.invoke() with ipcRendererInternal.invoke()

miniak:
#### Description of Change
Replace `ipcRendererUtils.invoke()` with newly added `ipcRendererInternal.invoke()`, which is re-using the existing mojo implementation of `ipcRenderer.invoke()` by just adding the `internal` flag.

Stop using separate request / response IPCs. Simplifies the code by removing duplicate abstraction.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/ipc-main-internal-utils.ts
+++ b/lib/browser/ipc-main-internal-utils.ts
@@ -1,26 +1,16 @@
 import { ipcMainInternal } from '@electron/internal/browser/ipc-main-internal'
 import * as errorUtils from '@electron/internal/common/error-utils'
 
-type IPCHandler = (event: ElectronInternal.IpcMainInternalEvent, ...args: any[]) => any
-
-const callHandler = async function (handler: IPCHandler, event: ElectronInternal.IpcMainInternalEvent, args: any[], reply: (args: any[]) => void) {
-  try {
-    const result = await handler(event, ...args)
-    reply([null, result])
-  } catch (error) {
-    reply([errorUtils.serialize(error)])
-  }
-}
+type IPCHandler = (event: Electron.IpcMainInvokeEvent, ...args: any[]) => any
 
 export const handle = function <T extends IPCHandler> (channel: string, handler: T) {
-  ipcMainInternal.on(channel, (event, requestId, ...args) => {
-    callHandler(handler, event, args, responseArgs => {
-      if (requestId) {
-        event._replyInternal(`${channel}_RESPONSE_${requestId}`, ...responseArgs)
-      } else {
-        event.returnValue = responseArgs
-      }
-    })
+  ipcMainInternal.handle(channel, handler)
+  ipcMainInternal.on(channel, async (event, ...args) => {
+    try {
+      event.returnValue = [null, await handler(event, ...args)]
+    } catch (error) {
+      event.returnValue = [errorUtils.serialize(error)]
+    }
   })
 }
 
--- a/lib/browser/ipc-main-internal-utils.ts
+++ b/lib/browser/ipc-main-internal-utils.ts
@@ -3,8 +3,7 @@ import * as errorUtils from '@electron/internal/common/error-utils'
 
 type IPCHandler = (event: Electron.IpcMainInvokeEvent, ...args: any[]) => any
 
-export const handle = function <T extends IPCHandler> (channel: string, handler: T) {
-  ipcMainInternal.handle(channel, handler)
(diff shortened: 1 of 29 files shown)


miniak:
@nornagon can you please have a look?

lib/renderer/ipc-renderer-internal.ts
```diff
@@ -20,3 +20,9 @@ ipcRendererInternal.sendTo = function (webContentsId, channel, ...args) {
 ipcRendererInternal.sendToAll = function (webContentsId, channel, ...args) {
   return ipc.sendTo(internal, true, webContentsId, channel, args)
 }
+
+ipcRendererInternal.invoke = async function<T> (channel: string, ...args: any[]) {
+  const { error, result } = await ipc.invoke<T>(internal, channel, args)
+  if (error) { throw new Error(`Error invoking remote method '${channel}': ${error}`) }
```
MarshallOfSound: The throw here should be on a new line

lib/renderer/ipc-renderer-internal.ts
```diff
@@ -20,3 +20,9 @@ ipcRendererInternal.sendTo = function (webContentsId, channel, ...args) {
 ipcRendererInternal.sendToAll = function (webContentsId, channel, ...args) {
   return ipc.sendTo(internal, true, webContentsId, channel, args)
 }
+
+ipcRendererInternal.invoke = async function<T> (channel: string, ...args: any[]) {
+  const { error, result } = await ipc.invoke<T>(internal, channel, args)
```
MarshallOfSound: Am I blind, where is `internal` coming from?

MarshallOfSound: I think this should just be `true`

lib/renderer/ipc-renderer-internal-utils.ts
```diff
@@ -4,38 +4,22 @@ import * as errorUtils from '@electron/internal/common/error-utils'
...
+      event.sender.send(replyChannel, null, await handler(event, ...args))
+    } catch (error) {
+      event.sender.send(replyChannel, errorUtils.serialize(error))
+    }
   })
 }
 
-let nextId = 0
-
 export function invoke<T> (command: string, ...args: any[]) {
```
MarshallOfSound: This helper seems unnecessary, can we remove and inline the `ipcRendererInternal.invoke` calls instead

lib/renderer/ipc-renderer-internal.ts
miniak: it's declared at the top of the file, this was done long time ago to "document" the arguments, all other calls are doing the same

lib/renderer/ipc-renderer-internal-utils.ts
miniak: I wanted to do it in a separate PR

miniak:
@MarshallOfSound we have tests for APIs built on top of `invoke` / `handle`

lib/renderer/ipc-renderer-internal.ts
miniak: I can do that, I kept the same formatting as https://github.com/electron/electron/blob/master/lib/renderer/api/ipc-renderer.ts#L26

miniak: done

miniak:
@MarshallOfSound are you ok with the PR?

miniak:
@deepak1556 can you please review?

miniak:

lib/browser/ipc-main-internal-utils.ts
```diff
@@ -1,26 +1,16 @@
...
-  ipcMainInternal.on(channel, (event, requestId, ...args) => {
-    callHandler(handler, event, args, responseArgs => {
-      if (requestId) {
-        event._replyInternal(`${channel}_RESPONSE_${requestId}`, ...responseArgs)
-      } else {
-        event.returnValue = responseArgs
-      }
-    })
+  ipcMainInternal.handle(channel, handler)
+  ipcMainInternal.on(channel, async (event, ...args) => {
```
nornagon: Registering two separate handlers seems like a bad idea to me. Can we just call `ipcMainInternal.handle` directly for when we want to call with `invoke`?

miniak: done

PR #19575, opened 2019-08-01
refactor: move file path gin converter to new file

DeerMichel:
#### Description of Change
[Beginning of an endless PR series]

This converter was already implemented in `gin` and simply adapted to `mate` - however, now that we'll migrate completely, it deserves its own file.

cc @MarshallOfSound @deepak1556 @codebytere 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/common/gin_converters/file_path_converter.h
+++ b/shell/common/gin_converters/file_path_converter.h
@@ -5,11 +5,8 @@
 #ifndef SHELL_COMMON_GIN_CONVERTERS_FILE_PATH_CONVERTER_H_
 #define SHELL_COMMON_GIN_CONVERTERS_FILE_PATH_CONVERTER_H_
 
-// #include <string>
-
 #include "base/files/file_path.h"
-// #include "shell/common/gin_converters/string16_converter.h" //
-// TODO(deermichel): check windows
+#include "shell/common/gin_converters/string16_converter.h"
 
 namespace gin {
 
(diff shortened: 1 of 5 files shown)


shell/common/gin_converters/file_path_converter.h
```diff
@@ -0,0 +1,40 @@
...
+// found in the LICENSE file.
+
+#ifndef SHELL_COMMON_GIN_CONVERTERS_FILE_PATH_CONVERTER_H_
+#define SHELL_COMMON_GIN_CONVERTERS_FILE_PATH_CONVERTER_H_
+
+// #include <string>
+
+#include "base/files/file_path.h"
+// #include "shell/common/gin_converters/string16_converter.h" //
+// TODO(deermichel): check windows
```
nornagon: what does this mean?

shell/common/native_mate_converters/file_path_converter.h
```diff
@@ -5,36 +5,7 @@
...
-      *out = base::FilePath(path);
-      return true;
-    } else {
-      return false;
-    }
-  }
-};
-
-}  // namespace gin
+#include "shell/common/gin_converters/file_path_converter.h"
```
nornagon: instead of redirecting, just update the files that include this.

shell/common/gin_converters/file_path_converter.h
DeerMichel: I didn't see why these includes are required and therefore wait for CI to see whether it fails on Windows.

shell/common/native_mate_converters/file_path_converter.h
DeerMichel: Sorry, I didn't get that. Do you mean we should create the gin converters without changing the existing native converters to use them?

nornagon: ah, nvm, I see now that the native_mate version just calls the gin version. this is all good 👍

PR #19576, opened 2019-08-01
docs: Fix `process.enablePromiseAPI` typo in promisification.md

flotwig:
#### Description of Change
The docs indicate that you can use `process.enablePromiseAPI`, but the actual API seems to be `process.enablePromiseAPIs`. Fixed and a link added.

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/docs/api/modernization/promisification.md
+++ b/docs/api/modernization/promisification.md
@@ -2,7 +2,7 @@
 
 The Electron team is currently undergoing an initiative to convert callback-based functions in Electron to return Promises. During this transition period, both the callback and Promise-based versions of these functions will work correctly, and will both be documented.
 
-To enable deprecation warnings for these updated functions, use the [`process.enablePromiseAPIs` runtime flag](https://github.com/electron/electron/blob/master/docs/api/process.md#processenablepromiseapis).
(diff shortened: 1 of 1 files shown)


docs/api/modernization/promisification.md
```diff
@@ -2,7 +2,7 @@
 
 The Electron team is currently undergoing an initiative to convert callback-based functions in Electron to return Promises. During this transition period, both the callback and Promise-based versions of these functions will work correctly, and will both be documented.
 
-To enable deprecation warnings for these updated functions, use the `process.enablePromiseAPI` runtime flag.
+To enable deprecation warnings for these updated functions, use the [`process.enablePromiseAPIs` runtime flag](https://github.com/electron/electron/blob/master/docs/api/process.md#processenablepromiseapis).
```
MarshallOfSound: This should be a relative path link

flotwig: All of the other doc links in the file are absolute, do you still want me to change it?

MarshallOfSound: Yeah, relative paths will be linted to ensure that if we ever move files around we get warned that the link no longer works.  Absolute URLs like that aren't checked and may break

flotwig: Gotcha, fixed, thanks

PR #19577, opened 2019-08-01
refactor: move native window gin converter

DeerMichel:
#### Description of Change
[Continuation of an endless PR series]

I put this converter in `message_box_converter.h` at first because it was sufficient for the `dialog` migration (#19482). However, this temporary approach quickly turned out to be a bad decision for further migrations. ~So let's place it right beneath the original one.~ Now, it's moved to a new file.

cc @nornagon @deepak1556 @codebytere 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


deepak1556:
Also is future refactors make sure to move out converters,

* If converter is inlined in a header file
* If converter is inlined in source file, you can leave it as it is unless the same converter is to be duplicated elsewhere.

DeerMichel:
I'm 💯 with you... this will mitigate some weird issues i'm encountering with further migrations :)

shell/common/native_mate_converters/file_dialog_converter.cc
```diff
@@ -8,6 +8,7 @@
 #include "shell/browser/api/atom_api_browser_window.h"
```
deepak1556: Api header shouldn't be needed now.

DeerMichel:
huh, ill take a look

shell/common/native_mate_converters/file_dialog_converter.cc
DeerMichel: We need it in [L.65](https://github.com/electron/electron/pull/19577/files#diff-2c4b6a568b2c88fa116f70644e3fd219R65), at least my compiler says so 😄.

DeerMichel:
All right, should be fine now 👍

PR #19582, opened 2019-08-01
ci: enable crashReporter tests on mac

nornagon:
#### Description of Change
These have been disabled for ages but they work, so un-disable them.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


spec/api-crash-reporter-spec.js
```diff
@@ -15,15 +15,15 @@ const { app, BrowserWindow, crashReporter } = remote
...
+const ifdescribe = (condition) => {
+  return condition ? describe : describe.skip
+}
 
+// TODO(alexeykuzmin): [Ch66] Fails on linux. Fix it and enable back.
+ifdescribe(!process.mas && process.platform !== 'linux')('crashReporter module', () => {
   let originalTempDirectory = null
   let tempDirectory = null
-  const specTimeout = 180000
+  const specTimeout = 10000
```
zcbenz: The tests still seem to be flaky on CI, as you can see in the CI results of this PR.

nornagon: I don't think the results are flaky, I think they're legitimately failing. cc @nitsakh

nornagon:
NB. the test "should send minidump with updated extra parameters when node processes crash" has been removed, because with crashpad, you can't update the parameters in child processes any more.

PR #19586, opened 2019-08-02
fix: handle edge behavior for about panel on Linux

erickzhao:
#### Description of Change
This PR fixes two small issues:
* If `app.showAboutPanel()` would be called before `app.setAboutPanelOptions()`, the app would crash. This PR handles this edge case gracefully by skipping the option logic if none are there and logging a warning to the user.
* Upon closing the About dialog, there were previously warnings from GTK stemming from the `g_clear_object(&dialog)` call.
Screenshot from 2019-08-01 18-49-47
This has now been replaced with a call to destroy the `GtkWidget` after closing the About dialog.

cc @codebytere @ckerr 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed bug where app would crash if `app.showAboutPanel()` was called before setting any About panel options on Linux.

### DIFF
--- a/shell/browser/browser_linux.cc
+++ b/shell/browser/browser_linux.cc
@@ -147,60 +147,62 @@ bool Browser::IsEmojiPanelSupported() {
 }
 
 void Browser::ShowAboutPanel() {
+  const auto& opts = about_panel_options_;
+
+  if (!opts.is_dict()) {
+    LOG(WARNING) << "Called showAboutPanel(), but didn't use "
+                    "setAboutPanelSettings() first";
+    return;
+  }
+
   GtkWidget* dialogWidget = gtk_about_dialog_new();
   GtkAboutDialog* dialog = GTK_ABOUT_DIALOG(dialogWidget);
 
-  const auto& opts = about_panel_options_;
   const std::string* str;
   const base::Value* val;
 
-  if ((opts.is_dict())) {
-    if ((str = opts.FindStringKey("applicationName"))) {
-      gtk_about_dialog_set_program_name(dialog, str->c_str());
-    }
-    if ((str = opts.FindStringKey("applicationVersion"))) {
-      gtk_about_dialog_set_version(dialog, str->c_str());
-    }
-    if ((str = opts.FindStringKey("copyright"))) {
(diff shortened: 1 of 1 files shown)


shell/browser/browser_linux.cc
```diff
@@ -147,58 +147,64 @@ bool Browser::IsEmojiPanelSupported() {
...
+      }
+      if (cstrs.empty()) {
+        LOG(WARNING) << "No author strings found in 'authors' array";
+      } else {
+        cstrs.push_back(nullptr);  // null-terminated char* array
+        gtk_about_dialog_set_authors(dialog, cstrs.data());
+      }
     }
+  } else {
+    LOG(WARNING) << "Called showAboutPanel(), but didn't use "
```
MarshallOfSound: Unclear what we do on other platforms but should this throw an error?

Also should we short-circuit this and not even allocate the GtkWidget if no options are set?

codebytere: I'm in agreement w the above: this should be short-circuited i would say.

shell/browser/browser_linux.cc
```diff
@@ -147,58 +147,64 @@ bool Browser::IsEmojiPanelSupported() {
...
-    // set preserve_aspect_ratio to true
-    GdkPixbuf* icon =
-        gdk_pixbuf_new_from_file_at_size(str->c_str(), width, height, &error);
-    if (error != nullptr) {
-      g_warning("%s", error->message);
-      g_clear_error(&error);
-    } else {
-      gtk_about_dialog_set_logo(dialog, icon);
-      g_clear_object(&icon);
+  if ((opts.is_dict())) {
```
codebytere: extra parens here

shell/browser/browser_linux.cc
erickzhao: > Unclear what we do on other platforms but should this throw an error?
* On macOS, I believe there are defaults based on the `.plist` file.
* Windows behaviour is TBD, as the PR is still open (#19420).

> Also should we short-circuit this and not even allocate the GtkWidget if no options are set?
So just not display the panel at all? I could make that happen.

codebytere:
@erickzhao please backport this to 5 & 6!

Issue #19474, opened 2019-07-26
Silent printing not working in combination with deviceName (in v5.0.7 +)


ibra038:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success. 
  * #17260 Not relevant: addresses the same issue, but for a version before the pull request that supposedly fixed silent printing. is already closed.
  * #17077 Not relevant: addresses the same issue, but for an older version before pull request #19038 that should fix silent printing. 
  * #18123 Not relevant: doesn't address the same issue. Is already closed
  * #17523 A list of issues related to printing. 

### Issue Details

* **Electron Version:**
  * 5.0.7 and 6.0.0-beta.15 
* **Operating System:**
  * MacOS Mojave v10.14.4
* **Last Known Working Electron version:**
  * 3.1.11

### Expected Behaviors
According to the pull request #19038 which was merged in v.5.0.7, silent printing should work. 

&#35; | Code | Expected Behavior
------ | ----- | -------------------
1 | `webContents.print()` | Opens the print preview
2 | `webContents.print({ silent: true })` | Prints silently to default printer
3 | `webContents.print({ silent: true, printBackground: true })` | Prints silently with background to default printer
4 | `webContents.print({ silent: true, deviceName: 'Canon_MF633C_635C' })` | Prints silently to my Canon printer
5 | `webContents.print({ silent: true, deviceName: printBackground: true, deviceName: 'Canon_MF633C_635C' })` | Prints silently with background to my Canon printer

Original documentation of WebContents.print() can be found [here](https://electronjs.org/docs/api/web-contents#contentsprintoptions-callback).

### Actual Behaviors
&#35; | Actual Behavior
------ |  ---------------
1 | :white_check_mark: As expected
2 | :white_check_mark: As expected
3 | :white_check_mark: As expected
4 | :x: Nothing will happen, callback is never called
5 | :x: Nothing will happen, callback is never called

Unfortunately when providing a deviceName, nothing is send to the printer and the callback will never be called (see image of terminal logs below).

### To Reproduce

1. Clone this fork: https://github.com/ibra038/electron-quick-start
2. Run `yarn install` and `yarn start` 
[...]

Issue #19122, opened 2019-07-05
webContents.print() doesn't work under Windows when 'deviceName' is provided


copystyle:
* Electron Version 6.0.0-beta.12
* Operating System Windows 10 64 bit

Expected Behavior
webContents.print() should send print a job to the specified printer

Actual behavior
Nothing happens. Callback gets called much latter, for example on program exit.

To Reproduce
let print_options = { 
    silent: true,
    printBackground: true,
    deviceName: barcode_device_name 
}
print_window.webContents.print(print_options, (data) => {
    console.log(`print result: ${data}`)
})

Related to #17260

zlab:
electron 6.0.0, the same problem on mac

jriffel:
I guess I don't understand the release process for Electron.  I've been following this bug and its in the "Fixed for Next Release" column on the 6.0.x project.  I've been waiting for 6.0.3 to release so I could get this bug fix which we've been desperately waiting on.  6.0.3 just released and this bug is still sitting in the "Fixed for Next Release" column.  What is the release process so I can understand when this fix might be released?

jriffel:
@sofianguy / @jkleinsc can I ask when this bug fix might be released on 6.0.x?  I see that 6.0.4 is releasing and it this still did not make it.  Is there a release process documented anywhere that would help me understand when this bug fix might move to release?  In the 6.0.x project it is sitting in the Fixed for Next Release column but there have been at least 2 releases since it was placed there.  Thanks for your contributions!

jriffel:
6.0.5 and 6.0.6 have now all been released and this bug still remains "Fixed for Next Release":

https://github.com/electron/electron/projects/20

Who can I pay to merge this for the next release :)

popod:
@jriffel it seems that "project" cannot be trusted to know when an issue will be released..

I've opened this issue to find ways to pay for issues : #18113.

jkleinsc:
@jriffel @popod Open source is largely volunteer-driven and collaborative. Because of this, time and energy to get things done are often the scarcest resources. We want Electron to be the best that it can be, but we also have to be realistic about what the team maintaining Electron can achieve. This means that:

* We might not have an ETA for when an issue will get investigated or fixed.
* Something you want implemented might not get implemented.
* A change you want merged might not get merged.
* Insisting on or making demands for any of the above is considered against [the Electron Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md)

[...]

Issue #19108, opened 2019-07-04
Electron crashed after print


YiiGuxing:
Version: v5.0.5, v5.0.6
JavaScript code: `document.querySelector('iframe').contentWindow.print()`

MarshallOfSound:
Thank you for taking the time to report this issue and helping to make Electron better.

Please provide a standalone test that a tester could run to reproduce the issue. This makes fixing issues go more smoothly: not only does it ensure the right thing is tested, but it also prevents triage from bottlenecking on making testers write new tests for each new issue.

I'm setting the more-information-needed label for the above reasons. After you've responded, please @ me in a followup comment.

Thanks in advance! Your help is appreciated.

MarshallOfSound:
Thanks for reaching out!

We require the template to be filled out on all new issues and pull requests. We do this so that we can be certain we have all the information we need to address your submission efficiently. This allows the maintainers to spend more time fixing bugs, implementing enhancements, and reviewing and merging pull requests.

Thanks for understanding and meeting us half way :grinning:

YiiGuxing:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.5
  * 5.0.6
* **Operating System:**
  * Windows 10 1903 x64
* **Last Known Working Electron version:**
  * 5.0.4

### Expected Behavior
1. Electron does not crash.
2. Electron responds to UI events.

### Actual Behavior
1. Electron crashed.
2. Electron does not respond to UI events.

### To Reproduce
1. `$ git clone https://github.com/electron/electron-api-demos.git`
2. Modify `package.json`: `"electron": "^5.0.1"` -> `"electron": "^5.0.6"`
3. `$ npm install`
4. Add a button to `index.html`: `<button onclick="window.print()">Print</button>`
5. `$ npm run dev`
6. Click the print button
    - Click the print buuton  on the dialog: *Electron crashed after print.*
    - Click the cancel buuton  on the dialog: *Electron does not respond to UI events after the dialog closes.*

### Screenshots
[...]

Issue #17077, opened 2019-02-21
webContents.print() not print as per the default printer settings when putting device name 


shabeer-mdy:
* **Electron Version** (output of `node_modules/.bin/electron --version`):
  * v4.0.5
* **Operating System** (Platform and Version):
  *  Windows 10 

### Expected Behavior
using  webContents api to print an html file oriention  set to landscape in default printer settings and page size set to A5, 

mainWindow.webContents.print({ silent: true, printBackground: false, deviceName: 'HP LaserJet 1020' })

My issue is : it is not printing as defined in the default printer settings

but it currently printing is deviceName not set

and need to send mainWindow.webContents.print() api to silently and printer name 

### To Reproduce
If you provide a URL, please list the commands required to clone/setup/run your repo e.g.
```sh
$ git clone git@github.com:shabeer-mdy/electron-quick-start.git
$ npm install
$ npm start || electron .
```
and set any printer name to  mainWindow.webContents.print({ silent: true, printBackground: false, deviceName: 'HP LaserJet 1020' })

set any installed printer name

rhys101:
It looks like it's related to https://github.com/electron/electron/issues/16792. The arguments aren't being passed through correctly inside webContents.print -  here's the pull request that should fix it: https://github.com/electron/electron/pull/17052

popod:
@shabeer-mdy is your issue resolved by #17052 (electron 4.0.7) ? If yes, please close this issue ;)

shabeer-mdy:
#17052 it is not working as I expected, printer option set silent is true it is not called the printer

shabeer-mdy:
#17052 wouldn't solve my problem, In printer options when changing silent set to true then it's not calling any printer and silent is set to false then the print popup is getting. also same issue in device name property

PR #19598, opened 2019-08-02
fix: crash on print cancellation and silent print settings

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/19122.
Fixes https://github.com/electron/electron/issues/17077.
Fixes https://github.com/electron/electron/issues/19108.
Fixes https://github.com/electron/electron/issues/19474.

In an earlier PR, I introduced the ability to set custom print options in `webContents.print()`.

However, this meant that the `base::DictionaryValue` passed in with these settings would never be `null`, because if even one option is set than they _all_ have to be set with a default to prevent a DCHECK that gets hit if [this function](https://cs.chromium.org/chromium/src/printing/print_settings_conversion.cc?type=cs&q=printsettingsfromjob&g=0&l=78) returns false. As a result, only the `PrintHostMsg_UpdatePrintSettings` [IPC message](https://github.com/electron/electron/blob/4bea46cccdc8f8ed0dd9c015dc8b496a3fc9a268/patches/chromium/printing.patch#L504-L505) would be sent, and printing would silently fail as no defaults were properly initialized. More specifically, since no defaults were set, when a new `PrinterQuery` was created it would have no idea about previously existing or defaulted settings, since settings were reset at the beginning of print updates.

This fixes that by changing codepaths to _intentionally_ only send the IPC message to update print settings (`PrintHostMsg_UpdatePrintSettings`), and then each time pursuing the following codepath: 
1) Reset settings from previous job
2) Call `printing_context_->UseDefaultSettings()` in order to set the `settings_` member with default settings
3) Call `printing_context_->UpdatePrintSettings()` with our new settings to only change the options that were passed in.

These custom settings should work now for both silent and non-silent printing: this can be seeing by calling:

```
win.webContents.print({ copies: 3 }, (success, reason) => {
})
```

to see that '3' is automatically placed in the UI settings of the native print dialog.

**Secondary Fix:**

Due to changes in the lifetime of PrinterQuery, a crash occurred if we posted a task to notify of print cancellation and did _not_ also call the callback. This removes that gating and now will call the callback regardless of cancellation status.

I chose not to put this secondary fix in a standalone PR because the primary issue above meant that it isn't even being surfaced to users as it causes a silent failure before a UI printing dialog is ever shown.

cc @jkleinsc @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash on manual print cancellation as well as an issue with `deviceName` not working.

### DIFF
--- a/patches/chromium/printing.patch
+++ b/patches/chromium/printing.patch
@@ -11,7 +11,7 @@ majority of changes originally come from these PRs:
 This patch also fixes callback for manual user cancellation and success.
 
 diff --git a/chrome/browser/printing/print_job_worker.cc b/chrome/browser/printing/print_job_worker.cc
-index 63f432b58371cfa0f8079fa78a51c8865a00c183..7b39523e0b8b840191ea517d5f5e8eda701995bc 100644
+index 63f432b58371cfa0f8079fa78a51c8865a00c183..d20d803f55ca67fb6993facc69c3431767786053 100644
 --- a/chrome/browser/printing/print_job_worker.cc
 +++ b/chrome/browser/printing/print_job_worker.cc
 @@ -21,12 +21,12 @@
@@ -28,11 +28,28 @@ index 63f432b58371cfa0f8079fa78a51c8865a00c183..7b39523e0b8b840191ea517d5f5e8eda
  #include "printing/print_job_constants.h"
  #include "printing/printed_document.h"
  #include "printing/printing_utils.h"
-@@ -224,7 +224,14 @@ void PrintJobWorker::UpdatePrintSettingsFromPOD(
+@@ -206,9 +206,14 @@ void PrintJobWorker::SetSettingsFromPOD(
+ void PrintJobWorker::UpdatePrintSettings(base::Value new_settings,
+                                          SettingsCallback callback) {
+   DCHECK_CURRENTLY_ON(BrowserThread::UI);
+-  PrintingContext::Result result =
+-      printing_context_->UpdatePrintSettings(std::move(new_settings));
+-  GetSettingsDone(std::move(callback), result);
++  // Reset settings from previous print job
++  printing_context_->ResetSettings();
++  PrintingContext::Result get_default_result = printing_context_->UseDefaultSettings();
++  if (get_default_result == PrintingContext::Result::OK) {
++    PrintingContext::Result update_result =
++            printing_context_->UpdatePrintSettings(std::move(new_settings));
++    GetSettingsDone(std::move(callback), update_result);
++  }
+ }
+ 
+ #if defined(OS_CHROMEOS)
+@@ -224,6 +229,13 @@ void PrintJobWorker::UpdatePrintSettingsFromPOD(
  
  void PrintJobWorker::GetSettingsDone(SettingsCallback callback,
                                       PrintingContext::Result result) {
--  std::move(callback).Run(printing_context_->settings(), result);
 +  if (result == PrintingContext::CANCEL) {
 +    print_job_->PostTask(
 +      FROM_HERE,
@@ -40,10 +57,9 @@ index 63f432b58371cfa0f8079fa78a51c8865a00c183..7b39523e0b8b840191ea517d5f5e8eda
 +                    JobEventDetails::USER_INIT_CANCELED, 0,
 +                    base::RetainedRef(document_)));
 +  }
-+  std::move(callback).Run(printing_context_->settings(), result);
+   std::move(callback).Run(printing_context_->settings(), result);
  }
  
- void PrintJobWorker::GetSettingsWithUI(int document_page_count,
 diff --git a/chrome/browser/printing/print_view_manager_base.cc b/chrome/browser/printing/print_view_manager_base.cc
 index c4e0992f6265b34659514ef5f15eb8d78645161c..1aca2f88da5d8e96a0f16a667a8a86a7873dfdf9 100644
 --- a/chrome/browser/printing/print_view_manager_base.cc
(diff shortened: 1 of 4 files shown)


fleabitten232:
I'm currently using your changes in a small, heavily print dependent project that needs to silently print to both a standard A4 size/laser printer combination and a notoriously temperamental 6x4inch thermal label printer. I'm happy to share my experiences if you would find that helpful -- just let me know where if here isn't appropriate. 

I've now mostly finished moving from electron@latest to electron@beta which broke a label design that was previously working. In latest, it would correctly print landscape (albeit not silently). 

In the beta (without any code/css changes), the same label was adamant that it wanted to be portrait. It would print massively scaled down so that the landscape design would fit to the width of a portrait page. This happened with 'silent' sent to both true and false.

I couldn't immediately fix the problem through any print dialog settings and unfortunately I ran out of labels before I could explore deeper. The last thing I noticed was that setting the page to 'landscape' would actually rotate it 180 degrees, rather than 90. Although the printer does have a setting to rotate labels 180 degrees, I'm wondering if the page is somehow being 'landscaped' twice.

I'll have some more labels in the next 24 hours so I'll keep you updated. Unfortunately the project has a looming deadline so if I can't get it working soon I'll have to just rotate the label design 90deg using CSS as a hack and call it done.

spec/ts-smoke/electron/main.ts
```diff
@@ -83,7 +83,7 @@ app.on('ready', () => {
   mainWindow.webContents.setVisualZoomLevelLimits(50, 200)
   mainWindow.webContents.setLayoutZoomLevelLimits(50, 200)
 
-  mainWindow.webContents.print({ silent: true, printBackground: false })
+  mainWindow.webContents.print({ silent: true })
```
jkleinsc: Is this change necessary?

codebytere:
@deepak1556 i've managed to fix the underlying issue, and have updated the PR body accordingly :)

spec/ts-smoke/electron/main.ts
codebytere: nope, removed with latest changes!

patches/chromium/printing.patch
```diff
@@ -576,3 +580,55 @@ index 71c0c15217b62cd7a6087c6d9ae50481f9041d5f..e2250a66517dbd909cd3b5407173ad91
...
+diff --git a/printing/print_settings_conversion.cc b/printing/print_settings_conversion.cc
+index 2563ae6a87b2354ff2f2b45c17f61d2f44910efa..1f34f905791f38a44ae8c892cc9cfb38d15b659d 100644
+--- a/printing/print_settings_conversion.cc
++++ b/printing/print_settings_conversion.cc
+@@ -190,11 +190,13 @@ bool PrintSettingsFromJobSettings(const base::Value& job_settings,
+ 
+   settings->set_dpi_xy(dpi_horizontal.value(), dpi_vertical.value());
+ #endif
++  std::string print_device_name = *device_name;
++  if (print_device_name != "")
```
deepak1556: Can be `!device_name->empty()`

shell/browser/api/atom_api_web_contents.cc
```diff
@@ -1643,6 +1651,8 @@ void WebContents::Print(mate::Arguments* args) {
   options.Get("landscape", &landscape);
   settings.SetBoolean(printing::kSettingLandscape, landscape);
 
+  // We set the default to empty string here and only update
+  // if at the Chromium level if it's non-empty
```
deepak1556: What happens when an empty device name is passed along ?

patches/chromium/printing.patch
```diff
@@ -496,24 +509,17 @@ index ef580254bd8feba84ac02924b77b9b4feaf14d96..57a8264a7174b440ed1d6bfe9c2c3e21
...
-+    bool canceled = false;
-+    Send(new PrintHostMsg_UpdatePrintSettings(
-+        routing_id(), cookie, new_settings, &settings, &canceled));
-+    if (canceled)
-+      return false;
-+  }
++  // new_settings will never be empty, always send the update IPC message
++  bool canceled = false;
++  Send(new PrintHostMsg_UpdatePrintSettings(
++      routing_id(), 0, new_settings, &settings, &canceled));
```
deepak1556: So by setting the cookie value to be zero we create a new PrinterQuery every time instead of reusing one if exists for same settings. Any reason this is the case ?

codebytere: @deepak1556 this was codification of implicit behavior; becuase we set the defaults ourselves the params member would never been initialized, so that ternary would always have evaluated to 0 (i did some testing to ensure this was the case)

shell/browser/api/atom_api_web_contents.cc
codebytere: https://github.com/electron/electron/pull/19598#discussion_r310895212 that's what is checked here; if no value is passed by the user it is set to the value initialized in `printing_context_->UseDefaultValues()` which is the system's default printer.

codebytere:
CI failures are unrelated and will be fixed by: https://github.com/electron/electron/pull/19652

shell/browser/api/atom_api_web_contents.cc
deepak1556: Sorry for being unclear, I meant what happens when we pass along empty device string, does it still pick the default value or crash ? I am trying to understand if its necessary for an empty device name check.

codebytere: it errors if we don't do this, since it sets the printer to empty string which necessarily won't exist!

deepak1556: Got it, thanks for clarifying.

Issue #19588, opened 2019-08-02
Win: app.clearRecentDocuments does not work


dospolov:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details
app.clearRecentDocuments does not work on Windows.

* **Electron Version:**
  * 5.0.8
* **Operating System:**
  * Windows 10 pro; build 1903
* **Last Known Working Electron version:**
  * 

### Expected Behavior
List of recent documents should be cleared

### Actual Behavior
List of recent documents continues to accumulate

### To Reproduce
Try `app.clearRecentDocuments();` on Windows

### Screenshots
![image](https://user-images.githubusercontent.com/670682/62343120-307b8f00-b4f2-11e9-853e-a8aac23b8ecd.png)
![image](https://user-images.githubusercontent.com/670682/62343111-20fc4600-b4f2-11e9-8277-189fe6f0a69b.png)

codebytere:
@dospolov when was the last time this worked?

dospolov:
@codebytere I'm new to electron, so I don't know ¯\_(ツ)_/¯

dospolov:
Issue is not fixed :(
electron version: 6.0.3

codebytere:
@dospolov i think this may be an issue with your app packaging or the way it's interfacing with roles and recent documents - this was tested to work on Windows 10 🤔

dospolov:
@codebytere I've checked source code of Electron Builder, and I haven't found any interactions with roles

PR #19599, opened 2019-08-02
fix: clearRecentDocuments role on Windows

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/19588.

Per [documentation](https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shaddtorecentdocs), `SHAddToRecentDocs` can also be used to clear all recent documents. This switches our impl to use that fn and passes `nullptr` as the second argument to clear all recent docs:

> Set this parameter to NULL to clear all usage data on all items.

cc @DeerMichel 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed the `clearRecentDocuments` MenuItem role on Windows.

### DIFF
--- a/shell/browser/browser_win.cc
+++ b/shell/browser/browser_win.cc
@@ -111,13 +111,10 @@ void Browser::AddRecentDocument(const base::FilePath& path) {
 }
 
 void Browser::ClearRecentDocuments() {
-  CComPtr<IApplicationDestinations> destinations;
-  if (FAILED(destinations.CoCreateInstance(CLSID_ApplicationDestinations, NULL,
-                                           CLSCTX_INPROC_SERVER)))
+  if (base::win::GetVersion() < base::win::Version::WIN7)
     return;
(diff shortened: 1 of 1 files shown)


shell/browser/browser_win.cc
```diff
@@ -111,13 +111,10 @@ void Browser::AddRecentDocument(const base::FilePath& path) {
 }
 
 void Browser::ClearRecentDocuments() {
-  CComPtr<IApplicationDestinations> destinations;
-  if (FAILED(destinations.CoCreateInstance(CLSID_ApplicationDestinations, NULL,
-                                           CLSCTX_INPROC_SERVER)))
-    return;
-  if (FAILED(destinations->SetAppID(GetAppUserModelID())))
+  if (base::win::GetVersion() < base::win::Version::WIN7)
```
zcbenz: We only support Windows 7 and later, so this is no needed.

PR #19601, opened 2019-08-02
fix: crash on window.print()

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/19539.

`window.print()` doesn't invoke `PrintViewManagerBase::PrintNowInternal`, so when we moved 

```
  registrar_.Add(this, chrome::NOTIFICATION_PRINT_JOB_EVENT,
                content::NotificationService::AllSources());
```

into that function from `PrintViewManagerBase::CreateNewPrintJob()` we inadvertently created a side-effect wherein in `PrintViewManagerBase::ReleasePrintJob()` it would try to unregister a notification listener which wasn't registered, and would therefore CHECK on ` found != registered_.end()`. 

This fixes that by guarding the unregistration with a null check on `callback_`, which would only be non-null when in the call chain of our `webContents.print()` implementation, since `callback_` is a member we patched in ourselves as part of that method.

cc @jkleinsc @PalmerAL @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

### DIFF
--- a/patches/chromium/printing.patch
+++ b/patches/chromium/printing.patch
@@ -46,7 +46,7 @@ index 63f432b58371cfa0f8079fa78a51c8865a00c183..7b39523e0b8b840191ea517d5f5e8eda
  
  void PrintJobWorker::GetSettingsWithUI(int document_page_count,
 diff --git a/chrome/browser/printing/print_view_manager_base.cc b/chrome/browser/printing/print_view_manager_base.cc
-index c4e0992f6265b34659514ef5f15eb8d78645161c..e95f62c0761d1a5829cc4403434fd643e45a0a69 100644
+index c4e0992f6265b34659514ef5f15eb8d78645161c..1aca2f88da5d8e96a0f16a667a8a86a7873dfdf9 100644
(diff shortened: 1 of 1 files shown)


rgarcia1990:
i have the same problem in electron v9, will it be solved soon?

codebytere:
@rgarcia1990 this should be fixed in v9. If you are still having issues, please fill out a new issue with a proper repro and other aspects of the template filled.

rgarcia1990:
> @rgarcia1990 this should be fixed in v9. If you are still having issues, please fill out a new issue with a proper repro and other aspects of the template filled.

it seems that the problem is when you try to print a page from version 9 using pdf viewer... it easy to reproduce just load a browserwindow from an url with pdf and try to print.

rgarcia1990:
> @rgarcia1990 this should be fixed in v9. If you are still having issues, please fill out a new issue with a proper repro and other aspects of the template filled.

  i created the issue for that problem. It is very important for me if you cant help me! Thanks for considering my request!
  https://github.com/electron/electron/issues/24458

PR #19612, opened 2019-08-03
fix: Patch download architecture for ARM (fixes ARM broken install)

pi1024e:
#### Description of Change
Electron no longer produces an 'arm' build, and now labels it 'armv7l'. As NPM still uses 'arm', we need to patch the variable name here or the download fails with a 404 error as such:

Error: GET https://github.com/electron/electron/releases/download/v3.1.3/electron-v3.1.3-linux-arm.zip returned 404
#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Redirected "arm" to "armv7l". Based on PR #16962, which was recently closed due to inactivity.


pi1024e:
Made this changes!

pi1024e:
cc: @miniak

pi1024e:
The function you mention being called only does something if the arch is arm. Calling this function for every arch is unnecessary then. Would you say so?

pi1024e:
Not needed at this time, as get does the work. Could do better, but I am closing this for now.

PR #19639, opened 2019-08-05
fix: return correct bounds on will-resize

### LINKED ISSUE #19611: BrowserWindow's "will-resize" event's callback is provided with bizarre newBounds value
### Preflight Checklist

### Issue Details

* **Electron Version:** 
  * 6.0.0
* **Operating System:** 
  * MacOS any
  * none that I know of

### Expected Behavior
The actual new bounds (correct x, y, width and height) to be provided as the new bounds to the callback of `will-resize` event

### Actual Behavior
On mac (and only mac) it provides some very bizarre values. It seems that the `y` and `x` are actually calculated from the BOTTOM-RIGHT corner of the DISPLAY bounds to the BOTTOM-RIGHT of the window bounds. Instead of the usual (by that I mean what you get from `getBounds`) `x` and `y` that is calculated from the TOP-LEFT of the DISPLAY to the TOP-LEFT of the window.

in other words imagine this:

My display is 100x100 and I have a window at 10, 20 with width: 30, and height 40

the `getBounds` function gives me (10, 20, 30, 40) all good.

when I resize from the TOP, UP for 1 pixel, the new bounds would be: (10, 19, 10, 41)

because the height is increased by one, while the `y` is reduced by one.

however the `will-resize` event on MAC gives me: (60, 40, 10, 41)

60 is the `display.width - bounds.width - bounds.x`.
40 is the `display.height - bounds.height - bounds.y`. why?!

and what's worse? there is no way to transform these, because not only the `x` and the `y` are wrong, they never reflect what is going on, that means, even if I resize from BOTTOM in the UP direction, it still gives me the same WRONG newBounds.

this means, there is absolutely no way to even work around this issue. I have no way of knowing whether the user dragged the TOP edge down or the BOTTOM edge down. the data is so bizarre that I spent a whole day trying to figure out what the values meant and how I could fix them. there is no way. It has to be fixed here

### To Reproduce
Simply listen to `will-resize` of any window and log the `newBounds` argument.

DeerMichel:
#### Description of Change
Fixes #19611.

Use `gfx::ScreenRectFromNSRect` to refer to the window's top-left corner as origin instead of the bottom-left one.

No tests since `will-resize` cannot be triggered programmatically - correct me if i'm wrong :)

cc @codebytere @pmkary

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed BrowserWindow's `will-resize` event returning wrong bounds on macOS.

### DIFF
--- a/shell/browser/ui/cocoa/atom_ns_window_delegate.mm
+++ b/shell/browser/ui/cocoa/atom_ns_window_delegate.mm
@@ -12,6 +12,7 @@
 #include "shell/browser/native_window_mac.h"
 #include "shell/browser/ui/cocoa/atom_preview_item.h"
 #include "shell/browser/ui/cocoa/atom_touch_bar.h"
+#include "ui/gfx/mac/coordinate_conversion.h"
 #include "ui/views/cocoa/native_widget_mac_ns_window_host.h"
 #include "ui/views/widget/native_widget_mac.h"
 
@@ -120,8 +121,10 @@ using TitleBarStyle = electron::NativeWindowMac::TitleBarStyle;
 
   {
     bool prevent_default = false;
-    gfx::Rect new_bounds(gfx::Point(sender.frame.origin), gfx::Size(newSize));
-    shell_->NotifyWindowWillResize(new_bounds, &prevent_default);
+    NSRect new_bounds = NSMakeRect(sender.frame.origin.x, sender.frame.origin.y,
+                                   newSize.width, newSize.height);
+    shell_->NotifyWindowWillResize(gfx::ScreenRectFromNSRect(new_bounds),
+                                   &prevent_default);
     if (prevent_default) {
       return sender.frame.size;
     }


alitaheri:
Will this also take position change into account?

I'm not a expert at reading the electron code base. What I'm worried about is that when the user grabs one of the `top`, `top-left`, `top-right`, `left`, `bottom-left` corners and tries to resize from there. it's not only the size that changes. the position will change too.

For example:

If I have a window at (x=10, y=10, width=20, height=30) and drag the left edge and move my mouse 5 pixels the `newBounds` passed to the `will-resize` callback must be (x=5, y=10, width=25, height=30) this is to keep the right edge at it's place. and it's crucial for our application to actually receive the correct `newBounds`. otherwise it won't be able to figure out which edge is being dragged.

So, please check this and make sure this behavior is provided. it is on windows.

Thank you very much 👍

DeerMichel:
I think so:

jj

alitaheri:
Perfect 😍 Thank you

pmkary:
So nice, thanks a lot @DeerMichel 😍

davej:
I believe there is a bug here. The `y` value changes on `newBounds` when the window is dragged from the bottom edge. It can be seen in @deermichel's gif above. This behaviour is also inconsistent with Windows where the `y` value doesn't change (correctly IMO).

ndr34:
+1 to @davej comment. When resizing from the bottom edge, I see y origin in newBounds differs from y origin in current bounds. Is there any way to workaround this issue and/or a plan to fix? Looking for some way to detect that a user has dragged the browser window from the bottom edge.

PR #19653, opened 2019-08-06
fix: fall back to default logs path in getPath('logs')

codebytere:
#### Description of Change

Follow-up to https://github.com/electron/electron/pull/19545#issuecomment-517044623.

If users try to call `app.getPath('logs')` before they've called `app.setAppLogsPath()` then it would previously error. Apps that relied on getPath('log') to work would get an exception, which is not great API.

Now, calling `app.getPath('logs')` without having previously called `app.setAppLogsPath()` will result in a directory being created and set to `/Library/Logs/YourAppName` on macOS, and inside the `userData` directory on Linux and Windows.

This also adds a converter for `base::Optional<T>` such that we don't introduce the possibility for `mate::Arguments*` pollution a la https://xkcd.com/327.

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where `app.getPath('logs')` would throw an error if the logs path was not previously set.

### DIFF
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -614,6 +614,8 @@ Returns `String` - The current application directory.
 Returns `String` - A path to a special directory or file associated with `name`. On
 failure, an `Error` is thrown.
 
+Calling `app.getPath('logs')` without having previously called `app.setAppLogsPath()` will result in a directory being created and set to `/Library/Logs/YourAppName` on _macOS_, and inside the `userData` directory on _Linux_ and _Windows_.
+
 ### `app.getFileIcon(path[, options])`
 
 * `path` String
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -614,7 +614,7 @@ Returns `String` - The current application directory.
 Returns `String` - A path to a special directory or file associated with `name`. On
 failure, an `Error` is thrown.
 
-Calling `app.getPath('logs')` without having previously called `app.setAppLogsPath()` will result in a directory being created and set to `/Library/Logs/YourAppName` on _macOS_, and inside the `userData` directory on _Linux_ and _Windows_.
+If `app.getPath('logs')` is called without called `app.setAppLogsPath()` being called first, a default log directory will be used equivalent to calling `app.setAppLogsPath()` without a `path` parameter.
 
 ### `app.getFileIcon(path[, options])`
 
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -614,7 +614,7 @@ Returns `String` - The current application directory.
 Returns `String` - A path to a special directory or file associated with `name`. On
 failure, an `Error` is thrown.
 
-If `app.getPath('logs')` is called without called `app.setAppLogsPath()` being called first, a default log directory will be used equivalent to calling `app.setAppLogsPath()` without a `path` parameter.
+If `app.getPath('logs')` is called without called `app.setAppLogsPath()` being called first, a default log directory will be created equivalent to calling `app.setAppLogsPath()` without a `path` parameter.
 
 ### `app.getFileIcon(path[, options])`
 
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -582,7 +582,7 @@ them.
 
 Sets or creates a directory your app's logs which can then be manipulated with `app.getPath()` or `app.setPath(pathName, newPath)`.
 
-Calling `app.setAppLogsPath()` without a `path` parameter will result in this directory being set to `/Library/Logs/YourAppName` on _macOS_, and inside the `userData` directory on _Linux_ and _Windows_.
+Calling `app.setAppLogsPath()` without a `path` parameter will result in this directory being set to `~/Library/Logs/YourAppName` on _macOS_, and inside the `userData` directory on _Linux_ and _Windows_.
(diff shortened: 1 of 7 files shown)


shell/browser/api/atom_api_app.cc
```diff
@@ -868,18 +868,21 @@ void App::SetAppLogsPath(mate::Arguments* args) {
...
-  if (!succeed) {
-    if (name == "logs") {
-      args->ThrowError("Failed to get '" + name +
-                       "' path: setAppLogsPath() must be called first.");
-    } else {
-      args->ThrowError("Failed to get '" + name + "' path");
+    // If users try to get the logs path before setting a logs path,
+    // set the path to a sensible default and then try to get it again
+    if (name == "logs" && !succeed) {
+      SetAppLogsPath(args);
```
ckerr: This is probably safe, but it's a little weird to pass `args` here: (1) we actually _won't_ get our expected behavior if it happens to have anything in it, and (2) SetAppLogsPath() calls GetNext() and so the args->next_ field gets incremented.

It would be slightly safer to pass the address of a local temporary Args

shell/browser/api/atom_api_app.cc
```diff
@@ -868,18 +868,21 @@ void App::SetAppLogsPath(mate::Arguments* args) {
...
     succeed = base::PathService::Get(key, &path);
-  if (!succeed) {
-    if (name == "logs") {
-      args->ThrowError("Failed to get '" + name +
-                       "' path: setAppLogsPath() must be called first.");
-    } else {
-      args->ThrowError("Failed to get '" + name + "' path");
+    // If users try to get the logs path before setting a logs path,
+    // set the path to a sensible default and then try to get it again
+    if (name == "logs" && !succeed) {
```
ckerr:
```suggestion
    if (!succeed && name == "logs") {
```

micro-nit: evaluating a bool is cheaper than comparing a string, so try it first

docs/api/app.md
```diff
@@ -614,6 +614,8 @@ Returns `String` - The current application directory.
 Returns `String` - A path to a special directory or file associated with `name`. On
 failure, an `Error` is thrown.
 
+Calling `app.getPath('logs')` without having previously called `app.setAppLogsPath()` will result in a directory being created and set to `/Library/Logs/YourAppName` on _macOS_, and inside the `userData` directory on _Linux_ and _Windows_.
```
ckerr:
```suggestion
If `app.getPath('logs')` is called without called `app.setAppLogsPath()` being called first, a default log directory will be used equivalent to calling `app.setAppLogsPath()` without a `path` parameter.
```

(point back to the description in `app.setAppLogsPath()` instead of re-describing it here)

ckerr: Unresolving due to lack of response + lack of change?

codebytere: whoops, thought this made it into the changes and did not. resolved.

docs/api/app.md
```diff
@@ -614,6 +614,8 @@ Returns `String` - The current application directory.
 Returns `String` - A path to a special directory or file associated with `name`. On
 failure, an `Error` is thrown.
 
+If `app.getPath('logs')` is called without called `app.setAppLogsPath()` being called first, a default log directory will be used equivalent to calling `app.setAppLogsPath()` without a `path` parameter.
```
MarshallOfSound:
```suggestion
If `app.getPath('logs')` is called without called `app.setAppLogsPath()` being called first, a default log directory will be created equivalent to calling `app.setAppLogsPath()` without a `path` parameter.
```

native_mate/native_mate/arguments.h
```diff
@@ -34,6 +35,17 @@ class Arguments {
     return ConvertFromV8(isolate_, info_->Data(), out);
   }
 
+  template <typename T>
+  bool GetNext(base::Optional<T>* out) {
+    if (next_ >= info_->Length())
+      return false;
```
MarshallOfSound: This has to be `true` @codebytere 

If there is no next but the type is optional then we should just give it an empty optional

MarshallOfSound: (and mark the `GetNext` as succeeded)

MarshallOfSound: That's the difference between this `GetNext` and the one below

codebytere: done!

codebytere:
@MarshallOfSound should be all set now!

shell/common/native_mate_converters/value_converter.h
```diff
@@ -33,6 +35,29 @@ struct Converter<base::Value> {
...
+template <typename T>
+struct Converter<base::Optional<T>> {
+  static bool FromV8(v8::Isolate* isolate,
+                     v8::Local<v8::Value> val,
+                     base::Optional<T>* out) {
+    if (val->IsNull() || val->IsUndefined()) {
+      return true;
+    }
+    T converted;
+    if (Converter<T>::FromV8(isolate, val, &converted)) {
```
zcbenz: @codebytere I just ran into this code when doing refactoring. I believe this should be `if not` otherwise the `out` would never be filled?

PR #19656, opened 2019-08-06
feat: add new nativeTheme API

MarshallOfSound:
This adds a new `nativeTheme` API which extracts APIs from `systemPreferences` (which was becoming a bit of a dumping ground) and consolidates them all to be backed by Chromium's `NativeTheme` module which is better implemented, more resilient and means we don't have to maintain the theme code 👍 

It also allows us to easily implement the `updated` event.

Notes: Added new `nativeTheme` API to read and respond to changes in the OS's theme and color scheme


DeerMichel:
Now we're talking 🎉

lib/browser/api/system-preferences.js
```diff
@@ -12,4 +12,8 @@ if ('appLevelAppearance' in systemPreferences) {
   deprecate.fnToProperty(SystemPreferences.prototype, 'appLevelAppearance', '_getAppLevelAppearance', '_setAppLevelAppearance')
 }
 
+if ('effectiveAppearance' in systemPreferences) {
+  deprecate.fnToProperty(SystemPreferences.prototype, 'effectiveAppearance', '_getEffectiveAppearance')
+}
+
```
DeerMichel: Can we also add deprecation warnings for the 3 methods moved to `nativeTheme`?

shell/common/api/atom_api_native_theme.cc
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2019 Slack Technologies, Inc.
```
DeerMichel: Hehe, gotcha 😉

shell/common/api/atom_api_native_theme.h
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) 2019 Slack Technologies, Inc.
+// Use of this source code is governed by the MIT license that can be
+// found in the LICENSE file.
+
+#ifndef SHELL_BROWSER_API_ATOM_API_NATIVE_THEME_H_
```
DeerMichel:
```suggestion
#ifndef SHELL_COMMON_API_ATOM_API_NATIVE_THEME_H_
```

shell/common/api/atom_api_native_theme.h
```diff
@@ -0,0 +1,44 @@
...
+#define SHELL_BROWSER_API_ATOM_API_NATIVE_THEME_H_
```
DeerMichel:
```suggestion
#define SHELL_COMMON_API_ATOM_API_NATIVE_THEME_H_
```

shell/common/api/atom_api_native_theme.cc
```diff
@@ -0,0 +1,77 @@
...
+
+// static
+void NativeTheme::BuildPrototype(v8::Isolate* isolate,
+                                 v8::Local<v8::FunctionTemplate> prototype) {
+  prototype->SetClassName(mate::StringToV8(isolate, "NativeTheme"));
+  mate::ObjectTemplateBuilder(isolate, prototype->PrototypeTemplate())
+      .SetProperty("shouldUseDarkColors", &NativeTheme::ShouldUseDarkColors)
+      .SetProperty("shouldUseHighContrastColors",
+                   &NativeTheme::ShouldUseHighContrastColors)
+      .SetProperty("shouldUseInvertedColorScheme",
```
DeerMichel: In favor of consistency, imo there should be `NativeWindow::ShouldUseInvertedColorScheme` as well

shell/common/api/atom_api_native_theme.h
```diff
@@ -0,0 +1,44 @@
...
+  ui::NativeTheme* theme_;
+
+  DISALLOW_COPY_AND_ASSIGN(NativeTheme);
+};
+
+}  // namespace api
+
+}  // namespace electron
+
+#endif  // SHELL_BROWSER_API_ATOM_API_NATIVE_THEME_H_
```
DeerMichel:
```suggestion
#endif  // SHELL_COMMON_API_ATOM_API_NATIVE_THEME_H_
```

shell/common/api/atom_api_native_theme.cc
MarshallOfSound: There is no need for this (it would just be an empty method that called another method, more efficient to bind directly to the target)

MarshallOfSound:
> Did you miss module-keys.js or isn't it used anymore?

`module-keys.js` is only used for browser APIs.  These APIs are in `common`.

> nativeTheme.shouldUseInvertedColorScheme doesn't work on macOS for me (always false), but update event gets fired when I toggle it

This API is only implemented on Windows inside Chromium, it's a one liner to implement on macOS so I'll do that here as well.

> Afaik, the nativeTheme methods on Linux aren't implemented and default to some value - we should state that

Potentially, will have to review which APIs are implemented for linux 👍 

> Oh and we have some fake tests for the sysPref methods - do we wanna move them to a new nativeThemeSpec?

New tests should be added here that the APIs "work"

docs/api/native-theme.md
```diff
@@ -0,0 +1,35 @@
...
+
+A `Boolean` for if the OS / Chromium currently has a dark mode enabled or is
+being instructed to show a dark-style UI.
+
+### `nativeTheme.shouldUseHighContrastColors` _Readonly_
+
+A `Boolean` for if the OS / Chromium currently has high-contrast mode enabled
+or is being instructed to show a high-constrast UI.
+
+## `nativeTheme.shouldUseInvertedColorScheme` _Readonly_
```
DeerMichel:
```suggestion
### `nativeTheme.shouldUseInvertedColorScheme` _Readonly_
```

docs/api/native-theme.md
```diff
@@ -0,0 +1,35 @@
+# nativeTheme
+
+> Read and respond to changes in Chromium's native color theme
```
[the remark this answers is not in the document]
MarshallOfSound: `nativeTheme` lines up with our `nativeImage` API hence the naming.  IMO it is worth highlighting that this is based off the the theme that the internal Chromium APIs use rather than the OS theme as although they are mostly the same their are differences that could come into play (E.g. the `--force-dark-mode` flag)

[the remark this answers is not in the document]
DeerMichel: I'm ok with this - although i still think `appearance` is a more intuitive name. (Apache vote: -0).

shell/common/api/atom_api_native_theme.cc
```diff
@@ -0,0 +1,97 @@
...
+const CFStringRef WhiteOnBlack = CFSTR("whiteOnBlack");
+const CFStringRef UniversalAccessDomain = CFSTR("com.apple.universalaccess");
+#endif
+
+// TODO(MarshallOfSound): Implement for Linux
+bool NativeTheme::ShouldUseInvertedColorScheme() {
+#if defined(OS_MACOSX)
+  CFPreferencesAppSynchronize(UniversalAccessDomain);
+  Boolean keyExistsAndHasValidFormat = false;
+  Boolean is_inverted = CFPreferencesGetAppBooleanValue(
```
codebytere: I think we can use type coercion here to clean things up a bit:
```cpp
bool keyExistsAndHasValidFormat = false;
bool is_inverted = CFPreferencesGetAppBooleanValue(
                WhiteOnBlack, UniversalAccessDomain, &keyExistsAndHasValidFormat);
return keyExistsAndHasValidFormat && is_inverted;
```

lib/common/api/deprecate.ts
```diff
@@ -51,7 +51,15 @@ const deprecate: ElectronInternal.DeprecationUtil = {
     const warn = warnOnce(`${fn.name} function`, `${newName} function`)
     return function (this: any) {
       warn()
-      fn.apply(this, arguments)
+      return fn.apply(this, arguments)
+    }
+  },
+
+  moveAPI: (fn: Function, oldUsage: string, newUsage: string) => {
```
codebytere: Can we add a test for this?

damienallen:
Just updated today and didn't even know this was coming, awesome 🙌

PR #19672, opened 2019-08-07
feat: add property customization to save dialogs

codebytere:
#### Description of Change

This PR allows property customization on save dialogs in addition to what already exists for open dialogs. More specifically, this introduces the following properties on save dialogs:

* `showHiddenFiles` - Show hidden files in dialog.
* `createDirectory` _macOS_ - Allow creating new directories from dialog.
* `treatPackageAsDirectory` _macOS_ - Treat packages, such as `.app` folders,
      as a directory instead of a file.
* `showOverwriteConfirmation` _Linux_ - Sets whether the user will be presented a confirmation dialog if the user types a file name that already exists.
* `dontAddToRecent` _Windows_ - Do not add the item being saved to the recent documents list.

In order to accomplish this, i separated the dialog property enums into those for `SAVE` and `OPEN` so that it is more clear which properties apply to which dialog subtypes.

cc @ckerr @miniak @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added property customization to save dialogs.

### DIFF
--- a/docs/api/dialog.md
+++ b/docs/api/dialog.md
@@ -171,6 +171,13 @@ dialog.showOpenDialog(mainWindow, {
     displayed in front of the filename text field.
   * `showsTagField` Boolean (optional) _macOS_ - Show the tags input box,
     defaults to `true`.
+  * `properties` String[] (optional)
+    * `showHiddenFiles` - Show hidden files in dialog.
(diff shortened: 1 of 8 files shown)


codebytere:
Removed `window.print()` test per offline conversation with @deepak1556 👇

PR #19677, opened 2019-08-07
fix: crash on window.print()

codebytere:
Backport of https://github.com/electron/electron/pull/19601.

See that PR for more details.


misbeliever:
Checked v5 v6 and v7 after fix releases, because I need to specify the printer as device name, result only v7 works as intended with deviceName name specified

codebytere:
Yes @misbeliever, i have not yet been able to backport this fix to v6/v5; it's a Chromium patch so it's a bit more complex unfortunately. I'll do my best to get to it soon.

PR #19678, opened 2019-08-07
fix: crash on window.print()

codebytere:
Backport of https://github.com/electron/electron/pull/19601.

See that PR for more details.


misbeliever:
When the device name specified as parameter, still doesn't work for me.
Checked v5, v6 and v7 after window.print() release, only v7 works with deviceName specified

Issue #19602, opened 2019-08-03
network serviceification followup


nornagon:
Since #19488, the network service is enabled and the old path has been deleted. Some features are broken as a result and will need to be fixed before 8.0.0. This is a tracking issue for those features.

- [x] webRequest API (zcbenz)
- [x] service workers on file protocol (#20132)
- [x] session.clearAuthCache - tests (#20015)
- [x] session.getBlobData (#20041)
- [x] session.setUserAgent (#20014)
- [x] custom standard schemes / registerSchemesAsPrivileged
- [x] Rewire Login Handler from NetworkDelegate into ProxyingURLLoaderFactory
- [x] fix 'login' event not firing on net requests

cc @deepak1556 @jkleinsc @zcbenz

zcbenz:
I'm putting my name on the tasks that I'm working on, in case of any duplicate work.

PR #19679, opened 2019-08-08
feat: migrate webRequest module to NetworkService (Part 4)

zcbenz:
#### Description of Change

Adds empty stubs for the webRequest APIs.

I plan to use `gin` directly since this class is rewritten from scratch, so there are also gin related changes.

#### Release Notes

Notes: no-notes


shell/browser/api/atom_api_web_request_ns.h
```diff
@@ -5,27 +5,59 @@
...
-#include "shell/browser/api/trackable_object.h"
 
 namespace electron {
 
 class AtomBrowserContext;
 
 namespace api {
 
-class WebRequestNS : public mate::TrackableObject<WebRequestNS> {
+class WebRequestNS : public gin::Wrappable<WebRequestNS> {
```
deepak1556: will `gin::Wrappable` lifetime be made compatible with `mate::TrackableObject` ? Do we plan to eventually remove the `TrackableObject` ?

zcbenz: For `WebRequest`, it does not really need to be a `TrackableObject` as it is managed by `Session` and it does not hold native resources.

Previously it held a reference to `BrowserContext` and thus had to be a `TrackableObject`, I think it was a design error.

`TrackableObject` is still needed for cases that need to clean up resources on shutdown.

DeerMichel:
**This** is awesome 🎉

Issue #17523, opened 2019-03-23
Please, make printing work with Electron !


popod:
### Problem Description

There are many printing problems since Electron 3 and this prevents people who need a working print solution to use Electron 3, 4 or 5 !

So please, can you take a look at all this **printings issues** and trying to fix it ? We are a lot of developers waiting for this since a long time..

Thanks a lot for your work ! And I hope you could give us some feedback about this.

cc: @codebytere @MarshallOfSound @nornagon 

### Related issues

All this issues seems related to `webContents.print()` which seems to be buggy:

- ~~#17265 `webContents.print()` callback always return `false`~~ :heavy_check_mark:
  - **Fixed since v5.0.5**
- ~~#16085 Fix callback for `webContents.print()` cancel event~~ :heavy_check_mark:
  - **Fixed since v5.0.5**
- ~~#18123 `'silent: true'` for printing not working in version 5~~
  - ~~#18858 `webContents.print()` doesn't respond when adding parameters~~
  - ~~#17670 Don't ask user for print settings.~~
  - ~~#16362 Silent print not working in mac/osX~~
  - **Fixed since v5.0.7**
- ~~#18980 Better callback value for `webContents.print()`~~
  - **Available since v7**
- ~~#18917 print blurring~~
  - **Fixed since v7** (by #18984 feat: add a series of new printing options)
- ~~`webContents.print()` doesn't work when `deviceName` is provided and `silent: true`~~
   - #17077, #17260, #18123,  #17077 

Other printing issues:

- #16813 Unable to print large documents in Electron 4, 5
- ~~#16534 `webview.print()` not working properly in 4.0.X~~
  - **Seems to be fixed since v5.0.5**
- #15847 `webContents.print()` random missing letters
- ~~#7849 Printing needs more options~~
  - ~~#15568 Add error information to `webContents.print()` callback~~
  - **Available since v7**
- ~#19056 Add an option to choose the paper format (size) on `webContents.print()`~
- #18093 `webContents.printToPDF()` regression in v5
- ~~#19108 Electron crashed after print~~
   - **Fixed by https://github.com/electron/electron/pull/19598**
   - Will be released in `v6.0.2` and `v5.0.10`

(Non-exhaustive list)

mojason-81:
[...]

PR #19688, opened 2019-08-08
feat: allow customization of print page header and footer

codebytere:
#### Description of Change

Refs: https://github.com/electron/electron/issues/17523

This PR allows users to customize print settings further by setting a custom header and footer by adding the following to print setting options:

```
  * `headerFooterInfo` Object (optional)
    * `header` String - String to be printed as page header.
    * `footer` String - String to be printed as page footer.
```

Tested to work with the following:

```js
win.webContents.print({
   copies: 2,
   headerFooterInfo: {
      header: 'ALL HAIL BLINI CAT',
      footer: 'https://imgur.com/gallery/jQecn'
   }
}, (success, reason) => {
})
```

and an html page containing one link to [this image](https://user-images.githubusercontent.com/2036040/62734356-b493b200-b9dd-11e9-9471-4246a00dee81.jpg)

cc @deepak1556 @brenca @jkleinsc 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added options to enable customization of print page headers and footers.

### DIFF
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -1264,8 +1264,8 @@ Returns [`PrinterInfo[]`](structures/printer-info.md).
   * `dpi` Object (optional)
     * `horizontal` Number (optional) - The horizontal dpi.
     * `vertical` Number (optional) - The vertical dpi.
-  * `header` String - String to be printed as page header.
-  * `footer` String - String to be printed as page footer.
+  * `header` String (optional) - String to be printed as page header.
+  * `footer` String (optional) - String to be printed as page footer.
 * `callback` Function (optional)
   * `success` Boolean - Indicates success of the print call.
   * `failureReason` String - Called back if the print fails; can be `cancelled` or `failed`.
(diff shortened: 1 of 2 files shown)


docs/api/web-contents.md
```diff
@@ -1264,6 +1264,9 @@ Returns [`PrinterInfo[]`](structures/printer-info.md).
   * `dpi` Object (optional)
     * `horizontal` Number (optional) - The horizontal dpi.
     * `vertical` Number (optional) - The vertical dpi.
+  * `headerFooterInfo` Object (optional)
```
jkleinsc: Why is an object needed here?  I think it would be more straightforward as:
```suggestion
  * `header` String - String to be printed as page header.
  * `footer` String - String to be printed as page footer.
```

codebytere: This was done to simplify logic: header and footer can only be set if `printing::kSettingHeaderFooterEnabled` is set to `true`; and the methodology for this involving the least conditional checking was implicitly checking for the object itself. If an object was passed then both header _and_ footer must be set, and so we can proceed slightly more cleanly :)

jkleinsc: @codebytere maybe I missed this, but what if you only want to set header and not footer or vice versa?

codebytere: hmm, i think i might be able to rework the logic a bit to pass empty strings, although i think Chromium intends for both to be set if any are set 🤔

codebytere: @jkleinsc all set ⭐️

Issue #19687, opened 2019-08-08
`window.print()` not working in electron 7.0.0-beta.2


lesleyandreza:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.0.0-beta-2
* **Operating System:**
  * MacOs Mojave 10.14.6
* **Last Known Working Electron version:**
  * 7.0.0-beta.1

### Expected Behavior
Show system print dialog.

### Actual Behavior
Nothing happens.

### To Reproduce
Add `<script>window.print();</script>` into `index.html` in [*Electron Fiddle*](https://electronjs.org/fiddle).

codebytere:
I know what did this, PR incoming.

PR #19690, opened 2019-08-08
fix: crash in window.print()

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/19687.

`window.print()` also calls `InitPrintSettings` and would not receive the settings received when we call `webContents.print()` so we need to send `PrintHostMsg_GetDefaultPrintSettings` over IPC in that case.

cc @jkleinsc @deepak1556 

#### Checklist

- [x] `npm test` passes
- [x] specs are added

#### Release Notes


codebytere:
@deepak1556 the test itself passes but the native printer dialog pops when `window.print()` gets called and blocks so the tests time out every time. How would you suggest we proceed here?

deepak1556:
Can you test with silent code path for `window.print` ?

Issue #8124, opened 2016-12-01
Cache data is written to `userData` dir instead of `cache` dir


smyrman:
* Electron version: 1.3.9 (via Atom 1.12.6) and unknown (via Slack 2.3.2)
* Operating system: Linux

### Expected behavior

Files stored by Electron on behalf of electron apps, should follow the XDG Base Directory specification and place only config files in $XDG_CONFIG_HOME/<App name> (default ~/.config/). cache and data files should be placed in XDG_CACHE_HOME (default: ~/.cache) and XDG_DATA_HOME (default: ~/.local/share) respectively.

https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html

### Actual behavior

As seen in Atom and Slack on Linux, they leave non-config files in ~/.config/Atom and ~/.config/Slack/ in particular, they leave folders such as "databases", "GPUCache", "Cache", "logs", and "Local Storage". Atom mention this as an Electron issue in https://github.com/atom/atom/issues/8281 ("where electron stores it's files"). 

Could still be it's not Electron's fault that the apps store non-config files in these locations, but hopefully you know more than me here.

kevinsawicki:
> Files stored by Electron on behalf of electron apps

Chrome provides the paths to these files, it looks like they've labeled issue as a won't fix in the past.

https://bugs.chromium.org/p/chromium/issues/detail?id=129861

smyrman:
It seams to me though that chromium still uses XDG_CACHE_HOME correctly? Although they do store config and browser data together.

Is it possible for electron to move fodlers like GPUCache Cache to XDG_CACHE_HOME, or are those still defined by Chrome/chromium?

kevinsawicki:
> Is it possible for Electron to move folders like GPUCache and Cache to XDG_CACHE_HOME, or are those still defined by Chrome/chromium?

Those files are completely managed by Chrome I believe, are you seeing standalone Chrome store them in a different location?

smyrman:
I find ~/.cache/google-chrome/Default/
Cache  'Media Cache'
and:
~/.config/google-chrome/Default/GPUCache

, so actually, it just seams inconsistent...

Thanks for the quick reply!

deepak1556:
> Is it possible for Electron to move folders like GPUCache and Cache to XDG_CACHE_HOME, or are those still defined by Chrome/chromium?

Yup its possible. Currently we set the cache path to app data path in brightray. @kevinsawicki are we ok on changing this ?

[...]

PR #19693, opened 2019-08-09
feat: Custom cache path

emmkimme:
#### Description of Change
This MR fixes the following issues:
- https://github.com/electron/electron/issues/18309.
- https://github.com/desktop/desktop/issues/6523.

Notes: Allow to customize Electron/Chromium cache location

#### Checklist
- [x] `npm test` passes
- [x] tests are unchanged 

#### Technical Notes
The notion of cache (DIR_CACHE, DIR_USER_CACHE) is already present in the code but not used.

atom\browser\api\atom_api_app.cc
```
int GetPathConstant(const std::string& name) {
...
  else if (name == "cache")
    return DIR_CACHE;
  else if (name == "userCache")
    return DIR_USER_CACHE;
...
```

atom/browser/atom_paths.h
```
...
  DIR_USER_CACHE,              // Directory where user cache can be written.
#if defined(OS_POSIX)
  DIR_CACHE = base::DIR_CACHE,  // Directory where to put cache data.
#else
  DIR_CACHE = base::DIR_APP_DATA,
#endif
...
```
lib\browser\api\app.ts
```
app._setDefaultAppPaths = (packagePath) => {
...
  app.setPath('userData', path.join(app.getPath('appData'), app.name!))
  app.setPath('userCache', path.join(app.getPath('cache'), app.name!))
```

There is 2 levels of cache : 
- userCache (DIR_USER_CACHE) and cache (DIR_CACHE) 
equivalent to the 2 levels of user data
- userData (DIR_USER_DATA) and appData (DIR_APP_DATA) 

Purpose was to use the userCache value for customizing the cache

atom\browser\atom_browser_context.cc
Before
```
if (!base::PathService::Get(DIR_USER_DATA, &path_)) {
    base::PathService::Get(DIR_APP_DATA, &path_);
    path_ = path_.Append(base::FilePath::FromUTF8Unsafe(GetApplicationName()));
    base::PathService::Override(DIR_USER_DATA, path_);
  }
```

After
```
  base::PathService::Get(DIR_USER_CACHE, &path_);
```
I added a AppPathService which provides default paths (data, cache, logs) if there are not overridden.

#### Usage
```
import { app } from 'electron';

// Must be called before application is ready
app.setPath('userCache', myOwnUserCache);
```
#### Backward compatibility
As before, by default, 
- 'userData' is computed from 'appData' and app name
- 'userCache' from 'cache' and app name

But not at the same moment. previously it was done very early, before app main script was executed. Now it is done when the app is ready, giving some opportunities to customize 'appData', app name, ..
Previous behavior can be restored easily.

#### Unit tests
Unit tests are now covering 'userData', 'userCache' and 'logs'.

#### Additional notes
For concurrent access violation, the cache folder must not be shared by different users.
For performance purpose, it must be set on a fast drive.


emmkimme:
Hi, there is WIP (in progress) and 'build-mac' (failing) both stuck on a old commit. 
It is supposed to be fixed. 
is there a way to force checks ?

emmkimme:
Dear @codebytere or any reviewers
My MR is ready but there is still tests not related to my code which are failing  !
If you do not mind to plan a review please
Thanks

codebytere:
@emmkimme it looks like https://github.com/electron/electron/pull/19693/files#diff-c0db39733aba3af6dedd3c8a42082f92 accidentally got changed more than intended?

emmkimme:
> @emmkimme it looks like https://github.com/electron/electron/pull/19693/files#diff-c0db39733aba3af6dedd3c8a42082f92 accidentally got changed more than intended?

You right, I have been again trapped with CR/LF mess on windows. I do not get why it happens only on Markdown files, it works for all others. I will fix that. Thanks

emmkimme:
I deactivated my tests related to setAppLogsPath (BTW original tests are wrong), as it does not work on Windows. 

If I change the current implem C++ signature from
```
void App::SetAppLogsPath(util::ErrorThrower thrower,
                         base::Optional<base::FilePath> custom_path) {
```
to
```
void App::SetAppLogsPath(util::ErrorThrower thrower,
                         const base::FilePath& custom_path) {
```
Removing `base::Optional` works !

Another point, AppLogs on Mac takes for application name
```
NSString* bundle_name =
      [[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleName"];
```
All others code takes 
Browser::Get()->GetName()
is it expected ?

emmkimme:
Dear @codebytere, @poiru, @ckerr, reviewers
Do you know when it is possible to plan a review of this MR please
Thanks

emmkimme:
Hi @MarshallOfSound, 
do you know when it can be possible to schedule a review of this please.
Thanks

emmkimme:
Could it be possible to review this MR please, I keep it update against master since August 2019 but it is a pain. Thanks.

shell/browser/atom_paths.cc
```diff
@@ -0,0 +1,149 @@
...
+                                     base::nix::kDotConfigDir);
+}
+#endif
+
+}  // namespace
+
+// We can not use PathService with a provider !
+// PathService caches the returned value of the first request
+// If the path depends on another values that changes later
+// PathService returns the cached value not the updated value
```
ckerr: I _really_ don't like this -- this sets us up to work against base::PathService instead of with it, and over time that's likely to be a maintenance headache. In addition, it makes it more complicated to understand which path gets picked for any given key because of the extra layers of choices.

I've made a basic proof-of-concept branch at https://github.com/electron/electron/commits/custom-cache-path which adds an AtomPathProvider that gets registered. The trick is that it relies on a modified `base/path-service.cc`:

```diff
diff --git a/base/path_service.cc b/base/path_service.cc
index cc29cab2f1ff..b27f63eeaec1 100644
--- a/base/path_service.cc
+++ b/base/path_service.cc
@@ -336,4 +336,13 @@ void PathService::DisableCache() {
   path_data->cache_disabled = true;
 }
 
+// static
+void PathService::InvalidateCache() {
+  PathData* path_data = GetPathData();
+  DCHECK(path_data);
+
+  AutoLock scoped_lock(path_data->lock);
+  path_data->cache.clear();
+}
+
 }  // namespace base
diff --git a/base/path_service.h b/base/path_service.h
index 9b4715f073c0..3551e8c150f7 100644
--- a/base/path_service.h
+++ b/base/path_service.h
@@ -79,6 +79,9 @@ class BASE_EXPORT PathService {
   // Disable internal cache.
   static void DisableCache();
 
+  // Invalidate internal cache.
+  static void InvalidateCache();
+
  private:
   friend class ScopedPathOverride;
   FRIEND_TEST_ALL_PREFIXES(PathServiceTest, RemoveOverride);
```

but IMO that's a patch that upstream might accept (preferrable); and in the case that they didn't, we could patch it locally (ok).

Still, as a proof-of-concept, https://github.com/electron/electron/commit/6f150ca4f50eb188b82f5ac194fac03163947c8d is pretty straightforward and it does pass your (very nice) new tests for the app path module:

```
$ e test --runners=main -g 'app path module' 
Running "node ./script/spec-runner.js --runners main -g app path module" in /home/charles/electron/gn/master/src/electron
Only running: [ 'main' ]

Running: Main process specs
[15937:1104/211436.764817:INFO:content_main_runner_impl.cc(956)] Chrome is running in full browser mode.
[15979:1104/211437.164519:ERROR:sandbox_linux.cc(372)] InitializeSandbox() called with multiple threads in process gpu-process.
1..18
ok 1 app path module computes 'userData' from 'appData' and app name by default
ok 2 app path module computes 'userData' from 'appData' and app name app.name='myAppName'
ok 3 app path module computes 'userData' from 'appData' and app name setPath('appData', '/tmp/myappdata')
ok 4 app path module customizes 'userData' setPath('userData', '/tmp/myuserdata')
ok 5 app path module customizes 'userData' --user-data-dir='/tmp/myuserdata')
ok 6 app path module computes 'userCache' from 'cache', 'appData' and app name by default
ok 7 app path module computes 'userCache' from 'cache', 'appData' and app name app.name='myAppName'
ok 8 app path module computes 'userCache' from 'cache', 'appData' and app name setPath('appData', '/tmp/myappdata')
ok 9 app path module customizes 'userCache' setPath('userCache', '/tmp/myusercache')
ok 10 app path module getPath(name) returns paths that exist
ok 11 app path module getPath(name) throws an error when the name is invalid
ok 12 app path module getPath(name) returns the overridden path
ok 13 app path module setPath(name, path) does not create a new directory by default
ok 14 app path module getAppPath works for directories with package.json
ok 15 app path module getAppPath works for directories with index.js
ok 16 app path module getAppPath works for files without extension
ok 17 app path module getAppPath works for files
ok 18 app path module setAppLogsPath by default
# tests 18
# pass 18
# fail 0
[15937:1104/211453.697558:ERROR:process_posix.cc(333)] Unable to terminate process 15992: No such process (3)
[15937:1104/211453.697608:WARNING:internal_linux.cc(64)] Failed to read /proc/15992/stat
[15937:1104/211453.697700:WARNING:discardable_shared_memory_manager.cc(431)] Some MojoDiscardableSharedMemoryManagerImpls are still alive. They will be leaked.
✓ Electron main process tests passed.
```

IMO what should happen is this PR should be modified to be a provider + add an InvalidateCache patch to `electron/patches`, and we try to upstream it. I'd be much more comfortable with that than adding another slightly-incompatible layer on top of PathService.

docs/api/app.md
```diff
@@ -598,6 +598,8 @@ Returns `String` - The current application directory.
     * `~/Library/Application Support` on macOS
   * `userData` The directory for storing your app's configuration files, which by
     default it is the `appData` directory appended with your app's name.
+    * `userCache` The directory for storing your app's Electron cache files, which by
+    default it is the `appData` directory appended with your app's name.
```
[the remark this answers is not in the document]
emmkimme: You are right, except on Windows 'cache' is equivalent to 'appData', it would explain the current sentence.

shell/browser/api/atom_api_app.cc
```diff
@@ -848,43 +809,25 @@ void App::SetAppPath(const base::FilePath& app_path) {
...
   } else {
-    base::FilePath path;
-    if (base::PathService::Get(DIR_USER_DATA, &path)) {
-      path = path.Append(base::FilePath::FromUTF8Unsafe(GetApplicationName()));
-      path = path.Append(base::FilePath::FromUTF8Unsafe("logs"));
-      base::PathService::Override(DIR_APP_LOGS, path);
-    }
+    AppPathService::GetDefault(DIR_APP_LOGS, &logs_path);
   }
+  AppPathService::Override(DIR_APP_LOGS, logs_path);
```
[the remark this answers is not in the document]
emmkimme: According to the SetAppLogsPath behavior, providing an empty string should revert to the default value. Currently there is no way to come back to default value, I mean clear the PathService cache, so I override the value with the default.

shell/browser/atom_paths.cc
emmkimme: To be fair, I was not fond of this solution myself. 
I implemented a provider in my first implem but stuck by the cache not possible to be resetted. I was reluctant to patch Chromium / PathService . If it is not a concern, I'm happy to do this way.

shell/browser/atom_paths.cc
```diff
@@ -0,0 +1,149 @@
...
+
+bool AppPathService::Override(const std::string& name,
+                              const base::FilePath& path) {
+  return AppPathService::Override(GetPathConstant(name), path);
+}
+
+bool AppPathService::Override(int key, const base::FilePath& path) {
+  bool succeed = false;
+  if (key >= 0)
+    // Path must be  absolute and never created by default
```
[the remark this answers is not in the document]
emmkimme: We do not create any paths until the App is ready, so 'create' must be set to false. Paths are explicitly created before the app-ready event.

emmkimme:
Last master does not compile anymore on my machine. I have to investigate.

shell/browser/atom_paths.h
```diff
@@ -45,6 +51,13 @@ enum {
 
 static_assert(PATH_START < PATH_END, "invalid PATH boundaries");
 
+class AtomPaths {
+ public:
+  static void Register();
+
+  static bool GetDefault(int key, base::FilePath* path);
```
ckerr: This method doesn't need to be visible in the header; it could just as easily be implemented in an anonymous namespace in atom_paths.cc. That would also allow the forward declaration of base::FilePath to be dropped

ckerr: There's no special state being held in AtomPaths, so the `GetDefault()` method could easily be implemented in an anonymous namespace -- the only customer that needs to see it is the implementation of `Register()`. If implemented that way, then this public static method could be removed, as could the forward declaration of FilePath above

shell/browser/browser.cc
```diff
@@ -154,9 +160,21 @@ void Browser::WillFinishLaunching() {
 void Browser::DidFinishLaunching(base::DictionaryValue launch_info) {
   // Make sure the userData directory is created.
   base::ThreadRestrictions::ScopedAllowIO allow_io;
+
+  // 'userData' (DIR_USER_DATA) and 'userCache' (DIR_USER_CACHE)
+  // can be customized changing the 'appData' or the application name.
+  // At this stage, it is no more allowed, paths must be frozen.
+  // So if it is not yet done, we 'override' the values.
+  // It does not prevent to explicitely changing userData/userCache later.
+  // May be, would be fine to prevent such actions.
```
ckerr: Maybe something like

```
// 'userData' (`DIR_USER_DATA) and `userCache` (DIR_USER_CACHE) can
// be customized by the user up until the `ready` event is fired.
// Now that we've reached that point, it's time to freeze their values.
// FIXME: setPath() should reject changes to these paths after this point
```

shell/browser/atom_paths.cc
```diff
@@ -0,0 +1,78 @@
...
+void GetLinuxAppDataPath(base::FilePath* path) {
+  std::unique_ptr<base::Environment> env(base::Environment::Create());
+  *path = base::nix::GetXDGDirectory(env.get(), base::nix::kXdgConfigHomeEnvVar,
+                                     base::nix::kDotConfigDir);
+}
+#endif
+
+}  // namespace
+
+// We can not use PathService with a provider !
```
ckerr: I think this comment paragraph is out-of-date, referring to a previous version of the PR?

shell/browser/atom_paths.h
```diff
@@ -5,6 +5,8 @@
 #ifndef SHELL_BROWSER_ATOM_PATHS_H_
 #define SHELL_BROWSER_ATOM_PATHS_H_
 
+#include <string>
```
ckerr: I don't think this is needed here?

shell/browser/browser.cc
```diff
@@ -122,6 +122,12 @@ std::string Browser::GetName() const {
 
 void Browser::SetName(const std::string& name) {
   OverrideApplicationName(name);
+  if (!is_ready()) {
+    // DIR_USER_DATA, DIR_USER_CACHE and DIR_APP_LOGS depend
+    // on application name.
+    // We have to force re-computation
+    base::PathService::InvalidateCache();
+  }
```
ckerr: The intent and implementation is fine but that `if (!is_ready())` line is going to be confusing to future maintainers. IMO the comment should take it into account. What would you think of

```cc
// Some of the paths in `app.getPath()` are based on `name` and should
// update when `name` is changed before `app.ready` is fired. So if app
// isn't ready yet, invalidte the PathService cache to force a refresh.
if (!is_ready()) {
  base::PathService::InvalidateCache();
}
```

shell/browser/atom_paths.h
emmkimme: Will manage that

emmkimme:
Hi, any hope ?

emmkimme:
Dear @ckerr, @codebytere, hope you going well in the current circumstances.
Guess, it is not the best conditions but if you can have a look at this 8 months old MR, I would be grateful. Thanks

emmkimme:
This MR is in our production code for more than 1 year now, any hope to have it integrate in Electron ?
Thanks

ThomasMoutel:
Hi @ckerr / @wg-upgrades,
What is preventing this PR from being reviewed and integrated?

jkleinsc:
@emmkimme given the change here, this needs to be reviewed by @electron/wg-api.  Also it appears that the change to  docs/api/menu-item.md is unneccessary; lint and build failures need to also be resolved.

emmkimme:
Thanks for coming back to me.
To be fair, this MR is pending for more than 18 months and I stop rebasing it from long time ago as it is a pain. I will manage that in the coming weeks.

emmkimme:
@jkleinsc
Single check failure is not related with my MR (session module ses.cookies should survive an app restart for persistent partition - should survive an app restart for persistent partition)
Then could we review this MR please ?

jkleinsc (changes requested):
Test failures appear to be flakes unrelated to this PR.  However it appears that this PR is adding two git submodules that shouldn't be there:
Screen Shot 2021-01-20 at 11 27 50 AM

emmkimme:
Sorry for that, should be good know (finally!)

spec-main/api-app-path-spec.ts
```diff
@@ -0,0 +1,315 @@
...
+  //      expect(fs.existsSync(osLogPath)).to.be.false()
+  //    })
+
+  //    it('creates a new logs directory if one does not exist', () => {
+  //      expect(() => { app.getPath('logs') }).to.not.throw()
+
+  //      const osLogPath = (logsPaths as any)[process.platform]
+  //      expect(fs.existsSync(osLogPath)).to.be.true()
+  //    })
+  //  })
```
ckerr, line 294: These appear to be new tests but are commented out. Am I reading that right, and if so, what is your intent for these?

spec-main/api-app-path-spec.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  // console.log(`ouput=${output}`)
+  try {
+    const outputJSON = JSON.parse(output);
+    return outputJSON as Payload;
+  } catch (err) {
+    console.log(`Error ${err}) with ${output}`);
+    throw err;
+  }
+}
```
ckerr, line 315: This util is very similar to one in spec-main/api-app-spec.ts.

Maybe it should be extracted into spec-helpers.ts where the same code can be used by both?

spec-main/api-app-path-spec.ts
```diff
@@ -0,0 +1,315 @@
...
+    // app path module setAppLogsPath() setAppLogsPath(/tmp/mylogs) - setAppLogsPath(/tmp/mylogs)
+    // /home/builduser/project/src/electron/spec-main/api-app-path-spec.ts
+    // AssertionError: expected '/home/builduser/.config/app-custom-path/logs' to equal '/tmp/mylogs'
+
+    // if (process.platform === 'win32') {
+    //   it(`setAppLogsPath('${appLogsPath}')`, async () => {
+    //     const output = await runTestApp('app-custom-path', `-custom-applogs=${appLogsPath}`)
+    //     expect(output.appLogs).to.equal(appLogsPath)
+    //   })
+    // }
```
ckerr, line 256: These appear to be new tests but are commented out. Am I reading that right, and if so, what is your intent for these?

shell/browser/browser.cc
```diff
@@ -174,9 +182,24 @@ void Browser::WillFinishLaunching() {
 void Browser::DidFinishLaunching(base::DictionaryValue launch_info) {
   // Make sure the userData directory is created.
   base::ThreadRestrictions::ScopedAllowIO allow_io;
+
+  // 'userData' (`DIR_USER_DATA) and `userCache` (DIR_USER_CACHE) can
```
ckerr, line 186: _nit_ unclosed backtick in comment

shell/browser/browser.cc
```diff
@@ -174,9 +182,24 @@ void Browser::WillFinishLaunching() {
...
+  // be customized by the user up until the `ready` event is fired.
+  // Now that we've reached that point, it's time to freeze their values.
```
ckerr, line 188: This is a good explanation of this block :+1:

docs/api/app.md
```diff
@@ -720,9 +722,9 @@ In that case, the directory should be created with `fs.mkdirSync` or similar.
 
 You can only override paths of a `name` defined in `app.getPath`.
 
-By default, web pages' cookies and caches will be stored under the `userData`
+By default, web pages' cookies and caches will be stored under the `userCache`
 directory. If you want to change this location, you have to override the
-`userData` path before the `ready` event of the `app` module is emitted.
+`userCache` path before the `ready` event of the `app` module is emitted.
```
ckerr, line 727: @jkleinsc this arguably makes this `semver/major` in the case of people who were already customizing their cookie location with code that overrides `userData`?

patches/chromium/expose_a_method_for_invalidating_internal_cache.patch
```diff
@@ -0,0 +1,43 @@
...
+@@ -78,6 +78,9 @@ class BASE_EXPORT PathService {
+   // Disable internal cache.
+   static void DisableCache();
+ 
++  // Clear cache
++  static void InvalidateCache();
++
+  private:
+   friend class ScopedPathOverride;
+   FRIEND_TEST_ALL_PREFIXES(PathServiceTest, RemoveOverride);
```
ckerr, line 43: Just leaving a breadcrumb: we should try to upstream this patch. @emmkimme do you want to do that?

I'm still :+1: on this approach of having upstream have an "invalidate cache" API so that we can use this PathService code rather than reinventing the wheel with a custom PathService in Electron's code.

ckerr:
BTW @emmkimme you have really gone above & beyond with your persistence on this PR. :tada: :heart:

spec-main/api-app-path-spec.ts
emmkimme: In fact, there is another duplicate in api-ipc-main-spec.ts#51. If you do not mind, would prefer to focus on the main feature.

docs/api/app.md
emmkimme: If people are customizing 'userData' the behavior remains the same. 'userCache' is a new way to move the cache out of its default location in 'userData' (in my case on Windows, move it out of the roaming profile of the user)

shell/browser/browser.cc
emmkimme: Fixed

spec-main/api-app-path-spec.ts
emmkimme: intermediate  tests, not needed, I removed them

spec-main/api-app-path-spec.ts
emmkimme: intermediate tests, not needed, I removed them

shell/browser/browser.cc
emmkimme: Thanks

patches/chromium/expose_a_method_for_invalidating_internal_cache.patch
emmkimme: I never did it.
But to be fair, according to the time this MR is pending could we bypass this step please, would be so happy to have it in master as soon as possible.

patches/chromium/expose_a_method_for_invalidating_internal_cache.patch
```diff
@@ -0,0 +1,43 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: "emmanuel.kimmerlin@thomsonreuters.com"
+ <emmanuel.kimmerlin@thomsonreuters.com>
+Date: Mon, 18 Nov 2019 12:42:23 +0100
+Subject: Expose a method for invalidating internal cache
+
+Allow to customize Electron/Chromium cache location
+See MR: https://github.com/electron/electron/pull/19693
```
nornagon, line 8: Per https://www.electronjs.org/docs/development/patches#patch-justification, every patch we carry is a significant ongoing maintenance burden. We ask that the description of a patch includes a justification of why they're worth that maintenance burden, in addition to a description of the functionality that the patch adds.

In order of descending desirability:

1. Can this patch be removed entirely by achieving the desired functionality without having to include a patch?
2. Can this patch be merged upstream into Chromium itself? (see [here](https://chromium.googlesource.com/chromium/src/+/master/docs/contributing.md) for a guide on contributing to Chromium)
3. In the patch description, describe the reasons why neither of the above two are possible, and include guidance for a future maintainer looking to clean up patches about how they can check whether this patch is still required (maybe as simple as "if it compiles without this patch, the patch is no longer needed").

Thanks for helping us keep maintaining Electron sustainable 🙇

shell/browser/electron_browser_main_parts.cc
```diff
@@ -242,7 +231,6 @@ void ElectronBrowserMainParts::RegisterDestructionCallback(
 int ElectronBrowserMainParts::PreEarlyInitialization() {
   field_trial_list_ = std::make_unique<base::FieldTrialList>(nullptr);
 #if defined(OS_LINUX)
```
zcbenz, line 233: The `if defined` can also be removed.

zcbenz (changes requested):
Session data (like cookies and indexDB) should be considered as user data instead of cache, apps usually store login tokens and un-uploaded works in it. So if we silently change the place where session data are stored in Electron, users will lose login sessions and unsaved works after an app update, which is not acceptable.

It is totally Chromium's fault that it put session data and browser cache in one directory, and is impossible to separate them without a massive patch to Chromium, but I think we can use a different way other than `userCache` to make things less confusing.

I propose:
1. Do not list `userCache` in public API.
2. Add `browserData` path which is used to store cookies and cache, i.e. the "cache path" in this PR.
3. The default value of `browserData` should be the same with `userData`.

In this way we can achieve:
1. a clear API that does not give a wrong assumption of what "cache" means;
2. backward-compatibility without risk losing users' data;
2. future-proof that if in future it became possible to separate locations of cookies and caches, we could still introduce `userCache` path.

shell/browser/electron_browser_main_parts.cc
emmkimme: I'm afraid, there is some misunderstandings
1. while new userCache path is not modified, the location remains the same, it is **backward compatible**.
2. if as an Electron app developer I want to control where the whole cache is located (cookies, db, session, file cache..), this is under my responsibility. I do not get the point to split the cache into different pieces. From my point of view, it is less error-prone, simpler and more consistent, to consider all 'temporary' files as a whole. my2cents

emmkimme:
Hi @zcbenz,
I'm afraid, there is some misunderstandings
- while new userCache path is not modified, the location remains the same, it is backward compatible.
- if as an Electron app developer I want to control where the whole cache is located (cookies, db, session, file cache..), this is under my responsibility. I do not get the point to split the cache into different pieces. From my point of view, it is less error-prone, simpler and more consistent, to consider all 'temporary' files as a whole.
my2cents

shell/browser/electron_browser_main_parts.cc
emmkimme: done

zcbenz:
> while new userCache path is not modified, the location remains the same, it is backward compatible.

The `userCache` path points to `DIR_CACHE/AppName`:
https://github.com/electron/electron/blob/aef75682a3cadd12cfaa1ac977969b7275077429/shell/app/electron_default_paths.cc#L66-L70

I'm not sure for Windows, but for macOS and Linux the `DIR_CACHE` points to a different location than `DIR_APP_DATA`. So if we change `BrowserContext`'s path from `DIR_USER_DATA` to `DIR_USER_CACHE` it will not be backward compatible.

> if as an Electron app developer I want to control where the whole cache is located (cookies, db, session, file cache..), this is under my responsibility. I do not get the point to split the cache into different pieces. From my point of view, it is less error-prone, simpler and more consistent, to consider all 'temporary' files as a whole.

For desktop apps, cache is usually something that can be safely deleted without affecting current app's state, for example, for a chat app the login session should be kept after deleting the app's cache. This is different from the cache under browser's context, where cookies and sessions are also part of cache.

With the name of `userCache`, it is not clear whether things like cookies and localStorage belong to "cache", and according to the discussion in https://github.com/electron/electron/issues/8124, I believe lots of people would be surprised if cookies are treated as cache in this API.

So I'm proposing using `browserData` as the path's name instead, which will not cause confusions like `userCache`, and it will be clear that if this path is changed all cookies will be lost.

As for separating cookies from other cache, I agree it is error-prone and we should not do it unless it is supported by upstream Chromium.

emmkimme:
I'm still looking at a Mac machine for investigating. Should have one soon !

ratze90:
> I'm still looking for a Mac machine for investigating. Should have one soon !

Maybe i can help you? I will look through the failing checks.

emmkimme:
> > I'm still looking for a Mac machine for investigating. Should have one soon !
> 
> Maybe i can help you? I will look through the failing checks.

That would be great. Lot of thanks. Let me know if you need any help

emmkimme:
I did not rebase for 5 months and meanwhile the code change a lot, especially new DIR_* constant appeared and a new management of them. I'm afraid I have to redo this feature from scratch.

emmkimme:
Has been closed and replaced by this one: [Redirect Electron/Chromium cache location](https://github.com/electron/electron/pull/30665/files)

PR #19698, opened 2019-08-09
feat: add property to accessible BrowserWindow window title

mshoho:
#### Description of Change

Sometimes it's necessary to convey more information about the window to the screen reader users only (simply putting everything to the window title might be unnecessarily noisy).

For example, Chromium uses that technique to tell the screen reader users that the window is in incognito mode (the incognito window looks differently and doesn't have «incognito» in the title, but for the blind users the screen reader will announce that it's incognito).

Chromium has GetAccessibleWindowTitle() method internally. This PR exposes a way to customize the accessible title via Electron API.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `win.accessibleTitle property` to be able to augment the window title for the screen readers.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -826,7 +826,9 @@ Menu.setApplicationMenu(menu)
 
 #### `win.accessibleTitle`
 
-A `String` property that defines a part of window title which is visible for the screen readers only.
+A `String` property that defines an alternative title provided only to
+accessibility tools such as screen readers. This string is not directly
+visible to users.
 
 ### Instance Methods
 
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -379,8 +379,9 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
     * `disableHtmlFullscreenWindowResize` Boolean (optional) - Whether to
       prevent the window from resizing when entering HTML Fullscreen. Default
       is `false`.
-    * `accessibleTitle` String (optional) - a part of Window title visible for
-      the screen readers only.
+    * `accessibleTitle` String (optional) - An alternative title string provided only
+      to accessibility tools such as screen readers. This string is not directly
+      visible to users.
 
 When setting minimum or maximum window size with `minWidth`/`maxWidth`/
 `minHeight`/`maxHeight`, it only constrains the users. It won't prevent you from
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -1227,6 +1227,16 @@ Returns `String` - The title of the native window.
 **Note:** The title of the web page can be different from the title of the native
 window.
 
+#### `win.setAccessibleTitle(title)`
+
+* `title` String
+
+Changes the accessible title of native window to `title`.
+
+#### `win.getAccessibleTitle()`
+
+Returns `String` - The accessible title of native window (or an empty string if none).
+
 #### `win.setSheetOffset(offsetY[, offsetX])` _macOS_
 
 * `offsetY` Float
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -379,6 +379,8 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
     * `disableHtmlFullscreenWindowResize` Boolean (optional) - Whether to
       prevent the window from resizing when entering HTML Fullscreen. Default
       is `false`.
+    * `accessibleTitle` String (optional) - a part of Window title visible for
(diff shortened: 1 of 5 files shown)


alexeykuzmin:
@mshoho please use the PR template for description.

shell/browser/native_window.h
```diff
@@ -132,6 +132,12 @@ class NativeWindow : public base::SupportsUserData,
   virtual void Invalidate() = 0;
   virtual void SetTitle(const std::string& title) = 0;
   virtual std::string GetTitle() = 0;
+
+  // Ability to augment the window title for the screen readers.
+  base::string16 GetAccessibleWindowTitle() const override;
```
miniak: please move to the section with other `WidgetDelegate` overrides

shell/browser/native_window.cc
```diff
@@ -582,6 +582,22 @@ const views::Widget* NativeWindow::GetWidget() const {
   return widget();
 }
 
+base::string16 NativeWindow::GetAccessibleWindowTitle() const {
```
miniak: please move to the section with other `WidgetDelegate` overrides

docs/api/browser-window.md
```diff
@@ -1227,6 +1227,16 @@ Returns `String` - The title of the native window.
 **Note:** The title of the web page can be different from the title of the native
 window.
 
+#### `win.setAccessibleTitle(title)`
```
miniak: please add to `docs/api/modernization/property-updates.md`

miniak:
please use the correct PR template

docs/api/browser-window.md
codebytere: this is new, so it doesn't need to go there!

MarshallOfSound: That said this should definitely be a property `win.accessibleTitle`

shell/browser/api/atom_api_top_level_window.cc
```diff
@@ -1108,6 +1116,8 @@ void TopLevelWindow::BuildPrototype(v8::Isolate* isolate,
       .SetMethod("getPosition", &TopLevelWindow::GetPosition)
       .SetMethod("setTitle", &TopLevelWindow::SetTitle)
       .SetMethod("getTitle", &TopLevelWindow::GetTitle)
+      .SetMethod("setAccessibleTitle", &TopLevelWindow::SetAccessibleTitle)
```
MarshallOfSound: This should be a property `accessibleTitle` with a setter/getter 👍

docs/api/browser-window.md
```diff
@@ -379,6 +379,8 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
     * `disableHtmlFullscreenWindowResize` Boolean (optional) - Whether to
       prevent the window from resizing when entering HTML Fullscreen. Default
       is `false`.
+    * `accessibleTitle` String (optional) - a part of Window title visible for
```
miniak: I think it should be visible to instead of for

miniak:
@MarshallOfSound can you please review the current version?

shell/browser/native_window.cc
```diff
@@ -582,6 +582,22 @@ const views::Widget* NativeWindow::GetWidget() const {
...
+  if (accessible_title_.empty()) {
+    return views::WidgetDelegate::GetAccessibleWindowTitle();
+  }
+
+  return base::UTF8ToUTF16(accessible_title_);
```
MarshallOfSound: Is there a reason we can't store this as `base::string16`.  Storing as `std::string` just to convert to `string16` every time _something_ in chrome calls this getter seems a bit rough

docs/api/browser-window.md
MarshallOfSound: `String (optional) - An alternative title string provided only to accessibility tools such as screen readers.  This string is not directly visible to users.`

shell/browser/native_window.cc
mshoho: The only reason is that I believe I've seen something similar somewhere else. I've switched to storing it as `base::string16`.

shell/browser/native_window.cc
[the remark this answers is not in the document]
mshoho: It's a Chromium's method and Chromium gives the ability to override it. And Chromium calls it internally constructing the title for the screen readers.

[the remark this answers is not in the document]
mshoho: More precisely it's a method of Chromium's WidgetDelegate and I'm just overriding it.

Issue #4974, opened 2016-03-31
shell.moveItemToTrash does not work when Recycle Bin is disabled


50Wliu:
- Electron version: 0.36.8
- Operating system: Windows 10 x64

I have the Recycle Bin set up to immediately delete files.  This seems to cause `shell.moveItemToTrash()` to fail.  As this is a native Windows option (at least for Windows 7+), it should be supported.

![recycle-bin-immediately-delete](https://cloud.githubusercontent.com/assets/2766036/14189087/90a8576e-f759-11e5-96bb-1518472c6f47.png)

felixrieseberg:
Did a quick investigation: This won't be easy.

baconface:
Would ```fs.unlink``` or ```fs.unlinkSync``` work for you since you have no need to put it in the bin?

allistercsmith:
Hey guys,

Any progress on this issue yet? Just wanted to add that this is an issue for OS X too. I have almost no experience with objective-C coding, so unfortunately can't be much help coding/testing wise.

However, just browsing around the web a bit ([here](https://developer.apple.com/library/mac/releasenotes/Foundation/RN-FoundationOlderNotes/), just search for NSFileManager Trash APIs), I noticed that there is a NSFeatureUnsupportedError returned if trash isn't supported for the volume.

Perhaps we can make a change in the `platform_util_mac.mm` file to the following lines:

```
bool MoveItemToTrash(const base::FilePath& full_path) {
  NSString* path_string = base::SysUTF8ToNSString(full_path.value());
  BOOL status = [[NSFileManager defaultManager]
                trashItemAtURL:[NSURL fileURLWithPath:path_string]
                resultingItemURL:nil
                error:nil];
  
  // Check if error = NSFeatureUnsupportedError.
  // If so, make a call to removeItemAtPath instead to directly delete the file.
  // Otherwise, continue as normal.
  
  if (!path_string || !status)
    LOG(WARNING) << "NSWorkspace failed to move file " << full_path.value()
                 << " to trash";
  return status;
}
```

There may be a bit more work involved with error messages/logging, but is there anyone who could try this out?

slikts:
This is a rather crippling issue, but thankfully a workaround exists for Windows (see atom/tree-view#856).

codebytere:
closing due to abandonment.

jcj52436999:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
[...]

PR #19700, opened 2019-08-09
feat: macOS removal fallback when moveItemToTrash fails

codebytere:
#### Description of Change

Refs https://github.com/electron/electron/issues/4974.

Some macOS volumes may not support a Trash folder or it may be disabled, so these methods will report failure by returning NO or `nil` and an `NSError` with `NSFeatureUnsupportedError`.

This enables consumers to handle this by optionally passing a boolean `deleteOnFail` value to indicate whether or not the item should be unilaterally deleted if the `NSFileManager` fails to execute `trashItemAtURL`.

cc @miniak @ckerr @MarshallOfSound 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Enable macOS users to fallback to item removal when when `shell.moveItemToTrash` fails.

### DIFF
--- a/shell/common/api/atom_api_shell.cc
+++ b/shell/common/api/atom_api_shell.cc
@@ -71,6 +71,13 @@ v8::Local<v8::Promise> OpenExternal(const GURL& url, mate::Arguments* args) {
   return handle;
 }
 
+bool MoveItemToTrash(const base::FilePath& full_path, mate::Arguments* args) {
+  bool delete_on_fail = false;
+  args->GetNext(&delete_on_fail);
+
+  return platform_util::MoveItemToTrash(full_path, delete_on_fail);
+}
+
 #if defined(OS_WIN)
 bool WriteShortcutLink(const base::FilePath& shortcut_path,
                        mate::Arguments* args) {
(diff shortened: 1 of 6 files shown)


shell/common/platform_util.h
```diff
@@ -41,7 +42,7 @@ void OpenExternal(const GURL& url,
                   OpenExternalCallback callback);
 
 // Move a file to trash.
-bool MoveItemToTrash(const base::FilePath& full_path);
+bool MoveItemToTrash(const base::FilePath& full_path, mate::Arguments* args);
```
deepak1556: The function signature can be just `bool MoveItemToTrash(mate::Arguments* args)` and the `const base::FilePath& full_path` value can be still be checked for non-optionality with `args->Length()`.

MarshallOfSound: Putting the filrpath in the signature though ensures the error message for missing arguments is consistent across methods 

#justmy2cents

deepak1556: In that case the implementation declaration should not be made aware of `mate::Arguments`, I would suggest doing the runtime parsing in `shell/common/api/*.cc` and call the implementation methods with right values.

shell/common/api/atom_api_shell.cc
```diff
@@ -71,6 +71,13 @@ v8::Local<v8::Promise> OpenExternal(const GURL& url, mate::Arguments* args) {
   return handle;
 }
 
+bool MoveItemToTrash(const base::FilePath& full_path, mate::Arguments* args) {
```
deepak1556: `bool MoveItemToTrash(mate::Arguments* args)` , try to reuse args and perform the required processing

PR #19701, opened 2019-08-09
docs: mention QuickEdit in Windows troubleshooting

DeerMichel:
#### Description of Change
Stumbled across this "feature" while troubleshooting our Windows env 😏 

ref https://stackoverflow.com/questions/30418886/how-and-why-does-quickedit-mode-in-command-prompt-freeze-applications

cc @erickzhao @MarshallOfSound @nornagon 

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/docs/development/build-instructions-windows.md
+++ b/docs/development/build-instructions-windows.md
@@ -113,3 +113,10 @@ This can happen during build, when Debugging Tools for Windows has been installe
 ### ImportError: No module named win32file
 
 Make sure you have installed `pywin32` with `pip install pywin32`.
+
+### build scripts occasionally hang until keypress
+
+This bug is a "feature" of Windows' command prompt. It happens when clicking inside the prompt window with
+`QuickEdit` enabled and is intended to allow selecting and copying output text easily.
+Since each accidental click will pause the build process, you might want to disable this
+feature in the command prompt properties.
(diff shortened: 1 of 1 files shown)


docs/development/build-instructions-windows.md
```diff
@@ -113,3 +113,10 @@ This can happen during build, when Debugging Tools for Windows has been installe
 ### ImportError: No module named win32file
 
 Make sure you have installed `pywin32` with `pip install pywin32`.
+
+### build scripts occasionally hang until keypress
```
codebytere: nit: this should be :

```
### Keypress Troubleshooting

Build scripts occasionally hang until keypress. This bug is a "feature" of Windows' command prompt. It happens when clicking inside the prompt window with
`QuickEdit` enabled and is intended to allow selecting and copying output text easily.
Since each accidental click will pause the build process, you might want to disable this
feature in the command prompt properties.
```

DeerMichel: imo, if I searched for this issue, the original heading would feel more suitable to me

codebytere: that's why i moved the original text to the subheading: it'd still be greppable, but the header style now fits in with the style of all the rest of our documentation headers.

DeerMichel: how about `Building stuck until keypress`? - tbh none of them will fit the style bc all other headings are error messages.. and we don't get a message here 😄

DeerMichel: @codebytere so what's preferred?

codebytere: Let's go with `Build Script Hang on Keypress`?

Issue #18460, opened 2019-05-27
Missing additionalFeatures in new-window


zarubond:
Callback 'new-window' of BrowserWindow.webContents should pass additionalFeatures parameter as array of string which normally work until you enable on on parent window nativeWindowOpen feature, after that they parameter becomes always undefined. Native windows open is required for script-able child windows and not having additionalFeatures is bad because there is no other way how to pass window parameters from render to main process.

### Preflight Checklist

* [x ] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x ] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x ] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  3.0.0 and higher (probably older ones too)
* **Operating System:**
All platforms 
* **Last Known Working Electron version:**
None

### Expected Behavior
new-window callback should pass additionalFeatures as array of strings even if nativeWindowOpen is enabled.

### Actual Behavior
additionalFeatures is returned as undefiend if nativeWindowOpen is enabled.

### To Reproduce
Open up a BrowserWindow with nativeWindowOpen enabled and listening on 'new-window'. Then open up child window from render process with window.open('', 'child', 'testing=true');. The callback in 'new-window' will pass additionalFeatures as undefined instead of array of strings.

codebytere:
Looks like this is happening [here](https://github.com/electron/electron//blob/e81afed66d9e49fcca3679b663bd7c0319067c31/lib/browser/api/web-contents.js#L378-L399): the IPC message is being sent when the internal `'-add-new-contents'` event is emitted and it doesn't have a concept of `additionalFeatures` so it's undefined. 

I'll try to see what can be done about this.

lucasyvas:
It would be great if this could be made to work.

Issue #9340, opened 2017-05-01
postMessage between windows missing fields when contextIsolation enabled


kevinsawicki:
* Electron version: 1.6.6
* Operating system: All

### Expected behavior

`postMessage` over a `window.open` window or `window.opener` should contain all the standard event fields.

### Actual behavior

Event is missing `data`, `source`, and `origin` when `contextIsolation` is enabled. This is because the event is manually created and the secure proxies for built-ins such as `Event` prevent additional properties from hopping across contexts.

### How to reproduce

* Open a window with `contextIsolation` enabled
* Open a window from the page from the main context using `window.open`
* Post a message to the window, `event.data` will be missing, also other fields from https://github.com/electron/electron/blob/f84c75aa1aaa903d893a87a2111517ffdfec5cc1/lib/renderer/window-setup.js#L152-L154

shubham2892:
@kevinsawicki What in your opinion would be the best way to approach this.

kevinsawicki:
> What in your opinion would be the best way to approach this.

I'm not sure there will be a straight-forward way to address this based on the current implementation since the proxy objects used by `contextIsolation` deliberately prevent added properties from being accessible across contexts.

I think the best path forward it to make the `nativeWindowOpen` option as solid as can be and then eventually we can remove the custom `window.open`, `window.postMessage`, etc. implementation from `window-setup` and just rely on Chrome's support for this.

codebytere:
We are no longer implementing bugfixes for versions of Electron <= `1.7.x`, so i'm going to close this issue but if it is still persisting in more recent versions of Electron we can certainly reopen it!

asnow003:
Can we reopen this issue.  We are using the latest versions of Electron (2.0.10) and when we turn on context isolation the issue still exists.  Just wondering if this is something that will be fixed soon?  Looks like the suggested fix would be to remove the open/postMessage from the window-setup.js.

alexeykuzmin:
Reopening per the comment above.

alexeykuzmin:
@asnow003 Is using the [`nativeWindowOpen`](https://electronjs.org/docs/api/window-open#using-chromes-windowopen-implementation) an option for your task?

asnow003:
@alexeykuzmin I did try the nativeWindowOpen but still have the issue.  Also, with that switched on my child windows no longer will center on the screen (might be something fixable).

asnow003:
[...]

PR #19703, opened 2019-08-09
feat: enhance native window.open to match the custom implementation's behavior

### LINKED ISSUE #10852: Send postData with the "new-window" event for sandbox
We're writing a browser and we've been forced to ignore disposition because we're trying to create a new tab as a result of a form submission but don't have the postData in the `new-window` event to correctly handle the form post.

### LINKED ISSUE #18513: Can't focus or close windows when nativeWindowOpen is true
### Preflight Checklist

### Issue Details
* **Electron Version:** 4.0.4
* **Operating System:** Windows 10
* **Last Known Working Electron version:** n/a

### Expected Behavior
I open a native window with `w = window.open(...)`. `w.focus()` focuses the window. `w.close()` closes the window.

### Actual Behavior
`w.focus()` doesn't focus the window and `w.close()` doesn't close the window if it's cross-domain.

### To Reproduce
Seems related to [this issue](https://github.com/electron/electron/issues/8969), but I can reproduce it without sandboxing. Turning `nativeWindowOpen` off resolves the issue, so it appears related to that.

Download the electron quick start app and:

main.js:
```

function createWindow () {
  var mainWindow = new BrowserWindow({webPreferences: {nativeWindowOpen: true}});
  mainWindow.loadFile('index.html');
}

app.on('ready', createWindow)
```

index.html's body:
```
<webview id="w" src="about:blank;" allowpopups></webview>
<script>
   require('./renderer.js')
</script>
```

renderer.js:
```
setTimeout(function() {
	var w = document.getElementById("w");
	w.loadURL("https://google.com");
	w.openDevTools();
}, 1000); // after webview has loaded
```

After `npm start`ing the app, execute the following in the devtools:

    w = window.open("https://google.com", "a", "width=500");
    w.focus();

Observe that the window doesn't get focus.

Similarly, `w.close()` doesn't close the window if it points to a cross-domain site, but it sets `closed` to `true`.

Turning off `nativeWindowOpen` makes it work.

Tried in Chrome, and it works there.

Electron 5 and 6 also appear to be broken, while Canary (which uses the Chromium that Electron 6 uses I believe) works.

Same issue when using BrowserView instead of webview.

brenca:
#### Description of Change

This PR aims to bring the native `window.open` implementation and the custom shimmed one closer together, while also adding functionality that makes them more usable. That functionality includes:
 - passing `referrer` properly wherever it's possible to do so
 - passing `additionalFeatures` in the native `window.open` 
 - passing `POST` data (and headers for it) in both implementations to the `new-window` event, so the handler of the event can pass it along in `loadURL` to the custom `BrowserWindow`, if needed

Fixes #10852.
Fixes https://github.com/electron/electron/issues/18460.
Fixes https://github.com/electron/electron/issues/9340.
Fixes #18513.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Enhanced window.open functionality with the ability to handle `POST` data in the `new-window` event.


psiservices-ccounterman:
This solves significant issues for me, I hope it can be incorporated soon.

patches/chromium/can_create_window.patch
```diff
@@ -5,68 +5,129 @@ Subject: can_create_window.patch
...
+ #include "content/common/content_export.h"
++#include "content/common/frame.mojom.h"
+ #include "content/public/browser/bluetooth_chooser.h"
+ #include "content/public/browser/bluetooth_scanning_prompt.h"
+ #include "content/public/browser/invalidate_type.h"
+@@ -334,6 +335,12 @@ class CONTENT_EXPORT WebContentsDelegate {
+       const std::string& partition_id,
+       SessionStorageNamespace* session_storage_namespace);
+
++  virtual void WebContentsCreated(WebContents* source_contents,
```
zcbenz: I prefer using a different name than `WebContentsCreated` to make code easier to understand, otherwise it is hard distinguish between the 2 methods with the same name.

lib/browser/guest-view-manager.js
```diff
@@ -128,17 +128,6 @@ const createGuest = function (embedder, params) {
     }
   })
 
-  // Forward internal web contents event to embedder to handle
-  // native window.open setup
-  guest.on('-add-new-contents', (...args) => {
```
zcbenz: I thought this was needed to make native `window.open` work in webview, is it no longer needed?

zcbenz:
This PR requires review from @electron/wg-upgrades.

lib/browser/guest-view-manager.js
brenca: It's no longer needed - I think this was there to make sure that the `-add-new-contents` is handled because only `BrowserWindow`s and `BrowserView`s had this event handled, but I expanded that to be handled for every `WebContents` that's not `remote`.
https://github.com/electron/electron/pull/19703/files#diff-940e61f1a69c6bbbeb31968b6fc49d1cR378

blackie4:
need postData! How to get this branch release packager?

patches/chromium/can_create_window.patch
brenca: @zcbenz I resolved this, can I have another look?

zcbenz:
Do you mind rebasing the branch to solve the conflict?

And /cc @electron/wg-upgrades as I can not merge the PR without approval from the group.

nornagon:
Tests are failing legitimately:
```
Unhandled exception in main spec runner: electron/spec-main/api-web-contents-spec.ts(703,55): error TS2322: Type 'string' is not assignable to type 'NativeImage'.
```

pushkin-:
@brenca If this PR also fixes [this issue](https://github.com/electron/electron/issues/18513), can that be documented somewhere, so that issue is closed when this PR goes through? Thank you

brenca:
Ready for review from @electron/wg-upgrades

loc:
👋 Going to slip in and try to get this landed, as I have some work that depends on it.

lib/browser/guest-window-manager.js
```diff
@@ -80,6 +84,19 @@ const mergeBrowserWindowOptions = function (embedder, options) {
   return options
 }
 
+// Figure out appropriate headers for post data
+const headersForPostData = function (postData) {
+  if (postData.length > 0) {
```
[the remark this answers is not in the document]
loc: Good catch. I assume this is a mistake — reverting.

patches/chromium/can_create_window.patch
```diff
@@ -3,49 +3,67 @@ From: Cheng Zhao <zcbenz@gmail.com>
...
-index d5845d0263477b01ddf1e8bf5075e1a813751360..3d0bb0760b222eb3dd916ec032179fec62b3e6cc 100644
+index 4e0376f47bcec0afda2341695a6d951b023bd913..950ba21bcc2cf2328a2e779ce08362826df3b6e4 100644
 --- a/content/browser/frame_host/render_frame_host_impl.cc
 +++ b/content/browser/frame_host/render_frame_host_impl.cc
-@@ -4023,6 +4023,7 @@ void RenderFrameHostImpl::CreateNewWindow(
+@@ -3913,6 +3913,7 @@ void RenderFrameHostImpl::CreateNewWindow(
            last_committed_origin_, params->window_container_type,
            params->target_url, params->referrer.To<Referrer>(),
            params->frame_name, params->disposition, *params->features,
-+          params->additional_features, params->body,
```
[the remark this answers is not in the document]
loc: Yep, this was not working.

docs/api/web-contents.md
```diff
@@ -150,6 +150,9 @@ Returns:
 * `referrer` [Referrer](structures/referrer.md) - The referrer that will be
   passed to the new window. May or may not result in the `Referer` header being
   sent, depending on the referrer policy.
+* `postBody` [PostBody](structures/post-body.md) (optional) - The post data that
+  will be sent to the new window, along with the appropriate headers that will
+  be set. If no post data is to be sent, the value will be `null`.
```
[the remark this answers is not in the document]
loc: That's my understanding as well. Added a note here to make that clear.

lib/browser/guest-window-manager.js
```diff
@@ -131,14 +147,7 @@ const createGuest = function (embedder, url, referrer, frameName, options, postD
...
       loadOptions.postData = postData
-      loadOptions.extraHeaders = 'content-type: application/x-www-form-urlencoded'
-      if (postData.length > 0) {
-        const postDataFront = postData[0].bytes.toString()
-        const boundary = /^--.*[^-\r\n]/.exec(postDataFront)
-        if (boundary != null) {
-          loadOptions.extraHeaders = `content-type: multipart/form-data; boundary=${boundary[0].substr(2)}`
-        }
-      }
+      loadOptions.extraHeaders = headersForPostData(postData)
```
nornagon: is 'postData` guaranteed to be a buffer here? because the docs suggest it could be any of a number of different things.

patches/chromium/can_create_window.patch
```diff
@@ -77,10 +143,12 @@ index 2ddadfe3aa8aa8be885001429c5e817063788199..08ef43b9dcd13ee48598dd50b8933476
...
  #include "content/renderer/media/webrtc/rtc_peer_connection_handler.h"
  #include "content/renderer/render_frame_impl.h"
-@@ -1341,6 +1342,8 @@ WebView* RenderViewImpl::CreateView(
+@@ -1341,6 +1342,10 @@ WebView* RenderViewImpl::CreateView(
    }
    params->features = ConvertWebWindowFeaturesToMojoWindowFeatures(features);
- 
+
++  params->raw_features = features.raw_features.Utf8(
++      WTF::UTF8ConversionMode::kStrictUTF8ConversionReplacingUnpairedSurrogatesWithFFFD);
```
nornagon: my favourite utf8 conversion mode

lib/browser/guest-window-manager.js
loc: Ugh, no, it's not at all guaranteed 😕

loc: Actually, good news, I read the code wrong. Added comments to make it clear what is happening here, but this is safe in my tests of target="_blank" window openings.

docs/api/structures/post-body.md
```diff
@@ -0,0 +1,6 @@
+# PostBody Object
+
+* `data` Array<[PostData](./post-data.md)> - The post data to be sent to the
+  new window.
+* `headers` String - The appropriate `content-type` headers for the data. Can be
```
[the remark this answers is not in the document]
loc: It will never be more than `content-type`, so I think your initial comment is correct; we should change the name to contentTypeHeader.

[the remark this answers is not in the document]
loc: Maybe `rawContentTypeHeader`?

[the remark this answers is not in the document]
nornagon: Is this structure controlled/controllable by the user? Or only something that we give to the user informationally?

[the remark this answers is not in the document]
loc: It's informational.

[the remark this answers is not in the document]
nornagon: Then we should probably rephrase the docs slightly, too, e.g.

```md
* `contentType` String - The `content-type` header used for the data. One of
  `application/x-www-form-urlencoded` or `multipart/form-data`.
```

Also I don't see any reason to include the `content-type: ` part here.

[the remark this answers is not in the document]
loc: Sounds good to me.

pushkin-:
does anyone have an estimate of which Electron version this will land in? Thanks

miniak:
@brenca are you sure it fixes #9340?

lib/browser/guest-window-manager.js
```diff
@@ -81,6 +85,40 @@ const mergeBrowserWindowOptions = function (embedder, options) {
...
+}
+
+// Figure out appropriate headers for post data.
+const parseContentTypeFormat = function (postData) {
+  if (postData.length) {
+    // For multipart forms, the first element will start with the boundary
+    // notice, which looks something like `------WebKitFormBoundary12345678`
+    // Note, this regex would fail when submitting a urlencoded form with an
+    // input attribute of name="--theKey", but, uhh, don't do that?
+    const postDataFront = postData[0].bytes.toString()
```
nornagon: Why is it OK to assume that `postData[0].bytes` is defined here (as opposed to `filePath`, `blobUUID`, or any of the other options for specifying data on a PostData object)?

loc: Because the first element in the postData array (if it's an array, it means it's a multipart form) is always describing the boundary, not the actual content (which would be the next element in the array).

zcbenz:
@jkleinsc Can you take another look at this PR since you had requested changes?

lib/browser/guest-window-manager.js
```diff
@@ -83,6 +87,40 @@ const mergeBrowserWindowOptions = function (embedder, options) {
...
+  return header
+}
+
+// Figure out appropriate headers for post data.
+const parseContentTypeFormat = function (postData) {
+  if (postData.length) {
+    // For multipart forms, the first element will start with the boundary
+    // notice, which looks something like `------WebKitFormBoundary12345678`
+    // Note, this regex would fail when submitting a urlencoded form with an
+    // input attribute of name="--theKey", but, uhh, don't do that?
```
jkleinsc: Should this be documented in our documentation?

jkleinsc: Something along the lines of urlencoded forms cannot contain inputs that start with `--`

loc: Hmm. It's only when you're using a form that's `target="_blank"` _and_ when your webPreferences include `nativeWindowOpen: false`. I think I'm more inclined to make it a bug, since it's not something we can't solve, it's just that we haven't; what do you think? If you still think it should go in the docs, where should it go?

jkleinsc: @loc I think we should document it in the docs/api/structures/post-body.md doc: https://github.com/electron/electron/pull/19703/files#diff-5f5fc877666cac38e61b5292645c18f1R1

jkleinsc: The reason for documented this is if/when someone runs into this and files an issue saying that its a bug we can point to docs that state it is not supported.

loc: Done!

loc:
Merged master and added docs; should be ready to go.

loc:
@jkleinsc @nornagon @zcbenz somebody with merge access want to hit it? 🙏

pushkin-:
Should the release notes include the other issues that were fixed?

pushkin-:
@loc Is this going to be backported to 8?

pushkin-:
@loc @brenca @nornagon  This is currently listed as Fixed for Next Release in the Electron 8 project. Is that accurate? Will it be backported to 8, or should we just expect it in 9+? Thanks

loc:
@pushkin- you're correct, this isn't currently backported to 8. I'll take some time next week to look into fixing that.

pushkin-:
@loc At this point, I'm fine with it being backported to 9 only. Is that still the plan, or will this only be available in 10?

Issue #18745, opened 2019-06-12
Undo/Redo/Cut/Copy/Paste & more Not work using Menubar in Windows


damianaiamad:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.3
* **Operating System:**
  * Windows 10 (1809)
* **Last Known Working Electron version:**
  * 4.2.4

### Expected Behavior
Selecting Undo/Redo/Cut/Copy/Paste/ZoomIn/ZoomOut/ResetZoom from menu bar with mouse performs those various functions.

### Actual Behavior
Selecting these options from the menubar using the mouse seems to have no effect.
Using keyboard shortcuts does work ok.

Both mouse & keyboard works ok for all functions in MacOS 10.14.5.

### To Reproduce

https://github.com/electron/fiddle/releases/tag/v0.8.1
Download and run, then try Edit and View menus.
This uses 5.0.1

I have same problem in my own app using 5.0.3. Problem eliminated by downgrading to 4.2.4.

### Additional Information

My app uses standard MenuItem roles:
```
        { role: 'resetzoom' },
        { role: 'zoomin' },
        { role: 'zoomout' },

```

Electron Fiddle does the same:
https://github.com/electron/fiddle/blob/88ce4a76e97c4c93821306b59b3de9a147378058/src/main/menu.ts
```
     item.submenu.push({ type: 'separator' }, { role: 'resetzoom' }, { role: 'zoomin' }, { role: 'zoomout' }); // Add zooming actions
```

codebytere:
cc @felixrieseberg, perhaps this is a fiddle issue?

damianaiamad:
My app has the same problems, so I don't think it's just a fiddle problem.

Also, these "role" based menu behaviours are provided by Electron, not by application code.

damianaiamad:
I did some more investigation and can reproduce all problems with **Electron API Demo**.
https://github.com/electron/electron-api-demos

Current version uses Electron@4, so upgrade to Electron@5 as follows:

**To Reproduce**

$ git clone https://github.com/electron/electron-api-demos
$ cd electron-api-demos

[...]

PR #19710, opened 2019-08-09
fix: make sure that menu bar gets focus even when you click an item to focus it first

brenca:
#### Description of Change

Pretty much what the title says, right now, if you click on the menu bar and hit `Esc`, you can't use the arrow keys to navigate because the menu bar itself never got focus. This fixes that.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed a bug where you couldn't navigate the menu bar with the keyboard if you used the mouse to focus it.

### DIFF
--- a/shell/browser/ui/views/menu_bar.cc
+++ b/shell/browser/ui/views/menu_bar.cc
@@ -268,6 +268,12 @@ void MenuBar::OnMenuButtonClicked(views::Button* source,
   if (!window_->HasFocus())
     window_->RequestFocus();
 
+  // This ensures that if you focus the menubar by clicking on an item, you can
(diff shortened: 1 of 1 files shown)


nornagon:
This seems to have reintroduced an issue that https://github.com/electron/electron/pull/19708 fixed, namely that clicking the menu bar causes the main window to lose focus.

erickzhao:
Looking back, that does make sense as giving focus to the menu bar might un-focus the input fields. I wasn't careful enough to check the Fiddle example for issue #18745 w/ this PR.

I don't have access to a dev setup on Windows or Linux at the moment, but maybe @brenca would have an idea?

PR #19711, opened 2019-08-10
build: fix nightly install command in release notes

MarshallOfSound:
We publish nightlies to `electron-nightly` nowadays

Notes: no-notes


script/release/prepare-release.js
```diff
@@ -90,7 +90,7 @@ async function createRelease (branchToTarget, isBeta) {
       releaseBody = `Note: This is a nightly release.  Please file new issues ` +
         `for any bugs you find in it.\n \n This release is published to npm ` +
         `under the nightly tag and can be installed via npm install electron@nightly, ` +
```
HashimotoYT: This installs 4.0.0 nightly for me

```suggestion
         `under the nightly tag and can be installed via npm install electron-nightly, ` +
```

codebytere: @HashimotoYT yes, we changed it after 4.
https://www.npmjs.com/package/electron you can see not a single nightly is here past 4

PR #19714, opened 2019-08-11
feat: migrate webRequest module to NetworkService (Part 5)

zcbenz:
#### Description of Change

This PR makes `ProxyingURLLoaderFactory::InProgressRequest` actually invoke the methods of `webRequest` module.

The next step would be actually invoking the listeners.

#### Release Notes

Notes: no-notes

### DIFF
--- a/shell/browser/api/atom_api_session.cc
+++ b/shell/browser/api/atom_api_session.cc
@@ -575,7 +575,7 @@ v8::Local<v8::Value> Session::Protocol(v8::Isolate* isolate) {
 
 v8::Local<v8::Value> Session::WebRequest(v8::Isolate* isolate) {
   if (web_request_.IsEmpty()) {
-    auto handle = WebRequestNS::Create(isolate, browser_context());
+    auto handle = WebRequestNS::FromOrCreate(isolate, browser_context());
     web_request_.Reset(isolate, handle.ToV8());
   }
   return v8::Local<v8::Value>::New(isolate, web_request_);
--- a/shell/browser/api/atom_api_session.cc
+++ b/shell/browser/api/atom_api_session.cc
@@ -577,7 +577,7 @@ v8::Local<v8::Value> Session::Protocol(v8::Isolate* isolate) {
 
 v8::Local<v8::Value> Session::WebRequest(v8::Isolate* isolate) {
   if (web_request_.IsEmpty()) {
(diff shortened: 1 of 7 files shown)


shell/browser/api/atom_api_web_request_ns.cc
```diff
@@ -94,15 +191,62 @@ void WebRequestNS::SetListener(Event event, gin::Arguments* args) {
...
     v8::Isolate* isolate,
-    AtomBrowserContext* browser_context) {
-  return gin::CreateHandle(isolate, new WebRequestNS(isolate, browser_context));
+    content::BrowserContext* browser_context) {
+  auto* web_request = From(browser_context);
+  if (!web_request) {
+    // Create.
+    web_request = new WebRequestNS(isolate, browser_context);
+    // Make sure the |Session| object has the |webRequest| property managed.
+    Session::CreateFrom(isolate,
```
deepak1556: This seems circular, `Session::WebRequest` => `WebRequest::FromOrCreate` => `Session::WebRequest`.

shell/browser/atom_browser_client.cc
```diff
@@ -979,13 +980,14 @@ bool AtomBrowserClient::WillCreateURLLoaderFactory(
     bool* bypass_redirect_checks) {
   content::WebContents* web_contents =
       content::WebContents::FromRenderFrameHost(frame_host);
-  if (!web_contents) {
```
deepak1556: This was added for service workers that crashed here since they don't have any associated `web_contents`. https://github.com/electron/electron/pull/19488

shell/browser/api/atom_api_session.cc
```diff
@@ -206,8 +206,10 @@ Session::Session(v8::Isolate* isolate, AtomBrowserContext* browser_context)
 Session::~Session() {
   content::BrowserContext::GetDownloadManager(browser_context())
       ->RemoveObserver(this);
+  // TODO(zcbenz): Now since URLRequestContextGetter is gone, is this still
+  // needed?
+  // Refs https://github.com/electron/electron/pull/12305.
   DestroyGlobalHandle(isolate(), cookies_);
```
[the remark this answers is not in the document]
zcbenz: `api::Cookies` and others are keeping strong references to `BrowserContext` and thus the headache of destructions, I think we can make them use weak reference (plain pointer) and things should be much simpler.

PR #19719, opened 2019-08-12
fix: don't handle browser messages before document element is created (6-0-x)

zcbenz:
Backport of #19718.

Notes: Fix async Node APIs not working after received IPC message.


l7s:
Do we still need this commit c0e9761b or perhaps now we have a proper fix and it should be reverted?

zcbenz:
I think it is no longer needed, will start a PR to revert it.

PR #19535, opened 2019-07-30
fix: normalize behavior of `win.setOpacity()` for invalid number values across operating systems

erickzhao:
### Description of Change
Small fix to normalize behavior for invalid values for the [`win.setOpacity()`](https://electronjs.org/docs/api/browser-window#winsetopacityopacity-windows-macos).

For example: `win.setOpacity(100)` or `win.setOpacity(-100)`.

#### Current behavior with invalid values
* Windows:
  * `win.setOpacity(opacity)`: supposed to normalize the [0,1] opacity values to 255. However, for invalid values, passing in `opacity * 255` leads to unexpected behavior.
  * `win.getOpacity()`: retrieves the invalid `opacity` value that was previously passed in.
* macOS:
  * `win.setOpacity(opacity)`: bounds opacity value to the [0,1] range (i.e. `(opacity < 0) -> (opacity = 0)` and `(opacity > 1) -> (opacity = 1)`)
  * `win.getOpacity()`: retrieves the invalid `opacity` value that was previously passed in.

* Linux (unsupported):
  * `win.setOpacity(opacity)`: takes no action on the actual window opacity, but stores the `opacity` value.
  * `win.getOpacity()`: retrieves the previously stored `opacity` value.

#### Proposed behavior

* macOS/Windows:
  * `win.setOpacity(opacity)`: bounds opacity value to the [0,1] range
  * `win.getOpacity()`: retrieves the bounded opacity value.
* Linux:
  * `win.setOpacity(opacity)`:  takes no action.
  * `win.getOpacity()`: always returns 1.

### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

### Release Notes

Notes: Normalized out-of-bound value behavior for the `setOpacity()` API in `BrowserWindow`.

PR #19723, opened 2019-08-12
fix: normalize behavior of `win.setOpacity()` for invalid number values across operating systems

erickzhao:
Backport of #19535.

See that PR for more details.

Notes: Normalized out-of-bound value behavior for the `setOpacity()` API in `BrowserWindow`.


codebytere:
@erickzhao `6-0-x` doesn't use `expect` in that file: those tests all need to be `assert`-based, which is why CI is all failing.

Issue #18942, opened 2019-06-22
Set use_vaapi=true build flag


TRPB:
According to https://aur.archlinux.org/packages/chromium-vaapi 

> I have learned today that we don't need to patch Chromium sources to get VA-API support, an interesting alternative is to simply compile Chromium with use_vaapi=true flag and libva as a dependency, and then VA-API stays disabled by default, but users can enable it on case-by-case basis using --ignore-gpu-blacklist.

I'd like to see Electron built with VA-API if possible. An it looks like Chromium now supports it. 

I haven't been able to test this as running `gn gen out/Release --args="import(\"//electron/build/args/release.gn\") use_vaapi=true"`

gives me the following error:

```
ERROR at //printing/BUILD.gn:151:22: Script returned non-zero exit code.
      cups_version = exec_script("cups_config_helper.py",
                     ^----------
Current dir: /home/media/electron-gn/src/out/Release/
Command: python /home/media/electron-gn/src/printing/cups_config_helper.py --api-version /home/media/electron-gn/src/build/linux/debian_sid_amd64-sysroot
Returned 1.
stderr:

  File "/home/media/electron-gn/src/printing/cups_config_helper.py", line 70
    print 'cups-config not found: %s' % cups_config
                                    ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print('cups-config not found: %s' % cups_config)?

See //BUILD.gn:142:7: which caused the file to be included.
      "//printing:printing_unittests",
```

and I have no idea where to start on fixing so cannot tell whether or not this will work, but it looks like it should as long as electron uses the version of Chromium that looks for use_vaapi.

MarshallOfSound:
@TRPB That error looks like you're using the wrong version of python.  It needs to be using Python 2.7 for all of Chromiums build scripts

TRPB:
thanks, yes, I managed to build it. Unfortunately, although it did something, I couldn't get any videos to play with hardware acceleration enabled.

tcd156:
I have had the same issue - I attempted to add the patches from:
https://chromium-review.googlesource.com/c/chromium/src/+/532294 
and
https://github.com/saiarcot895/chromium-ubuntu-build/blob/master/debian/patches/enable_vaapi_on_linux_2.diff

with no luck. 

I take the dist.zip and use it in electron-builder, and build an AppImage. I haven't been able to see hardware acceleration work yet. The ubuntu build from above has worked for me.

[...]

PR #19736, opened 2019-08-13
build: enable vaapi support by default

tcd156:
#### Description of Change

As a result of this conversation: https://github.com/electron/electron/issues/18942 I would like to propose that `use_vaapi = true` is added as a build flag to Electron builds so that Linux users can have hardware acceleration if they choose.

This does not change the default behavior of Electron, but compiles Chromium with the ability to use VAAPI drivers which can be enabled on Linux systems by using the `--ignore-gpu-blacklist` flag at runtime, as well as by installing any necessary dependencies (in that conversation, I mention Ubunutu specific deps that must be installed to work).

#### Checklist

#### Release Notes

Notes: Added use_vaapi as a build time flag to enable hardware acceleration for Linux builds

### DIFF
--- a/build/args/all.gn
+++ b/build/args/all.gn
@@ -24,3 +24,8 @@ is_cfi = false
 # TODO: Remove this and update CI to contain 10.14 SDK once
 # crbug.com/986701 is fixed.
 mac_sdk_min = "10.13"
+
+# This will build electron with vaapi support in order to enable hardware
+# acceleration - however, it will need to be enabled intentionally by using
+# the --ignore-gpu-blacklist flag and installing necessary dependencies.
+use_vaapi = true


MarshallOfSound:
@tcd156 We are trying to remain as close as possible to upstream Chromium to avoid us having to handle upstream issues in code paths that upstream does not exercise.  Once Chromium enables this by default we will inherit that flag.

Closing out as per above

tcd156:
@MarshallOfSound I can understand that logic, but is there no way to get this into a build so it can be enabled? I mean, Chromium is probably never going to enable this by default (or at least, based on all of the information available out there, at least not any time soon).

I mean it's not difficult enough at least for us to maintain a branch that has this enabled, I guess I'd just like to see some barrier lowered for Linux users wanting hardware acceleration.

MarshallOfSound:
@tcd156 You can definitely build your own versions of Electron and use them, as you can see it's just changing a single flag and building yourself.  Unfortunately it's not something we can maintain or support.

> Chromium is probably never going to enable this by default (or at least, based on all of the information available out there, at least not any time soon).

That unfortunately means that we will probably never do it either

gardotd426:
>Chromium is probably never going to enable this by default (or at least, based on all of the information available out there, at least not any time soon).

This was actually recently proven false, and Chromium now enables it by default (and it's now even included in all chromium-based browsers including Chrome, Brave, and Vivaldi). That change anything?

Issue #19738, opened 2019-08-13
ELECTRON_RUN_AS_NODE can be erroneously set for render processes


stevebaxter:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v4.2.5
* **Operating System:**
  * macOS 10.13.6 / Windows 10 (1803)
* **Last Known Working Electron version:**
  * N/A

### Expected Behavior
ELECTRON_RUN_AS_NODE should never be set in the Electron main `process.env`.

### Actual Behavior
`ELECTRON_RUN_AS_NODE` is set in the Electron main `process.env` after calling `child_process.fork()`, passing in `process.env` directly as `options.env`.

This can cause a number of problems, the main symptom is that reloading the browser window will no longer work. It is likely that creating additional browser windows will also no longer work after this point.

### To Reproduce
See fiddle:

https://gist.github.com/cdc803312b25b151d53629d0c13bc896

The console output is:
```
Before fork, ELECTRON_RUN_AS_NODE is  undefined
After fork, ELECTRON_RUN_AS_NODE is  1
Hello world
```

ELECTRON_RUN_AS_NODE should not be set!

### Additional Information
The Electron version of Node gets patched to set `ELECTRON_RUN_AS_NODE=1` in the environment of the child process:

https://github.com/electron/electron/blob/9713fa09e75342094e382399be87633246e5437b/patches/node/refactor_alter_child_process_fork_to_use_execute_script_with.patch

It's quite common though for code to pass in `process.env` as an option to child_process.fork() (worker-farm does this for instance). In this case, this line changes the actual process.env for the current process, not a copy:

https://github.com/electron/electron/blob/9713fa09e75342094e382399be87633246e5437b/patches/node/refactor_alter_child_process_fork_to_use_execute_script_with.patch#L21

Initially this is harmless, however the next time any new process is spawned it will inherit the current environment including `ELECTRON_RUN_AS_NODE`, even if it should be run as Electron. This causes all sorts of malfunctions.

I suspect the solution will be to ensure we modify a copy of the environment we have been passed, for instance changing:

```
[...]

PR #19742, opened 2019-08-14
fix: ensure child_process.fork() doesn't modify main process env

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/19738.

As stated in the issue, `ELECTRON_RUN_AS_NODE` is set in the Electron main `process.env` after `calling child_process.fork()`, passing in `process.env` directly as options.env. This means that the value for `ELECTRON_RUN_AS_NODE` is actually changed in the main process env, which causes unwanted side effects.

This PR remedies that by making a new `process.env` object using the existing one as a prototype
to avoid changing the main process env. 

Verified with https://gist.github.com/cdc803312b25b151d53629d0c13bc896:

```sh
Before fork, ELECTRON_RUN_AS_NODE is  undefined
After fork, ELECTRON_RUN_AS_NODE is  undefined
```

cc @MarshallOfSound

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where a call to `child_process.fork()` would set `ELECTRON_RUN_AS_NODE` in the main process.

### DIFF
--- a/patches/node/refactor_alter_child_process_fork_to_use_execute_script_with.patch
+++ b/patches/node/refactor_alter_child_process_fork_to_use_execute_script_with.patch
@@ -7,17 +7,16 @@ Subject: refactor: alter child_process.fork to use execute script with
 When forking a child script, we setup a special environment to make the Electron binary run like the upstream node. On Mac, we use the helper app as node binary.
 
 diff --git a/lib/child_process.js b/lib/child_process.js
-index 43257e53dfe0319d6c110e94529cbc991c2bcfb3..ec429a082b081f0289507c15aa9ecc5305345ca2 100644
+index 43257e53dfe0319d6c110e94529cbc991c2bcfb3..2f478625c8220fc86c34876c17c8f532d88ba922 100644
 --- a/lib/child_process.js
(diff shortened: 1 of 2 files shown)


patches/node/refactor_alter_child_process_fork_to_use_execute_script_with.patch
```diff
@@ -7,15 +7,21 @@ Subject: refactor: alter child_process.fork to use execute script with
...
 --- a/lib/child_process.js
 +++ b/lib/child_process.js
-@@ -102,6 +102,16 @@ function fork(modulePath /* , args, options */) {
+@@ -102,6 +102,22 @@ function fork(modulePath /* , args, options */) {
      throw new ERR_CHILD_PROCESS_IPC_REQUIRED('options.stdio');
    }
  
++  // When forking a child script, we setup a special environment to make
++  // the electron binary run like upstream Node.js
 +  if (!options.env) {
```
MarshallOfSound: Can this patch just become

```
options.env = Object.create(options.env || process.env)
```

To simplify this `if` 👍

PR #19759, opened 2019-08-14
docs: add exemplary fiddle for `launch in fiddle` feat

DeerMichel:
#### Description of Change
Adds a simple fiddle to demonstrate the `screen` module. 

Used to verify the `Launch in Fiddle` feat on [electronjs.org](https://electronjs.org/).

ref https://github.com/electron/electronjs.org/pull/2848

cc @MarshallOfSound @codebytere @ckerr 

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/screen.md
+++ b/docs/api/screen.md
@@ -14,7 +14,7 @@ property, so writing `let { screen } = require('electron')` will not work.
 
 An example of creating a window that fills the whole screen:
 
-```javascript fiddle='electron/master/docs/fiddles/screen/fit-screen'
+```javascript fiddle='screen/fit-screen'
 const { app, BrowserWindow, screen } = require('electron')
 
 let win
--- a/docs/api/screen.md
+++ b/docs/api/screen.md
@@ -14,7 +14,7 @@ property, so writing `let { screen } = require('electron')` will not work.
 
 An example of creating a window that fills the whole screen:
 
-```javascript fiddle='screen/fit-screen'
+```javascript fiddle='docs/fiddles/screen/fit-screen'
 const { app, BrowserWindow, screen } = require('electron')
 
 let win
--- a/docs/api/screen.md
+++ b/docs/api/screen.md
@@ -14,7 +14,7 @@ property, so writing `let { screen } = require('electron')` will not work.
 
 An example of creating a window that fills the whole screen:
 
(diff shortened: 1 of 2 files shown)


docs/api/screen.md
```diff
@@ -14,7 +14,7 @@ property, so writing `let { screen } = require('electron')` will not work.
 
 An example of creating a window that fills the whole screen:
 
-```javascript
+```javascript fiddle='electron/master/docs/fiddles/screen/fit-screen'
```
MarshallOfSound: Hm, at scale this is gonna suck because it means when we make a version branch we have to adjust all these strings from `master` to `my-version`.  Can't we make it so that the website repo infers the `master` part and replaces it with the commit SHA the docs were pulled from?

DeerMichel: you mean pulling it from:
image

and just write `electron/docs/fiddles/screen/..`

MarshallOfSound: Personally I'd just put it as 

```
javascript fiddle='docs/fiddles/screen/fit-screen'
```

and infer the version + put the `electron` there.  i.e. `electron/{version}/{path}`

We shouldn't ever put a `gist` fiddle in the docs because then it isn't versioned and can be manipulated outside of the PR review process

DeerMichel: then how about also getting rid of the `docs/fiddles` part 🤔 and just write `fiddle='screen/fit-screen'`

MarshallOfSound: One day we might want to change that path or put fiddles in "guides" it should allow us to do that still IMO

MarshallOfSound:
https://github.com/electron/electron/pull/19759#discussion_r314170582

docs/api/screen.md
DeerMichel: i mean it's not a big deal 😆 - so changing the root to be `e/e` and you'll give me green light 🍏?

MarshallOfSound: Haha this has descended into violent amounts of future-proofing 😆 

Technically I've already approved 😆 but I'll be 100% happy if the root is the root of `e/e`

DeerMichel:
Oki, defaulting to root. Ready to merge @MarshallOfSound  :)

PR #19756, opened 2019-08-14
fix: use GTK3 stock i18n strings

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/19755.

We should be using GTK stock labels on GTK3; the translations will be broken for those using GTK4 but it's in our best interest for now to wait and provide internationalized strings when GTK provides support for it.

cc @ckerr @erickzhao 

#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Correctly internationalize i18n in GTK-based Linux dialogs.

PR #19760, opened 2019-08-14
fix: use i18nized strings in gtk dialog buttons

ckerr:
Backport of #19756

See that PR for details.

NB: trop failed so this is a manual backport.

CC @codebytere 

#### Release Notes

Notes: Fixed i18n of GTK dialog buttons.


codebytere:
(took off the 5/4 labels since they failed and it's easier for trop if all the `needs-manual` are consolidated)

PR #19808, opened 2019-08-16
fix: disable LayoutNG to prevent random crashes in selection calcuations

MarshallOfSound:
Upstream Bug: https://bugs.chromium.org/p/chromium/issues/detail?id=992929
CL that caused it: https://chromium-review.googlesource.com/c/chromium/src/+/1598319

Even Chrome doesn't have LayoutNG enabled in C76 and if you try to turn it on Chrome violently crashes so I think we're right to turn this off.

In particular LayoutNG causes a crash when selecting "wbr" elements in certain fonts under certain layout conditions.  (I can't be much more specific or do a repro outside of a Slack specific message).

Notes: Fixed occasional crashes while selecting text


alfredxing:
Thanks Sam! 😍

PR #18288, opened 2019-05-13
fix: update the FileSelectHelper to support the new promise API

MarshallOfSound:
Fixes #18254

So it turns out we've successfully introduced a way to write
non-typesafe C++.

This fixes two things:
* Uses the object the promise resolves
* Ensures we attach the Then handler before moving the promise

Notes: Fixed the `input[type=file]` element not showing a file dialog

PR #19809, opened 2019-08-16
refactor: make util::Promise type safe when chaining in native

MarshallOfSound:
This solves a long-standing TODO whereby using `util::Promise::Then` is **not** type safe (in cpp I know) and could result in nasty issues like this one --> https://github.com/electron/electron/pull/18288 where cpp compiles but throws an uncatchable promise rejection internally because the two types native mate tries to convert between are incompatible.

This quote kinda sums it up:

> So it turns out we've successfully introduced a way to write non-typesafe C++.

This PR makes all `util::Promise` instances templated as `util::Promise<ResolveType>` and adds some helpful static_asserts to ensure the following three things

* You can only chain a `base::Callback` onto ` util::Promise` if the callback takes a single parameter that matches the `util::Promise` resolve type
* You can only resolve promises without a value (`Resolve()`) if the resolve type is `void*`
* You can't resolve promises with a value (`Result(v)`) if the resolve type is `void*`

This PR also removes the unused `CopyablePromise` helper, removes the generic `Reject<T>` method as IMO we should always reject with an error message and nothing was using it.  Also added a new `ResolveWithGin` helper to make the gin migration smoother (it uses gin to convert to v8)

This also IMO makes code involving our promise helper much easier to understand at a glance.  i.e. `util::Promise` before was a wildcard, and now `util::Promise<gfx::Image>` is really obvious what's going on.

Please review this quite well and _question anything that looks off to you_ this entire PR had to be done by hand.

Notes: no-notes

### DIFF
--- a/shell/common/promise_util.h
+++ b/shell/common/promise_util.h
@@ -6,6 +6,8 @@
 #define SHELL_COMMON_PROMISE_UTIL_H_
 
 #include <string>
+#include <tuple>
+#include <type_traits>
 #include <utility>
 
 #include "base/task/post_task.h"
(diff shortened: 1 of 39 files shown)


shell/common/promise_util.h
```diff
@@ -162,23 +163,36 @@ class Promise {
...
+    v8::Context::Scope context_scope(
+        v8::Local<v8::Context>::New(isolate(), GetContext()));
+
+    v8::Local<v8::String> error_message =
+        v8::String::NewFromUtf8(isolate(), string.c_str(),
+                                v8::NewStringType::kNormal,
+                                static_cast<int>(string.size()))
+            .ToLocalChecked();
+    v8::Local<v8::Value> error = v8::Exception::Error(error_message);
+    return GetInner()->Reject(GetContext(), (error));
```
ckerr: _(nit)_ the parens around error don't seem to do anything?

MarshallOfSound: I just copied this over from the impl, you're right though :D will update

MarshallOfSound:
> why use void* instead of void for the no-value case?

I didn't think `void` could be used as a function param type.  I.e. `Resolve(T value)` is invalid when `T == void`

deepak1556:
Everything looks good, sharing the same concern as @ckerr about `void*` to denote a null type, I would try to use `std::nullopt_t` https://en.cppreference.com/w/cpp/utility/optional/nullopt_t for this scenario.

PR #19811, opened 2019-08-17
fix: race condition in NodeStreamLoader

brenca:
#### Description of Change

Currently, it's possible to break the `protocol.registerStreamProtocol` API if you supply data to the stream too quickly (e.g. sending a file as a response), because of how node streams work. The problem specifically was that we listen to the `readable` event, and the [stream emits one right before emitting `end`](https://nodejs.org/api/stream.html#stream_event_readable). This can lead to the `end` event arriving while `NodeStreamLoader::ReadMore` is calling `stream.read()`, but before `is_writing_` is set, which causes a crash, since `NodeStreamLoader` deletes itself in this case.

Another thing that was happening was that `ReadMore` was called from two sources - whenever the stream emitted `readable`, and whenever a `DidWrite` call was successful, which lead to two `ReadMore` calls running at once, which could lead to chunks being sent in the wrong order. This PR fixes that problem as well.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a race condition that occurred when using protocol.registerStreamProtocol.

### DIFF
--- a/shell/browser/net/node_stream_loader.cc
+++ b/shell/browser/net/node_stream_loader.cc
@@ -71,12 +71,18 @@ void NodeStreamLoader::Start(network::ResourceResponseHead head) {
      base::BindRepeating(&NodeStreamLoader::NotifyComplete, weak, net::OK));
   On("error", base::BindRepeating(&NodeStreamLoader::NotifyComplete, weak,
                                   net::ERR_FAILED));
-  On("readable", base::BindRepeating(&NodeStreamLoader::ReadMore, weak));
+  On("readable", base::BindRepeating(&NodeStreamLoader::NotifyReadable, weak));
+}
+
+void NodeStreamLoader::NotifyReadable() {
+  if (!readable_)
+    ReadMore();
+  readable_ = true;
 }
 
 void NodeStreamLoader::NotifyComplete(int result) {
   // Wait until write finishes or fails.
-  if (is_writing_) {
+  if (is_reading_ || is_writing_) {
     ended_ = true;
     result_ = result;
     return;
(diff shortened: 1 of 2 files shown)


brenca:
trop run backport-to 7-0-x,6-0-x,5-0-x

deepak1556:
@codebytere @brenca there is no need to backport this on any of our existing release lines, the code path is not enabled in any of them and we won't be enabling it.

brenca:
Yeah, I realized that when looking at `6-0-x` as well, thanks for removing the labels!

HaNdTriX:
@brenca Is it possible that this PR broke the ability to server HTML pages to a browser window?

## Example

**Electron Version:** `7.0.0-beta.4`
**OS:** `macOS`

The following codes implements a [webpack devserver proxy](https://webpack.js.org/configuration/dev-server/).

### URL Mapping Example

| Electron URL            | Target URL                  |
|-------------------------|-----------------------------|
| `some-scheme://app`     | `http://localhost:1234/`    |
| `some-scheme://app/foo` | `http://localhost:1234/foo` |

### Code

```js
const { app, protocol, session } = require('electron')
const http = require('http')

const host = 'app'
const scheme = 'some-scheme'
const targetURL = 'http://localhost:1234/'

app.on('ready', () => {
  session.defaultSession.protocol.registerStreamProtocol(scheme, (request, callback) => {
    // Build proxy options
    const { url, ...options } = request

    // Rewrite url
    const webpackDevServerUrl = url
      .replace(`${scheme}://${host}`, targetURL)
      .replace(/\/$/, '')

    // Proxy request to webpack dev server endpoint
    http.request(webpackDevServerUrl, options, callback).end()
  })
})
```

### Result in Browser Window

Bildschirmfoto 2019-09-04 um 13 20 11

Issue #19775, opened 2019-08-15
How to prevent Electron Security Warning (Insecure Content-Security-Policy)


TomasHubelbauer:
I have a question about CSP in Electron 6 which has already been asked on Stack Overflow here:
https://stackoverflow.com/q/48854265/2715716

However, the question was not answered (there are only recommendations on how to disable the warning, not how to actually prevent the warning from showing).

There is a security checklist for Electron available at https://electronjs.org/docs/tutorial/security, which is very helpful. However, even in the simplest case, I still run afoul of these warnings even when following the most basic simplest Electron tutorial! https://electronjs.org/docs/tutorial/first-app

I don't believe that such a basic issue has been undiscovered for all this time and I'm the first person just coming across this, but alas I have not been able to find a solution.

My repro is simple, it consists of only 4 files:

`package.json`:

```json
{
  "main": "main.js",
  "scripts": {
    "start": "electron ."
  },
  "devDependencies": {
    "electron": "^6.0.2"
  }
}
```

`main.js`:

```js
const electron = require('electron');

electron.app.on('ready', () => {
  const window = new electron.BrowserWindow({ width: 640, height: 480 });
  window.loadFile('./index.html');
  window.webContents.openDevTools();
});
```

`index.html`:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <script src="index.js"></script>
  </head>
  <body></body>
</html>
```

`index.js`:

```js
window.addEventListener('load', () => {
  document.body.append(document.createTextNode('Hello, World!'));
});
```

This is how far I got with Electron before I ran into my first roadblock. Now how does one go about fixing up this code to avoid the warning?

From what I read, `nodeIntegration` has been `false` by default since version 5.

Hiding these warnings using `process.env['ELECTRON_DISABLE_SECURITY_WARNINGS'] = 'true';` is not a fix, it's hiding the warning.

I cannot attach CSP headers to the `index.html` file because it is loaded from the disk, not from the network.

Ignoring this warning because it is not present when the app is packaged is not a solution either, because it equals to hiding the warning (by ignoring it) instead of fixing it.

In #12035 @MarshallOfSound says to follow the security checklist and the warning will go away. What am I running afoul of in the security checklist in the above repro that they don't go away for me?

What is there left to do to get this warning to go away?

It would be awesome to have an answer for this in the Stack Overflow question as that's where Google takes you on your quest to find out.

TomasHubelbauer:
[...]

PR #19819, opened 2019-08-19
docs: Add a CSP meta tag to make the tutorial compiant with the security checklist

TomasHubelbauer:
I've asked #19775 because I was frustrated with how hard it was to find a way to fix (instead of hide) the CSP warning in Electron and I complained that even the official quick start guide wasn't compliant with the security checklist at https://electronjs.org/docs/tutorial/security. Someone helped me out with a CSP meta tag which I have later noticed is indeed mentioned in the checklist, too: https://electronjs.org/docs/tutorial/security#csp-meta-tag. I have not used the checklist one verbatim because it prevents a `script` tag from working when serving `index.html` through the `file:` protocol as the quick start does. I instead used the one the person in my issue recommended which seems to work well to me. I am not that well versed in CSP so there might be a better policy to include with the quick start, but this is what I've got for now.

notes: no-notes

### DIFF
--- a/docs/tutorial/first-app.md
+++ b/docs/tutorial/first-app.md
@@ -195,6 +195,8 @@ Finally the `index.html` is the web page you want to show:
   <head>
     <meta charset="UTF-8">
     <title>Hello World!</title>
+    <!-- https://electronjs.org/docs/tutorial/security#csp-meta-tag -->
+    <meta http-equiv="Content-Security-Policy" content="script-src 'self';" />
   </head>
   <body>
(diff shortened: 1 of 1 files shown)


betosimo:
this do not help me, I tried EVERY possible combination and still receiving this error: because it violates the following Content Security Policy directive: "default-src capacitor-electron://* 'unsafe-inline' data:". Note that 'script-src-elem' was not explicitly set, so 'default-src' is used as a fallback

PR #19820, opened 2019-08-19
feat: migrate webRequest module to NetworkService (Part 7)

zcbenz:
#### Description of Change

This PR makes the webRequest module work again.

Lots of related tests are still failing, and I'll fix them in other PRs.

#### Release Notes

Notes: no-notes

### DIFF
--- a/shell/browser/api/atom_api_web_request_ns.cc
+++ b/shell/browser/api/atom_api_web_request_ns.cc
@@ -316,16 +316,29 @@ template <typename Listener, typename Listeners, typename Event>
 void WebRequestNS::SetListener(Event event,
                                Listeners* listeners,
                                gin::Arguments* args) {
+  v8::Local<v8::Value> arg;
+
   // { urls }.
   std::set<URLPattern> patterns;
   gin::Dictionary dict(args->isolate());
-  args->GetNext(&dict) && dict.Get("urls", &patterns);
+  if (args->GetNext(&arg) && !arg->IsFunction()) {
+    // Note that gin treats Function as Dictionary when doing convertions, so we
+    // have to explicitly check if the argument is Function before try to
+    // convert it to Dictionary.
+    if (gin::ConvertFromV8(args->isolate(), arg, &dict)) {
+      dict.Get("urls", &patterns);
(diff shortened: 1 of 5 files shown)


shell/browser/api/atom_api_web_request_ns.cc
```diff
@@ -316,16 +320,29 @@ template <typename Listener, typename Listeners, typename Event>
...
+    // Note that gin treats Function as Dictionary when doing convertions, so we
+    // have to explicitly check if the argument is Function before try to
+    // convert it to Dictionary.
+    if (gin::ConvertFromV8(args->isolate(), arg, &dict)) {
+      dict.Get("urls", &patterns);
+      args->GetNext(&arg);
+    }
+  }
+
+  if (arg.IsEmpty()) {
```
deepak1556: Maybe its better to move it down and have a single check ?

`if (arg.IsEmpty() || !(gin::ConvertFromV8(args->isolate(), arg, &listener) || arg->IsNull()))`

shell/browser/api/atom_api_web_request_ns.cc
```diff
@@ -395,7 +418,11 @@ void WebRequestNS::OnListenerResult(uint64_t id,
       ReadFromResponse(isolate, &dict, out);
   }
 
-  std::move(callbacks_[id]).Run(result);
+  // The ProxyingURLLoaderFactory expects the callback to be executed
+  // asynchronously, because it used to work on IO thread before NetworkService.
+  base::SequencedTaskRunnerHandle::Get()->PostTask(
```
deepak1556: What happens when its not executed async ?

zcbenz: ProxyingURLLoaderFactory's logic is:
Post callback in next tick => pause network request => continue network request when callback is called

If we execute the callback in current tick then it becomes:
Run callback in current tick => continue network request(no-op) => pause network request

So the network request would be paused forever.

deepak1556: Thanks for the explanation!

PR #19831, opened 2019-08-19
ci: add macOS debug builds

jkleinsc:
#### Description of Change

As we now have resources to do so, we should enable debug builds for macOS.  This PR enables those builds.

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -582,9 +582,7 @@ source_set("electron_lib") {
   }
 
   if (enable_desktop_capturer) {
-    if (is_component_build && is_win) {
-      # On windows the implementation relies on unexported
-      # DxgiDuplicatorController class.
+    if (is_component_build && !is_linux) {
       deps += [ "//third_party/webrtc/modules/desktop_capture" ]
     }
     sources += [
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -707,9 +707,11 @@ steps-electron-build-for-tests: &steps-electron-build-for-tests
     - *step-depot-tools-add-to-path
     - *step-setup-env-for-build
     - *step-restore-brew-cache
+    - *step-get-more-space-on-mac
     - *step-install-npm-deps-on-mac
     - *step-fix-sync-on-mac
     - *step-gn-gen-default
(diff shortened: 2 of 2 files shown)


BUILD.gn
```diff
@@ -582,9 +582,7 @@ source_set("electron_lib") {
   }
 
   if (enable_desktop_capturer) {
-    if (is_component_build && is_win) {
-      # On windows the implementation relies on unexported
-      # DxgiDuplicatorController class.
+    if (is_component_build && !is_linux) {
       deps += [ "//third_party/webrtc/modules/desktop_capture" ]
```
deepak1556: What changed to add this dependency on mac ? Also please retain the comment for windows.

jkleinsc: @deepak1556 my bad on the comment.  As far as the mac dependency, it was need on mac to resolve this error on the debug build:
```Undefined symbols for architecture x86_64: "webrtc::GetWindowOwnerPid(unsigned int)" error.```   I can put up another PR correcting the comment.

Comment (jkleinsc) [reply BUILD.gn]: @deepak1556 PR to update comment is here: https://github.com/electron/electron/pull/19848

Comment (deepak1556) [reply BUILD.gn]: I don't see any relevant changes on the desktop capturer api code path for these new undefined symbols. Do you know what needed these ?

Comment (deepak1556) [reply BUILD.gn]: Ignore my comment, found https://github.com/electron/electron/pull/18926

Comment (deepak1556) [reply BUILD.gn]: @jkleinsc I would suggest to guard those implementations added in the above PR behind `enable_desktop_capturer` since they are dependent on desktop capturer. When building with `enable_desktop_capturer=false` , things will compile horribly.

```
void TopLevelWindow::MoveAbove(const std::string& sourceId,
                               mate::Arguments* args) {
#if  BUILDFLAG(ENABLE_DESKTOP_CAPTURER)
  if (!window_->MoveAbove(sourceId))
    args->ThrowError("Invalid media source id");
#endif
}
```

jkleinsc: Ok I'll add it to #19848

PR #19837, opened 2019-08-20
refactor: use Error util when only throwing errors

codebytere:
#### Description of Change

This PR applies abilities conferred by https://github.com/electron/electron/pull/19803 to use a bespoke error throwing utility in `atom_api_app` when a given function is only taking `mate:Arguments` as a parameter to throw arguments. There are not many and specific instances where we were using the previous antipattern, so this change should likely cause minimal churn.

cc @MarshallOfSound @ckerr 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/atom_api_app.cc
+++ b/shell/browser/api/atom_api_app.cc
@@ -846,11 +846,11 @@ void App::SetAppPath(const base::FilePath& app_path) {
 }
 
 #if !defined(OS_MACOSX)
-void App::SetAppLogsPath(base::Optional<base::FilePath> custom_path,
-                         mate::Arguments* args) {
+void App::SetAppLogsPath(util::ErrorThrower thrower,
+                         base::Optional<base::FilePath> custom_path) {
   if (custom_path.has_value()) {
     if (!custom_path->IsAbsolute()) {
-      args->ThrowError("Path must be absolute");
+      thrower.ThrowError("Path must be absolute");
       return;
     }
     base::PathService::Override(DIR_APP_LOGS, custom_path.value());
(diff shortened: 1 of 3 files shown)


shell/browser/api/atom_api_app.cc
```diff
@@ -846,11 +846,11 @@ void App::SetAppPath(const base::FilePath& app_path) {
 }
 
 #if !defined(OS_MACOSX)
-void App::SetAppLogsPath(base::Optional<base::FilePath> custom_path,
-                         mate::Arguments* args) {
+void App::SetAppLogsPath(util::ErrorThrower thrower,
```
jkleinsc: Why reverse the order of the args here?  It seems more logicall that the first arg would be the path.

shell/browser/api/atom_api_app.h
```diff
@@ -182,10 +183,10 @@ class App : public AtomBrowserClient::Delegate,
   bool RequestSingleInstanceLock();
   void ReleaseSingleInstanceLock();
   bool Relaunch(mate::Arguments* args);
-  void DisableHardwareAcceleration(mate::Arguments* args);
-  void DisableDomainBlockingFor3DAPIs(mate::Arguments* args);
+  void DisableHardwareAcceleration(util::ErrorThrower thrower);
+  void DisableDomainBlockingFor3DAPIs(util::ErrorThrower thrower);
   bool IsAccessibilitySupportEnabled();
-  void SetAccessibilitySupportEnabled(bool enabled, mate::Arguments* args);
+  void SetAccessibilitySupportEnabled(util::ErrorThrower thrower, bool enabled);
```
jkleinsc: Same comment here about order of args.

shell/browser/api/atom_api_app.cc
codebytere: fair! i'll change this 🙇‍♀

zcbenz: Currently we have 2 patterns for passing args:

1. `Func(v8::Isolate* isolate, arg)`: which follows convention of V8 and Node.js
2. `Func(arg, mate::Arguments* args)`: as `args` acts as rest arguments in js

Since `ErrorThrower` is a thin wrapper of `v8::Isolate`, I think it makes sense to put it as first arg.

PR #19841, opened 2019-08-20
feat: migrate webRequest module to NetworkService (Part 8)

zcbenz:
#### Description of Change

This PR fixes tests in `api-web-request-spec`.

The tests in `api-net` are still failing, it seems that `webRequest` has some problems with `net` module.

#### Release Notes

Notes: no-notes

### DIFF
--- a/shell/browser/api/atom_api_web_request_ns.cc
+++ b/shell/browser/api/atom_api_web_request_ns.cc
@@ -207,7 +207,7 @@ void ReadFromResponse(v8::Isolate* isolate,
   if (!response->Get("statusLine", &status_line))
     status_line = headers.second;
   v8::Local<v8::Value> value;
-  if (response->Get("responseHeaders", &value)) {
+  if (response->Get("responseHeaders", &value) && value->IsObject()) {
     *headers.first = new net::HttpResponseHeaders("");
     (*headers.first)->ReplaceStatusLine(status_line);
     gin::Converter<net::HttpResponseHeaders*>::FromV8(isolate, value,
--- a/shell/browser/api/atom_api_web_request_ns.cc
+++ b/shell/browser/api/atom_api_web_request_ns.cc
@@ -113,15 +113,10 @@ v8::Local<v8::Value> HttpResponseHeadersToV8(
     std::string key;
     std::string value;
     while (headers->EnumerateHeaderLines(&iter, &key, &value)) {
-      if (response_headers.FindKey(key)) {
-        base::ListValue* values = nullptr;
-        if (response_headers.GetList(key, &values))
-          values->AppendString(value);
-      } else {
-        auto values = std::make_unique<base::ListValue>();
-        values->AppendString(value);
-        response_headers.Set(key, std::move(values));
-      }
+      base::Value* values = response_headers.FindListKey(key);
+      if (!values)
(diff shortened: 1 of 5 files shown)


shell/browser/api/atom_api_web_request_ns.cc
```diff
@@ -176,7 +207,7 @@ void ReadFromResponse(v8::Isolate* isolate,
   if (!response->Get("statusLine", &status_line))
     status_line = headers.second;
   v8::Local<v8::Value> value;
-  if (response->Get("responseHeaders", &value)) {
+  if (response->Get("responseHeaders", &value) && value->IsObject()) {
```
allydurks: wouldn't you want to flip this condition? so that if the value isn't an object it's immediately false and there is no Get called on that response*?
```suggestion
  if (value->IsObject() && response->Get("responseHeaders", &value)) {
```

deepak1556: @allydurks the type of value is undefined until its extracted from the `response` dictionary, so the condition can't be flipped.

shell/browser/api/atom_api_web_request_ns.cc
```diff
@@ -99,6 +100,33 @@ bool MatchesFilterCondition(extensions::WebRequestInfo* info,
...
+  base::DictionaryValue response_headers;
+  if (headers) {
+    size_t iter = 0;
+    std::string key;
+    std::string value;
+    while (headers->EnumerateHeaderLines(&iter, &key, &value)) {
+      if (response_headers.FindKey(key)) {
+        base::ListValue* values = nullptr;
+        if (response_headers.GetList(key, &values))
+          values->AppendString(value);
```
ckerr: AppendString is deprecated, we could use GetList().emplace_back() instead. _(minor)_ Also we can avoid the second lookup if we use FindListKey()

```c++
Value* values = response_headers.FindListKey(key);
if (values != nullptr) {
  values->GetList().emplace_back(value);
}
```
What's the difference between `FindKey()` + `GetList()` and `FindListKey()`?

shell/browser/api/atom_api_web_request_ns.cc
```diff
@@ -99,6 +100,33 @@ bool MatchesFilterCondition(extensions::WebRequestInfo* info,
...
+      } else {
+        auto values = std::make_unique<base::ListValue>();
+        values->AppendString(value);
```
ckerr: AppendString is deprecated, but I think we can pass a ListStorage into the ListValue ctor, so something like

```c++
auto values = std::make_unique<base::ListValue>({ value }); // not sure if this will compile?
auto values = std::make_unique<base::ListValue>(base::Value::ListStorage(value));
```

spec/api-web-request-spec.js
```diff
@@ -12,7 +12,7 @@ chai.use(dirtyChai)
 /* The whole webRequest API doesn't use standard callbacks */
 /* eslint-disable standard/no-callback-literal */
 
-describe.skip('webRequest module', () => {
+describe('webRequest module', () => {
```
ckerr: :tada:

shell/browser/api/atom_api_web_request_ns.cc
zcbenz: They should do the same work.

PR #19850, opened 2019-08-20
build: add gn-check to precommit linting

codebytere:
#### Description of Change

This PR adds the GN check step to `lint-staged` for changes to `*.{gn,gni}` files. This is intended to prevent changes to the GN file which cause errors in CI on the `gn-check` step that could easily have been caught had they been tested locally first.

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,8 +114,8 @@
       "remark -qf"
     ],
     "*.{gn,gni}": [
-      "python script/run-gn-format.py",
       "npm run gn-check",
+      "python script/run-gn-format.py",
       "git add"
     ],
     "*.py": [
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "create-typescript-definitions": "npm run create-api-json && electron-typescript-definitions --api=electron-api.json && node spec/ts-smoke/runner.js",
     "gn-typescript-definitions": "npm run create-typescript-definitions && shx cp electron.d.ts",
     "pre-flight": "pre-flight",
+    "gn-check": "node ./script/gn-check.js",
     "preinstall": "node -e 'process.exit(0)'",
     "prepack": "check-for-leaks",
     "repl": "node ./script/start.js --interactive",
@@ -114,6 +115,7 @@
     ],
     "*.{gn,gni}": [
       "python script/run-gn-format.py",
+      "npm run gn-check",
       "git add"
     ],
     "*.py": [
--- a/script/gn-check.js
+++ b/script/gn-check.js
@@ -1,21 +1,16 @@
 const sh = require('shelljs')
-const path = require('path')
-const fs = require('fs')
 
-const SRC_DIR = path.resolve(__dirname, '..', '..')
-let GN_CHECK_DIR
+const { getOutDir } = require('./lib/utils')
 
-if (process.env.ELECTRON_OUT_DIR) {
-  GN_CHECK_DIR = `../out/${process.env.ELECTRON_OUT_DIR}`
-} else {
-  for (const buildType of ['Debug', 'Testing', 'Release']) {
-    const outPath = path.resolve(SRC_DIR, 'out', buildType)
-    if (fs.existsSync(outPath)) {
-      GN_CHECK_DIR = `../out/${buildType}`
-      break
-    }
-  }
+const OUT_DIR = getOutDir()
+if (!OUT_DIR) {
+  throw new Error(`No viable out dir: one of Debug, Testing, or Release must exist.`)
 }
 
-const gnCheckString = `gn check ${GN_CHECK_DIR} //electron:electron_lib`
+const gnCheckString = `
+gn check ../out/${OUT_DIR} //electron:electron_lib &&
+gn check ../out/${OUT_DIR} //electron:electron_app &&
+gn check ../out/${OUT_DIR} //electron:manifests &&
+gn check ../out/${OUT_DIR} //electron/shell/common/api:mojo
+`
 sh.exit(sh.exec(gnCheckString).code)
--- a/script/gn-check.js
+++ b/script/gn-check.js
@@ -1,0 +1,21 @@
+const sh = require('shelljs')
+const path = require('path')
+const fs = require('fs')
+
+const SRC_DIR = path.resolve(__dirname, '..', '..')
+let GN_CHECK_DIR
+
+if (process.env.ELECTRON_OUT_DIR) {
+  GN_CHECK_DIR = `../out/${process.env.ELECTRON_OUT_DIR}`
+} else {
+  for (const buildType of ['Debug', 'Testing', 'Release']) {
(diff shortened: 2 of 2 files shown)


script/gn-check.js
```diff
@@ -0,0 +1,23 @@
...
+  GN_CHECK_DIR = `../out/${process.env.ELECTRON_OUT_DIR}`
+} else {
+  for (const buildType of ['Debug', 'Testing', 'Release']) {
+    const outPath = path.resolve(ELECTRON_DIR, 'out', buildType)
+    if (fs.existsSync(outPath)) {
+      GN_CHECK_DIR = `../out/${buildType}`
+      break
+    }
+  }
+}
```
ckerr: Does it matter that our method of picking a fallback directory iff `ELECTRON_OUT_DIR` isn't set differs here from other parts of our `script/`directory?

It might make sense to port `get_out_dir()` from `script/lib/util.py` to `script/lib/utils.js` and use that ported function here (and also in `utils.getElectronExec()`)

codebytere: `get_out_dir()` is not used by us really though, it's all in CI, whereas most of us use testing builds 🤔

script/gn-check.js
```diff
@@ -0,0 +1,23 @@
...
+
+const gnCheckString = `gn check ${GN_CHECK_DIR} //electron:electron_lib`
+if (sh.exec(gnCheckString).code !== 0) {
+  sh.exit(1)
+}
```
ckerr: Not a blocker for this use case, but as a rule-of-thumb it's better to pass the exit code along -- it's not uncommon for apps to return various codes to signal different errors.

Consider a hypothetical case where this script is called by another script that wants to do different things depending on the error code. We wouldn't want all nonzero codes to be mapped to 1.

```js
sh.exit(sh.exec(gnCheckString).code)
```

script/gn-check.js
```diff
@@ -0,0 +1,23 @@
+const sh = require('shelljs')
+const path = require('path')
+const fs = require('fs')
+
+const ELECTRON_DIR = path.resolve(__dirname, '..', '..')
```
ckerr: This looks like it would point to the directory `above` Electron. (Is this a copypaste error from `script/lib/utils.js`? :slightly_smiling_face:)

```suggestion
const { ELECTRON_DIR } = require('./lib/utils')
```

script/gn-check.js
ckerr: The logic of `get_out_dir()` is also what's used by `utils.getElectronExec()`, which we do use for our testing builds' spec runs though?

I'm not against having one set of rules for human runs and another set for CI runs; I just don't want to introduce inconsistency for the point of inconsistency. If this new behavior is preferable, maybe that's what we should be doing in `utils.getElectronExec()` too

script/gn-check.js
codebytere: @ckerr that's correct - this is intended to point to `src`, _not_ Electron. happy to change the name if that's clearer

script/gn-check.js
codebytere: hmm, ok!

codebytere: hmm, so `utils.getElectronExec()` just...creates `out/Debug` if none exists, which i don't feel is ideal behavior. I'd also rather change it, but perhaps in a follow-up PR?

codebytere: if there's perhaps a cleaner way to evaluate descending out dir type i'd be in favor of that though 😁

script/gn-check.js
ckerr: ahh of course, that makes sense. Maybe `ELECTRON_GN_DIR` to avoid confusion with the extant use of `ELECTRON_DIR` in unrelated `script/`s

ckerr: NVM, your new variable `GN_CHECK_DIR` is even better

script/gn-check.js
ckerr: Resolving the inconsistencies in a followup sounds good to me.

codebytere: https://github.com/electron/electron/pull/19857 🌟

script/gn-check.js
```diff
@@ -0,0 +1,25 @@
...
+      break
+    }
+  }
+}
+
+if (!GN_CHECK_DIR) {
+  throw new Error(`No viable out dir: one of Debug, Testing, or Release must exist.`)
+}
+
+const gnCheckString = `gn check ${GN_CHECK_DIR} //electron:electron_lib`
```
jkleinsc: We should run the same checks we do in CI:
https://github.com/electron/electron/blob/master/.circleci/config.yml#L264

codebytere: ooh good point, will do!

script/gn-check.js
```diff
@@ -0,0 +1,16 @@
...
+
+const { getOutDir } = require('./lib/utils')
+
+const OUT_DIR = getOutDir()
+if (!OUT_DIR) {
+  throw new Error(`No viable out dir: one of Debug, Testing, or Release must exist.`)
+}
+
+const gnCheckString = `
+gn check ../out/${OUT_DIR} //electron:electron_lib &&
```
MarshallOfSound: Can we use `child_process` to `spawnSync` these instead of executing raw bash?

package.json
```diff
@@ -113,6 +114,7 @@
       "remark -qf"
     ],
     "*.{gn,gni}": [
+      "npm run gn-check",
```
MarshallOfSound: This depends on header files as well so this should be put in a new block `*.{gn,gni,h,cc,mm}`

codebytere: i thought about this but i was concerned about the lag time - it's not awful but slow enough that i wanted it to take more minimal effect 🤔

ckerr:
@MarshallOfSound your turn :D

script/gn-check.js
```diff
@@ -0,0 +1,23 @@
...
+const gnCheckDirs = [
+  '//electron:electron_lib',
+  '//electron:electron_app',
+  '//electron:manifests',
+  '//electron/shell/common/api:mojo'
+]
+
+for (const dir of gnCheckDirs) {
+  const args = ['check', `../out/${OUT_DIR}`, dir]
+  const result = cp.spawnSync('gn', args, { stdio: 'inherit' })
```
zcbenz: We need to execute the `gn` in src instead of global PATH to avoid wrong results when switching to old branches, can you reference `script/lint.js` to set the required environment values?

ckerr: We still need to pass `env` into the spawnSync options to pick up the new & improved path

script/gn-check.js
```diff
@@ -0,0 +1,34 @@
...
```
ckerr:
```suggestion
  const result = cp.spawnSync('gn', args, { env, stdio: 'inherit' })
```

PR #19852, opened 2019-08-20
build: ninja count should be 2*cores + 2

MarshallOfSound:
Noticed while wondering why macOS CI was going slower than normal, with 8 core machines we should be using 18 processes.  With 16 core machines we should be using 34 processes.

Notes: no-notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,13 +91,13 @@ env-linux-medium: &env-linux-medium
   NUMBER_OF_NINJA_PROCESSES: 3
 
 env-linux-2xlarge: &env-linux-2xlarge
-  NUMBER_OF_NINJA_PROCESSES: 18
+  NUMBER_OF_NINJA_PROCESSES: 34
 
 env-machine-mac: &env-machine-mac
   NUMBER_OF_NINJA_PROCESSES: 6
 
 env-mac-large: &env-mac-large
-  NUMBER_OF_NINJA_PROCESSES: 10
+  NUMBER_OF_NINJA_PROCESSES: 18
 
 env-disable-crash-reporter-tests: &env-disable-crash-reporter-tests
   DISABLE_CRASH_REPORTER_TESTS: true


.circleci/config.yml
```diff
@@ -91,13 +91,13 @@ env-linux-medium: &env-linux-medium
   NUMBER_OF_NINJA_PROCESSES: 3
 
 env-linux-2xlarge: &env-linux-2xlarge
-  NUMBER_OF_NINJA_PROCESSES: 18
+  NUMBER_OF_NINJA_PROCESSES: 34
```
ckerr: 1. Where does the `2 * cores + 2` heuristic come from? Not disagreeing with it; just asking to learn

2. It seems kind of brittle to code the current value instead of the heuristic that was used to calculate it. Would it be possible to use something like this instead:

```sh
NUMBER_OF_CORES=`cat /proc/cpuinfo | grep processor | wc -l`
NUMBER_OF_NINJA_PROCESSES=$(expr $NUMBER_OF_CORES + $NUMBER_OF_CORES + 2)
```

IIUC this would work as a one-stop setting regardless off the size profile

MarshallOfSound: @ckerr The `2 * cores + 2` comes from ninja's built in default.  The reason we have to hard code this and not calculate on the fly is because the processor count in CircleCI is reported incorrectly which causes ninja to think it can spin up like 80 processes and they eventually deadlock.

So in answer to `2` we unfortunately have to hardcode these numbers.

MarshallOfSound:
Difference in reported build speed:

#### Linux x64 Testing

##### Before
`[18 processes, 16135/19547 @ 96.5/s : 167.121s]`

##### After
`[34 processes, 16043/19547 @ 133.5/s : 120.130s]`

Around a 40% improvement

#### Mac x64 Testiing

##### Before

`[10 processes, 12020/19029 @ 14.2/s : 849.090s]`

##### After

`[18 processes, 12020/19029 @ 28.2/s : 426.209s]`

Around a 100% improvement??!?!?!?!?!???

.circleci/config.yml
ckerr: The answer to `2` appears to be that we have to provide some numbers, but it's not clear from what you said that we have to hardcode them.

Do either `grep -c ^processor /proc/cpuinfo` or `nproc` return a usable number in these environments?

MarshallOfSound: @ckerr They return incorrect numbers, that's what made us hard code these in the first place.  Something about circleci's platform makes those numbers incorrect.

PR #19853, opened 2019-08-20
build: trim branch name before comparing to master

MarshallOfSound:
If I have to trim one more string....

Notes: no-notes


codebytere:
this needs backports, no?

PR #19857, opened 2019-08-21
build: dynamically calculate out dir

codebytere:
#### Description of Change

This PR updates our `out` directory subfolder calculation in `utils.js` dynamically instead of always falling back to `Debug` if `process.env.ELECTRON_OUT_DIR` is not set. Associated uses of the previous const export `OUT_DIR` haven been updated accordingly.

cc @ckerr @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/script/lib/utils.js
+++ b/script/lib/utils.js
@@ -23,13 +23,16 @@ function getElectronExec () {
   }
 }
 
-function getOutDir () {
+function getOutDir (shouldLog) {
   if (process.env.ELECTRON_OUT_DIR) {
     return process.env.ELECTRON_OUT_DIR
   } else {
     for (const buildType of ['Debug', 'Testing', 'Release']) {
       const outPath = path.resolve(SRC_DIR, 'out', buildType)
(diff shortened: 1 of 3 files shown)


script/lib/utils.js
```diff
@@ -23,6 +23,17 @@ function getElectronExec () {
   }
 }
 
+function getOutDir () {
+  if (process.env.ELECTRON_OUT_DIR) {
+    return process.env.ELECTRON_OUT_DIR
+  } else {
+    for (const buildType of ['Debug', 'Testing', 'Release']) {
+      const outPath = path.resolve(SRC_DIR, 'out', buildType)
+      if (fs.existsSync(outPath)) return buildType
```
jkleinsc: We should add console output here so that folks see what dir was automatically used.

codebytere: hmm, could we perhaps do this conditionally? Since this will be used in the GN check lint step, i don't imagine folks would want the extra console spew in their terminals every time 🤔

PR #19873, opened 2019-08-21
fix: mitigate white screen flash after occlusion by disabling compositor recycling 

### LINKED ISSUE #19858: White screen appears quickly when the app focus from background.
### Preflight Checklist
*  [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
*  [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
*  [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v6.0.3
* **Operating System:**
  * macOS 10.14.6
  * v1.8.3

### Expected Behavior
The same behavior as other APPs

### Actual Behavior
White screen appears quickly when the app focus from background, Ensure that the app is in the background for a while.

### To Reproduce
https://github.com/electron/electron-quick-start
Put the app in the background for a while and focus main window, the issue will be reproduced.

### Screenshots
Screen recording: https://s0.static.lunkr.cn/cab/publish/electron_bug_report.mov

### Additional Information
none

loc:
#### Description of Change

Fix #19858, caused by compositor recycling in response to OS occlusion events.

Compositor recycling is useful for Chrome because there can be many tabs and spinning up a compositor for each one would be costly. In practice, Chrome uses the parent compositor code path of browser_compositor_view_mac.mm; the NSView of each tab is detached when it's hidden and attached when it's shown. For Electron, there is no parent compositor, so we're forced into the "own compositor" code path, which seems to be non-optimal and pretty ruthless in terms of the release of resources. Electron has no real concept of multiple tabs per window, so it should be okay to disable this ruthless recycling altogether in Electron.

There's some risk:
a) it might degrade CPU performance
b) it might have a memory cost
c) I largely don't know what I'm doing

To satiate a), I loaded giphy.com before and after the patch with the window both visible and occluded. After the patch, the renderer and GPU processes still drop to zero, as we would hope.

No patch, visible:
without-patch-before

No patch, occluded:
without-patch-after

Patch, visible:
with-patch-before

Patch, occluded:
with-patch-after

Through inspection, I think I can allay concern b): the recycleable_compositor keeps around at least one compositor of the pool that it manages (per window). So, with this change, I don't think we're actually affecting anything. It's possible that detaching from the compositor frees up some surface buffers or something, but I imagine that's negligible.

For c) reviewers will have to weigh in with more context on potential risks to other features that I'm unaware of.

As a bonus, with this change, we're able to remove a workaround introduced in #18661. The occlusion event from macOS was errantly firing when fullscreening, causing the compositor to be recycled. Now that we don't do that, we no longer need the debounce.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed white flash after restoring an app from the background.


patches/chromium/disable_compositor_recycling.patch
```diff
@@ -0,0 +1,20 @@
...
+Compositor recycling is useful for Chrome because there can be many tabs and spinning up a compositor for each one would be costly. In practice, Chrome uses the parent compositor code path of browser_compositor_view_mac.mm; the NSView of each tab is detached when it's hidden and attached when it's shown. For Electron, there is no parent compositor, so we're forced into the "own compositor" code path, which seems to be non-optimal and pretty ruthless in terms of the release of resources. Electron has no real concept of multiple tabs per window, so it should be okay to disable this ruthless recycling altogether in Electron.
+
+diff --git a/content/browser/renderer_host/browser_compositor_view_mac.mm b/content/browser/renderer_host/browser_compositor_view_mac.mm
+index 59e58d693c971742951434f6582140d9179235f2..135e7a384a560f55e5201f108fe1ac2db621fbca 100644
+--- a/content/browser/renderer_host/browser_compositor_view_mac.mm
++++ b/content/browser/renderer_host/browser_compositor_view_mac.mm
+@@ -209,7 +209,7 @@
+ }
+ 
+ void BrowserCompositorMac::SetRenderWidgetHostIsHidden(bool hidden) {
```
codebytere: In what circumstances is this being called with `hidden: true`?

I might say it would be more effective to consider changing `hidden` at the callsite(s) as opposed to unconditionally changing it inside the `SetRenderWidgetHostIsHidden` impl, as that might give us a better opportunity to mitigate risk and understand when this is being set to cause this specific error.

MarshallOfSound: hidden is true when the view is marked as occluded.  The goal of this change was to prevent the compositor recycle while still allowing every other thing that handles that observer call to know the window is occluded.

This is why we tested the CPU usage when backgrounded.  We wanted occlusion detection to still work, just this one observer to stop doing it's thing

deepak1556:
> Electron has no real concept of multiple tabs per window, so it should be okay to disable this ruthless recycling altogether in Electron.

While this is true in general, @loc have you tested this change's impact for macOS native tabs code path ?

/cc requesting review from @brenca @zcbenz for compositor code path.

MarshallOfSound:
> have you tested this change's impact for macOS native tabs code path ?

We hadn't but now I have and the same got the results as our other tests.  Renderers backgrounded in native tabs don't white flash but they do drop to 0% CPU usage when occluded 👍

MarshallOfSound:
This appears to have destabilized a visibilityState test.  Will investigate today 😄

MarshallOfSound:
> This appears to have destabilized a visibilityState test. Will investigate today 😄

So my debounce patch was hiding the fact that this test has been broken for a while.  I've disabled it in this PR to unblock it and will investigate after it has landed.

poiru:
@MarshallOfSound @zcbenz Unfortunately this patch caused a regression with BrowserViews. In our app, switching tabs switches the visible BrowserView. Due to a change in this PR, the BrowserViews are momentarily blank when you switch to them. Prior to this change and e.g. in Chrome (the browser), switching to a tab makes it immediately visible without any sort of flicker.

Reverting a bit of this PR restores the old and correct behavior:

```patch
 void BrowserCompositorMac::SetRenderWidgetHostIsHidden(bool hidden) {
-  render_widget_host_is_hidden_ = false;
+  render_widget_host_is_hidden_ = hidden;
   UpdateState();
 }
```

I'm not sure what to do here - should we revert this PR or should we try to find another solution? This is a critical issue for us, we cannot upgrade to 6.x and beyond due to this.

MarshallOfSound:
> should we revert this PR or should we try to find another solution? 

IMO no, this fixes a pretty nasty bug which affects the _majority_ of users.  I'm unsure why it would negatively impact BrowserView users but that shouldn't mean we revert this.

I'd be open to a forward-fix PR here but I don't have time to dig into this right now.

poiru:
@MarshallOfSound Would you be willing to accept a patch that introduces a opt-in command-line flag to restore the old behavior until we have a proper solution? We'd like to ship Electron 6 ASAP, but this is blocking us.

PR #19881, opened 2019-08-22
fix: notify views of content view size change (4-2-x)

zcbenz:
Backport of https://github.com/electron/electron/pull/19878

Notes: Fix web page displayed with offset when using native tab on macOS.


zcbenz:
This PR requires approval from @electron/wg-releases.

PR #8047, opened 2016-11-22
Fix incorrect window size after restore on Windows

MarshallOfSound:
These reverts #7766 

It fixes #7630 in a different way and also fixes #7951.

Issue #7951, opened 2016-11-12
Windows: Restore (after minimize) to wrong window size


simast:
* Electron version: 1.4.6
* Operating system: Windows 10

### Expected behavior

After minimizing the window it should restore to the same size.

### Actual behavior

After minimizing the window it is restored to a larger width.

### How to reproduce

The issue started with Electron 1.4.5 and will only appear when you use `minWidth` property in browser window config. As an example:

```javascript
const windowConfig = {
	title: "Test app",
	width: 800,
	height: 600,
	minWidth: 800
};

const mainWindow = new BrowserWindow(windowConfig);
```

MarshallOfSound:
Interesting on my setup I get the following results.

| Config `width` | Config `minWidth` | `width` after restore |
|-----------------|---------------------|------------------------|
| 800 | 800 | 944 |
| 500 | 500 | 644 |
| 700 | 500 | 700 |

Looks to incorrectly have a minWidth bound of 144 too big.  I'll look into this more

MarshallOfSound:
@zcbenz I started fixing this then remembered I saw similar code to what I was writing before.  It was removed in a recent PR.

https://github.com/electron/electron/pull/7766/files#diff-a2a9d45b70c38bec45928975c3a3a6a1L214

Perhaps we should roll that PR back and attempt to fix the issue in #7630 in a different way that is consistent with the fixes that were deleted in that PR that we really still need.

zcbenz:
@MarshallOfSound Sounds good to me to revert the PR.

Luci404:
Why is this still a problem?

CahierX:
This problem is still present on Window11

Issue #2498, opened 2015-08-13
[Windows] Frameless window size increase on a maximize, unmaximize sequence


etiktin:
When I have a frameless resizeable window with `min-width` and `min-height` set, the width and height increases after returning from maximize. The width increases by 6 px while the height increases by 8 px.

How to reconstruct:
1) Create a window using:
```javascript
new BrowserWindow({
  'frame': false,
  'resizable': true,
  'width': 1020,
  'height': 720,
  'min-width': 1000,
  'min-height': 700,
}); 
```
2) Open the devtools.
3) When you start the program you should see something like this (1020x720):
![initialstate](https://cloud.githubusercontent.com/assets/4381579/9259327/1bc9823e-4209-11e5-9f06-0e58c1e1c4f1.png)
4) Resize the window to it's minimum size (1000x700):
![resizetominimal](https://cloud.githubusercontent.com/assets/4381579/9259355/4f9c5fd2-4209-11e5-8b58-650cd5413323.png)
5) Maximize the window using `WinKey+Up` (for me this was 1924x1042 which is already weird because the screen size is 1920x1080 so the width already has 4 extra pixels):
![maximize](https://cloud.githubusercontent.com/assets/4381579/9259422/c58fca9e-4209-11e5-8e82-491ea0a31268.png)
6) Unmaximize using `WinKey+Down` (100*6*x70*8*):
![afterunmaximize](https://cloud.githubusercontent.com/assets/4381579/9259511/41a76786-420a-11e5-95b6-165ff9e0a86f.png)

If you skip the 4th step (resize to minimum), the size still changes but from my experiments it seems only the height will increase (it increases by 8 pixels).

If I do the same without setting `min-width` and `min-height` it seems to work as expected (returns to the same size).

In an app I'm working on, there are a couple of scenarios when I open an iframe that is set to the minimum size. It looks good in the initial state and in the maximize state, but once I unmaximize the window instead of returning to it's last size, it has this extra pixels that show up around the iframe.

I tested it with Windows 8.1 64 bit using Electron v0.30.4.

etiktin:
@zcbenz, could this be a result of [this](https://github.com/atom/electron/commit/0ab32bfe17c81a03c91f79088e71f6c58b2306fb) commit?

etiktin:
@zcbenz, could this issue be a result of [this](https://github.com/atom/electron/commit/0ab32bfe17c81a03c91f79088e71f6c58b2306fb) commit?

[...]

PR #19886, opened 2019-08-22
fix: don't call SetBounds on restore

### LINKED ISSUE #13043: Window size changes when being moved if resizable is false
* Electron Version: 2.0.0
* Last known working Electron version: before commit bb5eecc16cff8053c7b04e0b9396803fbbb8c4df

**Expected Behavior**
Width and height of a window should not change while being dragged 

**Actual behavior**
Width and height become smaller.

**To Reproduce**
Create a window with:
  var mainWindow = new BrowserWindow({width: 1024, height: 880, x: 30, y: 30,
            resizable: false })

Then drag the window.  Width and height become smaller than specified (1024, 880)

**Additional Information**
NativeWindowViews::SetBounds calls SetMin/MaxSize(bounds) when resizable_ is false.  bounds get adjusted to smaller  numbers by WindowBoundsToContentBounds.

brenca:
#### Description of Change

This is a revert of #8047, the issues mentioned there are not reproducible after the revert (in fact #7951 resurfaced, and this fixes that). Since the issue mentioned in the comments (#2498) is about frameless window sizing and the PR I created to improve that landed in `6-0-x`, I'm only backporting until that to make sure not to re-introduce the original bug.

Fixes #13043.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fix a bug where an unresizable window would shrink when restored after being minimized on Windows.

### DIFF
--- a/shell/browser/native_window_views.cc
+++ b/shell/browser/native_window_views.cc
@@ -305,7 +305,6 @@ NativeWindowViews::NativeWindowViews(const mate::Dictionary& options,
     last_window_state_ = ui::SHOW_STATE_FULLSCREEN;
   else
     last_window_state_ = ui::SHOW_STATE_NORMAL;
-  last_normal_bounds_ = GetBounds();
 #endif
 
 #if defined(OS_LINUX)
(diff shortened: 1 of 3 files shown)


zcbenz:
There is conflict needs to be resolved.

brenca:
@zcbenz I resolved the conflict!

PR #19889, opened 2019-08-22
feat: add session.downloadURL()

miniak:
#### Description of Change
Add `session.downloadURL()` allowing to trigger downloads without a `BrowserWindow`

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `session.downloadURL()` allowing to trigger downloads without a `BrowserWindow`.

### DIFF
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -408,6 +408,17 @@ Returns `String` - The user agent for this session.
 
 Returns `Promise<Buffer>` - resolves with blob data.
 
+#### `ses.downloadURL(url)`
+
+* `url` String
+
+Initiates a download of the resource at `url`.
+The API will generate a [DownloadItem](download-item.md) that can be accessed
+with the [will-download](#event-will-download) event.
+
+**Note:** This does not perform any security checks that relate to a page's origin,
+unlike [`webContents.downloadURL`](web-contents.md#contentsdownloadurlurl).
+
 #### `ses.createInterruptedDownload(options)`
 
 * `options` Object
--- a/shell/browser/api/atom_api_session.cc
+++ b/shell/browser/api/atom_api_session.cc
@@ -19,6 +19,7 @@
 #include "chrome/browser/browser_process.h"
 #include "chrome/common/pref_names.h"
 #include "components/download/public/common/download_danger_type.h"
+#include "components/download/public/common/download_url_parameters.h"
 #include "components/prefs/pref_service.h"
 #include "components/prefs/value_map_pref_store.h"
 #include "components/proxy_config/proxy_config_dictionary.h"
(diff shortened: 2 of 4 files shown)


docs/api/session.md
```diff
@@ -386,6 +386,14 @@ Returns `String` - The user agent for this session.
 
 Returns `Promise<Buffer>` - resolves with blob data.
 
+#### `contents.downloadURL(url)`
```
codebytere: `contents.downloadURL(url)` typo, perhaps?

miniak: fixed

deepak1556:
@miniak can you redirect the implementation of `WebContents::DownloadURL` to be based on this session api ?

miniak:
@deepak1556 `WebContents::DownloadURL()` associates the download with the main frame, which enforces some security checks, etc. I think the implementation has to be kept separate. Also it's not that much extra code, it's just few lines that trigger the download.

miniak:
@deepak1556 https://cs.chromium.org/chromium/src/components/download/public/common/download_url_parameters.h?g=0&l=69-81

deepak1556:
> associates the download with the main frame of the webContents, which enforces some security checks, etc

Yup but those are configurable via the download parameters which could be passed into, something like

```
void DownloadURL(const GURL& url,
                          std::unique_ptr<download::DownloadUrlParameters> params) {}
```

Having these api separate sounds good too, but I would like to maintain single point of implementation that uses content api if possible , so that we don't update multiple places if chromium decides to break api signature.

miniak:
@deepak1556 I've updated the test to use `session.defaultSession` instead of `w.webContents.session` to make it explicit that the API does not need a window

miniak:

nornagon:
Should this replace `webContents.downloadURL()`? i.e. should we deprecate & remove that function with this as the replacement?

MarshallOfSound:
@nornagon See https://github.com/electron/electron/pull/19889#issuecomment-524011431

docs/api/session.md
```diff
@@ -386,6 +386,14 @@ Returns `String` - The user agent for this session.
 
 Returns `Promise<Buffer>` - resolves with blob data.
 
+#### `ses.downloadURL(url)`
+
+* `url` String
+
+Initiates a download of the resource at `url`.
+The API will generate a [DownloadItem](download-item.md) that can be accessed
+with the [will-download](#event-will-download) event.
```
nornagon: Document how this differs from a download initiated from a frame?

miniak: I am trying to figure it out. Those checks might be Chromium browser specific.

miniak: the request headers sent by Electron are the same regardless whether the download is associated with a frame or not (no referer)
```
Connection: keep-alive
Sec-Fetch-Site: none
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) divergent-garbage-integrate/1.0.0 Chrome/78.0.3892.0 Electron/8.0.0-nightly.20190823 Safari/537.36
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US
```

miniak: https://cs.chromium.org/chromium/src/content/browser/download/download_manager_impl.cc?sq=package:chromium&g=0&l=1288-1297
```cpp
  // Check if the renderer is permitted to request the requested URL.
  if (params->render_process_host_id() >= 0 &&
      !DownloadRequestUtils::IsURLSafe(params->render_process_host_id(),
                                       params->url())) {
    CreateInterruptedDownload(
        std::move(params),
        download::DOWNLOAD_INTERRUPT_REASON_NETWORK_INVALID_REQUEST,
        weak_factory_.GetWeakPtr());
    return;
  }
```

Issue #19795, opened 2019-08-16
Can't show file chooser dialog when clicking <input type="file"> after canceled the dialog.


agata:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.0.0-beta.3
* **Operating System:**
  * macOS 10.14.5
* **Last Known Working Electron version:**
  * 6.0.2

### Expected Behavior
Show file chooser dialog when clicking input type="file" after canceled the dialog.

### Actual Behavior
Can't show file chooser dialog when clicking input type="file" after canceled the dialog.

### To Reproduce

1. Clone and Install source code.
```sh
$ git clone https://github.com/agata/electron-quick-start.git -b bug-for-file-chooser-after-canceled
$ npm install
$ npm start || electron .
```
2. Click <input type="file"> -> Show file chooser dialog.
3. Click cancel button of the dialog -> Close the dialog
4. Click <input type="file"> again -> Can't show file chooser dialog.

### Screenshots
![bug-for-file-chooser-after-canceled](https://user-images.githubusercontent.com/40473/63176157-60916900-c080-11e9-8715-f263319f0e68.gif)

### Additional Information
WebView has same problem.

PR #19897, opened 2019-08-22
fix: crash on input file handler dialog

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/19795.

This was crashing on the following DCHECK:
```
DCHECK(was_file_select_listener_function_called_)
          << "Should call either FileSelectListener::FileSelected() or "
             "FileSelectListener::FileSelectionCanceled()";
```

We introduced proper support for the webkitdirectory input attr in https://github.com/electron/electron/pull/18343, which required us to override much of the functionality previously [handled by Chromium](https://cs.chromium.org/chromium/src/content/browser/frame_host/render_frame_host_impl.cc). We told the ListenerProxy that files were selected with
```
listener_->FileSelected(std::move(file_info), base_dir, mode_);
```
but when the cancellation button was clicked, we never explicitly told the `ListenerProxy` that the dialog was cancelled and so the DCHECK was hit. This fixes that issue for both save and open dialogs without altering existing functionality.

cc @MarshallOfSound @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash on `<input type="file">` dialog cancellation.

### DIFF
--- a/shell/browser/web_dialog_helper.cc
+++ b/shell/browser/web_dialog_helper.cc
@@ -153,10 +153,12 @@ class FileSelectHelper : public base::RefCounted<FileSelectHelper>,
                                                 paths[0].DirName());
         }
       }
+      // We should only call this if we have not cancelled the dialog
+      if (ready_to_call_listener)
+        OnFilesSelected(std::move(file_info), lister_base_dir_);
+    } else {
+      OnSelectionCancelled();
     }
-
-    if (ready_to_call_listener)
-      OnFilesSelected(std::move(file_info), lister_base_dir_);
   }
 
   void OnSaveDialogDone(mate::Dictionary result) {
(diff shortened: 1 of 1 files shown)


deepak1556:
The patch LGTM, the actual point of failure was not the DCHECK, which wouldn't be triggered with release builds but rather the callback to blink was never called in this situation.

shell/browser/web_dialog_helper.cc
```diff
@@ -331,6 +338,7 @@ void WebDialogHelper::EnumerateDirectory(
     content::WebContents* web_contents,
     std::unique_ptr<content::FileSelectListener> listener,
     const base::FilePath& dir) {
+  LOG(INFO) << "WebDialogHelper::EnumerateDirectory";
```
deepak1556: Can you remove this log ?

codebytere: oops, yes!

codebytere: done, & updated it to make it slightly more robust in the case that `listener_` was not initialized

shell/browser/web_dialog_helper.cc
```diff
@@ -153,10 +153,12 @@ class FileSelectHelper : public base::RefCounted<FileSelectHelper>,
                                                 paths[0].DirName());
         }
       }
+      // We should only call this if we have not cancelled the dialog
+      if (ready_to_call_listener)
+        OnFilesSelected(std::move(file_info), lister_base_dir_);
+    } else {
```
MarshallOfSound: This indentation is funky 🤔

codebytere: resolved in dm - it just looks funny but is correct 😆

ckerr: two-birds-one-stone-dept: the indentation will look less funky if we use positive rather than negative logic in the surrounding if-else:

```cc
if (cancelled) {
  OnSelectionCancelled();
} else {
  ...
  if (ready_to_call_listener)
    OnFIlesSelected(std::move(file_info), lister_base_dir_);
}

File: shell/browser/web_dialog_helper.cc [hunk 3]
```diff
@@ -171,8 +173,11 @@ class FileSelectHelper : public base::RefCounted<FileSelectHelper>,
             blink::mojom::NativeFileInfo::New(
                 path, path.BaseName().AsUTF16Unsafe())));
       }
+      // We should only call this if we have not cancelled the dialog
+      OnFilesSelected(std::move(file_info), base::FilePath());
+    } else {
+      OnSelectionCancelled();
```
ckerr: Same as above wrt removing negative logic now that we have actions for both the cancelled and not-cancelled state

```cc
if (cancelled) {
  OnSelectionCancelled();
} else {
  ...
}
```

PR #19903, opened 2019-08-22
feat: route frame based permission checks through our permission check handler

### LINKED ISSUE #14544: [C67] Review synchronous permission check implementation
See https://github.com/electron/electron/blob/54ad38526095f4a415ad75a1757feae4336c9db4/atom/browser/atom_permission_manager.cc#L213 and https://github.com/electron/electron/blob/54ad38526095f4a415ad75a1757feae4336c9db4/atom/browser/atom_permission_manager.cc#L246

Should probably call out to https://github.com/electron/electron/blob/54ad38526095f4a415ad75a1757feae4336c9db4/atom/browser/atom_permission_manager.cc#L231 and be documented as such

MarshallOfSound:
Bringing back #14544

Fixes #14544

Notes: Additional permission checks are now routed through `session.setPermissionCheckHandler`.  These include `Notification.permission`, and `permission.query`.  Please note that the `webContents` parameter to the check handler can now be null.

### DIFF
--- a/docs/breaking-changes.md
+++ b/docs/breaking-changes.md
@@ -14,6 +14,28 @@ This document uses the following convention to categorize breaking changes:
 
 ## Planned Breaking API Changes (13.0)
 
+### API Changed: `session.setPermissionCheckHandler(handler)`
+
+The `handler` methods first parameter was previously always a `webContents`, it can now sometimes be `null`.  You should use the `requestingOrigin`, `embeddingOrigin` and `securityOrigin` properties to respond to the permission check correctly.  As the `webContents` can be `null` it can no longer be relied on.
+
+```js
+// Old code
+session.setPermissionCheckHandler((webContents, permission) => {
+  if (webContents.getURL().startsWith('https://google.com/') && permission === 'notification') {
+    return true
+  }
+  return false
+})
+
+// Replace with
+session.setPermissionCheckHandler((webContents, permission, requestingOrigin) => {
+  if (new URL(requestingOrigin).hostname === 'google.com' && permission === 'notification') {
+    return true
+  }
+  return false
+})
+```
+
 ### Removed: `shell.moveItemToTrash()`
 
 The deprecated synchronous `shell.moveItemToTrash()` API has been removed. Use
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -500,6 +500,7 @@ win.webContents.session.setCertificateVerifyProc((request, callback) => {
     * `pointerLock` - Request to directly interpret mouse movements as an input method. Click [here](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API) to know more.
     * `fullscreen` - Request for the app to enter fullscreen mode.
     * `openExternal` - Request to open links in external applications.
+    * `unknown` - An unrecognized permission request
   * `callback` Function
     * `permissionGranted` Boolean - Allow or deny the permission.
   * `details` Object - Some properties are only available on certain permission types.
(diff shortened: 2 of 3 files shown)


MarshallOfSound:
This is tagged as `semver/major` because `webContents` can be `null` in the permission check handler now.  This can't be deprecated, but can only go into a major release (8.0.0)

nornagon:
Could we perhaps make this not a breaking change by having a default implementation of `PermissionCheckHandler` which returns whatever the last return value from the `PermissionRequestHandler` was?

i.e.

```
session.setPermissionRequestHandler((webContents, permission, callback) => {
  if (can(webContents, permission)) callback(true)
})

// => subsequent permission checks return `true` by default
// installing your own PermissionCheckHandler overrides this behavior
```

Does Chromium ever call the check before the request..?

Another way to handle this API would be something like:

```
session.setPermissionRequestHandler((webContents, permission, callback) => {
  webContents.grantPermission(permission)
  callback(true)
})

// => the permission has been set to 'enabled'. At any future time `revokePermission` can be called to revoke permission. `grantPermission` can also be called proactively.
```

MarshallOfSound:
@nornagon This isn't a breaking change because of that.  The only reason this is marked as `semver/major` is because the `webContents` parameter to an existing event can now be `null`.

nornagon:
> @nornagon This isn't a breaking change because of that. The only reason this is marked as `semver/major` is because the `webContents` parameter to an existing event can now be `null`.

I'm pretty sure requiring a new permission handler for e.g. web notifications or fullscreen is also a breaking change.

MarshallOfSound:
@nornagon it doesn't require a new permission handler, the handler defaults to granted so it's not a breaking change

nikitakot:
@nornagon @MarshallOfSound Hi there, what is the status of this PR? When it can be ready? Also will it be available in electron 8 and 9? Thank you.

nornagon:
@nikitakot as you can see from the comments in this PR:

1. It's old, and hasn't been recently worked on.
2. It's a breaking change.

We do not merge breaking changes to old branches. The soonest this could possibly be available is Electron 12, as Electron 11 is a quiet release.

docs/breaking-changes.md
```diff
@@ -14,6 +14,28 @@ This document uses the following convention to categorize breaking changes:
...
+// Old code
+session.setPermissionCheckHandler((webContents, permission) => {
+  if (webContents.getURL().startsWith('https://google.com/') && permission === 'notification') {
+    return true
+  }
+  return false
+})
+
+// Replace with
+session.setPermissionCheckHandler((webContents, permission, requestingOrigin) => {
```
nornagon, line 31: This is called `requestingUrl` in the other docs. Is it the URL or the origin?

MarshallOfSound: `requestingUrl` is different.  This param is `requestingOrigin`, the URL is an optional property on the `details` object

nornagon (approved):

ChALkeR:
@MarshallOfSound This PR (commit 771e34a53a03bfaa8299a0eb28589baeae756a7b) introduced an [undocumented](https://www.electronjs.org/docs/api/session#sessetpermissioncheckhandlerhandler) `window-placement` event in permission check handler:

To reproduce:
```js

app.whenReady().then(() => {
  const win = new BrowserWindow({ width: 800, height: 600 })

  const ses = win.webContents.session
  console.log(ses.getUserAgent())
  ses.setPermissionRequestHandler((...args) => console.log('request', args))
  ses.setPermissionCheckHandler((...args) => console.log('check', args))

  win.loadURL('about:blank')
})
```

To trigger the event do any screen configuration changes: e.g. connect or disconnect an external display, or change screen rotation or resolution, or just turn the monitor off and on again. Even that might not be needed, for some setups this just happens on launch apparently.

This happens even if `--disable-blink-features=WindowPlacement` is added to the options.

PR #19904, opened 2019-08-23
fix: i18n of gtk msgbox buttons

ckerr:
#### Description of Change

Does for GTK messageboxes what #19756 (12df0e8) did for GTK file dialogs: fixes the i18n of the stock buttons in a way that doesn't break the buttons' keyboard shortcuts.

CC @codebytere 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed i18n of GTK messageboxes' stock buttons.

### DIFF
--- a/shell/browser/ui/message_box_gtk.cc
+++ b/shell/browser/ui/message_box_gtk.cc
@@ -4,8 +4,6 @@
 
 #include "shell/browser/ui/message_box.h"
 
-#include <glib/gi18n.h>
-
 #include "base/callback.h"
 #include "base/strings/string_util.h"
 #include "base/strings/utf_string_conversions.h"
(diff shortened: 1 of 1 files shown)


shell/browser/ui/message_box_gtk.cc
```diff
@@ -33,6 +31,27 @@ MessageBoxSettings::~MessageBoxSettings() = default;
 
 namespace {
 
+// Copied from L40-L55 in
+// https://cs.chromium.org/chromium/src/chrome/browser/ui/libgtkui/select_file_dialog_impl_gtk.cc
```
codebytere: i don't love that we straight up copy this exact big block twice...can we somehow reuse this across both instances perhaps? 🤔

ckerr: Good suggestion! Fixed 4ac3f14258becac3ed2f1bf07e659755ad24c504.

shell/browser/ui/gtk_util.cc
```diff
@@ -0,0 +1,36 @@
...
+
+#include "shell/browser/ui/gtk_util.h"
+
+#include <gtk/gtk.h>
+
+namespace gtk_util {
+
+// Copied from L40-L55 in
+// https://cs.chromium.org/chromium/src/chrome/browser/ui/libgtkui/select_file_dialog_impl_gtk.cc
+#if GTK_CHECK_VERSION(3, 90, 0)
```
MarshallOfSound: This whole thing needs to be linux OS ifdef checked

shell/browser/ui/gtk_util.cc
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2019 GitHub, Inc.
+// Use of this source code is governed by the MIT license that can be
+// found in the LICENSE file.
+
+#include "shell/browser/ui/gtk_util.h"
+
+#include <gtk/gtk.h>
```
MarshallOfSound: In particular this header

shell/browser/ui/gtk_util.cc
codebytere: we may also want to put this alongside similar utils in `shell/common` i think!

zcbenz: GTK is only used by Linux so we should avoid having it compiled for other platforms at all, we can add some rules in GN or renamed the file to `util_gtk`.

And GTK is only available in the main process, putting it under `shell/common` might not be a good idea.

PR #19928, opened 2019-08-24
fix: handle WM_GETMINMAXINFO instead of letting chromium do it

### LINKED ISSUE #19423: Electron 6, maximize/restore not working correctly
### Preflight Checklist
* [X ] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [ X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [X ] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 6.0.0-beta.15
* **Operating System:**
  * Windows 10 64bit
  * 5.0.8

### Expected Behavior
The window restores its size to what it used to be

### Actual Behavior
The window is shown in the smallest size

### To Reproduce
1) Maximize Window
2) Minimize Window
3) Restore Window (click on the icon in the taskbar)
4) Unmaximize Window

### Screenshots
grafik

### LINKED ISSUE #19816: Electron 6 - restore after minimize has wrong window size

### Issue Details

* **Electron Version:**
  * `6.0.2`
* **Operating System:**
  * Windows 10 (1903)
  * `5.0.7`

### Expected Behavior
1. App is not maximized 
2. click minimize 
3. click on taskbar icon
4. **app restores to original size**

### Actual Behavior
1. App is not maximized 
2. click minimize 
3. click on taskbar icon
4. **screen flickers as the app returns from _minimum allowed window size_ to the original (not maximized) size.**

### To Reproduce
I logged the behavior:
```ts
  win.on('minimize', (e) => {
    console.log('minimizing');
    const bounds: any = win.getBounds();
    console.log(bounds);
  });

  win.on('restore', (e) => {
    console.log('restoring');
    const bounds: any = win.getBounds();
    console.log(bounds);
  });
```

Result:
```
minimizing
{ x: 100, y: 200, width: 1400, height: 800 }
restoring
{ x: 100, y: 200, width: 420, height: 250 }
```

The dimensions come from 
```ts
  win = new BrowserWindow({
    ...
    minWidth: 420,
    minHeight: 250,
    frame: false // this line might be relevant for the current issue
```

I have the problem happening in my app, in `master` - tracking with https://github.com/whyboris/Video-Hub-App/issues/260

Can be reproduced by cloning the repository, running `npm install`, having _Angular CLI_ installed (`npm i -g @angular/cli`), and running the app `npm start`. Needless to say the error occurs in the built version of the app (`npm run electron:windows`).

_Note:_ the error does _not_ occur when the window is _maximized_, minimized, and restored.

Error might not occur on Mac -- can't be sure because the Mac has a stupid minimize/maximize animation 🤷‍♂ 

Thank you _Electron Team_ for your amazing work 🙇 my app would be impossible without you ❤️ 

### Possibly related tickets:
https://github.com/electron/electron/issues/19423
https://github.com/electron/electron/issues/13043
https://github.com/electron/electron/issues/16900

### Possible solution:
https://github.com/electron/electron/issues/19423#issuecomment-521621266

brenca:
#### Description of Change

There is a workaround I introduced that's no longer needed and is causing bugs, so this PR removes that. We had that workaround because during the `WM_GETMINMAXINFO` message Windows reports a position that's incorrect (somewhere around -32000 for both coordinates), and the handler in chromium calculates the scale factor for the window based on that position, which will be incorrect if you have more than one monitor (the leftmost will be used).

To work around this, we instead handle the `WM_GETMINMAXINFO` message ourselves, similarly to how chromium does it, but without a size compensation that's not required for us (it compensates for the window frame size, but we expect the minimum and maximum sizes to apply to the window anyway by default, and that compensation code is from 2011, doesn't care about scale factors, and probably should be removed/corrected in chromium anyway).

I'm going to look into upstreaming this new workaround/fix so that we can later remove it altogether.

Depends on #19886. 

Fixes #19423.
Fixes #19816. 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a bug where windows would sometimes shrink to 0 size after being restored on Windows.

### DIFF
--- a/shell/browser/native_window_views.h
+++ b/shell/browser/native_window_views.h
@@ -244,8 +244,6 @@ class NativeWindowViews : public NativeWindow,
 
   ui::WindowShowState last_window_state_;
 
-  gfx::Rect last_normal_placement_bounds_;
-
   // There's an issue with restore on Windows, that sometimes causes the Window
   // to receive the wrong size (#2498). To circumvent that, we keep tabs on the
   // size of the window while in the normal state (not maximized, minimized or
--- a/shell/browser/native_window_views.h
+++ b/shell/browser/native_window_views.h
@@ -244,6 +244,8 @@ class NativeWindowViews : public NativeWindow,
 
   ui::WindowShowState last_window_state_;
 
+  gfx::Rect last_normal_placement_bounds_;
+
   // There's an issue with restore on Windows, that sometimes causes the Window
   // to receive the wrong size (#2498). To circumvent that, we keep tabs on the
   // size of the window while in the normal state (not maximized, minimized or
--- a/shell/browser/native_window_views_win.cc
+++ b/shell/browser/native_window_views_win.cc
@@ -335,26 +335,6 @@ bool NativeWindowViews::PreHandleMSG(UINT message,
 
       return false;
     }
-    case WM_GETMINMAXINFO: {
-      WINDOWPLACEMENT wp;
-      wp.length = sizeof(WINDOWPLACEMENT);
-
-      // We do this to work around a Windows bug, where the minimized Window
-      // would report that the closest display to it is not the one that it was
-      // previously on (but the leftmost one instead). We restore the position
-      // of the window during the restore operation, this way chromium can
-      // use the proper display to calculate the scale factor to use.
-      if (!last_normal_placement_bounds_.IsEmpty() &&
-          GetWindowPlacement(GetAcceleratedWidget(), &wp)) {
-        last_normal_placement_bounds_.set_size(gfx::Size(0, 0));
-        wp.rcNormalPosition = last_normal_placement_bounds_.ToRECT();
-        SetWindowPlacement(GetAcceleratedWidget(), &wp);
-
-        last_normal_placement_bounds_ = gfx::Rect();
-      }
-
-      return false;
-    }
     case WM_NCCALCSIZE: {
       if (!has_frame() && w_param == TRUE) {
         NCCALCSIZE_PARAMS* params =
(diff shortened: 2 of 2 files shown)


whyboris:
Thank you @brenca for the PR -- it will likely fix my issue: https://github.com/whyboris/Video-Hub-App/issues/260 🙇

brenca:
@ckerr thanks for the quick review! I realized that removing the workaround does re-introduce a bug, so I changed the code to address that bug and still fix the issues mentioned above.

I'll do some additional testing to make sure things work alright.

brenca:
@ckerr I added some explanatory comments and made some changes, can I get another review from you?

whyboris:
Thank you @brenca for working on this bugfix 🙇 . Your PR probably closed the other bug, but mine persists:
https://github.com/electron/electron/issues/19816#issuecomment-526161552

whyboris:
False alarm -- seems like this PR has _not_ been merged into `6.0.5` but instead will be released later: https://github.com/electron/electron/issues/19816#issuecomment-526423893 👌

carterbs:
@codebytere this PR is causing issues for us on Electron 6 and 7. When we call setBounds on a hidden window, and then show that window on a monitor with a different display scale than the primary....it's the wrong size. But Electron thinks it's the correct size. We've narrowed it down to this PR causing the bad behavior. 

The reason we do this is that we have menus that are hidden, and change size based on the contents within the menus. So this bug makes our app look awful in the latest electron versions.

I've reached out to @brenca on slack to see if he has any ideas of how to fix this. It's presently preventing us from keeping up with the latest electron release.

sgd2z:
> @codebytere this PR is causing issues for us on Electron 6 and 7. When we call setBounds on a hidden window, and then show that window on a monitor with a different display scale than the primary....it's the wrong size. But Electron thinks it's the correct size. We've narrowed it down to this PR causing the bad behavior.
> 
> The reason we do this is that we have menus that are hidden, and change size based on the contents within the menus. So this bug makes our app look awful in the latest electron versions.
> 
> I've reached out to @brenca on slack to see if he has any ideas of how to fix this. It's presently preventing us from keeping up with the latest electron release.

After some testing, I found that I can only reproduce when resizable is false. I'm working on a reproducer.

PR #15229, opened 2018-10-17
chore: don't set allow_new_privs in new processes

nornagon:
#### Description of Change
WIP because I'm not sure if this breaks anything, and I'm hoping CI will tell us. cc @zcbenz 

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

#### Release Notes

Notes: no-notes

PR #19953, opened 2019-08-26
fix: allow unsandboxed renderers to request new privileges

### LINKED ISSUE #18521: Executing SUID binaries not possible in v5
Since v5, it's no longer possible to start setuid binaries from within an Electron renderer process - even with sandboxing disabled.

nornagon:
#### Description of Change
This used to be set for all processes, but was removed in #15229. This re-enables `allow_new_privs` but only for unsandboxed renderers. All other child processes will continue to have the default behaviour, which is to prevent new privileges via `prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0)` (see [prctl(2)](http://man7.org/linux/man-pages/man2/prctl.2.html)).

Fixes #18521.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed an issue that prevented setuid binaries from being launched from the renderer process on Linux.

### DIFF
--- a/spec/node-spec.js
+++ b/spec/node-spec.js
@@ -155,6 +155,19 @@ describe('node feature', () => {
         })
       })
     })
+
+    describe('child_process.exec', () => {
+      (process.platform === 'linux' ? it : it.skip)('allows executing a setuid binary from non-sandboxed renderer', () => {
+        // Chrome uses prctl(2) to set the NO_NEW_PRIVILEGES flag on Linux (see
+        // https://github.com/torvalds/linux/blob/40fde647cc/Documentation/userspace-api/no_new_privs.rst).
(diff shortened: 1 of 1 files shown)


the-j0k3r:
> A maintainer has manually backported this PR to "6-0-x", please check out #20023

Any chance this will be backported to Electron 5.0.13? Seems there is a regression for this see https://github.com/bus-stop/terminus/issues/84

Atom 1.47.0 latest version is using electron 5.0.13 which fails, earlier Atom 1.46.0 with electron 4.2.7 worked.

the-j0k3r:
Has anyone checked this fix exists in newer electron version greater than 7.x?

Issue #19932, opened 2019-08-25
nativeTheme API extensions


MarshallOfSound:
In Electron 7 there is a new API called [`nativeTheme`](https://github.com/electron/electron/blob/master/docs/api/native-theme.md) which maps to Chromium's `NativeTheme` class.

To unlock as many use cases as possible and to unify all the cross-platform trickiness of dark mode I'm proposing (and will implement) the following additions to the API.

* Build an internal API `browserWindow.setGTKDarkThemeEnabled(true | false)` that is a setter for the current `darkTheme` option in the browser window constructor options.  This is to allow better dynamic dark theming on Linux.
* BrowserWindow's constructed should set the default value of the `darkTheme` option to `nativeTheme.shouldUseDarkColors`
* `shouldUseDarkColors` should have a setter that will do the following
  * Update the return value of `shouldUseDarkColors`
  * Do the equivalent of `systemPreferences.setAppLevelAppearance(shouldUseDarkColors ? 'dark' : 'light')` to update the native UI on macOS
  * Override the internal value of `NativeTheme::ShouldUseDarkColors` in Chromium.  This will require a patch that should be upstreamed
  * Automatically call `setGTKDarkThemeEnabled(shouldUseDarkColors)`  on all current BrowserWindow instances
  * Dispatch the `updated` event both in the Electron API and internally in Chromium

This list may be updated over time but that's all I've got at the moment.  The end goal of doing this is that `nativeTheme.shouldUseDarkColors` is correct 100% of the time and overriding that value should update **all** UI elements that Electron has control over.

fxha:
@MarshallOfSound I don't know whether I missed something but `systemPreferences.isDarkMode()` is deprecated as of version 7 and native theme's `shouldUseDarkColors` depends on the `themeSource` value. If I manually overwrite `themeSource`, I cannot longer get the preferred OS theme mode until I set `themeSource` to `system` and read the `shouldUseDarkColors` value and undo `themeSource` afterwards. Is it possible to add a function to read the OS theme mode independently of the `themeSource` value or is there already a function for this?

MarshallOfSound:
@fxha On macOS Catalina and above its impossible to tell what the OSs theme mode is.  You can only tell what theme mode your application has.  As such if you override your theme mode that's the only value you can see.

fxha:
That's very unfavourable because the `nativeTheme` API has different behavior across OSs - maybe depending on OS and Chromium version too. FYI, I would add/change the documentation a bit:

[...]

PR #19960, opened 2019-08-26
feat: add nativeTheme.themeSource to allow apps to override Chromiums theme choice

MarshallOfSound:
This is Phase 1 of #19932

A lot of the documentation and expected usage of this property has been added to `native-theme.md` so I recommend reviewers read that new documentation.

At a high level this property is added to allow apps to override Chromiums theme choice.  By default Chromium uses light colors unless your OS supports dark mode and has dark mode enabled, in which case it uses dark colors.  Some apps need to be able to override these colors choices for things like menus so that they can implement their own state machine of

* Follow System Mode
* Force Dark Mode
* Force Light Mode

The patch in this PR is deliberately kept minimal (and not the best cpp) to avoid conflicts.  My intention is to upstream this change so we can remove this patch.

Notes: Added `nativeTheme.themeSource` to allow apps to override Chromium and the OS's theme choice

### DIFF
--- a/docs/api/native-theme.md
+++ b/docs/api/native-theme.md
@@ -23,23 +23,24 @@ The `nativeTheme` module has the following properties:
 
 A `Boolean` for if the OS / Chromium currently has a dark mode enabled or is
 being instructed to show a dark-style UI.  If you want to modify this value you
-should use `shouldUseDarkColorsOverride` below.
+should use `themeSource` below.
 
-### `nativeTheme.shouldUseDarkColorsOverride`
+### `nativeTheme.themeSource`
 
-A `Boolean | null` property used to override and supercede the value that Chromium has
-chosen to use internally.  Setting this property to `null` will remove the override and
-everything will be reset to the OS default.  By default `shouldUseDarkColors` tracks the
-OS dark mode setting.
+A `String` property that can be `system`, `light` or `dark`.  It is used to override and supercede
+the value that Chromium has chosen to use internally.
 
-Settings this property to `true` will have the following effects:
+Setting this property to `system` will remove the override and
+everything will be reset to the OS default.  By default `themeSource` is `system`.
+
+Settings this property to `dark` will have the following effects:
 * `nativeTheme.shouldUseDarkColors` will be `true` when accessed
 * Any UI Electron renders on Linux and Windows including context menus, devtools, etc. will use the dark UI.
 * Any UI the OS renders on macOS including menus, window frames, etc. will use the dark UI.
 * The [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) CSS query will match `dark` mode.
 * The `updated` event will be emitted
 
-Settings this property to `false` will have the following effects:
+Settings this property to `light` will have the following effects:
 * `nativeTheme.shouldUseDarkColors` will be `false` when accessed
 * Any UI Electron renders on Linux and Windows including context menus, devtools, etc. will use the light UI.
 * Any UI the OS renders on macOS including menus, window frames, etc. will use the light UI.
@@ -48,9 +49,9 @@ Settings this property to `false` will have the following effects:
 
 The usage of this property should align with a classic "dark mode" state machine in your application
 where the user has three options.
-* `Follow OS` --> `shouldUseDarkColorsOverride = null`
-* `Dark Mode` --> `shouldUseDarkColorsOverride = true`
-* `Light Mode` --> `shouldUseDarkColorsOverride = false`
+* `Follow OS` --> `themeSource = 'system'`
+* `Dark Mode` --> `themeSource = 'dark'`
+* `Light Mode` --> `themeSource = 'light'`
 
 Your application should then always use `shouldUseDarkColors` to determine what CSS to apply.
 
--- a/docs/api/native-theme.md
+++ b/docs/api/native-theme.md
@@ -22,7 +22,37 @@ The `nativeTheme` module has the following properties:
 ### `nativeTheme.shouldUseDarkColors` _Readonly_
 
 A `Boolean` for if the OS / Chromium currently has a dark mode enabled or is
-being instructed to show a dark-style UI.
+being instructed to show a dark-style UI.  If you want to modify this value you
+should use `shouldUseDarkColorsOverride` below.
+
+### `nativeTheme.shouldUseDarkColorsOverride`
+
+A `Boolean | null` property used to override and supercede the value that Chromium has
+chosen to use internally.  Setting this property to `null` will remove the override and
+everything will be reset to the OS default.  By default `shouldUseDarkColors` tracks the
+OS dark mode setting.
+
+Settings this property to `true` will have the following effects:
+* `nativeTheme.shouldUseDarkColors` will be `true` when accessed
+* Any UI Electron renders on Linux and Windows including context menus, devtools, etc. will use the dark UI.
+* Any UI the OS renders on macOS including menus, window frames, etc. will use the dark UI.
+* The [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) CSS query will match `dark` mode.
+* The `updated` event will be emitted
+
+Settings this property to `false` will have the following effects:
+* `nativeTheme.shouldUseDarkColors` will be `false` when accessed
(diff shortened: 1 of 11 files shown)


docs/api/native-theme.md
```diff
@@ -22,7 +22,37 @@ The `nativeTheme` module has the following properties:
...
-being instructed to show a dark-style UI.
+being instructed to show a dark-style UI.  If you want to modify this value you
+should use `shouldUseDarkColorsOverride` below.
+
+### `nativeTheme.shouldUseDarkColorsOverride`
+
+A `Boolean | null` property used to override and supercede the value that Chromium has
+chosen to use internally.  Setting this property to `null` will remove the override and
+everything will be reset to the OS default.  By default `shouldUseDarkColors` tracks the
+OS dark mode setting.
```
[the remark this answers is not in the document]
MarshallOfSound: @jkleinsc `null`, no override by default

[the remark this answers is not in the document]
jkleinsc: @MarshallOfSound can you note that in the docs?

patches/chromium/feat_add_set_theme_source_to_allow_apps_to.patch
```diff
@@ -0,0 +1,91 @@
...
+index 70389e0245993faa2c17e9deefeb000280ef2368..4d56998a391c1c7508bb911eea0c16cdde443a34 100644
+--- a/ui/native_theme/native_theme.h
++++ b/ui/native_theme/native_theme.h
+@@ -21,6 +21,11 @@
+ #include "ui/native_theme/native_theme_export.h"
+ #include "ui/native_theme/native_theme_observer.h"
+ 
++#define HANDLE_OVERRIDDEN_SHOULD_USE_DARK_COLORS \
++  if (theme_source() == ThemeSource::kForcedLight) return false; \
++  if (theme_source() == ThemeSource::kForcedDark) return true;
```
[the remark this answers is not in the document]
MarshallOfSound: Like trick with the `do {} while (0)`.  Was concerned about mis-use of the macro as well and that covers it nicely.

docs/api/native-theme.md
```diff
@@ -22,7 +22,38 @@ The `nativeTheme` module has the following properties:
 ### `nativeTheme.shouldUseDarkColors` _Readonly_
 
 A `Boolean` for if the OS / Chromium currently has a dark mode enabled or is
-being instructed to show a dark-style UI.
+being instructed to show a dark-style UI.  If you want to modify this value you
+should use `themeSource` below.
+
+### `nativeTheme.themeSource`
```
nornagon: . o O ( Naming thoughts )

Perhaps this should be `nativeTheme.overrideTheme` and the values should be `'light'`, `'dark'`, or `undefined`?

patches/chromium/feat_add_set_theme_source_to_allow_apps_to.patch
```diff
@@ -0,0 +1,91 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <sattard@slack-corp.com>
+Date: Mon, 26 Aug 2019 14:32:41 -0700
+Subject: feat: add set_theme_source to allow apps to
+ override chromiums internal theme choice
+
+
```
nornagon: This patch requires a justification. Please provide one or more of:

1. A reason this can't be done without a patch
2. Instructions for how to remove this patch by upstreaming or alternative implementation
3. How to know if this patch is still necessary

patches/chromium/feat_add_set_theme_source_to_allow_apps_to.patch
```diff
@@ -0,0 +1,91 @@
...
+index 70389e0245993faa2c17e9deefeb000280ef2368..4d56998a391c1c7508bb911eea0c16cdde443a34 100644
+--- a/ui/native_theme/native_theme.h
++++ b/ui/native_theme/native_theme.h
+@@ -21,6 +21,11 @@
+ #include "ui/native_theme/native_theme_export.h"
+ #include "ui/native_theme/native_theme_observer.h"
+ 
++#define RETURN_IF_FORCED_THEME do { \
++  if (theme_source() == ThemeSource::kForcedLight) return false; \
++  if (theme_source() == ThemeSource::kForcedDark) return true; } while (0)
```
nornagon: please don't add #defines to class headers. Just repeat this logic, or find an alternative approach.

docs/api/native-theme.md
MarshallOfSound: @nornagon That's what is used to be but the underlying reason for this API is to back a 3-state switch and making a 3-state switch backed by a 2-state API (with the third state being "no value") felt worse than implementing the 3-state switch directly.

patches/chromium/feat_add_set_theme_source_to_allow_apps_to.patch
```diff
@@ -66,26 +62,28 @@ index 70389e0245993faa2c17e9deefeb000280ef2368..4d56998a391c1c7508bb911eea0c16cd
...
 --- a/ui/native_theme/native_theme_win.cc
 +++ b/ui/native_theme/native_theme_win.cc
-@@ -611,6 +611,7 @@ bool NativeThemeWin::ShouldUseDarkColors() const {
+@@ -611,6 +611,8 @@ bool NativeThemeWin::ShouldUseDarkColors() const {
    // ...unless --force-dark-mode was specified in which case caveat emptor.
    if (UsesHighContrastColors() && !IsForcedDarkMode())
      return false;
-+  RETURN_IF_FORCED_THEME
++  if (theme_source() == ThemeSource::kForcedLight) return false;
++  if (theme_source() == ThemeSource::kForcedDark) return true;
```
nornagon: Isn't this handled by `NativeTheme::ShouldUseDarkColors`? Do we need to insert the logic twice?

spec-main/api-native-theme-spec.ts
```diff
@@ -8,6 +10,58 @@ describe('nativeTheme module', () => {
...
+      nativeTheme.themeSource = 'dark'
+      let called = false
+      nativeTheme.once('updated', () => {
+        called = true
+      })
+      nativeTheme.themeSource = 'dark'
+      expect(called).to.equal(false)
+    })
+
+    describe('on macOS 10.14', () => {
```
nornagon: prefer `ifdescribe`:
```suggestion
    ifdescribe(process.platform === 'darwin' && semver.ge(os.release(), '18.0.0'))('on macOS 10.14', () => {
```

patches/chromium/feat_add_set_theme_source_to_allow_apps_to.patch
MarshallOfSound: Yes but if this block changes to be more similar to the dark aura one (not call the base `NativeTheme`) I don't want to have that bug.  Safer to call in all places IMO and it doesn't cause any significant overhead. Future proofing against future chromium patches, would rather get a conflict than it mysteriously stop working.

spec-main/api-native-theme-spec.ts
```diff
@@ -8,6 +10,58 @@ describe('nativeTheme module', () => {
...
+      nativeTheme.themeSource = 'dark'
+      expect(called).to.equal(false)
+    })
+
+    ifdescribe(process.platform === 'darwin' && semver.gte(os.release(), '18.0.0'))('on macOS 10.14', () => {
+      before(function () {
+        if (process.platform !== 'darwin') return this.skip()
+        // Darwin 18.0.0 === Mojave 10.14
+        if (semver.lt(os.release(), '18.0.0')) return this.skip()
+      })
```
nornagon: no need to skip twice

patches/chromium/feat_add_set_theme_source_to_allow_apps_to.patch
nornagon: hm... won't our tests catch that if it happens?

MarshallOfSound: Depends on the conditions required.  i.e. if it breaks on windows when the native dark theme is enabled our tests can't catch that.

jkleinsc:
@MarshallOfSound @electron/wg-releases decided that given how late we are in the 7.0 beta cycle, we are not going to backport this for 7.0.  We did discuss targeting this for a 7.1 release but wanted to follow up with you as to the urgency of this feature as well as future plans for #19932.

PR #19988, opened 2019-08-27
fix: ensure document.visibilityState aligns with the visibility of the TopLevelWindow

MarshallOfSound:
The best way to see what this changes is to review the bunch of new tests I added.  Basically this fixes the following two things and a bunch of edge cases

* Creating a window with `show: false` results in `document.visibilityState === 'visible'`.  This is fundamentally wrong so now it's hidden, but due to a technical constraint imposed by the `ready-to-show` event we only set it to hidden initially if folks use a new `paintWhenInitiallyHidden: false` option when constructing their window.
* Calling `win.hide()` and `win.show()` did not modify visibilityState resulting in backgrounded windows not being throttled correctly and CPU usage being insane while backgrounded.  This is fixed in this PR, no other changes for users needed.

Long term we should make `paintWhenInitiallyHidden: false` the default but for now let's keep this change non breaking.  The reason we need `paintWhenInitiallyHidden` is because previously `show: false` didn't set the renderer as hidden so it painted, that first paint is used to fire the `ready-to-show` event.  If we don't paint that event never fires (and I've documented this in this PR).

Notes: Fixed several cases where `document.visibilityState` would be incorrect.  The `win.show()` and `win.hide()` methods now correctly update visibility state

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -51,6 +51,9 @@ This event is usually emitted after the `did-finish-load` event, but for
 pages with many remote resources, it may be emitted before the `did-finish-load`
 event.
 
+Please note that using this event implies that the renderer will be considered "visible" and
+paint even though `show` is false.  This event will never fire if you use `paintWhenInitiallyHidden: false`
+
 ## Setting `backgroundColor`
 
 For a complex app, the `ready-to-show` event could be emitted too late, making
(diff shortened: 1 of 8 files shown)


MarshallOfSound:
> Would this affect the behavior of setBackgroundThrottling API?

Initial tests showed it didn't change the behavior of that API, our patch short circuits this logic so even these extra calls don't affect it.

> And should this be a feat as it adds a new option?

🤷‍♂ Technically yes, I added that new option to retain backwards compatibility.  I'd like to find some form of this change that can land in `7-0-x`.

PR #19996, opened 2019-08-28
feat: allow macOS tray visibility changes

codebytere:
#### Description of Change

Introduces `tray.visible` instance property, which is a  `Boolean` property indicating if the menu bar currently displays the status item. This allows users to show and hide a given Tray icon at will on `macOS` without having to create and destroy them.

cc @miniak @nornagon @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `tray.visible` instance property allowing users to show and hide a given Tray icon in the status bar on macOS.

### DIFF
--- a/shell/browser/api/atom_api_tray.cc
+++ b/shell/browser/api/atom_api_tray.cc
@@ -188,6 +188,20 @@ bool Tray::GetIgnoreDoubleClickEvents() {
 #endif
 }
 
+void Tray::SetVisible(bool visible) {
+#if defined(OS_MACOSX)
+  tray_icon_->SetVisible(visible);
+#endif
+}
+
+bool Tray::GetVisible() {
+#if defined(OS_MACOSX)
+  return tray_icon_->GetVisible();
+#else
+  return false;
+#endif
+}
+
 void Tray::DisplayBalloon(mate::Arguments* args,
                           const mate::Dictionary& options) {
   TrayIcon::BalloonOptions balloon_options;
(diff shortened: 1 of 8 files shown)


shell/browser/api/atom_api_tray.cc
```diff
@@ -188,6 +188,20 @@ bool Tray::GetIgnoreDoubleClickEvents() {
...
+#if defined(OS_MACOSX)
+  tray_icon_->SetVisible(visible);
+#endif
+}
+
+bool Tray::GetVisible() {
+#if defined(OS_MACOSX)
+  return tray_icon_->GetVisible();
+#else
+  return false;
```
miniak: I would rather not expose the property on other platform than return `false`. Or at least return `true`, as you cannot hide the icon on other platforms.

zcbenz: I agree this property should not be exposed to keep consistency with other platform-dependent APIs.

But the problem with property is, user has to explicitly test whether the property exists. Code like this would run without runtime exception:

```js
if (tray.visible) {
}
```

This is probably another reason not to do #12745.

zcbenz:
I think we should have an conclusion for #12745 before making this API a property.

shell/browser/api/atom_api_tray.cc
miniak: But `'visible' in tray` would work, right?

zcbenz: Yeah there are multiple ways to test whether the property exists, but I think we should make the API hard-to-misuse, instead of hard-to-use-correctly.

codebytere:
Closing for now.

PR #20001, opened 2019-08-28
fix: handle WM_GETMINMAXINFO instead of letting chromium do it

brenca:
Backport of #19928

See that PR for details.

Notes: Fixed a bug where windows would sometimes shrink to 0 size after being restored on Windows.


whyboris:
Thank you @brenca  🙇 Confirming that this bugfix, just released with Electron `6.0.6` closes my issue: https://github.com/electron/electron/issues/19816 🙇

PR #20006, opened 2019-08-28
feat: add aggregated renderer-crashed event with all affected webContents

miniak:
#### Description of Change
When a scriptable popup is created, there is another `BrowserWindow` instance, which has its own `webContents`. When the renderer process hosting both the parent and the child crashes, both `webContents` emit the `crashed` event. 

Adding new `renderer-crashed` event on `app` that's only emitted once per single crashed renderer process listing all affected `webContents`.

The implementation is taking advantage of the fact, that all `crashed` events are emitted in the same iteration of the message loop and therefore can be aggregated using `setImmediate()`.

#### Checklist
- [x] `npm test` passes

#### Release Notes

Notes: Added new `renderer-crashed` event on `app`, which is emitted only once per group of all `webContents` affected by a renderer process crash.

### DIFF
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -366,6 +366,20 @@ Returns:
 
 Emitted when the renderer process of `webContents` crashes or is killed.
 
+**Note:** This event is emitted separately for each `webContents` hosted in the same crashed renderer.
+
+**Deprecated:** Should use the new [`renderer-crashed`](#event-renderer-crashed) event.
+
+### Event: 'renderer-crashed'
+
+Returns:
+
+* `event` Event
+* `webContents` [WebContents](web-contents.md)[]
+* `killed` Boolean
+
+Emitted when the renderer process hosting the `webContents` crashes or is killed.
+
 ### Event: 'accessibility-support-changed' _macOS_ _Windows_
 
 Returns:
--- a/docs/api/breaking-changes.md
+++ b/docs/api/breaking-changes.md
@@ -20,6 +20,19 @@ const { remote } = require('electron')
 remote.webContents.fromId(webview.getWebContentsId())
 ```
 
+### `app` event `renderer-process-crashed`
+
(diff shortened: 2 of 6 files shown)


zcbenz:
I think we should use a new event name, so it is possible to print warning for the deprecated event. 

Silently changing the parameter type without warning would be frustrating for users.

miniak:
@MarshallOfSound the event should have been implemented as suggested when I added it initially.

miniak:
@zcbenz added a new event and deprecated the old one as suggested.

MarshallOfSound:
> Also, we have multiple apps that need this event, we don't want to duplicate the implementation in userland.

This is not sufficient reason to cause API churn in Electron.  I don't see any additional benefit to this API as opposed to the old one and changing events is significantly more churn than changing methods.  I.e. Events stop being emitted silently, methods will cause a JS error if you try to call something that does not exist.

> I don't think userland code should be depending on details like the crashed event being emitted in the same tick for example. 

Why not? If you want us to depend on it internally why is not dependable from userland?

I think I'm missing the justification for the API change, i.e. normally for an API change the reasoning would be

> The old one didn't work for common use-case X --> See this issue that folks raised (See nativeTheme)

or

> The old one no longer works due to a Chromium upgrade, we should replace it and that requires a breaking change.  --> See memory APIs

or

> A new change in Chromium allows us to enhance an existing API.  --> See contentTracing

API churn is something we should start avoiding so without a solid justification for this change and reasoning why it can't live in userland (where IMO it definitely can) I'm still -1 on this.

codebytere:
@miniak do you plan to work on this? if yes, it's probably best to bring it to the API WG.

miniak:
@codebytere Yes, I would like to discuss it in the API WG meeting.

nornagon:
Closing this for now; it seems like there's not sufficient reason to make this API change currently.

PR #20019, opened 2019-08-29
fix: Revert "fix: make sure that menu bar gets focus even when you click an item to focus it first"

MarshallOfSound:
Reverts electron/electron#19710

As outlined in comments in that PR this is re-causing issues where when you open the menu the webContents loses focus and it is never restored.  This breaks things like "copy", "paste" and the `webContents.getFocusedWebContents()` method.

Notes: Fixed issue where using the menu on Linux would un-focus the webContents


MarshallOfSound:
cc @brenca

Issue #10668, opened 2017-10-02
Honor the OS's cursor blink settings


davidhfo:
* Electron version: Various, as tested by apps mentioned below
* Operating system: Windows, macOS, Linux

### Expected behavior

Electron should honor OS-level cursor blink settings.

Using Windows and Firefox as an example, here is what should be happening (thanks @jonathansampson for the gif):

<img src="https://cloud.githubusercontent.com/assets/815158/24823693/131bb71c-1bc7-11e7-8dd4-b969b1c86bbe.gif">

### Actual behavior

The cursor still blinks. The choices made by the user in his/her OS are ignored.

Aside from the OS’s settings not being honored, note that this is an accessibility issue for some ([example 1](https://discussions.apple.com/thread/7689487), [example 2](https://github.com/brave/browser-laptop/issues/8123)), and an irritation or frustration for many others. A google search for `turn off blinking cursor` produces [387,000 results](https://www.google.com/search?q=turn+off+blinking+cursor).

I tried various apps built using Electron ([Tusk](https://champloohq.github.io/tusk/), [Wavebox](https://wavebox.io/), [Wire](https://wire.com/en/), [Boostnote](https://boostnote.io/), [Brave](https://brave.com/), and [Min](https://minbrowser.github.io/min/)). In every app and on any OS, the result was the same: the cursor blinked, though I have the OS settings for cursor blinking turned off.

### How to reproduce

1. Change your respective OS’s setting for cursor blink to off.

    #### On Windows:

     - Set `HKEY_CURRENT_USER\Control Panel\Desktop\CursorBlinkRate` in the registry to `-1`.
     - OR (this will set the above value) navigate to Control Panel -> Keyboard, and under “Cursor blink rate,” move the slider all the way to the left.

    #### On macOS:
    Set values in the terminal approximating the following:
    ```
    defaults write -g NSTextInsertionPointBlinkPeriodOn -float 30000
    defaults write -g NSTextInsertionPointBlinkPeriodOff -float 1
    ```
    This input causes the cursor to be on for 30 seconds, off for just a millisecond, then on again. Though this is not technically “off,” with these values the blink is not noticeable and is as good as being off.

    #### On Linux:
    ##### GNOME
    Enter `gsettings set org.gnome.desktop.interface cursor-blink false` in a terminal.

     ##### Xfce
     - Enter `xfconf-query -c xsettings -p /Net/CursorBlink -s false` in a terminal.
[...]

PR #20020, opened 2019-08-29
fix: honor cursor blink rate

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/10668.

Ensures that WebPreferences honor the system cursor blink rate on creation.

cc @zcbenz @nornagon @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where cursor blink rate was not honored.

### DIFF
--- a/shell/browser/api/atom_api_web_contents.cc
+++ b/shell/browser/api/atom_api_web_contents.cc
@@ -100,10 +100,12 @@
 
 #if !defined(OS_MACOSX)
 #include "ui/aura/window.h"
+#else
+#include "ui/base/cocoa/defaults_utils.h"
 #endif
 
-#if defined(OS_MACOSX)
-#include "ui/base/cocoa/defaults_utils.h"
+#if defined(OS_LINUX)
+#include "ui/views/linux_ui/linux_ui.h"
 #endif
 
 #if defined(OS_LINUX) || defined(OS_WIN)
@@ -466,6 +468,10 @@ void WebContents::InitWithSessionAndOptions(
   base::TimeDelta interval;
   if (ui::TextInsertionCaretBlinkPeriod(&interval))
     prefs->caret_blink_interval = interval;
+#elif defined(OS_LINUX)
+  views::LinuxUI* linux_ui = views::LinuxUI::instance();
+  if (linux_ui)
+    prefs->caret_blink_interval = linux_ui->GetCursorBlinkInterval();
 #endif
 
   // Save the preferences in C++.
--- a/shell/browser/api/atom_api_web_contents.cc
+++ b/shell/browser/api/atom_api_web_contents.cc
@@ -102,6 +102,10 @@
 #include "ui/aura/window.h"
 #endif
 
+#if defined(OS_MACOSX)
+#include "ui/base/cocoa/defaults_utils.h"
+#endif
+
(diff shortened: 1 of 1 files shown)


shell/browser/api/atom_api_web_contents.cc
```diff
@@ -458,6 +468,23 @@ void WebContents::InitWithSessionAndOptions(
...
+  views::LinuxUI* linux_ui = views::LinuxUI::instance();
+  if (linux_ui)
+    prefs->caret_blink_interval = linux_ui->GetCursorBlinkInterval();
+#endif
+
+#if defined(OS_WIN)
+  static const size_t system_value = ::GetCaretBlinkTime();
+  if (system_value != 0 && system_value != INFINITE)
+    prefs->caret_blink_interval =
+        base::TimeDelta::FromMilliseconds(system_value);
```
ckerr: _(minor)_ One thing that might be nice to copy from [TextField::GetCaretBlinkInterval](https://cs.chromium.org/chromium/src/ui/views/controls/textfield/textfield.cc?sq=package:chromium&dr=CSs&g=0&l=266) is that if `::GetCaretBlinkTime` fails, it uses a 500msec blink as a fallback

e.g.

```cpp
const auto system_msec = ::GetCaretBlinkTime();
if (system_msec == INFINITE)
  prefs->caret_blink_interval = base::TimeDelta()
else if (system_msec == 0)
  prefs->caret_blink_interval = base::TimeDelta::FromMilliseconds(500);
else
  prefs->caret_blink_interval = base::TimeDelta::FromMilliseconds(system_msec);
```

shell/browser/api/atom_api_web_contents.cc
```diff
@@ -458,6 +468,23 @@ void WebContents::InitWithSessionAndOptions(
   prefs->subpixel_rendering = params->subpixel_rendering;
 #endif
 
+#if defined(OS_MACOSX)
+  base::TimeDelta interval;
+  if (ui::TextInsertionCaretBlinkPeriod(&interval))
+    prefs->caret_blink_interval = interval;
```
ckerr: 500msec would be a good fallback default here too

codebytere: macOS automatically falls back to 500 :)

shell/browser/api/atom_api_web_contents.cc
```diff
@@ -100,6 +100,16 @@
 
 #if !defined(OS_MACOSX)
 #include "ui/aura/window.h"
+#else
+#include "ui/base/cocoa/defaults_utils.h"
+#endif
+
+#if defined(TOOLKIT_VIEWS)
+#include "ui/views/controls/textfield/textfield.h"
```
miniak: we don't need this unless we use `views::Textfield::GetCaretBlinkInterval();`

miniak:
@codebytere can we use `views::Textfield::GetCaretBlinkInterval();`? it wraps `GetCaretBlinkTime`

codebytere:
@miniak it does wrap that method, but it also wraps things for mac and so for clarity & maintainability i’d prefer we explicitly set all three ourselves since we’d need to call the linux one separately regardless.

Perhaps we could consider upstreaming the linux one into that call and then using that in the future?

PR #20028, opened 2019-08-29
test: make getPath(logs) spec more idempotent

codebytere:
#### Description of Change

Improve granularity of `getPath('logs')` spec by checking the specific folder that the logs path would be created as. Also makes the test more idempotent by removing the created directory once the test has completed.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/spec-main/api-app-spec.ts
+++ b/spec-main/api-app-spec.ts
@@ -673,26 +673,29 @@ describe('app module', () => {
   })
 
   describe('getPath("logs")', () => {
+    // this won't be deterministic except on CI
+    before(function () {
+      if (!isCI) this.skip()
+    })
+  
     const logsPaths = {
-      'darwin': path.resolve(homedir(), 'Library', 'Logs'),
-      'linux': path.resolve(homedir(), 'AppData', app.name),
-      'win32': path.resolve(homedir(), 'AppData', app.name),
+      'darwin': path.resolve(homedir(),'Library', 'Logs', 'Electron'),
+      'linux': path.resolve(app.getPath('userData'), app.name, 'logs'),
+      'win32': path.resolve(app.getPath('userData'), app.name, 'logs'),
     }
 
     it('has no logs directory by default', () => {
-      // this won't be deterministic except on CI since
-      // users may or may not have this dir
-      if (!isCI) return
-
       const osLogPath = (logsPaths as any)[process.platform]
-      expect(fs.existsSync(osLogPath)).to.be.false
+      expect(fs.existsSync(osLogPath)).to.be.false(`${osLogPath} exists`)
     })
 
     it('creates a new logs directory if one does not exist', () => {
       expect(() => { app.getPath('logs') }).to.not.throw()
 
       const osLogPath = (logsPaths as any)[process.platform]
-      expect(fs.existsSync(osLogPath)).to.be.true
+      expect(fs.existsSync(osLogPath)).to.be.true(`${osLogPath} does not exist`)
+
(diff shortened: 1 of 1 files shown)


spec-main/api-app-spec.ts
```diff
@@ -674,25 +674,30 @@ describe('app module', () => {
...
-      'linux': path.resolve(homedir(), 'AppData', app.name),
-      'win32': path.resolve(homedir(), 'AppData', app.name),
+      'darwin': path.resolve(homedir(), 'Library', 'Logs', 'Electron'),
+      'linux': path.resolve(homedir(), 'AppData', 'Electron', 'logs'),
+      'win32': path.resolve(homedir(), 'AppData', 'Electron', 'logs'),
     }
 
+    after(() => {
+      const osLogPath = (logsPaths as any)[process.platform]
+      fs.rmdirSync(osLogPath)
```
ckerr: I don't think we want to call this unconditionally. It looks like this will remove the directory for any human who happens to run this test, as well.

spec-main/api-app-spec.ts
```diff
@@ -674,25 +674,30 @@ describe('app module', () => {
 
   describe('getPath("logs")', () => {
     const logsPaths = {
-      'darwin': path.resolve(homedir(), 'Library', 'Logs'),
-      'linux': path.resolve(homedir(), 'AppData', app.name),
-      'win32': path.resolve(homedir(), 'AppData', app.name),
+      'darwin': path.resolve(homedir(), 'Library', 'Logs', 'Electron'),
+      'linux': path.resolve(homedir(), 'AppData', 'Electron', 'logs'),
```
ckerr: The second component in this call should be '.config' rather than 'AppData'. This would cause the 'has no logs directory by default' test to incorrectly pass if  ~/.config/Electron/logs existed

That bug looks like a copy-paste error that predates this PR

codebytere: fixed!

spec-main/api-app-spec.ts
```diff
@@ -673,26 +673,29 @@ describe('app module', () => {
   })
 
   describe('getPath("logs")', () => {
+    before(() => {
+      // this won't be deterministic except on CI
+      if (!isCI) return
+    })
```
ckerr: I think this function literally has no effect and that the tests run either way?

```js
before(function () {
  if (!isCI) this.skip()
})
```

shell/browser/api/atom_api_app.cc
```diff
@@ -857,7 +857,6 @@ void App::SetAppLogsPath(util::ErrorThrower thrower,
   } else {
     base::FilePath path;
     if (base::PathService::Get(DIR_USER_DATA, &path)) {
-      path = path.Append(base::FilePath::FromUTF8Unsafe(GetApplicationName()));
```
codebytere: this was adding an extra name directory to the logs:
`~/.config/Electron Tests Main/Electron Tests Main/logs` instead of `~/.config/Electron Tests Main/logs`, which was incorrect and causing the previous test failure as that's not what documentation outlined should be the case.

ckerr: But this is a `semver/major` change because the pre-existing behavior is in 4-2-x, 5-0-x, and 6-0-x

codebytere: agreed that this is a point of discussion - arguably it was previously incorrect which makes this, to some extent, a fix 🤔

PR #20031, opened 2019-08-29
ci: run tests on WOA hardware

jkleinsc:
#### Description of Change

This PR add WOA testing to our CI by running our tests on Windows on Arm hardware

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/azure-pipelines-woa.yml
+++ b/azure-pipelines-woa.yml
@@ -6,8 +6,8 @@ steps:
 
 - script: |
     cd src\electron
-    npm install --verbose
-  displayName: 'NPM install'
+    node script/yarn.js install --frozen-lockfile
+  displayName: 'Yarn install'
 
 - powershell: |
     $localArtifactPath = "$pwd\dist.zip"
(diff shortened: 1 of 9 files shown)


azure-pipelines-woa.yml
```diff
@@ -78,7 +78,7 @@ steps:
   displayName: 'Verify ffmpeg'
 
 - script: |
-    cd src
-    echo "Verifying mksnapshot"
-    python electron\script\verify-mksnapshot.py --build-dir out\Default --source-root %cd%
-  displayName: 'Verify mksnapshot'
+    taskkill /F /IM electron.exe
+    taskkill /F /IM MicrosoftEdge.exe
```
deepak1556: So the processes don't exit cleanly, is this an issue ? Also why is there a edge process running in ci ?

brenca: I think the edge process is because of the `shell.openExternal` test where we open a page in the default browser.

jkleinsc: @deepak1556 those lines are in there to handle bad builds.  The tests don't run in a container, so we need to make sure that the OS is cleaned up for the next test run.   @brenca is right about the edge process running as a result of `shell.openExternal`.

jkleinsc: @deepak1556 those lines are in there to handle bad builds. The tests don't run in a container, so we need to make sure that the OS is cleaned up for the next test run. @brenca is right about the edge process running as a result of shell.openExternal.

PR #20033, opened 2019-08-29
fix: use rounded scaling for window sizing on Windows

### LINKED ISSUE #19769: set the x of BrowserWindow to 1021-1079,the width of the win will increase 2
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * Electron v6.0.2
* **Operating System:**
  * Windows 10 1903
  * just had used v6.0.2

### Expected Behavior
set bound.x of browserWin to a new number (1021-1079)

### Actual Behavior
the bound.width increase 2

### To Reproduce
`setInterval(()=>{
        win.setBounds({
            x:1079,
        })
        console.log(win.getBounds().width)
    },100);`

### Screenshots

image

brenca:
#### Description of Change

Chromium uses ceiled sizes when translating between DIP coordinates and screen coordinates to ensure that content of native widgets isn't cut off ever. This however has some side effects, mostly that window sizing behaves very weirdly when using fractional scaling (1.25 for example) - if you set the bounds of a window, you might not get the same bounds back when you query the window for it.

This can also lead to weird behaviour where you set the `x` coordinate of a window and you end up with a window that has:
 - the incorrect `x` coordinate
 - different size than before you set the `x` coordinate

Fixes #19769.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed a bug where window bounds didn't match the set values on Windows.

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -77,3 +77,4 @@ picture-in-picture.patch
 disable_compositor_recycling.patch
 allow_new_privileges_in_unsandboxed_child_processes.patch
 expose_setuseragent_on_networkcontext.patch
+win-window-scaling.patch
--- a/patches/chromium/win-window-scaling.patch
+++ b/patches/chromium/win-window-scaling.patch
@@ -1,0 +1,48 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Heilig Benedek <benecene@gmail.com>
+Date: Thu, 29 Aug 2019 22:17:17 +0200
+Subject: fix: use rounded scaling for window sizing on Windows
+
+Chromium uses ceiled sizes when translating between DIP coordinates and screen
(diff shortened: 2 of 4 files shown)


brenca:
@deepak1556 Would the folks at chromium accept that code change just to make things more embedder friendly?

deepak1556:
They have been accepting patches that helps specific embedders but this patch is more general and not tied to any specific embedder, so I would be confident in them accepting this, ofcourse they might have some code style suggestions :)

brenca:
Alright, I'll adjust the patch tomorrow and I'll try upstreaming it as well, thanks for the suggestion!

zcbenz:
@brenca Do you still plan to work on this PR?

codebytere:
@zcbenz Benedek has moved on from Electron, so i think this should be closed unless someone else wants to pick it up

PR #20050, opened 2019-08-30
fix: emit `will-navigate` event when sandbox is enabled

### LINKED ISSUE #8841: `will-navigate` event not fired when in sandbox mode
Thanks for Electron! Amazing project :)

* Electron version: 1.6.2
* Operating system: macOS Sierra (10.12.3)

It happens when `BrowserWindow` is opened with `sandbox: true`. It might be on purpose, but I couldn't find any Electron/Chromium documentation that explains why. So I'm suspecting it's a bug. I can submit a PR with a failing test to this repo if you agree that it's a bug.

### Expected behavior

`webContents`' `will-navigate` event should fire when user reloads page fx.

### Actual behavior

It does not fire.

### How to reproduce

package.json:

```json
{
  "name": "electron-sandbox-will-navigate-bug",
  "version": "0.0.0",
  "main": "main.js",
  "dependencies": {
    "electron": "^1.6.2"
  }
}
```

main.js:

```js
const path = require('path')

app.on('ready', function() {
    webPreferences: {
      sandbox: true
    }
  })

  const {webContents} = win
  webContents.on('will-navigate', () => {
    console.log('will-navigate')
  })
  webContents.on('did-navigate', () => {
    console.log('did-navigate')
  })

  win.loadURL('file://' + path.join(__dirname, 'index.html'))
})
```

index.html:

```html
<!doctype html>
<html>
  <body>
    <button onclick="window.location.reload()">Reload</button>
  </body>
</html>
```

Perform these steps:

- Run `npm install`
- Run `./node_modules/.bin/electron .`
- Click the Reload button in the browser window
- Observe the Terminal output

Expected output:

```
did-navigate
will-navigate
did-navigate
```

Actual output:

```
did-navigate
did-navigate
```

(will-navigate is not fired).

If you outcomment the line `sandbox:true` in `main.js`, then it works as expected.

aegarbutt:
#### Description of Change
`will-navigate` events were not being delivered when `sandbox` was
enabled. This is because the `AtomSandboxedRendererClient` did not
override the `ShouldFork` method found in `ContentRendererClient` while
the `AtomRendererClient` did.

Fixes #8841

CC @codebytere 

#### Checklist
- [X] `npm test` passes

#### Release Notes

Notes: Fixed delivery of `will-navigate` events when sandbox is enabled.

### DIFF
--- a/shell/renderer/renderer_client_base.h
+++ b/shell/renderer/renderer_client_base.h
@@ -86,6 +86,11 @@ class RendererClientBase : public content::ContentRendererClient {
   void RunScriptsAtDocumentStart(content::RenderFrame* render_frame) override;
   void RunScriptsAtDocumentEnd(content::RenderFrame* render_frame) override;
   void RunScriptsAtDocumentIdle(content::RenderFrame* render_frame) override;
+  bool ShouldFork(blink::WebLocalFrame* frame,
+                  const GURL& url,
+                  const std::string& http_method,
+                  bool is_initial_navigation,
+                  bool is_server_redirect) override;
 
  protected:
 #if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
--- a/shell/renderer/atom_renderer_client.cc
+++ b/shell/renderer/atom_renderer_client.cc
@@ -43,12 +43,6 @@ AtomRendererClient::~AtomRendererClient() {
   asar::ClearArchives();
 }
 
-void AtomRendererClient::RenderFrameCreated(
-    content::RenderFrame* render_frame) {
-  new AtomRenderFrameObserver(render_frame, this);
-  RendererClientBase::RenderFrameCreated(render_frame);
-}
-
 void AtomRendererClient::RunScriptsAtDocumentStart(
     content::RenderFrame* render_frame) {
   RendererClientBase::RunScriptsAtDocumentStart(render_frame);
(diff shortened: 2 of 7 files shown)


shell/renderer/renderer_client_base.h
```diff
@@ -86,6 +86,11 @@ class RendererClientBase : public content::ContentRendererClient {
   void RunScriptsAtDocumentStart(content::RenderFrame* render_frame) override;
   void RunScriptsAtDocumentEnd(content::RenderFrame* render_frame) override;
   void RunScriptsAtDocumentIdle(content::RenderFrame* render_frame) override;
+  bool ShouldFork(blink::WebLocalFrame* frame,
```
zcbenz: Returning `true` for `ShouldFork` has quite a lot of side effects, and one reason `sandbox` mode was introduced was to avoid that, I don't think this solution can be accepted.

ridesky: so  is there a  good resolution we can  fire 'will-navigation'  in  sandbox mode?

codebytere:
@aegarbutt do you plan to continue this change?

aegarbutt:
Apologies; I've been sidetracked at work and home.

If someone else wants to pursue a patch that would solve the issue of delivering will-navigate events properly, please do.

Aside; I may misunderstand why sandbox is protecting against ShouldFork? I thought the renderer process was forking from a zygote anyway, no?

zcbenz:
This approach seems to be a dead end: the implementation of `will-navigate` requires the renderer process to be restarted for navigations (i.e. returning `true` in `ShouldFork`), while the purpose of `sandbox` mode is to avoid that.

To really solve the issue, we have to implement `will-navigate` in a different way, which I'm not sure if it is possible without patching Chromium.

So I'm closing this PR as we can not tweak current approach to solve the side effects..

PR #20065, opened 2019-08-31
docs: Improve wording of setBrowserView and getBrowserView

koralcem:
#### Description of Change
Improving the wording in the description of `setBrowserView` and `getBrowserView`

cc: @codebytere, @zcbenz, @MarshallOfSound. cc'ing you folks only because you seem to be a few of the recent committers. 

#### Checklist

notes: no-notes

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -1739,14 +1739,14 @@ removed in future Electron releases.
 
 #### `win.setBrowserView(browserView)` _Experimental_
 
-* `browserView` [BrowserView](browser-view.md) | null - Attach browserView to win.
-If there is some other browserViews was attached they will be removed from
+* `browserView` [BrowserView](browser-view.md) | null - Attach `browserView` to `win`.
+If there are other `BrowserView`s attached, they will be removed from
 this window.
 
 #### `win.getBrowserView()` _Experimental_
 
-Returns `BrowserView | null` - an BrowserView what is attached. Returns `null`
-if none is attached. Throw error if multiple BrowserViews is attached.
+Returns `BrowserView | null` - The `BrowserView` attached to `win`. Returns `null`
+if none is attached. Throws an error if multiple `BrowserView`s are attached.
 
 #### `win.addBrowserView(browserView)` _Experimental_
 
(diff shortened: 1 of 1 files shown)


docs/api/browser-window.md
```diff
@@ -1739,14 +1739,14 @@ removed in future Electron releases.
 
 #### `win.setBrowserView(browserView)` _Experimental_
 
-* `browserView` [BrowserView](browser-view.md) | null - Attach browserView to win.
-If there is some other browserViews was attached they will be removed from
+* `browserView` [BrowserView](browser-view.md) | null - Attach `browserView` to `win`.
+If there are other `BrowserView`s attached, they will be removed from
```
koralcem: Aside from the wording changes, I have a question too. Could there be something wrong with markdown rendering over at `electronjs.org`? 

I ask because type of the `browserView` argument here is supposedly `BrowserView | null`. And indeed [on GitHub](https://github.com/electron/electron/blob/master/docs/api/browser-window.md) I see this:
ongithub

But over at [the docs given at electronjs.org](https://electronjs.org/docs/api/browser-window#winsetbrowserviewbrowserview-experimental) I see:
onsite

The difference doesn't seem trivial; whether `null` is a valid input or not. Should the argument and return types be in backticks?

docs/api/browser-window.md
```diff
@@ -1739,14 +1739,14 @@ removed in future Electron releases.
...
 this window.
 
 #### `win.getBrowserView()` _Experimental_
 
-Returns `BrowserView | null` - an BrowserView what is attached. Returns `null`
-if none is attached. Throw error if multiple BrowserViews is attached.
+Returns `BrowserView | null` - The `BrowserView` attached to `win`. Returns `null`
+if none is attached. Throws an error if multiple `BrowserView`s are attached.
```
codebytere: let's maybe say `if one is not attached` here?

docs/api/browser-window.md
codebytere: @koralcem that's because the documentation matches the latest **stable** version of Electron, and the `null` argument was added [recently](https://github.com/electron/electron/commit/9910507bc4536342cc3a851675d633dff097b827). It's not a rendering issue :)

koralcem: Ah, thanks. I learned something new today 😄

docs/api/browser-window.md
koralcem: That does sound better. Changing...

PR #20095, opened 2019-09-03
test: move robotjs-based devtools paste test to main runner

nornagon:
#### Description of Change
This also enables it for non-Linux platforms. TBD if it'll work there on CI.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: none

PR #20099, opened 2019-09-03
test: move more web contents specs

nornagon:
#### Description of Change
Last of the webContents specs apart from #20095.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash that could occur if a WebContents was destroyed during a navigation event.


shell/browser/api/atom_api_web_contents.cc
```diff
@@ -927,10 +927,14 @@ void WebContents::DidFinishLoad(content::RenderFrameHost* render_frame_host,
   bool is_main_frame = !render_frame_host->GetParent();
   int frame_process_id = render_frame_host->GetProcess()->GetID();
   int frame_routing_id = render_frame_host->GetRoutingID();
+  auto weak_this = GetWeakPtr();
   Emit("did-frame-finish-load", is_main_frame, frame_process_id,
        frame_routing_id);
 
-  if (is_main_frame)
+  // ⚠️WARNING!⚠️
```
codebytere: did you see this crash specifically when you moved the tests to the main process? or was it something you noticed and added incidentally?

shell/browser/api/event_emitter.h
```diff
@@ -107,15 +107,15 @@ class EventEmitter : public Wrappable<T> {
                      v8::Local<v8::Object> event,
                      Args&&... args) {
     DCHECK_CURRENTLY_ON(content::BrowserThread::UI);
-    v8::Locker locker(isolate());
-    v8::HandleScope handle_scope(isolate());
-    EmitEvent(isolate(), GetWrapper(), name, event,
-              std::forward<Args>(args)...);
-    auto context = isolate()->GetCurrentContext();
+    auto* isolate = this->isolate();
```
codebytere: what motivated this change?

shell/browser/api/atom_api_web_contents.cc
nornagon: Yes, this crash happens when running in the main process because the `destroy` is synchronous, vs. asynchronous when running in the remote process.

shell/browser/api/event_emitter.h
nornagon: `this` is destroyed by `EmitEvent`. so `this->isolate()` crashes.

nornagon:
```suggestion
    // It's possible that |this| will be deleted by EmitEvent, so save anything we need
    // from |this| before calling EmitEvent.
    auto* isolate = this->isolate();
```

codebytere:
Once rebased this is good to go :)

PR #20103, opened 2019-09-04
test: remove usage of remote module from node tests

nornagon:
#### Description of Change
Moves some tests for node that were mostly just calling spawn/fork into the main runner, to avoid usage of the remote module.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


spec/node-spec.js
```diff
@@ -526,21 +343,7 @@ describe('node feature', () => {
...
-      expect(process.stdout.isTTY).to.be.a('boolean')
-    })
-
-    it('should have isTTY undefined on Windows', function () {
-      if (isCI || process.platform !== 'win32') {
-        this.skip()
-        return
-      }
-
+    it('isTTY should be undefined in the renderer process', function () {
```
nornagon:
```suggestion
    // TODO: figure out why process.stdout.isTTY is true on Darwin but not Linux/Win.
    ifit(process.platform !== 'darwin)('isTTY should be undefined in the renderer process', function () {
```

nornagon:
@jkleinsc the test was previously skipped on all platforms on CI, so I've updated the test to only be skipped on Mac now, since that's where it was failing.

PR #20131, opened 2019-09-05
chore: remove native_mate (Part 3)

zcbenz:
#### Description of Change

This PR removes some uses of native_mate version of `event_emitter_caller.h` and `callback_converter.h`.

The `ErrorThrower` class is also migrated to used gin.

#### Release Notes

Notes: no-notes


patches/chromium/gin_dictionary_default_constructor.patch
```diff
@@ -0,0 +1,37 @@
...
+
+diff --git a/gin/dictionary.cc b/gin/dictionary.cc
+index 95e00072700c..7643347890a5 100644
+--- a/gin/dictionary.cc
++++ b/gin/dictionary.cc
+@@ -6,6 +6,10 @@
+ 
+ namespace gin {
+ 
++Dictionary::Dictionary()
```
deepak1556: For such call sites, can we resort to having `gin::Arguments` as the argument and extract `gin::Dictionary` from that in the callback, which would avoid this patch. ex:

```
void AtomDownloadManagerDelegate::OnDownloadSaveDialogDone(..., gin::Arguments* args) {
  gin::Dictionary result(args->isolate());
  // extract result from args.
}
```


Ideally it would be better if we rewrite `ArgumentsHolder` to not expect a default constructor.

MarshallOfSound: As an alternative (because I don't like the idea of duplicate that dictionary construction code everywhere).  Can we just extend `gin::Dictionary` into `electron::util::Dictionary` and implement the empty constructor there?

zcbenz: We already have an extended `gin_helper::Dictionary` and it should be possible to use it to avoid patching `gin`. I'll put it in the next PR.

PR #20137, opened 2019-09-05
fix: emit updated on NativeTheme on the UI thread to avoid DCHECK

MarshallOfSound:
Found while experimenting locally, this `OnNativeThemeUpdated` call can come from any thread apparently so let's post it to the UI thread.,

Related to #19932

Notes: no-notes

### DIFF
--- a/shell/browser/api/atom_api_native_theme.cc
+++ b/shell/browser/api/atom_api_native_theme.cc
@@ -6,6 +6,9 @@
 
 #include <string>
 
+#include "base/task/post_task.h"
+#include "content/public/browser/browser_task_traits.h"
+#include "content/public/browser/browser_thread.h"
 #include "native_mate/dictionary.h"
 #include "native_mate/object_template_builder.h"
 #include "shell/common/node_includes.h"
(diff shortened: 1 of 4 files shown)


shell/common/api/atom_api_native_theme.cc
```diff
@@ -26,10 +29,17 @@ NativeTheme::~NativeTheme() {
...
-void NativeTheme::OnNativeThemeUpdated(ui::NativeTheme* theme) {
+void NativeTheme::OnNativeThemeUpdatedOnUI() {
   Emit("updated");
 }
 
+void NativeTheme::OnNativeThemeUpdated(ui::NativeTheme* theme) {
+  base::PostTaskWithTraits(
+      FROM_HERE, {content::BrowserThread::UI},
+      base::BindOnce(&NativeTheme::OnNativeThemeUpdatedOnUI,
+                     base::Unretained(this)));
```
codebytere: are we 100% sure the lifetime of `this` lasts beyond when this callback can be invoked?

MarshallOfSound: `api::NativeTheme` lasts as long as they app does, as does `ui::NativeTheme`.  They are both process-lived singletons

codebytere:
```
not ok 588 nativeTheme module nativeTheme.themeSource should emit the "updated" event when it is set and the resulting "shouldUseDarkColors" value changes
  AssertionError: expected false to equal true
      at Context.<anonymous> (electron/spec-main/api-native-theme-spec.ts:38:25)
```
seems this change broke the spec

MarshallOfSound:
> seems this change broke the spec

Now the event is async so the previous logic that relied on it being sync doesn't work.  Lemme hack around it a bit

PR #20138, opened 2019-09-05
feat: default gtk darkTheme option to nativeTheme.shouldUseDarkColors for better platform support

MarshallOfSound:
This is done so that dark mode / `nativeTheme.themeSource` make everything else Just Work out of the box.

This also updates `darkTheme` when `themeSource` is modified in the `nativeTheme` API to match macOS behavior.

Phase 3 (and almost final phase) of #19932

Notes: BrowserWindow `darkTheme` option now defaults to `nativeTheme.shouldUseDarkColors`

### DIFF
--- a/shell/browser/native_window_views.cc
+++ b/shell/browser/native_window_views.cc
@@ -34,6 +34,7 @@
 #include "ui/base/hit_test.h"
 #include "ui/gfx/image/image.h"
 #include "ui/gfx/native_widget_types.h"
+#include "ui/native_theme/native_theme.h"
 #include "ui/views/background.h"
 #include "ui/views/controls/webview/unhandled_keyboard_event_handler.h"
 #include "ui/views/controls/webview/webview.h"
@@ -212,7 +213,8 @@ NativeWindowViews::NativeWindowViews(const gin_helper::Dictionary& options,
   window_state_watcher_ = std::make_unique<WindowStateWatcher>(this);
 
   // Set _GTK_THEME_VARIANT to dark if we have "dark-theme" option set.
-  bool use_dark_theme = false;
+  bool use_dark_theme =
+      ui::NativeTheme::GetInstanceForNativeUi()->ShouldUseDarkColors();
   if (options.Get(options::kDarkTheme, &use_dark_theme) && use_dark_theme) {
     SetGTKDarkThemeEnabled(use_dark_theme);
   }
--- a/shell/browser/native_window_views.cc
+++ b/shell/browser/native_window_views.cc
@@ -215,9 +215,8 @@ NativeWindowViews::NativeWindowViews(const gin_helper::Dictionary& options,
   // Set _GTK_THEME_VARIANT to dark if we have "dark-theme" option set.
   bool use_dark_theme =
       ui::NativeTheme::GetInstanceForNativeUi()->ShouldUseDarkColors();
-  if (options.Get(options::kDarkTheme, &use_dark_theme) && use_dark_theme) {
-    SetGTKDarkThemeEnabled(use_dark_theme);
-  }
+  options.Get(options::kDarkTheme, &use_dark_theme);
+  SetGTKDarkThemeEnabled(use_dark_theme);
 
   // Before the window is mapped the SetWMSpecState can not work, so we have
   // to manually set the _NET_WM_STATE.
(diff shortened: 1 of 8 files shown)


shell/browser/api/atom_api_top_level_window.h
```diff
@@ -32,13 +32,16 @@ class TopLevelWindow : public mate::TrackableObject<TopLevelWindow>,
...
                              v8::Local<v8::FunctionTemplate> prototype);
+  static std::vector<TopLevelWindow*> GetAllWindows() { return all_windows; }
 
   base::WeakPtr<TopLevelWindow> GetWeakPtr() {
     return weak_factory_.GetWeakPtr();
   }
 
   NativeWindow* window() const { return window_.get(); }
 
+  void SetGTKDarkThemeEnabled(bool use_dark_theme);
```
ckerr: This doesn't seem to be used?

shell/browser/native_window_views.cc
```diff
@@ -212,7 +213,8 @@ NativeWindowViews::NativeWindowViews(const mate::Dictionary& options,
   window_state_watcher_.reset(new WindowStateWatcher(this));
 
   // Set _GTK_THEME_VARIANT to dark if we have "dark-theme" option set.
-  bool use_dark_theme = false;
+  bool use_dark_theme =
+      ui::NativeTheme::GetInstanceForNativeUi()->ShouldUseDarkColors();
   if (options.Get(options::kDarkTheme, &use_dark_theme) && use_dark_theme) {
```
ckerr: Not new to this PR, but now that I read it again -- are you sure we want the `&& use_dark_theme` clause here? Do we want to call `SetGTKDarkThemeEnabled()` here even if `use_dark_theme` is false?

MarshallOfSound: Good call, we should call it no matter what 👍

shell/browser/api/atom_api_top_level_window.h
MarshallOfSound: It's used in `atom_api_native_theme`

shell/browser/api/atom_api_top_level_window.cc
```diff
@@ -118,6 +121,8 @@ TopLevelWindow::TopLevelWindow(v8::Isolate* isolate,
 }
 
 TopLevelWindow::~TopLevelWindow() {
+  all_windows.erase(std::remove(all_windows.begin(), all_windows.end(), this),
+                    all_windows.end());
```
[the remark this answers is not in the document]
ckerr: Since you left this unresolved, I'll expand on the previous comment with rationale:

 * If using a set or unordered set, this line would be a very simple `all_windows_.erase(this);`

 * It looks like `all_windows_` doesn't care about the order of the items and the same window will never occur twice, so `unordered_set` seems like the simplest & fastest container.

That said, the current impl should run. If you have a reason for usign a vector, e.g. some future work you intend to do that works better with a vector, :+1:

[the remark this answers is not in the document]
codebytere: cc @MarshallOfSound, do you want to address this prior to merge?

jkleinsc:
@MarshallOfSound looks like this needs a rebase

shell/browser/api/atom_api_native_theme.cc
```diff
@@ -36,6 +37,10 @@ void NativeTheme::SetThemeSource(ui::NativeTheme::ThemeSource override) {
   // Update the macOS appearance setting for this new override value
   UpdateMacOSAppearanceForOverrideValue(override);
 #endif
+  const bool dark_enabled = ShouldUseDarkColors();
+  for (auto* window : electron::api::TopLevelWindow::GetAllWindows()) {
```
zcbenz: You can iterate all windows with the `electron::WindowList` class. Also the `electron::` namespace prefix can be omitted here.

shell/browser/api/atom_api_native_theme.cc
```diff
@@ -36,6 +37,10 @@ void NativeTheme::SetThemeSource(ui::NativeTheme::ThemeSource override) {
...
+    window->SetGTKDarkThemeEnabled(dark_enabled);
+  }
   // TODO(MarshallOfSound): Update all existing browsers windows to use GTK dark
```
zcbenz: The `TODO` can be removed.

PR #20139, opened 2019-09-05
fix: don't expose nativeTheme in the renderer process

MarshallOfSound:
Exposing these in the renderer didn't make sense as they weren't backed by the same instance / value store.  This API should be browser only especially now that we have nativeTheme.themeSource.  Exposing in //common was a mistake from the beginning.

This along with all other `nativeTheme` API changes should be sent into `7-0-x` to avoid these changes having to land after the `nativeTheme` API lands in a stable release.

Phase 4 of #19932

Notes: Limited the `nativeTheme` API to the browser process only


MarshallOfSound:
> what's your plan for this in 7?

Due to the new / experimental nature of `nativeTheme` I plan to try land this (and the other follow up PRs to nativeTheme) into `7-0-x`

jkleinsc:
This was manually backported to 7-0-x via #20486

PR #20146, opened 2019-09-06
chore: remove native_mate (Part 4)

zcbenz:
#### Description of Change

This PR removes the patch on `gin::Dictionary` by using the `gin_helper::Dictionary` helper instead.

This PR also migrates some uses of `mate::Dictionary` to `gin`.

#### Release Notes

Notes: no-notes


shell/common/gin_converters/std_converter.h
```diff
@@ -30,6 +30,23 @@ struct Converter<unsigned long> {  // NOLINT(runtime/int)
...
+template <>
+struct Converter<const char*> {
+  static v8::Local<v8::Value> ToV8(v8::Isolate* isolate, const char* val) {
+    return v8::String::NewFromUtf8(isolate, val, v8::NewStringType::kNormal)
+        .ToLocalChecked();
+  }
+};
+
+template <size_t n>
+struct Converter<const char[n]> {
```
deepak1556: Can we replace the callsites to `StringPiece` so that this converter can be avoided.

zcbenz: This is for converting string literals, as C++ can not deduce their type to `const char*` in template. Without this converter it would be annoying that some string can be automatically converted while some can not, and lots of developers might find the error message hard to understand.

```c++
dict.Set("type", "renderer");
```

would emit error:

```
../../gin/converter.h:245:24: error: no member named 'ToV8' in 'gin::Converter<char [9], void>'
  return Converter<T>::ToV8(isolate, input);
         ~~~~~~~~~~~~~~^
```

Personally I think the converter has the value that C++ gives us the length of string in compile time so we can save a `strlen` call. (The performance gain does not really matter though.)

Issue #12458, opened 2018-03-28
Timeout parameter for Notification for Linux


uroybd:
Libnotify supports a timeout parameter which allows developers to control for how long the notification will stay. Some DEs has also implemented options to create persistent notification if the value is -1 for this parameter. Which is particularly helpful in GNOME since they removed System Tray altogether. Developers can create persistent notification instead of Tray Icon if the timeout is supported.

Please add this.

PR #20153, opened 2019-09-06
feat: allow Linux/Windows users to set notification timeout

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/12458.

Allows users to set timeout type of Linux and Windows notifications - either `default` or `never`.

cc @ckerr 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added `timeoutType` property to allows users to set timeout type for Linux  and Windows notifications.

### DIFF
--- a/docs/api/notification.md
+++ b/docs/api/notification.md
@@ -35,7 +35,7 @@ Returns `Boolean` - Whether or not desktop notifications are supported on the cu
   * `silent` Boolean (optional) - Whether or not to emit an OS notification noise when showing the notification.
   * `icon` (String | [NativeImage](native-image.md)) (optional) - An icon to use in the notification.
   * `hasReply` Boolean (optional) _macOS_ - Whether or not to add an inline reply option to the notification.
-  * `timeoutType` String (optional) _Linux_ - The timeout duration of the notification.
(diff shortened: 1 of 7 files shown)


docs/api/notification.md
```diff
@@ -144,6 +145,12 @@ A `Boolean` property representing whether the notification is silent.
 
 A `Boolean` property representing whether the notification has a reply action.
 
+#### `notification.timeoutType` _Linux_ _Windows_
+
+A `String` property representing the type of timeout duration for the notification. Can be 'default' or 'never'.
+
+If 'never' is set, te notification never expires. It stays open until closed by the calling API or the user.
```
jkleinsc:
```suggestion
If 'never' is set, the notification never expires. It stays open until closed by the calling API or the user.
```

shell/browser/api/atom_api_notification.h
```diff
@@ -78,6 +80,7 @@ class Notification : public mate::TrackableObject<Notification>,
   bool has_icon_ = false;
   bool silent_ = false;
   bool has_reply_ = false;
+  base::string16 timeout_type_;
```
miniak: Can we make it an enum class instead of string?

Issue #19908, opened 2019-08-23
desktopCapturer.getSources memory leaks


osan15:
### Issue Details

* **Electron Version:**
  * v4.0.0
* **Operating System:**
  * Ubuntu 18.04.3 LTS 
* **Last Known Working Electron version:**
  * Unknown

### Expected Behavior
The Resident set size of the Main process when using desktopCapturer.getSources not growing.

### Actual Behavior
Memory leak in the Main process (rss memory growing) when using the desktopCapturer.getSources method in the Rendering process.

### To Reproduce

> main.js
```javascript
const {
    app,
    BrowserWindow
} = require('electron');
const url = require('url');
const path = require('path');

let win;

const logBytes = x => `${x[0]}=${x[1] / (1000.0 * 1000)}MB`;

const logMemory = () => {
    console.log(
        Object.entries(process.memoryUsage())
        .map(logBytes)
        .join(', ')
    );
};

const createWindow = () => {
    win = new BrowserWindow();
    win.loadURL(url.format({
        pathname: path.join(__dirname, 'index.html'),
        protocol: 'file:',
        slashes: true
    }));
};

try {
    app.on('ready', createWindow);
} catch (e) {
    console.err(e);
}

logMemory();
setInterval(logMemory, 10000);

```

> index.html 
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8">
    <title>Hello World!</title>
  </head>
  <body>
    <h1>Hello World!</h1>
    We are using Node.js <span id="node-version"></span>,
    Chromium <span id="chrome-version"></span>,
    and Electron <span id="electron-version"></span>.

    <!-- You can also require other files to run in this process -->
    <script src="./renderer.js"></script>
  </body>
```

> renderer.js
```javascript
const electron = require('electron');
const desktopCapturer = electron.desktopCapturer;

const getSourcesLoop = () => {
    console.log('getSources() call');
    desktopCapturer.getSources({
        types: ['screen', 'window']
    }).then(sources => {
        console.log('getSources() success', sources);
        setTimeout(getSources, 5000);
    });
};
getSourcesLoop();
```

### Screenshots

![image](https://user-images.githubusercontent.com/10058753/63587467-64534d00-c5a4-11e9-9622-630e62cebbdf.png)

CapOM:
Can you try with electron 4.2.10 ?

miniak:
```
Electron v4.2.12 started.
[...]

PR #14835, opened 2018-09-27
refactor: desktop capturer module

deepak1556:
##### Description of Change

* Reverts https://github.com/electron/electron/commit/5a28759fea358c117c55121c1cd370abd143a207, which was added during 66 upgrade to fix desktop capturer. The specs were failing because the capturer was being created on a thread that disallows sync primitives

```
 Check failed: !g_base_sync_primitives_disallowed.Get().Get(). Waiting on a //base sync primitive is not allowed on this thread to prevent jank and deadlock. If waiting on a //base sync primitive is unavoidable, do it within the scope of a ScopedAllowBaseSyncPrimitives. If in a test, use ScopedAllowBaseSyncPrimitivesForTesting.
```
hence the workaround of creating it on a separate thread which was causing incorrect usage of weak ptr as seen in https://github.com/electron/electron/pull/14816, this PR instead fixes it be creating the capturers before the thread primitives is initialized similar to chrome.

* Creates the capturers for the lifetime of the app, the original design of this module relied on constantly creating and destroying the capturers for each api call which caused some race conditions during destruction and hence required to introduce a patch https://github.com/electron/libchromiumcontent/pull/418 . Its no longer required with this PR.

##### Checklist

- [ ] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [ ] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

##### Release Notes

Notes: no-notes

PR #20156, opened 2019-09-07
fix: free screen capturers immediately after we're finished with them

### LINKED ISSUE #17937: desktopCapturer.getSources with 'screen' causes binocular icon on lock screen
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 4.1.3
* **Operating System:**
  * macOS 10.14.4
  * None.

### Expected Behavior
A call to `desktopCapturer.getSources( { types: [ 'screen'] })` should not cause the OS to show the binoculars icon when the user locks the screen, this icon should only be shown when the screen is being recorded. 

As far as I know this icon is used to show that the user is recording a screen. [Here](https://macreports.com/binoculars-icon-on-mac-what-does-that-mean/) is explained this icon.

### Actual Behavior
Calling `desktopCapturer.getSources( { types: [ 'screen'] })` causes the binoculars icon to be shown to the user when he locks the PC, even when the screen is not being recorded in any other way. The icon isn't shown when the PC is unlocked, only when the PC is locked, until the users closes the electron app that made the call.

This doesn't happen when `desktopCapturer.getSources({ types: 'window' })` is called. 

### To Reproduce
Clone this repository

```
git clone https://github.com/electron/electron-quick-start
npm install
npm start
```
Then, in the `renderer.js` file, put:
```
const { desktopCapturer } = require('electron');

desktopCapturer.getSources( { types: [ 'screen'] }, (error, sources) => {
    if (error) throw error
    for (let i = 0; i < sources.length; ++i) {
      console.log('source: ' + sources[i].name); 
    }
});
```
Finally, while keeping the app opened, lock your screen and you will notice on the login screen the binoculars icon, even though there aren't any apps recording the screen. This icon will be seen on your lock screen as long as you have the electron app opened.  

### Screenshots

Screen Shot 2019-04-23 at 18 01 18

loc:
#### Description of Change
Fixes #17937 and #19908 

Originally fixed in #10223, but regressed in #14835, the fix here is to make sure our one-shot DesktopCapturer class destroys its `DesktopMediaList` instances (which in turn hold the real capturers) after it finishes gathering the thumbnails. If you'll recall, we use `DesktopMediaList` in a way that's different from Chrome; theirs is longer lived and capable of updating the thumbnails, ours is single-use, with the periodic thumbnail update overridden in a patch.

For good hygiene, we now send the reference to the cpp DesktopCapturer out to pasture after we're done with it (although after this change, a leak of DesktopCapturer would be much less costly, given the `DesktopMediaList`s are being reset).

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed desktopCapturer leak.

### DIFF
--- a/lib/browser/desktop-capturer.ts
+++ b/lib/browser/desktop-capturer.ts
@@ -19,7 +19,13 @@ export const getSources = (event: Electron.IpcMainEvent, options: ElectronIntern
   }
 
   const getSources = new Promise<ElectronInternal.GetSourcesResult[]>((resolve, reject) => {
+    let capturer: ElectronInternal.DesktopCapturer | null = createDesktopCapturer()
+
     const stopRunning = () => {
+      if (capturer) {
+        capturer.emit = null
+        capturer = null
+      }
       // Remove from currentlyRunning once we resolve or reject
       currentlyRunning = currentlyRunning.filter(running => running.options !== options)
     }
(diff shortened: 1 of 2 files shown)


Syed-Umair:
Hi team, will it be backported to 6-0-x

Issue #17059, opened 2019-02-20
Error using stream protocol in webview:  this.listenerCount is not a function


mjbvz:
* **Electron Version** (output of `node_modules/.bin/electron --version`):
  * 4.0.4
* **Operating System** (Platform and Version):
  *  macOS 10.14.2
* **Last known working Electron version** (if applicable):
  * None

### Expected Behavior
In a webview that uses its own partition and registers a custom stream protocol in that partition, we should be able to use the protocol as expected. This works with buffer protocols:

```js
// in renderer

const webview = document.createElement('webview');
webview.setAttribute('partition', `webview${Date.now()}`);

webview.addEventListener('did-start-loading', () => {
    // register protocol in webview
    const contents = webview.getWebContents();
    contents.session.protocol.registerStreamProtocol('atom', (request, callback) => {
        http.get({ hostname: 'example.com', }, res => {
            callback({
                statusCode: 200,
                headers: {
                    'content-type': 'text/html'
                },
                data: res
            })
        })
    }, (error) => {
        if (error) console.error('Failed to register protocol')
    });
});

webview.src = `data:text/html;charset=utf-8;base64,${Buffer.from(`<!DOCTYPE html>
<html lang="en" style="width: 100%; height: 100%">
<head>
	<title>Virtual Document</title>
</head>
<body style="margin: 0; overflow: hidden; width: 100%; height: 100%">
    <iframe src="atom://bla"></iframe>
</body>
</html>`).toString('base64')}`

document.body.append(webview);
```

This works as expected if the protocol is registered in the main process and an iframe is created in the renderer process

### Actual behavior
This does not work. In the main developer console, see:

```
Uncaught TypeError: this.listenerCount is not a function
    at Readable.on (_stream_readable.js:805)
    at CallbacksRegistry.apply (/Users/matb/Library/Application Support/Electron Fiddle/electron-bin/4.0.5/Electron.app/Contents/Resources/electron.asar/renderer/callbacks-registry.js:47)
    at handleMessage (/Users/matb/Library/Application Support/Electron Fiddle/electron-bin/4.0.5/Electron.app/Contents/Resources/electron.asar/renderer/api/remote.js:285)
    at EventEmitter.ipcRenderer.on.args (/Users/matb/Library/Application Support/Electron Fiddle/electron-bin/4.0.5/Electron.app/Contents/Resources/electron.asar/renderer/api/remote.js:275)
    at EventEmitter.emit (events.js:182)
```

### To Reproduce
[...]

PR #20184, opened 2019-09-10
fix: remote.protocol.{register|intercept}StreamProtocol can work now under certain conditions

ianho:
#### Description of Change
For issue #17059
`remote.protocol.{register|intercept}StreamProtocol` not work.
This commit wont fix all the case because the `remote ipc` is sync but `stream` is async.
At least `remote.{register|intercept}StreamProtocol` can work under certain conditions now.

#### Checklist
- [x] `npm test` passes

#### Release Notes

Notes: Fixed remote stream protocol now working

### DIFF
--- a/lib/browser/rpc-server.js
+++ b/lib/browser/rpc-server.js
@@ -15,6 +15,7 @@ const ipcMainUtils = require('@electron/internal/browser/ipc-main-internal-utils
 const objectsRegistry = require('@electron/internal/browser/objects-registry')
 const guestViewManager = require('@electron/internal/browser/guest-view-manager')
 const bufferUtils = require('@electron/internal/common/buffer-utils')
+const streamUtils = require('@electron/internal/common/readable-stream-utils')
 const errorUtils = require('@electron/internal/common/error-utils')
 const clipboardUtils = require('@electron/internal/common/clipboard-utils')
 const { isPromise } = require('@electron/internal/common/is-promise')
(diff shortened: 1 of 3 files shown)


vincentdchan:
How long it can be merged! Can't wait to use it!

ianho:
How can I retrigger the CI pipline?
I found that win-ia32 build succeeded but win-x64 build failed.
I think my changes wont cause this problem.

ianho:
I think “reading all the data into memory takes very long time” is very convincing for me.
But this issue block my work in our project. I must use api `remote.protocol.{register|intercept}StreamProtocol` because of some complicated reasons which I cant control.
Api `remote.protocl.{register|intercept}BufferProtocol` cant return with headers, so it wont work too.
Otherwise, if api `remote.protocol.{register|intercept}StreamProtocol` cant work, maybe we can tell people that in doc. Or we can tell people using this api with remote will take very long time?

ianho:
> I agree with @zcbenz it makes no sense to use a `stream` protocol then load the entire stream into memory. This is terrible for performance and is terribly misleading for folks using a stream protocol.
> 
> Because of the way streams work making them work over remote is basically impossible. If this is what you want to achieve you should use a buffer protocol not a stream protocol.

I said I agree with the opinion. But I think if we dont want to support stream protocl with remote way, maybe we should tell people that instead of provide a api with bug.
And about buffer protocol, it cant response with headers. Maybe we should support response headers via buffer protocol ?

zcbenz:
I'm closing this PR as so far no maintainer thinks we should solve the problem in this direction.

PR #20214, opened 2019-09-12
refactor: use v8 serialization for ipc

nornagon:
#### Description of Change

BREAKING CHANGE

This changes IPC communication to use v8's [Structured Clone](https://html.spec.whatwg.org/multipage/structured-data.html#safe-passing-of-structured-data) algorithm instead of using the `base::Value` serialization defined in `native_mate_converters/v8_value_converter.cc`. This is be faster, more featureful, and less surprising than the existing logic, since it more-or-less matches the logic that backs `postMessage`.

This brings about a 2x performance boost for large buffers and complex objects. Latency for small messages is not significantly affected.

User-observable differences from the existing IPC API:
- It's about 2x faster.
- `NaN`, `Infinity`, and `undefined` are transferred as such, rather than being converted to `null`.
- Cyclic objects can be transmitted.
- `Set` and `Map`, `Error`, `RegExp`, `Date` and `BigInt` can be transmitted.
- `Buffer` will be converted to `Uint8Array`.
- Typed arrays (`Float32Array` and friends) will be transmitted as they are, instead of being converted to `Buffer`.
- Sparse arrays will be transferred as sparse arrays, instead of being converted to dense arrays with `null`s.

> **NOTE**: Objects that aren't serializable with V8's Structured Clone algorithm, such as functions, DOM objects, special Node/Electron objects like `process.env` or `WebContents`, or any objects containing such items will be serialized with the old `base::Value`-based algorithm. However, this behavior is deprecated and will throw an exception beginning with Electron 9.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: IPC between main and renderer processes now uses the Structured Clone Algorithm.


lib/browser/chrome-extension.js
```diff
@@ -240,7 +240,7 @@ const getMessagesPath = (extensionId) => {
 
 ipcMainUtils.handleSync('CHROME_GET_MESSAGES', async function (event, extensionId) {
   const messagesPath = getMessagesPath(extensionId)
-  return fs.promises.readFile(messagesPath)
+  return fs.promises.readFile(messagesPath).then(b => b.toString('utf8'))
```
MarshallOfSound: Can we not send buffers over IPC anymore??

nornagon: you totally can, but there's no particular reason to here, and the remote end will end up as a Uint8Array, which doesn't have this form of `toString()`.

nornagon:
Remaining TODOs as of latest commit:

- [ ] Handle errors gracefully & correctly.
- [ ] Investigate whether we can discriminate between `Buffer` and `Uint8Array` wisely, and deserialize `Buffer`s as `Buffer`s where we can, and `Uint8Array` where we can't (e.g. in sandboxed renderers).
- [ ] Convert the remaining methods to the new serialization algorithm (`webContents.send`, `messageTo`, and `messageHost`).

nornagon:
OK, there's a choice to make about how we handle when non-POJO objects are passed to IPC functions (things like DOM objects, `WebContents`, `process.env`, etc.).

1. Throw an exception.
    - This is the safest option, but also the most disruptive. (e.g. many of Electron's internal APIs pass such objects to IPC methods, and will need to be refactored.)
2. JSON.stringify()/JSON.parse() them.
    - This is roughly similar to current behavior of the IPC methods, in that it can't handle NaN or Infinity, or cyclic objects. However, it might be somewhat confusing to have some objects work differently based on whether they're referenced from a non-POJO object or not. It's a fairly obscure corner case though, so maybe nobody would notice?
3. Iterate the keys/values of such objects and recurse in the serialization algorithm.
    - This gives us cyclic object support and correct handling of Infinity/NaN/etc.
    - But most non-trivial non-POJO objects contain unserializable things like functions.
    - We can attempt to replace those unserializable things with `undefined`, but it only works one layer deep. i.e. if there's a non-POJO object like `{ fn: () => {} }`, it'll be turned into `{ fn: undefined }`, but if it's like `{ events: { fn: () => {} } }`, then the entire serialization operation must be aborted and an error thrown.

I'm leaning towards (1), since I don't think it really makes sense to be trying to send `window.location` or a `WebContents` value over IPC, and places where we're doing that are likely error-prone. Thoughts?

MarshallOfSound:
I don't like the idea of existing code throwing errors in an obscure and un-lintable way.  i.e. we can't provide a lint rule or other tool to detect invalid usages of `ipc.send` so folks will only find out at runtime when it's too late and hard to spot.

I'm leaning towards something equivalent to (2) where if the v8 serializer fails we fall back to the old one.  Maybe log a console warning that you tried to send an unserializable object and we fell back to a less accurate converter?

nornagon:
Does that leave any path towards eventually eliminating the old converter? I'm disinclined to leave it there forever... but maybe it would make sense as a less-dramatic deprecation strategy? e.g. "Looks like you sent a non-standard object over IPC. This is probably not what you intended to do, but we obliged and serialized it anyway. Look out, this will throw an exception in Electron 9."

MarshallOfSound:
I'd be down for a **very** long term deprecation strategy here.  Like we should support serialization for I'd say 4 versions beyond when this lands (number picked to match the number of versions we used for the native module deprecation).

The deprecation log should include the object itself as well if possible 🤔

nornagon:
@MarshallOfSound I'm not sure that's warranted. 99% of cases where IPC is used, the new version should still work fine. Only in weird cases where it only kind of semi-works currently are problematic.

For example. The `new-window` event has as one of its arguments the `options` parameter, which is the options passed to `new BrowserWindow()`. One of these options is `webContents`, which specifies which WebContents will go in the new window. For WebViews, the `new-window` event is proxied over to the host frame. However, during that transmission, the `webContents` key stops being an actual `WebContents`. Is anything depending on that key being there in the webview host? I don't think so. Can anything actually useful be done with it there? Again, I don't think so. The fact that it's there is an accident, and it never really performed any kind of useful function.

Also, it's trivial to get something close to the old behavior with the transformation `x => JSON.parse(JSON.stringify(x))`, as long as `x` isn't cyclic.

tldr most people wouldn't need to fix anything if this usage started throwing an error, and for those that do, the fix is easy.

MarshallOfSound:
> if this sort of usage started throwing an error

My only concern here was that you can't glance at a bit of code and say "that's gonna cause a problem".  In fact I don't think you could describe the kinds of things that would cause this problem very easily at all.  That's the only reason I'm advocating for a deprecation cycle, it will take time for people to find all the edge cases in their app that might do this.  Maybe 4 versions was a bit over-eager 🤷‍♂But it should definitely go through a cycle where the old code still works but tells you you're doing something bad.

nornagon:
The things which will throw exceptions with structured clone that don't with base::Value serialization:
- DOM objects (`Element`, `Location`, etc.)
- special Node objects like `process.env` or http internals
- special Electron API objects like WebContents
- functions

I've updated the PR to fall back to the base::Value-based serialization if structured clone fails, and emit a deprecation warning.

nornagon:
The last remaining tricky thing is that sending a `Buffer` will result in the other side receiving a `Uint8Array`. Without patching V8, I don't see a good way to get around this. Fortunately, you can wrap a `Buffer` around a `Uint8Array` quite easily:

```
const uint8Array = await ipcRenderer.invoke('return-a-buffer')
const buf = Buffer.from(uint8Array.buffer)
```

nornagon:
Taking off the [wip] as I believe this is ready for full review.

shell/renderer/api/atom_api_renderer_ipc.cc
```diff
@@ -69,45 +70,71 @@ class IPCRenderer : public gin::Wrappable<IPCRenderer> {
...
 
   v8::Local<v8::Promise> Invoke(v8::Isolate* isolate,
                                 bool internal,
                                 const std::string& channel,
-                                const base::Value& arguments) {
-    electron::util::Promise<base::Value> p(isolate);
+                                v8::Local<v8::Value> arguments) {
+    blink::CloneableMessage message;
+    if (!mate::ConvertFromV8(isolate, arguments, &message)) {
+      return v8::Local<v8::Promise>();
```
MarshallOfSound: Is this promise rejected or does the method throw an error and return an empty promise?

nornagon: When the serialization fails, the method throws an error. Also, this isn't a `Promise` value, it's an empty `Local`.

MarshallOfSound: My concern here was is that error still catchable with .catch?

If it's an empty promise and the method throws then .catch won't work which is incompatible with a promise based API

nornagon: yes, it's still catchable.

```js
ipcRenderer.invoke('foo', ()=>{}).catch(e => console.log('bloop'))
// => bloop
```

patches/chromium/export_fetchapi_mojo_traits_to_fix_component_build.patch
```diff
@@ -0,0 +1,36 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jeremy Apthorp <jeremya@chromium.org>
+Date: Fri, 20 Sep 2019 16:44:18 -0400
+Subject: export FetchAPI mojo traits to fix component build
+
+Without these, we get link errors in the component build when using the
+blink::CloneableMessage mojo traits.
```
deepak1556: Is there a CL for this ?

docs/api/breaking-changes.md
```diff
@@ -6,6 +6,59 @@ Breaking changes will be documented here, and deprecation warnings added to JS c
 
 The `FIXME` string is used in code comments to denote things that should be fixed for future releases. See https://github.com/electron/electron/search?q=fixme
 
+## Planned Breaking API Changes (8.0)
```
jkleinsc: Are we targeting this for 8 or 9?  The PR says 9.

nornagon: This PR is targeted to 8. When 9 rolls around, the behavior will change again (the deprecated things will start throwing an exception instead of being serialized with the old algorithm).

jkleinsc: @nornagon doesn't that mean it will be `Planned Breaking API Changes (9.0)` in the doc?

jkleinsc: Ehh... nevermind I misread this.

patches/chromium/export_fetchapi_mojo_traits_to_fix_component_build.patch
nornagon: Not yet, but I talked to pfeldman@ and he said he'd approve it if i uploaded it :P will do soon.

lib/browser/guest-view-manager.js
```diff
@@ -48,6 +47,13 @@ const supportedWebViewEvents = [
 const guestInstances = {}
 const embedderElementsMap = {}
 
+function sanitizeOptionsForGuest (options) {
+  const ret = { ...options }
```
miniak: what about?
```js
const { webContents, ...ret } = options
return ret
```

lib/browser/chrome-extension.js
miniak: what about?
```js
const buffer = await fs.promises.readFile(messagesPath)
return toString('utf8')
```

miniak: you can do just
```js
return fs.promises.readFile(messagesPath, 'utf8')
```

lib/browser/guest-view-manager.js
miniak: what about?
```js
// WebContents values can't be sent over IPC.
const { webContents, ...ret } = options
return ret
```

shell/common/api/api.mojom
```diff
@@ -1,19 +1,25 @@
...
       bool internal,
       bool send_to_all,
       string channel,
-      mojo_base.mojom.ListValue arguments,
+      blink.mojom.CloneableMessage arguments,
       int32 sender_id);
 
   UpdateCrashpadPipeName(string pipe_name);
 
+  // This is an API specific to the "remote" module, and will ultimately be
```
miniak: `#if BUILDFLAG(ENABLE_REMOTE_MODULE)`

shell/browser/api/atom_api_web_contents.cc
```diff
@@ -1047,24 +1047,35 @@ void WebContents::MessageTo(bool internal,
...
 
 void WebContents::MessageHost(const std::string& channel,
-                              base::Value arguments) {
+                              blink::CloneableMessage arguments) {
   // webContents.emit('ipc-message-host', new Event(), channel, args);
   EmitWithSender("ipc-message-host", bindings_.dispatch_context(),
                  base::nullopt, channel, std::move(arguments));
 }
 
+void WebContents::DereferenceRemoteJSObject(const std::string& context_id,
```
miniak: #if BUILDFLAG(ENABLE_REMOTE_MODULE)

shell/common/api/api.mojom
```diff
@@ -64,11 +70,18 @@ interface ElectronBrowser {
...
     string channel,
-    mojo_base.mojom.ListValue arguments);
+    blink.mojom.CloneableMessage arguments);
 
   MessageHost(
     string channel,
-    mojo_base.mojom.ListValue arguments);
+    blink.mojom.CloneableMessage arguments);
+
+  // This is an API specific to the "remote" module, and will ultimately be
```
miniak: #if BUILDFLAG(ENABLE_REMOTE_MODULE)

shell/renderer/electron_api_service_impl.cc
```diff
@@ -168,12 +172,36 @@ void ElectronApiServiceImpl::Message(bool internal,
...
         v8::Local<v8::Context> child_context =
             renderer_client_->GetContext(child->ToWebLocalFrame(), isolate);
-        EmitIPCEvent(child_context, internal, channel, arguments.GetList(),
-                     sender_id);
+        EmitIPCEvent(child_context, internal, channel, args, sender_id);
       }
   }
 }
 
+void ElectronApiServiceImpl::DereferenceRemoteJSCallback(
```
miniak: #if BUILDFLAG(ENABLE_REMOTE_MODULE)

shell/renderer/electron_api_service_impl.h
```diff
@@ -28,8 +28,10 @@ class ElectronApiServiceImpl : public mojom::ElectronRenderer,
   void Message(bool internal,
                bool send_to_all,
                const std::string& channel,
-               base::Value arguments,
+               blink::CloneableMessage arguments,
                int32_t sender_id) override;
+  void DereferenceRemoteJSCallback(const std::string& context_id,
```
miniak: #if BUILDFLAG(ENABLE_REMOTE_MODULE)

docs/api/breaking-changes.md
```diff
@@ -6,6 +6,59 @@ Breaking changes will be documented here, and deprecation warnings added to JS c
...
+  arrays with `null`s.
+- `Date` objects will be transferred as `Date` objects, instead of being
+  converted to their ISO string representation.
+- Typed Arrays (such as `Uint8Array`, `Uint16Array`, `Uint32Array` and so on)
+  will be transferred as such, instead of being converted to Node.js `Buffer`.
+- Node.js `Buffer` objects will be transferred as `Uint8Array`s. You can
+  convert a `Uint8Array` back to a Node.js `Buffer` by wrapping the underlying
+  `ArrayBuffer`:
+```js
+Buffer.from(someUint8Array.buffer)
```
miniak: it needs to be `Buffer.from(value.buffer, value.byteOffset, value.byteLength)`

docs/api/breaking-changes.md
```diff
@@ -6,6 +6,59 @@ Breaking changes will be documented here, and deprecation warnings added to JS c
...
+
+// From Electron 8:
+ipcRenderer.send('channel', { value: 3, someFunction: () => {} })
+// => throws Error("() => {} could not be cloned.")
+```
+- `NaN`, `Infinity` and `-Infinity` will now be correctly serialized, instead
+  of being converted to `null`.
+- Objects containing cyclic references will now be correctly serialized,
+  instead of being converted to `null`.
+- `Set`, `Map`, `Error` and `RegExp` values will be correctly serialized,
```
miniak: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm
> Error and Function objects cannot be duplicated by the structured clone algorithm; attempting to do so will throw a DATA_CLONE_ERR exception.

docs/api/breaking-changes.md
nornagon: Doesn't seem like it:

```
> u8b = Uint8Array.from([1,2,3])
Uint8Array [ 1, 2, 3 ]
> Buffer.from(u8b.buffer)
<Buffer 01 02 03>
```

nornagon: oh i see, you mean for when it's a view on a larger buffer. sure.

docs/api/breaking-changes.md
nornagon: DevTools_-_github_com_electron_electron_pull_20214

spec/webview-spec.js
```diff
@@ -1068,7 +1068,7 @@ describe('<webview> tag', function () {
       await loadWebView(webview, { src })
 
       const data = await webview.printToPDF({})
-      expect(data).to.be.an.instanceof(Buffer).that.is.not.empty()
+      expect(data).to.be.an.instanceof(Uint8Array).that.is.not.empty()
```
miniak: the documentation for `<webview>.printToPDF(options)` should specify the return value as `Promise<Uint8Array >` instead of `Promise<Buffer>`

docs/api/breaking-changes.md
miniak: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm
> Error and Function objects cannot be duplicated by the structured clone algorithm; attempting to do so will throw a DATA_CLONE_ERR exception.

docs/api/breaking-changes.md
miniak: it needs to be `Buffer.from(value.buffer, value.byteOffset, value.byteLength)` as the underlying `ArrayBuffer` can be larger with some invalid data beyond the current view

miniak:
this will allow us to also simplify the serialization in the `remote` module. #20427

shell/common/api/api.mojom
nornagon: Mojom files can't have `#if` in them unfortunately. I'm not sure it's worth splitting this out into an optionally-offered interface, but I can make the implementations `NOTREACHED()`.

lib/browser/guest-view-manager.js
nornagon: I think that'd cause an 'unused variable' warning...

docs/api/breaking-changes.md
nornagon: I'm not sure why you deleted my previous comment displaying a screenshot of this working, but here it is again:

DevTools_-_chrome-search___local-ntp_local-ntp_html

shell/common/api/api.mojom
deepak1556: https://cs.chromium.org/chromium/src/mojo/public/tools/bindings/mojom.gni?l=274 can help build optional features into mojom files.

spec/webview-spec.js
nornagon: Unfortunately this currently produces an error in our typings generator:

```
../electron.d.ts(10212,62): error TS2694: Namespace 'Electron' has no exported member 'Uint8Array'.
```

nornagon: https://github.com/electron/typescript-definitions/pull/156 fixes

docs/api/breaking-changes.md
miniak: @nornagon that must have been a mistake, sorry for that. I was deleting my own comments as I've added them as single comments first, but then switched them to review comments.

spec-main/api-ipc-renderer-spec.ts
```diff
@@ -104,7 +105,7 @@ describe('ipcRenderer module', () => {
       expect(childValue).to.deep.equal(child)
     })
 
-    it('inserts null for cyclic references', async () => {
+    it('can handle cyclic references', async () => {
```
ckerr: :tada:

lib/browser/rpc-server.js
```diff
@@ -114,7 +114,7 @@ ipcMainUtils.handleSync('ELECTRON_BROWSER_SANDBOX_LOAD', async function (event)
     process: {
       arch: process.arch,
       platform: process.platform,
-      env: process.env,
+      env: { ...process.env },
```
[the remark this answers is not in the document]
nornagon: `process.env` is a C++-backed "special" JS object, so it can't be serialized by Structured Clone. This code would work unmodified currently, but it would fall back on the old `base::Value` serialization and emit a deprecation warning, and begin throwing an exception in Electron 9.

shell/browser/api/atom_api_web_contents.cc
```diff
@@ -1017,7 +1017,7 @@ void WebContents::OnElectronBrowserConnectionError() {
 
 void WebContents::Message(bool internal,
                           const std::string& channel,
-                          base::Value arguments) {
+                          blink::CloneableMessage arguments) {
```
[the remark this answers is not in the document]
nornagon: `blink::CloneableMessage`'s copy constructor is deleted, which ensures that this is called with `std::move` or `CloneableMessage::ShallowClone`.

shell/common/native_mate_converters/blink_converter.cc
```diff
@@ -527,4 +532,182 @@ bool Converter<network::mojom::ReferrerPolicy>::FromV8(
...
+  explicit V8Serializer(v8::Isolate* isolate,
+                        bool use_old_serialization = false)
+      : isolate_(isolate),
+        serializer_(isolate, this),
+        use_old_serialization_(use_old_serialization) {}
+  ~V8Serializer() override = default;
+
+  bool Serialize(v8::Local<v8::Value> value, blink::CloneableMessage* out) {
+    serializer_.WriteHeader();
+    if (use_old_serialization_) {
```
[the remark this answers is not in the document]
nornagon: It is used, see L564 in this file. It will be deleted, along with the serialization tags, in Electron 9.

shell/common/native_mate_converters/blink_converter.cc
```diff
@@ -527,4 +532,182 @@ bool Converter<network::mojom::ReferrerPolicy>::FromV8(
...
+
+  void WriteTag(uint8_t tag) { serializer_.WriteRawBytes(&tag, 1); }
+
+  // v8::ValueSerializer::Delegate
+  void* ReallocateBufferMemory(void* old_buffer,
+                               size_t size,
+                               size_t* actual_size) override {
+    data_.resize(size);
+    *actual_size = data_.capacity();
+    return data_.data();
```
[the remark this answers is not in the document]
nornagon: Chromium (and thus Electron) is built with `-fno-exceptions`, so unfortunately I don't think we can catch the error here. I'm not totally sure what the stdlib's behavior will be when there's an exception thrown during vector resize...

shell/renderer/api/atom_api_renderer_ipc.cc
```diff
@@ -69,45 +70,71 @@ class IPCRenderer : public gin::Wrappable<IPCRenderer> {
...
-  void Send(bool internal,
+  void Send(v8::Isolate* isolate,
+            bool internal,
             const std::string& channel,
-            const base::ListValue& arguments) {
-    electron_browser_ptr_->get()->Message(internal, channel, arguments.Clone());
+            v8::Local<v8::Value> arguments) {
+    blink::CloneableMessage message;
+    if (!mate::ConvertFromV8(isolate, arguments, &message)) {
+      return;
```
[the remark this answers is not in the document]
nornagon: In Electron 7 this will only be hit if both the V8 serializer and the `base::Value` serializer fail. I'm not exactly sure what circumstances the `base::Value` serializer can fail in, but those ones. In Electron 8 this will throw.

The `EmitDeprecationWarning` is taken care of by the `base::Value` path in the serializer.

docs/api/ipc-renderer.md
```diff
@@ -69,9 +75,15 @@ The main process handles it by listening for `channel` with the
 
 Returns `Promise<any>` - Resolves with the response from the main process.
 
-Send a message to the main process asynchronously via `channel` and expect an
-asynchronous result. Arguments will be serialized as JSON internally and
-hence no functions or prototype chain will be included.
+Send a message to the main process via `channel` and expect a result
```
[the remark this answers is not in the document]
nornagon: I don't think the distinction is worth the grammatical awkwardness, plus there are certainly elements of the sending that happen synchronously (i.e. the serialization).

spec/webview-spec.js
nornagon: electron/typescript-definitions#157 _actually_ fixes.

shell/common/native_mate_converters/blink_converter.cc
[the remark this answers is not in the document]
nornagon: I asked about this in #cxx on the Chromium workspace and it looks like the the exception, if thrown, will crash the process. That seems like reasonable behavior to me, so I'm going to resolve this.

lib/browser/rpc-server.js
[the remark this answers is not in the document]
ckerr: Sounds good, thanks for the explanation. I didn't realize `process.env` was a special case compared to other objects.

Xylobol:
For anyone else interested, these changes have landed in nightly - https://github.com/electron/nightlies/releases/tag/v8.0.0-nightly.20191009

gdavidkov:
Will this be backported to 6.x?

nornagon:
No, since it is a breaking change.

Slapbox:
This PR is noted as a **_Breaking Changes_** for Electron 8.0.0 & 9.0.0 betas.  Is there any difference in the IPC between versions 8 and 9?

pushkin-:
@Slapbox  I believe the plan in 8 is to use the new algorithm, but if it fails, emit a warning and fallback to the old algorithm.

In Electron 9, they're going remove the fallback and warning and serializing bad data will start throwing exceptions.

Qrysto:
Is there any way to check if a variable is safe to be transferred through RPC calls? Could `JSON.stringify` be used to test it?

nornagon:
@Qrysto The most reliable way to check for serializability is to try sending the message, and catching the error. If for some odd reason you need to check in a way that's guaranteed not to send a message, you could try sending to a dummy channel that has no listeners, or if you're in the main process or an unsandboxed renderer, you can try the [v8.serialize Node.JS API](https://nodejs.org/api/v8.html#v8_v8_serialize_value). `JSON.stringify` isn't sufficient to determine serializability.

PR #20224, opened 2019-09-13
docs: improve and add examples for clipboard

codebytere:
#### Description of Change

Most lacked examples and some lacked contextual info. I've added that, and clarified some ambiguous terminology.

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/clipboard.md
+++ b/docs/api/clipboard.md
@@ -4,18 +4,12 @@
 
 Process: [Main](../glossary.md#main-process), [Renderer](../glossary.md#renderer-process)
 
-The following example shows how to write a string to the clipboard:
-
-```javascript
-const { clipboard } = require('electron')
-clipboard.writeText('Example String')
-```
-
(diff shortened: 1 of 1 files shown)


docs/api/clipboard.md
```diff
@@ -28,56 +22,122 @@ The `clipboard` module has the following methods:
 
 ### `clipboard.readText([type])`
 
-* `type` String (optional) - Can be `selection` or `clipboard`. `selection` is only available on Linux.
+* `type` String (optional) - Can be `selection` or `clipboard`. Default is 'clipboard'.
```
jkleinsc: I think it is important to keep the qualifier here that selection is only available on Linux instead of putting that in the description.
```suggestion
* `type` String (optional) - Can be `selection` or `clipboard`. `selection` is only available on Linux. Default is 'clipboard'.
```

codebytere: i'll undo that change!

PR #20230, opened 2019-09-13
docs: Performance checklist

felixrieseberg:
#### Description of Change

This PR introduces a performance checklist. It's a bit self-documenting, but I'd like to highlight that it's not meant to be prescriptive or exhaustive.

For internal discussion, see `#proj-performance-docs`.

#### Checklist

#### Release Notes

Notes: Added a performance checklist


PalmerAL:
This seems fairly comprehensive, and it covers a lot of the things I've run into in the past. A few more things that it might make sense to add:

* Even if your app shares code with your website, you should bundle resources in your app rather than loading them over HTTP.
* In most cases, it's better to bundle your code together rather than trying to require() everything on startup.
* I know you mentioned it already, but it might be a good idea to emphasize more just how slow `remote` is, since it's one of the largest Electron-specific sources of slowness.
* Minimizing resource usage when the app is running in the background.

It might also be interesting to talk about the differences in user expectations between a website and a desktop app. I think most people have gotten used to websites being somewhat slow, but don't have the same expectations for desktop apps, so if you're working with Electron for the first time, you may need to focus on performance more than you would in the past.

MarshallOfSound:
> Even if your app shares code with your website, you should bundle resources in your app rather than loading them over HTTP.

I wouldn't recommend this this point blank.  There are significant cons to this, namely version drift and the fact that updating your website now means you have to update your app as well.  Their are significant pros and cons to live-loading and shipping in-app and I don't think we can recommend one without a comprehensive list of those pros and cons which I don't think belongs in this document.

> In most cases, it's better to bundle your code together rather than trying to require() everything on startup.

👍 Especially on windows, file system ops are slow ™️ 

> Minimizing resource usage when the app is running in the background.

Mentioning putting things into `requestAnimationFrame` or checking `document.visibilityState` might be good here

issacgerges:
Really great stuff @felixrieseberg! This is going to make a massive impact. Just some stuff that  comes to mind when reading:

* My number one tool for demonstrating the cost of work in the main process has been Johannes' fiddle. Curious if it would be worth including examples  like this one when it makes sense https://twitter.com/johannesrieken/status/1086311961128620032

* Love the warning around `remote`,  but I wonder if `remote.getGlobal` deserves special attention. I think people rarely understand the performance trade-off there, partly because the API hides a lot of the complexity around this almost magic front. I think it may be hard for people to understand that calls like `remote.getGlobal('settings').getTheme().topBar.color` is around 10 `ipc` messages back and forth.

* For unnecessary polyfills, it might be worthwhile to mention that `electron` is a valid target as a `browserlist` and can be used with `@babel/preset-env`

* It's well documented, but I wonder if the `Measure,  Measure, Measure` section deserves to link out to https://electronjs.org/docs/tutorial/debugging-main-process.

* Should `devtron` be called out? Might be some overlap with the require graph / ipc stuff and some of what you've called out here.

docs/tutorial/performance.md
```diff
@@ -0,0 +1,347 @@
...
+will require you to go beyond a number of steps. Instead, you will have to
+closely examine all the code running in your app by carefully profiling and
+measuring. Where are the bottlenecks? When the user clicks a button, what
+operations take up the brunt of the time? While the app is simply idling, which
+objects take up the most memory?
+
+Time and time again, we have seen that the most successful strategy for building
+a performant Electron app is to profile the running code, find the most
+resource-hungry piece of it, and to optimize it. Repeating this seemingly process
+over and over again will dramatically reduce your app's performance. Experience
```
Kilian: reduce -> increase

docs/tutorial/performance.md
```diff
@@ -0,0 +1,347 @@
...
+easy to implement. However, building the most performant version of your app
+will require you to go beyond a number of steps. Instead, you will have to
+closely examine all the code running in your app by carefully profiling and
+measuring. Where are the bottlenecks? When the user clicks a button, what
+operations take up the brunt of the time? While the app is simply idling, which
+objects take up the most memory?
+
+Time and time again, we have seen that the most successful strategy for building
+a performant Electron app is to profile the running code, find the most
+resource-hungry piece of it, and to optimize it. Repeating this seemingly process
```
Kilian: `Repeating this seemingly [something] process`

laborous? repetetive?

docs/tutorial/performance.md
```diff
@@ -0,0 +1,430 @@
...
+
+### Recommended Reading
+
+ * [Get Started With Analyzing Runtime Performance][chrome-devtools-tutorial]
+ * [Talk: "Visual Studio Code - The First Second"][vscode-first-second]
+
+## Checklist
+
+Chances are that your app could be a little leaner, faster, and generally less
+resource-hungry if you attempt these steps.
```
Kilian: 1 through 6 are titled as things you should not do, 7 is something you _should_ do. Combined with "attempt these steps" above, it seems like all things you should do, so maybe rewrite the intro line to:

```
Chances are that your app could be a little leaner, faster, and generally less 
resource-hungry if you review the following checks.
```

Or change the titles of 1 through 6 to be the thing you _should_ do, which is mostly adding "Don't" in front of the current titles.

docs/tutorial/performance.md
```diff
@@ -0,0 +1,430 @@
...
+5) [Unnecessary polyfills](#5-unnecessary-polyfills)
+6) [Unnecessary or blocking network requests](#6-unnecessary-or-blocking-network-requests)
+7) [Bundle your code](#7-bundle-your-code)
+
+## 1) Carelessly including modules
+
+Before adding a Node.js module to your application, examine said module. How
+many dependencies does that module include? What kind of resources does said
+module need to simply be called in a `require()` statement? You might find
+that the module with the most downloads on `npm` or the most stars on GitHub
```
malept: :bike: :derelict_house: I'd say that `npm` is the CLI, not the package registry.

```suggestion
that the module with the most downloads on the NPM package registry or the most stars on GitHub
```

docs/tutorial/performance.md
```diff
@@ -0,0 +1,430 @@
...
+storing in memory information that it does not actually need.
+
+In short, a seemingly excellent module written primarily for Node.js servers
+running Linux might be bad news for your app's performance. In this particular
+example, the correct solution was to use no module at all, and to instead use
+connectivity checks included in later versions of Chromium.
+
+### How?
+
+When considering a module, we recommend that you check 1) the size of dependencies
```
malept: I think a legit ordered list is easier to read.

```suggestion
When considering a module, we recommend that you check:

1. the size of dependencies included
```

docs/tutorial/performance.md
```diff
@@ -0,0 +1,430 @@
...
+included; 2) resources required to load (`require()`) it; 3) resources required
```
malept:
```suggestion
2) the resources required to load (`require()`) it
```

docs/tutorial/performance.md
```diff
@@ -0,0 +1,430 @@
...
+to perform the action you're interested in.
```
malept:
```suggestion
3. the resources required to perform the action you're interested in
```

docs/tutorial/performance.md
```diff
@@ -0,0 +1,430 @@
...
+
+It is rare for a JavaScript-based polyfill to be faster than the equivalent
+native feature in Electron. Do not slow down your Electron app by shipping your
+own version of standard web platform features.
+
+### How?
+
+Operate under the assumption that polyfills in current versions of Electron
+are unnecessary. If you have doubts, check [caniuse.com][https://caniuse.com/]
+and check if the version of Chromium used in your Electron version supports
```
malept:
```suggestion
and check if the [version of Chromium used in your Electron version](/docs/api/process#processversionschrome) supports
```

docs/tutorial/performance.md
```diff
@@ -0,0 +1,430 @@
...
+will take care of the rest.
+
+When building an Electron app, your users are better served if you download
+the fonts and include them in your app's bundle.
+
+### How?
+
+In an ideal world, your application wouldn't need the network to operate at
+all. To get there, you must understand what resources your app is downloading
+- and how large those resources are.
```
malept: This is rendering as a bullet point:

```suggestion
\- and how large those resources are.
```

docs/tutorial/performance.md
```diff
@@ -0,0 +1,432 @@
...
+
+It is rare for a JavaScript-based polyfill to be faster than the equivalent
+native feature in Electron. Do not slow down your Electron app by shipping your
+own version of standard web platform features.
+
+### How?
+
+Operate under the assumption that polyfills in current versions of Electron
+are unnecessary. If you have doubts, check [caniuse.com][https://caniuse.com/]
+and check if the [version of Chromium used in your Electron version](/docs/api/process#processversionschrome) supports
```
malept: The linter did not like my link :cry:

Maybe this instead?

```suggestion
and check if the [version of Chromium used in your Electron version](../api/process#processversionschrome) supports
```

malept:
This needs to go into the 7-0-x branch for it to be deployed to the website, right? (Ignoring the fact that it's not in the docs index page yet...)

PR #20239, opened 2019-09-15
fix: backgroundThrottling to throttle render view

gerhardberger:
#### Description of Change

* Previously when a render view was created the corresponding RenderWidgetHost was fetched with the wrong routing ID which caused that the `backgroundThrottling` option could not be set for the RWH, which meant that various throttle mechanisms, could not be disabled (e.g. throttling a video stream being sent through WebRTC). This PR fixes this fetching therefore making the `backgroundThrottling` option having an effect.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: fixes `backgroundThrottling: false` not having an effect.

### DIFF
--- a/shell/browser/api/atom_api_web_contents.cc
+++ b/shell/browser/api/atom_api_web_contents.cc
@@ -844,9 +844,8 @@ void WebContents::BeforeUnloadFired(bool proceed,
 }
 
 void WebContents::RenderViewCreated(content::RenderViewHost* render_view_host) {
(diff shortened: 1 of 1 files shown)


deepak1556:
Is it possible to add a test for this regression ?

miniak:
@gerhardberger which Electron version are affected?

miniak:
You forgot to also fix `WebContents::SetBackgroundThrottling()`

miniak:
Is this supposed to also fix the issue with timers? #20974 still reproduces after this fix.

PR #20240, opened 2019-09-16
build: ensure we strip all binaries when zipping up dist

MarshallOfSound:
As part of the [electron/data](https://github.com/electron/data) project I noticed that earlier in the year several of our binaries went up in size including the `mksnapshot` dist and the `ffmpeg` zip file.  See this graph:

image.  The ones that spiked are the linux binaries

This PR ensures that we strip all binaries during every `dist_zip` generation.  Check CI for the difference but here is a quick proof screenshot.

**Before**
image

**After**
image

Notes: Reduced size of the `mksnapshot` binary and the secondary `ffmpeg` distribution on linux

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -338,7 +338,7 @@ step-electron-chromedriver-build: &step-electron-chromedriver-build
     name: Build chromedriver.zip
     command: |
       cd src
-      ninja -C out/Default electron:electron_chromedriver_zip
+      ninja -C out/Default electron:electron_chromedriver_zip -j $NUMBER_OF_NINJA_PROCESSES
 
 step-electron-chromedriver-store: &step-electron-chromedriver-store
   store_artifacts:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -335,8 +335,6 @@ step-electron-chromedriver-build: &step-electron-chromedriver-build
     name: Build chromedriver.zip
     command: |
       cd src
-      ninja -C out/Default chrome/test/chromedriver -j $NUMBER_OF_NINJA_PROCESSES
-      electron/script/strip-binaries.py --target-cpu="$TARGET_ARCH" --file $PWD/out/Default/chromedriver
       ninja -C out/Default electron:electron_chromedriver_zip
 
 step-electron-chromedriver-store: &step-electron-chromedriver-store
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -338,8 +338,6 @@ step-electron-chromedriver-build: &step-electron-chromedriver-build
     name: Build chromedriver.zip
     command: |
       cd src
-      ninja -C out/Default chrome/test/chromedriver -j $NUMBER_OF_NINJA_PROCESSES
-      electron/script/strip-binaries.py --target-cpu="$TARGET_ARCH" --file $PWD/out/Default/chromedriver
       ninja -C out/Default electron:electron_chromedriver_zip
```
nornagon: do we need the `-j` here?

BUILD.gn
```diff
@@ -1194,10 +1196,27 @@ template("dist_zip") {
...
+      ":$_runtime_deps_target",
+    ]
+    args = [
+      "--directory",
+      rebase_path(root_out_dir),
+      "--target-cpu",
+      target_cpu,
+    ]
+    outputs = [
+      "$target_gen_dir/$_strip_binaries_target/.strip",
```
nornagon: GN hates steps that modify files in the output, and this will probably lead to weird situations where GN doesn't properly detect that a rebuild is required. Can we do this another way?

.circleci/config.yml
MarshallOfSound: Probably, good call

BUILD.gn
MarshallOfSound: Not without changing how literally everything else works.

I set it up like this so it "outputs" a file that it never actually creates.  This way GN will always run the script (never think it's done)

MarshallOfSound: Not the best, but it works for now and doesn't require substantial changes to everything else

zcbenz:
This PR needs rebasing.

BUILD.gn
nornagon: i'm not sure where we landed on this, but if you're going to keep this hack in, at least name this file `.strip_never_created` and add a comment.

zcbenz:
@MarshallOfSound Do you still plan to work on this?

zcbenz:
I'm closing this PR since there has been no activity for a long time.

PR #20245, opened 2019-09-16
ci: use CircleCI 2.1 config and v2 APIs

jkleinsc:
#### Description of Change

This PR changes our CircleCI config to use the CircleCI 2.1 config which will allow us to use some newer CircleCI features like [conditional steps](https://circleci.com/docs/2.0/configuration-reference/#the-when-step-requires-version-21) and [orbs](https://circleci.com/docs/2.0/creating-orbs/).  This should also pave the way to use something like https://github.com/marketplace/actions/trigger-circleci-workflows to limit what CI runs when a change is made (eg docs just need lint and no build run).

Additionally moving to the 2.1 config requires changing how we trigger release builds, so this PR changes our release script to use CircleCI's v2 API which is compatible with the 2.1 config.

Please note that this PR is just doing the bare minimum to convert to 2.1.  Once this is merged in we should be able to introduce additional optimizations that take advantage of the newer config.

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/release/ci-release-build.js
+++ b/script/release/ci-release-build.js
@@ -3,6 +3,7 @@ if (!process.env.CI) require('dotenv-safe').load()
 const assert = require('assert')
 const request = require('request')
 const buildAppVeyorURL = 'https://ci.appveyor.com/api/builds'
+const circleCIPipelineURL = 'https://circleci.com/api/v2/project/gh/electron/electron/pipeline'
 const vstsURL = 'https://github.visualstudio.com/electron/_apis/build'
 
 const appVeyorJobs = {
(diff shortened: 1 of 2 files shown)


script/release/ci-release-build.js
```diff
@@ -55,30 +56,80 @@ async function makeRequest (requestOptions, parseResponse) {
...
+  } catch (err) {
+    console.log('Error calling CircleCI: ', err)
+  }
+}
+
+async function getCircleCIWorkflowId (pipelineId) {
+  const pipelineInfoUrl = `https://circleci.com/api/v2/pipeline/${pipelineId}`
+  for (let i = 0; i < 5; i++) {
+    const pipelineInfo = await circleCIRequest(pipelineInfoUrl, 'GET')
+    if (pipelineInfo.workflows.length !== 1) {
```
MarshallOfSound: The response has a status property we should check otherwise this could hang for ever

I think the possible values are `pending`, `created` and `error`.  Not 100% sure though

jkleinsc: it won't hang for ever because we only retry 5 times.  I'll add a check for status though

jkleinsc:
@MarshallOfSound in regards to sudowoodo, this PR creates workflows that only have 1 job so it will still work the way sudowoodo expects.  The `ci-release-build` script has been updated to handle the fact that we are triggering workflows with 1 job assigned to them.  The new output from this script will look like this which is compatible with sudowoodo:
```
Triggering CircleCI to run build job: linux-x64-publish on branch: circleci-2-1 with release flag.
1 jobs were requested.
CircleCI release build pipeline 73427c04-ee98-4ad3-a871-fac7fa21545e for linux-x64-publish triggered.
CircleCI release build workflow running at https://circleci.com/workflow-run/dfaf5f25-8a76-40e8-b344-94ad06f1f36e for linux-x64-publish.
CircleCI release build request for linux-x64-publish successful.  Check https://circleci.com/gh/electron/electron/331918 for status.
```

PR #20246, opened 2019-09-16
docs: remove dash that broke the return type parser

### LINKED ISSUE #20243: Property 'canceled' does not exist on type 'OpenDialogReturnValue'
I noticed that `OpenDialogReturnValue` has a property called canceled and it's even documented but it has no type definition in ` OpenDialogReturnValue` interface 

`canceled - Boolean - whether or not the dialog was canceled.`

I can define it if I know where to get the types definition files.

MarshallOfSound:
Fixes #20243

The new docs parser caught this so it is fixed on `7-0-x` and `master`.  This API only exists on `6-0-x` so we only need to fix it here.

Notes: no-notes

### DIFF
--- a/docs/api/dialog.md
+++ b/docs/api/dialog.md
@@ -117,7 +117,7 @@ dialog.showOpenDialogSync(mainWindow, {
 
 Returns `Promise<Object>` - Resolve wih an object containing the following:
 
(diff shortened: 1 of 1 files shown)


jkleinsc:
Merging as CI failure unrelated to doc change.

PR #20264, opened 2019-09-17
chore: remove native_mate (Part 5)

zcbenz:
#### Description of Change

This PR removes some uses of `mate::ObjectTemplateBuilder`, which is a major blocker to convert class methods to `gin`.

More uses will be removed in the next PR.

#### Release Notes

Notes: no-notes

### DIFF
--- a/shell/common/gin_helper/object_template_builder.h
+++ b/shell/common/gin_helper/object_template_builder.h
@@ -1,0 +1,72 @@
+// Copyright (c) 2019 GitHub, Inc. All rights reserved.
+// Use of this source code is governed by the MIT license that can be
+// found in the LICENSE file.
+
+#ifndef SHELL_COMMON_GIN_HELPER_OBJECT_TEMPLATE_BUILDER_H_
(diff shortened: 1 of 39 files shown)


shell/common/gin_helper/object_template_builder.h
```diff
@@ -0,0 +1,72 @@
...
+#include "shell/common/gin_helper/function_template.h"
+
+namespace gin_helper {
+
+// This class works like gin::ObjectTemplateBuilder, but operates on existing
+// prototype template instead of creating a new one.
+//
+// It also uses gin_helper::CreateFunctionTemplate for function templates to
+// support gin_helper types.
+class ObjectTemplateBuilder {
```
deepak1556: Maybe as a follow up we can add this constructor to upstream gin, so that class can reused without duplicating ?

codebytere:
@zcbenz this needs rebase then we should be g2g :)

shell/common/gin_helper/object_template_builder.h
zcbenz: This class also relies on our fork of `function_template.h` so simply patching gin would not be enough, I'll add comment for it.

PR #20270, opened 2019-09-18
fix: reset capturers at the very end

codebytere:
#### Description of Change

Follow-up to https://github.com/electron/electron/pull/20156. Move `capturer_.reset()` into our patch to more closely emulate [how it was done](https://github.com/electron/electron/pull/14835/files#diff-22735fa1b90e850a4ac270613eb16dcbL214-L215) prior to the desktopCapturer refactor. Fixes a linux test crash.

cc @loc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/patches/chromium/desktop_media_list.patch
+++ b/patches/chromium/desktop_media_list.patch
@@ -133,7 +133,7 @@ index 47401abc984e6fe26c7f4c5399aa565c687060b0..ca6a527ffac877c27aac94337ec5a7b5
   protected:
    virtual ~DesktopMediaListObserver() {}
 diff --git a/chrome/browser/media/webrtc/native_desktop_media_list.cc b/chrome/browser/media/webrtc/native_desktop_media_list.cc
-index b2005c70acbc1c05c59bb2059b190ab78fb63a68..1a7188e2df76672d66da3206d4448df35a065754 100644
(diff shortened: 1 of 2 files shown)


patches/chromium/desktop_media_list.patch
```diff
@@ -153,3 +153,12 @@ index b2005c70acbc1c05c59bb2059b190ab78fb63a68..1a7188e2df76672d66da3206d4448df3
  #include "media/base/video_util.h"
  #include "third_party/libyuv/include/libyuv/scale_argb.h"
  #include "third_party/skia/include/core/SkBitmap.h"
+@@ -218,6 +219,8 @@ void NativeDesktopMediaList::Worker::RefreshThumbnails(
+       FROM_HERE, {BrowserThread::UI},
+       base::BindOnce(&NativeDesktopMediaList::UpdateNativeThumbnailsFinished,
+                      media_list_));
++
++  capturer_.reset();
```
MarshallOfSound: It looks like there is a `PostTask` just above this that does work, are we sure resetting here doesn't interfere with that task?

codebytere: i don't believe so - it's the exact same flow as occurred [here](https://github.com/electron/electron/pull/14835/files#diff-22735fa1b90e850a4ac270613eb16dcbL214-L215) with the same PostTask

jkleinsc:
Merging as CI failure unrelated to fix

PR #20272, opened 2019-09-18
perf: only access remote module when <webview>.getWebContents() is called

miniak:
#### Description of Change
Fixes small performance regression introduced in #17200, where the `remote` module was imported unconditionally even though it's only used to implement `<webview>.getWebContents()`.

#### Checklist
- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/lib/renderer/web-view/web-view-impl.ts
+++ b/lib/renderer/web-view/web-view-impl.ts
@@ -1,10 +1,11 @@
-import { remote, webFrame } from 'electron'
+import * as electron from 'electron'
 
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
Sorry I missed this for initial review but this change is not required and IMO just makes the code harder to read.  The typescript transpilation target lazily accessed exported members.

See this example:

image

I don't think that was actually needed

miniak:
@MarshallOfSound I didn't realize that. Should I revert the change? Is it a good idea to rely on this?

PR #20293, opened 2019-09-20
chore: enable native unittesting

codebytere:
#### Description of Change

This PR enables native unittesting of Electron's C++/ObjC++ code.

cc @ckerr @nornagon @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


BUILD.gn
```diff
@@ -1189,6 +1198,49 @@ if (is_mac) {
...
+    ]
+
+    # Resolve paths owing to different test executable locations
+    ldflags = [
+      "-F",
+      rebase_path("external_binaries", root_build_dir),
+      "-rpath",
+      "@loader_path",
+      "-rpath",
+      "@executable_path/Electron.app/Contents/Frameworks",
```
nornagon: This is the wrong rpath to use here, because it depends on the app having been built. `Electron.app/Contents/Frameworks` contains symlinks to the "real" files, so instead, we should use the paths to the directory(s) containing real frameworks directly here.

BUILD.gn
```diff
@@ -1189,6 +1198,49 @@ if (is_mac) {
...
+    libs = [
+      "AVFoundation.framework",
+      "Carbon.framework",
+      "LocalAuthentication.framework",
+      "QuartzCore.framework",
+      "Quartz.framework",
+      "Security.framework",
+      "SecurityInterface.framework",
+      "ServiceManagement.framework",
+      "StoreKit.framework",
```
nornagon: This list of frameworks shouldn't be duplicated... the fact that they're needed here probably means they need to be moved from `electron_framework` to `electron_lib`.

script/spec-runner.js
```diff
@@ -48,7 +50,7 @@ if (args.runners) {
   }
   console.log('Only running:', runnersToRun)
 } else {
-  console.log(`Triggering both ${[...runners.keys()].join(' and ')} runners`)
+  console.log(`Triggering ${[...runners.keys()].join(' and ')} runners`)
```
nornagon:
```suggestion
  console.log(`Triggering runners: ${[...runners.keys()].join(', ')}`)
```

script/spec-runner.js
```diff
@@ -141,6 +143,70 @@ async function runRemoteBasedElectronTests () {
   console.log(`${pass} Electron remote process tests passed.`)
 }
 
+async function runNativeElectronTests () {
+  const TEST_TARGETS = require('./native-test-targets.json')
```
nornagon:
```suggestion
  const testTargets = require('./native-test-targets.json')
```

script/spec-runner.js
```diff
@@ -141,6 +143,70 @@ async function runRemoteBasedElectronTests () {
...
+    const run = childProcess.spawnSync(`./${outDir}/${args.target}`, {
+      cwd: path.resolve(__dirname, '../..'),
+      stdio: 'inherit'
+    })
+
+    if (run.status !== 0) {
+      console.log(`${fail} ${args.target} test failed.`)
+      process.exit(1)
+    }
+  } else {
```
nornagon: I think we can deduplicate this so that we share logic between the "one test" and "all tests" cases.

vsts-arm-test-steps.yml
```diff
@@ -53,6 +53,15 @@ steps:
   env:
     CIRCLE_TOKEN: $(CIRCLECI_TOKEN)
 
+- bash: |
+   export NATIVE_UNITTESTS_DEST=$PWD/src/out/Default
+   cd src/electron
+   node script/download-circleci-artifacts.js --buildNum=$CIRCLE_BUILD_NUM --name=shell_browser_ui_unittests --dest=$NATIVE_UNITTESTS_DEST
+   chmod +x $NATIVE_UNITTESTS_DEST/shell_browser_ui_unittests
```
nornagon: it's surprising to me that we have to set +x here...

codebytere: me too - for some reason in testing the version that gets pulled down isn't executable 🤔

BUILD.gn
codebytere: i'm not quite sure what you mean here? i don't see the custom (non-system, like Squirrel) frameworks existing anywhere else 🤔

nornagon: Those frameworks get assembled into the `.app` bundle as part of the `electron_app_framework_bundle_data` target, so if that target doesn't get built (e.g. if you build `shell_browser_ui_unittests` without having built `electron_app`) then those frameworks won't have been copied there.

Maybe something like:
```suggestion
      rebase_path("external_binaries", root_build_dir),
```

? That will produce a binary with an absolute-path link command, which isn't great, but the test binary isn't likely to get run from anywhere but the exact place it was built, so it's probably fine?

codebytere: hmm; that's what's already at the top in the existing PR, right? https://github.com/electron/electron/pull/20293/files#diff-4eaebbcd1dc0da11e3a3ce488290bfc6R1224

That doesn't work for the above reasons; it fails to find all local frameworks.

nornagon: `-F` and `rpath` are different flags that do different things. `-F` is where to find frameworks at build-time, `rpath` is where to find them at run-time.

PR #20307, opened 2019-09-20

MarshallOfSound:
Putting this up without tests to get API and Implementation feedback.  I've talked about this 1:1 with a few folks, both app devs and electron maintainers so I feel we've got a good handle here but as always, open to feedback 👍 Now into the fun stuff.

#### What is this?

This PR adds a new renderer-side module called `contextBridge` that allows you to expose an API synchronously from the isolated context to the main world (and optionally back again) to achieve the following goals.

1. Make it really easy for folks currently doing the `window.api = {}` pattern to enable contextIsolation.  The migration becomes `bindAPIInMainWorld('api', {})`.
2. Make it possible for synchronous APIs to exist between the main world and the isolated world.  Some app logic is heavily tied to being synchronous and rather than ask folks to refactor their entire app we can just make this possible.
3. Kind of tied to (1) this would solve the "developer experience" issues we briefly touched on at summit around enabling contextIsolation.  Once this is in, working and tested enabling `contextIsolation` by default no longer seems like a daunting task.

#### How?

The docs outline a lot of how this works from a users perspective so I'd recommend reading those.  From a Behind The Scenes perspective it is all implemented using our current converter and binding infrastructure.

Primitives / Values get converted to `base::Value` and then converted back to `v8::Value` but in the other context.  Functions get stored as `v8::Global` and then proxied into a bound cpp method that calls the original function in context A.  Then does the primitive move trick and moves it to be the return value for context B.  Their is special logic to proxy promise return values but you can see that both in the docs and the source code.

#### TODO:

* [x] Freeze the API object that gets bound to ensure it is immutable in the other world
  * [x] Document that the API object is frozen and immutable
* [x] Function arguments need to be proxied
* [x] Only expose `debugGC` in a testing build
* [x] Code clean up
* [x] CI Tests that it works
* [x] CI Tests that we don't leak prototypes across the bridge
* [x] CI Tests that sending functions isn't a memory leak
* [x] Naming? Main World vs Main Process -- How do we make that clear?
* [x] Prevent recursion destroying us.
  * [x] Keep a weakmap of objects as we traverse, use already proxied versions if we can
  * [x] Throw an error if we handle a thing with a recursion depth >= 10000?
* [x] Use the V8 serializer - Copy from @nornagon PR and resolve conflicts with whoever lands first


felixrieseberg:
1) This thing will allow us to explain context isolation in a way that won't have developers wanting to look for a different job. It'll literally be night and day. I can't express enough how big of a boon for `contextIsolation` this is.

2) I'm not a big fan of using `mainWorld` as I'm afraid that it'll get confused with the `main process`. I don't have a better solution right now but I'd like to brainstorm some alternatives to see if this is the best we can come up with.

kwonoj:
> I'm not a big fan of using mainWorld

I liked way naming comlink did, things like `expose` - something like `contextBridge.exposeFromPreload`?

As shared separately, few thoughts I have in mind are

- exposed object attached to global:

I am aware this may not be significant issue as a consumer (as conumser always can wrap any way they want to use), but default electron api advocate its behavior to attach / pollute object into global context (especially for any possible remote webcontents) would be preferred to be avoided. attaching global exposes one more way to list out what surface is being provided in preload. Personally, something like built-in module approach (https://developers.google.com/web/updates/2019/03/kv-storage) if it is available.

- streategy for un-interopable types:

it is indeed understandable some of types are not able to be transferred, but some of practices are common like `addEventListener(cb)  some(promise)` -style of passing some types of object unsupported. If it's not supported out of box by given api, there might need some workaround for those special cases handling.

docs/api/context-bridge.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Main World
+
+The "Main World" is the javascript context that your main renderer code runs in.  By default the page you load in your renderer
+executes code in this world.
+
+### Isolated World
+
+When `contextIsolation` is enabled in your `webPreferences` your `preload` scripts run in an "Isolated World".  You can read more about
+context isolation and what it affects in the [BrowserWindow](browser-window.md) docs.
```
codebytere: I think since contextIsolation is mentioned in a BW constructor option and isn't _super_ obvious, that it would be worthwhile to lower that barrier and re-copy it here or move it into a standalone document we could link to

docs/api/context-bridge.md
```diff
@@ -0,0 +1,138 @@
...
+
+### API Objects
+
+The `api` object provided to both [`bindAPIInMainWorld`](#contextbridgebindapiinmainworldapikey-api-options) and [Reverse Bindings (`window[apiKey].bind`)](#reverse-bindings) must be an object
+whose keys are strings and values are a `Function`, `String`, `Number`, `Array` or another nested object that meets the same conditions.
+
+`Function` values are proxied to the other context and all other values are **copied** and **frozen**.  I.e. Any data / primtives sent in
+the API object become immutable and updates on either side of the bridge do not result in an update on the other side.
+
+**NOTE:** Arrays containing Functions are currently not supported, Functions must not be a child of an array.
```
samiskin: How bout Arrays containing promises 🤔

docs/api/context-bridge.md
```diff
@@ -0,0 +1,138 @@
...
+| `Symbol` | ❌ | ❌ | Symbols cannot be copied across contexts so they are dropped |
+
+
+If the type you care about is not in the above table it is probably not supported.
+
+### Reverse Bindings
+
+The `contextBridge.bindAPIInMainWorld` allows you to expose APIs from the **Isolated World** the preload script runs in to the **Main World** the renderer runs in.  If you want
+to allow the main world to provide an API back to the isolated world you can set the `allowReverseBinding` option to `true`.  By setting that option the API exposed in the main
+world will have a new method injected on it on the `bind` key.  You can use this method by providing it a single `api` object in the same structure as the original `bindAPIInMainWorld`
```
samiskin: If `bind` already exists, I assume we'll just override it.  We may want to error / warn if that happens.

> You can use this method by providing it a single `api` object in the same structure as the original `bindAPIInMainWorld`

I'd use "same limitations" rather than "same structure" since structure could also be interpreted as "the same keys and values"

docs/api/context-bridge.md
```diff
@@ -0,0 +1,138 @@
...
+// Renderer (Main World)
+
+window.electron.doThing()
+```
+
+## Glossary
+
+### Main World
+
+The "Main World" is the javascript context that your main renderer code runs in.  By default the page you load in your renderer
```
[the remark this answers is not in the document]
samiskin: Are there also more details that we need to mention around different frames?  Don't fully remember the distinctions

docs/api/context-bridge.md
```diff
@@ -0,0 +1,138 @@
...
+has been included below for completeness.
+
+| Type | Parameter Support | Return Value Support | Limitations |
+| ---- | ----------------- | -------------------- | ----------- |
+| `String` | ✅ | ✅ | N/A |
+| `Number` | ✅ | ✅ | N/A |
+| `Boolean` | ✅ | ✅ | N/A |
+| `Object` | ✅ | ✅ | Keys and values must be supported in this table.  Prototype modifications are dropped.  Sending custom classes will copy values but not the prototype. |
+| `Error` | ✅ | ✅ | Errors that are thrown are also copied, this can result in the message and stack trace of the error changing slightly due to being thrown in a different context |
+| `Promise` | ✅ | ✅ | Promises are proxied through the bridge **only** if it is the return value, promises as function parameters are dropped |
```
samiskin: This limitation is gone now right

lib/renderer/api/context-bridge.ts
```diff
@@ -0,0 +1,25 @@
...
+binding._setReverseBindingStore(reverseBindings)
+
+const contextIsolationEnabled = hasSwitch('context-isolation')
+
+const checkContextIsolationEnabled = () => {
+  if (!contextIsolationEnabled) throw new Error('contextBridge API can only be used when contextIsolation is enabled')
+}
+
+const contextBridge = {
+  bindAPIInMainWorld: (key: string, api: Record<string, any>, options: { allowReverseBinding: boolean }) => {
```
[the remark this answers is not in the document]
MarshallOfSound: @samiskin Unfortunately due to technicaly reasons this type is not the one exposed to end users.  Typing this part strictly wouldn't be all that helpful.

That type would be correct though if we can get it into the type generator

docs/api/context-bridge.md
```diff
@@ -0,0 +1,138 @@
...
+* `apiKey` String - The key for the API that you bound into the main world that corresponds to the reverse binding you wish to fetch.
+
+Returns `Record<String, any> | null` the reverse bound API provided by the main world.
+
+## Usage
+
+### API Objects
+
+The `api` object provided to both [`bindAPIInMainWorld`](#contextbridgebindapiinmainworldapikey-api-options) and [Reverse Bindings (`window[apiKey].bind`)](#reverse-bindings) must be an object
+whose keys are strings and values are a `Function`, `String`, `Number`, `Array` or another nested object that meets the same conditions.
```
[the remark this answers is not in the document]
MarshallOfSound: Yup 👍

MarshallOfSound:
> Can we implement the primitives that would be necessary to build this as a 3rd-party library and experiment with the specific API independently of the Electron release cycle until we can settle on an API that's good?

To keep this answer short, no, the way this is written it requires access to v8 contexts that native modules don't have access to let alone raw JS.  Also to be fair this is already _very_ barebones, we don't prescribe API shape or structure, just "give us some functions and we'll put them in the other context".

> I have some reservations about the API as written (particularly regarding the "reverse bindings" part), and I don't want to enshrine this in Electron without giving it a bit of a test-drive first.

This API is flagged as "experimental" to make it clear the API (a) is new and relatively untested and (b) may be changed in the future.  If you have specific usability / API concerns I'd love to hear the specifics so that they can be addressed.

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,426 @@
...
+namespace {
+
+void Initialize(v8::Local<v8::Object> exports,
+                v8::Local<v8::Value> unused,
+                v8::Local<v8::Context> context,
+                void* priv) {
+  using namespace electron::api;  // NOLINT(build/namespaces)
+
+  v8::Isolate* isolate = context->GetIsolate();
+  mate::Dictionary dict(isolate, exports);
```
zcbenz: Can you use `gin` APIs in this file?

`mate::Dictionary` is replaced by `gin_helper::Dictionary` in `shell/common/gin_helper/dictionary.h`. Other converters for gin are defined in `shell/common/gin_converters`.

MarshallOfSound: @zcbenz My intention is to request a `6.1` and backport this PR.  To avoid churn I chose to use `mate` in this file to ease the backport.  I'll refactor to `gin` once it's landed 👍

zcbenz: This is reasonable, I'm good with `mate` in this file then.

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,426 @@
...
+};
+
+std::map<content::RenderFrame*, RenderFramePersistenceStore*> store_map_;
+
+RenderFramePersistenceStore* GetOrCreateStore(
+    content::RenderFrame* render_frame) {
+  if (store_map_.find(render_frame) == store_map_.end()) {
+    store_map_[render_frame] = new RenderFramePersistenceStore(render_frame);
+  }
+  return store_map_[render_frame];
```
zcbenz: Not a big issue but we can avoid duplicate lookup:

```c++
auto it = store_map_.find(render_frame);
if (it == end())
  ...
return it->first;
```

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,426 @@
...
+                                             v8::Local<v8::Value> value) {
+  // Proxy promises as they have a safe and guarunteed memory lifecycle (unlike
+  // functions)
+  if (value->IsPromise()) {
+    v8::Context::Scope destination_scope(destination);
+
+    auto v8_promise = v8::Local<v8::Promise>::Cast(value);
+    util::Promise<v8::Local<v8::Value>> promise(destination->GetIsolate());
+    v8::Local<v8::Promise> handle = promise.GetHandle();
+    UnsafePromiseHandle* p_handle = new UnsafePromiseHandle(std::move(promise));
```
zcbenz: The `util::Promise` can be allocated on heap, so `UnsafePromiseHandle` should not be needed:

```c++
auto* promise = new util::Promise<v8::Local<v8::Value>>(isolate);
auto then_cb = base::BindOnce([](util::Promise* promise) {
  promise->Resolve();
  delete promise;
}, promise);
```

docs/api/context-bridge.md
```diff
@@ -0,0 +1,127 @@
...
+### Isolated World
+
+When `contextIsolation` is enabled in your `webPreferences` your `preload` scripts run in an "Isolated World".  You can read more about
+context isolation and what it affects in the [BrowserWindow](browser-window.md) docs.
+
+## Methods
+
+The `contextBridge` module has the following instance methods:
+
+### `contextBridge.bindAPIInMainWorld(apiKey, api)`
```
miniak: what about calling it just `bindInMainWorld`?

MarshallOfSound: Updated to `exposeInMainWorld`

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+        new util::Promise<v8::Local<v8::Value>>(destination->GetIsolate());
+    v8::Local<v8::Promise> handle = promise->GetHandle();
+
+    auto then_cb = base::BindOnce(
+        [](util::Promise<v8::Local<v8::Value>>* promise, v8::Isolate* isolate,
+           v8::Global<v8::Context> source, v8::Global<v8::Context> destination,
+           RenderFramePersistenceStore* store, v8::Local<v8::Value> result) {
+          promise->Resolve(PassValueToOtherContext(
+              source.Get(isolate), destination.Get(isolate), result, store));
+          delete promise;
```
nornagon: What happens if the RenderFrame is destroyed before the promise resolves? Does this leak?

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+        source,
+        v8::Local<v8::Function>::Cast(
+            mate::ConvertToV8(destination->GetIsolate(), then_cb)),
+        v8::Local<v8::Function>::Cast(
+            mate::ConvertToV8(destination->GetIsolate(), catch_cb))));
+
+    return handle;
+  }
+
+  // Errors aren't serializable currently, we need to pull the message out and
```
nornagon: Once https://github.com/electron/electron/pull/20214 lands, errors should be serializable through the V8 value serializer.

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+  // re-construct in the destination context
+  if (value->IsNativeError()) {
+    v8::Context::Scope destination_context_scope(destination);
+    return v8::Exception::Error(
+        v8::Exception::CreateMessage(destination->GetIsolate(), value)->Get());
+  }
+
+  // Manually go through the array and pass each value individually into a new
+  // array so that deep functions get proxied or arrays of promises.
```
nornagon: I'm not sure I understand this sentence ("deep functions get proxied or arrays of promises"...?)

Could you rephrase?

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+  if (value->IsArray()) {
+    v8::Context::Scope destination_context_scope(destination);
+    v8::Local<v8::Array> arr = v8::Local<v8::Array>::Cast(value);
+    size_t length = arr->Length();
+    v8::Local<v8::Array> cloned_arr =
+        v8::Array::New(destination->GetIsolate(), length);
+    for (size_t i = 0; i < length; i++) {
+      ignore_result(cloned_arr->Set(
+          destination, static_cast<int>(i),
+          PassValueToOtherContext(source, destination,
```
nornagon: This allows the main world to crash the process through overflowing the stack by passing a deeply-nested array to a bound function.

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+  if (value->IsNull()) {
+    v8::Context::Scope destination_context_scope(destination);
+    return v8::Null(destination->GetIsolate());
+  }
+
+  base::Value ret;
+  {
+    v8::Context::Scope source_context_scope(source);
+    // TODO(MarshallOfSound): What do we do if serialization fails? Throw an
+    // error here?
```
nornagon: Throwing an error seems reasonable to me

shell/renderer/api/atom_api_context_bridge.h
```diff
@@ -0,0 +1,56 @@
...
+#include "content/public/renderer/render_frame_observer.h"
+#include "native_mate/dictionary.h"
+#include "shell/common/node_includes.h"
+
+namespace electron {
+
+namespace api {
+
+using FunctionContextPair =
+    std::tuple<v8::Global<v8::Function>, v8::Global<v8::Context>>;
```
nornagon: Perhaps define a struct for this?

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+        error_message =
+            "An unknown exception occurred in the isolated context, an error "
+            "occurred but a valid exception was not thrown.";
+      }
+    }
+  }
+
+  if (did_error) {
+    v8::Context::Scope calling_context_scope(calling_context);
+    args->ThrowError(error_message);
```
nornagon: Should we [sanitize errors](https://cs.chromium.org/chromium/src/third_party/blink/renderer/bindings/core/v8/sanitize_script_errors.h?q=sanitizescripterror&sq=package:chromium&g=0&l=10)?

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+  v8::Context::Scope target_context_scope(target_context);
+  uint32_t length = keys->Length();
+  std::string key_str;
+  int key_int;
+  for (uint32_t i = 0; i < length; i++) {
+    v8::Local<v8::Value> key = keys->Get(target_context, i).ToLocalChecked();
+    // Try get the key as a string
+    if (!mate::ConvertFromV8(api.isolate(), key, &key_str)) {
+      // Try get the key as an int
+      if (!mate::ConvertFromV8(api.isolate(), key, &key_int))
```
nornagon: When are the keys for an object integers?

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+  if (maybe_return_value.IsEmpty())
+    return v8::Undefined(args->isolate());
+
+  auto return_value = maybe_return_value.ToLocalChecked();
+
+  return PassValueToOtherContext(func_owning_context, calling_context,
+                                 return_value, store);
+}
+
+mate::Dictionary CreateProxyForAPI(mate::Dictionary api,
```
nornagon: Why take a `mate::Dictionary` here rather than a `v8::Local<v8::Object>`?

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+}
+#endif
+
+void ExposeAPIInMainWorld(const std::string& key,
+                          mate::Dictionary api,
+                          mate::Arguments* args) {
+  auto* render_frame = GetRenderFrame(api.GetHandle());
+  RenderFramePersistenceStore* store = GetOrCreateStore(render_frame);
+  auto* frame = render_frame->GetWebFrame();
+  DCHECK(frame);
```
nornagon:
```suggestion
  CHECK(frame);
```

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+  mate::Dictionary ret = mate::Dictionary::CreateEmpty(empty.isolate());
+  ret.Set("functionCount", store->functions().size());
+  return ret;
+}
+#endif
+
+void ExposeAPIInMainWorld(const std::string& key,
+                          mate::Dictionary api,
+                          mate::Arguments* args) {
+  auto* render_frame = GetRenderFrame(api.GetHandle());
```
nornagon:
```suggestion
  auto* render_frame = GetRenderFrame(api.GetHandle());
  CHECK(render_frame);
```

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+    args->ThrowError(
+        "Cannot bind an API on top of an existing property on the window "
+        "object");
+    return;
+  }
+
+  v8::Local<v8::Context> isolated_context =
+      frame->WorldScriptContext(api.isolate(), World::ISOLATED_WORLD);
+
+  {
```
nornagon: No need for the extra scope.

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+
+}  // namespace electron
+
+namespace {
+
+void Initialize(v8::Local<v8::Object> exports,
+                v8::Local<v8::Value> unused,
+                v8::Local<v8::Context> context,
+                void* priv) {
+  using namespace electron::api;  // NOLINT(build/namespaces)
```
nornagon: instead of disabling lint, just qualify the function name

default_app/preload.ts
```diff
@@ -15,6 +42,12 @@ async function initialize () {
...
+
+  for (const element of document.querySelectorAll<HTMLSpanElement>('.octicon')) {
+    loadSVG(element)
+  }
 }
 
-document.addEventListener('DOMContentLoaded', initialize)
+contextBridge.exposeInMainWorld('electronDefaultApp', {
+  initialize
+})
```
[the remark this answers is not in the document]
MarshallOfSound: Better is questionable (it removes any race but minimal benefit) I'm half using this as an example of `contextBridge`

docs/api/context-bridge.md
```diff
@@ -0,0 +1,127 @@
...
+
+* `apiKey` String - The key to inject the API onto `window` with.  The API will be accessible on `window[apiKey]`.
+* `api` Record<String, any> - Your API object, more information on what this API can be and how it works is available below.
+
+## Usage
+
+### API Objects
+
+The `api` object provided to [`exposeInMainWorld`](#contextbridgeexposeInMainWorldapikey-api-options) must be an object
+whose keys are strings and values are a `Function`, `String`, `Number`, `Array`, `Boolean` or another nested object that meets the same conditions.
```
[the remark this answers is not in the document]
MarshallOfSound: Currently no, that depends on V8 serialization afaik

docs/api/context-bridge.md
```diff
@@ -0,0 +1,127 @@
...
+
+`Function` values are proxied to the other context and all other values are **copied** and **frozen**.  I.e. Any data / primtives sent in
+the API object become immutable and updates on either side of the bridge do not result in an update on the other side.
```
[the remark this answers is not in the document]
MarshallOfSound: Proxy'ing getters and setters is infinitely more complicated and may result in Hidden Cost issues where an object gets passed around until it isn't clear it's a magical proxy object.  Then when folks use `a.b = myBigObject` it's quite slow because it gets serialized twice.

Opinionated but IMO it's better to remove that chance of the Hidden Cost.

docs/api/context-bridge.md
```diff
@@ -0,0 +1,127 @@
...
+| `Promise` | Complex | ✅ | ✅ | Promises are only proxied if they are a the return value or exact parameter.  Promises nested in arrays or obejcts will be dropped. |
+| `Function` | Complex | ✅ | ✅ | Prototype modifications are dropped.  Sending classes or constructors will not work. |
+| `Symbol` | N/A | ❌ | ❌ | Symbols cannot be copied across contexts so they are dropped |
+
+
+If the type you care about is not in the above table it is probably not supported.
+
+### Performance
+
+Electron will attempt to proxy any object, array or other structure you provide the `contextBridge` module.  This includes method parameters, method return types, etc.  This can be quite costly when sending large data blobs.  If you need more performance out of the bridge and you are sending data, no methods or dynamic content then we reccomend you `JSON.stringify` and `JSON.parse` the content across the bridge.  E.g.
```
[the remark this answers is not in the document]
MarshallOfSound: > I don't think JSON.stringify / parse will provide any performance benefit if you're using the v8 serialization algorithm.

We only use the V8 serialization algorithm for root nodes of the tree.  Anything in the middle is traversed by Electron's logic and proxied through Electron.  I benchmarked it and it's ~20% slower than using V8 serialization for everything but it's required in order to properly proxy functions and promises and such.

lib/renderer/api/context-bridge.ts
```diff
@@ -0,0 +1,20 @@
+const { hasSwitch } = process.electronBinding('command_line')
+const binding = process.electronBinding('context_bridge')
+
+const contextIsolationEnabled = hasSwitch('context-isolation')
```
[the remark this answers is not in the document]
MarshallOfSound: I think @miniak is doing work to move these options from command line flags and into webprefs but until that lands all the code in Electron has the same issue.

lib/renderer/api/context-bridge.ts
```diff
@@ -0,0 +1,20 @@
...
+const checkContextIsolationEnabled = () => {
+  if (!contextIsolationEnabled) throw new Error('contextBridge API can only be used when contextIsolation is enabled')
+}
+
+const contextBridge = {
+  exposeInMainWorld: (key: string, api: Record<string, any>) => {
+    checkContextIsolationEnabled()
+    return binding.exposeAPIInMainWorld(key, api)
+  },
+  debugGC: () => binding._debugGCMaps({})
```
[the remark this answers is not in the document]
MarshallOfSound: I'm going to use it in tests 👍

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+}  // namespace
+
+RenderFramePersistenceStore::RenderFramePersistenceStore(
+    content::RenderFrame* render_frame)
+    : content::RenderFrameObserver(render_frame) {}
+
+RenderFramePersistenceStore::~RenderFramePersistenceStore() = default;
+
+void RenderFramePersistenceStore::OnDestruct() {
+  delete this;
```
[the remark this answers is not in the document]
MarshallOfSound: > Or do the v8 objects get cleaned up before OnDestruct runs?

This is the case (at least from my own testing).  The context is destroyed before the frame is destroyed

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: My understanding is that the context is destroyed before the frame is destroyed, when the context is destroyed all V8 objects are GC'ed so this callback that is bound to V8 will be GC'ed which will then release everything else.

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: Rephrased

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: Any recommendation here?  Limit the max recursion to like 100 or something?  I think that is what mojo does

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: I can't remember the case where I need this, let me test some things and maybe remove it

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: We're just gonna insta-wrap it anyway, felt easier and means we only wrap it once

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: One of the last remnants of a copy-paste from somewhere else

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+      v8::Local<v8::Object>::Cast(value)->CreationContext();
+  if (context.IsEmpty())
+    return nullptr;
+  blink::WebLocalFrame* frame = blink::WebLocalFrame::FrameForContext(context);
+  if (!frame)
+    return nullptr;
+  return content::RenderFrame::FromWebFrame(frame);
+}
+
+std::map<content::RenderFrame*, RenderFramePersistenceStore*> store_map_;
```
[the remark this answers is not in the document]
MarshallOfSound: What's wrong with the pointers here? Haven't used `NoDestructor` in our code before

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+// Recursively freezes every v8 object on |object|.
+void DeepFreeze(const v8::Local<v8::Object>& object,
+                const v8::Local<v8::Context>& context) {
+  v8::Local<v8::Array> property_names =
+      object->GetOwnPropertyNames(context).ToLocalChecked();
+  for (uint32_t i = 0; i < property_names->Length(); ++i) {
+    v8::Local<v8::Value> child =
+        object->Get(context, property_names->Get(context, i).ToLocalChecked())
+            .ToLocalChecked();
+    if (child->IsObject())
```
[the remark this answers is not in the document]
MarshallOfSound: Arrays are objects with keys, this freezes deep inside arrays

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,408 @@
...
+      DeepFreeze(v8::Local<v8::Object>::Cast(child), context);
```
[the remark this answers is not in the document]
MarshallOfSound: Fixed 👍

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: These errors don't flow across origins right? Technically it's the same origin but a different context

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: Added a test, double checked we need this.

docs/api/context-bridge.md
```diff
@@ -0,0 +1,127 @@
...
+contextBridge.exposeInMainWorld(
+  'electron',
+  {
+    doThing: () => ipcRenderer.send('do-a-thing')
+  }
+)
+```
+
+```javascript
+// Renderer (Main World)
```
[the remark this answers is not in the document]
MarshallOfSound: Confusing -- Yes

Not sure what the alternative is though, we already deviated naming with `browser`/`main` process.  I don't want to invent a new name for worlds as well

shell/renderer/api/atom_api_context_bridge.cc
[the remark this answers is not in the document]
nornagon: I mean to use it for the map itself. We use `base::LazyInstance` elsewhere, but Chromium recommends using `NoDestructor` over `LazyInstance`: https://cs.chromium.org/chromium/src/base/no_destructor.h?q=base::nodes&sq=package:chromium&g=0&l=37

shell/renderer/api/atom_api_context_bridge.cc
nornagon: I think there are a couple of options.

1. Use a custom stack and a loop, instead of recursion. Then you have to run out of memory to crash the process, which JS can do anyway but which is a lot harder than running out of stack.
2. Detect stack overflow and throw a `RangeError: Maximum call stack size exceeded` exception. The `ValueSerializer` class in V8 does something like this (look for calls to `isolate->StackOverflow()` or `STACK_CHECK`).

shell/renderer/api/atom_api_context_bridge.cc
nornagon: The reason for the errors not to flow across origins is because you can inadvertently leak information through an error, and information isn't supposed to flow across origins unintentionally. I think the same is true of errors from the isolated world, right? I'd hesitate to say that the preload content has the same origin as the page, given the code comes from somewhere else and is controlled by something else. (I think in Chrome the isolated context is treated as having a `chrome-extension://` origin..? Maybe?)

I suppose it's not as much of a concern as cross-origin stuff, because the error was generated in the same process as the page, so if the page really wanted access to that information, [it could have it](https://spectreattack.com).

docs/api/context-bridge.md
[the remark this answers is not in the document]
nornagon: I'm not sure I understand this:

> We only use the V8 serialization algorithm for root nodes of the tree. Anything in the middle is traversed by Electron's logic and proxied through Electron.

Can you rephrase?

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: FWIW I left this running on a thing nested like a million times and it didn't crash.  Not sure how deep it has to go for it to actually overflow 🤔

docs/api/context-bridge.md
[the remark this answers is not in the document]
nornagon: > serialized twice

I'm not sure I get what would be serialized twice unexpectedly here?

I was imagining a read-only proxy layer, so that updates in the isolated world are reflected in the main world. But perhaps that's enough of an edge case that we should go with the simpler option for now :+1:

shell/renderer/api/atom_api_context_bridge.cc
nornagon: Not sure what you did but here's a fiddle which crashes the renderer: https://gist.github.com/57c3730ec8bd3f170ca0a17e2b8fbdd7

docs/api/context-bridge.md
```diff
@@ -0,0 +1,127 @@
...
+### Isolated World
+
+When `contextIsolation` is enabled in your `webPreferences` your `preload` scripts run in an "Isolated World".  You can read more about
+context isolation and what it affects in the [BrowserWindow](browser-window.md) docs.
+
+## Methods
+
+The `contextBridge` module has the following instance methods:
+
+### `contextBridge.exposeInMainWorld(apiKey, api)`
```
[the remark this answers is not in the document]
MarshallOfSound: Discussed: We're going to stick with Main World

docs/api/context-bridge.md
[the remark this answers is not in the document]
MarshallOfSound: Discussed: We're going to stick with Main World

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: I don't think error santization is what we want, partly because of the whole "they're kind of the same origin" and also because of this:

> But in the promise unhandled rejection case error existence is hidden when kSanitize is specified.

I.e. promise rejections are not supposed to be transferred when error sanitization is used.  That isn't that helpful for a bridge like this.

I think the current measure of stripping the stack trace and only transferring the message hides enough information for now.

docs/api/context-bridge.md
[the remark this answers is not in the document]
loc: I also don't think "main world" is Good. Is there record of your discussion so I might change my mind or argue further?

docs/api/context-bridge.md
```diff
@@ -0,0 +1,129 @@
+# contextBridge
+
+> Create a safe, bi-directional, synchronous bridge across isolated contexts
```
loc: Will this only ever be preload -> "main world"? This line (and the name contextBridge come to think of it) suggest that you can expose to arbitrary contexts, which, as far as I can tell, is not true.

docs/api/context-bridge.md
```diff
@@ -0,0 +1,129 @@
...
+When `contextIsolation` is enabled in your `webPreferences` your `preload` scripts run in an "Isolated World".  You can read more about
+context isolation and what it affects in the [BrowserWindow](browser-window.md) docs.
+
+## Methods
+
+The `contextBridge` module has the following instance methods:
+
+### `contextBridge.exposeInMainWorld(apiKey, api)`
+
+* `apiKey` String - The key to inject the API onto `window` with.  The API will be accessible on `window[apiKey]`.
```
loc: nit: I think we can do better than `apiKey`, which is avoidably loaded.

I'm having trouble envisioning when a consumer would want to expose multiple API objects on `window` in the main world. It seems more likely that they'll want to expose a single object under one key, almost every time. Would it make sense to include a reasonable default and make the API slightly simpler? Maybe a signature like:
```ts
contextBridge.exposeInMainWorld(
  api: { [key: string]: ContextBridgeSerializeable },
  name?: string = 'electron'
);
``` 
And your example becomes:
```ts
contextBridge.exposeInMainWorld({
  doThing: () => ipcRenderer.send('do-a-thing')
})
```

docs/api/context-bridge.md
MarshallOfSound: preload -> main world
and main world -> preload (bi-directional)

The code in theory could also bridge any worlds but currently it's only intended to support a bridge between these two.

docs/api/context-bridge.md
MarshallOfSound: I've got a few use cases in mind for exposing multiple things it helps in tests + actual apps might want to use it (including us 😄 ).  Not sure I want to standardize `window.electron`

default_app/index.ts
```diff
@@ -1,30 +0,0 @@
-async function getOcticonSvg (name: string) {
```
miniak: why move this to the preload script if the code can run just fine in the web context?

miniak: I get that we want to use the new API, but that's only strictly necessary for getting the versions. On the other hand that logic is also default-app specific... OK, this at least moves the code into a single place :)

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,473 @@
...
+                          v8::Local<v8::Object> api_object,
+                          mate::Arguments* args) {
+  auto* render_frame = GetRenderFrame(api_object);
+  CHECK(render_frame);
+  context_bridge::RenderFramePersistenceStore* store =
+      GetOrCreateStore(render_frame);
+  auto* frame = render_frame->GetWebFrame();
+  CHECK(frame);
+  v8::Local<v8::Context> main_context = frame->MainWorldScriptContext();
+  mate::Dictionary global(main_context->GetIsolate(), main_context->Global());
```
deepak1556: `gin_helper::Dictionary`

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,473 @@
...
+  if (!cached_value.IsEmpty()) {
+    return cached_value;
+  }
+
+  // Proxy functions and monitor the lifetime in the new context to release
+  // the global handle at the right time.
+  if (value->IsFunction()) {
+    auto func = v8::Local<v8::Function>::Cast(value);
+    v8::Global<v8::Function> global_func(source->GetIsolate(), func);
+    v8::Global<v8::Context> global_source(source->GetIsolate(), source);
```
deepak1556: would be better maintain consistency with naming to avoid confusion, for contexts the variable can end with `_context`

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,473 @@
...
+    size_t func_id,
+    mate::Arguments* args) {
+  // Context the proxy function was called from
+  v8::Local<v8::Context> calling_context = args->isolate()->GetCurrentContext();
+  // Context the function was created in
+  v8::Local<v8::Context> func_owning_context =
+      std::get<1>(store->functions()[func_id]).Get(args->isolate());
+
+  v8::Context::Scope func_owning_context_scope(func_owning_context);
+  v8::Local<v8::Function> func =
```
deepak1556: its better to block scope to avoid any compiler optimizations here since we clearly want this call to execute after entering a context scope.

```
v8::Context::Scope scope();
{
  // scope related operatios.
}
```

also in other places in this PR.

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,473 @@
...
+  }
+
+  // Proxy promises as they have a safe and guaranteed memory lifecycle
+  if (value->IsPromise()) {
+    v8::Context::Scope destination_scope(destination);
+
+    auto v8_promise = v8::Local<v8::Promise>::Cast(value);
+    auto* promise =
+        new util::Promise<v8::Local<v8::Value>>(destination->GetIsolate());
+    v8::Local<v8::Promise> handle = promise->GetHandle();
```
deepak1556: naming can be `source_promise` , `proxied_promise`

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,473 @@
...
+  }
+  return mate::internal::IsTrue(
+      object->SetIntegrityLevel(context, v8::IntegrityLevel::kFrozen));
+}
+
+bool IsPlainObject(const v8::Local<v8::Value>& object) {
+  if (!object->IsObject())
+    return false;
+
+  return !(object->IsNullOrUndefined() || object->IsDate() ||
```
deepak1556: Can you try upstreaming this utility to `v8.h` , manually tracking this variation everytime is not good. Maybe upstream can suggest other alternatives to test this.

native_mate/native_mate/arguments.h
```diff
@@ -31,7 +31,7 @@ class Arguments {
 
   template <typename T>
   bool GetHolder(T* out) {
-    return ConvertFromV8(isolate_, info_->Holder(), out);
+    return mate::ConvertFromV8(isolate_, info_->Holder(), out);
```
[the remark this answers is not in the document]
MarshallOfSound: It shouldn't, but it does 😢

native_mate/native_mate/dictionary.h
```diff
@@ -40,6 +40,12 @@ class Dictionary {
 
   static Dictionary CreateEmpty(v8::Isolate* isolate);
 
+  bool Has(base::StringPiece key) const {
```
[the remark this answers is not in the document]
MarshallOfSound: Outlined above when talking to @zcbenz this PR is written for easy backporting (so it uses native mate).  Once this lands I'll follow up and migrate it to `gin`

shell/common/promise_util.h
```diff
@@ -126,6 +126,16 @@ class Promise {
     return GetInner()->Reject(GetContext(), v8::Undefined(isolate()));
   }
 
+  v8::Maybe<bool> Reject(v8::Local<v8::Value> exception) {
```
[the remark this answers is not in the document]
MarshallOfSound: I actually prefer `RejectWithErrorMessage`.  It enforces the pattern that we reject promises with errors, the generic one let's people to the _wrong thing_ which isn't great IMO.  If I could get away with it I wouldn't have added this helper but we need it 😢

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,473 @@
...
+  for (uint32_t i = 0; i < property_names->Length(); ++i) {
+    v8::Local<v8::Value> child =
+        object->Get(context, property_names->Get(context, i).ToLocalChecked())
+            .ToLocalChecked();
+    if (child->IsObject() && !child->IsTypedArray()) {
+      if (!DeepFreeze(v8::Local<v8::Object>::Cast(child), context, frozen))
+        return false;
+    }
+  }
+  return mate::internal::IsTrue(
```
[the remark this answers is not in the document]
MarshallOfSound: Same comment as above RE `mate` and backporting

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: Same comment as above RE `mate` and backporting

miniak:
@MarshallOfSound considering it's a single method, wouldn't it make more sense to make it `webFrame.exposeInMainWorld()`? are you planning to add more functionality in the future?

docs/api/context-bridge.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### Parameter / Error / Return Type support
+
+Because parameters, errors and return values are **copied** when they are sent over the bridge there are only certain types that can be used.
+At a high level if the type you want to use can be serialized and un-serialized into the same object it will work.  A table of type support
+has been included below for completeness.
+
+| Type | Complexity | Parameter Support | Return Value Support | Limitations |
+| ---- | ---------- | ----------------- | -------------------- | ----------- |
+| `String` | Simple | ✅ | ✅ | N/A |
```
miniak: what about `Date`, `RegExp`, `ArrayBuffer`, `ArrayBufferView`? those are now natively supported by the IPC payload serializer. It would be nice if those were supported here as well

miniak: maybe reference https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm?

MarshallOfSound: Those are already supported, just need to put them or structured clone into this table

spec-main/api-context-bridge-spec.ts
```diff
@@ -0,0 +1,599 @@
...
+          }
+        })
+        await w.loadFile(path.resolve(fixturesPath, 'empty.html'))
+      }
+    
+      const callWithBindings = async (fn: Function) => {
+        return await w.webContents.executeJavaScript(`(${fn.toString()})(window)`)
+      }
+    
+      // const forceGC = async () => {
```
jkleinsc: Should this be removed?

docs/api/context-bridge.md
```diff
@@ -0,0 +1,129 @@
...
+| `Promise` | Complex | ✅ | ✅ | Promises are only proxied if they are a the return value or exact parameter.  Promises nested in arrays or obejcts will be dropped. |
+| `Function` | Complex | ✅ | ✅ | Prototype modifications are dropped.  Sending classes or constructors will not work. |
+| `Symbol` | N/A | ❌ | ❌ | Symbols cannot be copied across contexts so they are dropped |
+
+
+If the type you care about is not in the above table it is probably not supported.
+
+### Performance
+
+Electron will attempt to proxy any object, array or other structure you provide the `contextBridge` module.  This includes method parameters, method return types, etc.  This can be quite costly when sending large data blobs.  If you need more performance out of the bridge and you are sending data, no methods or dynamic content then we recommend you `JSON.stringify` and `JSON.parse` the content across the bridge.  E.g.
```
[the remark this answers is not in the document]
MarshallOfSound: Nope

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: Yeah can do 👍

shell/common/promise_util.h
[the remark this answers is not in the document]
deepak1556: The `promise_util` module also has a `Reject(void)` util, so I don't think it enforces this rule in a hard way, but nevertheless doesn't hurt to have a utility.

docs/api/context-bridge.md
```diff
@@ -0,0 +1,127 @@
+# contextBridge
+
+> Create a safe, bi-directional, synchronous bridge across isolated contexts
+
+Process: [Renderer](../glossary.md#renderer-process)
+
```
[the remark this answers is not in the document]
nornagon: This didn't get addressed...?

nornagon:
I saw a comment here about moving this to `webFrame`, but I've lost it in GitHub's collapsed comments UI. Did that get addressed?

shell/renderer/api/context_bridge/render_frame_context_bridge_store.h
```diff
@@ -0,0 +1,84 @@
...
+namespace api {
+
+namespace context_bridge {
+
+using FunctionContextPair =
+    std::tuple<v8::Global<v8::Function>, v8::Global<v8::Context>>;
+
+using WeakGlobalPair = std::tuple<v8::Global<v8::Value>, v8::Global<v8::Value>>;
+
+struct WeakGlobalPairNode {
```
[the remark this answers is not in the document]
MarshallOfSound: Try this, copy / move semantics did not play nicely together.  This code is clean enough so gonna stick with it for now.

spec-main/api-context-bridge-spec.ts
```diff
@@ -9,7 +9,7 @@ import { emittedOnce } from './events-helpers'
 
 const fixturesPath = path.resolve(__dirname, 'fixtures', 'api', 'context-bridge')
 
-describe('contextBridge', () => {
+describe.only('contextBridge', () => {
```
deepak1556: accidental `only` change ?

MarshallOfSound: Whoops

MarshallOfSound:
> Did that get addressed?

The comment was removed.

 I was going to reply with something along the lines of “webFrames are traversable and have an 1:N relationship with a preload script”.  We only want this method to be calling logically from the preload script targetting the main world so currently it makes no sense to expose on webFrame”

In particular note that subframes may be out-of-process and therefore this API would not function on them

docs/api/context-bridge.md
```diff
@@ -0,0 +1,111 @@
...
+executes code in this world.
+
+### Isolated World
+
+When `contextIsolation` is enabled in your `webPreferences` your `preload` scripts run in an "Isolated World".  You can read more about
+context isolation and what it affects in the [BrowserWindow](browser-window.md) docs.
+
+## Methods
+
+The `contextBridge` module has the following instance methods:
```
nornagon: These aren't really "instance methods" because there's nothing for `contextBridge` to be an instance of

docs/api/context-bridge.md
nornagon: We should update this to point to SCA probably.

lib/renderer/api/context-bridge.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+if (!binding._debugGCMaps) delete contextBridge.debugGC
```
nornagon: nit: can we invert this?

```js
if (binding._debugGCMaps)
  contextBridge.debugGC = () => binding._debugGCMaps({})
```

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,507 @@
...
+  if (context.IsEmpty())
+    return nullptr;
+  blink::WebLocalFrame* frame = blink::WebLocalFrame::FrameForContext(context);
+  if (!frame)
+    return nullptr;
+  return content::RenderFrame::FromWebFrame(frame);
+}
+
+std::map<content::RenderFrame*, context_bridge::RenderFramePersistenceStore*>
+    store_map_;
```
nornagon: This should be `base::LazyInstance` or `base::NoDestructor`

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,507 @@
...
+  }
+  return mate::internal::IsTrue(
+      object->SetIntegrityLevel(context, v8::IntegrityLevel::kFrozen));
+}
+
+bool IsPlainObject(const v8::Local<v8::Value>& object) {
+  if (!object->IsObject())
+    return false;
+
+  return !(object->IsNullOrUndefined() || object->IsDate() ||
```
nornagon: nit:
```suggestion
  return object->IsObject() && !(object->IsNullOrUndefined() || object->IsDate() ||
```

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,507 @@
...
+           object->IsSharedArrayBuffer() || object->IsProxy() ||
+           object->IsWebAssemblyCompiledModule() ||
+           object->IsModuleNamespaceObject());
+}
+
+bool IsPlainArray(const v8::Local<v8::Value>& arr) {
+  if (!arr->IsArray())
+    return false;
+
+  return !arr->IsTypedArray();
```
nornagon:
```suggestion
  return arr->IsArray() && !arr->IsTypedArray();
```

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,507 @@
...
+    context_bridge::RenderFramePersistenceStore* store,
+    int recursion_depth) {
+  if (recursion_depth >= kMaxRecursion) {
+    v8::Context::Scope source_scope(source_context);
+    {
+      source_context->GetIsolate()->ThrowException(v8::Exception::TypeError(
+          mate::StringToV8(source_context->GetIsolate(),
+                           "Electron contextBridge recursion depth exceeded.  "
+                           "Nested objects "
+                           "deeper than 10000 are not supported.")));
```
nornagon: `kMaxRecursion` is 1000, but this error message states 10x that

shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,507 @@
...
+  context_bridge::RenderFramePersistenceStore* store =
+      GetOrCreateStore(render_frame);
+  auto* frame = render_frame->GetWebFrame();
+  CHECK(frame);
+  v8::Local<v8::Context> main_context = frame->MainWorldScriptContext();
+  mate::Dictionary global(main_context->GetIsolate(), main_context->Global());
+
+  if (global.Has(key)) {
+    args->ThrowError(
+        "Cannot bind an API on top of an existing property on the window "
```
nornagon: Is there a technical reason not to do this or is it just sort of a bad idea? If a user wants to do this, they can easily do it:

```js
contextBridge.exposeInMainWorld('_myFakeLocation', {...})
webFrame.executeJavaScript('window.location = _myFakeLocation')
```

so it seems a bit weird to throw an error here.

shell/renderer/api/context_bridge/render_frame_context_bridge_store.cc
```diff
@@ -0,0 +1,89 @@
...
+    auto global_proxy =
+        v8::Global<v8::Value>(v8::Isolate::GetCurrent(), proxy_value);
+    // Do not retain
+    global_from.SetWeak();
+    global_proxy.SetWeak();
+    auto iter = proxy_map_.find(hash);
+    auto* node = new WeakGlobalPairNode(
+        std::make_tuple(std::move(global_from), std::move(global_proxy)));
+    if (iter == proxy_map_.end()) {
+      proxy_map_.emplace(hash, node);
```
nornagon: How do objects get removed from this map?

MarshallOfSound: They don't, they're all weak references so the underlying object is not retained

lib/renderer/api/context-bridge.ts
MarshallOfSound: It's done in the current way so that typescript understands what is going on without `any` casting

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: I swear I did that, must have pushed over it.  Will do

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: I deliberately kept these as two separate checks to make it cleared to the reader that we first care if it an object at all and then we care what type of object.

The compile will optimize this down for us.

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: Same as above, this is done for readability and consistency

shell/renderer/api/atom_api_context_bridge.cc
MarshallOfSound: > Is there a technical reason not to do this or is it just sort of a bad idea?

Overriding some globals will throw errors and also it's kinda a bad idea.  Stopping users shooting themselves in the foot or at least making it harder was important here.

shell/renderer/api/context_bridge/render_frame_context_bridge_store.cc
nornagon: So this map will get bigger every time a contextBridge function is called? that doesn't seem good.

shell/renderer/api/atom_api_context_bridge.cc
nornagon: Counterpoint: if a user exposes an API at a key that isn't currently present on `window`, but later becomes present due to a new Web API, their app will break for no good reason. I get the idea of trying to help users not shoot themselves in the foot, i'm just not sure the juice is worth the squeeze here.

MarshallOfSound: Further counter point, wouldn't you want to know if you were suddenly overwriting a new web API that any one of your thousands of dependencies could be using?

shell/renderer/api/context_bridge/render_frame_context_bridge_store.cc
MarshallOfSound: I made a new object life monitor to track the keys and clean up as things get GC'ed.  Memory impact was small before but this is self-cleaning now 👍

shell/renderer/api/context_bridge/render_frame_context_bridge_store.h
```diff
@@ -0,0 +1,71 @@
...
+#include "third_party/blink/public/web/web_local_frame.h"
+
+namespace electron {
+
+namespace api {
+
+namespace context_bridge {
+
+using FunctionContextPair =
+    std::tuple<v8::Global<v8::Function>, v8::Global<v8::Context>>;
```
nornagon: i thought i commented on it before but it seems to have gotten lost; can we use a struct for this instead of `std::tuple`?
```suggestion
struct FunctionContextPair {
  v8::Global<v8::Function> function;
  v8::Global<v8::Context> context;
};
```

shell/renderer/api/context_bridge/render_frame_context_bridge_store.h
```diff
@@ -0,0 +1,71 @@
...
+
+using WeakGlobalPair = std::tuple<v8::Global<v8::Value>, v8::Global<v8::Value>>;
```
nornagon: here too
```suggestion
struct WeakGlobalPair {
  v8::Global<v8::Value> value_in_isolated_context;
  v8::Global<v8::Value> value_in_main_context;
};
```

reZach:
@MarshallOfSound, since you wrote this code, can you think of a case where the values/objects passed through the contextBridge do not fully "come out" in the other context? I'll try to parse through your commits too when I have time.

ipcRenderer does not have access to the **.on(channel, args)** method after coming through the contextBridge.
**preload.js**
```
const {
    contextBridge,
    ipcRenderer
} = require("electron");

contextBridge.exposeInMainWorld(
    "electron", {
        ipcRenderer: ipcRenderer
    }
);
``` 

**index.html**
```
<!doctype html>
<html lang="en-US">
  <head>
    <meta charset="UTF-8">
    <title>Getting Started!</title>
  </head>
  <body>
    <script>
        console.log("In index.html");
        try {
          window.electron.ipcRenderer.on("event2", (a, b) => {});
        } catch (err){
          // log error
          console.error(err); // logs error; ".on" is not a method
        }
    </script>
  </body>
</html>
```

MarshallOfSound:
@reZach This is all in the documentation, prototypes are not sent over the bridge.

https://electronjs.org/docs/api/context-bridge#parameter--error--return-type-support

Also as a side note, exposing the entire `ipcRnederer` module kind of defeats the purpose of context isolation.  You should expose the **minimal** required APIs, not wildcard style APIs

reZach:
Thank you for clarifying that @MarshallOfSound. I missed that bit!

There is no other way to configure the `ipcRenderer`'s listeners that are based on logic in the renderer process without setting **contextIsolation=false**. I could copy a minimal object via `contextBridge`, but what I tested, the entire set of ipcRenderer properties of the existing instance must be sent along or else the ipcRenderer doesn't work in the renderer process.

The entire `ipcRenderer` instance is ideal to have in the renderer process because electron libraries might need to configure listeners based on library data itself. Ie. a call _couldn't_ exist like this in the preload script because the preload script doesn't have access to this data.

## Desired
**renderer**
```
class LibraryClass {
    constructor(myValue = 1) {
        this.classValue = myValue;

        window.electron.ipcRenderer.on("response", (IpcRendererEvent, args) => {
            if (args.success) this.classValue++;
        });
    }

    send() {
        window.electron.ipcRenderer.send("request", {
            data: this.classValue
        });
    }
}
```

**main**
```
let win = new BrowserWindow({...});

ipcMain.on("request", (IpcMainEvent, args) => {
    // Use node module (ie. "fs");
    // perhaps save value of args.data to file with a timestamp

    win.webContents.send("response", {
        success: true
    });
});
```

I've posted a [$230 bounty](https://www.bountysource.com/issues/85135915-ipcrenderer-does-not-execute-normally-in-preload-js) for this issue. I imagine as library authors switch to use `contextIsolation`, they will run into this same problem too.

## Workaround
For the time being I'm setting `contextIsolation` to false, and my preload script looks something like this:
```
const { ipcRenderer } = require("electron");

window.electron= {
    ipcRenderer
};
```

PR #20324, opened 2019-09-23
refactor: convert some more remote stuff to typescript

nornagon:
#### Description of Change
This converts the browser side of the `remote` module to TypeScript.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


lib/browser/remote/objects-registry.ts
```diff
@@ -1,15 +1,18 @@
...
+
 const v8Util = process.electronBinding('v8_util')
 
-const getOwnerKey = (webContents, contextId) => {
+const getOwnerKey = (webContents: WebContents, contextId: string) => {
   return `${webContents.id}-${contextId}`
 }
 
 class ObjectsRegistry {
+  nextId: number = 0
```
MarshallOfSound: This can/should be private?

```suggestion
  private nextId: number = 0
```

lib/browser/remote/objects-registry.ts
```diff
@@ -1,15 +1,18 @@
...
+  storage: Record<number, { count: number, object: any }>
```
MarshallOfSound:
```suggestion
  private storage: Record<number, { count: number, object: any }> = {}
```

lib/browser/remote/objects-registry.ts
```diff
@@ -1,15 +1,18 @@
...
+  owners: Record<string, Map<number, number>>
```
MarshallOfSound:
```suggestion
  private owners: Record<string, Map<number, number>> = {}
```

lib/browser/remote/objects-registry.ts
```diff
@@ -1,15 +1,18 @@
...
   constructor () {
```
MarshallOfSound: We can completely remove the constructor here

lib/browser/remote/objects-registry.ts
```diff
@@ -117,18 +120,18 @@ class ObjectsRegistry {
   }
 
   // Private: Clear the storage when renderer process is destroyed.
-  registerDeleteListener (webContents, contextId) {
+  registerDeleteListener (webContents: WebContents, contextId: string) {
     // contextId => ${processHostId}-${contextCount}
     const processHostId = contextId.split('-')[0]
-    const listener = (event, deletedProcessHostId) => {
+    const listener = (event: any, deletedProcessHostId: string) => {
```
MarshallOfSound: If `event` is unused it should be `_` and `unknown` or left undefiend.
```suggestion
    const listener = (_, deletedProcessHostId: string) => {
```

lib/browser/remote/server.ts
```diff
@@ -234,29 +333,29 @@ const unwrapArgs = function (sender, frameId, contextId, args) {
...
+        throw new TypeError(`Unknown type: ${(meta as any).type}`)
     }
   }
   return args.map(metaToValue)
 }
 
-const isRemoteModuleEnabledImpl = function (contents) {
-  const webPreferences = contents.getLastWebPreferences() || {}
+const isRemoteModuleEnabledImpl = function (contents: electron.WebContents) {
+  const webPreferences = (contents as any).getLastWebPreferences() || {}
```
miniak: Why do you need to cast to `any` here?

nornagon: bc `getLastWebPreferences` isn't in the typings. i guess i could put it there? but i'm lazy :D

PR #20335, opened 2019-09-24
feat: add --enable-api-filtering-logging command-line switch

miniak:
#### Description of Change
Add `--enable-api-filtering-logging` command-line switch, which enables caller stack logging for the following APIs (filtering events):
- `desktopCapturer.getSources()` / `desktop-capturer-get-sources`
- `remote.require()` / `remote-require`
- `remote.getGlobal()` / `remote-get-builtin`
- `remote.getBuiltin()` / `remote-get-global`
- `remote.getCurrentWindow()` / `remote-get-current-window`
- `remote.getCurrentWebContents()` / `remote-get-current-web-contents`
- `remote.getGuestWebContents()` / `remote-get-guest-web-contents`

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `--enable-api-filtering-logging` command-line switch, which enables caller stack logging for `desktopCapturer` and `remote` APIs that can be blocked or filtered.


MarshallOfSound:
I feel as though there is a cleaner and less impactful implementation out there for this.

What if we captures the trace inside ipcRenderer.send and attached it to the `event` object itself. I.e. `event.senderStackTrace`?

miniak:
@MarshallOfSound that seems quite intrusive as well

miniak:
@nornagon when blocking the remote module, sometimes it's hard to figure out where it's being used from. especially when the renderer process is sandboxed, the preload script cannot be debugged properly to see where the exception is thrown.

zcbenz:
From past experience it is generally bad idea to change API's behavior with environment variables.

miniak:
@zcbenz I agree with that. My concern is the performance impact of always capturing the stack traces if it's not optional.

miniak:
@zcbenz what about not changing the API, but just logging the call stacks when the variable is set?

lib/renderer/api/remote.js
```diff
@@ -281,6 +282,16 @@ function handleMessage (channel, handler) {
   })
 }
 
+const enableStacks = hasSwitch('enable-api-filtering-logging')
+
+function getCurrentStack () {
+  const target = {}
+  if (enableStacks) {
+    Error.captureStackTrace(target)
```
nornagon: TIL `Error.captureStackTrace`. What a weird API! If I'm reading https://v8.dev/docs/stack-trace-api right, it looks like we could make this stack a little cleaner by passing `getCurrentStack` as the 2nd argument to `captureStackTrace`, which would exclude that function from the trace?

miniak: @nornagon that would be great. I will check it out

miniak: @nornagon I've actually passed the `remote` method instead of `getCurrentStack` as the 2nd argument, as the log message already says which remote method was called.

lib/renderer/api/desktop-capturer.ts
```diff
@@ -1,12 +1,24 @@
...
+
 // |options.types| can't be empty and must be an array
 function isValid (options: Electron.SourcesOptions) {
   const types = options ? options.types : undefined
   return Array.isArray(types)
 }
 
+const enableStacks = hasSwitch('enable-api-filtering-logging')
+
+function getCurrentStack (caller: Function) {
```
nornagon: I'm not sure that `caller` is the right thing to pass here, as it means the stack trace doesn't include the name of the function of interest. e.g.

```
> function withCaller(caller) { const x = {}; Error.captureStackTrace(x, caller); return x.stack }
> function foo() { console.log(withCaller(foo)) }
> function withSelf() { const x = {}; Error.captureStackTrace(x, withSelf); return x.stack }
> function bar() { console.log(withSelf()) }
> foo()
Error
    at repl:1:1 
    at Script.runInThisContext (vm.js:123:20)
    at REPLServer.defaultEval (repl.js:332:29)
    at bound (domain.js:393:14)
    at REPLServer.runBound [as eval] (domain.js:406:12)
    at REPLServer.onLine (repl.js:642:10)
    at REPLServer.emit (events.js:202:15)
    at REPLServer.EventEmitter.emit (domain.js:439:20)
    at REPLServer.Interface._onLine (readline.js:308:10)
    at REPLServer.Interface._line (readline.js:654:8)
undefined
> bar()
Error
    at bar (repl:1:30)
    at repl:1:1
    at Script.runInThisContext (vm.js:123:20)
    at REPLServer.defaultEval (repl.js:332:29)
    at bound (domain.js:393:14)
    at REPLServer.runBound [as eval] (domain.js:406:12)
    at REPLServer.onLine (repl.js:642:10)
    at REPLServer.emit (events.js:202:15)
    at REPLServer.EventEmitter.emit (domain.js:439:20)
    at REPLServer.Interface._onLine (readline.js:308:10)
undefined
```

calling `bar()` lists `bar` at the top of the stack trace, but `foo()` doesn't.

miniak: it shows
```
WebContents (1): remote.require('fs') Error
    at foobar (/private/var/folders/8r/wrg27t2s7dj035jmgf4kwqvm0000gn/T/tmp-46573iyoYocpIXfBO/renderer.js:5:8)
    at Object.<anonymous> (/private/var/folders/8r/wrg27t2s7dj035jmgf4kwqvm0000gn/T/tmp-46573iyoYocpIXfBO/renderer.js:8:1)
    at Object.<anonymous> (/private/var/folders/8r/wrg27t2s7dj035jmgf4kwqvm0000gn/T/tmp-46573iyoYocpIXfBO/renderer.js:10:3)
    at Module._compile (internal/modules/cjs/loader.js:948:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:960:10)
    at Module.load (internal/modules/cjs/loader.js:794:32)
    at Module._load (internal/modules/cjs/loader.js:707:12)
    at Module._load (electron/js2c/asar.js:732:26)
    at Function.Module._load (electron/js2c/asar.js:732:26)
    at Object.<anonymous> (electron/js2c/renderer_init.js:2382:16)
```

miniak: otherwise it would be
```
WebContents (1): remote.require('fs') Error
    at Object../lib/renderer/api/remote.js.exports.require (electron/js2c/renderer_init.js:1535:73)
    at foobar (/private/var/folders/8r/wrg27t2s7dj035jmgf4kwqvm0000gn/T/tmp-46573pedsayHGW7Yr/renderer.js:5:8)
    at Object.<anonymous> (/private/var/folders/8r/wrg27t2s7dj035jmgf4kwqvm0000gn/T/tmp-46573pedsayHGW7Yr/renderer.js:8:1)
    at Object.<anonymous> (/private/var/folders/8r/wrg27t2s7dj035jmgf4kwqvm0000gn/T/tmp-46573pedsayHGW7Yr/renderer.js:10:3)
    at Module._compile (internal/modules/cjs/loader.js:948:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:960:10)
    at Module.load (internal/modules/cjs/loader.js:794:32)
    at Module._load (internal/modules/cjs/loader.js:707:12)
    at Module._load (electron/js2c/asar.js:715:26)
    at Function.Module._load (electron/js2c/asar.js:715:26)
```

miniak: maybe it's better like even even though the function name is both in the log message and the stack trace

miniak:
@nornagon modified `getCurrentStack` as suggested, should be ready to land

PR #20338, opened 2019-09-25
docs: Minor fixes to build instructions

ajmacd:
#### Description of Change
1. Checkout `master` before attempting to set its upstream.
2. Direct readers to the unit testing page, rather than having them manually perform some of the tasks `spec-runner.js` handles.

#### Checklist

#### Release Notes
notes: no-notes

### DIFF
--- a/docs/development/build-instructions-gn.md
+++ b/docs/development/build-instructions-gn.md
@@ -72,6 +72,7 @@ origin URLs.
 $ cd src/electron
 $ git remote remove origin
 $ git remote add origin https://github.com/electron/electron
+$ git checkout master
 $ git branch --set-upstream-to=origin/master
 $ cd -
 ```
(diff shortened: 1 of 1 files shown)


ajmacd:
Thanks for taking a look. I was deluding myself into thinking providing both was solving the issue, but I really do think that the relative path is incorrect. Check out the current working directory when the error is reported:
```
gyp: /Users/ajm/electron-gn/src/electron/spec/fixtures/out/Debug/gen/node_headers/common.gypi not found (cwd: /Users/ajm/electron-gn/src/electron/spec/fixtures/native-addon/echo) while reading includes of binding.gyp while trying to load binding.gyp
```

I think the right way to express this is with an absolute path:
```
$ npm i --nodedir=$PWD/../../out/Debug/gen/node_headers
```

which is the same thing the `spec-runner.js` script does:
https://github.com/electron/electron/blob/80af35e0cd391f1a154c58a3552eae73be708312/script/spec-runner.js#L218

But, speaking of which, and now that I understand a little more what I'm doing, I think we should just direct users to run that script. I've done that in the latest commit. WDYT?

PR #20342, opened 2019-09-25
fix: allow paths to asar archives to contain the .asar extension in directories

miniak:
#### Description of Change
There is an issue with the code parsing paths to determine whether it's an ASAR archive:
https://github.com/electron/electron/blob/0653e31767a4e5aa1220042a7a8a838d045ad561/lib/common/asar.js#L43-L73

It can handle `//some/path.asar/to/archive.asar/then/internal.file`, however if the file ends up being mapped to `//some/path.asar/to/archive.asar.unpacked/then/internal.file`, it incorrectly returns:
```
{
  isAsar: true,
  asarPath: "//some/path.asar",
  filePath: "to/archive.asar.unpacked/then/internal.file"
}
```
This PR:
- moves the core of the `splitPath` implementation to native code reusing `asar::GetAsarArchivePath()` to eliminate code duplication
- no longer considers directories with `.asar` extension an archive (the results of is directory check need to be cached for performance reasons)

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Fixed parsing of paths with `.asar` in directory name extensions.


miniak:
@zcbenz we have an issue with a customer, who has the app installed to 
`C:\Users\t.asar\AppData\Local\Microsoft\Teams`

Loading a node module, which is located in app.asar.unpacked fails with this exception:
```
Uncaught Exception:
Error: Invalid package \\?\C:\Users\t.asar
    at invalidArchiveError (ELECTRON_ASAR.js:147:19)
    at process.module.(anonymous function) [as dlopen] (ELECTRON_ASAR.js:171:9)
    at Object.Module._extensions..node (internal/modules/cjs/loader.js:740:18)
    at Object.module.(anonymous function) [as .node] (ELECTRON_ASAR.js:180:18)
    at Module.load (internal/modules/cjs/loader.js:620:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:559:12)
    at Function.Module._load (internal/modules/cjs/loader.js:551:3)
    at Module.require (internal/modules/cjs/loader.js:658:17)
    at require (C:\Users\t.asar\AppData\Local\Microsoft\Teams\current\resources\app.asar\external\v8-compile-cache\v8-compile-cache.js:173:28)
    at Object.<anonymous> (C:\Users\t.asar\AppData\Local\Microsoft\Teams\current\resources\app.asar\node_modules\native-utils\index.js:1:173)
```

miniak:
@zcbenz I am not sure what's the best solution. Ideally only files would be treated as .asar archives, not directories. But we would need to check the type every time we parse the patch, which would be inefficient.

zcbenz:
@miniak I think it is not just a problem with reading file under `archive.asar.unpacked`, it seems that reading any normal file under `C:\Users\xyz.asar\` would result in the same error?

The `//some/path.asar/to/archive.asar/then/internal.file` works because the path is parsed backwards, paths like `//some/path.asar/to/file` would fail if `path.asar` is a directory.

> But we would need to check the type every time we parse the path, which would be inefficient.

Checking type every time seems to be only correct solution. We can probably use caching to avoid performance problem:

```js
if (cachedArchives.has(archivePath))
  isAsar = true
else if (cacedFolders.has(archivePath))
  isAsar = false
else {
  if (fs.isFolder(archivePath)) {
    cacedFolders.add(archivePath)
    isAsar = false
  } else
    isAsar = true
}
```

we would have problems when user replaces a folder with a real asar archive though, but I think we don't need to be prepared for that.

miniak:
@zcbenz I've added the check for directories with caching. I've also moved some of parsing to native to eliminate code duplication.

PR #20354, opened 2019-09-27
fix: when building with enable_plugins=false

miniak:
#### Description of Change
Fix building with `enable_plugins=false`

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes


patches/chromium/fix_when_building_with_enable_basic_printing_false.patch
```diff
@@ -0,0 +1,31 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Milan Burda <milan.burda@gmail.com>
+Date: Sat, 28 Sep 2019 10:39:58 +0200
+Subject: fix: when building with enable_basic_printing=false
+
+
```
nornagon: Patches must include justification. Also, it seems like this patch should join the other printing mega-patch?

patches/chromium/fix_when_building_with_enable_plugins_false.patch
```diff
@@ -0,0 +1,38 @@
...
+ namespace content {
+ 
+ namespace {
+@@ -152,7 +156,9 @@ void SetupPPAPISandboxParameters(sandbox::SeatbeltExecClient* client) {
+   SetupCommonSandboxParameters(client);
+ 
+   std::vector<content::WebPluginInfo> plugins;
++#if BUILDFLAG(ENABLE_PLUGINS)
+   PluginService::GetInstance()->GetInternalPlugins(&plugins);
++#endif
```
nornagon: Should this be upstreamed?

patches/chromium/fix_when_building_with_enable_basic_printing_false.patch
miniak: merged with `printing.patch`

patches/chromium/fix_when_building_with_enable_plugins_false.patch
miniak: @deepak1556 can you pleas help with this?

miniak:
@nornagon do you want us to upstream the Chromium patch before merging this PR?

deepak1556:
Patch uploaded at https://chromium-review.googlesource.com/c/chromium/src/+/1845624

deepak1556:
Have updated the patches from upstream. Should be good to go once build is finished.

Issue #20233, opened 2019-09-14
High CPU on Windows upon crash when not starting crash reporter


bpasero:
This is a regression from 4.x as far as I can tell.

Steps:
* `git clone https://github.com/bpasero/electron-crashreporter`
* `yarn`
* open `constants.js` and configure:
  * `enableCrashReporter: false`
  * `crashMain` (or renderer, or child process): `true`
  * `tmpDir`: some location that exists
* `yarn start`

=> 🐛 the process that crashes goes to 100% CPU and does not terminate. My feeling is that this is a regression from adopting crashpad for Windows.

This is a blocker for shipping E6 for VSCode.

//cc @deepak1556

bpasero:
This seems to be happening on Linux as well but with much lower CPU pressure:

![image](https://user-images.githubusercontent.com/900690/64905258-df0c1580-d6d5-11e9-898d-1aa30568bf0a.png)

bpasero:
@sofianguy @MarshallOfSound ping, any updates? I am getting a bit worried that we have to pull out the E6 update for VSCode next week when we close for a monthly stable release.

zcbenz:
@bpasero I could not reproduce it on either Linux or Windows 8.1, when crashing each process.

Can you check if you have any `ELECTRON_` environment variable set that may be related?

And can you use Process Explorer to capture the stack of the process that takes 100% CPU?

<img width="1133" alt="Screen Shot 2019-09-30 at 10 51 18" src="https://user-images.githubusercontent.com/639601/65844291-965c8980-e370-11e9-9584-06320b3c071b.png">

bpasero:
@zcbenz thanks for looking into. I am not really able anymore to reproduce this on Linux for some reason, so let's focus on Windows for now. 

There I can reproduce it from my desktop that runs Windows natively (no VM) with version: 1903 (18362.295).

The stack is:

```
0x0000000000000000
ntdll.dll!NtQueryVirtualMemory+0x14
ntdll.dll!RtlProtectHeap+0x141
ntdll.dll!RtlProtectHeap+0x95
ntdll.dll!RtlIsGenericTableEmpty+0xd8
ntdll.dll!RtlRaiseException+0x1e6
ntdll.dll!KiUserExceptionDispatcher+0x2e
electron.exe!v8::HeapStatistics::number_of_detached_contexts+0x34a0
electron.exe!uv_loadavg+0xa0e66
electron.exe!uv_loadavg+0xa0e14
[...]

PR #20388, opened 2019-10-01
fix: correctly crash when there is no crashReporter

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/20233.

When there is exception and we do not have crashReporter set up, we should let the execution continue and crash, which is the default behavior of Windows programs.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix process taking 100% CPU when crashed with no crashReporter set up on Windows x64.


jkleinsc:
Merging as Windows 32 bit failure is known flake unrelated to this PR.

PR #20390, opened 2019-10-01
fix: remove unnecessary patch

CezaryKulakowski:
Patch which this pr removes adds new version of
BrowserCompositorMac::GetCompositor. There is this function in
BrowserCompositorMac with specifier const. Chromium's function
returns parent's compositor if parent is set. Electron's version
doesn't and - as a result - constrained dialogs won't be displayed
correctly if they don't have its own compositor.

Notes: removed unnecessary chromium patch for browser compositor on mac


CezaryKulakowski:
Patch which this pr removes was created by Cheng Zhao so he should know if removing it now is safe.

nornagon:
@reviewers, please review release notes before merging a PR! Our checklist says:
- [ ] PR release notes describe the change in a way relevant to app developers, and are capitalized, punctuated, and past tense.

The notes in this PR do not meet those criteria.

PR #20397, opened 2019-10-01
chore: update patches for chromium roll

loc:
### setAlwaysOnTop
Fixes the setAlwaysOnTop tests for the macOS build in the in-progress Chrome roll. Also cleans up the code, ever-so-slightly.

Chrome changed the API around SetZOrderLevel/GetZOrderLevel in commit [7b02e28175d17da363ade2940b86001a48a44cf4](https://chromium.googlesource.com/chromium/src/+/7b02e28175d17da363ade2940b86001a48a44cf4%5E%21/#F4). Instead of reading the window level from the NSWindow, they now store the level in a local variable. Previously, we didn't bother calling set, since it has a more limited range of the levels it accepts. With this change, we call it anyway, in order to ensure correctness at the lower level (though I'm not sure if it's actually used).

I plan to improve this API in the future; it's kinda janky.

### DesktopMediaList
Conveniently, Chrome [just added a mode for using the DesktopMediaList in a very similar way](https://chromium-review.googlesource.com/c/chromium/src/+/1669775) to how we use it for desktopCapturer. This presented the opportunity to vastly simplify the patch.

notes: no-notes


shell/browser/native_window_mac.mm
```diff
@@ -875,52 +875,62 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
...
-  if (newLevel >= minWindowLevel && newLevel <= maxWindowLevel) {
-    was_maximizable_ = IsMaximizable();
-    [window_ setLevel:newLevel];
-    // Set level will make the zoom button revert to default, probably
-    // a bug of Cocoa or macOS.
-    [[window_ standardWindowButton:NSWindowZoomButton]
-        setEnabled:was_maximizable_];
-  } else {
-    *error = std::string([[NSString
-        stringWithFormat:@"relativeLevel must be between %d and %d",
```
loc: I removed this error in favor of a clamp because it's an absurdly wide range (-2147483643, 2147483631) and I didn't see much value in the error. Willing to add back on strong feelings, though.

shell/browser/api/atom_api_desktop_capturer.cc
```diff
@@ -94,43 +94,23 @@ void DesktopCapturer::StartHandling(bool capture_window,
           content::DesktopMediaID::TYPE_WINDOW,
           content::desktop_capture::CreateWindowCapturer());
       window_capturer_->SetThumbnailSize(thumbnail_size);
-      window_capturer_->AddObserver(this);
-      window_capturer_->StartUpdating();
+      window_capturer_->Update(base::BindOnce(
+          &DesktopCapturer::UpdateSourcesList, weak_ptr_factory_.GetWeakPtr(),
+          window_capturer_.get()));
```
loc: I'm not confident enough in my C++ to be sure this is the optimal way to pass these pointers; please double check.

codebytere:
@loc do you have links for the changes prompting the mas private api changes? normally i link the whole CL (ex. https://chromium-review.googlesource.com/c/chromium/src/+/1794322) so that it's a bit easier to grab contect for review :)

shell/browser/api/atom_api_desktop_capturer.h
```diff
@@ -64,6 +52,7 @@ class DesktopCapturer : public mate::TrackableObject<DesktopCapturer>,
   bool capture_window_ = false;
   bool capture_screen_ = false;
   bool fetch_window_icons_ = false;
+  base::WeakPtrFactory<DesktopCapturer> weak_ptr_factory_{this};
```
codebytere: we want `weak_ptr_factory_` is the last member variable in `DesktopCapturer` so that it's
destroyed first (ensures that if any class methods bound to a wk ptr instance
are `Run()` during teardown, they wont actually be executed)

loc:
@codebytere: I updated the commits with CL links 🙇

patches/chromium/ssl_security_state_tab_helper.patch
```diff
@@ -20,17 +20,22 @@ index 887cdd11fbbd9df44edef1548a89891c5cf23cee..12da72449622289ec467d284f616bbf9
  #include "chrome/browser/safe_browsing/safe_browsing_service.h"
  #include "chrome/browser/safe_browsing/ui_manager.h"
 +#endif
+ #include "chrome/browser/ssl/tls_deprecation_config.h"
```
deepak1556: This file should be included in the sources list under `chromium_src/BUILD.gn`

patches/chromium/ssl_security_state_tab_helper.patch
```diff
@@ -20,17 +20,22 @@ index 887cdd11fbbd9df44edef1548a89891c5cf23cee..12da72449622289ec467d284f616bbf9
...
+ #include "chrome/common/chrome_features.h"
+ #include "chrome/common/chrome_switches.h"
+ #include "chrome/common/pref_names.h"
  #include "chrome/common/secure_origin_whitelist.h"
 +#if 0
  #include "components/omnibox/browser/omnibox_field_trial.h"
  #include "components/omnibox/common/omnibox_features.h"
  #include "components/password_manager/core/browser/password_manager_metrics_util.h"
+ #include "components/prefs/pref_service.h"
```
deepak1556: We build the pref service, this need not be under `#if 0`

patches/chromium/ssl_security_state_tab_helper.patch
```diff
@@ -39,7 +44,19 @@ index 887cdd11fbbd9df44edef1548a89891c5cf23cee..12da72449622289ec467d284f616bbf9
  #include "chrome/browser/safe_browsing/chrome_password_protection_service.h"
  #endif
  
-@@ -66,8 +70,9 @@ void RecordSecurityLevel(
+@@ -70,6 +74,7 @@ void RecordSecurityLevel(
+   }
+ }
+ 
++#if 0
```
deepak1556: I don't see a reason why the warning of TLS deprecation utility should be excluded. The components are all in place and the only additional source needed is `tls_deprecation_config.h`

loc: The `IsLegacyTLS` function uses `g_browser_process` which we'd previously excluded, so I took it out. Without `IsLegacyTLS`, `SSLProtocolVersionFromString` is otherwise unused.

patches/chromium/ssl_security_state_tab_helper.patch
loc: Is it worth adding a deletion to the patch? My intuition is to try and leave as much Chromium source unchanged in the patch as possible (but I'm new to this!).

patches/chromium/ssl_security_state_tab_helper.patch
deepak1556: We mock the `g_browser_process` here https://github.com/electron/electron/blob/master/shell/browser/browser_process_impl.cc#L33 , so its harmless to include it. The place where we excluded was using a preference we don't set, hence it was excluded.

Here `prefs::kSSLVersionMin` is a preference that is loaded from command line https://cs.chromium.org/chromium/src/chrome/browser/prefs/chrome_command_line_pref_store.cc?l=123, we support this particular command line switch, hence I would suggest to parse the command line directly in `IsLegacyTLS` instead of reading pref value to retain the functionality.

patches/chromium/ssl_security_state_tab_helper.patch
deepak1556: My suggestion is not deletion here but rather retain the functionality from upstream https://github.com/electron/electron/pull/20397#discussion_r330804439 

```
//electron/chromium_src/BUILD.gn

static_library("chrome") {
...
sources = [
...
"//chrome/browser/ssl/security_state_tab_helper.h",
"//chrome/browser/ssl/tls_deprecation_config.h",
....
]
}
```

> My intuition is to try and leave as much Chromium source unchanged in the patch as possible

Yes that is the usual policy.

nornagon: I'm not sure it's that useful to list the header files in the sources list for our `chromium_src` target, GN will already pick up the .h dependency via `-MMD` analysis so listing the .h here is somewhat superfluous. Also it would be really hard to keep up with changes in #includes for every cc file we have here!

deepak1556: Oops yeah that actually is a better option to follow. Thanks for noticing!

Issue #20364, opened 2019-09-28
Regression of recentDocuments Menu role on macOS


bigtimebuddy:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

Defining a menu item like this no longer works in v6 where prior versions worked correctly. In v6, the menu item shows up in the menu disabled and does not have a submenu.

```js
{
    label: 'Open Recent',
    role: 'recentDocuments',
    submenu: [
        { role: 'clearRecentDocuments' }
    ]
}
```

* **Electron Version:**
  * 6.0.10
* **Operating System:**
  * macOS 10.14.5 
* **Last Known Working Electron version:**
  * 5.0.11

### Expected Behavior

`Open Recent` menu item should be enabled and contain a submenu that is expandable.

### Actual Behavior

`Open Recent` menu item is disabled without a submenu.

### To Reproduce

Example, see `main.js` for the menu setup.

```sh
$ git clone git@github.com:bigtimebuddy/electron-recent-docs-bug.git -b master
$ npm install
$ npm start || electron .
```

### Screenshots

v6.x
![Screen Shot 2019-09-27 at 5 24 51 PM](https://user-images.githubusercontent.com/864393/65808808-ce36b780-e14b-11e9-9a51-b1877f66c9d1.png)

v5.x
![Screen Shot 2019-09-27 at 5 25 19 PM](https://user-images.githubusercontent.com/864393/65808811-d2fb6b80-e14b-11e9-9d71-3f37fe6b8fc3.png)

codebytere:
Caused by https://github.com/electron/electron/pull/18626; looking into a fix now.

bigtimebuddy:
Yeah! Thanks so much @codebytere for looking into this.

bigtimebuddy:
cc @codebytere 

Could we reopen this @zcbenz? `recentDocuments` seems like it is still broken in 6.0.12. The menu items work (are now enabled) but the `recentDocuments` menu does not populate with files. I have updated the repo I provided above:

[...]

PR #20398, opened 2019-10-01
fix: recentDocuments menu role on macOS

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/20364.

In the past when we constructed our menu through `MainMenu.xib`, we got Chromium's submenu construction defaults for free, which is why prior to `v6.0.0-beta.1` the `recentDocuments` role worked as we intended. The `recentDocuments` NSMenuItem would never have a `nil` submenu, since barring any actual recent documents it had a default `Clear Menu` item (see `.xib` file in linked CL). When they removed their dependence on MainMenu.xib via  [this CL](https://chromium-review.googlesource.com/c/chromium/src/+/1627242) it had the unfortunate side-effect that [this call](https://github.com/electron/electron/blob/eb2d2264d02abd7f34df36809d44752aaf93da93/shell/browser/ui/cocoa/atom_menu_controller.mm#L196-L197) would always return a null pointer, and so `[item setSubmenu:recentDocumentsMenu]` would _never set a submenu_ and thus never make the NSMenuItem's `action` selector `submenuAction:`. As a result, `validateUserInterfaceItem` would return `NO` and the item would both have no submenu and also be mysteriously disabled. 

To fix this, we now allocate an `NSMenu` for recent documents and append the old default menu item to it if no submenu exists (ie we're appending it for the first time). I've also gone ahead and added a pair of DCHECKs to prevent a similar issue from occurring in the future.
 
cc @nornagon @jkleinsc @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a regression in the `recentDocuments` MenuItem role on macOS.

### DIFF
--- a/shell/browser/ui/cocoa/atom_menu_controller.mm
+++ b/shell/browser/ui/cocoa/atom_menu_controller.mm
@@ -126,11 +126,13 @@ static base::scoped_nsobject<NSMenu> recentDocumentsMenuSwap_;
   if (!menu_)
     return;
 
+  // Locate & retain the recent documents menu item
   if (!recentDocumentsMenuItem_) {
-    // Locate & retain the recent documents menu item
-    recentDocumentsMenuItem_.reset(
-        [[[[[NSApp mainMenu] itemWithTitle:@"Electron"] submenu]
-            itemWithTitle:@"Open Recent"] retain]);
+    base::string16 title = base::ASCIIToUTF16("Open Recent");
+    NSString* openTitle = l10n_util::FixUpWindowsStyleLabel(title);
+
+    recentDocumentsMenuItem_.reset([[[[[NSApp mainMenu]
+        itemWithTitle:@"Electron"] submenu] itemWithTitle:openTitle] retain]);
   }
 
   model_ = model;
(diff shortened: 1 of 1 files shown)


shell/browser/ui/cocoa/atom_menu_controller.mm
```diff
@@ -193,8 +193,18 @@ - (void)moveMenuItems:(NSMenu*)source to:(NSMenu*)destination {
...
-  NSMenu* recentDocumentsMenu =
-      [[[recentDocumentsMenuItem_ submenu] retain] autorelease];
+  NSMenu* recentDocumentsMenu = nil;
+  if ([recentDocumentsMenuItem_ submenu]) {
+    recentDocumentsMenu =
+        [[[recentDocumentsMenuItem_ submenu] retain] autorelease];
+  } else {
+    recentDocumentsMenu = [[NSMenu alloc] initWithTitle:@""];
+    NSMenuItem* clearItem =
+        [[NSMenuItem alloc] initWithTitle:@"Clear Menu"
```
nornagon: Does this need to be localized?

codebytere: not particularly :)

shell/browser/ui/cocoa/atom_menu_controller.mm
```diff
@@ -193,8 +193,18 @@ - (void)moveMenuItems:(NSMenu*)source to:(NSMenu*)destination {
 // Replaces the item's submenu instance with the singleton recent documents
 // menu. Previously replaced menu items will be recovered.
 - (void)replaceSubmenuShowingRecentDocuments:(NSMenuItem*)item {
-  NSMenu* recentDocumentsMenu =
-      [[[recentDocumentsMenuItem_ submenu] retain] autorelease];
+  NSMenu* recentDocumentsMenu = nil;
+  if ([recentDocumentsMenuItem_ submenu]) {
+    recentDocumentsMenu =
+        [[[recentDocumentsMenuItem_ submenu] retain] autorelease];
```
zcbenz: `[[retain] autorelease]` is no-op, `recentDocumentsMenu = [recentDocumentsMenuItem_ submenu]` is enough.

shell/browser/ui/cocoa/atom_menu_controller.mm
```diff
@@ -193,8 +193,18 @@ - (void)moveMenuItems:(NSMenu*)source to:(NSMenu*)destination {
...
+  } else {
+    recentDocumentsMenu = [[NSMenu alloc] initWithTitle:@""];
```
zcbenz: Should do `autorelease` here.

shell/browser/ui/cocoa/atom_menu_controller.mm
```diff
@@ -193,8 +193,18 @@ - (void)moveMenuItems:(NSMenu*)source to:(NSMenu*)destination {
...
+    [recentDocumentsMenu
+        addItem:[[NSMenuItem alloc]
```
zcbenz: Should do `autorelease` for the menu item.

shell/browser/ui/cocoa/atom_menu_controller.mm
nornagon: Will this show in English to users whose system language is set to a non-English language?

codebytere: ah great catch, ty 🙇‍♀

PR #20399, opened 2019-10-01
feat: add app.getApplicationNameForProtocol API

ajmacd:
#### Description of Change
This API returns the name of the application registered as the default handler for a protocol (e.g. `irc://`). We're going to use this functionality in Slack to enable a "native app linking" feature, whereby Slack apps will optionally provide URLs to native applications alongside http URLs, which we'll attempt to launch if the correct native application is installed on the user's machine.

It fits well with the other protocol functionality provided by the `Browser` class. `GetApplicationNameForUrl` is an existing method in Chromium:
https://chromium.googlesource.com/chromium/src/+/ed716f08f330dda84f38b7f40f3e381cd6af571a/chrome/browser/shell_integration.h#65

and my initial attempt was simply to call down to it. Sadly, this would appear to require depending on the entire `//chrome/browser` module, which brings in something like a highly undesirable 3k files. Instead I copied the implementations out to Electron, which again is consistent with some of the other methods having similar or identical implementations in Chromium:
* `SetAsDefaultProtocolClient`
* `IsDefaultProtocolClient`

The Chromium Linux implementation is fairly useless, simply returning `"xdg-open"`:
https://chromium.googlesource.com/chromium/src/+/eaa08411bd1f60a4e9d4b3253ba4710ae05bff3c/chrome/browser/shell_integration_linux.cc#720

but I discovered a useful implementation relying on `xdg-mime` wasn't all that hard. I'd like to upstream this to Chromium later. (Again, we won't be relying on that implementation, but just to share the spoils as it were.) This also avoids some `DCHECK`s I hit around disallowed blocking calls on Linux, in both the new and existing functionality. 

#### Checklist

- [x] `npm test` passes

#### Release Notes

notes: Added `app.getApplicationNameForProtocol()` API


shell/browser/browser_linux.cc
```diff
@@ -113,6 +131,16 @@ bool Browser::RemoveAsDefaultProtocolClient(const std::string& protocol,
   return false;
 }
 
+base::string16 Browser::GetApplicationNameForProtocol(const GURL& url) {
+  std::vector<std::string> argv;
```
codebytere: kinda-nit: for simplicity we can probably just do:
```
const std::string& handler = std::string("x-scheme-handler/") + url.scheme();
argv = {"xdg-mime", "query", "default", handler};
```

spec-main/api-app-spec.ts
```diff
@@ -876,6 +876,32 @@ describe('app module', () => {
         })
       })
     })
+
+    it('sets the default client such that getApplicationNameForProtocol returns Electron', (done) => {
+      app.setAsDefaultProtocolClient(protocol);
+      expect(app.getApplicationNameForProtocol(`${protocol}://`)).to.equal('Electron')
+      done()
```
codebytere: if this isn't an explicitly async new method call (docs don't indicate it is unless i missed something) we probably don't need this `done` here

spec-main/api-app-spec.ts
```diff
@@ -876,6 +876,32 @@ describe('app module', () => {
...
+      const protocols = [
+        'http://',
+        'https://',
+      ]
+      // We can't expect particular app names here, but these protocols should
+      // at least have _something_ registered.
+      protocols.forEach((protocol) => {
+        expect(app.getApplicationNameForProtocol(protocol)).to.not.equal('')
+      })
+      done()
```
codebytere: same as above

docs/api/app.md
```diff
@@ -770,6 +770,21 @@ macOS machine. Please refer to
 
 The API uses the Windows Registry and LSCopyDefaultHandlerForURLScheme internally.
 
+### `app.getApplicationNameForProtocol(url)`
+
+* `url` String - a URL with the protocol name to check. Unlike the other
+  methods in this family, this accepts an entire URL, including `://` at a
+  minimum (e.g. `https://`).
+
+Returns String - Name of the application handling the protocol, or an empty
```
codebytere:
```suggestion
Returns `String` - Name of the application handling the protocol, or an empty
```

spec-main/api-app-spec.ts
codebytere: if this isn't an explicitly async new method call (docs don't indicate it is unless i missed something?) we probably don't need this `done` here

spec-main/api-app-spec.ts
codebytere: same as above

shell/browser/browser_linux.cc
codebytere: kinda-nit: for simplicity we can probably just do:
```cpp
const std::string& handler = std::string("x-scheme-handler/") + url.scheme();
argv = {"xdg-mime", "query", "default", handler};
```

shell/browser/browser_mac.mm
```diff
@@ -131,6 +132,22 @@
...
 }
 
+base::string16 Browser::GetApplicationNameForProtocol(const GURL& url) {
+  NSURL* ns_url = [NSURL
+      URLWithString:base::SysUTF8ToNSString(url.possibly_invalid_spec())];
+  base::ScopedCFTypeRef<CFErrorRef> out_err;
+  base::ScopedCFTypeRef<CFURLRef> openingApp(LSCopyDefaultApplicationURLForURL(
+      (CFURLRef)ns_url, kLSRolesAll, out_err.InitializeInto()));
+  if (out_err) {
+    // likely kLSApplicationNotFoundErr
```
codebytere: let's `LOG(ERROR)` this error here?

miniak:
I am not sure if this API belongs to the app module, it does not operate on the current application instance.

spec-main/api-app-spec.ts
ajmacd: Ah right, thanks.

shell/browser/browser_linux.cc
ajmacd: Good idea. Updated the other similar uses as well.

ajmacd:
> I am not sure if this API belongs to the app module, it does not operate on the current application instance. What about moving it to systemPreferences instead?

I can see your point, and in a vacuum, I agree, it seems like this new API would be a better fit for `systemPreferences`. However, it's closely related to the existing protocol APIs and I think they should be grouped together. Thoughts?

ajmacd:
Hello fine reviewers: any more thoughts on my comments? Happy to discuss further if you disagree!

ajmacd:
Sorry should have @codebytere and @miniak in the last comment ^ 🙇

ajmacd:
@codebytere, @miniak any comments on this? I'd like to land it, but am happy to discuss further changes.

ajmacd:
@zcbenz would you be OK with landing this, or would you prefer to wait for @codebytere and @miniak's approval?

zcbenz:
If no opposition I would like to land this.

@ajmacd Do you mind rebasing this branch on master as there is conflict?

ajmacd:
I actually chatted quickly with @codebytere about this today. She mentioned wanting to bring it up at the API working group meeting.

codebytere:
@ajmacd this needs a final conflict fix: 

```
Unhandled exception in main spec runner: electron/spec-main/api-app-spec.ts(866,43): error TS2304: Cannot find name 'isCI'.
```

ajmacd:
@codebytere I think we're good to go here. The build-linux failure is unrelated to this change.

jkleinsc:
Approved by the Releases WG for backport to 8-x-y

PR #20411, opened 2019-10-03
test: close window before destroying browserView

zcbenz:
#### Description of Change

This PR fixes the flaky Windows CI caused by `BrowserView` tests.

In the `BrowserView` tests, the `browserView` is destroyed before the parent window gets closed, and this may result in the parent window accessing destroyed pointer.

This PR avoids the crash by closing parent window before destroying `browserView`.

A complete fix would be correctly handle the case when an in-use `browserView` gets destroyed, but it is not easy to get right and I don't think it is really necessary for now.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/spec-main/api-browser-view-spec.ts
+++ b/spec-main/api-browser-view-spec.ts
@@ -27,11 +27,11 @@ describe('BrowserView module', () => {
   })
 
   afterEach(async () => {
+    await closeWindow(w)
+
     if (view) {
       view.destroy()
     }
-
-    await closeWindow(w)
   })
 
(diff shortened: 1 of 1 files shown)


deepak1556:
> A complete fix would be correctly handle the case when an in-use browserView gets destroyed, but it is not easy to get right and I don't think it is really necessary for now.

Cannot agree more, the shutdown sequence of embedded views must be revisited and see if we can manage both browserview and webview destruction in the same code path.

PR #20416, opened 2019-10-03
fix: enable worker threads in ELECTRON_RUN_AS_NODE

codebytere:
#### Description of Change

Fix Node.js' `worker_threads` module when Electron is run with `ELECTRON_RUN_AS_NODE`. The fundamental issue here is that Electron sets `NODE_USE_V8_PLATFORM` to `false`, because Electron initializes the v8 platform itself (in `javascript_environment.cc`), and Node.js worker thread initialization relies on the `platform` it uses having been set inside code guarded by `NODE_USE_V8_PLATFORM`.

This commit fixes this problem by changing `node_worker.cc` to use the three-arg
implementation of `NewIsolate` to prevent it trying to use a possibly-null ptr.
cc @ckerr @nornagon @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed Node.js' `worker_threads` in `ELECTRON_RUN_AS_NODE`.


patches/node/fix_enable_worker_threads.patch
```diff
@@ -0,0 +1,83 @@
...
++++ b/src/node_v8_platform-inl.h
+@@ -148,9 +148,22 @@ struct V8Platform {
+   NodePlatform* platform_;
+ #else   // !NODE_USE_V8_PLATFORM
+   inline void Initialize(int thread_pool_size) {}
+-  inline void Dispose() {}
+-  inline void DrainVMTasks(v8::Isolate* isolate) {}
+-  inline void CancelVMTasks(v8::Isolate* isolate) {}
++  inline void Dispose() {
++    if (platform_) {
```
deepak1556: This means platform is owned by node context ? `platform_` was originally leaked from electron context. With this that case must be guarded, otherwise possible UAF scenario. check https://github.com/electron/electron/blob/master/shell/app/node_main.cc#L139-L141

codebytere: i added guards based on my understanding of what you mean here but please let me know if i missed something!

Inomezi:
Hi, is it working now ? I'm Using electron 6.0.11 , but still got error The V8 platform used by this instance of Node does not support creating Workers

Inomezi:
Hello, it still isn't work ?

codebytere:
@Inomezi what _released version_ did you test with?

Inomezi:
> @Inomezi what _released version_ did you test with?

I tested it on version 6.0.12 and 7.0.0-beta.6. When i calling `new Worker("path")` i getting error: Uncaught Error: The V8 platform used by this instance of Node does not support creating Workers.
What's version i need to use ?

addaleax:
> @Inomezi that's a different issue; i'm still working on that! It's not fixed yet.

@codebytere Was that something we discussed in the last weeks? Is that anything that I could help with?

PR #20417, opened 2019-10-03
test: move crashReporter specs to the main process

nornagon:
#### Description of Change
This moves the specs for the crashReporter module to the main process, and substantially refactors them to be more readable & straightforward. It's not quite testing the same thing now—we're starting the crash reporter from the main process instead of from the renderer—but I think that's better actually, since starting the crash reporter from the renderer is kind of weird.

There's no way to shut down the crash reporter, though, so this might affect tests that run after this module? I think that was also the case with the old test suite though.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


nornagon:
Ah, I see the old version had special code to guarantee that these tests run last.

docs/api/crash-reporter.md
```diff
@@ -119,6 +119,14 @@ Remove a extra parameter from the current set of parameters so that it will not
 
 See all of the current parameters being passed to the crash reporter.
 
+### `crashReporter.getCrashesDirectory()`
+
+Returns `String` - The directory where crashes are temporarily stored before being uploaded.
+
+### `crashReporter.getProductName()`
```
miniak: I am not sure we want to make this a public API

miniak: I think it was only mean to be used by the tests

spec-main/api-crash-reporter-spec.ts
```diff
@@ -0,0 +1,448 @@
...
+      nodeIntegration: true
+    }
+  })
+  generateSpecs('with sandbox', {
+    webPreferences: {
+      sandbox: true,
+      preload: path.join(fixtures, 'module', 'preload-sandbox.js')
+    }
+  })
+  generateSpecs('with remote module disabled', {
```
miniak: I don't think we need this anymore

docs/api/crash-reporter.md
happys33: miniak  u are great am new here thx for ur help >  that guys name nornagon and Shelley Vohr Jeremy Apthorp and one  googleapis/nodejs remove my job coz its there pic and madia voice ... i dont know how stop that can u help me .. u can check main and see

nornagon: In that case can we delete the function? It doesn't seem to be used by anything...

miniak: I would delete them

PR #20421, opened 2019-10-04
feat: Implement `webContents.menuShortcutPriority` to control shortcut handling order

poiru:
This adds the ability to prioritize application menu shortcuts over web
content shortcuts. Previously this was only possible by listening to the
`before-input-event` event and calling `preventDefault()` on the relevant
shortcuts. This also deprecates `webContents.setIgnoreMenuShortcuts` in
favor of the new property.

See added docs for more info.

Notes: Added `webContents.menuShortcutPriority` to control shortcut handling order


poiru:
I can look into adding tests after people are happy with the API.

poiru:
I now think it makes more sense to have this as a property on `MenuItem`s so that you can prioritize only certain shortcuts. Will open a new PR after changing!

PR #20427, opened 2019-10-04
refactor: take advantage of structured clone algorithm in the remote module

miniak:
#### Description of Change
Take advantage of structured clone algorithm based IPC introduced by #20214:
- `Date`, `ArrayBuffer` and `ArrayBufferView` can now be serialized as values.
- Keeping special handling to preserve `Buffer`, otherwise it will be converted to `UInt8Array`.
- Use native serialization for `Boolean`, `Number`, `String` and `RegExp`.
- `Error` is now serialized natively. Preserve custom properties in the `remote` module.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: The `remote` module now properly serializes `Boolean`, `Number`, `String` and `RegExp` instances.


miniak:

lib/browser/remote/server.ts
```diff
@@ -186,8 +182,6 @@ const valueToMeta = function (sender: electron.WebContents, contextId: string, v
       value: value.name
     })
     return { type, members }
-  } else if (type === 'date') {
-    return { type, value: value.getTime() }
   } else {
     return {
       type: 'value',
```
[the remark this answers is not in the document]
miniak: I know, but the default serialization looses custom properties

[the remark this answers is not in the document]
miniak: There is a test explicitly enforcing that, which started failing when I did the change for `Error` as well

lib/browser/remote/server.ts
```diff
@@ -128,14 +124,14 @@ const valueToMeta = function (sender: electron.WebContents, contextId: string, v
     // Recognize certain types of objects.
     if (value === null) {
```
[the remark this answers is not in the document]
miniak: moved the check for `null` into the `isSerializable`, and removed from `lib/renderer/api/remote.js` as well

lib/browser/remote/server.ts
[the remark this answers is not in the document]
miniak: done + added code to preserve custom properties

miniak:
@nornagon do we mind loosing the `from` property indicating which process the `Error` came from, which was added by the old `errorUtils` serializer?

lib/common/remote/type-utils.ts
```diff
@@ -10,3 +10,16 @@ export function isPromise (val: any) {
...
   )
 }
+
+const serializableTypes = [
+  Boolean,
+  Number,
+  String,
+  Date,
+  RegExp,
+  ArrayBuffer
```
nornagon: `Map`, `Set`?

lib/renderer/api/remote.js
```diff
@@ -43,17 +41,17 @@ function wrapArgs (args, visited = new Set()) {
       }
       visited.delete(value)
```
nornagon: it's still a mystery to me why we have two separate serialization algorithms

lib/common/remote/type-utils.ts
miniak: those can contain other objects, which require custom serialization.

lib/renderer/api/remote.js
miniak: do you mean main -> renderer and renderer -> main?

nornagon: yes

PR #20437, opened 2019-10-05
chore: remove unused abort/assert exposure patch

codebytere:
#### Description of Change

We no longer use `node::Abort()` or `node::Assert()` in our codebase, and thus no longer need to expose these private methods from Node.

We also need to include `node.h` in `node_native_module_env.h` after this change, as `NODE_EXTERN` won't be defined without it and it was previously compiling through transitive includes as a result of its inclusion in the deleted patch.

cc @nitsakh 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


codebytere:
Welp, looks like this was needed by the debug builds. Ah well - i'll leave this in for now and try to figure out a better way to solve this in future.

Issue #20392, opened 2019-10-01
Filter cookies by domain is not working in 7.0.0


btzsoft:
### Issue Details

* **Electron Version:**
  * 7.0.0-beta.5
* **Operating System:**
  * Windows
* **Last Known Working Electron version:**
  * 6

### Expected Behavior
```js
const filter = {
  domain: 'facebook.com'
}
const cookies = await webContents.session.cookies.get(filter) // will resolve with actual cookies
```

### Actual Behavior
```js
const filter = {
  domain: 'facebook.com',
  // url: 'https://facebook.com' - NOTE: filtering by url is working
}
const cookies = await webContents.session.cookies.get(filter) // is returning empty array
```

### To Reproduce
```js
const filter = {
  domain: 'facebook.com'
}
const cookies = await webContents.session.cookies.get(filter)
```

Filter by `domain` is returning empty array, but if I try to filter using `url` is returning the actual cookies.

Tested on all `7` beta versions.
On `6` filtering by domain was worked.

codebytere:
cc @nornagon/@zcbenz this seems like a nw13n issue?

btzsoft:
@codebytere when it's expected to have a new build which will include the fix? thx.

PR #20471, opened 2019-10-08
fix: cookies.get should be able to filter domain

zcbenz:
#### Description of Change

Fix https://github.com/electron/electron/issues/20392.

This PR fixes a bug that wrong API was used when `url` is not specified, this bug was introduced when migrating to NetworkService APIs.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix `cookies.get` not working when `url` is not specified in filter.


btzsoft:
@zcbenz what about `.set` method, will work without url?

zcbenz:
Yeah the `.set` API works without url.

PR #20482, opened 2019-10-08
fix: set proper constraints for windows with defined maxWidth

CezaryKulakowski:
When BrowserWindow has set constraints for width (max or min) it
won't behave correctly during first attempt of resizing it. When
maxWidth is defined and maxWidth equals its width it will shrink
rapidly when user tries to expand its width. On the other hand
when minWidth is defined and minWidth equals its width it's
possible to decrease its width with a few pixels.

Here is a sample app which can be used to reproduce the problem:

```

async function createWindow() {
  let win = new BrowserWindow({
    width: 800,
    height: 600,
    maxWidth: 800
  });
  win.loadURL('http://www.example.com');
}

app.on('ready', createWindow)
```

Notes: Fixed improper behaviour of window with width constraint set during resize.
#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:

### DIFF
--- a/shell/browser/api/atom_api_cookies.cc
+++ b/shell/browser/api/atom_api_cookies.cc
@@ -15,6 +15,7 @@
 #include "content/public/browser/storage_partition.h"
 #include "gin/dictionary.h"
 #include "gin/object_template_builder.h"
+#include "native_mate/dictionary.h"
 #include "net/cookies/canonical_cookie.h"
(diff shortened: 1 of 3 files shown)


CezaryKulakowski:
It's rebased to master now.

CezaryKulakowski:
I've created new pr with updated description and with more generic change which fixes more scenarios.

PR #20489, opened 2019-10-08
docs: add a dev doc describing the patch system

nornagon:
#### Description of Change
This adds documentation for how we apply patches to upstream, including the `git-import-patches` and `git-export-patches` tool.

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/docs/development/patches.md
+++ b/docs/development/patches.md
@@ -6,7 +6,7 @@ Electron is built on two major upstream projects: Chromium and Node.js. Each of 
 
 Every patch in Electron is a maintenance burden. When upstream code changes, patches can break—sometimes without even a patch conflict or a compilation error. It's an ongoing effort to keep our patch set up-to-date and effective. So we strive to keep our patch count at a minimum. To that end, every patch must describe its reason for existence in its commit message. That reason must be one of the following:
 
-1. The patch is temporary, and is intended to be (or has been) committed upstream. Include a link to an upstream PR or code review if available.
+1. The patch is temporary, and is intended to be (or has been) committed upstream or otherwise eventually removed. Include a link to an upstream PR or code review if available, or a procedure for verifying whether the patch is still needed at a later date.
 2. The patch allows the code to compile in the Electron environment, but cannot be upstreamed because it's Electron-specific (e.g. patching out references to Chrome's `Profile`). Include reasoning about why the change cannot be implemented without a patch (e.g. by subclassing or copying the code).
 3. The patch makes Electron-specific changes in functionality which are fundamentally incompatible with upstream.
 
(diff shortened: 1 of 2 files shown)


docs/development/patches.md
```diff
@@ -0,0 +1,84 @@
...
+⋮
+```
+
+To help manage these patch sets, we provide two tools: `git-import-patches` and `git-export-patches`. `git-import-patches` imports a set of patch files into a git repository by applying each patch in the correct order and creating a commit for each one. `git-export-patches` does the reverse; it exports a series of git commits in a repository into a set of files in a directory and an accompanying `.patches` file.
+
+> Side note: the reason we use a `.patches` file to maintain the order of applied patches, rather than prepending a number like `001-` to each file, is because it reduces conflicts related to patch ordering. It prevents the situation where two PRs both add a patch at the end of the series with the same numbering and end up both getting merged resulting in a duplicate identifier, and it also reduces churn when a patch is added or deleted in the middle of the series.
+
+### Usage
+
+#### Adding a new patch
```
jkleinsc: I'm wondering if we should explain why git commit is important, eg something like:
```suggestion
#### Adding a new patch
To add a new patch, edit the file you wish to apply changes to and then `git commit` that file so that the `git-export-patches` script can use that commit to create the patch.
```

docs/development/patches.md
```diff
@@ -0,0 +1,84 @@
...
+Re-exporting patches will sometimes cause shasums in unrelated patches to change. This is generally harmless and can be ignored (but go ahead and add those changes to your PR, it'll stop them from showing up for other people).
+
+#### Editing an existing patch
+```
+$ cd src/v8
+$ vim some/code/file.cc
+$ git log
+# Find the commit sha of the patch you want to edit.
+$ git commit --fixup [COMMIT_SHA]
+$ git rebase --autosquash -i [COMMIT_SHA]^
```
jkleinsc: This is a much easier way of editing a patch than the way I was doing it.  🤦‍♂

docs/development/patches.md
```diff
@@ -0,0 +1,84 @@
...
+
+#### Resolving conflicts
+When updating an upstream dependency, patches may fail to apply cleanly. Often, the conflict can be resolved automatically by git with a 3-way merge. You can instruct `git-import-patches` to use the 3-way merge algorithm by passing the `-3` argument:
+
+```
+$ cd src/third_party/electron_node
+$ ../../electron/script/git-import-patches -3 ../../electron/patches/node
+```
+
+If `git-import-patches -3` encounters a merge conflict that it can't resolve automatically, it will pause and allow you to resolve the conflict manually. Once you have resolved the conflict, `git add` the resolved files and continue to apply the rest of the patches by running `git am --continue`.
```
jkleinsc: In the upgrade hackmd we say to run:
 `git am --abort`, then `git checkout refs/patches/upstream-head` before running the 3 way merge.  Is that necessary and should it be listed here?

docs/development/patches.md
nornagon: Glad to be of service!

PR #20496, opened 2019-10-09
fix: cookies.get should be able to filter domain (7-0-x)

zcbenz:
Backport of https://github.com/electron/electron/pull/20471.

Notes: Fix `cookies.get` not working when `url` is not specified in filter.

### DIFF
--- a/shell/browser/api/atom_api_cookies.cc
+++ b/shell/browser/api/atom_api_cookies.cc
@@ -15,6 +15,7 @@
 #include "content/public/browser/storage_partition.h"
 #include "gin/dictionary.h"
 #include "gin/object_template_builder.h"
+#include "native_mate/dictionary.h"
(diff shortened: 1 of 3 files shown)


btzsoft:
@zcbenz setting cookie isn't working in any way.
I try `webContents.session.cookies.set({domain: '.facebook.com', name: 'my cookie', value: 'my value'})`
but I receive `Failed to get cookies domain` error.
When I try with url the promise is resolved but no any cookies is set in any way.
@codebytere @jkleinsc cc

btzsoft:
@jkleinsc `v7.0.0-beta.7` installed few minutes ago.

zcbenz:
@btzsoft While the documentation says `url` is optional for `cookies.set`, but it is actually required. This is a documentation error and I believe this behavior has been there at least since 5.x.

PR #20542, opened 2019-10-11
ci: skip build on doc only changes

jkleinsc:
#### Description of Change

This PR changes CI to only run lint and not run builds when a commit only contains doc changes.
For CircleCI every job still runs but is short circuited if it is determined that the change is a doc only change.  We may be able to eventually use GitHub actions to be more efficient and not run all jobs, but this gets the job done for now.

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -711,7 +711,7 @@ step-check-for-doc-only-change: &step-check-for-doc-only-change
     command: |
       cd src/electron
       node script/yarn install --frozen-lockfile
-      if [ node script/doc-only-change.js --prNumber=$CIRCLE_PR_NUMBER --prURL=$CIRCLE_PULL_REQUEST`" -eq 1 ]; then
+      if [ node script/doc-only-change.js --prNumber=$CIRCLE_PR_NUMBER --prURL=$CIRCLE_PULL_REQUEST -eq 1 ]; then
         echo "true" > .skip-ci-build
       else
         touch .skip-ci-build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -704,6 +704,50 @@ step-save-src-cache: &step-save-src-cache
     key: v5-src-cache-{{ arch }}-{{ checksum "src/electron/.depshash" }}
     name: Persisting src cache
 
+# Check for doc only change
+step-check-for-doc-only-change: &step-check-for-doc-only-change
+  run:
+    name: Check if commit is doc only change
+    command: |
+      cd src/electron
+      node script/yarn install --frozen-lockfile
+      export DOC_ONLY_CHANGE="`node script/doc-only-change.js --prNumber=$CIRCLE_PR_NUMBER --prURL=$CIRCLE_PULL_REQUEST`"
+      if [ "$DOC_ONLY_CHANGE" = "true" ]; then
+        echo $DOC_ONLY_CHANGE > .skip-ci-build
+      else
+        touch .skip-ci-build
+      fi
+
+step-persist-doc-only-change: &step-persist-doc-only-change
+  persist_to_workspace:
+    root: .
+    paths:
+      - src/electron/.skip-ci-build
+
+step-maybe-early-exit-doc-only-change: &step-maybe-early-exit-doc-only-change
+  run:
+    name: Shortcircuit build if doc only change
+    command: |
+      if [ -s src/electron/.skip-ci-build ]; then
+        circleci-agent step halt
+      fi
+
+step-maybe-early-exit-no-doc-change: &step-maybe-early-exit-no-doc-change
+  run:
+    name: Shortcircuit job if change is not doc only
+    command: |
+      if [ ! -s src/electron/.skip-ci-build ]; then
+        circleci-agent step halt
+      fi
+
+step-ts-compile: &step-ts-compile
+  run:
+    name: Run TS/JS compile on doc only change
+    command: |
+      cd src
+      ninja -C out/Default electron:default_app_js -j $NUMBER_OF_NINJA_PROCESSES
+      ninja -C out/Default electron:atom_js2c -j $NUMBER_OF_NINJA_PROCESSES
+
 # Lists of steps.
 steps-lint: &steps-lint
   steps:
@@ -752,6 +796,9 @@ steps-lint: &steps-lint
 steps-checkout-fast: &steps-checkout-fast
   steps:
     - *step-checkout-electron
+    - *step-check-for-doc-only-change
+    - *step-persist-doc-only-change
+    - *step-maybe-early-exit-doc-only-change
     - *step-depot-tools-get
     - *step-depot-tools-add-to-path
     - *step-restore-brew-cache
@@ -783,6 +830,9 @@ steps-checkout-fast: &steps-checkout-fast
 steps-checkout-and-save-cache: &steps-checkout-and-save-cache
   steps:
     - *step-checkout-electron
+    - *step-check-for-doc-only-change
+    - *step-persist-doc-only-change
+    - *step-maybe-early-exit-doc-only-change
     - *step-depot-tools-get
     - *step-depot-tools-add-to-path
     - *step-restore-brew-cache
@@ -813,6 +863,7 @@ steps-electron-gn-check: &steps-electron-gn-check
   steps:
     - attach_workspace:
         at: .
+    - *step-maybe-early-exit-doc-only-change
     - *step-depot-tools-add-to-path
     - *step-setup-env-for-build
     - *step-gn-gen-default
@@ -822,6 +873,7 @@ steps-electron-build: &steps-electron-build
   steps:
     - attach_workspace:
         at: .
+    - *step-maybe-early-exit-doc-only-change
     - *step-depot-tools-add-to-path
     - *step-setup-env-for-build
     - *step-restore-brew-cache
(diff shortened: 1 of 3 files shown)


jkleinsc:
@nornagon this change still runs `lint` for doc only changes which in turn runs `create-typescript-definitions` which should end up catching TS errors

nornagon:
I don't think `create-typescript-definitions` catches errors during the _build_ of the .ts files in lib/, only errors in the descriptions themselves.

jkleinsc:
CI run with non doc changes:
https://github.com/electron/electron/pull/20542/checks?sha=88e4f66d30ac69ecff670ba58bc3b1a4015b84ac
https://ci.appveyor.com/project/electron-bot/electron-ia32-testing/builds/28052013
https://ci.appveyor.com/project/electron-bot/electron-ldhmv/builds/28052012
https://ci.appveyor.com/project/electron-bot/electron-ljo26/builds/28052011

nornagon:
> build time per-platform and per-feature exclusions

we only CI with one set of feature definitions so I think that's OK. I don't see any platform-specific TS files..?

MarshallOfSound:
> I don't see any platform-specific TS files..?

Yet, it will be possible quite soon

jkleinsc:
Successful doc only change here:
https://github.com/electron/electron/runs/262604321
https://ci.appveyor.com/project/electron-bot/electron-ia32-testing/builds/28155557
https://ci.appveyor.com/project/electron-bot/electron-ldhmv/builds/28155554
https://ci.appveyor.com/project/electron-bot/electron-ljo26/builds/28155556

jkleinsc:
I refactored this PR a bit because I realized that what we really need to check is if the PR as a whole is a doc change versus a particular commit being a doc change.  If we just look at a particular commit and someone pushes two (or more) commits at a time we might accidentally skip needed build and testing

.circleci/config.yml
```diff
@@ -704,6 +704,50 @@ step-save-src-cache: &step-save-src-cache
...
 
+# Check for doc only change
+step-check-for-doc-only-change: &step-check-for-doc-only-change
+  run:
+    name: Check if commit is doc only change
+    command: |
+      cd src/electron
+      node script/yarn install --frozen-lockfile
+      export DOC_ONLY_CHANGE="`node script/doc-only-change.js --prNumber=$CIRCLE_PR_NUMBER --prURL=$CIRCLE_PULL_REQUEST`"
+      if [ "$DOC_ONLY_CHANGE" = "true" ]; then
```
nornagon: better to use the return code, e.g.

```bash
if node script/doc-only-change.js ...; then
  # ...
```

script/doc-only-change.js
```diff
@@ -0,0 +1,58 @@
+const args = require('minimist')(process.argv.slice(2))
+const octokit = require('@octokit/rest')()
+const path = require('path')
+
+const SOURCE_ROOT = path.normalize(path.dirname(__dirname))
+
+async function checkIfDocOnlyChange () {
+  if (args.prNumber || args.prBranch || args.prURL) {
```
nornagon: Hm... I don't think you need the PR number to get the diff. We should be able to use `git merge-base` to figure out the root of the branch, e.g.

```
$ git merge-base HEAD origin/master
```

should give you the base commit of the branch. Then you can do the diff from that base to `HEAD`.

appveyor.yml
```diff
@@ -50,6 +50,14 @@ build_script:
   - ps: >-
       if(($env:APPVEYOR_PULL_REQUEST_HEAD_REPO_NAME -split "/")[0] -eq ($env:APPVEYOR_REPO_NAME -split "/")[0]) {
         Write-warning "Skipping PR build for branch"; Exit-AppveyorBuild
+      } else {
+        npx.cmd yarn@1.15.2 install --frozen-lockfile
+
+        node script/doc-only-change.js --prNumber=$env:APPVEYOR_PULL_REQUEST_NUMBER --prBranch=$env:APPVEYOR_REPO_BRANCH
+        
+        if($DOC_ONLY_CHANGE -eq "true" ) {
```
nornagon: how does `DOC_ONLY_CHANGE` get set here...?

.circleci/config.yml
jkleinsc: @nornagon a non zero error code will stop CI execution

appveyor.yml
jkleinsc: doh!  It was there -- I must have deleted it.

script/doc-only-change.js
jkleinsc: @nornagon will that work for fork PRs?

jkleinsc: Also, I don't think we know in CI which branch a PR is targeting

MarshallOfSound: I've done this dance before for https://circleci.com/gh/electron/archaeologist you need the PR number to get the `base` of the PR.  We don't know if it's targeting master or another branch

appveyor.yml
```diff
@@ -50,6 +50,14 @@ build_script:
   - ps: >-
       if(($env:APPVEYOR_PULL_REQUEST_HEAD_REPO_NAME -split "/")[0] -eq ($env:APPVEYOR_REPO_NAME -split "/")[0]) {
         Write-warning "Skipping PR build for branch"; Exit-AppveyorBuild
+      } else {
+        npx.cmd yarn@1.15.2 install --frozen-lockfile
```
MarshallOfSound:
```suggestion
        node script/yarn.js install --frozen-lockfile
```

jkleinsc:
@MarshallOfSound @nornagon this is ready for re-review.  I changed the logic to check on exit code versus output.  Also #20640 shows how a doc only change PR would act with this change.

.circleci/config.yml
```diff
@@ -704,6 +704,51 @@ step-save-src-cache: &step-save-src-cache
...
     name: Persisting src cache
 
+# Check for doc only change
+step-check-for-doc-only-change: &step-check-for-doc-only-change
+  run:
+    name: Check if commit is doc only change
+    command: |
+      cd src/electron
+      node script/yarn install --frozen-lockfile
+      if node script/doc-only-change.js --prNumber=$CIRCLE_PR_NUMBER --prURL=$CIRCLE_PULL_REQUEST; then
```
nornagon: minor nit, exit code convention is that 0 means "true", i.e. this script should probably return 0 if it _is_ a doc-only change.

.circleci/config.yml
```diff
@@ -704,6 +704,51 @@ step-save-src-cache: &step-save-src-cache
...
+        #PR is not a doc only change; create empty file to indicate check has been done
+        touch .skip-ci-build
```
nornagon: my first thought seeing this file name is that its _presence_ indicates we should skip the build, rather than its content. I don't see anything that uses the presence of an empty file as an indicator, so can we use the `-e` rather than `-s`?

appveyor.yml
```diff
@@ -50,6 +50,12 @@ build_script:
   - ps: >-
       if(($env:APPVEYOR_PULL_REQUEST_HEAD_REPO_NAME -split "/")[0] -eq ($env:APPVEYOR_REPO_NAME -split "/")[0]) {
         Write-warning "Skipping PR build for branch"; Exit-AppveyorBuild
+      } else {
+        node script/yarn.js install --frozen-lockfile
+
+        if ($(node script/doc-only-change.js --prNumber=$env:APPVEYOR_PULL_REQUEST_NUMBER --prBranch=$env:APPVEYOR_REPO_BRANCH;$LASTEXITCODE)) {
```
nornagon: oh no, is the default sense of exit codes on windows reversed? I guess this could be `$LASTEXITCODE == 0`?

script/doc-only-change.js
```diff
@@ -0,0 +1,58 @@
...
+    try {
+      let pullRequestNumber = args.prNumber
+      if (!pullRequestNumber || isNaN(pullRequestNumber)) {
+        if (args.prBranch) {
+          // AppVeyor doesn't provide a PR number for branch builds - figure it out from the branch
```
nornagon: appveyor has `APPVEYOR_REPO_BRANCH` that looks to specify the base branch, can we use that?

script/doc-only-change.js
```diff
@@ -0,0 +1,58 @@
...
+          })
+          if (prsForBranch.data.length === 1) {
+            pullRequestNumber = prsForBranch.data[0].number
+          } else {
+            // If there is more than one PR on a branch, just assume that this is more than a doc change
+            process.exit(0)
+          }
+        } else if (args.prURL) {
+          // CircleCI doesn't provide the PR number for branch builds, but it does provide the PR URL
+          const pullRequestParts = args.prURL.split('/')
```
nornagon: 💡 can we perhaps directly extract the base branch from `CIRCLE_COMPARE_URL`?

.circleci/config.yml
jkleinsc: Yeah, my first thought was to look for file existence, but `step-persist-doc-only-change` fails if the file doesn't exist 😦

script/doc-only-change.js
jkleinsc: not 100% sure but I don't think that value is always set.

script/doc-only-change.js
jkleinsc: `CIRCLE_COMPARE_URL` isn't alway set, eg see: https://circleci.com/gh/electron/electron/348952

script/doc-only-change.js
jkleinsc: Its not the base branch but rather the branch that the changes are on.  See https://ci.appveyor.com/project/electron-bot/electron-ljo26/builds/28165500

.circleci/config.yml
nornagon: 😭

script/doc-only-change.js
nornagon: ugh, https://www.appveyor.com/docs/environment-variables/ says it's the **base** branch, but i guess it's lying

jkleinsc:

PR #20547, opened 2019-10-11
fix: don't hang on SendSync if ElectronBrowser receiver is destroyed (6-0-x)

loc:
#### Description of Change

If the receiver end of the `ElectronBrowser` mojo interface disconnects, we should signal the WaitableEvent to continue execution. 

We are hitting this in the very specific case of a `event.preventDefault()`-ed `nativeWindowOpen` child window with a preload script that sends a synchronous IPC. Because the nativeWindowOpen API is kinda janky, the WebContents is already created by the time we get the preventDefault from the consumer, so we do the only thing we can do: [destroy the WebContents](https://github.com/electron/electron/blob/9bd63864e487532e0adb7d36473fb72d7f83c09a/atom/browser/api/atom_api_web_contents.cc#L578). But the preload script has already started running, interestingly, first _in the opener renderer_, because site isolation may not have forced the new window into its own process yet (if it turns out to need a new process, it seems that Chrome kills the context and starts again). This is what caused the freeze we saw: the preload for the child window was running in the parent and, once the WebContents was destroyed, responsible for responding to the synchronous IPC we'd sent, hung.

This is targeting 6-0-x because the 7-0-x/master fix is slightly more complicated, given that we switched to a thread safe interface pointer which doesn't allow us to set an error handler. I'll patch in the error handler to the thread safe interface pointer in an upcoming PR. Then I'll upstream that patch, as Chrome has expressed openness to that addition.

Longer term, ideally we would allow the user to opt-out of the window creation earlier, to avoid creating a WebContents at all. It's very strange that the preload starts running twice, once in the renderer and once in the new process (if applicable). Changing this should be taken on as part of a broader effort to improve the native window open API.

#### Checklist

#### Release Notes

Notes: Fix IPC hang when opening a specific case of `nativeWindowOpen` child window

### DIFF
--- a/atom/renderer/api/atom_api_renderer_ipc.cc
+++ b/atom/renderer/api/atom_api_renderer_ipc.cc
@@ -99,6 +99,7 @@ class IPCRenderer : public mate::Wrappable<IPCRenderer> {
     auto interface_info = electron_browser_ptr_.PassInterface();
     task_runner->PostTask(
         FROM_HERE, base::BindOnce(&IPCRenderer::SendMessageSyncOnWorkerThread,
+                                  base::Unretained(this),
                                   base::Unretained(&interface_info),
                                   base::Unretained(&response_received_event),
                                   base::Unretained(&result), internal, channel,
@@ -109,7 +110,7 @@ class IPCRenderer : public mate::Wrappable<IPCRenderer> {
   }
 
  private:
-  static void SendMessageSyncOnWorkerThread(
+  void SendMessageSyncOnWorkerThread(
       atom::mojom::ElectronBrowserPtrInfo* interface_info,
       base::WaitableEvent* event,
       base::Value* result,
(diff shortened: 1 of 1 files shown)


atom/renderer/api/atom_api_renderer_ipc.cc
```diff
@@ -109,32 +110,50 @@ class IPCRenderer : public mate::Wrappable<IPCRenderer> {
...
       bool internal,
       const std::string& channel,
       const base::ListValue* arguments) {
     atom::mojom::ElectronBrowserPtr browser_ptr(std::move(*interface_info));
-    browser_ptr->MessageSync(
+    worker_thread_electron_browser_ptr_ = std::move(browser_ptr);
+
+    worker_thread_electron_browser_ptr_.set_connection_error_handler(
+        base::BindOnce(&IPCRenderer::HandleMojoConnectionErrorOnWorkerThread,
+                       base::Unretained(&worker_thread_electron_browser_ptr_),
```
nornagon: it might be simpler to pass `base::Unretained(this)` and have the method be non-static? that way we don't have to pass all these member variables around

atom/renderer/api/atom_api_renderer_ipc.cc
```diff
@@ -109,32 +110,50 @@ class IPCRenderer : public mate::Wrappable<IPCRenderer> {
...
+    *interface_info = ptr->PassInterface();
+    event->Signal();
+  }
+  // If the other end of the message pipe disconnects, ensure we don't hang the
+  // main thread forever.
+  static void HandleMojoConnectionErrorOnWorkerThread(
+      atom::mojom::ElectronBrowserPtr* ptr,
+      atom::mojom::ElectronBrowserPtrInfo* interface_info,
+      base::WaitableEvent* event) {
+    *interface_info = ptr->PassInterface();
```
nornagon: We should probably record somewhere that an error occurred, so we can throw instead of returning an empty value.

atom/renderer/api/atom_api_renderer_ipc.cc
loc: Yep, could do; was just thinking I'd match the style of the surrounding code, which passes things around. It of course changes in 7-0-x, so this might not be worth debating.

atom/renderer/api/atom_api_renderer_ipc.cc
nornagon: hm... what does it mean to return the interface once the connection is severed? perhaps we should leave the interface empty in this case? also we should make sure not to attempt to send more messages across the interface once the first failure has been detected.

miniak:
related to #20632

atom/renderer/api/atom_api_renderer_ipc.cc
loc: Intuitively, it makes sense to PassInterface — that's the unbind operation. Changing the question slightly but with similar intent: should one be allowed to re-bind after an error? This change represents no regression on that front, as now we're simply aware of errors, with no effect on how we or Mojo responds to them. I agree, it would be a good thing to consider, but it feels reasonably out of scope for this PR, especially since this change is targeted at stale 6-0-x.

Issue #20442, opened 2019-10-05
Porting electron-api-demos to Electron Fiddle examples in docs


erickzhao:
## Context

This meta-issue addresses a need to bridge the gap between our website's online docs, `electron-api-demos`, and Electron Fiddle.

- <details><summary>What's <code>electron-api-demos</code>?</summary>
   <p>
   As the name indicates, <code>electron-api-demos`</code> is a collection of examples for Electron APIs wrapped in an application. Each demo contains a executable example of the API along with sample code and additional explanations.
    </p>
    </details>

- <details><summary>What's Electron Fiddle?</summary>
   <p>
   Electron Fiddle is a code playground allowing developers to tinker around with small experiments that can later be exported into full-fledged Electron applications. The interface allows users to create and run small Electron apps with a single <code>main.js</code>, <code>renderer.js</code>, and <code>index.html</code>.
   </p>
   </details>
- <details><summary>Motivation</summary>
   <ul>
   <li>Although collecting runnable API demos is nice, having the sample code gated behind an application increases friction, and maintaining the app in its current state for more than 1 release line is difficult.</li>
    <li> Having the code available online makes it easier to find and more accessible to read, but having to manually launch </li>
    <li> <strong>Porting over <code>electron-api-demos</code> into Fiddle-friendly examples is a big step into ultimately having API code samples available in the docs as runnable Fiddles.</strong> </li> </details>
- <details>
   <summary>Current Progress</summary>
   <ul><li>Recently, Electron Fiddle added the capability of loading code directly from the docs on the <code>electron/electron repo</code> (see electron/fiddle/pull/203). However, there are currently no examples!</li>
     <li>The ElectronJS website currently has an open PR (electron/electronjs.org/pull/2848]) adding the ability to launch the Fiddle protocol from the Electron website.</li>
</details>
   </ul>

## Task

We will be attempting to port over all of the `electron-api-demos` examples in the codebase. For context on how API demos are structured, see the docs [here](https://github.com/electron/electron-api-demos/blob/master/docs.md). 

The task is to ultimately convert each demo into an Electron Fiddle sample that's equivalent in functionality. **Please include the explanatory text from the HTML files as well.** Note that many of these demos are tightly coupled in logic, and a single PR can be submitted for multiple demos within a Section.

[...]

PR #20480, opened 2019-10-08
docs: New fiddle example 'Create a new window'

wedob13:
#### Description of Change
Refs #20442

Adds the Create a new window example from electron-api-demos into a runnable/executable Fiddle example.

Gist link to Fiddle (same as code submitted in this PR): https://gist.github.com/7c0d457d67ffbeff02107303fb2127ee

#### Checklist
- [x] PR description included and stakeholders cc'd
- [x] Lint test case passed
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes]

#### Release Notes

notes: Adds the "Create a new window" fiddle example for Manage Windows section

PR #20571, opened 2019-10-14
docs: New fiddle example 'Window Crash'

ValerianClerc:
#### Description of Change
Added a fiddle example for window crash handling. Based off of code from [electron-api-demos](https://github.com/electron/electron-api-demos)

For: #20442 

#### Checklist

#### Release Notes

Notes: Added "Relaunch window after process crashes" fiddle example to docs

### DIFF
--- a/docs/fiddles/windows/crashes-and-hangs/crash/index.html
+++ b/docs/fiddles/windows/crashes-and-hangs/crash/index.html
@@ -1,0 +1,26 @@
+<!DOCTYPE html>
+<html><head><style>
+  body {
+    padding: 20px;
+    font-family: system, -apple-system, '.SFNSText-Regular', 'SF UI Text', 'Lucida Grande', 'Segoe UI', Ubuntu, Cantarell, sans-serif;
+    color: #fff;
+    background-color: #8aba87;
+    text-align: center;
+    font-size: 40px;
+  }
+
+  #crash {
+    color: white;
+    opacity: 0.7;
+    position: absolute;
+    bottom: 20px;
+    left: 50%;
+    transform: translateX(-50%);
+    font-size: 12px;
+    text-decoration: none;
+  }
+</style>
+
+</head><body><p>Click the text below to crash and then reload this process.</p>
+<a id="crash" href="javascript:process.crash()" >Crash this process</a>
+</body></html>
--- a/docs/fiddles/windows/crashes-and-hangs/crash/index.html
+++ b/docs/fiddles/windows/crashes-and-hangs/crash/index.html
@@ -1,26 +1,8 @@
 <!DOCTYPE html>
-<html><head><style>
-  body {
-    padding: 20px;
-    font-family: system, -apple-system, '.SFNSText-Regular', 'SF UI Text', 'Lucida Grande', 'Segoe UI', Ubuntu, Cantarell, sans-serif;
-    color: #fff;
-    background-color: #8aba87;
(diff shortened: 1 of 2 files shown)


docs/fiddles/windows/crashes-and-hangs/crash/index.html
```diff
@@ -0,0 +1,26 @@
...
+    bottom: 20px;
+    left: 50%;
+    transform: translateX(-50%);
+    font-size: 12px;
+    text-decoration: none;
+  }
+</style>
+
+</head><body><p>Click the text below to crash and then reload this process.</p>
+<a id="crash" href="javascript:process.crash()" >Crash this process</a>
```
felixrieseberg:
```suggestion
<a id="crash" href="javascript:process.crash()">Crash this process</a>
```

docs/fiddles/windows/crashes-and-hangs/crash/index.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<html><head><style>
+  body {
```
felixrieseberg: Can you do us a favor and cut out the `style`? Thank you!

ValerianClerc:
> Thank you very much! We just need to cut the CSS, it's otherwise good to go.

Alright all done! @felixrieseberg 

Also I noticed that the `appveyor: win-ia32-testing-pr` build failed, do you know what that's about? It's saying that tests failed, but I didn't write any tests or change any tested code.

Thanks!

erickzhao:
@ValerianClerc CI logs seem to show that the tests are failing: `Electron tests failed with code 0xc0000374.`

As this is a documentation-only PR and doesn't modify Electron internals, I wouldn't worry about it. Tests flake fairly often!

zcbenz:
@ValerianClerc Do you still plan to work on this PR per @erickzhao's review?

ValerianClerc:
Hi @zcbenz, thank you for the reminder, I'll keep working on this one and wrap it up.

@erickzhao Thanks for the comments and sorry for the delay, I promisified the dialog MessageBox to fit the newer APIs! (tested on 7.1.3)

I tried to do the crash recovery with `mainWindow.reload()`/`mainWindow.close()` initially, because that's what they do in the `electron-api-demos` app. But I get the same stack trace crash as you.

Could you point me in the direction to tackle this? My thoughts so far:
1. From my understanding, in `electron-api-demos`, the code is being executed as a child window to the main app, in the `renderer.js`. However, from your comments on the `createWindow` PR #20480, I understand that `electron-fiddle` doesn't support loading multiple HTML windows? So I can't make a child window with a button to crash itself, since I already have an HTML file loaded for the parent window? And when the parent window (app) crashes, I have to reload the entire app?
2. Could also just be that `electron-api-demos` is outdated

zcbenz:
@ValerianClerc I'm jumping into the discussion.

The usual way of recovering crash is to close the window and create a new one, which is used by apps like Slack. Reloading a crashed window might not work.

I'm not sure why `mainWindow.close()` crashes, but we should use the right approach here and fix the crash instead. If the crash still happens can you create an issue for it?

> 1. From my understanding, in `electron-api-demos`, the code is being executed as a child window to the main app, in the `renderer.js`

The fiddle examples are executed as independent apps, so it is fine create multiple windows in this app.

codebytere:
Closing as abandoned.

PR #20591, opened 2019-10-15
docs: added windows -> create frameless window example

kristofkalocsai:
#### Description of Change
This adds the _Create a frameless window_ example from [electron-api-demos](https://github.com/electron/electron-api-demos/blob/master/sections/windows/windows.html) to a Fiddle

#### Checklist

- [x] PR is merging into master branch.
- [x] File structure adheres to the requirements above.
- [x] Code in PR runs successfully in Electron Fiddle with the latest release of Electron 6.
- [x] Code in PR passes a linting check with Standard.

#### Release Notes

Notes: Replaced modal HTML file from the original example with an inline [DataURL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) with the same content.


erickzhao:
Refs #20442

PR #20598, opened 2019-10-15
chore: remove native_mate (Part 8)

zcbenz:
#### Description of Change

This PR converts the methods of `auto_updater`, `crash_reporter`, `native_image` to `gin`.

#### Release Notes

Notes: no-notes


shell/browser/auto_updater.h
```diff
@@ -55,7 +58,10 @@ class AutoUpdater {
   static void SetDelegate(Delegate* delegate);
 
   static std::string GetFeedURL();
-  static void SetFeedURL(mate::Arguments* args);
+  // FIXME(zcbenz): We should not do V8 in this file, this method should only
+  // accept C++ struct as parameter, and atom_api_auto_updater.cc is responsible
+  // for parsing the parameter from JavaScript.
+  static void SetFeedURL(gin_helper::Arguments* args);
```
deepak1556: Since this is `static` the refactor wouldn't be much and would be good to do it in this PR ?

shell/common/crash_reporter/crash_reporter.h
```diff
@@ -25,7 +28,10 @@ class CrashReporter {
   typedef std::pair<int, std::string> UploadReportResult;  // upload-date, id
 
   static CrashReporter* GetInstance();
-  static void StartInstance(const mate::Dictionary& options);
+  // FIXME(zcbenz): We should not do V8 in this file, this method should only
+  // accept C++ struct as parameter, and atom_api_crash_reporter.cc is
+  // responsible for parsing the parameter from JavaScript.
+  static void StartInstance(const gin_helper::Dictionary& options);
```
deepak1556: Same here.

shell/browser/auto_updater.h
zcbenz: I would like to keep this PR focused on native_mate related changes.

Issue #20599, opened 2019-10-16
systemPreferences.getColors() throws exceptions with some macOS colors


yourhead:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 
v7.0.0-beta.6
* **Operating System:**
  * 
macOS 10.14.6
* **Last Known Working Electron version:**
  * 
the bug exists in all recent betas

### Expected Behavior

a color value as a string
no exception thrown

### Actual Behavior

exception thrown
```
"Error: Unknown color: disabled-control-text
    at Function.updateAppearanceTheme (/Users/iac/Desktop/gluon/dist/App.js:138:54)
    at App.createWindow (/Users/iac/Desktop/gluon/dist/App.js:9:13)
    at App.emit (events.js:208:15)"
```

### To Reproduce

```
systemPreferences.getColor('disabled-control-text');
systemPreferences.getColor('scrubber-textured-background');
systemPreferences.getColor('selected-menu-item');
```

For reference, here are the macOS headers for these colors and their availability in the macOS 10.14 SDK

```
@property (class, strong, readonly) NSColor *disabledControlTextColor;                     // Text on disabled controls

@property (class, strong, readonly) NSColor *scrubberTexturedBackgroundColor NS_AVAILABLE_MAC(10_12_2); // Patterned background color for use in NSScrubber

@property (class, strong, readonly) NSColor *selectedMenuItemColor NS_DEPRECATED_MAC(10_0, API_TO_BE_DEPRECATED, "Use NSVisualEffectMaterialSelection");
```

All three are available in macOS 10.14 (one is deprecated, but still available).

But even if they weren't, the Electron docs don't specify limitations on the macOS version for this function. Shouldn't I expect *some* value rather than throwing an exception even when colors are not available on the user's system.

Or perhaps the documentation should reflect that in some cases the function will throw an exception -- and what those cases are.

A fiddle for such basic stuff seems overkill, but it also took me about 1 minute -- so why not. 😃 

https://gist.github.com/yourhead/1db43ba064651648c673d686775b7afc

MarshallOfSound:
[...]

PR #20611, opened 2019-10-16
fix: macOS getColor inconsistencies

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/20599.

Updates documentation to reflect macOS 10.14 specific colors and fixes a mistyped color. Also notes a deprecated color.

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes
Notes: Fixes some `systemPreferences.getColor()` inconsistencies on macOS.

### DIFF
--- a/shell/browser/api/atom_api_system_preferences_mac.mm
+++ b/shell/browser/api/atom_api_system_preferences_mac.mm
@@ -26,6 +26,7 @@
 #include "shell/browser/mac/atom_application.h"
 #include "shell/browser/mac/dict_util.h"
 #include "shell/browser/ui/cocoa/NSColor+Hex.h"
+#include "shell/common/deprecate_util.h"
 #include "shell/common/native_mate_converters/gurl_converter.h"
 #include "shell/common/native_mate_converters/value_converter.h"
 #include "ui/native_theme/native_theme.h"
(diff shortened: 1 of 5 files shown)


shell/browser/api/atom_api_system_preferences_mac.mm
```diff
@@ -503,19 +503,23 @@ AVMediaType ParseMediaType(const std::string& media_type) {
...
+    EmitDeprecationWarning(env, "Use selected-content-background instead.",
+                           "electron");
   } else if (color == "control-background") {
     sysColor = [NSColor controlBackgroundColor];
   } else if (color == "control") {
     sysColor = [NSColor controlColor];
   } else if (color == "control-text") {
     sysColor = [NSColor controlTextColor];
-  } else if (color == "disabled-control") {
+  } else if (color == "disabled-control-text") {
```
MarshallOfSound: How long has this been wrong? Could someone have figured out the wrong value and should we specifically support the old value with a deprecation warning similar to above to be nicer to those folks?

codebytere: (resolved offline)

shell/browser/api/atom_api_system_preferences_mac.mm
```diff
@@ -503,19 +504,25 @@ AVMediaType ParseMediaType(const std::string& media_type) {
...
 
-std::string SystemPreferences::GetColor(const std::string& color,
-                                        mate::Arguments* args) {
+std::string SystemPreferences::GetColor(gin_helper::ErrorThrower thrower,
+                                        const std::string& color) {
   NSColor* sysColor = nil;
   if (color == "alternate-selected-control-text") {
     sysColor = [NSColor alternateSelectedControlTextColor];
+    EmitDeprecationWarning(node::Environment::GetCurrent(thrower.isolate()),
+                           "Use selected-content-background instead.",
```
MarshallOfSound:
```suggestion
                           "'alternate-selected-control-text' is deprecated as an input to getColor.  Use 'selected-content-background' instead.",
```

shell/browser/api/atom_api_system_preferences_mac.mm
```diff
@@ -503,19 +504,25 @@ AVMediaType ParseMediaType(const std::string& media_type) {
...
     sysColor = [NSColor controlBackgroundColor];
   } else if (color == "control") {
     sysColor = [NSColor controlColor];
   } else if (color == "control-text") {
     sysColor = [NSColor controlTextColor];
-  } else if (color == "disabled-control") {
+  } else if (color == "disabled-control-text") {
     sysColor = [NSColor disabledControlTextColor];
+    EmitDeprecationWarning(node::Environment::GetCurrent(thrower.isolate()),
+                           "Use a color that matches the semantics being used.",
```
MarshallOfSound: Not sure what this is for, if I don't understand the warning users won't either 😄

codebytere: Looks like it's not deprecated anymore? i'm very confused but rolling with it. Removed.

Issue #20503, opened 2019-10-09
Regression of recentDocuments file list on macOS


bigtimebuddy:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 6.0.12
* **Operating System:**
  * macOS 10.14.6
* **Last Known Working Electron version:**
  * 5.0.11

### Expected Behavior

Using `app.addRecentDocument`, should add files to the `recentDocuments` menu.

### Actual Behavior

The files are missing from the `recentDocuments` menu (the files, however, show up correctly when right-clicking on the dock icon--assuming that `addRecentDocument` works fine).

* `recentDocuments` seems like it is still broken in 6.0.12. The menu items work (are now enabled) but the `recentDocuments` menu does not populate with files.

### To Reproduce

```sh
$ git clone git@github.com:bigtimebuddy/electron-recent-docs-bug.git -b master
$ npm install
$ npm start || electron .
```

### Screenshots

**v6.0.12**
![Screen Shot 2019-10-08 at 4 27 44 PM](https://user-images.githubusercontent.com/864393/66440652-9996eb80-e9e8-11e9-9517-b3651677b207.png)

**v5.0.11**
![Screen Shot 2019-10-08 at 4 27 19 PM](https://user-images.githubusercontent.com/864393/66440648-969bfb00-e9e8-11e9-8d3d-7422cfdf08b5.png)

### Additional Information

Follow up to #20364 & #20398

cc @codebytere

bigtimebuddy:
I'm currently still being blocked by this regression.

codebytere:
It's on my backlog, but it's a fairly complex issue. We used to use a xib file but then removed it unaware that it would cause this. Chromium has removed their use of xib files so we can't trivially add it back and will now need to manually implement the backing logic. I'll do my best in the next few weeks :)

bigtimebuddy:
Thank you for the update, makes sense. I appreciate your help.

codebytere:
[...]

PR #20615, opened 2019-10-16
fix: use xib file to construct macOS Menu

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/20503.

When we removed the `.xib` file menu construction, it unfortunately nuked the `recentDocuments` role, which unexpectedly is near-impossible to replicate manually with obj-c without doing terrifying hacky things or using private APIs (see [this blog post](http://lapcatsoftware.com/blog/2007/07/10/working-without-a-nib-part-5-open-recent-menu/)). As a result, our only real path forward here was to revert to xib-based functionality removed originally removed in [this commit](https://github.com/electron/electron/pull/18626/commits/44ee1f4de23fe634a794cc3bddf3a4acc0624809). I needed to re-implement the xib template in GN removed upstream but otherwise I mostly just needed to re-add the xib file and revert to the older `recentDocuments` menu role business logic in the `menu_controller`.

Tested to work with [this repro case](http://github.com/bigtimebuddy/electron-recent-docs-bug).

cc @MarshallOfSound @nornagon (& @bigtimebuddy!)

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a regression in the `recentDocuments` role on macOS.

### DIFF
--- a/build/rules.gni
+++ b/build/rules.gni
@@ -5,6 +5,8 @@ import("//build/config/mac/mac_sdk.gni")
 # See https://chromium-review.googlesource.com/c/chromium/src/+/1648695
 import("//build/config/ios/rules.gni")
 
+# Template is copied here from Chromium but was removed in
+# https://chromium-review.googlesource.com/c/chromium/src/+/1637981
 # Template to compile and package Mac XIB files as bundle data.
 # Arguments
 #     sources:
--- a/build/rules.gni
+++ b/build/rules.gni
@@ -1,0 +1,55 @@
+import("//build/config/mac/mac_sdk.gni")
+
+# This is imported from /ios becuase this functionality was moved
+# after Chromium stopped using xib files for macOS menu functionality
+# See https://chromium-review.googlesource.com/c/chromium/src/+/1648695
+import("//build/config/ios/rules.gni")
+
+# Template to compile and package Mac XIB files as bundle data.
+# Arguments
+#     sources:
+#         list of string, sources to comiple
+#     output_path:
+#         (optional) string, the path to use for the outputs list in the
+#         bundle_data step. If unspecified, defaults to bundle_resources_dir.
+template("mac_xib_bundle_data") {
+  _target_name = target_name
+  _compile_target_name = _target_name + "_compile_ibtool"
+
+  compile_ib_files(_compile_target_name) {
+    forward_variables_from(invoker, [ "testonly" ])
+    visibility = [ ":$_target_name" ]
+    sources = invoker.sources
+    output_extension = "nib"
+    ibtool_flags = [
+      "--minimum-deployment-target",
+      mac_deployment_target,
+
+      # TODO(rsesek): Enable this once all the bots are on Xcode 7+.
+      # "--target-device",
+      # "mac",
+    ]
+  }
+
+  bundle_data(_target_name) {
+    forward_variables_from(invoker,
+                           [
+                             "testonly",
+                             "visibility",
+                           ])
+
+    public_deps = [
+      ":$_compile_target_name",
+    ]
+    sources = get_target_outputs(":$_compile_target_name")
+
+    _output_path = "{{bundle_resources_dir}}"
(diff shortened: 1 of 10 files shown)


deepak1556:
Not a review but some thoughts , the reason chrome is doing this refactor is they are trying to remove dependency on system XCode and make builds deterministic https://bugs.chromium.org/p/chromium/issues/detail?id=330262, which again aligns with one of our upcoming projects from security WG , making electron builds deterministic. 

Once this transition is complete on chrome end and we still deviate from it then its going to be a technical debt similar to libc++ one which I would like to avoid at all cost. So reiterating on this PR, I would suggest to look into the programmatic implementation or revisit the position of this api through the API WG.

nornagon:
Thanks for the context @deepak1556! Since this is just a dependency on `ibtool`, not a dependency on any compiler/linker tool from Xcode, do you think it would be OK? A quick test on my machine produced identical output from multiple runs of `ibtool` on identical inputs, showing that it's at least timestamp-independent. I also tested running in a different directory (different pwd, input and output paths) and got identical bytes out of `ibtool`.

In general I agree with wanting to drop Xcode as a dependency, and aligning with Chromium, but this doesn't seem like it would impact our ability to produce deterministic builds. Additionally, doing this programmatically requires calling private APIs which may or may not be rejected by MAS.

codebytere:
> Additionally, doing this programmatically requires calling private APIs which may or may not be rejected by MAS.

This is also my primary consideration here; i'd rather have the system handle this where possible and not endanger users' ability to ship via the mac app store.

At the very least, i'd like to get this into versions where it's supposed to work and then perhaps consider deprecating it in future versions when Chromium does totally drop Xcode 🤔

deepak1556:
Sorry for the delayed response, comments inline.

> Since this is just a dependency on ibtool, not a dependency on any compiler/linker tool from Xcode, do you think it would be OK?

Yeah dependency on ibtool is fine, it used to be non-deterministic in the past. I was originally worried about a situation when chromium gets rid of all its xib files, they would start shipping the minimal set of binaries and SDK for the build. We would have to rely on the system installation for this one, but since it doesn't change the determinism of the build, should be ok! 


> Additionally, doing this programmatically requires calling private APIs which may or may not be rejected by MAS.

Moving forward, implementations that require this should be considered by the API WG and see if it can be made as a userland module or expose the right parts to make it possible.

codebytere:
> Moving forward, implementations that require this should be considered by the API WG and see if it can be made as a userland module or expose the right parts to make it possible.

@deepak1556 do you wish for the API group to vote on _this_ PR, or are you comfortable with this PR as it stands and wish for us to vote on _future_ PRs affecting similar build components?

deepak1556:
@codebytere sorry my comment wasn't clear, I was talking about future PR's . This one is good as it stands.

build/rules.gni
```diff
@@ -0,0 +1,55 @@
+import("//build/config/mac/mac_sdk.gni")
+
+# This is imported from /ios becuase this functionality was moved
+# after Chromium stopped using xib files for macOS menu functionality
+# See https://chromium-review.googlesource.com/c/chromium/src/+/1648695
+import("//build/config/ios/rules.gni")
+
+# Template to compile and package Mac XIB files as bundle data.
```
nornagon: Add a note that this was copied from chromium & link the CL in which it was removed

shell/browser/resources/mac/Info.plist
```diff
@@ -24,6 +24,8 @@
     <string>public.app-category.developer-tools</string>
     <key>LSMinimumSystemVersion</key>
     <string>10.10.0</string>
+    <key>NSMainNibFile</key>
+	  <string>MainMenu</string>
```
nornagon: indentation

bigtimebuddy:
Could someone manually back-port this to 6.1.x? Doesn't look like it showed up in the latest release.

PR #20620, opened 2019-10-17
build: add support for automated minor releases

jkleinsc:
#### Description of Change

Starting with the 8-x series we are going to switch to having an 8-x-y branch from which we can publish 8-0-x releases as well as (if necessary) 8-x-y releases.  This means that if we want to release a minor version we will no longer have to create a new branch for that minor.  

This PR changes our release scripts to allow for minor bumps.  There will need to be corresponding changes made to sudowoodo and trop to support this change.
#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/release/prepare-release.js
+++ b/script/release/prepare-release.js
@@ -23,7 +23,7 @@ const pass = '✓'.green
 const fail = '✗'.red
 
 if (!bumpType && !args.notesOnly) {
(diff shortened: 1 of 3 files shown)


jkleinsc:
@MarshallOfSound good point. I forgot about those tests. I added a test for the minor bump.

jkleinsc:
@MarshallOfSound this is ready for re-review

PR #20389, opened 2019-10-01
feat: allow inspection of specific shared workers

miniak:
#### Description of Change
Added APIs to allow electron to get a list of shared web workers and then inspect 
a shared worker based on its ID.

#### Checklist
- [X] PR description included and stakeholders cc'd
- [X] `npm test` passes
- [X] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [X] relevant documentation is changed or added
- [X] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes
Notes: Added ability to inspect specific shared workers: `contents.getAllSharedWorkers()` / `contents.inspectSharedWorkerById(workerId)`.

PR #20624, opened 2019-10-17
feat: add API for receiving logs from service workers

MarshallOfSound:
Pretty simple API here, plans are for the following

* `console-message` event.  Vaguely equivalent to the `console-message` event on the `webContents` API so that apps can capture logs from their service workers.
* `getWorkerInfoFromID()` Get service worker scope and script url for a given ID
* `getRunningWorkerInfos()` Get service worker scope, script url and id for all active workers

The second two APIs are blocked on a Chromium roll hence `WIP`, the event is already implemented.

TODO:
* [x] `console-message`
* [x] `getAll()`
* [x] `getFromVersionId()`
* [x] Docs
* [ ] Tests

Notes: Added `session.serviceWorkerContext` API to access basic service worker info and receive console logs from service workers.


miniak:
@MarshallOfSound Is `getRunningWorkerInfos()` a replacement for `webContents.getAllSharedWorkers()` added in #20389?

miniak:
@MarshallOfSound should we also deprecate `webContents.inspectServiceWorker()` and `webContents.inspectSharedWorker()` and replace it with a new API that supports any type of worker?

MarshallOfSound:
@miniak `SharedWorker` !== `ServiceWorker` afaik.  This APIs primary purpose is getting logs from service workers.  If we feel it's appropriate for this to work for _all_ workers that is substantially more complicated as service workers, shared workers and web workers are each very different things in Chrome

MarshallOfSound:
Removing WIP as this is ready for code review.  I'll work on tests shortly 👍

shell/browser/api/atom_api_service_worker_context.cc
```diff
@@ -0,0 +1,142 @@
...
+#include "shell/common/gin_helper/object_template_builder.h"
+#include "shell/common/node_includes.h"
+
+namespace electron {
+
+namespace api {
+
+namespace {
+
+std::string MessageSourceToString(
```
nornagon: gin converter?

shell/browser/api/atom_api_service_worker_context.cc
```diff
@@ -0,0 +1,142 @@
...
+    return "recommendation";
+  return "other";
+}
+
+base::DictionaryValue ServiceWorkerRunningInfoToDict(
+    const content::ServiceWorkerRunningInfo& info) {
+  base::DictionaryValue dict;
+  dict.SetStringPath("scriptUrl", info.script_url.spec());
+  dict.SetStringPath("scope", info.scope.spec());
+  dict.SetIntPath("renderProcessId", info.render_process_id);
```
nornagon: prefer `gin::Dictionary` unless there's a particular reason to go via `base::Value`. Even better, use `gin::DataObjectBuilder`.

shell/browser/api/atom_api_service_worker_context.cc
```diff
@@ -0,0 +1,142 @@
...
+
+}  // namespace
+
+ServiceWorkerContext::ServiceWorkerContext(v8::Isolate* isolate,
+                                           AtomBrowserContext* browser_context)
+    : browser_context_(browser_context), weak_ptr_factory_(this) {
+  Init(isolate);
+  service_worker_context_ =
+      content::BrowserContext::GetDefaultStoragePartition(browser_context_)
+          ->GetServiceWorkerContext();
```
nornagon: I don't think this object is guaranteed to have a shorter lifetime than the `ServiceWorkerContext` to which it refers, since it's retained by V8.

shell/browser/api/atom_api_service_worker_context.cc
```diff
@@ -0,0 +1,142 @@
...
+void ServiceWorkerContext::OnReportConsoleMessage(
+    int64_t version_id,
+    const content::ConsoleMessage& message) {
+  base::DictionaryValue details;
+  details.SetDoublePath("versionId", static_cast<double>(version_id));
+  details.SetStringPath("source", MessageSourceToString(message.source));
+  details.SetIntPath("level", static_cast<int32_t>(message.message_level));
+  details.SetStringPath("message", message.message);
+  details.SetIntPath("lineNumber", message.line_number);
+  details.SetStringPath("sourceUrl", message.source_url.spec());
```
nornagon: `DataObjectBuilder` here too

shell/browser/api/atom_api_service_worker_context.cc
```diff
@@ -0,0 +1,142 @@
...
+                            new ServiceWorkerContext(isolate, browser_context));
+}
+
+// static
+void ServiceWorkerContext::BuildPrototype(
+    v8::Isolate* isolate,
+    v8::Local<v8::FunctionTemplate> prototype) {
+  prototype->SetClassName(mate::StringToV8(isolate, "ServiceWorkerContext"));
+  gin_helper::ObjectTemplateBuilder(isolate, prototype->PrototypeTemplate())
+      .SetMethod("getAll", &ServiceWorkerContext::GetAllWorkerInfo)
```
nornagon: `getAllRunning` perhaps?

shell/browser/api/atom_api_service_worker_context.cc
```diff
@@ -0,0 +1,142 @@
...
+
+ServiceWorkerContext::~ServiceWorkerContext() {
+  service_worker_context_->RemoveObserver(this);
+}
+
+void ServiceWorkerContext::OnReportConsoleMessage(
+    int64_t version_id,
+    const content::ConsoleMessage& message) {
+  base::DictionaryValue details;
+  details.SetDoublePath("versionId", static_cast<double>(version_id));
```
nornagon: `int64_t` is bigger than `double` can hold. Use a `BigInt` or a string.

docs/api/service-workers.md
```diff
@@ -0,0 +1,62 @@
...
+### Instance Events
+
+The following events are available on instances of `ServiceWorkers`:
+
+#### Event: 'console-message'
+
+Returns:
+
+* `event` Event
+* `messageDetails` Object - Information about the console message
```
nornagon: this sort of makes me wonder why we even have an `Event` object

docs/api/service-workers.md
```diff
@@ -0,0 +1,62 @@
...
+  * `sourceUrl` String - The URL the message came from
+  * `lineNumber` Number - The line number of the source that triggered this console message
+
+Emitted when a service worker logs something to the console.
+
+### Instance Methods
+
+The following methods are available on instances of `ServiceWorkers`:
+
+#### `serviceWorkers.getAllRunning()`
```
nornagon: Would it be reasonable to add a `getAllRegistered` method here? Not necessarily in this PR, but as a followup?

docs/api/structures/service-worker-info.md
```diff
@@ -0,0 +1,5 @@
+# ServiceWorkerInfo Object
+
+* `scriptUrl` String - The full URL to the script that this service worker runs
+* `scope` String - The base URL that this service worker is active for.
+* `renderProcessId` Number - The virtual ID of the process that this service worker is running in.  This is not an OS level PID.  This aligns with the ID set used for `webContents.getProcessId()`.
```
nornagon: If we wanted to add e.g. methods for killing / restarting service workers, would they go on this object? e.g.

```
session.defaultSession.serviceWorkers.getAllRunning().forEach(sw => sw.shutdown())
```

or something else?

spec-main/api-service-workers-spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  describe('getAllRunning()', () => {
+    it('should initially report none are running', () => {
+      expect(ses.serviceWorkers.getAllRunning()).to.deep.equal({})
+    })
+
+    it('should report one as running once you load a page with a service worker', async () => {
+      await emittedOnce(ses.serviceWorkers, 'console-message', () => w.loadURL(`${baseUrl}/index.html`))
+      const workers = ses.serviceWorkers.getAllRunning()
+      const ids = Object.keys(workers) as any[] as number[]
```
nornagon:
```suggestion
      const ids = Object.keys(workers)
```

spec-main/api-service-workers-spec.ts
```diff
@@ -0,0 +1,99 @@
...
+      expect(ids).to.have.lengthOf(1, 'should have one worker running')
+    })
+  })
+
+  describe('getFromVersionID()', () => {
+    it('should report the correct sciprt url and scope', async () => {
```
nornagon:
```suggestion
    it('should report the correct script url and scope', async () => {
```

docs/api/service-workers.md
MarshallOfSound: I think that's a large api-wg discussion about why we have this `event` object for everything

docs/api/service-workers.md
MarshallOfSound: If there's a Chromium API for it, yeah, can do 👍

docs/api/structures/service-worker-info.md
MarshallOfSound: Either that, or have `serviceWorkers.stopByVersionId(versionid)`

Issue #20557, opened 2019-10-13
SharedWorker broken in 7.0 beta


krisdages:
### Preflight Checklist

* [X ] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [X ] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [X ] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details
This is a bug report specifically for the 7.0 beta. 

* **Electron Version:**
  * 7.0.0-beta.6
* **Operating System:**
  * Ubuntu 18.04 x64, Windows 10
* **Last Known Working Electron version:**
  * 6.0.12

### Expected Behavior
SharedWorker should be functional and *shared* when using
  a custom protocol to provide a proper origin.

### Actual Behavior
SharedWorker does not execute when loaded from custom protocol.
It does execute when page is loaded with the `file://` protocol, but this 
  defeats the purpose, since it is not shared between contexts.

Also noticed while diagnosing the issue:
Creating a SharedWorker with node integration enabled using a Data URL crashes 
the renderer.  I verified that this works in Electron 6, figured the cause might be related.

### To Reproduce
Install and run the following repo:

https://github.com/krisdages/electron-7-shared-worker-broken

$ git clone https://github.com/krisdages/electron-7-shared-worker-broken
$ npm install
$ npm start

Sorry, had it implemented before seeing the "electron fiddle" option. May try that next time.

Thanks for looking into this :)

PR #20625, opened 2019-10-18
fix: loading dedicated/shared worker scripts over custom protocol

deepak1556:
#### Description of Change

Closes https://github.com/electron/electron/issues/20557.
Closes https://github.com/electron/electron/issues/24715.

This is a two part fix,

**Part I: Fix crash when loading over data: protocol mentioned in the above issue**

This issue will appear only with nodeintegration enabled in the worker. Adds patch `worker_feat_add_hook_to_notify_script_ready.patch` that explains the root cause of this issue, I am gonna try upstreaming this, will see how it goes. The access to global properties from Node.js happens in `node::MaybeInitializeContext` which modifies the global object for primordials.

**Part II: Cannot load worker scripts over custom protocol**

After https://chromium-review.googlesource.com/c/chromium/src/+/1798250 we need to explicitly register custom URLLoaderFactories for dedicated/shared workers.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:
* fix loading shared worker scripts over custom protocol
* fix crash when loading worker scripts with nodeIntegration enabled


deepak1556:
`WIP` to add some tests.

Dan-Shields:
Wouldn't know where to begin with writing the tests, but would love to see this merged, especially as v6 is now not supported and was the last version where shared workers worked!

krisdages:
I thought this had been ready to go for a while and just needed to be merged... I assumed there were tests in the last commit since it was approved...

codebytere:
macOS failures are: https://github.com/electron/electron/pull/24748. Will merge when woa finishes.

PR #9389, opened 2017-05-06
Keep track of remote object reference count.

tarruda:
This is required to ensure page reloads are handled properly.

Close #9387

Issue #20350, opened 2019-09-26
Using sandbox and affinity together crashes render process on window close


sgd2z:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 6.0.10
* **Operating System:**
  * Windows 10 (1903)
* **Last Known Working Electron version:**
  * 5.0.8

### Expected Behavior
Render process should not crash when closing one window in the process using browserWindow.close() if sandbox is true and affinity is set.

### Actual Behavior
When using affinity and sandbox together closing any window in code causes the render process to crash making the rest of the windows sharing that affinity to become unusable.

### To Reproduce
See this branch on my fork of quick-start:
https://github.com/sgd2z/electron-quick-start/tree/sandbox-affinity

It launches two windows with the same affinity and closes one of them. The remaining window goes blank and content never shows.

Remove sandbox: true from the config makes things all work fine.

```
$ git clone https://github.com/sgd2z/electron-quick-start.git -b sandbox-affinity
$ npm install
$ npm run start
```

### Screenshots
Crashed Window:
![image](https://user-images.githubusercontent.com/18012933/65708804-64d87b00-e05d-11e9-86f6-4a2f83cfa82f.png)

Actual Content:
![image](https://user-images.githubusercontent.com/18012933/65708949-b6810580-e05d-11e9-8d92-5152e1c88499.png)

kirilpopov:
The Electron team and @MarshallOfSound are working for deprecating affinity (ignoring the set of apps that depend on it), so I guess their answer will be "affinity is experimental feature, you should not depend on it", as in several similar tickets.

lkonstantinov:
This is actually an issue that is related to #17406 and removing the affinity overrides wont help in all cases (i.e. it will happen again if you use the Chromium process per site option).

The renderer process doesnt crash, but rather waits for a mutex that never is signaled. The issue is caused by this:
```
// Notify the main process when current context is going to be released.
// Note that when the renderer process is destroyed, the message may not be
// sent, we also listen to the "render-view-deleted" event in the main process
// to guard that situation.
process.on('exit', () => {
[...]

PR #20632, opened 2019-10-18
fix: send ELECTRON_BROWSER_CONTEXT_RELEASE asynchronously

### LINKED ISSUE #20086: Closing child window hangs main window
### Preflight Checklist

### Issue Details

I have enabled nativeWindowOpen and nodeIntegrationInSubFrames for main window, and I have a preload script which required a remote from electron. Which this configuration I open a child window and then I close it. The main window render process hangs for ever so I have to restart whole app. Here is and example app https://github.com/zarubond/electron-quick-start/tree/child_hang

* **Electron Version:**
6.0.7, 7.0.0-beta3
* **Operating System:**
Windows, Linux, MacOS
5.0.10

### Expected Behavior
Closing child window should not hang main window for ever

### Actual Behavior
Main window hangs after closing child window. 

### To Reproduce
Open child window and then close it

miniak:
#### Description of Change
Fixes parts of #20086 by sending `ELECTRON_BROWSER_CONTEXT_RELEASE` asynchronously.
The synchronous send was added in #9389, which was necessary back then as we didn't have a unique identifier for each JS context.

Fixing `ipcRenderer.sendSync()` to not hang when the mojo connection is lost is also important, but more complicated.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Fixed hang when closing a scriptable popup window using the `remote` module.

### DIFF
--- a/lib/browser/remote/objects-registry.ts
+++ b/lib/browser/remote/objects-registry.ts
@@ -127,6 +127,10 @@ class ObjectsRegistry {
         this.clear(webContents, contextId)
       }
     }
+    // Note that the "render-view-deleted" event may not be emitted on time when
+    // the renderer process get destroyed because of navigation, we rely on the
+    // renderer process to send "ELECTRON_BROWSER_CONTEXT_RELEASE" message to
+    // guard this situation.
     webContents.on('render-view-deleted' as any, listener)
   }
 }
--- a/lib/browser/remote/server.ts
+++ b/lib/browser/remote/server.ts
@@ -529,7 +529,6 @@ handleRemoteCommand('ELECTRON_BROWSER_DEREFERENCE', function (event, contextId, 
 
 handleRemoteCommand('ELECTRON_BROWSER_CONTEXT_RELEASE', (event, contextId) => {
   objectsRegistry.clear(event.sender, contextId)
-  return null
 })
 
 handleRemoteCommand('ELECTRON_BROWSER_GUEST_WEB_CONTENTS', function (event, contextId, guestInstanceId, stack) {
--- a/lib/renderer/api/remote.js
+++ b/lib/renderer/api/remote.js
@@ -19,7 +19,7 @@ const contextId = v8Util.getHiddenValue(global, 'contextId')
 // to guard that situation.
 process.on('exit', () => {
   const command = 'ELECTRON_BROWSER_CONTEXT_RELEASE'
-  ipcRendererInternal.sendSync(command, contextId)
+  ipcRendererInternal.send(command, contextId)
 })
 
 // Convert the arguments object into an array of meta data.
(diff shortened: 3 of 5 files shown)


nornagon:
cc @loc

MarshallOfSound:
Also this doesn't fix the root cause of #20086 it just hides the underlying issue

zcbenz:
### To add some background here:

The `ELECTRON_BROWSER_CONTEXT_RELEASE` message used to be required to be sent synchronously, so the remote references for current WebContents can be cleared before loading the new page, otherwise race conditions would happen.

However later there was some changes in Chromium that, the new page is pre-loaded before the old page gets destroyed, so race condition would happen even though the message was sent synchronously.

The race condition was then fixed by passing the ID of current context with the `ELECTRON_BROWSER_CONTEXT_RELEASE` message (https://github.com/electron/electron/pull/13603), so the remote references are cleared per-context instead of per-WebContents, and the sequence of messages does not matter anymore since then.

 So technically the `ELECTRON_BROWSER_CONTEXT_RELEASE` message no longer has to be sent synchronously. However we have to ensure that the message has actually been sent when navigation happens, and I think we should add a test for it.

### Also some note on how remote references are cleared:

Currently we have 2 ways of clearing remote references: one is listening to `render-view-deleted` event, and the other is sending the `ELECTRON_BROWSER_CONTEXT_RELEASE` message.

Both ways are required because `render-view-deleted` does not work for certain page navigations (depending on which Chromium version), and the `ELECTRON_BROWSER_CONTEXT_RELEASE` message can not be sent successfully when page is destroyed directly.

zcbenz:
For the original issue #20086, before IPC was migrated to mojo, the `sendSync` would simply fail when the page is being destroyed; but with mojo the `sendSync` is waiting for signal manually and Chromium would not unlock it for us automatically.

zcbenz:
I have added tests to ensure the remote references can be cleared for all cases (on navigation and on `destroy()`).

There is still problem that `sendSync` would hang when used in `exit` event while page is being destroyed directly, but I think it should be a separate problem.

lkonstantinov:
I have to agree with @MarshallOfSound - this is just fixing piecemeal and sweeping under the carpet the underlying issue. After we've patched this particular sendSync (about a month or so ago - see #20350) we've been hitting slew of IPC deadlock issues in other situations.

miniak:
@lkonstantinov This PR is not pretending to be fixing the root cause, it's fixing #20086. The change to use an async IPC could have been made sooner as a refactoring as it does not have to be sync anymore as @zcbenz explained.

zcbenz:
@MarshallOfSound Can you take another look of this?

zcbenz:
I'm merging this and keeping the original issue open.

PR #20639, opened 2019-10-18

MarshallOfSound:
Backport of #20307

See that PR for details


MarshallOfSound:
> this is supposed to be for 6-1-x not 6-0-x right?

Yes but that branch doesn't exist yet, I put this PR up for CI.

MarshallOfSound:
This required cherry picking the V8 serializer and a `base::OnceCallback` converter in addition to the raw changes from the original PR 👍

Issue #15202, opened 2018-10-16
Multi-Language Spellchecker doesn't work.


ccorcos:
* Output of `node_modules/.bin/electron --version`: v2.0.5
* Operating System (Platform and Version): MacOS

The following will not check spelling for words with russian letters.

```ts
electron.webFrame.setSpellCheckProvider({
    language: "en",
    ...
})
```

The following will not check spelling for words with english letters.

```ts
electron.webFrame.setSpellCheckProvider({
    language: "ru",
    ...
})
```

The following will not check spelling for words with russian letters.

```ts
electron.webFrame.setSpellCheckProvider({
    language: "en,ru",
    ...
})
```

At the very least, I would hope that `en,ru` should work.

ccorcos:
The example you showed me is still just one language at a time. 

I'm using [node-cld](https://github.com/dachev/node-cld) which can detect multiple languages in the same string. For example `Предложение на русскм языке без ошиб hello world. does this work well? ` returns Russian and English. 

Using [node-spellchecker](https://github.com/atom/node-spellchecker) I can check if a word is misspelled in either language, but the electron spellchecking api is not checking all words so I'm fundamentally limited by Electron now.

Honestly, I'm not sure why I have to provide a language to setSpellCheckProvider. Ideally it would always check all words and leave it up to the developer. 

Please reconsider opening this issue. Chrome spellchecks multiple languages at the same time very gracefully. Ideally we can have a similar functionality in electron apps.

ccorcos:
This looks like the heart of the issue:

https://github.com/electron/electron/blob/6f91af93433df4e9c0e7fb592e5b2dcb0b1c7888/chromium_src/chrome/renderer/spellchecker/spellcheck_worditerator.h#L37-L41

ccorcos:
Thanks for the breakdown. Looks like you're ahead of me. Are you using your hunspell library in an application? Mind if I ask which app?

kwonoj:
> Are you using your hunspell library in an application

yes. https://github.com/kwonoj/electron-hunspell uses https://github.com/kwonoj/hunspell-asm in behind, and I'm using https://github.com/kwonoj/cld3-asm for language detection.

> Mind if I ask which app

Sorry, allow me to not share which app in here.

Issue #7244, opened 2016-09-18
Spellchecker provider character sets & multiple spell check languages


Thomas101:
* Electron version: 1.4.0
* Operating system: osx

I've noticed that when setting the spellcheck provider for a certain language it doesn't run spell checking over character sets not used in that language (e.g. american characters vs cyrillics). For example if you set..

```js
webFrame.setSpellCheckProvider('en-US', true, {
  spellCheck (text) {
    console.log(text)
    return false
  }
})
```

Then if you start typing in a text field you'll find the following...
* Type `english` you see `english`
* Type `français` you see `français`
* Type `русский` nothing gets logged out.

Swap the provider language to `ru` you'll see the results in reverse.

Taking this into account is it possible to...

1.) Run the spellchecker over different character sets? (I'm guessing the answer to this is going to be no as I can't imagine chrome supports it...)
2.) Provide spellchecking for multiple languages at once? I noticed that if you run the `setSpellCheckProvider` multiple times the last one wins
3.) Give any information on the character sets used for the different languages so I can build a matrix of compatible languages?

Thanks for your help :-)

enlight:
I'm not familiar with the spellchecker implementation so I can't answer your questions, but I can point you at https://github.com/paulcbetts/electron-spellchecker which seems to support switching between multiple languages on the fly.

Thomas101:
Thanks @enlight. I've had a look at that lib and it works around this issue by checking if there's been a [number of unchecked words by the spellchecker](https://github.com/paulcbetts/electron-spellchecker/blob/7bda47a5395f599ebcb418a1f1348e9e97f16f68/src/spell-check-handler.js#L183), then uses that to infer if it should try to re-detect language and auto change the language from there.

I was looking at getting more than one language spell checking at once though. Some users have been asking for suggestions in multiple languages as they quite often type bilingually

MarshallOfSound:
@Thomas101 I've never tested this, but I can't imagine that Chrome or even Microsoft Words does multilingual spell checking.  For one main reason I can think of.

If a word is valid in one language and invalid in another.  Is it an error?  The spell checker would have to contextually understand what language the block of test the word is in is written in.  And even then, quoting a different language would result in errors.

Spell checking multiple languages at the same time will either result in an increase in false positives or ignoring real spelling mistakes.  I don't see how this could possible be implementing in a consistent and precise manner 😢 

> Run the spellchecker over different character sets?

This however definitely seems to be some form of a bug 😆

[...]

PR #20648, opened 2019-10-19
feat: allow multi language spellcheckprovider

kwonoj:
#### Description of Change
- Relates to #15202, #7244

This PR attempts to expand `setSpellCheckProvider` to support multi-language spellchecker. Consumer of interface can opt-in by specifying multiple languages when initializing spellcheck provider. SpellcheckClient will create wordIterator per each language specified, will call provider function with object have language as keys, and word to check spell as its values.

Marking misspelling logic is not changed, as long as async callback returns word, it'll be marked as error. It's up to consumer's decision as same as before.
 
#### Checklist

- [ ] `npm test` passes
- [x] tests are [changed](https://github.com/electron/electron/blob/master/docs/development/testing.md)

#### Release Notes

Notes: `webFrame.setSpellcheckerProvider` supports multiple languages.

### DIFF
--- a/docs/api/web-frame.md
+++ b/docs/api/web-frame.md
@@ -65,6 +65,15 @@ Sets the maximum and minimum layout-based (i.e. non-visual) zoom level.
 
 ### `webFrame.setSpellCheckProvider(language, provider)`
 
+* `language` String[]
+* `provider` Object
+  * `spellCheck` Function
+    * `words` Record<String, String[]>
+    * `callback` Function
+      * `misspeltWords` String[]
+
+### `webFrame.setSpellCheckProvider(language, provider)`
+
 * `language` String
 * `provider` Object
   * `spellCheck` Function
@@ -75,7 +84,8 @@ Sets the maximum and minimum layout-based (i.e. non-visual) zoom level.
 Sets a provider for spell checking in input fields and text areas.
 
 The `provider` must be an object that has a `spellCheck` method that accepts
-an array of individual words for spellchecking.
+an array of individual words for spellchecking for single language,
+or an object for multiple languages.
 The `spellCheck` function runs asynchronously and calls the `callback` function
 with an array of misspelt words when complete.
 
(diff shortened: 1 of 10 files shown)


kwonoj:
@nitsakh and I discussed that actually, maybe it's possible - but it requires languages param (`string` | `Array<string>`) and spellcheck fn's param (`Array<string> | Record<string, Array<string>`) both. 

> when you have to check Electron version in your code

Majority of case application relies on certain version of electron and when they update it they just need to apply breaking changes, isn't it? I agree breaking changes could require effort to migration, don't think many cases single code runs on multiple electron version.

miniak:
@kwonoj in our case unfortunately due to the complicated way we deploy changes we need to be able to build the same code with multiple versions of Electron. We had to do a version check to handle the previous breaking change made in Electron 5 (which we backported to out internal Electron 4 builds to make it even more complicated).

kwonoj:
OK. I personally think this would be ok to be breaking (especially migration should be easy enough) but don't have strong opinion. Will try to update changes to interop with current interfaces.

docs/api/web-frame.md
```diff
@@ -65,6 +65,15 @@ Sets the maximum and minimum layout-based (i.e. non-visual) zoom level.
...
 ### `webFrame.setSpellCheckProvider(language, provider)`
 
+* `language` String[]
+* `provider` Object
+  * `spellCheck` Function
+    * `words` Record<String, String[]>
+    * `callback` Function
+      * `misspeltWords` String[]
+
+### `webFrame.setSpellCheckProvider(language, provider)`
```
kwonoj: Not sure what would be recommended way to create overload definition. Specifying union type (`language: String | String[]` and `words: String[] | Record<String, String[]>`) cannot narrow down param types, especially for `words` param `String[] | Record<String, String[]>` does not have interesection it's force required to cast types (which could be breaking changes in terms of types).

To make this work, listed each definition separately to create overloaded fn signatures.

kwonoj:
@miniak @MarshallOfSound PR updated to support existing interfaces.

shell/renderer/api/atom_api_web_frame.cc
```diff
@@ -314,6 +314,16 @@ void SetSpellCheckProvider(mate::Arguments* args,
     return;
   }
 
+  std::vector<std::string> language_vector;
+  bool is_multi_lang_spell_check = languages->IsArray();
+  if (is_multi_lang_spell_check) {
+    gin::ConvertFromV8(args->isolate(), languages, &language_vector);
```
MarshallOfSound: This conversion could fail, it should be handled

shell/renderer/api/atom_api_web_frame.cc
```diff
@@ -314,6 +314,16 @@ void SetSpellCheckProvider(mate::Arguments* args,
...
+  } else {
+    std::string language;
+    gin::ConvertFromV8(args->isolate(), languages, &language);
```
MarshallOfSound: Same with this conversion, if both fail we should throw an error

miniak:
Is this still relevant after #20692?

zcbenz:
I'm closing this PR since with #20692 this PR should be unnecessary. Please let me know if I'm wrong and I will reopen this.

PR #20666, opened 2019-10-21
refactor: use [Sync,NoDeadlockPrevention] in for MessageSync

nornagon:
#### Description of Change
This removes a bunch of hacks we have to pretend that the `ElectronBrowser.Message` mojo function is asynchronous, by adding a new `[NoDeadlockPrevention]` flag to Mojo.

Some background: Mojo already has the functionality to be able to send synchronous messages; i.e. messages on which the sender will block until a response is received. In order to prevent a situation in which two processes are waiting for each other (or other deadlock situations), if process (A) sends a synchronous message to (B) and is waiting on the result, and then process (A) receives a synchronous request from process (C), it will process it immediately (even though it's still waiting for the response to its first message). This can result in synchronous messages being received out-of-order with respect to asynchronous messages.

There are some IPC messages in Electron which rely on strict ordering between synchronous and asynchronous messages. Most of the time, the main process doesn't send any synchronous IPC messages and `[Sync]` would work just fine (because the reordering only kicks in if a process receives a synchronous message while it's waiting for a response to a synchronous message that it sent). However, there are a few cases in window shutdown where the main process _does_ send a synchronous message, resulting in out-of-order message delivery.

This new `[NoDeadlockPrevention]` flag disables the deadlock-breaking feature for the method, meaning that correct ordering will be guaranteed w.r.t. other asynchronous messages sent on the same channel. The sending side still sees the message as synchronous, but the receiving side won't reorder the message on receipt if it's in the middle of making a synchronous request itself.

The main reason to do this is that our existing approach with `ThreadSafeElectronBrowserPtr` cannot listen for channel errors, since the `ThreadSafeInterfacePtr` class doesn't have that functionality, resulting in a possible issue where the process issues a request and then blocks waiting for a response which it will never receive, because the remote end has hung up. This PR doesn't fix that issue, but it does make it _possible_ to fix.

Upstream CL: https://chromium-review.googlesource.com/c/chromium/src/+/1872981

Ref #20086 #20547 #20632

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -81,3 +81,4 @@ ui_views_fix_jumbo_build.patch
 export_fetchapi_mojo_traits_to_fix_component_build.patch
 build_fix_when_building_with_enable_plugins_false.patch
 add_zoom_limit_setters_to_webcontents.patch
+add_nodeadlockprevention_mojo_flag.patch
--- a/patches/chromium/add_nodeadlockprevention_mojo_flag.patch
+++ b/patches/chromium/add_nodeadlockprevention_mojo_flag.patch
@@ -5,56 +5,62 @@ Subject: add [NoDeadlockPrevention] mojo flag
 
 
 diff --git a/ipc/ipc_mojo_bootstrap.cc b/ipc/ipc_mojo_bootstrap.cc
-index 8817d8e23d72d6a60cdd4ece5809ac0a6d2f59fa..5e45ebe9c763b4ce78d1c5d93621f4822c51d891 100644
+index 8817d8e23d72d6a60cdd4ece5809ac0a6d2f59fa..fa3322b4698f84d68e81d7e38e4efe73f16c6636 100644
 --- a/ipc/ipc_mojo_bootstrap.cc
 +++ b/ipc/ipc_mojo_bootstrap.cc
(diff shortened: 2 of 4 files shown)


deepak1556:
This is cool!

> The main reason to do this is that our existing approach with ThreadSafeElectronBrowserPtr cannot listen for channel errors

Can you also please give some background on why the current approach was taken instead of fixing this ?

nornagon:
`ThreadSafeElectronBrowserPtr` doesn't have the `set_connection_error_handler` method that `ElectronBrowserPtr` has, so it's not possible to listen for connection errors through that class.

deepak1556:
`ThreadSafeInterfacePtr` operates on the `InterfacePtr` and has access to the ptr state , so we could add a method to set the connection handler right https://cs.chromium.org/chromium/src/mojo/public/cpp/bindings/thread_safe_interface_ptr.h?l=169-176 ?

deepak1556:
Having said that, I don't have the best expertise to judge which approach is better here, if upstream accepts the deadlock flag, we should very well use it as its much more straight forward and makes our implementation simpler.

nornagon:
update: chances of upstream accepting my patch are looking gloomy.

PR #20678, opened 2019-10-22
feat: Add Secure Keyboard Entry APIs in macOS

mtgto:
#### Description of Change

Add two APIs to `app` to control `Secure Keyboard Entry`.

In macOS, there is a way to protect from other processes listen keyboard input event.
https://developer.apple.com/library/archive/technotes/tn2150/_index.html

To enable it, we call `EnableSecureEventInput`, which available since Mac OS X 10.3.

This functions helps the apps become more security like https://github.com/bitwarden/desktop/issues/298

`app.setSecureKeyboardEntryEnabled(enabled)` protects your application (your process) from listening keyboard input events. You have to disable before quit application.

`app.isSecureKeyboardEntryEnabled()` returns true if you already called `app.setSecureInputEnabled(true)`.

To confirm your application is protected or not, below command is helpful.

```console
ioreg -l -w 0 | grep kCGSSessionSecureInputPID
```

I write sample Electron app to test this PR is working. (2020-05-17 added)
https://github.com/mtgto/electron-secure-event-input-sample

I test with [Swift-Keylogger](https://github.com/SkrewEverything/Swift-Keylogger) whether this API protect from keylogger.
[Here is my electron sample](https://gist.github.com/mtgto/6f26b79e3f23b466357440e04544bc45).

Password textfield is already protected by Chromium, but plain textfield is not.
I confirmed this API protect plain textfield from keylogger.

Here is the list of other open source softwares use the API.

Chromium
https://cs.chromium.org/chromium/src/ui/base/cocoa/secure_password_input.mm

iTerm2
https://github.com/gnachman/iTerm2/blob/master/sources/iTermSecureKeyboardEntryController.m

#### Checklist

- [ ] `npm test` passes(https://github.com/electron/electron/blob/master/docs/development/testing.md)

`npm test` failed in my environment:
```
error /Users/user/work/js/electron/electron-gn/src/electron/spec/node_modules/is-valid-window: Command failed.
Exit code: 1
Command: node-gyp rebuild
Arguments:
Directory: /Users/user/work/js/electron/electron-gn/src/electron/spec/node_modules/is-valid-window
Output:
gyp: /Users/user/work/js/electron/electron-gn/src/out/Debug/gen/node_headers/common.gypi not found (cwd: /Users/user/work/js/electron/electron-gn/src/electron/spec/node_modules/is-valid-window) while reading includes of binding.gyp while trying to load binding.gyp
```

#### Release Notes

Notes: add `app.isSecureInputEnabled()` and `app.setSecureInputEnabled(enabled)` to manage [Secure Keyboard Entry](https://developer.apple.com/library/archive/technotes/tn2150/_index.html) in macOS.

### DIFF
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -1280,33 +1280,28 @@ app.moveToApplicationsFolder({
 
 Would mean that if an app already exists in the user directory, if the user chooses to 'Continue Move' then the function would continue with its default behavior and the existing app will be trashed and the active app moved into its place.
 
-### `app.isSecureInputEnabled()` _macOS_
+### `app.isSecureKeyboardEntryEnabled()` _macOS_
 
 Returns `Boolean` - whether `Secure Keyboard Entry` is enabled.
 
-This function might return `true` even if you have not called `app.setSecureInputEnabled(true)`
-but other process already called.
-For instance, Terminal.app has a menu item `Secure Keyboard Entry` in `Terminal` menu.
-If you already set enabled it, this method always returns `true`.
-You need to `app.setSecureInputEnabled(true)` to protect from other process listens input events.
+By default this API will return `false`.
 
-For details, see [TN2150](https://developer.apple.com/library/content/technotes/tn2150/_index.html).
+On _Linux_ and _Windows_, this API will return `false`.
 
-### `app.setSecureInputEnabled(enabled)` _macOS_
+### `app.setSecureKeyboardEntryEnabled(enabled)` _macOS_
 
 * `enabled` Boolean - Enable or disable `Secure Keyboard Entry`
 
-Set `Secure Keyboard Entry` is enabled.
+Set the `Secure Keyboard Entry` is enabled in your application.
 
-This function is protect from other process listens input events.
+By using this API, important information such as password and other sensitive information can be prevented from being intercepted by other processes.
 
-After you enable it, you should disable before application quited.
+See [Apple's documentation](https://developer.apple.com/library/archive/technotes/tn2150/_index.html) for more
+details.
 
-```js
-app.on('quit', () => {
-  app.setSecureInputEnabled(false)
-})
-```
+**Note:** Enable only when it is needed and disable it when it is no longer needed.
+
+On _Linux_ and _Windows_, this API will do nothing.
 
 ## Properties
 
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -1304,7 +1304,7 @@ After you enable it, you should disable before application quited.
 
 ```js
 app.on('quit', () => {
-  app.setSecureInputEnabled(false);
+  app.setSecureInputEnabled(false)
 })
 ```
 
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -1280,6 +1280,34 @@ app.moveToApplicationsFolder({
 
 Would mean that if an app already exists in the user directory, if the user chooses to 'Continue Move' then the function would continue with its default behavior and the existing app will be trashed and the active app moved into its place.
 
+### `app.isSecureInputEnabled()` _macOS_
+
+Returns `Boolean` - whether `Secure Keyboard Entry` is enabled.
+
+This function might return `true` even if you have not called `app.setSecureInputEnabled(true)`
+but other process already called.
+For instance, Terminal.app has a menu item `Secure Keyboard Entry` in `Terminal` menu.
+If you already set enabled it, this method always returns `true`.
+You need to `app.setSecureInputEnabled(true)` to protect from other process listens input events.
+
+For details, see [TN2150](https://developer.apple.com/library/content/technotes/tn2150/_index.html).
+
+### `app.setSecureInputEnabled(enabled)` _macOS_
+
+* `enabled` Boolean - Enable or disable `Secure Keyboard Entry`
+
+Set `Secure Keyboard Entry` is enabled.
+
+This function is protect from other process listens input events.
+
+After you enable it, you should disable before application quited.
+
+```js
+app.on('quit', () => {
+  app.setSecureInputEnabled(false);
+})
+```
+
 ## Properties
 
 ### `app.accessibilitySupportEnabled` _macOS_ _Windows_
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -1286,8 +1286,6 @@ Returns `Boolean` - whether `Secure Keyboard Entry` is enabled.
 
 By default this API will return `false`.
 
-On _Linux_ and _Windows_, this API will return `false`.
-
 ### `app.setSecureKeyboardEntryEnabled(enabled)` _macOS_
 
 * `enabled` Boolean - Enable or disable `Secure Keyboard Entry`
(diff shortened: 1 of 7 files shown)


docs/api/app.md
```diff
@@ -1265,6 +1265,34 @@ app.moveToApplicationsFolder({
 
 Would mean that if an app already exists in the user directory, if the user chooses to 'Continue Move' then the function would continue with its default behavior and the existing app will be trashed and the active app moved into its place.
 
+### `app.isSecureInputEnabled()` _macOS_
+
+Returns `Boolean` - whether `Secure Keyboard Entry` is enabled.
+
+This function might return `true` even if you have not called `app.setSecureInputEnabled(true)`
+but other process already called.
```
miniak: but other process already called **it**

docs/api/app.md
```diff
@@ -1265,6 +1265,34 @@ app.moveToApplicationsFolder({
...
+For instance, Terminal.app has a menu item `Secure Keyboard Entry` in `Terminal` menu.
+If you already set enabled it, this method always returns `true`.
```
miniak: there is a redundant it here

shell/browser/browser_mac.mm
```diff
@@ -393,4 +394,26 @@ void RemoveFromLoginItems() {
...
+bool Browser::IsSecureInputEnabled() {
+  // See
+  // https://developer.apple.com/library/content/technotes/tn2150/_index.html
+  return IsSecureEventInputEnabled();
+}
+
+void Browser::SetSecureInputEnabled(bool enabled) {
+  // See
+  // https://developer.apple.com/library/content/technotes/tn2150/_index.html
+  if (enabled) {
```
miniak: calling `setSecureInputEnabled()` with the same value twice should have no effect IMO.

mtgto:
@miniak I apologize for wasting your time for my first PR. I almost rewrote.

There are several changes since first pull request:

1. Rename API
    - User easily realize what the APIs are. The name is from menuitem of Terminal.app.
1. Use [ScopedPasswordInputEnabler](https://cs.chromium.org/chromium/src/ui/base/cocoa/secure_password_input.mm) in Chromium instead of call `EnableSecureEventInput()` by API itself
    - While debugging, I found `app.SetSecureKeyboardEntryEnabled(true)` causes crash Electron. This is because Chromium already use `Secure Keyboard Entry` in password textfield.
    - Please grep with `ScopedPasswordInputEnabler` in [textfield.cc](https://cs.chromium.org/chromium/src/ui/views/controls/textfield/textfield.cc)
1. Rewrite API doc

shell/browser/browser_mac.mm
```diff
@@ -393,4 +393,17 @@ void RemoveFromLoginItems() {
...
 }
 
+bool Browser::IsSecureKeyboardEntryEnabled() {
+  return password_input_enabler_.get() != nullptr;
+}
+
+void Browser::SetSecureKeyboardEntryEnabled(bool enabled) {
+  if (enabled) {
+    password_input_enabler_ =
+        std::make_unique<ui::ScopedPasswordInputEnabler>();
```
mtgto: initializer enable Secure Keyboard Entry.
https://cs.chromium.org/chromium/src/ui/base/cocoa/secure_password_input.mm

zcbenz:
@miniak Can you take another look at this?

docs/api/app.md
```diff
@@ -1265,6 +1265,29 @@ app.moveToApplicationsFolder({
 
 Would mean that if an app already exists in the user directory, if the user chooses to 'Continue Move' then the function would continue with its default behavior and the existing app will be trashed and the active app moved into its place.
 
+### `app.isSecureKeyboardEntryEnabled()` _macOS_
+
+Returns `Boolean` - whether `Secure Keyboard Entry` is enabled.
+
+By default this API will return `false`.
+
+On _Linux_ and _Windows_, this API will return `false`.
```
miniak, line 1274: This API should probably be only exposes on Mac.

docs/api/app.md
```diff
@@ -1265,6 +1265,29 @@ app.moveToApplicationsFolder({
...
+Set the `Secure Keyboard Entry` is enabled in your application.
+
+By using this API, important information such as password and other sensitive information can be prevented from being intercepted by other processes.
+
+See [Apple's documentation](https://developer.apple.com/library/archive/technotes/tn2150/_index.html) for more
+details.
+
+**Note:** Enable only when it is needed and disable it when it is no longer needed.
+
+On _Linux_ and _Windows_, this API will do nothing.
```
miniak, line 1289: The same here, the method documentation says it's only available on Mac.

shell/browser/browser.h
```diff
@@ -259,6 +263,11 @@ class Browser : public WindowListObserver {
 
   void RemoveObserver(BrowserObserver* obs) { observers_.RemoveObserver(obs); }
 
+  // Returns whether secure input is enabled
+  bool IsSecureKeyboardEntryEnabled();
+
```
miniak, line 268: why the space?

mtgto:
Thanks for your review. I'll fix this weekend.

mtgto:
- rebase current HEAD of master
- Add example app https://github.com/mtgto/electron-secure-event-input-sample

zcbenz:
Can you fix the lint error?

```
 
shell/browser/browser_mac.mm:440:  Add #include <memory> for make_unique<>  [build/include_what_you_use] [4]
 
Total errors found: 1
 
```

jkleinsc:
@electron/wg-api reviewed on May-18-2020.

mtgto:
> In particular, I'm worried that this might interact poorly with the existing password field security, where calling setSecureKeyboardEntryEnabled(true), then focusing a password field, then focusing out would result in the secure keyboard entry being disabled.

I try to reproduce it by using sample app, but I can't.
https://github.com/mtgto/electron-secure-event-input-sample
Please tell me the way and your environment.

mtgto:
@nornagon

> Would it be possible to write a test for this?

added. https://github.com/electron/electron/pull/20678/commits/619319f275c4c18189735e296beea8600ba8bff7

> In particular, I'm worried that this might interact poorly with the existing password field security, where calling setSecureKeyboardEntryEnabled(true), then focusing a password field, then focusing out would result in the secure keyboard entry being disabled.

I try to reproduce it by using sample app, but I can't.
https://github.com/mtgto/electron-secure-event-input-sample
Please tell me the way and your environment.

nornagon:
> @nornagon
> 
> > Would it be possible to write a test for this?
> 
> added. [619319f](https://github.com/electron/electron/commit/619319f275c4c18189735e296beea8600ba8bff7)

Thanks. I was hoping that it would be possible to test that the underlying feature was working—i.e. that it was protecting input from being read by other applications—but that seems hard and this test at least exercises the code path.

> > In particular, I'm worried that this might interact poorly with the existing password field security, where calling setSecureKeyboardEntryEnabled(true), then focusing a password field, then focusing out would result in the secure keyboard entry being disabled.
> 
> I try to reproduce it by using sample app, but I can't.
> https://github.com/mtgto/electron-secure-event-input-sample
> Please tell me the way and your environment.

Thanks, your test looks good. Looking at the underlying code in chromium it looks like there's a counter to track how many ScopedPasswordInputEnablers are present, so it should be fine.

Issue #20606, opened 2019-10-16
shell.openItem returns always true on Linux


xmedeko:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details
```js
let ret = shell.openItem('file:///noexistent');
console.log(ret);
```
`ret` is always `true` on Linux even if opening fails, e.g.
```sh
xdg-open 'file:///noexistent'
echo $?
```
Shows nonzero error code.

* **Electron Version:**
  * v6.0.11
* **Operating System:**
  *Linux / Ubuntu 16.04 x64

### Expected Behavior
`ret` should be `false`, if `xdg-open` returns non-zero code.

Optional: It would be very nice if the `xdg-open` error (catch stderr) would be displayed in `dialog.showErrorBox` if it returns non-zero code. Just take care that `xdg-open` may write some warnings to stderr even if it returns 0.

codebytere:
@xmedeko hey! you're right - the way we've implemented this it just returns true when the command spawns successfully but doesn't wait for the exit code. I'll look into improving this soon!

Issue #7210, opened 2016-09-14
shell.openItem() is not async


WholeMilk:
- Electron version: 1.3.6
- Operating system: Windows 7 x64
- Other: Microsoft Office 2013 (will make more sense in a bit)

`shell.openItem()` is a blocking call, it would be nice if it's async or has an async version to handle synchronous external application behavior.

Reproduce: 
Open a password-protected Microsoft Excel file, `shell.openItem()` will be blocking until the Password window is submitted or dismissed. Consequently Microsoft Word does not have this issue.

Related to #2761.

kevinsawicki:
Merging this into #6889 which had already requested something similar for `shell.openExternal`

dominic-simplan:
@kevinsawicki could you reopen this, as it seems that #6889 has been closed without implementing async behaviour for openItem().

PR #20682, opened 2019-10-22
feat: convert shell.openItem to async shell.openPath

codebytere:
#### Description of Change

BREAKING CHANGE.

Resolves https://github.com/electron/electron/issues/20606.
Resolves https://github.com/electron/electron/issues/7210.

Implements an asynchronous version of `shell.openItem(path)` as a new method `shell.openPath(path)`.

Adheres to [this spec](https://github.com/electron/governance/blob/master/wg-api/RFCs/shell-openitem.md).

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Split `shell.openItem(path)` into synchronous and asynchronous methods.

### DIFF
--- a/docs/api/shell.md
+++ b/docs/api/shell.md
@@ -24,11 +24,14 @@ The `shell` module has the following methods:
 
 Show the given file in a file manager. If possible, select the file.
 
-### `shell.openItem(fullPath)`
+### `shell.openPath(path)`
 
-* `fullPath` String
+* `path` String
 
-Returns `Boolean` - Whether the item was successfully opened.
+Returns `Promise<Object>` - Resolve with an object containing the following:
+
+* `success` Boolean - whether or not the path was successfully opened in the desktop's default manner.
+* `errorMessage` String (optional) - The error message corresponding to the failure if a failure occurred, otherwise empty string.
 
 Open the given file in the desktop's default manner.
 
--- a/shell/browser/ui/inspectable_web_contents_impl.cc
+++ b/shell/browser/ui/inspectable_web_contents_impl.cc
@@ -559,7 +559,9 @@ void InspectableWebContentsImpl::ShowItemInFolder(
     return;
 
   base::FilePath path = base::FilePath::FromUTF8Unsafe(file_system_path);
-  platform_util::OpenItem(path);
+
+  // Pass empty callback here; we can ignore errors
+  platform_util::OpenPath(path, platform_util::OpenCallback());
 }
 
 void InspectableWebContentsImpl::SaveToFile(const std::string& url,
--- a/shell/common/api/atom_api_shell.cc
+++ b/shell/common/api/atom_api_shell.cc
@@ -44,8 +44,8 @@ struct Converter<base::win::ShortcutOperation> {
 
 namespace {
 
-void OnOpenExternalFinished(gin_helper::Promise<void> promise,
-                            const std::string& error) {
+void OnOpenFinished(gin_helper::Promise<void> promise,
+                    const std::string& error) {
   if (error.empty())
     promise.Resolve();
   else
@@ -66,8 +66,21 @@ v8::Local<v8::Promise> OpenExternal(const GURL& url, gin::Arguments* args) {
   }
 
   platform_util::OpenExternal(
-      url, options,
-      base::BindOnce(&OnOpenExternalFinished, std::move(promise)));
+      url, options, base::BindOnce(&OnOpenFinished, std::move(promise)));
+  return handle;
+}
+
+v8::Local<v8::Promise> OpenPath(v8::Isolate* isolate,
+                                const base::FilePath& full_path) {
+  gin_helper::Promise<const std::string&> promise(isolate);
+  v8::Local<v8::Promise> handle = promise.GetHandle();
+
+  platform_util::OpenPath(
+      full_path,
+      base::BindOnce(
+          [](gin_helper::Promise<const std::string&> promise,
+             const std::string& err_msg) { promise.Resolve(err_msg); },
+          std::move(promise)));
   return handle;
 }
 
@@ -142,7 +155,7 @@ void Initialize(v8::Local<v8::Object> exports,
                 void* priv) {
   gin_helper::Dictionary dict(context->GetIsolate(), exports);
   dict.SetMethod("showItemInFolder", &platform_util::ShowItemInFolder);
-  dict.SetMethod("openItem", &platform_util::OpenItem);
+  dict.SetMethod("openPath", &OpenPath);
   dict.SetMethod("openExternal", &OpenExternal);
   dict.SetMethod("moveItemToTrash", &MoveItemToTrash);
   dict.SetMethod("beep", &platform_util::Beep);
--- a/shell/common/platform_util_linux.cc
+++ b/shell/common/platform_util_linux.cc
@@ -19,7 +19,25 @@
 
 namespace {
 
-bool XDGUtil(const std::vector<std::string>& argv, const bool wait_for_exit) {
+// Descriptions pulled from https://linux.die.net/man/1/xdg-open
+std::string GetErrorDescription(int error_code) {
+  switch (error_code) {
+    case 1:
+      return "Error in command line syntax";
+    case 2:
+      return "The item does not exist";
+    case 3:
+      return "A required tool could not be found";
+    case 4:
+      return "The action failed";
+    default:
+      return "";
+  }
+}
+
+bool XDGUtil(const std::vector<std::string>& argv,
+             const bool wait_for_exit,
+             platform_util::OpenCallback callback) {
   base::LaunchOptions options;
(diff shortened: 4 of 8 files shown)


deepak1556:
@codebytere this will conflict https://github.com/electron/electron/pull/10902 in behavior. /cc @zcbenz

codebytere:
@deepak1556 hmm - why don't we allow in an optional callback with an error code and description string if one exists? i think that would solve both issues 🤔

deepak1556:
There is no reason to not implement it https://github.com/electron/electron/issues/7210 😄

shell/common/api/atom_api_shell.cc
```diff
@@ -67,8 +67,17 @@ v8::Local<v8::Promise> OpenExternal(const GURL& url, gin::Arguments* args) {
...
 
   platform_util::OpenExternal(
-      url, options,
-      base::BindOnce(&OnOpenExternalFinished, std::move(promise)));
+      url, options, base::BindOnce(&OnOpenFinished, std::move(promise)));
+  return handle;
+}
+
+v8::Local<v8::Promise> OpenItem(const base::FilePath& full_path,
+                                gin::Arguments* args) {
```
zcbenz: Should pass `v8::Isolate* isolate` if only the isolate is needed.

shell/common/platform_util_linux.cc
```diff
@@ -13,13 +11,32 @@
...
+    case 4:
+      return "The action failed.";
+    default:
+      return "";
+  }
+}
+
+bool XDGUtil(const std::vector<std::string>& argv,
+             const bool wait_for_exit,
+             base::Optional<platform_util::OpenCallback> callback) {
```
zcbenz: The `base::Callback` supports null values so there is no need to use `base::Optional`.

```c++
if (callback.is_null()) {
}
```

shell/common/platform_util_mac.mm
```diff
@@ -58,6 +58,28 @@
       stringByAppendingString:@".loginhelper"];
 }
 
+std::string OpenItemOnThread(const base::FilePath& full_path) {
+  DCHECK([NSThread isMainThread]);
+
+  NSString* path_string = base::SysUTF8ToNSString(full_path.value());
+  if (!path_string)
```
zcbenz: We do not need to check result of `SysUTF8ToNSString`, if it returns nullptr then it is out of memory error, not because of invalid path passed.

The original code was quite ancient, and it did lots of unnecessary checks.

shell/common/platform_util_mac.mm
```diff
@@ -58,6 +58,28 @@
...
+    return "Invalid path string";
+
+  NSURL* url = [NSURL fileURLWithPath:path_string];
+  if (!url)
+    return "Invalid URL";
```
zcbenz: Should warn invalid path here.

docs/api/shell.md
```diff
@@ -28,6 +28,14 @@ Show the given file in a file manager. If possible, select the file.
 
 * `fullPath` String
 
+Returns `Promise<String>` - Resolves with description corresponding to the error if the operation failed, or empty string if the operation succeeded.
+
+Open the given file in the desktop's default manner.
+
+### `shell.openItemSync(fullPath)`
```
zcbenz: The document should explain what `Sync` means here. It does not mean the API would always wait until the opening application exits, it means the API would only wait on Linux for certain applications.

We probably do not need this API at all, as the behavior of this API is confusing, and no one has requested for it.

zcbenz:
Generally I think we should not change API's behavior and return value silently, otherwise we would not be able to show deprecation warnings, and it would very hard to users to debug if any bug was introduced to their apps due to the behavior change.

For example, previous silent errors would now turn into unhandled exceptions.

A safe choice would be deprecating `openItem` and renaming the APIs to `openPath`/`openPathSync`.

/cc @electron/wg-api for more inputs.

codebytere:
@zcbenz fair point! i'll address these mechanical changes and then add this to the agenda for tomorrow :)

shell/common/platform_util_linux.cc
```diff
@@ -32,54 +49,60 @@ bool XDGUtil(const std::vector<std::string>& argv, const bool wait_for_exit) {
...
-  return XDGUtil({"xdg-open", path}, wait_for_exit);
+bool XDGOpen(const std::string& path,
+             const bool wait_for_exit,
+             platform_util::OpenCallback callback) {
+  return XDGUtil({"xdg-open", path}, wait_for_exit, std::move(callback));
 }
 
 bool XDGEmail(const std::string& email, const bool wait_for_exit) {
-  return XDGUtil({"xdg-email", email}, wait_for_exit);
+  return XDGUtil({"xdg-email", email}, wait_for_exit, base::nullopt);
```
zcbenz: Should pass `OpenCallback()` for null callback.

shell/browser/ui/inspectable_web_contents_impl.cc
```diff
@@ -559,7 +559,9 @@ void InspectableWebContentsImpl::ShowItemInFolder(
     return;
 
   base::FilePath path = base::FilePath::FromUTF8Unsafe(file_system_path);
-  platform_util::OpenItem(path);
+
+  // TODO(codebytere): figure out best apth for this
+  // platform_util::OpenPath(path);
```
zcbenz: I think you can just pass empty callback `platform_util::OpenCallback` here to ignore errors.

codebytere:
This should be ready to go per the spec!

shell/common/platform_util_linux.cc
```diff
@@ -32,54 +49,61 @@ bool XDGUtil(const std::vector<std::string>& argv, const bool wait_for_exit) {
   if (!process.IsValid())
     return false;
 
-  if (wait_for_exit) {
+  if (wait_for_exit || !callback.is_null()) {
```
deepak1556: redundant check.

shell/common/platform_util_linux.cc
```diff
@@ -32,54 +49,61 @@ bool XDGUtil(const std::vector<std::string>& argv, const bool wait_for_exit) {
...
     int exit_code = -1;
     if (!process.WaitForExit(&exit_code))
```
deepak1556: If the process doesn't wait for exit then the api's promise will be unhandled. The callback should be handled here.

shell/common/platform_util_linux.cc
```diff
@@ -32,54 +49,61 @@ bool XDGUtil(const std::vector<std::string>& argv, const bool wait_for_exit) {
...
 
-  XDGOpen(dir.value(), false);
+  XDGOpen(dir.value(), false, platform_util::OpenCallback());
 }
 
-bool OpenItem(const base::FilePath& full_path) {
-  return XDGOpen(full_path.value(), false);
+void OpenPath(const base::FilePath& full_path, OpenCallback callback) {
+  // This is async, so we don't care about the return value.
+  XDGOpen(full_path.value(), false, std::move(callback));
```
deepak1556: Isn't the purpose of async version to extract error message from process exit value, why don't we care about exit value then ?

codebytere: this may have been unclear - by this i mean that we don't care about what `XDGOpen` returns; we're still calling back the exit code with that call.

deepak1556: `wait_for_exit` should be set to `true` since we depend on the `exit_code`, and simplify the above check

```
if (wait_for_exit) {
  ...
  bool success = process.WaitForExit(&exit_code);
  if (!callback.is_null())
   std::move(callback).Run(..);
  return success ? (exit_code == 0) : false;
```

shell/common/platform_util_linux.cc
```diff
@@ -13,13 +11,32 @@
 #include "base/process/kill.h"
 #include "base/process/launch.h"
 #include "chrome/browser/ui/libgtkui/gtk_util.h"
+#include "shell/common/platform_util.h"
```
deepak1556: nit: this shouldn't be moved, https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

anitakurz:
The release-notes of this breaking change says: Split `shell.openItem(path)` into synchronous and asynchronous methods.
This is misleading as `openItem` was replaced by `openPath`. The link to the spec in the first comment is also broken (not https://github.com/electron/governance/blob/master/wg-api/RFCs/shell-openitem.md anymore but https://github.com/electron/governance/blob/master/wg-api/spec-documents/shell-openitem.md). This is important as the API-documentation in this change is wrong and got corrected with https://github.com/electron/electron/pull/21641 (which is not linked anywhere).
Comming from https://www.electronjs.org/releases/stable#breaking-changes it is quite hard to find out what has been going on. I'd be glad if the release notes of a breaking change would state the nature of the change more clearly and/or outline the actions required for a successful upgrade.
Thanks for considering it.

PR #20684, opened 2019-10-22
fix: call InitializeContextRuntime in InitializeContext

codebytere:
#### Description of Change

Node added the following [commit](https://github.com/nodejs/node/commit/1ec4154e507ba71d7aefca0a6e5c155be34e989a) in `v12.10.0` that caused a crash in renderer process initialization becuase we don't directly call `NewContext`, and so `InitializeContext` would not create primordials correctly and the following error would occur:

```cpp
node::Environment::CreateProperties(): Assertion `primordials->IsObject()' failed.
```

cc @jkleinsc @ckerr @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash in Node.js initialization in renderer processes.


shell/renderer/atom_renderer_client.cc
```diff
@@ -122,8 +122,10 @@ void AtomRendererClient::DidCreateScriptContext(
 
   // Setup node environment for each window.
   DCHECK(node::InitializeContext(renderer_context));
-  node::Environment* env =
-      node_bindings_->CreateEnvironment(renderer_context, nullptr, true);
+  node::InitializeContextRuntime(renderer_context)
```
MarshallOfSound: This appears to be added by mistake

PR #20692, opened 2019-10-22
feat: enable builtin spellchecker

MarshallOfSound:
This does all the work to get the spellchecker working on all supported platforms.  Testing matrix I'm using is outlined below:

* [x] Windows 10
* [x] Windows 8.1
* [x] Windows 7
* [x] Ubuntu 18.04
* [x] Fedora
* [ ] RHEL
* [x] macOS 10.14
* [x] macOS 10.15

This is behind a build flag that is now **enabled** by default.  The addition of this introduces a few new APIs.

##### `session.setSpellCheckerLanguages(languageCodes: string[])`

This set the required pref that tells the spellchecker which languages it should use, example usage is.  `session.setSpellCheckLanguages(['en-US', 'fr'])` to enable french and american english.  This only impacts hunspell based platforms.

##### `session.availableSpellCheckerLanguages: string[]` (Readonly)

This is a read only property with an array of supported language codes that you can provide to the set method above.

##### `session.setSpellCheckerDictionaryDownloadURL(url: string)`

This is a writeable property that allows you to modify the base URL that Electron will use to download hunspell dictionaries from.  By default we will fetch them from the chromium provided CDN, if you don't want your app to do that you can host these yourself and modify this URL here.

For more info on these new APIs check out the new documentation.

#### Releasing

There is also a new published asset `hunspell_dictionaries.zip` which is published once (identical across platforms) that includes all the dictionary files and their licenses that you would need to host if you wanted to override the `dictionarySourceURL` property.

#### Existing APIs?

Personally I think we should remove the `webFrame.setSpellChecker` API as this supersedes in every way it but I'm open to other thoughts there.  For now I've added a `spellcheck` option in webPreferences to enable / disable the spellchecker.  While we make sure this is gonna work for everyone we can keep both code paths with the intention to deprecate and remove the webFrame method at some point in the future.

Notes: Added support for the built-in spellchecker.  We will use the OS spellchecker on macOS and hunspell on all other platforms.


buildflags/buildflags.gni
```diff
@@ -33,4 +33,7 @@ declare_args() {
 
   # Enable Chrome extensions support.
   enable_electron_extensions = false
+
+  # Enable Spellchecker support
+  enable_builtin_spellchecker = true
```
jkleinsc: @MarshallOfSound Looks like it is enabled by default but PR body says it will be disabled by default?

MarshallOfSound: I enabled it to test full CI

patches/chromium/remove_usage_of_incognito_apis_in_the_spellchecker.patch
```diff
@@ -0,0 +1,21 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <sattard@slack-corp.com>
+Date: Wed, 23 Oct 2019 11:43:58 -0700
+Subject: remove usage of incognito APIs in the spellchecker
+
```
codebytere: Can you explain the why behind this patch and if/when we can potentially remove it? (for posterity)

patches/chromium/use_electron_resources_not_chrome.patch
```diff
@@ -0,0 +1,32 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <sattard@slack-corp.com>
+Date: Wed, 23 Oct 2019 14:17:18 -0700
+Subject: use electron resources not chrome
+
```
codebytere: Can you explain the why behind this patch and if/when we can potentially remove it? (for posterity)

buildflags/buildflags.gni
MarshallOfSound: New plan, I want to turn this on for Realz in this PR 👍 

It's still behind a flag for easy removal / disable by other folks.

patches/chromium/remove_usage_of_incognito_apis_in_the_spellchecker.patch
MarshallOfSound: Explained

patches/chromium/use_electron_resources_not_chrome.patch
MarshallOfSound: Removed this patch

docs/api/session.md
```diff
@@ -456,10 +456,38 @@ this session just before normal `preload` scripts run.
...
+
+* `languages` String[] - An array of language codes to enable the spellchecker for.
+
+The built in spellchecker does not automatically detect what language a user is typing in.  In order for the
+spell checker to correctly check their words you must call this API with an array of language codes.  You can
+get the list of supported language codes with the `ses.availableSpellCheckerLanguages` property.
+
+#### `ses.getSpellCheckerLanguages()`
+
+Returns `String[]` asn array of language codes the spellchecker is enabled for.  If this list is empty the spellchecker
```
codebytere:
```suggestion
Returns `String[]` - an array of language codes the spellchecker is enabled for.  If this list is empty the spellchecker
```

BUILD.gn
```diff
@@ -478,6 +537,15 @@ source_set("electron_lib") {
     ]
   }
 
+  if (enable_builtin_spellchecker) {
+    deps += [
+      ":chrome_spellchecker",
+      "//components/spellcheck/browser",
```
deepak1556: `chrome_spellchecker` already depends on `//components/spellcheck`.

BUILD.gn
```diff
@@ -303,10 +305,15 @@ source_set("manifests") {
...
 
   deps = [
+    "buildflags",
     "//electron/shell/common/api:mojo",
     "//printing/buildflags",
     "//services/service_manager/public/cpp",
   ]
+
+  if (enable_builtin_spellchecker) {
+    deps += [ ":chrome_spellchecker" ]
```
deepak1556: Would be good to just declare the mojom `//components/spellcheck/common:interfaces` target rather than complete target, minimal dependency. Also manifest based services will go away soon so this will only be short lived https://bugs.chromium.org/p/chromium/issues/detail?id=977637 :)

BUILD.gn
```diff
@@ -335,6 +342,58 @@ templated_file("electron_version_header") {
...
+  if (use_browser_spellchecker) {
+    sources += [
+      "//chrome/browser/spellchecker/spelling_request.cc",
+      "//chrome/browser/spellchecker/spelling_request.h",
+    ]
+  }
+
+  public_configs = [
+    ":branding",
+    ":electron_lib_config",
```
deepak1556: Why does this need `electron_lib_config` ?

shell/browser/api/atom_api_app.cc
```diff
@@ -898,9 +898,14 @@ void App::SetPath(gin_helper::ErrorThrower thrower,
 
   bool succeed = false;
   int key = GetPathConstant(name);
-  if (key >= 0)
+  if (key >= 0) {
     succeed =
         base::PathService::OverrideAndCreateIfNeeded(key, path, true, false);
+    if (key == DIR_USER_DATA) {
+      succeed |= base::PathService::OverrideAndCreateIfNeeded(
+          chrome::DIR_USER_DATA, path, true, false);
```
deepak1556: why are we creating `chrome::DIR_USER_DATA` ?

shell/browser/api/atom_api_session.cc
```diff
@@ -646,6 +652,42 @@ void Session::Preconnect(const gin_helper::Dictionary& options,
...
+
+void SetSpellCheckerDictionaryDownloadURL(gin_helper::ErrorThrower thrower,
+                                          const GURL& url) {
+  if (!url.is_valid()) {
+    thrower.ThrowError(
+        "The URL you provided to setSpellCheckerDictionaryDownloadURL is not a "
+        "valid URL");
+    return;
+  }
+  SpellcheckHunspellDictionary::SetDownloadURLForTesting(url);
```
deepak1556: Can we avoid using a testing api ? maybe get a new one from upstream ?

shell/browser/api/atom_api_session.cc
```diff
@@ -646,6 +652,42 @@ void Session::Preconnect(const gin_helper::Dictionary& options,
...
+  for (const std::string& lang : languages) {
+    std::string code = spellcheck::GetCorrespondingSpellCheckLanguage(lang);
+    if (code.empty()) {
+      thrower.ThrowError("Invalid language code provided: \"" + lang +
+                         "\" is not a valid language code");
+      return;
+    }
+    language_codes.AppendString(code);
+  }
+  browser_context_->prefs()->Set(spellcheck::prefs::kSpellCheckDictionaries,
```
deepak1556: This could use a in memory preference store and flush to the file system on app shutdown.

shell/app/manifests.cc
```diff
@@ -22,3 +26,19 @@ const service_manager::Manifest& GetElectronContentBrowserOverlayManifest() {
...
+
+#if BUILDFLAG(ENABLE_BUILTIN_SPELLCHECKER)
+const service_manager::Manifest& GetElectronContentRendererOverlayManifest() {
+  static base::NoDestructor<service_manager::Manifest> manifest {
+    service_manager::ManifestBuilder()
+#if defined(OS_MACOSX)
+        .ExposeInterfaceFilterCapability_Deprecated(
+            "navigation:frame", "browser",
+            service_manager::Manifest::InterfaceList<
+                spellcheck::mojom::SpellCheckPanel>())
```
deepak1556: After looking through , I don't see the `SpellCheckPanel` initialized anywhere in the renderer , can you remove this dependency if not needed ?

miniak:
@MarshallOfSound how do you do spelling suggestions in the context menu? This was handled by getting those from hunspell directly in the `context-menu` event before.

BUILD.gn
MarshallOfSound: Yes but some files in this target depend on some spellcheck headers when this buildflag is enabled and `gn check` isn't smart enough to figure it out.  Adding this as a dep seemed cleaner than adding `NOCHECK` comments

BUILD.gn
MarshallOfSound: Include paths iirc

shell/browser/api/atom_api_app.cc
MarshallOfSound: Chrome's hunspell implementation uses it to determine the user data dir to download hunspell dictionaries to.  I just set it up to exactly mirror our user data preference.

shell/browser/api/atom_api_session.cc
MarshallOfSound: Using testing was easy.  I can file a CL upstream to get a proper setter if we want 👍

shell/app/manifests.cc
MarshallOfSound: Without it some internal logic crashes.  I think it _can_ get triggered from a menu item role on macOS.

MarshallOfSound:
@miniak Spelling suggestions are exposed in the `context-menu` event.  Looks like I didn't document that bit, I'll do that tomorrow 👍

shell/browser/api/atom_api_session.cc
MarshallOfSound: It's all tied to the prefs store we construct in AtomBrowserContext which has to be the syncable one to fulfil api requirements for the spellchecker classes.  Not sure how easy it is to make it half in memory, also worth noting that this prefs store is required for extensions to work as well (we just use the same logic)

BUILD.gn
deepak1556: `public_deps` are there to help for this scenario.

BUILD.gn
deepak1556: Hmm, why would a chrome sources set require include paths for `//electron` ?

shell/app/manifests.cc
deepak1556: Thats weird, this just provides access to the service interface for the renderer process and I don't see any use of this interface in here. Can you track down where it is coming from ?

shell/browser/api/atom_api_session.cc
deepak1556: Yup, in memory store can mock syncable, you can check the ones used for `session.setProxy`, we can flush it to file system if needed. But I see that you are editing the pref registry directly here, its fine to leave it as such.

BUILD.gn
MarshallOfSound: @deepak1556 That patch to include the electron grit header

shell/browser/api/atom_api_app.cc
MarshallOfSound: @deepak1556 That's what this does (makes the two values the same thing).  `chrome::DIR_USER_DATA` is `1002` afaics.  `electron::DIR_USER_DATA` is `11000`.

shell/app/manifests.cc
MarshallOfSound: I'll give that another look

shell/browser/api/atom_api_app.cc
MarshallOfSound: Resolved in DM.  This will be part of the whole "extract spellchecker from upstream into //component"

shell/app/manifests.h
```diff
@@ -7,6 +7,7 @@
 
 #include <vector>
 
+#include "electron/buildflags/buildflags.h"
```
deepak1556: Don't need this header.

BUILD.gn
```diff
@@ -303,10 +305,15 @@ source_set("manifests") {
   include_dirs = [ "//electron" ]
 
   deps = [
+    "buildflags",
```
deepak1556: Not required.

BUILD.gn
```diff
@@ -303,10 +305,15 @@ source_set("manifests") {
...
     "//electron/shell/common/api:mojo",
     "//printing/buildflags",
     "//services/service_manager/public/cpp",
   ]
+
+  if (enable_builtin_spellchecker) {
```
deepak1556: Not required.

astoilkov:
I was very happily surprised to see Electron doing significant improvements to the spellchecker. Congrats on the work!

@MarshallOfSound I have a quick question related to the future of spellchecking in Electron. Do you see a path to deprecating the need for [node-spellchecker](https://github.com/atom/node-spellchecker)?

We are building a text editor where we can't use the spellchecking of HTML elements because we need a more advanced implementation. We need methods to manually check if a text is misspelled.

dziudek:
@MarshallOfSound  - is the spellchecker feature working properly at this moment outside macOS? I have experimentally implement it in my electron-based app and it works only on macOS 10.15. On Windows 10 and Ubuntu linux 19.04 there are no visible results of the spellchecking :/ 

Initially I thought that I have to set language through `session.setSpellCheckerLanguages` (as it is dedicated to the hunspell platforms), but I have discovered that `session.getSpellCheckerLanguages` returns `['en-US']` at start, so it shouldn't be a reason. 

Also `session.availableSpellCheckerLanguages` returns 50 items, so the dictionaries seems to be available (and there is a new output directory called "Dictionaries" on Windows 10).

Tested on the newest beta: 8.0.0-beta.6

vladimiry:
> Existing APIs?

Please don't drop existing APIs support.

dziudek:
@MarshallOfSound - to avoid an impact of the third party reasons - I have created a super simple app based on this tutorial on Windows 10: https://electronjs.org/docs/tutorial/first-app

The only difference - I have add to the `webPreferences` option `spellcheck: true`

Unfortunately spellchecker is not working :( 

Screenshot 2020-01-15 at 21 15 23

aabuhijleh:
Adding to @dziudek.

I don't know If we're doing something wrong or if the spellchecker is not working properly on Windows.

I left a comment [here](https://github.com/electron/electron/pull/20897#issuecomment-574138357)

#### Using Windows 10 + Electron 8.0.0-beta.5 inside the simple quick start repo with minimal changes



I'm not getting anything.

### + Additional question

Using `session.setSpellCheckerLanguages`  => we can add a list of supported spellchecker languages.

Does that mean that this spellchecker will **automatically** detect the language the user is typing?

MarshallOfSound:
Quick thing folks, commenting on a closed and merged PR is not going to get people to really look at your problem.  If you think there's a bug in Electron, raise a new issue with detailed repro steps and it will be triaged appropriately.

PR #20732, opened 2019-10-25
fix: pass frameId to v8Util.setRemoteCallbackFreer()

miniak:
#### Description of Change
`RemoteCallbackFreer` was always sending the `ELECTRON_RENDERER_RELEASE_CALLBACK` message to the main frame instead of the source frame.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Fixed memory leaks caused by callbacks not being released when the `remote` module is used in sub-frames (`<iframe>` or scriptable popup).

### DIFF
--- a/lib/browser/remote/server.ts
+++ b/lib/browser/remote/server.ts
@@ -309,7 +309,7 @@ const unwrapArgs = function (sender: electron.WebContents, frameId: number, cont
         v8Util.setHiddenValue(callIntoRenderer, 'location', meta.location)
         Object.defineProperty(callIntoRenderer, 'length', { value: meta.length })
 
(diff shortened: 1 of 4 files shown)


miniak:
@nornagon I am not sure how to do that. `ELECTRON_RENDERER_RELEASE_CALLBACK` is an internal message

nornagon:
@miniak presumably this situation is observable from user code, otherwise we wouldn't need to fix it :)

miniak:
@nornagon it's similar to the previous issue, you could just observe a memory leak. But as we don't have weak references in JS yet, you cannot tell programmatically that the callbacks were released.

Issue #20737, opened 2019-10-25
crash in AtomMenuController.menuDidClose


jeremyspiegel:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  *  7.0.0, 6.1.2, 5.0.11
* **Operating System:**
  *  macOS 10.15
* **Last Known Working Electron version:**
  *  unknown

### Expected Behavior

When using the Menu.popup API, Electron should not crash in AtomMenuController.menuDidClose.

### Actual Behavior

When using the Menu.popup API, Electron will sometimes crash in AtomMenuController.menuDidClose.

### To Reproduce

Add this to main.js after creating mainWindow:
```javascript
  mainWindow.webContents.on('context-menu', () => {
    Menu.buildFromTemplate([{role: 'paste'}]).popup(mainWindow);
  });
```
Here is a gist with that change: https://gist.github.com/jeremyspiegel/53afd0127ec3fc3e1cad5bbef0e4c390

Then keep right-clicking to invoke the menu repeatedly until you see a crash.

You can make the crash happen reliably on the first time invoking the menu by forcing garbage collection — add `setTimeout(() => global.gc(), 0);` after running `Menu.popup`, and run Electron with `--js-flags="--expose-gc"` on the command line.

### Additional Information

@MarshallOfSound added a speculative fix in https://github.com/electron/electron/pull/20111 but it doesn't fix the issue.

The problem isn't that `model_` is null, but rather it appears to be that the `MenuMac` API object is destroyed while `MenuMac::PopupOnUI` is running (inside the `[menu popUpMenuPositioningItem:item atLocation:position inView:view]` call).

I added the following to the top of `MenuMac::PopupOnUI` to try to make it so that the object could not be garbage collected while in that function:
```c++
v8::HandleScope handle_scope(isolate());
GetWrapper();
```
That seemed to make the problem go away.

jeremyspiegel:
This only start coming up recently, so I think there is something about macOS 10.15 that triggers this (or is more likely to trigger this).

miniak:
The bug is not fixed

MarshallOfSound:
[...]

PR #20745, opened 2019-10-25
fix: prevent menu gc during popup

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/20737.

Ensures that the garbage collector will track the object stored in the handle and they won't get gc'd accidentally while they're open.

cc @MarshallOfSound @zcbenz 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed a crash in Menus related to `menu.popup()`.


codebytere:
@jkleinsc should be set now!

Issue #20242, opened 2019-09-16
macOS Catalina 10.15 beta 


roy2651:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description

I have updated my mac os to Catalina 10.15 beta , and test my electron app.
I share screen via eletron, as webrtc but  it does not work，and remote is black

### Proposed Solution

mac os Catalina 10.15 beta is new os featrue such as screen recorder authority，when I give authority to my electron app is ok。

### Alternatives Considered

is anybody has update mac os to beta，testing electron is ok，and if any soultion work this。

varisbhalala:
navigator.geolocation.getcurrentposition not working in electron mac os catalina

dwilson6:
I am seeing issues too with the desktopCapturer and screen capture on macOS Catalina. I've tried electron 5.0.8 and electron 7 beta 6. I know Catalina added a permission request for screen capture/recording and [the webrtc stack in Chromium had to be updated](https://groups.google.com/forum/#!msg/discuss-webrtc/C1T2wGEaakQ/agrPP_wMBQAJ) to account for it. 

The behavior I see in electron is that `desktopCapturer.getSources` and `navigator.mediaDevices.getUserMedia({ video: { mandatory: { chromeMediaSource: 'desktop' } } })` both resolve before the permission is granted or denied. If permission is denied (or not explicitly granted), the media stream captures the background and top bar and you can see the cursor moving. If permission is granted, the next stream you capture will capture correctly. Denying permission while a stream is active doesn't seem to affect the stream. Only subsequent requests for streams.

Here's a guide that shows what the UX for the permission request is like: https://dropshare.zendesk.com/hc/en-us/articles/360033453434-Enabling-Screen-Recording-Permission-on-macOS-Catalina-10-15-

Also, I haven't been able to find any reference to the actual api Apple added to wait for the permission request other than [this discussion](https://stackoverflow.com/questions/56597221/detecting-screen-recording-settings-on-macos-catalina). I guess there's also probably something in the webrtc source code.

dwilson6:
[...]

PR #20764, opened 2019-10-27
feat: add 'screen' to systemPreferences.getMediaAccessStatus()

miniak:
#### Description of Change
Adding `screen` to `systemPreferences.getMediaAccessStatus()` as macOS 10.15 Catalina requires consent for capturing the screen contents. Related to #20242.

This permission can only be granted manually in the System Preferences. Therefore `systemPreferences.askForMediaAccess()` cannot be extended in the same way.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `screen` to `systemPreferences.getMediaAccessStatus()` for detecting the new macOS Catalina permissions.

### DIFF
--- a/docs/api/system-preferences.md
+++ b/docs/api/system-preferences.md
@@ -434,11 +434,13 @@ Returns `Boolean` - `true` if the current process is a trusted accessibility cli
 
 ### `systemPreferences.getMediaAccessStatus(mediaType)` _macOS_
 
-* `mediaType` String - `microphone` or `camera`.
+* `mediaType` String - `microphone`, `camera` or `screen`.
 
 Returns `String` - Can be `not-determined`, `granted`, `denied`, `restricted` or `unknown`.
 
-This user consent was not required until macOS 10.14 Mojave, so this method will always return `granted` if your system is running 10.13 High Sierra or lower.
+This user consent was not required on macOS 10.13 High Sierra or lower so this method will always return `granted`.
+macOS 10.14 Mojave or higher requires consent for `microphone` and `camera` access.
+macOS 10.15 Catalina or higher requires consent for `screen` access.
 
 ### `systemPreferences.askForMediaAccess(mediaType)` _macOS_
 
--- a/docs/api/system-preferences.md
+++ b/docs/api/system-preferences.md
@@ -434,7 +434,7 @@ Returns `Boolean` - `true` if the current process is a trusted accessibility cli
 
 ### `systemPreferences.getMediaAccessStatus(mediaType)` _macOS_
 
-* `mediaType` String - `microphone`, `camera` or `screen`.
+* `mediaType` String - Can be `microphone`, `camera` or `screen`.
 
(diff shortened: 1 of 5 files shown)


nornagon:
Is a boolean the right return value for this API? Can we capture the difference between "screen capture allowed", "screen capture denied", and "screen capture permission not yet requested"? (Assuming the latter is a possible situation, I'm not too familiar with the Catalina APIs.) Even if that's not yet possible, it might be a good idea to leave the door open to more than 2 states in the future.

dwilson6:
> What is the proposed behavior / use case of this API for the failure / false case?
> 
> What can an app do when this returns false that it couldn't do before? What additional information does this provide that a rejected get media promise doesn't?

The getUserMedia promise is currently resolves if permissions aren't granted (at least on 7.x.x, I haven't tried on 8.x.x). The stream captures the user's background and the first frame of the app which initiated the capture and will capture the cursor's movement.

That said, getUserMedia rejecting if the permission isn't granted would make more sense.

> Is a boolean the right return value for this API? Can we capture the difference between "screen capture allowed", "screen capture denied", and "screen capture permission not yet requested"? (Assuming the latter is a possible situation, I'm not too familiar with the Catalina APIs.) Even if that's not yet possible, it might be a good idea to leave the door open to more than 2 states in the future.

Currently there are no proper apis in Catalina to determine the state of the screen capture permission. The code referenced by this PR uses a heuristic to determine if the permissions have been granted or not.

docs/api/system-preferences.md
```diff
@@ -450,6 +450,12 @@ Returns `Promise<Boolean>` - A promise that resolves with `true` if consent was
 
 This user consent was not required until macOS 10.14 Mojave, so this method will always return `true` if your system is running 10.13 High Sierra or lower.
 
+### `systemPreferences.isScreenCaptureAllowed()` _macOS_
```
loc: I think it'd be worth adding a link to this method from `systemPreferences.getMediaAccessStatus` and `desktopCapturer` to make things a little easier for folks and promote a proactive approach to solving this inevitable problem.

docs/api/system-preferences.md
```diff
@@ -434,11 +434,13 @@ Returns `Boolean` - `true` if the current process is a trusted accessibility cli
 
 ### `systemPreferences.getMediaAccessStatus(mediaType)` _macOS_
 
-* `mediaType` String - `microphone` or `camera`.
+* `mediaType` String - `microphone`, `camera` or `screen`.
```
MarshallOfSound:
```suggestion
* `mediaType` String - Can be `microphone`, `camera` or `screen`.
```

dwilson6:
Could this change be backported to 7-x-y as well so it can be used in an already stable version of electron? I believe the changes used here were present in chromium 77

CapOM:
Could this change be back ported to 4-2-x branch as well ? Thx!

PR #20775, opened 2019-10-28
build: disable `strip_absolute_paths_from_debug_symbols` on debug.gn

ymmuse:
#### Description of Change
We can't set breakpoints directly in the Chromium source on XCode, Because from [this commit](https://chromium.googlesource.com/chromium/src/+/87a677dfe577d9411be85d08b466a03f53c480c5%5E%21/#F0) it was set strip_absolute_paths_from_debug_symbols default to true. This PR make strip_absolute_paths_from_debug_symbols to false in debug/build.gn.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Make strip_absolute_paths_from_debug_symbols to false in debug.gn

### DIFF
--- a/build/args/debug.gn
+++ b/build/args/debug.gn
@@ -2,6 +2,8 @@ import("all.gn")
 is_debug = true
 is_component_build = true
 
+strip_absolute_paths_from_debug_symbols = false
+
 # This may be guarded behind is_chrome_branded alongside
 # proprietary_codecs https://webrtc-review.googlesource.com/c/src/+/36321,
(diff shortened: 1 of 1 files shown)


miniak:
FYI @alexeykuzmin

deepak1556:
Shouldn't be an issue with only enabling for debug config, but there is also https://chromium.googlesource.com/chromium/src/+/lkgr/docs/lldbinit.md to help when this flag is enabled by default. So I am neutral on this change.

PR #20789, opened 2019-10-28

MarshallOfSound:
Backport of #20307 see that PR for details.


shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -0,0 +1,515 @@
...
+
+#include <set>
+#include <utility>
+#include <vector>
+
+#include "base/no_destructor.h"
+#include "base/strings/string_number_conversions.h"
+#include "shell/common/api/object_life_monitor.h"
+#include "shell/common/native_mate_converters/blink_converter.h"
+#include "shell/common/native_mate_converters/callback_converter_deprecated.h"
```
zcbenz: Can be removed as `once_callback.h` is already included.

PR #20794, opened 2019-10-28
test: there is only --ci

nornagon:
#### Description of Change
This removes the option to run the remote tests _without_ `--ci`, i.e., the only option now is to run the tests the same as in CI.

The `isCI` variable was often used to disable tests running in CI, but leave them enabled when run locally. I claim that a test that doesn't run in CI might as well not exist. Further, no current tests in the `spec/` directory rely on the variable.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


spec-main/api-menu-spec.ts
```diff
@@ -840,97 +840,4 @@ describe('Menu module', function () {
       expect(Menu.getApplicationMenu()).to.be.null('application menu')
     })
   })
-
-  describe('menu accelerators', async () => {
```
codebytere: can we note somewhere that we should _try_ to test this on CI

PR #20797, opened 2019-10-29
refactor: try just using regular [Sync] for MessageSync

nornagon:
#### Description of Change
This aligns us closer with how Mojo expects to work, and as a bonus fixes some tricky race conditions that can happen if the browser process closes its end of the mojo pipe while we're waiting for a synchronous response.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/lib/browser/api/browser-window.js
+++ b/lib/browser/api/browser-window.js
@@ -116,7 +116,7 @@ BrowserWindow.getFocusedWindow = () => {
 
 BrowserWindow.fromWebContents = (webContents) => {
   for (const window of BrowserWindow.getAllWindows()) {
-    if (window.webContents.equal(webContents)) return window
+    if (window.webContents && window.webContents.equal(webContents)) return window
   }
 
   return null
--- a/shell/common/api/api.mojom
+++ b/shell/common/api/api.mojom
@@ -55,10 +55,7 @@ interface ElectronBrowser {
 
   // Emits an event on |channel| from the ipcMain JavaScript object in the main
   // process, and waits synchronously for a response.
-  //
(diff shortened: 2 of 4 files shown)


nornagon:
This seems to be passing CI... could it be that we've removed our need for synchronous IPC to be well-ordered w.r.t asynchronous IPC?

nornagon:
Hm, I don't think the scenario you laid out would be a problem in the usual case. The only time that sync calls will be reordered with respect to async calls is if the main process is waiting on a sync call, which [almost never happens](https://source.chromium.org/chromium/chromium/src/+/master:mojo/public/cpp/bindings/sync_call_restrictions.h;l=71-85;drc=f58dc6a1dc47d88e2736b2a40ef9e32b30725d60?originalUrl=https:%2F%2Fcs.chromium.org%2F). We previously observed some issues relating to window close, because there's a [compositor shutdown call that the browser invokes synchronously](https://source.chromium.org/chromium/chromium/src/+/master:components/viz/host/host_frame_sink_manager.cc;l=105;drc=060a4d929a2a6dcebe7dfdc33a45d8989f41b6cb?originalUrl=https:%2F%2Fcs.chromium.org%2F) during window close. If I recall correctly, it was because we were triggering synchronous calls during window close to release objects, which I don't think is the case any more? But you're right that we're exercising the remote module a lot less.

nornagon:
Let's merge this and see how it goes. We can always revert if we discover issues.

miniak:
@nornagon what about `8-x-y`?

nornagon:
it should.

@MarshallOfSound want to do the honors?

PR #20809, opened 2019-10-29
refactor: pass base::Value by value in JS API implementations 

zcbenz:
#### Description of Change

There are a few cases that the passed `base::Value` has to be cloned because they are passed by const references. 

For JS API implementations, the callees actually own the passed `base::Value`s, and by passing them by value we can eliminate unnecessary clones.

Note that `base::Value` is move-only so there would be no accidental copies.

This PR is needed for removing some native_mate classes.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/shell/browser/browser_mac.mm
+++ b/shell/browser/browser_mac.mm
@@ -141,7 +141,7 @@ bool Browser::SetBadgeCount(int count) {
 }
 
 void Browser::SetUserActivity(const std::string& type,
-                              const base::DictionaryValue& user_info,
+                              base::DictionaryValue user_info,
                               gin_helper::Arguments* args) {
   std::string url_string;
   args->GetNext(&url_string);
@@ -167,7 +167,7 @@ void Browser::ResignCurrentActivity() {
 }
 
 void Browser::UpdateCurrentActivity(const std::string& type,
-                                    const base::DictionaryValue& user_info) {
+                                    base::DictionaryValue user_info) {
   [[AtomApplication sharedApplication]
       updateCurrentActivity:base::SysUTF8ToNSString(type)
                withUserInfo:DictionaryValueToNSDictionary(user_info)];
@@ -187,7 +187,7 @@ void Browser::DidFailToContinueUserActivity(const std::string& type,
 }
 
 bool Browser::ContinueUserActivity(const std::string& type,
-                                   const base::DictionaryValue& user_info) {
+                                   base::DictionaryValue user_info) {
   bool prevent_default = false;
   for (BrowserObserver& observer : observers_)
     observer.OnContinueUserActivity(&prevent_default, type, user_info);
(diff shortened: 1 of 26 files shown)


shell/browser/browser_mac.mm
```diff
@@ -373,15 +373,15 @@ void RemoveFromLoginItems() {
...
-void Browser::SetAboutPanelOptions(const base::DictionaryValue& options) {
+void Browser::SetAboutPanelOptions(base::DictionaryValue options) {
   about_panel_options_.Clear();
 
-  for (const auto& pair : options) {
-    std::string key = pair.first;
-    const auto& val = pair.second;
+  for (auto& pair : options) {
+    std::string& key = pair.first;
+    std::unique_ptr<base::Value>& val = pair.second;
```
deepak1556: Any reason for this to be a reference since we are eventually moving it ?

shell/common/native_mate_converters/value_converter.cc
```diff
@@ -39,7 +40,7 @@ bool Converter<base::Value>::FromV8(v8::Isolate* isolate,
   std::unique_ptr<base::Value> value(
       converter.FromV8Value(val, isolate->GetCurrentContext()));
```
deepak1556: Can we avoid creating a unique_ptr here just for temporary storage ?

shell/common/gin_helper/function_template.h
```diff
@@ -225,7 +226,7 @@ class Invoker<IndicesHolder<indices...>, ArgTypes...>
   void DispatchToCallback(base::Callback<void(ArgTypes...)> callback) {
     v8::MicrotasksScope script_scope(args_->isolate(),
                                      v8::MicrotasksScope::kRunMicrotasks);
-    callback.Run(ArgumentHolder<indices, ArgTypes>::value...);
+    callback.Run(std::move(ArgumentHolder<indices, ArgTypes>::value)...);
```
deepak1556: Will this be future proof ? Expecting all arguments to be movable ?

shell/browser/browser_mac.mm
zcbenz: Didn't realize this is unneeded. 👍

shell/common/gin_helper/function_template.h
zcbenz: If the parameter is a const reference, the `ArgTypes&&` would be converted to `const ArgTypes&` automatically.

If the parameter is a value, the `ArgTypes&&` would be converted to `ArgTypes` and C++ would try to invoke the copy constructor.

The only difference is when the parameter is a rvalue reference or when it is a value but has move constructor, and in this case move constructor would be used. So this should be future proof.

shell/common/native_mate_converters/value_converter.cc
zcbenz: The `V8ValueConverter::FromV8Value` returns `std::unique_ptr` as result so it can not be avoided.

I tried to refactor `V8ValueConverter` to make it return `base::Value` directly, but it is then impossible to distinguish between conversion failures and null results.

PR #20829, opened 2019-10-29
fix: Disable compositor recycling only for attached views

poiru:
In #19873, we completely disabled compositor recycling. This has adverse
effects in our tabbed app where switching tabs (i.e. `BrowserView`s) now
results in a flicker because we now also switch compositors.

To fix this without regressing the original fix, we now recycle the
compositor when the view is removed from a window or explicitly marked
as hidden. These can only happen with `BrowserView`s so `BrowserWindow`s
are unaffected.

### DIFF
--- a/patches/chromium/disable_compositor_recycling.patch
+++ b/patches/chromium/disable_compositor_recycling.patch
@@ -5,16 +5,20 @@ Subject: fix: disabling compositor recycling
 
 Compositor recycling is useful for Chrome because there can be many tabs and spinning up a compositor for each one would be costly. In practice, Chrome uses the parent compositor code path of browser_compositor_view_mac.mm; the NSView of each tab is detached when it's hidden and attached when it's shown. For Electron, there is no parent compositor, so we're forced into the "own compositor" code path, which seems to be non-optimal and pretty ruthless in terms of the release of resources. Electron has no real concept of multiple tabs per window, so it should be okay to disable this ruthless recycling altogether in Electron.
 
-diff --git a/content/browser/renderer_host/browser_compositor_view_mac.mm b/content/browser/renderer_host/browser_compositor_view_mac.mm
-index 8ddd790decc43af9820c97121a3b359e7cbb49ee..41269301e92d96757066229842333c8981994234 100644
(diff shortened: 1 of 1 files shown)


patches/chromium/disable_compositor_recycling.patch
```diff
@@ -1,20 +1,25 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
-From: Andy Locascio <andy@slack-corp.com>
-Date: Wed, 21 Aug 2019 12:09:10 -0700
+From: Birunthan Mohanathas <birunthan@mohanathas.com>
```
deepak1556: Can you comma separate the authors ?

poiru: I reverted this bit.

loc:
This regresses the original fix. With the original fix, the RenderWidgetHost is never hidden as far as the compositor is concerned, whereas your patch allows it to be hidden if the NSView is hidden. I think the logic we actually want is more like:
```
render_widget_host_is_hidden_ = [GetInProcessNSView() window] ? false : hidden;
```
with `hidden` being the passed bool.

poiru:
@loc I actually just fixed that albeit in a slightly different way 😊

With the updated patch and this logging statement...

```patch
 void BrowserCompositorMac::SetRenderWidgetHostIsHidden(bool hidden) {
+  LOG(ERROR) << (hidden ? "COMPOSITOR HIDDEN" : "COMPOSITOR SHOWN");
   render_widget_host_is_hidden_ = hidden;
   UpdateState();
 }
```

... the results are as follows:

* Minimizing and restoring window

```
COMPOSITOR SHOWN
COMPOSITOR SHOWN
```

* Occluding and unoccluding window

```
COMPOSITOR SHOWN
COMPOSITOR SHOWN
```

* Removing a BrowserView from a window and adding it back

```
COMPOSITOR HIDDEN
COMPOSITOR SHOWN
```

So I think we should be good to go!

poiru:
>Edit: Oh I see you just pushed changes. Now it won't regress, but I still think we want to set it to the passed bool for the BrowserView case?

We do set it for the BrowserView case, see my reply above!

PR #20833, opened 2019-10-29
fix: do not DCHECK non-const methods

codebytere:
#### Description of Change

`InitializeContext` was not being properly called in release builds due to the nature of DCHECKS.

cc @MarshallOfSound @ckerr 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue with Node.js context initialization in renderer processes.

### DIFF
--- a/shell/renderer/atom_renderer_client.cc
+++ b/shell/renderer/atom_renderer_client.cc
@@ -121,7 +121,9 @@ void AtomRendererClient::DidCreateScriptContext(
     node::tracing::TraceEventHelper::SetAgent(node::CreateAgent());
 
   // Setup node environment for each window.
-  DCHECK(node::InitializeContext(renderer_context));
+  bool initialized = node::InitializeContext(renderer_context);
+  CHECK(initialized);
+
   node::Environment* env =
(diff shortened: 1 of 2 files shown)


deepak1556:
Thanks for fixing this!

But the title and commit message are misleading, `DCHECK` in non debug builds noops `std::ostream` with `operator<<` which doesn't evaluate the conditions. How would having `const` methods change this ?

```
std::ostream* g_swallow_stream;

class LogMessageVoidify {
 public:
  LogMessageVoidify() = default;
  // This has to be an operator with a precedence lower than << but
  // higher than ?:
  void operator&(std::ostream& g) {  }
};

#define EAT_STREAM_PARAMETERS \
  true ? (void)0              \
       : LogMessageVoidify() & (*g_swallow_stream)
       
#define DCHECK(condition) EAT_STREAM_PARAMETERS << !(condition)

```

codebytere:
oh yes sorry there's some missing context - we want to make it in a future PR so that we lint against non-const methods being used in DCHECKs we know it doesn't work now :)

PR #20834, opened 2019-10-29
fix: Disable compositor recycling only for attached views (6-1-x)

poiru:
Backport of #20829


jkleinsc:
Merging as CI failure unrelated to PR change

PR #20846, opened 2019-10-30
fix: Disable compositor recycling only for attached views (7-0-x)

poiru:
Backport of #20829


jkleinsc:

PR #20849, opened 2019-10-30
chore: properly dispose of isolate

codebytere:
#### Description of Change

We should be ensuring that we're disposing of the Isolate properly when we're finished with it.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


codebytere:
Hold on this one; i think we may have to do it elsewhere?

Issue #20831, opened 2019-10-29
There is no way to set `--max-http-header-size` in Electron 5.x


flotwig:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.10
* **Operating System:**
  * all
* **Last Known Working Electron version:**
  * n/a

### Issue Summary

* Node 11.11.0 introduced a `--max-http-header-size` config option, to change `http.maxHeaderSize` (default 80kb)
* In Electron 5, Node 12.0.0 was introduced, which made the default `http.maxHeaderSize` only 8kb
* This made some of our users' tests fail with Electron 5, since the header size changed:
	* cypress-io/cypress#5431
* We want to increase the header size by doing `--max-http-header-size={1 MB}` at Electron startup time
	* However, Electron does not accept this command line argument. If you try passing it to an unpackaged app or a packaged app, the result is the same, `http.maxHeaderSize` is still 8kb.
	* You can pass it as a `NODE_OPTIONS`, and this works great. We updated our Electron launcher to pass this because we thought this was the solution: 
		* cypress-io/cypress#5452, cypress-io/cypress#5492
	* However, this only works in development. As the docs note:
		> https://electronjs.org/docs/all#node_options
		> NODE_OPTIONS are explicitly disallowed in packaged apps.
* So, it appears there is no real way to set `--max-http-header-size` in Electron. Not sure if this counts as a bug, or as a feature request, but either way, a working method to set this flag would be much appreciated!

P.S. thanks for all the work you do on Electron, the fact that this is the first issue I've had to open after 10 months of working with your product is a testament to its quality.

### Related

* #13402 - "Electron ignores NODE_OPTIONS"

codebytere:
Closing this as i've added it, but per our API policies it will only be available in newer versions of Electron.

flotwig:
@codebytere Awesome, thanks! Do you know when this will be released? I don't see it in the changelog for any of the versions since it was merged on October 30th.

flotwig:
[...]

PR #20857, opened 2019-10-30
feat: allow some NODE_OPTIONS in packaged apps

codebytere:
#### Description of Change

Addresses https://github.com/electron/electron/issues/20831.

Refactors our `NODE_OPTIONS` parsing to allow, for now, `--max-http-header-size` to be passed in packaged apps with the future potential for more specific options as well. Splits out handling into two encapsulated functions for cleanliness. I don't believe that there are security implications to this specific argument being whitelisted, but I would like to get clearance from @electron/wg-security just to be sure.

cc @deepak1556 @MarshallOfSound @ckerr 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Allowed some specific options to be passed via NODE_OPTIONS to packaged apps.

### DIFF
--- a/shell/common/node_bindings.cc
+++ b/shell/common/node_bindings.cc
@@ -126,12 +126,13 @@ bool g_is_initialized = false;
 bool IsPackagedApp() {
   base::FilePath exe_path;
   base::PathService::Get(base::FILE_EXE, &exe_path);
-  base::FilePath::StringType path = base::ToLowerASCII(exe_path.value());
+  base::FilePath::StringType base_name =
+      base::ToLowerASCII(exe_path.BaseName().value());
 
 #if defined(OS_WIN)
-  return path == FILE_PATH_LITERAL("electron.exe");
+  return base_name != FILE_PATH_LITERAL("electron.exe");
 #else
-  return path == FILE_PATH_LITERAL("electron");
+  return base_name != FILE_PATH_LITERAL("electron");
 #endif
 }
 
@@ -151,11 +152,13 @@ void SetNodeOptions(base::Environment* env) {
     std::vector<std::string> parts = base::SplitString(
         options, " ", base::TRIM_WHITESPACE, base::SPLIT_WANT_NONEMPTY);
 
+    bool is_packaged_app = IsPackagedApp();
+
     for (const auto& part : parts) {
       // Strip off values passed to individual NODE_OPTIONs
       std::string option = part.substr(0, part.find("="));
 
-      if (IsPackagedApp() &&
+      if (is_packaged_app &&
           allowed_in_packaged.find(option) == allowed_in_packaged.end()) {
         // Explicitly disallow majority of NODE_OPTIONS in packaged apps
         LOG(ERROR) << "Most NODE_OPTIONs are not supported in packaged apps."
--- a/shell/common/node_bindings.cc
+++ b/shell/common/node_bindings.cc
@@ -6,6 +6,7 @@
 
 #include <algorithm>
 #include <memory>
+#include <set>
 #include <string>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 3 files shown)


shell/common/node_bindings.cc
```diff
@@ -122,6 +123,67 @@ void stop_and_close_uv_loop(uv_loop_t* loop) {
 
 bool g_is_initialized = false;
 
+bool IsPackagedApp() {
+  base::FilePath exe_path;
+  base::PathService::Get(base::FILE_EXE, &exe_path);
+
+#if defined(OS_WIN)
+  std::string path = base::UTF16ToUTF8(exe_path.value());
```
zcbenz: There is no need for conversion here:

```c++
base::FilePath::StringType path = exe_path.value();
path = base::ToLowerASCII(path);
#if defined(OS_WIN)
  return path == L"electron.exe";
#else
  return path == "electron";
#endif
```

shell/common/node_bindings.cc
```diff
@@ -122,6 +123,67 @@ void stop_and_close_uv_loop(uv_loop_t* loop) {
...
+#else
+  std::string path = exe_path.value();
+#endif
+
+  std::transform(path.begin(), path.end(), path.begin(), ::tolower);
```
zcbenz: There is `base::ToLowerASCII`.

shell/common/node_bindings.cc
```diff
@@ -122,6 +123,67 @@ void stop_and_close_uv_loop(uv_loop_t* loop) {
...
+
+#if defined(OS_WIN)
+  return path == "electron.exe";
```
zcbenz: I believe this should be `!=`.

shell/common/node_bindings.cc
```diff
@@ -122,6 +123,67 @@ void stop_and_close_uv_loop(uv_loop_t* loop) {
...
+  if (env->HasVar("NODE_OPTIONS")) {
+    std::string options;
+    env->GetVar("NODE_OPTIONS", &options);
+    std::vector<std::string> parts = base::SplitString(
+        options, " ", base::TRIM_WHITESPACE, base::SPLIT_WANT_NONEMPTY);
+
+    std::string new_options = options;
+    for (const auto& part : parts) {
+      // Strip off values passed to individual NODE_OPTIONs
+      std::string option = part.substr(0, part.find("="));
```
zcbenz: `part.find` might return `std::npos`.

codebytere: right; i think that's what we want here, no? if there's no value, we'll just be passing the flag regardless

shell/common/node_bindings.cc
codebytere: we want this to return `true` when it _is_ packaged though, so i think this is what we want to check?

zcbenz: I think when app is packaged the executable name should not be `electron`? The `App.isPackaged` is implemented so.

shell/common/node_bindings.cc
zcbenz: It appears `substr` does accepts `npos`, thanks for testing.

codebytere:
i'll try to add tests but am not 100% sure the best way at the moment since an error would just no-op and log a warning 🤔

zcbenz:
For unpackaged app you can just run it via childProcess and check its output, there are some examples in the tests of `--inspect`.

I don't have a good idea for packaged app, it is probably fine leave them for now.

shell/common/node_bindings.cc
```diff
@@ -122,6 +123,60 @@ void stop_and_close_uv_loop(uv_loop_t* loop) {
...
+  // Subset of options allowed in packaged apps
+  const std::set<std::string> allowed_in_packaged = {"--max-http-header-size"};
+
+  if (env->HasVar("NODE_OPTIONS")) {
+    std::string options;
+    env->GetVar("NODE_OPTIONS", &options);
+    std::vector<std::string> parts = base::SplitString(
+        options, " ", base::TRIM_WHITESPACE, base::SPLIT_WANT_NONEMPTY);
+
+    std::string new_options = options;
```
deepak1556: `options` doesn't seem to be used anywhere other than for copying here, can instead use it directly and avoid the copy.

spec-main/node-spec.ts
```diff
@@ -8,7 +8,7 @@ import { webContents, WebContents } from 'electron';
 
 const features = process.electronBinding('features')
 
-describe('node feature', () => {
+describe.only('node feature', () => {
```
deepak1556: Accidental `.only` ?

codebytere: ooops yes will remove

shell/common/node_bindings.cc
codebytere: @deepak1556:
```
    // overwrite new NODE_OPTIONS without unsupported variables
    if (new_options != options)
      env->SetVar("NODE_OPTIONS", new_options);
```
needs that i believe?

deepak1556: That check seems unnecessary, whats the harm in setting the options irrespective of the check.

codebytere: ah ok! will do 🙇‍♀

shell/common/node_bindings.cc
```diff
@@ -122,6 +123,58 @@ void stop_and_close_uv_loop(uv_loop_t* loop) {
...
 
+bool IsPackagedApp() {
+  base::FilePath exe_path;
+  base::PathService::Get(base::FILE_EXE, &exe_path);
+  base::FilePath::StringType path = base::ToLowerASCII(exe_path.value());
+
+#if defined(OS_WIN)
+  return path != FILE_PATH_LITERAL("electron.exe");
+#else
+  return path != FILE_PATH_LITERAL("electron");
```
deepak1556: Wouldn't this return the path to helper executable on macOS, so this comparison will not be valid then.

codebytere:
@deepak1556 fixed that - i'm honestly not sure why CI is failing since it passes perfectly for me locally on macOS for both new specs 🤔

shell/common/node_bindings.cc
```diff
@@ -122,6 +123,58 @@ void stop_and_close_uv_loop(uv_loop_t* loop) {
...
 
+bool IsPackagedApp() {
+  base::FilePath exe_path;
+  base::PathService::Get(base::FILE_EXE, &exe_path);
+  base::FilePath::StringType path = base::ToLowerASCII(exe_path.value());
+
+#if defined(OS_WIN)
+  return path == FILE_PATH_LITERAL("electron.exe");
+#else
+  return path == FILE_PATH_LITERAL("electron");
```
deepak1556: Its still not sufficient, helper executables end with `Electron Helper` etc, in this case `Electron Helper (Renderer)` comparing to `electron` wouldn't be right. The tests we have right now will pass even for this condition because we pass unilaterally invalid arguments in the test case. This issue will show up when the app is packaged.

zcbenz: Should be `!=` here, the change seems to be reverted back when resetting.

codebytere: @zcbenz if i make if the other way around (the way you're suggesting), both specs fail, meaning it's returning packaged when it shouldn't be 🤔

zcbenz:
I have pushed a fix which should be able to fix the tests.

PR #20867, opened 2019-10-31
refactor: remove references to non-existent webui

nornagon:
#### Description of Change
This removes references to the PdfViewerUI class that was removed a while ago. Chromium now uses an extension to show PDFs, and if/when we eventually re-enable PDF support, we'll do it through //extensions.

This also lays some preliminary groundwork for allowing the application/x-google-chrome-pdf plugin to be loaded by the extension in future.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


BUILD.gn
```diff
@@ -675,6 +676,87 @@ source_set("electron_lib") {
...
+  if (enable_pdf_viewer) {
+    deps += [
+      "//components/pdf/browser",
+      "//components/pdf/renderer",
+      "//pdf",
+    ]
+  }
+}
+
+source_set("plugins") {
```
deepak1556: Can you move this to `chromium_src/BUILD.gn`, better to keep all `//chrome` references in one place. Maybe it could now be `chromium_src` => `chrome_src`

patches/chromium/pepper_plugin_support.patch
```diff
@@ -1,9 +1,10 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: deepak1556 <hop2deep@gmail.com>
-Date: Thu, 20 Sep 2018 17:48:09 -0700
-Subject: pepper_flash.patch
+Date: Thu, 20 Sep 2018 17:46:17 -0700
+Subject: pepper plugin support
 
-Allows building chrome pepper flash integration for Electron.
+This tweaks Chrome's pepper flash and PDF plugin support to make it
```
deepak1556: Since the flash sources are now built as `source_set`, the component build specific patch is not needed

`class COMPONENT_EXPORT(PEPPER_FLASH) ChromeBrowserPepperHostFactory : public ppapi::host::HostFactory` => `class ChromeBrowserPepperHostFactory : public ppapi::host::HostFactory`

shell/app/atom_content_client.cc
```diff
@@ -32,9 +32,10 @@
 #endif  // defined(WIDEVINE_CDM_AVAILABLE)
 
 #if BUILDFLAG(ENABLE_PDF_VIEWER)
-#include "pdf/pdf.h"
-#include "shell/common/atom_constants.h"
-#endif  // BUILDFLAG(ENABLE_PDF_VIEWER)
+#include "pdf/pdf.h"                      // nogncheck
```
deepak1556: why `//nogncheck` ?

shell/browser/common_web_contents_delegate.cc
```diff
@@ -184,6 +189,24 @@ CommonWebContentsDelegate::CommonWebContentsDelegate()
 
 CommonWebContentsDelegate::~CommonWebContentsDelegate() = default;
 
+#if BUILDFLAG(ENABLE_PDF_VIEWER)
+class ElectronPDFWebContentsHelperClient
```
deepak1556: Can you move this to separate file under `shell/browser/pdf` ?

shell/renderer/pepper_helper.cc
```diff
@@ -0,0 +1,33 @@
...
+
+PepperHelper::~PepperHelper() {}
+
+void PepperHelper::DidCreatePepperPlugin(content::RendererPpapiHost* host) {
+  // TODO(brettw) figure out how to hook up the host factory. It needs some
+  // kind of filter-like system to allow dynamic additions.
+  host->GetPpapiHost()->AddHostFactoryFilter(
+      std::make_unique<ChromeRendererPepperHostFactory>(host));
+  host->GetPpapiHost()->AddHostFactoryFilter(
+      std::make_unique<ElectronRendererPepperHostFactory>(host));
```
deepak1556: What is `ElectronRendererPepperHostFactory` gonna handle that is not handled by the chrome couterpart ?

shell/app/atom_content_client.cc
nornagon: beacuse gn check doesn't understand `#if`

shell/renderer/pepper_helper.cc
nornagon: UMA blackholing

shell/app/atom_content_client.cc
deepak1556: It should iirc if you have these as `public_deps` , I remember doing it for other guarded sources we have in `chromium_src/BUILD.gn`

shell/renderer/pepper_helper.cc
deepak1556: We already do that in `pepper_plugin_support.patch`, is the plan to reduce the patch surface ?

shell/app/atom_content_client.cc
nornagon: we already have the deps, they're just conditional in the gn file based on the `enable_pdf_viewer` build flag. i suppose i could contort to do it by putting this stuff in a separate target, but i don't think it's worth it.

shell/browser/common_web_contents_delegate.cc
nornagon: There's no `pdf/` directory currently, do you think this warrants a new dir?

deepak1556: Yeah its better, future implementations can go in there, avoids polluting this file.

shell/renderer/pepper_helper.cc
nornagon: that patch doesn't blackhole UMA, it just doesn't create the UMA resource host at all. That ends up resulting in a DCHECK() when pepper plugins try to load the UMA interface (which the PDF plugin does). The UMA host in chrome relies on a bunch of chrome-specific stuff which I wanted to avoid, so building our own stub version seemed like the best way to go. An alternative path would be to patch the PDF plugin to not call any UMA hooks.

deepak1556: Thanks for the context! in that case, this is fine as it is :)

patches/chromium/pepper_plugin_support.patch
nornagon: done

PR #20886, opened 2019-10-31
refactor: add Error to isSerializableObject()

miniak:
#### Description of Change
When sending `Error` from main to renderer in the `remote` we handle it first to preserve custom properties, so this change is safe:
https://github.com/electron/electron/blob/f0b0614dd79f741b5018c3187a438c24571c722e/lib/browser/remote/server.ts#L129-L132

Follow-up to #20427

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes


miniak:

PR #20897, opened 2019-10-31
feat: enable builtin spellchecker

MarshallOfSound:
Backport of #20692

Only difference is the default value of the `spellcheck` webPreference is `false `(disabled by default in 8-x-y as per discussion)

Notes: Added support for the built-in spellchecker.  We will use the OS spellchecker on macOS and hunspell on all other platforms.


docs/api/web-frame.md
```diff
@@ -74,6 +74,17 @@ Sets the maximum and minimum layout-based (i.e. non-visual) zoom level.
 
 Sets a provider for spell checking in input fields and text areas.
 
+If you want to use this method you must disable the builtin spellchecker when you
```
jkleinsc: Is this section needed in 8-x-y since the default of `spellcheck` is already `false`?

MarshallOfSound: Left it in to be clear that the two settings are incompatible.

aabuhijleh:
Hi @MarshallOfSound,

Thanks for this PR.

I'm a little confused on how this new built-in spellchecker works

This is what I'm doing (Windows 10)
image

Misspelled words are not highlighted with red lines. Can you please clear up how to use this spellchecker?

tallev266:
Adding to @aabuhijleh .
I don't get anything in the 'dictionarySuggestions' nor in the 'misspelledWord'

Do we need to install something on windows in order for it to work?

dziudek:
@aabuhijleh @tallev266 - you were probably affected by this issue: https://github.com/electron/electron/issues/21798 but in stable release of Electron v.8.0 spellchecker is working fine :)

sergeyd63:
not working in stable 8.0..0

eteisman:
I am trying to upgrade my application to use Electron 8.0.0,

The spellchecker does not seem to work. I do not see any spelling errors being marked.

adventofryz:
Okay, I love the fact that we have native spellchecker but I honestly have no clue how to use it. I see the red underline on an incorrect word and naturally, I right-click it to see the context menu. No context menu appears. 
Okay, so I guess I need to build one by myself. But where do I get the **suggestions** to display from the session spellchecker? I read the documentation and could not find any mention of it whatsoever. 
So my question is: Do we actually build a suggestion menu (can we?) or is there a built-in tool to bring up a context menu on right-clicking the spellchecked word.

PR #15596, opened 2018-11-06
fix: bypass DOM storage quota

jacobq:
#### Description of Change
This updates the `dom_storage_map` patch to include bypassing the 2 additional locations of quota checking. It also adds a test to ensure that at least 50MiB of data can be stored in `localStorage` and `sessionStorage` (I'm surprised that I didn't find something like that already in place. Perhaps that's how the regression sneaked in.) Since the usual limit is about 10MiB the test will fail if the patch has not been applied or is not working.
Fixes #8337.

cc @MarshallOfSound you [mentioned](https://github.com/electron/electron/issues/897#issuecomment-412779881) you might be willing to help land this. (Thanks in advance!)
cc @zcbenz It looks like you [wrote the original patch](https://github.com/electron/electron/issues/897#issuecomment-66856924). (Thank you! That helped me understand how electron's build system applies chromium patches)
cc @asinbow You [pointed out the additional checks](https://github.com/electron/electron/issues/897#issuecomment-412461917) that needed to be disabled. (Thank you!)

#### Checklist
- [X] PR description included and stakeholders cc'd
- [X] `npm test` passes (CI is green)
- [X] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [X] relevant documentation is changed or added (see note below)
- [X] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines) (once WIP prefix is removed)

**Note about documentation:**
Since this is fixing a regression (doing what the previous patch used to do), I don't think there's any need to change documentation. However, it might make sense to record this behavior in the documentation somewhere (I couldn't find it). Perhaps in the FAQ where [`localStorage` is mentioned](https://electronjs.org/docs/faq#how-to-share-data-between-web-pages).

**Note about running tests locally:**
It seems that I can't run the test suite locally, though I can build a release and manually test in the console. Here's what I get on my (Debian 9.5) machine when I try to build the debug version (needed for running tests, right?) It seems like my system's libc version does not match the required one.
</summary><details>

```sh
$ ninja -C out/Debug electron
ninja: Entering directory `out/Debug'
[7817/18095] ACTION //components/url_formatter/top_domains:generate_top_domains_trie(//build/toolchain/linux:clang_x64)
FAILED: gen/components/url_formatter/top_domains/alexa_domains-trie-inc.cc 
[...]

PR #20899, opened 2019-10-31
fix: prevent silent failure when DOM storage quota exceeded

jacobq:
#### Description of Change
This PR modifies [`dom_storage_limits.patch`](https://github.com/electron/electron/blob/ba85d4c3bbbaddb53795340053c273dfdf51aca1/patches/chromium/dom_storage_limits.patch) so that instead of (incompletely / incorrectly) circumventing the enforcement of the DOM storage quota, it is simply increased (from 10MiB to 100MiB). The primary motivation for this change is that the previous patch (see PR #15596) does not actually work properly ([example](https://github.com/jacobq/electron-quick-start/tree/storageLimit)). 

As described in the [updated patch notes](https://github.com/electron/electron/pull/20899/files#diff-557eb2fa2d8baedff18ed8d626af55c4), bypassing the [maximum size enforcement in
`StorageAreaImpl::Put`](https://github.com/chromium/chromium/blob/62521aaefdef5b345dd470f14eef6e44732ccf23/content/browser/dom_storage/storage_area_impl.cc#L306-L315) partially resolves the problem but can lead to undesirable side-effects (e.g. instability / poor performance). 

**Cc:**
@deepak1556 
@alexeykuzmin 
@nornagon 

#### Checklist
- [x] `npm test` passes

#### Release Notes

Notes: DOM storage quota enforcement re-enabled, but limit changed from 10MiB to 100MiB

### DIFF
--- a/patches/chromium/dom_storage_limits.patch
+++ b/patches/chromium/dom_storage_limits.patch
@@ -1,80 +1,46 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Jacob Quant <jacobq@gmail.com>
-Date: Tue, 6 Nov 2018 15:26:00 -0600
+Date: Thu, 31 Oct 2019 14:00:00 -0500
 Subject: dom_storage_limits.patch
 
-This patch circumvents the restriction on DOM storage objects,
-namely `localStorage` and `sessionStorage`, which chromium otherwise
-limits to approximately 10MiB.
+This patch increases the DOM storage (e.g. `localStorage`
+and `sessionStorage`) size quota from 10MiB to 100MiB.
+Previous versions of this patch attempted to circumvent
+the restriction altogether.
+However, this can lead to other problems, such as crashing
+the Dev Tools when attempting to read or write values that exceed
+`IPC::Channel::kMaximumMessageSize` (128MiB).
 
-That restriction originates from a recommendation
-[in the Web Storage API specification](https://html.spec.whatwg.org/multipage/webstorage.html#disk-space-2)
-that is motivated by the concern that hostile code could abuse this
-feature to exhaust available storage capacity.
-However, in the case of  Electron, where the application developers
-have control over all of the code being executed,
-this safety precaution becomes a hindrance that does not add much value.
-For example, if a malicious developer wanted to consume disk space
-on a victim's machine they could do so via Node's native file system API.
-
-By disabling this restriction or increasing the quota,
-electron application developers can use `localStorage`
-as their application's "back end", without being having
-to limit the amount of data stored to 10MiB.
-
-There may still be some benefit to keeping this restriction for applications that load remote content.
-Although all remote data should be from a trusted source and transferred using
-a secure channel, it is nevertheless advisable to include additional layers of protection
-to mitigate risks associated with potential compromise of those other technologies.
-With that in mind, an acceptable alternative to disabling the limit at compile-time
-(as this patch currently does) would be to instead allow it to be disabled at run-time
-for a given `BrowserWindow` via a `webPreferences` option,
-similar to [`nodeIntegration`](https://electronjs.org/docs/tutorial/security#2-disable-nodejs-integration-for-remote-content).
+Increasing the quota rather than bypassing it reduces the
+amount of chromium code that needs to be changed for Electron
+as well as keeps these storage areas limited to a bounded
+size meanwhile giving application developers more space to work with.
 
 diff --git a/content/browser/dom_storage/dom_storage_types.h b/content/browser/dom_storage/dom_storage_types.h
-index 6c0b831ebaaa2c1749bbc7436ce1025656588310..b67767751cadc6072c133297c7a6cdcc6bfd0c98 100644
+index 6c0b831ebaaa2c1749bbc7436ce1025656588310..96d1c73adb09d33cf591ca569f46de9b21f9a4df 100644
 --- a/content/browser/dom_storage/dom_storage_types.h
 +++ b/content/browser/dom_storage/dom_storage_types.h
-@@ -21,6 +21,7 @@ typedef std::map<base::string16, base::NullableString16> DOMStorageValuesMap;
+@@ -21,7 +21,8 @@ typedef std::map<base::string16, base::NullableString16> DOMStorageValuesMap;
  
  // The quota for each storage area.
  // This value is enforced in renderer processes and the browser process.
-+// However, Electron's dom_storage_limits.patch removes the code that checks this limit.
- const size_t kPerStorageAreaQuota = 10 * 1024 * 1024;
+-const size_t kPerStorageAreaQuota = 10 * 1024 * 1024;
++// Electron's dom_storage_limits.patch increased this value from 10MiB to 100MiB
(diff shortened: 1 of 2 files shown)


jacobq:
Looks like my test isn't passing yet -- marking as WIP for now

jacobq:
It seems that the patch is not being applied at the right time or something because I can see the following in `src/out/Debug/gen/third_party/blink/public/mojom/dom_storage/storage_area.mojom.h`:

```c++
  static constexpr uint32_t kPerStorageAreaQuota = 10485760U;
  
  static constexpr uint32_t kPerStorageAreaOverQuotaAllowance = 102400U;
```

which I suspect is derived from the unpatched `src/content/browser/dom_storage/dom_storage_types.h`

jacobq:
FYI The failing mac test appears to be a fluke (failure was caused by time out and the corresponding code should be completely unrelated to this PR).

jacobq:
Please let me know if there's anything that I can do to help land this fix.

zcbenz:
This PR requires approval from @electron/wg-upgrades.

spec/chromium-spec.js
```diff
@@ -375,16 +375,38 @@ describe('chromium feature', () => {
...
+        it(`throws when attempting to use more than 128MiB in ${storageName}`, () => {
+          expect(() => {
+            const testKeyName = '_electronDOMStorageQuotaStillEnforcedTest'
+            const length = 128 * Math.pow(2, 20) - testKeyName.length
+            try {
+              storage.setItem(testKeyName, 'X'.repeat(length))
+            } finally {
+              storage.removeItem(testKeyName)
+            }
+          })
```
jacobq: Did I forget [`.to.throw()`](https://www.chaijs.com/api/bdd/#method_throw) here?

jacobq: Looks like I did. I will force push to correct this now.

patches/chromium/put_back_deleted_colors_for_autofill.patch
```diff
@@ -94,7 +94,7 @@ index f1330721a711c61211ea136286ee884bc00e1975..e479fc84f7013984fc53d18cad8812fb
      kColorId_ThrobberSpinningColor,
      kColorId_ThrobberWaitingColor,
 diff --git a/ui/native_theme/native_theme_win.cc b/ui/native_theme/native_theme_win.cc
-index 5b1804538c2e78af23c797004f250369580b4178..97e72bc38a7231ce295ad943e977c69586cc6e07 100644
+index 5b1804538c2e..97e72bc38a72 100644
 --- a/ui/native_theme/native_theme_win.cc
 +++ b/ui/native_theme/native_theme_win.cc
 @@ -673,6 +673,17 @@ SkColor NativeThemeWin::GetSystemColor(ColorId color_id,
```
nornagon: Please revert these unrelated changes.

patches/chromium/dom_storage_limits.patch
```diff
@@ -1,80 +1,46 @@
...
-   // shrinking changes to pre-existing files that are over budget.
-   if (check_quota && new_item_size > old_item_size && new_quota_used > quota_)
-     return false;
-+#endif
- 
-   keys_values_.Set(key, value);
-   ResetKeyIterator();
+ // accomodate concurrent writes from different renderers
+diff --git a/third_party/blink/public/mojom/dom_storage/storage_area.mojom b/third_party/blink/public/mojom/dom_storage/storage_area.mojom
+index 1f1b2c6fa109..d8f15eed9be8 100644
```
nornagon: Please use the `git-export-patches` script as described in [the patches documentation](https://github.com/electron/electron/blob/master/docs/development/patches.md#editing-an-existing-patch). It's important that we get the full hashes here, amongst other formatting things that the script does.

patches/chromium/put_back_deleted_colors_for_autofill.patch
jacobq: Sure, I will drop that commit. (Would you accept a separate PR to fix the "trailing whitespace" warning that gets produced without it?)

patches/chromium/dom_storage_limits.patch
jacobq: Ah, I should've read that. (I was just using `git format-patch`) I'll fix that shortly and push again.

patches/chromium/put_back_deleted_colors_for_autofill.patch
nornagon: sure :)

spec/chromium-spec.js
```diff
@@ -371,16 +371,38 @@ describe('chromium feature', () => {
...
+          // translation between character and byte counts. However,
+          // a string of 40 * 2^20 characters will require at least 40MiB
+          // and presumably no more than 80MiB, a size guaranteed to
+          // to exceed the original 10MiB quota yet stay within the
+          // new 100MiB quota.
+          // Note that both the key name and value affect the total size.
+          const testKeyName = '_electronDOMStorageQuotaIncreasedTest'
+          const length = 40 * Math.pow(2, 20) - testKeyName.length
+          storage.setItem(testKeyName, 'X'.repeat(length))
+          // Wait at least one turn of the event loop to help avoid false positives
```
nornagon: What false positives might these be? Ideally we could simplify this test to not require a tick.

jacobq: Good question. This might not be applicable any more, but when we were circumventing the restriction prior to this PR, a write followed immediately by a read would "appear to work" (no exceptions throw & reading shows the expected value). Now that we're letting chromium behave more like its usual self (i.e. throwing an exception when a write would exceed the allowed limit) these "false positives" should've be possible any more. Would you like me to remove the `setTimeout`?

nornagon: ah, okay. In that case it seems fine to keep the extra tick, it will let us know if things start behaving weirdly again. It'd be great if you could add that context into this comment, though, as it may help the next person to work with this code!

jacobq: Updated -- hope that's useful (I'm often wary of comments as they can accidentally be misleading)

jacobq:
Could you give the CI a kick? (lint failed due to "Error: socket hang up")

jacobq:
Looks like win-x64 could use a restart as well... the error looks unrelated:
```
not ok 443 <webview> tag <webview>.capturePage() returns a Promise with a NativeImage
  AssertionError: expected undefined to equal 6
      at Context.<anonymous> (C:\projects\src\electron\spec\webview-spec.js:1035:32)
[4576:1113/203523.801:WARNING:ipc_message_attachment_set.cc(49)] MessageAttachmentSet destroyed with unconsumed attachments: 0/1
```

jacobq:
@deepak1556 I'm sure you're busy, but would you mind taking a quick peek at this and letting me know if you notice anything major that needs to be addressed?

jacobq:
:pray: Thanks, Robo.

jacobq:
> @jacobq has manually backported this PR to "8-x-y", please check out #21380

Hopefully this is the right process.... I just cherry-picked the squashed commit from this onto `8-x-y` since I noticed that this didn't land in the latest beta (v8.0.0-beta.4).

PR #20925, opened 2019-11-02
fix: capture the promise global to avoid userland mutation

MarshallOfSound:
A common module `Bluebird` overrides the `Promise` global with a non A+ implementation that doesn't support thennables (`Promise.resolve({ then: () => Promise.resolve(123) })`).

This modifies Electrons code to make it impossible for us to rely on the global, instead we cache the global at runtime in the `capturedGlobals` object that is auto-injected by webpack where it is used.

There is also an eslint rule to prevent accidental re-introduction of this issue.

This is kinda similar to node's primordials but doesn't strip prototypes and deep clone, it just retains a reference to the original constructor.

Notes: Fixed issue where proxied remote promises might not resolve if `Bluebird` was installed in the renderer


MarshallOfSound:
This was caught as we adopted the promisified methods and noticed them not resolving in the renderer

lib/renderer/api/remote.js
```diff
@@ -7,6 +7,10 @@ const { CallbacksRegistry } = require('@electron/internal/renderer/remote/callba
 const { isPromise, isSerializableObject } = require('@electron/internal/common/remote/type-utils')
 const { ipcRendererInternal } = require('@electron/internal/renderer/ipc-renderer-internal')
 
+const {
+  Promise
+} = capturedGlobals
```
nornagon: can we have webpack automatically inject this instead of modifying every single JS file?

MarshallOfSound: Technically, yes.  We could make webpack provide `Promise`.  I did it this way to somewhat align with how primordials _look_.  I can do a Promise provide if that's the kind of magic we want to do 👍

nornagon: ah ok, I'm not familiar with node's primordials. I'm fine with this style if it's matching node!

PR #20949, opened 2019-11-04
build: include windows toolchain profiler

thypon:
#### Description of Change

This commit allows any electron user to compile and verify electron
artifacts reproducibility. When the source code
is reproducibly buildable with a given toolchain this change
allows them to verify the hash of their build, with the same
(windows) toolchain againss the CI version.

This change was tested on top of version https://chromium.googlesource.com/chromium/src/+/fb9837799bb6388ecaee010fc80cc1438e6bb134
which was buildable and reproducible for chromium and electron.

Further tests may be introduced to allow checking for local reproducibility.
EG: building twice an artifact on the same machine, with different time
and output directory should yield similar hashes.

#### Release Notes

Notes: no-notes

### DIFF
--- a/build/profile_toolchain.py
+++ b/build/profile_toolchain.py
@@ -1,0 +1,98 @@
+from __future__ import with_statement
+import contextlib
+import sys
+import os
+import optparse
+import json
+
+sys.path.append("%s/../../build" % os.path.dirname(os.path.realpath(__file__)))
+
+import find_depot_tools
+from vs_toolchain import \
+    SetEnvironmentAndGetRuntimeDllDirs, \
+    SetEnvironmentAndGetSDKDir, \
+    GetVisualStudioVersion, \
+    NormalizePath
+
+sys.path.append("%s/win_toolchain" % find_depot_tools.add_depot_tools_to_path())
+
+from get_toolchain_if_necessary import CalculateHash
+
+
+@contextlib.contextmanager
+def cwd(dir):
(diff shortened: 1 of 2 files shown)


thypon:
Cc: @nornagon @felixrieseberg

This fix includes both the windows toolchain profiler and the appveyor uploader.
I am not sure if it's a good idea to enable a second build on windows to check potential reproducible builds regressions.

build/profile_toolchain.py
```diff
@@ -0,0 +1,98 @@
...
+        'sdks': [
+            {'name': 'vs', 'path': path, 'hash': calculate_hash(path)},
+            {'name': 'wsdk', 'path': win_sdk_dir, 'hash': calculate_hash(win_sdk_dir)}
+        ],
+        'runtime_lib_dirs': runtime_dll_dirs,
+    }
+
+
+def main(options):
+    if os.name == 'nt':
```
nornagon: elsewhere we use `sys.platform in ['win32', 'cygwin']` (though i don't think we support building on cygwin...). For consistency,
```suggestion
    if sys.platform == 'win32':
```

thypon: Thanks, fixed in the rebased commit!

nornagon:
Looks like there are conflicts in appveyor.yml, please take a look!

thypon:
Rebases. It should be okay now.

Issue #20752, opened 2019-10-25
webContents.login event not triggered in 7.0.0


binux:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.0.0
* **Operating System:**
  * macOS
* **Last Known Working Electron version:**
  * 6.1.2

### Expected Behavior
https://github.com/electron/electron/blob/v7.0.0/docs/api/web-contents.md#event-login triggered

### Actual Behavior
login event not triggered

### To Reproduce
run following code

```js
const {app, BrowserWindow} = require('electron')
app.once('ready', async () => {
  win = new BrowserWindow()
  app.on('login', console.log)
  win.webContents.on('login', console.log)
  win.webContents.once('did-finish-load', async () => {
    const html = await win.webContents.executeJavaScript('document.documentElement.innerHTML')
    console.log(html)
    app.quit()
  })
  await win.webContents.loadURL('https://httpbin.org/basic-auth/user/passwd')
})
```

nornagon:
Yeah, unfortunately the 'login' event was a casualty of us refactoring on top of Chromium's new 'network service' code. It's currently just ... not attached at all. Whoops. https://github.com/electron/electron/blob/7-0-x/shell/browser/login_handler.cc#L37-L39

sowelie:
Any thoughts on when this will be implemented? This is holding me back from updating to 7.x. Is there anything I can do to help with the fix?

gavinkalika:
any movement on this getting fixed?

PR #20954, opened 2019-11-04
fix: implement 'login' event for WebContents

nornagon:
#### Description of Change
The 'login' event was disabled during the Great Network Servicification (see https://github.com/electron/electron/issues/19602). This reimplements it on top of the new stuff.

Fixes https://github.com/electron/electron/issues/20752

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed the 'login' event not being emitted when HTTP Basic auth was requested from a WebContents.


nornagon:
TODO: this event this commit implements doesn't include request details (url, referrer, method, etc.), because they aren't easily accessible at the time the event is emitted. We should either figure out a way to get that info or update the docs to reflect that the info is no longer present.

nornagon:
Also TODO: events from the 'net' module...

spec-main/api-web-contents-spec.ts
```diff
@@ -1518,4 +1518,37 @@ describe('webContents module', () => {
...
+        serverUrl = 'http://127.0.0.1:' + (server.address() as AddressInfo).port
+        done()
+      })
+    })
+
+    after(() => {
+      server.close()
+    })
+
+    it('receives login event on webcontents', (done) => {
```
MarshallOfSound: Can we make this test actually verify the `cb` worked.  Like callback with valid creds and test the page loaded and callback with invalid creds and assert it didn't

nornagon: done

nornagon:
Since the issue with the `net` module is blocked waiting on a response from upstream, i think it would be a good idea to merge this and not wait.

gavinkalika:
any movement on getting this fixed merged

PR #20962, opened 2019-11-05
fix: allow iframe-initiated HTML fullscreen to exit while in macOS fullscreen

loc:
#### Description of Change

Since Electron 6, exiting HTML fullscreen from an out-of-process iframe (OOPIF) has not worked correctly while in macOS fullscreen. As far as I can find, this has not been reported.

Repro:
1. Open fiddle: https://gist.github.com/loc/e6d622ababe7dda9ce4a783dba1cb8c6
2. Play and fullscreen the youtube embed
3. Try to exit fullscreen
4. Note that you can't (until you swipe to a different Space or otherwise force a resize).

The issue is that we're not triggering a manual resize (since the view isn't actually changing size when changing from HTML fullscreen to OS fullscreen) in the same way that Chrome does in /browser/ui.

The fix is for macOS only, but the added test covers all three platforms. Testing is kind of a pain, since iframes are a pain — I got it to a place where it wasn't flakey locally, but am open to suggestions on how to make sure it doesn't become flakey. 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix exiting HTML fullscreen for cross-origin iframes (e.g. YouTube) while in macOS fullscreen

### DIFF
--- a/shell/browser/common_web_contents_delegate.cc
+++ b/shell/browser/common_web_contents_delegate.cc
@@ -355,6 +355,13 @@ void CommonWebContentsDelegate::ExitFullscreenModeForTab(
     return;
   SetHtmlApiFullscreen(false);
   owner_window_->NotifyWindowLeaveHtmlFullScreen();
+
+  if (native_fullscreen_) {
+    // Explicitly trigger a view resize, as the size is not actually changing if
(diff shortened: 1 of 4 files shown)


codebytere:
@loc looks like this still needs one to `8-x-y`

PR #20963, opened 2019-11-05
fix: Fix broken globalShortcuts.registerAll() on non-macOS platforms

poiru:
This was a regression in #16125, which unintentionally put
`GlobalShortcutListener::RegisterAccelerator` into a
`#if defined(OS_MACOSX)` block.

Notes: Fix broken `globalShortcut.registerAll()` on Windows and Linux

### DIFF
--- a/shell/browser/api/atom_api_global_shortcut.cc
+++ b/shell/browser/api/atom_api_global_shortcut.cc
@@ -74,19 +74,13 @@ bool GlobalShortcut::RegisterAll(
   std::vector<ui::Accelerator> registered;
 
   for (auto& accelerator : accelerators) {
-#if defined(OS_MACOSX)
-    if (RegisteringMediaKeyForUntrustedClient(accelerator))
-      return false;
-
-    GlobalShortcutListener* listener = GlobalShortcutListener::GetInstance();
-    if (!listener->RegisterAccelerator(accelerator, this)) {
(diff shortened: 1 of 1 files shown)


poiru:
cc @codebytere

poiru:
... actually, given that this has not been reported, how about we (also) deprecate `registerAll`, `unregisterAll`, etc. They are trivial to implement on top of `register`, `unregister`, etc. so I don't see the value of providing them as part of our own API.

codebytere:
it should just be:
```cpp
bool GlobalShortcut::RegisterAll(
    const std::vector<ui::Accelerator>& accelerators,
    const base::Closure& callback) {
  std::vector<ui::Accelerator> registered;

  for (auto& accelerator : accelerators) {
#if defined(OS_MACOSX)
    if (RegisteringMediaKeyForUntrustedClient(accelerator))
      return false;
#endif
    GlobalShortcutListener* listener = GlobalShortcutListener::GetInstance();
    if (!listener->RegisterAccelerator(accelerator, this)) {
      // unregister all shortcuts if any failed
      UnregisterSome(registered);
      return false;
    }
    registered.push_back(accelerator);
    accelerator_callback_map_[accelerator] = callback;
  }
  return true;
}
```

poiru:
@codebytere The code now calls `GlobalShortcut::Register`, which does that check.

PR #20965, opened 2019-11-05
fix: disable remote layer APIs in MAS build

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/20027.

This PR disables following private macOS APIs used in Chromium:

* `CAContext`
* `CALayerHost`

Chromium has 2 ways of rendering on macOS: one is using official `IOSurface` way, and one is `CARemoteLayer` API. The `IOSurface` way is used when software compositing is enabled, and when the remote layer API is not available.

This PR forces using `IOSurface ` even when remote layer API is available, it essentially behaves the same with passing the `--disable-remote-core-animation` command line flag.

More information can be found at https://bugs.chromium.org/p/chromium/issues/detail?id=312462.

* `NSAccessibilityRemoteUIElement`

All related code using this class are disabled. It would affect accessibility when remote layer API is used, but we should be fine since remote layer API is disabled too.

* `NSNextStepFrame`
* `NSThemeFrame`

Chromium overrides these private classes to implement custom frame. All related code are disabled so custom frame would not work.

Since in Electron we only support standard frame and simple frameless window, we are not affected.

* `NSURLFileTypeMappings`

Chromium uses it to guess mime type from extension names, the iOS version of Chromium does not use it. The fallback detection should be enough for us.

#### Checklist

- [x] `npm test` passes

#### Release Notes

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -31,6 +31,7 @@ scroll_bounce_flag.patch
 mas-cfisobjc.patch
 mas-cgdisplayusesforcetogray.patch
 mas-audiodeviceduck.patch
+mas_disable_remote_layer.patch
 mas_disable_remote_accessibility.patch
 mas_disable_custom_window_frame.patch
 chrome_key_systems.patch
--- a/patches/chromium/mas_disable_remote_layer.patch
+++ b/patches/chromium/mas_disable_remote_layer.patch
@@ -1,0 +1,137 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Cheng Zhao <zcbenz@gmail.com>
+Date: Thu, 20 Sep 2018 17:48:49 -0700
+Subject: mas_disable_remote_layer.patch
+
+Disable remote layer APIs (CAContext and CALayerHost) for MAS build.
+
+Accordin to ccameron:
+
+For apps that spend a lot of time watching video (especially fullscreen video),
+the power/battery usage will likely increase 1.5x to 2x. For something that is,
+e.g, scrolling, it'll be smaller, more like 1.15x or 1.25x.
+
+In terms of performance, the impact will likely be fairly small -- any app that
+could hit 60fps before will likely still be able to hit 60fps. There may even be
+cases where performance improves when disabling remote CoreAnimation (remote
+CoreAnimation is really only about battery usage).
+
+diff --git a/gpu/ipc/service/image_transport_surface_overlay_mac.mm b/gpu/ipc/service/image_transport_surface_overlay_mac.mm
+index c1af03a268dc..b883883ad52b 100644
+--- a/gpu/ipc/service/image_transport_surface_overlay_mac.mm
++++ b/gpu/ipc/service/image_transport_surface_overlay_mac.mm
+@@ -63,6 +63,7 @@
+ template <typename BaseClass>
+ bool ImageTransportSurfaceOverlayMacBase<BaseClass>::Initialize(
+     gl::GLSurfaceFormat format) {
++#ifndef MAS_BUILD
+   // Create the CAContext to send this to the GPU process, and the layer for
+   // the context.
+   if (use_remote_layer_api_) {
+@@ -71,6 +72,7 @@
+         [CAContext contextWithCGSConnection:connection_id options:@{}] retain]);
+     [ca_context_ setLayer:ca_layer_tree_coordinator_->GetCALayerForDisplay()];
+   }
++#endif
+   return true;
+ }
+ 
+@@ -139,7 +141,9 @@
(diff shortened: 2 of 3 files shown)


zcbenz:
The original issue should be kept open since this PR does not patch `fileport_makefd` and `fileport_makeport`, and I don't know if it is needed.

(I hope it is not needed since these two are really difficult to get rid of.)

patches/chromium/mas_disable_remote_accessibility.patch
```diff
@@ -0,0 +1,283 @@
...
+diff --git a/ui/base/BUILD.gn b/ui/base/BUILD.gn
+index c5d4da4baef2..0101dbb1cf77 100644
+--- a/ui/base/BUILD.gn
++++ b/ui/base/BUILD.gn
+@@ -287,6 +287,13 @@ jumbo_component("base") {
+     "window_open_disposition.h",
+   ]
+ 
++  if (is_mas_build) {
++    sources -= [
```
deepak1556: why are we excluding them here ?

deepak1556:
`3-1-x` and `4-2-x` are not supported release lines, hence removed their backport labels

patches/chromium/mas_disable_remote_accessibility.patch
zcbenz: `remote_accessibility_api.mm` implements `ui::RemoteAccessibility` which uses private APIs, excluding it is simpler than commenting out the whole file.

zcbenz:
> We should look into the possibility of pushing the remote implementation behind a buildflag in upstream to reduce our patch in future. Thoughts ?

It is definitely good idea, it would also keep Chromium maintaining the code path we use.

patches/chromium/mas_disable_remote_accessibility.patch
deepak1556: But we also seem to be patching them below is that needed ?

zcbenz: In theory it is fine leaving `remote_accessibility_api.h` untouched, but by commenting out the declarations defined in `remote_accessibility_api.h`, any other file that uses the private APIs would trigger compilation error early, so we won't miss things when updating/backporting the patch.

MarshallOfSound:
Switching to `IOSurface` will (according to that linked issue) result in a performance hit.  Do we have any idea how to quantify that hit?

zcbenz:
> Switching to `IOSurface` will (according to that linked issue) result in a performance hit. Do we have any idea how to quantify that hit?

I don't know how to measure the performance hit, they did not do any actual measurement neither in the issue.

zcbenz:
Status update: as disabling remote layer APIs might drag down performance significantly, we are currently waiting for responses from Apple to see if it is possible to unblock `CAContext` and `CALayerHost` for us, or whether is there a better way of patching to avoid performance hit.

ngehlert:
any new updates here?

ncodeyx:
Also waiting for update. https://github.com/electron/electron/issues/20027#issuecomment-554785974

JCBsystem:
any news on this ? 
for us with a live app in MAS with live customers this is extremely important

soundar78:
Any idea when this can be merged @MarshallOfSound @nornagon ?

gaodeng:
What can we do to help advance this matter quickly? Or can we release a beta version containing this patch first, so that everyone can easily update the current app on the mac app store

JCBsystem:
new week .. 
please keep us updated

gaodeng:
slack was updated yesterday, does anyone know how they pass the review？I also submitted a new version yesterday, but it was rejected this morning 。

JCBsystem:
> slack was updated yesterday, does anyone know how they pass the review？I also submitted a new version yesterday, but it was rejected this morning 。

@gaodeng what reason did you get rejected for ?
and did you build with the zip or the electron:...path_to_patch_file 
i think path_to_patch_file can have more libs that don't belong i the MAS

gaodeng:
> @gaodeng what reason did you get rejected for ?

Private API Rejection .

> About slack it looks like the use
> "electron": "6.0.10"
> when i look in the package.json
> but they can have used the Zip build hack..

No,I don't think so.
otool -ov with slack's electron framwork 
屏幕快照 2019-12-07 下午4 59 24

Why slack can pass the review？ the answer may be simple  , "All developers are equal, but some developers are more equal than others."

JCBsystem:
Maybe this is a good thing ....
 "we are currently waiting for responses from Apple"  
we can hope apple did a exception for electron :)

snokier:
any news here？

gtamas:
@zcbenz 

Any news? 

Exactly which electron version has this patch? Is there an electron version which is 100% supported  by Apple Store? 

If there isn't, when can we expect such a version to be released?

PR #20966, opened 2019-11-05
fix: disable more private macOS APIs in MAS build (8-x-y)

zcbenz:
Backport of https://github.com/electron/electron/pull/20965.


jkleinsc:
Closed in favor of #21574

PR #20985, opened 2019-11-05
fix: correctly emplace optional values in the value converter

### LINKED ISSUE #20941: app.setAppLogsPath() does not set the `logs` path
### Preflight Checklist

### Issue Details

* **Electron Version:** 7.0.0, 7.0.0-beta.4
* **Operating System:** macOS 10.14.6 (Mojave)
* **Last Known Working Electron version:** 7.0.0-beta.5

### Expected Behavior
`app.setAppLogsPath(path)` should change the default logging path

### Actual Behavior
`app.setAppLogsPath(desktopPath)` seems to have no effect, logs are always created in `~/Library/Logs/[appName]`

### To Reproduce
Fiddle: https://gist.github.com/eb7c5c66066afba3ab6dbb0353f79657

```js
const {app} = require('electron')
const path = require('path')

app.on('ready', () => {
  const desktopPath = app.getPath(`desktop`)
  app.setAppLogsPath(desktopPath)

  const logPath = app.getPath('logs')
  console.assert(logPath === desktopPath, `Expected ${logPath} to equal ${desktopPath}`);

  app.quit();
})
```

MarshallOfSound:
I don't wanna talk about it....

Fixes #20941

Will follow up with a test for master, we do not need to fix anything in master as the gin converter migration fixed this by accident

Notes: Fixed issue where `app.setAppLogsPath` did not work when you provided a valid path

### DIFF
--- a/shell/common/native_mate_converters/value_converter.h
+++ b/shell/common/native_mate_converters/value_converter.h
@@ -44,7 +44,7 @@ struct Converter<base::Optional<T>> {
       return true;
     }
     T converted;
-    if (Converter<T>::FromV8(isolate, val, &converted)) {
+    if (!Converter<T>::FromV8(isolate, val, &converted)) {
       return true;
     }
     out->emplace(converted);
--- a/shell/common/native_mate_converters/value_converter.h
+++ b/shell/common/native_mate_converters/value_converter.h
@@ -41,10 +41,12 @@ struct Converter<base::Optional<T>> {
                      v8::Local<v8::Value> val,
                      base::Optional<T>* out) {
     if (val->IsNull() || val->IsUndefined()) {
+      *out = base::nullopt;
       return true;
     }
     T converted;
     if (!Converter<T>::FromV8(isolate, val, &converted)) {
+      *out = base::nullopt;
       return true;
     }
     out->emplace(converted);


felixrieseberg:
What. The. Hell.

shell/common/native_mate_converters/value_converter.h
```diff
@@ -44,7 +44,7 @@ struct Converter<base::Optional<T>> {
       return true;
     }
     T converted;
-    if (Converter<T>::FromV8(isolate, val, &converted)) {
+    if (!Converter<T>::FromV8(isolate, val, &converted)) {
```
nornagon: it's unlikely to affect usages, but we should also set the optional to empty when we get undefined/null, rather than leaving it as it was.
```suggestion
    if (!Converter<T>::FromV8(isolate, val, &converted)) {
      *out = base::nullopt;
```

MarshallOfSound: Did this in both places 👍

shell/common/native_mate_converters/value_converter.h
```diff
@@ -41,10 +41,12 @@ struct Converter<base::Optional<T>> {
                      v8::Local<v8::Value> val,
                      base::Optional<T>* out) {
     if (val->IsNull() || val->IsUndefined()) {
+      *out = base::nullopt;
       return true;
     }
     T converted;
-    if (Converter<T>::FromV8(isolate, val, &converted)) {
+    if (!Converter<T>::FromV8(isolate, val, &converted)) {
+      *out = base::nullopt;
```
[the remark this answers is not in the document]
MarshallOfSound: It isn't a failed conversion because this is an optional param.  If this param fails to convert the optional will be empty and the next arg will use this param.

This emulates expected behavior of optional arguments in the middle of argument lists.

PR #20988, opened 2019-11-06
fix: disable more private macOS APIs in MAS build (5-0-x)

zcbenz:
Backport of #20965.


jack-arms:
At least for 5.0.x I think this code also need a patched `#ifndef MAS_BUILD ... #endif` around it: https://chromium.googlesource.com/chromium/src.git/+/refs/tags/73.0.3683.121/content/browser/renderer_host/render_widget_host_view_mac.mm#1929

PR #21003, opened 2019-11-06
fix: disable more private macOS APIs in MAS build (4-2-x)

cmgustavo:
Backport of #20965 .

### DIFF
--- a/patches/common/chromium/.patches
+++ b/patches/common/chromium/.patches
@@ -61,6 +61,10 @@ windows_cc_wrapper.patch
 enable_osr_components.patch
 ignore_rc_check.patch
 enable_widevine.patch
+mas_disable_remote_layer.patch
+mas_disable_remote_accessibility.patch
+mas_disable_custom_window_frame.patch
+mas_disable_spi_file_type_mappings.patch
 chrome_key_systems.patch
 allow_nested_error_trackers.patch
 blink_initialization_order.patch
--- a/patches/common/chromium/mas_disable_custom_window_frame.patch
+++ b/patches/common/chromium/mas_disable_custom_window_frame.patch
@@ -1,0 +1,146 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Cheng Zhao <zcbenz@gmail.com>
+Date: Thu, 20 Sep 2018 17:48:49 -0700
+Subject: mas_disable_custom_window_frame.patch
+
+Disable private window frame APIs (NSNextStepFrame and NSThemeFrame) for MAS
+build.
+
+diff --git a/components/remote_cocoa/app_shim/browser_native_widget_window_mac.mm b/components/remote_cocoa/app_shim/browser_native_widget_window_mac.mm
+index 677279352ba9..2194a762c6a1 100644
+--- a/components/remote_cocoa/app_shim/browser_native_widget_window_mac.mm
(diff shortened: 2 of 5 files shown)


miniak:
@cmgustavo Electron 4 is not supported anymore

codebytere:
@cmgustavo `4.x` is [no longer supported](https://electronjs.org/docs/tutorial/support#currently-supported-versions) and as such we are no longer running releases from that branch

ffflorian:
@codebytere I am really unhappy about the fact that this backport was closed. I understand that you have a strict release cycle but right now it's impossible to publish Electron 4 apps to the Apple App Store.

Would you maybe make an exception for this one so that people running Electron 4 don't have to upgrade to Electron 5 just to be able to publish to the Apple App Store?

jarek-foksa:
I second that, could you please make an exception here? This is a special case since perfectly stable Electron 4 builds are no longer usable due to sudden and unannounced Apple policy changes.

 Electron 5.x (or later) has too many serious bugs to be used in production. For example my app is completely broken on some devices because of bug https://github.com/electron/electron/issues/19626.

MarshallOfSound:
@ffflorian @jarek-foksa It's not even necessarily a case of making an exception.  We have complex release automation that changes quite rapidly and once a version is EOL we no longer ensure compatibility with the current system.

This PR also did not appear to contain valid patches, based on the output of the CI runs the patch files are invalid the version of Chromium that `4-2-x` depends on.

FWIW there are seval large apps using Electron 6 and 7 in production.  Updating to our latest versions of Electron is still our strong reccomendation here.

jarek-foksa:
Are you sure those large apps are actually published on the Mac App Store?

I did try updating to newer versions of Electron, all of them were crashing on my machine due to bug https://github.com/electron/electron/issues/19626. Therefore, now I can't publish any updates on the Mac App Store.

circleapps:
@MarshallOfSound 

My app is depending on Native Client to play local video files. 

After Electron 5.0.x,  this feature has been dropped, so I have to depend on Electron 4.x forever.

If you don't provide patch for Electron 4.x,  I can't publish any updates on the Mac App Store, which is a disaster for myself and all my customers.

PLEASE reconsider this patch, thanks.

miniak:
@circleapps you can always build Electron yourself with any custom backports / patches.

miniak:
@circleapps are you using NaCl or just a regular Pepper plugin? Native pepper plugins may not work anymore due to the host process being sandboxed, which can be fixed.

circleapps:
Thanks miniak,

I will learn how to build Electron 4.x with this patch by myself.

I remember the "sandbox" was introduced in Electron 5.x. So I assume if I stay with 4.x,  my plugin will still work?

circleapps:
I'm using mpv.js to play video files.
https://github.com/Kagami/mpv.js

miniak:
So it’s a pepper plugin, I can help you get it working with latest Electron

miniak:
it's tracked by #18516

circleapps:
Thanks! I will try this method to see if it works for me.

circleapps:
@miniak
 It works! Really appreciate your help!

semireg:
My app is dependent on many native modules (usb, font, printer to name a few) that do not yet support Node NAN in the way Electron >4 expect. Electron 4 was the last major electron version to work with these modules. In order for me to upgrade to Electron 5+ I have to become an expert with NAN and maintain forks of these projects. This is going to be a real challenge... 💪🏼

PR #21004, opened 2019-11-06
feat: add net.online / net.isOnline()

miniak:
#### Description of Change
It allows the app to tell that the machine is online without creating a `BrowserWindow`, in the main process.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `net.online` for detecting whether there is currently internet connection.

### DIFF
--- a/docs/api/net.md
+++ b/docs/api/net.md
@@ -8,7 +8,7 @@ The `net` module is a client-side API for issuing HTTP(S) requests. It is
 similar to the [HTTP](https://nodejs.org/api/http.html) and
 [HTTPS](https://nodejs.org/api/https.html) modules of Node.js but uses
 Chromium's native networking library instead of the Node.js implementation,
-offering better support for web proxies.
+offering better support for web proxies. It also supports checking network status.
 
 The following is a non-exhaustive list of why you may consider using the `net`
 module instead of the native Node.js modules:
@@ -62,3 +62,25 @@ Creates a [`ClientRequest`](./client-request.md) instance using the provided
 `options` which are directly forwarded to the `ClientRequest` constructor.
 The `net.request` method would be used to issue both secure and insecure HTTP
 requests according to the specified protocol scheme in the `options` object.
+
+### `net.isOnline()`
+
+Returns `Boolean` - Whether there is currently internet connection.
+
+A return value of `false` is a pretty strong indicator that the user
+won't be able to connect to remote sites. However, a return value of
+`true` is inconclusive; even if some link is up, it is uncertain
+whether a particular connection attempt to a particular remote site
+will be successful.
+
+## Properties
+
+### `net.online` _Readonly_
+
+A `Boolean` property. Whether there is currently internet connection.
+
+A return value of `false` is a pretty strong indicator that the user
+won't be able to connect to remote sites. However, a return value of
+`true` is inconclusive; even if some link is up, it is uncertain
+whether a particular connection attempt to a particular remote site
+will be successful.
--- a/lib/browser/api/net.ts
+++ b/lib/browser/api/net.ts
@@ -4,6 +4,7 @@ import { app } from 'electron/main';
 import type { ClientRequestConstructorOptions, UploadProgress } from 'electron/main';
 
 const {
+  isOnline,
   isValidHeaderName,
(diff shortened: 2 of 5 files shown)


docs/api/net.md
```diff
@@ -62,3 +62,13 @@ Creates a [`ClientRequest`](./client-request.md) instance using the provided
 `options` which are directly forwarded to the `ClientRequest` constructor.
 The `net.request` method would be used to issue both secure and insecure HTTP
 requests according to the specified protocol scheme in the `options` object.
+
+### `net.isOffline()`
```
nornagon: Prefer positive conditions, i.e. this should be `net.isOnline()` (this would also match the `navigator.onLine` API that @jkleinsc mentioned)

sindresorhus:
It would be useful to also have an event when this changes. For example, you want to detect when the user comes back online.

sindresorhus:
> why is this needed when navigator.onLine is available: developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/Online_and_offline_events?

That API is useless in practice: https://stackoverflow.com/a/14283180/64949

docs/api/net.md
sindresorhus: It should, IMHO, also be a property, not a method. (https://github.com/electron/electron/issues/12745)

MarshallOfSound:
@sindresorhus This API is backed by the same logic.

It can definitely tell when you are offline but can declare you "online" when you don't have a actual internet connection as the docs say.

sindresorhus:
@MarshallOfSound If that's true, that's great. Then it must have changed/improved at some point as it used to only be able to detect whether you have a connection, not specifically whether you're actually online (able to load websites).

MarshallOfSound:
@sindresorhus It still has that limitation, detecting if you are able to actually load a website is virtually impossible without just trying to load that website and declaring yourself offline if it fails.

sindresorhus:
@MarshallOfSound Then I think the docs should be clear that it works like `navigator.onLine` and that it does not guarantee that the user is actually online (able to load websites).

MarshallOfSound:
@sindresorhus The docs state that afaics.

> A return value of `true` is a pretty strong indicator that the user won't be able to connect to remote sites. However, a return value of `false` is inconclusive; even if some link is up, it is uncertain whether a particular connection attempt to a particular remote site will be successful.

miniak:
@jkleinsc I added the reason to the PR description, forgot to do that yesterday.

miniak:
@sindresorhus I want to add `network-changed` event using the `net::NetworkChangeNotifier::NetworkChangeObserver`, but I wasn't working for me on Mac for some reason. I didn't have time yet to figure out why or test on Windows.

docs/api/net.md
miniak: changed to `net.isOnline()`

codebytere:
@miniak are you still working on this or can it be closed?

miniak:
@codebytere I still want to land this.

miniak:
@jkleinsc added both property and getter method

zcbenz (approved):

jkleinsc:
@miniak the API WG reviewed this at our Oct 19, 2020 meeting.  Generally the WG approves of the PR but would like the following changes made:
1. The getter should be `net.isOnline()` instead of `net.getIsOnline`
2. The property should be named `online` instead of `isOnline`
3. The net doc says the following in its introduction: ```The net module is a client-side API for issuing HTTP(S) requests.```.  This introduction should be changed to accomodate the fact that the module now also supports checking network status.

miniak:
@jkleinsc feedback incorporated

codebytere (approved):

jkleinsc (approved):

ericblade:
fwiw, one way people have solved the "online is true, but can't actually connect to anything is a bad state" problem, is to use "online === false" to mean "i am definitely offline", but if the state changes to "online === true", then do a DNS resolution to confirm that a) the outside world is actually reachable and b) DNS resolution works, so that you have a hope of actually reaching a site you're interested in.  Granted it's still not 100%, but it's easily implementable on the node side, and is much closer to a useful meaning of that for most applications.

Issue #19008, opened 2019-06-27
Wrong behavior after canceling the printer dialog box


brunto:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.6
* **Operating System:**
  * macOS 10.14.5
* **Last Known Working Electron version:**
  * 4.2.5

### Expected Behavior
When I click on a **button** who uses the `print()` method
The print dialog opens
Then I click on "Cancel" button
I must be able to go to another page by clicking on a link

### Actual Behavior
When I click on a **button** who uses the `print()` method
The print dialog opens
Then I click on "Cancel" button
I can not click on another link, I have to quit the application and open it again

### To Reproduce
You can download the PIA 2.1.0 version at this URL https://github.com/LINCnil/pia-app/releases/tag/2.1.0 then:
* Create a new PIA
* Click on the **Preview** icon
* Click on the **Printer** icon (Wait for the printer dialog box)
* Click on the **Cancel** button
* Try to go on another page of the application

fxha:
@codebytere FYI, this issue also affect the button click event and links (`href="javascript:foo();"`). I also noticed that URLs which were opened in the default browser are loaded by Electron after cancelling the print dialog ([details](https://github.com/electron/electron/issues/16085#issuecomment-506869309)).

popod:
I think I've had something similar (only on Windows) by using `window.confirm()`. In this case, using [`dialog.showMessageBox()`](https://electronjs.org/docs/api/dialog#dialogshowmessageboxbrowserwindow-options-callback) will get around this problem.

fxha:
> Have you the same problem if you replace the print dialog by a `window.confirm()` ?

No, I haven't.

zadam:
@popod I have very similar issue - after calling confirm() I can't set focus to the text input - specifically the cursor does not appear after clicking the input, I can't write into input, but backspace will remove characters (which is pretty weird).

Opening dev tools fixes the problem, also switching to another window and then back fixes the problem as well.

This happens only on windows (linux works fine, I did not test mac). I'm using latest 6.0.0-beta.11

[...]

PR #20839, opened 2019-10-30
build: depend on chromium version not SHA

MarshallOfSound:
As in title, depend on tag not SHA

Notes: no-notes

PR #21026, opened 2019-11-07
fix: refactor printing for mojo

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/19008.

Fixes printing in the aftermath of its mojoification.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed printing.

### DIFF
--- a/patches/chromium/printing.patch
+++ b/patches/chromium/printing.patch
@@ -63,7 +63,7 @@ index ab32d5475a0e269d32f6ab71f7dc82aa9e02035a..9ada750416f02d23c2f9faae0c41e4a7
  }
  
 diff --git a/chrome/browser/printing/print_view_manager_base.cc b/chrome/browser/printing/print_view_manager_base.cc
-index 81cb954f02363c829947dde830f340a761c80a77..2a26bb6d5367337fa13da07cb07588684d1ea05a 100644
+index 81cb954f02363c829947dde830f340a761c80a77..98043d6ec0cb2c55493ef7446a6de0c4eeed184e 100644
 --- a/chrome/browser/printing/print_view_manager_base.cc
 +++ b/chrome/browser/printing/print_view_manager_base.cc
 @@ -27,10 +27,7 @@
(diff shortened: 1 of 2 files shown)


patches/chromium/printing.patch
```diff
@@ -321,10 +318,11 @@ index 40762a36024bc48dfe5259520161dc203197bfd0..e38aa442df858ce362645230f7642b2e
    std::unique_ptr<PrinterQuery> printer_query = queue_->PopPrinterQuery(0);
    if (!printer_query) {
      printer_query =
-@@ -228,11 +234,13 @@ void PrintingMessageFilter::OnScriptedPrintReply(
+@@ -228,11 +234,14 @@ void PrintingMessageFilter::OnScriptedPrintReply(
  void PrintingMessageFilter::OnUpdatePrintSettings(int document_cookie,
                                                    base::Value job_settings,
                                                    IPC::Message* reply_msg) {
++LOG(INFO) << "PrintingMessageFilter::OnUpdatePrintSettings";
```
codebytere: whoops, will delete

shell/browser/api/atom_api_web_contents.cc
```diff
@@ -1785,58 +1786,60 @@ void WebContents::Print(gin_helper::Arguments* args) {
...
-      }
-    }
-    if (page_range_list->GetSize() > 0)
-      settings.SetList(printing::kSettingPageRange, std::move(page_range_list));
-  }
+  settings.SetIntKey(printing::kSettingPrinterType, printing::kLocalPrinter);
+  settings.SetBoolKey(printing::kSettingShouldPrintSelectionOnly, false);
+  settings.SetBoolKey(printing::kSettingRasterizePdf, false);
+
+  // TODO(codebytere): figure out the value conversion for this
```
codebytere: i just wanted to get the thing compiling and mostly working - after this part is fixed this PR should be good to go

patches/chromium/printing.patch
```diff
@@ -117,28 +117,34 @@ index 81cb954f02363c829947dde830f340a761c80a77..2a26bb6d5367337fa13da07cb0758868
...
-@@ -123,7 +126,8 @@ PrintViewManagerBase::~PrintViewManagerBase() {
+@@ -123,7 +126,9 @@ PrintViewManagerBase::~PrintViewManagerBase() {
    DisconnectFromCurrentPrintJob();
  }
  
 -bool PrintViewManagerBase::PrintNow(content::RenderFrameHost* rfh) {
 +bool PrintViewManagerBase::PrintNow(content::RenderFrameHost* rfh,
-+                                    CompletionCallback callback) {
++                                    CompletionCallback callback,
++                                    base::Value settings) {
```
nornagon: nit: convention is for the callback to come last in the args list

patches/chromium/printing.patch
```diff
@@ -117,28 +117,34 @@ index 81cb954f02363c829947dde830f340a761c80a77..2a26bb6d5367337fa13da07cb0758868
...
 +  callback_ = std::move(callback);
-+  // FIXME: pass 'silent' through from the caller..?
-+  GetPrintRenderFrame(rfh)->PrintRequestedPages(false /* silent */);
++
++  if (!callback_.is_null()) {
++    registrar_.Add(this, chrome::NOTIFICATION_PRINT_JOB_EVENT,
++            content::NotificationService::AllSources());
++  }
++
++  GetPrintRenderFrame(rfh)->PrintRequestedPages(false /* silent */, std::move(settings));
```
nornagon: do you need to pass `silent` here..?

patches/chromium/printing.patch
```diff
@@ -381,20 +378,29 @@ index 9fbea6d0a2dbe55b1d600fbc217dee5aa8ae8cd5..de9bd267e408c02fd4da7d903523c0e6
...
+ // Interface implemented by a class that desires to render print documents for
+ // Chrome print preview.
+ interface PrintRenderer {
+@@ -16,7 +18,7 @@ interface PrintRenderer {
  interface PrintRenderFrame {
    // Tells the RenderFrame to switch the CSS to print media type, render every
    // requested page, and then switch back the CSS to display media type.
 -  PrintRequestedPages();
-+  PrintRequestedPages(bool silent);
++  PrintRequestedPages(bool silent, mojo_base.mojom.DictionaryValue settings);
```
nornagon: what's even in these settings? Would it make sense to use a proper Mojo struct here rather than an untyped `DictionaryValue`?

patches/chromium/printing.patch
codebytere: whoops, i forgot to plumb this through correctly 🙈 
We do though - Chromium doesn't have a concept of silent printing so it's not included in the bespoke print settings the way that the other options are

patches/chromium/printing.patch
codebytere: I would say we do unfortunately need to leave this untyped - we could define a whole new struct outlining the settings integral to printing but i think that might introduce more arguably unncessary overhead. 

Since they're all interpreted as an untyped DictionaryValue in the end, this for now feels like a more tenable approach.

codebytere:
Hold on this for now - discovered a new problem i need to work out before this should be merged.

patches/chromium/printing.patch
codebytere: this is tested to be good now ✅

codebytere:
note: this can't be backported to 8 until https://github.com/electron/electron/pull/20839 lands

gordonel:
Can this be backported to v6?

gordonel:
@codebytere sorry for bothering, but can this be backported to Electron v5 or 6?

codebytere:
@gordonel it cant; this relies on changes which didn't exist in those versions of Chromium sadly.

gordonel:
Thanks.
-------- Original Message --------

Issue #20974, opened 2019-11-05
backgroundThrottling: false not obeyed in Electron 4+


mbman:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
 All versions starting from 4
* **Operating System:**
 Tested on Win 10, don't know if other platforms are affected.
* **Last Known Working Electron version:**
  3.1.13

### Expected Behavior
When an electron window is opened with `backgroundThrottling: false` set in `webPreferences` it shouldn't throttle `setTimeout` while window is minimized, but it does throttle it.

### Actual Behavior
Minimized window's `setTimeout `is throttled, despite disabling it.

### To Reproduce
Please load the Electron Fiddle: [https://gist.github.com/987afc63925e8ab2ad9c1e02dd7c3309](https://gist.github.com/987afc63925e8ab2ad9c1e02dd7c3309) and follow instructions in the opened window.

### Additional Information
Possible similar issue: https://github.com/electron/electron/issues/17402

codebytere:
@mbman i believe this was fixed in https://github.com/electron/electron/pull/20239, but was not backported. I'll see if I can get some backports up to address this in versions 7 and 6.

mbman:
@codebytere thanks for the info - does that PR also fixes `window.setBackgroundThrottling() `not working?

codebytere:
That'll be fixed by the PR i put up a little while ago: https://github.com/electron/electron/pull/21037

Slapbox:
#21037 says it's backported to `6.1.x`, but I'm getting `document.visibilityState === 'hidden'` when my window is `minimized` or `hidden` on `6.1.7` - any thoughts?

Much more strangely, static content in my `BrowserWindow`s disappears when I focus a separate window in my project and many other oddities, 3 shown below

![image](https://user-images.githubusercontent.com/6835891/71941216-f2605680-3186-11ea-8ec8-4cfa3ea92670.png)

Slapbox:
[...]

PR #21037, opened 2019-11-07
fix: backgroundThrottling rwh assignment

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/20974.
Follow-up to https://github.com/electron/electron/pull/20239

Previously when a render view was created the corresponding `RenderWidgetHost` was fetched with the wrong routing ID which meant that the `backgroundThrottling` option was not properly set for the RWH and thus the API had no effect.

This PR fixes that.

cc @miniak 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed throttling in `webContents.setBackgroundThrottling`.


shell/browser/api/atom_api_web_contents.cc
```diff
@@ -1286,31 +1286,26 @@ void WebContents::SetBackgroundThrottling(bool allowed) {
...
   auto* render_process_host = render_view_host->GetProcess();
-  if (!render_process_host) {
+  if (!render_process_host)
     return;
-  }
 
-  auto* render_widget_host_impl = content::RenderWidgetHostImpl::FromID(
-      render_process_host->GetID(), render_view_host->GetRoutingID());
-  if (!render_widget_host_impl) {
+  auto* impl = static_cast<content::RenderWidgetHostImpl*>(
```
deepak1556:
```suggestion
  auto* render_widget_host_impl = static_cast<content::RenderWidgetHostImpl*>(
```

codebytere:
hmm - looks like this is causing a CHECK failure now 🤔

deepak1556:
Based on https://cs.chromium.org/chromium/src/content/browser/renderer_host/render_widget_host_impl.h?l=818-822 it would make sense that our version would crash given it doesn't check for a valid process when extracting the widget associated with the view.

I think the solution here would be fix the underlying routing id issue by extracting the widget from `RenderFrameHost`. /cc @zcbenz for second opinion.

miniak:
This PR still does not fix #20974. It only fixes `requestAnimationFrame` throttling.

shell/browser/api/atom_api_web_contents.cc
```diff
@@ -1286,31 +1286,26 @@ void WebContents::SetBackgroundThrottling(bool allowed) {
...
     return;
-  }
 
   auto* render_view_host = contents->GetRenderViewHost();
-  if (!render_view_host) {
+  if (!render_view_host)
     return;
-  }
 
   auto* render_process_host = render_view_host->GetProcess();
```
miniak: `render_process_host` is not used anymore, please remove

shell/browser/api/atom_api_web_contents.cc
```diff
@@ -1286,31 +1286,26 @@ void WebContents::SetBackgroundThrottling(bool allowed) {
   background_throttling_ = allowed;
 
   auto* contents = web_contents();
```
miniak: this is also not necessary, we don't check the return value of `web_contents()` anywhere else

deepak1556:
@miniak I think this PR needs to address https://github.com/electron/electron/pull/21037#issuecomment-551223741 before making any further changes.

miniak:
@deepak1556, @codebytere this issue is quite high priority for us

deepak1556:
@miniak I am working on this today.

deepak1556:
Update: The reason timers aren't affected by this change is because they are controlled separately on the blink end using `blink::PageScheduler` and they throttle based on the page visibility, you can disable this throttling by using `--disable-features="WebContentsOcclusion"` but this will also affect the value of `document.visibilityState`.

I will update the disable_hidden.patch to account for this code path as well without impacting the value of `document.visibilityState`

patches/chromium/feat_allow_disbaling_blink_scheduler_throttling_per_renderview.patch
```diff
@@ -0,0 +1,155 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: deepak1556 <hop2deep@gmail.com>
+Date: Fri, 29 Nov 2019 16:08:14 -0800
+Subject: feat: allow disbaling blink scheduler throttling per RenderView
```
miniak: typo `disbaling` -> `disabling`

deepak1556:
Please merge this after merging https://github.com/electron/electron/pull/21343 for a clean backport on all tags listed here.

deepak1556:
failing tests are not related, merging.

PR #20904, opened 2019-11-01
fix: don't copy tray image when it's set

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/20850.

We were making unnecessary copies of NSImage instances, which lead to unnecessary RAM consumption.

Benchmarks for a few minutes of runtime with [this repro](https://github.com/D0miH/electron-bug-repo):

Before:
```js
{                                                                              
  rss: 195792896,
  heapTotal: 6774784,
  heapUsed: 4962656,
  external: 2211750
}
```

After:
```js
{                                                                              
  rss: 155697152,
  heapTotal: 6774784,
  heapUsed: 5315472,
  external: 2211750
}
```

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Fixed a memory leak issue when setting Tray images.

PR #21040, opened 2019-11-07
fix: don't copy tray image when it's set

codebytere:
Backport of #20904

See that PR for details.

Notes: Fixed a memory leak issue when setting Tray images.


codebytere:
I don't actually think we can backport this here - we changed the way we set images/titles
from 
```cpp
title_.reset(title);
```
to 
```cpp
[[statusItem_ button] setTitle:title];
```

so i think this is a no-fly now

codebytere:
Closing given above - we do need to copy here.

PR #21055, opened 2019-11-08
fix: allow reading body from non-2xx responses in net.request

kwonoj:
#### Description of Change

Context: https://cs.chromium.org/chromium/src/services/network/public/cpp/simple_url_loader.h?dr=C&g=0&l=262-276

`SimpleURLLoader` by default do not allow non-2xx http response as results, raises `ERR_HTTP_RESPONSE_CODE_FAILURE` and it makes attempt to read response body for those response not possible in net module's clientrequest.

This PR changes urlloader configuration to allow - this behavior matches to previous `net` module implementation, as well as browser native like `fetch`.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed net module request raises error when non-2xx response received.

### DIFF
--- a/spec-main/api-net-spec.ts
+++ b/spec-main/api-net-spec.ts
@@ -717,6 +717,59 @@ describe('net module', () => {
       expect(abortsEmitted).to.equal(1, 'request should emit exactly 1 "abort" event')
     })
 
+    it('shold allow to read response body from non-2xx response', (done) => {
+      const bodyData = randomString(kOneKiloByte)
+      respondOnce.toSingleURL((request, response) => {
+        response.statusCode = 404
+        response.end(bodyData)
+      }).then(serverUrl => {
+        let requestResponseEventEmitted = false
+        let responseDataEventEmitted = false
+        let responseEndEventEmitted = false
+        let requestCloseEventEmitted = false
+
+        function maybeDone (done: () => void) {
+          if (!requestCloseEventEmitted || !responseEndEventEmitted) {
+            return
+          }
+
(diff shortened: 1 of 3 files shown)


nornagon:
Hm, looks like the tests are consistently failing now. Perhaps this had some unintended consequences?

kwonoj:
Relevant test case failures are corrected, but each time rebase to master some of tests are failing inconsistently, not sure if it's actually related tests & how to fix.

spec-main/api-net-spec.ts
```diff
@@ -769,6 +774,59 @@ describe('net module', () => {
...
+        let responseDataEventEmitted = false
+        let responseEndEventEmitted = false
+        let requestCloseEventEmitted = false
+
+        function maybeDone (done: () => void) {
+          if (!requestCloseEventEmitted || !responseEndEventEmitted) {
+            return
+          }
+
+          expect(requestResponseEventEmitted).to.equal(true)
```
nornagon: Calling `expect()` from an event / callback will result in the entire test process being aborted, instead of just failing the test. Please refactor similarly to https://github.com/electron/electron/pull/21265/files#diff-c31378831f4597a16ed943c2f7060323R566

kwonoj: this is mostly copy-paste from existing test cases like https://github.com/electron/electron/blob/master/spec-main/api-net-spec.ts#L337 fyi

nornagon: yep, these changed with https://github.com/electron/electron/pull/21265

Issue #21058, opened 2019-11-09
The printerinfo status does not determine whether the printer is online or offline


shaonianla1997:
### Preflight Checklist

* [ ] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [ ] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [ ] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description

when calling webContents.getPrinters(), printer status does not determine whether the printer is online or offline.

codebytere:
We are only able to plumb through the data that's available to use through Chromium, and as you can [see here](https://cs.chromium.org/chromium/src/printing/backend/print_backend.h?sq=package:chromium&dr=CSs&g=0&l=40-44) we don't have the ability to access printer online status. Should something on their end change, we're more than happy to revisit this in the future.

PR #21081, opened 2019-11-11
feat: enhance PrinterInfo object values

codebytere:
#### Description of Change

Resolves https://github.com/electron/electron/issues/21058.

Improves both documentation and returned object data for `webContents.getPrinters()` in the following ways:
1) Adds a `displayName` field to distinguish the name of the printer as understood by the OS from the name of the printer as shown in Print Preview.
2) Adds descriptions for `PrinterInfo` structure data.

cc @miniak @nornagon @MarshallOfSound 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Improved the returned types and descriptions for the `PrinterInfo` API structure returned by `webContents.getPrinters`.

### DIFF
--- a/docs/api/structures/printer-info.md
+++ b/docs/api/structures/printer-info.md
@@ -1,9 +1,13 @@
 # PrinterInfo Object
 
-* `name` String
-* `description` String
-* `status` Number
-* `isDefault` Boolean
+* `name` String - the name of the printer as understood by the OS.
+* `displayName` String - the name of the printer as shown in Print Preview.
+* `description` String - a longer description of the printer's type.
+* `status` Number - the current status of the printer.
+* `isDefault` Boolean - whether or not a given printer is set as the default printer on the OS.
+* `options` Object - an object containing a variable number of platform-specific printer information.
+
+The number represented by `status` means different things on different platforms: on Windows it's potential values can be found [here](https://docs.microsoft.com/en-us/windows/win32/printdocs/printer-info-2), and on Linux and macOS they can be found [here](https://www.cups.org/doc/cupspm.html).
 
 ## Example
 
@@ -12,13 +16,14 @@ may be different on each platform.
 
 ```javascript
 {
-  name: 'Zebra_LP2844',
-  description: 'Zebra LP2844',
+  name: 'Austin_4th_Floor_Printer___C02XK13BJHD4',
+  displayName: 'Austin 4th Floor Printer @ C02XK13BJHD4',
+  description: 'TOSHIBA ColorMFP',
   status: 3,
   isDefault: false,
   options: {
     copies: '1',
-    'device-uri': 'usb://Zebra/LP2844?location=14200000',
+    'device-uri': 'dnssd://Austin%204th%20Floor%20Printer%20%40%20C02XK13BJHD4._ipps._tcp.local./?uuid=71687f1e-1147-3274-6674-22de61b110bd',
     finishings: '3',
     'job-cancel-after': '10800',
     'job-hold-until': 'no-hold',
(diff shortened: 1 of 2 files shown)


shell/browser/api/atom_api_web_contents.cc
```diff
@@ -124,15 +124,34 @@
 
 namespace gin {
 
+std::string GetPrinterDescription(int status) {
+  switch (status) {
+    case 0:
+      return "idle";
+    case 1:
```
nornagon: is this intentionally falling-through? If so, please use `FALLTHROUGH`.
```suggestion
    case 1:
      FALLTHROUGH;
```

shell/browser/api/atom_api_web_contents.cc
```diff
@@ -124,15 +124,34 @@
...
+    case 2:
+      return "error";
+    case 3:
+      return "offline";
+    case 4:
+      return "in use";
+    case 5:
+      return "paused";
```
nornagon: Where did you find the meaning of these codes?

docs/api/structures/printer-info.md
```diff
@@ -1,9 +1,11 @@
 # PrinterInfo Object
 
-* `name` String
-* `description` String
-* `status` Number
-* `isDefault` Boolean
+* `name` String - the name of the printer as understood by the OS.
+* `displayName` String - the name of the printer as shown in Print Preview.
+* `description` String - a longer description of the printer's type.
+* `status` String - the current status of the printer. Can be 'idle', 'paused', 'in use', 'offline', or 'unknown'.
```
nornagon: is changing the status from a number to a string a breaking change?

shell/browser/api/atom_api_web_contents.cc
codebytere: they're not documented anywhere unfortunately - i manually tested all of them in order to determine the statuses they corresponded to. This was done via triggering local printers, checking statuses via system preferences ui:

Screen Shot 2019-11-11 at 6 01 37 PM

and then mapping those statuses to the numbers that was being returned in the object.

docs/api/structures/printer-info.md
codebytere: On principle yes - but i would also say that the previous output was essentially unintelligible in that we neither described what 'status' meant nor provided a way to interpret it.

shell/browser/api/atom_api_web_contents.cc
nornagon: - Do these have the same meaning on macOS, Windows and Linux?
- Is this an exhaustive list?
- Do we want to become the documentation for this?

If no to any of the above, it seems like it might be better for us (though obviously a worse API) to just pass through the number directly and pass the buck to the app developer to interpret the status code.

codebytere: primarily - i changed this in service of wanting to move away from APIs that essentially just dump implementation details with no filtering or parsing. What would you think about trying to confirm this with Chromium developers to ensure that my current statuses are correct?

nornagon: Yeah, it'd be great to get some clarity on this. I'm down with providing a better API but I want to be careful about how we do it. e.g. i have no idea if any of these status codes might have multiple or different meanings in other situations...

docs/api/structures/printer-info.md
nornagon: Ha, fair enough. It seems like an improvement, I agree. Is it small enough of a change that we can skip putting it in the breaking changes doc?

shell/browser/api/atom_api_web_contents.cc
codebytere: i talked to some folks who gave me some more context and informed me that windows and unix use different meanings, so i've instead opted to retain the numeric value and instead link to locations users can go to parse meanings from the integers :)

docs/api/structures/printer-info.md
codebytere: (resolved, no longer applicable!)

PR #21086, opened 2019-11-12
build: make CI faster, magic contained within

MarshallOfSound:
So I tried a lot of things here, mostly based off of the idea @jkleinsc had around caching and restoring the out directory.  On linux that works really well, on macOS the network speed is lower (appears to be only gigabit) so we have to make trade offs around out size vs cache upload / download time.

I think I found the sweet spot.  Raising this PR for 👀 now, I just reset the cache keys so the first run on this PR will be the old slow path.  The second run will be ultra-fast.

Build time estimates are:
Linux ~13 minutes from start to tests complete
macOS ~28 minutes from start to tests complete

Notes: no-notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ parameters:
 # Build machines configs.
 docker-image: &docker-image
   docker:
-    - image: electronjs/build:697b894f36d127155e020f4e8ad4b2e5f6a09613
+    - image: electronbuilds/electron:0.0.10
 
 machine-linux-medium: &machine-linux-medium
   <<: *docker-image
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -693,7 +693,7 @@ step-restore-out-cache: &step-restore-out-cache
     paths:
       - ./src/out/Default
     keys:
-      - v6-out-cache-{{ checksum "src/electron/.depshash" }}-{{ checksum "src/electron/.depshash-target" }}
+      - v7-out-cache-{{ checksum "src/electron/.depshash" }}-{{ checksum "src/electron/.depshash-target" }}
     name: Restoring out cache
 
 step-set-git-cache-path: &step-set-git-cache-path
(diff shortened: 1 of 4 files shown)


MarshallOfSound:
Hm, looks like there's a double-build going on due to my size reduction logic.  Will fix it up tomorrow.

MarshallOfSound:
Ok I got this working.  I could explain how it works but I'm using a bunch of new tricks in here that would be _quite_ a write up.

image

.circleci/config.yml
```diff
@@ -684,10 +694,18 @@ step-maybe-restore-git-cache: &step-maybe-restore-git-cache
     paths:
       - ~/.gclient-cache
     keys:
-      - v2-gclient-cache-{{ arch }}-{{ checksum "src/electron/.circle-sync-done" }}-{{ checksum "src/electron/DEPS" }}
-      - v2-gclient-cache-{{ arch }}-{{ checksum "src/electron/.circle-sync-done" }}
+      - v2-gclient-cache-{{ checksum "src/electron/.circle-sync-done" }}-{{ checksum "src/electron/DEPS" }}
```
nornagon: why is the arch no longer part of the cache key?

BUILD.gn
```diff
@@ -1374,12 +1374,18 @@ dist_zip("electron_chromedriver_zip") {
...
+  public_deps = mksnapshot_deps
+}
+
 dist_zip("electron_mksnapshot_zip") {
-  data_deps = [
-    "//v8:mksnapshot($v8_snapshot_toolchain)",
-    "//tools/v8_context_snapshot:v8_context_snapshot_generator",
-    ":licenses",
-  ]
+  data_deps = mksnapshot_deps
```
nornagon: maybe just make this depend on `electron_mksnapshot`?

MarshallOfSound: I did that first, it was not impressed with the idea.

.circleci/config.yml
MarshallOfSound: Because it's just a git blob, not arch specific.  Also we checkout macos on a linux machine anyway 🤷‍♂

PR #21096, opened 2019-11-12

nornagon:
#### Description of Change
This is a followup to #20954, implementing the 'login' event on `net.ClientRequest`. Unfortunately this requires patching Chromium.

#### Checklist

- [x] `npm test` passes

#### Release Notes


shell/browser/api/atom_api_url_request.cc
```diff
@@ -166,7 +173,9 @@ class ChunkedDataPipeGetter : public UploadDataPipeGetter,
   mojo::ReceiverSet<network::mojom::ChunkedDataPipeGetter> receiver_set_;
 };
 
-URLRequest::URLRequest(gin::Arguments* args) : weak_factory_(this) {
+URLRequest::URLRequest(gin::Arguments* args)
+    : id_(GetAllRequests().Add(this)), weak_factory_(this) {
```
deepak1556: Can we maintain an IDMap of weak ptrs rather than raw ptrs , we already check before using it in `AuthResponder::OnAuthRequired`, this will just make it more stronger.

shell/browser/atom_browser_context.cc
```diff
@@ -342,6 +344,42 @@ AtomBrowserContext::GetURLLoaderFactory() {
...
+    }
+  }
+
+  mojo::Receiver<network::mojom::TrustedAuthClient> receiver_;
+};
+
+void AtomBrowserContext::OnLoaderCreated(
+    int32_t request_id,
+    mojo::PendingReceiver<network::mojom::TrustedAuthClient> auth_client) {
+  new AuthResponder(std::move(auth_client));
```
deepak1556: Who owns `AuthResponder` ?

spec-main/api-web-contents-spec.ts
```diff
@@ -1534,7 +1534,7 @@ describe('webContents module', () => {
         }
         response
           .writeHead(401, { 'WWW-Authenticate': 'Basic realm="Foo"' })
-          .end()
+          .end('401')
```
deepak1556: confused about why this was needed, until i saw the below test where you check the page content :)

shell/browser/api/atom_api_url_request.cc
nornagon: no point in having a weak ptr, the object removes itself from the map on destruction so a weak ptr wouldn't add anything here.

shell/browser/atom_browser_context.cc
nornagon: good question... it should probably be a [strong binding](https://cs.chromium.org/chromium/src/mojo/public/cpp/bindings/strong_binding.h?l=6&rcl=4644ed252018ddf3800d5e652bb7ab376b57c021).

nornagon: by which i mean, [self-owned receiver](https://cs.chromium.org/chromium/src/mojo/public/cpp/bindings/self_owned_receiver.h?l=6&rcl=4644ed252018ddf3800d5e652bb7ab376b57c021)

PR #21097, opened 2019-11-12
fix: implement login event for WebContents

nornagon:
Backport of #20954. See that PR for details.

Notes: Fixed the 'login' event not being emitted when HTTP Basic auth was requested from a WebContents.

### DIFF
--- a/shell/browser/login_handler.cc
+++ b/shell/browser/login_handler.cc
@@ -10,7 +10,7 @@
 #include <vector>
 
 #include "base/callback.h"
-#include "mate/dictionary.h"
+#include "native_mate/dictionary.h"
 #include "shell/browser/api/atom_api_web_contents.h"
 #include "shell/common/gin_converters/callback_converter_gin_adapter.h"
 #include "shell/common/gin_converters/net_converter_gin_adapter.h"
--- a/shell/browser/login_handler.cc
+++ b/shell/browser/login_handler.cc
@@ -10,7 +10,7 @@
 #include <vector>
 
 #include "base/callback.h"
-#include "gin/dictionary.h"
+#include "mate/dictionary.h"
 #include "shell/browser/api/atom_api_web_contents.h"
 #include "shell/common/gin_converters/callback_converter_gin_adapter.h"
 #include "shell/common/gin_converters/net_converter_gin_adapter.h"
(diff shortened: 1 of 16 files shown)


shell/browser/login_handler.cc
```diff
@@ -5,96 +5,82 @@
...
-  DCHECK_CURRENTLY_ON(BrowserThread::UI);
-  // TODO(deepak1556): fix with network service
-  // tracking issue: #19602
-  CHECK(false) << "fix with network service";
-  return web_contents_getter_.Run();
+  bool default_prevented =
+      api_web_contents->Emit("login", std::move(details), auth_info,
+                             base::BindOnce(&LoginHandler::CallbackFromJS,
+                                            weak_factory_.GetWeakPtr()));
+  if (!default_prevented) {
```
deepak1556: Minor nit: Although there is no explicit need to check for the auth callback as it is not moved between different sequences unlike pre network service. But since other release branches have this check, better to maintain consistency.

```
if (!default_prevented && auth_required_callback_)
```

shell/browser/login_handler.cc
```diff
@@ -5,96 +5,82 @@
...
-                           const base::string16& password) {
-  DCHECK_CURRENTLY_ON(BrowserThread::IO);
-  if (!auth_callback_.is_null()) {
-    credentials_->Set(username, password);
-    std::move(auth_callback_)
-        .Run(net::NetworkDelegate::AUTH_REQUIRED_RESPONSE_SET_AUTH);
-  }
+void LoginHandler::CallbackFromJS(base::string16 username,
+                                  base::string16 password) {
+  std::move(auth_required_callback_)
```
deepak1556: same here

shell/browser/login_handler.cc
nornagon: I was planning to merge that in with the followup here: https://github.com/electron/electron/pull/21135 but i can add it here.

nornagon: also, there _is_ a need to check this callback. because it's a OnceCallback and we `std::move` it, if the JS calls the callback synchronously and doesn't call `event.preventDefault`, then we can hit this case and end up attempting to call an empty callback. There's a test for this added in #21135.

PR #13039, opened 2018-05-22
Remove command line argument black-list

alespergl:
Per our agreement we will instead block all arguments following a URL.

Users of apps that accept multiple URLs, or other arguments following a URL, will need to use a `--` argument as a prefix.

PR #21110, opened 2019-11-13
feat: enable passing Node.js cli flags

codebytere:
#### Description of Change

This PR enables passing safelisted cli flags to Node.js to allow doing useful things like enabling experimental Node.js features.

For reasons outlined below, it seems safer to make different allowances for Electron itself vs Electron in `ELECTRON_RUN_AS_NODE` mode.  As such, this PR only allows passing `DebugOptions` via cli for prototypical Electron, and allows all options in `ELECTRON_RUN_AS_NODE` mode.

I've upstreamed the patch necessary to enable this at https://github.com/nodejs/node/pull/30466.

cc @nornagon @MarshallOfSound @ckerr

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Enable native Electron handling and passing of Node.js cli options.


codebytere:
Current issue/consideration: Node.js will spit back a non-zero exit code if any non-node switches are appended to the command line, which is the case by default for some switches like `--allow-file-access-from-files` and `--enable-avfoundation` set in `atom_main_delegate`.

The Node.js flags passed will still work and be correctly set, but this makes me think that a more robust approach might instead be to safelist allowed flags instead of skipping a preset list of disallowed flags, which would ameliorate this issue entirely. What do you all think?

deepak1556:
What is the difference between this and the `NODE_OPTIONS` support we already provide ? AFAIU you can pass these command line params through options https://nodejs.org/api/cli.html#cli_node_options_options ?

I am slightly not okay with maintaining any kind of blacklist for volatile structures, we tried to do that for chromium in the past but quickly proved inefficient https://github.com/electron/electron/pull/13039.

codebytere:
At the moment, we introduce a lot of extra patch surface by exposing `DebugOptions`, as we do currently do a small amount of parsing cli flags in this manner. As ux goes - passing flags in this manner is a lot more user-friendly and is most likely what users expect, and this will, more importantly allow us to remove a notable amount of patch surface and simplify the way we're passing settings to node for the inspector agent, a process i've already started [here](https://github.com/electron/electron/commit/d18811ba10f2a421986d8a089fafacadd506b696).

I would be in favor of moving away from a blacklist approach and into a whitelist approach based on concerns you outlined above, as i also think that would also be safer generally.

deepak1556:
> What is the difference between this and the NODE_OPTIONS support we already provide ? AFAIU you can pass these command line params through options https://nodejs.org/api/cli.html#cli_node_options_options ?

I still don't understand why this is needed over `NODE_OPTIONS`.


I think we can agree inspector related flags are special case and it wouldn't be hard to handle them separately. But I don't see why we have to bring in the whole set of node cli switches support when there is already another way to achieve it `NODE_OPTIONS` ?  if there are flags that aren't allowed in this mode users should get that fixed in Node.js and not through us.

deepak1556:
> Node.js will not extern DebugOptions, whereas the patch in this PR has already been approved upstream

If the aim is to reduce the patch around `DebugOptions`, we can handle only the cli arguments related to inspector through this new method. I am still -1 on handling all node cli options through this new method.

codebytere:
I'd be happy to adapt this PR to just allow the set of `DebugOptions`-specific options and then we could perhaps open the wider conversation of what we want to eventually enable for users, be it flags, NODE_OPTIONS, or both?

I would say that i personally feel `NODE_OPTIONS` was added as a shortest-path approach to enable handling these options at all, but is less ergonomic/usable an approach in contrast to being able to pass those selfsame options via bespoke flags

deepak1556:
> I'd be happy to adapt this PR to just allow the set of DebugOptions-specific options and then we could perhaps open the wider conversation of what we want to eventually enable for users, be it flags, NODE_OPTIONS, or both?

Yup that would be great, thanks!

> I would say that i personally feel NODE_OPTIONS was added as a shortest-path approach to enable handling these options at all, but is less ergonomic/usable an approach in contrast to being able to pass those selfsame options via bespoke flags

True I agree with that as well but that comes to our personal choice, what I am trying to avoid is double parsing stuff, we already maintain a code path with blacklisted options for `NODE_OPTIONS` which I am not so happy about, I am declined to add another layer on top of that. Also node gave this feature support despite cli because its easier to pass across environments to child processes, so I think its sufficient.

The primary goal here is to eliminate the patches around `DebugOptions` for which maintaining a whitelist of node inspector related flags are much easier and solves the issue at hand rather than maintaining an ever growing list of all node cli options.

docs/api/command-line-switches.md
```diff
@@ -16,78 +16,64 @@ app.on('ready', () => {
 })
 ```
 
-## --ignore-connections-limit=`domains`
```
codebytere: I organized these flags to group Electron-specific flags separately from Node.js flags and also alphabetize them.

shell/app/node_main.cc
```diff
@@ -5,11 +5,15 @@
 #include "shell/app/node_main.h"
 
 #include <memory>
+#include <set>
```
ckerr:
```suggestion
#include <unordered_set>
```

shell/app/node_main.cc
```diff
@@ -5,11 +5,15 @@
...
 #include <string>
 #include <utility>
+#include <vector>
 
 #include "base/command_line.h"
 #include "base/feature_list.h"
+#include "base/strings/string_util.h"
+#include "base/strings/utf_string_conversions.h"
```
ckerr: need to include string_piece.h too

shell/app/node_main.cc
```diff
@@ -29,6 +33,44 @@
...
+      "--force-fips", "--enable-fips"};
+
+  const auto argv = base::CommandLine::ForCurrentProcess()->argv();
+  std::vector<std::string> args;
+  args.reserve(argv.size());
+  for (const auto& arg : argv) {
+#if defined(OS_WIN)
+    const auto option = base::UTF16ToUTF8(arg);
+#else
+    const auto option = arg;
```
ckerr:
```suggestion
    const auto& option = arg;
```

shell/app/node_main.cc
```diff
@@ -29,6 +33,44 @@
 #include "shell/common/crash_reporter/crash_reporter_win.h"
 #endif
 
+namespace {
+
+// Initialize Node.js cli options to pass to Node.js
+// See https://nodejs.org/api/cli.html#cli_options
+void SetNodeCliFlags() {
+  // Options that are unilaterally disallowed
+  std::unordered_set<base::StringPiece, base::StringPieceHash> disallowed = {
```
ckerr: this can be `const`

shell/common/node_bindings.cc
```diff
@@ -136,7 +136,51 @@ bool IsPackagedApp() {
 #endif
 }
 
+// Initialize Node.js cli options to pass to Node.js
+// See https://nodejs.org/api/cli.html#cli_options
+void SetNodeCliFlags() {
+  // Only allow DebugOptions in non-ELECTRON_RUN_AS_NODE mode
+  std::unordered_set<base::StringPiece, base::StringPieceHash> allowed = {
```
ckerr: const

shell/common/node_bindings.cc
```diff
@@ -136,7 +136,51 @@ bool IsPackagedApp() {
...
+  const auto argv = base::CommandLine::ForCurrentProcess()->argv();
+  std::vector<std::string> args;
+  args.reserve(argv.size() + 1);
+  args.push_back("electron");
+
+  for (const auto& arg : argv) {
+#if defined(OS_WIN)
+    std::string option = base::UTF16ToUTF8(arg);
+#else
+    std::string option = arg;
```
ckerr: both versions of `option` can be const

the one in this #ifdef can also be a reference

shell/common/node_bindings.cc
```diff
@@ -136,7 +136,51 @@ bool IsPackagedApp() {
...
+#endif
+    auto stripped = option.substr(0, option.find("="));
```
ckerr:
```suggestion
    const auto stripped = base::StringPiece(option).substr(0, option.find('='));
```

* const

* avoid creating a new string by using a StringPiece

* faster delimiter finding by using `'='` instead of `"="`

shell/common/node_bindings.cc
```diff
@@ -136,7 +136,51 @@ bool IsPackagedApp() {
...
+    if (allowed.find(stripped) != allowed.end())
```
ckerr:
```suggestion
    if (allowed.count(stripped) != 0)
```

_(nit)_ your version is fine too but this is more idiomatic. (In C++20 we'll have `includes()` which is equivalent to count()!=0)

shell/common/node_bindings.cc
```diff
@@ -136,7 +136,51 @@ bool IsPackagedApp() {
...
+      args.push_back(option);
+  }
+
+  std::vector<std::string> exec_args;
+  std::vector<std::string> errors;
+  const int exit_code = ProcessGlobalArgs(&args, &exec_args, &errors,
```
ckerr: As above, we don't need `exec_args`

shell/common/node_bindings.cc
```diff
@@ -136,7 +136,51 @@ bool IsPackagedApp() {
...
+      "--inspect-brk-node", "--inspect-publish-uid",
+  };
+
+  // TODO(codebytere): We need to set the first entry in args to the
+  // program name owing to src/node_options-inl.h#L286-L290 but this is
+  // redundant and so should be refactored upstream.
+  const auto argv = base::CommandLine::ForCurrentProcess()->argv();
+  std::vector<std::string> args;
+  args.reserve(argv.size() + 1);
+  args.push_back("electron");
```
ckerr:
```suggestion
  args.emplace_back("electron");
```

shell/app/node_main.cc
```diff
@@ -29,6 +33,44 @@
...
+    auto stripped = base::StringPiece(option).substr(0, option.find('='));
+    if (disallowed.count(stripped) != 0) {
+      LOG(ERROR) << "The Node.js cli flag " << stripped
+                 << " is not supported in Electron";
+    } else {
+      args.push_back(option);
+    }
+  }
+
+  std::vector<std::string> exec_args;
```
ckerr: exec_args can be removed

shell/app/node_main.cc
```diff
@@ -29,6 +33,44 @@
...
+  const auto argv = base::CommandLine::ForCurrentProcess()->argv();
+  std::vector<std::string> args;
+  args.reserve(argv.size());
+  for (const auto& arg : argv) {
+#if defined(OS_WIN)
+    const auto option = base::UTF16ToUTF8(arg);
+#else
+    const auto option = arg;
+#endif
+    auto stripped = base::StringPiece(option).substr(0, option.find('='));
```
ckerr:
```suggestion
    const auto stripped = base::StringPiece(option).substr(0, option.find('='));
```

shell/common/node_bindings.cc
```diff
@@ -136,7 +137,51 @@ bool IsPackagedApp() {
...
+
+  // TODO(codebytere): We need to set the first entry in args to the
+  // process name owing to src/node_options-inl.h#L286-L290 but this is
+  // redundant and so should be refactored upstream.
+  args.reserve(argv.size() + 1);
+  args.emplace_back("electron");
+
+  for (const auto& arg : argv) {
+#if defined(OS_WIN)
+    const auto& option = base::UTF16ToUTF8(arg);
```
ckerr: Same here

shell/app/node_main.cc
```diff
@@ -64,6 +111,9 @@ int NodeMain(int argc, char* argv[]) {
     // Explicitly register electron's builtin modules.
     NodeBindings::RegisterBuiltinModules();
 
+    // Parse and set Node.js cli flags.
+    SetNodeCliFlags();
+
     int exec_argc;
     const char** exec_argv;
     node::Init(&argc, const_cast<const char**>(argv), &exec_argc, &exec_argv);
```
deepak1556: Can't we have `node::Init` do the call to `ProcessGlobalArgs` since it has access to commandline and also we are allowing all node cli flags for this mode, it only makes sense for node to handle rather than us.

Only reason we want it here is for the crypto related flags, what happens if we let them through to node ? We don't build with those crypto config support, so ideally node should error in those situations which is acceptable behavior as an embedder.

codebytere: We don’t just want it here for crypto-related flags - without explicitly setting these from our command line they simply don’t get set. `ProcessGlobalArgs` doesn’t have access to the command line - it just parses what we pass to it via `argv`

deepak1556: `argv` passed to `node::Init` is the same one `base::CommandLine` uses, check the start of this function, why would they not be available to node to process further ?

codebytere: Right, i see that, but do you feel it's _bad_  to save users from experiencing what will be deterministic and unexpected failures?

deepak1556: What does node throw at the moment when we don't handle the parsing ? Just an exit code and no descriptive erro ?

codebytere: no - it'll fail when the user then goes to invoke the crypto-related methods. Since we're bundling BoringSSL, which is not always clear to users, I personally feel that it's fair for us to assume responsibility for helping users not run into failing functionality we're already clearly aware will be failing when they go to invoke it.

codebytere:
I'll take a look at these test failures soon but as soon as that's handled this should be g2g.

zcbenz:
I'm adding `wip` tag since there are pending works (fixing tests) in this PR.

codebytere:
@zcbenz thanks - i should be able to get back to this soon!

jkleinsc:
@codebytere rebasing this against master should resolve the linux failures

PR #21118, opened 2019-11-13
chore: remove openssl from process.versions

codebytere:
#### Description of Change

Previously, we were leaving the OpenSSL version in Node.js in our `process.versions.openssl` since we set `node_use_openssl = true` in our `BUILD.gn` file for parity with Node.js, but this is confusing to users since that does not exist in Electron. This PR checks for `process.versions.openssl` being set, and then if it is, removes it. In the future, i'd like to actually show the BoringSSL version there instead, but I feel that can be done in a separate PR.

This also has the handy side effect of letting us remove a bunch of disabled Node.js crypto tests.

cc @MarshallOfSound @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


shell/app/node_main.cc
```diff
@@ -113,6 +113,10 @@ int NodeMain(int argc, char* argv[]) {
     gin_helper::Dictionary versions;
     if (process.Get("versions", &versions)) {
       versions.SetReadOnly(ELECTRON_PROJECT_NAME, ELECTRON_VERSION_STRING);
+
+      // We don't build with OpenSSL
+      if (versions.Has("openssl"))
+        versions.Delete("openssl");
```
poiru: Do we need `Has()`? I think `Delete()` should just no-ops if the key doesn't exist.

codebytere: oh i think you're right - will update.

codebytere:
Do you see any possible path forward here? It would be nice to indicate to users that they are not actually getting OpenSSL in a more programmatic way, which is what this intended to accomplish.

nornagon:
You can easily determine that you're not running openssl by checking `process.versions.electron`. I'm not aware of any other major Node-like environments that ship non-OpenSSL, so that seems like a reasonable way to check to me. Even better would be to check for the specific features that you need, since BoringSSL is _mostly_ compatible with OpenSSL.

PR #21099, opened 2019-11-13
fix: webRequest should be able to modify CORS headers

zcbenz:
#### Description of Change

Fix https://github.com/electron/electron/issues/20923.

With NetworkService `webRequest` module is not able to modify CORS headers by default, this PR turns on the ability in Electron.

The code path is basically the same with the `extraHeaders: true` option turned on in Chromium.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Fix `webRequest` module unable to modify CORS headers.

PR #21123, opened 2019-11-14
fix: webRequest should be able to modify CORS headers (7-1-x)

zcbenz:
Backport of #21099.

Notes: Fix `webRequest` module unable to modify CORS headers.


zcbenz:
The fix somehow does not work for 7.x, I will look into it later.

zcbenz:
It turns out the added tests have to run in main process because of `remote`'s limitation, I have also backported https://github.com/electron/electron/pull/19992 (`test: move webRequest spec to main runner`) to fix the failing tests.

PR #21133, opened 2019-11-14

nornagon:
Backport of #21096. See that PR for details.


nornagon:
NB: this includes https://github.com/electron/electron/pull/21098 to avoid having to resolve conflicts twice. Will rebase once that lands.

true-zero:
This issue was never resolved.

true-zero:
https://github.com/electron/electron/issues/21323

nornagon:
@true-zero do you mean you're still encountering problems with the 'login' event? Or do you mean that #21323 should now be closed?

PR #21135, opened 2019-11-14

nornagon:
Backport of #21096. See that change for details.


nornagon:
NB. this also includes https://github.com/electron/electron/pull/21097, to reduce conflicts. Once that backport lands, I'll rebase this.

PR #21157, opened 2019-11-17
fix: prevent urlrequest exception when upload

### LINKED ISSUE #21151: Intermittent crash on net module PUT request
### Preflight Checklist

### Issue Details

* **Electron Version:**
7.1.1
* **Operating System:**
  * 
  * 6.x

Windows 10 / Mac Os catalina

### Actual Behavior
DCHECK failure

```
[9628:1115/193639.015:FATAL:atom_api_url_request_ns.cc(467)] Check failed: producer_.
Backtrace:
        base::debug::CollectStackTrace [0x00007FF7A1C5B152+18] (o:\base\debug\stack_trace_win.cc:284)
        base::debug::StackTrace::StackTrace [0x00007FF7A1C56F72+18] (o:\base\debug\stack_trace.cc:203)
        logging::LogMessage::~LogMessage [0x00007FF7A1C75863+115] (o:\base\logging.cc:630)
        electron::api::URLRequestNS::DoWrite [0x00007FF79F0C0E83+99] (o:\electron\shell\browser\api\atom_api_url_request_ns.cc:468)
        electron::api::URLRequestNS::Write [0x00007FF79F0C0271+881] (o:\electron\shell\browser\api\atom_api_url_request_ns.cc:276)
        mate::internal::Invoker<mate::internal::IndicesHolder<0,1,2>,electron::api::URLRequestNS *,v8::Local<v8::Value>,bool>::DispatchToCallback<bool> [0x00007FF79F0C3E7A+102] (o:\electron\native_mate\native_mate\function_template.h:200)
        mate::internal::Dispatcher<bool (electron::api::URLRequestNS *, v8::Local<v8::Value>, bool)>::DispatchToCallback [0x00007FF79F0C3DAF+159] (o:\electron\native_mate\native_mate\function_template.h:244)
        v8::internal::FunctionCallbackArguments::Call [0x00007FF7A082CF98+1080] (o:\v8\src\api\api-arguments-inl.h:159)
        v8::internal::`anonymous namespace'::HandleApiCallHelper<0> [0x00007FF7A07C1D9A+826] (o:\v8\src\builtins\builtins-api.cc:113)
        v8::internal::Builtin_Impl_HandleApiCall [0x00007FF7A07C03C5+453] (o:\v8\src\builtins\builtins-api.cc:141)
        v8::internal::Builtin_HandleApiCall [0x00007FF7A07BFEC0+160] (o:\v8\src\builtins\builtins-api.cc:129)
        Builtins_CEntry_Return1_DontSaveFPRegs_ArgvOnStack_BuiltinExit [0x00007FF7A15B2F44+68]
        Builtins_InterpreterEntryTrampoline [0x00007FF7A12E159A+954]
        Builtins_InterpreterEntryTrampoline [0x00007FF7A12E159A+954]
        Builtins_ArgumentsAdaptorTrampoline [0x00007FF7A12C5DF9+185]
        Builtins_InterpreterEntryTrampoline [0x00007FF7A12E159A+954]
        Builtins_InterpreterEntryTrampoline [0x00007FF7A12E159A+954]
        (No symbol) [0x000003E6BB7EF04C]
Task trace:
Backtrace:
        electron::NodeBindings::WakeupMainThread [0x00007FF79F178E97+235] (o:\electron\shell\common\node_bindings.cc:427)
```

### To Reproduce

@nornagon and I occasionally hit similar exception recently but so far not able to create isolated repro case can share here. So far this happens PUT + chunkedEncoding.

kwonoj:
#### Description of Change

- closes #21151 .

Context: https://cs.chromium.org/chromium/src/services/network/public/mojom/chunked_data_pipe_getter.mojom?type=cs&sq=package:chromium&g=0&l=30-31

`ChunkedDataPipeGetter::StartReading()` is gauranteed to be called after `ChunkedDataPipeGetter::GetSize()`, but not gauranteed _when_. In result, depends on size of file / timing of stream closes, when `Write(_data, is_last:true)` synchronously calls `StartWriting`, `producer_` is not gauranteed to exist.

This PR changes logic to trigger last part writing,  let DataPipeGetter callback handles writing and emit event necessarily. It required to change some of event emission like close event as well, as DataPipeGetter callback can be called after close. For those case, set state to pending to close and let emit once writing is done.

It's bit unclear how to add test cases for this, as these are intermittent, timing involved.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: fix clientrequest crashes intermittently when uploading

### DIFF
--- a/shell/browser/api/atom_api_url_request.cc
+++ b/shell/browser/api/atom_api_url_request.cc
@@ -65,6 +65,7 @@ enum State {
   STATE_CANCELED = 1 << 2,
   STATE_FAILED = 1 << 3,
   STATE_CLOSED = 1 << 4,
+  STATE_PENDING_CLOSE = 1 << 5,
   STATE_ERROR = STATE_CANCELED | STATE_FAILED | STATE_CLOSED,
 };
 
@@ -259,6 +260,11 @@ void URLRequest::Cancel() {
 }
 
 void URLRequest::Close() {
+  if (!pending_writes_.empty() && !(request_state_ & STATE_CANCELED)) {
+    request_state_ |= STATE_PENDING_CLOSE;
+    return;
+  }
+
   if (!(request_state_ & STATE_CLOSED)) {
     request_state_ |= STATE_CLOSED;
     if (response_state_ & STATE_STARTED) {
@@ -321,10 +327,9 @@ bool URLRequest::Write(v8::Local<v8::Value> data, bool is_last) {
     if (!pending_writes_.empty()) {
       last_chunk_written_ = true;
       StartWriting();
+    } else {
+      EmitFinished();
     }
-
-    request_state_ |= STATE_FINISHED;
-    EmitEvent(EventType::kRequest, true, "finish");
   }
   return true;
 }
@@ -509,8 +514,15 @@ void URLRequest::OnWrite(MojoResult result) {
 
   // Continue the pending writes.
   pending_writes_.pop_front();
-  if (!pending_writes_.empty())
+  if (!pending_writes_.empty()) {
     DoWrite();
+  } else if (last_chunk_written_) {
+    EmitFinished();
+    size_callback_.Reset();
+    if (request_state_ & (STATE_PENDING_CLOSE | STATE_ERROR)) {
+      Close();
+    }
+  }
 }
 
 void URLRequest::DoWrite() {
@@ -524,8 +536,13 @@ void URLRequest::DoWrite() {
 }
 
 void URLRequest::StartWriting() {
-  if (!last_chunk_written_ || size_callback_.is_null())
+  if (!last_chunk_written_ || size_callback_.is_null() || !producer_)
     return;
+
+  if (request_state_ & STATE_CANCELED) {
+    EmitFinished();
+    return;
+  }
 
   size_t size = 0;
   for (const auto& data : pending_writes_)
@@ -552,6 +569,11 @@ void URLRequest::EmitError(EventType type, base::StringPiece message) {
   v8::HandleScope handle_scope(isolate());
   auto error = v8::Exception::Error(gin::StringToV8(isolate(), message));
   EmitEvent(type, false, "error", error);
+}
+
+void URLRequest::EmitFinished() {
+  request_state_ |= STATE_FINISHED;
+  EmitEvent(EventType::kRequest, true, "finish");
 }
 
 template <typename... Args>
--- a/shell/browser/api/atom_api_url_request.cc
+++ b/shell/browser/api/atom_api_url_request.cc
@@ -260,6 +260,8 @@ void URLRequest::Cancel() {
 }
 
 void URLRequest::Close() {
+  // When write called for is_last for upload, last_chunk_written is set already
+  // with pending bytes
   if (!pending_writes_.empty() && last_chunk_written_) {
     request_state_ |= STATE_PENDING_CLOSE;
     return;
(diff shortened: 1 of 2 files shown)


nornagon:
@kwonoj iirc this is 100% reproducible in Slack; is it possible we could recreate those conditions in a test?

shell/browser/api/atom_api_url_request.cc
```diff
@@ -259,6 +260,11 @@ void URLRequest::Cancel() {
 }
 
 void URLRequest::Close() {
+  if (!pending_writes_.empty() && last_chunk_written_) {
```
nornagon: How can the pending writes be non-empty but `last_chunk_written_` is true? Perhaps add a comment here?

kwonoj:
@nornagon it is not 100%. I tried various time while creating this fix and noticed it's frequent but intermittent. Secondly, using same code / same combination also partially triggered issue on bare electron, but it's exactly samr as existing test case (`chunkEncoding` test case)

shell/browser/api/atom_api_url_request.cc
```diff
@@ -524,8 +536,13 @@ void URLRequest::DoWrite() {
 }
 
 void URLRequest::StartWriting() {
-  if (!last_chunk_written_ || size_callback_.is_null())
+  if (!last_chunk_written_ || size_callback_.is_null() || !producer_)
```
nornagon: Do we need to emit some event here in the case of `!producer_`?

shell/browser/api/atom_api_url_request.cc
```diff
@@ -554,6 +571,11 @@ void URLRequest::EmitError(EventType type, base::StringPiece message) {
   EmitEvent(type, false, "error", error);
 }
 
+void URLRequest::EmitFinished() {
+  request_state_ |= STATE_FINISHED;
```
nornagon: can we `DCHECK` that we didn't already emit this event?

nornagon:
also cc @deepak1556 for a second look

shell/browser/api/atom_api_url_request.cc
kwonoj: https://github.com/electron/electron/pull/21157/files#diff-83b26000afb8e61c607f454ace5a06d2R328-R329

We  set flags, and let start writing - until ns calls startreading and completes, we have pending bytes ready but flag is set. I also felt this is bit interesting path.

shell/browser/api/atom_api_url_request.cc
kwonoj: Personally don't think so, `StartWriting` is now eager and either ns or urlrequest calls it, and if we call it producer doesn't exist we'll wait for next ns callback.

shell/browser/api/atom_api_url_request.cc
kwonoj: updated to emit once try to follow https://github.com/electron/electron/pull/21157/files#diff-83b26000afb8e61c607f454ace5a06d2R248-R249, but if DCHECK makes more sense will update again.

MarshallOfSound: @kwonoj If finishing twice is unexpected / invalid we should `DCHECK` on it to ensure we notice in the tests.  If it's an unnaceptable state we should `CHECK` on it.  Otherwise if it's just a state we can ignore then the current code is fine.  Personally I think it should be some form of check

kwonoj: I think trying to finish multiple time in URLRequest itself is fine, as long as consumer gauranteed to have single finish event before close.

nornagon:
How do we feel about landing this vs. backporting https://github.com/electron/electron/pull/21244? On the one hand, this change is a lot smaller than #21244. On the other hand, I am far from convinced that this change doesn't introduce other, more sinister bugs.

deepak1556:
I am more inclined to backport the refactor, as mentioned earlier by @nornagon the current implementation is becoming a more complex state mention.

kwonoj:
I don't have a strong opinion and close out this PR as needed.

MarshallOfSound:
Replaced by #21244

PR #21164, opened 2019-11-18
fix: allow chromium to handle WM_NCCALCSIZE for frameless windows

deepak1556:
#### Description of Change

The event wasn't handled well for frameless window because of how we claim non client view with no system frame drawn https://cs.chromium.org/chromium/src/ui/views/win/hwnd_message_handler.cc?l=1421-1423

We handle it on our end https://github.com/electron/electron/pull/16596 but this has led to some perf regressions when maximizing and restoring windows, however if we provide the client area insets for frameless window, chromium can handle this event very well without us having to dedupe code paths.

Refs  https://github.com/microsoft/vscode/issues/84343, https://github.com/microsoft/vscode/issues/81102

Verified not to regress the issues previously fixed by https://github.com/electron/electron/pull/16596, https://github.com/electron/electron/pull/18321, https://github.com/electron/electron/pull/19883, https://github.com/electron/electron/pull/19886, https://github.com/electron/electron/pull/19928

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: fix flickering when maximizing and restoring frameless windows

### DIFF
--- a/filenames.gni
+++ b/filenames.gni
@@ -358,6 +358,7 @@ filenames = {
     "shell/browser/ui/tray_icon_observer.h",
     "shell/browser/ui/tray_icon_win.cc",
     "shell/browser/ui/views/atom_views_delegate.cc",
+    "shell/browser/ui/views/atom_views_delegate_win.cc",
     "shell/browser/ui/views/atom_views_delegate.h",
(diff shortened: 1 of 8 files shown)


deepak1556:
The failing tests are not related to this PR, merging.

PR #21169, opened 2019-11-18
fix: menu should not be garbage-collected when popuping

### LINKED ISSUE #20737: crash in AtomMenuController.menuDidClose

zcbenz:
#### Description of Change
Fixes #20737

This PR makes sure the Menu object is being retained when popuping, so it won't be garbage-collected until the menu is closed. This is done by keeping a strong reference to the menu when `popup` is called, and release it when the close callback is called.

Most changes in this PR are about changing the callback of `popup` from `RepeatingCallback` to `OnceCallback`, so we can ensure the strong reference is released correctly during compile time.

#### Checklist

- [x] `npm test` passes

#### Release Notes

### DIFF
--- a/shell/browser/api/atom_api_menu.cc
+++ b/shell/browser/api/atom_api_menu.cc
@@ -5,6 +5,7 @@
 #include "shell/browser/api/atom_api_menu.h"
 
 #include <map>
+#include <utility>
 
 #include "shell/browser/native_window.h"
 #include "shell/common/gin_converters/accelerator_converter.h"
@@ -215,6 +216,16 @@ void Menu::OnMenuWillClose() {
 void Menu::OnMenuWillShow() {
   g_menus[weak_map_id()] = v8::Global<v8::Object>(isolate(), GetWrapper());
   Emit("menu-will-show");
+}
+
+base::OnceClosure Menu::BindSelfToClosure(base::OnceClosure callback) {
+  // return ((callback, ref) => { callback() }).bind(null, callback, this)
+  v8::Global<v8::Value> ref(isolate(), GetWrapper());
+  return base::BindOnce(
+      [](base::OnceClosure callback, v8::Global<v8::Value> ref) {
+        std::move(callback).Run();
+      },
+      std::move(callback), std::move(ref));
 }
 
 // static
--- a/spec-main/api-menu-spec.ts
+++ b/spec-main/api-menu-spec.ts
@@ -822,6 +822,31 @@ describe('Menu module', function () {
         menu.closePopup()
       })
     })
+
+    it('prevents menu from getting garbage-collected when popuping', (done) => {
+      const menu = Menu.buildFromTemplate([{ role: 'paste' }])
+      menu.popup({ window: w })
+
+      // Keep a weak reference to the menu.
+      const v8Util = process.electronBinding('v8_util')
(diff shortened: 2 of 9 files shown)


spec-main/api-menu-spec.ts
```diff
@@ -822,6 +822,31 @@ describe('Menu module', function () {
...
       })
     })
+
+    it('prevents menu from getting garbage-collected when popuping', (done) => {
+      const menu = Menu.buildFromTemplate([{ role: 'paste' }])
+      menu.popup({ window: w })
+
+      // Keep a weak reference to the menu.
+      const v8Util = process.electronBinding('v8_util')
+      const map = (v8Util as any).createIDWeakMap() as any
```
miniak: can you add `createIDWeakMap()` to `V8UtilBinding` in `typings/internal-ambient.d.ts` instead?

miniak: ref #21171

shell/browser/api/atom_api_menu.cc
```diff
@@ -217,6 +218,16 @@ void Menu::OnMenuWillShow() {
   Emit("menu-will-show");
 }
 
+base::OnceClosure Menu::BindSelfToClosure(base::OnceClosure callback) {
+  // return ((callback, ref) => { callback() }).bind(null, callback, this)
+  v8::Global<v8::Value> ref(isolate(), GetWrapper());
```
deepak1556: Cool hack 👍

spec-main/api-menu-spec.ts
miniak: let's merge this one first, I will fix this in the other PR

miniak:
rebased to fix the unrelated flaky tests

PR #21184, opened 2019-11-18
build: use python3 to download external binaries

MarshallOfSound:
On master, on macOS the version of `vpython` chrome ships in depot tools does not have a new enough version of openssl to download things from github.  This updates our script to work in both python 2 and python 3 and updates our DEPS file to use python 3 to download the external binaries.

This fixes the issue with nightlies not releasing.

Notes: no-notes

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -114,7 +114,7 @@ hooks = [
     'pattern': 'src/electron/script/update-external-binaries.py',
     'condition': 'download_external_binaries',
     'action': [
-      'python',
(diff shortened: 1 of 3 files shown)


codebytere:
Do we need to backport this to 8?

jkleinsc:
Merging as CI failures unrelated to PR change

nornagon:
FWIW this broke my build on Windows as I don't have python3 installed there.

PR #21190, opened 2019-11-19
feat: expose executeJavaScriptInIsolatedWorld on webContents

MarshallOfSound:
This exposes `executeJavaScriptInIsolatedWorld` in a similar way to `executeJavaScript` so that folks can call it from the main process.  Also cleans up the code we use for the main-process side `executeJavaScript`

Notes: Exposed `executeJavaScriptInIsolatedWorld` on the `webContents` API

### DIFF
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -1033,7 +1033,7 @@ contents.executeJavaScript('fetch("https://jsonplaceholder.typicode.com/users/1"
 
 #### `contents.executeJavaScriptInIsolatedWorld(worldId, scripts[, userGesture])`
 
-* `worldId` Integer - The ID of the world to run the javascript in, `0` is the default world, `999` is the world used by Electrons `contextIsolation` feature.  You can provide any integer here.
+* `worldId` Integer - The ID of the world to run the javascript in, `0` is the default world, `999` is the world used by Electron's `contextIsolation` feature.  You can provide any integer here.
 * `scripts` [WebSource[]](structures/web-source.md)
 * `userGesture` Boolean (optional) - Default is `false`.
 
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -1031,6 +1031,17 @@ contents.executeJavaScript('fetch("https://jsonplaceholder.typicode.com/users/1"
   })
 ```
 
+#### `contents.executeJavaScriptInIsolatedWorld(worldId, scripts[, userGesture])`
+
+* `worldId` Integer - The ID of the world to run the javascript in, `0` is the default world, `999` is the world used by Electrons `contextIsolation` feature.  You can provide any integer here.
+* `scripts` [WebSource[]](structures/web-source.md)
+* `userGesture` Boolean (optional) - Default is `false`.
+
+Returns `Promise<any>` - A promise that resolves with the result of the executed code
+or is rejected if the result of the code is a rejected promise.
+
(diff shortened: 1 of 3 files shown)


docs/api/web-contents.md
```diff
@@ -1031,6 +1031,17 @@ contents.executeJavaScript('fetch("https://jsonplaceholder.typicode.com/users/1"
   })
 ```
 
+#### `contents.executeJavaScriptInIsolatedWorld(worldId, scripts[, userGesture])`
+
+* `worldId` Integer - The ID of the world to run the javascript in, `0` is the default world, `999` is the world used by Electrons `contextIsolation` feature.  You can provide any integer here.
```
loc:
```suggestion
* `worldId` Integer - The ID of the world to run the javascript in, `0` is the default world, `999` is the world used by Electron's `contextIsolation` feature.  You can provide any integer here.
```

lib/browser/api/web-contents.js
```diff
@@ -183,22 +183,25 @@ for (const method of webFrameMethods) {
   }
 }
 
-const executeJavaScript = (contents, code, hasUserGesture) => {
-  return ipcMainUtils.invokeInWebContents(contents, false, 'ELECTRON_INTERNAL_RENDERER_WEB_FRAME_METHOD', 'executeJavaScript', code, hasUserGesture)
+const waitTillCanExecuteJavaScript = async (webContents) => {
```
loc:
```suggestion
const waitTillCanExecuteJavaScript = (webContents) => {
```

spec-main/api-web-contents-spec.ts
```diff
@@ -214,6 +214,25 @@ describe('webContents module', () => {
...
+
+    before(async () => {
+      w = new BrowserWindow({ show: false, webPreferences: { contextIsolation: true } })
+      await w.loadURL('about:blank')
+    })
+
+    it('resolves the returned promise with the result', async () => {
+      await w.webContents.executeJavaScriptInIsolatedWorld(999, [{ code: 'window.X = 123' }])
+      const isolatedResult = await w.webContents.executeJavaScriptInIsolatedWorld(999, [{ code: 'window.X' }])
+      console.log(isolatedResult)
```
loc:
```suggestion
```

spec-main/api-web-contents-spec.ts
```diff
@@ -214,6 +214,25 @@ describe('webContents module', () => {
...
+      const mainWorldResult = await w.webContents.executeJavaScript('window.X')
+      console.log(mainWorldResult)
```
loc:
```suggestion
```

docs/api/web-contents.md
loc: For my own learning: do we have any docs that guide when you'd want to use a world with an ID between 1-998?

MarshallOfSound: Not really, it's just not a world we use.

download (1)

lib/browser/api/web-contents.js
MarshallOfSound: We can't do that because of the next line doing a short-circuit `return`.  Putting `return Promise.resolve()` just seems weird when putting `async` here works fine

docs/api/web-contents.md
loc: Would probably be a little less confusing for consumers to use constants provided by Electron, I would think?

jkleinsc:
Approved for backport to 8-x-y by the releases WG.

PR #21224, opened 2019-11-20
fix: menu should not be garbage-collected when popuping (8-x-y)

zcbenz:
Backport of https://github.com/electron/electron/pull/21169.


jkleinsc:

PR #21225, opened 2019-11-20
fix: menu should not be garbage-collected when popuping (7-1-x)

zcbenz:
Backport of #21169.


momakes:
@zcbenz Thank you so much, but can we please have a release on 7.1.x channel with this? 6 and 8 have been released, only 7 doesn't include this 💜 Would really be awesome if we could have a new patch release.

PR #21236, opened 2019-11-20
feat: introduce new window.open customization API

loc:
#### Description of Change
Depends on #19703 

Commits are atomic — The two "easier review" commits are meant to be ignored, as they're file moves. The `feat: add webContents.setWindowOpenOverride API` commit contains the actual changes.

This is an implementation of the draft spec defined here: https://github.com/electron/governance/pull/208. A few minor things had to change based on ease of implementation: notably the override handler passed to `setWindowOpenOverride` will receive fewer arguments than initially anticipated. This is to avoid patching. I'm unsure of the process — should I go update that document? Is that going to be a source of truth?

Read the draft spec for full rationale, but a quick summary: this adds an API, `webContents.setWindowOpenOverride`, which 1) addresses a fundamental flaw in our current implementation: the user can only cancel a window while it's initializating 2) improves the ergonomics of an API burgeoning in importance.

Code is better than words.

**The old way:**

```javascript
// main.js
mainWindow.webContents.on('new-window', (event, url, frameName, disposition, options, additionalFeatures) => {
  // to cancel:
  event.preventDefault()
  // or...
  // to customize:
  Object.assign(options, {
    frame: false
  })
  event.newGuest = new BrowserWindow(options)
})

// renderer.js
window.open('about:blank')
```
**The new way:**
```javascript
// main.js
mainWindow.webContents.setWindowOpenOverride({ url, frameName }) => {
  // to cancel:
  return false
  // or...
  // to customize:
  return {
    frame: false
  }
})

mainWindow.webContents.on('did-create-window', (win, { url, frameName, options: browserWindowOptions, disposition, additionalFeatures, referrer, postData }) => {
  myApp.subscribeToWindowEvents(win)
})

// renderer.js
window.open('about:blank')
```

**Alignment of code paths**
Another major pillar of this PR is the continuation of the work that @brenca was doing in #19703, where he more closely aligned the `nativeWindowOpen` and `BrowserWindowProxy` code paths. Before his work, it was unclear if and why things were different between them. Now they are just about entirely the same. To that end, I moved the BrowserWindowProxy-specific pieces to their own file.

**Code cleanup**
Finally, the code was ripe for cleanup and relatively hard to understand without reading several entire files. I tried to break out logic into contained functions and document the things I'd learned while puzzling over how the code worked. I think the most important piece of that was understanding how the browser options are created — now the precedence is much more clear.

**Deprecations**
I'm proposing the following deprecations. I'm not really sure the best way to go about deprecations in behavior, so please advise.
- `new-window` event - This event is problematic, as described in detail in the draft spec, and is no longer necessary.
- `additionalFeatures` - AFAICT, this is an Electron creation. Its inclusion also forces our window.open implementation to go against the html spec, as it says bare keys should be interpreted to be value `true`.
- Parent option inheritance - currently, most options are inheritable from the parent if the window.open doesn't specify otherwise. This can be annoying for consumers, as it forces them to find the inverse of parent-specified options when they create the new window and broadly doesn't align with the paradigm laid out in the rest of our API in creating windows. It's much easier for a developer to reuse BrowserWindowConstructorOptions if they want similarly styled windows than it is for a developer to find the inverse of the parent-specified options should they choose to diverge.

Note that these are not removed in this PR as it's designed to be backportable.

**Snapshots**
Because this is rather heavy refactoring of the code underlying `new-window`, I employed snapshotting to prove to myself (and you, the reviewer) I hadn't broken things. I'm not wed to the idea that they actually are merged, but I do think they're generally useful.

There are a couple small changes in the snapshot (which you can review in the commit `test: update new-window event snapshots` that I found to be permissible and in my mind represent "bug fixes" in the API. I recognize there is a blurred line between bug fix and API change so I want to call it out so we don't break applications in a minor bump. The most significant: previously, the `show` option was inherited from the `parent's visibility` (not their `show` property,  which could be different). I think this is incorrect behavior. If I open a window and don't specify show=false, it should be shown. However, I am willing to budge on the inclusion of this change in the backported version.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Introduced a new API for customizing renderer window.open() creation.


spec-main/chromium-spec.ts
```diff
@@ -425,7 +425,7 @@ describe('chromium features', () => {
 
   describe('window.open', () => {
     for (const show of [true, false]) {
-      it(`inherits parent visibility over parent {show=${show}} option`, (done) => {
+      it(`does not inherit parent visibility {show=${show}}`, (done) => {
```
loc: See the PR description for this change — I believe old behavior was wrong.

spec-main/chromium-spec.ts
```diff
@@ -476,37 +476,6 @@ describe('chromium features', () => {
       expect(preferences.javascript).to.be.false()
     })
 
-    it('handles cycles when merging the parent options into the child options', (done) => {
```
loc: I don't think this was a necessary requirement? Happy to be shown that I'm wrong and reinstate the functionality, but I don't see how cycle prevention is helpful.

spec-main/fixtures/snapshots/native-window-open.snapshot.txt
```diff
@@ -2,23 +2,18 @@
   [
     "top=5,left=10,resizable=no",
     {
-      "sender": "[WebContents]",
-      "returnValue": 2
```
loc: This was an internal implementation return value leaking out.

spec-main/fixtures/snapshots/native-window-open.snapshot.txt
```diff
@@ -2,23 +2,18 @@
...
     "frame name",
     "new-window",
     {
       "show": true,
-      "width": 800,
+      "title": "cool",
+      "width": 200,
       "height": 600,
-      "webContents": "[WebContents]",
-      "top": "5",
```
loc: Note the wrong type here, not that it matters for the BrowserWindow constructor (as it will use x, y)

spec-main/fixtures/snapshots/native-window-open.snapshot.txt
```diff
@@ -41,33 +40,32 @@
...
         "sandbox": true,
         "backgroundColor": "blue",
+        "zoomFactor": "2",
         "nodeIntegration": false,
         "webviewTag": false,
         "nodeIntegrationInSubFrames": false,
         "openerId": "placeholder-opener-id"
       },
       "resizable": false,
-      "x": null,
```
loc: Note the bug from incorrect boolean conversion of an integer key.

spec-main/fixtures/snapshots/proxy-window-open.snapshot.txt
```diff
@@ -136,20 +140,21 @@
     "frame name",
     "new-window",
     {
-      "show": "false",
```
loc: Note the wrong type

.eslintrc.json
```diff
@@ -12,7 +12,7 @@
     "@typescript-eslint/no-unused-vars": ["error", {
       "vars": "all",
       "args": "after-used",
-      "ignoreRestSiblings": false
+      "ignoreRestSiblings": true
```
loc: Useful for emulating `_.pick` / `_.omit` via destructuring, as done here:

https://github.com/electron/electron/pull/21236/commits/051706e4a8b620f46b3e7aa1f2096806e3b1653f#diff-39ff6eb84462bf6b40517f852555ac6dR262

miniak:
@loc can you rebase please?

chevonc:
Hi @loc our team is very interested in this new API and wondering if this PR is still actively being worked on?

loc:
@chevonc depends on what you mean by active, but yes, I still intend to merge this. Biggest roadblock right now is that I initially didn't understand that webPreferences are necessarily tied to the process, and thus you won't be able to change them in sandbox mode (nativeWindowOpen) even with this shiny new API. Sooo I need to figure out how to surface that constraint in the API.

chevonc:
Thanks for replying back @loc. Glad to know this is still being actively looked at 🎉. You made an intersting point: `I initially didn't understand that webPreferences are necessarily tied to the process, and thus you won't be able to change them in sandbox mode`. 

Follow up question here, related to a use-case I was hoping this shiny api would unlock - If we cannot modify `webPreferences` in sandbox mode, does this mean *all* preferences will always be inherited from the parent window? In particular, the `offscreen` preference.  What we were looking for is the ability to call `window.open` from a visible onscreen window and setting `webPreferences: {offscreen: true}}` **before** the native `webContents` is created. The issue we're seeing is when doing `const windowRef = window.open(...)`, in the current api, the [initial `windowRef` becomes invalid ](https://github.com/electron/electron/blob/master/lib/browser/guest-window-manager.js#L260) after I create a new offscreen BrowserWindow inside the `new-window` event. This renders `windowRef` useless for manipulating the DOM of the newly created offscreen window

loc:
@chevonc yep, this is exactly the sort of use case I'd love to enable. As it sits now, this PR will not enable your use case because webPreferences are sent via the command line when the process gets forked (the parent webContents gets created). In the future, I'd like to change that so that we can modify webPreferences over time for child windows via IPC, but no promises on the timeline for that.

Thanks for detailing your use case. It helps me know we're on the right track!

chevonc:
@loc Yup definitely on the right track. You mentioned:
 
```
As it sits now, this PR will not enable your use case because webPreferences are sent via the command line... In the future, I'd like to change that so that we can modify webPreferences over time for child windows via IPC, but no promises on the timeline for that.
```

I have a few days of bandwidth to help take a look here, if you'll have it. Could you link me to 
1) Where the process fork happens 
2) Where you imagine getting an IPC event to swap out the `webContents`. 

I'd like to get some context on the code change we'd need here - so far I've successfully checked out and built the repo on MacOSX 😅

loc:
Closing in favor of https://github.com/electron/electron/pull/24517

PR #21243, opened 2019-11-21
chore: improve nativeImage path converter error

codebytere:
#### Description of Change

In fixing the Tray-from-path issue earlier, I realized that we don't throw a very intuitive error if there are conversion failures originating from files actually not existing on disk. This PR thus adds a check for the `base::FilePath` not actually being a valid path and throws a more user-friendly error should that be the case.

Before:
Screen Shot 2019-11-21 at 9 43 39 AM

After:
Screen Shot 2019-11-22 at 8 37 02 AM

cc @ckerr 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/common/api/atom_api_native_image.cc
+++ b/shell/common/api/atom_api_native_image.cc
@@ -12,6 +12,7 @@
 #include "base/files/file_util.h"
 #include "base/strings/pattern.h"
 #include "base/strings/string_util.h"
+#include "base/strings/utf_string_conversions.h"
 #include "base/threading/thread_restrictions.h"
 #include "net/base/data_url.h"
 #include "shell/common/asar/asar_util.h"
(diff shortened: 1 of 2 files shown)


codebytere:
Windows path error:
```
not ok 946 tray module new Tray throws a descriptive error for a missing file
  AssertionError: expected [Function] to throw error matching /C:\projects\src\I\Do\Not\Exist does not exist on disk/ but got 'C:\\projects\\src\\I\\Do\\Not\\Exist does not exist on disk'
      at Context.<anonymous> (electron\spec-main\api-tray-spec.ts:22:18)
```
I'll figure out how best to account for Windows fs paths and then update this accordingly.

shell/common/api/atom_api_native_image.cc
```diff
@@ -541,8 +542,21 @@ bool Converter<electron::api::NativeImage*>::FromV8(
   // Try converting from file path.
   base::FilePath path;
   if (ConvertFromV8(isolate, val, &path)) {
+    // Throw when path is invalid, i.e file doesn't exist
+    base::ThreadRestrictions::ScopedAllowIO allow_io;
```
deepak1556: We should avoid operations on the main thread that involve adding `ScopedAllowIO` where possible.

Also it seems redundant to check for path to throw the bellow error as `!(*out)->image().IsEmpty()` already gives you whether an image could be initialized or not. Please use that check to throw the error.

codebytere: The issue there, though, is that it doesn't tell you _why_ the image couldn't be initialized, does it? My goal here is to make it clear to users when the fault lies with their path and not with a more oblique and unclear failure.

deepak1556: Well whatever error you require can be obtained from `electron::api::NativeImage::CreateFromPath` since it has all the knowledge about the image and its path when the initialization happened, you can try to retrieve the info from there.

codebytere: Updated!

codebytere:
Failure on arm is:
```
not ok 1259 document.visibilityState should become visible when a window is restored
  AssertionError: expected 'hidden' to equal 'visible'
      at Context.<anonymous> (electron\spec-main\visibility-state-spec.ts:88:54)
      at Context.<anonymous> (electron\spec-main\visibility-state-spec.ts:21:13)
```
Merging as this is unrelated.

Issue #21200, opened 2019-11-19
request.write does not add a body in second negotiate request


Spenhouet:
### Issue Details

* **Electron Version:** 7.1.2.
* **Operating System:** Windows 10 (1803)

### Expected Behavior

`request.write` adds a request body to the request.

### Actual Behavior

The request does not contain a body.

### Additional Information

We are using SSO via Kerberos.
The first POST contains the body, then comes the 401 Negotiate response and then the second POST request does not contain the body.

I'm using this simple method for request:

```javascript
const requestPromise = new Promise((resolve, reject) => {
        const request = net.request({url, method});

        if (headers) {
            for (let key in headers) {
                request.setHeader(key, headers[key]);
            }
        }

        if (body && Object.entries(body).length !== 0 && body.constructor === Object) {
            request.write(JSON.stringify(body));
        }

        request.on('response', (response) => {
            let responseDataString = '';

            response.on('data', (chunk) => {
                responseDataString += chunk;
            });

            response.on('end', () => {
                let responseData = JSON.parse(responseDataString);
                if (response.statusCode === 200) {
                    resolve({statusCode: response.statusCode, data: responseData});
                } else {
                    reject({errorMessage: `${response.statusCode} - ${responseData.errorMessage}`});
                }
            });
        });

        request.on('error', (error) => {
            reject({errorMessage: JSON.stringify(error)});
        });

        request.end();
    });
```

I did see that `net` is totally broken with v7.1.x. - so is this also broken?
If so, is there any workaround that I can apply?

Spenhouet:
Further details:

We are using SSO via Kerberos.
The first POST contains the body, then comes the 401 Negotiate response and then **the second POST request does not contain the body**.

vladimiry:
I'd suggest trying recently released 7.1.2 as it got #20923 fixed. It's not the same issue but I think it's worth trying.

Spenhouet:
@vladimiry Thank you for the suggestion. I just tested 7.1.2 but it didn't solve this issue.

nornagon:
Working on cleaning things up with #21244, but as a workaround you might try just re-issuing the request after the credentials have been provided. Credentials should be stored in the auth cache, and sent along with subsequent requests, so a retry ought to work.

nornagon:
I tried adding a test for this in #21244, where it works. However, on master, this test crashes the main process:

```patch
diff --git a/spec-main/api-net-spec.ts b/spec-main/api-net-spec.ts
[...]

PR #21244, opened 2019-11-21
refactor: rewrite the net module to simplify state tracking

nornagon:
#### Description of Change
This is a major rewrite of the `net` module. The main goals here are:

1. Reduce complexity of state management. The existing `URLRequest` class has a lot of delicate logic around managing the lifetime of a request, which [has subtle bugs](https://github.com/electron/electron/pull/21157). This refactor aims to substantially remove that state tracking through careful choice of interface.
2. Move stream logic into JS. Working with JS streams in C++ is challenging. This moves all stream logic into JavaScript, simplifying the C++ code and replacing it with smaller, more modular and easier-to-understand JS.
3. Make `ClientRequest` a real `stream.Readable`.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed some crashes that could occur when using the `net` module.


nornagon:
TODO:

- [ ] implement redirect logic
- [ ] clean up 'abort' flow
- [ ] write tests for upload body (POST/PUT) and redirect / auth. See #21200

Additionally, I'd like to expose some new features on this module, though that will come in a followup. Specifically:

- `ResourceRequestBody` has support for directly uploading a file from the FS without going through JS. I'd like to expose that as `net.request({ bodyFromFile: 'path/to/file' })`.
- This PR maintains the behavior that streaming data into the request buffers until the complete upload data is in memory before starting the request. This is because the full size of the stream must be known in advance. I'd like to introduce a change that will use proper streaming if the `content-length` header is set in advance.

shell/browser/api/atom_api_url_loader.cc
```diff
@@ -0,0 +1,396 @@
...
+    bytes_written_ += buffer->ByteLength();
+    auto backing_store = buffer->Buffer()->GetBackingStore();
+    data_producer_->Write(
+        std::make_unique<BufferDataSource>(base::make_span(
+            static_cast<char*>(backing_store->Data()) + buffer->ByteOffset(),
+            buffer->ByteLength())),
+        base::BindOnce(
+            &JSChunkedDataPipeGetter::OnWriteChunkComplete,
+            base::Unretained(
+                this),  // TODO: should this be a weak ptr? what do we do with
```
zcbenz: We can probably add an option to `gin_helper::Promise`, to make it throw in the C++ destructor if it has not been resolved or thrown before. So if `JSChunkedDataPipeGetter` goes away in the middle of writing, the returned JS promise would throw, other than pending forever.

It should probably even be the default behavior.

nornagon: Ah, yes, that would make sense!

nornagon:
NB. this PR changes the way redirects are handled with `net.request`. Previously, there was an option to `net.request` called `redirect`, which could be `"follow"`, `"error"` or `"manual"`. We used that to fill the `redirect_mode` field of `ResourceRequest`.

However, that field is intended to be used for the `fetch` parameter of the same name (see here: https://fetch.spec.whatwg.org/#concept-request-redirect-mode) and doesn't work the way it was documented for `net.request`. Instead, I've replaced this with an API somewhat similar to the `"login"` event, where the default behavior if no listeners are registered is to follow redirects, but if a listener is registered then it's up to that listener to continue or cancel the request.

nornagon:
Removing [wip] because this is passing the tests now :)

nornagon:
... actually, readding it because there are still unaddressed TODOs in the code, and I want to pull out the test changes into a separate PR.

shell/browser/api/atom_api_url_loader.cc
nornagon: Actually... I'm not so sure this is a good idea. It would mean entering JS at a possibly unexpected time, when it's not necessarily safe to begin executing JS. Probably best that it remains explicit.

nornagon:
okie doke, tests are separated into https://github.com/electron/electron/pull/21265 which should be reviewed first. all TODOs removed. let's go 🚀

nornagon:
If this is gonna be backportable (per https://github.com/electron/electron/pull/21157#issuecomment-558339073), I'll have to patch in the old redirect behavior.

nornagon:
I've returned the pre-existing `'redirect'` behavior, though the docs were previously incorrect about when `followRedirect()` could be called. It's unclear to me if that was a change that happened with the network service, or was just always wrong. In any case, I believe this to now be a non-API-changing PR with respect to 7-1-x and 8-x-y and thus backportable.

nornagon:
Superseded by https://github.com/electron/electron/pull/21294.

nornagon:
good luck trop

nornagon:
```
          .--. .-,       .-..-.__
        .'(`.-` \_.-'-./`  |\_( "\__
     __.>\ ';  _;---,._|   / __/`'--)
    /.--.  : |/' _.--.<|  /  | |
_..-'    `\     /' /`  /_/ _/_/
 >_.-``-. `Y  /' _;---.`|/))))
'` .-''. \|:  .'   __, .-'"`
 .'--._ `-:  \/:  /'  '.\             _|_
     /.'`\ :;   /'      `-           `-|-`
    -`    |     |                      |
          :.; : |                  .-'~^~`-.
          |:    |                .' _     _ `.
          |:.   |                | |_) | |_) |
          :. :  |                | | \ | |   |
          : ;   |                |           |
          : ;   |                |   trop    |
          : ;   |                |           |
          : ;   |                | One patch |
          : ;   |                |  conflict |
        .jgs. : ;                | too many  |
-."-/\\\/:::.    `\."-._'."-"_\\-|           |///."-
" -."-.\\"-."//.-".`-."_\\-.".-\\`=.........=`//-".
```

Issue #21159, opened 2019-11-17
checkbox in dialog not working 


otsybulsky:
### Issue Details

* **Electron Version:**
  * 7.1.1
* **Operating System:**
  *  Windows 10 (1803) 
* **Last Known Working Electron version:**
  * 6.1.3

### Expected Behavior
dialog.showMessageBox
https://electronjs.org/docs/api/dialog#dialogshowmessageboxbrowserwindow-options
Returns Promise<Object> - resolves with a promise containing the following properties:
- response Number - The index of the clicked button.
- **checkboxChecked** Boolean - The checked state of the checkbox if checkboxLabel was set. Otherwise false.

### Actual Behavior
checkboxChecked do not change

sofianguy:
Thank you for taking the time to report this issue and helping to make Electron better.

Would it be possible for you to fork [electron-quick-start](https://github.com/electron/electron-quick-start) for a small app that reproduces the issue by itself?

Standalone test cases make fixing issues go more smoothly: it ensure everyone's looking at the same issue, it removes all unnecessary variables from the equation, and it can also provide the basis for automated regression tests.

xland:
@sofianguy 

Yes. This is a bug.
I test it on my computer.


```js
let { remote } = require("electron");
let result = await remote.dialog.showMessageBox({
  checkboxLabel: "test",
  checkboxChecked: false,
  title: "test",
  message: "test,test,test"
});
console.log(result.checkboxChecked);
```

Whether you check the CheckBox or not, It will all print `false`

codebytere:
@otsybulsky thanks for the report, i just put up a PR that should fix the issue!

xland:
@codebytere 

^_^

otsybulsky:
Thank you very much for your help!

jrdiard:
How do I get this fix? My app exhibits the checkbox not working using 7.1.7. I've tried Electron Fiddle with 6.1.5, 7.1.4, 7.1.7, and V8.0.0.-beta.5 and 6.1.5 is the only one that works.  Note this is on Windows 10. I have not had any issues on Mac OS.   Here is the fiddle test (this is main.js):

/ Display native system dialogs for opening and saving files, alerting, etc.
//
// For more info, see:
// https://electronjs.org/docs/api/dialog

const { app, BrowserWindow, dialog } = require('electron')

let mainWindow = null

app.on('ready', () => {
  mainWindow = new BrowserWindow({ height: 600, width: 600 })

[...]

PR #21248, opened 2019-11-21
fix: correctly plumb checkboxChecked on win

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/21159.

We were passing the `MessageBoxSettings` struct with `const`  qualifier, so we weren't properly updating the user's change of the `checkboxChecked` value in the resolved Promise. This resolves that issues and also adds an extra error check so that we don't silently no-op when users pass `checkboxChecked` but no `checkboxLabel`.

cc @ckerr 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed an issue where the `checkboxChecked` value wasn't being updated in `showMessageBox` on Windows.

### DIFF
--- a/shell/browser/ui/message_box.h
+++ b/shell/browser/ui/message_box.h
@@ -6,6 +6,7 @@
 #define SHELL_BROWSER_UI_MESSAGE_BOX_H_
 
 #include <string>
+#include <utility>
 #include <vector>
 
 #include "base/callback_forward.h"
@@ -28,6 +29,8 @@ enum MessageBoxOptions {
   MESSAGE_BOX_NONE = 0,
   MESSAGE_BOX_NO_LINK = 1 << 0,
 };
+
+using DialogResult = std::pair<int, bool>;
 
 struct MessageBoxSettings {
   electron::NativeWindow* parent_window = nullptr;
--- a/shell/browser/ui/message_box_win.cc
+++ b/shell/browser/ui/message_box_win.cc
@@ -171,13 +171,10 @@ DialogResult ShowTaskDialogUTF16(NativeWindow* parent,
   }
 
   int button_id;
-  bool cb_checked = false;
 
   int id = 0;
   BOOL verificationFlagChecked = FALSE;
   TaskDialogIndirect(&config, &id, nullptr, &verificationFlagChecked);
-  if (checkbox_checked)
-    cb_checked = verificationFlagChecked;
 
   if (id_map.find(id) != id_map.end())  // common button.
     button_id = id_map[id];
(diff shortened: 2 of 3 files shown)


codebytere:
Thanks @deepak1556! that's a cool solution - i've updated it to that end.

ckerr:
At the risk of bikeshedding, I'm not a fan of this alternative. IMO it's more readable / understandable / maintainable to have the function return the entire return state (button id, checkbox bool) than to return half of it in the return value and the other half inside a const struct whose name includes the word "Settings". If I were coming into this code cold, a Settings struct is not where I'd be looking for a return state.

codebytere:
That's fair - i'm happy with both solutions but i do agree with you @ckerr that the latter is cleaner and more understandable for someone not familiar with this aspect of the codebase.

deepak1556:
I can agree with @ckerr concern, in that case `checkbox_checked` shouldn't be a part of `MessageBoxSettings` in the original solution since its configurable and is ending up being a return state.

shell/browser/ui/message_box.h
```diff
@@ -6,6 +6,7 @@
 #define SHELL_BROWSER_UI_MESSAGE_BOX_H_
 
 #include <string>
+#include <utility
```
ckerr: tyop

shell/browser/ui/message_box_win.cc
```diff
@@ -148,10 +148,8 @@ int ShowTaskDialogUTF16(NativeWindow* parent,
 
   if (!checkbox_label.empty()) {
     config.pszVerificationText = checkbox_label.c_str();
-
-    if (checkbox_checked && *checkbox_checked) {
+    if (checkbox_checked && *checkbox_checked)
```
ckerr: Since `checkbox_checked` is a bool, this line can be just `if (checkbox_checked)`

shell/browser/ui/message_box_win.cc
```diff
@@ -172,22 +170,26 @@ int ShowTaskDialogUTF16(NativeWindow* parent,
       config.dwFlags |= TDF_USE_COMMAND_LINKS;  // custom buttons as links.
   }
 
+  int button_id;
+  bool cb_checked = false;
```
deepak1556: Minor style nit: can eliminate this temporary state by doing,

`return std::make_pair(button_id, checkbox_checked ? verificationFlagChecked : false);`

codebytere:
Failing on gtk icon issue - this is a windows bug so it is unrelated.

PR #21266, opened 2019-11-22
feat: add session.addWordToSpellCheckerDictionary to allow custom words in the dictionary

MarshallOfSound:
This feature adds the missing capability of the "Add To Dictionary" context menu item for a mis-spelled word.

Notes: Added `session.addWordToSpellCheckerDictionary` API to support custom words in the 
dictionary

### DIFF
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -491,7 +491,7 @@ to host here.
 
 Returns `Boolean` - Whether the word was successfully written to the custom dictionary.
 
-**Note:** On macOS this word will be written to the OS custom dictionary as well
+**Note:** On macOS and Windows 10 this word will be written to the OS custom dictionary as well
 
 ### Instance Properties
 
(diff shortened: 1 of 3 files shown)


nornagon:
Why is this on `session`? Are dictionaries specific to a BrowserContext? It looks like they're not when using Chrome's spellchecker...

MarshallOfSound:
>  Are dictionaries specific to a BrowserContext? It looks like they're not when using Chrome's spellchecker...

Yes they are.  https://github.com/electron/electron/pull/21266/files#diff-52cca0c5435ff413826d402cfdc8c0e4R717-R722

Each `SpellcheckService` is per-context.  Same reason the other spell check methods are here as well

docs/api/session.md
```diff
@@ -479,6 +483,16 @@ behavior you can use this API to point the dictionary downloader at your own hos
...
 
+**Note:** On macOS the OS spellchecker is used and therefore we do not download any dictionary files.  This API is a no-op on macOS.
+
+#### `ses.addWordToSpellCheckerDictionary(word)`
+
+* `word` String - The word you want to add to the dictionary
+
+Returns `Boolean` - Whether the word was successfully written to the custom dictionary.
+
+**Note:** On macOS this word will be written to the OS custom dictionary as well
```
nornagon: Also Win10?

PR #18676, opened 2019-06-06
ci: generate debug symbols on Linux

CapOM:
#### Description of Change

For `mac` and `win` there is currently `electron-v4.2.3-darwin-x64-dsym.zip` and `electron-v4.2.3-win32-x64-pdb.zip` of about 500MB. There is no such debug package for `linux`. This PR just fixes that.

Electron uses breakpad to generate minidumps but there are cases where it does not work properly on Linux so better to use usual coredumps and proper debug symbols on Linux.

https://github.com/electron/electron/issues/13315

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Generate debug symbols on Linux

PR #21278, opened 2019-11-25
ci: generate debug symbols on Linux

nornagon:
Backport of #18676, see that PR for details.

Notes: Generate debug symbols on Linux.


zcbenz:
Not sure why CI is failing.

Issue #18397, opened 2019-05-22
[Discussion] Requiring Native Modules in the Renderer Process to be NAPI or Context Aware


MarshallOfSound:
### Native modules in Electron

Loading native modules in Electron has always been tricky.  It's a pain to rebuild, to link to the right `node.lib/dll`, and to ensure your `NODE_MODULE_VERSION` is correct.  And there's another problem behind the scenes that most users have never had to deal with: Node doesn't let you load multiple instances of a native module in the same process, even in different v8 contexts.  Imagine this scenario:

1. Load https://google.com in BrowserWindow
    * Process is launched.
    * Preload script runs and loads native module X
2. Load https://google.com/foo in BrowserWindow
    * Chromium decides to reuse the process
    * Preload script runs and loads native module X
    * FAILURE - native module X loaded twice in the same process

We address this in Electron by patching Chromium to create a new process for **every** navigation.  This mostly works but still has a few edge cases (e.g. #4025, #12045, #17576) and a few side effects:
* We lose all the performance gains from the Chromium plz-navigate changes :cry: 
* We have strange behavior with `window.open` and `window.opener` as we force child windows into separate processes sometimes. :cry: 
* Our patches occasionally have regressions because this isn't a process model that Chromium accepts / tests. :cry: 

We've always thought "wouldn't it be cool if we could change this". Now, with Node 12's [Worker Threads](https://nodejs.org/api/worker_threads.html), we see a way to do that.

### Worker Threads

Node ran into the same issue when they added worker threads: native modules could not be loaded in multiple workers.  Node solved this by introducing the concept of ["Context Aware"](https://nodejs.org/api/addons.html#addons_context_aware_addons) native modules.  This is how native modules tell Node that they are safe to be loaded in multiple `v8::Context`s.  [Nan](https://github.com/nodejs/nan) has a handy helper ([`NAN_MODULE_WORKER_ENABLED`](https://github.com/nodejs/nan/pull/792)) for doing this.

Context Aware modules and NAPI modules can be instantated multiple times in Node even without our patched process model. As they become more common, our patches will bring less benefit and be more redundant.

### What does this mean for Electron?

Right now, nothing. However, at some point in the future -- probably in a few major versions -- we intend to remove our process model patches. At that point, any native Node modules loaded in the renderer process **must** be either NAPI or Context Aware.

[...]

PR #21287, opened 2019-11-26
chore: add deprecation warning for the default of allowRendererProcessReuse

MarshallOfSound:
Refs: https://github.com/electron/electron/issues/18397

This is the `8-x-y` step that is required in that issue's timeline.

Notes: Deprecated the default value of `app.allowRendererProcessReuse` (See #18397)

### DIFF
--- a/lib/browser/api/web-contents.js
+++ b/lib/browser/api/web-contents.js
@@ -322,7 +322,7 @@ let warnedAboutRendererProcessReuse = false
 
 // Add JavaScript wrappers for WebContents class.
 WebContents.prototype._init = function () {
-  if (!warnedAboutRendererProcessReuse && app._allowRendererProcessReuseIsDefaultValue) {
+  if (!process.noDeprecation && !warnedAboutRendererProcessReuse && app._allowRendererProcessReuseIsDefaultValue) {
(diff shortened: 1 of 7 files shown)


lib/browser/api/web-contents.js
```diff
@@ -318,8 +318,14 @@ const addReturnValueToEvent = (event) => {
   })
 }
 
+let warnedAboutRendererProcessReuse = false
+
 // Add JavaScript wrappers for WebContents class.
 WebContents.prototype._init = function () {
+  if (!warnedAboutRendererProcessReuse && app._allowRendererProcessReuseIsDefaultValue) {
+    deprecate.log('The default value of app.allowRendererProcessReuse is deprecated, it is currently "false".  It will change to be "true" in Electron 9.  For more information please check https://github.com/electron/electron/issues/18397')
```
nornagon: want to use `deprecate.warnOnce` or similar? that pays attention to `process.noDeprecation`

MarshallOfSound: `warnOnce` logs that a thing will be removed, I can respect `process.noDeprecation` here though

PR #8537, opened 2017-01-30
webContents: manage zoom changes on the browser side with HostZoomMap

deepak1556:
With this PR:
- Using zoom api on webContents will persist in a session on per-origin basis (132.content.foo and content.foo will share zoom changes if they are on the same session).
- Webview will still inherit embedder zoom changes. (this behavior is already present and it will continue)
- The zoom changes of each host in a session are persisted with the Preference file, so they will be available on the next load of the app. (If ZoomFactor is set on the webPreference then it will take precedence over the values from preference file, this can be overriden by using the zoom apis). In short the order of precedence is Preference File < ZoomFactor on WebPreference < Zoom api on webContents.
- Use the webframe api if the zoom changes are required only to be temporary. (Does not persist across navigations)

Would like feedback if the above behavior sounds good, thanks!

- [ ] Add Spec
- [x] Map webframe zoom calls to HostZoomMap temporary apis.

Depends on https://github.com/electron/brightray/pull/273

Fixes https://github.com/electron/electron/issues/6958

PR #8041, opened 2016-11-21
Add new method to set layout-based zoom level limit

paulcbetts:
The current behavior of `setZoomLevelLimits` affects the visual-based zoom (i.e. no layout changes) despite all of the rest of the methods affecting layout-based zoom. While this zoom survives in-page navigations such as iframes being loaded, it also causes pages to be cut-off.

The ideal fix for this would to be to just change the behavior of this method but that would be a breaking change. This PR introduces a new zoom level limit method to limit layout-based zoom.

This allows you to work around #6958 by doing the following:

```js
function setZoomHarder(wv, zoomLevel) {
  wv.setLayoutZoomLevelLimits(zoomLevel, zoomLevel);
  wv.setZoomLevel(zoomLevel);
}
```

Refs #6958

Issue #6958, opened 2016-08-24
WebView zoomFactor reset on navigate


Thomas101:
- Electron version: 1.3.4
- Operating system: osx, Ubuntu 16.04 LTS

It looks like the zoomFactor is reset on webview tags when the page navigates. You can replicate this by adding the following to a webview preload script...

``` js
const webFrame = require('electron').webFrame
webFrame.setZoomFactor(2)
setTimeout(function () {
    window.location.hash = '123'
}, 1000)
```

 There's also some discussion around this issue [here](https://discuss.atom.io/t/why-does-webview-reset-its-zoomfactor-on-url-navigate/27064).

Would you consider this a bug or the desired behaviour? If it is the desired behaviour is there anyway to suppress it or configure it to be more permanent? I can imagine there are plenty of instances where you would want to provide a consistent zoom level to some remote content.

Thanks :)

zcbenz:
It is expected behavior when using `webFrame.setZoomFactor`, since it only changes current frame and current frame is gone after navigation.

The correct way is to use `webview.setZoomFactor`, which is supposed to keep the zoom factor through navigations. However when I just tried it I found the zoom factor still goes away after navigation, so this is a bug.

MarshallOfSound:
@zcbenz I think this is the issue that I am experiencing (but never tracked down) over at MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/issues/1421

This issue didn't always exist and appears to have been introduced some time between `1.1.0` and `1.3.2`.  I'll try track it down to a specific version.

Thomas101:
@zcbenz Didn't know you could set zoomFactor directly on the web view, is this just missing from the docs or up and coming?

@MarshalOfSound as a workaround I found you can use setZoomLevelLimits(1.5, 1.5). It's not great, but keeps it set

zcbenz:
> @zcbenz Didn't know you could set zoomFactor directly on the web view, is this just missing from the docs or up and coming?

I think it is just missing in docs.

Thomas101:
@zcbenz I really hate to ask cause you're already doing such great work, but can you share any info on when a fix for this might land?

zcbenz:
@Thomas101 Usually a bug is only fixed when someone is interested in it, so if no one assigns themselves to the issue, you can assume one one is working on it.

Thomas101:
Thanks for the update on this :)

andymartinwork:
[...]

PR #21296, opened 2019-11-26
fix: deprecate setLayoutZoomLevelLimits

nornagon:
#### Description of Change
This hasn't worked [since Chromium removed it](https://chromium-review.googlesource.com/c/chromium/src/+/1809500) and I believe it's beyond our capacity to continue to support it. The function doesn't work, but instead of throwing an error it seems nicer to warn about it in 8.x and wait till 9.x to delete it completely.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: `setLayoutZoomLevelLimits` no longer works in Electron 8, so emit a warning message when using it. The function will be removed in Electron 9.


deepak1556:
This was added in https://github.com/electron/electron/pull/8041 ,  the issue it mentioned was fixed in another way https://github.com/electron/electron/pull/8537, can you please verify if there is a regression before merging this.

nornagon:
@deepak1556 do you mean this issue? https://github.com/electron/electron/issues/6958

deepak1556:
Yup that one, users seem to have been using this api as a workaround to persist the zoom level across navigations but we later fixed it on electron core, just want to make sure its still fixed, so that we can remove this api safely.

nornagon:
https://gist.github.com/8ace446c40111aca67391104c2a9cc76 I tested this in a Fiddle and it seems to work OK. However, I'm not totally sure what the original failing behavior was so I can't say for certain that I didn't accidentally dodge it somehow. Wish there had been a test written :(

PR #21286, opened 2019-11-25
fix: restore --ignore-connections-limit functionality

nornagon:
#### Description of Change
This restores the [inadvertently removed](https://github.com/electron/electron/pull/19644/files#diff-8c1eed0d78f8872226e8dd910fccb551L245) `--ignore-connections-limit` functionality.

Fixes #21255.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Fixed `--ignore-connections-limit`, ironically, being ignored.

PR #21298, opened 2019-11-26
fix: restore --ignore-connections-limit functionality

nornagon:
Backport of #21286, see that PR for details.

Notes: Fixed `--ignore-connections-limit`, ironically, being ignored.


zcbenz:
The failing tests are not related to this PR.

PR #21315, opened 2019-11-27
feat: expose sync js exec methods on WebFrame

### LINKED ISSUE #19161: promisifying WebFrame.executeJavaScript(InIsolatedWorld) is a breaking change without a workaround
### Preflight Checklist

### Issue Details

1. not an improvement

### Expected Behavior

### Actual Behavior

### To Reproduce
    **`index.js`**
    ```js
    'use strict';
    
          path = require('path');
    
    let win;
    
    app.on('ready', () => {
    
      win = new BrowserWindow({
        width: 1440,
        height: 900,
        useContentSize: true,
        webPreferences: {
          nodeIntegration: false,
          contextIsolation: true,
        },
      });
    
      win.loadURL('about:blank');
    
    });
    ```
    **`preload.js`**
    ```js
    'use strict';
    
      let result;
      return result;
    };
    ```
5. it prints `1`
7. it prints `undefined`

bughit:
#### Description of Change

Before the promisification (#17312) of `WebFrame#executeJavaScript*` their interface was async (callback) but behavior was sync (callback called before the methods returned). So they could be used in sync code (or async, you could wrap them in a promise yourself). The switch from a callback to a promise interface made it impossible to use them in sync mode (#19161), which was a breaking change without a workaround.

This PR exposes the sync js exec methods on WebFrame, providing a workaround for code that used the original methods synchronously. In general, a sync in-process js eval makes perfect sense, because it's sync under the hood. 

Fixes #19161

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: add sync javascript execution methods to WebFrame

### DIFF
--- a/docs/api/web-frame.md
+++ b/docs/api/web-frame.md
@@ -154,6 +154,23 @@ or is rejected if the result of the code is a rejected promise.
 
 Works like `executeJavaScript` but evaluates `scripts` in an isolated context.
 
+### `webFrame.executeJavaScriptSync(code)`
+
+* `code` String
+
+Returns `any` - The result of the executed code.
+
+Evaluates `code` in page.
+
+### `webFrame.executeJavaScriptInIsolatedWorldSync(worldId, code)`
+
+* `worldId` Integer - The ID of the world to run the javascript in, `0` is the default world, `999` is the world used by Electrons `contextIsolation` feature.  You can provide any integer here.
+* `code` String
+
+Returns `any` - The result of the executed code.
+
+Works like `executeJavaScriptSync` but evaluates `scripts` in an isolated context.
+
 ### `webFrame.setIsolatedWorldInfo(worldId, info)`
 * `worldId` Integer - The ID of the world to run the javascript in, `0` is the default world, `999` is the world used by Electrons `contextIsolation` feature. Chrome extensions reserve the range of IDs in `[1 << 20, 1 << 29)`. You can provide any integer here.
 * `info` Object
--- a/shell/renderer/api/atom_api_web_frame.cc
+++ b/shell/renderer/api/atom_api_web_frame.cc
@@ -436,6 +436,24 @@ v8::Local<v8::Promise> ExecuteJavaScriptInIsolatedWorld(
   return handle;
 }
 
+v8::Local<v8::Value> ExecuteJavaScriptSync(gin_helper::Arguments* args,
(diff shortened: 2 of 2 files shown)


bughit:
It appears some commit hook is converting all .md filed to crlf. I am checking it as lf and have no conversion enabled.

bughit:
`RequestExecuteScriptAndReturnValue` used by the [current method](https://github.com/electron/electron/blob/master/shell/renderer/api/atom_api_web_frame.cc#L387) is also [deprecated](https://cs.chromium.org/chromium/src/third_party/blink/public/web/web_local_frame.h?l=347). So it did not seem wrong to use a deprecated method.

Is it also going to be removed? Then what's the right method to use? The deprecation comment says "Prefer requestExecuteScriptInIsolatedWorld", assuming they mean `RequestExecuteScriptInIsolatedWorld`, it does not work for the main world: "worldID must be > 0 (as 0 represents the main world)."

MarshallOfSound:
@bughit As our current implementation returns a promise we can transparently switch to using `RequestExecuteScriptInIsolatedWorld` internally without any API changes.

There is the current limitation on Chromium's side that that API does not allow you to pass in the main world ID, but that seems easily fix-able and we will probably do that.

We can't support a synchronous version here with Chromium's intention to remove all synchronous versions of those APIs.

bughit:
> Chromium's intention to remove all synchronous versions of those APIs.

@MarshallOfSound not sync execution

https://cs.chromium.org/chromium/src/third_party/blink/public/web/web_local_frame.h?l=366

The chrome APIs have a async interface but they can be executed synchronously and that's not going away. Please reopen this, I can at least try adding sync versions using the exiting API's. Please understand that for *years* these could be used synchronously without any issue, consistently, with no exceptions, despite having callback interface, and still can be.

bughit:
@MarshallOfSound It is definitely possible to add a sync method based on RequestExecuteScript, that returns the value delivered by the sync callback and throws if the callback was not called (it will quietly delete itself when called later). The only reason it would ever not be called sync  is if the FrameLifecycleState is not kRunning, 

https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/execution_context/execution_context.cc?l=158

something I have never seen, I am guessing it never just happens spontaneously, but regardless, that possibility would be covered.

Do object to even that? Why?

Issue #21288, opened 2019-11-26
Mircotaskscope in the renderer interferes with debugging async/await


deepak1556:
**Electron Version:** 6-1-x, 7-1-x, 8-xy

### Expected Behavior

Devtools should perform normally without crash

### Actual Behavior

Renderer crashes with following,

https://gist.github.com/deepak1556/29458392e72d01c097b822cbb36848f9 -> interesting frame starts from 33

### To Reproduce

https://gist.github.com/deepak1556/f9ce7fd4d695361f1568c787affa3795

This originally surfaced in https://github.com/microsoft/vscode/issues/83994 as a different issue, this is just one example based on the task event loop was processing when we originally saw the crash but we have explicit Microtaskscope spread out in a couple of places in the renderer and depending on the task queue at the time of debugger pause, they will cause a similar issue.

/cc @zcbenz

zcbenz:
The `blink::PromiseRejectHandlerInMainThread` is called when promise is rejected, and crash happened inside it because blink could not could a valid DOM window at current scope.

From the crash report, it seems that we need to make `electron::EmitIPCEvent` call JS via blink.

electron-triage:
The Electron version reported on this issue is no longer supported. See our [supported versions documentation](https://www.electronjs.org/docs/tutorial/support#supported-versions).

If this is still reproducible on a supported version, please open a new issue with any other new information that a maintainer should know.

Thank you for taking the time to report this issue and helping to make Electron better! Your help is appreciated.

deepak1556:
Can confirm the issue doesn't repro with Electron >=11

PR #21351, opened 2019-12-02
fix: execute JS function via blink in renderer process

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/21288.

When IPC message is received there is no blink window bound to current scope, and Electron would crash if `blink::ToLocalDOMWindow` is called during JS executions.

By using `blink::WebLocalFrame::RequestExecuteV8Function` to execute JS functions, we can correctly setup a blink environment.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix crash when debugging async/await code.

### DIFF
--- a/shell/renderer/electron_api_service_impl.cc
+++ b/shell/renderer/electron_api_service_impl.cc
@@ -44,7 +44,8 @@ v8::Local<v8::Object> GetIpcObject(v8::Local<v8::Context> context) {
   return value->ToObject(context).ToLocalChecked();
 }
 
-void InvokeIpcCallback(v8::Local<v8::Context> context,
+void InvokeIpcCallback(blink::WebLocalFrame* frame,
+                       v8::Local<v8::Context> context,
                        const std::string& callback_name,
                        std::vector<v8::Local<v8::Value>> args) {
   TRACE_EVENT0("devtools.timeline", "FunctionCall");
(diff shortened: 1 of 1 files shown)


shell/renderer/electron_api_service_impl.cc
```diff
@@ -54,24 +55,19 @@ void InvokeIpcCallback(v8::Local<v8::Context> context,
   if (ipcNative.IsEmpty())
     return;
 
-  // Only set up the node::CallbackScope if there's a node environment.
-  // Sandboxed renderers don't have a node environment.
-  node::Environment* env = node::Environment::GetCurrent(context);
-  std::unique_ptr<node::CallbackScope> callback_scope;
```
MarshallOfSound: I think we still need this when there is a node environment?

shell/renderer/electron_api_service_impl.cc
```diff
@@ -54,24 +55,19 @@ void InvokeIpcCallback(v8::Local<v8::Context> context,
...
-
   auto callback_key = gin::ConvertToV8(isolate, callback_name)
                           ->ToString(context)
                           .ToLocalChecked();
   auto callback_value = ipcNative->Get(context, callback_key).ToLocalChecked();
   DCHECK(callback_value->IsFunction());  // set by init.ts
   auto callback = callback_value.As<v8::Function>();
-  ignore_result(callback->Call(context, ipcNative, args.size(), args.data()));
+
+  frame->RequestExecuteV8Function(context, callback, ipcNative, args.size(),
```
MarshallOfSound: This isn't guaranteed to be synchronous anymore correct? I don't think that's a problem just pointing it out.

shell/renderer/electron_api_service_impl.cc
nornagon: I believe the node callback scope is responsible for invoking async hooks, which are totally broken in Electron currently, but shouldn't be 🙃

deepak1556:
@zcbenz the problem seems to be somewhere with task runners being shared between chromium and node, here is a different code path.

In the test case from the issue replace the html file with following contents.

```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8">
    <title>Hello World!</title>
  </head>
  <body>
    <h1>Hello World!</h1>
  </body>
  <script>
	class Foo {
		constructor() {
			document.body.addEventListener('click', () => {
				this.foo();
			});
		}
		async foo() {
			require('fs').promises.readdir('electron-gn/src').then((err, files) => {
        console.log(files.length);
      });
			await this.foo2(); // set breakpoint in this line
			await this.foo2();
		}
		async foo2() {
			return Promise.resolve();
		}
	}
	new Foo();
  </script>
</html>
```


This doesn't go through the `EmitIPCEvent` code path on purpose, rather the promise resolution from node fs , check from frame 30 https://gist.github.com/deepak1556/2e60fb4c89134026fa27aaf845aabfb5 will trigger a microtask checkpoint and cause a similar crash.

shell/renderer/electron_api_service_impl.cc
deepak1556: The problem is not with async hooks, we can bypass those checks with `export NODE_OPTIONS="--no-force-async-hooks-checks"`, although not a ideal flow.

Here the problem is about `node::CallbackScope` triggering a microtask checkpoint at an undesired time.

zcbenz:
I'll try fixing it in another way.

PR #21356, opened 2019-12-02
chore: prevent user code polluting internal EventEmitter's

MarshallOfSound:
Please note I'm well aware this is an infinite and impossible mountain, but we can at least make the other side of the mountain harder for everyone else 😄 

Notes: no-notes


lib/common/event-emitter.ts
```diff
@@ -0,0 +1,13 @@
...
+ * not propogate to Electron's internal instances.  This means that user
+ * code can't pollute Electron logic.
+ */
+import { EventEmitter as NodeEventEmitter } from 'events'
+
+export const EventEmitter = function EventEmitter (this: NodeEventEmitter) {
+  NodeEventEmitter.call(this)
+} as any as typeof NodeEventEmitter
+
+Object.assign(EventEmitter.prototype, NodeEventEmitter.prototype)
```
nornagon: Isn't this still easily modifiable from user code by:

```js
require('electron').BrowserWindow.prototype.emit = ...
```

MarshallOfSound: It's been a while but I think I was trying to avoid userland modules that mutate `EventEmitter` breaking internal electron logic.

Some module I was using was extending `EventEmitter` in a way that broke ipc.

lib/common/event-emitter.ts
```diff
@@ -0,0 +1,13 @@
+/**
```
zcbenz: I think we still prefer `//` for multiple lines comment?

lib/common/event-emitter.ts
zcbenz: Can you add comment for the line in the file?

zcbenz:
@nornagon Have your question been resolved in this PR?

lib/common/event-emitter.ts
nornagon: How does Node handle this with their primordials..?

zcbenz: @nornagon  Node does nothing to prevent users to change internal states, they just use `EventEmitter` and say do not modify things:
https://github.com/nodejs/node/blob/master/lib/net.js#L35
https://nodejs.org/api/http.html#http_agent_requests

Should we just keep things as they are since if users mess up with `EventEmitter` they would break Node anyway?

nornagon: @zcbenz if Node would also be broken by this, then it seems reasonable to me that we should be broken by it too. I'm happy to leave things as-is then.

zcbenz:
I'm closing this PR since as discussed above if users pollute internal EventEmitter they would break Node.js too.

PR #21371, opened 2019-12-03
feat: expose sync js exec interface on WebFrame (try 2)

### LINKED ISSUE #19161: promisifying WebFrame.executeJavaScript(InIsolatedWorld) is a breaking change without a workaround

bughit:
#### Description of Change

Before the promisification (#17312) of `WebFrame#executeJavaScript*` their interface was async (callback) but behavior was sync (callback called before the methods returned). So they could be used in sync code (or async, you could wrap them in a promise yourself).

The switch from a callback to a promise interface made it impossible to use them in sync mode (#19161), which was a breaking change without a workaround.

This PR exposes the sync js exec methods on WebFrame, providing a workaround for code that used the original methods synchronously. In general, a sync in-process js eval makes perfect sense, because it's sync under the hood.

This attempt uses the same underlying methods `WebLocalFrame::RequestExecuteScript*` as the "async" versions.

Fixes #19161

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: add sync javascript execution methods to WebFrame

### DIFF
--- a/shell/renderer/api/atom_api_web_frame.cc
+++ b/shell/renderer/api/atom_api_web_frame.cc
@@ -140,6 +140,47 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
   DISALLOW_COPY_AND_ASSIGN(ScriptExecutionCallback);
 };
 
+class ScriptSyncExecutionCallback : public blink::WebScriptExecutionCallback {
+ public:
+  bool completed_ = false;
+  bool delete_self_ = false;
+  v8::Global<v8::Value> result_;
+
+  std::string error_message_;
+
+  explicit ScriptSyncExecutionCallback(v8::Isolate* isolate)
+      : isolate_(isolate) {}
+
+  ~ScriptSyncExecutionCallback() override = default;
+
+  void Completed(
+      const blink::WebVector<v8::Local<v8::Value>>& result) override {
+    if (delete_self_) {
+      delete this;
+      return;
+    }
+    completed_ = true;
+    if (!result.empty()) {
+      if (!result[0].IsEmpty()) {
+        // Right now only single results per frame is supported.
+        result_ = v8::Global<v8::Value>(isolate_, result[0]);
+      } else {
+        error_message_ =
+            "Script failed to execute, this normally means an error "
+            "was thrown. Check the renderer console for the error.";
+      }
+    } else {
+      error_message_ =
+          "WebFrame was removed before script could run. This normally means "
+          "the underlying frame was destroyed";
+    }
+  }
+
+ private:
+  v8::Isolate* isolate_;
+  DISALLOW_COPY_AND_ASSIGN(ScriptSyncExecutionCallback);
+};
+
 class FrameSetSpellChecker : public content::RenderFrameVisitor {
  public:
   FrameSetSpellChecker(SpellCheckClient* spell_check_client,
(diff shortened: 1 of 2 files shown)


shell/renderer/api/atom_api_web_frame.cc
```diff
@@ -436,6 +477,102 @@ v8::Local<v8::Promise> ExecuteJavaScriptInIsolatedWorld(
...
+    } else {
+      args->ThrowError(error_message);
+      return v8::Null(isolate);
+    }
+  } else {
+    sync_callback->delete_self_ = true;
+
+    // this can only happen if the frame is not in the
+    // mojom::FrameLifecycleState::kRunning state
+    args->ThrowError("execution did not complete synchronously");
```
MarshallOfSound: This makes no sense as an API, their is no way the app can handle this in any useful way.

bughit: I have never seen the top frame not in the mojom::FrameLifecycleState::kRunning state.

Can you point out when it wouldn't be?

When a frame is in the kRunning state, execution is guaranteed to be synchronous, that's not an "implementation detail" it's a public api guarantee

```c++
  enum ScriptExecutionType {
    // Execute script synchronously, unless the page is suspended.
    kSynchronous,
    // Execute script asynchronously.
    kAsynchronous,
    // Execute script asynchronously, blocking the window.onload event.
    kAsynchronousBlockingOnload
  };
```

Only if the page is suspended, would execution not be synchronous.

If the top frame being suspended is an unavoidable, unpredictable possibility outside of one's control, which seems unlikely, I would agree with you. Otherwise these sync methods are useful and will work with 100% reliability and so should not be rejected.

shell/renderer/api/atom_api_web_frame.cc
```diff
@@ -436,6 +477,102 @@ v8::Local<v8::Promise> ExecuteJavaScriptInIsolatedWorld(
...
+  v8::Isolate* isolate = args->isolate();
+  bool has_user_gesture = false;
+  args->GetNext(&has_user_gesture);
+
+  // return v8::Null(isolate);
+
+  ScriptSyncExecutionCallback* sync_callback =
+      new ScriptSyncExecutionCallback(isolate);
+
+  GetRenderFrame(window)->GetWebFrame()->RequestExecuteScriptAndReturnValue(
```
nornagon: This method on `WebFrame` is deprecated: https://cs.chromium.org/chromium/src/third_party/blink/public/web/web_local_frame.h?l=347&rcl=879cf27d2d56f2ff63e2ba7041bfabdd2d96deb3

shell/renderer/api/atom_api_web_frame.cc
nornagon: I believe this could happen if the window is displaying an `alert` or similar: https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/page/page.h?l=237-244&rcl=879cf27d2d56f2ff63e2ba7041bfabdd2d96deb3

shell/renderer/api/atom_api_web_frame.cc
bughit: This is what the async versions are using as well, there's currently no alternative as the suggested requestExecuteScriptInIsolatedWorld does not work for the main world

nornagon: Oh? Even if you pass `0`, which is the ID for the main world?

shell/renderer/api/atom_api_web_frame.cc
bughit: I'll test it. If so, and if this is something that could be detected, the sync methods could fail early. That would not be unreasonable behavior.

shell/renderer/api/atom_api_web_frame.cc
nornagon: Ah, I see, the API checks that this is > 0. OK, it seems like there's no non-deprecated way to do this; presumably one will be added before this one is removed. Hopefully 😅

shell/renderer/api/atom_api_web_frame.cc
nornagon: I'd also test what happens if the executed script calls `alert`; that seems like it would pause the script also and prevent a synchronous return from happening.

MarshallOfSound:
@bughit I've come up with a middle ground that should suit your use case.  Implement this API to support **both** callbacks and promises.  Then if you _really_ want to do this, you can implement the "fake sync" behavior in JS by using the callback.

nornagon:
> @bughit I've come up with a middle ground that should suit your use case. Implement this API to support **both** callbacks and promises. Then if you _really_ want to do this, you can implement the "fake sync" behavior in JS by using the callback.

Do you mean to support the old behavior where you could pass in a callback? Seems reasonable to me, and as a bonus old code that depends on the synchronous behavior would continue to work unchanged.

bughit:
@MarshallOfSound That works for me, I am already using the "fake sync" approach in v4.

Please give more details of how this should work.

- should I modify existing methods?
- to take an optional callback?
- that gets invoked if present in addition to the promise?

MarshallOfSound:
Yep exactly that.  The current API should work like it currently does (return a promise).  But **if** a callback is also provided it will be called in addition to resolving the promise

bughit:
@MarshallOfSound 

could use some help with callback conversion

bughit/electron@948d2a38870fd931e4c8f7f5d966aa975878a6a5

callback signature
```c++
  // for compatibility with the older version of this, error is after result
  using CompletionCallback =
      base::Callback<void(const v8::Local<v8::Value>& result,
                          const v8::Local<v8::Value>& error)>;
```

which results in a conversion error

```
../..\gin/converter.h(284,24): error: no member named 'FromV8' in 'gin::Converter<base::RepeatingCallback<void (const v8::Local<v8::Value> &, const v8::Local<v8
::Value> &)>, void>'
  return Converter<T>::FromV8(isolate, input, result);
         ~~~~~~~~~~~~~~^
```

bughit:
needed "shell/common/gin_converters/callback_converter.h"

bughit:
closing in favor of #21423

PR #21382, opened 2019-12-04
docs: cookie.set api documentation and sample code

uallmight:
#### Description of Change

notes: Update sample code to set cookie
notes: Update the API description documentation for `cookie.set` to show `name` and `expirationDate` are required

#### Checklist

- [x] PR description included

#### Release Notes

Notes:


zcbenz:
After looking into Chromium's implementation, I found that even the `name` parameter should not be required. I'll fix this bug in another PR.

I'm closing this PR since it is a bug instead of documentation error, but thanks for bringing this up anyway!

PR #21384, opened 2019-12-04
test: clean up some stale code in renderer spec runner

nornagon:
#### Description of Change
Just stuff that wasn't used any more.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


zcbenz:
I think the crash is caused by the removal of `protocol.registerSchemesAsPrivileged` call.

nornagon:
Hm... I'm not sure I understand why removing that call would cause a crash in `legacy_cache_storage_manager.cc`. Can you explain your reasoning? Restoring the call doesn't seem like the right fix if that's the case.

zcbenz:
Certain code in Chromium used to assert whether the origin of URL has privileges, and if we did not call `protocol.registerSchemesAsPrivileged` for the custom protocols, some operations would crash Chromium. That is why the `protocol.registerSchemesAsPrivileged` call was added.

The crash in Windows CI looks very similar to what we saw:
```
[6088:1210/193848.572:FATAL:legacy_cache_storage_manager.cc(170)] Check failed: !origin.GetURL().is_empty(). 
```

But it might be unrelated though.

nornagon:
The renderer specs no longer use any custom schemes, though.

nornagon:
Did some more investigation, it looks like the crash happens when reading info from the service worker cache. On my machine, the crash happens after reading this file:

```
/Users/nornagon/Library/Application Support/Electron Test/Service Worker/CacheStorage/33300f42568342e80d69cb8f0f8781d2c70179ad/index.txt
```

which contains a protobuf, and the string `app://fake-host/`. That URL is used in many of our other tests, so probably one of those tests is creating this index file and filling it with data that the renderer tests crash on.

Clearing out the user-data directory for the test app fixes the issue.

PR #21396, opened 2019-12-05
fix: Fix compositor recycling when creating new BrowserView (6-1-x)

poiru:
Backport of #21372. This also backports a tiny bit of #19988.

Notes: Fix flicker when switching between BrowserViews after creating new BrowserView


poiru:
@codebytere Seems like trop picked the second #PR from the description instead of the first #PR because of a greedy match at:

https://github.com/electron/trop/blob/ad44de04e948f8272103861cc95f6bd45d5ccb67/src/index.ts#L101

I think this should fix it:

```
const backportPattern = /(?:^|\n)(?:manual |manually )?backport.*?(?:#(\d+)|\/pull\/(\d+))/im;
```

Could someone manually fix up the labels?

PR #19813, opened 2019-08-18
fix: only override default window.close() in windows opened with nativeWindowOpen: false

miniak:
#### Description of Change
Fixes #19194

#### Checklist
- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes
Notes: `window.close()` does no longer close the window when called in `BrowserView`.

Issue #19194, opened 2019-07-10
use window.close in Browserview-webcontents(web page) will close the app!!!!!!!


spcBackToLife:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

Electron Version: 5.0.6
Platform: macOs 10.14

Expected Behavior: 
    when use `window.close()` in web page of Browserview, which page with `onbeforunload` hook. and it will fire `webContents.on('will-prevent-unload', xxx)`, so that i can open a tips dialog of leaving page.

Actual Behavior:
   The Electron app is closed.

I have saw the source code. and it had rewrite the function `window.close` just like this:

```

export const windowSetup = (
  guestInstanceId: number, openerId: number, isHiddenPage: boolean, usesNativeWindowOpen: boolean
) => {
  if (guestInstanceId == null) {
    // Override default window.close.
    window.close = function () {
      ipcRendererInternal.sendSync('ELECTRON_BROWSER_WINDOW_CLOSE')
    }
  }
......
}
```

and in rpc-server.js ,just like this:

```
ipcMainInternal.on('ELECTRON_BROWSER_WINDOW_CLOSE', function (event) {
  const window = event.sender.getOwnerBrowserWindow()
  if (window) {
    window.close()
  }
  event.returnValue = null
})

```

so if like this, how i can close the web page in Browserview, and can get the thing i want?

pushkin-:
>how i can close the web page in Browserview, and can get the thing i want?

I don't understand. From what you said earlier, it sounds like you *don't* want the window to close. What is the "thing you want?"

I think closing the window is the right thing to do here - that's how browsers behave. You can alter the behavior by latching onto the `close` event of BrowserWindow and doing whatever you want there in response to the close.

PalmerAL:
I think the issue is that if you have a BrowserWindow with a BrowserView attached to it, calling `window.close` from inside the view will close the parent window, which isn't what I would expect to happen. Since the `close` event on BrowserWindow doesn't provide a way to tell whether the window itself or just the view is closing, there's no way to prevent this behavior without also making it impossible to close the window.

[...]

PR #21416, opened 2019-12-06
fix: load window-setup in sandboxed renderer

miniak:
#### Description of Change
Fixes the following issues in sandboxed renderers:
- `window.opener` is `null` when opened with `nativeWindowOpen: false`
- `window.prompt()` returns `null` instead of throwing an exception
- `document.visibilityState` is broken in sandboxed `<webview>`

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Fixed an issue that could prevent communication between a sandboxed child window opened with `nativeWindowOpen: false` and an unsandboxed parent window. Also fixed `document.visibilityState` not working in sandboxed `<webview>`.

### DIFF
--- a/filenames.auto.gni
+++ b/filenames.auto.gni
@@ -165,6 +165,7 @@ auto_filenames = {
     "lib/renderer/web-view/web-view-element.ts",
     "lib/renderer/web-view/web-view-impl.ts",
     "lib/renderer/web-view/web-view-init.ts",
+    "lib/renderer/window-setup.ts",
     "lib/sandboxed_renderer/api/exports/electron.ts",
     "lib/sandboxed_renderer/api/module-list.ts",
     "lib/sandboxed_renderer/init.js",
--- a/lib/browser/guest-window-manager.js
+++ b/lib/browser/guest-window-manager.js
@@ -75,8 +75,10 @@ const mergeBrowserWindowOptions = function (embedder, options) {
     }
   }
 
-  // Sets correct openerId here to give correct options to 'new-window' event handler
-  options.webPreferences.openerId = embedder.id
+  if (!webPreferences.nativeWindowOpen) {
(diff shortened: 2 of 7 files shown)


MarshallOfSound:
> We should be relying on the native stuff for this, no?

Yes, we shouldn't (and can't in a context isolated world) be overriding `window.history.back` and friends.  We need to either use the original Chromium impl or hook into `window.history.back` from the native side of things.

miniak:
@nornagon I am explicitly not overriding the history APIs

miniak:
> Can we avoid overriding as much as possible when `nativeWindowOpen: true`? Specifically, `window.close` and `window.prompt` seem to be newly overridden.

`window.prompt` has nothing to do with `nativeWindowOpen`, it should behave consistently regardless of the `sandbox` option. Making it throw as it's not implemented in Electron seems like the better option.
`window.close` has some issues, as reported by #19194 for example. I would suggest unifying the behavior first and then fixing it in a follow up PR.

> We shouldn't enable any of the `*_GUEST_*` functionality if webview tag is disabled. Can we flag it behind that web preference?

which exact code do you mean?

> The visibility propagation stuff is gross but at least it only comes into play if you're using .

agree

miniak:
I've did some experiments, and it seems like we need a custom `window.close()` implementation, otherwise these tests are failing https://github.com/electron/electron/blob/9c442e0a7e1ee5945c4d1f9f67a814c855b7e440/spec-main/api-browser-window-spec.ts#L153-L176 as you can see in https://github.com/electron/electron/pull/19813

nornagon:
Also @miniak I updated the release notes to not mention internal implementation details but instead to try to describe it from what a developer might observe. Please edit further if needed.

miniak:
@nornagon tweaked the release notes as suggested, should be ready to merge now.

PR #21423, opened 2019-12-06
feat: bring back optional callbacks for WebFrame#executeJavaScript* methods

### LINKED ISSUE #19161: promisifying WebFrame.executeJavaScript(InIsolatedWorld) is a breaking change without a workaround

bughit:
#### Description of Change

Before the promisification (#17312) of `WebFrame#executeJavaScript*` their interface was async (callback) but behavior (with some exceptions, e.g. page being suspended due to a modal alert) was sync (callback called before the methods returned). So they could be used in sync mode.

This PR brings back optional callbacks, which are invoked in addition to promise resolution.

Fixes #19161

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Allow an optional callback parameter for WebFrame.executeJavaScript* methods, which is called synchronously unless the target context is paused.

### DIFF
--- a/docs/api/web-frame.md
+++ b/docs/api/web-frame.md
@@ -122,13 +122,20 @@ by its key, which is returned from `webFrame.insertCSS(css)`.
 
 Inserts `text` to the focused element.
 
-### `webFrame.executeJavaScript(code[, userGesture])`
+### `webFrame.executeJavaScript(code[, userGesture, callback])`
 
 * `code` String
 * `userGesture` Boolean (optional) - Default is `false`.
+* `callback` Function (optional) - Called after script has been executed. Unless
+  the frame is suspended (e.g. showing a modal alert), execution will be
+  synchronous and the callback will be invoked before the method returns. For
+  compatibility with an older version of this method, the error parameter is
+  second.
+  * `result` Any
+  * `error` Error
 
-Returns `Promise<any>` - A promise that resolves with the result of the executed code
-or is rejected if the result of the code is a rejected promise.
+Returns `Promise<any>` - A promise that resolves with the result of the executed
+code or is rejected if execution throws or results in a rejected promise.
 
 Evaluates `code` in page.
 
@@ -136,14 +143,24 @@ In the browser window some HTML APIs like `requestFullScreen` can only be
 invoked by a gesture from the user. Setting `userGesture` to `true` will remove
 this limitation.
 
-### `webFrame.executeJavaScriptInIsolatedWorld(worldId, scripts[, userGesture])`
+### `webFrame.executeJavaScriptInIsolatedWorld(worldId, scripts[, userGesture, callback])`
 
-* `worldId` Integer - The ID of the world to run the javascript in, `0` is the default world, `999` is the world used by Electrons `contextIsolation` feature.  You can provide any integer here.
+* `worldId` Integer - The ID of the world to run the javascript
+            in, `0` is the default main world (where content runs), `999` is the
+            world used by Electron's `contextIsolation` feature. Accepts values
+            in the range 1..536870911.
 * `scripts` [WebSource[]](structures/web-source.md)
 * `userGesture` Boolean (optional) - Default is `false`.
+* `callback` Function (optional) - Called after script has been executed. Unless
+  the frame is suspended (e.g. showing a modal alert), execution will be
+  synchronous and the callback will be invoked before the method returns.  For
+  compatibility with an older version of this method, the error parameter is
+  second.
+  * `result` Any
+  * `error` Error
 
-Returns `Promise<any>` - A promise that resolves with the result of the executed code
-or is rejected if the result of the code is a rejected promise.
+Returns `Promise<any>` - A promise that resolves with the result of the executed
+code or is rejected if execution throws or results in a rejected promise.
 
 Works like `executeJavaScript` but evaluates `scripts` in an isolated context.
 
--- a/shell/renderer/api/electron_api_web_frame.cc
+++ b/shell/renderer/api/electron_api_web_frame.cc
@@ -15,6 +15,7 @@
 #include "services/service_manager/public/cpp/interface_provider.h"
 #include "shell/common/api/api.mojom.h"
 #include "shell/common/gin_converters/blink_converter.h"
+#include "shell/common/gin_converters/callback_converter.h"
 #include "shell/common/gin_helper/dictionary.h"
 #include "shell/common/gin_helper/promise.h"
 #include "shell/common/node_includes.h"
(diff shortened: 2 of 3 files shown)


bughit:
@MarshallOfSound @nornagon 

New PR.

I mentioned before, I think there is a commit hook that's converting .md line endings to crlf. I am editing the file as lf.

MarshallOfSound:
> I am editing the file as lf.

@bughit revert the line endings change and then use `git commit --ammend --no-verify` and it should bypass the commit hook

bughit:
> @bughit revert the line endings change and then use `git commit --ammend --no-verify` and it should bypass the commit hook

done

bughit:
@MarshallOfSound Is there anything I need to do for this?

shell/renderer/api/atom_api_web_frame.cc
```diff
@@ -110,32 +111,57 @@ class RenderFrameStatus final : public content::RenderFrameObserver {
 
 class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
  public:
+  // for compatibility with the older version of this, error is after result
```
nornagon: I think this is worth mentioning in the docs, too, as it's contrary to node style.

shell/renderer/api/atom_api_web_frame.cc
```diff
@@ -110,32 +111,57 @@ class RenderFrameStatus final : public content::RenderFrameObserver {
...
+  using CompletionCallback =
+      base::Callback<void(const v8::Local<v8::Value>& result,
```
nornagon:
```suggestion
      base::OnceCallback<void(const v8::Local<v8::Value>& result,
```

zcbenz:
I don't think it is a good idea to change the behavior of API depending on whether a parameter is passed, it would make the API hard to understand and easy to misuse.

It is probably easier to add a new API instead of changing the current one.

bughit:
> I don't think it is a good idea to change the behavior of API depending on whether a parameter is passed

The optional callback does not change the core behavior of the api, nothing about how js is executed changes. There are or have been other APIs that take optional callbacks. I don't think you can describe them as "changing their behavior", they do the same core work, and just call back optionally.

> it would make the API hard to understand

Here's the documentation for the optional callback:

> `callback` Function (optional) - Called after script has been executed. Unless
  the frame is suspended (e.g. showing a modal alert), execution will be
  synchronous and the callback will be invoked before the method returns.

Everything important about it described, it's function, it's optional, it's called after script execution, synchronously (with the modal dialog exception) Why do you say that's hard to understand?

> and easy to misuse.

How?

> It is probably easier to add a new API instead of changing the current one.

I've already tried (#21315, #21371). This PR is a compromise that I understood was acceptable, @nornagon just asked for a test which I haven't had a chance to add. Is that no longer the case? Then what would be acceptable?

zcbenz:
Thanks for the clarification, I'm good with this PR then.

shell/renderer/api/atom_api_web_frame.cc
bughit: this required using std::move which I'm not confident about, so please check those.

bughit:
> Mostly LGTM. A few tweaks suggested, and needs tests. Please test at least:
> 
>     * that both the promise and the callback are called, in both success and error cases
> 

@nornagon What can be done about the failing test for the error case? It seems the synthetic error is being caught by the test framework, which alters the normal flow of the call.
```
not ok 48 webFrame module executeJavaScript() yields errors via a promise and a sync callback
  ReferenceError: thisShouldProduceAnError is not defined
      at <anonymous>:1:1
      at WebFrame../lib/renderer/api/web-frame.ts.WebFrame.<computed> [as executeJavaScript] (electron/js2c/renderer_init.js:1513:33)
      at Context.<anonymous> (/home/builduser/project/src/electron/spec/api-web-frame-spec.js:55:8)
ok 49 webFrame module executeJavaScript() yields errors via a promise and a sync callback
not ok 50 webFrame module executeJavaScript() yields errors via a promise and a sync callback
  TypeError: Cannot read property 'isPending' of undefined
```

nornagon:
@bughit not sure how to answer that as the test you're referring to doesn't seem to exist in the code..?

bughit:
> not sure how to answer that as the test you're referring to doesn't seem to exist in the code..?

I removed the test to get everything else checked. I haven't been able to run tests locally due to first #22343, then #22400. Here it is:

```js
  it('executeJavaScript() yields errors via a promise and a sync callback', done => {
    let callbackResult, callbackError

    webFrame
      .executeJavaScript('thisShouldProduceAnError()', (result, error) => {
        callbackResult = result
        callbackError = error
      })
      .then(
        promiseResult => {
          done(new Error('error is expected'))
        },
        promiseError => {
          expect(promiseError).to.be.an('error')
          done()
        }
      )

    expect(callbackResult).to.be.undefined()
    expect(callbackError).to.be.an('error')
  })
```

I manually tested it of course, it's fine, but the test framework appears to treat the synthetic eval error as a test failure.

bughit:
@nornagon So can you accept this without the error test? Keep in mind that there were no executeJavascript tests in api-web-frame-spec before, and now there are some.

If not, then please help with #22432 so I can experiment with tests locally and/or suggest how the error test I posted can be fixed.

nornagon:
> ```js
>   it('executeJavaScript() yields errors via a promise and a sync callback', done => {
>     let callbackResult, callbackError
> 
>     webFrame
>       .executeJavaScript('thisShouldProduceAnError()', (result, error) => {
>         callbackResult = result
>         callbackError = error
>       })
>       .then(
>         promiseResult => {
>           done(new Error('error is expected'))
>         },
>         promiseError => {
>           expect(promiseError).to.be.an('error')
>           done()
>         }
>       )
> 
>     expect(callbackResult).to.be.undefined()
>     expect(callbackError).to.be.an('error')
>   })
> ```

Looks like you're not waiting for the promise to be completed before calling `expect(callbackError).to.be.an('error')`. Make the test function `async` and then `await` the `executeJavaScript` call perhaps?

bughit:
> Looks like you're not waiting for the promise to be completed before calling expect(callbackError).to.be.an('error')

The callback is synchronous, so callbackError is available right after the call. In fact this assertion is what tests the syncness. The same thing is done in the working test. So this is not the problem.

Did you see the the test runner output (`ReferenceError: thisShouldProduceAnError is not defined`)? It's referring to the the synthetic error which I am deliberately generating to test the error scenario. It's treating this error as a test failure.

nornagon:
Ah, I see the issue. The test is executing the error-producing function in the frame that's running the tests itself, and mocha is catching "uncaught" errors.

Here's a way around that by using `webFrame` to execute in a child frame instead: https://gist.github.com/nornagon/033e8497e16dae0e4216652cb3bf83ae

bughit:
Adding a test for `executeJavascriptInIsolatedWorld` revealed a problem in the original error detection logic added by @MarshallOfSound in 0fc172fcaf8ad0158d89d5447be39983fd5614b5

You can confirm it with `try { await webFrame.executeJavaScriptInIsolatedWorld(999, [{code: 'thisShouldProduceAnError()'}]); console.log('this should not be printed');} catch {}` in 9.0

Fixing this is not really in the scope of this PR, as the problem applies equally to the callback and promise, but if it's quick, and since the test is already here, perhaps it makes to include it?

How would you like to proceed?

bughit:
in this case `result[0].IsEmpty()` is false

nornagon:
If the bug is already present in existing code, I think it's fine to leave that to a followup PR.

bughit:
ok, I commented out (with an explanation) the executeJavaScriptInIsolatedWorld error test. I believe everything else you requested is there.

bughit:
executeJavascriptInIsolatedWorld rejection issue: #22490

`linux-ia32-testing-tests ` failure does not appear to be related to this PR

bughit:
Can this be backported to 9?

bughit:
@nornagon My understanding is that you allow backporting to the beta branch, is that not so? And have an automated way to do that, assuming it merges cleanly. Can that be done here?

PR #21431, opened 2019-12-08
fix: load window-setup in sandboxed renderer

miniak:
Backport of #21416

See that PR for details.

Notes: Fixed an issue that could prevent communication between a sandboxed child window opened with `nativeWindowOpen: false` and an unsandboxed parent window. Also fixed `document.visibilityState` not working in sandboxed `<webview>`.


miniak:
@codebytere the PR to master is merged now

PR #21434, opened 2019-12-09
fix: window menu should handle keys correctly

zcbenz:
#### Description of Change

Fix a few weird behaviors and crashes when controlling window menu by keys, for example:
* Press Alt => Press Esc => Crash
* Press Alt => Press Alt again => Focus does not move back

This was caused by Chromium removing a few virtual methods. I have added the `virtual`s back, and I'll try to upstream the patch.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix a weird behaviors and crashes when controlling window menu by keys.


patches/chromium/accessible_pane_view.patch
```diff
@@ -0,0 +1,30 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Cheng Zhao <zcbenz@gmail.com>
+Date: Thu, 4 Oct 2018 14:57:02 -0700
+Subject: fix: add back virtual methods in AccessiblePaneView
+
+Mark SetPaneFocus and RemovePaneFocus as virtual in AccessiblePaneView, as we
```
codebytere: Can you open the CL to chromium and add that CL to this patch so that we can more easily follow/track this patch to ensure it's deleted in future?

zcbenz: Submitted https://crrev.com/c/1959189.

Issue #20788, opened 2019-10-28
Media key global shortcut no longer working on macOS


salomvary:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v8.0.0-beta.1, all versions from and including 6 seem to be affected
* **Operating System:**
  * macOS 10.14.6 (18G103)
* **Last Known Working Electron version:**
  * 5

### Expected Behavior

If an application is a trusted accessibility client and `globalShortcut.register("MediaPlayPause", handlerFn)` returns `true`, the  registered `handlerFn` should be invoked whenever the play/pause key is pressed. 

### Actual Behavior

The handler is not invoked.

### To Reproduce
```js
app.on("ready", () => {
  const isTrusted = systemPreferences.isTrustedAccessibilityClient(true);
  console.log("isTrustedAccessibilityClient", isTrusted);

  const registered = globalShortcut.register("MediaPlayPause", () => {
    console.log("MediaPlayPause pressed");
  });
  console.log("Registered MediaPlayPause?", registered);

  const win = new BrowserWindow();
  win.loadURL("http://example.com");
});
```

If you provide a URL, please list the commands required to clone/setup/run your repo e.g.
```sh
$ git clone https://github.com/salomvary/electron-media-keys-bug -b another-bug
$ npm install
$ cd electron-media-keys-bug
$ npm start
```

When prompted, allow Terminal to control the computer, restart the app and repeat the test.

PR #21442, opened 2019-12-09
fix: MediaKey globalShortcuts not working on macOS

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/20788.

This PR overrides `MediaKeysListener::IsMediaKeycode` to also listen for `VolumeUp`, `VolumeDown`, and `VolumeMute`.

We also need to patch out Chromium's usage of `RemoteCommandCenterDelegate`, as it uses `MPRemoteCommandCenter` in a change added in [this CL](https://chromium-review.googlesource.com/c/chromium/src/+/1480702). `MPRemoteCommandCenter` makes it such that `globalShortcuts` in Electron will not work as intended, because by design an app does not receive remote control events until it begins playing audio. This means that a media shortcut would not kick into effect until you, for example, began playing a YouTube video which sort of defeats the purpose of `globalShortcuts`.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed media-specific globalShortcuts not working on macOS.

### DIFF
--- a/patches/chromium/command-ismediakey.patch
+++ b/patches/chromium/command-ismediakey.patch
@@ -3,48 +3,22 @@ From: Jeremy Apthorp <jeremya@chromium.org>
 Date: Wed, 10 Oct 2018 15:07:34 -0700
 Subject: command-ismediakey.patch
 
-define Command::IsMediaKey on mac; copied from //chrome/common/extensions/command.cc,
-which also defines a bunch of other stuff that depends on extensions.
-since we only need IsMediaKey, and we don't want the extensions stuff
-(and aren't compiling command.cc), it's safe to duplicate the
-definition. A candidate for upstreaming would be to move the IsMediaKey
-function into //ui.
+Override MediaKeysListener::IsMediaKeycode to also listen for Volume Up, Volume Down,
+and Mute. We also need to patch out Chromium's usage of RemoteCommandCenterDelegate, as
+it uses MPRemoteCommandCenter. MPRemoteCommandCenter makes it such that GlobalShortcuts 
+in Electron will not work as intended, because by design an app does not receive remote
+control events until it begins playing audio. This means that a media shortcut would not kick
+into effect until you, for example, began playing a YouTube video which sort of defeats the
+purpose of GlobalShortcuts.
+
+At the moment there is no upstream possibility for this; but perhaps Chromium may 
+consider some kind of switch, enabled by default, which would conditionally choose to avoid usage of
+RemoteCommandCenterDelegate on macOS.
 
 Also apply electron/electron@0f67b1866a9f00b852370e721affa4efda623f3a
 and electron/electron@d2368d2d3b3de9eec4cc32b6aaf035cc89921bf1 as
 patches.
 
-diff --git a/chrome/browser/extensions/global_shortcut_listener_mac.mm b/chrome/browser/extensions/global_shortcut_listener_mac.mm
-index befe726af9c10b1563a7fc0bb77cc55f65943d5c..46c6fe08bab8471007f78d3ef227e5195bfdf0e1 100644
---- a/chrome/browser/extensions/global_shortcut_listener_mac.mm
-+++ b/chrome/browser/extensions/global_shortcut_listener_mac.mm
-@@ -21,6 +21,26 @@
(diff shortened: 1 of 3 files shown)


patches/chromium/command-ismediakey.patch
```diff
@@ -3,48 +3,22 @@ From: Jeremy Apthorp <jeremya@chromium.org>
...
- namespace extensions {
- 
-+// NOTE: this is defined in command.cc, but command.cc is full of
-+// chrome-extensions-specific logic that we don't want to depend on.
-+// Since we don't build command.cc in Electron, it's safe to re-define this
-+// function here. Ideally, though, `IsMediaKey` would be the responsibility of
-+// `ui::Accelerator`, rather than `extensions::Command`.
-+
-+// static
-+bool Command::IsMediaKey(const ui::Accelerator& accelerator) {
```
codebytere: `IsMediaKey` in Chromium now [invokes `IsMediaKeyCode`](https://cs.chromium.org/chromium/src/chrome/common/extensions/command.cc?q=ismediakey&sq=package:chromium&l=421) which is also used in other places, so it makes more sense to override there instead of at the `IsMediaKey` level.

patches/chromium/command-ismediakey.patch
```diff
@@ -116,3 +104,23 @@ index f4e3126a4efd66f05c4f13e40ba23db10b8cca96..bb4c1a891dd13855227b39a0e582fd4d
      return event;
    }
  
+@@ -221,19 +230,6 @@ static CGEventRef EventTapCallback(CGEventTapProxy proxy,
+ std::unique_ptr<MediaKeysListener> MediaKeysListener::Create(
+     MediaKeysListener::Delegate* delegate,
+     MediaKeysListener::Scope scope) {
+-  // For Mac OS 10.12.2 or later, we want to use MPRemoteCommandCenter for
```
deepak1556: Extensions allow registering shortcuts with `MPRemoteCommandCenter` code path, can you check how they achieve it https://cs.chromium.org/chromium/src/chrome/browser/extensions/extension_keybinding_registry.cc?l=172-202 and avoid the patch if possible.

codebytere: Right - they do, but under the same constraints that i outlined above. All media key options will only work when there is something to play (a Now Playing item) that will trigger the remote receiver; and will similarly not work in the way that we want them to work for users.

deepak1556: > Extensions face the same problem that we do,

Was this verified ? As I see the shortcut registration is part of https://developer.chrome.com/extensions/commands which doesn't mention about restrictions to media key accelerators.

I maybe wrong here, but if this behavior is verified for the extensions then I am good with the patch.

codebytere: @deepak1556 I verified the behavior in Chrome proper with the following gist: https://gist.github.com/codebytere/52aad32238c74894b6b4eaa80e40893d.

If you load that as an unpacked extension, open the background page, and then press `MediaPlayPause`, you'll see nothing logged to the background page console. if you open youtube, start a video, and then try again, messages will begin to log to console.

deepak1556: Great thanks for verifying, can confirm the behavior.

Style nit: Add this code path behind a buildflag which we can disable instead of deleting in the patch, so that its easier to upstream if that is ever considered.

patches/chromium/command-ismediakey.patch
```diff
@@ -116,3 +104,19 @@ index f4e3126a4efd66f05c4f13e40ba23db10b8cca96..bb4c1a891dd13855227b39a0e582fd4d
      return event;
    }
  
+@@ -233,6 +233,7 @@ static CGEventRef EventTapCallback(CGEventTapProxy proxy,
+   // For Mac OS 10.12.2 or later, we want to use MPRemoteCommandCenter for
+   // getting media keys globally if there is a RemoteCommandCenterDelegate
+   // available.
++#if !BUILDFLAG(ENABLE_MEDIA_KEY_OVERRIDES)
```
deepak1556: File will need `#include "electron/buildflags/buildflags.h"`

Issue #21435, opened 2019-12-09
Refresh in-app-purchase App Store receipt for Mac


thomasdao:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description

Some users reported that they have purchased in app products but it's not reflected in the app. It happens in my app and several others as in this tickets #14945. The in-app receipt is not updated after user purchased the product.

### Proposed Solution
It would be helpful to have a way to refresh the in-app receipt using `SKReceiptRefreshRequest`.

### Alternatives Considered

If user purchases the product again, it works. I also checked with Apple and they said that "if the user deletes the app, and re-installs it from the App Store, the new appStoreReceipt should include the recent transaction". However these solutions are quite bad from user perspective. Having a way to refresh the receipt would make it a lot more reliable to implement in-app purchase.

thomasdao:
Hi @codebytere, would it be possible to look at this feature request again if you have some time? 
Having an API to refresh receipt would really help us a lot, since we are receiving more and more complaints from our users that when they purchase the app, the Premium is not granted to them because the receipt is not updated.

We had created a standalone binary from Swift command line to refresh the receipt but always see the error `The operation couldn’t be completed. (SKErrorDomain error 0.)`. It's probably because StoreKit needs to be setup differently, but we have tried a lot of things but still can't get it to work. I hope this function can be added to Electron - it will help Electron apps to be more reliable. Thanks!

thomasdao:
@kevinlinv I'm using `restoreCompletedTransactions` to refresh the receipt

kevinlinv:
Ah, cool, thanks!

ChrisChou-freeman:
> @kevinlinv I'm using `restoreCompletedTransactions` to refresh the receipt

Hi, I'm sorry to bother you. Can you please tell me if the restoreCompletedTransactions interface still has any effect. I have no effect in the macOS 14 system version in the electron 28 version.

thomasdao:
@ChrisChou-freeman yes it works fine for me, I haven't had any issue with it for a long time.

ChrisChou-freeman:
OK, thanks for response.

PR #21451, opened 2019-12-10
[wip] feat: refresh in-app receipts when fetching products

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/21435.

This change ensures that in-app-purchase App Store receipts for macOS are refreshed when a user calls `inAppPurchase.getProducts`.

cc @MarshallOfSound @miniak 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Ensured that in-app receipts are refreshed when fetching products on macOS.

### DIFF
--- a/shell/browser/mac/in_app_purchase_product.mm
+++ b/shell/browser/mac/in_app_purchase_product.mm
@@ -55,12 +55,20 @@
  * @param productIDs - The products' id to fetch.
  */
 - (void)getProducts:(NSSet*)productIDs {
-  SKProductsRequest* productsRequest;
-  productsRequest =
+  SKProductsRequest* productsRequest =
       [[SKProductsRequest alloc] initWithProductIdentifiers:productIDs];
 
   productsRequest.delegate = self;
   [productsRequest start];
+}
+
+/**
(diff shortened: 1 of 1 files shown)


thomasdao:
@codebytere thanks so much for working on this. I only have a minor feedback, would it be possible to have a separate method like `inAppPurchase.refreshReceipt`? The usage is like this: after user purchased the product, we would call `inAppPurchase.refreshReceipt`, wait for it to complete, then send the receipt to our own server to validate it. Having it in `inAppPurchase .getProducts` is a little unintuitive, and it's not clear that when we receive the response, does the receipt really get refreshed? Thanks again for your work!

codebytere:
@thomasdao sure, honestly i'm not super opinionated either way! my initial approach was just the one that required the least amount of refactoring to enable the new functionality. I'll keep poking at this in the imminent future to see how i can make it more intuitive :)

zcbenz:
I'm adding `wip` tag since according to the comments above this PR would need more work.

codebytere:
Closing for now - i'll come back to this when i can.

Issue #20688, opened 2019-10-22
session.cookies.set is broken in 7.0


sgd2z:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.0.0
* **Operating System:**
  * Windows 10 (1903)
* **Last Known Working Electron version:**
  * 6.0.12

### Expected Behavior
session.defaultSession.cookies.set(cookie) should set a cookie that can be then retrieved using session.defaultSession.cookies.get(filter)

### Actual Behavior
session.defaultSession.cookies.set doesnt seem to be doing anything.

### To Reproduce
Gist:
This gist logs an empty array in 7.0.0, In 6.0.12, it logs the cookie object that was set:
https://gist.github.com/sgd2z/af6abf3cea4bddb1cd62a3e8265c359f

nornagon:
I believe the the issue here is with cookie expiry time. Can you try adding an expiry time to the cookie?

sgd2z:
Thanks, adding an expirationDate made it work. The docs say the parameter is optional though.

binux:
I noticed the same issue, per document
> If omitted then the cookie becomes a session cookie

With some test, seems I have to set `expirationDate` to 0 to set a session cookie, which is a breaking changing.

Treverix:
Thanks for that ticket, saved my day! I was caching and restoring cookies and suddenly lost all session cookies because of that.

a1q123456:
adding expirationDate: 0 to detail object sovled the problem, thx

ChrisHSandN:
`session.cookies.set()` also fails with a blank `name`:

According to the documentation: https://electronjs.org/docs/api/cookies#cookiessetdetails
> name String (optional) - The name of the cookie. Empty by default if omitted.

```js
session.cookie.set({
  url: 'https://example.com',
});
// ❌ Error: Failed to parse cookie
```

```js
session.cookie.set({
  url: 'https://example.com',
  name: '',
});
// ❌ Error: Failed to parse cookie
```

```js
session.cookie.set({
  url: 'https://example.com',
  name: 'foo',
});
// ✔ success
```

[...]

PR #21454, opened 2019-12-10
fix: name and expirationDate should be optional when setting cookie

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/20688.

According to our docs of cookies API (which is derived from Chrome extension's cookies API), the `name` should be optional.

And for `expirationDate`, it should pass `base::Time()` when the option is omitted:
https://cs.chromium.org/chromium/src/chrome/browser/extensions/api/cookies/cookies_api.cc?sq=package:chromium&g=0&l=384
Otherwise it would set a cookie the expires now instead of setting a session cookie, and as result `cookies.set` API does not work correctly without setting `expirationDate`.

This is a regression since 7.x.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix `cookies.set` not working correctly when `name` or `expirationDate` is omitted.

### DIFF
--- a/shell/browser/api/atom_api_cookies.cc
+++ b/shell/browser/api/atom_api_cookies.cc
@@ -139,6 +139,17 @@ void FilterCookieWithStatuses(const base::Value& filter,
                 net::cookie_util::StripStatuses(list));
 }
 
+// Read date property from the dictionary.
+base::Time GetTimeFromDict(const base::DictionaryValue& dict,
+                           base::StringPiece key) {
+  base::Optional<double> value = dict.FindDoubleKey(key);
+  if (!value)  // empty time means ignoring the parameter
+    return base::Time();
+  if (*value == 0)  // FromDoubleT would convert 0 to empty Time
+    return base::Time::UnixEpoch();
+  return base::Time::FromDoubleT(*value);
+}
+
 std::string InclusionStatusToString(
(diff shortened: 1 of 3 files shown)


spec-main/api-session-spec.ts
```diff
@@ -79,8 +88,31 @@ describe('session module', () => {
...
+    })
+
+    it('sets cookies without name', async () => {
+      const { cookies } = session.defaultSession
+      const value = '3'
+
+      await cookies.set({ url, value })
       const cs = await cookies.get({ url })
-      expect(cs.some(c => c.name === name && c.value === value)).to.equal(true)
+      expect(cs.some(c => c.name === '' && c.value === value)).to.equal(true)
```
codebytere: maybe explicitly check `c.name.isEmpty()` here?

shell/browser/api/atom_api_cookies.cc
```diff
@@ -139,6 +139,17 @@ void FilterCookieWithStatuses(const base::Value& filter,
                 net::cookie_util::StripStatuses(list));
 }
 
+// Read date property from the dictionary.
+base::Time GetTimeFromDict(const base::DictionaryValue& dict,
```
nornagon: nit: it would be a little more flexible if this fetched a time from `base::Optional<double>` rather than from a dictionary.

Issue #14945, opened 2018-10-03
Restoring in-app purchases


ivandroid:
* Electron version: 3.0.2
* Operating System (Platform and Version): MacOS High Sierra latest

### Expected behavior

There should be a method in the in-app purchase API to restore all purchased products at once. Apple has rejected my current app build because of the absence of this feature. They want to have a button to restore in-app purchases. 

### Actual behavior

Currently it’s not possible to restore purchased products. It’s possible to buy, but not to restore. 

Thank you for your help, guys!

thomasdao:
I think you can restore purchases by using receipt. Electron already has API for retrieving receipts.

MarshallOfSound:
Sounds like this was resolved

semireg:
It's not clear what `inAppPurchase.getReceiptURL` is doing. Which receipt is this for? Is it always the current transaction?

Additionally, is there an equivalent that calls `[[SKPaymentQueue defaultQueue] restoreCompletedTransactions]`?

nikbabchenko:
> You're right. It's possible indeed. If a product is already purchased, you can restore it without paying again. Apple has approved my current build. Thank you.

How have you implemented that? 
Has the same issue from appstore)

thomasdao:
"Restore purchase" button that Apple required, is a way for users to maintain access to content they’ve already purchased. According to [Apple documentation](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/StoreKitGuide/Chapters/Restoring.html):

`In most cases, all your app needs to do is refresh its receipt and deliver the products in its receipt. The refreshed receipt contains a record of the user’s purchases in this app, on this device or any other device.`

So you could get the receipt in your app, send to Apple server to decrypt and inspect the receipt to find out what user has purchased and then deliver that product to them. You can find more about validating the receipt from [Apple documentation](https://developer.apple.com/library/archive/releasenotes/General/ValidateAppStoreReceipt/Introduction.html#//apple_ref/doc/uid/TP40010573).

ivandroid:
> > You're right. It's possible indeed. If a product is already purchased, you can restore it without paying again. Apple has approved my current build. Thank you.
> 
> How have you implemented that?
> Has the same issue from appstore)

I added a 'restore' button for each in app purchase. Clicking on this button forces the user to buy the corresponding product without paying again, if it's already purchased.

semireg:
[...]

PR #21461, opened 2019-12-10
feat: enable explicit IAP restoration

codebytere:
#### Description of Change

Refs https://github.com/electron/electron/issues/14945.

Explicitly allow restoration for IAPs with `inAppPurchase.restoreCompletedTransactions()`. This method can be called either to install purchases on additional devices, or to restore purchases for an application that the user deleted and reinstalled. 

Previously, this would be unintuitive and difficult for users, since they would need to mock this behavior by re-purchasing the product (which would not charge the user more). This added convenience method adds very little overhead and makes it more straightforward to programmatically ensure the desired behavior.

@MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Explicitly allow product restoration to an app with 
`inAppPurchase.restoreCompletedTransactions()`

### DIFF
--- a/shell/browser/mac/in_app_purchase_observer.mm
+++ b/shell/browser/mac/in_app_purchase_observer.mm
@@ -183,14 +183,14 @@ Transaction::Transaction(const Transaction&) = default;
 Transaction::~Transaction() = default;
 
 TransactionObserver::TransactionObserver() : weak_ptr_factory_(this) {
-  obeserver_ = [[InAppTransactionObserver alloc]
+  observer_ = [[InAppTransactionObserver alloc]
       initWithCallback:base::BindRepeating(
                            &TransactionObserver::OnTransactionsUpdated,
                            weak_ptr_factory_.GetWeakPtr())];
 }
 
 TransactionObserver::~TransactionObserver() {
-  [obeserver_ release];
+  [observer_ release];
 }
 
 }  // namespace in_app_purchase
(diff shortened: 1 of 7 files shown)


shell/browser/mac/in_app_purchase_observer.mm
```diff
@@ -183,14 +183,14 @@ - (void)paymentQueue:(SKPaymentQueue*)queue
 Transaction::~Transaction() = default;
 
 TransactionObserver::TransactionObserver() : weak_ptr_factory_(this) {
-  obeserver_ = [[InAppTransactionObserver alloc]
+  observer_ = [[InAppTransactionObserver alloc]
```
codebytere: typo fix

spec-main/api-in-app-purchase-spec.ts
```diff
@@ -29,22 +34,23 @@ describe('inAppPurchase module', function () {
...
-
-  xit('purchaseProduct() accepts optional arguments', async () => {
-    const success = await inAppPurchase.purchaseProduct('non-exist')
-    expect(success).to.be.false('failed to purchase non-existent product')
-  })
-
-  xit('getProducts() returns an empty list when getting invalid product', async () => {
-    const products = await inAppPurchase.getProducts(['non-exist'])
-    expect(products).to.be.an('array').of.length(0)
+  xdescribe('handles product purchases', () => {
```
codebytere: simpler to group disable these for local testing

codebytere:
```
not ok 1083 webContents module destroy() should not crash when invoked synchronously inside did-finish-load handler
```

is a flake since this PR only touches macOS.

Issue #13722, opened 2018-07-18
Window title in third party app switcher


lasekio:
* Electron Version: 3.0.0.beta.3
* Operating System (Platform and Version): Mac OS X 10.13.6
* Last known working Electron version: 2.0.5

**Expected Behavior**
Window title (`document.title`) shows as window title using third party window switcher like [Contexts](https://contexts.co/)

**Actual behavior**
It shows app name (`Electron`) instead of window title

**To Reproduce**
1. Install contexts app or other window switcher that shows window titles
2. Clone and run untouched [electron-quick-start](https://github.com/electron/electron-quick-start) with `npm start`
3. Run app switcher and you see app with `Hello World!` title (see screenshot)
4. Install electron beta 3 `npm install --save-dev electron@3.0.0-beta.3` and run app again `npm start`
5. Run app switcher and you see app with `Electron` title (see screenshot)

**Screenshots**
Electron v2:
<img width="678" alt="ss1" src="https://user-images.githubusercontent.com/8137000/42895558-94145726-8aba-11e8-8c23-c9dcce152e06.png">

Electron Beta:
<img width="683" alt="ss2" src="https://user-images.githubusercontent.com/8137000/42895724-05a06fce-8abb-11e8-8f8a-7f42c774ef17.png">

**Additional Information**
I know is it some sort of corner case but it is still a regression plus this can indicate non-native windowing

nitsakh:
Interesting.. I use HyperSwitch[https://bahoom.com/hyperswitch] for switching windows and this doesn't repro with it.

lasekio:
Can you try with Contexts? Its free to download

nitsakh:
Yeah! I am able to see the issue with Contexts trial. Not sure what's different/what's going on. 😕 
Will look into this further.
(Also, Contexts works pretty cool!)

nitsakh:
Pretty sure this is a focus related issue in 3.0.x. Probably related to #13913.

codebytere:
^ moved to card, not removed

nitsakh:
Closing as the PR has been merged now.

PR #21462, opened 2019-12-10
fix: restore accessibility window title on macOS

jryans:
#### Description of Change

Electron's `AtomNSWindow` [implements](https://github.com/electron/electron/blob/a6d142a112ccc0cecd33c83ff59479f4485614ba/shell/browser/ui/cocoa/atom_ns_window.mm#L105) `accessibilityAttributeValue` to provide various accessibility info to the OS, including window titles.

Chromium 75 [changed to](https://chromium.googlesource.com/chromium/src/+/69a6b7a326b993bb2faa33ae93ccc80101719d35%5E%21/) Apple's [newer accessibility API](https://developer.apple.com/documentation/appkit/nsaccessibility?language=objc) for window titles in the super class that `AtomNSWindow` inherits from. macOS still supports both the old and new style APIs, but it will prefer the new style if it is implemented. This means the Electron window title is being ignored because the newer API at the Chromium level has taken precedence.

By implementing the newer accessibility API in `AtomNSWindow`, this restores correct accessibility window titles in macOS Electron apps. VS Code is [known to be affected](https://github.com/microsoft/vscode/issues/84195), but presumably all Electron apps have a similar issue.

This is a regression that has been present since Electron 6.0.0 (the first release including the Chromium change above).

It looks like @codebytere has modified these files in the past, potentially they are interested.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed accessibility window title on macOS

### DIFF
--- a/shell/browser/ui/cocoa/atom_ns_window.mm
+++ b/shell/browser/ui/cocoa/atom_ns_window.mm
@@ -103,8 +103,6 @@ bool ScopedDisableResize::disable_resize_ = false;
 }
 
 - (id)accessibilityAttributeValue:(NSString*)attribute {
-  if ([attribute isEqual:NSAccessibilityTitleAttribute])
-    return base::SysUTF8ToNSString(shell_->GetTitle());
   if ([attribute isEqual:NSAccessibilityEnabledAttribute])
     return [NSNumber numberWithBool:YES];
   if (![attribute isEqualToString:@"AXChildren"])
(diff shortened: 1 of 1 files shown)


codebytere:
Thanks! Do you by chance have before and after screenshots or a fiddle to help verify the fix?

deepak1556:
@codebytere can be verified with similar steps from https://github.com/electron/electron/issues/13722

jryans:
I should be able to post before / after screenshots soon as well... I forgot to capture them, and then I messed up my local build a bit by switching versions, so just waiting on fixed build to finish... 😅

deepak1556:
Screen Shot 2019-12-10 at 2 12 24 PM

PR #21469, opened 2019-12-10
fix: restore POST forms that open a new window with target=_blank

loc:
#### Description of Change
Fix a regression introduced in #20719 that caused HTML form submissions to not actually POST data when using `target=_blank`. The original implementation in that PR was in `shell/common/native_mate_converters/network_converter.cc`. Not sure if the removal of those fields was intentional or an oversight. @zcbenz to advise.

#### Checklist

- [x] `npm test` passes
- [x] tests are [added](https://github.com/electron/electron/blob/master/docs/development/testing.md)

#### Release Notes

Notes: Fixed POST-ing HTML forms with target=_blank.

### DIFF
--- a/spec/fixtures/pages/form-with-data.html
+++ b/spec/fixtures/pages/form-with-data.html
@@ -1,0 +1,8 @@
+<html>
+<body>
+  <form id="form" method="post">
+    <input name="greeting" value="hello">
+    <input type="submit" value="submit">
+  </form>
+</body>
+</html>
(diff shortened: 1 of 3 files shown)


spec/fixtures/pages/form-with-data.html
```diff
@@ -0,0 +1,8 @@
+<html>
```
nornagon: please put this fixture in `spec-main`

loc: Wasn't sure of the convention, since all the other fixtures used in `chromium-spec.ts` (via  `fixturesPath`) point to `spec/`. Should I change this one or all of them in `chromium-spec.ts`? I guess I don't see the value in separating them if there's a chance they'll be used in both.

PR #21480, opened 2019-12-11
docs: replace systemPreferences.isDarkMode() with nativeTheme.shouldUseDarkColors

xland:
`systemPreferences.isDarkMode()` is deprecated,
so replace it with `nativeTheme.shouldUseDarkColors` .

Notes: none


docs/tutorial/mojave-dark-mode-guide.md
```diff
@@ -21,12 +21,12 @@ If your app has its own dark mode you should toggle it on and off in sync with t
 this by listening for the theme changed event on Electron's `systemPreferences` module.  E.g.
 
 ```js
-const { systemPreferences } = require('electron')
+const { nativeTheme } = require('electron');
```
MarshallOfSound: This semicolon needs to be removed

PR #21488, opened 2019-12-12
fix: quit after Chromium is fully started

zcbenz:
#### Description of Change

Flush pending tasks before quitting Electron.

This is to fix crash on exit which happens when Electron quits while there are still some initialization work pending in message loop.

This should be able to fix some flaky tests on CI.

This does not completely fix the flaky tests that destroy WebContents, which I believe have other reasons for crashing.

#### Release Notes

Notes: no-notes

### DIFF
--- a/shell/browser/browser.cc
+++ b/shell/browser/browser.cc
@@ -26,6 +26,23 @@
 
 namespace electron {
 
+namespace {
+
+// Call |quit| after Chromium is fully started.
+//
+// This is important for quitting immediately in the "ready" event, when
+// certain initialization task may still be pending, and quitting at that time
+// could end up with crash on exit.
+void RunQuitClosure(base::OnceClosure quit) {
+  // Consume remaining tasks in message loop.
+  base::RunLoop().RunUntilIdle();
+  // On Linux/Windows the "ready" event is emitted in "PreMainMessageLoopRun",
+  // make sure we quit after message loop has run for once.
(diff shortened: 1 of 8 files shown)


zcbenz:
I'm backporting all the way to 5-0-x since this reduces flaky tests.

shell/browser/browser.cc
```diff
@@ -26,6 +26,23 @@
...
+//
+// This is important for quitting immediately in the "ready" event, when
+// certain initialization task may still be pending, and quitting at that time
+// could end up with crash on exit.
+void RunQuitClosure(base::OnceClosure quit) {
+  // Consume remaining tasks in message loop.
+  base::RunLoop().RunUntilIdle();
+  // On Linux/Windows the "ready" event is emitted in "PreMainMessageLoopRun",
+  // make sure we quit after message loop has run for once.
+  base::ThreadTaskRunnerHandle::Get()->PostTask(FROM_HERE, std::move(quit));
```
nornagon: If we're calling `PostTask`, do we also need to be calling `RunUntilIdle`?

zcbenz: Hmm `RunUntilIdle` does seem unnecessary here, let me try without it.

zcbenz:
The `electron-woa-testing` is failing for clipboard tests, I'm confident it is not related to changes here since just a while ago this branch was green, and there are other PRs (https://github.com/electron/electron/pull/21501) also affected by the same failures.

PR #21496, opened 2019-12-12
fix: enforce parent-child relationship in custom postMessage() handler

miniak:
#### Description of Change
Adding missing child-parent relationship enforcement in the `ELECTRON_GUEST_WINDOW_MANAGER_WINDOW_POSTMESSAGE` handler.

Follow-up to #19077

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/guest-window-manager.js
+++ b/lib/browser/guest-window-manager.js
@@ -289,8 +289,8 @@ const handleMessageSync = function (channel, handler) {
   ipcMainUtils.handleSync(channel, makeSafeHandler(handler))
 }
 
-const assertCanAccessWindow = function (contents, guestContents) {
-  if (!canAccessWindow(contents, guestContents)) {
+const securityCheck = function (contents, guestContents, check) {
+  if (!check(contents, guestContents)) {
     console.error(`Blocked ${contents.getURL()} from accessing guestId: ${guestContents.id}`)
     throw new Error(`Access denied to guestId: ${guestContents.id}`)
   }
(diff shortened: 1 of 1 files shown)


lib/browser/guest-window-manager.js
```diff
@@ -321,6 +321,8 @@ handleMessage('ELECTRON_GUEST_WINDOW_MANAGER_WINDOW_POSTMESSAGE', (event, guestC
   // The W3C does not seem to have word on how postMessage should work when the
   // origins do not match, so we do not do |canAccessWindow| check here since
   // postMessage across origins is useful and not harmful.
+  assertCanAccessWindow(event.sender, guestContents, isRelatedWindow)
```
nornagon: comment is wrong now?

miniak: @nornagon `assertCanAccessWindow` defaults to `canAccessWindow` for the check, I am passing `isRelatedWindow` here. Maybe I should rename `assertCanAccessWindow` to something else to avoid confusion?

lib/browser/guest-window-manager.js
```diff
@@ -289,8 +289,8 @@ const handleMessageSync = function (channel, handler) {
   ipcMainUtils.handleSync(channel, makeSafeHandler(handler))
 }
 
-const assertCanAccessWindow = function (contents, guestContents) {
-  if (!canAccessWindow(contents, guestContents)) {
+const securityCheck = function (contents, guestContents, check) {
```
nornagon: Rather than passing the function in, maybe:

```
const failAccessCheck = function(senderContents, receiverContents) {
  console.error(`Blocked ${contents.getURL()} from accessing guestId: ${guestContents.id}`)
  throw new Error(`Access denied to guestId: ${guestContents.id}`)
}
```

and then the usage can be
```
if (!myCheck(...)) {
  failAccessCheck(...)
}
```
which is clearer, IMO.

miniak: maybe, but I need to pass the `event.sender` and `guestContents` arguments to both functions now

PR #21497, opened 2019-12-12
fix: workaround for hang when preventDefault-ing nativeWindowOpen (7-1-x)

loc:
### Description of Change
Addresses the same issues as #21236, except without replacing the current API and cleaning up the involved code paths. Briefly, the issue is that canceling a `nativeWindowOpen`-ed window in the `new-window` event causes a hang because the WebContents is already initializing. This change adds a hook earlier, so we can cancel before it starts loading.

#21236 is the "right way", this change is because we need the workaround sooner than time allows. We'd prefer not to rush the addition of a new API!

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added workaround for nativeWindowOpen hang.


lib/browser/api/web-contents.js
```diff
@@ -286,6 +286,17 @@ WebContents.prototype.loadFile = function (filePath, options = {}) {
   }))
 }
 
+WebContents.prototype.setWindowOpenOverride = function (handler) {
+  this.windowOpenOverrideHandler = handler
```
nornagon: since this variable isn't supposed to be accessed directly, maybe
```suggestion
  this._windowOpenOverrideHandler = handler
```

lib/browser/api/web-contents.js
```diff
@@ -286,6 +286,17 @@ WebContents.prototype.loadFile = function (filePath, options = {}) {
...
+}
+
+WebContents.prototype._callWindowOpenOverride = function (event, url, frameName) {
```
nornagon: inline this function IMO

codebytere:
```
webContents module getAllWebContents() API returns an array of web contents - returns an array of web contents
```
has died, but will ✅ when green:)

PR #21501, opened 2019-12-13
fix: avoid contextBridge crash when RenderFrame address is reused

loc:
#### Description of Change
Fix a crash that was happening because 
a) the RenderFramePersistenceStore map was keyed on a pointer, which are not guaranteed to be unique in the process and
b) the RenderFramePersistanceStore was not removing itself from the map when it was released, causing a bad access exception.

h/t @nornagon for help debugging.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed contextBridge crash when opening and closing many windows.

### DIFF
--- a/shell/renderer/api/atom_api_context_bridge.cc
+++ b/shell/renderer/api/atom_api_context_bridge.cc
@@ -47,20 +47,12 @@ content::RenderFrame* GetRenderFrame(const v8::Local<v8::Object>& value) {
   return content::RenderFrame::FromWebFrame(frame);
 }
 
-std::map<content::RenderFrame*, context_bridge::RenderFramePersistenceStore*>&
-GetStoreMap() {
-  static base::NoDestructor<std::map<
-      content::RenderFrame*, context_bridge::RenderFramePersistenceStore*>>
-      store_map;
-  return *store_map;
-}
-
 context_bridge::RenderFramePersistenceStore* GetOrCreateStore(
     content::RenderFrame* render_frame) {
-  auto it = GetStoreMap().find(render_frame);
-  if (it == GetStoreMap().end()) {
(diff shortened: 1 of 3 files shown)


shell/renderer/api/atom_api_context_bridge.cc
```diff
@@ -47,20 +47,12 @@ content::RenderFrame* GetRenderFrame(const v8::Local<v8::Object>& value) {
...
-      content::RenderFrame*, context_bridge::RenderFramePersistenceStore*>>
-      store_map;
-  return *store_map;
-}
-
 context_bridge::RenderFramePersistenceStore* GetOrCreateStore(
     content::RenderFrame* render_frame) {
-  auto it = GetStoreMap().find(render_frame);
-  if (it == GetStoreMap().end()) {
+  auto it = context_bridge::GetStoreMap().find(render_frame->GetRoutingID());
```
deepak1556: Will there be a possibility of cross process access to this function ? `GetRoutingID` is not unique across processes.

MarshallOfSound: @deepak1556 No, the bridge operates across two contexts in the same process

deepak1556: Thanks for confirming.

shell/renderer/api/context_bridge/render_frame_context_bridge_store.h
```diff
@@ -58,11 +58,15 @@ class RenderFramePersistenceStore final : public content::RenderFrameObserver {
   // proxy maps are weak globals, i.e. these are not retained beyond
   // there normal JS lifetime.  You must check IsEmpty()
 
+  int32_t routing_id_;
```
ckerr: _(nit)_ This is set in the constructor and never changes, so might as well be `const`

Issue #21408, opened 2019-12-05
Disable the 'remote' module by default


nornagon:
The `remote` module is a serious security liability and a sizable footgun, as I wrote about [here](https://medium.com/@nornagon/electrons-remote-module-considered-harmful-70d69500f31). We should disable it by default and ultimately move it to `electron-userland`.

This is a fairly disruptive change as it will affect any app that's using the `remote` module, which is most of them.

## Deprecation Strategy
1. In Electron 9.x, start emitting a deprecation warning when using `remote` without explicitly enabling the `remote` module via the `enableRemoteModule` WebPreference.
2. In Electron 10.x, set the default value of `enableRemoteModule` to `false`. Apps that use `remote` will need to update to explicitly set `enableRemoteModule` to `true`.
3. Once [WeakRef](https://v8.dev/features/weak-references) is available, refactor `remote` to use that instead of our custom GC hook callbacks.
4. Once `remote` no longer relies on native code, create `electron-userland/remote` and start warning when apps use `require('electron').remote` or set `enableRemoteModule: true`, directing them to the userland version.
5. After 2 major versions of warning on `require('electron').remote`, remove it.

I can see an argument for delaying (1)–(2) until `WeakRef` is available, so we can skip straight to the userland module and only have one change instead of two. However, I think it might be valuable to start warning ASAP so that apps can start the process of migrating away from `remote` if they decide they want to.

AlexSHoffman:
A few questions:

1. What features will we loose with this and is there a way to implement those features in an app without using `remote`?

2. Since `remote` connects main and renderer processes, will we loose the ability to send messages between them via `ipc` and APIs like "native file drag & drop" that rely on it?
And what about things like `remote.getCurrentWindow()` and `remote.app.getPath()` ? Will it become impossible to get current window and get paths of the app directories then?

nornagon:
IPC primitives (`ipcRenderer` and `ipcMain`) will be unaffected. This only affects the `remote` module, which is a convenience layer wrapped on top of those IPC primitives.

The recommended migration path for apps is to replace usage of the remote module with explicit use of the IPC primitives, for example [`ipcRenderer.invoke`](https://electronjs.org/docs/api/ipc-renderer#ipcrendererinvokechannel-args) / [`ipcMain.handle`](https://electronjs.org/docs/api/ipc-main#ipcmainhandlechannel-listener).

[...]

PR #21546, opened 2019-12-16
feat: warn when remote is used without enableRemoteModule: true

nornagon:
#### Description of Change
This is the first step of #21408, issuing a warning when using the 'remote' module without `enableRemoteModule: true`.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Began issuing a deprecation warning when the 'remote' module is used when enableRemoteModule: true isn't explicitly specified. See https://github.com/electron/electron/issue/21408 for more details.

### DIFF
--- a/lib/browser/remote/server.ts
+++ b/lib/browser/remote/server.ts
@@ -321,7 +321,7 @@ const unwrapArgs = function (sender: electron.WebContents, frameId: number, cont
 
 const isRemoteModuleEnabledImpl = function (contents: electron.WebContents) {
   const webPreferences = (contents as any).getLastWebPreferences() || {}
-  return !!webPreferences.enableRemoteModule
+  return webPreferences.enableRemoteModule != null ? !!webPreferences.enableRemoteModule : true
 }
 
 const isRemoteModuleEnabledCache = new WeakMap()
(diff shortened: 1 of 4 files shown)


felixrieseberg:
Looks solid to me, but I wonder if we should add a link. That can happen in a later PR though, I'd first have to write said document.

nornagon:
@felixrieseberg where should we link to?

lib/browser/remote/server.ts
```diff
@@ -321,7 +321,7 @@ const unwrapArgs = function (sender: electron.WebContents, frameId: number, cont
 
 const isRemoteModuleEnabledImpl = function (contents: electron.WebContents) {
   const webPreferences = (contents as any).getLastWebPreferences() || {}
-  return !!webPreferences.enableRemoteModule
+  return webPreferences.enableRemoteModule != null ? !!webPreferences.enableRemoteModule : true
```
miniak: I think the reason why the default was `false` is that devtools webContents don’t have webPreferences and the remote module should be disabled there

nornagon: Hm, do you have suggestions for how we can detect devtools webcontents but also know whether remote module was explicitly enabled?

miniak: I can do this in a follow-up PR if I don’t forget

Issue #20631, opened 2019-10-18
Electron 7, request using net module never contains cookies in response


t57ser:
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.0.0-beta.7
* **Operating System:**
  * Windows 10
* **Last Known Working Electron version:**
  * 6

### Expected Behavior
Cookies are received

### Actual Behavior
response.headers["set-cookie"] never contains cookies

### To Reproduce
Do a request using the net module as described here: https://electronjs.org/docs/api/net
to any site that will return cookies.

codebytere:
cc @zcbenz, this seems like a nws13n issue?

zcbenz:
Yeah it looks like a regression from nws13n.

t57ser:
Also not working in: 8.0.0-beta.2

t57ser:
I have found a workaround:
```
const session = electron.session.fromPartition("AUTH" + Math.random());
const request = electron.net.request({
	url: "http://localhost:3000/logon",
	method: "GET",
	session,
});
request.on("login", function (_, callback) {
	callback("credentials.username", "credentials.password");
});
request.on("response", (response) => {
	response.on("error", (e) => {
		console.error(e);
	});
	session.cookies.get({}).then((c) => {console.log("cookies in session", c);});
});
request.on("error", (e) => {
	console.error(e);
});
request.end();
```
Passing a session will put the cookies into the session.
Note: This does not work when passing "partition" instead of "session"
This workaround also does not work if:
`app.commandLine.appendSwitch("enable-experimental-web-platform-features");` was set.

arantes555:
Author of [electron-fetch](https://www.npmjs.com/package/electron-fetch) here. Seeing the same problem (plus many others...) in the net module in electron7.

arantes555:
(it seems somewhat linked to https://github.com/electron/electron/issues/8891 , but it is not the same issue)

PR #21552, opened 2019-12-17
fix: pass full response headers in net module

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/20631.

Use raw headers instead of filtered headers in net module. This is a regression caused by network service refactoring.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix `set-cookie` header not passed in net module.

### DIFF
--- a/lib/browser/api/net.js
+++ b/lib/browser/api/net.js
@@ -53,18 +53,26 @@ class IncomingMessage extends Readable {
 
   get headers () {
     const filteredHeaders = {}
-    const rawHeaders = this._responseHead.headers
-    Object.keys(rawHeaders).forEach(header => {
-      if (header in filteredHeaders && discardableDuplicateHeaders.has(header)) {
+    const { rawHeaders } = this._responseHead
+    rawHeaders.forEach(header => {
+      if (header.key in filteredHeaders && discardableDuplicateHeaders.has(header.key)) {
         // do nothing with discardable duplicate headers
       } else {
-        if (header === 'set-cookie') {
+        if (header.key === 'set-cookie') {
           // keep set-cookie as an array per Node.js rules
           // see https://nodejs.org/api/http.html#http_message_headers
-          filteredHeaders[header] = rawHeaders[header]
+          if (filteredHeaders[header.key]) {
+            filteredHeaders[header.key].push(header.value)
+          } else {
+            filteredHeaders[header.key] = [header.value]
+          }
         } else {
           // for non-cookie headers, the values are joined together with ', '
-          filteredHeaders[header] = rawHeaders[header].join(', ')
+          if (filteredHeaders[header.key]) {
+            filteredHeaders[header.key] += `, ${header.value}`
+          } else {
+            filteredHeaders[header.key] = header.value
+          }
         }
       }
     })
--- a/lib/browser/api/net.js
+++ b/lib/browser/api/net.js
@@ -55,20 +55,21 @@ class IncomingMessage extends Readable {
     const filteredHeaders = {}
     const { rawHeaders } = this._responseHead
     rawHeaders.forEach(header => {
-      if (header.key in filteredHeaders && discardableDuplicateHeaders.has(header.key)) {
+      if (Object.prototype.hasOwnProperty.call(filteredHeaders, header.key) &&
+          discardableDuplicateHeaders.has(header.key)) {
         // do nothing with discardable duplicate headers
       } else {
         if (header.key === 'set-cookie') {
           // keep set-cookie as an array per Node.js rules
           // see https://nodejs.org/api/http.html#http_message_headers
-          if (filteredHeaders[header.key]) {
+          if (Object.prototype.hasOwnProperty.call(filteredHeaders, header.key)) {
             filteredHeaders[header.key].push(header.value)
           } else {
             filteredHeaders[header.key] = [header.value]
           }
         } else {
           // for non-cookie headers, the values are joined together with ', '
-          if (filteredHeaders[header.key]) {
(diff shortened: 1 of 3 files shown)


shell/browser/api/atom_api_url_loader.cc
```diff
@@ -29,6 +29,28 @@
 #include "shell/common/gin_helper/object_template_builder.h"
 #include "shell/common/node_includes.h"
 
+namespace gin {
+
+template <>
+struct Converter<mojo::InlinedStructPtr<network::mojom::HttpRawHeaderPair>> {
```
nornagon:
```suggestion
struct Converter<network::mojom::HttpRawHeaderPairPtr> {
```

shell/browser/api/atom_api_url_loader.cc
```diff
@@ -29,6 +29,28 @@
...
+  static v8::Local<v8::Value> ToV8(
+      v8::Isolate* isolate,
+      const mojo::InlinedStructPtr<network::mojom::HttpRawHeaderPair>& pair) {
```
nornagon:
```suggestion
      const network::mojom::HttpRawHeaderPairPtr& pair) {
```

lib/browser/api/net.js
```diff
@@ -53,18 +53,26 @@ class IncomingMessage extends Readable {
...
+    rawHeaders.forEach(header => {
+      if (header.key in filteredHeaders && discardableDuplicateHeaders.has(header.key)) {
         // do nothing with discardable duplicate headers
       } else {
-        if (header === 'set-cookie') {
+        if (header.key === 'set-cookie') {
           // keep set-cookie as an array per Node.js rules
           // see https://nodejs.org/api/http.html#http_message_headers
-          filteredHeaders[header] = rawHeaders[header]
+          if (filteredHeaders[header.key]) {
```
nornagon: This has a bug if `header.key` is one of `Object.prototype`'s keys.

e.g.

```js
const obj = {}
const key = 'constructor'
if (obj[key]) {
  obj[key].push('foo')
}
// TypeError: obj[key].push is not a function
```

Instead, check:
```suggestion
          if (Object.prototype.hasOwnProperty(filteredHeaders, header.key)) {
```

lib/browser/api/net.js
```diff
@@ -53,18 +53,26 @@ class IncomingMessage extends Readable {
...
+            filteredHeaders[header.key].push(header.value)
+          } else {
+            filteredHeaders[header.key] = [header.value]
+          }
         } else {
           // for non-cookie headers, the values are joined together with ', '
-          filteredHeaders[header] = rawHeaders[header].join(', ')
+          if (filteredHeaders[header.key]) {
```
nornagon: here too.

shell/browser/api/atom_api_url_loader.cc
zcbenz: The type of `response_head.raw_request_response_info` is `std::vector< mojo::InlinedStructPtr<network::mojom::HttpRawHeaderPair>>`, the `mojo::InlinedStructPtr` can not be omitted here otherwise compilation error would happen.

zcbenz:
> What other kinds of filtering is Chromium doing here? Is it doing any sort of normalization that we might not want to lose? Do we _only_ need the `set-cookie` headers here?

The raw headers are populated from `net::URLRequest::response_headers()`, and it is the same with 6.x, which uses `net::URLRequest` directly: 
https://cs.chromium.org/chromium/src/services/network/loader_util.cc?type=cs&sq=package:chromium&g=0&l=102-120

Currently I'm only aware the `set-cookie` has been filtered because Chromium blocks third party cookies. I couldn't find a way to work around this without patching Chromium as the requests of `net` module are sent from browser process, and setting cookie sites does not seem to work.

zcbenz:
@nornagon Can you take another look at this?

shell/browser/api/atom_api_url_loader.cc
nornagon: I'm confused,

```
../out/Testing/gen/services/network/public/mojom/http_raw_headers.mojom-forward.h
29:using HttpRawHeaderPairPtr = mojo::InlinedStructPtr<HttpRawHeaderPair>;
```

these two types are defined to be identical. how could one cause a compile error?

lib/browser/api/net.js
```diff
@@ -53,18 +53,26 @@ class IncomingMessage extends Readable {
 
   get headers () {
     const filteredHeaders = {}
-    const rawHeaders = this._responseHead.headers
-    Object.keys(rawHeaders).forEach(header => {
-      if (header in filteredHeaders && discardableDuplicateHeaders.has(header)) {
+    const { rawHeaders } = this._responseHead
+    rawHeaders.forEach(header => {
+      if (filteredHeaders.hasOwnProperty(header.key) && discardableDuplicateHeaders.has(header.key)) {
```
nornagon: Ugh, I'm sorry about JavaScript, but actually this has to be written:
```suggestion
      if (Object.prototype.hasOwnProperty.call(filteredHeaders, header.key) && discardableDuplicateHeaders.has(header.key)) {
```

otherwise a header called `hasOwnProperty` will break this code.

shell/browser/api/atom_api_url_loader.cc
zcbenz: Sorry I did not notice the `Ptr` in class name.

lib/browser/api/net.js
zcbenz: 👍 Always learn something new about JavaScript when I thought I knew it enough.

PR #21562, opened 2019-12-18
refactor: update node main instance management

codebytere:
#### Description of Change

Node has long since updated their approach to main instance handling, but we haven't done the same in several years. This updates to take the same approach to libuv polling as is taken by the latest changes in https://github.com/nodejs/node/pull/27220.

cc @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/app/node_main.cc
+++ b/shell/app/node_main.cc
@@ -116,28 +116,40 @@ int NodeMain(int argc, char* argv[]) {
     }
 
     node::LoadEnvironment(env);
+    v8::Isolate* isolate = env->isolate();
 
-    bool more;
-    do {
-      more = uv_run(env->event_loop(), UV_RUN_ONCE);
-      gin_env.platform()->DrainTasks(env->isolate());
-      if (more == false) {
-        node::EmitBeforeExit(env);
+    {
+      v8::SealHandleScope seal(isolate);
+      bool more;
+      do {
+        uv_run(env->event_loop(), UV_RUN_DEFAULT);
+
+        gin_env.platform()->DrainTasks(env->isolate());
+
+        more = uv_loop_alive(env->event_loop());
+        if (more && !env->is_stopping())
+          continue;
+
+        if (!uv_loop_alive(env->event_loop())) {
+          EmitBeforeExit(env);
+        }
 
         // Emit `beforeExit` if the loop became alive either after emitting
         // event, or after running some callbacks.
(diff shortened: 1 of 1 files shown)


shell/app/node_main.cc
```diff
@@ -116,28 +116,40 @@ int NodeMain(int argc, char* argv[]) {
     }
 
     node::LoadEnvironment(env);
+    v8::Isolate* isolate = env->isolate();
+
+    {
+      v8::SealHandleScope seal(isolate);
```
nornagon: TIL what a `SealHandleScope` is!

```
/**
 * A SealHandleScope acts like a handle scope in which no handle allocations
 * are allowed. It can be useful for debugging handle leaks.
 * Handles can be allocated within inner normal HandleScopes.
 */
```

shell/app/node_main.cc
```diff
@@ -116,28 +116,40 @@ int NodeMain(int argc, char* argv[]) {
...
+        }
 
         // Emit `beforeExit` if the loop became alive either after emitting
         // event, or after running some callbacks.
         more = uv_loop_alive(env->event_loop());
-        if (uv_run(env->event_loop(), UV_RUN_NOWAIT) != 0)
-          more = true;
-      }
-    } while (more == true);
+      } while (more == true && !env->is_stopping());
```
nornagon:
```suggestion
      } while (more && !env->is_stopping());
```

shell/app/node_main.cc
```diff
@@ -116,28 +116,40 @@ int NodeMain(int argc, char* argv[]) {
...
+        more = uv_loop_alive(env->event_loop());
+        if (more && !env->is_stopping())
+          continue;
+
+        if (!uv_loop_alive(env->event_loop())) {
+          EmitBeforeExit(env);
+        }
 
         // Emit `beforeExit` if the loop became alive either after emitting
         // event, or after running some callbacks.
```
nornagon: should this comment be moved above the call to `EmitBeforeExit`?

codebytere: @nornagon i'm fairly impartial on this - i felt it better to directly emulate Node.js itself, and this is where the comment is in that codebase

nornagon: ah ok, didn't realise this was copy/pasted from node directly.

PR #21572, opened 2019-12-19
fix: set enable_negotiate_port to false in allowNTLMCredentialsForDom…

nornagon:
Backport of #21571. See that change for details.

Notes: Fixed an issue where calling allowNTLMCredentialsForDomains() could cause a change in Kerberos SPN generation behavior.


naiiooofy12:
naiioooofy012@gmail.com

PR #20967, opened 2019-11-05
fix: disable more private macOS APIs in MAS build (7-1-x)

zcbenz:
Backport of #20965.

Notes: Fix Electron apps getting rejected to Mac App Store.

Issue #21439, opened 2019-12-09
App "freezes" after modal window with print options is closed/cancelled


clusterberries:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.12, 6.1.4
* **Operating System:**
  *  Windows 10
* **Last Known Working Electron version:**
  * 5.0.2

### Expected Behavior
Application responses to user actions after print is cancelled. (log statements are in the console)

### Actual Behavior
Application doesn't handle events (e.g. click) after print is cancelled. (no log statements in the console)

### To Reproduce
Precondition: clone this repository to reproduce the issue:
https://github.com/clusterberries/electron-sample/tree/print-page
```sh
$ git clone https://github.com/clusterberries/electron-sample.git -b print-page
$ npm install
$ npm start
```
1. Click 'print me' button, so modal window appears
2. Close or cancel modal window
3. Click 'print me' and/or 'test' buttons
4. Observe the console in dev tools.

### Additional Information
The issue is not reproduced with Electon 7.1.3.

codebytere:
This is unfortunately known, and can't be fixed in 5-0-x given some changes that were made in Chromium at the time. 5-0-x is also only receiving security fixes at this time, so sadly, your best bet at this time is to upgrade to a version of Electron where this issue is not present.

vladimiry:
Simply referencing the original and still open issue #19768.

PR #21575, opened 2019-12-19
fix: disable private macOS APIs in MAS build except for CAContext/CALayerHost (7-1-x)

zcbenz:
#### Description of Change

Refs #20027.

This PR is split from https://github.com/electron/electron/pull/20967 to only include patches that are safe to merge.

#### Release Notes

Notes: no-notes

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -38,6 +38,7 @@ mas-lssetapplicationlaunchservicesserverconnectionstatus.patch
 ignore_rc_check.patch
 mas_disable_remote_accessibility.patch
 mas_disable_custom_window_frame.patch
+mas_disable_spi_file_type_mappings.patch
 chrome_key_systems.patch
 allow_nested_error_trackers.patch
 blink_initialization_order.patch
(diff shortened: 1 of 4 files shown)


ymmuse:
Hi @MarshallOfSound, The newest 7-1-x mas version still has some private APIs in use. 
There is missing a patch file look at this https://github.com/electron/electron/pull/20965/commits/4fcc2b03bb332bcf995ece96831acef5728e7fb7

soundar78:
@zcbenz 
will CAContext & CALayerHost apis be removed or anything has been decided regarding these apis ?
just checking as apps using 7.x.x depend on this decision.

zcbenz:
@soundar78 Removing those APIs would have significant affects on app's battery performance and we are still seeking from Apple to provide an alternative solution or unblock those APIs. So currently CAContext/CALayerHost are still being included in 7.x and use it for your app may result in getting rejected by Apple.

jarek-foksa:
As of February 9th Apple is still rejecting Electron >= 7 apps because of the usage of CAContext and CALayerHost.

Wouldn't it make more sense to remove those APIs now (just like it was done in Electron 5.x and 6.x) and bring them back when a better solution is provided by Apple? It doesn't look like they are willing to do anything about it anytime soon. A version with poor battery performance is always better than a version that can't be used at all.

Besides, battery performance is not that important for desktop apps, so far none of my existing users have complained about it (I'm using a patched Electron 5.x MAS version, but I have received several complains about [the broken print dialog](https://github.com/electron/electron/issues/21439) that can fixed without updating to Electron 7.

Issue #19447, opened 2019-07-25
Tracking issue: //extensions integration


nornagon:
This is a tracking issue for the work remaining to migrate our Chrome Extensions support away from our ad-hoc polyfills to the official //extensions code in Chromium.

- [x] Preliminary support behind a feature flag https://github.com/electron/electron/pull/17440
- [ ] Sketch out complete API
  - Should include:
    - Loading and unloading extensions
    - Callbacks for relevant events
    - Exposing relevant info, e.g. badging, notifications, other requests from extensions that the main process might want or need to intervene in
- [x] Load extensions on the IO thread instead of the main thread
- [x] Support for background pages #21591
- [x] Support for devtools extensions #19515 #21779
- [x] API Documentation #21809
- [x] Deprecate polyfill-based extensions support #21812
- [x] Enable feature flag by default #21814
- [x] Remove old extensions code #23340
- [ ] Finalize API
- [ ] Assess all TODOs

nornagon:
Building on the API that @samuelmaddock suggested in #17440, here's a draft to be evolved as requirements become evident.

```ts
interface Extension {
  id: string,
  path: string,
  manifest: chrome.Manifest
}

class Session {
  /** Emitted when extension is installed and ready. */
  on(event: 'extension-ready', listener: (event: Event, extension: Extension) => void): this;

  loadExtension(path: string): Extension;
  removeExtension(extensionId: string): void;
  enableExtension(extensionId: string): void;
  disableExtension(extensionId: string): void;

  getExtension(extensionId: string): Extension;
}
```

nornagon:
To remove in the new world:
- `BrowserWindow.addDevToolsExtension`
- `BrowserWindow.removeDevToolsExtension`
- `BrowserWindow.getDevToolsExtensions`

//extensions doesn't distinguish between devtools extensions and non-devtools extensions. We can deprecate these functions and delegate them to `defaultSession.loadExtension` &c. to avoid breaking users without warning.

sentialx:
@nornagon How `chrome.tabs` related functions will be handled? Could I listen for example to `chrome.tabs.create`? I suggest adding for example a `extensions-tabs-create` event which is fired when an extension calls a `chrome.tabs.create` function to the `Session` class.

nornagon:
[...]

PR #21587, opened 2019-12-19
feat: flesh out the api for //extensions

nornagon:
#### Description of Change
This adds a few more APIs to the //extensions system, which is still behind the `enable_electron_extensions` build flag.

No release notes because none of these features are available in released versions of Electron yet.

Ref #19447

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/atom_api_session.cc
+++ b/shell/browser/api/atom_api_session.cc
@@ -634,14 +634,6 @@ void Session::RemoveExtension(const std::string& extension_id) {
   extension_system->RemoveExtension(extension_id);
 }
 
-void Session::EnableExtension(const std::string& extension_id) {
-  NOTIMPLEMENTED() << "TODO(nornagon)";
-}
-
-void Session::DisableExtension(const std::string& extension_id) {
-  NOTIMPLEMENTED() << "TODO(nornagon)";
-}
-
 v8::Local<v8::Value> Session::GetExtension(const std::string& extension_id) {
   auto* registry = extensions::ExtensionRegistry::Get(browser_context());
   const extensions::Extension* extension =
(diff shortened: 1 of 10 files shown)


samuelmaddock:
This looks to be going in a good direction to me! 👍 

The only opinion I have is about the `getLoadedExtensions` method name. I could see there being some confusion around whether loaded extensions includes disabled extensions. `getAllExtensions` might be more clear, but I'm not strongly opinionated about this.

nornagon:
Makes sense to me! Also updated the sketch over in #19447.

shell/browser/api/atom_api_session.cc
```diff
@@ -606,10 +608,59 @@ std::vector<base::FilePath::StringType> Session::GetPreloads() const {
...
+}
+
+void Session::RemoveExtension(const std::string& extension_id) {
   auto* extension_system = static_cast<extensions::AtomExtensionSystem*>(
       extensions::ExtensionSystem::Get(browser_context()));
-  extension_system->LoadExtension(extension_path);
+  extension_system->RemoveExtension(extension_id);
+}
+
+void Session::EnableExtension(const std::string& extension_id) {
```
jkleinsc: Should we wait until these are implemented to add them?

nornagon:
> LGTM, but I think the API WG should weigh in on this.

FWIW, none of these functions are available in any released version of Electron, because these are behind a build flag that is off by default. I agree that we should do API review of the proposed APIs for extensions, over in #19447. I don't think API review should block this PR though as it does not change any exposed APIs of Electron.

shell/browser/api/atom_api_session.cc
nornagon: Sure, seems reasonable. I added them here because it was easy to do them all together but no particular need to keep them in this PR.

shell/browser/api/atom_api_session.h
```diff
@@ -96,7 +96,10 @@ class Session : public gin_helper::TrackableObject<Session>,
 #endif
 
 #if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
-  void LoadChromeExtension(const base::FilePath extension_path);
+  v8::Local<v8::Promise> LoadExtension(const base::FilePath extension_path);
```
zcbenz: Should be `const base::FilePath&`.

PR #21592, opened 2019-12-20
fix: avoid contextBridge double free on garbage collection

loc:
#### Description of Change
Speculatively fix a crash in contextBridge. Because the elements in the hash-collided linked list are each subject to an object life monitor, they will free themselves. So it's redundant (and crash-y) to free next in the destructor.

Additionally, while we're at it, correct the pointer resetting of the linked list shuffling in the destructor.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash in contextBridge that happens on garbage collection

### DIFF
--- a/shell/renderer/api/context_bridge/render_frame_context_bridge_store.cc
+++ b/shell/renderer/api/context_bridge/render_frame_context_bridge_store.cc
@@ -78,11 +78,7 @@ WeakGlobalPairNode::WeakGlobalPairNode(WeakGlobalPair pair) {
   this->pair = std::move(pair);
 }
 
-WeakGlobalPairNode::~WeakGlobalPairNode() {
-  if (next) {
-    delete next;
-  }
-}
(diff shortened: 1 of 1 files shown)


shell/renderer/api/context_bridge/render_frame_context_bridge_store.cc
```diff
@@ -76,11 +78,7 @@ WeakGlobalPairNode::WeakGlobalPairNode(WeakGlobalPair pair) {
   this->pair = std::move(pair);
 }
 
-WeakGlobalPairNode::~WeakGlobalPairNode() {
-  if (next) {
```
nornagon: how... was this ever sensible?

cc @MarshallOfSound

MarshallOfSound: I think past me reasoned that deleting the head of a list should delete everything in the list without thinking about how this particular linked list was being used

PR #21622, opened 2019-12-26
build: document goma usage and add the external binaries

MarshallOfSound:
Initial work at documenting Goma usage, please note that it doesn't actually work yet 😆We're working on it.

Notes: no-notes


MarshallOfSound:
Gonna land this, we can iterate later 👍

PR #21630, opened 2019-12-28
fix: SimpleURLLoaderWrapper redirects (#21566)

### LINKED ISSUE #21566: SimpleURLLoaderWrapper broken with redirects in > 7.1.2
### Preflight Checklist

### Issue Details

* **Electron Version:**
 v7.1.4
* **Operating System:**
 Ubuntu + MacOS
 7.1.2

### Expected Behavior

electron and electron-updater should follow redirect URLs

### Actual Behavior

electron-updater / electron-builder breaks with Electron > 7.1.2 

```
Redirect was cancelled at SimpleURLLoaderWrapper. (electron/js2c/browser_init.js:2561:23) at SimpleURLLoaderWrapper.emit (events.js:203:13)
```

This NUKED my app update system and I think it's a really critical bug that needs to be fixed.

### To Reproduce

I don't have an easy way to reproduce.

You could TRY to fetch this URL:

```
https://github.com/burtonator/polar-bookshelf/releases/download/v1.80.3/Polar-Bookshelf-1.80.3-mac.zip
```

which is what the auto-updater is fetching and then breaks on.

PS ... Electron rocks!  Except for this issue of course.

lutzroeder:
#### Description of Change

Fix #21566.

PR #21304 regressed `electron-updater` as it raises an error if `request.abort()` is called in a redirect handler. For all apps that shipped with `electron-updater` since 7.1.3+ auto-update is not working and users will be stranded with outdated bits. Fix is to not raise an error if the request was aborted within the handler.

@nornagon @MarshallOfSound, I was unable to build and verify this change (mac stuck for many hours in `gclient sync`). Since you introduced this regression, can you help or commit the correct fix and publish an Electron 7.1.x release including the fix?

#### Checklist

- [ ] `npm test` passes

@nornagon @MarshallOfSound

Notes: Fixed an issue in the `net` module where aborting a request during a redirect could cause an error to be thrown.

### DIFF
--- a/lib/browser/api/net.js
+++ b/lib/browser/api/net.js
@@ -398,7 +398,7 @@ class ClientRequest extends Writable {
           this.emit('redirect', statusCode, newMethod, newUrl, headers)
         } finally {
           this._followRedirectCb = null
-          if (!_followRedirect) {
+          if (!_followRedirect && !this._aborted) {
             this._die(new Error('Redirect was cancelled'))
           }
         }


nornagon:
Also, it'd be great if you could add a test for this behavior so we can avoid regressing it in future!

lutzroeder:
@nornagon What would be great is if you can take over this PR, create a validation build, do the testing, add the necessary tests and port the fix to 7-1-x. If you are looking for a repro, download [zip](https://github.com/lutzroeder/netron/releases/download/v3.6.8/Netron-3.6.8-mac.zip) and start the app from command line. It will crash in the right location as will pretty much every recently shipped app using `electron-builder` and `electron-updater`.

sidneys:
This fix is urgently needed, as nearly all Electron In-App updates via GitHub using `electron-updater` are broken as of v7.1.3.

Please fast-track this.
Happy near year!

Issue #21619, opened 2019-12-25
crashReporter.getLastCrashReport always throw error


Leaking:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

```
crashReporter.getLastCrashReport()
```
it always throw this error

```
 'Error processing argument at index 0, conversion failure from undefined'
```

* **Electron Version:**
  * 7.1.2
* **Operating System:**
  * Windows 10 (10.0.17134)

codebytere:
@Leaking this happens if no `crashesDirectory` has been set, which requires initializing the `crashReporter`:

No Crash:
```js
const { crashReporter } = require('electron')

crashReporter.start({
  productName: 'YourName',
  companyName: 'YourCompany',
  submitURL: 'https://your-domain.com/url-to-submit',
  uploadToServer: false
})

crashReporter.getLastCrashReport()
```

Crash:
```js
const { crashReporter } = require('electron')

crashReporter.getLastCrashReport()
```

PR #21643, opened 2019-12-30
refactor: throw an error for getLastCrashReport if crashReporter not started

codebytere:
#### Description of Change

Refs https://github.com/electron/electron/issues/21619.

Previously, if you tried to call `crashReporter.getLastCrashReport()` without starting the crash reporter, it would fail with a very unhelpful error for end users:

before

This PR refactors the error to now show: 

after

Which more clearly indicates to the user what the underlying issue is and the steps they need to take to remedy it.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Updated `crashReporter` to throw an error for `getLastCrashReport` if `crashReporter` not started.

### DIFF
--- a/lib/common/crash-reporter.js
+++ b/lib/common/crash-reporter.js
@@ -54,7 +54,12 @@ class CrashReporter {
   }
 
   getUploadedReports () {
-    return binding.getUploadedReports(this.getCrashesDirectory())
+    const crashDir = this.getCrashesDirectory()
+    if (!crashDir) {
+      throw new Error('crashReporter has not been started')
+    }
+
+    return binding.getUploadedReports(crashDir)
   }
 
   getCrashesDirectory () {


spec-main/api-crash-reporter-spec.ts
```diff
@@ -183,6 +183,24 @@ ifdescribe(!process.mas && !process.env.DISABLE_CRASH_REPORTER_TESTS && process.
...
+  describe('getUploadedReports', () => {
+    it('throws an error if start has not been called', () => {
+      expect(() => {
+        crashReporter.getUploadedReports()
+      }).to.throw(/crashReporter has not been started/)
+    })
+
+    it('returns an array of reports', () => {
+      crashReporter.start({
+        companyName: 'Umbrella Corporation',
```
jkleinsc: Love the companyName 😄

spec-main/api-crash-reporter-spec.ts
```diff
@@ -183,6 +183,24 @@ ifdescribe(!process.mas && !process.env.DISABLE_CRASH_REPORTER_TESTS && process.
     }
   })
 
+  describe('getUploadedReports', () => {
```
jkleinsc: Maybe move this test up higher in the execution order to ensure that a previous test didn't start the crash reporter?

codebytere: I dug into this and unfortunately it looks like the smoke tests do it somehow, so i don't think it'll be possible to replicate the environment needed to throw the error :(

miniak:
@codebytere what about 6-1-x?

PR #21658, opened 2020-01-01
test: retry for a few flaky tests

zcbenz:
#### Description of Change

Add retries for a few flaky tests.

#### Release Notes

Notes: no-notes

### DIFF
--- a/spec-main/api-web-contents-spec.ts
+++ b/spec-main/api-web-contents-spec.ts
@@ -1122,7 +1122,12 @@ describe('webContents module', () => {
       { name: 'did-fail-load', url: '/net-error' }
     ]
     for (const e of events) {
-      it(`should not crash when invoked synchronously inside ${e.name} handler`, async () => {
+      it(`should not crash when invoked synchronously inside ${e.name} handler`, async function () {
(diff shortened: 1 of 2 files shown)


spec-main/api-web-contents-spec.ts
```diff
@@ -1122,7 +1122,12 @@ describe('webContents module', () => {
       { name: 'did-fail-load', url: '/net-error' }
     ]
     for (const e of events) {
-      it(`should not crash when invoked synchronously inside ${e.name} handler`, async () => {
+      it(`should not crash when invoked synchronously inside ${e.name} handler`, async function () {
+        // This test is flaky on Windows CI and we don't know why, but the
+        // purpose of this test is to make sure Electron does not crash so it
+        // is fine to retry this test for a few times.
+        this.retries(3)
```
nornagon: Are we sure this isn't timing out _because it's crashing_, i.e. the test is legitimately failing?

zcbenz: This tests the case that when destroy is called in the event handler the main process would crash due to pointer used after free. So if it is crashing the whole tests would be aborted due to main process crashing instead of timing out.

Issue #21663, opened 2020-01-02
shell.showItemInFolder slow response and double error msg box when access is denied


diegozhu:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v5.0.6
* **Operating System:**
  * Windows 10 (1803)
* **Last Known Working Electron version:**
  * Not sure

### Expected Behavior
response should be faster and "access is denied" should only popup once.

### Actual Behavior
"access is denied" message popup and UI kind of partially hang. cursor style not changed while UI dom element style change properly like hover event.

### To Reproduce
make sure you haven't grant privilege of "C:\\ProgramData\\Microsoft\\Search\\Data
```
app.on('ready', () => {
  console.log(new Date());
shell.showItemInFolder("C:\\ProgramData\\Microsoft\\Search\\Data\\Applications\\Windows\\Windows.edb");
console.log(new Date())
});
```

result:
`2020-01-02T07:37:25.277Z`
`2020-01-02T07:37:45.404Z`

takes 10+ second to continue working and two "access is denied" popup.

### Additional Information
the implementation of showItemInFolder is [here line 290](https://github.com/electron/electron/blob/d3622f9c37e8cd94e47206bb6211346011d5f76b/shell/common/platform_util_win.cc) 

code:
```
hr = SHOpenFolderAndSelectItems(dir_item, base::size(highlight), highlight,
                                  NULL);
  if (FAILED(hr)) {
    // On some systems, the above call mysteriously fails with "file not
    // found" even though the file is there.  In these cases, ShellExecute()
    // seems to work as a fallback (although it won't select the file).
    if (hr == ERROR_FILE_NOT_FOUND) {
      ShellExecute(NULL, L"open", dir.value().c_str(), NULL, NULL, SW_SHOW);
    } else {
      LOG(WARNING) << " " << __func__ << "(): Can't open full_path = \""
                   << full_path.value() << "\""
                   << " hr = " << logging::SystemErrorCodeToString(hr);
      ui::win::OpenFolderViaShell(dir);
    }
  }
```
there's a fallback call OpenFolderViaShell, I think this is the root cause.
[...]

PR #21668, opened 2020-01-03
fix: don't fallback to OpenFolderViaShell

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/21663.

This fallback call breaks from Chromium's approach as seen [here](https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/platform_util_win.cc;l=76-81?q=SHOpenFolderAndSelectItems&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2Fsearch%2F); it was added in [this commit](https://github.com/electron/electron/commit/05aeceeb52eede7d11a1dcfd0761cc8b2e1b6a56) four years ago to account for shortcomings in Chromium's approach which i feel at this point is significantly more robust. This PR thus removes the fallback so that we do not cause unexpected end-user behavior with duplicate popups and lag.

cc @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue with potential duplicate error popups when calling `shell.showItemInFolder` on Windows

### DIFF
--- a/shell/common/platform_util_win.cc
+++ b/shell/common/platform_util_win.cc
@@ -272,19 +272,15 @@ void ShowItemInFolderOnWorkerThread(const base::FilePath& full_path) {
   hr = desktop->ParseDisplayName(NULL, NULL,
                                  const_cast<wchar_t*>(dir.value().c_str()),
                                  NULL, &dir_item, NULL);
-  if (FAILED(hr)) {
-    ui::win::OpenFolderViaShell(dir);
+  if (FAILED(hr))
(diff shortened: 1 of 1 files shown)


miniak:
@codebytere there are still two other calls to `ui::win::OpenFolderViaShell`, is that ok?

codebytere:
@miniak i left those initially but you may be right; all three break from Chromium technically so i'd personally be ok with removing them all.

PR #21698, opened 2020-01-08
docs: event.newGuest for new-window in WebContents and webContents in BrowsweWindow's constructor

weilonge:
#### Description of Change

According to the example codes in the documentation of [`new-window` event](https://electronjs.org/docs/api/web-contents#event-new-window) in `WebContents`, the following properties are both existing but documented:
  - `webContents` in `BrowsweWindow` constructor options
  - `newGuest` in `event` argument of `new-window` handler of `WebContents`

This patch is for adding the related documentations. Also, it provides typescript-definitations for these two properties.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: `newGuest` in `new-window` event of `WebContents` and `webContents` in `BrowsweWindow` constructor were well-documented.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -190,6 +190,7 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
   * `paintWhenInitiallyHidden` Boolean (optional) - Whether the renderer should be active when `show` is `false` and it has just been created.  In order for `document.visibilityState` to work correctly on first load with `show: false` you should set this to `false`.  Setting this to `false` will cause the `ready-to-show` event to not fire.  Default is `true`.
   * `frame` Boolean (optional) - Specify `false` to create a
     [Frameless Window](frameless-window.md). Default is `true`.
+  * `webContents` WebContents (optional) - Specify `webContents` if there is any existing one.
   * `parent` BrowserWindow (optional) - Specify parent window. Default is `null`.
   * `modal` Boolean (optional) - Whether this is a modal window. This only works when the
     window is a child window. Default is `false`.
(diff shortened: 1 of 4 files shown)


weilonge:
Hi @nornagon , this is my first Electron PR, could you help to review it or forward to someone?

This PR is just for fixing the document of `event` object in `new-window` event. Thanks.

docs/api/browser-window.md
```diff
@@ -190,6 +190,7 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
   * `paintWhenInitiallyHidden` Boolean (optional) - Whether the renderer should be active when `show` is `false` and it has just been created.  In order for `document.visibilityState` to work correctly on first load with `show: false` you should set this to `false`.  Setting this to `false` will cause the `ready-to-show` event to not fire.  Default is `true`.
   * `frame` Boolean (optional) - Specify `false` to create a
     [Frameless Window](frameless-window.md). Default is `true`.
+  * `webContents` WebContents (optional) - Specify `webContents` if there is any existing one.
```
MarshallOfSound: I don't think we want to document this as it implies that users are creating their own webContents which isn't documented and is not officially supported.

There is a trade-off here though that `nativeWindowOpen` requires you pass this webContents into the constructor.  There is some work that @loc is doing to make that not-be-a-thing any more though so I'd rather hold for that solution

weilonge: @MarshallOfSound Thanks for the feedback! I am fine with removing this comment since the other part is still a progress for fixing `newGuest` type definition.

Some more context here to explain why we need this.
Our main goal is to get the reference of a new created `BrowserWindow` by calling `window.open`. The example code mentioned in the [doc](https://electronjs.org/docs/api/web-contents#event-new-window) is exactly what we need.

Based on the example code, it looks like there are two missing type definitions:
1. `newGuest` in `new-window` event
2. `webContents` in `BrowserWindow` constructor options.

> I don't think we want to document this as it implies that users are creating their own webContents which isn't documented and is not officially supported.

Does this mean only assigning a customized `webContents` is not officially supported? but `newGuest` usage is fine?

Since it is using in our code already, I would like to know what's the plan for the above example. Would it be deprecated soon or using a different way to achieve the purpose I mentioned above?

weilonge:
Hi @MarshallOfSound  @loc , could you take a look my PR? Thanks.

nornagon:
FYI I've updated the release notes for this to "no release notes", as we do not typically put docs-only changes in our release notes.

PR #14122, opened 2018-08-15
feat: allow window above full screen windows on mac

paulius005:
continuation of https://github.com/electron/electron/pull/11599

tested on macOS 10.13.6

I have followed https://github.com/electron/electron/blob/master/CONTRIBUTING.md and tested locally. Please let me know if I have missed anything or how I can improve this code

notes: Add visibleOnFullScreen option for SetVisibleOnAllWorkspaces for macOS

PR #21706, opened 2020-01-08
feat: remove visibleOnFullscreen option

codebytere:
#### Description of Change

Refs https://github.com/electron/electron/pull/14122.

This was originally added and tested on macOS 10.13.6 High Sierra; unfortunately, however, Apple changed the underlying functionality of `NSWindow`s starting with 10.14 to disallow `NSWindow`s from floating on top of fullscreen apps. The only way to achieve this now is with an `NSPanel`, which we don't utilize within `BrowserWindow`s.

As such, I believe this should be removed for now as it will not work for the majority of macOS users and was a macOS-specific API options as it stood. I also believe this should perhaps target 8, as it simply does not work and thus users stand to gain nothing from having it there to avoid the appearance of a breaking change.

BREAKING CHANGE

cc @electron/wg-api 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Removed the `visibleOnFullscreen` option in `setVisibleOnAllWorkspaces.`

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -1627,12 +1627,9 @@ can still bring up the menu bar by pressing the single `Alt` key.
 
 Returns `Boolean` - Whether the menu bar is visible.
 
-#### `win.setVisibleOnAllWorkspaces(visible[, options])`
+#### `win.setVisibleOnAllWorkspaces(visible)`
 
 * `visible` Boolean
-* `options` Object (optional)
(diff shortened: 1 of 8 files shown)


codebytere:
@nornagon in that case, would you advocate merging this removal into `master` and adding a deprecation warning to 8?

nornagon:
@codebytere sgtm

codebytere:
```
not ok 1089 webContents module destroy() should not crash when invoked synchronously inside did-frame-finish-load handler
```

is a known windows flake.

Issue #18701, opened 2019-06-09
Notification close() causes app crash at exit


johaven:
* **Electron Version:**
  * 5.0.x
* **Operating System:**
  * macOS 10.14.5

### Expected Behavior
Electron does not crash at exit.

### Actual Behavior
The notifications are sent well, i use the close() function to avoid persistence in Notification Center. 

By doing this every time I quit the application I have a message that warns me that the application did not stop properly.

### To Reproduce
```javascript
const notif = new Notification({title: title, body: body, silent: true})
notif.on('click', () => appEvent.emit('show-window'))
notif.show()
setTimeout(() => notif.close(), 10000)
```
After sending notification(s), close application (app.quit()) and the error will appear.

To no longer have the error after quitting app, i do not use the function close () but my application generates a lot of notifications and I would have preferred to handle that.

codebytere:
This is not in itself a reproducible piece of code, as it doesn't give me enough context about how you're using it. Please attach a more complete piece of error-inducing code for the maintainers to triage this issue. I put:

```
  const notif = new Notification({title: 'title', body: 'body', silent: true})
  notif.on('click', () => console.log('hi'))
  notif.show()
  setTimeout(() => notif.close(), 10000)
```

in a quick-start main process and could not reproduce this.

johaven:
@codebytere could you try this:

```javascript
let count = 0

send(title, body) {
        const notif = new Notification({title: title, body: body, silent: true})
        notif.once('click', () => console.log('tick'))
        notif.show()
        setTimeout(() => {notif.close(); if (count === 5){app.quit()}}, 10000)
}

setInterval(() => {this.send('test', 'test'); count += 1},15000)
```

MarshallOfSound:
Fwiw I think I reproed this (either that or I found a different notification crash).  I think it might be related to my recent notification reset fix

codebytere:
Repro'd successfully @johaven, thanks!

Crash report:
```
Received signal 10 BUS_ADRERR 7fff92d041d8

Thread 0 Crashed:: CrBrowserMain  Dispatch queue: com.apple.main-thread
0   ???                           	0x00007fff92d041d8 __CFConstantStringClassReference + 0
1   com.github.Electron.framework 	0x000000010af898de 0x108d6a000 + 35780830
2   com.github.Electron.framework 	0x000000010af8b46f 0x108d6a000 + 35787887
[...]

PR #21709, opened 2020-01-08
fix: Notification crash in destructor on app quit

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/19842.
Closes https://github.com/electron/electron/issues/18701.

Previously, when you close the api notification, that notification forgets about the platform notification. As such, when it is destroyed it will not call [`set_delegate(nullptr)`](https://github.com/electron/electron/blob/0fe6767d6b8992d0ac3036d553ead1e529b854f3/shell/browser/api/atom_api_notification.cc#L77-L80). The notification can be reset [here](https://github.com/electron/electron/blob/0fe6767d6b8992d0ac3036d553ead1e529b854f3/shell/browser/api/atom_api_notification.cc#L205-L210) without the delegate having been reset, which creates the conditions necessary for this crash. This PR fixes that by ensuring that the delegate is _always_ set to `nullptr` before the notification is reset.

cc @MarshallOfSound @nornagon  

h/t to @charlag for great work investigating this!

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash that would occur when Notifications were closed in concert with app termination.

### DIFF
--- a/shell/browser/api/atom_api_notification.cc
+++ b/shell/browser/api/atom_api_notification.cc
@@ -205,6 +205,7 @@ void Notification::NotificationClosed() {
 void Notification::Close() {
   if (notification_) {
     notification_->Dismiss();
+    notification_->set_delegate(nullptr);
     notification_.reset();
   }
 }


codebytere:
@nornagon see https://github.com/electron/electron/issues/19842#issuecomment-559557495 for more context

PR #21732, opened 2020-01-10
chore: deprecate visibleOnFullScreen option

codebytere:
#### Description of Change

Refs https://github.com/electron/electron/pull/21706.

Deprecate the nonfunctional `visibleOnFullScreen` option within `win.setVisibleOnAllWorkspaces` prior to its removal in the next major release version.

cc @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -1627,7 +1627,7 @@ Returns `Boolean` - Whether the menu bar is visible.
 * `visible` Boolean
 * `options` Object (optional)
   * `visibleOnFullScreen` Boolean (optional) _macOS_ - Sets whether
-    the window should be visible above fullscreen windows
+    the window should be visible above fullscreen windows _deprecated_
 
 Sets whether the window should be visible on all workspaces.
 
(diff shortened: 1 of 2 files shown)


docs/api/browser-window.md
```diff
@@ -1627,7 +1627,7 @@ Returns `Boolean` - Whether the menu bar is visible.
 * `visible` Boolean
 * `options` Object (optional)
   * `visibleOnFullScreen` Boolean (optional) _macOS_ - Sets whether
-    the window should be visible above fullscreen windows
+    the window should be visible above fullscreen windows _deprecated_
```
nornagon: I think we usually put `_Deprecated_` next to `_macOS_` etc.?

codebytere: for options (not top-level apis themselves) it seems to be a mix: https://github.com/electron/electron/blob/14cc902ad83a21c1dab7765166e9f3402cb48c3c/docs/api/system-preferences.md#L294

i'm impartial though, happy to move it!

PR #21779, opened 2020-01-15
feat: [extensions] implement a couple of tabs APIs

nornagon:
#### Description of Change
This (sort of) implements chrome.tabs.executeScript. It only works with non-negative tab IDs (i.e. the default of "active tab" will throw a validation error), and it makes no special effort to prevent extensions from accessing "tabs" they haven't requested permission to (though there does seem to be at least some protections in place in the code I've added, mostly copied from Chrome. I'm not sure if those are watertight though.)

Also implements chrome.tabs.sendMessage, which the Ember DevTools extension depends on.

This is enough to get the React and Ember DevTools extensions working when `enable_electron_extensions = true`.

Ref #19447

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


shell/renderer/extensions/electron_extensions_dispatcher_delegate.cc
```diff
@@ -0,0 +1,214 @@
...
+  source_map->RegisterSource("platformKeys.utils", IDR_PLATFORM_KEYS_UTILS_JS);
+  source_map->RegisterSource("terminalPrivate",
+                             IDR_TERMINAL_PRIVATE_CUSTOM_BINDINGS_JS);
+
+  // IME service on Chrome OS.
+  source_map->RegisterSource("chromeos.ime.mojom.input_engine.mojom",
+                             IDR_IME_SERVICE_MOJOM_JS);
+  source_map->RegisterSource("chromeos.ime.service",
+                             IDR_IME_SERVICE_BINDINGS_JS);
+#endif  // defined(OS_CHROMEOS)
```
samuelmaddock: Could remove this section since it's related to Chrome OS

nornagon:
win32 failure unrelated

Issue #8237, opened 2016-12-19
Is there a way to set a custom inset for the macos traffic lights


matt3224:
Hi,

I am currently using `titleBarStyle: 'hidden-inset',` to inset the traffic lights, however I would very much like to inset it further as many designs feature chunky headers. Is there a way to do this?

Screenshot to illustrate my issue using inbox as an example

![screen shot 2016-12-19 at 11 49 37](https://cloud.githubusercontent.com/assets/2401925/21311864/4c84f28e-c5e1-11e6-9824-be2159213e02.png)

Thanks,

MarshallOfSound:
Currently this is not possible, the `hidden-inset` values are hard set at the moment I believe.

MarshallOfSound:
Oh yeah, the good ol'

GitHub issues are for feature requests and bug reports, questions about using Electron should be directed to the [community](https://github.com/electron/electron#community) or to the [Slack Channel](http://atom-slack.herokuapp.com/).

matt3224:
Can we make this a feature request?

MarshallOfSound:
> The APIs provided by macOS only allow us to set the titlebar in two styles, i.e. hidden and hidden-inset, I don't know a way to perfectly reposition the buttons to arbitrary place, and I'm closing this as won't fix since it is very difficult to implement and we will not work on this.

But it you have an idea on implement this, feel free to send a pull request and we would be happy to merge.

Referenced #6821

margox:
You can create a frameless window by setting frame to false in BrowserWindow’s options, and then make control buttons yourself in your view layer, just some button and css, electron provided some api that you can call to close/minimize/maxmize the window

fantasticsid:
Just throwing ideas here:

It's possible to take traffic lights out of the toolbar and put them in the content view. That way we can position them any way we like, using auto layouts etc.

The reason we need to take them out of the toolbar is that the toolbar adds extra constraints to these buttons so it's hard to position them freely 

I experimented with this idea here: https://zhenchao.li/posts/2018-07-04-positioning-traffic-lights-of-your-cocoa-app.html

Happy to work on a PR if there's no strong objections :)

matt3224:
@fantasticsid any joy on a PR?

fantasticsid:
[...]

PR #21781, opened 2020-01-15
feat: custom positioning for traffic light buttons

TonyWuu:
#### Description of Change
[There seems to be a high demand for custom inset traffic light buttons in Electron](https://github.com/electron/electron/issues/8237). 

- I added a new option in `BrowserWindow`, `trafficLightPosition` which sets the `(x,y)` position of the traffic light buttons. 
- It needs to be used in conjunction with `titleBarStyle: 'hidden'` to work. 
- Default to a value of (0, 0), which results in the standard default position, identical to the regular placement in `titleBarStyle: 'hidden'`. 

e.g.
```
titleBarStyle: 'hidden'
trafficLightPosition: {x: 12, y: 42}
```
image

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: add `trafficLightPosition` option in `BrowserWindow` API to allow custom positioning of traffic lights

### DIFF
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -335,6 +335,8 @@ NativeWindowMac::NativeWindowMac(const gin_helper::Dictionary& options,
   options.Get(options::kZoomToPageWidth, &zoom_to_page_width_);
   options.Get(options::kFullscreenWindowTitle, &fullscreen_window_title_);
   options.Get(options::kSimpleFullScreen, &always_simple_fullscreen_);
+  options.Get(options::kTrafficLightOffsetX, &traffic_light_offsetX_);
+  options.Get(options::kTrafficLightOffsetY, &traffic_light_offsetY_);
 
   bool minimizable = true;
   options.Get(options::kMinimizable, &minimizable);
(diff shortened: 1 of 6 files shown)


shell/browser/native_window_mac.mm
```diff
@@ -509,6 +511,46 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
     [NSEvent removeMonitor:wheel_event_monitor_];
 }
 
+void NativeWindowMac::RepositionTrafficLights() {
+  if (!traffic_light_offsetX_ && !traffic_light_offsetY_) {
+    NSLog(@"in here");
```
zcbenz: This can be removed.

shell/browser/native_window_mac.mm
```diff
@@ -622,12 +664,23 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
 bool NativeWindowMac::IsVisible() {
   bool occluded = [window_ occlusionState] == NSWindowOcclusionStateVisible;
 
+  if (title_bar_style_ == TitleBarStyle::HIDDEN && !entering_fullscreen()) {
+    RepositionTrafficLights();
```
zcbenz: It does not seem appropriate to put this in `IsVisible`.

shell/browser/native_window_mac.mm
```diff
@@ -509,6 +511,45 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
     [NSEvent removeMonitor:wheel_event_monitor_];
 }
 
+void NativeWindowMac::RepositionTrafficLights() {
+  if (!traffic_light_offsetX_ && !traffic_light_offsetY_) {
```
MarshallOfSound: This does not allow offsets set to `0` which someone may want

TonyWuu: My thinking was that an offset of 0 should result in the default position of the traffic lights (in `titleBarStyle: 'hidden'` mode). Are you suggesting there's a use case where someone might want an offset smaller than that default position? In that case, negative numbers would work here. 

IMO (x: 0, y: 0) should have a frame of reference relative to the default position, and not the absolute (0, 0) position of the frame

jameskerr:
I'm excited for this! Thanks!

Issue #21708, opened 2020-01-08
webContents.printToPDF rejects in v8.0.0-beta.5


jacobq:
### Preflight Checklist

* [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [X] I have searched the issue tracker for an issue that matches the one I want to file, without success.

Note: related but different issues:

* #21603 -- that one was using the older (callback) API, but this is using the latest (promise) API (and #21639 appears to be a duplicate of that)
* #18093 -- in that case some output is produced, but here a rejection occurs (and the affected electron versions are different).
* #20634 -- that one never ends, but this one rejects shortly after the call occurs

### Issue Details

* **Electron Version:**
  * v8.0.0-beta.5 (also happens in 9.0.0-nightly.20191220 & 9.0.0-nightly.20200108)
* **Operating System:**
  * Debian (bullseye/sid) x64, 5.3.0 kernel
* **Last Known Working Electron version:**
  * 7.1.8, 8.0.0-beta.4, See [bisect](https://github.com/electron/electron/issues/21708#issuecomment-573079603)

### Expected Behavior
Calling `someBrowserWindow.webContents.printToPDF` should not throw exception / generate promise rejection.

### Actual Behavior
Calling `someBrowserWindow.webContents.printToPDF` rejects with `[Error: Failed to generate PDF]`.

### To Reproduce
Fiddle: https://gist.github.com/jacobq/cc5e5e1736b88cb32604f55369afd1c7
Repo: https://github.com/jacobq/electron-quick-start/tree/bug/print-to-pdf

```js
mainWindow.webContents.once('did-finish-load', async function() {
  try {
    const pdfData = await mainWindow.webContents.printToPDF({})
    console.log('Got PDF data -->', pdfData) // never gets here
  } catch(e) {
    // output --> Call to printToPDF threw exception [Error: Failed to generate PDF]
    console.error('Call to printToPDF threw exception', e)
  }
});
```

jacobq:
FWIW I've started bisecting and 2ab56802cb9ea7549c9020731bdc69618e201f0c appears OK.

jacobq:
[...]

Issue #21438, opened 2019-12-09
Enable print-preview specs on master


deepak1556:
Disabled in https://github.com/electron/electron/pull/21000

Upstream change https://chromium-review.googlesource.com/c/chromium/src/+/1917685

PR #21783, opened 2020-01-15
fix: printToPDF failing to generate PDF

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/21708.
Closes https://github.com/electron/electron/issues/21438.

Fixes an issue where `contents.printToPDF` rejected with `[Error: Failed to generate PDF]`. This was caused by a change in Chromium CL [1917685](https://chromium-review.googlesource.com/c/chromium/src/+/1917685): Chromium made a change such that when Skia metafiles were passed to the PDF compositor, the full document PDF would get composited using the same individual page metafile objects as those that are sent to the utility process for composition.

This behavior hinges on access to the Print Preview UI, which we do not make use of within Electron.  More specifically, Chromium stopped copying the correct metafile data to the `metafile_data_region` for our use case, and so when we received the final IPC message that the metafile was ready, it wouldn't have the data we wanted in it. This fixes the issue by ensuring that the metafile data is _always_ copied to the `metafile_data_region`, which both addresses our use case and doesn't interfere with the goal of the referenced CL.

Tested in Fiddle with the following code:

```js
  mainWindow.webContents.once('did-finish-load', async function() {
    try {
      const pdfData = await mainWindow.webContents.printToPDF({})
      const outPath = '/Users/codebytere/Desktop/print_test.pdf'
    fs.writeFile(outPath, pdfData, (error) => {
        if (error) throw error
        console.log(`Wrote PDF to: ${outPath}`)
      })
    } catch(e) {
      console.error(e)
    }
  })
```

and observing that a PDF is successfully created at the target path.

cc @deepak1556 @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where `webContents.printToPDF` failed to properly generate the PDF document.

### DIFF
--- a/patches/chromium/printing.patch
+++ b/patches/chromium/printing.patch
@@ -395,7 +395,7 @@ index 34e1cda035c042c8c4c55acd7716cf9473856fcf..199267d873f2343cdf401bb9a335dd75
    // Tells the RenderFrame to switch the CSS to print media type, render every
    // requested page using the print preview document's frame/node, and then
 diff --git a/components/printing/renderer/print_render_frame_helper.cc b/components/printing/renderer/print_render_frame_helper.cc
-index 41f6c6ba9920c7fe2a0cf91458162afb70762c16..7315a2b5962f42b57d04655cea97a938bd61768f 100644
+index 41f6c6ba9920c7fe2a0cf91458162afb70762c16..5c23f9fba61c791365d9c69910ebb12154a6f238 100644
 --- a/components/printing/renderer/print_render_frame_helper.cc
 +++ b/components/printing/renderer/print_render_frame_helper.cc
 @@ -39,6 +39,7 @@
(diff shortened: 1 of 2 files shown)


deepak1556:
The composition optimization seems to be used only for document type `SkiaDocumentType::MSKP` , is there any way we can set the `printed_doc_type` to `SkiaDocumentType::PDF` which can eliminate the patch ?

codebytere:
@deepak1556 that's calculated [here](https://source.chromium.org/chromium/chromium/src/+/master:components/printing/browser/print_manager_utils.cc;drc=f7927a386d02777cb71b70ccfff48c26f3d73da0;bpv=1;bpt=1;l=70?originalUrl=https:%2F%2Fcs.chromium.org%2F) and would also require a patch; i considered that but felt that would introduce more potential unintended consequences which is why i opted for the above patch 🤔

deepak1556:
Thanks for testing it, I am good with the above fix. But I would also like to find a bit more about if we can live with type `PDF` moving forward even if content is modifiable since we don't support print preview, technically we are generating static content, its a use case chrome doesn't need hence maybe code path tested above aborted. I am just worried if in future the ability to copy the metafile regions of individual pages goes away.

deepak1556:
Fixes https://github.com/electron/electron/issues/21438

PR #21794, opened 2020-01-16
feat: enable pdf viewer

### LINKED ISSUE #12337: Enable PDF Viewer
It was disabled during the Chromium 63 upgrade in #11459.
Please turn `enable_pdf_viewer` flag in the **features.gypi** on, fix compilation and tests.

nornagon:
#### Description of Change
This builds on top of the native extensions support work that's been ongoing (see #19447) to implement the pdfium-based PDF viewer, the same one that Chrome uses.

Fixes #12337.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Restored support for pdfium-based PDF viewer.


BUILD.gn
```diff
@@ -677,6 +677,7 @@ source_set("electron_lib") {
   ]
   if (enable_electron_extensions) {
     sources += filenames.lib_sources_extensions
+    deps += [ "//chrome/browser/resources:component_extension_resources" ]
```
nornagon: NB. this includes some other stuff we don't care about, but it was easier than making a new .grd file and adding it to the resource_ids list.

shell/app/atom_content_client.cc
```diff
@@ -155,7 +156,7 @@ void ComputeBuiltInPlugins(std::vector<content::PepperPluginInfo>* plugins) {
   pdf_info.is_out_of_process = true;
   pdf_info.name = "Chromium PDF Viewer";
   pdf_info.description = "Portable Document Format";
-  pdf_info.path = base::FilePath(FILE_PATH_LITERAL("internal-pdf-viewer"));
+  pdf_info.path = base::FilePath(FILE_PATH_LITERAL("internal-pdf-viewer2"));
```
nornagon: This definitely needs to be cleaned up, I'll follow Chrome a little more closely when I come back to refactor this.

shell/browser/atom_browser_client.cc
```diff
@@ -1045,28 +1065,130 @@ void AtomBrowserClient::RegisterNonNetworkNavigationURLLoaderFactories(
     protocol->RegisterURLLoaderFactories(factories);
 }
 
+namespace {
+
+// The FileURLLoaderFactory provided to the extension background pages.
+// Checks with the ChildProcessSecurityPolicy to validate the file access.
+class FileURLLoaderFactory : public network::mojom::URLLoaderFactory {
```
nornagon: This might not be required actually? TODO: try removing it and see if the pdf extension still works

shell/browser/atom_browser_client.cc
```diff
@@ -1045,28 +1065,130 @@ void AtomBrowserClient::RegisterNonNetworkNavigationURLLoaderFactories(
...
+
+  // Extension with a background page get file access that gets approval from
+  // ChildProcessSecurityPolicy.
+  extensions::ExtensionHost* host =
+      extensions::ProcessManager::Get(web_contents->GetBrowserContext())
+          ->GetBackgroundHostForExtension(extension->id());
+  if (host) {
+    factories->emplace(url::kFileScheme, std::make_unique<FileURLLoaderFactory>(
+                                             render_process_id));
+  }
```
nornagon: this is largely copied from chrome, we might be able to get away with cutting it down a bit by removing stuff we don't really need

shell/browser/extensions/atom_extensions_browser_client.cc
```diff
@@ -127,9 +137,203 @@ base::FilePath AtomExtensionsBrowserClient::GetBundleResourcePath(
...
+  }
+
+  mojo::Receiver<network::mojom::URLLoader> receiver_{this};
+  mojo::Remote<network::mojom::URLLoaderClient> client_;
+  scoped_refptr<net::HttpResponseHeaders> response_headers_;
+  base::WeakPtrFactory<ResourceBundleFileLoader> weak_factory_{this};
+
+  DISALLOW_COPY_AND_ASSIGN(ResourceBundleFileLoader);
+};
+}  // namespace
```
nornagon: this is all copy/pasted from Chromium

shell/browser/extensions/electron_component_extension_resource_manager.cc
```diff
@@ -0,0 +1,98 @@
+// Copyright 2014 The Chromium Authors. All rights reserved.
```
nornagon: This file is copied from Chrome, then I removed a bunch of stuff that isn't relevant to Electron.

shell/browser/plugins/plugin_response_interceptor_url_loader_throttle.cc
```diff
@@ -0,0 +1,200 @@
+// Copyright 2018 The Chromium Authors. All rights reserved.
```
nornagon: This file is copied from chrome with some modifications.

shell/browser/plugins/plugin_utils.cc
```diff
@@ -0,0 +1,61 @@
+// Copyright 2016 The Chromium Authors. All rights reserved.
```
nornagon: also copied from chrome

nornagon:
This PR is ready for review, although I'm still yet to write tests.

shell/browser/extensions/electron_extension_system.cc
```diff
@@ -84,6 +88,34 @@ void ElectronExtensionSystem::InitForRegularProfile(bool extensions_enabled) {
...
+  if (!manifest.get() || !manifest->is_dict()) {
+    LOG(ERROR) << "Failed to parse extension manifest.";
+    return std::unique_ptr<base::DictionaryValue>();
+  }
+  return base::DictionaryValue::From(std::move(manifest));
+}
+
+void ElectronExtensionSystem::LoadComponentExtensions() {
+  std::string utf8_error;
+  std::string pdf_manifest_string = pdf_extension_util::GetManifest();
```
jkleinsc: Should the variable names be made more generic in case we ever want to load other component extensions?

shell/browser/extensions/electron_extensions_browser_client.cc
```diff
@@ -127,9 +137,203 @@ base::FilePath ElectronExtensionsBrowserClient::GetBundleResourcePath(
...
+    std::string temp_str = ui::ReplaceTemplateExpressions(input, *replacements);
+    DCHECK(!temp_str.empty());
+    return base::RefCountedString::TakeString(&temp_str);
+  } else {
+    return bytes;
+  }
+}
+// Loads an extension resource in a Chrome .pak file. These are used by
+// component extensions.
+class ResourceBundleFileLoader : public network::mojom::URLLoader {
```
jkleinsc: Should this class be in its own file?

patches/chromium/add_didinstallconditionalfeatures.patch
```diff
@@ -0,0 +1,137 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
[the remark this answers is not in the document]
nornagon: See #22054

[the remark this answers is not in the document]
jkleinsc: 👍

shell/browser/extensions/electron_extension_system.cc
nornagon: As-is this is hardcoded to only load a single component extension; if we were to want to load others we should pull in more machinery from Chrome for handling them. I elided that machinery here because it seemed like overkill.

shell/browser/extensions/electron_extensions_browser_client.cc
nornagon: updated to directly use the one from chromium.

shell/browser/electron_browser_client.cc
```diff
@@ -1143,28 +1163,136 @@ void ElectronBrowserClient::RegisterNonNetworkNavigationURLLoaderFactories(
...
+  if (!web_contents)
+    return;
+
+  extensions::ElectronExtensionWebContentsObserver* web_observer =
+      extensions::ElectronExtensionWebContentsObserver::FromWebContents(
+          web_contents);
+
+  // There is nothing to do if no ElectronExtensionWebContentsObserver is
+  // attached to the |web_contents|.
+  if (!web_observer)
```
deepak1556: The lifetime of the observer based on `WebContentsUserData` should be based on the lifetime of the `WebContents` , so the above check for `if (!web_contents)` would be good, this is redundant.

shell/browser/electron_browser_client.cc
```diff
@@ -1143,28 +1163,136 @@ void ElectronBrowserClient::RegisterNonNetworkNavigationURLLoaderFactories(
...
+    // Components of chrome that are implemented as extensions or platform apps
+    // are allowed to use chrome://resources/ and chrome://theme/ URLs.
+    allowed_webui_hosts.emplace_back(content::kChromeUIResourcesHost);
+  }
+  if (extension->is_extension()) {
+    // Extensions are allowed to use chrome://favicon/,
+    // chrome://extension-icon/ and chrome://app-icon URLs. Hosted apps are not
+    // allowed because they are served via web servers (and are generally never
+    // given access to Chrome APIs).
+    allowed_webui_hosts.emplace_back(chrome::kChromeUIExtensionIconHost);
```
deepak1556: Does the pdf-viewer access resources from these origins ?

shell/browser/electron_browser_client.cc
```diff
@@ -1143,28 +1163,136 @@ void ElectronBrowserClient::RegisterNonNetworkNavigationURLLoaderFactories(
...
+                                      std::move(allowed_webui_hosts)));
+  }
+
+  // Extension with a background page get file access that gets approval from
+  // ChildProcessSecurityPolicy.
+  extensions::ExtensionHost* host =
+      extensions::ProcessManager::Get(web_contents->GetBrowserContext())
+          ->GetBackgroundHostForExtension(extension->id());
+  if (host) {
+    factories->emplace(url::kFileScheme, std::make_unique<FileURLLoaderFactory>(
```
deepak1556: Will this override the file url loader handled by the network service https://source.chromium.org/chromium/chromium/src/+/master:content/browser/frame_host/render_frame_host_impl.cc;l=5594 ?

shell/browser/extensions/api/tabs/tabs_api.cc
```diff
@@ -130,4 +165,149 @@ bool TabsExecuteScriptFunction::ShouldInsertCSS() const {
...
+  EXTENSION_FUNCTION_VALIDATE(params);
+
+  int tab_id = params->tab_id ? *params->tab_id : -1;
+  auto* contents = electron::api::WebContents::FromWeakMapID(
+      v8::Isolate::GetCurrent(), tab_id);
+  if (!contents)
+    return RespondNow(Error("No such tab"));
+
+  auto* web_contents = contents->web_contents();
+  GURL url(web_contents->GetVisibleURL());
```
deepak1556: Why visible url instead of the committed url ?

shell/browser/extensions/api/tabs/tabs_api.cc
```diff
@@ -130,4 +165,149 @@ bool TabsExecuteScriptFunction::ShouldInsertCSS() const {
...
+
+  return RespondNow(NoArguments());
+}
+
+ExtensionFunction::ResponseAction TabsGetZoomFunction::Run() {
+  std::unique_ptr<tabs::GetZoom::Params> params(
+      tabs::GetZoom::Params::Create(*args_));
+  EXTENSION_FUNCTION_VALIDATE(params);
+
+  int tab_id = params->tab_id ? *params->tab_id : -1;
```
deepak1556: Can extract this common tab check used across all runners into a macro ?

shell/browser/plugins/plugin_utils.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace content {
+class BrowserContext;
+}
+
+class PluginUtils {
+ public:
+  // If there's an extension that is allowed to handle |mime_type|, returns its
+  // ID. Otherwise returns an empty string.
+  static std::string GetExtensionIdForMimeType(
```
deepak1556: Any necessity for the class if the only members are static ?

shell/renderer/renderer_client_base.cc
```diff
@@ -337,6 +346,50 @@ void RendererClientBase::DidSetUserAgent(const std::string& user_agent) {
...
+    const std::string& mime_type) {
+#if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS) && BUILDFLAG(ENABLE_PLUGINS)
+  DCHECK(plugin_element.HasHTMLTagName("object") ||
+         plugin_element.HasHTMLTagName("embed"));
+  content::WebPluginInfo info;
+  info.type = content::WebPluginInfo::PLUGIN_TYPE_BROWSER_PLUGIN;
+  info.name = base::UTF8ToUTF16("Chromium PDF Viewer");
+  info.path = base::FilePath::FromUTF8Unsafe(extension_misc::kPdfExtensionId);
+  info.background_color = content::WebPluginInfo::kDefaultBackgroundColor;
+  info.mime_types.emplace_back("application/pdf", "pdf",
```
deepak1556: would be good to keep the same constants between the utility class plugin declaration and here.

spec-main/chromium-spec.ts
```diff
@@ -973,117 +973,25 @@ describe('chromium features', () => {
...
-      expect(status).to.equal(200)
-      expect(title).to.not.equal('cat.pdf')
-    })
-
-    it('opens when loading a pdf resource in a iframe', (done) => {
-      testPDFIsLoadedInSubFrame('pdf-in-iframe.html', 'preload-pdf-loaded-in-subframe.js', done)
+      w.loadURL(pdfSource)
+      const [, contents] = await emittedOnce(app, 'web-contents-created')
+      expect(contents.getURL()).to.equal('chrome-extension://mhjfbmdgcfjbbpaeojofohoefgiehjai/index.html')
+      await emittedOnce(contents, 'did-finish-load')
```
deepak1556: Is there a better signal or chrome api we can run to see if the pdf is actually loaded , maybe https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/resources/pdf/pdf_viewer.js;bpv=0;bpt=0;l=1521 ? `did-finish-load` on the webcontents wouldn't exactly denote that the extension rendered the pdf.

shell/browser/extensions/api/resources_private/resources_private_api.cc
```diff
@@ -0,0 +1,112 @@
...
+// 2. Create an AddStringsForMyComponent(base::DictionaryValue * dict) method.
+// 3. Tie in that method to the switch statement in Run()
+
+namespace extensions {
+
+namespace {
+
+void AddStringsForPdf(base::DictionaryValue* dict) {
+#if BUILDFLAG(ENABLE_PDF)
+  static constexpr webui::LocalizedString kPdfResources[] = {
```
deepak1556: Have we disabled the `Print` feature in the pdf UI, as I don't think we can provide the capability at the moment ?

shell/browser/plugins/plugin_utils.h
nornagon: just copying from //chrome/browser/plugins/plugin_utils.h

shell/browser/extensions/api/tabs/tabs_api.cc
nornagon: this is largely copied from chrome's version of this, and i think a macro's probably not the right way to refactor this code.

spec-main/chromium-spec.ts
nornagon: i tried a lot of things to try to get a better "loaded" signal but nothing i tried worked. the core issue is that the MimeHandlerGuestView is similar to a <webview>, and it doesn't forward any kind of messages out of the <embed>. the best idea i have is to make preloads work in extension pages and use that to add some hooks, but we don't have that capability yet.

shell/browser/extensions/api/tabs/tabs_api.cc
deepak1556: Cool, yeah the macro idea was simply based on `EXTENSION_FUNCTION_VALIDATE` :)

spec-main/chromium-spec.ts
ericshung: I'm currently having some issue loading [larger PDF file](https://www.hq.nasa.gov/alsj/a17/A17_FlightPlan.pdf). Is this somehow related?

shell/browser/extensions/api/resources_private/resources_private_api.cc
ericshung: somewhat related question, how will the Download feature in the pdf UI be supported?

spec-main/chromium-spec.ts
nornagon: @ericshung no, I don't think that's related, but I'm seeing some other issues with that PDF (including discovering a bug in Chromium itself: https://crbug.com/1051548). Thanks for the report.

BUILD.gn
```diff
@@ -642,12 +640,22 @@ source_set("electron_lib") {
   ]
   if (enable_electron_extensions) {
     sources += filenames.lib_sources_extensions
+    deps += [
+      "shell/browser/extensions/api:api_registration",
+      "shell/common/extensions/api",
+      "shell/common/extensions/api:extensions_features",
+      "//chrome/browser/resources:component_extension_resources",
```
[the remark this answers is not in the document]
nornagon: It is needed, otherwise:
```
Undefined symbols for architecture x86_64:
  "_kComponentExtensionResources", referenced from:
      extensions::ElectronComponentExtensionResourceManager::ElectronComponentExtensionResourceManager() in electron_component_extension_resource_manager.o
      extensions::ElectronComponentExtensionResourceManager::ElectronComponentExtensionResourceManager() in electron_component_extension_resource_manager.o
  "_kComponentExtensionResourcesSize", referenced from:
      extensions::ElectronComponentExtensionResourceManager::ElectronComponentExtensionResourceManager() in electron_component_extension_resource_manager.o
      extensions::ElectronComponentExtensionResourceManager::ElectronComponentExtensionResourceManager() in electron_component_extension_resource_manager.o
```

shell/browser/electron_browser_client.cc
nornagon: nope, removed

shell/browser/extensions/api/tabs/tabs_api.cc
nornagon: This is what Chrome uses.

shell/browser/extensions/api/resources_private/resources_private_api.cc
nornagon: disabled it.

i'd like the download feature to work the same as other downloads, but it currently just does nothing.

shell/browser/electron_browser_client.cc
nornagon: I'm not sure... I copied this logic from Chrome so maybe? How would we tell?

shell/app/electron_content_client.cc
```diff
@@ -166,6 +168,22 @@ void ComputeBuiltInPlugins(std::vector<content::PepperPluginInfo>* plugins) {
       chrome_pdf::PPP_ShutdownModule;
   pdf_info.permissions = ppapi::PERMISSION_PDF | ppapi::PERMISSION_DEV;
   plugins->push_back(pdf_info);
+
+  // NB. in Chrome, this plugin isn't registered until the PDF extension is
+  // loaded. However, in Electron, we load the PDF extension unconditionally
+  // when it is enabled in the build, so we're OK to load the plugin eagerly
+  // here.
+  content::WebPluginInfo info;
```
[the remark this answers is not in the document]
nornagon: Yeah, I agree it's a little weird to see. The reason I did it this way was to reduce the amount of machinery we need to pull in. Currently we have nothing that listens for extension loads, and it was easy enough to put it here. If there's a good reason to make it lazy later on, we can do so. I think this is fairly light in terms of the amount of work it does though so I think it's OK for it to be eager.

shell/browser/electron_browser_client.cc
nornagon: I'm not 100% sure that we never create a WebContents without an ElectronExtensionWebContentsObserver, so I think this check should stay...

(fwiw, I'm pretty sure we sometimes create `content::WebContents` without an `electron::api::WebContents` wrapper, and in that case we might not have created the observer I think?)

shell/renderer/renderer_client_base.cc
nornagon: agree, I think Chrome does some other stuff to deduplicate this. I'll clean up in a followup.

nornagon:
There's still some work to do on this, but I think this is a good first pass and is ready to merge. I'll open issues for:
- Download functionality. This should work similarly to regular download requests from web pages.
- Printing.
- Some incrementally-loaded PDFs don't load past the first page (e.g. https://www.hq.nasa.gov/alsj/a17/A17_FlightPlan.pdf)

tomsontom:
I tried the beta.2 support but my PDFs are rendering in there! If I use the 10.0.0-nightly they do! I don't see any error message, ... so I'm not sure how I could debug why things work in the nightly but fail in 9.0.0

exentrich:
@tomsontom  Same issue

nornagon:
See https://github.com/electron/electron/issues/22286

emmkimme:
Hi @nornagon,
first of all, lot of thanks for restoring the pdf support.

Sorry, if it is obvious for you, but is it normal to not support pdf in memory context ?
`
  if (!browser_context_->IsOffTheRecord())
    LoadComponentExtensions();
}
`
Side effect is when we create a BrowserWindow relying on a session.fromPartition('mysession'), it opens an empty pdf viewer, with errors in devtools : not allowed to open local resource.
If it is expected, would be fine to document this restriction else I can create a defect, let me know.
Regards

rgilotra:
Hi I am posting this comment w.r.t. nightly version 10.x. I see PDF viewer working in development mode. but as soon as it is packaged it reflects same issue "Save As dialog" box as with stable version. 
npm i electron-nightly@10.0.0-nightly.20200310
Could anybody suggest something ?

PR #21803, opened 2020-01-16
fix: disable contextBridge object identity caching

MarshallOfSound:
There's a long story here but here's the short version:

```js
const o = { n: 0 };

contextBridge.expose('foo', {
  get: () => {
    o.n++;
    return o;
  }
}

// Main world
assert(foo.get().n === 1);  // PASS
assert(foo.get().n === 2);  // FAIL
```

This fails because the proxied object was cached but mutations are not proxied (and can't be)

Notes: Removed assurance of object identity equality from the `contextBridge` as it is incapable of handling mutations


buildflags/BUILD.gn
```diff
@@ -22,6 +22,7 @@ buildflag_header("buildflags") {
     "ENABLE_BUILTIN_SPELLCHECKER=$enable_builtin_spellchecker",
     "ENABLE_PICTURE_IN_PICTURE=$enable_picture_in_picture",
     "ENABLE_MEDIA_KEY_OVERRIDES=$enable_media_key_overrides",
+    "ENABLE_CONTEXT_BBRIDGE_OBJECT_IDENTITY_CACHING=$enable_context_bridge_object_identity_caching",
```
nornagon:
```suggestion
    "ENABLE_CONTEXT_BRIDGE_OBJECT_IDENTITY_CACHING=$enable_context_bridge_object_identity_caching",
```

MarshallOfSound:
@nornagon take two, this time support for in-transaction object caching is still there so that recursive objects work 👍

PR #21809, opened 2020-01-17
docs: document new extensions APIs

nornagon:
#### Description of Change
Ref #19447

These APIs aren't yet switched on, but when they are, we should have documentation for them.

I plan to stub out the `BrowserWindow` versions to call `session.defaultSession.*` when `enable_electron_extensions` is turned on.

#### Checklist

#### Release Notes

Notes: none


nornagon:
I'm going to merge this with #21814, since they should be both merged at the same time.

Issue #21597, opened 2019-12-21
Crash when going to http://google.com with any `webRequest` listener


sentialx:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 8.0.0-beta.5
* **Operating System:**
  * Windows 10
* **Last Known Working Electron version:**
  * 8.0.0-beta,4

### Expected Behavior
http://google.com should load without any crash

### Actual Behavior
Going to http://google.com works for the first time. However, when going to the same URL after the cache has been saved in `userData`, it crashes with `3221225477` exit code. The strange thing is that when going to https://google.com, it works perfectly.

### To Reproduce
(I couldn't use Fiddle, since it saves the `userData` temporarily)
https://github.com/sentialx/electron-quick-start/tree/google-crash
Branch `google-crash`

sentialx:
It seems that it also crashes on http://detectadblock.com.

zcbenz:
Can you share more about how to reproduce the crash? I have tried with following steps but could not reproduce:
1. Open the app, click random links
2. Close the app
3. Reopen the app

sentialx:
@zcbenz Ah sorry. I forgot to replace the line
```
mainWindow.loadURL('https://google.com');
```
with
```
mainWindow.loadURL('http://google.com');
```

zcbenz:
I can reproduce this on Windows but not macOS. This seems to be a Windows-only bug.

PR #21838, opened 2020-01-20
fix: crash when doing redirect navigation with webRequest listener

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/21597.

This PR contains 2 fixes:
1. Fix a DCHECK due to lack of `navigation_ui_data`. 
2. Fix crash due to `current_response_` being moved instead of cloned. This is caused by Chromium changing the behavior of their API without changing the signature.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix crash when doing redirect navigation with `webRequest` listener.

### DIFF
--- a/shell/browser/net/proxying_url_loader_factory.cc
+++ b/shell/browser/net/proxying_url_loader_factory.cc
@@ -523,7 +523,7 @@ void ProxyingURLLoaderFactory::InProgressRequest::ContinueToResponseStarted(
   proxied_client_receiver_.Resume();
 
   factory_->web_request_api()->OnResponseStarted(&info_.value(), request_);
-  target_client_->OnReceiveResponse(std::move(current_response_));
+  target_client_->OnReceiveResponse(current_response_.Clone());
 }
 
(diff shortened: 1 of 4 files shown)


shell/browser/net/proxying_url_loader_factory.cc
```diff
@@ -538,8 +541,7 @@ void ProxyingURLLoaderFactory::InProgressRequest::ContinueToBeforeRedirect(
 
   factory_->web_request_api()->OnBeforeRedirect(&info_.value(), request_,
                                                 redirect_info.new_url);
-  target_client_->OnReceiveRedirect(redirect_info,
-                                    std::move(current_response_));
+  target_client_->OnReceiveRedirect(redirect_info, current_response_.Clone());
```
deepak1556: Should we also clone the response in `target_client_->OnReceiveResponse` ?

zcbenz: Yeah it should also be changed, thanks for noticing this.

PR #21842, opened 2020-01-21
test: retry a few more flaky tests

zcbenz:
I have noticed a few constantly flaky tests in recent CI runnings, I think it is safe to retry a few of them.

notes: no-notes

### DIFF
--- a/spec-main/api-menu-spec.ts
+++ b/spec-main/api-menu-spec.ts
@@ -882,7 +882,12 @@ describe('Menu module', function () {
       expect(output).to.include('Window has no menu')
     })
 
-    ifit(process.platform !== 'darwin')('does not override null menu on startup', async () => {
+    ifit(process.platform !== 'darwin')('does not override null menu on startup', async function () {
+      // This test is flaky on WOA.
+      if (process.platform === 'win32' && process.arch === 'arm64') {
+        this.retries(3)
+      }
+
       const appPath = path.join(fixturesPath, 'api', 'test-menu-null')
       const appProcess = cp.spawn(process.execPath, [appPath])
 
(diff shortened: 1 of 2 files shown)


spec-main/chromium-spec.ts
```diff
@@ -1089,7 +1089,11 @@ describe('chromium features', () => {
 
   describe('window.history', () => {
     describe('window.history.pushState', () => {
-      it('should push state after calling history.pushState() from the same url', (done) => {
+      it('should push state after calling history.pushState() from the same url', function (done) {
+        // In Electron the navigation history is maintained by listening to
+        // WebContents events, so window.history related tests are flaky.
+        this.retries(3)
```
nornagon: Is there some event we can wait for to have a deterministic test instead of retrying?

spec-main/api-menu-spec.ts
```diff
@@ -882,7 +882,12 @@ describe('Menu module', function () {
       expect(output).to.include('Window has no menu')
     })
 
-    ifit(process.platform !== 'darwin')('does not override null menu on startup', async () => {
+    ifit(process.platform !== 'darwin')('does not override null menu on startup', async function () {
+      // This test is flaky on WOA.
```
nornagon: this is surprising to me. do we have any theories as to why?

zcbenz: It is because the child process used for testing crashed on exit. Hmm I think I have just got an idea of how to work around this, will do it in another PR.

spec-main/chromium-spec.ts
zcbenz: There might be some events in navigation controller, I'll have a try.

zcbenz:
Closing since I have a better idea for these tests.

Issue #21596, opened 2019-12-21
youtube.com not loading on 8.0.0-beta.5


sentialx:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 8.0.0-beta.5
* **Operating System:**
  * Windows 10
* **Last Known Working Electron version:**
  * 8.0.0-beta.4

### Expected Behavior
https://youtube.com should load.

### Actual Behavior
https://youtube.com is not loading.

### To Reproduce
https://gist.github.com/e9ea437d3562b951a6619540aaf3da15

### Screenshots
![image](https://user-images.githubusercontent.com/11065386/71311012-519a3900-241b-11ea-90b5-2eb959bc74ea.png)

### Additional information

This may be related to #21242

nornagon:
Bisected on master, the issue was introduced on master within in this range: https://github.com/electron/electron/compare/113b47d871bf97cb901ed5eef5074959c2eef14b...7161b2f57c8a0da87d8e2d54f240c16f5b2afe4b

nornagon:
looks like `--enable-features="WebComponentsV0Enabled"` fixes the issue as a workaround. I suspect that in Chrome, youtube has this enabled as an origin trial or similar, which Electron does not support.

nornagon:
Specifying `enableBlinkFeatures: 'ShadowDOMV0,CustomElementsV0,HTMLImports'` in the WebPreferences also works around the issue.

PR #21858, opened 2020-01-22
fix: some websites using WebComponents V0 not loading

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/21596.
Refs: https://chromium-review.googlesource.com/c/chromium/src/+/1869562

Fixes an issue where any website using Polymer 2.x would not load, as Polymer 2.x [still uses HTML Imports as a loading mechanism](https://www.polymer-project.org/blog/2018-10-02-webcomponents-v0-deprecations). This adds the flag to enable them back for the time being, since Electron does not support origin trials and so wouldn't respect individual sites enabling them. 

Another potential approach to this would be to add `HTMLImports` to the `enableBlinkFeatures` defaults, since that's more granular. Finally, we may choose to simply force consumers to work around the issue themselves, but I would say that for now I personally am in favor of leaving this enabled while it affects a nontrivial number of sites.

cc @nornagon @deepak1556 @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where sites using `ShadowDOMV0`, `CustomElementsV0`, or `HTMLImports` would not load properly.

### DIFF
--- a/shell/browser/feature_list.cc
+++ b/shell/browser/feature_list.cc
@@ -27,6 +27,13 @@ void InitializeFeatureList() {
   // when node integration is enabled.
   disable_features +=
       std::string(",") + features::kSpareRendererForSitePerProcess.name;
+
+  // https://www.polymer-project.org/blog/2018-10-02-webcomponents-v0-deprecations
+  // https://chromium-review.googlesource.com/c/chromium/src/+/1869562
+  // Any website which uses older WebComponents will fail in without this
+  // enabled, since Electron does not support origin trials.
+  enable_features += std::string(",") + "WebComponentsV0Enabled";
+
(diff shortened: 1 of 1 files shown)


deepak1556:
I think build flag is overkill for this one, apps that don't want the feature can just pass it through `--disable-features` to disable it.

LGTM otherwise, nice find. I am surprised youtube is an affected site based on the description from the CL.

zcbenz:
I'm good with this, but all Windows PRs are failing. They seem all flaky, we should have a retry of the tests before merging.

nornagon:
Hm, this shouldn't have been merged as `new-pr` was still applied

zcbenz:
Sorry for merging this too quickly, I didn't know the meaning of `new-pr` label.

PR #21865, opened 2020-01-22
test: use exit event to wait for process exit

zcbenz:
#### Description of Change

For tests that wait for child app to exit, use `exit` event instead of `close` event.

On Windows the node pipe for stdio is not working, so the behavior of `close` event is undefined. From my observation, all timeout tests that started child apps were using `close` event instead of `exit`, so it is possible that the tests failed because they received `close` event before receiving all console output, or because the `close` event did not emit at all.

Even if this change does not really fix the flaky tests, using `exit` event would still be the right way.

#### Release Notes

Notes: no-notes

### DIFF
--- a/spec-main/api-app-spec.ts
+++ b/spec-main/api-app-spec.ts
@@ -160,7 +160,7 @@ describe('app module', () => {
       if (appProcess && appProcess.stdout) {
         appProcess.stdout.on('data', data => { output += data })
       }
-      const [code] = await emittedOnce(appProcess, 'close')
+      const [code] = await emittedOnce(appProcess, 'exit')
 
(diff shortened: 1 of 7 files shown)


zcbenz:
> Is this a known upstream issue ?

Yes, libuv and Node.js are designed to be used in console programs, but Electron is compiled as GUI program. The Win32 API used by libuv to handle stdio do not work for GUI programs. We don't have an idea on how to fix this.

This had been an issue in Electron since its first day, and we have been working around this problem by redirecting node's stdio output to `std::cout`:
https://github.com/electron/electron/blob/master/lib/browser/init.ts#L18-L36

PR #21875, opened 2020-01-23
test: make window.history.pushState test more robust

zcbenz:
#### Description of Change

This PR improves the flaky `window.history.pushState` test with:
1. Load a real page instead of `about:blank`, the latter may have strange behaviors sometimes.
2. Listen to `navigation-entry-commited` event for navigation change.

#### Release Notes

Notes: no-notes

### DIFF
--- a/spec-main/chromium-spec.ts
+++ b/spec-main/chromium-spec.ts
@@ -1089,19 +1089,17 @@ describe('chromium features', () => {
 
   describe('window.history', () => {
     describe('window.history.pushState', () => {
-      it('should push state after calling history.pushState() from the same url', (done) => {
+      it('should push state after calling history.pushState() from the same url', async () => {
         const w = new BrowserWindow({ show: false })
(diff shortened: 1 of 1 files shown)


spec-main/chromium-spec.ts
```diff
@@ -1089,19 +1089,17 @@ describe('chromium features', () => {
...
-          })
-        })
-        w.loadURL('about:blank')
+        await w.loadFile(path.join(fixturesPath, 'pages', 'blank.html'))
+        // History should have current page by now.
+        expect((w.webContents as any).length()).to.equal(1)
+
+        const waitCommit = emittedOnce(w.webContents, 'navigation-entry-commited')
+        w.webContents.executeJavaScript('window.history.pushState({}, "")')
+        await waitCommit
```
nornagon: fwiw there's no need to split the promise from the `await` here, since we're no longer using `remote` for this test.

PR #21880, opened 2020-01-23
fix: About Panel credits should be dark mode aware

codebytere:
#### Description of Change

Fixes an issue surfaced by @bnb where by the Credits option in `app.showAboutPanel()` did not appropriately adapt to dark mode.

Before:
Screen Shot 2020-01-23 at 3 04 34 PM

After:
Screen Shot 2020-01-23 at 3 04 22 PM

cc @MarshallOfSound @zcbenz @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where the credits set in About Panel credits were not dark mode aware on macOS.

### DIFF
--- a/shell/browser/browser_mac.mm
+++ b/shell/browser/browser_mac.mm
@@ -379,20 +379,16 @@ void Browser::ShowAboutPanel() {
   // Credits must be a NSAttributedString instead of NSString
   NSString* credits = (NSString*)options[@"Credits"];
   if (credits != nil) {
-    // Check if app is running in dark mode
-    NSString* mode = [[NSUserDefaults standardUserDefaults]
-        stringForKey:@"AppleInterfaceStyle"];
-    BOOL isDarkMode = [mode isEqualToString:@"Dark"];
+    base::scoped_nsobject<NSMutableDictionary> mutable_options(
+        [options mutableCopy]);
+    base::scoped_nsobject<NSAttributedString> creditString(
+        [[NSAttributedString alloc]
+            initWithString:credits
+                attributes:@{
+                  NSForegroundColorAttributeName : [NSColor textColor]
+                }]);
 
-    // Set color of credits depending on if we're in dark mode or not.
-    NSColor* color = isDarkMode ? [NSColor whiteColor] : [NSColor blackColor];
-    NSAttributedString* creditString = [[NSAttributedString alloc]
-        initWithString:credits
-            attributes:@{NSForegroundColorAttributeName : color}];
-
-    // Cast back to NSDictionary with updated options
-    NSMutableDictionary* mutable_options = [options mutableCopy];
-    mutable_options[@"Credits"] = creditString;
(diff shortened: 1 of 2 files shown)


shell/browser/browser_mac.mm
```diff
@@ -377,11 +377,22 @@ void RemoveFromLoginItems() {
...
-  id credits = options[@"Credits"];
+  NSString* credits = (NSString*)options[@"Credits"];
   if (credits != nil) {
+    // Check if app is running in dark mode
+    NSString* mode = [[NSUserDefaults standardUserDefaults]
+        stringForKey:@"AppleInterfaceStyle"];
+    BOOL isDarkMode = [mode isEqualToString:@"Dark"];
+
+    // Set color of credits depending on if we're in dark mode or not.
+    NSColor* color = isDarkMode ? [NSColor whiteColor] : [NSColor blackColor];
```
sindresorhus: It would be more correct to use `NSColor.textColor` here. It also has the benefit of automatically adapting to dark mode.

codebytere: thanks for the tip, i updated it 🙇‍♀

shell/browser/browser_mac.mm
```diff
@@ -377,11 +377,14 @@ void RemoveFromLoginItems() {
   NSDictionary* options = DictionaryValueToNSDictionary(about_panel_options_);
 
   // Credits must be a NSAttributedString instead of NSString
-  id credits = options[@"Credits"];
+  NSString* credits = (NSString*)options[@"Credits"];
   if (credits != nil) {
+    NSAttributedString* creditString = [[NSAttributedString alloc]
```
zcbenz: Should call `autorelease` or put it in a `base::scoped_nsobject`.

shell/browser/browser_mac.mm
```diff
@@ -377,11 +377,14 @@ void RemoveFromLoginItems() {
...
+        initWithString:credits
+            attributes:@{NSForegroundColorAttributeName : [NSColor textColor]}];
+
     NSMutableDictionary* mutable_options = [options mutableCopy];
```
zcbenz: Not a problem of this PR, but it should also have `autorelease` or be put in a `base::scoped_nsobject`.

shell/browser/browser_mac.mm
```diff
@@ -374,17 +374,24 @@ void RemoveFromLoginItems() {
 }
 
 void Browser::ShowAboutPanel() {
-  NSDictionary* options = DictionaryValueToNSDictionary(about_panel_options_);
+  NSMutableDictionary* mutable_options =
+      [DictionaryValueToNSDictionary(about_panel_options_) mutableCopy];
```
zcbenz: `mutableCopy` returns a copy, the result has also to be managed with `autorelease` or `base::scoped_nsobject`.

And it is better to put the copy inside the `if (credits != nil)` so the copy only happens when necessary.

PR #21882, opened 2020-01-23
ci: Use Windows VM initialized via script

jkleinsc:
#### Description of Change

In order to better support reproducible builds on Windows, this PR changes our Windows CI to use a Windows VM that was initialized by the included script, ` script/setup-win-for-dev.bat`.  Developers can use this script to setup a Windows install with the same exact tooling used by our CI.

Also, I fixed an issue using the source cache on Appveyor.  It turns out that because we kill running builds if a new commit comes in for that branch and the job running is trying to save the source to the cache, we end up in a situation where the source archive is corrupt - eg https://ci.appveyor.com/project/electron-bot/electron-ldhmv/builds/30321997.  I changed the appveyor script to check if the unzip of the archive was successful and if not, it will do a fresh fetch of the source.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,7 +29,7 @@
 
 version: 1.0.{build}
 build_cloud: libcc-20
-image: vs2019-16.3-10.0.18362
+image: vs2019bt-16.4.0
 environment:
   GIT_CACHE_PATH: C:\Users\electron\libcc_cache
   ELECTRON_OUT_DIR: Default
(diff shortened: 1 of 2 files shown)


zcbenz:
The `shell.openExternal` test is failing, I wonder if it is caused by Windows showing some message boxes.

jkleinsc:
@zcbenz yes that was it.  That VM has both Internet Explorer and Microsoft Edge installed and there was a prompt asking which one to use.  I updated the VM to use Microsoft Edge.

PR #2328, opened 2015-07-24
Propagate User App Model ID to Tray Icon

paulcbetts:
This PR prevents dozens of items showing up in the notification area preferences when using Squirrel for Windows, by ensuring that notification tray items are tied to the User App Model ID.

## TODO:

- [ ] Make sure this actually works

PR #21891, opened 2020-01-24
feat: allow GUID parameter to avoid systray demotion on Windows 

bitdisaster:
#### Description of Change
Enhances the Tray API on Windows to avoid demotion of the icon position in the system tray. With normal usage of the Windows API https://docs.microsoft.com/en-us/windows/win32/api/shellapi/ns-shellapi-notifyicondataa a user customized position of the icon is linked to the absolute file path of the executable. If the path changes, mostly by updater frameworks such Squirrel.Windows, the position is lost, and the icon will be placed back in the overflow area.
However, the combination of a code signed binary and GUID parameter can make the positioning of the icon persistent even if the path to the executable changes.
The use of the GUID parameter also fixes a behavior where the position of a second or third icon will be lost on each restart of the App. Even if the path has not changed. 

I’m aware of a previous attempt https://github.com/electron/electron/pull/2328  to introduce the GUID parameter. The reason why it failed and had to be rolled back is bases on Microsofts somewhat  incomplete/confusing documentation. A couple of gotchas:

- The GUID is not a replacement/alternative for the uID. The uID parameter is still needed to identify the icon via Hwnd messages. 
- The GUID parameter is no global Id for the app that seeks to create icons in the system tray. Each icon needs its own GUID if multiple icons are created.
- The GUID is best used in combination with a code signed binary. Once it is used with code signature, the GUID is permanently bound to it.
- The code signature has to contain CN, OU, O in the subject line. 
- The GUID parameter can be used with none code signed binary. However, now the GUID is permanently bound to the path of the binary. If the path changes and the same GUID is used the creation the Tray icon will fail. Since the GUID was statically derived from the AppUserModelId, PR 2328 had to be reverted because it broke all Apps without code signature when they updated and changed there executable path.

Based on above behaviors of the Windows API its best to make this an opt-in. The developer has to be aware of the behavior/requirements using the GUID parameter. It also allows the mitigation of unexpected problems with the Tray creation by providing a new GUID.

#### Checklist

- [X] `npm test` passes

#### Release Notes

Notes: Added GUID parameter to Tray API to avoid system tray icon demotion on Windows

### DIFF
--- a/docs/api/tray.md
+++ b/docs/api/tray.md
@@ -58,9 +58,10 @@ If you want to keep exact same behaviors on all platforms, you should not
 rely on the `click` event and always attach a context menu to the tray icon.
 
 
-### `new Tray(image)`
+### `new Tray(image, [guid])`
 
 * `image` ([NativeImage](native-image.md) | String)
+* `guid` String (optional) _Windows_ - Assigns a GUID to the tray icon. If the executable is signed and the signature contains an organization in the subject line then the GUID is permanently associated with that signature. OS level settings like the position of the tray icon in the system tray will persist even if the path to the executable changes. If the executable is not code-signed then the GUID is permanently associated with the path to the executable. Changing the path to the executable will break the creation of the tray icon and a new GUID must be used. However, it is highly recommended to use the GUID parameter only in conjunction with code-signed executable. If an App defines multiple tray icons then each icon must use a separate GUID.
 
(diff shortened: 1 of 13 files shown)


bitdisaster:
@zcbenz I solved the mac/linux problem a bit differently. @MarshallOfSound recommended the use of converter to me and I like the approach. Does the typedef via conditional compiling work for you to?

PR #21897, opened 2020-01-26
feat: allow fine grain control of BrowserWindow.focus behavior on mac

### LINKED ISSUE #19920: macOS: window.focus() does not focus window anymore
### Issue Details

* **Electron Version:**
  * 6.0.4
* **Operating System:**
  * macOS 10.14.6
  * 4.x

### Expected Behavior
BrowserWindow.focus() puts the window to the front.

### Actual Behavior
Window is not moved to front. 

### To Reproduce
Open a hello world app and add this code:

```js
setTimeout(() => {
	mainWindow.focus();
}, 5000);
```

Click on another app. The window is not focused. It works fine with Electron 4.

tonyferrell:
#### Description of Change

Fixes #19920

Add an optional parameter dictionary to `BrowserWindow.focus(...)` that allows you to control the state of `activateIgnoringOtherApps`. This gives the application creator the option to aggressively steal focus from other applications. The default value is left to to "false" to respect Apple's application development guidelines. I opted for the optional parameter dictionary because a passing a boolean to Focus already has a general meaning (as evidenced by some of the code I replaced) and I felt it could lead to confusion.

I also ended up pushing the concept of `Blur()` a little deeper in the stack, as I thought it improved readability with the mixed set of parameters.

#### Checklist

- [x] `npm test` passes

cc @MarshallOfSound 

#### Release Notes

Notes: Added an optional parameter to `BrowserWindow.focus` providing better control of focus behavior on macOS.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -857,7 +857,10 @@ Try to close the window. This has the same effect as a user manually clicking
 the close button of the window. The web page may cancel the close though. See
 the [close event](#event-close).
 
-#### `win.focus()`
+#### `win.focus([options])`
+
+* `options` Object (optional)
+  * `ignoreOtherApps` Boolean (optional) _macOS_ - Controls whether focus should be stolen when other apps are active. Default is `false`.
 
 Focuses on the window.
 
--- a/shell/browser/api/atom_api_browser_window.cc
+++ b/shell/browser/api/atom_api_browser_window.cc
@@ -295,11 +295,13 @@ void BrowserWindow::OnWindowLeaveFullScreen() {
 #endif
 }
 
-void BrowserWindow::Focus() {
-  if (api_web_contents_->IsOffScreen())
+void BrowserWindow::Focus(
+    const base::Optional<gin_helper::Dictionary>& options) {
+  if (api_web_contents_->IsOffScreen()) {
     FocusOnWebView();
-  else
-    TopLevelWindow::Focus();
+  } else {
+    TopLevelWindow::Focus(options);
+  }
 }
 
 void BrowserWindow::Blur() {
--- a/shell/browser/api/atom_api_browser_window.cc
+++ b/shell/browser/api/atom_api_browser_window.cc
@@ -273,7 +273,7 @@ void BrowserWindow::OnWindowFocus() {
     rwhv->SetActive(true);
 #else
   if (!api_web_contents_->IsDevToolsOpened())
-    web_contents()->Focus();
+    web_contents()->Focus(base::nullopt);
 #endif
 
   TopLevelWindow::OnWindowFocus();
--- a/shell/browser/api/atom_api_browser_window.cc
+++ b/shell/browser/api/atom_api_browser_window.cc
@@ -273,7 +273,7 @@ void BrowserWindow::OnWindowFocus() {
     rwhv->SetActive(true);
 #else
   if (!api_web_contents_->IsDevToolsOpened())
-    web_contents()->Focus(base::nullopt);
+    web_contents()->Focus();
 #endif
 
   TopLevelWindow::OnWindowFocus();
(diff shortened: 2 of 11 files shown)


shell/browser/api/atom_api_browser_window.cc
```diff
@@ -295,11 +295,11 @@ void BrowserWindow::OnWindowLeaveFullScreen() {
 #endif
 }
 
-void BrowserWindow::Focus() {
+void BrowserWindow::Focus(const gin_helper::Dictionary& options) {
```
MarshallOfSound: Because the `options` object is optional you can use `base::Optional`

```suggestion
void BrowserWindow::Focus(const base::Optional<gin_helper::Dictionary>& options) {
```

This means you don't have to do the extra bit of JS 👍

lib/browser/api/browser-window.js
```diff
@@ -25,6 +25,14 @@ BrowserWindow.prototype._init = function () {
     nativeSetBounds.call(this, bounds, ...opts)
   }
 
+  const nativeFocus = this.focus
```
MarshallOfSound: We won't need this if we use `base::Optional`

shell/browser/native_window.h
```diff
@@ -70,7 +70,8 @@ class NativeWindow : public base::SupportsUserData,
   virtual void Close() = 0;
   virtual void CloseImmediately() = 0;
   virtual bool IsClosed() const;
-  virtual void Focus(bool focus) = 0;
+  virtual void Focus(const gin_helper::Dictionary& options) = 0;
```
MarshallOfSound: I think this should be a strict struct that we pass around rather than a "Dictionary" which is basically a wildcard.  I'm thinking just a `FocusOptions` struct.

In general I think we try to draw the line that `api_*` can use gin and v8 and the other things (in this case `native_window`) should use actual structs and such.

tonyferrell: Makes sense to me. Can I just define this in `native_window.h` or is there a better place for it?

MarshallOfSound: @tonyferrell  `native_window.h` makes sense I think 👍

docs/api/browser-window.md
```diff
@@ -859,7 +859,10 @@ Try to close the window. This has the same effect as a user manually clicking
 the close button of the window. The web page may cancel the close though. See
 the [close event](#event-close).
 
-#### `win.focus()`
+#### `win.focus([options])`
+
+* `options` Object (optional)
+  * `ignoreOtherApps` Boolean (optional) - Controls whether focus should be stolen when other apps are active on mac. Default is `false`.
```
ckerr:
```suggestion
  * `ignoreOtherApps` Boolean (optional) _macOS_ - Controls whether focus should be stolen when other apps are active. Default is `false`.
```

shell/browser/native_window.h
```diff
@@ -70,7 +70,8 @@ class NativeWindow : public base::SupportsUserData,
...
+  virtual void Blur() = 0;
```
ckerr: I like that this revision replaces `Focus(true)` and `Focus(false)` with `Focus()` and `Blur()` which is more intuitive IMO

shell/browser/api/atom_api_browser_window.cc
tonyferrell: Ah, cool! I wasn't aware of `baser::Optional` - I'll get that integrated once I've fixed my local build (🤦‍♂) and update the diff.

tonyferrell:
Any thoughts on testing this? I'm not sure the right way to verify this parameter that only impacts focus behavior on mac when other applications are active? I didn't see any obvious examples in other tests, but maybe I looked in the wrong place.

miniak:
@tonyferrell are you still planning to land this PR?

tonyferrell:
I had kind of ignored this because it seemed like the issue I was originally working toward fixing was resolved, but looking through some of the discussion, maybe this is still valuable :)

I'd be happy to take another look at this (or let someone else adopt it?) if there's a use for it!

codebytere:
Closing as abandoned. Happy to re-open if there's movement though!

PR #21904, opened 2020-01-27
fix: ensure tray icon is the proper size on linux

### LINKED ISSUE #21445: No tray icon on Linux (electron@^8)
### Preflight Checklist

### Issue Details

* **Electron Version:** 8.0.0-beta.4
* **Operating System:** Linux (tested on: Manjaro LXDE, Ubuntu 19.04/19.10 Gnome Shell)
* **Last Known Working Electron version:** 7.1.3

### Expected Behavior
Tray icon created.

### Actual Behavior
No tray icon created.

### To Reproduce
https://github.com/vladimiry-playground/electron-quick-start-tray-issue

### Additional Information

The issue has been partially fixed within https://github.com/electron/electron/issues/21154

CC @codebytere

ckerr:
Fixes #21445.

This solves an issue where icons that are larger than the preferred size are not being rendered correctly. This issue surfaced as a side-effect of gfx's change from using the deprecated AppIndicator to now using KdeStatusNotifierItem over dbus.

More info: https://github.com/electron/electron/pull/21904#discussion_r371510492

#### Description of Change

If the icon is too large for the tray, tries to resize it to the correct size.

@codebytere 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed Linux desktop tray icon size regression introduced in the 8.0 betas.

### DIFF
--- a/shell/browser/ui/tray_icon_gtk.cc
+++ b/shell/browser/ui/tray_icon_gtk.cc
@@ -47,8 +47,9 @@ gfx::ImageSkia GetIconFromImage(const gfx::Image& image) {
   if (!size.IsEmpty() &&
       ((size.width() > max_size) || (size.height() > max_size))) {
     const double ratio =
-        max_size / double(std::max(size.width(), size.height()));
-    size = gfx::Size(int(ratio * size.width()), int(ratio * size.height()));
+        max_size / static_cast<double>(std::max(size.width(), size.height()));
+    size = gfx::Size(static_cast<int>(ratio * size.width()),
+                     static_cast<int>(ratio * size.height()));
     icon = gfx::ImageSkiaOperations::CreateResizedImage(
         icon, skia::ImageOperations::RESIZE_BEST, size);
   }
--- a/shell/browser/ui/tray_icon_gtk.cc
+++ b/shell/browser/ui/tray_icon_gtk.cc
@@ -4,12 +4,18 @@
 
 #include "shell/browser/ui/tray_icon_gtk.h"
 
+#include <algorithm>
+
+#include "base/optional.h"
 #include "base/strings/stringprintf.h"
 #include "base/strings/utf_string_conversions.h"
 #include "chrome/browser/ui/views/status_icons/status_icon_linux_dbus.h"
 #include "shell/browser/browser.h"
 #include "shell/common/application_info.h"
 #include "ui/gfx/image/image.h"
+#include "ui/gfx/image/image_skia_operations.h"
+
+#include <gtk/gtk.h>
 
 namespace electron {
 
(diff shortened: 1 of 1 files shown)


shell/browser/ui/tray_icon_gtk.cc
```diff
@@ -4,21 +4,62 @@
...
 TrayIconGtk::TrayIconGtk() = default;
 
 TrayIconGtk::~TrayIconGtk() = default;
 
+int GetPreferredTrayIconPixelSize() {
+  static base::Optional<int> pixels;
+
+  if (!pixels) {
+    static constexpr GtkIconSize icon_size = GTK_ICON_SIZE_LARGE_TOOLBAR;
+    static constexpr int fallback_px = 22;
```
deepak1556: Can you add a comment on how the fallback value is chosen ?

ckerr: Done!

I'm glad you requested that: after some more Googling, I've found the history to not be what I expected. A lot of desktops straight-up coded to an expected height of 22 px, _not_ any GtkIconSize. In my defense, libappindicator's example code does indeed use a GtkIconSize -- but it looks like that works by the coincidence of its values being close to 22 px anyway. I've revised the code to skip GtkIconSize and just pin the height to 22 pixels.

This might not be desirable long-term as it could cause suboptimal icons on high-dpi displays -- but for 8.0.0 it is better than the 8.0.0-beta.X behavior of having 1x1 px tray icons. [This active Chromium issue](https://bugs.chromium.org/p/chromium/issues/detail?id=419673) discusses ongoing upstream work.

More background, if desired, can be found in [this oldish blog post](http://blog.martin-graesslin.com/blog/2014/03/system-tray-in-plasma-next/).

shell/browser/ui/tray_icon_gtk.cc
```diff
@@ -4,21 +4,44 @@
...
 #include "ui/gfx/image/image.h"
+#include "ui/gfx/image/image_skia_operations.h"
 
 namespace electron {
 
 TrayIconGtk::TrayIconGtk() = default;
 
 TrayIconGtk::~TrayIconGtk() = default;
 
+gfx::ImageSkia GetIconFromImage(const gfx::Image& image) {
```
zcbenz: Utility functions should be put in anonymous namespace to avoid name conflicts.

An example can be found at:
https://cs.chromium.org/chromium/src/base/guid.cc?type=cs&g=0&l=16-45

deepak1556: Ah good catch!

ckerr: agreed, nice catch. fixed 418de05.

jkleinsc:
Merging as WOA CI failures unrelated to PR changes.

exander77:
@ckerr This breaks icon size on Linux, Electron icons are tiny compared to other icons, locked to 22px.
Screenshot_2020-04-27_00-02-33

It looks hideous (all my other icons are fine) and if I use numbers in corner, they are unreadable. This is not the correct fix for anything.

shell/browser/ui/tray_icon_gtk.cc
exander77: @deepak1556 @ckerr This is not desirable at all
Screenshot_2020-04-27_00-02-33

The icon is disproportionately small. No other application I use has this problem. And I am not using large icons. If I used 48px icons, this would be crazy disproportionate.

PR #21908, opened 2020-01-27
fix: window.print() only working once

codebytere:
Backport of https://github.com/electron/electron/pull/21893.

See that PR for more details.

Notes: Fixed an issue where `window.print()` only worked once on a single `BrowserWindow`.


codebytere:
Windows failure is flake:

```
not ok 981 webContents module will-prevent-unload event emits if beforeunload returns false
```

PR #21911, opened 2020-01-27
fix: window.print() only working once

codebytere:
Backport of https://github.com/electron/electron/pull/21893.

See that PR for more details.

Notes: Fixed an issue where `window.print()` only worked once on a single `BrowserWindow`.


aitldeveloper:
I am developing a POS application where I need to print on 2 seperate printers (Kitchen Printer & Bill printer) but since 3.1.6 either silent:true is not working or device name not working..

I am in urgent need..please do help me guys....my production is stuck now

PR #21946, opened 2020-01-28
fix: prevent print crash on bad deviceName

codebytere:
#### Description of Change

On [Windows](https://source.chromium.org/chromium/chromium/src/+/master:printing/printing_context_win.cc;l=157?originalUrl=https:%2F%2Fcs.chromium.org%2F) and [macOS](https://source.chromium.org/chromium/chromium/src/+/master:printing/printing_context_mac.mm;l=226?originalUrl=https:%2F%2Fcs.chromium.org%2F), Chromium's printing code will crash if a `deviceName` is provided, but is invalid on the network, i.e does not correspond to a printer that the caller machine is aware of. This bug has been present for a long time, but after playing around a bit I've determined that it's possible to catch faulty `deviceName`s by pulling out some of the printer validation code from Chromium. This PR adds that check for Windows and macOS as well as tests for the same.

Tested with:
```js
win.webContents.on('did-finish-load', () => {
   win.webContents.print({deviceName: 'bad-printer-name'}, (success, reason) => {
   })
})
```

cc @ckerr @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a potential crash on faulty `deviceName`s in `webContents.print()`.

### DIFF
--- a/docs/api/webview-tag.md
+++ b/docs/api/webview-tag.md
@@ -545,9 +545,28 @@ Stops any `findInPage` request for the `webview` with the provided `action`.
 
 * `options` Object (optional)
   * `silent` Boolean (optional) - Don't ask user for print settings. Default is `false`.
-  * `printBackground` Boolean (optional) - Also prints the background color and image of
+  * `printBackground` Boolean (optional) - Prints the background color and image of
     the web page. Default is `false`.
-  * `deviceName` String (optional) - Set the printer device name to use. Default is `''`.
+  * `deviceName` String (optional) - Set the printer device name to use. Must be the system-defined name and not the 'friendly' name, e.g 'Brother_QL_820NWB' and not 'Brother QL-820NWB'.
+  * `color` Boolean (optional) - Set whether the printed web page will be in color or grayscale. Default is `true`.
+  * `margins` Object (optional)
+    * `marginType` String (optional) - Can be `default`, `none`, `printableArea`, or `custom`. If `custom` is chosen, you will also need to specify `top`, `bottom`, `left`, and `right`.
+    * `top` Number (optional) - The top margin of the printed web page, in pixels.
+    * `bottom` Number (optional) - The bottom margin of the printed web page, in pixels.
+    * `left` Number (optional) - The left margin of the printed web page, in pixels.
+    * `right` Number (optional) - The right margin of the printed web page, in pixels.
+  * `landscape` Boolean (optional) - Whether the web page should be printed in landscape mode. Default is `false`.
+  * `scaleFactor` Number (optional) - The scale factor of the web page.
+  * `pagesPerSheet` Number (optional) - The number of pages to print per page sheet.
+  * `collate` Boolean (optional) - Whether the web page should be collated.
+  * `copies` Number (optional) - The number of copies of the web page to print.
+  * `pageRanges` Record<string, number> (optional) - The page range to print. Should have two keys: `from` and `to`.
+  * `duplexMode` String (optional) - Set the duplex mode of the printed web page. Can be `simplex`, `shortEdge`, or `longEdge`.
+  * `dpi` Object (optional)
+    * `horizontal` Number (optional) - The horizontal dpi.
+    * `vertical` Number (optional) - The vertical dpi.
+  * `header` String (optional) - String to be printed as page header.
+  * `footer` String (optional) - String to be printed as page footer.
 
 Returns `Promise<void>`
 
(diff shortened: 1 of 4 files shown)


shell/browser/api/atom_api_web_contents.cc
```diff
@@ -346,6 +350,23 @@ base::Optional<base::TimeDelta> GetCursorBlinkInterval() {
...
 
+#if BUILDFLAG(ENABLE_PRINTING)
+// This will return false if no printer with the provided device_name can be
+// found on the network. We need to check this because Chromium does not do
+// sanity checking of device_name validity and so will crash on invalid names.
+bool DeviceNameValid(const std::string& device_name) {
+#if defined(OS_MACOSX)
+  base::ScopedCFTypeRef<CFStringRef> new_printer_id(
+      base::SysUTF8ToCFStringRef(device_name));
+  return PMPrinterCreateFromPrinterID(new_printer_id.get());
```
zcbenz: Must call `PMRelease` to free the return value.

shell/browser/api/atom_api_web_contents.cc
```diff
@@ -346,6 +350,23 @@ base::Optional<base::TimeDelta> GetCursorBlinkInterval() {
...
 }
 
+#if BUILDFLAG(ENABLE_PRINTING)
+// This will return false if no printer with the provided device_name can be
+// found on the network. We need to check this because Chromium does not do
+// sanity checking of device_name validity and so will crash on invalid names.
+bool DeviceNameValid(const std::string& device_name) {
+#if defined(OS_MACOSX)
+  base::ScopedCFTypeRef<CFStringRef> new_printer_id(
+      base::SysUTF8ToCFStringRef(device_name));
```
zcbenz: `SysUTF16ToCFStringRef` can be used here so `DeviceNameValid` can take `base::string16` as parameter to avoid unnecessary conversions on Windows.

shell/browser/api/atom_api_web_contents.cc
```diff
@@ -346,6 +350,23 @@ base::Optional<base::TimeDelta> GetCursorBlinkInterval() {
   return base::nullopt;
 }
 
+#if BUILDFLAG(ENABLE_PRINTING)
+// This will return false if no printer with the provided device_name can be
+// found on the network. We need to check this because Chromium does not do
+// sanity checking of device_name validity and so will crash on invalid names.
+bool DeviceNameValid(const std::string& device_name) {
```
zcbenz: For function returning bool, `IsDeviceNameValid` might be a better name.

docs/api/webview-tag.md
```diff
@@ -545,9 +545,28 @@ Stops any `findInPage` request for the `webview` with the provided `action`.
 
 * `options` Object (optional)
   * `silent` Boolean (optional) - Don't ask user for print settings. Default is `false`.
-  * `printBackground` Boolean (optional) - Also prints the background color and image of
+  * `printBackground` Boolean (optional) - Prints the background color and image of
     the web page. Default is `false`.
-  * `deviceName` String (optional) - Set the printer device name to use. Default is `''`.
+  * `deviceName` String (optional) - Set the printer device name to use. Must be the system-defined name and not the 'friendly' name, e.g 'Brother_QL_820NWB' and not 'Brother QL-820NWB'.
+  * `color` Boolean (optional) - Set whether the printed web page will be in color or grayscale. Default is `true`.
```
deepak1556: Is documenting these other parameters intentional as part of this PR ?

spec-main/api-web-contents-spec.ts
```diff
@@ -99,20 +99,34 @@ describe('webContents module', () => {
...
     })
 
+    ifit(process.platform !== 'linux')('throws when an invalid deviceName is passed', () => {
+      expect(() => {
+        w.webContents.print({ deviceName: 'i-am-a-nonexistent-printer' }, () => {})
+      }).to.throw('webContents.print(): Invalid deviceName provided.')
+    })
+
     it('does not crash', () => {
       const w = new BrowserWindow({ show: false })
```
deepak1556: Remove this, as its taken care in `beforeEach`

docs/api/webview-tag.md
codebytere: yes, see PR body above:
> Also fixes missing documentation for <webview>.print().

deepak1556: Sorry missed that, its better to pull that into separate PR. The release notes and PR title mislead.

deepak1556: Since docs are using facing change, its better to do it in a separate PR for visibility. Can you please make this change. Thanks!

codebytere: will do!

codebytere: done!

PR #21949, opened 2020-01-28
test: check actual output generated from printToPDF

codebytere:
#### Description of Change

In the interest of better testing our printing code, we should be testing some of the actual substance of the output of `printToPDF`, not just that it throws correct errors for incorrect parameters.

This PR adds one new dependency in `spec-main`, `pdfjs-dist`, which is Mozilla's PDF library in order to test that parameters passed to `contents.printToPDF()` take the desired effect. This will provide a far better alarm for potential upstream breakages to functionality in `contents.printToPDF()`.

cc @ckerr @nornagon @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/spec-main/api-deprecate-spec.ts
+++ b/spec-main/api-deprecate-spec.ts
@@ -2,9 +2,16 @@ import { expect } from 'chai'
 import { deprecate } from 'electron'
 
 describe('deprecate', () => {
+  let throwing: boolean
+
   beforeEach(() => {
+    throwing = process.throwDeprecation
     deprecate.setHandler(null)
     process.throwDeprecation = true
+  })
+
+  afterEach(() => {
+    process.throwDeprecation = throwing
   })
 
   it('allows a deprecation handler function to be specified', () => {
--- a/spec-main/api-web-contents-spec.ts
+++ b/spec-main/api-web-contents-spec.ts
@@ -9,6 +9,7 @@ import { emittedOnce } from './events-helpers'
 import { closeAllWindows } from './window-helpers'
 import { ifdescribe, ifit } from './spec-helpers'
 
+const pdfjs = require('pdfjs-dist')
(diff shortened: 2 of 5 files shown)


codebytere:
Fixed test failures so this should be good to go 🚀

spec-main/api-deprecate-spec.ts
```diff
@@ -7,6 +7,10 @@ describe('deprecate', () => {
     process.throwDeprecation = true
   })
 
+  after(() => {
```
codebytere: The test was failing as a side effect of this not being properly reset after the suite was run.

ckerr: Fantastic find. :rocket: I've no idea how you found this  

nit 1: since `process.throwDeprecation = true;` is set in `beforeEach()`, it would be slightly more correct to reset it in an `afterEach()`.

nit 2:  it would be more futureproof to store the current value in `beforeEach()` and restore _that_ value in `afterEach()` rather than unconditionally setting it to `false`

spec-main/api-web-contents-spec.ts
```diff
@@ -1447,6 +1448,29 @@ describe('webContents module', () => {
...
+
+      const doc = await pdfjs.getDocument(data).promise
+
+      // Check that correct # of pages are rendered.
+      expect(doc.numPages).to.equal(3)
+
+      // Check that PDF is generated in landscape mode.
+      const firstPage = await doc.getPage(1)
+      const { width, height } = firstPage.getViewport({ scale: 100 })
+      expect(width).to.be.greaterThan(height)
```
ckerr: This is great

spec-main/api-deprecate-spec.ts
codebytere: necessary props to sam for assisting me on this find :D

PR #21950, opened 2020-01-29
fix: compilation of native modules on windows with older msvc versions

deepak1556:
#### Description of Change

https://chromium-review.googlesource.com/c/v8/v8/+/1847361 started using C++ `[[deprecated]]` attribute which has issues when building with msvc from VS2015. Native modules compiling with this version will have error of the format ` error C2416: attribute 'deprecated("..")' cannot be applied in this context` from v8 headers

This can be verified with the following test case:

```cpp
class A {
  public:
    [[deprecated("test")]]
    A(int a);
};

int main() {
  return 0;
}
```

with 2015 toolchain there will be error:

```ps
>  & 'C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\amd64\cl.exe' /c /I"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include" test.cc
```

with 2017 toolchain it will compile:

```ps
> & 'C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\bin\Hostx64\x64\cl.exe' /c /I"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include" test.cc
```

There is a msvc bug report out there which fixed this specific case, don't have the link. The PR just uses `__declspec(deprecated)` for those versions.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: fix compilation error for native modules building with VS 2015


patches/v8/fix_build_deprecated_attirbute_for_older_msvc_versions.patch
```diff
@@ -0,0 +1,43 @@
...
++++ b/include/v8config.h
+@@ -388,10 +388,13 @@
+ # define V8_NOINLINE /* NOT SUPPORTED */
+ #endif
+ 
+-
+ // A macro (V8_DEPRECATED) to mark classes or functions as deprecated.
+ #if defined(V8_DEPRECATION_WARNINGS)
+-# define V8_DEPRECATED(message) [[deprecated(message)]]
++# if defined(_MSC_VER) && _MSC_VER <= 1900
```
deepak1556: The version number is extracted from https://devblogs.microsoft.com/cppblog/side-by-side-minor-version-msvc-toolsets-in-visual-studio-2017/

deepak1556:
Yeah its better to upstream this, will do it tomorrow.

MarshallOfSound:
@deepak1556 Can you take this back to 8?

deepak1556:
Upstream CL https://chromium-review.googlesource.com/c/v8/v8/+/2032196

deepak1556:
Please see the above CL and the v8 team has responded with their supported release lines wrt msvc, so its better we remove this patch for Electron 9 and above. Let users fix this properly on their end using VS2017 build tools.

PR #21956, opened 2020-01-29
fix: default printer if none is provided

codebytere:
#### Description of Change

Related to https://github.com/electron/electron/pull/21946, but distinct enough a unit of work that i've chosen to make this its own PR. This removes a section of `printing.patch` by choosing the system default if no printer is passed by the user. This is what happened previously at the chromium level, but as a result of not setting the `kSettingDeviceName` key; we can instead take control of this ourselves.

cc @deepak1556 @ckerr 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/atom_api_web_contents.cc
+++ b/shell/browser/api/atom_api_web_contents.cc
@@ -1801,10 +1801,14 @@ void WebContents::Print(gin_helper::Arguments* args) {
   options.Get("landscape", &landscape);
   settings.SetBoolKey(printing::kSettingLandscape, landscape);
 
-  // We set the default to empty string here and only update
-  // if at the Chromium level if it's non-empty
+  // We set the default to the system's default printer and only update
+  // if at the Chromium level if the user overrides.
+  auto print_backend = printing::PrintBackend::CreateInstance(
+      nullptr, g_browser_process->GetApplicationLocale());
+  std::string default_printer = print_backend->GetDefaultPrinterName();
+  base::string16 device_name = base::UTF8ToUTF16(default_printer);
+
   // Printer device name as opened by the OS.
-  base::string16 device_name;
   options.Get("deviceName", &device_name);
   if (!device_name.empty() && !IsDeviceNameValid(device_name)) {
     args->ThrowError("webContents.print(): Invalid deviceName provided.");
--- a/shell/browser/api/atom_api_web_contents.cc
+++ b/shell/browser/api/atom_api_web_contents.cc
@@ -113,15 +113,6 @@
 #include "ui/gfx/font_render_params.h"
 #endif
 
-#if BUILDFLAG(ENABLE_PRINTING)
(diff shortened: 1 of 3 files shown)


shell/browser/api/atom_api_web_contents.cc
```diff
@@ -1778,10 +1778,14 @@ void WebContents::Print(gin_helper::Arguments* args) {
   options.Get("landscape", &landscape);
   settings.SetBoolKey(printing::kSettingLandscape, landscape);
 
-  // We set the default to empty string here and only update
-  // if at the Chromium level if it's non-empty
+  // We set the default to the system's default printer and only update
+  // if at the Chromium level if the user overrides.
+  auto print_backend = printing::PrintBackend::CreateInstance(
+      nullptr, g_browser_process->GetApplicationLocale());
+  std::string default_printer = print_backend->GetDefaultPrinterName();
```
deepak1556: This is a network operation, can you make it async with `base::PostTaskAndReplyWithResult` , check `LocalPrinterHandlerDefault::GetDefaultPrinter` for example.

shell/browser/api/atom_api_web_contents.cc
```diff
@@ -1868,8 +1879,23 @@ void WebContents::Print(gin_helper::Arguments* args) {
...
+              printing::CompletionCallback print_callback,
+              base::string16 default_printer) {
+            if (printer_name->empty())
+              *printer_name = default_printer;
+            print_settings.SetStringKey(printing::kSettingDeviceName,
+                                        *printer_name);
+            print_view_manager->PrintNow(rfh, silent, std::move(print_settings),
+                                         std::move(print_callback));
+          },
+          &device_name, std::move(settings), std::move(callback)));
```
ckerr: with `&device_name` here it looks like we're passing in the address of a stack local that's going out-of-scope by the time this lambda is invoked?

shell/browser/api/atom_api_web_contents.cc
```diff
@@ -1737,6 +1739,27 @@ bool WebContents::IsCurrentlyAudible() {
...
+    printing::CompletionCallback print_callback,
+    base::string16 device_name,
+    bool silent,
+    base::string16 default_printer) {
+  base::string16 printer_name =
+      device_name.empty() ? default_printer : device_name;
+  print_settings.SetStringKey(printing::kSettingDeviceName, printer_name);
+
+  auto* print_view_manager =
+      printing::PrintViewManagerBasic::FromWebContents(web_contents());
```
deepak1556: Continuing discussion about https://github.com/electron/electron/pull/21986#issuecomment-580972199 , the crash is caused by this line.

This function is registered as a weak callback to the task runner below, so that its not invoked when the `api::WebContents` is not available. Based on the crash it seems `content::WebContents` is not available in the callback, @zcbenz is it possible for `api::WebContents` to be alive while the `content::WebContents` is not ?

PR #21957, opened 2020-01-29
docs: windows don't need to be retained

nornagon:
#### Description of Change
BrowserWindows retain references to themselves at least until they're closed, so this isn't an issue for windows.

It remains an issue for Tray, at least for now, though I think that's something we should address if possible.

#### Checklist

#### Release Notes

Notes: none


aabbccsmith:
As an end user of Electron, does this mean we can avoid `let mainWindow;` for all platforms? We don't need to retain it for all platforms (Mac and windows).

nornagon:
Correct; this change is not platform-specific.

PR #21959, opened 2020-01-29
fix: use the new MediaPlayPause key listener for internal chrome logic

MarshallOfSound:
This stops a previous fix (#21442) from regressing https://bugs.chromium.org/p/chromium/issues/detail?id=930180

This patch should be upstreamable in a similar form in a way that will make our patch a single line

Notes: fixed issue with when playing videos or audio Electron would ask for accessibility access


patches/chromium/fix_use_the_new_mediaplaypause_key_listener_for_internal_chrome.patch
```diff
@@ -0,0 +1,62 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <sattard@slack-corp.com>
+Date: Wed, 29 Jan 2020 12:28:48 -0800
+Subject: fix: use the new MediaPlayPause key listener for internal chrome
+ logic
+
```
codebytere: needs description and upstream plans

patches/chromium/command-ismediakey.patch
```diff
@@ -112,18 +112,3 @@ index 85378bb565de617b1bd611d28c8714361747a357..94a899e76586d2c7bb199828bfa4aa1e
      return event;
    }
  
-@@ -223,12 +233,14 @@ static CGEventRef EventTapCallback(CGEventTapProxy proxy,
-   // For Mac OS 10.12.2 or later, we want to use MPRemoteCommandCenter for
-   // getting media keys globally if there is a RemoteCommandCenterDelegate
-   // available.
-+#if !BUILDFLAG(ENABLE_MEDIA_KEY_OVERRIDES)
```
deepak1556: Can you also remove this buildflag related changes.

MarshallOfSound: Done

patches/chromium/fix_use_the_new_mediaplaypause_key_listener_for_internal_chrome.patch
MarshallOfSound: Done

Issue #21970, opened 2020-01-30
Request for the ability to control app activation policy on macOS


MauricioPellizzon:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description

app.dock.hide() crash on OS X Catalina when running as root #19896

### Proposed Solution

In macOS Catalina the dock.hide() will cause a crash when activated programmatically by root.
The same function works well on 10.14 and earlier versions.

Could we have an implementation as mentioned in  #19896, please?

```
// The application does not appear in the Dock and does not have a menu
// bar, but it may be activated programmatically or by clicking on one
// of its windows.
[NSApp setActivationPolicy: NSApplicationActivationPolicyAccessory];
```

### Additional Information

In my case the Electron app will be open via root from a package installer or LaunchDaemon, as subsequent processes will need to be run as root.
End users do not have admin rights and therefore any UI to request root access will not work.
As well as not showing in the Dock, the app will not have a menu bar too.

Thank you for considering this request.

Issue #21810, opened 2020-01-17
app.dock.hide() && app.dock.show() called in quick succession can duplicate icons


Slapbox:
**Edit: I'm guessing this is known and just unavoidable without some extra working around using the resolution of the promise returned from `dock.show()`?**

### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** `7.1.9`
* **Operating System:** `macOS 10.13.6`
* **Last Known Working Electron version:** `Unk

### Expected Behavior
Duplicate app icons shouldn't be possible.

### Actual Behavior
Duplicate icons can accumulate. After quitting the application the extra icons remain and cannot be quit or force quit to get rid of them. Oddly, simply left-clicking them does get them to go away.

### To Reproduce
```js
  remote.app.dock.show();
  setTimeout(remote.app.dock.hide, 50);
  setTimeout(remote.app.dock.show, 100);
  setTimeout(remote.app.dock.hide, 150);
  setTimeout(remote.app.dock.show, 200);
  setTimeout(remote.app.dock.hide, 250);
```
> Note: `remote` not required, just easier

### Additional Information

If a user tries to force quit a zombie icon, macOS will prevent re-opening the app without the user acknowledging that it crashed last time, as it would if you actually force quit the app.

This should be easy enough for users to workaround, but I assume this isn't the intended behavior.

Edit: Simply checking for `!dock.isVisible()` before calling `dock.show()` is insufficient to prevent this behavior. The bug persists

The real world situation for this is an application that hides/shows with a global hotkey - a user may hit the hotkey by accident and then again to hide it, but they may end up with duplicated dock icons because of this quick action. I toggle back and forth 3x in the repro code above to ensure reproduction, but it can be reproduced with less.

![image](https://user-images.githubusercontent.com/6835891/72575960-398fcb00-389c-11ea-8a16-9f06309965b3.png)

codebytere:
Unfortunately this is a `wontFix` on our end - not because we don't wish to but because for consumer apps, Apple as a company allows very little control over the Dock, and the way that an app exists on it. Apps can only set, by normal means, one of a few activation policies:
```
[...]

Issue #19896, opened 2019-08-22
app.dock.hide() crash on OS X Catalina when running as root


nicko2n:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
6.0.3. Also observed on 4.2.9
* **Operating System:**
OSX Catalina 10.15 Beta (19A536g)
* **Last Known Working Electron version:**
Can't reproduce on non-Catalina variants of OSX - at least Yosemite and High Sierra
### Expected Behavior
App icon not shown in Dock
### Actual Behavior
Error message followed by Electron app crash

2019-08-22 13:18:37.765 Electron[73668:2759878] -[__NSCFConstantString objectAtIndex:]: 
unrecognized selector sent to instance 0x7fff9812d5c0

### To Reproduce

App crashes when calling app.dock.hide() but only when running as root, e.g. started as a service by launchd.

package.json:
{
  "name": "electron-dock",
  "version": "1.0.0",
  "main": "main.js",
  "scripts": {
    "start": "electron ."
  },
  "devDependencies": {
    "electron": "^6.0.3"
  }
}

main.js:
const {app, BrowserWindow} = require('electron')
app.dock.hide()
console.log('reached')
process.exit(0)

As non-privileged user:
$ npm start

> electron-dock@1.0.0 start /path/to/app
> electron .
reached

As root, via sudo
$ sudo npm start

> electron-dock@1.0.0 start path/to/app
> electron .
reached

As root, via su - (or via launchd)

$ sudo su -
root# npm start

> electron-dock@1.0.0 start /path/to/app
> electron .

2019-08-22 13:23:11.096 Electron[73780:2762770] -[__NSCFConstantString objectAtIndex:]: unrecognized selector sent to instance 0x7fff9812d5c0

### Additional Information
My app crashes with "Illegal instruction: 4" after printing the error message but the example above doesn't; not sure what the difference is.

Unsure why it runs ok via "sudo npm start", but not after "sudo su -" . It doesn't appear to be related to differenced in environment variables.

Reproduced on two OS X laptops, but on Catalina only (latest beta as of submission).

[...]

Issue #16093, opened 2018-12-16
app.dock.hide() and app.hide() cannot be used together


cagdas001:
* Output of `node_modules/.bin/electron --version`: v3.0.8
 
* Operating System (Platform and Version): 

  * macOS High Sierra 10.12 (**no issue**)
  * macOS High Sierra 10.13.6 (**issue exists**)
  * macOS Mojave 10.14.1 (**issue exists**)

**Expected Behavior**
**app.hide()** should properly hide the app even if **app.dock.hide()** is being used.

**Actual behavior**
After hiding the dock icon with **app.dock.hide()**, **app.hide()** cannot hide the app itself. 

**To Reproduce**
You just need to use **app.hide()** after **app.dock.hide().** I prepared a minimal example for you to reproduce the issue.

1. Clone my repo.
```sh
git clone --single-branch -b master https://github.com/cagdas001/electron-quick-start.git
```

2. Go to the root directory of project and install dependencies/modules.
```sh
cd electron-quick-start/
npm install
```

3. Start the electron app.
```sh
npm start
```

4. Click on the **app.dock.hide** button and observe, it will work properly.
5. Click on the **app.hide** button and observe, it won't work.

You can show the app again by pressing CommandOrCtrl + X, if you need.

**P.S.** Reversing the order (calling **app.hide()** first, and then**app.dock.hide()**) will work only once. After showing/unhiding the app again, it won't work.

**Additional Information**

**What is wrong with BrowserWindow.hide?**
**BrowserWindow.hide** function can be used on Windows, the previously active window automatically regains  the focus after hiding the window. But this is not the case on macOS.

**Technical Investigation**

I have investigated this issue and it seems that, calling the **app.dock.hide()** sets **canHide** flag to NO. 

See

https://github.com/electron/electron/blob/master/atom/browser/browser_mac.mm#L326

```cpp
void Browser::DockHide() {
  for (auto* const& window : WindowList::GetWindows())
    [window->GetNativeWindow() setCanHide:NO];

  ProcessSerialNumber psn = {0, kCurrentProcess};
  TransformProcessType(&psn, kProcessTransformToUIElementApplication);
}
```

I guess this is the main reason. Is it necessary to set this flag?

This code wasn't enough for hiding dock icon?
```cpp
TransformProcessType(&psn, kProcessTransformToUIElementApplication);
```

[...]

PR #21983, opened 2020-01-31
feat: allow macOS apps to set activation policies

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/21970.

Right now, we allow users access to some dock functionality by hooking into pretty low-level components, but that can cause bugs as evidenced [here](https://github.com/electron/electron/issues/19896), [here](https://github.com/electron/electron/issues/16093), and [here](https://github.com/electron/electron/issues/21810). For apps that want to set more clear and understandable app activation policies, this new small function on `app`, `setActivationPolicy`, should allow them to do that without adding much API maintenance or surface area for bugs.

Alternatives considered:
* Apps can set the [`LSUIElement`](https://developer.apple.com/documentation/bundleresources/information_property_list/lsuielement?language=objc) value to indicate that their app is an agent app that runs in the background, but that's not intuitive by comparison

cc @zcbenz @MarshallOfSound @ckerr 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added the ability to set app activation policy on `macOS`.

### DIFF
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -994,6 +994,17 @@ Updates the current activity if its type matches `type`, merging the entries fro
 
 Changes the [Application User Model ID][app-user-model-id] to `id`.
 
+### `app.setActivationPolicy(policy)` _macOS_
+
+* `policy` String - Can be 'regular', 'accessory', or 'prohibited'.
+
+Sets the activation policy for a given app.
+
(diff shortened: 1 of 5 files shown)


codebytere:
Linux failures are unrelated.

PR #21985, opened 2020-01-31
fix: return path from netLog.stopLogging

nornagon:
#### Description of Change
The docs say we return a string, so let's do that. This was broken by the refactor in #18289.

It's a bit weird that we save the path on behalf of the user—all we're doing is keeping track of the value they passed in. That should really be the responsibility of the user, not the API, but since that's how it's documented to work, we might as well make it work the way the docs say (and the way it used to work).

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed netLog.stopLogging returning undefined instead of the path to the log.

### DIFF
--- a/lib/browser/api/session.js
+++ b/lib/browser/api/session.js
@@ -36,8 +36,9 @@ NetLog.prototype.startLogging = function (path, ...args) {
 
 const _originalStopLogging = NetLog.prototype.stopLogging
 NetLog.prototype.stopLogging = function () {
+  const logPath = this._currentlyLoggingPath
   this._currentlyLoggingPath = null
-  return _originalStopLogging.call(this)
+  return _originalStopLogging.call(this).then(() => logPath)
 }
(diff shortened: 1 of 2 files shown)


lib/browser/api/session.js
```diff
@@ -36,8 +36,9 @@ NetLog.prototype.startLogging = function (path, ...args) {
 
 const _originalStopLogging = NetLog.prototype.stopLogging
 NetLog.prototype.stopLogging = function () {
+  const logPath = this._currentlyLoggingPath
   this._currentlyLoggingPath = null
-  return _originalStopLogging.call(this)
+  return _originalStopLogging.call(this).then(() => logPath)
```
felixrieseberg: So simple

nornagon:
Throwing on [fast-track] to try to get this in before the 8.0.0 stable

PR #22008, opened 2020-02-03
docs: correct the usage of China mirror

g1eny0ung:
#### Description of Change

Change the mirror URL to the newest.

And also, because of the electron version dirs in Taobao mirror doesn't with a **`v`** prefix. Refer to the handling method of `electronic/get`:

https://github.com/electron/get/blob/ba2119f4b15e33f64f11b054add51d8491d1264a/src/utils.ts#L32

```ts
export function normalizeVersion(version: string) {
  if (!version.startsWith('v')) {
    return `v${version}`;
  }
  return version;
}
```

This will make `electron/get` can't find the corresponding download link. So, it's needed to specific ELECTRON_CUSTOM_DIR to which version you want to use.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Correct the usage of the China mirror.

### DIFF
--- a/docs/tutorial/installation.md
+++ b/docs/tutorial/installation.md
@@ -66,7 +66,8 @@ url = ELECTRON_MIRROR + ELECTRON_CUSTOM_DIR + '/' + ELECTRON_CUSTOM_FILENAME
 For instance, to use the China mirror:
 
 ```plaintext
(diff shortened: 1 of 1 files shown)


g1eny0ung:
@malept @MarshallOfSound I just tested it again. Works now......Maybe it's just my network problem. Thank you, guys. Close now.

PR #22013, opened 2020-02-03
chore: wrap cb with default invocation

codebytere:
#### Description of Change

In Chrome's [own](https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/printing/print_preview_message_handler.cc;l=262-267?q=print_preview_message&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2Fsearch%2F) print preview message handling, they wrap their pdf composition callback in a Mojo default invoker, which is designed to be run on cases where the connection is dropped or similar. This ensures that the callback is not dropped on the floor and is run so the function chain can complete. We should do the same.

cc @deepak1556 @nornagon 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/printing/print_preview_message_handler.cc
+++ b/shell/browser/printing/print_preview_message_handler.cc
@@ -23,6 +23,7 @@
 #include "content/public/browser/browser_thread.h"
 #include "content/public/browser/render_frame_host.h"
 #include "content/public/browser/web_contents.h"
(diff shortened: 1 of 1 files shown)


codebytere:
Failures are Chromedriver and mksnapshot and thus unrelated to this PR; merging.

PR #22033, opened 2020-02-04
build: don't use component_ffmpeg for chromedriver

jkleinsc:
#### Description of Change

Chromedriver builds were failing on cross arch compiles because they were trying to bring in the wrong ffmpeg component.  We really don't need/use ffmpeg for chromedriver but due to how it is setup in chromium, it gets pulled in.  This PR changes to use the non component version of ffmpeg for the chromedriver build.
#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


zcbenz:
The failing tests are not related to this PR.

PR #22041, opened 2020-02-05
fix: check WebContents in OnGetDefaultPrinter

zcbenz:
#### Description of Change

This should be able to fix the crash on CI when running printer tests.

#### Release Notes

Notes: no-notes


jkleinsc:

PR #22054, opened 2020-02-05
refactor: saner blink_initialization_order.patch

nornagon:
#### Description of Change
This refactors our patch to `LocalWindowProxy::Initialize`, so that instead of changing when `DidCreateScriptContext` is called, we add a new callback called (somewhat clumsily) `DidInstallConditionalFeatures`.

The main effect of this is that some Chromium code which uses `DidCreateScriptContext` to set up flags that `InstallConditionalFeatures` reads will function correctly. In particular, the flags for whether to expose Mojo primitives to the renderer are set up in `DidCreateScriptContext`, which is relevant for the PDF viewer extension.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


nornagon:
> Probability of upstreaming this hook?

Unlikely since it doesn't seem like it would be useful anywhere in Chrome :\

castroCrea:
Hello with the Beta-3 i can see the PDF but I don't see the zoom and title like in the picture.
Is that just me ?
Thanks
Capture d’écran 2020-02-26 à 20 35 48

castroCrea:
I found out just add  `#page=1&zoom=300` at the end of your data property like here
Capture d’écran 2020-02-26 à 20 43 38

PR #22056, opened 2020-02-05
fix: use a WeakPtr so we do not UAF the store in FunctionLifetimeMonitor

MarshallOfSound:
We already did this for the ObjectLifeMonitor, should also do it for the FunctionLifetimeMonitor

Notes: Fixed issue where renderers could crash during GC when using the `contextBridge` module

### DIFF
--- a/shell/renderer/api/electron_api_context_bridge.cc
+++ b/shell/renderer/api/electron_api_context_bridge.cc
@@ -111,25 +111,31 @@ bool IsPlainArray(const v8::Local<v8::Value>& arr) {
 
 class FunctionLifeMonitor final : public ObjectLifeMonitor {
  public:
-  static void BindTo(v8::Isolate* isolate,
-                     v8::Local<v8::Object> target,
-                     context_bridge::RenderFramePersistenceStore* store,
-                     size_t func_id) {
+  static void BindTo(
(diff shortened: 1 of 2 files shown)


shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -111,25 +111,31 @@ bool IsPlainArray(const v8::Local<v8::Value>& arr) {
...
-  void RunDestructor() override { store_->functions().erase(func_id_); }
+  void RunDestructor() override {
+    if (!store_)
+      return;
+    store_->functions().erase(func_id_);
+  }
 
  private:
-  context_bridge::RenderFramePersistenceStore* store_;
+  base::WeakPtr<context_bridge::RenderFramePersistenceStore> store_;
```
nornagon: Would it be reasonable to eagerly delete the FLMs when the store is deleted? rather than leaving them hanging around until they happen to wake up and notice they're useless now since their store is gone

MarshallOfSound: @nornagon They will all be deleted at the same time (give or take some milliseconds).  The store is only ripped out when the context is destroyed.  The function monitors will be ripped apart during the final GC run

Issue #20744, opened 2019-10-25
Flash SWF content not correctly loaded in Electron 7


frankieta:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.0.0
* **Operating System:**
  * Windows 10 (1803)
* **Last Known Working Electron version:**
  * 6.1.2

### Expected Behavior
Loading flash content through an swf file results in the correct visualization of said content

### Actual Behavior
An error "GET http://interactive-mirror3.nfb.ca/FrameLoader.swf net::ERR_INVALID_ARGUMENT" appears:

![Annotazione 2019-10-25 170420](https://user-images.githubusercontent.com/6009920/67582177-8a829e00-f749-11e9-8e0d-1d08c4930084.png)

### To Reproduce
Try embedding this flash website in an electron project (http://bear71.nfb.ca/#/bear71) following this tutorial: https://electronjs.org/docs/tutorial/using-pepper-flash-plugin

### Screenshots
![Annotazione 2019-10-25 170420](https://user-images.githubusercontent.com/6009920/67582177-8a829e00-f749-11e9-8e0d-1d08c4930084.png)

GhostlyDark:
Introduced with v7.0.0-beta.4

georgelesica-wf:
This also seems to manifest on Mac.

ncodeyx:
Confirmed on Windows and Mac.

misbeliever:
Same in 7.1.2 :(

eric-zhangqi:
some in 7.1.2  not in 6.0.7

misbeliever:
Same in 8.0.0-beta.3

lovoror:
Same in 8.0.0-beta.3 and  7.1.2

kkannan-carecloud:
Anyone here know which version of electron supports flash in MacOS ?

Rosslington:
This is also still broken in 7.1.3

kkannan-carecloud:
If someone here is looking for the last version known to work with flash, it will be 6.1.2

GhostlyDark:
[...]

PR #22072, opened 2020-02-06
fix: flash plugin

t57ser:
#### Description of Change
Close https://github.com/electron/electron/issues/20744

#### Checklist
- [x] `npm test` passes

#### Release Notes

Notes: Fix flash plugin not working

### DIFF
--- a/shell/browser/net/system_network_context_manager.cc
+++ b/shell/browser/net/system_network_context_manager.cc
@@ -10,7 +10,9 @@
 #include "base/command_line.h"
 #include "chrome/browser/browser_process.h"
 #include "chrome/browser/net/chrome_mojo_proxy_resolver_factory.h"
+#include "components/variations/net/variations_http_headers.h"
 #include "content/public/browser/browser_thread.h"
+#include "content/public/browser/cors_exempt_headers.h"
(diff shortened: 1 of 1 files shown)


t57ser:
Note: This should also be backported to version 8 and 7

andreieftimie:
Awesome!

fp07:
Review status?

shell/browser/net/system_network_context_manager.cc
```diff
@@ -154,6 +156,9 @@ SystemNetworkContextManager::CreateDefaultNetworkContextParams() {
   network::mojom::NetworkContextParamsPtr network_context_params =
       network::mojom::NetworkContextParams::New();
 
+  content::UpdateCorsExemptHeader(network_context_params.get());
+  variations::UpdateCorsExemptHeaderForVariations(network_context_params.get());
```
deepak1556: `variations::UpdateCorsExemptHeaderForVariations` is not needed, as chrome finch experiment is not supported by Electron.

Also can you add a comment that

`This is required to avoid blocking X-Requested-With headers sent by PPAPI plugins`

t57ser: All done

PR #22083, opened 2020-02-07
refactor: remove use of Node's DebugOptions

codebytere:
#### Description of Change

Refs https://github.com/electron/electron/pull/21110.

Now that we're processing args with `ProcessGlobalArgs`, DebugOptions will already have been set. As a result, we can now just pull the ones from there to pass to the `InspectorAgent`. This enables us to remove some redundant patches.

cc @zcbenz @nornagon @deepak1556 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


shell/common/node_bindings.cc
```diff
@@ -164,21 +164,23 @@ void SetNodeCliFlags() {
     const auto& option = arg;
 #endif
     const auto stripped = base::StringPiece(option).substr(0, option.find('='));
-    if (allowed.count(stripped) != 0)
+
+    // Only allow in no-op (--) option or DebugOptions
+    if (allowed.count(stripped) != 0 || stripped == "--")
```
deepak1556: @codebytere why was this check for no-op separator become explicit in this PR ? Having this is good, just curious about the cause.

codebytere: @deepak1556 previous to this we were parsing the `DebugOptions` separately, so while we were stripping them in this function, it wasn't the results of this function that DebugOptions was seeing. This would have been wrong before, but we just weren't seeing the consequences of it until we used it for its intended purpose in this PR.

PR #22094, opened 2020-02-07
build: preserve timestamps when stripping files

jkleinsc:
#### Description of Change

In investigating issues with our release builds I discovered that mksnapshot.zip on x64-linux was rather large (~1.4gb) because the executables in it were not getting stripped.  Actually they were getting stripped but doing so caused the mksnapshot zip target to think that the binaries needed to be recompiled.  This PR resolves that issue by preserving the timestamp on the stripped files.

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/strip-binaries.py
+++ b/script/strip-binaries.py
@@ -22,7 +22,7 @@ def strip_binary(binary_path, target_cpu):
     strip = 'mips64el-redhat-linux-strip'
   else:
     strip = 'strip'
-  execute([strip, '-p', binary_path])
+  execute([strip, '--preserve-dates', binary_path])
 
 def main():
   args = parse_args()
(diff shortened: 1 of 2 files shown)


script/strip-binaries.py
```diff
@@ -22,7 +22,7 @@ def strip_binary(binary_path, target_cpu):
     strip = 'mips64el-redhat-linux-strip'
   else:
     strip = 'strip'
-  execute([strip, binary_path])
+  execute([strip, '-p', binary_path])
```
nornagon: I generally prefer to use long-form flags in scripts to reduce the mystery somewhat.
```suggestion
  execute([strip, '--preserve-dates', binary_path])
```

jkleinsc:

PR #22107, opened 2020-02-08
chore: check if file exists before unlinkSync

lutzroeder:
@MarshallOfSound @ckerr @codebytere @nornagon @miniak

notes: no-notes


lutzroeder:
It will break in the debugger each time without need. Startup and shutdown of a basic Electron shouldn’t result in debugger breaks...

lutzroeder:
@nornagon The code still exists in `master`. Are you saying it is no longer used and will be removed?

nornagon:
https://github.com/electron/electron/pull/21814

PR #22121, opened 2020-02-10
refactor: use NSVisualEffectMaterial* constants directly

codebytere:
#### Description of Change

We wre previously static casting these because our minimum SDK was too low to directly use the constants, but now that we're on 10.15 we can use them directly, and so this PR updates them to do so.

cc @ckerr @jkleinsc @zcbenz 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -1455,8 +1455,7 @@ void NativeWindowMac::SetVibrancy(const std::string& type) {
                            relativeTo:nil];
   }
 
-  std::string dep_warn =
-      " has been deprecated and will be removed in a future version of macOS.";
+  std::string dep_warn = " has been deprecated and removed as of macOS 10.15.";
   node::Environment* env =
       node::Environment::GetCurrent(v8::Isolate::GetCurrent());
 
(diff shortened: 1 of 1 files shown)


shell/browser/native_window_mac.mm
```diff
@@ -1455,8 +1455,7 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
                            relativeTo:nil];
   }
 
-  std::string dep_warn =
-      " has been deprecated and will be removed in a future version of macOS.";
+  std::string dep_warn = " has been deprecated and removed as of macOS 10.14.";
```
codebytere: Apple's docs indicate that all below deprecated APIs were removed in 10.15.

shell/browser/native_window_mac.mm
```diff
@@ -1455,8 +1455,7 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
                            relativeTo:nil];
   }
 
-  std::string dep_warn =
-      " has been deprecated and will be removed in a future version of macOS.";
+  std::string dep_warn = " has been deprecated and removed as of macOS 10.15.";
```
codebytere: Apple's docs indicate that all below deprecated APIs were removed in 10.15.

shell/browser/native_window_mac.mm
```diff
@@ -1477,61 +1476,44 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
   }
 
   if (@available(macOS 10.11, *)) {
```
jkleinsc: If we are changing to make the minimum version of the SDK 10.15, do we need this if?

shell/browser/native_window_mac.mm
```diff
@@ -1477,61 +1476,44 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
...
     } else if (type == "ultra-dark") {
-      // NSVisualEffectMaterialUltraDark
       EmitWarning(env, "NSVisualEffectMaterialUltraDark" + dep_warn,
                   "electron");
-      vibrancyType = static_cast<NSVisualEffectMaterial>(9);
+      vibrancyType = NSVisualEffectMaterialUltraDark;
     }
   }
 
   if (@available(macOS 10.14, *)) {
```
jkleinsc: Ditto to above question about min sdk.

shell/browser/native_window_mac.mm
codebytere: @jkleinsc minimum SDK and minimum supported version are not the same: we _support_ 10.10 and beyond but use the SDK from 10.15 :)

shell/browser/native_window_mac.mm
codebytere: Same as above.

PR #22131, opened 2020-02-11
feat: enable pdf viewer

nornagon:
Backport of #21794

Notes: Restored support for pdfium-based PDF viewer.


zcbenz:
Building is failing.

PR #22140, opened 2020-02-11
ci: fix script to initialize Windows VM

thypon:
@nornagon @jkleinsc @ikkisoft 

#### Description of Change

* remove python3 rename statements, python3 is not required nor
installed

* fix a chicken-egg condition where refreshenv is not installed, after
chocolatey install.

#### Checklist

#### Release Notes
Notes: no-notes


thypon:
@jkleinsc do you think we can even remove the vcredist140 from chocolatey if we install Microsoft.VisualStudio.Workload.VCTools ?

thypon:
@jkleinsc I was also wondering if it is possible to avoid installing `windows-sdk-10-version-1903-windbg` without involving `chocolatey`. I find `chocolatey` a nice tool for development but I find difficult to thrust for not established packages (eg. `yarn`, `nodejs` and `python`).

jkleinsc:
@thypon I don't know how else you would get the debug tools needed.  Installing the SDK through VS build tools doesn't do it automatically.

PR #22040, opened 2020-02-05
fix: make webRequest work with WebSocket

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/21308.

This PR replaces https://github.com/electron/electron/pull/21754.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Fix `webRequest` API not working with WebSockets.

PR #22141, opened 2020-02-11
fix: make webRequest work with WebSocket (7-1-x)

zcbenz:
Backport of https://github.com/electron/electron/pull/22040.

Notes: Fix `webRequest` API not working with WebSockets.


codebytere:
@zcbenz this looks like it might need some work before review

zcbenz:
I have fixed the failing tests, this is ready for review.

PR #22146, opened 2020-02-11
feat: add two-finger-swipe webContents event

iliabylich:
#### Description of Change

Added `two-finger-swipe` webContents event.

100% based on Chromium's implementation. These files under `chromium_src/` directory had a bunch of code related to Chromium internal (like bookmarks and other things), so I removed it and kept only what's needed.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added `two-finger-swipe` webContents event

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 [![Electron Logo](https://electronjs.org/images/electron-logo.svg)](https://electronjs.org)
-
+-
 
 [![CircleCI Build Status](https://circleci.com/gh/electron/electron/tree/master.svg?style=shield)](https://circleci.com/gh/electron/electron/tree/master)
 [![AppVeyor Build Status](https://ci.appveyor.com/api/projects/status/4lggi9dpjc1qob7k/branch/master?svg=true)](https://ci.appveyor.com/project/electron-bot/electron-ljo26/branch/master)


iliabylich:
Makes sense.

> let's look for an alternate path here

These files contain a lot of code related to Chromium's internals (bookmarks, drag and drop) that can't be compiled without adding more modules into Electron that will be unused (I'm not even sure if it's possible to include them).

What would you suggest?

deepak1556:
@iliabylich the necessary hooks to implement the two finger swipe are available through https://source.chromium.org/chromium/chromium/src/+/master:content/public/browser/render_widget_host_view_mac_delegate.h 

You have to create `WebContentsViewDelegate` through `ElectronBrowserClient::GetWebContentsViewDelegate` and create the `RenderWidgetHostViewDelegate` from it. Now the only implementation required is the core logic of the two finger swipe. You can eliminate all the chrome related sources.

iliabylich:
@deepak1556 That's exactly what I tried initially.

I have a method `ElectronBrowserClient::GetWebContentsViewDelegate` (well, in this PR it's for `AtomBrowserClient`, initially I implemented it for Electron 7, I'll fix it later) that creates this delegate.

The problem comes from the main class that maintains a state machine and combines multiple low-level events into a single 2-finger swipe event - HistorySwiper.

First of all, it depends on some UI parts of Chromium - https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/renderer_host/chrome_render_widget_host_view_mac_history_swiper.mm;l=7. I had to include a ton of files to progress and finally I gave up.

Also, it's not intended for using anywhere outside Chromium because to perform a navigation it manually invokes `chrome::GoForward/GoBack` - https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/renderer_host/chrome_render_widget_host_view_mac_history_swiper.mm;l=494. What we need is to fire some event that will be caught by observer.

Is it possible to decouple it from Chromium? Or is it worth including more Chromium's modules to build this file?

zcbenz:
@iliabylich If you can move the files from `chromium_src` to `shell`, remove the includes from `chrome/`, and rename classes/filenames/namespaces from `chrome` to `electron`, then I think this PR is mostly ready to go.

Putting files under `chromium_src` implies they have strong dependency on `chrome/` and need lots of maintenance work when doing Chromium upgrades, and this is why this PR gets strong oppositions. However judging from your stripped implementation, I think they have none dependency on `chrome`, and require very little maintenance work, treating them as part of Electron's source tree would be acceptable.

iliabylich:
I was going to rework this PR by turning it into a patch that wraps some places in Chromium files into `if 0/endif`. I have it locally (and it's really small comparing to what we have in this PR), but it's for the older version of Chromium, and so I need to re-apply it on the latest version. I'll try to find some time later this week.

PR #22149, opened 2020-02-11
refactor: use NSVisualEffectMaterial* constants directly

codebytere:
Backport of https://github.com/electron/electron/pull/22121.

Notes: none.


codebytere:
@jkleinsc that shouldn't affect this PR; it doesn't need the 10.15 sdk :)

Issue #22108, opened 2020-02-08
Calling console.log on windows with no arguments causes 'undefined' in stdout


piksel:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.1.10
* **Operating System:**
  *  Windows 10 (1909)
* **Last Known Working Electron version:**
  * v2.0.18 but probably <= v5.x

### Expected Behavior
When using electron on windows and using `console.log()`, an empty row should be written to stdout.

### Actual Behavior
The string 'undefined' is written to stdout.

### To Reproduce
```js
const { app } = require('electron')

console.log('console.log(""):');
console.log("");

console.log('console.log():');
console.log();

app.quit();
```

### Screenshots
![image](https://user-images.githubusercontent.com/807383/74092640-2e107800-4ac7-11ea-81e9-12446c43ada4.png)

### Additional Information
This is likely caused by #16441, when the arguments for the consoleLog method was altered:
https://github.com/electron/electron/pull/16441/files#diff-fd84dda09f0dbd7e74e2822ff6cf19a8R27

The behaviour can be reproduced in node REPL:
```js
const consoleLog = (format, ...args) => require('util').format(format, ...args); 
consoleLog(); // 'undefined'
```

This was discovered when running cypress tests on windows ever since they upgraded their electron version.
Ref: https://github.com/cypress-io/cypress/issues/5995

PR #22165, opened 2020-02-12
fix: no-arg console.log is undefined

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/22108.

`@types/node` incorrectly [types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6a3064cc7408b4e71d0d0feb6f92457d4c22e15c/types/node/util.d.ts#L3) the signature for `util.format()`; it should in fact  per [documentation](https://nodejs.org/api/util.html#util_util_format_format_args) be:

```js
function format(format?: any, ...param: any[]): string;
```

This corrects the error and fixed the undefined behavior. I'll also try to open a PR to the upstream `@types/node` repo.

cc @MarshallOfSound 

#### Checklist

#### Release Notes

Notes: Fixes an issue where `undefined` was printed from `console.log` on Window when no arguments were passed.

### DIFF
--- a/lib/browser/init.ts
+++ b/lib/browser/init.ts
@@ -18,8 +18,10 @@ require('@electron/internal/common/init')
 if (process.platform === 'win32') {
   // Redirect node's console to use our own implementations, since node can not
   // handle console output when running as GUI program.
-  const consoleLog = (format: any, ...args: any[]) => {
-    return process.log(util.format(format, ...args) + '\n')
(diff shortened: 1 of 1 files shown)


ckerr:
@MarshallOfSound I'm :+1: on merging this but am confused why it was done while still a new-pr w/o a fast track?

PR #21951, opened 2020-01-29
feat: [extensions] support extension.getBackgroundPage

nornagon:
#### Description of Change
Builds on top of #21814.

This adds support for `chrome.extension.getBackgroundPage` for synchronous scripting of the background page. Additionally, it fixes loading subresources in extension pages (e.g. when navigating to `chrome://some-extension/page.html`, that page will be able to load scripts with `<script src="some-subresource.js">` and so on), and makes it so that all extension pages live in the same process.

#### Checklist

- [ ] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [ ] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [ ] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Added support for the chrome.extension.getBackgroundPage API when building with enable_electron_extensions.

PR #22177, opened 2020-02-13
feat: [extensions] support extension.getBackgroundPage (#21951)

nornagon:
Backport of #21951.

Notes: Added support for the chrome.extension.getBackgroundPage API.


zcbenz:
Does this need approval from @electron/wg-releases? While it is a `feat` I think it is more about fixing extensions.

ckerr:
Approved for backport in 2020-02-19 wg-releases meeting

jkleinsc:
@nornagon do you mind rebasing this so that we can get it merged into 9-x-y

PR #22187, opened 2020-02-13
fix: crash on custom printing margins

codebytere:
Backport of https://github.com/electron/electron/pull/22164.

See that PR for details.

Notes: Fixed a crash in `webContents.print()` with custom print margins.


zcbenz:
Build is failing due to API changes of `base::Value`.

PR #22188, opened 2020-02-13
fix: emit will-navigate for sandboxed contents

### LINKED ISSUE #8841: `will-navigate` event not fired when in sandbox mode

nornagon:
#### Description of Change
Fixes #8841.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed "will-navigate" event not being emitted for sandboxed contents.

### DIFF
--- a/shell/browser/common_web_contents_delegate.cc
+++ b/shell/browser/common_web_contents_delegate.cc
@@ -294,6 +294,11 @@ content::WebContents* CommonWebContentsDelegate::OpenURLFromTab(
   load_url_params.is_renderer_initiated = params.is_renderer_initiated;
   load_url_params.initiator_origin = params.initiator_origin;
   load_url_params.should_clear_history_list = true;
+  if (params.post_data) {
+    load_url_params.load_type =
+        content::NavigationController::LOAD_TYPE_HTTP_POST;
+    load_url_params.post_data = params.post_data;
+  }
 
   source->GetController().LoadURLWithParams(load_url_params);
   return source;
--- a/shell/browser/common_web_contents_delegate.cc
+++ b/shell/browser/common_web_contents_delegate.cc
@@ -292,8 +292,16 @@ content::WebContents* CommonWebContentsDelegate::OpenURLFromTab(
   load_url_params.should_replace_current_entry =
       params.should_replace_current_entry;
   load_url_params.is_renderer_initiated = params.is_renderer_initiated;
+  load_url_params.started_from_context_menu = params.started_from_context_menu;
   load_url_params.initiator_origin = params.initiator_origin;
-  load_url_params.should_clear_history_list = true;
+  load_url_params.source_site_instance = params.source_site_instance;
+  load_url_params.frame_tree_node_id = params.frame_tree_node_id;
+  load_url_params.redirect_chain = params.redirect_chain;
+  load_url_params.has_user_gesture = params.user_gesture;
+  load_url_params.blob_url_loader_factory = params.blob_url_loader_factory;
+  load_url_params.href_translate = params.href_translate;
+  load_url_params.reload_type = params.reload_type;
+
   if (params.post_data) {
     load_url_params.load_type =
         content::NavigationController::LOAD_TYPE_HTTP_POST;
--- a/spec-main/api-browser-window-spec.ts
+++ b/spec-main/api-browser-window-spec.ts
@@ -2088,8 +2088,7 @@ describe('BrowserWindow module', () => {
             'did-finish-load',
             'did-frame-finish-load',
             'did-navigate-in-page',
-            // TODO(nornagon): sandboxed pages should also emit will-navigate
-            // 'will-navigate',
+            'will-navigate',
             'did-start-loading',
             'did-stop-loading',
             'did-frame-finish-load',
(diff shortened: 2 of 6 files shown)


nornagon:
NB. this approach was previously attempted here https://github.com/electron/electron/pull/20050, but wasn't accepted because of an assertion that `ShouldFork` causes the renderer process to restart. I've tested this code locally and that doesn't seem to be the case—the navigation does not cause a new process to start. Cross-domain navigations still cause a new renderer, as expected, but within a single domain the process is reused.

Here's a test fiddle: https://gist.github.com/f4dfb4cd13e1a433b79a118a94a98965

Try watching `ps` or equivalent process monitor and watch for changing PIDs.

shell/browser/common_web_contents_delegate.cc
```diff
@@ -292,8 +292,21 @@ content::WebContents* CommonWebContentsDelegate::OpenURLFromTab(
   load_url_params.should_replace_current_entry =
       params.should_replace_current_entry;
   load_url_params.is_renderer_initiated = params.is_renderer_initiated;
+  load_url_params.started_from_context_menu = params.started_from_context_menu;
   load_url_params.initiator_origin = params.initiator_origin;
-  load_url_params.should_clear_history_list = true;
```
deepak1556: We maintain our own navigation controller, so its necessary to clear the chromium history list.

deepak1556:
To add some notes on this mechanism:

`ShouldFork` by itself doesn't cause process restart, it gives access to us for all render initiated navigations and lets us decide whether we want it to turn into a browser process initiated navigation so that we can force our process policy through `ShouldOverrideSiteInstanceForNavigation`. There used to be bug with sandbox renderers to use this mechanism but it seems to have improved.

/cc @zcbenz incase I missed a scenario.

spec-main/api-browser-window-spec.ts
```diff
@@ -434,110 +434,128 @@ describe('BrowserWindow module', () => {
...
-
-    describe('will-navigate event', () => {
-      it('allows the window to be closed from the event listener', (done) => {
-        w.webContents.once('will-navigate', () => {
-          w.close()
-          done()
-        })
-        w.loadFile(path.join(fixtures, 'pages', 'will-navigate.html'))
+  for (const sandbox of [false, true]) {
+    describe(`navigation events${sandbox ? ' with sandbox' : ''}`, () => {
```
deepak1556: Can you add a test for `will-navigate` in the case when navigating from non-file scheme to file scheme ?

shell/browser/common_web_contents_delegate.cc
nornagon: The tests aren't failing. How would I detect that this is causing incorrect behavior?

nornagon: FWIW removing this fixed a `DCHECK(!replace)` in the navigation controller.

spec-main/api-browser-window-spec.ts
nornagon: It seems that results in `"Not allowed to load local resource: file://..."`, in both the sandbox and non-sandbox situations, with this patch.

shell/browser/common_web_contents_delegate.cc
deepak1556: Hmm the non-sandboxed renderer history are unaffected by this change, tested with:

* load `file://index.html`
* window.history.length == 1
* window.location.reload() // creates new process
* window.location = "https://example.com" // creates new process
* window.location.reload() // creates new process
* window.history.length == 4 but should be 2 , this bug is not related to this change.

So seems like custom navigation controller still takes control for non-sandboxed renderers.

nornagon: Sounds like we're all OK on this front :)

Issue #17031, opened 2019-02-18
Add ability to get NSUserDefaults item from a specific suite


sindresorhus:
**Is your feature request related to a problem? Please describe.**

Electron has the `systemPreferences.getUserDefault()` method to get a NSUserDefaults key in the [standard suite](https://developer.apple.com/documentation/foundation/nsuserdefaults/1416603-standarduserdefaults?language=objc). However, there's no way to get a key from a different suite.

**Describe the solution you'd like**

I propose adding a `systemPreferences.getUserDefaultFromSuite(suite, key, type)` method to get a key from a given suite.

Internally, this call would look like:

```objective-c
[[[[NSUserDefaults alloc] initWithSuiteName:@"name-of-suite"] objectForKey:@"key"] boolValue];
```

https://developer.apple.com/documentation/foundation/nsuserdefaults/1409957-initwithsuitename?language=objc

**Describe alternatives you've considered**

None.

**Additional context**

My specific use-case is to get the system "Do Not Disturb" status. I would use the new method like this:

```js
systemPreferences.getUserDefaultFromSuite('com.apple.notificationcenterui', 'doNotDisturb', 'boolean');
```

PR #22193, opened 2020-02-14
feat: add APIs for accessing NSUserDefaults in the given domain

miniak:
#### Description of Change
Add APIs requested in #17031:
- `systemPreferences.getUserDefaultInDomain(domain, key, type)`
- `systemPreferences.setUserDefaultInDomain(domain, key, type, value)`
- `systemPreferences.removeUserDefaultInDomain(domain, key)`

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added APIs for accessing `NSUserDefaults` in the given domain on macOS.

### DIFF
--- a/shell/browser/api/electron_api_system_preferences.h
+++ b/shell/browser/api/electron_api_system_preferences.h
@@ -81,13 +81,22 @@ class SystemPreferences : public gin_helper::EventEmitter<SystemPreferences>
   int SubscribeWorkspaceNotification(const std::string& name,
                                      const NotificationCallback& callback);
   void UnsubscribeWorkspaceNotification(int request_id);
-  v8::Local<v8::Value> GetUserDefault(const std::string& name,
+  v8::Local<v8::Value> GetUserDefault(const std::string& key,
                                       const std::string& type);
+  v8::Local<v8::Value> GetUserDefaultInDomain(const std::string& domain,
+                                              const std::string& key,
+                                              const std::string& type);
   void RegisterDefaults(gin_helper::Arguments* args);
-  void SetUserDefault(const std::string& name,
+  void SetUserDefault(const std::string& key,
                       const std::string& type,
                       gin_helper::Arguments* args);
-  void RemoveUserDefault(const std::string& name);
+  void SetUserDefaultInDomain(const std::string& domain,
+                              const std::string& key,
+                              const std::string& type,
+                              gin_helper::Arguments* args);
+  void RemoveUserDefault(const std::string& key);
+  void RemoveUserDefaultInDomain(const std::string& domain,
+                                 const std::string& key);
   bool IsSwipeTrackingFromScrollEventsEnabled();
 
   std::string GetSystemColor(gin_helper::ErrorThrower thrower,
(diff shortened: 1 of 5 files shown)


shell/browser/api/electron_api_system_preferences.h
```diff
@@ -83,6 +83,9 @@ class SystemPreferences : public gin_helper::EventEmitter<SystemPreferences>
   void UnsubscribeWorkspaceNotification(int request_id);
   v8::Local<v8::Value> GetUserDefault(const std::string& name,
                                       const std::string& type);
+  v8::Local<v8::Value> GetUserDefaultFromSuite(const std::string& suite,
+                                               const std::string& name,
```
zarubond: For consistency this should be named key

codebytere:
@miniak if you'd like this to go to 9 can you add it to the `Releases` agenda? I also think this may need some API WG review.

shell/browser/api/electron_api_system_preferences_mac.mm
```diff
@@ -121,6 +121,121 @@ AVMediaType ParseMediaType(const std::string& media_type) {
...
+                        gin_helper::Arguments* args) {
+  const auto throwConversionError = [&] {
+    args->ThrowError("Unable to convert value to: " + type);
+  };
+
+  NSString* key = base::SysUTF8ToNSString(name);
+  if (type == "string") {
+    std::string value;
+    if (!args->GetNext(&value)) {
+      throwConversionError();
```
zcbenz: I think we can refactor the code flow to avoid duplicating the `throwConversionError` call.

```c++
if (type == "string") {
  std::string value;
  if (args->GetNext(&value)) {
    [defaults setObject:base::SysUTF8ToNSString(value) forKey:key];
    return;
  }
} else if (...) {
  ...
} else {
  args->ThrowError("Invalid type: " + type);
  return;
}

args->ThrowError("Unable to convert value to: " + type);
```

miniak: I just moved the function as is with minimal changes to be able to re-use it. I can refactor it in a follow-up PR.

miniak: actually, let's do it now

jkleinsc:
It appears that this PR slipped through the cracks, and for that I apologize @miniak.
This was reviewed at the March 23rd API WG meeting and the working group decided to decline this feature due to the following reasons:

- Not accessible from MAS
- Not possible to subscribe to changes, which might encourage polling
- Can be implemented straightforwardly as a native module.
See: https://github.com/electron/governance/blob/master/wg-api/meeting-notes/2020-03-23.md

miniak:
@jkleinsc thanks for the reply

sindresorhus:
I just want to clear up some inaccuracies:

> Not accessible from MAS

External UserDefaults suites are not available, but someone could have a bundled helper app that they want to share UserDefaults with using [App Groups](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_application-groups). This is common for app extensions. Also, most Electron apps are not on the Mac App Store.

> Not possible to subscribe to changes, which might encourage polling

`UserDefaults.didChangeNotification` does work with different domains, but it does not work across processes. However, you can use KVO on the individual UserDefault keys, which does work across processes.

> Can be implemented straightforwardly as a native module.

Nothing about native modules are straightforward.

PR #22197, opened 2020-02-14
docs: NSPrincipalClass should be ElectronApplication

nornagon:
This was renamed, just updating the docs.

Notes: none


nornagon:
maybe we should revert the rename instead...?

MarshallOfSound:
> maybe we should revert the rename instead...?

Sounds like it's the safest option 👍

nornagon:
I have no idea what this thing does; what would be the expected behavior of this being broken?

MarshallOfSound:
@nornagon Deep links would stop working, i.e. `my-app://`

PR #22198, opened 2020-02-14
chore: update chromium to 82.0.4058.2

nornagon:
Notes: Updated Chromium to 82.0.4058.2


zcbenz:
There is a patch needs to be updated.

spec/fixtures/api/gpu-info.js
```diff
@@ -1,6 +1,7 @@
 const { app } = require('electron')
 
 app.commandLine.appendSwitch('--disable-software-rasterizer')
+app.commandLine.appendSwitch('--no-delay-for-dx12-vulkan-info-collection')
```
deepak1556: This is just for debugging, will remove before review.

patches/chromium/command-ismediakey.patch
```diff
@@ -41,26 +41,17 @@ index c5125495b4d178ffb18be4d2d9670f7556412cbd..cddb321abb938c667a4a2089f87eab99
...
 +  modifiers |= accelerator.IsCmdDown() ? Mod4Mask : 0;
  
    return modifiers;
  }
-@@ -151,6 +152,8 @@ void GlobalShortcutListenerX11::OnXKeyPressEvent(::XEvent* x_event) {
-   modifiers |= (x_event->xkey.state & ShiftMask) ? ui::EF_SHIFT_DOWN : 0;
-   modifiers |= (x_event->xkey.state & ControlMask) ? ui::EF_CONTROL_DOWN : 0;
-   modifiers |= (x_event->xkey.state & Mod1Mask) ? ui::EF_ALT_DOWN : 0;
-+  // For Windows key
-+  modifiers |= (x_event->xkey.state & Mod4Mask) ? ui::EF_COMMAND_DOWN: 0;
```
nornagon: Is it OK to remove this? This seems important.

patches/chromium/disable_color_correct_rendering.patch
```diff
@@ -19,28 +19,11 @@ This can be removed once web content (including WebGL) learn how
...
-+++ b/cc/trees/layer_tree_host_impl.cc
-@@ -1814,6 +1814,12 @@ const gfx::ColorSpace& LayerTreeHostImpl::GetRasterColorSpace() const {
- 
- const gfx::ColorSpace& LayerTreeHostImpl::GetRasterColorSpaceAndId(
-     int* id) const {
-+  if (!settings_.enable_color_correct_rendering) {
-+    static gfx::ColorSpace invalid_color_space;
-+    *id = -1;
-+    return invalid_color_space;
-+  }
```
nornagon: fyi @poiru

patches/node/avoid_calling_deprecated_method.patch
```diff
@@ -0,0 +1,22 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <samuel.r.attard@gmail.com>
+Date: Thu, 13 Feb 2020 10:34:31 -0800
+Subject: Avoid calling deprecated method
+
+The {SetExpectInlineWasm} method is deprecated and non-functional since
+V8 v8.1.
+Thus node should stop calling it, so that it can be fully removed in a
+future v8 version.
```
nornagon: Presumably this has already been PR'd to Node?

patches/chromium/command-ismediakey.patch
deepak1556: Yeah we were handling it specifically because of `XEvent` but now the conversion is done by `ui::accelerator` which will handle the modifier. It was removed in https://github.com/electron/electron/pull/22025/commits/3fb6ab63da83cce10db1f0f52ca89794cd0b3765

CL: https://chromium-review.googlesource.com/c/chromium/src/+/2022280

patches/node/avoid_calling_deprecated_method.patch
deepak1556: This was cherry-pick'd from https://github.com/electron/electron/pull/22064/commits/7ed51ee3ad22d16e530b3c8ea7a15075e2ec3689.  I don't think its upstreamed yet.

/cc @codebytere

deepak1556:
@jkleinsc mksnapshot test fails on x86 even after https://github.com/electron/electron/pull/21991 , can you please take a look. Thanks!

patches/chromium/gpu_notify_when_dxdiag_request_fails.patch
```diff
@@ -0,0 +1,69 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Deepak Mohan <hop2deep@gmail.com>
+Date: Fri, 28 Feb 2020 15:08:26 -0800
+Subject: gpu: notify when dxdiag request for gpu info fails
+
+We rely on the signal OnGpuInfoUpdate to process gpu info.
```
deepak1556: The spec `app module getGPUInfo() API succeeds with complete GPUInfo` failure on windows in this chromium roll revealed two issues:

1) The complete gpu info involves requesting the gpu process for info about DxDiagnostics and Dx12Vulkan specific infos, but our test was not giving enough time to launch the gpu process and thus only obtained basic info, this is fixed in https://github.com/electron/electron/pull/22198/commits/64aa2fae5e011a8413b44a098da582e9112ef911

2) There is a flaw in the current `getGPUInfo` api where it waits for the above mentioned values from the gpu process https://github.com/electron/electron/blob/master/shell/browser/api/gpuinfo_manager.cc#L37 hoping that it will always succeed https://source.chromium.org/chromium/chromium/src/+/master:content/browser/gpu/gpu_data_manager_impl_private.cc;l=837, but there are situations when this may not succeed https://source.chromium.org/chromium/chromium/src/+/master:content/browser/gpu/gpu_data_manager_impl_private.cc;l=630-647 in which case the api just waits endlessly, this patch fixes that case by reusing the failure signals from chromium. 

If the api was instead an event like `app.on('gpu-info-update')` users can continuously perceive the updates from gpu process and we don't have to worry about handling these edge cases similar to `chrome://gpu` page.

PR #22244, opened 2020-02-17
feat(extensions): add more properties to extension object

samuelmaddock:
#### Description of Change
Adds `manifest`, `path`, and `url` properties to the `Extension` structure.

These are converted directly from `extensions/common/extension.h` and serve as shortcuts to prevent the user from needing to construct the data themselves.

Ref #19447

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

Chrome extensions are still not released in a stable build yet so no notes are to be added.

### DIFF
--- a/shell/common/gin_converters/extension_converter.cc
+++ b/shell/common/gin_converters/extension_converter.cc
@@ -6,6 +6,9 @@
 
 #include "extensions/common/extension.h"
 #include "gin/dictionary.h"
+#include "shell/common/gin_converters/file_path_converter.h"
+#include "shell/common/gin_converters/gurl_converter.h"
+#include "shell/common/gin_converters/value_converter.h"
 
 namespace gin {
 
@@ -16,7 +19,11 @@ v8::Local<v8::Value> Converter<const extensions::Extension*>::ToV8(
   auto dict = gin::Dictionary::CreateEmpty(isolate);
   dict.Set("id", extension->id());
   dict.Set("name", extension->name());
+  dict.Set("path", extension->path());
+  dict.Set("url", extension->url());
   dict.Set("version", extension->VersionString());
+  dict.Set("manifest", *(extension->manifest()->value()));
+
   return gin::ConvertToV8(isolate, dict);
 }
 
--- a/docs/api/structures/extension.md
+++ b/docs/api/structures/extension.md
@@ -1,5 +1,8 @@
 # Extension Object
 
 * `id` String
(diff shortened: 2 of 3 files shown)


shell/common/gin_converters/extension_converter.cc
```diff
@@ -4,8 +4,12 @@
 
 #include "shell/common/gin_converters/extension_converter.h"
 
+#include "content/public/renderer/v8_value_converter.h"
```
zcbenz: This header is not needed.

shell/common/gin_converters/extension_converter.cc
```diff
@@ -16,7 +20,13 @@ v8::Local<v8::Value> Converter<const extensions::Extension*>::ToV8(
   auto dict = gin::Dictionary::CreateEmpty(isolate);
   dict.Set("id", extension->id());
   dict.Set("name", extension->name());
+  dict.Set("path", extension->path());
+  dict.Set("url", extension->url());
   dict.Set("version", extension->VersionString());
+
+  auto manifest = extension->manifest()->value()->Clone();
```
zcbenz: There is no need for clone, converting a `base::Value` to V8 already implies cloning.

shell/common/gin_converters/extension_converter.cc
samuelmaddock: oops, thanks for catching this!

shell/common/gin_converters/extension_converter.cc
samuelmaddock: I ended up needing to write another converter template for this. c38dce3

zcbenz: I think you can just do `dict.Set("manifest", *(extension->manifest()->value()));`?

samuelmaddock: Looking more closely at `converter.ToV8Value`, it seems like this performs close to the same operation as `Clone()`. Please let me know which you prefer.

samuelmaddock: > 
> 
> I think you can just do `dict.Set("manifest", *(extension->manifest()->value()));`?

Ah, yes. Fixed in f7a4f5d

sentialx:
Could it be backported to 9.0.0?

samuelmaddock:
> Could it be backported to 9.0.0?

You can open a new PR with the backport if you'd like. Target the `9-x-y` branch and include "Backport of #" in the description. That's what I did for #22570, if you wanted to have a reference.

PR #22245, opened 2020-02-18
fix: load devtools://remote assets so that lighthouse works

### LINKED ISSUE #13008: Launch throwing breaking devtools error
* Electron Version:
* Operating System (Platform and Version): macOS sierra 10.12.6
* Last known working Electron version: 2.0.0
I had upgraded to 2.0.1 previously but the error started after that so I tried downgrading to 2.0.0. The error is still there though..

**Expected Behavior**
Launch my app normally. This was working fine prior to today.

**Actual behavior**
When I launch the app if I click ANYTHING on the screen that is registered as a click event in my renderer main.js file, this error spams the console.

**To Reproduce**
1. Open app
2. Do anything in the app.
3. Console spams

Error - 
`[2287:0518/162444.516714:ERROR:CONSOLE(0)] "Failed to load https://chrome-devtools-frontend.appspot.com/serve_file/@7accc8730b0f99b5e7c0702ea89d1fa7c17bfe33/product_registry_impl/product_registry_impl_module.js: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'chrome-devtools://devtools' is therefore not allowed access. The response had HTTP status code 404.", source: chrome-devtools://devtools/bundled/inspector.html?remoteBase=https://chrome-devtools-frontend.appspot.com/serve_file/@7accc8730b0f99b5e7c0702ea89d1fa7c17bfe33/&can_dock=true&toolbarColor=rgba(223,223,223,1)&textColor=rgba(0,0,0,1)&experiments=true (0)
[2287:0518/162444.516839:ERROR:CONSOLE(22)] "Empty response arrived for script 'https://chrome-devtools-frontend.appspot.com/serve_file/@7accc8730b0f99b5e7c0702ea89d1fa7c17bfe33/product_registry_impl/product_registry_impl_module.js'", source: chrome-devtools://devtools/bundled/inspector.js (22)
`
Then this message spams about 20 times every time I click anything.

`"Uncaught (in promise) Error: Could not instantiate: ProductRegistryImpl.Registry", source: chrome-devtools://devtools/bundled/inspector.js (105)`

I am using devtron as well.

MarshallOfSound:
This is basically some more copy-pasting into our `devtools_ui.cc` impl (which is already a lot of copy-pasting).  In order to avoid duplicating the sanitization logic as that getting out of sync would be a Bad Thing I patched some chrome logic out of `devtools_bindings` and used the sanitization logic directly from there.

This fixes issues folks are seeing with "failed to load devtools://devtools/remote/*" or "empty response arrived for script devtools://devtools/remote/*" messages.

Fixes #13008

We had all the logic to set the remote URL but not the handling for the remote requests

Notes: fixed remote devtools assets not loading so that lighthouse works in devtools


shell/browser/ui/devtools_ui.cc
```diff
@@ -68,13 +99,30 @@ class BundledDataSource : public content::URLDataSource {
...
     }
 
+    std::string remote_path_prefix(kChromeUIDevToolsRemotePath);
+    remote_path_prefix += "/";
+    if (base::StartsWith(path, remote_path_prefix,
+                         base::CompareCase::INSENSITIVE_ASCII)) {
+      GURL url(kRemoteFrontendBase + path.substr(remote_path_prefix.length()));
+
+      CHECK_EQ(url.host(), kRemoteFrontendDomain);
+      if (url.is_valid() && DevToolsUIBindings::IsValidRemoteFrontendURL(url)) {
```
deepak1556: Our `inspectable_web_contents_impl` is based on `devtools_ui_bindings` , the sanitization logic can be copied there. We already have a couple of copied code path for devtools ui actions, so wouldn't hurt to have this one part of it.

MarshallOfSound: @deepak1556 It's a decent chunk of code 🤷‍♂ If we're ok with copying it over I can do that 👍

MarshallOfSound:
@deepak1556 This isn't for the product registry, it's for the lighthouse worker

https://source.chromium.org/chromium/chromium/src/+/master:third_party/devtools-frontend/src/BUILD.gn;l=2165-2169?q=lighthouse_worker%20BUILD.gn&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2F

`lighthouse_worker/lighthouse_worker_module.js` is a script that is required for lighthouse to work and it is one of three scripts that is still "remote" for some reason

MarshallOfSound:
@deepak1556 If you don't want to bake in loading remote hosts, would a patch to make that worker module "bundled" be better for you (Haven't tried but it should be possible)

deepak1556:
> lighthouse_worker/lighthouse_worker_module.js is a script that is required for lighthouse to work and it is one of three scripts that is still "remote" for some reason

Ah thanks for the clarification, looking into the lighthouse_worker code https://source.chromium.org/chromium/chromium/src/+/master:third_party/devtools-frontend/src/front_end/lighthouse_worker/lighthouse/README.chromium seems to be the reason they have kept it as remote.

> would a patch to make that worker module "bundled" be better for you (Haven't tried but it should be possible)

Worth trying, it should definitely be possible, but can you measure the increase in size of `resources.pak' with this change ?

MarshallOfSound:
Closing out for now as abandoned

PR #22270, opened 2020-02-18
feat: Add OtherItemsProxy TouchBar item

erickzhao:
#### Description of Change

Adds a new `TouchBarOtherItemsProxy` class that inherits from `TouchBarItem`. This allows Electron developers to customize the positioning of touch bars inherited from Chromium using an `NSTouchBarItemIdentifierOtherItemsProxy`.

By default, we used to have this proxy enabled at the end of each touch bar instance. It still defaults to that position if not manually defined.

Also, you can't add more than one of these per touch bar. I've added a guard to catch this at the JS level.

See: https://developer.apple.com/documentation/appkit/nstouchbaritemidentifierotheritemsproxy

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: The position of nested touch bars inherited from Chromium can now be customized (macOS).

### DIFF
--- a/lib/browser/api/touch-bar.js
+++ b/lib/browser/api/touch-bar.js
@@ -51,10 +51,22 @@ class TouchBar extends EventEmitter {
         item.child.ordereredItems.forEach(registerItem)
       }
     }
+
+    let hasOtherItemsProxy = false
     items.forEach((item) => {
       if (!(item instanceof TouchBarItem)) {
         throw new Error('Each item must be an instance of TouchBarItem')
       }
+
+      // only allow one OtherItemsProxy per touchbar
+      if (item.type === 'other_items_proxy') {
+        if (!hasOtherItemsProxy) {
+          hasOtherItemsProxy = true
+        } else {
+          throw new Error('Must only have one OtherItemsProxy per TouchBar')
+        }
+      }
+
       this.ordereredItems.push(item)
       registerItem(item)
     })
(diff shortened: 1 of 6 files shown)


spec-main/api-touch-bar-spec.ts
```diff
@@ -47,32 +54,35 @@ describe('TouchBar module', () => {
...
-        }),
-        new TouchBarSegmentedControl({ segments: [] }),
-        new TouchBarScrubber({
-          items: [{ label: 'foo' }, { label: 'bar' }, { label: 'baz' }],
-          selectedStyle: 'outline',
-          mode: 'fixed',
-          showArrowButtons: true
-        })
-      ] })
+      const touchBar = new TouchBar({
```
erickzhao: I added a line of code to the spec, but it seems like formatting on save changed a lot of whitespace.

Let me know if this needs to be reverted.

lib/browser/api/touch-bar.js
```diff
@@ -334,4 +346,12 @@ TouchBar.TouchBarScrubber = class TouchBarScrubber extends TouchBarItem {
   }
 }
 
+TouchBar.TouchBarOtherItemsProxy = class TouchBarOtherItemsProxy extends TouchBarItem {
+  constructor (config) {
+    super()
+    if (config == null) config = {} // no config options anyways
```
MarshallOfSound: This is a noop, we can just ignore `config` altogether here

lib/browser/api/touch-bar.js
```diff
@@ -51,10 +51,22 @@ class TouchBar extends EventEmitter {
...
       if (!(item instanceof TouchBarItem)) {
         throw new Error('Each item must be an instance of TouchBarItem')
       }
+
+      // only allow one OtherItemsProxy per touchbar
+      if (item.type === 'other_items_proxy') {
+        if (!hasOtherItemsProxy) {
+          hasOtherItemsProxy = true
+        } else {
+          throw new Error('Must only have one OtherItemsProxy per TouchBar')
```
MarshallOfSound: We need to ensure we throw this error before we call `registerIteam` (probably in a different loop) otherwise we're going to have dangling event listeners.

Issue #22305, opened 2020-02-20
BrowserWindow.webContents.focus doesn't work on mac


CezaryKulakowski:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
10.0.0-nightly.20200220
* **Operating System:**
macOS 10.15.1
* **Last Known Working Electron version:**
I don't know (judging by implementation it never worked)

### Expected Behavior:
Call to BrowserWindow.webContents.focus focuses proper window

### Actual Behavior:
Call  to BrowserWindow.webContents.focus does nothing

### To Reproduce:

Run following simple app:
```
const { app, BrowserWindow } = require('electron')

function createWindow () {
  let win1 = new BrowserWindow({ x:100, y: 300, width: 300, height: 200 })
  win1.on('closed', () => {
    win1 = null
  })
  win1.loadURL('https://github.com')

  let win2 = new BrowserWindow({ x:300, y: 300, width: 300, height: 200 })
  win2.on('closed', () => {
    win1 = null
  })
  win2.loadURL('http://www.example.com')

  setTimeout(() => {
    win1.webContents.focus();
  }, 3000);
}

app.on('ready', createWindow)
```

It works as expected on Windows: it creates two windows and after three seconds it focuses the first one. On mac it doesn't work and call to `webContents.focus` doesn't change the focus (it stays on the second window).

codebytere:
We require the template to be filled out on all new issues and pull requests. We do this so that we can be certain we have all the information we need to address your submission efficiently. This allows the maintainers to spend more time fixing bugs, implementing enhancements, and reviewing and merging pull requests.

We will be able to more closely look at your issue once you do so. This issue will be closed in 10 days if the above is not addressed.

CezaryKulakowski:
I've create pr for the issue: https://github.com/electron/electron/pull/22323

PR #22323, opened 2020-02-21
fix: call focus on native window on call to webContents.focus on mac

CezaryKulakowski:
Closes https://github.com/electron/electron/issues/22305.

On mac call to web_contents()->Focus() is not enough so it's
necessary to call it on native window.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed `WebContents.focus` not focusing window on macOS and Linux.

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -2070,11 +2070,11 @@ void WebContents::CopyImageAt(int x, int y) {
     host->CopyImageAt(x, y);
 }
 
+#if !defined(OS_MACOSX)
 void WebContents::Focus() {
   web_contents()->Focus();
 }
 
-#if !defined(OS_MACOSX)
 bool WebContents::IsFocused() const {
   auto* view = web_contents()->GetRenderWidgetHostView();
   if (!view)
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -2070,11 +2070,17 @@ void WebContents::CopyImageAt(int x, int y) {
     host->CopyImageAt(x, y);
 }
 
-#if !defined(OS_MACOSX)
 void WebContents::Focus() {
+  // Focusing on WebContents does not automatically focus the window on macOS,
+  // do it manually to match the behavior on Windows.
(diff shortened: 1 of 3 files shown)


CezaryKulakowski:
I've pushed a fixup with a test. This test fails without my change and passes with it. Unfortuately my change causes regressions in some other tests. I'm going to check why.

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2064,11 +2064,11 @@ void WebContents::CopyImageAt(int x, int y) {
     host->CopyImageAt(x, y);
 }
 
+#if !defined(OS_MACOSX)
```
zcbenz: The `web_contents()->Focus()` should still be called on macOS otherwise the specified WebContents may not be focused when the window has multiple WebContents.

shell/browser/api/electron_api_web_contents_mac.mm
```diff
@@ -11,6 +11,13 @@
 
 namespace api {
 
+void WebContents::Focus() {
+  if (GetType() != Type::BACKGROUND_PAGE) {
```
zcbenz: We should avoid testing type as new types may be added in future and things could break.

You can just test `owner_window()` here and there is no need for Cocoa code. And there should be a comment saying why this is needed:


```c++
void WebContents::Focus() {
  // Focusing on WebContents does not automatically focus the window on macOS,
  // do it manually to match the behavior on Windows.
#if defined(OS_MACOSX)
  if (owner_window())
    owner_window()->Focus(true);
#endif
  web_contents()->Focus();
}
```

spec-main/api-browser-window-spec.ts
```diff
@@ -3751,6 +3751,21 @@ describe('BrowserWindow module', () => {
     })
   })
 
+  describe('window.webContents.focus()', () => {
+    it('focuses window', (done) => {
+      const w1 = new BrowserWindow({ x:100, y: 300, width: 300, height: 200 })
```
zcbenz:
```
  3756:40  error  Missing space before value for key 'x'  key-spacing
  3758:40  error  Missing space before value for key 'x'  key-spacing
```

PR #22329, opened 2020-02-21
fix: emit will-navigate for sandboxed contents (#22188)

nornagon:
Backport of #22188.

Notes: Fixed "will-navigate" event not being emitted for sandboxed contents.


zcbenz:
Build is failing.

PR #22334, opened 2020-02-21
chore: wrap process restart flag in renderer process reuse check

MarshallOfSound:
This is in preparation for the removal of our process hacks.  Make it clearer what logic is required / runs under the two states of this flag.

Refs: #18397

Notes: no-notes

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -1504,17 +1504,23 @@ void WebContents::Stop() {
 }
 
 void WebContents::GoBack() {
-  electron::ElectronBrowserClient::SuppressRendererProcessRestartForOnce();
+  if (!ElectronBrowserClient::Get()->CanUseCustomSiteInstance()) {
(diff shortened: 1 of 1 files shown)


deepak1556:
Should also add a switch around loading our custom `navigation-controller` https://github.com/electron/electron/blob/master/lib/browser/navigation-controller.js which is not required in the process reuse code path , either in this PR or https://github.com/electron/electron/pull/22335 

Also add a switch around `RendererClient::ShouldFork` and follow what chromium does there.

MarshallOfSound:
@deepak1556 I think `RendererClient::ShouldFork` is tricky as without it we don't get `will-navigate`

MarshallOfSound:
Also yeah, I'm planning on going through the navigation controller, there is some code we'll still need there though like the `loadURL` being a promise

deepak1556:
> I think RendererClient::ShouldFork is tricky as without it we don't get will-navigate

Hmm yeah, we can leave it for cleanup post electron 11.

PR #22336, opened 2020-02-21
feat: set app.enableRendererProcessReuse to true by default

MarshallOfSound:
As per the timeline in #18397 we should flip this for Electron 9.

BREAKING CHANGE

Notes: Changed the default value of `app.allowRendererProcessReuse` to `true`, this will prevent loading of non-context-aware native modules in renderer processes.  See #18397 for more information on this change.


MarshallOfSound:
There are two spec fixes in this PR as well:

* The `window.open` fix is because the behavior of `nodeIntegration` in child windows changes when `allowRendererProcessReuse` is enabled.
* The `remote` fix is because the `webContents` is not marked as destroyed (as it is reused on reload) which makes the "function does not exist" error async as it has to bounce through the renderer to determine that the function does not exist

Issue #21907, opened 2020-01-27
'warn' event being handled twice


nornagon:
v5.0.10:
```
> process.emit('warning', new Error('warn'))
(node:55756) Error: warn
```
v6.1.3+:
```
> process.emit('warning', new Error('warn'))
(node:55797) Error: warn
(node:55797) Error: warn
```

I think we're running the `setupWarningHandler` (and probably other init code) twice during startup.

call sites:
```
Error: here
    at prepareMainThreadExecution (internal/bootstrap/pre_execution.js:11:11)
    at internal/bootstrap/environment.js:12:1
Error: here
    at prepareMainThreadExecution (internal/bootstrap/pre_execution.js:11:11)
    at internal/main/run_main_module.js:7:1
```

PR #22342, opened 2020-02-22
fix: don't run environment bootstrapper

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/21907.

In https://github.com/nodejs/node/pull/26788 (v12.0.0), Node.js added a bootstrapper to `CreateEnvironment`, which would prepare the main thread for execution any time an embedder created a new environment. However, this caused an unfortunate doubling-up effect; Node.js also ran bootstrapping (called `prepareMainThreadExecution`) for all other execution paths (like the repl, the actual main module, eval, etc). This meant that we'd see weird stuff like

```
> process.emit('warning', new Error('warn'))
(node:55797) Error: warn
(node:55797) Error: warn
```

since both execution pathways (env and whatever the actual execution was) were being exercised. To fix this, we can just remove bootstrapping code from `CreateEnvironment`. This also saves us a bunch of workaround code on our end, like the unsafe env block we had to prevent strange behavior during debugger setup. An upstream PR is open for this against Node.js itself, linked at bottom.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where some logging would double-print.


codebytere:
@nornagon oh good reminder - there are a few Node.js tests we had disabled that were actually commented out because of the undesirable behavior. I'll go re-add those and that should serve as a regression test!

Issue #3995, opened 2016-01-05
Document cryptographic algorithms used by Electron


jarek-foksa:
In order to submit Atom Electron app to the Mac App Store you have to provide information about the cryptography algorithms that your app "uses, accesses, implements or incorporates". Check [this tutorial](https://carouselapps.com/2015/12/15/legally-submit-app-apples-app-store-uses-encryption-obtain-ern/) for more info.

I would propose to create a documentation page that would list all cryptography algorithms used by Atom Electron so that we could reference it when filling the legal forms.

jarek-foksa:
Here is a list of algorithms I have pulled from Chromium's [WebCrypto](https://www.chromium.org/blink/webcrypto) page:
- RSASSA-PKCS1-v1_5 [RFC 3447](http://www.ietf.org/rfc/rfc3447)
- RSA-PSS [RFC 3447](http://www.ietf.org/rfc/rfc3447)
- RSA-OAEP - [RFC 3447](http://www.ietf.org/rfc/rfc3447)
- ECDSA - ANS X9.62–2005
- ECDH - ANS X9.63–2001
- AES-CTR - [NIST SP 800-38A](http://csrc.nist.gov/publications/nistpubs/800-38a/sp800-38a.pdf)
- AES-CBC - [NIST SP 800-38A](http://csrc.nist.gov/publications/nistpubs/800-38a/sp800-38a.pdf)
- AES-GCM - [NIST SP 800-38D](http://csrc.nist.gov/publications/nistpubs/800-38D/SP-800-38D.pdf) 
- AES-KW - [RFC3394](http://www.ietf.org/rfc/rfc3394.txt)
- HMAC - [FIPS 198-1](http://csrc.nist.gov/publications/fips/fips198-1/FIPS-198-1_final.pdf)
- SHA-1 - [FIPS 180-4](http://csrc.nist.gov/publications/fips/fips180-4/fips-180-4.pdf)
- SHA-256 - [FIPS 180-4](http://csrc.nist.gov/publications/fips/fips180-4/fips-180-4.pdf)
- SHA-384 - [FIPS 180-4](http://csrc.nist.gov/publications/fips/fips180-4/fips-180-4.pdf)
- SHA-512 - [FIPS 180-4](http://csrc.nist.gov/publications/fips/fips180-4/fips-180-4.pdf)
- HKDF - [NIST SP 800-56C](http://csrc.nist.gov/publications/nistpubs/800-56C/SP-800-56C.pdf)
[...]

PR #22358, opened 2020-02-24
docs: update MD2 link in cryptographic algorithms list

ferdnyc:
Replace link to MD2 RFC ([RFC 1319](https://tools.ietf.org/html/rfc1319)) with link to [RFC 6149](https://tools.ietf.org/html/rfc6149), which obsoletes and retires the algorithm.

(In addition to providing the most current information on the algorithm, this creates parity with the following entry, MD4, which links to its own obsoleting RFC ([RFC 6150](https://tools.ietf.org/html/rfc6150)), rather than to the original (and now-historical) protocol-defining RFC.

#### Checklist

#### Release Notes

Notes: no-notes


nornagon:
Hm, I'm actually not sure that this list is accurate, given we no longer ship OpenSSL. BoringSSL, for example, does not include the MD2 algorithm.

zcbenz:
Current list came from https://github.com/electron/electron/issues/3995, which was for OpenSSL and now outdated. 

I'm closing this PR since MD2 is no longer included.

(I'm not sure where to get an updated list, BoringSSL does not seem to provide such documentation.)

PR #22382, opened 2020-02-25
fix: add patch to set the base download URL rather than override it completely

### LINKED ISSUE #22381: Unable to get setSpellCheckerDictionaryDownloadURL to download dictionary
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * v8.0.1
* **Operating System:**
  * Arch Linux
  * N/A (new feature)

### Expected Behavior
I expect that when I use `win.webContents.session.setSpellCheckerDictionaryDownloadURL`, Electron downloads `hunspell_dictionaries.zip` and spell checks input boxes. 

### Actual Behavior
Dictionary file is not downloaded to `~/.config/Electron/Dictionaries` and no spellchecking is done. 

### To Reproduce
See demo repo at https://github.com/heronhaye/electron-8-spellchecker-test. Relevant excerpt below.

```js
function createWindow () {
  win = new BrowserWindow({
    width: 800,
    height: 600,
    webPreferences: {
      nodeIntegration: true,
      spellcheck: true
    }
  })

  win.webContents.session.setSpellCheckerDictionaryDownloadURL(
    'https://modalduality.keybase.pub/hunspell801.zip'
  )

  win.loadFile('index.html')

  win.webContents.openDevTools()

  win.on('closed', () => {
    win = null
  })
}
```

When I comment out the call to setSpellCheckerDictionaryDownloadURL, everything works as expected.

I verified that the sha256 hash of the zip file at the URL (`bc64c413308f6582df625833e4d9f35871ca438ba93d72afc3a1cf84779e2d50`) is the same as the hash of the file that was released in Electron 8.0.1.

```sh
$ git clone https://github.com/heronhaye/electron-8-spellchecker-test.git
$ npm install
$ npm start
```

### Additional Information
Let me know if I can help by doing other tests, thanks!

MarshallOfSound:
Closes #22381 

The old testing-only method we were using does not actually do what we want, we need a patch to support this use case.

Notes: fixed `session.setSpellCheckerDictionaryDownloadURL` to append the dictionary name at the end of requests instead of requesting the same URL for all dictionaries


patches/chromium/feat_add_support_for_overriding_the_base_spellchecker_download_url.patch
```diff
@@ -0,0 +1,61 @@
...
+@@ -254,8 +261,11 @@ GURL SpellcheckHunspellDictionary::GetDictionaryURL() {
+   std::string bdict_file = dictionary_file_.path.BaseName().MaybeAsASCII();
+   DCHECK(!bdict_file.empty());
+ 
++  if (g_base_download_url_override.Get() != GURL())
++    return GURL(g_base_download_url_override.spec() + base::ToLowerASCII(bdict_file));
++
+   static const char kDownloadServerUrl[] =
+-      "https://redirector.gvt1.com/edgedl/chrome/dict/";
++     "https://redirector.gvt1.com/edgedl/chrome/dict/";
```
nornagon: this change seems spurious...?

MarshallOfSound: yeah noticed that after push, got rid of it 👍

MarshallOfSound:
Self-validating on windows as the spellchecker tests haven't landed yet

MarshallOfSound:
Validated on windows 👍

PR #22383, opened 2020-02-25
docs: improve documentation on spellchecker download URL

MarshallOfSound:
Had to explain this in an issue, so it makes sense that the documentation is more clear about what this URL is and how it is used.

Notes: no-notes

### DIFF
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -483,6 +483,10 @@ behavior you can use this API to point the dictionary downloader at your own hos
 dictionaries.  We publish a `hunspell_dictionaries.zip` file with each release which contains the files you need
 to host here.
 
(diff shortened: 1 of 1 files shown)


docs/api/session.md
```diff
@@ -483,6 +483,10 @@ behavior you can use this API to point the dictionary downloader at your own hos
 dictionaries.  We publish a `hunspell_dictionaries.zip` file with each release which contains the files you need
 to host here.
 
+If the files present in `hunspell_dictionaries.zip` are available at `https://example.com/dictionaries/language-code.bdic`
+then you should call this api with `ses.setSpellCheckerDictionaryDownloadURL('https://example.com/dictionaries/')`.  Please
+note the trailing slash.  The URL to the dictionaries is formed as `${baseURL}${filename}`.
```
erickzhao: On this line, could we make the `${baseURL}` variable match the name of the parameter variable on line 479?

PR #22386, opened 2020-02-25
fix: allow persistent media salts

### LINKED ISSUE #22144: mediaDevices.enumerateDevices returns different deviceIds on every reload
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * Tested with 8.0.0 and v9.0.0-beta.1
* **Operating System:**
  * macOS 10.15.3 (19D76)
  * 7.1.12

### Expected Behavior

`navigator.mediaDevices.enumerateDevices()` should return a list of MediaDevices with a `label` and a `deviceId` property. The `deviceId` property should stay the same across reloads and restarts.

### Actual Behavior
The `deviceId` property changes with every reload/restart.

### To Reproduce
Simply execute `navigator.mediaDevices.enumerateDevices()` in the renderer process and compare the ids of the media devices to a different execution.

See this fork for an example:
https://github.com/codedrift/electron-quick-start

```sh
$ git clone https://github.com/codedrift/electron-quick-start -b media-device-ids
$ npm install
$ npm start || electron .
```

### Additional Information
It seems to work fine in a regular Chrome 80 browser.

MarshallOfSound:
Fixes #22144

A new BrowserClient method was added and we didn't implement it to retain the old behavior, this fills that gap 👍 

Notes: deviceId from `navigator.mediaDevices.enumerateDevices` is now consistent across reloads

### DIFF
--- a/shell/browser/electron_browser_client.cc
+++ b/shell/browser/electron_browser_client.cc
@@ -872,6 +872,14 @@ bool ElectronBrowserClient::ShouldUseProcessPerSite(
 #endif
 }
 
+bool ElectronBrowserClient::ArePersistentMediaDeviceIDsAllowed(
+    content::BrowserContext* browser_context,
+    const GURL& scope,
+    const GURL& site_for_cookies,
+    const base::Optional<url::Origin>& top_frame_origin) {
+  return true;
+}
+
 void ElectronBrowserClient::SiteInstanceDeleting(
     content::SiteInstance* site_instance) {
   // We are storing weak_ptr, is it fundamental to maintain the map up-to-date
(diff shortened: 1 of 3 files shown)


shell/browser/electron_browser_client.cc
```diff
@@ -872,6 +872,14 @@ bool ElectronBrowserClient::ShouldUseProcessPerSite(
 #endif
 }
 
+bool ElectronBrowserClient::ArePersistentMediaDeviceIDsAllowed(
+    content::BrowserContext* browser_context,
+    const GURL& scope,
+    const GURL& site_for_cookies,
+    const base::Optional<url::Origin>& top_frame_origin) {
+  return true;
```
deepak1556: Might be better to return `false` for `browser_context->IsOffTheRecord()` types, I don't see upstream covering that condition.

deepak1556: Nm: I misunderstood the persistent level of id , the spec makes it clear https://w3c.github.io/mediacapture-main/#privacy-and-security-considerations

LuoJinghua:
Could we make it optional or let the app decide whether this feature should be enabled for a site? Some web sites is already using the persistent media device id to tracking users.

koenoe:
Will this fix be released in 8.0.3? We're currently struggling with this issue.

PR #22397, opened 2020-02-26
chore: ensure correct scopes are in place

codebytere:
#### Description of Change

In ongoing efforts to match `ELECTRON_RUN_AS_NODE` as closely as possible to Node.js main instance execution, we should ensure that that scopes we have in place are correct.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/app/node_main.cc
+++ b/shell/app/node_main.cc
@@ -138,9 +138,15 @@ int NodeMain(int argc, char* argv[]) {
     // Initialize gin::IsolateHolder.
     JavascriptEnvironment gin_env(loop);
 
+    v8::Isolate* isolate = gin_env.isolate();
+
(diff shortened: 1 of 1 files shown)


shell/app/node_main.cc
```diff
@@ -138,10 +138,16 @@ int NodeMain(int argc, char* argv[]) {
     // Initialize gin::IsolateHolder.
     JavascriptEnvironment gin_env(loop);
 
+    v8::Isolate* isolate = gin_env.isolate();
```
codebytere: the isolate we were pulling out of the Node.js env is the same as the one in the gin env, so this simply makes that clearer to the reader.

PR #22404, opened 2020-02-27
feat: MessagePorts in the main process

nornagon:
#### Description of Change

This adds a new IPC primitive to Electron based on [MessageChannel][], the DOM primitive. MessageChannel already worked fine in the renderer process, and between renderer processes opened through `window.open` (when `nativeWindowOpen` or `sandbox` is enabled). This change allows passing `MessagePort`s from renderer to the main process. From there, the ports can be forwarded on to other renderers, which wouldn't otherwise necessarily be able to communicate with one another.

Communication over `MessagePort` between two renderer processes is not proxied through the main process. (Caveat: if messages are sent on the port before the other end of it is fully set up in the target renderer, those messages may be queued in the main process until the channel is ready.)

Initial proposal: https://hackmd.io/bwzFRWZzRIyewIBc1Vf7DQ?both

This change adds some new APIs:
#### `ipcRenderer.postMessage(channel, message, [options])`

This is similar to the existing `ipcRenderer.send`, with two key differences. First, you can only send a single argument, as opposed to `ipcRenderer.send` which forwards all arguments after the channel name. Secondly, the 3rd argument is an array of "transferable objects", i.e. `MessagePort`s. To send a `MessagePort` to the main process, you must _transfer_ it by passing it in the 3rd argument of `postMessage`, like so:

```js
const {port1, port2} = new MessageChannel
ipcRenderer.postMessage("my-message", {some: "message"}, [port1])
```

This is intentionally similar to [`window.postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage).

Currently, the only kind of transferable object that may be passed in the transfer list is `MessagePort`. In future, more kinds of transferable objects such as `SharedArrayBuffer` may be supported.

TODO: currently this function only accepts an array, but we should match `Window.postMessage` and also accept an object of the form `{transfer: [...]}`.

#### `WebContents.postMessage(channel, message, [options])`

This is the same as `ipcRenderer.postMessage` but going in the other direction, from main to renderer.

#### `MessagePortMain`

As the JS context in the main process in Electron is not a Blink context, we do not have access to Blink's implementation of `MessagePort`. Thus, we provide a polyfill, which communicates over the same underlying Mojo pipe.

This has a similar API to [MessagePort][], but instead of `port.addEventListener('message', ...)` or `port.onmessage = ...`, we adopt the Node.js convention and offer the EventEmitter-style `port.on('message', ...)`.

This object can be passed as a transferable to `WebContents.postMessage` in order to transfer ownership of the message port to a renderer.

---

[MessageChannel]: https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel
[MessagePort]: https://developer.mozilla.org/en-US/docs/Web/API/MessagePort

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added support for `MessagePort` in the main process.


nornagon:
One potential improvement: as written there's currently no way to send a `MessagePort` to a renderer frame other than the main frame.

nornagon:
TODO:
- [ ] There's currently no way to create a `MessageChannel` from the browser process; only to receive them from the renderer. There should be such a way.
- [ ] Error handling is very haphazard currently.
- [ ] `postMessageToFrame` (though I think it would be OK to leave this to a followup PR)
- [ ] Consider patching in an `onclose` event: https://github.com/w3ctag/design-reviews/issues/269#issuecomment-407584290
- [ ] Allow passing an object for the 3rd argument of `postMessage`, to match `Window.postMessage`.
- [ ] `BrowserSideMessagePort.postMessage` doesn't yet extract transferables that are passed to it, so it's not possible to pass a message port across another message port from the browser.
- [ ] References to a transferred `MessagePort` in the message object being sent should resolve to that `MessagePort`. i.e. `ipcRenderer.postMessage("foo", { port: port1 }, [port1])` should work.
- [ ] Blink implements rate-limiting on MessagePort messages. Evaluate if we should also do that on the browser side.
- [ ] Tests for failure/edge cases:
  - [ ] Neutered ports
  - [ ] Sending messages to closed ports
  - [ ] Attempting to transfer things other than MessagePorts, both on the browser and renderer side
  - [ ] Attempting to transfer duplicate or neutered ports

nornagon:
Some other things to think about:
- Once we support Node.js `worker_threads` in the main process, can we enable passing `MessagePort`s to workers in the main process?
- `MessagePort`s in web workers / service workers in the renderer process ought to work fine, but it's probably worth writing a couple of tests to make sure.

nornagon:
Removing the [wip] tag as this is 90% ready. The only remaining major chunk of work is validating that the GC behavior of `MessagePortMain` is correct & fixing it if it isn't.

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1086,6 +1095,57 @@ void WebContents::Invoke(bool internal,
...
+    if (!gin::ConvertFromV8(isolate(), transferables, &wrapped_ports)) {
+      args->ThrowError();
+      return;
+    }
+  }
+
+  blink::TransferableMessage transferable_message;
+  if (!electron::SerializeV8Value(args->isolate(), message_value,
+                                  &transferable_message)) {
+    args->ThrowError();
```
zcbenz: The error would be complaining the 3rd arg instead of 2rd arg because of the `GetNext` above, can probably use `ThrowTypeError` to throw a custom error.

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1086,6 +1095,57 @@ void WebContents::Invoke(bool internal,
                  std::move(callback), internal, channel, std::move(arguments));
 }
 
+void WebContents::ReceivePostMessage(const std::string& channel,
+                                     blink::TransferableMessage message) {
+  auto wrapped_ports = MessagePort::EntanglePorts(isolate(), message.ports);
```
zcbenz: Should this be `std::move(message.ports )`?

shell/browser/api/electron_api_web_contents.cc
nornagon: Switched the ordering instead, and added a test.

shell/browser/api/electron_api_web_contents.cc
nornagon: sure, that saves us a vector copy. it's semantically the same though.

docs/api/ipc-renderer.md
```diff
@@ -143,6 +143,35 @@ Sends a message to a window with `webContentsId` via `channel`.
 Like `ipcRenderer.send` but the event will be sent to the `<webview>` element in
 the host page instead of the main process.
 
+### `ipcRenderer.postMessage(channel, message, [options])`
+
+* `channel` String
+* `message` any
+* `options` MessagePort[] (optional)
```
MarshallOfSound: Calling this options when it's an array of ports feels weird.  Either this should be an object of `{ ports[] }` or this should be called `ports`

docs/api/message-port-main.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Instance Events
+
+#### Event: 'message'
+
+Returns:
+
+* `messageEvent` Object
+  * `data` any
+  * `ports` MessagePortMain[]
```
MarshallOfSound: If no ports are provided is this an empty array or undefined?

docs/api/message-port-main.md
```diff
@@ -0,0 +1,45 @@
...
+`MessagePortMain` is an [EventEmitter][event-emitter].
+
+## Class: MessagePortMain
+
+### Instance Methods
+
+#### `port.postMessage(message, [transfer])`
+
+* `message` any
+* `transfer` MessagePortMain[] (optional)
```
MarshallOfSound: We should standardize on `options` vs `transfer` so that the main and renderer docs at least align

docs/api/structures/ipc-main-event.md
```diff
@@ -3,6 +3,7 @@
 * `frameId` Integer - The ID of the renderer frame that sent this message
 * `returnValue` any - Set this to the value to be returned in a synchronous message
 * `sender` WebContents - Returns the `webContents` that sent the message
+* `ports` MessagePortMain[] - A list of MessagePorts that were transferred with this message
```
MarshallOfSound: Same question if none are provided, empty array or undefined?

docs/api/web-contents.md
```diff
@@ -1593,6 +1593,32 @@ ipcMain.on('ping', (event) => {
 })
 ```
 
+#### `contents.postMessage(channel, message, [options])`
+
+* `channel` String
+* `message` any
+* `options` MessagePortMain[] (optional)
```
MarshallOfSound: options vs transfer

lib/browser/message-port-main.ts
```diff
@@ -0,0 +1,25 @@
+import { EventEmitter } from 'events'
+
+export class MessagePortMain extends EventEmitter {
+  _internalPort: any
```
MarshallOfSound: Is it worth making this a v8 hidden or otherwise hiding this from end users, or is it ok that this is theoretically usable

patches/chromium/.patches
```diff
@@ -73,6 +73,7 @@ expose_setuseragent_on_networkcontext.patch
 feat_add_set_theme_source_to_allow_apps_to.patch
 revert_cleanup_remove_menu_subtitles_sublabels.patch
 export_fetchapi_mojo_traits_to_fix_component_build.patch
+add_webmessageportconverter_entangleandinjectmessageportchannel.patch
```
MarshallOfSound: Is this upstreamable?

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1086,6 +1095,59 @@ void WebContents::Invoke(bool internal,
...
+  auto wrapped_ports =
+      MessagePort::EntanglePorts(isolate(), std::move(message.ports));
+  v8::Local<v8::Value> message_value =
+      electron::DeserializeV8Value(isolate(), message);
+  EmitWithSender("-ipc-ports", bindings_.dispatch_context(), InvokeCallback(),
+                 false, channel, message_value, std::move(wrapped_ports));
+}
+
+void WebContents::PostMessage(gin::Arguments* args) {
+  std::string channel;
```
MarshallOfSound: Why are we pulling the arguments out manually here instead of letting gin to the work?

```
void WebContents::PostMessage(std::string channel, v8::Local<v8::Value> message_value, ...)
```

shell/browser/api/message_port.cc
```diff
@@ -0,0 +1,276 @@
...
+  // Walk the incoming array - if there are any duplicate ports, or null ports
+  // or cloned ports, throw an error (per section 8.3.3 of the HTML5 spec).
+  for (unsigned i = 0; i < ports.size(); ++i) {
+    auto* port = ports[i].get();
+    if (!port || port->IsNeutered() || visited.find(port) != visited.end()) {
+      std::string type;
+      if (!port)
+        type = "null";
+      else if (port->IsNeutered())
+        type = "already neutered";
```
MarshallOfSound: Can we explain in this error or in docs what a "neutered" port is 👍

shell/browser/api/electron_api_web_contents.cc
nornagon: done

typings/internal-ambient.d.ts
```diff
@@ -22,6 +22,7 @@ declare namespace NodeJS {
     sendToHost(channel: string, args: any[]): void;
     sendTo(internal: boolean, sendToAll: boolean, webContentsId: number, channel: string, args: any[]): void;
     invoke<T>(internal: boolean, channel: string, args: any[]): Promise<{ error: string, result: T }>;
+    postMessage(channel: string, message: any, transferables: any[] | Object): void;
```
MarshallOfSound: I don't think `any[] | Object` is right here, these have to be ports right?

shell/renderer/api/electron_api_renderer_ipc.cc
```diff
@@ -95,14 +99,52 @@ class IPCRenderer : public gin::Wrappable<IPCRenderer> {
...
+    if (!electron::SerializeV8Value(isolate, message_value,
+                                    &transferable_message)) {
+      // SerializeV8Value sets an exception.
+      return;
+    }
+
+    std::vector<v8::Local<v8::Object>> transferables;
+    if (transfer) {
+      if (!gin::ConvertFromV8(isolate, *transfer, &transferables)) {
+        isolate->ThrowException(v8::Exception::Error(
```
MarshallOfSound: Can we use `ErrorThrower` here instead of throwing exceptions directly.

nornagon:
Pending approval from @electron/wg-api on https://github.com/electron/governance/pull/204

docs/api/ipc-renderer.md
```diff
@@ -143,6 +143,35 @@ Sends a message to a window with `webContentsId` via `channel`.
 Like `ipcRenderer.send` but the event will be sent to the `<webview>` element in
 the host page instead of the main process.
 
+### `ipcRenderer.postMessage(channel, message, [transfer])`
```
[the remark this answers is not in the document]
nornagon: No, the intention here isn't to replace the existing methods.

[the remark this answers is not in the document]
nornagon: We should definitely clearly document when to use which one of `invoke`, `send` and `postMessage`, though.

nornagon:
Removing api-review-requested label as the spec has been approved electron/governance#204

willium:
A couple quick/naive questions re: this change!

1. Are there performances implications of using a hidden BrowserWindow to do work versus a child_process? It seems like child_process would be more performant. If that’s right, @nornagon—would you recent exciting work on MessagePorts work with child_process as well?

2. It seems this would allow for transferrables between the main process and either renderer process... would there be a similar way to allow transferables between the renderer processses

nornagon:
> A couple quick/naive questions re: this change!
> 
>     1. Are there performances implications of using a hidden BrowserWindow to do work versus a child_process? It seems like child_process would be more performant. If that’s right, @nornagon—would you recent exciting work on MessagePorts work with child_process as well?

Depends what you're optimizing for. Hidden BrowserWindows seem OK to me, I'm not aware of any major performance reasons to avoid them. The MessagePorts in this PR would not be transferable to child processes.

>     2. It seems this would allow for transferrables (no-cost shared memory?) between the main process and either renderer process... would there be a similar way to allow transferables between the renderer processses?

Sharing transferables other than `MessagePort` with the main process is not currently supported, but it will work fine across `MessagePort`s in renderers, which the main process can facilitate. For example:

Main:
```js
const w1 = new BrowserWindow()
const w2 = new BrowserWindow()

ipcMain.on('port', (e) => {
  // Forward the ports to w2
  w2.webContents.postMessage('port', null, e.ports)
})

await w2.loadFile('w2.html')
w1.loadFile('w1.html')
```

In w1:
```js
const {port1, port2} = new MessageChannel
const buf = new SharedArrayBuffer(16)
port2.postMessage(buf)
ipcRenderer.postMessage('port', null, [port2])
```

In w2:
```js
ipcRenderer.on('port', (e) => {
  const [port] = e.ports
  port.onmessage = (e) => {
    const buf = e.data
    // buf here is the same SharedArrayBuffer as in w1.
  }
})
```

This is the same pattern as when using `window.postMessage` on the Web, but it allows the main process to permit cross-origin messaging.

Mike-Dax:
@nornagon Do you know of the performance implications of this versus using `ipcRenderer.sendTo` between renderer processes?

nornagon:
> @nornagon Do you know of the performance implications of this versus using `ipcRenderer.sendTo` between renderer processes?

I have not performed any benchmark, but `sendTo` will pass the message via the main process, whereas `MessagePort` allows a direct connection between two renderer processes. Additionally, `sendTo` will result in the message being deserialized and reserialized in the main process, while `MessagePort` requires only a single serialize/deserialize pair.

nornagon:
Linux failure is unrelated; merging.

Mike-Dax:
@nornagon Thank you for your insight. I'll do some benchmarks on the various options.

Is it likely for this feature to be back-ported for v8.2.0 or v9.0.0?

nornagon:
No, this feature will land in v10.

Mike-Dax:
Hi @nornagon

I build Electron 10.0.0-nightly.20200316 and did some benchmarks.

I'm actually finding `ipcRenderer.sendTo` to not only be faster than using this MessageChannel API, but also have a tighter standard deviation.

To benchmark I'm setting up two BrowserWindows, designating one as a 'server' and one as a 'client'. The client merely echos data back to the server.

I'm then sending 10k messages in a row and waiting for the last 'pong' back. I'm sending small objects that look like this:

```
{
  keyA: index,
  keyB: index * 1000
}
```

Results

```
sendTo x 1.51 ops/sec ±0.46% (12 runs sampled)
MessageChannel x 1.19 ops/sec ±2.43% (10 runs sampled)
```

The exact code used is available here:

https://github.com/Mike-Dax/pipe-benchmarks

I'm finding similar results with large binary messages, with sendTo being about 7% faster.

nornagon:
@Mike-Dax thanks for those benchmarks, that's really interesting! I expanded on the benchmarks and went digging in Blink to try to figure out what was going on. Here's what I found.

1. Blink's postMessage implementation [yields the thread](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/messaging/message_port.cc;l=269-273;drc=539e6b94b4f95c1710718085c0e526a69611b2bd) after 200 messages have been received or 50ms has passed. I tried disabling this logic in my local build and the two benchmarks became approximately equal. Blink has a TODO in it saying the logic should be removed once [this](https://chromium-review.googlesource.com/c/chromium/src/+/1145692) lands, which it has. I'll look into whether it's reasonable to remove that logic upstream.
2. This benchmark measures throughput, rather than end-to-end latency, so the performance difference between bouncing to the main process vs. going direct isn't likely to be dominant, because Mojo pipelines all its messages. So the indirection to the main process only adds a constant overhead per cycle.

I tried modifying the benchmark to send 10 1MB ArrayBuffers instead of 10000 small objects, and found `MessageChannel` to be about 10% faster:

```
MessageChannel x 29.36 ops/sec ±1.57% (68 runs sampled)
sendTo x 26.41 ops/sec ±0.82% (62 runs sampled)
```

And that difference is exacerbated in my local build with the Blink yielding logic removed:

```
MessageChannel x 5.27 ops/sec ±3.35% (27 runs sampled)
sendTo x 1.95 ops/sec ±1.09% (14 runs sampled)
```

nornagon:
@Mike-Dax update, I was able to get the changes merged upstream into Blink: https://chromium-review.googlesource.com/c/chromium/src/+/2112954, so we'll pick them up in the next Chromium roll. After that, throughput should be similar between `MessageChannel` and `sendTo` for small objects. Performance for large arrays should be much better over `MessageChannel`. And end-to-end latency ought to still be better over `MessageChannel`; I'll try to find some time to set up a benchmark for that.

Also, as a bonus, `MessageChannel` in Blink (and thus Chrome, Edge and Opera) will get more performant too :)

Mike-Dax:
@nornagon Oh incredible. Thank you so much!

I'll likely have some time in the coming week or two to do more benchmarks - I'll make one for latency as well so you don't have to. Benchmark.js isn't a good fit for measuring latency so I'll need to make something bespoke.

Yeah it's interesting, a named pipe implementation (that transfers all serialisation / deserialisation responsibility to the dev) is about 3x faster on the throughput side for me, and lower latency as well, however I'm getting weird cases where it just won't connect on Windows, specifically on the first 'install' of Electron.

willium:
> Currently, the only kind of transferable object that may be passed in the transfer list is MessagePort. In future, more kinds of transferable objects such as SharedArrayBuffer may be supported.
> TODO: currently this function only accepts an array, but we should match Window.postMessage and also accept an object of the form {transfer: [...]}.

When other types of transferable objects are supported would you expect support for  [OffscreenCanvas](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/transferControlToOffscreen)?

nornagon:
@willium no. That would require support for Canvas in the main process, which is not a Blink context.

willium:
The new message channel APIs are awesome, but the docs are a little unclear ~ I'm curious if there is there any implementation difference or tradeoff between:
1. using new MessageChannelMain() in the main process and and sending the ports to two renderer processes (described [here](https://www.electronjs.org/docs/api/message-channel-main) and in @nornagon's [original proposal](https://hackmd.io/bwzFRWZzRIyewIBc1Vf7DQ?both#Example-usage-of-new-API))
2. using new MessageChannel() in one renderer process and send one of the ports to main and then forwarding it to another renderer?  (described here in this PR)

Also, as a follow-up, in your [example](https://github.com/electron/electron/pull/22404#issuecomment-597929387) @nornagon, would there not be a race condition with the postMessage call before the other port is even sent to the main process (let alone to the other renderer which will listen on the MessageChannel)? If so, is there a solid pattern to avoid this besides facilitating a handshake?

```ts
const {port1, port2} = new MessageChannel
const buf = new SharedArrayBuffer(16)
port2.postMessage(buf)                                      // posting message before sending port
ipcRenderer.postMessage('port', null, [port2])
```

willium:
Also, if there is a better place to ask these questions, my apologies. Let me know, happy to engage where you are!

nornagon:
@willium there's no difference in performance based on how the two ends of the channel are created. Regardless of how the ports arrive at their destination, performance characteristics will thereafter be the same.

There's no race condition in that example because messages are queued until the port is started (which happens automatically when an `onmessage` handler is registered). See https://developer.mozilla.org/en-US/docs/Web/API/MessagePort/start.

willium:
I'm having a hell of a time getting MessageChannels to work while also using 
```
contextIsolation: true,
nodeIntegration: false,
```
I was able to attach listeners on the port's events within the preload. But oddly serializing the message and forward it removes all properties except `isTrusted` but you can manually pull off those properties and send them to the renderer. 
```

contextBridge.exposeInMainWorld("intercom", {
	on: (channel: string, callback: (...args: any[]) => any) => {
		// Filtering the event param from ipcRenderer
		// eslint-disable-next-line @typescript-eslint/no-unsafe-return
		ipcRenderer.on(channel, (e: IpcRendererEvent, data: any) => {
			// eslint-disable-next-line @typescript-eslint/no-unsafe-return
			// callback(e)
			e.ports.forEach((port, i) => {
				console.log("port", port)
				port.onmessage = (m: MessageEvent) => {
					callback(
						data,
						"message",
						i,
						m.data,
						m.bubbles,
						m.timeStamp,
						JSON.stringify(m),
					)
				}
				port.onmessageerror = (m: MessageEvent) => {
					callback(data, "messageerror", i, m.data, JSON.stringify(m))
				}
				port.postMessage("BAR")
				// port.start()
			})
		})
	},
})
```
output within `callback`:
Screen Shot 2020-12-14 at 11 45 30 PM

Additionally, I haven't found a way to allow the renderer to attach a callback to the port's message event(s) and handle the arguments itself. After inspecting the messageEvent type, it seems everything is available in the DOM context, so I'm not sure why that's not working—am I missing something?

willium:
Also, though it's mostly syntactical: I believe that typically with MessageChannel's you can put your transferable inside the message, but that seems to throw an error with MessageChannelMain because it can't serialize MessagePorts. I'm not sure if this is something worth adding to the roadmap or not :)

willium:
Are there any docs or examples on combining message channels with contextBridge? As methods on the port aren’t available outside of the closure of the ipc event, it's not clear how they can exposed on the contextBridge

nornagon:
@willium i'm having a hard time understanding your example as it's incomplete. Would you be able to provide a gist (preferably using [Electron Fiddle](/electron/fiddle)) that shows the issue you're having?

Also, you might have better luck with this question at one of the links at [the project's Community page](https://www.electronjs.org/community) -- in particular, be sure to give [the Electron Discord server](https://discord.com/invite/electron) a try.

willium:
Hi Jeremy, thanks for your reply! I'll work on a minimal reproduction now, but I don't believe Fiddle allows for multiple renderer processes just yet, so that may not be as trivial to emulate. 

Perhaps I can best isolate the questions for you: 
1. Transferables seems to be copied as they go over the ContextBridge, is there anyway to send Transferables using MessagePorts without incurring a copy step as they go over the ContextBridge? 
2. What is the best way to interface with the MessagePort on the renderer thread with context isolation? Is there someway to attach the MessagePort to the `window` with ContextBridge, or must it done within a closure?

[The docs](https://www.electronjs.org/docs/api/context-bridge#contextbridge) suggest this pattern:
```ts
contextBridge.exposeInMainWorld(
  'electron',
  {
    doThing: () => ipcRenderer.send('do-a-thing')
  }
)
```

but it's not straightforward to use this pattern with a MessagePort, as its only accessible within the callback of 
```ts
ipcRender.on("port", ({ port }) => { ... }
```

 e.g.
 
 ```ts

contextBridge.exposeInMainWorld("electron", {
	connect: (port: PortName, connection: Connection) => {
		if (!validPorts.includes(port)) {
			throw new Error(`Invalid Channel "${port}"`)
		}

		const wrappedCallback = ({ ports }: IpcRendererEvent, _: Payload) => {
			const [messagePort] = ports
			if (!messagePort) throw new Error("Connection requires (1) port")

			const post = (outgoing: Payload, transferables: Transferable[]) => {
				messagePort.postMessage(outgoing, transferables)
			}

			const reply: Reply = (outgoing: Payload) => {
				// same channel (channel should also be in outgoing payload)
				messagePort.postMessage(outgoing)
			}

			const disconnect = () => {
				ipcRenderer.removeListener(port, wrappedCallback)
			}

			messagePort.onmessage = ({ data }: MessageEvent<Payload>) => {
				connection(data, post, reply, disconnect)
			}
			messagePort.onmessageerror = ({ data }: MessageEvent<Payload>) => {
				connection(data, post, reply, disconnect)
			}

			messagePort.start()
		}

		ipcRenderer.on(`${port}:port`, wrappedCallback)
	},
})
```

**Comment by willium** [comment]
By the way, I appreciate your responsiveness and I really don't mean to overuse issues as a support forum! I've attempted to engage in the community support channels, but both MessageChannels and ContextBridge are fairly new patterns without a ton of documentation (especially not when used together). And, with ContextBridge especially, much of what you find on StackOverflow/etc employs dark patterns of exposing `ipcRenderer`. 

I do believe this might be pointing a to a bug or at least a blocker for allowing MessageChannel use with Context Isolation active.

 In any case, I'd be happy to help contribute documentation once I figure this out!

**Comment by nornagon** [comment]
Let's continue discussion at #27024.

**Comment by bpasero** [comment]
@nornagon this API seems great for a use-case we have in VSCode: direct window-window communication without going through the main process, thanks a lot for adding it 👍 

I was able to come up with a minimal sample for how to communicate between 2 windows, maybe this is something that would make the docs clearer:

```ts

//main
const window1 = new BrowserWindow(...);
const window2 = new BrowserWindow(...);

ipcMain.on('port', e => {
  window2.webContents.postMessage('port', null, e.ports);
});

//window1
const { port1, port2 } = new MessageChannel();
ipcRenderer.postMessage('port', null, [port2]);

port1.onmessage = msg => console.log('[From Renderer 1', msg);
port1.postMessage('Hello From Renderer 1');

//window2
ipcRenderer.on('port', event => {
  [port] = event.ports;

  port.onmessage = msg => console.log('From Renderer 2', msg);
  port.postMessage('Hello From Renderer 2');
});
```

If this seems like a good sample, I am happy to open a PR.

One question: the purpose of `MessageChannelMain` is not immediately obvious to me. What is the use-case over just `ipcRenderer` and `ipcMain` methods? Is there any benefit? 

One use-case I think could be to prepare the ports on the main side and send them over into the 2 windows? But isn't this something covered by my example already?

Thanks for clarification 👍

Issue #22242, opened 2020-02-17
webRequest onBeforeSendHeaders + onHeadersReceived CORS net::ERR_FAILED error


vladimiry:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** v9.0.0-beta.2, 10.0.0-nightly.20200216
* **Operating System:** Linux
* **Last Known Working Electron version:** 8.0.1

### Expected Behavior
The following text gets printed on screen on "Send cross-domain XHR request" button click:
```
{"status":422,"statusText":"Unprocessable Entity","responseText":"{\"Code\":10080,\"Error\":\"some-user-name@protonmail.ch does not exist. Did you mean SomeUsername@protonmail.com?\",\"ErrorDescription\":\"\",\"Details\":{}}"}
```
![ok](https://user-images.githubusercontent.com/1560781/74688785-cdc9b680-51e9-11ea-9c43-3f06fadee40b.png)

### Actual Behavior
The following text gets printed on screen on "Send cross-domain XHR request" button click:
```
{"status":0,"statusText":"","responseText":"EMPTY"}
XMLHttpRequest error
```
![error](https://user-images.githubusercontent.com/1560781/74688794-d1f5d400-51e9-11ea-93c3-b8a4e427572f.png)

### To Reproduce

https://github.com/vladimiry-playground/electron-quick-start-cors-issue2

samuelmaddock:
There's a good chance this is related to this work ongoing in #19447. Chrome extensions now use their native implementation and could still need work to fully support some `chrome.*` extension APIs.

vladimiry:
Can confirm issue resolving, thanks @zcbenz.

airking05:
I repro in 8.1.1-9.0.0-beta.13 that responseHeaders cannot be changed by "onHeadersReceived".
I cannot add responseHeaders to "Access-Control-Allow-Origin" attribute at all...

vladimiry:
@airking05 I think you need to open a new issue then, regression like.

airking05:
@vladimiry thank you for reply.
I'll open a new issue :)

vladimiry:
[...]

PR #22407, opened 2020-02-27
fix: make webRequest work for CORS preflight requests

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/22242.

Recent Chromium upgrade has changed how CORS requests are handled, this PR updates our webRequest implementation to correctly handle CORS preflight requests.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix `webRequest` not working for CORS requests.

### DIFF
--- a/shell/browser/net/proxying_url_loader_factory.cc
+++ b/shell/browser/net/proxying_url_loader_factory.cc
@@ -20,6 +20,7 @@
 #include "shell/common/options_switches.h"
 
 namespace electron {
+
 ProxyingURLLoaderFactory::InProgressRequest::FollowRedirectParams::
     FollowRedirectParams() = default;
 ProxyingURLLoaderFactory::InProgressRequest::FollowRedirectParams::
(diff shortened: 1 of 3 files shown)


shell/browser/net/proxying_url_loader_factory.cc
```diff
@@ -465,15 +533,34 @@ void ProxyingURLLoaderFactory::InProgressRequest::
       current_response_->headers = override_headers_;
     }
   }
+
+  if (for_cors_preflight_ && !redirect_url_.is_empty()) {
+    OnRequestError(network::URLLoaderCompletionStatus(net::ERR_FAILED));
```
deepak1556: Can you add a test to check for modifying headers with redirect url fails in `onBeforeSendHeaders` and `onHeadersReceived`.

zcbenz: The `redirect_url_` can only be changed in `onBeforeRequest`, but once we do that this code path won't be run. It seems that this check is only to guard edge cases that do not usually happen.

Issue #17558, opened 2019-03-26
macOS: BrowserWindow.destroy() can cause segfaults


sa-MatteoHausner:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details
* **Electron Version:**
  * 4.1.1
* **Operating System:**
  * macOS 10.14.3
* **Last Known Working Electron version:**
  * Unknown

### Expected Behavior
Creating a **modal** `BrowserWindow` and closing it via `destroy()` on macOS, followed by creating a second **modal** `BrowserWindow` and closing it via `close()` should not produce any visual artifcats, nor leave the application in a inoperable state, nor lead to segmentation faults.

### Actual Behavior
Creating a **modal** `BrowserWindow` and closing it via `destroy()`, followed by creating a second **modal** `BrowserWindow` and closing it via `close()` summons parts (visual artifacts) of the previously destroyed first BrowserWindow (see our test-case and attached screenshot).  
Afterwards the application does not react to any input anymore and must be terminated forcefully.  
On MacBookPros with a touchbar, in many cases, the application crashes with a segmentation fault or illegal hardware instruction:
```
Application Specific Information:
Crashing on exception: -[__NSTaggedDate makeTouchBar]: unrecognized selector sent to instance 0x800007fc55bf8e7d

Application Specific Backtrace 1:
0   CoreFoundation                      0x00007fff2bc3aded __exceptionPreprocess + 256
1   libobjc.A.dylib                     0x00007fff57d02720 objc_exception_throw + 48
2   CoreFoundation                      0x00007fff2bcb8195 -[NSObject(NSObject) __retain_OA] + 0
3   CoreFoundation                      0x00007fff2bbdca60 ___forwarding___ + 1486
4   CoreFoundation                      0x00007fff2bbdc408 _CF_forwarding_prep_0 + 120
5   Electron Framework                  0x000000010d421c38 _ZN4atom19AtomNativeWidgetMac14CreateNSWindowERKN5views6Widget10InitParamsE + 568
6   AppKit                              0x00007fff291d4ba8 -[NSResponder(NSTouchBarProvider) touchBar] + 56
[...]

PR #22410, opened 2020-02-27
fix: prevent potential modal window close segfault

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/17558.

We were previously trying to call `close` on a sheet itself, instead of calling `[NSWindow endSheet:]` on the sheet's parent. As such, the window would either freeze since the modal session wasn't ended properly, or crash. To fix this, we need to end the sheet properly by calling `[sheetParent endSheet:window];`, where `window` is the sheet itself.

However! If we just call that synchronously, we block the thread with an animation, so we instead post a task. This fix was verified with the repro provided in the issue linked above.

cc @zcbenz @nornagon @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an occasional segfault with modal windows being closed or destroyed.

### DIFF
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -588,7 +588,14 @@ void NativeWindowMac::SetContentView(views::View* view) {
 void NativeWindowMac::Close() {
   // When this is a sheet showing, performClose won't work.
   if (is_modal() && parent() && IsVisible()) {
-    [parent()->GetNativeWindow().GetNativeNSWindow() endSheet:window_];
+    NSWindow* window = parent()->GetNativeWindow().GetNativeNSWindow();
+    if (NSWindow* sheetParent = [window sheetParent]) {
+      base::ThreadTaskRunnerHandle::Get()->PostTask(
+          FROM_HERE, base::BindOnce(base::RetainBlock(^{
+            [sheetParent endSheet:window];
+          })));
+    }
+
     // Manually emit close event (not triggered from close fn)
     NotifyWindowCloseButtonClicked();
     CloseImmediately();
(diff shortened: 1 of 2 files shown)


codebytere:
@MarshallOfSound we're not - if you look at the change we used to disable _all_ modal tests for darwin, and my change makes it such that we disable individual tests and not the suite, since i do want the new test to run on darwin

shell/browser/native_window_mac.mm
```diff
@@ -588,7 +588,14 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
...
+    if (NSWindow* sheetParent = [window sheetParent]) {
+      base::ThreadTaskRunnerHandle::Get()->PostTask(
+          FROM_HERE, base::BindOnce(base::RetainBlock(^{
+            [sheetParent endSheet:window];
+          })));
+    }
+
     // Manually emit close event (not triggered from close fn)
     NotifyWindowCloseButtonClicked();
     CloseImmediately();
```
deepak1556: The sheet is being closed async, will this function be safe to close the window synchronously ?

codebytere: @deepak1556 i believe so - this is the same approach that [chromium took](https://source.chromium.org/chromium/chromium/src/+/master:components/remote_cocoa/app_shim/views_nswindow_delegate.mm;l=137-149?q=%22%5BsheetParent%20endSheet:window%5D;%22&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2F)

deepak1556: Thanks for the pointer.

Issue #22409, opened 2020-02-27
Print crash in Electron 8.0.2, Mac OS Catalina 10.15.3


thomasdao:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details
App crashes when call `webContents.print()`
* **Electron Version:**
8.0.2
* **Operating System:**
Mac OS Catalina 10.15.3 (19D76)
* **Last Known Working Electron version:**
Electron 6 did not crash when print but has other bugs with printing

### Expected Behavior

App should not crash
### Actual Behavior

App crashes
### To Reproduce

Can be reproduced with this Fiddle
https://gist.github.com/1aced5931b191c07adb6a111f937a4e1

codebytere:
Do you have a crash dump @thomasdao? I ran this fiddle and experienced no issues.

thomasdao:
@codebytere please check this crash dump, thanks https://drive.google.com/file/d/1pJn0XTqiQoIzR69thIA3DIgWwrY5H7qz/view?usp=sharing

thomasdao:
By the way the app crashes immediately when I clicked "Print" button, and it did not show any print dialog. My Mac doesn't have any printer connected. However in Electron 6, I can see the print dialog. Please let me know if you need any other info, thanks!

codebytere:
@thomasdao ah, this was the crucial piece:
> My Mac doesn't have any printer connected.

I nuked local printers from my mac and could repro. Looks like this edge case is specific only to when no local printers are connected to your mac, which is annoying, but also makes triage a little easier 😛 I'll get on this but in the meantime if you connect literally any printer to your mac (doesn't matter if it's connected to your current network or not) this should stop happening.

PR #22418, opened 2020-02-27
fix: bail early if no printers on the network

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/22409.

Fixes an edge case in the event that no valid printers are known to a given device.

cc @nornagon @deepak1556 @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a potential crash on devices which had not connected any printers to their network.

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -1765,6 +1765,14 @@ void WebContents::OnGetDefaultPrinter(
 
   base::string16 printer_name =
       device_name.empty() ? default_printer : device_name;
+
+  // If there are no valid printers available on the network, we bail.
+  if (printer_name.empty() || !IsDeviceNameValid(printer_name)) {
+    if (print_callback)
+      std::move(print_callback).Run(false, "no valid printers available");
+    return;
+  }
+
   print_settings.SetStringKey(printing::kSettingDeviceName, printer_name);
 
   auto* print_view_manager =
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -1286,7 +1286,7 @@ Returns [`PrinterInfo[]`](structures/printer-info.md)
   `A4`, `A5`, `Legal`, `Letter`, `Tabloid` or an Object containing `height`.
 * `callback` Function (optional)
   * `success` Boolean - Indicates success of the print call.
-  * `failureReason` String - Called back if the print fails; can be `cancelled` or `failed`.
+  * `failureReason` String - Error description called back if the print fails.
 
 Prints window's web page. When `silent` is set to `true`, Electron will pick
 the system's default printer if `deviceName` is empty and the default settings for printing.


shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1763,8 +1763,17 @@ void WebContents::OnGetDefaultPrinter(
     return;
   }
 
-  base::string16 printer_name =
-      device_name.empty() ? default_printer : device_name;
+  base::string16 printer_name;
```
deepak1556: Can you remove https://github.com/electron/electron/blob/master/shell/browser/api/electron_api_web_contents.cc#L1848-L1856 and consolidate this block too,

```
base::string16 printer_name =
      device_name.empty() ? default_printer : device_name;
// If there are no valid printers available on the network, bail.
if (printer_name.empty() || !IsDeviceNameValid(printer_name)) {
   std::move(print_callback).Run(false, "no valid printers available");
   return;
 }
```

The current condition seems to omit device_name always.

thomasdao:
Hi @codebytere, thanks for working on this. Just wanna ask if printer is not available, will Electron still show the print dialog? I tried print with `Preview` on Mac and although I don't have any printer available, the print dialog still shows up for me. I hope Electron would show the dialog, but if it's difficult I think fixing the crash is definitely more important.

codebytere:
@thomasdao it's difficult enough that for now i'm just going to mitigate the crash, but I can spend a bit of time in the next few weeks searching for a more robust solution!

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1765,6 +1765,13 @@ void WebContents::OnGetDefaultPrinter(
 
   base::string16 printer_name =
       device_name.empty() ? default_printer : device_name;
+
+  // If there are no valid printers available on the network, bail.
+  if (printer_name.empty() || !IsDeviceNameValid(printer_name)) {
```
deepak1556: https://github.com/electron/electron/blob/master/shell/browser/api/electron_api_web_contents.cc#L1851-L1856 check is now redundant, only advantage is bailing early which doesn't save much, we can remove that.

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1765,6 +1765,15 @@ void WebContents::OnGetDefaultPrinter(
 
   base::string16 printer_name =
       device_name.empty() ? default_printer : device_name;
+
+  // We set the default to the system's default printer and only update
+  // if at the Chromium level if the user overrides.
+  // If there are no valid printers available on the network, we bail.
+  if (printer_name.empty() || !IsDeviceNameValid(printer_name)) {
+    std::move(print_callback).Run(false, "no valid printers available");
```
zcbenz: `print_callback` might be `null` if user does not pass an callback, should check `if (print_callback)` here.

zcbenz:
The tests are failing on macOS.

shell/browser/api/electron_api_web_contents.cc
codebytere: @deepak1556 hmm - i'm rethinking this in light of the newly failing test. I think there is in fact value in telling the user that _they_ passed in an invalid printer independent of printers being available on the network. One throws an error and the other calls a callback, and i think as a rule user-error should be surfaced early with a true Error and not just a callback.

zcbenz: I'm good with either solution. But we should explicitly document that the `print` throws on invalid input, currently the API design implies that all errors are passed via callback.

deepak1556: I am fine with it as well, was suggesting this because errors can be raised through callbacks.

codebytere: done!

PR #22422, opened 2020-02-27
fix: backport fix for icu crash

nornagon:
#### Description of Change
This backports a fix from Chromium for a crash in ICU. See https://bugs.chromium.org/p/chromium/issues/detail?id=1044570

#### Release Notes

Notes: Fixed an integer overflow crash in ICU (https://crbug.com/1044570).


nornagon:
Mac and WoA failures are flake. Merging.

PR #22449, opened 2020-02-29
feat: add events for spellcheck dictionary downloads

MarshallOfSound:
As in title and notes 😄 Pretty standard wiring, had to patch in a way to attach observers to the dictionaries that the spellcheck service creates but that's pretty minimal.

Notes: Added new `spellcheck-dictionary-*` events on the `session module to help track and debug dictionary initialization


jkleinsc:
This was approved for backport to 9-x-y and 8-x-y by the releases WG.

Issue #20679, opened 2019-10-22
The new audioMuted property should work on WebViews


HaNdTriX:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.0.0
  * 8.0.0
  * 8.0.1
  * 9.0.0-beta.1
  * 9.0.0-beta.2
* **Operating System:**
  * MacOS: 10.14.06

### Expected Behavior

`webview.audioMuted = true` should mute the webview. 
`webview.audioMuted = false` should unmute the webview. 

### Actual Behavior

`webview.audioMuted = true` does nothing but triggers the following warning:

```log
(electron) 'setAudioMuted function' is deprecated and will be removed. Please use 'audioMuted property' instead.
```

### To Reproduce

Check out the following electron fiddle

https://gist.github.com/e80bbbc93d80be9889d8aceb4b8c20a2

codebytere:
@HaNdTriX welp, this does not work in 6, 5, or 4, so it's at least not an issue with the property migration. I'll try to poke at this more.

HaNdTriX:
@codebytere thanks for having a look at this 🙏! For my understanding all "old" setter and getter methods will be deprecated in in the next versions of electron. 

Currently I moved back to `setAudioMuted` as a temporary fix. Nevertheless I can't turn of that misleading deprecating warning electron is triggering:

```
(electron) 'setAudioMuted function' is deprecated and will be removed. Please use 'audioMuted property' instead.
```

codebytere:
@HaNdTriX wait, does the non-property version work? Neither worked when i tested your repro.

HaNdTriX:
`webview.setAudioMuted(true)` works fine but this one triggers a deprecation warning. Tested on `macOS` via 
[Electron Fiddle](https://electronjs.org/fiddle).

KochiyaOcean:
I've noticed that if webview is set as muted, sometimes when a iframe inside webview reloads, it will still play sounds.
It's hard to create a stable reproduce. If I find a way to make it, I'll make a demo and publish it here.

HaNdTriX:
[...]

PR #22485, opened 2020-03-02
fix: properly forward properties to webview

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/20679

Adds handling for properties so that they are properly forwarded to webviews. Tested with [this gist](https://gist.github.com/HaNdTriX/e80bbbc93d80be9889d8aceb4b8c20a2).

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixes some properties not working in webview tags.

### DIFF
--- a/lib/renderer/web-view/web-view-impl.ts
+++ b/lib/renderer/web-view/web-view-impl.ts
@@ -4,7 +4,7 @@ import { ipcRendererInternal } from '@electron/internal/renderer/ipc-renderer-in
 import * as ipcRendererUtils from '@electron/internal/renderer/ipc-renderer-internal-utils'
 import * as guestViewInternal from '@electron/internal/renderer/web-view/guest-view-internal'
 import { WEB_VIEW_CONSTANTS } from '@electron/internal/renderer/web-view/web-view-constants'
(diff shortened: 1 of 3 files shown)


lib/renderer/web-view/web-view-impl.ts
```diff
@@ -249,6 +249,25 @@ export const setupMethods = (WebViewElement: typeof ElectronInternal.WebViewElem
   WebViewElement.prototype.capturePage = async function (...args) {
     return deserialize(await ipcRendererInternal.invoke('ELECTRON_GUEST_VIEW_MANAGER_CAPTURE_PAGE', this.getWebContentsId(), args))
   }
+
+  const createPropertyHandler = function (method: string, type: 'SET' | 'GET') {
```
MarshallOfSound: The enum here seems unnecessary, can we have `createPropertyGetterHandler` and `createPropertySetterHandler` as two separate things.

codebytere: i'm impartial, this was just slightly less code but if you'd prefer that i'll get it updated :)

codebytere: done!

PR #22492, opened 2020-03-03
fix: do not reposition traffic lights when fullscreened

MarshallOfSound:
As in notes

Notes: updating the document title while in fullscreen with custom traffic light positions no longer makes the traffic lights invisible


alfredxing:
Thanks Sam

PR #22493, opened 2020-03-03
build: fix macOS build on 8.x

zcbenz:
Tests on `8-x-y` are failing because it was upgraded to a new SDK while the Chromium version in `8-x-y` was not really ready for it https://github.com/electron/electron/commit/69598ae5c68af53f6b80f023bc8e492b91e85858.

```
process module process.getProcessMemoryInfo() resolves promise successfully with valid data - resolves promise successfully with valid data
/Users/distiller/project/src/electron/spec/api-process-spec.js
AssertionError: expected 0 to be above 0
```

I can't find how the change got into the `8-x-y` branch, seems to be an accidental push. This PR reverts the SDK changes.

Notes: no-notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,12 +65,12 @@ machine-linux-2xlarge: &machine-linux-2xlarge
 
 machine-mac: &machine-mac
   macos:
-    xcode: "11.1.0"
+    xcode: "10.3.0"
 
 machine-mac-large: &machine-mac-large
   resource_class: large
   macos:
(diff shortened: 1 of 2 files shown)


MarshallOfSound:
@zcbenz we need to fix this forward rather than revert the SDK change if possible, we updated the SDK to get more disk space on CI.

We may need to look at other options for disk space if this SDK truly is causing these breakages

zcbenz:
Closed in favor of https://github.com/electron/electron/pull/22496.

PR #22494, opened 2020-03-03
test: give arm machines more time to load spellchecker

zcbenz:
The spellchecker tests are flaky on arm CI.

Notes: no-notes


MarshallOfSound:
Once we land https://github.com/electron/electron/pull/22449 these tests can become much more stable (we can just wait for the dictionary initialized event)

PR #22523, opened 2020-03-04
fix: add patch to fix linux arm build

zcbenz:
Somehow the arm compiler failed to automatically set the function's symbol to `STT_FUNC`, it is probably caused by Chromium upgrading its toolchain and this problem might be gone in future, but anyway explicitly set `.type marl_fiber_swap, %function` can fix the problem for now.

We should remove this patch once it is solved in Chromium https://bugs.chromium.org/p/chromium/issues/detail?id=1058107.

Notes: no-notes

### DIFF
--- a/patches/config.json
+++ b/patches/config.json
@@ -5,5 +5,7 @@
 
   "src/electron/patches/v8":  "src/v8",
 
-  "src/electron/patches/node": "src/third_party/electron_node"
+  "src/electron/patches/node": "src/third_party/electron_node",
+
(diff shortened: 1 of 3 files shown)


gregk-git:
The Chromium issue was fixed, should this change be undone and instead use the Chromium fix?

jkleinsc:
@gregk-git that already happened here: https://github.com/electron/electron/commit/d8737734bff17bccbfc7291f948b705d4c490d74.

Our roller-bot automatically pulls in Chromium updates and when an issue like this is fixed upstream, we delete the patch in Electron.

PR #22524, opened 2020-03-04
test: pass same args to v8_context_snapshot_generator with mksnapshot

zcbenz:
#### Description of Change

The verify mksnapshot test sometimes fails with this error:

```
#
# Fatal error in ../../v8/src/execution/isolate.cc, line 3511
# The Isolate is incompatible with the embedded blob. This is usually caused by incorrect usage of mksnapshot. When generating custom snapshots, embedders must ensure they pass the same flags as during the V8 build process (e.g.: --turbo-instruction-scheduling).
#
```

This PR tries to fix it by passing the same args to v8_context_snapshot_generator with mksnapshot.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/verify-mksnapshot.py
+++ b/script/verify-mksnapshot.py
@@ -34,8 +34,8 @@ def main():
         context_snapshot_path = os.path.join(app_path, context_snapshot)
         gen_binary = get_binary_path('v8_context_snapshot_generator', \
                                     app_path)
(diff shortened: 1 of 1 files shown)


zcbenz:
Yeah you are right, this should not be needed. The snapshot error looks really weird then.

PR #22571, opened 2020-03-06
fix: disallow loading extensions in temp sessions (#22090)

samuelmaddock:
#### Description of Change
Backport of #22090

I removed the docs additions since I assume those don't matter for backports. Extensions docs don't exist in 9-x-y either.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


sentialx:
Does it mean I won't be able to load extensions into incognito sessions? In Chrome it's possible by checking the `Allow in incognito` switch.

samuelmaddock:
> 
> 
> Does it mean I won't be able to load extensions into incognito sessions? In Chrome it's possible by checking the `Allow in incognito` switch.

I believe the current issue is mostly the potential instability of doing so in Electron. As @nornagon mentioned, a `DCHECK()` statement was hit which usually indicates something isn't quite right. I think if someone were to resolve these issues, maintainers might be receptive to supporting extensions in temporary/incognito sessions.

sentialx:
@samuelmaddock Where does the DCHECK trigger?

samuelmaddock:
> 
> 
> @samuelmaddock Where does the DCHECK trigger?

@sentialx I don't know off hand. If you checkout `master` and remove the changes made in #22090, you should be able to reproduce it.

nornagon:
We don't currently support the "allow in incognito" behavior that's present in Chromium, unfortunately. There's some fairly significant machinery required for it.

PR #22588, opened 2020-03-06
chore: don't delete nightly tag after draft

codebytere:
#### Description of Change

Tags on the nightlies repository are  only created at the time a release is published, so if we delete a draft it will always fail to delete the associated ref as it will not exist yet. We should also always be deleting the electron tag regardless of draft deletion status.

cc @MarshallOfSound 

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/script/release/release-artifact-cleanup.js
+++ b/script/release/release-artifact-cleanup.js
@@ -44,7 +44,6 @@ async function deleteDraft (releaseId, targetRepo) {
       repo: targetRepo,
       release_id: parseInt(releaseId, 10)
     })
-    console.log(result)
     if (!result.data.draft) {
       console.log(`${fail} published releases cannot be deleted.`)
       return false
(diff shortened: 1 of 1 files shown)


script/release/release-artifact-cleanup.js
```diff
@@ -85,14 +84,14 @@ async function cleanReleaseArtifacts () {
...
-      if (deletedNightlyDraft) {
-        await Promise.all([
-          deleteTag(args.tag, 'electron'),
-          deleteTag(args.tag, 'nightlies')
-        ])
+        // We only need to delete the Electron tag since the
+        // nightly tag is only created at publish-time.
+        await deleteTag(args.tag, 'electron')
+      } catch (err) {
+        console.log(`${fail} Couldn't complete cleanup for nightly: `, err)
```
ckerr: I'm not sure this error handling will work as intended?

Neither `deleteDraft()` nor `deleteTag()` will throw errors because they wrap everything in internal try-catch blocks and return a success flag, so I think (1) `deleteTag()` will be called even if `deleteDraft()` fails and (2) the `catch` clause will never be reached

ckerr: For example, previously we didn't try to delete tags if `deleteDraft()` failed, but the new code would try to delete the tag `'electron'` tag anyway

codebytere: that's correct @ckerr 
> new code would try to delete the tag 'electron' tag anyway

that's is what's intended!

PR #22589, opened 2020-03-07
docs: fix error china mirror url address

xland:
#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:


malept:
`cdn.npm.taobao.org` is the correct address, however there seems to be a delay in their mirroring new versions of the Electron prebuilt binary. I would suggest contacting the maintainers to see what they can do about that.

If you'll notice below that section, it does mention `npm.taobao.org` and the way to configure for that non-CDN mirror.

PR #22592, opened 2020-03-07
fix(extensions): add more properties to port.sender.tab

sentialx:
Currently, the `Port.sender.tab` object has only 1 property `id`. Some extensions like Dark Reader or `uBlock Origin` require also other properties like `url` or `title`. This PR brings support for `url`, `title`, `audible` properties.

Ref #19447 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/shell/browser/extensions/electron_messaging_delegate.cc
+++ b/shell/browser/extensions/electron_messaging_delegate.cc
@@ -20,10 +20,9 @@
 #include "extensions/browser/pref_names.h"
 #include "extensions/common/api/messaging/port_id.h"
 #include "extensions/common/extension.h"
+#include "shell/browser/api/electron_api_web_contents.h"
 #include "ui/gfx/native_widget_types.h"
 #include "url/gurl.h"
-
-#include "shell/browser/api/electron_api_web_contents.h"
 
 namespace extensions {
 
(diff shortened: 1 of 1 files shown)


shell/browser/extensions/electron_messaging_delegate.cc
```diff
@@ -48,6 +47,13 @@ ElectronMessagingDelegate::MaybeGetTabInfo(content::WebContents* web_contents) {
...
       tab->SetWithoutPathExpansion(
           "id", std::make_unique<base::Value>(api_contents->ID()));
+      tab->SetWithoutPathExpansion(
+          "url", std::make_unique<base::Value>(api_contents->GetURL()));
+      tab->SetWithoutPathExpansion(
+          "title", std::make_unique<base::Value>(api_contents->GetTitle()));
+      tab->SetWithoutPathExpansion(
+          "audible",
+          std::make_unique<base::Value>(api_contents->IsCurrentlyAudible()));
       return tab;
```
sentialx: Still, it doesn't contain all of the `tabs.Tab` properties, since they are rather related to the UI. I suggest sending a sync IPC message `extensions-get-tab` to allow users to add their own properties to `tabs.Tab` object, for example `active`, `highlighted` etc.

shell/browser/extensions/electron_messaging_delegate.cc
```diff
@@ -48,6 +47,13 @@ ElectronMessagingDelegate::MaybeGetTabInfo(content::WebContents* web_contents) {
       auto tab = std::make_unique<base::DictionaryValue>();
       tab->SetWithoutPathExpansion(
           "id", std::make_unique<base::Value>(api_contents->ID()));
+      tab->SetWithoutPathExpansion(
+          "url", std::make_unique<base::Value>(api_contents->GetURL()));
```
nornagon:
```suggestion
          "url", std::make_unique<base::Value>(api_contents->GetURL().spec()));
```

sentialx:
Could it be backported to 9-x-y with some super commands like /trop run backport or should I do it manually?

Issue #8891, opened 2017-03-10
Cookie handling in net module


arantes555:
* Electron version: 1.6.2
* Operating system: macOS 10.12

### Expected behavior

Requests sent with the net module should behave one of two ways regarding cookies.
- Either not care about cookies at all, and neither store them nor send them back automatically.
- Or, fully handle cookies, store them when getting a `set-cookie` header, and send them back when necessary, automatically (with the possibility of an option to disable this behaviour)

### Actual behavior

Requests sent with the net module store cookies in the Session used by the request when getting a `set-cookie` header, but do not send them back afterwards.

### How to reproduce

```bash
git clone https://github.com/arantes555/electron-net-cookies-testcase.git && cd electron-net-cookies-testcase
npm i
npm start
```
output: 
```
Server started
REQUEST-HEADERS { 'Accept-Encoding': 'gzip, deflate',
  'Accept-Language': 'fr',
  'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/56.0.2924.87 Electron/1.6.2 Safari/537.36' }
Response SET-COOKIE: undefined
STORED COOKIES: []
REQUEST-HEADERS { 'Accept-Encoding': 'gzip, deflate',
  'Accept-Language': 'fr',
  'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/56.0.2924.87 Electron/1.6.2 Safari/537.36' }
Response SET-COOKIE: a=1,b=1
STORED COOKIES: [{"name":"a","value":"1","domain":"localhost"},{"name":"b","value":"1","domain":"localhost"}]
REQUEST-HEADERS { 'Accept-Encoding': 'gzip, deflate',
  'Accept-Language': 'fr',
  'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/56.0.2924.87 Electron/1.6.2 Safari/537.36' }
Response SET-COOKIE: a=1,b=1
```

Btw, if I'm reporting multiple issues with the net module these days, it's because I'm working on https://github.com/arantes555/electron-fetch , if you wanna take a look :)

deepak1556:
[...]

PR #22596, opened 2020-03-08
fix: don't clobber cookie header with session cookies

### LINKED ISSUE #22541: Net request replaces cookie header with session cookies
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * From v7.1.14 to v9.0.0-beta.5
* **Operating System:**
  * macOS 10.15.3
  * v6.1.9

### Expected Behavior
A `net` request that sets a cookie using `request.setHeader` should include the specified header.

### Actual Behavior
Instead, when cookies from the session cookie jar are present, they are sent instead and the specified header is ignored.

### To Reproduce
Run this code:
```js
const electron = require('electron');
electron.app.on('ready', async () => {
  await makeRequest();
  // A long timeout is required for it to be reproducible when running this script on the first try, in a clean session
  setTimeout(async () => {
    console.log(await electron.session.defaultSession.cookies.get({}));
    await makeRequest();
  }, 10000);
});

function makeRequest() {
    const request = electron.net.request({ url: 'https://www.twitter.com' });
    request.setHeader('cookie', 'foo=bar');
    request.on('close', resolve);
    request.end();
  });
}
```

### Screenshots
When running the above code, all requests after the initial one shows a cookie header with this value, which reflects the value that is present in the session's cookies.
image

### Additional Information
* To see the incorrect headers, you must inspect the network request in a proxy such as Fiddler.
* This was previously reported in #8223, fixed in #8258, but now seems to have been regressed.
* Also #8891 seems to be possibly implemented by accident now?

NilSet:
Fixes #22541

cc @deepak1556 @nornagon 

#### Description of Change
If a cookie header was added to a net.request, do not send cookies from
the session's cookie jar, and instead respect the explicitly passed in
header.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Stopped overwriting explicit cookie header value with values from the session cookies when making requests using `net`.

### DIFF
--- a/spec-main/api-net-spec.ts
+++ b/spec-main/api-net-spec.ts
@@ -592,8 +592,9 @@ describe('net module', () => {
           url: `${serverUrl}`,
           name: 'test',
           value: '11111',
-          expirationDate: 0
-        }).then(() => { // resolved
+          expirationDate: new Date().getTime() + 30000
+        }).then(() => customSession.cookies.flushStore()
+        ).then(() => { // resolved
           const urlRequest = net.request({
             method: 'GET',
             url: serverUrl,
--- a/shell/browser/api/electron_api_url_loader.cc
+++ b/shell/browser/api/electron_api_url_loader.cc
@@ -12,11 +12,13 @@
 #include <vector>
 
 #include "base/containers/id_map.h"
+#include "base/strings/string_util.h"
 #include "gin/handle.h"
 #include "gin/object_template_builder.h"
 #include "gin/wrappable.h"
 #include "mojo/public/cpp/bindings/remote.h"
 #include "mojo/public/cpp/system/data_pipe_producer.h"
+#include "net/base/load_flags.h"
 #include "services/network/public/cpp/resource_request.h"
 #include "services/network/public/cpp/simple_url_loader.h"
 #include "services/network/public/mojom/chunked_data_pipe_getter.mojom.h"
@@ -353,6 +355,12 @@ gin_helper::WrappableBase* SimpleURLLoaderWrapper::New(gin::Arguments* args) {
           !net::HttpUtil::IsValidHeaderValue(it.second)) {
         args->ThrowTypeError("Invalid header name or value");
         return nullptr;
+      }
+      // If we explicitly set a cookie header, do not clobber it with cookies
+      // from the session
+      // TODO(nilset): find other late added headers to allow overriding
+      if (base::LowerCaseEqualsASCII(it.first, "cookie")) {
+        request->load_flags |= net::LOAD_DO_NOT_SEND_COOKIES;
       }
       request->headers.SetHeader(it.first, it.second);
     }


NilSet:
The strange thing is that there already appears to be a passing unit test for this. I need to figure out how to make it fail without my change.

NilSet:
Also, is master broken? Because a number of checks are failing in ways I cannot rationally attribute to my changes.

deepak1556:
@NilSet please rebase on latest master to fix those unrelated errors.

NilSet:
@nornagon test has been updated

NilSet:
Remaining question I have is how this behavior should be documented. I believe reading from the cookie store itself is currently undocumented. Can I just leave that work for #8891?

deepak1556:
Its not an easy problem, depends on how we want the net module to behave.

1) If we want it to behave like fetch in the renderer, then web requests api only rely on the cookies from the cookie store, there is a uniform way for webpages to access and edit it. Hence the requesting module doesn't have to care about them.

Good thing we also provide apis for users to do it in the main process as well, so we can let the net module behave like where it always loads from the store and ignore anything set inline in the request api.

2) If we want to behave like a node request module, then we should respect the cookie line set via the headers and send only that, with nothing from the cookie store.

We are trying to behave like 2) now but question is are there use cases where the cookie sent via the header line will want to be reused for successive requests on the same session, in which case we have to store to them and resend but cookie store for a session is the same across renderer and main, so we can't just retain these on successive requests, we will be sending everything thats available in the store.


The current solution in this PR that was previously made avoids any kind of persistence and hence reduced the problem surface but the underlying problem still remains and whatever change is made for it should happen in a separate PR, as it will be breaking.

NilSet:
> but question is are there use cases where the cookie sent via the header line will want to be reused for successive requests on the same session

I don't know who would expect this. Persisted cookies are set by the server, not the client. I don't expect any of my request options to be sticky.

deepak1556:
@NilSet the behavior is expected by users in https://github.com/electron/electron/issues/8891

@MarshallOfSound suggested an idea where we can have an option on the net module that either respects the cookie line for requests or use cookies from the cookie store, this seems like an ideal behavior from user perspective as the control is given to them in terms of where the cookies come from

`net.request({ useCookieStore: true })`

spec-main/api-net-spec.ts
```diff
@@ -592,8 +592,9 @@ describe('net module', () => {
           url: `${serverUrl}`,
           name: 'test',
           value: '11111',
-          expirationDate: 0
-        }).then(() => { // resolved
+          expirationDate: new Date().getTime() + 30000
```
nornagon: Leaving out the expiration date should be fine since https://github.com/electron/electron/pull/21454

spec-main/api-net-spec.ts
```diff
@@ -592,8 +592,9 @@ describe('net module', () => {
...
+        }).then(() => customSession.cookies.flushStore()
```
nornagon: This is kind of awkward syntax, if you feel like it it would be great to convert this test to use `async`! But not needed for the PR.

zcbenz:
Closing in favor of https://github.com/electron/electron/pull/22704.

Issue #22554, opened 2020-03-05
app.focus() not working


devinfoley:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

It would appear that `app.focus()` stopped working in Electron 6, and remains broken in Electron 8.

* **Electron Version:**
  * 8.0.3
* **Operating System:**
  * macOS
* **Last Known Working Electron version:**
  * 5.0.13

### Expected Behavior
`app.focus()` should bring the application to the foreground.

### Actual Behavior
`app.focus()` does nothing.

### To Reproduce

Fiddle: https://gist.github.com/a772265b312329e3f3bc1feb7b6aff7d

1) Start this fiddle.
2) As soon as the window pops up, select another application so that Electron is in the background.
3) Within 5 seconds, the app will attempt to focus itself to no avail.

codebytere:
@devinfoley this appears to have been intentional: https://github.com/electron/electron/pull/16400

MarshallOfSound:
The use case here is for things like dragging files from app A to app B.  App B needs to be able to focus itself after receiving a file, currently it cant.

mifi:
Will this make it into 8.x version?

PR #22612, opened 2020-03-09
feat: add force option to app.focus()

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/22554. 

Adds a new optional `force` parameter to `app.focus()`, which allows for apps to make themselves active even if other apps are active.

cc @MarshallOfSound @nornagon @ckerr 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Add a new `force` parameter to `app.focus()` on macOS to allow apps to forcefully take focus.

### DIFF
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -554,10 +554,15 @@ Returns `Promise<void>` - fulfilled when Electron is initialized.
 May be used as a convenient alternative to checking `app.isReady()`
 and subscribing to the `ready` event if the app is not ready yet.
 
-### `app.focus()`
+### `app.focus([force])`
+
+* `force` String (optional) _macOS_ - Make the receiver the active app even if another app is
+currently active.
 
 On Linux, focuses on the first visible window. On macOS, makes the application
 the active app. On Windows, focuses on the application's first window.
+
+On macOS, you should seek to use the `force` option as sparingly as possible.
 
 ### `app.hide()` _macOS_
 
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -554,15 +554,16 @@ Returns `Promise<void>` - fulfilled when Electron is initialized.
 May be used as a convenient alternative to checking `app.isReady()`
 and subscribing to the `ready` event if the app is not ready yet.
 
-### `app.focus([force])`
+### `app.focus([options])`
 
-* `force` String (optional) _macOS_ - Make the receiver the active app even if another app is
-currently active.
+* `options` Object (optional)
+  * `steal` Boolean _macOS_ - Make the receiver the active app even if another app is
+  currently active.
 
 On Linux, focuses on the first visible window. On macOS, makes the application
 the active app. On Windows, focuses on the application's first window.
 
-On macOS, you should seek to use the `force` option as sparingly as possible.
+You should seek to use the `steal` option as sparingly as possible.
 
 ### `app.hide()` _macOS_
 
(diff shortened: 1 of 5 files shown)


docs/api/app.md
```diff
@@ -554,11 +554,16 @@ Returns `Promise<void>` - fulfilled when Electron is initialized.
 May be used as a convenient alternative to checking `app.isReady()`
 and subscribing to the `ready` event if the app is not ready yet.
 
-### `app.focus()`
+### `app.focus([force])`
+
+* `force` String (optional) _macOS_ - Make the receiver the active app even if another app is
```
MarshallOfSound: Might seem overkill but can we make this. `app.focus({ force: true })` better readability and expandability

codebytere: yeah, we can avoid the boolean trap here :)

shell/browser/browser.h
```diff
@@ -57,7 +57,7 @@ class Browser : public WindowListObserver {
   void Shutdown();
 
   // Focus the application.
-  void Focus();
+  void Focus(gin_helper::Arguments* args);
```
ckerr: We should handle the function declaration for `Browser::Focus()` the same in both `shell/browser/browser_win.cc` and `shell/browser/browser_linux.cc`. Right now, the latter is not declaring the `gin_helper::Arguments*` parameter

shell/browser/browser_win.cc
```diff
@@ -163,7 +163,7 @@ Browser::UserTask::UserTask() = default;
 Browser::UserTask::UserTask(const UserTask&) = default;
 Browser::UserTask::~UserTask() = default;
 
-void Browser::Focus() {
+void Browser::Focus(gin_helper::Arguments* args) {
```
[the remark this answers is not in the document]
codebytere: i think we could do that, but we don't do it as a rule anywhere else in the codebase at the moment so I think i'd prefer to either do it in a follow-up or discuss switching to that convention on a larger scale?

docs/api/app.md
```diff
@@ -554,11 +554,17 @@ Returns `Promise<void>` - fulfilled when Electron is initialized.
 May be used as a convenient alternative to checking `app.isReady()`
 and subscribing to the `ready` event if the app is not ready yet.
 
-### `app.focus()`
+### `app.focus([opts])`
+
+* `options` Object (optional)
+  * `force` Boolean _macOS_ - Make the receiver the active app even if another app is
+  currently active.
```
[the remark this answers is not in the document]
codebytere: i'd be fine with this :)

docs/api/app.md
```diff
@@ -554,11 +554,17 @@ Returns `Promise<void>` - fulfilled when Electron is initialized.
 May be used as a convenient alternative to checking `app.isReady()`
 and subscribing to the `ready` event if the app is not ready yet.
 
-### `app.focus()`
+### `app.focus([opts])`
```
zcbenz: `opts` => `options` to match the description bellow.

shell/browser/browser.h
ckerr: Fixed in 928979e

sofianguy:
this was approved by the Releases WG on 5/6. removing pending-vote label

Goldiserv:
Solution's a bit hacky but works for me:

ipcMain.handle('focus', () => {
      // mainWindow?.moveTop(); // doesn't work
      mainWindow?.setAlwaysOnTop(true);
      mainWindow?.setAlwaysOnTop(false);
      mainWindow?.focus();
});

PR #22616, opened 2020-03-09
fix: remove bad usages of for-in and guard against it

### LINKED ISSUE #22615: Adding any Object prototype throws an error
* **Electron Version:**
  v8.1.0
* **Operating System:**
  macOS 10.15.3
  N/A

### Expected Behavior
I expect to be able to add an Object prototype without Electron throwing an error.

### Actual Behavior
Electron throws an error on startup, then otherwise seems to work fine.

### To Reproduce
Electron fiddle: https://gist.github.com/d3238b39f2b33493e0d7dbfc4c3b6afe

### Screenshots
A screenshot by another user already exists here (different code than in the Fiddle, though) on this old Stackoverflow Question: https://stackoverflow.com/questions/54179169/object-prototype-break-electron

MarshallOfSound:
`for (... in ...)` is a bad idea as it will pick up everything on the prototype as well not just keys on that object.  Where appropriate I've replaced it with `for (const key of Object.keys...)` and enabled the eslint rule that guards against it so that future occurrences to not slip through.

Fixes #22615

Notes: fixed issue where mutating the global `Object` prototype could cause internal Electron logic to throw errors

### DIFF
--- a/lib/browser/guest-view-manager.js
+++ b/lib/browser/guest-view-manager.js
@@ -74,7 +74,7 @@ const createGuest = function (embedder, params) {
 
   // Clear the guest from map when it is destroyed.
   guest.once('destroyed', () => {
-    if (guestInstanceId in guestInstances) {
+    if (Object.keys(guestInstances).includes(guestInstanceId)) {
       detachGuest(embedder, guestInstanceId)
     }
   })
@@ -288,7 +288,7 @@ const watchEmbedder = function (embedder) {
 
   // Forward embedder window visiblity change events to guest
   const onVisibilityChange = function (visibilityState) {
-    for (const guestInstanceId in guestInstances) {
+    for (const guestInstanceId of Object.keys(guestInstances)) {
       const guestInstance = guestInstances[guestInstanceId]
       guestInstance.visibilityState = visibilityState
       if (guestInstance.embedder === embedder) {
@@ -302,7 +302,7 @@ const watchEmbedder = function (embedder) {
     // Usually the guestInstances is cleared when guest is destroyed, but it
     // may happen that the embedder gets manually destroyed earlier than guest,
     // and the embedder will be invalid in the usual code path.
-    for (const guestInstanceId in guestInstances) {
+    for (const guestInstanceId of Object.keys(guestInstances)) {
       const guestInstance = guestInstances[guestInstanceId]
       if (guestInstance.embedder === embedder) {
         detachGuest(embedder, parseInt(guestInstanceId))
--- a/lib/browser/guest-view-manager.js
+++ b/lib/browser/guest-view-manager.js
@@ -74,7 +74,7 @@ const createGuest = function (embedder, params) {
 
   // Clear the guest from map when it is destroyed.
   guest.once('destroyed', () => {
-    if (Object.keys(guestInstances).includes(guestInstanceId)) {
+    if (guestInstances.hasOwnProperty(guestInstanceId)) {
       detachGuest(embedder, guestInstanceId)
     }
   })
--- a/lib/browser/guest-view-manager.js
+++ b/lib/browser/guest-view-manager.js
@@ -74,7 +74,7 @@ const createGuest = function (embedder, params) {
 
   // Clear the guest from map when it is destroyed.
   guest.once('destroyed', () => {
-    if (guestInstances.hasOwnProperty(guestInstanceId)) {
+    if (Object.prototype.hasOwnProperty.call(guestInstances, guestInstanceId)) {
       detachGuest(embedder, guestInstanceId)
     }
   })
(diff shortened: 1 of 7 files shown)


lib/browser/utils.ts
```diff
@@ -18,7 +18,7 @@ export function createLazyInstance (
 ) {
   let lazyModule: Object
   const module: any = {}
-  for (const method in (holder as any).prototype) {
+  for (const method in (holder as any).prototype) { // eslint-disable-line guard-for-in
```
nornagon: hmm is this correct? we should probably guard with `hasOwnProperty` here...

lib/renderer/api/remote.js
```diff
@@ -72,7 +72,7 @@ function wrapArgs (args, visited = new Set()) {
         members: []
       }
       visited.add(value)
-      for (const prop in value) {
+      for (const prop in value) { // eslint-disable-line guard-for-in
```
nornagon: I think this could be `Object.keys`?

lib/browser/utils.ts
MarshallOfSound: We want to lazily access inherited prototypes here, so I think `in` is right 🤔

lib/renderer/api/remote.js
MarshallOfSound: Maybe, but I'm scared it will result in a side-effect breaking change.  We can flip it to `Object.keys` in a major version or when we extract `remote`

lib/browser/guest-view-manager.js
```diff
@@ -74,7 +74,7 @@ const createGuest = function (embedder, params) {
 
   // Clear the guest from map when it is destroyed.
   guest.once('destroyed', () => {
-    if (guestInstanceId in guestInstances) {
+    if (Object.keys(guestInstances).includes(guestInstanceId)) {
```
samuelmaddock:
```suggestion
    if (guestInstances.hasOwnProperty(guestInstanceId)) {
```

lib/renderer/api/remote.js
```diff
@@ -219,7 +219,7 @@ function metaToValue (meta) {
     exception: () => { throw metaToError(meta.value) }
   }
 
-  if (meta.type in types) {
+  if (Object.keys(types).includes(meta.type)) {
```
samuelmaddock:
```suggestion
  if (types.hasOwnProperty(meta.type)) {
```

lib/browser/guest-view-manager.js
```diff
@@ -74,7 +74,7 @@ const createGuest = function (embedder, params) {
 
   // Clear the guest from map when it is destroyed.
   guest.once('destroyed', () => {
-    if (Object.keys(guestInstances).includes(guestInstanceId)) {
+    if (guestInstances.hasOwnProperty(guestInstanceId)) {
```
nornagon: I have sad news, this breaks if `guestInstances` has a property named `"hasOwnProperty"`. It is, I think, _never_ correct to call `x.hasOwnProperty(y)`, always always use `Object.prototype.hasOwnProperty.call(x, y)`.
```suggestion
    if (Object.prototype.hasOwnProperty.call(guestInstances, guestInstanceId)) {
```

lib/renderer/api/remote.js
```diff
@@ -219,7 +219,7 @@ function metaToValue (meta) {
     exception: () => { throw metaToError(meta.value) }
   }
 
-  if (Object.keys(types).includes(meta.type)) {
+  if (types.hasOwnProperty(meta.type)) {
```
nornagon:
```suggestion
  if (Object.prototype.hasOwnProperty(types, meta.type)) {
```

lib/browser/guest-view-manager.js
MarshallOfSound: Ah good catch 👍

ckerr:
I'm :heavy_check_mark: on putting this in new releases, but before we backport this -- is this fixing some known existing problems?

I _think_ some of these objects being looped through are coming from users, e.g. extraHeaders is coming from the 'headers' property of the options object passed to ClientRequest's constructor. It's possible that some oddball setup out in the wild works under `for...in` but not under `for..of Object.keys()`.

Which is why I ask if this is fixing any known bugs. If this is a cleanliness PR, I don't see the upside in backporting.

ckerr:
...or I could've just, you know, read the description and click through to #22615 to see what this fixes.

Perhaps a more limited scope of changes in the backports?

MarshallOfSound:
@ckerr Sending properties of the `headers` object prototype through would also be considered a bug (hence the change instead of the lint-ignore).  These cases were carefully chosen as to whether to update the usage or ignore the lint rule.  Things like `remote` were left alone out of fear of side-effect breakages and will be dealt with later, the `headers` object though it is never intended or expected behavior to send a stringified prototype function as a header so I switched it to not do that any more.

Issue #22528, opened 2020-03-04
Fail Browserwindow creation with wrong x/y


DrNiels:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description
I create a BrowserWindow and want to define its position via x and y. Due to some computations, the numbers become float and I insert them as option parameters in the constructor of BrowserWindow.
As the numbers are not int, they are not as required and the provided coordinates are ignored. I was searching for quite some time to detect that issue.
When providing the same coordinates via setBounds, the execution fails with an error message "TypeError: Error processing argument at index 0, conversion failure from ..."

### Proposed Solution
When a BrowserWindow is created with incorrect x/y values, throw the same error that would come from setBounds. Thus, a developer can immediately detect that the values are not fitting and adjust accordingly

### Alternatives Considered
There is no real alternative. In both cases, I need to adjust my code. Only the search for the error is significantly faster if an error is thrown.

codebytere:
We require the template to be filled out on all new issues and pull requests. We do this so that we can be certain we have all the information we need to address your submission efficiently. This allows the maintainers to spend more time fixing bugs, implementing enhancements, and reviewing and merging pull requests. Specifically, there is no information about platform, version of Electron, or a reproducible code snippet we might run to replicate the problem.

We will be able to more closely look at your issue once you do so. This issue will be closed in 10 days if the above is not addressed.

DrNiels:
* **Electron Version:** 8.0.2
* **Operating System:** Windows 10 (1909)

A code snippet could be:
````
let window = new BrowserWindow({x: 0.5, y: 0.5}); // Working fine, just ignoring x/y
window.setBounds({x: 0.5, y: 0.5}); // Throwing exception
````

I considered this more of a feature request, so I used the corresponding template and filled it out completely (except for "Additional Information"). Maybe you want to add these fields to the template?

codebytere:
@DrNiels there's a fully separate template for feature requests, in fact - please fill that out instead.

DrNiels:
[...]

PR #22620, opened 2020-03-09
fix: throw an error if bw width/height aren't ints

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/22528.

We should be typechecking a little better for incorrect types - this would mysteriously be ignored instead of throwing a descriptive error, which it now does.

cc @MarshallOfSound @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where incorrect width and height values were ignore and didn't properly error.

### DIFF
--- a/shell/browser/api/electron_api_top_level_window.cc
+++ b/shell/browser/api/electron_api_top_level_window.cc
@@ -92,15 +92,22 @@ TopLevelWindow::TopLevelWindow(v8::Isolate* isolate,
   }
 #endif
 
+#if defined(TOOLKIT_VIEWS)
+  // Sets the window icon.
+  gin::Handle<NativeImage> icon;
+  options.Get(options::kIcon, &icon);
+#endif
+
   // Creates NativeWindow.
   window_.reset(NativeWindow::Create(
       options, parent.IsEmpty() ? nullptr : parent->window_.get()));
   window_->AddObserver(this);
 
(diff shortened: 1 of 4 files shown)


shell/common/gin_helper/dictionary.h
```diff
@@ -134,6 +134,8 @@ class Dictionary : public gin::Dictionary {
 
   bool IsEmpty() const { return isolate() == nullptr || GetHandle().IsEmpty(); }
 
+  v8::Isolate* GetIsolate() const { return isolate(); }
```
zcbenz: I think users can just use `options.isolate()`? All `gin::Dictionary` members should be visible to users.

codebytere:
@MarshallOfSound hmm - this is an interesting one actually. Before it would just fail, so i think this is conceptually perceivable as a fix, but I do see your point in that it's breaking. Shall we perhaps discuss it more in the Releases WG mtg?

ckerr:
Weighing in here b/c the Releases mtg was deferred -- you are both right. @codebytere is correct that this fixes a bug of Electron silently discarding bad input, and @MarshallOfSound is right that this would throw an error were one wasn't thrown before.

This is a bugfix; but since it also increases the severity of Electron's response to bad input, it is also a breaking change

codebytere:
Closing this for now to consider a better approach, perhaps through gin.

PR #22643, opened 2020-03-10
chore: make converter errors descriptive with a subject to make it easier to debug issues

MarshallOfSound:
This is kind of an experiment / seeking feedback on this idea.  Currently if you call an Electron method with invalid arguments you get the super helper (not really)

> Failed to convert argument at index {n}

This adjusts a function template so that converters can throw errors.  Potential problems:

* Converters have never thrown errors before, this might have unintended side effects in our code base
* Folks may be depending on the error message for things like tests or log filtering

Notes: no-notes

### DIFF
--- a/shell/common/gin_converters/gurl_converter.h
+++ b/shell/common/gin_converters/gurl_converter.h
@@ -8,6 +8,7 @@
 #include <string>
 
 #include "gin/converter.h"
+#include "shell/common/gin_helper/error_thrower.h"
 #include "url/gurl.h"
 
 namespace gin {
@@ -25,6 +26,8 @@ struct Converter<GURL> {
       *out = GURL(url);
       return true;
     } else {
+      gin_helper::ErrorThrower(isolate).ThrowConverterError(
+          "Attempted to convert a non-string value to a URL", val);
       return false;
     }
   }
--- a/shell/common/gin_helper/error_thrower.cc
+++ b/shell/common/gin_helper/error_thrower.cc
@@ -5,6 +5,7 @@
 #include "shell/common/gin_helper/error_thrower.h"
 
 #include "gin/converter.h"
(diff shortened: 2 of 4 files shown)


zcbenz:
This is going to break things, for example when the argument may have multiple types:
https://github.com/electron/electron/blob/master/shell/common/gin_converters/blink_converter.cc#L177-L181
https://github.com/electron/electron/blob/master/shell/common/api/electron_api_native_image.cc#L542-L558

zcbenz:
I think the correct way to handle this is to improve `gin::Arguments`.

Currently `gin::Arguments` already prints the type of the argument:

```c++
"Error processing argument at index %d, conversion failure from %s"
```

we can make it better by printing the target type:

```c++
void Arguments::ThrowError(const char* target_type) const {
  return ThrowTypeError(base::StringPrintf(
      "Error processing argument at index %d, conversion failure from %s to %s",
      ...,
      target_type);
```

We can achieve this by optionally adding type name for converters:

```c++
template <>
struct Converter<GURL> {
  static const char* type_name = "url";
  static v8::Local<v8::Value> ToV8(v8::Isolate* isolate, const GURL& val);
};
```

and then pass the target type name in `ArgumentHolder` when parsing arguments:

```c++
    ok = GetNextArgument(args, create_flags, index == 0, &value);
    if (!ok) {
      args->ThrowError(Converter<ArgType>::type_name);
    }
```

Since we have our own forks of `gin_helper/argument.h` and `gin_helper/function_template.h`, this only needs minimal change inside Chromium.

This requires some C++ hacks as we want to keep compatibility with existing converters that do not specify the type's name.

I can give this approach a try if you think it is something worth trying.

zcbenz:
@MarshallOfSound Do you mind closing this in favor of #22668?

Issue #22175, opened 2020-02-12
'webRequest' or 'webRequestBlocking' extension permissions prevents loading


samuelmaddock:
relates to #19447

### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v9.0.0-beta.1
* **Operating System:**
  * Windows 10.0.18362 Build 18362 64-bit
* **Last Known Working Electron version:**
  *  N/A

### Expected Behavior

When an extension is loaded using `Session.loadExtension` with `webRequest` or `webRequestBlocking` permissions, no effect should be taken since these are currently unsupported Chrome APIs.

### Actual Behavior

Loading an extension with `webRequest` or `webRequestBlocking` permissions causes all app network events to hang without resolving.

Adding a delay to calling `Window.loadURL` can fix the issue (seems like a race condition).

### To Reproduce

This is likely the fastest way to test this issue:

1. Modify `spec-main/fixtures/extensions/content-script/manifest.json` to include the following:
```json
"permissions": ["webRequest", "webRequestBlocking", "<all_urls>"]
```
2. Re-run `spec-main/extensions-spec.ts` tests and it should prevent loading any content in the window.

Otherwise you can try this repo:
https://github.com/samuelmaddock/electron-extensions-loading-blocked-bug

### Screenshots

App not loading after calling `win.loadURL('https://www.electronjs.org/')`
![electron_2020-02-12_18-16-08aZeL](https://user-images.githubusercontent.com/1656324/74386393-c78ba280-4dc3-11ea-951e-f7112e6c447b.png)

samuelmaddock:
Closing as this seems to be fixes by #22655

FeeshDev:
I am currently encountering the exact same issue in Electron v12.0.2

PR #22655, opened 2020-03-11
fix(extensions): implement missing web_request hooks

sentialx:
#### Description of Change

Currently the `chrome.webRequest` listeners don't do anything.

This PR adds missing `web_request_api` hooks required for intercepting requests using `chrome.webRequest`.

Ref #19447 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed `chrome.webRequest` API not intercepting any requests.


shell/browser/electron_browser_client.cc
```diff
@@ -1383,7 +1429,8 @@ bool ElectronBrowserClient::WillCreateURLLoaderFactory(
 
   if (bypass_redirect_checks)
     *bypass_redirect_checks = true;
-  return true;
+
+  return use_proxy;
```
sentialx: I'm not sure about this line

sentialx:
@nornagon @zcbenz Could you give me some advice on how to fix the web socket unit test?

sentialx:
It's hitting a DCHECK here: https://source.chromium.org/chromium/chromium/src/+/master:extensions/browser/api/web_request/web_request_api.cc;l=827?originalUrl=https:%2F%2Fcs.chromium.org%2F

sentialx:
My PR is ready for review.

sentialx:
@jkleinsc I think I've reverted the rebase, I can't see those commits anymore

shell/browser/electron_browser_client.cc
```diff
@@ -1350,6 +1376,19 @@ bool ElectronBrowserClient::WillCreateURLLoaderFactory(
...
+  auto* web_request_api =
+      extensions::BrowserContextKeyedAPIFactory<extensions::WebRequestAPI>::Get(
+          browser_context);
+
+  // NOTE: Some unit test environments do not initialize
+  // BrowserContextKeyedAPI factories for e.g. WebRequest.
+  if (web_request_api) {
+    web_request_api->MaybeProxyURLLoaderFactory(
+        browser_context, frame_host, render_process_id, type,
+        std::move(navigation_id), factory_receiver, header_client);
```
nornagon, line 1389: It seems like you'd need to `return` here, otherwise we'd try to also pass the request through Electron's `webRequest` API, which I don't think would work. Not least, it would result in double-`std::move`ing of `navigation_id`.

sentialx: Well, I was thinking of keeping support for both Electron's `webRequest` and `chrome.webRequest`.

nornagon: I don't know, but the code as it is does not do that correctly :)

sentialx: I assume that this would be tricky, so what I propose for now is that registering any `chrome.webRequest` would override the Electron's `webRequest` module. But I think that's not the behavior we want?

shell/browser/electron_browser_client.cc
```diff
@@ -1377,16 +1377,26 @@ bool ElectronBrowserClient::WillCreateURLLoaderFactory(
...
+        web_request_api->MaybeProxyURLLoaderFactory(
+            browser_context, frame_host, render_process_id, type, navigation_id,
+            factory_receiver, header_client);
+
+    if (bypass_redirect_checks)
+      *bypass_redirect_checks = use_proxy_for_web_request;
+    use_proxy |= use_proxy_for_web_request;
+
+    if (use_proxy)
+      return true;
```
nornagon, line 1399:
```suggestion
    if (use_proxy_for_web_request)
      return true;
```

shell/browser/electron_browser_client.cc
```diff
@@ -1313,6 +1315,15 @@ bool ElectronBrowserClient::WillInterceptWebSocket(
   if (!web_request.get())
     return false;
 
+#if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
+  const auto* web_request_api =
+      extensions::BrowserContextKeyedAPIFactory<extensions::WebRequestAPI>::Get(
+          browser_context);
+
+  if (web_request_api)
+    return web_request->HasListener() || web_request_api->MayHaveProxies();
```
nornagon, line 1324:
```suggestion
  if (web_request_api)
    return web_request->HasListener() || web_request_api->MayHaveProxies() || web_request->HasListener();
```

?

It's possible that the `WebRequestAPI` is available but doesn't have a listener, but electron's `api::WebRequest` does have a listener.

nornagon: oh nevermind i'm blind

shell/browser/electron_browser_client.cc
```diff
@@ -1327,6 +1338,21 @@ void ElectronBrowserClient::CreateWebSocket(
...
+
+#if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
+  auto* web_request_api =
+      extensions::BrowserContextKeyedAPIFactory<extensions::WebRequestAPI>::Get(
+          browser_context);
+
+  DCHECK(web_request_api);
+
+  if (web_request_api->MayHaveProxies()) {
+    web_request_api->ProxyWebSocket(frame, std::move(factory), url,
```
nornagon, line 1350: I think this needs the same treatment as `WillCreateURLLoaderFactory`? i.e. we need to check the return value here and early-exit if `extensions::WebRequestAPI` decided to take over the request.

sentialx:
@nornagon I can't seem to load an extension using a testing version of Electron from the artifacts page you showed me. It's just crashing #23007

shell/browser/electron_browser_client.cc
```diff
@@ -47,6 +47,8 @@
 #include "content/public/common/web_preferences.h"
 #include "electron/buildflags/buildflags.h"
 #include "electron/grit/electron_resources.h"
+#include "extensions/browser/api/web_request/web_request_api.h"
```
deepak1556, line 50: Headers must be behind `#if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)`

sentialx: Is it good now?

deepak1556: yup thanks!

spec-main/extensions-spec.ts
```diff
@@ -175,9 +177,62 @@ describe('chrome extensions', () => {
...
+        w.webContents.once('dom-ready', async () => {
+          await customSession.loadExtension(path.join(fixtures, 'extensions', 'chrome-webRequest'));
+
+          customSession.webRequest.onBeforeRequest((details: any, callback: any) => {
+            callback({
+              cancel: true
+            });
+          });
+
+          await expect(fetch(w.webContents, url)).to.eventually.be.rejectedWith(TypeError);
```
sentialx, line 225: Is there actually a better way to get the `net::ERR_BLOCKED_BY_CLIENT` error? `fetch` throws only a `TypeError` with `Failed to fetch` message.

sentialx:
@nornagon I've added tests and I think the PR is ready to merge. Could this be backported to 9-x-y? I think these are not breaking changes.

spec-main/extensions-spec.ts
```diff
@@ -175,9 +190,98 @@ describe('chrome extensions', () => {
...
+        const customSession = session.fromPartition(`persist:${uuid.v4()}`);
+        const w = new BrowserWindow({ show: false, webPreferences: { session: customSession, sandbox: true } });
+
+        w.webContents.once('dom-ready', async () => {
+          await customSession.loadExtension(path.join(fixtures, 'extensions', 'chrome-webRequest'));
+
+          await expect(fetch(w.webContents, url)).to.eventually.be.rejectedWith(TypeError);
+        });
+
+        w.loadURL(url);
```
nornagon, line 209: All `expect`ations should be at the top level, otherwise the test runner will not continue running tests after a failure.

```suggestion
        w.loadURL(url);
        await emittedOnce(w.webContents, 'dom-ready')
        await customSession.loadExtension(path.join(fixtures, 'extensions', 'chrome-webRequest'));
        await expect(fetch(w.webContents, url)).to.eventually.be.rejectedWith(TypeError);
```

spec-main/extensions-spec.ts
```diff
@@ -175,9 +190,98 @@ describe('chrome extensions', () => {
...
+
+      it('does not cancel http requests', async () => {
+        const customSession = session.fromPartition(`persist:${uuid.v4()}`);
+        const w = new BrowserWindow({ show: false, webPreferences: { session: customSession, sandbox: true } });
+
+        w.webContents.once('dom-ready', async () => {
+          await expect(fetch(w.webContents, url)).to.have.property('status', 200);
+        });
+
+        w.loadURL(url);
```
nornagon, line 220:
```suggestion
        w.loadURL(url);
        await emittedOnce(w.webContents, 'dom-ready');
        await expect(fetch(w.webContents, url)).to.have.property('status', 200);
```

spec-main/extensions-spec.ts
```diff
@@ -175,9 +190,98 @@ describe('chrome extensions', () => {
...
+
+      w.webContents.once('dom-ready', async () => {
+        await customSession.loadExtension(path.join(fixtures, 'extensions', 'chrome-webRequest'));
+
+        customSession.webRequest.onBeforeRequest(() => {
+          throw new Error('Electron onBeforeRequest callback has been called');
+        });
+
+        await expect(fetch(w.webContents, url)).to.eventually.be.rejectedWith(TypeError);
+      });
```
nornagon, line 236: etc

spec-main/extensions-spec.ts
```diff
@@ -175,9 +190,98 @@ describe('chrome extensions', () => {
...
+
+      w.loadURL(url);
+    });
+
+    it('takes precedence over Electron webRequest - WebSocket', async () => {
```
nornagon, line 241: Should this test that it only takes precedence on urls that the extension specifically intercepts?

e.g. if an extension registers webRequest handlers for `*.google.com`, and the main process registers handlers through the Electron API for `*.slack.com`, both handlers should work.

shell/browser/electron_browser_client.cc
```diff
@@ -1429,6 +1440,22 @@ void ElectronBrowserClient::CreateWebSocket(
   v8::Isolate* isolate = v8::Isolate::GetCurrent();
   v8::HandleScope scope(isolate);
   auto* browser_context = frame->GetProcess()->GetBrowserContext();
+
+#if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
+  auto* web_request_api =
+      extensions::BrowserContextKeyedAPIFactory<extensions::WebRequestAPI>::Get(
+          browser_context);
+
+  DCHECK(web_request_api);
```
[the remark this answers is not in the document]
sentialx: If I recall correctly, I saw that DCHECK in Chromium code, but sure, I can handle this gracefully.

spec-main/extensions-spec.ts
sentialx: Is that okay?
https://github.com/sentialx/electron/blob/a1cdc70435d953e0ad45fb8316b08dabbadc053d/spec-main/extensions-spec.ts#L232-L242

nornagon: I'm a little confused how that test works. How does the `onBeforeRequest` handler get invoked given that it's supposed to only trigger on google.com urls, which aren't ever requested afaict?

sentialx: Ah yes, I forgot, sorry.

sentialx:
It seems to be still related #22175

sentialx:
@nornagon I've added a note about `chrome.webRequest` API.

nornagon:
@sentialx looks like there are conflicts now, whoops

sentialx:
@nornagon Fixed.

jkleinsc:
mac arm64 build failures are unrelated to this PR.

docs/api/extensions.md
```diff
@@ -1,104 +1,110 @@
...
-
-The following methods of `chrome.tabs` are supported:
-
-- `chrome.tabs.sendMessage`
-- `chrome.tabs.executeScript`
-
-> **Note:** In Chrome, passing `-1` as a tab ID signifies the "currently active
-> tab". Since Electron has no such concept, passing `-1` as a tab ID is not
-> supported and will raise an error.
+# Chrome Extension Support
```
nornagon, line 1: Why is this entire file listed as having changed...? it looks the same 😕

jkleinsc: @nornagon if you look at the rich view diff, it only shows the new `chrome.webRequest` entry.  I'm wondering if the text view diff is a GitHub hiccup.

shell/browser/electron_browser_client.cc
```diff
@@ -1447,6 +1477,25 @@ bool ElectronBrowserClient::WillCreateURLLoaderFactory(
     bool* bypass_redirect_checks,
     bool* disable_secure_dns,
     network::mojom::URLLoaderFactoryOverridePtr* factory_override) {
+#if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
+  auto* web_request_api =
+      extensions::BrowserContextKeyedAPIFactory<extensions::WebRequestAPI>::Get(
+          browser_context);
+
+  // NOTE: Some unit test environments do not initialize
+  // BrowserContextKeyedAPI factories for e.g. WebRequest.
```
nornagon, line 1486: Is this relevant to us? Perhaps we should just `DCHECK` here?

shell/browser/extensions/electron_extension_loader.cc
```diff
@@ -110,6 +110,9 @@ void ElectronExtensionLoader::FinishExtensionLoad(
   if (extension) {
     extension_registrar_.AddExtension(extension);
   }
+  ExtensionPrefs* extension_prefs = ExtensionPrefs::Get(browser_context_);
+  extension_prefs->OnExtensionInstalled(extension.get(), Extension::ENABLED,
```
nornagon, line 114: What is this needed for?

shell/common/extensions/electron_extensions_client.cc
```diff
@@ -130,9 +130,7 @@ const GURL& ElectronExtensionsClient::GetWebstoreUpdateURL() const {
 }
 
 bool ElectronExtensionsClient::IsBlacklistUpdateURL(const GURL& url) const {
-  // TODO(rockot): Maybe we want to do something else here. For now we accept
-  // any URL as a blacklist URL because we don't really care.
-  return true;
+  return false;
```
nornagon, line 133: What does this allow?

spec-main/extensions-spec.ts
```diff
@@ -193,9 +208,108 @@ describe('chrome extensions', () => {
...
+      const customSession = session.fromPartition(`persist:${uuid.v4()}`);
+      const w = new BrowserWindow({ show: false, webPreferences: { session: customSession, sandbox: true } });
+
+      w.loadURL(url);
+      await emittedOnce(w.webContents, 'dom-ready');
+      await customSession.loadExtension(path.join(fixtures, 'extensions', 'chrome-webRequest'));
+      customSession.webRequest.onBeforeRequest({ urls: ['*://*.google.com/*'] }, async (details, callback) => {
+        callback({ cancel: true });
+      });
+      await expect(fetch(w.webContents, 'https://google.com')).to.eventually.be.rejectedWith(TypeError);
```
nornagon, line 260: Can we change these to `https://does.not.exist` or `https://127.127.127.127` some other obviously nonsense URL? Our tests should never make external requests.

nornagon (changes requested):
I think this is close to landable!

shell/browser/extensions/electron_extension_loader.cc
sentialx: There was somewhere a DCHECK in web_request_api.cc code requiring this.

shell/common/extensions/electron_extensions_client.cc
sentialx: @nornagon Returning `true` basically blocks all URLs to be processed through code in `web_request_api.cc`.
https://source.chromium.org/chromium/chromium/src/+/master:extensions/browser/api/web_request/web_request_permissions.cc;l=329;drc=5ba77dad98cf94506cf3700f9e12fcdc65fadfb6?originalUrl=https:%2F%2Fcs.chromium.org%2F

shell/browser/extensions/electron_extension_loader.cc
sentialx: https://source.chromium.org/chromium/chromium/src/+/master:extensions/browser/api/web_request/web_request_api.cc;l=2658;drc=6e8b402a6231405b753919029c9027404325ea00?q=web_request_api&ss=chromium%2Fchromium%2Fsrc

shell/browser/electron_browser_client.cc
```diff
@@ -1410,6 +1411,18 @@ bool ElectronBrowserClient::WillInterceptWebSocket(
   if (!web_request.get())
     return false;
 
+#if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
+  if (!frame) {
+    return false;
+  }
```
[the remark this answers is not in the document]
sentialx: I didn't notice, sorry

shell/browser/extensions/electron_extension_loader.cc
sentialx: And that's because there is no `ExtensionService` in Electron.
https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/extensions/extension_service.cc;l=1705;drc=c1908ef394f04e64cb4cdb8cc49e0e3b7f1c7527;bpv=0;bpt=1

shell/browser/electron_browser_client.cc
```diff
@@ -1423,8 +1424,18 @@ bool ElectronBrowserClient::WillInterceptWebSocket(
...
     return false;
+  
+  bool has_listener = web_request->HasListener();
+#if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
+  const auto* web_request_api =
+      extensions::BrowserContextKeyedAPIFactory<extensions::WebRequestAPI>::Get(
+          browser_context);
+
+  if (web_request_api)
+    has_listener |= web_request_api->MayHaveProxies();
```
nornagon, line 1435: nit:
```suggestion
    has_listener ||= web_request_api->MayHaveProxies();
```

shell/browser/electron_browser_client.cc
```diff
@@ -1438,6 +1449,20 @@ void ElectronBrowserClient::CreateWebSocket(
   v8::Isolate* isolate = JavascriptEnvironment::GetIsolate();
   v8::HandleScope scope(isolate);
   auto* browser_context = frame->GetProcess()->GetBrowserContext();
+
+#if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
+  auto* web_request_api =
+      extensions::BrowserContextKeyedAPIFactory<extensions::WebRequestAPI>::Get(
+          browser_context);
+
+  if (web_request_api && web_request_api->MayHaveProxies()) {
```
nornagon, line 1458: It looks like `MayHaveProxies` will return true even if there isn't a proxy for this specific URL. So if there's any web request handler registered for any URL, Electron's webRequest API won't get a chance at proxying the websocket.

Can we add a test that does something like:

1. Extension registers a webRequest handler for ws://a.com
2. Electron registers a webRequest handler for ws://b.com
3. Page connects to ws://a.com, extension gets the event
4. Page connects to ws://b.com, Electron gets the event

shell/browser/electron_browser_client.cc
sentialx: There was a compilation error, so I changed it to `|=` ¯\_(ツ)_/¯

nornagon: oh, right you are. I guess `||=` isn't in C++. oh well. 🤷‍♂️

samuelmaddock:
@sentialx Would you be open to granting me access to your fork? I'd like to help with landing this work.

sentialx:
@samuelmaddock Sure!

shell/browser/extensions/electron_extension_loader.cc
samuelmaddock, line 114: This addition comes with it persistence of a number of properties for each extension installed.

image

This goes back to some discussion in https://github.com/electron/electron/pull/25198#discussion_r487389995 . I can think of a few ways around this:
- Manually write the `install_time` property to avoid the DCHECK
- Patch //extensions to remove the DCHECK
- Switch to using the `InMemoryPrefStore` to avoid any of this from getting written to disk

I'll have to think on this more to see if there are other alternatives.

samuelmaddock: This is still persisted to the `Preferences` JSON store, but I modified the code to write only the `install_time` pref in 
8d837e4

image

sentialx:
I was discussing with @nornagon which `webRequest` should take precedence over the other. `chrome.webRequest` over Electron's webRequest or the opposite? There are limitations to both options though:
- if extensions come first, that means that Electron’s webRequest API has no opportunity to override what an extension does.
- if Electron comes first, it means you couldn’t have a “wildcard” base webRequest listener in Electron and allow extensions to override it

However, both cases seem to be weird for me personally. I was thinking of removing Electron's `webRequest` backend and replacing it with `chrome.webRequest's` backend, keeping similar API in the main process. This would help us avoid conflicts between two separate `webRequest` backends, but it would involve patching `web_request_api.cc` file quite extensively.

@samuelmaddock Maybe you've got some other ideas?

zcbenz:
On the priority of the 2 APIs, I prefer making Electron's `webRequest` have higher priority, because loading an extension should not change the app's behavior, it is better having a broken extension then having a broken app.

In the long term we should find a real solution for this, but I think currently having a broken webRequest API in extension is acceptable, and it is probably better than adding a large patch to extensions code.

sentialx:
@zcbenz Okay, is there a way to check if Electron webRequest has a listener registered for a certain request? Similarly to `web_request_api->MaybeProxyURLLoaderFactory` which is returning boolean.

zcbenz:
Yeah, there is a `HasListener` method in `electron::api::WebRequest`, you can find a use case in `ElectronBrowserClient::WillInterceptWebSocket`.

nornagon:
It looks like this PR is still pending changes, so I'll hold off on re-reviewing for now. Please ping me when the above concerns have been addressed :)

sentialx:
@nornagon I think it's ready for review. I've made Electron webRequest to take precedence over chrome.webRequest.

docs/api/extensions.md
```diff
@@ -115,3 +115,9 @@ The following methods of `chrome.management` are supported:
 - `chrome.management.getPermissionWarningsByManifest`
 - `chrome.management.onEnabled`
 - `chrome.management.onDisabled`
+
+### `chrome.webRequest`
+
+All features of this API are supported.
+
+> **NOTE:** `chrome.webRequest` takes precedence over Electron's [`webRequest`](web-request.md) module.
```
nornagon, line 123: I think this needs to be updated

shell/browser/extensions/electron_extension_loader.cc
```diff
@@ -110,6 +113,26 @@ void ElectronExtensionLoader::FinishExtensionLoad(
...
+    ExtensionPrefs::ScopedDictionaryUpdate update(
+        extension_prefs, extension.get()->id(),
+        extensions::pref_names::kPrefPreferences);
+    auto preference = update.Create();
+    const base::Time install_time = base::Time().Now();
+    std::unique_ptr<base::Value> install_time_str =
+        base::Value(base::NumberToString(install_time.ToInternalValue()))
+            .CreateDeepCopy();
+    preference->SetWithoutPathExpansion("install_time",
+                                        std::move(install_time_str));
```
nornagon, line 133: `SetWithoutPathExpansion` is deprecated, see https://source.chromium.org/chromium/chromium/src/+/master:base/values.h;l=661;drc=2a06d6e94a216850e057c4ba7d36c64032448c77

```suggestion
    preference->SetStringKey("install_time", base::NumberToString(install_time.ToInternalValue()));
```

nornagon (approved):
Looks good to me. Just a couple of minor suggestions.

sentialx:
@nornagon Can this be backported to 9-x-y, since I don't think it introduces any breaking changes?

nornagon:
@sentialx backports to the oldest supported branch are generally speaking only security patches. A backport to 10/11/12 might be accepted though, I've applied the relevant tags and @electron/wg-releases will make a decision.

VerteDinde:
@sentialx @nornagon Just wanted to confirm - the Releases WG met today and confirmed that we'll backport this to 10, 11 and 12 🙂

nornagon (approved):

PR #22531, opened 2020-03-04
fix: remove catch-all HandleScope

nornagon:
#### Description of Change
This removes the catch-all HandleScope in JavascriptEnvironment. This could cause memory leaks when creating v8 objects in C++ without an explicit HandleScope on the stack. Now, such usage will cause a crash.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Fixed several memory leaks related to V8 handles not being properly scoped.

PR #22657, opened 2020-03-12
fix: enter handle scope when creating custom event

zcbenz:
Followup of https://github.com/electron/electron/pull/22531 to fix a case where custom event objects are not guarded with a handle scope.

Notes: no-notes

### DIFF
--- a/shell/common/gin_helper/event_emitter.h
+++ b/shell/common/gin_helper/event_emitter.h
@@ -62,6 +62,8 @@ class EventEmitter : public gin_helper::Wrappable<T> {
   // this.emit(name, new Event(flags), args...);
   template <typename... Args>
   bool EmitWithFlags(base::StringPiece name, int flags, Args&&... args) {
+    v8::Locker locker(isolate());
+    v8::HandleScope handle_scope(isolate());
     return EmitCustomEvent(name,
                            internal::CreateEventFromFlags(isolate(), flags),
                            std::forward<Args>(args)...);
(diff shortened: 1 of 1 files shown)


shell/common/gin_helper/event_emitter.h
```diff
@@ -108,8 +110,6 @@ class EventEmitter : public gin_helper::Wrappable<T> {
     // It's possible that |this| will be deleted by EmitEvent, so save anything
     // we need from |this| before calling EmitEvent.
     auto* isolate = this->isolate();
-    v8::Locker locker(isolate);
-    v8::HandleScope handle_scope(isolate);
```
nornagon: I'm not sure it's correct to remove this scope. Check all callers of `EmitWithEvent`?

shell/common/gin_helper/event_emitter.h
```diff
@@ -62,6 +62,8 @@ class EventEmitter : public gin_helper::Wrappable<T> {
   // this.emit(name, new Event(flags), args...);
   template <typename... Args>
   bool EmitWithFlags(base::StringPiece name, int flags, Args&&... args) {
+    v8::Locker locker(isolate());
+    v8::HandleScope handle_scope(isolate());
```
nornagon: If you observed a crash related to this handle scope being missing, could you add a test which exercises the crashing behavior?

shell/common/gin_helper/event_emitter.h
zcbenz: This method takes `v8::Local` as argument, so all callers are required to be in a handle scope and I think it is not needed here.

shell/common/gin_helper/event_emitter.h
zcbenz: This require user interaction with Tray and currently we don't have a way to test that.

I think it is obvious that `EmitWithFlags` requires a handle scope due to use of `internal::CreateEventFromFlags` so it should be fine without a test.

shell/common/gin_helper/event_emitter.h
nornagon: Oh good point.

PR #22668, opened 2020-03-12
feat: prompt expected type in argument type error

zcbenz:
#### Description of Change

Currently when wrong type of argument is passed, an error like this would be thrown:

```
Error processing argument at index 1, conversion failure from 
```

(gin can not recognize most types, so the type is empty above)

With this PR, the error would become:

```
Error processing argument at index 1, conversion failure from Boolean to URL
```

To support this, we only need to add a `type_name` member to custom converters:

```c++
template <>
struct Converter<GURL> {
  static constexpr const char* type_name = "url";
  static v8::Local<v8::Value> ToV8(v8::Isolate* isolate, const GURL& val);
};
```

If there is no `type_name` member, the old-style error would be thrown:

```
Error processing argument at index 1, conversion failure from Boolean
```

In this PR I have added  `type_name` member to a few custom gin converters, we should gradually add it to all converters in future.

A patch is added in this PR because we have to be able to access the private `next_` member of `gin::Arguments` to support custom error.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Prompt expected type in argument type error.

### DIFF
--- a/filenames.gni
+++ b/filenames.gni
@@ -526,6 +526,7 @@ filenames = {
     "shell/common/gin_helper/promise.h",
     "shell/common/gin_helper/trackable_object.cc",
     "shell/common/gin_helper/trackable_object.h",
+    "shell/common/gin_helper/v8_type_traits.h",
     "shell/common/gin_helper/wrappable.cc",
     "shell/common/gin_helper/wrappable.h",
     "shell/common/gin_helper/wrappable_base.h",
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -15,6 +15,7 @@ web_contents.patch
 webview_cross_drag.patch
 disable_user_gesture_requirement_for_beforeunload_dialogs.patch
 gin_enable_disable_v8_platform.patch
+gin_expose_arguments_members.patch
 blink-worker-enable-csp-in-file-scheme.patch
 disable-redraw-lock.patch
 v8_context_snapshot_generator.patch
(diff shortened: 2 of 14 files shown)


deepak1556:
Maybe move the `gin_helper::Arguments` as patch to `gin::Arguments`, that way members of classes being converted to `gin::Wrappable` can still get benefit from this feature through `gin::Arguments`.

Ref https://github.com/electron/electron/pull/22726 for completeness

zcbenz:
@MarshallOfSound 

To have a nested "string to integer" error you have to also patch the gin converters in Chromium, which would then break much more things. For example every code that uses `gin::Dictionary::Get` might break as it now throws exception when it was not supposed to.

I believe throwing error in converters is a dead end.

---

And throwing nested type error for structures would not be helpful if it does not carry the information of argument index and property name.

For example, with the approach in #22643 the `new BrowserWindow({webPreferences: {session: 1}}}` would throw:

`
Attempted to convert a non-string value to a string.
`

which mentions no source type, no index of the argument, and no information on which property failed to convert. It would be very confusing when there are multiple arguments with complex options.

I don't think it is better than a clear message like:

`
Error processing argument at index 0, conversion failure from Object to BrowserWindowOptions
`

An ideal error message that satisfies your expectation for nested type error would be like:

`
Error processing argument at index 0, conversion failure from Object to BrowserWindowOptions: unable to convert webPreferences property from Object to WebPreferences: unable to convert session property from Number to String
`

It is technically possible since we have all necessary information in `docs/` and we could automatically generate parsing code, but it would be massive work.

I think the approach in this PR is the middle group between naive type error and ultimate nested type error: it is simple to implement and understand, it is an improvement without any sacrifice, and we can even try to upstream the change to Chromium. If we do come up with a way to report nested type error, we can just extend the error message.

This PR does not kill the idea of reporting nested type error, this PR is compatible with it, and it improves what we have.

zcbenz:
I plan to upstream the changes to Chromium directly.

miniak:
@zcbenz are you planning to move this forward?

zcbenz:
Yeah, not recently but I'll try to submit a patch to Chromium some time later.

PR #22672, opened 2020-03-12
fix: don't assign NSAlert to window which is not visible

### LINKED ISSUE #22671: message box can't be dismissed on mac
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 10.0.0-nightly.20200309
* **Operating System:**
  * macOS 10.15.2
  * Probably never

### Expected Behavior
It's possible to dismiss message box created with `dialog.showMessageBoxSync`

### Actual Behavior
It's not possible to dismiss message box created with `dialog.showMessageBoxSync` if window which is not visible is provided as parameter (it is not possible to focus such box and clicking its ok button doesn't close it).

### To Reproduce
Run following app:
```const { app, dialog, BrowserWindow } = require('electron')

function createWindow() {
  let win = new BrowserWindow({
    show: false
  });
  dialog.showMessageBoxSync(win, { message: "Message" });
}

app.on('ready', createWindow)
```

CezaryKulakowski:
Closes #22671.

Fixes an issue with possible creation of a messageBox which cannot be dismissed on macOS.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue with possible creation of a messageBox which cannot be dismissed on macOS.

### DIFF
--- a/shell/browser/ui/message_box_mac.mm
+++ b/shell/browser/ui/message_box_mac.mm
@@ -97,7 +97,9 @@ int ShowMessageBoxSync(const MessageBoxSettings& settings) {
   NSAlert* alert = CreateNSAlert(settings);
 
   // Use runModal for synchronous alert without parent, since we don't have a
-  // window to wait for.
+  // window to wait for. Also use it when window is provided but it is not
+  // shown as it would be impossible to dismiss the alert if it is connected
+  // to invisible window (see #22671).
   if (!settings.parent_window || !settings.parent_window->IsVisible())
     return [[alert autorelease] runModal];
 
--- a/shell/browser/ui/message_box_mac.mm
+++ b/shell/browser/ui/message_box_mac.mm
@@ -98,7 +98,7 @@ int ShowMessageBoxSync(const MessageBoxSettings& settings) {
 
   // Use runModal for synchronous alert without parent, since we don't have a
   // window to wait for.
-  if (!settings.parent_window)
+  if (!settings.parent_window || !settings.parent_window->IsVisible())
     return [[alert autorelease] runModal];
 
   __block int ret_code = -1;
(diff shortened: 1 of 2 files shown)


codebytere:
@CezaryKulakowski this appears to be causing a crash on Linux.

CezaryKulakowski:
@codebytere this fix changes only `mm` file which is not being compiled on linux or windows.

codebytere:
@CezaryKulakowski misread - you're correct. Thought it was a `.cc` file on first glance.

codebytere:
@MarshallOfSound in that case, would we want to show the window ourselves if it's hidden before invoking the message box? I wanted to avoid the breaking change in this PR but it might be something to consider

CezaryKulakowski:
@MarshallOfSound what do you mean by documenting this behavior and telling folks to avoid this scenario to call window.show() before opening the modal? Leaving it as it is now and adding some information to documentation? The scenario leading to it isn't very unusual and it breaks the ui.

MarshallOfSound:
> Leaving it as it is now and adding some information to documentation?

Yes exactly that

zcbenz:
@MarshallOfSound @CezaryKulakowski Maybe we can do this fix and document it explicitly? The `showMessageBoxSync` is totally broken under this case, I don't think we should just leave a broken case there and tell users not to do it in the documentation.

shell/browser/ui/message_box_mac.mm
```diff
@@ -98,7 +98,7 @@ int ShowMessageBoxSync(const MessageBoxSettings& settings) {
 
   // Use runModal for synchronous alert without parent, since we don't have a
   // window to wait for.
-  if (!settings.parent_window)
+  if (!settings.parent_window || !settings.parent_window->IsVisible())
```
zcbenz: Should leave a comment on why `IsVisible` is checked here.

zcbenz:
@CezaryKulakowski Can you also document this behavior? (On macOS showing message box on invisible parent window would behave the same with no parent window.)

@MarshallOfSound Are you OK with both documenting this scenario and having a workaround?

CezaryKulakowski:
@zcbenz I've pushed a fixup which updates documentation. I hope this is what you meant.

PR #22673, opened 2020-03-12
fix: crash on invalid zoomFactor

codebytere:
#### Description of Change

This PR fixes a crash caused by passing a value of `0.0` to `webContents.setZoomFactor()`, seen below:

```console
[38965:0312/111238.132477:FATAL:values.cc(211)] Check failed: false. Non-finite (i.e. NaN or positive/negative infinity) values cannot be represented in JSON
0   Electron Framework                  0x0000000117ce2839 base::debug::CollectStackTrace(void**, unsigned long) + 9
1   Electron Framework                  0x0000000117badb13 base::debug::StackTrace::StackTrace() + 19
2   Electron Framework                  0x0000000117bcc639 logging::LogMessage::~LogMessage() + 249
3   Electron Framework                  0x0000000117cd11c9 base::Value::Value(double) + 153
4   Electron Framework                  0x0000000112b2f420 electron::ZoomLevelDelegate::OnZoomLevelChanged(content::HostZoomMap::ZoomLevelChange const&) + 768
5   Electron Framework                  0x0000000116e79548 void base::CallbackList<void (content::HostZoomMap::ZoomLevelChange const&)>::Notify<content::HostZoomMap::ZoomLevelChange&>(content::HostZoomMap::ZoomLevelChange&) + 56
```

It will now throw a descriptive error informing the user that values must be greater than 0.0. Also corrects docs which incorrectly stated that the zoomFactor was a Number and not a Double.

cc @MarshallOfSound @jkleinsc @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a potential crash on invalid `zoomFactor` values when setting the zoom factor of a webpage.

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -4,6 +4,7 @@
 
 #include "shell/browser/api/electron_api_web_contents.h"
 
+#include <limits>
 #include <memory>
 #include <set>
 #include <string>
@@ -2489,7 +2490,13 @@ double WebContents::GetZoomLevel() const {
   return zoom_controller_->GetZoomLevel();
 }
 
-void WebContents::SetZoomFactor(double factor) {
+void WebContents::SetZoomFactor(gin_helper::ErrorThrower thrower,
+                                double factor) {
+  if (factor < std::numeric_limits<double>::epsilon()) {
+    thrower.ThrowError("'zoomFactor' must be a double greater than 0.0");
+    return;
+  }
+
   auto level = blink::PageZoomFactorToZoomLevel(factor);
   SetZoomLevel(level);
 }
--- a/shell/renderer/api/electron_api_web_frame.cc
+++ b/shell/renderer/api/electron_api_web_frame.cc
@@ -2,6 +2,7 @@
 // Use of this source code is governed by the MIT license that can be
 // found in the LICENSE file.
 
+#include <limits>
 #include <memory>
 #include <string>
 #include <utility>
@@ -275,7 +276,14 @@ double GetZoomLevel(v8::Local<v8::Value> window) {
   return result;
 }
 
-void SetZoomFactor(v8::Local<v8::Value> window, double factor) {
+void SetZoomFactor(gin_helper::ErrorThrower thrower,
+                   v8::Local<v8::Value> window,
+                   double factor) {
+  if (factor < std::numeric_limits<double>::epsilon()) {
+    thrower.ThrowError("'zoomFactor' must be a double greater than 0.0");
+    return;
+  }
+
   SetZoomLevel(window, blink::PageZoomFactorToZoomLevel(factor));
 }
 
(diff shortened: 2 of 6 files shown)


shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2489,7 +2489,13 @@ double WebContents::GetZoomLevel() const {
   return zoom_controller_->GetZoomLevel();
 }
 
-void WebContents::SetZoomFactor(double factor) {
+void WebContents::SetZoomFactor(gin_helper::ErrorThrower thrower,
+                                double factor) {
+  if (factor == 0.0) {
+    thrower.ThrowError("'zoomFactor' must be a double greater than 0.0");
```
MarshallOfSound: Could it technically be negative? Idk what happens if you do that

codebytere: yeah - i guess this could be altered to `<= 0.0`?

codebytere: done!

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2489,7 +2489,13 @@ double WebContents::GetZoomLevel() const {
   return zoom_controller_->GetZoomLevel();
 }
 
-void WebContents::SetZoomFactor(double factor) {
+void WebContents::SetZoomFactor(gin_helper::ErrorThrower thrower,
+                                double factor) {
+  if (factor <= 0.0) {
```
ckerr: There are all kinds of ways for comparing equality in floating-points to fail.

If we're trying to ensure that `factor` is positive, it's safer to add `#inclue <cmath>` to the top and use `std::signbit()`:

```suggestion
  if (!std::signbit(factor)) {
```

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -275,7 +275,14 @@ double GetZoomLevel(v8::Local<v8::Value> window) {
   return result;
 }
 
-void SetZoomFactor(v8::Local<v8::Value> window, double factor) {
+void SetZoomFactor(gin_helper::ErrorThrower thrower,
+                   v8::Local<v8::Value> window,
+                   double factor) {
+  if (factor <= 0.0) {
```
ckerr: Same as above

spec-main/api-web-contents-spec.ts
```diff
@@ -800,6 +800,15 @@ describe('webContents module', () => {
...
 
+    it('throws on an invalid zoomFactor', async () => {
+      const w = new BrowserWindow({ show: false })
+      await w.loadURL('about:blank')
+
+      expect(() => {
+        w.webContents.setZoomFactor(0.0)
+      }).to.throw(/'zoomFactor' must be a double greater than 0.0/)
+    })
+
```
ckerr: Might be worthwhile to confirm on negative as well?

shell/browser/api/electron_api_web_contents.cc
ckerr: The analysis is right, but the suggestion is wrong: `signbit()` is not going to do what we want here.

Revised suggestion: either use `std::numeric_limits<double>::epsilon()` or KISS with `< 0.001` or the like :slightly_smiling_face:

PR #22675, opened 2020-03-12
chore: fix default_app dimensions

MarshallOfSound:
On `master` the nightly version tag is too long for the window and causes text wrapping which makes a scrollbar appear.  It looks a lot nicer without the scrollbar and everything on one line 😄 

This also removes the white flash when showing the window by adjusting it to be the same background color as the page 👍 

Notes: no-notes


ckerr:
> the nightly version tag is too long for the window and causes text wrapping

:joy:

Issue #22646, opened 2020-03-11
auto re-enable maximizable even if set to false


RoderickQiu:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, with only a solution not working https://github.com/electron/electron/issues/11756.

### Issue Details

* **Electron Version:**
  * 7.1.14
* **Operating System:**
  * macOS 10.15.2

### Expected Behavior
`maximizable` property should always work.

### Actual Behavior
`maximizable` property will be ignored when some other properties are changed to the window or an alert is shown.

### To Reproduce

1.

From https://github.com/electron/electron/issues/11756.
> If i create a window like this:
> 
> ```javascript
> win = new BrowserWindow({ title: "My App", backgroundColor:'#FF0000', maximizable:false, width: 400, height: 200, resizable: false})
> ```
> 
> The window maximize button get disabled as wanted, but if we active the alert function inside
> 
> the index.html file by putting `<script> alert("test"); </script>` inside the file then the maximize button get enabled again and we can use it again.

2.
Also, for other kinds of code, like
```javascript
var mainWindow = new BrowserWindow({
    width: 800,
    height: 600,
    maximizable: false,
    webPreferences: {
      nodeIntegration: true
    }
  })

nativeTheme.on('updated', function theThemeHasChanged() {
    if (nativeTheme.shouldUseDarkColors) {
      if (mainWindow != null) {
        mainWindow.setBackgroundColor('#393939');
      }
    } else {
      if (mainWindow != null) {
        mainWindow.setBackgroundColor('#fefefe');
      }
    }
  })
```
can also produce this problem.
(how to: set dark mode / light mode, and then maximizable property is ignored.)

### Additional Information

These are *extremely old* resources that I found:

- https://github.com/electron/electron/pull/6736
- https://github.com/electron/electron/pull/18425

codebytere:
@RoderickQiu some research tells me this is becuase the `maximizable` setting does not persist through an NSWindow redraw event, which is triggered when a user changes between light and dark modes as discussed [here](https://developer.apple.com/documentation/xcode/supporting_dark_mode_in_your_interface?language=objc):

[...]

PR #22677, opened 2020-03-13
fix: persist maximizable state through theme change

codebytere:
#### Description of Change

(Partially) fixes https://github.com/electron/electron/issues/22646.

NSWindows reset their traffic light enabled states on redraw, which happens on theme changes (i.e changing from Light to Dark Mode) and so breaks old maximizable settings set by users. Luckily, we already hook into theme changes to reposition traffic lights, so this was a fairly straightforward fix in that we can use that function to persist maximizable state.

cc @MarshallOfSound @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where changing theme on macOS would break window maximizability state.

### DIFF
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -47,13 +47,13 @@
 // This view would inform Chromium to resize the hosted views::View.
 //
 // The overrided methods should behave the same with BridgedContentView.
-@interface ElectronAdapatedContentView : NSView {
+@interface ElectronAdaptedContentView : NSView {
  @private
   views::NativeWidgetMacNSWindowHost* bridge_host_;
 }
 @end
 
-@implementation ElectronAdapatedContentView
+@implementation ElectronAdaptedContentView
 
 - (id)initWithShell:(electron::NativeWindowMac*)shell {
   if ((self = [self init])) {
@@ -519,7 +519,10 @@ NativeWindowMac::~NativeWindowMac() {
     [NSEvent removeMonitor:wheel_event_monitor_];
 }
 
-void NativeWindowMac::RepositionTrafficLights() {
+void NativeWindowMac::RedrawTrafficLights() {
+  // Ensure maximizable options retain pre-existing state.
+  SetMaximizable(maximizable_);
+
   if (!traffic_light_position_.x() && !traffic_light_position_.y()) {
     return;
   }
@@ -708,7 +711,7 @@ void NativeWindowMac::SetExitingFullScreen(bool flag) {
 
 void NativeWindowMac::OnNativeThemeUpdated(ui::NativeTheme* observed_theme) {
   base::PostTask(FROM_HERE, {content::BrowserThread::UI},
-                 base::BindOnce(&NativeWindowMac::RepositionTrafficLights,
+                 base::BindOnce(&NativeWindowMac::RedrawTrafficLights,
                                 base::Unretained(this)));
 }
 
@@ -930,6 +933,7 @@ bool NativeWindowMac::IsMinimizable() {
 }
 
 void NativeWindowMac::SetMaximizable(bool maximizable) {
+  maximizable_ = maximizable;
   [[window_ standardWindowButton:NSWindowZoomButton] setEnabled:maximizable];
 }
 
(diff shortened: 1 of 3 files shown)


shell/browser/native_window_mac.mm
```diff
@@ -47,13 +47,13 @@
 // This view would inform Chromium to resize the hosted views::View.
 //
 // The overrided methods should behave the same with BridgedContentView.
-@interface ElectronAdapatedContentView : NSView {
+@interface ElectronAdaptedContentView : NSView {
```
codebytere: this was just a typo i fixed

shell/browser/native_window_mac.mm
```diff
@@ -930,11 +933,18 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
 }
 
 void NativeWindowMac::SetMaximizable(bool maximizable) {
+  if (maximizable != maximizable_)
```
ckerr: _(nit)_ this `if` is unnecessary

shell/browser/native_window_mac.mm
```diff
@@ -520,6 +520,9 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
 }
 
 void NativeWindowMac::RepositionTrafficLights() {
+  // Ensure maximizable options retain pre-existing state.
+  [[window_ standardWindowButton:NSWindowZoomButton] setEnabled:maximizable_];
+
```
ckerr: :+1: on handling this on theme changes, but putting this code in `RepositionTrafficLights()` is coincidental cohesion -- for example, `NativeWindowMac::SetTitle()` now refreshes the maximizable button for no reason.

Maybe the browser ui thread tasks posted in `NativeWindowMac::OnNativeThemeUpdated()` should be a new `OnNativeThemeUpdatedInUI()` type function that calls both the maximizable update and also `RepositionTrafficLights()`

codebytere: I've fixed the cohesion and made it such that `SetMaximizable` is a no-op if there is no state change. Unfortunately, moving the update call into OnNativeThemeUpdated doesn't fix the bug and continues to show the bad behavior, so this seemed like the most effective solution which wouldn't require a bit of a refactor which seemed scope-creepy.

ckerr: I wasn't suggesting moving the upate call to OnNativeThemeUpdated -- I was suggesting that OnNativeThemeUpdated to call a new wrapper function OnNatveThemeUpdatedOnUI and _that_ function call both SetMaximizable and RepositionTrafficLights.

Similar idea here wrt having OnNativeThemeUpdated be a pass-through to a sibling OnUI function: https://github.com/electron/electron/blob/master/shell/browser/api/electron_api_native_theme.cc#L38

shell/browser/native_window_mac.mm
```diff
@@ -930,11 +933,18 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
 }
 
 void NativeWindowMac::SetMaximizable(bool maximizable) {
-  [[window_ standardWindowButton:NSWindowZoomButton] setEnabled:maximizable];
+  if (maximizable_ != maximizable) {
+    maximizable_ = maximizable;
+    [[window_ standardWindowButton:NSWindowZoomButton] setEnabled:maximizable];
+  }
```
ckerr: With this new revision, [setEnabled] is not reached in this call from the ctor if the user specified a `maximizable` as `true` (because `maximizable_` is already set to `true` via the header declaration):

```
    // Set maximizable state last to ensure zoom button does not get reset
    // by calls to other APIs.
    SetMaximizable(maximizable);
```

I'm 90% sure this is safe but wanted to ask to confrm -- when a new window is created on macOS, is its NSWindowZoomButton is enabled by default?

codebytere: yes - it's default true! frameless windows and a few other things mean it's false, but there's no case where it would be false where it's not specifically _set_ to false in a fn or ctor!

shell/browser/native_window_mac.mm
```diff
@@ -930,11 +933,18 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
 }
 
 void NativeWindowMac::SetMaximizable(bool maximizable) {
-  [[window_ standardWindowButton:NSWindowZoomButton] setEnabled:maximizable];
+  if (maximizable_ != maximizable) {
```
zcbenz: Wouldn't the `SetMaximizable(maximizable_)` call in `RepositionTrafficLights` be no-op since `maximizable_ != maximizable` would always be false here?

shell/browser/native_window_mac.mm
codebytere: @ckerr this gets called whenever macOS redraws a window - it's fine that it does both of these things i'd say although i've renamed it to `RedrawTrafficLights` to clarify intent. macOS/Cocoa also optimizes this away so it's not called excessively when not needed.

PR #22686, opened 2020-03-14
chore: support props/fns for BrowserWindow

codebytere:
#### Description of Change

Continuation of decision to dual-support both properties and function for primitive module properties, this time for `BrowserWindows`.

cc @wg-api

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -1115,15 +1115,11 @@ Returns `Integer[]` - Contains the window's maximum width and height.
 
 * `resizable` Boolean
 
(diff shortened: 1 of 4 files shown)


ckerr:
In the docs, would it make sense to make explicit to users that the two approaches will co-exist and that the properties and functions are equivalent to each other?

codebytere:
@ckerr perhaps, but there will be a bunch of these PRs grouped by module for a bit so i can open up a specific one to document that? i imagine we'll bikeshed how we do it a bit anyways.

PR #22696, opened 2020-03-15
fix: script invocation in 'first app' documentation 

malikasinger1:
notes: fix for error discussed in this issue https://github.com/electron/electron-quick-start/issues/363

### DIFF
--- a/docs/tutorial/first-app.md
+++ b/docs/tutorial/first-app.md
@@ -54,7 +54,7 @@ to execute the current package:
   "version": "0.1.0",
   "main": "main.js",
   "scripts": {
-    "start": "node ."
+    "start": "electron ."
   }
 }
 ```


zcbenz:
With this change the document now read:

```
__Note__: If the `main` field is not present in `package.json`, Electron will
attempt to load an `index.js` (as Node.js does). If this was actually
a simple Node application, you would add a `start` script that instructs `node`
to execute the current package:

{
  "name": "your-app",
   "version": "0.1.0",
   "main": "main.js",
   "scripts": {
     "start": "node ."
   }
 }
```

Which is actually wrong as the code example is intended to demonstrate the case of starting the script with node.

However I fully understand how confusing the document is and I'll start a PR to remove this note completely.

zcbenz:
Closed in favor of https://github.com/electron/electron/pull/23929.

PR #22701, opened 2020-03-16
feat: add desktopCapturer.getMediaSourceIdForWebContents() to get stream source id from web contents

### LINKED ISSUE #4776: Capture media for specified WebContents
- Electron version: `0.37.0`
- Operating system: All

I've been trying for weeks now to figure this out, I need to somehow capture the audio output from my electron application.  But the `desktopCapturer` API appears to exclude that exact use case :cry: 

Is there an alternative way of performing this capture WITHOUT capturing the entire computers audio through the "screen" capturer?

https://github.com/atom/electron/blob/master/docs/api/desktop-capturer.md

Also open to any JS methods of capturing a stream from an Audio element

Volune:
#### Description of Change
New feature: add `desktopCapturer.getWebContentsStream()` to provide a way to get a MediaStream from a WebContents using `getUserMedia`.

An example code is provided in `docs/api/desktop-capturer.md`

This should be enough to resolve #4776

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added desktopCapturer.getMediaSourceIdForWebContents(), can be used with getUserMedia to get a stream for a WebContent.

#### Comments

Feedbacks on the API (especially on functions and arguments naming) are welcomed

These two tests failed locally. Since they don't seem related with the code changes, I assume it's caused by the test environment.
```
not ok 746 Notification module (dbus) Notification module using org.freedesktop.Notifications "before all" hook
  Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/mnt/c/dev/github.volune/electron-gn/src/electron/spec-main/api-notification-dbus-spec.ts)
      at listOnTimeout (internal/timers.js:549:17)
      at processTimers (internal/timers.js:492:7)


not ok 324 BrowserWindow module "webPreferences" option "sandbox" option exposes "exit" event to preload script
  Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/mnt/c/dev/github.volune/electron-gn/src/electron/spec-main/api-browser-window-spec.ts)
      at listOnTimeout (internal/timers.js:549:17)
      at processTimers (internal/timers.js:492:7)
```

### DIFF
--- a/docs/api/desktop-capturer.md
+++ b/docs/api/desktop-capturer.md
@@ -143,9 +143,9 @@ which can detected by [`systemPreferences.getMediaAccessStatus`].
 * `webContentsId` number - Id of the WebContents to get stream of
 
 Returns `Promise<string>` - Resolves with the identifier of a WebContents stream, this identifier can be
-    used with [`navigator.mediaDevices.getUserMedia`].
-    The identifier is **only valid for 10 seconds**.
-    The identifier may be empty if not requested from a renderer process.
+used with [`navigator.mediaDevices.getUserMedia`].
+The identifier is **only valid for 10 seconds**.
+The identifier may be empty if not requested from a renderer process.
 
 [`navigator.mediaDevices.getUserMedia`]: https://developer.mozilla.org/en/docs/Web/API/MediaDevices/getUserMedia
 [`systemPreferences.getMediaAccessStatus`]: system-preferences.md#systempreferencesgetmediaaccessstatusmediatype-macos
--- a/docs/api/desktop-capturer.md
+++ b/docs/api/desktop-capturer.md
@@ -72,6 +72,50 @@ const constraints = {
 }
 ```
 
+This example shows how to capture a video from a [WebContents](web-contents.md)
+
+```javascript
+// In the renderer process.
+const { desktopCapturer, remote } = require('electron')
+
+desktopCapturer.getMediaSourceIdForWebContents(remote.getCurrentWebContents().id).then(async mediaSourceId => {
+  try {
+    const stream = await navigator.mediaDevices.getUserMedia({
+      audio: {
+        mandatory: {
+          chromeMediaSource: 'tab',
+          chromeMediaSourceId: mediaSourceId
+        }
+      },
+      video: {
+        mandatory: {
+          chromeMediaSource: 'tab',
+          chromeMediaSourceId: mediaSourceId,
+          minWidth: 1280,
+          maxWidth: 1280,
+          minHeight: 720,
+          maxHeight: 720
+        }
+      }
+    })
+    handleStream(stream)
+  } catch (e) {
+    handleError(e)
+  }
+})
+
+function handleStream (stream) {
+  const video = document.querySelector('video')
+  video.srcObject = stream
+  video.onloadedmetadata = (e) => video.play()
+}
+
+function handleError (e) {
+  console.log(e)
+}
+```
+
+
 ## Methods
 
 The `desktopCapturer` module has the following methods:
@@ -93,6 +137,15 @@ Returns `Promise<DesktopCapturerSource[]>` - Resolves with an array of [`Desktop
 
 **Note** Capturing the screen contents requires user consent on macOS 10.15 Catalina or higher,
 which can detected by [`systemPreferences.getMediaAccessStatus`].
+
+### `desktopCapturer.getMediaSourceIdForWebContents(webContentsId)`
+
+* `webContentsId` number - Id of the WebContents to get stream of
+
+Returns `Promise<string>` - Resolves with the identifier of a WebContents stream, this identifier can be
+    used with [`navigator.mediaDevices.getUserMedia`].
+    The identifier is **only valid for 10 seconds**.
+    The identifier may be empty if not requested from a renderer process.
 
 [`navigator.mediaDevices.getUserMedia`]: https://developer.mozilla.org/en/docs/Web/API/MediaDevices/getUserMedia
 [`systemPreferences.getMediaAccessStatus`]: system-preferences.md#systempreferencesgetmediaaccessstatusmediatype-macos
--- a/docs/api/structures/web-contents-stream-result.md
+++ b/docs/api/structures/web-contents-stream-result.md
@@ -1,0 +1,6 @@
+# WebContentsStreamResult Object
+
+* `id` String - The identifier of a WebContents stream, this identifier can be
+  used with [`navigator.webkitGetUserMedia`]. The identifier may be empty if
+  not requested from a renderer process.
+* `mediaId` String - The DesktopMediaID associated with the WebContents.
--- a/docs/api/structures/web-contents-stream-result.md
+++ b/docs/api/structures/web-contents-stream-result.md
@@ -1,6 +1,0 @@
-# WebContentsStreamResult Object
-
-* `id` String - The identifier of a WebContents stream, this identifier can be
-  used with [`navigator.webkitGetUserMedia`]. The identifier may be empty if
-  not requested from a renderer process.
-* `mediaId` String - The DesktopMediaID associated with the WebContents.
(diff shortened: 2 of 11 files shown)


docs/api/desktop-capturer.md
```diff
@@ -72,6 +72,50 @@ const constraints = {
 }
 ```
 
+This example shows how to capture a video from a [WebContents](web-contents.md)
+
+```javascript
+// In the renderer process.
+const { desktopCapturer, remote } = require('electron')
+
+desktopCapturer.getWebContentsStream({ webContentsId: getCurrentWebContents().id }).then(async result => {
```
zcbenz:
```suggestion
desktopCapturer.getWebContentsStream({ webContentsId: remote.getCurrentWebContents().id }).then(async result => {
```

shell/browser/media/media_stream_devices_controller.cc
```diff
@@ -209,4 +217,28 @@ void MediaStreamDevicesController::HandleUserMediaRequest() {
...
+    content::DesktopMediaID video_media_id =
+        content::DesktopMediaID::Parse(request_.requested_video_device_id);
+
+    if (video_media_id.type != content::DesktopMediaID::TYPE_NONE) {
+      devices.emplace_back(blink::mojom::MediaStreamType::DISPLAY_VIDEO_CAPTURE,
+                           video_media_id.ToString(), "Display Video");
+    }
+  }
+
+  VLOG(0) << "DEVICES " << devices.size();
```
zcbenz: This can be removed.

Volune: Removed that whole file change, it was actually an experiment to use `getDisplayMedia` and should not have been committed.

nornagon:
cc @electron/wg-api

docs/api/structures/web-contents-stream-result.md
```diff
@@ -0,0 +1,6 @@
+# WebContentsStreamResult Object
+
+* `id` String - The identifier of a WebContents stream, this identifier can be
+  used with [`navigator.webkitGetUserMedia`]. The identifier may be empty if
+  not requested from a renderer process.
+* `mediaId` String - The DesktopMediaID associated with the WebContents.
```
loc: When might a consumer need to know what the `mediaId` is?

docs/api/desktop-capturer.md
```diff
@@ -94,6 +138,13 @@ Returns `Promise<DesktopCapturerSource[]>` - Resolves with an array of [`Desktop
 **Note** Capturing the screen contents requires user consent on macOS 10.15 Catalina or higher,
 which can detected by [`systemPreferences.getMediaAccessStatus`].
 
+### `desktopCapturer.getWebContentsStream(options)`
```
loc: Maybe `getWebContentsStreamInfo`? It's not really a stream yet.

docs/api/structures/web-contents-stream-result.md
Volune: I exposed it thinking it may be useful for debugging. I don't think it has any other uses at the moment.
`id` looks like `jzR0r6ua6Ixl5//h8NZv5w==`, `mediaId` looks like `web-contents-media-stream://4:3`

docs/api/desktop-capturer.md
Volune: Right

At some point I was considering returning a `DesktopCapturerSource`, but:
- the description of `DesktopCapturerSource.id` would not match this case (not "screen:XX" format, that's the "mediaId")
- don't know if it should return the "mediaId" in the `display_id` field or not, since that's different format
- `name`, `thumbnail`, `appIcon` are not provided
- At this point, the returned `id` is expected to be used, otherwise it will leak inside `DesktopStreamsRegistry`. This behavior is not similar to screen/application sources.

I don't know if `getWebContentsSource` would be a good name or more confusing.

loc:
@Volune what do you think about moving this to be on WebContents? I think `contents.getCaptureMediaId()` or something like that would provide better ergonomics. Otherwise I think this is looking pretty good.

jpmcgrath:
@MarshallOfSound just wondering if this is close to getting approved/merged? I'd love to be able to try of this feature. It would be great to know If this is just around the corner as I could abort my n00bish attempt to build electron from this branch :-)

Apologies if this question sounds demanding/entitled, or is not in keeping with electron community expectations/conventions. Thanks everyone for your efforts with this great framework!

Volune:
> @Volune what do you think about moving this to be on WebContents? I think `contents.getCaptureMediaId()` or something like that would provide better ergonomics. Otherwise I think this is looking pretty good.

@loc  The issue I have with that is I need to know the requesting WebContents in order to call `content::DesktopStreamsRegistry::GetInstance()->RegisterStream`. I don't know how to do that from WebContents. Also a benefit of `desktopCapturer` is it's available in rendering thread, so there always is a "request_web_contents"
One solution could be to split into `WebContents.getMediaId()` and `deskopCapturer.registerWebContentsStream()`
Maybe another solution would be to patch Blink to stop verifying that the stream has been registered, which could make sense for Electron.

(I feel like I already replied that, I guess I forgot to submit my response ☹️ )

----
> just wondering if this is close to getting approved/merged?

The API is not stabilized, so probably not. I will try to update the PR (rebase on top of master) during next week.

gamerz1990:
I am in need for this feature hope  it will 🙏 done soon 😊

loc:
Sorry for dropping the ball on this @Volune. I spent some time today thinking about how we could make this better, but couldn't come up with anything that didn't involve significant patching. In the future, I'll try to find an opportunity to improve Chrome's deference to embedders on the issue of screen capture.

One last minor suggestion. I've gone back and forth, but I think I'd prefer to change the signature from:
```
desktopCapturer.getWebContentsStream({ webContentsId: number }): Promise<{ id: string, mediaId: string}>
```
to
```
desktopCapturer.getMediaSourceIdForWebContents(webContentsId: number): Promise<string>
```
The name is more clear about what it's returning and, though it's reasonable to want to use options objects here to make this more extensible, I think that is pretty unlikely and comes at the cost of a stranger API. Also, since the mediaId is not useful to the consumer, we can safely get rid of it.

Also, we should add a note in the docs that [this media source ID is only valid for 10 seconds after it's called](https://source.chromium.org/chromium/chromium/src/+/master:content/browser/media/capture/desktop_streams_registry_impl.cc;l=74;bpv=1;bpt=1?q=DesktopStreamsRegistryImpl::RegisterStream&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2Fsearch%2F).

WDYT?

Volune:
@loc ok for me

I rebased on top of master, and made the API changes. Tests passed on my Windows, and a small demo app worked too.

lib/renderer/api/desktop-capturer.ts
```diff
@@ -37,3 +37,7 @@ export async function getSources (options: Electron.SourcesOptions) {
 
   return deserialize(sources);
 }
+
+export function getMediaSourceIdForWebContents (webContentsId: number) {
+  return ipcRendererInternal.invoke<Electron.WebContentsStreamResult>('ELECTRON_BROWSER_DESKTOP_CAPTURER_GET_MEDIA_SOURCE_ID_FOR_WEB_CONTENTS', webContentsId, getCurrentStack());
```
loc, line 42:
```suggestion
  return ipcRendererInternal.invoke<string>('ELECTRON_BROWSER_DESKTOP_CAPTURER_GET_MEDIA_SOURCE_ID_FOR_WEB_CONTENTS', webContentsId, getCurrentStack());
```

Volune:
- updated PR (sorry should have better reviewed last changes)
- CI test issue on mac seems unrelated to this PR (but I can't test on mac)
- Do I need to do something about release notes?

jkleinsc:
@Volune the failing mac tests were flakes that disappeared on a re-run.  In regards to the release notes, I updated your PR description to have a release note that matches the feature.  The `release-notes` status being marked as pending is an issue unrelated to this PR.

yozzh:
Hello, @Volune. I have a question about this feature. Is it possible to add alpha channel (transparency) when capturing tab? It will be very useful for third party widgets (ex. stickers)

PR #22704, opened 2020-03-16
feat: add support for net requests to use the session cookie store

### LINKED ISSUE #8891: Cookie handling in net module

### LINKED ISSUE #22541: Net request replaces cookie header with session cookies

MarshallOfSound:
Closes #8891
Fixes #22541

There is a common use case for wanting to send a `fetch`-like request from the main process.  The `net` module is designed to serve this use case but it unfortunately lacks support for using cookies the same way `fetch` would.  This leads to hacks and insecure direct usage of the `Cookies` header.

This new flag makes any `net` request use Chromium's internal cookie store logic.

Comes with tests and stuff 👍 

Notes: Added new `useSessionCookies` flag to `net` requests to allow them to use the session cookie store.

### DIFF
--- a/spec-main/api-net-spec.ts
+++ b/spec-main/api-net-spec.ts
@@ -511,6 +511,131 @@ describe('net module', () => {
       expect(response.headers['set-cookie']).to.have.same.members(cookie);
     });
 
+    it('should not use the sessions cookie store by default', async () => {
+      const serverUrl = await respondOnce.toSingleURL((request, response) => {
+        response.statusCode = 200;
+        response.statusMessage = 'OK';
+        response.setHeader('x-cookie', `${request.headers.cookie!}`);
+        response.end();
+      });
+      const sess = session.fromPartition('cookie-tests-1');
+      const cookieVal = `${Date.now()}`;
+      await sess.cookies.set({
+        url: serverUrl,
+        name: 'wild_cookie',
+        value: cookieVal
+      });
+      const urlRequest = net.request({
+        url: serverUrl,
+        session: sess
+      });
+      const response = await getResponse(urlRequest);
+      expect(response.headers['x-cookie']).to.equal(`undefined`);
+    });
+
+    it('should be able to use the sessions cookie store', async () => {
+      const serverUrl = await respondOnce.toSingleURL((request, response) => {
+        response.statusCode = 200;
+        response.statusMessage = 'OK';
+        response.setHeader('x-cookie', request.headers.cookie!);
+        response.end();
+      });
+      const sess = session.fromPartition('cookie-tests-2');
+      const cookieVal = `${Date.now()}`;
+      await sess.cookies.set({
+        url: serverUrl,
+        name: 'wild_cookie',
+        value: cookieVal
+      });
+      const urlRequest = net.request({
+        url: serverUrl,
+        session: sess,
+        useSessionCookies: true
+      });
+      const response = await getResponse(urlRequest);
+      expect(response.headers['x-cookie']).to.equal(`wild_cookie=${cookieVal}`);
+    });
+
+    it('should be able to use the sessions cookie store with set-cookie', async () => {
+      const serverUrl = await respondOnce.toSingleURL((request, response) => {
+        response.statusCode = 200;
+        response.statusMessage = 'OK';
+        response.setHeader('set-cookie', 'foo=bar');
+        response.end();
+      });
+      const sess = session.fromPartition('cookie-tests-3');
+      let cookies = await sess.cookies.get({});
+      expect(cookies).to.have.lengthOf(0);
+      const urlRequest = net.request({
+        url: serverUrl,
+        session: sess,
+        useSessionCookies: true
+      });
+      await collectStreamBody(await getResponse(urlRequest));
+      cookies = await sess.cookies.get({});
+      expect(cookies).to.have.lengthOf(1);
+      expect(cookies[0]).to.deep.equal({
+        name: 'foo',
+        value: 'bar',
+        domain: '127.0.0.1',
+        hostOnly: true,
+        path: '/',
+        secure: false,
+        httpOnly: false,
+        session: true
+      });
+    });
+
+    it('should be able to use the sessions cookie store safely across redirects', async () => {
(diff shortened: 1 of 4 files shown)


MarshallOfSound:
Side note, gonna follow this up later with a PR to make the `net` tests `async` because they're quite wild at the moment 😄

spec-main/api-net-spec.ts
```diff
@@ -633,6 +633,142 @@ describe('net module', () => {
...
+        response.setHeader('x-cookie', `${request.headers.cookie!}`)
+        response.end()
+      }).then(serverUrl => {
+        const sess = session.fromPartition('cookie-tests-1')
+        const cookieVal = `${Date.now()}`
+        sess.cookies.set({
+          url: serverUrl,
+          name: 'wild_cookie',
+          value: cookieVal
+        }).then(() => {
```
nornagon: in https://github.com/electron/electron/pull/22596 it looks like a flush might be required to ensure that this makes it to the store..? is that test wrong or is this one wrong?

MarshallOfSound: This one works 😄 🤷‍♂ Ran it quite a few times locally, flush is not required afaics

NilSet: The test previously was erroneously passing, and with your changes may be coincidentally passing. Can you check if this test still passes when you reverse the flag? It was my experience that the cookies needed to flush for the chromium machinery to pick them up.

MarshallOfSound: @NilSet There are two tests here, one for and one against.  They can't both be erroneously passing.

i.e. These tests are testing both the positive and negative cases 👍

deepak1556: Cookie flush is needed only if you have a short running app and want to ensure that they get persisted on shutdown. Adding flush here has no negative side effect but it isn't needed.

spec-main/api-net-spec.ts
```diff
@@ -633,6 +633,142 @@ describe('net module', () => {
...
+            url: localhostUrl,
+            name: 'wild_cookie',
+            value: cookieLocalVal
+          })
+        ]).then(() => {
+          const urlRequest = net.request({
+            url: serverUrl,
+            session: sess,
+            useSessionCookies: true
+          })
```
deepak1556: Can you handle `urlRequest.on('redirect')` block explicitly and check it has the `serverUrl` cookies. Just for the sake of completeness.

jkleinsc:
The releases WG has approved this for backporting to 8-x-y and 9-x-y.

MarshallOfSound:
Releases WG async approved this for `7.2.x` --> https://electronhq.slack.com/archives/CC80G2R6H/p1584742028067600

PR #22706, opened 2020-03-16
refactor: precache the IsWindowStateEvent() XAtom

ckerr:
#### Description of Change

Don't look up the `_NET_WM_STATE` XAtom every time we get an XEvent. Instead, get it when the `WindowStateWatcher` is constructed and keep it in a private const field. XEvents are high volume -- for example, moving a mouse across the screen can trigger dozens of XMouseEvents per second -- so looking it up in each event handler call is expensive over time.

Background info: what we're avoiding in each event handler call is a call to [XInternAtom()](https://github.com/mirror/libX11/blob/b676e62377483df77bcb6472d26b24f901323fa9/src/IntAtom.c#L56) which hashes the key, then does a linear walk through a list of interned keys & returns a match when found. 

No particular stakeholder; reviews welcomed from anyone :slightly_smiling_face: 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Improved window events handler efficiency on Linux.

### DIFF
--- a/shell/browser/ui/x/window_state_watcher.cc
+++ b/shell/browser/ui/x/window_state_watcher.cc
@@ -9,7 +9,9 @@
 namespace electron {
 
 WindowStateWatcher::WindowStateWatcher(NativeWindowViews* window)
-    : window_(window), widget_(window->GetAcceleratedWidget()) {
+    : window_(window),
+      widget_(window->GetAcceleratedWidget()),
+      window_state_atom_(gfx::GetAtom("_NET_WM_STATE")) {
   ui::X11EventSource::GetInstance()->AddXEventObserver(this);
 }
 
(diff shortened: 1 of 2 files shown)


ckerr:
> Curious if you tried a benchmark on how much this perf gains ?

Actually yes that's how I found it -- I was profiling startup time and in perf and hotspot and looking for low-hanging fruit. Unfortunately nearly all of it is out of our control -- e.g. parsing the JS and instantiating Isolates are very expensive -- so this isn't a *huge* win but it's not bad for such a small-scope patch:

| Function | Version | Cycles (Incl.) | 
|:---|:---|:---|
| WindowStateWatcher::WillProcessXEvent | before | 0.119% | 
| WindowStateWatcher::WillProcessXEvent | PR     | 0.00794% |
| WindowStateWatcher::DidProcessXEvent  | before | 0.0379% |
| WindowStateWatcher::DidProcessXEvent  | PR     | 0.00833% |

PR #22707, opened 2020-03-16
build: update release scripts to handle a remote master

MarshallOfSound:
One day these scripts will be flawless, today is not that day

Notes: no-notes

### DIFF
--- a/script/lib/utils.js
+++ b/script/lib/utils.js
@@ -82,7 +82,7 @@ async function getCurrentBranch (gitDir) {
       '--remote'
     ], gitDir)).split('\n')
 
-    branch = branches.filter(b => b.trim() === 'master' || RELEASE_BRANCH_PATTERN.test(b.trim()))[0]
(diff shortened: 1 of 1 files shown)


script/lib/utils.js
```diff
@@ -82,7 +82,7 @@ async function getCurrentBranch (gitDir) {
       '--remote'
     ], gitDir)).split('\n')
 
-    branch = branches.filter(b => b.trim() === 'master' || RELEASE_BRANCH_PATTERN.test(b.trim()))[0]
+    branch = branches.filter(b => b.trim() === 'master' || b.trim() === 'origin/master' || RELEASE_BRANCH_PATTERN.test(b.trim()))[0]
```
codebytere: this should probably be a little more general for forks: e.g `upstream/master` should work as well

MarshallOfSound: We don't handle `upstream/` as a prefix below, is there a case in sudowoodo where the remote could be called `upstream`?

codebytere: verdict: follow-up concern :)

PR #22710, opened 2020-03-16
fix: crash on invalid zoomFactor

codebytere:
Backport of https://github.com/electron/electron/pull/22673.

See that PR for more details.

Notes: Fixed a potential crash on invalid `zoomFactor` values when setting the zoom factor of a webpage.

### DIFF
--- a/shell/browser/api/atom_api_web_contents.cc
+++ b/shell/browser/api/atom_api_web_contents.cc
@@ -4,6 +4,7 @@
 
 #include "shell/browser/api/atom_api_web_contents.h"
 
+#include <limits>
 #include <memory>
 #include <set>
 #include <string>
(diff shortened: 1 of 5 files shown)


shell/browser/api/atom_api_web_contents.cc
```diff
@@ -2304,7 +2305,12 @@ double WebContents::GetZoomLevel() const {
   return zoom_controller_->GetZoomLevel();
 }
 
-void WebContents::SetZoomFactor(double factor) {
+void WebContents::SetZoomFactor(mate::Arguments* args, , double factor) {
```
jkleinsc:
```suggestion
void WebContents::SetZoomFactor(mate::Arguments* args, double factor) {
```

zcbenz:
The test is failing:

```
webContents module "after each" hook for "throws on an invalid zoomFactor" - "after each" hook for "throws on an invalid zoomFactor"
/Users/distiller/project/src/electron/spec/api-web-contents-spec.js
moreError: Object has been destroyed
```

zcbenz:
Maybe using `webContents.create` to create a headless WebContents to avoid the exception?

jkleinsc:

PR #22726, opened 2020-03-17
refactor: EventEmitters without gin_helper

nornagon:
#### Description of Change
This adds the infrastructure needed to refactor EventEmitter objects away from `gin_helper` and onto almost-completely-standard `gin`. (One small patch to gin is required to allow us to customize prototype chains.)

Also included, an example refactoring of the `Debugger` API object to use the new infrastructure. I expect to refactor several more classes using this infrastructure as followup PRs.

Of note, this also means we no longer need to call `Object.setPrototype(SomeApi.prototype, EventEmitter.prototype)` from JavaScript on EventEmitter objects, as that step will be done automatically in C++.

As a followup, I'd like to move the new EventEmitter infrastructure from `shell/common` to `shell/browser`, as it is not supported in the renderer process.

The method for passing `EventEmitter`'s prototype in to C++ was inspired by Node's similar handling of `Buffer.prototype`: [1](https://github.com/nodejs/node/blob/ec0dd6fa1cfa03b3983419acedb27ef95d297c82/lib/internal/bootstrap/node.js#L303), [2](https://github.com/nodejs/node/blob/d4660aba6324096198bc17772ba16fce6259967d/src/node_buffer.cc#L1147-L1153), [3](https://github.com/nodejs/node/blob/605615e5f33df41a307a0921f451b6dabb89410c/src/node_buffer.cc#L246-L257). Thanks to @codebytere for finding this reference.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


lib/browser/init.ts
```diff
@@ -15,6 +16,8 @@ require('../common/reset-search-paths')
 // Import common settings.
 require('@electron/internal/common/init')
 
+process.electronBinding('event_emitter').setEventEmitterPrototype(EventEmitter.prototype)
```
MarshallOfSound: For sanity / safety this should probably be called before `reset-search-paths` as that's what prepares the `electron` module that you "require"

nornagon: Happy to do so but I'm not quite sure why that would make a difference here? Can any user code run between `reset-search-paths` and this line?

nornagon:
CI failures are unrelated.

shell/browser/event_emitter_mixin.h
```diff
@@ -0,0 +1,80 @@
...
+ private:
+  // this.emit(name, event, args...);
+  template <typename... Args>
+  static bool EmitWithEvent(v8::Isolate* isolate,
+                            v8::Local<v8::Object> wrapper,
+                            base::StringPiece name,
+                            v8::Local<v8::Object> event,
+                            Args&&... args) {
+    v8::Locker locker(isolate);
+    v8::HandleScope handle_scope(isolate);
```
zcbenz: The locker and handle scope is unneeded here as the method takes `v8::Local` as argument.

deepak1556:
`EventEmitter` are extended currently through `TrackableObject` , will that be separated in follow-up PR as refactor happens on a case-by-case basis ?

nornagon:
@deepak1556 I'd like to move away from `TrackableObject` as much as possible. I'm not quite sure yet what my approach will be for classes which actually use the functionality of `TrackableObject` (which afaict is just maintaining a list of all the ones that exist), but many classes that inherit from `TrackableObject` currently don't appear to use that functionality.

PR #22732, opened 2020-03-18
refactor: ginify session.netLog

nornagon:
#### Description of Change
This refactors `netLog` to use `gin::Wrappable` instead of `gin_helper::TrackableObject`.

Additionally, it removes the deprecated `currentlyLoggingPath` and the `String` return value of `netLog.stopLogging`.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Removed the deprecated `currentlyLoggingPath` property of `netLog`. Additionally, `netLog.stopLogging` no longer returns the path to the recorded log.

### DIFF
--- a/shell/browser/api/electron_api_session.cc
+++ b/shell/browser/api/electron_api_session.cc
@@ -307,7 +307,6 @@ Session::~Session() {
   // Refs https://github.com/electron/electron/pull/12305.
   DestroyGlobalHandle(isolate(), cookies_);
   DestroyGlobalHandle(isolate(), protocol_);
-  DestroyGlobalHandle(isolate(), net_log_);
   g_sessions.erase(weak_map_id());
 }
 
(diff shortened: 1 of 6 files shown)


shell/browser/api/electron_api_session.cc
```diff
@@ -307,7 +307,6 @@ Session::~Session() {
   // Refs https://github.com/electron/electron/pull/12305.
   DestroyGlobalHandle(isolate(), cookies_);
   DestroyGlobalHandle(isolate(), protocol_);
-  DestroyGlobalHandle(isolate(), net_log_);
```
nornagon: NB. this is no longer needed because the object will be GC'd by V8.

deepak1556: Some session properties hold `scoped_refptr` to `ElectronBrowserContext` while some hold a weak-reference, we should ensure `ElectronBrowserContext` is released when session is destroyed and no other reference exits, otherwise bad things can happen on shutdown.

PR #22737, opened 2020-03-18
test: no need to loadURL in menu test

zcbenz:
There is no need to call `loadURL` when testing window menus. This should be able to make the `app.setApplicationMenu` tests less flaky.

#### Release Notes

Notes: no-notes

### DIFF
--- a/spec-main/fixtures/api/test-menu-null/main.js
+++ b/spec-main/fixtures/api/test-menu-null/main.js
@@ -3,7 +3,6 @@ const { app, BrowserWindow } = require('electron')
 let win
 app.whenReady().then(function () {
   win = new BrowserWindow({})
-  win.loadURL('about:blank')
   win.setMenu(null)
 
   setTimeout(() => {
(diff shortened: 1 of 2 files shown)


deepak1556:
What kind of flakiness was observed with `loadURL` ? navigation initiation should have been async.

Removing it is fine here but would like to understand the flakiness impact first.

zcbenz:
`electron-woa-testing` is failing frequently at the `Menu.setApplicationMenu` test:

```
Menu module Menu.setApplicationMenu does not override null menu on startup
Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\vsts-agent-win-x86-2.148.2\_work\1\s\src\electron\spec-main\api-menu-spec.ts)
```

Currently I only know the child app did not quit and I doubt it was caused by some dead lock on exit, I want to reduce factors of affects and see if the flaky test could be fixed.

PR #22738, opened 2020-03-18
feat(view-api): add ImageView

zcbenz:
#### Description of Change

This PR adds a simple `ImageView` with a single `setImage` method, this class is hidden under the `enable_view_api` build flag.

One use of this class, is to show a splash image when the window is created, and then replace the image view with WebContents once after the page is loaded. I have a written a simple demo:
https://gist.github.com/zcbenz/e0048264a2890c2b9e85fa91e5682b76

Once this is merged, I will start working on an RFC to make parts of the view APIs public, which would be a minimum set of APIs to enable implementing loading screens for apps (i.e. the APIs used by the demo above).

The feature of `ImageView` is currently rather limited, and I'll start improving it after the RFC is accepted.

Note that this PR does not touch our public API. 

#### Release Notes

Notes: no-notes

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -606,6 +606,8 @@ source_set("electron_lib") {
       "shell/browser/api/views/electron_api_box_layout.h",
       "shell/browser/api/views/electron_api_button.cc",
       "shell/browser/api/views/electron_api_button.h",
+      "shell/browser/api/views/electron_api_image_view.cc",
+      "shell/browser/api/views/electron_api_image_view.h",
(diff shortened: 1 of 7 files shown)


zcbenz:
> Wondering if this conflicts with refactor from #22726 and the class should be made a pure gin wrappable

Currently we have to use our `gin_helper::Wrappable` because:
1. we need to support the `new ImageView` style constructor;
2. the base class and child class must be able to have different prototypes.

They probably don't need to inherit from `gin_helper::TrackableObject` though.

PR #22747, opened 2020-03-18
chore: update app module property support

codebytere:
Backport of https://github.com/electron/electron/pull/22713.

See that PR for details.

Notes: none.

### DIFF
--- a/lib/browser/api/app.ts
+++ b/lib/browser/api/app.ts
@@ -1,7 +1,7 @@
 import * as fs from 'fs'
 import * as path from 'path'
 
-import { deprecate, Menu } from 'electron'
+import { Menu } from 'electron'
 import { EventEmitter } from 'events'
 
 const bindings = process.electronBinding('app')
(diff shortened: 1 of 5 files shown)


lib/browser/api/app.ts
```diff
@@ -17,6 +17,29 @@ let dockMenu: Electron.Menu | null = null
 Object.setPrototypeOf(App.prototype, EventEmitter.prototype)
 EventEmitter.call(app as any)
 
+// Properties.
+
+const nativeASGetter = app.isAccessibilitySupportEnabled
+const nativeASSetter = app.setAccessibilitySupportEnabled
+Object.defineProperty(App.prototype, 'accessibilitySupportEnabled', {
+  get: () => nativeASGetter.call(app),
```
zcbenz: Maybe use `get: function () { this.isAccessibilitySupportEnabled() }` or 
`get: () => app.isAccessibilitySupportEnabled()`?

One benefit is, if user monkey-patches the method the property would also be patched automatically.

PR #22751, opened 2020-03-18
refactor: ginify powerMonitor

nornagon:
#### Description of Change
This migrates `powerMonitor` to gin. A challenge here was to make sure that the `powerMonitor` instance was lazily initialized. I think this would be better handled directly in C++, delaying the internal initialization of the object until the app is ready but allowing the JS API to be immediately usable. However, to keep the refactoring to a minimum, I replicated the old behavior but avoided the use of the `PowerMonitor` constructor, since we're not currently exposing that for gin objects.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


deepak1556:
> I think this would be better handled directly in C++, delaying the internal initialization of the object until the app is ready but allowing the JS API to be immediately usable

I would prefer its done as part of the PR, seems much cleaner. Are you thinking of something like 

```c++
PowerMonitor::Powermonitor() {
  Browser::Get()->ready_promise()->Then(base::BindOnce([](){
   // delayed stuff
 ));
}
```

and the js should only need to create the proxy

```js
// powerMonitor must be lazily accessed
const powerMonitor = new Proxy(createPowerMonitor(), {
  get (obj, prop) {
    if (!app.isReady()) throw new Error("The 'powerMonitor' module can't be used before the app is ready.")
    return obj[prop]
  }
})
```

nornagon:
@deepak1556 I was imagining make the entire API usable before `app.on('ready')` is emitted; just, it won't actually fire any events or do anything until the app is started. Then there's no need for any proxy shenanigans.

nornagon:
@deepak1556 refactor done PTAL

deepak1556:
Build failure with calling non-static members on linux.

LGTM on the refactor otherwise.

Does this need an wg-api approval ?

nornagon:
@deepak1556 it doesn't change the API significantly IMO, just removes a previously-illegal usage.

nornagon:
@zcbenz / @deepak1556 PTAL

lib/browser/api/power-monitor.ts
```diff
@@ -1,38 +1,46 @@
...
+          // keeps the C++ code informed about whether there are any listeners.
+          pm.setListeningForShutdown(this.listenerCount('shutdown') > 0);
+          this.on('newListener', (event) => {
+            if (event === 'shutdown') {
+              pm.setListeningForShutdown(this.listenerCount('shutdown') + 1 > 0);
+            }
+          });
+          this.on('removeListener', (event) => {
+            if (event === 'shutdown') {
+              pm.setListeningForShutdown(this.listenerCount('shutdown') - 1 > 0);
```
deepak1556: if there is some usage like

```
pm.on('shutdown', l1);
pm.on('shutdown', l2);
pm.off('shutdown', l1);  // listenerCount will be 1 but the call will 
                         // be `pm.setListeningForShutdown(false)` 
                         // and reset the shutdown handler, also unblocks
                        // shutdown while there is an active listener
```

nornagon: hm, i don't think that's true—adding a listener increases listener count by 1, removing a listener decreases it by 1. but the change in count doesn't happen until after the `removeListener` / `newListener` event is fired, so this `- 1` accounts for that.

nornagon: oops, missed this in the docs:

> The 'removeListener' event is emitted after the listener is removed.

will fix.

deepak1556: https://repl.it/repls/ExcitingColorlessDataset yeah that seems to be true for newListener, but removeListener seems to do the opposite.

PR #22760, opened 2020-03-18
fix: print from PDF viewer not working

codebytere:
#### Description of Change

Fixes the print button functionality in the PDF viewer.

pdf-print

cc @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed the print button functionality in the PDF viewer extension.

### DIFF
--- a/shell/renderer/renderer_client_base.cc
+++ b/shell/renderer/renderer_client_base.cc
@@ -68,6 +68,7 @@
 
 #if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
 #include "base/strings/utf_string_conversions.h"
+#include "chrome/renderer/pepper/chrome_pdf_print_client.h"
 #include "content/public/common/webplugininfo.h"
 #include "extensions/common/constants.h"
 #include "extensions/common/extensions_client.h"
@@ -155,6 +156,9 @@ void RendererClientBase::RenderThreadStarted() {
 
   extensions_renderer_client_.reset(new ElectronExtensionsRendererClient);
   extensions::ExtensionsRendererClient::Set(extensions_renderer_client_.get());
+
+  // Enables printing from Chrome PDF viewer.
+  pdf::PepperPDFHost::SetPrintClient(new ChromePDFPrintClient());
 
   thread->AddObserver(extensions_renderer_client_->GetDispatcher());
 #endif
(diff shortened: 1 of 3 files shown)


shell/renderer/renderer_client_base.cc
```diff
@@ -156,6 +157,9 @@ void RendererClientBase::RenderThreadStarted() {
   extensions_renderer_client_.reset(new ElectronExtensionsRendererClient);
   extensions::ExtensionsRendererClient::Set(extensions_renderer_client_.get());
 
+  // Enables printing from Chrome PDF viewer.
+  pdf::PepperPDFHost::SetPrintClient(new ChromePDFPrintClient());
```
deepak1556: NB: `SetPrintClient` stores the client in a thread local storage which doesn't do lifetime management, but in this case it is fine to leak as there is only one RenderThread per RenderClient in multi process mode and we don't support single process mode.

x4080:
Hi, I'm using "electron": "^9.0.2", and still can't print using print button (it like freeze), what version should I use to enable native printing ? Thanks

sgil-carecloud:
i still cannot save after i hit print

gavi:
It freezes on hitting the print button - Does this have something to do with permissions?

hozikm:
Same issue on Electron 10.1.3 – when I hit the print button, I can select printer in system dialog, but nothing is printed and whole PDF-viewer GUI became unresponsive. The problem occurs both on Windows and Mac. 

(PDF is client-generated and opened as the internal blob URL. Every other controls in PDF-viewer works fine.)

rlaffers:
I can confirm this issue exists on Linux as well.
Electron 10.1.3, Ubuntu 20.04

Clicking the print icon in Pdf viewer --> opens the system print dialog correctly.
Next, selecting a printer and clicking the Print button --> the file is never printed (not even the Print to PDF option works). The whole PDF viewer freezes.

heitara:
I can confirm that this issue is still presented on macOS Catalina 10.15.6. I'm using  "electron": "11.0.0-beta.11". The issue was there when using electron 10.X (haven't checked the most recent release).

The problem is that once I try to save the pdf from the native print dialog - the PDF preview is frozen. If this iframe is closed the app still works. So, I guess, only the plugin hangs somehow. This is the same on windows 10.

codebytere:
Yep! We're aware and working on it. There is already a tracking issue as noted in the comment directly above yours.

dtboy1995:
mark

sujiewen:
test 11.1.x still not work

heitara:
I'm using "^11.3.0" and I can confirm it's working. I'm not sure which is the minimal version that introduces the fix.

m4heshd:
Tested on `12.0.0`. Works perfectly fine.

PR #22771, opened 2020-03-19
feat: add property support for remainder of BrowserWindow

codebytere:
#### Description of Change

Adds property-based support for the remainder of primitive gette/setter pairs on `BrowserWindow`.

Namely:
- `win.simpleFullScreen`
- `win.title`
- `win.visibleOnAllWorkspaces`
- `win.documentEdited`
- `win.representedFilename`
- `win.shadow`
- `win.kiosk`
- `win.menuBarVisible`

cc @electron/wg-api

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added support for property access to some getter/setter pairs on `BrowserWindow`.


zcbenz:
Should the properties be put in `top-level-window.js` instead of `browser-window.js`? They are essentially methods in `TopLevelWindow`, and it would be helpful when making `TopLevelWindow` public after https://github.com/electron/governance/pull/254.

lib/browser/api/browser-window.js
```diff
@@ -80,11 +80,51 @@ BrowserWindow.prototype._init = function () {
...
+    set: (simple) => this.setSimpleFullScreen(simple)
+  })
+
+  Object.defineProperty(this, 'kiosk', {
+    get: () => this.isKiosk(),
+    set: (kiosk) => this.setKiosk(kiosk)
+  })
+
+  Object.defineProperty(this, 'documentEdited', {
+    get: () => this.isFullscreen(),
```
zcbenz: `typo` here.

lib/browser/api/browser-window.js
```diff
@@ -80,11 +80,51 @@ BrowserWindow.prototype._init = function () {
...
+    set: (edited) => this.setDocumentEdited(edited)
+  })
+
+  Object.defineProperty(this, 'fullscreen', {
```
zcbenz: Should be `fullScreen` to match the method name and `simpleFullScreen` property above.

lib/browser/api/browser-window.js
```diff
@@ -80,11 +80,51 @@ BrowserWindow.prototype._init = function () {
...
+    get: () => this.isFullScreen(),
+    set: (full) => this.setFullScreen(full)
+  })
+
+  Object.defineProperty(this, 'shadow', {
```
zcbenz: I think we should use `hasShadow` otherwise it would cause problems when writing documentations.

zcbenz:
I think we should probably write a helper to automatically generate the setter/getters, since the pattern is fixed.

zcbenz:
And we should define the property on the prototype instead of `this`, so we only need to do it once instead of every time when a new instance is created.

lib/browser/api/browser-window.js
codebytere: @zcbenz i did this intentionally since the constructor property is `fullscreen` with no capitalization and i felt it might confuse people 🤔

codebytere:
@zcbenz the problem is that it's fixed but not quite fixed _enough_, for example:

```
  Object.defineProperty(this, 'menuBarVisible', {
    get: () => this.isMenuBarVisible(),
    set: (visible) => this.setMenuBarVisibility(visible)
  })
```
seems like it should be `setMenuBarVisible` but isn't. Also, defining it on the prototype like so:

```
Object.defineProperty(BrowserWindow.prototype, 'title', {
  get: () => BrowserWindow.prototype.getTitle(),
  set: (title) => BrowserWindow.prototype.setTitle(title)
})
```

Results in this error:

Screen Shot 2020-03-20 at 8 50 03 AM

which is why i took the approach i did

lib/browser/api/browser-window.js
zcbenz: Maybe leave this to discussion on next meeting? I don't have a strong opposition on this but I think we should be consistent with new APIs and fix the inconsistent old ones.

zcbenz:
> the problem is that it's fixed but not quite fixed _enough_

This is reasonable.

> Also, defining it on the prototype like so Results in this error:

The correct way should be:

```js
Object.defineProperty(BrowserWindow.prototype, 'title', {
  get: function () { return this.getTitle() },
  set: function (title) { this.setTitle(title) }
})
```

The point is to use `function` to have `this` passed.

lib/browser/api/browser-window.js
codebytere: Works for me! i agree that we should aim for consistency here.

zcbenz: @codebytere Maybe leave `fullscreen` out in this PR, or use `fullScreen` temporarily? From the discussion in @electron/wg-api it seems that everyone agrees we should have a consistent API but the problem is to whether use `fullscreen` or `fullScreen`.

If we are going to use `fullscreen` then we have to deprecate a few APIs, so no whether we are doing it or not the change would be better put in a separate PR.

lib/browser/api/browser-window.js
codebytere: The getter is already called `hasShadow`, so it'll error on duplicate identifiers if i do so

lib/browser/api/browser-window.js
```diff
@@ -72,44 +72,84 @@ BrowserWindow.prototype._init = function () {
...
+};
 
-  // Properties
+// Properties
 
-  Object.defineProperty(this, 'autoHideMenuBar', {
-    get: () => this.isMenuBarAutoHide(),
-    set: (autoHide) => this.setAutoHideMenuBar(autoHide)
-  });
+Object.defineProperty(BrowserWindow.prototype, 'autoHideMenuBar', {
```
zcbenz, line 79: Can you move the properties to `TopLevelWindow`? These are essentially properties of `TopLevelWindow` and inherited by `BrowserWindow`, once we make `TopLevelWindow` public per https://github.com/electron/governance/pull/254 we will have to move the properties to it anyway.

PR #22774, opened 2020-03-19
feat: add support for optimizing simple objects and functions over the context bridge

MarshallOfSound:
The contextBridge clone algorithm is slow (about on par with our `base::Value` converter, but slower than the V8 structured clone algorithm).  This PR adds two new methods to allow apps to force different parts of the context bridge (individual functions and entire objects) to be sent over the bridge in "simple" mode that exclusively uses the structured cloning algorithm.  This can result in upwards of 3x speed improvements for sending large objects over the bridge.

Screen Shot 2020-03-19 at 11 41 40 AM

Recommend reviewing with [whitespace diff disabled](https://github.com/electron/electron/pull/22774/files?diff=split&w=1), there aren't actually that many changes here.

Notes: Added new `contextBridge.prepareSimpleFunction` and `contextBridge.exposeSimpleObjectInMainWorld` APIs to allow apps to potentially send data over the context bridge faster.

### DIFF
--- a/docs/api/context-bridge.md
+++ b/docs/api/context-bridge.md
@@ -44,7 +44,30 @@ The `contextBridge` module has the following methods:
 ### `contextBridge.exposeInMainWorld(apiKey, api)` _Experimental_
 
 * `apiKey` String - The key to inject the API onto `window` with.  The API will be accessible on `window[apiKey]`.
-* `api` Record<String, any> - Your API object, more information on what this API can be and how it works is available below.
+* `api` Record<String, any> - Your API object, more information on what this API can be and how it works is available [below](#parameter--error--return-type-support)).
+
+### `contextBridge.exposeSimpleObjectInMainWorld(objectKey, object)` _Experimental_
+
+* `objectKey` String - The key to inject the object onto `window` with.  The object will be accessible on `window[objectKey]`.
+* `object` Record<String, any> - Your object, the keys and values of this object must be **simple** data types from the [Type Support](#parameter--error--return-type-support)) table below.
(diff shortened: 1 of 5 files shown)


nornagon:
Oof... that sure is a bit of a bummer of an API.

Can we think of any way to get some/most of this performance win without pushing on the developer to decide whether their objects or functions are is "simple", and remembering to "prepare" all their "simple" functions?

MarshallOfSound:
> Can we think of any way to get some/most of this performance win without pushing on the developer to decide whether their objects or functions are is "simple", and remembering to "prepare" all their "simple" functions?

We can try and make our existing traversal and conversion logic faster but it'll never going to be as fast as V8's alogorithm.  We also can't "detect" automatically whether to use smart or simple.  I tried happy-path optimizing for "try to use the V8 serializer and if it fails fallback to our logic" but that made it noticeably slower for quite a few cases

nornagon:
@MarshallOfSound could you expand on the cases where that's slower? I'd imagine that large objects are generally the ones you'd really get the win on for having them be "simple", whereas small objects are more likely to contain functions / other non-serializable things. If we get slower on small objects (which are generally fast anyway), but faster on big objects, that seems like a reasonable compromise in exchange for a much simpler API.

jkleinsc:
This PR was reviewed by the API WG on [April 27th ](https://github.com/electron/governance/blob/master/wg-api/meeting-notes/2020-04-27.md).

zcbenz:
I'm adding `wip` tag since there are pending work to do.

PR #22777, opened 2020-03-19
build: update eslint + eslint plugins

MarshallOfSound:
Split this into easy commits to review, most of the changes are automated.  Then a few new rules kicked in and the following commits are one-commit-per-new-rule

In good news (and part of the reason for doing this) one of the new rules blocks `.hasOwnProperty` which would have prevented a bug that crept in a few releases back.

Notes: no-notes


lib/browser/api/net.js
```diff
@@ -399,7 +401,7 @@ class ClientRequest extends Writable {
     this._urlLoader.on('redirect', (event, redirectInfo, headers) => {
       const { statusCode, newMethod, newUrl } = redirectInfo
       if (this._redirectPolicy === 'error') {
-        this._die(new Error(`Attempted to redirect, but redirect policy was 'error'`))
+        this._die(new Error('Attempted to redirect, but redirect policy was \'error\''))
```
nornagon: this seems worse IMO

lib/renderer/web-view/guest-view-internal.ts
```diff
@@ -24,10 +24,10 @@ const WEB_VIEW_EVENTS: Record<string, Array<string>> = {
   'did-frame-navigate': ['url', 'httpResponseCode', 'httpStatusText', 'isMainFrame', 'frameProcessId', 'frameRoutingId'],
   'did-navigate-in-page': ['url', 'isMainFrame', 'frameProcessId', 'frameRoutingId'],
   'focus-change': ['focus', 'guestInstanceId'],
-  'close': [],
-  'crashed': [],
+  close: [],
+  crashed: [],
   'plugin-crashed': ['name', 'version'],
-  'destroyed': [],
+  destroyed: [],
```
nornagon: this seems worse, but meh i guess?

lib/content_script/init.js
```diff
@@ -15,7 +15,7 @@ v8Util.setHiddenValue(global, 'ipc-internal', v8Util.getHiddenValue(isolatedWorl
 // The process object created by webpack is not an event emitter, fix it so
 // the API is more compatible with non-sandboxed renderers.
 for (const prop of Object.keys(EventEmitter.prototype)) {
-  if (process.hasOwnProperty(prop)) {
+  if (Object.prototype.hasOwnProperty.call(process, prop)) {
```
nornagon: This could perhaps be better done with `Object.getOwnPropertyNames`?

lib/renderer/web-view/web-view-impl.ts
```diff
@@ -173,7 +173,7 @@ export class WebViewImpl {
     }
 
     for (const attributeName in this.attributes) {
-      if (this.attributes.hasOwnProperty(attributeName)) {
+      if (Object.prototype.hasOwnProperty.call(this.attributes, attributeName)) {
```
nornagon: `Object.getOwnPropertyNames`

lib/sandboxed_renderer/init.js
```diff
@@ -20,7 +20,7 @@ v8Util.setHiddenValue(global, 'ipc-internal', new EventEmitter())
 // The process object created by webpack is not an event emitter, fix it so
 // the API is more compatible with non-sandboxed renderers.
 for (const prop of Object.keys(EventEmitter.prototype)) {
-  if (process.hasOwnProperty(prop)) {
+  if (Object.prototype.hasOwnProperty.call(process, prop)) {
```
nornagon: here too

lib/content_script/init.js
MarshallOfSound: They are not functionally equivalent

image

lib/renderer/web-view/web-view-impl.ts
MarshallOfSound: See above

lib/content_script/init.js
nornagon: Ah... because `constructor` is included in `getOwnPropertyNames`. ok.

PR #22783, opened 2020-03-20
build: enable JS semicolons

MarshallOfSound:
ASI is an **error correcting strategy**, it has exceptions instead of being a steadfast rule and all our new projects in the `electron` org are standardizing on `prettier` with semicolons enabled.  This is a quick `--fix` run to put semicolons everywhere, making code more legible, easier to parse and standard across the electron org

Notes: no-notes

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,6 +6,7 @@
     "browser": true
   },
   "rules": {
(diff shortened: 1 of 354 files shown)


adjobscc-del (commented):
ok

PR #22788, opened 2020-03-20
fix: allow net requests to use Same-Site cookies

MarshallOfSound:
It's expected that when the `net` module sends cookies, it sends all eligible cookies.  The default behavior of Chromium is to _not_ send same site cookies.  We should so that users can achieve with `net` what they could do with `fetch` 👍 

Currently targeting the `net-cookie-option` branch till that lands as it uses the same testing infrastructure.

Notes: Fixed issue where `SameSite` cookies would not be attached to outgoing requests from the `net` module


felixrieseberg:
<3 that there is a test, thank you so much!

PR #22789, opened 2020-03-20
feat: expose the sameSite value for cookies

MarshallOfSound:
This would have helped a lot while comparing cookies debugging a `net` issue earlier.  Exposing this is easy and aligns with the information about a cookie that devtools exposes.

Notes: Added a `sameSite` property to the Cookie structure allowing you to get and set the same site policy for cookies

### DIFF
--- a/docs/api/cookies.md
+++ b/docs/api/cookies.md
@@ -96,7 +96,7 @@ the response.
   * `expirationDate` Double (optional) - The expiration date of the cookie as the number of
     seconds since the UNIX epoch. If omitted then the cookie becomes a session
     cookie and will not be retained between sessions.
-  * `sameSite` String (optional) - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy to apply to this cookie.  Can be `unspecified`, `norestriction`, `lax` or `strict`.  Default is `no_restriction`.
+  * `sameSite` String (optional) - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy to apply to this cookie.  Can be `unspecified`, `norestriction`, `lax` or `strict`.  Default is `norestriction`.
 
 Returns `Promise<void>` - A promise which resolves when the cookie has been set
 
--- a/docs/api/cookies.md
+++ b/docs/api/cookies.md
@@ -96,7 +96,7 @@ the response.
   * `expirationDate` Double (optional) - The expiration date of the cookie as the number of
     seconds since the UNIX epoch. If omitted then the cookie becomes a session
     cookie and will not be retained between sessions.
-  * `sameSite` String (optional) - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy to apply to this cookie.  Can be `unspecified`, `norestriction`, `lax` or `strict`.  Default is `norestriction`.
+  * `sameSite` String (optional) - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy to apply to this cookie.  Can be `unspecified`, `no_restriction`, `lax` or `strict`.  Default is `no_restriction`.
 
 Returns `Promise<void>` - A promise which resolves when the cookie has been set
 
--- a/docs/api/cookies.md
+++ b/docs/api/cookies.md
@@ -96,6 +96,7 @@ the response.
   * `expirationDate` Double (optional) - The expiration date of the cookie as the number of
     seconds since the UNIX epoch. If omitted then the cookie becomes a session
     cookie and will not be retained between sessions.
+  * `sameSite` String (optional) - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy to apply to this cookie.  Can be `unspecified`, `norestriction`, `lax` or `strict`.  Default is `no_restriction`.
 
 Returns `Promise<void>` - A promise which resolves when the cookie has been set
 
--- a/docs/api/structures/cookie.md
+++ b/docs/api/structures/cookie.md
@@ -12,4 +12,4 @@
 * `expirationDate` Double (optional) - The expiration date of the cookie as
   the number of seconds since the UNIX epoch. Not provided for session
   cookies.
-* `sameSite` String - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy applied to this cookie.  Can be `unspecified`, `norestriction`, `lax`, `strict` or `unknown`.
+* `sameSite` String - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy applied to this cookie.  Can be `unspecified`, `no_restriction`, `lax`, `strict` or `unknown`.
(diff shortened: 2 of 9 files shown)


docs/api/cookies.md
```diff
@@ -96,6 +96,7 @@ the response.
   * `expirationDate` Double (optional) - The expiration date of the cookie as the number of
     seconds since the UNIX epoch. If omitted then the cookie becomes a session
     cookie and will not be retained between sessions.
+  * `sameSite` String (optional) - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy to apply to this cookie.  Can be `unspecified`, `norestriction`, `lax` or `strict`.  Default is `no_restriction`.
```
ckerr:
```suggestion
  * `sameSite` String (optional) - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy to apply to this cookie.  Can be `unspecified`, `norestriction`, `lax` or `strict`.  Default is `norestriction`.
```

shell/browser/api/electron_api_cookies.cc
```diff
@@ -171,6 +190,22 @@ std::string InclusionStatusToString(
   return "Setting cookie failed";
 }
 
+net::CookieSameSite StringToCookieSameSite(const std::string* str_ptr) {
+  if (!str_ptr)
+    return net::CookieSameSite::NO_RESTRICTION;
+  std::string str = *str_ptr;
```
ckerr: Copying`str_ptr` is unnecessary. If you want `str` for the readability of saying `str` instead of `*str_ptr`, you could use a const reference:

```suggestion
  const std::string& str = *str_ptr;
```

docs/api/cookies.md
```diff
@@ -96,6 +96,7 @@ the response.
   * `expirationDate` Double (optional) - The expiration date of the cookie as the number of
     seconds since the UNIX epoch. If omitted then the cookie becomes a session
     cookie and will not be retained between sessions.
+  * `sameSite` String (optional) - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy to apply to this cookie.  Can be `unspecified`, `norestriction`, `lax` or `strict`.  Default is `norestriction`.
```
flotwig: `no_restriction` (with an underscore) is the string used by the WebExtension API, should this API match that since it's already a convention? https://developer.chrome.com/extensions/cookies#type-SameSiteStatus

MarshallOfSound: @flotwig Thanks for the link, that's actually a good reason to put it back to having an underscore 👍

MarshallOfSound:
```suggestion
  * `sameSite` String (optional) - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy to apply to this cookie.  Can be `unspecified`, `no_restriction`, `lax` or `strict`.  Default is `no_restriction`.
```

shell/browser/api/electron_api_cookies.cc
```diff
@@ -29,6 +29,24 @@ using content::BrowserThread;
...
 
+template <>
+struct Converter<net::CookieSameSite> {
+  static v8::Local<v8::Value> ToV8(v8::Isolate* isolate,
+                                   const net::CookieSameSite& val) {
+    switch (val) {
+      case net::CookieSameSite::UNSPECIFIED:
+        return ConvertToV8(isolate, "unspecified");
+      case net::CookieSameSite::NO_RESTRICTION:
+        return ConvertToV8(isolate, "norestriction");
```
MarshallOfSound:
```suggestion
        return ConvertToV8(isolate, "no_restriction");
```

shell/browser/api/electron_api_cookies.cc
```diff
@@ -171,6 +190,22 @@ std::string InclusionStatusToString(
   return "Setting cookie failed";
 }
 
+net::CookieSameSite StringToCookieSameSite(const std::string* str_ptr) {
+  if (!str_ptr)
+    return net::CookieSameSite::NO_RESTRICTION;
+  const std::string& str = *str_ptr;
+  if (str == "unspecified") {
+    return net::CookieSameSite::UNSPECIFIED;
+  } else if (str == "norestriction") {
```
MarshallOfSound:
```suggestion
  } else if (str == "no_restriction") {
```

docs/api/structures/cookie.md
```diff
@@ -12,3 +12,4 @@
 * `expirationDate` Double (optional) - The expiration date of the cookie as
   the number of seconds since the UNIX epoch. Not provided for session
   cookies.
+* `sameSite` String - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy applied to this cookie.  Can be `unspecified`, `norestriction`, `lax`, `strict` or `unknown`.
```
MarshallOfSound:
```suggestion
* `sameSite` String - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy applied to this cookie.  Can be `unspecified`, `no_restriction`, `lax`, `strict` or `unknown`.
```

docs/api/structures/cookie.md
```diff
@@ -12,3 +12,4 @@
 * `expirationDate` Double (optional) - The expiration date of the cookie as
   the number of seconds since the UNIX epoch. Not provided for session
   cookies.
+* `sameSite` String - The [Same Site](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) policy applied to this cookie.  Can be `unspecified`, `no_restriction`, `lax`, `strict` or `unknown`.
```
nornagon, line 15: What does `unknown` mean here?

MarshallOfSound: It means we didn't update the converter :/

c++ isn't smart enough to know I covered every enum so we need a default case in the converter.  We have this `unknown` value in a few places where we convert from an enum

shell/browser/api/electron_api_cookies.cc
```diff
@@ -168,7 +188,24 @@ std::string InclusionStatusToString(
   if (status.HasExclusionReason(net::CanonicalCookie::CookieInclusionStatus::
                                     EXCLUDE_NONCOOKIEABLE_SCHEME))
     return "Cannot set cookie for current scheme";
-  return "Setting cookie failed";
+  LOG(WARNING) << "Cookie inclusion debug string: "
+               << status.GetDebugString() return "Setting cookie failed";
+}
+
+net::CookieSameSite StringToCookieSameSite(const std::string* str_ptr) {
```
nornagon, line 195: This should result in a JS exception being thrown if you typo the name, rather than silently defaulting to `no_restriction`

yarn.lock
```diff
@@ -25,20 +25,35 @@
...
     lodash.camelcase "^4.3.0"
-    markdown-it "^8.4.2"
+    markdown-it "^10.0.0"
     minimist "^1.2.0"
-    ora "^3.4.0"
-    pretty-ms "^5.0.0"
+    ora "^4.0.3"
+    pretty-ms "^5.1.0"
+
+"@electron/docs-parser@^0.7.2":
```
nornagon, line 43: we probably shouldn't be depending on two different versions of docs-parser..?

MarshallOfSound: I'm going to address this in a follow up, it looks awkward but doesn't actually negatively impact anything

shell/browser/api/electron_api_cookies.cc
MarshallOfSound: Done

PR #22810, opened 2020-03-23
build: fix beta version bumper logic for betas beyond 10

MarshallOfSound:
Manually parse out and sort the pre-release version number as a number instead of as a string to determine the next beta.  Otherwise `10 < 2` which is mathematically, uh, wrong

Notes: no-notes


jkleinsc:

PR #22812, opened 2020-03-23
refactor: ginify protocol

nornagon:
#### Description of Change
This refactors the `protocol` module to use `gin::Wrappable` instead of `gin_helper::TrackableObject`. Additionally, it also extracts the business logic into a new class, `ProtocolRegistry`, which is owned by `ElectronBrowserContext`. This simplifies the process of fetching the protocol information for a particular session.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


nornagon:
Blocked on #22820

shell/browser/protocol_registry.h
```diff
@@ -0,0 +1,53 @@
...
+#include "content/public/browser/content_browser_client.h"
+#include "shell/browser/net/electron_url_loader_factory.h"
+
+namespace content {
+class BrowserContext;
+}  // namespace content
+
+namespace electron {
+
+class ProtocolRegistry {
```
deepak1556: `ProtocolRegistry` can be a `BrowserContextKeyedServiceFactory` which is a singleton for the lifetime of a context.
https://source.chromium.org/chromium/chromium/src/+/master:components/keyed_service/content/browser_context_keyed_service_factory.h

nornagon: It looks like we don't use that pattern much in Electron. Can you expand on why you think it would be a good fit here?

deepak1556: BrowserContextKeyServiceFactory has the following benefits

1) Clearly defined lifetime and has access to the BrowserContext in a safe way
2) Ability to provide dependencies on different interfaces/services that also has lifetime tied to a BrowserContext
3) Can provide different interfaces depending on the BrowserContext

2 and 3 is not useful with the current implementation of `ProtocolRegistry`, my suggestion to adopt this now was just for 1 , since I think declaring it as a member restricts the flexibility of the interface in terms of what it can depend on or order of declaration, for extending in future. And yeah we don't use this pattern much except for the extension system and `NetworkContextServiceFactory`

The current implementation looks good and we can refactor into a keyed factory if need arises.

nornagon: Yeah, I think the lifetime of this is pretty clear as is, with it being a member of the BrowserContext. The safer access seems like it could potentially prevent some misuse, though I'm not aware of it being possible to create a `content::BrowserContext` that's not an `ElectronBrowserContext` currently in Electron, so it seems like that would be of limited value.

I can imagine it being useful to specify arbitrary orderings in future, but yeah, not needed right now. Good to know for future though!

PR #22821, opened 2020-03-24
refactor: ginify Notification

nornagon:
#### Description of Change
Also, add some tests.

#### Checklist

- [x] `npm test` passes
#### Release Notes

Notes: none


nornagon:
cc @sethlu

nornagon:
@zcbenz hmm, that seems like it might be the intended behavior on Windows in the underlying API?

If an app crashes, it won't be able to call any OS functions to remove notifications.

zcbenz:
It think this is similar to Tray icons, it is fine to leave a ghost icon there if the process crashed, but it should clear the icon if it exited normally.

But anyway I tried to find out the original issue about this but didn't succeed, let's just do this way.

zcbenz:
Also I think this should be marked as `feat` and documented for the same reason with https://github.com/electron/electron/pull/22822#issuecomment-604153428.

nornagon:
Blocking on #22822 for solutions to the same problems described above.

nornagon:
NB this removes the `.destroy` method from `Notification`, because it's not documented and not tested.

PR #22822, opened 2020-03-24
refactor: ginify Tray

nornagon:
#### Description of Change
This refactors the `Tray` object to use `gin::Wrappable` instead of `gin_helper::TrackableObject`.

Additionally, `gin_helper::EventEmitter::CreateWithFlags` has been refactored into a helper in `ui_event.h`, since it's called from both Menu and Tray. I don't think a dedicated method on `EventEmitter` makes sense here as `EmitCustomEvent` is easily callable with a custom event created by the helper.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


shell/browser/api/electron_api_tray.cc
```diff
@@ -236,14 +262,20 @@ void Tray::DisplayBalloon(gin_helper::ErrorThrower thrower,
...
+    return;
   tray_icon_->Focus();
 }
 
-void Tray::PopUpContextMenu(gin_helper::Arguments* args) {
+void Tray::PopUpContextMenu(gin::Arguments* args) {
+  if (!CheckDestroyed())
+    return;
   gin::Handle<Menu> menu;
   args->GetNext(&menu);
```
zcbenz: The `gin::Arguments` would increase `nextArg` counter even when `GetNext` fails, so `tray.popUpContextMenu(pos)` would not work with current parsing code.

zcbenz:
Would `((new Tray) instanceof Tray) === true`?

shell/browser/api/electron_api_tray.cc
nornagon: Ah, good catch. I've cleaned up the argument parsing logic and added some tests. The new code should be better in that it will properly throw errors for invalid argument types, rather than silently using the default behavior.

nornagon:
> Would `((new Tray) instanceof Tray) === true`?

Looks like no. Is this a requirement?

zcbenz:
> > Would `((new Tray) instanceof Tray) === true`?
> 
> Looks like no. Is this a requirement?

It is not an requirement, but I think we should mark this as `feat` since it breaks runtime type check, and it should be documented too as users would not expect this behavior from an `new Tray` API.

nornagon:
@zcbenz I updated the PR to make `Tray` constructible directly and factor out a generic mixin for this behavior, similar to `EventEmitterMixin`. It's a bit of a hack because it has to check directly for if you're trying to make an EventEmitter, but hopefully reusable.

zcbenz:
@nornagon Can you add code to ensure the tray icon is destroyed on exit? On Windows if you leak it, there would be a ghost icon left in the notification area even after the process is exited.

Previously it was done with `ElectronBrowserMainParts::RegisterDestructionCallback` by `TrackableObject`.

The other changes look good to me.

Issue #5107, opened 2016-04-10
Not allowed to load local resource: file://index.html/ after webpacking main.js


MihaiValentin:
I am trying to webpack all the main.js script & its dependencies in a single file (I want to have one file for the UI app, and one file for the server app).

If I use the normal source, index.html loads just fine. However, when webpacking, I get the error mentioned in the title.

Here's the webpack config I used:
```javascript
webpack({
    entry: [
        './main'
    ],
    output: {
        path: path.join(__dirname, 'asar'),
        filename: 'main.js',
        libraryTarget: 'commonjs2'
    },
    externals: ['electron'],
    target: 'node'
});
```

I load the file like this:
`mainWindow.loadURL('file://' + __dirname + '/index.html');`

I think that its perhaps due to webpack calling/evaling stuff outside of the electron context that allows serving local files.

Any ideas/suggestions? Thanks!

* Electron version: 0.37.2
* Operating system: Windows 10 Home

zcbenz:
You can probably try turning off `webSecurity` in `webPreferences` of `BrowserWindow`. For more questions I suggest seeking help from the [community](https://github.com/electron/electron#community).

singhshashi:
@MihaiValentin Hey, did you find a solution for this?

eiriarte:
@MihaiValentin @singhshashi I had the same problem. It seems webpack, by default, tries to "mock" Node globals like `__dirname`. I disabled this behavior with the `node.__dirname` option and… it worked!

Just in case, I'm also using [webpack-target-electron-renderer](https://github.com/chentsulin/webpack-target-electron-renderer) for the `target` option.

This is my config so far. I hope it helps:

```
var webpack = require('webpack');
var path = require('path');
var webpackTargetElectronRenderer = require('webpack-target-electron-renderer');

var BUILD_DIR = path.resolve(__dirname, 'build');
var APP_DIR = path.resolve(__dirname, 'src');

var config = {
  entry: APP_DIR + '/index.js',
  output: {
    path: BUILD_DIR,
    filename: 'index.js'
  },
  node: {
    __dirname: false,
    __filename: false
  },
  module : {
    loaders : [
      {
        test : /\.jsx?/,
        include : APP_DIR,
        exclude: /node_modules/,
        loader : 'babel'
      }
    ]
  }
};

config.target = webpackTargetElectronRenderer(config);

module.exports = config;
```

singhshashi:
[...]

PR #22824, opened 2020-03-24
chore: surface webcontents load error more readily

loc:
#### Description of Change
Surface failed load error messages a little more readily so that new folks aren't left as confused. 

e.g. https://github.com/electron/electron/issues/5107#issuecomment-281200144

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -118,6 +118,7 @@
 #include "shell/common/mouse_util.h"
 #include "shell/common/node_includes.h"
 #include "shell/common/options_switches.h"
+#include "shell/common/process_util.h"
 #include "shell/common/v8_value_serializer.h"
 #include "storage/browser/file_system/isolated_context.h"
 #include "third_party/blink/public/common/associated_interfaces/associated_interface_provider.h"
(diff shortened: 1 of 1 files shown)


nornagon (dismissed):
Hm, I don't think `LOG(ERROR)` is the right place for a load failure. Things can fail to load for a variety of reasons and it's not generally an application error when they do. Further, `LOG` is often not easy to see on Windows.

Can you expand a bit on what your goal is here?

loc:
Sure. I believe it's too easy to make a pretty reasonable mistake (misspelling or otherwise getting your file wrong for loadFile/loadURL) and not get an actionable error from it. The error that you do get comes in the renderer console, and it's a side-effect with a pretty obtuse and misdirecting error: `Not allowed to load local resource: file://misspelled-file.html/`.

There are two ways to get the correct error, `ERR_FILE_NOT_FOUND`, neither of which are immediately obvious, nor currently "recommended" by [electron-quick-start](https://github.com/electron/electron-quick-start/blob/master/main.js#L16):
1. You can `.catch()` the error from the loadFile promise or
2. You can subscribe to the webContents's `did-fail-load` event.

_A_ satisfactory solution could be to change our examples to include error handling, but I think that expecting log output when a webcontents fails to load is pretty reasonable. Ideally, actually, would be to surface `ERR_FILE_NOT_FOUND` in the renderer console too, but there doesn't seem to be that much prior art for doing that elsewhere.

nornagon:
Can we have this be a `console.warn` (in JS) instead of `LOG(ERROR)` in C++?

And can we make it dev-only?

zcbenz:
@loc Do you still plan to work on this PR?

loc:
I should! I wasn't immediately sure how to ferry the error to the renderer to console.warn in JS. Any thoughts on how I should approach? Naively I want to set up an IPC handler n window-setup.ts in the renderer, but not sure if there's a cleaner way?

MarshallOfSound:
@loc you can use the `EmitWarning` helper from `process_util.cc`

zcbenz (approved):

zcbenz (changes requested):
@loc The "app module getGPUInfo() API succeeds with basic GPUInfo" test keeps failing on Windows CI, which I believe it is caused by the warning message leaked to the output of `gpu-info.js`, can you fix it?

codebytere:
@zcbenz Andy no longer works at Slack and hasn't for some time - i can try to fix this up.

codebytere (approved):

zcbenz (approved):

PR #22834, opened 2020-03-25
docs: categorize breaking changes

nornagon:
#### Description of Change
This change hopes to make scanning through the breaking changes document easier by categorizing each breaking change, and including _what broke_ in the title.

So far, the categories are:
- API Changed
  - If you use the old API, you can reasonably expect that your code will throw an exception after this change.
- Behavior Changed
  - The behavior of Electron has changed in a way that might cause your app to malfunction, but not necessarily by throwing an exception and not necessarily at the point in your code that requires change.
- Default Changed
  - If you were implicitly relying on the old default, your code may break. The old behavior can be restored by making the option explicit.
- Deprecated
  - This API will continue to function as before in this release, but it is expected to be removed in future. Technically speaking, this isn't exactly a breaking change.
- Removed
  - This API or functionality was removed.

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/docs/breaking-changes.md
+++ b/docs/breaking-changes.md
@@ -2,6 +2,16 @@
 
 Breaking changes will be documented here, and deprecation warnings added to JS code where possible, at least [one major version](tutorial/electron-versioning.md#semver) before the change is made.
 
+### Types of Breaking Changes
+
+This document uses the following convention to categorize breaking changes:
+
+- **API Changed:** An API was changed in such a way that code that has not been updated is guaranteed to throw an exception.
+- **Behavior Changed:** The behavior of Electron has changed, but not in such a way that an exception will necessarily be thrown.
+- **Default Changed:** Code depending on the old default may break, not necessarily throwing an exception. The old behavior can be restored by explicitly specifying the value.
+- **Deprecated:** An API was marked as deprecated. The API will continue to function, but will emit a deprecation warning, and will be removed in a future release.
+- **Removed:** An API or feature was removed, and is no longer supported by Electron.
+
 ## Planned Breaking API Changes (10.0)
 
 ### Removed: Browser Window Affinity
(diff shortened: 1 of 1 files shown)


docs/breaking-changes.md
```diff
@@ -249,11 +245,11 @@ webFrame.setIsolatedWorldInfo(
   })
 ```
 
-### Removal of deprecated `marked` property on getBlinkMemoryInfo
+### Removed: `marked` property on `getBlinkMemoryInfo`
 
 This property was removed in Chromium 77, and as such is no longer available.
 
-### `webkitdirectory` attribute for `<input type="file"/>`
+### API Changed: `webkitdirectory` attribute for `<input type="file"/>` now lists directory contents
```
jkleinsc: The API didn't change here, but how it behaves did.
```suggestion
### Behavior Changed: `webkitdirectory` attribute for `<input type="file"/>` now lists directory contents
```

docs/breaking-changes.md
```diff
@@ -403,16 +400,16 @@ const w = new BrowserWindow({
...
-### `nativeWindowOpen`
+### Behavior Changed: `nodeIntegration` in child windows opened via `nativeWindowOpen`
 
-Child windows opened with the `nativeWindowOpen` option will always have Node.js integration disabled, unless `nodeIntegrationInSubFrames` is `true.
+Child windows opened with the `nativeWindowOpen` option will always have Node.js integration disabled, unless `nodeIntegrationInSubFrames` is `true`.
 
-### Privileged Schemes Registration
+### API Changed: Registering privileged schemes must now be done before app ready
 
 Renderer process APIs `webFrame.setRegisterURLSchemeAsPrivileged` and `webFrame.registerURLSchemeAsBypassingCSP` as well as browser process API `protocol.registerStandardSchemes` have been removed.
```
jkleinsc: This isn't needed for this PR, but its something I found when working on the Electronegativity upgrade checks.  There never was a function called `setRegisterURLSchemeAsPrivileged` - it should have been `registerURLSchemeAsPrivileged`
```suggestion
Renderer process APIs `webFrame.registerURLSchemeAsPrivileged` and `webFrame.registerURLSchemeAsBypassingCSP` as well as browser process API `protocol.registerStandardSchemes` have been removed.
```

Issue #22833, opened 2020-03-25
Opening and closing a modal window prevents system modals from opening


kenpowers-signal:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 8.1.0
* **Operating System:**
  * macOS Catalina 10.15.3
* **Last Known Working Electron version:**
  * 8.0.3

### Expected Behavior

After opening and closing a modal window in an Electron application we should be able to open system modals (e.g., browsing for files or saving a file)

### Actual Behavior

After opening and closing a modal window in an Electron application we are unable to open system modals (e.g., browsing for files or saving a file)

### To Reproduce

Use this electron-fiddle gist: https://gist.github.com/c38df808f0ebd23f65e0c5e287d084cd

1. Set Electron fiddle to use `8.0.3`
1. Start the application
1. Press the "choose file" button, notice that it works
1. Close the file browser
1. Press the "open modal" button
1. Press the "close window" button
1. Press the "choose file" button, you can browse for a file
1. Repeat with `8.1.0`, `8.1.1`, `8.2.0`, and , `9.0.0-beta.10`. You can browse for a file before opening the modal, but not after closing the modal

### Additional Information

Potentially related: #22481

PR #22858, opened 2020-03-27
fix: dialog fails to show after modal close

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/22833.

The solution to https://github.com/electron/electron/issues/17558 unfortunately created a new issue, whereby native dialogs sometimes failed to open after modal close.

This PR fixes this by shifting the modal handling case logic to our delegate as [Chromium does](https://cs.chromium.org/chromium/src/components/remote_cocoa/app_shim/views_nswindow_delegate.mm?type=cs&q=retainblock&sq=package:chromium&g=0&l=137-148).

Tested with:
- https://gist.github.com/c38df808f0ebd23f65e0c5e287d084cd
- https://gist.github.com/8ff50362409e14fceb3e28bddbbdbfc4

cc @zcbenz @nornagon @kenpowers-signal

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where native macOS dialogs sometimes failed to show after modal close.

### DIFF
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -597,28 +597,18 @@ void NativeWindowMac::SetContentView(views::View* view) {
 }
 
 void NativeWindowMac::Close() {
-  // When this is a sheet showing, performClose won't work.
-  if (is_modal() && parent() && IsVisible()) {
-    NSWindow* window = parent()->GetNativeWindow().GetNativeNSWindow();
-    if (NSWindow* sheetParent = [window sheetParent]) {
-      base::ThreadTaskRunnerHandle::Get()->PostTask(
-          FROM_HERE, base::BindOnce(base::RetainBlock(^{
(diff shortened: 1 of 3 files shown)


shell/browser/native_window_mac.mm
```diff
@@ -597,28 +597,16 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
...
-  }
-
   if (!IsClosable()) {
     WindowList::WindowCloseCancelled(this);
     return;
   }
 
   [window_ performClose:nil];
+
+  if (is_modal() && parent() && IsVisible()) {
```
zcbenz: Should add a comment on why this is needed, here and in `electron_ns_window_delegate.mm`.

zcbenz:
I don't see any macOS error noise though, probably related to macOS version or SDK version.

PR #22861, opened 2020-03-27
chore: remove unused classes of views API

zcbenz:
#### Description of Change

This PR removes unused views classes, future new views APIs would be made public directly.

The `ImageView` class is moved to public export as it has been approved by @electron/wg-api.

There some pending work to be done and I will do them in new PRs:
1. Write documentations to make `ImageView` and related classes public.
2. Try to make `api::View` class use `gin::Wrappable` instead of `gin_helper::Wrappable`.

#### Release Notes

Notes: no-notes


deepak1556:
Can we keep the buildflag around and flip the default value to `true` ? There is no strict guideline about what goes behind buildflag, but I think if there is a chance to make the code base modular with a component then buildflag should be used.

Maybe once the views component is completely developed and the modularity doesn't offer significant purpose then we can remove it.

Also can we keep the directory structure around `shell/browser/api/views` ? Its easier to navigate and if there are platform specific implementations, can go behind respective sub-directories `shell/browser/api/views/linux` etc.

zcbenz:
> Can we keep the buildflag around and flip the default value to `true` ?
> Also can we keep the directory structure around `shell/browser/api/views` ?

I think it is actually very good idea, I have updated the PR.

PR #22867, opened 2020-03-27
fix: ensure standard schemes are registered in nw service process

deepak1556:
#### Description of Change

Crash in Network service process due to https://github.com/electron/electron/pull/20546 regressed after https://github.com/electron/electron/pull/22064/commits/f5b9e49be99f2b85c0381d791b381aa400db9751 .

This PR fixes by making a clear indication of how the schemes are registered in respective Browser, Renderer and Network process.

Identified while testing https://github.com/electron/electron/pull/20625 but ci logs in master also show the crash.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: fix crash in network service process when using protocol.registerSchemeAsPrivileged api

### DIFF
--- a/spec-main/api-protocol-spec.ts
+++ b/spec-main/api-protocol-spec.ts
@@ -2,6 +2,7 @@ import { expect } from 'chai';
 import { protocol, webContents, WebContents, session, BrowserWindow, ipcMain } from 'electron';
 import { promisify } from 'util';
 import { AddressInfo } from 'net';
+import * as ChildProcess from 'child_process';
 import * as path from 'path';
 import * as http from 'http';
 import * as fs from 'fs';
@@ -634,6 +635,22 @@ describe('protocol module', () => {
 
     it('returns error when scheme is not intercepted', async () => {
       await expect(uninterceptProtocol('http')).to.be.eventually.rejectedWith(Error);
+    });
+  });
+
+  describe('protocol.registerSchemeAsPrivileged', () => {
+    it('does not crash on exit', async () => {
+      const appPath = path.join(__dirname, 'fixtures', 'api', 'custom-protocol-shutdown.js');
+      const appProcess = ChildProcess.spawn(process.execPath, ['--enable-logging', appPath]);
+      let output = '';
+      appProcess.stdout.on('data', data => { output += data; });
+      appProcess.stderr.on('data', data => { output += data; });
+      const [code] = await emittedOnce(appProcess, 'exit');
+      if (code !== 0) {
+        console.log(code, output);
+      }
+      expect(code).to.equal(0);
+      expect(output).to.not.contain('VALIDATION_ERROR_DESERIALIZATION_FAILED');
     });
   });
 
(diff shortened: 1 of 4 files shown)


deepak1556:
@nornagon thoughts on how to test this ? when this crash happens it doesn't affect the resource loading on those custom schemes, so thats why existing test suites where passing. Only the service process goes on a loop restarting.

nornagon:
> @nornagon thoughts on how to test this ? when this crash happens it doesn't seem to affect the resource loading on those custom schemes, so thats why existing test suites where passing. Only the service process goes on a loop restarting.

Maybe I'm confused then... how is this change visible to the user then?

deepak1556:
Just infinite logs from network service

```
[14085:0331/150914.566974:WARNING:pref_notifier_impl.cc(61)] Init observer found at shutdown.
[14085:0331/150914.568450:ERROR:validation_errors.cc(76)] Invalid message: VALIDATION_ERROR_DESERIALIZATION_FAILED (network.mojom.NetworkContext.1731934142 )
```

And just observed that shutdown of app isn't graceful. Let me see if I can write a test that verifies app shutdown when using this api.

deepak1556:
@nornagon added simple test, PTAL.

spec-main/api-protocol-spec.ts
```diff
@@ -637,6 +638,22 @@ describe('protocol module', () => {
     });
   });
 
+  describe('protocol.registerSchemeAsPrivileged', () => {
+    it('does not crash on exit', async () => {
+      const appPath = path.join(__dirname, 'fixtures', 'api', 'custom-protocol-shutdown.js');
+      const appProcess = ChildProcess.spawn(process.execPath, ['--enable-logging', appPath]);
+      let output = '';
+      appProcess.stdout.on('data', data => { output += data; });
+      appProcess.stderr.on('data', data => { output += data; });
```
nornagon, line 647:
```suggestion
      let stdout = '';
      let stderr = '';
      appProcess.stdout.on('data', data => { stdout += data; });
      appProcess.stderr.on('data', data => { stderr += data; });
```

spec-main/api-protocol-spec.ts
```diff
@@ -637,6 +638,22 @@ describe('protocol module', () => {
...
+      const [code] = await emittedOnce(appProcess, 'exit');
+      if (code !== 0) {
+        console.log(code, output);
+      }
+      expect(code).to.equal(0);
+      expect(output).to.not.contain('VALIDATION_ERROR_DESERIALIZATION_FAILED');
```
nornagon, line 653:
```suggestion
      expect(stdout).to.not.contain('VALIDATION_ERROR_DESERIALIZATION_FAILED');
      expect(stderr).to.not.contain('VALIDATION_ERROR_DESERIALIZATION_FAILED');
```

stdout and stderr aren't necessarily synced, so this prevents potential mixing of output

deepak1556:

Issue #12640, opened 2018-04-17
Do not use base::Passed


alexeykuzmin:
It's going to be removed from Chromium, see https://crbug.com/812523.

/cc @electron/code-owners

deepak1556:
This has to be done alongside https://github.com/electron/electron/issues/12639

PR #22871, opened 2020-03-28
chore: remove discouraged base::Passed

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/12640.

Remove discouraged `base::Passed` from Bind calls.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


deepak1556:
For the other usage in `App::ImportCertificate` , use `base::Value` instead of `base::DictionaryValue` which is a movable type.

shell/browser/api/electron_api_app.cc
```diff
@@ -1071,16 +1071,15 @@ Browser::LoginItemSettings App::GetLoginItemSettings(
 }
 
 #if defined(USE_NSS_CERTS)
-void App::ImportCertificate(const base::DictionaryValue& options,
+void App::ImportCertificate(const base::Value& options,
```
nornagon: Can we make this [passed by value](https://source.chromium.org/chromium/chromium/src/+/master:base/values.h;l=72;drc=b01268ba46dd26506ea62c7ccfa63e00060bcd29) instead of by reference? I'm not sure it makes sense to `std::move` a const reference.

shell/browser/api/electron_api_app.cc
```diff
@@ -1090,12 +1089,11 @@ void App::ImportCertificate(const base::DictionaryValue& options,
...
 
 void App::OnCertificateManagerModelCreated(
-    std::unique_ptr<base::DictionaryValue> options,
+    base::Value options,
     net::CompletionOnceCallback callback,
     std::unique_ptr<CertificateManagerModel> model) {
   certificate_manager_model_ = std::move(model);
-  int rv =
-      ImportIntoCertStore(certificate_manager_model_.get(), *(options.get()));
+  int rv = ImportIntoCertStore(certificate_manager_model_.get(), &options);
```
nornagon: kinda weird to pass `&options` here, can we `std::move` it instead since this function doesn't use it after this call?

shell/browser/api/electron_api_app.cc
```diff
@@ -1071,31 +1070,34 @@ Browser::LoginItemSettings App::GetLoginItemSettings(
...
                             net::CompletionRepeatingCallback callback) {
+  DCHECK(options.is_dict());
+
   auto browser_context = ElectronBrowserContext::From("", false);
   if (!certificate_manager_model_) {
-    auto copy = base::DictionaryValue::From(
-        base::Value::ToUniquePtrValue(options.Clone()));
     CertificateManagerModel::Create(
         browser_context.get(),
         base::BindRepeating(&App::OnCertificateManagerModelCreated,
```
deepak1556: The `CreationCallback` used by `CertificateManagerModel::Create` is currently a repeating callback for no reason, the callback is passed to other threads and which always bind as `OnceCallback` , it would be better to upstream a CL that changes `CreationCallback` to `OnceCallback`.

```
typedef base::Callback<void(std::unique_ptr<CertificateManagerModel>)>
      CreationCallback;

// static
void CertificateManagerModel::Create(
    content::BrowserContext* browser_context,
    CertificateManagerModel::Observer* observer,
    const CreationCallback& callback);
```

to 

```
using CreationCallback = base::OnceCallback<void(std::unique_ptr<CertificateManagerModel>)>;

// static
void CertificateManagerModel::Create(
    content::BrowserContext* browser_context,
    CertificateManagerModel::Observer* observer,
    CreationCallback callback);
```

codebytere: @deepak1556 would you prefer i add a patch and implement this change in this PR or a follow-up?

deepak1556: We maintain a copy of the `CertificateManagerModel` class https://github.com/electron/electron/blob/master/chromium_src/chrome/browser/certificate_manager_model.cc, you can make the changes in this PR.

I would also recommend uploading a similar patch to upstream `CertificateManagerModel` code.

chromium_src/chrome/browser/certificate_manager_model.cc
```diff
@@ -69,7 +69,7 @@ net::NSSCertDatabase* GetNSSCertDatabaseForResourceContext(
 
 // static
 void CertificateManagerModel::Create(content::BrowserContext* browser_context,
-                                     const CreationCallback& callback) {
+                                     CreationCallback callback) {
```
deepak1556: `callback`  should be `std::move`'d wherever it is passed, and also the invocation should be `std::move(callback).Run`

codebytere: whoops, late-night mishap.

also, what does this mean for `GetCertDBOnIOThread`? callback is used twice there it appears

deepak1556: Okay a couple of things to get this correct,

1) Remove the copy of `CertificateManagerModel` from `chromium_src` and apply our changes as patch to upstream code, similar to how we do it for printing etc.

2) Make the OnceCallback change to CreationCallback in this patch.

3) Upstream part of that patch.

shell/browser/api/electron_api_app.cc
```diff
@@ -1071,31 +1070,33 @@ Browser::LoginItemSettings App::GetLoginItemSettings(
 }
 
 #if defined(USE_NSS_CERTS)
-void App::ImportCertificate(const base::DictionaryValue& options,
+void App::ImportCertificate(base::Value options,
                             net::CompletionRepeatingCallback callback) {
+  DCHECK(options.is_dict());
```
nornagon, line 1075: Is `DCHECK` the appropriate thing here? This method is exposed to JS, so anything could be passed as `options`. gin will happily pass an array here, for example, and DCHECKs won't be hit in release builds.

Perhaps
```suggestion
  if (!options.is_dict()) {
    thrower.ThrowTypeError("Expected options to be an object");
  }
```
? Where `thrower` is a `gin_helper::ErrorThrower`

shell/browser/api/electron_api_app.cc
```diff
@@ -1071,31 +1070,33 @@ Browser::LoginItemSettings App::GetLoginItemSettings(
...
+      ImportIntoCertStore(certificate_manager_model_.get(), std::move(options));
   std::move(callback).Run(rv);
 }
 
 void App::OnCertificateManagerModelCreated(
-    std::unique_ptr<base::DictionaryValue> options,
+    base::Value options,
     net::CompletionOnceCallback callback,
     std::unique_ptr<CertificateManagerModel> model) {
+  DCHECK(options.is_dict());
```
nornagon, line 1095: same here

chromium_src/BUILD.gn
```diff
@@ -74,8 +74,20 @@ static_library("chrome") {
...
     sources += [ "//chrome/browser/icon_loader_auralinux.cc" ]
     sources += [
+      "//chrome/browser/certificate_manager_model.cc",
+      "//chrome/browser/certificate_manager_model.h",
       "//chrome/browser/extensions/global_shortcut_listener_x11.cc",
       "//chrome/browser/extensions/global_shortcut_listener_x11.h",
+      "//chrome/browser/net/nss_context.h",
+      "//chrome/browser/net/nss_context_linux.cc",
+      "//chrome/browser/ui/browser_dialogs.cc",
+      "//chrome/browser/ui/browser_dialogs.h",
```
nornagon, line 84: it's surprising to me that these are needed, what requires them?

chromium_src/BUILD.gn
```diff
@@ -74,8 +74,20 @@ static_library("chrome") {
...
+      "//chrome/browser/ui/crypto_module_password_dialog_nss.cc",
+      "//chrome/browser/ui/crypto_module_password_dialog_nss.h",
+      "//chrome/browser/ui/views/chrome_layout_provider.cc",
+      "//chrome/browser/ui/views/chrome_layout_provider.h",
+      "//chrome/browser/ui/views/crypto_module_password_dialog_view.cc",
+      "//chrome/browser/ui/views/crypto_module_password_dialog_view.h",
```
nornagon, line 90: these too; we shouldn't be building chrome password dialog UI since we don't use it.

shell/browser/api/electron_api_app.cc
```diff
@@ -497,14 +497,16 @@ void OnClientCertificateSelected(
...
-  base::string16 password;
+int ImportIntoCertStore(CertificateManagerModel* model, base::Value options) {
+  std::string file_data;
   net::ScopedCERTCertificateList imported_certs;
   int rv = -1;
-  options.GetString("certificate", &cert_path);
-  options.GetString("password", &password);
+
+  std::string* cert_path_ptr = options.FindStringPath("certificate");
+  std::string cert_path = *cert_path_ptr;
```
nornagon, line 506: if `certificate` isn't present in the dictionary, `cert_path_ptr` will be `nullptr` and this will SIGSEGV. We should handle that more gracefully.

shell/browser/api/electron_api_app.cc
```diff
@@ -497,14 +497,16 @@ void OnClientCertificateSelected(
...
+
+  std::string* pwd = options.FindStringPath("password");
+  base::string16 password = base::UTF8ToUTF16(*pwd);
```
nornagon, line 509: here too

chromium_src/BUILD.gn
codebytere: @nornagon it's an unfortunate little undefined symbol chain - `certificate_manager_model.cc` requires `GetNSSCertDatabaseForResourceContext` (defined in `nss_context_linux.cc`) as well as `chrome::UnlockSlotsIfNecessary` (defined in `crypto_module_password_dialog_nss.cc`) so i aded those, which in turn introduced a few more undefined symbols, requiring the last several files be added.

chromium_src/BUILD.gn
codebytere: see above

shell/browser/api/electron_api_app.cc
```diff
@@ -497,14 +497,19 @@ void OnClientCertificateSelected(
...
-                        const base::DictionaryValue& options) {
+int ImportIntoCertStore(CertificateManagerModel* model, base::Value options) {
   std::string file_data, cert_path;
   base::string16 password;
   net::ScopedCERTCertificateList imported_certs;
   int rv = -1;
-  options.GetString("certificate", &cert_path);
-  options.GetString("password", &password);
+
+  std::string* cert_path_ptr = options.FindStringPath("certificate");
```
ckerr, line 506: FindStringPath() creates a `PathSplitter` to walk through nested key-based Values, but that extra work is unnecessary because we know that the path is a single key `"certificate"` -- we should use `FindStringKey()` instead.

```suggestion
  std::string* cert_path_ptr = options.FindStringKey("certificate");
```

shell/browser/api/electron_api_app.cc
```diff
@@ -497,14 +497,19 @@ void OnClientCertificateSelected(
...
+  if (cert_path_ptr)
+    cert_path = *cert_path_ptr;
+
+  std::string* pwd = options.FindStringPath("password");
```
ckerr, line 510:
```suggestion
  std::string* pwd = options.FindStringKey("password");
```

chromium_src/BUILD.gn
deepak1556: I will look at reducing the dependency chain  before merging this PR, if you don't mind @codebytere . We should be able to omit these by patching some code in `certificate_manager_model`

codebytere: Sounds good - i was avoiding extra patch surface but if we think that's a better alternative i'll swap to that :)

codebytere:
@nornagon CL opened at https://chromium-review.googlesource.com/c/chromium/src/+/2138564.

shell/browser/api/electron_api_app.cc
```diff
@@ -1071,31 +1076,35 @@ Browser::LoginItemSettings App::GetLoginItemSettings(
 }
 
 #if defined(USE_NSS_CERTS)
-void App::ImportCertificate(const base::DictionaryValue& options,
-                            net::CompletionRepeatingCallback callback) {
+void App::ImportCertificate(gin_helper::ErrorThrower thrower,
+                            base::Value options,
+                            net::CompletionOnceCallback callback) {
+  if (!options.is_dict()) {
+    thrower.ThrowTypeError("Expected options to be an object");
```
nornagon, line 1083:
```suggestion
    thrower.ThrowTypeError("Expected options to be an object");
    return;
```

PR #22878, opened 2020-03-28
fix: Do not ignore electron exit signals in cli.js wrapper

andersk:
#### Description of Change

When the electron child process exits with a signal, the close event handler receives code `null` and the cli wrapper would silently exit successfully.  Fix it to log a message and exit with a nonzero code in this case.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/npm/cli.js
+++ b/npm/cli.js
@@ -5,7 +5,11 @@ var electron = require('./')
 var proc = require('child_process')
 
 var child = proc.spawn(electron, process.argv.slice(2), { stdio: 'inherit', windowsHide: false })
-child.on('close', function (code) {
+child.on('close', function (code, signal) {
+  if (code === null) {
+    console.error(electron, 'exited with signal', signal)
+    process.exit(1)
+  }
   process.exit(code)
 })
 


npm/cli.js
```diff
@@ -5,7 +5,11 @@ var electron = require('./')
 var proc = require('child_process')
 
 var child = proc.spawn(electron, process.argv.slice(2), { stdio: 'inherit', windowsHide: false })
-child.on('close', function (code) {
+child.on('close', function (code, signal) {
+  if (code === null) {
+    console.error(electron, 'exited with signal', signal)
+    process.exit(1)
```
MarshallOfSound: We can pass the signal through to this process and actually self-kill with a signal

```suggestion
    process.kill(0, signal)
```

andersk: I thought about that strategy. Note that we’d need to turn off the `SIGINT` and `SIGTERM` handlers first, and some signals aren’t fatal so we’d want an `exit` fallback anyway.

My concern was that it might be weird for the wrapper script to write a core dump just because the binary did, and also that the added complexity might make it harder to get review. But if you want to go with that, I can make the changes.

zcbenz: /cc @MarshallOfSound ☝️

MarshallOfSound: > My concern was that it might be weird for the wrapper script to write a core dump just because the binary did,

Hm this makes sense, maybe exiting with code 1 is ok

PR #22887, opened 2020-03-30
ci: Use smaller vms with goma

jkleinsc:
#### Description of Change

Now that we are building with goma we can scale back our CI VMs for testing.
This PR changes our Linux builds to use 8-core machines and our Windows builds to use 16-core machines for testing.  Using these sized machines results in builds that complete in about the same time as our current configuration.  I tested using an 8-core machine on Windows but that was too slow due to those jobs running gclient sync/unzipping chromium source.
#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,6 +67,10 @@ machine-linux-medium: &machine-linux-medium
   <<: *docker-image
   resource_class: medium
 
+machine-linux-xlarge: &machine-linux-xlarge
+  <<: *docker-image
+  resource_class: xlarge
+
 machine-linux-2xlarge: &machine-linux-2xlarge
   <<: *docker-image
   resource_class: 2xlarge+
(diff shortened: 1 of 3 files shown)


script/release/ci-release-build.js
```diff
@@ -206,7 +206,8 @@ function buildAppVeyor (targetBranch, options) {
 async function callAppVeyor (targetBranch, job, options) {
   console.log(`Triggering AppVeyor to run build job: ${job} on branch: ${targetBranch} with release flag.`);
   const environmentVariables = {
-    ELECTRON_RELEASE: 1
+    ELECTRON_RELEASE: 1,
+    APPVEYOR_BUILD_WORKER_CLOUD: 'libcc-20'
```
jkleinsc: This is needed so that release builds use the larger 32-core machines.

.circleci/config.yml
```diff
@@ -1621,9 +1625,9 @@ jobs:
 
   # Layer 2: Builds.
   linux-x64-testing:
-    <<: *machine-linux-2xlarge
+    <<: *machine-linux-xlarge
     environment:
-      <<: *env-linux-2xlarge
+      <<: *env-global
```
jkleinsc: testing builds do not need the number of cores in `env-linux-2xlarge` because those builds always use goma.

PR #22893, opened 2020-03-30
docs: document what experimental means explicitly

MarshallOfSound:
As discussed at API / releases this gives us a starting point for explicit "experimental" API documentation and untags some experimental things that have been around for years and definitely aren't experimental anymore.

Notes: no-notes

### DIFF
--- a/docs/experimental.md
+++ b/docs/experimental.md
@@ -19,3 +19,5 @@ happen at an API WG meeting.  Things to consider when discussing / nominating:
 * The above "two major stables release lines" condition must have been met
 * During that time no major bugs / issues should have been caused by the adoption of this feature
 * The API is stable enough and hasn't been heavily impacted by Chromium upgrades
+* Is anyone using the API?
+* Is the API fulfilling the original proposed usecases, does it have any gaps?
--- a/docs/experimental.md
+++ b/docs/experimental.md
@@ -1,0 +1,21 @@
+# Experimental APIs
+
+Some of Electrons APIs are tagged with `_Experimental_` in the documentation,
+this tag indicates that the API may not be considered stable and the API may
+be removed or modified more frequently than other APIs with less warning.
+
+## Conditions for an API to be tagged as Experimental
+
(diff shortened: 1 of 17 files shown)


docs/experimental.md
```diff
@@ -0,0 +1,21 @@
+# Experimental APIs
+
+Some of Electrons APIs are tagged with `_Experimental_` in the documentation,
+this tag indicates that the API may not be considered stable and the API may
```
nornagon:
```suggestion
Some of Electrons APIs are tagged with `_Experimental_` in the documentation.
This tag indicates that the API may not be considered stable and the API may
```

docs/experimental.md
```diff
@@ -0,0 +1,21 @@
...
+## Conditions for an API to be tagged as Experimental
+
+Anyone can request an API be tagged as experimental in a feature PR, disagreements
+on the experimental nature of a feature can be discussed in the API WG if they
+can't be resolved in the PR.
+
+## Process for removing the Experimental tag
+
+Once an API has been stable and in at least two major stable release lines it
+can be nominated to have it's experimental tag removed.  This discussion should
```
nornagon:
```suggestion
can be nominated to have its experimental tag removed.  This discussion should
```

docs/experimental.md
```diff
@@ -0,0 +1,21 @@
...
+happen at an API WG meeting.  Things to consider when discussing / nominating:
+
+* The above "two major stables release lines" condition must have been met
+* During that time no major bugs / issues should have been caused by the adoption of this feature
+* The API is stable enough and hasn't been heavily impacted by Chromium upgrades
```
nornagon: We should probably also review use cases and examine whether they're well served by the API.

* Is anyone using the API?
* Is it working for them? Any pain points?

The API will get significantly harder to change after we remove the experimental tag, so we should make sure we're happy with the current state of the API when we do so!

MarshallOfSound:
> should we link to the experimental doc anywhere?

We can link it on the website automatically for _Experimental_ tags

PR #22896, opened 2020-03-30
fix: propagate preferred color scheme to the renderer

### LINKED ISSUE #22846: prefers-color-scheme media query not updated when changing nativeTheme.themeSource
### Preflight Checklist

### Issue Details

* **Electron Version:**
  *  v8.2.0
* **Operating System:**
  *  Windows 10 (17763) 
  *  n/a

### Expected Behavior

When setting `nativeTheme.themeSource`, the `prefers-color-scheme` media query updates accordingly and all elements using the query on the page are restyled.

### Actual Behavior

Setting `nativeTheme.themeSource` does not appear to update the media query until a system-level change to the light/dark theme is made.

e.g. if we do `nativeTheme.themeSource = "dark"`, then running `matchMedia("(prefers-color-scheme: dark)").matches` returns `false`.

The documentation appears to suggest the media query should update: https://www.electronjs.org/docs/api/native-theme#nativethemethemesource

### To Reproduce

Fiddle Gist: https://gist.github.com/6e571fb4e8aa3d319a3109d61dda79c3
1. Set your windows theme to light in system settings.
2. Open the fiddle. The menu bar and page are in the light theme.
3. From the `Theme` menu, click `Dark`. This sets `nativeTheme`.
4. Notice that the menu changes colour but the page does not (as per screenshot).
5. Set your windows theme to dark in system settings.
6. Notice that the page is now dark.
7. Set your windows theme to light in system settings.
8. The page remains dark as expected.
9. From the `Theme` menu, click `Light`.
10. The menu changes colour but the page does not. It only updates when changing the system settings.

### Screenshots

image

### Additional Information

- Using the remote module to set `nativeTheme`.

MarshallOfSound:
Fixes #22846

This regressed in a recent Chromium upgrade:

https://github.com/electron/electron/commit/032552df574008047b4672b49e70556524601c4f#diff-8a19e7f384ba27f625eb0e55641ee5feL414-L418

This PR fixes the issue and adds a test to ensure it does not regress by accident in the future.  It also fixes a crash that was revealed by the test I wrote.  If you close a window as the native theme updates, only possible if you do this with `nativeTheme` directly but fixed it 👍 

It also updates our `native_theme` implementation to override both `web` and `native_ui` themes that Chromium uses.  Currently Chromium doesn't use webUI and we opt to use native for everything, but there's no harm overriding and it might help us in the future with edge cases.

Notes: Fixed issue where `prefers-color-scheme` would not be updated / set correctly when your OS was in dark mode

### DIFF
--- a/shell/browser/electron_browser_client.cc
+++ b/shell/browser/electron_browser_client.cc
@@ -524,6 +524,11 @@ void ElectronBrowserClient::OverrideWebkitPrefs(
   prefs->picture_in_picture_enabled = false;
 #endif
 
+  ui::NativeTheme* native_theme = ui::NativeTheme::GetInstanceForNativeUi();
+  prefs->preferred_color_scheme = native_theme->ShouldUseDarkColors()
+                                      ? blink::PreferredColorScheme::kDark
+                                      : blink::PreferredColorScheme::kLight;
+
   SetFontDefaults(prefs);
 
   // Custom preferences of guest page.
--- a/shell/browser/api/electron_api_native_theme.cc
+++ b/shell/browser/api/electron_api_native_theme.cc
@@ -23,14 +23,16 @@ namespace electron {
 
 namespace api {
 
-NativeTheme::NativeTheme(v8::Isolate* isolate, ui::NativeTheme* theme)
-    : theme_(theme) {
-  theme_->AddObserver(this);
+NativeTheme::NativeTheme(v8::Isolate* isolate,
+                         ui::NativeTheme* ui_theme,
+                         ui::NativeTheme* web_theme)
+    : ui_theme_(ui_theme), web_theme_(web_theme) {
+  ui_theme_->AddObserver(this);
   Init(isolate);
 }
 
 NativeTheme::~NativeTheme() {
-  theme_->RemoveObserver(this);
+  ui_theme_->RemoveObserver(this);
 }
 
 void NativeTheme::OnNativeThemeUpdatedOnUI() {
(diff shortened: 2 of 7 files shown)


shell/browser/electron_browser_client.cc
```diff
@@ -524,6 +524,11 @@ void ElectronBrowserClient::OverrideWebkitPrefs(
   prefs->picture_in_picture_enabled = false;
 #endif
 
+  ui::NativeTheme* native_theme = ui::NativeTheme::GetInstanceForNativeUi();
+  prefs->preferred_color_scheme = native_theme->ShouldUseDarkColors()
```
deepak1556: Can you use the new pipeline to set the color scheme for renderer process as mentioned in https://chromium-review.googlesource.com/c/chromium/src/+/1845575

shell/browser/electron_browser_client.cc
```diff
@@ -524,6 +524,11 @@ void ElectronBrowserClient::OverrideWebkitPrefs(
   prefs->picture_in_picture_enabled = false;
 #endif
 
+  ui::NativeTheme* native_theme = ui::NativeTheme::GetInstanceForNativeUi();
```
deepak1556: should be `GetInstanceForWeb` since we use the same for deciding `ShouldUseDarkColors` in native theme api.

deepak1556:
Why is this PR under `fast-track` ?

shell/browser/electron_browser_client.cc
MarshallOfSound: That was a mistake, updated the `ShouldUseDarkColors` back to be native UI

MarshallOfSound:
@deepak1556 Was quite a bad regression in 8 and it's basically just restoring old code.  Can we land this restoration of old logic first and I'll follow up with using the new pipeline thing?

deepak1556:
@MarshallOfSound the content layer no longer reads the value from web preferences https://chromium-review.googlesource.com/c/chromium/src/+/1845575/18/content/renderer/render_view_impl.cc , shouldn't we use the api from native theme directly based on the CL ?

MarshallOfSound:
@deepak1556 It must or that CL must have been reverted because if I match the changes in that CL it doesn't work and the test fails.  With these changes it does work and the test passes.  Not entirely sure what's going on but I know this works and it's what we had before.

Issue #22370, opened 2020-02-25
webRequest API stopped working with file:// urls since 7.0.0 (and up to 9.0.0-beta.3 at least)


ChALkeR:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.0.0
  * 7.1.13
  * 8.0.1
  * 8.0.2
  * 9.0.0-beta.2
  * 9.0.0-beta.3
* **Operating System:**
  * `Linux xps 5.5.4-arch1-1 #1 SMP PREEMPT Sat, 15 Feb 2020 00:36:29 +0000 x86_64 GNU/Linux`, Arch Linux current
* **Last Known Working Electron version:**
  * 6.1.8

### Expected Behavior
E.g. how 6.x behaves: `onBeforeRequest` callback gets fired before requests are made.

Note: `touch /tmp/1.html` to get `onCompleted`, otherwise check `onErrorOccured`.

```
Electron 6.1.8
onBeforeRequest: file:///tmp/1.html
onCompleted: file:///tmp/1.html
Load finished!
```

See code to reproduce below.

#### With `file:///tmp/1.html` (does not work):

```
Electron 7.0.0
Load finished!
```

```
Electron 7.1.13
Load finished!
```

```
Electron 8.0.2
Load finished!
```

```
Electron 9.0.0-beta.3
Load finished!
```

#### With `https://example.com/` (works):

```
Electron 9.0.0-beta.3
onBeforeRequest: https://example.com/
onCompleted: https://example.com/
Load finished!
```

```
Electron 8.0.2
onBeforeRequest: https://example.com/
onCompleted: https://example.com/
Load finished!
```

```
Electron 7.0.0
onBeforeRequest: https://example.com/
onCompleted: https://example.com/
Load finished!
```

### To Reproduce

```js
const { BrowserWindow, app } = require('electron')

app.on('ready', () => {
  const window = new BrowserWindow({})
  console.log(`Electron ${process.versions.electron}`)
  window.webContents.session.webRequest.onBeforeRequest((details, callback) => {
    console.log(`onBeforeRequest: ${details.url}`)
[...]

PR #22903, opened 2020-03-31
fix: webRequest module should work with file:// protocol

zcbenz:
#### Description of Change

Fix https://github.com/electron/electron/issues/22370.

When supporting asar archives in `file://` URL, overwrite the default `FileURLLoaderFactory` directly instead of intercepting the protocol in `ProxyingURLLoaderFactory`. Otherwise `webRequest` won't be able to proxy the `file://` requests.

This is actually the correct way to override `file://` protocol, previous implementation was essentially a hack.

#### Checklist

- [x] `npm test` passes

#### Release Notes

### DIFF
--- a/shell/browser/protocol_registry.cc
+++ b/shell/browser/protocol_registry.cc
@@ -3,11 +3,45 @@
 // found in the LICENSE file.
 
 #include <memory>
+#include <utility>
 
 #include "shell/browser/electron_browser_context.h"
+#include "shell/browser/net/asar/asar_url_loader.h"
 #include "shell/browser/protocol_registry.h"
 
 namespace electron {
+
+namespace {
+
+// Provide support for accessing asar archives in file:// protocol.
+class AsarURLLoaderFactory : public network::mojom::URLLoaderFactory {
+ public:
+  AsarURLLoaderFactory() {}
+
+ private:
+  // network::mojom::URLLoaderFactory:
+  void CreateLoaderAndStart(
+      mojo::PendingReceiver<network::mojom::URLLoader> loader,
+      int32_t routing_id,
+      int32_t request_id,
+      uint32_t options,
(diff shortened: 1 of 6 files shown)


shell/browser/protocol_registry.cc
```diff
@@ -21,6 +55,12 @@ ProtocolRegistry::~ProtocolRegistry() = default;
 
 void ProtocolRegistry::RegisterURLLoaderFactories(
     content::ContentBrowserClient::NonNetworkURLLoaderFactoryMap* factories) {
+  // Override the default FileURLLoaderFactory to support asar archives.
+  //
+  // Note that |insert| or |emplace| should not be used here, as they do nothing
+  // if the key already exists, while we would like to overwrite the factory.
+  (*factories)[url::kFileScheme] = std::make_unique<AsarURLLoaderFactory>();
```
deepak1556, line 62: When extension system is enabled, it also registers a factory for `file://` https://github.com/electron/electron/blob/master/shell/browser/electron_browser_client.cc#L1281 , it seems to override this one for sub resources. Will that be okay ?

deepak1556: Actually the problem is other way around, the extension system uses `emplace` to register, so it won't override our factory. But then question, its trying to register a file provider with stricter rules https://github.com/electron/electron/blob/master/shell/browser/electron_browser_client.cc#L1202 and that won't be available, is that okay ? Or can we accommodate that configuration here ?

zcbenz: The url factory installed by extension system is to support `file://` under `chrome-extension://` schemes, Chromium only installs file url factory for `file://` and `about:blank` so it is fine for extension system to use `emplace`.

The problem is we are unconditionally installing asar url factory for all protocols, I have updated the code to only install it when needed.

Issue #22744, opened 2020-03-18
Undo/Redo has no effect inside <webview>


adityamudgil2505:
Calling Cmd+Z (webContents.undo()) and Shift+Cmd+Z (webContents.redo()) have no effect inside a <webview> tag

### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * Electron v8.1.1
* **Operating System:**
  * macOS 10.15

### Expected Behavior
Calling Cmd+Z (webContents.undo()) and Shift+Cmd+Z (webContents.redo()) inside a <webview> tag will do the functionality that are defined for them i.e, undo and redo. Also by calling them from menu bar inside edit.

### Actual Behavior
Calling undo and redo inside <webview> has no effect. It gives no response. I remains unchanged after calling undo/redo it.

### To Reproduce
I use Electron fiddler to reproduce this issue and the gist link is https://gist.github.com/958fea4350204faf6778121b97af90f7
(no effect of undo and redo inside google search bar)

### Screenshots
![ezgif com-video-to-gif-2](https://user-images.githubusercontent.com/25535045/76966288-bea26980-694b-11ea-84ed-215097205276.gif)

andersk:
This Gist has working undo/redo for me on both Windows and Linux (via both menu entry and keyboard shortcut), so this must be a macOS specific bug.

codebytere:
@andersk some preliminary digging tells me the functionality isn't totally broken: if you run [this gist](https://gist.github.com/a28818a2355047854908e7f6c87529d2) you'll see that undo and redo work correctly on button click. I thus suspect it's a menu or shortcut issue; i'll look a bit more.

pyrocto:
This makes editing comments very annoying in the mac Zulip desktop app, so those of us who use it are very interested in getting this fixed.

PR #22911, opened 2020-03-31
fix: undo and redo broken in webviews

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/22744.

When propagating the undo and redo events from the menu, Chromium was not
properly sending the events to the focused webcontents. This fixes that by ensuring that the events are sent to the proper webcontents.

cc @zcbenz @MarshallOfSound @nornagon 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed an issue where undo and redo shortcuts did not work in webviews.

Co-authored-by: Samuel Attard <sattard@slack-corp.com>


deepak1556:
Is this only an issue for macOS ?

codebytere:
@deepak1556 yes i believe so, see [this comment](https://github.com/electron/electron/issues/22744#issuecomment-601492527).

deepak1556:
Thanks, can you mention the platform affected in the release notes. The notes reflects webview in general is affected, so can assume all platforms.

deepak1556:
Just tested the following sample on web with OOPIF and in electron with webview. The problem is not seen with OOPIF, I don't think the patch completely identifies the issue , we seem to workaround it.

```html
// main.html
<iframe src="http://localhost:8000/test.html"></iframe>
```

```html
//embed.html
<input type="text"/>
```

codebytere:
@deepak1556 would you prefer we not merge this solution and try to find another one? Or shall we merge this and see if other issues come up later and we can regroup then?

deepak1556:
I would suggest to identify the reason why it works for plain OOPIF and not for our webview implemented on top of OOPIF. I am not comfortable merging the solution in its current state, a more closer understanding of the problem would help.

MarshallOfSound:
@deepak1556 I believe this patch is correct, OOPIF are not separate webContents.  WebViews are implemented as separate webContents that are attached via "inner" webContents.

The documentation for `GetFocusedWebContents` says this is explicitly what we want when taking into account guest views and inner web contents.

deepak1556:
@MarshallOfSound thanks, that explains it.

I was able to repro the issue with chrome's webview https://developer.chrome.com/apps/tags/webview which is based on `//compoent/guest_view` that is similar design to our webview. Can you please open an upstream issue and link it in the patch. Thanks!

patches/chromium/fix_undo_redo_broken_in_webviews.patch
```diff
@@ -5,7 +5,8 @@ Subject: fix: undo redo broken in webviews
 
 When propagating the undo and redo events from the menu, Chromium was not
 properly sending the events to the focused webcontents on macOS. This fixes
-that error, and will be upstreamed into Chromium.
+that error. A crbug has been opened at https://bugs.chromium.org/p/chromium/issues/detail?id=1067284
```
deepak1556, line 8: Thank you! 🙇

akashnimare:
@sofianguy is this being added in the v9?

ckerr:
It's been awhile since the question was asked so @akashnimare you've probably already found the answer, but I'll add this comment in case anyone else comes across this PR in the future:

https://github.com/electron/electron/pull/23370, which did land in 9-x-y, superceded this PR

PR #22916, opened 2020-04-01
refactor: ginify Menu

nornagon:
#### Description of Change
Refactors `Menu` to use `gin::Wrappable`, and also refactors some of the internal JS away from an overcomplicated 'delegate' model.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


shell/common/gin_helper/event_emitter_caller.h
```diff
@@ -69,11 +69,11 @@ v8::Local<v8::Value> CallMethod(v8::Isolate* isolate,
                                 gin::Wrappable<T>* object,
                                 const char* method_name,
                                 Args&&... args) {
-  v8::HandleScope scope(isolate);
+  v8::EscapableHandleScope scope(isolate);
```
deepak1556, line 72: Nice Catch!

shell/browser/api/electron_api_menu_mac.mm
```diff
@@ -177,8 +177,11 @@
 }
 
 // static
-gin_helper::WrappableBase* Menu::New(gin::Arguments* args) {
-  return new MenuMac(args);
+gin::Handle<Menu> Menu::New(gin::Arguments* args) {
+  auto handle =
+      gin::CreateHandle(args->isolate(), static_cast<Menu*>(new MenuMac(args)));
+  gin_helper::CallMethod(args->isolate(), handle.get(), "_init");
```
deepak1556, line 183: `_init` was a notion from `native_mate` https://github.com/electron/native-mate/blob/master/native_mate/wrappable.cc#L41 , if this is being maintained in `gin` world as well, then can this be made a part of `gin_helper::constructible` ?

nornagon: I'd prefer to avoid it completely, though in this case removing it would have meant a more major refactor than I wanted to embark on. For now I'd like to put it in on a case-by-case basis and see how we go. It's a 1-liner to call it so it's not too much trouble.

PR #22919, opened 2020-04-01
fix: webRequest module should work with file:// protocol (9-x-y)

zcbenz:
Backport of https://github.com/electron/electron/pull/22903.


deepak1556:

PR #22920, opened 2020-04-01
fix: webRequest module should work with file:// protocol (8-x-y)

zcbenz:
Backport of https://github.com/electron/electron/pull/22903.


zcbenz:
There seems to be some problems with asar subresource requests, I'll try to fix it later.

ckerr:
@zcbenz does this still need to be landed in 8-x-y?

zcbenz:
Don't have time for this and it is not a critical fix that has to be backported, closing for now.

PR #22922, opened 2020-04-01
test: do not trigger unhandled promise rejections

zcbenz:
Fix the tests of https://github.com/electron/electron/pull/22903 so they do not throw unhandled promise rejections.

I have included this commit in the backports of https://github.com/electron/electron/pull/22903.

Notes: none


jkleinsc:

PR #22924, opened 2020-04-01
refactor: ginify DownloadItem

nornagon:
#### Description of Change
Refactors `DownloadItem` to use `gin::Wrappable`, and also fix a bug which could crashes due to duplicate download IDs. I'm not sure what I did that triggered the crash, but the usage seemed wrong so I fixed it.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/electron_api_download_item.cc
+++ b/shell/browser/api/electron_api_download_item.cc
@@ -99,17 +99,17 @@ DownloadItem::~DownloadItem() {
   }
 }
 
-bool DownloadItem::CheckDestroyed() const {
+bool DownloadItem::CheckAlive() const {
   if (!download_item_) {
     gin_helper::ErrorThrower(v8::Isolate::GetCurrent())
-        .ThrowError("Object has been destroyed");
+        .ThrowError("DownloadItem used after being destroyed");
     return false;
   }
   return true;
 }
 
 void DownloadItem::OnDownloadUpdated(download::DownloadItem* item) {
-  if (!CheckDestroyed())
+  if (!CheckAlive())
     return;
   if (download_item_->IsDone()) {
     Emit("done", item->GetState());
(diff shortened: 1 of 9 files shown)


shell/browser/api/electron_api_download_item.cc
```diff
@@ -71,65 +97,89 @@ DownloadItem::~DownloadItem() {
...
     download_item_->Remove();
   }
+}
 
-  // Remove from the global map.
-  g_download_item_objects.erase(weak_map_id());
+bool DownloadItem::CheckDestroyed() const {
+  if (!download_item_) {
+    gin_helper::ErrorThrower(v8::Isolate::GetCurrent())
+        .ThrowError("Object has been destroyed");
```
ckerr, line 105: Minor suggestion to make life a little easier for people hitting this exception someday:

```suggestion
        .ThrowError("DownloadItem used after being destroyed");
```

shell/browser/api/electron_api_download_item.cc
```diff
@@ -71,65 +97,89 @@ DownloadItem::~DownloadItem() {
     download_item_->RemoveObserver(this);
     download_item_->Remove();
   }
+}
 
-  // Remove from the global map.
-  g_download_item_objects.erase(weak_map_id());
+bool DownloadItem::CheckDestroyed() const {
```
ckerr, line 102: _(nit)_

Slightly weird negative logic here where a method named "CheckDestroyed" returning false if the item is destroyed.

Not a blocker, but consider renaming to `CheckAlive` but keeping the logic the same? That way the usage would read like:

```cc
if (!CheckAlive())
  return;
if (download_item_->IsDone())) {
  ...
```

shell/browser/api/electron_api_download_item.cc
```diff
@@ -71,65 +97,89 @@ DownloadItem::~DownloadItem() {
...
 
 void DownloadItem::OnDownloadUpdated(download::DownloadItem* item) {
+  if (!CheckDestroyed())
+    return;
   if (download_item_->IsDone()) {
     Emit("done", item->GetState());
-    // Destroy the item once item is downloaded.
-    base::ThreadTaskRunnerHandle::Get()->PostTask(FROM_HERE,
-                                                  GetDestroyClosure());
+    Unpin();
```
ckerr, line 116: Is this guaranteed to always be reached s.t. the `Pin();` call is always balanced out?

nornagon: Hm good question! The previous logic was that the V8 object would be kept alive through the `g_download_item_objects` map until the destructor of the C++ object was called, which was only ever triggered by this `GetDestroyClosure` call here. But it might have been leaky before. I'll look into it. I'm fairly confident at least that this doesn't introduce a _new_ memory leak.

shell/browser/api/electron_api_download_item.cc
nornagon: This is matching a pattern I started in `electron_api_tray.h`, so I'll update both.

shell/browser/api/electron_api_download_item.cc
nornagon: In `download_item.h`:

```
  // Returns true if the download is in a terminal state. This includes
  // completed downloads, cancelled downloads, and interrupted downloads that
  // can't be resumed.
  virtual bool IsDone() const = 0;
```

So I think between this and the `Unpin()` I added after this comment (in `OnDownloadDestroyed`) we should be good.

I could imagine an argument that we should Unpin when a download is paused and Pin again when it's resumed (since if you pause a download and then lose the reference, you're never going to resume it), but I think this is OK as it is.

nornagon:
CI failures unrelated. Merging.

Issue #22910, opened 2020-03-31
Crash in ExecuteJavaScript with preload, nodeIntegrationInSubFrames and invoke


boutetnico:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 8.2.0
  * 9.0.0-beta.12
* **Operating System:**
  * MacOS 10.14.6
* **Last Known Working Electron version:**
  * Unknown

### Expected Behavior
The renderer do not crash.

### Actual Behavior
The renderer is crashing as soon as the app is starting.

### To Reproduce

```sh
$ git clone https://github.com/boutetnico/electron-bug-report.git -b fpcollect
$ npm install
$ npm start || electron .
```

### Additional Information

I have spent a lot of time researching this and came to the conclusion the following conditions are required for the crash to occur:

- `nodeIntegrationInSubFrames`=`true`
- An invokable IPC is sent from preload to the main process
- Followed by invokation of `executeJavaScript`
- index.html loads [fp-collect](https://github.com/antoinevastel/fp-collect/blob/master/src/fpCollect.js)

Crash report from Sentry:
```
OS Version: macOS 10.14.6 (18G3020)
Report Version: 104

Crashed Thread: 33484704

Application Specific Information:
Fatal Error: EXC_BAD_ACCESS / KERN_INVALID_ADDRESS

Thread 33484704 Crashed:
0   Electron Framework              0x104a799de         electron::api::ExecuteJavaScript (electron_api_web_frame.cc:385)
1   Electron Framework              0x104a7e68b         mate::internal::Dispatcher<T>::DispatchToCallback (callback.h:132)
2   Electron Framework              0x1056028fb         v8::internal::FunctionCallbackArguments::Call (api-arguments-inl.h:158)
3   Electron Framework              0x105601df9         v8::internal::(anonymous namespace)::HandleApiCallHelper<T> (builtins-api.cc:111)
4   Electron Framework              0x1056013d1         v8::internal::Builtin_Impl_HandleApiCall (builtins-api.cc:141)
[...]

PR #22925, opened 2020-04-02
fix: webframe crashes for removed render frame

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/22910.

The above issue surfaced an error transitively through a node module which was calling:
```
const iframe = document.createElement('iframe');
iframe.srcdoc = 'blank page';
document.body.appendChild(iframe);

const result = typeof iframe.contentWindow.chrome;
iframe.remove();
```

This led to the realization we had no nullptr safety checks in `web_frame`, and so calling methods in `web_frame` right after a frame had been removed would cause renderer crashes. This fixes those errors by bailing or throwing errors if the render frames have for some reason been removed.

Tested with https://gist.github.com/bfeb118b0c5a0b9858538db4623a3cad.

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed occasional WebFrame crashes caused by removed iframes.

### DIFF
--- a/shell/renderer/api/electron_api_web_frame.cc
+++ b/shell/renderer/api/electron_api_web_frame.cc
@@ -18,6 +18,7 @@
 #include "shell/common/gin_converters/blink_converter.h"
 #include "shell/common/gin_converters/callback_converter.h"
 #include "shell/common/gin_helper/dictionary.h"
+#include "shell/common/gin_helper/error_thrower.h"
 #include "shell/common/gin_helper/promise.h"
 #include "shell/common/node_includes.h"
 #include "shell/renderer/api/electron_api_spell_check_client.h"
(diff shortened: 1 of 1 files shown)


shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -258,17 +259,32 @@ void SetName(v8::Local<v8::Value> window, const std::string& name) {
...
 }
 
-void SetZoomLevel(v8::Local<v8::Value> window, double level) {
+void SetZoomLevel(gin_helper::ErrorThrower thrower,
+                  v8::Local<v8::Value> window,
+                  double level) {
   content::RenderFrame* render_frame = GetRenderFrame(window);
+  if (!render_frame) {
+    thrower.ThrowError(
+        "Render frame was torn down before script could be executed");
```
MarshallOfSound, line 268: Can we change `script` in all these places to the method name.  E.g.

```suggestion
        "Render frame was torn down before script could be webFrame.setZoomLevel");
```

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -258,17 +259,33 @@ void SetName(v8::Local<v8::Value> window, const std::string& name) {
...
 }
 
-void SetZoomLevel(v8::Local<v8::Value> window, double level) {
+void SetZoomLevel(gin_helper::ErrorThrower thrower,
+                  v8::Local<v8::Value> window,
+                  double level) {
   content::RenderFrame* render_frame = GetRenderFrame(window);
+  if (!render_frame) {
+    thrower.ThrowError(
+        "Render frame was torn down before SetZoomLevel could be executed");
```
MarshallOfSound, line 268:
```suggestion
        "Render frame was torn down before webFrame.setZoomLevel could be executed");
```

PR #22937, opened 2020-04-02
feat: expose electron/{process} typed modules

MarshallOfSound:
## What's New?

This PR exposes three new modules.

```typescript
import * as electronMain from 'electron/main';
import * as electronCommon from 'electron/common';
import * as electronRenderer from 'electron/renderer';
```

The existing `electron` module is left untouched.

These three modules use our typescript generator to correctly type their interface based on what is available in each process.  E.g.

```typescript
// This works according to typescript
// But we in Electron land know you can't have both
import { BrowserWindow, desktopCapturer } from 'electron';

// The new recommendation for typescript users will be to do this
import { BrowserWindow } from 'electron/main';
import { desktopCapturer } from 'electron/renderer';

// Then users can use either linting rules or typescript workspaces to
// ensure that they only load from "electron/main" or "electron/renderer"
// in their main process or renderer process code.
```

This PR has zero impact on existing apps and is left undocumented while I work out any kinks / adjust the methodology after feedback from some folks.  I want to land this so that we can start experimenting with how useful drawing these typescript boundaries can be.

Notes for when we document it:

> Exposed new `electron/main`, `electron/common` and `electron/renderer` helper modules that limit the APIs available in the typescript definitions to just those available in the given process.  These can be used in combination with linting rules or typescript workspaces to safely limit which Electron process's APIs your typescript code can use

Notes: no-notes


MarshallOfSound:
> How does this choose what's in main vs. renderer?

Each docs file has a "Main Process" or "Renderer Process" or both at the top of the class / module definition.  That is read and interpreted by the docs parser and the TS defs then generate these interfaces based on that

> Also, what's in "common"?

Common === Things available in both renderer and main

MarshallOfSound:
@nornagon You can see the diff in the TS defs in the check run here:

https://github.com/electron/electron/pull/22937/checks?check_run_id=556675350

nornagon:
Hm, lots of the things in "Common" seem to be main-process-only, e.g. `shell` & many of the types.

MarshallOfSound:
`shell` is documented as being in both processes.

https://github.com/electron/electron/blob/master/docs/api/shell.md

All the interfaces are exposed in all three modules as these are types not values and they aren't "owned" by a certain process.

nornagon:
> `shell` is documented as being in both processes.
> 
> https://github.com/electron/electron/blob/master/docs/api/shell.md

Oh huh, TIL! It won't work with in a sandboxed renderer, though, which should probably be noted in that doc. I'll make a PR.

> All the interfaces are exposed in all three modules as these are types not values and they aren't "owned" by a certain process.

Hm... it seems weird to me to expose types that are only ever relevant to the main process from a module called 'renderer'. It's also weird that you could import a type from `common` in one file and import the same type from `renderer` in another file.

MarshallOfSound:
> Hm... it seems weird to me to expose types that are only ever relevant to the main process from a module called 'renderer'. It's also weird that you could import a type from common in one file and import the same type from renderer in another file.

The idea here is we block imports of "electron/renderer" for an apps main process files through a linting rule / TS workspaces.  However this is a valid usecase for the main process wanting a type the renderer process needs and vice-versa.  (E.g. sending an object over IPC for the other process to use).  The API itself isn't exposed, but the _type_ is exposed.

The TS compiler knows these are all aliases and it doesn't matter where you actually import the type.  In the same way you can use the namespace directly if you wanted.  Just different ways of accessing the same type.

nornagon:
I noticed this doesn't have any release notes, do we want to document this somewhere?

PR #22941, opened 2020-04-02
chore: bump chromium to 83.0.4102.3

nornagon:
[Chromium is skipping M82](https://chromereleases.googleblog.com/2020/03/chrome-and-chrome-os-release-updates.html).

Notes: none


patches/chromium/.patches
```diff
@@ -91,3 +91,4 @@ feat_add_support_for_overriding_the_base_spellchecker_download_url.patch
 gpu_notify_when_dxdiag_request_fails.patch
 feat_allow_embedders_to_add_observers_on_created_hunspell.patch
 fix_account_for_print_preview_disabled_when_printing_to_pdf.patch
+revert_enable_samesitebydefaultcookies_and.patch
```
nornagon, line 94: How come we have to cherry-pick this? is it not included in the roll?

deepak1556: Yeah it didn't come with this version in the PR, will come with the next roller bump. I just didn't update the version manually.

PR #22945, opened 2020-04-02
chore: cherry-pick b5950ad76471 from chromium

nornagon:
Protect automatic pull handlers with Mutex

In some cases, |rendering_automatic_pull_handlers_| in
DeferredTaskHandler can be touched from both the main thread and the
audio rendering thread. This CL adds a lock when it is updated,
processed, and cleared.

crash on the repro case after 30 min.

Test: Locally confirmed that the ASAN build with this patch does not
Bug: 1061018
Change-Id: I5f4440edcdc26e4a3afbfe8fad88492bdb49c323
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2101712
Cr-Commit-Position: refs/heads/master@{#750000}

Notes: Security: backport fix for CVE-2020-6451: Use after free in WebAudio

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -99,3 +99,4 @@ allow_restricted_clock_nanosleep_in_linux_sandbox.patch
 move_readablestream_requests_onto_the_stack_before_iteration.patch
 streams_convert_state_dchecks_to_checks.patch
 cherry-pick-fd211b44535c.patch
+cherry-pick-b5950ad76471.patch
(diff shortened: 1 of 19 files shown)


.circleci/config.yml
```diff
@@ -217,7 +217,34 @@ step-gclient-sync: &step-gclient-sync
...
+          else
+            echo
+            echo "======================================================================"
+            echo "Sync failed even with 3way patch application. You'll need to resolve"
+            echo "this conflict locally."
+            echo "======================================================================"
+            echo
+          fi
+          exit 1
+        fi
```
jkleinsc, line 247: I like this but I'm not sure it belongs in this backport PR?

nornagon: Yeah, I'm still experimenting with it. Once it works I'll move it to another PR. The patch in this PR still doesn't apply.

Issue #14161, opened 2018-08-17
EXC_BAD_ACCESS: atom::WindowList::CloseAllWindows() [window_list.cc : 88 + 0x3]


bpasero:
* Electron Version: 2.0.5
* Operating System (Platform and Version): macOS

Data from our crash reporter shows this as one of the top crashes:

```

Package: VSCode
Version: 1.26.0
OS: Mac OS X 10.13.6 (17G65)

Operating system: Mac OS X
                  10.13.6 17G65
CPU: amd64
     family 6 model 60 stepping 3
     4 CPUs

GPU: UNKNOWN

Crash reason:  EXC_BAD_ACCESS / EXC_I386_GPFLT
Crash address: 0x10ac4d8e6
Process uptime: 14298 seconds

Thread 0 (crashed)
 0  Electron Framework!atom::WindowList::CloseAllWindows() [window_list.cc : 88 + 0x3]
    rax = 0x6000000000000000   rdx = 0x00007ffee764fa88
    rcx = 0x0000000000000000   rbx = 0x00006040002004d8
    rsi = 0x0000000000003830   rdi = 0x00007fe126f490d0
    rbp = 0x00007ffee764fc20   rsp = 0x00007ffee764fbf0
     r8 = 0x00000000000003ff    r9 = 0x000060000039c630
    r10 = 0x00007fe127390a30   r11 = 0x00007fe127a6df10
    r12 = 0x0000000000000064   r13 = 0x000060000017bc00
    r14 = 0x00006040002004e0   r15 = 0x0000000000000000
    rip = 0x000000010ac4d8e6
    Found by: given as instruction pointer in context
 1  Electron Framework!<name omitted> [browser.cc : 52 + 0x5]
    rbp = 0x00007ffee764fc40   rsp = 0x00007ffee764fc30
    rip = 0x000000010abfa516
    Found by: previous frame's frame pointer
 2  Electron Framework!-[AtomApplication terminate:] [atom_application.mm : 38 + 0x8]
    rbp = 0x00007ffee764fc60   rsp = 0x00007ffee764fc50
    rip = 0x000000010ac04e55
    Found by: previous frame's frame pointer
 3  AppKit + 0x404008
    rbp = 0x00007ffee764fd00   rsp = 0x00007ffee764fc70
    rip = 0x00007fff553ff008
    Found by: previous frame's frame pointer
 4  AppKit + 0x403e6d
[...]

PR #22948, opened 2020-04-03
fix: nullptr check when closing windows

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/14161.

We weren't performing a nullptr check, so we would experience an `EXC_BAD_ACCESS` crash
if an object in the WindowVector is destroyed during cleanup. This fixes that by adding said nullptr check.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an occasional crash when closing all BrowserWindows.

### DIFF
--- a/shell/browser/window_list.cc
+++ b/shell/browser/window_list.cc
@@ -84,16 +84,19 @@ void WindowList::CloseAllWindows() {
 #if defined(OS_MACOSX)
   std::reverse(windows.begin(), windows.end());
 #endif
-  for (auto* const& window : windows)
-    if (!window->IsClosed())
+  for (auto* const& window : windows) {
+    if (window && !window->IsClosed())
       window->Close();
+  }
 }
 
 // static
 void WindowList::DestroyAllWindows() {
   WindowVector windows = GetInstance()->windows_;
-  for (auto* const& window : windows)
-    window->CloseImmediately();  // e.g. Destroy()
+  for (auto* const& window : windows) {
+    if (window)
+      window->CloseImmediately();
+  }
 }
 
 WindowList::WindowList() = default;
--- a/shell/browser/window_list.cc
+++ b/shell/browser/window_list.cc
@@ -10,6 +10,18 @@
 #include "shell/browser/native_window.h"
 #include "shell/browser/window_list_observer.h"
 
+namespace {
+template <typename T>
+std::vector<base::WeakPtr<T>> ConvertToWeakPtrVector(std::vector<T*> raw_ptrs) {
+  std::vector<base::WeakPtr<T>> converted_to_weak;
+  for (auto* raw_ptr : raw_ptrs) {
+    base::WeakPtr<T> weak = raw_ptr->GetWeakPtr();
+    converted_to_weak.push_back(weak);
+  }
+  return converted_to_weak;
+}
+}  // namespace
+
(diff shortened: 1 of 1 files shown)


deepak1556:
changing to `std::vector<base::WeakPtr<NativeWindow>>` will help rather than maintaining a raw map, the ideal solution is for the map to be maintained by `AddWindow` , `RemoveWindow` , but I don't know when is case this happens. Open to solutions.

MarshallOfSound:
If we make this a vector of weak pointers this solution makes sense, without it I don't think this fix actually helps.

deepak1556:
@zcbenz any reason not maintain the `WindowList` vector  as weakptr rather than converting at the call site ?

zcbenz:
I think the internal records should still be a vector of raw ptrs, the purpose of `WindowList` is to provide a precise list of existing windows and it is doing the work correctly. Changing the internal vector to store weak ptrs would be a weird design, because `WindowList` already correctly manages the window list so there is no way for its elements to be null. 

There is crash because `CloseAllWindows` and `DestroyAllWindows` are deleting windows while iterating them, so I think we should only fix these methods.

shell/browser/window_list.cc
```diff
@@ -84,16 +96,22 @@ void WindowList::CloseAllWindows() {
 #if defined(OS_MACOSX)
   std::reverse(windows.begin(), windows.end());
 #endif
-  for (auto* const& window : windows)
-    if (!window->IsClosed())
+  for (auto* const& window : windows) {
```
deepak1556, line 99: Weakptr vector here too

shell/browser/window_list.cc
```diff
@@ -81,19 +93,27 @@ void WindowList::RemoveObserver(WindowListObserver* observer) {
 // static
 void WindowList::CloseAllWindows() {
   WindowVector windows = GetInstance()->windows_;
+  std::vector<base::WeakPtr<NativeWindow>> weak_windows =
+      ConvertToWeakPtrVector(windows);
```
zcbenz, line 97: `WindowVector windows = GetInstance()->windows_` is doing a copy of vector, can just pass `ConvertToWeakPtrVector(GetInstance()->windows_)` here to save the copy.

shell/browser/window_list.cc
```diff
@@ -10,6 +10,18 @@
 #include "shell/browser/native_window.h"
 #include "shell/browser/window_list_observer.h"
 
+namespace {
+template <typename T>
+std::vector<base::WeakPtr<T>> ConvertToWeakPtrVector(std::vector<T*> raw_ptrs) {
+  std::vector<base::WeakPtr<T>> converted_to_weak;
```
zcbenz, line 16: Can do a `converted_to_weak.reserve(raw_ptrs.size())` here;

shell/browser/window_list.cc
```diff
@@ -10,6 +10,18 @@
...
+  for (auto* raw_ptr : raw_ptrs) {
+    base::WeakPtr<T> weak = raw_ptr->GetWeakPtr();
+    converted_to_weak.push_back(weak);
```
zcbenz, line 19: Can do `push_back(raw_ptr->GetWeakPtr()` so it is moved instead of copied.

Issue #21018, opened 2019-11-07
Stream protocols and fs.createReadStream sometimes never complete in Electron 7


itsananderson:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.0.0-beta.4 - 8.0.0-beta.2
* **Operating System:**
  * macOS 10.14.6, Windows 10
* **Last Known Working Electron version:**
  * 7.0.0-beta.3

### Expected Behavior

According to the docs, one should be able to register a stream protocol handler which serves files from the file system.

The following code snippet is suggested by the [docs](https://electronjs.org/docs/api/protocol#protocolregisterstreamprotocolscheme-handler-completion):

```
const { protocol } = require('electron')
const fs = require('fs')

protocol.registerStreamProtocol('atom', (request, callback) => {
  callback(fs.createReadStream('index.html'))
}, (error) => {
  if (error) console.error('Failed to register protocol')
})
```

The expected behavior is that Electron should read from the stream and (assuming no read errors etc.) serve a valid response for the custom protocol request.

### Actual Behavior

Starting with Electron 7.0.0-beta.4 it looks like the request sometimes randomly never completes on the Renderer side. This can be seen in the Network tab as well as by inspecting AJAX calls at runtime. The read stream seems to be created successfully, and the response starts to be sent back to the renderer, but the final state change never happens.

### To Reproduce

Run the following Fiddle to see the behavior change. In Electron < 7.0.0-beta4, the page can make an AJAX request to the custom scheme in a loop and print the contents. In Electron >= 7.0.0-beta4 the page can occasionally make a few AJAX requests, but it will eventually halt when one request never reaches a `XMLHttpRequest.DONE` state.

https://gist.github.com/33a9b357c606c7cbe7c2cac99bbda973

Good console output (7.0.0-beta3 and bellow) looks like:

```
index.html:22 ajaxStateChange readyState: 1 statusCode: 0
index.html:22 ajaxStateChange readyState: 2 statusCode: 200
index.html:22 ajaxStateChange readyState: 3 statusCode: 200
[...]

PR #22955, opened 2020-04-03
fix: video/audio playing on custom stream protocols

pfrazee:
#### Description of Change

Currently, `<video>` and `<audio>` tags fail to play for custom streaming protocols. `blink::WebMediaPlayerImpl` - which provides the `<video>` and `<audio>` behaviors - needs to know whether a data source will stream or fully buffer the response. It determines this behavior with `MultibufferDataSource::AssumeFullyBuffered()` which has http/s hardwired. An incorrect determination will cause the video/audio to fail playing.

This PR includes a chromium patch which adds a list of "streaming protocols" in order to allow other protocols to register their streaming behavior. `MultibufferDataSource::AssumeFullyBuffered()` then refers to the registry so that it can correctly determine the data source's settings.

Reference issue: https://github.com/electron/electron/issues/21018#issuecomment-581529309

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added {stream:} option to registerSchemeAsPrivileged to enable custom protocols to stream video and audio


patches/chromium/add_streaming-schemes_registry.patch
```diff
@@ -0,0 +1,133 @@
...
++  };
++
+   // Schemes that are allowed for referrers.
+   std::vector<SchemeWithType> referrer_schemes = {
+       {kHttpsScheme, SCHEME_WITH_HOST_PORT_AND_USER_INFORMATION},
+@@ -551,6 +557,10 @@ void AddStandardScheme(const char* new_scheme, SchemeType type) {
+                       &GetSchemeRegistryWithoutLocking()->standard_schemes);
+ }
+ 
++void AddStreamingScheme(const char* new_scheme) {
```
deepak1556, line 78: Argument `SchemeType` is not defined here, hence undefined symbols with overloaded functions.

pfrazee: Aha, I'll give that a fix.

shell/browser/api/electron_api_protocol.cc
```diff
@@ -173,6 +178,11 @@ ProtocolError Protocol::RegisterProtocol(ProtocolType type,
                                          const std::string& scheme,
                                          const ProtocolHandler& handler) {
   bool added = protocol_registry_->RegisterProtocol(type, scheme, handler);
+  if (added) {
+    if (type == ProtocolType::kHttp || type == ProtocolType::kStream) {
+      g_streaming_schemes.push_back(scheme);
```
deepak1556, line 183: Necessary review, this call assumes that `registerSchemeAsPrivileged` will be called after , but we document users to make the 'registerSchemeAsPrivileged` as the first call in the app as early as possible before app ready due to threading constraints in `//url`.

The right way would be to provide a new option `stream: true` in the `registerSchemeAsPrivileged` api and go from there.

pfrazee: My assumption was that we should automatically register known-streaming protocols, but I guess you're saying that's not an option due to when `registerProtocol()` is called?

deepak1556: > I guess you're saying that's not an option due to when registerProtocol() is called?

Yup

pfrazee:
@deepak1556 The PR has been fixed and updated with your suggestions. You can find a test fiddle here (be sure to update the video path in main.js): https://gist.github.com/283332286b015993700cabf1262ff6d4

Per @nornagon's suggestion, I'm going to submit the chromium patch to chromium to see if they'll upstream it. I'll keep this PR updated with progress on that front.

patches/chromium/feat_add_streaming-schemes_registry.patch
```diff
@@ -0,0 +1,135 @@
...
+Subject: feat: add streaming-schemes registry
+
+blink::WebMediaPlayerImpl - which provides the <video> and <audio> behaviors - needs to know
+whether a data source will stream or fully buffer the response. It determines this behavior
+with MultibufferDataSource::AssumeFullyBuffered() which has http/s hardwired. An incorrect
+determination will cause the video/audio to fail playing.
+
+This patch adds a list of "streaming protocols" to the url SchemeRegistry in order to allow
+other protocols to register their streaming behavior. MultibufferDataSource::AssumeFullyBuffered()
+then refers to the registry so that it can correctly determine the data source's settings.
```
deepak1556, line 13: Should also provide a way to register via `ContentClient::AddAdditionalSchemes` https://source.chromium.org/chromium/chromium/src/+/master:content/public/common/content_client.h;l=117-148 so that we can register the scheme in `ElectronContentClient::AddAdditionalSchemes` that will help the network service process understand the scheme.

pfrazee: Done

pfrazee:
Issue opened in Chromium tracker: https://bugs.chromium.org/p/chromium/issues/detail?id=1068034#c3

pfrazee:
Chromium CI created: https://chromium-review.googlesource.com/c/chromium/src/+/2161417/

ahallicks:
Just to note that the notes read:

>Fix to video/audio *playblack* on custom stream protocols

I think you mean *playback*.

pfrazee:
Ah, yes. Whoops, thanks for the correction

pfrazee:
Chromium just closed as `wontfix` https://bugs.chromium.org/p/chromium/issues/detail?id=1068034#c7

pfrazee:
@nornagon Okay! I've rewritten the PR so that the chromium patch adds the "streaming protocols" list to the media namespace (in the MultibufferDataSource code). Let me know if that's looking better to you.

pfrazee:
Okay looks like the builds are failing because of something related to my patch. I'll check into that.

I'll also add a test and improve the docs a bit.

pfrazee:
Test and more docs are added. Also rebased onto master. Let's see how the builds go...

pfrazee:
A note about the test: I found that very small videos will play correctly even without the fix in this PR. The included video is the smallest I could find which still triggers the behavior (63kb)

shell/renderer/renderer_client_base.cc
```diff
@@ -82,6 +82,10 @@
 #include "shell/renderer/extensions/electron_extensions_renderer_client.h"
 #endif  // BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
 
+namespace media {
+void AddStreamingScheme(const char*);
+}
```
nornagon, line 87: Shouldn't this come from `#include "media/blink/multibuffer_data_source.h"`?

patches/chromium/feat_add_streaming-protocol_registry_to_multibuffer_data_source.patch
```diff
@@ -0,0 +1,68 @@
...
++++ b/media/blink/multibuffer_data_source.cc
+@@ -65,10 +65,20 @@ const int kUpdateBufferSizeFrequency = 32;
+ // How long to we delay a seek after a read?
+ constexpr base::TimeDelta kSeekDelay = base::TimeDelta::FromMilliseconds(20);
+ 
++// Schemes that can stream their responses.
++std::vector<std::string> streaming_schemes = {
++    url::kHttpsScheme,
++    url::kHttpScheme
++};
```
nornagon, line 27: Use `base::NoDestructor` here (see `GetSchemeRegistryWithoutLocking` in `//url/url_util.cc` for an example.)

spec/fixtures/pages/video.html
```diff
@@ -0,0 +1,14 @@
+<html>
+<body>
+<video id="videoPlayer" src="/video.mp4" autoplay muted></video>
```
nornagon, line 3: It looks like the mp4 in this PR is 1MB? That's a little on the larger side to commit to the repo forever :/

pfrazee: Yeah any thoughts on what to do there? We need a larger video for the test to work. I guess we could try to come up with a script to generate the video on test-run?

shell/renderer/renderer_client_base.cc
pfrazee: I'm not totally clear on when we should do includes or forward declarations. In another PR I was advised to just do a forward decl. Happy to do either

patches/chromium/feat_add_streaming-protocol_registry_to_multibuffer_data_source.patch
pfrazee: Went ahead and copied their entire approach of putting the schemes in a struct and applying locking schemes

pfrazee:
Pushed an update as requested, questions remaining on 2 items

shell/renderer/renderer_client_base.cc
deepak1556: https://chromium.googlesource.com/chromium/src/+/master/styleguide/c++/c++.md#forward-declarations-vs_includes

In short, try to forward declare in header files where possible especially when dealing with headers that might be included often, check `content/public/browser/browser_context.h` as a good example. But you definitely don't forward declare in an implementation file `.cc` , this has no advantage.

nornagon: Yeah, good rule of thumb is:

- Never forward-declare in a .cc file
  - we use forward-declarations in .h files to avoid "propagating" a dependency that might not actually force a recompile. however, in the case that a .cc file depends on something in a header, we _do_ want to recompile that .cc file when the header changes. In this case, for instance, if the function's signature changed, we'd want to get a compile error.
- Never forward-declare a function (only `class`)
  - Mismatched function signatures between what's forward-declared and what's in the header can cause odd problems at link-time, and it can be easy to get them subtly wrong. Best avoid it altogether.

spec/fixtures/pages/video.html
nornagon: you mentioned [here](https://github.com/electron/electron/pull/22955#issuecomment-622066127) that you found a 63kb video that triggers the issue. That seems sufficiently small. If that's not the case, then yeah a script to generate a video seems reasonable.

Would it trigger the issue to have the custom protocol handler stream back a very small amount of the video in the first chunk? e.g. just a few KB. Then wait a bit and return the rest.

shell/renderer/renderer_client_base.cc
pfrazee: Ohh right okay, that makes sense. Sorry, I'm a C++ rookie. I'll fix this.

spec/fixtures/pages/video.html
pfrazee: @nornagon sorry that comment was unclear, the the smaller video doesn't trigger the issue.

I'll try your second suggestion first. Generating an mp4 would be a bit harder.

shell/renderer/renderer_client_base.cc
pfrazee: Fixed

spec/fixtures/pages/video.html
pfrazee: I was able to find a relatively small webm generator called Whammy and used that.

pfrazee:
Pushed an update (hopefully) resolving issues

docs/api/protocol.md
```diff
@@ -68,7 +68,7 @@ The `protocol` module has the following methods:
 module gets emitted and can be called only once.
 
 Registers the `scheme` as standard, secure, bypasses content security policy for resources,
-allows registering ServiceWorker and supports fetch API.
+allows registering ServiceWorker, supports fetch API, and streams video/audio.
```
nornagon, line 71:
```suggestion
allows registering ServiceWorker, supports fetch API, and streaming video/audio.
```

patches/chromium/feat_add_streaming-protocol_registry_to_multibuffer_data_source.patch
```diff
@@ -0,0 +1,97 @@
...
++struct SchemeRegistry {
++  // Schemes that can stream their responses.
++  std::vector<std::string> streaming_schemes = {
++    url::kHttpsScheme,
++    url::kHttpScheme
++  };
++};
++
++// Ensure that the schemes aren't modified after first use.
++static std::atomic<bool> g_scheme_registries_used{false};
```
nornagon, line 41: I think this lock is superfluous for this use case & can be removed.

patches/chromium/feat_add_streaming-protocol_registry_to_multibuffer_data_source.patch
```diff
@@ -0,0 +1,97 @@
...
++
++// Gets the scheme registry without locking the schemes. This should *only* be
++// used for adding schemes to the registry.
++SchemeRegistry* GetSchemeRegistryWithoutLocking() {
++  static base::NoDestructor<SchemeRegistry> registry;
++  return registry.get();
++}
```
nornagon, line 48: Since we only have one thing in our registry, there's no need for a struct. Instead:
```suggestion
+std::vector<std::string>* GetStreamingSchemes() {
+  static base::NoDestructor<std::vector<std::string>> streaming_schemes;
+  return streaming_schemes.get();
+}
```

patches/chromium/feat_add_streaming-protocol_registry_to_multibuffer_data_source.patch
pfrazee: Removed

docs/api/protocol.md
pfrazee: Done

patches/chromium/feat_add_streaming-protocol_registry_to_multibuffer_data_source.patch
pfrazee: Done, I had to pre-populate the list with http/s, lmk if my solution works for yall

patches/chromium/feat_add_streaming-protocol_registry_to_multibuffer_data_source.patch
```diff
@@ -0,0 +1,87 @@
...
++std::vector<std::string>* GetStreamingSchemes() {
++  static base::NoDestructor<std::vector<std::string>> streaming_schemes;
++
++  // initialize default schemes
++  std::vector<std::string>* schemes = streaming_schemes.get();
++  if (!g_streaming_schemes_initialized) {
++    schemes->push_back(url::kHttpsScheme);
++    schemes->push_back(url::kHttpScheme);
++    g_streaming_schemes_initialized = true;
++  }
```
nornagon, line 43: I think this can be done with an initializer, e.g.
```suggestion
+  static base::NoDestructor<std::vector<std::string>> streaming_schemes = {url::kHttpsScheme, url::kHttpScheme};
```

pfrazee: Aha, yes you can if you pass the initialization value into the constructor. Just pushed an update.

pfrazee:
You bet, thanks to yall for helping me through this despite my weak C++ chops!

pfrazee:
Hey I hope yall don't mind the bump, just wanted to see if we can get this merged!

nornagon:
@pfrazee looks like there are merge conflicts, would you mind resolving them?

pfrazee:
Sure thing

pfrazee:
Okay still waiting on the tests but I think the conflicts are all set

pfrazee:
Ugh why are the builds failing now. Testing locally.

pfrazee:
Appveyor is failing with

```
ERROR at //electron/shell/renderer/renderer_client_base.cc:23:11: Can't include this header from here.
#include "media/blink/multibuffer_data_source.h"
          ^------------------------------------
The target:
  //electron:electron_lib
  //media/blink:blink
```

But my build does not. Trying to diagnose...

pfrazee:
Okay I can't sort out why online builds are failing over this but not in my local env. I hate to punt but do yall have any pointers on why this might be happening?

deepak1556:
@pfrazee for linux and mac we have that failing step you see on appveyor as a separate pipeline which also fails [linux-x64-testing-gn-check](https://circleci.com/gh/electron/electron/573535)

The check thats failing is basically https://gn.googlesource.com/gn/+/master/docs/reference.md#cmd_check which is to validate that included headers match the build dependency graph and makes sure that appropriate targets are being built. Its a pre-build sanity check.

If you look at the full output, it reads

```
ERROR at //electron/shell/renderer/renderer_client_base.cc:23:11: Can't include this header from here.
#include "media/blink/multibuffer_data_source.h"
          ^------------------------------------
The target:
  //electron:electron_lib
  //media/blink:blink

to include headers.

  //electron:electron_lib -->
  //content/public/app:app --[private]-->
  //content/public/renderer:renderer_sources --[private]-->
  //content/renderer:renderer --[private]-->
  //media/blink:blink
```

`//media/blink:blink` is built via a transitive private dependency, and you are trying to depend on a header from that target directly, its always possible that one of the transitive targets can remove this dependency at some point, at which your compilation is gonna fail. In this case, the fix is add `//media/blink:blink` to deps of `//electron:electron_lib`. Ofcourse these errors will always be caught during link step, thats why I said its a pre-build sanity check which is quite useful on a target with nearly 30000 sources.

pfrazee:
@deepak1556 I guess you could say I was in the deps of despair. Appreciate the explanation. I'll look a little more closely next time -- if I had noticed which command that was failing, I might have been able to trace the meaning of the error.

Just pushed again, I'll ping if we're all-clear.

pfrazee:
Appears one (presumably unrelated) test in windows failed due to a timeout. Otherwise all clear.

pfrazee:
🎉 Thanks again for all the help with this PR

obsius:
I recently ran into this issue on Friday, and what lucky timing with this PR.  Now that this code is merged, will it be in the next bugfix release of 9, or will this be scheduled for a later release?

nornagon:
This PR has not been targeted to be backported to any previous version, so with no further action it will be available in v11.

This is a new feature, so backporting it to an old version would mean bumping the minor version and require approval from the Releases WG. See https://github.com/electron/governance/tree/master/wg-releases#feature-backport-requests for more details on the process for requesting feature backports.

pfrazee:
Sorry to be a pest but this didn't make it into today's 10 beta.3, what's the plan? Is it just in the backlog?

pfrazee:
Oh man I'm so sorry, I must have had a cached page or something because I didn't see @nornagon's reponse. I hate being a pest on merges to begin with so now I really feel bad 😅 See you in v11!

PR #22958, opened 2020-04-03
build: strip uuid from darwin binaries

thypon:
#### Description of Change

with these configuration we are removing one other
source of nondeterminism. In particular this change will
strip LC_UUID from any darwin binary.

This is needed since "random" UUID are added in the MACH header.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


deepak1556:
Just want to point that breakpad symbol generation use the `LC_UUID` for generating the final identifier for the symbols, but it does fallback to other option when `LC_UUID` is not available. https://github.com/google/breakpad/blob/master/src/common/mac/file_id.cc#L79

I would suggest this patch to be upstreamed first or get opinions from upstream authors, so we don't break something non-obvious.

patches/chromium/build_strip_uuid_from_darwin_binaries.patch
```diff
@@ -0,0 +1,24 @@
...
+From: Andrea Brancaleoni <abc@pompel.me>
+Date: Fri, 3 Apr 2020 22:08:22 +0200
+Subject: [PATCH] build: strip uuid from darwin binaries
+
+---
+ build/config/mac/BUILD.gn | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/build/config/mac/BUILD.gn b/build/config/mac/BUILD.gn
+index de8233bba4a1..2806cb8e0cb2 100644
```
nornagon, line 11: Please use `git-export-patches` to export patch files in the standard format.

patches/chromium/build_strip_uuid_from_darwin_binaries.patch
```diff
@@ -0,0 +1,20 @@
...
+diff --git a/build/config/mac/BUILD.gn b/build/config/mac/BUILD.gn
+index de8233bba4a1dec56f2b429d99642dae2b9eb7f4..2806cb8e0cb270e93cde7a0867ae142c0d9c2ec2 100644
+--- a/build/config/mac/BUILD.gn
++++ b/build/config/mac/BUILD.gn
+@@ -95,6 +95,6 @@ config("mac_dynamic_flags") {
+ # preserve specific symbols.
+ config("strip_all") {
+   if (enable_stripping) {
+-    ldflags = [ "-Wcrl,strip,-x,-S" ]
++    ldflags = [ "-Wcrl,strip,-x,-S,-no_uuid" ]
```
nornagon, line 18: NB. you can do this without a patch by removing the `strip_all` config from the relevant targets in Electron and adding a new config that has your desired ldflags added. Grep for `configs -=` in GN files in Chromium for examples.

andrea-doyensec: I was thinking of using that approach too, then I thought `no_uuid` can be a good addition also to chromium codebase.

andrea-doyensec:
https://bugs.chromium.org/p/chromium/issues/detail?id=1068970 I added a new bug in chromium issue tracker to discuss even there the implication of removing `LC_UUID`

thypon:
I created a new bug in chromium issue tracker to discuss further there the implication of removing `LC_UUID`.

https://bugs.chromium.org/p/chromium/issues/detail?id=1068970

nornagon:
https://bugs.chromium.org/p/chromium/issues/detail?id=1068970#c4

> Removing UUIDs will absolutely break crash reporting. The UUID is used on the crash server to match debug information and the original code module up with the crash dump.

> [...] I’m heavily inclined against -no_uuid anywhere.

also,

> LC_UUID is supposed to be deterministic.

thypon:
I am planning to investigate how LC_UUID is different in the two machines tomorrow.
Seeing what chromium team is saying about debugging breaking without that UUID, it's not a good idea to remove it during stripping.

thypon:
So I did a test maintaining the same output directory as the CI `/Users/distiller/project/src/out`. Seems like in this particular case the `LC_UUID` is the same. My guess is that since debug symbols are including full output paths and the `LC_UUID` links them to the executable, this `LC_UUID` is output path dependant.

Since Windows builds are output path dependant as well, but we can infer this path from files inside the repo and the CI output, I'm closing this pull request and adding a limitation in the reproducible `README.md`.

Screen Shot 2020-04-10 at 11 39 13

thypon:
This has been tested for `electron-v10.0.0-nightly.20200408-darwin-x64.zip`.

PR #22968, opened 2020-04-06
build: strip local/dbg symbols, from Linux binaries

thypon:
#### Description of Change

Non full stripping seems to be the cause of limited
non-determinism. Employ similar default as macOS when
stripping away symbols.

--discard-all       Remove all local symbols (saving only global symbols).
--strip-debug     Remove the debugging symbol table entries (those created by the -g option to cc(1) and other compilers).

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/strip-binaries.py
+++ b/script/strip-binaries.py
@@ -22,7 +22,7 @@ def strip_binary(binary_path, target_cpu):
     strip = 'mips64el-redhat-linux-strip'
   else:
     strip = 'strip'
-  execute([strip, '--preserve-dates', binary_path])
+  execute([strip, '--discard-all', '--strip-debug', '--preserve-dates', binary_path])
 
 def main():
   args = parse_args()


script/strip-binaries.py
```diff
@@ -22,7 +22,7 @@ def strip_binary(binary_path, target_cpu):
     strip = 'mips64el-redhat-linux-strip'
   else:
     strip = 'strip'
-  execute([strip, '--preserve-dates', binary_path])
+  execute([strip, '-x', '-S', '--preserve-dates', binary_path])
```
nornagon, line 25: Prefer long-style switches in scripts, as they self-document to some extent.

Also, from the man-page, `-x` means "remove all non-global symbols", and `-S` means "remove debugging symbols only". How can these not conflict?

```suggestion
  execute([strip, '--discard-all', '--strip-debug', '--preserve-dates', binary_path])
```

andrea-doyensec: These are the two flags included by default in chromium macOS. Isn't possible to have global debugging symbols?

nornagon: Hm, I'm not sure. If Chromium combines them that's a reasonable hint that it might be useful to do so.

andrea-doyensec: Just to be noted that `--discard-all` and `--strip-debug` does not seem to be POSIX compatible flags. Most probably they are only available on GNU strip. Do we still want to use them verbatim?

nornagon: This script is Linux-only.

jkleinsc:

PR #22969, opened 2020-04-06
feat: include verify.* scripts

thypon:
#### Description of Change

Reproducible builds are a set of software
development practices that create an
independently-verifiable path from source to binary code.

These scripts allow to verify both local reprodicibility
of:
- Windows x64 (See verify-reproducible.bat)
- Linux and macOS x64 (See verify-reproducible.sh)

Don't merge these scripts before determinism of the build system
has been fully tested on macOS and Linux. ( https://github.com/electron/electron/pull/22968 , https://github.com/electron/electron/pull/22958 )

Cc: @nornagon @jkleinsc @ikkisoft 

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/reproducible/README.md
+++ b/script/reproducible/README.md
@@ -1,0 +1,36 @@
+Reproducible builds are a set of software development practices that
+create an independently-verifiable path from source to binary code.
+See [https://reproducible-builds.org/](https://reproducible-builds.org/)
+for more information.
+
+Electron release determinism may be verified by calling one of the two verify scripts
+in this directory.
+
+In particular on macOS and Linux, an user may invoke
+
+```bash
+$ ./verify.sh [-b $buildir] [-n] [CHECK_VERSION]
+```
+
+Where:
+
+* `-b` build directory in which electron will be build
+* `-n` don't check for minimum required space, in this case 200GB
+* The first argument `CHECK_VERSION` represents a given tag of which the user is willing to verify determinism (eg. `v10.0.0`)
+
+While on Windows 
+```bash
+$ verify.bat [CHECK_VERSION] [-buildir $buildir]
+```
+
+Where:
+
+* `-buildir` build directory in which electron will be build
+* The first argument `CHECK_VERSION` represents a given tag of which the user is willing to verify determinism (eg. `v10.0.0`)
+
+## Notes
+
+It is important to note that while Linux releases is, at the time of writing,
+generating path independant symbols, it is not the case on Windows and macOS where the output
+directory will be important to be set respectively to `C:\projects\src`
+and `/Users/distiller/project/src` as reported in the current `toolchain_profile.json`.
--- a/script/reproducible/verify.bat
+++ b/script/reproducible/verify.bat
@@ -1,0 +1,129 @@
+@ECHO OFF
+set CHECK_VERSION=%1
+SHIFT
+
+set buildir=C:\projects\
+
+REM Interpret arguments
+:loop
+IF NOT "%1"=="" (
+    IF "%1"=="-buildir" (
+        SET buildir=%2
+        SHIFT
+    )
+    SHIFT
+    GOTO :loop
+)
+
+@ECHO ON
+
+set ELECTRON_OUT_DIR=Default
+set ELECTRON_ENABLE_STACK_DUMPING=1
+
+REM Check for disk space
+Rem        543210987654321
+Set "Blank=               "
+Set "GB100=   107374182400"
+
+for /f "tokens=2" %%A in (
+  'wmic LogicalDisk Get DeviceID^,FreeSpace ^| find /i "C:"'
+) Do Set "FreeSpace=%Blank%%%A"
+Set "FreeSpace=%FreeSpace:~-15%"
+
+Echo FreeSpace="%FreeSpace%"
+Echo    100 GB="%GB100%"
+
+If "%FreeSpace%" gtr "%GB100%" (
+  Echo yes enough free space
+) else (
+  Echo not enough free space - 100GB 
+  exit 5
+)
+
+REM Install missing dependencies
+powershell -command "& { iwr https://raw.githubusercontent.com/electron/electron/%CHECK_VERSION%/script/setup-win-for-dev.bat -OutFile C:\install.bat }"
+cmd /c C:\install.bat
+
+
+set GN_CONFIG=release
+
+echo "Building $env:GN_CONFIG build"
+SET PATH=%PATH%;C:\depot_tools\
+SET PATH=%PATH%;C:\Python27\
+git config --global core.longpaths true
+if not exist %buildir% mkdir %buildir%
+cd %buildir%
+set CHROMIUM_BUILDTOOLS_PATH=%buildir%\src\buildtools
+set SCCACHE_PATH=%buildir%\src\electron\external_binaries\sccache.exe
+REM set GYP_MSVS_OVERRIDE_PATH=C:\BuildTools\
+set GYP_MSVS_VERSION=2019
+set DEPOT_TOOLS_WIN_TOOLCHAIN=0
+REM set GCLIENT_EXTRA_ARGS="%GCLIENT_EXTRA_ARGS% --custom-var=\"checkout_boto=True\" --custom-var=\"checkout_requests=True\""
+cmd /C gclient config --name "src\electron" --unmanaged %GCLIENT_EXTRA_ARGS% "https://github.com/electron/electron"
+cmd /C gclient sync -f --with_branch_heads --with_tags
+cd src
+cd electron
+git fetch
+git checkout %CHECK_VERSION%
+cmd /c gclient sync -f
+cd ..
+cd ..
+set BUILD_CONFIG_PATH="//electron/build/args/release.gn"
+REM build out/Default
+cd src
+REM XXX: leave extra space after GN_EXTRA_ARG
+set GN_EXTRA_ARG= 
+cmd /C gn gen out/Default --args="import(\"//electron/build/args/release.gn\") %GN_EXTRA_ARG% cc_wrapper=\"%SCCACHE_PATH%\""
+ninja -C out/Default electron:electron_app
+cmd /C gn gen out/ffmpeg "--args=import(\"//electron/build/args/ffmpeg.gn\") %GN_EXTRA_ARGS%"
+ninja -C out/ffmpeg electron:electron_ffmpeg_zip
+ninja -C out/Default electron:electron_dist_zip
+ninja -C out/Default electron:electron_mksnapshot_zip
+ninja -C out/Default electron:hunspell_dictionaries_zip
+ninja -C out/Default electron:electron_chromedriver_zip
+ninja -C out/Default third_party/electron_node:headers
+cmd /C %SCCACHE_PATH% --show-stats
+cmd /C %SCCACHE_PATH% --stop-server
+
+cd out
+REM loop needed since sometimes the output directory is locked before completing the compilation
+:renloop
+timeout 1
+ren Default Default2 || goto :renloop
+ren ffmpeg ffmpeg2
+cd ..
+
+cmd /C gn gen out/Default --args="import(\"//electron/build/args/release.gn\") %GN_EXTRA_ARG% cc_wrapper=\"%SCCACHE_PATH%\""
+ninja -C out/Default electron:electron_app
+cmd /C gn gen out/ffmpeg "--args=import(\"//electron/build/args/ffmpeg.gn\") %GN_EXTRA_ARGS%"
+ninja -C out/ffmpeg electron:electron_ffmpeg_zip
+ninja -C out/Default electron:electron_dist_zip
+ninja -C out/Default electron:electron_mksnapshot_zip
+ninja -C out/Default electron:hunspell_dictionaries_zip
+ninja -C out/Default electron:electron_chromedriver_zip
+ninja -C out/Default third_party/electron_node:headers
+cmd /C %SCCACHE_PATH% --show-stats
+
+REM Check for local determinism, there is no reason in checking remote determinism if local is not satisfied
+pip install python-stripzip
+for %%f in (out/Default/*.zip) do (
+  stripzip out/Default/%%f
+)
+for %%f in (out/Default2/*.zip) do (
+  stripzip out/Default/%%f
+)
+python tools/determinism/compare_build_artifacts.py -f out/Default -s out/Default2
+
+REM Check for remote determinism comparing output zip hashes
+powershell -command "& { iwr https://github.com/electron/electron/releases/download/%CHECK_VERSION%/SHASUMS256.txt -OutFile C:\TEMP\SHASUMS256.txt }"
+
+ECHO Expected SHA256SUM
+type C:\TEMP\SHASUMS256.txt
+
+ECHO Calculated SHA256SUM
+for %%f in (out/Default/*.zip) do (
+  certutil -hashfile out/Default/%%f SHA256
+)
+for %%f in (out/Default2/*.zip) do (
+  certutil -hashfile out/Default/%%f SHA256
+)
--- a/script/reproducible/verify.sh
+++ b/script/reproducible/verify.sh
@@ -1,0 +1,195 @@
+#!/bin/bash -e
+SCRIPT=`realpath $0`
+buildir=/mnt
+
+usage() { echo "Usage: $0 [-b $buildir] [-n] [CHECK_VERSION]" 1>&2; exit 1; }
+
+setup_mac_for_dev() {
+    export CHECK_VERSION=$1
+
+    which -s brew
+    if [[ $? != 0 ]] ; then
+        # Install Homebrew
+        ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+    else
+        brew update
+    fi
+
+    which -s python2
+    if [[ $? != 0 ]] ; then
+        # Install python2
+        brew install python2
+    fi
+
+    which -s python3
+    if [[ $? != 0 ]] ; then
+        # Install python3
+        brew install python3
+    fi
+
+    if [ ! -f /usr/local/bin/gtar ]; then
+      brew install gnu-tar
+      ln -fs /usr/local/bin/gtar /usr/local/bin/tar
+    fi
+
+    XCODE_VERSION=$(curl -v https://raw.githubusercontent.com/electron/electron/${CHECK_VERSION}/.circleci/config.yml | grep "xcode:"  | head -n1 | awk '{print $2}' | tr -d '"')
+    echo "Since there is no way to automate XCode installation. Install XCode version $XCODE_VERSION manually."
+    echo "Press Key to continue."
+    open 'https://developer.apple.com/download/more/?q=xcode'
+    read -n1 ans
+}
+
+nocheckdisk=false
+
+while getopts ":b:n" o; do
+    case "${o}" in
+        b)
+            buildir=${OPTARG}
+            ;;
+        n)
+            nodiskcheck=true
+            ;;
+        *)
+            usage
+            ;;
+    esac
+done
+shift $((OPTIND-1))
+export CHECK_VERSION=$1
+
+if [[ "$OSTYPE" == "linux-gnu" ]] && cat /proc/1/cgroup | grep docker &>/dev/null; then
+  # If you are in docker run youself and that's it
+  echo "Executing under docker image, continuing"
+  git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git $HOME/depot_tools
+  export GCLIENT_PY3=0
+  export PATH=$HOME/depot_tools:$PATH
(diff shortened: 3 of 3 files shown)


script/setup-mac-for-dev.sh
```diff
@@ -0,0 +1,34 @@
...
+    # Install python3
+    brew install python3
+fi
+
+if [ ! -f /usr/local/bin/gtar ]; then
+	brew install gnu-tar
+	ln -fs /usr/local/bin/gtar /usr/local/bin/tar
+fi
+
+XCODE_VERSION=$(curl -v https://raw.githubusercontent.com/electron/electron/${CHECK_VERSION}/.circleci/config.yml | grep "xcode:"  | head -n1 | awk '{print $2}' | tr -d '"')
```
jkleinsc, line 30: Why doesn't this use the locally checked out .circleci/config.yml file?  If you are working on an older version of Electron, it will most like be using an older version of xcode

script/verify-reproducible.bat
```diff
@@ -0,0 +1,129 @@
...
+
+If "%FreeSpace%" gtr "%GB100%" (
+  Echo yes enough free space
+) else (
+  Echo not enough free space - 100GB 
+  exit 5
+)
+
+REM Install missing dependencies
+powershell -command "& { iwr https://raw.githubusercontent.com/electron/electron/%CHECK_VERSION%/script/setup-win-for-dev.bat -OutFile C:\install.bat }"
```
jkleinsc, line 44: Why not use local version?

script/setup-mac-for-dev.sh
thypon: My initial idea was to let the user download the script alone and pass all the needed arguments to the `verify-reproducibility.sh` script. That alone would suffice to actually download and verify electron.

That would be enough to bootstrap all the process from a single script. Do you think is better to download the whole repository before instead?

jkleinsc: ok that makes sense about downloading the script, but that means that it wouldn't be helpful for older versions of Electron.

andrea-doyensec: No it won't for macOS unfortunately. It should instead work for Linux. Do you think is better to embed the setup phase inside the same script for macOS too, to have it self contained?

thypon: No it won't for macOS unfortunately. It should instead work for Linux. Do you think is better to embed the setup phase inside the same script for macOS too, to have it self contained?

zcbenz:
Maybe put the scripts in a sub-dir under `scripts/` with a README on what they are and how to use them?

script/reproducible/README.md
```diff
@@ -0,0 +1,35 @@
...
+Where:
+
+* `-buildir` build directory in which electron will be build
+* The first argument `CHECK_VERSION` represents a given tag of which the user is willing to verify determinism (eg. `v10.0.0`)
+
+## Notes
+
+It is important to note that while macOS and Linux releases are, at the time of writing,
+generating path independant symbols, it is not the case on Windows where the output
+directory will be important to be set to `C:\\projects\\src` as reported in the current `toolchain_profile.json`.
```
jkleinsc, line 35:
```suggestion
directory will be important to be set to `C:\projects\src` as reported in the current `toolchain_profile.json`.
```

script/setup-mac-for-dev.sh
jkleinsc: I guess its fine to leave the scripts as is.

nornagon:
I'm not thrilled about the idea of maintaining a fairly sizable Windows batch script, & one with no tests, to boot.

Would it make sense to reduce the amount that these bash/batch scripts do and move the bulk of the logic to python? Also, how will we know if we do something that breaks these scripts?

andrea-doyensec:
@nornagon at the moment both `verify.{sh,bat}` scripts are a direct translation of, respectively, `appveyor.yml` file and `.circleci/config.yml`. A possible future improvement when we verified these two scripts are fully useable is implementing a unifiying python script that get both the steps from `appveyor.yml` and `.circleci/config.yml` and executes them sequentially. At the end it checks both reproducibility for the two configurations.

Unless we want to test these scripts every release I find it difficult to test these integration scripts.
Their aim is testing for local reproducibility and remote reproducibility. So they compile electron twice and then they download the tarball sha256sum to verify if they are correct. In this particular design we put the burden of verifying local reproduciblity to the verifier. Once this local reproducibility is verifyied every release in electron CI, we can reduce these to build and check the checksum.

I have a couple of questions too:

- Is there any plan of switching from appveyor to circleci windows in the foreseable future?
- Is having local reproducibility tests something electron team may consider for releases?

thypon:
@nornagon at the moment both `verify.{sh,bat}` scripts are a direct translation of, respectively, `appveyor.yml` file and `.circleci/config.yml`. A possible future improvement when we verified these two scripts are fully useable is implementing a unifiying python script that get both the steps from `appveyor.yml` and `.circleci/config.yml` and executes them sequentially. At the end it checks both reproducibility for the two configurations.

Unless we want to test these scripts every release I find it difficult to test these integration scripts.
Their aim is testing for local reproducibility and remote reproducibility. So they compile electron twice and then they download the tarball sha256sum to verify if they are correct. In this particular design we put the burden of verifying local reproduciblity to the verifier. Once this local reproducibility is verifyied every release in electron CI, we can reduce these to build and check the checksum.

I have a couple of questions too:

- Is there any plan of switching from appveyor to circleci windows in the foreseable future?
- Is having local reproducibility tests something electron team may consider for releases?

jkleinsc:
@thypon
> Is there any plan of switching from appveyor to circleci windows in the foreseable future?

While we would love to move all our CI over to CircleCI, right now the Windows functionality in CircleCI doesn't work for Electron.  We are talking to them about this, but my gut says it will be a while before we could migrate over.

> Is having local reproducibility tests something electron team may consider for releases?

Do you mean to run reproducibility tests as part of our release process?  Given how long release builds take, I'm currently not comfortable running release builds twice.  Perhaps other members of @electron/wg-releases have other thoughts on this.

script/reproducible/verify.bat
```diff
@@ -0,0 +1,129 @@
...
+set GN_CONFIG=release
+
+echo "Building $env:GN_CONFIG build"
+SET PATH=%PATH%;C:\depot_tools\
+SET PATH=%PATH%;C:\Python27\
+git config --global core.longpaths true
+if not exist %buildir% mkdir %buildir%
+cd %buildir%
+set CHROMIUM_BUILDTOOLS_PATH=%buildir%\src\buildtools
+set SCCACHE_PATH=%buildir%\src\electron\external_binaries\sccache.exe
```
nornagon, line 57: I'm confused why sccache would be involved in a reproducibility test. Shouldn't we be building from scratch both times?

script/reproducible/verify.bat
```diff
@@ -0,0 +1,129 @@
...
+REM set GYP_MSVS_OVERRIDE_PATH=C:\BuildTools\
```
nornagon, line 58: Remove comment cruft

script/reproducible/verify.bat
```diff
@@ -0,0 +1,129 @@
...
+set GYP_MSVS_VERSION=2019
+set DEPOT_TOOLS_WIN_TOOLCHAIN=0
+REM set GCLIENT_EXTRA_ARGS="%GCLIENT_EXTRA_ARGS% --custom-var=\"checkout_boto=True\" --custom-var=\"checkout_requests=True\""
```
nornagon, line 61: here too

script/reproducible/verify.bat
```diff
@@ -0,0 +1,129 @@
...
+cmd /C gclient config --name "src\electron" --unmanaged %GCLIENT_EXTRA_ARGS% "https://github.com/electron/electron"
+cmd /C gclient sync -f --with_branch_heads --with_tags
```
nornagon, line 63: Why is this run with `cmd /c` instead of just directly invoking `gclient`?

script/reproducible/verify.bat
```diff
@@ -0,0 +1,129 @@
...
+cd electron
+git fetch
+git checkout %CHECK_VERSION%
+cmd /c gclient sync -f
+cd ..
+cd ..
+set BUILD_CONFIG_PATH="//electron/build/args/release.gn"
+REM build out/Default
+cd src
+REM XXX: leave extra space after GN_EXTRA_ARG
```
nornagon, line 74: Why?

script/reproducible/verify.bat
```diff
@@ -0,0 +1,129 @@
...
+set GN_EXTRA_ARG= 
+cmd /C gn gen out/Default --args="import(\"//electron/build/args/release.gn\") %GN_EXTRA_ARG% cc_wrapper=\"%SCCACHE_PATH%\""
+ninja -C out/Default electron:electron_app
```
nornagon, line 77:
```suggestion
ninja -C out/Default electron:electron

```

script/reproducible/verify.bat
```diff
@@ -0,0 +1,129 @@
...
+cmd /C gn gen out/ffmpeg "--args=import(\"//electron/build/args/ffmpeg.gn\") %GN_EXTRA_ARGS%"
```
nornagon, line 78: This script seems to be inconsistent w.r.t `GN_EXTRA_ARG` vs `GN_EXTRA_ARGS`. Are either of these variables needed?

script/reproducible/verify.sh
```diff
@@ -0,0 +1,195 @@
...
+  docker run -ti -v ${buildir}:${buildir} ${EXTRA_DOCKER_ARGS} -v ${SCRIPT}:/verify.sh ${IMAGE} bash -e /verify.sh -b ${buildir} $([ "$nodiskcheck" = "true" ] && printf " -n") ${CHECK_VERSION}
+  exit 0
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+  # If you are on mac run youself pointing to the correct version of toolchain
+  setup_mac_for_dev "${CHECK_VERSION}" || true
+  echo "Executing under macOS image, continuing"
+fi
+
+export ELECTRON_OUT_DIR=Default
+export ELECTRON_ENABLE_STACK_DUMPING=1
```
nornagon, line 78: I don't believe this is relevant to the build process, it's only relevant for running.

script/reproducible/verify.sh
```diff
@@ -0,0 +1,195 @@
...
+  echo not enough free space - ${MINSPACE}GB 
+  exit 5
+fi
+
+export GN_CONFIG=release
+export STRIP_BINARIES=true 
+export GENERATE_SYMBOLS=true
+export CHECK_DIST_MANIFEST=1
+
+echo "Building $env:GN_CONFIG build"
```
nornagon, line 96:
```suggestion
echo "Building $GN_CONFIG build"
```

script/reproducible/verify.sh
```diff
@@ -0,0 +1,195 @@
...
+git config --global core.longpaths true
```
nornagon, line 97: core.longpaths is only relevant on windows

script/reproducible/verify.sh
```diff
@@ -0,0 +1,195 @@
+#!/bin/bash -e
```
nornagon, line 1: recommend `set -eo pipefail`

script/reproducible/verify.sh
```diff
@@ -0,0 +1,195 @@
...
+export GENERATE_SYMBOLS=true
+export CHECK_DIST_MANIFEST=1
+
+echo "Building $env:GN_CONFIG build"
+git config --global core.longpaths true
+mkdir -p $buildir
+cd $buildir
+export CHROMIUM_BUILDTOOLS_PATH=$buildir/src/buildtools
+export SCCACHE_PATH=$buildir/src/electron/external_binaries/sccache
+#export GCLIENT_EXTRA_ARGS="$GCLIENT_EXTRA_ARGS --custom-var=\"checkout_boto=True\" --custom-var=\"checkout_requests=True\""
```
nornagon, line 102: Remove comment cruft

script/reproducible/verify.sh
```diff
@@ -0,0 +1,195 @@
...
+gclient config --name "src/electron" --unmanaged $GCLIENT_EXTRA_ARGS "https://github.com/electron/electron"
+gclient sync -f --with_branch_heads --with_tags
+cd src
+cd electron
+git fetch
+git checkout $CHECK_VERSION
+gclient sync -f
+cd ..
+cd ..
```
nornagon, line 111:
```suggestion
(
  cd src/electron
  git fetch
  git checkout $CHECK_VERSION
)
gclient sync -f
```

script/reproducible/verify.sh
nornagon: Also, the `GN_CONFIG` variable doesn't seem to be used anywhere. Perhaps remove it?

script/reproducible/verify.bat
```diff
@@ -0,0 +1,129 @@
...
+cmd /C %SCCACHE_PATH% --stop-server
+
+cd out
+REM loop needed since sometimes the output directory is locked before completing the compilation
+:renloop
+timeout 1
+ren Default Default2 || goto :renloop
+ren ffmpeg ffmpeg2
+cd ..
+
```
nornagon, line 95: If using sccache to match CI, you should clear the sccache cache between builds.

script/reproducible/verify.sh
```diff
@@ -0,0 +1,195 @@
...
+export BUILD_CONFIG_PATH="//electron/build/args/release.gn"
+# build out/Default
+$SCCACHE_PATH --show-stats
+cd src
+# XXX: leave extra space after GN_EXTRA_ARG
+# GN_EXTRA_ARG=
```
nornagon, line 117:
```suggestion
```

script/reproducible/verify.sh
```diff
@@ -0,0 +1,195 @@
...
+cd ..
+cd ..
+export BUILD_CONFIG_PATH="//electron/build/args/release.gn"
+# build out/Default
+$SCCACHE_PATH --show-stats
+cd src
+# XXX: leave extra space after GN_EXTRA_ARG
+# GN_EXTRA_ARG= 
+gn gen out/Default --args="import(\"//electron/build/args/release.gn\") $GN_EXTRA_ARG cc_wrapper=\"$SCCACHE_PATH\""
+ninja -C out/Default electron:electron_app
```
nornagon, line 119:
```suggestion
ninja -C out/Default electron:electron
```

script/reproducible/verify.sh
```diff
@@ -0,0 +1,195 @@
...
+  echo yes enough free space
+else
+  echo not enough free space - ${MINSPACE}GB 
+  exit 5
+fi
+
+export GN_CONFIG=release
+export STRIP_BINARIES=true 
+export GENERATE_SYMBOLS=true
+export CHECK_DIST_MANIFEST=1
```
nornagon, line 94:
```suggestion
```

script/reproducible/verify.sh
```diff
@@ -0,0 +1,195 @@
...
+ninja -C out/Default electron:electron_dist_zip
+ninja -C out/Default electron:electron_mksnapshot_zip
+ninja -C out/Default electron:hunspell_dictionaries_zip
+ninja -C out/Default electron:electron_chromedriver_zip
+ninja -C out/Default third_party/electron_node:headers
+$SCCACHE_PATH --show-stats
+$SCCACHE_PATH --stop-server
+
+cd out
+# loop needed since sometimes the output directory is locked before completing the compilation
```
nornagon, line 145: is this true? i don't think linux or mac "lock" directories.

script/reproducible/verify.sh
```diff
@@ -0,0 +1,195 @@
...
+while true; do
+  sleep 1
+  mv Default Default2 || continue
+  mv ffmpeg ffmpeg2 || continue
+  break
+done
+cd ..
+
```
nornagon, line 153: clear the sccache cache.

script/reproducible/verify.bat
thypon: So, since the cache seems to be a major offender in this particular case, maybe 1 build without cache, 2 with the cache should be better.

What we want to actually test is that build is:
1. local reproducible when built twice in the same setup of the remote machine (so sccache enabled)
2. that sccache is not introducing weird caching strategies that breaks the correctness of the executable.

script/reproducible/verify.bat
thypon: remember to move to `call` every `cmd /c`. What is happening here is that it is exiting after the gclient sync execution. If you execute a nested .bat script, for legacy reason, when you `exit` also the father `.bat` is signalled to exit.

thypon: `cmd /C` spawn a new shell with a separated environment, (inherithed), `call` should instead spawn in the caller environment, and can change it accordingly.

script/reproducible/verify.bat
thypon: If removed, during the expansion at line 76, it won't work properly, maybe I should say why in the comment.

script/reproducible/verify.sh
thypon: we need to remove this line, I ported from verify.bat and these lines are a leftover.

script/reproducible/verify.sh
thypon: Isn't electron_app correct to build the app for macOS?

script/reproducible/verify.sh
thypon: This one I think is needed for Linux, since this is done on that platform.

script/reproducible/verify.sh
thypon: I ported straight the code from the Windows part. What it was happening there was that sccache was getting lock of out directory and, until the server was stopped completely, the out directory was locked .  I think this one may even be removed, if we assume that --stop-server will always succeed.

script/reproducible/verify.sh
nornagon: These variables are only used in the ci script, which isn't being referenced here, so they're not relevant.

script/reproducible/verify.sh
nornagon: `electron:electron` is the right target to build on all platforms.

script/reproducible/verify.bat
nornagon: This variable doesn't seem to have any function. Can it be removed?

script/reproducible/verify.bat
nornagon: ah, so it's to ignore failures?

script/reproducible/verify.bat
nornagon: it's fine to use sccache in the build i suppose, as long as the cache is cleared between builds. Otherwise it's not really testing reproducibility, right? Since it's just reusing the build outputs.

script/reproducible/verify.sh
thypon: Honestly not knowing where these variables are used, one by one, I tried to use the most similar environment of the build machine. It may even be some of them will end in the build, and during a test like that you prefer to avoid false positives due to small environment changes.

script/reproducible/verify.bat
thypon: That is actually a better solution than running the build 3 times. We lost on correctness check for the caching system.

script/reproducible/verify.sh
nornagon: I am confident that this variable will not affect the built output. If it does, that is a bug.

script/reproducible/verify.bat
thypon: not really failures, in `.bat` failures are always ignored unless you check the return value of every call/exe invocation. It's that whenever you invoke another .bat script (not an .exe), the exit signal will bounce back to the root .bat unless you have a `call` or `cmd /C` and will trigger a exit on the main root .bat, in our case `verify.bat`. I was confused too by this behavior initially, but it's due to `bat` being very legacy friendly.

thypon: This behavior does not have any similarity with how exit and shell nesting is handled on posix shells. Basically you have this weird behaviour or:

1. `call another.bat` that is similar to `. another.sh`
2. `cmd /C another.sh` that is similar to `sh another.sh`

`another.bat` invocation without anything is not equivalent to `./another.sh`. It's more like both `root.bat` and `another.bat` are having a shared `exit` signal handler and when the first bat invoked exit, it invokes the `exit` signal and bounce the exit.

thypon:
Thanks @nornagon for the feedback. I will go through a major cleanup of both the Linux and Windows scripts based on that.

miniak:
@thypon are you still planning to land this PR?

ikkisoft:
Hi @miniak - @thypon is no longer working at Doyensec, however he mentioned that would be happy to help finalizing the PR. Additionally, we're also available to allocate some days of another engineer to get to the finish line. One way or another, we can make it happen.

The real question is whether the maintainers team is still interested in reproducible builds, and whether adjusting these sh / bat scripts is sufficient. For instance, we won't have the bandwidth to port all to Python.

codebytere:
Can this be closed?

jkleinsc:
Closing as this PR is quite outdated and we may potentially pursue sigstore instead of reproducible builds.

PR #22976, opened 2020-04-06
ci: auto-3way patches and detect changes

nornagon:
#### Description of Change
This change does two things:
1. It automatically attempts 3-way merging of patches, and
2. Requires that `git-export-patches` in master always produces 0 changes. i.e. that the `patches` folder always represents exactly what you'd get if you exported all the patches from all the patched repos.

There are two goals I'm going for here. The first is that I'd like CI to be able to provide better support in the case that a simple 3-way merge is sufficient to correct a non-applying patch. I'd love it if CircleCI was able to create a commit back in the PR with a patch that would reflect the result of the 3-way merge, but in lieu of that, I figure at least a patch file uploaded into the CI artifacts would be helpful & can be applied manually without having to do a full `gclient sync`.

The second goal is to reduce spurious patch changes which come from patches being exported without a full `git-export-patches`. It's common for the repository to be in a state where the patches directory doesn't accurately reflect what you'd get from a `git-export-patches` run, which means that when you do run `git-export-patches` after making a change, you end up with a bunch of changes to unrelated patches.

#### Checklist

#### Release Notes

Notes: none


nornagon:
> Should we also run on Windows?

I don't see a particular reason to do so. Patch output should be the same on all platforms.

patches/chromium/mas_disable_remote_layer.patch
```diff
@@ -74,15 +74,15 @@ diff --git a/ui/accelerated_widget_mac/display_ca_layer_tree.mm b/ui/accelerated
...
+@@ -97,6 +97,7 @@ void DisplayCALayerTree::UpdateCALayerTree(
  }
  
  void DisplayCALayerTree::GotCALayerFrame(uint32_t ca_context_id) {
 +#ifndef MAS_BUILD
    // Early-out if the remote layer has not changed.
    if ([remote_layer_ contextId] == ca_context_id)
      return;
-@@ -121,6 +122,9 @@ - (void)setContentsChanged;
+@@ -121,6 +122,9 @@ void DisplayCALayerTree::GotCALayerFrame(uint32_t ca_context_id) {
```
nornagon, line 85: this seems better...

patches/chromium/mas_no_private_api.patch
```diff
@@ -53,39 +53,39 @@ diff --git a/content/browser/accessibility/browser_accessibility_cocoa.mm b/cont
...
+@@ -479,6 +482,7 @@ NSAttributedString* GetAttributedTextForTextMarkerRange(id marker_range) {
    AddMisspelledTextAttributes(ax_range, attributed_text);
    return attributed_text;
  }
 +#endif
  
  // Returns an autoreleased copy of the AXNodeData's attribute.
  NSString* NSStringForStringAttribute(BrowserAccessibility* browserAccessibility,
-@@ -772,7 +776,9 @@ + (void)initialize {
+@@ -772,7 +776,9 @@ NSString* const NSAccessibilityRequiredAttributeChrome = @"AXRequired";
```
nornagon, line 88: ... but this seems worse

DEPS
```diff
@@ -155,3 +155,5 @@ hooks = [
 recursedeps = [
   'src',
 ]
+
+# Touch deps again to fool circleci
```
jkleinsc, line 159: Maybe remove this bit?

PR #22977, opened 2020-04-06
chore: cherry-pick adc8f05aa3ab from chromium

nornagon:
Handle ERR_CACHE_RACE in DoDoneHeadersAddToEntryComplete

Crashes have been observed in the wild that may be due to
ERR_CACHE_RACE being sent to STATE_DONE_HEADERS_ADD_TO_ENTRY_COMPLETE.
These crashes, however, cannot be reproduced and code investigation
does not suggest how they might happen. If the crashes stop occurring
following this change, we will be sure that this is the cause and will
perform more analysis based on that understanding.

Bug: 959194
Change-Id: I01649ee31024b598e41c0beda7c7ed54980cb133
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/1814843
Commit-Queue: Alex Turner <alexmt@chromium.org>
Reviewed-by: Shivani Sharma <shivanisha@chromium.org>
Cr-Commit-Position: refs/heads/master@{#698597}

Notes: none

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -99,3 +99,4 @@ allow_restricted_clock_nanosleep_in_linux_sandbox.patch
 move_readablestream_requests_onto_the_stack_before_iteration.patch
 streams_convert_state_dchecks_to_checks.patch
 cherry-pick-fd211b44535c.patch
+cherry-pick-adc8f05aa3ab.patch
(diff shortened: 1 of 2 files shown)


jkleinsc:
Merging as CI failures are flakes unrelated to PR change.

PR #22985, opened 2020-04-06
build: set merge=union for .patches

nornagon:
#### Description of Change
Hopefully this will reduce patch conflicts in .patches files. Not sure if github understands it but it's worth a shot!

Notes: none


ZakreyaAlsaiha:
05117088

PR #22997, opened 2020-04-07
chore: make WebContentsView take webPreferences as parameter

zcbenz:
#### Description of Change

This PR changes the public JS constructor of `WebContentsView` to take `web_preferences` as parameter and create `WebContents` inside it. This is required for making the APIs public (https://github.com/electron/governance/pull/254). I'll write documentations in another PR.

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/electron_api_web_contents_view.cc
+++ b/shell/browser/api/electron_api_web_contents_view.cc
@@ -45,6 +45,25 @@ namespace electron {
 
 namespace api {
 
+namespace {
+
+// Helper to create WebContentsView by calling from constructor.
+gin::Handle<WebContentsView> CallConstructor(
+    v8::Isolate* isolate,
+    v8::Local<v8::Function> constructor,
+    v8::Local<v8::Value> arg) {
+  v8::Local<v8::Context> context = isolate->GetCurrentContext();
+  v8::Local<v8::Object> obj;
+  if (constructor->NewInstance(context, 1, &arg).ToLocal(&obj)) {
+    gin::Handle<WebContentsView> web_contents_view;
+    if (gin::ConvertFromV8(isolate, obj, &web_contents_view))
+      return web_contents_view;
+  }
+  return gin::Handle<WebContentsView>();
+}
+
+}  // namespace
+
 WebContentsView::WebContentsView(v8::Isolate* isolate,
                                  gin::Handle<WebContents> web_contents)
 #if defined(OS_MACOSX)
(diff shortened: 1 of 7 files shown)


shell/browser/api/electron_api_web_contents_view.cc
```diff
@@ -78,38 +57,78 @@ WebContentsView::~WebContentsView() {
...
+  if (GetConstructor(isolate)->NewInstance(context, 1, &arg).ToLocal(&obj)) {
+    gin::Handle<WebContentsView> web_contents_view;
+    if (gin::ConvertFromV8(isolate, obj, &web_contents_view))
+      return web_contents_view;
+  }
+  return gin::Handle<WebContentsView>();
+}
+
+// static
+v8::Local<v8::Function> WebContentsView::GetConstructor(v8::Isolate* isolate) {
```
deepak1556, line 85: Is it possible to make it a `gin_helper::Constructible`, would match the style of upcoming classes or is it better to handle it in a separate PR ?

shell/browser/api/electron_api_web_contents_view.cc
```diff
@@ -78,38 +57,78 @@ WebContentsView::~WebContentsView() {
...
 }
 
+// static
+gin::Handle<WebContentsView> WebContentsView::Create(
+    v8::Isolate* isolate,
+    const gin_helper::Dictionary& web_preferences) {
+  v8::Local<v8::Context> context = isolate->GetCurrentContext();
+  v8::Local<v8::Value> arg = gin::ConvertToV8(isolate, web_preferences);
+  v8::Local<v8::Object> obj;
+  if (GetConstructor(isolate)->NewInstance(context, 1, &arg).ToLocal(&obj)) {
```
deepak1556, line 76: Why does `Create` have to go through the V8 binding to call `New` and wrap the handle ? If the call is only happening on c++ land, can it not directly call `New` and wrap the result ?

shell/browser/api/electron_api_web_contents_view.cc
zcbenz: `Constructible` uses `FillObjectTemplate` instead of `BuildPrototype` to fill the prototype, and it requires declaring a static `kWrapperInfo` member, I would like to use it after converting all `View` classes to `gin::Wrappable`, otherwise the classes would be in mixed styles.

shell/browser/api/electron_api_web_contents_view.cc
zcbenz: Calling `New` is equivalent to calling `WebContentsView()` in JS which does not allocate a holder, we have to call `new WebContentsView()` in JS and the only way to do that in JS is to call the `v8::Function::NewInstance` API.

zcbenz: Classes based on `gin::Wrappable` does not need to do this because the objects are generated from `v8::ObjectTemplate`, while the `View` classes have to use `v8::FunctionTemplate` to implement prototype inheritance.

zcbenz: We should be able to call New directly once we figured out a way to move the `View` classes to `gin::Wrappable`, but currently a `Create` method is required.

PR #22998, opened 2020-04-07
build: implement zip64 stripping

thypon:
#### Description of Change

When zip contains files larger than 4GB, 4bytes length headers are
not sufficient anymore. Zip64 defines an extra header 0x0001 which
may contain 8byte lengthed (16 exabytes) lengths [uncompressed and compressed].

Read this value when performing extra data cleaning and override the
bogus value if the header is available.

Read https://blog.yaakov.online/zip64-go-big-or-go-home/ for more
information on Zip64 extra header.

This is the first known implementation of zip64 stripping.

This change was needed since electron CI explicitly enable zip64
and some files inside these zips may exceed the maximum zip file size (4GB).
This is particularly true on Linux when debuggings symbols are enabled.

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/script/release/uploaders/upload.py
+++ b/script/release/uploaders/upload.py
@@ -182,7 +182,7 @@ def zero_zip_date_time(fname):
 
 
 def _zero_zip_date_time(zip_):
-  def purify_extra_data(mm, offset, length):
+  def purify_extra_data(mm, offset, length, compressed_size=0):
     extra_header_struct = Struct("<HH")
     # 0. id
     # 1. length
@@ -194,6 +194,16 @@ def _zero_zip_date_time(zip_):
     UNIX_EXTRA_DATA = 0x7875
     # Unix extra data; UID / GID stuff, see
     # ftp://ftp.info-zip.org/pub/infozip/src/zip30.zip ./proginfo/extrafld.txt
+    ZIP64_EXTRA_HEADER = 0x0001
+    zip64_extra_struct = Struct("<HHQQ")
+    # ZIP64.
+    # When a ZIP64 extra field is present his 8byte length
+    # will override the 4byte length defined in canonical zips.
+    # This is in the form:
+    # - 0x0001 (header_id)
+    # - 0x0010 [16] (header_length)
+    # - ... (8byte uncompressed_length)
+    # - ... (8byte compressed_length)
     mlen = offset + length
 
     while offset < mlen:
@@ -205,15 +215,17 @@ def _zero_zip_date_time(zip_):
 
       if header_id in (EXTENDED_TIME_DATA, UNIX_EXTRA_DATA):
         values[0] = STRIPZIP_OPTION_HEADER
-        for i in xrange(2, len(values)):
+        for i in range(2, len(values)):
           values[i] = 0xff
         extra_struct.pack_into(mm, offset, *values)
-      elif header_id != STRIPZIP_OPTION_HEADER:
-        return False
+      if header_id == ZIP64_EXTRA_HEADER:
+        assert header_length == 16
+        values = list(zip64_extra_struct.unpack_from(mm, offset))
+        header_id, header_length, uncompressed_size, compressed_size = values
 
       offset += extra_header_struct.size + header_length
 
-    return True
+    return compressed_size
 
   FILE_HEADER_SIGNATURE = 0x04034b50
   CENDIR_HEADER_SIGNATURE = 0x02014b50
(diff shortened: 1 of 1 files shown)


nornagon:
Are you including debugging symbols in the reproducible build? Are non-debugging-symbol zips exceeding 4GB sizes?

thypon:
For what I can see zip64 is enabled regardless of 4GB+ files when we are zipping. This, sometimes, causes the zipfile to not be stripped **even if** the file is not 4GB+. I noticed this in mksnapshot zips on Linux for example.

Zip64 spec tells that the size can be coded as 8+8 bytes in the extra header, regardless of the size of the file that is going to be zipped. Supporting Zip64 length encoding guards us also from other edge cases.

script/release/uploaders/upload.py
```diff
@@ -205,15 +213,17 @@ def purify_extra_data(mm, offset, length):
...
-        for i in xrange(2, len(values)):
+        for i in range(2, len(values)):
           values[i] = 0xff
         extra_struct.pack_into(mm, offset, *values)
-      elif header_id != STRIPZIP_OPTION_HEADER:
-        return False
+      if header_id == ZIP64_EXTRA_HEADER:
+        assert header_length == 16
+        values = list(zip64_extra_struct.unpack_from(mm, offset))
+        header_id, header_length, uncompressed_size, compressed_size = values[0], values[1], values[2], values[3]
```
nornagon, line 222:
```suggestion
        header_id, header_length, uncompressed_size, compressed_size = values
```

script/release/uploaders/upload.py
```diff
@@ -205,15 +213,17 @@ def purify_extra_data(mm, offset, length):
...
       if header_id in (EXTENDED_TIME_DATA, UNIX_EXTRA_DATA):
         values[0] = STRIPZIP_OPTION_HEADER
-        for i in xrange(2, len(values)):
+        for i in range(2, len(values)):
           values[i] = 0xff
         extra_struct.pack_into(mm, offset, *values)
-      elif header_id != STRIPZIP_OPTION_HEADER:
-        return False
+      if header_id == ZIP64_EXTRA_HEADER:
+        assert header_length == 16
```
nornagon, line 220:
```suggestion
        assert header_length == zip64_extra_struct.size
```

script/release/uploaders/upload.py
```diff
@@ -194,6 +194,14 @@ def purify_extra_data(mm, offset, length):
     UNIX_EXTRA_DATA = 0x7875
     # Unix extra data; UID / GID stuff, see
     # ftp://ftp.info-zip.org/pub/infozip/src/zip30.zip ./proginfo/extrafld.txt
+    ZIP64_EXTRA_HEADER = 0x0001
+    zip64_extra_struct = Struct("<HHQQ")
+    # ZIP64 header
+    # This is usually in the form:
```
nornagon, line 200: "usually"?

script/release/uploaders/upload.py
andrea-doyensec: zip_extra_struct contains both 2 8 byte (QQ) and two 2 bytes (header_id, header_length), so the total should be 20 and not 16.

script/release/uploaders/upload.py
thypon: Isn't values unpacking without the index only working on python3 (it is not python2 compatible)

nornagon:
```
$ python2.7
>>> a, b, c = [1,2,3]
>>> a
1
>>> b
2
>>> c
3
```

andrea-doyensec: Oh right, I confused that with the `a, b, *rest = [1,2,3,4]` syntax

```
>>> a, b, *rest = [1,2,3,4]
  File "<stdin>", line 1
    a, b, *rest = [1,2,3,4]
          ^
SyntaxError: invalid syntax
```

script/release/uploaders/upload.py
nornagon: Why's the header being parsed twice..?

andrea-doyensec: It is parsed three times actually.

1. First time I only get the `header_id` and `header_length` in `extra_header_struct` ("<HH")
2. second time I get also `header_id` + `header_length` and all his body bytes in `extra_struct` ("<HHB*" where B numbers are equal to the second H)
3. third time, only if it is a `ZIP64_EXTRA_HEADER` I read in the `zip64_extra_struct` ("<HHQQ")

HH is alway the header_id and header_length
body meaning changes wrt what is the header_id, and teoretically can be used to store even stenography. It is really dependant on the implementation what data ends in the Extras of each file. What I try to do in the purify_extra_data is handling the most common and, from this commit, also handling zip64, that, even if it's rarely used, we do use it to compress electron releases.

andrea-doyensec: I preferred read 3 times in the worst case and render the code more readable instead of playing with back and forth offsets.

nornagon:
@jkleinsc can we get a release build run against this PR, since the changed script is not covered by CI.

jkleinsc:
@nornagon already done.

PR #23009, opened 2020-04-08
chore: cherry-pick 4c57222340cf from chromium

nornagon:
Make finished_source_handlers_ hold scoped_refptrs

Previously, finished_source_handlers_ held raw pointers to
AudioHandlers and assumed that active_source_handlers_ also had a
copy.  But when the context goes away, active_source_handlers_ would
be cleared, but not finished_source_handlers_, leaving pointers to
deleted objects.

So do two things:
1. Change finished_source_handlers_ to hold scoped_refptrs to manage
   lifetime of the objects
2. Clear finished_source_handler_ in ClearHandlersToBeDeleted()

Either of these fix the repro case, but let's do both.  Don't want to
leaving dangling objects.

Manually tested the repro case which no longer reproduces.

Bug: 1059686
Change-Id: I2f30c996e8589fa5c3890d32500c4bb4f3bc4286
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2098260
Cr-Commit-Position: refs/heads/master@{#749302}

Notes: Security: backported fix for CVE-2020-6449: Use after free in audio.


deepak1556:
Build needs to be fixed.

nornagon:
Should wait for https://github.com/electron/electron/pull/23013

PR #23013, opened 2020-04-08
chore: cherry-pick 3ac8883297e1 from chromium

nornagon:
Break connections before removing from active_source_handlers_.

In DeferredTaskHandler::BreakConnections, we want to remove finished
handlers and break the connection.  when a finished handler is removed
from active_source_handlers_, it might be deleted, but we were still
using that to create the connection.  Instead, break the connection
first and then remove it.

Manually ran test from the bug and it passes with this change.  Without
this, it failed right away.

Bug: 1057593
Change-Id: I3c9346a6842f412100d608876adb268befb80470
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2083436
Cr-Commit-Position: refs/heads/master@{#746142}

Notes: Security: backported fix for CVE-2020-6428: Use after free in audio.


deepak1556:
Changes are in webaudio codepath, but content tracing specs have failed in both mas and osx builds. Not sure whats happening here, but should be a flake. Retriggered tests.

nornagon:
skipping most CI as there's minimal changes since last green.

PR #23016, opened 2020-04-08
chore: cherry-pick 2cd0af7ea205 from chromium

nornagon:
Use WeakPtr for cross-thread posting

{IIR,Biquad}FilterNodes check the state of the filter and notify the
main thread when it goes bad. In this process, the associated context
can be collected when a posted task is performed sometime later
in the task runner's queue.

By using WeakPtr, the task runner will not perform a scheduled task
in the queue when the target object is invalid anymore.

Test: Locally confirmed that the repro case does not crash after 30 min.
Bug: 1055788
Change-Id: Icdb3a7378d0345936b5b50e12ec2b187e58a611c
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2074807
Reviewed-by: Kentaro Hara <haraken@chromium.org>
Cr-Commit-Position: refs/heads/master@{#744936}

Notes: Security: backported fix for CVE-2020-6427: Use after free in audio.

### DIFF
--- a/patches/common/chromium/.patches
+++ b/patches/common/chromium/.patches
@@ -102,3 +102,4 @@ move_readablestream_requests_onto_the_stack_before_iteration.patch
 streams_convert_state_dchecks_to_checks.patch
 audiocontext_haspendingactivity_unless_it_s_closed.patch
 protect_automatic_pull_handlers_with_mutex.patch
+cherry-pick-2cd0af7ea205.patch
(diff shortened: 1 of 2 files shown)


deepak1556:
Build needs to be fixed.

nornagon:
Based on patch conflicts, I don't think this backport is needed.

Issue #22513, opened 2020-03-03
Navigating away from a page (possibly with an iframe) crashes Electron


pfrazee:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 9.0.0-beta.5
* **Operating System:**
  * MacOS 10.15.3
* **Last Known Working Electron version:**
  * Unknown

### Expected Behavior

Navigating to a page (afaict when it has an iframe) and then navigating away via "Back" should not crash Electron.

### Actual Behavior

Electron crashes with the following trace (with debug symbols):

<details>
```
Received signal 11 SEGV_MAPERR 058c0000057c
0   Electron Framework                  0x0000000117ad7739 base::debug::CollectStackTrace(void**, unsigned long) + 9
1   Electron Framework                  0x00000001179a8df3 base::debug::StackTrace::StackTrace() + 19
2   Electron Framework                  0x0000000117ad75e1 base::debug::(anonymous namespace)::StackDumpSignalHandler(int, __siginfo*, void*) + 2385
3   libsystem_platform.dylib            0x00007fff652b242d _sigtramp + 29
4   ???                                 0x00007f9d56173880 0x0 + 140313730955392
5   Electron Framework                  0x0000000117231c1c network_hints::mojom::NetworkHintsHandlerStubDispatch::Accept(network_hints::mojom::NetworkHintsHandler*, mojo::Message*) + 1148
6   Electron Framework                  0x0000000117ebc45f mojo::InterfaceEndpointClient::HandleValidatedMessage(mojo::Message*) + 991
7   Electron Framework                  0x0000000117ec09fb mojo::MessageDispatcher::Accept(mojo::Message*) + 283
8   Electron Framework                  0x0000000117ebdb08 mojo::InterfaceEndpointClient::HandleIncomingMessage(mojo::Message*) + 120
[...]

PR #23039, opened 2020-04-08
fix: reset node env earlier during shutdown

nornagon:
#### Description of Change
Speculative fix for #22835

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed a crash that could happen when attempting to access a session during shutdown.

### DIFF
--- a/shell/browser/electron_browser_main_parts.cc
+++ b/shell/browser/electron_browser_main_parts.cc
@@ -495,6 +495,7 @@ void ElectronBrowserMainParts::PostMainMessageLoopRun() {
   // Destroy node platform after all destructors_ are executed, as they may
   // invoke Node/V8 APIs inside them.
   node_debugger_->Stop();
+  node_env_.reset();
   js_env_->OnMessageLoopDestroying();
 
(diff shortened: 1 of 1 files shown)


codebytere:
@nornagon that multiplex failure can be pretty consistently repro'd with https://github.com/electron/electron/issues/22513, and i tried building with this fix and sadly still experienced the crash

nornagon:
> It would be better to see if we can disable running any user script before `PostMainMessageLoopRun`, like adding a `blink::ScriptForbiddenScope` to our gin bindings that call into c++ from v8 ?

do you mean `after`?

deepak1556:
Yeah sorry meant after, something like

```
void ElectronBrowserMainParts::PostMainMessageLoopRun() {
   ScriptForbiddenScope forbid_script;
}
```

and add check in the bindings for `ScriptForbiddenScope::IsForbidden()`

nornagon:
Agreed that would be helpful for catching this sort of thing ahead of time in future.

Also agreed that this isn't necessarily early enough, but it's a start!

PR #23041, opened 2020-04-08
chore: cherry-pick e246871765f5 from chromium

nornagon:
Worker: Stop passing creator's origin for starting a dedicated worker

This CL makes DedicatedWorkerHostFactoryImpl use its
|parent_execution_origin_| (renamed to |creator_origin| by this CL) for
starting a dedicated worker instead of an origin passed from a renderer
process.

This was not feasible before because |parent_execution_origin_| is
provided from parent's |RenderFrameHostImpl::last_committed_origin_|
that is set during navigation commit. Worker creation IPC from the
renderer to browser could race with navigation commit, and could see the
wrong last committed origin.

Now this is feasible. This is because worker creation IPC is now tied
with RenderFrameHostImpl's BrowserInterfaceBroker that is re-bound
during navigation commit[*]. This ensures that worker creation requests
issued before the navigation commit are discarded by the previous
BrowserInterfaceBroker, and new requests via the new
BrowserInterfaceBroker are scoped to the new last committed origin.

[*] The call path between binding BrowserInterfaceBroker and updating
the last committed origin is as follows. These are synchronously done.

- RenderFrameHostImpl::DidCommitNavigation() re-binds the interface broker
https://source.chromium.org/chromium/chromium/src/+/master:content/browser/frame_host/render_frame_host_impl.cc;l=7489;drc=d54ee0c3d25dfc644282b50c5f57e23b7ab4dda4?originalUrl=https:%2F%2Fcs.chromium.org%2F
  -> RenderFrameHostImpl::DidCommitNavigationInternal()
    -> NavigatorImpl::DidNavigate()
      -> RenderFrameHostImpl::DidNavigate()
        -> RenderFrameHostImpl::SetLastCommittedOrigin()

Change-Id: Id69c3d66e50aa8cbb7fee520a1479b28970de1c6
Bug: 906991, 1030909
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/1971660
Reviewed-by: Daniel Cheng <dcheng@chromium.org>
Reviewed-by: Matt Falkenhagen <falken@chromium.org>
Commit-Queue: Hiroki Nakagawa <nhiroki@chromium.org>
Cr-Commit-Position: refs/heads/master@{#748127}

Notes: Security: backported fix for site isolation bypass in dedicated workers.

### DIFF
--- a/patches/common/chromium/.patches
+++ b/patches/common/chromium/.patches
@@ -104,3 +104,4 @@ audiocontext_haspendingactivity_unless_it_s_closed.patch
 protect_automatic_pull_handlers_with_mutex.patch
 cherry-pick-adc8f05aa3ab.patch
 mojovideoencodeacceleratorservice_handle_potential_later.patch
+cherry-pick-e246871765f5.patch
--- a/patches/common/chromium/cherry-pick-e246871765f5.patch
+++ b/patches/common/chromium/cherry-pick-e246871765f5.patch
@@ -1,0 +1,519 @@
+From e246871765f5fb2dbe0f9b05726f63745dee03fc Mon Sep 17 00:00:00 2001
+From: Hiroki Nakagawa <nhiroki@chromium.org>
+Date: Mon, 9 Mar 2020 07:07:57 +0000
+Subject: [PATCH] Worker: Stop passing creator's origin for starting a
+ dedicated worker
+
+This CL makes DedicatedWorkerHostFactoryImpl use its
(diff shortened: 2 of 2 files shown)


nornagon:
This is a significant amount of manual work to backport for a fairly obscure fix in an old branch, so I'm making the decision to close this PR and skip backporting this fix for 6-1-x. I recommend upgrading to at least 7.x!

PR #23044, opened 2020-04-08
fix: backport b52f7fb5933a from WebRTC.

ppontes:
#### Description of Change

[DirectX] Fix vector allocation for raw data handling.

std::vector::reserve has the effect to reserve space in memory but does
not affect the result of size(), which keeps on returning 0. If size is
0, however, data() might either return null or not [1].

This CL fixes the use of reserve() in favour of resize() which
effectively allocates the memory in the vector and updates its size.
This way size() returns a value bigger than 0 and data() returns a valid
pointer.

[1] https://en.cppreference.com/w/cpp/container/vector/data

Fixed: chromium:1059764
Change-Id: Ida3dbe643710c6895f09b9da87b0075b7d7b28df
Reviewed-on: https://webrtc-review.googlesource.com/c/src/+/170470
Reviewed-by: Jamie Walch <jamiewalch@chromium.org>
Commit-Queue: Armando Miraglia <armax@webrtc.org>
Cr-Commit-Position: refs/heads/master@{#30836}

#### Release Notes

Notes: Security: backported the fix to CVE-2020-6452: potential container-overflow in MediaStream mojo.

### DIFF
--- a/patches/config.json
+++ b/patches/config.json
@@ -13,5 +13,7 @@
 
   "src/electron/patches/v8":  "src/v8",
 
-  "src/electron/patches/node": "src/third_party/electron_node"
+  "src/electron/patches/node": "src/third_party/electron_node",
+
+  "src/electron/patches/webrtc": "src/third_party/webrtc"
 }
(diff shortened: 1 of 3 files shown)


jkleinsc:
Merging as CI failure is known flake unrelated to this PR.

PR #23049, opened 2020-04-09
chore: cherry-pick 85f708fa7ab8 from chromium

nornagon:
Use KeepSelfAlive on AudioContext to keep it alive until rendering stops

When an ExecutionContext is abruptly/unexpectedly destroyed (e.g.
shutting down of document or iframe), an AudioContext can also
go away. This type of shutdown can be problematic because the render
thread still might be touching resources in the AudioContext allocated
by the main thread.

This CL introduces a self-referencing pointer to the AudioContext,
and it is cleared after the underlying render thread is stopped. In
that way, the destruction of AudioContext can be done safely.

Test: Locally confirmed the repro case doesn't crash (UAP) after 1hr.
Bug: 1043446
Change-Id: I2e40b7d58ca9d647eed8a5971fc69dc87ee3d1fe
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2049912
Reviewed-by: Michael Lippautz <mlippautz@chromium.org>
Cr-Commit-Position: refs/heads/master@{#742338}

Notes: Security: backported fix for CVE-2020-6423: Use after free in audio.


nornagon:
Closing this as the code seems significantly different in M76 from what it is on M80 where this fix was applied.

PR #23054, opened 2020-04-09
feat: separate TopLevelWindow from BrowserWindow

zcbenz:
#### Description of Change

Makes `TopLevelWindow` class public and make `BrowserWindow` class inherit from it, as discussed in https://github.com/electron/governance/pull/254. Note that this is almost a documentation-only change, the actually implementation had been in Electron for quite a long time.

This PR mostly separates the documentation of `BrowserWindow` into different classes/structures, so while it is a large PR it does not include significant changes. I plan to add documentations on how `BrowserWindow` is different from `TopLevelWindow`, and I'll do it in other PRs so they can be reviewed easily.

This PR does:
* Move parts of `BrowserWindow` APIs into `TopLevelWindow` class.
* Add `BrowserWindowOptions`, `TopLevelWindowOptions`, `WebPreferences` structures, and `BrowserWindowOptions` inherits from `TopLevelWindowOptions`.
* Move notes about `beforeunload` from the description of `close` event into a new chapter in `BrowserWindow`.
* In documentations of `BrowserWindow`, reference the instance object as `browserWindow`, to make a different from `win` used by `TopLevelWindow`.

#### Release Notes

Notes: Add `TopLevelWindow` class.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -133,7 +133,7 @@ Process: [Main](../glossary.md#main-process)
 
 `BrowserWindow` is an [EventEmitter][event-emitter].
 
-It creates a new `BrowserWindow` with native properties as set by the `options`.
+`BrowserWindow` inherits from [`TopLevelWindow`](./top-level-window.md).
 
 ### `new BrowserWindow([options])`
 
@@ -385,6 +385,8 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       visible to users.
     * `spellcheck` Boolean (optional) - Whether to enable the builtin spellchecker.
       Default is `true`.
+
+Creates a new `BrowserWindow` with native properties as set by the `options`.
 
 When setting minimum or maximum window size with `minWidth`/`maxWidth`/
 `minHeight`/`maxHeight`, it only constrains the users. It won't prevent you from
--- a/docs/api/top-level-window.md
+++ b/docs/api/top-level-window.md
@@ -1,0 +1,730 @@
+# TopLevelWindow
+
+> Create and control native windows.
+
+Process: [Main](../glossary.md#main-process)
+
+```javascript
+// In the main process.
+const { TopLevelWindow } = require('electron')
+const win = new TopLevelWindow({ width: 800, height: 600 })
+```
+
+## Class: TopLevelWindow
+
+> Create and control native windows.
+
+Process: [Main](../glossary.md#main-process)
+
+`TopLevelWindow` is an [EventEmitter][event-emitter].
+
+### `new TopLevelWindow(options)` _Experimental_
+
+* `options` Omit<BrowserWindowConstructorOptions, 'webPreferences'>
+
+Creates a new `TopLevelWindow` with native properties as set by the `options`.
+
+### Instance Events
+
+Objects created with `new TopLevelWindow` emit the following events:
+
+#### Event: 'close'
+
+Refs ['close' event of BrowserWindow](./browser-window.md#event-close).
+
+#### Event: 'closed'
+
+Refs ['closed' event of BrowserWindow](./browser-window.md#event-closed).
+
+#### Event: 'session-end' _Windows_
+
+Refs ['session-end' event of BrowserWindow](./browser-window.md#event-session-end).
+
+#### Event: 'blur'
+
+Refs ['blur' event of BrowserWindow](./browser-window.md#event-blur).
+
+#### Event: 'focus'
+
+Refs ['focus' event of BrowserWindow](./browser-window.md#event-focus).
+
+#### Event: 'show'
+
+Refs ['show' event of BrowserWindow](./browser-window.md#event-show).
+
+#### Event: 'hide'
+
+Refs ['hide' event of BrowserWindow](./browser-window.md#event-hide).
+
+#### Event: 'maximize'
+
+Refs ['maximize' event of BrowserWindow](./browser-window.md#event-maximize).
+
+#### Event: 'unmaximize'
+
+Refs ['unmaximize' event of BrowserWindow](./browser-window.md#event-unmaximize).
+
+#### Event: 'minimize'
+
+Refs ['minimize' event of BrowserWindow](./browser-window.md#event-minimize).
+
+#### Event: 'restore'
+
+Refs ['restore' event of BrowserWindow](./browser-window.md#event-restore).
+
+#### Event: 'will-resize' _macOS_ _Windows_
+
+Refs ['will-resize' event of BrowserWindow](./browser-window.md#event-will-resize).
+
+#### Event: 'resize'
+
+Refs ['resize' event of BrowserWindow](./browser-window.md#event-resize).
+
+#### Event: 'will-move' _macOS_ _Windows_
+
+Refs ['will-move' event of BrowserWindow](./browser-window.md#event-will-move).
+
+#### Event: 'move'
+
+Refs ['move' event of BrowserWindow](./browser-window.md#event-move).
+
+#### Event: 'moved' _macOS_
+
+Refs ['moved' event of BrowserWindow](./browser-window.md#event-moved).
+
+#### Event: 'enter-full-screen'
+
+Refs ['enter-full-screen' event of BrowserWindow](./browser-window.md#event-enter-full-screen).
+
+#### Event: 'leave-full-screen'
+
+Refs ['leave-full-screen' event of BrowserWindow](./browser-window.md#event-leave-full-screen).
+
+#### Event: 'always-on-top-changed'
+
+Refs ['always-on-top-changed' event of BrowserWindow](./browser-window.md#event-always-on-top-changed).
+
+#### Event: 'app-command' _Windows_ _Linux_
+
+Refs ['app-command' event of BrowserWindow](./browser-window.md#event-app-command).
+
+#### Event: 'scroll-touch-begin' _macOS_
+
+Refs ['scroll-touch-begin' event of BrowserWindow](./browser-window.md#event-scroll-touch-begin).
+
+#### Event: 'scroll-touch-end' _macOS_
+
+Refs ['scroll-touch-end' event of BrowserWindow](./browser-window.md#event-scroll-touch-end).
+
+#### Event: 'swipe' _macOS_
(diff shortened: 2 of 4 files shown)


default_app/default_app.ts
```diff
@@ -44,7 +44,7 @@ ipcMain.handle('bootstrap', (event) => {
 async function createWindow () {
   await app.whenReady();
 
-  const options: Electron.BrowserWindowConstructorOptions = {
+  const options: Electron.BrowserWindowOptions = {
```
malept, line 47: Changing from `Electron.BrowserWindowConstructorOptions` to `Electron.BrowserWindowOptions` is a breaking change for folks who use TypeScript. Can you please document this in `docs/breaking-changes.md` ?

nornagon:
API nit: `TopLevelWindow` isn't my favourite name for this. It implies that there are other kinds of windows that are not top-level, which there aren't.

Some other possibilities:
- `Window`, keep it simple. Though this would perhaps clash with [`Window` from the web](https://developer.mozilla.org/en-US/docs/Web/API/Window)? though we don't have blink in the main process, so it might be fine.
- `Widget`, to match the `//views` terminology
- `NativeWindow`

docs/api/browser-window.md
```diff
@@ -116,302 +84,46 @@ The [Page Visibility API][page-visibility-api] works as follows:
...
+  // Unlike usual browsers that a message box will be prompted to users,
+  // returning a non-void value will silently cancel the close. It is
+  // recommended to use the dialog API to let the user confirm closing the
+  // application.
+  e.returnValue = false // equivalent to `return false` but not recommended
+}
+```
+_**Note**: There is a subtle difference between the behaviors of `window.onbeforeunload = handler` and `window.addEventListener('beforeunload', handler)`. It is recommended to always set the `event.returnValue` explicitly, instead of only returning a value, as the former works more consistently within Electron._
+
+## Class: BrowserWindow extends `TopLevelWindow`
```
nornagon, line 110: It's sort of strange that TopLevelWindow gets backticks here but BrowserWindow doesn't

docs/api/browser-window.md
```diff
@@ -116,302 +84,46 @@ The [Page Visibility API][page-visibility-api] works as follows:
...
-  `notification`.
-* On macOS, possible types are `desktop`, `textured`.
-  * The `textured` type adds metal gradient appearance
-    (`NSTexturedBackgroundWindowMask`).
-  * The `desktop` type places the window at the desktop background window level
-    (`kCGDesktopWindowLevel - 1`). Note that desktop window will not receive
-    focus, keyboard or mouse events, but you can use `globalShortcut` to receive
-    input sparingly.
-* On Windows, possible type is `toolbar`.
+* `options` [BrowserWindowOptions](structures/browser-window-options.md) (optional)
```
nornagon, line 118: I think this might be one of the more heavily-used parts of Electron's documentation (@electron/wg-ecosystem perhaps has some stats?), so moving it is a big change. Also, it moves the `WebPreferences` documentation behind two separate clicks (first through to `BrowserWindowOptions` and then through to `WebPreferences`), where before they were directly available here.

I'm not necessarily against this reorganization, I just wanted to point out that it's a fairly major docs change and have some discussion about it.

docs/api/browser-window.md
```diff
@@ -116,302 +84,46 @@ The [Page Visibility API][page-visibility-api] works as follows:
...
+
+Creates a new `BrowserWindow`, the `options` will also be passed to its base
+class `TopLevelWindow`.
 
 ### Instance Events
 
 Objects created with `new BrowserWindow` emit the following events:
```
nornagon, line 126: Should we link to the documentation for the parent class's events here? It seems like it might be easy to miss the fact that `BrowserWindow` has more events than are listed here if you come straight to this part of the documentation.

docs/api/browser-window.md
```diff
@@ -774,1044 +291,102 @@ Objects created with `new BrowserWindow` have the following properties:
...
 
-#### `win.destroy()`
-
-Force closing the window, the `unload` and `beforeunload` event won't be emitted
-for the web page, and `close` event will also not be emitted
-for this window, but it guarantees the `closed` event will be emitted.
+#### `browserWindow.focusOnWebView()`
 
-#### `win.close()`
+#### `browserWindow.blurWebView()`
```
nornagon, line 315: Are these new methods? They're lacking documentation.

docs/api/top-level-window.md
```diff
@@ -0,0 +1,1240 @@
+# TopLevelWindow
+
+> Create and control native windows.
+
+Process: [Main](../glossary.md#main-process)
```
nornagon, line 5: We should probably mention `BrowserWindow` here as the main way to create windows with web content.

docs/api/top-level-window.md
```diff
@@ -0,0 +1,1240 @@
...
+  when parent window moves, while on Windows and Linux child windows will not
+  move.
+* On Linux the type of modal windows will be changed to `dialog`.
+* On Linux many desktop environments do not support hiding a modal window.
+
+## Class: TopLevelWindow
+
+> Create and control native windows.
+
+Process: [Main](../glossary.md#main-process)
```
nornagon, line 50: It's odd that this information is repeated..?

docs/breaking-changes.md
```diff
@@ -14,6 +14,11 @@ This document uses the following convention to categorize breaking changes:
 
 ## Planned Breaking API Changes (10.0)
 
+### API Changed: `BrowserWindowConstructorOptions` is renamed to `BrowserWindowOptions`
+
+For TypeScript users, the type name of `BrowserWindow`'s options has changed
+form `BrowserWindowConstructorOptions` to `BrowserWindowOptions`.
```
nornagon, line 20: We don't typically list type changes as breaking changes to the API. I think we should continue that tradition for a few reasons:

1. We fairly regularly make relatively minor changes to types (e.g. refining types from `any` to a more specific type, or fixing incorrect types). Ensuring we including all such changes in the breaking changes notes would be a fairly significant undertaking and would result in a lot of new breaking change notes.
2. The TS compiler will automatically catch these changes and notify folks who are upgrading anyway.

That said, I could see an argument for having a single item under the "breaking API changes" for "here's what types changed: ..."

docs/api/top-level-window.md
```diff
@@ -0,0 +1,1240 @@
...
+
+`TopLevelWindow` is an [EventEmitter][event-emitter].
+
+### `new TopLevelWindow([options])`
```
nornagon, line 54: This constructor should be marked experimental, with a note that creating a TopLevelWindow directly (as opposed to via BrowserWindow) is experimental.

docs/api/browser-window.md
erickzhao: > I think this might be one of the more heavily-used parts of Electron's documentation

Good point! Google Analytics says it's the 5th most viewed page in the past 90 days, and within `/docs`, it's the 3rd one beyond the root docs page and the first app tutorial.

zcbenz:
> API nit: `TopLevelWindow` isn't my favourite name for this. It implies that there are other kinds of windows that are not top-level, which there aren't.

We used to use `Window` as name for browser windows, but it was too confusing to have both DOM `Window` and OS `Window`, so we ended up renaming it to `BrowserWindow`. Using `Window` as type name would make the term context sensitive, we have to read the context to know which window we are talking about. And it would make learning Electron harder, as new comers have to know the differences between 2 `Window`s.

The name `Widget`, is used to reference UI elements (controls/views) in GTK and Qt, I think it would be a strange term to reference windows.

`NativeWindow` sounds parallel to `BrowserWindow`, their relationship feels more like `NativeWindow` vs `BrowserWindow`, instead of `BrowserWindow` inheriting from `NativeWindow`.

The good thing about `TopLevelWindow` is, this name has no ambiguity. The term `Window` could be referenced as UI elements in some UI toolkit. In Win32 and X11 controls are windows, in GTK some widgets are windows (GDK window), in Qt any widget without a parent is a window. The term `TopLevelWindow` is dumb but people usually would not mistake it for other things.

But I agree `TopLevelWindow` is of course not the best name, and I'm open to renaming it to better names.

docs/api/browser-window.md
zcbenz: This follows the style of structures, which use backticks to reference base objects (for example `MouseInputEvent`), I don't know why backticks was used there. /cc @MarshallOfSound

docs/api/browser-window.md
zcbenz: I moved `BrowserWindowOptions` to a separate place because it now inherits from the `TopLevelWindowOptions`.

The `WebPreferences` is moved to structures because the new `WebContentsView` class will take it as parameter.

I don't know what's the best way to handle this. Maybe we can add a button on docs website to fill the information? (Sorry for pushing away the problem 😄 )

docs/api/browser-window.md
zcbenz: Sounds good to me, what would be the best way to do this? /cc @electron/wg-ecosystem

docs/api/browser-window.md
zcbenz: They are not new, they did not have description in the original documentation. I can add descriptions in new PRs.

docs/api/top-level-window.md
zcbenz: The original documentation does so, I think it is because one is for the module and one is for the class? I don't know if the docs parser needs 2 of this. @MarshallOfSound

docs/breaking-changes.md
zcbenz: This was requested by @malept, I don't have a strong feeling for documenting this.

docs/api/top-level-window.md
zcbenz: I'll add detailed information for `TopLevelWindow` and its relationship with `BrowserWindow` in other PRs, so the reviews can be done easier.

docs/breaking-changes.md
malept: I've added this topic to the Ecosystem WG meeting agenda.

loc:
drive-by suggestion on the question of naming: maybe `BaseWindow`?

PR #23076, opened 2020-04-10
chore: remove patch: fix_breakpad_symbol_generation_on_linux_arm

nornagon:
The patch description doesn't say what the error was, so let's try removing it to find out!

Notes: none


nornagon:
@jkleinsc we might need release builds on this one

codebytere:
@nornagon it was a release build error - we'll need to run release builds on this

i'll do it monday!

jkleinsc:
@nornagon @codebytere I kicked off release builds on this one.

nornagon:
Looks to me like the linux arm release build succeeded, but I'm not sure what the "Build Error" build is.

jkleinsc:
`Build Error` is from an incorrectly kicked off release build.  It can safely be ignored.

jkleinsc:
Merging as CI failures are unrelated to PR change.

PR #23096, opened 2020-04-13
feat: add new render-process-gone event

MarshallOfSound:
The `crashed` event could fire for a number of reasons, some of which are nothing to do with crashes.  It also was losing some helpful information about **why** the process crashed.  This PR leaves that event untouched (as per our new breaking changes policy) and adds a new event that contains more information (the actual `reason` the process crashed) and puts it in an object for easier expandability later.

Notes: Added new `render-process-gone` event to replace the `crashed` event.

### DIFF
--- a/shell/browser/api/electron_api_web_contents.h
+++ b/shell/browser/api/electron_api_web_contents.h
@@ -60,6 +60,42 @@ namespace network {
 class ResourceRequestBody;
 }
 
+namespace gin {
+
+template <>
+struct Converter<base::TerminationStatus> {
+  static v8::Local<v8::Value> ToV8(v8::Isolate* isolate,
+                                   const base::TerminationStatus& status) {
+    switch (status) {
+      case base::TERMINATION_STATUS_NORMAL_TERMINATION:
+        return gin::ConvertToV8(isolate, "clean-exit");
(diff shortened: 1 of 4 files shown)


shell/browser/api/electron_api_web_contents.h
```diff
@@ -60,6 +60,41 @@ namespace network {
...
+        return gin::ConvertToV8(isolate, "oom");
+#if defined(OS_WIN)
+      case base::TERMINATION_STATUS_INTEGRITY_FAILURE:
+        return gin::ConvertToV8(isolate, "integrity-failure");
+#endif
+      case base::TERMINATION_STATUS_MAX_ENUM:
+        DCHECK(false);
+        return gin::ConvertToV8(isolate, "");
+    }
+    return gin::ConvertToV8(isolate, "");
```
nornagon, line 92: also DCHECK(false) here maybe?
```suggestion
    NOTREACHED();
    return gin::ConvertToV8(isolate, "");
```

pushkin-:
@MarshallOfSound  Can I safely access the `webContents`'s id even though I would expect the object to be destroyed? It seems like I can without any crashes, but is that guaranteed?

pushkin-:
Maybe [this PR](https://github.com/electron/electron/pull/27038) addresses that concern

PR #23100, opened 2020-04-13
chore: rename atom things to electron things in our docs / scripts

MarshallOfSound:
Finishing up the removal of `atom` from our source.  Some of the documentation was outdated / wrong so I fixed that too.

Notes: none

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -139,32 +139,7 @@ When using Electron's built-in module you might encounter an error like this:
 Uncaught TypeError: Cannot read property 'setZoomLevel' of undefined
 ```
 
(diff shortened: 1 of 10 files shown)


docs/faq.md
```diff
@@ -139,32 +139,7 @@ When using Electron's built-in module you might encounter an error like this:
 Uncaught TypeError: Cannot read property 'setZoomLevel' of undefined
 ```
 
-This is because you have the [npm `electron` module][electron-module] installed
-either locally or globally, which overrides Electron's built-in module.
-
-To verify whether you are using the correct built-in module, you can print the
```
MarshallOfSound, line 145: This is no longer possible as we more directly override nodes module resolution

PR #23103, opened 2020-04-14
build: check patches on update

codebytere:
#### Description of Change

This PR ensures that patch indices are updated when patches are changed. It will try to commit as the current user, and if none is found will fall back on `Electron Bot`. This change will only be run when patches are changed, as a precommit hook via `lint-staged`.

cc @MarshallOfSound @nornagon 

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/script/check-patch-diff.js
+++ b/script/check-patch-diff.js
@@ -1,0 +1,18 @@
+const { spawnSync } = require('child_process');
+const path = require('path');
+const { inspect } = require('util');
+
+const srcPath = path.resolve(__dirname, '..', '..', '..');
+const patchExportFnPath = path.resolve(__dirname, 'export_all_patches.py');
+const configPath = path.resolve(__dirname, '..', 'patches', 'config.json');
+
+// Re-export all the patches to check if there were changes.
+const proc = spawnSync('python', [patchExportFnPath, configPath, '--dry-run'], {
+  cwd: srcPath
+});
+
+// Fail if patch exporting returned 1, e.g dry run failed.
+if (proc.status === 1) {
+  console.log(proc.stderr.toString('utf8'));
+  process.exit(1);
+}
--- a/package.json
+++ b/package.json
@@ -133,7 +133,8 @@
       "git add filenames.auto.gni"
     ],
     "{*.patch,.patches}": [
-      "node script/lint.js --patches --only --"
+      "node script/lint.js --patches --only --",
+      "node script/check-patch-diff.js"
     ],
     "DEPS": [
       "node script/gen-hunspell-filenames.js"
--- a/script/export_all_patches.py
+++ b/script/export_all_patches.py
@@ -8,9 +8,9 @@ import sys
 from lib import git
 
 
-def export_patches(dirs):
+def export_patches(dirs, dry_run):
   for patch_dir, repo in dirs.iteritems():
-    git.export_patches(repo=repo, out_dir=patch_dir)
+    git.export_patches(repo=repo, out_dir=patch_dir, dry_run=dry_run)
 
 
 def parse_args():
@@ -18,13 +18,17 @@ def parse_args():
   parser.add_argument('config', nargs='+',
                       type=argparse.FileType('r'),
                       help='patches\' config(s) in the JSON format')
+  parser.add_argument("-d", "--dry-run",
+    help="Checks whether the exported patches need to be updated.",
+    default=False, action='store_true')
   return parser.parse_args()
 
 
 def main():
   configs = parse_args().config
+  dry_run = parse_args().dry_run
   for config_json in configs:
-    export_patches(json.load(config_json))
+    export_patches(json.load(config_json), dry_run)
 
 
 if __name__ == '__main__':
--- a/script/lib/git.py
+++ b/script/lib/git.py
@@ -260,7 +260,7 @@ def remove_patch_filename(patch):
     force_keep_next_line = l.startswith('Subject: ')
 
 
-def export_patches(repo, out_dir, patch_range=None):
+def export_patches(repo, out_dir, patch_range=None, dry_run=False):
   if patch_range is None:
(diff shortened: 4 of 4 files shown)


script/check-patch-diff.js
```diff
@@ -0,0 +1,40 @@
+const { execSync, spawnSync } = require('child_process');
+const path = require('path');
+const { inspect } = require('util');
+
+const srcPath = path.resolve(__dirname, '..', '..', '..');
+const patchExportPath = path.resolve(__dirname, 'export_all_patches.py');
+const configPath = path.resolve(__dirname, '..', 'patches', 'config.json');
+
+// Re-export all the patches to check if there were changes.
+execSync(`python ${patchExportPath} ${configPath}`, { cwd: srcPath });
```
nornagon, line 10: don't rely on shell quoting for this. as written this will fail if `__dirname` has a space in it, for example.
```suggestion
spawnSync('python', [patchExportPath, configPath], { cwd: srcPath });
```

script/check-patch-diff.js
```diff
@@ -0,0 +1,40 @@
...
+
+// Update indices.
+execSync('git update-index --refresh || true');
```
nornagon, line 13: here too, use `spawnSync` and then you can ignore the exit code by just ... ignoring it :)

script/check-patch-diff.js
```diff
@@ -0,0 +1,40 @@
...
+const { status } = spawnSync('git', ['diff-index', '--quiet', 'HEAD']);
+
+if (status !== 0) {
+  const committerName = execSync('git config user.name');
+  const committerEmail = execSync('git config user.email');
+
+  const env = Object.assign({}, process.env, {
+    GIT_COMMITTER_NAME: committerName || 'Electron Bot',
+    GIT_COMMITTER_EMAIL: committerEmail || 'anonymous@electronjs.org'
+  });
```
nornagon, line 25: I don't think there's any need for this bit here as it's running on a user's local machine, prior to commit. So we don't need to make a new commit inside of the lint step.

script/check-patch-diff.js
```diff
@@ -0,0 +1,40 @@
...
+  // Commit updated patches.
+  const { status } = spawnSync('git', ['commit', '-m', 'chore: update patch indices'], {
+    env,
+    stdio: 'inherit'
+  });
+
+  if (status !== 0) {
+    console.error('Failed to update patches.');
+    process.exit(1);
+  }
```
nornagon, line 39: I think rather rely on the lint step to run `git add`, by analogy with the other linters we have that edit code files.

script/check-patch-diff.js
```diff
@@ -0,0 +1,16 @@
+const { spawnSync } = require('child_process');
+const path = require('path');
+const { inspect } = require('util');
+
+const srcPath = path.resolve(__dirname, '..', '..', '..');
+const patchExportPath = path.resolve(__dirname, 'export_all_patches.py');
```
nornagon, line 6: One potential gotcha I can imagine happening here is if someone manually adds a patch to `patches/` without going through the `export-patches` flow—this would delete that patch. Then the `git add` would commit the deletion to the index.

It seems a bit rude to delete the file that someone just added. Perhaps we could skip the `git add` and just fail the lint step if there are changes as a result of `export_all_patches`?

codebytere: oh - good point 🤔 i agree - we maybe shouldn't be nuking things out from underneath users with no warning, but on the other hand i guess it does force use of what arguably should be done anyway? Lint-staged typically doesn't fail that way, but i'm definitely open to it

MarshallOfSound: >  and just fail the lint step if there are changes as a result of export_all_patches?

Yeah I think we should do this, let users figure out what to do with their _bad things_, would suck to lose a change because you didn't sync or run export or something

script/check-patch-diff.js
```diff
@@ -0,0 +1,18 @@
...
+// Re-export all the patches to check if there were changes.
+const proc = spawnSync('python', [patchExportFnPath, configPath, '--dry-run'], {
+  cwd: srcPath
+});
+
+// Fail if patch exporting returned 1, e.g dry run failed.
+if (proc.status === 1) {
+  console.log(proc.stderr.toString('utf8'));
+  process.exit(1);
+}
```
nornagon, line 18: can this be done by just `python export_all_patches --dry-run` from the package.json list? or is the cwd not stable for that script?

codebytere: @nornagon unfortunately no - i tried that initially to golf down changes by just putting `python script/export_all_patches.py patches/config.json`  in the lint-staged script and it's not stable:
```
fatal: cannot change to 'src': No such file or directory
fatal: cannot change to 'src': No such file or directory
Traceback (most recent call last):
File "script/export_all_patches.py", line 35, in <module>
  main()
File "script/export_all_patches.py", line 31, in main
  export_patches(json.load(config_json), dry_run)
File "script/export_all_patches.py", line 13, in export_patches
  git.export_patches(repo=repo, out_dir=patch_dir, dry_run=dry_run)
File "/Users/codebytere/Developer/electron-gn/src/electron/script/lib/git.py", line 265, in export_patches
  patch_range, num_patches = guess_base_commit(repo)
File "/Users/codebytere/Developer/electron-gn/src/electron/script/lib/git.py", line 184, in guess_base_commit
  return subprocess.check_output(args).rsplit('-', 2)[0:2]
File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 223, in check_output
  raise CalledProcessError(retcode, cmd, output=output)
subprocess.CalledProcessError: Command '['git', '-C', u'src', 'describe', '--tags']' returned non-zero exit status 128
husky > pre-commit hook failed (add --no-verify to bypass)
```

PR #23111, opened 2020-04-14
fix: allow blocking IO for setAppLogsPath to avoid DCHECK

loc:
#### Description of Change
Avoid a DCHECK that Slack hits on startup.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


shell/browser/api/electron_api_app.cc
```diff
@@ -846,6 +846,7 @@ void App::SetAppPath(const base::FilePath& app_path) {
 #if !defined(OS_MACOSX)
 void App::SetAppLogsPath(gin_helper::ErrorThrower thrower,
                          base::Optional<base::FilePath> custom_path) {
+  base::ThreadRestrictions::ScopedAllowIO allow_io;
```
nornagon, line 849: Hm, it's unclear to me which bit of this function does IO. Can we wrap this tighter by any chance? Put it closer to where the IO happens?

shell/browser/net/proxying_url_loader_factory.cc
```diff
@@ -810,6 +810,7 @@ void ProxyingURLLoaderFactory::CreateLoaderAndStart(
   network::ResourceRequest request = original_request;
 
   if (ShouldIgnoreConnectionsLimit(request)) {
+    request.priority = net::RequestPriority::MAXIMUM_PRIORITY;
```
nornagon, line 813: This seems unrelated to the PR title, is it intentionally here?

shell/browser/api/electron_api_app.cc
loc: base::PathService::Override checks if the directory exists (and thus hits the filesystem). I suppose i can wrap the call in a block and scope it there? but since it happens in both branches, that seemed excessive.

shell/browser/net/proxying_url_loader_factory.cc
loc: I can move this into its own PR.

shell/browser/api/electron_api_app.cc
nornagon: It's best to scope this as narrowly as possible IMO, both because it reduces the likelihood we'll accidentally do some IO where we shouldn't be, and because it helps self-document (i.e. i wouldn't have asked this question if it was wrapped in a block with only one call to `PathService::Override`!)

loc: Good point! Fixed.

PR #23112, opened 2020-04-14
ci: fix v8 cppgc files in wrong location

codebytere:
#### Description of Change

Refs [2139585](https://chromium-review.googlesource.com/c/v8/v8/+/2139585).

V8 moved a few files used by Node.js to another directory in the most recent Chromium roll, which clashed with the changes made in the latest Node.js roll. This fixes those incorrect locations.

cc @nornagon @MarshallOfSound @jkleinsc 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


patches/node/build_add_gn_build_files.patch
```diff
@@ -944,7 +944,25 @@ index 0000000000000000000000000000000000000000..78dff9e0458aaa6b864259f3001764c7
...
++        "//v8/include/cppgc/source-location.h",
++        "//v8/include/cppgc/trace-trait.h",
++        "//v8/include/cppgc/type-traits.h",
++        "//v8/include/cppgc/visitor.h"
++      ]
++    },
++    {
++      "dest_dir": "include/node//cppgc/internal/",
++      "files": [
++        "//v8/include/cppgc/internal/accessors.h",
```
deepak1556, line 961: Why are we exposing internal headers through `node_headers` ?

deepak1556: The CL commit says,

```
cppgc: Move implementation headers in internal/ dir

Headers containing only implementation details are better to stay in a
separate dir to indicate the user that they shouldn't be included.
```

codebytere: @deepak1556 hmm - i just ran `python tools/generate_gn_filenames_json.py` from the node dir

nornagon: @deepak1556 the non-internal headers include the internal ones. it just means "shouldn't be included _directly_".

deepak1556: Okay in that case there is nothing we can do, I was hoping the non-internal ones are like `v8.h`

PR #23118, opened 2020-04-15
fix: explicitly mark unlimited requests as max priority

loc:
#### Description of Change
Avoid a DCHECK that Slack hits on startup.

Slack ignores connection limits for its domains and [hits this DCHECK](https://source.chromium.org/chromium/chromium/src/+/master:net/url_request/url_request.cc;l=433;drc=0e1e59c69a449885aa58c403e05feb8274a67c3b;bpv=1;bpt=1) when requests aren't initialized with a max priority?

The priority of these requests is reset after the DCHECK for production builds, so this is essentially a no-op other than avoiding the check.
```
if ((load_flags_ & LOAD_IGNORE_LIMITS) != 0)
  SetPriority(MAXIMUM_PRIORITY);
```

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


deepak1556:
Please rebase on master to fix the build.

PR #23128, opened 2020-04-15
refactor: be more precise when creating api::WebContents

nornagon:
#### Description of Change
This removes one usage of `GetWebContentsFromProcessID`, which is important as there can potentially be several `WebContents` relating to a given process ID.

Previously, we were using `RenderProcessReady` to be a "backstop" for catching certain unusual paths for creation of WebContents instances. In particular, the WebContents that are created for devtools, and those that are created for guest views (specifically, mime viewer guest views, i.e. for the pdfium pdf viewer).

This PR shifts the creation of the `api::WebContents` wrapper for devtools to be explicitly created at the same time as the devtools `content::WebContents`, and similarly for extension background pages and mime view handlers (i.e. pdf viewer guest views).

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/electron_api_app.cc
+++ b/shell/browser/api/electron_api_app.cc
@@ -877,17 +877,6 @@ void App::BrowserChildProcessCrashedOrKilled(
 void App::RenderProcessReady(content::RenderProcessHost* host) {
   ChildProcessLaunched(content::PROCESS_TYPE_RENDERER, host->GetID(),
                        host->GetProcess().Handle());
-
-  // TODO(jeremy): this isn't really the right place to be creating
-  // `WebContents` instances, but this was implicitly happening before in
-  // `RenderProcessPreferences`, so this is at least more explicit...
-  content::WebContents* web_contents =
-      ElectronBrowserClient::Get()->GetWebContentsFromProcessID(host->GetID());
-  if (web_contents) {
-    v8::Isolate* isolate = JavascriptEnvironment::GetIsolate();
-    v8::HandleScope scope(isolate);
-    WebContents::FromOrCreate(isolate, web_contents);
-  }
 }
 
 void App::RenderProcessExited(content::RenderProcessHost* host) {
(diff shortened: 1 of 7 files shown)


shell/browser/ui/inspectable_web_contents_impl.cc
```diff
@@ -456,6 +458,10 @@ void InspectableWebContentsImpl::ShowDevTools(bool activate) {
     managed_devtools_web_contents_ = content::WebContents::Create(
         content::WebContents::CreateParams(web_contents_->GetBrowserContext()));
     managed_devtools_web_contents_->SetDelegate(this);
+    v8::Isolate* isolate = v8::Isolate::GetCurrent();
+    v8::HandleScope scope(isolate);
+    api::WebContents::FromOrCreate(isolate,
```
deepak1556, line 463: For devtools, we already create the js wrapper in `api::WebContents::DevToolsOpened`

zcbenz (dismissed):
There are tests failing:

```
Unhandled exception in main spec runner: AssertionError: expected 9 to equal 10
    at Immediate.<anonymous> (/home/builduser/project/src/electron/spec-main/api-app-spec.ts:410:39)
    at processImmediate (internal/timers.js:456:21) {
  message: 'expected 9 to equal 10',
  showDiff: true,
  actual: 9,
  expected: 10
}
not ok 23 app module BrowserWindow events should emit web-contents-created event when a webContents is created
  AssertionError: expected 9 to equal 10
      at Immediate.<anonymous> (electron/spec-main/api-app-spec.ts:410:39)
      at processImmediate (internal/timers.js:456:21)
```

shell/browser/ui/inspectable_web_contents_impl.cc
nornagon: That wrapper is not created early enough to respond to the sync message that is sent when the renderer boots.

MarshallOfSound (approved):

PR #23130, opened 2020-04-15
refactor: improve feature string parsing

loc:
#### Description of Change
Splitting up https://github.com/electron/electron/pull/21236 so it's easier to review and merge.

This piece refactors the feature string parsing employed by window.open. There were a few bugs in the old version (see my notes in the "update snapshots" commit), and a cleaner implementation more readily brings those to light. We also add some compile-time safety with regard to type coercion.

Since this is a rather integral part of window opening, I added a snapshot test to make sure I didn't break anything.

I will actually deprecate additionalFeatures / bare key parsing behavior in a future PR.

#### Bugs revealed:
For example, previously, the string `"show=false,top=1,left=1"` passed to `window.open` would yield an options object
```
{
  show: false,
  top: true,
  left: true,
  x: null,
  y: null,
}
```

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed feature string parsing type coercion for window.open


spec-main/fixtures/snapshots/native-window-open.snapshot.txt
```diff
@@ -13,11 +13,6 @@
       "width": 800,
       "height": 600,
       "webContents": "[WebContents]",
-      "top": "5",
-      "left": "10",
```
loc: notice the previously-incorrect types here

spec-main/fixtures/snapshots/native-window-open.snapshot.txt
```diff
@@ -61,9 +62,8 @@
         "nodeIntegrationInSubFrames": false
       },
       "resizable": false,
-      "x": null,
```
loc: previously incorrect boolean coercion

lib/browser/api/web-contents.js
```diff
@@ -511,11 +511,13 @@ WebContents.prototype._init = function () {
...
     this.on('-new-window', (event, url, frameName, disposition,
       rawFeatures, referrer, postData) => {
-      const { options, additionalFeatures } = convertFeaturesString(rawFeatures, frameName);
+      const { options, webPreferences, additionalFeatures } = parseFeatures(rawFeatures);
       const mergedOptions = {
         show: true,
         width: 800,
         height: 600,
+        title: frameName,
+        webPreferences,
```
loc, line 520: This got slightly more obtuse for now, but it will be improved in a future PR.

lib/browser/guest-window-manager.js
```diff
@@ -224,7 +224,10 @@ ipcMainInternal.on('ELECTRON_GUEST_WINDOW_MANAGER_WINDOW_OPEN', (event, url, fra
   if (features == null) features = '';
 
   const disposition = 'new-window';
-  const { options, additionalFeatures } = convertFeaturesString(features, frameName);
+  const { options, webPreferences, additionalFeatures } = parseFeatures(features);
+  if (!options.title) options.title = frameName;
+  options.webPreferences = webPreferences;
```
loc, line 229: same comment as above; slightly more obtuse now but will make more sense in a later refactoring PR.

lib/browser/api/web-contents.js
```diff
@@ -511,11 +511,13 @@ WebContents.prototype._init = function () {
...
         ...options
```
nornagon, line 521: The code was previously careful to not let a `webPreferences=...` string in the feature string override the `webPreferences` that's built from other things. having `...options` come after `webPreferences` in this list makes me wonder if it's possible for `options` to contain a `webPreferences` key that could cause issues?

lib/common/parse-features-string.ts
```diff
@@ -0,0 +1,115 @@
+/**
+ * Utilities to parse comma-separated key value pairs used in browser APIs.
+ * For example: "x=100,y=200,width=500,height=500"
+ */
+import { BrowserWindowConstructorOptions } from 'electron';
+
+type A = Required<BrowserWindowConstructorOptions>;
```
nornagon, line 7: Can I request a more descriptive name than `A`?

lib/common/parse-features-string.ts
```diff
@@ -0,0 +1,115 @@
...
+type KeysOfTypeInteger = {
```
nornagon, line 8:
```suggestion
type IntegerBrowserWindowOptionKeys = {
```
?

lib/common/parse-features-string.ts
```diff
@@ -0,0 +1,115 @@
...
+    const [key, value] = keyValuePair.split('=').map(str => str.trim());
+
+    if (useSoonToBeDeprecatedBehaviorForBareKeys && value === undefined) {
+      bareKeys.push(key);
+      return map;
+    }
+
+    map[key] = coerce(key, value);
+    return map;
+  }, {} as { [key: string]: any });
```
nornagon, line 71: i'm a big fan of `reduce` but i think a `forEach` or a loop would be easier to read here.

```suggestion
  const parsed = {} as { [key: string]: any };
  for (const keyValuePair of source.split(',')) {
    const [key, value] = keyValuePair.split('=').map(str => str.trim());

    if (useSoonToBeDeprecatedBehaviorForBareKeys && value === undefined) {
      bareKeys.push(key);
      continue;
    }

    parsed[key] = coerce(key, value);
  }
```

lib/common/parse-features-string.ts
```diff
@@ -0,0 +1,115 @@
...
+  features: string,
+  useSoonToBeDeprecatedBehaviorForBareKeys: boolean = true
+): {
+  options: Omit<BrowserWindowConstructorOptions, 'webPreferences'> & { [key: string]: CoercedValue };
+  webPreferences: BrowserWindowConstructorOptions['webPreferences'];
+  additionalFeatures: string[];
+} {
+  const { parsed, bareKeys } = parseCommaSeparatedKeyValue(features, useSoonToBeDeprecatedBehaviorForBareKeys);
+
+  const webPreferences = {};
```
nornagon, line 100:
```suggestion
  const webPreferences = {} as Record<string, any>;
```

lib/common/parse-features-string.ts
```diff
@@ -0,0 +1,115 @@
...
+  allowedWebPreferences.forEach((key) => {
+    if (parsed[key] === undefined) return;
+    (webPreferences as any)[key] = parsed[key];
```
nornagon, line 103:
```suggestion
    webPreferences[key] = parsed[key];
```

spec-main/guest-window-manager-spec.ts
```diff
@@ -0,0 +1,108 @@
...
+      snapshotFileName: 'proxy-window-open.snapshot.txt',
+      browserWindowOptions: {
+        show: false
+      }
+    }
+  };
+
+  let testName : keyof typeof testConfig;
+  // eslint-disable-next-line guard-for-in
+  for (testName in testConfig) {
```
nornagon, line 42:
```suggestion
  for (testName of Object.keys(testConfig)) {
```

lib/common/parse-features-string.ts
loc: I'm into it.

lib/common/parse-features-string.ts
```diff
@@ -0,0 +1,110 @@
...
+ */
+export function parseFeatures (
+  features: string,
+  useSoonToBeDeprecatedBehaviorForBareKeys: boolean = true
+) {
+  const { parsed, bareKeys } = parseCommaSeparatedKeyValue(features, useSoonToBeDeprecatedBehaviorForBareKeys);
+
+  const webPreferences: { [K in AllowedWebPreference]?: any } = {};
+  allowedWebPreferences.forEach((key) => {
+    if (parsed[key] === undefined) return;
```
nornagon, line 97: Just to be defensive against the difference between `parsed[key]` being explicitly set to `undefined` vs not present in the object,, perhaps rather test `key in parsed`? Or I suppose lint will make you do `Object.prototype.hasOwnProperty.call(parsed, key)`

lib/common/parse-features-string.ts
```diff
@@ -0,0 +1,110 @@
...
+    case 'no':
+      return false;
+    default:
+      return value;
+  }
+}
+
+export function parseCommaSeparatedKeyValue (source: string, useSoonToBeDeprecatedBehaviorForBareKeys: boolean) {
+  const bareKeys = [] as string[];
+  const parsed = {} as { [key: string]: any };
```
nornagon, line 61:
```suggestion
  const parsed = {} as { [key: string]: CoercedValue };
```

I think?

lib/common/parse-features-string.ts
```diff
@@ -0,0 +1,110 @@
...
+ * not respect this. In order to not break any applications, this will be
+ * flipped in the next major version.
+ */
+export function parseFeatures (
+  features: string,
+  useSoonToBeDeprecatedBehaviorForBareKeys: boolean = true
+) {
+  const { parsed, bareKeys } = parseCommaSeparatedKeyValue(features, useSoonToBeDeprecatedBehaviorForBareKeys);
+
+  const webPreferences: { [K in AllowedWebPreference]?: any } = {};
```
nornagon, line 95:
```suggestion
  const webPreferences: { [K in AllowedWebPreference]?: CoercedValue } = {};
```
?

spec-main/fixtures/snapshots/proxy-window-open.snapshot.txt
```diff
@@ -0,0 +1,161 @@
...
+      "title": "frame name",
+      "webPreferences": {
+        "nodeIntegration": false,
+        "webviewTag": false,
+        "nodeIntegrationInSubFrames": false,
+        "openerId": "placeholder-opener-id"
+      },
+      "width": 800,
+      "height": 600
+    },
```
nornagon, line 153: should we perhaps test just this chunk, rather than the whole event?

lib/common/parse-features-string.ts
loc: Yeah, considered that, but the keys of webPreferences in the BrowserWindowConstructor are typed specifically, so the error would be something like "CoercedValue is not of type boolean", for example. Might come back to this and figure something better out later.

lib/common/parse-features-string.ts
loc: There is no way to set undefined in the windowFeatures string. So there's only the latter "not present in the object" case to worry about.

spec-main/fixtures/snapshots/proxy-window-open.snapshot.txt
loc: Testing all the args makes more sense in the scope of the broader new-window refactor, of which this is but one piece. Lots of things are changing and I wanted to make sure I wasn't losing anything. IMO, it's a useful thing to have in place, even if I weren't changing things drastically in the coming PRs.

lib/common/parse-features-string.ts
nornagon: Ok. I think it's a good idea to do this anyway to defend against future refactors which might let an `undefined` value through, but I won't block the PR on it.

pushkin-:
@loc It *seems* like this PR introduced a change where the options object that's passed to the new-window event handler now has undefined x/y values whereas before, they were defaulted to 0. Is that an intentional change?

pushkin-:
@loc Any thought on this? Should I create an issue for this change, or is it expected?

nornagon:
@pushkin- opening an issue is the right way to go I think. Would be great to include an example gist/fiddle!

Issue #23139, opened 2020-04-16
tray.click action causes the app to get crashed


vladimiry:
* **Electron Version:** 10.0.0-nightly.20200415
* **Operating System:** Linux, Windows 10 (at least)
* **Last Known Working Electron version:** 10.0.0-nightly.20200327 (so it's broken since the next npm-published release which is `10.0.0-nightly.20200330`)

### Expected Behavior
- The app doesn't get segfaulted.
- `tray.click handler, NOOP action` message gets printed to console output.

### Actual Behavior
The app gets segfaulted with `ERROR:electron_bindings.cc(40)] Fatal error in V8: v8::HandleScope::CreateHandle() Cannot create a handle without a HandleScope` message printed to the console.

### To Reproduce
- git clone https://github.com/vladimiry-playground/electron-quick-start-tray-click-segfault
- cd electron-quick-start-tray-click-segfault
- npm ci
- npm start
- click the tray icon using primary/left mouse button

### Additional Information

The issue looks like a blocker.

vladimiry:
Can confirm issue resolving in 10.0.0-nightly.20200428, thanks everyone involved in fixing it.

PR #23140, opened 2020-04-16
fix: missing handlescopes in event emission

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/23139.

Fixes crashes in `CreateEventFromFlags` and `EmitCustomEvent` owing to missing `HandleScopes`.

cc @deepak1556 @zcbenz @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed some event listener crashes in Tray.

### DIFF
--- a/shell/browser/api/electron_api_tray.cc
+++ b/shell/browser/api/electron_api_tray.cc
@@ -93,14 +93,20 @@ gin::Handle<Tray> Tray::New(gin_helper::ErrorThrower thrower,
 void Tray::OnClicked(const gfx::Rect& bounds,
                      const gfx::Point& location,
                      int modifiers) {
+  v8::Isolate* isolate = JavascriptEnvironment::GetIsolate();
+  v8::HandleScope scope(isolate);
   EmitCustomEvent("click", CreateEventFromFlags(modifiers), bounds, location);
 }
 
 void Tray::OnDoubleClicked(const gfx::Rect& bounds, int modifiers) {
+  v8::Isolate* isolate = JavascriptEnvironment::GetIsolate();
+  v8::HandleScope scope(isolate);
   EmitCustomEvent("double-click", CreateEventFromFlags(modifiers), bounds);
 }
 
 void Tray::OnRightClicked(const gfx::Rect& bounds, int modifiers) {
+  v8::Isolate* isolate = JavascriptEnvironment::GetIsolate();
+  v8::HandleScope scope(isolate);
   EmitCustomEvent("right-click", CreateEventFromFlags(modifiers), bounds);
 }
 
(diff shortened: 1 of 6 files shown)


shell/browser/api/electron_api_tray.cc
```diff
@@ -63,7 +63,7 @@ gin::WrapperInfo Tray::kWrapperInfo = {gin::kEmbedderNativeGin};
 Tray::Tray(gin::Handle<NativeImage> image,
            base::Optional<UUID> guid,
            gin::Arguments* args)
-    : tray_icon_(TrayIcon::Create(guid)) {
+    : tray_icon_(TrayIcon::Create(guid)), isolate_(args->isolate()) {
```
nornagon, line 66: there's no need to save isolates in the main process because there's only ever 1. it's `v8::Isolate::GetCurrent()`.

codebytere: @nornagon calling `v8::isolate::GetCurrent()` repeatedly is much more costly, since there are atomic loads in v8::isolate::GetCurrent(), so it's preferable to avoid doing so

shell/browser/api/electron_api_tray.cc
```diff
@@ -93,14 +93,17 @@ gin::Handle<Tray> Tray::New(gin_helper::ErrorThrower thrower,
 void Tray::OnClicked(const gfx::Rect& bounds,
                      const gfx::Point& location,
                      int modifiers) {
+  v8::HandleScope scope(isolate());
   EmitCustomEvent("click", CreateEventFromFlags(modifiers), bounds, location);
```
deepak1556, line 97: I think this is better handled in `event_emitter_mixin.h` for these event methods `Emit`, `EmitCustomEvent` etc

shell/browser/api/electron_api_tray.cc
nornagon: I don't think it makes a difference on the scale of this API. But if we're seriously worried about the performance overhead of the atomic load there, we should create a singleton that holds the isolate reference. It's a bit silly to pass it around everywhere.

shell/browser/event_emitter_mixin.h
```diff
@@ -39,6 +39,7 @@ class EventEmitterMixin {
                        v8::Local<v8::Object> custom_event,
                        Args&&... args) {
     v8::Isolate* isolate = v8::Isolate::GetCurrent();
+    v8::HandleScope scope(isolate);
```
zcbenz, line 42: The `EmitCustomEvent` takes a `v8::Local` as parameter, I think we should add scope in the caller instead of here, otherwise the passed parameter is still leaked.

codebytere: @deepak1556 what do you think here? The former approach (from the caller) is a little weirder since we need to likely store the isolate as a member to prevent unnecessary loads, but i agree that we may wish to avoid leaking the passed parameter.

deepak1556: Scoping `electron::api::CreateEventFromFlags` with an `EscapableHandleScope` should solve this instead, and other callers of `EmitCustomEvent` can do the same but right now there is only one usage, thoughts ?

shell/browser/api/electron_api_tray.cc
nornagon: I don't see how that would solve this here as `CreateEventFromFlags` returns a `v8::Local`, so needs a handle scope.

shell/browser/event_emitter_mixin.h
nornagon: If there's a value returned from an escapable handle scope, then the calling function must have a handlescope anyway, so I don't see how that addresses the issue here?

deepak1556: The value returned is passed directly to `EmitCustomEvent` , so the copied value lands in here rather than the function calling `EmitCustomEvent`, shouldn't that be safe ?

`EmitCustomEvent('..., CreateEventFromFlags(), ...');`

deepak1556: But I am onboard with the singleton isolate holder, it will serve other use cases, this was just subtle that I was looking at other solutions.

nornagon: > The value returned is passed directly to `EmitCustomEvent` , so the copied value lands in here rather than the function calling `EmitCustomEvent`, shouldn't that be safe ?
> 
> `EmitCustomEvent('..., CreateEventFromFlags(), ...');`

No. Escapable handle scopes allow you to "push" a value into a handlescope that's higher up in the stack. If there's only one handlescope on the stack, "pushing" it up will fail.

PR #23153, opened 2020-04-17
fix: use Node's microtasks policy in node_main.cc

### LINKED ISSUE #21515: await Promise with Timer behavior regress from previous version
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 5.0.12 - 7.1.10
* **Operating System:**
  * Windows 10/macOS 10.15
  * 4.2.5

### Expected Behavior
We have been using the piece of code to perform asynchronous operations.
```js
await new Promise((resolve) => { setTimeout(resolve, 1000); }
```
It would pause the execution of the code path containing it for about 1 second and resume it then. This has been working well for a long time.

### Actual Behavior
Since we upgrade to Electron 7, in forked node processes, only the first several timer callbacks are called on time, while the rest of them are called extremely late for unknown reason.

Suppose we call the above piece of code for 10 times in a loop. Only the first two promises are created and resolved correctly. In addition, the process exits for unknown reason. In our production code, since there are other things going on (not computationally intensive) which keeps the process alive, we are seeing the 3rd and the rest of the Timers being created and invoked unreasonably late (around 12s after the 2nd one resolves).

The output in the renderer console is
```
Spawned child process <some_pid>
child process send a message INIT_ACK
child process send a message INIT_ACK
child process exits
```

### To Reproduce
https://github.com/JasonYeMSFT/electron-quick-start/tree/bug/await-Promise

To run the repro steps
```sh
$ git clone https://github.com/JasonYeMSFT/electron-quick-start -b bug/await-Promise
$ npm install
$ npm start
```

### Additional Information
Electron 4.2.5 behaves correctly for exactly the same code in this repo. The behavior doesn't reproduce in renderer process. Also it doesn't seem to be a nodejs problem because I can run the renderer.js directly with node on the same version and it behaves correctly.

I did some extra experiment and it turns out that `setInterval` is producing much more stable behaviors compared to `setTimeout` (update, I was wrong, `setInterval` also has the weird behavior).

ckerr:
Fixes #21515. See that ticket for a looong writeup.

#### Description of Change

Node [uses](https://github.com/nodejs/node/blob/666c67e078bddc32f25409f4a929c1e9b5f47373/src/api/environment.cc#L191) a `kExplicit` microtasks policy; previously we were using the V8 default value of `kAuto`. This PR ensures we use the same policy as Node by extracting it from `node::IsolateSettings`.

CC @codebytere,@nornagon

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed Promise timeout issue when running Electron as Node.

### DIFF
--- a/shell/app/node_main.cc
+++ b/shell/app/node_main.cc
@@ -146,6 +146,10 @@ int NodeMain(int argc, char* argv[]) {
     JavascriptEnvironment gin_env(loop);
 
     v8::Isolate* isolate = gin_env.isolate();
+    // TODO(ckerr) and/or TODO(codebytere) use node::SetIsolateMiscHandlers()
+    node::IsolateSettings is;
+    isolate->SetMicrotasksPolicy(is.policy);
+
     v8::Isolate::Scope isolate_scope(isolate);
     v8::Locker locker(isolate);
     node::Environment* env = nullptr;
--- a/spec-main/node-spec.ts
+++ b/spec-main/node-spec.ts
@@ -307,4 +307,17 @@ describe('node feature', () => {
     const result = childProcess.spawnSync(process.execPath, [path.resolve(fixtures, 'api', 'electron-main-module', 'app.asar')]);
     expect(result.status).to.equal(0);
   });
+
+  it('handles Promise timeouts correctly', (done) => {
+    const scriptPath = path.join(fixtures, 'module', 'node-promise-timer.js');
+    const child = childProcess.spawn(process.execPath, [scriptPath], {
+      env: { ELECTRON_RUN_AS_NODE: 'true' }
+    });
+    emittedOnce(child, 'exit').then(([code, signal]) => {
+      expect(code).to.equal(0);
+      expect(signal).to.equal(null);
+      child.kill();
+      done();
+    });
+  });
 });
--- a/spec/fixtures/module/node-promise-timer.js
+++ b/spec/fixtures/module/node-promise-timer.js
@@ -1,0 +1,23 @@
+const waitMs = (msec) => new Promise((resolve) => setTimeout(resolve, msec));
+
+const intervalMsec = 100;
+const numIterations = 2;
+let curIteration = 0;
+let promise;
+
+for (let i = 0; i < numIterations; i++) {
+  promise = (promise || waitMs(intervalMsec)).then(() => {
+    ++curIteration;
+    return waitMs(intervalMsec);
+  });
+}
+
+// https://github.com/electron/electron/issues/21515 was about electron
+// exiting before promises finished. This test sets the pending exitCode
+// to failure, then resets it to success only if all promises finish.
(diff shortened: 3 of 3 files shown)


shell/app/node_main.cc
```diff
@@ -146,6 +146,10 @@ int NodeMain(int argc, char* argv[]) {
     JavascriptEnvironment gin_env(loop);
 
     v8::Isolate* isolate = gin_env.isolate();
+    // TODO(ckerr) and/or TODO(codebytere) use node::SetIsolateMiscHandlers()
+    node::IsolateSettings is;
```
codebytere, line 150: for posterity's sake, can we note that this means `v8::MicrotasksPolicy::kExplicit`?

nornagon:
This seems like a potentially dangerous change since we aren't particularly careful to execute microtasks when we exit a C++ context.

The [V8 docs on MicrotasksPolicy](https://source.chromium.org/chromium/chromium/src/+/master:v8/include/v8.h;l=7335-7343;drc=c986fbbe38e50ba884e20026fe00abe47be94f02) say that `auto` is "microtasks are invoked when the script call depth decrements to zero".


I'm a little hazy as to what the implications are of not having good hygiene around executing microtasks when we ought to be. Could there be issues where e.g. we get an event in C++ (like a tray icon click, for instance), which we then run a JS callback for, but we forget to run microtasks and then the main thread goes back to sleep again without having executed microtasks? Could that result in weird things like promises or timers not executing when they're supposed to?

shell/app/node_main.cc
ckerr: I guess we could; but if that's the goal, we'd be better off just using `kExplicit` here instead of IsolateSettngs IMO

My rationale for using `node::IsolateSettings` wasn't so much "show that we're using kExplicit" as it was "show that we're using whatever Node is using" -- e.g. if they were to change from kExplicit to something else in their code, we would pick up that change automatically.

codebytere:
@nornagon we don't run into this problem in non-emulation mode bc we already [override it](https://github.com/electron/electron/blob/0a78ab4b98c078814e6aed498ff96947733888cd/shell/common/node_bindings.cc#L399). @ckerr has a great writeup linked in PR :)

nornagon:
@codebytere ah, I missed the bit in the writeup that mentions that we set this elsewhere during non-RUN_AS_NODE mode.

That's exactly the thing I was concerned about though, which is that we're setting the `explicit` microtasks mode (at least in the main process), but we possibly aren't consistently invoking microtasks when we need to be.

codebytere:
To my knowledge we do via our [microtasks runner](https://github.com/electron/electron/blob/49b47ee4edb6e90545ca56ed2d3c051437df9521/shell/browser/microtasks_runner.h) - we have a  specific task observer to flush the queued microtasks which runs at the end of every task and is based off of Blink's `EndOfTaskRunner`

PR #23180, opened 2020-04-20
build: fix crashpad+asan

nornagon:
#### Description of Change
This fixes an issue where the crashpad reporter wouldn't start when building with asan, due to not being able to find the asan dylib, e.g.

```
dyld: Library not loaded: @executable_path/libclang_rt.asan_osx_dynamic.dylib
  Referenced from: /Users/nornagon/work/electron-gn/src/out/Testing/Electron.app/Contents/Frameworks/Electron Framework.framework/Resources/crashpad_handler
  Reason: image not found
```

This would manifest when running an app as a log message when starting the crashpad handler:

```
[85013:0420/143027.954855:ERROR:file_io.cc(89)] ReadExactly: expected 8, observed 0
```

This snippet is copied from [`//chrome/BUILD.gn:675`](https://chromium.googlesource.com/chromium/src/+/3bdd2420941cba3c4754923d842e44c9b30a1960/chrome/BUILD.gn#675).

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -757,6 +757,12 @@ if (is_mac) {
     outputs = [ "{{bundle_resources_dir}}/{{source_file_part}}" ]
 
     public_deps = [ "//third_party/crashpad/crashpad/handler:crashpad_handler" ]
+
+    if (is_asan) {
+      # crashpad_handler requires the ASan runtime at its @executable_path.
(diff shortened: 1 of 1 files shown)


nornagon:
NB. CI is irrelevant here (other than lint) because we don't build with `is_asan` in CI.

PR #23187, opened 2020-04-20
test: fix type errors in devToolsWebContents

codebytere:
#### Description of Change

Fixes type errors in `api-web-contents-spec.ts`. We can assert that `devToolsWebContents` is not `null` safely when `w.webContents.openDevTools()` has been invoked.

cc @nornagon 

Notes: none


deepak1556:
Failing tests are unrelated, merging

PR #23188, opened 2020-04-21
fix: block custom window.open when nativeWindowOpen is true

nornagon:
#### Description of Change
This prevents windows that have `nativeWindowOpen: true` (or `sandbox: true` which implies it) from being able to use the `ELECTRON_GUEST_WINDOW_MANAGER_WINDOW_OPEN` ipc event, which handles non-native window.open calls.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/lib/browser/guest-window-manager.js
+++ b/lib/browser/guest-window-manager.js
@@ -219,6 +219,12 @@ const canAccessWindow = function (sender, target) {
 
 // Routed window.open messages with raw options
 ipcMainInternal.on('ELECTRON_GUEST_WINDOW_MANAGER_WINDOW_OPEN', (event, url, frameName, features) => {
+  // This should only be allowed for senders that have nativeWindowOpen: false
+  const webPreferences = event.sender.getLastWebPreferences();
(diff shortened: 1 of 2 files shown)


lib/browser/guest-window-manager.js
```diff
@@ -219,6 +219,8 @@ const canAccessWindow = function (sender, target) {
 
 // Routed window.open messages with raw options
 ipcMainInternal.on('ELECTRON_GUEST_WINDOW_MANAGER_WINDOW_OPEN', (event, url, frameName, features) => {
+  // This should only be allowed for senders that have nativeWindowOpen: false
+  if (event.sender.getLastWebPreferences().nativeWindowOpen || event.sender.getLastWebPreferences().sandbox) { throw new Error('GUEST_WINDOW_MANAGER_WINDOW_OPEN denied: expected native window.open'); }
```
MarshallOfSound, line 223: sandbox: true does not imply nativeWindowOpen afaik

nornagon: it does, since https://github.com/electron/electron/pull/18273

PR #23194, opened 2020-04-21
refactor: port parts of window-setup to use ctx bridge instead of being run in the main world

MarshallOfSound:
This changes uses all the work I did for `contextBridge` and exposes some new helpers internally that allow us to do some of the work `window-setup` was doing without punching a hole between the isolated worlds.  This covers every function other than the `window.open` and `window.opener` bits as they will require slightly more work.

The end goal of this work is to completely remove the `isolated_bundle` that punches a hole between contexts.

Notes: no-notes

### DIFF
--- a/lib/renderer/api/context-bridge.ts
+++ b/lib/renderer/api/context-bridge.ts
@@ -18,3 +18,14 @@ const contextBridge = {
 if (!binding._debugGCMaps) delete contextBridge.debugGC;
 
 export default contextBridge;
+
+export const internalContextBridge = {
+  contextIsolationEnabled,
+  overrideGlobalMethodFromIsolatedWorld: (keys: string[], method: Function) => {
+    return binding._overrideGlobalMethodFromIsolatedWorld(keys, method);
+  },
+  overrideGlobalPropertyFromIsolatedWorld: (keys: string[], getter: Function, setter?: Function) => {
+    return binding._overrideGlobalPropertyFromIsolatedWorld(keys, getter, setter || null);
+  },
+  isInMainWorld: () => binding._isCalledFromMainWorld({}) as boolean
+};
--- a/lib/renderer/api/context-bridge.ts
+++ b/lib/renderer/api/context-bridge.ts
@@ -27,5 +27,5 @@ export const internalContextBridge = {
   overrideGlobalPropertyFromIsolatedWorld: (keys: string[], getter: Function, setter?: Function) => {
     return binding._overrideGlobalPropertyFromIsolatedWorld(keys, getter, setter || null);
   },
-  isInMainWorld: () => binding._isCalledFromMainWorld({}) as boolean
+  isInMainWorld: () => binding._isCalledFromMainWorld() as boolean
 };
--- a/lib/renderer/window-setup.ts
+++ b/lib/renderer/window-setup.ts
@@ -5,6 +5,10 @@ import { internalContextBridge } from '@electron/internal/renderer/api/context-b
 const inMainWorld = internalContextBridge.isInMainWorld();
 const { contextIsolationEnabled } = internalContextBridge;
 
+// Should we inject APIs into this world, if ctx isolation is enabled then only inject in the isolated world
+// else inject everywhere
+const shouldInjectGivenContextIsolationIsMaybeEnabled = contextIsolationEnabled ? !inMainWorld : true;
+
 // This file implements the following APIs Directly:
 // - window.open()
 // - window.opener.blur()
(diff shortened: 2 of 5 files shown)


lib/renderer/window-setup.ts
```diff
@@ -230,28 +245,31 @@ export const windowSetup = (
     });
   }
 
-  if (!process.sandboxed && !rendererProcessReuseEnabled) {
+  if (!process.sandboxed && !rendererProcessReuseEnabled && !(inMainWorld && contextIsolationEnabled)) {
```
nornagon, line 248: this conditional is super confusing, perhaps we can do de Morgan + naming to it?
```suggestion
  const usesNativeHistory = process.sandboxed || rendererProcessReuseEnabled;
  const inPrivilegedContext = !contextIsolationEnabled || !inMainWorld;
  if (!usesNativeHistory && inPrivilegedContext) {
```

lib/renderer/window-setup.ts
```diff
@@ -230,28 +245,31 @@ export const windowSetup = (
...
-        return ipcRendererInternal.sendSync('ELECTRON_NAVIGATION_CONTROLLER_LENGTH');
-      },
+      get: getHistoryLength,
       set () {}
     });
+    if (contextIsolationEnabled) internalContextBridge.overrideGlobalPropertyFromIsolatedWorld(['history', 'length'], getHistoryLength);
   }
 
-  if (guestInstanceId != null) {
+  if (guestInstanceId != null && !(inMainWorld && contextIsolationEnabled)) {
```
nornagon, line 272:
```suggestion
  if (guestInstanceId != null && inPrivilegedContext) {
```

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -484,6 +484,104 @@ void ExposeAPIInMainWorld(const std::string& key,
...
+  auto* frame = render_frame->GetWebFrame();
+  CHECK(frame);
+  v8::Local<v8::Context> main_context = frame->MainWorldScriptContext();
+  gin_helper::Dictionary global(main_context->GetIsolate(),
+                                main_context->Global());
+
+  const std::string final_key = key_path[key_path.size() - 1];
+  gin_helper::Dictionary target_object = TraceKeyPath(global, key_path);
+
+  v8::Context::Scope main_context_scope(main_context);
```
nornagon, line 515: Put the scope inside the `{}` to control its lifetime. If code is added to the end of this function it shouldn't accidentally be inside the main context scope.

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -484,6 +484,104 @@ void ExposeAPIInMainWorld(const std::string& key,
...
+        PassValueToOtherContext(method->CreationContext(), main_context, method,
+                                store, &object_cache, 1);
+    DCHECK(!maybe_proxy.IsEmpty());
+    auto proxy = maybe_proxy.ToLocalChecked();
+
+    target_object.Set(final_key, proxy);
+  }
+}
+
+bool OverrideGlobalPropertyFromIsolatedWorld(
```
nornagon, line 528: The `bool` return value seems to never be used?

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -484,6 +484,104 @@ void ExposeAPIInMainWorld(const std::string& key,
...
+    const std::vector<std::string>& key_path,
+    v8::Local<v8::Object> getter,
+    v8::Local<v8::Value> setter,
```
nornagon, line 531: Why is `getter` an `Object` but `setter` a `Value`? Shouldn't they both be `Function`s?

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -484,6 +484,104 @@ void ExposeAPIInMainWorld(const std::string& key,
...
+    }
+
+    v8::PropertyDescriptor desc(getter_proxy, setter_proxy);
+    return IsTrue(target_object->DefineProperty(
+        main_context, gin::StringToV8(args->isolate(), final_key), desc));
+  }
+}
+
+bool IsCalledFromMainWorld(v8::Local<v8::Object> sample) {
+  auto* render_frame = GetRenderFrame(sample);
```
nornagon, line 577: I don't think we need to pass an object through here; we should be able to get the calling context without it. `isolate->GetCurrentContext()` should be correct.

shell/renderer/api/electron_api_context_bridge.cc
MarshallOfSound: Yeah but cpp doesn't know that 😉 The return value I provide has a warn unused 😄 It is also handy for debugging when it returned false for a while when I wasn't using it right

shell/renderer/api/electron_api_context_bridge.cc
MarshallOfSound: `setter` can be null / undefined.  `getter` doesn't have to be a function it just has to be callable.  The API we pass it to expects an object not a function 🤔

shell/renderer/api/electron_api_context_bridge.cc
nornagon: `CHECK` or `DCHECK` seems like a better way to handle that.

lib/renderer/api/context-bridge.ts
```diff
@@ -18,3 +18,14 @@ const contextBridge = {
...
+
+export const internalContextBridge = {
+  contextIsolationEnabled,
+  overrideGlobalMethodFromIsolatedWorld: (keys: string[], method: Function) => {
+    return binding._overrideGlobalMethodFromIsolatedWorld(keys, method);
+  },
+  overrideGlobalPropertyFromIsolatedWorld: (keys: string[], getter: Function, setter?: Function) => {
+    return binding._overrideGlobalPropertyFromIsolatedWorld(keys, getter, setter || null);
+  },
+  isInMainWorld: () => binding._isCalledFromMainWorld({}) as boolean
```
nornagon, line 30:
```suggestion
  isInMainWorld: () => binding._isCalledFromMainWorld() as boolean
```

PR #23207, opened 2020-04-21
fix: ensure that functions are not retained beyond their context being released

MarshallOfSound:
We were cleaning up the `RenderFrameFunctionStore` instances when a render frame was destroyed.  For same site navigations / reloads the render frame isn't destroyed, it's reused so we were leaking the old context and all the functions.  This updates our impl to listen for WillReleaseScriptContext and release our holds on objects in that context once that context is about to be released.

Also adds a test for good measure 😄 Had to update the contextBridge specs to use an actual HTTP server so that same-site rules applied

Notes: fixed memory leak when using contextBridge with sandbox=true

### DIFF
--- a/shell/renderer/api/context_bridge/render_frame_function_store.cc
+++ b/shell/renderer/api/context_bridge/render_frame_function_store.cc
@@ -32,6 +32,16 @@ void RenderFrameFunctionStore::OnDestruct() {
   delete this;
 }
 
+void RenderFrameFunctionStore::WillReleaseScriptContext(
+    v8::Local<v8::Context> context,
+    int32_t world_id) {
+  base::EraseIf(functions_, [context](auto const& pair) {
+    v8::Local<v8::Context> func_owning_context =
+        std::get<1>(pair.second).Get(context->GetIsolate());
+    return func_owning_context == context;
+  });
+}
+
 }  // namespace context_bridge
 
 }  // namespace api
--- a/spec-main/api-context-bridge-spec.ts
+++ b/spec-main/api-context-bridge-spec.ts
@@ -2,17 +2,33 @@ import { BrowserWindow, ipcMain } from 'electron/main';
 import { contextBridge } from 'electron/renderer';
 import { expect } from 'chai';
 import * as fs from 'fs-extra';
+import * as http from 'http';
 import * as os from 'os';
 import * as path from 'path';
 
 import { closeWindow } from './window-helpers';
 import { emittedOnce } from './events-helpers';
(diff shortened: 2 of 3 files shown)


shell/renderer/api/context_bridge/render_frame_function_store.cc
```diff
@@ -32,6 +32,20 @@ void RenderFrameFunctionStore::OnDestruct() {
...
+    int32_t world_id) {
+  std::vector<size_t> to_remove;
+  for (auto const& pair : functions_) {
+    v8::Local<v8::Context> func_owning_context =
+        std::get<1>(pair.second).Get(context->GetIsolate());
+    if (func_owning_context == context)
+      to_remove.push_back(pair.first);
+  }
+  for (auto remove : to_remove)
+    functions_.erase(remove);
```
nornagon, line 46: use `base::EraseIf`

spec-main/api-context-bridge-spec.ts
```diff
@@ -2,17 +2,33 @@ import { BrowserWindow, ipcMain } from 'electron/main';
...
 
 describe('contextBridge', () => {
   let w: BrowserWindow;
   let dir: string;
+  let server: http.Server;
+
+  before(async () => {
+    server = http.createServer((req, res) => {
+      res.setHeader('Content-Type', 'text/html');
+      res.end(fs.readFileSync(path.resolve(fixturesPath, 'empty.html')));
```
nornagon, line 23:
```suggestion
      res.end('');
```

spec-main/api-context-bridge-spec.ts
```diff
@@ -2,17 +2,33 @@ import { BrowserWindow, ipcMain } from 'electron/main';
...
 
 describe('contextBridge', () => {
   let w: BrowserWindow;
   let dir: string;
+  let server: http.Server;
+
+  before(async () => {
+    server = http.createServer((req, res) => {
+      res.setHeader('Content-Type', 'text/html');
+      res.end('<html></html>');
```
nornagon, line 23: is the `<html>` necessary?

MarshallOfSound: Just keeping it consistent with what we had before.  Plus I don't like the idea of a zero-byte document 😆

nornagon: this server didn't exist before, so there's nothing to be consistent with. and zero-byte documents are just fine :)

MarshallOfSound: Updated

PR #23213, opened 2020-04-22
refactor: use Node.js isolate setup in node mode

codebytere:
#### Description of Change

I did some digging around and realized that while we cannot easily use `SetupIsolateForNode` in our `node_bindings`, there's not a reason I can see that we can't use it in emulation mode.

This allows us to remove a significant amount of cruft we had to replicate isolate setup behavior and also turns out to be the cause of `async_hooks` failures in emulation mode. I plan to dig further into the `async_hooks` differences and port what i find over to our node bindings to hopefully fix associated issues in Electron itself.

cc @MarshallOfSound @ckerr 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


shell/app/node_main.cc
```diff
@@ -167,13 +146,11 @@ int NodeMain(int argc, char* argv[]) {
       // TODO(codebytere): we shouldn't have to call this - upstream?
       env->InitializeDiagnostics();
 
-      // This is needed in order to enable v8 host weakref hooks.
-      // TODO(codebytere): we shouldn't have to call this - upstream?
-      isolate->SetHostCleanupFinalizationGroupCallback(
-          HostCleanupFinalizationGroupCallback);
+      // We do not want to set the error level.
+      node::IsolateSettings is;
+      is.flags = ~node::IsolateSettingsFlags::MESSAGE_LISTENER_WITH_ERROR_LEVEL;
```
deepak1556, line 151: Why don't we want to set this ? what happens if we do it ?

codebytere: Electron relies on error reporting matching that of Blink/Chromium, which Node.js changes. If we set this it kills ~40 previously-passing tests (see [here](https://app.circleci.com/pipelines/github/electron/electron/23480/workflows/98efc491-72d9-46fe-9310-ff737775a34c/jobs/528118/tests)) since Electron is more sensitive to errors and will exit the process instead of deferring to other listeners in place.

deepak1556: > Electron relies on error reporting matching that of Blink/Chromium.
> .., since Electron is more sensitive to errors and will exit the process instead of deferring to other listeners in place.

Can you expand on how this behavior is inherited in `ELECTRON_RUN_AS_NODE` mode ?

I would expect the node env thats run in the child process to not have interference from the chromium process management.

codebytere: Sorry, 'relies' is not the best term. It collides with changes we've already made in [this patch](https://github.com/electron/electron/blob/fdf7e288bb049a681a46351e7770b68f41b5075e/patches/node/feat_add_flags_for_low-level_hooks_and_exceptions.patch#L1). Although, we could just edit the patch to handle that instead.

deepak1556: Thanks! completely missed that patch.

I am not surprised `BrowserEnvironment` is considered `BROWSER` for the `RUN_AS_NODE` mode. I think the fix here would be to

1) Rename `BrowserEnvironment` enum to `NodeEnvironment`
2) Add a new type `NODE`
3) Set this type when initializing `NodeBindings` from `node_main.cc`

This would avoid mixing `g_standalone_mode` in `ELECTRON_RUN_AS_NODE`. It is better to make the change in this PR.

deepak1556: Thanks! completely missed that patch, the fix here would be to

1) Rename `BrowserEnvironment` enum to `NodeEnvironment`
2) Add a new type `NODE`
3) Set this type when initializing `NodeBindings` from `node_main.cc`

This would avoid mixing `g_standalone_mode` in `ELECTRON_RUN_AS_NODE`. It is better to make the change in this PR.

deepak1556: Actually wait, since we don't initialize `NodeBindings` in `node_main` , I think simpler fix is to set `g_standalone_mode=false`

codebytere: I thought that'd work as well but it does not seem to.

deepak1556: Can you set the variable before `NodeBindings::RegisterBuiltinModules();` call , otherwise its not gonna affect where it needs to.

shell/app/node_main.cc
```diff
@@ -108,6 +108,10 @@ int NodeMain(int argc, char* argv[]) {
     crash_reporter::CrashReporterWin::SetUnhandledExceptionFilter();
 #endif
 
+    // We do not want to double-set the error level and promise rejection
+    // callback.
+    node::g_standalone_mode = false;
```
deepak1556, line 113: One more style nit, can you also add `node::g_upstream_node_mode=true` here  and remove the following from the patch

```
diff --git a/src/node.cc b/src/node.cc
index 7dda6b39c3..c084a44544 100644
--- a/src/node.cc
+++ b/src/node.cc
@@ -129,9 +129,6 @@ using v8::Undefined;
 using v8::V8;
 using v8::Value;
 
-bool g_standalone_mode = true;
-bool g_upstream_node_mode = true;
-
 namespace per_process {
 
 // node_revert.h
```

Its better and clear the extern variables are defined either in `node_main` or in electron through `node_bindings`.

deepak1556:
https://ci.appveyor.com/project/electron-bot/electron-ia32-testing/builds/32484753 -> win_32 failure is not a flake.

patches/node/feat_add_flags_for_low-level_hooks_and_exceptions.patch
```diff
@@ -52,26 +52,20 @@ index fee485f86e17f1987e140aea874656e5048cdbfe..7dda6b39c37cdf96f59a34ca6bc75ea2
...
-@@ -343,6 +347,13 @@ MaybeLocal<Value> Environment::RunBootstrapping() {
+@@ -343,6 +347,10 @@ MaybeLocal<Value> Environment::RunBootstrapping() {
  
    CHECK(!has_run_bootstrapping_code());
  
 +  if (g_standalone_mode) {
 +    isolate()->AddMessageListener(errors::PerIsolateMessageListener);
 +  }
-+  if (g_upstream_node_mode) {
-+    isolate()->SetFatalErrorHandler(OnFatalError);
```
deepak1556, line 73: is this necessary ?

codebytere: nope! already deleted :)

codebytere: yes - it's already invoked in `SetIsolateUpForNode` so this needlessly does it twice.

PR #23221, opened 2020-04-22

MarshallOfSound:
Doing this for a combination of reasons.

* We're planning on removing the `nativeWindowOpen: false` codepath in an upcoming release
* This option is more secure and we should be secure by default
* Our custom `window.open` is bad and not technically correct.  This change will make people have a better first-experience.

Notes: Enabled the `nativeWindowOpen` webPreference by default


docs/breaking-changes.md
```diff
@@ -40,6 +40,12 @@ const w = new BrowserWindow({
 We [recommend moving away from the remote
 module](https://medium.com/@nornagon/electrons-remote-module-considered-harmful-70d69500f31).
 
+### Default Changed: `nativeWindowOpen` defaults to `true`
+
+In Electron 10 the `nativeWindowOpen` webPreference will default to `true`.  This changes
+the provider of the `window.open` API to use Chromiums logic instead of Electrons.  This
+flag will be removed in a future Electron version.
```
nornagon, line 47:
```suggestion
### Default Changed: `nativeWindowOpen` defaults to `true`

In Electron 10 the `nativeWindowOpen` webPreference will default to `true`.  This changes
the provider of the `window.open` API to use Chromium's logic instead of Electron's.

This flag will be removed in a future Electron version, with the `nativeWindowOpen: true`
behavior becoming the only available option.
```

MarshallOfSound:
Need to rewrite all the tests, they are Bad ™️

PR #23224, opened 2020-04-22
fix: block custom window.open when nativeWindowOpen is true (#23188)

nornagon:
Backport of #23188.

Notes: Fixed an issue where windows without `nativeWindowOpen: true` could invoke the non-native-open path.


spec-main/chromium-spec.ts
```diff
@@ -72,6 +72,36 @@ describe('reporting api', () => {
...
+      });
+      w.loadURL('about:blank');
+
+      const previousListeners = process.listeners('uncaughtException');
+      process.removeAllListeners('uncaughtException');
+      try {
+        const uncaughtException = new Promise<Error>(resolve => {
+          process.once('uncaughtException', resolve);
+        });
+        expect(await w.webContents.executeJavaScript(`(${function () {
```
MarshallOfSound, line 94: TS isn't happy with this line apparently?

PR #23237, opened 2020-04-22
ci: robustify doc only change check

jkleinsc:
#### Description of Change

This PR updates our check for doc only changes to handle the use case where the GitHub API reports no files for a PR.  It shouldn't happen but if it does we should assume that the change is more than doc changes just to be safe.
#### Checklist

#### Release Notes

Notes: none


jkleinsc:

jkleinsc:
This code was not in 7-2-x and therefore cannot be backported there.

Issue #18829, opened 2019-06-16
Choosing appropriate fallback font list by system locale (especially on Windows)


sorah:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description

Under a Windows machine using English as a display language, Japanese kanji texts are rendered with Chinese font `SimSum`. This behavior matches with Chrome's one if the machine doesn't list Japanese as an alternative preferred locale, and HTML content language is set to `en-US`.
But if the machine lists Japanese as an alternative, Electron should render the text with Japanese fonts.

AFAIK, Chrome renders kanji texts (Han script) with Japanese fonts when the accept-language configuration includes Japanese, as reading Chromium (blink)'s [font_fallback_win.cc](https://cs.chromium.org/chromium/src/third_party/blink/renderer/platform/fonts/win/font_fallback_win.cc). This behavior is done by considering the system locales including secondary/alternative locales. Because of this, I don't encounter the same problem in Chrome.

Looks like Electron currently honors only primary system locale and content locale (HTML `lang`), but not secondary/alternative locales. Thus I think Electron's current behavior is not following Chrome's one.

As I'm not familiar with the Electron codebase, I'm still not sure why the current implementation doesn't make this problem on Mac. I'm listing Japanese as an alternative locale both on my Mac and Windows machine, but this problem happens only on Windows; as looking into font related codes in Chromium, I guess macOS' CoreText is doing good for choosing a font.

Related to #15486 #18197.

In real life, for Slack workspace/Discord servers where its primary language is Japanese, this problem forces people using English as a primary system language to read Japanese with an unmatched font, which is rendered ugly as a Japanese text. I don't think letting app developers using Electron to implement font configuration is happy, just for this problem. Electron should provide a good default to devs.

### Proposed Solution

Get _all_ preferred locales from OS and set them to Chromium engine. I'm guessing https://github.com/electron/electron/pull/15532 fixes this problem.

#### Japanese text shown on Slack app (Windows)

![Japanese text](https://img.sorah.jp/x/20190616_130936_PbjXWPa70A.png)

[...]

PR #15532, opened 2018-11-01
fix: navigator.languages to use OS preferred languages

nitsakh:
#### Description of Change
Until recent versions of chrome (i.e. for electron version < 4.x), `navigator.language` returned the system language, and the first item in `navigator.languages` array, was also the same system locale.
However, this behaviour has changed, (it also follows the HTML spec). The language preferences are obtained through the preferences list that chrome maintains. The user has to set these language preferences manually, through the settings menu.
For an electron app, there's no way to set these preferences. Besides, being a desktop app, it seems reasonable for the app to follow the users desktop language preferences.
This change makes the needed changes to set the right OS languages. This also fixes `app.getLocale` which always returned `en-US`.
Adding a spec for this would be difficult, as the os language preferences need to be changed to see the difference.

#### Checklist

- [ ] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [ ] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

#### Release Notes

Notes: Fixed behaviour of navigator.language/s and app.getLocale to use OS locale

PR #23247, opened 2020-04-23
fix: respect system language preferences on Windows & macOS

sorah:
#### Description of Change

This PR aims to fix https://github.com/electron/electron/issues/18829. Please refer to the issue for more details.

Add implementations to set user's OS language preferences to `Accept-Language` and `navigator.languages`, on Windows and macOS.

This also improves the CJK fallback font selection on Windows. Chromium uses `Accept-Language` to determine fallback fonts on Windows, especially kanji/han characters in CJK.

Previously the full preferences set to OS was not given to Chromium.  For instance, when user sets `en-US, ja-JP` to `Accept-Language`, while Chromium chooses Japanese font for kanji text, but Electron chooses Chinese font.  This is because only the first language is given to `Accept-Language` on Electron.

This patch is based on the idea given at https://github.com/electron/electron/pull/15532 /cc @nitsakh

image

##### fix or feat?

❓ No sure this a feat or a fix. Might be a fix, and backports would be appreciated as I'm thinking this is a bug for CJK speakers.

##### WinRT API

<s>In this implementation, [`GetThreadPreferedUILanguages`](https://docs.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-getthreadpreferreduilanguages) API is used on Windows. But, I noticed this API doesn't provide the full list of user preferences.
Only the system locale (i.e. used for logon screen) and the user locale (i.e. UI used while logged in) are returned.

So, how do you think about adding `ELECTRON_ACCEPT_LANGUAGE` environment variable to give the preferences correctly, in addition to this change? </s>

✅ Resolved using WinRT utility in Chromium https://github.com/electron/electron/pull/23247#issuecomment-618269313

#### Checklist

- [x] `npm test` passes

#### Release Notes

notes: Fixed an issue with `navigator.languages` and `Accept-Language` did not fully respect users' language preferences on Windows and macOS. This also improved fallback font selection for CJK texts on Windows.


hanazuki:
I have a sample application to retrieve Windows 10's Preferred Language setting: https://github.com/hanazuki/win-accept-languages-sample

This program makes use of [GlobalizationPreferences.Languages](https://docs.microsoft.com/en-us/uwp/api/windows.system.userprofile.globalizationpreferences.languages?view=winrt-18362#Windows_System_UserProfile_GlobalizationPreferences_Languages) from WinRT library, which is available since Windows 10 as with the Preferred Languages setting itself.
We can try this method first, and if not supported by the running OS, fallback to GetThreadPreferedUILanguages.

Note on compatibility for old Windows:
- Windows 7 or earlier does not support WinRT; we have to load `RoInitialize` and `RoGetActivationFactory` functions at runtime (i.e. using GetProcAddress) instead of linking runtimeobject.lib.
  - My sample does not do this. [Chromium seems to have a mechanism for it](https://source.chromium.org/chromium/chromium/src/+/master:base/win/core_winrt_util.cc).
- Windows 8 does not have GlobalizationPreferences; we can detect this case by checking the result of RoGetActivationFactory.

I hope this will help.

sorah:
Nice. I regret that I was not thinking that Chrome should have such utility as Chrome is utilizing modern Windows APIs...

sorah:
Changed to utilize Globalization API when available. Now trying to fix failing CIs.

sorah:
CI Passed. Lint failure seems not relevant to my changes..

sorah:
Opened https://github.com/electron/electron/pull/23266 for lint failure.

sorah:
rebased and only linux-ia32 is failing with flaky timeout. my change is ready.

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -535,7 +535,7 @@ void WebContents::InitWithSessionAndOptions(
   managed_web_contents()->GetView()->SetDelegate(this);
 
   auto* prefs = web_contents()->GetMutableRendererPrefs();
-  prefs->accept_languages = g_browser_process->GetApplicationLocale();
+  SetAcceptLanguages(prefs);
```
zcbenz, line 538: Can you move the language preferences code into separate files and then reference it here? Like this:

```c++
#include "shell/common/lang_util.h"

prefs->accept_languages = GetSystemLanguage();
```

It would make the code easier to maintain, and we might want to reuse the code in other places.

sorah: I wasn't sure where to put this kind of code. Extracted: d947ed9e7db679576d92045b7bdfe8254b37f645

zcbenz:
Can you fix the lint waring?

```
$ node ./script/lint.js
shell/common/language_util_mac.mm:14:  Add #include <string> for string  [build/include_what_you_use] [4]
shell/common/language_util_mac.mm:14:  Add #include <vector> for vector<>  [build/include_what_you_use] [4]
Total errors found: 2
```

sorah:
Fixed, rebased & squashed.

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -535,7 +536,22 @@ void WebContents::InitWithSessionAndOptions(
...
+  // Collect preferred languages from OS and browser process. accept_languages
+  // effects HTTP header, navigator.languages, and CJK fallback font selection.
+  //
+  // Note that an application locale set to the browser process might be
+  // different with the one set to the preference list.
+  // (e.g. overridden with --lang)
+  std::string accept_languages =
+      g_browser_process->GetApplicationLocale() + ",";
+  for (auto const& language : electron::GetPreferredLanguages()) {
+    if (language == g_browser_process->GetApplicationLocale())
```
deepak1556, line 549: would there be issue with fallback if there was a duplicate in the final string ?

deepak1556: Curious if we can remove this check incase its not necessary.

sorah: confirmed `navigator.languages`, and `Accept-Language` header would contain duplicated entry; this should have no issue but look strange from end-users maybe?

deepak1556: Cool, thanks for checking. In that case we can leave it in.

sorah:
nah, `l10n_util::GetApplicationLocale()` would block. changed to return empty vector (as getting one from a browser process can satisfy) at b8f4e09, but suggestions welcomed for possible Linux implementation.

deepak1556:
> l10n_util::GetApplicationLocale() would block...

Instead of overriding `g_browser_process->SetApplicationLocale` for `kLang` on linux, we can append the custom locale and the OS preferred locale to the final list in `ElectronMainDelegate::PostEarlyInitialization` which should avoid the blocking call check.

sorah:
do you mean generate the list at the MainDelegate and refer to it at WebContents? (I'm not familiar with the structure/hierarchy between MainDelegate and WebContents)

deepak1556:
Something like

```
diff --git a/shell/app/electron_main_delegate.cc b/shell/app/electron_main_delegate.cc
index 3e608d603..22a3d3763 100644
--- a/shell/app/electron_main_delegate.cc
+++ b/shell/app/electron_main_delegate.cc
@@ -238,6 +238,8 @@ bool ElectronMainDelegate::BasicStartupComplete(int* exit_code) {
 
 void ElectronMainDelegate::PostEarlyInitialization(bool is_running_tests) {
   std::string custom_locale;
+  std::string preferred_locale = l10n_util::GetApplicationLocale("");
+  ALLOW_UNUSED_LOCAL(preferred_locale);
   ui::ResourceBundle::InitSharedInstanceWithLocale(
       custom_locale, nullptr, ui::ResourceBundle::LOAD_COMMON_RESOURCES);
   auto* cmd_line = base::CommandLine::ForCurrentProcess();
@@ -263,7 +265,11 @@ void ElectronMainDelegate::PostEarlyInitialization(bool is_running_tests) {
   LoadResourceBundle(custom_locale);
 
   ElectronBrowserClient::SetApplicationLocale(
-      l10n_util::GetApplicationLocale(custom_locale));
+      l10n_util::GetApplicationLocale(custom_locale) 
+#if defined(OS_LINUX)
+      + "," + preferred_locale
+#endif
+  );
 }
 
 void ElectronMainDelegate::PreSandboxStartup() {
```

this does the same as `language_util_linux`, just quite early in the app startup that would avoid a blocking call. We don't have to refer it separately in `webContents` separately, the value set in `ElectronBrowserClient::SetApplicationLocale` will be reflected via `g_browser_process->GetApplicationLocale` when you access it.

sorah:
I don't think we have to check both `GetApplicationLocale("")` and `GetApplicationLocale(custom_locale)`  as custom_locale is brought by user (`--lang`). My initial intention to use `GetApplicationLocale("")` was a attempt to return something meaningful rather than empty. `util` codes returning empty sound unnatural to me; from my perspective, returning empty is fine.

deepak1556:
Thanks for the context. In that case the current state of PR is good.

sorah:
thank you for treating this patch as a fix! :hugs:

PR #23261, opened 2020-04-23
fix: isolate callbacks in node_bindings

codebytere:
#### Description of Change

While some aspects of embedding are necessarily different between emulated Node.js and Node.js as it's embedded in the main and renderer processes of Electron, there's been some skew recently in our embedding processes resultant of the fact that we only test some of Node's functionality in emulated mode and therefore miss potential holes in non-emulated mode.

In the future, we should seek to test these more effectively, but for now we should at least close that gap by ensuring that isolate setup is better executed in non-emulated mode. There remains some work to be done here around `async_hooks` (see https://github.com/electron/electron/pull/23213) but that will require a little more investigation to determine what of `SetupIsolateForNode` can be safely executed in a non-emulated environment without causing collateral damage.

cc @nornagon @ckerr @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed some Wasm and diagnostics issues in main and renderer process execution of Node.js

### DIFF
--- a/shell/common/node_bindings.cc
+++ b/shell/common/node_bindings.cc
@@ -219,6 +219,23 @@ void SetNodeOptions(base::Environment* env) {
   }
 }
 
+void HostCleanupFinalizationGroupCallback(
+    v8::Local<v8::Context> context,
+    v8::Local<v8::FinalizationGroup> group) {
+  node::Environment* env = node::Environment::GetCurrent(context);
+  if (env == nullptr) {
+    return;
+  }
+  env->RegisterFinalizationGroupForCleanup(group);
+}
+
+bool AllowWasmCodeGenerationCallback(v8::Local<v8::Context> context,
+                                     v8::Local<v8::String>) {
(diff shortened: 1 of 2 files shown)


shell/common/node_bindings.cc
```diff
@@ -394,15 +411,33 @@ node::Environment* NodeBindings::CreateEnvironment(
   }
 
   if (browser_env_ == BrowserEnvironment::BROWSER) {
-    // SetAutorunMicrotasks is no longer called in node::CreateEnvironment
-    // so instead call it here to match expected node behavior
+    // This policy ensures microtask checkpoints are explicitly invoked.
```
nornagon, line 414: I don't think that's what this policy does...? Rather, it _doesn't_ automatically invoke any microtask checkpoints and instead leaves it up to the embedder to invoke them.

It might also be useful to mention that this is what Node does & thus is required by Node.

codebytere: you're right and is what i meant but mangled explaining - i'll clarify the phrasing :)

PR #23271, opened 2020-04-24
fix: fetch fresh frame_host reference on use to avoid holding destroyed references

pfrazee:
#### Description of Change

Fixes https://github.com/electron/electron/issues/22513

Previously `NetworkHintsHandlerImpl` acquired a `content::RenderFrameHost` pointer on creation. This pointer lived longer than the lifetime of the render host in some cases, leading to a bad access on certain navigation events.

EDIT:
The original PR updated `NetworkHintsHandlerImpl` to hold the process & frame ID instead and then acquire a fresh pointer at time-of-access in order to better track the lifetime of the render_host. However, this required iterating the frame tree on each event-emit, which was inefficient.

The final PR updates `NetworkHintsHandlerImp` to hold a browser-context pointer instead, which solves the lifetime issues.

cc @nornagon as stakeholder (thanks for helping me solve this)

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash which could occur during page navigations

### DIFF
--- a/shell/browser/network_hints_handler_impl.h
+++ b/shell/browser/network_hints_handler_impl.h
@@ -9,7 +9,8 @@
 
 namespace content {
 class RenderFrameHost;
-}
+class BrowserContext;
+}  // namespace content
 
 class NetworkHintsHandlerImpl
     : public network_hints::SimpleNetworkHintsHandlerImpl {
@@ -27,7 +28,7 @@ class NetworkHintsHandlerImpl
  private:
   explicit NetworkHintsHandlerImpl(content::RenderFrameHost*);
 
-  content::RenderFrameHost* render_frame_host_ = nullptr;
+  content::BrowserContext* browser_context_ = nullptr;
 };
 
 #endif  // SHELL_BROWSER_NETWORK_HINTS_HANDLER_IMPL_H_
--- a/shell/browser/network_hints_handler_impl.cc
+++ b/shell/browser/network_hints_handler_impl.cc
@@ -21,7 +21,7 @@ NetworkHintsHandlerImpl::NetworkHintsHandlerImpl(
     : network_hints::SimpleNetworkHintsHandlerImpl(
           frame_host->GetProcess()->GetID(),
           frame_host->GetRoutingID()),
-      render_frame_host_(frame_host) {}
+      browser_context_(frame_host->GetProcess()->GetBrowserContext()) {}
 
 NetworkHintsHandlerImpl::~NetworkHintsHandlerImpl() = default;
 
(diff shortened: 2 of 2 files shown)


deepak1556:
I was bit confused about the lifetime of `NetworkHintsHandlerImpl` here, because `WebPrescientNetworkingImpl` on the renderer which is the Remote end of the pipe has its lifetime associated with a `RenderFrame` and `NetworkHintsHandlerImpl` is strongly binded to the mojo pipe, so if the connection is terminated on one end of the pipe it should have deleted the interface on the browser side.

I expect the `RenderFrameHost` to be alive until the unload handler is run on the corresponding `RenderFrame` but there is a case when lifetime of `RenderFrameHost` is managed a bit differently. It is when cross-site navigations are happening with site isolated child `RenderFrameHost` in the `FrameTreeNode` tree. https://goo.gl/85KSff explains it clearly.

So this issue can only happen when there is a cross-site navigation, in which case we should do one of the following for the browser interfaces defined for a frame,

1) If the interface is managed by a service with its own lifetime, then no action is necessary
2) If the interface is bound to the lifetime of the mojo connection then we should ensure it isn't coupled tightly to a  `RenderFrameHost`

Coming back to this specific case, the `RenderFrameHost` is only used to obtain the `BrowserContext` for event emission, but since the event doesn't carry any frame specific details and unless its going to be added, I think it should be okay to cache the `BrowserContext` in `NetworkHintsHandlerImpl` class.

pfrazee:
@deepak1556 I'm only scratching the surface of understanding here, so apologies if this is obvious -

Is there any reason not to always do what this PR does, hold onto the IDs and then fetch the pointer at the time of usage?

deepak1556:
@pfrazee what you are doing in this PR is the right way.

But in the end the event data emitted in the `Preconnect` handler has nothing associated with a frame, its just has data specific to a session. So what I was suggesting is to have something like

```
NetworkHintsHandlerImpl::NetworkHintsHandlerImpl(
     content::RenderFrameHost* frame_host)
    : network_hints::SimpleNetworkHintsHandlerImpl(
               frame_host->GetProcess()->GetID(),
               frame_host->GetRoutingID()),
       browser_context_(frame_host->GetProcess()->GetBrowserContext()) {}

void NetworkHintsHandlerImpl::Preconnect(const GURL& url,
                                         bool allow_credentials) {

     if (browser_context_) {

    } 
}
```


which would avoid iterating over the frame tree unnecessarily for every preconnect signal. But if the event had any frame specific data, I wouldn't suggest this route :)

pfrazee:
Got it! I'll give that a run

shell/browser/network_hints_handler_impl.h
```diff
@@ -6,6 +6,7 @@
 #define SHELL_BROWSER_NETWORK_HINTS_HANDLER_IMPL_H_
 
 #include "components/network_hints/browser/simple_network_hints_handler_impl.h"
+#include "content/public/browser/browser_context.h"
```
deepak1556, line 9: Header is not needed, it should be fine to forward declare raw pointers.

```
namespace content {
class BrowserContext;
....
}
```

pfrazee: Done!

deepak1556:
@pfrazee can you rebase the PR on master to fix the lint errors. Thanks!

pfrazee:
Done! Not sure why the release-notes check is failing, I think it's correct

deepak1556:
It was missing the `Notes:` hint, fixed it.

PR #23301, opened 2020-04-27
fix: do not add 104 to the history length

MarshallOfSound:
I don't wanna talk about it.

Notes: no-notes


nornagon:
does this need backports?

MarshallOfSound:
@nornagon I'm fixing it in the backports of the PR that accidentally introduced it

Issue #23282, opened 2020-04-25
Crash when pressing ALT and setMenu is called with Menu.buildFromTemplate([])


dezsiszabi:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v8.2.3

* **Operating System:**
  * Windows 10 (64 bit)
* **Last Known Working Electron version:**
  * N/A
### Expected Behavior

Doesn't crash
### Actual Behavior

Crashes
### To Reproduce

```
const { app, BrowserWindow, Menu } = require('electron');

function createWindow () {
  let win = new BrowserWindow();
  win.setMenu(Menu.buildFromTemplate([]));
  win.loadFile('index.html');
}

app.whenReady().then(createWindow);
```

Press ALT in the window that pops up. Result: electron.exe crashes. Probably `Menu.buildFromTemplate` shouldn't be called with an empty array... BUT, this still shouldn't crash the application in my opinion.

### Call stack from crashdump

```
electron.exe!views::FocusSearch::FindNextFocusableViewImpl(views::View * starting_view, views::FocusSearch::StartingViewPolicy check_starting_view, bool can_go_up, bool can_go_down, views::FocusSearch::AnchoredDialogPolicy can_go_into_anchored_dialog, int skip_group_id, base::internal::flat_tree<views::View *,views::View *,base::internal::GetKeyFromValueIdentity<views::View *>,std::__1::less<void>> * seen_views, views::FocusTraversable * * focus_traversable, views::View * * focus_traversable_view) Line 186	C++
electron.exe!views::FocusSearch::FindNextFocusableView(views::View * starting_view, views::FocusSearch::SearchDirection search_direction, views::FocusSearch::TraversalDirection traversal_direction, views::FocusSearch::StartingViewPolicy check_starting_view, views::FocusSearch::AnchoredDialogPolicy can_go_into_anchored_dialog, views::FocusTraversable * * focus_traversable, views::View * * focus_traversable_view) Line 63	C++
[Inline Frame] electron.exe!views::AccessiblePaneView::GetFirstFocusableChild() Line 138	C++
[...]

PR #23308, opened 2020-04-28
fix: prevent Menu.buildFromTemplate with empty array

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/23282.

Prevent issues with menu creation and subsequent pane focus from menu bar by preventing menus from being created from an empty array. I can't conceive a valid use case for this, since if one wants to remove a menu they should be be passing `null` to `win.setMenu()` or calling `win.removeMenu()`. This issue is also specific to top-level menus, and not submenus, so the new check and exception is scoped to top-level menus.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a potential crash when menu is created from an empty template.

### DIFF
--- a/lib/browser/api/menu-item.js
+++ b/lib/browser/api/menu-item.js
@@ -16,7 +16,7 @@ const MenuItem = function (options) {
   }
   this.submenu = this.submenu || roles.getDefaultSubmenu(this.role);
   if (this.submenu != null && this.submenu.constructor !== Menu) {
-    this.submenu = Menu.buildFromTemplate(this.submenu);
+    this.submenu = Menu.buildFromTemplate(this.submenu, true);
   }
   if (this.type == null && this.submenu != null) {
     this.type = 'submenu';
--- a/lib/browser/api/menu.js
+++ b/lib/browser/api/menu.js
@@ -173,13 +173,17 @@ Menu.setApplicationMenu = function (menu) {
   }
 };
 
-Menu.buildFromTemplate = function (template) {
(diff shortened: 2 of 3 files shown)


zcbenz:
Would the crash in original issue still happen when calling `win.setMenu(new Menu)`?

dezsiszabi:
Just to give a usecase.

We have a wrapper package around electron window creation/menu creation. It exposes a WindowBuilder, TrayBuilder, MenuBuilder etc (using a fluent API/builder pattern to compose the electron application).

In this wrapper code we called a code similar to

```
window.setMenu(Menu.buildFromTemplate(this.menuBuilder.map(mb => mb.build())));
```

Of course the issue should be prevented in our code by adding defensive checks and not call `setMenu` when there are no menu builders.

That said, I'd still prefer an early failure which doesn't result in a crash.

Either:

1. Fail early by throwing an exception if `Menu.buildFromTemplate` is called with an empty array
OR
2. `Menu.buildFromTemplate` should return `null` if called with an empty array.

In both cases the documentation should mention the behavior in case `Menu.buildFromTemplate` is called with an empty array.

codebytere:
@zcbenz it shouldn't from my read of [this function](https://github.com/electron/electron/blob/6159066c261a286a2e7ac4bb3cff3800c0a6d2b3/shell/browser/api/electron_api_top_level_window.cc#L685-L700)

codebytere:
New information came to light about how this may break existing codepaths - i'm going to open a new PR that avoids this breakage and backport that.

PR #23311, opened 2020-04-28
feat: add enableWebSQL webpreference

deepak1556:
#### Description of Change

Allows to disable websql api in a backward compatible way

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: enableWebSQL is a new webpreference option to enable/disable websql api

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -385,6 +385,8 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       visible to users.
     * `spellcheck` Boolean (optional) - Whether to enable the builtin spellchecker.
       Default is `true`.
+    * `enableWebSQL` Boolean (optional) - Whether to enable the [WebSQL api](https://www.w3.org/TR/webdatabase/).
+      Default is `true`.
 
 When setting minimum or maximum window size with `minWidth`/`maxWidth`/
 `minHeight`/`maxHeight`, it only constrains the users. It won't prevent you from
--- a/shell/renderer/content_settings_observer.cc
+++ b/shell/renderer/content_settings_observer.cc
@@ -5,6 +5,7 @@
 #include "shell/renderer/content_settings_observer.h"
 
 #include "content/public/renderer/render_frame.h"
+#include "shell/common/options_switches.h"
 #include "third_party/blink/public/platform/url_conversion.h"
 #include "third_party/blink/public/platform/web_security_origin.h"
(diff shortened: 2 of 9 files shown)


docs/api/browser-window.md
```diff
@@ -385,6 +385,8 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       visible to users.
     * `spellcheck` Boolean (optional) - Whether to enable the builtin spellchecker.
       Default is `true`.
+    * `enableWebSQL` Boolean (optional) - Whether to enable the [WebSQL api](https://www.w3.org/TR/webdatabase/).
+      Default is `true`.
```
nornagon, line 389: Do we plan to set the default to false in a followup?

shell/renderer/content_settings_observer.cc
```diff
@@ -20,6 +21,11 @@ ContentSettingsObserver::ContentSettingsObserver(
 ContentSettingsObserver::~ContentSettingsObserver() = default;
 
 bool ContentSettingsObserver::AllowDatabase() {
+  if (!base::CommandLine::ForCurrentProcess()->HasSwitch(
+          switches::kEnableWebSQL)) {
+    return false;
+  }
```
nornagon, line 27: Is this a purely renderer-side check? Is there any way we can disable this browser-side as well?

docs/api/browser-window.md
deepak1556: Thats up for discussion, I didn't know if we wanted to keep it enabled by default upto a certain electron version and then disable it ?

shell/renderer/content_settings_observer.cc
deepak1556: We can't disable on browser side, it only manages quota for storage and establishes connection to SQLite https://source.chromium.org/chromium/chromium/src/+/master:storage/browser/database/

The renderer side check is pretty consistent and powerful.

There is flag `--disable-databases` https://source.chromium.org/chromium/chromium/src/+/master:content/public/common/content_switches.cc;l=116 but it would disable both indexedDB and webSQL

deepak1556:
Thanks @nornagon , but we don't currently expose a way to set per RenderFrame flags from the browser process. We have to expose the `RenderFrameHost` as a property on the `webContents` to achieve it or some other similar mechanism. 

Is there another approach you have in mind ?

nornagon:
Not currently, I think this is something @loc is exploring.

deepak1556:
Cool, thanks!

PR #23321, opened 2020-04-28
fix: do not leak IPC or context bridge promises

MarshallOfSound:
This fixes three main memory leaks to do with our promise helper mostly:

* `IPCRenderer` would hold pending promises for `invoke` calls even after the render frame was removed, those promises had a strong ref to the `Context` so the context was never freed
* `contextBridge` proxy promise had a similar issue with heap allocating the promise, if it was never resolved or rejected it would never be released, thus never releasing the strong ref to `Context`
* And a final leak again in the proxy promise impl where we passed strong global of source and destination context into the then and catch functions, if the promise was left pending the globals would not be freed. These have been converted to weak refs.

Notes: Fixed memory leaks in sandbox mode when using `contextBridge` with promises or `ipcRenderer.invoke`

### DIFF
--- a/shell/renderer/api/electron_api_context_bridge.cc
+++ b/shell/renderer/api/electron_api_context_bridge.cc
@@ -5,6 +5,7 @@
 #include "shell/renderer/api/electron_api_context_bridge.h"
 
 #include <map>
+#include <memory>
 #include <set>
 #include <string>
 #include <utility>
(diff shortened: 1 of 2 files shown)


shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -195,53 +196,69 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
     v8::Context::Scope destination_scope(destination_context);
     {
       auto source_promise = v8::Local<v8::Promise>::Cast(value);
-      auto* proxied_promise = new gin_helper::Promise<v8::Local<v8::Value>>(
-          destination_context->GetIsolate());
+      std::shared_ptr<gin_helper::Promise<v8::Local<v8::Value>>>
```
nornagon, line 199: a comment about why this is a `shared_ptr` would be 💯

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -195,53 +196,69 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
...
+          proxied_promise(new gin_helper::Promise<v8::Local<v8::Value>>(
+              destination_context->GetIsolate()));
       v8::Local<v8::Promise> proxied_promise_handle =
           proxied_promise->GetHandle();
 
+      v8::Global<v8::Context> global_then_source_context(
+          source_context->GetIsolate(), source_context);
+      v8::Global<v8::Context> global_then_destination_context(
+          destination_context->GetIsolate(), destination_context);
```
nornagon, line 208: This is redundant with the promise's context right? Should we expose a reference to the promise helper's context?

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -195,53 +196,69 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
...
+      global_then_source_context.SetWeak();
+      global_then_destination_context.SetWeak();
       auto then_cb = base::BindOnce(
-          [](gin_helper::Promise<v8::Local<v8::Value>>* proxied_promise,
+          [](std::shared_ptr<gin_helper::Promise<v8::Local<v8::Value>>>
+                 proxied_promise,
              v8::Isolate* isolate,
              v8::Global<v8::Context> global_source_context,
```
nornagon, line 215: The source context should be the current context when `then_cb` is invoked right?

shell/renderer/api/electron_api_ipc_renderer.cc
```diff
@@ -36,27 +37,41 @@ RenderFrame* GetCurrentRenderFrame() {
...
   }
 
+  void OnDestruct() override { electron_browser_ptr_.reset(); }
+
+  void WillReleaseScriptContext(v8::Local<v8::Context> context,
+                                int32_t world_id) override {
+    if (weak_context_.IsEmpty() ||
+        weak_context_.Get(context->GetIsolate()) == context)
+      electron_browser_ptr_.reset();
+  }
```
nornagon, line 68: If we make the promise helper's context reference weak, is this still needed?

MarshallOfSound: _maybe_, not sure, and it's probably more safe / correct to free this at the right time anyway

shell/renderer/api/electron_api_context_bridge.cc
MarshallOfSound: I don't think the promise should expose it's context, I'd rather the promise didn't hold a context at all 😆

shell/renderer/api/electron_api_context_bridge.cc
MarshallOfSound: That's making some assumptions about which context the promise was called from which isn't always gonna be the main world, E.g. an iframe could call it

PR #23331, opened 2020-04-29
refactor: use Node.js isolate setup in node mode

codebytere:
Backport of https://github.com/electron/electron/pull/23213.

See that PR for more details.

Notes: none.


codebytere:
Looks like we can't backport this owing to `node::IsolateSettings` not yet having been exposed. Ah well.

PR #23339, opened 2020-04-29
fix: do not leak IPC or context bridge promises (#23321)

MarshallOfSound:
Backport of #23321

See that PR for details.

Notes: Fixed memory leaks in sandbox mode when using `contextBridge` with promises or `ipcRenderer.invoke`

### DIFF
--- a/shell/renderer/api/electron_api_context_bridge.cc
+++ b/shell/renderer/api/electron_api_context_bridge.cc
@@ -4,6 +4,7 @@
 
 #include "shell/renderer/api/electron_api_context_bridge.h"
 
+#include <memory>
 #include <set>
 #include <string>
 #include <utility>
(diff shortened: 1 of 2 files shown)


shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -371,7 +391,8 @@ v8::Local<v8::Value> ProxyFunctionWrapper(
             !mate::ConvertFromV8(args->isolate(), message->Get(),
                                  &error_message)) {
           error_message =
-              "An unknown exception occurred in the isolated context, an error "
+              "An unknown exception occurred in the isolated context, an "
+              "error "
               "occurred but a valid exception was not thrown.";
```
nornagon, line 396: interesting choice of indentation

PR #23342, opened 2020-04-30
fix: ensure guest-embedder map is updated when webview is removed

deepak1556:
#### Description of Change

There are use cases of webview where the container holding the webview is not
actually destroyed first, instead just webview gets removed from DOM, in such
situations the browser process map is not updated accordingly and holds reference
to stale guest contents, and any window operations like scroll, resize or keyboard
events that has to chain through browser embedder will lead to UAF crash.

Ref:

https://github.com/microsoft/vscode/issues/92420
https://github.com/microsoft/vscode/issues/96492

I was unable to isolate a test case to add for this.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: fix crash with webview during some window management events like resize, scroll etc.

### DIFF
--- a/lib/browser/guest-view-manager.js
+++ b/lib/browser/guest-view-manager.js
@@ -251,6 +251,9 @@ const attachGuest = function (event, embedderFrameId, elementInstanceId, guestIn
 // Remove an guest-embedder relationship.
 const detachGuest = function (embedder, guestInstanceId) {
   const guestInstance = guestInstances[guestInstanceId];
+
+  if (!guestInstance) return;
+
   if (embedder !== guestInstance.embedder) {
     return;
   }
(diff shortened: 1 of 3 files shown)


pelmers:
thanks once more for the quick fix @deepak1556 
Do you know when to expect a new version of Electron 7.2 containing this patch to be released?

deepak1556:
Not sure of the timeline, but there will be one out sometime this week.

Linskeyd:
@deepak1556 thanks again for the fix here.  Any updates on when a new 7.2 release will be made? Looks like it is these 14 commits (which includes this one) that haven't been released as a part of 7.2 yet: https://github.com/electron/electron/compare/v7.2.4...7-2-x

bestander:
This is be released in 7.3.0 https://github.com/electron/electron/commits/v7.3.0/lib/browser/guest-view-manager.js

But looks like 7.2.4 is the last 7.2 release?

bestander:
I guess for vscode it is not relevant since it's switched to 8.3 now?

deepak1556:
vscode has reverted to Electron 7 for the upcoming release due to performance concerns with Electron 8. I have updated the version to 7.3.0 to pick up this fix.

PR #23381, opened 2020-05-01
feat: app.getPath('recent')

felixrieseberg:
#### Description of Change

This PR introduces `app.getPath('recent')`, which returns `FOLDERID_recent`, more commonly known as `CSIDL_RECENT`. It's the folder where Windows keeps all its recent files. If you're building an app, `SHGetFolderPath` is your only way to get this path - there is no environment variable, little PowerShell script, or other simple solution to get it. C++ or FFI is the only way.

#### Why?
Given that we're already offering developers a handy way to get `Desktop`, `Pictures`, `Videos`, or `Music` locations, it makes sense that developers expect that they can get other special folders with `getPath()`. You probably guessed it, but this PR fills an actual need in our app. 

If we don't add it, we're asking devs to write a full native Node module to get this path, which feels insanely cumbersome. 

This PR adds lines, but I don't expect any forced maintenance burden - this code hasn't changed on in Windows since we've first had Windows.

#### Why not?

On the other hand, I'd understand if we don't want to keep adding folders. Windows has [many of them](https://docs.microsoft.com/en-us/previous-versions/bb762584(v=vs.85)), and this does feel like a decent chunk of code in our core files for a single method.

#### Alternatives

@nornagon and I briefly discussed using a [PathServiceProvider](https://source.chromium.org/chromium/chromium/src/+/master:base/path_service.h;l=23?q=pathservice&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2F), but requirements on _when_ we'd need to register that thing made us believe that this solution is less involved.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added the ability to get the "Recent" folder on Windows with `app.getPath('recent')`

### DIFF
--- a/shell/browser/api/electron_api_app.cc
+++ b/shell/browser/api/electron_api_app.cc
@@ -917,9 +917,10 @@ base::FilePath App::GetPath(gin_helper::ErrorThrower thrower,
 
 #if defined(OS_WIN)
     // If we get the "recent" path before setting it, set it
-    if (!succeed && name == "recent") {
-      SetRecentPath(thrower, base::Optional<base::FilePath>());
-      succeed = base::PathService::Get(key, &path);
+    if (!succeed && name == "recent" &&
+        platform_util::GetFolderPath(DIR_RECENT, &path)) {
+      base::ThreadRestrictions::ScopedAllowIO allow_io;
+      succeed = base::PathService::Override(DIR_RECENT, path);
     }
 #endif
   }
--- a/shell/browser/api/electron_api_app.cc
+++ b/shell/browser/api/electron_api_app.cc
@@ -54,6 +54,7 @@
 #include "shell/common/gin_helper/object_template_builder.h"
 #include "shell/common/node_includes.h"
 #include "shell/common/options_switches.h"
+#include "shell/common/platform_util.h"
 #include "ui/gfx/image/image.h"
 
 #if defined(OS_WIN)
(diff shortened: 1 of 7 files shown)


felixrieseberg:
I believe the failed tests are flakes.

shell/common/platform_util_win.cc
```diff
@@ -392,6 +393,25 @@ bool MoveItemToTrash(const base::FilePath& path, bool delete_on_fail) {
          SUCCEEDED(pfo->PerformOperations());
 }
 
+bool GetFolderPath(int key, base::FilePath* result) {
+  wchar_t system_buffer[MAX_PATH];
+  system_buffer[0] = 0;
+
+  base::FilePath cur;
```
deepak1556, line 400: The temp variable is not required, better to populate `*result` directly.

shell/common/platform_util_win.cc
```diff
@@ -392,6 +393,25 @@ bool MoveItemToTrash(const base::FilePath& path, bool delete_on_fail) {
          SUCCEEDED(pfo->PerformOperations());
 }
 
+bool GetFolderPath(int key, base::FilePath* result) {
+  wchar_t system_buffer[MAX_PATH];
+  system_buffer[0] = 0;
```
deepak1556, line 398: Shouldn't it be okay to have the buffer uninitialized ?

felixrieseberg: You'd think, but that's how Chromium does it. Happy to leave it uninitialized though!

deepak1556: No problem in having it but if the call the failed it gonna point to some empty string and we don't use the result in the case, so its not necessary to initialize it here.

shell/browser/api/electron_api_app.cc
```diff
@@ -871,6 +876,30 @@ void App::SetAppLogsPath(gin_helper::ErrorThrower thrower,
 }
 #endif
 
+#if defined(OS_WIN)
+void App::SetRecentPath(gin_helper::ErrorThrower thrower,
```
deepak1556, line 880: Method is unused, can be removed.

Issue #23102, opened 2020-04-13
Can't set Referer header


sd-soleaio:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 8.2.1
* **Operating System:**
  * macOS 10.14.6

### Expected Behavior
The Referer header should be able to be set on ClientRequest objects via request.setHeader(name, value).

### Actual Behavior
Referer header isn't being set at all (headers can be viewed by using a proxy like Charles or Fiddler).

### To Reproduce
```const { app } = require('electron')
app.on('ready', () => {
  const { net } = require('electron')
  const request = net.request('https://github.com')
  request.on('response', (response) => {
    console.log(response.statusCode})
    console.log(JSON.stringify(response.headers))
    response.on('data', (chunk) => {
      console.log(chunk)
    })
    response.on('end', () => {
      console.log('No more data in response.')
    })
  })

  # this referer header should be set
  request.setHeader('referer', 'https://github.com')

  request.end()
})
```

### Additional Information
There does seem to be a Referrer property on the request that can be seen when using webRequest.onBeforeRequest but as far as I can see it's always empty and cannot be modified.

Being able to set the referer header is a basic and fundamental requirement for working with some websites, would be great if this issue could be fixed.

Massively appreciate all the developers taking the time working and updating this project.

theyezzyman:
Electron team fix this sh*t so bots can eat

We need that referer header to cook :(

LuoJinghua:
We got this problem too which is blocking our product from working.

LuoJinghua:
The referer of details is blank, is there any way to set this property for the request?

jjason685:
+1

mitchdavies:
+1

walmat:
Seems to be fixed in v8.x.x and up. v7 broke a lot of functionality with `net` that this was more than likely related to.

nornagon:
Closing as this is fixed in 8.x. v7 is receiving security fixes only.

PR #23386, opened 2020-05-02
fix: Don't ignore the referer header in net.request

### LINKED ISSUE #22411: Cant't set referer header when using `net.request`
### Preflight Checklist

### Issue Details

* **Electron Version:**
  8.0.1
* **Operating System:**
  macOS 10.14.6

### Expected Behavior
Send a request with `referer` header. It's worked with Node's `http/https` module, but failed when I change to `net.request`.

### Actual Behavior
No referer header

### To Reproduce
```js
ses.protocol.interceptStreamProtocol("https", (request, callback) => {
	const resp = sendRequestWithNetRequest({
       url: request.url,
       method: request.method,
	   headers: {
			...request.headers,
			referer: request.referrer  // only referrer has been ignored!!
       }
    })
    callback({
        data: resp.body,
        statusCode: resp.statusCode,
        headers: resp.headers
      });
});
```

LuoJinghua:
#### Description of Change
Don't ignore the referrer header for a request, fixes #22411 #23102.

@zcbenz @MarshallOfSound 

#### Checklist

#### Release Notes

Notes: Don't ignore the referrer header in net.request

### DIFF
--- a/shell/browser/api/electron_api_url_loader.cc
+++ b/shell/browser/api/electron_api_url_loader.cc
@@ -352,6 +352,7 @@ gin::Handle<SimpleURLLoaderWrapper> SimpleURLLoaderWrapper::Create(
   request->force_ignore_site_for_cookies = true;
   opts.Get("method", &request->method);
   opts.Get("url", &request->url);
+  opts.Get("referrer", &request->referrer);
   std::map<std::string, std::string> extra_headers;
   if (opts.Get("extraHeaders", &extra_headers)) {
     for (const auto& it : extra_headers) {
--- a/typings/internal-ambient.d.ts
+++ b/typings/internal-ambient.d.ts
@@ -58,6 +58,7 @@ declare namespace NodeJS {
     body: Uint8Array | BodyFunc;
     session?: Electron.Session;
     partition?: string;
+    referrer?: string;
   }
   type ResponseHead = {
     statusCode: number;
--- a/lib/browser/api/net.ts
+++ b/lib/browser/api/net.ts
@@ -386,6 +386,7 @@ class ClientRequest extends Writable implements Electron.ClientRequest {
       }
       return ret;
     };
+    this._urlLoaderOptions.referrer = this._urlLoaderOptions.extraHeaders.referer || '';
     const opts = { ...this._urlLoaderOptions, extraHeaders: stringifyValues(this._urlLoaderOptions.extraHeaders) };
     this._urlLoader = createURLLoader(opts);
     this._urlLoader.on('response-started', (event, finalUrl, responseHead) => {
--- a/spec-main/api-net-spec.ts
+++ b/spec-main/api-net-spec.ts
@@ -1253,6 +1253,38 @@ describe('net module', () => {
         setTimeout(resolve, 50);
       });
     });
+
+    it('should remove the referer header when no referrer url specified', async () => {
+      const serverUrl = await respondOnce.toSingleURL((request, response) => {
+        expect(request.headers.referer).to.equal(undefined);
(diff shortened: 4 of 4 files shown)


LuoJinghua:
Done, please review my changes again.

LuoJinghua:
@nornagon @zcbenz any comment for my latest change?

shell/browser/api/electron_api_url_loader.cc
```diff
@@ -352,6 +352,7 @@ gin::Handle<SimpleURLLoaderWrapper> SimpleURLLoaderWrapper::Create(
   request->force_ignore_site_for_cookies = true;
   opts.Get("method", &request->method);
   opts.Get("url", &request->url);
+  opts.Get("referrer", &request->referrer);
```
MarshallOfSound, line 355: Is there a reason this has to be a new `option` and we can't just read it off the `headers` object if it's defined?  That would make it easier to use I think / be more expected that the header _just works_ ™️

LuoJinghua: En, this way works too and less burden for user, but the current approach matches the window.fetch() better and even better if we'll introduce the referrer policy later. I don't have opinion which one is better.

nornagon: I think we should probably go with what @MarshallOfSound said and read this out of the headers option.

This API is designed to be similar to the NodeJS http API (though it doesn't match it in every respect). In the NodeJS API, setting the `Referer` header is done through the `headers` object (or `setHeader`), so it would be more consistent with NodeJS for us to respect that here. Additionally, the two linked bugs in the PR description show two examples of how people _expect_ the API to work, which is similarly to the NodeJS API. I think we can avoid confusion by matching NodeJS here.

LuoJinghua: Ok, I just pushed the branch forcely and changed the code to read the referrer url from headers instead.

typings/internal-ambient.d.ts
```diff
@@ -58,6 +58,7 @@ declare namespace NodeJS {
     body: Uint8Array | BodyFunc;
     session?: Electron.Session;
     partition?: string;
+    referrer?: string;
```
jkleinsc, line 61: This change is no longer needed.

LuoJinghua: But I got build failures without this change.

nornagon: @jkleinsc this is in the internal typings

0xdeadbae:
I have the same issue on Windows. Electron 8.2.5.

Does this fix also solve the issue for Windows user?

LuoJinghua:
> I have the same issue on Windows 1909. Electron 8.2.5.
> 
> Does this fix also solve the issue for Windows user?

Yes, this pull request fixes the issue for all platforms.

typings/internal-ambient.d.ts
LuoJinghua: I have added this line back to fix the build failure.

LuoJinghua:
Thank you for merging，how to backport this to stable branches atomatically？

nornagon:
welp

LuoJinghua:
Should be backport the 7e841ceb to stable branches first? If we don't want backport it, I think I have to backport this pull manually.

nornagon:
I don't think we should backport 7e841ce to older branches, please go ahead and create manual backports.

(there's no need to backport to 7.x since this was only an issue since the 8.x branch)

LuoJinghua:
Ok. I'm going to backport this manually.

PR #23307, opened 2020-04-28
fix: build when some buildflags are disabled

deepak1556:
#### Description of Change

Fix build when the following flags are disabled

```
enable_basic_printing = false
enable_builtin_spellchecker = false
enable_desktop_capturer = false
enable_osr = false
enable_pdf = false
enable_pdf_viewer = false
enable_pepper_flash = false
enable_plugins = false
enable_print_preview = false
enable_widevine = false
enable_extensions = false
enable_electron_extensions = false
```

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: no-notes

PR #23404, opened 2020-05-04
fix: guard pdf behind correct buildflags

deepak1556:
#### Description of Change

Follow-up to https://github.com/electron/electron/pull/23307

Missed a case where pdf flags are disabled but extension system is enabled. We can never have the other case because of the dependency and also its explicitly guarded https://github.com/electron/electron/blob/master/BUILD.gn#L64

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: no-notes


deepak1556:

PR #23414, opened 2020-05-05
fix: fixed possible problem with visibility state on mac

CezaryKulakowski:
I don't know if it's possible to create Electron's app to present
the problem but in our project we added possibility to turn on/off
the frame on the window and after adding new container to the window
it becomes blank because of the wrong visibility state of the window.

Notes: fixed possible problem with visibility state on mac
#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:

### DIFF
--- a/patches/chromium/disable_hidden.patch
+++ b/patches/chromium/disable_hidden.patch
@@ -46,3 +46,16 @@ index 12e1923fd2943af92aaf6f691d28212bdd2a73dd..3b66abb180df5f3da7c4fa8de7727962
      host()->WasHidden();
      aura::WindowTreeHost* host = window_->GetHost();
      if (delegated_frame_host_) {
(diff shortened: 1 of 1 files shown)


CezaryKulakowski:
" there normally isn't a good reason to need to patch Chromium like this" I couldn't agree more. But this problem is caused by Electron's patch to Chromium (this change is just an expansion of this patch). I will try to find an example application which shows the problem on clean Electron when I have time.

codebytere:
@CezaryKulakowski do you plan to pursue this or can it be closed?

CezaryKulakowski:
I didn't have time to look for app which would demonstrate the problem on clean Electron. Closing this pr: I'm going to keep this change on our side as it fixes serious bug in our app.

PR #23415, opened 2020-05-05
test: skip "handles Promise timeouts correctly" when ELECTRON_RUN_AS_NODE is disabled

miniak:
#### Description of Change
This test fails when `ELECTRON_RUN_AS_NODE` is disabled, let's skip it in this case.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: none

### DIFF
--- a/spec-main/node-spec.ts
+++ b/spec-main/node-spec.ts
@@ -308,7 +308,7 @@ describe('node feature', () => {
     expect(result.status).to.equal(0);
   });
 
-  it('handles Promise timeouts correctly', (done) => {
+  ifit(features.isRunAsNodeEnabled())('handles Promise timeouts correctly', (done) => {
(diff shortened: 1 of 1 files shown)


codebytere:
It's concerning to me that it fails, and i'd prefer to investigate the underlying cause and potential fix before considering skipping the test and removing the only signal we had for the behavior.

zcbenz:
@codebytere @nornagon This test was added to test `ELECTRON_RUN_AS_NODE` especially, and it fails when `ELECTRON_RUN_AS_NODE` is disabled because in this case the node process would behave the same with the main process.

Also the main process of Electron handles micro tasks differently from Node because of Node integration, so I don't think it is possible to make this test pass in the main process.

miniak:
@codebytere, @nornagon the test depends on `ELECTRON_RUN_AS_NODE`, there is no point running it when the feature is disabled.

PR #23419, opened 2020-05-05
fix: run Node.js at-exit callbacks in renderer proc

codebytere:
#### Description of Change

When destroying the Node.js environment, we should be running all at-exit callbacks and we were not doing so in the renderer process. This fixes that.

cc @MarshallOfSound @nornagon 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Ensured that exit callbacks are run for Node.js in the renderer process.


codebytere:
@nornagon what do you envision tests for this looking like? This is extremely difficult to test in JS and i'm not 100% sure what that would look like. All Node.js tests for this are native c++ tests.

miniak:
@codebytere I think this deserves to have release notes

PR #23433, opened 2020-05-06
chore: support --remote-debugging-pipe command line flag

pavelfeldman:
#### Description of Change

This change introduces a new `--remote-debugging-pipe` command line flag.

When passed, Electron exposes Chrome remote debugging protocol over the process Pipes 3 & 4, similarly to how Chrome itself does.

This option is handy for the debugging and testing scenarios where binding a port for the remote debugging web socket is problematic or insecure.

#### Checklist

- [x] `npm test` passes
- [x] tests are added

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/electron_browser_main_parts.cc
+++ b/shell/browser/electron_browser_main_parts.cc
@@ -439,10 +439,14 @@ void ElectronBrowserMainParts::PreMainMessageLoopRun() {
   content::WebUIControllerFactory::RegisterFactory(
       ElectronWebUIControllerFactory::GetInstance());
 
-  // --remote-debugging-port
   auto* command_line = base::CommandLine::ForCurrentProcess();
-  if (command_line->HasSwitch(switches::kRemoteDebuggingPort))
+  if (command_line->HasSwitch(switches::kRemoteDebuggingPipe)) {
+    // --remote-debugging-pipe
+    content::DevToolsAgentHost::StartRemoteDebuggingPipeHandler();
+  } else if (command_line->HasSwitch(switches::kRemoteDebuggingPort)) {
+    // --remote-debugging-port
     DevToolsManagerDelegate::StartHttpHandler();
+  }
 
 #if !defined(OS_MACOSX)
   // The corresponding call in macOS is in ElectronApplicationDelegate.
@@ -511,6 +515,7 @@ void ElectronBrowserMainParts::PostMainMessageLoopRun() {
   node_env_.reset();
 
   fake_browser_process_->PostMainMessageLoopRun();
+  content::DevToolsAgentHost::StopRemoteDebuggingPipeHandler();
 }
 
 #if !defined(OS_MACOSX)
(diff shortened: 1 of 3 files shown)


shell/browser/electron_browser_main_parts.cc
```diff
@@ -439,8 +439,11 @@ void ElectronBrowserMainParts::PreMainMessageLoopRun() {
   content::WebUIControllerFactory::RegisterFactory(
       ElectronWebUIControllerFactory::GetInstance());
 
-  // --remote-debugging-port
   auto* command_line = base::CommandLine::ForCurrentProcess();
+  // --remote-debugging-pipe
+  if (command_line->HasSwitch(switches::kRemoteDebuggingPipe))
+    content::DevToolsAgentHost::StartRemoteDebuggingPipeHandler();
```
MarshallOfSound, line 445: I can see in Chromium they have their conditions set up so that `StartRemoteDebuggingPipeHandler` and `StartRemoteDebuggingServer` aren't called in the same instance.  Is it ok that we'd launch both the pipe and the server here if both arguments are passed in?

pavelfeldman: Yes, it is totally fine. Chromium supports multiple concurrent CDP clients and they can be from both Web Sockets and Pipe. Content shell does [both](https://cs.chromium.org/chromium/src/content/shell/browser/shell_devtools_manager_delegate.cc?type=cs&q=StartRemoteDebuggingPipeHandler&sq=package:chromium&g=0&l=168). But since you asked, there is no reason to be smarter than Chrome here, so let me mimic the condition.

pavelfeldman: Done!

spec-main/chromium-spec.ts
```diff
@@ -267,6 +268,38 @@ describe('command line switches', () => {
     it('should not set an invalid locale', (done) => testLocale('asdfkl', currentLocale, done));
   });
 
+  describe('--remote-debugging-pipe switch', () => {
+    it('should expose CDP via pipe', async () => {
```
pavelfeldman, line 272: @MarshallOfSound : Now that I have two tests here, on my Linux box I can see Electron app started in a headful manner for a brief moment. Is that expected?

MarshallOfSound: @pavelfeldman Reading through these tests, yes.  They spawn the electron executable with no app argument so it will launch the default app (which will open a window)

spec-main/chromium-spec.ts
```diff
@@ -267,6 +268,38 @@ describe('command line switches', () => {
...
+      const electronPath = process.execPath;
+      const appProcess = ChildProcess.spawn(electronPath, ['--remote-debugging-pipe'], {
```
MarshallOfSound, line 274: What cleans this app? It looks like this spawned application will be left open and running even after the test ends.  Should probably use an `after` or `afterEach` to ensure that the child process is killed

pavelfeldman: Aha! I thought there was some global harness somewhere that was doing that, but I realize that i should have appProcess.kill()-d.

pavelfeldman: (not sure how you roll and whether it should be try/finally)

MarshallOfSound: Normally we scope `appProcess` outside the test and then use an `afterEach` block to do

```js
afterEach(() => {
  if (appProcess && !appProcess.killed) appProcess.kill()
});
```

PR #23436, opened 2020-05-07
chore: handle Browser.close over CDP

pavelfeldman:
#### Description of Change

This change makes Electron handle [`Browser.close`](https://chromedevtools.github.io/devtools-protocol/tot/Browser/#method-close) CDP command for automation. Automation driver can start Electron process, but it needs a graceful way of shutting it down to prevent core dump pollution. Chrome and Headless have similar logic implemented in [BrowserHandler::Close](https://cs.chromium.org/search/?q=BrowserHandler::Close&sq=package:chromium&type=cs).

#### Checklist

- [x] `npm test` passes
- [x] tests are added

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/ui/devtools_manager_delegate.cc
+++ b/shell/browser/ui/devtools_manager_delegate.cc
@@ -27,7 +27,9 @@
 #include "net/base/net_errors.h"
 #include "net/socket/stream_socket.h"
 #include "net/socket/tcp_server_socket.h"
+#include "shell/browser/browser.h"
 #include "shell/common/electron_paths.h"
+#include "third_party/inspector_protocol/crdtp/dispatch.h"
 #include "ui/base/resource/resource_bundle.h"
 
 namespace electron {
@@ -80,6 +82,8 @@ std::unique_ptr<content::DevToolsSocketFactory> CreateSocketFactory() {
       new TCPServerSocketFactory("127.0.0.1", port));
 }
 
+const char kBrowserCloseMethod[] = "Browser.close";
+
 }  // namespace
 
 // DevToolsManagerDelegate ---------------------------------------------------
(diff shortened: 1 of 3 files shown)


pavelfeldman:
`chore: support --remote-debugging-pipe` has been reviewed as https://github.com/electron/electron/pull/23433

shell/browser/ui/devtools_manager_delegate.cc
```diff
@@ -27,9 +27,13 @@
 #include "net/base/net_errors.h"
 #include "net/socket/stream_socket.h"
 #include "net/socket/tcp_server_socket.h"
+#include "shell/browser/browser.h"
 #include "shell/browser/electron_paths.h"
+#include "third_party/inspector_protocol/crdtp/dispatch.h"
 #include "ui/base/resource/resource_bundle.h"
 
+const char kBrowserCloseMethod[] = "Browser.close";
```
deepak1556, line 35: can you move this to the anonymous namespace below

shell/browser/ui/devtools_manager_delegate.cc
```diff
@@ -102,6 +106,20 @@ void DevToolsManagerDelegate::HandleCommand(
...
+  if (crdtp::SpanEquals(crdtp::SpanFrom(kBrowserCloseMethod),
+                        dispatchable.Method())) {
+    // In theory, we should respond over the protocol saying that the
+    // Browser.close was handled. But doing so requires instantiating the
+    // protocol UberDispatcher and generating proper protocol handlers.
+    // Since we only have one method and it is supposed to close Electron,
+    // we don't need to add this complexity. Should we decide to support
+    // metohds like Browser.setWindowBounds, we'll need to do it though.
+    base::PostTask(FROM_HERE, {content::BrowserThread::UI},
+                   base::BindOnce([]() { Browser::Get()->Shutdown(); }));
```
deepak1556, line 120: Based on the protocol documentation, this is supposed to be a graceful shutdown but `Browser::Get()->Shutdown()` isn't graceful when there are multiple windows with embedded `webContents`. Ideally this should be `Browser::Get()->Quit()` but it has handlers on the app side that are allowed to prevent shutdown. Also by graceful, is the window allowed to run its unload handlers ?

deepak1556:
I have merged #23433 can you rebase, thanks!

shell/browser/ui/devtools_manager_delegate.cc
pavelfeldman: In `Browser.close` graceful means not generating the core dump. I.e. browser initiates the exit rather than process is killed. No unload handlers are allowed.

shell/browser/ui/devtools_manager_delegate.cc
pavelfeldman: Done

shell/browser/ui/devtools_manager_delegate.cc
deepak1556: Thanks for the context! But in chrome's case I see they are trying to force close any open browser windows https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/lifetime/application_lifetime.cc;l=302-323 . Because quitting the message loop before cleaning up the windows seems like a path for crash, especially when there are user scripts running in the main process, simple demo https://gist.github.com/5ecb1eb5a7dca08d2681ab18595f9a45 , in the local electron build change `app.quit` to call `Browser::Get()->Shutdown()` , the app will core dump.

pavelfeldman: Ah, got it! Ok, in this case let me call `Browser::Get()->Quit()` to keep things simple. If we need to force no handlers on the app side, we can follow up. Thanks for the hints!

deepak1556: Sounds good, thanks!

PR #23460, opened 2020-05-07
[wip] feat: add hooks for profiling startup performance

ckerr:
#### Description of Change

This is a draft of some code that's been used by an internal app for awhile. Since it could be useful to other users too, this is a draft of an API that could be both generally usable and also extensible.

The documentation and tests are incomplete since this PR is still in the feedback-gathering phase...

Feedback welcomed @miniak @alexeykuzmin 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added startup profiling API.

### DIFF
--- a/docs/api/process.md
+++ b/docs/api/process.md
@@ -149,6 +149,23 @@ The time is represented as number of milliseconds since epoch. It returns null i
 
 Returns [`CPUUsage`](structures/cpu-usage.md)
 
+### `process.getProfilingMarks()`
+
+Returns an object of name/timestamp pairs that can be used in profiling an application's performance.
+
+FIXME: add more information
+
+Returns `Object`:
+
+* `main` Date
+* `basicStartupComplete` Date
+* `nodeLoadEnvironment` Date
+* `initBegin` Date
+* `initEnd` Date
+* `didCreateScriptContext` Date | null
+* `renderFrameCreated` Date | null
+* `renderThreadStarted` Date | null
+
 ### `process.getIOCounters()` _Windows_ _Linux_
 
 Returns [`IOCounters`](structures/io-counters.md)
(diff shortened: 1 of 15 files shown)


nornagon:
Hm, this seems like something that the [chromium tracing framework](https://www.electronjs.org/docs/api/content-tracing) should be able to provide.

Can you say more about the use case here?

MarshallOfSound:
In agreement with the above, would love more context on this.  As an example of us using Chromium's tracing framework you can search for these kinds of `TRACE` calls.

https://github.com/electron/electron/blob/master/shell/browser/api/electron_api_web_contents.cc#L1190

jkleinsc:
@electron/wg-api reviewed on [May-18-2020](https://github.com/electron/governance/blob/master/wg-api/meeting-notes/2020-05-18.md) and is waiting for this PR to be updated by @ckerr.

ckerr:
Just a courtesy pong since this has gotten some attention, though it wasn't meant to have. As marked above, this is a WIP that got moved back in the queue because I got some higher priority issues on my TODO list. I didn't consider this ready for prime time, which is why I didn't request @electron/wg-api review.

The context is that Teams has a downstream patch similar to this to benchmark how long specific parts take to startup and makes that information available via a JS API for consumption. This seems like it might be generally useful & not Teams-specific, so my intent was to try to upstream a more flexible variant of it.

Agree with @nornagon and @MarshallOfSound that the Tracing API could be useful here too.

nornagon:
@ckerr it's been a while since this PR was opened, do you still plan to work on this?

ckerr:
TBH I didn't realize it was still open until your bot pinged it @nornagon :slightly_smiling_face: 

This draft is pretty stale and the suggestions that you and @MarshallOfSound provided would probably be a better starting point IMO. I'm going to close this PR and revisit as necessary.

PR #23461, opened 2020-05-07
fix: crashReporter incompatible with sandbox on Linux (#23265)

nornagon:
Backport of #23265.

Notes: `crashReporter` is now explicitly initialized only in the main process, and implicitly initialized in other child processes. This fixes an issue preventing the crash reporter from functioning in sandboxed renderers on Linux.


vyunikov:
Does `crashReporter` get initialized in child processes created using `child_process` module?
If not, how can I use it in the case I mentioned?

nornagon:
@vyunikov on Linux, it does not. I don't know what use case you are talking about.

To initialize the crash reporter in a Node child process on Linux, you'll need to use `process.crashReporter.start()`. Sadly this doesn't seem to be documented anywhere, I'll update the docs soon. You can see an example [in our tests](https://github.com/electron/electron/blob/0d72bf3ed20bd897da22bcbe1b38d8b02683adf2/spec-main/fixtures/apps/crash/node-crash.js).

nornagon:
Ref https://github.com/electron/electron/pull/25579

vyunikov:
@nornagon Thanks for your response. Sorry, I was probably a bit unclear.
My app works on Mac and Windows and I need to enable crash reporting on these platforms. 
On Windows `process.crashReporter.start()` seems to be unavailable. Looks like child process only has access to those 3 `crashReporter methods`: `getParameters`, `addExtraParameter`, `removeExtraParameter`

nornagon:
`process.crashReporter.start()` is only relevant in _node child processes_ and _only on Linux_.

vyunikov:
@nornagon I am using Electron 9.2.0 and it doesn't look like `crashReporter` monitors my child process when initialized in main one. At least I see that crashes in main and renderer are reporter correctly, but there's no report for the crash in my child process.
Does Crashpad monitor child processes in this Electron version?

nornagon:
@vyunikov perhaps you could open a new issue with a small repro illustrating the issue you're having?

vyunikov:
Will do, thanks

PR #23467, opened 2020-05-08
fix: NSImageName string conversion

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/23428.

Ah, Apple. Previously, we assumed that `NSImageName`s constant names matched their underlying strings. After some digging, I determined that the string representations of `NSImageName`s do _not_ in fact match the strings
themselves and instead follow the following pattern:

 * NSImageNameActionTemplate -> "NSActionTemplate"
 * NSImageNameMultipleDocuments -> "NSMultipleDocuments"

To account for this, we can strip out "ImageName" from the passed string so that it matches the system string.

Tested with https://gist.github.com/fb71b7a822216ad886263f5a48856502.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where `startDrag` could crash if some specific strings were passed into `nativeImage.createFromImage` to create the `icon`.

### DIFF
--- a/shell/common/api/electron_api_native_image_mac.mm
+++ b/shell/common/api/electron_api_native_image_mac.mm
@@ -33,12 +33,24 @@ double safeShift(double in, double def) {
   return def;
 }
 
-gin::Handle<NativeImage> NativeImage::CreateFromNamedImage(
-    gin::Arguments* args,
-    const std::string& name) {
+gin::Handle<NativeImage> NativeImage::CreateFromNamedImage(gin::Arguments* args,
+                                                           std::string name) {
   @autoreleasepool {
     std::vector<double> hsl_shift;
+
+    // The string representations of NSImageNames don't match the strings
+    // themselves; they instead follow the following pattern:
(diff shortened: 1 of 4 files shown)


shell/common/api/electron_api_native_image_mac.mm
```diff
@@ -4,6 +4,7 @@
 
 #include "shell/common/api/electron_api_native_image.h"
 
+#include <regex>
```
deepak1556, line 7: This is a banned library in chromium https://chromium-cpp.appspot.com/ , hence the lint error.

Can do instead:

```c++
size_t pos = name.find(to_remove);
if (pos != std::string::npos) {
   name.erase(pos, to_remove.length());
}
```

codebytere:
@deepak1556 sadly no but i agree it feels bad - I did a good bit of research to try and avoid this but Apple didn't really consider a case where one would be converting strings to `NSString*` constant representations and hoping they matched :(

PR #23501, opened 2020-05-11
chore: add TRACE calls for heavy ctx bridge methods

MarshallOfSound:
This allows us to view on the performance impact of the context bridge on content tracing reports

Notes: no-notes

### DIFF
--- a/shell/renderer/api/electron_api_context_bridge.cc
+++ b/shell/renderer/api/electron_api_context_bridge.cc
@@ -149,6 +149,7 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
     context_bridge::ObjectCache* object_cache,
     bool support_dynamic_properties,
     int recursion_depth) {
+  TRACE_EVENT0("electron", "ContextBridge::PassValueToOtherContext");
   if (recursion_depth >= kMaxRecursion) {
     v8::Context::Scope source_scope(source_context);
     {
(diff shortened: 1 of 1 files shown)


shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -516,6 +517,7 @@ gin_helper::Dictionary DebugGC(gin_helper::Dictionary empty) {
 void ExposeAPIInMainWorld(const std::string& key,
                           v8::Local<v8::Object> api_object,
                           gin_helper::Arguments* args) {
+  TRACE_EVENT0("electron", "ContextBridge::ExposeAPIInMainWorld");
```
nornagon, line 520: maybe `EVENT1` and include the key?

MarshallOfSound: Good call 👍

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -348,6 +349,7 @@ v8::Local<v8::Value> ProxyFunctionWrapper(
     size_t func_id,
     bool support_dynamic_properties,
     gin_helper::Arguments* args) {
+  TRACE_EVENT_INSTANT0("electron", "ContextBridge::ProxyFunctionWrapper");
```
nornagon, line 352: Why is this an instant rather than measuring a duration?

MarshallOfSound: The duration isn't actually the cost of this method, the duration would include the duration of the V8 function actually being called.  I didn't think it was good to give the misleading impression that `ProxyFunctionWrapper` was slow when it was actually the underlying method.

Thoughts?

nornagon: That should also be visible in the trace via the separate v8 function call tracing. I think this should be a duration. Instants are for things which are instant, which this is not. It's generally the case that durations in tracing include other things, and the tracing UI is equipped to illustrate that.

MarshallOfSound: Fair enough, updated 👍

Issue #23506, opened 2020-05-11
[Discussion] Changing the defaults for nodeIntegration and contextIsolation to improve the default security posture of Electron applications


MarshallOfSound:
### Change the default of `contextIsolation` from `false` to `true`

Without `contextIsolation` any code running in a renderer process can quite easily reach into Electron internals or your preload script and perform privileged actions that you don't want arbitrary websites to be doing.

For more information on `contextIsolation`, how to enable it easily and it's security benefits please see our dedicated [Context Isolation Document](https://github.com/electron/electron/blob/master/docs/tutorial/context-isolation.md).

We're making this change to improve the **default** security of Electron apps so that your app is only insecure if you have deliberately opted in to the insecure behaviour.

#### Timeline

* Deprecate the current default of `contextIsolation` in Electron 10
* Change to the new default (`true`) in Electron 12

### Remove the `nodeIntegration` flag completely

Historically we have recommended that apps use `nodeIntegration: false` to prevent renderers from having access to Electron internals or the `require` function.  Over time it has become clear that this flag actually has negligible security impact and can easily be bypassed.  This was the original motivation for adding the `contextIsolation` flag.

We are now confident enough in the `contextIsolation` feature that we intend to remove the misleading `nodeIntegration` flag and instead strongly recommend usage of `contextIsolation`.

#### Timeline

* Deprecate the flag and instruct folks to use `contextIsolation: true` instead of `nodeIntegration: false` in Electron 10
* Remove the flag and any effect it had in Electron 12

reZach:
This is marked as discussion, so I hope I'm allowed to post, if not, please limit to just collaborators. 

I feel this is a good change, but not many are prepared for. There are a lot of packages/templates that make use of remote and do not use IPC to talk between the processes. I've built a template that's supportive of this change, but I feel a lot will be discussing once this change is implemented.

PalmerAL:
I'm not sure if this has been discussed anywhere (or if I'm not thinking through it correctly), but it does seem like there's a scenario in which:

* You're creating a webContents with `nodeIntegration: false, contextIsolation: false` in an old Electron version.
* You upgrade directly from an old version to 12, or you don't read the deprecation warnings in the intermediate versions.

The effect would be that node integration would be silently enabled in your app after the upgrade, which would potentially cause severe security issues. (I suppose that wouldn't matter if `nodeIntegration` can already be bypassed, but as far as I know it's still reasonably secure as long as you don't have a preload script).

[...]

PR #23507, opened 2020-05-11
chore: add deprecation warning for the default of contextIsolation

MarshallOfSound:
Refs: https://github.com/electron/electron/issues/23506

Notes: Deprecated the default of `contextIsolation`, it will change from `false` to `true` in a future Electron release

### DIFF
--- a/docs/breaking-changes.md
+++ b/docs/breaking-changes.md
@@ -14,6 +14,15 @@ This document uses the following convention to categorize breaking changes:
 
 ## Planned Breaking API Changes (12.0)
 
+### Default Changed: `contextIsolation` defaults to `true`
+
+In Electron 12, `contextIsolation` will be enabled by default.  To restore
+original behavior, `contextIsolation: false` must be specified in WebPreferences:
+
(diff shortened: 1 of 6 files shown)


docs/breaking-changes.md
```diff
@@ -14,6 +14,15 @@ This document uses the following convention to categorize breaking changes:
 
 ## Planned Breaking API Changes (12.0)
 
+### Default Changed: `contextIsolation` defaults to `true`
+
+In Electron 12, `contextIsolation` will be enabled by default.  To restore
+original behavior, `contextIsolation: false` must be specified in WebPreferences:
```
nornagon, line 20:
```suggestion
the previous behavior, `contextIsolation: false` must be specified in WebPreferences.
```

baltpeter:
In #15045, a deprecation warning for a default change from `false` to `true` for `contextIsolation` was already (mentioning Electron 5.0). Seems like this was never actually implemented and will only be changed in Electron 12.0?

Should the note for Electron 5.0 be removed?

PR #23508, opened 2020-05-11
docs: add documentation on how to use the built-in spellchecker

MarshallOfSound:
This adds a top-level document for understanding how to use the new spellchecker and the new APIs for it.

Notes: no-notes

### DIFF
--- a/docs/tutorial/spellchecker.md
+++ b/docs/tutorial/spellchecker.md
@@ -1,0 +1,74 @@
+# SpellChecker
+
+Electron has built-in support for Chromium's spellchecker since Electron 8.  On Windows and Linux this is powered by Hunspell dictionaries, and on macOS it makes use of the native spellchecker APIs.
+
+## How to enable the spellchecker?
+
+For Electron 9 and higher the spellchecker is enabled by default.  For Electron 8 you need to enable it in `webPreferences`.
+
+```js
+const myWindow = new BrowserWindow({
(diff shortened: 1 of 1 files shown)


docs/tutorial/spellchecker.md
```diff
@@ -0,0 +1,74 @@
+# SpellChecker
+
+Electron has built-in support for Chromium's spellchecker since Electron 8.  On Windows and Linux this is powered by Hunspell dictionaries, on macOS it makes use of the native spellchecker APIs.
```
codebytere, line 3: nit(comma splice): On Windows and Linux this is powered by Hunspell dictionaries, **and** on macOS it makes use of the native spellchecker APIs.

docs/tutorial/spellchecker.md
```diff
@@ -0,0 +1,74 @@
...
+const myWindow = new BrowserWindow({
+  webPreferences: {
+    spellcheck: true
+  }
+})
+```
+
+## How to set the languages the spellchecker uses?
+
+On macOS as we use the native APIs there is no way to set the language that the spellchecker uses, by default on macOS the native spellchecker will automatically detect the language being used for you.
```
codebytere, line 19: nit(comma splice): On macOS as we use the native APIs there is no way to set the language that the spellchecker uses. By default, on macOS, the native spellchecker will automatically detect the language being used for you.

PR #23512, opened 2020-05-11
fix: leave behind the unmodified XDG_CURRENT_DESKTOP variable

MarshallOfSound:
By mutating this variable apps can't actually check the original value for things like platform/OS/UA detection.

Notes: no-notes

### DIFF
--- a/lib/browser/init.ts
+++ b/lib/browser/init.ts
@@ -186,6 +186,7 @@ function currentPlatformSupportsAppIndicator () {
 }
 
 // Workaround for electron/electron#5050 and electron/electron#9046
+process.env.ORIGINAL_XDG_CURRENT_DESKTOP = process.env.XDG_CURRENT_DESKTOP;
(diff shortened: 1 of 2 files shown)


lib/browser/init.ts
```diff
@@ -186,6 +186,7 @@ function currentPlatformSupportsAppIndicator () {
 }
 
 // Workaround for electron/electron#5050 and electron/electron#9046
+process.env.ORIGINAL_XDG_CURRENT_DESKTOP = process.env.XDG_CURRENT_DESKTOP;
```
deepak1556, line 189: Should this be documented ?

MarshallOfSound: Probably, but not sure where 🤔

deepak1556: https://github.com/electron/electron/blob/master/docs/api/environment-variables.md would be good spot.

MarshallOfSound: Good call, done 👍

PR #23538, opened 2020-05-12
fix: exclude crashpad_handler binary on linux

nornagon:
#### Description of Change
This binary was being built & included on Linux despite the fact that we do not use crashpad on Linux.

#### Checklist

#### Release Notes

Notes: Removed extraneous crashpad_handler binary from the Linux distribution files.

### DIFF
--- a/build/zip.py
+++ b/build/zip.py
@@ -31,7 +31,7 @@ PATHS_TO_SKIP = [
   # app bundle.
   # On Linux, we don't use crashpad, but this binary is still built for some
   # reason. Exclude it from the zip.
-  'crashpad_handler',
+  './crashpad_handler',
 ]
 
 def skip_path(dep, dist_zip, target_cpu):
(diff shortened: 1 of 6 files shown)


build/zip.py
```diff
@@ -25,6 +25,13 @@
   # //chrome/browser/resources/ssl/ssl_error_assistant, but we don't need to
   # ship it.
   'pyproto',
+
+  # On Windows, this binary doesn't exist (the crashpad handler is built-in).
+  # On MacOS, the binary is called 'chrome_crashpad_handler' and is inside the
+  # app bundle.
+  # On Linux, we don't use crashpad, but this binary is still built for some
+  # reason. Exclude it from the zip.
+  'crashpad_handler',
```
deepak1556, line 34: `./crashpad_handler`

> This binary was being built & included on Linux despite the fact that we do not use crashpad on Linux.

`gn refs out/Testing //third_party/crashpad/crashpad/handler:crashpad_handler` shows
https://source.chromium.org/chromium/chromium/src/+/master:components/crash/core/app/BUILD.gn;l=131

nornagon: Yep, upstream is experimenting with crashpad on linux (they have it enabled on Android). We aren't though.

nornagon, line 34:
```suggestion
  './crashpad_handler',
```

Issue #23444, opened 2020-05-07
Cannot use Icon in MenuItem: "Could not call remote method 'buildFromTemplate'" error


dropsonic:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 8.0.2 (also checked on 8.2.5.)
* **Operating System:**
  * Windows 10 Pro (1903)
* **Last Known Working Electron version:**
  * 

### Expected Behavior
The menu is created without any errors

### Actual Behavior
Error:

```
render.js:48 Uncaught (in promise) Error: Could not call remote method 'buildFromTemplate'. Check that the method signature is correct. Underlying error: Error processing argument at index 1, conversion failure from 
Underlying stack: TypeError: Error processing argument at index 1, conversion failure from 
    at Menu../lib/browser/api/menu.js.Menu.insert (electron/js2c/browser_init.js:1855:23)
    at Menu../lib/browser/api/menu.js.Menu.append (electron/js2c/browser_init.js:1835:15)
    at electron/js2c/browser_init.js:1913:12
    at Array.forEach (<anonymous>)
    at Function../lib/browser/api/menu.js.Menu.buildFromTemplate (electron/js2c/browser_init.js:1909:10)
    at electron/js2c/browser_init.js:6745:67
    at IpcMainImpl.<anonymous> (electron/js2c/browser_init.js:6610:27)
    at IpcMainImpl.emit (events.js:210:5)
    at WebContents.<anonymous> (electron/js2c/browser_init.js:3873:23)
    at WebContents.emit (events.js:210:5)

    at electron/js2c/browser_init.js:6748:21
    at IpcMainImpl.<anonymous> (electron/js2c/browser_init.js:6610:27)
    at IpcMainImpl.emit (events.js:210:5)
    at WebContents.<anonymous> (electron/js2c/browser_init.js:3873:23)
    at WebContents.emit (events.js:210:5)
getVideoSources @ render.js:48
async function (async)
getVideoSources @ render.js:32
```

### To Reproduce

```sh
$ git clone https://github.com/dropsonic/223-electron-screen-recorder.git -b master
$ npm install
[...]

PR #23543, opened 2020-05-12
fix: nativeImage remote serialization

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/23444.

We weren't serializing `nativeImages` properly in the remote module, leading to gin conversion errors  when trying to, for example, create a new context menu in the renderer with icons using `nativeImage`. This fixes that by adding a new special case to handle them.

Tested with https://gist.github.com/d8f1bcee2aa65c85b9cf3c8e4d34ad82.

cc @nornagon @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where `nativeImage`s might throw conversion errors in the renderer process.


spec-main/fixtures/api/nativeimage-serialization.html
```diff
@@ -0,0 +1,24 @@
...
+      const { ipcRenderer, remote, nativeImage } = require('electron')
+      const template = [
+        {
+          label: 'hello',
+          icon: nativeImage.createEmpty(),
+        }
+      ]
+
+      try {
+        const menu = remote.Menu.buildFromTemplate(template)
```
nornagon, line 17: This is a bit of a roundabout way to check the serialization. Perhaps just `remote.getGlobal('someFunction')(nativeImage.createEmpty())`?

nornagon: Also, to avoid having to make an actual global, you can intercept the global filtering stuff to provide a test-specific response. see the other tests in `api-remote-spec` for examples.

lib/browser/remote/server.ts
```diff
@@ -259,6 +262,8 @@ const fakeConstructor = (constructor: Function, name: string) =>
 const unwrapArgs = function (sender: electron.WebContents, frameId: number, contextId: string, args: any[]) {
   const metaToValue = function (meta: MetaTypeFromRenderer): any {
     switch (meta.type) {
+      case 'nativeimage':
+        return electron.nativeImage.createFromBitmap(meta.value.buffer, meta.value.size as any);
```
zcbenz, line 266: The `scaleFactor` option should also be passed, and we should add a way to pass the `isTemplateImage` information.

spec-main/fixtures/api/nativeimage-serialization.html
codebytere: @nornagon i'm not sure i can use that approach - i believe we specifically only handle simpler objects over [`valueToMeta`](https://github.com/electron/electron/blob/f0737174efe3664f7e6e830b339517680e352ba5/lib/browser/remote/server.ts#L120)? If i set a native image on the return value for that kind of filtering it'll just show up as empty object 🤔

lib/browser/remote/server.ts
nornagon: If we're going to do a different serialization of native images here we should also update `type-utils`.

codebytere: @nornagon i was planning to handle that in a subsequent pr 🤔 i can do it here if preferable though.

shell/common/api/electron_api_native_image.cc
```diff
@@ -272,11 +281,11 @@ gin::Handle<NativeImage> NativeImage::Resize(v8::Isolate* isolate,
   if (width_set && !height_set) {
     // Scale height to preserve original aspect ratio
     size.set_height(width);
-    size = gfx::ScaleToRoundedSize(size, 1.f, 1.f / GetAspectRatio());
+    size = gfx::ScaleToRoundedSize(size, 1.f, 1.f / GetAspectRatio(args));
```
nornagon, line 284: I don't think this technique of passing `args` through works, because `GetScaleFactorFromOptions` calls `args->GetNext()`, so it'll be iterating through the arguments. This function will end up ultimately calling `args->GetNext()` twice.

Better to pull the scale factor out in each function which takes them, and then pass the scale factor along where necessary.

spec-main/api-remote-spec.ts
```diff
@@ -222,6 +223,74 @@ ifdescribe(features.isRemoteModuleEnabled())('remote module', () => {
     });
   });
 
+  describe('nativeImage serialization', () => {
+    const w = makeWindow();
+    const remotely = makeRemotely(w);
+
+    it('can serialize an empty nativeImage', async () => {
+      const getEmptyImage = (img: nativeImage) => {
```
nornagon, line 231:
```suggestion
      const getEmptyImage = (img: NativeImage) => {
```

spec-main/api-remote-spec.ts
```diff
@@ -222,6 +223,74 @@ ifdescribe(features.isRemoteModuleEnabled())('remote module', () => {
...
+        return new Promise((resolve, reject) => {
+          if (img.isEmpty()) {
+            resolve('empty image');
+          } else {
+            reject(new Error('incorrectly serialized image'));
+          }
+        });
```
nornagon, line 238: It seems weird that this is a promise because it doesn't do any asynchronous things—can it be refactored to not use a promise? e.g.
```suggestion
        return img.isEmpty()
```

spec-main/api-remote-spec.ts
```diff
@@ -222,6 +223,74 @@ ifdescribe(features.isRemoteModuleEnabled())('remote module', () => {
...
+        return new Promise((resolve, reject) => {
+          if (img.isEmpty()) {
+            reject(new Error('incorrectly serialized image'));
+          } else {
+            const size = img.getSize();
+            if (size.height !== 0 && size.width !== 0) {
+              resolve('non-empty image');
+            }
+          }
+        });
```
nornagon, line 262: same here—
```suggestion
        return img.getSize();
```

spec-main/api-remote-spec.ts
```diff
@@ -222,6 +223,74 @@ ifdescribe(features.isRemoteModuleEnabled())('remote module', () => {
...
+
+      w().webContents.once('remote-get-global', (event) => {
+        event.returnValue = getNonEmptyImage;
+      });
+
+      await expect(remotely(() => {
+        const { nativeImage } = require('electron');
+        const nonEmptyImage = nativeImage.createFromDataURL('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAIAAAACCAIAAAD91JpzAAAAFklEQVQYlWP8//8/AwMDEwMDAwMDAwAkBgMBBMzldwAAAABJRU5ErkJggg==');
+        return require('electron').remote.getGlobal('someFunction')(nonEmptyImage);
+      })).to.eventually.equal('non-empty image');
```
nornagon, line 273:
```suggestion
      })).to.eventually.deep.equal({width: 123, height: 123});
```

spec-main/api-remote-spec.ts
```diff
@@ -222,6 +223,55 @@ ifdescribe(features.isRemoteModuleEnabled())('remote module', () => {
...
+          enableRemoteModule: true
+        }
+      });
+
+      ipcMain.once('make-menu', (event, message) => {
+        expect(message).to.match(/^No error thrown/);
+        done();
+      });
+
+      win.loadFile(path.join(fixtures, 'api', 'nativeimage-serialization.html'));
```
nornagon, line 271: nit: i think this would be better written inline using `remotely`. e.g.
```suggestion
      await expect(remotely(() => {
        const { remote, nativeImage } = require('electron');
        remote.Menu.buildFromTemplate([
          {
            label: 'hello',
            icon: nativeImage.createEmpty(),
          }
        ]);
      })).to.be.fulfilled();
```

miniak:
@codebytere The `getScaleFactors()` change is a good one and should also be used in https://github.com/electron/electron/blob/master/lib/common/type-utils.ts

shell/common/api/electron_api_native_image.h
```diff
@@ -84,15 +85,16 @@ class NativeImage : public gin_helper::Wrappable<NativeImage> {
...
-  gin::Handle<NativeImage> Resize(v8::Isolate* isolate,
+  gin::Handle<NativeImage> Resize(gin::Arguments* args,
                                   base::DictionaryValue options);
   gin::Handle<NativeImage> Crop(v8::Isolate* isolate, const gfx::Rect& rect);
   std::string ToDataURL(gin::Arguments* args);
   bool IsEmpty();
-  gfx::Size GetSize();
-  float GetAspectRatio();
+  gfx::Size GetSize(const base::Optional<float> scale_factor);
+  float GetAspectRatio(float scale_factor);
```
nornagon, line 97: These functions are exposed to JS, is it your intention to add extra parameters without documenting it?

In particular `getAspectRatio` is changed in a breaking way here.

I think it might be better to move the bodies of both of these functions out to helper methods are called both from the JS-exposed functions and from other methods that need to pass in the scale factor. i.e. the `GetSize` and `GetAspectRatio` functions that are exposed to JS should continue to take 0 parameters, but internally they call `GetSizeImpl(1.0)` or something.

miniak:
@codebytere but I agree that it's kind of confusing that we have nativeImage available both in renderer and main, and you can have a proxy nativeImage from main in the renderer vs actual nativeImage.

miniak:
@codebytere with `desktopCapturer` accessed via `remote`, the actual data will stay in the main process and only the data necessary for creating the proxy objects would be transferred via IPC. With the approach this PR is using, you would transfer the image main -> renderer and then back renderer -> main when creating the menu.

miniak:
@codebytere also even better would be to recommend against using the remote module, in which case it makes sense to be able to use `desktopCapturer` in the main process, which my PR is addressing.

nornagon:
@miniak native images are used in a few other places than `desktopCapturer`, so this wouldn't cover those use cases.

agreed that the best solution for everyone involved is to not use `remote` though 😁

miniak:
@nornagon ok, but in that case we should be able to transfer nativeImage both ways, ideally extend the code in type-utils with the scale factors and re-use in the remote module?

shell/common/api/electron_api_native_image.h
codebytere: i did intend to do that, but i'm happy to do something else

codebytere:
@miniak see comment above; i plan to follow this PR with one for typeUtils.

PR #23561, opened 2020-05-13
chore: cherry-pick 1288aa12369e from angle

ppontes:
Update the active texture cache before changing the texture binding.

When a new texture is bound, the texture binding state is updated before
updating the active texture cache. With this ordering, it is possible to delete
the currently bound texture when the binding changes and then use-after-free it
when updating the active texture cache.

BUG=angleproject:1065186

Change-Id: Id6d56b6c6db423755b195cda1e5cf1bcb1ee7aee
Reviewed-on: https://chromium-review.googlesource.com/c/angle/angle/+/2124588
Commit-Queue: Geoff Lang <geofflang@chromium.org>
Reviewed-by: Jamie Madill <jmadill@chromium.org>

Notes: Security: backported fix for CVE-2020-6463: use-after-free in Angle.

### DIFF
--- a/patches/angle/.patches
+++ b/patches/angle/.patches
@@ -1,0 +1,1 @@
+update_the_active_texture_cache_before_changing_the_texture_binding.patch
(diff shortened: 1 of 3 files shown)


jkleinsc:

PR #23564, opened 2020-05-13
fix: run Node.js at-exit callbacks in renderer proc

codebytere:
Backport of #23419

See that PR for details.

Notes: Ensured that exit callbacks are run for Node.js in the renderer process.


jkleinsc:

PR #23569, opened 2020-05-13
refactor: ginify Session

nornagon:
#### Description of Change
This refactors `Session` to use `gin::Wrappable` instead of `gin_helper::TrackableObject`.

Additionally, it fixes a segfault that could sometimes happen on shutdown in `KeyWeakMap` (related to `DataPipeHolder`).

It also fixes `constructor.name` not being set on `gin_helper::EventEmitterMixin` objects, and adds a test for the previously untested `session-created` event.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


lib/browser/api/net.js
```diff
@@ -244,7 +243,8 @@ function parseOptions (options) {
     }
   }
   if (options.session) {
-    if (options.session instanceof Session) {
+    // Weak check, but it should be enough to catch 99% of accidental misuses.
+    if (options.session.constructor.name === 'Session') {
```
MarshallOfSound, line 247: This check needs to be way safer.  `options.session.constructor` could be falsey / not an object

nornagon, line 247:
```suggestion
    if (options.session?.constructor?.name === 'Session') {
```

nornagon:
> Are their any actions that were taken with `sess.destroy()` that aren't possible through alternative methods? i.e. Did destroying the session synchronously remove any files from disk

not to my knowledge

> have any effect on existing webContents using that session, etc.?

I believe doing so would cause crashes. The `BrowserContext` in which a `WebContents` resides is not expected to be deleted.

PR #23570, opened 2020-05-13
test: run tests under asan on linux-x64

nornagon:
#### Description of Change
This adds a step to build + run tests under asan on one platform (linux-x64).

I believe with Goma this is possibly reasonable to run on every PR.

#### Checklist

#### Release Notes

Notes: none


jkleinsc:
Nice!  I see the asan x64-linux build took 15 minutes which isn't bad for an asan build!

nornagon:
@jkleinsc i'm disabling ffmpeg builds under asan explicitly in this PR because they are built with `is_official_build = true`, which is incompatible with `is_asan = true`, by `assert` in the build files.

nornagon:
https://github.com/electron/electron/pull/23570/files#diff-1d37e48f9ceff6d8030570cd36286a61R1680

nornagon:
er.. hm, it should still be building it for ARM though.

nornagon:
Two bugs discovered so far: https://github.com/electron/electron/issues/23640 https://github.com/electron/electron/issues/23642

nornagon:
There are a bunch of tests that seem to be failing consistently when run in CI, but I can't reproduce those failures either locally or when SSHed into the CircleCI machine. Any ideas?

jkleinsc:
@nornagon when you tested via ssh do you run the whole test suite?  The only other thing I can think of is that on CircleCI we split the tests into parallel testing runs, so only 1/3rd of the tests run at once.

nornagon:
@jkleinsc yeah, I copy/pasted the command that CI ran directly, including the list of filenames that was split out.

.circleci/config.yml
```diff
@@ -1266,18 +1276,6 @@ steps-verify-ffmpeg: &steps-verify-ffmpeg
...
-    - attach_workspace:
-        at: .
-    - *step-depot-tools-add-to-path
-    - *step-electron-dist-unzip
-    - *step-chromedriver-unzip
-    - *step-setup-linux-for-headless-testing
-
-    - *step-verify-chromedriver
-    - *step-maybe-notify-slack-failure
-
```
nornagon, line 1280: These are apparently unused

nornagon:
I believe this is now ready for consideration for merging. This includes #27827 and #27774, which fix failures that ASan detects, so it should be merged only after those two PRs are merged.

ckerr (approved):
This is fantastic

WALLOUD:
> #### Description of Change
> This adds a step to build + run tests under asan on one platform (linux-x64).
> 
> I believe with Goma this is possibly reasonable to run on every PR.
> 
> #### Checklist
> * [x]  PR description included and stakeholders cc'd
> * [x]  PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
> 
> #### Release Notes
> Notes: none

PR #23572, opened 2020-05-13
fix: use a weakptr for the context bridge function store

MarshallOfSound:
Love a good WeakPtr...

Crash flow is quite complicated but traced to:

* Expose function via ctxBridge from context A to context B
* Somehow pass the function from context B to context C (through a child window for instance via window.opener)
* Context C stores a reference to the function
* Release context A and B (by reloading the window)
* Call function in context C

**Before:** Crash
**After:** Error thrown about calling a function from a released context

Notes: Fixed crash when calling functions passed over the contextBridge into a third context after the primary context has been released.


MarshallOfSound:
Superseded by #23592

PR #23578, opened 2020-05-13
refactor: ginify BrowserView

nornagon:
#### Description of Change
This refactors `BrowserView` to use `gin::Wrappable` instead of `gin_helper::TrackableObject`. It also makes some changes to the API, removing the `id` and related methods, and making the object self-owned instead of forcing manual memory management on the user.

I'm not yet convinced that the new code doesn't leak `BrowserView`s forever in certain (or all) situations, but putting it up for review because at least the tests pass!

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Removed experimental APIs: `BrowserView.{fromId, fromWebContents, getAllViews}` and the `id` property of `BrowserView`.


nornagon:
@zcbenz removed APIs are marked experimental (mostly, though I think the one omission is a mistake and should have been marked experimental). Do we still list breaking changes when APIs are experimental?

zcbenz:
I could not find any documents on what to do with removing experimental APIs, I'm good without mentioning it then.

pushkin-:
@nornagon I am currently relying on `BrowserView.fromWebContents`. What is the expected transition path forward? Am I expected to create and maintain my own object mapping webcontents to browserview?

nornagon:
@pushkin- yes, I'd advise you to implement your own mapping.

The previous implementation simply looped through all the `BrowserView`s in existence until it found one with the right WebContents. You can likely make a more efficient mapping system. e.g.

```js
const webContentsToBrowserView = new WeakMap

// ... when creating a BrowserView
webContentsToBrowserView.set(browserView.webContents, browserView)

// to replace BrowserView.fromWebContents
function browserViewFromWebContents(browserView) {
  return webContentsToBrowserView.get(webContents)
}
```

PalmerAL:
@nornagon Just to make sure I understand the new behavior here: If I was previously calling `view.destroy()`, I should just stop doing that? And the view will be kept around as long as I keep a reference to it, and once I delete that reference, the view will be destroyed automatically?

nornagon:
@PalmerAL yep, just like with a regular JS object, you should be able to trust it to maintain a lifetime that makes sense.

- if it's attached to a non-closed BrowserWindow, it won't be destroyed
- if you have a reference to it, it won't be destroyed

if it's not attached to any window, and you have no reference to it, then eventually V8's GC will come along and delete it.

PalmerAL:
@nornagon Is there any way to force the view to stop using resources immediately if I want it to? The usecase I'm testing is a page that is playing audio; previously I could call `destroy()` when the view was removed from the window, and the audio would stop playing immediately. In Electron 11, the audio continues playing for a while before finally stopping (I'm assuming when GC occurs).

nornagon:
You should be able to destroy the WebContents inside: `browserView.webContents.destroy()` (this is a private API, not documented). Alternatively, you can cause the webContents to clear itself by navigating it, e.g. `browserView.webContents.loadURL('about:blank')`.

I'd eventually like to move away from `destroy()` calls altogether, as having the underlying C++ object disappear is problematic for the V8/C++ integration, but there should always be a way to manually end a renderer!

nornagon:
You should be able to destroy the WebContents inside: `browserView.webContents.destroy()` (this is a private API, not documented). Alternatively, you can cause the webContents to clear itself by navigating it, e.g. `browserView.webContents.loadURL('about:blank')`.

I'd eventually like to move away from `destroy()` calls altogether, as having the underlying C++ object disappear is problematic for the V8/C++ integration, but there should always be a way to manually end a renderer!

pushkin-:
@nornagon I may be extrapolating, or misinterpreting, but regarding your comment about clearing the webcontents by navigating to "about:blank", are you saying that that's expected to kill the renderer process, or no? posted [question here](https://stackoverflow.com/questions/60661917/does-electron-have-a-standard-way-of-killing-a-useless-renderer-process) a while ago

nornagon:
@pushkin- not necessarily, but it should end the page itself in most circumstances.

nornagon:
If you want a reliable way to force-exit a renderer process, one was [recently added](https://github.com/electron/electron/pull/25580).

pushkin-:
@nornagon Another clarifying question. In your example, the browser view is a strong reference, but its webcontents is a weak reference. So the BrowserView itself won't be GC'd if it's not attached to a window, but its webcontents could be if there isn't another reference to it?

In my case, I don't want to GC anything until window is closed entirely, so maybe I should just use a regular map that I clean up when the window closes.

nornagon:
BrowserViews are kept alive while the underlying WebContents is alive. If you create a BrowserView and don't load anything in its WebContents and don't attach it to a window, it will be GC'd as normal. If it's attached to a window then it will be kept alive at least as long as the window is alive (see [here](https://github.com/electron/electron/blob/b4196ca4862327f7012fd6fdc67bc2a2957a89ae/shell/browser/api/electron_api_base_window.h#L262)).

HTH.

pushkin-:
@nornagon Ok and if it's navigated to a page, but *isn't* attached to a window, then what happens? I guess it will stay alive if I'm interpreting "kept alive while the underlying WebContents is alive" correctly.

nornagon:
@pushkin- yep, i believe it will stay alive in that case.

yohan1234:
@pushkin- @nornagon 

It doesn't matter if it's been bound to a BrowserWindow.

BrowserView always leaks if you don't run the internal destroy function on webContents:
`browserView.webContents.destroy();`

This creates 40 zombies:
```
for (let i = 0; i < 40; i++) {
  let browserView: any = new BrowserView({
    webPreferences: {},
  });
  browserView.webContents.loadURL("http://example.com");
}
```
This nukes those 40 zombies by calling destroy:
```
for (let i = 0; i < 40; i++) {
  let browserView: any = new BrowserView({
    webPreferences: {},
  });
  browserView.webContents.loadURL("http://example.com");
  browserView.webContents.destroy();
}
```

Tested on electron 11.0.4. Would have been great to have tested this before the merge.

kevinlinv:
@yohan1234 I have been looking at this too! Thanks for putting the comment up. 

I don't have a solid understanding of how V8's GC work (_e.g. how often is it running?_). 
Thus, I can't really tell whether the leak here:
- is a legit resource leaking, or
- is only temporary, because V8's GC has not yet been scheduled to run

I was thinking about perhaps trying to call GC manually to verify, but I would need to figure out how to do that first 😭 

Aside: calling the undocumented `browserView.webContents.destroy();` API seems to clear it immediately.

PalmerAL:
> I was thinking about perhaps trying to call GC manually to verify, but I would need to figure out how to do that first 😭

If you run Electron with --expose-gc, there’s a gc() function that I think should work.

yohan1234:
@kevinlinv This does not have to do with the garbage collector. Running global.gc() does not influence the behavior. When the webContents runs loadURL(), references get created somewhere internally that are never released until you perform a destroy(). In my example, there are no references reachable from main and the leak persists.

kevinlinv:
@yohan1234 You're right! I tried running GC on my app and the renderer processes did not go away 😞

nornagon:
WebContents will stay alive as long as there's a page loaded in it. As you've discovered, the BrowserView can be GC'd while the WebContents stays alive, if there's a page loaded. To make sure the WebContents is closed, you must call `destroy` or otherwise close the WebContents (`window.close()` from the loaded page also works fine).

This is a common pattern in JS, consider e.g. creating sockets in Node.js:

```js
for (let i = 0; i < 40; i++) {
  let socket = net.connect({port: 1234}, (s) => { /* ... */ })
}
```

Those sockets will remain alive unless explicitly closed.

yohan1234:
@nornagon You're right, but the difference is that sockets have a destroy function whereas BrowserView and webContents do not. 

I don't mind explicit resource management, we were destroying BrowserView using BrowserView.destroy before. The problem is:

- BrowserView.destroy was removed
- webContents does not have a public API destroy function

So to remove the process leak, one needs to call an undocumented function which is unknowable by users of Electron and without it, there's no proper lifecycle management. 

This is an issue in our case because we dynamically create BrowserViews and use them as tabs in our application.

nornagon:
@yohan1234 ah, yes, you're right. I had it in my head that `webContents.destroy` was publicly documented.

I agree that this is a missing piece of our API. I've opened a new issue to track it: #26929

yohan1234:
@nornagon Excellent!

pushkin-:
@nornagon Calling `destroy()` and then closing the window crashes Electron for me actually 
[dump.zip](https://github.com/electron/electron/files/5675284/dump.zip)

yohan1234:
@pushkin- Is it bound to a BrowserWindow when you destroy it? to remove it prior to destroy: BrowserWindow.setBrowserView(null)

pushkin-:
@yohan1234  removing it from the window fixed the crash. Thanks!

pushkin-:
So destroying the webcontents fixes the process leak, but the entries are still not removed from my `WeakMap<WebContents, value>` for some reason. Maybe it just takes a really long time? (I waited 15 minutes). Or maybe there's something else going on. I might have to use a regular map and just delete the entries myself.

yohan1234:
@pushkin- Not sure what you mean. All references need to be removed for garbage collection. Do you mean some kind of internal map in BrowserWindow or something in your application?

pushkin-:
@yohan1234 The WeakMap structure that I created from above (what nornagon posted) still has entries for destroyed webcontents:

    webContentsToBrowserView.set(browserView.webContents, browserView)

yohan1234:
@pushkin- Sorry, where is the code?

pushkin-:
@yohan1234  I'm referring to [this](https://github.com/electron/electron/pull/23578#issuecomment-697003802).

My specific code is like:
```
interface IWebContentsInfo {
	browserView: Electron.BrowserView;
	browserWindow: Electron.BrowserWindow;
}
let webContentsInfo = new WeakMap<Electron.WebContents, IWebContentsInfo>();
```

And when I close the window, I do:
```
window.setBrowserView(null);
const webContentsIds = getAllWebContentsIdThatAreAssociatedWithThatWindow(); // each WC is a tab
for each webContents Id:
    webContents.fromId(id).destroy();
```

And the entries are never removed from the WeakMap.

yohan1234:
@pushkin- You need to call .delete on the map to remove the references in that map, .destroy() only destroys the internal V8 stuff in webcontents, it doesn't nuke the chain of things

yohan1234:
@MatthieuLeMee we seem to have the same use case. @nornagon opened up a proposal but I don't know who the maintainer is:
https://github.com/electron/electron/issues/26929

Perhaps the practical approach is to unmerge this PR?

nornagon:
`BrowserView` will be GCable when its WebContents is destroyed and the reference to it is dropped. `bv.webContents.destroy()` should do the trick.

pushkin-:
@nornagon Unless it's a bug in my code, I pasted [an example](https://github.com/electron/electron/pull/23578#issuecomment-743387401) above where I destroyed the webcontents, but the entry is not cleared from the `WeakMap` (I waited 15 minutes for GC; maybe it takes longer?).

I could open a separate issue for this if you want (if we don't already have one)

nornagon:
@pushkin- yeah, let's go ahead and open a new issue, I think. If the BVs aren't being GC'd even after the webContents is destroyed then that'd be a bug. I don't think WeakMap would factor into it.

pushkin-:
@nornagon  I posted an issue [here](https://github.com/electron/electron/issues/27071)

eric-burel:
Hi guys, I am currently upgrading from v7 (long way ahead!), and I don't get how I should replace those methods.

Here's my use case: I want some renderer code to open a dialog. To do so, I've created an IPC handler, and I expose a method like this in preload script: `showOpenDialogSync: (...args) => {     return ipcRenderer.sendSync('show-open-dialog-sync', ..args);}`. 

First, I am not sure if I am allowed to call `dialog` in the preload script, but that's another point.

Then, the handler is like this:
```
const showOpenDialogSyncHandler: IpcMainHandlerDef = [
    showOpenDialogSyncChannel,
    async (evt: any, ...args: any) => {
        const parent = evt.sender
            ? BrowserWindow.fromWebContents(evt.sender)
            : null;
        // @ts-ignore
        evt.returnValue = dialog.showOpenDialogSync(parent, ...args);
        return evt;
    },
];
```

So to get the parent, I need to get the BrowserWindow based on the event sender WebContents. How should I achieve that without those methods?

By the way they are still documented and nothing tells that they are deprecated or removed in current doc: https://www.electronjs.org/docs/latest/api/browser-window#browserwindowfromwebcontentswebcontents

I have some hard time upgrading the code from "remote" to the new APIs

PR #23592, opened 2020-05-14
refactor: remove the RenderFrameFunctionStore and use privates to memory manage

MarshallOfSound:
Replaces #23572

This PR completely removes context bridges custom memory management and let's V8 do the hard work for us.  It appears to increase GC performance with regard to things sent over the bridge and dramatically simplifies the logic.

Notes: Fixed issues where objects sent over the context bridge could be freed prematurely resulting in crashes when accessing them


shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -169,24 +126,27 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
...
-          base::BindRepeating(&ProxyFunctionWrapper, store, func_id,
-                              support_dynamic_properties));
-      FunctionLifeMonitor::BindTo(destination_context->GetIsolate(),
-                                  v8::Local<v8::Object>::Cast(proxy_func),
-                                  store->GetWeakPtr(), func_id);
-      object_cache->CacheProxiedObject(value, proxy_func);
-      return v8::MaybeLocal<v8::Value>(proxy_func);
+      gin_helper::Dictionary state = gin_helper::Dictionary::CreateEmpty(
+          destination_context->GetIsolate());
+      DCHECK(IsTrue(state.GetHandle()->SetPrivate(
```
nornagon, line 134: anything inside `DCHECK(...)` will be compiled out in non-debug builds. This should be `CHECK`.

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -169,24 +126,27 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
...
+      state.Set("support_dynamic_properties", support_dynamic_properties);
+      v8::MaybeLocal<v8::Function> maybe_proxy_func = v8::Function::New(
+          destination_context, ProxyFunctionWrapper, state.GetHandle());
+      v8::Local<v8::Value> proxy_func;
+      if (maybe_proxy_func.ToLocal(&proxy_func)) {
+        object_cache->CacheProxiedObject(value, proxy_func);
+        return v8::MaybeLocal<v8::Value>(proxy_func);
+      }
+
+      return v8::MaybeLocal<v8::Value>();
```
nornagon, line 149:
```suggestion
      v8::Local<v8::Value> proxy_func;
      if (!v8::Function::New(destination_context, ProxyFunctionWrapper,
                             state.GetHandle()).ToLocal(&proxy_func))
        return v8::MaybeLocal<v8::Value>();
      object_cache->CacheProxiedObject(value, proxy_func);
      return v8::MaybeLocal<v8::Value>(proxy_func);
```

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -169,24 +126,27 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
...
-                                  store->GetWeakPtr(), func_id);
-      object_cache->CacheProxiedObject(value, proxy_func);
-      return v8::MaybeLocal<v8::Value>(proxy_func);
+      gin_helper::Dictionary state = gin_helper::Dictionary::CreateEmpty(
+          destination_context->GetIsolate());
+      DCHECK(IsTrue(state.GetHandle()->SetPrivate(
+          destination_context,
+          v8::Private::ForApi(destination_context->GetIsolate(),
+                              gin::StringToV8(destination_context->GetIsolate(),
+                                              "electron_ref")),
```
nornagon, line 138: Does this still need to be set using `v8::Private` now that it's being passed as `data` to the function callback rather than being attached to the function? Could it simply be a key in the dict?

MarshallOfSound: @nornagon I'm not entirely sure whether `data` is completely safe whereas v8 privates have certain guarantees and not too much overhead.

MarshallOfSound: Actually, I remember thinking about this, hard no.  The object prototypes in the target context could be overridden to intercept the cross-context function if we don't use a private.  The private here protects us against context-side prototype pollution

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -109,43 +97,32 @@ bool IsPlainArray(const v8::Local<v8::Value>& arr) {
...
-  void RunDestructor() override {
-    if (!store_)
-      return;
-    store_->functions().erase(func_id_);
-  }
+bool SetPrivate(v8::Local<v8::Context> context,
+                v8::Local<v8::Object> target,
+                const std::string& key,
+                v8::Local<v8::Value> value) {
+  return IsTrue(target->SetPrivate(
```
nornagon, line 104: maybe `.Check()` instead of `IsTrue`

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -169,22 +146,22 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
...
-      v8::Local<v8::Value> proxy_func = gin_helper::CallbackToV8Leaked(
-          destination_context->GetIsolate(),
-          base::BindRepeating(&ProxyFunctionWrapper, store, func_id,
-                              support_dynamic_properties));
-      FunctionLifeMonitor::BindTo(destination_context->GetIsolate(),
-                                  v8::Local<v8::Object>::Cast(proxy_func),
-                                  store->GetWeakPtr(), func_id);
+      gin_helper::Dictionary state = gin_helper::Dictionary::CreateEmpty(
+          destination_context->GetIsolate());
+      CHECK(SetPrivate(destination_context, state.GetHandle(), "bridged_fn",
```
nornagon, line 154: private keys are a global namespace so probably best to put `electron` and `context bridge` in here somewhere.

also, named constants for the strings to catch typos

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -169,22 +146,22 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
...
       v8::Context::Scope destination_scope(destination_context);
-      v8::Local<v8::Value> proxy_func = gin_helper::CallbackToV8Leaked(
-          destination_context->GetIsolate(),
-          base::BindRepeating(&ProxyFunctionWrapper, store, func_id,
-                              support_dynamic_properties));
-      FunctionLifeMonitor::BindTo(destination_context->GetIsolate(),
-                                  v8::Local<v8::Object>::Cast(proxy_func),
-                                  store->GetWeakPtr(), func_id);
+      gin_helper::Dictionary state = gin_helper::Dictionary::CreateEmpty(
+          destination_context->GetIsolate());
```
nornagon, line 153:
```suggestion
      v8::Local<v8::Object> state = v8::Object::New(destination_context->GetIsolate());
```
since we're not using any of `gin_helper::Dictionary`'s features.

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -334,51 +309,60 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
...
+
+  // Pull the original function and its context off of the data private key
+  v8::MaybeLocal<v8::Value> sdp_value =
+      GetPrivate(calling_context, data, "support_dynamic_properties");
+  v8::MaybeLocal<v8::Value> maybe_func =
+      GetPrivate(calling_context, data, "bridged_fn");
+  v8::Local<v8::Value> func_value;
+  if (sdp_value.IsEmpty() || maybe_func.IsEmpty() ||
+      !gin::ConvertFromV8(args.isolate(), maybe_func.ToLocalChecked(),
+                          &support_dynamic_properties) ||
```
nornagon, line 329: am i blind or is this trying to convert `maybe_func` to a bool?

lib/sandboxed_renderer/init.js
```diff
@@ -113,6 +113,12 @@ function preloadRequire (module) {
 // Process command line arguments.
 const { hasSwitch } = process.electronBinding('command_line');
 
+// Similar to nodes --expose-internals flag, this exposes electronBinding so
+// that tests can call it to get access to some test only bindings
+if (hasSwitch('unsafely-expose-electron-internals-for-testing')) {
+  preloadProcess.electronBinding = process.electronBinding;
+}
```
nornagon, line 120: since the test in question relies on `DCHECK_IS_ON` (i.e. won't run on release builds), do we want to make this flag also only active when `DCHECK_IS_ON`?

MarshallOfSound: @nornagon That `DCHECK_IS_ON` flag isn't available in JS land

shell/renderer/api/electron_api_context_bridge.cc
MarshallOfSound: Wat, how did this pass tests lol

lib/sandboxed_renderer/init.js
nornagon: @MarshallOfSound not with that attitude ;)

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -109,43 +97,34 @@ bool IsPlainArray(const v8::Local<v8::Value>& arr) {
...
+                v8::Local<v8::Object> target,
+                const std::string& key,
+                v8::Local<v8::Value> value) {
+  target
+      ->SetPrivate(
+          context,
+          v8::Private::ForApi(context->GetIsolate(),
+                              gin::StringToV8(context->GetIsolate(), key)),
+          value)
+      .Check();
```
nornagon, line 110: nice e.e. cummings poem

shell/renderer/api/electron_api_context_bridge.h
```diff
@@ -17,20 +17,19 @@ namespace electron {
...
-}
 
-v8::Local<v8::Value> ProxyFunctionWrapper(
-    context_bridge::RenderFrameFunctionStore* store,
-    size_t func_id,
-    bool support_dynamic_properties,
-    gin_helper::Arguments* args);
+const char* const kProxyFunctionPrivateKey = "electron_contextBridge_proxy_fn";
+const char* const kSupportsDynamicPropertiesPrivateKey =
+    "electron_contextBridge_supportsDynamicProperties";
```
nornagon, line 23: These are only ever referenced in the .cc file, so keep them private.

shell/renderer/api/electron_api_context_bridge.h
```diff
@@ -16,21 +16,12 @@ namespace electron {
...
-}
-
-v8::Local<v8::Value> ProxyFunctionWrapper(
-    context_bridge::RenderFrameFunctionStore* store,
-    size_t func_id,
-    bool support_dynamic_properties,
-    gin_helper::Arguments* args);
+void ProxyFunctionWrapper(const v8::FunctionCallbackInfo<v8::Value>& info);
 
 v8::MaybeLocal<v8::Object> CreateProxyForAPI(
```
nornagon, line 21: These functions are never referenced outside of the .cc file either. Can we just delete this header?

MarshallOfSound: @nornagon These functions reference each other so need to be "forward declared" in a sense

nornagon: It's totally fine to forward-declare inside the C++ file—the header is only necessary if you're referencing the functions from somewhere else.

PR #23595, opened 2020-05-14
chore: align crash patch with upstream

nornagon:
#### Description of Change
This aligns our crash patch with the version that landed upstream: https://chromium-review.googlesource.com/c/chromium/src/+/2194585

Notes: none


nornagon:
fast-track to get this in before the next roll.

PR #23598, opened 2020-05-15
feat: deprecate uncompressed crash uploads

nornagon:
#### Description of Change
This deprecates setting the `compress` option of `crashReporter.start` to false (which is the default).

The reason for this is twofold:
1. upstream denied our patch to allow configuring compression, so this is on the path to removing that patch, and
2. almost all crash ingestion servers support gzip, so this will save bandwidth for everyone at little cost.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Deprecated uncompressed crash uploads in `crashReporter`.


docs/api/crash-reporter.md
```diff
@@ -59,7 +59,7 @@ The `crashReporter` module has the following methods:
   * `rateLimit` Boolean (optional) _macOS_ _Windows_ - If true, limit the
     number of crashes uploaded to 1/hour. Default is `false`.
   * `compress` Boolean (optional) - If true, crash reports will be compressed
-    and uploaded with `Content-Encoding: gzip`. Default is `false`.
+    and uploaded with `Content-Encoding: gzip`. Default is `true`.
```
MarshallOfSound, line 62: The default here is still `false` right?

nornagon: oops, yep.
```suggestion
    and uploaded with `Content-Encoding: gzip`. Default is `false`.
```

PR #23650, opened 2020-05-18
docs: Update application-architecture.md

AqibMukhtar:
Few lines of documentation are added to show how main and renderer threads are similar to clients and server in web apps. Understanding main and renderer threads sometimes become difficult for beginners. An example from the client-server model can help them in better understanding of the two threads.

#### Checklist

- [ ] `npm test` passes

#### Release Notes
notes: none
Notes:

### DIFF
--- a/docs/tutorial/application-architecture.md
+++ b/docs/tutorial/application-architecture.md
@@ -44,6 +44,15 @@ and renderer processes, such as [`ipcRenderer`](../api/ipc-renderer.md) and
 [remote](../api/remote.md) module for RPC style communication. There is also
 an FAQ entry on [how to share data between web pages][share-data].
 
+## Main-Renderer threads and Client-Server architecture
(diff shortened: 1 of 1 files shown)


AqibMukhtar:
> I don't agree with this assessment especially that the browser/renderer model should be compared to or implemented in a similar way to client-server.
> 
> It can lead people to the incorrect assumption they should actually run a server in the main process, will confuse folks with this comparison. The relationship between browser/renderer process's has fundamentally different security, performance and communication constraints than a traditional client-server model that can not be sufficiently explaining in this comparison

Yeah, that's true. Different differences between the two may exist, but my aim is to point out similarities between client-server architecture and main-renderer thread model so that the developers coming from the web development background can easily grasp the concept of main-renderer thread.

AqibMukhtar:
Okay, I got it. Thanks a lot @MarshallOfSound and @erickzhao for your reviews. Closing this pull request.

Issue #15760, opened 2018-11-19
Enable sandbox by default


nornagon:
In Electron 5, sandbox mode will no longer be experimental, and running with --enable-sandbox or --enable-mixed-sandbox will be the recommended way to use Electron. In Electron 6, --enable-mixed-sandbox will be on by default and the `sandbox` webPreferences option will be true by default.

This is still an early-stage proposal, so definitely keen to hear any feedback or concerns people might have!

paulcbetts:
While I think that this tactic is appropriate for remote content, it pretty thoroughly nukes the entire Electron ecosystem. Developers will be faced with the choice of, "Use any electron- package by disabling security, or enable security and not be able to do anything". Nuking the ecosystem has been a pretty huge disaster for many OSS projects (Python 3, QT, .NET Core), and while security is one of the few Good Reasons to do so, I think that we should think Really Hard on what we can do in order to get a Secure *and* Usable Electron. 

People do _not_ get IPC. Full stop. Understanding IPC requires you to thoroughly understand the multiprocess model, the most confounding part of Electron, and people _really_ don't understand how to use it effectively - even good developers will end up inadvertently making their app into an undebuggable spaghetti maze of message handoffs. Moving to a model where IPC is literally _required_ to do anything will have a huge effect on developer productivity.

At the end of the day, the issues brought up by @bcrypt et al about Electron Security, are all around way that local-only apps can get tricked into running remote content ("Every XSS now becomes an exploit"). To that end, I think the solution here is to limit the executability of content *not* coming from the app bundle (i.e. `file://` but restricted to the app i.e. `app://`). In Local-Only Mode even If you put up an `img` tag of an SVG from an HTTPS domain, we do _not_ execute its attached JS. 

This model might require more deviation from Chrome (and more work!), but I believe we're at a pretty significant juncture here - if we make Electron so difficult to use in the name of Security that only Experts can build something more than "a webpage in a frame", Electron ceases to be an interesting platform.

emmkimme:
Hi,
Regarding IPC, I fully agree, we faced the same confusing situation, 3 IPCs (Electron: ipcRenderer, ipcMain + Node.js: process.send, ..,), inconsistent behaviors and apis, impacted by options like sandbox, nodeIntegration, and OS/s.... 
We built our own Electron IPC: [electron-common-ipc](https://github.com/emmkimme/electron-common-ipc). 
[...]

PR #23651, opened 2020-05-18
docs: remove note about experimental nature of sandbox

nornagon:
#### Description of Change
Sandbox is no longer experimental, don't call it out as such.

cc @pfrazee for the Beaker mention

Ref https://github.com/electron/electron/issues/15760#issuecomment-629732526

#### Release Notes

Notes: none


pfrazee:
Okay well ❤️ I'm honored by the shoutout

Issue #22490, opened 2020-03-02
WebFrame#executeJavascriptInIsolatedWorld is not rejecting on error


bughit:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v9.0.0-beta.5 x86
* **Operating System:**
  * win7 x64 sp1

### Expected Behavior
The returned promise should reject if js eval fails.

### Actual Behavior
The promise resolves with `undefined`.

### To Reproduce
1. run this app
   ```js
   'use strict';
   
   const {app, BrowserWindow} = require('electron');
   
   let win;
   
   app.on('ready', async () => {
   
     win = new BrowserWindow({
       width: 1440,
       height: 900,
       useContentSize: true,
       webPreferences: {
         nodeIntegration: true,
       },
     });
   
     win.webContents.openDevTools({mode: 'right'});
     win.loadURL('about:blank');
   
   });
   ```
2. eval the following in the js console
   ```js
   try {const r = await require('electron').webFrame.executeJavaScriptInIsolatedWorld(999, [{code: 'thisShouldProduceAnError()'}]); console.log(`executeJavaScriptInIsolatedWorld resolved with ${r}, instead of rejecting`);} catch {}
   ```

zcbenz:
This is unfortunately by design in Chromium, see:
https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/frame/pausable_script_executor.cc;l=62-72

```
    // Note: An error event in an isolated world will never be dispatched to
    // a foreign world.
```

Fix this requires patching Chromium, and we need to be very careful to not break other things. So I prefer documenting this behavior instead of fixing it.

bughit:
@zcbenz please take a look at @MarshallOfSound's commit 0fc172fcaf8ad0158d89d5447be39983fd5614b5

ScriptExecutionCallback was enhanced to detect execution failure and reject, not with the original error but nevertheless an error, which is better than a resolve. This code is common to executeJavascript and executeJavascriptInIsolatedWorld. It works for executeJavascript but not for executeJavascriptInIsolatedWorld.

[...]

PR #23653, opened 2020-05-19
docs: errors in isolated world are not dispatched to foreign worlds

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/22490.

Chromium does not dispatch errors of isolated worlds to foreign worlds, so the promise returned by `webFrame.executeJavaScriptInIsolatedWorld` will not reject when execution fails.

https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/frame/pausable_script_executor.cc;l=62-72

```
    // Note: An error event in an isolated world will never be dispatched to
    // a foreign world.
```

It is possible to make the errors pass through worlds by patching Chromium, but it is risky and has few benefits, so I think we should just document the behavior.

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/web-frame.md
+++ b/docs/api/web-frame.md
@@ -162,9 +162,13 @@ this limitation.
   * `error` Error
 
 Returns `Promise<any>` - A promise that resolves with the result of the executed
-code or is rejected if execution throws or results in a rejected promise.
+code or is rejected if execution could not start.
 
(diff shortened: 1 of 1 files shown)


jkleinsc:
Merging as CI failures are unrelated to doc change.

PR #23671, opened 2020-05-19
docs: document optional scaleFactor param to nativeImage fns

codebytere:
#### Description of Change

Refs https://github.com/electron/electron/pull/23543.

Documents new optional `scaleFactor` parameter on `nativeImage.getSize()` and  `nativeImage.getAspectRatio()` as well as new convenience function `nativeImage.getScaleFactors()`. Previously, those functions would return the size and aspect ratio for the default image representation.

cc @nornagon @MarshallOfSound 
 
#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/native-image.md
+++ b/docs/api/native-image.md
@@ -266,9 +266,13 @@ image instead of a copy, so you _must_ ensure that the associated
 
 Returns `Boolean` - Whether the image is empty.
 
-#### `image.getSize()`
+#### `image.getSize([scaleFactor])`
 
-Returns [`Size`](structures/size.md)
+* `scaleFactor` Double (optional) - Defaults to 1.0.
+
+Returns [`Size`](structures/size.md).
+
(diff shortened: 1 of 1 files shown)


docs/api/native-image.md
```diff
@@ -266,9 +266,13 @@ image instead of a copy, so you _must_ ensure that the associated
...
 
-#### `image.getSize()`
+#### `image.getSize([scaleFactor])`
 
-Returns [`Size`](structures/size.md)
+* `scaleFactor` Double (optional) - Defaults to 1.0.
+
+Returns [`Size`](structures/size.md).
+
+If `scaleFactor` is passed, this will return the aspect ratio corresponding to the image representation most closely matching the passed value.
```
MarshallOfSound, line 275:
```suggestion
If `scaleFactor` is passed, this will return the size corresponding to the image representation most closely matching the passed value.
```

Issue #23448, opened 2020-05-07
Problem with Tray Icon after sleep mode for Linux


trae-op:
Hi. I have a problem with the Tray icon.

* **Electron Version:**
  * 8.2.5
* **Operating System:**
  * Ubuntu 18.04
* **Last Known Working Electron version:**
  * 7.2.4

### Expected Behavior
The Tray Icon always display after go out from sleep mode

### Actual Behavior
After go out from sleep mode the tray icon is hidden.

### Screenshots
If just start app after `npm start` see screen https://prnt.sc/scnq37 then go to sleep mode.
After go out from sleep mode the tray icon is hidden see screen https://prnt.sc/scnu75

### Additional Information
The problem in versions 8.0.0-8.2.5.
I installed a less v7.2.4 and it works. After sleep mode icon is display
My example https://github.com/trae-op/test-electron-app

vladimiry:
Looks like it's a duplicate of https://github.com/electron/electron/issues/22443.

trae-op:
> Looks like it's a duplicate of #22443.
I saw the link, but the problem I see is still not resolved

Thatoo:
Same for Riot : https://github.com/vector-im/riot-web/issues/13692

codebytere:
Closed by https://github.com/electron/electron/pull/23674.

Issue #23287, opened 2020-04-26
Tray icon size broken on Linux, locked to 22px.


exander77:
Broken by this pull request (details there): 
https://github.com/electron/electron/pull/21904

![Screenshot_2020-04-27_00-14-48](https://user-images.githubusercontent.com/2256039/80321566-49656680-881e-11ea-9b11-7d7853d546e0.png)

The icon should have a proper size.

I am using Xfce, but it is probably broken for anybody using larger than 22px icons (which I would say is pretty common with current display resolution).

exander77:
I have made an instruction-level patch, to jump over the code as a workaround:

```
#!/usr/bin/env python3
import sys
import re

# 17b4693:      85 c9                   test   %ecx,%ecx
# 17b4695:      74 6f                   je     17b4706 <_ZN2v88internal17ExternalReference18address_of_jslimitEPNS0_7IsolateE@@Base+0x93d6>
# 17b4697:      83 f9 16                cmp    $0x16,%ecx
# 17b469a:      74 6a                   je     17b4706 <_ZN2v88internal17ExternalReference18address_of_jslimitEPNS0_7IsolateE@@Base+0x93d6>

f = open(sys.argv[1], 'rb+')
data = f.read()
match = re.search(rb'(\x85\xc9.\x6f\x83\xf9\x16\x74)', data)
if match:
    result = match.group(1)
    print('Found:');
    print(match.span());
    new = list(result)
    print(new);
    new[7] = 0xeb
    print(new);
    result = bytes(new)
    f.seek(match.span()[0])
    f.write(result)
    print('Replaced.')
else:
    print('No match.')
f.close()
print('Done.');
```
It is small and should probably work with newer versions as well. I didn't want to recompile the whole electron to get rid of that stupid 22px.

electron-triage:
The Electron version reported on this issue is no longer supported. See our [supported versions documentation](https://www.electronjs.org/docs/tutorial/support#supported-versions).

**If you're still experiencing this issue on a supported version, please open a new issue with an updated repro - a [Fiddle](https://www.electronjs.org/fiddle) is very appreciated.**

Electron has a large issues backlog. To help our team prioritize, we're closing older issues and asking for new issues with updated repro steps if it affects a supported version. This helps sort what issues are still relevant and helps us fix them more quickly.

Thanks for your patience and understanding!

Issue #22443, opened 2020-02-28
Tray icon disappear from Topbar when system goes in stand by


p-fusi:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
8.0.0
* **Operating System:**
Ubuntu 18.04.4 LTS x64
* **Last Known Working Electron version:**
7.1.13

### Expected Behavior
When the system goes in stand by the Tray icon should stay visible in the Topbar

### Actual Behavior
When system goes stand by the Tray icon disappear

### To Reproduce
I set the icon and the Menu in main process:

```
let tray;
const icon = path.join(__dirname, "./icon.png");
const tryMenu = Menu.buildFromTemplate([
  { label: "Reload", role: "reload" },
  { label: "Toggle dev tools", role: "toggleDevTools" }
]);
tray = new Tray(icon);
tray.setContextMenu(tryMenu);
```
When starting electron everything works fine: the icon (and the menu) is visible in the Topbar.  
But if system goes stand by, when I log in again the icon is disappeared.

```sh
$ git clone https://github.com/p-fusi/electron-quick-start -b master
$ npm install
$ npm start || electron .
```

beltrame:
This happens with all electron applications, I can confirm with Ubuntu 19.10

jhasse:
This is a Chromium bug, see https://bugs.chromium.org/p/chromium/issues/detail?id=1042098

felixrieseberg:
Interesting tidbit from @sbuschjaeger: 

> By random chance, I noticed a [similar](https://github.com/jiahaog/nativefier/issues/945) bugreport which mentions a [fix](https://github.com/telegramdesktop/tdesktop/issues/7541) for the telegram desktop app. There seems to be a problem with the icon counter, but the telegram app does not use electron. However, maybe it helps.

vladimiry:
The above-referenced issue says it's Gnome related but I face the issue of getting disappeared icon after the suspend mode using Xfce Desktop Environment, happens quite rarely though.

MatMaul:
After reading various bugs, here is my understanding of the problem of the disappearing icon on lock/unlock.

[...]

PR #23674, opened 2020-05-20
fix: restore original GTK/appindicator implementation of tray icons

MarshallOfSound:
#### Description of Change
In https://chromium-review.googlesource.com/c/chromium/src/+/1820024 Chromium removed the implementation of `Tray` icons we were using on Linux (GTK + Appindicator backed) in favour of their new DBUS backed implementation.

However the new DBUS implementation is not production ready and has a decent number of bugs that would prevent folks shipping to prod.

* #23448
* #22443
* #23287

This PR carefully restores the original code that Chromium deleted.  Tested so far on Ubuntu and will be testing on other distros shortly 😄 

**Note:** I know this looks like a lot of code, but 99% of it is literally just copy-pasting code from Chromium that we had already shipped with for years.

The intention is to remove this implementation was the DBUS implementation has achieved production ready-ness which I will be actively monitoring.  In a follow up PR (as this PR is supposed to be backported) I'll add a build flag to toggle between the implementations.

#### Release Notes

Notes: Restored old implementation of Linux Tray icons to fix a collection of issues where the tray icon wouldn't appear, would be the wrong size or would randomly disappear

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -322,6 +322,7 @@ source_set("electron_lib") {
     "shell/common/api:mojo",
     "//base:base_static",
     "//base/allocator:buildflags",
+    "//chrome/app:command_ids",
     "//chrome/app/resources:platform_locale_settings",
     "//chrome/services/printing/public/mojom",
     "//components/certificate_transparency",
@@ -523,6 +524,16 @@ source_set("electron_lib") {
 
     sources += filenames.lib_sources_nss
     sources += [
+      "shell/browser/ui/gtk/app_indicator_icon.cc",
+      "shell/browser/ui/gtk/app_indicator_icon.h",
+      "shell/browser/ui/gtk/app_indicator_icon_menu.cc",
+      "shell/browser/ui/gtk/app_indicator_icon_menu.h",
+      "shell/browser/ui/gtk/gtk_status_icon.cc",
+      "shell/browser/ui/gtk/gtk_status_icon.h",
+      "shell/browser/ui/gtk/menu_util.cc",
+      "shell/browser/ui/gtk/menu_util.h",
+      "shell/browser/ui/gtk/status_icon.cc",
+      "shell/browser/ui/gtk/status_icon.h",
       "shell/browser/ui/gtk_util.cc",
       "shell/browser/ui/gtk_util.h",
     ]
--- a/shell/browser/ui/gtk/app_indicator_icon.cc
+++ b/shell/browser/ui/gtk/app_indicator_icon.cc
@@ -1,0 +1,379 @@
+// Copyright 2013 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "shell/browser/ui/gtk/app_indicator_icon.h"
+
+#include <dlfcn.h>
+#include <gtk/gtk.h>
+
+#include <algorithm>
+#include <string>
+#include <vector>
+
+#include "base/bind.h"
+#include "base/environment.h"
+#include "base/files/file_util.h"
+#include "base/hash/md5.h"
+#include "base/memory/ref_counted_memory.h"
+#include "base/strings/string_number_conversions.h"
+#include "base/strings/stringprintf.h"
+#include "base/strings/utf_string_conversions.h"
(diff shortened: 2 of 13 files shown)


tarruda:
To build on 8-x-y branch, this patch is necessary:

```diff
diff --git a/shell/browser/ui/gtk/menu_util.cc b/shell/browser/ui/gtk/menu_util.cc
index 3549e71f5..c80be3349 100644
--- a/shell/browser/ui/gtk/menu_util.cc
+++ b/shell/browser/ui/gtk/menu_util.cc
@@ -18,7 +18,6 @@
 #include "third_party/skia/include/core/SkUnPreMultiply.h"
 #include "ui/base/accelerators/accelerator.h"
 #include "ui/base/accelerators/menu_label_accelerator_util_linux.h"
-#include "ui/base/models/image_model.h"
 #include "ui/base/models/menu_model.h"
 #include "ui/events/event_constants.h"
 #include "ui/events/keycodes/keyboard_code_conversion_x.h"
@@ -155,6 +154,7 @@ void BuildSubmenuFromModel(ui::MenuModel* model,
   std::map<int, GtkWidget*> radio_groups;
   GtkWidget* menu_item = nullptr;
   for (int i = 0; i < model->GetItemCount(); ++i) {
+    gfx::Image icon;
     std::string label = ui::ConvertAcceleratorsFromWindowsStyle(
         base::UTF16ToUTF8(model->GetLabelAt(i)));
 
@@ -188,9 +188,8 @@ void BuildSubmenuFromModel(ui::MenuModel* model,
       }
       case ui::MenuModel::TYPE_SUBMENU:
       case ui::MenuModel::TYPE_COMMAND: {
-        auto icon_model = model->GetIconAt(i);
-        if (!icon_model.IsEmpty())
-          menu_item = BuildMenuItemWithImage(label, icon_model.GetImage());
+        if (model->GetIconAt(i, &icon))
+          menu_item = BuildMenuItemWithImage(label, icon);
         else
           menu_item = BuildMenuItemWithLabel(label);
 #if !GTK_CHECK_VERSION(3, 90, 0)
@@ -287,10 +286,9 @@ void SetMenuItemInfo(GtkWidget* widget, void* block_activation_ptr) {
 #if !GTK_CHECK_VERSION(3, 90, 0)
         G_GNUC_BEGIN_IGNORE_DEPRECATIONS;
         if (GTK_IS_IMAGE_MENU_ITEM(widget)) {
-          auto icon_model = model->GetIconAt(id);
-          if (!icon_model.IsEmpty()) {
-            GdkPixbuf* pixbuf = gtk_util::GdkPixbufFromSkBitmap(
-                *icon_model.GetImage().ToSkBitmap());
+          gfx::Image icon;
+          if (model->GetIconAt(id, &icon)) {
+            GdkPixbuf* pixbuf = gtk_util::GdkPixbufFromSkBitmap(*icon.ToSkBitmap());
             gtk_image_menu_item_set_image(GTK_IMAGE_MENU_ITEM(widget),
                                           gtk_image_new_from_pixbuf(pixbuf));
             g_object_unref(pixbuf);
```

vladimiry:
I wish this change happened in along with Electron v8 release since it would save tons of developers time and effort in working around the DBus-based tray icons/menus behavior on different systems. But better late than never :+1: @MarshallOfSound, thanks for taking an initiative.

MarshallOfSound:
Tested and validated.  Gonna kick off these merges

hovancik:
I am actually sorry to hear the `StatusIconLinuxDbus` is getting ripped of. For me, it solved many tray related issues and my app was finally working well on Linux ;/

ArisuOngaku:
@hovancik as of what I understand, it should be brought back when it's more widely spread across DEs. Probably years to wait for that though.

Kodehawa:
I know this is an already-merged PR, but why does this use Ubuntu's AppIndicators? They've been dead for years and the active fork is [Ayatana Indicators](https://github.com/AyatanaIndicators/) ([Explanation](https://ayatanaindicators.github.io/))

It should be compatible with the old AppIndicator logic, but I think it only makes sense to use what everyone else (see: LightDM, MATE, etc) uses now, instead of the old Unity AppIndicators which are as dead as everything Unity7 related

Issue #18161, opened 2019-05-04
process.windowsStore returns undefined even in AppX packages in Electron 5.0.x


dangeredwolf:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * Tested on electron 5.0.0, through 5.0.6
* **Operating System:**
  * Windows 10 (Tested on both 1803 and 1809)

### Expected Behavior
`process.windowsStore` should return true if running inside of an appx package, as described in https://electronjs.org/docs/api/process#processwindowsstore

### Actual Behavior
`process.windowsStore` returns undefined, even when running in an appx package

### To Reproduce
I forked electron-quick-start to demonstrate the issue I found.
[https://github.com/dangeredwolf/electron-quick-start](https://github.com/dangeredwolf/electron-quick-start)

You can then use `npm run buildWindows` to create the appx.

If you're unfamiliar with appx deployment, appx files need to be signed to be installed, even in developer mode. [This is basically how to do that (from Microsoft's docs)](https://docs.microsoft.com/en-gb/windows/uwp/packaging/sign-app-package-using-signtool)

`process.windowsStore` is undefined in both the main and renderer processes when installed as an AppX

### Screenshots

I set it up to throw a dialog box of whatever process.windowsStore is

![image](https://user-images.githubusercontent.com/2636265/57185775-bb230280-6e9f-11e9-901b-0bb384de9fe8.png)

It's obvious from electron itself that it's running from C:\Program Files\WindowsApps

![image](https://user-images.githubusercontent.com/2636265/57185783-d726a400-6e9f-11e9-8d05-bfc38a959897.png)

### Additional Information

After further testing, I confirmed that this is an issue introduced in Electron 5.0.0. You can test this by cloning [the electron-quick-start fork with process.windowsStore checks](https://github.com/dangeredwolf/electron-quick-start) and changing `package.json`'s `"devDependencies"`>`"electron"` from `"^5.0.0"` to `"^4.2.0"`

![image](https://user-images.githubusercontent.com/2636265/57185900-e9561180-6ea2-11e9-84b7-75a4e3ef7a48.png)

[...]

PR #23677, opened 2020-05-20
docs: Add warning about process.windowsStore

mifi:
## Description of Change

I think it's good to warn people about using this variable, because it does not work, and now every developer who uses it in their app will get an unpleasant surprise when their app gets complaints from a user in the Windows Store.

See https://github.com/electron/electron/issues/18161

#### Checklist

#### Release Notes

none


codebytere:
Closing as this is being fixed in https://github.com/electron/electron/pull/23785.

Issue #21427, opened 2019-12-07
Always matches `prefers-color-scheme: light` media-query (no dark mode)


konsumer:
### Preflight Checklist

* [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [X] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.1.3 & 8.0.0-beta.4
* **Operating System:**
  * Linux Manjaro 18.1.3

### Expected Behavior
Electron should behave like Chromium with `prefers-color-scheme` media-queries.

### Actual Behavior
Electron always matches `prefers-color-scheme: light` so OS dark-mode isn't selected in media-query.

### To Reproduce

I made an electron-fiddle [here](https://gist.github.com/dc86e8931bef1781db8fbd0a6faf76ee). You can reproduce issue by switching to dark theme in your OS. See `<link>` tags in `<head>`.

### Screenshots
This is a comparison of same web-page in electron 7.1.3 (right) and chromium 78.0.3904.97 (left) on Arch/Manjaro. I am switching between "Adwaita-dark" and "Adwaita" themes on gnome.

![Peek 2019-12-06 20-06](https://user-images.githubusercontent.com/83857/70368820-84adcb80-1864-11ea-876a-21ad7c612c1e.gif)

gombosg:
Yep, same for me.
See also [VSCode](https://github.com/microsoft/vscode/issues/91169), [Joplin](https://discourse.joplinapp.org/t/feature-idea-automatic-dark-light-theme-switching-according-to-os/6010/8).

For me `window.matchMedia((prefers-color-scheme: light)).matches === true` also returns always true in Chromium regardless of theme. Works in FF properly though.

Prinzhorn:
@konsumer for me this is fixed in 9.0.0-beta.13 (see #22900). Can you confirm?

gombosg:
@Prinzhorn 
I tried an Electron fiddle with 9.0.0-beta.13, but still doesn't seem to work in GNOME 3.34. :disappointed: 
[...]

PR #23678, opened 2020-05-20
fix: read GTK dark theme setting on Linux

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/21427.

Chromium does not respect the dark theme settings from GTK, according to [a Chromium issue](https://bugs.chromium.org/p/chromium/issues/detail?id=998903) it is intended and because of Chromium having its own settings of themes (custom theme, WebUI theme, forced dark theme).

This PR makes sure the default dark theme setting of Electron is synced with GTK.

I don't have a test for this though, since it involves desktop environment settings of the host machine. (We don't test dark theme for macOS neither.)

#### Release Notes

Notes: Fix GTK dark theme setting not respected in Electron on Linux.

### DIFF
--- a/shell/browser/electron_browser_main_parts.cc
+++ b/shell/browser/electron_browser_main_parts.cc
@@ -67,6 +67,7 @@
 #include "ui/base/x/x11_util.h"
 #include "ui/base/x/x11_util_internal.h"
 #include "ui/events/devices/x11/touch_factory_x11.h"
+#include "ui/gfx/color_utils.h"
 #include "ui/gfx/x/x11_types.h"
 #include "ui/gtk/gtk_ui.h"
 #include "ui/gtk/gtk_ui_delegate.h"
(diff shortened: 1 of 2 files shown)


shell/browser/electron_browser_main_parts.cc
```diff
@@ -375,10 +376,19 @@ void ElectronBrowserMainParts::ToolkitInitialized() {
...
   views::LinuxUI::SetInstance(BuildGtkUi(ui::GtkUiDelegate::instance()));
-#endif
-
-#if defined(USE_AURA) && defined(USE_X11)
   views::LinuxUI::instance()->Initialize();
+
+  // GTK does not provide a way to check if current theme is dark, so we compare
+  // the text and background luminosity to get a result.
+  // This trick comes from FireFox.
+  float bg = color_utils::GetRelativeLuminance(gtk::GetBgColor("GtkLabel"));
```
MarshallOfSound, line 384: Can we observe changes to the GTK color theme somehow and do this logic on theme change?

zcbenz: I have updated the code to subscribe to GTK theme changes.

codebytere:
@zcbenz can this be backported to 8? It utilizes `gtk_ui_delegate_` which doesn't exist on that branch.

zcbenz:
It should be able to be backported, I'll take care of it.

PR #23684, opened 2020-05-20
fix: add patch to prevent crash during frame swap with ctx isolation enabled

MarshallOfSound:
Electron's Context Isolation implementation has a side-effect of initializing the isolated worlds WindowProxy during the initialization of the main world WindowProxy as a result of creating the isolated world inside the DidCreateScriptContext hook.  This results in an assertion failing in Chromium during a frame swap where it expects to be able to set a new global_proxy in the WindowProxy of the isolated world BEFORE it is initialized.

To meet this assumption this patch splits SetGlobalProxy into two calls, SetGlobalProxyWithoutInitializing and InitializeIfNeeded which has the same resultant effect but means that all of the global_proxy objects are set BEFORE any WindowProxy's are initialized.

This could probably be upstreamed as it doesn't affect the way Chromium works but also it has no benefit for them at this time.

Notes: Fixed crash when navigating between origins in a child window with nativeWindowOpen and contextIsolation enabled


deepak1556:
Is there a minimal repro for this, its hard to understand from the patch what kind of assertion failure is hit and which type of frame swap triggers this.

Curious, mainly because this implementation hasn't changed much in years, so why hasn't ci caught this ? Can we improve the test if so.

MarshallOfSound:
@deepak1556 I'm working on a test for this edge case, automating it out is quite complex but it should be possible.  I've got a fiddle I'll DM you

patches/chromium/fix_swap_global_proxies_before_initializing_the_windows_proxies.patch
```diff
@@ -0,0 +1,79 @@
...
+@@ -55,8 +55,11 @@ void WindowProxyManager::ReleaseGlobalProxies(
+ 
+ void WindowProxyManager::SetGlobalProxies(
+     const GlobalProxyVector& global_proxies) {
++  for (const auto& entry : global_proxies) {
++    WindowProxyMaybeUninitialized(*entry.first)->SetGlobalProxyWithoutInitializing(entry.second);
++  }
+   for (const auto& entry : global_proxies)
+-    WindowProxyMaybeUninitialized(*entry.first)->SetGlobalProxy(entry.second);
++    WindowProxyMaybeUninitialized(*entry.first)->InitializeIfNeeded();
```
deepak1556, line 76: Can you call `Initialize` instead of `InitializeIfNeeded` to maintain what `SetGlobalProxy` intended.

MarshallOfSound: @deepak1556 Unfortunately not, the entire point of this change is that `SetGlobalProxy` may have a side-effect of initializing a different global proxy.  We don't want to end up double-initializing.  + `Initialize` is not a public method

deepak1556: Oh I definitely misunderstood the patch, I didn't realize it was to prevent double-initializing. I thought with the context isolation being initialized in the `DidInstallConditionalFeatures` hook which is something eventually called by the main world `Initialize`, setting the global proxies during frame swap tries to setup stuff on an uninitialized isolate world and fails. And this patch tries to split that up, by first creating global proxies for all the worlds and then running the initialization stuff on them.

Question, why isn't the global proxy of isolated world cleared during frame swap ?

PR #23699, opened 2020-05-21
test: fix flaky beforeunload tests

zcbenz:
#### Description of Change

The `beforeunload` tests are flaky for following reasons:
1. Chromium disallows `beforeunload` handler to block window closing if there is no user interaction. (The purpose is to block ads.)
2. Sending sync message in `beforeunload` handler sometimes do not work.

For 1), in our tests we rely on `BrowserWindow.close` and `WebContents.loadURL` to trigger `beforeunload` event, and they used to imply user interaction. After recent Chromium upgrades, they no longer imply so. The tests are still passing for most times because we are running lots of `executeJavaScript` and other APIs that make Chromium think there is user interaction.

For 2), I'm still very confused of it and I'm not even sure whether it is a correct statement, I only ran into it for a few times when testing and it went away once I tried to debug it.

So in order to fix the `beforeunload` tests, this PR does:

1. Explicitly mark `close` and `loadURL` as user interaction to recover old behaviors.
2. Use the `BeforeUnloadFired` event of `WebContents` instead of IPC messages to detect if `beforeunload` has ran.
3. Due to Chromium not emitting `BeforeUnloadFired` for navigations, some tests are using `console-message` to get signals from `beforeunload` handler, which is still more robust than sending IPC messages directly.
4. Tests are refactored so we don't run `window.close()` directly in the web pages, otherwise Chromium might not allow `beforeunload` to block closing.

#### Release Notes

Notes: none

### DIFF
--- a/spec-main/fixtures/api/beforeunload-false-prevent3.html
+++ b/spec-main/fixtures/api/beforeunload-false-prevent3.html
@@ -1,15 +1,14 @@
 <html>
 <body>
 <script type="text/javascript" charset="utf-8">
-  function preventNextBeforeUnload() {
+  function installBeforeUnload(removeAfterNTimes) {
+    let count = 0
     window.addEventListener('beforeunload', function handler(e) {
-      e.preventDefault();
-      e.returnValue = '';
-      window.removeEventListener('beforeunload', handler)
-      setTimeout(function() {
-        require('electron').ipcRenderer.sendSync('onbeforeunload')
-      }, 0);
-      return false;
+      setTimeout(() => console.log('beforeunload'))
+      if (++count <= removeAfterNTimes) {
+        e.preventDefault();
+        e.returnValue = '';
+      }
     })
   }
 </script>
--- a/shell/browser/api/electron_api_browser_window.cc
+++ b/shell/browser/api/electron_api_browser_window.cc
@@ -266,6 +266,9 @@ void BrowserWindow::OnCloseButtonClicked(bool* prevent_default) {
     // Already closed by renderer
     return;
 
+  // Required to make beforeunload handler work.
+  api_web_contents_->NotifyUserActivation();
+
   if (web_contents()->NeedToFireBeforeUnloadOrUnload())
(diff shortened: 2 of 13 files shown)


zcbenz:
The beforeunload tests only seem to be flaky on master, so I'm not backporting this to old branches.

deepak1556:
> Sending sync message in beforeunload handler sometimes do not work.

If the beforeunload handler returns a [non-null value](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/dom/document.cc;l=4148) then the renderer will put the page into [paused state](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/page/chrome_client.cc;l=149,) to run the javascript dialog, I don't know if this is the scenario you had encountered.

> Due to Chromium not emitting BeforeUnloadFired for navigations, some tests are using console-message to get signals from beforeunload handler, which is still more robust than sending IPC messages directly.

TIL, chromium seems to let the FrameTreeNode navigator handle it https://source.chromium.org/chromium/chromium/src/+/master:content/browser/frame_host/render_frame_host_impl.cc;l=3033 . But since the FrameTreeNode is very much alive, wonder if we can propose a patch in upstream to notify the `WebContentsDelegate::BeforeUnloadFired` for this case as well.

spec-main/fixtures/api/beforeunload-false-prevent3.html
```diff
@@ -1,15 +1,15 @@
...
-      window.removeEventListener('beforeunload', handler)
-      setTimeout(function() {
-        require('electron').ipcRenderer.sendSync('onbeforeunload')
-      }, 0);
-      return false;
+      setTimeout(() => console.log('beforeunload'))
+      if (++count <= removeAfterNTimes) {
+        e.preventDefault();
+        e.returnValue = '';
+        return false;
```
deepak1556, line 11: `return false` is not needed when `e.returnValue` is provided.

PR #23719, opened 2020-05-22
chore: tsify sandboxed init

MarshallOfSound:
Does what it says on the tin, converts the sandboxed renderer init script to TS

Notes: none

### DIFF
--- a/lib/sandboxed_renderer/init.ts
+++ b/lib/sandboxed_renderer/init.ts
@@ -1,11 +1,11 @@
-'use strict';
-
 /* eslint no-eval: "off" */
 /* global binding, Buffer */
-const events = require('events');
+import { electronBindingSetup } from '@electron/internal/common/electron-binding-setup';
+import * as events from 'events';
+
 const { EventEmitter } = events;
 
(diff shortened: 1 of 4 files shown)


lib/sandboxed_renderer/init.ts
```diff
@@ -80,7 +80,7 @@ webFrameInit();
 
 // Pass different process object to the preload script(which should not have
 // access to things like `process.electronBinding`).
-const preloadProcess = new EventEmitter();
+const preloadProcess: NodeJS.Process = new EventEmitter() as any;
```
nornagon, line 83: Does this work...?
```suggestion
const preloadProcess: NodeJS.Process & EventEmitter = new EventEmitter() as any;
```

typings/internal-ambient.d.ts
```diff
@@ -1,4 +1,5 @@
 declare var internalBinding: any;
+declare var binding: { get: any; process: any; createPreloadScript: (src: string) => Function };
```
nornagon, line 2: Can we be a little more specific here?

lib/sandboxed_renderer/init.ts
MarshallOfSound: NodeJS.Process is an event emitter already

PR #23720, opened 2020-05-22
chore: remove IPC hiddens

MarshallOfSound:
We no longer use these (and when we did it was wrong and we did bad things with them 😆 ) so this cleans up the globals so we don't accidentally start using them again

Notes: none

### DIFF
--- a/lib/renderer/ipc-renderer-internal.ts
+++ b/lib/renderer/ipc-renderer-internal.ts
@@ -1,33 +1,32 @@
-const { ipc } = process.electronBinding('ipc');
-const v8Util = process.electronBinding('v8_util');
+import { EventEmitter } from 'events';
 
-// Created by init.js.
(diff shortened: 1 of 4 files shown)


lib/renderer/ipc-renderer-internal.ts
```diff
@@ -1,33 +1,32 @@
...
+
+ipcRendererInternal.invoke = async function<T> (channel: string, ...args: any[]) {
+  const { error, result } = await ipc.invoke<T>(internal, channel, args);
+  if (error) {
+    throw new Error(`Error invoking remote method '${channel}': ${error}`);
+  }
+  return result;
+};
+
+export { ipcRendererInternal };
```
nornagon, line 32: be consistent between ipc-renderer-internal and ipc-renderer w.r.t. `export default`

MarshallOfSound: exporting named variables is IMO better but unfortunately publicly exposed APIs have to be default exports.  Consistency here would result in us using a Worse Thing

PR #23721, opened 2020-05-22
chore: tsify more of lib

MarshallOfSound:
Does what it says on the tin, converts a bunch of the easier / smaller files to TS.  Mostly just file renames and `s/require/import`

Notes: none


lib/browser/api/session.ts
```diff
@@ -0,0 +1,10 @@
+const { fromPartition } = process.electronBinding('session');
+
+const sessionAPI = {
+  fromPartition,
+  get defaultSession () {
+    return fromPartition('');
+  }
+};
+
+export default sessionAPI;
```
nornagon, line 10:
```suggestion
export default {
  fromPartition,
  get defaultSession () {
    return fromPartition('');
  }
};
```

PR #23727, opened 2020-05-22
fix: trigger activate event when app is activated via app switcher

isnotgood:
When application is activated thru macOS app switcher (cmd+tab) the App's activate event is note emitted. [Here](https://gist.github.com/isnotgood/5ad90ba50f45ad26be052de8646fdb1b) is Electron Fiddle with steps to repro (tested on macOS 10.15.4). 

#### Description of Change
The reason is that `applicationShouldHandleReopen:hasVisibleWindows:` is sent only when app
is activated via Dock not the other ways like app switcher. Using `applicationDidBecomeActive:` is handling all the cases properly. Also to maintain backward compatibility I added logic to avoid firing 'activate' event after the application launch.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/mac/electron_application_delegate.mm
+++ b/shell/browser/mac/electron_application_delegate.mm
@@ -41,7 +41,10 @@ static base::mac::ScopedObjCClassSwizzler* g_swizzle_imk_input_session;
 @end
 #endif  // BUILDFLAG(USE_ALLOCATOR_SHIM)
 
-@implementation ElectronApplicationDelegate
+@implementation ElectronApplicationDelegate {
+ @private
+  bool isFirstActivation_;
+}
 
 - (void)setApplicationDockMenu:(electron::ElectronMenuModel*)model {
   menu_controller_.reset([[ElectronMenuController alloc] initWithModel:model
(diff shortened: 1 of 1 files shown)


isnotgood:
Tests are failing but the same one are failing on develop as well. I guess it's being taken care of in https://github.com/electron/electron/pull/23699.

Also I didn't provided any tests for this, but if anyone will give me some hints how to test this functionality I'll be happy to do so 😄

isnotgood:
@zcbenz yeah, of course! Updated the comment in the code. Thanks for taking look at it 🙇

bpasero:
@isnotgood I do not fully understand this change. Does this mean `app.on('activate')` will fire when you switch to the Electron app and before it was not? That sounds like a breaking change to me because e.g. VSCode opens a window when none is opened and I would not expect that to happen unless you click the dock icon.

isnotgood:
@bpasero ouch, unfortunately you understand correctly😢 In that case we need to revert this one back to use `applicationShouldHandleReopen:hasVisibleWindows:` not to break existing apps which use it for reopen handling and add new mac specific event for actual activation of the app `app.on('did-become-active')` as I don't see any other way how to handle that use case. @zcbenz wdyt?

bpasero:
Yes, adding a separate event seems better.

zcbenz:
Yeah this should be reverted and adding a new event looks good to me. I'll revert the PRs, thanks for bringing this up.

PR #23755, opened 2020-05-25
fix: ensure valid `userData` and `userCache` paths

### LINKED ISSUE #23754: `app.getPath('userData')` gives invalid path if Product Name has invalid characters
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 8.2.5
* **Operating System:**
  * Windows 7
  * Unknown

### Expected Behavior

Calling [`app.getPath('userData')`](https://www.electronjs.org/docs/api/app#appgetpathname) should always give a valid path

### Actual Behavior

If my app's Product Name contains characters which are not valid for a folder name, such as `:` in windows, `app.getPath('userData')` will generate an invalid path (since it simply concatenates the Product Name):

```
C:\Users\Pedro\AppData\Roaming\Foo: Bar
```

### To Reproduce

Create any electron app with `productName: "Foo: Bar"` in `package.json` and call `app.getPath('userData')` and see the invalid path be returned.

If you want a step-by-step reproduction guidance:

* Clone `https://github.com/electron/electron-quick-start`
* Add `productName: "Foo: Bar"` to `package.json`
* Add `console.log(app.getPath('userData'))` to `main.js`
* Run and see it log `C:\Users\Pedro\AppData\Roaming\Foo: Bar`

### Additional Information

Suggestion: [filenamify](https://github.com/sindresorhus/filenamify) could be used internally by electron.

papb:
#### Description of Change

Closes #23754

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed `app.getPath('userData')` and `app.getPath('userCache')` returning invalid paths in some cases


papb:
Hello @MarshallOfSound, thank you for the fast response! Indeed I was surprised to observe that electron doesn't have any dependencies (only `devDependencies`). However I didn't find any guidance on this matter in the contribution guidelines...

What is the standard practice when electron needs something that another existing npm package already provides? Do you just repeat the code entirely? :thinking: 

For this specific case, copying might be feasible because the `filenamify` package is quite small, and I guess @sindresorhus wouldn't mind having his code copied to here.

Note: what is a chromium helper? Maybe that's the answer but I don't know what it is :sweat_smile:

MarshallOfSound:
> What is the standard practice when electron needs something that another existing npm package already provides? Do you just repeat the code entirely? 🤔

We've never really run into this case because as mentioned below, Chromium normally has all the helpers / utilities we could dream of.  In most cases we just implement the code ourselves or use Chromium's methods.

> For this specific case, copying might be feasible because the filenamify package is quite small, and I guess @sindresorhus wouldn't mind having his code copied to here.

It's MIT licensed so I believe we could inline the equivalent code that we'd need in our own codebase with appropriate referencing if we wanted to.

> Note: what is a chromium helper? Maybe that's the answer but I don't know what it is 😅

When I say "Chromium Helper" I mean that chromium probably has a method we can use in it's `base` or `path` libraries to do this kind of filename sanitization.

papb:
> We've never really run into this case because as mentioned below, Chromium normally has all the helpers / utilities we could dream of. In most cases we just implement the code ourselves or use Chromium's methods.

:+1:

> It's MIT licensed so I believe we could inline the equivalent code that we'd need in our own codebase with appropriate referencing if we wanted to.

:+1:

> When I say "Chromium Helper" I mean that chromium probably has a method we can use in it's `base` or `path` libraries to do this kind of filename sanitization.

Hmm, this is my first time contributing to Electron, where can I find docs on this so I can try to search for an existing method? Is it a JavaScript library like any other? I assume not because otherwise it would be in `dependencies`, right? :grimacing:

papb:
> PR description included and stakeholders cc'd

How do I know who are the stakeholders so I can cc them?

papb:
Ok, I have no idea why CI is failing on linux. Seem to be some C++ issues? If it has something to do with my PR, that's pretty obscure :open_mouth: 

@MarshallOfSound Can you please help me understand what is wrong?

codebytere:
hey @papb! sorry for the delay here. what @MarshallOfSound means by Chromium helper is that we want to leverage an abstraction that already exists in the [Chromium](https://source.chromium.org/chromium/chromium/src/+/) codebase - in this case, we'll probably want to make a change here:
https://github.com/electron/electron/blob/7b55a70a3673fc76ee6ff9e50577ca72536606fd/shell/browser/api/electron_api_app.cc#L910-L933

papb:
Hello @codebytere! Please let me know if I understood you correctly: what you are saying basically is that instead of applying this fix in the TypeScript code as I currently have done in this PR, I should do an equivalent fix in this C++ code instead?

If yes, may I ask why do it in `App::SetPath` and not in `base::PathService::OverrideAndCreateIfNeeded`? (Note: I am just guessing based on the names, I don't know anything about these functions)

Also your link to Chromium is broken :sweat_smile:

papb:
Can anyone help me move this forward please?

ckerr:
ping @MarshallOfSound and possibly @codebytere for followup

miniak:
@papb the code that you are modifying has been refactored by #29649, please submit a new PR in case you are still interested in fixing this issue.

PR #23760, opened 2020-05-26
chore: remove obsolete native node module patch

codebytere:
#### Description of Change

We can now remove this patch as it's been released in [`node-gyp v5.1.1`](https://github.com/nodejs/node-gyp/releases/tag/v5.1.1). 

Release proposal PR here: https://github.com/nodejs/node-gyp/pull/2125

cc @zcbenz @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


deepak1556:
Thanks for following up on this.

Question:

I believe this change gets picked up for users who have node-gyp installed separately and who can update the version. For users who use a node version manager, node-gyp usually comes from npm dependency, so this change will not be picked up until a corresponding npm version is out. Is this correct ?

In which case I would suggest to float this patch in electron until v12 hits maintenance mode, that way the change is available to all users independent of their local node toolchain installation.

codebytere:
@deepak1556 that's correct - the only caveat is that the version bundled and distributed within Node.js is not made directly available for developers to use and is intended only to support the ability to use `npm install` to compile and install addons. 

`npm` does get updated in LTS versions though (last update was `v12.16.2`) so i think there's potential for it to pick up this patch version and remove the potential issue before maintenance mode.

deepak1556:
> I don't see a particular urgency to remove this patch really so i'm happy to hinge this patch's removal on the next roll of npm into Node.js in v12 if that feels like the better option!

That would be great, since we have control on the gyp file used for compilation, its good to have this change around until the toolchains are updated. Lets make this PR a draft until then.

codebytere:
@deepak1556 Node.js v12 is now in maintenance mode - I think we can land this now.

deepak1556 (approved):
Great!

PR #23777, opened 2020-05-26
chore: ensure release notes always come from Clerk

ckerr:
As discussed in last week's WG Releases meeting in the 9.0.0 postmortem. Now with tests.

CC @electron/wg-releases 

#### Description of Change

* Ensure that the release notes generator always pulls text from Clerk's PR comments.
* Allow the release notes cache folder to be specified by environment variable
* Add tests

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


package.json
```diff
@@ -52,6 +53,7 @@
     "request": "^2.88.0",
     "semver": "^5.6.0",
     "shx": "^0.3.2",
+    "sinon": "^9.0.1",
```
jkleinsc, line 56: These changes should be in spec-main/package.json

yarn.lock
```diff
@@ -135,6 +135,42 @@
   dependencies:
     any-observable "^0.3.0"
 
+"@sinonjs/commons@^1", "@sinonjs/commons@^1.6.0", "@sinonjs/commons@^1.7.0", "@sinonjs/commons@^1.7.2":
```
jkleinsc, line 138: These changes should be in spec-main/yarn.lock

package.json
ckerr: I've moved the sinon dependencies into spec-main/ in 839eebe12f4721c6f0d3e2f5f2bfefe1b8dce9eb

I'm a little unclear on the division of dependencies? The change you suggested is what makes intuitive sense to me, but I don't undersatnd why the top-level package.json is where the dependencies on @types/chai, @types/chai-as-promised, @types/dirty-chai, @types/mocha, eslint-plugin-mocha are specified...?

yarn.lock
ckerr: Moved in 839eebe

package.json
```diff
@@ -142,4 +142,4 @@
   "dependencies": {
     "@types/temp": "^0.8.34"
   }
-}
```
jkleinsc, line 145: This change shouldn't be committed

package.json
jkleinsc: The way I see it is that the dependencies should be located where they are used.  If we are using them in multiple places then it makes sense to put them at the top level.

PR #23782, opened 2020-05-27
fix: volume key globalShortcut registration

codebytere:
#### Description of Change

Fixes a  NOTREACHED DCHECK crash [here](https://source.chromium.org/chromium/chromium/src/+/master:ui/base/accelerators/system_media_controls_media_keys_listener.cc;l=69?q=system_media_controls_media_keys_listener.cc&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2Fsearch%2F) borne of the fact that we patched  `MediaKeysListener::IsMediaKeycode` and then did not add the three extra keys to the switch statement. Also adds the three missing keys to the Accessibility check.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue with volume-related `globalShortcut` registration.

### DIFF
--- a/patches/chromium/command-ismediakey.patch
+++ b/patches/chromium/command-ismediakey.patch
@@ -3,17 +3,8 @@ From: Jeremy Apthorp <jeremya@chromium.org>
 Date: Wed, 10 Oct 2018 15:07:34 -0700
 Subject: command-ismediakey.patch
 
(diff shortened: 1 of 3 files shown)


codebytere:
@deepak1556 that was just the automatic patch linter failing, not the patch itself. Fixed!

PR #23785, opened 2020-05-27
fix: make process.windowsStore work reliably

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/18161.

The `length` was not initialized when passed to `GetPackageFamilyName`, resulting in `process.windowsStore` being random undefined in difference versions.

#### Release Notes

Notes: Fix `process.windowsStore` returning undefined in AppX packages.


deepak1556:

PR #23819, opened 2020-05-28
chore: Revert "fix: trigger activate event when app is activated via app switcher"

zcbenz:
Reverts electron/electron#23727 since it is causing a breaking change. See https://github.com/electron/electron/pull/23727#issuecomment-635119103 for more.

Notes: none


jkleinsc:

PR #23842, opened 2020-05-29
feat: add V8CacheOptions webpreference

deepak1556:
#### Description of Change

Refs https://github.com/microsoft/vscode/issues/98682#issuecomment-635159734 for context

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: V8CacheOptions is a new webpreference option to enforce code caching policy

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -387,7 +387,7 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       Default is `true`.
     * `enableWebSQL` Boolean (optional) - Whether to enable the [WebSQL api](https://www.w3.org/TR/webdatabase/).
       Default is `true`.
-    * `V8CacheOptions` String (optional) - Enforces the v8 code caching policy
+    * `v8CacheOptions` String (optional) - Enforces the v8 code caching policy
       used by blink. Accepted values are
       * `none` - Disables code caching
       * `code` - Heuristic based code caching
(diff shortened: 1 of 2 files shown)


bpasero:
This is awesome, and my 2 cents are that Electron by default should cache on the first startup as much as possible because an Electron app deserves to be cached right from the beginning, as opposed to any website I visit.

docs/api/browser-window.md
```diff
@@ -387,6 +387,13 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       Default is `true`.
     * `enableWebSQL` Boolean (optional) - Whether to enable the [WebSQL api](https://www.w3.org/TR/webdatabase/).
       Default is `true`.
+    * `V8CacheOptions` String (optional) - Enforces the v8 code caching policy
```
jkleinsc, line 390:
```suggestion
    * `v8CacheOptions` String (optional) - Enforces the v8 code caching policy
```
To match the other properties (start with lowercase).

deepak1556: Done, thanks!

pushkin-:
@deepak1556 So is this recommendation to enable this for our apps? Are there case where we wouldn't want to do that? Should this be the default?

deepak1556:
Couple of things to remember,

* This flag only affects scripts served by chromium in the render process, scripts served by Node.js will not be affected.

* If your app uses service workers to serve resources, then you need not use these flags. You will get the more aggressive code caching for free.

* `bypassHeatCheck` is a sane default for all electron apps.

* `bypassHeatCheckAndEagerCompile` should be used cautiously, depending on the amount of code being eagerly compiled and cached, it can cause increase in memory and add to GC pressure, also causes increased time in deserializing the cached data which translates to increased code load time and then there is bytecode flushing GC that will clear up inactive code, so most of the cached data may not end up being used.

deepak1556:
> Should this be the default?

Electron 11 is set to be a quiet release, so we don't do any breaking changes in master at the moment. Will switch the default in near future.

Issue #11587, opened 2018-01-08
Accessibility issue: Electron system tray context menu on Windows is not keyboard navigable


josimi:
* Electron version: 1.7.10
* Operating system: Windows 10

### Expected behavior

It is possible to access system tray items and their respective context menus on Windows using the keyboard only. This functionality is important for accessibility reasons.

The system tray is selected using the key combination **WINDOWS+B** and then opened using the **ENTER** key. It is then possible to navigate to a specific icon using the arrow keys and select it with **ENTER** to open the icon's context menu. It _should_ be possible to then navigate through this menu using the arrow keys and select a specific item with **ENTER**.

### Actual behavior

Arrow key navigation and ENTER selection is not available for Electron apps' system tray icon context menus. Pressing the arrow keys while the system tray icon's context menu is open, after being opened using the keyboard steps above, produces no result.

### How to reproduce

1. **WINDOWS+B** and then **ENTER** to open the list of icons
2. Arrow keys and to select the Electron app's icon
3. Press **SHIFT+F-10** to open the icon's context menu
4. Observe that arrow keys are not useable within this icon's menu, but are for other icons.

### Technical details

See **notify_icon.cc**'s function `NotifyIcon::PopUpContextMenu` which makes a call to the `->RunMenuAt`. I am new to the Electron codebase, but as far as I can tell, this is `RunMenuAt` call is defined in Chromium, meaning we can't change it directly, but will need to find another way to show context menus that is keyboard accessible.

josimi:
Looks like the missing key press handling functionality should be in Chromium's `/ui/views/controls/menu/menu_controller.cc`.

brentgti:
Did regression occur with this? I created an application with a tray menu, and I can't navigate it with the arrow keys on my keyboard. My code is here: https://github.com/brentgti/electron-Teams-links.

brentgti:
@josimi, is this issue actually fixed, or maybe there's a regression issue here? My tray menu doesn't allow for navigation via keyboard. See https://github.com/brentgti/electron-Teams-links.

danikane:
I second that. Can't navigate with the keyboard.

- Electron v7.0.1
- electron-builder 21.2.0

danikane:
Update:
Upgraded to Electron v8.1.1 and electron-builder 22.4.1, same behavior.
Thanks!

brentgti:
[...]

PR #23843, opened 2020-05-29
fix: do not use CONTEXT_MENU flag for tray menu

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/11587.

Tray menu does not popup on a window, so it should not use `CONTEXT_MENU` flag otherwise some things would not work. The workaround used in https://github.com/electron/electron/pull/15302 stopped working after some Chromium upgrades.

This PR makes our tray menu code aligned with Chromium `status_icon_win.cc`.

#### Release Notes

Notes: Fix tray menu on Windows not keyboard navigable.


ckerr:
https://github.com/electron/electron/pull/23897

Issue #22951, opened 2020-04-03
Mac OS - HTTP authentication - NTLMv2 authentication not working


christophpre:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 6.1.9, 7.2.1, 8.2.0, 9.0.X
* **Operating System:**
  * macOS 10.15.4
* **Last Known Working Electron version:**
  * Never

### Expected Behavior
1. When authenticating via HTTP authentication and Proxy/Server only allows NTLMv2, authentication should work.
2. When authenticating via HTTP authentication and Proxy/Server negotiates protocol and allows NTLMv1 and NTLMv2, Electron should always use NTLMv2.

### Actual Behavior
1. When authenticating via HTTP and the Proxy/Server only allows NTLMv2 and 'login' event is triggered in Electron and **Password/Username is entered correctly**, authentication does not work. Windows Event Log Error Code 0xc000006d / Substatus 0xc000006a (https://docs.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4625)
2. When authenticating via HTTP authentication and Proxy/Server negotiates protocol, Electron always uses NTLMv1.

### To Reproduce
1. Configure all Domain Controllers for Windows server to only allow NTLMv2
2. Try to authenticate with correct credentials via Electron 'login'-Event
-> 'login'-Event in Electron is triggered again and the described Error Codes can be seen in Windows Event Log

### Screenshots
![image](https://user-images.githubusercontent.com/6893212/78341915-be83aa00-7598-11ea-929b-836cca0097a5.png)
![image](https://user-images.githubusercontent.com/6893212/78341931-c4798b00-7598-11ea-8c8e-c44d85b41ff6.png)

### Additional Information
* This issue #17111 states, that Chrome should use NTLMv2 per default since Chromium 65
* NTLMv2 in Electron works on Windows 10 Version 1903 (OS Build 18362.476)
* NTLMv2 is used and works per default on Chromium Browser Version 80.0.3987.158, which is used in current stable Electron Version 8.2.0, where it does not work

christophpre:
Any news on this?

[...]

PR #23846, opened 2020-05-29
fix: default to NTLM v2 in the network service for POSIX platforms

deepak1556:
#### Description of Change

Fixes https://github.com/electron/electron/issues/22951

This has been the default for a while at the //net layer https://bugs.chromium.org/p/chromium/issues/detail?id=22532 ,  the network service configures the default incorrectly. I have reported it in upstream, will update here once I have a response.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: enable NTLM v2 for POSIX platforms and added --disable-ntlm-v2 switch to disable it.


deepak1556:
Marking WIP pending upstream fix.

deepak1556:
Upstream PR has been merged.

deepak1556:
Since this is a breaking change, I will change the PR to keep the default as NTLM v1 for the current release lines, will add a flag `--enable-ntlm-v2` for apps and since `11-x-y` is a quiet release, will do the same for master. After `11-x-y` is branched out will change the default to NTLM v2 and flip the flag to `--disable-ntlm-v2`

deepak1556:
Looking into this a bit more, NTLM v2 was made default in [68.0.3416.0](https://chromium-review.googlesource.com/c/chromium/src/+/1030250/) which went out with our Electron 4 release. This broke accidentally after network service codepath was used to initialize the http auth preferences since Electron 6. Hence this should be treated as a bug fix rather than breaking change,

Also NTLM v2 is more widely used authentication protocol than v1 if at all NTLM is used. Hence enabling this as default with a flag to disable is a good change.

deepak1556:
PR is ready for review. Thanks!

PR #23874, opened 2020-06-01
test: wait for beforeunload handler to be installed

zcbenz:
#### Description of Change

After the beforeunload handler is installed, the render will send a message to main process to update the terminate status of frame host (it is called SuddenTerminationStatus message). So there is a race condition when we try to close the window immediately after installing the handler.

This PR makes the test wait for a console message before closing, which should be sent after the SuddenTerminationStatus message.

There is also a cleanup of the beforeunload fixture files.

Note that this PR only changes the tests that install handler dynamically, I'll start a new PR to change the tests that hardcoded the handler in html files, if needed in future.

#### Release Notes

Notes: none


spec-main/fixtures/api/beforeunload-false-prevent3.html
```diff
@@ -2,6 +2,7 @@
 <body>
 <script type="text/javascript" charset="utf-8">
   function installBeforeUnload(removeAfterNTimes) {
+    setTimeout(() => console.log('installed'))
```
deepak1556, line 5: wouldn't it be better if this is added below the event listener, so that its more explicit about the order.

```
window.addEventListener('beforeunload', function handler(e) {})
setTimeout(() => console.log('installed'));
```

deepak1556:
> it is called SuddenTerminationStatus message

TIL, wow there are so many internal details around unload handlers. Should document these findings if we get a chance.

PR #23901, opened 2020-06-02
test: load page with some content when testing extension

zcbenz:
#### Description of Change

When loading a completely blank page, the `did-finish-load` event may emit immediately instead of waiting for the script in extensions to run. Adding some content in the page can probably reduce the flakiness of extension tests, and it would help debugging if the test is still flaky. 

There is also a slight change to the extension so its code can overwrite page's content.

#### Release Notes

Notes: none


deepak1556:
Would additionally relying on [WebContentsObserver::DOMContentLoaded](https://source.chromium.org/chromium/chromium/src/+/master:content/public/browser/web_contents_observer.h;l=248-251) be a good signal, we already expose an event on it `dom-ready`.

zcbenz:
> Would additionally relying on [WebContentsObserver::DOMContentLoaded](https://source.chromium.org/chromium/chromium/src/+/master:content/public/browser/web_contents_observer.h;l=248-251) be a good signal, we already expose an event on it `dom-ready`.

Yeah that's better 👍

PR #23903, opened 2020-06-02
fix: correctly support the --inspect-brk-node flag

MarshallOfSound:
Very useful for profiling Electron's boot process.

Notes: none


shell/browser/node_debugger.cc
```diff
@@ -35,6 +35,11 @@ void NodeDebugger::Start() {
                            debug_options.host_port),
                        true /* is_main */))
     DCHECK(env_->inspector_agent()->IsListening());
+
+  if (env_->inspector_agent()->options().break_node_first_line) {
```
deepak1556, line 39: `env_->inspector_agent()` => `inspector`

deepak1556:
Any reason not to backport to `8-x-y` ?

PR #23917, opened 2020-06-02
fix: throw when using ipcRenderer after context released

nornagon:
... instead of crashing.

#### Description of Change
This fixes a possible segfault that was introduced [here](https://github.com/electron/electron/pull/23321/files#diff-c330abcb8a74593f6eced7f04ff261fcR67). This crash could happen if the `ipcRenderer` module was used after the context was released.

This change makes this pattern throw a JS error instead of crashing.

I'm not convinced that this is the right behavior, ultimately. The initial fix in #23321 works around a memory leak that can happen if the main process leaks an `invoke` promise without resolving it. I think that #23321 is potentially preventing legitimate uses that retain the context after blink has released it, though I don't have any concrete examples. We should consider reverting #23321 and leaving it up to apps to resolve their own memory leaks rather than patching over the problem by neutering the IPC object on context release.

However, it's better to throw a JS exception than to crash, so this is at least not worse than the current behavior.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash that could occur when using the `ipcRenderer` module after blink had released the context. Instead, a JS exception will be thrown.

### DIFF
--- a/shell/renderer/api/electron_api_ipc_renderer.cc
+++ b/shell/renderer/api/electron_api_ipc_renderer.cc
@@ -16,6 +16,7 @@
 #include "shell/common/api/api.mojom.h"
 #include "shell/common/gin_converters/blink_converter.h"
 #include "shell/common/gin_converters/value_converter.h"
+#include "shell/common/gin_helper/error_thrower.h"
 #include "shell/common/gin_helper/function_template_extensions.h"
 #include "shell/common/gin_helper/promise.h"
 #include "shell/common/node_bindings.h"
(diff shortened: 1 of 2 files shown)


shell/renderer/api/electron_api_ipc_renderer.cc
```diff
@@ -83,9 +87,14 @@ class IPCRenderer : public gin::Wrappable<IPCRenderer>,
 
  private:
   void SendMessage(v8::Isolate* isolate,
+                   gin_helper::ErrorThrower thrower,
```
codebytere, line 90: If we're introducing an `ErrorThrower` we can pull the isolate off of it with `thrower.isolate()` - we probably don't need to keep the isolate as a separate param

nornagon: True, though it doesn't seem particularly compelling to me to do so. Do you feel strongly about it?

codebytere: not per se but we've preferred to consolidate and avoid excess isolates everywhere else in the codebase so this is a bit of a break from existing style

Issue #23211, opened 2020-04-22
regression: Session#setUserAgent ignores its second acceptLanguages param


bughit:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 9.0.0-beta.18 x86
* **Operating System:**
  * win7 x64 sp1
* **Last Known Working Electron version:**
  * 4.2.9 x86

### Expected Behavior
Per the docs and the original implementation, `acceptLanguages` param should be controlling the Accept-Language header.

https://github.com/electron/electron/blob/4-2-x/atom/browser/api/atom_api_session.cc#L612
https://www.electronjs.org/docs/api/session#sessetuseragentuseragent-acceptlanguages

### Actual Behavior
`acceptLanguages` param is ignored, Accept-Language remains at its default.

### To Reproduce
- the current user os (win7) lang is "en-US"
- run the following app in 4.2 and 9.0
- inspect the request in the renderer devtools
- in 4.2, Accept-Language is "fr"
- in 9.0 it's "en-US"

```js
'use strict';

const {app, BrowserWindow} = require('electron');

let win;

app.once('ready', () => {

  win = new BrowserWindow({
    webPreferences: {
      nodeIntegration: false
    }
  });

  win.webContents.session.setUserAgent('', 'fr');

  win.webContents.openDevTools({mode: 'right'});
  win.loadURL('about:blank');
  setTimeout(() => win.loadURL('https://example.org/'), 3000);

});
```

### Additional Information
The original implementation of setUserAgent was removed at some point and it was later re-implemented by #20014 without acceptLanguages  support.

bughit:
in case it matters, this is not windows specific

PR #20014, opened 2019-08-28
fix: nws13n: make ses.setUserAgent work 

nornagon:
#### Description of Change
This fixes `session.setUserAgent` in the post-network-service world. Ref https://chromium-review.googlesource.com/c/chromium/src/+/1585083. Ref #19602.

#### Checklist

- [ ] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [ ] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [ ] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: none

Issue #2396, opened 2015-08-01
webview node integeration on & require 'module'


skandasoft:
I have a webrower-plus(http://atom.io/packages/browser-plus) package for atom. and I am using webview with node-integeration-on but this has problem when showing web pages which has require module...https://github.com/skandasoft/browser-plus/issues/18

frankhale:
Looks like the node require is getting in the way. I suspect you'd need to do some sort of "intelligent" switching around to move it out of the way, eg. assigning the node require to another variable so requirejs's require could work. The "intelligent" part is magic because if you are using node-integration and loading random pages off the net that use requirejs then things are going to get ugly unless you can "know" eg parse the pages to see if they use requirejs or some other require that would break your webview. 

Having node integration turned on for a web browser plugin seems very risky unless you know exactly what and have control over the pages you are loading.

skandasoft:
I have move away from node-integeration on for the browser-plus..I am not sure if this ticked is need to open to check the possibility for node-integeration with require module

zcbenz:
This is a quite classic problem of browsers with node integration, see also:
https://github.com/nwjs/nw.js/wiki/faq-name-conflict

PR #23944, opened 2020-06-03
fix: use acceptLanguages argument in session.setUserAgent()

miniak:
#### Description of Change
The `acceptLanguages` argument was lost when re-implementing [`session.setUserAgent()`](https://github.com/electron/electron/blob/master/docs/api/session.md#sessetuseragentuseragent-acceptlanguages) in #20014. It's still documented.

Fixes https://github.com/electron/electron/issues/23211

#### Checklist
- [x] `npm test` passes

#### Release Notes

Notes: Fixed the `acceptLanguages` argument being ignored in `session.setUserAgent()`.

### DIFF
--- a/shell/browser/api/electron_api_session.cc
+++ b/shell/browser/api/electron_api_session.cc
@@ -40,6 +40,7 @@
 #include "net/http/http_auth_handler_factory.h"
 #include "net/http/http_auth_preferences.h"
 #include "net/http/http_cache.h"
+#include "net/http/http_util.h"
 #include "services/network/network_service.h"
 #include "services/network/public/cpp/features.h"
 #include "shell/browser/api/electron_api_app.h"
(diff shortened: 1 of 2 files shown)


shell/browser/api/electron_api_session.cc
```diff
@@ -590,10 +590,15 @@ void Session::AllowNTLMCredentialsForDomains(const std::string& domains) {
 
 void Session::SetUserAgent(const std::string& user_agent,
                            gin::Arguments* args) {
+  std::string accept_lang = g_browser_process->GetApplicationLocale();
```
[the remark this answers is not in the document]
miniak: but I can change it

shell/browser/api/electron_api_session.cc
```diff
@@ -591,9 +591,15 @@ void Session::AllowNTLMCredentialsForDomains(const std::string& domains) {
...
-      ->GetNetworkContext()
-      ->SetUserAgent(user_agent);
+  auto* network_context = content::BrowserContext::GetDefaultStoragePartition(
+                              browser_context_.get())
+                              ->GetNetworkContext();
+  network_context->SetUserAgent(user_agent);
+
+  std::string accept_lang;
+  if (args->GetNext(&accept_lang)) {
+    network_context->SetAcceptLanguage(accept_lang);
```
deepak1556, line 601: Should be

`network_context->SetAcceptLanguage(net::HttpUtil::GenerateAcceptLanguageHeader(accept_lang));`

deepak1556: Also should add a test that gives an input of accept languages like `en-US,fr,de` , so the generated header is validated and turns out as `en-US,fr;q=0.9,de;q=0.8`

codebytere:
@miniak it looks like this fails on arm on older release lines - my tendency to ignore arm failures got the best of me. 

See https://github.com/electron/electron/pull/2396 for example - it's passing everywhere else which is weird.

PR #23960, opened 2020-06-04
docs: fix darwin auto-quit template

BenSova:
#### Description of Change
I fixed the macOS auto-quit template in `first-app.md`. 
Before it said if the platform was **not** darwin it would quit the app (which makes the code do nothing.) So, I changed it so it will quit if it **is** darwin.

#### Release Notes

Notes: none


malept:
Thanks, but the code works as intended. Here's what the comment right above it says:

> On macOS it is common for applications and their menu bar to stay active until the user quits explicitly with Cmd + Q

BenSova:
@malept I am a macOS user (that's why I added the code in the first place) and I used the code from the documentation and it didn't quit the app, it just stayed open waiting for my to press cmd+q or create a new instance. Once I changed the code from `!==` to `===` it worked. It quit when the last instance of the app was closed.

erickzhao:
@BenSova Mac apps are generally expected to stay open even after you click the big red button.

See [macOS User Guide](https://support.apple.com/guide/mac-help/quit-apps-mchl834d18c2/mac):
> Note: When you click the red close button  in the top-left corner of an app’s window, the window closes, but the app remains open. See Manage windows.

BenSova:
@malept @erickzhao I understand what you guys are talking about. I use macOS all the time, so I am used to that behavior. However, the code there designed to close the app when all windows are closed. It says so right above:
> Quit when all windows are closed.
I'm not necessarily mad, I'm sure anyone who's cared about doing that figured it out. I just want to point out that the code doesn't do what it's designed to do.

malept:
Arguably, the comment "Quit when all windows are closed" should be changed, not the code itself.

BenSova:
@malept, good point. I was wondering why that was even included until I ran a test without that code and the app quit by itself. I'm going to fix the comment (and undo my change), so it makes more sense.

Issue #23413, opened 2020-05-05
Handle custom protocol in filter of webRequest.onBeforeRequest


mahnunchik:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

I've faced with the issue that I couldn't catch request by `webRequest.onBeforeRequest` if custom protocol was used in `filter`.

```js
mainWindow.webContents.session.webRequest.onBeforeRequest({
    urls: ['myscheme://*/*'],
}, (details, cb) => {
    console.log('onBeforeRequest:', details.url);
    cb({ cancel: false });
});
```

* **Electron Version:**
  * v8.2.5
* **Operating System:**
  * macOS 10.15.4

### Expected Behavior

Message in console like:

```bash
onBeforeRequest: myscheme://action/foobar
```

### Actual Behavior

The error on `onBeforeRequest` a call.

```
Uncaught Exception:
TypeError: Invalid url pattern myscheme://*/*: Wrong scheme type.
```

### To Reproduce

```js
mainWindow.webContents.session.webRequest.onBeforeRequest({
    urls: ['myscheme://*/*'],
}, (details, cb) => {
    console.log('onBeforeRequest:', details.url);
    cb({ cancel: false });
});
```

### Additional Information

Original question: https://stackoverflow.com/questions/61610086/typeerror-invalid-url-pattern-myscheme-wrong-scheme-type

deepak1556:
@mahnunchik can you explain why you would want to intercept custom protocol requests with webrequest module, you should be able to handle any request related to the protocol in the handler you have registered.

mahnunchik:
@deepak1556 

1. Window related (local) protocol handler.
2. Inspect any requests in `onBeforeRequest`.

deepak1556:
Thanks for the context,

> Window related (local) protocol handler.

These requests are not handled by the webRequest module

If its a protocol not registered by your app then you can use [session.setPermissionRequestHandler](https://github.com/electron/electron/blob/master/docs/api/session.md#sessetpermissionrequesthandlerhandler) and listen for the `openExternal` permission to reject or accept it.

[...]

PR #23972, opened 2020-06-04
fix: allow custom protocols as onBeforeRequest schemes

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/23413.

Invalid scheme errors match only against Chromium's [valid schemes](https://source.chromium.org/chromium/chromium/src/+/master:extensions/common/url_pattern.cc;drc=4a8c52892b63e88acb5508f9b4c9f805a4b244e1;bpv=1;bpt=1;l=31?originalUrl=https:%2F%2Fcs.chromium.org%2F) which means that custom protocols would false-positively error.

To fix this, we should choose to be more permissive and allow potentially invalid schemes rather than disallow any custom procotols.

cc @nornagon @zcbenz @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where custom protocols would error in `webRequest.onBeforeRequest`.

### DIFF
--- a/shell/browser/api/electron_api_web_request.cc
+++ b/shell/browser/api/electron_api_web_request.cc
@@ -376,10 +376,20 @@ void WebRequest::SetListener(Event event,
     if (result == URLPattern::ParseResult::kSuccess) {
       patterns.insert(pattern);
     } else {
-      const char* error_type = URLPattern::GetParseResultString(result);
-      args->ThrowTypeError("Invalid url pattern " + filter_pattern + ": " +
-                           error_type);
-      return;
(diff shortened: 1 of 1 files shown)


deepak1556:
Custom protocols have handlers registered by users which by itself acts like a webrequest module, all requests for the protocol will hit the handler. IMO webrequest module should only be used to intercept requests for protocols that users don't have control over, thats why we didn't support this pattern from the beginning.

I would like to see a use case for this kind of interception before pushing this fix forward. I have commented on the issue.

codebytere:
Closing in light of comments on the original issue.

PR #22770, opened 2020-03-19
fix: don't allow window to go behind menu bar on mac

CezaryKulakowski:
This change fixes #22667.

Notes: don't allow window to go behind menu bar on mac

PR #23976, opened 2020-06-04
fix: Allow windows behind macOS elements if frame = false

felixrieseberg:
#### Description of Change

Closes https://github.com/electron/electron/issues/24000.

In #22770, we ensured that windows cannot be moved behind the menu bar or be too close to the dock. That was a good idea - we wanted to make sure that windows can always be moved by the user.

However, some apps (Slack) are using frameless windows to purposefully draw content onto the screen. In Slack's case, that's a red border, only six pixels tall, that goes around the screen. With the fix mentioned above, we couldn't get the window as close to the screen's boundary as we wanted to. That turned the PR above into a breaking change without workarounds available.

With this proposed fix, I'm allowing frame-less windows to be positioned exactly where the developer wants them to be, even if that's behind the dock or the menu bar. I'm hoping that it's a good compromise between "helping the developer" and ensuring that previously covered use cases remain possible.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix: Allow windows behind macOS elements if "frame" is false.

### DIFF
--- a/shell/browser/ui/cocoa/electron_ns_window.mm
+++ b/shell/browser/ui/cocoa/electron_ns_window.mm
@@ -91,8 +91,20 @@ bool ScopedDisableResize::disable_resize_ = false;
 
   NSRect result = [super constrainFrameRect:frameRect toScreen:screen];
   // Enable the window to be larger than screen.
-  if ([self enableLargerThanScreen])
-    result.size = frameRect.size;
+  if ([self enableLargerThanScreen]) {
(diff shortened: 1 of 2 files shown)


CezaryKulakowski:
This change causes problem when application creates frameless window and creates its own bar for dragging on top of it. In such case you can place window behind main menu bar and make it impossible to drag it again. I can attach testing application which shows the problem.

PR #23979, opened 2020-06-04
fix: throw instead of crash when using ipcRenderer after context released

nornagon:
Backport of #23917

Notes: Fixed a crash that could occur when using the `ipcRenderer` module after blink had released the context. Instead, a JS exception will be thrown.

### DIFF
--- a/spec-main/api-ipc-renderer-spec.ts
+++ b/spec-main/api-ipc-renderer-spec.ts
@@ -9,7 +9,7 @@ describe('ipcRenderer module', () => {
 
   let w: BrowserWindow
   before(async () => {
-    w = new BrowserWindow({ show: false, webPreferences: { nodeIntegration: true } })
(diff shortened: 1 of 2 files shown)


spec-main/api-ipc-renderer-spec.ts
```diff
@@ -192,4 +192,25 @@ describe('ipcRenderer module', () => {
       expect(result).to.deep.equal([])
     })
   })
-})
+
+  describe('after context is released', () => {
+    it('throws an exception', async () => {
+      const error = await w.webContents.executeJavaScript(`(${() => {
+        const child = window.open('', 'child', 'show=no,nodeIntegration=yes')! as any;
+        const childIpc = child.require('electron').ipcRenderer;
```
nornagon, line 200: @MarshallOfSound do you have any idea why `child.require` might not be defined here? This same logic works in 9.x+.

codebytere: @nornagon i suspect it's related to https://github.com/electron/electron/pull/19703

PR #23985, opened 2020-06-05
ci: use call instead of cmd /C

thypon:
#### Description of Change

Since we want to import the environment change from RefreshEnv.cmd
we also want to share the same execution context of the
caller.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes


nornagon:
Is this script actually invoked anywhere in our CI processes?

jkleinsc:

Issue #629, opened 2014-09-06
[build] [windows] Build failed Error: %1 is not a valid Win32 application


nhandl3:
I have tried with:
- nodejs 32bit 0.10.31
- build nodejs from master branch (0.13.0-pre)
- cygwin
- cmd
- Visual Studio cmd.

All cause the same error as below.
Move to shorter path (d:\toollib\atom-editor\atom-shell -> c:\atomshell) can't help.
I cannot figure out what kind of this problem is. Please help, thank you.

```
Microsoft Windows [Version 6.1.7601]
Copyright (c) 2009 Microsoft Corporation.  All rights reserved.

d:\toollib\atom-editor\atom-shell>python script\bootstrap.py -v
Running in verbose mode
Synchronizing submodule url for 'vendor/breakpad'
Synchronizing submodule url for 'vendor/brightray'
Synchronizing submodule url for 'vendor/depot_tools'
Synchronizing submodule url for 'vendor/native_mate'
Synchronizing submodule url for 'vendor/node'



> atom-shell@0.15.9 preinstall d:\toollib\atom-editor\atom-shell
> node -e 'process.exit(0)'



module.js:356
  Module._extensions[extension](this, filename);
                               ^
Error: %1 is not a valid Win32 application.
d:\toollib\atom-editor\atom-shell\node_modules\atom-package-manager\node_modules\keytar\build\Release\keytar.node
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (d:\toollib\atom-editor\atom-shell\node_modules\atom-package-manager\node_modules\keytar\lib\keytar.js:4:12)
    at Object.<anonymous> (d:\toollib\atom-editor\atom-shell\node_modules\atom-package-manager\node_modules\keytar\lib\keytar.js:58:4)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)

Traceback (most recent call last):
  File "script\bootstrap.py", line 119, in <module>
    sys.exit(main())
  File "script\bootstrap.py", line 25, in main
    update_atom_modules('spec')
  File "script\bootstrap.py", line 73, in update_atom_modules
    execute([apm, 'install'])
[...]

PR #23993, opened 2020-06-05
chore: cherry-pick 6e6b3096ffa9 from chromium

nornagon:
[m83] fido: improve guards against adding authenticators with identical IDs

Make FidoRequestHandler::AuthenticatorAdded() return early when an
FidoAuthenticator is added whose ID matches that of a previously added
authenticator. The request handler  previously did not add the
duplicate authenticator into its |active_authenticators_| map, but then
attempted to dispatch its request to it (or rather to an invalid
reference).

Also better guard against authenticators being removed during
initialization by making the (asynchronously run)
InitializeAuthenticatorAndDispatchRequest() method look up the
AuthenticatorState for the authenticator to be initialized by its ID
rather than passing around AuthenticatorState pointers that may have
been freed by the time the method runs because the authenticator went
away.

Lastly, derive VirtualFidoDevice IDs randomly. It previously used its
instance pointer address for "randomness" which, aside from being weird,
could lead to re-use of IDs. (FidoAuthenticator ID reuse in itself
_should_ not be a problem, but certainly could lead to bugs if the rest
of the code is less than careful about it.)

(cherry picked from commit bdbf8ce8f546e6bc3e2c2fe7e68ac4ae6156c5b6)

Bug: 1082105
Change-Id: Ie4e3fd39c3360bf0131cdd6dd33b2be4dbb225a8
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2203776
Commit-Queue: Martin Kreichgauer <martinkr@google.com>
Reviewed-by: Christopher Thompson <cthomp@chromium.org>
Reviewed-by: Adam Langley <agl@chromium.org>
Cr-Original-Commit-Position: refs/heads/master@{#770190}
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2220073
Reviewed-by: Martin Kreichgauer <martinkr@google.com>
Cr-Commit-Position: refs/branch-heads/4103@{#629}
Cr-Branched-From: 8ad47e8d21f6866e4a37f47d83a860d41debf514-refs/heads/master@{#756066}

Notes:


nornagon:
Closing because:
1. patch significantly conflicts,
2. upstream bug is inacessible, so no known repro,
3. fido is not a commonly used feature in electron.

PR #24022, opened 2020-06-08
fix: correctly handle nexttick scheduling in stream reads

pfrazee:
#### Description of Change

We discovered some cases where protocol streaming responses would stall before completing. We isolated the issue to the case where a stream's read handler scheduled a `push()` using `nextTick()`, which (according to @mafintosh, who I trust to know about this sort of thing) will get scheduled to run prior to control being returned to C++. This means the call into js' `read()` will return nothing, but the `'readable'` event will get missed.

We were able to fix this by flagging when that occurs and automatically retrying the read.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix an issue which would cause streaming protocol responses to stall in some cases

### DIFF
--- a/shell/browser/net/node_stream_loader.cc
+++ b/shell/browser/net/node_stream_loader.cc
@@ -68,6 +68,8 @@ void NodeStreamLoader::Start(network::mojom::URLResponseHeadPtr head) {
 void NodeStreamLoader::NotifyReadable() {
   if (!readable_)
     ReadMore();
+  else if (is_reading_)
+    has_read_waiting_ = true;
   readable_ = true;
 }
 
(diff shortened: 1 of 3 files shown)


mafintosh:
Just to add a little context. nextTicks in Node er technically afterTicks as they run in the same event tick as the last thing (legacy naming).

In practice this bug manifest itself almost immediately on any stream returned that’s the last stream in a pipeline as they will have lots of these “read returns null but data is added later in the tick” events

nornagon:
NodeStreamLoader needs a rewrite. It's extremely subtle and full of incorrect behaviors like this 😞

PR #24054, opened 2020-06-10
refactor: remove renderer-side refcount in remote

nornagon:
#### Description of Change
The refcount on the renderer side is always 1.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


deepak1556:
@nornagon can you please expand the PR description to why this change is being made, there is not much context at a glance.

nornagon:
@deepak1556 I'm working on refactoring remote to use WeakRef, and this is a step along the way of untangling and understanding the existing implementation to ensure that I reproduce it correctly using WeakRef. This is a piece of complexity that can be removed, as it serves no function, and avoiding re-implementing it saves effort.

deepak1556:
Thanks for the context!

PR #24055, opened 2020-06-10
fix: allow IME insert zero length composition string

rebornix:
#### Description of Change

Bug:
* Multiple IMEs no longer terminate after pressing shift https://bugs.chromium.org/p/chromium/issues/detail?id=1091069#c9
* Issue in VS Code https://github.com/microsoft/vscode/issues/98968

Expected behavior:
* IME quits composition mode after pressing shift

cc @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: fix issue with some IMEs on windows (for ex: Zhuyin) don't terminate after pressing shift

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -124,3 +124,4 @@ cherry-pick-86c02c5dcd37.patch
 fix_hunspell_crash.patch
 introduce_a_mutex_for_the_rendering_loop_in_baseaudiocontext.patch
 fix_default_to_ntlm_v2_in_network_service.patch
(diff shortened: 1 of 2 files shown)


deepak1556:
Originally @rebornix backported to `7-3-x` for verifying the fix locally, also this is his first PR in electron repo :) I will create backports to other branches.

deepak1556:
> but I recommend upgrading from Electron 7!

Currently Electron 8 upgrade blocked by https://github.com/microsoft/vscode/issues/97638 :( startup time is quite bad on windows. Still pending investigation.

Issue #24049, opened 2020-06-10
Tray menu creates taskbar item


zarubond:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
9.0.2
* **Operating System:**
  Windows 10
* **Last Known Working Electron version:**
9.0.1
### Expected Behavior
Opening tray icon should not create taskbar item

### Actual Behavior
Empty taskbar item is created when tray icon menu is opened

### To Reproduce
Create tray menu and open it

### Screenshots

![tray](https://user-images.githubusercontent.com/6961047/84270140-85346100-ab2a-11ea-9b75-45a40f4c2470.png)

MarshallOfSound:
For now if we don't have an immediate solution I think a revert makes sense.  Let's go back to the state we were in which is arguably better than having this thing appear

ckerr:
I'm going to keep look at this. I don't see an immediate solution but we need an answer that doesn't have visual artifacts *and* doesn't break tray menus for a11y users.

PR #24076, opened 2020-06-11
fix: Revert "fix: do not use CONTEXT_MENU flag for tray menu"

MarshallOfSound:
Reverts electron/electron#23843

As per #24049

Notes: no-notes


ckerr:
I don't love this, since it means we're re-breaking a11y. This change is a tradeoff between an important bug for a small set of a11y users vs a visual annoyance for a larger set of users.

Since tray a11y on Windows has been broken for several versions, it's understandable as an interim move -- but we still need to find a better fix for #11587.

PR #24097, opened 2020-06-12
fix: do not use CONTEXT_MENU flag for tray menu (reland)

### LINKED ISSUE #11587: Accessibility issue: Electron system tray context menu on Windows is not keyboard navigable

### LINKED ISSUE #24049: Tray menu creates taskbar item

zcbenz:
Close #11587, close #24049.

Relands https://github.com/electron/electron/pull/23843 with a fix for https://github.com/electron/electron/issues/24049.

By adding the `WS_EX_TOOLWINDOW` style, the menu window would no longer show in taskbar.

I'll file a bug and submit the patch to Chromium next week, I haven't figured out how to make Chrome browser show a status icon.

Notes: Fix tray menu showing in taskbar on Windows.


patches/chromium/remove_menu_window_task_item.patch
```diff
@@ -0,0 +1,19 @@
...
+
+diff --git a/ui/views/widget/widget_hwnd_utils.cc b/ui/views/widget/widget_hwnd_utils.cc
+index 40e66a212e3ea158b61f11d804a3c1644023b2d7..d3cdffc67216efa81f1c7394ee653a1d66207bd2 100644
+--- a/ui/views/widget/widget_hwnd_utils.cc
++++ b/ui/views/widget/widget_hwnd_utils.cc
+@@ -109,6 +109,7 @@ void CalculateWindowStylesFromInitParams(
+       break;
+     case Widget::InitParams::TYPE_MENU:
+       *style |= WS_POPUP;
++      *ex_style |= WS_EX_TOOLWINDOW;
```
MarshallOfSound, line 16: Is this safe, this looks like a well traversed function

deepak1556:
> I'll file a bug and submit the patch to Chromium next week, I haven't figured out how to make Chrome browser show a status icon.

@zcbenz as long as the exercised code path is same, you can create an issue with `content_shell` as an example. It is an accepted bug filing method when external developers can't make an example case with chrome.

patches/chromium/remove_menu_window_task_item.patch
zcbenz: This function is executed when calculating the styles of windows, so it is the correct place to add a style for a type of window.

My only concern is whether this would cause other side effects (I believe it won't according to the docs of this style), we will know once the patch gets reviewed in Chromium.

zcbenz:
Patch has been submitted to upstream: https://chromium-review.googlesource.com/c/chromium/src/+/2245941.

zcbenz:
The added patch has been approved in upstream.

deepak1556:
@zcbenz can you update the patch to match upstream CL ?

deepak1556:

Issue #18872, opened 2019-06-19
repl.start Causes Crash in Renderer


dsanders11:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 5.0.4
  * 6.0.0-beta.9
* **Operating System:** 
  * macOS 10.14.5
* **Last Known Working Electron version:**
  * 4.1.4
  * 5.0.0-beta.8

### Expected Behavior

`repl.start` should create a new `repl.REPLServer` in the renderer process

### Actual Behavior

A crash happens

### To Reproduce

* Branch with crash: https://github.com/dsanders11/electron-quick-start/tree/repl-renderer-crash
* Branch without crash (v4.1.4): https://github.com/dsanders11/electron-quick-start/tree/repl-renderer-crash-v4

### Additional Information

Does not crash in the main process. I ran through the individual v5 betas and see that the crash began happening with **5.0.0-beta.9**. It does not crash if you test on 5.0.0-beta.8.

Crash stacks are slightly different depending on which version of Electron you run, but all seem to go through `node::inspector::Agent::GetWorkerManager` in one of the frames.

I realize it doesn't make sense to start a `REPLServer` in this manner in the renderer process since it will default to `stdin`/`stdout`, but the crash also happens even if you provide your own stream for `input` and `output`. It's just a more concise reproduce case to let it use the default.

codebytere:
I've narrowed this to `NodeInspectorClient` being null at the time `node::inspector::Agent::Connect` is called, so it dies at `CHECK_NOT_NULL(client_);`.

I highly suspect it's the result of [this Node.js PR](https://github.com/nodejs/node/pull/21182)

ak239:
I believe that I can reproduce similar crash by following steps:
1. clone [`electron-quick-start`](https://github.com/electron/electron-quick-start)
2. enable node integration in Node.js by adding `nodeIntegration: true` in `webPreferences`.
3. add `require('inspector').open(9229, undefined, false)` to `renderer.js`

cc @eugeneo , who is an author of https://github.com/nodejs/node/pull/21182

eugeneo:
[...]

PR #24109, opened 2020-06-12
fix: enable Node.js repl in the renderer

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/18872.

Fixes an issue where `repl.start` caused a crash in the renderer process. This is most simply a result of `node::inspector::StartIoThread` being called without prior call to` node::inspector::Start`. However, we can't just add something like:

```
auto node_debugger = std::make_unique<NodeDebugger>(env);
node_debugger->Start();
```

to `electron_renderer_client.cc` since `V8Inspector `was designed with an idea that there is only one `V8Inspector` per isolate. Since the renderer process already has one via Blink, this meant that if we tried to start the  Node.js debugger in the renderer it would crash. We can work around this by adding a new contructor to `NodeInspectorClient` that allows it to be initialized with an existing `V8InspectorClient`, and pass it the one Blink created previously.

I don't know if this is necessarily the best approach, but it was the one that required the fewest alterations within pre-existing constraints - leaving it as a draft to solicit feedback.

Tested with [this repo](https://github.com/junga-com/elelectron-inspectorInRenderCrashes/tree/bg-rendererREPL-crashesOnTab).

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed a crash when calling `repl.Start()` in the renderer process.


codebytere:
I might come back to this someday but for now it's best to close as I don't have cycles.

PR #24114, opened 2020-06-12
feat: add worldSafe flag for executeJS results

MarshallOfSound:
Previously the return values of `webFrame.executeJavaScript` crossed the world boundary when context isolation was enabled.  This allows apps to makes themselves insecure by accidentally sending objects from the isolated world back to the main world.  To help devs avoid this we're adding this new flag, and this flag will be turned on by default in Electron 12 (and removed) ensuring that this kind of issue can't become a thing again.

This PR is also requesting new minors of 8 and 9 😄 

Notes: Added new `worldSafeExecuteJavaScript` webPreference to ensure that the return values from `webFrame.executeJavaScript` are world safe when context isolation is enabled


shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -130,10 +134,34 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
+          bool success;
+          {
+            v8::TryCatch try_catch(isolate);
+            success = gin::ConvertFromV8(isolate, result[0], &ret);
+          }
+          if (!success) {
+            // Failed convert so we send undefined everywhere
+            if (!callback_.is_null())
+              std::move(callback_).Run(v8::Undefined(isolate),
+                                       v8::Undefined(isolate));
```
nornagon, line 154: should we throw here? or at least print a warning?

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -130,10 +134,34 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
+            promise_.Resolve(v8::Undefined(isolate));
```
nornagon, line 155: maybe reject the promise?

lib/renderer/api/web-frame.ts
```diff
@@ -47,12 +48,18 @@ class WebFrame extends EventEmitter {
...
+const { hasSwitch } = process.electronBinding('command_line');
+const worldSafeJS = hasSwitch('world-safe-execute-javascript') && hasSwitch('context-isolation');
+
 // Populate the methods.
 for (const name in binding) {
   if (!name.startsWith('_')) { // some methods are manually populated above
     // TODO(felixrieseberg): Once we can type web_frame natives, we could
     // use a neat `keyof` here
     (WebFrame as any).prototype[name] = function (...args: Array<any>) {
+      if (!worldSafeJS && name.startsWith('executeJavaScript')) {
```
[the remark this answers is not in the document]
MarshallOfSound: I'll update the impl to only do the "world safe" thing if the creation context isn't the calling context

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -130,10 +134,54 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
-        // Right now only single results per frame is supported.
-        if (!callback_.is_null())
-          std::move(callback_).Run(result[0], v8::Undefined(isolate));
-        promise_.Resolve(result[0]);
+        // Either world safe results are disabled or the result was created in
+        // the same world as the caller
+        if (!world_safe_result_ ||
+            (result[0]->IsObject() &&
+             promise_.GetContext() ==
+                 result[0].As<v8::Object>()->CreationContext())) {
```
nornagon, line 142: if `result[0]` is a `Primitive` can we return it directly?

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -130,10 +134,54 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
+            promise_.RejectWithErrorMessage(error_message);
+          } else {
+            v8::Local<v8::Context> context = promise_.GetContext();
+            v8::Context::Scope context_scope(context);
+            v8::Local<v8::Value> cloned_value = gin::ConvertToV8(isolate, ret);
+            if (!callback_.is_null())
+              std::move(callback_).Run(cloned_value, v8::Undefined(isolate));
+            promise_.Resolve(cloned_value);
+          }
+        }
```
nornagon, line 184: getting highly indented here. Can we break some of this functionality out into a helper function?

shell/renderer/api/electron_api_web_frame.cc
MarshallOfSound: No. Primitives still have prototypes right?

nornagon: I don't think so? `GetPrototype` is defined on `v8::Object`.

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -120,20 +121,73 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
+            !gin::ConvertFromV8(isolate, message->Get(), &error_message)) {
+          error_message =
+              "An unknown exception occurred while getting the result of "
+              "the script";
+        }
+      }
+    }
+    if (!success) {
+      // Failed convert so we send undefined everywhere
+      if (!callback_.is_null())
```
nornagon, line 154:
```suggestion
      if (callback_)
```
https://source.chromium.org/chromium/chromium/src/+/master:base/callback_internal.h;l=124;drc=a6a4ba2cbd8841c1d62705ed8a24d44042466153

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -120,20 +121,73 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
+            v8::Undefined(isolate),
+            v8::Exception::Error(
+                v8::String::NewFromUtf8(isolate, error_message.c_str())
+                    .ToLocalChecked()));
+      promise_.RejectWithErrorMessage(error_message);
+    } else {
+      v8::Local<v8::Context> context = promise_.GetContext();
+      v8::Context::Scope context_scope(context);
+      v8::Local<v8::Value> cloned_value = gin::ConvertToV8(isolate, ret);
+      if (!callback_.is_null())
```
nornagon, line 165:
```suggestion
      if (callback_)
```

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -120,20 +121,73 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
-          std::move(callback_).Run(result[0], v8::Undefined(isolate));
-        promise_.Resolve(result[0]);
+        // Either world safe results are disabled or the result was created in
+        // the same world as the caller or the result is not an object and
+        // therefore does not have a prototype chain to protect
+        if (!world_safe_result_ ||
+            (result[0]->IsObject() &&
+             promise_.GetContext() ==
+                 result[0].As<v8::Object>()->CreationContext()) ||
+            !result[0]->IsObject()) {
```
nornagon, line 183:
```suggestion
        v8::Local<v8::Value> value = result[0];
        bool should_clone = world_safe_result_ &&
            value->IsObject() &&
            promise_.GetContext() != value.As<v8::Object>()->CreationContext();
        if (!should_clone) {
```

probs also switch `if (!x) a else b` -> `if (x) b else a`

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -120,20 +121,73 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
+          // Right now only single results per frame is supported.
+          if (!callback_.is_null())
+            std::move(callback_).Run(result[0], v8::Undefined(isolate));
+          promise_.Resolve(result[0]);
```
nornagon, line 187:
```suggestion
          if (callback_)
            std::move(callback_).Run(value, v8::Undefined(isolate));
          promise_.Resolve(value);
```

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -120,20 +121,76 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
       if (!result[0].IsEmpty()) {
-        // Right now only single results per frame is supported.
-        if (!callback_.is_null())
-          std::move(callback_).Run(result[0], v8::Undefined(isolate));
-        promise_.Resolve(result[0]);
+        v8::Local<v8::Value> value = reuslt[0];
+        // Either world safe results are disabled or the result was created in
+        // the same world as the caller or the result is not an object and
+        // therefore does not have a prototype chain to protect
+        bool should_send_directly =
```
nornagon, line 180: i specifically inverted the sense of this boolean in my comment because i thought it was significantly more readable that way.

lib/renderer/api/web-frame.ts
```diff
@@ -47,14 +48,26 @@ class WebFrame extends EventEmitter {
...
+        deprecate.log(`Security Warning: webFrame.${name} was called without worldSafeExecuteJavaScript enabled. This is considered unsafe. worldSafeExecuteJavaScript will be enabled by default in Electron 12.`);
```
nornagon, line 61: Should we make this a one-time warning rather than logging every time?

lib/renderer/web-frame-init.ts
```diff
@@ -12,6 +12,11 @@ export const webFrameInit = () => {
   ipcRendererUtils.handle('ELECTRON_INTERNAL_RENDERER_WEB_FRAME_METHOD', (
     event, method: keyof WebFrameMethod, ...args: any[]
   ) => {
+    // TODO(MarshallOfSound): Remove once the world-safe-execute-javascript deprecation warning is removed
+    if (method.startsWith('executeJavaScript')) {
+      return (webFrame as any)[`_${method}`](...args);
+    }
```
nornagon, line 18: does this... bypass the deprecation warning?

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -120,20 +121,76 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
+              "the script";
+        }
+      }
+    }
+    if (!success) {
+      // Failed convert so we send undefined everywhere
+      if (callback_)
+        std::move(callback_).Run(
+            v8::Undefined(isolate),
+            v8::Exception::Error(
```
nornagon, line 157: What context is this error created in?

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -120,20 +121,76 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
...
+        v8::Local<v8::Value> value = reuslt[0];
+        // Either world safe results are disabled or the result was created in
+        // the same world as the caller or the result is not an object and
+        // therefore does not have a prototype chain to protect
+        bool should_clone_value =
+            world_safe_result_ &&
+            !(value->IsObject() &&
+              promise_.GetContext() ==
+                  value.As<v8::Object>()->CreationContext()) &&
+            value->IsObject() if (should_clone_value) {
```
nornagon, line 185:
```suggestion
            value->IsObject(); if (should_clone_value) {
```

lib/renderer/api/web-frame.ts
MarshallOfSound: I think logging every time is ok, historically I think logging once means the first call at the start of the app triggers the warning and then no one ever sees it again.

lib/renderer/web-frame-init.ts
MarshallOfSound: Yes, because these results are immediately sent over IPC and not touched in any way it didn't make sense to me to log a deprecation warning for `webContents.executeJavaScript` as no change in behavior would be noticed when the flag is toggled

shell/renderer/api/electron_api_web_frame.cc
MarshallOfSound: Pretty sure it ends up in the caller context scope but I've made this explicit.

docs/api/browser-window.md
```diff
@@ -348,6 +348,9 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       You can access this context in the dev tools by selecting the
       'Electron Isolated Context' entry in the combo box at the top of the
       Console tab.
+    * `worldSafeExecuteJavaScript` Boolean (optional) - If true, values returned from `webFrame.executeJavaScript` will be sanitized to ensure JS values
+      can't unsafely cross between worlds when using `contextIsolation`.  The default
+      is `false`. In Electron 12, the default will be changed to `true`. _Deprecated_
```
nornagon, line 353: I wonder if this might be a good test case for the `base::Feature`-based stuff we've been talking about in api-wg?

MarshallOfSound: Maybe, but I don't super want to figure out that system / pioneer it in this PR

nornagon: Someone's gotta be the first :\

bughit:
@MarshallOfSound 

Please don't remove the option. I have code the defines (and returns) anon functions in the main world. That's not even hypothetically unsafe. Also code that returns other main world objects (e.g. window). The 'window' property is unwritable and unconfigurable, so that's also not unsafe, you're always get the expected window even in untrusted content.  It's reasonable for this raw access be restricted by default, but the EIC should retain it optionally, it can be safe, especially if the content is trusted.

If contextIsolation can be disabled and nodeIntegration enabled, security is clearly ultimately up to the dev, it can't be forced like this.

bughit:
@nornagon 

Could we please have some discussion about the planned removal of worldSafeExecuteJavaScript. I have code that depends on it. Retrieves the main world window. Defines anon functions for later execution. 

```
const contentWin = webFrame.executeJavaScript('window')
const contentFn = webFrame.executeJavaScript('() => {...}')
```

I have no issue with maximally secure defaults, but there shouldn't be a need to completely block this, breaking working code, forcing redesigns. . This can be used safely, judiciously. Considering security can be disabled completely, it's not unreasonable to optionally allow EIC to retrieve content world objects. Please don't remove the choice.

docs/api/browser-window.md
poiru, line 353: @MarshallOfSound Why is this marked `Deprecated`? I turned this on to prepare for Electron 12, but my code reviewer was confused why I'm using a deprecated option. Presumably `false` is deprecated, but I think we should be explicit about it.

MarshallOfSound: @poiru This option is a migration feature, it's kinda weird and you're right we should probably document it better.  The flow here will be

* Add this option
* Change default of this option
* Deprecate this option in code
* Remove this option

We aren't going to leave this flag around forever / for long at all

PR #24017, opened 2020-06-08
refactor: replace v8Util.createIDWeakMap() with native WeakRef in tests

miniak:
#### Checklist

- [ ] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [ ] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [ ] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).
- [ ] This is **NOT A BREAKING CHANGE**. Breaking changes may not be merged to master until 11-x-y is branched.

#### Release Notes

Notes:

PR #24115, opened 2020-06-12
refactor: use WeakRef on main process side of remote

nornagon:
#### Description of Change
Followup to #24037, converting main process side.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


lib/browser/remote/server.ts
```diff
@@ -14,16 +14,48 @@ if (!features.isRemoteModuleEnabled()) {
...
+  const mapKey = id[0] + '~' + id[1];
+  const ref = rendererFunctionCache.get(mapKey);
+  if (ref !== undefined) {
+    const deref = ref.deref();
+    if (deref !== undefined) return deref;
+  }
+}
+function setCachedRendererFunction (id: RendererFunctionId, wc: electron.WebContents, frameId: number, value: CallIntoRenderer) {
+  // eslint-disable-next-line no-undef
+  const wr = new (globalThis as any).WeakRef(value) as WeakRef<CallIntoRenderer>;
```
MarshallOfSound, line 49: Is there any way we can get types for these things, update TS maybe?

nornagon: The approach here https://github.com/electron/electron/pull/24017 is to copy/paste them in. I don't think the types have landed yet.

Issue #23838, opened 2020-05-29
Incorrect UnhandledPromiseRejectionWarning for asynchronous called async function


ivanggq:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v9.0.0
* **Operating System:**
  * Windows 10 (1909)
* **Last Known Working Electron version:**
  * Not sure

### Expected Behavior
UnhandledPromiseRejectionWarning warnings should not be printed since all promise rejections are handled.
Similar script executed in plain Node.js project (not Electron) with recent Node version (v12.16.1) does not produce these warnings.

### Actual Behavior
UnhandledPromiseRejectionWarning warnings are printed even though all promise rejections are handled.

### To Reproduce
```sh
git clone https://github.com/ivanggq/electron-async-vs-promise-issue.git -b master
cd electron-async-vs-promise-issue/
npm install
npm start
```

### Additional Information
The focus of this bug is that the UnhandledPromiseRejectionWarning is printed only under the following conditions:
 - only for function f3(),
 - only when executed under Electron (it is not printed under plain Node project), and, 
 - only when the function is called asynchronously (from a setTimeout() callback)

If any of the above conditions is changed, the warning is not printed. It is not printed
 - for functions f1() and f2(), or,
 - when executed from plain Node.js project (Electron-specific parts should be removed from the script), or, 
 - when the functions are called directly from the main script (not from setTimeout() callback)

I am looking to get answer/discussion to the  following questions:
 - Why the UnhandledPromiseRejectionWarning warning is printed when executed from Electron but not when executed in a plain Node.js project?
 - Technical explanation of under-the-hood difference between f1/f2 and f3? Is an additional promise created/destroyed leading to some internal detection logic to complain? Why is the warning printed twice?
 - Why this happens only when executed from an asynchronous callback?
 - Is there any incorrect behavior in Electron or Node.js that should be fixed? If the outcome is correct, this means the code I submit is wrong. Can you point me to documentation or technical specification explaining why f3 is wrong or discouraged, if so?

```
const { app } = require('electron');

const f1 = async () => {
    throw 'oops';
}

[...]

Issue #20013, opened 2019-08-28
DCHECK when closing BrowserWindow from microtask


nornagon:
Release builds don't crash on this because it's a DCHECK.

Fiddle: https://gist.github.com/316a0d5e34603dc3a352e8289f0065de

```js
const {app, BrowserWindow} = require('electron')

async function ensureWindowIsClosed(window) {
  new Promise(resolve => {
    window.once('closed', () => resolve())
  }).then(() => {}) // <-- this then() is critical! The crash doesn't happen without it.
  window.destroy()
}
const closeWindow = async (window) => {
  await ensureWindowIsClosed(window)
}

async function createWindow () {
  const w = new BrowserWindow({show: false})
  try {
    await w.webContents.session.cookies.get({ url: 'http://127.0.0.1' })
  } finally {
    await closeWindow(w)
  }
}

app.on('ready', createWindow)
```

Stack trace:
```
#
# Fatal error in ./../../v8/src/execution/microtask-queue.cc, line 126
# Debug check failed: !*flag_.
#
#
#
#FailureMessage Object: 0x7ffee8dc3970Received signal 4 <unknown> 000112059212
0   Electron Framework                  0x000000010f56bda9 base::debug::CollectStackTrace(void**, unsigned long) + 9
1   Electron Framework                  0x000000010f416b53 base::debug::StackTrace::StackTrace() + 19
2   Electron Framework                  0x000000010f56bc41 base::debug::(anonymous namespace)::StackDumpSignalHandler(int, __siginfo*, void*) + 2385
3   libsystem_platform.dylib            0x00007fff66ab1b5d _sigtramp + 29
4   ???                                 0x0000ffff00001fab 0x0 + 281470681751467
5   Electron Framework                  0x0000000112052773 V8_Fatal(char const*, int, char const*, ...) + 339
6   Electron Framework                  0x00000001120522c5 v8::base::(anonymous namespace)::DefaultDcheckHandler(char const*, int, char const*) + 21
7   Electron Framework                  0x000000010ca16341 v8::internal::MicrotaskQueue::RunMicrotasks(v8::internal::Isolate*) + 1345
8   Electron Framework                  0x0000000114b61b54 node::InternalCallbackScope::Close() + 260
9   Electron Framework                  0x0000000114b61edd node::InternalMakeCallback(node::Environment*, v8::Local<v8::Object>, v8::Local<v8::Function>, int, v8::Local<v8::Value>*, node::async_context) + 525
[...]

PR #24131, opened 2020-06-15
fix: let Node.js perform microtask checkpoint in the main process

deepak1556:
#### Description of Change

Closes https://github.com/electron/electron/issues/23838.
Refs https://github.com/electron/electron/issues/20013#issuecomment-643550663

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: fix delayed execution of some Node.js callbacks in the main process

### DIFF
--- a/filenames.gni
+++ b/filenames.gni
@@ -525,6 +525,8 @@ filenames = {
     "shell/common/gin_helper/function_template_extensions.h",
     "shell/common/gin_helper/locker.cc",
     "shell/common/gin_helper/locker.h",
+    "shell/common/gin_helper/microtasks_scope.cc",
+    "shell/common/gin_helper/microtasks_scope.h",
     "shell/common/gin_helper/object_template_builder.cc",
     "shell/common/gin_helper/object_template_builder.h",
     "shell/common/gin_helper/persistent_dictionary.cc",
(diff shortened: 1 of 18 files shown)


codebytere:
@deepak1556 i think this fixes https://github.com/electron/electron/issues/23838

```js
const f3 = async () => {
    return new Promise((_, reject) => {
        reject('oops');
    });
}
 f3().catch((error) => {
        console.log('3 caught', error);
});
```

should function as a minimal test case if run in the main proc

Right now it errors like so:
```
(node:86234) UnhandledPromiseRejectionWarning: oops
(node:86234) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:86234) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
3 caught oops
(node:86234) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 1)
```

See https://github.com/electron/electron/issues/23838#issuecomment-639886588 :)

deepak1556:
Thanks @codebytere , yep that does serve as a valid test case.

zcbenz:
On the failed getGPUInfo test, it is because with this change the resolve/reject callbacks get called earlier and then results in a call like this:

```js
w.webContents.once('did-finish-load', () => {
  app.exit(1)
})
```

Which then crashes.

Schedule the `app.exit` into next tick should be able to fix the failed test, though we should look into solving the crash in future.

Issue #24052, opened 2020-06-10
Media Keys not working when registered with globalShortcut


somebeaver:
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

Note: I am aware of these issues, but they have been closed and are for 8.1.x and 8.2.x releases, not 8.3.x+.
https://github.com/electron/electron/issues/20788
https://github.com/electron/electron/issues/22405

### Issue Details

* **Electron Version:**
  * 8.3.1 (and 9.0.3)
* **Operating System:**
  * macOS 10.15.5
* **Last Known Working Electron version:**
  * 7.3.1

### Expected Behavior
The expected behavior can be seen in Electron 7.3.1, where the media key event handler is invoked when the media key is pressed.

### Actual Behavior
Electron 8.3.1 says that it registers the Media Key event handlers correctly, but in fact it does not. Pressing "Play" on my keyboard will open Apple Music because the Electron globalShortcut is not working.

### To Reproduce
**Fiddle: https://gist.github.com/e502767f6dc26942c825a73029f0b20a**

Steps:

1. Open Fiddle with my gist and Electron 7.3.1.
2. Launch the Electron app, and grant the accessibility permission in the macOS system settings if needed (there will be a popup if you need to).
3. On your keyboard, press the Next/Prev/PlayPause media keys, and see that in the main process console, the events are being captured and that this code works.
4. Repeat the steps with Electron 8.3.1; the events are not being captured even though Electron returns `true` for four `globalShortcut.register()` invocations.

PR #24145, opened 2020-06-16
fix: global shortcut media keys working with accessibility

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/24052.

Initially, I tried to fix this in https://github.com/electron/electron/pull/21442, by patching out Chromium's usage of `RemoteCommandCenterDelegate` entirely, as it uses `MPRemoteCommandCenter` in a change added in [this CL](https://chromium-review.googlesource.com/c/chromium/src/+/1480702). `MPRemoteCommandCenter` makes it such that `globalShortcuts` in Electron will not work as intended, because by design an app does not receive remote control events until it begins playing audio. However, as of macOS Catalina, the alternate implementation requires accessibility access to have been granted, which means that it created a situation where simply trying to watch a YouTube video and hitting play/pause would trigger Electron to request a user for Accessibility permissions. For many large apps this is extremely undesirable behavior. 

This second-pass solution splits the difference. What we want is for unilateral override _only_ when registering `globalShortcuts`, which means disabling internal media key handling specifically in that use case. Chromium's media key listener manager already kept track of `media_key_handling_enabled` as a member, so I added a new static method to `GlobalShortcutListenerMac` that allows us to enable and disable it when registering and unregistering shortcuts. Then, we can ensure it's off when registering global shortcuts, and on otherwise (preserving previous behavior).

Tested with https://gist.github.com/codebytere/a7e33c6f03a84dbcc2878859237422dd.

cc @MarshallOfSound @ckerr @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed media keys working unilaterally when registered with `globalShortcut`

### DIFF
--- a/patches/chromium/fix_use_the_new_mediaplaypause_key_listener_for_internal_chrome.patch
+++ b/patches/chromium/fix_use_the_new_mediaplaypause_key_listener_for_internal_chrome.patch
@@ -8,6 +8,54 @@ The new kGlobalRequiresAccessibility Scope type should be upstreamed
 and then we can use that and minimize this patch to just the change in
 global_shortcut_listener_mac.mm
 
+diff --git a/chrome/browser/extensions/global_shortcut_listener.cc b/chrome/browser/extensions/global_shortcut_listener.cc
+index bc009606d01469125052e68a9cdc82aaa697c764..ff18043cb07d748a49adea9874517fb29e3e7f9f 100644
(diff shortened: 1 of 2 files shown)


somebeaver:
@MarshallOfSound Hello, I do not mean to pester, I am just looking for an update on if this will be merged any time soon, as it has been manually blocked for about a month. Thank you :)

Issue #24144, opened 2020-06-16
TypeScript: MenuItem click handler is not typed


bpasero:
Referring to https://www.electronjs.org/docs/api/menu-item#new-menuitemoptions

The `click` handler is typed as `Function` inside the `electron.d.ts`. And the documentation is wrong around the `browserWindow` property being `BrowserWindow`. It can be `undefined` on macOS when no window is opened.

PR #24147, opened 2020-06-16
docs: fix MenuItem click handler type

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/24144.

Corrects potentially undefined `browserWindow` in click handler for MenuItems.

#### Checklist

#### Release Notes

Notes: none.

### DIFF
--- a/docs/api/menu-item.md
+++ b/docs/api/menu-item.md
@@ -12,7 +12,7 @@ See [`Menu`](menu.md) for examples.
   * `click` Function (optional) - Will be called with
     `click(menuItem, browserWindow, event)` when the menu item is clicked.
     * `menuItem` MenuItem
-    * `browserWindow` [BrowserWindow](browser-window.md)
+    * `browserWindow` [BrowserWindow](browser-window.md) | undefined - This will not be defined if no window is open.
     * `event` [KeyboardEvent](structures/keyboard-event.md)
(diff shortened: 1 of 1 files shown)


docs/api/menu-item.md
```diff
@@ -12,7 +12,7 @@ See [`Menu`](menu.md) for examples.
   * `click` Function (optional) - Will be called with
     `click(menuItem, browserWindow, event)` when the menu item is clicked.
     * `menuItem` MenuItem
-    * `browserWindow` [BrowserWindow](browser-window.md)
+    * `browserWindow` [BrowserWindow](browser-window.md) | undefined - This will not be defined if no window is open.
```
MarshallOfSound, line 15: Should use `(optional)` to define an optional type

codebytere: @MarshallOfSound docs-parser currently precludes using `optional` for a variable followed by a non-optional variable, and switching them would be a breaking change which is why i did it this way

MarshallOfSound: TS might preclude it but our docs parser doesn't and is smart enough to generate this correctly if you use (optional)

codebytere: resolved - this is not in fact possible 😛

PR #24156, opened 2020-06-16
Correct for the deprecated API

railsjack:
We need to mention about the correct URL for the deprecated API.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:

### DIFF
--- a/docs/tutorial/devtools-extension.md
+++ b/docs/tutorial/devtools-extension.md
@@ -45,6 +45,9 @@ The extension will be remembered so you only need to call this API once per
 extension. If you try to add an extension that has already been loaded, this method
 will not return and instead log a warning to the console.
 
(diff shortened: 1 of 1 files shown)


docs/tutorial/devtools-extension.md
```diff
@@ -45,6 +45,9 @@ The extension will be remembered so you only need to call this API once per
 extension. If you try to add an extension that has already been loaded, this method
 will not return and instead log a warning to the console.
 
+**Note:** The `BrowserWindow.addDevToolsExtension` API is deprecated. Here is an alternative.
```
codebytere, line 48: Instead of adding a note, it's probably more effective to just update the tutorial to use the correct new methods so let's do that instead.

PR #23831, opened 2020-05-28
build: remove dead symlink from MAS build

MarshallOfSound:
Fixes #23828

This fixes the stray symbolic link and adds a script to ensure it doesn't happen again

Notes: Fixed mac app store rejection notice for invalid symbolic link in bundle.

PR #24158, opened 2020-06-16
build: remove dead symlink from MAS build

### LINKED ISSUE #23828: Electron 9.0.0: Mac App Store rejection - ITMS-90332: Invalid Symbolic Link
### Preflight Checklist

### Issue Details

* **Electron Version:** 9.0.0
* **Operating System:** MacOS
* **Last Known Working Electron version:** Electron 8

Release to Mac App Store failed with below error:

ITMS-90332: Invalid Symbolic Link - The symbolic link 'com.my.app.pkg/Payload/MyApp.app/Contents/Frameworks/Electron Framework.framework/Helpers' resolves to an invalid location. Make sure that the symbolic link does not resolve to itself, and that the location exists and is contained within the app bundle.

I've checked `electron-v9.0.0-mas-x64.zip` file from `~/Library/Caches/electron`, open the package content of `Electron.app` and I found it contains an invalid alias at `Electron.app/Contents/Frameworks/Electron Framework.framework/Helpers`. I am not sure if this file is added on purpose - it is not in electron 8 release.

### Expected Behavior

App should be able to release to App Store

### Actual Behavior
Apple rejected the build

MarshallOfSound:
Fixes #23828

This fixes the stray symbolic link and adds a script to ensure it doesn't happen again

Notes: Fixed mac app store rejection notice for invalid symbolic link in bundle.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -831,7 +831,7 @@ step-restore-out-cache: &step-restore-out-cache
     paths:
       - ./src/out/Default
     keys:
-      - v7-out-cache-{{ checksum "src/electron/.depshash" }}-{{ checksum "src/electron/.depshash-target" }}
+      - v8-out-cache-{{ checksum "src/electron/.depshash" }}-{{ checksum "src/electron/.depshash-target" }}
     name: Restoring out cache
 
 step-set-git-cache-path: &step-set-git-cache-path
@@ -855,7 +855,7 @@ step-save-out-cache: &step-save-out-cache
   save_cache:
     paths:
       - ./src/out/Default
-    key: v7-out-cache-{{ checksum "src/electron/.depshash" }}-{{ checksum "src/electron/.depshash-target" }}
+    key: v8-out-cache-{{ checksum "src/electron/.depshash" }}-{{ checksum "src/electron/.depshash-target" }}
     name: Persisting out cache
 
 step-run-electron-only-hooks: &step-run-electron-only-hooks
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -418,6 +418,7 @@ step-electron-build: &step-electron-build
       fi
       cd src
       ninja -C out/Default electron -j $NUMBER_OF_NINJA_PROCESSES
+      node electron/script/check-symlinks.js
 
 step-native-unittests-build: &step-native-unittests-build
   run:
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -326,7 +326,6 @@ source_set("electron_lib") {
     "//chrome/app/resources:platform_locale_settings",
     "//chrome/services/printing/public/mojom",
     "//components/certificate_transparency",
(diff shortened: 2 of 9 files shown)


nornagon:
This is a reland of #23831?

MarshallOfSound:
@nornagon Yes but with a more correct fix for the gn check failures, the MAS build now truly doesn't depend on any parts of the crash component

deepak1556:
Some crash symbols still leak through because of this path

```
//electron:electron_lib --[private]-->
//electron/chromium_src:chrome --[public]-->
//chrome/common:common --[private]-->
//components/crash/core/app:app
```

if symbols should be completely removed in MAS build then this path should be addressed.

nornagon:
I don't _think_ symbols are the issue here, I think it's API calls?

deepak1556:
I am not sure too, @MarshallOfSound can you please clarify the end goal of conditionally including those deps. If symbols are not an issue, I would like to add `//components/crash/core/app` back and remove the `nogncheck`

MarshallOfSound:
@deepak1556 It's a combination.  The original goal was to _completely remove_ the crash component as a dependency to guarantee that MAS build can't call anything in there that straight up won't work.  If there's a dependency chain I missed I can follow up to remove that.

The actual _problem_ this PR solves is twofold.

(1) The `Helpers` folder needs to not exist on MAS builds, this requires changing some GN config
(2) The MAS build needs to not end up binding to, or calling, any crash methods (so this introduces some new noop / stub methods)

I think the dependency chain you posted can be removed by [patching out the dependency for mas builds](https://source.chromium.org/chromium/chromium/src/+/master:chrome/common/BUILD.gn;l=261?q=%22%2F%2Fcomponents%2Fcrash%2Fcore%2Fapp%22&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2F)

deepak1556:
Thanks for the context!

Removing the dependency chain from `//chrome` is a patch that we have to maintain always. With the electron crash api sources, `Helpers` framework and `crashpad_handler_syms` excluded, can we just build the deps unconditionally ?

MarshallOfSound:
> With the electron crash api sources, Helpers framework and crashpad_handler_syms excluded, can we just build the deps unconditionally ?

The Helpers and syms are completely excluded but the `electron_api_crash_reporter` files and a few others are left in and we rely on `MAS_BUILD` `defined` checks to ensure we aren't doing The Wrong Thing in a mas build.  It would be way safer to ensure it's impossible link those symbols into the MAS build accidentally.  I don't know how big of a patch that would end up looking like though.

deepak1556:
Hmm, this is a unique case where it is not completely necessary to remove the symbols but at the same time want to ensure headers are behind `#if defined`. I am fine with ignoring the symbols coming from `//chrome`, I don't think a patch is worth it.

davidwinter:
Will this be backported to v9?

PR #24179, opened 2020-06-17
refactor: remove uses of v8::Isolate::GetCurrent

codebytere:
#### Description of Change

This PR moves away from discouraged usage of `v8::Isolate::GetCurrent`. 

Several motivations here:
 - `v8::Isolate::GetCurrent()` is slated for deprecation (see [this doc](https://docs.google.com/document/d/1g8JFi8T_oAE_7uAri7Njtig7fKaPDfotU6huOa1alds/edit) and [this conversation](https://groups.google.com/forum/#!topic/v8-users/bhfAVJySjIA)) - the conversation is a bit old but it remains the case that v8 is trying to get to a state where the embedder always specifies the isolate to use and doesn't use TLS to store it internally.
- `v8::Isolate::GetCurrent()` uses costly atomic loads
- A static `v8::Isolate` accessor was added in https://github.com/electron/electron/commit/ac5c30a70775838994d27946c3bb62688617421a, and it's confusing to have multiple ways to accomplish the same goal with no clear usage demarcation.

cc @deepak1556 @nornagon @zcbenz 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/ui/certificate_trust_win.cc
+++ b/shell/browser/ui/certificate_trust_win.cc
@@ -10,6 +10,7 @@
 
 #include "net/cert/cert_database.h"
 #include "net/cert/x509_util_win.h"
+#include "shell/browser/javascript_environment.h"
 
 namespace certificate_trust {
 
@@ -60,7 +61,7 @@ v8::Local<v8::Promise> ShowCertificateTrust(
     electron::NativeWindow* parent_window,
     const scoped_refptr<net::X509Certificate>& cert,
     const std::string& message) {
-  v8::Isolate* isolate = v8::Isolate::GetCurrent();
+  v8::Isolate* isolate = electron::JavascriptEnvironment::GetIsolate();
   gin_helper::Promise<void> promise(isolate);
   v8::Local<v8::Promise> handle = promise.GetHandle();
 
--- a/shell/browser/api/electron_api_app.cc
+++ b/shell/browser/api/electron_api_app.cc
@@ -39,6 +39,7 @@
 #include "shell/browser/api/gpuinfo_manager.h"
 #include "shell/browser/electron_browser_context.h"
 #include "shell/browser/electron_browser_main_parts.h"
+#include "shell/browser/javascript_environment.h"
 #include "shell/browser/login_handler.h"
 #include "shell/browser/relauncher.h"
 #include "shell/common/application_info.h"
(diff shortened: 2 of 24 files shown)


shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -127,7 +128,7 @@ class ScriptExecutionCallback : public blink::WebScriptExecutionCallback {
 
   void Completed(
       const blink::WebVector<v8::Local<v8::Value>>& result) override {
-    v8::Isolate* isolate = v8::Isolate::GetCurrent();
+    v8::Isolate* isolate = JavascriptEnvironment::GetIsolate();
```
nornagon, line 131: `JavascriptEnvironment` is not initialized in the renderer

codebytere: whoops, overdid it on the batch replace

shell/common/gin_helper/event_emitter_caller.cc
```diff
@@ -4,6 +4,7 @@
 
 #include "shell/common/gin_helper/event_emitter_caller.h"
 
+#include "shell/browser/javascript_environment.h"
```
MarshallOfSound, line 7: Is this `include` needed?

shell/common/gin_helper/event_emitter_caller.h
```diff
@@ -82,7 +83,7 @@ template <typename T, typename... Args>
 v8::Local<v8::Value> CallMethod(gin::Wrappable<T>* object,
                                 const char* method_name,
                                 Args&&... args) {
-  v8::Isolate* isolate = v8::Isolate::GetCurrent();
+  v8::Isolate* isolate = electron::JavascriptEnvironment::GetIsolate();
```
MarshallOfSound, line 86: A file in `shell/common` shouldn't depend on `shell/browser`, this won't work in a renderer context

shell/common/gin_helper/event_emitter_caller.cc
codebytere: nope - i guess for a few of these we have no choice at present so i'll update them

shell/common/gin_helper/event_emitter_caller.h
codebytere: you're right - only `shell/browser` file changes left now!

shell/browser/ui/certificate_trust_win.cc
```diff
@@ -60,7 +60,7 @@ v8::Local<v8::Promise> ShowCertificateTrust(
     electron::NativeWindow* parent_window,
     const scoped_refptr<net::X509Certificate>& cert,
     const std::string& message) {
-  v8::Isolate* isolate = v8::Isolate::GetCurrent();
+  v8::Isolate* isolate = JavascriptEnvironment::GetIsolate();
```
deepak1556, line 63: `electron::JavascriptEnvironment`

Issue #24189, opened 2020-06-18
Error: No such module was linked: atom_browser_v8_util


lutzroeder:
### Preflight Checklist
* [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [X] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**: 9.0.4
* **Operating System:** macOS 10.15.5 (19F101)

### Expected Behavior
Debugging with `All Exceptions` and `Uncaught Exceptions` enabled should not break in the debugger during load.

### Actual Behavior
```
Error: No such module was linked: atom_browser_v8_util
    at _linkedBinding (internal/bootstrap/loaders.js:118:34)
    at process.electronBinding (electron/js2c/browser_init.js:7570:20)
    at Object../lib/browser/api/menu.js (electron/js2c/browser_init.js:1788:24)
    at __webpack_require__ (electron/js2c/browser_init.js:20:30)
    at loader (electron/js2c/browser_init.js:2087:35)
    at Object.Menu (electron/js2c/browser_init.js:7536:19)
    at App../lib/browser/default-menu.ts.exports.setDefaultApplicationMenu (electron/js2c/browser_init.js:4914:29)
    at Object.onceWrapper (events.js:313:26)
    at App.emit (events.js:228:7)
```

### To Reproduce
1. Install Visual Studio Code
2. Install Debugger for Chrome extension
3. `git clone https://github.com/lutzroeder/netron.git`
4. `npm install`
5. Open in Visual Studio Code
6. Debugger > `Launch Application`
7. Enable `All Exceptions` and `Uncaught Exceptions` under `Breakpoints`
8. <kbd>F5</kbd>

### Screenshots

<img width="1057" alt="screenshot" src="https://user-images.githubusercontent.com/438516/84973748-b2f34a00-b0d6-11ea-909b-2c7a0654d658.png">

@codebytere @tarruda

codebytere:
@lutzroeder i'm not sure i understand what the issue is here 🤔 this is an expected error - `v8_util` is a `common` module which is why this is wrapped in a try/catch to fall back onto `electron_common_v8_util` (which is what's actually linked)

lutzroeder:
[...]

PR #24191, opened 2020-06-18
refactor: prevent consistent early exception

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/24189.

This PR alters our `process.electronBinding` parameters to pass the process type alongside the binding name.

cc @MarshallOfSound @deepak1556 @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue with a consistent unwanted exception appearing during a debugging session with the VSCode debugger.


MarshallOfSound:
We should probably just make all call sites pass in the process type.  E.g. process.electronBinding('app', 'browser')

codebytere:
That also works - i'll rework this a bit

deepak1556:
Can't we just refactor to use the full module name at the call site ?

`const { nativeImage, NativeImage } = process._linkedBinding('electron_common_native_image');`

I don't see an advantage with this `electronBindingSetup` helper.

nornagon:
Flagging that some 3rd-party tools might break with this change. It's a private API so it seems OK to break it, but it's been that way for a while and it's used to do some things which aren't otherwise accomplishable (e.g. get/setHiddenValue).

deepak1556:
In that case we can leave the original `process.electronBinding` definition as such for 3rd-party tools and just replace the internal call sites to `process._linkedBinding`

codebytere:
Works for me - it's a bit clearer anyway.

codebytere:
I'm missing one somewhere 🤔 just two tests failing

PR #24195, opened 2020-06-18
fix: do not use CONTEXT_MENU flag for tray menu (reland)

zcbenz:
Backport of #24097

See that PR for details.

Notes: Fix tray menu showing in taskbar on Windows.


codebytere:
@zcbenz `7-3-x` is security only, so we probably don't want to open this here i don't think

zcbenz:
Hmm https://github.com/electron/electron/pull/23897 and https://github.com/electron/electron/pull/24080 are already in the branch, but since one reverts the other one we should be fine then.

PR #24204, opened 2020-06-19
refactor: improve the REPL

codebytere:
#### Description of Change

This PR improves the Electron REPL experience. It adds a welcome message to the REPL to let users know what versions of Node.js and Electron they're running, as well as overriding the completer function in the REPL to preload and add tab autocompletion for Electron's own modules.

Screen Shot 2020-06-18 at 10 27 15 PM

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Improved the default REPL experience when running Electron with the `--interactive` flag.

### DIFF
--- a/default_app/main.ts
+++ b/default_app/main.ts
@@ -149,10 +149,10 @@ function startRepl () {
   // Prevent quitting.
   app.on('window-all-closed', () => {});
 
-  const colors = { GREEN: '32', MAGENTA: '35' };
+  const GREEN = '32';
   const colorize = (color: string, s: string) => `\x1b[${color}m${s}\x1b[0m`;
-  const electronVersion = colorize(colors.MAGENTA, `v${process.versions.electron}`);
-  const nodeVersion = colorize(colors.GREEN, `v${process.versions.node}`);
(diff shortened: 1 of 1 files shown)


default_app/main.ts
```diff
@@ -145,13 +146,57 @@ function startRepl () {
     process.exit(1);
   }
 
-  // prevent quitting
+  // Prevent quitting.
   app.on('window-all-closed', () => {});
 
-  const repl = require('repl');
-  repl.start('> ').on('exit', () => {
+  const colors = { GREEN: '32', MAGENTA: '35' };
```
jkleinsc, line 152: Not sure about using these colors as they are not great for accessibility.

codebytere: hmm good point - do you know what might perhaps be better here? i mostly just want them to stand out 🤔

jkleinsc: It's all about contrast.  Looks like a green against a black background is ok.  Why not use the same color for both?  You can play around with color contrast here: https://webaim.org/resources/contrastchecker/

PR #24241, opened 2020-06-22
feat: Electron Fuses, package time feature toggles

MarshallOfSound:
#### Problem Statement

Certain features or Electron are undesirable for specific deployments / environments and although they can be disabled / not-utilized at the app level certain security models take account for user-level code execution.  E.g. Being able to run electron apps with `ELECTRON_RUN_AS_NODE` and just get a node binary is not-ideal.  We want a way to toggle certain features in a way that both doesn't require app developers to build Electron in-house and ensures that OS level app integrity / code-signing protection (gatekeeper / app-locker) correctly prevent modification of these toggles.

#### Solution: Electron Fuses

Utilizing a known address / section of the app binary we can create a series of "fuses" (toggles) that can be overwritten after Electron has been built into a `.app` / `.exe`.  These can be overwritten by writing raw data to the correct address in the binary.  In order to locate this known address we use a fixed sentinel string that is guaranteed to directly prefix the fuses.

The format of this string is `{sentinel}{fuse_version}{...fuses}`.

This allows packaging tools to find the sentinel, and then know that the next byte is the fuse version and each sequential byte is a fuse.  The `fuse_version` should be used to determine what each fuse is and how many there are.  A handy tool `@electron/fuses` will be built to abstract away most of this complexity for developers.

#### Example use cases

This PR implements one use case (disabling `ELECTRON_RUN_AS_NODE`), other potential use cases which could be implemented in the future using this system.

* Disabling bad flags (`--disable-sandbox`, `--remote-debugging-port`, etc.)
* Disabling the node debugger
* Disable the `app`, `app.asar`, `default_app.asar` load path search and enforce only one of those is ever searched
* Enforce certain webPrefs (E.g. all webContents must have `sandbox`, `contextIsolation`, etc.)
* Other wild stuff that depends on other crazy ideas

#### Implementation Details

Adding a new fuse is as easy as adding a new key to the `fuses.json` file, the fuse implementation is automatically generated as part of the build step so if you add a new `foo_thing_enabled` fuse a new method `electron::fuses::IsFooThingEnabled()` will automatically exist.

#### To Do

* [x] Documentation
* [ ] Build `@electron/fuses` for controlling the fuses at the packaging phase
* [ ] Tests

Notes: Implemented [Electron Fuses](https://example.com) for customizing certain Electron features at package time.

### DIFF
--- a/build/fuses/build.py
+++ b/build/fuses/build.py
@@ -12,6 +12,12 @@ TEMPLATE_H = """
 #ifndef ELECTRON_FUSES_H_
 #define ELECTRON_FUSES_H_
 
+#if defined(WIN32)
+#define FUSE_EXPORT __declspec(dllexport)
+#else
+#define FUSE_EXPORT __attribute__((visibility("default")))
+#endif
+
 namespace electron {
 
 namespace fuses {
@@ -29,8 +35,6 @@ extern const volatile char kFuseWire[];
 
 TEMPLATE_CC = """
 #include "electron/fuses.h"
-
-#include <iostream>
 
 namespace electron {
 
@@ -50,6 +54,7 @@ with open(os.path.join(dir_path, "fuses.json"), 'r') as f:
 
 fuse_version = fuse_defaults['_version']
 del fuse_defaults['_version']
+del fuse_defaults['_schema']
 del fuse_defaults['_comment']
 
 if fuse_version >= pow(2, 8):
@@ -60,20 +65,20 @@ fuses = fuse_defaults.keys()
 initial_config = ""
 getters_h = ""
 getters_cc = ""
-index = len(SENTINEL)
+index = len(SENTINEL) + 1
 for fuse in fuses:
   index += 1
-  initial_config += "1" if fuse_defaults[fuse] else "0"
+  initial_config += fuse_defaults[fuse]
   name = ''.join(word.title() for word in fuse.split('_'))
   getters_h += "bool Is{name}Enabled();\n".replace("{name}", name)
   getters_cc += """
-__attribute__((__visibility__("default"))) bool Is{name}Enabled() {
+FUSE_EXPORT bool Is{name}Enabled() {
   return kFuseWire[{index}] == '1';
 }
 """.replace("{name}", name).replace("{index}", str(index))
 
 header = TEMPLATE_H.replace("{getters}", getters_h.strip())
-impl = TEMPLATE_CC.replace("{sentinel}", SENTINEL).replace("{fuse_version}", chr(fuse_version)).replace("{initial_config}", initial_config).replace("{getters}", getters_cc.strip())
+impl = TEMPLATE_CC.replace("{sentinel}", SENTINEL).replace("{fuse_version}", chr(fuse_version) + chr(len(fuses))).replace("{initial_config}", initial_config).replace("{getters}", getters_cc.strip())
 
 with open(sys.argv[1], 'w') as f:
   f.write(header)
--- a/build/fuses/build.py
+++ b/build/fuses/build.py
@@ -70,9 +70,9 @@ for fuse in fuses:
   index += 1
   initial_config += fuse_defaults[fuse]
   name = ''.join(word.title() for word in fuse.split('_'))
-  getters_h += "bool Is{name}Enabled();\n".replace("{name}", name)
+  getters_h += "FUSE_EXPORT bool Is{name}Enabled();\n".replace("{name}", name)
   getters_cc += """
-FUSE_EXPORT bool Is{name}Enabled() {
+bool Is{name}Enabled() {
   return kFuseWire[{index}] == '1';
 }
 """.replace("{name}", name).replace("{index}", str(index))
--- a/build/fuses/build.py
+++ b/build/fuses/build.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+
+import json
+import os
+import sys
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+SENTINEL = "dL7pKGdnNz796PbbjQWNKmHXBZaB9tsX"
+
+TEMPLATE_H = """
+#ifndef ELECTRON_FUSES_H_
+#define ELECTRON_FUSES_H_
+
+namespace electron {
+
+namespace fuses {
+
+extern const volatile char kFuseWire[];
+
+{getters}
+
+}  // namespace fuses
+
+}  // namespace electron
+
+#endif  // ELECTRON_FUSES_H_
+"""
+
+TEMPLATE_CC = """
+#include "electron/fuses.h"
+
+#include <iostream>
+
+namespace electron {
+
+namespace fuses {
+
+const volatile char kFuseWire[] = "{sentinel}{fuse_version}{initial_config}";
+
+{getters}
+
+}
+
+}
+"""
+
+with open(os.path.join(dir_path, "fuses.json"), 'r') as f:
+  fuse_defaults = json.load(f)
+
+fuse_version = fuse_defaults['_version']
+del fuse_defaults['_version']
+del fuse_defaults['_comment']
+
+if fuse_version >= pow(2, 8):
+  raise Exception("Fuse version can not exceed one byte in size")
+
+fuses = fuse_defaults.keys()
+
+initial_config = ""
+getters_h = ""
+getters_cc = ""
+index = len(SENTINEL)
+for fuse in fuses:
+  index += 1
+  initial_config += "1" if fuse_defaults[fuse] else "0"
+  name = ''.join(word.title() for word in fuse.split('_'))
+  getters_h += "bool Is{name}Enabled();\n".replace("{name}", name)
+  getters_cc += """
+__attribute__((__visibility__("default"))) bool Is{name}Enabled() {
+  return kFuseWire[{index}] == '1';
+}
+""".replace("{name}", name).replace("{index}", str(index))
+
+header = TEMPLATE_H.replace("{getters}", getters_h.strip())
+impl = TEMPLATE_CC.replace("{sentinel}", SENTINEL).replace("{fuse_version}", chr(fuse_version)).replace("{initial_config}", initial_config).replace("{getters}", getters_cc.strip())
+
+with open(sys.argv[1], 'w') as f:
+  f.write(header)
+
+with open(sys.argv[2], 'w') as f:
+  f.write(impl)
--- a/build/fuses/build.py
+++ b/build/fuses/build.py
@@ -40,7 +40,7 @@ namespace electron {
 
 namespace fuses {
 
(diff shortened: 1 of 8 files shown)


build/fuses/build.py
```diff
@@ -0,0 +1,82 @@
...
+
+}  // namespace electron
+
+#endif  // ELECTRON_FUSES_H_
+"""
+
+TEMPLATE_CC = """
+#include "electron/fuses.h"
+
+#include <iostream>
```
nornagon, line 33:
```suggestion
```

shell/app/electron_library_main.mm
```diff
@@ -25,6 +26,9 @@ int ElectronMain(int argc, char* argv[]) {
 
 #if BUILDFLAG(ENABLE_RUN_AS_NODE)
 int ElectronInitializeICUandStartNode(int argc, char* argv[]) {
+  if (!electron::fuses::IsRunAsNodeEnabled())
+    return 1;
```
nornagon, line 30: `CHECK` perhaps? seems like this function should never be reached if `IsRunAsNodeEnabled` is false.

build/fuses/build.py
```diff
@@ -0,0 +1,82 @@
...
+getters_h = ""
+getters_cc = ""
+index = len(SENTINEL)
+for fuse in fuses:
+  index += 1
+  initial_config += "1" if fuse_defaults[fuse] else "0"
+  name = ''.join(word.title() for word in fuse.split('_'))
+  getters_h += "bool Is{name}Enabled();\n".replace("{name}", name)
+  getters_cc += """
+__attribute__((__visibility__("default"))) bool Is{name}Enabled() {
```
nornagon, line 70: I think this needs to be `__declspec` something on windows? and also just `visibility` instead of `__visibility__` on non-windows. see e.g. `//mojo/public/c/system/system_export.h`.

build/fuses/fuses.json
```diff
@@ -0,0 +1,5 @@
+{
+  "_comment": "Modifying this file in any way (removing, adding or changing values) should result in the _version prop being incremented",
```
nornagon, line 2: trying to think of if there's a way we can make this work so that an older tool can (at least sometimes) successfully edit a newer exe. which would be useful because apps wouldn't have to constantly update their `@electron/fuses` dep in order to continue setting the same fuse.

If fuses were only ever added to the end (never removed or changed), and _version was just the number of flags, then the tool could just check that _version > the index of the flag it wants to change.

We could indicate removed flags with a value of `f` instead of `0`/`1`, then `@electron/fuses` could first read the flag to see if it's non-removed, and warn if it tries to set a flag that's been removed.

If we did that scheme, we'd probably also want an escape-hatch that would prevent an out-of-date fuser from modifying the exe, in case we change this fuse scheme. We could either add a 2nd version field (maybe rename `_version` to `_length` and use `_version` for this backwards-incompatible indicator), or we could regenerate the sentinel, though the latter would make the case of trying to edit the wrong binary indistinguishable from the version being wrong, so we'd get better error messages if we keep the sentinel stable.

And of course, `_length` can be automatically computed, so there'd be no need to write it in the JSON file directly :)

build/fuses/build.py
```diff
@@ -0,0 +1,87 @@
...
+  name = ''.join(word.title() for word in fuse.split('_'))
+  getters_h += "FUSE_EXPORT bool Is{name}Enabled();\n".replace("{name}", name)
+  getters_cc += """
+bool Is{name}Enabled() {
+  return kFuseWire[{index}] == '1';
+}
+""".replace("{name}", name).replace("{index}", str(index))
+
+header = TEMPLATE_H.replace("{getters}", getters_h.strip())
+impl = TEMPLATE_CC.replace("{sentinel}", SENTINEL).replace("{fuse_version}", chr(fuse_version) + chr(len(fuses))).replace("{initial_config}", initial_config).replace("{getters}", getters_cc.strip())
```
nornagon, line 81: can we use an `\x00`-style encoding for the fuse version in the generated file?

build/fuses/fuses.json
```diff
@@ -0,0 +1,6 @@
+{
+  "_comment": "Modifying the fuse schema in any way should result in the _version prop being incremented.  NEVER REMOVE A FUSE, instead mark it as removed with '-1'",
```
nornagon, line 2: should the version be incremented even when e.g. adding a fuse?

MarshallOfSound: @nornagon No, just when modifying the scheme (i.e. the 0 vs 1 vs r bit)

nornagon: then this comment should probably read "modifying an existing fuse ..." or something like that.

build/fuses/build.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
+
+import json
+import os
+import sys
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+SENTINEL = "dL7pKGdnNz796PbbjQWNKmHXBZaB9tsX"
```
nornagon, line 9: . o O ( instead of using this sentinel, could we export this as a symbol and then parse ELF/PE/mach-o to directly find the location? )

docs/tutorial/fuses.md
```diff
@@ -0,0 +1,52 @@
...
+
+For a subset of Electron functionality it makes sense to disable certain features for an entire application.  For example, 99% of apps don't make use of `ELECTRON_RUN_AS_NODE`, these applications want to be able to ship a binary that is incapable of using that feature.  We also don't want Electron consumers building Electron from source as that is both a massive technical challenge and has a high cost of both time and money.
+
+Fuses are the solution to this problem, at a high level they are "magic bits" in the Electron binary that can be flipped when packaging your Electron app to enable / disable certain features / restrictions.  Because they are flipped at package time before you code sign your app the OS becomes responsible for ensuring those bits aren't flipped back via OS level code signing validation (Gatekeeper / App Locker).
+
+## How do I flip the fuses?
+
+### The easy way
+
+We've made a handy module `@electron/fuses` to make flipping these fuses easy.  Check out the README of that module for more details on usage and potential error cases.
```
nornagon, line 15: Does this exist yet? We probably shouldn't commit this unless this exists 🤣

docs/tutorial/fuses.md
```diff
@@ -0,0 +1,52 @@
...
+```js
+require('@electron/fuses').flipFuses(
+  // Path to electron
+  require('electron'),
+  // Fuses to flip
+  {
+    runAsNode: false
+  }
+)
+```
```
nornagon, line 26: would be good if this had a command-line interface too, e.g. `npx @electron/fuses path/to/binary --set-fuse run_as_node=1`

docs/tutorial/fuses.md
```diff
@@ -0,0 +1,52 @@
...
+
+### The hard way
+
+#### Quick Glossary
+
+* **Fuse Wire**: A sequence of bytes in the Electron binary used to control the fuses
+* **Sentinel**: A static known sequence of bytes you can use to locate the fuse wire
+* **Fuse Schema**: The format / allowed values for the fuse wire
+
+Manually flipping fuses requires editing the correct Electron binary and modifying the fuse wire to be the sequence of bytes that represent the state of the fuses you want.
```
nornagon, line 36:
```suggestion
Manually flipping fuses requires editing the Electron binary and modifying the fuse wire to be the sequence of bytes that represent the state of the fuses you want.
```

docs/tutorial/fuses.md
```diff
@@ -0,0 +1,52 @@
...
+
+Somewhere in the Electron binary there will be a sequence of bytes that look like this:
+
+```text
+| ...binary | sentinel_bytes | fuse_version | fuse_wire_length | fuse_wire | ...binary |
+```
+
+* `sentinel_bytes` is always this exact string `dL7pKGdnNz796PbbjQWNKmHXBZaB9tsX`
+* `fuse_version` is a single byte whose integer value represents the version of the fuse schema
+* `fuse_wire_length` is a single byte whose integer value represents the number of fuses in the following fuse wire
```
nornagon, line 46:
```suggestion
* `fuse_version` is a single byte whose unsigned integer value represents the version of the fuse schema
* `fuse_wire_length` is a single byte whose unsigned integer value represents the number of fuses in the following fuse wire
```

docs/tutorial/fuses.md
```diff
@@ -0,0 +1,52 @@
...
+* `fuse_wire` is a sequence of N bytes, each byte represents a single fuse and its state.
+  * "0" indicates the fuse is disabled
+  * "1" indicates the fuse is enabled
+  * "r" indicates the fuse is removed and is now a noop
```
nornagon, line 50:
```suggestion
  * "0" (0x30) indicates the fuse is disabled
  * "1" (0x31) indicates the fuse is enabled
  * "r" (0x72) indicates the fuse has been removed and changing the byte to either 1 or 0 will have no effect.
```

docs/tutorial/fuses.md
```diff
@@ -0,0 +1,52 @@
...
+
+To flip a fuse you find its position in the fuse wire and change it to "0" or "1" depending on the state you'd like.
```
nornagon, line 52: We should put a link to the fuse schema here.

docs/tutorial/fuses.md
MarshallOfSound: It does but I wanted to test it against a real Electron release before `npm publish`'ing 😆

docs/tutorial/fuses.md
MarshallOfSound: Potentially, I'm not a fan of adding a `bin` entry point for every module we write because it just adds extra overhead and writing a simple wrapper script isn't that hard.  Can discuss on the `fuses` repo

docs/tutorial/fuses.md
nornagon: I guess this won't show up on the website until Electron 12 becomes stable so go ahead :)

docs/tutorial/fuses.md
nornagon: it'll be handy for testing and manually editing, I think. also makes it easier to integrate into people's CI flows if they're not already all javascripty.

shell/app/electron_library_main.mm
```diff
@@ -25,6 +26,11 @@ int ElectronMain(int argc, char* argv[]) {
 
 #if BUILDFLAG(ENABLE_RUN_AS_NODE)
 int ElectronInitializeICUandStartNode(int argc, char* argv[]) {
+  if (!electron::fuses::IsRunAsNodeEnabled()) {
+    CHECK(false);
```
nornagon, line 30:
```suggestion
    CHECK(false) << "run_as_node fuse is disabled";
```

nornagon (approved):

phamminhvu:
Hi Team,

We tried to follow the instruction for installing the library "fuses" to our electron project?
We used electron version 12.x.x
But looks like It is not ready to use yet.

Thank you very much for feedback!

Best regards,
Jason

PR #24246, opened 2020-06-23
fix: use try/catch for base class converter error

codebytere:
#### Description of Change

Even though this worked previously, we shouldn't be throwing errors in converters because they call into JS and therefore can destroy `this` and cause crashes later. This will be followed up with a refactor to remove the exception from the converter and refactor callsites, but for now this PR mitigates the crash.

cc @nornagon @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/electron_api_base_window.cc
+++ b/shell/browser/api/electron_api_base_window.cc
@@ -98,10 +98,15 @@ BaseWindow::BaseWindow(v8::Isolate* isolate,
   window_->AddObserver(this);
 
 #if defined(TOOLKIT_VIEWS)
+  v8::TryCatch try_catch(isolate);
+
   // Sets the window icon.
   gin::Handle<NativeImage> icon;
   if (options.Get(options::kIcon, &icon) && !icon.IsEmpty())
     SetIcon(icon);
+
+  if (try_catch.HasCaught())
+    LOG(ERROR) << "Failed to convert NativeImage";
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


shell/browser/api/electron_api_base_window.cc
```diff
@@ -98,10 +98,15 @@ BaseWindow::BaseWindow(v8::Isolate* isolate,
   window_->AddObserver(this);
 
 #if defined(TOOLKIT_VIEWS)
+  v8::TryCatch try_catch(isolate);
```
nornagon, line 101: It's good practice to put this `TryCatch` in a scope along with the code it's supposed to surround, e.g.
```cpp
{
  // ... code which might throw an exception ...
}
```
this pattern prevents accidentally catching exceptions which are thrown by code that isn't meant to be wrapped.

in this case, the behavior would be unchanged since there's no code later in this function. It's still good practice to wrap it in a scope though because it defends against code being added to the end of the function later on and accidentally being "inside" the try/catch.

shell/browser/api/electron_api_base_window.cc
```diff
@@ -98,10 +98,15 @@ BaseWindow::BaseWindow(v8::Isolate* isolate,
...
+
   // Sets the window icon.
   gin::Handle<NativeImage> icon;
   if (options.Get(options::kIcon, &icon) && !icon.IsEmpty())
     SetIcon(icon);
+
+  if (try_catch.HasCaught())
+    LOG(ERROR) << "Failed to convert NativeImage";
```
nornagon, line 109: Can we print the path here maybe?

codebytere: unfortunately no - we only have access to the path in the converter

Issue #24244, opened 2020-06-23
powerMonitor does not fire 'suspend" event 


gupta-shivam:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 9.0.2
* **Operating System:**
  * Windows 10
* **Last Known Working Electron version:**
  * 

### Expected Behavior
electron.remote.powerMonitor should have fired the 'suspend' event and eventually 'The system is going to sleep' should have been printed in the console.

### Actual Behavior
No event is received.

### To Reproduce
Try to put your computer on sleep from the Start menu of windows.
And then check the logs after unlocking the computer after some time.

[Github Gist link](https://gist.github.com/gupta-shivam/d364e186dfc0a2cd177c9078c9a576eb)

gupta-shivam:
I have tried it again after your changes, but this issue has not been fixed for me

codebytere:
@gupta-shivam a version of 9 with the changes has not been released yet.

gupta-shivam:
I tried with a local electron build which has these changes by taking help from @alexeykuzmin

PR #24251, opened 2020-06-23
feat: support suspend/resume on Windows

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/24244.

Add support for `suspend` and `resume` on Windows.

cc @zcbenz @MarshallOfSound @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added support for suspend and resume events to Windows.

### DIFF
--- a/docs/api/power-monitor.md
+++ b/docs/api/power-monitor.md
@@ -8,11 +8,11 @@ Process: [Main](../glossary.md#main-process)
 
 The `powerMonitor` module emits the following events:
 
-### Event: 'suspend'
+### Event: 'suspend' _Linux_ _Windows_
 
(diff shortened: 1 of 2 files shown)


deepak1556:
`base/power_monitor` seems to handle suspend and resume events on windows https://source.chromium.org/chromium/chromium/src/+/master:base/power_monitor/power_monitor_device_source_win.cc . Can you verify if this is a broken behavior.

codebytere:
@deepak1556 looks like we can pretty easily use that approach too and support it - updated the PR. I also have suspend/resume support working for macOS but i'll put it in another PR.

Issue #19313, opened 2019-07-17
Build the node_snapshot through GN


MarshallOfSound:
Currently we use the stub snapshot in our build system, we should support `node_mksnapshot` properly.

Refs: https://github.com/nodejs/node/blob/master/node.gyp#L454-L474

Issue #19311, opened 2019-07-17
Build the node code cache through GN


MarshallOfSound:
Currently we use the stub node code cache, we should support the `mkcodecache` binary in our build system.

Refs: https://github.com/nodejs/node/blob/master/node.gyp#L429-L449

nornagon:
does that conflict with mksnapshot?

MarshallOfSound:
They're two different things afaik.

MarshallOfSound:
So quick update, made it build, then everything crashes, so that's fun 🤷‍♂ I'll get back to this at some point but not going to invest to much time rn

PR #24252, opened 2020-06-23
chore: minify internal JS code

MarshallOfSound:
This improves boot performance by 40-50ms on my `testing` build.  The raw size of the JS + the overwhelmingly large sourcemap results in the JS script compiler doing far too much work.  If we make it easier for it, unsurprisingly, it gets faster.

There is a trade off here that JS exceptions internal to Electron may be harder to read, but that's reasonable IMO and it's definitely not that hard to read minified JS.  This also removes the sourcemap because it's stupidly big and doesn't work properly anyway.

Notes: no-notes

### DIFF
--- a/build/webpack/webpack.config.base.js
+++ b/build/webpack/webpack.config.base.js
@@ -1,6 +1,7 @@
 const fs = require('fs')
 const path = require('path')
 const webpack = require('webpack')
+const TerserPlugin = require('terser-webpack-plugin');
 
 const electronRoot = path.resolve(__dirname, '../..')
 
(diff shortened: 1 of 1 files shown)


nornagon:
Should we consider packaging a snapshot rather than the raw js?

MarshallOfSound:
Yeah, long term we should definitely figure out how to pre-compile some of this JS or even better put it (along with all of the other js2c code) into a snapshot:

https://github.com/electron/electron/issues/19313
https://github.com/electron/electron/issues/19311

But that's quite technically challenging still so we're stuck doing this for a while

codebytere:
>  may be harder to read

do you have a sample of what the new traces would look like?

MarshallOfSound:
> do you have a sample of what the new traces would look like?

They'll just be minified variable names, so instead of "Cannot read property X of "niceNameHere".  It'll be "Cannot read property X of a", breakpoints and all that jazz will still work as normal.  It's just our code is minified so variable names might be changed.

PR #24254, opened 2020-06-23
feat: support suspend/resume on macOS

codebytere:
#### Description of Change

Refs https://github.com/electron/electron/issues/24244 and https://github.com/electron/electron/pull/24251.

This adds suspend/resume support to macOS.

cc @deepak1556 @MarshallOfSound @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added support for suspend and resume events to macOS.

### DIFF
--- a/shell/browser/api/electron_api_power_monitor_mac.mm
+++ b/shell/browser/api/electron_api_power_monitor_mac.mm
@@ -32,6 +32,21 @@
                    selector:@selector(onScreenUnlocked:)
                        name:@"com.apple.screenIsUnlocked"
                      object:nil];
+
+    // A notification that the workspace posts before the machine goes to sleep.
+    [[[NSWorkspace sharedWorkspace] notificationCenter]
+        addObserver:self
+           selector:@selector(isSuspending:)
+               name:NSWorkspaceWillSleepNotification
+             object:nil];
+
+    // A notification that the workspace posts when the machine wakes from
+    // sleep.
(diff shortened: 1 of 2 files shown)


shell/browser/api/electron_api_power_monitor_mac.mm
```diff
@@ -45,6 +60,18 @@ - (void)addEmitter:(electron::api::PowerMonitor*)monitor_ {
   self->emitters.push_back(monitor_);
 }
 
+- (void)isSuspending:(NSNotification*)notify {
+  for (auto*& emitter : self->emitters) {
```
deepak1556, line 64: This was not introduced in this PR, but there is no advantage to taking a reference to the pointer here.

`for (auto*& emitter : self->emitters)` => `for (auto* emitter : self->emitters)`

same for all below.

codebytere: @deepak1556 this was done to fix chromium style errors: https://github.com/electron/electron/commit/cdc502230561d7e67c019034a6ffce9d7d748c40

deepak1556: Yes that was because we were not specifying the type for raw pointer.

`for (auto* emitter : self->emitters)` // valid
`for (auto emitter : self->emitters)` // not valid

PR #15620, opened 2018-11-07
feat: Windows implementation of powerMonitor "shutdown" event

tarruda:
#### Description of Change
The powerMonitor "shutdown" event is only handled on Linux/Mac, this PR adds the Windows implementation.

Handling shutdown correctly on Windows is more complex than Linux/Mac, since any process can be killed during shutdown. That means renderer processes can be killed first (and usually are) , which could result in loss of data even if the event is handled in the browser process.

#### Checklist

- [ ] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [ ] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

#### Release Notes

Notes: Add Windows implementation of powerMonitor "shutdown" event

Issue #15880, opened 2018-11-29
session-end doesn't fire at all


agjs:
Electron version: 3.0.3
OS: Windows 10 Pro, version: 1803

Yesterday I've opened [session-end doesn't fire on Electron 1.7.10](https://github.com/electron/electron/issues/15864) thinking that it's an issue with Electron 1.7.10. This now at least for us, turns to be the case with Electron in general, regardless of the version (we are currently on 3.0.3).

To minimize any side-effects when testing, I've used Electron Starter and here is the test code:

```js
const { app, BrowserWindow } = require('electron');
const path = require('path');
const fs = require('fs');

let mainWindow;

function createWindow() {
  mainWindow = new BrowserWindow({ width: 800, height: 600 });
  mainWindow.loadFile(path.join(__dirname, '../renderer/index.html'));
  mainWindow.on('closed', function() {
    mainWindow = null;
  });

  mainWindow.on('session-end', function() {
    fs.mkdirSync(path.join(app.getPath('downloads'), 'test123'));
  });
}

app.on('ready', createWindow);
app.on('window-all-closed', function() {
  if (process.platform !== 'darwin') {
    app.quit();
  }
});

app.on('activate', function() {
  if (mainWindow === null) {
    createWindow();
  }
});
```
**Expectation:** 
1. Run the app
2. Restart your machine 
3. A new folder inside Downloads folder should be created

**Result** 
Folder is not created. It seams that the event is not fired at all.

Based on my research and very minimal understanding of Windows API's, this is how session-end should be invoked:

1. We restart the computer
2. System sends WM_QUERYENDSESSION messages
3. System sends WM_ENDSESSION to all the applications that returned TRUE.

By using [SendMessage](https://tools.stefankueng.com/SendMessage.html) I was able to play with this a bit. If I directly send WM_ENDSESSION to our app process with _WPARAM: endsession_, event fires.

This issue could also be very well related to [callbacks for hookWindowMessage should have their return value sent to OS](https://github.com/electron/electron/issues/8762).

If the context from above article is true and we indeed don't return anything on *WM_QUERYENDSESSION*, why do we then expect to get *WM_ENDSESSION*? Doesn't WM_ENDSESSION only fires for those applications that return TRUE to *WM_QUERYENDSESSION*?

Please, feel free to ask for any more information, I'm willing to help.

agjs:
@juturu By reading the source code, I noticed you implemented this event. Any idea of what could be wrong?

[...]

PR #24261, opened 2020-06-23
feat: support 'shutdown' on Windows

codebytere:
#### Description of Change

Take 2 of https://github.com/electron/electron/pull/15620.

From initial PR:

> Handling shutdown correctly on Windows is more complex than Linux/Mac, since any process can be killed during shutdown. That means renderer processes can be killed first (and usually are) , which could result in loss of data even if the event is handled in the browser process.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Add Windows implementation of powerMonitor "shutdown" event


shell/browser/lib/shutdown_blocker_win.cc
```diff
@@ -0,0 +1,130 @@
...
+      ShutdownBlockerWin* blocker = reinterpret_cast<ShutdownBlockerWin*>(
+          GetWindowLongPtr(hwnd, GWLP_USERDATA));
+      LOG(INFO) << "Received WM_QUERYENDSESSION"
+                << (!blocker ? ""
+                             : (blocker->dedicated_message_loop_
+                                    ? " on renderer process"
+                                    : " on browser process"));
+      if (blocker && !blocker->OnQueryEndSession()) {
+        LOG(INFO) << "Shutdown blocked";
+        ShutdownBlockReasonCreate(hwnd, L"Ensure a clean shutdown");
```
anaisbetts, line 120: I would consider making this string configurable - this gets shown on shutdown to users and if it's hard-coded to English, this will result in a Confusing Experience™™™ to people using non-English locales!

codebytere: oh excellent point, will do!

shell/browser/lib/shutdown_blocker_win.cc
```diff
@@ -0,0 +1,130 @@
...
+  MSG msg;
+  while (GetMessage(&msg, NULL, 0, 0)) {
+    TranslateMessage(&msg);
+    DispatchMessage(&msg);
+  }
+  DestroyHiddenWindow();
+}
+
+void ShutdownBlockerWin::CreateHiddenWindow() {
+  ATOM atom = RegisterWindowClass();
```
jkleinsc, line 64: :atom: ???

jkleinsc: :electron: instead?

lib/renderer/win32-queryendsession-setup.ts
```diff
@@ -0,0 +1,8 @@
+import { ipcRenderer } from 'electron';
+
+export function setup (binding: typeof process['_linkedBinding']) {
+  // TODO(codebytere): fix typedef here.
+  (binding as any).setShutdownHandler(() => {
```
zcbenz, line 5: The shutdown handler in renderer returns `undefined`, which would then always cancel the shutdown.

shell/browser/lib/shutdown_blocker_win.h
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) 2020 Microsoft, Inc.
+// Use of this source code is governed by the MIT license that can be
+// found in the LICENSE file.
+//
+// The purpose of this class is to block system shutdown on windows so the
+// application can exit cleanly. It uses the WM_QUERYENDSESSION message to do
+// so, as explained here
+// https://msdn.microsoft.com/en-us/library/windows/desktop/aa376890(v=vs.85).aspx
+//
+// The WM_QUERYENDSESSION message must be handled in every every process managed
```
zcbenz, line 10: I'm not familiar with the `WM_QUERYENDSESSION` message, but if it must be handled in every process then we also need to handle it in GPU/utility/network/... processes then.

shell/browser/lib/shutdown_blocker_win.cc
zcbenz: The `ATOM` here is a win32 thing: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerclassa.

jkleinsc: thanks @zcbenz I didn't know that!

LorenzoBiondani:
Any updates on this?

codebytere:
@LorenzoBiondani none at the moment but i haven't forgotten about it! i'll work on it more this week

LorenzoBiondani:
@codebytere thank you, I don't have much time that I can dedicate to this but if you need something let me know, I will try to help.

shell/browser/lib/shutdown_blocker_win.h
mkrufky, line 10: typo / repeated word:

"handled in every every process" should be "handled in every process"

shell/browser/lib/shutdown_blocker_win.cc
mkrufky: Calling `ShutdownBlockReasonCreate()` inside the `WM_QUERYENDSESSION` handler allows us to notify our code to clean up and shutdown quickly, but we can also use it preemptively when we know that we're running a critical section, like burning a cd, or saving a file to disk.

It would be *really* nice if an app could call a "block" or "unblock" API when the programmer knows in advance where a shutdown should be blocked, releasing the block afterwards.  The "block" API would call `ShutdownBlockReasonCreate(hwnd, L"A critical task is running.");`   The "unblock" API would call `ShutdownBlockReasonDestroy(hwnd);`  The string passed into `ShutdownBlockReasonCreate()` is shown by windows as the reason for the shutdown being held up.

[Application Shutdown Changes in Windows](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ms700677(v=vs.85))

shell/browser/lib/shutdown_blocker_win.cc
```diff
@@ -0,0 +1,130 @@
...
+      ShutdownBlockerWin* blocker = reinterpret_cast<ShutdownBlockerWin*>(
+          GetWindowLongPtr(hwnd, GWLP_USERDATA));
+      LOG(INFO) << "Received WM_QUERYENDSESSION"
+                << (!blocker ? ""
+                             : (blocker->dedicated_message_loop_
+                                    ? " on renderer process"
+                                    : " on browser process"));
+      if (blocker && !blocker->OnQueryEndSession()) {
+        LOG(INFO) << "Shutdown blocked";
+        ShutdownBlockReasonCreate(hwnd, L"Ensure a clean shutdown");
```
mkrufky, line 120: Calling `ShutdownBlockReasonCreate()` inside the `WM_QUERYENDSESSION` handler allows us to notify our code to clean up and shutdown quickly, but we can also use it preemptively when we know that we're running a critical section, like burning a cd, or saving a file to disk.

It would be *really* nice if an app could call a "block" or "unblock" API when the programmer knows in advance where a shutdown should be blocked, releasing the block afterwards.  The "block" API would call `ShutdownBlockReasonCreate(hwnd, L"A critical task is running.");`   The "unblock" API would call `ShutdownBlockReasonDestroy(hwnd);`  The string passed into `ShutdownBlockReasonCreate()` is shown by windows as the reason for the shutdown being held up.

[Application Shutdown Changes in Windows](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ms700677(v=vs.85))

giftedunicorn:
Is anyone pushing this merge?

cberescu:
any updates on this?

zhushenwudi:
> @LorenzoBiondani none at the moment but i haven't forgotten about it! i'll work on it more this week

hello, your pull request is not merge successful, can you fix it and pull again? 3Q

nornagon:
Related: https://github.com/electron/electron/issues/15880

gergof:
Any updates on this?

savely-krasovsky:
Can someone explain why we need to handle events from both browser and renderer?

cc @codebytere

dladeira:
Was this issue addressed in the end?

Docs say that shutdown for powerMonitor is Linux and MacOS only
https://www.electronjs.org/docs/latest/api/power-monitor#event-shutdown-linux-macos

PR #24264, opened 2020-06-23
feat: deprecate nodeIntegration option

nornagon:
#### Description of Change
This deprecates the `nodeIntegration` option, per #23506. In Electron 12, `nodeIntegration` will be enabled in all non-sandboxed renderers.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Deprecated the `nodeIntegration` option.

### DIFF
--- a/lib/browser/api/web-contents.js
+++ b/lib/browser/api/web-contents.js
@@ -556,6 +556,9 @@ WebContents.prototype._init = function () {
     if (prefs.webviewTag && prefs.contextIsolation) {
       electron.deprecate.log('Security Warning: A WebContents was just created with both webviewTag and contextIsolation enabled.  This combination is fundamentally less secure and effectively bypasses the protections of contextIsolation.  We strongly recommend you move away from webviews to OOPIF or BrowserView in order for your app to be more secure');
     }
+    if (prefs.nodeIntegration != null && !prefs.nodeIntegration) {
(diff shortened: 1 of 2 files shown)


lib/browser/api/web-contents.js
```diff
@@ -556,6 +556,9 @@ WebContents.prototype._init = function () {
     if (prefs.webviewTag && prefs.contextIsolation) {
       electron.deprecate.log('Security Warning: A WebContents was just created with both webviewTag and contextIsolation enabled.  This combination is fundamentally less secure and effectively bypasses the protections of contextIsolation.  We strongly recommend you move away from webviews to OOPIF or BrowserView in order for your app to be more secure');
     }
+    if (prefs.nodeIntegration != null && !prefs.nodeIntegration) {
+      electron.deprecate.log('Deprecation Warning: Disabling nodeIntegration is deprecated. In Electron 12, all non-sandboxed renderers will have node integration enabled. See https://github.com/electron/electron/issues/23506 for more details.');
```
MarshallOfSound, line 560: Can we fit a "You should be using contextIsolation instead" kind of message in here as well?

stevenjoezhang:
Any updates on this thread?

nornagon:
Closing this for now. Removing the `nodeIntegration` option and defaulting to `nodeIntegration: true` in all non-sandboxed renderers would break web content that includes AMD-type modules that test for the presence of node globals to work out whether they're in a browser or not. Possibly we should remove the option for `nodeIntegration: true` and require a preload script to expose globals, but I'm less convinced about the utility of that option.

PR #24275, opened 2020-06-24
feat: implement systemPreferences.getMediaAccessStatus() on Windows

miniak:
#### Description of Change
Implement systemPreferences.getMediaAccessStatus() on Windows

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Implemented `systemPreferences.getMediaAccessStatus()` on Windows.

### DIFF
--- a/docs/api/system-preferences.md
+++ b/docs/api/system-preferences.md
@@ -416,7 +416,7 @@ This API itself will not protect your user data; rather, it is a mechanism to al
 
 Returns `Boolean` - `true` if the current process is a trusted accessibility client and `false` if it is not.
 
-### `systemPreferences.getMediaAccessStatus(mediaType)` _macOS_
+### `systemPreferences.getMediaAccessStatus(mediaType)` _Windows_ _macOS_
 
 * `mediaType` String - Can be `microphone`, `camera` or `screen`.
 
@@ -425,6 +425,9 @@ Returns `String` - Can be `not-determined`, `granted`, `denied`, `restricted` or
 This user consent was not required on macOS 10.13 High Sierra or lower so this method will always return `granted`.
 macOS 10.14 Mojave or higher requires consent for `microphone` and `camera` access.
 macOS 10.15 Catalina or higher requires consent for `screen` access.
+
+Windows 10 has a global setting controlling `microphone` and `camera` access for all win32 applications.
+It will always return `granted` for `screen` and for all media types on older versions of Windows.
 
 ### `systemPreferences.askForMediaAccess(mediaType)` _macOS_
 
--- a/shell/browser/api/electron_api_system_preferences_win.cc
+++ b/shell/browser/api/electron_api_system_preferences_win.cc
@@ -3,10 +3,13 @@
 // found in the LICENSE file.
 
 #include <dwmapi.h>
+#include <windows.devices.enumeration.h>
+#include <wrl/client.h>
 #include <iomanip>
 
 #include "shell/browser/api/electron_api_system_preferences.h"
 
+#include "base/win/core_winrt_util.h"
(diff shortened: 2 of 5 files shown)


shell/browser/api/electron_api_system_preferences_win.cc
```diff
@@ -117,6 +166,24 @@ std::string SystemPreferences::GetColor(gin_helper::ErrorThrower thrower,
...
+    return ConvertDeviceAccessStatus(
+        GetDeviceAccessStatus(DeviceClass::DeviceClass_VideoCapture));
+  } else if (media_type == "microphone") {
+    return ConvertDeviceAccessStatus(
+        GetDeviceAccessStatus(DeviceClass::DeviceClass_AudioCapture));
+  } else if (media_type == "screen") {
+    return ConvertDeviceAccessStatus(
+        DeviceAccessStatus::DeviceAccessStatus_Allowed);
+  } else {
+    args->ThrowError("Invalid media type");
```
deepak1556, line 182: Not originating from this PR but when we have a `unknown` state in the docs, why do we want to throw an error ? seem to do it for macOS as well.

shell/browser/api/electron_api_system_preferences_win.cc
```diff
@@ -117,6 +166,24 @@ std::string SystemPreferences::GetColor(gin_helper::ErrorThrower thrower,
...
+    gin_helper::Arguments* args) {
+  if (media_type == "camera") {
+    return ConvertDeviceAccessStatus(
+        GetDeviceAccessStatus(DeviceClass::DeviceClass_VideoCapture));
+  } else if (media_type == "microphone") {
+    return ConvertDeviceAccessStatus(
+        GetDeviceAccessStatus(DeviceClass::DeviceClass_AudioCapture));
+  } else if (media_type == "screen") {
+    return ConvertDeviceAccessStatus(
+        DeviceAccessStatus::DeviceAccessStatus_Allowed);
```
codebytere, line 180: We should note in the documentation that `screen` access is always allowed on Windows, i'd say.

shell/browser/api/electron_api_system_preferences_win.cc
MarshallOfSound: @deepak1556 The return type can have `unknown`, the `media_type` argument can't be `unknown`

docs/api/system-preferences.md
```diff
@@ -426,6 +426,9 @@ This user consent was not required on macOS 10.13 High Sierra or lower so this m
 macOS 10.14 Mojave or higher requires consent for `microphone` and `camera` access.
 macOS 10.15 Catalina or higher requires consent for `screen` access.
 
+Windows 10 has a global setting controlling `microphone` and `camera` access for all win32 applications.
```
MarshallOfSound, line 429: If this is specific for win32 applications should we safety check if for APPX / Window Store builds?

miniak: The system API returns the global value, it should work fine. I took the code from Chromium.

shell/browser/api/electron_api_system_preferences_win.cc
deepak1556: For an unknown media type why can't the return type by unknown, the current system calls both on windows and macOS never return an unknown state, the value is being unused. I wouldn't mind if this state wasn't documented but it is :)

deepak1556: Would be good to bring this up with api-WG

miniak: I can address it in a separate PR.

deepak1556: Thanks!

PR #19043, opened 2019-06-29
refactor: pass webPreferences over mojo IPC instead of command-line

miniak:
#### Description of Change
Pass all required `webPreferences` (such as the list of preload scripts) via native mojo IPC as the command-line is shared when the same renderer process is hosting multiple `BrowserWindows` (scriptable popup with  `nativeWindowOpen: true` for example).

Fixes #18466

#### Checklist
- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes
Notes: Preload scripts configured for popups when handling the `new-window` event now work properly regardless of whether a new renderer process is hosting the contents (cross-origin) or the same renderer (same-origin).

Issue #18466, opened 2019-05-27
Background color ignored on scriptable BrowserWindow


zarubond:
`BrowserWindow` parameter `backgroundColor` is ignored (the background stays white) if it is created with native chrome `window.open`.

### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
3.1.6
* **Operating System:**
Windows 
* **Last Known Working Electron version:**
Never

### Expected Behavior
The window should start-up with defined color

### Actual Behavior
The window start with white color

### To Reproduce
Create a child window with native Chrome window.open (scriptable window) as shown here https://electronjs.org/docs/api/window-open , then set backgroundColor constructor property of BrowserWindow to some color. The result is that it is ignored. The color is set just fine if I do not use the window.open.

miniak:
@zarubond the reason is that the scriptable popup shares the render process with the parent window. The background color is passed via the command-line, which is therefore shared.

zarubond:
Unfortunately there is a bug in passing color from parent too. If I add the backgroundColor to the constructor parameter of parent BrowserWindow, then the color is passed to the script-able window. But if I set the color to parent with setBackgroundColor and then open up the script-able window, the color is not passed to it. So I can not change background color during lifetime of application which is very inconvenient for users who want to switch themes.

iMoses:
> @zarubond the reason is that the scriptable popup shares the render process with the parent window. The background color is passed via the command-line, which is therefore shared.

I tried setting the parent `backgroundColor` hoping it will effect it's children, but instead the parent itself simply flushes that color for a second before resetting it to white, and then both the children and the parent background color is white.

miniak:
Should be fixed by #24517, which landed in Electron 12

PR #24291, opened 2020-06-24
feat: add preloadSrc web preference

nornagon:
#### Description of Change
This allows specifying the source of a preload script as a string, rather than by a file path.

This is marked [WIP] for two reasons:

1. This API should be considered within the broader scope of what we're thinking about for the future of preload scripts, including but not limited to:
  - defining multiple different kinds of preload scripts depending on the context type (e.g. an extensions preload script)
  - precompiling preload scripts
  - bundling precompiled preload scripts with the app
  - automatically precompiling/caching preload scripts in the main process, rather than reading them off disk every time
2. The way this system works in renderers with node integration enabled is odd, in that preload scripts are loaded with `Module._load`, which requires a file path to function. How this should interact with `preloadSrc` is unclear to me.

However, I'm putting this PR up in its WIP state to gather feedback and design critique to hopefully move this feature forward & make preload scripts faster and better.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added the `preloadSrc` web preference, which allows preload scripts to be specified by their source code instead of by file name.


miniak:
@nornagon we should later use `ELECTRON_BROWSER_LOAD` to pass more webPreferences to the renderer instead of using the commandline, which is causing issues when multiple webContents live in the same renderer process, such as #18466

MarshallOfSound:
@miniak the act of moving webprefs to be IPC based instead of cmd line based is being worked on but I don't think that's directly related to this PR.

miniak:
@MarshallOfSound I have an old PR where I was experimenting with that, but using a native mojo call (#19043). Are you working on it?

MarshallOfSound:
I'm not, @loc is though

deepak1556:
Precompiling and loading cached data for preload scripts is a good one to have. https://nodejs.org/docs/latest-v12.x/api/vm.html#vm_script_createcacheddata should serve for spitting out v8 code cache

For the preload strings, do we generate them on every run of the app ?. And for preload scripts, even if we generate cache data we would still have to write them to disk, then we would be comparing the performance of loading via `Module._load` vs loading cached data from disk.

Do you have other ideas for caching ?

Certainly benchmarks on these would be helpful for general design of the api.

nornagon:
> Do you have other ideas for caching ?

My plan was to expose the necessary functionality and leave the strategy to developers / 3rd party libraries. There are real tradeoffs between convenience, flexibility and performance and I'm not sure we can choose the right way for every use case.

e.g.

- it's convenient to re-read from disk every time during development, because then you can edit / cmd+r and not have to restart the whole app. so we shouldn't cache forever.
- compiling on first run might ultimately make launch slower.
- holding onto the preload script's source code in the main process could be an undesirable memory cost

the api we decide on should allow apps to make the right trade-off for their use case.

deepak1556:
That sums up pretty well, thanks!

Just to add, incase the app decides the necessity for a cached preload src then we are responsible for generating it. And if we are looking to promote different kinds of preload then maybe [session.setPreloads](https://github.com/electron/electron/blob/master/docs/api/session.md#sessetpreloadspreloads) can serve as the main source of serving preloads instead of the webPreference container.

```
session.setPreloads([
  {
     source: <Buffer> | <Path> | <DataView>,
     kind: 'extension' | 'BrowserWindow' | 'webView',
     produceCachedData: <Boolean>,
     origin: <url::Origin>
  }
])
```

nornagon:
. o O ( I wonder if we should support something like [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob)s as preload sources...? )

deepak1556:
That would be a cool exploration for storing and transferring large preload sources.

nornagon:
I still think this idea is worth pursuing, but this PR is pretty stale by now and I haven't had time to work on it.

PR #24301, opened 2020-06-25
fix: do not register the node ESM loader in renderer processes

MarshallOfSound:
Only one ESM loader can be registered per **isolate**, in renderer processes this should be blink.  This patches node so that it won't register its handler (overriding blinks) in non-browser processes.

Notes: no-notes


deepak1556:
Is it possible to add test for this change ?

codebytere:
@deepak1556 and @MarshallOfSound i think yes - we added the sandbox change to ARM to account for this, so we should be able to remove it now. If it fails, we know this didn't have the intended effect.

deepak1556:
Failing vanilla node tests are related.

codebytere:
oh boy missed those - i think we'll need to ensure it's enabled in run as node mode?

codebytere:
@MarshallOfSound can you rebase this against the accepted upstream solution: https://github.com/nodejs/node/pull/34060

Issue #24290, opened 2020-06-24
Web Notifications API: notifications don't always close from Mac Notification Center when close function is called


jflasch:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v8.3.1
* **Operating System:**
  * macOS 10.14.6
* **Last Known Working Electron version:**
  * N/A

### Expected Behavior
I have a simple electron app that generates a notification when a button is clicked. Before the new notification is generated, `close()` is called on the last notification. I would expect that each time the button is clicked, the existing notification closes and a new one pops up.

### Actual Behavior
When the button is clicked a new notification is generated, but the old notification does not close. It is important to note that I am logging in the `onclose` event handler for all notifications, and this log shows that all notifications are being closed successfully even though they remain in the Mac Notification Center. I also have logs in the `onerror` event handler for each notification, and these logs are never getting printed, so no errors are occurring.

### To Reproduce
Using the sample app provided in the fiddle link below, click on the 'Generate Notification' button ~20 times. Do some clicks back to back, and spread out some clicks by a few seconds. Check the Mac Notification Center and you will see that most of the notifications are still there, even though only the most recent notification should be there. There are rare occasions where the close does work, but it is still unknown how to reproduce this result consistently. Check the renderer logs as well to see that the `onclose` function was triggered for each notification, even though they are still in the Notification Center.

**Electron Fiddle Link**: https://gist.github.com/fd16a663434667d2096cd31ec7f2662f

### Additional Information
This problem is only seen when running an electron app on Mac. If you run the exact renderer.js and index.html files from the fiddle code in a real chrome web browser, it works correctly (old notifications always clear when new ones come in).
I have confirmed that `nodeIntegration` and the notification option `requireInteraction` have no effect on this regardless of if their values are true or false.

Issue #17706, opened 2019-04-05
Notifications that shares the same tag won't trigger onclick on every second notification


r0stig:
### Issue Details

* **Electron Version:**
  * 4.1.4
* **Operating System:**
  * macOS 10.13.6
* **Last Known Working Electron version:**:

### Expected Behavior
The onclick-handler for a notification in the renderer process should always be fired upon click when the notification object shares the same tag.

### Actual Behavior
Only every second notification click fires the onclick handler when the notifications share the same tag.

### To Reproduce

https://github.com/r0stig/electron-quick-start

```sh
$ git clone git@github.com:r0stig/electron-quick-start.git
$ npm install
$ npm start
```
* Click on the button "send notification"
* Click on the notification and notice the log messages that indicate the click-handler fired
* Click on the button "send notification"
* Click on the notification and see that this time the onclick-handler did not fire

The same occur when you just wait for the notification to close from timeout:
* Click on the button "send notification"
* Wait for the notification to be closed automatically
* Click on the button "send notification"
* Click on the notification and notice that the onclick-handler did not fire

### Screenshots
![notification_electron](https://user-images.githubusercontent.com/1148042/55620492-5048b380-579b-11e9-885c-bc4081c6a01e.gif)
As shown in the gif, when I click on the notification a second time the onclick handler won't trigger.

r0stig:
Letting you know that this is resolved in 5.0.0 but still present in 4.1.5.

Ivorfason:
> Letting you know that this is resolved in 5.0.0 but still present in 4.1.5.

It still exist in 5.0.0

r0stig:
> > Letting you know that this is resolved in 5.0.0 but still present in 4.1.5.
> 
> It still exist in 5.0.0

Are you able to reproduce with the above demo? If not how do you reproduce?

electron-triage:
Thank you for taking the time to report this issue and helping to make Electron better.

The version of Electron you reported this on has been superseded by newer releases. 

If you're still experiencing this issue in Electron 6.x.y or later, please add a comment specifying the version you're testing with and any other new information that a maintainer trying to reproduce the issue should know.

I'm setting the `blocked/need-info` label for the above reasons. This issue will be closed 7 days from now if there is no response.

Thanks in advance! Your help is appreciated.

electron-triage:
Thank you for your issue!

[...]

PR #24302, opened 2020-06-25
fix: remove same-tag notifications before showing new ones

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/24290.

Fixes an issue where notifications where not being closed properly when more than one notification was created in the renderer process with the same tag.

Per spec: https://notifications.spec.whatwg.org/#showing-a-notification

Screen Shot 2020-07-01 at 9 15 14 PM

Tested with https://gist.github.com/fd16a663434667d2096cd31ec7f2662f.

cc @MarshallOfSound @zcbenz  

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where some old notifications were not properly removed from the Notification Center on macOS.

### DIFF
--- a/shell/browser/notifications/mac/cocoa_notification.mm
+++ b/shell/browser/notifications/mac/cocoa_notification.mm
@@ -111,8 +111,6 @@ void CocoaNotification::Dismiss() {
 
   NotificationDismissed();
 
-  this->LogAction("dismissed");
-
   notification_.reset(nil);
 }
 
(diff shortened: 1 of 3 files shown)


shell/browser/notifications/mac/cocoa_notification.mm
```diff
@@ -165,8 +166,10 @@
 void CocoaNotification::LogAction(const char* action) {
   if (getenv("ELECTRON_DEBUG_NOTIFICATIONS")) {
     NSString* identifier = [notification_ valueForKey:@"identifier"];
-    LOG(INFO) << "Notification " << action << " (" << [identifier UTF8String]
-              << ")";
+    if (identifier) {
```
codebytere, line 169: if there's no identifier on a notification this will crash on nullptr, so add a safety check

r0stig:
I think this PR introduced an old bug, explained in this issue: https://github.com/electron/electron/issues/17706

Quick tldr: when showing a new notification with the same tag as the one before both the newly created notification and the one before gets it's onclose callback called and the user can't click on the notification (the onclick handler won't fire).

codebytere:
@r0stig this PR was already reverted - it shouldn't be the cause.

PR #24318, opened 2020-06-26
build: build squirrel and its dependencies from source

MarshallOfSound:
This updates our build to build `Squirrel.framework`, `Mantle.framework` and `ReactiveObjC.framework` from source (Used to be `ReactiveCocoa.framework`).  Things that changed:

* Removed old frameworks from `external-binaries`
* Added `BUILD.gn` file to Squirrel.Mac to build the three frameworks
* Updated Squirrel.Mac and Electron to use `ReactiveObjC` (a renamed and slightly newer version of ReactiveCocoa)
* Apply one patch to `ReactiveCocoa` as the way it consumes headers doesn't quite work with the _interesting_ way that GN builds framework files

Squirrel.Mac GN Changes: https://github.com/electron/Squirrel.Mac/commit/c7186061da815eb82dfba181caaac6f98e729b5c

This needs to be a fork commit as it needs a `DEPS` file to be processed.

We have a E2E updater test (phew) so this is as safe as this change is gonna get.

Notes: no-notes


patches/ReactiveObjC/.patches
```diff
@@ -0,0 +1 @@
+build_conditionally_import_ext_headers_from_framework_or_from.patch
```
nornagon, line 1: why do we have a fork _and_ patches? could we settle on one or the other?

MarshallOfSound: The fork is for Squirrel.Mac, as in the PR:

> This needs to be a fork commit as it needs a DEPS file to be processed.

We can't patch in a `DEPS` file.

This patch is not for Squirrel.Mac, it is for ReactiveObjC a separate repo.  Here we can use a patch so we're using our existing system.  Squirrel.Mac is just an unfortunate exception

nornagon: we could fold the `DEPS` content from squirrel.mac into `//electron/DEPS`?

MarshallOfSound: @nornagon That would be semantically incorrect 🤔  I did it this way so that the DEPS chain actually made sense.  Electron depends on squirrel which depends on these two frameworks

nornagon: maybe, but there's a real benefit to having patches in-repo, which is why we did it for all our dependencies. i don't think it's worth sacrificing real benefit for the abstract concept of "semantic correctness".

BUILD.gn
```diff
@@ -863,17 +862,49 @@ if (is_mac) {
     }
   }
 
+  template("striped_framework") {
```
ahallicks, line 865: Should this be `stripped_framework`?

patches/squirrel.mac/build_add_gn_config.patch
```diff
@@ -0,0 +1,764 @@
...
++import("//build/config/mac/rules.gni")
++import("build/xcrun.gni")
++import("filenames.gni")
++
++bundle_data("mantle_headers") {
++  sources = []
++  public_deps = []
++  sources += mantle_filenames.headers
++  outputs = [ "{{bundle_contents_dir}}/Headers/{{source_file_part}}" ]
++}
```
nornagon, line 41: do we still need these if we're no longer bundling the headers?

patches/squirrel.mac/build_add_gn_config.patch
```diff
@@ -0,0 +1,764 @@
...
++  framework_version = "A"
++  framework_contents = [
++    "Headers",
++    "Resources",
++  ]
++  info_plist = "vendor/Mantle/Mantle/Info.plist"
++  extra_substitutions = [ "CURRENT_PROJECT_VERSION=0.0.0" ]
++
++  configs -= [
++    "//build/config/compiler:chromium_code",
```
nornagon, line 54: + `:no_chromium_code`

patches/squirrel.mac/build_add_gn_config.patch
```diff
@@ -0,0 +1,764 @@
...
++  ]
++  info_plist = "vendor/ReactiveObjC/ReactiveObjC/Info.plist"
++  extra_substitutions = [
++    "CURRENT_PROJECT_VERSION=0.0.0",
++    "PRODUCT_BUNDLE_IDENTIFIER=com.electron.reactive",
++  ]
++  defines = [ "BUILDING_RAC_FRAMEWORK=1" ]
++
++  configs -= [
++    "//build/config/compiler:chromium_code",
```
nornagon, line 118: + `:no_chromium_code`

patches/squirrel.mac/build_add_gn_config.patch
```diff
@@ -0,0 +1,764 @@
...
++    "//build/config/gcc:symbol_visibility_hidden",
++  ]
++  public_deps = [
++    ":reactiveobjc_disposableprovider_dtrace",
++    ":reactiveobjc_headers",
++    ":reactiveobjc_signalprovider_dtrace",
++  ]
```
nornagon, line 125: Can these be non-public?

.circleci/config.yml
```diff
@@ -855,7 +855,7 @@ step-save-out-cache: &step-save-out-cache
   save_cache:
     paths:
       - ./src/out/Default
-    key: v8-out-cache-{{ checksum "src/electron/.depshash" }}-{{ checksum "src/electron/.depshash-target" }}
+    key: v9-out-cache-{{ checksum "src/electron/.depshash" }}-{{ checksum "src/electron/.depshash-target" }}
```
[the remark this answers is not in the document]
MarshallOfSound: @nornagon GN's framework building logic sucks and doesn't clean up things in the `.framework` folder that shouldn't be there.  E.g. The old `ReactiveCocoa.framework` which caused zip manifest failures

patches/squirrel.mac/build_add_gn_config.patch
MarshallOfSound: We need the headers during build time, but we don't want them when we zip up the app

BUILD.gn
```diff
@@ -863,17 +862,49 @@ if (is_mac) {
...
+    action(target_name) {
+      assert(defined(invoker.framework))
+
+      script = "//electron/build/strip_framework.py"
```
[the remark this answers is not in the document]
MarshallOfSound: Yes, otherwise the zip manifest checks would fail locally + the build might pass locally but fail unexpectedly for prod builds

patches/ReactiveObjC/build_conditionally_import_ext_headers_from_framework_or_from.patch
```diff
@@ -0,0 +1,533 @@
...
++++ b/ReactiveObjC/NSControl+RACTextSignalSupport.m
+@@ -7,7 +7,11 @@
+ //
+ 
+ #import "NSControl+RACTextSignalSupport.h"
++#if defined(BUILDING_RAC_FRAMEWORK)
++#import "extobjc/EXTScope.h"
++#else
+ #import <ReactiveObjC/EXTScope.h>
++#endif
```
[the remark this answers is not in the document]
MarshallOfSound: ReactiveObjC itself hasn't changed in over a year, I think this patch is OK

.circleci/config.yml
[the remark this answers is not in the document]
nornagon: do these things actually need to be frameworks or can we statically link them?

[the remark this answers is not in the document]
MarshallOfSound: I want to avoid drastic shifts in this PR and maintain the status quo.  Happy to investigate building these things inline in the future though, their are issues with for instance hard coded path expectations and the ShipIt helper binary.  But it might be possible

PR #24322, opened 2020-06-26
chore: tsify net-log

nornagon:
more types good

Notes: none


lib/browser/api/net-log.ts
```diff
@@ -0,0 +1,22 @@
+// TODO(deepak1556): Deprecate and remove standalone netLog module,
+// it is now a property of session module.
+import { app, session } from 'electron';
+
+export const startLogging: typeof session.defaultSession.netLog.startLogging = async (path, options) => {
```
MarshallOfSound, line 5: Can we make these exports:

```js
export default {
  startLogging: ...,
  stopLogging: ...,
  get currentlyLogging() {
    ...
  }
};
```

To avoid the `Object.defineProperty` call

PR #24323, opened 2020-06-26
feat: MessagePorts in the main process

miniak:
Backport of #22404

Notes: Added support for MessagePort in the main process.


MarshallOfSound:
Just noting here but will also discuss at the WG meeting, IMO this is far too big / untested to backport into a minor release at this time

nornagon:
I think the risk here is relatively low as there's not much code changed when not using MessagePort. The only diff here is that the serialize/deserialize functions got moved into a new file. haven't gone over them to check for potential differences / issues, but that's the only area of concern i have re: breaking existing apps.

PR #24325, opened 2020-06-26
chore: tsify web-contents

nornagon:
Also, deleted `WebContents.sendToAll` which isn't documented or used anywhere.

Notes: none


lib/browser/api/web-contents.ts
```diff
@@ -67,8 +75,8 @@ const PDFPageSizes = {
 // Default printing setting
 const defaultPrintingSetting = {
```
MarshallOfSound, line 76: Instead of all the `as` casts below we should type this as `Electron.WhateverThePrintingSettingsInterfaceIs`

lib/browser/api/web-contents.ts
```diff
@@ -436,14 +433,30 @@ const addReturnValueToEvent = (event) => {
 // Add JavaScript wrappers for WebContents class.
 WebContents.prototype._init = function () {
   // The navigation controller.
-  NavigationController.call(this, this);
+  const navigationController = new NavigationController(this);
+  this.loadURL = navigationController.loadURL.bind(navigationController);
```
MarshallOfSound, line 437: This is subtly different behaviour to what we had before.  Before we were assigning to the prototype, here we are actually creating instance properties.  Not sure if that's a problem but worth raising as a difference

lib/browser/api/web-contents.ts
```diff
@@ -360,14 +357,14 @@ WebContents.prototype.print = function (options = {}, callback) {
           throw new Error('height and width properties are required for pageSize');
         }
         // Dimensions in Microns - 1 meter = 10^6 microns
-        options.mediaSize = {
+        (options as any).mediaSize = {
```
MarshallOfSound, line 360: These `any` casts shouldn't be needed if the original `options` object is typed correctly

lib/browser/api/web-contents.ts
nornagon: yep, the reason it is this way is that you can't call the constructor of a class without `new`.

lib/browser/api/web-contents.ts
nornagon: I don't think there exists any type in electron.d.ts that matches this.

```
$ rg '\bpageRange\b'
lib/browser/api/web-contents.ts
78:  pageRange: [] as {from: number, to: number}[],
279:    printSettings.pageRange = [{
$
```

MarshallOfSound: @nornagon That's because according to the docs it's `pageRanges`

nornagon: This is also true of several other members of this struct:

```
$ rg '\bshouldPrintBackgrounds\b'
lib/browser/api/web-contents.ts
84:  shouldPrintBackgrounds: false,
258:    printSettings.shouldPrintBackgrounds = options.printBackground;
$ rg '\bheaderFooterEnabled\b'
lib/browser/api/web-contents.ts
81:  headerFooterEnabled: false,
287:    printSettings.headerFooterEnabled = true;
$ rg '\bgenerateDraftData\b'
lib/browser/api/web-contents.ts
96:  generateDraftData: true,
```

I don't think the struct you're thinking of is the same as this struct.

nornagon: I believe the type of this struct is related to [these](https://source.chromium.org/chromium/chromium/src/+/master:printing/print_job_constants.h;l=15-77;drc=e454f2607b49d8eee3f5d40d09c3d4b9c0dd2572)

PR #24327, opened 2020-06-26
test: wrap expect() in try/catch in spec/asar-spec.js

miniak:
#### Description of Change
Makes the tests fail properly instead of timing out.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes


alexeykuzmin:
Wouldn't it be better to use fsPromises here?

PR #24344, opened 2020-06-29
fix: crash due to race between attach and destruction of webview

deepak1556:
#### Description of Change

This is an edge case that was discovered in ci when a page holding webview is created and destroyed rapidly but after some debugging this is possible to trigger by users too depending on usage. 

Thanks @rebornix for help with the detailed debugging.

Refs https://github.com/microsoft/vscode/issues/99408 , https://github.com/microsoft/vscode/issues/100321

The sequence of action before this PR were as follows:

1) `<webview>` DOM element is created and attached to DOM tree by the user
2) starts parsing the `src` attribute which inturn signals the browser process via an async ipc to create the backing webcontents via `CreateGuest` in `guest-view-manager`
3) The above ipc returns a guestInstanceId that is used to uniquely identify the webview element
4) The renderer then signals the browser process via another async ipc to attach the backing webcontents in the browser frame tree
5) When the `<webview>` is removed from the DOM tree by the user this will signal a sync ipc to cleanup stuff on the browser process.

All this works well most of the time, but there can be situations where a page holding the webview is destroyed between step  3) and 4) , which causes the attach ipc call to cause UAF crashes in `electron::WebViewGuestDelegate::AttachToIframe`

This patch fixes the scenario by combining the creation and attaching of backing webcontents in a single ipc call that will avoid the race.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed crash in webview creation casued by UAF in the browser process


deepak1556:
@nornagon hence the draft state 😄

rebornix:
I spent a few days trying to get a minimal reproduce on vanilla Electron but I found the test is simply about removing the webview **shortly** after the webview is created (sequence `1, 2, 3, 5, 4` per https://github.com/electron/electron/pull/24344#issue-441193606). On a relatively powerful machine, that's nanoseconds.

For example, I wrote a test like below and I can only crash the process 1-2 out of 100 times on some bad days (I'm guessing it's when my machine is on heavy loads)

```js
const v8Util = process.electronBinding('v8_util');
const ipcRendererInternal = v8Util.getHiddenValue(global, 'ipc-internal');
const { ipc } = process.electronBinding('ipc');
function timeoutAsync(n) {
        setTimeout(() => {
            resolve();
        }, n);
    });
}
const loadWebView = async (id, container, webview, attributes = {}) => {
    for (const [name, value] of Object.entries(attributes)) {
        webview.setAttribute(name, value);
    }
    container.appendChild(webview);
    setTimeout(() => {
        ipcRendererInternal.sendSync('ELECTRON_GUEST_VIEW_MANAGER_DETACH_GUEST', id + 1);
    }, 2);
};
async function test() {
    for (let i = 0; i < 500; i++) {
        let webview = new WebView();
        const container = document.createElement('div');
        document.body.appendChild(container);
        loadWebView(i + 1, container, webview, {
            src: `https://s3.amazonaws.com/duhaime/blog/visualizations/isolation-forests.html`
        });
        await timeoutAsync(10);
    }
}
window.onload = function () {
    test()
}
```

Even on our CI https://dev.azure.com/vscode/VSCode/_build/results?buildId=48185&view=logs&j=a5e52b91-c83f-5429-4a68-c246fc63a4f7&t=855ed636-0983-5316-9b28-7efa8f10a65b, on which we are seeing most of the crashes, we only saw crashes 1 out of 5 to 10 times. It made me think that if we are going to have a test, which can crash the process 100% of the time, we need to either 

* mimic the CI environment and make the CI machine under as heavy load as VS Code integration tests.
* or, have code in electron/chromium directly to slow down the execution, to trigger the racing condition.

but none of above sounds valuable. I'm wondering if there is any other approach that we can take to move this forward?

zcbenz:
I'm good without a test in this case.

lib/browser/guest-view-manager.js
```diff
@@ -142,7 +143,11 @@ const createGuest = function (embedder, params) {
     }
   });
 
-  return guestInstanceId;
+  if (attachGuest(event, embedderFrameId, elementInstanceId, guestInstanceId, params)) {
+    return guestInstanceId;
+  } else {
```
codebytere, line 148: nit: we can probably drop the unnecessary `else` here :)

lib/browser/guest-view-manager.js
```diff
@@ -166,11 +171,13 @@ const attachGuest = function (event, embedderFrameId, elementInstanceId, guestIn
...
   // If this isn't a valid guest instance then do nothing.
   if (!guestInstance) {
-    throw new Error(`Invalid guestInstanceId: ${guestInstanceId}`);
+    console.error(new Error(`Guest attach failed: Invalid guestInstanceId: ${guestInstanceId}`));
+    return false;
   }
   const { guest } = guestInstance;
   if (guest.hostWebContents !== event.sender) {
-    throw new Error(`Access denied to guestInstanceId: ${guestInstanceId}`);
+    console.error(new Error(`Guest attach failed: Access denied to guestInstanceId: ${guestInstanceId}`));
```
codebytere, line 179: if this no longer throws, will a user still be made as clearly aware of failure if they don't see the console log?

nornagon:
I'm a little uneasy that this fixes a UAF by shuffling some JavaScript around. It should be impossible to cause a UAF from JS. Can we address this on the C++ side?

deepak1556:
>  It should be impossible to cause a UAF from JS.

For webview it becomes possible because its backing `webContents` lifetime is tied to either the embedder on the main process or the DOM node on the renderer unlike the other JS objects we manage. Hence this mostly boils down to the mapping we maintain in `guest-view-manager`, there is a case that is not updating the mapping correctly during destruction of the embedder which would the root cause here. But our destruction sequence is quite complex to mess with, hence this turned out to be the simplest solution. I agree that this doesn't fix the root cause but eliminates the abnormalities around webview creation.

> Can we address this on the C++ side?

any ideas ?

miniak:
@deepak1556 can you please rebase on top of latest master?

nornagon:
We now have ASan tests enabled in master (on Linux only). That might make it easier to create a reliably-failing test.

miniak:
I am getting this crash when running tests locally using testing build:
```
./out/testing/Electron.app/Contents/MacOS/Electron electron/spec --grep 'webview'
1..79
ok 1 chromium feature window.open disables the <webview> tag when it is disabled on the parent window
[82992:0507/021415.448658:FATAL:web_view_guest_delegate.cc(39)] Check failed: embedder_web_contents_ == content::WebContents::FromRenderFrameHost(embedder_frame) (0x7fd2b289ca00 vs. 0x0)
0   Electron Framework                  0x0000000113d3ae79 base::debug::CollectStackTrace(void**, unsigned long) + 9
1   Electron Framework                  0x0000000113c56fe3 base::debug::StackTrace::StackTrace() + 19
2   Electron Framework                  0x0000000113c6fa6f logging::LogMessage::~LogMessage() + 175
3   Electron Framework                  0x0000000113c7096e logging::LogMessage::~LogMessage() + 14
4   Electron Framework                  0x000000010f70b1f9 electron::WebViewGuestDelegate::AttachToIframe(content::WebContents*, int) + 217
5   Electron Framework                  0x000000010f6644ca gin_helper::Dispatcher<void (electron::api::WebContents*, content::WebContents*, int)>::DispatchToCallback(v8::FunctionCallbackInfo<v8::Value> const&) + 474
6   Electron Framework                  0x0000000111021dc2 v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo) + 994
7   Electron Framework                  0x0000000111020354 v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) + 1444
8   Electron Framework                  0x000000011101e988 v8::internal::Builtin_Impl_HandleApiCall(v8::internal::BuiltinArguments, v8::internal::Isolate*) + 456
9   Electron Framework                  0x000000011101e52d v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) + 109
10  ???                                 0x0000003a000c56d8 0x0 + 249108911832
11  ???                                 0x0000003a0004b1e1 0x0 + 249108410849
12  ???                                 0x0000003a0004b1e1 0x0 + 249108410849
13  ???                                 0x0000003a0004b1e1 0x0 + 249108410849
Task trace:
0   Electron Framework                  0x00000001141147de mojo::internal::MultiplexRouter::MaybePostToProcessTasks(base::SequencedTaskRunner*) + 158
1   Electron Framework                  0x0000000118d7ce38 ui::RecyclableCompositorMacFactory::RecycleCompositor(std::__1::unique_ptr<ui::RecyclableCompositorMac, std::__1::default_delete<ui::RecyclableCompositorMac> >) + 472
2   Electron Framework                  0x000000010f63dead electron::api::WebContents::Destroy() + 109
3   Electron Framework                  0x000000011413273f mojo::SimpleWatcher::Context::Notify(unsigned int, MojoHandleSignalsState, unsigned int) + 447

Trace/BPT trap: 5
```

miniak (approved):

miniak:
I am not seeing those crashes anymore.

miniak:
@zcbenz, @codebytere can you please review?

zcbenz (approved):

ckerr (approved):

deepak1556:

PR #24359, opened 2020-06-29
feat: add name to app.getAppMetrics() output

miniak:
#### Description of Change
Add `name` to `app.getAppMetrics()` output.

Examples:
- `Audio Service`
- `Content Decryption Module Service`
- `Network Service`
- `Video Capture`

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `name` to `app.getAppMetrics()` output.

### DIFF
--- a/docs/api/structures/process-metric.md
+++ b/docs/api/structures/process-metric.md
@@ -11,6 +11,8 @@
   * `Pepper Plugin`
   * `Pepper Plugin Broker`
   * `Unknown`
+* `name` String (optional) - The name of the process. i.e. for plugins it might be Flash.
+    Examples for utility: `Audio Service`, `Content Decryption Module Service`, `Network Service`, `Video Capture`, etc.
 * `cpu` [CPUUsage](cpu-usage.md) - CPU usage of the process.
 * `creationTime` Number - Creation time for this process.
     The time is represented as number of milliseconds since epoch.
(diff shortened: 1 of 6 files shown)


docs/api/structures/process-metric.md
```diff
@@ -11,6 +11,8 @@
   * `Pepper Plugin`
   * `Pepper Plugin Broker`
   * `Unknown`
+* `serviceName` String (optional) - Name of the service hosted in the utility process.
+    Examples: `Audio Service`, `Content Decryption Module Service`, `Network Service`, `Video Capture`, etc.
```
alexeykuzmin, line 15: Do we have a whole list of services? Can we use an `enum` here instead of a `string`?

miniak: I was trying to get it, but it's complicated, there a lot of services and I don't know, which ones apply to Electron

deepak1556: If there is interest only for the services and not all the utility process then `content::ChildProcessData::sandbox_type` can be used which is a predefined enum for each service.

miniak: for the pepper plugin host this given the plugin name, so it cannot be an enum

deepak1556: well pepper plugin host is not a service but a generic child process, I was suggesting the sandboxed enum because of 

> serviceName` String (optional) - Name of the service hosted in the utility process.

but I see that you have corrected it to `name String (optional) - The name of the process. i.e. for plugins it might be Flash.`. Thanks this looks better.

PR #24367, opened 2020-06-30
feat: add utility-process-gone event to app

miniak:
#### Description of Change
Allows reporting utility process issues in application telemetry.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `utility-process-gone` event to `app`.


docs/api/app.md
```diff
@@ -403,6 +408,36 @@ Returns:
 Emitted when the renderer process unexpectedly dissapears.  This is normally
 because it was crashed or killed.
 
+#### Event: 'child-process-gone'
```
MarshallOfSound, line 411: This is awkward naming because of the child_process module in node.  Is there another name we can use for this that wouldn't conflict with that existing concept

nornagon: IMO this is a different enough context that the namespace collision doesn't bother me, and it's valuable i think to match Chromium's naming where possible.

jkleinsc:
The API WG approved this PR at our July 13th 2020 meeting

PR #24373, opened 2020-06-30
test: convert more tests to async / await

miniak:
#### Description of Change
Makes the tests more readable.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/spec-main/api-app-spec.ts
+++ b/spec-main/api-app-spec.ts
@@ -5,6 +5,7 @@ import * as http from 'http';
 import * as net from 'net';
 import * as fs from 'fs';
 import * as path from 'path';
+import { promisify } from 'util';
 import { app, BrowserWindow, Menu, session } from 'electron/main';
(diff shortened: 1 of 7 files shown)


spec-main/api-app-spec.ts
```diff
@@ -891,53 +892,37 @@ describe('app module', () => {
...
 
-      classesKey.keys((error: Error, keys: any[]) => {
-        if (error) throw error;
-
-        const exists = !!keys.find(key => key.key.includes(protocol));
-        expect(exists).to.equal(true);
-
-        done();
-      });
+      const keys = await promisify(classesKey.keys).call(classesKey) as any[];
```
jkleinsc, line 898: @miniak why does this need to be async/await? It doesn't appear to be an async operation?

jkleinsc: Oops... I missed the fact that it was callback based before... nevermind.

PR #24376, opened 2020-07-01
chore: TS-ify Guest View Manager

silvestrijonathan:
#### Description of Change
Converts guest-view-manager.js to TypeScript

#### Checklist

- [ ] `npm test` passes
- [X] This is **NOT A BREAKING CHANGE**. Breaking changes may not be merged to master until 11-x-y is branched.

Failing tests from `npm test` are due to async timeouts.

Notes: none

### DIFF
--- a/lib/browser/guest-view-manager.ts
+++ b/lib/browser/guest-view-manager.ts
@@ -1,16 +1,30 @@
-'use strict';
+import { NewWindowEvent, Referrer, IpcMainEvent, IpcMainInvokeEvent, webContents } from 'electron/main';
+import { ipcMainInternal } from '@electron/internal/browser/ipc-main-internal';
+import * as ipcMainUtils from '@electron/internal/browser/ipc-main-internal-utils';
+import { parseWebViewWebPreferences } from '@electron/internal/common/parse-features-string';
+import { syncMethods, asyncMethods, properties } from '@electron/internal/common/web-view-methods';
+import { serialize } from '@electron/internal/common/type-utils';
 
-const { webContents } = require('electron');
-const { ipcMainInternal } = require('@electron/internal/browser/ipc-main-internal');
-const ipcMainUtils = require('@electron/internal/browser/ipc-main-internal-utils');
-const { parseWebViewWebPreferences } = require('@electron/internal/common/parse-features-string');
-const { syncMethods, asyncMethods, properties } = require('@electron/internal/common/web-view-methods');
-const { serialize } = require('@electron/internal/common/type-utils');
+import type { BrowserWindow, WebContents } from 'electron/main';
+
+enum Disposition {
+  default = 'default',
+  ['foreground-tab'] = 'foreground-tab',
+  ['background-tab'] = 'background-tab',
+  ['new-window'] = 'new-window',
+  ['save-to-disk'] = 'save-to-disk',
+  other = 'other'
+}
+
+interface WebVM {
+  addGuest(guestInstanceId: number, elementInstanceId: number, embedder: WebContents, guest: any, webPreferences: {[key: string]: any}): void;
+  removeGuest(embedder: WebContents, guestInstanceId: number): void
+}
 
 // Doesn't exist in early initialization.
-let webViewManager = null;
+let webViewManager: WebVM | null = null;
 
-const supportedWebViewEvents = [
+const supportedWebViewEvents: Array<string> = [
   'load-commit',
   'did-attach',
   'did-finish-load',
@@ -46,10 +60,10 @@ const supportedWebViewEvents = [
   'update-target-url'
 ];
 
-const guestInstances = {};
-const embedderElementsMap = {};
+const guestInstances: { [key: string]: WebContents & Electron.WebContentsInternal | any; } = {};
+const embedderElementsMap: { [key: string]: number } = {};
 
-function sanitizeOptionsForGuest (options) {
+function sanitizeOptionsForGuest (options: BrowserWindow) {
   const ret = { ...options };
   // WebContents values can't be sent over IPC.
   delete ret.webContents;
@@ -57,12 +71,12 @@ function sanitizeOptionsForGuest (options) {
 }
 
 // Create a new guest instance.
-const createGuest = function (embedder, params) {
+const createGuest = function (embedder: Electron.WebContentsInternal & WebContents, params: Record<string, any>): number {
   if (webViewManager == null) {
     webViewManager = process._linkedBinding('electron_browser_web_view_manager');
   }
 
-  const guest = webContents.create({
+  const guest = (webContents as any).create({
     type: 'webview',
     partition: params.partition,
     embedder: embedder
@@ -81,7 +95,7 @@ const createGuest = function (embedder, params) {
   });
 
   // Init guest web view after attached.
-  guest.once('did-attach', function (event) {
+  guest.once('did-attach', function (this: WebContents, event: Electron.IpcRendererEvent): void {
     params = this.attachParams;
     delete this.attachParams;
 
@@ -94,7 +108,7 @@ const createGuest = function (embedder, params) {
     }
 
     if (params.src) {
-      const opts = {};
+      const opts: {[key: string]: string} = {};
       if (params.httpreferrer) {
         opts.httpReferrer = params.httpreferrer;
       }
@@ -106,15 +120,15 @@ const createGuest = function (embedder, params) {
     embedder.emit('did-attach-webview', event, guest);
   });
 
-  const sendToEmbedder = (channel, ...args) => {
+  const sendToEmbedder = (channel: string, ...args: any[]) => {
     if (!embedder.isDestroyed()) {
-      embedder._sendInternal(`${channel}-${guest.viewInstanceId}`, ...args);
+      embedder._sendInternal(`${channel}-${guest.viewInstanceId}`, args);
     }
   };
 
   // Dispatch events to embedder.
-  const fn = function (event) {
-    guest.on(event, function (_, ...args) {
+  const fn = function (event: string) {
+    guest.on(event, function (_: any, ...args: any[]) {
       sendToEmbedder('ELECTRON_GUEST_VIEW_INTERNAL_DISPATCH_EVENT', event, ...args);
     });
   };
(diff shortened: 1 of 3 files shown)


silvestrijonathan:
Notes: This is "complete" in that things are typed and `e build` brings back no errors, but I had to do a few things that I'm not entirely sure are correct. I left comments above the odd areas. I also think some more typings for the returns of functions, where appropriate, could be added.

lib/browser/guest-view-manager.ts
```diff
@@ -46,23 +60,24 @@ const supportedWebViewEvents = [
...
 // Create a new guest instance.
-const createGuest = function (embedder, params) {
+const createGuest = function (embedder: Electron.WebContentsInternal & WebContents, params: Record<string, any>): number {
   if (webViewManager == null) {
     webViewManager = process._linkedBinding('electron_browser_web_view_manager');
   }
 
-  const guest = webContents.create({
+  // TS saying .create doesn't exist, could not find it as I walked back the dep tree.
+  const guest = (webContents as any).create({
```
MarshallOfSound, line 80: This is a private internal method so it should be declared in `internal-electron.d.ts`.  The syntax is:

```
create(opts: Electron.WebPreferences): Electron.WebContents
```

lib/browser/guest-view-manager.ts
```diff
@@ -81,7 +96,7 @@ const createGuest = function (embedder, params) {
   });
 
   // Init guest web view after attached.
-  guest.once('did-attach', function (event) {
+  guest.once('did-attach', function (this: any, event: Electron.IpcRendererEvent): void {
```
MarshallOfSound, line 99: `this` is the "guest" so should be typed as `WebContents`

lib/browser/guest-view-manager.ts
```diff
@@ -94,7 +109,7 @@ const createGuest = function (embedder, params) {
     }
 
     if (params.src) {
-      const opts = {};
+      const opts: {[key: string]: string} = {};
```
MarshallOfSound, line 112:
```suggestion
      const opts: Electron.LoadURLOptions = {};
```

lib/browser/guest-view-manager.ts
```diff
@@ -192,7 +208,8 @@ const attachGuest = function (event, embedderFrameId, elementInstanceId, guestIn
       ? parseWebViewWebPreferences(params.webpreferences)
       : null;
 
-  const webPreferences = {
+  // Typing satisfy's no index signature error on webPreferences in for loop
+  const webPreferences: { [key: string]: any } = {
```
MarshallOfSound, line 212:
```suggestion
  const webPreferences: Electron.WebPreferences = {
```

lib/browser/guest-view-manager.ts
```diff
@@ -203,13 +220,10 @@ const attachGuest = function (event, embedderFrameId, elementInstanceId, guestIn
     webSecurity: !params.disablewebsecurity,
     enableBlinkFeatures: params.blinkfeatures,
     disableBlinkFeatures: params.disableblinkfeatures,
+    ...Boolean(params.preload) && { preloadUrl: params.preloadURL },
```
MarshallOfSound, line 223: This `Boolean` trick shouldn't be needed, can we restore the original `if`

lib/browser/guest-view-manager.ts
```diff
@@ -223,7 +237,8 @@ const attachGuest = function (event, embedderFrameId, elementInstanceId, guestIn
   ]);
 
   // Inherit certain option values from embedder
-  const lastWebPreferences = embedder.getLastWebPreferences();
+  // Typing satisfy's no index signature error on lastWebPreferences in for loop
+  const lastWebPreferences: { [key: string]: any } = embedder.getLastWebPreferences();
```
MarshallOfSound, line 241:
```suggestion
  const lastWebPreferences: Electron.WebPreferences = embedder.getLastWebPreferences();
```

lib/browser/guest-view-manager.ts
```diff
@@ -245,23 +260,25 @@ const attachGuest = function (event, embedderFrameId, elementInstanceId, guestIn
   guestInstance.elementInstanceId = elementInstanceId;
 
   watchEmbedder(embedder);
-
-  webViewManager.addGuest(guestInstanceId, elementInstanceId, embedder, guest, webPreferences);
+  if (webViewManager) {
```
MarshallOfSound, line 263: We should assert that `webViewManager` exists here instead of truthy checking, we want to do `webViewManager!` and let it error if it doesn't exist, it's a theoretical impossible case.

lib/browser/guest-view-manager.ts
```diff
@@ -179,7 +195,7 @@ const attachGuest = function (event, embedderFrameId, elementInstanceId, guestIn
     delete embedderElementsMap[oldKey];
 
     // Remove guest from embedder if moving across web views
-    if (guest.viewInstanceId !== params.instanceId) {
+    if (webViewManager && guest.viewInstanceId !== params.instanceId) {
```
MarshallOfSound, line 198: We should assert that `webViewManager` exists here instead of truthy checking, we want to do `webViewManager!` and let it error if it doesn't exist, it's a theoretical impossible case.

lib/browser/guest-view-manager.ts
```diff
@@ -245,23 +260,25 @@ const attachGuest = function (event, embedderFrameId, elementInstanceId, guestIn
...
 
   if (!guestInstance) return;
 
   if (embedder !== guestInstance.embedder) {
     return;
   }
-
-  webViewManager.removeGuest(embedder, guestInstanceId);
-  delete guestInstances[guestInstanceId];
+  if (webViewManager) {
```
MarshallOfSound, line 278: We should assert that `webViewManager` exists here instead of truthy checking, we want to do `webViewManager!` and let it error if it doesn't exist, it's a theoretical impossible case.

lib/browser/guest-view-manager.ts
```diff
@@ -305,8 +322,9 @@ const watchEmbedder = function (embedder) {
 };
 
 const isWebViewTagEnabledCache = new WeakMap();
+type isWebViewTagEnabledCacheType = typeof isWebViewTagEnabledCache
```
MarshallOfSound, line 325: This is just `WeakMap` so we shouldn't need this type alias?

lib/browser/guest-view-manager.ts
```diff
@@ -326,32 +344,32 @@ const makeSafeHandler = function (channel, handler) {
...
 });
 
-handleMessageSync('ELECTRON_GUEST_VIEW_MANAGER_DETACH_GUEST', function (event, guestInstanceId) {
+handleMessageSync('ELECTRON_GUEST_VIEW_MANAGER_DETACH_GUEST', function (event: IpcMainEvent, guestInstanceId: number): void {
   return detachGuest(event.sender, guestInstanceId);
 });
 
 // this message is sent by the actual <webview>
-ipcMainInternal.on('ELECTRON_GUEST_VIEW_MANAGER_FOCUS_CHANGE', function (event, focus, guestInstanceId) {
+ipcMainInternal.on('ELECTRON_GUEST_VIEW_MANAGER_FOCUS_CHANGE', function (event: ElectronInternal.IpcMainInternalEvent, focus: WindowProxy['focus'], guestInstanceId: number): boolean | void {
```
MarshallOfSound, line 372:
```suggestion
ipcMainInternal.on('ELECTRON_GUEST_VIEW_MANAGER_FOCUS_CHANGE', function (event: ElectronInternal.IpcMainInternalEvent, focus: boolean, guestInstanceId: number): boolean | void {
```

silvestrijonathan:
@MarshallOfSound Thank you for all your feedback! I'm going to implement it and see if anything changes w/ the tests.

I posted in #ask-anything, but I'll cross-post here, as it relates to your q:

"When on master, if you run `e test --runners=main --trace-uncaught` in your terminal, with the tests in webview-spec::webview<tag> focused, does anyone else experience timeouts with some of the tests? They are breaking on my open PR right now and I want to fix them before getting a review, but when I switched back to master and ran the same tests, the same thing happened. The first test passes, but the second test is where things block up"

lib/browser/guest-view-manager.ts
silvestrijonathan: Appreciate that, this is the one that had me the most confused.

lib/browser/guest-view-manager.ts
silvestrijonathan: Thanks for the guidance on how to treat this!

lib/browser/guest-view-manager.ts
silvestrijonathan: Changing to this typing causes the error that I mention in my comment on line 211.

lib/browser/guest-view-manager.ts
silvestrijonathan: Going to also add `preloadURL` to `WebPreferences` in this case as it isn't on that typing right now.

lib/browser/guest-view-manager.ts
silvestrijonathan: See comment on line 240 for why I could not do this.

Ran into this error for each associated variable in the loop:

```javascript
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'WebPreferences'.
  No index signature with a parameter of type 'string' was found on type 'WebPreferences'.
```

Googled around for a bit but couldn't find something that worked for this case.

lib/browser/guest-view-manager.ts
silvestrijonathan: Will also be adding `attachParams` and `viewInstanceId` to `WebContents` to satisfy TS.

lib/browser/guest-view-manager.ts
silvestrijonathan: hmm, adding this to the typings in that file still does not resolve the issue where .create does not exist.

silvestrijonathan:
Ack, something really weird happen w/ a rebase. Please ignore as of now.

silvestrijonathan:
@codebytere @MarshallOfSound I've gone about as far as I can w/ this branch solo.

Something of note - locally, a number of tests that have been starting to fail w/ my changes stopped failing when I imported `create` from this file: `lib/browser/api/web-contents.ts` as used it where I currently am using `webContents as any`. Still resulted in test failures, however, just less.

Sam, I'm afraid adding the typing for this function did not fix the associated `.ts` error.

miniak:
@silvestrijonathan can you please rebase on top of latest master?

silvestrijonathan:
@miniak done!

miniak:
@silvestrijonathan would you mind if I take this over? I've done `guest-window-manager` recently.

silvestrijonathan:
go for it @miniak. I've had a lot less time to work on this than I would have liked to lately and I'm still not certain why my results are producing breaking tests.

Issue #24393, opened 2020-07-02
webContents.executeJavaScript throws error around serializing arguments for IPC


Galkon:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 9.0.5 and 8.3.4
* **Operating System:**
  * macOS 10.14.5 (18F203)
* **Last Known Working Electron version:**
  * 8.x (throws deprecation warnings constantly, but it works)

### Expected Behavior
I expect to be able to use `webContents.executeJavaScript(codeString, callback)` where `codeString` is just a string of JS to execute, without getting any deprecation warnings for IPC serialization on electron 8.x, and without serialization errors on 9.x.

### Actual Behavior
Using `webContents.executeJavaScript` in any way triggers the serialization warning/errors, even when the code is a very simple string.

### To Reproduce
In my app we have a splash screen which we update a loading message programmatically using the following:
```
await new Promise (resolve => webContents.executeJavaScript(`document.getElementById('update-message').innerHTML = '${state}'`, resolve));
```

But this results in the following error:
```
This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
Error: Failed to serialize arguments
    at WebContents../lib/browser/api/web-contents.js.WebContents._sendInternal (electron/js2c/browser_init.js:3789:15)
    at electron/js2c/browser_init.js:6224:20
    at new Promise (<anonymous>)
    at Object.invokeInWebContents (electron/js2c/browser_init.js:6204:12)
    at WebContents../lib/browser/api/web-contents.js.WebContents.executeJavaScript (electron/js2c/browser_init.js:3854:23)
```

nornagon:
`executeJavaScript` does not take a callback. It returns a Promise. The 2nd argument is `hasUserGesture`. https://www.electronjs.org/docs/api/web-contents#contentsexecutejavascriptcode-usergesture

PR #24396, opened 2020-07-02
fix: coerce executeJavaScript arguments

nornagon:
#### Description of Change
Hopefully this will result in less confusing failures when the wrong type of
thing is passed. Ref #24393.

Also fixes an issue with an incorrect argument list in `executeJavaScriptInIsolatedWorld` resulting in the 3rd argument (`hasUserGesture`) being ignored.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed executeJavaScriptInIsolatedWorld not respecting `hasUserGesture` argument.


codebytere:
Is this testable somehow?

lib/browser/api/web-contents.js
```diff
@@ -207,11 +207,11 @@ const waitTillCanExecuteJavaScript = async (webContents) => {
...
 WebContents.prototype.executeJavaScript = async function (code, hasUserGesture) {
   await waitTillCanExecuteJavaScript(this);
-  return ipcMainUtils.invokeInWebContents(this, false, 'ELECTRON_INTERNAL_RENDERER_WEB_FRAME_METHOD', 'executeJavaScript', code, hasUserGesture);
+  return ipcMainUtils.invokeInWebContents(this, false, 'ELECTRON_INTERNAL_RENDERER_WEB_FRAME_METHOD', 'executeJavaScript', String(code), !!hasUserGesture);
 };
-WebContents.prototype.executeJavaScriptInIsolatedWorld = async function (code, hasUserGesture) {
+WebContents.prototype.executeJavaScriptInIsolatedWorld = async function (worldId, code, hasUserGesture) {
   await waitTillCanExecuteJavaScript(this);
-  return ipcMainUtils.invokeInWebContents(this, false, 'ELECTRON_INTERNAL_RENDERER_WEB_FRAME_METHOD', 'executeJavaScriptInIsolatedWorld', code, hasUserGesture);
+  return ipcMainUtils.invokeInWebContents(this, false, 'ELECTRON_INTERNAL_RENDERER_WEB_FRAME_METHOD', 'executeJavaScriptInIsolatedWorld', worldId, code, !!hasUserGesture);
```
miniak, line 214: `String(code)`

nornagon:
```suggestion
  return ipcMainUtils.invokeInWebContents(this, false, 'ELECTRON_INTERNAL_RENDERER_WEB_FRAME_METHOD', 'executeJavaScriptInIsolatedWorld', worldId, String(code), !!hasUserGesture);
```

nornagon: ah now i remember why i took this out. `code` isn't a string for this function.

nornagon:
> Is this testable somehow?

yeah, we can test that calling the function with weird argument types results in a more clear error than "Failed to serialize arguments"

nornagon:
Ah, we should also probably test that `hasUserGesture` works in both forms of `executeJavaScript`.

PR #24440, opened 2020-07-07
chore: don't duplicate devtools constants

codebytere:
#### Description of Change

We had duplicate devtools constants everywhere and already have access to those selfsame constants in `chrome/common/webui_url_constants.h`.

cc @MarshallOfSound @deepak1556 @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/electron_web_ui_controller_factory.cc
+++ b/shell/browser/electron_web_ui_controller_factory.cc
@@ -6,17 +6,12 @@
 
 #include <string>
 
+#include "chrome/common/webui_url_constants.h"
 #include "content/public/browser/web_contents.h"
 #include "electron/buildflags/buildflags.h"
 #include "shell/browser/ui/devtools_ui.h"
 
 namespace electron {
-
-namespace {
-
-const char kChromeUIDevToolsBundledHost[] = "devtools";
-
-}  // namespace
 
 // static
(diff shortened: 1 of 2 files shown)


shell/common/extensions/electron_extensions_api_provider.cc
```diff
@@ -32,7 +33,7 @@ namespace keys = manifest_keys;
 namespace errors = manifest_errors;
 
 constexpr APIPermissionInfo::InitInfo permissions_to_register[] = {
-    {APIPermission::kDevtools, "devtools",
+    {APIPermission::kDevtools, chrome::kChromeUIDevToolsHost,
```
deepak1556, line 36: "devtools" here is just name of permission, should not be replaced.

shell/renderer/electron_sandboxed_renderer_client.cc
```diff
@@ -32,7 +33,7 @@ const char kModuleCacheKey[] = "native-module-cache";
 
 bool IsDevTools(content::RenderFrame* render_frame) {
   return render_frame->GetWebFrame()->GetDocument().Url().ProtocolIs(
-      "devtools");
+      chrome::kChromeUIDevToolsHost);
```
deepak1556, line 36: we are checking the scheme here not host, its better to leave the raw version.

PR #24452, opened 2020-07-08
refactor: match upstream print preview handling

codebytere:
#### Description of Change

Printing refactor to more closely match upstream upstream. Once this is merged to the master roll we can cherry-pick it to 10-x-y

cc @deepak1556 

Notes: none


shell/browser/printing/print_preview_message_handler.cc
```diff
@@ -107,14 +112,45 @@ void PrintPreviewMessageHandler::OnMetafileReadyForPrinting(
   }
 }
 
+void PrintPreviewMessageHandler::OnPrepareForDocumentToPdfDone(
+    const PrintHostMsg_PreviewIds& ids,
+    printing::mojom::PrintCompositor::Status status) {
+  DCHECK_CURRENTLY_ON(BrowserThread::UI);
+  if (status != printing::mojom::PrintCompositor::Status::kSuccess) {
+    LOG(ERROR) << "Preparing document for pdf failed with error " << status;
```
deepak1556, line 120: Will `OnMetafileReadyForPrinting` be called even when this fails ? want to make sure we don't have an unresolved promise.

codebytere: yes - this is always triggered from local testing. if we try and reject before this, in page composition failure, it results in duplicate rejections

deepak1556:
Good to see a patch go away.

codebytere:
merging as the failure:
> AssertionError: is not alwaysOnTop: expected false to be true

is unrelated.

PR #24459, opened 2020-07-08
build: update ts defs generator

MarshallOfSound:
Notes: no-notes


codebytere:
> contentTracing stopRecording creates a temporary file when an empty string is passed

is unrelated - merging.

Issue #24260, opened 2020-06-23
Add a possibility to create a sealed app


antelle:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description
Imagine there's an app built on Electron. Using flags and environment variables provided by Electron, it's possible to use this app to run any code, not only what was provided by the app developer. For example:

```
ELECTRON_RUN_AS_NODE=1 /Applications/Visual\ Studio\ Code.app/Contents/MacOS/Electron
```

Now my script has a possibility to act as if it was Visual Studio Code.

This poses several security risks:

* any other app can use the identity of our app, for example, for:
  * reading keychain items using node-keytar module or similar, which must be limited to only one app by design; 
  * hiding network activity behind our app and, since nowadays there are a lot of apps build on Electron, it can be an easy way to bypass any firewall; 
  * running untrusted code as if it was code signed; there's no need to sign the code, just find another electron app, take its signed binary, create a launcher, and use it; 
* it's very easy to run a modified/patched version the app itself by running some script before the app has a chance to launch its first bit of code; app developers can't be sure if their app has not been modified, which effectively defeats the purpose of code signing, this can be used for:
  * creating patches and cracks for Electron apps; 
  * building modified versions of the app with malicious purposes without redistributing a patched binary.

There can be other applications. Overall, this way developers don't have control over code that is executed by the app signed and distributed with their name.

### Proposed Solution
I would be happy to see a "hardened" build with these features disabled in production:
- `NODE_OPTIONS`
- `ELECTRON_RUN_AS_NODE `
- possibility to provide a path to script that will be executed first

The list may be incomplete and, if the feature is accepted, this should be reviewed by Electron team members, who have more knowledge about possible ways of changing the entry point.

### Alternatives Considered
- build Electron from source with mentioned flags disabled (there are options for that): this is complicated, requires setting up CI, which is a bit hard for an average open-source project
- let the app decide if overriding the entry point is possible, for example, if the app is requested to run as node, the entry point must be a script in app resources, which can abort the execution or make another decision based on developer's needs.

MarshallOfSound:
[...]

Issue #2570, opened 2015-08-24
Is the source really protected with building apps with Electron?


ashking:
Hey guys, 

This is not entirely a bug. 

I was wondering; are the apps built with Electron protects the source? I tried JxCore and it does not protect source. How about electron? 

Cheers

zcbenz:
No, it is very easy to get your source code even when you packaged them in `asar` archive, the V8 JavaScript engine is never designed to hide source code. NW.js does succeed to hide one file of the source code completely, but with the price of significant performance downgrade.

So if you really to want to protect your source code, you have to write them in C++ and make a native Node module.

etiktin:
There are some options, but the level of protection they provide is limited by the reasons @zcbenz mentioned:
- You could implement some decryption logic on your own, in a C++ node addon, but it will probably still be breakable (e.g. debug the program and copy the source after decryption but before eval).
- There's also [EncloseJS](http://enclosejs.com/), that claims to compile your code for node.js/io.js, so it might work with Electron.
- [nw.js](https://github.com/nwjs/nw.js), which is a similar project, has support for v8 snapshots which provide some kind of protection (see details [here](https://github.com/nwjs/nw.js/wiki/Protect-JavaScript-source-code-with-v8-snapshot)).

Note that all of this methods, will incur some performance penalty.

fritx:
@etiktin I'm running into this too.
Is it possible to just apply node.js code in C++ addon directly?

etiktin:
@fritx, Node has a module called vm which can eval code. I guess you could find a way to call it or it's underlining C++ API from your addon and use that to load your code. 
Keep in mind that the user could still manage to open the developer tools and see the loaded script. Even if you manage to block that using a custom Electron build and the addon checking to see that your Electron build wasn't tampered with, the user could still see your code as a string in memory.

fizzvr:
Sounds good protect the source! :key:

alexis57:
Then I have a question, how closed code electron app such as GitKraken, Whatsapp or Slack can hide the code ?

Hum4n01d:
@alexis57 I have the same question; maybe it’s just that inspect element is disabled or something

alexis57:
Or maybe the main code comes from a C/C++ library and then they use electron just for the GUI, don't know...

nabtron:
Would like to know how whatsapp and slack protect their source code too

boeserwolf:
[...]

PR #24470, opened 2020-07-08
refactor: mmap asar files

nornagon:
#### Description of Change
In preparation for being able to load asar files from a dll/dylib, which are implicitly memory mapped.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


deepak1556:
First of all, interesting exploration 👍 

QQ: wouldn't this increase the memory footprint of the main process considering the entire asar file is being kept in memory ?

nornagon:
> QQ: wouldn't this increase the memory footprint of the main process considering the entire asar archive is being kept in memory ?

This isn't how mmap works. The OS can page out chunks of the mmap'd file.

nornagon:
Some relevant commentary from Sublime: https://www.sublimetext.com/blog/articles/use-mmap-with-care

deepak1556:
thanks for the context, follow-up question why does `base::MemoryMappedFile` provide to load a specific region of file into mmap https://source.chromium.org/chromium/chromium/src/+/master:base/files/memory_mapped_file.h;l=102 and there are some uses of that in the codebase, what is the advantage over loading the whole file ? would it be useful for asar archive which can be separated to regions.

nornagon:
Not 100% sure, but I think the main reason to map a subregion is to limit usage of virtual address space, which is important mainly on 32-bit systems. It could also be a convenience method to avoid having to pass around an offset everywhere along with the mmap'd file.

MarshallOfSound:
@nornagon Have you performance tested this at all? Can we see if there's an increase / decrease in perf of reading a large amount of small files sequentially.  Or reading large files.  etc.

deepak1556:
Thinking a bit more, can you run this experiment behind a `base::FeatureList` flag, seems like a good candidate.

nornagon:
> Thinking a bit more, can you run this experiment behind a `base::FeatureList` flag, seems like a good candidate.

Hm, I'm not sure I understand why this would qualify for that. There's no change in functionality.

deepak1556:
> In preparation for being able to load asar files from a dll/dylib

wouldn't this require anymore implementation changes, having both code paths then would help in comparison as well as a way to opt in.

> There's no change in functionality.

Hmm but the underlying implementation has changed, this is similar to blink enabling its new rendering engine `LayoutNG` which had no user facing changes but was behind feature flag to experiment any regressions in perf or other stuff.

nornagon:
> > In preparation for being able to load asar files from a dll/dylib
> 
> wouldn't this require anymore implementation changes, having both code paths then would help in comparison as well as a way to opt in.

You'll be able to opt in to loading from a dylib by putting your app in a dylib and enabling the relevant fuse. That part will always remain optional, and no feature flag will be necessary.

> > There's no change in functionality.
> 
> Hmm but the underlying implementation has changed, this is similar to blink enabling its new rendering engine `LayoutNG` which had no user facing changes but was behind feature flag to experiment any regressions in perf or other stuff.

LayoutNG was a major rewrite of a massive and extraordinarily complex system. This is not. I think the risk is substantially and meaningfully lower here.

I don't think we should ask that all refactors in Electron come with a feature flag and a deprecation cycle. If there are concrete reasons to suspect that this may cause issues, then I'd be happy to test & address them prior to merging. Otherwise, let's treat this as any other refactor: roll it out, watch for reports of breakage, and fix as necessary.

The only thing I'm cautious about is potential issues when loading from a network drive that goes away (as mentioned in the sublime post). Previously, that would have resulted in an I/O failure (with unknown effects on the app; it would be up to the app to handle it). After this change, I think it will result in SIGBUS.

There might be other issues when using mmap as opposed to fs.read(), but we won't find them until we enable this by default anyway, so putting it behind a flag will just delay the gathering of such information and force us to maintain both implementations in parallel for longer.

I think flags are best suited for large, relatively isolated functionality which app developers have a reason to want to opt into (e.g. it enables a feature for them, so they'll switch on the flag in order to use it). The value of this refactor by itself isn't obvious, so I don't see any reason a developer would choose to enable it.

shell/common/api/electron_api_asar.cc
```diff
@@ -103,15 +107,67 @@ class Archive : public gin_helper::Wrappable<Archive> {
...
   }
 
  private:
-  std::unique_ptr<asar::Archive> archive_;
+  static std::unique_ptr<v8::BackingStore> ReadOnIO(
+      v8::Isolate* isolate,
+      std::shared_ptr<asar::Archive> archive,
+      uint64_t offset,
+      uint64_t length) {
+    auto backing_store = v8::ArrayBuffer::NewBackingStore(isolate, length);
```
zcbenz, line 155: This API is not marked as thread safe, and [its implementation](https://source.chromium.org/chromium/chromium/src/+/master:v8/src/objects/backing-store.cc;drc=65c53735b58c961c225a878df11c841a6614d6ab;l=219?originalUrl=https:%2F%2Fcs.chromium.org%2F) does access things from `isolate`. I think the `backing_store` should be created in UI thread and then passed to IO.

shell/common/api/electron_api_asar.cc
```diff
@@ -103,15 +107,67 @@ class Archive : public gin_helper::Wrappable<Archive> {
...
+                              uint64_t offset,
+                              uint64_t length) {
+    gin_helper::Promise<v8::Local<v8::ArrayBuffer>> promise(thrower.isolate());
+    v8::Local<v8::Promise> handle = promise.GetHandle();
+
+    base::CheckedNumeric<uint64_t> safe_offset(offset);
+    base::CheckedNumeric<uint64_t> safe_end = safe_offset + length;
+    if (!safe_end.IsValid() ||
+        safe_end.ValueOrDie() > archive_->file()->length()) {
+      thrower.ThrowError("Out of bounds read requested in ASAR");
```
zcbenz, line 136: This error should be passed in promise otherwise it would not be captured user's callback.

lib/common/asar.js
```diff
@@ -562,13 +557,10 @@
...
 
       const { encoding } = options;
-      const buffer = Buffer.alloc(info.size);
-      const fd = archive.getFd();
-      if (!(fd >= 0)) throw createError(AsarError.NOT_FOUND, { asarPath, filePath });
 
       logASARAccess(asarPath, filePath, info.offset);
-      fs.readSync(fd, buffer, 0, info.size, info.offset);
-      return (encoding) ? buffer.toString(encoding) : buffer;
+      const buffer = Buffer.from(archive.readSync(info.offset, info.size));
```
zcbenz, line 562: The `Buffer.from` copies data from the argument, can the method return `Buffer` directly to save the extra copy?

shell/common/api/electron_api_asar.cc
```diff
@@ -103,15 +107,67 @@ class Archive : public gin_helper::Wrappable<Archive> {
...
-      return -1;
-    return archive_->GetFD();
+  v8::Local<v8::ArrayBuffer> ReadSync(gin_helper::ErrorThrower thrower,
+                                      uint64_t offset,
+                                      uint64_t length) {
+    base::CheckedNumeric<uint64_t> safe_offset(offset);
+    base::CheckedNumeric<uint64_t> safe_end = safe_offset + length;
+    if (!safe_end.IsValid() ||
+        safe_end.ValueOrDie() > archive_->file()->length()) {
+      thrower.ThrowError("Out of bounds read requested in ASAR");
```
zcbenz, line 117: The error should have an error code, and include file path and asar path in description, like the other asar errors.

zcbenz:
> In preparation for being able to load asar files from a dll/dylib, which are implicitly memory mapped.

I'm not sure what does loading asar files from dll means, can you explain a bit more about what it is and when it would be used?

Mmap is very useful for editors when editing very large files, where app has to constantly read parts of files. But in Electron apps most files in asar archive are only read for once, I'm not sure if there would be benefits on performance. So I'm a bit hesitate with a refactoring that brings little benefits and increases stability risks. It would be great if you can explain more about the new features it would enable, and probably do some benchmarks on performance.

lib/common/asar.js
nornagon: It does not copy in this case: https://github.com/nodejs/node/blob/090ad7d4a1c32ed6fcd861ce215b1205afe4146c/lib/buffer.js#L293

nornagon:
> I'm not sure what does loading asar files from dll means, can you explain a bit more about what it is and when it would be used?

The idea is to allow protecting script sources with codesigning. Currently, .asar files are not considered "code" by codesigning verification systems in macOS and Windows:

- macOS's Gatekeeper system validates non-code resources at first launch, but not thereafter. However, enabling the hardened runtime on a binary can prevent it from loading any unsigned dylibs, regardless of whether it's "first launch" or not.
  - The level of protection that this would add is debatable. If an attacker can modify files on disk, they can just as easily remove the code signature and hardened runtime flags from the main executable as edit the .asar. In current versions of macOS, unsigned binaries run without any user prompt. There are, however, some minor restrictions that apply to unsigned apps: for instance, unsigned apps cannot access the keychain.
  - I've heard that on modern macOS, Gatekeeper does "periodic" checks in addition to first-launch checks. I'm not sure what triggers this.
  - There have been some signs from Apple that in future versions of macOS, unsigned code will not run by default.
- On Windows, AppLocker can restrict which publishers' signatures are respected for all loaded DLLs. It has no system for requiring that non-DLL files are signed, though. Putting resources in a DLL on Windows would allow AppLocker to apply its restrictions to those resources.
- I don't know enough about codesigning on Linux to comment as to whether this approach would be relevant on that platform. It seems there are several systems and ongoing development. e.g. [IMA appraisal](https://lwn.net/Articles/733431/).

The goal here is to allow Electron apps to take advantage of the resource-integrity systems provided by the underlying operating system. Native apps have access to this by default, but given that the code of Electron apps is loaded dynamically from a non-"code" file, these systems do not effectively apply to Electron apps currently.

My plan for supporting this is roughly:
1. Introduce the ability to load ASAR files from a new kind of bundle file, `.asar.dll` (or `.asar.dylib` on macOS). These would be loaded with `LoadLibrary` / `dlopen`, and would expose a two well-known symbols: a pointer to a byte sequence, and the size of that byte sequence.
2. Add `app.asar.dll` to the front of the ["path search"](https://github.com/electron/electron/blob/6b1de271a7cba4ae18e3bbc95afa98627a823030/lib/browser/init.ts#L109) code when searching for app code.
3. Add a fuse which turns off the path search code and sets it to only ever load from `app.asar.dll`.

The ultimate goal is to provide apps with tools that allow them to do just as well as the OS will permit them to in attempting to prevent unvalidated code from running.

lib/common/asar.js
zcbenz: Thanks for the reference, I confirm the content is not copied in this case.

So Node's document of `Buffer.from` is wrong, but it probably does not matter except for low-level library writers.

zcbenz:
Thank for the clarification, code signing asar files would be very exciting and I think it is a fair use case. 👍

shell/common/api/electron_api_asar.cc
nornagon: The ArrayBuffer allocator is noted to be thread-safe: https://source.chromium.org/chromium/chromium/src/+/master:v8/include/v8.h;l=5041;drc=76b53b66b0a441494748b82250b7d3ea3f1ef6fe. Additionally, SharedArrayBuffer::NewBackingStore is surely thread-safe, as that is the entire reason for the existence of SharedArrayBuffer, and it [calls the same allocation function](https://source.chromium.org/chromium/chromium/src/+/master:v8/src/api/api.cc;l=7890;drc=76b53b66b0a441494748b82250b7d3ea3f1ef6fe).

shell/common/api/electron_api_asar.cc
nornagon: Hm, interesting point. When using `async/await` either are captured by `try/catch`, the only difference is when using `.catch()` (or the 2-arg `.then()`). It would also make a difference if not `await`ing the result of such a promise.

e.g.


```js
throwsErrorsSynchronously(badArguments)
  .then(() => { /* ... */ })
  .catch((err) => { /* never reached */ })
// never reached:
doOtherThings().then(/* ... */)
```

vs. throwing errors by rejecting the promise:

```js
throwsErrorsAsynchronously(badArguments)
  .then(() => { /* ... */ })
  .catch((err) => { /* error is thrown here */ })
doOtherThings().then(/* ... this will end up happening ... */)
```


It looks like when you write an `async function`, errors are always thrown by rejecting the promise rather than synchronously thrown, even if the error happens synchronously:


```js
async function errorThrower() { throw new Error('no anchovies unless otherwise specified') }
errorThrower()
  .catch((err) => { /* error appears here */ })
thisDoesGetCalled()
```

So I'll go with the way you suggest.

nornagon:
@zcbenz please take another look.

shell/common/api/electron_api_asar.cc
zcbenz: Yeah the allocator is thread-safe and it is safe to share the backing stores between threads, but I'm not sure if the `SharedArrayBuffer::NewBackingStore` API is thread-safe.

From my understanding of V8's threading model, the isolate should not be used in multiple threads without locks. In Node.js, the shared array buffers (and their backing stores) are always created in the same thread with their isolates, and only then their backing stores can be passed to other threads. It is similar to how we can write to a `node::Buffer` in a new thread, while still having to create it in the thread of the isolate.

Also in the [the implementation of BackingStore::Allocate](https://source.chromium.org/chromium/chromium/src/+/master:v8/src/objects/backing-store.cc;drc=65c53735b58c961c225a878df11c841a6614d6ab;l=219?originalUrl=https:%2F%2Fcs.chromium.org%2F), it has a call of `isolate->heap()->AllocateExternalBackingStore`, which could invoke V8's garbage collection. It should not be a thread-safe API.

nornagon: makes sense, thanks for the info. updated.

nornagon:
all 4 (!) ci failures are unrelated; merging.

gpetrov:
@nornagon packaging asar files in dlls/gylibs and code signing them will be really an amazing feature! Can't wait till it is available! 

I presume that the asar.dll files will be loaded by Electron only when the code signature checks out and refused otherwise.

Loading from a memory mapped files, should really give also a welcome speed boost as well, specially when dealing with tons of small node_modules files.

Also another welcome addition will be also to be able to encrypt the asar files, so next to code signing also code protection is available. When packing an asar archive we can already specify a stream transform, now we just need to at the loading as well.

gpetrov:
A small but powerful addition will be also the just un-gzip asar dll files runtime on load. This will save storage as well and make asar files much smaller.

Also if the asar memory can be directly smartly mapped with the node_modules cache, so run time memory is saved. You don't want to load a 100MB asar file with node_modules, that will take then twice the memory because of the asar mmap file self and all the require modules cache... but that might be quite a technical challenge.

nornagon:
> I presume that the asar.dll files will be loaded by Electron only when the code signature checks out and refused otherwise.

This is up to the OS. On Windows, code signatures are not checked by default, but there are some 3rd-party tools like AppLocker which can be employed to harden the system.

> Loading from a memory mapped files, should really give also a welcome speed boost as well, specially when dealing with tons of small node_modules files.

There is no appreciable speed difference between reading from an asar and reading from a memory-mapped file, as far as I can tell.

> encrypt the asar files

We have no plans to add encryption of any sort.

> un-gzip asar dll files runtime on load

I don't believe this is useful to implement in Electron. Enable NTFS compression if you want this.

> directly smartly mapped with the node_modules cache

I'm not sure what you mean by this.

gpetrov:
> This is up to the OS. On Windows, code signatures are not checked by default, but there are some 3rd-party tools like AppLocker which can be employed to harden the system.

Well we can't really rely on the OS as we don't have control of it. Specially on windows indeed there is no code signing validation per default or it can be easily switched off.

So loading and validating the app and its asar file is really a responsibility of the app self. It has being asked before many times: https://github.com/electron/electron/issues/24260 and encryption just for that as well https://github.com/electron/electron/issues/24681 https://github.com/electron/electron/issues/2570

maybe something like the content verification with own public key, like in NWJS:
https://nwjs.readthedocs.io/en/latest/For%20Users/Advanced/Content%20Verification/

but then combined with Electron fuses @MarshallOfSound  or other more easier way to embed the public key in the electron binary without the need to recompile it and also have a way from the asar content to verify back that the main binary is also not tempered and bot use the same code signing key.

> _un-gzip asar dll files runtime on load_ 
> I don't believe this is useful to implement in Electron. Enable NTFS compression if you want this.

Again we don't have control of the user system, so can just let them enable NTFS compression. Just like loading files from a website the browser accepts gzip and unzip on the fly - this should be also possible when loading asar files. NWJS also supports having zip file as a content package to load. But unfortunately it unzips it fully first and is not that smart as asar to do that on the fly.

> _directly smartly mapped with the node_modules cache_ > I'm not sure what you mean by this.

well the node modules code resides in the asar file, so when the asar file is loaded, it will get in memory (well maybe not completely depending on the mmap implementation) but then when a node module is required, the require builds a global cache in memory of all required modules, so those don't have to be loaded again. So you can actually the code of the same module twice - once in the asar memory file and once in the require cache https://nodejs.org/api/modules.html#modules_caching

fabiospampinato:
> > un-gzip asar dll files runtime on load

> I don't believe this is useful to implement in Electron. Enable NTFS compression if you want this.

@nornagon Could you expand on why you think this wouldn't be useful? One complaint I hear fairly often from people is that Electron apps are just big, if there's something that can deliver double-digits improvements in this area I don't think Electron should just dismiss the opportunity like that. 

Also I don't think it's reasonable to say to people "Yeah we could improve on this, but just switch to a filesystem that supports compressed partitions", that's obviously not going to work, people don't even know what a filesystem is.

I have opened an issue about this [here](https://github.com/electron/electron/issues/29556). IMHO working with compressed asar archives may speed things up too, since the bottleneck for almost all machine is going to be reading from disk rather than decompressing.

Other than the potential difficulties of switching to compressed asar archives it seems to me like the potential pros clearly outweigh the potential cons 🤷‍♂️

Issue #16417, opened 2019-01-16
Vibrancy should not default to active state


sindresorhus:
* Output of `node_modules/.bin/electron --version`: v4.0.1
* Operating System (Platform and Version): macOS 10.14.2

**Expected Behavior**

The default behavior of native macOS apps with vibrancy is to disable the vibrancy when the window is inactive. This makes it easier for the user to see the active app.

**Actual behavior**

Electron apps have the vibrancy forced to always be active since Electron manually sets the state to [`NSVisualEffectStateActive`](https://developer.apple.com/documentation/appkit/nsvisualeffectstate/nsvisualeffectstateactive?language=objc): https://github.com/electron/electron/blob/52fe92d02ef94712fc7b19b3f03d3363cfa71a4d/atom/browser/native_window_mac.mm#L1272

Electron should instead default to `NSVisualEffectStateFollowsWindowActiveState`, which is the default.

Electron could let the user optionally override this behavior through `BrowserWindow({vibrancyState: 'active'})` and `win.setVibrancy('sidebar', {state: 'active'})`.

**To Reproduce**

Nothing to reproduce. The docs are clear on this one.

**Additional Information**

More info here: https://github.com/sindresorhus/caprine/issues/701

electron-triage:
Thank you for taking the time to report this issue and helping to make Electron better.

The version of Electron you reported this on has been superseded by newer releases. 

If you're still experiencing this issue in Electron 6.x.y or later, please add a comment specifying the version you're testing with and any other new information that a maintainer trying to reproduce the issue should know.

I'm setting the `blocked/need-info` label for the above reasons. This issue will be closed 7 days from now if there is no response.

Thanks in advance! Your help is appreciated.

sindresorhus:
Still an issue in `master`: https://github.com/electron/electron/blob/fe5318d753637c3903e23fc1ed1b263025887b6a/shell/browser/native_window_mac.mm#L1440

flawnn:
@codebytere I wasn't sure whether to open a new issue but this issue hasn't been implemented like described, there is still an option missing to change the vibrancyState over the runtime :)

PR #24471, opened 2020-07-09
fix: use default NSVisualEffectState enum case

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/16417.

In [Apple documentation](https://developer.apple.com/documentation/appkit/nsvisualeffectstate/nsvisualeffectstatefollowswindowactivestate?language=objc) `NSVisualEffectState` is an enum with three options detailing how the material appearance should reflect window activity state. We were previously using `NSVisualEffectStateActive`, meaning that the backdrop would always appear active. This breaks from the default as specified in the first link, which is `NSVisualEffectStateFollowsWindowActiveState` and means that the backdrop should automatically appear active when the window is active, and inactive when it is not. This thus changes our effect state to the more expected default.

cc @miniak @zcbenz @jkleinsc 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where macOS window vibrancy active state did not always match the active state of the window.

### DIFF
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -1451,7 +1451,7 @@ void NativeWindowMac::SetVibrancy(const std::string& type) {
 
     [effect_view setAutoresizingMask:NSViewWidthSizable | NSViewHeightSizable];
     [effect_view setBlendingMode:NSVisualEffectBlendingModeBehindWindow];
-    [effect_view setState:NSVisualEffectStateActive];
+    [effect_view setState:NSVisualEffectStateFollowsWindowActiveState];
 
     // Make frameless Vibrant windows have rounded corners.
     if (!has_frame() && !is_modal()) {


MarshallOfSound:
> Notes: Fixed an issue where the effect state on macOS was not using the appropriate default for NSVisualEffectState.

I had to do some googling to figure this out, can we phrase this better in a way Electron devs would understand what changed.  i.e. does this only impacts folks using vibrancy?

codebytere:
@MarshallOfSound yes - it's invoked in `SetVibrancy`. Honestly not sure how to distill this down any better, since it is a detail that's somewhat specific to macOS-specific APIs.

MarshallOfSound:
I assume this is a visual change, what will users see differently now?

codebytere:
@MarshallOfSound essentially - the vibrancy active state did not match that of the window. The default behavior of native macOS apps with vibrancy is to disable the vibrancy when the window is inactive, which makes it easier for the user to see the active app. Before this fix we forced it to be active all the time, which was visually confusing

Issue #24456, opened 2020-07-08
Printing hangs with pageSize Object


TheSinding:
* **Electron Version:**
v9.0.1
* **Operating System:**
Linux 5.7.4-arch1-1
### Expected Behavior
I expect it to print.

### Actual Behavior
The printing hangs.

### To Reproduce
```javascript
const { remote } = require ('electron');

const { BrowserWindow } = remote;
const win = new BrowserWindow();

win.loadURL('https://google.com');

win.webContents.on('did-finish-load', () => {
  win.webContents.print({ pageSize: { height: 88.6, width: 35.7 } }, (success, error) => {
    if(!success) console.error(error);
  })
})
```
Here is a fiddle of the error `https://gist.github.com/81597c2a1f33779c7083194b4590f55c`

codebytere:
@TheSinding i see the error - i'll try to get this patched up soon :)

TheSinding:
@codebytere - Arh okay that's great.
It would be great it the documentation were more detailed.

Right now it looks like this:
`pageSize` String | Size (optional) - Specify page size of the printed document. Can be `A3`, `A4`, `A5`, `Legal`, `Letter`, `Tabloid` or an Object containing `height`.

It should properly look like this.
`pageSize` String | Size (optional) - Specify page size of the printed document. Can be `A3`, `A4`, `A5`, `Legal`, `Letter`, `Tabloid` or an Object containing `height` and `width` measured in **microns**.


Second:
I would be awesome if I could enter the actual paper name in the PPD driver as well. I my case I'm trying to print with a DYMO LabelWriter 450, which uses the `30321 Large Address` in the driver, but I can't specify it in the `pageSize`. 

I don't know if that would be useful to anyone other. :)

TheSinding:
This GIST still wont work in my system, is it my setup ? 
`https://gist.github.com/81597c2a1f33779c7083194b4590f55c`

codebytere:
@TheSinding what does "won't work" mean in this context? are you seeing errors?

TheSinding:
There is not error or any output, it just hangs until I close the BrowserWindow.

PR #24476, opened 2020-07-09
fix: always callback error with invalid print settings

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/24456.

The minimum micron size Chromium accepts is that where (per `printing/units.h`):
* kMicronsPerInch - Length of an inch in 0.001mm unit.
* kPointsPerInch - Length of an inch in CSS's 1pt unit.

Adhere to: `(kPointsPerInch / kMicronsPerInch) * size >= 1`

Practically, this means microns need to be > 352 microns.

We therefore need to verify this or it will fail with a logging error of:

```sh
[77469:0708/232856.127937:ERROR:print_view_manager_base.cc(71)] Invalid printer settings The selected printer is not available or not installed correctly. <br> Check your printer or try selecting another printer.
```

This also adds in a callback call in the event that print setting validation fails, so there are no silent failures.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where `webContents.print()` would sometimes hang with invalid settings.

### DIFF
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -1334,13 +1334,12 @@ win.webContents.print(options, (success, errorType) => {
   * `landscape` Boolean (optional) - `true` for landscape, `false` for portrait.
   * `marginsType` Integer (optional) - Specifies the type of margins to use. Uses 0 for
     default margin, 1 for no margin, and 2 for minimum margin.
-    and `width` in microns.
   * `scaleFactor` Number (optional) - The scale factor of the web page. Can range from 0 to 100.
   * `pageRanges` Record<string, number> (optional) - The page range to print.
     * `from` Number - the first page to print.
     * `to` Number - the last page to print (inclusive).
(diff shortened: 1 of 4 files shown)


codebytere:
@jkleinsc unfortunately we cannot, for a few reasons:
* Printing infrastructure requires that a given system has a default printer configured, and without that will throw an error to that effect
* Printing dialogs are process blocking, and so we don't test for successful dialog triggering elsewhere for the same reasons we can't currently test any other dialogs except in that they throw appropriately for bad user input

I tested locally on macOS to ensure that the dialog was appropriately triggered and can try to do so on my Linux machine as well, but i think this is the best we can do for now 😕

jkleinsc:
> I tested locally on macOS to ensure that the dialog was appropriately triggered and can try to do so on my Linux machine as well, but i think this is the best we can do for now 😕

@codebytere did you test with small values like 
```
        w.webContents.print({
          pageSize: {
            width: 352,
            height: 352
          }
        });
```

PR #24479, opened 2020-07-09
build: update to standard 14

MarshallOfSound:
Does what is says on the tin, updates our standard deps (standard-eslint-config + standard-markdown) to standard 14. The rest of the changes are automated as the result of `--fix`

Notes: no-notes


MarshallOfSound:
@codebytere We don't use prettier at all in this repo, we use standard as the base for everything.  Once all these dep bumps have landed I can look at pulling in prettier

codebytere:
Oh i assumed we swapped when you enabled semicolons - if that's not the case we should at least use the same rules in docs that we do in internal code

Issue #24451, opened 2020-07-07
Missing "visible-on-all-workspaces-changed" event


fabiospampinato:
### Problem Description

The internal states that apps keep of whether each window is visible on all workspaces or not can't be kept in sync with the actual state of the window (assuming polling is not acceptable) because there currently doesn't seem to be a way to detect when the "visibleOnAllWorkspaces" property changes from outside the app, e.g. like by manually toggling this from macOS' dock.

### Proposed Solution

There should be a "visible-on-all-workspaces-changed" event, like there is a "always-on-top-changed" event.

### Alternatives Considered

Polling? But it's unclear whether the `BrowserWindow#visibleOnAllWorkspaces` is even kept synchronized with external changes, and polling would be an unreasonable strategy for this anyway.

codebytere:
Seems reasonable. I'll try to add it sometime soon.

codebytere:
See PR for why this ended up being closed - apologies!

PR #24485, opened 2020-07-09
feat: add visible-on-all-workspaces-changed event

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/24451.

At present, there is not a way to detect when the "visibleOnAllWorkspaces" property changes from outside the app, which are possible by e.g. toggling from the macOS dock. A developer could attempt to poll for changes, but that's somewhat unwieldy when we could instead add an event to more effectively notify developers of changes to this window property.

cc @zcbenz @miniak @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added a` visible-on-all-workspaces-changed` event to poll for changes to window workspace visibility.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -580,6 +580,15 @@ Returns:
 
 Emitted when the window is set or unset to show always on top of other windows.
 
+#### Event: 'visible-on-all-workspaces-changed'
+
+Returns:
+
+* `event` Event
+* `isVisibleOnAllWorkspaces` Boolean
+
+Emitted when a window is set or unset to be visible on all workspaces.
+
 #### Event: 'app-command' _Windows_ _Linux_
 
 Returns:
--- a/shell/browser/api/electron_api_base_window.cc
+++ b/shell/browser/api/electron_api_base_window.cc
@@ -280,6 +280,10 @@ void BaseWindow::OnWindowAlwaysOnTopChanged() {
   Emit("always-on-top-changed", IsAlwaysOnTop());
 }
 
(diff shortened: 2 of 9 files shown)


docs/api/browser-window.md
```diff
@@ -580,6 +580,15 @@ Returns:
 
 Emitted when the window is set or unset to show always on top of other windows.
 
+#### Event: 'visible-on-all-workspaces-changed'
```
ckerr, line 583: Since the tests run iff platform != win32, does this line need `_macOS_ _Linux_` ?

jkleinsc:
The API WG approved this PR at our July 13th 2020 meeting

docs/api/browser-window.md
codebytere: no - this runs on all platforms but i believe CI makes this unreliable :)

codebytere:
@zcbenz i'm not 100% sure that's currently possible - i don't know that there's a notification that is posted when this changes by the system 🤔

zcbenz:
I could not find a system notification for it neither, but I think that's the whole point of adding this new API, as it was mentioned in #24451:

> there currently doesn't seem to be a way to detect when the "visibleOnAllWorkspaces" property changes from outside the app, e.g. like by manually toggling this from macOS' dock.

codebytere:
@zcbenz i think i see a path forward! will update this soon

codebytere:
Closing this as from an API perspective it turns out that the “assign to all desktops” option is not 1:1 with “can join all spaces”. As such, observing `collectionBehavior` on NSwindow does not trigger a new event when the dock setting is toggled.

PR #24489, opened 2020-07-09
fix: default to simplex for printing DuplexMode

codebytere:
#### Description of Change

Linux requires that `duplex_mode` have a default, which is `kSimplex`. We need to set that before attempting to convert so that it doesn't fall down into this [`NOTREACHED()`](https://source.chromium.org/chromium/chromium/src/+/master:ui/gtk/printing/print_dialog_gtk.cc;l=276?q=print_dialog_gtk.cc&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2Fsearch%2F) DCHECK failure.

cc @jkleinsc @MarshallOfSound @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed potentially invalid duplex mode settings on Linux.

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -2025,7 +2025,8 @@ void WebContents::Print(gin_helper::Arguments* args) {
   }
 
   // Duplex type user wants to use.
-  printing::mojom::DuplexMode duplex_mode;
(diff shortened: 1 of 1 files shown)


jkleinsc:
I verified that the dcheck no longer happens on Linux if duplexmode is not specified

PR #24492, opened 2020-07-09
build: update octicons

MarshallOfSound:
Next in line for dependency updates is octicons, some icons were renamed / redesigned so I found the new equivs.

Notes: no-notes


Erraach-walid:
------

PR #24495, opened 2020-07-10
build: tsify asar and move to webpack js2c pipeline

MarshallOfSound:
Converts the asar init scripts to typescript and move them to be processed by our webpack --> js2c pipeline.

Notes: no-notes


BUILD.gn
```diff
@@ -139,36 +148,26 @@ webpack_build("electron_isolated_renderer_bundle") {
...
+    ":electron_asar_bundle",
     ":electron_browser_bundle",
     ":electron_isolated_renderer_bundle",
-    ":electron_js2c_copy",
     ":electron_renderer_bundle",
     ":electron_sandboxed_renderer_bundle",
     ":electron_worker_bundle",
   ]
 
   webpack_sources = [
```
nornagon, line 161: i think this intermediate variable is no longer useful

lib/asar/fs-wrapper.ts
```diff
@@ -0,0 +1,779 @@
...
+    gid,
+    0, // rdev
+    undefined, // blksize
+    ++nextInode, // ino
+    stats.size,
+    undefined, // blocks,
+    msec(stats.atime), // atim_msec
+    msec(stats.mtime), // mtim_msec
+    msec(stats.ctime), // ctim_msec
+    msec(stats.birthtime) // birthtim_msec
```
nornagon, line 91: `stats.atime` and friends are never actually set, so this always returns fakeTime

lib/asar/fs-wrapper.ts
```diff
@@ -0,0 +1,779 @@
+import { Buffer } from 'buffer';
+import * as path from 'path';
+import * as util from 'util';
+
+const asar = process._linkedBinding('electron_common_asar');
```
nornagon, line 5: should probably type this binding

lib/asar/fs-wrapper.ts
```diff
@@ -0,0 +1,779 @@
...
+// Convert asar archive's Stats object to fs's Stats object.
+let nextInode = 0;
+
+const uid = process.getuid != null ? process.getuid() : 0;
+const gid = process.getgid != null ? process.getgid() : 0;
+
+const fakeTime = new Date();
+const msec = (date: Date) => (date || fakeTime).getTime();
+
+const asarStatsToFsStats = function (stats: any) {
```
codebytere, line 64: we can probably type this, no? We know what the object's properties are: https://github.com/electron/electron/blob/602913cb4c98d102548e082c97401415849fe082/shell/common/api/electron_api_asar.cc#L71-L75

lib/asar/fs-wrapper.ts
MarshallOfSound: Trying to avoid changing implementation during ts stuff, also why I had to type a few things as `any`.  Turns our there's a lot of inaccuracies in this file.  They can all be addressed individually once this has landed.

MarshallOfSound:
Updated with strongly typed bindings

PR #24512, opened 2020-07-10
build: ensure that electron/lib/browser can only use browser or common imports

MarshallOfSound:
I noticed that some of our bundles incorrectly end up resolving files from a bad place.  E.g. `worker` and `renderer` depend on a `browser` TS file.  Before fixing that I'm introducing linting rules to ensure that we don't do this by accident / anywhere else too.

Only done the `lib/browser` folder first:

* You can't import from `electron` anymore, you have to import from `electron/main` or `electron/common`
* Relative imports are disabled, you need to use the `@electron/internal/...` path
* To go with the above in the browser folder you can only import from `internal/browser` and `internal/common`

I plan to do the same thing for the other `lib` folders, fixing issues as they're found.

This also has the nice side effect of using the new `electron/{main|common|renderer}` imports which introduce process-level type safety which is neat. 

Notes: no-notes

### DIFF
--- a/build/webpack/webpack.config.base.js
+++ b/build/webpack/webpack.config.base.js
@@ -90,7 +90,7 @@ module.exports = ({
     resolve: {
       alias: {
         '@electron/internal': path.resolve(electronRoot, 'lib'),
-        'electron$': electronAPIFile,
+        electron$: electronAPIFile,
         'electron/main$': electronAPIFile,
         'electron/renderer$': electronAPIFile,
         'electron/common$': electronAPIFile,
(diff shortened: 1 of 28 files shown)


nornagon:
Why disable relative imports?

MarshallOfSound:
@nornagon Otherwise you can bypass the rules by just relatively importing something from one of the disallowed directories.

nornagon:
@MarshallOfSound is the `no-restricted-imports` module not smart enough to resolve the path first?

MarshallOfSound:
Unfortunately not

PR #24517, opened 2020-07-10
feat: add webContents.setWindowOpenHandler API

loc:
## API changes:

* ### **Deprecate webContents `new-window` event in favor of `webContents.setWindowOpenHandler(handler)` and `did-create-window`**
  `new-window` is fired _after_ the contents is already created by Chrome. This is a problem when the application wants to prevent that window from being opened, as killing the window during initialization causes many problems and leads to the inevitable crash. Instead, we'll allow cancellation/customization of the new window _before_ it's created. In congress with a change to set WebPreferences via IPC instead of the command line, this allows us to completely* change the child window's BrowserWindowConstructorOptions, even for in-process windows.

   \*With the exception of a few outstanding preferences that we need early in initialization, like `offscreen` and `nodeIntegrationInWorker`. These could be fixed with some effort.

  **Old way:**
  ```javascript
  mainWindow.webContents.on('new-window', (event, url, frameName, disposition, options, additionalFeatures) => {
      Object.assign(options, {
        modal: true,
      });
      childWindow = new BrowserWindow(options);
      event.newGuest = childWindow;
  });

  // or

  mainWindow.webContents.on('new-window', (event, url, frameName, disposition, options, additionalFeatures) => {
      event.preventDefault();
  });
  ```

  **New way:**
  ```javascript
  mainWindow.webContents.setWindowOpenHandler(({ url, frameName }) => {
    return {
      action: 'allow',
      overrideBrowserWindowOptions: {
        modal: true,  
      }
    }
  });
  mainWindow.webContents.on('did-create-window', (win, { url, frameName, options, disposition, additionalFeatures, referrer, postData }) => {
    childWindow = win;
  });

  // or

  mainWindow.webContents.setWindowOpenHandler(({ url, frameName }) => {
    return { action: 'deny' };
  });
  ```

* ### (internal API) `webFrame.getWebPreference(prefName)` 
  Previously, preferences were stuffed into command line switches so that the renderer could look and see what settings it was supposed to be using. However, this wouldn't work for in-process child windows. Instead, we'll stuff Electron preferences into the struct that Chrome already uses for this sort of thing and add an API for the renderer to retreive those preferences. At the moment, only Electron-specific WebPreferences are available for lookup, but it would be easy to add Chromium-defined prefs like defaultFontSize, for example.

  ```javascript
  const { webFrame } = require('electron');
  const isContextIsolationEnabled = webFrame.getWebPreference('contextIsolation');
  ```

## Implementation details:
- Allowing for in-process WebPreference changes

    Prior to this change, WebPreferences were set via the command line, which meant that in-process child windows were forced to inherit WebPreferences from their parent. Now, we overload the Chrome WebPreference struct with our own properties so they are accessible in the renderer (via a new `webFrame.getWebPreference`).

    window-open

- Some behavior changes that make window opening less, um, surprising.
  - Non-security related BrowserWindowConstructorOptions are no longer inherited from the parent. This was confusing when you had to "undo" preferences that may have been set in a different part of your app.
  - window.open `features` keys without values are now interpreted as `true`, to w3c spec.
  - We no longer check for cycles in BrowserWindowOptions. (https://github.com/electron/electron/pull/8340)

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Add setWindowOpenHandler API for renderer-created child windows, and deprecate new-window event.


docs/api/web-contents.md
```diff
@@ -1081,6 +1114,17 @@ Works like `executeJavaScript` but evaluates `scripts` in an isolated context.
 
 Ignore application menu shortcuts while this web contents is focused.
 
+#### `contents.setWindowOpenOverride(handler)`
```
MarshallOfSound, line 1117: `setWindowOpenHandler`? You've even called the parameter `handler` so I think it fits a bit better

docs/api/web-contents.md
```diff
@@ -1081,6 +1114,17 @@ Works like `executeJavaScript` but evaluates `scripts` in an isolated context.
...
+
+* `handler` Function
+  * `details` Object
+    * `url` String - URL passed to `window.open()`
```
MarshallOfSound, line 1121: Can we specify, is this the _resolved_ URL or the string that was passed to that method. i.e. does `windoe.open('/foo')` result in `http://example.com/foo` or just `/foo`

docs/api/window-open.md
```diff
@@ -1,34 +1,57 @@
...
+untrusted content within a renderer. Windows can be created from the renderer in two ways:
+
+- clicking on links or submitting forms adorned with `target=_blank`
+- JavaScript calling `window.open()`
+
+By default, this results in the creation of a
+[`BrowserWindowProxy`](browser-window-proxy.md), a light wrapper around
+`BrowserWindow`.
+
+However, when the `contextIsolation` or `sandbox` (or directly,
```
MarshallOfSound, line 13: This `contextIsolation` comment is incorrect, only `nativeWindowOpen` or `sandbox` (which implies `nativeWindowOpen`) will result in a real Window instance

lib/browser/api/web-contents.ts
```diff
@@ -434,6 +434,28 @@ WebContents.prototype.loadFile = function (filePath, options = {}) {
...
+    const response = this.windowOpenOverrideHandler({ url, frameName });
+
+    switch (typeof response) {
+      case 'boolean':
+        if (response === false) event.preventDefault();
+        return;
+      case 'object':
+        return response;
+      default:
+        throw new TypeError(
```
MarshallOfSound, line 452: This error is uncatchable, we should either just log this or refactor the API so the error is catchable

lib/browser/api/web-contents.ts
```diff
@@ -553,43 +577,60 @@ WebContents.prototype._init = function () {
...
+          frameName,
+          features: rawFeatures
+        }
+      });
+    });
 
-      internalWindowOpen(event, url, referrer, frameName, disposition, mergedOptions, additionalFeatures, postData);
+    let windowOpenOverriddenOptions: BrowserWindowConstructorOptions | null = null;
+    this.on('-will-add-new-contents' as any, (event: any, url: string, frameName: string) => {
+      process.stdout.write('test\ntest\nhere\ncool\n');
```
MarshallOfSound, line 598: Noice

lib/browser/api/web-contents.ts
```diff
@@ -553,43 +577,60 @@ WebContents.prototype._init = function () {
...
+        windowOpenArgs: {
+          url,
+          frameName,
+          features: rawFeatures
+        }
+      });
+    });
 
-      internalWindowOpen(event, url, referrer, frameName, disposition, mergedOptions, additionalFeatures, postData);
+    let windowOpenOverriddenOptions: BrowserWindowConstructorOptions | null = null;
```
MarshallOfSound, line 596: Can we guarantee that there is no possible race condition here?  I'd be concerned that enough `window.open` calls could result in us assigning the wrong `webPreferences` to the new window

lib/browser/api/web-contents.ts
```diff
@@ -553,43 +577,60 @@ WebContents.prototype._init = function () {
...
+    this.on('-will-add-new-contents' as any, (event: any, url: string, frameName: string) => {
+      process.stdout.write('test\ntest\nhere\ncool\n');
+      windowOpenOverriddenOptions = this._callWindowOpenOverride(event, url, frameName);
+      if (!event.defaultPrevented && windowOpenOverriddenOptions) {
+        this._setNextChildWebPreferences({
```
MarshallOfSound, line 601: Same question as above, I'd be way more comfortable with this if we had a test that called `window.open` like 50 times in a row (synchronously) and asserts that the webPrefs end up on the right window.

lib/browser/guest-window-manager.ts
```diff
@@ -0,0 +1,285 @@
...
+import { parseFeatures } from '../common/parse-features-string';
+
+type PostData = LoadURLOptions['postData']
+export type WindowOpenArgs = {
+  url: string,
+  frameName: string,
+  features: string,
+}
+
+const frameNamesToWindow = new Map();
```
MarshallOfSound, line 18: `Map` of what to what? `Map<K, V>`

patches/chromium/allow_in_process_windows_to_have_different_web_prefs.patch
```diff
@@ -0,0 +1,132 @@
...
+diff --git a/content/public/common/web_preferences.cc b/content/public/common/web_preferences.cc
+index b543f98a34266efdb771482c5717042700a47ed0..4ab4ae9855ad470fe8c79de7a294b6630a5f60f6 100644
+--- a/content/public/common/web_preferences.cc
++++ b/content/public/common/web_preferences.cc
+@@ -166,6 +166,27 @@ WebPreferences::WebPreferences()
+       navigate_on_drag_drop(true),
+       v8_cache_options(blink::mojom::V8CacheOptions::kDefault),
+       record_whole_document(false),
++
++      preloads(base::EmptyString()),
```
MarshallOfSound, line 79: `preloads` vs `preload`?

shell/browser/web_contents_preferences.cc
```diff
@@ -397,6 +338,7 @@ void WebContentsPreferences::AppendCommandLineSwitches(
     }
   }
 
+  std::string s;
```
MarshallOfSound, line 341: Noice

shell/browser/electron_browser_client.cc
```diff
@@ -604,13 +604,21 @@ void ElectronBrowserClient::OverrideWebkitPrefs(
                                       ? blink::PreferredColorScheme::kDark
                                       : blink::PreferredColorScheme::kLight;
 
+  auto* web_contents = content::WebContents::FromRenderViewHost(host);
+  auto preloads =
+      SessionPreferences::GetValidPreloads(web_contents->GetBrowserContext());
+  if (!preloads.empty())
+    prefs->preloads = base::JoinString(preloads, kPathDelimiter);
```
MarshallOfSound, line 611: Can we pass this as a vector instead of a weirdly joined string?

shannawhaley0:
in-process-web-prefs

lib/browser/api/web-contents.ts
loc: it's marked as a DRAFT, Sam 😂

shell/browser/web_contents_preferences.cc
loc: This is just moved from L355.

patches/chromium/allow_in_process_windows_to_have_different_web_prefs.patch
loc: I'm preserving the distinction already present between `kPreloadScript` and `kPreloadScripts`

docs/api/web-contents.md
loc: Good point — it is the resolved URL. Will add a note.

lib/browser/api/web-contents.ts
loc: Yes it is guaranteed, since both the `-will-add-new-contents` and `-add-new-contents` are fired sequentially from the blue box in the diagram. Specifically, they are both fired from the same method in Chrome: `WebContents::CreateNewWindow`, in the delegate callbacks for `delegate_->IsWebContentsCreationOverridden` and `delegate_->AddNewContents`.

docs/api/web-contents.md
loc: I think I agree with you, but will let the API WG weigh in before making the change.

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -684,6 +688,15 @@ void WebContents::WebContentsCreatedWithFullParams(
...
   tracker->body = params.body;
+
+  v8::HandleScope scope(isolate());
+  v8::Local<v8::Value> val =
+      gin::ConvertToV8(isolate(), pending_child_web_preferences_);
+  gin_helper::Dictionary dict;
+  gin::ConvertFromV8(isolate(), val, &dict);
+  pending_child_web_preferences_.DictEmpty();
+
+  new WebContentsPreferences(new_contents, dict);
```
loc, line 699: I would very much appreciate a less insane way to do this.

loc:
A couple tests are still failing, but I'm going to get the review ball rolling while I address them.

shell/browser/api/electron_api_web_contents.cc
MarshallOfSound: @loc You can temporarily put the `v8::Value` into a `v8::Global`, that allows it to exist for longer than the stack it was allocated on.

docs/api/web-contents.md
```diff
@@ -1081,6 +1114,17 @@ Works like `executeJavaScript` but evaluates `scripts` in an isolated context.
 
 Ignore application menu shortcuts while this web contents is focused.
 
+#### `contents.setWindowOpenOverride(handler)`
+
+* `handler` Function<Boolean | BrowserWindowConstructorOptions>
+  * `details` Object
+    * `url` String - The _resolved_ version of the URL passed to `window.open()`. e.g. opening a window with `window.open('foo')` will yield something like `https://the-origin/the/current/path/foo`.
+    * `frameName` String - Name of the window provided in `window.open()`
+  Returns `Boolean | BrowserWindowConstructorOptions` - `false` cancels the creation of the new window. `true` results in the default behavior: a window is created. Returning `BrowserWindowConstructorOptions` allows customization of the created window. Returning `undefined` will result in a console error and cancellation of the window.
```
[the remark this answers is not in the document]
loc: I'd argue that the API you've proposed is also odd, and features a magical property with unclear semantics that lack precedent. From the function signature, it's unclear that it's even possible to override the options. It makes use of parameter mutation, which is [widely considered to be bad practice](https://eslint.org/docs/rules/no-param-reassign.html). Whereas in what I've proposed, the semantics are clear and idiomatic: the handler will be passed inputs and return outputs, and those outputs will affect the course of the program. I'd hardly call it magical that the program's response changes given a function's varying return values.

At first blush, it does seem that event semantics would be appropriate, but _notification_ is not the primary goal of this API, _consultation_ is. And there's no real precedent for consultation semantics in javascript that I can think of, because normally it's possible to get consultation with the union of notification (`event.preventDefault()`) and action (`window.doTheThingMyWay()`). But in this case, due to the precariousness of where we've had to hook into Chrome, the program has to take the action, rather than being allowed to delegate to the consumer. I agree that it's not an ideal API, but as so often happens in Electron, the ideal API is technically infeasible.

docs/api/web-frame.md
```diff
@@ -170,6 +170,28 @@ Note that when the execution of script fails, the returned promise will not
 reject and the `result` would be `undefined`. This is because Chromium does not
 dispatch errors of isolated worlds to foreign worlds.
 
+### `webFrame.getWebPreference(prefName)`
```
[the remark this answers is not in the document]
loc: Yeah, I think you're right. I could see value in allowing the renderer to introspect its given webPreferences, but given that only a rather inexplicable subset of the options are available, keeping it internal makes more sense.

patches/chromium/allow_in_process_windows_to_have_different_web_prefs.patch
```diff
@@ -0,0 +1,135 @@
...
+@@ -3097,12 +3101,6 @@ RenderFrameHostDelegate* WebContentsImpl::CreateNewWindow(
+     AddDestructionObserver(new_contents_impl);
+   }
+ 
+-  if (delegate_) {
+-    delegate_->WebContentsCreatedWithFullParams(this, render_process_id,
+-                                                opener->GetRoutingID(),
+-                                                params, new_contents_impl);
+-  }
+-
```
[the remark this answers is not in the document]
loc: Oops, good catch.

lib/browser/api/web-contents.ts
loc: You're right, it needs to be safer. Hmm. Back to the drawing board 🤔

loc: Upon further review, it actually is safe. The reason it's safe is because the params are associated with the new WebContents in `WebContentsCreatedWithFullParams` (our patched version of `WebContentsCreated`) via the `WebContentsPreferences` which associates the WebPreferences with the WebContents by `content::WebContentsUserData`. This is called before the AddNewContents paths branch. Obviously this is not clear enough because we both missed it and I wrote the code 😂 😭  So I'll add some comments for clarity.

jkleinsc:
The @electron/wg-api reviewed this at the August 24, 2020 meeting and decided to defer voting on it.

lib/browser/api/web-contents.ts
nornagon: Is there some way we can make this more direct in the code? It feels very fragile and prone to breaking in obscure situations, especially if the upstream code changes at all.

docs/api/web-contents.md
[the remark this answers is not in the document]
nornagon: I don't think `event.foo = bar` lacks precedent; we do the same thing in sync IPC with `event.returnValue` ([docs](https://www.electronjs.org/docs/api/ipc-main#sending-messages)). FWIW, that API is also present in the web, for instance in the [beforeunload event](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event). I'll grant it's unusual, though.

With the return-style API, is there any way to inspect what the default options would be? How would you express e.g. "do the default, but set the x/y coordinates"?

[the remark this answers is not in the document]
loc: Yes, the way to represent it would just be

```
webContents.setWindowOpenOverride(() => ({
  x: 500,
  y: 500,
}))
```

> is there any way to inspect what the default options would be?

It would be the same as creating a new BrowserWindow insofar as there are defaults, but they are inscrutable to the consumer. This is a departure from the way things work now, as I mention in the PR description:

> Non-security related BrowserWindowConstructorOptions are no longer inherited from the parent. This was confusing when you had to "undo" preferences that may have been set in a different part of your app.

lib/browser/api/web-contents.ts
loc: I mean, we could always patch the hell out of Chrome, but I was really trying to avoid that. 

If I could find a more direct way, I definitely would've written it that way 😄 Not to say there isn't one, but it is apparently beyond my ability. I welcome suggestions though!

patches/chromium/allow_in_process_windows_to_have_different_web_prefs.patch
```diff
@@ -0,0 +1,135 @@
...
++  IPC_STRUCT_TRAITS_MEMBER(preload)
++  IPC_STRUCT_TRAITS_MEMBER(native_window_open)
++  IPC_STRUCT_TRAITS_MEMBER(node_integration)
++  IPC_STRUCT_TRAITS_MEMBER(node_integration_in_worker)
++  IPC_STRUCT_TRAITS_MEMBER(node_leakage_in_renderers)
++  IPC_STRUCT_TRAITS_MEMBER(node_integration_in_sub_frames)
++  IPC_STRUCT_TRAITS_MEMBER(enable_spellcheck)
++  IPC_STRUCT_TRAITS_MEMBER(enable_plugins)
++  IPC_STRUCT_TRAITS_MEMBER(enable_websql)
++  IPC_STRUCT_TRAITS_MEMBER(webview_tag)
```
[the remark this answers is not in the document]
loc: I dunno, is there? 😄

typings/internal-electron.d.ts
```diff
@@ -64,6 +64,9 @@ declare namespace Electron {
     equal(other: WebContents): boolean;
     _initiallyShown: boolean;
     browserWindowOptions: BrowserWindowConstructorOptions;
+    _windowOpenHandler: any;
```
miniak, line 67: can we avoid `any` here?

shell/browser/web_contents_preferences.cc
```diff
@@ -506,6 +414,89 @@ void WebContentsPreferences::OverrideWebkitPrefs(
   if (GetAsString(&preference_, "defaultEncoding", &encoding))
     prefs->default_encoding = encoding;
 
+  // // --background-color.
```
miniak, line 417: double comment

patches/chromium/allow_in_process_windows_to_have_different_web_prefs.patch
```diff
@@ -0,0 +1,260 @@
...
+   }
+diff --git a/third_party/blink/public/mojom/webpreferences/web_preferences.mojom b/third_party/blink/public/mojom/webpreferences/web_preferences.mojom
+index ac1099c8fb923184a014bf9cbaa35c496cf85cd9..d9804ff04ec80ca4458d141d3010c29557bb6fd5 100644
+--- a/third_party/blink/public/mojom/webpreferences/web_preferences.mojom
++++ b/third_party/blink/public/mojom/webpreferences/web_preferences.mojom
+@@ -202,6 +202,29 @@ struct WebPreferences {
+   V8CacheOptions v8_cache_options;
+   bool record_whole_document;
+ 
++  // Begin Electron-specific WebPreferences.
```
miniak, line 228: can this be delegated somehow to avoid having to modify the patch when adding new preferences?

nornagon: Asked before: https://github.com/electron/electron/pull/24517#discussion_r463266797

lib/browser/api/web-contents.ts
nornagon: It looks to me like `ShowCreatedWindow` isn't guaranteed to be called after `IsWebContentsCreationOverridden`, not immediately, and not necessarily _ever_.

Here's the way that can happen:
1. Renderer calls [CreateNewWindow](https://source.chromium.org/chromium/chromium/src/+/master:content/renderer/render_view_impl.cc;l=440;drc=04ccda060610036333f90b365e21fa76bbc684b9)
2. Renderer creates a OnceCallback that will call `ShowCreatedWindow`. It passes the callback to...
3. [RenderView::Create](https://source.chromium.org/chromium/chromium/src/+/master:content/renderer/render_view_impl.cc;l=519;drc=04ccda060610036333f90b365e21fa76bbc684b9) (this is guaranteed to happen in all cases). RenderView::Create calls...
4. [RenderView::Initialize](https://source.chromium.org/chromium/chromium/src/+/master:content/renderer/render_view_impl.cc;l=327;drc=929e790ad17a54008b3dac86cbe30751c968c2ce).
5. If `!local_main_frame`, the `ShowCreatedWindow` callback is dropped, and the callback will never be invoked.

I'm not 100% sure of all the cases in which this can occur, but possibly when opening cross-origin windows?

Separately, in the case that `local_main_frame` is true, the callback isn't invoked immediately. Instead, it's [saved](https://source.chromium.org/chromium/chromium/src/+/master:content/renderer/render_widget.cc;l=260;drc=929e790ad17a54008b3dac86cbe30751c968c2ce) in RenderWidget::Initialize, and only later invoked in [RenderWidget::Show](https://source.chromium.org/chromium/chromium/src/+/master:content/renderer/render_widget.cc;l=399;drc=929e790ad17a54008b3dac86cbe30751c968c2ce). The latter is not guaranteed to be called, in particular [if an old window was reused](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/page/create_window.cc;l=316;drc=929e790ad17a54008b3dac86cbe30751c968c2ce).

I think the way to trigger this case would be to try to `window.open` twice with the same frame name.

nornagon: I tried opening multiple windows with the same frame name, and also opening multiple cross-origin child windows, and I wasn't able to observe any interleaving.

nornagon: I'm going to resolve this for now. I'd like to add a test for this but I don't really know what to test for. I'm not able to observe any misbehavior currently, and I'm not sure how to anticipate what might trigger misbehavior in the future. Let's be on the lookout for failures in this area in future and be prepared to respond + refactor.

nornagon:
Just collecting the outstanding comments on this PR as it's a long one and there's been a lot of comments. There's just 2 remaining:

1. The thing where you return a boolean or a BrowserWindowConstructorOptions is kinda wonky. https://github.com/electron/electron/pull/24517#discussion_r463133325
2. It sucks to have to patch in our WebPreferences. https://github.com/electron/electron/pull/24517#discussion_r463266797

For (2) I'm okay eating the patch for now, until such a time as we discover or create a better option. It's a reasonably manageable patch.

I've thought about (1) for a while and here are the alternatives I've considered:

1. Use an event, and pass options using a property of the event object.
```js
contents.on('will-create-window', (event) => {
  if (!isAllowed(event.url)) {
    event.preventDefault()
    return
  }
  if (shouldBeModal(event.url)) {
    event.browserWindowOptions = { modal: true }
  }
})
```

This approach has the advantage that it works almost like a normal event: you can cancel it with `preventDefault`, or let it happen by not doing that. However, it's not common in JS for event handlers to mutate the event object. There are a few examples of it in Electron and the Web (e.g. `e.returnValue`), but it's not very discoverable. Additionally, this doesn't allow statically checking that the event handler correctly made a determination on whether the window-open should continue or not.

2. Use an event, and pass a callback that must be called synchronously.
```js
contents.on('will-create-window', (event, createWindow) => {
  if (!isAllowed(event.url)) {
    return
  }
  const options = shouldBeModal(event.url) ? { modal: true } : {}
  createWindow(options)
})
```

This avoids mutating the event argument, but introduces a different difficulty, which is that generally speaking in JavaScript, callback arguments are allowed to be invoked asynchronously. A developer would likely see the above snippet and conclude that it would be OK to call the callback asynchronously if they weren't familiar with the API. There's no way to enforce the must-be-called-synchronously requirement statically, even with TypeScript. Further, there's no way to statically check that the callback is called at most once. And finally, this API doesn't really make sense as an event, as there can be multiple handlers for an event, and it's unclear how those handlers should interact with one another to collaborate on making a decision about opening the window.

3. Use a handler, and pass a callback that must be called synchronously.
```js
contents.setWindowOpenHandler((params, createWindow) => {
  if (isAllowed(params.url)) {
    const options = shouldBeModal(params.url) ? { modal: true } : {}
    createWindow(options)
  }
})
```

This avoids the "many event listeners, how do they cooperate?" issue from (2), but shares the other issues about ensuring at-most-once, synchronous calling of the callback.

4. Use a handler, but return a more descriptive and extensible object.
```js
contents.setWindowOpenHandler((params) => {
  if (!isAllowed(params.url))
    return { action: 'deny' }
  return {
    action: 'allow',
    overrideBrowserWindowOptions: {
      modal: shouldBeModal(params.url)
    }
  }
})
```

This is essentially the same as the current API, but using a different return type. Instead of `bool | BrowserWindowConstructorOptions`, we always return an object, and choose what to do based on the `action` property. Additional parameters can be specified in the object. This is more readable, and also more extensible—we can add further actions or further options for existing actions in future.


---

Other than the slightly awkward ergonomics of the current approach, I see three remaining potential areas we could improve on in the current API design:

1. Composability. The current approach allows only for a single handler per WebContents. There's no built-in option for multiple different parts of the app to register handlers and collaborate (e.g. one handler that checks for blocked origins, another that checks for certain windows that should be started frameless, and a third that checks for repeated popup requests). That would all have to be coordinated within the one handler, or else by a custom coordination system supplied by the app.
2. Extensibility. As it stands, this API can choose to do one of three things: deny the creation of the window, allow the creation of the window, or allow the creation of the window and specify custom options. If, in future, we want to add a fourth option to this list, we'll have some trouble extending the API to accommodate that.
3. Boilerplate. I suspect that the most common usage of this will be to apply the same handler to every web contents. Thus, the app will register an `app.on('web-contents-created')` handler which then sets the handler for every created WebContents. It might be better to provide an option for a catch-all handler.

The handler-based approach with a more descriptive return type (alternative (4) above) improves on the extensibility of the current approach, but is not any more composable. I think alternative (4) is a strict improvement over the current approach.

The event-based approach (alternative (1) above) improves on the handler-based approach on all three of the above qualities. It's an event, so multiple handlers can be registered, and they can cooperate without knowing about each other. We can add further mutable properties to the event in future if needed. And we can add a secondary event emission on `app` after the event has been emitted on `webContents`. And, as a bonus, this is a fairly familiar shape for an API like this to take; there are many others like it in Electron and elsewhere.

I'm still torn between event approach—alternative (1) above—and the return-an-action-object approach—alternative (4) above. The most compelling thing about the handler-based approach, to me, is that it is statically verifiable that the response is synchronous. There's no other statically-checkable way in JS/TS to assert that a result be specified synchronously than with a return value. Additionally, it avoids the awkward mutate-the-event API. However, I'm not sure that outweighs the advantages of the event-based API outlined above. It's a relatively common pattern in JS APIs to have operations that must be handled synchronously (e.g. 'paste' events, IndexedDB transactions, onbeforeunload, calls to preventDefault(), etc.), so I don't think it's unreasonable to require that here. Mutating the event is unusual but not unprecedented.

Thoughts? @electron/wg-api?

nornagon:
I updated the PR to implement alternative (4) as I feel it was an improvement on the previous implementation. Pending https://github.com/electron/docs-parser/pull/46.

jkleinsc:
I'm in favor of approach (4).  Mutating the event feels clunky even if there is precedence to do so.  I also like the declarative nature of the return object on (4).

jkleinsc:

nornagon (approved):

sentialx:
Is it now possible to set `e.newGuest` to a BrowserView?

stewartlord:
@nornagon Is there a way to influence how the new window is created? Specifically, we are looking at how to open into a BrowserView

stewartlord:
@loc @nornagon Just wanted to share an idea that @sentialx and I discussed to add support for browser views. We tried to think of a way to do this without changing too much about the new `setWindowOpenHandler()` API. I'm not sure it's ideal, but it looks like the design is being driven by a requirement to leave creation of the new guest up to electron internally? With that in mind:

```
mainWindow.webContents.setWindowOpenHandler(({ url, frameName }) => {
  return {
    action: 'allow',
    guestType: 'browserView', // default: 'window'
    overrideBrowserViewOptions: {
      ...
    }
  }
});
mainWindow.webContents.on('did-create-browser-view', (view, { url, frameName, options, disposition, additionalFeatures, referrer, postData }) => {
  browserWindow.addBrowserView(view)
  ...
});
```

I believe there might be an outstanding issue re: 'close'? Eryk was going to think about this a bit more and refine the idea, but I wanted to get some initial feedback on the above.

nornagon:
@stewartlord hm, that proposal raises a few concerns for me, specifically it seems difficult to correctly correlate the window open handler invocation with the `did-create-browser-view` event.

However, I like the direction this is going in and to collect discussion I'd encourage you to open a proposal over at https://github.com/electron/governance/tree/master/wg-api/spec-documents (see https://github.com/electron/governance/blob/master/wg-api/spec-documents/api-spec-template.md for a template)

stewartlord:
@nornagon Thanks for the feedback. Eryk had a different idea and opened a PR that should solve the issue with less effort:
https://github.com/electron/electron/pull/26802

pushkin-:
Is there a way with this new paradigm to pend window creation? I only want to create the window if the URL is in my allowlist (whitelist). Currently, I'm preventing the `new-window` event, checking my allowlist asynchronously, and then creating the window.

Is the way to do this using the new paradigm to create the window in a hidden state, then after it's been created, check with the allowlist and close the window if the URL doesn't pass the check; otherwise, show the window? I presume I can do that bit in `mainWindow.webContents.on('did-create-window'` since I expect that can be asynchronous.

nornagon:
There is not. Chromium code forces that a decision about whether to create the window or not be made synchronously.

PR #24528, opened 2020-07-13
fix: Node.js cpu and heap profiling

codebytere:
#### Description of Change

What it says on the tin. In Node.js this is done during `CreateMainInstance` inside `InitializeInspector()`. We don't call `InitializeInspector()` because we handle node debugging setup ourselves, and neglected to appropriately start heap and cpu profilers. This fixes that.

cc @nornagon @MarshallOfSound @jkleinsc 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where cpu and heap profiling in Node.js did not work properly with `--cpu-prof`, `--heap-prof`, and related CLI flags.


nornagon:
Is this related to some node-specific command-line flags? It would be nice to mention the flags in the release notes if so!

PR #24534, opened 2020-07-13
fix: ensure that errors thrown in the context bridge are created in the correct context

MarshallOfSound:
This PR fixes two cases where errors were being created in context A but thrown in context B.

* When the V8 serializer threw an error, it always threw it in the source context, now it throws in the appropriate error context
* If we throw an error as a result of a return value failing to be converted, it created the error in the source context when it should create it in the destination context as that is where the error will be thrown / caught.

Notes: no-notes

### DIFF
--- a/shell/renderer/api/electron_api_context_bridge.cc
+++ b/shell/renderer/api/electron_api_context_bridge.cc
@@ -127,6 +127,22 @@ v8::MaybeLocal<v8::Value> GetPrivate(v8::Local<v8::Context> context,
                           gin::StringToV8(context->GetIsolate(), key)));
 }
 
+// Where the context bridge should create the exception it is about to throw
+enum BridgeErrorTarget {
+  // The source / calling context.  This is default and correct 99% of the time,
+  // the caller / context asking for the conversion will receive the error and
+  // therefore the error should be made in that context
+  kSource,
+  // The destination / target context.  This should only be used when the source
+  // won't catch the error that results from the value it is passing over the
+  // bridge.  This can **only** occur when returning a value from a function as
+  // we convert the return value after the method has terminated and execution
+  // has been returned to the caller.  In this scenario the error will the be
+  // catchable in the "destination" context and therefore we create the error
+  // there.
+  kDestination
+};
+
 }  // namespace
 
 v8::MaybeLocal<v8::Value> PassValueToOtherContext(
@@ -135,10 +151,13 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
     v8::Local<v8::Value> value,
     context_bridge::ObjectCache* object_cache,
     bool support_dynamic_properties,
-    int recursion_depth) {
+    int recursion_depth,
+    BridgeErrorTarget error_target = BridgeErrorTarget::kSource) {
   TRACE_EVENT0("electron", "ContextBridge::PassValueToOtherContext");
   if (recursion_depth >= kMaxRecursion) {
-    v8::Context::Scope source_scope(source_context);
+    v8::Context::Scope error_scope(error_target == BridgeErrorTarget::kSource
+                                       ? source_context
+                                       : destination_context);
     source_context->GetIsolate()->ThrowException(v8::Exception::TypeError(
         gin::StringToV8(source_context->GetIsolate(),
                         "Electron contextBridge recursion depth exceeded.  "
@@ -303,9 +322,12 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
   // Serializable objects
   blink::CloneableMessage ret;
   {
-    v8::Context::Scope source_context_scope(source_context);
+    v8::Local<v8::Context> error_context =
+        error_target == BridgeErrorTarget::kSource ? source_context
+                                                   : destination_context;
+    v8::Context::Scope error_scope(error_context);
     // V8 serializer will throw an error if required
-    if (!gin::ConvertFromV8(source_context->GetIsolate(), value, &ret))
+    if (!gin::ConvertFromV8(error_context->GetIsolate(), value, &ret))
       return v8::MaybeLocal<v8::Value>();
   }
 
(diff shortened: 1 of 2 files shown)


shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -303,7 +322,9 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
   // Serializable objects
   blink::CloneableMessage ret;
   {
-    v8::Context::Scope source_context_scope(source_context);
+    v8::Context::Scope error_scope(error_target == BridgeErrorTarget::kSource
+                                       ? source_context
+                                       : destination_context);
     // V8 serializer will throw an error if required
     if (!gin::ConvertFromV8(source_context->GetIsolate(), value, &ret))
```
nornagon, line 329: is it ok to use `source_context->GetIsolate()` here or should we use the isolate from the destination context when appropriate? Are they guaranteed to be the same isolate?

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -303,7 +322,9 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
   // Serializable objects
   blink::CloneableMessage ret;
   {
-    v8::Context::Scope source_context_scope(source_context);
+    v8::Context::Scope error_scope(error_target == BridgeErrorTarget::kSource
+                                       ? source_context
+                                       : destination_context);
```
nornagon, line 327: won't this result in the value produced from the converter being in a different context, as well as the error? if so, `error_scope` isn't the right name for this. additionally, i'm not sure that it's correct to produce the value in the destination context in this situation.

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -127,6 +127,22 @@ v8::MaybeLocal<v8::Value> GetPrivate(v8::Local<v8::Context> context,
...
+  // the caller / context asking for the conversion will receive the error and
+  // therefore the error should be made in that context
+  kSource,
+  // The destination / target context.  This should only be used when the source
+  // won't catch the error that results from the value it is passing over the
+  // bridge.  This can **only** occur when returning a value from a function as
+  // we convert the return value after the method has terminated and execution
+  // has been returned to the caller.  In this scenario the error will the only
+  // catchable in the "destination" context and therefore we create the error
+  // there.
```
nornagon, line 142:
```suggestion
  // has been returned to the caller.  In this scenario the error will only be
  // catchable in the "destination" context and therefore we create the error
  // there.
```

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -391,10 +412,10 @@ void ProxyFunctionWrapper(const v8::FunctionCallbackInfo<v8::Value>& info) {
...
       return;
 
-    auto ret =
-        PassValueToOtherContext(func_owning_context, calling_context,
-                                maybe_return_value.ToLocalChecked(),
-                                &object_cache, support_dynamic_properties, 0);
+    auto ret = PassValueToOtherContext(
+        func_owning_context, calling_context,
+        maybe_return_value.ToLocalChecked(), &object_cache,
+        support_dynamic_properties, 0, BridgeErrorTarget::kDestination);
```
nornagon, line 418: Would it also be correct to use `TryCatch` here to capture the error and throw a new error in the correct context? e.g.
```suggestion
    v8::TryCatch try_catch(isolate);
    auto ret = PassValueToOtherContext(...);
    if (try_catch.HasCaught()) {
      ThrowException("Conversion of return value failed");
      return;
    }
```

spec-main/api-context-bridge-spec.ts
```diff
@@ -472,6 +472,39 @@ describe('contextBridge', () => {
...
+        const result = await callWithBindings((root: any) => {
+          const getError = (fn: Function) => {
+            try {
+              fn();
+            } catch (e) {
+              return e;
+            }
+            return null;
+          };
+          const normalIsError = Object.getPrototypeOf(getError(() => root.example.throwNormal())) === Error.prototype;
```
nornagon, line 499:
```suggestion
          const normalIsError = Object.getPrototypeOf(getError(root.example.throwNormal)) === Error.prototype;
```

shell/renderer/api/electron_api_context_bridge.cc
MarshallOfSound: Technically they're guaranteed to be the same isolate, but for "correctness" I've used the correct context everywhere else, may as well use it here.

shell/renderer/api/electron_api_context_bridge.cc
MarshallOfSound: No it won't, this conversion is **from** V8 and this context scope is only used for that conversion, the only V8 value that could get created during the conversion from V8 to `blink::Cloneable` is the error

shell/renderer/api/electron_api_context_bridge.cc
MarshallOfSound: This is handled inside `PassValueToOtherContext`, the way that function is written and intended to be used is that it always does everything in the right context.  Any errors thrown are thrown in the correct context too so this try catch would be unnecessary.

spec-main/api-context-bridge-spec.ts
MarshallOfSound: Not sure it matters in this particular case so I could change this but I'm not a huge fan of passing methods by property access as it rebinds / loses the `this` binding in many cases and can do so silently.  It's normally safer / more consistent to pass a lamda style function.

shell/renderer/api/electron_api_context_bridge.cc
nornagon: Yeah, I was just thinking about this as an alternative implementation that would be less intrusive (i.e. there would no need for this extra enum).

MarshallOfSound: It means we'd have to `trycatch` every call to `PassValueToOtherContext` which would be considerably more intrusive?

PR #24542, opened 2020-07-14
fix: Node.js cpu and heap profiling

codebytere:
Backport of https://github.com/electron/electron/pull/24528.

See that PR for more details.

Notes: Fixed an issue where cpu and heap profiling in Node.js did not work properly with `--cpu-prof`, `--heap-prof`, and related CLI flags.


codebytere:
We can't re-enabled tests here owing to: https://github.com/electron/electron/pull/23213 not being in this branch as it hasn't been exposed yet. The tests use `assert.ok` which dies as we don't use Node.js' prepareStackTrace callback in this PR. This should be otherwise good to go though.

PR #24545, opened 2020-07-14
build: add apple silicon support

MarshallOfSound:
Does what it says on the tin, the patch surface area has been reduced to basically 0 as a result of upstream work.  There are still a few things being upstreamed and the remaining patches should disappear then.

Notes: Added new experimental apple silicon (darwin arm64) builds


.circleci/config.yml
```diff
@@ -1939,22 +1957,14 @@ jobs:
...
       <<: *env-testing-build
     <<: *steps-electron-gn-check
 
-  osx-chromedriver:
-    <<: *machine-mac
-    environment:
-      <<: *env-machine-mac
-      <<: *env-release-build
-      <<: *env-send-slack-notifications
-    <<: *steps-chromedriver-build
```
nornagon, line 1955: Was it intentional to remove the chromedriver build?

.circleci/config.yml
```diff
@@ -2009,7 +2036,7 @@ jobs:
           checkout-and-assume-cache: true
           attach: false
 
-  mas-testing-gn-check:
+  mas-testing-x64-gn-check:
```
nornagon, line 2039: do we need to build MAS versions of arm64 as well?

patches/chromium/chore_disable_failing_blink_font_dcheck_on_big_sur.patch
```diff
@@ -0,0 +1,26 @@
...
+   SkFontParameters::Variation::Axis axes_parameters[num_axes];
+-  int returned_axes =
+-      typeface->getVariationDesignParameters(axes_parameters, num_axes);
+-  DCHECK_EQ(num_axes, returned_axes);
+-  DCHECK_GE(num_axes, 0);
++  // int returned_axes =
++  //     typeface->getVariationDesignParameters(axes_parameters, num_axes);
++  // DCHECK_EQ(num_axes, returned_axes);
++  // DCHECK_GE(num_axes, 0);
+
```
nornagon, line 24: This doesn't seem _super_ upstreamable. When will we know when we can remove this?

patches/chromium/fix_check_issecureeventinputenabled_in_constructor_before_setting.patch
```diff
@@ -0,0 +1,21 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: John Kleinschmidt <jkleinsc@github.com>
+Date: Thu, 9 Jul 2020 11:16:30 -0400
+Subject: fix: check IsSecureEventInputEnabled in constructor before setting
+ SetPasswordInputEnabled to true
+
+It appears that macOS 12 has IsSecureEventInputEnabled set to true initially.
```
nornagon, line 7: this is upstreamed already, right?

patches/v8/chore_add_v8_apple_silicon_patches.patch
```diff
@@ -0,0 +1,134 @@
...
+--- a/src/codegen/arm64/macro-assembler-arm64.cc
++++ b/src/codegen/arm64/macro-assembler-arm64.cc
+@@ -2973,6 +2973,35 @@ void TurboAssembler::PrintfNoPreserve(const char* format,
+ 
+   int arg_count = kPrintfMaxArgCount;
+ 
++#if V8_OS_MACOSX && !USE_SIMULATOR
++  CPURegList tmp_list = kCallerSaved;
++  tmp_list.Remove(x0);      // Used to pass the format string.
++  tmp_list.Remove(arg0, arg1, arg2, arg3);
```
nornagon, line 19: extremely terrifying patch

script/spec-runner.js
```diff
@@ -225,6 +225,9 @@ async function installSpecModules (dir) {
     npm_config_nodedir: nodeDir,
     npm_config_msvs_version: '2019'
   });
+  if (fs.existsSync(path.resolve(dir, 'node_modules'))) {
+    await fs.remove(path.resolve(dir, 'node_modules'));
+  }
```
nornagon, line 230: big oof

patches/chromium/fix_check_issecureeventinputenabled_in_constructor_before_setting.patch
jkleinsc: Yes - https://chromium-review.googlesource.com/c/chromium/src/+/2295749

.circleci/config.yml
MarshallOfSound: This build was unused and completely unreferenced

.circleci/config.yml
MarshallOfSound: Yes but not right now

patches/chromium/fix_check_issecureeventinputenabled_in_constructor_before_setting.patch
MarshallOfSound: Yes but we haven't absorbed it yet

patches/chromium/chore_disable_failing_blink_font_dcheck_on_big_sur.patch
MarshallOfSound: Chrome are aware of the failure on Big Sur and are working themselves on fixing it

script/spec-runner.js
MarshallOfSound: We can remove this once my `node-gyp` changes are absorbed upstream

patches/v8/chore_add_v8_apple_silicon_patches.patch
MarshallOfSound: V8 is mostly terrifying

PR #24559, opened 2020-07-15
chore: cherry-pick fix from chromium issue 1074317

zcbenz:
[1074317] [High] [CVE-2020-6511]: Security: The CSP reports and stacktraces of errors leaks post-redirect URL for <script>
Backport https://chromium.googlesource.com/chromium/src/+/0b707cbaa2cb806162797be55caf9f8074fbdccf

Notes: fix: remove leaks of post-redirect URL for `<script>`  in the CSP reports and stacktraces of errors (Chromium security issue 1074317)


deepak1556:
The CL indicates a change in behavior that might break some users, should that be explicitly called out in the release notes ?

Issue #24577, opened 2020-07-15
assert.ok(false) throws TypeError instead of AssertionError


ianhattendorf:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 8.0.0 (also tested with `electron-nightly@11.0.0-nightly.20200708`)
* **Operating System:**
  * macOS 10.15.5
* **Last Known Working Electron version:**
  * 7.3.2

### Expected Behavior
```
$ yarn add electron@8.0.0
$ node_modules/electron/dist/Electron.app/Contents/MacOS/Electron -i
> require('assert').ok(false)
Thrown:
AssertionError [ERR_ASSERTION]: false == true
    at repl:1:19
    at Script.runInContext (vm.js:137:20)
    at REPLServer.defaultEval (repl.js:386:29)
    at bound (domain.js:420:14)
    at REPLServer.runBound [as eval] (domain.js:433:12)
    at REPLServer.onLine (repl.js:700:10)
    at REPLServer.emit (events.js:203:13)
    at REPLServer.EventEmitter.emit (domain.js:476:20)
    at REPLServer.Interface._onLine (readline.js:316:10)
    at REPLServer.Interface._line (readline.js:693:8) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}
```

### Actual Behavior
```
$ yarn add electron@8.0.0
$ node_modules/electron/dist/Electron.app/Contents/MacOS/Electron -i
> require('assert').ok(false)
Thrown:
TypeError: call.getFileName is not a function
    at getErrMessage (assert.js:270:25)
    at innerOk (assert.js:361:17)
    at Function.ok (assert.js:381:3)
    at repl:1:19
    at Script.runInContext (vm.js:127:20)
    at REPLServer.defaultEval (repl.js:406:29)
    at bound (domain.js:420:14)
    at REPLServer.runBound [as eval] (domain.js:433:12)
    at REPLServer.onLine (repl.js:715:10)
    at REPLServer.emit (events.js:210:5)
```
### To Reproduce
See expected/actual behavior.

[...]

PR #24579, opened 2020-07-16
fix: use Node.js isolate setup logic in bindings

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/24577.

This fixes an issue whereby any code in Node.js that invokes various assert functionality in the browser or renderer processes will throw with `TypeError: call.getFileName is not a function`. The reason this happens is that Node.js overrides the Promise rejection callback used by the Isolate, so the `WeakMap` used in their assert logic is never populated.

To remedy this, the best long-term solution, I feel, is to better align our isolate setup functionality with that available to us via the embedder-focused `SetIsolateUpForNode`. We already approximated a lot of this functionality, so this allows us to remove that. It does, however create a small new issue wherein we do not want to use the promise rejection callback that Node.js uses in the renderer process, because it does not send PromiseRejectionEvents to the global script context. This causes code like:

```
window.addEventListener('unhandledrejection', function (event) {
  // ...your code here to handle the unhandled rejection...

  // Prevent the default handling (such as outputting the
  // error to the console)

  event.preventDefault();
});
```

to fail since the `'unhandledrejection'` rejection event would never be fired. We need thus to use the one Blink already provides, and to augment `IsolateSettings` accordingly.

I've upstreamed this at https://github.com/nodejs/node/pull/34387.

cc @nornagon @zcbenz @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where many uses of the Node.js `assert` module would throw in both the browser and renderer processes.


deepak1556:
Couple of question,

* >The reason this happens is that Node.js overrides the Promise rejection callback used by the Isolate,

AFAIK in the browser process we setup the callbacks in https://github.com/electron/electron/blob/master/patches/node/feat_add_flags_for_low-level_hooks_and_exceptions.patch as `g_standalone_mode` and in the renderer blink does it. Where does node override here ? Maybe the problem/solution statement is a bit confusing, can you please elaborate whats happening ?

* Also there is no blink in the browser process, so why are those isolate settings being removed for the browser process too ?

* We setup our own fatal error handlers for all the process in https://github.com/electron/electron/blob/master/shell/common/api/electron_bindings.cc#L89 , now will that be overriden by the handler from `SetIsolateUpForNode` ?

codebytere:
@deepak1556 you're right - we can remove a lot of that low-level flag logic now, as it's just being redundantly set twice. 
I've also moved the flags into process-specific locations. With regard to:

> AFAIK in the browser process we setup the callbacks in https://github.com/electron/electron/blob/master/patches/node/feat_add_flags_for_low-level_hooks_and_exceptions.patch as g_standalone_mode and in the renderer blink does it

I removed the invocation in the patch, so it's now just disabled for the renderer process where Blink does it.

> We setup our own fatal error handlers for all the process in https://github.com/electron/electron/blob/master/shell/common/api/electron_bindings.cc#L89 , now will that be overriden by the handler from SetIsolateUpForNode ?

Yes - we can remove this now as the embedder interface handles it.

MarshallOfSound:
I think we should avoid letting node control blinks isolate, can't we fix the `assert` module to work without nodes hooks?  Isolate wide modifications are bad in the renderer as not every context in the given isolate are actually node contexts.  So this can have unexpected effects / context leakage in the renderer process.

codebytere:
The embedder interface this change leverages, `SetIsolateUpForNode`, calls 8 discrete pieces of isolate functionality:

```cpp
isolate->AddMessageListenerWithErrorLevel( errors::PerIsolateMessageListenerg);
isolate->SetAbortOnUncaughtExceptionCallback(abort_callback);
isolate->SetFatalErrorHandler(fatal_error_cb);
isolate->SetPrepareStackTraceCallback(prepare_stack_trace_cb);
isolate->SetMicrotasksPolicy(s.policy);
isolate->SetAllowWasmCodeGenerationCallback(allow_wasm_codegen_cb);
v8::CpuProfiler::UseDetailedSourcePositionsForProfiling(isolate);
```

Renderer Isolate Setup Prior to Change
```
 isolate->SetMicrotasksPolicy(s.policy);
 isolate->SetAllowWasmCodeGenerationCallback(allow_wasm_codegen_cb);
 v8::CpuProfiler::UseDetailedSourcePositionsForProfiling(isolate);
 isolate->AddMessageListenerWithErrorLevel( errors::PerIsolateMessageListeners);
```

Renderer Isolate Setup After Change
```cpp
isolate->SetMicrotasksPolicy(s.policy);
isolate->SetAllowWasmCodeGenerationCallback(allow_wasm_codegen_cb);
v8::CpuProfiler::UseDetailedSourcePositionsForProfiling(isolate);
isolate->AddMessageListenerWithErrorLevel( errors::PerIsolateMessageListeners);
isolate->SetAbortOnUncaughtExceptionCallback(abort_callback);
isolate->SetFatalErrorHandler(fatal_error_cb);
```

That leaves the difference consequent to this PR as:
```cpp
 isolate->SetAbortOnUncaughtExceptionCallback(abort_callback);
isolate->SetFatalErrorHandler(fatal_error_cb);  
```

We choose to call `isolate->SetFatalErrorHandler(fatal_error_cb)` ourselves in Node bindings and bind it to the following function:
```cpp
 void FatalErrorCallback(const char* location, const char* message) {
  LOG(ERROR) << "Fatal error in V8: " << location << " " << message;
  ElectronBindings::Crash();
} 
```

  Which compares with the one we’d now be invoking with Node.js:

```cpp
 void OnFatalError(const char* location, const char* message) {
  if (location) {
    FPrintF(stderr, "FATAL ERROR: %s %s\n", location, message);
  } else {
    FPrintF(stderr, "FATAL ERROR: %s\n", message);
  }

  Isolate* isolate = Isolate::GetCurrent();
  Environment* env = Environment::GetCurrent(isolate);
  bool report_on_fatalerror;
  {
    Mutex::ScopedLock lock(node::per_process::cli_options_mutex);
    report_on_fatalerror = per_process::cli_options->report_on_fatalerror;
  }

  if (report_on_fatalerror) {
    report::TriggerNodeReport(
        isolate, env, message, "FatalError", "", Local<Object>());
  }

  fflush(stderr);
  ABORT();
}
```

  Stripping out cli functionality we _already_ prevent in the renderer process: 
```cpp 
void OnFatalError(const char* location, const char* message) {
  if (location) {
    FPrintF(stderr, "FATAL ERROR: %s %s\n", location, message);
  } else {
    FPrintF(stderr, "FATAL ERROR: %s\n", message);
  }

  fflush(stderr);
  ABORT();
}
```

  This introduces no side effects and is a comparable output, I would say.

Lastly, we’re left with: `isolate->SetAbortOnUncaughtExceptionCallback(abort_callback);`

 Blink currently [never calls this](https://source.chromium.org/search?q=SetAbortOnUncaughtExceptionCallback&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2Fsearch%2F).  This callback is used explicitly by embedders to help V8 determine if it should abort when it throws and no internal handler is predicted to catch the exception.

Node.js currently chooses to do this if:
 1) The user has passed the `—abort-on-uncaught-exception` flag, which is irrelevant to is in the renderer as we disallow it 
2) The user is in a scope where uncaught exceptions should not result in an abort; this only applies in the context of the Node.js ESM loader which is also already irrelevant in a renderer context as we disallow it in the renderer process as of [#24301](https://github.com/electron/electron/pull/24301). 

The only other case resulting in an abort in Node.js’ callback is when the Environment itself is stopping. As such, I personally feel comfortable allowing this setup functionality to run, but I would definitely be open to perhaps not choosing to do this and instead upstreaming a toggle to prevent this to Node.js.  In the net, though, the comparable surface area of difference is small enough, and the net benefit great enough of using embedding functionality provided to us, that I would ideally prefer to utilize it where possible.

PR #24592, opened 2020-07-16
chore: cherry-pick fix from chromium issue 1072165

zcbenz:
[[1072165](https://crbug.com/1072165)] [**Medium**]: libjingle_xmpp_xmlparser_fuzzer: Incorrect-function-pointer-type with empty stacktrace
Backport https://chromium.googlesource.com/chromium/src/+/34f7d701cd3a21f455a6fbccff366b6eb184fb7c

Notes: fix: Incorrect-function-pointer-type with empty stacktrace. (Chromium security issue 1072165)

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -130,3 +130,4 @@ backport_1063177.patch
 backport_1065122.patch
 backport_1074317.patch
 backport_1090543.patch
+backport_1072165.patch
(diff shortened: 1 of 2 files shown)


zcbenz:
> We don't have fuzzer tests running, is this required ?

No we don't actually need this, thanks for checking.

PR #24598, opened 2020-07-16
chore: cherry-pick fix from chromium issue 1052492

zcbenz:
[[1052492](https://crbug.com/1052492)] [**Medium**]: Use-of-uninitialized-value in blink::ImageDataBuffer::ImageDataBuffer
Backport https://chromium.googlesource.com/chromium/src/+/bad271a777cf623ddcd6908e9de6ec5bbcd51f6a

Notes: fix: use-of-uninitialized-value in blink::ImageDataBuffer::ImageDataBuffer. (Chromium security issue 1052492)

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -117,3 +117,4 @@ avoid_using_x11_shm_for_remote_connections.patch
 backport_1065122.patch
 backport_1074317.patch
 backport_1090543.patch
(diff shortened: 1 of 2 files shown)


zcbenz:
> Considering this is behind MSAN flag the changes won't be present in any of our builds, does this need a backport ?

You are right this should not need a backport.

PR #24603, opened 2020-07-16
refactor: clean up Session with CleanedUpAtExit

nornagon:
#### Description of Change
Use the CleanedUpAtExit gin helper to clean up sessions at exit.

Also, CleanedUpAtExit now deletes things in the reverse order of their creation.

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/javascript_environment.cc
+++ b/shell/browser/javascript_environment.cc
@@ -46,7 +46,6 @@ JavascriptEnvironment::~JavascriptEnvironment() {
   {
     v8::Locker locker(isolate_);
     v8::HandleScope scope(isolate_);
-    gin_helper::CleanedUpAtExit::DoCleanup();
     context_.Get(isolate_)->Exit();
   }
   isolate_->Exit();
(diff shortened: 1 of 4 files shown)


shell/browser/javascript_environment.cc
```diff
@@ -100,6 +99,11 @@ void JavascriptEnvironment::OnMessageLoopDestroying() {
   base::MessageLoopCurrent::Get()->RemoveTaskObserver(microtasks_runner_.get());
   platform_->DrainTasks(isolate_);
   platform_->UnregisterIsolate(isolate_);
+  {
+    v8::Locker locker(isolate_);
+    v8::HandleScope scope(isolate_);
+    gin_helper::CleanedUpAtExit::DoCleanup();
```
deepak1556, line 105: Should this block be moved up to the top of this function, so that it gets the benefit of microtask runner and the v8::platform draining the event loop ?

Considering this block was handled in the destructor before and nothing seemed to break, it should be fine as it is,

PR #24604, opened 2020-07-16
refactor: use gin::Arguments in WebContents

nornagon:
#### Description of Change
In preparation for converting WebContents to gin.

Notes: none


shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2341,11 +2346,16 @@ void WebContents::SendInputEvent(v8::Isolate* isolate,
...
 
-void WebContents::BeginFrameSubscription(gin_helper::Arguments* args) {
+void WebContents::BeginFrameSubscription(gin::Arguments* args) {
   bool only_dirty = false;
   FrameSubscriber::FrameCaptureCallback callback;
 
-  args->GetNext(&only_dirty);
+  if (args->Length() > 1) {
+    if (!args->GetNext(&only_dirty)) {
+      args->ThrowError();
```
deepak1556, line 2355: Descriptive error message ?

nornagon: I think this will show "conversion failure at index 0" or something, which is consistent with other APIs (if not exactly the most useful)

PR #24607, opened 2020-07-16
build: free up space on macOS VM in background

MarshallOfSound:
The delete takes ~2-3 minutes and there is no reason to not do anything while we wait, most builds download a cache directly after this command or start building.  We should be able to do things at the same time and save that 2-3 minutes by not blocking for the `rm` commands

Notes: no-notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -312,17 +312,23 @@ step-get-more-space-on-mac: &step-get-more-space-on-mac
     name: Free up space on MacOS
     command: |
       if [ "`uname`" == "Darwin" ]; then
-        sudo rm -rf /Library/Developer/CoreSimulator
-        sudo rm -rf $(xcode-select -p)/Platforms/AppleTVOS.platform
-        sudo rm -rf $(xcode-select -p)/Platforms/iPhoneOS.platform
-        sudo rm -rf $(xcode-select -p)/Platforms/WatchOS.platform
-        sudo rm -rf $(xcode-select -p)/Platforms/WatchSimulator.platform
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -312,17 +312,22 @@ step-get-more-space-on-mac: &step-get-more-space-on-mac
...
+        tmpify $(xcode-select -p)/Platforms/WatchOS.platform
+        tmpify $(xcode-select -p)/Platforms/WatchSimulator.platform
+        tmpify $(xcode-select -p)/Platforms/AppleTVSimulator.platform
+        tmpify $(xcode-select -p)/Platforms/iPhoneSimulator.platform
+        tmpify ~/.rubies
+        tmpify ~/Library/Caches/Homebrew
+        tmpify /usr/local/Homebrew
+        sudo rm -rf $TMPDIR/del-target
       fi
+    background: true
```
ckerr, line 330: This may be overkill but I do want to ask, is there any risk at having these directories' locations in an indeterminate state, e.g. maybe not all these are done being `mv`ed by the time the caller moves on to the next step because the `mv` is happening in the background?

Maybe this should be broken into a foreground step that does the `tmpify()` calls, and then a background one that blows away del-target afterwards?

MarshallOfSound: The `mv` calls on macOS are basically instantaneous.  The reason I had to introduce `tmpify` is a task that occurred a minute or two after this job got confused by a half-existing developer platform.  I think it's safe to say it's not possible for the problematic task to spawn quick enough that the `mv` tasks have not completed 🤔

Issue #23980, opened 2020-06-05
Bad window resize performance in electron 9.0.0


crazytoucan:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 9.0.0
* **Operating System:**
  * macOS 10.14.6
* **Last Known Working Electron version:**
  * 8.x

### Expected Behavior

Resizing a window is smooth

### Actual Behavior

Resizing a window is unexpectedly laggy, even when the window is completely empty and there are no other active listeners

### To Reproduce

The following simple program exhibits the poor behavior:

```sh
$ electron index.js
```

```js
// index.js
const BrowserWindow = require("electron").BrowserWindow;
const app = require("electron").app;

app.on("ready", () => {
  new BrowserWindow();
})
```

### Screenshots

![slow-resize](https://user-images.githubusercontent.com/3346671/83829285-5f750b00-a697-11ea-849f-a4abe1dfc609.gif)

### Additional Information

Notes:

- To the best of my knowledge, this worked well in previous versions of Electron
- Interestingly, at super tiny sizes, like making the window 80px x 60px, resizing works just fine. The resizing performance degrades _significantly_ starting at about 600 x 400.

Thanks much for such an awesome technology. Let me know if there's anything I can do to help diagnose potential confounding issues here!

nornagon:
I can repro. Looks like it was introduced with the first beta of v9, so it's likely related to a Chromium roll.

johnsusek:
Also seeing this on MacOS 10.15.3, iMac Pro

zcbenz:
Closed with https://github.com/electron/electron/pull/24642.

Issue #10886, opened 2017-10-23
Frameless Vibrant windows have square corners (no border radius)


natew:
* Electron version: 1.8.0
* Operating system: MacOS Sierra / High Sierra

### Expected behavior

Some sort of control over this would be nice.

### Actual behavior

Only square windows show.

### More

Keeping this ticket purposely narrow. In the ideal you'd have control over regions where it shows + the border radius. Or really ideal would be just to have backdrop-filter work through to the desktop (perhaps configurable). But, baby steps...

I noticed that the original PR for vibrancy [seems to show a screenshot](https://github.com/electron/electron/pull/7898#issuecomment-259158609) that implies it was rounded at one point, but for me I see this:

<img width="329" alt="screen shot 2017-10-23 at 11 04 14 am" src="https://user-images.githubusercontent.com/12100/31905109-ee246370-b7e1-11e7-86b6-68707674f73e.png">

felixrieseberg:
Hm, I tried to replicate the behavior but was unable to do so with the latest master (running macOS Sierra). Do you have a solid repro?

<img width="987" alt="screen shot 2017-10-24 at 4 38 46 pm" src="https://user-images.githubusercontent.com/1426799/31973259-da0a636e-b8d9-11e7-86e6-f90e8b51d000.png">

sandor:
Guys, getting something similar here: https://github.com/electron/electron/issues/10823

sandor:
If you need a working example: https://github.com/sandor/floido-designer

natew:
@sandor I actually don't see the light/darkbackground behavior, it's sharp cornered on any background:

![image](https://user-images.githubusercontent.com/12100/33108933-aa5d2c16-cef3-11e7-808c-3b08c4e2b7d1.png)

sandor:
@natew I think if you make the background of your header transparent you will get the same results (light/dark/bg) but can not test it here now – sorry...

sandor:
Hey Guys, just saw this in my terminal (think this appears since I updated to the latest beta). To be honest I am not experienced enough to see what this is implying, but it seems to have some relavanve to the behaviours described in this thred...

`2017-11-24 13:55:57.923 Electron[7973:767197] *** WARNING: Textured window <AtomNSWindow: 0x7fe68d86a8b0> is getting an implicitly transparent titlebar. This will break when linking against newer SDKs. Use NSWindow's -titlebarAppearsTransparent=YES instead.`

okonet:
[...]

PR #24642, opened 2020-07-19
fix: remove corner mask override to increase window resize performance

gerhardberger:
#### Description of Change

This PR addresses issue #23980 that has been around since electron 9.0.0, when the override of an internal method on NSWindow (to have smooth corners on a vibrant window), caused very bad/laggy resize performance for the whole window, even if vibrancy not applied.

Removing this produced still smooth corners on macOS 10.15.5, so probably it's already obsolete.

#### Checklist

#### Release Notes

Notes: fix: remove unnecessary corner mask overriding to increase window resize performance

### DIFF
--- a/shell/browser/ui/cocoa/electron_ns_window.mm
+++ b/shell/browser/ui/cocoa/electron_ns_window.mm
@@ -24,7 +24,6 @@ bool ScopedDisableResize::disable_resize_ = false;
 @synthesize disableAutoHideCursor;
 @synthesize disableKeyOrMainWindow;
 @synthesize vibrantView;
-@synthesize cornerMask;
 
 - (id)initWithShell:(electron::NativeWindowMac*)shell
           styleMask:(NSUInteger)styleMask {
@@ -161,12 +160,6 @@ bool ScopedDisableResize::disable_resize_ = false;
 
 - (NSView*)frameView {
   return [[self contentView] superview];
-}
-
-// By overriding this built-in method the corners of the vibrant view (if set)
-// will be smooth.
-- (NSImage*)_cornerMask {
-  return [self cornerMask];
 }
 
 // Quicklook methods
--- a/shell/browser/ui/cocoa/electron_ns_window.mm
+++ b/shell/browser/ui/cocoa/electron_ns_window.mm
@@ -17,6 +17,10 @@ bool ScopedDisableResize::disable_resize_ = false;
 
 }  // namespace electron
 
+@interface NSWindow (PrivateAPI)
+- (NSImage*)_cornerMask;
+@end
+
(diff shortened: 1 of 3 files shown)


shell/browser/ui/cocoa/electron_ns_window.mm
```diff
@@ -163,12 +162,6 @@ - (NSView*)frameView {
   return [[self contentView] superview];
 }
 
-// By overriding this built-in method the corners of the vibrant view (if set)
-// will be smooth.
-- (NSImage*)_cornerMask {
-  return [self cornerMask];
```
zcbenz, line 169: I'm not sure if removing this would cause side effects on older macOS versions, could something like this solve the problem?

```objective-c
- (NSImage*)_cornerMask {
  if ([self hasVibrantView])
    return [self cornerMask];
  else
    return [super _cornerMask];
}
```

gerhardberger: yea i'm not sure as well what could this cause on older macOS versions, but calling `[super _cornerMask]` for me results in an error:

```
../../electron/shell/browser/ui/cocoa/electron_ns_window.mm:182:17: error: 'NativeWidgetMacNSWindow' may not respond to '_cornerMask' [-Werror]
  return [super _cornerMask];
          ~~~~~ ^
1 error generated.
```

and if i just return `nil` always in that method the resize has the same lagginess.

zcbenz: You can declare the API like this:

```objective-c
@interface NSWindow (PrivateAPI)
- (NSImage*)_cornerMask;
@end
```

gerhardberger: with this it works, so we could do that above, the question is, having smooth corners are better than not having poor resize performance in vibrancy mode?

zcbenz: I have tried removing `_cornerMask` on 10.14 and the corners are still smooth, and looking at original issue https://github.com/electron/electron/issues/10886 this might only be a problem for macOS <= 10.13.

So how about only doing `_cornerMask` hack when macOS version <= 10.13?

gerhardberger: actually I realised now that if I only return our cornerMask for vibrant view, the resize performance is the same as calling the base class implementation, so possibly it will work well on all platforms this way

msavin:
Would the corner fix be difficult to bring to Electron 8?

zcbenz:
@msavin The problem being solved by this PR does not exist in 8.

PR #24655, opened 2020-07-20
fix: Allow VoiceOver to navigate "back into" web contents

felixrieseberg:
#### Description of Change

After some consultation with our friends at Apple, I've figured out why exactly VoiceOver has such trouble getting "back into" web contents when it was first navigated out. As it turns out, by default, there is no `RenderWidgetHostViewCocoa` in the AXChildren array, so our custom filter ensuring that it's included doesn't do anything. Instead, the web contents are part of a class named `BrowserAccessibilityCocoa`.

[Note: I've previously tried to fix this by manually adding the `contentView`, which worked in some cases, but not all]

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixes an issue where VoiceOver was unable to navigate from the top-level window back into the web contents

### DIFF
--- a/shell/browser/ui/cocoa/electron_ns_window.mm
+++ b/shell/browser/ui/cocoa/electron_ns_window.mm
@@ -130,7 +130,7 @@ bool ScopedDisableResize::disable_resize_ = false;
   // the window title instead of using Cmd+C to get the selected text.
   NSPredicate* predicate = [NSPredicate
       predicateWithFormat:@"(self isKindOfClass: %@) OR (self.className == %@)",
-                          [NSButtonCell class], @"ElectronAdaptedContentView"];
+                          [NSButtonCell class], @"BrowserAccessibilityCocoa"];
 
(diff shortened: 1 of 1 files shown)


codebytere:
@felixrieseberg do you by chance have a reference for this or are we just sort of trusting manual testing and what Apple says? i'm ok with this but have mild concern it might change again in the future and would love to know where we can look for this kind of information 🤔

felixrieseberg:
@codebytere God, I wish I had a good answer for you - that would have made this a lot easier. This PR is the result of some manual testing by Slack's a11y people and an email thread with Apple where some helpful AX experts told me to please fix the parent-child relationship. 

Actually, as I'm writing this out, we'd probably be better off _removing_ the thing we don't want in the array instead of manually keeping things. Let me amend the PR, because that'd solve our problem here - if things change again in the future, we wouldn't have to change anything in this code.

PR #24659, opened 2020-07-21
fix: export libuv symbols

### LINKED ISSUE #24428: undefined symbol error when importing a native module that uses `uv_dlopen`
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 9.0.5
* **Operating System:**
  * Ubuntu 18.04 x64

### Expected Behavior
Able to import and use a native module that references `uv_dlopen` and other related functions.

### Actual Behavior
Importing a native module that references `uv_dlopen`, `uv_dlsym` and other related functions causes an undefined symbol error.

Error:
```bash
/home/teokp/source/electron-uv-test/node_modules/electron/dist/electron test.js: symbol lookup error: /home/teokp/source/electron-uv-test/build/Release/uv_module.node: undefined symbol: uv_dlopen
```

### To Reproduce
```bash
git clone https://github.com/koonpeng/electron-uv-test.git
npm install
npx electron test.js
```

### Additional Information
Comparing the exported symbols of node and electron,

node v12.18.1
```bash
$ nm -D $(which node) | grep uv_dl
0000000001332420 T uv_dlclose
00000000013324d0 T uv_dlerror
00000000013323a0 T uv_dlopen
0000000001332460 T uv_dlsym
```

electron 9.0.5
```bash
$ nm -D node_modules/electron/dist/electron | grep uv_dl
<empty>
```

It appears that some symbols of libuv are not exported, I tried building a custom version of electron 9.0.5 with these changes

```diff
diff --git a/BUILD.gn b/BUILD.gn
index f70dbcba1..e46592144 100644
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -1143,7 +1143,7 @@ if (is_mac) {
       ]
     }
     if (is_linux) {
-      ldflags = [ "-pie" ]
+      ldflags = [ "-pie", "-Wl,--whole-archive", "obj/third_party/electron_node/deps/uv/libuv.a", "-Wl,--no-whole-archive" ]
 
       if (!is_component_build && is_component_ffmpeg) {
         configs += [ "//build/config/gcc:rpath_for_built_shared_libraries" ]
```

to force the missing symbols to be exported, which does appears to fix the error.

koonpeng:
#### Description of Change

Forces libuv symbols to be exported, this fixes #24428.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed undefined symbol error when loading native modules that uses `uv_dlopen`

### DIFF
--- a/spec-main/yarn.lock
+++ b/spec-main/yarn.lock
@@ -193,11 +193,6 @@ nise@^4.0.1:
     just-extend "^4.0.2"
     path-to-regexp "^1.7.0"
 
-node-addon-api@^3.0.0:
-  version "3.0.0"
-  resolved "https://registry.yarnpkg.com/node-addon-api/-/node-addon-api-3.0.0.tgz#812446a1001a54f71663bed188314bba07e09247"
-  integrity sha512-sSHCgWfJ+Lui/u+0msF3oyCgvdkhxDbkCS6Q8uiJquzOimkJBvX6hl5aSSA7DR1XbMpdM8r7phjcF63sF4rkKg==
-
 node-ensure@^0.0.0:
   version "0.0.0"
   resolved "https://registry.yarnpkg.com/node-ensure/-/node-ensure-0.0.0.tgz#ecae764150de99861ec5c810fd5d096b183932a7"
@@ -273,10 +268,8 @@ uri-js@^4.2.2:
   dependencies:
     punycode "^2.1.0"
 
-"uv-dlopen@file:fixtures/native-addon/uv-dlopen":
+uv-dlopen@./fixtures/native-addon/uv-dlopen/:
   version "0.0.1"
-  dependencies:
-    node-addon-api "^3.0.0"
 
 worker-loader@^2.0.0:
   version "2.0.0"
--- a/spec-main/yarn.lock
+++ b/spec-main/yarn.lock
@@ -193,6 +193,11 @@ nise@^4.0.1:
     just-extend "^4.0.2"
     path-to-regexp "^1.7.0"
 
+node-addon-api@^3.0.0:
+  version "3.0.0"
+  resolved "https://registry.yarnpkg.com/node-addon-api/-/node-addon-api-3.0.0.tgz#812446a1001a54f71663bed188314bba07e09247"
+  integrity sha512-sSHCgWfJ+Lui/u+0msF3oyCgvdkhxDbkCS6Q8uiJquzOimkJBvX6hl5aSSA7DR1XbMpdM8r7phjcF63sF4rkKg==
+
 node-ensure@^0.0.0:
   version "0.0.0"
   resolved "https://registry.yarnpkg.com/node-ensure/-/node-ensure-0.0.0.tgz#ecae764150de99861ec5c810fd5d096b183932a7"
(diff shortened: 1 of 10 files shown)


koonpeng:
Thanks for the quick respond! For context, I got an undefined symbol error when trying to load a native module that works in node, the reason I found is because node exports a few extra symbols that electron is missing.

I followed the docs and stripped the symbols made a patched release build here https://github.com/koonpeng/electron/releases/tag/v9.0.5. Comparing that to the official release,

```sh
teokp@teokp-Desktop:~/Downloads$ nm -D ./electron-v9.0.5-linux-x64/electron | awk '{print $3}' > default.txt 
teokp@teokp-Desktop:~/Downloads$ nm -D ./electron-v9.0.5-linux-x64-koonpeng/electron | awk '{print $3}' > patched.txt
teokp@teokp-Desktop:~/Downloads$ git diff default.txt patched.txt
diff --git a/default.txt b/patched.txt
index c188ec7..29e92b7 100644
--- a/default.txt
+++ b/patched.txt
@@ -1589,6 +1589,10 @@ uv_cpu_info
 uv_cwd
 uv_default_loop
 uv_disable_stdio_inheritance
+uv_dlclose
+uv_dlerror
+uv_dlopen
+uv_dlsym
 uv_err_name
 uv_err_name_r
 uv_exepath
```

It seems that only 4 more libuv symbols are added, filesize remains pretty much identical. I'm not very sure if this is the correct way to fix it so I'm happy to hear other suggestions.

koonpeng:
I explored abit further on this, it looks like node's build script also uses `--whole-archive` https://github.com/nodejs/node/blob/b0b52b2023f5cd0df4ae921850815586b4313dca/node.gypi#L176. This flag is used when `'force_load=="true"'`. `force_load` is only set when building as an executable, I think electron's node is built as a static library so it is correct that it doesn't use the flag.

I'm not sure how to go about this, in node's docs it says that libuv is purposefully re-exported and should be usable by native modules. If electron aims to support native modules, should it add `--whole-archive` even though it's building node as a static library?
```
Node.js includes other statically linked libraries including OpenSSL. These other libraries are located in the deps/ directory in the Node.js source tree. Only the libuv, OpenSSL, V8 and zlib symbols are purposefully re-exported by Node.js and may be used to various extents by Addons. See Linking to libraries included with Node.js for additional information.
```

koonpeng:
node uses `--whole-archive` to export all the symbols in libuv, I think electron should follow it instead of explicitly adding certain symbols with `-u`?

Actually the problem doesn't happen in Windows because libuv is built as a shared library, I wasn't able to test on mac so I don't know if it works. I have a test project here https://github.com/koonpeng/electron-uv-test. I can try adding a test case that use that module and see if it works on the CI, in the meantime I will convert this to a draft.

zcbenz:
Thanks for providing the test project, I confirm that this problem also happens on macOS.

I'm good with the `--whole-archive` solution if you can add the flag for both Linux and macOS. I think it is fair use here since the lack of symbols is predictable, and exporting all symbols of libuv is expected behavior for users.

koonpeng:
I tried using force_load and whole-archive flags like in node but I'm still getting symbol errors in the mac tests. It's hard for me to test without a mac.

zcbenz:
Thanks for adding the test, I have added a commit to make it work on mac, it turns out we should use the `-force_load` flag there.

codebytere:
@koonpeng can you please rebase this on master?

spec-main/yarn.lock
```diff
@@ -193,6 +193,11 @@ nise@^4.0.1:
     just-extend "^4.0.2"
     path-to-regexp "^1.7.0"
 
+node-addon-api@^3.0.0:
```
nornagon, line 196: This is a fairly simple module; could we use the C API of napi directly rather than adding a dependency?

koonpeng: yes, i actually started off with the c api before finding out there is c++ extensions.

codebytere:
Failure is known on macOS - will merge when the Linux rerun passes.

zcbenz:
The failing mac test is unrelated.

Issue #24319, opened 2020-06-26
Apple Silicon / macOS Big Sur Support


MarshallOfSound:
I'm opening this issue so that when people search for it, they find this 😄  We are currently reviewing the work that needs to be done across Electron, the ecosystem and our dependencies to provide support for Apple Silicon and macOS 11.  Once we have a good picture of the work that needs to be done we will raise tracking issues for that work or just update this issue.

You should follow this issue for updates, don't raise new ones asking for updates or information because they will simply be closed.

MarshallOfSound:
### Known Issues

* [ ] Corners aren't rounded properly on most windows, this appears to be true across all applications and is therefore probably a Big Sur bug
* [ ] Electron doesn't run under Rosetta 2 (no chromium based things do).  We should set `LSRequiresNativeExecution` to indicate this to the OS.  It might be possible to fix using a determined-at-runtime page size but that might have an untold performance cost

MarshallOfSound:
Closing this tracking issue out, for information on how to get your app running on Apple silicon please refer to our blog post.

https://www.electronjs.org/blog/apple-silicon

If you find any Apple silicon specific issues, please raise them as new issues on this repository.

PR #24670, opened 2020-07-21
fix: disable rosetta as Electron does not run under rosetta

MarshallOfSound:
Electron crashes under rosetta, we should prevent execution under rosetta.

Refs: #24319  

Notes: no-notes

### DIFF
--- a/shell/app/resources/mac/loginhelper-Info.plist
+++ b/shell/app/resources/mac/loginhelper-Info.plist
@@ -12,5 +12,7 @@
 	<string>APPL</string>
 	<key>LSBackgroundOnly</key>
 	<true/>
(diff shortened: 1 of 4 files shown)


deepak1556:
considering upstream now looking into this https://bugs.chromium.org/p/chromium/issues/detail?id=1102194#c15 , will we revisit this decision if there is a fix ?

MarshallOfSound:
@deepak1556 Yes definitely, for now though versions we publish won't and will never work under Rosetta so we should flag them as such.  Once things start working under Rosetta we can remove this flag.  This is purely a UX thing 😄

deepak1556:
cool, sgtm 👍

PR #24671, opened 2020-07-21
perf: do not convert object keys in ctx bridge as they are always primitives

MarshallOfSound:
Another performance improvement for the ctx bridge based on the work in #24531

Object keys by spec are either strings, numbers or symbols (numbers can be used interchangeably with their string representations).  These are all primitives and therefore we do not need to convert / clone them.  For object with lots of keys this is another quick win.

Notes: Improved the performance of sending wide objects over the context bridge

### DIFF
--- a/shell/renderer/api/electron_api_context_bridge.cc
+++ b/shell/renderer/api/electron_api_context_bridge.cc
@@ -470,7 +470,7 @@ v8::MaybeLocal<v8::Object> CreateProxyForAPI(
 
             v8::PropertyDescriptor desc(getter_proxy, setter_proxy);
             ignore_result(proxy.GetHandle()->DefineProperty(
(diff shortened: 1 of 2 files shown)


shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -476,22 +470,21 @@ v8::MaybeLocal<v8::Object> CreateProxyForAPI(
 
             v8::PropertyDescriptor desc(getter_proxy, setter_proxy);
             ignore_result(proxy.GetHandle()->DefineProperty(
-                destination_context, gin::StringToV8(api.isolate(), key_str),
-                desc));
+                destination_context, v8::Local<v8::Name>::Cast(key), desc));
```
nornagon, line 473:
```suggestion
                destination_context, key.As<v8::Name>(), desc));
```

PR #24675, opened 2020-07-21
refactor: ginify SystemPreferences

nornagon:
Notes: none

### DIFF
--- a/shell/browser/api/electron_api_system_preferences_mac.mm
+++ b/shell/browser/api/electron_api_system_preferences_mac.mm
@@ -125,7 +125,7 @@ std::string ConvertSystemPermission(
 
 void SystemPreferences::PostNotification(const std::string& name,
                                          base::DictionaryValue user_info,
-                                         gin_helper::Arguments* args) {
+                                         gin::Arguments* args) {
   bool immediate = false;
   args->GetNext(&immediate);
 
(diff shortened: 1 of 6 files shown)


shell/browser/api/electron_api_system_preferences_mac.mm
```diff
@@ -651,13 +654,13 @@ AVMediaType ParseMediaType(const std::string& media_type) {
...
 
-void SystemPreferences::SetAppLevelAppearance(gin_helper::Arguments* args) {
+void SystemPreferences::SetAppLevelAppearance(gin::Arguments* args) {
   if (@available(macOS 10.14, *)) {
     NSAppearance* appearance;
     if (args->GetNext(&appearance)) {
       [[NSApplication sharedApplication] setAppearance:appearance];
     } else {
-      args->ThrowError("Invalid app appearance provided as first argument");
+      args->ThrowError();
```
MarshallOfSound, line 663: We've lost the Better Error here?

nornagon: Yeah, kinda. It matches other gin errors, "Conversion error at index 0". I'm on the fence as to whether consistency or descriptiveness wins here, as the error message here isn't much different to the default one that gin provides.

PR #24708, opened 2020-07-23
fix: Resolve bookmarks without mounting network drives

CharlieHess:
#### Description of Change
Slack had several customers reporting hangs, reproducible on a version where we had started using security-scoped bookmarks on MAS builds. After collecting a sample, we found the culprit–network drives being mounted during `URLByResolvingBookmarkData`:

```
2623 LSSharedFileListItemCopyResolvedURL  (in SharedFileList)
2623 +[SFLList(LSSharedFileListSupport) resolveItem:resolutionFlags:error:]  (in SharedFileList)
2623 -[SFLBookmark resolveWithOptions:relativeToURL:error:]  (in SharedFileList)
2623 -[NSData(SFLBookmarkAdditions) __sfl_resolveBookmarkWithOptions:relativeToURL:bookmarkDataIsStale:error:]  (in SharedFileList)
2623 +[NSURL(NSURL) URLByResolvingBookmarkData:options:relativeToURL:bookmarkDataIsStale:error:]  (in Foundation)
2623 -[NSURL(NSURL) initByResolvingBookmarkData:options:relativeToURL:bookmarkDataIsStale:error:]  (in Foundation)
2623 _CFURLCreateByResolvingBookmarkData  (in CoreServicesInternal)
2623 createByResolvingBookmarkDataInternal(__CFAllocator const*, __CFData const*, unsigned long, __CFURL const*, __CFArray const*, unsigned char*, __CFError**, __CFURL const**)  (in CoreServicesInternal)
2623 CFURLCreateByResolvingDataInBookmark(__CFAllocator const*, BookmarkData&, unsigned int, unsigned long, __CFArray const*, unsigned char*, __CFError**, __CFURL const**)  (in CoreServicesInternal)
2623 NetFSMountURLSync  (in NetFS)
2623 NAConnectToServerSync  (in NetAuth)
2623 _NetAuthConnectToServerSync  (in NetAuth)
2623 NAAA_MountURL  (in NetAuth)
```

Some web searching turned up a few other folks encountering the same problem, specifically [here](https://github.com/quicksilver/Quicksilver/issues/1977) and [here](http://tangent405.com/dealing-with-slow-security-scoped-bookmarks). Both fixed it by appending the `NSURLBookmarkResolutionWithoutUI|NSURLBookmarkResolutionWithoutMounting` flags to the options here.

cc @nornagon 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Resolve app-scoped bookmarks without mounting network drives


shell/browser/api/electron_api_app_mas.mm
```diff
@@ -34,7 +34,7 @@ void OnStopAccessingSecurityScopedResource(NSURL* bookmarkUrl) {
   NSError* error = nil;
   NSURL* bookmarkUrl =
       [NSURL URLByResolvingBookmarkData:bookmarkData
-                                options:NSURLBookmarkResolutionWithSecurityScope
+                                options:NSURLBookmarkResolutionWithSecurityScope|NSURLBookmarkResolutionWithoutUI|NSURLBookmarkResolutionWithoutMounting
```
nornagon, line 37: `WithoutMounting` makes sense to me here to address the specific problem being encountered. I don't know enough about this system to know if `WithoutUI` is desirable here.

cc @acheronfail who wrote this and @alexeykuzmin who reviewed it

CharlieHess: >`NSURLBookmarkResolutionWithoutUI`
Specifies that no UI feedback should accompany resolution of the bookmark data.

Seems totes reasonable that we want this as well.

nornagon: According to https://stackoverflow.com/a/2235076, that controls showing the auth dialog if the user needs to reauth, which maybe they do?

semi-related, http://michaellynn.github.io/2015/10/24/apples-bookmarkdata-exposed/ is a fascinating dive through some internals of this stuff.

MarshallOfSound: Seems like these should just be exposed as options `{ tryMountIfRequired, showUIIfRequired }`

nornagon: @MarshallOfSound in the interest of this being a fix rather than a feat, should we perhaps avoid changing the api surface?

MarshallOfSound: Fair, then we're just discussing whether to include the UI flag or not ❓ If the goal is completely silent bookmark usage then the no-ui flag makes sense IMO

nornagon: As I understand it, the goal is to stop hanging indefinitely when there are stuck network drives, and it's not clear to me that the UI flag is causing that problem.

CharlieHess: yeah I'm cool with removing the UI flag, it just seemed reasonable that this rando method should not be popping dialogs on our users either (although I'm yet to see a customer report with one)

zcbenz:
Rebasing this branch on master should be able to fix the lint errors.

nornagon:
ugh sorry

nornagon:
Hi @electron/wg-releases, this fixes an issue for us where using a bookmark could result in the app hanging. It'd be great to get it backported. I'm not familiar enough with these APIs to know if this change could potentially break an app though.

Thoughts?

zcbenz:
It seems that with this change some symbolic links to network drives will not work under MAS, but I think it is safe to backport this as those accesses would be very easy to hang anyway.

Issue #24715, opened 2020-07-23
Using `nodeIntegrationInWorker: true` with a service worker crashes


lgrahl:
### Preflight Checklist

* [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [X] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 9.1.1
  * 10.0.0-beta.4
* **Operating System:**
  * Linux x64

### Expected Behavior

Don't crash when `nodeIntegrationInWorker` is `true` and a service worker is being registered. :slightly_smiling_face:

### Actual Behavior

A crash occurs once the service worker is attempted to be registered (blank page, DevTools disconnects).

### To Reproduce

Attempt to register a service worker with `nodeIntegrationInWorker` set to `true`.

```sh
$ git clone https://github.com/lgrahl/electron-node-integration-service-worker-crash
$ npm install
$ npm start
```

Commit https://github.com/lgrahl/electron-node-integration-service-worker-crash/commit/97b14f2c599bfd25ce416b8a7ee91d9122bbfe2c

codebytere:
I've located the cause of this crash but the fix is less clear - i'll hopefully have a PR soon.

PR #24736, opened 2020-07-27
fix: service worker crash with Node.js

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/24715.

This crash occurs because any calls into JS in `WebWorkerObserver::ContextCreated` trigger a CHECK [here](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/workers/worker_global_scope.cc;l=155?q=%22WorkerGlobalScope::url%22&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2F).

I initially tried:

- Preventing gin from emitting if the execution context is a service worker, but it turns out that in Node.js v12 `LoadEnvironment` also calls into JS so this solution wouldn't be tenable until we roll in the updates to the embedder API added in [this commit](https://github.com/nodejs/node/pull/30467/commits/7336b7fc1ba97d2b7d1028252631748cf12a2fed) in Node.js v14
- Setting the URL so that it's not null - `InitializeURL` is a [protected member](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/workers/worker_global_scope.h;l=237;bpv=1;bpt=1?originalUrl=https:%2F%2Fcs.chromium.org%2F) of `WorkerGlobalScope` though so this would also require a patch.

For now i think thus that the best path is just to disable the CHECK as there exists a path for its removal in Node.js v14 via something like:

```
blink::ExecutionContext* execution_context = blink::ToExecutionContext(worker_context);
bool should_emit = !execution_context->IsServiceWorkerGlobalScope();

// Load everything.
node_bindings_->LoadEnvironment(env, should_emit);
```

cc @nornagon @deepak1556 @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash in service workers when `nodeIntegrationInWorker`s was enabled.

### DIFF
--- a/patches/chromium/dcheck.patch
+++ b/patches/chromium/dcheck.patch
@@ -45,6 +45,22 @@ index 954d5401b71bb090c1c533783398298f28eeefbd..dd9051afe099cb1ef3aebb825cf05c34
    // The browser requested to clear the session history when it initiated the
    // navigation. Now we know that the renderer has updated its state accordingly
    // and it is safe to also clear the browser side history.
+diff --git a/third_party/blink/renderer/core/workers/worker_global_scope.cc b/third_party/blink/renderer/core/workers/worker_global_scope.cc
+index d56e4c2a3ad775144a1c913bd592eebfdfc1a0c7..07d3d47cfd2ca997a9e911b030778ba1a315d6cf 100644
+--- a/third_party/blink/renderer/core/workers/worker_global_scope.cc
++++ b/third_party/blink/renderer/core/workers/worker_global_scope.cc
+@@ -152,7 +152,10 @@ WorkerGlobalScope::~WorkerGlobalScope() {
+ }
+ 
+ NOINLINE const KURL& WorkerGlobalScope::Url() const {
+-  CHECK(url_.IsValid());
++  // Whenever JS is called into during context creation, this CHECK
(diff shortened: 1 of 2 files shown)


deepak1556:
Thanks for looking into it, the crash is not specific to service worker, it can happen for any worker with nodeIntegration enabled that relies on fetching its script off the main thread, this is an optimization that landed a while back.

https://github.com/electron/electron/pull/20625/files#diff-d3280e60e04d19f26a043d784d154227 has the details

> it turns out that in Node.js v12 LoadEnvironment also calls into JS so this solution wouldn't be tenable until we roll in the updates to the embedder API added in this commit in Node.js v14

If there is a way to delay access into global properties from Node.js then thats better.

For path forward, I am hesitant to disable the check, can we instead take the patch from https://github.com/electron/electron/pull/20625 and remove it once the node API is available ?

codebytere:
ah - is there any reason the other patch can't land? happy to close this in favor of that one!

deepak1556:
I didn't get time to finish it, main reason was to add some tests, as currently we don't cover worker scripts with nodeIntegration in the test suite.

codebytere:
Closing in favor of https://github.com/electron/electron/pull/20625.

PR #24771, opened 2020-07-29
chore: add V8 crash information to crashReporter

codebytere:
#### Description of Change

This PR restores our use of a custom V8 fatal error handler with the purpose of improving crash reports for crashes originating in V8. When a v8 crash occurs, the message and location will be added as extra parameters to the crash report for more effective debugging.

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added V8 crash message and location information to crashReport parameters.

### DIFF
--- a/shell/common/api/electron_api_v8_util.cc
+++ b/shell/common/api/electron_api_v8_util.cc
@@ -103,16 +103,6 @@ void RequestGarbageCollectionForTesting(v8::Isolate* isolate) {
       v8::Isolate::GarbageCollectionType::kFullGarbageCollection);
 }
 
-// This causes a fatal error by creating a circular extension dependency.
-void TriggerFatalErrorForTesting(v8::Isolate* isolate) {
-  static const char* aDeps[] = {"B"};
-  v8::RegisterExtension(std::make_unique<v8::Extension>("A", "", 1, aDeps));
-  static const char* bDeps[] = {"A"};
-  v8::RegisterExtension(std::make_unique<v8::Extension>("B", "", 1, bDeps));
-  v8::ExtensionConfiguration config(1, bDeps);
-  v8::Context::New(isolate, &config);
-}
-
 bool IsSameOrigin(const GURL& l, const GURL& r) {
   return url::Origin::Create(l).IsSameOriginWith(url::Origin::Create(r));
 }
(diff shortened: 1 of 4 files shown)


shell/common/api/electron_api_v8_util.cc
```diff
@@ -103,6 +103,16 @@ void RequestGarbageCollectionForTesting(v8::Isolate* isolate) {
       v8::Isolate::GarbageCollectionType::kFullGarbageCollection);
 }
 
+// This causes a fatal error by creating a circular extension dependency.
```
MarshallOfSound, line 106: Can we wrap this in a `DCHECK_ENABLED` check or something so it is literally only exposed in a testing build?

shell/common/node_bindings.cc
```diff
@@ -132,6 +137,15 @@ bool IsPackagedApp() {
 #endif
 }
 
+void FatalErrorCallback(const char* location, const char* message) {
+  LOG(ERROR) << "Fatal error in V8: " << location << " " << message;
+
+  electron::crash_keys::SetCrashKey("electron.v8-fatal.message", message);
```
MarshallOfSound, line 143: This needs to be wrapped in a `!defined(MAS_BUILD)` check as well

shell/common/node_bindings.cc
```diff
@@ -132,6 +137,15 @@ bool IsPackagedApp() {
...
+  electron::crash_keys::SetCrashKey("electron.v8-fatal.location", location);
+
+  electron::ElectronBindings::Crash();
```
MarshallOfSound, line 146: This is kinda a /nit and the compiler might do it for us anyway, but can we rename this method `V8FatalErrorCallback` and actually do the crash inline here?  That way the stacktrace when it is reported ends with `V8FatalErrorCallback` instead of `ElectronBindings::Crash` which isn't super helpful

shell/common/node_bindings.cc
```diff
@@ -132,6 +137,18 @@ bool IsPackagedApp() {
...
+void V8FatalErrorCallback(const char* location, const char* message) {
+  LOG(ERROR) << "Fatal error in V8: " << location << " " << message;
+
+#if !defined(MAS_BUILD)
+  electron::crash_keys::SetCrashKey("electron.v8-fatal.message", message);
+  electron::crash_keys::SetCrashKey("electron.v8-fatal.location", location);
+#endif
+
+  volatile int* zero = nullptr;
+  *zero = 0;
```
nornagon, line 149: Hm, why introduce new crash code rather than calling ElectronBindings::Crash()?

codebytere: @nornagon https://github.com/electron/electron/pull/24771#discussion_r462014987

PR #24782, opened 2020-07-29
fix: increase max crash key value length

nornagon:
#### Description of Change
This allows storing more complex data in the crash values, and will enable Sentry's reporter to switch away from their custom uploader.

This will waste memory for smaller crash keys, but most apps use only a small number of crash keys, so the excess won't be significant. In future, we can optimize this further by either separating crash keys into various classes by length, or expanding the API to allow the user to specify the maximum length of the key, or some other system.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Increased maximum length for crash keys from 127B to 20KB.

### DIFF
--- a/spec-main/api-crash-reporter-spec.ts
+++ b/spec-main/api-crash-reporter-spec.ts
@@ -196,6 +196,20 @@ ifdescribe(!isLinuxOnArm && !process.mas && !process.env.DISABLE_CRASH_REPORTER_
   });
 
   ifdescribe(!isLinuxOnArm)('extra parameter limits', () => {
+    function stitchLongCrashParam (crash: any, paramKey: string) {
+      if (crash[paramKey]) return crash[paramKey];
+      let chunk = 1;
+      if (crash[`${paramKey}__${chunk}`]) {
+        let stitched = '';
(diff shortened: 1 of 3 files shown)


spec-main/api-crash-reporter-spec.ts
```diff
@@ -196,19 +196,33 @@ ifdescribe(!isLinuxOnArm && !process.mas && !process.env.DISABLE_CRASH_REPORTER_
...
+      let chunk = 1;
+      if (crash[`${paramKey}__${chunk}`]) {
+        let stitched = '';
+        while (crash[`${paramKey}__${chunk}`]) {
+          stitched += crash[`${paramKey}__${chunk}`];
+          chunk++;
+        }
+        return stitched;
+      }
+      return null;
```
nornagon, line 210:
```suggestion
      let stitched = '';
      while (crash[`${paramKey}__${chunk}`]) {
        stitched += crash[`${paramKey}__${chunk}`];
        chunk++;
      }
      return stitched;
```

Doesn't seem like other code here uses the distinction between `''` and `null` so I dropped that.

PR #24785, opened 2020-07-29
chore: replace OS_MACOSX usage with OS_MAC

MarshallOfSound:
OS_MACOSX --> OS_MAC

Refs: https://bugs.chromium.org/p/chromium/issues/detail?id=1105907

Notes: no-notes


MarshallOfSound:
Looks like this needs a chromium roll first

PR #24798, opened 2020-07-30
fix(extensions): enable WebSQL in background pages

sentialx:
#### Description of Change
Ref #19447

Fixes a security error when calling `openDatabase` in an extension background page. An example major extension which uses it is LastPass.

Could this be backported to 9-x-y?

cc @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed not working `WebSQLDatabase` in extension background pages.


shell/browser/electron_browser_client.cc
```diff
@@ -769,6 +770,13 @@ void ElectronBrowserClient::AppendExtraCommandLineSwitches(
     content::WebContents* web_contents =
         GetWebContentsFromProcessID(process_id);
     if (web_contents) {
+      // Extension background pages don't have WebContentsPreferences, but they
+      // support WebSQL by default.
+#if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
+      if (extensions::GetViewType(web_contents) !=
+          extensions::VIEW_TYPE_INVALID)
+        command_line->AppendSwitch(switches::kEnableWebSQL);
```
nornagon, line 778: Perhaps
```suggestion
      if (extensions::ProcessMap::Get(browser_context)->Contains(process->GetID()))
        command_line->AppendSwitch(switches::kEnableWebSQL);
```

By analogy with //chrome/browser/extensions/chrome_content_browser_client_extensions_part.cc (ChromeContentBrowserClientExtensionsPart::AppendExtraRendererCommandLineSwitches)

I think it's logically more correct to associate this with the _process_ rather than the WebContents, though the effect will presently be the same, I think.

sentialx: So I assume it should be out of the `if (web_contents)` block?

nornagon: yep, also you can get the BrowserContext similarly to this: https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/chrome_content_browser_client.cc;l=2189-2193;drc=398ae53c39c1011dfc62bc95e59edbdf62f66374

spec-main/extensions-spec.ts
```diff
@@ -37,6 +37,18 @@ describe('chrome extensions', () => {
...
+    const w = new BrowserWindow({ show: false, webPreferences: { session: customSession, sandbox: true } });
+    w.loadURL('about:blank');
+
+    await emittedOnce(w.webContents, 'dom-ready');
+    await customSession.loadExtension(path.join(fixtures, 'extensions', 'persistent-background-page'));
+    const args: any = await emittedOnce(app, 'web-contents-created');
+    const wc: Electron.WebContents = args[1];
+    await expect(wc.executeJavaScript('(()=>{try{openDatabase("t", "1.0", "test", 2e5);return true;}catch(e){throw e}})()')).to.not.be.rejected();
+  });
+
```
zcbenz, line 51: The trailing space should be removed to fix the lint warning.

sentialx: @zcbenz Fixed!

codebytere:
@sentialx could you please backport this to 10 and 9?

sentialx:
@codebytere Sure

PR #24802, opened 2020-07-31
feat: add nativeImage.createThumbnailFromPath API

georgexu99:
#### Description of Change

This api returns a preview thumbnail of a file given its file path. The thumbnail is the same as a file's preview picture in explorer/finder.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added app.createThumbnailFromPath() API that returns a preview image of a file given its file path and a maximum thumbnail size.


shell/browser/browser_mac.mm
```diff
@@ -95,6 +98,26 @@
...
+gfx::Image Browser::CreateThumbnailFromPath(const base::FilePath& path,
+                                            int width,
+                                            int height) {
+  NSString* ns_path = base::mac::FilePathToNSString(path);
+  NSSize size = NSMakeSize(width, height);
+  CFURLRef cfurl = (__bridge CFURLRef)[NSURL fileURLWithPath:ns_path];
+
+  QLThumbnailRef ql_thumbnail = QLThumbnailCreate(
+      kCFAllocatorDefault, cfurl, CGSizeMake(size.width, size.height), NULL);
+  CGImageRef cg_thumbnail = QLThumbnailCopyImage(ql_thumbnail);
```
MarshallOfSound, line 110: We should be using `QLThumbnailDispatchAsync` first here to ensure that we don't block the main thread while we wait for quicklook to generate the thumbnail.  This API will need to return a promise.

Refs: https://developer.apple.com/documentation/quicklook/1402621-qlthumbnailcopyimage?language=objc

> If you call this function directly without using QLThumbnailDispatchAsync, this function blocks the calling thread until it finishes computing the thumbnail.

MarshallOfSound:
I think it fits with `nativeImage` nicely (create from x).  I think it was originally put in `app` because of `app.getFileIcon` which is kinda weird to be in `app` as well 🤷

docs/api/native-image.md
```diff
@@ -119,6 +119,13 @@ Returns `NativeImage`
 
 Creates an empty `NativeImage` instance.
 
+### `nativeImage.createThumbnailFromPath(path, maxSize)` _macOS_ _Windows_
+
+* `path` String - path to a file that we intend to construct a thumbnail out of.
+* `maxSize` [Size](structures/size.md) - the maximum width and height the thumbnail returned can be. The Windows implementation will ignore this parameter and scale the height according to `width`.
```
ckerr, line 125: nits:

* Should document that `width` and `height` must be positive

* Guessing without peeking: this sounds like we're saying that `maxSize.width` *will* be honored? If so, we shouldn't say Windows will ignore `maxSize` and instead say it will ignore `maxSize.height`

* Similar to previous, "...according to `width`" would be marginally clearer as "...according to `maxSize.width`"

shell/common/api/electron_api_native_image_mac.mm
```diff
@@ -34,6 +39,64 @@
...
+  if (size.height() <= 0) {
+    promise.RejectWithErrorMessage(
+        "invalid height, please enter a positive number");
+    return handle;
+  }
+
+  if (!path.IsAbsolute()) {
+    promise.RejectWithErrorMessage("path must be absolute");
+    return handle;
+  }
```
ckerr: We don't need 2-3 copies of the input sanitizing code. Better to have one function in electron-api-native-image.cc that does these checks, then delegates the work out to per-platform functions that take presanitized input.

shell/common/api/electron_api_native_image_mac.mm
```diff
@@ -34,6 +39,49 @@
...
+
+  if (size.IsEmpty()) {
+    promise.RejectWithErrorMessage("size must not be empty");
+    return handle;
+  }
+
+  CGSize cg_size = size.ToCGSize();
+  base::ScopedCFTypeRef<CFURLRef> cfurl = base::mac::FilePathToCFURL(path);
+  base::ScopedCFTypeRef<QLThumbnailRef> ql_thumbnail(
+      QLThumbnailCreate(kCFAllocatorDefault, cfurl, cg_size, NULL));
```
ckerr, line 58: This function is deprecated. Is it the best option?

spec/api-native-image-spec.js
```diff
@@ -515,6 +515,41 @@ describe('nativeImage module', () => {
...
+    });
+
+    ifit(process.platform === 'win32')('throws when a bad path is passed (Windows)', async () => {
+      const badPath = '\\hey\\hi\\hello';
+      const goodSize = { width: 100, height: 100 };
+
+      await expect(
+        nativeImage.createThumbnailFromPath(badPath, goodSize)
+      ).to.eventually.be.rejected();
+    });
```
ckerr, line 543: These are the same test -- they could be merged into one test with something like

```js
const badPath = process.platform === 'win32' ? '\\hey\\hi\\hello' : '/hey/hi/hello';
```

nornagon:
NB. release notes don't match impl

shell/common/api/electron_api_native_image_mac.mm
georgexu99: This is removed in the latest commit, it was my mistake where I was unaware that gfx::Size does not allow negative inputs

shell/common/api/electron_api_native_image_mac.mm
```diff
@@ -34,6 +39,49 @@
...
+        } else {
+          dispatch_async(dispatch_get_main_queue(), ^{
+            p.RejectWithErrorMessage("unable to retrieve thumbnail preview "
+                                     "image for the given path");
+          });
+        }
+      });
+  return handle;
+}
+
```
ckerr, line 84: Do we need per-platform implementations of this?

For example, what if we did something like this, in a non-blocking thread:

```
gfx::ImageSkia original;
electron::util::PopulateImageSkiaRepsFromPath(&original, filename);
auto scaled = gfx::ImageSkiaOperations::CreateResizedImage(original, skia::ImageOperations::RESIZE_GOOD);
gfx::Image image(scaled);
return Create(isolate, image);
```

To be clear, I haven't tested this pseudocode and don't know what the quality of the output is. The reason I think this would be worth trying is:

* Better to maintain one codebase instead of two
* If it works, it would eliminate the "width but not height supported on Windows" wart
* Picks up Linux for free

shell/common/api/electron_api_native_image_mac.mm
georgexu99: I believe the other option is QLTHumbnailImageCreate (https://developer.apple.com/documentation/quicklook/1402623-qlthumbnailimagecreate?language=objc) 

but @MarshallOfSound mentioned that we don't want to use that as it is a synchronous function

spec/api-native-image-spec.js
georgexu99: ur totally right thanks!

shell/common/api/electron_api_native_image_win.cc
```diff
@@ -0,0 +1,106 @@
...
+// static
+v8::Local<v8::Promise> NativeImage::CreateThumbnailFromPath(
+    v8::Isolate* isolate,
+    const base::FilePath& path,
+    const gfx::Size& size) {
+  gin_helper::Promise<gfx::Image> promise(isolate);
+  v8::Local<v8::Promise> handle = promise.GetHandle();
+  HRESULT hr;
+
+  if (size.IsEmpty()) {
```
ckerr, line 33: IsEmpty() is terser than the previous "width <= 0 ... height <= 0" checks but this revision lets negative sizes get by, e.g. `maxSize { width: -1, height: -1 }`. Changing that to `isEmpty()` isn't necessarily a blocker if you *intend* to let those values through, but since nothing's mentioned in the new docs, I thought I'd ask if you meant to let those through

shell/common/api/electron_api_native_image_mac.mm
georgexu99: What this API does is create a *new* image from a path rather than resizing it rather than making it a smaller version of a original image

we have different implementations because of the different APIs needed to extract a thumbnail from a path

shell/common/api/electron_api_native_image_win.cc
```diff
@@ -0,0 +1,106 @@
...
+
+#include "shell/common/api/electron_api_native_image.h"
+
+#include <windows.h>  // NOLINT(build/include_order)
+
+#include <thumbcache.h>  // NOLINT(build/include_order)
+#include <wrl/client.h>  // NOLINT(build/include_order)
+
+#include <string>  // NOLINT(build/include_order)
+#include <vector>  // NOLINT(build/include_order)
```
zcbenz, line 13: I think the `NOLINT`s should no longer be needed here.

zcbenz (approved):

shell/common/api/electron_api_native_image_mac.mm
MarshallOfSound: To be clear this constructs a thumbnail / preview / image _representation_ of a file at a given path.  i.e. we aren't resizing an image.  This means you call this API with a path to a word document and the OS will send you back a thumbnail image for that word document.

lib/browser/init.ts
```diff
@@ -214,3 +215,7 @@ if (packagePath) {
   console.error('Failed to locate a valid package to load (app, app.asar or default_app.asar)');
   console.error('This normally means you\'ve damaged the Electron package somehow');
 }
+
+ipcMainInternal.handle('ELECTRON_NATIVE_IMAGE_CREATE_THUMBNAIL_FROM_PATH', async (_, path: string, size: Electron.Size) => {
```
MarshallOfSound, line 219: Can you move this to `rpc-server.ts`

jkleinsc:
The @electron/wg-api approved this via Slack on Aug 20th, 2020

gaodeng:
I tried to get app icon and got an error 
```
Error: unable to retrieve thumbnail preview image for the given path
```
Does this API only get thumbnails of images?
In that case, there doesn't seem to be much difference between this api and nativeImage.createFromPath(path) + resize

MentalGear:
@gaodeng I'm also trying to get app icons in a cross-platform, high-res way. Could you share whether nativeImage.createThumbnailFromPath worked out for you, or what your successful approach was?

fanchenio:
@MentalGear Facts have proved that high-definition icons can only be obtained through native mode, and electron does not support it.

PR #24821, opened 2020-08-03
feat: add mas_build crash key

nornagon:
#### Description of Change
This adds a new crash key, `mas_build`, that will be present in mac crashes. Its value is `1` if the crash is from a MAS build, or `0` otherwise.

It's useful to be able to distinguish between MAS and non-MAS builds when debugging crashes, both because the symbolication is different (since the addresses of functions are different) and because the runtime environment is different.

I considered changing the value of the `platform` key from `darwin` to `darwin-mas` on MAS builds, but decided against it because (1) the `platform` key is redundant with a key that `//components/crash` adds, and might eventually be removed, (2) having both `platform` and `plat` exist but take different values is surprising, and (3) this approach is less likely to break anyone depending on particular crash keys existing and having certain expected values.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added a new default crash key on macOS builds, `mas_build`, for distinguishing between crashes from MAS vs. non-MAS builds.

### DIFF
--- a/shell/common/crash_keys.cc
+++ b/shell/common/crash_keys.cc
@@ -144,17 +144,16 @@ void SetPlatformCrashKey() {
   platform_key.Set("win32");
 #elif defined(OS_MACOSX)
   platform_key.Set("darwin");
-#elif defined(OS_LINUX)
-  platform_key.Set("linux");
-#else
-#error Unknown platform.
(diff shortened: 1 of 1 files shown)


nornagon:
doh

PR #24836, opened 2020-08-04
fix: patch io_surface.cc for macOS ARM64 builds

erickzhao:
#### Description of Change
As per the patch description:

This patch makes io_surface.cc build for macOS ARM64 and Big Sur. An alternative patch from [Chromium](https://chromium-review.googlesource.com/c/chromium/src/+/2315078) is not ideal because it references a private symbol.

cc @MarshallOfSound @CharlieHess 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -117,3 +117,4 @@ backport_1074340.patch
 backport_1016278.patch
 backport_1042986.patch
 a11y_axplatformnodebase_getindexinparent_returns_base_optional.patch
+build_io_surface_macos_arm64.patch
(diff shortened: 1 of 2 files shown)


MarshallOfSound:
@erickzhao You need to remove the chromium patch that references the private symbol.

```
#if defined(__MAC_11_0) && __MAC_OS_X_VERSION_MAX_ALLOWED >= __MAC_10_16
// https://crbug.com/1108561: This is supposed to be deprecated (but available)
// through 10.15.4, but the macOS 11.0 20A5323l SDK from Xcode 12b3 12A8169g
// doesn’t declare it at all when the SDK version is 10.16 or later, which of
// course it always is. It’s a bug in the SDK. Work around it with the
// declaration that’s hidden.
//
// TODO(https://crbug.com/1108627): remove this workaround if and when the SDK
// bug is fixed.
CG_EXTERN const CFStringRef kCGColorSpaceITUR_2020_PQ_EOTF
CG_AVAILABLE_BUT_DEPRECATED(10.14.6, 10.15.4, 12.6, 13.4);
#endif
```

PR #24838, opened 2020-08-04
fix: provide AXTextChangeValueStartMarker for macOS a11y value change notifications

MarshallOfSound:
Backport of https://github.com/electron/electron/pull/24801

See that PR for details.

This PR includes the final upstreamed CL so it looks slightly different.

Notes: Fixed issued where voice over would not read words as you typed them


codebytere:
@MarshallOfSound this needs a rebase and small fixup i think

Issue #24647, opened 2020-07-20
All child windows have active title bar


zarubond:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  6,7,8,9
* **Operating System:**
  Window 10
* **Last Known Working Electron version:**
5.0.13

### Expected Behavior
Only one window should have active title bar at one time.

### Actual Behavior
All electron windows have active titlebar.

### To Reproduce
**index.html**
```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8">
  </head>
  <body>
    <script>
      // You can also require other files to run in this process
      function openWin() {
        let modal = window.open(\'\',\'modal\');
        modal.document.write('<h1>Hello</h1>');
      }
    </script>
    <button onclick="openWin()">Open window</button>
  </body>
</html>
```
**main.js**
```
// Modules to control application life and create native browser window
const {app, BrowserWindow} = require('electron')

function createWindow () {
  // Create the browser window.
  const mainWindow = new BrowserWindow({
    width: 800,
    height: 600,
    webPreferences: {
      nativeWindowOpen: true
    }
  })

  mainWindow.webContents.on('new-window', (event, url, frameName, disposition, options, additionalFeatures) => {
  if (frameName === 'modal') {
    // open window as modal
    event.preventDefault()
    Object.assign(options, {
      width: 300,
      height: 300
    })
    event.newGuest = new BrowserWindow(options)
  }
})

  // and load the index.html of the app.
  mainWindow.loadFile('index.html')

  // Open the DevTools.
  // mainWindow.webContents.openDevTools()
}

// This method will be called when Electron has finished
// initialization and is ready to create browser windows.
// Some APIs can only be used after this event occurs.
app.on('ready', createWindow)

// Quit when all windows are closed.
app.on('window-all-closed', function () {
  // On OS X it is common for applications and their menu bar
  // to stay active until the user quits explicitly with Cmd + Q
  if (process.platform !== 'darwin') {
    app.quit()
  }
})

app.on('activate', function () {
  // On OS X it's common to re-create a window in the app when the
[...]

PR #24847, opened 2020-08-05
fix: do not render inactive titlebar as active on Windows

zcbenz:
#### Description of Change

Chromium renders titlebar as active as long as there is any views window active, even when current window is not active, which is different from most other win32 apps. We prevent this behavior by overriding the `WM_NCACTIVATE` message to default behavior.

Close https://github.com/electron/electron/issues/24647.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix inactive windows having active titlebar on Windows

### DIFF
--- a/shell/browser/ui/win/electron_desktop_window_tree_host_win.cc
+++ b/shell/browser/ui/win/electron_desktop_window_tree_host_win.cc
@@ -24,6 +24,14 @@ bool ElectronDesktopWindowTreeHostWin::PreHandleMSG(UINT message,
   return native_window_view_->PreHandleMSG(message, w_param, l_param, result);
 }
 
+bool ElectronDesktopWindowTreeHostWin::ShouldPaintAsActive() const {
+  // Tell Chromium to use system default behavior when rendering inactive
+  // titlebar, otherwise it would render inactive titlebar as active under
+  // some cases.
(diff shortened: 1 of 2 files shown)


shell/browser/native_window_views_win.cc
```diff
@@ -301,6 +301,15 @@ bool NativeWindowViews::PreHandleMSG(UINT message,
       }
       return false;
     }
+    case WM_NCACTIVATE: {
+      // Chromium renders titlebar as active as long as there is a views window
+      // active, which is different from most other win32 apps. We prevent this
+      // behavior by overriding the WM_NCACTIVATE message to default behavior.
+      // See also https://github.com/electron/electron/issues/24647.
+      *result =
```
deepak1556, line 309: Can't we instead override [HWNDMessageHandlerDelegate::ShouldPaintAsActive()](https://source.chromium.org/chromium/chromium/src/+/master:ui/views/win/hwnd_message_handler.cc;l=2093) to `false` which would still resort to the default behavior because of `paint_as_active || active`?

zcbenz: This is a good idea 👍 , I have update the PR.

codebytere:
Failures are previously known and unrelated - merging.

Issue #23757, opened 2020-05-26
9.0.0 does not display local images


whyboris:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v9.0.0
* **Operating System:**
  * Windows 10
* **Last Known Working Electron version:**
  * v8.3.0

### Expected Behavior
My app shows a gallery of images from local disk. Electron 8.2.0 would show images. v9.0.0 shows blanks with error:
```
GET file:///C:/VHA/vha-videos/filmstrips/6f3acc0aeee45a7af5404e3849d8719f.jpg 
net::ERR_UNKNOWN_URL_SCHEME
```

### Actual Behavior
App does not show local images

### To Reproduce
Clone https://github.com/whyboris/Video-Hub-App
Run the app, create a "hub" (give it a folder with videos) and it will create a gallery of images.
Exit the app, update to Electron 9 (comment out 2 lines of `shell.openItem`) and run app again

### Screenshots

![image](https://user-images.githubusercontent.com/17264277/82849048-31880d80-9ec4-11ea-95fc-6639cf36e858.png)

codebytere:
@whyboris could you please provide a more minimal sample? That app is rather large which makes it harder for us to isolate the Electron-mediated issue separate from the rest of the app's workings.

whyboris:
Thank you @codebytere -- I will see if I can create a minimal sample 👍

astoilkov:
I am experiencing the same problem. I think it is related to the fact that I am running `webpack-dev-server` on `localhost` but requesting images from the filesystem via `file://` links. The same approach worked in version 8.

lyswhut:
same problem.
In v9.0.0, when browserWindow uses http protocol to load the page, the audio tag cannot load the local audio file,
When the file protocol is used to load the page, the local audio file can be loaded normally.

The last working version: v8.3.0

codebytere:
@lyswhut @astoilkov as i mentioned above - a repro will get this moving a lot faster!

lyswhut:
[...]

PR #24849, opened 2020-08-05
feat: add session.setCorsOriginAccessList API

lishid:
#### Description of Change

Summary: A new "local" scheme privilege is added, which grants access to file:// resources without needing to disable webSecurity.

Rationale:

- It is common for electron apps to define a custom protocol to serve app resources [(as seen from first example here)](https://www.electronjs.org/docs/api/protocol).
- Apps that registers their custom protocol as `secure` cannot access files from the `file:///` protocol.
- This rejection comes from deep within chromium and no Electron API seem to be able to override that, except `webSecurity: false`. Things I've tried:
  - `registerFileProtocol('file'...)`
  - `interceptFileProtocol('file'...)`
  - `session.defaultSession.webRequest.onBeforeRequest`
- `webSecurity: false` is not a recommended practice for security purposes by https://www.electronjs.org/docs/tutorial/security

This PR adds a `local` privilege to scheme registration, which adds the scheme to `blink::SchemeRegistry::RegisterURLSchemeAsLocal`. 
Doing so marks the protocol as local, thus allowing it to access local resources via the pathway:
[`SecurityOrigin::CanDisplay()`](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/platform/weborigin/security_origin.cc;l=417;drc=8d3351f2b93dc4fef60b794c47ec22ebf54f3e8b;bpv=1;bpt=1)
[`SecurityOrigin::can_load_local_resources_`](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/platform/weborigin/security_origin.cc;l=144;drc=8d3351f2b93dc4fef60b794c47ec22ebf54f3e8b)
[`SecurityOrigin::IsLocal()`](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/platform/weborigin/security_origin.cc;l=471;drc=8d3351f2b93dc4fef60b794c47ec22ebf54f3e8b)

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added "local" scheme privilege, which grants access to file:// without the need to disable webSecurity.

### DIFF
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -662,6 +662,29 @@ event. The [DownloadItem](download-item.md) will not have any `WebContents` asso
 the initial state will be `interrupted`. The download will start only when the
 `resume` API is called on the [DownloadItem](download-item.md).
 
+#### `ses.setCorsOriginAccessList(origin, allowPatterns, blockPatterns)`
+
+* `origin` String - The Origin URL for which to set the CORS policy.
+* `allowPatterns` String[] - An array of URL patterns to allow.
+* `blockPatterns` String[] - An array of URL patterns to block.
+
+Returns `Promise<void>` - Resolves when the operation is complete.
+
+Adds additional URL patterns to the allow and block list of the CORS policy.
+This is often used to securely permit pages from your custom protocol to access resources that would otherwise be blocked by CORS, such as `file:///` URLs, without having to use `webSecurity: false`.
+This can also be used to block access to certain resources.
+
+```javascript
+const { app, session } = require('electron')
+
+app.whenReady().then(async () => {
+  await session.defaultSession.setCorsOriginAccessList(
+    'custom://abc/hello.html',
+    ['https://*.github.com/*', '*://electron.github.io', 'file'], /* allow list */
+    ['http://*/*'] /* block list */)
+})
+```
+
 #### `ses.clearAuthCache()`
 
 Returns `Promise<void>` - resolves when the session’s HTTP authentication cache has been cleared.
--- a/docs/tutorial/security.md
+++ b/docs/tutorial/security.md
@@ -292,7 +292,9 @@ Do not disable `webSecurity` in production applications.
 
 Disabling `webSecurity` will disable the same-origin policy and set
 `allowRunningInsecureContent` property to `true`. In other words, it allows
-the execution of insecure code from different domains.
+the execution of insecure code from different domains. It also allows any
+website from any URL to access files on the local computer using
+the `file:///` protocol.
 
 ### How?
 
@@ -316,6 +318,24 @@ const mainWindow = new BrowserWindow()
 
 <!-- Good -->
 <webview src="page.html"></webview>
+```
+
+### Allow custom protocol to access `file:///` resources
+
+Instead of disabling `webSecurity`, consider using the
+[`session.setCorsOriginAccessList`][set-cors-origin-access-list] API
+to allow cross-origin resource access for your custom protocol.
+
+```javascript
+const { app, session } = require('electron')
+
+app.whenReady().then(async () => {
+  await session.defaultSession.setCorsOriginAccessList(
+    'my-custom-protocol://abc/index.html',
+    ['file'], /* allow list */
+    [] /* block list */
+  )
+})
 ```
 
 ## 6) Define a Content Security Policy
@@ -694,3 +714,4 @@ which potential security issues are not as widely known.
 [open-external]: ../api/shell.md#shellopenexternalurl-options
 [sandbox]: ../api/sandbox-option.md
 [responsible-disclosure]: https://en.wikipedia.org/wiki/Responsible_disclosure
+[set-cors-origin-access-list]: ../api/session.md#sessetcorsoriginaccesslistorigin-allowpatterns-blockpatterns
--- a/shell/browser/api/electron_api_session.cc
+++ b/shell/browser/api/electron_api_session.cc
@@ -318,6 +318,38 @@ class DictionaryObserver final : public SpellcheckCustomDictionary::Observer {
 };
 #endif  // BUILDFLAG(ENABLE_BUILTIN_SPELLCHECKER)
 
+uint16_t GetEffectivePort(const std::string& port_string) {
+  int port_int = 0;
+  bool success = base::StringToInt(port_string, &port_int);
+  // The URLPattern should verify that |port| is a number or "*", so conversion
+  // should never fail.
+  DCHECK(success) << port_string;
+  return port_int;
+}
+
+void AddPatternSetToAccessList(
+    const std::set<URLPattern>& pattern_set,
+    std::vector<network::mojom::CorsOriginPatternPtr>* list,
+    network::mojom::CorsOriginAccessMatchPriority priority) {
+  for (const URLPattern& pattern : pattern_set) {
+    network::mojom::CorsDomainMatchMode domain_match_mode =
+        pattern.match_subdomains()
+            ? network::mojom::CorsDomainMatchMode::kAllowSubdomains
+            : network::mojom::CorsDomainMatchMode::kDisallowSubdomains;
+    network::mojom::CorsPortMatchMode port_match_mode =
+        (pattern.port() == "*")
+            ? network::mojom::CorsPortMatchMode::kAllowAnyPort
+            : network::mojom::CorsPortMatchMode::kAllowOnlySpecifiedPort;
+    uint16_t port = (port_match_mode ==
+                     network::mojom::CorsPortMatchMode::kAllowOnlySpecifiedPort)
+                        ? GetEffectivePort(pattern.port())
+                        : 0u;
+    list->push_back(network::mojom::CorsOriginPattern::New(
+        pattern.scheme(), pattern.host(), port, domain_match_mode,
+        port_match_mode, priority));
+  }
+}
+
 struct UserDataLink : base::SupportsUserData::Data {
   explicit UserDataLink(Session* ses) : session(ses) {}
 
@@ -749,6 +781,32 @@ void Session::DownloadURL(const GURL& url) {
   download_manager->DownloadUrl(std::move(download_params));
 }
 
+v8::Local<v8::Promise> Session::SetCorsOriginAccessList(
+    const GURL& url,
+    const std::set<URLPattern>& allow_pattern_set,
+    const std::set<URLPattern>& block_pattern_set) {
+  auto* isolate = JavascriptEnvironment::GetIsolate();
+  gin_helper::Promise<void> promise(isolate);
+  v8::Local<v8::Promise> handle = promise.GetHandle();
+
+  std::vector<network::mojom::CorsOriginPatternPtr> allow_list;
+  std::vector<network::mojom::CorsOriginPatternPtr> block_list;
+  AddPatternSetToAccessList(
+      allow_pattern_set, &allow_list,
+      network::mojom::CorsOriginAccessMatchPriority::kDefaultPriority);
+  AddPatternSetToAccessList(
+      block_pattern_set, &block_list,
+      network::mojom::CorsOriginAccessMatchPriority::kDefaultPriority);
+
+  browser_context_->SetCorsOriginAccessListForOrigin(
+      content::BrowserContext::TargetBrowserContexts::kSingleContext,
+      url::Origin::Create(url), std::move(allow_list), std::move(block_list),
+      base::BindOnce(gin_helper::Promise<void>::ResolvePromise,
+                     std::move(promise)));
+
+  return handle;
+}
+
 void Session::CreateInterruptedDownload(const gin_helper::Dictionary& options) {
   int64_t offset = 0, length = 0;
   double start_time = base::Time::Now().ToDoubleT();
@@ -1174,6 +1232,7 @@ gin::ObjectTemplateBuilder Session::GetObjectTemplateBuilder(
       .SetMethod("setSSLConfig", &Session::SetSSLConfig)
       .SetMethod("getBlobData", &Session::GetBlobData)
       .SetMethod("downloadURL", &Session::DownloadURL)
+      .SetMethod("setCorsOriginAccessList", &Session::SetCorsOriginAccessList)
       .SetMethod("createInterruptedDownload",
                  &Session::CreateInterruptedDownload)
       .SetMethod("setPreloads", &Session::SetPreloads)
--- a/shell/browser/api/electron_api_session.h
+++ b/shell/browser/api/electron_api_session.h
@@ -5,6 +5,7 @@
 #ifndef SHELL_BROWSER_API_ELECTRON_API_SESSION_H_
 #define SHELL_BROWSER_API_ELECTRON_API_SESSION_H_
 
+#include <set>
 #include <string>
 #include <vector>
 
@@ -124,6 +125,10 @@ class Session : public gin::Wrappable<Session>,
   v8::Local<v8::Value> NetLog(v8::Isolate* isolate);
   void Preconnect(const gin_helper::Dictionary& options, gin::Arguments* args);
   v8::Local<v8::Promise> CloseAllConnections();
+  v8::Local<v8::Promise> SetCorsOriginAccessList(
+      const GURL& url,
+      const std::set<URLPattern>& allow_pattern_set,
+      const std::set<URLPattern>& block_pattern_set);
 #if BUILDFLAG(ENABLE_BUILTIN_SPELLCHECKER)
   base::Value GetSpellCheckerLanguages();
   void SetSpellCheckerLanguages(gin_helper::ErrorThrower thrower,
--- a/shell/browser/api/electron_api_web_request.cc
+++ b/shell/browser/api/electron_api_web_request.cc
@@ -30,19 +30,6 @@
 namespace gin {
 
 template <>
-struct Converter<URLPattern> {
-  static bool FromV8(v8::Isolate* isolate,
-                     v8::Local<v8::Value> val,
-                     URLPattern* out) {
-    std::string pattern;
-    if (!ConvertFromV8(isolate, val, &pattern))
-      return false;
-    *out = URLPattern(URLPattern::SCHEME_ALL);
-    return out->Parse(pattern) == URLPattern::ParseResult::kSuccess;
-  }
-};
-
-template <>
 struct Converter<extensions::WebRequestResourceType> {
   static v8::Local<v8::Value> ToV8(v8::Isolate* isolate,
                                    extensions::WebRequestResourceType type) {
--- a/shell/browser/net/network_context_service.cc
+++ b/shell/browser/net/network_context_service.cc
@@ -87,6 +87,11 @@ void NetworkContextService::ConfigureNetworkContextParams(
   network_context_params->enable_ftp_url_support = true;
 #endif  // !BUILDFLAG(DISABLE_FTP_SUPPORT)
 
+  network_context_params->cors_origin_access_list =
+      browser_context_->GetSharedCorsOriginAccessList()
+          ->GetOriginAccessList()
+          .CreateCorsOriginAccessPatternsList();
+
   proxy_config_monitor_.AddToNetworkContextParams(network_context_params);
 
   BrowserProcessImpl::ApplyProxyModeFromCommandLine(
--- a/shell/common/gin_converters/net_converter.cc
+++ b/shell/common/gin_converters/net_converter.cc
@@ -12,6 +12,7 @@
 #include "base/strings/string_number_conversions.h"
 #include "base/strings/string_util.h"
 #include "base/values.h"
+#include "extensions/common/url_pattern.h"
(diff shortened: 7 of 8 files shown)


lishid:
The main issue we're trying to solve here is that `webSecurity: false` allows external embedded <iframe>s (think `https://`) to load `file:///` URLs, which is a security issue.
Allowing local access for your custom `app://` URL means the page on `app://index.html` can access `file:///` resources, but an embedded <iframe> from `https://` cannot access `file:///`.

deepak1556:
Any reason you can't register another custom protocol that handles file paths ? what is the reason behind mixing file uri and custom protocol in the app ?

lishid:
@deepak1556 
> Any reason you can't register another custom protocol that handles file paths ?

That's what I'm currently using, but it has some limitations and complications when we render end-user generated contents. It would be extremely convenient if `file:///` paths just worked, rather than being blocked by `Not allowed to load local resource`.

For context, we render markdown content created by end users. We'd like to support 
- `file:///` image embeds
- Copying the rendered image to yield the `file:///` link in the clipboard as HTML (rather than `custom://`)
- <iframe> a local `.html` file

I imagine it's possible to work around these issues with an HTML post processor that replaces `file:///` links with `custom://`, then hook the `copy`, `cut`, `dragstart` events to replace the links back to `file:///`, but that's quite a poor solution with many easy-to-miss edge cases, for example, remembering to translate ALL src-like properties from `<video>`, `<audio>`, `<source>`, etc.

There's currently another bug which I have not yet filed, where `registerFileProtocol('custom')` does not work with the built-in PDF plugin using `custom://path/to/file.pdf`.

> what is the reason behind mixing file uri and custom protocol in the app?

If I understand correctly, custom protocol provides many advantages over file:/// for serving app resources:
- It can be registered as a standard scheme to allow absolute paths to resolve properly
- It can be registered as a secure protocol to access various 'secure' gated features, such as SubtleCrypto
- It sets a Origin properly when connecting with remote resources behind `https://`, which can be used to manage CORS securely.

lishid:
I'd like to add that while workarounds do exist, such as another custom protocol for files, developers using Electron are incentivized to take the easy route and just set `webSecurity: false`. (Related: https://github.com/electron/electron/issues/23757)

Most people don't really think about the implications of turning off webSecurity, other than that it enables the desired functionality. [By principle](https://www.electronjs.org/docs/tutorial/security) we should provide a safer way to access/embed local resources.

deepak1556:
Thanks for the added context, really helps!

* When you register a scheme with this new local privilege, it will get the same security policies as a `file` protocol, that means you can no longer access resources on this custom protocol from other privileged schemes. Would this custom scheme still be useful for your app ?

* Have a scheme with `secure, standard and local` privileges is not a valid path in chromium, so there are more undefined behaviours if users arbitrarily configure these options.

I think it should be hard to configure bad custom schemes in electron, if the api were to make `local` as exclusive from other privileges (i-e) throw proper error messages, it should be fine.

With the above conditions, an example `app` scheme configured with `local` privilege will behave exactly like another `file` scheme. Would this still serve the purpose for your app ?

lishid:
> * When you register a scheme with this new local privilege, it will get the same security policies as a `file` protocol, that means you can no longer access resources on this custom protocol from other privileged schemes. Would this custom scheme still be useful for your app ?

I think that should be fine for most use cases. I imagine in the most common case, it would disallow `https://` pages access to include things from `my-app://`, which may even be considered a desirable behavior. It should be blocked by CORS anyway assuming the developer hasn't disabled CORS.

> * Have a scheme with `secure, standard and local` privileges is not a valid path in chromium, so there are more undefined behaviours if users arbitrarily configure these options.

That's interesting, definitely something I wasn't aware of. I must have missed some of the checks when browsing the code (please excuse my first time working with chromium source; would also appreciate if you can give some pointers to how chromium checks for such an invalid scheme).
I don't mind trying another approach, since I did notice there might be side effects from marking the custom scheme as local.

> I think it should be hard to configure bad custom schemes in electron, if the api were to make `local` as exclusive from other privileges (i-e) throw proper error messages, it should be fine.
> 
> With the above conditions, an example `app` scheme configured with `local` privilege will behave exactly like another `file` scheme. Would this still serve the purpose for your app ?

Given your insight that `local` must be exclusive from `standard` and `secure`, I would say that indeed makes this solution rather useless.

While digging, another possible solution I think could work is calling [`WebSecurityPolicy::AddOriginAccessAllowListEntry()`](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/exported/web_security_policy.cc;l=76;drc=8d3351f2b93dc4fef60b794c47ec22ebf54f3e8b;bpv=1;bpt=1), which would satisfy [this part](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/platform/weborigin/security_origin.cc;l=418;drc=8d3351f2b93dc4fef60b794c47ec22ebf54f3e8b) of the `CanDisplay()` check. The only problem is `AddOriginAccessAllowListEntry` requires an Origin (scheme + host) rather than just a scheme.

deepak1556:
> While digging, another possible solution I think could work is calling WebSecurityPolicy::AddOriginAccessAllowListEntry(), which would satisfy this part of the CanDisplay() check

that sounds better providing more finer control, it can be a new api on session module. I think the api works well with empty hosts and `file://` origin, can you test if this works for your test case ?

lishid:
Sure, I'll give it a try. Thanks for the tip!

lishid:
Sorry for the delay - I haven't been able to get Electron to build. After several attempts it seems to constantly get stuck on
`[14424/14849] STAMP obj/tools/v8_context_snapshot/v8_context_snapshot.stamp`

Is there any resources I can look for to debug this issue?

deepak1556:
you can run the ninja command with `-v` flag to get more verbose output. Also does your system have a minimum of 8GB RAM, it is bare minimum required for linking step.

lishid:
My VM has 16GB RAM, not sure if that's enough, I'll reboot it with 48GB after to see.

In my search, I found this thread https://groups.google.com/a/chromium.org/d/msg/chromium-dev/0g7D-ubb87c/dbpGxBSoBQAJ where this user ran into something similar.
According to his log, he's stuck on the LINK step, but I think my setup got stuck before that.
His log from the google groups link:
```
[4/5] STAMP obj/tools/v8_context_snapshot/v8_context_snapshot.stamp
[5/5] LINK ./chrome
```
---
I've restarted with `e build -v`, somehow it seems to get stuck on a `touch` command (makes sense that's what STAMP does) but I can run it just fine manually
image

deepak1556:
Is it still stuck at that step after you ran it manually ? Can you try removing that file and re-try again.

lishid:
That's right, I've tried a couple of times removing that file and resuming build - suspecting it may be a hardware issue, I'm formatting a new SSD to install a fresh OS and setup from scratch. Will report back once that's done.

lishid:
I figured it out (it was a hard drive issue) and I'm now able to build & run.

Quick question while I'm at it: 
The session module seems to belong to the main process, so I'm unsure what pathway this should be communicated to renderer processes. I've concluded that since I'm not seeing any reference from electron_api_session to the blink namespace.

lishid:
I've also confirmed that the `WebSecurityPolicy::AddOriginAccessAllowListEntry()` indeed works when placed in renderer_client_base.cc around this area: https://github.com/electron/electron/blob/c6c022dc46258b439c311da7feaa15d24dc08b16/shell/renderer/renderer_client_base.cc#L230

```c++
blink::WebURL webUrl(GURL("app://./"));
blink::WebSecurityPolicy::AddOriginAccessAllowListEntry(
  webUrl, "file", "", 
  network::mojom::CorsDomainMatchMode::kAllowSubdomains,
  network::mojom::CorsPortMatchMode::kAllowAnyPort,
  network::mojom::CorsOriginAccessMatchPriority::kDefaultPriority
);
```

There's a few sanitization I'll need to perform (ex: input url must contain host) to avoid crashes, but adding this indeed solves the `Not allowed to load local resource` issue.

lishid:
After much exploring in the source code, it still seems the only place that uses WebSecurityPolicy is from `renderer_client_base` https://github.com/electron/electron/blob/c6c022dc46258b439c311da7feaa15d24dc08b16/shell/renderer/renderer_client_base.cc#L39

I haven't yet found a good example of an Electron API that accesses something similar from the main process. I'm hoping someone can point me to one such example so I can ~~copy paste~~ get some inspirations.

deepak1556:
Sorry for the delayed response

Yes `WebSecurityPolicy` is only meant to be accessed on the renderer side and the `RendererClientBase::RenderThreadStarted` is ideal point to invoke it, mainly because the policies need to be configured before the document starts loading.

Currently we have protocol module which configures something similar on the renderer side via [RegisterSchemeAsPrivileged](https://github.com/electron/electron/blob/master/shell/browser/api/electron_api_protocol.cc#L91) through command line [ElectronBrowserClient::AppendExtraCommandLineSwitches](https://github.com/electron/electron/blob/master/shell/browser/electron_browser_client.cc#L740-L752)

Reason I asked the api to be on session is because the origin list can be shared across different webContents using similar session. Also we don't want this api on the renderer side (for ex: via webFrame) because the renderer contents can be untrusted and should not get access to this powerful api.

For this particular api the CORS check is eventually performed by the network service process https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/platform/weborigin/security_policy.cc;l=58 , so we don't have to pass this to the renderer and call the WebSecurityPolicy api.

There is way to configure the origin list per network context via the context create param https://source.chromium.org/chromium/chromium/src/+/master:services/network/public/mojom/network_context.mojom;l=429 , all we need to do is setup [ElectronBrowserContext::GetSharedCorsOriginAccessList](https://source.chromium.org/chromium/chromium/src/+/master:content/public/browser/browser_context.h;l=323) and the session module api can call [SharedCorsOriginAccessList::SetForOrigin](https://source.chromium.org/chromium/chromium/src/+/master:content/public/browser/shared_cors_origin_access_list.h;l=48)

deepak1556:
Since this api might require a patch to `network_context.mojom` for updating the access list dynamically and also the api form needs a bit more discussion, looping in @electron/wg-api members

lishid:
Sorry for the lack of activity - shifting priorities. Though I'll get back to more experimenting and come up with a new commit soon. Thanks again for all the assistance.

lishid:
Ok so I've hooked up a method in the session API that calls `browser_context()->SetCorsOriginAccessListForOrigin`.
Next, I've modified `ElectronBrowserContext::SetCorsOriginAccessListForOrigin` and `GetSharedCorsOriginAccessList` using the source in Chromium (https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/profiles/profile_impl.cc;l=1275?q=SetCorsOriginAccessListForOrigin&ss=chromium%2Fchromium%2Fsrc)

If I understand the code correctly, `SetCorsOriginAccessListForOrigin` calls `CorsOriginPatternSetter::SetLists` on each partition, which proxies to `partition->GetNetworkContext()->SetCorsOriginAccessListsForOrigin`.

This doesn't seem to do anything in my testing just yet, so I'm not too sure what the next step is.
In your last reply, you mentioned that we'll need to patch and use `network_context.mojom`, so maybe that's what I'm missing? I was under the impression that `partition->GetNetworkContext()->SetCorsOriginAccessListsForOrigin` would apply it to the `NetworkContext`.

Perhaps I should gather and commit my changes just so we can look at them? Let me know.

deepak1556:
Yup it would be better to push your changes, thanks!

lishid:
The API is still there just for testing purpose, will change it to whatever makes sense once the team approves the method and agrees on an API design.

deepak1556:
I haven't gotten around to this PR yet, will have a look at the implementation tomorrow. Thanks!

lishid:
Oops wrong branch... Sorry about that

lishid:
Quick ping and also to clarify this code doesn't actually achieve the intended behavior just yet. I have not yet find out why that is.

jkleinsc:
The @electron/wg-api reviewed this at the August 24, 2020 meeting.

deepak1556:
Hi this PR was discussed today in the api-WG meeting, couple of questions asked there that I lacked context.

1) For the use case mentioned https://github.com/electron/electron/pull/24849#issuecomment-669503460 , have you tried using the webrequest api to resolve the file paths via custom protocol ? If that didn't work, would fixing it help your use case ?

2) For the markdown case where file paths are possible, any reason the page is not served over file protocol ?

the api is seen as a risk since it is an easy foot gun that apps can misuse to bypass the origin security policies laid out by chromium and the api-wg would like more context on why the existing solutions don't work. Thanks!

lishid:
### 1. For the use case mentioned #24849 (comment) , have you tried using the webrequest api to resolve the file paths via custom protocol ? If that didn't work, would fixing it help your use case?

I have tried that as written in the original PR:

> Things I've tried:
> - ...
> - session.defaultSession.webRequest.onBeforeRequest

I think if that API worked, it would be a reasonable workaround to solve the problem.

### 2. For the markdown case where file paths are possible, any reason the page is not served over file protocol ?

My app runs on a custom "app://" protocol for several of the reasons highlighted in a previous comment https://github.com/electron/electron/pull/24849#issuecomment-669503460

> If I understand correctly, custom protocol provides many advantages over file:/// for serving app resources:
> - It can be registered as a standard scheme to allow absolute paths to resolve properly
> - It can be registered as a secure protocol to access various 'secure' gated features, such as SubtleCrypto
> - It sets a Origin properly when connecting with remote resources behind https://, which can be used to manage CORS securely.

- Standard schemes for absolute paths: `<script src="/lib/app.js"></script>`
- Secure scheme to access SubtleCrypto `crypto.subtle.digest('SHA-256', data);`
- Set an origin properly `Origin: app://obsidian.md` which is checked my an API server.

### the api is seen as a risk since it is an easy foot gun that apps can misuse to bypass the origin security policies laid out by chromium and the api-wg would like more context on why the existing solutions don't work.

The current existing solution everyone seems to use is `webSecurity: false`, which disables all origin security policies. I would argue this is far worse than misuse from this API. See https://github.com/electron/electron/issues/23757 for examples of many apps/developers who would benefit from this API to securely allow access to `file:///` without disabling webSecurity.

deepak1556:
Thanks for the response, looping in @nornagon for additional thoughts.

lishid:
Yup, I would argue that `webSecurity: false` is actually quite common practice, despite clearly marked as a bad one which has serious security implications. This is most likely because it's the only thing that works, and developers in my knowledge don't really have the time to make a PR for a big project like Electron to get their job done.

lishid:
API design aside, does anyone have an idea why the current code **doesn't** actually work? From the looks of it, that should work, but in practice this doesn't seem to accomplish anything. I don't know the underlying code well enough to tell what I'm missing.

lishid:
@deepak1556 Just a quick follow-up, is there any way I can make progress on this PR? Thanks!

deepak1556:
@lishid sorry do the delayed response, the hold is mainly because I was gonna further evaluate if your use case can be solved without adding this api.

Looked a bit more into the case described in https://github.com/electron/electron/pull/24849#issuecomment-669503460 and vscode has a similar use case for its builtin markdown preview. It is currently done by extensions in this app, and extensions are hosted under `<webview>` which are served over a custom protocol, while the main app resources are under a different protocol. But the catch is they control the rendering of markdown and where the resources are anchored at.

If an app is building a super generic html previewer then it is not possible to achieve the same result without a processor that rewrites all the src points in the markup, certainly can have interesting edge cases.

Given this @nornagon I think it is safe move forward with this proposal unless there are other workarounds for the said use case. The api implementation needs work but that is another discussion.

lishid:
Thank you for the response. It's understandable given you guys have a lot on your plates!

Let me know if there's anything I can do to help 😄

deepak1556:
@lishid i have cleaned up the PR a bit, the api now looks like

```js
await session.setCorsOriginAccessList(
   "custom://abc/hello.html",
   [ "https://*.github.com/*", "*://electron.github.io", "file" ], /* allow list */
   [ "http://*/*" ] /* block list */);
```

can you work on writing some tests and adding documentation, so that it can be reviewed again by api-wg next week. Thanks!

lishid:
Will do, thanks!

lishid:
Just added docs, will work on tests tomorrow when my updated sources builds all 14k files again 😂

bpasero:
Hey, is this PR moving on?

lishid:
Doesn't seem like the API has been reviewed a second time yet, probably fallen through the cracks - @jkleinsc any updates?

lishid:
Hey @deepak1556 @jkleinsc is there anything we can do to push this PR forward?

IMO it's still a serious security issue for Electron because a lot of Electron apps use `webSecurity: false` and/or `bypassCSP: true` as the current workaround for the lack of this API.

jkleinsc:
@lishid there are a couple things that are needed for this PR to move forward:
1. Update your branch to the latest from master to resolve the conflicting files:
2. Resolve the lint errors - you can verify linting locally by running `npm run lint`
3. The API WG needs to approve this PR.  I'll ping the group today on this.

lishid:
I could not run lint because `cpplint.py: not found`. Tried looking it up and I don't see any cpplint.py in the repo.

I don't see any instructions for cpplint anywhere in any of these docs pages either:
- https://www.electronjs.org/docs/development/pull-requests
- https://www.electronjs.org/docs/development/build-instructions-linux
- https://www.electronjs.org/docs/development/coding-style
- https://github.com/electron/build-tools/search?q=cpplint

Saw a package named cpplint on pip but not sure if that's compatible with the current python 2.7 setup.

Please advise.

lishid:
@deepak1556 I seem to be getting an exception for "conversion failure from" (there's nothing after "from") when I specify the allow list to be `[ "file" ]`. That can't seem right? I'm not familiar with the conversion process, hopefully someone can help. 🙏

lishid:
@deepak1556 Upon further debugging, I believe that the `CanDisplay()` check only uses the local WebSecurityPolicy as a shortcut for checking, and the request only makes it to the shared network manager process if this check passes in the first place. Thus we're pretty much stuck passing it through AppendExtraCommandLineSwitches like we do for schemes.

@jkleinsc Given this scenario, I'd propose a new API inspired by `protocol.registerSchemesAsPrivileged(customSchemes)`, let's say `protocol.registerFileAccessibleOrigin(origin)`, which would similarly be added to as an extra command line switch.

I'd argue that making this API more general purpose (i.e. not just allowing access to `file:///` protocol) doesn't make much sense, because there's little need to have fine grained control over CORS for any other scheme/protocol.

zcbenz:
@lishid Do you mean `setCorsOriginAccessList(origin, ['file'], [])` is not enough to make `origin` access `file:` and we have to add a new `registerFileAccessibleOrigin` API for that purpose?

lishid:
@zcbenz That is correct. In my experiments, only applying the WebSecurityPolicy API directly from the renderer's side had an impact over the `file:///` block. (As described by https://github.com/electron/electron/pull/24849#issuecomment-670825092)

In addition, in my testing, I was able to allow access to `file:///` by only using the WebSecurityPolicy API in the renderer process, without making use of the shared CORS allow list that is currently in this PR.

Some more observations that may be relevant:
- The error for `file:///` protocol access is different from regular CORS errors. It specifically says `Not allowed to load local resource: file:///path/to/resource`
- In the network tab, the request lacks any kind of information, including headers, and the timing tab calculates it to be 20us.
- I've attempted to use any API available for intercept the request, such as registering a protocol handler, and using webRequests, but none of it ever catches the request.

When I traced through the specific error message in Chromium's source, it seemed to be produced through a code path that had very basic checks, which was specially coded for protocols marked as `local`. The only two ways to bypass this check was to either 1) mark the origin protocol as `local` as well, or 2) have `file:` in the WebSecurityPolicy allow list for that origin.

zcbenz:
Thanks for explaining, I'm good with either `setCorsOriginAccessList` or `registerFileAccessibleOrigin` API.

But if we go with `setCorsOriginAccessList`, it should handle `setCorsOriginAccessList(origin, ['file'], [])` as a special case under the hood without involving another API, since most users do not understand the internal security model of blink and it would be very confusing to have to use 2 APIs together to achieve one thing.

lishid:
Sounds good. I will attempt a prototype with just `registerFileAccessibleOrigin` because I think `setCorsOriginAccessList` will most likely go unused, and will just be additional maintenance surface without clear benefit or use-case.

codebytere:
@lishid is this PR still something you're planning on pursuing?

lishid:
Yeah... I haven't had time to follow up but it should definitely be implemented at some point so that people don't just turn off `webSecurity` to achieve the desired effect. I will catch up once I get some more free time on my hands.

miniak:
@lishid do you have any updates on the PR?

lishid:
Apologies, I still have my hands full and haven't been able to make any progress. Feel free to close until someone more capable can implement this.

nornagon:
I'm going to close this for now as there's not been much activity lately. If there's renewed interest in this we can open a new PR.

PR #24863, opened 2020-08-05
refactor: use aws-sdk-js instead of boto

nornagon:
#### Description of Change
`boto` is deprecated, and `boto3` has substantially more dependencies. We don't currently `pip install` anything during CI machine setup, relying instead on `gclient` to fetch deps. `boto3` has 5 more dependencies. Rather than upgrade to `boto3` and add 5 new manually tracked python dependencies to `DEPS`, switch to using `aws-sdk` in JS to upload files to S3, so we can lean on npm.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


script/lib/util.py
```diff
@@ -30,8 +30,7 @@
   os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
 )
 SRC_DIR = os.path.abspath(os.path.join(__file__, '..', '..', '..', '..'))
-BOTO_DIR = os.path.abspath(os.path.join(__file__, '..', '..', '..', 'vendor',
-                                        'boto'))
+BOTO_DIR = os.path.join(ELECTRON_DIR, 'vendor', 'boto')
```
jkleinsc, line 33: Is this needed anymore?

nornagon: it was, until i removed the last usage of boto in the next commit

script/release/release.js
```diff
@@ -218,6 +220,36 @@ function uploadIndexJson () {
   console.log(`${pass} Done uploading index.json to S3.`);
 }
 
+async function mergeShasums (pkgVersion) {
+  // Download individual checksum files for Electron zip files from S3,
+  // concatenate them, and upload to GitHub.
+
+  const bucket = process.env.ELECTRON_S3_BUCKET;
+  const accessKey = process.env.ELECTRON_S3_ACCESS_KEY;
+  const secretKey = process.env.ELECTRON_S3_SECRET_KEY;
```
jkleinsc, line 229: `secretKey` does not appear to be used.

script/lib/s3put.js
```diff
@@ -0,0 +1,38 @@
...
+  file = path.resolve(file);
+  if (file.startsWith(prefix)) file = file.substr(prefix.length);
+  return key_prefix + file.replace(path.sep, '/');
+}
+
+let anErrorOccurred = false;
+function next (done) {
+  const file = files.shift();
+  if (!file) return done();
+  s3.upload({
```
jkleinsc, line 23: Where does s3 get authenticated?

script/release/release.js
nornagon: good catch! fixed.

script/lib/s3put.js
nornagon: It automatically pulls the environment variables: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EnvironmentCredentials.html

nornagon:
hm... seeing "access denied" in the logs. I'm not sure if that's due to the environment (since this is a PR and not the normal release flow) or an issue with the code though. @jkleinsc any insights?

MarshallOfSound:
@nornagon Access denied is not expected here, the release builds do have a real token associated with them and should work

jkleinsc:
Merging as CI failure is known flake.

PR #24864, opened 2020-08-05
chore: add V8 crash information to crashReporter

codebytere:
Backport of https://github.com/electron/electron/pull/24771.

See that PR for more details.

Notes: Added V8 crash message and location information to crashReport parameters.


zcbenz:
The added test seems to be failing in CI.

MarshallOfSound (approved):

PR #24881, opened 2020-08-06
fix: send guid with linux crashes

### LINKED ISSUE #24053: Linux crashReporter does not send "guid" attribute
### Issue Details

* **Electron Version:**
  * All up to and including `9-x-y`
* **Operating System:**
  * Linux
  * Never 😢 

### Expected Behavior

When a crash occurs the `crashReporter` module should send the crash dump to your reporting URL along with some extra metadata including a `guid` attribute that is a UUID for the user sending the crash to you.

### Actual Behavior
No `guid` 😢

nornagon:
#### Description of Change
Fixes #24053.

This saves the crash reporter ID in a file in the crash dumps directory, which roughly matches how Crashpad works.

Chrome stores this as a browser preference, and also has some complicated logic around when they regenerate the token, as it's used for a lot of other metrics-related things. I don't believe it's important for Electron to replicate that logic, especially given that Chrome's Crashpad reporter doesn't seem to do so.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed missing `guid` parameter in Linux crash reports.

### DIFF
--- a/shell/browser/api/electron_api_crash_reporter.cc
+++ b/shell/browser/api/electron_api_crash_reporter.cc
@@ -40,7 +40,11 @@
 #endif
 
 #if defined(OS_LINUX)
+#include "base/containers/span.h"
+#include "base/files/file.h"
+#include "base/guid.h"
 #include "components/crash/core/app/breakpad_linux.h"
+#include "components/crash/core/common/crash_keys.h"
 #include "v8/include/v8-wasm-trap-handler-posix.h"
 #include "v8/include/v8.h"
 #endif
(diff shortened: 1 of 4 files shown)


nornagon:
NB. not backporting to 8.x as that version predates the crashReporter refactor.

shell/browser/api/electron_api_crash_reporter.cc
```diff
@@ -81,6 +85,40 @@ bool IsCrashReporterEnabled() {
...
+}
+
+std::string ReadClientId() {
+  base::ThreadRestrictions::ScopedAllowIO allow_io;
+  std::string client_id;
+  // "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".length == 36
+  if (!base::ReadFileToStringWithMaxSize(GetClientIdPath(), &client_id, 36) ||
+      client_id.size() != 36)
+    return std::string();
+  return client_id;
```
MarshallOfSound, line 102: Do we have to check if it is a valid UUID or will breakpad take any 36 char string?

nornagon: i don't think breakpad cares at all what this string is

PR #24895, opened 2020-08-07
build: update git-export-patches to strip hunk header function names

MarshallOfSound:
This portion of the diff hunk header is only used by humans not by git itself, it's designed to make patches slightly more readable but it is also generated by user-specified regexps that vary from git version to git version especially across different OS's (macOS vs linux).  This makes exporting patches even more deterministic 😄 

For reviewers, first commit is the actual python change.  Second commit is running `e patches all` to re-export

Notes: no-notes


MarshallOfSound:
> hunk headers are useful context when reading patches and IMO it's sad to lose them. Perhaps we could just strip .mm files?

The issue here is anyone can define custom regexps for any file, the only case we've found thats Out Of The Box is the difference on macOS with `.mm`, `.m` and `.h` files.  But I assume this _could_ happen with any file depending on your local git config, and the whole point of these scripts is we want it to be consistent output regardless of local settings

MarshallOfSound:
Closing as true source of issue was located

PR #24937, opened 2020-08-11
feat: Added support for all proxy modes

LuoJinghua:
#### Description of Change

This pull request introduces following changes:
*  Extended the setProxy() to support all proxy modes including `direct`, `auto_detect`, `fixed_servers`, `pac_script` and `system`.
* Api for reloading proxy configuration.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added support for explicitly specifying `direct`, `auto_detect` or `system` modes in `session.setProxy()`.

### DIFF
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -283,6 +283,27 @@ Writes any unwritten DOMStorage data to disk.
 #### `ses.setProxy(config)`
 
 * `config` Object
+  * `mode` String (optional) - The proxy mode. Should be one of `direct`,
+    `auto_detect`, `pac_script`, `fixed_servers` or `system`. If it's
+    unspecified, it will be automatically determined based on other specified
+    options.
+    * `direct`
+      In direct mode all connections are created directly, without any proxy involved.
+    * `auto_detect`
+      In auto_detect mode the proxy configuration is determined by a PAC script that can
+      be downloaded at http://wpad/wpad.dat.
+    * `pac_script`
+      In pac_script mode the proxy configuration is determined by a PAC script that is
+      retrieved from the URL specified in the `pacScript`. This is the default mode
+      if `pacScript` is specified.
+    * `fixed_servers`
+      In fixed_servers mode the proxy configuration is specified in `proxyRules`.
+      This is the default mode if `proxyRules` is specified.
+    * `system`
+      In system mode the proxy configuration is taken from the operating system.
+      Note that the system mode is different from setting no proxy configuration.
+      In the latter case, Electron falls back to the system settings
+      only if no command-line options influence the proxy configuration.
   * `pacScript` String (optional) - The URL associated with the PAC file.
   * `proxyRules` String (optional) - Rules indicating which proxies to use.
   * `proxyBypassRules` String (optional) - Rules indicating which URLs should
@@ -292,7 +313,7 @@ Returns `Promise<void>` - Resolves when the proxy setting process is complete.
 
 Sets the proxy settings.
 
-When `pacScript` and `proxyRules` are provided together, the `proxyRules`
+When `mode` is unspecified, `pacScript` and `proxyRules` are provided together, the `proxyRules`
 option is ignored and `pacScript` configuration is applied.
 
 You may need `ses.closeAllConnections` to close currently in flight connections to prevent
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -387,6 +387,10 @@ The `proxyBypassRules` is a comma separated list of rules described below:
 
 Returns `Promise<String>` - Resolves with the proxy information for `url`.
 
+#### `ses.forceReloadProxyConfig()`
+
+Returns `Promise<void>` - Resolves when the all internal states of proxy service is reset and the latest proxy configuration is reapplied if it's already available. The pac script will be fetched from `pacScript` again if the proxy mode is `pac_script`.
+
 #### `ses.setDownloadPath(path)`
 
 * `path` String - The download location.
--- a/shell/browser/api/electron_api_session.cc
+++ b/shell/browser/api/electron_api_session.cc
@@ -557,6 +557,19 @@ v8::Local<v8::Promise> Session::SetProxy(gin::Arguments* args) {
   return handle;
 }
 
+v8::Local<v8::Promise> Session::ForceReloadProxyConfig() {
+  v8::Isolate* isolate = JavascriptEnvironment::GetIsolate();
+  gin_helper::Promise<void> promise(isolate);
+  auto handle = promise.GetHandle();
+
+  content::BrowserContext::GetDefaultStoragePartition(browser_context_)
+      ->GetNetworkContext()
+      ->ForceReloadProxyConfig(base::BindOnce(
+          gin_helper::Promise<void>::ResolvePromise, std::move(promise)));
+
+  return handle;
+}
+
 void Session::SetDownloadPath(const base::FilePath& path) {
   browser_context_->prefs()->SetFilePath(prefs::kDownloadDefaultDirectory,
                                          path);
@@ -1125,6 +1138,7 @@ gin::ObjectTemplateBuilder Session::GetObjectTemplateBuilder(
       .SetMethod("clearStorageData", &Session::ClearStorageData)
       .SetMethod("flushStorageData", &Session::FlushStorageData)
       .SetMethod("setProxy", &Session::SetProxy)
+      .SetMethod("forceReloadProxyConfig", &Session::ForceReloadProxyConfig)
       .SetMethod("setDownloadPath", &Session::SetDownloadPath)
       .SetMethod("enableNetworkEmulation", &Session::EnableNetworkEmulation)
       .SetMethod("disableNetworkEmulation", &Session::DisableNetworkEmulation)
--- a/shell/browser/api/electron_api_session.cc
+++ b/shell/browser/api/electron_api_session.cc
@@ -27,6 +27,7 @@
 #include "components/prefs/value_map_pref_store.h"
 #include "components/proxy_config/proxy_config_dictionary.h"
 #include "components/proxy_config/proxy_config_pref_names.h"
+#include "components/proxy_config/proxy_prefs.h"
 #include "content/public/browser/browser_task_traits.h"
 #include "content/public/browser/browser_thread.h"
 #include "content/public/browser/download_item_utils.h"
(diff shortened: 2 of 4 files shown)


deepak1556:
Thanks for the PR!

Can you split the `closeAllConnections()` and `closeIdleConnections()` to a separate PR.

Also can you add some tests, it would be difficult to test `system` and `auto_detect` mode but you can unittest the api and also verify the other modes.

shell/browser/api/electron_api_session.cc
```diff
@@ -427,14 +427,36 @@ v8::Local<v8::Promise> Session::SetProxy(gin::Arguments* args) {
...
+    }
+  }
   // pacScript takes precedence over proxyRules.
-  if (!pac_url.empty()) {
+  if (mode == "direct") {
+    browser_context_->in_memory_pref_store()->SetValue(
+        proxy_config::prefs::kProxy,
+        std::make_unique<base::Value>(ProxyConfigDictionary::CreateDirect()),
+        WriteablePrefStore::DEFAULT_PREF_WRITE_FLAGS);
+  } else if (mode == "sysytem") {
```
deepak1556, line 449: "sysytem" => "system"

shell/browser/api/electron_api_session.cc
```diff
@@ -427,14 +427,36 @@ v8::Local<v8::Promise> Session::SetProxy(gin::Arguments* args) {
...
+  if (!options.Get("mode", &mode)) {
+    if (!pac_url.empty()) {
+      mode = "pac_script";
+    } else {
+      mode = "fixed_servers";
+    }
+  }
   // pacScript takes precedence over proxyRules.
-  if (!pac_url.empty()) {
+  if (mode == "direct") {
```
deepak1556, line 444: Can you use [StringToProxyMode](https://source.chromium.org/chromium/chromium/src/+/master:components/proxy_config/proxy_prefs.h;l=65) util and compare against the `ProxyPrefs::ProxyMode` , it will guarantee a level of upstream api breakages wrt the strings used.

LuoJinghua: ok.

LuoJinghua:
> Thanks for the PR!
> 
> Can you split the `closeAllConnections()` and `closeIdleConnections()` to a separate PR.
> 
> Also can you add some tests, it would be difficult to test `system` and `auto_detect` mode but you can unittest the api and also verify the other modes.

Finished, please review my changes.

shell/browser/api/electron_api_session.cc
```diff
@@ -455,6 +487,19 @@ v8::Local<v8::Promise> Session::SetProxy(gin::Arguments* args) {
   return handle;
 }
 
+v8::Local<v8::Promise> Session::ForceReloadProxyConfig() {
+  auto* isolate = v8::Isolate::GetCurrent();
```
deepak1556, line 491: please use `auto* isolate = JavascriptEnvironment::GetIsolate();` instead

spec-main/api-session-spec.ts
```diff
@@ -398,6 +406,50 @@ describe('session module', () => {
...
+
+    it('allows configuring proxy settings with mode `invalid`', async () => {
+      const config = { mode: 'invalid' as any };
+      await expect(customSession.setProxy(config)).to.eventually.be.rejectedWith(/Invalid mode/);
+    });
+
+    it('reload proxy configuration', async () => {
+      const config = { mode: 'fixed_servers' as any, proxyRules: 'http=myproxy:80' };
+      await customSession.setProxy(config);
+      await customSession.forceReloadProxyConfig();
```
deepak1556, line 449: does this reload have any effect here ? I was imagining the reload is required after the config is updated.

```
const config = { mode: 'direct' as any, proxyRules: 'http=myproxy:80' };
await customSession.setProxy(config);
const proxy = await customSession.resolveProxy('http://example.com/');
expect(proxy).to.equal('DIRECT');
const newConfig = { mode: 'fixed_servers' as any, proxyRules: 'http=myproxy:80' };
await customSession.setProxy(config);
await customSession.forceReloadProxyConfig();
const newProxy = await customSession.resolveProxy('http://example.com/');
expect(newProxy).to.equal('PROXY myproxy:80');
```

LuoJinghua: For this case it already works without forceReloadProxyConfig(). I think I just figured a ways to test this functionality.
```javascript
      let proxyPort = 8132;
      server = http.createServer((req, res) => {
        const pac = `
          function FindProxyForURL(url, host) {
            return "PROXY myproxy:${proxyPort}";
          }
        `;
        res.writeHead(200, {
          'Content-Type': 'application/x-ns-proxy-autoconfig'
        });
        res.end(pac);
      });
      await new Promise(resolve => server.listen(0, '127.0.0.1', resolve));
      const config = { mode: 'pac_script' as any, pacScript: `http://127.0.0.1:${(server.address() as AddressInfo).port}` };
      await customSession.setProxy(config);
      {
        const proxy = await customSession.resolveProxy('https://google.com');
        expect(proxy).to.equal(`PROXY myproxy:${proxyPort}`);
      }
      {
        proxyPort = 8133;
        customSession.forceReloadProxyConfig()
        const proxy = await customSession.resolveProxy('https://google.com');
        expect(proxy).to.equal(`PROXY myproxy:${proxyPort}`);
      }
```

deepak1556: yeah that works too, as long as there is a change in configuration within the same session then testing this api makes sense.

shell/browser/api/electron_api_session.cc
LuoJinghua: Ok, got it.

spec-main/api-session-spec.ts
LuoJinghua: > LGTM with minor nits

Fixed, could you please review my changes again.

deepak1556:
changes look good, but this will need review from api working group before being merged.

jkleinsc:
The @electron/wg-api approved this at the August 24, 2020 meeting

LuoJinghua:
> The new APIs looks good to me, but I'd like to know a couple of things:
> 
> 1. Can the mode be set using the already-existing `proxyRules` string? Does the `mode` parameter provide access to functionality that's not currently accessible?


> 2. Can you say a little more about the use-case that motivates this API?
> 
> Thanks for the well-formatted PR (and test cases)!



> The new APIs looks good to me, but I'd like to know a couple of things:
> 
> 1. Can the mode be set using the already-existing `proxyRules` string? Does the `mode` parameter provide access to functionality that's not currently accessible?
If it's unspecified, it will be automatically determined based on other specified options and matches the current behaviour. The mode is set to pac_script if pacScript is specified, otherwise it's set to fixed_servers.

Without the mode parameter: 
* It's unclear how to set the proxy to direct mode, it's much clear to call `setProxy({ mode: 'direct' })` then `setProxy({proxyRules: ''})`
* There's no way to set proxy back to system/auto_detect mode after setting a proxy for a session

> 2. Can you say a little more about the use-case that motivates this API?
> 

Let me quote a part of reply from pull request #24945
> The real world use case is a user want to access to a site requires a proxy and he get a proxy list from somewhere.
> 1. Our product randomly select a proxy from the list and try to navigate the target site
> 2. Do some basic checks to verify whether the proxy is good and the website is accessable
> 3. Goto step 1 if the check failed
> 
> The problem is these proxies maybe from the same provider and the origin is same too. Everything is ok when the origin of desn't match previous one, otherwise the connection is reused when the server supports http 2 or keep-alived connections which are true for most of sites.

There's no way to switch the proxy back to system mode without this change to `setProxy` after finishing the dirty work.

> Thanks for the well-formatted PR (and test cases)!

docs/api/session.md
```diff
@@ -179,6 +179,25 @@ Writes any unwritten DOMStorage data to disk.
...
+    * `direct`
+      In direct mode all connections are created directly, without any proxy involved.
+    * `auto_detect`
+      In auto_detect mode the proxy configuration is determined by a PAC script that can
+      be downloaded at http://wpad/wpad.dat.
+    * `pac_script`
+      In pac_script mode the proxy configuration is determined by a PAC script that is
+      retrieved from the URL specified in the `pacScript`.
+    * `fixed_servers`
+      In fixed_servers mode the proxy configuration is specified in `proxyRules`.
```
nornagon, line 195:
```suggestion
      In fixed_servers mode the proxy configuration is specified in `proxyRules`.
      This is the default mode if `proxyRules` is specified.
```

docs/api/session.md
```diff
@@ -179,6 +179,25 @@ Writes any unwritten DOMStorage data to disk.
...
+    unspecified, it will be automatically determined based on other specified
+    options.
+    * `direct`
+      In direct mode all connections are created directly, without any proxy involved.
+    * `auto_detect`
+      In auto_detect mode the proxy configuration is determined by a PAC script that can
+      be downloaded at http://wpad/wpad.dat.
+    * `pac_script`
+      In pac_script mode the proxy configuration is determined by a PAC script that is
+      retrieved from the URL specified in the `pacScript`.
```
nornagon, line 193:
```suggestion
      retrieved from the URL specified in the `pacScript`. This is the default mode
      if `pacScript` is specified.
```

docs/api/session.md
```diff
@@ -259,6 +278,10 @@ The `proxyBypassRules` is a comma separated list of rules described below:
 
 Returns `Promise<String>` - Resolves with the proxy information for `url`.
 
+#### `ses.forceReloadProxyConfig()`
+
+Returns `Promise<void>` - Resolves when the proxy configuration is reloaded.
```
nornagon, line 283: This needs a description of what the method does.

What configuration is reloaded? From where?

shell/browser/api/electron_api_session.cc
```diff
@@ -427,14 +428,45 @@ v8::Local<v8::Promise> Session::SetProxy(gin::Arguments* args) {
...
+          "Invalid mode, must be one of direct, auto_detect, pac_script, "
+          "fixed_servers or system");
+      return handle;
+    }
+  }
+  if (proxy_mode == ProxyPrefs::MODE_DIRECT) {
+    browser_context_->in_memory_pref_store()->SetValue(
+        proxy_config::prefs::kProxy,
+        std::make_unique<base::Value>(ProxyConfigDictionary::CreateDirect()),
+        WriteablePrefStore::DEFAULT_PREF_WRITE_FLAGS);
```
nornagon, line 457: Let's factor out the common part of these branches. e.g.

```cpp
std::unique_ptr<base::Value> proxy_config;
if (proxy_mode == ProxyPrefs::MODE_DIRECT) {
  proxy_config = ProxyConfigDictionary::CreateDirect();
} else if (...) {
  // ...
}
browser_context_->in_memory_pref_store()->SetValue(
    proxy_config::prefs::kProxy, proxy_config,
    WriteablePrefStore::DEFAULT_PREF_WRITE_FLAGS);
```

nornagon (changes requested):
Thanks for the clarifications, and sorry for dropping the ball on this review. The general shape of the API looks good to me, just a couple of minor comments before we can get this landed.

nornagon (approved):

nornagon:
@LuoJinghua looks like the build is failing, but otherwise lgtm :)

LuoJinghua:
> @LuoJinghua looks like the build is failing, but otherwise lgtm :)

Fixed.

zcbenz (approved):

PR #24945, opened 2020-08-12
feat: add ability to close connections for a session

LuoJinghua:
#### Description of Change
These methods are needed for closing idle and active connections after
switching to another proxy for a session, otherwise these connections may
be reused for future requests which is unexpected for most of users.

this pull request was a part of #24937

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: New api for closing connections for a session

@deepak1556

### DIFF
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -295,6 +295,9 @@ Sets the proxy settings.
 When `pacScript` and `proxyRules` are provided together, the `proxyRules`
 option is ignored and `pacScript` configuration is applied.
 
+You may need `ses.closeAllConnections` to close currently in flight connections to prevent
+pooled sockets using previous proxy from being reused by future requests.
+
 The `proxyRules` has to follow the rules below:
 
 ```sh
@@ -403,6 +406,12 @@ window.webContents.session.enableNetworkEmulation({ offline: true })
   * `numSockets` Number (optional) - number of sockets to preconnect. Must be between 1 and 6. Defaults to 1.
 
 Preconnects the given number of sockets to an origin.
+
+#### `ses.closeAllConnections()`
+
+Returns `Promise<void>` - Resolves when all connections are closed.
+
+**Note:** It will terminate / fail all requests currently in flight.
 
 #### `ses.disableNetworkEmulation()`
 
--- a/shell/browser/api/electron_api_session.cc
+++ b/shell/browser/api/electron_api_session.cc
@@ -903,6 +903,19 @@ void Session::Preconnect(const gin_helper::Dictionary& options,
                      url, num_sockets_to_preconnect));
 }
 
+v8::Local<v8::Promise> Session::CloseAllConnections() {
+  v8::Isolate* isolate = JavascriptEnvironment::GetIsolate();
+  gin_helper::Promise<void> promise(isolate);
+  auto handle = promise.GetHandle();
+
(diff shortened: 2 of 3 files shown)


itsananderson:
It sounds like failing to call these methods when switching proxies could actually cause bugs in an app's code. Is there a use-case for having this as a separate API rather than closing out connections automatically when switching to a new proxy?

What happens to any active connections when you call `closeAllConnections()`? Would that cause any pending requests to fail? Would it be more user-friendly to "drain" the connection pool by not assigning new requests to the old connections, and then removing the connections from the pool once their pending requests are complete? I haven't yet dug into the underlying networking code to see how difficult it would be to implement that behavior, but what are your thoughts about implementing the API that way?

jkleinsc:
The @electron/wg-api reviewed this at the August 24, 2020 meeting.

LuoJinghua:
> It sounds like failing to call these methods when switching proxies could actually cause bugs in an app's code. Is there a use-case for having this as a separate API rather than closing out connections automatically when switching to a new proxy?
> 
> What happens to any active connections when you call `closeAllConnections()`? Would that cause any pending requests to fail? Would it be more user-friendly to "drain" the connection pool by not assigning new requests to the old connections, and then removing the connections from the pool once their pending requests are complete? I haven't yet dug into the underlying networking code to see how difficult it would be to implement that behavior, but what are your thoughts about implementing the API that way?

The real world use case is a user want to access to a site requires a proxy and he get a proxy list from somewhere.
1. Our product randomly select a proxy from the list and try to navigate the target site
2. Do some basic checks to verify whether the proxy is good and the website is accessable
3. Goto step 1 if the check failed

The problem is these proxies maybe from the same provider and the origin is same too. Everything is ok when the origin of desn't match previous one, otherwise the connection is reused when the server supports http 2 or keep-alived connections which is true for most of site.

> What happens to any active connections when you call `closeAllConnections()`? Would that cause any pending requests to fail? 

Yes.

> Would it be more user-friendly to "drain" the connection pool by not assigning new requests to the old connections, and then removing the connections from the pool once their pending requests are complete?

We're using the browser view and there's on way to controll the navigation requests. I don't think it's possible to drain these connection pool atomatically since they share the same origin, they're same from view point of chrome network stack. The network stack is happy to reuse these connections since the target websites supports http2 and origin of proxy is the same.
The `closeIdleConections` is enough to drain these connections, works fine for us. We prefer closeAllConections since we want to abort other pending non-navigation requests too. closeIdleConections does what you seggest extractly.

MarshallOfSound:
Hey @LuoJinghua The API WG discussed this and based on your comments and the example use cases we're OK with adding `closeAllConnections` but don't feel `closeIdleConnections` makes sense at this time.  If you remove the idle implementation we can merge this once CI is green 👍 

Also please explicitly document that `closeAllConnections` will terminate / fail all requests currently in flight

LuoJinghua:
Ok

> Hey @LuoJinghua The API WG discussed this and based on your comments and the example use cases we're OK with adding `closeAllConnections` but don't feel `closeIdleConnections` makes sense at this time. If you remove the idle implementation we can merge this once CI is green 👍
> 
> Also please explicitly document that `closeAllConnections` will terminate / fail all requests currently in flight

Ok

codebytere (approved):

LuoJinghua:
Thanks for merging this. Is possible to backport this to stable branches?

Issue #22178, opened 2020-02-13
PDF download button doesn't do anything


nornagon:
When viewing a PDF, clicking the "download" button has no effect. It should trigger a download (including the `"will-download"` event and associated machinery).

aaronsarnat:
Not sure if this is possible, but it would also be cool if the download button could be optionally hidden or disabled if, say for instance, user is reading a PDF from the local file system (e.g. user manual).

agata:
This is very useful. Now, if the user is using BrowserWindow or WebView, the PDF The files are automatically previewed (really great). However, I think it's getting harder to download the PDF files.

vishnu0320:
"dom- ready" console is working but "will-download" is not working ???

```
win.webContents.on('dom-ready', () => {
        console.log("dom ready")
      })
win.webContents.on('will-download', () => {
        console.log("download")
      })
```

PR #24946, opened 2020-08-12
fix: pdf preview download button not working

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/22178.

Before: Nothing.

After:

Screen Shot 2020-08-11 at 10 44 18 PM

Unfortunately this can't be hooked up to `will-download` infrastructure in its current form - it runs through fairly separate logic and adding the ability to intercept and prevent default a la the existing event implementation would require some patching that at this juncture i don't think is in our best interest.

cc @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where the Save button did not function in PDF previews.

### DIFF
--- a/shell/browser/extensions/electron_extensions_api_client.cc
+++ b/shell/browser/extensions/electron_extensions_api_client.cc
@@ -8,8 +8,14 @@
 #include <string>
 
 #include "extensions/browser/guest_view/mime_handler_view/mime_handler_view_guest_delegate.h"
+#include "printing/buildflags/buildflags.h"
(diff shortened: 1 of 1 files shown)


zapjelly:
Curious if anyone else is experiencing this, loading local packaged pdf, save works perfectly when served but when packaged it shows the save prompt but there is no pdf file created.

msgc:
Hi, is there any solution to this saving pdf issue? 
I tested this in PC and Mac and is not working in neither

thank you in advance

```
node -v
v14.16.1

```
```
npm -v
7.15.1
```

Issue #24939, opened 2020-08-11
Why visibleOnFullScreen is deprecated, it works fine on my mac 10.15.4


wyqpzm:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description

My electron version is 6.1.11, and I can't update because visibleOnFullScreen option on setVisibleOnAllWorkspaces is deprecated. It works fine on mac 10.15.4 and newer version.
To code like this:

  app.dock.hide()
  window.setAlwaysOnTop(true, 'floating')
  window.setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true })

The visibleOnFullScreen option is very important to my electron app and I'm so appreciate if you can check this feature again!

codebytere:
Huh, you're right. We removed it because it doesn't work without `app.dock.hide()`, but i can see about reinvigorating it.

wyqpzm:
Thank you so much! I'm looking forward to the new version!

Issue #24011, opened 2020-06-08
Extensions don't work with file:// protocol since 9.0.0


ChALkeR:
Extracted from #23662, https://github.com/electron/electron/issues/23662#issuecomment-632344005, per https://github.com/electron/electron/issues/23662#issuecomment-637077040 (cc @nornagon).

Extensions get loaded for `http://` and `https://` protocols, but don't work for `file://` protocol, which is commonly used.

This also affects devtools extensions.

Appears to be an undocumented regression from 8.x.

Chrome by default behaves the same, but it has an `Allow access to file URLs` switch (the `fileAccess` flag) on extension page at `chrome://extensions/`, which enables the extension to run on `file://` urls (which can be verified on `Content scripts` tab).
Any hint how to do that in Electron?
React-devtools works on `file://` protocol in Chrome with that flag enabled, but does not work in Electron since 9.0.0, as there are no ways to toggle that setting.

### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 9.0.0, 9.0.2
* **Operating System:**
  * `Linux xps 5.6.14-arch1-1 #1 SMP PREEMPT Wed, 20 May 2020 20:43:19 +0000 x86_64 GNU/Linux`
* **Last Known Working Electron version:**
  * 8.3.1

### Expected Behavior

Installed extensions should load for `file://` protocol automatically, or/and there should be an opt-in/out-out for that (like opt-in in Chrome).

Extension content script present in `Content scripts` tab.

### Actual Behavior

Extensions are not loaded if the for pages requested over `file://` protocol.

No extension content script present in `Content scripts` tab.

### To Reproduce

`index.js`:
```js
const { app, session, BrowserWindow } = require('electron');
const path = require('path');

(async () => {
  await app.whenReady()
  await session.defaultSession.loadExtension(path.join(__dirname, 'extensions/test'))
  const bw = new BrowserWindow({webPreferences: {sandbox: true}})
  //await bw.loadURL('about:blank')
  //await bw.loadURL('https://example.org/')
  await bw.loadURL(`file://${path.resolve('simple.html')}`)
  bw.show()
[...]

PR #24956, opened 2020-08-12
feat: reinvigorate visibleOnFullscreen option

codebytere:
#### Description of Change

Refs https://github.com/electron/electron/pull/21706.

Closes https://github.com/electron/electron/issues/24939.

We originally removed this as we could not determine a way to appropriately compensate for Apple's changes to the underlying functionality of NSWindows starting with 10.14 to disallow NSWindows from floating on top of fullscreen apps. At the time we weren't aware of a possible workaround with `app.dock.hide` but now that this is known, and presents value to a nontrivial number of apps, I feel that it would be appropriate to allow apps access to this functionality once more.

Tested on macOS Catalina and observed working as intended.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added back a previously broken `visibleOnFullScreen` option for `setVisibleOnAllWorkspaces`.

### DIFF
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -1351,8 +1351,21 @@ void NativeWindowMac::SetProgressBar(double progress,
 void NativeWindowMac::SetOverlayIcon(const gfx::Image& overlay,
                                      const std::string& description) {}
 
-void NativeWindowMac::SetVisibleOnAllWorkspaces(bool visible) {
+void NativeWindowMac::SetVisibleOnAllWorkspaces(bool visible,
+                                                bool visibleOnFullScreen) {
+  // In order for NSWindows to be visible on fullscreen we need to functionally
+  // mimic app.dock.hide() since Apple changed the underlying functionality of
+  // NSWindows starting with 10.14 to disallow NSWindows from floating on top of
+  // fullscreen apps.
+  if (visibleOnFullScreen) {
+    [window_ setCanHide:NO];
+    ProcessSerialNumber psn = {0, kCurrentProcess};
+    TransformProcessType(&psn, kProcessTransformToUIElementApplication);
+  }
+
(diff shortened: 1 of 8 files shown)


shell/browser/native_window_mac.mm
```diff
@@ -1351,8 +1351,21 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
...
                                      const std::string& description) {}
 
-void NativeWindowMac::SetVisibleOnAllWorkspaces(bool visible) {
+void NativeWindowMac::SetVisibleOnAllWorkspaces(bool visible,
+                                                bool visibleOnFullScreen) {
+  // In order for NSWindows to be visible on fullscreen we need to functionally
+  // mimic app.dock.hide() since Apple changed the underlying functionality of
+  // NSWindows starting with 10.14 to disallow NSWindows from floating on top of
+  // fullscreen apps.
+  if (visibleOnFullScreen) {
```
MarshallOfSound, line 1360: Don't we need to do the inverse of this when `visibleOnFullScreen=false`? i.e. we need an `else` block

codebytere:
Updated to ensure dock behavior restored - tested with https://gist.github.com/1188123a35b35e8c9cd46f34a48ae633

codebytere:
This change is macOS-only, so Window and WOA failures are unrelated.

andrewschreiber:
Any chance we could backport this to electron 8? #24011 is a blocker on upgrading past 8 for now for many of us.

ajay-pipaliya-simform:
> appreciate for bringing this back, however it doesnt seem to work on electron 14 for me
> 
> ```
> win.setAlwaysOnTop(true, 'pop-up-menu');
> win.setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true });
> ```
> 
> did it work for you guys?

Nope, if you're using chrome browser in full-screen mode and try to toggle electron app it will show the browserwindow but when you move your cursor towards the browser window it will loss the focus and hide the browser window

Issue #22393, opened 2020-02-26
Frameless minimum sized window is not restored to previous size after minimize and restore


christian-judt:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.1.10 - 9.0.0-beta.3 (and presumably later versions until this bug is fixed)
* **Operating System:**
  * Windows 10 pro (1909)
* **Last Known Working Electron version:**
  * 7.1.9

### Preconditions

* Window is frameless
* Window has minimum size
* Window size is minimum size

### Expected behavior

* Window is minimized
* Window is restored
* Window has the same size as before the minimization

### Actual behavior

* Window is minimized
* Window is restored
* Window is slightly higher and wider than before the minimization

### To Reproduce

```
git clone -b frameless-minsize-minimize-restore-bug https://github.com/christian-judt/electron-quick-start.git
cd electron-quick-start
npm install
npm start
```

### Additional Information

* A similar problem with "maximize-unmaximize" instead of "minimize-restore" has appeared in Electron Version 7.1.2. (see #22440)

### Maybe related issues

* Windows: Restore (after minimize) to wrong window size #7951
* Frameless minimum sized window is not restored to previous size after maximize and unmaximize #13533
* maximize, minimize, restore incorrect size #15702
* Electron 6 - restore after minimize has wrong window size #19816

tallev266:
Glad to see this issue is reported.
Any plans on fixing this for version 8?
If not, are there any workarounds we can implement other than saving the bounds on minimize and resetting them on restore?

orangeChu:
> Glad to see this issue is reported.
> Any plans on fixing this for version 8?
> If not, are there any workarounds we can implement other than saving the bounds on minimize and resetting them on restore?

Maybe possible to temporarily solve this problem by resetting the size of the window when the window's `restore` event is activated.

pejman1:
Does anybody know, how long does it take to fix ?

Issue #22440, opened 2020-02-28
Frameless minimum sized window is not restored to previous size after maximize and unmaximize


christian-judt:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.1.2 - 9.0.0-beta.3 (and presumably later versions until this bug is fixed)
* **Operating System:**
  * Windows 10 pro (1909)
* **Last Known Working Electron version:**
  * 7.1.1

### Preconditions

* Window is frameless
* Window has minimum size
* Window size is minimum size

### Expected behavior

* Window is maximized
* Window is unmaximized
* Window has the same size as before the maximization

### Actual Behavior
* Window is maximized
* Window is unmaximized
* Window is slightly higher and wider than before the maximization

### To Reproduce

```
git clone -b frameless-minsize-maximize-unmaximize-bug https://github.com/christian-judt/electron-quick-start.git
cd electron-quick-start
npm install
npm start
```

### Additional Information

* This seems to be a reappearing problem. (for the last occurence see #13533)
* A similar problem with "minimize-restore" instead of "maximize-unmaximize" has appeared in Electron Version 7.1.10. (see #22393)

### Other maybe related issues

* [Windows] Frameless window size increase on a maximize, unmaximize sequence #2498
* Windows: Restore (after minimize) to wrong window size #7951
* maximize, minimize, restore incorrect size #15702
* Electron 6 - restore after minimize has wrong window size #19816

aleksey-hoffman:
I reproduced it  in version `8.0.3`. I see the same behavior:

- Resized window to the defined minimum (500x300)
- Maximize, un-maximize
- The window size changes to 515x315

```JS
mainWindow = new BrowserWindow({
    title: "Title",
    width: 1280,
    height: 720,
    minWidth: 500,
    minHeight: 300,
    frame: false,
    transparent: false,
    resizable: true
})
```

PsykickSam:
I made a solution for window resizing, I don't know the problem but It may cause for minWidth and minHeight so I remove the minWidth and minHeight from BrowserWindow and add an resize event, 

mainWindow.on("resize", () => {
[...]

PR #25014, opened 2020-08-18
fix: calculate frame when setting window placement

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/22393.

When calling `SetWindowPlacement`, we should let Chromium handle the `WM_NCCALCSIZE` message otherwise the window size would be wrong for frameless window.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix frameless window's size being changed when restored from minimized state.

### DIFF
--- a/spec-main/api-browser-window-spec.ts
+++ b/spec-main/api-browser-window-spec.ts
@@ -1005,6 +1005,26 @@ describe('BrowserWindow module', () => {
           await restore;
           expectBoundsEqual(w.getNormalBounds(), bounds);
         });
+        it('does not change size for a frameless window with min size', async () => {
+          w.destroy();
+          w = new BrowserWindow({
+            show: false,
+            frame: false,
+            width: 300,
+            height: 300,
+            minWidth: 300,
+            minHeight: 300
+          });
+          const bounds = w.getBounds();
+          w.once('minimize', () => {
+            w.restore();
+          });
+          const restore = emittedOnce(w, 'restore');
+          w.show();
+          w.minimize();
+          await restore;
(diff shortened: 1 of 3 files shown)


spec-main/api-browser-window-spec.ts
```diff
@@ -1005,6 +1005,24 @@ describe('BrowserWindow module', () => {
           await restore;
           expectBoundsEqual(w.getNormalBounds(), bounds);
         });
+        it('does not change size for a frameless window', async () => {
+          w.destroy();
+          w = new BrowserWindow({
+            show: false,
+            frame: false,
+            width: 300,
+            height: 300
```
deepak1556, line 1014: I was only able to repro when the window had `minWidth` and `minHeight` set, were you able to repro without it ?

deepak1556:
This doesn't seem to fix https://github.com/electron/electron/issues/22440 , I thought it would be same codepath for both.

spec-main/api-browser-window-spec.ts
zcbenz: I have updated the test to set min size, they are required for reproducing the bug.

zcbenz:
> This doesn't seem to fix #22440 , I thought it would be same codepath for both.

I confirm this PR does not fix it, I'll take a look at it too.

codebytere:
https://ci.appveyor.com/project/electron-bot/electron-ia32-testing/builds/34730405 is green, idk why it's still showing as pending.

deepak1556:
@pejman1 it is fixed by https://github.com/electron/electron/pull/25051

PR #25015, opened 2020-08-18
docs: Fixed click handler function arguments of MenuItem

gitcommitshow:
As can be seen on line no. 50 of [`menu-item.js`](https://github.com/electron/electron/blob/v9.2.0/lib/browser/api/menu-item.js), the arguments of click handler function were documented incorrectly. Fixed it.

#### Description of Change
Changed incorrect MenuItem's click handler arguments `click(menuItem, browserWindow, event)` to correct version `click(event, focusedWindow, focusedWebContents)`

#### Checklist

#### Release Notes

Notes: Fixed click handler function schema of MenuItem

### DIFF
--- a/docs/api/menu-item.md
+++ b/docs/api/menu-item.md
@@ -10,10 +10,10 @@ See [`Menu`](menu.md) for examples.
 
 * `options` Object
   * `click` Function (optional) - Will be called with
-    `click(menuItem, browserWindow, event)` when the menu item is clicked.
-    * `menuItem` MenuItem
-    * `browserWindow` [BrowserWindow](browser-window.md) | undefined - This will not be defined if no window is open.
+    `click(event, focusedWindow, focusedWebContents)` when the menu item is clicked.
     * `event` [KeyboardEvent](structures/keyboard-event.md)
+    * `focusedWindow` [BrowserWindow](browser-window.md)  | undefined - This will not be defined if no window is open.
+    * `focusedWebContents` [WebContents](web-contents.md)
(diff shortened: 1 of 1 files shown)


gitcommitshow:
This mistake in documentation had been a cause of a bug that I couldn't track for quite some time

gitcommitshow:
You're right. My bad. I should have tested it earlier and instead of PR, should have posted a question. 

I'm still trying to get my head around the code of menu-item, roles and electron.binding('menu'). 
And the docs for this seems confusing to me https://github.com/electron/electron/blob/v9.2.1/docs/api/menu-item.md

I will post question with more details after testing on desktop.

EricBalingit:
I've arrived after this change in the docs and am unable to programmatically trigger a click on a menuItem, no matter what arguments I provide to the click handler.

I always get the error:
Uncaught Exception:
TypeError: Assignment to constant variable.
    at click ( ...
    at MenuItem.click ( ...

additionally, the documentation on KeyboardEvent, as the event argument is described, suggest it can be any object because all properties are optional, however if it does require a typed object, there appears to be no documentation on how to construct any type of event.

[](https://www.electronjs.org/docs/latest/api/menu-item#menuitemclick)

[](https://www.electronjs.org/docs/latest/api/structures/keyboard-event)

EricBalingit:
Ignore my previous comment, my mistake.

Though this version of the docs needs clarification.

reading from here it looks like the click hander is called with a different signature than later in the page when it talks about triggering the handler manually.

https://www.electronjs.org/docs/latest/api/menu-item#new-menuitemoptions
> click Function (optional) - Will be called with click(menuItem, browserWindow, event) when the menu item is clicked.

and here
https://www.electronjs.org/docs/latest/api/menu-item#menuitemclick
> A Function that is fired when the MenuItem receives a click event. It can be called with menuItem.click(event, focusedWindow, focusedWebContents).

PR #25023, opened 2020-08-18
fix: backport performance fix for CSS counters

MarshallOfSound:
Refs: https://chromium-review.googlesource.com/c/chromium/src/+/2363252

Notes: Fixed performance regression with CSS counters


nornagon:
We'll pick this up in a roll, right? any reason to backport early?

MarshallOfSound:
@nornagon This won't get picked up in `10-x-y`, it will get picked up in master in the next roll though

nornagon:
Merge requested upstream: https://bugs.chromium.org/p/chromium/issues/detail?id=1112978#c7

Issue #24755, opened 2020-07-28
Packages cannot be selected in <input file="type"> file selector on macOS


simoncozens:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** v9.0.4
* **Operating System:** mac OS 10.14.6
* **Last Known Working Electron version:**:  2.0.2

### Expected Behavior

[Document packages](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/DocumentPackages/DocumentPackages.html#//apple_ref/doc/uid/10000123i-CH106-SW1) should be selectable in an `<input type="file">`. (As they are in Firefox, Safari and Chrome.)

### Actual Behavior

Packages cannot be selected in an `<input type="file">`.

### Additional Information

This is the exact opposite of #13219. Sometimes you *do* want the user to be able to choose a package file - not for apps, perhaps, but certainly for `.pkg`, old-style or large `.pages`, `.keynote` and `.numbers` files, or in the case I'm working with, `.ufo` font source files. Currently, if I have `<input type="file" accept=".ufo"/>`, I can't select the package itself because the dialog opens it as a directory and shows the contained files, and if I have `<input type="file" accept=".ufo" directory/>`, I can't select the package itself because the dialog treats it as a file and not a directory! 🤷

#13220, which addressed #13219, decided to "patch out the broken behavior and look into getting the optimum behavior later". But I don't think it was ever looked into.

ckerr:
(Am I reading between the lines correctly that you're upgrading an app from Electron 2 to Electron 9?! That's impressive! :1st_place_medal:)

TBH I am not very familiar with this piece of the code. It looks like @miniak and @MarshallOfSound did the work on #13219 / #13220, so pinging them here for eyeballs, but I @simoncozens I also wanted to ask you if it would be possible to write an isolated test case, e.g. from a fork of [electron-quick-start](https://github.com/electron/electron-quick-start).

Ordinarily a repro request gets the `blocked/needs-repro` label but I'd like @miniak or @MarshallOfSound to have a chance to sound off on whether the existing info is enough.

[...]

PR #25030, opened 2020-08-19
fix: transparently package bundles as zip archives

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/24755.

Refactors and cleans up our input element handing for files and directories. Moves `FileSelectHelper` to its own fileset and hooks up functionality to transparently package bundles on macOS into zip archives.

Screen Shot 2020-08-18 at 8 50 13 PM

Tested with https://gist.github.com/6ed9ee05659956e33c15343e49e26ed3. Also tested that the `multiple` and `webkitdirectory` attributes still work as expected. 

cc @nornagon @zcbenz @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where packages could not be selected with <input file="type"> on macOS.


shell/browser/file_select_helper_mac.mm
```diff
@@ -0,0 +1,136 @@
...
+        file_info.file_path = result;
+        file_info.display_name.append(".zip");
+      }
+    }
+  }
+
+  content::GetUIThreadTaskRunner({})->PostTask(
+      FROM_HERE,
+      base::BindOnce(&FileSelectHelper::ProcessSelectedFilesMacOnUIThread,
+                     base::Unretained(this), files_out, temporary_files));
```
deepak1556, line 122: `FileSelectHelper` is a refcounted class, how is it guaranteed here that the object will outlive this call ? I see that the chromium implementation explicitly manages the lifetime for the class, we should do the same https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/file_select_helper.cc;l=566-571.

shell/browser/file_select_helper_mac.mm
```diff
@@ -0,0 +1,136 @@
+// Copyright 2014 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
deepak1556, line 3: Are there any additional changes apart from the upstream implementation, can instead include `//chrome/browser/file_select_helper_mac.mm` ?

shell/browser/file_select_helper.cc
```diff
@@ -0,0 +1,244 @@
...
+
+// net::DirectoryLister::DirectoryListerDelegate
+void FileSelectHelper::OnListDone(int error) {
+  std::vector<FileChooserFileInfoPtr> file_info;
+  for (const auto& path : lister_paths_)
+    file_info.push_back(FileChooserFileInfo::NewNativeFile(
+        NativeFileInfo::New(path, base::string16())));
+
+  OnFilesSelected(std::move(file_info), lister_base_dir_);
+  Release();
```
deepak1556, line 110: move the `Release` call to `OnFilesSelected`, it will also help `AddRed` added for `ShowOpenDialog` , `ShowSaveDialog`

shell/browser/file_select_helper.cc
```diff
@@ -0,0 +1,244 @@
...
+void FileSelectHelper::ShowOpenDialog(
+    const file_dialog::DialogSettings& settings) {
+  v8::Isolate* isolate = electron::JavascriptEnvironment::GetIsolate();
+  v8::HandleScope scope(isolate);
+  gin_helper::Promise<gin_helper::Dictionary> promise(isolate);
+
+  auto callback = base::BindOnce(&FileSelectHelper::OnOpenDialogDone, this);
+  ignore_result(promise.Then(std::move(callback)));
+
+  file_dialog::ShowOpenDialog(settings, std::move(promise));
```
deepak1556, line 63: explicit `AddRef` here.

shell/browser/file_select_helper.cc
```diff
@@ -0,0 +1,244 @@
...
+void FileSelectHelper::ShowSaveDialog(
+    const file_dialog::DialogSettings& settings) {
+  v8::Isolate* isolate = electron::JavascriptEnvironment::GetIsolate();
+  v8::HandleScope scope(isolate);
+  gin_helper::Promise<gin_helper::Dictionary> promise(isolate);
+
+  auto callback = base::BindOnce(&FileSelectHelper::OnSaveDialogDone, this);
+  ignore_result(promise.Then(std::move(callback)));
+
+  file_dialog::ShowSaveDialog(settings, std::move(promise));
```
deepak1556, line 75: explicit `AddRef` here.

shell/browser/file_select_helper.cc
```diff
@@ -0,0 +1,244 @@
...
+  // until web_contents_ is destroyed, so that this instance can delete the
+  // temporary files.
+  if (!temporary_files_.empty())
+    return;
+
+  if (listener_)
+    listener_->FileSelectionCanceled();
+
+  render_frame_host_ = nullptr;
+  web_contents_ = nullptr;
```
deepak1556, line 99: explicit `Release` for file selection cancelled case.

shell/browser/file_select_helper.cc
```diff
@@ -0,0 +1,244 @@
...
+
+  lister_paths_.push_back(data.path);
+}
+
+void FileSelectHelper::RunSelectionEnd() {
+  // If there are temporary files, then this instance needs to stick around
+  // until web_contents_ is destroyed, so that this instance can delete the
+  // temporary files.
+  if (!temporary_files_.empty())
+    return;
```
deepak1556, line 93: why does the instance have to wait until web_contents_ is destroyed, can't we do the following ?

```
if (!temporary_files_.empty()) {
  DeleteTemporaryFiles();
  Release();
}
```

shell/browser/file_select_helper.cc
```diff
@@ -0,0 +1,244 @@
...
+  if (deleted_host == render_frame_host_)
+    render_frame_host_ = nullptr;
+}
+
+// content::WebContentsObserver:
+void FileSelectHelper::WebContentsDestroyed() {
+  render_frame_host_ = nullptr;
+  web_contents_ = nullptr;
+
+  DeleteTemporaryFiles();
```
deepak1556, line 243: explicit `Release` here

shell/browser/file_select_helper_mac.mm
codebytere: @deepak1556 we can potentially - we'd need to patch though :(

```
In file included from ../../chrome/browser/file_select_helper_mac.mm:5:
In file included from ../../chrome/browser/file_select_helper.h:27:
In file included from ../../chrome/browser/safe_browsing/cloud_content_scanning/deep_scanning_dialog_delegate.h:19:
../../chrome/browser/enterprise/connectors/common.h:11:10: fatal error: 'components/enterprise/common/proto/connectors.pb.h' file not found
#include "components/enterprise/common/proto/connectors.pb.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

shell/browser/file_select_helper.cc
codebytere: @deepak1556 the `temporary` is a bit of a misnomer - these files are the zipped files stored in the `zip_cache`. They should remain alive as long as the webcontents is alive, since that's likely how long the user would in some fashion need to interact with them

codebytere:
@nornagon see https://github.com/electron/electron/pull/25030#discussion_r473289490

shell/browser/file_select_helper_mac.mm
nornagon: It looks like we could set `BUILDFLAG(FULL_SAFE_BROWSING)` to false in order to omit that include...

essentially just `safe_browsing_mode = 0` in args.gn.

codebytere:
Note that i ended up putting back `file_select_helper_mac` owing to the fact that it includes `chrome/browser/file_select_helper.h` instead of `shell/browser/file_select_helper.h` and we would thus have had to patch. As we want to keep patch count down and this is a small surface area unlikely to change, I think it's probably a better choice to just copy it.

deepak1556:
while testing found the following DCHECK that can also be seen on chrome canary,

* Launch the test app
* Open a `.app` file that is relative large, for my example tried `Docker.app` or you can select multiple  `*.app` that would end up a large zip cache
* It will take a while to populate the zip
* Close the app or navigate away
* DCHECK!!!

```
FATAL:file_chooser_impl.cc(21)] Check failed: was_file_select_listener_function_called_. Must call either FileSelectListener::FileSelected() or FileSelectListener::FileSelectionCanceled()
0   Electron Framework                  0x000000011c1529f9 base::debug::CollectStackTrace(void**, unsigned long) + 9
1   Electron Framework                  0x000000011c04d073 base::debug::StackTrace::StackTrace() + 19
2   Electron Framework                  0x000000011c06bb15 logging::LogMessage::~LogMessage() + 261
3   Electron Framework                  0x000000011c06c81e logging::LogMessage::~LogMessage() + 14
4   Electron Framework                  0x000000011b186441 content::FileChooserImpl::FileSelectListenerImpl::~FileSelectListenerImpl() + 97
5   Electron Framework                  0x000000011b1864ce content::FileChooserImpl::FileSelectListenerImpl::~FileSelectListenerImpl() + 14
6   Electron Framework                  0x000000011b187197 content::FileChooserImpl::~FileChooserImpl() + 103
7   Electron Framework                  0x000000011b187f03 mojo::StrongBinding<blink::mojom::FileChooser>::Close() + 83
8   Electron Framework                  0x000000011b187c3c mojo::StrongBinding<blink::mojom::FileChooser>::OnConnectionError(unsigned int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 76
9   Electron Framework                  0x000000011c5f3155 mojo::InterfaceEndpointClient::NotifyError(base::Optional<mojo::DisconnectReason> const&) + 357
10  Electron Framework                  0x000000011c5fc3df mojo::internal::MultiplexRouter::ProcessNotifyErrorTask(mojo::internal::MultiplexRouter::Task*, mojo::internal::MultiplexRouter::ClientCallBehavior, base::SequencedTaskRunner*) + 447
11  Electron Framework                  0x000000011c5f936e mojo::internal::MultiplexRouter::ProcessTasks(mojo::internal::MultiplexRouter::ClientCallBehavior, base::SequencedTaskRunner*) + 878
12  Electron Framework                  0x000000011c5f7942 mojo::internal::MultiplexRouter::OnPipeConnectionError(bool) + 914
13  Electron Framework                  0x000000011c5ea31e mojo::Connector::HandleError(bool, bool) + 494
14  Electron Framework                  0x000000011c5eb501 mojo::Connector::OnHandleReadyInternal(unsigned int) + 113
15  Electron Framework                  0x00000001168b9820 mojo::SimpleWatcher::DiscardReadyState(base::RepeatingCallback<void (unsigned int)> const&, unsigned int, mojo::HandleSignalsState const&) + 80
16  Electron Framework                  0x000000011c61d5da mojo::SimpleWatcher::OnHandleReady(int, unsigned int, mojo::HandleSignalsState const&) + 410
17  Electron Framework                  0x000000011c61dacb base::internal::Invoker<base::internal::BindState<void (mojo::SimpleWatcher::*)(int, unsigned int, mojo::HandleSignalsState const&), base::WeakPtr<mojo::SimpleWatcher>, int, unsigned int, mojo::HandleSignalsState>, void ()>::RunOnce(base::internal::BindStateBase*) + 171
18  Electron Framework                  0x000000011c0e2a3b base::TaskAnnotator::RunTask(char const*, base::PendingTask*) + 411
19  Electron Framework                  0x000000011c101a27 base::sequence_manager::internal::ThreadControllerWithMessagePumpImpl::DoWorkImpl(base::sequence_manager::LazyNow*) + 519
20  Electron Framework                  0x000000011c101698 base::sequence_manager::internal::ThreadControllerWithMessagePumpImpl::DoWork() + 168
21  Electron Framework                  0x000000011c173191 base::MessagePumpCFRunLoopBase::RunWork() + 65
22  Electron Framework                  0x000000011c168222 base::mac::CallWithEHFrame(void () block_pointer) + 10
23  Electron Framework                  0x000000011c172b5f base::MessagePumpCFRunLoopBase::RunWorkSource(void*) + 63
24  CoreFoundation                      0x00007fff3251ed52 __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 17
25  CoreFoundation                      0x00007fff3251ecf1 __CFRunLoopDoSource0 + 103
26  CoreFoundation                      0x00007fff3251eb0b __CFRunLoopDoSources0 + 209
27  CoreFoundation                      0x00007fff3251d83a __CFRunLoopRun + 927
28  CoreFoundation                      0x00007fff3251ce3e CFRunLoopRunSpecific + 462
29  HIToolbox                           0x00007fff31149abd RunCurrentEventLoopInMode + 292
30  HIToolbox                           0x00007fff311497d5 ReceiveNextEventCommon + 584
31  HIToolbox                           0x00007fff31149579 _BlockUntilNextEventMatchingListInModeWithFilter + 64
32  AppKit                              0x00007fff2f78f039 _DPSNextEvent + 883
33  AppKit                              0x00007fff2f78d880 -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 1352
34  AppKit                              0x00007fff2f77f58e -[NSApplication run] + 658
35  Electron Framework                  0x000000011c17414c base::MessagePumpNSApplication::DoRun(base::MessagePump::Delegate*) + 348
36  Electron Framework                  0x000000011c172662 base::MessagePumpCFRunLoopBase::Run(base::MessagePump::Delegate*) + 130
37  Electron Framework                  0x000000011c10225d base::sequence_manager::internal::ThreadControllerWithMessagePumpImpl::Run(bool, base::TimeDelta) + 349
38  Electron Framework                  0x000000011c0b5bac base::RunLoop::Run() + 956
39  Electron Framework                  0x000000011aa247eb content::BrowserMainLoop::RunMainMessageLoopParts() + 203
40  Electron Framework                  0x000000011aa26bb2 content::BrowserMainRunnerImpl::Run() + 82
41  Electron Framework                  0x000000011aa2130b content::BrowserMain(content::MainFunctionParams const&) + 267
42  Electron Framework                  0x000000011a8371ce content::ContentMainRunnerImpl::RunServiceManager(content::MainFunctionParams&, bool) + 1406
43  Electron Framework                  0x000000011a836c23 content::ContentMainRunnerImpl::Run(bool) + 467
44  Electron Framework                  0x000000011e886a5e service_manager::Main(service_manager::MainParams const&) + 3310
45  Electron Framework                  0x000000011876df68 content::ContentMain(content::ContentMainParams const&) + 120
46  Electron Framework                  0x0000000116863386 ElectronMain + 134
47  Electron                            0x000000010a4f5551 main + 289
48  libdyld.dylib                       0x00007fff6c589cc9 start + 1
Task trace:
0   Electron Framework                  0x000000011c61d936 mojo::SimpleWatcher::Context::Notify(unsigned int, MojoHandleSignalsState, unsigned int) + 518
```

shell/browser/file_select_helper.cc
```diff
@@ -0,0 +1,299 @@
...
+
+  // Will be released in one of the following
+  // OnFilesSelected - Callback flow completed with response
+  // RunFileChooserEnd - If user cancelled dialog or render_frame_host was
+  // destroyed WebContentsDestroyed
+  AddRefWithCheck();
+  DCHECK(HasAtLeastOneRef());
+
+  auto callback = base::BindOnce(&FileSelectHelper::OnOpenDialogDone,
+                                 base::Unretained(this));
```
nornagon, line 70: It's surprising to me that we'd use `base::Unretained` here, given the class is refcounted. It's not clear to me why this class needs to be refcounted, but either we should just past `this` here, or we should change the class to not be refcounted.
```suggestion
                                 this);
```

shell/browser/file_select_helper.cc
```diff
@@ -0,0 +1,299 @@
...
+void FileSelectHelper::ShowOpenDialog(
+    const file_dialog::DialogSettings& settings) {
+  v8::Isolate* isolate = electron::JavascriptEnvironment::GetIsolate();
+  v8::HandleScope scope(isolate);
+  gin_helper::Promise<gin_helper::Dictionary> promise(isolate);
+
+  // Will be released in one of the following
+  // OnFilesSelected - Callback flow completed with response
+  // RunFileChooserEnd - If user cancelled dialog or render_frame_host was
+  // destroyed WebContentsDestroyed
```
nornagon, line 65:
```suggestion
  // destroyed (and thus WebContentsDestroyed was called)
```

shell/browser/file_select_helper.cc
```diff
@@ -0,0 +1,299 @@
...
+  AddRefWithCheck();
+  DCHECK(HasAtLeastOneRef());
+
+  auto callback = base::BindOnce(&FileSelectHelper::OnSaveDialogDone,
+                                 base::Unretained(this));
+  ignore_result(promise.Then(std::move(callback)));
```
nornagon, line 91: It's very confusing that `file_dialog::Show{Open,Save}Dialog` call back to C++ code using a promise and a v8 dictionary, rather than a `base::OnceCallback` and plain C++ data.

Am I missing something about the reason the code is written that way? Can we refactor it to be more straightforward and avoid using v8 data structures to communicate between two C++ functions?

The relevant upstream code uses //ui/shell_dialogs, which exposes a reasonable interface for this that I think both the `dialog` API and this code could use.

shell/browser/file_select_helper.cc
deepak1556: Refcounting is really not needed for this class, for a better DCHECK I was trying to maintain a deterministic refcount. Will rework this.

shell/browser/file_select_helper.cc
deepak1556: Yeah this mostly historic trying to use the exposed interface from `file_dialog` class and currently it only has the JS interfaces used for dialog apis.

> Can we refactor it to be more straightforward and avoid using v8 data structures to communicate between two C++ functions?

Yup we can and that actually is better. Thanks for identifying this!

> The relevant upstream code uses //ui/shell_dialogs, which exposes a reasonable interface for this that I think both the dialog API and this code could use.

This seems good to try, but maybe we do it in a follow-up PR, it would just increase the scope of the current PR.

codebytere: Planning on doing this in a refactor following this pr :)

shell/browser/native_browser_view_mac.mm
```diff
@@ -129,8 +129,8 @@ - (void)mouseDragged:(NSEvent*)theEvent {
 // #ifdef DEBUG_DRAG_REGIONS
 // - (void)drawRect:(NSRect)aRect
 // {
-//     [[NSColor redColor] set];
-//     NSRectFill([self bounds]);
+//   [[NSColor redColor] set];
```
nornagon, line 132: this change seems unrelated

nornagon (approved):

shell/browser/native_browser_view_mac.mm
codebytere: dangit, yes 😛  debris from unrelated

shell/browser/web_dialog_helper.cc
```diff
@@ -305,18 +111,19 @@ WebDialogHelper::~WebDialogHelper() = default;
...
   file_dialog::DialogSettings settings;
   settings.force_detached = offscreen_;
   settings.filters = GetFileTypesFromAcceptType(params.accept_types);
   settings.parent_window = window_;
   settings.title = base::UTF16ToUTF8(params.title);
 
-  auto file_select_helper = base::MakeRefCounted<FileSelectHelper>(
-      render_frame_host, std::move(listener), params.mode);
+  auto* fsc =
+      new FileSelectHelper(render_frame_host, std::move(listener), params.mode);
```
deepak1556, line 122: where does `FileSelectHelper` get deleted since it no longer ref counted ?

deepak1556: Class should be explicitly deleted wherever `Release` was called previously https://github.com/electron/electron/pull/25030/commits/7c6f600b2424b1fd7d61b17a35394dc7f590728a

deepak1556 (approved):
LGTM

codebytere:
> document.visibilityState should be hidden when the window is initially shown but hidden before the page is loaded - should be hidden when the window is initially shown but hidden before the page is loaded

is unrelated.

codebytere:
PDF failure tests are unrelated.

PR #25048, opened 2020-08-19
chore: cherry-pick 88b1d19329df from chromium

StevenEBarbaro:
Don't crash when using 'revert' in var() fallback

CSS-wide keywords should not be allowed here in general, but they
currently are by Chrome and FF. (And WPT requires this behavior).

It would be easy to make revert-in-fallback actually behave as
'revert', but I don't want to ship this behavior since the spec doesn't
currently define how to handle this. So for now I'm just adding a unit
test that verifies that we don't crash.

Bug: 1105635, 1105782
Change-Id: Ia8c9100484c3c351f67aada850211a0ff6d2367f
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2300079
Commit-Queue: Anders Hartvoll Ruud <andruud@chromium.org>
Reviewed-by: Oriol Brufau <obrufau@igalia.com>
Cr-Commit-Position: refs/heads/master@{#788624}

Notes:

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -121,3 +121,4 @@ worker_feat_add_hook_to_notify_script_ready.patch
 reconnect_p2p_socket_dispatcher_if_network_service_dies.patch
 allow_focus_to_move_into_an_editable_combobox_s_listbox.patch
 cherry-pick-70579363ce7b.patch
+cherry-pick-88b1d19329df.patch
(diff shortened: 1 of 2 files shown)


codebytere:
@StevenEBarbaro doesn't look like this patch applied correctly - also missing release notes

nornagon:
@StevenEBarbaro if you're going to do that i'd recommend the wip tag, rather than changing the title. I don't think it's necessary to use the wip tag though.

We should automatically fill in the release notes when opening the PRs—I can make that change.

StevenEBarbaro:
@nornagon Happy to do whatever in regards to WIP. The scenario I'm trying to solve is raising transparency on the CI/test phase of the backport and hopefully saving others time on knowing to skip the review. I set it up last night with the intention of revisiting the status today. 

Regarding the release notes: Sounds good! 🍻

PR #25058, opened 2020-08-20
feat: expose currencyCode on Product object in IAP module

svirs:
#### Checklist

- [X] `npm test` passes
- [X] This is **NOT A BREAKING CHANGE**. Breaking changes may not be merged to master until 11-x-y is branched.

#### Release Notes

Notes: Added the `currencyCode` field that Apple's StoreKit in-app-purchasing library provides but has not been added to the Product object that `inAppPurchase.getProducts` returns.


svirs:
would love to have this backported to older version :)

svirs:
First PR to `electron`, exciting! Here's some reasoning for this (which I'm not sure if it belongs in this comment or up top):

Currency codes are often used in the front and backend for various reasons. Having that information readily available let's us categorize the expected income better and create estimations of revenue much more easily since we can take the currency that's there and get an idea of the value of the transaction. It is also often used by 3rd party billing services for the magic blackbox that is their products. On the front end, while we currently are able to get properly formatted string for the price, we can't really use the info that is there to 'continue the conversation' so to speak, here's an example:

```
const products = inAppPurchase.getProducts(['mysku']);
if (products.length > 0) {
    const productPrice = products[0].price;
    const currency = products[0].currencyCode;
    const locale = 'de-DE';
    const currencyLocalizer = new Intl.NumberFormat(locale, { style: 'currency', currency })
    const twoLocalizedBucks = currencyLocalizer.format(2)
    alert(`You can get a subscription for {productPrice}! Isn't savings incredible? Tip your barista {twoLocalizedBucks} the next time you get a latte!`);
}
```
If the currency code came back as 'EUR' we'd see "2,00 €", but if it came back as something else from Apple such as "XPF" we'd see "2 CFPF" instead.

Apple is the source of truth for the currency a user is shown, we can't rely on anything else if we want the way we show pricing to be consistent across not only the app but also the native checkout dialogs.
```

**Comment by codebytere** [comment]
@svirs looks like lint is failing:

```
shell/browser/mac/in_app_purchase_product.mm:153:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
Total errors found: 1
```
this lgtm otherwise tho!

svirs:
@codebytere fixed!

svirs:
Would it all right to label this with `target/9-x-y`?

MarshallOfSound:
@svirs That would require a minor bump and therefore needs to go through the release WG.  I can tag it for review though 😄

Issue #22829, opened 2020-03-25
Expose `spellcheck_platform::CheckSpelling` API


craftzdog:
Hi, thank you for creating such a great framework!
I have a feature request which I think I can contribute, but let me check if you are interested in supporting it.

### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description
The builtin spellchecker has been supported in https://github.com/electron/electron/pull/20692.
But it does not allow us to check spells programmatically, like [node-spellchecker](https://github.com/felixrieseberg/node-spellchecker)'s `isMisspelled` method.
Chrome's spellcheck_platform provides [CheckSpelling](https://chromium.googlesource.com/chromium/src/+/46ee392970e491fb9ee2523348e23222b40d5119/components/spellcheck/browser/spellcheck_platform.h#60) to do that exactly.

### Proposed Solution
Add a method to call `spellcheck_platform::CheckSpelling` like so:

```cc
#if BUILDFLAG(USE_BROWSER_SPELLCHECKER)
bool Session::CheckSpelling(const std::string& word) {
  if (spellcheck::UseBrowserSpellChecker()) {
    return spellcheck_platform::CheckSpelling(base::UTF8ToUTF16(word), 0);
  }
  return true;
}
#endif
```

in `shell/browser/api/atom_api_session.cc`.

It's simple and won't affect other features.

### Alternatives Considered
I don't know.

### Additional Information
None

Thanks!

MarshallOfSound:
@craftzdog That method is not cross platform, can you clarify your usecase for this method?

craftzdog:
@MarshallOfSound Thanks for the reply.

My usecase is that I'm building a Markdown editor based on [CodeMirror](https://codemirror.net/) which browser's spellchecker doesn't work.
So, we have to use a plugin like [codemirror-spell-checker](https://github.com/sparksuite/codemirror-spell-checker) which checks spellings in JS.

[...]

PR #25060, opened 2020-08-20
feat: Expose renderer spellcheck API

lishid:
#### Description of Change

Exposes renderer side browser spellcheck API to the javascript environment.
Resolves https://github.com/electron/electron/issues/22829

Rationale: Electron-based apps that uses CodeMirror (a hugely popular editor) are unable to benefit from the built-in spellchecker since CodeMirror does not use a ContentEditable (everything is faked). CodeMirror exposes a way to highlight errors, but given that we are currently unable to ask for a spellcheck from javascript, all Electron+CodeMirror apps use a js-based spellcheck library like https://github.com/cfinke/Typo.js which is considerably slower than the built-in native spellchecker.

This PR adds the following API:

- `isWordMisspelled(word: string): boolean`
- `getWordSuggestions(word: string): string[]`

**Current WIP adds the methods to the global `process` object which is just so I can test that it works.**
I've tested this and it works. Now to figure out where to put the API.

Ideally I would like this to be a module on `require('electron')`, similar to [IPCRenderer](https://github.com/electron/electron/blob/master/shell/renderer/api/electron_api_ipc_renderer.cc).
In practice, I have not found a way to pass the `WeakPtr<Spellcheck>` to [the renderer API stack](https://github.com/electron/electron/blob/master/lib/renderer/api/module-list.ts). I was also unable to find a reference back to `ElectronRendererClient`/`RenderClientBase` from an API module.
If someone with experience can give me some tips on how to achieve it that would be great!

Also currently missing tests and documentation.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added spellcheck API to renderer.

### DIFF
--- a/shell/renderer/api/electron_api_web_frame.cc
+++ b/shell/renderer/api/electron_api_web_frame.cc
@@ -10,6 +10,8 @@
 
 #include "base/command_line.h"
 #include "base/memory/memory_pressure_listener.h"
+#include "base/strings/utf_string_conversions.h"
+#include "components/spellcheck/renderer/spellcheck.h"
 #include "content/public/renderer/render_frame.h"
 #include "content/public/renderer/render_frame_observer.h"
 #include "content/public/renderer/render_frame_visitor.h"
@@ -24,6 +26,7 @@
 #include "shell/common/node_includes.h"
 #include "shell/common/options_switches.h"
 #include "shell/renderer/api/electron_api_spell_check_client.h"
+#include "shell/renderer/electron_renderer_client.h"
 #include "third_party/blink/public/common/page/page_zoom.h"
 #include "third_party/blink/public/common/web_cache/web_cache_resource_type_stats.h"
 #include "third_party/blink/public/platform/web_cache.h"
@@ -99,6 +102,24 @@ content::RenderFrame* GetRenderFrame(v8::Local<v8::Value> value) {
   if (!frame)
     return nullptr;
   return content::RenderFrame::FromWebFrame(frame);
+}
+
+bool SpellCheckWord(v8::Isolate* isolate,
+                    v8::Local<v8::Value> window,
+                    const std::string& word,
+                    std::vector<base::string16>* optional_suggestions) {
+  size_t start;
+  size_t length;
+
+  ElectronRendererClient* client = ElectronRendererClient::Get();
+  auto* render_frame = GetRenderFrame(window);
+  if (!render_frame)
+    return true;
+
+  base::string16 w = base::UTF8ToUTF16(word);
+  int id = render_frame->GetRoutingID();
+  return client->GetSpellCheck()->SpellCheckWord(
+      w.c_str(), 0, word.size(), id, &start, &length, optional_suggestions);
 }
 
 class RenderFrameStatus final : public content::RenderFrameObserver {
@@ -671,6 +692,20 @@ blink::WebCacheResourceTypeStats GetResourceUsage(v8::Isolate* isolate) {
   return stats;
 }
 
+bool IsWordMisspelled(v8::Isolate* isolate,
+                      v8::Local<v8::Value> window,
+                      const std::string& word) {
+  return !SpellCheckWord(isolate, window, word, nullptr);
+}
+
+std::vector<base::string16> GetWordSuggestions(v8::Isolate* isolate,
+                                               v8::Local<v8::Value> window,
+                                               const std::string& word) {
+  std::vector<base::string16> suggestions;
+  SpellCheckWord(isolate, window, word, &suggestions);
+  return suggestions;
+}
+
 void ClearCache(v8::Isolate* isolate) {
   isolate->IdleNotificationDeadline(0.5);
   blink::WebCache::Clear();
@@ -826,6 +861,10 @@ void Initialize(v8::Local<v8::Object> exports,
                  &ExecuteJavaScriptInIsolatedWorld);
   dict.SetMethod("setIsolatedWorldInfo", &SetIsolatedWorldInfo);
   dict.SetMethod("getResourceUsage", &GetResourceUsage);
+#if BUILDFLAG(ENABLE_BUILTIN_SPELLCHECKER)
+  dict.SetMethod("isWordMisspelled", &IsWordMisspelled);
+  dict.SetMethod("getWordSuggestions", &GetWordSuggestions);
+#endif
   dict.SetMethod("clearCache", &ClearCache);
   dict.SetMethod("_findFrameByRoutingId", &FindFrameByRoutingId);
   dict.SetMethod("_getFrameForSelector", &GetFrameForSelector);
--- a/shell/renderer/electron_renderer_client.cc
+++ b/shell/renderer/electron_renderer_client.cc
@@ -34,13 +34,25 @@ bool IsDevToolsExtension(content::RenderFrame* render_frame) {
 
 }  // namespace
 
+// static
+ElectronRendererClient* ElectronRendererClient::self_ = nullptr;
+
 ElectronRendererClient::ElectronRendererClient()
     : node_bindings_(
           NodeBindings::Create(NodeBindings::BrowserEnvironment::RENDERER)),
-      electron_bindings_(new ElectronBindings(node_bindings_->uv_loop())) {}
+      electron_bindings_(new ElectronBindings(node_bindings_->uv_loop())) {
+  DCHECK(!self_) << "Cannot have two ElectronRendererClient";
+  self_ = this;
+}
 
 ElectronRendererClient::~ElectronRendererClient() {
   asar::ClearArchives();
+}
+
+// static
+ElectronRendererClient* ElectronRendererClient::Get() {
+  DCHECK(self_);
+  return self_;
 }
 
 void ElectronRendererClient::RenderFrameCreated(
--- a/docs/api/web-frame.md
+++ b/docs/api/web-frame.md
@@ -257,6 +257,20 @@ renderer process.
 
 Returns `WebFrame` - that has the supplied `routingId`, `null` if not found.
 
+### `webFrame.isWordMisspelled(word)`
+
+* `word` String - The word to be spellchecked.
+
+Returns `Boolean` - True if the word is misspelled according to the built in
+spellchecker, false otherwise. If no dictionary is loaded, always return false.
+
+### `webFrame.getWordSuggestions(word)`
+
+* `word` String - The misspelled word.
+
+Returns `String[]` - A list of suggested words for a given word. If the word
+is spelled correctly, the result will be empty.
+
 ## Properties
 
 ### `webFrame.top` _Readonly_
--- a/shell/renderer/electron_renderer_client.h
+++ b/shell/renderer/electron_renderer_client.h
@@ -26,6 +26,8 @@ class ElectronRendererClient : public RendererClientBase {
   ElectronRendererClient();
   ~ElectronRendererClient() override;
 
+  static ElectronRendererClient* Get();
+
   // electron::RendererClientBase:
   void DidCreateScriptContext(v8::Handle<v8::Context> context,
                               content::RenderFrame* render_frame) override;
@@ -69,6 +71,8 @@ class ElectronRendererClient : public RendererClientBase {
   // its script context. Doing so in a web page without scripts would trigger
   // assertion, so we have to keep a book of injected web frames.
   std::set<content::RenderFrame*> injected_frames_;
+
+  static ElectronRendererClient* self_;
 
   DISALLOW_COPY_AND_ASSIGN(ElectronRendererClient);
 };
--- a/spec-main/spellchecker-spec.ts
+++ b/spec-main/spellchecker-spec.ts
@@ -13,7 +13,10 @@ ifdescribe(features.isBuiltinSpellCheckerEnabled())('spellchecker', () => {
 
   beforeEach(async () => {
     w = new BrowserWindow({
-      show: false
+      show: false,
+      webPreferences: {
+        nodeIntegration: true
+      }
     });
     await w.loadFile(path.resolve(__dirname, './fixtures/chromium/spellchecker.html'));
   });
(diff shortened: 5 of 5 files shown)


codebytere:
cc @MarshallOfSound

shell/renderer/electron_renderer_client.cc
```diff
@@ -141,6 +190,13 @@ void ElectronRendererClient::DidCreateScriptContext(
   gin_helper::Dictionary process_dict(env->isolate(), env->process_object());
   process_dict.SetReadOnly("isMainFrame", render_frame->IsMainFrame());
 
+#if BUILDFLAG(ENABLE_BUILTIN_SPELLCHECKER)
+  base::WeakPtr<SpellCheck> c = GetSpellCheck()->AsWeakPtr();
+
+  process_dict.SetMethod("isWordMisspelled", base::BindRepeating(&IsWordMisspelled, c));
```
MarshallOfSound, line 196: Hm, so this definitely shouldn't be on `process`.  My guess is it should probably be on something like `webFrame` 🤔

lishid: Yeah agreed. I'm debating whether to put it on `webFrame` or adding its own module, something like `require('electron').spellcheck`.

Here's what I wrote in the PR description:

> In practice, I have not found a way to pass the `WeakPtr<Spellcheck>` to [the renderer API stack](https://github.com/electron/electron/blob/master/lib/renderer/api/module-list.ts). I was also unable to find a reference back to `ElectronRendererClient`/`RenderClientBase` from an API module.
> If someone with experience can give me some tips on how to achieve it that would be great!

lishid:
@MarshallOfSound Quick ping on this. Can you please advise on how I can get a reference from:
[RendererClientBase](https://github.com/electron/electron/blob/master/shell/renderer/renderer_client_base.h#L84) to [the renderer API modules](https://github.com/electron/electron/blob/master/lib/renderer/api/module-list.ts) such as [IPCRenderer](https://github.com/electron/electron/blob/master/shell/renderer/api/electron_api_ipc_renderer.cc)?

This is blocking further progress on this PR. Thanks.

shell/renderer/electron_renderer_client.cc
zcbenz: @lishid You can find some examples on adding new modules under `shell/renderer/api/` folder. 

Adding the APIs to `webFrame` would be simplest and you only need to move the code to `shell/renderer/api/electron_api_web_frame.cc`.

lishid: @zcbenz Thank you for the help! In my working copy I was already able to move the code & API to `webFrame`, however, I was unable to find a reference to [the spellchecker object](https://github.com/electron/electron/blob/master/shell/renderer/renderer_client_base.h#L84) which is located on RefererClientBase. Any tips? It seems that all the samples I've seen only references `content::RenderFrame` gotten from a [v8::Context](https://github.com/electron/electron/blob/master/shell/renderer/api/electron_api_web_frame.cc#L93)

zcbenz: Since the `RendererClientBase` is guaranteed to have only one instance per process, I think you can add a `RendererClientBase::Get()` method to return the global instance, like what we do in `ElectronBrowserMainParts::Get`.

Chromium also does a similar thing for `ChromeExtensionsRendererClient`, so this pattern should be fine.

lishid: Great having some code I can reference makes this a lot easier. Will update the PR soon, thanks again!

lishid:
Did a quick test and it seems to work pretty well.
I'll start working on tests and documentation if this looks all good 🎉

lishid:
Also I can't seem to be able to view the lint results from CircleCI without "logging in with Github" and giving it read/write access to all my repos. The details link gives a 404 "You may have been logged out."

Curious if this is intentionally configured, and if there's an alternate way to check the result details.

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -672,6 +675,38 @@ blink::WebCacheResourceTypeStats GetResourceUsage(v8::Isolate* isolate) {
   return stats;
 }
 
+bool SpellCheckWord(v8::Isolate* isolate,
```
zcbenz, line 678: Can you move this to anonymous namespace? i.e. where the `GetRenderFrame` are defined.

zcbenz:
The lint error is:

```patch
--- a/shell/renderer/api/electron_api_web_frame.cc
+++ b/shell/renderer/api/electron_api_web_frame.cc
@@ -675,9 +675,9 @@
   return stats;
 }
 
-bool SpellCheckWord(v8::Isolate* isolate, 
+bool SpellCheckWord(v8::Isolate* isolate,
                     v8::Local<v8::Value> window,
-                    const std::string& word, 
+                    const std::string& word,
                     std::vector<base::string16>* optional_suggestions) {
   size_t start;
   size_t length;
```

It seems that you need to remove some trailing whitespaces.

I'm not sure why the lint CI result is not visible to you. /cc @jkleinsc

zcbenz:
Current implementation looks good to me, and the API will be discussed at next wg-api's meeting.

Can you also add some documentation for the APIs?

lishid:
Sounds good, will do.

jkleinsc:
@lishid @zcbenz I checked with CircleCI and in regards to viewing CircleCi results (eg lint results), you will need to login to CircleCI to see the details.  You can see the overall status on GitHub, but viewing the details requires logging into CircleCI.  You should be able to pick and choose what repos you give CircleCI access to.

lishid:
@jkleinsc I see. It seems like the only choice given was "All repos" or "Public repos only"

image

Choosing "Public repos only" asks for read-write access to all public repos:

image

IIRC the problem is with Github's OAuth not having granular control over specific repos, but I guess in the meantime it shouldn't be much trouble registering a separate account if access control is an issue.

nornagon (changes requested):
Please add a test!

lishid:
Test and docs have both been added. Excited to finally get native spellchecker support for our CodeMirror editor!
Let me know if this PR still needs any tweaking.

jkleinsc:
The API WG reviewed this on Sept 21, 2020.

lishid:
Quick follow-up: is there anything still blocking this PR?

nornagon (approved):

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -101,6 +104,24 @@ content::RenderFrame* GetRenderFrame(v8::Local<v8::Value> value) {
...
+
+  ElectronRendererClient* client = ElectronRendererClient::Get();
+  auto* render_frame = GetRenderFrame(window);
+  if (!render_frame)
+    return true;
+
+  base::string16 w = base::UTF8ToUTF16(word);
+  int id = render_frame->GetRoutingID();
+  return client->GetSpellCheck()->SpellCheckWord(
+      w.c_str(), 0, word.size(), id, &start, &length, optional_suggestions);
```
nornagon, line 122: Hang on, I think we can just specify `kNoTag` for the tag? Which would mean we wouldn't need to fetch the render frame, which makes me think maybe this API shouldn't be on `webFrame` at all?

nornagon (changes requested):

shell/renderer/api/electron_api_web_frame.cc
nornagon: hm... https://source.chromium.org/chromium/chromium/src/+/master:components/spellcheck/renderer/spellcheck_provider.cc;l=310;drc=c83ae1d48c762639e626be9be0c77a209fdea359 is a reference that uses the routing id. there are other places in the code that use `kNoTag`. cc @MarshallOfSound for further review

nornagon: It seems to be used in the macOS platform spellchecker impl: https://source.chromium.org/chromium/chromium/src/+/master:components/spellcheck/browser/spellcheck_platform_mac.mm;l=201;drc=da08b7363d0947b0e7d98c5282b4cca513f62100

nornagon (approved):
switching my review back to approved because i'm satisfied that the routing id is probably the right thing to be passing here.

shell/renderer/electron_renderer_client.cc
```diff
@@ -34,15 +34,27 @@ bool IsDevToolsExtension(content::RenderFrame* render_frame) {
...
     : node_bindings_(
           NodeBindings::Create(NodeBindings::BrowserEnvironment::RENDERER)),
-      electron_bindings_(new ElectronBindings(node_bindings_->uv_loop())) {}
+      electron_bindings_(new ElectronBindings(node_bindings_->uv_loop())) {
+  DCHECK(!self_) << "Cannot have two ElectronRendererClient";
+  self_ = this;
+}
 
 ElectronRendererClient::~ElectronRendererClient() {
   asar::ClearArchives();
```
miniak, line 49: `self_ = nullptr`

lishid: I could add this, but it didn't seem necessary given it wasn't added for ElectronBrowserMainParts https://github.com/electron/electron/blob/01dc50d8295690058fbde34b540e3b44b486bdbc/shell/browser/electron_browser_main_parts.cc#L205

lishid:
Quick follow-up: is there anything still blocking this PR? Seems like it's ready to merge.

jkleinsc (changes requested):
Looks like the newly added tests are failing on Linux

lishid:
> Looks like the newly added tests are failing on Linux

Good catch, seems like I forgot to add the `spellcheck` flag to the test runner.

jkleinsc:
The API WG meeting approved this API at our Oct 5, 2020 meeting.

lishid:
Anyone has an idea why the test script doesn't seems to have spellchecker properly enabled?

I've enabled them here:
https://github.com/electron/electron/commit/c160b8e35fb6278e7e921bcd4fdef609725c95e3#diff-391ce3c5c52c03cc68b42bd9c0fdc0deR115

I've also added a delay so the dictionary has time to load:
https://github.com/electron/electron/commit/c160b8e35fb6278e7e921bcd4fdef609725c95e3#diff-9260609f068ed314da5e8498971a6457R31

Neither seems to have loaded the spellchecker dictionary properly for the test suite...

lishid:
@MarshallOfSound Since you [wrote the original tests for spellchecker](https://github.com/electron/electron/commit/4323b6d618d41376c7fcef85752ebff48602651d) would it be possible to take a look at why the tests here aren't properly setup to enable spellchecker? Thanks in advance!
https://github.com/electron/electron/commit/c160b8e35fb6278e7e921bcd4fdef609725c95e3#diff-9260609f068ed314da5e8498971a6457

zcbenz:
@lishid Have you tested this API on Linux?

lishid:
> @lishid Have you tested this API on Linux?

Yes I have tested the API (Ubuntu 18.04) and it does work from the developer console when loading the app via `e start`.

lishid:
I'm on a slightly older nightly but I don't think that should have affected things:
image

lishid:
Just pulled the latest nightly and spent the whole day rebuilding - Seems like it still works correctly from the developer console.

zcbenz:
What if you run the Electron tests instead? Running code from devtools may behave differently from when running tests.

lishid:
I did run the test and as expected it does fail just like it did in CI.

I've made the test window show up and added a `<textarea>` with spellchecker turned on - it doesn't seem to work there either.

lishid:
I tried various functions and commands but the app seems to "think" spellchecker is enabled overall, the renderer's command line has the `--enable-spellcheck` flag, etc.

Running the functions in the console (while the test is running) doesn't seem to work. Weird given it works in `e start`. I'm running the tests using `e tests --runners=remote`.

zcbenz:
I think Chromium probably requires a textarea to present in _current_ page, and a context menu to be opened to actually initialize the spell checker.

So to make the test pass, you can probably dynamically add a textarea to the page and popup a context menu on it first.

lishid:
I don't think that's the case: I've added a `<textarea>` with a misspelled word, and I tried right clicking on it while the tests were running. I'd expect it should turn red with squiggly line but nothing happened. Subsequent calls to the spellchecker API also fails.

Also in my `e start` testing, which is literally just a clone of `electron-quick-start`, it doesn't have a `<textarea>` but the spellchecker API works right away.

lishid:
Would it be ok to move the tests to https://github.com/electron/electron/blob/master/spec-main/spellchecker-spec.ts instead? I can try if that works.

zcbenz:
I'm good with moving the tests to `spellchecker-spec.ts`.

lishid:
Apologies for the delay. New test is in and I have verified that it works on my Ubuntu 18.04 box. Hopefully the CI is green and we should be getting closer to a merge.

jkleinsc (approved):
LGTM... the mac arm64 CI failures are not related to this PR.  Rebasing this PR to the latest on master will resolve those failures.

lishid:
I did a rebase which seemed to have fixed the specific build issue with mac, but latest master has some other build failures :(

Anything else I need to do before a merge can happen? 🚀

zcbenz:
The CI is failing on master branch, we need to fix it first. https://github.com/electron/electron/pull/25992

zcbenz (approved):

craftzdog:
Congrats!!!!!

lishid:
@zcbenz Thanks for the merge!
I'm not too familiar with how the Electron team's roadmap planning works, but I'm curious to know which release version of Electron would contain these changes?
It doesn't seem like it's being backported to v11 so I'm guessing we're aiming for v12?

Thanks again for all the help on this PR!

zcbenz:
Yeah it will be included in v12, for backporting to v11 please check https://github.com/electron/governance/blob/master/wg-releases/feature-backport-requests.md.

lishid:
Sounds good.

> By requesting backport of a feature, the submitter of the PR is assuming responsibility for this feature on the requested branches.

I can take responsibility for this PR if that means we can backport it to v11. Given there's still over 3 weeks until v11 goes stable, let me know what I should do to initiate the backport request.

zcbenz:
I have added this PR to @electron/wg-releases 's meeting items, once it is approved you can start backporting it to 11-x.

lishid:
Seems like we got less than a week to go before v11 hits stable so I'm guessing a backport isn't likely to happen, is that correct?

jkleinsc:
@lishid the @electron/wg-releases missed this one, but we will vote on it.  It probably won't make 11.0.0, but if approved it could go in as a 11.1.0 release.

daviddev1:
What's the current state of this request, when is it going to be implemented into Electron?

PR #25065, opened 2020-08-20
refactor: wire will-navigate up to a navigation throttle instead of OpenURL

MarshallOfSound:
Was talking to some Chromium folks and apparently this is the more _correct_ way of doing this.  This also aligns our impl of `will-redirect` and `will-navigate` so that's good too.

Notes: no-notes


CezaryKulakowski:
@MarshallOfSound with this change there is no way for javascript to disallow navigation if it was initiated by browser process. Was it intended?

MarshallOfSound:
That was always how this event worked, and how it was documented.

Issue #25079, opened 2020-08-21
window.open doesn't work correctly in child window


CezaryKulakowski:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 11.0.0-nightly.20200814
* **Operating System:**
  * Windows 10
* **Last Known Working Electron version:**
  * unknown

### Expected Behavior
`window.open` called in child window opens new window with requested url

### Actual Behavior
If speculative render view host is deleted during naviagation to page we try to open with call to `window.open` window is destroyed right after it is created.

### To Reproduce
1. Create main window with web preference `nativeWindowOpen` set to `true`
1. Create child window with call to `window.open('about:blank')` in the main window's dev tools console
3. Create second child window with call to `window.open('http://www.google.com')` in the child's window dev tools console

Third window will be destroyed right after it is created.

### Additional Information
I attached an example application which reproduces the problem.
[window-open-app.zip](https://github.com/electron/electron/files/5107974/window-open-app.zip)

codebytere:
Closed by https://github.com/electron/electron/pull/25080

PR #25080, opened 2020-08-21
fix: window.open doesn't work correctly in child window

CezaryKulakowski:
#### Description of Change
If speculative render view host is deleted during naviagation to
page we try to open with call to window.open window is destroyed
right after it is created. It may happen when naviagation triggers
redirect from http to https. To void that we should emit event
current-render-view-deleted only when render view host which was
destroyed is the current webcontent's rvh.

#### Checklist

#### Release Notes

Notes: Fixed window.open called from child window

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -1065,22 +1065,13 @@ void WebContents::RenderFrameCreated(
     rwh_impl->disable_hidden_ = !background_throttling_;
 }
 
-void WebContents::RenderViewHostChanged(content::RenderViewHost* old_host,
-                                        content::RenderViewHost* new_host) {
-  currently_committed_process_id_ = new_host->GetProcess()->GetID();
-}
-
 void WebContents::RenderViewDeleted(content::RenderViewHost* render_view_host) {
   // This event is necessary for tracking any states with respect to
   // intermediate render view hosts aka speculative render view hosts. Currently
   // used by object-registry.js to ref count remote objects.
   Emit("render-view-deleted", render_view_host->GetProcess()->GetID());
 
-  if (-1 == currently_committed_process_id_ ||
-      render_view_host->GetProcess()->GetID() ==
-          currently_committed_process_id_) {
-    currently_committed_process_id_ = -1;
-
+  if (web_contents()->GetRenderViewHost() == render_view_host) {
     // When the RVH that has been deleted is the current RVH it means that the
     // the web contents are being closed. This is communicated by this event.
     // Currently tracked by guest-window-manager.js to destroy the
--- a/spec-main/api-web-contents-spec.ts
+++ b/spec-main/api-web-contents-spec.ts
@@ -1156,6 +1156,10 @@ describe('webContents module', () => {
             res.end();
           } else if (req.url === '/redirected') {
             res.end('<html><script>window.localStorage</script></html>');
+          } else if (req.url === '/first-window-open') {
+            res.end(`<html><script>window.open('${serverUrl}/second-window-open', 'first child');</script></html>`);
+          } else if (req.url === '/second-window-open') {
+            res.end('<html><script>window.open(\'wrong://url\', \'second child\');</script></html>');
           } else {
             res.end();
           }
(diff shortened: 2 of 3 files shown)


CezaryKulakowski:
@MarshallOfSound I've attached sample application to this ticket: https://github.com/electron/electron/issues/25079. I will try to create unit test for it.

CezaryKulakowski:
@MarshallOfSound I've pushed fixup with unit test. I couldn't use external page like www.google.com in the test but I noticed the the problem also occurs when we use url with invalid protocol in second call to window.open so I used it in the test. Unfortunately I used timeout as I didn't find a proper way to check if child window was destroyed without using it. If you have an idea how to achieve that please let me know.

dimww (approved):

spec-main/api-web-contents-spec.ts
```diff
@@ -1155,6 +1155,10 @@ describe('webContents module', () => {
             res.end();
           } else if (req.url === '/redirected') {
             res.end('<html><script>window.localStorage</script></html>');
+          } else if (req.url === '/first-window-open') {
+            res.end(`<html><script>window.open('${serverUrl}/second-window-open', 'first child');</script></html>`);
```
deepak1556, line 1159: Does this have to open another child window in `/second-window-open` to repro the issue, wouldn't it repro if you used `res.end(<html><script>window.open('${serverUrl}/redirect-cross-site', 'first child');</script></html>);` ?

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1003,22 +1003,13 @@ void WebContents::RenderFrameCreated(
...
   // intermediate render view hosts aka speculative render view hosts. Currently
   // used by object-registry.js to ref count remote objects.
   Emit("render-view-deleted", render_view_host->GetProcess()->GetID());
 
-  if (-1 == currently_committed_process_id_ ||
-      render_view_host->GetProcess()->GetID() ==
-          currently_committed_process_id_) {
-    currently_committed_process_id_ = -1;
-
+  if (web_contents()->GetRenderViewHost() == render_view_host) {
```
deepak1556, line 1012: Does this mean `RenderViewHostChanged` is now emitted for speculative hosts ?

spec-main/api-web-contents-spec.ts
CezaryKulakowski: I tried that and it didn't work (i.e. problem doesn't reproduce). It should reproduce if there would be redirection from http to https (first reproduction and the most probable in normal usage was on http://www.google.com which redirects to https://www.google.com) but to achieve it in the test it would be necessary to start https server with valid certificate next to http server.

shell/browser/api/electron_api_web_contents.cc
CezaryKulakowski: You mean `current-render-view-deleted`? If you do - quite the contrary. When we do a call to window.open in child window and there is a redirection from http to https speculative rvh is being destroyed - and that is expected. As there was no swap of rvh `currently_committed_process_id_` is set to `-1` so we generate `current-render-view-deleted` and destroy browser window. With this change it won't happen as rvh is assigned to web contents only when it is commited and it won't happen for speculative rvh.

spec-main/api-web-contents-spec.ts
```diff
@@ -1191,6 +1195,23 @@ describe('webContents module', () => {
       expect(currentRenderViewDeletedEmitted).to.be.false('current-render-view-deleted was emitted');
     });
 
+    it('does not destroy child window when wrong url and nativeWindowOpen set to true', async () => {
```
ppontes, line 1198: Wouldn't it be better to test directly the (not) emission of the relevant event ("current-render-view-deleted") instead of one consequence of that event? This is what the surrounding tests are doing.

ppontes (commented):
> @ppontes Can you take a look at this change? The related code was first added in #15821 and I'm not sure if this change would cause side effects.

Thanks for the notification, @zcbenz . I like this approach more as it's simpler and more explicit and can't recall why we didn't use it in the first place. I don't see side effects and the fact that the existing tests aren't broken hopefully confirms that.

spec-main/api-web-contents-spec.ts
deepak1556: the protocol doesn't matter here `http://www.google.com/` and `https://www.google.com/` are different origins for chromium, it is required for a cross origin redirect to repro this issue. But weird, maybe the testing needs a network delay. Anyways the current test is good, wanted to see if we can simplify it. Thanks!

shell/browser/api/electron_api_web_contents.cc
deepak1556: ah makes sense, thanks for the details!

ppontes (approved):

spec-main/api-web-contents-spec.ts
```diff
@@ -1191,6 +1195,27 @@ describe('webContents module', () => {
...
+      const destroyed = emittedOnce(parentWindow.webContents, 'destroyed');
+      const renderViewDeletedHandler = () => {
+        currentRenderViewDeletedEmitted = true;
+      };
+      app.once('browser-window-created', (event, window) => {
+        childWindow = window;
+        window.webContents.on('current-render-view-deleted' as any, renderViewDeletedHandler);
+      });
+      parentWindow.loadURL(`${serverUrl}/first-window-open`);
+      setTimeout(() => {
```
deepak1556, line 1211: Can we listen on navigation events on the childWindow before closing the parentWindow ? In this case it would be `did-fail-load` since the second url is a invalid url. I am bit worried about `setTimeout` being a potential for flaky tests.

CezaryKulakowski: I've checked that `did-fail-load` is not emitted for this url. I've tried to find another event which we could use but with no luck so far.

deepak1556 (approved):

CezaryKulakowski:
I've pushed branch rebased to the newest master (there was a conflict in one file so it was necessary) and with all fixups squashed. It should be ready to be merged to master.

CezaryKulakowski:
@MarshallOfSound test was created. Could you close the change request if it is sufficient for you?

MarshallOfSound (approved):

spec-main/api-web-contents-spec.ts
jkleinsc: @deepak1556 FWIW... it appears you are right as WOA is now failing on this test fairly regularly:
https://github.visualstudio.com/electron/_build/results?buildId=85266&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=5caf77c8-9b10-50ef-b5c7-ca89c63e1c86&l=2139

PR #25083, opened 2020-08-21
feat: make BrowserWindow vibrancy visual effect state customizable

codebytere:
#### Description of Change

Refs https://github.com/electron/electron/pull/24471#issuecomment-678366151.

Adds a new `visualEffectState` option to `BrowserWindow` options. We reasonably default to `NSVisualEffectStateFollowsWindowActiveState` since that is what is generally expected, but there are situations where an app developer may want to make the app visual effect state be inactive or active regardless of window activation state. We should thus allow them to do that.

cc @MarshallOfSound @jkleinsc @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Add `visualEffectState` option to BrowserWindows to allow customization of vibrancy effect state on macOS.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -212,6 +212,10 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
     Default is `false`. On Windows, does not work unless the window is frameless.
   * `type` String (optional) - The type of window, default is normal window. See more about
     this below.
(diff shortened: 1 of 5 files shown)


jkleinsc:
Merging as WOA failure is unrelated to PR change.

PR #25088, opened 2020-08-21
chore: cherry-pick 88b1d19329df from chromium

StevenEBarbaro:
Don't crash when using 'revert' in var() fallback

CSS-wide keywords should not be allowed here in general, but they
currently are by Chrome and FF. (And WPT requires this behavior).

It would be easy to make revert-in-fallback actually behave as
'revert', but I don't want to ship this behavior since the spec doesn't
currently define how to handle this. So for now I'm just adding a unit
test that verifies that we don't crash.

Bug: 1105635, 1105782
Change-Id: Ia8c9100484c3c351f67aada850211a0ff6d2367f
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2300079
Commit-Queue: Anders Hartvoll Ruud andruud@chromium.org
Reviewed-by: Oriol Brufau obrufau@igalia.com
Cr-Commit-Position: refs/heads/master@{#788624}

Notes: Security: Backported fix for CVE-2020-6539

--- depends on ---

Add parser support for the 'revert' keyword

This CL adds support for 'revert' in the parser, behind a flag. Using
the keyword doesn't do anything yet. It temporarily behaves like
"unset" for both regular declarations and keyframes.

Bug: 579788
Change-Id: I0882b41ff6f5c162be8ac1d1cd203ad4ec84c60f
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2130249
Commit-Queue: Anders Hartvoll Ruud <andruud@chromium.org>
Reviewed-by: Rune Lillesveen <futhark@chromium.org>
Cr-Commit-Position: refs/heads/master@{#756203}

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -121,3 +121,4 @@ worker_feat_add_hook_to_notify_script_ready.patch
 reconnect_p2p_socket_dispatcher_if_network_service_dies.patch
 allow_focus_to_move_into_an_editable_combobox_s_listbox.patch
 cherry-pick-70579363ce7b.patch
+cherry-pick-88b1d19329df.patch
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -122,3 +122,4 @@ reconnect_p2p_socket_dispatcher_if_network_service_dies.patch
 allow_focus_to_move_into_an_editable_combobox_s_listbox.patch
 cherry-pick-70579363ce7b.patch
 cherry-pick-88b1d19329df.patch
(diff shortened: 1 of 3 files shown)


StevenEBarbaro:
I've been considering the same; It's why I haven't bothered fixing the 7/8 branches. I've already tracked the related SHAs in the group's backport list so it won't be lost to time. I'm happy to close all 3 PRs

nornagon:
@StevenEBarbaro that seems reasonable to me.

PR #25089, opened 2020-08-21
feat: bring --enable-logging functionality in line with Chromium

ckerr:
#### Description of Change

This adds support for several features of Chromium's logging system that were previously not exposed in Electron. In particular:

- It's now possible to send Chromium logs to a file. This can be done either by passing `--log-file=.../path/to/file.log` along with `--enable-logging`. Passing `--enable-logging=file` without `--log-file` will send logs to a default path, `electron_debug.log` in the user-data directory.
    - This is particularly relevant on Windows, on which OS it is not possible to collect log messages from child processes when logging to stderr.
    - `ELECTRON_ENABLE_LOGGING=[1|file]` is equivalent to passing `--enable-logging` / `--enable-logging=file`
    - `ELECTRON_LOG_FILE=.../path/to/file.log` is equivalent to passing `--log-file=.../path/to/file.log`
- We now support `--log-level=n` to set the minimum level of log messages printed.
- It's now possible to enable logging from JavaScript, if done in the first tick. Logging switches added by `app.commandLine.appendSwitch()` will be read immediately after the first run of JS.

CC'ing @electron/wg-api 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added support for directing Chromium logging to a file with `--log-file=.../path/to/file.log`. Also, it's now possible to enable logging from JavaScript by appending command-line switches during the first JS tick.


jkleinsc:
The @electron/wg-api approved this at the August 24, 2020 meeting

codebytere:
@ckerr could you fix this up so we can merge?

codebytere:
> 534 tests failed out of 1386

i think something may have happened here haha

codebytere:
@ckerr do you plan to pursue this

nornagon:
Some notes on the latest update: the app was crashing on launch on macOS because child processes were attempting to open the log file, but they were forbidden from doing so by Seatbelt, which is (at least partially?) initialized before the child process launches. There is a specific hole carved in the seatbelt policy for log files: https://source.chromium.org/chromium/chromium/src/+/main:sandbox/policy/mac/common.sb;l=84-88;drc=4d4d32015c3683a0c29b0381d0c06372e53112b9, which is driven from [ContentBrowserClient::GetLoggingFileName](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/sandbox_parameters_mac.mm;l=113-114;drc=c268b7b5b3b1f0ef9545e35540dea0e57c944913).

So: we must provide a default logging file path, and we must know this path before child processes are launched, and the child process must use the same path. Otherwise the child will try to write somewhere else and will fail to do so because it is disallowed by Seatbelt.

By default, Chrome puts the log file into the user data directory. This is a little troublesome for Electron because our user data directory is only known in the main process, and it depends on the app name, which could e.g. come from package.json or by `app.setName()`. I've addressed this issue by passing `--user-data-dir` to child processes (+ correctly handling this flag in ElectronMainDelegate).

NB. This swaps out our custom `electron::DIR_USER_DATA` to use `chrome::DIR_USER_DATA` before. I'm not sure why it was the way it was though, cc @zcbenz @deepak1556.

Also, we deviate from Chromium in that `--enable-logging` in Chromium logs to a file by default, whereas to maintain compatibility, this PR preserves the behavior that `--enable-logging` in Electron logs to stderr. We introduce `--enable-logging=file` to log to the default log path.

nornagon:
Proposed behavior for various combinations of `--enable-logging`, `--log-file`, `ELECTRON_ENABLE_LOGGING` and `ELECTRON_LOG_FILE`:

`--enable-logging` ==> LOG_TO_STDERR
`--enable-logging=file` ==> LOG_TO_FILE with the default log path
`--enable-logging --log-file=…` ==> LOG_TO_FILE with the given log path
`--enable-logging=file --log-file=…` ==> LOG_TO_FILE with the given log path
`--enable-logging=stderr` ==> :boom:, not supported, print an error message
`ELECTRON_ENABLE_LOGGING=1` ==> LOG_TO_STDERR
`ELECTRON_ENABLE_LOGGING=file` ==> LOG_TO_FILE with the default log path
`ELECTRON_ENABLE_LOGGING=1 --log-file=...` ==> LOG_TO_FILE with the given log path
`--enable-logging ELECTRON_LOG_PATH=...` ==> LOG_TO_FILE with the given log path
`--enable-logging=file ELECTRON_LOG_PATH=...` ==> LOG_TO_FILE with the given log path

(not yet implemented)

deepak1556:
> This swaps out our custom electron::DIR_USER_DATA to use chrome::DIR_USER_DATA before. I'm not sure why it was the way it was though

It was originally added in https://github.com/electron-archive/brightray/pull/89 , I think there is no explicit reason except that we weren't building with `chrome_paths.{cc|h}` at that point.

QQ:

1) This change seems to depend on the order of the following calls https://github.com/electron/electron/blob/330b7849a3b6faf559f904a77dfe92b77c910277/shell/app/electron_main_delegate.cc#L224-L225 (i-e) when there is no override for a key, [base::PathService::Get](https://source.chromium.org/chromium/chromium/src/+/main:base/path_service.cc;l=209) looks for the first provider that returns a successful value for the key. Should those calls be rearranged ?

2) Is it okay manipulate a key that is outside the range of a provider ? (i-e) Currently `base::RegisterProvider` uses the key_start and key_end values only for a check on path range collision between different providers but if the implementation of `base::PathService` were to change in a way that only providers matching a particular key range will be invoked then we can't manipulate the `chrome::` key from our electron path provider.

nornagon:
> > This swaps out our custom electron::DIR_USER_DATA to use chrome::DIR_USER_DATA before. I'm not sure why it was the way it was though
> 
> It was originally added in [electron-archive/brightray#89](https://github.com/electron-archive/brightray/pull/89) , I think there is no explicit reason except that we weren't building with `chrome_paths.{cc|h}` at that point.
> 
> QQ:
> 
> 1. This change seems to depend on the order of the following calls https://github.com/electron/electron/blob/330b7849a3b6faf559f904a77dfe92b77c910277/shell/app/electron_main_delegate.cc#L224-L225
>     (i-e) when there is no override for a key, [base::PathService::Get](https://source.chromium.org/chromium/chromium/src/+/main:base/path_service.cc;l=209) looks for the first provider that returns a successful value for the key. Should those calls be rearranged ?

Yeah, they should be, good catch.

> 2. Is it okay manipulate a key that is outside the range of a provider ? (i-e) Currently `base::RegisterProvider` uses the key_start and key_end values only for a check on path range collision between different providers but if the implementation of `base::PathService` were to change in a way that only providers matching a particular key range will be invoked then we can't manipulate the `chrome::` key from our electron path provider.

I admit this is cheating a bit, but given PathService doesn't even compile the key_start / key_end checks when `NDEBUG` is defined (which it always is in Electron, even in debug builds), I'd say we're pretty safe here. PathService has been around a long, long time and hasn't changed much.

nornagon:
ah crap, i just realised we can't save logs to the user-data directory because we technically can't know it during `PreSandboxStartup`, since `PreSandboxStartup` is called before JS runs.

nornagon:
It looks like `InitLogging` is allowed to be called twice:

```
// This function may be called a second time to re-direct logging (e.g after
// loging in to a user partition), however it should never be called more than
// twice.
inline bool InitLogging(const LoggingSettings& settings) {
```
https://source.chromium.org/chromium/chromium/src/+/main:base/logging.h;l=274-277;drc=d29fc436d1245e58e476c6ab07cf6c144684fef0

so maybe we could do that?

nornagon:
Ok, the above is done now, but I'm yet to test interaction with `app.setAppLogsPath` or `app.setPath('logs', ...)` (why are there two different methods for that...?)

deepak1556:
`app.setPath('logs', ...)` should be a noop at the moment as we don't respect the change. You are right, they are redundant, maybe unify them for now and then deprecate one of them.

docs/api/command-line-switches.md
```diff
@@ -67,13 +67,27 @@ Enables caller stack logging for the following APIs (filtering events):
...
-This switch can not be used in `app.commandLine.appendSwitch` since it is parsed
-earlier than user's app is loaded, but you can set the `ELECTRON_ENABLE_LOGGING`
-environment variable to achieve the same effect.
+This switch can not be used in `app.commandLine.appendSwitch` since it is read
+before the first line JavaScript is run.
+
+The `ELECTRON_ENABLE_LOGGING` environment variable has the same effect as
+passing `--enable-logging`.
+
+If `destination` is:
```
MarshallOfSound, line 80: Change to `file` or rename the argument above?

docs/api/command-line-switches.md
```diff
@@ -67,13 +67,27 @@ Enables caller stack logging for the following APIs (filtering events):
...
+
+* an empty string: logs will be printed on stderr.
+* `file`: logs will be saved to the file specified by `--log-file=...`, or to
+  `electron_debug.log` in the user-data directory if `--log-file` is not
```
MarshallOfSound, line 84: Could we be nice here and do `{app_name}_debug.log` just so it's, idk, branded 😆

shell/app/electron_main_delegate.cc
```diff
@@ -281,6 +277,19 @@ void ElectronMainDelegate::PreSandboxStartup() {
   std::string process_type =
       command_line->GetSwitchValueASCII(::switches::kProcessType);
 
+  base::FilePath user_data_dir =
+      command_line->GetSwitchValuePath(::switches::kUserDataDir);
+  if (!user_data_dir.empty()) {
+    base::PathService::OverrideAndCreateIfNeeded(chrome::DIR_USER_DATA,
```
MarshallOfSound, line 283: How does this interact with things like https://github.com/electron/electron/blob/main/lib/browser/api/app.ts#L61-L72

shell/browser/api/electron_api_app.cc
```diff
@@ -998,13 +998,6 @@ void App::SetPath(gin_helper::ErrorThrower thrower,
   if (key >= 0) {
     succeed =
         base::PathService::OverrideAndCreateIfNeeded(key, path, true, false);
-    if (key == DIR_USER_DATA) {
-      succeed |= base::PathService::OverrideAndCreateIfNeeded(
-          chrome::DIR_USER_DATA, path, true, false);
-      succeed |= base::PathService::Override(
-          chrome::DIR_APP_DICTIONARIES,
```
MarshallOfSound, line 1005: I assume `DIR_APP_DICTIONARIES` now works correctly as it is resolved based on `DIR_USER_DATA`? Can you double check this?

docs/api/command-line-switches.md
nornagon: child processes can't get the app name, so we'd need to add another command-line argument, which I don't think is worth it.

shell/browser/api/electron_api_app.cc
nornagon: yeah, that's the idea. How do I check it?

shell/app/electron_main_delegate.cc
nornagon: This is here because this needs to be correct in child processes. I think we might be able to delete (some of) that app.ts code now.

nornagon:
> `app.setPath('logs', ...)` should be a noop at the moment as we don't respect the change. You are right, they are redundant, maybe unify them for now and then deprecate one of them.

It looks like none of these things have ever interacted with Chromium logging, so I'll leave this for a followup.

shell/app/electron_main_delegate.cc
nornagon: Turns out we can delete all of it I'm pretty sure 👍

nornagon:
> > > This swaps out our custom electron::DIR_USER_DATA to use chrome::DIR_USER_DATA before. I'm not sure why it was the way it was though
> > 
> > 
> > It was originally added in [electron-archive/brightray#89](https://github.com/electron-archive/brightray/pull/89) , I think there is no explicit reason except that we weren't building with `chrome_paths.{cc|h}` at that point.
> > QQ:
> > 
> > 1. This change seems to depend on the order of the following calls https://github.com/electron/electron/blob/330b7849a3b6faf559f904a77dfe92b77c910277/shell/app/electron_main_delegate.cc#L224-L225
> >    
> >    (i-e) when there is no override for a key, [base::PathService::Get](https://source.chromium.org/chromium/chromium/src/+/main:base/path_service.cc;l=209) looks for the first provider that returns a successful value for the key. Should those calls be rearranged ?
> 
> Yeah, they should be, good catch.

Nevermind, PathService checks the providers in reverse order: https://source.chromium.org/chromium/chromium/src/+/main:base/path_service.cc;l=343-344;drc=3f0b3892c73048740e9a98f174974fa1fafd78df

docs/api/command-line-switches.md
```diff
@@ -67,13 +67,25 @@ Enables caller stack logging for the following APIs (filtering events):
...
-### --enable-logging
+### --enable-logging[=file]
 
-Prints Chromium's logging into console.
+Prints Chromium's logging to stderr (or a log file).
 
-This switch can not be used in `app.commandLine.appendSwitch` since it is parsed
-earlier than user's app is loaded, but you can set the `ELECTRON_ENABLE_LOGGING`
-environment variable to achieve the same effect.
+This switch can not be used in `app.commandLine.appendSwitch` since it is read
```
deepak1556, line 74: Should be adjusted a little that we support for the first tick ?

shell/common/logging.cc
```diff
@@ -0,0 +1,132 @@
...
+#include "base/logging.h"
+#include "base/path_service.h"
+#include "base/strings/string_number_conversions.h"
+#include "chrome/common/chrome_paths.h"
+#include "content/public/common/content_switches.h"
+#include "shell/common/electron_paths.h"
+
+namespace logging {
+
+constexpr base::StringPiece kLogFileName("ELECTRON_LOG_FILE");
```
deepak1556, line 22: Do we want to document this new env variable ?

shell/browser/electron_browser_main_parts.cc
```diff
@@ -280,6 +266,11 @@ void ElectronBrowserMainParts::PostEarlyInitialization() {
   // Initialize field trials.
   InitializeFieldTrials();
 
+  // Reinitialize logging now that the app has had a chance to set the app name
+  // and/or user data directory.
+  logging::InitElectronLogging(*base::CommandLine::ForCurrentProcess(),
```
deepak1556, line 271: Should this be avoided for `--enable-logging --log-file` case as mentioned in your comment or is it a follow-up ?

docs/api/command-line-switches.md
nornagon: ah yep, I think I fixed this somewhere else but missed this one. 👍

shell/common/logging.cc
nornagon: It was mentioned in `--log-file` docs, but i've added it to the environment variables doc too.

jkleinsc (approved):

deepak1556 (approved):
LGTM

PR #25091, opened 2020-08-22
build: update to typescript 4

MarshallOfSound:
Lot's of fun stuff in typescript 4, this just updates us so we could in theory use it 😄 

Notes: no-notes


vladimiry:
Re https://github.com/electron/electron/commit/8baa9deccdbd2473c99811eaf2291c3d00fce0ff#diff-dd0e415ac6643adc75f041a4617e141bR168-R174 I think this would be a more type-safe way since you don't cast stuff to any:
```typescript
for (const prop of ["Buffer", "setImmediate", "clearImmediate", "global", "root", "GLOBAL"] as const) {
    delete (global as Partial<Pick<typeof global, typeof prop>>)[prop];
}
```

PR #25112, opened 2020-08-24
fix: prevent crash if BrowserView webContents was destroyed

codebytere:
#### Description of Change

Fixes an issue where interacting with a BrowserWindow which had a BrowserView attached whose `webContents` had been destroyed could cause a crash in `BaseWindow::ResetBrowserViews()`.

cc @nornagon @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none.

### DIFF
--- a/docs/api/browser-view.md
+++ b/docs/api/browser-view.md
@@ -40,16 +40,6 @@ A [`WebContents`](web-contents.md) object owned by this view.
 
 Objects created with `new BrowserView` have the following instance methods:
 
-#### `view.destroy()`
-
-Force closing the view, the `unload` and `beforeunload` events won't be emitted
-for the web page. After you're done with a view, call this function in order to
-free memory and other resources as soon as possible.
-
-#### `view.isDestroyed()`
-
-Returns `Boolean` - Whether the view is destroyed.
-
 #### `view.setAutoResize(options)` _Experimental_
 
 * `options` Object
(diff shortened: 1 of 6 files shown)


codebytere:
This might be more complex than originally anticipated bc `BrowserView::WebContentsDestroyed()` is destroyed async as of https://github.com/electron/electron/commit/746beb0d8b28809ec96d6ce9ad1ee252402584a7 so when we check for `browser_view->web_contents()` not being nullptr it might falsely still show as existing so this fix isn't deterministic

docs/api/browser-view.md
```diff
@@ -40,16 +40,6 @@ A [`WebContents`](web-contents.md) object owned by this view.
 
 Objects created with `new BrowserView` have the following instance methods:
 
-#### `view.destroy()`
```
codebytere, line 43: this does not exist

zcbenz (approved):

vladimiry:
@codebytere 

> This is NOT A BREAKING CHANGE. Breaking changes may not be merged to master until 11-x-y is branched.

The change looks like a breaking change for me since some public methods got removed.

How do I explicitly destroy the BrowserView instance in v11? Does the linked BrowserView isntance get automatically destroyed when I call `browserWindow.setBrowserView(null)`?

codebytere:
@vladimiry this PR didn't remove the methods - it followed up on another PR which did so. However, we do agree with you - the PR which did remove these methods did not align with above stated goals as you point out and the methods will be restored.

vladimiry:
> this PR didn't remove the methods - it just addressed documentation as a follow-up on another PR which did so

Thanks for clarification, I missed looking into that another PR.

Issue #25115, opened 2020-08-25
Electron 10.0.0 App Store Rejected - Private API - Framework/_CGFontRenderingGetFontSmoothingDisabled


bkervaski:
### Preflight Checklist

* [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.

### Issue Details

* **Electron Version:** 
  * 10.0.0

### Rejection Email
Guideline 2.5.1 - Performance - Software Requirements

Your app includes a version of an SDK from Electron that violates the App Store Review Guidelines. The version of the Electron SDK you are using in your app attempts to hide the use of private APIs. This is a violation Section 2.5.1 of the App Store Review Guidelines.

• Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework/_CGFontRenderingGetFontSmoothingDisabled

Guideline 2.5.1: Apps may only use public APIs.

PR #25117, opened 2020-08-25
fix: mitigate use of private macOS font API

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/25115.

Fixes an issue where apps could be rejected in MAS becuase of the presence of a new private API: `CGFontRenderingGetFontSmoothingDisabled()`. The feature was added to ensure that macOS LCD text behavior was respected, but as it causes rejections we should disable the feature on MAS builds.

cc @MarshallOfSound @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where some apps might be rejected from the Mac App Store owing to a private font-related API added in Chromium.

### DIFF
--- a/patches/chromium/mas_no_private_api.patch
+++ b/patches/chromium/mas_no_private_api.patch
@@ -420,6 +420,44 @@ index 28ca1646af0b0cce40d27baec71cbe65adc334fa..bae65c1f485bc02eb9ef2ebf7018af4a
  }
  
  }  // namespace
+diff --git a/content/renderer/theme_helper_mac.mm b/content/renderer/theme_helper_mac.mm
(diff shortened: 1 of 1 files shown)


zcbenz (approved):

codebytere:
Failure is 

```
chromium features PDF Viewer opens when loading a pdf resource in a iframe - opens when loading a pdf resource in a iframe
/Users/distiller/project/src/electron/spec-main/chromium-spec.ts
Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/distiller/project/src/electron/spec-main/chromium-spec.ts)
    at listOnTimeout (internal/timers.js:549:17)
    at processTimers (internal/timers.js:492:7)
```

which has previously been seen and is not related.

Issue #15370, opened 2018-10-24
shell.moveItemToTrash() does not properly return false if file path is in use


bpasero:
* Output of `node_modules/.bin/electron --version`: 2.0.12
* Operating System (Platform and Version): Windows 10

**Expected Behavior**
`moveItemToTrash` should return `false` if the file is locked.

**Actual behavior**
`moveItemToTrash` does not return `false` when a file is locked.

**To Reproduce**
* open a folder in a command prompt (cd into it)
* run `shell.moveItemToTrash(path of folder)`
* observe the returned value is `true` indicating this operation was successful

sofianguy:
@bpasero is this bug still present in version 4.x or later?

bpasero:
@sofianguy reproduces with Electron 6.0.1

niik:
@codebytere @nornagon This was fixed in #25124 but I forgot to use the proper syntax for auto-closing it on merge. Would either of you be so kind and close it?

bpasero:
Thanks a lot for working on this 👍

PR #3337, opened 2015-11-04
Fixing MoveItemToTrash behavior in Windows

tylergibson:
Fixes #2715 by implementing IFileOperation (used in Windows Vista forward), replacing SHFileOperation.  This allows us to do a filesystem PreDeleteItem check to ensure the file can actually go into the Recycle Bin.  The existing behavior actually deletes files that cannot be recycled.

Before moving to IFileOperation, I attempted to extend the current implementation by adding the three  checks (Type of Drive, Existence of Recycle Bin on that drive, and free space available in the Bin) with: 

    INT drive_type = GetDriveType((LPCWSTR)(path_components.at(0) + L"\\").c_str());
    QueryDosDevice((LPCWSTR)path_components.at(0).c_str(), dos_path_chars, 100);
    rbinfo.cbSize = sizeof(SHQUERYRBINFO);
    SHQueryRecycleBin((LPCWSTR)path.value().c_str(), &rbinfo);

But these operations can be extremely slow, especially the SHQueryRecycleBin, which can take seconds to execute, locking the UI in the mean time.  And there are still corner cases where this can fail.

So I move to reimplement with IFileOperation.  It needs a full implementation of the interface IFileOperationProgressSink to be able to check Pre-Delete for the Recycle Bin.  I authored the class implementation inline in platform_util_win.cc and only implemented the minimum methods required for MoveItemToTrash.

It may make sense to actually refactor the IFileOperationProgressSink class externally and use it more generally for other special case file I/O issues on Windows.  SHFileOperation is still faster in execution, but has many limitations in use.

Behavior is now correct.  Files that cannot be recycled will not be deleted and the function will return false.

PR #25113, opened 2020-08-24
test: add tests for shell.moveItemToTrash

nornagon:
#### Description of Change
We didn't have tests for this before. now we do!

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).
- [x] This is **NOT A BREAKING CHANGE**. Breaking changes may not be merged to master until 11-x-y is branched.

#### Release Notes

Notes: none

PR #25124, opened 2020-08-25
fix: make shell.moveItemToTrash return false on Windows when move is unsuccessful

niik:
#### Description of Change

This is a shot in the dark attempt at fixing #15370. I haven't been able to build electron locally and I don't even know if this will compile let alone solve the issue. I'm hoping to have electron's CI do the hard work for me so that I can then test using the CI artifacts.

Minimal test case: https://github.com/niik/electron-win-moveitemtotrash-test-case

From: https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ifileoperation-getanyoperationsaborted

> You should call IFileOperation::GetAnyOperationsAborted regardless of whether IFileOperation::PerformOperations returned a success or failure code. A success code can be returned even if the operation was stopped by the user or the system.

The [original implementation](https://github.com/electron/electron/commit/bedea89dcf3642e512934567e99ee4e65bdea3c5) had something very similar to this:

https://github.com/electron/electron/blob/bedea89dcf3642e512934567e99ee4e65bdea3c5/atom/common/platform_util_win.cc#L181-L186

Which I believe was unintentionally removed in https://github.com/electron/electron/commit/4ff5ce4d6d4b8104c1dfca2b95f15ab847a2a3ed (https://github.com/electron/electron/pull/3337).

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed shell.moveItemToTrash on Windows so that it returns false when move was unsuccessful.


niik:
The Windows build failed but to my untrained eye it looks like a flake. Even though it failed I was able to grab the compiled dist from the build artifacts and run the [reproduction case] on it.

### Before
Screen Shot 2020-08-25 at 23 09 18

### After

Screen Shot 2020-08-25 at 23 09 00

> Would it be possible to add a test?

Yup, shouldn't be too hard. Probably creating a file and acquiring an exclusive read lock on it before trying to move it to trash. It probably needs to be a Windows-specific test though as the other platforms don't lock files like Windows does. I'll check back in tomorrow and pull in your PR and see about that test.

niik:
> * PR description included and stakeholders cc'd

cc @deepak1556 since you wrote the initial implementation for returning false when delete failed (https://github.com/electron/electron/commit/bedea89dcf3642e512934567e99ee4e65bdea3c5)
cc @tylergibson since you wrote the more comprehensive delete sink in https://github.com/electron/electron/pull/3337 which  (inadvertently?) omitted the AnyOperationsAborted check

niik:
I've merged #25113 into this branch in order to have a base to start from when writing tests. I'll move this PR back to a draft until #25113 has been merged and these tests have been finalized.

nornagon:
FYI, #25113 has now been merged :)

niik:
@nornagon Thanks so much! I've merged master and I'll open this up for review again. Assuming this gets merged what's the process for getting it backported? We're hitting this in GitHub Desktop and we're currently on Electron 9 and I expect it'll take us a fair while to get onto 11.

codebytere:
@niik i tagged it!

niik:
> @niik i tagged it - it'll go out when the backports get merged! Looks like they'll need manual though if you're willing to pop them up targeting each of those branches (they didn't apply cleanly)

@codebytere I'll give it a shot!

tylergibson:
LGTM, yes completely inadvertent that I missed this check in my PR

PR #25126, opened 2020-08-25
feat: reinvigorate visibleOnFullscreen option

codebytere:
Backport of https://github.com/electron/electron/pull/24956.

See that PR for more details.

Notes: Added back a previously broken visibleOnFullScreen option for setVisibleOnAllWorkspaces.


codebytere:
> chrome extensions devtools extensions loads a devtools extension

is unrelated.

PR #25128, opened 2020-08-25
feat: add support for will-navigate and will-fail-load custom error pages

MarshallOfSound:
### Use Cases

Closes https://github.com/electron/electron/issues/30369.

Sometimes users navigation failures occur, mostly due to network issues (proxy, DNS failure, etc.) currently those failures send users to a blank page that is very hard to customize or show any meaningful information on.

This new event and new capability to provide custom HTML error pages allows app devs to provide actually _useful_ error pages in the same style as their app for a much nicer "app" experience vs the "white screen of doom" that people associate with web apps failing to load.

### Example Usage

```js
mainWindow.webContents.on('will-fail-request', (e, url, isInPlace, isMainFrame, frameProcessId, frameRoutingId, errorCode, errorDescription) => {
  e.returnValue = {
    errorCode: -2,
    errorPage: `
<html>
<body>
  <h1>Oh no, something went wrong</h1>
  <p>URL: ${url}</p>
  <p>Error: ${errorCode} - ${errorDescription}</p>
</body>
</html>`
  };
});
```

### Screenshot
Screen Shot 2020-08-25 at 1 53 53 PM

Notes:
* Added new `will-fail-load` event that is emitted when a navigation is about to fail for a known reason
* Added capability to `will-fail-load`, `will-navigate` and `will-redirect` events to provide a custom error page as raw HTML.  This allows end users to display custom UI when preventing navigations, when network failures occur or other issues cause navigation failures

### DIFF
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -73,8 +73,8 @@ Returns:
 This event will be emitted after `did-start-loading` and always before the
 `did-fail-load` event for the same navigation.
 
-Settings `event.returnValue` to an HTML string will result in a custom error page being
-displayed using that HTML.
+Settings `event.returnValue` to the appropriate object will result in a custom error page being
+displayed using custom HTML.
 
 #### Event: 'did-fail-load'
 
@@ -262,8 +262,8 @@ this purpose.
 
 Calling `event.preventDefault()` will prevent the navigation.
 
-Settings `event.returnValue` to an HTML string will result in a custom error page being
-displayed using that HTML and the navigation being cancelled.
+Settings `event.returnValue` to the appropriate object will result in a custom error page being
+displayed using custom HTML and the navigation being cancelled.
 
 #### Event: 'did-start-navigation'
 
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -61,7 +61,7 @@ spinning, and the `onload` event was dispatched.
 
 Returns:
 
-* `event` Event
+* `event` [CancellableNavigationEvent](structures/cancellable-navigation-event.md)
 * `url` String
 * `isInPlace` Boolean
 * `isMainFrame` Boolean
@@ -247,7 +247,7 @@ See [`window.open()`](window-open.md) for more details and how to use this in co
 
 Returns:
 
-* `event` Event
+* `event` [CancellableNavigationEvent](structures/cancellable-navigation-event.md)
 * `url` String
 
 Emitted when a user or the page wants to start navigation. It can happen when
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -57,6 +57,25 @@ Process: [Main](../glossary.md#main-process)
 Emitted when the navigation is done, i.e. the spinner of the tab has stopped
 spinning, and the `onload` event was dispatched.
 
+#### Event: 'will-fail-load'
+
+Returns:
+
+* `event` Event
+* `url` String
+* `isInPlace` Boolean
+* `isMainFrame` Boolean
+* `frameProcessId` Integer
+* `frameRoutingId` Integer
+* `errorCode` Integer
+* `errorDescription` String
+
+This event will be emitted after `did-start-loading` and always before the
+`did-fail-load` event for the same navigation.
+
+Settings `event.returnValue` to an HTML string will result in a custom error page being
+displayed using that HTML.
+
 #### Event: 'did-fail-load'
 
 Returns:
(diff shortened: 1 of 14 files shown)


shell/browser/electron_navigation_throttle.cc
```diff
@@ -37,10 +59,35 @@ ElectronNavigationThrottle::WillStartRequest() {
     return PROCEED;
   }
 
-  if (handle->IsRendererInitiated() && handle->IsInMainFrame() &&
-      api_contents->EmitNavigationEvent("will-navigate", handle)) {
+  std::shared_ptr<bool> is_delay_called(new bool(false));
```
zcbenz, line 62: `std::shared_ptr` is banned in Chromium: https://chromium-cpp.appspot.com. We do have a few code breaking this rule but we should not add more usages.

For this check, I think implementing it in JS would be much more easy to understand, and also more efficient.

```js
      let emitted = false
      Object.defineProperty(event, 'returnValue', {
         set: (value) => {
           if (emitted) return
         },
         get: () => {}
       });
       this.emit(eventName, event, ...args);
      emitted = true
```

shell/browser/electron_navigation_throttle.cc
```diff
@@ -37,10 +59,35 @@ ElectronNavigationThrottle::WillStartRequest() {
...
+  std::shared_ptr<std::string> error_html(new std::string);
```
zcbenz, line 63: It would be safe to pass a raw pointer since there is already code checking delayed call.

```c++
std::string error_html;
base::BindOnce(&HandleEventReturnValue, &error_html);
```

docs/api/web-contents.md
```diff
@@ -57,6 +57,25 @@ Process: [Main](../glossary.md#main-process)
...
+* `isMainFrame` Boolean
+* `frameProcessId` Integer
+* `frameRoutingId` Integer
+* `errorCode` Integer
+* `errorDescription` String
+
+This event will be emitted after `did-start-loading` and always before the
+`did-fail-load` event for the same navigation.
+
+Settings `event.returnValue` to an HTML string will result in a custom error page being
```
zcbenz, line 76: Since this API is essentially a wrapper of `NavigationThrottle`, I wonder if it would be more future proof by returning an Object instead:

```js
{ action: 'cancel', error: errorCode, errorPage: 'html' }
```

shell/browser/electron_navigation_throttle.cc
```diff
@@ -37,10 +59,35 @@ ElectronNavigationThrottle::WillStartRequest() {
...
+ElectronNavigationThrottle::WillFailRequest() {
+  return DelegateEventToWebContents("-will-fail-load",
+                                    navigation_handle()->GetNetErrorCode());
+}
+
+content::NavigationThrottle::ThrottleCheckResult
+ElectronNavigationThrottle::WillStartRequest() {
+  auto* handle = navigation_handle();
+  if (handle->IsRendererInitiated() && handle->IsInMainFrame())
+    return DelegateEventToWebContents("-will-navigate", net::ERR_FAILED);
```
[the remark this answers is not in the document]
MarshallOfSound: How would this emit `will-navigate` multiple times for the same request?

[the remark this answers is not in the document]
zcbenz: Sorry I forgot that the implementation of `will-navigate` had been changed.

docs/api/web-contents.md
```diff
@@ -57,6 +57,25 @@ Process: [Main](../glossary.md#main-process)
...
```
zcbenz, line 76: The documentation also needs to be updated as `returnValue` is now an object.

zcbenz (approved):

jkleinsc:
The API WG approved this at the Sept 21, 2020 meeting with the condition that this be added as an experimental feature.

zcbenz:
@MarshallOfSound Can you fix the docs and rebase the PR? Other than that I think this PR is ready for merge.

pushkin-:
Will this allow us to put custom UI into iframes as well as entire windows, or is that not possible?

jkleinsc (changes requested):
@MarshallOfSound can you make this an experimental feature as we discussed in the API WG?  Also, can we add test(s)?

MarshallOfSound:
[Add support for will-navigate and will-fail-load custom error pages](https://trello.com/c/w7Gxy3IM/12-add-support-for-will-navigate-and-will-fail-load-custom-error-pages)

gdavidkov:
Will be possible to return the error page for the iframes?

pushkin-:
Is there a reason to not also support supplying an absolute path to an html page rather than just using raw html?

samuelmaddock:
Marking as `wip` until @MarshallOfSound (or another contributor) has interest in continuing this work.

samuelmaddock:
I'm closing this out for now. Feel free to reopen, of course. However, I think this would be great to see if someone has the capacity to update this to work with `WebFrameMain`.

PR #25139, opened 2020-08-26
fix: do not reset process_id in URLLoaderFactoryParams

### LINKED ISSUE #24991: Two BrowserView loadUrl (or loadFile) without  await will cause upload file error in renderer process
OS: `windows 10`
Electron: `9.2.0` `9.1.2` `9.0.0` `8.5.0` `7.3.0`

Main process code:
```javascript
  width: 800,
  height: 600,
  webPreferences: {
    nodeIntegration: true,
    webSecurity:false
  }
})
let view1 = new BrowserView({
    webPreferences: {
      nodeIntegration: true,
      devTools: true,
      webSecurity: false,
    },
  });
mainWindow.setBrowserView(view1);
let size = mainWindow.getSize();
view1.setBounds({
  x: 0,
  y: 0,
  width: size[0],
  height: size[1],
});
view1.setAutoResize({
  width: true,
  height: true,
});
view1.webContents.loadFile(`index.html`)
view1.webContents.openDevTools({ mode: "undocked" })
let view2 = new BrowserView({
  webPreferences: {
    nodeIntegration: true,
    devTools: true,
    webSecurity: false,
  },
});
view2.webContents.loadFile(`index.html`);
```

Renderer process code
```javascript
//HTML
// <input id="fileInput" type="file" />
postFile = (data)=> {
  let _url = 'https://******.com/upload'
  return fetch(_url, {
    method: 'POST',
    body: data,
  }).then(response => response.json())
    .catch(error => {
      console.error(error)
    })
}
let fileInput = document.getElementById("fileInput")
fileInput.addEventListener('change', function() {
  var file = fileInput.files[0];
  let formData = new FormData()
  formData.append('filetype', 'image')
  formData.append('file', file)
  postFile(formData);
});
```
Error:

> POST https://******.com/upload net::ERR_ACCESS_DENIED

Description:
When add `await` before `view1.webContents.loadFile`,the renderer process will work correctly.

zcbenz:
#### Description of Change

Fix #24991.

In `OverrideURLLoaderFactoryParams` when disabling CORB, the `process_id` parameter was also changed.

This would cause a DCHECK assertion under certain cases (see https://github.com/electron/electron/issues/24991#issuecomment-680413218), and under Release build while the DHCECK assertion would be disabled, it would cause weird permission errors due to wrong process id being set (see https://github.com/electron/electron/issues/24991#issuecomment-677724346).

This behavior was introduced by https://github.com/electron/electron/pull/19488/commits/6d3e6abf9f43dab8773dd27c82333554ed55fca4 from https://github.com/electron/electron/pull/19488, according to its description, the purpose was to disable CORB when web security is disabled. The original change did need to set the `process_id` because it was constructing a new Param struct, but after later Chromium upgrades there was no longer need to set `process_id` because the method was changed to override the parameters instead of creating a new one.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix network permission error when there are multiple WebContents sharing same session are created with web security disabled.

### DIFF
--- a/shell/browser/electron_browser_client.cc
+++ b/shell/browser/electron_browser_client.cc
@@ -1506,15 +1506,10 @@ void ElectronBrowserClient::OverrideURLLoaderFactoryParams(
     const url::Origin& origin,
     bool is_for_isolated_world,
     network::mojom::URLLoaderFactoryParams* factory_params) {
-  for (const auto& iter : process_preferences_) {
-    if (iter.second.browser_context != browser_context)
-      continue;
-
-    if (!iter.second.web_security) {
-      // bypass CORB
-      factory_params->process_id = iter.first;
-      factory_params->is_corb_enabled = false;
-    }
+  // Bypass CORB when web security is disabled.
+  auto it = process_preferences_.find(factory_params->process_id);
+  if (it != process_preferences_.end() && !it->second.web_security) {
+    factory_params->is_corb_enabled = false;
   }
 
   extensions::URLLoaderFactoryManager::OverrideURLLoaderFactoryParams(
--- a/spec-main/chromium-spec.ts
+++ b/spec-main/chromium-spec.ts
@@ -245,6 +245,14 @@ describe('web security', () => {
       <script src="${serverUrl}"></script>`);
     await p;
   });
+
+  it('does not crash when multiple WebContent are created with web security disabled', () => {
+    const options = { webPreferences: { webSecurity: false } };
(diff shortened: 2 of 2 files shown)


deepak1556:
> Fix network permission error when there are multiple WebContents created with web security disabled.

maybe reword to highlight same session since the comparison was done based on browser_context.

`Fix network permission error when there are multiple WebContents sharing same session are created with web security disabled.`

PR #25146, opened 2020-08-26
fix: prevent crash if BrowserView webContents was destroyed

codebytere:
Backport of https://github.com/electron/electron/pull/25112.

See that PR for more details.

Notes: none


codebytere:
@jkleinsc yeah 🤔 the big difference between 10/11 is this: https://github.com/electron/electron/pull/23578

codebytere:
@nornagon might have thoughts on this one - we removed `getAllViews()` in the above linked PR (for good reason) but that's likely intersecting with this in some weird way i'm not seeing

PR #25151, opened 2020-08-27
fix: allow extensions to access local files

ChALkeR:
#### Description of Change

Allows file access for _all extensions_ (except probably the pdf viewer which is created separately).

This is how it was pre-9, until 9.0.0 broke react devtools on local `file://` urls (#24011). 

I'm not sure if this is a good solution.
An alternative would be to add an explicit opt-in to `loadExtension()`, but that would require API changes.

I would actually prefer an explicit opt-in over this, but I didn't heard a feedback on https://github.com/electron/electron/issues/24011#issuecomment-674316115, and this is a far easier change, so I started with this first. Hence a draft until there is a resolution on which approach should be followed.

Fixes: #24011.

Confirmed that react-devtools now works over `file://` with this patch.

cc @nornagon @samuelmaddock @RangerMauve

#### Checklist
- ~`npm test` passes~ can't test locally, tests are hardcoded to expect CI env, see https://github.com/electron/electron/pull/25151#issuecomment-681536581
- ~tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)~

#### Release Notes

Notes: Fixed extensions access to `file://` protocol to unbreak react-devtools.

### DIFF
--- a/shell/browser/extensions/electron_extension_loader.cc
+++ b/shell/browser/extensions/electron_extension_loader.cc
@@ -37,7 +37,8 @@ std::pair<scoped_refptr<const Extension>, std::string> LoadUnpacked(
     return std::make_pair(nullptr, err);
   }
 
-  int load_flags = Extension::FOLLOW_SYMLINKS_ANYWHERE;
+  int load_flags =
+      Extension::FOLLOW_SYMLINKS_ANYWHERE | Extension::ALLOW_FILE_ACCESS;
   std::string load_error;
   scoped_refptr<Extension> extension = file_util::LoadExtension(
       extension_dir, Manifest::COMMAND_LINE, load_flags, &load_error);


ChALkeR:
Local tests fail at:
```
ok 7 app module app.getLocale() should not be empty
not ok 8 app module app.getLocaleCountryCode() should be empty or have length of two
  AssertionError: expected 'RU' to have a length of 0 but got 2
      at Context.<anonymous> (electron/spec-main/api-app-spec.ts:130:69)
      at processImmediate (internal/timers.js:456:21)
```
which seeems unrelated?

sentialx:
@ChALkeR Would be really nice to have it as an option

nornagon:
It might be a good idea to land this first since https://github.com/electron/electron/pull/25198 is still pending some changes and this fix is quite simple. @sentialx @samuelmaddock how would you feel about landing this?

nornagon:
Also, this should get some tests.

nornagon:
Hm, on further reflection I think #25198 can be landed without the fix of remembering the file access across reloads. We can address that in a followup. I think #25198 is a safer approach, so I'm closing this in favor of that.

PR #25171, opened 2020-08-27
fix: make shell.moveItemToTrash return false on Windows when move is unsuccessful

niik:
#### Description of Change

Backport of #25124

#### Release Notes

Notes: Fixed shell.moveItemToTrash on Windows so that it returns false when move was unsuccessful.


codebytere:
Failure:
> chrome extensions devtools extensions loads a devtools extension

is unrelated.

PR #25189, opened 2020-08-28
Add context to ipcRenderer docs

Raynos:
It was not clear to me why exposing `invoke()` was bad but exposes a function that does `invoke('api', commandName)` was good.

My app has many IPC commands and writing explicit messages that whitelist each one is tedious and boilerplatey. It took me a while to understand that electron also uses `ipcRenderer` internally and that it's that which needs protection for security reasons.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:

### DIFF
--- a/docs/tutorial/context-isolation.md
+++ b/docs/tutorial/context-isolation.md
@@ -68,3 +68,6 @@ contextBridge.exposeInMainWorld('myAPI', {
 })
 ```
 
+The reason arbitary IPC messages are insecure is because electron uses `ipcRenderer` internally for communication. You only
+want to expose the IPC messages important for your app, not the electron internals.
+


Raynos:
If we validate in `ipcMain` is there a benefit to duplicating the validation logic in the `preload.js` script ?

Raynos:
> is that allows that webpage to call any of your IPC handlers with any arguments completely unsanitized.

Ah, so the electron documentation best practices assumes your electron app has multiple `BrowserWindow` with their own `preload.js` that each expose a subset of the apis available on `ipcMain` ?

Aka there's one `main` process with one `ipcMain` and there are N `BrowserWindow` with N `ipcRenderer` .

PR #25198, opened 2020-08-29
feat: implement allowFileAccess loadExtension option

ChALkeR:
#### Description of Change

Alternative to #25151.

Allows file access for extensions via an opt-in on per-extension basis.
Pre-9 this was essentially "enabled" by default, until 9.0.0 broke react devtools on local `file://` urls (#24011). 

* API / doc is done.
* Implementation works but is WiP (see below)
* I don't like how flags are being passed this way, also a FIXME in `LoadExtensionForReload()` -- not sure yet when that is triggered. Probably replacing that with an additional method that sets the flags for a certain extension id and is stored in-memory would be a better approach.

Fixes: #24011.

Confirmed that react-devtools now works over `file://` if `allowFileAccess: true` is passed and doesn't work otherwise.

cc @nornagon @samuelmaddock @RangerMauve

Thoughts? It would be helpful to hear which approach should be continued, this one or #25151.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added `allowFileAccess` option to `loadExtension()` API to unbreak react-devtools.


shell/browser/api/electron_api_session.cc
```diff
@@ -697,10 +698,19 @@ v8::Local<v8::Promise> Session::LoadExtension(
     return handle;
   }
 
+  int load_flags = extensions::Extension::FOLLOW_SYMLINKS_ANYWHERE;
+  gin_helper::Dictionary options;
+  if (args->GetNext(&options)) {
+    bool allow_file_access = false;
+    options.Get("allow_file_access", &allow_file_access);
```
samuelmaddock, line 705: Camel case is more natural for Electron properties.
```suggestion
    options.Get("allowFileAccess", &allow_file_access);
```

samuelmaddock: I think this approach is good and lends itself to further extensibility down the road.

shell/browser/extensions/electron_extension_loader.cc
```diff
@@ -155,9 +156,10 @@ void ElectronExtensionLoader::LoadExtensionForReload(
     LoadErrorBehavior load_error_behavior) {
   CHECK(!path.empty());
 
+  int load_flags = Extension::FOLLOW_SYMLINKS_ANYWHERE; // FIXME
```
samuelmaddock, line 159: I think this is a place where `ExtensionPrefs` would be useful. Looking at [Chrome's installed extension loader](https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/extensions/installed_loader.cc;l=687;drc=02fc05043bd38d318ba9ce59ccc4c7602284f990), they use this to store global and specific extension preferences at runtime—including [methods for getting/setting file access.](https://source.chromium.org/chromium/chromium/src/+/master:extensions/browser/extension_prefs.h;l=517;drc=3c2bf33cc5112e89df2ebc668def33b9363f1723)

In `Session::LoadExtension`, we could call `ExtensionPrefs::SetAllowFileAccess` with the value of the `allowFileAccess` option. Then `ElectronExtensionLoader` can derive the `load_flags` from `ExtensionPrefs`. Chrome's [`InstalledLoader::GetCreationFlags`](https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/extensions/installed_loader.cc;l=687;drc=02fc05043bd38d318ba9ce59ccc4c7602284f990) is a good example of how this could work.

shell/browser/api/electron_api_session.cc
ChALkeR: Done, thanks!

shell/browser/extensions/electron_extension_loader.cc
ChALkeR: I thought about that, but the thing I noticed while testing (https://github.com/electron/electron/issues/24011#issuecomment-674315764) is that:
1) This preference doesn't appear to do anything on it's own, or it could do something I don't know yet.
2) In Chromium, this is apparently not what is used.
  * A call to `ExtensionPrefs::SetAllowFileAccess` writes to `{profile}/Preferences`.
  * But in Chromium that doesn't happen when file access is enabled in extension preferences.
  * Instead, Chromium stores that in`{profile}/Current Session` which is inexistent in Electron.

My concern here is that perhaps `ExtensionPrefs::SetAllowFileAccess` does something _else_ and can have some other side effect? Or is that not the case?

I can surely port to that if it would be safe to use it here.

docs/api/session.md
```diff
@@ -592,7 +596,10 @@ const { app, session } = require('electron')
 const path = require('path')
 
 app.on('ready', async () => {
-  await session.defaultSession.loadExtension(path.join(__dirname, 'react-devtools'))
+  await session.defaultSession.loadExtension(
+    path.join(__dirname, 'react-devtools'),
+    { allowFileAccess: true }
```
nornagon, line 601:
```suggestion
    // allowFileAccess is required to load the devtools extension on file:// URLs.
    { allowFileAccess: true }
```

docs/api/session.md
```diff
@@ -567,9 +567,13 @@ will not work on non-persistent (in-memory) sessions.
 
 **Note:** On macOS and Windows 10 this word will be removed from the OS custom dictionary as well
 
-#### `ses.loadExtension(path)`
+#### `ses.loadExtension(path[, options])`
 
 * `path` String - Path to a directory containing an unpacked Chrome extension
+* `options` Object (optional)
+  * `allowFileAccess` Boolean - Whether to allow the extension to read local files over `file://`
```
nornagon, line 574: This also allows injecting content scripts into `file:` urls, yes?

shell/browser/extensions/electron_extension_loader.cc
```diff
@@ -155,9 +156,10 @@ void ElectronExtensionLoader::LoadExtensionForReload(
     LoadErrorBehavior load_error_behavior) {
   CHECK(!path.empty());
```
nornagon, line 157: This whole function should probably be `NOTREACHED()` since we never end up invoking it.

nornagon (approved):

shell/browser/extensions/electron_extension_loader.cc
samuelmaddock: After looking more into this, Chrome implements a private API on the chrome://extensions page with a function `chrome.developerPrivate.updateExtensionConfiguration(config, callback)`. [It's internal implementation](https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/extensions/api/developer_private/developer_private_api.cc;l=891;drc=f8990f540193a784158e5e4d9ec3ac6d1cd76771) calls a utility which in turn calls `ExtensionPrefs::SetAllowFileAccess`.

After this function is called, the value will stay in memory and (I think) persist to disk soon after. An extension loader implements a [`GetCreationFlags`](https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/extensions/installed_loader.cc;l=687;drc=02fc05043bd38d318ba9ce59ccc4c7602284f990) method to then derive flags from `ExtensionPrefs`.

In my opinion, sticking with this pattern close to Chrome will allow us to more naturally build on the extension support in Electron later on.

As for the the location of prefs (`{profile}/Preferences`), I'm not sure what the right path would be us. I think we can go with the default and change it later if needed.

nornagon: I don't believe Electron persists such preferences to disk. Additionally, I'd prefer that the extensions loading stuff doesn't persist anything to disk internally, as that's often surprising for developers (e.g. "i removed the `loadExtension(...)` line from my app, why is the react devtools extension still loading??" behavior that used to be present.)

shell/browser/extensions/electron_extension_loader.cc
samuelmaddock: I looked more into this and it does end up getting called if an extension invokes `chrome.runtime.reload()`. A callchain starts from `ElectronRuntimeAPIDelegate::ReloadExtension`, to `ExtensionRegistrar::ReloadExtension`, and finally calls `LoadExtensionForReload`.

shell/browser/extensions/electron_extension_loader.cc
samuelmaddock: Update on `ExtensionPrefs`, they're mostly used by Chrome to persist extension settings for a profile.
- Whether an extension has file access or can run in incognito mode.
- The extension installation time.
- If an extension is disabled, the reason why.
- And many other Chrome-specific settings.

These, among many other preferences, aren't too relevant to an Electron app and could be managed outside of Electron internal code. However, special cases such as [`chrome.runtime.reload()`](https://developer.chrome.com/extensions/runtime#method-reload) will cause this setting to be dropped without relying on storage such as `ExtensionPrefs`.

Additionally, `ExtensionPrefs` _do_ end up persisting to disk using the same `JsonPrefStore` as `UserPrefs` which are setup in [`ElectronBrowserContext::InitPrefs`](https://github.com/electron/electron/blob/master/shell/browser/electron_browser_context.cc#L169-L185).

To avoid this behavior, I believe it will be necessary to create an [`InMemoryPrefStore`](https://source.chromium.org/chromium/chromium/src/+/master:components/prefs/in_memory_pref_store.h) exclusively for ExtensionPrefs which would be returned from `ElectronExtensionsBrowserClient::GetPrefServiceForContext`. Possibly a command-line flag could be used if a developer wanted to opt-in to persisting these settings?

jkleinsc:
The API WG reviewed this at the Sept 21, 2020 meeting

shell/browser/extensions/electron_extension_loader.cc
samuelmaddock: @ChALkeR to followup, we can simplify this by avoiding `ExtensionPrefs` altogether. Adding a map of `extensionId -> allowFileAccess` could be added to `ElectronExtensionLoader`. When an extension is unloaded, we can clear it. That way we can allow persisting file access if an extension is reloaded.

shell/browser/extensions/electron_extension_loader.cc
```diff
@@ -155,9 +156,10 @@ void ElectronExtensionLoader::LoadExtensionForReload(
...
 
+  int load_flags = Extension::FOLLOW_SYMLINKS_ANYWHERE;  // FIXME
```
nornagon, line 159:
```suggestion
  // TODO(nornagon): we should save whether file access was granted
  // when loading this extension and retain it here. As is, reloading an extension
  // will cause the file access permission to be dropped.
  int load_flags = Extension::FOLLOW_SYMLINKS_ANYWHERE;
```

docs/api/session.md
```diff
@@ -567,9 +567,13 @@ will not work on non-persistent (in-memory) sessions.
...
+    protocol. This is required e.g. for loading devtools extensions on `file://` URLs. Defaults to
+    false.
```
nornagon, line 576:
```suggestion
  * `allowFileAccess` Boolean - Whether to allow the extension to read local files over `file://`
    protocol and inject content scripts into `file://` pages. This is required e.g. for loading
    devtools extensions on `file://` URLs. Defaults to false.
```

nornagon:
I'm happy to land this without the fix for remembering file access across extension reload. The latter is a smaller bug than what this fixes.

rwwagner90:
What's the status on this?

andrewschreiber:
Very important PR to my project. What are the remaining blockers? I'd like to help if possible.

codebytere:
@ChALkeR would you still like to move this forward?

rwwagner90:
This is a necessary feature, so hopefully it can be moved forward 😃. I'm willing to help too, if more help is needed.

safareli:
We really want this to be merged and are ready to help move this forward!

bedney:
Wow! Thank you @nornagon!!

m4heshd:
Finally. Will this be included in v11.2.2?

rwwagner90:
@nornagon I have had a lot of issues with installing the nightly version, but with many forks, I finally got it to work and I can confirm Ember Inspector works again! 🎉 

Will this be released in an 11.x or 12.x version? I would love to get off of these forks and use a more stable version than a nightly.

ChALkeR:
@nornagon sorry for me dropping this on the floor and thanks for handling this!
I was unsure how to continue here and this seemed non-trivial, and then I switched to other tasks...
Thanks again! :+1:

VerteDinde (commented):
Approved for backports on behalf of @electron/wg-releases

bedney:
@nornagon -

Thank you for going the extra mile here with the back ports! We really appreciate it!

benjaffe:
Yes yes, thank you thank you! <3

twitharshil:
@nornagon Will the fix for this issue can be backported to electron v9 which is still on the supported line versions?

nornagon:
@twitharshil no, we are only backporting security fixes to v9.

PR #25222, opened 2020-08-31
docs: add verbiage for experimental API modification

codebytere:
#### Description of Change

Adds language for removal or modification of experimental APIs where none previously existed. 

cc @electron/wg-api 

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/docs/experimental.md
+++ b/docs/experimental.md
@@ -7,17 +7,25 @@ be removed or modified more frequently than other APIs with less warning.
 ## Conditions for an API to be tagged as Experimental
 
 Anyone can request an API be tagged as experimental in a feature PR, disagreements
-on the experimental nature of a feature can be discussed in the API WG if they
+on the experimental nature of a feature can be discussed in the [API Working Group](https://github.com/electron/governance/tree/master/wg-api) if they
 can't be resolved in the PR.
 
 ## Process for removing the Experimental tag
 
 Once an API has been stable and in at least two major stable release lines it
 can be nominated to have its experimental tag removed.  This discussion should
-happen at an API WG meeting.  Things to consider when discussing / nominating:
+occur in the context of an [API Working Group](https://github.com/electron/governance/tree/master/wg-api) meeting.
 
+Things to consider when discussing / nominating an API for stability:
 * The above "two major stables release lines" condition must have been met
 * During that time no major bugs / issues should have been caused by the adoption of this feature
 * The API is stable enough and hasn't been heavily impacted by Chromium upgrades
 * Is anyone using the API?
 * Is the API fulfilling the original proposed usecases, does it have any gaps?
+
+## Modifying Experimental APIs
+
+In the event that we want to nontrivially modify or remove an existing API that has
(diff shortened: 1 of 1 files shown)


docs/experimental.md
```diff
@@ -7,17 +7,25 @@ be removed or modified more frequently than other APIs with less warning.
...
 * During that time no major bugs / issues should have been caused by the adoption of this feature
 * The API is stable enough and hasn't been heavily impacted by Chromium upgrades
 * Is anyone using the API?
 * Is the API fulfilling the original proposed usecases, does it have any gaps?
+
+## Modifying Experimental APIs
+
+In the event that we want to nontrivially modify or remove an existing API that has
+been marked experimental, the process for doing so should follow the process for modifying
+or removing a stable API. The burden for modification or removal shall be lesser than that
```
nornagon, line 30: Lesser in what way?

What makes an experimental API different from a regular API?

codebytere: there may definitely be better ways to phrase this, but by that i meant that we have quite a high bar for removal or breaking modification of existing stable APIs. Since experimental APIs are explicitly stated as subject to change, they should still undergo approval when that occurs, but the bar need not (imo) be quite so high

nornagon: I think we should be more specific about in what ways the requirements should differ.

Perhaps, experimental APIs are not subject to the usual deprecation cycle? When we remove or change stable APIs, we always make sure to have a deprecation cycle where the old way still works but prints a warning before we remove it. Experimental APIs can change or be removed without a deprecation cycle.

MarshallOfSound: I think I would directly translate lesser into excluding or modifying requirements for removing stable APIs.  I don't think we have those documented anywhere so it's hard to actually define where this differs.  I do think we should have some kind of warning somewhere though for the removal of even experimental APIs.  Maybe for experimental APIs we are allowed to backport the deprecation warning?  Effectively backdate the removal?

We also have this problem at the moment where we have quite a few things tagged experimental that have actually been in our codebase for years. (Should be resolved separately)

nornagon: I disagree that we should warn about changes to experimental APIs other than in release notes. An API being experimental should mean it's subject to change without warning.

I think this conversation is muddled by the fact that many of our current APIs that are marked experimental are, in fact, not. When we think of changing experimental APIs, that should not include for example BrowserView, which has been in Electron for years without changing. It _should_ include for example the recently-added `clipboard` APIs, and the `contextBridge` API (though the latter I think is ready to graduate from experimental status).

We've talked previously about hiding experimental APIs behind some sort of flag. If this were the case, apps that wanted to use experimental functionality would have to much more explicitly opt into such functionality. One potential way of handling breaking changes to experimental APIs would be to add a version to that flag, e.g.

```js
app.setExperiments({
  clipboard: 1
})
```

Then, if we in future break the API, we can increment the version number and throw an exception if an unsupported version is requested. That would free us from having to detect the old API and still support it at least enough to throw a warning.

PR #25235, opened 2020-08-31
chore: cherry-pick 44bb4d7abee6 from angle

nornagon:
D3D11: Fix bug with static vertex attributes.

In some specific cases after binding a zero size buffer we could end
up trying to use a buffer storage that was no longer valid. Fix this
by ensuring we don't flush dirty bits when we have an early exit due
to a zero size buffer.

Also adds a regression test.

Bug: chromium:1107433
Change-Id: I9db560e8dd3699abed2bb7fe6d91060148ba1817
Reviewed-on: https://chromium-review.googlesource.com/c/angle/angle/+/2314216
Commit-Queue: Jamie Madill jmadill@chromium.org
Reviewed-by: Geoff Lang geofflang@chromium.org

Notes: Security: backported fix for 1107433.

### DIFF
--- a/patches/angle/.patches
+++ b/patches/angle/.patches
@@ -1,1 +1,2 @@
 update_the_active_texture_cache_before_changing_the_texture_binding.patch
+d3d11_fix_bug_with_static_vertex_attributes.patch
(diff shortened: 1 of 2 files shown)


nornagon:
CI failures are unrelated.

PR #25237, opened 2020-08-31
feat: add serial api support

### LINKED ISSUE #22478: Serial API support
### Preflight Checklist

### Problem Description
The [Serial API](https://wicg.github.io/serial/) is a draft specification from the Web Incubator Community Group. There is a prototype implementation [being developed](https://bugs.chromium.org/p/chromium/issues/detail?id=884928) inside the Chromium project (by me). Developers have asked whether this API will be supported in Electron in addition to being available in regular web browsers. As this API requires the user to grant permissions there is some work required on the Electron side before it will be supported.

### Proposed Solution
[A pattern](https://www.electronjs.org/docs/api/web-contents#event-select-bluetooth-device) for this already seems to exist for the Web Bluetooth API. When `navigator.serial.requestPort()` is called a `select-serial-port` could be fired with a list of available serial ports and a callback. Applications would have the option of presenting their own UI to the user or automatically selecting a particular port. Example,

```js
const { app, BrowserWindow } = require('electron');

let win = null;

app.on('ready', () => {
  win = new BrowserWindow({ width: 800, height: 600 });
  win.webContents.on('select-serial-port', (event, portList, callback) => {
    event.preventDefault()
    let result = portList.find((device) => {
      return port.portName === 'COM1'
    })
    if (!result) {
      callback('')
    } else {
      callback(result.portId)
    }
  })
})
```

In addition an application could choose to bypass the permission-based flow altogether and configure Electron to return all available ports in response to a call to `navigator.serial.getPorts()`.

The Chromium content layer API for this is [`SerialDelegate`](https://source.chromium.org/chromium/chromium/src/+/master:content/public/browser/serial_delegate.h).

### Alternatives Considered
There are existing NodeJS modules for communicating with serial ports, such as [node-serialport](https://serialport.io/). The advantage of supporting the Serial API is the ability for applications to share code between their Electron and web versions without the need for a polyfill layer. I am considering adding one to [the Serial API polyfill](https://github.com/google/web-serial-polyfill) anyways but that is likely more work than hooking up the API to work in Electron using Chromium's built-in implementation.

### Additional Information
[Chrome status entry](https://chromestatus.com/feature/6577673212002304) for the Serial API.

Whatever code is added for this API should be directly applicable for resolving #21311 to allow access to the WebHID API as well.

jkleinsc:
#### Description of Change

This PR enables use of the [Serial API](https://web.dev/serial/) in Electron by exposing several new events on webContents:
* `select-serial-port` - fired when `navigator.serial.requestPort()` is called so that an Electron developer can programmatically select a serial port
* `serial-port-added` - fired when a new serial port is detected.  Emitted after `navigator.serial.requestPort` has been called and `select-serial-port` has fired if a new serial port becomes available.  For example, this event will fire when a new USB device is plugged in.
* `serial-port-removed` - fired when a serial port is removed.  Emitted after `navigator.serial.requestPort` has been called and `select-serial-port` has fired if a serial port has been removed.  For example, this event will fire when a USB device is unplugged.
Resolves #22478
#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added Serial API support


shell/browser/serial/electron_serial_delegate.cc
```diff
@@ -0,0 +1,53 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
```
nornagon, line 1: nit: update copyright

shell/browser/serial/serial_chooser_context.cc
```diff
@@ -0,0 +1,76 @@
...
+
+#include "base/base64.h"
+#include "base/memory/singleton.h"
+#include "content/public/browser/device_service.h"
+#include "mojo/public/cpp/bindings/pending_remote.h"
+
+namespace electron {
+
+SerialChooserContext* SerialChooserContext::GetInstance() {
+  return base::Singleton<SerialChooserContext>::get();
```
nornagon, line 17: `base::Singleton`'s documentation says to avoid using it if possible.

Can we tie the lifetime of this to something that already has a well-defined lifetime? (e.g. `ElectronBrowserContext`)?

If we don't need to invoke the destructor, can we use `base::NoDestructor` instead?

shell/browser/serial/serial_chooser_controller.cc
```diff
@@ -0,0 +1,152 @@
...
+}  // namespace gin
+
+namespace electron {
+
+SerialChooserController::SerialChooserController(
+    content::RenderFrameHost* render_frame_host,
+    std::vector<blink::mojom::SerialPortFilterPtr> filters,
+    content::SerialChooser::Callback callback)
+    : filters_(std::move(filters)), callback_(std::move(callback)) {
+  api_web_contents_ = api::WebContents::From(
```
nornagon, line 56: This needs to listen for the destruction of the webcontents, otherwise this may result in use-after-free.

Perhaps this could be an event on `session` instead of on `WebContents`?

shell/browser/serial/electron_serial_delegate.cc
```diff
@@ -0,0 +1,53 @@
...
+
+ElectronSerialDelegate::ElectronSerialDelegate() = default;
+
+ElectronSerialDelegate::~ElectronSerialDelegate() = default;
+
+std::unique_ptr<content::SerialChooser> ElectronSerialDelegate::RunChooser(
+    content::RenderFrameHost* frame,
+    std::vector<blink::mojom::SerialPortFilterPtr> filters,
+    content::SerialChooser::Callback callback) {
+  chooser_controller_ = std::make_unique<SerialChooserController>(
```
nornagon, line 22: Does the fact that there's only one SerialChooserController at a time mean that multiple WebContents can't have outstanding serial port requests at the same time?

shell/browser/serial/electron_serial_delegate.cc
```diff
@@ -0,0 +1,53 @@
...
+
+bool ElectronSerialDelegate::CanRequestPortPermission(
+    content::RenderFrameHost* frame) {
+  return true;
+}
+
+bool ElectronSerialDelegate::HasPortPermission(
+    content::RenderFrameHost* frame,
+    const device::mojom::SerialPortInfo& port) {
+  return true;
```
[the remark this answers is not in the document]
jkleinsc: @nornagon I updated this code to more closely work like the chromium implementation

shell/browser/serial/serial_chooser_context.cc
jkleinsc: The chromium implementation uses a `BrowserContextKeyedServiceFactory`, so I changed our implementation to do the same.

shell/browser/serial/serial_chooser_controller.cc
jkleinsc: Changed to be an event on session.

shell/browser/serial/electron_serial_delegate.cc
jkleinsc: I think we are good now as the serial chooser context is tied to the browser context.

shell/browser/serial/electron_serial_delegate.cc
```diff
@@ -0,0 +1,53 @@
...
+
+ElectronSerialDelegate::~ElectronSerialDelegate() = default;
+
+std::unique_ptr<content::SerialChooser> ElectronSerialDelegate::RunChooser(
+    content::RenderFrameHost* frame,
+    std::vector<blink::mojom::SerialPortFilterPtr> filters,
+    content::SerialChooser::Callback callback) {
+  chooser_controller_ = std::make_unique<SerialChooserController>(
+      frame, std::move(filters), std::move(callback));
+  return nullptr;
```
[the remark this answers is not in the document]
jkleinsc: @nornagon I think that comment is misleading.  In the chromium implementation what is returned is used to close the UI dialog (eg see https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/ui/views/permission_bubble/chooser_bubble_ui.cc;l=198;drc=897cbf4390c9a5b2ce6d1dd94337c2aca7698e8c), but we don't have a UI dialog to cleanup.

shell/browser/serial/electron_serial_delegate.cc
nornagon: I'm confused, there's still only one chooser.

What happens if two different WebContents both try to choose a serial port at the same time? Will this `RunChooser` method be called twice, replacing the previous `SerialChooserController`?

shell/browser/serial/electron_serial_delegate.cc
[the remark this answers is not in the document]
nornagon: The embedder might show a UI dialog to allow the user to select a serial port, right? How will they know if the request is cancelled so they can close their UI?

shell/browser/serial/electron_serial_delegate.cc
```diff
@@ -0,0 +1,67 @@
...
+}
+
+bool ElectronSerialDelegate::CanRequestPortPermission(
+    content::RenderFrameHost* frame) {
+  return true;
```
nornagon, line 37: Should this be sent through to the permission check handler? https://www.electronjs.org/docs/api/session#sessetpermissioncheckhandlerhandler

MarshallOfSound: Yes, and it should default to rejected

shell/browser/serial/serial_chooser_context.cc
```diff
@@ -0,0 +1,166 @@
...
+  port_manager_->SetClient(client_receiver_.BindNewPipeAndPassRemote());
+}
+
+void SerialChooserContext::OnPortManagerConnectionError() {
+  port_manager_.reset();
+  client_receiver_.reset();
+
+  port_info_.clear();
+
+  std::vector<std::pair<url::Origin, url::Origin>> revoked_origins;
```
nornagon, line 159: We don't seem to do anything with this?

shell/browser/serial/serial_chooser_context.cc
```diff
@@ -0,0 +1,166 @@
...
+SerialChooserContext::~SerialChooserContext() = default;
+
+void SerialChooserContext::GrantPortPermission(
+    const url::Origin& requesting_origin,
+    const url::Origin& embedding_origin,
+    const device::mojom::SerialPortInfo& port) {
+  base::Value value = PortInfoToValue(port);
+  port_info_.insert({port.token, value.Clone()});
+
+  ephemeral_ports_[{requesting_origin, embedding_origin}].insert(port.token);
```
nornagon, line 65: when do we remove entries from `ephemeral_ports_`?

shell/browser/serial/electron_serial_delegate.cc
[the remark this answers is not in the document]
jkleinsc: 👍 I'll update to fire an event in this case.

shell/browser/serial/electron_serial_delegate.cc
jkleinsc: 👍 ok I'll do so.

shell/browser/serial/electron_serial_delegate.cc
jkleinsc: @nornagon let me test this and see what it does.

shell/browser/serial/serial_chooser_context.cc
jkleinsc: Looking at the chromium implementation: https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/serial/serial_chooser_context.cc?q=ephemeral_ports_&ss=chromium, it doesn't look they ever get removed - maybe because they are a list of ports  to which an origin has access to/been granted permission to:
https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/serial/serial_chooser_context.h;l=92

shell/browser/serial/serial_chooser_context.cc
jkleinsc: yup... copied over from original implementation - I'll remove.

shell/browser/serial/electron_serial_delegate.cc
jkleinsc: @nornagon @MarshallOfSound web serial uses [ContentSettingsType::SERIAL_GUARD](https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/serial/serial_chooser_context.cc;l=66) for permissioning, but it doesn't look like there is a corresponding [content::PermissionType](https://source.chromium.org/chromium/chromium/src/+/master:components/permissions/permission_util.cc;l=128).  Any suggestions with what we do with that?

shell/browser/serial/electron_serial_delegate.cc
jkleinsc: @nornagon you are correct.  I'll refactor this to handle multiple web contents requesting a serial port.

shell/browser/serial/electron_serial_delegate.cc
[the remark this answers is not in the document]
jkleinsc: @nornagon I've chased this down further... there is no mechanism to cancel navigator.serial.requestPort().  That return value is simply used to cleanup the UI once the serial service is destroyed: https://source.chromium.org/search?q=chooser_&ss=chromium%2Fchromium%2Fsrc:content%2Fbrowser%2Fserial%2F

shell/browser/serial/electron_serial_delegate.cc
jkleinsc: @MarshallOfSound FWIW the way https://www.electronjs.org/docs/api/session#sessetpermissioncheckhandlerhandler works it defaults to accepted unless you add a handler.

jkleinsc:
The API WG reviewed this at the Sept 21, 2020 meeting.

docs/api/session.md
```diff
@@ -144,6 +144,70 @@ Emitted when a hunspell dictionary file download fails.  For details
...
+Emitted when a serial port needs to be selected when a call to
+`navigator.serial.requestPort` is made. `callback` should be called with
+`portId` to be selected, passing an empty string to `callback` will
+cancel the request.  Additionally, permissioning on `navigator.serial` can
+be managed by using [ses.setPermissionCheckHandler(handler)](#sessetpermissioncheckhandlerhandler)
+with the `serial` permission.
+
+Because this is an experimental feature it is disabled by default.  To enable this feature, you
+will need to use the following command line switches: `--enable-experimental-web-platform-features` and
+`enable-features=ElectronSerialChooser`.
```
nornagon, line 166:
```suggestion
`--enable-features=ElectronSerialChooser`.
```

docs/api/session.md
```diff
@@ -144,6 +144,70 @@ Emitted when a hunspell dictionary file download fails.  For details
...
+
+```javascript
+const { app, BrowserWindow } = require('electron')
+
+let win = null
+app.commandLine.appendSwitch('enable-experimental-web-platform-features')
```
nornagon, line 172: we should probably recommend per-WebContents flagging of this feature via the `experimentalFeatures` WebPreference. Further, is there a way to enable _just_ the serial API rather than all experimental features?

jkleinsc: Ah yes there is `--enable-blink-features=Serial` or `enableBlinkFeatures: 'Serial'` via webpreference.

spec-main/chromium-spec.ts
```diff
@@ -1457,7 +1457,7 @@ describe('iframe using HTML fullscreen API while window is OS-fullscreened', ()
 });
 
 describe('navigator.serial', () => {
-  app.commandLine.appendSwitch('enable-experimental-web-platform-features');
+  app.commandLine.appendSwitch('enable-blink-features', 'Serial');
```
nornagon, line 1460: can this go in the WebPreferences?

nornagon (approved):
lgtm with nit

PR #25246, opened 2020-09-01
revert: "fix: reply notifs sometimes destroyed too early"

MarshallOfSound:
Reverts electron/electron#25100

Notes: Fixed issue where clicking notifications would no longer dismiss them as expected


codebytere:
It transpires that this conflicts with the HIG:

> Clicking a notification when the Mac is unlocked dismisses the notification, removes it from Notification Center, opens the corresponding app, and shows related information.

PR #25249, opened 2020-09-01
chore: upgrade to Node.js v14.9.0

codebytere:
#### Description of Change

Upgrade Node.js to `v14.9.0`

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Upgraded Node.js to v14.9.0.


codebytere:
This PR is ready for initial review - ARM is experiencing some Struggles™️ so i'll sort those out before looking to land this.

jkleinsc (approved):

patches/node/build_add_gn_build_files.patch
```diff
@@ -1176,17 +1163,17 @@ index 0000000000000000000000000000000000000000..2e07ea28437ee25ddaea4d730e1a65a5
...
++    "deps/v8/tools/codemap.js",
++    "deps/v8/tools/consarray.js",
++    "deps/v8/tools/csvparser.js",
++    "deps/v8/tools/profile.js",
++    "deps/v8/tools/profile_view.js",
++    "deps/v8/tools/logreader.js",
++    "deps/v8/tools/arguments.js",
++    "deps/v8/tools/tickprocessor.js",
++    "deps/v8/tools/SourceMap.js",
++    "deps/v8/tools/tickprocessor-driver.js",
```
nornagon, line 1176: hmm, we shouldn't be using the code in `deps/v8`. can we change this back to `//v8`?

patches/node/fix_comment_out_incompatible_crypto_modules.patch
```diff
@@ -9,58 +9,76 @@ with what's exposed through BoringSSL. I plan to upstream parts of this or
...
 +#endif
    BIGNUM* bn_p =
        BN_bin2bn(reinterpret_cast<const unsigned char*>(p), p_len, nullptr);
    BIGNUM* bn_g = BN_new();
-@@ -5168,6 +5170,7 @@ bool DiffieHellman::Init(const char* p, int p_len, int g) {
+@@ -5204,12 +5206,14 @@ bool DiffieHellman::Init(const char* p, int p_len, int g) {
+     BN_free(bn_g);
+     return false;
+   }
++
```
nornagon, line 35: we probably don't need this extra newline

spec-main/api-net-spec.ts
```diff
@@ -1075,20 +1075,6 @@ describe('net module', () => {
...
-        response.end();
-      });
-      const urlRequest = net.request({
-        url: serverUrl,
-        redirect: 'error'
-      });
-      urlRequest.end();
-      await emittedOnce(urlRequest, 'error');
-    });
-
```
nornagon, line 1091: uh? why is this test being deleted?

spec-main/api-net-spec.ts
```diff
@@ -1207,27 +1193,6 @@ describe('net module', () => {
...
-      const serverUrl = await respondOnce.toSingleURL((request, response) => {
-        request.destroy();
-        response.end();
-      });
-      const urlRequest = net.request(serverUrl);
-      urlRequest.end(randomBuffer(kOneMegaByte));
-      const [error] = await emittedOnce(urlRequest, 'error');
-      expect(error.message).to.be.oneOf(['net::ERR_CONNECTION_RESET', 'net::ERR_CONNECTION_ABORTED']);
-    });
-
```
nornagon, line 1230: here too

spec/ts-smoke/electron/main.ts
```diff
@@ -1072,17 +1072,17 @@ shell.writeShortcutLink('/home/user/Desktop/shortcut.lnk', 'update', shell.readS
 
   // Query all cookies.
   session.defaultSession.cookies.get({})
-    .then((cookies) => {
+    .then((cookies: Electron.Cookie[]) => {
```
nornagon, line 1075: these types (all the changes in this file) should be automatically inferred..?

spec/ts-smoke/electron/renderer.ts
```diff
@@ -34,7 +34,7 @@ remote.getCurrentWindow().on('close', () => {
 })
 
 remote.getCurrentWindow().capturePage().then(buf => {
-  fs.writeFile('/tmp/screenshot.png', buf, err => {
+  fs.writeFile('/tmp/screenshot.png', (buf as any), err => {
```
nornagon, line 37: can we do better than `any`? What's the error here?

spec/ts-smoke/electron/renderer.ts
```diff
@@ -109,7 +109,7 @@ crashReporter.start({
 
 const desktopCapturer = require('electron').desktopCapturer
 
-desktopCapturer.getSources({ types: ['window', 'screen'] }).then(sources => {
+desktopCapturer.getSources({ types: ['window', 'screen'] }).then((sources: Electron.DesktopCapturerSource[]) => {
```
nornagon, line 112: this type should be inferred

spec/ts-smoke/electron/renderer.ts
```diff
@@ -130,6 +130,27 @@ desktopCapturer.getSources({ types: ['window', 'screen'] }).then(sources => {
   }
 })
 
+desktopCapturer.getMediaSourceIdForWebContents(remote.getCurrentWebContents().id).then((mediaSourceId: any) => {
```
nornagon, line 133: this method has been removed

nornagon (changes requested):

spec-main/api-net-spec.ts
codebytere: @nornagon https://electronhq.slack.com/archives/C5VT8SQ8K/p1599188065017200

spec-main/api-net-spec.ts
codebytere: https://electronhq.slack.com/archives/C5VT8SQ8K/p1599188065017200

lib/browser/init.ts
```diff
@@ -62,7 +62,7 @@ process.on('uncaughtException', function (error) {
 // Emit 'exit' event on quit.
 const { app } = require('electron');
 
-app.on('quit', function (event, exitCode) {
+app.on('quit', function (_event: Event, exitCode: number) {
```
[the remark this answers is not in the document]
codebytere: i thought so as well, but it consistently fails 🤔

patches/node/build_add_gn_build_files.patch
codebytere: V8 changed this here: https://github.com/v8/v8/commit/56bf834a2b122616d00c6cc85243e4d80a1ca241 which is why this occurred but i agree - we should probably update our filename generation logic accordingly 🤔

lib/browser/remote/server.ts
```diff
@@ -353,7 +353,7 @@ handleRemoteCommand('ELECTRON_BROWSER_REQUIRE', function (event, contextId, modu
     if (customEvent.defaultPrevented) {
       throw new Error(`Blocked remote.require('${moduleName}')`);
     } else {
-      customEvent.returnValue = process.mainModule!.require(moduleName);
+      customEvent.returnValue = (process as any).mainModule.require(moduleName);
```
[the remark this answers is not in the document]
codebytere: It's deprecated: https://nodejs.org/api/process.html#process_process_mainmodule

spec/ts-smoke/electron/renderer.ts
codebytere: Node.js v14 types this more strongly - we can do `buf as unknown as Buffer`perhaps

codebytere: Screen Shot 2020-09-16 at 1 46 55 PM

spec-main/api-net-spec.ts
nornagon: I don't understand why the destroy event would prevent the behavior that this test is testing from changing. We should still not follow redirects when the mode is 'error'.

spec/ts-smoke/electron/renderer.ts
nornagon: surely we shouldn't be trying to write a `NativeImage` to a file?

lib/browser/init.ts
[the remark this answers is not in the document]
MarshallOfSound: This is a symptom of a problem with our TS definitions and depending on `@types/node@14` then and should be fixed before this lands

lib/browser/remote/server.ts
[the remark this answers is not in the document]
MarshallOfSound:
```suggestion
      customEvent.returnValue = require.main.require(moduleName);
```

Should replace with the new thing then?

spec/ts-smoke/electron/main.ts
MarshallOfSound: Yeah if these aren't being inferred anymore we have a big problem

MarshallOfSound (changes requested):
Blocking on the types problem as that will be a big regression for folks if some types stop being auto-inferred 🤔 

Happy to help look into it once I'm not fighting `nan` in the upgrades branch

codebytere:
@MarshallOfSound it's been resolved - just working on verifying one last thing with the net specs atm

lib/browser/init.ts
```diff
@@ -4,6 +4,9 @@ import * as fs from 'fs';
 import { Socket } from 'net';
 import * as path from 'path';
 import * as util from 'util';
+
+// Emit 'exit' event on quit.
+import { app } from 'electron/main';
```
MarshallOfSound, line 9: This can't work? The `electron` import isn't available until `reset-search-paths` and `common/init` are called

lib/browser/remote/server.ts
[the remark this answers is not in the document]
codebytere: This is causing issues so we'll follow up with it since it's still working in v14

spec-main/api-net-spec.ts
codebytere: fixed

lib/browser/api/net.ts
```diff
@@ -481,6 +481,10 @@ export class ClientRequest extends Writable implements Electron.ClientRequest {
   }
 
   _die (err?: Error) {
+    if ((this as any)._writableState.destroyed && err) {
```
nornagon, line 484: Can this just be `this.destroyed`? https://github.com/nodejs/node/blob/fe293e914c3b9a65d2024971ebbefcf8a93dc549/lib/_stream_writable.js#L754

Also, I'm not sure I understand why we need to check for `destroyed` here.

spec-main/api-context-bridge-spec.ts
```diff
@@ -375,7 +375,7 @@ describe('contextBridge', () => {
       if (!useSandbox) {
         it('should release the global hold on methods sent across contexts', async () => {
           await makeBindingWindow(() => {
-            require('electron').ipcRenderer.on('get-gc-info', e => e.sender.send('gc-info', { trackedValues: process._linkedBinding('electron_common_v8_util').getWeaklyTrackedValues().length }));
+            require('electron').ipcRenderer.on('get-gc-info', (e: Electron.IpcRendererEvent) => e.sender.send('gc-info', { trackedValues: process._linkedBinding('electron_common_v8_util').getWeaklyTrackedValues().length }));
```
nornagon, line 378: should be inferred, yes?

spec-main/api-ipc-spec.ts
```diff
@@ -455,7 +455,7 @@ describe('ipc module', () => {
         w.loadURL('about:blank');
         await w.webContents.executeJavaScript(`(${function () {
           const { ipcRenderer } = require('electron');
-          ipcRenderer.on('foo', (e, msg) => {
+          ipcRenderer.on('foo', (_e: Event, msg: string) => {
```
nornagon, line 458: here too?

spec-main/api-net-spec.ts
```diff
@@ -1089,6 +1089,27 @@ describe('net module', () => {
...
+    it('should emit error event on server request destroy', async () => {
+      const serverUrl = await respondOnce.toSingleURL((request, response) => {
+        request.destroy();
+        response.end();
+      });
+      const urlRequest = net.request(serverUrl);
+      urlRequest.end(randomBuffer(kOneMegaByte));
+      const [error] = await emittedOnce(urlRequest, 'error');
+      expect(error.message).to.be.oneOf(['net::ERR_CONNECTION_RESET', 'net::ERR_CONNECTION_ABORTED']);
+    });
```
nornagon, line 1111: can we move this to its original position to zero out the diff?

nornagon (approved):
lgtm with nits

lib/browser/api/net.ts
codebytere: Node.js assumes that any stream which is ended is no longer capable of emitted events - this is imho a faulty assumption (e.g. in the case of incoming data) but causes this since we are listening for an `error` event emitted _after_ `urlRequest.end()`. Node.js doesn't consider the case of an object that is _acting_ like a stream (our `urlRequest`) for which that behavior is not a reasonable expectation. We also need to check for readable/writable state destruction specifically since that's what Node.js [makes its decision based on](https://github.com/nodejs/node/pull/29197/files#diff-0da8feb832c813299da638cf7fdc65c9R10)

nornagon: ah, so if `destroyed` is true then it will emit the error event itself?

MarshallOfSound (approved):

linonetwo:
Is this added to the v11 beta? This fixes an important issue about worker_thread and we hope we can adapt this.

Issue #24407, opened 2020-07-03
Multiple app dock icons shown


CyrusRoshan:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  *  9.0.5
* **Operating System:**
  *  macOS 10.15.5
* **Last Known Working Electron version:**
  *  not sure

### Expected Behavior

With the example program at [https://github.com/CyrusRoshan/electron-multiple-dock-icon](https://github.com/CyrusRoshan/electron-multiple-dock-icon), I would expect the dock icon to be hidden after the initial click, or multiple quick clicks in succession.

### Actual Behavior

If clicking once, dock icon behaves as expected, and disappears. If quickly clicking multiple times, dock icons are quickly added. These dock icons persist after app shutdown.

### To Reproduce

Clone and run this [electron-quick-start fork](https://github.com/CyrusRoshan/electron-multiple-dock-icon), then quickly click the dock icon.

### Screenshots

![image](https://user-images.githubusercontent.com/8443502/86441896-c3ecaf80-bcc1-11ea-8fda-619506d19331.png), can also upload a video if that'd help.

### Additional Information

This may have something to do with the following lines:

```
win.show();
win.setVisibleOnAllWorkspaces(true);
win.setAlwaysOnTop(true, 'pop-up-menu');
win.hide();
```

which are used for keeping BrowserWindows always on top [over fullscreen apps](https://github.com/electron/electron/issues/10078#issuecomment-331581160).

SMLVG:
Try deleting 'pop-up-menu'

CyrusRoshan:
The point is more that `app.dock.hide()` `app.dock.show()` should just [show and hide the app dock icon](https://www.electronjs.org/docs/api/dock#dockhide-macos), whereas this example results in multiple icons. Changing the window positioning isn't supposed to affect that.

bpasero:
We have seen this issue in VSCode and our only workaround was to never call any dock related API.

PR #25269, opened 2020-09-02
fix: multiple dock icons when calling dock.show/hide

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/24407.

Work around a bug of macOS that, when calling dock.show/hide in very short interval, multiple dock icons would be left in the system.

I don't have a test for this, since it is a visual bug of Finder and the APIs are just lying.

#### Checklist

#### Release Notes

Notes: Fix multiple dock icons being left in system when calling `dock.show`/`hide` on macOS.


zcbenz:
@codebytere I have adjusted the order of the tests to make it pass, it is not really decent but I can't find a better way.

codebytere (approved):

PR #25284, opened 2020-09-02
feat: [net] add "credentials" option to net.request

nornagon:
#### Description of Change
This adds a new option to `net.request()`, "credentials", which more closely matches the behavior of `fetch()` than does "useSessionCookies".

If "credentials" is specified it overrides "useSessionCookies", otherwise the default behavior remains unchanged.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added a new "credentials" option for `net.request()`.


nornagon:
NB. this change depends on the current master roll: https://github.com/electron/electron/pull/24984

deepak1556 (approved):
Api looks better now 👍 

Also `useSessionCookies` was added for https://github.com/electron/electron/issues/8891 where users where only concerned with whether the cookies are completely handled by the store or not. So this change is reasonable for that too.

nornagon:
@MarshallOfSound this is not semver/major because it doesn't break old behavior. it'll be semver/major when we remove the old one.

shell/browser/api/electron_api_url_loader.cc
```diff
@@ -353,6 +376,9 @@ gin::Handle<SimpleURLLoaderWrapper> SimpleURLLoaderWrapper::Create(
   opts.Get("method", &request->method);
   opts.Get("url", &request->url);
   opts.Get("referrer", &request->referrer);
+  bool credentials_specified =
+      opts.Get("credentials", &request->credentials_mode);
+  LOG(INFO) << "credentials_specified = " << credentials_specified;
```
deepak1556, line 381: log can be removed ?

deepak1556 (approved):
LGTM

jkleinsc:

jkleinsc (changes requested):
This needs a rebase from master to fix the mac arm64 compile issues.

nornagon:
@jkleinsc ping

jkleinsc (approved):

PR #25291, opened 2020-09-02
docs: update menu example to avoid remote

nornagon:
#### Description of Change
Replaced with an example using `ipcRenderer.invoke()`.

I'm not thrilled with this example, but the `Menu` API as it is leaves little choice. This would be much easier with an API like `Menu.on('item-clicked')` and `Menu.on('did-close-menu')`.

Ref https://github.com/electron/electron/issues/21408#issuecomment-674519033

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/menu.md
+++ b/docs/api/menu.md
@@ -255,55 +255,29 @@ To create menus initiated by the renderer process, send the required
 information to the main process using IPC and have the main process display the
 menu on behalf of the renderer.
 
-Below is an example of creating a menu with dynamic information in a web page,
-and showing it when the user right clicks the page:
-
-```html
-<!-- index.html -->
-<script>
-const { ipcRenderer } = require('electron')
-
-window.addEventListener('contextmenu', (e) => {
-  e.preventDefault()
-  const template = [
-    { label: 'Menu Item 1', id: 'menu-item-1' },
-    { type: 'separator' },
-    { label: 'Menu Item 2', type: 'checkbox', checked: true, id: 'menu-item-2' },
-  ]
-  ipcRenderer.invoke('show-menu', template).then((clickedId) => {
-    if (clickedId === 'menu-item-1') {
-      console.log('item 1 clicked')
-    }
-  })
-})
-</script>
-```
+Below is an example of showing a menu when the user right clicks the page:
 
 ```js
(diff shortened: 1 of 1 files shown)


docs/api/menu.md
```diff
@@ -257,28 +251,62 @@ Menu.setApplicationMenu(menu)
...
 ```
 
+```js
+// main.js
+const { BrowserWindow, ipcMain, Menu } = require('electron')
+
+ipcMain.handle('show-menu', (event, template) => {
+  const menu = Menu.buildFromTemplate(template)
+  menu.popup(BrowserWindow.fromWebContents(event.sender))
+  return new Promise((resolve, reject) => {
```
MarshallOfSound, line 289: This is kinda nasty, is there a better way of doing this over IPC? It's a common use case and this isn't super ideal 🤔  We should think about a nice way of doing this before recommending this particular technique.

nornagon: I'm open to suggestions, this is the best I've thought of so far.

See the PR description for proposed changes to the `Menu` API that would simplify this.

zcbenz: How about using a simple example like this:

```js
// renderer
window.addEventListener('contextmenu', (e) => {
   e.preventDefault()
   ipcRenderer.send('show-context-menu')
 })

ipcRenderer.on('context-menu-command', () => { ... })

// main
ipcMain.on('show-context-menu', (event) => {
  const template = [
    { label: 'Menu Item 1', click: () => { event.sender.send('context-menu-command') } },
    { type: 'separator' },
    { label: 'Menu Item 2', type: 'checkbox', checked: true },
  ]
  const menu = Menu.buildFromTemplate(template)
  menu.popup(BrowserWindow.fromWebContents(event.sender))
})
```

The purpose of the example is to show how to show menu from renderer process, so a ping-pong style example could probably serve the purpose best.

nornagon: @zcbenz seems reasonable. Updated.

zcbenz (approved):

zcbenz:
There is a lint error needs to be solved.

MarshallOfSound (approved):

PR #25292, opened 2020-09-02
fix: only focus a webContents if the window was not initially hidden

### LINKED ISSUE #25287: Missing focus event on app activation, for window created with show: false
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 10.1.1
* **Operating System:**
  * macOS 10.15.4

We recently started launching as hidden on macOS (that is, `show: false` or `browserWindow.hide()` while loading the page, followed by `browserWindow.show()` when the dock icon is clicked). This results in some busted focus behavior on the first activation of the window. The second activation restores it.

### Expected Behavior
The guest page receives a `focus` event on the first activation of the app, when created with `show: false`.

### Actual Behavior
The guest page doesn't receive the first `focus` event. Subsequent activations do trigger a `focus` event.

### To Reproduce
[Electron Fiddle](https://gist.github.com/CharlieHess/c507468cd40630d04f58c6514100b997)

MarshallOfSound:
Fixes #25287

Notes: Fixed an issue where `document.hasFocus` and `document.activeElement` would be inconsistent upon showing a window that was initially created with `show: false`

### DIFF
--- a/lib/browser/api/browser-window.ts
+++ b/lib/browser/api/browser-window.ts
@@ -28,9 +28,11 @@ Object.setPrototypeOf(BrowserWindow.prototype, BaseWindow.prototype);
   // Though this hack is only needed on macOS when the app is launched from
   // Finder, we still do it on all platforms in case of other bugs we don't
   // know.
-  this.webContents.once('load-url' as any, function (this: WebContents) {
-    this.focus();
-  });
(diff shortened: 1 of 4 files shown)


zcbenz (approved):
I tested a few cases and I confirm it does not have side effects.

codebytere (approved):

PR #25293, opened 2020-09-02
chore: deprecate remote

nornagon:
#### Description of Change
Ref #21408.

This adds a new deprecation warning message when `require('electron').remote` is encountered.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Deprecated the `remote` module. It is replaced by [`@electron/remote`](https://github.com/electron/remote).


docs/breaking-changes.md
```diff
@@ -50,6 +50,25 @@ If your crash ingestion server does not support compressed payloads, you can
 turn off compression by specifying `{ compress: false }` in the crash reporter
 options.
 
+### Deprecated: `remote` module
+
+The `remote` module is deprecated in Electron 12, and will be removed in
```
MarshallOfSound, line 55: Can we utilize this message to further indicate that you shouldn't just replace and forget.  Let's use it as another chance to educate why using `remote` is bad.

lib/renderer/api/remote.ts
```diff
@@ -3,9 +3,12 @@ import { isPromise, isSerializableObject, serialize, deserialize } from '../../c
 import { MetaTypeFromRenderer, ObjectMember, ObjProtoDescriptor, MetaType } from '../../common/remote/types';
 import { ipcRendererInternal } from '../ipc-renderer-internal';
 import type { BrowserWindow, WebContents } from 'electron/main';
+import deprecate from '@electron/internal/common/api/deprecate';
 import { browserModuleNames } from '@electron/internal/browser/api/module-names';
 import { commonModuleList } from '@electron/internal/common/api/module-list';
 
+deprecate.log('The remote module is deprecated. Use https://github.com/electron/remote instead.');
```
MarshallOfSound, line 10: Same here

jkleinsc (changes requested):
We should also update this doc to note that remote is deprecated: 
https://www.electronjs.org/docs/api/remote

docs/breaking-changes.md
nornagon: The @electron/remote repository has a big warning at the top. I think that's enough.

nornagon:
@jkleinsc that's already been done in master. #24941

jkleinsc (approved):
Doh! @nornagon I missed that the doc had been updated.

codebytere (approved):

zcbenz (approved):

Issue #25141, opened 2020-08-26
Renderer crash after calling navigator.setAppBadge(42)


zarubond:
### Preflight Checklist

* [ ] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [ ] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [ ] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 9, 10
* **Operating System:**
  * Windows
* **Last Known Working Electron version:**
  *  none

### Expected Behavior
Calling navigator.setAppBadge(42) should create an application badge on taskbar with the number 42.

### Actual Behavior
Render process crashed after calling navigator.setAppBadge(42) 

### To Reproduce
Start Electron, open console and try to execute navigator.setAppBadge(42) 

### Additional Information

https://web.dev/badging-api/

codebytere:
Seems to be the following: 
> ```Terminating render process for bad Mojo message: Received bad user message: No binder found for interface blink.mojom.BadgeService for the frame/document scope```

zarubond:
Electron does support badges on Mac/Linux but not on Windows. On Windows I can use overlays but that is not very convenient. Having one standardized API would be simply be better.

thomas-darling:
@codebytere

Yep, this definitely needs to be at least mitigated - I just spend two whole days trying to debug a crash caused by this.

May I ask how you found that log message?
Because I never found _any_ way to get _any_ useful log messages when the renderer crashes - and honestly, even though it is working for us right now, it's been so frustrating to debug this, that we are seriously considering just giving up on Electron, and dropping our plans for a desktop product. Maintaining a desktop package is just not worth the development costs, if we cannot properly debug those types of errors. So I'm hoping I'm just missing something obvious here 🙂 

Also, regarding the badge API - I think Electron should actually support this API. Having your own API is fine, if that provides more options, but I think it makes sense that the standard browser API should work too. Either that, or the methods need to be removed from the `navigator` object, so our feature detection doesn't mistakenly think they can be called...

codebytere:
[...]

PR #25294, opened 2020-09-03
fix: bind fake mojo service for badging

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/25141.

Bind a fake mojo service for the [Badging API](https://web.dev/badging-api/). When users add e.g. `navigator.setAppBadge(42)` to the renderer, they will now see

```
The Chromium Badging API is not available in Electron
```

logged as a warning.

cc @nornagon @MarshallOfSound @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash in the renderer process when invoking the Badging API.

### DIFF
--- a/shell/browser/electron_browser_client.cc
+++ b/shell/browser/electron_browser_client.cc
@@ -97,6 +97,7 @@
 #include "shell/common/options_switches.h"
 #include "shell/common/platform_util.h"
 #include "third_party/blink/public/common/loader/url_loader_throttle.h"
+#include "third_party/blink/public/mojom/badging/badging.mojom.h"
 #include "ui/base/resource/resource_bundle.h"
 #include "ui/native_theme/native_theme.h"
 #include "v8/include/v8.h"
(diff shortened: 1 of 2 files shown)


MarshallOfSound (commented):
Quick test? 😄

jkleinsc (approved):

ckerr (approved):

nornagon (commented):
Is it possible we could actually support the badging service? We do expose the relevant APIs...

codebytere:
@nornagon not as it stands without massive patching, given that the Chromium impl is per-Profile.

For now, i think in the short term we should mitigate this crash, although I _do_ think there are some paths forward here. At minimum, we should look towards adding badging support for Windows. Once that's done, I think we can look to bind this to our own badging APIs 🤔  I can open a follow-up issue about adding Windows support if that seems like something we feel would be the next best step.

nornagon:
Can we expose the events to the main process and let the app figure it out?

codebytere:
Potentially? Wiring up the events would almost certainly require a fair bit of patching as well, and at that point it seems more prudent to actually support badging on Windows and wire it up more properly 🤔

nornagon:
I think it still makes sense to give the app control over what's allowed to badge, for instance.

zcbenz (approved):

zcbenz:
I assume it is OK to merge this PR? Providing some APIs to implement `navigator.setAppBadge` would be great, but this PR is only about fixing a crash and does not conflict with that purpose.

PR #25296, opened 2020-09-03
fix: avoid creating client_id file for empty DIR_CRASH_DUMPS

deepak1556:
#### Description of Change

Node child process don't set `DIR_CRASH_DUMPS` unless the undocumented `BREAKPAD_DUMP_LOCATION ` is specified, in these situations client_id file will be generated under the working directory. 

Refs https://github.com/microsoft/vscode/issues/105743

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: fix client_id file being generated in the working directory for node child process

### DIFF
--- a/shell/browser/api/electron_api_crash_reporter.cc
+++ b/shell/browser/api/electron_api_crash_reporter.cc
@@ -86,18 +86,22 @@ const std::map<std::string, std::string>& GetGlobalCrashKeys() {
   return GetGlobalCrashKeysMutable();
 }
 
-base::FilePath GetClientIdPath() {
-  base::FilePath path;
-  base::PathService::Get(electron::DIR_CRASH_DUMPS, &path);
-  return path.Append("client_id");
+bool GetClientIdPath(base::FilePath* path) {
+  if (base::PathService::Get(electron::DIR_CRASH_DUMPS, path)) {
+    *path = path->Append("client_id");
+    return true;
+  }
+  return false;
 }
 
 std::string ReadClientId() {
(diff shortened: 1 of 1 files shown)


jkleinsc (approved):

nornagon (approved):

nornagon:
hm... does this mean crashes in node child processes on linux won't come with a `guid` parameter?

Perhaps we should set `BREAKPAD_DUMP_LOCATION` for node child processes?

deepak1556:
Yeah thought about that, but I don't see a reasonable way to specify the value as `base::PathService` will be uninitialized in the node_main which blocks us from extracting the `DIR_USER_DATA`. If we propagate the user data path via the process object then we can set the value here https://github.com/electron/electron/blob/master/patches/node/refactor_alter_child_process_fork_to_use_execute_script_with.patch or else we can just document `BREAKPAD_DUMP_LOCATION ` for apps to specify.

nornagon:
I was imagining setting the env var in `childProcess.fork()` as in that patch. But I guess there are other ways that apps could launch a `RUN_AS_NODE` process...

nornagon:
I think the ultimate path forward here is to switch to crashpad on Linux, which we'll do eventually anyway.

deepak1556:
> I think the ultimate path forward here is to switch to crashpad on Linux, which we'll do eventually anyway.

yep

FWIW crash reporting from node child process wasn't working at all on linux before the refactor, so I think we can leave this as such until the crashpad move is made.

nornagon (approved):

PR #25297, opened 2020-09-03
fix: multiple fixes to redirects in intercepted requests

marcin-prochownik:
#### Description of Change
Minor fixes to handling intercepted requests:

- 'content-length' wasn't passed down the stream
- for redirects,  methods were not converted to 'GET' and caused too many redirects
- intermittent access violation during redirects

#### Checklist
- [ ] `npm test` passes
- [x] [PR release notes]

#### Release Notes

Notes: Fixed following issues for intercepted requests:
- Fixed issue where on redirects methods were not converted to 'GET' and caused too many redirects
- Fixed 'content-length' not being passed down the stream 
- Fixed intermittent access violation that could occur during handling redirects

### DIFF
--- a/shell/browser/net/electron_url_loader_factory.cc
+++ b/shell/browser/net/electron_url_loader_factory.cc
@@ -259,7 +259,8 @@ void ElectronURLLoaderFactory::StartLoading(
   // API in WebRequestProxyingURLLoaderFactory.
   std::string location;
   if (head->headers->IsRedirect(&location)) {
-    auto new_method = ComputeMethodForRedirect(request.method, head->headers->response_code());
+    auto new_method = ComputeMethodForRedirect(request.method,
+                                               head->headers->response_code());
     auto new_location = request.url.Resolve(location);
     auto new_site_for_cookies = net::SiteForCookies::FromUrl(new_location);
 
--- a/shell/browser/net/electron_url_loader_factory.cc
+++ b/shell/browser/net/electron_url_loader_factory.cc
@@ -194,6 +194,24 @@ void ElectronURLLoaderFactory::Clone(
   receivers_.Add(this, std::move(receiver));
 }
 
+std::string ComputeMethodForRedirect(const std::string& method,
+                                     int http_status_code) {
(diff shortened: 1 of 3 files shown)


shell/browser/net/proxying_url_loader_factory.h
```diff
@@ -238,6 +238,7 @@ class ProxyingURLLoaderFactory
   content::BrowserContext* const browser_context_;
   const int render_process_id_;
   uint64_t* request_id_generator_;  // managed by ElectronBrowserClient
+  std::set<int32_t> intercepted_requests_;
```
nornagon, line 241: I think this set grows without bound? Which would be a memory leak and thus not acceptable to merge.

shell/browser/net/electron_url_loader_factory.cc
```diff
@@ -189,6 +194,24 @@ void ElectronURLLoaderFactory::Clone(
...
+std::string ComputeMethodForRedirect(const std::string& method,
+                                     int http_status_code) {
+  // For 303 redirects, all request methods except HEAD are converted to GET,
+  // as per the latest httpbis draft.  The draft also allows POST requests to
+  // be converted to GETs when following 301/302 redirects, for historical
+  // reasons. Most major browsers do this and so shall we.  Both RFC 2616 and
+  // the httpbis draft say to prompt the user to confirm the generation of new
+  // requests, other than GET and HEAD requests, but IE omits these prompts and
+  // so shall we.
+  // See: https://tools.ietf.org/html/rfc7231#section-6.4
```
nornagon, line 206: I believe this should be implemented upstream, not by us.

See https://source.chromium.org/chromium/chromium/src/+/master:services/network/cors/cors_url_loader.cc;l=359-367;drc=e3aec06168fddef3f2a085fb1f6282b2b9920480

nornagon (changes requested):
This appears to be three separate changes. Would it be possible to break this apart into 3 PRs so they can be reviewed independently?

marcin-prochownik:
I'll create three separate PRs.

PR #25301, opened 2020-09-03
fix: multiple dock icons when calling dock.show/hide

zcbenz:
Backport of #25269

See that PR for details.

Notes: Fix multiple dock icons being left in system when calling dock.show/hide on macOS.


zcbenz:
Note that some changes in `api-app-spec.ts` are added by linter automatically.

codebytere (approved):

PR #25317, opened 2020-09-04
chore: log hint on renderer crash

codebytere:
#### Description of Change

Refs https://github.com/electron/electron/issues/25141#issuecomment-686719615.

It's clear that we can and should provide more helpful tips when debugging certain types of crashes in Electron, but this seems like a small and helpful step we can take towards helping users debug renderer crashes. This will now log a small hint to console when the renderer crashes, telling them that they can set the logging env var to potentially see more information.

cc @ckerr @jkleinsc @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added a small console hint to console to help debug renderer crashes.

### DIFF
--- a/lib/browser/api/web-contents.ts
+++ b/lib/browser/api/web-contents.ts
@@ -461,6 +461,11 @@ const addReturnValueToEvent = (event: any) => {
   });
 };
 
+const loggingEnabled = () => {
+  return Object.prototype.hasOwnProperty.call(process.env, 'ELECTRON_ENABLE_LOGGING') ||
+    process.argv.some(arg => arg.toLowerCase().startsWith('--enable-logging'));
+};
+
 // Add JavaScript wrappers for WebContents class.
 WebContents.prototype._init = function () {
   // The navigation controller.
@@ -545,8 +550,8 @@ WebContents.prototype._init = function () {
     app.emit('renderer-process-crashed', event, this, ...args);
 
     // Log out a hint to help users better debug renderer crashes.
-    if (!process.env.ELECTRON_ENABLE_LOGGING) {
-      console.info('Renderer process crashed - setting ELECTRON_ENABLE_LOGGING may provide more information.');
+    if (loggingEnabled()) {
+      console.info('Renderer process crashed - see https://www.electronjs.org/docs/tutorial/application-debugging for potential debugging information.');
     }
   });
 
--- a/lib/browser/api/web-contents.ts
+++ b/lib/browser/api/web-contents.ts
@@ -543,6 +543,11 @@ WebContents.prototype._init = function () {
 
   this.on('crashed', (event, ...args) => {
     app.emit('renderer-process-crashed', event, this, ...args);
+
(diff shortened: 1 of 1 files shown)


lib/browser/api/web-contents.ts
```diff
@@ -543,6 +543,11 @@ WebContents.prototype._init = function () {
 
   this.on('crashed', (event, ...args) => {
     app.emit('renderer-process-crashed', event, this, ...args);
+
+    // Log out a hint to help users better debug renderer crashes.
+    if (!process.env.ELECTRON_ENABLE_LOGGING) {
```
ckerr, line 548: * (nit) Electron's logging doesn't care if the value of `ELECTRON_ENABLE_LOGGING` is truthy or not; only that it exists. So it's more accurate to see if process.env hasOwnProperty named `ELECTRON_ENABLE_LOGGING`

* (not nit) We also need to check for `--enable-logging` which is currently supported according to docs, and is _definitely_ supported in the copy-logging-from-headless branch.

Possibly overkill, but more accurate:

```js
function isLogging () { 
  return Object.prototype.hasOwnProperty.call(process.env, 'ELECTRON_ENABLE_LOGGING') ||
    process.argv.some(arg => arg.toLowerCase().startsWith('--enable-logging'));
}
```

(The startsWith() check is because the copy-headless-logging PR allows for --enable-logging=foo)

zcbenz (approved):

lib/browser/api/web-contents.ts
```diff
@@ -461,6 +461,11 @@ const addReturnValueToEvent = (event: any) => {
   });
 };
 
+const loggingEnabled = () => {
+  return Object.prototype.hasOwnProperty.call(process.env, 'ELECTRON_ENABLE_LOGGING') ||
+    process.argv.some(arg => arg.toLowerCase().startsWith('--enable-logging'));
```
nornagon, line 466: better to use `app.commandLine` to check this sort of thing.

```suggestion
    app.commandLine.hasSwitch('enable-logging')
```

lib/browser/api/web-contents.ts
```diff
@@ -543,6 +548,11 @@ WebContents.prototype._init = function () {
 
   this.on('crashed', (event, ...args) => {
     app.emit('renderer-process-crashed', event, this, ...args);
+
+    // Log out a hint to help users better debug renderer crashes.
+    if (loggingEnabled()) {
+      console.info('Renderer process crashed - see https://www.electronjs.org/docs/tutorial/application-debugging for potential debugging information.');
```
nornagon, line 554: Can we give some more info about the crash here?

lib/browser/api/web-contents.ts
miniak: https://github.com/electron/electron/pull/25592

lib/browser/api/web-contents.ts
miniak: https://github.com/electron/electron/pull/25592

Issue #25039, opened 2020-08-19
MacOS powerMonitor events, fast user switching detection


camhart:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description
On MacOS, I can't find any way to detect when "fast user switching" has occurred.  The powerMonitor events lock-screen/unlock-screen don't fire (technically the screen isn't being locked).

### Proposed Solution
I'd propose 2 new events be added that signal user switch notifications (one for kEventSystemUserSessionActivated and one for kEventSystemUserSessionDeactivated).  See https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPMultipleUsers/Concepts/UserSwitchNotifications.html#//apple_ref/doc/uid/20002210-CJBJDAGF for more details.

### Alternatives Considered
A few potential alternatives are mentioned at https://stackoverflow.com/questions/8519051/api-to-detect-active-session-in-mac-os-x-with-fast-user-switch.

camhart:
@codebytere Sorry to bother you.  Could you explain how the PR that resolves this will get into production?  I was thinking it would show up in electron 11, but it's still now showing at https://www.electronjs.org/docs/api/power-monitor.

camhart:
It looks like this will be released with electron 13.

For anyone who needs a work around, you can watch /dev/console file's owner to determine who's currently logged into the gui.

```
fs.watch('/dev/console', async (eventType) => {
  let fileInfo = await fs.promises.stat('/dev/console')
  console.log(`current gui uid = ${fileInfo.uid}`)
})
```

This doesn't work on Windows.

PR #25321, opened 2020-09-04
feat: add macOS fast user switching evt

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/25039.

Adds fast user switching detection, which occurs when one logs out and then back in again as another user. This is distinct from `suspend`/`resume` - since no suspension occurs neither event is fired in this case.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Adds fast user switching event to powerMonitor on macOS.

### DIFF
--- a/docs/api/power-monitor.md
+++ b/docs/api/power-monitor.md
@@ -39,6 +39,14 @@ Emitted when the system is about to lock the screen.
 
 Emitted as soon as the systems screen is unlocked.
 
+### Event: 'user-did-become-active' _macOS_
+
+Emitted when a login session is activated. See [documentation](https://developer.apple.com/documentation/appkit/nsworkspacesessiondidbecomeactivenotification?language=objc) for more information.
+
+### Event: 'user-did-resign-active' _macOS_
+
+Emitted when a login session is deactivated. See [documentation](https://developer.apple.com/documentation/appkit/nsworkspacesessiondidresignactivenotification?language=objc) for more information.
+
 ## Methods
 
 The `powerMonitor` module has the following methods:
--- a/shell/browser/api/electron_api_power_monitor_mac.mm
+++ b/shell/browser/api/electron_api_power_monitor_mac.mm
@@ -45,6 +45,18 @@
                    selector:@selector(isResuming:)
                        name:NSWorkspaceDidWakeNotification
                      object:nil];
+    // A notification that the workspace posts when the user session becomes
+    // active.
+    [distCenter addObserver:self
+                   selector:@selector(onUserDidBecomeActive:)
+                       name:NSWorkspaceSessionDidBecomeActiveNotification
+                     object:nil];
+    // A notification that the workspace posts when the user session becomes
+    // inactive.
+    [distCenter addObserver:self
+                   selector:@selector(onUserDidResignActive:)
+                       name:NSWorkspaceSessionDidResignActiveNotification
+                     object:nil];
   }
   return self;
 }
@@ -79,6 +91,18 @@
 - (void)onScreenUnlocked:(NSNotification*)notification {
   for (auto* emitter : self->emitters) {
     emitter->Emit("unlock-screen");
+  }
+}
+
+- (void)onUserDidBecomeActive:(NSNotification*)notification {
+  for (auto* emitter : self->emitters) {
+    emitter->Emit("user-did-become-active");
+  }
+}
+
+- (void)onUserDidResignActive:(NSNotification*)notification {
+  for (auto* emitter : self->emitters) {
+    emitter->Emit("user-did-resign-active");
   }
 }
 


zcbenz (approved):

docs/api/power-monitor.md
```diff
@@ -39,6 +39,14 @@ Emitted when the system is about to lock the screen.
 
 Emitted as soon as the systems screen is unlocked.
 
+### Event: 'user-became-active' _macOS_
+
+Emitted when a login session is activated. See [documentation](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPMultipleUsers/Concepts/UserSwitchNotifications.html#//apple_ref/doc/uid/20002210-CJBJDAGF) for more information.
+
+### Event: 'user-became-inactive' _macOS_
```
nornagon, line 46: any reason to name this differently to the native event?

codebytere: @nornagon you mean e.g. `session-did-become-active`? i named it this way mostly bc i felt it was more clear from a user perspective, but am happy to change it.

nornagon: sorry, i specifically meant the native event being "resign-active" rather than "became-inactive".

I don't mind so much using the word `user`, as `session` has another unrelated meaning in Electron and I think it would be more confusing to overload that term than to follow the underlying naming in that respect.

however, given that we aren't writing docs on this, merely exposing the underlying event and pointing at the native docs, i think we should follow the naming in the docs we're linking as closely as possible for minimal confusion.

codebytere: done!

docs/api/power-monitor.md
```diff
@@ -39,6 +39,14 @@ Emitted when the system is about to lock the screen.
 
 Emitted as soon as the systems screen is unlocked.
 
+### Event: 'user-did-become-active' _macOS_
+
+Emitted when a login session is activated. See [documentation](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPMultipleUsers/Concepts/UserSwitchNotifications.html#//apple_ref/doc/uid/20002210-CJBJDAGF) for more information.
```
nornagon, line 44: Let's link to the current docs:
```suggestion
Emitted when a login session is activated. See [documentation](https://developer.apple.com/documentation/appkit/nsworkspacesessiondidbecomeactivenotification?language=objc) for more information.
```

docs/api/power-monitor.md
```diff
@@ -39,6 +39,14 @@ Emitted when the system is about to lock the screen.
...
+
+### Event: 'user-did-resign-active' _macOS_
+
+Emitted when a login session is deactivated. See [documentation](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPMultipleUsers/Concepts/UserSwitchNotifications.html#//apple_ref/doc/uid/20002210-CJBJDAGF) for more information.
```
nornagon, line 48:
```suggestion
Emitted when a login session is deactivated. See [documentation](https://developer.apple.com/documentation/appkit/nsworkspacesessiondidresignactivenotification?language=objc) for more information.
```

shell/browser/api/electron_api_power_monitor_mac.mm
```diff
@@ -45,6 +45,18 @@ - (id)init {
...
+    // A notification that the workspace posts when the user session becomes
+    // active.
+    [distCenter addObserver:self
+                   selector:@selector(onUserBecameActive:)
+                       name:NSWorkspaceSessionDidBecomeActiveNotification
+                     object:nil];
+    // A notification that the workspace posts when the user session becomes
+    // inactive.
+    [distCenter addObserver:self
+                   selector:@selector(onUserBecameInactive:)
```
nornagon, line 57:
```suggestion
                   selector:@selector(onUserDidResignActive:)
```

shell/browser/api/electron_api_power_monitor_mac.mm
```diff
@@ -82,6 +94,18 @@ - (void)onScreenUnlocked:(NSNotification*)notification {
   }
 }
 
+- (void)onUserBecameActive:(NSNotification*)notification {
+  for (auto* emitter : self->emitters) {
+    emitter->Emit("user-became-active");
+  }
+}
+
+- (void)onUserBecameInactive:(NSNotification*)notification {
```
nornagon, line 103:
```suggestion
- (void)onUserDidResignActive:(NSNotification*)notification {
```

shell/browser/api/electron_api_power_monitor_mac.mm
```diff
@@ -45,6 +45,18 @@ - (id)init {
                    selector:@selector(isResuming:)
                        name:NSWorkspaceDidWakeNotification
                      object:nil];
+    // A notification that the workspace posts when the user session becomes
+    // active.
+    [distCenter addObserver:self
+                   selector:@selector(onUserBecameActive:)
```
nornagon, line 51:
```suggestion
                   selector:@selector(onUserDidBecomeActive:)
```

shell/browser/api/electron_api_power_monitor_mac.mm
```diff
@@ -82,6 +94,18 @@ - (void)onScreenUnlocked:(NSNotification*)notification {
   }
 }
 
+- (void)onUserBecameActive:(NSNotification*)notification {
```
nornagon, line 97:
```suggestion
- (void)onUserDidBecomeActive:(NSNotification*)notification {
```

shell/browser/api/electron_api_power_monitor_mac.mm
```diff
@@ -82,6 +94,18 @@ - (void)onScreenUnlocked:(NSNotification*)notification {
...
+  for (auto* emitter : self->emitters) {
+    emitter->Emit("user-became-active");
```
nornagon, line 99:
```suggestion
    emitter->Emit("user-did-become-active");
```

shell/browser/api/electron_api_power_monitor_mac.mm
```diff
@@ -82,6 +94,18 @@ - (void)onScreenUnlocked:(NSNotification*)notification {
...
+  }
+}
+
+- (void)onUserBecameInactive:(NSNotification*)notification {
+  for (auto* emitter : self->emitters) {
+    emitter->Emit("user-became-inactive");
```
nornagon, line 105:
```suggestion
    emitter->Emit("user-did-resign-active");
```

nornagon (changes requested):

jkleinsc:

nornagon (approved):

codebytere:
Arm failure is:
```
app module app.requestSingleInstanceLock prevents the second launch of app
```

and unrelated (this is macOS only)

camhart:
Sorry for the noob question--but how do I know once this makes it released on npm?  And will the documentation (https://www.electronjs.org/docs/api/power-monitor) be updated at the same time?  (I can easily check the code changes out too if needed).

Thank you for this!!!!

PR #25324, opened 2020-09-04
fix: passing content-length down the stream for intercepted requests

marcin-prochownik:
#### Description of Change
Passing 'content-length' down the stream for intercepted request.

#### Checklist
- [x] `npm test` passes
- [x] tests are [changed or added]
- [x] [PR release notes]

#### Release Notes
Notes: Fixed an issue that file length is not available in 'will-download' event when file is downloaded thourgh intercepted request
Fixed an issue that videos are not played for custom protocols and buffered content.

### DIFF
--- a/shell/browser/net/electron_url_loader_factory.cc
+++ b/shell/browser/net/electron_url_loader_factory.cc
@@ -9,6 +9,7 @@
 #include <utility>
 
 #include "base/guid.h"
+#include "base/strings/string_number_conversions.h"
 #include "content/public/browser/browser_thread.h"
 #include "content/public/browser/storage_partition.h"
 #include "mojo/public/cpp/bindings/binding.h"
@@ -123,7 +124,6 @@ network::mojom::URLResponseHeadPtr ToResponseHead(
       } else {
         continue;
       }
-
       auto header_name_lowercase = base::ToLowerASCII(iter.first);
 
       if (header_name_lowercase == "content-type") {
@@ -131,8 +131,9 @@ network::mojom::URLResponseHeadPtr ToResponseHead(
         // in NetworkService.
         head->headers->GetMimeTypeAndCharset(&head->mime_type, &head->charset);
         has_content_type = true;
-      } else if (header_name_lowercase == "content-length") {
-        head->content_length = std::stol(iter.second.GetString());
+      } else if (header_name_lowercase == "content-length" &&
+                 iter.second.is_string()) {
+        base::StringToInt64(iter.second.GetString(), &head->content_length);
       }
     }
   }
--- a/shell/browser/net/electron_url_loader_factory.cc
+++ b/shell/browser/net/electron_url_loader_factory.cc
@@ -123,11 +123,16 @@ network::mojom::URLResponseHeadPtr ToResponseHead(
       } else {
         continue;
       }
-      // Some apps are passing content-type via headers, which is not accepted
-      // in NetworkService.
-      if (base::ToLowerASCII(iter.first) == "content-type") {
+
+      auto header_name_lowercase = base::ToLowerASCII(iter.first);
+
+      if (header_name_lowercase == "content-type") {
+        // Some apps are passing content-type via headers, which is not accepted
+        // in NetworkService.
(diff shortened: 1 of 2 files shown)


shell/browser/net/electron_url_loader_factory.cc
```diff
@@ -123,11 +123,16 @@ network::mojom::URLResponseHeadPtr ToResponseHead(
...
+
+      auto header_name_lowercase = base::ToLowerASCII(iter.first);
+
+      if (header_name_lowercase == "content-type") {
+        // Some apps are passing content-type via headers, which is not accepted
+        // in NetworkService.
         head->headers->GetMimeTypeAndCharset(&head->mime_type, &head->charset);
         has_content_type = true;
+      } else if (header_name_lowercase == "content-length") {
+        head->content_length = std::stol(iter.second.GetString());
```
zcbenz, line 135: Can you use the `base::StringToInt64` function in `base/strings/string_number_conversions.h` header?

```suggestion
        base::StringToInt64(iter.second.GetString(), &head->content_length);
```

Also we have to check `iter.second.is_string()` before calling `GetString()`.

shell/browser/net/electron_url_loader_factory.cc
```diff
@@ -123,11 +123,16 @@ network::mojom::URLResponseHeadPtr ToResponseHead(
       } else {
         continue;
       }
-      // Some apps are passing content-type via headers, which is not accepted
-      // in NetworkService.
-      if (base::ToLowerASCII(iter.first) == "content-type") {
+
```
zcbenz, line 126: This empty line can be removed.

zcbenz (changes requested):

shell/browser/net/electron_url_loader_factory.cc
marcin-prochownik: Thank you for your feedback. I applied the changes.

zcbenz (approved):

spec-main/api-protocol-spec.ts
```diff
@@ -919,8 +919,8 @@ describe('protocol module', () => {
       await protocol.unregisterProtocol('stream');
     });
 
-    it('does not successfully play videos with stream: false on streaming protocols', async () => {
-      await streamsResponses(standardScheme, 'error');
+    it('successfully plays videos with stream: false on streaming protocols', async () => {
+      await streamsResponses(standardScheme, 'play');
```
deepak1556, line 923: Seems like this fixes the root cause for https://github.com/electron/electron/pull/22955 and the stream option seems unnecessary now ?

/cc @pfrazee

deepak1556 (dismissed):
Confirming if stream option is necessary for custom protocol with this change.

spec-main/api-protocol-spec.ts
pfrazee: I'll need to look into this - the `stream` flag tells the video/audio player to use stream behaviors instead of fully-buffered behaviors, so what might have happened is this PR fixed the fully-buffered behavior but it still isn't streaming, which would mean playback requires a full download.

deepak1556: Thanks! that clarifies my question. Then the test case should be modified to rather denote streaming content when `stream: true` vs completely buffered content download when `stream: false`.

The current change to the test case doesn't differentiate this fix from https://github.com/electron/electron/pull/22955

nornagon:
The release notes here could be improved, I think. What does "down the stream" mean? Can you expand the notes a little so that someone reading them could understand whether this fix might be relevant to them?

spec-main/api-protocol-spec.ts
marcin-prochownik: I've updated the test names to indicate the difference, is that what you meant?

marcin-prochownik:
Our intention was to fix an issue that getTotalBytes on an item doesn't return file length in 'will-download' event, but it looks like this change also fixes issue that videos are not played when content is buffered for custom protocols. I'll put these two things in release notes.

zcbenz (approved):

PR #25328, opened 2020-09-04
fix: handle electron script errors better

MarshallOfSound:
There are some super rare occurrences where Electron JS can throw errors (we're not perfect either 😉 ).  This updates it so these errors don't impact things like renderer process creation negatively by catching, logging and absorbing them.

Notes: no-notes

### DIFF
--- a/shell/common/node_bindings.cc
+++ b/shell/common/node_bindings.cc
@@ -400,9 +400,24 @@ node::Environment* NodeBindings::CreateEnvironment(
   std::unique_ptr<const char*[]> c_argv = StringVectorToArgArray(args);
   isolate_data_ =
       node::CreateIsolateData(context->GetIsolate(), uv_loop_, platform);
-  node::Environment* env = node::CreateEnvironment(
-      isolate_data_, context, args.size(), c_argv.get(), 0, nullptr);
-  DCHECK(env);
+
+  node::Environment* env;
+  if (browser_env_ != BrowserEnvironment::BROWSER) {
+    v8::TryCatch try_catch(context->GetIsolate());
+    env = node::CreateEnvironment(isolate_data_, context, args.size(),
+                                  c_argv.get(), 0, nullptr);
+    DCHECK(env);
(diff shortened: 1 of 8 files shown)


shell/common/node_util.cc
```diff
@@ -17,15 +17,22 @@ v8::MaybeLocal<v8::Value> CompileAndCall(
     std::vector<v8::Local<v8::Value>>* arguments,
     node::Environment* optional_env) {
   v8::Isolate* isolate = context->GetIsolate();
+  v8::TryCatch try_catch(isolte);
```
nornagon, line 20:
```suggestion
  v8::TryCatch try_catch(isolate);
```

shell/common/node_bindings.cc
```diff
@@ -400,9 +400,24 @@ node::Environment* NodeBindings::CreateEnvironment(
...
+  if (browser_env_ != BrowserEnvironment::BROWSER) {
+    v8::TryCatch try_catch(context->GetIsolate());
+    env = node::CreateEnvironment(isolate_data_, context, args.size(),
+                                  c_argv.get(), 0, nullptr);
+    DCHECK(env);
+    // This will only be caught when something has gone terrible wrong as all
+    // electron scripts are wrapped in a try {} catch {} in run-compiler.js
+    if (try_catch.HasCaught()) {
+      LOG(ERROR) << "Failed to initialize node environment in process: "
+                 << process_type;
```
nornagon, line 414: Perhaps `DumpWithoutCrashing` to record these?

🧠  unrelated to this PR, but had the thought: what if we could enable UKM but allow the app to control the endpoint to which the data is sent? 🤔

nornagon (approved):

shell/common/node_bindings.cc
MarshallOfSound: I prefer making an event that users could beacon themselves instead of forcing it through the crash reporter 🤔

PR #25357, opened 2020-09-07
fix(extensions): define platform info to prevent renderer crash

samuelmaddock:
#### Description of Change

Fully implements `chrome.runtime.getPlatformInfo` such that the calling renderer won't crash/hang.

The implementation is copied from [chrome/browser/extensions/api/runtime/chrome_runtime_api_delegate.cc](https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/extensions/api/runtime/chrome_runtime_api_delegate.cc;l=285;drc=89ce2d1e6202423c9c0f0c83cec1f198cf55eded).

Ref #19447

cc @nornagon @sentialx 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed `chrome.runtime.getPlatformInfo` crashing the background process upon being invoked.

### DIFF
--- a/shell/browser/extensions/api/runtime/electron_runtime_api_delegate.cc
+++ b/shell/browser/extensions/api/runtime/electron_runtime_api_delegate.cc
@@ -7,6 +7,7 @@
 #include <string>
 
 #include "build/build_config.h"
+#include "components/update_client/update_query_params.h"
 #include "extensions/common/api/runtime.h"
 #include "shell/browser/extensions/electron_extension_system.h"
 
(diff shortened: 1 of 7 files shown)


spec-main/fixtures/extensions/chrome-runtime/main.js
```diff
@@ -1,6 +1,40 @@
 /* eslint-disable */
-document.documentElement.textContent = JSON.stringify({
-  manifest: chrome.runtime.getManifest(),
-  id: chrome.runtime.id,
-  url: chrome.runtime.getURL('main.js')
+
```
samuelmaddock, line 2: I converted this script to another pattern we use such that each API can be tested in isolation.

sentialx:
LGTM

nornagon:
Looks like you're missing a dependency declaration in the BUILD.gn file:

```
ERROR at //electron/shell/browser/extensions/api/runtime/electron_runtime_api_delegate.cc:10:11: Can't include this header from here.
#include "components/update_client/update_query_params.h"
          ^---------------------------------------------
The target:
  //electron:electron_lib
  //components/update_client:update_client
```

While it's a little questionable for us to be making a dependency on something called "update client" (given we don't use any of the actual functionality of that component), we're already transitively dependent on it via `//extensions/browser`, so adding it as an explicit dependency seems OK.

nornagon:
Also, I'm surprised this crashed before. Can you provide some more info about the crash this fixes?

samuelmaddock:
> 
> 
> Also, I'm surprised this crashed before. Can you provide some more info about the crash this fixes?

@nornagon what seems to happen is the API attempts to use `PlatformInfo` in a response, but doesn't handle it well when some of the fields aren't initialized. Specifically the call to [runtime::GetPlatformInfo::Results::Create(info)](https://source.chromium.org/chromium/chromium/src/+/master:extensions/browser/api/runtime/runtime_api.cc;l=723;drc=3003f448ef9a2d6407d1369d2aa8543f7a0a7c63). Then `runtime::ToString` is called on uninitialized data in a [PlatformInfo generated function.](https://source.chromium.org/chromium/chromium/src/+/master:out/Debug/gen/extensions/common/api/runtime.cc;l=225;drc=3003f448ef9a2d6407d1369d2aa8543f7a0a7c63)

shell/browser/extensions/api/runtime/electron_runtime_api_delegate.cc
```diff
@@ -42,10 +43,57 @@ bool ElectronRuntimeAPIDelegate::CheckForUpdates(
...
+  } else if (strcmp(nacl_arch, "x86-64") == 0) {
+    info->nacl_arch = extensions::api::runtime::PLATFORM_NACL_ARCH_X86_64;
+  } else if (strcmp(nacl_arch, "mips32") == 0) {
+    info->nacl_arch = extensions::api::runtime::PLATFORM_NACL_ARCH_MIPS;
+  } else if (strcmp(nacl_arch, "mips64") == 0) {
+    info->nacl_arch = extensions::api::runtime::PLATFORM_NACL_ARCH_MIPS64;
+  } else {
+    NOTREACHED();
+    return false;
+  }
```
nornagon, line 95: We never support NaCl. Can we change this to:
```suggestion
  info->nacl_arch = extensions::api::runtime::PLATFORM_NACL_ARCH_NONE;
```

spec-main/extensions-spec.ts
```diff
@@ -184,28 +184,38 @@ describe('chrome extensions', () => {
...
-        w.loadURL(url);
-        await emittedOnce(w.webContents, 'dom-ready');
-        content = JSON.parse(await w.webContents.executeJavaScript('document.documentElement.textContent'));
-        expect(content).to.be.an('object');
-      } finally {
-        w.destroy();
-      }
+      w = new BrowserWindow({ show: false, webPreferences: { session: customSession, nodeIntegration: true } });
+      w.loadURL(url);
+      await emittedOnce(w.webContents, 'dom-ready');
```
nornagon, line 199: do we need to wait for `dom-ready` here or can we just `await w.loadURL`?

spec-main/fixtures/extensions/chrome-runtime/main.js
nornagon: Nice change 👍

shell/browser/extensions/api/runtime/electron_runtime_api_delegate.cc
samuelmaddock: So the extensions code doesn't seem to like using this value. By doing so, the renderer crashes and devtools gets disconnected. 😞

I reverted the suggested changes, but removed unsupported platforms (mips, mips64).

spec-main/extensions-spec.ts
samuelmaddock: Nope. I've updated extension tests to only await `dom-ready` when necessary.

gittysachin (commented):
So far, LGTM

shell/browser/extensions/api/runtime/electron_runtime_api_delegate.cc
nornagon: hm, ok. /shrug then

nornagon (approved):

PR #25373, opened 2020-09-08
feat: honor `nativeTheme.themeSource = 'dark'` before creating BrowserWindow on Windows

### LINKED ISSUE #23479: [Windows] Title bar does not respect dark mode
### Preflight Checklist

### Issue Details
When setting nativeTheme.themeSource to 'dark' (or 'system' and setting the system theme to 'dark'), the title bar on Windows remains light. On macOS, the title bar accurately reflects the setting.

* **Electron Version:**
  * `npm ./node_modules/.bin/electron --version`: 6.9.0
  * package.json: `"electron": "^8.2.5",`
  * yarn.lock: 8.2.5
* **Operating System:**
  * Windows 10 (1903)

### Expected Behavior

(This works as expected on macOS)
|             | nativeTheme.themeSource = 'system' | nativeTheme.themeSource = 'light' | nativeTheme.themeSource = 'dark' |
|-------------|------------------------------------|-----------------------------------|----------------------------------|
| Light Theme | Light Title Bar                    | Light Title Bar                   | Dark Title Bar                   |
| Dark Theme  | Dark Title Bar                     | Light Title Bar                   | Dark Title Bar                   |

### Actual Behavior

|  | nativeTheme.themeSource = 'system' | nativeTheme.themeSource = 'light' | nativeTheme.themeSource = 'dark' |
|-------------|------------------------------------|-----------------------------------|----------------------------------|
| Light Theme | Light Title Bar | Light Title Bar | Light Title Bar |
| Dark Theme | Light Title Bar | Light Title Bar | Light Title Bar |

### To Reproduce

```js
nativeTheme.themeSource = 'dark'
```

### Screenshots

Command Prompt:
Command Prompt

File Explorer:
File Explorer

Title Bar in Electron with `nativeTheme.themeSource = 'dark'`
Windows Title Bar

ckerr:
#### Description of Change

Fixes #23479.

With this patch, setting `nativeTheme.themeSource = 'dark'` or `'light'` *before* creating the window is honored and the window will be rendered with the correct colors when shown.

As a side-effect of this change, the setting themeSource dynamically now works on Linux too.

There is still a wart: Changing themeSource dynamically when the window is already shown isn't perfect: the property is set, but the window still has to be poked before the change takes effect, e.g. changing focus or minimizing/restoring the window. The usual programmatic ways of forcing a window refresh don't seem to work here; perhaps this is my error but let's see about getting guidance for the right recipe to use.

This PR adds a lot of scaffolding to pull in the undocumented Win32 dark mode APIs. Much of this ground is covered by https://github.com/ysc3839/win32-darkmode, which is license-compatible (MIT).

TODO:
- [x] Can the titlebar issue described above be fixed? 
- [x] `win\dark_mode.cc` is really big. Audit it to see if any can be removed

CC @MarshallOfSound  @electron/wg-releases 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Improved dark mode support on Windows.

### DIFF
--- a/buildflags/BUILD.gn
+++ b/buildflags/BUILD.gn
@@ -20,7 +20,7 @@ buildflag_header("buildflags") {
     "ENABLE_ELECTRON_EXTENSIONS=$enable_electron_extensions",
     "ENABLE_BUILTIN_SPELLCHECKER=$enable_builtin_spellchecker",
     "ENABLE_PICTURE_IN_PICTURE=$enable_picture_in_picture",
-    "ENABLE_WIN_DARK_MODE=$enable_win_dark_mode",
+    "ENABLE_WIN_DARK_MODE_WINDOW_UI=$enable_win_dark_mode_window_ui",
     "OVERRIDE_LOCATION_PROVIDER=$enable_fake_location_provider",
   ]
 }
--- a/buildflags/BUILD.gn
+++ b/buildflags/BUILD.gn
@@ -20,6 +20,7 @@ buildflag_header("buildflags") {
     "ENABLE_ELECTRON_EXTENSIONS=$enable_electron_extensions",
     "ENABLE_BUILTIN_SPELLCHECKER=$enable_builtin_spellchecker",
     "ENABLE_PICTURE_IN_PICTURE=$enable_picture_in_picture",
+    "ENABLE_WIN_DARK_MODE=$enable_win_dark_mode",
     "OVERRIDE_LOCATION_PROVIDER=$enable_fake_location_provider",
   ]
 }
--- a/buildflags/buildflags.gni
+++ b/buildflags/buildflags.gni
@@ -35,5 +35,5 @@ declare_args() {
   enable_builtin_spellchecker = true
 
   # Undocumented Windows dark mode API
-  enable_win_dark_mode_window_ui = true
+  enable_win_dark_mode_window_ui = false
 }
--- a/buildflags/buildflags.gni
+++ b/buildflags/buildflags.gni
@@ -35,5 +35,5 @@ declare_args() {
   enable_builtin_spellchecker = true
 
   # Undocumented Windows dark mode API
-  enable_win_dark_mode = true
+  enable_win_dark_mode_window_ui = true
 }
--- a/buildflags/buildflags.gni
+++ b/buildflags/buildflags.gni
@@ -33,4 +33,7 @@ declare_args() {
 
   # Enable Spellchecker support
   enable_builtin_spellchecker = true
+
+  # Undocumented Windows dark mode API
+  enable_win_dark_mode = true
 }
--- a/shell/browser/win/dark_mode.cc
+++ b/shell/browser/win/dark_mode.cc
@@ -1,0 +1,435 @@
+// Copyright (c) 2020 Microsoft Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE-CHROMIUM file.
+
+#include "shell/browser/win/dark_mode.h"
+
+// #include <climits>
+// #include <cstdint>
+#include <mutex>
+#include <thread>
+
+// Including SDKDDKVer.h defines the highest available Windows platform.
+// If you wish to build your application for a previous Windows platform,
+// include WinSDKVer.h and set the _WIN32_WINNT macro to the platform you wish
+// to support before including SDKDDKVer.h.
+#include <SDKDDKVer.h>
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#include <CommCtrl.h>
+#include <Uxtheme.h>
+#include <Vssym32.h>
+#include <Windows.h>
+#include <WindowsX.h>
+
+#pragma comment(lib, "Comctl32.lib")
+#pragma comment(lib, "Uxtheme.lib")
+
+// This namespace contains code from
+// https://github.com/stevemk14ebr/PolyHook_2_0/blob/master/sources/IatHook.cpp
+// which is licensed under the MIT License.
+// See PolyHook_2_0-LICENSE for more information.
+namespace {
+
+template <typename T, typename T1, typename T2>
+constexpr T RVA2VA(T1 base, T2 rva) {
+  return reinterpret_cast<T>(reinterpret_cast<ULONG_PTR>(base) + rva);
+}
+
+template <typename T>
+constexpr T DataDirectoryFromModuleBase(void* moduleBase, size_t entryID) {
+  auto* dosHdr = reinterpret_cast<PIMAGE_DOS_HEADER>(moduleBase);
+  auto* ntHdr = RVA2VA<PIMAGE_NT_HEADERS>(moduleBase, dosHdr->e_lfanew);
+  auto* dataDir = ntHdr->OptionalHeader.DataDirectory;
+  return RVA2VA<T>(moduleBase, dataDir[entryID].VirtualAddress);
+}
+
+#if 0
+PIMAGE_THUNK_DATA FindAddressByName(void *moduleBase, PIMAGE_THUNK_DATA impName, PIMAGE_THUNK_DATA impAddr, const char *funcName)
+{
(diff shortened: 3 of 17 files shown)


MarshallOfSound (changes requested):
This is a tricky one because it "fixes" a thing that I would argue isn't expected behavior.  On Windows there are maybe 3 apps that do this and all of them are windows built-in.  The private API this utilizes to do this is super hacky, apparently depends on exact versions / builds of Windows and in theory could break at any time.

If there isn't a public API for this maybe we should ask for one? It doesn't feel right adding this amount of private API complexity to do a thing that by most definitions isn't expected behavior (and speaking for one app in particular we'd want to opt out of)

ckerr:
@MarshallOfSound it looks like we'll be using undocumented API if we want this to work in the short term. I agree this is a lot of code for a small amount of functionality -- but at the same time, I think having this feature work is still worthwhile. What if I shave down all the code to only the minimum subset that we need, and then we see how large that patch is?

MarshallOfSound:
@ckerr I'm _really_ not a fan of this much private API usage.  If this were to land against my recommendations the minimum requirements would be:

* We've done our damndest to reduce the private API usage by as much as possible, the smaller the surface area the better
* The code should at least be vaguely readable, it's almost there atm but it's still a lot of `_` and magic enums
* This feature should 100% be disabled by default (opt-in) and no code path should call these APIs unless the individual app has opted in

samuelmaddock:
If this feature were to land, would it allow apps to forcefully apply dark mode? As opposed to applying it only when the user has chosen that OS theme.

Aside from the issue of private API usage, this would be a nice addition for apps which are primarily dark in aesthetics, but aren't using a frameless window.

ckerr:
> * We've done our damndest to reduce the private API usage by as much as possible, the smaller the surface area the better

I've removed everything not necessary to fix the titlebar issue reported in #23479. This revised patch only loads enough API to determine how to see if dark mode is supported, if it's allowed, and how to set it for newly-created windows. dark_mode.cc is now about 60% smaller despite having new comments to address the next point --

> * The code should at least be vaguely readable, it's almost there atm but it's still a lot of `_` and magic enums

I've tried to eliminate confusing code where possible, added explanatory comments, limited variables' scope, and moved variables to be declared just before use. Looking up function pointers from the dlls is still verbose -- you have to declare the function type, then declare a variable, and then do the lookup -- but I've added some template helper code to make it more readable.

> * This feature should 100% be disabled by default (opt-in) and no code path should call these APIs unless the individual app has opted in

I'm not  sure this is a good idea -- it doesn't seem productive to require users to opt-in to a bugfix? This fixes an existing code in our dark mode:

```js
nativeTheme.themeSource = 'dark';
new BrowserWindow(...)
// windows' titlebar will not be dark

**Comment by ckerr** [comment]
> If this feature were to land, would it allow apps to forcefully apply dark mode? As opposed to applying it only when the user has chosen that OS theme.

Yes it would. The use case is literally setting `nativeTheme.themeSource = 'dark';` before creating a window. :)

File: buildflags/buildflags.gni [hunk 1]
```diff
@@ -33,4 +33,7 @@ declare_args() {
 
   # Enable Spellchecker support
   enable_builtin_spellchecker = true
+
+  # Undocumented Windows dark mode API
+  enable_win_dark_mode = true
```
MarshallOfSound, line 38: Is this supposed to be enabled by default?

buildflags/BUILD.gn
```diff
@@ -20,6 +20,7 @@ buildflag_header("buildflags") {
     "ENABLE_ELECTRON_EXTENSIONS=$enable_electron_extensions",
     "ENABLE_BUILTIN_SPELLCHECKER=$enable_builtin_spellchecker",
     "ENABLE_PICTURE_IN_PICTURE=$enable_picture_in_picture",
+    "ENABLE_WIN_DARK_MODE=$enable_win_dark_mode",
```
MarshallOfSound, line 23: Can we call it something more specific to this feature.  i.e. `ENABLE_WIN_DARK_MODE_WINDOW_UI` or something

ckerr: Renamed in 42214684722c2aad1548ee539aa2bf65046d009a

buildflags/buildflags.gni
ckerr: Disbled by default in 8508c62af5a9015dfebc9d38d68486a763b70aa6

typings/internal-ambient.d.ts
```diff
@@ -24,6 +24,7 @@ declare namespace NodeJS {
     isPictureInPictureEnabled(): boolean;
     isExtensionsEnabled(): boolean;
     isComponentBuild(): boolean;
+    isWinDarkModeEnabled(): boolean;
```
MarshallOfSound, line 27:
```suggestion
    isWinDarkModeWindowUiEnabled(): boolean;
```

MarshallOfSound (approved):
One missed replacement, other than that :shipit:

typings/internal-ambient.d.ts
ckerr: Fixed 2d54188d5c8e15a9d9474cff9d23cb10679c6fe1

shell/browser/win/dark_mode.cc
```diff
@@ -0,0 +1,197 @@
...
+  return (buildNumber == 17763 ||  // 1809
+          buildNumber == 18362 ||  // 1903
+          buildNumber == 18363 ||  // 1909
+          buildNumber == 19041);   // 2004
+}
+
+void InitDarkMode() {
+  HMODULE hDwmApi = LoadLibrary(L"DWMAPI");
+  if (hDwmApi != 0) {
+    Symbol(hDwmApi, &_DwmSetWindowAttribute, "DwmSetWindowAttribute");
```
deepak1556, line 87: Can you use `base::win::PEImage` instead of defining our own symbol import utility.

deepak1556 (changes requested):

shell/browser/win/dark_mode.cc
```diff
@@ -0,0 +1,197 @@
+// Copyright (c) 2020 Microsoft Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE-CHROMIUM file.
+
+#include "shell/browser/win/dark_mode.h"
+
+#pragma comment(lib, "Uxtheme.lib")
```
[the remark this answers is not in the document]
ckerr: moved to deps in da3108c868907c7c3daff5431aba6c984b9cdcbe

shell/browser/win/dark_mode.cc
ckerr: refactored to use `base/` in 46ec6e8a50207a8c62a8e329d4e3131168c2095d

shell/browser/win/dark_mode.cc
```diff
@@ -0,0 +1,195 @@
...
+      nt_pei.GetProcAddress("RtlGetNtVersionNumbers"));
+  if (!RtlGetNtVersionNumbers) {
+    return;
+  }
+
+  // check to see if the major/minor/build support these dark mode APIS
+  DWORD major = 0, minor = 0;
+  RtlGetNtVersionNumbers(&major, &minor, &g_buildNumber);
+  g_buildNumber &= ~0xF0000000;
+  if (major != 10 || minor != 0 || !CheckBuildNumber(g_buildNumber)) {
```
deepak1556, line 86: Can you use `base::win::GetVersion` instead, this seems like you will have to update `CheckBuildNumber` for every OS release.

ckerr: Yes for certain we can use `base::win::OSInfo` instead, that's a great idea.

But... yes, absolutely there will be upkeep on this because there is still no documented API for this and it could all change in a future version of Windows. See for example how ordinal 135 in uxtheme changed from `AllowDarkModeForApp()` to `SetPreferredAppMode()` in version 1903. It's better for dark mode to just stop working in newer versions than it is to potentially call an unknown function signature.

Until there is supported stable API for these calls, I don't see a safe alternative.

ckerr: Migrated build/version checks to `base::win::OSInfo` in b1d76af6d4f386c06e871bbe27e6af3c004b2f68

deepak1556 (approved):

gpetrov:
Can we have this in production now without the flag?

PR #25377, opened 2020-09-08
feat: allow path override with --ignore-scripts

skeggse:
If you --ignore-scripts when installing electron currently, it'll fail to write the path.txt file and thus fail to use the override dist path. Open to other solutions - just hoping to be able to use a prebuilt electron binary with the default package without having to muck around with it installing an unused version.

#### Description of Change

Default to `electron` if the `ELECTRON_OVERRIDE_DIST_PATH` environment variable is present.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/npm/index.js
+++ b/npm/index.js
@@ -4,11 +4,14 @@ const path = require('path');
 const pathFile = path.join(__dirname, 'path.txt');
 
 function getElectronPath () {
+  let executablePath;
   if (fs.existsSync(pathFile)) {
-    const executablePath = fs.readFileSync(pathFile, 'utf-8');
-    if (process.env.ELECTRON_OVERRIDE_DIST_PATH) {
-      return path.join(process.env.ELECTRON_OVERRIDE_DIST_PATH, executablePath);
-    }
+    executablePath = fs.readFileSync(pathFile, 'utf-8');
+  }
+  if (process.env.ELECTRON_OVERRIDE_DIST_PATH) {
+    return path.join(process.env.ELECTRON_OVERRIDE_DIST_PATH, executablePath || 'electron');
+  }
+  if (executablePath) {
     return path.join(__dirname, 'dist', executablePath);
   } else {
     throw new Error('Electron failed to install correctly, please delete node_modules/electron and try installing again');


npm/index.js
```diff
@@ -6,9 +6,11 @@ var pathFile = path.join(__dirname, 'path.txt')
 function getElectronPath () {
   if (fs.existsSync(pathFile)) {
     var executablePath = fs.readFileSync(pathFile, 'utf-8')
-    if (process.env.ELECTRON_OVERRIDE_DIST_PATH) {
-      return path.join(process.env.ELECTRON_OVERRIDE_DIST_PATH, executablePath)
-    }
+  }
+  if (process.env.ELECTRON_OVERRIDE_DIST_PATH) {
```
MarshallOfSound, line 10:
```js
  if (process.env.ELECTRON_OVERRIDE_DIST_PATH) {
    return path.join(process.env.ELECTRON_OVERRIDE_DIST_PATH, executablePath || 'electron')
  }
  if (fs.existsSync(pathFile)) {
    var executablePath = fs.readFileSync(pathFile, 'utf-8')
    return path.join(__dirname, 'dist', executablePath)
  } else {
      throw ...
```

I think that layout of if statements is more readable

skeggse: Admittedly I haven't given much thought to the structure of this implementation. That said, I think we'd be referencing `executablePath` before it's been initialized with that structure - yeah?

MarshallOfSound: Ah yeah, we should just not depend on `path.txt` for resolution.  Use `path.txt` to ensure it's installed but export `getPlatformPath` from `install.js` and use that to get the executable path.

codebytere: @skeggse do you still plan to pursue this PR?

Fercafig (approved):

npm/index.js
skeggse: It's a feature that no longer impacts me; I would be fine with finishing this, though I guess I don't follow @MarshallOfSound's last comment (maybe I did earlier?)

If I recall my use-case correctly, `path.txt` wouldn't exist because the install script would have been skipped - so checking for it when there's an override doesn't make sense to me. I'm also not clear on what `getPlatformPath` adds - I would expect the `ELECTRON_OVERRIDE_DIST_PATH` variable to be the path containing the binary, but that seems at odds with `getPlatformPath` returning a macos-app-specific string - wouldn't you expect `ELECTRON_OVERRIDE_DIST_PATH=/Applications/Electron.app/Contents/MacOS/Electron` to work?

zcbenz: This makes sense to me, I'm good with current implementation.

npm/index.js
```diff
@@ -6,9 +6,11 @@ var pathFile = path.join(__dirname, 'path.txt')
 function getElectronPath () {
   if (fs.existsSync(pathFile)) {
     var executablePath = fs.readFileSync(pathFile, 'utf-8')
```
zcbenz, line 8: `executablePath` should be moved out of the scope since it is used by other scopes, we should not rely on JS"s weird rules for undefined variables.

```js
var executablePath
if (...) {
  executablePath = ... 
}
```

skeggse: Done, and rebased.

zcbenz (approved):

skeggse:
:shipit: ?

codebytere (approved):

MarshallOfSound (approved):

PR #25385, opened 2020-09-09
feat(extensions): expose ExtensionRegistryObserver events in Session

samuelmaddock:
#### Description of Change
Extensions can be loaded or unloaded for various reasons. In some cases, this can occur by no means of the Electron programmer, such as in the case of [`chrome.runtime.reload()`](https://developer.chrome.com/extensions/runtime#method-reload).

In order to be able to manage state about extensions outside of Electron's APIs, events related to loading and unloaded extensions are needed. The only alternative would be to monkey-patch the `Session` prototype.

The use case I have is to manage state in my own implementation of [`chrome.browserAction`](https://developer.chrome.com/extensions/browserAction). I need to know when an extension is added or removed to set or delete saved state.

Ref #19447

cc @nornagon @sentialx 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Add `extension-loaded`, `extension-unloaded`, and `extension-ready` events to `Session`.

### DIFF
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -102,12 +102,9 @@ Emitted after an extension is loaded. This occurs whenever an extension is
 added to the "enabled" set of extensions. This includes:
 - Extensions being loaded from `Session.loadExtension`.
 - Extensions being reloaded:
+   * from a crash.
+   * from a disabled state (if the extension is disabled then enabled).
    * if the extension requested it ([`chrome.runtime.reload()`](https://developer.chrome.com/extensions/runtime#method-reload)).
-
-<!--
-Refer to //extensions/browser/extension_registry_observer.h for further
-explaination.
--->
 
 #### Event: 'extension-unloaded'
 
@@ -116,7 +113,9 @@ Returns:
 * `event` Event
 * `extension` [Extension](structures/extension.md)
 
-Emitted after an extension is unloaded.
+Emitted after an extension is unloaded. The extension no longer exists in
+the group of enabled extensions, but it can still be a member of another
+group, like disabled, blocklisted, or terminated.
 
 #### Event: 'extension-ready'
 
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -91,6 +91,43 @@ session.defaultSession.on('will-download', (event, item, webContents) => {
 })
 ```
 
+#### Event: 'extension-loaded'
+
+Returns:
+
+* `event` Event
+* `extension` [Extension](structures/extension.md)
+
+Emitted after an extension is loaded. This occurs whenever an extension is
+added to the "enabled" set of extensions. This includes:
+- Extensions being loaded from `Session.loadExtension`.
+- Extensions being reloaded:
+   * if the extension requested it ([`chrome.runtime.reload()`](https://developer.chrome.com/extensions/runtime#method-reload)).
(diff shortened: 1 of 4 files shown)


samuelmaddock:
Pinging @electron/wg-api since I think this will require their review.

docs/api/session.md
```diff
@@ -91,6 +91,43 @@ session.defaultSession.on('will-download', (event, item, webContents) => {
...
+Emitted after an extension is loaded. This occurs whenever an extension is
+added to the "enabled" set of extensions. This includes:
+- Extensions being loaded from `Session.loadExtension`.
+- Extensions being reloaded:
+   * if the extension requested it ([`chrome.runtime.reload()`](https://developer.chrome.com/extensions/runtime#method-reload)).
+
+<!--
+Refer to //extensions/browser/extension_registry_observer.h for further
+explaination.
+-->
```
nornagon, line 110: Perhaps include this in the visible part of the documentation?

Or, better, fully enumerate the situations that can cause this event in Electron.

docs/api/session.md
```diff
@@ -91,6 +91,43 @@ session.defaultSession.on('will-download', (event, item, webContents) => {
...
+
+#### Event: 'extension-unloaded'
+
+Returns:
+
+* `event` Event
+* `extension` [Extension](structures/extension.md)
+
+Emitted after an extension is unloaded.
```
nornagon, line 119: When does this happen in Electron?

shell/browser/api/electron_api_session.cc
```diff
@@ -294,6 +298,10 @@ Session::~Session() {
     service->SetHunspellObserver(nullptr);
   }
 #endif
+
+#if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
+  extensions::ExtensionRegistry::Get(browser_context())->RemoveObserver(this);
```
nornagon, line 303: I think this is handled by `ScopedObserver`, yes?

Given we're only ever only ever observing a single source, perhaps the scoped observer isn't needed?

nornagon (changes requested):
lgtm, just a few minor comments

shell/browser/api/electron_api_session.cc
samuelmaddock: The `ScopedObserver` wasn't really working how I thought it did so I just removed it. It would need to be passed the `browser_context()` which the template doesn't support.

docs/api/session.md
samuelmaddock: I've updated it to include more info, but the parts about enabled/disabled extensions doesn't make a lot of sense until #25386 is merged. However, enabled/disabled extensions still are technically part of the extension system implementation regardless.

docs/api/session.md
samuelmaddock: Updated to mention `Session.removeExtension`

docs/api/session.md
samuelmaddock: I'll hold off on mentioning the concept of enabled/disabled extensions for now. I'll plan to include them in the mentioned PR.

nornagon (approved):

jkleinsc:

jkleinsc (approved):

Issue #24977, opened 2020-08-14
Ability to disable pdf viewer


t57ser:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description
Since electron 9, pdf viewer has been implemented but does not fully work:
https://github.com/electron/electron/issues/22178
https://github.com/electron/electron/issues/24458
https://github.com/electron/electron/issues/22180
There should be an option to disable this feature so that it is possible to upgrade to a newer electron version and keep old working behavior until the pdf viewer gets fully fixed.

### Proposed Solution
Add a flag to disable the pdf viewer

### Alternatives Considered
I have checked if there is a chrome command line flag to handle this but could not find anything

leonheess:
Could be closed, right? All issues have been fixed.

Jelmerro:
I would still like an option to disable it because it's still broken in others ways: it claims the user keyboard input away from any listeners, doesn't allow custom listeners or shortcuts, and there is no easy option to download pdfs instead of using the (broken) viewer. Please add the option to disable it regardless of how well it works (or not), because some users just don't want it to be in Electron at all.

codebytere:
Hi, while we appreciate this enhancement request it's not something that our team will be able to implement in the near future, and since we use our issues tab as the team's backlog, we prefer to only retain issues that are on our immediate roadmap. If you'd like to make a PR for this request we would welcome that but in the meantime I'm going to go ahead and close this.

PR #25386, opened 2020-09-09
[wip] feat(extensions): add enable/disable extension methods to Session

samuelmaddock:
#### Description of Change

Allows the user to toggle an extension at runtime.

This is useful in cases where an extension should be loaded, but the user can decide whether or not it should run (ie. Chrome's extension toggle preferences).

TODO:
- [ ] Figure out whether to add `Session.isExtensionEnabled(boolean)` or modify `Session.getAllExtensions()` to include enabled state.
- [x] ~~Check if this can help solve #24977 (I did notice that the PDF extension shows up in an `extension-ready` event added in #25385).~~ PDF extension gets disabled, but it simply fails to load the plugin ([screenshot](https://user-images.githubusercontent.com/1656324/92865580-7f217b00-f3cc-11ea-817e-8f3bb0057c8c.png)).
- [ ] Verify that the modified `ExtensionPrefs` values are only in-memory.

Ref #19447

cc @sentialx - Let me know if you have any input on the direction here for these APIs.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added `Session.enableExtension(extensionId)` and `Session.disableExtension(extensionId)`.

### DIFF
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -615,6 +615,18 @@ Unloads an extension.
 **Note:** This API cannot be called before the `ready` event of the `app` module
 is emitted.
 
+#### `ses.enableExtension(extensionId)`
+
+* `extensionId` String - ID of extension to enable
+
+Enables an extension. If the extension is already enabled, does nothing.
+
+#### `ses.disableExtension(extensionId)`
+
+* `extensionId` String - ID of extension to disable
+
+Disables an extension. If the extension is already disabled, does nothing.
+
 #### `ses.getExtension(extensionId)`
 
 * `extensionId` String - ID of extension to query
--- a/shell/browser/api/electron_api_session.cc
+++ b/shell/browser/api/electron_api_session.cc
@@ -727,6 +727,18 @@ void Session::RemoveExtension(const std::string& extension_id) {
   extension_system->RemoveExtension(extension_id);
 }
 
+void Session::EnableExtension(const std::string& extension_id) {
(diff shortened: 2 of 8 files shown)


sentialx:
@samuelmaddock I think it would be convenient to include `enabled` property into the `Extension` object instead of making a separate method for that. Also, would be nice to have `extension-enabled` and `extension-disabled` events, unless `extension-loaded` and `extension-unloaded` do the same thing.

Other than that, I think we should consider making an `extensions` namespace to avoid `extension` word in every event and method.
i.e. `session.extensions.load(path)`
What are your thoughts on this?

samuelmaddock:
> @samuelmaddock I think it would be convenient to include enabled property into the Extension object instead of making a separate method for that. Also, would be nice to have extension-enabled and extension-disabled events, unless extension-loaded and extension-unloaded do the same thing.

I thought about adding `Extension.enabled` initially, but it would make directly converting from an `extensions::Extension` more complex. Additionally, it would need to be a fixed value which is set when serialized. That could cause confusion for folks thinking it would reflect its real-time value.

I'm not totally against the idea, I'm just not sure how best to implement it.

> Other than that, I think we should consider making an `extensions` namespace to avoid `extension` word in every event and method.
> i.e. `session.extensions.load(path)`
> What are your thoughts on this?

I think that would be a clean approach. I created an issue to track this idea: #25418

samuelmaddock:
Closing this for now. I'll re-open in the future if this becomes relevant again.

PR #25393, opened 2020-09-09
fix: access violation during redirects (intercepted requests)

marcin-prochownik:
#### Description of Change
Fixed intermittent access violation during redirects (intercepted requests)

#### Checklist
- [ ] `npm test` passes
- [x] [PR release notes

#### Release Notes
Notes: Fixed intermittent access violation during redirects (intercepted requests)

### DIFF
--- a/shell/browser/net/electron_url_loader_factory.cc
+++ b/shell/browser/net/electron_url_loader_factory.cc
@@ -22,7 +22,6 @@
 #include "shell/browser/electron_browser_context.h"
 #include "shell/browser/net/asar/asar_url_loader.h"
 #include "shell/browser/net/node_stream_loader.h"
-#include "shell/browser/net/proxying_url_loader_factory.h"
 #include "shell/browser/net/url_pipe_loader.h"
 #include "shell/common/electron_constants.h"
 #include "shell/common/gin_converters/file_path_converter.h"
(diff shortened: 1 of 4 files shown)


deepak1556:
Can you add some test to demonstrate the issue. Thanks!

marcin-prochownik:
It would be hard to write a test. We bumped into this issue while stress testing our electron based application. We ran multiple instances of an app and they were navigating in a loop to different websites. From time to time, one of the instances crash. It seemed not to crash if single instance of the app was running, but it could be just we were not patient enough.

Anyway, while debugging the issue, we noticed that this line of code:
https://github.com/electron/electron/blob/master/shell/browser/net/electron_url_loader_factory.cc#L272

is causing access violation exception. Electron tries to call CreateLoaderAndStart to follow redirect, but proxy_factory is already disposed.
The fix is, I  keep pending requests in a collection (for intercepted requests) to make ProxyingURLLoaderFactory aware that requests are pending and it is not safe to dispose itself. Very similar logic already exists for non-intercepted requests.

shell/browser/net/electron_url_loader_factory.cc
```diff
@@ -312,17 +325,20 @@ void ElectronURLLoaderFactory::StartLoading(
...
+                   proxy_factory);
         return;
       }
       StartLoading(std::move(loader), routing_id, request_id, options, request,
                    std::move(client), traffic_annotation, proxy_factory, type,
                    args);
       break;
   }
+
+  if (type != ProtocolType::kFree && proxy_factory) {
```
zcbenz, line 339: It is not very straightforward why the `type != ProtocolType::kFree` check is needed, how about doing `return` instead of `break` in `case ProtocolType::kFree`? It should make the code more easy to understand.

zcbenz (approved):
I'm good with this PR since it aligns with `extensions::WebRequestProxyingURLLoaderFactory`'s cleanup code.

I don't have an idea of how to test this, and I'm good without a test.

codebytere (approved):

deepak1556:
@marcin-prochownik thanks for the context!

@zcbenz some more thinking, wouldn't it be better to clone the proxy loader factory when redirect happens since essentially we want to restart the loader, that way we don't have to maintain a map of intercepted requests but rather check for the existence of `proxy_receivers_`

In https://github.com/electron/electron/blob/9f4a097e039aef04e3df4ee1f24ce3b817091b5d/shell/browser/net/electron_url_loader_factory.cc#L272

we change it to

```
mojo::Remote<network::mojom::URLLoaderFactory> loader_remote;
if (proxy_factory)
  proxy_factory->Clone(loader_remote.BindNewPipeAndPassReceiver());
loder_remote->CreateLoaderAndStart
```

In https://github.com/electron/electron/blob/9f4a097e039aef04e3df4ee1f24ce3b817091b5d/shell/browser/net/proxying_url_loader_factory.cc#L918

we add

```
if (target_factory_.is_bound() || !requests_.empty() || !proxy_receivers_.empty())
```

zcbenz:
@deepak1556 That sounds a very good idea to try, current implementation for redirects is not really ideal (you can see from its long comments).

marcin-prochownik:
@deepak1556 your solution seems better, it is more generic and maybe could be reused later on. I have one question though. If 'proxy_factory' is disposed, is it safe to call 'Clone' on it? Wouldn't that lead to access violation?

deepak1556:
You can clone before calling `ElectronURLLoaderFactory::StartLoading` in ProxyingURLLoaderFactory and only pass down the remote end to `ElectronURLLoaderFactory`

marcin-prochownik:
Done. Had some issues with taming mojo, but I hope I made it right.

PR #25401, opened 2020-09-10
feat: custom toast xml and failure reporting for notifications

bitdisaster:
#### Description of Change
This PR introduces two new features for Windows Toast Notification.
#### **Custom Toast XML**
Windows notifications allow many layouts and different activation methods. Exposing all possibilities through properties on the NotificationOption object would be extensive as well as the different code paths to achieve an arbitrary layout/design of a notification. Instead the options object can now take valid [Toast XML](https://docs.microsoft.com/en-us/uwp/schemas/tiles/toastschema/schema-root)  This also allows using protocol activation and will work with persisted notifications in Action Center. Something foreground activation can not do without a COM server implementation.
image  image

#### **Notification Failure event**
Unfortunately, Windows Notifications are not famous for reliability. A failure event on the Notification object now informs of a failed notification with a text hinting the location in the code path and a HRESULT error code coming from the Windows API call. 
` myNotification.on('failed', (e, err) => {
      console.log('C'mon Windows. Don't be like that.  ' + err);
    });`

#### Checklist

- [X] `npm test` passes

#### Release Notes

Notes:
* Implemented custom Toast XML notifications on Windows 
* Added failure reporting of notifications on Window

### DIFF
--- a/docs/api/notification.md
+++ b/docs/api/notification.md
@@ -104,8 +104,6 @@ Returns:
 
 Emitted when an error is encountered while creating and showing the native notification.
 
-#### Event: 'action' _macOS_
-
 ### Instance Methods
 
 Objects created with `new Notification` have the following instance methods:
@@ -174,7 +172,7 @@ If `timeoutType` is set to 'never', the notification never expires. It stays ope
 
 A [`NotificationAction[]`](structures/notification-action.md) property representing the actions of the notification.
 
-#### `notification.toastXml`
+#### `notification.toastXml` _Windows_
 
 A `String` property representing the custom Toast XML of the notification.
 
--- a/docs/api/notification.md
+++ b/docs/api/notification.md
@@ -29,9 +29,9 @@ Returns `Boolean` - Whether or not desktop notifications are supported on the cu
 ### `new Notification([options])`
 
 * `options` Object (optional)
-  * `title` String - A title for the notification, which will be shown at the top of the notification window when it is shown.
+  * `title` String (optional) - A title for the notification, which will be shown at the top of the notification window when it is shown.
   * `subtitle` String (optional) _macOS_ - A subtitle for the notification, which will be displayed below the title.
-  * `body` String - The body text of the notification, which will be displayed below the title or subtitle.
+  * `body` String (optional) - The body text of the notification, which will be displayed below the title or subtitle.
   * `silent` Boolean (optional) - Whether or not to emit an OS notification noise when showing the notification.
   * `icon` (String | [NativeImage](native-image.md)) (optional) - An icon to use in the notification.
   * `hasReply` Boolean (optional) _macOS_ - Whether or not to add an inline reply option to the notification.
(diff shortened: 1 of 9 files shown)


docs/api/notification.md
```diff
@@ -94,6 +95,17 @@ Returns:
...
+#### Event: 'failed' _Windows_
+
+Returns:
+
+* `event` Event
+* `error` String - The error encountered during execution of the `show()` method.
+
+Emitted when an error is encountered while creating and showing the native notification.
+
+#### Event: 'action' _macOS_
```
MarshallOfSound, line 107: ?

docs/api/notification.md
```diff
@@ -162,6 +174,10 @@ If `timeoutType` is set to 'never', the notification never expires. It stays ope
 
 A [`NotificationAction[]`](structures/notification-action.md) property representing the actions of the notification.
 
+#### `notification.toastXml`
```
MarshallOfSound, line 177:
```suggestion
#### `notification.toastXml` _Windows_
```

shell/browser/notifications/win/windows_toast_notification.cc
```diff
@@ -83,68 +105,70 @@ WindowsToastNotification::~WindowsToastNotification() {
   // Remove the notification on exit.
   if (toast_notification_) {
     RemoveCallbacks(toast_notification_.Get());
-    Dismiss();
+    // Dismiss();
```
MarshallOfSound, line 108: ?

docs/api/notification.md
```diff
@@ -29,9 +29,9 @@ Returns `Boolean` - Whether or not desktop notifications are supported on the cu
 ### `new Notification([options])`
 
 * `options` Object (optional)
-  * `title` String - A title for the notification, which will be shown at the top of the notification window when it is shown.
+  * `title` String (optional) - A title for the notification, which will be shown at the top of the notification window when it is shown.
```
MarshallOfSound, line 32: Technically `title` isn't optional.  Doesn't `toastXml` just completely replace all the other options in this structure.  Maybe this would make more sense if it was `new Notification(toastXml)`.  I.e. `options Object | String (optional)`

shell/browser/notifications/win/windows_toast_notification.cc
bitdisaster: Oh yeah I forget about that. This causes us not to show up in AC. I'm not sure why this was done.

docs/api/notification.md
bitdisaster: copy pasta 🤦

docs/api/notification.md
bitdisaster: yeah, its optional because when you provide Toast XML you don't need it. I thought about `options Object | String (optional)` Not sure how to implement on the API though.

bitdisaster: The notification will work without title and body. It actually did always because of a bug in the XML creation and Windows handling an empty notification gracefully. So if you provide an empty object nothing will blow up.

MarshallOfSound (approved):

jkleinsc:

miniak (approved):

PR #25404, opened 2020-09-10
fix: too many redirects when using POST method (intercepted requests)

marcin-prochownik:
#### Description of Change
For redirects, methods were not converted to 'GET' and caused too many redirects

#### Checklist
- [ ] `npm test` passes
- [ ] tests are [changed or added]
- [x] [PR release notes]

#### Release Notes

Notes: Fixed issue where on redirects methods were not converted to 'GET' and caused too many redirects

### DIFF
--- a/shell/browser/net/electron_url_loader_factory.cc
+++ b/shell/browser/net/electron_url_loader_factory.cc
@@ -189,6 +189,24 @@ void ElectronURLLoaderFactory::Clone(
   receivers_.Add(this, std::move(receiver));
 }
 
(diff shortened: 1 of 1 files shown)


marcin-prochownik:
I'm aware that this is not the best place to make this fix, however, could we merge this PR as a workaround, until it is fixed by chromium?

It is common for ASP websites to use POST and respond with redirect to 'return url'.

nornagon:
@marcin-prochownik I'd recommend submitting a fix upstream to Chromium.

I'm against patching this in Electron.

PR #25407, opened 2020-09-10
test: reduce flakiness in pdf test

nornagon:
#### Description of Change
This is a bit hacky, since really we ought to wait for the pdf frame to load, but I haven't been able to reproduce this test failing locally so I'm not sure what situation is causing it to fail.

For now, as this is one of our flakier tests, I think on balance there will be less total pain if we remove (what I assume is) the flaky bit from this test. Perhaps this is masking another bug, but I don't know what it is if so.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


samuelmaddock:
After #25385 is merged, the `extension-ready` event can be awaited for when the PDF extension is fully loaded. If the flakiness is caused by it not yet being loaded, this could fix it.

nornagon:
@samuelmaddock interesting, i'm not sure that's the case though as the pdf extension is a built-in extension, so it should be ready before any web contents are created, right?

jkleinsc (approved):

Issue #21666, opened 2020-01-02
Crashes in NativeBrowserViewViews::AutoResize function on Move/Resize


summeroff:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 6.0.7 and 6.1.5 
* **Operating System:**
  * Windows 10 x64 
* **Last Known Working Electron version:**
  * 

### Expected Behavior
App not crashing

### Actual Behavior
It crashes 

### To Reproduce
Was not able to reproduce myself or get reports with steps from user. 
Only information that we have is crash reports with backtraces.  

This backtraces have different call traces but all end with crash in "atom::NativeBrowserViewViews::AutoResize" file native_browser_view_views.cc line 54. 
Crashes of this two types of exceptions EXCEPTION_ACCESS_VIOLATION_READ and EXCEPTION_ACCESS_VIOLATION_EXEC. 
Calls to AutoResize leading to a crash originated from two different events Move and Resize. 

> atom::NativeBrowserViewViews::AutoResize native_browser_view_views.cc:54
> std::list<T>::list list:800
> atom::NativeWindowViews::GetBounds native_window_views.cc:588
> atom::NativeWindowViews::OnWidgetBoundsChanged native_window_views.cc:1222
> base::ObserverList<T>::Iter::EnsureValidIndex observer_list.h:211
> views::Widget::OnNativeWidgetMove widget.cc:1180
> display::win::ScreenWin::GetDisplayNearestPoint screen_win.cc:640
> views::DesktopWindowTreeHostWin::HandleMove desktop_window_tree_host_win.cc:836
> atom::NativeWindow::GetContentMaximumSize native_window.cc:287
> views::HWNDMessageHandler::_ProcessWindowMessage hwnd_message_handler.h:423
 

> atom::NativeBrowserViewViews::AutoResize native_browser_view_views.cc:54
> std::list<T>::list list:800
> atom::NativeWindowViews::OnWidgetBoundsChanged native_window_views.cc:1222
> base::ObserverList<T>::Iter::EnsureValidIndex observer_list.h:211
[...]

PR #25411, opened 2020-09-10
fix: prevent destroyed view references from causing crashes

mlaurencin:
#### Description of Change

Closes https://github.com/electron/electron/issues/21666.

This PR is fixing crashes caused by referencing and attempting to modify previously destroyed views.
Before, when a view was destroyed and then the contents were referenced for modification, the system would crash as undefined memory was accessed. This fix explicitly makes the pointer to the destroyed view's contents null, so that this will not happen.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed crashes caused by attempting to modify destroyed views


spec/api-browser-view-destroy.js
```diff
@@ -0,0 +1,31 @@
...
+    height: 600,
+    webPreferences: {
+      nodeIntegration: true
+    },
+  })
+  const view = new BrowserView();
+  mainWindow.addBrowserView(view);
+  view.webContents.destroy()
+  setTimeout(() => view.setBounds({ x: 0, y: 0, width: 300, height: 300 }), 10000) //Asynchronous Test
+  //view.setBounds({ x: 0, y: 0, width: 300, height: 300 }) //Synchronous Test
```
groundwater, line 16: We should move this back to the synchronous test for master, since it will always work.

MarshallOfSound: We actually have a folder for these things that get automatically run in a separate process to check for "did it crash".

Check out `spec-main/fixtures/crash-cases` @mlaurencin  👍

shell/browser/native_browser_view_views.cc
```diff
@@ -92,9 +92,35 @@ void NativeBrowserViewViews::ResetAutoResizeProportions() {
...
+  /*
   auto* view = GetInspectableWebContentsView()->GetView();
   view->SetBoundsRect(bounds);
   ResetAutoResizeProportions();
+  */
+
+  /*
+  This implementation is attempting to ignore setting bounds
+  if the view has been destroyed, (ie invalid).
+  A different check is most likely needed as the destroyed
```
MarshallOfSound, line 105: Very correct here, the best thing you can probably do is make `NativeBrowserViewViews` a `WebContentsObserver` and manually set `inspectable_web_contents_ = nullptr` when the webcontents is destroyed.

You can look at `BrowserView::WebContentsDestroyed` as an example of this pattern

spec/api-browser-view-destroy.js
mlaurencin: Awesome!

mlaurencin: > We should move this back to the synchronous test for master, since it will always work.

Alright, I replaced the asynchronous test with the synchronous one

codebytere: For a test of this nature, we have existing spec files which encompass similar functionality - as such we should look to run it there, as part of the test suite, instead of as its own file.

[This test](https://github.com/electron/electron/blob/c8a0b2b71d8622bbb848e364a842ea12bd5966fb/spec-main/api-browser-view-spec.ts#L145) provides an example of what we'd want to aim for - since only 3-4 lines are required to trigger the crash, we shouldn't need to spin up a bespoke file

shell/browser/native_browser_view_views.cc
```diff
@@ -92,9 +92,35 @@ void NativeBrowserViewViews::ResetAutoResizeProportions() {
...
+  pointer cannot be guaranteed NULL.
+  When the pointer is NULL, the program does not crash, but when
+  it is not, it proceeds to still crash.
+  */
+  auto* temp = GetInspectableWebContentsView();
+  LOG(INFO) << "NativeBrowserViewViews::SetBounds - temp: " << temp
```
codebytere, line 111: We do want to return early if there's no webContents. However, while this crash is specific to `setBounds`, if we take a look around this file, we'll notice that `GetInspectableWebContentsView()` is called in a large handful of other places. As such, any other code invoking those methods after a webContents is destructed will _also_ cause a crash, and we need to check those as well. In this case, `OnWidgetBoundsChanged` is called when the window is destructed during shutdown, and this method invokes `SetAutoResizeProportions`, which in turn makes the (incorrect) assumption that `GetInspectableWebContentsView()` is not a `nullptr`. Guarding against that case will likely fix the problem.

codebytere (commented):
This is a great start! as the crash in `setBounds` is but a symptom of a larger set of potential crashes, though, we'll want to make sure we address all similar potential crashes as well as try to add regression tests for all related crashes to really close up this bug :)

spec/api-browser-view-destroy.js
codebytere: Now that the other file has been added, this one should be deleted

shell/browser/native_browser_view_views.cc
codebytere: There are also some instances of this in the correspondent native_browser_view_mac.mm file which will also need to be checked and handled :)

shell/browser/native_browser_view_views.cc
codebytere: For the above, you can take a look [here](https://github.com/electron/electron/blob/3ac3fbdbfb4af9d7b35b02b0c3b4b8307c614c08/shell/browser/web_contents_zoom_controller.h#L20), and [here](https://github.com/electron/electron/blob/3ac3fbdbfb4af9d7b35b02b0c3b4b8307c614c08/shell/browser/web_contents_zoom_controller.h#L85-L89), though in this case we'll only want to override `WebContentsDestroyed()`

shell/browser/native_browser_view_views.cc
```diff
@@ -92,9 +92,35 @@ void NativeBrowserViewViews::ResetAutoResizeProportions() {
...
+            << " null: " << NULL;
+
+  if (temp != NULL) {
+    LOG(INFO)
+        << "NativeBrowserViewViews::SetBounds - NOT NULL, proceed as usual";
```
codebytere, line 116: We probably want to remove logs even though it's a wip :)

In addition - let's name the variable a bit more intuitively/readably, perhaps something like:

```cpp
auto* iwc_view = GetInspectableWebContentsView();
if (!iwc_view)
    return;
auto* view = iwc_view->GetView();
view->SetBoundsRect(bounds);
ResetAutoResizeProportions();
```

codebytere:
Looks like lint is still failing: see https://app.circleci.com/pipelines/github/electron/electron/29870/workflows/28683667-5c33-4ba1-a255-70e8d981c345/jobs/658611 for details :) You can check these locally with `npm run lint`

mlaurencin:
> Looks like lint is still failing: see https://app.circleci.com/pipelines/github/electron/electron/29870/workflows/28683667-5c33-4ba1-a255-70e8d981c345/jobs/658611 for details :) You can check these locally with `npm run lint`

Great! Pushing the fix now.

MarshallOfSound:
> I am getting local errors when running npm run lint saying "'gn' is not recognized as an internal or external command"

Ah that's because you're exclusively on build tools I think.  @codebytere we should either make `e lint` or come up with a better way of making those commands available.

@mlaurencin for now `e d npm run lint` would probably work (I know it looks funky but it works) 😆

shell/browser/native_browser_view_mac.mm
```diff
@@ -213,17 +221,22 @@ - (BOOL)mouseDownCanMoveWindow {
...
+  if (!contents)
+    return;
+  auto* view = contents->GetNativeView().GetNativeNSView();
   view.wantsLayer = YES;
   view.layer.backgroundColor = skia::CGColorCreateFromSkColor(color);
 }
 
 void NativeBrowserViewMac::UpdateDraggableRegions(
     const std::vector<gfx::Rect>& drag_exclude_rects) {
+  auto* inspectable_contents = GetInspectableWebContentsView();
```
codebytere, line 234: this should be `contents` for consistency although i think it might be clearer to use `auto* iwc_view = GetInspectableWebContentsView();` for this and other instances

mlaurencin:
> e d npm run lint



> > I am getting local errors when running npm run lint saying "'gn' is not recognized as an internal or external command"
> 
> Ah that's because you're exclusively on build tools I think. @codebytere we should either make `e lint` or come up with a better way of making those commands available.
> 
> @mlaurencin for now `e d npm run lint` would probably work (I know it looks funky but it works) 😆

I tried running that and ended up with this new error `ERROR Failed to run command, exit code was "null", error was 'Error: spawnSync npm ENOENT'` 😅  @codebytere

codebytere (approved):

MarshallOfSound (approved):

PR #25414, opened 2020-09-10
feat: remove getMediaSourceIdForWebContents()

nornagon:
This reverts #22701.

#### Description of Change
`desktopCapturer.getMediaSourceIdForWebContents` API has three serious problems as currently written:

1. It requires a WebContents ID, but the renderer process has no way to fetch WebContents IDs other than by IPC to the main process, and
2. No access control is applied to the caller, so any renderer process that can send an IPC message can get a token for a video stream of any other renderer.
3. The method is not exposed in the main process, despite being documented as being available in both main and renderer processes.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Removed the `desktopCapturer.getMediaSourceIdForWebContents()` API.

### DIFF
--- a/docs/api/desktop-capturer.md
+++ b/docs/api/desktop-capturer.md
@@ -72,50 +72,6 @@ const constraints = {
 }
 ```
 
-This example shows how to capture a video from a [WebContents](web-contents.md)
-
-```javascript
-// In the renderer process.
-const { desktopCapturer, remote } = require('electron')
-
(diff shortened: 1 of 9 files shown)


nornagon:
NB. this is marked as "security", but this API is only present in v11.0.0-beta.1 and higher. No stable versions are affected.

MarshallOfSound (approved):

nornagon:
cc @Volune

PR #25428, opened 2020-09-11
fix: Ensure electron delay loads the same modules as chromium

chrdavis:
This change adds the same module delay load list that chromium uses for electron.  Some modules were already getting delay loaded from other build files in chromium but not the main list via //build/config/win:delayloads.  We do not include the list of delay loads in delayloads_not_for_child_dll as those have issues being loaded in sandboxes processes.  This will reduce the overall reference set impact of the electron processes.
notes: Added missing module delay loads on windows to reduce per process reference set impact

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -1139,7 +1139,7 @@ if (is_mac) {
       ]
 
       ldflags = []
-      
+
       libs = [
         "comctl32.lib",
         "uiautomationcore.lib",
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -1138,13 +1138,13 @@ if (is_mac) {
         "//components/crash/core/app:run_as_crashpad_handler",
       ]
 
+      ldflags = []
+      
       libs = [
         "comctl32.lib",
         "uiautomationcore.lib",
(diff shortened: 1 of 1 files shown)


chrdavis:
Here is the list of delay loaded modules after the change:
image

nornagon (approved):

MarshallOfSound (approved):

deepak1556 (approved):

chrdavis:
Not sure what the build-mac failures are.  I can't access logs.  Also, my changes shouldn't impact anything but Windows.

nornagon:
mac failures seem to be flakes.

MarshallOfSound:
@chrdavis For future reference you should be able to view the CI build log just by signing in, you have to auth with github but the logs aren't behind any kind of permission barrier.  If you can't view them after signing in to CircleCI let me know and I can look into it 😄

chrdavis:
Do the mac tests have to be run again or is this ok to merge now?

deepak1556:
It is fine to merge, the failures are unrelated. Thanks for the pr!

PR #25430, opened 2020-09-11
fix: call node::Stop on exit

### LINKED ISSUE #25267: Electron 10 - sporadic `illegal access` errors resulting in exit code 7 during `process.exit`
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 10.0.0 and above
* **Operating System:**
  * Pop_OS 20.04, Debian 10
  * 9.2.1

### Expected Behavior

`process.exit(code)` exits the process with exit code `code`.

### Actual Behavior

After updating to Electron 10.0.0, `process.exit(0)` sporadically results in an error that looks like this:

```
undefined:0


illegal access
(Use `node --trace-uncaught ...` to show where the exception was thrown)
```

The process then exits with exit code `7`.

Passing `--trace-uncaught` to `electron` seems to have no effect on the output.

### To Reproduce

You can see this error popping up in the status checks on our Electron 10 upgrade PR: https://github.com/cypress-io/cypress/pull/8406

I attempted to make a reproducible example for this, but couldn't reproduce it outside of Cypress's tests.

To reproduce using Cypress's tests:

1. Clone Cypress, check out the Electron 10 branch, install dependencies:
	```
	git clone https://github.com/cypress-io/cypress.git
	cd cypress
	git checkout renovate/electron-10.x
	yarn
	```
2. Open `packages/server/test/e2e/1_deprecated_spec.ts` and add a `.only` to the `e2e.it` on line 22, so it reads:
	```
	e2e.it.only('push and no return - warns user exactly once', {
	```
3. Run the test until it fails:
	```
	cd packages/server
	# run this until it fails:
	yarn test-e2e 1_dep --browser electron
	# if you're using zsh, this will do the trick:
    while yarn test-e2e 1_dep --browser electron; do echo 'passed'; done
	```

You should see the test fail about 1/3 of the time with exit code `7` and the error mentioned above.

You can enable Cypress's debug logging by setting env var `DEBUG=cypress:*`, from which you can see that the last thing that happens before the crash is a call to `process.exit(0)`.

---

This is blocking us from upgrading to Electron 10, so please let me know if there is a way I can investigate further, or if I can provide any additional information to help debug.

nornagon:
#### Description of Change
I believe this fixes #25267, the "illegal access" message we sometimes see at exit.

I don't have a test or a full root cause analysis for this yet, but what appears to be happening is this:

1. When we call `node::FreeEnvironment` from `ElectronBrowserMainParts::PostMainMessageLoopRun`, it calls `node::CleanupHandles`.
2. `node::CleanupHandles` disallows JS execution, using [`Isolate::DisallowJavascriptExecutionScope`](https://github.com/nodejs/node/blob/b1831fed3a289d883944d23d2130368d90a48be2/src/env.cc#L595).
3. `node::CleanupHandles` [triggers a uv loop run](https://github.com/nodejs/node/blob/b1831fed3a289d883944d23d2130368d90a48be2/src/env.cc#L613), which triggers [`ElectronBindings::OnCallNextTick`](https://github.com/electron/electron/blob/ce865914594af79f0311a361a9d3c9a55d875f45/shell/common/api/electron_bindings.cc#L114).
4. `ElectronBindings::OnCallNextTick` [runs the microtask queue](https://github.com/electron/electron/blob/ce865914594af79f0311a361a9d3c9a55d875f45/shell/common/api/electron_bindings.cc#L123), which sometimes includes an entry into JS. (Not always, though. This is the bit I'm confused about: what are the conditions in which there's a pending JS microtask here?)
5. Attempting to execute JS while JS execution is disallowed [throws an "illegal access" exception](https://source.chromium.org/chromium/chromium/src/+/master:v8/src/execution/execution.cc;l=312;drc=2e96276c762073eb1e4de8a95bb944bd764f75fb).

This change causes `env->can_call_into_js()` to return false after `node::Stop()` is called, which means `InternalCallbackScope::Close()` [won't run any JS](https://github.com/nodejs/node/blob/a038199265519fa4db88a639b0e438675d23c4d1/src/api/callback.cc#L85), which avoids this error.

The remaining questions in my mind are:
1. Why were we not calling `node::Stop()` before? Was it simply an oversight, or was it an intentional choice?
2. Under what circumstances does `CleanupHandles` end up invoking JS? What is putting a JS task in the microtask queue during exit?
3. How can we write a test for this?

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue that could cause a normally-exiting process to fail with an "illegal access" message and exit code 7.

### DIFF
--- a/shell/app/node_main.cc
+++ b/shell/app/node_main.cc
@@ -290,6 +290,7 @@ int NodeMain(int argc, char* argv[]) {
 
     node::RunAtExit(env);
     node::FreeEnvironment(env);
+    node::Stop(env);
     node::FreeIsolateData(isolate_data);
 
     gin_env.platform()->DrainTasks(isolate);
--- a/shell/app/node_main.cc
+++ b/shell/app/node_main.cc
@@ -284,13 +284,12 @@ int NodeMain(int argc, char* argv[]) {
 
     node::ResetStdio();
 
-    env->set_can_call_into_js(false);
+    node::Stop(env);
     env->stop_sub_worker_contexts();
     env->RunCleanup();
 
     node::RunAtExit(env);
     node::FreeEnvironment(env);
-    node::Stop(env);
     node::FreeIsolateData(isolate_data);
 
     gin_env.platform()->DrainTasks(isolate);
--- a/shell/browser/electron_browser_main_parts.cc
+++ b/shell/browser/electron_browser_main_parts.cc
@@ -551,6 +551,7 @@ void ElectronBrowserMainParts::PostMainMessageLoopRun() {
   node_debugger_->Stop();
   node_env_->env()->set_trace_sync_io(false);
   js_env_->OnMessageLoopDestroying();
+  node::Stop(node_env_->env());
   node_env_.reset();
 
   ElectronBrowserContext::browser_context_map().clear();


nornagon:
cc @addaleax @codebytere

addaleax (approved):

zcbenz:
> * Why were we not calling `node::Stop()` before? Was it simply an oversight, or was it an intentional choice?

`node::Stop()` was only introduced in v11 and no one had noticed this new handy API since then.

zcbenz (approved):

codebytere (commented):
We'll also want to invoke this in `ELECTRON_RUN_AS_NODE` (e.g in `node_main`) @nornagon

nornagon:
The node embedder teardown API certainly seems a lot more complex than it needs to be 🤔

codebytere (approved):

MarshallOfSound (approved):

shawnchain:
It seems we encountered the same issue on 8.4.1. any hints to port this patch to 8.x.y branch ?

nornagon:
Yeah, if this was added in Node 11 then we should probably backport to 8.x.

Raynos:
When will this be release in 10.1.6 ?

Issue #25253, opened 2020-09-01
ready-to-show event is not fired and app window doesn't show


alexdevero:
### Preflight Checklist

* [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [X] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 10.1.1
* **Operating System:**
  * Windows 10 (20H2)
* **Last Known Working Electron version:**
  * 9.0.0

### Expected Behavior

Calling `mainWindow.once('ready-to-show' ...)` and inside it `mainWindow.show()` after creating a new BrowserWindow() on windows shouldshow the main window when it is ready to show.

### Actual Behavior

The `mainWindow.once('ready-to-show' ...)` is not fired and window doesn't show.

### To Reproduce

- Clone my [boilerplate](https://github.com/alexdevero/electron-react-webpack-boilerplate).
- Update electron to +10.0.0
- Start the app with `npm run`

### Additional Information

Content of `main.js`.

```
'use strict'

// Import parts of electron to use
const { app, BrowserWindow } = require('electron')
const path = require('path')
const url = require('url')

// Keep a global reference of the window object, if you don't, the window will
// be closed automatically when the JavaScript object is garbage collected.
let mainWindow

// Keep a reference for dev mode
let dev = false

// Broken:
// if (process.defaultApp || /[\\/]electron-prebuilt[\\/]/.test(process.execPath) || /[\\/]electron[\\/]/.test(process.execPath)) {
//   dev = true
// }

if (process.env.NODE_ENV !== undefined && process.env.NODE_ENV === 'development') {
  dev = true
}

// Temporary fix broken high-dpi scale factor on Windows (125% scaling)
// info: https://github.com/electron/electron/issues/9691
if (process.platform === 'win32') {
  app.commandLine.appendSwitch('high-dpi-support', 'true')
  app.commandLine.appendSwitch('force-device-scale-factor', '1')
}

function createWindow() {
  // Create the browser window.
  mainWindow = new BrowserWindow({
    width: 1024,
    height: 768,
    show: false,
    webPreferences: {
      nodeIntegration: true
    }
  })

  // and load the index.html of the app.
  let indexPath

[...]

Issue #7779, opened 2016-10-27
ready-to-show never fires


CharlieHess:
- Electron version: 1.3.7
- Operating system: macOS & Windows

I'm having some trouble creating a minimal repro for this but it seems that the `ready-to-show` event is not firing for us, since ~1.3.7, despite us having `show: false` in the window options. Filing this in case others are seeing the same behavior.

We had to switch to using `browserWindow.webContents.on('dom-ready', ...`).

kevinsawicki:
> Electron version: 1.3.7

Are you able to reproduce this on 1.4.x?

kevinsawicki:
> since ~1.3.7

So it fires in 1.3.6 but not in 1.3.7?

CharlieHess:
@kevinsawicki also repros on 1.4.4, 1.3.7. Let me step further back in time.

CharlieHess:
@kevinsawicki definitely working in 1.2.7.

kevinsawicki:
> definitely working in 1.2.7.

@CharlieHess what about 1.3.6?

CharlieHess:
@kevinsawicki works in 1.3.6.

kevinsawicki:
> works in 1.3.6

Hmm, that is really strange, the only change from 1.3.6 to 1.3.7 was a node upgrade, https://github.com/electron/electron/compare/v1.3.6...v1.3.7

I'm surprised that upgrade would impact this event firing unless something with the event loop integration broke/regressed.

kevinsawicki:
> I'm having some trouble creating a minimal repro

@CharlieHess any progress on this?

CharlieHess:
No, we swapped out the event and moved on to other things. 😞 I'll let you know if / when I get a simpler repro.

zcbenz:
As a work around, I think we can also emit `ready-to-show` on `dom-ready` event, so it would be guaranteed to emit.

Related: #6898.

mikecao:
I'm running into the same issue with 1.4.8 (windows x64). The strange thing is that it was working fine and then all of a sudden it just stopped working. No changes were made to the startup script. Had to resort the `dom-ready` event to get it working.

mikecao:
[...]

PR #25448, opened 2020-09-14

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/25253.

The `ready-to-show` event is not very stable that, sometimes on some machines it might not be fired, and the actual behavior depends on the version of Chromium.

To work around this bug, we ensure the `ready-to-show` event is emitted if it has not been emitted after `did-finish-load` event.

The real fix should be fixing the event itself, but unfortunately we haven't been able to find out why it did not emit so far, you can check https://github.com/electron/electron/issues/7779 on the history of this bug.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix `ready-to-show` event not emitted on some machines.


codebytere (commented):
@zcbenz looks like this is causing some test failures

zcbenz:
It seems that when loading `about:blank` Chromium would optimize the `did-finish-load` event to emit immediately, even earlier than `ready-to-show`. I have changed the test to actually draw something to work around this.

codebytere (approved):

PR #25461, opened 2020-09-14
feat: allow setting SSL config

nornagon:
#### Description of Change
This allows apps to configure min and max supported SSL versions. Chromium currently uses a minimum of TLS1.2 by default, whereas Electron defaults to TLS1.0 by default. This gives apps the ability to control supported TLS versions according to their needs.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `session.setSSLConfig()` to allow configuring SSL.

### DIFF
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -462,9 +462,10 @@ Returns `String` - The user agent for this session.
 
 * `config` Object
   * `minVersion` String - Can be `tls1`, `tls1.1`, `tls1.2` or `tls1.3`. The
-    minimum SSL version to allow when connecting to remote servers.
+    minimum SSL version to allow when connecting to remote servers. Defaults to
+    `tls1`.
   * `maxVersion` String - Can be `tls1.2` or `tls1.3`. The maximum SSL version
-    to allow when connecting to remote servers.
+    to allow when connecting to remote servers. Defaults to `tls1.3`.
 
 Sets the SSL configuration for the session. All subsequent network requests
 will use the new configuration. Existing network connections (such as WebSocket
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -458,6 +458,19 @@ will be temporary.
 
 Returns `String` - The user agent for this session.
 
+#### `ses.setSSLConfig(config)`
+
+* `config` Object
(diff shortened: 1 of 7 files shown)


deepak1556 (approved):
Implementation looks good.

One thing to make sure is that this doesn't configure the ssl client for the system network context, before network s13n the system network context was dependency for network fetching code of OCSP/CRL/AIA data but now the code path has been modified to use per-profile network context https://bugs.chromium.org/p/chromium/issues/detail?id=829029, so it should not be an issue. The only other use cases of system network context I was able to find was the one used for device_service (for fetching geolocation data) and creating tcp connection for viz devtools, both of which are fine without this change. Might want to ensure any other code paths I might have missed.

nornagon:
@deepak1556 hm, I'm not sure what I should be searching for. Can you give me some pointers?

Do you think we should add a `session.setDefaultSSLConfig()` method as well/instead?

deepak1556:
> I'm not sure what I should be searching for. Can you give me some pointers?

Checking for the usages of these apis https://github.com/electron/electron/blob/master/shell/browser/net/system_network_context_manager.h#L72-L82 should do.

> Do you think we should add a session.setDefaultSSLConfig() method as well/instead?

If there is no usage other than the ones I mentioned then its not required to add this, geolocation service is usually a google maps endpoint which the app doesn't have to control.

MarshallOfSound (approved):

nornagon:
I looked and I could only find those two as well.

As a precaution, what do you think about setting up the system network context to support the same TLS versions that Chrome does (without adding a new API to edit them)? i.e. min_version = TLS1.2.

deepak1556:
yeah that would be a good thing to do

jkleinsc:

jkleinsc (changes requested):
Just a small nit - can we note in the documentation what the defaults are if you do not set this config?

nornagon:
@jkleinsc NB. these defaults are set by a mojom file inside chromium and we don't have tests that will notify us if they change. So this documentation may go out of date.

jkleinsc (approved):

jkleinsc:
Merging as WOA failure is flake unrelated to this change.

Issue #24791, opened 2020-07-30
Suggestion: Specify domains to be excluded from websecurity


javaarchive:
### Preflight Checklist

* [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [X] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description
So I have this music app that utliizes youtube. I'd like to add stuff to the youtube embed like an audio visualizer. Due to cross origin iframe rules this is impossible even with header modifacation. I'd like to use the contentDocument and contentWindow attributes to connect audionodes from the parent page. I know i can do this by turning websecurity off but the music app also supports other sites some which can possibly be malicious and take advantage of websecurity being off. 

### Proposed Solution
A. An special header sent allowing a parent page with a domain in the list of domains specified in the header to execute javascript.
B. A list of hosts that don't have websecurity applied to them that can be passed to webpreferences. So 
Suppose site A is on the host list
If site A iframes site B:
site A can execute javascript but site B can't iframe another site and execute js on it. 

### Alternatives Considered
I've considered injecting javascript into the iframe but looking webcontents and webframe I can only find the frames for devtools and the page but I can't find a frame for the iframe so I can't execute the js. 

### Additional Information
I'd like a nice solution to this. I don't want to end up injecting the javascript via webcontents or something and transferring data via post message because I'd have to exchange data on both sides.

codebytere:
Hi, while we appreciate this enhancement request it's not something that our team will be able to implement in the near future, and since we use our issues tab as the team's backlog, we prefer to only retain issues that are on our immediate roadmap. If you'd like to make a PR for this request we would welcome that but in the meantime I'm going to go ahead and close this.

Issue #18371, opened 2019-05-20
webFrame.findFrameByRoutingId and webFrame.firstChild returns frames only if they belong to the same domain.


Lukich:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
5.0.1
* **Operating System:**
OSX 10.13.1
* **Last Known Working Electron version:**:
4.1.4

### Expected Behavior
When calling webFrame.findFrameByRoutingId and webFrame.firstChild in the preload script for a page with iframes I should be able to get a reference to the frame and perform executeJavaScript call on it regardless of the domain it belongs to.

### Actual Behavior
For all frames belonging to a domain different from the top domain of the page these two commands return null.

### To Reproduce

https://github.com/Lukich/webFrame-bug-report.git

```sh
$ git clone https://github.com/Lukich/webFrame-bug-report.git
$ electron .
```

bughit:
The WebFrame docs mention that only frames in the same renderer process can be returned. I'm guessing the iframe isolation model has changed to put different domains in different processes. Don't know if there's any control over that.

bughit:
Access to the the full frame hierarchy would have to be exposed in the browser (main process), #17252

electron-triage:
Thank you for taking the time to report this issue and helping to make Electron better.

The version of Electron you reported this on has been superseded by newer releases. 

If you're still experiencing this issue in Electron 6.x.y or later, please add a comment specifying the version you're testing with and any other new information that a maintainer trying to reproduce the issue should know.

I'm setting the `blocked/need-info` label for the above reasons. This issue will be closed 7 days from now if there is no response.

Thanks in advance! Your help is appreciated.

electron-triage:
Thank you for your issue!

We haven't gotten a response to our questions in our comment above. With only the information that is currently in the issue, we don't have enough information to take action. I'm going to close this but don't hesitate to reach out if you have or find the answers we need, we'll be happy to reopen the issue.

Issue #17252, opened 2019-03-06
Provide API for accessing subframes from `webContents`


samuelmaddock:
### Problem Description

As a test author, I'd like to be able to interact with subframes of a renderer process in a remote `BrowserWindow` (i.e. execute JavaScript).

### Proposed Solution

Adding a `webFrame` proxy API to the `webContents` API could offer a complete solution to interacting with frames in the renderer process. The following type definitions would be an example of this.

```ts
/**
 * Proxy for `webFrame` renderer API.
 * Could partially implement `webFrame` API.
 */
interface WebFrameProxy {
  /** Provide mappings to methods for getting subframes. */
  findFrameByRoutingId(routingId: number): Promise<WebFrameProxy>

  /** Example method which would run in the appropriate frame. */
  executeJavaScript<T>(code: string, userGesture: boolean): Promise<T>
}

interface WebContents {
  /** Top-level `webFrame` instance of `WebContents` instance. */
  webFrame: WebFrameProxy
}

// Internally, webFrame-related methods could map to the top-level webFrame.
// See `asyncWebFrameMethods` in web-contents.js for other methods.
const webContents: WebContents
webContents.executeJavaScript =
  webContents.webFrame.executeJavaScript.bind(webContents.webFrame)
```
### Alternatives Considered

For a less complete solution, additional methods could be added as needed to the `webContents` API for interacting with subframes.
```ts
webContents.executeJavaScriptInFrame(frameRoutingId: number, code: string, userGesture: boolean): Promise<T>
```

### Additional Information

This could be a potential fix for #14350 if the `WebFrameProxy` were to implement a `executeJavaScriptInIsolatedWorld` proxy method.

cumajkeee:
@samuelmaddock, is it possible to add in setSpellCheckProvider method in proposed proxy object? It would solve my issue where I'd like to have spellchecker works in nativeBrowserWindow.

Issue #14350, opened 2018-08-28
WebView and WebContents need executeJavaScriptInIsolatedWorld equivalent


TheGoddessInari:
**Is your feature request related to a problem? Please describe.**
This is related to #14288 where I'm trying to run remote untrusted  web content in isolation, as per documentation.

**Describe the solution you'd like**
WebView and WebContents need something like executeJavaScriptInIsolatedWorld, perhaps named executeJavaScriptInIsolatedContext, to be able to actually execute javascript in the isolated context that's created with a webPreferences that turns contextIsolation on.
This should behave according to the documentation, and still allow access to document and window.

**Describe alternatives you've considered**
I've tried looking into executeJavaScriptInIsolatedWorld, but can't get a reference to the specific webFrame needed from webview or webContents. Instead of making users do this manually (which currently doesn't work), exposing a relevant API would be preferential.

**Additional context**
My electron-based app is trying to be security-conscious, but the currently documented methods aren't developer-friendly and are experimental.

codebytere:
Given that this has been open for years, that this isn't currently on our roadmap, and in the interest of our issue tracker more closely reflecting our workstreams, i'm going to go ahead and label this a `wontfix`. However, should someone determine a workable solution we would be potentially open to accepting a PR!

PR #25464, opened 2020-09-15
feat: add webFrameMain API to the main process

samuelmaddock:
#### Description of Change

While working on extension support in Electron and in other cases such as in writing tests (#17252), I've found a need to be able to access information about subframes within an instance of `WebContents`. The `webFrame` API is supported in the renderer process, but there's no equivalent in the main process.

[Chromium's browser architecture](https://www.chromium.org/developers/design-documents/oop-iframes#TOC-Architecture-Overview) (since OOPIF) includes the concept of a "render frame" to represent a frame within the web content tree. This includes the `RenderFrame` and `RenderFrameHost` classes to represent a frame in the renderer and browser processes respectively.

The `webFrame` API available to the renderer process seems to cover the case of accessing information about a `RenderFrame`. These proposed changes aim to expose access to `RenderFrameHost`s from the main process.

I'd like to ask @electron/wg-api to please review early progress of this proposed API and let me know if it's something that the Electron team would be interested in.

-----

Here's an overview of what I envision a finalized API could look like:
```ts
/** An abstraction for a RenderFrameHost in the main process. */
interface WebFrameMain {
  processId: number;
  routingId: number;
  url: string;

  /**
   * This specific ID often comes into use from extensions. It corresponds to
   * the ID of the FrameTreeNode owned by a RenderFrameHost.
   **/
  frameTreeNodeId: number;

  // Traversal properties
  top: WebFrameMain;
  parent: WebFrameMain;
  frames: WebFrameMain[];

  /** Every frame in this frame's subtree, including itself. Useful for accessing the full tree as an array. */
  framesInSubtree: WebFrameMain[];

  // Methods
  executeJavaScript(code: string, useGesture: boolean): Promise<unknown>;
  reload(): boolean;
}

/** Global API for accessing WebFrames. */
interface WebFrameMainAPI {
  fromId(processId: number, routingId: number): WebFrameMain;
}

interface WebContents {
  /** Shortcut to get the top-level frame. */
  mainFrame: WebFrameMain;
}
```

With today's navigation events in Electron, this could offer a more complete API for interacting with frames:
```ts
webContents.on('did-start-navigation', (
  event: Electron.IpcMainEvent,
  url: string,
  isInPlace: boolean,
  isMainFrame: boolean,
  frameProcessId: number,
  frameRoutingId: number
) => {
  const frame = webFrameMain.fromId(frameProcessId, frameRoutingId);
  // ...
});

// OR it could be used to clean up navigation APIs
webContents.on('did-start-navigation', (
  event: Electron.IpcMainEvent,
  url: string,
  isInPlace: boolean,
  frame: WebFrame
) => {
  const { processId, routingId } = frame;
  const isMainFrame = frame.top === frame;
  // ...
})
```

-----

TODO:
- [x] Mark deleted `WebFrame` instances to prevent potential null pointer access of `RenderFrameHost`s.
- [x] Implement remainder of proposed API.
- [x] ~~Revise API to use similar interfaces to what's already offered by `webFrame`? I'm interested to hear what folks think about this.~~
- [x] Merge electron/typescript-definitions#175 to fix type definitions.

-----

Related issues which this could help solve:
#24791 - discovery of all frames of a domain
#18371 - access cross-domain iframes (not limited to single renderer process)
#14350 - access to all frames of webcontents

solves #17252

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `webFrameMain` API for accessing sub-frames of a `WebContents` instance from the main process.


michalgagat (approved):

jkleinsc:
The API WG reviewed this at the Sept 21, 2020 meeting.

samuelmaddock:
Making progress. I'm now able to inject JS into subframes. 🙂

image

```js
const { mainFrame } = win.webContents;

[mainFrame, ...mainFrame.frames].forEach(frame => {
  frame.executeJavaScript(`document.body.style.background = window.top === window.self ? 'red' : 'blue'`, false)
})
```

samuelmaddock:
@electron/wg-api This API should be good to start reviewing. I don't plan to add any additional features for this initial pass.

docs/api/web-frame-main.md
```diff
@@ -0,0 +1,91 @@
...
+## Methods
+
+These methods can be accessed from the `webFrameMain` module:
+
+```javascript
+const { webFrameMain } = require('electron')
+console.log(webFrameMain)
+```
+
+### `webFrameMain.fromId(processId, routingId)`
```
nornagon, line 16: I'm a little sad to add a new top-level module just for this one function.

Perhaps we could make this `webContents.frameFromId(...)`?

or, to follow `BrowserWindow.fromId()`, make it a static method on the `WebFrameMain` class?

docs/api/web-frame-main.md
```diff
@@ -0,0 +1,91 @@
...
+
+Process: [Main](../glossary.md#main-process)
+
+## Methods
+
+These methods can be accessed from the `webFrameMain` module:
+
+```javascript
+const { webFrameMain } = require('electron')
+console.log(webFrameMain)
```
nornagon, line 13: I think we can do better with this example :) Perhaps show an example of what you might do with the module?

shell/browser/api/electron_api_web_frame_main.cc
```diff
@@ -0,0 +1,230 @@
...
+// WebFrame can outlive its RenderFrameHost pointer so we need to check whether
+// its been disposed of prior to accessing it.
+#define CHECK_RENDER_FRAME(funcname, args, value)                      \
+  if (render_frame_disposed_) {                                        \
+    args->isolate()->ThrowException(v8::Exception::Error(              \
+        gin::StringToV8(args->isolate(),                               \
+                        "Render frame was disposed before "            \
+                        "WebFrame." #funcname " could be executed"))); \
+    return value;                                                      \
+  }
```
nornagon, line 46: Prefer a private method over a #define please. See e.g. `CheckAlive` in shell/browser/api/electron_api_tray.cc.

shell/browser/api/electron_api_web_frame_main.cc
```diff
@@ -0,0 +1,230 @@
...
+
+// static
+gin::Handle<WebFrame> WebFrame::From(v8::Isolate* isolate,
+                                     content::RenderFrameHost* rfh) {
+  if (rfh == nullptr)
+    return gin::Handle<WebFrame>();
+  auto frame_map = g_render_frame_map.Get();
+  auto iter = frame_map.find(rfh);
+  WebFrame* web_frame =
+      iter == frame_map.end() ? new WebFrame(rfh) : iter->second;
```
nornagon, line 156: Perhaps refactor this lookup into a common function? The code is repeated in a few places and I think a `WebFrame* FromRenderFrameHost(content::RenderFrameHost* rfh)` function would aid readability.

shell/browser/api/electron_api_web_frame_main.cc
```diff
@@ -0,0 +1,230 @@
...
+
+namespace {
+
+using electron::api::WebFrame;
+
+v8::Local<v8::Value> FromID(gin_helper::ErrorThrower thrower,
+                            int render_process_id,
+                            int render_frame_id) {
+  if (!electron::Browser::Get()->is_ready()) {
+    thrower.ThrowError("WebFrame can only be received when app is ready");
```
nornagon, line 211: nit:
```suggestion
    thrower.ThrowError("WebFrame is available only after app ready");
```

shell/browser/api/electron_api_web_frame_main.cc
```diff
@@ -0,0 +1,230 @@
...
+
+void WebFrame::MarkRenderFrameDisposed() {
+  g_render_frame_map.Get().erase(render_frame_);
+  render_frame_disposed_ = true;
+}
+
+v8::Local<v8::Promise> WebFrame::ExecuteJavaScript(const base::string16& code,
+                                                   bool has_user_gesture,
+                                                   gin::Arguments* args) {
+  gin_helper::Promise<base::Value> promise(args->isolate());
```
nornagon, line 64:
```suggestion
                                                   v8::Isolate* isolate) {
  gin_helper::Promise<base::Value> promise(isolate);
```

shell/browser/api/electron_api_web_frame_main.cc
```diff
@@ -0,0 +1,230 @@
...
+
+  render_frame_->ExecuteJavaScriptForTests(
+      code, base::BindOnce([](gin_helper::Promise<base::Value> promise,
+                              base::Value value) { promise.Resolve(value); },
+                           std::move(promise)));
+
+  return handle;
+}
+
+bool WebFrame::Reload(gin::Arguments* args) {
```
nornagon, line 89: These functions can all take `v8::Isolate*` instead of `gin::Arguments*`, which will cause them to correctly throw errors when they receive more arguments than expected.

nornagon (changes requested):

shell/browser/api/electron_api_web_frame_main.cc
samuelmaddock: I agree with your suggestion, however, this was copied from elsewhere in the codebase. The same message is used for the session and webframe APIs.

If we change it here, I imagine that'd make it harder for folks to find the solution if they searched the error message. Or we could change it for all of them. Or maybe I'm overthinking it. 😅

shell/browser/api/electron_api_web_frame_main.cc
samuelmaddock: I wanted to ask about this function's design. In other code we allow `userGesture` to be an optional parameter. Should we do the same here?

If so, is there a good example for implementing optional parameter with these gin types?

docs/api/web-frame-main.md
samuelmaddock: I like the latter suggestion of adding something like `WebFrameMain.fromId()`.

I noticed that APIs only seem to use this pattern when the class itself is constructable (with `new`). 
- `WebContents` doesn't have a constructor and it provides a top-level `webContents` API.
- `BrowserView` has a constructor and it provides static methods on the class.
- `BrowserWindow` has a constructor and it provides static methods on the class.

In this case, `WebFrameMain` doesn't have a constructor and it would provide static methods on the class. This seems unique, but if we're okay with it, I'll move forward with this design.

docs/api/web-frame-main.md
samuelmaddock: I blame the `webContents` docs 😂

I'll update once we decide on a path forward for static methods.

docs/api/web-frame-main.md
nornagon: hm, you're right. Let's be consistent, then.

shell/browser/api/electron_api_web_frame_main.cc
nornagon: Yeah... this error isn't consistent in our codebase (there are already a few variations). The one I picked here is one of the more grammatically fluent variations. I'm not sure I understand your point about searching the error message, it should be just as searchable as it remains unique.

samuelmaddock: If the error message remains similar to others, finding it in search engines might be easier. Since you mentioned we use the grammatically improved version elsewhere, I'll switch to that. 👍

docs/api/web-frame-main.md
samuelmaddock: Added usage examples in cf87997

shell/browser/api/electron_api_web_frame_main.cc
```diff
@@ -0,0 +1,255 @@
...
+#include "shell/browser/api/electron_api_web_frame_main.h"
+
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "base/lazy_instance.h"
+#include "base/logging.h"
+#include "content/browser/renderer_host/frame_tree_node.h"  // nogncheck
```
nornagon, line 14: can we avoid using private APIs from `//content`?

samuelmaddock: Currently this is required for executing JS with a user gesture. https://github.com/electron/electron/blob/cf87997176986deee6c6b64d242e2ccc7871fce7/shell/browser/api/electron_api_web_frame_main.cc#L94-L99

The public methods available on a `RenderFrameHost` lack the ability to execute JS with both a callback and a user gesture.

For reference, these are the two methods to choose from:
```cpp
// This runs the JavaScript, but without restrictions. THIS IS ONLY FOR TESTS.
virtual void ExecuteJavaScriptForTests(
    const base::string16& javascript,
    JavaScriptResultCallback callback,
    int32_t world_id = ISOLATED_WORLD_ID_GLOBAL) = 0;

// This runs the JavaScript, but without restrictions. THIS IS ONLY FOR TESTS.
// Unlike the method above, this one triggers a fake user activation
// notification to test functionalities that are gated by user
// activation.
virtual void ExecuteJavaScriptWithUserGestureForTests(
    const base::string16& javascript,
    int32_t world_id = ISOLATED_WORLD_ID_GLOBAL) = 0;
```

Also, the `*ForTests` methods were used because we have precedence for using one of them here: https://github.com/electron/electron/blob/69f1731bbbaa5211aefe64f3d155b11ab7bcb327/shell/browser/ui/inspectable_web_contents.cc#L1052

shell/browser/api/electron_api_web_frame_main.h
```diff
@@ -0,0 +1,94 @@
...
+namespace gin_helper {
+class Dictionary;
+}
+
+namespace electron {
+
+namespace api {
+
+// Bindings for accessing frames from the main process.
+class WebFrame : public gin::Wrappable<WebFrame> {
```
MarshallOfSound, line 34: Consistency says this class should actually be called `WebFrameMain`

spec-main/api-web-frame-main-spec.ts
```diff
@@ -0,0 +1,203 @@
...
+      // Wait for WebContents, and thus RenderFrameHost, to be destroyed.
+      await new Promise(resolve => setTimeout(resolve, 0));
+    });
+
+    it('throws upon accessing properties', () => {
+      expect(() => webFrame.url).to.throw();
+    });
+  });
+
+  // TODO: see https://github.com/electron/typescript-definitions/pull/175
```
MarshallOfSound, line 179: Update TS defs in this PR, your change landed

MarshallOfSound (commented):
API seems legit, two nits

docs/api/web-contents.md
```diff
@@ -1909,3 +1909,7 @@ A [`Debugger`](debugger.md) instance for this webContents.
 
 A `Boolean` property that determines whether or not this WebContents will throttle animations and timers
 when the page becomes backgrounded. This also affects the Page Visibility API.
+
+#### `contents.webFrame` _Readonly_
```
nornagon, line 1913: Perhaps this should be `mainFrame`?

jkleinsc:
The API WG approved this API on Oct 8th, 2020.

docs/api/web-contents.md
samuelmaddock: I originally was using this and changed it for some reason. I'll change it back since I think it improves semantics.

samuelmaddock: 9015ba1

shell/browser/api/electron_api_web_frame_main.h
samuelmaddock: 7980039

nornagon (approved):

samuelmaddock:
@electron/wg-releases would it make sense to backport this to `11-x-y`? 🤔

bughit:
Though I'd mention it here as it's kind of related. WebRequest does not currently provide direct frame info (#25952) in it's callbacks, which is potentially significant and useful.

ckerr:
@samuelmaddock 11-x-y is getting pretty close to stable and this is a pretty big patch. How urgent is your need for this in 11?

I'd lean towards saying 'no' to this but if you have an important use case s.t. this would need to be in 11, please let us know.

samuelmaddock:
> 
> 
> @samuelmaddock 11-x-y is getting pretty close to stable and this is a pretty big patch. How urgent is your need for this in 11?
> 
> I'd lean towards saying 'no' to this but if you have an important use case s.t. this would need to be in 11, please let us know.

This is a nice to have for me. If we're cutting releases of v12 when v11 is stable, it's fine by me to keep it out of v11. 👍

jkleinsc:
The releases WG declined backporting this PR to 11-x-y at the Oct 28, 2020 meeting

Issue #25427, opened 2020-09-11
Exit from HTML fullscreen and enter native fullscreen leads to appearance of unclickable fantom window image.


kirille:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v10.1.1
* **Operating System:**
  * macOS 10.15.6
* **Last Known Working Electron version:**
  * 

### Expected Behavior
consecutive exit from HTML fullscreen and enter native fullscreen do not create unclickable and indestructible fantom window image.

### Actual Behavior
consecutive exit from HTML fullscreen and enter native fullscreen create unclickable and indestructible fantom window image.

### To Reproduce

Modified electron quick-start project:

[kirille/electron-fullscreen-fantom-issue](https://github.com/kirille/electron-fullscreen-fantom-issue)

on start, opens app in html-fullscreen, on click on Do Bug button, exit html-fullscreen, wait for exit, enter native fullscreen.

### Screenshots

<img width="1143" alt="Screenshot 2020-09-11 at 18 19 32" src="https://user-images.githubusercontent.com/6457945/92944398-97d97180-f45c-11ea-973f-58cf68c9dd42.png">

hovancik:
Hi @codebytere ,

I am having similar issue with electron 9 on macOS

I have a fullscreen window and when I close it, in CMD+Tab I can still see it as an option. When I "open it", there is a phantom fullscreen window, like in a screenshot: (you can see that top and bottom bar are hidden by that window)

<img width="256" alt="image" src="https://user-images.githubusercontent.com/4745859/93663353-9cf37d80-fa67-11ea-9abf-61ac3d428862.png">

I did some digging and: 
- both versions 9.1 and 9.3 are broken
- version 9.1 with `app.dock.hide()` works fine

kirille:
Hi @codebytere,

The issue is still reproducible with electron 11.

Updated project to reproduce: https://github.com/kirille/electron-fullscreen-fantom-issue

Is there any hope to fix this in near future? It brings pretty terrible UX.

kirille:
Reproducible with electron version 12.0.0-beta.24 as well

[...]

Issue #25426, opened 2020-09-11
Native close window button disappears on mac after fullscreen off/on 


kirille:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v10.1.1
* **Operating System:**
  * macOS 10.15.6
* **Last Known Working Electron version:**
  * don't know

### Expected Behavior

sequence of:

window.setFullScreen(false);
window.setFullScreen(true);

does not lead to missing native window control buttons.

### Actual Behavior

sequence of:

window.setFullScreen(false);
window.setFullScreen(true);

leads to missing native window control buttons.

### To Reproduce

updated electron quick start app:
[no-native-window-buttons.zip](https://github.com/electron/electron/files/5209788/no-native-window-buttons.zip)

this will start app in fullscreen mode. Do Bug button just set fullscreen true and false consequentially.

### Screenshots

<img width="1920" alt="Screenshot 2020-09-11 at 17 53 12" src="https://user-images.githubusercontent.com/6457945/92941323-e7b63980-f458-11ea-9c28-695ca03c4afb.png">

PR #25470, opened 2020-09-15
fix: handle async nature of [NSWindow -toggleFullScreen]

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/25426.
Closes https://github.com/electron/electron/issues/25427.

`[NSWindow -toggleFullScreen]` is an asynchronous operation, which means that it's possible to call it while a fullscreen transition is currently in process. As such, this can create weird behavior (incl. phantom windows and missing traffic lights) and so we want to queue up transitions until the previous ones have completed. We do this by tracking when fullscreen transitions begin and end, and scheduling those transitions to execute in order after any current transitions complete.

cc @MarshallOfSound @nornagon @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where multiple calls to `window.setFullScreen` could cause problems.

### DIFF
--- a/shell/browser/native_window_mac.h
+++ b/shell/browser/native_window_mac.h
@@ -8,6 +8,7 @@
 #import <Cocoa/Cocoa.h>
 
 #include <memory>
+#include <queue>
 #include <string>
 #include <tuple>
 #include <vector>
@@ -164,6 +165,12 @@ class NativeWindowMac : public NativeWindow,
   void SetCollectionBehavior(bool on, NSUInteger flag);
   void SetWindowLevel(int level);
 
+  enum class FullScreenTransitionState { ENTERING, EXITING, NONE };
+
+  // Handle fullscreen transitions.
+  void SetFullScreenTransitionState(FullScreenTransitionState state);
+  void HandlePendingFullscreenTransitions();
+
   enum class VisualEffectState {
     kFollowWindow,
     kActive,
@@ -182,7 +189,6 @@ class NativeWindowMac : public NativeWindow,
   ElectronTouchBar* touch_bar() const { return touch_bar_.get(); }
   bool zoom_to_page_width() const { return zoom_to_page_width_; }
   bool always_simple_fullscreen() const { return always_simple_fullscreen_; }
-  bool exiting_fullscreen() const { return exiting_fullscreen_; }
 
  protected:
   // views::WidgetDelegate:
@@ -224,11 +230,16 @@ class NativeWindowMac : public NativeWindow,
   std::unique_ptr<RootViewMac> root_view_;
 
   bool is_kiosk_ = false;
-  bool was_fullscreen_ = false;
   bool zoom_to_page_width_ = false;
   bool resizable_ = true;
-  bool exiting_fullscreen_ = false;
   base::Optional<gfx::Point> traffic_light_position_;
+
+  std::queue<bool> pending_transitions_;
+  FullScreenTransitionState fullscreen_transition_state() const {
+    return fullscreen_transition_state_;
+  }
+  FullScreenTransitionState fullscreen_transition_state_ =
+      FullScreenTransitionState::NONE;
 
   NSInteger attention_request_id_ = 0;  // identifier from requestUserAttention
 
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -668,15 +668,7 @@ void NativeWindowMac::SetFullScreen(bool fullscreen) {
                                      ? FullScreenTransitionState::ENTERING
                                      : FullScreenTransitionState::EXITING;
 
-  // Until 10.13, AppKit would obey a call to -toggleFullScreen: made inside
-  // windowDidEnterFullScreen & windowDidExitFullScreen. Starting in 10.13,
-  // it behaves as though the transition is still in progress and just emits
-  // "not in a fullscreen state" when trying to exit fullscreen in the same
-  // runloop that entered it. To handle this, invoke -toggleFullScreen:
-  // asynchronously.
-  [window_ performSelector:@selector(toggleFullScreen:)
-                withObject:nil
-                afterDelay:0];
+  [window_ toggleFullScreenMode:nil];
 }
 
 bool NativeWindowMac::IsFullscreen() const {
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -557,6 +557,11 @@ bool NativeWindowMac::IsVisible() {
   return [window_ isVisible] && !occluded && !IsMinimized();
 }
 
+void NativeWindowMac::SetFullScreenTransitionState(
+    FullScreenTransitionState state) {
+  fullscreen_transition_state_ = state;
+}
+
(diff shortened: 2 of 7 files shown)


shell/browser/ui/cocoa/electron_ns_window.mm
```diff
@@ -224,6 +224,18 @@ - (void)toggleFullScreenMode:(id)sender {
...
+
+    // [NSWindow -toggleFullScreen] is an asynchronous operation, which means
+    // that it's possible to call it while a fullscreen transition is currently
+    // in process. This can create weird behavior (incl. phantom windows),
+    // so we want to block the operation until any current transitions have
+    // completed.
+    while (shell_->entering_fullscreen() || shell_->exiting_fullscreen()) {
+      [[NSRunLoop currentRunLoop]
+             runMode:NSDefaultRunLoopMode
+          beforeDate:[NSDate dateWithTimeIntervalSinceNow:.5]];
```
nornagon, line 236: 🚨 

Are we sure we want to run a nested run loop? I think this can, for example, re-enter JS. Which probably isn't what we want.

Also, using a specific time interval seems very flaky to me. Is this the same time interval against which the animation is measured? How will we know if macOS tweaks the length of the animation? If there's high CPU load, does the animation take longer? We should use some kind of event to know when this is done, instead of a time interval.

nornagon (changes requested):
the [docs](https://developer.apple.com/documentation/appkit/nswindow/1419527-togglefullscreen?language=objc) don't mention anything about not calling this twice.

shell/browser/ui/cocoa/electron_ns_window.mm
codebytere: Refactored to schedule transitions instead!

shell/browser/native_window_mac.mm
```diff
@@ -792,10 +796,29 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
...
+    return;
+
+  const auto& next_transition = pending_transitions_.front();
+  pending_transitions_.erase(pending_transitions_.begin());
+
+  SetFullScreen(next_transition);
+}
+
 void NativeWindowMac::SetFullScreen(bool fullscreen) {
   if (fullscreen == IsFullscreen())
```
zcbenz, line 810: This check should be modified to test the tail of `pending_transitions_` instead of current fullscreen state. Otherwise following calls would not work:

```is
// current state: not in fullscreen
window.setFullScreen(true);  // current state: fullscreen
window.setFullScreen(false); // pushed. current state: fullscreen
window.setFullScreen(true);  // ignored as current state is fullscreen
```

zcbenz (changes requested):

shell/browser/native_window_mac.h
```diff
@@ -214,9 +217,15 @@ class NativeWindowMac : public NativeWindow, public ui::NativeThemeObserver {
   bool zoom_to_page_width_ = false;
   bool fullscreen_window_title_ = false;
   bool resizable_ = true;
-  bool exiting_fullscreen_ = false;
   gfx::Point traffic_light_position_;
 
+  enum class FullScreenTransitionState { ENTERING, EXITING, NONE };
```
nornagon, line 222: It doesn't seem to me that there's any difference between `ENTERING` and `EXITING` in terms of behavior here. Can this be collapsed into two states?

nornagon (commented):
I think we could write some tests for this!

e.g.:

`requestFullScreen` immediately followed by `cancelFullScreen` should produce the [appropriate events](https://www.electronjs.org/docs/api/browser-window#event-enter-html-full-screen) and the window should ultimately end up not in fullscreen mode. `request` -> `cancel` -> `request` should end up in fullscreen mode and likewise generate the relevant events.

shell/browser/native_window_mac.mm
```diff
@@ -792,7 +793,24 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
...
+  bool next_transition = pending_transitions_.pop();
+  SetFullScreen(next_transition);
+}
+
 void NativeWindowMac::SetFullScreen(bool fullscreen) {
+  // [NSWindow -toggleFullScreen] is an asynchronous operation, which means
+  // that it's possible to call it while a fullscreen transition is currently
+  // in process. This can create weird behavior (incl. phantom windows),
+  // so we want to schedule a transition for when the current one has completed.
+  if (fullscreen_transition_state() != FullScreenTransitionState::NONE) {
```
zcbenz, line 809: We can check the tail of `pending_transitions_` to see if we actually need to push a new transition.

If `pending_transitions_.back() == fullscreen` then the call can be dropped:

```js
setFullScreen(true);
setFullScreen(true);  // can be dropped
```

if  `pending_transitions_.back() != fullscreen` then the call can be dropped:

```js
setFullScreen(false);
setFullScreen(true);  // pending
setFullScreen(false);  // can cancel the pending call
```

We probably does not need a queue at all since with this optimization the queue's length won't be larger than 1. Something like `base::Optional<bool> pending_transition_` might be enough to do the work.

```c++
if (pending_transition_) {
  if (pending_transition_.get() == fullscreen) {
    // ignored as they are the same call.
  } else {
    pending_transition_.reset();  // current call cancels the pending transition.
  }
  return;
}
```

shell/browser/native_window_mac.h
codebytere: @nornagon we specifically need to check entering for redrawing of traffic lights, which is why i kept these separate

shell/browser/native_window_mac.mm
codebytere: I see the goals for this but i also think devs may want to actually complete the transitions - optimizing away the visual behavior is perhaps better but also less intuitive and may lead to confused devs wondering why the transitions seem never to have occurred 🤔

zcbenz: I'm good not optimizing the transitions, but we still need to do something about duplicate calls. 

Since the duplicate calls won't actually do anything, the related events will not be emitted and the `HandlePendingFullscreenTransitions` method won't be called. So if there are still pending transitions they won't be processed after duplicate calls.


```js
setFullScreen(true);
setFullScreen(true);  // does not trigger events
setFullScreen(true);  // will be left in pending transitions unprocessed
```

codebytere:
Holding this for a bit to explore a few edge cases

spec-main/api-browser-window-spec.ts
```diff
@@ -4132,6 +4132,17 @@ describe('BrowserWindow module', () => {
...
 
+      it('handles several transitions in close proximity', async () => {
+        const w = new BrowserWindow();
+
+        w.setFullScreen(true);
+        w.setFullScreen(false);
+        w.setFullScreen(true);
+
+        const enterFullScreen = emittedNTimes(w, 'enter-full-screen', 2);
+        await enterFullScreen;
```
nornagon, line 4143: We should check the final fullscreen state with `w.isFullScreen()`

nornagon (approved):
ship it

MarshallOfSound (approved):

zcbenz (approved):

PR #25476, opened 2020-09-16

zcbenz:
Backport of #25448

See that PR for details.

Notes: Fix `ready-to-show` event not emitted on some machines.


codebytere (approved):

killghost:
This bug still exists in electron 13, when can it be fixed? grateful

Issue #25465, opened 2020-09-15
TouchBar: emoji button appears and cannot be removed


bpasero:
### Issue Details

* **Electron Version:**
  * `9.x`
* **Operating System:**
  * `macOS 10.15.x`
* **Last Known Working Electron version:**
  * `7.x`

### Expected Behavior
All touchbar buttons can be controlled by the app.

### Actual Behavior
An emoji picker appears on the touchbar that cannot be removed taking valuable space.

### To Reproduce
Start Electron Fiddle on a macBook with touchbar.

bpasero:
I just verified that this emoji button does not appear with Electron 7. It seems to be showing up for any `<input type="textarea">` element.

codebytere:
@bpasero looks like that happens here: https://source.chromium.org/chromium/chromium/src/+/master:content/app_shim_remote_cocoa/render_widget_host_view_cocoa.mm;l=2219-2238?q=emoji%20touchbar&ss=chromium

I'll try to figure out how to get rid of it or make it opt-in 🤔

bpasero:
Thanks a lot for the prompt PR!

deepak1556:
This has been handled in our fork as a patch to bring back https://github.com/microsoft/vscode/issues/106541#issuecomment-693433419, hence closing.

PR #25479, opened 2020-09-16
feat: allow disabling touchBar text input emoji picker

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/25465.

In https://chromium-review.googlesource.com/c/chromium/src/+/1953663, Chromium made it such that an emoji picker was included on the touch bar for most web text fields. This was added in a non-removable fashion and as @bpasero notes in https://github.com/electron/electron/issues/25465 takes up valuable space. As such, we should make it such that the emoji picker is an opt-out feature that can be removed by developers looking to maximize touchbar space.

Tested with https://gist.github.com/0c443233a57982129eb02a8dbe72272c.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Allowed for opt-out of the emoji picker placed in the touchBar on macOS in most web fields.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -192,6 +192,8 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
     `false`.
   * `disableAutoHideCursor` Boolean (optional) - Whether to hide cursor when typing.
     Default is `false`.
+  * `disableTouchBarEmojiPicker`  Boolean (optional) - Whether to automatically add an emoji picker to the touchBar
+    when inside a text input. Default is `false`.
   * `autoHideMenuBar` Boolean (optional) - Auto hide the menu bar unless the `Alt`
     key is pressed. Default is `false`.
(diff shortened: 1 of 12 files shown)


MarshallOfSound (commented):
Does the setter actually work instantly, or does it work the next time you happen to navigate into a text area?

codebytere:
@MarshallOfSound the next time you navigate into a text area - you need to focus out and then back in again for it to take effect

MarshallOfSound:
It should be possible to force a regeneration of the touch bar.  Our current TB API does and it would be better if the setter actually acted instantly

deepak1556:
Actually the root issue here is different https://github.com/microsoft/vscode/issues/106541#issuecomment-693433419 where the previous fix was reverted in upstream, we are currently following up on that. Fixing that would not require this api

codebytere:
Closing as this seems no longer necessary. Can reopen if needed.

PR #25493, opened 2020-09-16
feat: read/write Toast Activator CLSID in shortcuts

bitdisaster:
#### Description of Change
PArticipating in Action Center on Windows requires a Toast Activaotr CLSID in shortcuts. This PR adds support to read and write the CLSID in shortcuts on Windows. 

#### Checklist

- [X] `npm test` passes

#### Release Notes

Notes: added read/write support for Toast Activator CLSID in shortcuts


spec/api-shell-spec.js
```diff
@@ -76,7 +77,8 @@ describe('shell module', () => {
         args: 'args2',
         appUserModelId: 'appUserModelId2',
         icon: 'icon2',
-        iconIndex: 2
+        iconIndex: 2,
+        toastActivatorClsid: '{C51A3996-CAD9-4934-848B-16285D4A1496}'
```
nornagon, line 81: this doesn't actually check that this option had any effect—if the code broke, this test would still pass. Can we add a check? Perhaps using `ReadShortcutLink`?

shell/common/gin_converters/guid_converter.h
```diff
@@ -7,13 +7,17 @@
...
+#include "base/strings/sys_string_conversions.h"
+#include "base/win/win_util.h"
 #endif
 #include <string>
 
 #include "gin/converter.h"
 
 #if defined(OS_WIN)
 typedef GUID UUID;
+const GUID GUID_NULL = {0, 0, 0, {0, 0, 0, 0, 0, 0, 0, 0}};
```
nornagon, line 20: I think this will work fine and auto-initialize to 0: https://en.cppreference.com/w/cpp/language/zero_initialization
```suggestion
const GUID GUID_NULL;
```

shell/common/gin_converters/guid_converter.h
```diff
@@ -34,6 +38,9 @@ struct Converter<UUID> {
     UUID uid;
 
     if (guid.length() > 0) {
+      if (guid[0] == '{' && guid[guid.length() - 1] == '}') {
```
nornagon, line 41: any particular reason to support these extra optional curlies? should we just require that the raw uuid be passed in?

nornagon (changes requested):

shell/common/gin_converters/guid_converter.h
zcbenz: This behavior is behind [an option](https://reviews.llvm.org/D54604) and it seems that Chromium does not use auto zero initialization:
https://source.chromium.org/chromium/chromium/src/+/master:build/config/compiler/BUILD.gn;l=2514?ss=chromium

It uses pattern initialization to fill integers with 0xAA automatically so uninitialized variables are more liked to cause crashes.

spec/api-shell-spec.js
bitdisaster: It actually does. The 2 lines below write the change to disk and then read it back with a deep equal check.

shell/common/gin_converters/guid_converter.h
bitdisaster: Its very common to have the curly brackets on Windows. Our own `base::win::WStringFromGUID` method adds them. So for consistency, I added support to read them with our without.

shell/common/gin_converters/guid_converter.h
bitdisaster: @nornagon So are we ok to leave the explicit initialization?

nornagon: Will an explicitly-initialized but empty initializer list work?
```suggestion
const GUID GUID_NULL = {};
```

nornagon (approved):

shell/common/gin_converters/guid_converter.h
bitdisaster: Yep it works. Changed!

PR #25501, opened 2020-09-16
fix: call node::Stop on exit

nornagon:
Backport of #25430.

Notes: Fixed an issue that could cause a normally-exiting process to fail with an "illegal access" message and exit code 7.

### DIFF
--- a/shell/app/node_main.cc
+++ b/shell/app/node_main.cc
@@ -153,6 +153,7 @@ int NodeMain(int argc, char* argv[]) {
 
     exit_code = node::EmitExit(env);
     env->set_can_call_into_js(false);
+    node::Stop(env);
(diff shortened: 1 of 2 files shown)


shell/app/node_main.cc
```diff
@@ -153,6 +153,7 @@ int NodeMain(int argc, char* argv[]) {
 
     exit_code = node::EmitExit(env);
     env->set_can_call_into_js(false);
```
zcbenz, line 155: Does this line need to be removed?

nornagon: it's redundant, but it doesn't hurt to remove it.

zcbenz (approved):

Issue #25498, opened 2020-09-16
webContents.print() fails with 'No valid printers available'


andsmedeiros:
### Preflight Checklist

* [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [X] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  10.1.1
* **Operating System:**
  Ubuntu 20.04 + Linux 5.4.0-47-generic
* **Last Known Working Electron version:**
 7.1.2

### Expected Behavior
`webContents.getPrinters()` returns one printer. Besides, the OS provides additional virtual devices.
`webContents.print()` should open the print dialog and show those devices for selection.

### Actual Behavior
`webContents.print()` fails silently; in the print callback, `failureReason` is 'No valid printers available'

### To Reproduce
Just run `webContents.print()`

Electron fiddle: https://gist.github.com/35c5067af8ec104f86bd0cdd4a95b1f5

### Screenshots
![electron-issue](https://user-images.githubusercontent.com/5503002/93387857-3fc9b300-f840-11ea-9335-bf32a3ddea08.png)
![electron-issue](https://user-images.githubusercontent.com/5503002/93387860-40fae000-f840-11ea-938e-09c62f06f06b.gif)

### Additional Information
Upgrading an app built on Electron 7.1.2 that had printing working before

andsmedeiros:
@codebytere thank you so much for the rapid solution! any chance this lands on 10.0.3?

PR #25515, opened 2020-09-17
fix: check printer list when no default printers

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/25498.

It's possible for it to be the case on some operating systems that no default is selected, yet valid printers are available. In this case, we should check for available printers and default to the first available printer of those available on the system. Chromium will fail to print if no device name is provided at all, so this reduces the number of potential failure cases.

Tested with https://gist.github.com/35c5067af8ec104f86bd0cdd4a95b1f5

cc @jkleinsc @deepak1556 @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where an error would be displayed when using `webContents.print()` if no default was set and no device name provided.


zcbenz (approved):

codebytere:
Only failure is arm64 which is unrelated and caused by a CircleCI image upgrade.

PR #25522, opened 2020-09-17
refactor: replace USE_X11 with OS_LINUX for sharing code with Ozone

vially:
#### Description of Change

This pull-request is a *small first step* in trying to bring support for Wayland to Electron. I'm planning to follow up on this with more pull-requests in an effort to upstream [these Wayland patches](https://github.com/electron/electron/compare/10-x-y...hedgepigdaniel:arch-10?expand=1).

However, those patches will need some adjustments in order to make them compatible with the upcoming Chromium changes ([chromium#1085700](https://bugs.chromium.org/p/chromium/issues/detail?id=1085700)) which will soon be merged into Electron (#25488). In particular, it should be taken into consideration that the `USE_X11` and `USE_OZONE` build flags are no longer mutually exclusive.

Also, `USE_X11` is [going away](https://github.com/electron/electron/issues/10915#issuecomment-558515636) so it might be desirable as a long term goal to remove it from Electron too and replace it with the Ozone equivalents.

With that in mind, this pull-request replaces a few instances of `#if defined(USE_X11)` with `#if defined(OS_LINUX)` where it was identified that the guarded code could be shared with Ozone.

The more meaningful changes will come in the upcoming pull-requests but I'll be waiting for [these chromium changes](https://chromium-review.googlesource.com/c/chromium/src/+/2382834) to be pulled into Electron first (#24984, #25488).

#### Checklist

- [x] PR description included and stakeholders cc'd (@zcbenz @hedgepigdaniel)
- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/browser_linux.cc
+++ b/shell/browser/browser_linux.cc
@@ -15,7 +15,7 @@
 #include "shell/browser/window_list.h"
 #include "shell/common/application_info.h"
 
-#if defined(USE_X11)
+#if defined(OS_LINUX)
 #include "shell/browser/linux/unity_service.h"
 #include "ui/gtk/gtk_util.h"
 #endif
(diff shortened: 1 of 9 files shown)


sdimkov (approved):
It's a great start for Wayland support

zcbenz (approved):

PR #25526, opened 2020-09-18
refactor: remove native_browser_view web contents get function

mlaurencin:
#### Description of Change
This PR refactors the code by eliminating an unnecessary helper method. The method NativeBrowserView::GetWebContents has only one call site in the system, so to simplify the code, this helper was removed and the implementation was moved to the call site.

cc @codebytere 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: None


shell/browser/native_browser_view_mac.mm
```diff
@@ -232,10 +232,10 @@ - (BOOL)mouseDownCanMoveWindow {
 void NativeBrowserViewMac::UpdateDraggableRegions(
     const std::vector<gfx::Rect>& drag_exclude_rects) {
   auto* iwc_view = GetInspectableWebContentsView();
-  auto* web_contents = GetWebContents();
+  auto* web_contents = inspectable_web_contents_->GetWebContents();
```
codebytere, line 235: since we know that there's a possibility `inspectable_web_contents_` is `nullptr`, we need to check that it's not before calling `GetWebContents()`

codebytere (changes requested):

shell/browser/native_browser_view_mac.mm
codebytere: @mlaurencin -

> `auto* web_contents = inspectable_web_contents_->GetWebContents();`

as it's written right now makes it such that if `inspectable_web_contents_` is `nullptr`, there will still be a crash. This is because calling a function on a `nullptr` is undefined behavior, but in this case an access violation as it would be trying to access memory which is no longer allocated. We need to first check `> `inspectable_web_contents_ != nullptr` _before_ trying to call a member function on it.

shell/browser/native_browser_view_mac.mm
```diff
@@ -232,10 +232,12 @@ - (BOOL)mouseDownCanMoveWindow {
 void NativeBrowserViewMac::UpdateDraggableRegions(
     const std::vector<gfx::Rect>& drag_exclude_rects) {
   auto* iwc_view = GetInspectableWebContentsView();
-  auto* web_contents = GetWebContents();
-  if (!(iwc_view && web_contents))
+  if (!iwc_view)
```
codebytere, line 235: This would work, but for readability's sake:

```cpp
if (!inspectable_web_contents_)
     return;
auto* web_contents = inspectable_web_contents_->GetWebContents();
auto* iwc_view = GetInspectableWebContentsView();
```

lets us do fewer checks and is more readable :)

mlaurencin: ok got it

codebytere (approved):

zcbenz (approved):

PR #25531, opened 2020-09-18
fix: allow ClientRequest responses to be throttled

### LINKED ISSUE #25130: ClientRequest responses cannot be throttled
### Preflight Checklist

### Issue Details
I'm trying to throttle the download speeds and the expected method of throttling the stream does not work. In my example, I use `stream-throttle` to reduce my download speed to 1 KB/s. The file gets written out at this speed - but when I look at network tab of the activity monitor I see the download complete as fast as possible. 

I looked into why that is, and I see that the IncomingMessage class is implementing the Readable stream property correctly. However, its buffer will get filled without stopping by the `SimpleURLLoaderWrapper`. There doesn't seem to be a way for the IncomingMessage class to message back that it is full or that the request should pause. 

I can use a node library like https or request and the download will be throttled correctly. 

* **Electron Version:**
  * 9.2.0
* **Operating System:**
  * macOS 10.14.6

### Expected Behavior
When the IncomingMessage is full (`this._shouldPush = false`), the underlying networking should be throttled.

### Actual Behavior
The incomingMessage `_data` buffer continues to fill up and networking continues. I've gotten it as high as 1GB with it slowly streaming out. 

### To Reproduce

Here's my example code that downloads a 100MB files and attempts to write it to disk a 1KB/s. 
https://gist.github.com/burnhamup/fcb9269f294382d3bdc8fa66cb8655df
```
const exampleFunction = async () => {
  const {net} = require('electron');
  const {Throttle} = require('stream-throttle');
  const path = require('path');
  const fs = require('fs');

  const url = 'http://ipv4.download.thinkbroadband.com/100MB.zip';
  const request = net.request(url);
  request.end();
  const response = await new Promise(resolve => {
    request.on('response', resolve);
  });

  const throttle = new Throttle({rate: 1000}); // 1 KB/s

  const downloadLocation = path.join(app.getPath('downloads'), 'example.zip');
  const filestream = fs.createWriteStream(downloadLocation);

  response.pipe(throttle);
  throttle.pipe(filestream);
}
```

VerteDinde:
#### Description of Change

Closes #25130.

Previously, when making a client request, the IncomingMessage class was implementing the Readable stream property correctly. However, its buffer will get filled without stopping by the SimpleURLLoaderWrapper. 

This PR passes up a callback from the SimpleURLLoaderWrapper layer to the IncomingMessage class. This callback allows the IncomingMessage class to properly stop the readable stream when it is full and needs the request to pause, and then resume the stream when appropriate.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where net.request would continue downloading data even when the consuming stream was throttled.

### DIFF
--- a/shell/browser/api/electron_api_url_loader.cc
+++ b/shell/browser/api/electron_api_url_loader.cc
@@ -427,8 +427,7 @@ void SimpleURLLoaderWrapper::OnDataReceived(base::StringPiece string_piece,
   auto array_buffer = v8::ArrayBuffer::New(isolate, string_piece.size());
   auto backing_store = array_buffer->GetBackingStore();
   memcpy(backing_store->Data(), string_piece.data(), string_piece.size());
-  Emit("data", array_buffer);
-  std::move(resume).Run();
+  Emit("data", array_buffer, base::AdaptCallbackForRepeating(std::move(resume)));
 }
 
 void SimpleURLLoaderWrapper::OnComplete(bool success) {
--- a/shell/browser/api/electron_api_url_loader.cc
+++ b/shell/browser/api/electron_api_url_loader.cc
@@ -427,7 +427,8 @@ void SimpleURLLoaderWrapper::OnDataReceived(base::StringPiece string_piece,
   auto array_buffer = v8::ArrayBuffer::New(isolate, string_piece.size());
   auto backing_store = array_buffer->GetBackingStore();
   memcpy(backing_store->Data(), string_piece.data(), string_piece.size());
-  Emit("data", array_buffer, base::AdaptCallbackForRepeating(std::move(resume)));
+  Emit("data", array_buffer,
+       base::AdaptCallbackForRepeating(std::move(resume)));
 }
 
 void SimpleURLLoaderWrapper::OnComplete(bool success) {
--- a/spec-main/api-net-spec.ts
+++ b/spec-main/api-net-spec.ts
@@ -1412,8 +1412,8 @@ describe('net module', () => {
       const urlRequest = net.request(serverUrl);
       urlRequest.on('response', () => {});
       urlRequest.end();
-      await delay(1000);
-      expect(numChunksSent).to.be.at.most(10);
+      await delay(2000);
+      expect(numChunksSent).to.be.at.most(20);
     });
   });
 
--- a/spec-main/api-net-spec.ts
+++ b/spec-main/api-net-spec.ts
@@ -4,7 +4,7 @@ import * as http from 'http';
 import * as url from 'url';
 import { AddressInfo, Socket } from 'net';
 import { emittedOnce } from './events-helpers';
-import { defer } from './spec-helpers';
+import { defer, delay } from './spec-helpers';
 
 const kOneKiloByte = 1024;
 const kOneMegaByte = kOneKiloByte * kOneKiloByte;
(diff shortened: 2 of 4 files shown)


spec-main/api-net-spec.ts
```diff
@@ -1393,6 +1393,29 @@ describe('net module', () => {
...
+      let numChunksSent = 0;
+      const serverUrl = await respondOnce.toSingleURL((request, response) => {
+        const data = randomString(kOneMegaByte);
+        const encoding = 'utf8';
+        const write = () => {
+          let ok = true;
+          do {
+            numChunksSent++;
+            if (numChunksSent > 3000) return;
+            ok = response.write(data, encoding);
```
nornagon, line 1407:
```suggestion
            ok = response.write(data);
```

spec-main/api-net-spec.ts
```diff
@@ -1393,6 +1393,29 @@ describe('net module', () => {
       await collectStreamBody(nodeResponse);
       expect(nodeRequestProcessed).to.equal(true);
     });
+
+    it('should correctly throttle an incoming stream', async () => {
+      let numChunksSent = 0;
+      const serverUrl = await respondOnce.toSingleURL((request, response) => {
+        const data = randomString(kOneMegaByte);
+        const encoding = 'utf8';
```
nornagon, line 1401: I think we can skip the encoding as it will be utf8 by default (and anyway we don't really care about the encoding for this test)
```suggestion
```

spec-main/api-net-spec.ts
```diff
@@ -1393,6 +1393,29 @@ describe('net module', () => {
...
+        const write = () => {
+          let ok = true;
+          do {
+            numChunksSent++;
+            if (numChunksSent > 3000) return;
```
nornagon, line 1406: I think we can turn this down a bit.

```suggestion
            if (numChunksSent > 30) return;
```

nornagon (approved):
looks good to me!

nornagon:
I'd rewrite the release notes a little to not mention internal C++ class names. Maybe something like "Fixed an issue where net.request would continue downloading data even when the consuming stream was throttled"?

VerteDinde:
@nornagon Definitely! Will make that change now

nornagon:
Interesting! It looks like this test is failing on Linux, even though it's succeeding on Mac. I wonder if the kernel buffers are bigger on our CI machine?

I just checked on my Ubuntu VM and I found:

```
$ cat /proc/sys/net/ipv4/tcp_rmem
4096    131072  6291456
$ cat /proc/sys/net/ipv4/tcp_wmem
4096    16384   4194304
```

these three numbers are min/default/max size of the kernel's read and write buffers, respectively. The max size there is 6MB in the read buffer and 4MB in the write buffer, for a total of 10MB, which is roughly equal to the 9 chunks that the server successfully manages to send.

I think it would make sense to solve this by changing the expectation in the test to be <= 10 chunks, and increase the timeout to 1 second (just to be sure that if it was going to send, like, 100 chunks, then we'd definitely get >10). It's not a perfect test—I wish I could think of a way to test this without a `setTimeout`—but it'll at least pass.

shell/browser/api/electron_api_url_loader.cc
```diff
@@ -427,8 +427,7 @@ void SimpleURLLoaderWrapper::OnDataReceived(base::StringPiece string_piece,
   auto array_buffer = v8::ArrayBuffer::New(isolate, string_piece.size());
   auto backing_store = array_buffer->GetBackingStore();
   memcpy(backing_store->Data(), string_piece.data(), string_piece.size());
-  Emit("data", array_buffer);
-  std::move(resume).Run();
+  Emit("data", array_buffer, base::AdaptCallbackForRepeating(std::move(resume)));
```
zcbenz, line 430: Lint is failing:

```patch
--- a/shell/browser/api/electron_api_url_loader.cc
+++ b/shell/browser/api/electron_api_url_loader.cc
@@ -427,7 +427,8 @@
   auto array_buffer = v8::ArrayBuffer::New(isolate, string_piece.size());
   auto backing_store = array_buffer->GetBackingStore();
   memcpy(backing_store->Data(), string_piece.data(), string_piece.size());
-  Emit("data", array_buffer, base::AdaptCallbackForRepeating(std::move(resume)));
+  Emit("data", array_buffer,
+       base::AdaptCallbackForRepeating(std::move(resume)));
 }
 
 void SimpleURLLoaderWrapper::OnComplete(bool success) {
```

VerteDinde: @zcbenz Thanks so much! Ran a lint check and made sure this was fixed 🙇‍♀️

codebytere:
@VerteDinde looks like this is failing:

> net module IncomingMessage API should correctly throttle an incoming stream

codebytere (approved):

zcbenz (approved):

Issue #25491, opened 2020-09-16
Crash in RedrawTrafficLights


codebytere:
First seen here: https://app.circleci.com/pipelines/github/electron/electron/29873/workflows/923c3f7e-6bea-47bf-93af-a19497f42cf0/jobs/658750

```
Received signal 11 SEGV_MAPERR 07fba4266cd8
0   Electron Framework                  0x000000010a75eeb9 base::debug::CollectStackTrace(void**, unsigned long) + 9
1   Electron Framework                  0x000000010a659533 base::debug::StackTrace::StackTrace() + 19
2   Electron Framework                  0x000000010a75ed81 base::debug::(anonymous namespace)::StackDumpSignalHandler(int, __siginfo*, void*) + 2385
3   libsystem_platform.dylib            0x00007fff6a65f5fd _sigtramp + 29
4   ???                                 0xf6a3c00070001002 0x0 + 17772259662645694466
5   Electron Framework                  0x0000000104faaf35 electron::NativeWindowMac::RedrawTrafficLights() + 53
6   Electron Framework                  0x000000010a6eeefb base::TaskAnnotator::RunTask(char const*, base::PendingTask*) + 411
7   Electron Framework                  0x000000010a70dee7 base::sequence_manager::internal::ThreadControllerWithMessagePumpImpl::DoWorkImpl(base::sequence_manager::LazyNow*) + 519
8   Electron Framework                  0x000000010a70db58 base::sequence_manager::internal::ThreadControllerWithMessagePumpImpl::DoWork() + 168
9   Electron Framework                  0x000000010a77f651 base::MessagePumpCFRunLoopBase::RunWork() + 65
10  Electron Framework                  0x000000010a7746e2 base::mac::CallWithEHFrame(void () block_pointer) + 10
11  Electron Framework                  0x000000010a77f01f base::MessagePumpCFRunLoopBase::RunWorkSource(void*) + 63
12  CoreFoundation                      0x00007fff30468a54 __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 17
13  CoreFoundation                      0x00007fff304689f3 __CFRunLoopDoSource0 + 103
14  CoreFoundation                      0x00007fff3046880d __CFRunLoopDoSources0 + 209
15  CoreFoundation                      0x00007fff30467529 __CFRunLoopRun + 937
16  CoreFoundation                      0x00007fff30466b23 CFRunLoopRunSpecific + 466
[...]

PR #25554, opened 2020-09-21
fix: unsubscribe from observers when window is closing

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/25491.

Window related observers should be cleaned up when window is closing, instead of in the destructor of `NativeWindow`, otherwise crashes may happen due to invalid windows, as the destructor can be called much later after the window gets closed.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: chore


deepak1556 (approved):
Is this only needed to be backported for `11-x-y` branch ?

codebytere (approved):

zcbenz:
> Is this only needed to be backported for `11-x-y` branch ?

I only saw the CI crash on >= 1.x, but backporting to older branches should be fine.

PR #25557, opened 2020-09-21
build: Wrap bundles using webpack

juliekoubova:
#### Description of Change
Moved the `run-compiler.js` and `get-outputs.js` logic directly into webpack. This way, the source mappings for bundles—when enabled—are not offset by the number of lines of the try...catch & profiling wrappers.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


juliekoubova:
also, we could probably use production mode instead of development with `TerserPlugin` and manually disabled source mapping. unless there's a reason for that?

MarshallOfSound:
> also, we could probably use production mode instead of development with TerserPlugin and manually disabled source mapping. unless there's a reason for that?

I honestly don't remember why I did that, I think it's because initially we didn't have the minifier.  Then we relealized it made everything faster 😆 

As long as we can keep our custom terser settings then `production` mode might make sense here

juliekoubova:
> As long as we can keep our custom terser settings then production mode might make sense here

I'd rather do it in a separate PR, it does add some more things than just the already-overridden settings:

> Sets process.env.NODE_ENV on DefinePlugin to value production. Enables FlagDependencyUsagePlugin,
> FlagIncludedChunksPlugin, ModuleConcatenationPlugin, NoEmitOnErrorsPlugin, OccurrenceOrderPlugin,
> SideEffectsFlagPlugin and TerserPlugin.
https://webpack.js.org/configuration/mode/

Also, is there a way to re-trigger the build-mac workflow? I'm think it might have just randomly timed out?

MarshallOfSound (approved):

PR #25561, opened 2020-09-21
build: update nan dep for tests

MarshallOfSound:
Gonna add this to `roller` so we don't fall behind, no major changes incoming, just some deprecation warnings fixed.

Notes: no-notes


codebytere (approved):

zcbenz (changes requested):
The `linux-x64-testing-nan` tests keep failing and it does not seem to be flaky.

MarshallOfSound:
This will have to land as part of an upcoming chromium roll so closing this out while we deal with that

PR #25563, opened 2020-09-21
fix: order menu items before filtering excess separators

MarshallOfSound:
Previously we removed trailing separators before we re-ordered menu items into positions where the separators would not have been removed.  This flips the order of those operations to fix this issue.

Notes: Excess menu item separators are now only removed after item re-ordering is done to avoid accidental removal of needed separators

### DIFF
--- a/lib/browser/api/menu-utils.ts
+++ b/lib/browser/api/menu-utils.ts
@@ -157,9 +157,12 @@ function sortGroups<T> (groups: {id?: T}[][]) {
   return sortedGroupIndexes.map(i => groups[i]);
 }
 
-export function sortMenuItems (menuItems: {type?: string, id?: string}[]) {
-  const isSeparator = (item: {type?: string}) => item.type === 'separator';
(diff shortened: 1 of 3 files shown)


lib/browser/api/menu-utils.ts
```diff
@@ -158,8 +158,8 @@ function sortGroups<T> (groups: {id?: T}[][]) {
 }
 
 export function sortMenuItems (menuItems: {type?: string, id?: string}[]) {
-  const isSeparator = (item: {type?: string}) => item.type === 'separator';
-  const separators = menuItems.filter(i => i.type === 'separator');
+  const isSeparator = (i: any) => i.type === 'separator' && !i.before && !i.after && !i.beforeGroupContaining && !i.afterGroupContaining;
```
codebytere, line 161: we can probably type this properly, no?

MarshallOfSound: Done

zcbenz (approved):

codebytere (approved):

PR #25580, opened 2020-09-23
feat: add webContents.forcefullyCrashRenderer() to forcefully terminate a renderer process

MarshallOfSound:
Currently the only way to recover a hung renderer is to close the window or manually call `kill ${webContents.getProcessId()}` neither of which is ideal.  This adds a simple new method that aligns with what Chrome does to make killing a webConents in a hung state super easy.

Notes: Added `webContents.forcefullyCrashRenderer()` to forcefully terminate a renderer process to assist with recovering a hung renderer.

### DIFF
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -998,6 +998,34 @@ Navigates to the specified offset from the "current entry".
 
 Returns `Boolean` - Whether the renderer process has crashed.
 
+#### `contents.crashProcess()`
+
+Forcefully terminates the renderer process that is currently hosting this
+`webContents`. This will cause the `render-process-gone` event to be emitted
+with the `reason=killed`. Please note that some webContents share renderer
+processes and therefore calling this method may also crash the host process
+for other webContents as well.
+
+Calling `reload()` immediately after calling this
+method will force the reload to occur in a new process. This should be used
+when this process is unstable or unusable, for instance in order to recover
+from the `unresponsive` event.
+
+```js
+contents.on('unresponsive', async () => {
+  const { response } = await dialog.showMessageBox({
+    message: 'App X has become unresponsive',
+    title: 'Do you want to try forcefully reloading the app?',
+    buttons: ['OK', 'Cancel'],
+    cancelId: 1
+  })
+  if (response === 0) {
+    contents.crashProcess()
+    contents.reload()
+  }
+})
+```
+
 #### `contents.setUserAgent(userAgent)`
 
 * `userAgent` String
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -998,11 +998,11 @@ Navigates to the specified offset from the "current entry".
 
 Returns `Boolean` - Whether the renderer process has crashed.
 
-#### `contents.crashProcess()`
+#### `contents.forcefullyCrashRenderer()`
 
 Forcefully terminates the renderer process that is currently hosting this
 `webContents`. This will cause the `render-process-gone` event to be emitted
-with the `reason=killed`. Please note that some webContents share renderer
+with the `reason=killed || reason=crashed`. Please note that some webContents share renderer
 processes and therefore calling this method may also crash the host process
 for other webContents as well.
 
@@ -1020,7 +1020,7 @@ contents.on('unresponsive', async () => {
     cancelId: 1
   })
   if (response === 0) {
-    contents.crashProcess()
+    contents.forcefullyCrashRenderer()
     contents.reload()
   }
 })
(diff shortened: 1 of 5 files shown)


shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1757,6 +1758,30 @@ bool WebContents::IsCrashed() const {
...
+  if (rph) {
+#if defined(OS_LINUX) || defined(OS_CHROMEOS)
+    // A generic |CrashDumpHungChildProcess()| is not implemented for Linux.
+    // Instead we send an explicit IPC to crash on the renderer's IO thread.
+    rph->ForceCrash();
+#else
+    // Try to generate a crash report for the hung process.
+    CrashDumpHungChildProcess(rph->GetProcess().Handle());
+    rph->Shutdown(content::RESULT_CODE_HUNG);
+#endif
```
ckerr, line 1781: This implementation is consistent with what's in hung_renderer_view.cc and smells OK in terms of working.

I am wondering though what happens to these crash reports? If these get stored on the user's machine, does it make sense to have API that generates a lot of data files when working as intended?

These crash reports could be a boon for well-maintained apps that have people who mine these reports, but would not be great for less-supported apps iff the crash reports just keep piling up on the user's machine.

docs/api/web-contents.md
```diff
@@ -998,6 +998,34 @@ Navigates to the specified offset from the "current entry".
 
 Returns `Boolean` - Whether the renderer process has crashed.
 
+#### `contents.crashProcess()`
```
ckerr, line 1001: _(nit)_ If we made this method more descriptive about _which_ process is being crashed, the new n ame would be closer to its Chromium counterpart, e.g. `contents.forceCrashRenderer()`

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2915,6 +2940,7 @@ v8::Local<v8::ObjectTemplate> WebContents::FillObjectTemplate(
       .SetMethod("_goForward", &WebContents::GoForward)
       .SetMethod("_goToOffset", &WebContents::GoToOffset)
       .SetMethod("isCrashed", &WebContents::IsCrashed)
+      .SetMethod("crashProcess", &WebContents::CrashProcess)
```
DongHyunSong, line 2943: nit: could you keep this in alphabetical order?

shell/browser/api/electron_api_web_contents.cc
nornagon: @ckerr for apps that haven't configured crashReporter, no dumps will be generated when crashes occur. https://source.chromium.org/chromium/chromium/src/+/master:components/crash/core/app/crashpad_mac.mm;l=80-81;drc=bc576dbe56e7f97ceb1973c6cd4953afca86f175 the database is only present if `crashReporter.start()` has been called.

spec-main/api-web-contents-spec.ts
```diff
@@ -1245,6 +1245,35 @@ describe('webContents module', () => {
     });
   });
 
+  describe('crash behavior', () => {
+    let w: BrowserWindow;
+    beforeEach(async () => {
+      w = new BrowserWindow({ show: false, webPreferences: { nodeIntegration: true } });
+      await w.loadURL('data:text/html,<h1>Hello, World!</h1>');
```
nornagon, line 1252:
```suggestion
      await w.loadURL('about:blank');
```

should be fine yeah?

spec-main/api-web-contents-spec.ts
```diff
@@ -1245,6 +1245,35 @@ describe('webContents module', () => {
     });
   });
 
+  describe('crash behavior', () => {
+    let w: BrowserWindow;
+    beforeEach(async () => {
+      w = new BrowserWindow({ show: false, webPreferences: { nodeIntegration: true } });
```
nornagon, line 1251: is `nodeIntegration` relevant here?

Perhaps it's worth also testing `sandbox`?

nornagon (approved):

bpasero:
@MarshallOfSound may I ask how and when this method should be used? In VSCode we have always been listening to the `render-process-gone` and `unresponsive` event where we show the user a dialog offering to either "Reload" or "Close" the window. When the user picks "Reload", we do an attempt to set a new URL to the window and for the "Close" case, we actually call `BrowserWindow.destroy()`. 

image

I am trying to understand if I should now use `webContents.forcefullyCrashRenderer` in any of these cases to increase the chances of getting "Reload" or "Close" to work as advertised?

//cc @deepak1556

deepak1556:
@bpasero this is useful in sandbox mode when a reload doesn't necessarily spawn a new process, where you can call this method and then reload to safely put the unresponsive webContents in a new process.

PR #25587, opened 2020-09-23
chore: remove custom Node.js debugger

codebytere:
#### Description of Change

Removes our custom Node.js debugger and related patches. When we have `nodeIntegration` enabled in the renderer process, Blink creates the V8 inspector themselves, so that means two of them will be created, leading to a fun set of crashes. This PR adjusts a patch to ensure no inspector is created in the renderer process. 

This PR also has the side effect of making the `--inspect-brk-node` flag work again 🎉  

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


patches/node/fix_allow_preventing_initializeinspector_in_env.patch
```diff
@@ -1,18 +1,17 @@
...
 https://github.com/nodejs/node/commit/8c5ad1392f30cfe6b107e9bd85f4cb918ba04aab
 made it such that env->InitializeInspector was called in CreateEnvironment
-no matter what, which creates an issue for embedders seeking to manage
-the InspectorAgent themselves as Electron does. This adds a new
+no matter what, which creates an issue for Electron, as the V8 inspector
+already exists in the renderer process and therefore we only want to
+initialize it in the browser process. This adds a new
 EnvironmentFlags option which allows preventing that invocation.
 
-This will be upstreamed.
```
nornagon, line 12: are we no longer intending to upstream this?

nornagon (approved):

patches/node/fix_allow_preventing_initializeinspector_in_env.patch
codebytere: oh we are, i have a PR opened already! my bad

PR #25592, opened 2020-09-23
chore: improve renderer crash logging

miniak:
#### Description of Change
Follow up to #25317

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/api/web-contents.ts
+++ b/lib/browser/api/web-contents.ts
@@ -462,8 +462,7 @@ const addReturnValueToEvent = (event: any) => {
 };
 
 const loggingEnabled = () => {
-  return Object.prototype.hasOwnProperty.call(process.env, 'ELECTRON_ENABLE_LOGGING') ||
-    process.argv.some(arg => arg.toLowerCase().startsWith('--enable-logging'));
+  return process.env.ELECTRON_ENABLE_LOGGING || app.commandLine.hasSwitch('enable-logging');
 };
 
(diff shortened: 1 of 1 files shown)


lib/browser/api/web-contents.ts
```diff
@@ -462,8 +462,7 @@ const addReturnValueToEvent = (event: any) => {
 };
 
 const loggingEnabled = () => {
-  return Object.prototype.hasOwnProperty.call(process.env, 'ELECTRON_ENABLE_LOGGING') ||
-    process.argv.some(arg => arg.toLowerCase().startsWith('--enable-logging'));
+  return process.env.ELECTRON_ENABLE_LOGGING || app.commandLine.hasSwitch('enable-logging');
```
nornagon, line 465: teeeeechnically the logic for this in `base::Environment` is different. Ideally we'd just query this info directly from the C++ which actually controls logging, rather than trying to reimplement this.

Things that enable logging in the C++ without tripping this test, for example:
```
ELECTRON_ENABLE_LOGGING=0
electron_enable_logging=1

# only on non-Windows platforms:
ELECTRON_ENABLE_LOGGING=
```

I don't think it's very important to match exactly, but JFYI.

nornagon (approved):

codebytere (approved):

alexeykuzmin (approved):

PR #25606, opened 2020-09-23
fix: check printer list when no default printers

codebytere:
Backport of #25515

See that PR for details.

Notes: Fixed an issue where an error would be displayed when using `webContents.print()` if no default was set and no device name provided.


jkleinsc (approved):

alex-drocks:
Thanks for the fix!

PR #25609, opened 2020-09-23
fix: prevent destroyed view references from causing crashes

mlaurencin:
#### Description of Change

Backport of https://github.com/electron/electron/pull/25411

@codebytere 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed crashes caused by attempting to modify destroyed views

### DIFF
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -1247,8 +1247,11 @@ void NativeWindowMac::AddBrowserView(NativeBrowserView* view) {
   }
 
   add_browser_view(view);
-  auto* native_view =
-      view->GetInspectableWebContentsView()->GetNativeView().GetNativeNSView();
+  auto* iwc_view = view->GetInspectableWebContentsView();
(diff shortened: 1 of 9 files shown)


shell/browser/native_window_mac.mm
```diff
@@ -1247,8 +1247,9 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
   }
 
   add_browser_view(view);
-  auto* native_view =
-      view->GetInspectableWebContentsView()->GetNativeView().GetNativeNSView();
+  auto* iwc_view = view->GetInspectableWebContentsView() if (!iwc_view) return;
```
jkleinsc, line 1250:
```suggestion
  auto* iwc_view = view->GetInspectableWebContentsView();
  if (!iwc_view) 
    return;
```

jkleinsc (changes requested):
Looks like there are some legitimate test failures happening

codebytere (approved):

PR #25611, opened 2020-09-23
refactor: use chromium x11 error handler

vially:
#### Description of Change

Chromium recently extracted the X11 error handler into a separate file ([chromium#1093008](https://bugs.chromium.org/p/chromium/issues/detail?id=1093008)) in order to reuse that logic in Ozone/X11. However, this logic is also **very similar** to the X11 error handler implemented in Electron (originally inspired by the chromium implementation: https://github.com/electron-archive/brightray/pull/149).

This pull-request attempts to cleanup this code duplication by reusing chromium's X11 error handler in Electron. However, the Electron implementation does contain some changes which are not found in the chromium's implementation (e.g.: https://github.com/electron/electron/commit/b25175a19acb4fd2bb1491585307801674a2f2f3). This means that **if this pull-request is merged in its current form those changes will be lost**.

I'm not familiar enough with the Electron code base to understand whether these changes are still needed, but if they are (very likely) I'll try to find a way to port them over.

The reason I'm touching this code and trying to clean it up is because I'm working on a set of patches to enable Ozone/Wayland support in Electron (more details in #25522) and it would be nice if it was possible to remove some of the X11 specific code in the process.

### Warning

This pull-request contains subtle logic changes for which I don't fully understand the potential consequences. They also happen to be in areas which are not very easy to test (e.g.: X server errors while electron is running). **Please review it very carefully.**

#### Checklist

- [x] PR description included and stakeholders cc'd (@zcbenz) 
- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/electron_browser_main_parts.cc
+++ b/shell/browser/electron_browser_main_parts.cc
@@ -64,6 +64,7 @@
 #include "base/environment.h"
 #include "base/nix/xdg_util.h"
 #include "base/threading/thread_task_runner_handle.h"
+#include "ui/base/x/x11_error_handler.h"
 #include "ui/base/x/x11_util.h"
 #include "ui/events/devices/x11/touch_factory_x11.h"
 #include "ui/gfx/color_utils.h"
@@ -147,14 +148,7 @@ base::string16 MediaStringProvider(media::MessageId id) {
   }
 }
 
-#if defined(USE_X11)
-// Indicates that we're currently responding to an IO error (by shutting down).
-bool g_in_x11_io_error_handler = false;
-
-// Number of seconds to wait for UI thread to get an IO error if we get it on
-// the background thread.
-const int kWaitForUIThreadSeconds = 10;
-
+#if defined(OS_LINUX)
 void OverrideLinuxAppDataPath() {
   base::FilePath path;
   if (base::PathService::Get(DIR_APP_DATA, &path))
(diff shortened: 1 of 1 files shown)


nornagon:
cc @ckerr who has some experience in this area I think

zcbenz (approved):
Looks good to me.

vially:
I've force-pushed some changes to fix a [linter error](https://app.circleci.com/pipelines/github/electron/electron/30213/workflows/470496a0-a37c-4892-87fb-f780f6c68ca7/jobs/666011) caused by a duplicate import of `base/run_loop.h`.

ckerr (approved):
OK. As I read through this and look at the history of the x11 changes in e_b_main_parts and the new upstream changes -- this patch looks reasonable.

As for whether or not we need to keep the code shear... b25175a19acb4fd2bb1491585307801674a2f2f3 came during a Chromium roll so there's not a lot of context captured on GitHub about what prompted that. Since @zcbenz made that change _and_ approved this PR, that's encouraging :smile: 

When in doubt, we try to stick to upstream as closely as possible. There may be a blocker but I'm not seeing it in this PR, and so I'd vote for sticking to upstream, taking this into master, marking as do-not-backport, and seeing what surfaces. Unfortunately we don't have good CI specific to X11 error handling, so we may need to find these edge cases manually.

PR #25617, opened 2020-09-24
fix: submenu should be autoreleased

zcbenz:
#### Description of Change

The `submenu` is a strong property so it must be autoreleased before being assigned.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fix memory leak when creating "Services" menu.

### DIFF
--- a/shell/browser/ui/cocoa/electron_menu_controller.mm
+++ b/shell/browser/ui/cocoa/electron_menu_controller.mm
@@ -297,7 +297,7 @@ static base::scoped_nsobject<NSMenu> recentDocumentsMenuSwap_;
 
     [item setTarget:nil];
     [item setAction:nil];
-    NSMenu* submenu = [[NSMenu alloc] initWithTitle:label];
+    NSMenu* submenu = [[[NSMenu alloc] initWithTitle:label] autorelease];
     [item setSubmenu:submenu];
     [NSApp setServicesMenu:submenu];
   } else if (type == electron::ElectronMenuModel::TYPE_SUBMENU &&


codebytere (approved):

codebytere:
@zcbenz there's another place we allocate a new `NSMenu` - should those be autoreleased as well?

https://github.com/electron/electron/blob/7b0a84477ae8ae8817c560cd687e3222087b9d58/shell/browser/ui/cocoa/electron_menu_controller.mm#L80

ckerr (approved):
I'd be welcome another opinion but I _think_ this PR is correct as-is?

There are four places in the file where we alloc a new NSMenu. Two of them (lines 80 and 412) are managed by the `base::scoped_nsobject` wrapper, one (line 211) is already marked for autorelease, and this one (line 300) is the one now being marked.

zcbenz:
> @zcbenz there's another place we allocate a new `NSMenu` - should those be autoreleased as well?

`base::scoped_nsobject` behaves like `std::unique_ptr`, it takes a pointer of ref count 1 and releases it when destructed. So we should not pass an autoreleased pointer to it.

The behaviors of Cocoa APIs depend on the type of properties. For `@property(strong)` (for example `submenu` and `image`), the passed pointer would be retained and released, so we should usually pass an autoreleased pointer to it. For `@property(weak)` (for example `delegate`), nothing would happen to the passed pointer, so it is our responsibility to managed the pointer's lifetime and ensures the pointer outlives the target object.

Anyway the principle is to ensure the ref count of the pointer will eventually turn 0.

Also note that `base::RefCounted` and `GObject` objects are slightly different that, the newly allocated objects have ref count of 0, while newly allocated `NSObject` objects have ref count of 1.

MarshallOfSound (approved):

Issue #25254, opened 2020-09-01
App crashes after printing using webContents.print()


paronix:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 10.0.1
  * 10.1.0
  * 10.1.1
* **Operating System:**
  * Windows 10
* **Last Known Working Electron version:**
  * 9.2.1
  * 10.0.0

### Expected Behavior
Printing is succesful and the application does not crash after printing.

### Actual Behavior
Printing is succesful but immediately after the application closes with Exit status 3221225477

### To Reproduce
Use web Contents.print() in affected versions. See Fiddle at https://gist.github.com/3f840923e139cf11121a46e9bec3f5f9

### Additional Information
It does not seem to affect the V11 Beta's (tested in Electron Fiddle).

bjbk:
Confirmed same for macOS. 
Electron v10.1.1
OS Version:            Mac OS X 10.15.6
Crashed Thread:        0  CrBrowserMain

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000018
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [23587]

Issue #24947, opened 2020-08-12
Refactor printing code based on headless component print manager


deepak1556:
This is based on the discussion from https://bugs.chromium.org/p/chromium/issues/detail?id=1101596 , our current implementation of pdf generation is kind of hack using bits from `//chrome`. 

We have hit the limits of this hack, with oopif enabled our implementation is unable to composite the document with sub frames. Also we maintain considerable amount of patch set in `printing.patch` that cannot be upstreamed.

Let us try to minimize this based on the pointers from upstream.

/cc @codebytere

ckerr:
This is making its way into our roll in https://github.com/electron/electron/pull/33091 via https://chromium-review.googlesource.com/c/chromium/src/+/3111565, which is a precursor to https://chromium-review.googlesource.com/c/chromium/src/+/3082312/ .

I've put in stubs for our silent / job_settings / canceled parameters in https://github.com/electron/electron/pull/33091/commits/e17ada2846bccc6dfdc911623cdaf86143b9106f which unblocks the roll, but a more comprehensive refactor is called for with the kinds of goals described above by @deepak1556 about using upstream code instead of patching it

PR #25622, opened 2020-09-24
fix: crash in printing on Windows

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/25254.

This was crashing because `IDS_UTILITY_PROCESS_PRINTING_SERVICE_NAME` was being incorrectly loaded as an empty string.

cc @nornagon @ckerr @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash in printing on Windows.


deepak1556 (approved):
would be great if we can complete https://github.com/electron/electron/issues/24947  which would allow for some headless testing that can catch these issues early on.

ckerr (approved):
:+1: @deepak1556's suggestion on #24947, but it doesn't block this PR

Issue #24356, opened 2020-06-29
ready-to-show event doesn't fire when opening a window to "about:blank" and writing a small amount of content to it


pushkin-:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** 9.0.4
* **Operating System:** Windows 10
* **Last Known Working Electron version:** not sure

### Expected Behavior
The `ready-to-show` event should fire when opening a window to "about:blank" and writing a small amount of content to it.

### Actual Behavior
`ready-to-show` doesn't fire until triggering another `window.open` call.
`ready-to-show` does fire when writing a larger amount of content however

### To Reproduce
main.js:

```
(async () => {
  const {app, BrowserWindow} = require("electron");

  await app.whenReady()

  const win = new BrowserWindow({webPreferences: { nativeWindowOpen: true }});
  await win.loadURL("https://google.com");
  win.webContents.openDevTools();

  win.webContents.on("new-window", (e, url, frameName, disp, options) => {
    options.show = false;
  });

  app.on("browser-window-created", (e, win) => {
    win.once("ready-to-show", () => {
      win.show();
    });
  });
})();

```

1. `npm start` this app.
2. in the devtools that appears, execute `var w = window.open("about:blank", "a"); w.document.write('<body>a</body>"');`
3. notice the window doesn't appear because the ready-to-show event doesn't fire
4. if you do another `window.open("about:blank", "a")` (needs to be for the same window name), then the ready-to-show fires

If you don't make the `w.document.write` call and just open a window to `"about:blank"`, then the `ready-to-show` event fires appropriately.

[...]

PR #25632, opened 2020-09-25

CezaryKulakowski:
#### Description of Change

When this chromium change landed:
https://chromium-review.googlesource.com/c/chromium/src/+/2215143
event `ready-to-show` is not always emitted. This pull request removes following workaround for the problem:
https://github.com/electron/electron/pull/25448
and fixes the problem the way which ensures that event is emitted exactly at the same moment it would be emitted before problematic chromium change landed. Previous fix was not ideal as with it event `ready-to-show` is emitted only after `did-finish-load` was received what significantly delays moment the window is displayed (also: it may happen that `did-finish-load` is never emitted for certain pages). 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix ready-to-show event not emitted on some machines.

### DIFF
--- a/lib/browser/api/web-contents.ts
+++ b/lib/browser/api/web-contents.ts
@@ -615,6 +615,13 @@ WebContents.prototype._init = function () {
     app.emit('login', event, this, ...args);
   });
 
+  this.on('ready-to-show' as any, (event) => {
+    const owner = this.getOwnerBrowserWindow();
+    if (owner && !owner.isDestroyed()) {
+      owner.emit('ready-to-show');
+    }
+  });
+
   const event = process._linkedBinding('electron_browser_event').createEmpty();
   app.emit('web-contents-created', event, this);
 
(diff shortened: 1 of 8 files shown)


nornagon (approved):

pushkin-:
@CezaryKulakowski Which version of 11 did this land in? 11.0.0? I'm not seeing this in the [release notes](https://www.electronjs.org/releases/stable?page=2#fixes), only the notes for [this PR](https://github.com/electron/electron/pull/25448).

And I have case where ready-to-show isn't firing when I load a page into a BrowserView in the BrowserWindow, but don't load a page in the BrowserWindow itself. Is this what the bug is fixing, or no?

pushkin-:
@CezaryKulakowski @codebytere  this PR seems to have re-introduced [this issue](https://github.com/electron/electron/issues/24356)

killghost:
This bug still exists in electron 13, when can it be fixed? grateful

PR #25675, opened 2020-09-29
refactor: more efficient key string matching

ckerr:
#### Description of Change

This was a lunchtime experiment with constexpr that worked out well enough to PR it. My goal was to make more structured / efficient / readable string matching to replace long `if (str == "foo") ... else (foo == "bar") ...` chains. This adds a StringPiece-to-Value map that's all built at compile-time.

Advantages:
* The map is built at compile time and has negligible overhead -- in fact using this shrunk `keyboard_util.o` by about 25%
* makes code more readable by forcing a structured key/value mapping
* lookups are binary instead of linear

Disadvantages:
* Takes a few lines of setup. The code is cheap but the conceptual overhead is not worth it for short lists.
* In general we don't do _that_ many string lookups.

I like the struct, but I'm undecided on whether this is overkill and would like another opinion. Reviewers welcomed.

CC @nornagon

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/electron_api_session.cc
+++ b/shell/browser/api/electron_api_session.cc
@@ -71,6 +71,7 @@
 #include "shell/common/node_includes.h"
 #include "shell/common/options_switches.h"
 #include "shell/common/process_util.h"
+#include "shell/common/string_lookup.h"
 #include "ui/base/l10n/l10n_util.h"
 
 #if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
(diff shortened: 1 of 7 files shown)


shell/browser/api/electron_api_session.cc
```diff
@@ -120,27 +121,25 @@ struct ClearStorageDataOptions {
 };
 
 uint32_t GetStorageMask(const std::vector<std::string>& storage_types) {
+  static constexpr electron::StringLookup<uint32_t, 9> typemap{{{
```
nornagon, line 124: Having to specify the number of elements is a bit of a bummer, can we infer it?

It's also a bit surprising that we need three sets of curly braces here.

nornagon (commented):
Hm, I'm not sure the conceptual overhead is worth it. it's very clear what if/else does, whereas to understand how this non-standard StringLookup class behaves and performs takes extra mental space. None of the places we're doing this kind of if/else chain are performance-sensitive, I don't think.

shell/browser/api/electron_api_session.cc
ckerr: We [can](https://stackoverflow.com/questions/6114067/how-to-emulate-c-array-initialization-int-arr-e1-e2-e3-behaviou) infer it. It's cleaner in c++17, but it's doable in 14. If that's the make-or-break threshold for this PR then I'll make that change.

The curly braces are (I believe) necessary due to C++'s brace elision rule. std::pair isn't an aggregate, so it needs extra braces to be list-initialized. This holds true for the pairs _and_ the array. I agree it's an  unusual look :)

PR #25681, opened 2020-09-29
refactor: replace a few any-s with proper types

miniak:
#### Description of Change
Improves type safety.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/common/api/deprecate.ts
+++ b/lib/common/api/deprecate.ts
@@ -43,7 +43,7 @@ const deprecate: ElectronInternal.DeprecationUtil = {
     return function (this: any) {
       warn();
       fn.apply(this, arguments);
-    };
+    } as unknown as typeof fn;
   },
 
   // change the name of a function
@@ -52,7 +52,7 @@ const deprecate: ElectronInternal.DeprecationUtil = {
     return function (this: any) {
       warn();
       return fn.apply(this, arguments);
-    };
+    } as unknown as typeof fn;
   },
 
   moveAPI<T extends Function> (fn: T, oldUsage: string, newUsage: string): T {
@@ -60,7 +60,7 @@ const deprecate: ElectronInternal.DeprecationUtil = {
     return function (this: any) {
       warn();
       return fn.apply(this, arguments);
-    } as any;
+    } as unknown as typeof fn;
   },
 
   // change the name of an event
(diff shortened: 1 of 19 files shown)


lib/common/api/deprecate.ts
```diff
@@ -43,7 +43,7 @@ const deprecate: ElectronInternal.DeprecationUtil = {
     return function (this: any) {
       warn();
       fn.apply(this, arguments);
-    };
+    } as any;
```
nornagon, line 46: does this work? same effect i suppose from the outside, but it's a bit better documentation
```suggestion
    } as any as typeof fn;
```

lib/renderer/window-setup.ts
```diff
@@ -255,13 +255,13 @@ export const windowSetup = (
...
     // Make the browser window or guest view emit "new-window" event.
-    (window as any).open = function (url?: string, frameName?: string, features?: string) {
+    window.open = function (url?: string, frameName?: string, features?: string) {
       if (url != null && url !== '') {
         url = resolveURL(url, location.href);
       }
       const guestId = ipcRendererInternal.sendSync('ELECTRON_GUEST_WINDOW_MANAGER_WINDOW_OPEN', url, toString(frameName), toString(features));
       if (guestId != null) {
-        return getOrCreateProxy(guestId);
+        return getOrCreateProxy(guestId) as any as WindowProxy;
```
nornagon, line 264: can this not return `WindowProxy` directly?

miniak:
```
Type 'SafelyBoundBrowserWindowProxy' is missing the following properties from type 'Window': applicationCache, clientInformation, customElements, defaultStatus, and 237 more.ts(2740)
```

lib/common/api/deprecate.ts
```diff
@@ -52,7 +52,7 @@ const deprecate: ElectronInternal.DeprecationUtil = {
     return function (this: any) {
       warn();
       return fn.apply(this, arguments);
-    };
+    } as any;
```
nornagon, line 55: here too?

nornagon (approved):

alexeykuzmin (approved):

PR #25694, opened 2020-09-29
refactor: create IPC_MESSAGES enum for IPC message channels

miniak:
#### Description of Change
Now that almost everything is converted to TypeScript it's quite safe to do this. Eliminates duplicate string constants for internal IPC channels.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes


lib/browser/guest-view-manager.js
```diff
@@ -115,30 +115,30 @@ const createGuest = function (embedder, params) {
   // Dispatch events to embedder.
   const fn = function (event) {
     guest.on(event, function (_, ...args) {
-      sendToEmbedder('ELECTRON_GUEST_VIEW_INTERNAL_DISPATCH_EVENT', event, ...args);
+      sendToEmbedder('GUEST_VIEW_INTERNAL_DISPATCH_EVENT', event, ...args);
```
nornagon, line 118: can these be enums too?

lib/renderer/web-view/guest-view-internal.ts
```diff
@@ -66,18 +67,18 @@ const dispatchEvent = function (
 };
 
 export function registerEvents (webView: WebViewImpl, viewInstanceId: number) {
-  ipcRendererInternal.on(`ELECTRON_GUEST_VIEW_INTERNAL_DESTROY_GUEST-${viewInstanceId}`, function () {
+  ipcRendererInternal.on(`GUEST_VIEW_INTERNAL_DESTROY_GUEST-${viewInstanceId}`, function () {
```
nornagon, line 70: can we use the enum here?

nornagon: This seems like a potentially useful typescript feature coming in 4.1: https://devblogs.microsoft.com/typescript/announcing-typescript-4-1-beta/#template-literal-types

nornagon (changes requested):
Nice stuff! Requesting changes for usage of the enum in a couple more places.

Can we change the type signature of `ipcRendererInternal.send` so it requires its first argument to be a member of the enum?

lib/common/remote/ipc-messages.ts
```diff
@@ -0,0 +1,19 @@
+export const enum IPC_MESSAGES {
```
nornagon, line 1: Please also apply this change to https://github.com/electron/remote.

lib/renderer/web-view/guest-view-internal.ts
miniak: forgot this one

lib/browser/guest-view-manager.js
miniak: done in TypeScript, I don't think we can use `const enum` in JavaScript

miniak: they cannot be used here in JavaScript code, needs to be converted to TypeScript

miniak: or I can change the `enum` to be non-`const`

lib/common/remote/ipc-messages.ts
miniak: https://github.com/electron/remote/pull/10

miniak:
> Can we change the type signature of ipcRendererInternal.send so it requires its first argument to be a member of the enum?
@nornagon that's more complicated and might require converting the last 2 files to TypeScript.

alexeykuzmin (approved):

zcbenz (approved):

lib/browser/remote/server.ts
```diff
@@ -30,7 +31,7 @@ const finalizationRegistry = new FinalizationRegistry((fi: FinalizerInfo) => {
   const ref = rendererFunctionCache.get(mapKey);
   if (ref !== undefined && ref.deref() === undefined) {
     rendererFunctionCache.delete(mapKey);
-    if (!fi.webContents.isDestroyed()) { fi.webContents.sendToFrame(fi.frameId, 'ELECTRON_RENDERER_RELEASE_CALLBACK', fi.id[0], fi.id[1]); }
+    if (!fi.webContents.isDestroyed()) { fi.webContents.sendToFrame(fi.frameId, 'RENDERER_RELEASE_CALLBACK', fi.id[0], fi.id[1]); }
```
nornagon, line 34:
```suggestion
    if (!fi.webContents.isDestroyed()) { fi.webContents.sendToFrame(fi.frameId, IPC_MESSAGES.RENDERER_RELEASE_CALLBACK, fi.id[0], fi.id[1]); }
```

nornagon (approved):

PR #25702, opened 2020-09-30
fix: crash when application launched from UNUserNotificationCenter notification

miniak:
#### Description of Change
We have a custom node module, which is using the new `UNUserNotificationCenter` APIs to show notifications. When the application is launched from such notification, the `NSApplicationLaunchUserNotificationKey` points to a `UNNotificationResponse` instance instead of `NSNotification`. This leads to an unrecognized selector exception. Check for `NSNotification` before accessing the `userInfo` property.

We can also use the `NSApplicationLaunchUserNotificationKey` directly, as it was introduced in macOS 10.8
Checking `userInfo` for `nil` is not necessary, as it's done by `NSDictionaryToDictionaryValue`.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Fixed crash when application launched from `UNUserNotificationCenter` notification (via a native node module).

### DIFF
--- a/shell/browser/mac/electron_application_delegate.mm
+++ b/shell/browser/mac/electron_application_delegate.mm
@@ -69,9 +69,9 @@ static base::mac::ScopedObjCClassSwizzler* g_swizzle_imk_input_session;
 
 - (void)applicationDidFinishLaunching:(NSNotification*)notify {
   NSUserNotification* user_notification =
-      [notify userInfo][(id) @"NSApplicationLaunchUserNotificationKey"];
+      [notify userInfo][NSApplicationLaunchUserNotificationKey];
 
-  if (user_notification.userInfo) {
+  if ([user_notification isKindOfClass:[NSUserNotification class]]) {
     electron::Browser::Get()->DidFinishLaunching(
         electron::NSDictionaryToDictionaryValue(user_notification.userInfo));
   } else {


codebytere (approved):

MarshallOfSound (approved):

shell/browser/mac/electron_application_delegate.mm
```diff
@@ -69,9 +69,9 @@ - (void)applicationWillFinishLaunching:(NSNotification*)notify {
 
 - (void)applicationDidFinishLaunching:(NSNotification*)notify {
   NSUserNotification* user_notification =
-      [notify userInfo][(id) @"NSApplicationLaunchUserNotificationKey"];
+      [notify userInfo][NSApplicationLaunchUserNotificationKey];
 
-  if (user_notification.userInfo) {
+  if ([user_notification isKindOfClass:[NSUserNotification class]]) {
```
nornagon, line 74: The old code checked for `userInfo != null`, should we still be doing that?

miniak: https://github.com/electron/electron/blob/822b04406873e60dcbccc29c5e8e980329d4b426/shell/browser/mac/dict_util.mm#L74

miniak: Checking `userInfo` for `nil` is not necessary, as it's done by `NSDictionaryToDictionaryValue`.

nornagon (approved):

alexeykuzmin (approved):

PR #25757, opened 2020-10-02
fix: nativeImage.crop().toBitmap() returning garbage

### LINKED ISSUE #9351: Add image type to nativeImage.createFromBuffer
* Electron version: 1.4.16 / 1.6.6 
* Operating system: macOS

### Expected behavior

`toBitmap` method returns proper buffer for `nativeImage` created by cropping an existing `nativeImage`

### Actual behavior
The returned buffer's length is not correct, and `createBuffer` from it results in wrong image

### How to reproduce

The repo to reproduce is https://github.com/KagamiChan/electron-quick-start/tree/nativeimage-crop (branch `nativeimage-crop`). The renderer script reads the lenna picture, crops it and reconstructs the cropped part using `toBitmap` and `createFromBuffer`:

```js
const { nativeImage } = require('electron')
const { resolve } = require('path')
const { writeFile } = require('fs')

const image = nativeImage.createFromPath(resolve(__dirname, './lena_std.png'))
const { width, height } = image.getSize()
const bufferLength = image.toBitmap().length
console.log('original image', width, height, bufferLength, width * height * 4)

const crop = image.crop({x: 20, y: 20, width: 220, height: 330})
const { width:cropWidth, height:cropHeight } = crop.getSize()
const cropLength = crop.toBitmap().length
const _cropLength = crop.getBitmap().length
console.log('cropped image', cropWidth, cropHeight, cropLength, _cropLength, cropWidth * cropHeight * 4)

writeFile(resolve(__dirname, './crop.png'), crop.toPNG(), e => {if (e) console.error(e)})

const buffer = crop.toBitmap()
const crop2 = nativeImage.createFromBuffer(buffer, {
  width: cropWidth,
  height: cropHeight,
})
const cropLength2 = crop2.toBitmap().length
const _cropLength2 = crop2.getBitmap().length
console.log('cropped and restored image', cropWidth, cropHeight, cropLength2, _cropLength2, cropWidth * cropHeight * 4)

const pngBuffer = crop2.toPNG()
writeFile(resolve(__dirname, './crop2.png'), pngBuffer, e => {if (e) console.error(e)})
```

the console log:
```
original image 512 512 1048576 1048576
cropped image 220 330 674672 674672 290400
cropped and restored image 220 330 290400 290400 290400
```

Observations:
- the nativeImage created from lenna picture gives a buffer length that equals to `width * height * 4`
- the cropped nativeImage object returns a buffer with much larger size, since a node buffer's length is the memory allocated, this could happen
- however, the reconstructed nativeImage object's content is not what we expected

crop.png (what we expect)
crop

crop2.png (the reconstructed)
crop2

In my opinion `toBitmap` method is most suspicious but `crop` is another possibility because there's another issue (#8017) related to it

nornagon:
#### Description of Change
Fixes #9351. That PR hypothesized that createFromBuffer was parsing the bitmap as a PNG, which is not the case. What was actually happening was that `.crop()` was returning an image which shared the underlying data, but updated the bounds. This meant that when `.toBitmap()` was called, it was retrieving a reference to the pre-cropped pixel data, which included pixels off to the right of the image (because that is how the pixels are stored in memory).

```
memory:
0x0000: rgbargba rgbargba rgbargba rgbargba
0x0020: rgbargba rgbargba rgbargba rgbargba
0x0040: rgbargba rgbargba rgbargba rgbargba
0x0060: rgbargba rgbargba rgbargba rgbargba
[...]
```

The cropped image's pixel data, as it is shared, points to the same start position (0x0000 in the above example), but stores the information that only the first N pixels of each row should be read. For example, a crop starting at 0,0 of size 4x3 of the above image would look like:

```
0x0000:[rgbargba rgbargba]rgbargba rgbargba
0x0020:[rgbargba rgbargba]rgbargba rgbargba
0x0040:[rgbargba rgbargba]rgbargba rgbargba
0x0060: rgbargba rgbargba rgbargba rgbargba
[...]
```

`bitmap.pixelRef()` returns the pointer to the start, and `computeByteSize()` returns the total amount of bytes needed to encompass _all_ the data (including the bits that were cropped out to the left and right of the cropped image), but it does not specify which subset of each row to copy, which is why the returned buffer length was longer than expected.

The new code uses Skia's own functions for copying pixel data around instead of operating on raw memory, resolving the issue.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed NativeImage.crop().toBitmap() returning incorrect data.

### DIFF
--- a/shell/common/api/electron_api_native_image.cc
+++ b/shell/common/api/electron_api_native_image.cc
@@ -183,13 +183,20 @@ v8::Local<v8::Value> NativeImage::ToBitmap(gin::Arguments* args) {
 
   const SkBitmap bitmap =
       image_.AsImageSkia().GetRepresentation(scale_factor).GetBitmap();
-  SkPixelRef* ref = bitmap.pixelRef();
-  if (!ref)
-    return node::Buffer::New(args->isolate(), 0).ToLocalChecked();
-  return node::Buffer::Copy(args->isolate(),
-                            reinterpret_cast<const char*>(ref->pixels()),
-                            bitmap.computeByteSize())
-      .ToLocalChecked();
+
+  SkImageInfo info =
+      SkImageInfo::MakeN32Premul(bitmap.width(), bitmap.height());
+
+  auto array_buffer =
+      v8::ArrayBuffer::New(args->isolate(), info.computeMinByteSize());
+  auto backing_store = array_buffer->GetBackingStore();
+  if (bitmap.readPixels(info, backing_store->Data(), info.minRowBytes(), 0,
+                        0)) {
+    return node::Buffer::New(args->isolate(), array_buffer, 0,
+                             info.computeMinByteSize())
+        .ToLocalChecked();
+  }
+  return node::Buffer::New(args->isolate(), 0).ToLocalChecked();
 }
 
 v8::Local<v8::Value> NativeImage::ToJPEG(v8::Isolate* isolate, int quality) {
--- a/spec/api-native-image-spec.js
+++ b/spec/api-native-image-spec.js
@@ -468,6 +468,12 @@ describe('nativeImage module', () => {
       expect(cropB.getSize()).to.deep.equal({ width: 25, height: 64 });
       expect(cropA.toPNG().equals(cropB.toPNG())).to.be.false();
     });
+
+    it('toBitmap() returns a buffer of the right size', () => {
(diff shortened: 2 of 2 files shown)


codebytere (approved):

zcbenz (approved):

nornagon:
hm, actually 8.x should be security fixes only at this point i think, which this isn't, so i'll skip 8.

PR #25763, opened 2020-10-04
fix: output Unix-style paths for git diffs on Win

dsanders11:
#### Description of Change
The output of the linting script recommends using `git apply` to apply changes, but on Windows this fails because the paths in the patch aren't compatible with `git apply`. This PR ensures the paths look as expected when running on Windows.

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/script/run-clang-format.py
+++ b/script/run-clang-format.py
@@ -15,6 +15,7 @@ import fnmatch
 import io
 import multiprocessing
 import os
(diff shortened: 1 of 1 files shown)


zcbenz (approved):

jkleinsc (approved):
Thanks for the PR @dsanders11!

PR #25766, opened 2020-10-05
build: fix typos in script messages

dsanders11:
#### Description of Change
Just fixing a recurring typo I noticed in some scripts.

cc @zcbenz 

#### Checklist

- [x] PR description included and stakeholders cc'd

#### Release Notes

Notes: none

PR #25767, opened 2020-10-05
build: python3 compat and fix line endings on Win

dsanders11:
#### Description of Change
The `check-trailing-whitespace.py` script wasn't Python3 compatible and was also swapping the line endings for all files when run on Windows.

This includes a typo fix from #25766, so best to merge that PR first for no conflict here.

cc @zcbenz 

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/script/check-trailing-whitespace.py
+++ b/script/check-trailing-whitespace.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
 
-from __future__ import print_function
+from __future__ import print_function, unicode_literals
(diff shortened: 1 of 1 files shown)


dsanders11:
The lint check on this PR failed due to the PR changes actually detecting some trailing whitespace which wasn't caught before. That can be fixed by merging #25768 which should make the lint check in this PR pass.

zcbenz (approved):

PR #25772, opened 2020-10-05
chore: use no-op ShouldAbortOnUncaughtException cb

codebytere:
#### Description of Change

Previously we were letting Node.js use its default callback for deciding whether uncaught exceptions should abort the process. However, in the renderer process we definitely do not want to crash the process. We in fact already intercept uncaught exceptions in both the [renderer](https://github.com/electron/electron/blob/8baa9deccdbd2473c99811eaf2291c3d00fce0ff/lib/renderer/init.ts#L153) and [browser](https://github.com/electron/electron/blob/77038b7bda7c6633523541fc452bba1bc487e277/lib/browser/init.ts#L45) process to prevent quitting, and so should always return false in this isolate callback.

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/common/node_bindings.cc
+++ b/shell/common/node_bindings.cc
@@ -429,6 +429,12 @@ node::Environment* NodeBindings::CreateEnvironment(
   // crash message and location to CrashReports.
   is.fatal_error_callback = V8FatalErrorCallback;
 
+  // We don't want to abort either in the renderer or browser processes.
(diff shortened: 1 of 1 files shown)


MarshallOfSound (approved):

codebytere:
Arm failures are both unrelated known flakes.

Issue #25127, opened 2020-08-25
v10.0.0: cannot compile wasm when contextIsolation is enabled


ericmandel:
### Preflight Checklist

* [x ] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [ x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [ x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  *  10.0.0
* **Operating System:**
  *  macOS 10.14.6
* **Last Known Working Electron version:**
  *  9.2.1

### Expected Behavior

Wasm should be allowed to compile when contextIsolation is enabled, as it did in v9.2.1.

### Actual Behavior

Wasm compilation fails:

<img width="1125" alt="Screen Shot 2020-08-25 at 4 24 19 PM" src="https://user-images.githubusercontent.com/3028875/91225978-63db1e00-e6f2-11ea-9a70-7595d51969f9.png">

Turning off contextIsolation fixes the problem.

### Additional Information

Note that this error is different from #25066, in which it was clarified that Node is not available to the web page and emscripten when contextIsolation is turned on. In this case, it looks like the browser is not permitting wasm compilation at all.

quanglam2807:
I have the same problem when loading Figma with Electron 10.x

<img width="912" alt="Screen Shot 2020-08-27 at 9 15 43 PM" src="https://user-images.githubusercontent.com/1548835/91453900-81eb7e80-e8aa-11ea-896b-7d8674c636e0.png">

codebytere:
Thank you for taking the time to report this issue and helping to make Electron better.

Please provide a standalone test that a tester could run to reproduce the issue. This makes fixing issues go more smoothly: not only does it ensure the right thing is tested, but it also prevents triage from bottlenecking on making testers write new tests for each new issue.

I'm setting the more-information-needed label for the above reasons. After you've responded, please @ me in a followup comment.

Thanks in advance! Your help is appreciated.

ericmandel:
@codebytere A stand-alone test is attached in hello.zip that instantiates wasm successfully using v9.2.1 but not v10. To run, unzip and:
````
/Applications/Electron.app/Contents/MacOS/Electron main.js false    # turn off contextIsolation => success in 9.2.1, 10
[...]

PR #25777, opened 2020-10-05
fix: wasm code generation in the renderer

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/25127.

This issue arose from the fact that when `contextIsolation` is enabled, Node.js does not have control over aspects of the context being used in the renderer process. It therefore goes to use its own callback handlers by default, but has not set embedder data for the context in question, and so when it pulls the data piece representing whether Wasm codegen should be enabled from the index it _thinks_ such data will be at, this data piece is actually (don't ask) [`String.prototype.link`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/link) represented as a `v8::Function`. We handle this by asking for the current Environment, and if none exists, we know that we're in the renderer with `contextIsolation` enabled and should fall back to Blink's handler.

cc @MarshallOfSound @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where Wasm code generation erroneously showed as being disallowed by embedder when `contextIsolation` was enabled.


shell/common/node_bindings.cc
```diff
@@ -151,6 +152,25 @@ void V8FatalErrorCallback(const char* location, const char* message) {
...
+bool AllowWasmCodeGenerationCallback(v8::Local<v8::Context> context,
+                                     v8::Local<v8::String>) {
+  // If we're running with contextIsolation enabled, fall back to Blink's logic.
+  // We switch on the embedder data field count here because if we're in a ctx
+  // that Node.js doesn't control, the number of embedder data fields will
+  // always be less than the min index Node.js uses.
+  if (context->GetNumberOfEmbedderDataFields() <
+      node::ContextEmbedderIndex::kEnvironment) {
+    v8::Isolate* isolate = context->GetIsolate();
+    LOG(INFO) << __LINE__;
```
MarshallOfSound, line 164: 😄

shell/common/node_bindings.cc
```diff
@@ -151,6 +152,25 @@ void V8FatalErrorCallback(const char* location, const char* message) {
...
+    return blink::V8Initializer::WasmCodeGenerationCheckCallbackInMainThread(
+        context, v8::String::Empty(isolate));
+  }
+
+  v8::Local<v8::Value> wasm_code_gen = context->GetEmbedderData(
```
MarshallOfSound, line 169: Can we phone into nodes logic here instead of copying it?  This method would ideally be:

```js
if (isNodeContext(context))
  return nodeThing();

return blinkThing();
```

Might have to export nodes thing too

shell/common/node_bindings.cc
```diff
@@ -151,6 +152,25 @@ void V8FatalErrorCallback(const char* location, const char* message) {
   *zero = 0;
 }
 
+bool AllowWasmCodeGenerationCallback(v8::Local<v8::Context> context,
+                                     v8::Local<v8::String>) {
+  // If we're running with contextIsolation enabled, fall back to Blink's logic.
+  // We switch on the embedder data field count here because if we're in a ctx
+  // that Node.js doesn't control, the number of embedder data fields will
+  // always be less than the min index Node.js uses.
+  if (context->GetNumberOfEmbedderDataFields() <
```
codebytere, line 161: We may want to instead switch this on `node::Environment::GetCurrent` being `nullptr` 🤔

codebytere: (gonna do that)

MarshallOfSound: Yeah I think that makes sense

shell/common/node_bindings.cc
```diff
@@ -151,6 +152,23 @@ void V8FatalErrorCallback(const char* location, const char* message) {
...
 
+bool AllowWasmCodeGenerationCallback(v8::Local<v8::Context> context,
+                                     v8::Local<v8::String>) {
+  // If we're running with contextIsolation enabled, fall back to Blink's logic.
+  // We switch on the embedder data field count here because if we're in a ctx
+  // that Node.js doesn't control, the number of embedder data fields will
+  // always be less than the min index Node.js uses.
+  v8::Isolate* isolate = context->GetIsolate();
+  if (node::Environment::GetCurrent(isolate) == nullptr) {
+    return blink::V8Initializer::WasmCodeGenerationCheckCallbackInMainThread(
```
MarshallOfSound, line 163: Should we also _only_ do this in the renderer? I don't see a valid case for this to be called from the browser.  I think if we end up in this block in the browser process we should return false

MarshallOfSound: Should do a runtime process check

shell/common/node_bindings.cc
codebytere: node might be more reticent to do that - can we add it as a todo item and refactor this if they're amenable so we don't pre-emptively add a patch we can't get rid of?

MarshallOfSound: Without referencing it exactly I'm not sure how we'd keep it in sync 🤔  I think the one-line patch we might have to keep around forever is OK (though I'd still like it upstreamed).  It's just a `NODE_EXPORT` right?

MarshallOfSound (approved):

PR #25782, opened 2020-10-05
fix: prevent UAF in NativeImage.getBitmap

### LINKED ISSUE #25753: nativeImage.getBitmap() allows use-after-free
`nativeImage.getBitmap()` allows accessing raw memory after it is freed.

```js
const { nativeImage } = require('electron')

const img = nativeImage.createFromDataURL('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAAC0lEQVQYlWNgAAIAAAUAAdafFs0AAAAASUVORK5CYII=')
const bitmap = img.getBitmap()
console.log(bitmap)
setTimeout(() => {
  gc()
  console.log(bitmap)
}, 1000)
```

The above example produces the following output on my machine:

```
<Buffer 00 00 00 00>
<Buffer 50 b8 41 05>
```

The 2nd log is just an arbitrary chunk of memory. It's random each time I run the program.

nornagon:
#### Description of Change
Fixes #25753.

This holds a reference to the underlying pixel data for as long as the buffer
returned by getBitmap() is referenced.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where the buffer returned by NativeImage.getBitmap() could point to freed memory.

### DIFF
--- a/shell/common/api/electron_api_native_image.cc
+++ b/shell/common/api/electron_api_native_image.cc
@@ -103,8 +103,6 @@ base::win::ScopedHICON ReadICOFromPath(int size, const base::FilePath& path) {
 }
 #endif
 
-void Noop(char*, void*) {}
-
 }  // namespace
 
 NativeImage::NativeImage(v8::Isolate* isolate, const gfx::Image& image)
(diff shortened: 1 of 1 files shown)


zcbenz (approved):

zcbenz:
I wonder if the build change in https://github.com/electron/electron/pull/25396 can be reverted since this seems to be the direct cause of the nativeImage crash. /cc @jkleinsc

codebytere (approved):

jkleinsc:
@zcbenz I guess we could test if this fixes it

nornagon:
Hm, I don't recall all the details of the ARM failure, but just to note that this change is only relevant to `NativeImage.getBitmap()`; if there were failures that did not use that function then this change will have no effect on them.

PR #25795, opened 2020-10-06
feat: add support for preventing the system context menu

MarshallOfSound:
Docs explain this well enough hopefully, basically there is a menu that appears when you right click on the non-client area of your app.  This event lets you prevent that menu appearing and passes the coordinates so you can in theory replace it with your own menu if you wanted.

Pretty standard event wiring, nothing new here.

Notes: Added new `system-context-menu` event to allow preventing and overriding the system context menu

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -667,6 +667,20 @@ Emitted when the window has closed a sheet.
 
 Emitted when the native new tab button is clicked.
 
+#### Event: 'system-context-menu' _Windows_
+
+Returns:
+
+* `event` Event
+* `point` [Point](structures/point.md) - The screen coordinates the context menu was triggered at
+
+Emitted when the system context menu is triggered on the window, this is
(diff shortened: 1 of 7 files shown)


samuelmaddock:
Does this also prevent the menu from appearing when pressing the `Alt + Space` hotkey?

MarshallOfSound:
> Does this also prevent the menu from appearing when pressing the Alt + Space hotkey?

Unfortunately not, I can't seem to use wndproc to prevent that

samuelmaddock:
> 
> 
> > Does this also prevent the menu from appearing when pressing the Alt + Space hotkey?
> 
> Unfortunately not, I can't seem to use wndproc to prevent that

I've never seen a windowed app override that behavior, likely for accessibility reasons, so that's fine. I was mostly just curious whether that would be affected.

samuelmaddock (approved):

codebytere (approved):

Issue #22192, opened 2020-02-14
Notification timeoutType='never' doesn't work


p-fusi:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 8.0.0
* **Operating System:**
  * Windows 10 Pro (1709)

### Expected Behavior
Using Notification API in main process with `timeoutType: 'never` I expected that the notification disappears only when the user clicks on it.  

### Actual Behavior
After few seconds the notification disappers.

### To Reproduce

```
  const newNotification = new Notification({title: 'I am a main process Notification!', timeoutType: 'never'})
  newNotification.show()
```

```sh
$ git clone https://github.com/p-fusi/electron-quick-start -b master
$ npm install
$ npm run start
```

p-fusi:
Sorry, any update for this problem?

jp185318:
any one got the solution for above listed problem

mbuchner:
notifications are a pretty important thing when deciding for electron -> nobody could find a fix?

LaserYGD:
This issue is still ongoing, just tested today on the latest stable release. like mbu mentioned it's really important, sometimes you want to make sure the user sees the notification -- if not set to never, he could be AFK and by the time he's back it's gone

codebytere:
This is being worked on.

JohansenJunias20:
i fixed my problem. I putted title inside the object option instead parameter.
**Before :** 
`new Notification("hi there!",{ body: "helloworld", timeoutType: "never" })`
**After :**
`new Notification({ title:"hi there!",body: "helloworld", timeoutType: "never" })`

PR #25820, opened 2020-10-08
fix: notifications successfully never timeout with included flag

mlaurencin:
#### Description of Change

Closes https://github.com/electron/electron/issues/22192

This PR is fixing the bug that was appearing in using the timeoutType='never' option when creating a Windows toast notification. Before the notification was still disappearing after about 7 seconds of being on the screen. This fix adds a "Close" button to the notification, which causes the scenario='reminder' attribute of the notification XML to be respected and the notification to never timeout. This attribute is only respected when a button is present on the notification XML.

Before:

image (1)

After:

image

cc @codebytere 

#### Checklist
- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where Windows notifications with `timeoutType` of 'never' did not work properly.


shell/browser/notifications/win/windows_toast_notification.cc
```diff
@@ -265,13 +265,136 @@ HRESULT WindowsToastNotification::SetXmlScenarioReminder(IXmlDocument* doc) {
...
+  // Create content attribute
+  ComPtr<IXmlAttribute> content_attribute;
+  ScopedHString content_str(L"content");
+  RETURN_IF_FAILED(doc->CreateAttribute(content_str, &content_attribute));
+
+  ComPtr<IXmlNode> content_attribute_node;
+  RETURN_IF_FAILED(content_attribute.As(&content_attribute_node));
+
+  // Set content attribute to Dismiss
+  ScopedHString content_value(L"Dismiss");
```
codebytere, line 380: At the moment this is hardcoded to `Dismiss` - ideally we'll want to ensure that this gets localized to the user's current location. We can do this with the localization utility from Chromium:

```
#include "ui/base/l10n/l10n_util.h"
...
stuff
...
// Set content attribute to Dismiss
ScopedHString content_value(l10n_util::GetStringUTF16(IDS_TOAST_NOTIFICATION_DISMISS_TEXT));
```

We'll also want to add this string to our list of generated strings in `electron_strings.grdp` i would say, something like:

```
<message name="IDS_TOAST_NOTIFICATION_DISMISS_TEXT" desc="Text label for the Dismiss button in Windows Toast Notifications. This is used when the toast has a timeout type of never.">
      Dismiss
</message>
```

in the [windows-only block](https://github.com/electron/electron/blob/8bbdc224ac3d7756af930ce946a59ce5ff2533b5/electron_strings.grdp#L78).

codebytere (commented):
This looks great! Just one comment about localization.

shell/browser/notifications/win/windows_toast_notification.cc
```diff
@@ -265,13 +265,136 @@ HRESULT WindowsToastNotification::SetXmlScenarioReminder(IXmlDocument* doc) {
...
+  if (!content_value.success())
+    return E_FAIL;
```
ckerr, line 385: _(nit)_ Not a big deal, but for consistency with the rest of the code:

```suggestion
  RETURN_IF_FAILED(content_value.success());
```

shell/browser/notifications/win/windows_toast_notification.cc
mlaurencin: Done!

ckerr: Chromium has the same issue and has similar code to inject a button to prevent the notification from timing out -- can we piggyback on their strings, e.g. `IDS_APP_CLOSE` instead of creating another one?

codebytere (approved):
🥳

electron_strings.grdp
```diff
@@ -79,5 +79,9 @@
     <message name="IDS_UTILITY_PROCESS_UTILITY_WIN_NAME" desc="The name of the utility process used to handle Windows utility operations.">
       Windows Utilities
     </message>
+    <message name="IDS_TOAST_NOTIFICATION_DISMISS_TEXT" desc="Text label for the Dismiss button in Windows Toast Notifications. This is used when the toast has a timeout type of never.">
```
ckerr, line 82: I don't know if we need to create a new string? Upstream already has 

```
      <message name="IDS_APP_CLOSE" desc="A generic term for Close on buttons and menus.">
        Close
      </message>
```

in `ui/strings/`

shell/browser/notifications/win/windows_toast_notification.cc
codebytere: @ckerr since this is present in a bunch of other toast code as well - why don't we do it as a follow-up and change all of them at once?

electron_strings.grdp
codebytere: We also have https://github.com/electron/electron/blob/8bbdc224ac3d7756af930ce946a59ce5ff2533b5/electron_strings.grdp#L54-L56 - I think Dismiss is slightly more in line with what [Toast Content docs](https://docs.microsoft.com/en-us/windows/uwp/design/shell/tiles-and-notifications/adaptive-interactive-toasts#snoozedismiss) have but I don't feel strongly

shell/browser/notifications/win/windows_toast_notification.cc
ckerr: @codebytere oh, you're right, it is -- there are several places that use this exact phrasing in extant code. :+1:

electron_strings.grdp
ckerr: It just seems like pointless translation work & lack-of-translations when Chromium already has a fully-translated string sitting on the shelf :smiley: 

OTOH if we want to use "Dismiss" rather than "Close" for desktop consistency, we still may be able to use existing strings. That link says:

> Optionally specify a custom content string:
>
> If you don't provide a string, we'll automatically use localized strings for "Snooze" and "Dismiss".

It sounds like they're saying we can use `<action activationType="system" arguments="dismiss" content=""/>` and Windows will fill in the default string?

codebytere: Heh unfortunately that doc is a bit misleading - @mlaurencin figured out yesterday that that only applies to the _Windows_ API and not the generated XML - if `content` is left blank in our code (e.g. we don't supply a string) then the button simply does not show up 🙃

ckerr: OK. In that case I'd vote for reusing the Chromium string instead of reusing the nonexistent Windows string :smile:

codebytere: sound good to me! @mlaurencin would you mind updating that?

mlaurencin: Got it!

electron_strings.grdp
```diff
@@ -80,4 +80,5 @@
       Windows Utilities
     </message>
   </if>
+
```
codebytere, line 83: let's get rid of this extra newline :)

ckerr (approved):
LGTM. Thanks!

shell/browser/notifications/win/windows_toast_notification.cc
mlaurencin: @ckerr @codebytere I think it may be best to leave this as it is actually. `RETURN_IF_FAIL` will only return if the value is strictly less than 0, but with `success()` returning a boolean, it will never register as failing.

PR #25843, opened 2020-10-09
docs: gclient sync --ignore_locks was removed

dsanders11:
#### Description of Change

It seems that `gclient sync` has removed `--ignore_locks`, when you use it you get the following output (but execution continues): `Warning: ignore_locks is no longer used. Please remove its usage.`

It looks like this change [landed in May 2020](https://chromium.googlesource.com/chromium/tools/depot_tools.git/+/14a83aec5603e47208a4a41c8c085a1991d33d47). There does not appear to be an equivalent flag, I believe that change was meant to alleviate the issue and remove the need for the flag. Since `gclient` auto-updates, I assume pretty much anyone using the docs will be using a version with that change, so seems reasonable to drop this section from the docs.

cc @codebytere 

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/docs/development/build-instructions-gn.md
+++ b/docs/development/build-instructions-gn.md
@@ -254,10 +254,6 @@ New-ItemProperty -Path "HKLM:\System\CurrentControlSet\Services\Lanmanworkstatio
 
 ## Troubleshooting
 
-### Stale locks in the git cache
-If `gclient sync` is interrupted while using the git cache, it will leave
-the cache locked. To remove the lock, pass the `--ignore_locks` argument to `gclient sync`.
-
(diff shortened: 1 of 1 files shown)


codebytere (approved):

codebytere:
@dsanders11 thanks for these fixes! also, out of curiosity, are you aware of https://github.com/electron/build-tools? It generally abstracts most of these details away from you :)

dsanders11:
> are you aware of https://github.com/electron/build-tools? It generally abstracts most of these details away from you :)

@codebytere I wasn't aware of it, I'll take a look, thanks!

miniak (approved):

zcbenz (approved):

Issue #1083, opened 2015-02-02
Mailto: openExternal does not work on Windows.


marksalvin:
I've been trying to use openExternal to open the mail application from within my atom-shell app (with an email passed in), however it's not working on Windows (it's working fine on Mac). Using latest release (v0.21.0, non-symbols version).

Using:

`require("shell").openExternal("http://www.google.com")` - works.
`require("shell").openExternal("mailto:test@test.com")` - DOESN'T work. Nothing happens. No error is thrown either in WebInspector console.

lincleejun:
If  you doesn't install any mail client on your system, this calling will return nothing, but  if you install one it works correctly.

in OpenExternal implementation.
when system doesn't have any Mail client,  calling  ShellExecuteA will got SE_ERR_NOASSOC error.
but if I uninstall got SE_ERR_ACCESSDENIED error code(not restart computer).

maybe should change the API signature, give some response to caller.

bwin:
With 0.21.2 on Win7 it opens Thundebird without problems.
I'm not going to uninstall my mail-client, but I think it's as lincleejun says: There is no return value, so you don't know if it worked or not.
The title should be changed to "openExternal doesn't tell if there's no mail client" or something...

marksalvin:
I'll try again with the latest build, however I have outlook installed on my system, and mailto: links within the web browser open outlook just fine, it's only from within atom that nothing happens.

marksalvin:
Did not mean to close this issue. So I can confirm that with the latest version I'm still having issues. I have outlook installed on my system, and mailto: links within the web browser open outlook just fine, it's only from within atom that nothing happens. I installed thunderbird, and the openExternal worked just once but has not worked again since, bizarrely. 

So I wouldn't say that it's just a return value issue.

zcbenz:
This only seems to affect one machine and I didn't see any other report, no one else is able to reproduce this, so I'm blaming it to Windows.

bradstewart:
I'm having the problem exactly as described by @marksalvin on 0.36.2 on all Windows 7 machines I've tested. The machines did have a mail client (tried both Outlook and Thunderbird) installed and registered as the default for the Url:MAILTO scheme.

Ivshti:
Windows is to blame for this, definitely, but has anyone found a workaround? @bradstewart @markisadesignerd

marksalvin:
Hi guys.

So originally, my workaround was to modify the openExternal function on Windows:

[...]

PR #25858, opened 2020-10-09
chore: cherry-pick 8629cd7f8af3 from chromium

ppontes:
Add mutex for allowing the graph to be pulled

Basically add a mutex to protect access to
allow_pulling_audio_graph_.  The main thread waits until it has the
lock before setting this.  This prevents the main thread from deleting
things while the audio thread is pulling on the graph.

A try lock is used so as not to block the audio thread if it can't get
lock.

This is applied to both real time and offline contexts which required
moving the original real-time-only implementation to
audio_destination_node so we can use the same methods for the offline
context.

Tested the repro case from 1125635, and the issue does not reproduce.
We're assuming this will fix 1115901, but I've not been able to
reproduce that locally.

(cherry picked from commit 6785a406fd652f7a2f40933620ef4c979643c56b)

Bug: 1125635, 1115901
Change-Id: I1037d8c44225c6dcc8fe906c29a5a86740a15e1d
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2410743
Cr-Original-Commit-Position: refs/heads/master@{#809393}
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2436324
Cr-Commit-Position: refs/branch-heads/4240@{#1083}
Cr-Branched-From: f297677702651916bbf65e59c0d4bbd4ce57d1ee-refs/heads/master@{#800218}

Notes: Security: backported fix for 1125635, 1115901.

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -133,3 +133,4 @@ cherry-pick-0e61c69ebd47.patch
 cherry-pick-adc731d678c4.patch
 cherry-pick-52dceba66599.patch
 cherry-pick-abc6ab85e704.patch
+cherry-pick-8629cd7f8af3.patch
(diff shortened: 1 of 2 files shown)


nornagon (approved):

ppontes:
This fix isn't straightforward or safe to backport and it actually seems that M80 doesn't suffer from this vulnerability, as it syncs using a different mutex. Closing.

PR #25864, opened 2020-10-09
fix: update squirrel.mac to handle bug in Big Sur

### LINKED ISSUE #25626: autoUpdater.quitAndInstall fails to relaunch on Mac OS 11 Big Sur
### Preflight Checklist

### Issue Details

I wanted to post this in https://github.com/electron/electron/issues/24319 but the issue is locked.

[Related comment with more details](https://github.com/Squirrel/Squirrel.Mac/issues/251#issuecomment-698010499).

* **Electron Version:**
  * v8.5.1
* **Operating System:**
  * Mac OS 11 Big Sur Beta
  * n/a

### Actual Behavior

When restarting an upgraded electron application with [autoUpdater.quitAndInstall](https://www.electronjs.org/docs/api/auto-updater#autoupdaterquitandinstall), the application quits and upgrades successfully, but is NOT relaunched thereafter.

### Expected Behavior

The application should be launched after upgrading.

### To Reproduce

It's very hard for me to reproduce this issue because Squirrel.Mac requires that any MacOS app using auto-update is code-signed, and I can't figure out how to build Squirrel from source successfully so as to artificially remove that restriction. 

If there are any guides available for building Squirrel.Mac + Electron together, i would be very thankful to hear about them.

### Additional Information

As mentioned [here](https://github.com/Squirrel/Squirrel.Mac/issues/251#issuecomment-698010499), there are two lines in the Squirrel install log normally present with earlier versions of MacOS which are missing on the Big Sur log.

```
2020-09-17 17:55:15.344 ShipIt[4099:100431] Application launched at file:///Applications/Whatever.app/
2020-09-17 17:55:15.344 ShipIt[4099:100431] ShipIt quitting
```

Those lines are logged from these locations:

https://github.com/Squirrel/Squirrel.Mac/blob/44468f858ce0d25c27bd5e674abfa104e0119738/Squirrel/ShipIt-main.m#L128
https://github.com/Squirrel/Squirrel.Mac/blob/44468f858ce0d25c27bd5e674abfa104e0119738/Squirrel/ShipIt-main.m#L145

My original plan was to add some extra logging statements in Squirrel.Mac's source code and use that to determine why those lines are missing, but i hit a brick wall trying to build Squirrel and Electron together from lack of documentation.

MarshallOfSound:
Fixes #25626

This is not the greatest workaround but at least it works.

Notes: Fixed issue where `quitAndInstall()` would not restart the app upon update completion on macOS Big Sur

### DIFF
--- a/spec-main/api-autoupdater-darwin-spec.ts
+++ b/spec-main/api-autoupdater-darwin-spec.ts
@@ -307,7 +307,7 @@ describeFn('autoUpdater behavior', function () {
 
         await relaunchPromise;
         expect(requests).to.have.lengthOf(3);
-        expect(requests[2]).to.have.property('url', '/update-check/updated/2.0.0');
+        expect(requests[2].url).to.equal('/update-check/updated/2.0.0');
         expect(requests[2].header('user-agent')).to.include('Electron/');
       });
     });
--- a/DEPS
+++ b/DEPS
@@ -20,7 +20,7 @@ vars = {
   'nan_version':
     '2c4ee8a32a299eada3cd6e468bbd0a473bfea96d',
   'squirrel.mac_version':
-    '44468f858ce0d25c27bd5e674abfa104e0119738',
+    'a3a5b3f03b824441c014893b18f99a103b2603e9',
 
   'boto_version': 'f7574aa6cc2c819430c1f05e9a1a1a666ef8169b',
   'pyyaml_version': '3.12',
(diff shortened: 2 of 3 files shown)


spec-main/api-autoupdater-darwin-spec.ts
```diff
@@ -307,7 +307,7 @@ describeFn('autoUpdater behavior', function () {
 
         await relaunchPromise;
         expect(requests).to.have.lengthOf(3);
-        expect(requests[2]).to.have.property('url', '/update-check/updated/2.0.0');
+        expect(requests[2].url).to.equal('/update-check/updated/2.0.0');
```
nornagon, line 310: left side seems better..?

MarshallOfSound: It results in a very obscure error being thrown when that assertion fails (completely unrelated to the assertion).  Calling `inspect()` on a http request doesn't work and that's what the assertion library tries to log 😢

nornagon (approved):

codebytere (approved):

zcbenz (approved):

Issue #25496, opened 2020-09-16
Native multi-threaded N-API module callback not working in ES 9.x without allowRendererProcessReuse = false on Windows


f0zi:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * Any 9.x or higher with `app.allowRendererProcessReuse = true` on Windows
* **Operating System:**
  * Tested on Windows 10 and Mac OS Catalina
* **Last Known Working Electron version:**
  * Any 8.x or any 9.x with `allowRendererProcessReuse = false` or any on the Mac.

### Expected Behavior
We have a native module that we converted to N-API in the last release which shipped with ES 8.x. This native module does device access and exposes a `EventEmiter` interface to JS running in the renderer. Native code is calling `emit` on the main thread by using a [UV Async handle](http://docs.libuv.org/en/v1.x/async.html). Yes, we called [`napi_get_uv_event_loop`](https://nodejs.org/api/n-api.html#n_api_napi_get_uv_event_loop). For the sake of troubleshooting this issue I have also completely replaced the UV code with [thread safe functions](https://nodejs.org/api/n-api.html#n_api_asynchronous_thread_safe_function_calls) (with a `call_js_cb` and empty `func`) with the same results. Also for troubleshooting I have removed all code from the async worker except [debug outputs](https://docs.microsoft.com/en-us/windows/win32/api/debugapi/nf-debugapi-outputdebugstringa).

On ES 8.x or with `allowRendererProcessReuse = false` notification of the JS works fine and I can see from the debug outputs that the device is detected and the notification worker is called immediately.

### Actual Behavior
On ES 9.x with `allowRendererProcessReuse = true` the async worker seems to be called very inconsistently. It is usually called at least once and then stops being called entirely. I can still see from the debug output that our native module is detecting the device events and the call to `uv_async_send` returns 0 (success) however the async worker is not run any more. Sometimes it is called once more when the app exits, and on one of my colleagues' machine it got called after minutes of delay, but then stopped working again.

During all of this the UI is responsive, JS is running, I can start the JS debugger and do stuff.

### To Reproduce
[...]

Issue #25405, opened 2020-09-10
child_process exec callback not called (sometimes)


torakiki:
### Issue Details

* **Electron Version:**
  * 9.2.1
  * 10.1.1
* **Operating System:**
  * Windows 64 and 32
* **Last Known Working Electron version:**
  * 8.5.1

With version 9.2.1 and 10.1.1 it happens that _sometimes_ the child_process.exec callback is never called. It seems fine with version 8.5.1 (still testing) 

We have this code that sometimes doesn't call the callback, meaning it never print 'finished' on Windows 32 and 64bits. 
```
const exec = require('child_process').exec;

const execPromise = (cmd, execOpts) => {
    return new Promise(function(resolve, reject) {
	    console.log('start')
        exec(cmd, execOpts, (error, stdout, stderr) => {
		console.log('finished')
            if (error) {
                error.stdout = stdout;
                error.stderr = stderr
                reject(error);
                return;
            }
            resolve({stdout: stdout, stderr: stderr});
        });
    });
}
```
The command we exec is a Java command and we can see from the task manager the shell process is created and also the Java process, most of the times everything works as expected but sometimes **we can see the processes disappearing from the task manager and the callback is not called** (_finished_ is never printed in the console), causing a major issue with our application because we are not able to notify the user that the task has finished.

* Again, we see the processes disappearing from the task manager **and** the callback doesn't print anything
* It seems to be Windows specific, on Linux it doesn't happen.
* It's shouldn't be related to the cmd, it's always the same command with different args and, anyway, I'd expect the callback with an error if there's something wrong with cmd.
* It usually happens every 4 or 5 executions
* We downgraded to version 8.5.1 and it seems there issue is not there, we can run 20 tasks one after the other and we don't experience any issue
* Same thing happens with `util.promisify` of the require('child_process').exec;

Any idea?

W4545:
Can confirm I'm experiencing the same issue. 10% of the time the callback is not called.

hatton:
Same thing here, on Windows. I haven't tried on mac or linux yet. It can be very difficult for me to reproduce. Fortunately, the next time I do a `child_process.exec()`, the pending one will always return immediately. So for the moment I rely on this monstrosity:

```
  let didCallback = false;
  window.setTimeout(() => {
    if (!didCallback) {
      child_process.exec("call"); // does nothing on Windows, but causes a previously recalcitrant callback to fire
    }
  }, 2000);
```

codebytere:
[...]

Issue #25314, opened 2020-09-03
setImmediate works only on first page load, but not after subsequent reloads.


marcus13345:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** 
  * 10.1.0
* **Operating System:** 
  * windows 10
* **Last Known Working Electron version:** 
  * unknown

### Expected Behavior

`setImmediate` works the same regardless of whether you have loaded a page for the first time, or after a reload.

### Actual Behavior

`setImmediate` does execute on first page load, and does not after reloads.

### To Reproduce

I've created an [electron fiddle gist](https://gist.github.com/affedc59a009210f18d3d396bb9cc425) which shows the problem. The essence is that on first page load, setImmediate properly executes the given function, however, after reloading the page, with the same code, the invocation gets lost internally and never happens.

nornagon:
I can reproduce this on Windows, but not on macOS. 🤔

nornagon:
As a temporary workaround, `app.allowRendererProcessReuse = false` fixes this.

Issue #23910, opened 2020-06-02
[Win] libuv loop stops working in a native addon after page reload.


jphw:
### Issue Details

* **Electron Version:** 9.0.0 (beta versions of 9.0.0 too)
* **Operating System:** Windows 10 1903
* **Last Known Working Electron version:** 7.2.4

### Repro steps
1. **require** a native module in a Renderer process (on a page)
2. call libuv function **uv_queue_work** 5 times
3. get **uv_after_work_cb** invoked 5 times
4. reload the page 
5. call libuv function **uv_queue_work** 5 times once more

### Expected Behavior
6. get **uv_after_work_cb** invoked 5 times

### Actual Behavior
6. get **uv_after_work_cb** invoked 1 or 2 times, remaining ones never get invoked by libuv

### To Reproduce
I prepared a sample project with minimal code that demonstrates the problem:
>>> https://github.com/jphw/libuv_problems

Please read the README.MD and check src/libuv_problems_module.cpp sources.

### Additional Information
**app.allowRendererProcessReuse = false;**  works as a workaround for this issue

aabuhijleh:
I think this is causing the same issue I'm having #24173. Regression in Node.js APIs in preload script after page reload.

nornagon:
Related: #25314

ckerr:
This is still an issue in master & in the back branches even after the uv threading issue that was resolved in #25332. This appears to be an unrelated issue.

Jiahui-Yan:
This problem also exists in electron 12.0.0, but not in 9.4.4.

PR #25869, opened 2020-10-10
fix: uv loop polling when render process reuse enabled

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/22119.
Closes https://github.com/electron/electron/issues/25314.
Closes https://github.com/electron/electron/issues/23910.
Closes https://github.com/electron/electron/issues/25405.
Closes https://github.com/electron/electron/issues/25496.

Fixes an issue where some Node.js module API calls hung in the renderer process on reload when renderer process reuse was in effect. This was happening because when render process reuse was enabled, Chromium reused the same process, but the main thread was still bound to the old iocp (on windows, and respective interface for other platforms) because libuv didn't destroy it. The queue then went on to operate on this old iocp - when we reloaded, a new iocp was created but not bound to the main thread and so we would not properly be informed of events on it. This is fixed by preparing the message loop on _every_ reload when reuse is enabled with `node_bindings_->PrepareMessageLoop()`.

cc @zcbenz @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where some Node.js module API calls hung in the renderer process after reloads when render process reuse was enabled.

### DIFF
--- a/spec-main/api-browser-window-spec.ts
+++ b/spec-main/api-browser-window-spec.ts
@@ -4286,6 +4286,30 @@ describe('BrowserWindow module', () => {
     });
   });
 
+  describe('reloading with allowRendererProcessReuse enabled', () => {
+    it('does not cause Node.js module API hangs after reload', (done) => {
+      const w = new BrowserWindow({
+        show: false,
+        webPreferences: {
+          nodeIntegration: true
+        }
+      });
+
+      let count = 0;
+      ipcMain.on('async-node-api-done', () => {
+        if (count === 3) {
+          ipcMain.removeAllListeners('async-node-api-done');
+          done();
+        } else {
+          count++;
+          w.reload();
+        }
+      });
+
+      w.loadFile(path.join(fixtures, 'pages', 'send-after-node.html'));
+    });
+  });
+
   describe('window.webContents.focus()', () => {
     afterEach(closeAllWindows);
     it('focuses window', async () => {
--- a/shell/renderer/electron_renderer_client.cc
+++ b/shell/renderer/electron_renderer_client.cc
@@ -108,11 +108,11 @@ void ElectronRendererClient::DidCreateScriptContext(
 
   injected_frames_.insert(render_frame);
 
-  // If this is the first environment we are creating, prepare the node
-  // bindings.
   if (!node_integration_initialized_) {
     node_integration_initialized_ = true;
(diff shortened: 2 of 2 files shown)


zcbenz:
Busy looping is not a solution for listening to events. Apart from constant high CPU usage, when consuming streams it would add the delay for each chunk, making IO very slow.

codebytere:
@zcbenz that does make sense and i don't think this is the ideal solution either - it's not clear to me what specifically is lacking in the previous approach though. Do you have any thoughts on an alternative?

zcbenz:
I have not looked into the issue, but it seems that the issue is caused by ` GetQueuedCompletionStatus(uv_loop_->iocp)` no longer working after reloading. So I think there are few possibilities:

1. Node has created a new uv loop or iocp after reloading while we are still using the old one. (Unlikely though since other platforms are still working.)
2. Node does some initialization work to the iocp, after reloading the iocp was initialized again and stopped working.
3. After reloading Electron does some cleanup work and breaks iocp.

codebytere:
@zcbenz determined the true underlying cause - updated the solution & PR body accordingly

nornagon (approved):

nornagon:
Can we get a test?

spec-main/api-browser-window-spec.ts
```diff
@@ -4286,6 +4286,29 @@ describe('BrowserWindow module', () => {
...
+    it('does not cause Node.js module API hangs after reload', (done) => {
+      const w = new BrowserWindow({
+        show: false,
+        webPreferences: {
+          nodeIntegration: true
+        }
+      });
+
+      let count = 0;
+      ipcMain.on('async-node-api-done', () => {
```
zcbenz, line 4299: Should remove listener after test is over.

zcbenz (commented):
Test seems to be failing on mac CI:

```
Error: done() called multiple times
    at IpcMainImpl.<anonymous> (electron/spec-main/api-browser-window-spec.ts:4301:11)
    at Object.<anonymous> (electron/js2c/browser_init.js:157:8980)
```

spec-main/api-browser-window-spec.ts
codebytere: ah whoops, you're right!

zcbenz (approved):

PR #25874, opened 2020-10-10
feat: add webPreferences.enablePreferredSizeMode

samuelmaddock:
#### Description of Change

For Chrome extension popups and options windows, the size of their frame corresponds to the minimum size of their document's layout. To enable this, `RenderViewHost::EnablePreferredSizeMode` exists to notify the RenderView to send back updates with its size. This gives an interesting result where the document can dictate its frame's size.

In practice, it looks like this:

Example code:
```js
const view = new BrowserView({
  webPreferences: {
    preferredSizeMode: true,
  },
})
view.webContents.on('preferred-size-changed', (event, size) => {
  view.setBounds({ x: 0, y: 0, width: size.width, height: size.height })
})
```

I could see this being useful for specialized modal popups as well.

cc @electron/wg-api -- Question for you: is it worth adding tests for just adding bindings to existing Chromium functionality?

relates to #19447

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added `webPreferences.preferredSizeMode` to allow sizing views according to their document's minimum size.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -398,6 +398,10 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       * `bypassHeatCheck` - Bypass code caching heuristics but with lazy compilation
       * `bypassHeatCheckAndEagerCompile` - Same as above except compilation is eager.
       Default policy is `code`.
+    * `preferredSizeMode` Boolean (optional) - Whether to enable preferred size
+      mode. Enabling this will cause the `preferred-size-changed` event to be
+      emitted on the `WebContents` when the minimum size of the document's
+      layout changes.
 
 When setting minimum or maximum window size with `minWidth`/`maxWidth`/
 `minHeight`/`maxHeight`, it only constrains the users. It won't prevent you from
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -399,9 +399,10 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       * `bypassHeatCheckAndEagerCompile` - Same as above except compilation is eager.
       Default policy is `code`.
     * `preferredSizeMode` Boolean (optional) - Whether to enable preferred size
-      mode. Enabling this will cause the `preferred-size-changed` event to be
-      emitted on the `WebContents` when the minimum size of the document's
-      layout changes.
+      mode. The preferred size is the minimum size to contain the layout of the
+      document—without requiring scrolling. Enabling this will cause the
+      `preferred-size-changed` event to be emitted on the `WebContents` when
+      the preferred size changes. Default is `false`.
 
 When setting minimum or maximum window size with `minWidth`/`maxWidth`/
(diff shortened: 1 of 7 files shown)


shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1173,6 +1173,11 @@ void WebContents::DidStartLoading() {
 
 void WebContents::DidStopLoading() {
   Emit("did-stop-loading");
+
+  auto* web_preferences = WebContentsPreferences::From(web_contents());
+  if (web_preferences &&
+      web_preferences->IsEnabled(options::kPreferredSizeMode))
+    web_contents()->GetRenderViewHost()->EnablePreferredSizeMode();
```
samuelmaddock, line 1180: The location for calling this is based on Chromium's implementation: https://source.chromium.org/chromium/chromium/src/+/master:components/guest_view/browser/guest_view_base.cc;l=573;drc=d1252b1460e53a48e235c3bd9555bf5926c49100

sentialx:
Great job! Could this be backported to 9-x-y?

samuelmaddock:
> 
> 
> Great job! Could this be backported to 9-x-y?

At the will of @electron/wg-releases, probably. I'd like to target `11-x-y` at the least.

docs/api/browser-window.md
```diff
@@ -398,6 +398,10 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       * `bypassHeatCheck` - Bypass code caching heuristics but with lazy compilation
       * `bypassHeatCheckAndEagerCompile` - Same as above except compilation is eager.
       Default policy is `code`.
+    * `preferredSizeMode` Boolean (optional) - Whether to enable preferred size
```
zcbenz, line 401: Should document the default value.

Also I think there should be some descriptions on what "preferred size" is and what does it do, it is not very clear from current documents.

zcbenz (changes requested):

docs/api/browser-window.md
samuelmaddock: How's this? 97148e8

zcbenz (approved):

docs/api/browser-window.md
samuelmaddock: Added one minor improvement in 1daae04

ckerr (approved):
This is pretty nice.

ckerr:
linux CI failure seems  unrelated: `E1013 02:31:23.397542 22963 gomacc.cc:409] GOMA: compiler proxy not working?`

jkleinsc:
The API WG approved this at our Oct 19 2020 meeting.

codebytere (approved):

samuelmaddock:
@codebytere oof, I just realized the notes were not updated to reflect the renamed API. It should be `webPreferences.enablePreferredSizeMode`. How would one go about correcting that in the persisted release notes?

Issue #25888, opened 2020-10-12
electron crash when using node vm


CaiHuan:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
 9.3.1
* **Operating System:**
 Windows 10 (1904)

### Expected Behavior
 print error on dev devtools 

### Actual Behavior
 electron crash
### To Reproduce

1: git clone https://github.com/electron/electron-quick-start
2: cd electron-quick-start
3: Install dependencies
4: npm install
5: add nodeintegration=true webPreferences(main.js)
``` javascript
const mainWindow = new BrowserWindow({
    width: 800,
    height: 600,
    webPreferences: {
      preload: path.join(__dirname, 'preload.js'),
      nodeIntegration: true
    }
  })
```
5: add script to renderer.js
``` javascript
const vm = require('vm');

const script = new vm.Script('console.log(asasa);');

const sandbox = {globalVar: 1, console: console};

const contetifiedSandbox = vm.createContext(sandbox);

const result = script.runInContext(contetifiedSandbox);
```

### Screenshots
![image](https://user-images.githubusercontent.com/3890047/95700818-8b724f80-0c7a-11eb-89d4-8c5396408955.png)

codebytere:
Stacktrace:
```
Received signal 11 <unknown> 000000000000

0   Electron Framework                  0x000000011ef09cc9 base::debug::CollectStackTrace(void**, unsigned long) + 9

1   Electron Framework                  0x000000011ee04663 base::debug::StackTrace::StackTrace() + 19

2   Electron Framework                  0x000000011ef09b91 base::debug::(anonymous namespace)::StackDumpSignalHandler(int, __siginfo*, void*) + 2385

3   libsystem_platform.dylib            0x00007fff6adfe5fd _sigtramp + 29

4   ???                                 0x0000000000000000 0x0 + 0

5   Electron Framework                  0x000000012172ca5c blink::WorldSafeV8ReferenceInternal::MaybeCheckCreationContextWorld(blink::DOMWrapperWorld const&, v8::Local<v8::Value>) + 60

[...]

PR #25910, opened 2020-10-13
fix: ScriptState crash for non-Blink v8::Contexts

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/25888.

In the renderer process it's possible that a `v8::Context` will be created by Node.js and therefore not have an associated `ScriptState`. Blink assumes one will be associated always, and thus crashes when trying to
fetch the `DOMWrapperWorld` for a `ScriptState` created from a non-Blink `v8::Context`.

This adds extra safety checks to ensure that `ScriptContext::From` correctly returns `nullptr` for those cases.

Tested with https://gist.github.com/7d0072e815f31a0ef54a625bd0b17d96.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a potential crash when creating and using a new `vm.Context` in the renderer process.


CaiHuan:
but the code will hit the electron crash(Utils::ApiCheck) because i::EmbedderDataSlot(*data, index).ToAlignedPointer(&result) always false

ckerr (approved):

codebytere:
@CaiHuan i didn't experience any issues when testing with https://gist.github.com/7d0072e815f31a0ef54a625bd0b17d96 after this fix

CaiHuan:
> > [69847:1013/095537.481276:INFO:CONSOLE(1)] "10", source: evalmachine. (1)
the problem is post exception on node VM.  console.log can print an integer value.  you should test code like console.log(aaa).
the enviroment will post exception and hit the crash after cannot find aaa object

CaiHuan:
const vm = require('vm');
fs = require('fs')
timer = require('timers')
var script_data = "function print() {\
  console.log(sds);\
}\
setTimeout(print, 1000);";
const script = new vm.Script(script_data);

const sandbox = {setTimeout: setTimeout, console: console};

const contetifiedSandbox = vm.createContext(sandbox);

script.runInContext(contetifiedSandbox);

the case also crash because setTimeout will checkout world，but electron just replace web setTimeout with node setTimeout on the browser process

CaiHuan:
> @CaiHuan i didn't experience any issues when testing with https://gist.github.com/7d0072e815f31a0ef54a625bd0b17d96 after this fix -
> 
> > [69847:1013/095537.481276:INFO:CONSOLE(1)] "10", source: evalmachine. (1)
> 
> was successfully printed to console.

the problem is post exception on node VM. console.log can print an integer value. you should test code like console.log(aaa).
the enviroment will post exception and hit the crash after cannot find aaa object.
console.log print undefined object not basic type
release version will crash on SECURITY_CHECK(script_state);
debug version will crash on Utils::ApiCheck(i::EmbedderDataSlot(*data, index).ToAlignedPointer(&result))

CaiHuan:
> @CaiHuan i didn't experience any issues when testing with https://gist.github.com/7d0072e815f31a0ef54a625bd0b17d96 after this fix -
> 
> > [69847:1013/095537.481276:INFO:CONSOLE(1)] "10", source: evalmachine. (1)
> 
> was successfully printed to console.

const vm = require('vm');
fs = require('fs')
timer = require('timers')
var script_data = "function print() {
console.log(sds);
}
setTimeout(print, 1000);";
const script = new vm.Script(script_data);

const sandbox = {setTimeout: setTimeout, console: console};

const contetifiedSandbox = vm.createContext(sandbox);

script.runInContext(contetifiedSandbox);

the case also crash because setTimeout will checkout world，but electron just replace web setTimeout with node setTimeout on the browser process

CaiHuan:
> #### Description of Change
> Closes #25888.
> 
> In the renderer process it's possible that a `v8::Context` will be created by Node.js and therefore not have an associated `ScriptState`. Blink assumes one will be associated always, and thus crashes when trying to
> fetch the `DOMWrapperWorld` for a `ScriptState` created from a non-Blink `v8::Context`.
> 
> This adds extra safety checks to ensure that `ScriptContext::From` correctly returns `nullptr` for those cases.
> 
> Tested with https://gist.github.com/7d0072e815f31a0ef54a625bd0b17d96.
> 
> #### Checklist
> * [x]  PR description included and stakeholders cc'd
> * [x]  `npm test` passes
> * [x]  PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
> * [x]  [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).
> 
> #### Release Notes
> Notes: Fixed a potential crash when creating and using a new `vm.Context` in the renderer process.

const vm = require('vm');
fs = require('fs')
timer = require('timers')
var script_data = "function print() {
console.log(sds);
}
setTimeout(print, 1000);";
const script = new vm.Script(script_data);

const sandbox = {setTimeout: setTimeout, console: console};

const contetifiedSandbox = vm.createContext(sandbox);

script.runInContext(contetifiedSandbox);

the case also crash because setTimeout will checkout world，but electron just replace web setTimeout with node setTimeout on the browser process

zcbenz (commented):
Patching blink like this would be an endless game, if we fixed `console.log` it is still possible to crash other unpatched code by passing things like `WebAudio`.

Even if we managed to patch all places of blink to ignore non-blink contexts, I think it would then raise questions about security.

codebytere:
yeah - we may honestly want to simply leave this broken and mark it somewhere in docs

nornagon:
I agree with @zcbenz here, blink surely assumes in many places that all Contexts come with a ScriptState, and we're never going to be able to patch out all those assumptions.

Using node's `vm` module in the renderer process is asking for trouble.

MarshallOfSound:
> Using node's vm module in the renderer process is asking for trouble.

So let's deprecate it? The web platform has proposals for a worlds API iirc so The Web Way isn't a too distant future

codebytere:
we already override some stuff in the renderer - seems reasonable to intercept `vm` and prevent its usage there. I'll close this out and do that.

gzorin:
> > Using node's vm module in the renderer process is asking for trouble.
> 
> So let's deprecate it? The web platform has proposals for a worlds API iirc so The Web Way isn't a too distant future

I'm pretty interested in learning more about such a proposal, since at Bloomberg we've needed to hack Chromium quite a bit to get ordinary V8 contexts to cooperate with those created by Blink.

I couldn't find a link to any discussion or documents; do you have any references?

Issue #25649, opened 2020-09-27
mouse events in 'drag' and 'no-drag' unaligned after 10.1.1


rafma0:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  10.1.2
* **Operating System:**
  Windows 10 (2004)
* **Last Known Working Electron version:**
  10.1.0

### Expected Behavior
mouse events aligned to the element

### Actual Behavior
elements with 'drag' and 'no-drag' inside a 'drag' region are 'vertically unaligned' by 8 pixels for mouse hover, click, etc when window is maximized

### To Reproduce
https://gist.github.com/rafma0/e7bc9ad86c17e79b25130b4399d72506
double click any appbar to maximize
move your mouse to the bottom of any button, it wont change color until you go 8 pixels up
you can also click and drag from 8 pixels outside the top of the middle or bottom appbar and it will unmaximize

nornagon:
potentially related to https://github.com/electron/electron/pull/25052. cc @deepak1556

PR #25052, opened 2020-08-20
fix: client area inset calculation when maximized for framless windows

deepak1556:
#### Description of Change

Fixes the following regressions introduced by https://github.com/electron/electron/pull/21164

Refs

https://github.com/microsoft/vscode/issues/86260
https://github.com/microsoft/vscode/issues/85310
https://github.com/microsoft/vscode/issues/85592
https://github.com/microsoft/vscode/issues/85655

Also fixes a side effect of this bug that caused increased usage of GPU under certain conditions which was identified by teams app.

Gist: https://gist.github.com/deepak1556/6e672eea191d3e5f50e429d312a56446

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).
- [x] This is **NOT A BREAKING CHANGE**. Breaking changes may not be merged to master until 11-x-y is branched.

#### Release Notes

Notes:  Fixes the following issues for frameless when maximized on Windows
* fix unreachable task bar when auto hidden with position top
* fix 1px extending to secondary monitor
* fix 1px overflowing into taskbar at certain resolutions
* fix white line on top of window under 4k resolutions

Issue #1821, opened 2015-05-29
Electron windows causing repaint issues when DWM composition disabled


CharlieHess:
- Log on to a Windows 7 machine
- Switch your theme to Windows Basic or Windows Classic (this has the side-effect of disabling Aero's DWM composition)
- Launch an Electron app
- Switch to another app window
- The other window fails to repaint:

![window repaint issue](https://cloud.githubusercontent.com/assets/1865957/7892783/3bd6c34a-060c-11e5-8746-3815d26d6590.PNG)

This occurs with almost every app that I've tried, so it seems like something funky with the Electron window. It may or may not require the presence of a `webview` tag; I haven't isolated that yet. Until now we've been advising users to switch to an Aero theme (this fixes the problem, but is undesirable for many).

CharlieHess:
Also an issue with any remote desktop session, since that typically disables DWM composition.

jakebauer:
This is horrendous on a window sitting above outlook. The only real fix is to maximize and minimize slack, since there's so many panes in outlook you have to swirl your mouse pointer around in to clear the screen of repaint artifacts. Disabling gpu in both outlook and slack doesn't change anything either.

Sorry for the obfuscation in the images, I've provided extra to show the detail just in case.

![2015-05-16 00_45_46](https://cloud.githubusercontent.com/assets/12820558/8070076/9fd852d2-0eb3-11e5-81c9-fef652f013f2.png)
![2015-06-09 14_32_40-2015-05-17 22_01_40](https://cloud.githubusercontent.com/assets/12820558/8070214/a58e30f6-0eb4-11e5-94aa-7a50a44f3082.png)
![2015-06-09 14_30_06-2015-05-22 10_27_52](https://cloud.githubusercontent.com/assets/12820558/8070215/a5911546-0eb4-11e5-86e7-4f020c397427.png)
![2015-06-09 14_34_43-2015-05-22 11_31_11](https://cloud.githubusercontent.com/assets/12820558/8070246/d28ec9c6-0eb4-11e5-8169-7db70b430d0f.png)

jakebauer:
replaced with https://github.com/rfgamaral/SlackUI - works flawlessly

zcbenz:
@CharlieHess Does this happen for the default app? I tried with v0.29.1 on a fresh Window 7 image with basic theme, and the default app seems to work well.

CharlieHess:
@zcbenz I think it's exclusive to the `webview` tag.

[...]

PR #25940, opened 2020-10-14
fix: maximized frameless window bleeding to other monitors

zcbenz:
#### Description of Change

This is a followup to https://github.com/electron/electron/pull/25052, to fix a bug on Windows that, maximized frameless window would have its edge bleeding to other monitors.

The previous fix did not work with multiple monitors with different DIPs, this PR should be able to fix that.

Note that this bug causes some maximized apps to consume much more GPU resources, and this PR is needed by old branches to fix the performance issue.

Fixes https://github.com/electron/electron/issues/25649

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix maximized frameless window bleeding to other monitors.

### DIFF
--- a/shell/browser/ui/win/electron_desktop_window_tree_host_win.cc
+++ b/shell/browser/ui/win/electron_desktop_window_tree_host_win.cc
@@ -6,10 +6,7 @@
 
 #include "base/win/windows_version.h"
 #include "shell/browser/ui/views/win_frame_view.h"
-#include "ui/base/win/hwnd_metrics.h"
 #include "ui/base/win/shell.h"
-#include "ui/display/win/screen_win.h"
-#include "ui/views/win/hwnd_util.h"
 
 namespace electron {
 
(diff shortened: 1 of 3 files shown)


shell/browser/ui/win/electron_desktop_window_tree_host_win.cc
```diff
@@ -41,23 +38,24 @@ bool ElectronDesktopWindowTreeHostWin::HasNativeFrame() const {
   // Since we never use chromium's titlebar implementation, we can just say
   // that we use a native titlebar. This will disable the repaint locking when
   // DWM composition is disabled.
-  return !ui::win::IsAeroGlassEnabled();
+  return false;
```
ckerr, line 41: Should the three-line comment above be removed? It looks like this change invalidates the comment since `return false;` here is necessary for enabling repaint locking, not disabling it:

```cc
          should_lock_(owner_->IsVisible() && !owner->HasChildRenderingWindow() &&
                       ::IsWindow(hwnd_) &&
                       !owner_->HasNativeFrame() &&
```

shell/browser/ui/win/electron_desktop_window_tree_host_win.cc
```diff
@@ -41,23 +38,24 @@ bool ElectronDesktopWindowTreeHostWin::HasNativeFrame() const {
...
-    insets->Set(frame_height, frame_size, frame_size, frame_size);
+    // This would be equivalent to calling:
+    // DwmExtendFrameIntoClientArea({0, 0, 0, 0});
+    //
+    // which means do not extend window frame into client area. It is almost
+    // a no-op, but it can tell Windows to not extend the window frame to be
+    // larger than current workspace.
+    //
+    // See also:
+    // https://devblogs.microsoft.com/oldnewthing/20150304-00/?p=44543
```
ckerr, line 57: :+1:

ckerr (approved):

shell/browser/ui/win/electron_desktop_window_tree_host_win.cc
deepak1556: This seems like it will bring back https://github.com/electron/electron/issues/1821 for windows 7

shell/browser/ui/win/electron_desktop_window_tree_host_win.cc
```diff
@@ -66,24 +64,20 @@ bool ElectronDesktopWindowTreeHostWin::GetDwmFrameInsetsInPixels(
...
-      insets->Set(0, 0, border_thickness, border_thickness);
-    } else {
-      const int frame_thickness = ui::GetFrameThickness(monitor);
-      // Reduce the Windows non-client border size because we extend the border
-      // into our client area in UpdateDWMFrame(). The top inset must be 0 or
-      // else Windows will draw a full native titlebar outside the client area.
-      insets->Set(0, frame_thickness, frame_thickness, frame_thickness);
-    }
+    // The insets would be eventually passed to WM_NCCALCSIZE, which takes
+    // the metrics under the DPI of _main_ monitor instead of current moniotr.
```
deepak1556, line 74: surprising find, should this be reported to upstream ?

deepak1556 (approved):

shell/browser/ui/win/electron_desktop_window_tree_host_win.cc
zcbenz: Nice catch, I did not actually intend to change this line.

shell/browser/ui/win/electron_desktop_window_tree_host_win.cc
zcbenz: This only seems to be a problem for maximized frameless window, which Chrome only uses for deprecated Chrome Apps APIs. I can't find other way to reproduce it in Chrome .

PR #25943, opened 2020-10-14
build: fix missing dependency resulting in a random build failure

mzajaczkowski:
#### Description of Change
* build: pref_names.h #includes files generated by separate GN targets,
  any sources that include that file have to live in a target that depends
  on the target it comes from

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/chromium_src/BUILD.gn
+++ b/chromium_src/BUILD.gn
@@ -367,6 +367,7 @@ source_set("chrome_spellchecker") {
   sources = []
   deps = []
   libs = []
+  public_deps = []
 
   if (enable_builtin_spellchecker) {
     sources += [
@@ -406,15 +407,13 @@ source_set("chrome_spellchecker") {
       "//components/spellcheck:buildflags",
       "//components/sync",
     ]
+
+    public_deps += [ "//chrome/common:constants" ]
   }
 
-  public_deps = [
+  public_deps += [
     "//components/spellcheck/browser",
     "//components/spellcheck/common",
     "//components/spellcheck/renderer",
   ]
-
-  if (enable_builtin_spellchecker) {
-    public_deps += [ "//chrome/common:constants" ]
-  }
 }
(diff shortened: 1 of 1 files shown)


chromium_src/BUILD.gn
```diff
@@ -414,4 +413,8 @@ source_set("chrome_spellchecker") {
     "//components/spellcheck/common",
     "//components/spellcheck/renderer",
   ]
+
+  if (enable_builtin_spellchecker) {
+    public_deps += [ "//chrome/common" ]
```
ckerr, line 418: I might be reading the deps, it looks like pref_names.h comes in from `//chrome/common:constants`. Would it make sense to use that instead of `//chrome/common` here to pick up the various `#includes` deps?

MarshallOfSound: ☝️ In agreement, we shouldn't depend on all of `//chrome/common` here

mzajaczkowski: Yes, you are right, my bad. Following up with a fixup.

ckerr (approved):

MarshallOfSound (approved):

chromium_src/BUILD.gn
```diff
@@ -414,4 +413,8 @@ source_set("chrome_spellchecker") {
     "//components/spellcheck/common",
     "//components/spellcheck/renderer",
   ]
+
+  if (enable_builtin_spellchecker) {
+    public_deps += [ "//chrome/common:constants" ]
```
deepak1556, line 418: can you declare `public_deps = []`


and move this inside the previous `enable_builtin_spellchecker` block, would like to avoid repeated condition blocks.

deepak1556 (changes requested):

chromium_src/BUILD.gn
mzajaczkowski: Fixed up.

deepak1556 (approved):
thanks!

deepak1556:

PR #18286, opened 2019-05-13
feat: enable macOS apps to receive push notifications from APNs

kevinhu88:
#### Description of Change
I'm picking up where @frantic left off in https://github.com/electron/electron/pull/13777 :)

This exposes some API for registering for [Apple Push Notification service](https://developer.apple.com/documentation/usernotifications/registering_your_app_with_apns?language=objc) to let macOS apps receive push notifications even when the app is not running. Here is an [example of electron-quick-start](https://github.com/kevinhu88/electron-quick-start) that uses the stuff in this PR to register for APNS, receive a remote notification, launch the app from a user action, and displays the notification contents:
Screen Shot 2019-05-13 at 11 33 09 AM
Screen Shot 2019-05-13 at 11 33 23 AM
Screen Shot 2019-05-13 at 11 33 37 AM

This PR uses the [UserNotifications Framework](https://developer.apple.com/documentation/usernotifications?language=objc), which allows us to launch apps from custom user actions in the push notification, optionally with text input, on macOS mojave. **This requires building Electron with the macOS SDK 10.14+.** Launching the app from a notification on mojave results in a `UNNotificationResponse` mapped to `[notify userInfo][NSApplicationLaunchUserNotificationKey]` in `applicationDidFinishLaunching`, instead of `NSUserNotification`. This PR accounts for this and adds a function `UNNotificationResponseToNSDictionary` to convert the `UNNotificationResponse` to send as the `launch_info` dictionary with the `did-finish-launching` event so the Electron app can handle the notification payload.

As with any other macOS app, enabling push notifications requires setting up a provisioning profile with the APS Environment (macOS) Entitlement enabled in your Apple developer account. See [this doc](https://developer.apple.com/documentation/usernotifications/registering_your_app_with_apns?language=objc) for more info about setup. Because of this requirement, I'm not sure what other test cases I could add besides testing failed registration.

I'm looking forward to feedback on this and really hope we can get this feature into the official Electron release!

#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
[...]

PR #25950, opened 2020-10-14
feat: add support for UNNotificationResponse in app 'ready' event

miniak:
#### Description of Change
Fill `launchInfo` of `app` event `ready` with information from `UNNotificationResponse` when the app is launched via a notification presented by `UNUserNotificationCenter` (in a native node module).
Motivated by #18286. Follow-up to #25702.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: The `launchInfo` of `app` event `ready` is now populated with information from `UNNotificationResponse` when the app is launched via a notification presented by `UNUserNotificationCenter` (in a native node module).


BUILD.gn
```diff
@@ -480,6 +480,7 @@ source_set("electron_lib") {
       "SecurityInterface.framework",
       "ServiceManagement.framework",
       "StoreKit.framework",
+      "UserNotifications.framework",
```
MarshallOfSound, line 483: We can't do this, afaik just _linking_ to this framework breaks some usages of the `NSUserNotification` API.  We would need a complete migration to the new framework which is a larger amount of work we don't want to do right now

MarshallOfSound (changes requested):

BUILD.gn
miniak: Do you know which ones?

miniak: This is important for us, as we have a custom node module using the `UserNotifications.framework` and we are not getting the notification information when the app is launched from the notification. I am willing to spend time on the migration if that's necessary.

MarshallOfSound: @miniak I'm not 100% which ones, but to be safe we should (a) test it and figure it out or (b) start the process of using `@available` declarations and use the new `UN` APIs on 10.14 and above where they are available.

MarshallOfSound: @miniak I've thought about this a bit and you should actually be able to implement this entire thing using fake declarations, `respondsToSelector` and static casts without ever actually linking to the `UN` framework

miniak: btw how do you know that just linking `UserNotifications.framework` will caused `NSUserNotification` based  notifications to stop working?

miniak: @MarshallOfSound I've changed the code to load the framework dynamically, is that enough? Or do we have to avoid it completely?

MarshallOfSound: > btw how do you know that just linking UserNotifications.framework will cause NSUserNotification based notifications to stop working?

Horror stories from a while ago, I can't find the reference right now.

> I've changed the code to load the framework dynamically, is that enough? Or do we have to avoid it completely?

I think you have to avoid the symbols completely 🤔

miniak: I don't load the framework anymore, symbols are only needed to get the class for instance check, which I've replaced  by checking for selectors.

MarshallOfSound (approved):

deepak1556 (approved):

alexeykuzmin (approved):

PR #25958, opened 2020-10-15
perf: remove GC timer that fired once per minute.

### LINKED ISSUE #25954: ElectronBrowserMainParts wakes up every minute even if everything is idle
Found while investigating #25259 

ElectronBrowserMainParts has a periodic timer that is used to trigger GC once a minute even on an idle system. While one wakeup per minute isn't bad, it's also been in there [since 2014](https://github.com/electron/electron/commit/4dd78480846557370dd53789d50e6cb0d6a81bcd) and may or may not still be necessary. It might be worth seeing what happens to memory overhead if this is disabled and/or triggered with a different mechanism.

```c++
  // Start idle gc.
  gc_timer_.Start(
      FROM_HERE, base::TimeDelta::FromMinutes(1),
      base::Bind(&v8::Isolate::LowMemoryNotification,
                 base::Unretained(js_env_->isolate())));
```

ckerr:
#### Description of Change

Fixes #25954

ElectronBrowserMainParts has a periodic timer that is used to trigger GC once a minute even on an idle system. While one wakeup per minute isn't bad, it's also been in there since 2014 and doesn't seem to be necessary.

This heartbeat doesn't seem to be necessary, at least for a simple system. I see no difference in memory use with or without this timer when running the default app for an hour.

Any review welcomed. CC'ing @nornagon as he commented on #25954

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none.

### DIFF
--- a/shell/browser/electron_browser_main_parts.cc
+++ b/shell/browser/electron_browser_main_parts.cc
@@ -405,11 +405,6 @@ void ElectronBrowserMainParts::PreMainMessageLoopRun() {
   ui::TouchFactory::SetTouchDeviceListFromCommandLine();
 #endif
 
-  // Start idle gc.
-  gc_timer_.Start(FROM_HERE, base::TimeDelta::FromMinutes(1),
-                  base::BindRepeating(&v8::Isolate::LowMemoryNotification,
-                                      base::Unretained(js_env_->isolate())));
-
   content::WebUIControllerFactory::RegisterFactory(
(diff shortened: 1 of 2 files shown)


MarshallOfSound (approved):
Lol, weird 😆

nornagon (approved):

codebytere (approved):

zcbenz (approved):

PR #25986, opened 2020-10-16
chore: bump @typescript-eslint version

dsanders11:
#### Description of Change
Gets rid of the warning on linting JS about unsupported TS version, and supports TS 4 so that's nice. Getting rid of the warning is important for another PR I've got in the pipeline.

#### Checklist

#### Release Notes

Notes: none


dsanders11:
Gah. I ran the linter locally, but there's some issues with linters failing silently on Windows (another PR to come). Looks like it didn't actually run so I thought it was clean.

dsanders11:
Turns out the reason it looked fine locally was because of `.eslintcache`.

Updated `.eslintrc.json` so that linting passes cleanly on `master`. The only changes were:
  * Defining more globals (see [this FAQ on their repo for an explanation of that changing in v4](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/FAQ.md#i-get-errors-from-the-no-undef-rule-about-global-variables-not-being-defined-even-though-there-are-no-typescript-errors)).
  * Disabling `no-use-before-define` which was now having errors due to types used before declaration. This could be fixed with `@typescript-eslint/no-use-before-define`, but that caused more errors due to improved analysis in v4 (see ["TypeScript Scope Analysis"](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v4.0.0) for more details).

To enable `@typescript-eslint/no-use-before-define` there would be ~85 errors to fix, which should just be rearranging code in the files, so not too difficult, but I'm punting on that for now.

MarshallOfSound (commented):
Instead of adding all these global types can we instead rely on typescript as the eslint docs suggest

> We strongly recommend that you do not use the no-undef lint rule on TypeScript projects. The checks it provides are already provided by TypeScript without the need for configuration - TypeScript just does this significantly better.

Let's turn of `no-undef` for `.ts` files

dsanders11:
@MarshallOfSound, you're right, that's much better. I should have read that FAQ answer to the end. Change made, CI lint check ran clean.

zcbenz (approved):

codebytere (approved):

PR #26007, opened 2020-10-16
build: lint patches for trailing whitespace

dsanders11:
#### Description of Change
Perhaps a bit trivial, but there are warnings when applying the patches from git due to trailing whitespace, and no warnings is better than some. So this PR trims the trailing whitespace and adds a check to the patches linter to keep trailing whitespace out.

#### Checklist

#### Release Notes

Notes: none


dsanders11:
I'm not familiar with working with the patches so I'm guessing directly editing them wasn't the right way to do it, even if it was just removing whitespace.

codebytere (approved):
Thanks! we get a bit lazy about this, always appreciate the fix 😆 you can also run `e patches all` to regen the patches locally before push and avoid this issue (or scope it down more, `e patches node`, `e patches chromium` etc)

dsanders11:
@codebytere, thanks for the tip. I just didn't think about how the `index` line would change and how that would cause a mismatch when re-exporting the patches. Makes sense, and now I know - can't just edit the patches as is.

zcbenz (approved):

PR #26010, opened 2020-10-17
fix: support new PDF viewer update

codebytere:
#### Description of Change

Adds missing support for new PDF viewer UI.

Enable-able with `app.commandLine.appendSwitch('enable-features', 'PDFViewerUpdate')`.

Default:
Screen Shot 2020-10-16 at 6 27 44 PM

With feature enabled:

Screen Shot 2020-10-16 at 6 27 32 PM

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Enabled support for new Chromium experimental PDF viewer UI.


deepak1556 (approved):

jkleinsc (approved):

codebytere:
> document.visibilityState should be hidden when the window is initially shown but hidden before the page is loaded - should be hidden when the window is initially shown but hidden before the page is loaded

is unrelated.

PR #26014, opened 2020-10-17
build: fix running eslint on Windows

dsanders11:
#### Description of Change
This one probably needs a more thorough review and discussion, I don't love it.

First part of the fix is simple, use `shell: true` so that the command runs correctly on Windows, this is what was stopping it before.

However, once that was fixed it immediately runs into "The command line is too long." The full command with the list of hundreds of files maxes out Window's max length for a command line invocation. Looking at the [eslint command line options](https://eslint.org/docs/user-guide/command-line-interface) I don't see any other way to give it the list of files. There's `--stdin` where you could dump all the contents, but then you lose filenames, so that's not going to work.

~~So this PR runs the files one at a time through `eslint`. I don't know if there's much of a performance impact of doing so. A more complex implementation could chunk filenames into groups small enough to run on the Windows command line, either by keeping track of the cumulative length or just choose a reasonable N value for the number of filenames to run in one go. This could also be limited to just Windows and let other platforms run as they currently do, but I do wonder if eventually a similar length limit could be hit on other platforms given the length of filenames and a (presumably) ever growing list of them. EDIT: Looks like the lint took 3x as long (not sure why it failed) after this change, so I guess that answers the question about the performance impact. I'll look at refactoring it to do it in chunks of files rather than one at a time.~~

The latest version of this PR chunks filenames to ensure they don't run past the max command line length. On Windows that's 2047 and for any other platform the code assumes a max of 100kb.

This PR also needs (IMO) #25986, otherwise each invocation of `eslint` will output a warning which quickly blows out to a large amount of output. There's no way to silence that warning, so I just updated to the latest `@typescript-eslint`.

cc @codebytere @MarshallOfSound @nornagon 

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/script/lint.js
+++ b/script/lint.js
@@ -98,9 +98,39 @@ const LINTERS = [{
   test: filename => filename.endsWith('.js') || filename.endsWith('.ts'),
   run: (opts, filenames) => {
     const cmd = path.join(SOURCE_ROOT, 'node_modules', '.bin', 'eslint');
-    const args = ['--cache', '--ext', '.js,.ts', ...filenames];
+    const args = ['--cache', '--ext', '.js,.ts'];
     if (opts.fix) args.unshift('--fix');
-    spawnAndCheckExitCode(cmd, args, { cwd: SOURCE_ROOT });
+    // Windows has a max command line length of 2047 characters, so we can't provide
+    // all of the filenames without going over that. To work around it, run eslint
+    // multiple times and chunk the filenames so that each run is under that limit.
+    // Use a much higher limit on other platforms which will effectively be a no-op.
+    const MAX_FILENAME_ARGS_LENGTH = IS_WINDOWS ? 1900 : 100 * 1024;
+    const cmdOpts = { stdio: 'inherit', shell: IS_WINDOWS, cwd: SOURCE_ROOT };
+    if (IS_WINDOWS) {
+      // When running with shell spaces in filenames are problematic
+      filenames = filenames.map(filename => `"${filename}"`);
+    }
+    const chunkedFilenames = filenames.reduce((chunkedFilenames, filename) => {
+      const currentChunk = chunkedFilenames[chunkedFilenames.length - 1];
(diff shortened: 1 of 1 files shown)


dsanders11:
This has been updated to run the files on Windows in chunks to avoid the max command line length, and do all in one go on other platforms. I believe it only ends up taking three or four runs on Windows with the file names chunked up.

script/lint.js
```diff
@@ -98,9 +98,38 @@ const LINTERS = [{
   test: filename => filename.endsWith('.js') || filename.endsWith('.ts'),
   run: (opts, filenames) => {
     const cmd = path.join(SOURCE_ROOT, 'node_modules', '.bin', 'eslint');
-    const args = ['--cache', '--ext', '.js,.ts', ...filenames];
+    const args = ['--cache', '--ext', '.js,.ts'];
     if (opts.fix) args.unshift('--fix');
-    spawnAndCheckExitCode(cmd, args, { cwd: SOURCE_ROOT });
+    if (IS_WINDOWS) {
```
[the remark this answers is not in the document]
dsanders11: Fixed.

script/lint.js
```diff
@@ -98,9 +98,39 @@ const LINTERS = [{
...
+    const allOk = chunkedFilenames.map(filenames => {
+      const result = childProcess.spawnSync(cmd, [...args, ...filenames], cmdOpts);
+      if (result.error) {
+        console.error(result.error);
+        process.exit(result.status || 1);
+      }
+      return result.status === 0;
+    }).every(x => x);
+    if (!allOk) {
+      process.exit(1);
```
nornagon, line 132: I think this will never be hit? It looks like we `process.exit` in the loop if any individual execution fails

dsanders11: The loop only checks `result.error`. I believe that's only set if there's an invocation issue or a time out. A non-zero exit code from `eslint` wouldn't result in `result.error` being set. Pretty sure on that, since when it was failing on the filename with a space in it during testing the loop continued and it went on to the next run.

nornagon (approved):

codebytere (approved):

PR #26036, opened 2020-10-19
chore: backport fix for ARIA TreeGrid rows not being announced by VoiceOver

nornagon:
Expose AXIndex for accessibility TreeGrid and Tree rows on Mac

VoiceOver requires each row to have in index in order to announce it. We
expose this property for rows which either have role=row or
role=treeitem.

AX-Relnotes: Fix an issue where ARIA TreeGrid and Tree rows were not announced.
Bug: 1115267
Change-Id: Ia8917044a510b2467aa2480dff54fe226b90c391
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2426569
Auto-Submit: Martin Robinson <mrobinson@igalia.com>
Commit-Queue: Martin Robinson <mrobinson@igalia.com>
Reviewed-by: Dominic Mazzoni <dmazzoni@chromium.org>
Reviewed-by: Abigail Klein <abigailbklein@google.com>
Cr-Commit-Position: refs/heads/master@{#816089}

Notes: Backported fix for https://crbug.com/1115267.

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -99,3 +99,4 @@ remove_some_deps_that_do_not_work_on_arm64.patch
 fix_check_issecureeventinputenabled_in_constructor_before_setting.patch
 skip_atk_toolchain_check.patch
 worker_feat_add_hook_to_notify_script_ready.patch
+cherry-pick-2f5b8357dca2.patch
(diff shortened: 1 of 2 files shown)


deepak1556 (approved):
Is this only needed for 11-x-y ?

nornagon:
@deepak1556 older versions also have this bug, but in general we don't backport Chromium fixes unless there's a specific request for a backport. This fix is useful for Slack and having it in 11.x is good enough for us. If there's a need for this fix in older versions we can open further backports.

deepak1556:
Makes sense, Thanks!

codebytere (approved):

codebytere:
(pending patch application fixup)

codebytere:
> `session.serviceWorkers getFromVersionID() should report the correct script url and scope` is a known/unrelated failure

PR #26063, opened 2020-10-20
fix: wasm code generation in the renderer

codebytere:
Backport of #25777

See that PR for details.

Notes: Fixed an issue where Wasm code generation erroneously showed as being disallowed by embedder when `contextIsolation` was enabled.

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -104,3 +104,4 @@ cherry-pick-8629cd7f8af3.patch
 avoid_use-after-free.patch
 don_t_create_providers_if_context_is_lost.patch
 fix_use_electron_generated_resources.patch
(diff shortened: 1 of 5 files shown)


jkleinsc (approved):

codebytere:
> `<webview> tag devtools-focused event should fire when webview.openDevTools() is called`

is a known failure

poiru:
Thanks for backpackporting this! @electron/wg-releases, this is blocking Electron 10 in Figma so it would be great to have a new 10.x release sooner than later 😊

quanglam2807:
Thank you so much, @codebytere 

I agree with @poiru. It's already Friday, is it possible to cut the release before the weekends?

PR #26065, opened 2020-10-20
feat: intercepted protocol access to orig handler

### LINKED ISSUE #15434: protocol.intercept{Any}Protocol handler ability to call original handler
It's possible to intercept schemes (including built-ins like `http:`, `https:`, and `file:`), but doing so requires you to handle all requests for that scheme, there's no way to fall-through to the original handler as far as I can tell. Neglecting to call `callback(...)` simply makes the request hang.

Using `protocol. interceptHttpProtocol('http', ...)` will lead to an infinite loop if you use `callback(request)`. This prevents using the intercepts to do handy things like set a header on all HTTP requests to a certain domain, or blackholing another (useful for preventing tracking, ads, etc):

```javascript
const url = require('url')

protocol.interceptHttpProtocol('https', (request, callback) => {
  const { host, pathname } = url.parse(request.url)

  if (host === 'github.com') {
    request.headers['X-Foo'] = 'bar'
  } else if (host === 'bitbucket.com') {
    callback(403)
    return
  }

  callback(request)  // This will cause an infinite loop
}, (error) => {
  if (error) console.error('Failed to register protocol')
})
```

Being able to selectively handle requests for an intercepted protocol would be very useful. It would provide the ability to use `protocol.interceptFileProtocol('http', ...)` and only intercept certain requests, such as those with the origin `http://localhost/`. This would allow you to simulate local files as being served over HTTP, getting around several of the limitations of the `file://` protocol. Currently there's no easy way to do this as you'll intercept all HTTP requests and lose the ability to talk to the network (at least over HTTP). Even if your handler could retrieve the content, it would have to give it a file path so that the callback can return the path to the file, or you'd need to use the `string`/`buffer`/`stream` variants and open any files yourself and provide the content that way. Neither is very clean. You can also bundle a local web server, but this is overkill most of the time.

I've found this particularly painful for trying to get WebAuthn working without a server in the loop, as Chromium only allows the API for `https` domains, or `http://localhost`, not `file://` (or any other protocol). An intercept seems like a perfect fit for this, except you can't selectively handle that small case.

Selective handling also opens the ability to make protocol handlers in the future act more like middleware which is composable and reusable. If interceptors could be applied like a stack and pass to the next interceptor, you could have libraries which handle things like setting certain headers.

I think this can be solved with a somewhat minor change to the API. There may be dragons in the implementation, I've glanced at the code which implements the `protocol` module, but I'm not too familiar with it.

My suggested solution would be to add a third argument to the handler function for the `protocol.intercept{Any}Protocol` functions called `next`, and rename the `callback` argument to `done`. `done` would have the same behavior as the current `callback` but the name change makes it more explicit that it will end processing of the request. The new argument, `next`, would fall through to the original handler, the one which would be restored by `protocol.uninterceptProtocol`. `next` would optionally take a request object, allowing you to modify it before passing it along, such as setting a header, or changing the URL allowing redirecting between files.

So, the intercept earlier which currently causes an infinite loop could be written as:

```javascript
const url = require('url')

protocol.interceptHttpProtocol('https', (request, done, next) => {
  const { host, pathname } = url.parse(request.url)

  if (host === 'github.com') {
    request.headers['X-Foo'] = 'bar'
  } else if (host === 'bitbucket.com') {
    done(403)
    return
  }

  next(request)
}, (error) => {
  if (error) console.error('Failed to register protocol')
})
```

You could also write a handler which pretended to be `http://localhost` by reading files from disk:

```javascript
const url = require('url')

protocol.interceptFileProtocol('http', (request, done, next) => {
  const { host, pathname } = url.parse(request.url)

  if (host === 'localhost') {
    done({ path: path.normalize(`${__dirname}/${pathname}`) })
  } else {
    next()  // Invokes the original 'http:' handler for 'request'
  }
}, (error) => {
  if (error) console.error('Failed to register protocol')
})
```

Or make `file://` check cloud storage if the file isn't found locally:

```javascript
const fs = require('fs')
const url = require('url')

protocol.interceptStreamProtocol('file', (request, done, next) => {
  const { host, pathname } = url.parse(request.url)

  if (host === '') {
    // Absolute path, check if pathname exists, if not
    // check cloud storage (Dropbox, OneDrive, etc)
    if (fs.existsSync(pathname)) {
      next()  // Invoke the original 'file:' handler for 'request'
    } else {
      checkCloud(function (contents) => {
        if (contents) {
          done(/* stream to contents */)
        } else {
          done({ statusCode: 404, data: 'Not Found' })
        }
      })
    }
  } else {
    // Don't allow file:// with a host
    done({ statusCode: 400, data: 'Host Not Allowed' })
  }
}, (error) => {
  if (error) console.error('Failed to register protocol')
})
```

Combined with intercepting a custom protocol:

```javascript
const url = require('url')

protocol.registerFileProtocol('atom', (request, callback) => {
  const { host, pathname } = url.parse(request.url)

  callback({ path: path.normalize(`${__dirname}/${host + pathname}`) })
}, (error) => {
  if (error) console.error('Failed to register protocol')
})
protocol.interceptStreamProtocol('atom', (request, done, next) => {
  const { host, pathname } = url.parse(request.url)

  if (pathname.endsWith('.mp4') || pathname === '/dev/video') {
    done(/* Open a stream to video */)
  } else {
    next()
  }
}, (error) => {
  if (error) console.error('Failed to register protocol')
})
```

dsanders11:
#### Description of Change
An attempt to implement #15434. It's more of a proof-of-concept than anything, but trying to get some movement on that issue and hoping a PR might move the ball further.

This code was a little cleaner before #25393, in particular shell/browser/net/proxying_url_loader_factory.h#L196 in this PR uses a bare `ProxyingURLLoaderFactory*` in a way that's probably a no-no after those changes.

The patch is at least functional, and I've used it to play with the proposed API from #15434 

cc @deepak1556 @zcbenz @nornagon 

Resolves #15434.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: added callback to continue intercepted protocol request as normal


zcbenz:
I'm good with this direction 👍 .

dsanders11:
@zcbenz, I've made a new commit which takes a slightly different approach: separating out the interception functionality into it's own `mojo::URLLoaderFactory` instead of doing both in `ProxyingURLLoaderFactory`. This makes for a cleaner implementation after the #25393 changes. It is however a bigger change with more moving pieces.

dsanders11:
This is ready for review now. There's still a minor TODO if the separate `mojo::URLLoaderFactory` implementation is used, which is cleaning up the `#include`s after separating out the two classes.

spec-main/api-protocol-spec.ts
```diff
@@ -510,6 +521,30 @@ describe('protocol module', () => {
...
+      expect(r.data).to.be.equal(text);
+
+      const port = await createServer();
+      interceptStringProtocol('http', (request, callback, next) => {
+        next();
+        expect(() => callback('foobar')).to.throw(Error);
+      });
+      r = await ajax(`http://127.0.0.1:${port}`);
+      expect(r.data).to.equal(text);
+    });
```
nornagon, line 547: Can we test calling `next()` asynchronously?

jkleinsc:

jkleinsc (approved):

dsanders11:
There was some discussion at the API WG meeting about changing the documented name from `next` to something else (unfortunately `continue` is a keyword and can't be used). Suggestions welcome.

dsanders11:
@nornagon, @MarshallOfSound, any chance this could be reviewed soon?

shell/browser/electron_browser_client.cc
```diff
@@ -1520,14 +1521,19 @@ bool ElectronBrowserClient::WillCreateURLLoaderFactory(
...
-      render_process_id, &next_id_, std::move(navigation_ui_data),
-      std::move(navigation_id), std::move(proxied_receiver),
+      web_request.get(), render_process_id, &next_id_,
+      std::move(navigation_ui_data), std::move(navigation_id),
+      proxy_factory_remote.InitWithNewPipeAndPassReceiver(),
       std::move(target_factory_remote), std::move(header_client_receiver),
       type);
+  new InterceptingURLLoaderFactory(protocol_registry->intercept_handlers(),
+                                   std::move(intercepted_receiver),
+                                   std::move(proxy_factory_remote));
```
nornagon, line 1536: It seems odd to me that we need _two_ URLLoaderFactories. Can/should this be consolidated into one factory? Is there any performance benefit from doing so?

shell/browser/net/intercepting_url_loader_factory.cc
```diff
@@ -0,0 +1,189 @@
...
+}
+
+void InterceptingURLLoaderFactory::OnInterceptorBindingError() {
+  if (interceptor_receivers_.empty())
+    proxying_factory_remote_.reset();
+
+  MaybeDeleteThis();
+}
+
+void InterceptingURLLoaderFactory::MaybeDeleteThis() {
```
nornagon, line 182: Is this only ever deleted when there's an error...? Or is this called in the normal lifecycle? If the latter, perhaps the methods could be renamed `Error` -> `Disconnect`?

shell/browser/electron_browser_client.cc
dsanders11: > It seems odd to me that we need two URLLoaderFactories.

Having only one factory doing both intercepting and proxying made the change more awkward.

Since the factory is a `mojo::Remote`, the only entrypoint is `CreateLoaderAndStart`. By the time the `next()` is invoked from JS, the only reference to the proxying factory is as a `mojo::PendingRemote`, so the only thing you can really do with it is call back into `CreateLoaderAndStart`, which would just intercept the request again, unless there's some way to tell it not to.

> Can/should this be consolidated into one factory?

The original commit did that, you can [review that implementation](https://github.com/electron/electron/pull/26065/commits/690b711a092e52bd598f36573b46f771e18505a5) for a comparison. Landing #25393 made it become awkward which is why I abandoned it in favor of the latest implementation. Specifically what got awkward was I'd separated out the default loading implementation so avoid the issue mentioned above, but there's no way to access that when using the factory as a `mojo::Remote`, so I ended up using a raw pointer which felt wrong there.

> Is there any performance benefit from doing so?

Nothing intended there, I'd expect performance to be trivially worse just because of the extra calls involved.

nornagon: ok, makes sense. I think it's cheap + rare enough to not worry too much about the extra cost.

dsanders11: Sounds good. Definitely a lesser of two evils kind of deal, but there is the benefit that having each factory only doing one thing makes it a bit easier to follow the code (IMO), and moves the proxying factory closer to `extensions::WebRequestProxyingURLLoaderFactory` so it should be a bit easier to keep them in sync now that the extra functionality is pulled out.

I'll work the other comments now that this is resolved.

nornagon:
@zcbenz I think this is `semver/minor`, right? It shouldn't break existing code.

zcbenz:
Yeah it should be `semver/minor`, thanks for noticing.

Juice10:
This is really cool, thanks for the effort to get this working @dsanders11!
Any chance we could get the last two things fixed @nornagon mentioned. This PR looks nearly good to go.

Let me know if there is anything I can do to help move this along!

spec-main/api-protocol-spec.ts
```diff
@@ -510,6 +521,39 @@ describe('protocol module', () => {
...
+      let r = await ajax('http://fake-host');
+      expect(r.data).to.be.equal(text);
+
+      const port = await createServer();
+      interceptStringProtocol('http', (request, callback, next) => {
+        next();
+        expect(() => callback('foobar')).to.throw(Error);
+      });
+      r = await ajax(`http://127.0.0.1:${port}`);
+      expect(r.data).to.equal(text);
```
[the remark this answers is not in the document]
Juice10: @nornagon looks like this @dsanders11 fixed this in the test. Can we mark this comment as resolved?

dsanders11:
@Juice10, appreciate the enthusiasm for the PR. However, I'd like to take another pass at this, with a bit of rearranging some stuff. There's various other issues I've noticed with the `protocol` module in Electron, and I've learned more about the Chromium code. I think there's a better implementation possible which will fix some of those other issues. I'll try to get to it soon, in the mean time I've marked this as a draft.

DomiStyle:
@dsanders11 Is the PR as is compatible with the current Electron version? I really need this functionality and was thinking about just compiling a patched Electron version with this PR?

PR #26082, opened 2020-10-21
revert: "fix: reply notifs sometimes destroyed too early"

MarshallOfSound:
Backport of #25247

Notes: no-notes


zcbenz:
@MarshallOfSound This revert seems to be empty.

PR #26086, opened 2020-10-21
build: rename the v8 context snapshot on arm64 macOS builds

MarshallOfSound:
As in the patch description, in order to generate universal builds on macOS both the arm64 and x64 snapshots need to be available, this is only possible if they have unique filenames.  This adds a patch which I intend to upstream to Chromium to make this the default behavior (and make the filename powered by GN instead of being hardcoded everywhere).

Upstream attempt: https://chromium-review.googlesource.com/c/chromium/src/+/2490980

Notes: no-notes


jkleinsc (approved):

codebytere (approved):

zcbenz (changes requested):
arm64 build is failing with a strange error:

```
chruby: unknown Ruby: ruby-2.7
```

MarshallOfSound:
Looks like this is gonna land upstream, updated the patch to exactly match upstream

PR #9113, opened 2017-04-05
browser: make destruction of webContents async

deepak1556:
Fixes https://github.com/electron/electron/issues/8930
Fixes https://github.com/electron/electron/issues/9230

Original discussion https://groups.google.com/a/chromium.org/d/msg/embedder-dev/XTv6SpambTI/V51xPz1hFgAJ

PR #26090, opened 2020-10-21
chore: synchronously destroy WebContents on event prevented

nornagon:
#### Description of Change
Trying this out to see if CI breaks. From my reading of the Chromium code it should be ok.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -825,8 +825,7 @@ void WebContents::AddNewContents(
            initial_rect.x(), initial_rect.y(), initial_rect.width(),
            initial_rect.height(), tracker->url, tracker->frame_name,
            tracker->referrer, tracker->raw_features, tracker->body)) {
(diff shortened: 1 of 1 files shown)


zcbenz:
This was introduced by https://github.com/electron/electron/pull/9113, since the tests passed it would be fine. /cc @deepak1556

deepak1556 (approved):
I originally wanted to keep all top level webContent destruction async and this was added along side. Sounds good to have this sync when the webContents is not exposed to the user.

nornagon:
I think it's important for this to be sync because it prevents `window.open` from returning a window that's then destroyed (instead it'll return `null`).

zcbenz (approved):

mlaurencin:
@nornagon [This issue](https://github.com/electron/electron/issues/26133) was fixed when I applied this commit locally to 10, so I think this would be good to backport to 9, 10, and 11 as well.

PR #26094, opened 2020-10-22
build: drop use of --typeRoots for default_app

dsanders11:
#### Description of Change
Full disclaimer, I don't know much about GN and the *.gni files.

I bumped into this randomly while debugging a build issue (problem in my source tree).

I believe it's a correct change because:
* `node_modules/@types` is default behavior so it will still be used
* `typings` should be covered by the `include` in `tsconfig.default_app.json`
* `$target_gen_dir/tsc/electron/typings` doesn't exist - I believe this should be `$target_gen_dir/tsc/typings` if it were to have any effect, since I don't see `$target_gen_dir/tsc/electron/typings` existing.

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -184,7 +184,6 @@ target_gen_default_app_js = "$target_gen_dir/js/default_app"
 
 typescript_build("default_app_js") {
   deps = [ ":build_electron_definitions" ]
-  type_root = rebase_path("$target_gen_dir/tsc/electron/typings")
 
   sources = filenames.default_app_ts_sources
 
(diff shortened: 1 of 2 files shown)


ckerr (approved):
This looks correct to me but I'd like another opinion from someone who's done more with this code, e.g. @MarshallOfSound?

zcbenz (approved):

codebytere (approved):

MarshallOfSound (approved):
Yeah this logic predates a substantial amount of refactoring we did around how our TS compilation works.  `default_app` is actually the odd one out now

Issue #5577, opened 2016-05-18
Allow dialogs to be cancelled


50Wliu:
- Electron version: 1.1.0
- Operating system: N/A

Currently there isn't any way to remove dialog boxes (such as ones created through `dialog.showMessageBox`).  This can be useful in, for example, "Keep Waiting" dialogs.  After the window becomes responsive again it would be great if we could programmatically close the dialog.

jakubzitny:
This would also be useful e.g. when user opens a dialog and returns to the app and changes stuff (context) which makes the dialog unneccessary. From the app it'd make sense to cancel the dialog.

sante85:
up

sante85:
Please, could you tell me which component to modify to solve the problem?

Best Regards

codebytere:
Closing due to abandonment.

sagiegurari:
What kind of resolution is this? This is one of the annoying behaviours in atom and I hoped for a fix.

codebytere:
We're weeding out abandoned issues right now; if you'd like this issue to remain open & think it's an important fix that's totally fine and i'll reopen it :)

sagiegurari:
Yes please. First time boot is slow and this dialog hurts the useability

codebytere:
I'll see if i can take a look at this soon and work on a fix!

Ben3eeE:
:+1: on getting this fixed. Moving to an async dialog that we can programmatically close once Atom is responsive again would be great.

50Wliu:
@codebytere have you maybe looked into using [probot/stale](https://github.com/probot/stale)?  It's definitely helped the Atom organization clean up a lot of old issues, and you can configure it to give a warning message in advance so that it's not as surprising when it does close the issue.

thisisgabe:
Any update on this? I'm working on an "auto-logoff" feature that changes windows once the user has been idle for a certain amount of time. I display a warning message box 30 seconds before I switch back to the login window but I would like to close that message box once I go back to the login page (if the user doesn't return from idle).

bpasham:
this is a super important feature. I am looking for a "timeout" option on dialogs. The dialog should automatically close if no response for 30s or so. I build apps for publicly visible unattended scenarios, and automatic dialog closing is extremely important.... thanks

renambot-uic:
Any hope to be fixed in v2 ?

tcanusso:
[...]

PR #26102, opened 2020-10-22
feat: add signal option to dialog.showMessageBox

zcbenz:
#### Description of Change

This PR adds a `signal` option to close message box, once this API is approved, I'll implement same API to close file dialogs. Refs https://github.com/electron/electron/issues/5577.

```js
const controller = new AbortController();
dialog.showMessageBox(win, {
  signal: controller.signal,
  message: 'message',
})

controller.abort()
```

##### Duplicate code

There are some duplicate code when checking dialog IDs, which I would like to refactor after adding the API for closing file dialogs. Optimizing them in this PR is too early.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Add `signal` option to `dialog.showMessageBox`.

### DIFF
--- a/shell/browser/ui/message_box_gtk.cc
+++ b/shell/browser/ui/message_box_gtk.cc
@@ -42,7 +42,7 @@ MessageBoxSettings::~MessageBoxSettings() = default;
 namespace {
 
 // <ID, messageBox> map
-std::map<std::string, GtkWidget*> g_dialogs;
+std::map<int, GtkWidget*> g_dialogs;
 
 class GtkMessageBox : public NativeWindowObserver {
  public:
@@ -193,7 +193,7 @@ class GtkMessageBox : public NativeWindowObserver {
   electron::UnresponsiveSuppressor unresponsive_suppressor_;
 
   // The id of the dialog.
-  absl::optional<std::string> id_;
+  absl::optional<int> id_;
 
   // The id to return when the dialog is closed without pressing buttons.
   int cancel_id_ = 0;
@@ -240,7 +240,7 @@ void ShowMessageBox(const MessageBoxSettings& settings,
   (new GtkMessageBox(settings))->RunAsynchronous(std::move(callback));
 }
 
-bool CloseMessageBox(const std::string& id, std::string* error) {
+bool CloseMessageBox(int id, std::string* error) {
   DCHECK(error);
   auto it = g_dialogs.find(id);
   if (it == g_dialogs.end()) {
--- a/shell/browser/ui/message_box_gtk.cc
+++ b/shell/browser/ui/message_box_gtk.cc
@@ -7,6 +7,7 @@
 #include <map>
 
 #include "base/callback.h"
+#include "base/containers/contains.h"
 #include "base/strings/string_util.h"
 #include "base/strings/utf_string_conversions.h"
 #include "shell/browser/browser.h"
--- a/shell/browser/ui/message_box_gtk.cc
+++ b/shell/browser/ui/message_box_gtk.cc
@@ -213,9 +213,9 @@ void GtkMessageBox::OnResponseDialog(GtkWidget* widget, int response) {
   gtk_widget_hide(dialog_);
 
   if (response < 0)
-    std::move(callback_).Run("", cancel_id_, checkbox_checked_);
+    std::move(callback_).Run(cancel_id_, checkbox_checked_);
   else
-    std::move(callback_).Run("", response, checkbox_checked_);
+    std::move(callback_).Run(response, checkbox_checked_);
   delete this;
 }
 
@@ -231,24 +231,18 @@ int ShowMessageBoxSync(const MessageBoxSettings& settings) {
 
 void ShowMessageBox(const MessageBoxSettings& settings,
                     MessageBoxCallback callback) {
-  if (settings.id) {
-    if (base::Contains(g_dialogs, *settings.id)) {
-      std::move(callback).Run("Duplicate ID found", 0, false);
-      return;
-    }
-  }
+  if (settings.id && base::Contains(g_dialogs, *settings.id))
+    CloseMessageBox(*settings.id);
   (new GtkMessageBox(settings))->RunAsynchronous(std::move(callback));
 }
 
-bool CloseMessageBox(int id, std::string* error) {
-  DCHECK(error);
+void CloseMessageBox(int id) {
   auto it = g_dialogs.find(id);
   if (it == g_dialogs.end()) {
-    *error = "ID not found";
-    return false;
+    LOG(ERROR) << "CloseMessageBox called with unexist ID";
+    return;
   }
   gtk_window_close(GTK_WINDOW(it->second));
-  return true;
 }
 
 void ShowErrorBox(const std::u16string& title, const std::u16string& content) {
--- a/shell/browser/ui/message_box_gtk.cc
+++ b/shell/browser/ui/message_box_gtk.cc
@@ -2,8 +2,9 @@
 // Use of this source code is governed by the MIT license that can be
 // found in the LICENSE file.
 
-#include "shell/browser/ui/gtk_util.h"
 #include "shell/browser/ui/message_box.h"
+
+#include <map>
 
 #include "base/callback.h"
 #include "base/strings/string_util.h"
@@ -11,6 +12,7 @@
 #include "shell/browser/browser.h"
 #include "shell/browser/native_window_observer.h"
 #include "shell/browser/native_window_views.h"
+#include "shell/browser/ui/gtk_util.h"
 #include "shell/browser/unresponsive_suppressor.h"
 #include "ui/base/glib/glib_signal.h"
 #include "ui/gfx/image/image_skia.h"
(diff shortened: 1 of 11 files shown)


shell/browser/ui/message_box_gtk.cc
```diff
@@ -38,10 +40,14 @@ MessageBoxSettings::~MessageBoxSettings() = default;
 
 namespace {
 
+// <ID, messageBox> map
+std::map<std::string, GtkWidget*> g_dialogs;
```
miniak, line 44: you can use `std::unordered_map`

nornagon:
I don't love this ID system. Perhaps we could use something like AbortSignal? https://w3ctag.github.io/design-principles/#aborting

shell/browser/ui/message_box_gtk.cc
zcbenz: Chromium suggests avoiding `std::unordered_map` unless there is a very good reason:
https://chromium.googlesource.com/chromium/src/base/+/master/containers/#map-and-set-selection

zcbenz:
> I don't love this ID system. Perhaps we could use something like AbortSignal? https://w3ctag.github.io/design-principles/#aborting

I did not know it exists, it is a better design than a custom ID system. So I think the new API would look like this?

```js
const controller = new AbortController();
dialog.showMessageBox(win, {
  signal: controller.signal,
  message: 'message',
})

controller.abort();
```

There are 2 downsides though:
1. `AbortController` is a currently DOM only thing. Node is [adding its own `AbortController`](https://github.com/nodejs/node/pull/33527) since v15, but it is currently marked experimental and not supported by any promise API, using it seems a bit risky.
2. It can only be used to cancel the task, and does not provide a way to specify options. We might want to extend the API in future.

So I think maybe we can invent our own `AbortController` style API:

```js
const controller = new dialog.MessageBoxAbortController()

dialog.showMessageBox(win, {
  signal: controller.signal,
  message: 'message',
})

controller.abort();
// future extensions.
controller.close({button: 1})
```

nornagon:
I'm not convinced we need the option to "pretend" that a particular button was pressed. User code can handle that plumbing by itself if needed, e.g.

```js
const controller = new AbortController()
let abortedSignal = null

dialog.showMessageBox(win, {
  signal: controller.signal,
  message: 'message'
}).then(status => {
  if (status == null)
    status = abortedSignal
})

abortedSignal = 1
controller.abort();
```

as for Node adding its own, it seems like they are planning to match the DOM implementation (similar to their implementation of `URL`). The DOM version is [standardized](https://dom.spec.whatwg.org/#interface-abortcontroller), so it seems pretty safe to build our own shim that matches that and switch to Node's once it's available.

zcbenz:
I'm going to wait for the next major upgrade of Node in Electron, adding typescript declarations for our own `AbortSymbol` is going to have problems like https://github.com/node-fetch/node-fetch/issues/741.

zcbenz:
I have updated this PR to use `AbortSignal` as option.

MarshallOfSound (approved):

jkleinsc (approved):

shell/browser/ui/message_box_win.cc
```diff
@@ -198,33 +260,67 @@ DialogResult ShowTaskDialogUTF8(const MessageBoxSettings& settings) {
...
+              g_dialogs.erase(*id);
+            std::move(callback).Run(result.first, result.second);
+          },
+          std::move(callback), settings.id));
+}
+
+void CloseMessageBox(int id) {
+  auto it = g_dialogs.find(id);
+  if (it == g_dialogs.end()) {
+    LOG(ERROR) << "CloseMessageBox called with unexist ID";
```
nornagon, line 302:
```suggestion
    LOG(ERROR) << "CloseMessageBox called with nonexistent ID";
```

nornagon (commented):
overall lgtm, but a few suggestions

sogan66 (approved):
Submit feedback approve merging these changes

shell/browser/ui/message_box_win.cc
```diff
@@ -30,6 +34,32 @@ MessageBoxSettings::~MessageBoxSettings() = default;
...
+// will be passed between threads and we need to ensure the memory of HWND is
+// not changed while g_dialogs is modified.
+std::map<int, std::unique_ptr<HWND>> g_dialogs;
+
+// Speical HWND used by the g_dialogs map.
+//
+// - ID is used but window has not been created yet.
+HWND kHwndReserve = reinterpret_cast<HWND>(-1);
+// - Notification to cancel message box.
+HWND kHwndCancel = reinterpret_cast<HWND>(-2);
```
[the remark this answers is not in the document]
zcbenz: It seems that `reinterpret_cast` can not be used for `constexpr` in C++, I'm using `const` instead.

```
../../electron/shell/browser/ui/message_box_win.cc(49,16): error: constexpr variable 'kHwndReserve' must be initialized by a constant expression
```

shell/browser/ui/message_box_mac.mm
```diff
@@ -128,6 +133,12 @@ void ShowMessageBox(const MessageBoxSettings& settings,
     int ret = [[alert autorelease] runModal];
     std::move(callback).Run(ret, alert.suppressionButton.state == NSOnState);
   } else {
+    if (settings.id) {
+      if (base::Contains(g_dialogs, *settings.id))
```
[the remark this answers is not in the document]
zcbenz: It may happen if someone calls `process._linkedBinding('electron_browser_dialog').showMessageBox` directly, or when `nextId` overflows while some old dialogs are still open.

I don't think either would ever happen, but it is an edge case anyway and a simple check can protect us from leaking things.

shell/browser/ui/message_box_gtk.cc
```diff
@@ -38,10 +42,14 @@ MessageBoxSettings::~MessageBoxSettings() = default;
 
 namespace {
 
+// <ID, messageBox> map
+base::NoDestructor<std::map<int, GtkWidget*>> g_dialogs;
```
nornagon, line 46: All the usages in Chromium of `base::NoDestructor` wrap it inside a function like:
```
const std::string& GetLineSeparator() {
 // Forwards to std::string(size_t, char, const Allocator&) constructor.
  static const base::NoDestructor<std::string> s(5, '-');
  return *s;
}
```

Is it OK to have it at the global level like this...?

zcbenz: Just found the doc says 

```
// Note that since the destructor is never run, this *will* leak memory if used
// as a stack or member variable. Furthermore, a NoDestructor<T> should never
// have global scope as that may require a static initializer.
```

so yeah they must be wrapped in a function.

spec-main/api-dialog-spec.ts
```diff
@@ -121,6 +121,62 @@ describe('dialog module', () => {
...
+      const result = await p;
+      expect(result.response).to.equal(0);
+    });
+
+    it('closes message box after a while', async () => {
+      const controller = new AbortController();
+      const signal = controller.signal;
+      const w = new BrowserWindow();
+      const p = dialog.showMessageBox(w, { signal, message: 'i am message' });
+      await delay(500);
```
nornagon, line 142: Why 500ms here? Would 0ms work equally well? If not, is there some event we can listen for?

`delay(500)` smells very much like flake to me.

spec-main/api-dialog-spec.ts
```diff
@@ -121,6 +121,62 @@ describe('dialog module', () => {
...
+      const controller = new AbortController();
+      const signal = controller.signal;
+      const w = new BrowserWindow();
+      const p = dialog.showMessageBox(w, {
+        signal,
+        message: 'i am message',
+        buttons: ['OK', 'Cancel'],
+        cancelId: 1
+      });
+      await delay(500);
```
nornagon, line 173: as above

spec-main/api-dialog-spec.ts
zcbenz: On Windows it takes some time before we can know the dialog's HWND (the dialog runs in a new thread and the dialog HWND is created asynchronously), the delay here is to make the test run the code path that closes the window after HWND is available. 

Sometimes even 500ms is not enough for the HWND to be available, but this is fine as no matter which code path it runs the showMessageBox should be cancelled successfully.

nornagon: Hm, is there any other way we could detect this event? e.g. a blur event?

zcbenz: Showing a a child dialog does not remove the focus of parent window, so browser window's `blur` event will not emit. The only way I can think of is to load a web page and setup a `window.onblur` listener in it to send the event, which needs to careful to be not flaky.

Waiting for fixed time is usually bad because it makes test flaky, but this test is not flaky no matter what the timeout is, it only needs to confirm dialog can be closed after fully opened, and I think `delay(500)` does the work well.

nornagon: Yeah, it won't necessarily make the test fail, but it does mean the test will test _different things_, undetectably, depending on how fast the underlying OS is. I'd prefer that our tests test the same thing every time if possible.

I won't block the PR on this concern though.

nornagon (approved):

zcbenz:
@MarshallOfSound The `AbortSignal` was only added since Node 15, so this can not be back ported to Electron 14.

PR #26104, opened 2020-10-22
fix: window.open not accepting size values with "px" at the end

### LINKED ISSUE #26077: window.open does not accept size values with "px" at the end
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 10.1.4
* **Operating System:**
  * Windows
  * 9.3.2

### Expected Behavior
The window should open with the correct size

### Actual Behavior
The size is ignored

### To Reproduce
Use nativeWindowOpen and do: window.open("", "", "width=300px")

### Additional Information
This is working in chrome and firefox and was working in electron 9. If the event is intercepted via webContents.on("new-window") then the value is shown as NAN,

Code-Blooded-Human:
#### Description of Change
Close #26077 , correctly parses "px" at end of size string
#### Checklist
- [x] `npm test` passes

#### Release Notes
notes: Fix `window.open` not accepting size values with "px" at the end.

### DIFF
--- a/lib/common/parse-features-string.ts
+++ b/lib/common/parse-features-string.ts
@@ -38,7 +38,7 @@ const keysOfTypeNumber = ['top', 'left', ...Object.keys(keysOfTypeNumberCompileT
 type CoercedValue = string | number | boolean;
 function coerce (key: string, value: string): CoercedValue {
   if (keysOfTypeNumber.includes(key)) {
-    return Number(value);
+    return parseInt(value);
   }
 
   switch (value) {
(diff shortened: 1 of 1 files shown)


codebytere (commented):
@Code-Blooded-Human please correct your commit to be semantic (the commit _itself_ must start with `fix`) - this also needs proper release notes and a test as well. On the original issue, you can see that if the event is intercepted via `webContents.on("new-window")` then the value is shown as NaN - so a good test here would verify that does _not_ occur after this fix.

lib/common/parse-features-string.ts
```diff
@@ -38,7 +38,7 @@ const keysOfTypeNumber = ['top', 'left', ...Object.keys(keysOfTypeNumberCompileT
 type CoercedValue = string | number | boolean;
 function coerce (key: string, value: string): CoercedValue {
   if (keysOfTypeNumber.includes(key)) {
-    return Number(value);
+    return parseInt(value);
```
MarshallOfSound, line 41:
```suggestion
    return parseInt(value, 10);
```

The radix does not default to 10

zcbenz (approved):

pushkin-:
will this be backported to 10?

PR #26113, opened 2020-10-22
chore: use auto to avoid repeating type

dsanders11:
#### Description of Change
Another `clang-tidy` suggestion. Use `auto` when the type is already specified elsewhere in the statement.

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/electron_api_debugger.cc
+++ b/shell/browser/api/electron_api_debugger.cc
@@ -54,8 +54,7 @@ void Debugger::DispatchProtocolMessage(DevToolsAgentHost* agent_host,
                                        base::JSON_REPLACE_INVALID_CHARACTERS);
   if (!parsed_message || !parsed_message->is_dict())
     return;
-  base::DictionaryValue* dict =
-      static_cast<base::DictionaryValue*>(parsed_message.get());
+  auto* dict = static_cast<base::DictionaryValue*>(parsed_message.get());
   int id;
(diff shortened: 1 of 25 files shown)


ckerr (approved):

ckerr:
@dsanders11  We usually backport changes like this to older supported branches to reduce code shear between branches -- but this touches so many unrelated pieces, I'm curious to see how trop does with it :smile:

dsanders11:
I think the answer to that is... not well, judging by the failed checks. 😄

nornagon (changes requested):
I'm 👎  on this. Sometimes `auto` is better, but sometimes I specifically use the type name as documentation. I don't think being consistent across our codebase helps much here.

ckerr:
> I'm  on this. Sometimes `auto` is better, but sometimes I specifically use the type name as documentation. I don't think being consistent across our codebase helps much here.

@nornagon I think you might be misunderstanding PR? This is removing _redundant_ information, e.g.

```diff
- SomeTypeName* foo = static_cast<SomeTypeName*>(bar);
+ auto* foo = static_cast<SomeTypeName*>(bar);
```

s.t. only the redundancy is removed, not the explicit typing. I don't think you'd be losing any self-documenting properties with this use of `auto`.

nornagon (approved):
... you're right, sorry. 👍

MarshallOfSound (approved):

PR #26115, opened 2020-10-22
fix: ensure object is not used after move

dsanders11:
#### Description of Change
From `clang-tidy`:

```
electron/shell/browser/net/electron_url_loader_factory.cc:554:3: warning: 'write_data' used after it was moved [bugprone-use-after-move]
  write_data->producer->Write(
  ^
electron/shell/browser/net/electron_url_loader_factory.cc:558:7: note: move occurred here
      base::BindOnce(OnWrite, std::move(write_data)));
      ^
electron/shell/browser/net/electron_url_loader_factory.cc:554:3: note: the use and move are unsequenced, i.e. there is no guarantee about the order in which they are evaluated
  write_data->producer->Write(
  ^
```

I think the change is a bit clunky, but it does fix the potential order of evaluation issue. Suggestions for more elegant alternatives welcome.

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/net/electron_url_loader_factory.cc
+++ b/shell/browser/net/electron_url_loader_factory.cc
@@ -549,9 +549,10 @@ void ElectronURLLoaderFactory::SendContents(
   write_data->data = std::move(data);
   write_data->producer =
       std::make_unique<mojo::DataPipeProducer>(std::move(producer));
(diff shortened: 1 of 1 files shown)


deepak1556 (approved):
LGTM

MarshallOfSound (approved):

PR #26119, opened 2020-10-22
fix: re-enable the spellchecker when new language list set

MarshallOfSound:
Chromium recently added prefs logic to disable the spellchecker if the list of languages is empty, but the logic to re-enable if the languages are provided again lives in another part of Chromium.  This change makes it so our API re-enables the spellchecker correctly when required.

Refs: https://chromium-review.googlesource.com/c/chromium/src/+/2390300

Notes: Fixed issue where setting the spellchecker languages to an empty array would permanently disable the spellchecker till the end of time

### DIFF
--- a/spec-main/spec-helpers.ts
+++ b/spec-main/spec-helpers.ts
@@ -2,9 +2,19 @@ import * as childProcess from 'child_process';
 import * as path from 'path';
 import * as http from 'http';
 import * as v8 from 'v8';
+import { SuiteFunction, TestFunction } from 'mocha';
 
-export const ifit = (condition: boolean) => (condition ? it : it.skip);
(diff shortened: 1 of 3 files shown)


spec-main/spec-helpers.ts
```diff
@@ -2,9 +2,19 @@ import * as childProcess from 'child_process';
 import * as path from 'path';
 import * as http from 'http';
 import * as v8 from 'v8';
+import { SuiteFunction, TestFunction } from 'mocha';
```
MarshallOfSound, line 5: This stuff I just added because it was annoying me, you can now do `ifdescribe().only`

codebytere (approved):

PR #26134, opened 2020-10-23
fix: Don't sort the headers of ClientRequest

LuoJinghua:
#### Description of Change
This pull request make net.request to keep the order of headers for the request.

#### Checklist

- [ ] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: fix: Don't sort the headers of ClientRequest

PR #26135, opened 2020-10-23
feat: allow setting the Origin header and Sec-Fetch-* headers in net.request()

LuoJinghua:
#### Description of Change
This pull request is based on #26134. It's possible to reproduce a request send by Chrome Browser with #26134 and this pull request. 

Is ok to backport this PR to stable branches, such as 10-x-y?

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Allowed setting the 'Origin' header and Sec-Fetch-* headers in net.request().

### DIFF
--- a/docs/api/client-request.md
+++ b/docs/api/client-request.md
@@ -47,6 +47,7 @@ following properties:
     be aborted. When mode is `manual` the redirection will be cancelled unless
     [`request.followRedirect`](#requestfollowredirect) is invoked synchronously
     during the [`redirect`](#event-redirect) event.  Defaults to `follow`.
+  * `origin` String (optional) - The origin URL of the request.
 
 `options` properties such as `protocol`, `host`, `hostname`, `port` and `path`
 strictly follow the Node.js model as described in the
--- a/lib/browser/api/net.ts
+++ b/lib/browser/api/net.ts
@@ -256,7 +256,8 @@ function parseOptions (optionsIn: ClientRequestConstructorOptions | string): Nod
     headers: {},
     body: null as any,
     useSessionCookies: options.useSessionCookies,
-    credentials: options.credentials
+    credentials: options.credentials,
+    origin: options.origin
   };
   const headers: Record<string, string | string[]> = options.headers || {};
   for (const [name, value] of Object.entries(headers)) {
@@ -414,6 +415,7 @@ export class ClientRequest extends Writable implements Electron.ClientRequest {
       return ret;
     };
     this._urlLoaderOptions.referrer = this.getHeader('referer') || '';
+    this._urlLoaderOptions.origin = this._urlLoaderOptions.origin || this.getHeader('origin') || '';
     const opts = { ...this._urlLoaderOptions, extraHeaders: stringifyValues(this._urlLoaderOptions.headers) };
     this._urlLoader = createURLLoader(opts);
     this._urlLoader.on('response-started', (event, finalUrl, responseHead) => {
--- a/lib/browser/api/net.ts
+++ b/lib/browser/api/net.ts
@@ -197,7 +197,7 @@ class ChunkedBodyStream extends Writable {
 
 type RedirectPolicy = 'manual' | 'follow' | 'error';
 
-function parseOptions (optionsIn: ClientRequestConstructorOptions | string): NodeJS.CreateURLLoaderOptions & { redirectPolicy: RedirectPolicy, extraHeaders: Record<string, string> } {
+function parseOptions (optionsIn: ClientRequestConstructorOptions | string): NodeJS.CreateURLLoaderOptions & { redirectPolicy: RedirectPolicy, headers: Record<string, { name: string, value: string | string[] }> } {
   const options: any = typeof optionsIn === 'string' ? url.parse(optionsIn) : { ...optionsIn };
 
   let urlStr: string = options.url;
@@ -249,22 +249,25 @@ function parseOptions (optionsIn: ClientRequestConstructorOptions | string): Nod
     throw new TypeError('headers must be an object');
   }
 
-  const urlLoaderOptions: NodeJS.CreateURLLoaderOptions & { redirectPolicy: RedirectPolicy, extraHeaders: Record<string, string | string[]> } = {
+  const urlLoaderOptions: NodeJS.CreateURLLoaderOptions & { redirectPolicy: RedirectPolicy, headers: Record<string, { name: string, value: string | string[] }> } = {
     method: (options.method || 'GET').toUpperCase(),
     url: urlStr,
     redirectPolicy,
-    extraHeaders: options.headers || {},
+    headers: {},
     body: null as any,
     useSessionCookies: options.useSessionCookies,
     credentials: options.credentials
   };
-  for (const [name, value] of Object.entries(urlLoaderOptions.extraHeaders!)) {
+  const headers: Record<string, string | string[]> = options.headers || {};
+  for (const [name, value] of Object.entries(headers)) {
     if (!isValidHeaderName(name)) {
       throw new Error(`Invalid header name: '${name}'`);
     }
     if (!isValidHeaderValue(value.toString())) {
       throw new Error(`Invalid value for header '${name}': '${value}'`);
     }
+    const key = name.toLowerCase();
+    urlLoaderOptions.headers[key] = { name, value };
   }
   if (options.session) {
     // Weak check, but it should be enough to catch 99% of accidental misuses.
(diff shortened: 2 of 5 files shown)


shell/browser/api/electron_api_url_loader.cc
```diff
@@ -381,6 +381,11 @@ gin::Handle<SimpleURLLoaderWrapper> SimpleURLLoaderWrapper::Create(
   if (!origin.empty()) {
     request->request_initiator = url::Origin::Create(GURL(origin));
   }
+  bool has_user_activation;
+  if (opts.Get("hasUserActivation", &has_user_activation)) {
+    request->trusted_params = network::ResourceRequest::TrustedParams();
+    request->trusted_params->has_user_activation = has_user_activation;
```
nornagon, line 387: It looks like we'll start sending `Sec-Fetch-User: ?0` with all requests by default as of this change. Is that desirable?

spec-main/api-net-spec.ts
```diff
@@ -847,6 +847,36 @@ describe('net module', () => {
...
+        response.end();
+      });
+      const urlRequest = net.request({
+        url: serverUrl
+      });
+      await collectStreamBody(await getResponse(urlRequest));
+    });
+
+    ['navigate', 'cors', 'no-cors', 'same-origin'].forEach((mode) => {
+      it('should set sec-fetch-mode if requested', async () => {
```
nornagon, line 864:
```suggestion
      it(`should set sec-fetch-mode to ${mode} if requested`, async () => {
```

spec-main/api-net-spec.ts
```diff
@@ -877,6 +877,42 @@ describe('net module', () => {
...
+      await collectStreamBody(await getResponse(urlRequest));
+    });
+
+    [
+      'empty', 'audio', 'audioworklet', 'document', 'embed', 'font',
+      'frame', 'iframe', 'image', 'manifest', 'object', 'paintworklet',
+      'report', 'script', 'serviceworker', 'style', 'track', 'video',
+      'worker', 'xslt'
+    ].forEach((dest) => {
+      it('should set sec-fetch-dest if requested', async () => {
```
nornagon, line 900:
```suggestion
      it(`should set sec-fetch-dest to ${dest} if requested`, async () => {
```

nornagon (changes requested):
mostly lgtm! just want clarity on the (possibly) new `sec-fetch-user` behavior.

docs/api/client-request.md
```diff
@@ -47,6 +47,7 @@ following properties:
     be aborted. When mode is `manual` the redirection will be cancelled unless
     [`request.followRedirect`](#requestfollowredirect) is invoked synchronously
     during the [`redirect`](#event-redirect) event.  Defaults to `follow`.
+  * `origin` String (optional) - The origin URL of the request.
```
nornagon, line 50: Should we perhaps only allow setting this through the headers?

LuoJinghua: We need to set the origin url but not the origin header for sometime.

shell/browser/api/electron_api_url_loader.cc
LuoJinghua: I don’t think so. But I will add a test to ensure this.

LuoJinghua: It seems the chromimum network stack will remove the Sec-Fetch-User header if has_user_activation is false.

spec-main/api-net-spec.ts
LuoJinghua: Done.

spec-main/api-net-spec.ts
LuoJinghua: Done.

shell/browser/api/electron_api_url_loader.cc
LuoJinghua: I added a test to make sure this doesn't happen.

zcbenz (approved):

zcbenz:
> Is ok to backport this PR to stable branches, such as 10-x-y?

Once this API is approved by wg-api, we can ask wg-releases for approval of backport.

spec-main/api-net-spec.ts
```diff
@@ -756,6 +777,154 @@ describe('net module', () => {
...
+        expect(request.headers['sec-fetch-user']).to.not.equal('?0');
+        response.statusCode = 200;
+        response.statusMessage = 'OK';
+        response.end();
+      });
+      const urlRequest = net.request({
+        url: serverUrl
+      });
+      await collectStreamBody(await getResponse(urlRequest));
+    });
```
nornagon, line 847: This can be a single test:

```suggestion
    it('should not send sec-fetch-user header by default', async () => {
      const serverUrl = await respondOnce.toSingleURL((request, response) => {
        expect(request.headers).not.to.have.property('sec-fetch-user');
        response.statusCode = 200;
        response.statusMessage = 'OK';
        response.end();
      });
      const urlRequest = net.request({
        url: serverUrl
      });
      await collectStreamBody(await getResponse(urlRequest));
    });
```

nornagon (approved):
Generally speaking we don't backport new features to stable branches unless there's a specific need. Backporting a new feature to a stable branch requires a minor version bump, and it's up to @electron/wg-releases to decide on a case-by-case basis whether to accept a feature backport.

In my opinion (as a non-member of the Releases WG), this feature presents a non-zero but quite small stability risk. Code paths are changed by this PR, but in such a way that it seems quite unlikely to cause any observable behavior change unless the caller is using the added features.

spec-main/api-net-spec.ts
LuoJinghua: Done.

LuoJinghua:
Hi, the cors request always fail with net error -2 and cors error kPreflightWildcardOriginNotAllowed. Do you have any idea what happens for this case?

Response of preflight request:
``` 
HTTP/1.1 200 
date	Wed, 28 Oct 2020 00:49:47 GMT
access-control-allow-origin	*
access-control-allow-methods	GET, POST, OPTIONS
access-control-expose-headers	
access-control-max-age	3000
access-control-allow-headers	content-type,x-csrf-token
Content-Security-Policy	frame-ancestors 'self'
strict-transport-security	max-age=31536000; includeSubDomains
Transfer-Encoding	chunked
Connection	keep-alive
```

LuoJinghua:
> Hi, the cors request always fail with net error -2 and cors error kPreflightWildcardOriginNotAllowed. Do you have any idea what happens for this case?
> 
> Response of preflight request:
> 
> ```
> HTTP/1.1 200 
> date	Wed, 28 Oct 2020 00:49:47 GMT
> access-control-allow-origin	*
> access-control-allow-methods	GET, POST, OPTIONS
> access-control-expose-headers	
> access-control-max-age	3000
> access-control-allow-headers	content-type,x-csrf-token
> Content-Security-Policy	frame-ancestors 'self'
> strict-transport-security	max-age=31536000; includeSubDomains
> Transfer-Encoding	chunked
> Connection	keep-alive
> ```

Ok, figured out this finally. We need to backport 0e7d59dd7(#25284) too to fix this issue for stable branches.

lib/browser/api/net.ts
```diff
@@ -409,6 +410,10 @@ export class ClientRequest extends Writable implements Electron.ClientRequest {
       return ret;
     };
     this._urlLoaderOptions.referrer = this._urlLoaderOptions.extraHeaders.referer || '';
+    this._urlLoaderOptions.origin = this._urlLoaderOptions.origin || this._urlLoaderOptions.extraHeaders.origin || '';
```
MarshallOfSound, line 413: Headers are case insensitive when read by the networking stack but this assumes that `origin` is lowercase here.  To be accurate this should somehow pull `origin` in a case insensitive way right?

jkleinsc:

lib/browser/api/net.ts
LuoJinghua: The header name are stored in lowercase internally. This is another issue we may fix in the future.

```
  setHeader (name: string, value: string) {
    ...
    const key = name.toLowerCase();
    this._urlLoaderOptions.extraHeaders[key] = value;
  }
```

LuoJinghua: I just go ahead and fix this issue too.

lib/browser/api/net.ts
```diff
@@ -249,22 +249,26 @@ function parseOptions (optionsIn: ClientRequestConstructorOptions | string): Nod
     throw new TypeError('headers must be an object');
   }
 
-  const urlLoaderOptions: NodeJS.CreateURLLoaderOptions & { redirectPolicy: RedirectPolicy, extraHeaders: Record<string, string | string[]> } = {
+  const urlLoaderOptions: NodeJS.CreateURLLoaderOptions & { redirectPolicy: RedirectPolicy, headers: Record<string, [string, string | string[]]> } = {
```
nornagon, line 252: This tuple type is somewhat hard to read, I'd recommend instead using an object:
```suggestion
  const urlLoaderOptions: NodeJS.CreateURLLoaderOptions & { redirectPolicy: RedirectPolicy, headers: Record<string, {originalKey: string, value: string | string[]}> } = {
```

LuoJinghua: Fixed now.

PR #26138, opened 2020-10-23
perf: improve performance of heap snapshots.

### LINKED ISSUE #24509: takeHeapSnapshot on webview webContents is very slow in Electron 5+
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 5.x.x and higher
* **Operating System:**
  * Windows 10, Mac OS, Linux
  * 4.x.x

### Expected Behavior
duration of takeHeapSnapshot method called on webview webcontents should be comparable to the same operation on main window and with time in Electron prior v. 5

### Actual Behavior
takeHeapSnapshot on webview webcontents takes 10 and more times than in Electron 4, it performs especially slow on windows, but the regression seems to affect all the platforms. calling openDevTools before (but not immediatelly) "fixes" the problem and figures are then close to expected ones.

### To Reproduce

Call take takeHeapSnapshot on webview webconents on Electron 5 and higher, measure the time when the promise is resolved. Compare this time to Electron 4. In the same scenario, open the devtools on webcontents ~1s + before the snapshot is taken, observe the major reduction of the time for all subsequent snapshots.

Test app which shows the scenario  attached.

[wv-snapshot.zip](https://github.com/electron/electron/files/4904383/wv-snapshot.zip)

ckerr:
Fixes #24509.

#### Description of Change

Speed up `webContents.takeHeapSnapshot()`:

* Improve the efficiency of getting position info in slow mode. The current code does a linear walk through the source by calling [String::Get()](https://source.chromium.org/chromium/chromium/src/+/master:v8/src/objects/string-inl.h;l=419?q=String::Get&ss=chromium) for each character, which is somewhat expensive. This PR pulls the String's flat content into a local vector and then does a linear walk through that local vector.

* Cut in half the number of position lookups, regardless of whether we hit fast or slow mode in V8, by changing V8HeapExplorer::ExtractLocationForJSFunction() to call GetPositionInfo() instead of making separate calls for column and line number.

The test case at https://github.com/electron/electron/issues/24509 takes 7-8 seconds before this patch and 0.35 seconds with it.

Submitted upstream to V8 as well. Xref: https://chromium-review.googlesource.com/c/v8/v8/+/2496065

CC @codebytere 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Made `webContents.takeHeapSnapshot()` faster.


nornagon (changes requested):
Nice work! Leaving a 🛑  here so we don't merge before this is approved upstream.

deepak1556:
Looks like the upstream CL has been merged, @ckerr can you update the PR with the merged upstream commit. Thanks!

ckerr:
This is already in our `master` by virtue of the roll so I'm closing this PR.

Cherry-picks:
* 8-x-y: #26227
* 9-x-y: #26228 
* 10-x-y: #26229 
* 11-x-y: #26230

PR #26150, opened 2020-10-25
build: support for running clang-tidy

dsanders11:
##### Motivation
`clang-tidy` is a useful linter for catching code-style violations and potential issues like using null pointers, using an object after move, etc. Running it on the Electron code-base has already caught a couple of potential issues such as #26095 and #26115, with several more similar issues that haven't been fixed yet.

There are lots of potential checks in`clang-tidy`, some more noise than signal. For now it will only run the ones that Chromium uses (in `src/.clang-tidy`) but there are more useful checks (IMO) that can be enabled at a later date when it runs clean. One of the `clang-analyzer-webkit` checks causes a crash for me, in case anyone runs into that when enabling more checks.

##### Implementation
It's easy to checkout `clang-tidy` with Chromium using `checkout_clang_tidy` in DEPS. This brings in the executable during sync, so it adds a few dozen extra megabytes to the tree but that should be the only difference.

Created a runner script for it. It needs to know the build directory so it's not appropriate for adding to the general `yarn lint` command, and it's also very slow and memory hungry, so a bad idea anyway.

There's a `run-clang-tidy.py` in the LLVM project, but this implementation adds some sugar by making `compile_commands.json` for you, and is (IMO) a better parallel runner because it passes multiple filenames per invocation instead of just one, meaning less processes spinning up and down and parsing the compilation DB each time. For Electron the compilation DB is ~300 MB, and each invocation of `clang-tidy` uses ~1.2 GB of memory (on Windows) partially because of the size of the compilation DB. The runner script parses the compilation DB to filter filenames, so I brought in `stream-json` to minimize the memory footprint while doing so. Due to the large memory footprint, a single `clang-tidy` instance is the default, use `-j N` with N greater than 1 at your own risk, it doesn't take much to hit OOM errors.

Can be run with `yarn lint:clang-tidy --out-dir ../out/Testing`. Checks can be changed with `--checks=-*,performance*` style options. The `-` prefix disables checks. See `clang-tidy`'s documentation for a full list of checks.

Ideally this can be set up as a CI build once the noise is cleaned up.

cc @codebytere @ckerr @nornagon 

#### Checklist

#### Release Notes

Notes: none


dsanders11:
Gonna keep a list of interesting/fruitful checks:

* [readability-inconsistent-declaration-parameter-name](https://clang.llvm.org/extra/clang-tidy/checks/readability-inconsistent-declaration-parameter-name.html)

ckerr (commented):
I generally like this idea.

There is an open-ended question of how to use tools like this as a team, e.g. which rules to use and which to skip. @dsanders11 do you know how upstream Chromium projects make use of clang-tidy? Following upstream use is one thing that Electron tries to be good at :slightly_smiling_face:

dsanders11:
> @dsanders11 do you know how upstream Chromium projects make use of clang-tidy?

[Here's the doc for `clang-tidy` in the Chromium source source tree](https://chromium.googlesource.com/chromium/src/+/master/docs/clang_tidy.md). This PR uses the checks they've configured to use since `clang-tidy` looks up the tree for a `.clang-tidy` file and will find the one at `src/.clang-tidy`.

From the doc, relevant to your concern:

> We want these checks to cover as much as we reasonably can, but we also strive to strike a reasonable balance between signal and noise on code reviews. Hence, a large number of clang-tidy checks are disabled.

And here's the checks enabled in their `.clang-tidy`:

* bugprone-string-integer-assignment
* bugprone-unused-raii
* bugprone-use-after-move
* google-build-explicit-make-pair
* google-explicit-constructor
* google-readability-casting
* modernize-avoid-bind
* modernize-loop-convert
* modernize-make-shared
* modernize-make-unique
* modernize-redundant-void-arg
* modernize-replace-random-shuffle
* modernize-shrink-to-fit
* modernize-use-bool-literals
* modernize-use-default-member-init
* modernize-use-emplace
* modernize-use-equals-default
* modernize-use-equals-delete
* modernize-use-noexcept
* modernize-use-nullptr
* modernize-use-override
* modernize-use-transparent-functors
* readability-redundant-member-init

> There is an open-ended question of how to use tools like this as a team, e.g. which rules to use and which to skip

My personal preference would be a superset of the checks listed above. My above comment lists a few, and those have popped out some PRs that help tidy up the codebase. There's initial investment to bring the codebase into compliance with the checks, but after that if it runs as a CI job it should help keep things tidy moving forward, like the other linters do..

zcbenz (approved):
It would be valuable to have a document for this command, similar to `docs/development/clang-format.md` (which is a bit outdated though).

dsanders11:
@zcbenz, that's a good idea. As long as it looks like this is heading toward landing I can do that. I'd also planned to add it to `build-tools` since then you don't have to supply the output directory yourself since `build-tools` knows that, so it'll just become `e clang-tidy` or something similar.

dsanders11:
@zcbenz, I've added a `docs/development/clang-tidy.md`.

zcbenz:
Thanks for writing it!

@electron/wg-upgrades This PR requires review from wg-upgrades.

miniak (approved):

deepak1556 (approved):
LGTM on the DEPS change

PR #26162, opened 2020-10-26
chore: use consistent parameter names

dsanders11:
#### Description of Change
Results of a `clang-tidy` check: keep parameter names consistent between declaration and definition. There were more flagged, but these were the easy ones where it was only a change in the header (with one exception).

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/electron_download_manager_delegate.h
+++ b/shell/browser/electron_download_manager_delegate.h
@@ -41,7 +41,7 @@ class ElectronDownloadManagerDelegate
   // Get the save path set on the associated api::DownloadItem object
   void GetItemSavePath(download::DownloadItem* item, base::FilePath* path);
   void GetItemSaveDialogOptions(download::DownloadItem* item,
-                                file_dialog::DialogSettings* settings);
+                                file_dialog::DialogSettings* options);
 
   void OnDownloadPathGenerated(uint32_t download_id,
                                content::DownloadTargetCallback callback,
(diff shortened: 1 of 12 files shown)


shell/browser/electron_download_manager_delegate.h
```diff
@@ -41,7 +41,7 @@ class ElectronDownloadManagerDelegate
   // Get the save path set on the associated api::DownloadItem object
   void GetItemSavePath(download::DownloadItem* item, base::FilePath* path);
   void GetItemSaveDialogOptions(download::DownloadItem* item,
-                                file_dialog::DialogSettings* settings);
+                                file_dialog::DialogSettings* options);
```
nornagon, line 44: hypernit: imo `settings` is the better name here as it matches the type name.

shell/browser/native_window.cc
```diff
@@ -350,7 +350,7 @@ bool NativeWindow::AddTabbedWindow(NativeWindow* window) {
   return true;  // for non-Mac platforms
 }
 
-void NativeWindow::SetVibrancy(const std::string& filename) {}
```
nornagon, line 353: !!

nornagon (approved):

shell/browser/electron_download_manager_delegate.h
ckerr: agreed, we could change the body instead

ckerr (approved):

shell/browser/electron_download_manager_delegate.h
dsanders11: Yea, there were a dozen or so cases like that I skipped over because it would require changing more than just the header, but this one slipped in. There's also the fun case where an `override` has a different name in the definition and the declaration, and neither match the name used in the original declaration that's being overriden. :face_with_head_bandage:

PR #26164, opened 2020-10-26

nornagon:
#### Description of Change
Using the native implementation of window.open avoids many headaches and workarounds that have built up over the years. We should make this the default, and deprecate and eventually remove the non-native version.

Apps that require custom window.open behavior similar to the non-native window.open shim can implement similar behavior using IPC if required. But I'm not aware of any situations which require it.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: nativeWindowOpen is now enabled by default.


nornagon:
Closed in favor of #23221.

Issue #26166, opened 2020-10-26
Changes to the proxy config are not immediately reflected by resolveProxy at session creation time


nornagon:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * All current versions of Electron (8.x, 9.x, 10.x, 11.x and main branch)
* **Operating System:**
  * Tested on macOS, but probably all platforms
* **Last Known Working Electron version:**
  * 6.1.12

### Expected Behavior
The result of session.resolveProxy should immediately reflect proxy config updated by session.setProxy once the promise returned by `setProxy` resolves.

### Actual Behavior
When invoked shortly following session creation, resolveProxy returns `DIRECT` instead of using the configured proxy options.

### To Reproduce
https://gist.github.com/5daeaf25587086e09f13f218d1de2bf6

The first `resolveProxy` call returns `DIRECT`, but subsequent calls return the correct value.

Uncommenting the `setTimeout` delay after creating the session causes the first `resolveProxy` call to return the correct value, so this appears to be related to session startup.

Possibly related: https://source.chromium.org/chromium/chromium/src/+/master:components/proxy_config/pref_proxy_config_tracker_impl.cc;l=116;drc=5bafbb92230ad5f59e1fc6e1d3f5d638c34b9dad

walmat:
Any progress on this?

PR #26167, opened 2020-10-26
test: deflake ses.setProxy test

nornagon:
#### Description of Change
Example flake: https://app.circleci.com/pipelines/github/electron/electron/31724/workflows/ad370a6d-7c65-44dc-bf8b-e3978e263fe5/jobs/699105

Ref #26166

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/spec-main/api-session-spec.ts
+++ b/spec-main/api-session-spec.ts
@@ -341,9 +341,16 @@ describe('session module', () => {
   describe('ses.setProxy(options)', () => {
     let server: http.Server;
     let customSession: Electron.Session;
+    let created = false;
 
     beforeEach(async () => {
       customSession = session.fromPartition('proxyconfig');
+      if (!created) {
+        // Work around for https://github.com/electron/electron/issues/26166 to
+        // reduce flake
+        await delay(100);
+        created = true;
+      }
     });
 
     afterEach(() => {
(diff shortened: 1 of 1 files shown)


codebytere (approved):

ckerr (commented):
+1 if this is only something that shows up in CI situations -- but is that the case here? Is this something that could hit users in production as well?

nornagon:
@ckerr see referenced issue

deepak1556:
Maybe use `session.forceReloadProxyConfig` from https://github.com/electron/electron/pull/24937 instead of a delay

nornagon:
@deepak1556 I think the real fix is to have `ses.setProxy` return a promise that actually waits until the config has been applied, instead of what it does now which just a simple next-tick promise that doesn't wait for anything. this is a stopgap measure

deepak1556:
Sounds good

LuoJinghua:
The TEST_F(NetworkContextTest, StaticProxyConfig) from services/network/network_service_unittest.cc says:
``` c++
  // Kick the ConfiguredProxyResolutionService into action, as it doesn't start
  // updating its config until it's first used.
  proxy_resolution_service->ForceReloadProxyConfig();
```

nornagon:
@LuoJinghua good info. I'd still like to look into making the promise that `setProxy` returns more useful.

nornagon:
Also, hm, that comment seems inaccurate in our situation. Based on my experimentation in #26166, a `setTimeout` is sufficient to allow the change to propagate, no usage is required.

PR #26172, opened 2020-10-26
feat: correctly identify permissions when using setPermissionRequestHandler

### LINKED ISSUE #26089: navigator.clipboard.read() requests 'unknown' permission
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 11.0.0-beta.13
  * 10.0.1
* **Operating System:**
  * Windows 10 (2004)

When calling [`navigator.clipboard.read()`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read) in the renderer process, the session's permission handler receives a request for the permission "unknown".

### Expected Behavior
The permission handler should receive a request for "clipboard-read" permission.

### Actual Behavior
The permission handler receives a request for "unknown" permission.

### To Reproduce
Electron Fiddler: https://gist.github.com/5046d997fb4648edd08ad8a2645e84a0

jkleinsc:
#### Description of Change

When using `setPermissionRequestHandler` if  `navigator.clipboard.read()`, the permission would be described as `unknown`.  This PR fixes that permission to be `clipboard-read` which complies with https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read.

Additionally, I've mapped all of the permissions from https://source.chromium.org/chromium/chromium/src/+/master:content/browser/devtools/protocol/browser_handler.cc;l=141 so that in the future other permission types are not seen as `unknown`.

Resolves #26089
#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: updated setPermissionRequestHandler to correctly recognize permissions being requested.

### DIFF
--- a/shell/common/gin_converters/content_converter.cc
+++ b/shell/common/gin_converters/content_converter.cc
@@ -178,6 +178,10 @@ v8::Local<v8::Value> Converter<content::PermissionType>::ToV8(
   // Not all permissions are currently used by Electron but this will future
   // proof these conversions.
   switch (val) {
+    case content::PermissionType::ACCESSIBILITY_EVENTS:
+      return StringToV8(isolate, "accessibilityEvents");
+    case content::PermissionType::AR:
+      return StringToV8(isolate, "ar");
     case content::PermissionType::BACKGROUND_FETCH:
       return StringToV8(isolate, "background-fetch");
     case content::PermissionType::BACKGROUND_SYNC:
@@ -186,6 +190,8 @@ v8::Local<v8::Value> Converter<content::PermissionType>::ToV8(
       return StringToV8(isolate, "clipboard-read");
     case content::PermissionType::CLIPBOARD_SANITIZED_WRITE:
       return StringToV8(isolate, "clipboard-sanitized-write");
+    case content::PermissionType::FLASH:
+      return StringToV8(isolate, "flash");
     case content::PermissionType::CAMERA_PAN_TILT_ZOOM:
     case content::PermissionType::FONT_ACCESS:
       return StringToV8(isolate, "font-access");
@@ -216,11 +222,15 @@ v8::Local<v8::Value> Converter<content::PermissionType>::ToV8(
       return StringToV8(isolate, "screen-wake-lock");
     case content::PermissionType::SENSORS:
       return StringToV8(isolate, "sensors");
+    case content::PermissionType::STORAGE_ACCESS_GRANT:
+      return StringToV8(isolate, "storage-access");
+    case content::PermissionType::VR:
+      return StringToV8(isolate, "vr");
     case content::PermissionType::WAKE_LOCK_SYSTEM:
       return StringToV8(isolate, "system-wake-lock");
     case content::PermissionType::WINDOW_PLACEMENT:
       return StringToV8(isolate, "window-placement");
-    default:
+    case content::PermissionType::NUM:
       break;
   }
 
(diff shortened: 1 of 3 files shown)


MarshallOfSound (commented):
Should we also support `CLIPBOARD_SANITIZED_WRITE` here?

codebytere:
From the looks of it, we may want to make the converter handle most of https://source.chromium.org/chromium/chromium/src/+/master:components/permissions/permission_util.cc;l=130-190?q=PermissionType::CLIPBOARD_READ_WRITE&ss=chromium 🤔

jkleinsc:
@MarshallOfSound I couldn't figure out a way to make `CLIPBOARD_SANITIZED_WRITE` fire.  I suppose we could wire more up but I don't think we will be able to test them all.

jkleinsc:
@codebytere @MarshallOfSound I added all of the possible permissions based on https://source.chromium.org/chromium/chromium/src/+/master:content/browser/devtools/protocol/browser_handler.cc;l=141.

shell/common/gin_converters/content_converter.cc
```diff
@@ -188,6 +212,14 @@ v8::Local<v8::Value> Converter<content::PermissionType>::ToV8(
...
       return StringToV8(isolate, "midi");
+    case content::PermissionType::WAKE_LOCK_SCREEN:
+      return StringToV8(isolate, "screen-wake-lock");
+    case content::PermissionType::SENSORS:
+      return StringToV8(isolate, "sensors");
+    case content::PermissionType::WAKE_LOCK_SYSTEM:
+      return StringToV8(isolate, "system-wake-lock");
+    case content::PermissionType::WINDOW_PLACEMENT:
+      return StringToV8(isolate, "window-placement");
     default:
```
MarshallOfSound, line 223: If we've truly handled all possible permissions can we remove this `default` case as it should be unreachable?

codebytere: should we remove it? or just replace it with an `UNREACHABLE()` `DCHECK`?

MarshallOfSound: If we remove it the compiler will yell at us if we miss a new enum value

MarshallOfSound (commented):
Just noting, this isn't a `fix` anymore. It's a `feat`

shell/common/gin_converters/content_converter.cc
jkleinsc: @codebytere We don't want an unreachable here because there is another switch statement after this one to handle some additional permission types Electron has defined.  @MarshallOfSound As far as default, I thought it was a coding style thing to always include a default on a switch statement.

jkleinsc: Google's [style guide](https://google.github.io/styleguide/cppguide.html#Loops_and_Switch_Statements) says that:
>If not conditional on an enumerated value, switch statements should always have a default case

But PermissionType has this weird "NUM" type that isn't used: https://chromium.googlesource.com/chromium/src/+/master/content/public/browser/permission_type.h#55

jkleinsc: Ah... chromium just uses a break on `NUM`:
https://source.chromium.org/chromium/chromium/src/+/master:components/permissions/permission_manager.cc;l=133

I'll update to handle all the defined enum values and remove the default.

jkleinsc:
> Just noting, this isn't a `fix` anymore. It's a `feat`

Is it though?  You could argue that returning 'unknown' was a bug.

ckerr (approved):
LGTM.

As for the fix vs. feature question, it really does feel like it's kind of both. Though it looks like _if_ we treat this as a feature, it is a semver/minor one and backporting does not look problematic.

jkleinsc:
Merging as mac failure is flake unrelated to this PR.

PR #26176, opened 2020-10-26
chore: tidy up usage of PendingRemote

dsanders11:
#### Description of Change
Tidies up some of the usage of `mojo::PendingRemote` from #25393.

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/net/electron_url_loader_factory.cc
+++ b/shell/browser/net/electron_url_loader_factory.cc
@@ -213,9 +213,11 @@ void ElectronURLLoaderFactory::OnComplete(
     mojo::PendingRemote<network::mojom::URLLoaderClient> client,
     int32_t request_id,
     const network::URLLoaderCompletionStatus& status) {
-  mojo::Remote<network::mojom::URLLoaderClient> client_remote(
-      std::move(client));
-  client_remote->OnComplete(status);
+  if (client.is_valid()) {
+    mojo::Remote<network::mojom::URLLoaderClient> client_remote(
+        std::move(client));
+    client_remote->OnComplete(status);
+  }
 }
 
 // static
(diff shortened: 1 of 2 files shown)


shell/browser/net/proxying_url_loader_factory.cc
```diff
@@ -802,16 +802,16 @@ void ProxyingURLLoaderFactory::CreateLoaderAndStart(
   // Check if user has intercepted this scheme.
   auto it = intercepted_handlers_.find(request.url.scheme());
   if (it != intercepted_handlers_.end()) {
-    mojo::Remote<network::mojom::URLLoaderFactory> loader_remote;
-    this->Clone(loader_remote.BindNewPipeAndPassReceiver());
+    mojo::PendingRemote<network::mojom::URLLoaderFactory> loader_remote;
+    this->Clone(loader_remote.InitWithNewPipeAndPassReceiver());
```
dsanders11, line 806: Didn't seem to make sense to create a `mojo::Remote` just to unbind it, so go straight to `mojo::PendingRemote`.

shell/browser/net/electron_url_loader_factory.cc
```diff
@@ -294,14 +296,13 @@ void ElectronURLLoaderFactory::StartLoading(
     // module.
     //
     // I'm not sure whether this is an intended behavior in Chromium.
-    mojo::Remote<network::mojom::URLLoaderFactory> proxy_factory_remote(
-        std::move(proxy_factory));
+    if (proxy_factory.is_valid()) {
+      mojo::Remote<network::mojom::URLLoaderFactory> proxy_factory_remote(
+          std::move(proxy_factory));
```
dsanders11, line 301: You shouldn't bind invalid `mojo::PendingRemote` objects. For some reason the `mojo` code doesn't do that `DCHECK` if you bind via the constructor, rather than doing `proxy_factory_remote.Bind(std::move(proxy_factory))`. Binding then checking if bound is not a proper code path, the correct way is to check for validity first then bind.

deepak1556 (approved):

deepak1556:

Issue #25695, opened 2020-09-29
Electron apps based are not accessible via MacOS "Hover Text" unless Voice Over is turned on


rfairweather:
**Issue: Electron apps based are not accessible via Hover Text unless Voice Over is turned on.**

### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details
* **Electron Version:**
  * All versions
* **Operating System:**
  * MacOS Catalina
* **Last Known Working Electron version:**
  * None

### Expected Behavior
Hover Text appears for the text in the Electron app

### Actual Behavior
Hover Text does not appear (but it does on Chromium)

### To Reproduce
1. Install any Electron app.
2. Launch the Electron app.
3. In MacOS System Preferences, go to Accessibility > Zoom, and enable Hover Text.
4. Hover over text in the Electron app.

Now observe what happens when we enable Voice Over...

5. Turn on Voice Over
6. Hover over text in the Electron app.
7. Observe that Hover Text shows for the text in the Electron app now that Voice Over is enabled

### Additional resources
https://support.apple.com/guide/mac-help/use-hover-text-mchlb203bc78/mac#:~:text=%20%20%201%20On%20your%20Mac%2C%20choose,menu%2C%20then%20choose%20a%20location.%20If...%20More%20

PR #10305, opened 2017-08-19
Special attribute for macOS accessibility

ivmirx:
In the issue [#7206](https://github.com/electron/electron/issues/7206) we were discussing that Electron apps are inaccessible unless VoiceOver is enabled. While it's a working solution for users with vision impairment, all other users and apps that require accessibility can't interact with Electron-based software because they don't keep VoiceOver running.

I suggest adding `AXManualAccessibility` for programmatically enabling it in Electron apps. The reason for a new attribute is that `AXEnhancedUserInterface` is already reserved by VoiceOver.

Adding this attribute will allow both Electron developers and 3rd party developers to enable and disable accessibility from their code by calling `accessibilitySetValue:forAttribute:` on the application.

It will be also possible to create a small utility app to switch accessibility in Electron-based apps until there's a native UI solution (like the accessibility settings page in Chrome).

PR #26183, opened 2020-10-27
fix: hover text only working when VO enabled

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/25695.

Previously some screenreader-associated accessibility tooling was being enabled exclusively when VoiceOver was turned on, which broke Hover Text among some other things. This is because we only called `ax_state->OnScreenReaderDetected();` if VoiceOver was enabled, and is fixed by removing that condition. The previous condition seems to have been added as a result of https://github.com/atom/atom/issues/3288, but some local testing seems to ensure that text to speech still works after this removal.

cc @ckerr @jkleinsc @zcbenz  

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where Hover Text on macOS Catalina did not work without VoiceOver also being enabled.

### DIFF
--- a/shell/browser/mac/electron_application.mm
+++ b/shell/browser/mac/electron_application.mm
@@ -178,26 +178,21 @@ inline void dispatch_sync_main(dispatch_block_t block) {
 }
 
 - (void)accessibilitySetValue:(id)value forAttribute:(NSString*)attribute {
-  // Undocumented attribute that VoiceOver happens to set while running.
-  // Chromium uses this too, even though it's not exactly right.
-  if ([attribute isEqualToString:@"AXEnhancedUserInterface"]) {
-    [self updateAccessibilityEnabled:[value boolValue]];
-  } else if ([attribute isEqualToString:@"AXManualAccessibility"]) {
-    [self updateAccessibilityEnabled:[value boolValue]];
+  // Undocumented attribute that screen reader related functionality
+  // sets when running.
+  if ([attribute isEqualToString:@"AXEnhancedUserInterface"] ||
+      [attribute isEqualToString:@"AXManualAccessibility"]) {
+    auto* ax_state = content::BrowserAccessibilityState::GetInstance();
+    if ([value boolValue]) {
+      ax_state->OnScreenReaderDetected();
+    } else {
+      ax_state->DisableAccessibility();
+    }
+
+    electron::Browser::Get()->OnAccessibilitySupportChanged();
   }
+
   return [super accessibilitySetValue:value forAttribute:attribute];
-}
-
-- (void)updateAccessibilityEnabled:(BOOL)enabled {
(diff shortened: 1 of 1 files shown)


ckerr (commented):
It looks like this behavior was introduced by https://github.com/electron/electron/commit/6bc59cf2d77d794c9533e11575e5ca1c03e3b5d8 and one of the [comments](https://github.com/atom/atom/issues/3288#issuecomment-146653745) for that patch says

> This looks like something has changed in Chrome. With the VoiceOver patch, I mimicked Chrome's behavior of enabling the accessibility tree when VoiceOver was active, and disabling the tree when inactive.

This comment is 5 years old so it's likely stale information, but just to check -- did you confirm that we're still tracking to how Chrome behaves wrt the accessibility tree?

ckerr (approved):
LGTM. Other than the addition of [AXManualAccessibility](https://github.com/electron/electron/pull/10305), this matches what chrome_browser_application_mac.mm is doing.

shell/browser/mac/electron_application.mm
```diff
@@ -177,36 +177,22 @@ - (void)handleURLEvent:(NSAppleEventDescriptor*)event
...
-  }
-  return [super accessibilitySetValue:value forAttribute:attribute];
-}
-
-- (void)updateAccessibilityEnabled:(BOOL)enabled {
-  auto* ax_state = content::BrowserAccessibilityState::GetInstance();
+  // Undocumented attribute that screen reader related functionality
+  // sets when running.
+  if ([attribute isEqualToString:@"AXEnhancedUserInterface"] ||
+      [attribute isEqualToString:@"AXManualAccessibility"]) {
```
zcbenz, line 184: It seems that only Electron is using `AXManualAccessibility` and it leads me to an interesting article:
https://balatero.com/writings/hammerspoon/retrieving-input-field-values-and-cursor-position-with-hammerspoon/#downsides-of-the-accessibility-api

Anyway I'm good with this since it aligns with Chromium's behavior, and removing `AXManualAccessibility` might cause issues.

zcbenz (approved):

shell/browser/mac/electron_application.mm
codebytere: I wonder if this is still the case? With the new alignment, the accessibility tree should be read if _either_ attribute is set 🤔

codebytere:
```
not ok 1330 webContents module render view deleted events does not emit current-render-view-deleted when speculative RVHs are deleted and nativeWindowOpen is set to true
  TypeError: Cannot read property 'webContents' of undefined
      at Timeout._onTimeout (electron\spec-main\api-web-contents-spec.ts:1213:21)
      at listOnTimeout (internal/timers.js:554:17)
      at processTimers (internal/timers.js:497:7)
```

failure is unrelated, merging.

PR #26188, opened 2020-10-27
test: disable backgrounding occluded windows

nornagon:
#### Description of Change
This flag is designed for deflaking visibility tests in Chrome. I was able to produce flakes on my local machine by cmd+tabbing while the test was running, but with this flag I'm no longer able to reproduce them.

#### Release Notes

Notes: none

### DIFF
--- a/script/spec-runner.js
+++ b/script/spec-runner.js
@@ -26,6 +26,10 @@ for (const flag of unknownFlags) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


zcbenz:
I have restarted the CI for a few times, and the `document.visibilityState` tests now seem to fail constantly on macOS CI.

PR #26189, opened 2020-10-27
refactor: merge CommonWebContentsDelegate into api::WebContents

nornagon:
#### Description of Change
Based on #24517 to minimize conflicts, i'll rebase to master when that lands.

`CommonWebContentsDelegate` is no longer "common" to anything, as its only subclass is `electron::api::WebContents`. This simplifies the class hierarchy by merging `CommonWebContentsDelegate` into `api::WebContents`.

This has the effect of making `api::WebContents` a fair bit larger, which isn't great, but the split between what was in `CommonWebContentsDelegate` and what was in `api::WebContents` was more or less arbitrary, and didn't serve to improve a mental map of the code.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


deepak1556 (commented):
LGTM , can sign off once compilation succceeds.

As a follow-up we can move the devtools related functions to `InspectableWebContents` class, the whole delegate style was needed when devtools core was in a separate component and the ui management was in electron. Now that these classes are closer to each other, don't see the need for separation.

nornagon:
> As a follow-up we can move the devtools related functions to `InspectableWebContents` class, the whole delegate style was needed when devtools core was in a separate component and the ui management was in electron. Now that these classes are closer to each other, don't see the need for separation.

Hm, which functions are you referring to? afaict `DevToolsSaveToFile` depends on `owner_window()` and `offscreen_`, which aren't available in `InspectableWebContents`. Same for `DevToolsAddFileSystem`.

deepak1556:
> which functions are you referring to? afaict DevToolsSaveToFile depends on owner_window() and offscreen_, which aren't available in InspectableWebContents. Same for DevToolsAddFileSystem.

Yes those are the ones, and yes some parameters aren't available now, but we can refactor in way these functions can be moved. Ideally would be good to keep devtools related stuff in one place.

deepak1556 (approved):

zcbenz (approved):

kaycoinsUSA (commented):
Great!

PR #26216, opened 2020-10-27
feat: add 'resized' and 'moved' events to BrowserWindow

### LINKED ISSUE #24934: BrowserWindow resized event
### Preflight Checklist

### Problem Description

BrowserWindow currently offers "moved" event related to moving a window, but it only has "will-resize" and "resize" events. I want to save the current dimensions of a window when its size changes, but "will-resize" and "resize" events get fired at every single pixel change. This results in an unnecessarily large number of saving function calls or a need for debouncing the event.

### Proposed Solution

Adding "resized" event to BrowserWindow will solve the problem stated above.

samuelmaddock:
@electron/wg-api 
@electron/wg-releases I'd like to target 10-x-y if possible.

#### Description of Change

Emits a `resized` event after the user has finished manually resizing a window.
Similarly, a `moved` event now emits on Windows when the window has finished being moved.

I'm not quite sure whether it's possible to test this manual method of resizing the window. For now, I've added tests for macOS' animated resizing.

resolves #24934

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Added 'resized' (Windows/macOS) and 'moved' (Windows) events to BrowserWindow.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -543,6 +543,12 @@ Note that this is only emitted when the window is being resized manually. Resizi
 
 Emitted after the window has been resized.
 
+#### Event: 'resized' _macOS_ _Windows_
+
+Emitted once when the window has finished being resized.
+
+This is usually emitted when the window has been resized manually. On macOS, resizing the window with `setBounds`/`setSize` and setting the `animate` parameter to `true` will also emit this event once resizing has finished.
+
 #### Event: 'will-move' _macOS_ _Windows_
 
 Returns:
@@ -558,11 +564,11 @@ Note that this is only emitted when the window is being resized manually. Resizi
 
 Emitted when the window is being moved to a new position.
 
-__Note__: On macOS this event is an alias of `moved`.
-
-#### Event: 'moved' _macOS_
+#### Event: 'moved' _macOS_ _Windows_
 
 Emitted once when the window is moved to a new position.
+
+__Note__: On macOS this event is an alias of `move`.
 
 #### Event: 'enter-full-screen'
 
--- a/shell/browser/api/electron_api_base_window.cc
+++ b/shell/browser/api/electron_api_base_window.cc
@@ -222,6 +222,10 @@ void BaseWindow::OnWindowResize() {
   Emit("resize");
 }
 
+void BaseWindow::OnWindowResized() {
+  Emit("resized");
+}
+
 void BaseWindow::OnWindowWillMove(const gfx::Rect& new_bounds,
                                   bool* prevent_default) {
   if (Emit("will-move", new_bounds)) {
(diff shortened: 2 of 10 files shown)


docs/api/browser-window.md
```diff
@@ -558,12 +564,12 @@ Note that this is only emitted when the window is being resized manually. Resizi
...
 Emitted when the window is being moved to a new position.
 
-__Note__: On macOS this event is an alias of `moved`.
-
-#### Event: 'moved' _macOS_
+#### Event: 'moved' _macOS_ _Windows_
 
 Emitted once when the window is moved to a new position.
 
+__Note__: On macOS this event is an alias of `move`.
```
samuelmaddock, line 571: I moved the Note into 'moved' since now the behavior only differs on this event. On Windows, 'moved' will only emit after the user has stopped moving the window. On macOS, it will always emit along with 'move'.

zcbenz: After moving the note, it reads to me like this: the `moved` event is not really implemented on macOS, it is just an alias of `move` event. But I think the actual behavior is reverse?

The original note was confusing too, so I think maybe we should clearly describe the behavior in the note.

samuelmaddock: That's how I see it as well. You had left a comment mentioning needing to figure out a proper way to dispatch move.
https://github.com/electron/electron/blob/d3f32c7502412ec143e25da41589ab923cf0b163/shell/browser/ui/cocoa/electron_ns_window_delegate.mm#L171-L177

For me, I assume `move` should follow similar behavior to how [`mousemove`](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousemove_event) works. It should be emitted at each change of the position of the window.

`moved` should be emitted after moving is finished and should follow after the last `move` event. However, macOS doesn't have this behavior and will instead always emit after `move`.

Does this make sense?

MarshallOfSound:
@samuelmaddock Can you update the PR title / release notes to include a reference to the addition of the "moved" event on Windows

jkleinsc:
The releases WG approved this PR for backport to 10-x-y and 11-x-y at the Oct 28, 2020 meeting

samuelmaddock:
> 
> 
> @samuelmaddock Can you update the PR title / release notes to include a reference to the addition of the "moved" event on Windows

@MarshallOfSound I tried my best 😅

docs/api/browser-window.md
zcbenz: Yeah you are right.

I thought `windowDidMove` would emit when moving is finished, but after testing I found it actually keeps emitting during the moving. So moving the note does make the doc more clear 👍 .

zcbenz (approved):

miniak (approved):

jkleinsc:
The API WG approved this PR at our Nov 9, 2020 meeting.

jkleinsc (approved):

matt-huh:
> I was unable to backport this PR to "10-x-y" cleanly;
> you will need to perform this backport manually.

@jkleinsc @codebytere Hi, will this PR not be included in version 10?

PR #26232, opened 2020-10-28
build: only check patch diffs in testing builds

codebytere:
We should only be testing whether or not the patch diffs have changed in testing builds - diffs during the release process should not fail the release process itself.

cc @jkleinsc @MarshallOfSound 

Notes: none

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,6 +107,7 @@ env-release-build: &env-release-build
   STRIP_BINARIES: true
   GENERATE_SYMBOLS: true
   CHECK_DIST_MANIFEST: '1'
+  IS_RELEASE: true
 
 env-headless-testing: &env-headless-testing
   DISPLAY: ':99.0'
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -256,23 +258,25 @@ step-gclient-sync: &step-gclient-sync
...
-          # Export it
-          mkdir -p ../../patches
-          git format-patch -1 --stdout --keep-subject --no-stat --full-index > ../../patches/update-patches.patch
-          echo
-          echo "======================================================================"
-          echo "There were changes to the patches when applying."
-          echo "Check the CI artifacts for a patch you can apply to fix it."
-          echo "======================================================================"
-          exit 1
+        if [ "$IS_RELEASE" == "false" ]; then
```
MarshallOfSound, line 261:
```suggestion
        if [ "$IS_RELEASE" != "true" ]; then
```

We should fail into the "not release" state

.circleci/config.yml
```diff
@@ -101,12 +101,14 @@ machine-mac-large-arm: &machine-mac-large-arm
 env-testing-build: &env-testing-build
   GN_CONFIG: //electron/build/args/testing.gn
   CHECK_DIST_MANIFEST: '1'
+  IS_RELEASE: false
```
MarshallOfSound, line 104: Don't need this if we change the if condition below

MarshallOfSound (approved):

jkleinsc (approved):
LGTM

codebytere:
Merging `fast-track` as the failures are not germane to this config change.

Issue #26116, opened 2020-10-22
BrowserView mouse capture region has incorrect bounds.


ccorcos:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** v10.1.4
  * 
* **Operating System:** MacOS Big Sur
  * 

I created a full repro here: https://github.com/ccorcos/electron-browser-view-bounds

```
git clone git@github.com:ccorcos/electron-browser-view-bounds.git
cd electron-browser-view-bounds
npm install 
npm start
```

You can also view the repro example here: https://github.com/ccorcos/electron-browser-view-bounds/blob/main/example.mov

Basically we can resize the view with the window using `view.setAutoResize({ width: true, height: true })` or manually with: 
```
  mainWindow.on('resize', () => {
    const [width, height] = mainWindow.getSize()
    view.setBounds({ x: 0, y: 0, width, height })
  })
```

And the view does properly resize, but the mouse event capture region that gets passed on to the view remains the same size and fits into the bottom right region of the window.

codebytere:
@ccorcos do you know if this worked previously, and if so, in what version?

ccorcos:
@codebytere I just tested it out and it seems this has been broken. For a long time. I created a branch for each version.

```
git checkout electron-11; npm install; npm start # broken
git checkout electron-9; npm install; npm start # broken
git checkout electron-8; npm install; npm start # broken
git checkout electron-7; npm install; npm start # broken
git checkout electron-6; npm install; npm start # broken
git checkout electron-5; npm install; npm start # broken
git checkout electron-4; npm install; npm start # broken
git checkout electron-3; npm install; npm start # broken
```

Electron 2 has an error for some other reason about `app.whenReady` but this is ancient history so I get the feeling that this has always been an issue.
And

ccorcos:
I just added the following which makes it even more clear what the issue is:
```
  document.addEventListener('mouseenter', () => {
    document.body.style.color = 'blue'
  })
  document.addEventListener('mouseleave', () => {
    document.body.style.color = 'red'
  })
```

codebytere:
@ccorcos PR is up :)

[...]

Issue #22780, opened 2020-03-20
Cursor blocked after setBounds on frameless window


johnryan:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 8.1.1
* **Operating System:**
  * macOS 10.14.6
* **Last Known Working Electron version:**
  * n/a

### Background
It seems that when resizing a frameless window *with* a BrowserView inside it to a sufficiently narrow width there is an area created at the top of the window that blocks all touches and makes the window draggable (even if the window has movable = false)

### Expected Behavior
- Calling setBounds should allow the window to resize without disrupting the functionality of the BrowserView

### Actual Behavior
Approximately 25% of the window at the top stops accepting input and becomes a drag region

### To Reproduce
**Electron Fiddle:** https://gist.github.com/f80bfd413b7211c31e2feeb0608cafc8

### Screenshots
![ezgif-7-09195467100b](https://user-images.githubusercontent.com/416564/77129173-62157c00-6a29-11ea-9ca4-9bcf79cd3fab.gif)

praneetrohida:
Same thing happens for me on 9.1.0. I have a BrowserView inside a non-frameless window. The bounds of the BrowserView are calculated dynamically and at a certain point the BrowserView stops accepting the mouse input and becomes a drag region

PR #26233, opened 2020-10-28
fix: draggable region edge calculation on resize

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/26116.
Closes https://github.com/electron/electron/issues/22780.

Fixes an issue where draggable regions were not properly updated on BrowserViews when a containing `BrowserWindow` was resized. This was occurring because we were only updating draggable regions when the regions were non-empty, which was a logical error since the width and height will always need to be recalculated for a `BrowserView` contained within a frameless `BrowserWindow` on resize. Beyond that, it's the case that `BrowserView` bounds will not be set during the first call to `UpdateDraggableRegions` so we need to ensure the draggable regions are updated when the bounds _do_ get set. 

Tested with 
- https://gist.github.com/2c072f42d7abaf27e73961b1dde87811
- https://gist.github.com/codebytere/aab0eb3835ed4550e2498ef63c0a0be2

cc @deepak1556 @MarshallOfSound @zcbenz (and @ccorcos!)

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixes an issue where draggable regions were not properly updated on BrowserViews when a containing `BrowserWindow` was resized.

### DIFF
--- a/shell/browser/api/electron_api_browser_window.cc
+++ b/shell/browser/api/electron_api_browser_window.cc
@@ -14,6 +14,7 @@
 #include "content/public/browser/render_view_host.h"
 #include "shell/browser/api/electron_api_web_contents_view.h"
 #include "shell/browser/browser.h"
+#include "shell/browser/native_browser_view.h"
 #include "shell/browser/unresponsive_suppressor.h"
 #include "shell/browser/web_contents_preferences.h"
 #include "shell/browser/window_list.h"
@@ -304,9 +305,15 @@ void BrowserWindow::OnWindowIsKeyChanged(bool is_key) {
 
 void BrowserWindow::OnWindowResize() {
 #if defined(OS_MAC)
-  // Always update draggable regions here to ensure draggable bounds
-  // are recalculated for BrowserViews if any exist.
-  UpdateDraggableRegions(draggable_regions_);
+  if (!draggable_regions_.empty()) {
+    UpdateDraggableRegions(draggable_regions_);
+  } else {
+    // Ensure draggable bounds are recalculated for BrowserViews if any exist.
+    auto browser_views = window_->browser_views();
+    for (NativeBrowserView* view : browser_views) {
+      view->UpdateDraggableRegions(draggable_regions_);
+    }
+  }
 #endif
   BaseWindow::OnWindowResize();
 }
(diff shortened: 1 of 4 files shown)


shell/browser/api/electron_api_browser_window.cc
```diff
@@ -304,8 +304,9 @@ void BrowserWindow::OnWindowIsKeyChanged(bool is_key) {
 
 void BrowserWindow::OnWindowResize() {
 #if defined(OS_MAC)
-  if (!draggable_regions_.empty())
-    UpdateDraggableRegions(draggable_regions_);
+  // Always update draggable regions here to ensure draggable bounds
+  // are recalculated for BrowserViews if any exist.
+  UpdateDraggableRegions(draggable_regions_);
```
deepak1556, line 309: If there is a draggable region to update on resize then it would be signalled by `RenderFrameObserver::DraggableRegionsChanged` and the changes should get propagated via `ExtendedWebContentsObserver::OnDraggableRegionsUpdated`. Why should this update `Browserview` regions for empty  `draggable_regions_` ?

codebytere: @deepak1556 when draggable regions are empty on a frameless window, we create one giant `NSView` that is draggable for associated `BrowserViews`, and then on top of that add "exclusion zones", which for frameless windows with _no draggable regions_ will be the full size of the BrowserWindow. If we resize the window, what that means is that the BrowserWindow will resize, but the `NSView` we created _won't_ be resized, so we'll end up with incorrectly resizable _parts_ of the window.

deepak1556: In that case can we just call to update the draggable regions of browser view here. I am looking to short-circuit the situation where there is no browser view and there are empty draggable regions for window.

deepak1556 (approved):

zcbenz (approved):

castroCrea:
@codebytere Have you notice that 
```
  win.setVisibleOnAllWorkspaces(true, {
    visibleOnFullScreen: true,
  });
```
work with out app.dock.hide() in you test
ut when you do a modification on the browserWindow it doesn't work anymore with out ```app.dock.hide()```
just add     ```win.webContents.openDevTools({ mode: "detach" });``` and then you will need ```app.dock.hide()```

So It might has a bug somewhere

PR #26236, opened 2020-10-28
chore: don't self-include inspectable_web_contents.h

nornagon:
behavioral no-op. was fine because of header guards.

Notes: none


jkleinsc (approved):
Screen Shot 2020-10-28 at 4 51 37 PM

ckerr (approved):

nornagon:
CI failures appear to be unrelated.

PR #26185, opened 2020-10-27
fix: set app locale after user's script is loaded

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/5649.

This PR fixes the bug that, calling `app.commandLine.appendSwitch('lang')` could not change app's locale.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Fix calling `app.commandLine.appendSwitch('lang')` not changing app's locale.

PR #26241, opened 2020-10-29
fix: set app locale after user's script is loaded

zcbenz:
Backport of #26185

See that PR for details.

Notes: Fix calling `app.commandLine.appendSwitch('lang')` not changing app's locale.


codebytere (approved):

codebytere:
> <webview> tag did-change-theme-color event emits when theme color changes

is a known failure.

PR #26186, opened 2020-10-27
build: drop pywin32 usage

dsanders11:
#### Description of Change
Chromium has removed it's usage of `pywin32` in favor of not needing it as a dependency. The Electron code only directly uses it in one place, which I've swapped out for using an equivalent Powershell command.

Dropping its usage moves Windows a step closer to not needing to install anything other than `depot_tools` as a prerequisite, since `depot_tools` bootstraps Python and Git when `gclient` is used.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

#### Release Notes

Notes: none

PR #26091, opened 2020-10-21
build: Python3 compat and Win line-endings fixes

dsanders11:
#### Description of Change
* Fixes line endings when exporting patches on Windows
* Python3 compatibility fixes

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

#### Release Notes

Notes: none

PR #26252, opened 2020-10-29
chore: cleanup pylint violations

dsanders11:
#### Description of Change
Just file length violations and a couple unused variables and shadowing builtin names.

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/build/npm-run.py
+++ b/build/npm-run.py
@@ -15,5 +15,12 @@ args = [cmd, "run",
 try:
     subprocess.check_output(args, stderr=subprocess.STDOUT)
 except subprocess.CalledProcessError as e:
-    print("NPM script '" + sys.argv[2] + "' failed with code '" + str(e.returncode) + "':\n" + e.output)
+    print(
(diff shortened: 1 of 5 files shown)


codebytere (approved):

miniak:
I don't like this change, 80 characters per line seems backwards, can we have at least 100 or 120?

dsanders11:
@miniak, Electron follows Chromium's Python style guide, which is just PEP-8, and includes 80 characters per line. The `pylint` linter enforces that with a config that's in `depot_tools`.

I agree that 80 characters per line is often restricting, but considering Python is really only used for some scripting within Electron I think it makes the most sense to follow Chromium's style.

dsanders11:
Added fixes for `src/build/profile_toolchain.py` now that #26186 is on pause. This PR should now cover all `pylint` fixes outside of #26091.

zcbenz (approved):

miniak (approved):

PR #26253, opened 2020-10-29
build: fix pylint invocation

dsanders11:
#### Description of Change
Gets `pylint` running again.

Leaving as a draft for now as it needs several other PRs to land first. Refs electron/build-tools-installer#6, #26091, #26252

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/script/lint.js
+++ b/script/lint.js
@@ -11,7 +11,6 @@ const path = require('path');
 
 const ELECTRON_ROOT = path.normalize(path.dirname(__dirname));
 const SOURCE_ROOT = path.resolve(ELECTRON_ROOT, '..');
-const DEPOT_TOOLS = path.resolve(SOURCE_ROOT, 'third_party', 'depot_tools');
 
 const IGNORELIST = new Set([
   ['shell', 'browser', 'resources', 'win', 'resource.h'],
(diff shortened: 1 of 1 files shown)


dsanders11:
Blocking PRs on this repo have been merged, rebased this on `master`. Linting on CI passes now.

Still blocked by electron/build-tools#224, or merging this will break linting `python` and `gn` for those on `build-tools`.

zcbenz:
@dsanders11 It has been a long time and I believe everyone should have updated build-tools to needed version, so I think we should be able to merge this without https://github.com/electron/build-tools/pull/224/files?

dsanders11:
@zcbenz, unfortunately that blocking PR is still needed, for multiple reasons.

* While https://github.com/electron/build-tools-installer/pull/6 was merged, it was never released - `@electron/build-tools` on NPM was last published 2 years ago
* https://github.com/electron/build-tools/pull/224 is still needed because a new release of `@electron/build-tools` won't update the package on existing systems with `build-tools` installed - that's what the PR looks to solve, by checking if it needs to be updated

So the steps to unblock this PR are: 
* Publish a new version of `@electron/build-tools`
* Merge https://github.com/electron/build-tools/pull/224

Until I can get some movement on those other steps, this remains in limbo. Not sure who I need to ping for attention, I haven't had any luck getting movement in the past so I let it sit.

nornagon:
cc @MarshallOfSound

dsanders11:
Context on why this change is needed, from a Slack convo around the time this was created:

> The issue that has me mucking around in all of this is that `pylint` hasn't been running due to a path issue. Fixing the path issue leads to a different failure for me, in my tree that was set up with `build-tools`. Namely that issue is that `lint.js` specifically runs the in-tree `depot_tools` version of `pylint`, but the in-tree `depot_tools` was never bootstrapped on Windows (get's done when you run `gclient`, but all the `gclient` commands from `build-tools` run in the out-of-tree `depot_tools`). So it fails there. A fix would be to try to ensure it's running from the out-of-tree version instead, but that led me down the road of why there's two and when to use one or the other.

PR #26274, opened 2020-10-30
fix: Linux npm install error

tiagotakeda:
#### Description of Change

I faced this issue when I tried to install electron on Ubuntu 18.04 LTS. After I followed this steps the issue was resolved. 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:


zcbenz (commented):
Thanks for the contribution, and I think it is valuable information when setting up Node.js, but this documentation is more about installing Electron, and it assumes the Node.js environment has already been installed. So we can not merge this PR.

PR #26289, opened 2020-10-30
fix: [webview] fix missing properties on events when contextIsolation: true

### LINKED ISSUE #25143: Webview: the 'console-message' event doesn't have the appropriate properties when contextIsolation is true
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 10.0.0
* **Operating System:**
  * Windows 10 (1909)
  * unknown

### Expected Behavior
When listen to the 'console-message' event of a webview element the event should have a 'message' property.

### Actual Behavior
This works perfectly when contextIsolation is set to 'false' (or not at all in 10.0.0).

### To Reproduce
* Create a BrowserWindow with contextIsolation: false in webPreferences.
* Create a webview in the BrowserWindow.
* Add a 'console-message' listener to the webview.
* Load an URL in webview and print something in the console from the guest page.
* The console message should appear as a property of the event (event.message).
* Now set contextIsolation: true.
* Load the URL in the webview.
* The 'console-message' event will be fired, but the event.message property isn't available anymore.

https://gist.github.com/AmarBouchibane/80b89651d776f0b1cdf11748e6a2d090

nornagon:
#### Description of Change
Poke a hole in contextIsolation and let &lt;webview&gt; events through. This is
ill-advised from a security perspective, but `webviewTag: true` already pokes
holes in contextIsolation, so this is at least not making the problem
substantially worse.

If in future we want to close this hole, we can do so by
serializing/deserializing the event properties to make them "native" to the
main world.

Fixes #25143.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where events on webview elements were missing properties if contextIsolation was enabled.


miniak (approved):

codebytere (approved):

zcbenz (approved):

nornagon:
Backporting is not completely trivial, and this is a fairly obscure fix, so I'll hold on it for now.

PR #26296, opened 2020-11-01
fix: fail protocol request when OnWrite fails

dsanders11:
#### Description of Change

Wasn't sending a failed response for a protocol request in an error case, could have left the request without a response. I don't know of a way to easily test this, wasn't able to trigger it, seems like a corner-case.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: fixed error handling on write failure in the protocol module

### DIFF
--- a/shell/browser/net/electron_url_loader_factory.cc
+++ b/shell/browser/net/electron_url_loader_factory.cc
@@ -153,15 +153,13 @@ struct WriteData {
 };
 
 void OnWrite(std::unique_ptr<WriteData> write_data, MojoResult result) {
-  if (result != MOJO_RESULT_OK) {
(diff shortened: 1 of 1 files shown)


deepak1556:
This is a user facing change, can you add some release notes for it.

LGTM, otherwise.

dsanders11:
Added a release note. Wasn't sure how to word it since I'm still not clear on what condition would trigger it to begin with. 😆 I kept it generic.

aishwaryakarad:
Haha nice I respect that. I’m clueless about how to get even started to use electron for my web dev. Any shortcuts or guides y’all could recommend would be of real help

deepak1556 (approved):

codebytere (approved):

zcbenz (approved):

PR #26297, opened 2020-11-01
fix: improve redirect behavior in protocol module

### LINKED ISSUE #25894: `intercept*Protocol` Methods Break Redirect Handling Routines
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * `<=v10.1.3`
* **Operating System:**
  * All
  * N/A

### Expected Behavior
Redirect handling for non-GET requests in scenarios where HTTP/HTTPS requests are intercepted using the `interceptHttpProtocol` and `interceptStreamProtocol` handlers should follow the same standards as non-intercepted requests. This should be true for `BrowserWindow`, `BrowserView` and `webview`. For example, the following is an example of `GET` request being redirected and the subsequent request thereafter:

```http
client> GET / HTTP/1.1
client> Host: foo.com
client> 
client> 
server> HTTP/1.1 302 OK
server> Location: http://foo.com/new
server> Content-Length: 0
server>
server>
client> GET /new HTTP/1.1
client> Host: foo.com
client>
client>
```
The same should be true for non-GET requests that result in redirections (with exceptions listed here: [https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections)):

```http
client> POST /form HTTP/1.1
client> Host: foo.com
client> Content-Length: 3
client> Content-Type: application/x-www-form-urlencoded
client> 
client> a=b
server> HTTP/1.1 302 OK
server> Location: http://foo.com/new
server> Content-Length: 0
server>
server>
client> GET /new HTTP/1.1
client> Host: foo.com
client>
client>
```

Furthermore, the developer tools should display all the requests and responses that occurred (1 entry per request/response pair).

### Actual Behavior
Redirect handling for non-GET requests in scenarios where HTTP/HTTPS requests are intercepted using the `interceptHttpProtocol` and `interceptStreamProtocol` handlers are NOT following the same standards as non-intercepted requests. This is what happens, instead, with all 3xx responses:

```http
client> POST /form HTTP/1.1
client> Host: foo.com
client> Content-Length: 3
client> Content-Type: application/x-www-form-urlencoded
client> 
client> a=b
server> HTTP/1.1 302 OK
server> Location: http://foo.com/new
server> Content-Length: 0
server>
server>
client> POST /new HTTP/1.1
client> Host: foo.com
client> Content-Length: 3
client> Content-Type: application/x-www-form-urlencoded
client> 
client> a=b
```

This breaks interaction with a lot of websites; especially in scenarios where client authentication is being processed. For example, logging into Github.com successfully results in a 302 in scenarios where two factor authentication is required. The current implementation (without introducing some awful hacks) would break this interaction. 

Furthermore, it appears that 3xx responses are expected to be resolved prior to returning data to the interceptor callback. This creates an inconsistency in the Developer Tools network tab where 3xx responses automatically resolve to the final response code (i.e. 200) if the interaction resulted in a successful redirection. Meaning, the redirect request, response pair is lost in the log.

### To Reproduce

TBD

### Screenshots

The following screenshots demonstrate the issue using a login sequence on GitHub. After initial login request, the server redirects me to the 2FA form which is at `/sessions/two-factor`. The following is what happens when I don't implement redirection detection hacks:

Failed Login Without Redirection Hack

As can be seen, the redirect returns a 422 because the method and body remained unchanged during in the redirecting handling routine.

And the following demonstrates the proper result in the GUI with redirect detection hacks to modify the original request into a `GET` without a body, however, the developer tools network tab does not show an entry for the redirect/response pair to `/sessions/two-factor`:

Login With Redirection Hack but No Developer Tools Network Entry

As can be seen in the image above, the only request is to `/session` and the redirect gets handled internally and resolved to a 200 once the redirect has completed successfully.

### Additional Information

TBD

dsanders11:
#### Description of Change

Refs code in [`content/common/service_worker/service_worker_loader_helpers.cc`](https://source.chromium.org/chromium/chromium/src/+/5da0ed1e65305ab2c6d9de2bb4ce62f159520ba8:content/common/service_worker/service_worker_loader_helpers.cc;l=127-151).

Close #25894. It may fix it, but no reproduce case has been provided, so I can't definitively say that this fixes it.

I tried to add test coverage, but the test times out for some reason. I can see that it reaches the expected `callback(text)` after redirecting, but the promise from `executeJavaScript` never resolves. I've tried both `ajax` and `fetch` to no avail. The test was manually tested using that intercept code and acted as expected - `POST`s receiving a 302 or 303 redirects will be request the new location using `GET` rather than the current behavior of staying on `POST`. The test mentions matching Chromium's behavior because by-the-spec 302 is supposed to act like 307, but in Chromium (and I think every browser) it acts like 303 for backwards compatibility.

So the test failure seems to be a testing-only failure. Maybe someone else can spot an issue I'm not seeing.

cc @zcbenz

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed behavior of 302/303/307 redirect responses in the protocol module.

### DIFF
--- a/shell/browser/net/electron_url_loader_factory.cc
+++ b/shell/browser/net/electron_url_loader_factory.cc
@@ -17,6 +17,7 @@
 #include "mojo/public/cpp/system/string_data_source.h"
 #include "net/base/filename_util.h"
 #include "net/http/http_status_code.h"
+#include "net/url_request/redirect_util.h"
 #include "services/network/public/cpp/url_loader_completion_status.h"
 #include "services/network/public/mojom/url_loader_factory.mojom.h"
 #include "shell/browser/api/electron_api_session.h"
@@ -29,6 +30,7 @@
 #include "shell/common/gin_converters/gurl_converter.h"
 #include "shell/common/gin_converters/net_converter.h"
 #include "shell/common/gin_converters/value_converter.h"
+#include "third_party/blink/public/mojom/loader/resource_load_info.mojom-shared.h"
 
 #include "shell/common/node_includes.h"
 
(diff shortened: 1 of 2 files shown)


zcbenz (commented):
I'm not sure why `executeJavaScript` does not resolve, it is probably another bug we need to fix.

For the test itself, we can probably use `webRequest` module to get the results of the request instead.

dsanders11:
> For the test itself, we can probably use webRequest module to get the results of the request instead.

Do you have an example of what you're thinking? Not familiar with `webRequest` and the best way to use it in this situation.

zcbenz:
I'm thinking of something like this:

```js
interceptStringProtocol('http', (request, callback) => {...})

session.defaultSession.webRequest.onBeforeRedirect((request) => {
  expect(request.url).to.equal(...)
  expect(request.method).to.equal(...)
})

session.defaultSession.webRequest.onComplete((request) => {
  if (request.url === ...)
    done()
  else
    expect(request.url).to.equal(..)
})

after(() => session.defaultSession.webRequest.onBeforeRedirect(null))
after(() => session.defaultSession.webRequest.onComplete(null))

ajax('http://fake-host/redirect')
```

zcbenz (approved):

PR #26298, opened 2020-11-01
fix: recognize ?Cancel/?No as default cancel buttons

camilstaps:
#### Description of Change
This allows e.g. `?Cancel` to be recognized automatically as the cancel button.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/lib/browser/api/dialog.ts
+++ b/lib/browser/api/dialog.ts
@@ -186,7 +186,6 @@ const messageBox = (sync: boolean, window: BrowserWindow | null, options?: Messa
   const messageBoxType = messageBoxTypes.indexOf(type);
   if (messageBoxType === -1) throw new TypeError('Invalid message box type');
   if (!Array.isArray(buttons)) throw new TypeError('Buttons must be an array');
-  if (options.normalizeAccessKeys) buttons = buttons.map(normalizeAccessKey);
   if (typeof title !== 'string') throw new TypeError('Title must be a string');
   if (typeof noLink !== 'boolean') throw new TypeError('noLink must be a boolean');
   if (typeof message !== 'string') throw new TypeError('Message must be a string');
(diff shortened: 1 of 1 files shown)


ckerr (commented):
@camilstaps I'm not sure that I understand what bug this PR fixes. Could you give some more information on what the goal of this PR is? For example, is `?Cancel` a common use case that needs an easy path that can't be handled by `options.cancelId`?

camilstaps:
@ckerr thanks for the review!

I agree that 'bug' fix may be too harsh, but didn't want to call it a feature either.

The documentation for `cancelId` says:

> By default this is assigned to the first button with "cancel" or "no" as the label. If no such labeled buttons exist and this option is not set, 0 will be used as the return value.

It seems logical for me that if "Cancel" and "No" are recognized to determine the default, "&Cancel" and "&No" should be recognized as well. This is at least what I would expect reading the documentation, since I read it as if "cancel" and "no" refer to the label visible to the user, i.e., without access keys.

You asked:

> For example, is `?Cancel` a common use case that needs an easy path that can't be handled by `options.cancelId`?

Yes, I'd say that `?Cancel` (and `?No`) are common use cases. Of course they can be handled by `cancelId` as well, but the same goes for `Cancel` and `No`, for which the easy path is implemented.

I can imagine you're afraid of the added complexity(?). In that case maybe the documentation of `cancelId` can be updated to explicitly state that there should be no access keys for the default option resolution to work?

ckerr:
No it's not the complexity I'm concerned about -- I think I may be missing some context about the problem this PR solves, and that's what's confusing to me.

When would one want to pass `?Cancel` or `?No` as button text? That's not text that I can recall seeing on a button -- is there something else going on here, e.g. is the prepended `?` treated in a special way on some platform?

camilstaps:
Oh, I see the confusion now. `?` is used for access keys (see `normalizeAccessKeys`). It is not displayed, but the following character is used for a kind of shortcut key. `?Cancel` lets you cancel with Alt+C for example.

camilstaps:
@ckerr did I need to ping you? Apologies if you were already notified and it simply takes some time—that's fine!

zcbenz (changes requested):
Generally treating no/cancel as default cancel buttons has been proven a bad idea: it does not work for internationalization, it adds complexity for a very simple thing. 

We are not going to remove this feature, but we are also not going to add more complexity for it, so I'm closing this PR.

camilstaps:
@zcbenz OK, fair decision. Maybe it's a good idea to make this explicit in the documentation though.

PR #24832, opened 2020-08-04
ci: enable arm64 macOS tests

jkleinsc:
#### Description of Change

This PR enables testing of arm64 macOS builds.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).
- [x] This is **NOT A BREAKING CHANGE**. Breaking changes may not be merged to master until 11-x-y is branched.

#### Release Notes

Notes: none

PR #26313, opened 2020-11-03
fix: crash when loading wasm in mac arm64

deepak1556:
#### Description of Change

Instead of defaulting to always write protect, prefer v8 fix that does write and executable protection hop using background thread https://chromium-review.googlesource.com/c/v8/v8/+/2378307 .

Test: https://gist.github.com/deepak1556/c85a324822bf64ffeef450f3d1f33119
Refs https://github.com/microsoft/vscode/issues/106770#issuecomment-719877989

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: fix crash when loading wasm in mac arm64


jkleinsc (changes requested):
@deepak1556 in my work of getting CI testing working on a DTK, I found that this patch is not needed at all: https://github.com/electron/electron/pull/24832/

deepak1556:
@jkleinsc that would be ideal, is there a timeline for when the native compiled arm64 releases will be available ? The current cross compiled build is a no-go for vscode because of this wasm crash. Also while testing this fix realized that v8 made most of its fixes using `V8_HOST_ARCH_ARM64` so this change won't get picked up, really need native compiled bits.

jkleinsc:
@deepak1556 native arm64 releases are currently available for 11-x-y betas as well as on our nightly builds.

deepak1556:
Isn't it cross-compiled https://github.com/electron/electron/blob/master/.circleci/config.yml#L141 ?

jkleinsc:
@deepak1556 ah --I missed the cross-compiled bit of your comment.  Yes we cross-compile the arm64 binary on x64.  We won't have a native compiled arm64 release until our CircleCI provides arm64 macs and as of right now I don't have a date for that.

deepak1556:
@MarshallOfSound @jkleinsc can we remove the following patches from 11-x-y branch ? chrome arm64 beta (87.0.4280.67) is available without these, so I don't think we need them either ?

https://github.com/electron/electron/blob/11-x-y/patches/v8/chore_add_v8_apple_silicon_patches.patch
https://github.com/electron/electron/blob/11-x-y/patches/v8/fix_use_proper_page_size_for_mac_arm64.patch

jkleinsc:
@deepak1556 yes -- I'll put up a PR for that. #24832 is waiting until I have enough macs to run arm64 CI, but we can just address 11-x-y for now.

jkleinsc:
@deepak1556 PR is up here: #26650

PR #26314, opened 2020-11-03
chore: auto-label new issues based on template

dsanders11:
#### Description of Change

Auto-label issues based on the template used for creating it.

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -1,6 +1,7 @@
 ---
 name: Bug report
 about: Create a report to help us improve Electron
+labels: "bug :beetle:"
 
 ---
 
--- a/.github/ISSUE_TEMPLATE/Feature_request.md
+++ b/.github/ISSUE_TEMPLATE/Feature_request.md
@@ -1,6 +1,7 @@
 ---
 name: Feature request
 about: Suggest an idea for Electron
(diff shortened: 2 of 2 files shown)


05527223971:
Uueii3

MarshallOfSound (commented):
This is tricky, because in reality there is a solid quantity of issues that get raised with these templates that _aren't_ bugs or enhancement requests.  Until GitHub has strict template validation, I don't think we should be generous with auto-labels 😢

zcbenz:
When triaging issues, I think it is more helpful that new issues have no labels, otherwise it would be hard to tell if an issue has been triaged.

dsanders11:
> This is tricky, because in reality there is a solid quantity of issues that get raised with these templates that aren't bugs or enhancement requests.

That's a fair point.

> When triaging issues, I think it is more helpful that new issues have no labels, otherwise it would be hard to tell if an issue has been triaged.

That thought did occur to me, but it's also easy enough to add a label indicating it needs to be triaged, which would be removed on triage.

~10-12% of currently open issues are unlabeled, with an unknown percentage (hard to query) that only end up with a version label or the likes.

I think it might result in more issues being fully labeled if they were auto-labeled first, and the person triaging simply removed the auto-label if it wasn't applicable.

dsanders11:
Closing this PR since it's just outdated and hanging around, and looks like the newest version of the templates auto-label.

PR #26326, opened 2020-11-03
docs: update devtools extension tutorial

erickzhao:
#### Description of Change
With the switch to native extensions (see #21814), the set of `BrowserWindow` Extension APIs was deprecated in favour of their new `Session` counterparts.

This PR updates this tutorial to reflect that change, and improves a few other things (added links, clarifying notes, fixed copy errors).

cc @nornagon 

#### Checklist

#### Release Notes

Notes: none


docs/tutorial/devtools-extension.md
```diff
@@ -1,63 +1,73 @@
 # DevTools Extension
 
-Electron supports the [Chrome DevTools Extension][devtools-extension], which can
-be used to extend the ability of devtools for debugging popular web frameworks.
+Electron supports [Chrome DevTools extensions][devtools-extension], which can
+be used to extend the ability of Chrome's developer tools for debugging
+popular web frameworks.
 
-## How to load a DevTools Extension
+## Manually loading a DevTools extension
```
nornagon, line 7:
```suggestion
## How to load a DevTools extension
```

I don't think there's anything "manual" about this process, at least not more/less than it was before.

docs/tutorial/devtools-extension.md
```diff
@@ -1,63 +1,73 @@
...
 
-This document outlines the process for manually loading an extension.
-You may also try
-[electron-devtools-installer](https://github.com/GPMDP/electron-devtools-installer),
-a third-party tool that downloads extensions directly from the Chrome WebStore.
+To load an extension in Electron, you need to download it via Chrome,
+locate its filesystem path, and then load it into your [Session][session] by calling the
+[`ses.loadExtension`] API.
```
nornagon, line 11: It's possible to fetch the .crx with curl and then extract it, which is how I've done it. e.g. https://stackoverflow.com/a/14099762.

Though perhaps it's better to stick with recommending Chrome to download extensions, since the direct request method isn't documented/supported by google.

docs/tutorial/devtools-extension.md
```diff
@@ -1,63 +1,73 @@
...
+    const os = require('os')
+
+    // on macOS
+    const reactDevToolsPath = path.join(
+      os.homedir(),
+      '/Library/Application Support/Google/Chrome/Default/Extensions/fmkadmapgofadopljbjfkapdkoienihi/4.9.0_0'
+    )
+
+    app.whenReady().then(async () => {
+      await session.defaultSession.loadExtension(reactDevToolsPath)
```
nornagon, line 40: We should probably include the `allowFileAccess` option here once it lands: https://github.com/electron/electron/pull/25198

docs/tutorial/devtools-extension.md
```diff
@@ -1,63 +1,73 @@
...
+    })
    ```
 
-**Note:** The `BrowserWindow.addDevToolsExtension` API cannot be called before the
-ready event of the app module is emitted.
+**Notes:**
+
+* `loadExtension` returns a Promise with an [Extension object][extension-structure],
+which contains metadata about the extension that was loaded.
```
nornagon, line 47: We should probably note here that you have to await the promise before loading a page otherwise you're not guaranteed that the extension will be loaded when the page loads.

docs/tutorial/devtools-extension.md
```diff
@@ -1,63 +1,73 @@
...
+* `loadExtension` cannot be called before the `ready` event of the `app` module
+is emitted, nor can it be called on in-memory (non-persistent) sessions.
+* `loadExtension` must be called on every boot of your app if you want the
+extension to be loaded.
+
+### Removing a DevTools extension
+
+You can pass the extension's ID to the [`ses.removeExtension`][remove-extension] API to
+remove it from your Session.
```
nornagon, line 56:
```suggestion
remove it from your Session. Loaded extensions are not persisted between
app launches.
```

docs/tutorial/devtools-extension.md
```diff
@@ -1,63 +1,73 @@
...
 
-The extension will be remembered so you only need to call this API once per
-extension. If you try to add an extension that has already been loaded, this method
-will not return and instead log a warning to the console.
+## Using tooling
 
-### How to remove a DevTools Extension
+Alternatively, you can use third-party tooling to automate these steps for you.
+[electron-devtools-installer][electron-devtools-installer] is a popular NPM package
+that does just that.
```
nornagon, line 62: should we perhaps recommend this at the top of the page, as it seems like an easier approach for most?

(have we tested that it still works with the new api, also?)

docs/tutorial/devtools-extension.md
erickzhao: I wanted to contrast this method vs. the `electron-devtools-installer` way of doing things (both are H2-level headings now), so I figured this one was the more "manual" approach. Do you think that would make sense to readers? I'd like to add some differentiating descriptor here.

docs/tutorial/devtools-extension.md
erickzhao: Seems like new API support was added in https://github.com/MarshallOfSound/electron-devtools-installer/pull/131. I'm good with making this the first result!

docs/tutorial/devtools-extension.md
nornagon: Ah, ok, it makes sense if this section comes _after_ the "automatic" method.

docs/tutorial/devtools-extension.md
erickzhao: @nornagon I could set up a draft PR to be merged when the major this lands in becomes stable.

docs/tutorial/devtools-extension.md
erickzhao: >Though perhaps it's better to stick with recommending Chrome to download extensions, since the direct request method isn't documented/supported by google.
I'd lean towards this reasoning here!

zcbenz (approved):

Slapbox:
What about loading devtools extensions in packaged applications running on the `app://` protocol? It seems not to be possible to use React Devtools in a production mode with Electron over the `app://` protocol. Is this something that's not fixable?

PR #26331, opened 2020-11-04
fix: use the gin PageAllocator instead of V8::PageAllocator

MarshallOfSound:
This makes browser-process JS allocate pages using the base/gin allocator thus ensuring flags such as MAP_JIT are appropriately applied.

Without this changes apps were forced to codesign the browser process with `com.apple.security.cs.allow-unsigned-executable-memory` (which is a bad thing).

Notes: Updated internal memory allocation logic such that you no longer need to use the `com.apple.security.cs.allow-unsigned-executable-memory` codesign entitlement on macOS


deepak1556:
Is there path to upstream these patches ?

LGTM otherwise, nice find!

MarshallOfSound:
> Is there path to upstream these patches ?

The gin one, probably not.  The node one definitely

patches/chromium/export_gin_v8platform_pageallocator_for_usage_outside_of_the_gin.patch
```diff
@@ -0,0 +1,33 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <samuel.r.attard@gmail.com>
+Date: Tue, 3 Nov 2020 16:49:32 -0800
+Subject: export gin::V8Platform::PageAllocator for usage outside of the gin
+ platform
+
```
codebytere, line 6: Can you add an explanation here w/ upstream plans?

codebytere (approved):

patches/chromium/export_gin_v8platform_pageallocator_for_usage_outside_of_the_gin.patch
nornagon: lint failed here :( this needs a description @MarshallOfSound

MarshallOfSound: Oops, missed this comment.  How did it go green if lint was failing :/

nornagon: sorry, i meant lint failed to detect the missing description because of the two-line subject

PR #26335, opened 2020-11-04
docs: restructured docs directory

bandantonio:
#### Description of Change

First draft of the restructured docs directory

#### Checklist

#### Release Notes

Notes: first draft of the restructured docs directory


Kilian:
Nice! Here's my remarks, but this looks like a good step in the right direction so :+1: for me regardless of these remarks:

* I think it makes more sense to have the debugging section _before_ the distribution section.
* Given the state of Spectron and Devtron, should we still link them directly from this page?
* I think the security section, now that it is more a checklist like the "performance strategies" and less about the subsections linked here, might be better as "Security strategies" below "Application Architecture"
* I feel Accessibility is important enough to be higher up, perhaps directly below the "Application Architecture" section.

Again, just input, do with it as you please :)

Kilian (approved):
Nice! Here's my remarks, but this looks like a good step in the right direction so :+1: for me regardless of these remarks:

* I think it makes more sense to have the debugging section _before_ the distribution section.
* Given the state of Spectron and Devtron, should we still link them directly from this page?
* I think the security section, now that it is more a checklist like the "performance strategies" and less about the subsections linked here, might be better as "Security strategies" below "Application Architecture"
* I feel Accessibility is important enough to be higher up, perhaps directly below the "Application Architecture" section.

Again, just input, do with it as you please :)

docs/README.md
```diff
@@ -53,9 +46,20 @@ an issue:
...
+* [Boilerplates and CLIs](tutorial/boilerplates-and-clis.md)
+  * [Boilerplate vs CLI](tutorial/boilerplates-and-clis.md#boilerplate-vs-cli)
+  * [electron-forge](tutorial/boilerplates-and-clis.md#electron-forge)
+  * [electron-builder](tutorial/boilerplates-and-clis.md#electron-builder)
+  * [electron-react-boilerplate](tutorial/boilerplates-and-clis.md#electron-react-boilerplate)
+  * [Other Tools and Boilerplates](tutorial/boilerplates-and-clis.md#other-tools-and-boilerplates)
+
+### Advanced steps
+
+* [Application Architecture](tutorial/quick-start.md#application-architecture)
```
erickzhao, line 58: It might be a bit confusing to have `Application Architecture` linked twice. Maybe the solution is to rename `Application Architecture` the first time to `Electron's Process Model` or something, and remove the link in this header?

docs/README.md
```diff
@@ -55,10 +55,9 @@ an issue:
 
 ### Advanced steps
 
-* [Application Architecture](tutorial/quick-start.md#application-architecture)
-  * [Using Native Node.js Modules](tutorial/using-native-node-modules.md)
-  * [Performance Strategies](tutorial/performance.md)
-  * [Security Strategies](tutorial/security.md)
+* [Using Native Node.js Modules](tutorial/using-native-node-modules.md)
+* [Performance Strategies](tutorial/performance.md)
+* [Security Strategies](tutorial/security.md)
```
erickzhao, line 60: I think keeping Application Architecture as a top-level bullet point with no link would maybe make sense.

erickzhao (approved):

zcbenz (approved):

nornagon:
NB. docs changes shouldn't have release notes. I've updated the release-clerk message to prevent this from showing up when we generate release notes for the next release.

PR #26348, opened 2020-11-05
build: update dependencies

GChuf:
#### Description of Change

This PR updates/upgrades some dependencies without any breaking changes.
It also reduces the number of dependencies.

This PR sets the dugite version to exactly 1.87.0 (the one that was previously listed in yarn.lock) because using _^1.45.0_ now downloads a later version and produces some errors...
I did not test 1.88 for example.

Error logs for dugite v1.91.3 (latest):
https://app.circleci.com/pipelines/github/electron/electron/32311/workflows/c8049539-d88f-4f99-9928-c35ca2e0e4bc/jobs/711661

```
Error: Git could not be found at the expected path: '/home/builduser/project/src/electron/node_modules/dugite/git/bin/git'. This might be a problem with how the application is packaged, so confirm this folder hasn't been removed when packaging.
    at /home/builduser/project/src/electron/node_modules/dugite/build/lib/git-process.js:87:39
    at exithandler (child_process.js:324:5)
    at ChildProcess.errorhandler (child_process.js:336:5)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:275:12)
    at onErrorNT (internal/child_process.js:465:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
```

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


GChuf:
@dsanders11, can you check these two PRs?
The errors are just timeouts ... not sure what to do about that.
I can squash/change a few things if necessary.

dsanders11:
@GChuf, I'm not the person to review this, you'll need one of the maintainers to take a look.

> This PR sets the dugite version to exactly 1.87.0 (the one that was previously listed in `yarn.lock`) because using ^1.45.0 now downloads a later version and produces some errors...

Isn't that the point of `yarn.lock`? I'm not sure I understand what the problem was.

I'd focus on a single thing in the PR, if that's some `dugite` issue, then I'd focus on that. A big PR that changes a bunch of dependency versions doesn't seem like a good idea, IMO.

GChuf:
@dsanders11 thanks for your reply. 

regarding dugite: i'm not focusing on that, I can revert the changes. The point is that the latest dugite version produces some errors in tests - even though it still satisfies the "~1.45.0" requirement.

Who can I ask to take a look then - who are the maintainers?

dsanders11:
You don't need to ask anyone, PRs are reviewed by the maintainers, just be patient.

jkleinsc (changes requested):
@GChuf while the effort of this PR is appreciated, a PR that changes our dependencies and or our yarn.lock files is a security concern, eg see https://snyk.io/blog/why-npm-lockfiles-can-be-a-security-blindspot-for-injecting-malicious-modules/ and as such we typically only accept changes like this from known maintainers.  Please note that I'm not suggesting that you are trying to do something malicious here but rather that as a rule we generally reject these type of PRs.

GChuf:
@jkleinsc I see - thanks for the article. I understand your concerns.

The other PR is probably a bit easier to review, regarding yarn.lock file... this one's a bit big.
Just an idea: you could, for example, try recreating lock file from the package.json provided here, and compare the two versions . . .

Is there anything I can do to make this "easier" to review? Not updating dependencies can turn out to be a security issue itself after a while.

codebytere:
I agree with @jkleinsc above - while i also see your point that updating dependencies can turn out to be a security issue itself after a while, this type of contribution is not really one that can or should be done without deeper context for what the appropriate dependencies are to update to and why. We appreciate your efforts and would be happy to help guide you to other potential contributions based on issues or bugs on our issue tracker.

PR #26352, opened 2020-11-05
build: update spec dependencies

GChuf:
WIP

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/spec-main/package.json
+++ b/spec-main/package.json
@@ -16,6 +16,6 @@
   "dependencies": {
     "chai-as-promised": "^7.1.1",
     "dirty-chai": "^2.0.1",
-    "pdfjs-dist": "^2.2.228"
+    "pdfjs-dist": "~2.3.200"
   }
 }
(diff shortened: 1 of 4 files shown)


GChuf:
@dsanders11, can you check these two PRs?
The errors are just timeouts ... not sure what to do about that.
I can squash/change a few things if necessary.

jkleinsc (changes requested):
@GChuf while the effort of this PR is appreciated, a PR that changes our dependencies and or our yarn.lock files is a security concern, eg see https://snyk.io/blog/why-npm-lockfiles-can-be-a-security-blindspot-for-injecting-malicious-modules/ and as such we typically only accept changes like this from known maintainers.  Please note that I'm not suggesting that you are trying to do something malicious here but rather that as a rule we generally reject these type of PRs.

codebytere:
I agree with @jkleinsc above - while i also see your point that updating dependencies can turn out to be a security issue itself after a while, this type of contribution is not really one that can or should be done without deeper context for what the appropriate dependencies are to update to and why. We appreciate your efforts and would be happy to help guide you to other potential contributions based on issues or bugs on our issue tracker.

PR #26359, opened 2020-11-06
chore: improve docs relative link linting

dsanders11:
#### Description of Change

There's a wrinkle to the way anchors are generated for the website documentation it seems. While underscores are valid characters for a URL fragment, they're normally stripped *except* for when they're inside backticks.

An example from the documentation: <https://www.electronjs.org/docs/api/environment-variables#electron_force_window_menu_bar-linux>

That's a bit of a pain to try to represent with regex, the best I could do was multiple passes, first to separate out any backticked and not backticked sections, then another pass to allow underscores in one and not the other. I tried to keep it concise.

There was also weird indentation on that section (a single space) so it shows a couple lines changed which are only whitespace.

EDIT: Actually, now that I've done all that, I'm wondering if I got it inverted. Is it that `_Something_` has underscores stripped while they're normally not stripped? Or even the specific keywords like `_Linux_`, `_Deprecated_`, etc?

Maybe @MarshallOfSound knows more about the root behavior here. I'd imagine the "single source of truth" on this one is wherever the headers are turned into HTML IDs, but that could be down a rabbit hole of libraries.

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/script/check-relative-doc-links.py
+++ b/script/check-relative-doc-links.py
@@ -95,14 +95,27 @@ def getBrokenLinks(filepath):
 
 
 def checkSections(sections, lines):
-  sectionHeader = sections[1].replace('-', '')
+  invalidCharsRegex = '[^A-Za-z0-9_ \-]'
+  sectionHeader = sections[1]
   regexSectionTitle = re.compile('# (?P<header>.*)')
   for line in lines:
     matchHeader = regexSectionTitle.search(line)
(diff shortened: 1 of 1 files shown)


codebytere (approved):
lgtm as a stopgap (thanks!) but cc @MarshallOfSound in case you have more thoughts

Issue #26362, opened 2020-11-06
Call to BrowserWindow.center doesn't center the window on mac


CezaryKulakowski:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
12.0.0-nightly.20201104
* **Operating System:**
macOS 10.15.6
* **Last Known Working Electron version:**
I don't know, probably never.

### Expected Behavior
Window is placed on the center of the screen it was before call to BrowserWindow::center.

### Actual Behavior
Window is placed on the center of the screen horizontally but somewhat above center vertically.

### To Reproduce
Launch following application:
```
const { app, BrowserWindow } = require('electron')
  
function startTest() {
  let win = new BrowserWindow({
    show: true
  });
  
  win.loadURL('https://www.example.com');
  win.center();
}

app.on('ready', startTest)
```

CezaryKulakowski:
I've created pull request with fix for the issue: https://github.com/electron/electron/pull/26363

PR #26363, opened 2020-11-06
fix: center window properly after BrowserWindow::center is called

CezaryKulakowski:
This pr fixes https://github.com/electron/electron/issues/26362.

Call to BrowserWindow::center puts window somewhat above center vertically.
It's because Electron uses NSWindow::center:
https://developer.apple.com/documentation/appkit/nswindow/1419090-center
and according to documentation it doesn't center the window vertically
so we must do calculation and placement manually.
#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed call to BrowserWindow::center on mac.

### DIFF
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -38,6 +38,8 @@
 #include "shell/common/process_util.h"
 #include "skia/ext/skia_utils_mac.h"
 #include "third_party/webrtc/modules/desktop_capture/mac/window_list_utils.h"
+#include "ui/display/display.h"
+#include "ui/display/screen.h"
 #include "ui/gfx/skia_util.h"
 #include "ui/gl/gpu_switching_manager.h"
 #include "ui/views/background.h"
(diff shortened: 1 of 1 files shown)


MarshallOfSound (commented):
A case could be made here that this is expected behavior, and if you want to center the window _exactly_ then you can do exactly what this code does using the `screen` and `setPosition` API

codebytere:
@MarshallOfSound this is intended to https://github.com/electron/electron/issues/26362 it looks like. @CezaryKulakowski it's important when opening pull requests to indicate clearly which problem a change is intending to solve in detail and also to link to the relevant issue - without that it is hard for us to effectively review.

CezaryKulakowski:
@codebytere Is there a way to link issue with pr? Or I just should add link to issue in the pr's description?

jkleinsc:
@CezaryKulakowski add a link to the issue in the PR's description.  Something like fixes #26362.

CezaryKulakowski:
Closing this pr as according to electron current behaviour is expected. I'm going to keep the fix on our side as I don't agree with it (center means center to me and not somewhat above center vertically).

Issue #26055, opened 2020-10-20
setImmediate crash


hxkuc:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 
8-,9-,10-
* **Operating System:**
  * 
windows
* **Last Known Working Electron version:**
  * 
no

### Expected Behavior

console error

### Actual Behavior

crash

### To Reproduce

example code 

```
setImmediate(function () {
  aaa
})
```
Expected Behavior:
aaa is not defined

Actual Behavior:
crashed

ckerr:
gist of @hxkuc's test case: https://gist.github.com/ckerr/a0bbfb980262a38d4f6a5957d09720cd

ckerr:
I'm not able to reproduce this with the gist above when testing with Electron v10.1.4 on Windows 10 and Linux: the script errors out with the parse error, but Electron does not itself crash.

If you're able to reproduce this crash, could you provide more information on the crash itself and how to reproduce it? For example a crash backtrace would be helpful, as would a test case that reliably crashes.

hxkuc:
> I'm not able to reproduce this with the gist above when testing with Electron v10.1.4 on Windows 10 and Linux: the script errors out with the parse error, but Electron does not itself crash.
> 
> If you're able to reproduce this crash, could you provide more information on the crash itself and how to reproduce it? For example a crash backtrace would be helpful, as would a test case that reliably crashes.

i use your example and it crashed
![image](https://user-images.githubusercontent.com/17510186/96694366-f9c1bb00-13ba-11eb-9424-daf205896045.png)

codebytere:
@hxkuc I also cannot reproduce this. Can you please provide specific version numbers you're seeing this with?

hxkuc:
i don't know why you can't reproduce, i think it is very easy to reproduce, just use `electron-quick-start` and `npm i electron@v10.1.4`
openDevtools in the console write this 
```
setImmediate(function () {aaa})
```

ckerr:
[...]

PR #26365, opened 2020-11-06
fix: renderer crash on setImmediate

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/26055.

Fixes an issue whereby a corrupted `async_hooks` stack would crash the renderer when throwing some errors in the renderer process. This happened because Node.js resets the async hook stack when exceptions are thrown inside their uncaught exception handler, and in the renderer process we were using Blink's which did not do this. This is fixed by adding our own uncaught message listener to the renderer isolate - this resets the async hook stack to mitigate the crash and better align with Node.js' behavior. Isolate message listeners are additive (you can add multiple) so this does not interfere with Blink's own handling.

cc @ckerr @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixes an issue whereby a corrupted `async_hooks` stack would crash the renderer when throwing some errors in the renderer process.

### DIFF
--- a/spec-main/fixtures/crash-cases/setimmediate-renderer-crash/index.js
+++ b/spec-main/fixtures/crash-cases/setimmediate-renderer-crash/index.js
@@ -1,0 +1,22 @@
+const { app, BrowserWindow } = require('electron');
+const path = require('path');
+
+app.whenReady().then(() => {
+  const win = new BrowserWindow({
+    show: false,
+    webPreferences: {
+      nodeIntegration: true,
+      preload: path.resolve(__dirname, 'preload.js')
+    }
+  });
+
+  win.loadURL('data:text/html,<body>');
+
+  win.webContents.on('render-process-gone', () => {
+    process.exit(1);
+  });
+
+  win.webContents.on('did-finish-load', () => {
+    setTimeout(() => app.quit());
+  });
+});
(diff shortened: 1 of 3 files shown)


46967357 (approved):

MarshallOfSound (commented):
I know we talked about this on Slack, but I think `crash-cases` won't work for us here.  `crash-cases` detect if the main process crashes, because this crash is in the renderer we'll need a different system 🤔

MarshallOfSound (approved):
nvm I see what you're doing now 👍

spec-main/fixtures/crash-cases/setimmediate-renderer-crash/index.js
```diff
@@ -0,0 +1,22 @@
...
+app.whenReady().then(() => {
+  const win = new BrowserWindow({
+    show: false,
+    webPreferences: {
+      nodeIntegration: true,
+      preload: path.resolve(__dirname, 'preload.js')
+    }
+  });
+
+  win.loadURL('data:text/html,<body>');
```
nornagon, line 13:
```suggestion
  win.loadURL('about:blank');
```

jkleinsc (approved):

PR #26390, opened 2020-11-09
chore: move upload-to-github to TS

MarshallOfSound:
Type safety is always nice

Notes: no-notes

### DIFF
--- a/script/lib/util.py
+++ b/script/lib/util.py
@@ -25,11 +25,13 @@ from lib.config import is_verbose_mode
 ELECTRON_DIR = os.path.abspath(
   os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
 )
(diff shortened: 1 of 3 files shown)


codebytere (approved):

jkleinsc (approved):

dsanders11:
@MarshallOfSound, just a note about `ts-node` and the version that Electron currently uses, which I stumbled upon over the weekend. It has caching in that version, but [caching was removed in newer versions](https://github.com/TypeStrong/ts-node/issues/672) as it was considered unsafe due to not fully handling dependencies. In particular the original issue mentioned Mocha tests passing even though the code they were testing was changed and shouldn't have passed.

So, something to consider. `ts-node` with caching may lead to weird behavior, and it may be hard to realize it's happening. It's quite slow without caching though, so a bit between a rock and a hard place.

PR #26415, opened 2020-11-09
chore: add patch description to gin allocator patch

MarshallOfSound:
Missed a comment asking for a patch description for this one

Notes: no-notes

### DIFF
--- a/patches/chromium/export_gin_v8platform_pageallocator_for_usage_outside_of_the_gin.patch
+++ b/patches/chromium/export_gin_v8platform_pageallocator_for_usage_outside_of_the_gin.patch
@@ -4,6 +4,10 @@ Date: Tue, 3 Nov 2020 16:49:32 -0800
 Subject: export gin::V8Platform::PageAllocator for usage outside of the gin
  platform
 
+In order for memory allocation in the main process node environment to be
+correctly tagged with MAP_JIT we need to use gins page allocator instead
+of the default V8 allocator.  This probably can't be usptreamed.
+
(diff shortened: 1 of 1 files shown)


patches/chromium/export_gin_v8platform_pageallocator_for_usage_outside_of_the_gin.patch
```diff
@@ -4,6 +4,10 @@ Date: Tue, 3 Nov 2020 16:49:32 -0800
 Subject: export gin::V8Platform::PageAllocator for usage outside of the gin
  platform
 
+In order for memory allocation in the main process node environment to be
+correctly tagged with MAP_JIT we need to use gins page allocator instead
+of the default V8 allocator.  This probably can't be usptreamed.
```
nornagon, line 9: Could we upstream something that converted the non-static method that already exists for this into a static method? Since it doesn't actually use any instance state.

Alternatively, can we call the instance method that already exists instead of adding this patch?

MarshallOfSound: @nornagon We can't call the instance method because we don't use the gin platform.  And constructing one has side effects we don't want.  Converting the member to be a static though seems something upstream might take reasonably 🤔

nornagon (approved):
lgtm but please look into alternatives that are upstreamable!

jkleinsc (approved):

jkleinsc:
merging as CI failure is known flake unrelated to this PR.

PR #26428, opened 2020-11-10
chore: cherry-pick daad98ce from chromium (#26419)

### LINKED ISSUE #26419: getSystemVersion returns wrong version for Big Sur
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 10.1.5
* **Operating System:**
  * macOS 10.16 
  * N/A

### Expected Behavior
On Big Sur, Apple changed the versioning from 10.x.y to 11.x.y
https://en.wikipedia.org/wiki/MacOS_Big_Sur

Expected to get 11.0.1 from calling  _getSystemVersion_ 

### Actual Behavior
Calling  _getSystemVersion_  returns 10.16

This is true for an Intel based machine,  I haven't tried it with Apple Silicon which requires electron v11.

### To Reproduce

On Big Sur system, run electron and call to  _getSystemVersion_  

### Screenshots
N/A

### Additional Information
N/A

ckerr:
#### Description of Change

Fixes #26419.

The "IsOS/IsAtLeastOS/IsAtMostOS" functions are driven from the Darwin
version number, which is always accurate. Adjust the
SysInfo::OperatingSystemVersionNumbers() function to return accurate
version numbers (or the best we can deduce) so that all version
numbers returned by helper functions are accurate and consistent.
    
Change-Id: I2ca92478d76bf572cc55875cba443f4978482d10

#### Checklist

- [ ] `npm test` passes

CC @electron/wg-releases 

#### Release Notes

Notes: Fixed value of `getSystemVersion()` on Big Sur

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -111,3 +111,4 @@ cherry-pick-88f263f401b4.patch
 cherry-pick-229fdaf8fc05.patch
 cherry-pick-1ed869ad4bb3.patch
 crashpad-initialize-logging.patch
+make_macos_os_version_numbers_consistent.patch
--- a/patches/chromium/make_macos_os_version_numbers_consistent.patch
+++ b/patches/chromium/make_macos_os_version_numbers_consistent.patch
@@ -1,0 +1,45 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Avi Drissman <avi@chromium.org>
+Date: Sat, 25 Jul 2020 17:38:19 +0000
+Subject: Make macOS OS version numbers consistent
+
+The "IsOS/IsAtLeastOS/IsAtMostOS" functions are driven from the Darwin
+version number, which is always accurate. Adjust the
+SysInfo::OperatingSystemVersionNumbers() function to return accurate
+version numbers (or the best we can deduce) so that all version
+numbers returned by helper functions are accurate and consistent.
+
+Bug: 1101439
+Bug: 1108832
+Change-Id: I2ca92478d76bf572cc55875cba443f4978482d10
+Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2315490
+Reviewed-by: Mark Mentovai <mark@chromium.org>
+Commit-Queue: Avi Drissman <avi@chromium.org>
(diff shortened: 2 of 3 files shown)


MarshallOfSound (commented):
This change vaguely makes sense to Chromium because all they care about is the relative different between "a number" and "the current version".  This falls apart as soon as Apple releases 10.17 (it would be defined here as 12.0.0 which would more than likely be wrong).

The things apple have done to fake 10.16 make sense in the big scheme of things and I think we should let apples default behavior apply for older releases.  Apple is correct that changing from 10.X to X.0 is a breaking change and has correctly made the OS itself backwards compatible.  Us forcefully working around that is a bad plan IMO.

Unless I'm missing the need to directly override Apples _expected behavior_ I don't think we should backport this.  Worth noting on master we build with a new enough SDK that we don't worry about this at all (again aligning with apples expected behavior)

ckerr:
> This change vaguely makes sense to Chromium because all they care about is the relative different between "a number" and "the current version". This falls apart as soon as Apple releases 10.17 (it would be defined here as 12.0.0 which would more than likely be wrong).

This same code is live in 11-x-y, as well as master, as well as upstream. Unless you're saying it's not an issue in 11 and up because they are built with a newer version of the SDK?

The point about not forcing a breaking change when apple has tried to make things backwards compatible makes sense; but, I'm confused about treating this differently on old versions of Electron vs 11 and up -- surely there will be apps running in all combinations of [ electron version x OS version ]? Would we not want `getSystemVersion()` to return a consistent value across our supported versions?

MarshallOfSound:
> Unless you're saying it's not an issue in 11 and up because they are built with a newer version of the SDK?

Correct, Electron 11 and up are built on Xcode 12.2 for both x64 and arm64 and therefore will report the "correct" version

> Would we not want getSystemVersion() to behave consistently across builds?

I don't think so, the whole point here is it matches the expectations of the developer at the time that app was built.  This is a bit iffy given Electron devs don't build Electron itself.  But it makes sense IMO that until Electron 11 it reports as 10.16 as that was the belief of developers up until Electron 11 (which is the first version to officially "support" Big Sur and Apple silicon which is where the 11.0 version comes in to play)

Issue #26423, opened 2020-11-10
Segfault when calling app.quit in 'did-navigate' event handler


JOT85:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v10.1.5
* **Operating System:**
  * Arch Linux
* **Last Known Working Electron version:**
  * Can't find a working version, I've tried v6.1.12, v9.3.4, v10.1.5 (from the Arch repo) and v11.0.0-beta.20 from github. Notably, the error for v11 was SIGTRAP.

### Expected Behavior
The window to open and, upen navigation, to immediately quit.

### Actual Behavior
`“electron test.js” terminated by signal SIGSEGV (Address boundary error)`

### To Reproduce
```js
const { app, BrowserWindow } = require('electron')

app.whenReady().then(() => {
    const win = new BrowserWindow()
    win.webContents.on("did-navigate", () => app.quit())
    win.loadURL("https://example.com/")
})
```
Then run `electron test.js`

ckerr:
gist version of @JOT85's repro:

https://gist.github.com/ckerr/d6c4cb50635d1d2dca939ce51df84b92

JOT85:
Thanks so much!

DonAtMabl:
I'm seeing segfaults from this when a user selets a menu item with a role of `quit` as well. I haven't simplified the problem down, yet.

codebytere:
This has been resolved in recent versions of Electron

PR #15541, opened 2018-11-02
fix: destroy WebContents synchronously on shutdown

zcbenz:
#### Description of Change

Destroying WebContents asynchronously on shutdown may delay the destroying task after the shutdown task, resulting in an assertion in `ResourceDispatcherHostImpl::OnShutdown`.

This PR destroys WebContents synchronously on shutdown, so we can fix #15133 without breaking https://github.com/electron/electron/issues/8930.

I have also added notes for `DestroyWebContents` on how to use it correctly.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

#### Release Notes

Notes: Fix crash on exit when using `BrowserView`.

Issue #15133, opened 2018-10-13
4.0 crashes on shutdown when using BrowserView


PalmerAL:
* Output of `node_modules/.bin/electron --version`:  4.0.0-beta.3
* Operating System (Platform and Version): macOS 10.13.5
* Output of `node_modules/.bin/electron --version` on last known working Electron version (if applicable): 3.0.3

**Expected Behavior**
Electron shouldn't crash.

**Actual behavior**
A crash occurs.

**To Reproduce**
* Make a copy of `electron-quick-start`, and add the following to `main.js`:
```
const BrowserView = require("electron").BrowserView;
app.on('ready', function () {
  new BrowserView({})
})
```
* Quit the app.

Crash report:
```

Crashed Thread:        0  CrBrowserMain  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [0]

VM Regions Near 0:
--> 
    __TEXT                 000000010283e000-000000010283f000 [    4K] r-x/rwx SM=COW  /Users/USER/Documents/*/Electron.app/Contents/MacOS/Electron

Thread 0 Crashed:: CrBrowserMain  Dispatch queue: com.apple.main-thread
0   com.github.Electron.framework 	0x00000001081f8d45 0x10740c000 + 14601541
1   com.github.Electron.framework 	0x00000001081f9cfe 0x10740c000 + 14605566
2   com.github.Electron.framework 	0x000000010919563b 0x10740c000 + 30971451
3   com.github.Electron.framework 	0x000000010919574e 0x10740c000 + 30971726
4   com.github.Electron.framework 	0x00000001090e9ef2 0x10740c000 + 30269170
5   com.github.Electron.framework 	0x00000001091c4e31 0x10740c000 + 31166001
6   com.github.Electron.framework 	0x00000001091c59e2 0x10740c000 + 31168994
7   com.github.Electron.framework 	0x00000001091cf98e 0x10740c000 + 31209870
8   com.github.Electron.framework 	0x000000010902f828 0x10740c000 + 29505576
9   com.github.Electron.framework 	0x000000010a45e731 0x10740c000 + 50669361
[...]

PR #26435, opened 2020-11-10
fix: crash when invoking app.quit()

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/26423.

Fixes a crash seen when calling `app.quit()` inside navigation events like `did-finish-load`, `did-fail-load`, etc. This was happening because we were quitting synchronously in the same tick before all observers had finished notifying, as confirmed by the fact that calling `setTimeout(() => { app.quit() })` also solved the problem. We fix this by making `api_web_contents_->DestroyWebContents` asynchronous in all cases _including_ shutdown.

Tested with https://gist.github.com/ckerr/d6c4cb50635d1d2dca939ce51df84b92

cc @MarshallOfSound @nornagon @jkleinsc 

<details>
<Summary>Stacktrace</Summary>

```
55429:1110/134149.999064:FATAL:web_contents_impl.cc(879)] Check failed: !observers_.is_notifying_observers(). 
0   Electron Framework                  0x00000001143d3769 base::debug::CollectStackTrace(void**, unsigned long) + 9
1   Electron Framework                  0x00000001142c7c63 base::debug::StackTrace::StackTrace() + 19
2   Electron Framework                  0x00000001142e8115 logging::LogMessage::~LogMessage() + 261
3   Electron Framework                  0x00000001142e8fde logging::LogMessage::~LogMessage() + 14
4   Electron Framework                  0x00000001133c0360 content::WebContentsImpl::~WebContentsImpl() + 352
5   Electron Framework                  0x00000001133c278e content::WebContentsImpl::~WebContentsImpl() + 14
6   Electron Framework                  0x000000010eb55396 electron::InspectableWebContents::~InspectableWebContents() + 262
7   Electron Framework                  0x000000010eb5560e electron::InspectableWebContents::~InspectableWebContents() + 14
8   Electron Framework                  0x000000010ea27651 electron::api::BrowserWindow::Cleanup() + 193
9   Electron Framework                  0x000000010ea285cb electron::api::BrowserWindow::OnWindowClosed() + 27
10  Electron Framework                  0x000000010eb18363 electron::NativeWindow::NotifyWindowClosed() + 323
11  Electron Framework                  0x000000010ebfe7ad -[ElectronNSWindowDelegate windowWillClose:] + 45
```
</details>

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixes a crash seen when calling `app.quit()` inside some navigation events.

### DIFF
--- a/shell/browser/browser.h
+++ b/shell/browser/browser.h
@@ -306,8 +306,7 @@ class Browser : public WindowListObserver {
   void SetSecureKeyboardEntryEnabled(bool enabled);
 #endif
 
-  bool is_shutting_down() const { return is_shutdown_; }
-  bool is_quiting() const { return is_quiting_; }
+  bool is_shutting_down() const { return is_shutting_down_; }
   bool is_ready() const { return is_ready_; }
   v8::Local<v8::Value> WhenReady(v8::Isolate* isolate);
 
@@ -324,7 +323,7 @@ class Browser : public WindowListObserver {
   // Send the before-quit message and start closing windows.
   bool HandleBeforeQuit();
 
-  bool is_quiting_ = false;
+  bool is_quitting_ = false;
 
  private:
   // WindowListObserver implementations:
@@ -344,7 +343,7 @@ class Browser : public WindowListObserver {
   bool is_ready_ = false;
 
   // The browser is being shutdown.
-  bool is_shutdown_ = false;
+  bool is_shutting_down_ = false;
 
   // Null until/unless the default main message loop is running.
   base::OnceClosure quit_main_message_loop_;
--- a/shell/browser/api/electron_api_browser_view.cc
+++ b/shell/browser/api/electron_api_browser_view.cc
@@ -92,10 +92,8 @@ BrowserView::BrowserView(gin::Arguments* args,
 
 BrowserView::~BrowserView() {
   if (api_web_contents_) {  // destroy() is called
-    // Destroy WebContents asynchronously unless app is shutting down,
-    // because destroy() might be called inside WebContents's event handler.
     api_web_contents_->RemoveObserver(this);
-    api_web_contents_->DestroyWebContents(!Browser::Get()->is_shutting_down());
+    api_web_contents_->DestroyWebContents(true /* async */);
   }
 }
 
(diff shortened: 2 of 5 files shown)


nornagon (changes requested):
isn't this what the `is_shutting_down` check is supposed to catch?

I'm not a fan of adding more PostTasks to shutdown. It's already an extremely complex and fragile process and this makes it worse.

codebytere:
@nornagon the way this was previously means that it would destroy synchronously if it _was_ shutting down - perhaps it should just unilaterally do so

shell/browser/browser.h
```diff
@@ -306,8 +306,7 @@ class Browser : public WindowListObserver {
   void SetSecureKeyboardEntryEnabled(bool enabled);
 #endif
 
-  bool is_shutting_down() const { return is_shutdown_; }
-  bool is_quiting() const { return is_quiting_; }
```
codebytere, line 310: unused

deepak1556 (changes requested):
This change brings back https://github.com/electron/electron/pull/15541 , please refer https://github.com/electron/electron/issues/15133#issuecomment-434966976 thread for context.

I wouldn't encourage this api usage pattern from users, we can instead document how `app.quit` behaves and why it is not safe to call from within webContent event handlers.

codebytere:
@deepak1556 fair - i'd like to somehow prevent a hard crash though

deepak1556:
The shutdown behavior in chromium changes over versions, there are hard expectations from the runtime for the tasks that can be posted during shutdown and we have aligned our api with it. At the same time, webContents observers also has expectations about synchronous destructions. The case in question should be treated as a bad `CHECK`, we can instead throw an error for such use cases and have documentation for better design.

nornagon:
Should `app.quit()` be deprecated and replaced with, e.g. `app.quitSoon()` that calls `PostTask()`?

zcbenz (changes requested):
This is not a problem of `app.quit`, the same crash happens if you call `win.close()` or `win.webContents.destroy()`.

We usually solve this kind of crashes by delaying the event, for example https://github.com/electron/electron/pull/3643. But I'm fine documenting the crash to let users avoid it themselves, since the behavior may change after every Chromium upgrade.

PR #26444, opened 2020-11-11
feat: add app.runningUnderRosettaTranslation to detect running under rosetta

MarshallOfSound:
As in title and notes, simple API, code is straight from apples docs.

Docs reference for rosetta detection: https://developer.apple.com/documentation/apple_silicon/about_the_rosetta_translation_environment?language=objc#3616845

Notes: Added new `app.runningUnderRosettaTranslation` property to detect when running under rosetta on Apple silicon

### DIFF
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -1490,7 +1490,7 @@ on the direction Electron is going with renderer process restarts and usage of
 native modules in the renderer process please check out this
 [Tracking Issue](https://github.com/electron/electron/issues/18397).
 
-### `app.isRunningUnderRosettaTranslation` _macOS_ _Readonly_
+### `app.runningUnderRosettaTranslation` _macOS_ _Readonly_
 
 A `Boolean` which when `true` indicates that the app is currently running
 under the [Rosetta Translator Environment](https://en.wikipedia.org/wiki/Rosetta_(software)).
--- a/docs/api/app.md
+++ b/docs/api/app.md
@@ -1489,3 +1489,12 @@ which native modules you can use in the renderer process.  For more information
 on the direction Electron is going with renderer process restarts and usage of
 native modules in the renderer process please check out this
 [Tracking Issue](https://github.com/electron/electron/issues/18397).
+
+### `app.isRunningUnderRosettaTranslation` _macOS_ _Readonly_
+
+A `Boolean` which when `true` indicates that the app is currently running
+under the [Rosetta Translator Environment](https://en.wikipedia.org/wiki/Rosetta_(software)).
+
+You can use this property to prompt users to download the arm64 version of
+your application when they are running the x64 version under Rosetta
+incorrectly.
--- a/shell/browser/api/electron_api_app.cc
+++ b/shell/browser/api/electron_api_app.cc
@@ -1666,7 +1666,7 @@ gin::ObjectTemplateBuilder App::GetObjectTemplateBuilder(v8::Isolate* isolate) {
 #endif
 #if defined(OS_MAC)
       .SetProperty("dock", &App::GetDockAPI)
-      .SetProperty("isRunningUnderRosettaTranslation",
+      .SetProperty("runningUnderRosettaTranslation",
                    &App::IsRunningUnderRosettaTranslation)
 #endif
       .SetProperty("userAgentFallback", &App::GetUserAgentFallback,
@@ -1674,7 +1674,7 @@ gin::ObjectTemplateBuilder App::GetObjectTemplateBuilder(v8::Isolate* isolate) {
       .SetMethod("enableSandbox", &App::EnableSandbox)
       .SetProperty("allowRendererProcessReuse",
                    &App::CanBrowserClientUseCustomSiteInstance,
-                   &App::SetBrowserClientCanUseCustomSiteInstance)
+                   &App::SetBrowserClientCanUseCustomSiteInstance);
 }
 
 const char* App::GetTypeName() {
--- a/shell/browser/api/electron_api_app.cc
+++ b/shell/browser/api/electron_api_app.cc
@@ -1666,13 +1666,15 @@ gin::ObjectTemplateBuilder App::GetObjectTemplateBuilder(v8::Isolate* isolate) {
 #endif
 #if defined(OS_MAC)
       .SetProperty("dock", &App::GetDockAPI)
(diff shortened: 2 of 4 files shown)


docs/api/app.md
```diff
@@ -1489,3 +1489,12 @@ which native modules you can use in the renderer process.  For more information
 on the direction Electron is going with renderer process restarts and usage of
 native modules in the renderer process please check out this
 [Tracking Issue](https://github.com/electron/electron/issues/18397).
+
+### `app.isRunningUnderRosettaTranslation` _macOS_ _Readonly_
```
poiru, line 1493: Shouldn't properties be named like `app.runningUnderRosettaTranslation`?

shell/browser/api/electron_api_app.cc
```diff
@@ -1666,13 +1666,15 @@ gin::ObjectTemplateBuilder App::GetObjectTemplateBuilder(v8::Isolate* isolate) {
...
+      .SetProperty("isRunningUnderRosettaTranslation",
+                   &App::IsRunningUnderRosettaTranslation)
 #endif
       .SetProperty("userAgentFallback", &App::GetUserAgentFallback,
                    &App::SetUserAgentFallback)
       .SetMethod("enableSandbox", &App::EnableSandbox)
       .SetProperty("allowRendererProcessReuse",
                    &App::CanBrowserClientUseCustomSiteInstance,
-                   &App::SetBrowserClientCanUseCustomSiteInstance);
+                   &App::SetBrowserClientCanUseCustomSiteInstance)
```
poiru, line 1677: Need the semicolon here!

shell/browser/api/electron_api_app_mac.mm
```diff
@@ -58,6 +58,15 @@
   [NSApp setActivationPolicy:activation_policy];
 }
 
+bool App::IsRunningUnderRosettaTranslation() {
+  int ret = 0;
+  size_t size = sizeof(ret);
+  if (sysctlbyname("sysctl.proc_translated", &ret, &size, NULL, 0) == -1) {
```
miniak, line 64: `NULL` -> `nullptr`?

shell/browser/api/electron_api_app_mac.mm
```diff
@@ -58,6 +58,15 @@
...
+    return false;
+  }
+  return ret == = 1;
```
nornagon, line 67: that's a lot of `=`

shell/browser/api/electron_api_app_mac.mm
MarshallOfSound: Sticking with apples sample code here, makes for easier grep / replace if their recommendation changes in the future

ckerr (commented):
I'm not able to come up with a 2nd use case so maybe this is moot -- but `runningUnderRosettaTranslation()` is such a single-purpose call API that I wonder if something like `String? emulatorEnvironment()` that returns `rosetta` here would be more reusable for some currently-unknown future use.

:+1: on including this, whatever its final form, in 11

shell/browser/api/electron_api_app.h
```diff
@@ -221,6 +221,7 @@ class App : public ElectronBrowserClient::Delegate,
   bool MoveToApplicationsFolder(gin_helper::ErrorThrower, gin::Arguments* args);
   bool IsInApplicationsFolder();
   v8::Local<v8::Value> GetDockAPI(v8::Isolate* isolate);
+  bool IsRunningUnderRosettaTranslation();
```
ckerr, line 224: should be const

shell/browser/api/electron_api_app_mac.mm
```diff
@@ -58,6 +58,15 @@
   [NSApp setActivationPolicy:activation_policy];
 }
 
+bool App::IsRunningUnderRosettaTranslation() {
+  int ret = 0;
```
ckerr, line 62: _(nit)_ unlike the sample Apple code, this function does not return `ret` so the variable name is weird here. Maybe rename to "current_value" or "cur"?

shell/browser/api/electron_api_app_mac.mm
ckerr: You're going to grep on NULL rather than sysctl.proc_translated? :smile:

This line will break grep anyway if you change `ret` to `cur` or `curval`

MarshallOfSound:
> I'm not able to come up with a 2nd use case so maybe this is moot -- but runningUnderRosettaTranslation() is such a single-purpose call API that I wonder if something like String? emulatorEnvironment() that returns rosetta here would be more reusable for some currently-unknown future use.

Without building this is a super expansible way I don't think we can cover all the possible emulation scenarios.  i.e. technically you could be under multiple layers of emulation (Rosetta --> Qemu for instance is how docker is going to work).  Given the very targeted and _current_ use case for this I don't think making it partially generic gives us too much for the future.  Emulation is done differently on different platforms and each has it's own implications, i.e. I can see people being interested in a `process.arch` vs `process.archMyProcessIsBeingTreatedAs` but that isn't important for this rosetta use case where we just need to know _if_ rosetta is in play.  Even if rosetta isn't actually doing anything 🤷

ckerr (approved):

shell/browser/api/electron_api_app_mac.mm
```diff
@@ -58,6 +59,16 @@
   [NSApp setActivationPolicy:activation_policy];
 }
 
+bool App::IsRunningUnderRosettaTranslation() const {
+  int proc_translated = 0;
```
nornagon, line 63: Let's use Chrome's: https://source.chromium.org/chromium/chromium/src/+/master:base/mac/rosetta.mm;l=32;drc=ef4d087f2a9cbcc2c78c93dccc02695e4c2ed4ba

nornagon (commented):
api lgtm, but let's reuse chrome's impl.

shell/browser/api/electron_api_app_mac.mm
MarshallOfSound: @nornagon Chromes isn't available in 11-x-y.  I'll land this, clean backport and then refactor to use Chromes rosetta helpers on master.

nornagon: sounds good.

nornagon (approved):

PR #26449, opened 2020-11-11
doc: Can not set the BrowserWindow's webContents

xland:
#### Description of Change
One can not set the BrowserWindow's webContents when create a new BrowserWindow instance , so remove some lines of the example code.

#### Checklist

- [x] `npm test` passes


zcbenz:
The removed code is actually required for implementing standard `window.open` behavior, i.e. we have to create the window from the existing webContents.

Please create a bug report if you found the code not working for you.

I'm closing this PR since this is not a document error, and probably a bug.

PR #26461, opened 2020-11-12
Update dependencies

GChuf:
#### Description of Change

Updates extract-zip, pdfjs-dist and aws-sdk.

[Extract-zip version v1.4](https://github.com/maxogden/extract-zip/commits/master): fixed some bugs, better error handling, updated dependencies, comments, various checks.
[pdfjs-dist v2.3.200](https://github.com/mozilla/pdf.js/releases/tag/v2.3.200):  bugs/error fixes, perf improvements, dependency bumps
[aws-sdk@2.789.0](https://github.com/aws/aws-sdk-js/blob/master/CHANGELOG.md): new features, some bug fixes and performance improvements

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/npm/package.json
+++ b/npm/package.json
@@ -10,7 +10,7 @@
   "dependencies": {
     "@electron/get": "^1.0.1",
     "@types/node": "^14.6.2",
-    "extract-zip": "^1.0.3"
+    "extract-zip": "^1.4.1"
   },
   "engines": {
     "node": ">= 8.6"
--- a/package.json
+++ b/package.json
@@ -141,6 +141,6 @@
   },
   "dependencies": {
     "@types/temp": "^0.8.34",
-    "aws-sdk": "^2.727.1"
+    "aws-sdk": "^2.789.0"
   }
 }
--- a/spec-main/package.json
+++ b/spec-main/package.json
@@ -16,6 +16,6 @@
   "dependencies": {
     "chai-as-promised": "^7.1.1",
     "dirty-chai": "^2.0.1",
-    "pdfjs-dist": "^2.2.228"
+    "pdfjs-dist": "^2.3.200"
   }
 }
--- a/spec-main/yarn.lock
+++ b/spec-main/yarn.lock
@@ -205,10 +205,10 @@ path-to-regexp@^1.7.0:
   dependencies:
     isarray "0.0.1"
 
-pdfjs-dist@^2.2.228:
-  version "2.2.228"
-  resolved "https://registry.yarnpkg.com/pdfjs-dist/-/pdfjs-dist-2.2.228.tgz#777b068a0a16c96418433303807c183058b47aaa"
-  integrity sha512-W5LhYPMS2UKX0ELIa4u+CFCMoox5qQNQElt0bAK2mwz1V8jZL0rvLao+0tBujce84PK6PvWG36Nwr7agCCWFGQ==
+pdfjs-dist@^2.3.200:
+  version "2.3.200"
+  resolved "https://registry.yarnpkg.com/pdfjs-dist/-/pdfjs-dist-2.3.200.tgz#2920dc28ae7f2b8313cd6d6223c97ae85ac07eac"
+  integrity sha512-+8wBjU5h8LPZOIvR9X2uCrp/8xWQG1DRDKMLg5lzGN1qyIAZlYUxA0KQyy12Nw5jN7ozulC6v97PMaDcLgAcFg==
   dependencies:
     node-ensure "^0.0.0"
     worker-loader "^2.0.0"
--- a/yarn.lock
+++ b/yarn.lock
@@ -1059,10 +1059,10 @@ atob@^2.1.1:
   resolved "https://registry.yarnpkg.com/atob/-/atob-2.1.2.tgz#6d9517eb9e030d2436666651e86bd9f6f13533c9"
   integrity sha512-Wm6ukoaOGJi/73p/cl2GvLjTI5JM1k/O14isD73YML8StrH/7/lRFgmg8nICZgD3bZZvjwCGxtMOD3wWNAu8cg==
 
-aws-sdk@^2.727.1:
-  version "2.727.1"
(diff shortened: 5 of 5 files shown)


GChuf:
@codebytere @jkleinsc would you take a look at this PR?
I've only updated 3 dependencies that I think have the most impact - you can read the reasoning above in the 'description of change'. You can also peruse the changes in the changelogs/github repos.

spec-main/package.json
```diff
@@ -16,6 +16,6 @@
   "dependencies": {
     "chai-as-promised": "^7.1.1",
     "dirty-chai": "^2.0.1",
-    "pdfjs-dist": "^2.2.228"
+    "pdfjs-dist": "^2.3.200"
```
nornagon, line 19: We only use pdfjs to read the number of pages in a pdf generated by a test, so I'm not sure updating this is really useful.

Besides that, even if we do update, we should probably update to latest (2.5.207).

package.json
```diff
@@ -141,6 +141,6 @@
   },
   "dependencies": {
     "@types/temp": "^0.8.34",
-    "aws-sdk": "^2.727.1"
+    "aws-sdk": "^2.789.0"
```
nornagon, line 144: hm, these should probably be in devDependencies.

npm/package.json
```diff
@@ -10,7 +10,7 @@
   "dependencies": {
     "@electron/get": "^1.0.1",
     "@types/node": "^14.6.2",
-    "extract-zip": "^1.0.3"
+    "extract-zip": "^1.4.1"
```
nornagon, line 13: extract-zip@latest is 2.0.1. @malept is there any reason we shouldn't update to extract-zip@latest?

nornagon (changes requested):

npm/package.json
malept: `extract-zip@^2.0.0` requires Node 10, if we want to do that for Electron 12.

nornagon: Hm, seems friendlier to support more node versions if possible. In that case, we should perhaps go for extract-zip@1.7.0

malept: :man_shrugging: I don't know what sort of guarantees the Electron NPM module itself has, but Ecosystem WG modules generally only support Node versions that the Node.js folks actually support.

I'd also like to point out that extract-zip v2 has fewer transitive dependencies (which is why it requires a specific minor version of Node 10).

nornagon: good q. @electron/wg-ecosystem could maybe weigh in here.

malept: I think Releases/Upgrades owns the `electron` NPM module, I was just pointing out what Ecosystem WG's similar policy has been.

erickzhao: I don't have any context personally, and from what I can tell, our minimum Node version was last explicitly upgraded to Node 8 in Electron 7 so we could use `@electron/get` (ref https://github.com/electron/electron/pull/18413).

It makes sense to me to bump the minimum version of Node, though, especially since Node 8 reached EOL in 2019. Where do we document our minimum Node version for `electron`? We don't have an `engines` property in the `package.json`.

malept: > Where do we document our minimum Node version for `electron`? We don't have an `engines` property in the `package.json`.

https://github.com/electron/electron/blob/434f7749db9d3881695740d2a18fb1f4e652d0cc/npm/package.json#L15-L17

GChuf:
@malept @nornagon @erickzhao 
I should have mentioned the reasoning behind choosing these particular versions.

extract-zipv1.7 is the latest one that supports node8 I think, but given that v8 reached EOL, I'd say bumping to v10 is a very good idea.
However!, when I tried v1.7 and v1.5, both failed some tests. So I've tried to find the latest version which passes all the tests.

pdfjs-dist: v2.4 also fails some tests.

About the minimum engine requirement - I also saw it under /npm folder, would it be a good idea to move it inside the top level package.json?

spec-main/package.json
GChuf: I see - I thought this was being used to display PDFs. I guess that happens inside chromium then?

nornagon: That's correct, Chromium uses pdfium to display pdf. pdfjs is not present in the released version of Electron.

nornagon:
@GChuf do you mean that those packages failed their own tests? Or that you ran some Electron tests and observed failures related to those packages?

malept:
> About the minimum engine requirement - I also saw it under /npm folder, would it be a good idea to move it inside the top level package.json?

Considering that you're changing three different `package.json` files, have you read through the [source code structure docs](https://www.electronjs.org/docs/development/source-code-directory-structure)?

GChuf:
@malept that explains it - I read it now.
@nornagon the failed tests i was referring to were CircleCI/appveyor tests. I don't have any of those tests saved, and I wasnt able to get an url for those failed tests.

For those who didn't see my previous PRs, they're [here](https://github.com/electron/electron/pull/26348/files) and [here](https://github.com/electron/electron/pull/26352/files). They were basically attempts to update as much as possible without any failed tests.

spec-main/package.json
ckerr: As this is in spec-main, seems simple enough to confirm: remove the dep and see if CI is still happy

nornagon: @ckerr we use the dep (a grep confirms this). It is only used in one test, to count pages in a printed PDF from the `printToPDF` API.

zcbenz:
> the failed tests i was referring to were CircleCI/appveyor tests. I don't have any of those tests saved, and I wasnt able to get an url for those failed tests.

Some Electron tests are flaky so the failing tests might not be related to updating the dependencies.

Your previous PRs were rejected because those dependencies should not be updated without knowing why and how they are used, updating a dependency might still break things even when all tests pass.

I think we still can not accept this PR because it requires us to verify it does not introduce side effects and there is really no good reason for us to do that. /cc @codebytere @jkleinsc

codebytere:
Closing in agreement with the above. @GChuf we appreciate your efforts here, but as noted in previous PRs there are many better places to contribute that we are more able to accept from external contributors.

package.json
GChuf: @nornagon should i create a PR to move this to devDep?
Would also updating it at the same time be OK or not?

Issue #26460, opened 2020-11-12
Display Refresh Rate


JamesHannon:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description

Would it be possible to add the display refresh rate to the display object?

### Proposed Solution

Adding Refresh rate to the display object https://www.electronjs.org/docs/api/structures/display

### Alternatives Considered

We did look at doing the FPS tracker described in a couple of places but doesn't work with windows that band over multiple displays.

Exec platform-specific executables/plist and parse response

codebytere:
The Display object's properties mirror those [in Chromium](https://source.chromium.org/chromium/chromium/src/+/master:ui/display/display.h;bpv=0) and they do not currently have a refresh rate. Should they add one we will update our properties accordingly, but for now this is not unfortunately something we're able to do.

JamesHannon:
It will indeed, many thanks

PR #26472, opened 2020-11-12
feat: add displayFrequency to Display object

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/26460.

Refs https://chromium-review.googlesource.com/c/chromium/src/+/1621328.

Adds `displayFrequency` to the `Display` object to allow getting information about the refresh rate.

cc @jkleinsc @MarshallOfSound @zcbenz 

 
#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `displayFrequency` to the `Display` object to allow getting information about the refresh rate on Windows.

### DIFF
--- a/docs/api/structures/display.md
+++ b/docs/api/structures/display.md
@@ -10,6 +10,7 @@
 * `colorSpace` String -  represent a color space (three-dimensional object which contains all realizable color combinations) for the purpose of color conversions
 * `colorDepth` Number - The number of bits per pixel.
 * `depthPerComponent` Number - The number of bits per color component.
+* `displayFrequency` Number - The display refresh rate.
 * `bounds` [Rectangle](rectangle.md)
 * `size` [Size](size.md)
 * `workArea` [Rectangle](rectangle.md)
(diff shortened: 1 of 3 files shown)


docs/api/structures/display.md
```diff
@@ -10,6 +10,7 @@
 * `colorSpace` String -  represent a color space (three-dimensional object which contains all realizable color combinations) for the purpose of color conversions
 * `colorDepth` Number - The number of bits per pixel.
 * `depthPerComponent` Number - The number of bits per color component.
+* `displayFrequency` Number _Windows_ - The display refresh rate.
```
nornagon, line 13: It looks like this might be supported on Linux, and possibly on Mac as well? 
https://source.chromium.org/chromium/chromium/src/+/master:ui/base/x/x11_display_util.cc;l=313;drc=3980dd6e559b6d986dd29439ebbe5c9e6b7cb700
https://source.chromium.org/chromium/chromium/src/+/master:ui/display/mac/screen_mac.mm;l=139;drc=d29aea5fe2e3cafa6a84fda198d2e4c4ac5b1832

nornagon (changes requested):

docs/api/structures/display.md
codebytere: oh looks like you're right! the OG CL said windows-only but it seems they've updated since then.

zcbenz (approved):

nornagon (approved):

codebytere:
Failure is:
> Remove-Item : Cannot remove item C:\Users\joklein\AppData\Roaming\Electron Test 

and is not relevant.

PR #1430, opened 2015-04-16
Electron icon

simurai:
This PR is on top of https://github.com/atom/atom-shell/pull/1389 and replaces the Mac and Win icons.

I probably tweak it a bit more at some point. But in case you need it already now, here a first version.

screen shot 2015-04-17 at 12 21 34 am

I guess the icon files itself should also be renamed? Maybe `atom.icns` -> `icon.icns`. Then it's just generic and not project related.

PR #26477, opened 2020-11-12
chore: cherry-pick c244270e23 from chromium.

ppontes:
Ignore RenderFrameHostImpl::Detach() for speculative RFHs.

Currently, this all happens to work by chance, because the speculative
RFH or the entire FTN happens to be torn down before the browser process
ever processes a Detach() IPC for a speculative RFH.

However, there are a number of followup CLs that restructure how
provisional RenderFrames are managed and owned in the renderer process.
To simplify those CLs, explicitly branch in Detach() based on whether or
not the RFH is speculative. In the future, additional logic may be added
to the speculative branch (e.g. cancelling the navigation, if
appropriate).

(cherry picked from commit cf054220a2e1570a9149220494de8826c2e9d4db)

Bug: 1146709
Change-Id: I6490a90f7b447422d698676665b52f6f3a6f8ffd
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2524280
Commit-Queue: Daniel Cheng <dcheng@chromium.org>
Reviewed-by: Nasko Oskov <nasko@chromium.org>
Cr-Original-Commit-Position: refs/heads/master@{#825903}
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2530189
Reviewed-by: Adrian Taylor <adetaylor@chromium.org>
Cr-Commit-Position: refs/branch-heads/4240@{#1430}
Cr-Branched-From: f297677702651916bbf65e59c0d4bbd4ce57d1ee-refs/heads/master@{#800218}

Notes: Backported the fix to CVE-2020-16017: Browser UAF when detaching a provisional frame.

### DIFF
--- a/patches/chromium/ignore_renderframehostimpl_detach_for_speculative_rfhs.patch
+++ b/patches/chromium/ignore_renderframehostimpl_detach_for_speculative_rfhs.patch
@@ -1,0 +1,162 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Daniel Cheng <dcheng@chromium.org>
+Date: Wed, 11 Nov 2020 00:54:41 +0000
+Subject: Ignore RenderFrameHostImpl::Detach() for speculative RFHs.
+
+Currently, this all happens to work by chance, because the speculative
+RFH or the entire FTN happens to be torn down before the browser process
+ever processes a Detach() IPC for a speculative RFH.
+
+However, there are a number of followup CLs that restructure how
+provisional RenderFrames are managed and owned in the renderer process.
+To simplify those CLs, explicitly branch in Detach() based on whether or
+not the RFH is speculative. In the future, additional logic may be added
+to the speculative branch (e.g. cancelling the navigation, if
+appropriate).
+
+(cherry picked from commit cf054220a2e1570a9149220494de8826c2e9d4db)
+
+Bug: 1146709
+Change-Id: I6490a90f7b447422d698676665b52f6f3a6f8ffd
+Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2524280
+Commit-Queue: Daniel Cheng <dcheng@chromium.org>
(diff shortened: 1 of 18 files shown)


nornagon (approved):

patches/chromium/ignore_renderframehostimpl_detach_for_speculative_rfhs.patch
```diff
@@ -0,0 +1,132 @@
...
+diff --git a/content/browser/frame_host/render_frame_host_impl.cc b/content/browser/frame_host/render_frame_host_impl.cc
+index 8a3afc59f96e0f29997d0e239726217d490189d8..0c71cf19a2d8587e3e341d963da72c03a092b453 100644
+--- a/content/browser/frame_host/render_frame_host_impl.cc
++++ b/content/browser/frame_host/render_frame_host_impl.cc
+@@ -2403,6 +2403,9 @@ void RenderFrameHostImpl::UpdateRenderProcessHostFramePriorities() {
+ }
+ 
+ void RenderFrameHostImpl::OnDetach() {
++  if (frame_tree_node_->render_manager()->speculative_frame_host() == this)
++    return;
```
nornagon, line 39: This looks different to the upstream patch. It might be worth confirming that the issue still repros on 9.x? or otherwise checking quickly with the upstream patch author to see if this is the appropriate change for this version.

ppontes: > This looks different to the upstream patch. It might be worth confirming that the issue still repros on 9.x? or otherwise checking quickly with the upstream patch author to see if this is the appropriate change for this version.

The vulnerable code seems to be the same, so I expect the vulnerability to be there. The original fix, however, relies on a facility that doesn't exist in M83 or earlier. Great point about checking the adapted fix with the upstream patch author. I just got confirmation from him that this is the correct way to go for M83. 👍

PR #26490, opened 2020-11-13
test: support for adding extra module paths

dsanders11:
#### Description of Change

I'd like to be able to use a different `MOCHA_REPORTER`, which is almost fully supported, but currently that reporter must be in `spec/node_modules` or `spec-main/node_modules`. This PR adds ~`MOCHA_MODULE_PATHS`~ `ELECTRON_EXTRA_MODULE_PATHS` for adding to the module paths so that reporters can be loaded from outside the tree. This seemed like the most concise and small way to get the desired functionality.

Could also be used to remove `mocha-appveyor-reporter` as a dependency and use this functionality to load it on AppVeyor.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/spec-main/index.js
+++ b/spec-main/index.js
@@ -4,6 +4,13 @@ const v8 = require('v8');
 
 Module.globalPaths.push(path.resolve(__dirname, '../spec/node_modules'));
 
+// Extra module paths which can be used to load Mocha reporters
+if (process.env.ELECTRON_TEST_EXTRA_MODULE_PATHS) {
+  for (const modulePath of process.env.ELECTRON_TEST_EXTRA_MODULE_PATHS.split(':')) {
+    Module.globalPaths.push(modulePath);
+  }
+}
+
 // We want to terminate on errors, not throw up a dialog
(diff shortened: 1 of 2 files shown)


spec-main/index.js
```diff
@@ -4,6 +4,13 @@ const v8 = require('v8');
 
 Module.globalPaths.push(path.resolve(__dirname, '../spec/node_modules'));
 
+// Extra module paths which can be used to load Mocha reporters
+if (process.env.MOCHA_MODULE_PATHS) {
```
nornagon, line 8: I wouldn't name this `MOCHA` as there's nothing really mocha-specific about this functionality. `TEST_EXTRA_MODULE_PATHS` maybe?

dsanders11: Sure. Was just sticking with the naming convention of the other environment variables, but you're right, it's more generic.

nornagon (approved):

codebytere (approved):

MarshallOfSound (approved):

Issue #26355, opened 2020-11-05
Cannot click a button in a BrowserView -- MouseEvents are captured by drag?


ccorcos:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** v11.0.0-beta.19
  * 
* **Operating System:** macOS 11.0.1
  * 
* **Last Known Working Electron version:**
  * 

I've created a minimal repro for you. Download the repo:

```sh
git clone git@github.com:ccorcos/electron-browser-view-drag.git
cd electron-browser-view-drag
git checkout side-browser-view
npm install
npm start
```
There's a the default index.html inside the window's webview, and what I did is add a `BrowserView` to render `search.html` and you'll notice that button is not clickable. However, you can drag the whole window when you click in there...

### Expected Behavior
Should be able to click the button inside a BrowserView.

poiru:
It seems like there has been a regression in Electron 11. The BrowserView seems to always be a draggable region regardless of `-webkit-app-region`. Electron 10 works fine. cc @codebytere

codebytere:
Gist version: https://gist.github.com/060ad478da5b628726da4b4ea94a16a6

codebytere:
@poiru i'm not sure what you mean and did not see drag behavior in frameless windows where none was enabled 🤔 Do you have a specific repro in mind? 

@ccorcos PR is up to address this one

PR #26496, opened 2020-11-14
fix: webContents interaction with draggable BrowserViews

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/26355.

Fixed an issue where some buttons were un-clickable in some BrowserViews with draggable regions enabled.

This was happening because the way we construct draggable regions in `BrowserViews` is different than what we do in `BrowserWindows` - we implement them in the former by creating a new draggable `NSView` and overlaying regions that are not draggable. To accomplish this, we need to `hitTest` the view, so we can perform the correct actions when a user mouses down and tries to drag. Chromium explicitly chooses to ignore  mouse events in `RenderWidgetHostViewCocoa` when another view has returned itself in the `hitTest`, which is the case here and the root of the bug.

To fix the issue resultant of the above, we explicitly need to tell Chromium that it should _not_ use its own mouse ignore logic if a `NSEventTypeLeftMouseDragged` or `NSEventTypeLeftMouseDown` event - it should first check to see if we've told it not to do that and then proceed accordingly.

Tested with https://gist.github.com/060ad478da5b628726da4b4ea94a16a6.

cc @MarshallOfSound @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where some buttons were un-clickable in some BrowserViews with draggable regions enabled.

### DIFF
--- a/patches/chromium/render_widget_host_view_mac.patch
+++ b/patches/chromium/render_widget_host_view_mac.patch
@@ -10,10 +10,10 @@ kinds of utility windows. Similarly for `disableAutoHideCursor`.
 Additionally, disables usage of some private APIs in MAS builds.
 
 diff --git a/content/app_shim_remote_cocoa/render_widget_host_view_cocoa.mm b/content/app_shim_remote_cocoa/render_widget_host_view_cocoa.mm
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Did some testing and could not find any problem with this approach.

jkleinsc (approved):

PR #26517, opened 2020-11-16
fix: setCertVerifyProc canceling unrelated requests

### LINKED ISSUE #25338: calling setCertificateVerifyProc cancels requests in an unrelated session with net::ERR_CERT_DATABASE_CHANGED
### Preflight Checklist

* [ x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [ x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [ x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  *  10.1.1
* **Operating System:**
  *  Windows 10
  *  6.1.10

### Expected Behavior

The request to function normally except the certificate handler works as intended.

### Actual Behavior

When sending concurrent requests at once, the net request throws an error with net::ERR_CERT_DATABASE_CHANGED.

### To Reproduce

session.setCertificateVerifyProc((r, cb) => cb(0));

// Send net requests as normal

Soon enough, the code will start producing the error shown above.

nornagon:
#### Description of Change
This changes the behavior of ses.setCertificateVerifyProc so that it no longer
resets connections and clears the certificate verifier cache.

The original reason for clearing the cache was that our tests were written in
such a way that caching the result of the computation would interfere with the
functionality under test. Unfortunately, the only way we have to clear the
cache also has the side effect of canceling all open TLS requests in all
sessions. In lieu of a way to have this method affect only the session in which
it's called, I think that a better behavior is to not reset the cache. The
expected use for `setCertificateVerifyProc` is to call it once at the beginning
of a session's lifetime, so I believe that the impact of this behavior change
on real apps will be minimal.

Fixes #25338.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed `ses.setCertificateVerifyProc` canceling requests in unrelated sessions.

### DIFF
--- a/docs/api/session.md
+++ b/docs/api/session.md
@@ -483,6 +483,8 @@ win.webContents.session.setCertificateVerifyProc((request, callback) => {
 })
 ```
 
+> **NOTE:** The result of this procedure is cached by the network service.
+
 #### `ses.setPermissionRequestHandler(handler)`
 
 * `handler` Function | null
--- a/shell/browser/api/electron_api_session.cc
+++ b/shell/browser/api/electron_api_session.cc
@@ -619,9 +619,6 @@ void Session::SetCertVerifyProc(v8::Local<v8::Value> val,
   content::BrowserContext::GetDefaultStoragePartition(browser_context_)
       ->GetNetworkContext()
       ->SetCertVerifierClient(std::move(cert_verifier_client_remote));
-
-  // This causes the cert verifier cache to be cleared.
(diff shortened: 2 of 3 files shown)


deepak1556 (approved):

nornagon:
ci failures unrelated

PR #26518, opened 2020-11-16
refactor: move GetPrinterList off WebContents

nornagon:
#### Description of Change
This method doesn't need anything related to WebContents, so move it to a
separate module. Ultimately I think the API should be moved as well, but this
PR leaves the API unchanged.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/electron_api_web_contents.h
+++ b/shell/browser/api/electron_api_web_contents.h
@@ -45,12 +45,7 @@
 #if BUILDFLAG(ENABLE_PRINTING)
 #include "chrome/browser/printing/print_view_manager_basic.h"
 #include "components/printing/common/print_messages.h"
-#include "printing/backend/print_backend.h"  // nogncheck
 #include "shell/browser/printing/print_preview_message_handler.h"
-
-#if defined(OS_WIN)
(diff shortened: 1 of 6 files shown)


deepak1556 (approved):

shell/browser/api/electron_api_web_contents.h
```diff
@@ -45,12 +45,7 @@
 #if BUILDFLAG(ENABLE_PRINTING)
 #include "chrome/browser/printing/print_view_manager_basic.h"
 #include "components/printing/common/print_messages.h"
-#include "printing/backend/print_backend.h"  // nogncheck
 #include "shell/browser/printing/print_preview_message_handler.h"
-
-#if defined(OS_WIN)
-#include "printing/backend/win_helper.h"
```
deepak1556, line 52: Seems like we are using some classes from this header on windows.

nornagon: yep, just wanted to know which ones, so i asked CI :)

MarshallOfSound (approved):

nornagon:
ci failures unrelated

PR #26526, opened 2020-11-17
docs: update breaking changes for Electron 11

VerteDinde:
#### Description of Change

This PR updates our Planned Breaking Changes document to include two new breaking changes introduced in Electron 11, including links to the relevant PRs.

~Note: I added a brief description for why the `desktopCapturer.getMediaSourceIdForWebContents()` API method was removed, but that may be an unnecessary amount of detail. If we want to keep the description down to a simple "This API has been removed", very happy to change it.~ Removed this method entirely 🙂 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/breaking-changes.md
+++ b/docs/breaking-changes.md
@@ -103,7 +103,22 @@ shell.trashItem(path).then(/* ... */)
 
 ## Planned Breaking API Changes (11.0)
 
-There are no breaking changes planned for 11.0.
+### Removed: `BrowserView.{fromId, fromWebContents, getAllViews}` and `id` property of `BrowserView`
+The experimental APIs `BrowserView.{fromId, fromWebContents, getAllViews}`
+have now been removed. Additionally, the `id` property of `BrowserView`
+has also been removed.
+
+For more detailed information, see [#23578](https://github.com/electron/electron/pull/23578).
+
+### Removed: `desktopCapturer.getMediaSourceIdForWebContents()`
+
+The `desktopCapturer` API method `getMediaSourceIdForWebContents` has been
+removed. The method required a WebContents id, but required fetching the id
+through the main process. Any renderer process that could send an IPC message
+could get a token for a video stream in any other renderer, opening up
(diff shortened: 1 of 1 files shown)


docs/breaking-changes.md
```diff
@@ -103,7 +103,22 @@ shell.trashItem(path).then(/* ... */)
...
 
-There are no breaking changes planned for 11.0.
+### Removed: `BrowserView.{fromId, fromWebContents, getAllViews}` and `id` property of `BrowserView`
+The experimental APIs `BrowserView.{fromId, fromWebContents, getAllViews}`
+have now been removed. Additionally, the `id` property of `BrowserView`
+has also been removed.
+
+For more detailed information, see [#23578](https://github.com/electron/electron/pull/23578).
+
+### Removed: `desktopCapturer.getMediaSourceIdForWebContents()`
```
nornagon, line 113: This API was both added and removed during the 11.x cycle, so its removal is not a breaking change.

docs/breaking-changes.md
```diff
@@ -103,7 +103,22 @@ shell.trashItem(path).then(/* ... */)
 
 ## Planned Breaking API Changes (11.0)
 
-There are no breaking changes planned for 11.0.
+### Removed: `BrowserView.{fromId, fromWebContents, getAllViews}` and `id` property of `BrowserView`
+The experimental APIs `BrowserView.{fromId, fromWebContents, getAllViews}`
+have now been removed. Additionally, the `id` property of `BrowserView`
+has also been removed.
```
nornagon, line 109: Perhaps also mention that the `destroy` method was removed. It was never documented but I've seen a few issues of folks running into its removal.

VerteDinde: Sounds good, thanks! Will mention `destroy` as well

codebytere (approved):

ckerr (approved):

jkleinsc (approved):

PR #26531, opened 2020-11-17
refactor: use getWebPreference() for all options

miniak:
#### Description of Change
Follow-up to #24517

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/browser/guest-view-manager.ts
+++ b/lib/browser/guest-view-manager.ts
@@ -276,7 +276,7 @@ const watchEmbedder = function (embedder: Electron.WebContents) {
 
 const isWebViewTagEnabledCache = new WeakMap();
 
-export const isWebViewTagEnabled = function (contents: Electron.WebContents) {
+const isWebViewTagEnabled = function (contents: Electron.WebContents) {
   if (!isWebViewTagEnabledCache.has(contents)) {
     const webPreferences = contents.getLastWebPreferences() || {};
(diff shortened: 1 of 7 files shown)


codebytere:
Andy's no longer at Slack @miniak fyi :)

lib/renderer/window-setup.ts
```diff
@@ -261,7 +261,7 @@ export const windowSetup = (
         url = resolveURL(url, location.href);
       }
       const guestId = ipcRendererInternal.sendSync(IPC_MESSAGES.GUEST_WINDOW_MANAGER_WINDOW_OPEN, url, toString(frameName), toString(features));
-      if (guestId != null) {
+      if (guestId) {
```
nornagon, line 264: Should this (and the others in this file) still be `!= null`? This will now match 0 or '' or '0', which it didn't before.

miniak: guestId is now going to be 0 instead of `null` when missing

miniak: actually, I've reverted this change and just set `guestId` to `null` instead

zcbenz (approved):
Can you rebase on master to fix the lint errors?

alexeykuzmin (approved):

codebytere (approved):

PR #26537, opened 2020-11-17
fix: crash on exit in Event destructor

nornagon:
#### Description of Change
e.g. in CI: https://ci.appveyor.com/project/electron-bot/electron-ia32-testing/builds/36345418

Haven't repro'd locally, so this fix is speculative.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/event.cc
+++ b/shell/browser/api/event.cc
@@ -21,11 +21,15 @@ Event::Event() {}
 Event::~Event() {
   if (callback_) {
     v8::Isolate* isolate = electron::JavascriptEnvironment::GetIsolate();
-    v8::HandleScope scope(isolate);
-    auto message = gin::DataObjectBuilder(isolate)
-                       .Set("error", "reply was never sent")
-                       .Build();
-    SendReply(isolate, message);
+    // If there's no current context, it means we're shutting down, so we don't
+    // need to send an event.
+    if (!isolate->GetCurrentContext().IsEmpty()) {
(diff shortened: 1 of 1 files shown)


shell/browser/api/event.cc
```diff
@@ -21,11 +21,15 @@ Event::Event() {}
...
-    v8::HandleScope scope(isolate);
-    auto message = gin::DataObjectBuilder(isolate)
-                       .Set("error", "reply was never sent")
-                       .Build();
-    SendReply(isolate, message);
+    // If there's no current context, it means we're shutting down, so we don't
+    // need to send an event.
+    if (!isolate->GetCurrentContext().IsEmpty()) {
+      v8::HandleScope scope(isolate);
+      auto message = gin::DataObjectBuilder(isolate)
```
codebytere, line 28: Could we more specifically check this? e.g with `Browser::Get()->is_shutting_down()`?

nornagon: That would be a less specific check IMO. The crash is that the current context is empty. This check directly corresponds to the condition that matters here.

The current context is cleared partway through shutting down (https://github.com/electron/electron/blob/023c89265ffa8bdb21fac89474e1441822202f39/shell/browser/javascript_environment.cc#L88). This will work before then, and not afterwards.

codebytere (approved):

zcbenz (approved):

PR #26546, opened 2020-11-17
refactor: remove path from nativeImage converter

codebytere:
#### Description of Change

This refactor pulls the `base::FilePath` converter option out of the `NativeImage` converter.  Our converter contract is that converters return false when they fail -  they shouldn't touch the exception state.

Placed which previously leveraged the `NativeImage` converter to allow either a path or an image object are now changed to take `v8::Local<v8::Value>`s in their arguments instead of `gin::Handle<NativeImage>`, and detect the type passed and invoke the appropriate method for acquiring the image depending on whether the user passed a bespoke `NativeImage` or a string.

This also allows for better error handling by allowing the callers to decide whether to throw errors into JavaScript.

cc @nornagon @jkleinsc @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/electron_api_base_window.cc
+++ b/shell/browser/api/electron_api_base_window.cc
@@ -101,8 +101,7 @@ BaseWindow::BaseWindow(v8::Isolate* isolate,
 #if defined(TOOLKIT_VIEWS)
   v8::Local<v8::Value> icon;
   if (options.Get(options::kIcon, &icon)) {
-    gin_helper::ErrorThrower thrower(isolate);
-    SetIcon(thrower, icon);
+    SetIcon(gin_helper::ErrorThrower(isolate), icon);
   }
 #endif
 }
@@ -1008,12 +1007,13 @@ void BaseWindow::SetIcon(gin_helper::ErrorThrower thrower,
     native_image = electron::api::NativeImage::CreateFromPath(thrower.isolate(),
                                                               icon_path);
     if (native_image->image().IsEmpty()) {
-      thrower.ThrowError("Failed to convert path to nativeImage");
+      thrower.ThrowError("Failed to load image from path '" +
+                         icon_path.value() + "'");
       return;
     }
   } else {
     if (!gin::ConvertFromV8(thrower.isolate(), icon, &native_image)) {
-      thrower.ThrowError("Failed to convert nativeImage");
+      thrower.ThrowTypeError("Argument must be a file path or a NativeImage");
       return;
     }
   }
--- a/shell/browser/api/electron_api_base_window.cc
+++ b/shell/browser/api/electron_api_base_window.cc
@@ -99,13 +99,10 @@ BaseWindow::BaseWindow(v8::Isolate* isolate,
   window_->AddObserver(this);
 
 #if defined(TOOLKIT_VIEWS)
-  {
-    v8::TryCatch try_catch(isolate);
-    gin::Handle<NativeImage> icon;
-    if (options.Get(options::kIcon, &icon) && !icon.IsEmpty())
-      SetIcon(icon);
-    if (try_catch.HasCaught())
-      LOG(ERROR) << "Failed to convert NativeImage";
+  v8::Local<v8::Value> icon;
+  if (options.Get(options::kIcon, &icon)) {
+    gin_helper::ErrorThrower thrower(isolate);
+    SetIcon(thrower, icon);
   }
 #endif
 }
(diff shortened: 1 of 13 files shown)


shell/browser/api/electron_api_base_window.cc
```diff
@@ -998,14 +995,31 @@ bool BaseWindow::SetThumbarButtons(gin_helper::Arguments* args) {
...
+  if (gin::ConvertFromV8(thrower.isolate(), icon, &icon_path)) {
+    native_image = electron::api::NativeImage::CreateFromPath(thrower.isolate(),
+                                                              icon_path);
+    if (native_image->image().IsEmpty()) {
+      thrower.ThrowError("Failed to convert path to nativeImage");
+      return;
+    }
+  } else {
+    if (!gin::ConvertFromV8(thrower.isolate(), icon, &native_image)) {
+      thrower.ThrowError("Failed to convert nativeImage");
```
nornagon, line 1011:
```suggestion
      thrower.ThrowTypeError("Argument must be a file path or a NativeImage");
```

shell/browser/api/electron_api_tray.cc
```diff
@@ -61,19 +62,19 @@ namespace api {
 
 gin::WrapperInfo Tray::kWrapperInfo = {gin::kEmbedderNativeGin};
 
-Tray::Tray(gin::Handle<NativeImage> image,
+Tray::Tray(v8::Local<v8::Value> image,
            base::Optional<UUID> guid,
            gin::Arguments* args)
```
nornagon, line 67: nit: given this only uses `args->isolate()`, there's no need for it to take `gin::Arguments*`. Just `v8::Isolate*` will do.

shell/browser/api/electron_api_tray.cc
```diff
@@ -186,23 +187,62 @@ bool Tray::IsDestroyed() {
...
+          gin::StringToV8(isolate, "Failed to convert path to nativeImage")));
+      return;
+    }
+  } else {
+    if (!gin::ConvertFromV8(isolate, image, &native_image)) {
+      isolate->ThrowException(v8::Exception::Error(
+          gin::StringToV8(isolate, "Failed to convert nativeImage")));
+      return;
+    }
+  }
```
nornagon, line 239: here too

shell/browser/api/electron_api_tray.cc
```diff
@@ -283,7 +323,19 @@ void Tray::DisplayBalloon(gin_helper::ErrorThrower thrower,
...
+    icon = electron::api::NativeImage::CreateFromPath(thrower.isolate(),
+                                                      icon_path);
+    if (icon->image().IsEmpty()) {
+      thrower.ThrowError("Failed to convert path to nativeImage");
+      return;
+    }
+  } else if (options.Has("icon") && !options.Get("icon", &icon)) {
+    thrower.ThrowError("Failed to convert nativeImage");
+    return;
+  }
```
nornagon, line 337: here too

nornagon (dismissed):

shell/common/gin_converters/image_converter.cc
```diff
@@ -27,9 +27,19 @@ bool Converter<gfx::Image>::FromV8(v8::Isolate* isolate,
...
   gin::Handle<electron::api::NativeImage> native_image;
-  if (!gin::ConvertFromV8(isolate, val, &native_image))
-    return false;
+  base::FilePath icon_path;
+  if (gin::ConvertFromV8(isolate, val, &icon_path)) {
+    native_image =
+        electron::api::NativeImage::CreateFromPath(isolate, icon_path);
+    if (native_image->image().IsEmpty())
+      return false;
+  } else {
```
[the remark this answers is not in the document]
codebytere: this is used in `app.dock.setIcon` which takes a nativeImage or path - this would be the only place this gets handled

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2885,8 +2886,7 @@ void WebContents::StartDrag(const gin_helper::Dictionary& item,
     base::CurrentThread::ScopedNestableTaskAllower allow;
     DragFileItems(files, icon->image(), web_contents()->GetNativeView());
   } else {
-    gin_helper::ErrorThrower(args->isolate())
-        .ThrowError("Must specify either 'file' or 'files' option");
+    thrower.ThrowError("Must specify either 'file' or 'files' option");
```
nornagon, line 2889: nit: I prefer the previous code of constructing the thrower inline. Throwers have no state and it's not used anywhere else in this function.

shell/common/api/electron_api_native_image.cc
```diff
@@ -587,24 +616,6 @@ bool Converter<electron::api::NativeImage*>::FromV8(
...
-          StringToV8(isolate, "Image could not be created from " + img_path)));
-      return false;
-    }
-
-    return true;
-  }
-
   // reinterpret_cast is safe here because NativeImage is the only subclass of
   // gin::Wrappable<NativeImage>.
   *out = static_cast<electron::api::NativeImage*>(
```
nornagon, line 621: I think this converter is no longer necessary, it's identical to the default `gin::Wrappable` converter

shell/common/gin_converters/image_converter.cc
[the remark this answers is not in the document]
nornagon: possibly would be better to change that method to take `v8::Local<v8::Value>` instead of `gfx::Image` and use the new `TryConvertNativeImage`? That way you'd get better (+ consistent) error messages from `app.dock.setIcon`

shell/common/api/electron_api_native_image.cc
codebytere: do we not still need to `static_cast` to `electron::api::NativeImage`?

nornagon: https://source.chromium.org/chromium/chromium/src/+/master:gin/wrappable.h;l=132;drc=e1c3f75decd5a1097da39dc64cba32b7d559fac1

nornagon: this converter doesn't seem to have been removed.

codebytere: whoops, thought i got it! fixed.

shell/browser/browser_mac.mm
```diff
@@ -456,12 +456,17 @@ void RemoveFromLoginItems() {
...
-  api::NativeImage* native_image = nullptr;
-  if (!api::NativeImage::TryConvertNativeImage(isolate, icon, &native_image))
-    return;
+  gfx::Image image;
+
+  if (!icon->IsNull()) {
+    api::NativeImage* native_image = nullptr;
+    if (api::NativeImage::TryConvertNativeImage(isolate, icon, &native_image)) {
+      image = native_image->image();
+    }
```
nornagon, line 465: Best to not do anything if `TryConvertNativeImage` returns false, as it will have thrown an exception.

```suggestion
    if (!api::NativeImage::TryConvertNativeImage(isolate, icon, &native_image))
      return;
    image = native_image->image();
```

jkleinsc (approved):

codebytere:
Failure on step:
> Delete user app data directories

In WOA CI is not germane to this PR and all tests passed there, so I'm proceeding to merge this.

Issue #26480, opened 2020-11-13
Your app includes a version of an SDK from Electron that violates the App Store Review Guidelines. 


gaodeng:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.

### Issue Details

* **Electron Version:** 
  *  11.0.0-beta.22

### Rejection Email

Your app includes a version of an SDK from Electron that violates the App Store Review Guidelines. The version of the Electron SDK you are using in your app attempts to hide the use of private APIs. This is a violation Section 2.5.1 of the App Store Review Guidelines.
```
• Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework/__CFCopySystemVersionDictionary
• Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework/__kCFSystemVersionBuildVersionKey
• Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework/__kCFSystemVersionProductNameKey
• Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework/__kCFSystemVersionProductVersionExtraKey
• Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework/_CTFontDescriptorIsSystemUIFont
• Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework/__CFCopyServerVersionDictionary
• Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework/__kCFSystemVersionProductVersionKey
```
Guideline 2.5.1: Apps may only use public APIs.

Next Steps

To resolve this issue, please either remove the Electron SDK entirely from your app, or revise your app to utilize a corrected version of the Electron SDK.

### Additional Information

We need to use electron 11 to support apple silicon

MarshallOfSound:
Confirmed with `nm`, looks like we're accidentally linking to some crashpad files in the MAS build

gaodeng:
This problem doesn't seem to be fully resolved. I just submitted it for review with the electron 11 release version and it was rejected again!
@MarshallOfSound  @ckerr 

```

Guideline 2.5.1 - Performance - Software Requirements


[...]

PR #26548, opened 2020-11-18
fix: use public APIs in place of private CTFontDescriptorIsSystemUIFont in ui/gfx

MarshallOfSound:
This is the final API apple is dinging us for in #26480

This API was introduced to fix a test failure on Big Sur, I built and ran the tests with my change on Big Sur and they all passed so it seems legit.  Left the old impl there for non-mas just in case I missed something, will attempt to upstream this but it's unlikely it'll be accepted.

Notes: Stopped using private API `CTFontDescriptorIsSystemUIFont` in MAS build


codebytere:
Can we upstream this somehow?

MarshallOfSound:
Unlikely, but I can try.  I think this is slightly slower, and Chromium doesn't care a whole lot about private APIs

codebytere (approved):

Issue #25348, opened 2020-09-07
.isMaximized() always false with transparent:true


barkermn01:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.
*I found one but it's been closed since 2017 athough updates are still happening on it either close this issue and reopen that or make this the new issue #9092* 

### Issue Details

* **Electron Version:**
  * v10.1.1
* **Operating System:**
  * Windows 10 18363
* **Last Known Working Electron version:**
  * unknown

### Expected Behavior
calling `isMaximized()` to respond if the window is maximized 

### Actual Behavior
`isMaximized()` always returns false when transparent is set to true on BrowserWindow

### To Reproduce
https://gist.github.com/293ffab750b019714f12d8e1f203e611

### Additional Information
#9092 is a closed issue with people still reporting this problems but since it's closed I don't think admins are looking hence creating this issue.

arevi:
This has been an ongoing issue for years, it's possible to track this locally in your application and bypass the "isMaximized" method. Not a good solution, but it works.

barkermn01:
> This has been an ongoing issue for years, it's possible to track this locally in your application and bypass the "isMaximized" method. Not a good solution, but it works.

this is not a correct method, because a Maximized window when dragged should normalize it's size to then be moved, so the local state will come out of sync.

PR #26586, opened 2020-11-19
fix: add check in IsMaximized for non-WS_THICKFRAME windows

mlaurencin:
#### Description of Change

Closes https://github.com/electron/electron/issues/25348

This PR allows for windows without the WS_THICKFRAME style to be properly checked when determining if they are maximized. Transparent windows fall under this category. Since windows without the WS_THICKFRAME style cannot be put in a maximized state, we simulate it by increasing the window to the size of the display. Previously, when a window without the WS_THICKFRAME style was "maximized" in this way, the IsMaximized check did not account for this workaround. Now IsMaximized checks for both cases involving WS_THICKFRAME. 
(An additional criteria commented for maximization was also removed, since the implementation was removed at some point before this PR). 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: None

### DIFF
--- a/shell/browser/native_window_views.cc
+++ b/shell/browser/native_window_views.cc
@@ -632,8 +632,6 @@ void NativeWindowViews::SetBounds(const gfx::Rect& bounds, bool animate) {
   }
 #endif
 
-  LOG(INFO) << " NativeWindowViews::SetBounds - bounds: " << bounds.ToString()
-            << " " << __LINE__;
   widget()->SetBounds(bounds);
 }
 
--- a/shell/browser/native_window_views.cc
+++ b/shell/browser/native_window_views.cc
@@ -530,6 +530,20 @@ void NativeWindowViews::Unmaximize() {
 }
 
 bool NativeWindowViews::IsMaximized() {
+  // For window without WS_THICKFRAME style, we can not call IsMaximized().
+  // This path will be used for transparent windows as well.
+
+#if defined(OS_WIN)
+  gfx::Rect current_bounds = GetBounds();
+  if (!(::GetWindowLong(GetAcceleratedWidget(), GWL_STYLE) & WS_THICKFRAME)) {
+    // Compare the size of the window with the size of the display
+    auto display =
+        display::Screen::GetScreen()->GetDisplayNearestPoint(GetPosition());
+    return ((display.work_area().width() == current_bounds.width()) &&
+            (display.work_area().height() == current_bounds.height()))
+  }
+#endif
+
   return widget()->IsMaximized();
 }
 
@@ -618,6 +632,8 @@ void NativeWindowViews::SetBounds(const gfx::Rect& bounds, bool animate) {
   }
 #endif
 
+  LOG(INFO) << " NativeWindowViews::SetBounds - bounds: " << bounds.ToString()
+            << " " << __LINE__;
   widget()->SetBounds(bounds);
 }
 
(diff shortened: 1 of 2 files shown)


codebytere (commented):
Great find! We can probably add a test for this [here](https://github.com/electron/electron/blob/7c2ac6b7c180d0ae248540e9bf564a8e6dd207a2/spec-main/api-browser-window-spec.ts#L1022) to create a transparent window, and check maximized state before & after calling `maximize()`

shell/browser/native_window_views.cc
```diff
@@ -530,6 +530,20 @@ void NativeWindowViews::Unmaximize() {
...
 bool NativeWindowViews::IsMaximized() {
+  // For window without WS_THICKFRAME style, we can not call IsMaximized().
+  // This path will be used for transparent windows as well.
+
+#if defined(OS_WIN)
+  gfx::Rect current_bounds = GetBounds();
+  if (!(::GetWindowLong(GetAcceleratedWidget(), GWL_STYLE) & WS_THICKFRAME)) {
+    // Compare the size of the window with the size of the display
+    auto display =
+        display::Screen::GetScreen()->GetDisplayNearestPoint(GetPosition());
```
zcbenz, line 541: Should use `GetDisplayNearestWindow` to get window's display directly.

mlaurencin: Ok, sounds good. I switched to using that method with this next commit. Thanks

jkleinsc: @mlaurencin I think @zcbenz was saying to use GetDisplayNearestWindow here but maybe I'm missing something?

mlaurencin: @jkleinsc thank you for catching that. I changed it in in another function instead of this one. It should be correctly updated in the next commit.

shell/browser/native_window_views.cc
```diff
@@ -530,6 +530,20 @@ void NativeWindowViews::Unmaximize() {
...
+    return ((display.work_area().width() == current_bounds.width()) &&
```
MarshallOfSound, line 542: The dimensions being correct just means the _size_ is ok.  To be considered maximized the position of the window must match the top left coordinate of the display.  Otherwise this will incorrectly match a perfectly sized window that is positioned strangely.

mlaurencin: Great catch! Will add to next commit. Thank you!

codebytere:
@mlaurencin looks like this accidentally caught a release commit :) rebasing against master should fix that

shell/browser/native_window_views.cc
```diff
@@ -530,6 +530,23 @@ void NativeWindowViews::Unmaximize() {
...
+  gfx::Rect current_bounds = GetBounds();
+  if (!(::GetWindowLong(GetAcceleratedWidget(), GWL_STYLE) & WS_THICKFRAME)) {
+    // Compare the size of the window with the size of the display
+    auto display = display::Screen::GetScreen()->GetDisplayNearestWindow(
+        GetNativeWindow());
+    // Maximized if the window is the size of the display and in the top left
+    // corner
+    return ((display.work_area().width() == current_bounds.width()) &&
+            (display.work_area().height() == current_bounds.height()) &&
+            (GetPosition() == gfx::Point(0, 0)));
```
zcbenz, line 546: This does not work for multi-monitors, where the position might not be `(0, 0)` when window is maximized in an external window.  I think you can just do `return GetBounds() == display.work_area()`, I have not tested it though.

mlaurencin: Thank you for catching that. I tested using `return GetBounds() == display.work_area()` on my two monitors and it seemed to function properly, so I replaced those lines in this next commit.

zcbenz (approved):

codebytere (approved):

zcbenz:
@mlaurencin Can you rebase the PR on master? It should be able to fix the failing mac CI.

PR #26595, opened 2020-11-19
chore: fixup patches linting

dsanders11:
#### Description of Change

I'm not sure this linter was actually running. It was looking for `config.json` in individual patch directories, which doesn't seem to be the case with the current layout.

Unfortunately all of the linters seem to suffer from silently failing. They all follow a pattern of "no news is good news", but sometimes "no news" means the linter didn't actually run, or do meaningful work. I've tried to somewhat curtail that in this PR by requiring files be worked on for `allOk` to be true.

This PR also changes up how descriptions are checked for on patches, to better handle multi-line subjects. This was brought up with a recent patch where it wasn't detected, see https://github.com/electron/electron/pull/26331#discussion_r520152954.

~There are currently three patches in the repo which don't have a description, according to the revamped linter:~

* ~patches\ReactiveObjC\build_conditionally_import_ext_headers_from_framework_or_from.patch~
* ~patches\chromium\feat_allow_embedders_to_add_observers_on_created_hunspell.patch~
* ~patches\node\fix_expose_tracing_agent_and_use_tracing_tracingcontroller_instead.patch~

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/script/lint.js
+++ b/script/lint.js
@@ -164,65 +164,65 @@ const LINTERS = [{
 }, {
   key: 'patches',
   roots: ['patches'],
-  test: () => true,
+  test: filename => filename.endsWith('.patch'),
   run: (opts, filenames) => {
     const patchesDir = path.resolve(__dirname, '../patches');
-    for (const patchTarget of fs.readdirSync(patchesDir)) {
-      const targetDir = path.resolve(patchesDir, patchTarget);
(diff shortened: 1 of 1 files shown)


dsanders11:
Rebased after #26671 was merged. Lint runs clean now.

zcbenz (approved):

PR #26639, opened 2020-11-22
docs:  notification documentation fix

scharpentier0:
#### Description of Change
The new Notification parameters were saying that the title parameter is optional ans is supposed to be in the optional object containing every other parameter when in truth, the observed behavior is that the title is a mandatory first argument.

#### Checklist


dhoom56 (approved):

jkleinsc:
This documentation change is incorrect.  If you are having issues using the Notification module, please file an issue.

PR #26650, opened 2020-11-23
build: remove no longer needed arm64 macos patches

jkleinsc:
#### Description of Change

As mentioned here https://github.com/electron/electron/pull/26313#issuecomment-730988565, these patches are no longer needed.  This patch is just targeting 11-x-y because chore_add_v8_apple_silicon_patches.patch doesn't exist on master and 
fix_use_proper_page_size_for_mac_arm64.patch gets removed in #24832
#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


deepak1556 (approved):
Also confirmed with upstream that the wasm patches are not a correct https://bugs.chromium.org/p/chromium/issues/detail?id=1150060#c8

codebytere (approved):

PR #26657, opened 2020-11-23
refactor: simplify child process tracking for app.getAppMetrics()

miniak:
#### Description of Change
There is no need to map the child process ID to PID

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/shell/browser/api/electron_api_app.cc
+++ b/shell/browser/api/electron_api_app.cc
@@ -595,7 +595,7 @@ App::App() {
       ->set_delegate(this);
   Browser::Get()->AddObserver(this);
 
-  base::ProcessId pid = base::GetCurrentProcId();
+  auto pid = content::ChildProcessHost::kInvalidUniqueID;
   auto process_metric = std::make_unique<electron::ProcessMetric>(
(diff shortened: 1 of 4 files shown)


shell/browser/api/electron_api_app.cc
```diff
@@ -595,11 +595,10 @@ App::App() {
       ->set_delegate(this);
   Browser::Get()->AddObserver(this);
 
-  base::ProcessId pid = base::GetCurrentProcId();
   auto process_metric = std::make_unique<electron::ProcessMetric>(
       content::PROCESS_TYPE_BROWSER, base::GetCurrentProcessHandle(),
       base::ProcessMetrics::CreateCurrentProcessMetrics());
-  app_metrics_[pid] = std::move(process_metric);
+  app_metrics_[0] = std::move(process_metric);
```
zcbenz, line 601: Though `0` is guaranteed not to be returned in `GetID`, using `content::ChildProcessHost::kInvalidUniqueID` would make the code more future proof.

zcbenz (approved):

PR #25836, opened 2020-10-08
fix: delay emitting powerMonitor events on windows

nornagon:
#### Description of Change
This delays emitting certain powerMonitor events on Windows to avoid entering V8 during a WndProc callback, which can cause crashes.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Fixed a rare crash on Windows that could occur when emitting certain powerMonitor events.

PR #26668, opened 2020-11-24
fix: delay emitting NotifyIcon events on Windows

erickzhao:
#### Description of Change
Following prior work done in #25836, this PR delays emitting Balloon events from the `Tray` module to avoid entering V8 during a WndProc callback.

##### Implementation Details

Since `TrayIcon` objects are deleted when the `Tray` gets garbage collected, we don't want the callback to be processed if the icon is gone. This PR uses `WeakPtr` references to achieve this end.

This also means that the affected events won't be emitted if the WndProc callback is entered, but the `TrayIcon` is destroyed. This behaviour seems in line with Electron's existing `Tray` API design, as a reference to the `Tray` needs to be maintained for its API to be used at all.

cc @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a rare crash on Windows that could occur when emitting certain `Tray` events.

### DIFF
--- a/shell/browser/ui/win/notify_icon.cc
+++ b/shell/browser/ui/win/notify_icon.cc
@@ -49,7 +49,11 @@ NotifyIcon::NotifyIcon(NotifyIconHost* host,
                        HWND window,
                        UINT message,
                        GUID guid)
-    : host_(host), icon_id_(id), window_(window), message_id_(message) {
+    : host_(host),
+      icon_id_(id),
+      window_(window),
+      message_id_(message),
+      weak_factory_(this) {
   guid_ = guid;
   is_using_guid_ = guid != GUID_DEFAULT;
   NOTIFYICONDATA icon_data;
--- a/shell/browser/ui/win/notify_icon.cc
+++ b/shell/browser/ui/win/notify_icon.cc
@@ -49,11 +49,7 @@ NotifyIcon::NotifyIcon(NotifyIconHost* host,
                        HWND window,
                        UINT message,
                        GUID guid)
-    : host_(host),
-      icon_id_(id),
-      window_(window),
-      message_id_(message),
-      weak_factory_(this) {
+    : host_(host), icon_id_(id), window_(window), message_id_(message) {
   guid_ = guid;
   is_using_guid_ = guid != GUID_DEFAULT;
   NOTIFYICONDATA icon_data;
--- a/shell/browser/ui/win/notify_icon.h
+++ b/shell/browser/ui/win/notify_icon.h
@@ -71,6 +71,8 @@ class NotifyIcon : public TrayIcon {
   void SetContextMenu(ElectronMenuModel* menu_model) override;
   gfx::Rect GetBounds() override;
 
+  base::WeakPtr<NotifyIcon> GetWeakPtr() { return weak_factory_.GetWeakPtr(); }
+
  private:
   void InitIconData(NOTIFYICONDATA* icon_data);
 
@@ -100,6 +102,8 @@ class NotifyIcon : public TrayIcon {
 
   // Context menu associated with this icon (if any).
   std::unique_ptr<views::MenuRunner> menu_runner_;
+
+  base::WeakPtrFactory<NotifyIcon> weak_factory_;
 
   DISALLOW_COPY_AND_ASSIGN(NotifyIcon);
 };
--- a/shell/browser/ui/win/notify_icon.h
+++ b/shell/browser/ui/win/notify_icon.h
@@ -103,7 +103,7 @@ class NotifyIcon : public TrayIcon {
   // Context menu associated with this icon (if any).
   std::unique_ptr<views::MenuRunner> menu_runner_;
 
-  base::WeakPtrFactory<NotifyIcon> weak_factory_;
+  base::WeakPtrFactory<NotifyIcon> weak_factory_{this};
 
   DISALLOW_COPY_AND_ASSIGN(NotifyIcon);
 };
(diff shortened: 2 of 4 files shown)


codebytere (approved):

shell/browser/ui/win/notify_icon_host.h
```diff
@@ -10,6 +10,7 @@
 #include <vector>
 
 #include "base/macros.h"
+#include "base/memory/weak_ptr.h"
```
nornagon, line 13: This `#include` should go in the cc file if it's not needed in the header

shell/browser/ui/win/notify_icon.h
```diff
@@ -101,6 +103,8 @@ class NotifyIcon : public TrayIcon {
   // Context menu associated with this icon (if any).
   std::unique_ptr<views::MenuRunner> menu_runner_;
 
+  base::WeakPtrFactory<NotifyIcon> weak_factory_;
```
nornagon, line 106:
```suggestion
  base::WeakPtrFactory<NotifyIcon> weak_factory_{this};
```

shell/browser/ui/win/notify_icon.cc
```diff
@@ -49,7 +49,11 @@ NotifyIcon::NotifyIcon(NotifyIconHost* host,
                        HWND window,
                        UINT message,
                        GUID guid)
-    : host_(host), icon_id_(id), window_(window), message_id_(message) {
+    : host_(host),
+      icon_id_(id),
+      window_(window),
+      message_id_(message),
+      weak_factory_(this) {
```
nornagon, line 56: The modern style for initializing the weak factory is to use a default-initializer (see below)

```suggestion
      message_id_(message) {
```

shell/browser/ui/win/notify_icon_host.cc
```diff
@@ -142,14 +143,14 @@ LRESULT CALLBACK NotifyIconHost::WndProc(HWND hwnd,
...
-    NotifyIcon* win_icon = NULL;
+    base::WeakPtr<NotifyIcon> win_icon = NULL;
 
     // Find the selected status icon.
     for (NotifyIcons::const_iterator i(notify_icons_.begin());
          i != notify_icons_.end(); ++i) {
       auto* current_win_icon = static_cast<NotifyIcon*>(*i);
       if (current_win_icon->icon_id() == wparam) {
-        win_icon = current_win_icon;
+        win_icon = current_win_icon->GetWeakPtr();
```
nornagon, line 153: keep this loop as it was before and call `->GetWeakPtr()` after finding the icon. That way we call `GetWeakPtr` only once instead of _n_ times.

shell/browser/ui/win/notify_icon_host.cc
```diff
@@ -162,15 +163,30 @@ LRESULT CALLBACK NotifyIconHost::WndProc(HWND hwnd,
 
     switch (lparam) {
       case NIN_BALLOONSHOW:
-        win_icon->NotifyBalloonShow();
+        content::GetUIThreadTaskRunner({})->PostTask(
+            FROM_HERE, base::BindOnce(
+                           [](base::WeakPtr<NotifyIcon> wicon) {
+                             wicon->NotifyBalloonShow();
+                           },
+                           win_icon));
```
nornagon, line 171: The code as is will fail because it doesn't check `if (wicon)`. I recommend using the member-function variant of `base::Bind` instead, which handles this automatically:

```suggestion
        content::GetUIThreadTaskRunner({})->PostTask(
            FROM_HERE, base::BindOnce(
                           &NotifyIcon::NotifyBalloonShow,
                           win_icon));
```

(see [base/callback.h:46-49](https://source.chromium.org/chromium/chromium/src/+/master:base/callback.h;l=46-49;drc=788a3897a44bde5263e83172c84ebb85d99b5080))

Same for the cases below.

nornagon (changes requested):

shell/browser/ui/win/notify_icon_host.cc
```diff
@@ -160,17 +162,35 @@ LRESULT CALLBACK NotifyIconHost::WndProc(HWND hwnd,
...
+    // that the callback is aware if the NotifyIcon gets
+    // garbage-collected. This occurs when the tray gets
+    // GC'd, and the BALLOON events below will not emit.
+    base::WeakPtr<NotifyIcon> win_icon_weak = win_icon->GetWeakPtr();
+
+    int modifiers = GetKeyboardModifers();
+    bool left_button_click =
+        (lparam == WM_LBUTTONDOWN || lparam == WM_LBUTTONDBLCLK);
+    bool double_button_click =
+        (lparam == WM_LBUTTONDBLCLK || lparam == WM_RBUTTONDBLCLK);
```
nornagon, line 175: these can stay inlined

shell/browser/ui/win/notify_icon_host.cc
```diff
@@ -169,10 +169,6 @@ LRESULT CALLBACK NotifyIconHost::WndProc(HWND hwnd,
     base::WeakPtr<NotifyIcon> win_icon_weak = win_icon->GetWeakPtr();
 
     int modifiers = GetKeyboardModifers();
```
nornagon, line 171: this too i think (sorry!)

shell/browser/ui/win/notify_icon_host.cc
```diff
@@ -34,7 +36,7 @@ bool IsWinPressed() {
          ((::GetKeyState(VK_RWIN) & 0x8000) == 0x8000);
 }
 
-int GetKeyboardModifers() {
+int GetKeyboardModifiers() {
```
erickzhao, line 39: Unrelated typo fix in my last commit.

nornagon (approved):

zcbenz (approved):

PR #26676, opened 2020-11-24
feat: allow passing URL filter to intercept protocol functions

tarruda:
#### Description of Change
The goal is to allow passing an array of URL patterns to all intercept*Protocol functions, similarly to how it is done in the `webRequest` module. This should make the intercept API more useful, since right now one cannot intercept https without handling all requests.

Note that unlike webRequest, the protocol API allows intercepting requests at a lower level, before web security is aware. So the only way an app can intercept an https URL without breaking web security is by using the protocol API.

The first commit is a refactor which moves webRequest code that parses/matches URL patterns into a separate module. The second modifies protocol API to allow an optional second parameter to intercept functions. If no argument is passed, then all URLs will be intercepted (which is how it works now).

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Modified protocol API to allow passing URL filters to intercept functions.

### DIFF
--- a/docs/api/protocol.md
+++ b/docs/api/protocol.md
@@ -247,12 +247,12 @@ Unregisters the custom protocol of `scheme`.
 
 Returns `Boolean` - Whether `scheme` is already registered.
 
-### `protocol.interceptFileProtocol(scheme[, filter], handler)`
+### `protocol.interceptFileProtocol(scheme[, options], handler)`
 
 * `scheme` String
-* `filter` Object (optional)
-  * `urls` String[] - Array of URL patterns that will be used to filter out the
-        requests that do not match the URL patterns.
+* `options` Object (optional)
+  * `urls` String[] (optional) - Array of URL patterns that will be used to
+        filter out the requests that do not match the URL patterns.
 * `handler` Function
   * `request` [ProtocolRequest](structures/protocol-request.md)
   * `callback` Function
@@ -263,12 +263,12 @@ Returns `Boolean` - Whether the protocol was successfully intercepted
 Intercepts `scheme` protocol and uses `handler` as the protocol's new handler
 which sends a file as a response.
 
-### `protocol.interceptStringProtocol(scheme[, filter], handler)`
+### `protocol.interceptStringProtocol(scheme[, options], handler)`
 
 * `scheme` String
-* `filter` Object (optional)
-  * `urls` String[] - Array of URL patterns that will be used to filter out the
-        requests that do not match the URL patterns.
+* `options` Object (optional)
+  * `urls` String[] (optional) - Array of URL patterns that will be used to
+        filter out the requests that do not match the URL patterns.
 * `handler` Function
   * `request` [ProtocolRequest](structures/protocol-request.md)
   * `callback` Function
@@ -279,12 +279,12 @@ Returns `Boolean` - Whether the protocol was successfully intercepted
 Intercepts `scheme` protocol and uses `handler` as the protocol's new handler
 which sends a `String` as a response.
 
-### `protocol.interceptBufferProtocol(scheme[, filter], handler)`
+### `protocol.interceptBufferProtocol(scheme[, options], handler)`
 
 * `scheme` String
-* `filter` Object (optional)
-  * `urls` String[] - Array of URL patterns that will be used to filter out the
-        requests that do not match the URL patterns.
+* `options` Object (optional)
+  * `urls` String[] (optional) - Array of URL patterns that will be used to
+        filter out the requests that do not match the URL patterns.
 * `handler` Function
   * `request` [ProtocolRequest](structures/protocol-request.md)
   * `callback` Function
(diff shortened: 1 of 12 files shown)


docs/api/protocol.md
```diff
@@ -247,65 +247,70 @@ Unregisters the custom protocol of `scheme`.
 
 Returns `Boolean` - Whether `scheme` is already registered.
 
-### `protocol.interceptFileProtocol(scheme, handler)`
+### `protocol.interceptFileProtocol(scheme, handler[, filter])`
```
nornagon, line 250: Hm, I don't like the pattern of adding options as extra arguments. In general I think that "options" objects are a better pattern—they're self-documenting in code that uses them (since the options are named), and they're more extensible and flexible, especially if options are later removed.

There are a few possibilities here I think:
1. `protocol.interceptFileProtocol(scheme, handler[, options])` where `options` is `{ filter?: string[] }`. It's a little awkward to put the options after the handler though, as if specified inline the handler will be several lines long and the options object will be hard to spot in code that uses this.
2. `protocol.interceptFileProtocol(scheme[, options], handler)` as above, but put the optional argument before the handler function. A bit harder to handle in C++ but better ergonomically for the user.
3. `protocol.interceptFileProtocol(options, handler)` where `options` is `string | { scheme: string, filter?: string[] }`. Passing the first argument as a string is equivalent to passing `{ scheme: arg }`. This is similar to Node's [`http.request`](https://nodejs.org/api/http.html#http_http_request_options_callback).

I think either (2) or (3) is better than (1), and any of these would be an improvement.

tarruda: Is it ok to break API compatibility? I added the filters as the final argument to avoid breaking any existing apps that use intercept protocol.

nornagon: (2) wouldn't break API compatibility if correctly handled. The 2nd argument would be _optional_ — if not supplied, the behavior would be the same as before.

tarruda: I've implemented suggestion (2)

tarruda: Note that in the docs I've used `filter` as the argument name, to match what is used in web-request docs. Let me know if I should use `options` instead.

nornagon: The reason I suggested `options` is because we might add new options in future which aren't related to filtering.

tarruda: Ok, I will adjust the name and type

tarruda:
I've made the requested changes and added tests:

```
ok 1036 protocol module protocol.intercept(Any)Protocol with filter intercepts if URL matches any of the URL patterns
ok 1037 protocol module protocol.intercept(Any)Protocol with filter intercepts any URL if empty filter array is passed
ok 1038 protocol module protocol.intercept(Any)Protocol with filter intercepts any URL if empty options is passed
ok 1039 protocol module protocol.intercept(Any)Protocol with filter does not intercept if URL does not matches any of the URL patterns
ok 1040 protocol module protocol.intercept(Any)Protocol with filter throws error if an invalid pattern is passed
```

It seems the CI is failing for another reason:

```
not ok 1165 session.serviceWorkers getFromVersionID() should report the correct script url and scope
  Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\projects\src\electron\spec-main\api-service-workers-spec.ts)
      at listOnTimeout (internal/timers.js:554:17)
      at processTimers (internal/timers.js:497:7)
```

nornagon:
Just checking, would this use case be addressed satisfactorily by https://github.com/electron/electron/pull/26065? Then the filter could be implemented in JS rather than just being a simple URL match

tarruda:
> Just checking, would this use case be addressed satisfactorily by #26065? Then the filter could be implemented in JS rather than just being a simple URL match

Yes, #26065 implements a more generic/flexible solution since one can choose to intercept by looking at other request information such as headers.

However, I imagine that the common use case is to decide which requests should be intercepted based on URL patterns, similarly to how it is done with the `webRequest` API. In this simpler use case, it would probably be more efficient to pass URL patterns to C++ and avoid unnecessary JS function invocations.

Let me give you my own use case as an example: Our app (a single JS file), is loaded by an index.html which is modified via the protocol module (this index.html does some extra bootstrapping, such as creating service workers for offline load later), so all we need is to intercept the index.html URL.

If all we did was load the JS, it would be OK to have a JS callback invoked on every request our app makes (which is not that many). However, we load some external websites (from our own domain) which will do a lot of extra requests. So even though all we need is to inject a single HTML page and a service worker into our domain, our callback would be invoked on every request made.

I suppose JS is quite fast these days and this probably wouldn't even be a bottleneck for most apps (we probably wouldn't notice in our app), but I wanted to illustrate a scenario where avoiding unnecessary JS calls would be desirable. Also, one approach does not exclude the other: You can have your callback only invoked for certain URL patterns, and use invoke the original handler based on other conditions such as request headers/body.

Either way would be fine by me, if you want  I will close this PR in favor of #26065.

nornagon:
I think for now let's stick with just the more general method, implemented in #26065. We can revisit this in future if there are significant performance concerns.

hxkuc:
i think this is a good idea

PR #26682, opened 2020-11-24
feat: add MenuItem.userAccelerator property

MarshallOfSound:
This allows folks to read the user-assigned accelerator that macOS users can provide in system preferences. Useful for showing in-app shortcut help dialogs, you need to know if the accelerator you provided is not being used in favor of a user assigned one.

Notes: Added new `MenuItem.userAccelerator` property to read user-assigned macOS accelerator overrides

### DIFF
--- a/docs/api/menu-item.md
+++ b/docs/api/menu-item.md
@@ -159,7 +159,13 @@ A `String` (optional) indicating the item's role, if set. Can be `undo`, `redo`,
 
 #### `menuItem.accelerator`
 
-A `Accelerator` (optional) indicating the item's accelerator, if set.
+An `Accelerator` (optional) indicating the item's accelerator, if set.
+
+#### `menuItem.userAccelerator` _Readonly_ _macOS_
+
+An `Accelerator | null` indicating the item's [user-assigned accelerator](https://developer.apple.com/documentation/appkit/nsmenuitem/1514850-userkeyequivalent?language=objc) for the menu item.
+
+**Note:** This property is only initialized after the `MenuItem` has been added to a `Menu`. Either via `Menu.buildFromTemplate` or via `Menu.append()/insert()`.
 
 #### `menuItem.icon`
 
--- a/docs/api/menu-item.md
+++ b/docs/api/menu-item.md
@@ -165,7 +165,7 @@ An `Accelerator` (optional) indicating the item's accelerator, if set.
 
 An `Accelerator | null` indicating the item's [user-assigned accelerator](https://developer.apple.com/documentation/appkit/nsmenuitem/1514850-userkeyequivalent?language=objc) for the menu item.
 
-**Note:** This property is only initialized after the `MenuItem` has been added to a `Menu`. Either via `Menu.buildFromTemplate` or via `Menu.append()/insert()`.
+**Note:** This property is only initialized after the `MenuItem` has been added to a `Menu`. Either via `Menu.buildFromTemplate` or via `Menu.append()/insert()`.  Accessing before initialization will just return `null`.
 
 #### `menuItem.icon`
 
--- a/lib/browser/api/menu-item.ts
+++ b/lib/browser/api/menu-item.ts
@@ -43,6 +43,14 @@ const MenuItem = function (this: any, options: any) {
 
   this.overrideReadOnlyProperty('commandId', ++nextCommandId);
 
+  Object.defineProperty(this, 'userAccelerator', {
+    get: () => {
+      if (!this.menu) return null;
+      return this.menu._getUserAcceleratorAt(this.commandId);
+    },
+    enumerable: true
+  });
+
   const click = options.click;
   this.click = (event: Event, focusedWindow: BrowserWindow, focusedWebContents: WebContents) => {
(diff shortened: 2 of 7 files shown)


shell/browser/api/electron_api_menu_mac.mm
```diff
@@ -54,6 +73,29 @@
...
+      [[ElectronMenuController alloc] initWithModel:model()
+                              useDefaultAccelerator:NO]);
+
+  base::scoped_nsobject<NSMenuItem> item =
+      [controller makeMenuItemForIndex:GetIndexOfCommandId(command_id)
+                             fromModel:model()];
+  if ([[item userKeyEquivalent] length] == 0)
+    return v8::Null(isolate);
+
+  NSString* user_key_equivalent = [item keyEquivalent];
```
itsananderson, line 91: I'm not very familiar with these APIs, but it seems odd that you check `userKeyEquivalent` above, but then read the value from `keyEquivalent` and assign it to a variable named `user_key_equivalent`. Is this line also supposed to call `userKeyEquivalent`?

MarshallOfSound: You're correct that this is odd, but it is very much correct 😄 

`userKeyEquivalent` is a `NSString*` that contains either 0 or 1 character mapping to the literal key (ignoring modifiers) that the user has specified as the override.  If it's empty, there is no override.  The actual `keyEquivalent` and `keyEquivalentModifierMask` that the user specified are available in the standard format from the two keys we read from here.

It's weird but that's how Apple designed it 😆

shell/browser/api/electron_api_menu_mac.mm
```diff
@@ -24,6 +25,24 @@
 
 static scoped_nsobject<NSMenu> applicationMenu_;
 
+ui::Accelerator GetAcceleratorFromKeyEquivalentAndModifierMask(
```
codebytere, line 28: Can we not make use of [`GetKeyEquivalentAndModifierMaskFromAccelerator`](https://source.chromium.org/chromium/chromium/src/+/master:ui/base/accelerators/platform_accelerator_cocoa.mm;l=13-41q=%22%20NSEventModifierFlagShift%22&ss=chromium) in Chromium?

MarshallOfSound: @codebytere That function is the literal inverse of this one, the naming is similar because I manually reversed the Chromium impl for doing the opposite of what I wanted 😄

lib/browser/api/menu-item.ts
```diff
@@ -43,6 +43,14 @@ const MenuItem = function (this: any, options: any) {
 
   this.overrideReadOnlyProperty('commandId', ++nextCommandId);
 
+  Object.defineProperty(this, 'userAccelerator', {
+    get: () => {
+      if (!this.menu) return null;
+      return this.menu._getUserAcceleratorAt(this.commandId);
```
nornagon, line 49: Let's have this not fail when called on non-macOS, it can just return null safely I think.

docs/api/menu-item.md
```diff
@@ -159,7 +159,13 @@ A `String` (optional) indicating the item's role, if set. Can be `undo`, `redo`,
...
 #### `menuItem.accelerator`
 
-A `Accelerator` (optional) indicating the item's accelerator, if set.
+An `Accelerator` (optional) indicating the item's accelerator, if set.
+
+#### `menuItem.userAccelerator` _Readonly_ _macOS_
+
+An `Accelerator | null` indicating the item's [user-assigned accelerator](https://developer.apple.com/documentation/appkit/nsmenuitem/1514850-userkeyequivalent?language=objc) for the menu item.
+
+**Note:** This property is only initialized after the `MenuItem` has been added to a `Menu`. Either via `Menu.buildFromTemplate` or via `Menu.append()/insert()`.
```
nornagon, line 168: What will be returned if this is accessed before being added to a menu? null? undefined? JS exception? sigsegv?

MarshallOfSound: `null`, I'll update the docs 👍

shell/browser/api/electron_api_menu_mac.mm
```diff
@@ -54,6 +73,29 @@
...
+  if (![NSMenuItem usesUserKeyEquivalents])
+    return v8::Null(isolate);
+
+  auto controller = base::scoped_nsobject<ElectronMenuController>(
+      [[ElectronMenuController alloc] initWithModel:model()
+                              useDefaultAccelerator:NO]);
+
+  base::scoped_nsobject<NSMenuItem> item =
+      [controller makeMenuItemForIndex:GetIndexOfCommandId(command_id)
+                             fromModel:model()];
```
ckerr, line 87: do we need to test for index !== -1 to ensure an index was found before using it?

MarshallOfSound: I think it's a theoretical impossible state for that method to return `-1` given this is a non-public binding and we control the command_id passinto this method.  I guess it's a reasonable safety check though

nornagon (approved):

PR #26684, opened 2020-11-24
feat: add process.uptime() to sandboxed renderers

miniak:
#### Description of Change
Add [`process.uptime()`](https://nodejs.org/api/process.html#process_process_uptime) to sandboxed renderers.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `process.uptime()` to sandboxed renderers.

### DIFF
--- a/shell/renderer/electron_sandboxed_renderer_client.cc
+++ b/shell/renderer/electron_sandboxed_renderer_client.cc
@@ -90,6 +90,11 @@ v8::Local<v8::Value> CreatePreloadScript(v8::Isolate* isolate,
                                        preloadSrc);
 }
 
+double Uptime() {
+  return (base::Time::Now() - base::Process::Current().CreationTime())
+      .InSecondsF();
+}
+
 void InvokeHiddenCallback(v8::Handle<v8::Context> context,
                           const std::string& hidden_key,
                           const std::string& callback_name) {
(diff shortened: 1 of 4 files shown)


MarshallOfSound (approved):

shell/renderer/electron_sandboxed_renderer_client.cc
```diff
@@ -90,6 +90,12 @@ v8::Local<v8::Value> CreatePreloadScript(v8::Isolate* isolate,
                                        preloadSrc);
 }
 
+double Uptime() {
+  return (base::Time::NowFromSystemTime() -
```
nornagon, line 94: What's the difference between `base::Time::Now` and `base::Time::NowFromSystemTime`? In what situations will these return different values?

nornagon: From a bit of searching, it looks like the difference is only (a) during tests and (b) when "virtual time" is enabled, either through `--initial-virtual-time` or the [Emulation.setVirtualTimePolicy](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setVirtualTimePolicy) devtools protocol method is called.

It seems to me like there's not any particular reason to bypass the override in those cases, so perhaps `base::Time::Now` is the better choice here?

miniak: changed to `base::Time::Now`

nornagon (approved):

alexeykuzmin (approved):

zcbenz (approved):

Issue #9035, opened 2017-03-27
Method for retrieving file pointers from clipboard


darsain:
There is currently no mechanism to get/set file pointers from/to clipboard.

File pointers: The stuff in clipboard when you copy (`ctrl+c`) selected files in your OS file explorer.

After ~15 min research, there was #2244 issue which went a different direction and than got closed. Also there are some projects on hold waiting for this kind of functionality, such as https://github.com/bertofer/quercus/issues/1.

bompi88:
On mac I think I've used something like:

```js
var path = clipboard.read('public.file-url');
``

to retrieve the path of the copied file.

**Comment by evecalm** [comment]
@bompi88 this works well! But  seldom guys know this, and  can only get only file's path if multi files are copied. There should be an API like `clipboard.readPaths()` return an array of copied files' path

**Comment by timkendrick** [comment]
@evecalm From inspecting the clipboard with Apple's [clipboard viewer sample app](https://developer.apple.com/library/content/samplecode/ClipboardViewer/Introduction/Intro.html), it looks like you should be able to get the path of multiple files via `clipboard.read('NSFilenamesPboardType')`

This doesn't help with *setting* the clipboard though, seeing as Electron only allows you to copy plain text to the clipboard (not e.g. `NSFilenamesPboardType` data).

It'd be great if there were some way to set non-plaintext clipboard data.

**Comment by timkendrick** [comment]
Actually, after further investigation, it looks like copying files to the clipboard programmatically is also totally possible (on macOS at least) using the experimental [`clipboard.writeBuffer()`](https://github.com/electron/electron/blob/master/docs/api/clipboard.md#clipboardwritebufferformat-buffer-type-experimental) API:

```js
clipboard.writeBuffer(
  'NSFilenamesPboardType',
  Buffer.from(`
    <?xml version="1.0" encoding="UTF-8"?>
    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
    <plist version="1.0">
      <array>
        <string>/path/to/file1.ext</string>
        <string>/path/to/file2.ext</string>
      </array>
    </plist>
  `)
)
```

evecalm:
@timkendrick Awesome!

But I guess we need a much more developer-friendly api 😟

timkendrick:
[...]

PR #26693, opened 2020-11-25
feat: add clipboard.writeFilePaths/readFilePaths APIs

zcbenz:
#### Description of Change

Add `clipboard.writeFilePaths/readFilePaths` APIs that can be used to get/set file pointers in clipboard, close https://github.com/electron/electron/issues/9035.

Note that Chromium does not provide public APIs for writing file paths so we have to implement the API separately for each platform.

Also for the same reason we are unable to add `filePaths` to the `clipboard.write` API, otherwise a patch to Chromium would be required. But usually users would not need to write `filePaths` and other types of data into clipboard at the same time, and even if eventually we want to implement that, there will be no change to the `clipboard.writeFilePaths/readFilePaths` APIs.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Add `clipboard.writeFilePaths/readFilePaths` APIs.

### DIFF
--- a/shell/common/api/electron_api_clipboard_gtk.cc
+++ b/shell/common/api/electron_api_clipboard_gtk.cc
@@ -1,0 +1,73 @@
+// Copyright (c) 2020 Microsoft, Inc.
+// Use of this source code is governed by the MIT license that can be
+// found in the LICENSE file.
+
+#include "shell/common/api/electron_api_clipboard.h"
+
+#include <gtk/gtk.h>
+
+namespace electron {
+
+namespace api {
+
+namespace {
+
+void ClipboardGet(GtkClipboard* clipboard,
+                  GtkSelectionData* selection,
+                  guint info,
+                  gchar** uris) {
+  DCHECK_EQ(info, 0u);
+  gtk_selection_data_set_uris(selection, uris);
+}
+
+void ClipboardClear(GtkClipboard* clipboard, gchar** uris) {
+  g_strfreev(uris);
+}
+
+}  // namespace
+
+std::vector<base::FilePath> Clipboard::ReadFilePaths() {
+  std::vector<base::FilePath> result;
+  GtkClipboard* clipboard = gtk_clipboard_get(GDK_SELECTION_CLIPBOARD);
+  gchar** uris = gtk_clipboard_wait_for_uris(clipboard);
+  if (!uris)
+    return result;
+  for (int i = 0; uris[i]; i++) {
+    if (gchar* path = g_filename_from_uri(uris[i], nullptr, nullptr)) {
+      result.emplace_back(path);
+      g_free(path);
+    }
+  }
+  g_strfreev(uris);
+  return result;
+}
+
+void Clipboard::WriteFilePaths(const std::vector<base::FilePath>& paths) {
+  // Convert vector to gchar* array, will be destroyed by ClipboardClear.
+  gchar** uris = g_new(gchar*, paths.size() + 1);
+  if (!uris)
+    return;
+  size_t i;
+  for (i = 0; i < paths.size(); ++i) {
+    uris[i] = g_filename_to_uri(paths[i].value().c_str(), nullptr, nullptr);
+  }
+  uris[i] = nullptr;
+
+  GtkTargetList* targets = gtk_target_list_new(nullptr, 0);
+  gtk_target_list_add_uri_targets(targets, 0);
+  int number = 0;
+  GtkTargetEntry* table = gtk_target_table_new_from_list(targets, &number);
+  DCHECK_EQ(number, 1);
+
+  GtkClipboard* clipboard = gtk_clipboard_get(GDK_SELECTION_CLIPBOARD);
+  gtk_clipboard_set_with_data(clipboard, table, number,
+                              (GtkClipboardGetFunc)ClipboardGet,
+                              (GtkClipboardClearFunc)ClipboardClear, uris);
+
+  gtk_target_list_unref(targets);
+  gtk_target_table_free(table, number);
+}
+
+}  // namespace api
+
+}  // namespace electron
--- a/shell/common/api/electron_api_clipboard_mac.mm
+++ b/shell/common/api/electron_api_clipboard_mac.mm
@@ -2,8 +2,12 @@
 // Use of this source code is governed by the MIT license that can be
 // found in the LICENSE file.
 
-#include "base/strings/sys_string_conversions.h"
 #include "shell/common/api/electron_api_clipboard.h"
+
+#include <vector>
+
+#include "base/mac/foundation_util.h"
+#include "base/strings/sys_string_conversions.h"
 #include "ui/base/cocoa/find_pasteboard.h"
 
 namespace electron {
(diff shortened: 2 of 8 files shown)


shell/common/api/electron_api_clipboard_mac.mm
```diff
@@ -19,6 +23,27 @@
   return GetFindPboardText();
 }
 
+std::vector<base::FilePath> Clipboard::ReadFilePaths() {
+  std::vector<base::FilePath> result;
+  NSArray* paths = [[NSPasteboard generalPasteboard]
+      propertyListForType:NSFilenamesPboardType];
```
nornagon, line 29: similar code in chromium wraps with `@try` and also checks `[pasteboard types]` before calling `propertyListForType:`: https://source.chromium.org/chromium/chromium/src/+/master:ui/base/clipboard/clipboard_mac.mm;l=465;drc=bdf1e0a5273ce828b48801de30f88965c76eb6f9

should we do the same?

shell/common/api/electron_api_clipboard_win.cc
```diff
@@ -0,0 +1,83 @@
...
+
+#include "base/win/scoped_hglobal.h"
+
+namespace electron {
+
+namespace api {
+
+std::vector<base::FilePath> Clipboard::ReadFilePaths() {
+  std::vector<base::FilePath> result;
+  ::OpenClipboard(NULL);
```
nornagon, line 18: Chromium uses `ScopedClipboard` which has more complex logic: https://source.chromium.org/chromium/chromium/src/+/master:ui/base/clipboard/clipboard_win.cc;l=62;drc=510d3cc7ecba433b10d6854f3623ed4ca8fb72d7

Should we replicate it?

shell/common/api/electron_api_clipboard_gtk.cc
```diff
@@ -0,0 +1,73 @@
...
+void ClipboardClear(GtkClipboard* clipboard, gchar** uris) {
+  g_strfreev(uris);
+}
+
+}  // namespace
+
+std::vector<base::FilePath> Clipboard::ReadFilePaths() {
+  std::vector<base::FilePath> result;
+  GtkClipboard* clipboard = gtk_clipboard_get(GDK_SELECTION_CLIPBOARD);
+  gchar** uris = gtk_clipboard_wait_for_uris(clipboard);
```
nornagon, line 32: This seems like a dangerous function to call:

> This function waits for the data to be received using the main loop, so events, timeouts, etc, may be dispatched during the wait.

This means V8 could be re-entered (e.g. through an emitted event) while this method is running, which is undesirable IMO.

nornagon (changes requested):
This seems really hard to get right with a lot of platform-specific quirks :(

I'm OK with the API shape if we can find a satisfying way of implementing it.

shell/common/api/electron_api_clipboard_mac.mm
zcbenz: The referenced code is reading image which may throw, reading file paths is safe.

There is Chromium code reading file paths from drag drop clipboard:
https://source.chromium.org/chromium/chromium/src/+/master:ui/base/dragdrop/os_exchange_data_provider_mac.mm;l=244;drc=bdf1e0a5273ce828b48801de30f88965c76eb6f9?ss=chromium%2Fchromium%2Fsrc

shell/common/api/electron_api_clipboard_win.cc
zcbenz: On Windows the clipboard may be not available when it is being used by other apps, Chromium solves that by retrying after blocking for 5 seconds. I'm not sure if we need to replicate it, but it is easy to if you think it is necessary.

shell/common/api/electron_api_clipboard_gtk.cc
zcbenz: This is unfortunately also Chromium does, doing a nested run loop until data is received:
https://source.chromium.org/chromium/chromium/src/+/master:ui/base/clipboard/clipboard_x11.cc;l=354;drc=28d989a1fc83edb779c034cad48572775696bc66;bpv=0;bpt=1
https://source.chromium.org/chromium/chromium/src/+/master:ui/base/x/selection_requestor.cc;l=236;drc=510d3cc7ecba433b10d6854f3623ed4ca8fb72d7

For X11 reading clipboard is an async operation so this behavior can not be avoided.

Slapbox:
Is this still planned/in progress?

zcbenz:
I can't proceed without continued feedbacks and API approvals from @electron/wg-api.

shell/common/api/electron_api_clipboard_gtk.cc
nornagon: Chromium doesn't run V8 in the main process though, so reentry is much safer there. Is there some other way we could hook into the event loop here? Could we post a task?

zcbenz: Electron's clipboard APIs use Chromium's clipboard implementation, though Chromium does not run V8 in the main process, Electron does, so every time Electron app uses clipboard API on Linux, it has the reentry problem when waiting for X11 clipboard to respond.

I think using GTK's sync clipboard API has no difference from using Chromium's clipboard implementations.

jkleinsc (changes requested):
Chromium has a [Raw Clipboard Access API](https://crbug.com/897289) that I believe could provide the functionality being asked for in #9035.

zcbenz:
The Raw Clipboard Access API spec does not mention file paths, so with current spec to read/write file paths users have to manipulate the raw data directly, which is exactly the same with how users are trying to do with Electron's current `clipboard.read/write` APIs (https://github.com/electron/electron/issues/9035#issuecomment-722410816).

If we want to make Raw Clipboard Access API support read/write file paths, we have to request it explicitly to Chromium team, and probably write the specs and code, otherwise #9035 could not be considered fixed.

zcbenz:
> Per https://crbug.com/1217643, raw clipboard is actually being deprecated+removed now, so I guess this bug is no longer applicable. Sorry it never did end up getting fixed properly.

So we have to provide our own clipboard APIs as Raw Clipboard Access API is being removed.

shell/common/api/electron_api_clipboard_win.cc
```diff
@@ -0,0 +1,83 @@
...
+
+    size_t next_filename_offset = 0;
+    for (const auto& filename : paths) {
+      wcscpy(data + next_filename_offset, filename.value().c_str());
+      // Skip the terminating null character of the filename.
+      next_filename_offset += filename.value().length() + 1;
+    }
+  }
+
+  ::OpenClipboard(NULL);
```
kenan2002, line 75: What if `OpenClipboard()` fails? Should we free `hdata` in this case?

zcbenz: If `OpenClipboard` fails then `SetClipboardData` will also fail, and `hdata` will be released then.

samuelmaddock:
https://github.com/electron/governance/pull/454 includes mentions of a custom MIME type `electron/filepath`. Does that proposal supersede this PR? cc @codebytere @jkleinsc

nornagon:
Let's close this PR in favor of the [RFC](https://github.com/electron/governance/pull/454)

PR #26701, opened 2020-11-25
build: remove sccache

MarshallOfSound:
We don't use it for any builds anymore so let's save 20 seconds and not download it 😆 

Notes: no-notes


nornagon (approved):
rip to a real one

post malone pouring one out

codebytere (approved):

PR #26705, opened 2020-11-26
chore: use ESLint API for linting

dsanders11:
#### Description of Change

This is a better fix for #26014, and has some other benefits.

Count successes instead of expecting error output. Seems that the most common reason the linters get broken is that they aren't actually running what is expected, something silently fails, and silence is considered success. Invert that logic, and start with a fail state, so if nothing runs, it's a failure by default.

When in verbose mode, output names of files linted even if they ran clean. This again makes it easier to tell if a linter is running as intended, or silently failing.

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/script/lint.js
+++ b/script/lint.js
@@ -309,7 +309,7 @@ async function main () {
     const filenames = await findFiles(opts, linter);
     if (filenames.length) {
       if (opts.verbose) { console.log(`linting ${filenames.length} ${linter.key} ${filenames.length === 1 ? 'file' : 'files'}`); }
-      await inter.run(opts, filenames);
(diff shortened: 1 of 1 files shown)


dsanders11:
Ignore the extra commits, was a bit asleep at the wheel.

zcbenz (approved):

codebytere (approved):

miniak (approved):

miniak:
> Ignore the extra commits, was a bit asleep at the wheel.
@dsanders11 you can squash them into a single commit

Issue #26562, opened 2020-11-18
BrowserView draggable area not working in windows exclusively on BrowserViews


pejman1:
### Issue Details
in frameless windows 
While setting browserview style to draggable by webkit-app-region : drag. the area won't be draggable after that
and it can not move main window 
### Electron Version:
11.0.1
### Operating System:
Windows 10 (1803)
Last Known Working Electron version:
not found
### Expected Behavior
Browser view moves the main window while I set the style of the body or one of the area to
-webkit-app-region: drag;

### Actual Behavior
nothing happen in browserview after styling -webkit-app-region: drag;

I know it is mention that in 11 version it is fixed. but it still dosn't work 

### To Reproduce
https://gist.github.com/pejmanpesyan/8540aad65e7c6980c45f23e5eaafafb9

ckerr:
Thanks for reporting this and helping to make Electron better!

Would it be possible for you to make a standalone testcase with only the code necessary to reproduce the issue? For example, [Electron Fiddle](https://www.electronjs.org/fiddle) is a great tool for making small test cases and makes it easy to publish your test case to a [gist](https://gist.github.com) that Electron maintainers can use.

Stand-alone test cases make fixing issues go more smoothly: it ensure everyone's looking at the same issue, it removes all unnecessary variables from the equation, and it can also provide the basis for automated regression tests.

I'm adding the `blocked/needs-repro` label for this reason. After you make a test case, please link to it in a followup comment.

Thanks in advance! Your help is appreciated.

pejman1:
> Thanks for reporting this and helping to make Electron better!
> 
> Would it be possible for you to make a standalone testcase with only the code necessary to reproduce the issue? For example, [Electron Fiddle](https://www.electronjs.org/fiddle) is a great tool for making small test cases and makes it easy to publish your test case to a [gist](https://gist.github.com) that Electron maintainers can use.
> 
> Stand-alone test cases make fixing issues go more smoothly: it ensure everyone's looking at the same issue, it removes all unnecessary variables from the equation, and it can also provide the basis for automated regression tests.
> 
> I'm adding the `blocked/needs-repro` label for this reason. After you make a test case, please link to it in a followup comment.
> 
> Thanks in advance! Your help is appreciated.

https://gist.github.com/pejmanpesyan/8540aad65e7c6980c45f23e5eaafafb9

codebytere:
[...]

PR #26145, opened 2020-10-24
fix: draggable regions exclusively on BrowserViews

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/26114.

 Previously, it was impossible to set a draggable region _exclusively_ on a `BrowserView`. If one had `BrowserView(s)` on a `BrowserWindow`, and a draggable region on the `BrowserWindow` we made it such that the draggable region properly worked on top of the `BrowserView`, but you couldn’t set a draggable region _only_ on a `BrowserView`, where none existed on a `BrowserWindow`. This fixes that and pulls out related functionality into a helper util which already existed.

cc @zcbenz @deepak1556 @MarshallOfSound 

#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Fixed an issue where draggable regions did not work exclusively on `BrowserViews`.

PR #26738, opened 2020-11-30
fix: draggable views on BrowserViews on Windows

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/26562.

Prior to this PR draggable views did not work with BrowserViews except on macOS, which was fixed in https://github.com/electron/electron/pull/26145. This fixes that same functionality on Windows.

Tested with https://gist.github.com/7278c8c62ec976110afb47ef2d861a7a.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes:  Fixed an issue where draggable regions did not work exclusively on BrowserViews on Windows.

### DIFF
--- a/shell/browser/native_browser_view_views.h
+++ b/shell/browser/native_browser_view_views.h
@@ -5,7 +5,11 @@
 #ifndef SHELL_BROWSER_NATIVE_BROWSER_VIEW_VIEWS_H_
 #define SHELL_BROWSER_NATIVE_BROWSER_VIEW_VIEWS_H_
 
+#include <memory>
+#include <vector>
+
 #include "shell/browser/native_browser_view.h"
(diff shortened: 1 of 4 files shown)


ckerr (approved):

shell/browser/native_browser_view_views.h
```diff
@@ -40,6 +48,8 @@ class NativeBrowserViewViews : public NativeBrowserView {
   float auto_vertical_proportion_height_ = 0.;
   float auto_vertical_proportion_top_ = 0.;
```
Ivan-Kurochenko, line 50: where can I find `-webkit-app-region` parsing?
I wanna find out how Electron works if `-webkit-app-region` is applied.
@ckerr @codebytere @lsegal @vitaliel @anaisbetts

PR #26749, opened 2020-12-01
fix: internalModuleReadJSON for unpacked JSON

vadim-termius:
#### Description of Change

For some reasons, one `return` statement wasn't updated to the new output type consisting of an array in the next PR: https://github.com/electron/electron/pull/24707/files#diff-0c073d40320d69e1a416d4610fa7720f374be5f0d28ff9a0d8ad9c34b358e192R685

This broke import of modules which `package.json` is unpacked.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed import of unpacked node modules.

### DIFF
--- a/lib/asar/fs-wrapper.ts
+++ b/lib/asar/fs-wrapper.ts
@@ -689,7 +689,8 @@ export const wrapFsWithAsar = (fs: Record<string, any>) => {
     if (info.size === 0) return ['', false];
     if (info.unpacked) {
       const realPath = archive.copyFileOut(filePath);
-      return fs.readFileSync(realPath, { encoding: 'utf8' });
+      const str = fs.readFileSync(realPath, { encoding: 'utf8' });
+      return [str, str.length > 0];
     }
 
     logASARAccess(asarPath, filePath, info.offset);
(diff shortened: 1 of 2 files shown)


vadim-termius:
cc @codebytere

vadim-termius:
Does CI run tests? It seems like running tests requires full build of the project, and so I decide to skip running `npm run test` on my local machine because changes are quite small and I expect them to pass CI.

codebytere:
hey @vadim-termius - thanks for this! The two backports you opened are typically done automatically on merge unless the patch cannot be cherry-picked - you can ask us to add appropriate labels for check that in the future. The backports will show as `Invalid` right now because this PR is still open, but will become valid once this PR is merged.

codebytere (approved):
lgtm

vadim-termius:
@codebytere, cool, thank you for the information, I didn't know about that. Should I remove my backport PRs?

zcbenz (approved):

vladimiry:
> For some reasons, one return statement wasn't updated to the new output type consisting of an array

By the way, when it comes to TypeScript use it's a good practice to always explicitly specify the function return signature rather than just going with type inference. So then when you change the function logic and return an unexpected result the issue will be noticed by the compiler. And then if you actually need to change the return you will have to change the function signature clearly understanding the consequences.

momakes2:
When will we get this gem? <3

PR #26764, opened 2020-12-01
feat: add event.senderFrame property returning the webFrameMain

miniak:
#### Description of Change
Also add missing `event.processId`, which needs to be passed to `webFrameMain.fromId()`.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `event.senderFrame` property returning the originating `webFrameMain` of the IPC message.

### DIFF
--- a/docs/api/structures/ipc-main-event.md
+++ b/docs/api/structures/ipc-main-event.md
@@ -4,6 +4,7 @@
 * `frameId` Integer - The ID of the renderer frame that sent this message
 * `returnValue` any - Set this to the value to be returned in a synchronous message
 * `sender` WebContents - Returns the `webContents` that sent the message
+* `senderFrame` WebFrameMain _Readonly_ - The frame that sent this message
 * `ports` MessagePortMain[] - A list of MessagePorts that were transferred with this message
 * `reply` Function - A function that will send an IPC message to the renderer frame that sent the original message that you are currently handling.  You should use this method to "reply" to the sent message in order to guarantee the reply will go to the correct process and frame.
   * `channel` String
--- a/docs/api/web-frame-main.md
+++ b/docs/api/web-frame-main.md
@@ -26,7 +26,7 @@ win.webContents.on(
 )
 ```
 
-You can also access frames of existing pages by using the `webFrame` property
+You can also access frames of existing pages by using the `mainFrame` property
 of [`WebContents`](web-contents.md).
 
 ```javascript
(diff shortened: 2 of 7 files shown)


docs/api/structures/ipc-main-event.md
```diff
@@ -1,8 +1,10 @@
 # IpcMainEvent Object extends `Event`
 
 * `frameId` Integer - The ID of the renderer frame that sent this message
+* `processId` Integer - The ID of the process which owns the frame
```
nornagon, line 4: Is this the OS process ID or the internal process ID? The docs should say.

docs/api/web-frame-main.md
```diff
@@ -26,7 +26,7 @@ win.webContents.on(
 )
 ```
 
-You can also access frames of existing pages by using the `webFrame` property
+You can also access frames of existing pages by using the `mainFrame` property
```
nornagon, line 29: should this fix be a separate PR + backported to 12.x?

lib/browser/api/web-contents.ts
```diff
@@ -493,6 +493,12 @@ const addReplyInternalToEvent = (event: any) => {
   });
 };
 
+const addSenderFrameToEvent = (event: any) => {
+  Object.defineProperty(event, 'senderFrame', {
+    get: () => webFrameMain.fromId(event.processId, event.frameId)
+  });
```
nornagon, line 499: I'd prefer this to be resistant to the following:

```js
wc.on('event', (event) => {
  event.processId = 123
  event.senderFrame // ==> returns some other frame because the process ID changed
})
```

```suggestion
  const { processId, frameId } = event
  Object.defineProperty(event, 'senderFrame', {
    get: () => webFrameMain.fromId(processId, frameId)
  });
```

nornagon (changes requested):

docs/api/web-frame-main.md
miniak: I would like to backport this whole PR to 12-x-y if possible

nornagon (approved):
API LGTM

lib/browser/api/web-contents.ts
```diff
@@ -493,6 +493,13 @@ const addReplyInternalToEvent = (event: any) => {
   });
 };
 
+const addSenderFrameToEvent = (event: any) => {
+  const { processId, frameId } = event;
+  Object.defineProperty(event, 'senderFrame', {
+    get: () => webFrameMain.fromId(processId, frameId)
+  });
+};
```
samuelmaddock, line 501: This is great!

samuelmaddock (approved):
The getter is going to be very handy. Nice!

jkleinsc (approved):
API LGTM as well

nornagon:
looks like tests are failing. btw `processId` is also added by https://github.com/electron/electron/pull/26875

miniak:
@nornagon I've noticed that, I had it originally
```cpp
    dict.Set("processId", frame->GetProcess()->GetID());
    dict.Set("frameId", frame->GetRoutingID());
```
and then changed it to match your PR to avoid conflict
```cpp
    dict.Set("frameId", frame->GetRoutingID());
    dict.Set("processId", frame->GetProcess()->GetID());
```
which broke the snapshot for tests

nornagon:
@miniak the linked PR updates the tests to replace the process id with a placeholder in the snapshot.

nornagon (approved):

PR #26788, opened 2020-12-02
feat: add missing 'context-menu' values

codebytere:
#### Description of Change

This PR fills in a nontrivial amount of data missing from the object emitted by the 'context-menu' event. It'd been a significant amount of time since this was last updated, so a number of fields have been added to the object which may be useful to users handling the event and which we should therefore expose.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added missing fields to the parameters emitted as part of the `context-menu` event.

### DIFF
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -687,7 +687,6 @@ Returns:
     * `canPaste` Boolean - Whether the renderer believes it can paste.
     * `canDelete` Boolean - Whether the renderer believes it can delete.
     * `canSelectAll` Boolean - Whether the renderer believes it can select all.
-    * `canTranslate` Boolean - Whether the renderer believes it can perform translation.
     * `canEditRichly` Boolean - Whether the renderer believes it can edit text richly.
 
 Emitted when there is a new context menu that needs to be handled.
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -673,6 +673,7 @@ Returns:
     * `canToggleControls` Boolean - Whether the media element's controls are
       toggleable.
     * `canPrint` Boolean - Whether the media element can be printed.
+    * `canSave` Boolean - Whether or not the media element can be downloaded.
     * `canShowPictureInPicture` Boolean - Whether the media element can show picture-in-picture.
     * `isShowingPictureInPicture` Boolean - Whether the media element is currently showing picture-in-picture.
     * `canRotate` Boolean - Whether the media element can be rotated.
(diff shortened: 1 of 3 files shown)


docs/api/web-contents.md
```diff
@@ -674,6 +686,8 @@ Returns:
     * `canPaste` Boolean - Whether the renderer believes it can paste.
     * `canDelete` Boolean - Whether the renderer believes it can delete.
     * `canSelectAll` Boolean - Whether the renderer believes it can select all.
+    * `canTranslate` Boolean - Whether the renderer believes it can perform translation.
+    * `canEditRichly` Boolean - Whether the renderer believes it can edit text richly.
```
jkleinsc, line 691: `canSave` is missing from docs here

jkleinsc (changes requested):
LGTM apart from a small doc change needed

docs/api/web-contents.md
```diff
@@ -664,16 +672,23 @@ Returns:
...
     is able to perform the corresponding action.
     * `canUndo` Boolean - Whether the renderer believes it can undo.
     * `canRedo` Boolean - Whether the renderer believes it can redo.
     * `canCut` Boolean - Whether the renderer believes it can cut.
-    * `canCopy` Boolean - Whether the renderer believes it can copy
+    * `canCopy` Boolean - Whether the renderer believes it can copy.
     * `canPaste` Boolean - Whether the renderer believes it can paste.
     * `canDelete` Boolean - Whether the renderer believes it can delete.
     * `canSelectAll` Boolean - Whether the renderer believes it can select all.
+    * `canTranslate` Boolean - Whether the renderer believes it can perform translation.
```
MarshallOfSound, line 690: This will never be `true` in Electron as we don't (and won't) have a translation service

MarshallOfSound (changes requested):
Can we make sure this is tagged as `feat:`

MarshallOfSound (approved):

zcbenz (approved):

jkleinsc (approved):

jkleinsc:
API LGTM

nornagon (commented):
There are a few fields that still aren't exposed here, e.g. [writing_direction_*](https://source.chromium.org/chromium/chromium/src/+/master:content/public/common/untrustworthy_context_menu_params.h;l=127;drc=02f28360f2f6588ca2428ef93c8129e0a8cc231c), [impression](https://source.chromium.org/chromium/chromium/src/+/master:content/public/common/untrustworthy_context_menu_params.h;l=75;drc=02f28360f2f6588ca2428ef93c8129e0a8cc231c). Are there any particular reasons for these omissions?

miniak (approved):

sick78:
so what i not get paid ?

sick78:
so what i not get paid ?

PR #26802, opened 2020-12-02

### LINKED ISSUE #24833: Proposal: Allow BrowserView to be a newGuest in new-window
### Preflight Checklist

### Problem Description

The `BrowserView` is much like `BrowserWindow`, but it cannot be used as a "window" for the `new-window` event.

Imagine a tabbed layout with several `BrowserView`s, much like the regular browser window. The `webPreferences.nativeWindowOpen` is enabled.

*Use-case 1:* The page inside one `BrowserView` tries to open a new window via `window.open()`. There's no way to handle `new-window` and create a new `BrowserView` such that the `window.opener` JavaScript property refers to its opener.

*Use-case 2:* A `<form target="_blank" method="POST">` inside `BrowserView` is submitted. A new `BrowserView` can be created and its `WebContents` `loadURL()` method can be used to navigate using `POST` method. But AFAIK there's no way to know what exactly `postData` that form was going to POST.

It works fine if the `new-window` is default-prevented and a new `BrowserWindow` is created with the `options.webContents` as a constructor option. But it is ignored for a `BrowserView`

### Proposed Solution

- Allow setting `webContents` in a `BrowserView` constructor options.
- Allow `BrowserView` to be a `e.newGuest` in a default-prevented `new-window` event.

sentialx:
#### Description of Change
This PR adds support for `webContents` option in BrowserView, which works similarly as in BrowserWindow. Since the new `webContents.setWindowOpenHandler` doesn't support BrowserViews, we can at least listen to the internal `-add-new-contents` event and create a BrowserView with provided `webContents`. This PR will also help in adding support for BrowserView in `setWindowOpenHandler` in the future.

Closes #24833

Could this be backported to 12-x-y?

#### Checklist

- [x] `npm test` passes

### Release Notes

Since it's a hidden API, I think it's not worth mentioning this change in the release notes.

Notes: no-notes

### DIFF
--- a/shell/browser/api/electron_api_browser_view.cc
+++ b/shell/browser/api/electron_api_browser_view.cc
@@ -78,8 +78,17 @@ BrowserView::BrowserView(gin::Arguments* args,
       gin::Dictionary::CreateEmpty(isolate);
   options.Get(options::kWebPreferences, &web_preferences);
   web_preferences.Set("type", "browserView");
-  gin::Handle<class WebContents> web_contents =
-      WebContents::New(isolate, web_preferences);
+
+  v8::Local<v8::Value> value;
+
+  // Copy the webContents option to webPreferences. This is only used internally
+  // to implement nativeWindowOpen option.
+  if (options.Get("webContents", &value)) {
+    web_preferences.SetHidden("webContents", value);
+  }
+
+  auto web_contents =
+      WebContents::CreateFromWebPreferences(args->isolate(), web_preferences);
 
   web_contents_.Reset(isolate, web_contents.ToV8());
   api_web_contents_ = web_contents.get();
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -3765,6 +3765,33 @@ gin::Handle<WebContents> WebContents::FromOrCreate(
 }
 
 // static
+gin::Handle<WebContents> WebContents::CreateFromWebPreferences(
+    v8::Isolate* isolate,
+    const gin_helper::Dictionary& web_preferences) {
+  // Check if webPreferences has |webContents| option.
+  gin::Handle<WebContents> web_contents;
+  if (web_preferences.GetHidden("webContents", &web_contents) &&
+      !web_contents.IsEmpty()) {
+    // Set webPreferences from options if using an existing webContents.
+    // These preferences will be used when the webContent launches new
+    // render processes.
+    auto* existing_preferences =
+        WebContentsPreferences::From(web_contents->web_contents());
+    base::DictionaryValue web_preferences_dict;
+    if (gin::ConvertFromV8(isolate, web_preferences.GetHandle(),
+                           &web_preferences_dict)) {
+      existing_preferences->Clear();
+      existing_preferences->Merge(web_preferences_dict);
+    }
+  } else {
+    // Create one if not.
+    web_contents = WebContents::New(isolate, web_preferences);
+  }
+
+  return web_contents;
+}
+
+// static
 WebContents* WebContents::FromID(int32_t id) {
   return GetAllWebContents().Lookup(id);
 }
--- a/shell/browser/api/electron_api_web_contents.h
+++ b/shell/browser/api/electron_api_web_contents.h
@@ -135,6 +135,10 @@ class WebContents : public gin::Wrappable<WebContents>,
       v8::Isolate* isolate,
       content::WebContents* web_contents);
 
+  static gin::Handle<WebContents> CreateFromWebPreferences(
+      v8::Isolate* isolate,
+      const gin_helper::Dictionary& web_preferences);
+
   // gin::Wrappable
   static gin::WrapperInfo kWrapperInfo;
   static v8::Local<v8::ObjectTemplate> FillObjectTemplate(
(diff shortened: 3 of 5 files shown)


nornagon:
I'm not sure how I understand the way this would work—would you pass `show: false` in the BrowserWindowConstructor options in `setWindowOpenOverrideHandler`, so the BrowserWindow would still be created behind the scenes but hidden, and then yank the WebContents out of it and put it in a BrowserView? That seems like it might cause other problems 🤔

sentialx:
It allows us to override the internal `-add-new-contents` event and implement our own way to create a BrowserView
```ts
tab.webContents.removeAllListeners('-add-new-contents' as any);
  tab.webContents.on(
    '-add-new-contents' as any,
    (
      event: any,
      webContents: Electron.WebContents,
      disposition: string,
      _userGesture: boolean,
      _left: number,
      _top: number,
      _width: number,
      _height: number,
      url: string,
      frameName: string,
      referrer: Electron.Referrer,
      rawFeatures: string,
      postData: PostData,
    ) => {
      if (
        disposition !== 'foreground-tab' &&
        disposition !== 'new-window' &&
        disposition !== 'background-tab'
      ) {
        event.preventDefault();
        return;
      }
      new BrowserView({ webContents });
    },
  );
```

nornagon:
Given this API is intended to be undocumented, I'm downgrading from `semver/minor` to `semver/patch` and removing the API review request.

JFYI, we don't support overriding `-add-new-contents` and it's entirely possible that this workaround will break in future. I'd encourage you to define and document a public API that accomplishes this if you want this functionality to be stable.

nornagon (changes requested):
Please add some tests, thanks!

sentialx:
@nornagon done.

jkleinsc (approved):

nornagon (approved):

sentialx:
Sorry for changing my mind, but would it be possible to also backport to 10 and 11?

zcbenz (approved):

zcbenz:
> Sorry for changing my mind, but would it be possible to also backport to 10 and 11?

This is for @electron/wg-releases to decide, I'll add related tags.

VerteDinde:
Hey folks, the Releases WG approved this for backport - we wanted to confirm, this backport will result in a patch bump, not a minor bump, right? Thanks!

PalmerAL:
Is there any plan to turn this into a public API? This would be really useful for me, but I'm concerned about depending on it as long as it remains undocumented.

pOoOf:
Hey guys, i am trying to use this API to open a browserview as a guest window upon window.open and this is what i am doing to make it work

``` 
rightPaneWindowRef.webContents.removeAllListeners('-add-new-contents');
rightPaneWindowRef.webContents.on(
  '-add-new-contents',
  (
    newViewevent,
    webContents,
    disposition,
    _userGesture,
    _left,
    _top,
    _width,
    _height,
    newViewurl
  ) => {
    console.log({ disposition, newViewurl });
    const newView = new BrowserView({
      ...browserViewConfig,
      webContents,
    });
    mainWindow.addBrowserView(newView);
    newView.setBounds({
      x: leftViewWidth + 100,
      y: HEADER_HEIGHT + 100,
      width: 600,
      height: 600,
    });
  }
);
```

Now this works fine but when i try to close the BrowserView by clicking on any option on the UI(it may be calling `window.opener`), the application crashes.

Can anyone tell me what i could be doing wrong here?

Also what @sentialx suggested doesn't work, this code below doesn't open any `BrowserView` and nothing is visible on the UI utill i add this `BrowserView` into the BrowserWindow using `addBrowserWindow` method
```
tab.webContents.removeAllListeners('-add-new-contents' as any);
  tab.webContents.on(
    '-add-new-contents' as any,
    (
      event: any,
      webContents: Electron.WebContents,
      disposition: string,
      _userGesture: boolean,
      _left: number,
      _top: number,
      _width: number,
      _height: number,
      url: string,
      frameName: string,
      referrer: Electron.Referrer,
      rawFeatures: string,
      postData: PostData,
    ) => {
      if (
        disposition !== 'foreground-tab' &&
        disposition !== 'new-window' &&
        disposition !== 'background-tab'
      ) {
        event.preventDefault();
        return;
      }
      new BrowserView({ webContents });
    },
  );
```

PalmerAL:
In my case, removing the view from the window and calling `webContents.destroy()` ([code](https://github.com/minbrowser/min/blob/3dfc06bbe7e71eceb85312d0e6400c520f39fde6/main/viewManager.js#L111)). It doesn’t crash every time though, so there may be something else going on. I’ll put together a testcase sometime in the next couple days and see what I can figure out.

pOoOf:
> Now this works fine but when i try to close the BrowserView by clicking on any option on the UI(it may be calling `window.opener`), the application crashes.

When i said the application crashes, the application wasn't actually crashing which i later realized, what actually was happening was that, the new BrowserView(new Guest window) was closing the mainWindow upon calling window.close, i guess this is also an [existing issue](https://github.com/electron/electron/issues/28260) when we use BrowserView as a guest window

PalmerAL:
In my case, it is actually crashing, but after testing it again, I'm only seeing the crash infrequently (maybe 1/20 view closes?). I may still try to make a testcase at some point, but I don't think I can until I find a way to reproduce it consistently.

I don't have an actual crash report; this is what I get as console output:
```
[0514/162332.291416:WARNING:process_memory_mac.cc(93)] mach_vm_read(0x7ffee6d15000, 0x2000): (os/kern) invalid address (1)
[1] [0514/162332.520512:WARNING:crash_report_exception_handler.cc(240)] UniversalExceptionRaise: (os/kern) failure (5)
```
But that's probably not very helpful for diagnosing.

Issue #26713, opened 2020-11-27
Electron 11: `remote.screen` EventEmitter methods are undefined


deanylev:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 11.03
* **Operating System:**
  * macOS 11.0.1

### Expected Behavior
`remote.screen` has EventEmitter methods such as `on` defined, to listen to screen events such as `display-added`

### Actual Behavior
`remote.screen` does not have any defined EventEmitter methods

### To Reproduce
In the devtools, inspect `require('electron').remote.screen` and notice that it does not have the `on` method defined

### Additional Information
There was a PR to address other `remote.screen` methods not being defined that was merged in 11.0.3 (https://github.com/electron/electron/pull/26660), but it did not seem to address this bug

nornagon:
Duplicate of #26610.

nornagon:
Oops, I see now that this is a different issue.

PR #26620, opened 2020-11-21
fix: make screen wrapper remote-friendly again

andersk:
#### Description of Change

This restores accessibility of `screen` methods via `remote.screen`.

Fixes #26610.

Cc @nornagon.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Fixed `screen` methods not being accessible via `remote.screen`.

PR #26809, opened 2020-12-03
fix: screen EventEmitter methods with remote

codebytere:
#### Description of Change

Follow-up to https://github.com/electron/electron/pull/26620.
Closes https://github.com/electron/electron/issues/26713.

Fixes an issue whereby `remote.screen` EventEmitter methods are undefined in the renderer. This is fixed by implementing `getPrototypeOf` as a Proxy trap.

cc @nornagon @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue whereby `remote.screen` `EventEmitter` methods are undefined in the renderer.

### DIFF
--- a/lib/browser/api/screen.ts
+++ b/lib/browser/api/screen.ts
@@ -1,3 +1,5 @@
+import { EventEmitter } from 'events';
+
 const { createScreen } = process._linkedBinding('electron_common_screen');
 
 let _screen: Electron.Screen;
@@ -36,5 +38,11 @@ export default new Proxy({}, {
   getOwnPropertyDescriptor: (target, property: string) => {
     createScreenIfNeeded();
     return Reflect.getOwnPropertyDescriptor(_screen, property);
+  },
+  getPrototypeOf: () => {
+    // This is necessary as a result of weirdness with EventEmitterMixin
+    // and FunctionTemplate - we need to explicitly ensure it's returned
+    // in the prototype.
+    return EventEmitter.prototype;
   }
 });


MarshallOfSound (approved):

lib/browser/api/screen.ts
```diff
@@ -1,38 +1,41 @@
+import { EventEmitter } from 'events';
+
 const { createScreen } = process._linkedBinding('electron_common_screen');
 
 let _screen: Electron.Screen;
 
+const ensureScreen = () => {
+  _screen = createScreen();
+  EventEmitter.call(_screen);
```
nornagon, line 9: Is this `EventEmitter` constructor call necessary? We didn't need it before.

lib/browser/api/screen.ts
```diff
@@ -1,38 +1,41 @@
+import { EventEmitter } from 'events';
+
 const { createScreen } = process._linkedBinding('electron_common_screen');
 
 let _screen: Electron.Screen;
 
+const ensureScreen = () => {
+  _screen = createScreen();
```
nornagon, line 8: perhaps roll the check for `undefined` into this function too?

lib/browser/api/screen.ts
```diff
@@ -1,38 +1,41 @@
...
   ownKeys: () => {
-    if (_screen === undefined) {
-      _screen = createScreen();
-    }
+    if (_screen === undefined) ensureScreen();
     return Reflect.ownKeys(_screen);
   },
+  getPrototypeOf: (target) => {
+    const targetPrototype = Reflect.getPrototypeOf(target);
+    return Object.assign(EventEmitter.prototype, targetPrototype);
```
nornagon, line 31: yikes, you should not be overriding things in `EventEmitter.prototype` here!!

nornagon (changes requested):

lib/browser/api/screen.ts
codebytere: oops, typo. Should be swapped.

lib/browser/api/screen.ts
codebytere: @nornagon we did, it was done previously in `createLazyInstance`

lib/browser/api/screen.ts
nornagon: I don't think this should edit the target's prototype, either

lib/browser/api/screen.ts
nornagon: that was prior to the ginification though, yes? I don't believe the new code needs it.

codebytere: i'll check!

lib/browser/api/screen.ts
codebytere: @nornagon this is also what was done previously - https://github.com/electron/electron/commit/e6cf5906f60f060800d8368f80abf7019d2d4a5a#diff-3cab00a5b0a35804ff97b9c92fb7e719645bccb38c4a217957e5e58d153d5ceaL8

nornagon: @codebytere the prototype assignment is now done by C++, which is why I removed it.

lib/browser/api/screen.ts
```diff
@@ -1,38 +1,41 @@
...
-    }
+    ensureScreen();
     return Reflect.ownKeys(_screen);
   },
-  has: (target, prop: string) => {
-    if (_screen === undefined) {
-      _screen = createScreen();
-    }
+  getPrototypeOf: (target) => {
+    return Object.setPrototypeOf(target, EventEmitter.prototype);
```
nornagon, line 31: EventEmitter.prototype is already the prototype of `screen`.

AdrianMichigan: 67

AdrianMichigan:
> #### Description of Change
> 
> Follow-up to https://github.com/electron/electron/pull/26620.
> Closes https://github.com/electron/electron/issues/26713.
> 
> Fixes an issue whereby `remote.screen` EventEmitter methods are undefined in the renderer. This is fixed by implementing `getPrototypeOf` as a Proxy trap.
> 
> cc @nornagon @MarshallOfSound 
> 
> #### Checklist
> 
> 
> - [x] PR description included and stakeholders cc'd
> - [x] `npm test` passes
> - [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).
> 
> #### Release Notes
> 
> Notes: Fixed an issue whereby `remote.screen` `EventEmitter` methods are undefined in the renderer. 
>

lib/browser/api/screen.ts
```diff
@@ -1,38 +1,41 @@
...
-    }
+    ensureScreen();
     return Reflect.ownKeys(_screen);
   },
-  has: (target, prop: string) => {
-    if (_screen === undefined) {
-      _screen = createScreen();
-    }
+  getPrototypeOf: () => {
+    return EventEmitter.prototype;
```
MarshallOfSound, line 31: Is this 100% accurate? I would have thought this would be
```suggestion
    if (_screen === undefined) {
      _screen = createScreen();
    }
    return Reflect.getPrototypeOf(_screen);
```

codebytere: @MarshallOfSound that prototype is always null - it's some weirdness with the c++ prototype setters

nornagon (approved):

zcbenz (approved):

Issue #26677, opened 2020-11-24
Electron 11 breaks Error stack in a loaded iframe


sergeyd63:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  *  11.0.3
* **Operating System:**
  * All
* **Last Known Working Electron version:**
  * 10.1.6

### Expected Behavior

Should send a full Error stack

### Actual Behavior
In an iframe the error stack = "Error"

### To Reproduce
Website has an iframe. On load the stack is not loaded

### Screenshots
In a Browser (Chrome)
![image](https://user-images.githubusercontent.com/4738739/100126873-804f3980-2e4c-11eb-9d08-b56a53664c87.png)

In Electron 11
![image](https://user-images.githubusercontent.com/4738739/100127013-a8d73380-2e4c-11eb-8885-212aa8ec222f.png)

jnises:
I'm getting the same issue. No iframes involved though.
In electron 10.1.6 i get this in the devconsole: 
![image](https://user-images.githubusercontent.com/1059884/100872677-722f9780-34a2-11eb-848e-206bb94fd68e.png)
In electron 11.0.3 I get this: 
![image](https://user-images.githubusercontent.com/1059884/100872719-84113a80-34a2-11eb-8a26-ae847621549b.png)

This breaks my Sentry integration for example.

codebytere:
It would be helpful if either of you could please provide a reproducible sample. I can reproduce this with `contextIsolation` but not otherwise, and it's important for better triage that I know the _exact_ conditions under which you see this so I'm not forced to make guesses.

jnises:
Yes. Only with contextIsolation.
Here is a minimal example: https://github.com/jnises/electron_stack_test

codebytere:
thanks @jnises! I'll see what i can do.

codebytere:
@jnises @sergeyd63 I've located the error and a fix is up

jnises:
great :)
that was quick

Raynos:
When will we have a new version of electron 11 that we can use that fixes this bug ?

PR #26820, opened 2020-12-03
fix: stack traces in non-Node.js contexts

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/26677.

Node.js sets a stack trace handler specific to the context corresponding to the current `Environment`. When we're running in a non-Node.js context, there will be no correspondent `Environment` - we therefore need to unset this handler so that Blink falls back to its default handling and displays the correct stacktrace.

Upstreamed at https://github.com/nodejs/node/pull/36447.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where renderer process stack traces were broken with `contextIsolation` enabled.


shell/renderer/electron_renderer_client.cc
```diff
@@ -154,6 +154,15 @@ void ElectronRendererClient::DidCreateScriptContext(
   // Load everything.
   node_bindings_->LoadEnvironment(env);
 
+  // Node.js sets a stack trace handler specific to the v8::Context
+  // corresponding to the current Environment. When we're running in a
+  // non-Node.js v8::Context, there will be no correspondent Environment - we
+  // therefore need to unset this handler so that Blink falls back to its
+  // default handling and displays the correct stacktrace.
+  if (prefs.context_isolation) {
+    renderer_context->GetIsolate()->SetPrepareStackTraceCallback(nullptr);
```
MarshallOfSound, line 163: Don't we need to do the trick we do elsewhere and call into either blinks or nodes handler depending on the current context?  This will just fallback to the default for nodes case as well which we don't want right?

codebytere: I don't think so - we could just fall back in the renderer process unilaterally but the only case that presents issues is when `contextIsolation` is enabled.

nornagon (approved):

shell/common/node_bindings.cc
```diff
@@ -503,6 +503,12 @@ node::Environment* NodeBindings::CreateEnvironment(
     // context. We need to use the one Blink already provides.
     is.flags |=
         node::IsolateSettingsFlags::SHOULD_NOT_SET_PROMISE_REJECTION_CALLBACK;
+
+    // We do not want to use the promise rejection callback that Node.js uses,
```
deepak1556, line 507: `promise rejection callback` => `stack trace callback`

codebytere: whoops, thanks!

jkleinsc (approved):

PR #26821, opened 2020-12-03
docs: add description to quick-start.md

windwalkr:
I notice when I tried to run this,  The make fails because it lacks a description.  Adding the Description solves this issue.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


codebytere:
@windwalkr please correct your commit to match the title - it must be semantic and that is why the check is failing.

MarshallOfSound (approved):

windwalkr:
Not a problem,  Looking into how to do that.  May need to cancel the commit and recommit with a different title, as I used the GUI for the merge request, instead of pulling it down to my local machine.

PR #26851, opened 2020-12-07
refactor: define WebContents properties on the prototype

miniak:
#### Description of Change
`BaseWindow` properties are also defined on the prototype. Make it consistent.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes


zcbenz (changes requested):
I'm good with this change, but it seems to cause crashes on all CI.

PR #26874, opened 2020-12-07
chore: remove `app.allowRendererProcessReuse`  and BrowserWindow `affinity` options

MarshallOfSound:
As in title and notes, this is going to be a multiple step process and it'll take a decent amount of refactoring to remove the patch files that we actually want to remove.  But this is step 1 "remove the public API".

Notes: Removed the deprecated `app.allowRendererProcessReuse` and BrowserWindow `affinity` options


nornagon:
Would be good to note in the release notes that `allowRendererProcessReuse` is now always on.

nornagon (changes requested):
needs docs/breaking-changes.

miniak (approved):
Looks good, except for missing `breaking-changes` as mentioned by @nornagon

deepak1556 (changes requested):
Can we do the reverse order of removing the patches first and then flipping the public api bits.

* https://github.com/electron/electron/blob/master/patches/chromium/frame_host_manager.patch
* https://github.com/electron/electron/blob/master/patches/chromium/revert_remove_contentrendererclient_shouldfork.patch
* Navigation controller

Our custom navigation controller should not be used for process reuse mode and also the fork logic in the renderer client. This would ensure the right tests pass when the bits are flipped.

MarshallOfSound:
@deepak1556 Unfortunately not, there's a lot of logic that needs to be unwound here and the first thing we can (and need to) do is remove this public API.  I'll write up an issue shortly outlining all the work that still needs to happen, but this is a very tangled mess and untangling it going to take a lot of time, refactoring and pull requests.  Unfortunately the **final** step will be removing the patches.

hafixo:
Merge pull requested #27156

sentialx:
Does it also mean that Electron will use Chromium's NavigationController instead of its own?

MarshallOfSound:
> Does it also mean that Electron will use Chromium's NavigationController instead of its own?

Long term, yes

nornagon:
@MarshallOfSound 13.x is already in beta, are we sure we want to backport major breaking changes? Can this target 14.x instead?

MarshallOfSound:
> 13.x is already in beta, are we sure we want to backport major breaking changes? Can this target 14.x instead?

Eh sure.  Not super attached to when it goes out, just want to get it into master so I can start destroying our frame host manager patch 😄

docs/api/app.md
```diff
@@ -1426,19 +1426,6 @@ This is the user agent that will be used when no user agent is set at the
 app has the same user agent.  Set to a custom value as early as possible
 in your app's initialization to ensure that your overridden value is used.
 
-### `app.allowRendererProcessReuse`
```
nornagon, line 1429: we should mark this as deprecated in 13.x. did it already make it into docs/breaking-changes as a deprecation?

nornagon (approved):

docs/api/app.md
MarshallOfSound: It's definitely been deprecated since like, Electron 11.  There's been a warning.  I'm not sure if we put deprecations in breaking-changes.  The default flip back in Electron 9 is in the breaking-changes list though.

nornagon: image

We generally mark deprecations in the docs for a feature when we deprecate it, and for major things we put them in breaking changes too.

MarshallOfSound: Hm, I'll put a PR back into 13-x-y to add the @depreacted tag

samuelmaddock (approved):
API LGTM

PR #26875, opened 2020-12-07
fix: restrict sendToFrame to same-process frames by default

nornagon:
#### Description of Change
Fixes an issue where `event.reply` would sometimes not deliver a reply when
OOPIFs were present on a page.

This also changes `webContents.sendToFrame` to only send to same-process frames
by default. Out-of-process frames can be specified by passing a pair of numbers
`[processId, frameId]`.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where `event.reply` could sometimes not deliver a reply to an IPC message when cross-site iframes were present.


shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2718,17 +2718,30 @@ bool WebContents::SendIPCMessageToFrame(bool internal,
...
+    process_id = web_contents()->GetMainFrame()->GetProcess()->GetID();
+  } else {
+    std::vector<int32_t> id_pair;
+    if (gin::ConvertFromV8(isolate, frame, &id_pair) && id_pair.size() == 2) {
+      process_id = id_pair[0];
+      frame_id = id_pair[1];
+    } else {
+      isolate->ThrowException(v8::Exception::Error(gin::StringToV8(
+          isolate,
+          "frameId must be a number or a pair of [processId, routingId]")));
```
deepak1556, line 2733: `[processId, routingId]` => `[processId, frameId]` to be consistent with the docs.

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2718,17 +2718,30 @@ bool WebContents::SendIPCMessageToFrame(bool internal,
...
-  });
-  if (iter == frames.end())
-    return false;
-  if (!(*iter)->IsRenderFrameLive())
+  int32_t frame_id;
+  int32_t process_id;
+  if (gin::ConvertFromV8(isolate, frame, &frame_id)) {
+    process_id = web_contents()->GetMainFrame()->GetProcess()->GetID();
+  } else {
+    std::vector<int32_t> id_pair;
```
deepak1556, line 2726: `std::pair` ?

deepak1556 (approved):
LGTM

shell/browser/api/electron_api_web_contents.cc
nornagon:
```suggestion
          "frameId must be a number or a pair of [processId, frameId]")));
```

shell/browser/api/electron_api_web_contents.cc
nornagon: we have no converter for `std::pair`.

miniak (approved):

lib/browser/api/web-contents.ts
```diff
@@ -148,23 +148,23 @@ WebContents.prototype._sendInternal = function (channel, ...args) {
 
   return this._send(true /* internal */, channel, args);
 };
-WebContents.prototype.sendToFrame = function (frameId, channel, ...args) {
+WebContents.prototype.sendToFrame = function (frame, channel, ...args) {
   if (typeof channel !== 'string') {
     throw new Error('Missing required channel argument');
-  } else if (typeof frameId !== 'number') {
-    throw new Error('Missing required frameId argument');
+  } else if (!(typeof frame === 'number' || Array.isArray(frame))) {
```
miniak, line 154: extract to a helper function maybe?

nornagon: The C++ side will also catch this, so I don't think an extra check is needed. In fact this check isn't really needed either, but I'll leave it because it makes the diff smaller to keep it.

miniak:
API LGTM

nornagon:
CI failures are flakes.

PR #26876, opened 2020-12-08
build: fix JS linting

MarshallOfSound:
* Ensure --fix output is actually written to disk
* Cache bust on lint.js file changes
* Ensure CI does not use the linting cache

Notes: no-notes

### DIFF
--- a/script/lint.js
+++ b/script/lint.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+const crypto = require('crypto');
 const { GitProcess } = require('dugite');
 const childProcess = require('child_process');
 const { ESLint } = require('eslint');
@@ -99,19 +100,24 @@ const LINTERS = [{
   test: filename => filename.endsWith('.js') || filename.endsWith('.ts'),
   run: async (opts, filenames) => {
     const eslint = new ESLint({
-      cache: true,
+      // Do not use the lint cache on CI builds
+      cache: !process.env.CI,
+      cacheLocation: `node_modules/.eslintcache.${crypto.createHash('md5').update(fs.readFileSync(__filename)).digest('hex')}`,
       extensions: ['.js', '.ts'],
       fix: opts.fix
     });
     const formatter = await eslint.loadFormatter();
     let successCount = 0;
-    for (const result of await eslint.lintFiles(filenames)) {
+    const results = await eslint.lintFiles(filenames);
+    for (const result of results) {
       successCount += result.errorCount === 0 ? 1 : 0;
-      if (result.errorCount !== 0 || result.warningCount !== 0) {
-        console.log(formatter.format([result]));
-      } else if (opts.verbose) {
+      if (opts.verbose && result.errorCount === 0 && result.warningCount === 0) {
         console.log(`${result.filePath}: no errors or warnings`);
       }
+    }
+    console.log(formatter.format(results));
+    if (opts.fix) {
+      await ESLint.outputFixes(results);
     }
     if (successCount !== filenames.length) {
       console.error('Linting had errors');
--- a/.gitignore
+++ b/.gitignore
@@ -55,7 +55,7 @@ electron.d.ts
 spec/.hash
 
 # Eslint Cache
-.eslintcache
+.eslintcache*
 
 # Generated native addon files
 /spec-main/fixtures/native-addon/echo/build/


script/lint.js
```diff
@@ -99,20 +100,26 @@ const LINTERS = [{
   test: filename => filename.endsWith('.js') || filename.endsWith('.ts'),
   run: async (opts, filenames) => {
     const eslint = new ESLint({
-      cache: true,
+      // Do not use the lint cache on CI builds
+      cache: !process.env.CI,
+      cacheLocation: `node_modules/.eslintcache.${crypto.createHash('md5').update(fs.readFileSync(__filename)).digest('hex')}`,
```
nornagon, line 105: wait, what? why do we need to base the cache _location_ on the hash of the filename?

why is this in `node_modules/`?

MarshallOfSound: It's in `node_modules` so that it doesn't pollute the top level directory, this is just a safety measure.  There have been times where I've updated this script and the way it calls eslint and it doesn't bust the cache.  Seems like a safe idea.

script/lint.js
```diff
@@ -99,20 +100,26 @@ const LINTERS = [{
...
       extensions: ['.js', '.ts'],
       fix: opts.fix
```
dsanders11, line 107: So I guess the `fix` option only includes the fixes in the output but doesn't write it to disk? Seems like that was the main issue, and fixed by the `ESLint.outputFixes()` call below.

script/lint.js
```diff
@@ -99,20 +100,26 @@ const LINTERS = [{
...
     });
     const formatter = await eslint.loadFormatter();
     let successCount = 0;
+    const results = [];
```
dsanders11, line 111: Maybe just `const results = await eslint.lintFiles(filenames);` instead of rebuilding it?

script/lint.js
```diff
@@ -99,20 +100,26 @@ const LINTERS = [{
...
       successCount += result.errorCount === 0 ? 1 : 0;
-      if (result.errorCount !== 0 || result.warningCount !== 0) {
-        console.log(formatter.format([result]));
-      } else if (opts.verbose) {
+      results.push(result);
+      if (opts.verbose && result.errorCount === 0 && result.warningCount === 0) {
         console.log(`${result.filePath}: no errors or warnings`);
       }
     }
+    formatter.format(results);
```
dsanders11, line 119: I think this still needs to be output with `console.log()` or I think you won't get the output when running the linter.

script/lint.js
```diff
@@ -99,20 +100,26 @@ const LINTERS = [{
...
     const formatter = await eslint.loadFormatter();
     let successCount = 0;
+    const results = [];
     for (const result of await eslint.lintFiles(filenames)) {
       successCount += result.errorCount === 0 ? 1 : 0;
-      if (result.errorCount !== 0 || result.warningCount !== 0) {
-        console.log(formatter.format([result]));
-      } else if (opts.verbose) {
+      results.push(result);
+      if (opts.verbose && result.errorCount === 0 && result.warningCount === 0) {
```
dsanders11, line 115: I'm not sure this change is needed to fix anything. AFAIK `formatter.format()` is only creating the output for human consumption (what you see with the CLI). If there are no warnings or errors, the output (for the default formatter) is blank. Bumping it out to doing all results at once means that it won't be interleaved with the "no errors or warnings" in verbose mode.

script/lint.js
dsanders11: Could the cache be put under `.eslintcache` rather than `node_modules/`? Keeps it more visible and if it stops working as intended there's a better chance it would get caught since it would fall back to `.eslintcache` and if that's a directory, would probably choke on that.

`cacheLocation: `.eslintcache/${crypto.createHash('md5').update(fs.readFileSync(__filename)).digest('hex')}`,` seems to work as expected and makes the `.eslintcache` directory even if it doesn't exist. And everything is still covered by the `.gitignore` pattern.

script/lint.js
MarshallOfSound: Calling `formetter.format` on each individual file results in really hard to results when multiple files have warnings / errors.  This way is consistent with the old output we got by running `eslint` directly.

script/lint.js
MarshallOfSound: I think `node_modules` it the correct place for these, it's where things like typescripts and webpacks cache go as well. It's out of the way and no user should ever need to look at it 🤷

dsanders11:
LGTM

miniak (approved):

PR #26889, opened 2020-12-08
feat: enable world safe JS by default

### LINKED ISSUE #25806: (webContents|webFrame).executeJavaScript don't marshal evaled result promises when worldSafeExecuteJavaScript:true
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 10.1.3
* **Operating System:**
  * win7 sp1 x64

- without worldSafeExecuteJavaScript, `(webFrame|webContents).executeJavaScript('new Promise(resolve => setTimeout(() => resolve(2), 3000));')` works as expected; the executeJavaScript promise waits for the eval promise.
- in the case of webFrame the evaled promise is returned raw
- in the case of webContents the promise is marshaled
- with worldSafeExecuteJavaScript, both error with: "An object could not be cloned"
- webContents should continue marshaling promises, otherwise its an unnecessary breaking change
- webFrame should marshal promises for consistency with webContents
- the clone restriction would be on the promise resolve value

### To Reproduce
`(webFrame|webContents).executeJavaScript('new Promise(resolve => setTimeout(() => resolve(2), 3000));')`

MarshallOfSound:
Fixes #25806

As per the original deprecation message, enabling `worldSafeExecuteJavaScript` by default in Electron 12.  Also plans its removal for Electron 14.

In order to remain backwards-compatible (in regards to promise support and exception handling) I moved the copy-value logic to use the ctx bridge to support non-clonable types and also updated the ctx bridge logic to better handle exception messages

Notes: Changed the default value of `worldSafeExecuteJavaScript` to be `true`.

### DIFF
--- a/docs/breaking-changes.md
+++ b/docs/breaking-changes.md
@@ -20,6 +20,8 @@ In Electron 14, `worldSafeExecuteJavaScript` will be removed.  There is no alter
 ensure your code works with this property enabled.  It has been enabled by default since Electron
 12.
 
+You will be affected by this change if you use either `webFrame.executeJavaScript` or `webFrame.executeJavaScriptInIsolatedWorld`. You will need to ensure that values returned by either of those methods are supported by the [Context Bridge API](api/context-bridge.md#parameter--error--return-type-support) as these methods use the same value passing semantics.
+
 ## Planned Breaking API Changes (13.0)
 
 ### API Changed: `session.setPermissionCheckHandler(handler)`
--- a/docs/breaking-changes.md
+++ b/docs/breaking-changes.md
@@ -12,6 +12,14 @@ This document uses the following convention to categorize breaking changes:
 * **Deprecated:** An API was marked as deprecated. The API will continue to function, but will emit a deprecation warning, and will be removed in a future release.
 * **Removed:** An API or feature was removed, and is no longer supported by Electron.
 
+## Planned Breaking API Changes (14.0)
+
+### Removed: `worldSafeExecuteJavaScript`
+
+In Electron 14, `worldSafeExecuteJavaScript` will be removed.  There is no alternative, please
+ensure your code works with this property enabled.  It has been enabled by default since Electron
+12.
+
 ## Planned Breaking API Changes (13.0)
 
 ### API Changed: `session.setPermissionCheckHandler(handler)`
@@ -96,6 +104,15 @@ session.defaultSession.getAllExtensions()
 Chromium has removed support for Flash, and so we must follow suit. See
 Chromium's [Flash Roadmap](https://www.chromium.org/flash-roadmap) for more
 details.
+
+### Default Changed: `worldSafeExecuteJavaScript` defaults to `true`
+
+In Electron 12, `worldSafeExecuteJavaScript` will be enabled by default.  To restore
+the previous behavior, `worldSafeExecuteJavaScript: false` must be specified in WebPreferences.
+Please note that setting this option to `false` is **insecure**.
+
+This option will be removed in Electron 14 so please migrate your code to support the default
+value.
 
 ### Default Changed: `contextIsolation` defaults to `true`
 
--- a/shell/browser/web_contents_preferences.cc
+++ b/shell/browser/web_contents_preferences.cc
@@ -137,7 +137,7 @@ WebContentsPreferences::WebContentsPreferences(
                 "electron");
   }
   SetDefaultBoolIfUndefined(options::kContextIsolation, false);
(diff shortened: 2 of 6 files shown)


nornagon (approved):

shell/browser/web_contents_preferences.cc
```diff
@@ -137,7 +137,7 @@ WebContentsPreferences::WebContentsPreferences(
                 "electron");
   }
   SetDefaultBoolIfUndefined(options::kContextIsolation, false);
-  SetDefaultBoolIfUndefined(options::kWorldSafeExecuteJavaScript, false);
+  SetDefaultBoolIfUndefined(options::kWorldSafeExecuteJavaScript, true);
```
miniak, line 140: Will it work when this code is executed?
https://github.com/electron/electron/blob/e641a747fb8bc5a7115e4b0581b756aca039b864/shell/browser/api/electron_api_web_contents_view.cc#L112-L113

docs/breaking-changes.md
```diff
@@ -12,6 +12,14 @@ This document uses the following convention to categorize breaking changes:
 * **Deprecated:** An API was marked as deprecated. The API will continue to function, but will emit a deprecation warning, and will be removed in a future release.
 * **Removed:** An API or feature was removed, and is no longer supported by Electron.
 
+## Planned Breaking API Changes (14.0)
+
+### Removed: `worldSafeExecuteJavaScript`
+
+In Electron 14, `worldSafeExecuteJavaScript` will be removed.  There is no alternative, please
+ensure your code works with this property enabled.  It has been enabled by default since Electron
+12.
```
nornagon, line 21: Can we add a link to more info about what `worldSafeExecuteJavaScript` does and how to change your code to compensate?

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -289,6 +273,18 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
...
+    // crosses the bridge we fallback to the v8::Message approach if we can't
+    // pull "message" for some reason
+    v8::MaybeLocal<v8::Value> maybe_message = value.As<v8::Object>()->Get(
+        source_context,
+        gin::ConvertToV8(source_context->GetIsolate(), "message"));
+    v8::Local<v8::Value> message;
+    if (maybe_message.ToLocal(&message) && message->IsString()) {
+      return v8::MaybeLocal<v8::Value>(
+          v8::Exception::Error(message.As<v8::String>()));
+    }
```
nornagon, line 287: This looks new—is it fixing some behavior issue? Should it be separated out and backported to older versions?

nornagon (dismissed):
I feel like the behavior change of switching to use `PassValueToOtherContext` should be separated out and potentially backported to older versions.

Thoughts?

shell/renderer/api/electron_api_context_bridge.cc
MarshallOfSound: Yes, my plan was to extract is separately as a backport but land together with this change.  Otherwise we end up with a chain of 3 or 4 PRs to land to land this 😆

MarshallOfSound:
> I feel like the behavior change of switching to use PassValueToOtherContext should be separated out and potentially backported to older versions.

Switching to use `PassValueToOtherContext` is a thing that I want to backport as well, it's technically a fix IMO.  Effectively my plan is to backport this entire PR without the default change as everything else is fixes.

docs/breaking-changes.md
MarshallOfSound: The docs for `worldSafeExecuteJavaScript` explain "values returned from `webFrame.executeJavaScript` will be sanitized to ensure JS values can't unsafely cross between worlds when using `contextIsolation`".  I'm not sure how much detail we want here.

nornagon: I think this document should contain enough information to:

1. Enable apps to check whether or not their app will be affected by the change, and
2. Give a brief outline of how to adapt to the change.

For (1) we should mention greppable functions (e.g. `webFrame.executeJavaScript`, maybe also `executeJavaScriptInIsolatedWorld`?). For (2) we don't need to go into a lot of detail, maybe just "ensure that objects returned from `executeJavaScript` are serializable"?

shell/browser/web_contents_preferences.cc
miniak: @MarshallOfSound related to #18369

miniak: you need to use `IsEnabled(options::kWorldSafeExecuteJavaScript, true);` as well

codebytere (approved):

ckerr (approved):

PR #26890, opened 2020-12-08
feat: enable context isolation by default

MarshallOfSound:
As per the plan in #23506 and in the breaking changes document, lets enable context isolation by default in Electron 12.

Notes: Changed the default value of `contextIsolation` to `true`.


itsananderson:
Should there be any docs changes included along with this?

shell/browser/web_contents_preferences.cc
```diff
@@ -136,7 +136,7 @@ WebContentsPreferences::WebContentsPreferences(
                 "more information",
                 "electron");
   }
-  SetDefaultBoolIfUndefined(options::kContextIsolation, false);
+  SetDefaultBoolIfUndefined(options::kContextIsolation, true);
```
miniak, line 139: you also need to use `IsEnabled(options::kContextIsolation, true);` check #18369 for details

Randomstand (approved):

codebytere (approved):

codebytere:
Appears the visibility state tests are struggling on macOS

VerteDinde (approved):

codebytere (approved):

jkleinsc (approved):

Issue #10424, opened 2017-09-01
BrowserWindow with customButtonsOnHover and frame:false no longer rounds the corners on Mac


emilylaguna:
* Electron version: 1.7.5
* Operating system: macOS 10.12.6

### Expected behavior
The corners should be rounded like they were in 1.6.x

### Actual behavior
The browser windows corners are no longer rounded when running with electron 1.7.5

### How to reproduce
```
git clone https://github.com/emilylaguna/electron-quick-start.git
cd electron-quick-start
npm install
./node_modules/.bin/electron .
```

envomer:
I have the same issue. Any update on this?

jszaszvari:
Seeing this issue too.

After setting 'customButtonsOnHover' and frame: false

![image](https://user-images.githubusercontent.com/872463/31859696-c57e9f94-b75a-11e7-9851-197a83c22cbd.png)

Looking for a workaround or fix but have not found anything yet.

astoilkov:
I can verify that the issue still exists in Electron 4.0.

bjoernm:
Confirmed still an issue on "electron": "^5.0.4"
When using titleBarStyle: 'customButtonsOnHover', and frame: false, window edges become square, and not rounded.

pronebird:
Suddenly became an issue on Electron 6 as well. 😢

Jip-Hop:
For a multi-window video application I need the corners of my Browser Windows on MacOS not to be rounded. So I'm making use of this bug in Electron 4.2.10. But would be nice if there'd be a choice to round or not. However seems like that's not going to happen: https://github.com/electron/electron/issues/414.

PR #26901, opened 2020-12-09
fix: window with customButtonsOnHover should have rounded corner

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/10424.

The `customButtonsOnHover` option relied on the `NSWindowStyleMaskFullSizeContentView` style to hide window title and buttons, which is actually a weird code path and unnecessary.

This PR makes it use the normal window styles like the other `titleBarStyle` options, which fixes the problem that the window has no rounded corner.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix window with `customButtonsOnHover` option not having rounded corner on macOS.

### DIFF
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -392,10 +392,6 @@ NativeWindowMac::NativeWindowMac(const gin_helper::Dictionary& options,
   }
 
   NSUInteger styleMask = NSWindowStyleMaskTitled;
-  bool customOnHover = title_bar_style_ == TitleBarStyle::kCustomButtonsOnHover;
-  if (customOnHover && (!useStandardWindow || transparent() || !has_frame()))
-    styleMask = NSWindowStyleMaskFullSizeContentView;
-
   if (minimizable)
     styleMask |= NSMiniaturizableWindowMask;
   if (closable)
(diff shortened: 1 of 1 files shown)


xdumsidedx (approved):

xdumsidedx (approved):

xdumsidedx (approved):

xdumsidedx (approved):

xdumsidedx (approved):

xdumsidedx (approved):

xdumsidedx:
💙💚💜🖤

codebytere (approved):

codebytere:
Failure is:

> reporting api sends a report for a deprecation

And is known.

nornagon:
Is it still possible to get a frameless window _without_ rounded corners on macOS after this PR?

zcbenz:
> Is it still possible to get a frameless window _without_ rounded corners on macOS after this PR?

No, but we can add a new API to turn off rounded corner if there is request, we only need to apply the `NSWindowStyleMaskFullSizeContentView` style.

zcbenz:
I think we should hold on backporting it to 10/11, there are some users relying on this option to create a frameless window without rounded corner.

PR #26940, opened 2020-12-11
refactor: create Object.defineProperty() wrapper

miniak:
#### Description of Change
There is a repeating pattern that can be wrapped in a helper function.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/lib/common/property-utils.ts
+++ b/lib/common/property-utils.ts
@@ -1,0 +1,6 @@
+export function defineProperty(target: Object, propertyName: string, getter: () => any, setter?: (value: any) => void) {
+  Object.defineProperty(target, propertyName, {
+    get: () => getter.call(target),
+    set: setter ? (value) => setter.call(target, value) : undefined
+  });
+}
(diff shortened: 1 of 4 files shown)


lib/common/property-utils.ts
```diff
@@ -0,0 +1,6 @@
+export function defineProperty(target: Object, propertyName: string, getter: () => any, setter?: (value: any) => void) {
```
MarshallOfSound, line 1: This is less type-safe than the original approach, I'm not personally sure we need this helper as it just introduces a level of abstraction that feels not-needed.  If we do want it though this should definitely by typed like so.

```suggestion
export function defineProperty<T extends object, K extends keyof T>(target: T, propertyName: K, getter: () => T[K], setter?: (value: T[K]) => void) {
```

nornagon (changes requested):
I think this is needlessly obtuse; I prefer the named arguments of `Object.defineProperty` over the positional arguments of this helper.

PR #26941, opened 2020-12-11
feat: make win.setAspectRatio() work on Windows

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/8036.

This PR makes `win.setAspectRatio` work on Windows. It didn't work previously because the Windows implementation requires a non-zero maximum size, while we provided (0, 0) when maximum size is not set.

There is also a patch added to allow setting aspect ratio to 0 on Windows.

I'm unable to write a test for this because the aspect ratio is not respected when window is resized programmingly with APIs like `win.setSize`, it is also why we did not have a test for it at the first place.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Make `win.setAspectRatio()` work on Windows.

### DIFF
--- a/patches/chromium/enable_reset_aspect_ratio.patch
+++ b/patches/chromium/enable_reset_aspect_ratio.patch
@@ -19,10 +19,10 @@ index a409755330351e7e1684c31f7c7cc6882a2dc7af..3349f38e1df8ff7e5c70f1c177b11914
                                     aspect_ratio.height());
  }
 diff --git a/ui/views/win/hwnd_message_handler.cc b/ui/views/win/hwnd_message_handler.cc
-index 34b58eb81f42e70a17250b99dfaa58891dc076c1..10c584016c427b1969bf707f79e74fa9f7e3e882 100644
+index dc2cdf7467912d36261583c91c8f46fbe041c5e0..a23108d0a38af3388b35bd5a0ae3cc3d5e565d93 100644
 --- a/ui/views/win/hwnd_message_handler.cc
 +++ b/ui/views/win/hwnd_message_handler.cc
-@@ -924,8 +924,11 @@ void HWNDMessageHandler::SetFullscreen(bool fullscreen) {
+@@ -919,8 +919,11 @@ void HWNDMessageHandler::SetFullscreen(bool fullscreen) {
  }
  
  void HWNDMessageHandler::SetAspectRatio(float aspect_ratio) {
--- a/patches/chromium/enable_reset_aspect_ratio.patch
+++ b/patches/chromium/enable_reset_aspect_ratio.patch
@@ -1,0 +1,38 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Cheng Zhao <zcbenz@gmail.com>
+Date: Thu, 4 Oct 2018 14:57:02 -0700
+Subject: feat: enable setting aspect ratio to 0
+
+Make SetAspectRatio accept 0 as valid input, which would reset to null.
+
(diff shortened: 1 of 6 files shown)


samuelmaddock (approved):
API LGTM. A test might be nice for the aspect ratio of 0 edge case to ensure the patch continues to work.

patches/chromium/enable_reset_aspect_ratio.patch
```diff
@@ -0,0 +1,37 @@
...
+
+diff --git a/ui/views/widget/desktop_aura/desktop_window_tree_host_win.cc b/ui/views/widget/desktop_aura/desktop_window_tree_host_win.cc
+index 3b132d4d7618e11b851b315e82fa67442c852fb3..6cee855828fc8069b7cd7d429dcda5dc9bd7ecf8 100644
+--- a/ui/views/widget/desktop_aura/desktop_window_tree_host_win.cc
++++ b/ui/views/widget/desktop_aura/desktop_window_tree_host_win.cc
+@@ -484,7 +484,6 @@ void DesktopWindowTreeHostWin::SetOpacity(float opacity) {
+ }
+ 
+ void DesktopWindowTreeHostWin::SetAspectRatio(const gfx::SizeF& aspect_ratio) {
+-  DCHECK(!aspect_ratio.IsEmpty());
```
jkleinsc, line 16: @zcbenz why are we removing this DCHECK?  if aspect_ratio.IsEmpty() then the next line won't work will it?

zcbenz: `aspect_ratio.IsEmpty()` could also be true when it is something like `(0, 100)`.

codebytere: @zcbenz true, but i think this will crash on divide-by-zero if we don't at least ensure that height > 0, won't it?

zcbenz: Yeah it will crash if height == 0, but I think it should be fine crashing here? If someone pass a 0 height then the code will crash anyway with or without the DCHECK. The purpose of the DCHECK is prevent passing 0 to `SetAspectRatio`, instead of prevent the divide-by-zero crash.

Also we won't trigger the divide-by-zero crash in Electron since we are always passing `(ratio, 100)` to this API.

jkleinsc: @zcbenz I guess to ask the question another way, it is necessary to remove this DCHECK?  What problem does it solve to do so?

zcbenz: The Electron `win.setAspectRatio` API accepts `0` to cancel the aspect ratio, which internally passes `(0, 100)` to `DesktopWindowTreeHostWin::SetAspectRatio`, and it would cause an assertion with this DCHECK since `(0, 100)` is considered as an empty size.

jkleinsc: @zcbenz couldn't we change the DCHECK to 
`DCHECK(!aspect_ratio.height());` to avoid division by zero?

codebytere: I think i'd prefer that as well, since it's a clearer error

zcbenz: @jkleinsc @codebytere Sure, I have changed the DCHECK to validate the height.

jkleinsc (approved):

codebytere (approved):

samuelmaddock:
@electron/wg-releases @zcbenz I'm interested in seeing this work in 11.x.y. Any chance we could get it backported?

PR #26953, opened 2020-12-11
chore: cherry-pick 2d18de63acf1 from chromium

mlaurencin:
Convert strides with padding in skia::SkBitmapToN32OpaqueOrPremul().

Code using bitmaps converted with SkBitmapToN32OpaqueOrPremul() can
easily assume that the pixels are one contiguous (width*4*height)-sized
buffer. If it's not then out-of-bounds read/write can occur.

Also adds tests for SkBitmapToN32OpaqueOrPremul().

R=fmalita@chromium.org

Bug: 1147431, 1144462
Change-Id: I21f7a958a8c9231bf5f052f8ff246f2c249bd70b
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2544032
Commit-Queue: danakj <danakj@chromium.org>
Reviewed-by: Florin Malita <fmalita@chromium.org>
Cr-Commit-Position: refs/heads/master@{#828406}

Notes: Security: backported fix for 1147431, 1144462.


codebytere (approved):

jkleinsc (changes requested):
It looks like on the resolve of conflicts, the change https://chromium.googlesource.com/chromium/src/+/2d18de63acf122e306de0dae9bbef566c71b5f31%5E%21/#F0 was not included.  If that is intentional we should also remove https://chromium.googlesource.com/chromium/src/+/2d18de63acf122e306de0dae9bbef566c71b5f31%5E%21/#F3 because it will never get built.

mlaurencin:
> It looks like on the resolve of conflicts, the change https://chromium.googlesource.com/chromium/src/+/2d18de63acf122e306de0dae9bbef566c71b5f31%5E%21/#F0 was not included. If that is intentional we should also remove https://chromium.googlesource.com/chromium/src/+/2d18de63acf122e306de0dae9bbef566c71b5f31%5E%21/#F3 because it will never get built.

That first change was intentionally not included. The latter test changes have been removed from the patch.

jkleinsc (approved):

PR #26988, opened 2020-12-14
fix: screen EventEmitter methods with remote

codebytere:
Backport of #26809.

See that PR for details.

Notes: Fixed an issue whereby `remote.screen` `EventEmitter` methods are undefined in the renderer.


MarshallOfSound (approved):

codebytere:
Error is not related.

PR #26998, opened 2020-12-14
fix: install Node.js signal handlers in ELECTRON_RUN_AS_NODE

codebytere:
#### Description of Change

In `ELECTRON_RUN_AS_NODE`, Electron should respond to signals the way that
Node.js does. They perform platform-specific initialization for
signal handling in a method which we never invoke indirectly nor which
is directly exposed to embedders. This exposes it so we can perform
that initialization and correctly handle signals like `SIGXFSZ`.

Upstreamed at https://github.com/nodejs/node/pull/36518.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where signal handling was misaligned between Node.js and `ELECTRON_RUN_AS_NODE`.


MarshallOfSound (approved):

nornagon:
Would it be possible to have a more concrete release note? What behavior change might a user notice?

nornagon (approved):

zcbenz (approved):

codebytere:
@nornagon I can try but it's probably hard to coalesce into a single item - it's a lot of small nuances

deepak1556:
Looking at https://github.com/nodejs/node/pull/36518#issuecomment-744676590 , specifically point 3) 

> Replacing missing stdio handles with dummies

We already do it in https://github.com/electron/electron/blob/master/shell/app/electron_main.cc#L80

> Raising the open file descriptor limit

Why do we need this ?

Can we wait to land this until the upstream PR is refactored, so that we can opt-in based on the flags ?

codebytere:
Closing until this is upstreamed.

Issue #26975, opened 2020-12-12
BrowserWindow.addBrowserView doesn't remove BrowserView from old owner window


rumkin:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * v11.1.0
* **Operating System:**
  * –
* **Last Known Working Electron version:**
  * –
* **Place in code:**
  * [atom/browser/api/atom_api_top_level_window.cc:696](https://github.com/electron/electron/blob/5ae3d1a1b2dbe11d3091d366467591d9cb21fdfe/atom/browser/api/atom_api_top_level_window.cc#L696)
  * [atom/browser/api/atom_api_top_level_window.cc:709](https://github.com/electron/electron/blob/5ae3d1a1b2dbe11d3091d366467591d9cb21fdfe/atom/browser/api/atom_api_top_level_window.cc#L709)

### Expected Behavior

When `BrowserWindow#addBrowserView` got called, the BrowserView, passed as argument, receives new BrowserWindow and old BrowserWindow could be closed and has no control over BrowserView anymore.

### Actual Behavior

If BrowserView changes owner window (by `BrowserView#addBrowserView` call) and then old parent window closes, then BrowserView disappears from the new parent window.

### To Reproduce

```js
const {app, BrowserWindow, BrowserView} = require('electron')
const {once} = require('events')

async function main () {
  await app.whenReady()

  const width = 400
  const height = 300

  // Open two windows
  const win1 = new BrowserWindow({
    title: 'Window-1',
    x: 300,
    y: 300,
    width,
    height,
  })

  const win2 = new BrowserWindow({
    title: 'Window-2',
    y: 300,
    x: 300 + width,
    width,
    height,
  })

  // Create BrowserView and append it to Window-1
  const view = new BrowserView()
  win1.addBrowserView(view)
  view.setBounds({
    x: 0, y: 0, width, height,
  })
  // Change BrowserView's owner window to Window-2 (e.g. to move tab between windows)
  win2.addBrowserView(view)

  view.webContents.loadURL('https://electronjs.org')
  await once(view.webContents, 'did-finish-load')

[...]

PR #27000, opened 2020-12-14
fix: handle BrowserView reparenting

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/26975.

Fixes an issue where a `BrowserView` couldn't be reparented properly (moved from one `BrowserWindow` to another) because it wasn't properly detached from the first `BrowserWindow`.

A `BrowserWindow` has a one-to-many relationship with `BrowserViews`, but a `BrowserView` can only have a single owner window. As such, if a `BrowserView` is added to a `BrowserWindow` which has an owner window which is  both not null and _not_ the current window in question, then it needs to be properly detached from the previous owner window.

Tested with https://gist.github.com/a786242649a5d2e462c3328c905f0c9f.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixes an issue where BrowserViews couldn't be effectively reparented.

### DIFF
--- a/shell/browser/api/electron_api_base_window.cc
+++ b/shell/browser/api/electron_api_base_window.cc
@@ -759,6 +759,14 @@ void BaseWindow::AddBrowserView(v8::Local<v8::Value> value) {
     auto get_that_view = browser_views_.find(browser_view->ID());
     if (get_that_view == browser_views_.end()) {
       if (browser_view->web_contents()) {
+        // If we're reparenting a BrowserView, ensure that it's detached from
+        // its previous owner window.
+        auto* owner_window = browser_view->web_contents()->owner_window();
+        if (owner_window && owner_window != window_.get()) {
+          owner_window->RemoveBrowserView(browser_view->view());
+          browser_view->web_contents()->SetOwnerWindow(nullptr);
+        }
+
         window_->AddBrowserView(browser_view->view());
         browser_view->web_contents()->SetOwnerWindow(window_.get());
       }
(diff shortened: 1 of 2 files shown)


MarshallOfSound (commented):
Testable? 🤔

shell/browser/api/electron_api_base_window.cc
```diff
@@ -759,6 +759,14 @@ void BaseWindow::AddBrowserView(v8::Local<v8::Value> value) {
     auto get_that_view = browser_views_.find(browser_view->ID());
     if (get_that_view == browser_views_.end()) {
       if (browser_view->web_contents()) {
+        // If we're reparenting a BrowserView, ensure that it's detached from
+        // its previous owner window.
+        auto* owner_window = browser_view->web_contents()->owner_window();
+        if (owner_window && owner_window != window_.get()) {
+          owner_window->RemoveBrowserView(browser_view->view());
+          browser_view->web_contents()->SetOwnerWindow(nullptr);
+        }
```
nornagon, line 768: getting the owner window via the WebContents feels pretty indirect. Can we get the parent BrowserWindow from the BrowserView more directly?

shell/browser/api/electron_api_base_window.cc
```diff
@@ -1071,9 +1079,12 @@ void BaseWindow::ResetBrowserViews() {
...
                            &browser_view) &&
         !browser_view.IsEmpty()) {
-      if (browser_view->web_contents()) {
-        window_->RemoveBrowserView(browser_view->view());
+      // There's a chance that the BrowserView may have been reparented - only
+      // reset if the owner window is *this* window.
+      auto* owner_window = browser_view->web_contents()->owner_window();
+      if (browser_view->web_contents() && owner_window == window_.get()) {
         browser_view->web_contents()->SetOwnerWindow(nullptr);
+        owner_window->RemoveBrowserView(browser_view->view());
```
deepak1556, line 1087: Does the order of calls matter ? Does the view cleanup code path depend on having a valid browser window ?

i-e Removing view from window before unsetting the window pointer or vice-versa.

shell/browser/api/electron_api_base_window.cc
codebytere: not at present - browserview doesn't currently have a concept of its immediate parent window so we'd be adding a getter for this specific case which seems like a net neutral change given that we already have this pre-existing way to get it 🤔

shell/browser/api/electron_api_base_window.cc
codebytere: it doesn't that I was able to determine through any testing - this is also identical to the codepath we take when removing a `BrowserView` directly via API:

https://github.com/electron/electron/blob/e4c386bbd51a510fb8096e272670c71f0765b843/shell%2Fbrowser%2Fapi%2Felectron_api_base_window.cc#L784-L787

MarshallOfSound (approved):

PR #27001, opened 2020-12-14
refactor: use public node::CallbackScope where possible

codebytere:
#### Description of Change

We should be using the publicly exposed `CallbackScope` where possible, since `InternalCallbackScope` is not public and ideally we should be working to get rid of our patch exposing it. We can't fully get rid of it yet, since our integration with `async_hooks` is broken and we need to specifically skip them when we're adding a callback scope to load the `Environment` in `ELECTRON_RUN_AS_NODE`, but i'll be doing that as a follow up.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/microtasks_runner.cc
+++ b/shell/browser/microtasks_runner.cc
@@ -29,11 +29,9 @@ void MicrotasksRunner::DidProcessTask(const base::PendingTask& pending_task) {
   // up Node.js dealying its callbacks. To fix this, now we always lets Node.js
   // handle the checkpoint in the browser process.
   {
-    auto* node_env = electron::ElectronBrowserMainParts::Get()->node_env();
     v8::HandleScope scope(isolate_);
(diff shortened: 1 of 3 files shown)


deepak1556 (approved):

shell/browser/microtasks_runner.cc
```diff
@@ -29,11 +29,9 @@ void MicrotasksRunner::DidProcessTask(const base::PendingTask& pending_task) {
   // up Node.js dealying its callbacks. To fix this, now we always lets Node.js
   // handle the checkpoint in the browser process.
   {
-    auto* node_env = electron::ElectronBrowserMainParts::Get()->node_env();
```
deepak1556, line 32: node_env getter is unused, can be removed.

zcbenz (approved):

miniak (approved):

PR #27002, opened 2020-12-14
fix: throw when using globalShortcut before ready

### LINKED ISSUE #26986: CHECK when globalShortcut.register called before app.on('ready')
I have something like this:

```
const { app, BrowserWindow, BrowserView, globalShortcut} = require('electron');
...
globalShortcut.register("F5", () => {
    alert();
});
```

Adding this `globalShortcut.register` makes Electron crash on startup.

What's the best way to do an action when F5 is pressed, globally (even if there are other browserviews in the main window)?

nornagon:
#### Description of Change
Fixes #26986.

We should throw in JS instead of CHECK()ing.

No release notes because this module was already not usable before
app.on('ready'), it just changes a CHECK to an exception.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/global-shortcut.md
+++ b/docs/api/global-shortcut.md
@@ -9,7 +9,7 @@ with the operating system so that you can customize the operations for various
 shortcuts.
 
 **Note:** The shortcut is global; it will work even if the app does
-not have the keyboard focus. You should not use this module until the `ready`
+not have the keyboard focus. This module cannot be used before the `ready`
 event of the app module is emitted.
 
 ```javascript
--- a/shell/browser/api/electron_api_global_shortcut.cc
+++ b/shell/browser/api/electron_api_global_shortcut.cc
@@ -13,6 +13,7 @@
 #include "gin/dictionary.h"
 #include "gin/object_template_builder.h"
 #include "shell/browser/api/electron_api_system_preferences.h"
+#include "shell/browser/browser.h"
 #include "shell/common/gin_converters/accelerator_converter.h"
 #include "shell/common/gin_converters/callback_converter.h"
 #include "shell/common/node_includes.h"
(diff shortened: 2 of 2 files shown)


codebytere (approved):

MarshallOfSound (changes requested):
In order to be API consistent shouldn't this just `return false`.  I don't really like introducing exceptions into methods that didn't throw before

codebytere:
@MarshallOfSound i see the concern but i think from a user perspective it's a lot less friendly to fail without giving the user a clear notice as to why. Our docs say this, yes, but that's not always intuitive to the user and to that end i think introducing exceptions to clarify _extant_ behavior is a net positive.

MarshallOfSound:
> Our docs say this, yes, but that's not always intuitive to the user and to that end i think introducing exceptions to clarify extant behavior is a net positive.

Yes, but... Breaking change.  If something that never threw, now throws that's a breaking change IMO.  The "hitting the CHECK" is new.  So if we specifically only backport this to branches that hit the CHECK then we're OK.  If we backport beyond that, then we need to not throw an error on those older branches.

nornagon:
@MarshallOfSound this will only throw in cases which previously `CHECK`ed.

The return value from these functions (`false`) is ignored, as an exception is thrown.

nornagon:
The `CHECK` was added in 2013.

MarshallOfSound:
> To double check, I confirmed that all currently supported versions of electron currently CHECK when using globalShortcut prior to app.on('ready').

Sounds good to me then 👍  Would have crashed, now errors, so it's Better ™️ 😆

nornagon:
@MarshallOfSound switch your ❌  to a ✅  then?

MarshallOfSound (approved):

miniak (approved):

PR #27006, opened 2020-12-15
fix: incorrect case in content::PermissionType mapping

nharper:
#### Description of Change

When mapping content::PermissionType enum values from C++ to javascript, the content::PermissionType::CAMERA_PAN_TILT_ZOOM case statement fell through to the next case, which was FONT_ACCESS. This meant that if the CAMERA_PAN_TILT_ZOOM permission were requested, the JS permission request handler would get called with the string 'font-access' as the permission.

This change changes that mapping so that CAMERA_PAN_TILT_ZOOM corresponds to the string 'media'. Having CAMERA_PAN_TILT_ZOOM be the same string as VIDEO_CAPTURE matches https://source.chromium.org/chromium/chromium/src/+/master:content/browser/devtools/protocol/browser_handler.cc;l=150?q=CAMERA_PAN_TILT_ZOOM%20-f:%5Eout%20lang:cc&start=11.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Apps requesting the CAMERA_PAN_TILT_ZOOM permission will have the permission request handler called with a permission string of "media" instead of "font-access".

### DIFF
--- a/shell/common/gin_converters/content_converter.cc
+++ b/shell/common/gin_converters/content_converter.cc
@@ -190,7 +190,6 @@ v8::Local<v8::Value> Converter<content::PermissionType>::ToV8(
       return StringToV8(isolate, "clipboard-read");
     case content::PermissionType::CLIPBOARD_SANITIZED_WRITE:
       return StringToV8(isolate, "clipboard-sanitized-write");
-    case content::PermissionType::CAMERA_PAN_TILT_ZOOM:
     case content::PermissionType::FONT_ACCESS:
       return StringToV8(isolate, "font-access");
     case content::PermissionType::IDLE_DETECTION:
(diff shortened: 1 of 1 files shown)


nornagon (approved):

nornagon:
Thanks! Can you please update the PR description and include a release note describing the change in a way that would make sense to someone whose app would be impacted by this issue?

nharper:
I added a release note. Are you also looking for a change to the description?

nornagon:
@nharper thanks, your change to the `Notes:` field was exactly what I was requesting :)

deepak1556 (approved):

miniak (approved):

PR #27007, opened 2020-12-15

stewartlord:
#### Description of Change

This PR adds win.setTopBrowserView() so that browser views can be raised without removing them and re-adding them to the window. Raising browser views is necessary to control which view receives pointer events when there are overlapping views. Removing and re-adding a view also works, but has the problem of introducing a noticeable flicker. 

I'm unable to write a test for this because it does not appear possible to test which browser view receives pointer events.

#### Checklist

- [x] `npm test` passes

#### Release Notes


zcbenz (dismissed):
~It seems that this API is essentially `win.removeBrowserView(view) && win.addBrowserView(view)`?~

zcbenz:
I think we should probably allow moving browser view to arbitrary position, like `win.reorderBrowserView(view, index)`, otherwise we might have to add another API to move a BrowserView to bottom in future.

shell/browser/api/electron_api_base_window.cc
```diff
@@ -783,6 +783,25 @@ void BaseWindow::RemoveBrowserView(v8::Local<v8::Value> value) {
...
 }
 
+void BaseWindow::SetTopBrowserView(v8::Local<v8::Value> value) {
+  gin::Handle<BrowserView> browser_view;
+  if (value->IsObject() &&
+      gin::ConvertFromV8(isolate(), value, &browser_view)) {
+    if (!browser_view->web_contents())
+      return;
+    auto get_that_view = browser_views_.find(browser_view->ID());
+    if (get_that_view == browser_views_.end()) {
```
codebytere, line 793: We need to be careful here to ensure that the `BrowserView` which is being raised on a given `BrowserWindow` is not already attached to another window. If for example I call:

```js
  const win1 = new BrowserWindow()
  const win2 = new BrowserWindow()

  const view = new BrowserView()
  win1.addBrowserView(view)
  view.setBounds({ x: 0, y: 0, width: 100, height: 100 })
  win2.addBrowserView(view)
```

and then later call 
```js
win1.setTopBrowserView(view)
```

what behavior would a user reasonably expect to happen? Would this reparent the existing `BrowserView`? (Also see https://github.com/electron/electron/pull/27000 for related behavior)

I would say here that since `setTopBrowserView` makes the implicit assumption that the view _already_ exists on the parent window, that we should not be reparenting it in the event of the above - we should likely instead raise an exception or otherwise act in a way that least surprises the user.

codebytere (changes requested):

shell/browser/api/electron_api_base_window.cc
stewartlord: Good call. I assumed it was safe because of how `addBrowserView()` was written. Seeing your change makes me think we should throw (rather than copying the fix).

stewartlord:
> I think we should probably allow moving browser view to arbitrary position, like `win.reorderBrowserView(view, index)`, otherwise we might have to add another API to move a BrowserView to bottom in future.

Yes this was on my mind was well. Ultimately I went for the simplest possible solution, but am happy to entertain alternatives. 

My hesitation with arbitrary reordering is that I didn't want to take on any additional bookkeeping as I don't believe that the existing `browser_views_` (`std::map`) reflects stacking order. That's not necessarily a problem for Windows and Linux as the underlying Chromium API supports this via `ReorderChildView`. The Mac/NSView API doesn't work that way, but it is doable with a touch more effort.

Do we think users would be confused that they can set the indexes, but `getBrowserViews` doesn't reflect the order they set?

zcbenz:
> Do we think users would be confused that they can set the indexes, but `getBrowserViews` doesn't reflect the order they set?

I think if we allow reordering the views then we should make `getBrowserViews` reflect the order, since this API returns an Array instead of Set.

But yeah it requires additional bookkeeping so it might not really worth the efforts. Not sure how others in @electron/wg-api think.

itsananderson:
Bumped to `semver/minor` since this is adding new functionality

stewartlord:
> > Do we think users would be confused that they can set the indexes, but `getBrowserViews` doesn't reflect the order they set?
> 
> I think if we allow reordering the views then we should make `getBrowserViews` reflect the order, since this API returns an Array instead of Set.
> 
> But yeah it requires additional bookkeeping so it might not really worth the efforts. Not sure how others in @electron/wg-api think.

Does the wg-api group meet regularly to consider questions like this? Just curious what the next steps are. Thanks!

stewartlord:
An alternative change that might provoke less anxiety would be to add a `moveTop` method to `BrowserView` directly. This would align with the moveTop method on BrowserWindow.

zcbenz:
On the API itself, I prefer having APIs supporting arbitrary reordering, but I'm also OK with a simple `setTopBrowserView` API.

sentialx:
@zcbenz @codebytere I addressed the above code reviews

lucasbaze:
As a current user watching this PR, even the simple approach of `setTopBrowserView` would suffice for our use cases! Very much excited for this.

zcbenz (approved):

codebytere (approved):

codebytere:
@stewartlord 👍🏻  from me but if possible can we also add a test to address something like the case in https://github.com/electron/electron/pull/27007#discussion_r543032064?

stewartlord:
> @stewartlord 👍🏻 from me but if possible can we also add a test to address something like the case in [#27007 (comment)](https://github.com/electron/electron/pull/27007#discussion_r543032064)?

Yup, that's a good idea! Do you want to handle that @sentialx or should I? @codebytere Can we also tag this one for backport to 11?

sentialx:
I think I added tests for that case

codebytere:
@sentialx i only see https://github.com/electron/electron/pull/27007/files#diff-e7ecb5ed5bcc44ece4b87b456ac322a3661f7c6e8e8b608a2378f8f579992583R183-R191 which doesn't quite cover the reparent case

sentialx:
@codebytere Is this what you meant?

stewartlord:
I think this one is ready to merge @codebytere and @zcbenz? Maybe? Hopefully!

zcbenz:
The test is crashing:

```
[548:0130/180121.366467:FATAL:view.cc(249)] Check failed: view->parent_ == this (0x1aaff870ec10 vs. 0x1aaff7872e10)
#0 0x564ae89602d9 base::debug::CollectStackTrace()
#1 0x564ae88a8e13 base::debug::StackTrace::StackTrace()
#2 0x564ae88c05bf logging::LogMessage::~LogMessage()
#3 0x564ae88c0f0e logging::LogMessage::~LogMessage()
#4 0x564aed22ef29 views::View::ReorderChildView()
#5 0x564ae4c32ece electron::NativeWindowViews::SetTopBrowserView()
#6 0x564ae4abb969 electron::api::BaseWindow::SetTopBrowserView()
#7 0x564ae4ac5530 gin_helper::Invoker<>::DispatchToCallback()
#8 0x564ae4ac53ea gin_helper::Dispatcher<>::DispatchToCallback()
#9 0x564ae611e407 v8::internal::FunctionCallbackArguments::Call()
#10 0x564ae611cb21 v8::internal::(anonymous namespace)::HandleApiCallHelper<>()
#11 0x564ae611ae35 v8::internal::Builtin_Impl_HandleApiCall()
#12 0x564ae611a91a v8::internal::Builtin_HandleApiCall()
#13 0x564ae725f81f Builtins_CEntry_Return1_DontSaveFPRegs_ArgvOnStack_BuiltinExit
```

sentialx:
@codebytere @zcbenz I think the tests finally work now. The CI errors seem to be unrelated.

redhat:
I can't wait to get this feature

zcbenz (approved):

zcbenz:
Can you try rebasing on the latest master branch? It seems that your branch is based on an slightly old commit, which causes the lint error.

sentialx:
@zcbenz done

sentialx:
@zcbenz Can we backport this to 10?

stewartlord:
@zcbenz I need it in 11 - please and thank you 🙏

nornagon:
@stewartlord @sentialx feature backports to old versions are at the discretion of @electron/wg-releases, who will discuss this at their next meeting.

PR #27040, opened 2020-12-16
chore: put NativeWindowMac's methods in order

zcbenz:
#### Description of Change

The `NativeWindowMac` class has become a bit hard to work on, because its overrides do not match the base class's order, and its method definitions do not match the declarations' order. So I'm putting them in order in this PR.

This will make certain backports harder, but I think it worths the efforts otherwise the class will only become more messy, and I already find it very difficult to add new methods to this class.

There is no actual change to the code except for simple moves.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/shell/browser/native_window.h
+++ b/shell/browser/native_window.h
@@ -236,7 +236,7 @@ class NativeWindow : public base::SupportsUserData,
                            const std::string& display_name);
   virtual void CloseFilePreview();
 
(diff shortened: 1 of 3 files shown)


miniak (approved):

codebytere (approved):

nornagon (approved):

zcbenz:
I'm merging since the `new-pr` label is not removed only due to a bug.

PR #27067, opened 2020-12-17
fix: navigator.setAppBadge/clearAppBadge

### LINKED ISSUE #25141: Renderer crash after calling navigator.setAppBadge(42)

jkleinsc:
#### Description of Change

This PR fixes using `navigator.setAppBadge` and `navigator.clearAppBadge` in Electron.

While #25294 resolved a crash calling `navigator.setAppBadge`, this resolves #25141 by implementing the APIs in Electron.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed using `navigator.setAppBadge` and `navigator.clearAppBadge` in Electron.


shell/browser/badging/badge_manager.cc
```diff
@@ -0,0 +1,81 @@
...
+        "|value| should not be zero when it is |number| (ClearBadge should be "
+        "called instead)!");
+    return;
+  }
+
+  base::Optional<int> value =
+      mojo_value->is_flag() ? base::nullopt
+                            : base::make_optional(mojo_value->get_number());
+
+  electron::Browser::Get()->SetBadgeCount(value);
```
zcbenz, line 74: Should `navigator.setAppBadge` only set badge for the source window instead of all windows in the app?

zcbenz:
Also this PR has some overlap with https://github.com/electron/electron/pull/24051.

shell/browser/badging/badge_manager.cc
jkleinsc: @zcbenz I'm not sure I understand what you are asking here? `navigator.setAppBadge` is not window specific but for the whole app, therefore, I just tied it into [app.setBadgeCount(count)](https://www.electronjs.org/docs/api/app#appsetbadgecountcount-linux-macos).  Am I missing something?

shell/browser/browser_win.cc
```diff
@@ -603,8 +620,93 @@ v8::Local<v8::Promise> Browser::GetApplicationInfoForProtocol(
...
+    // Case 3.
+    badge_alt_string_ =
+        l10n_util::GetStringUTF8(IDS_BADGE_UNREAD_NOTIFICATIONS_UNSPECIFIED);
+    badge_count_ = 0;
+  }
+  return true;
+}
+
+void Browser::UpdateBadgeContents(HWND hwnd) {
+  auto badge = std::make_unique<SkBitmap>();
```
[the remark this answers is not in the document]
jkleinsc: makes sense.  I'll update

shell/browser/ui/win/taskbar_host.cc
```diff
@@ -166,13 +166,12 @@ bool TaskbarHost::SetProgressBar(HWND window,
 }
 
 bool TaskbarHost::SetOverlayIcon(HWND window,
-                                 const gfx::Image& overlay,
+                                 SkBitmap* overlay,
```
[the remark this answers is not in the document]
jkleinsc: makes sense.  I'll update

shell/browser/browser_win.cc
```diff
@@ -603,8 +620,93 @@ v8::Local<v8::Promise> Browser::GetApplicationInfoForProtocol(
...
-bool Browser::SetBadgeCount(int count) {
-  return false;
+bool Browser::SetBadgeCount(base::Optional<int> count) {
+  if (count.has_value() && count.value() == 0) {
+    badge_content_ = base::nullopt;
+  } else {
+    badge_content_ = badging::BadgeManager::GetBadgeString(count);
+  }
+  badge_process_id = GetCurrentProcessId();
+  EnumWindows(WindowsEnumerationBadgeHandler,
```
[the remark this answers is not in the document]
jkleinsc: @zcbenz is there a path to get a HWND from electron::WindowList?  I followed what was done here https://github.com/electron/electron/blob/master/shell/browser/browser_win.cc#L46 because I needed a HWND.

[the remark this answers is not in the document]
zcbenz: You can get the HWND via `GetAcceleratedWidget()` method:

```c++
for (auto* window : WindowList::GetWindows()) {
  HWND hwnd = window->GetAcceleratedWidget();
}
```

I'll do a rewrite of `Browser::Focus` implementation, it is rather outdated code.

shell/browser/badging/badge_manager.cc
zcbenz: On Windows the badge icons are set per-window, and this PR's implementation of `app.setBadgeCount` sets badge icon for all current windows to emulate the app-wide badge icons on macOS.

But to emulate app-wide badge icons we also need to automatically set badge icon for new windows, otherwise the new windows won't have the badges:

```js
app.setBadgeCount(1)
new BrowserWindow()  // window has no badge
```

So to correctly implement `app.setBadgeCount` on Windows we would end up with a rather complex system. And in future if we decide to add a per-window `BrowserWindow.setBadgeCount` API for Windows, this will get more complicated.

I think we don't really need to implement `app.setBadgeCount` on Windows, we can just add an `BrowserWindow.setBadgeCount` API and let users decide which one to use. This would also align with existing APIs, e.g. `app.dock.setIcon` and `BrowserWindow.setOverlayIcon`.

For the `navigator.setAppBadge` API, while it claims to set app badge, but it actually only works in PWA apps that run in standalone windows, so on Windows this API essentially sets badge for current window instead of all windows in the browser:
https://github.com/w3c/badging/blob/master/explainer.md#ux-treatment

jkleinsc: @zcbenz ok I understand what you are asking.  I can change the code to set the badge for all existing windows.  Since the main reason for this PR is to support navigator.setAppBadge/clearAppBadge I could just make that work in Windows and not add Windows support to app.get/setBadgeCount.  Another option would be to document in our documentation that on Windows app.setBadgeCount will only affect existing windows.  What do you think?

zcbenz: I prefer not adding `app.get/setBadgeCount` APIs for Windows, setting same badge for all windows is a weird behavior and I don't think there is any app doing that.

jkleinsc: @zcbenz I removed the changes to app.get/setBadgeCount for Windows, so this PR is now just a fix to make navigator.setAppBadge/clearAppBadge work in Electron.

shell/browser/browser_win.cc
```diff
@@ -584,8 +589,120 @@ v8::Local<v8::Promise> Browser::GetApplicationInfoForProtocol(
...
+    badge_count_ = 0;
+  }
+  for (auto* window : WindowList::GetWindows()) {
+    // Update all the windows
+    HWND hwnd = window->GetAcceleratedWidget();
+    DWORD process_id = 0;
+
+    GetWindowThreadProcessId(hwnd, &process_id);
+    // On Windows set the badge on the first window found for this process.
+    if (process_id == GetCurrentProcessId()) {
```
zcbenz, line 647: There is no need to get and compare `process_id`, `WindowList` only returns windows for current app.

shell/browser/browser.h
```diff
@@ -107,10 +108,21 @@ class Browser : public WindowListObserver {
...
+#else
+  bool SetBadgeCount(base::Optional<int> count);
   int GetBadgeCount();
+#endif
 
 #if defined(OS_WIN)
+  // Used by WindowsEnumerationBadgeHandler to get the window to set badge for
+  DWORD badge_process_id;
+
+  void UpdateBadgeContents(HWND hwnd);
```
zcbenz, line 124: This line can be removed.

shell/browser/browser.h
```diff
@@ -364,6 +376,15 @@ class Browser : public WindowListObserver {
   base::DictionaryValue about_panel_options_;
 #endif
 
+#if defined(OS_WIN)
+  void UpdateBadgeContents(HWND hwnd,
+                           const base::Optional<std::string> badge_content,
+                           const std::string badge_alt_string);
```
zcbenz, line 382: Should use const reference `const std::string&`.

shell/browser/browser.h
```diff
@@ -364,6 +376,15 @@ class Browser : public WindowListObserver {
   base::DictionaryValue about_panel_options_;
 #endif
 
+#if defined(OS_WIN)
+  void UpdateBadgeContents(HWND hwnd,
+                           const base::Optional<std::string> badge_content,
```
zcbenz, line 381: Should use const reference `const base::Optional<std::string>&`.

shell/browser/browser_win.cc
```diff
@@ -584,8 +589,120 @@ v8::Local<v8::Promise> Browser::GetApplicationInfoForProtocol(
...
+    return badge_count_;
+  } else {
+    return 0;
+  }
+}
+
+bool Browser::SetBadgeCount(base::Optional<int> count,
+                            base::Optional<bool> shouldSetBadge) {
+  if (!shouldSetBadge.has_value() || !shouldSetBadge.value()) {
+    // app.setBadgeCount(count) is not support on Windows, but
```
zcbenz, line 603: I think it is fine to make `app.setBadgeCount` silently work on Windows, as long as it is an undocumented thing we are free to do whatever we want. So there is no need to add a special `shouldSetBadge` for Windows.

docs/api/app.md
```diff
@@ -1188,7 +1188,7 @@ On macOS, it shows on the dock icon. On Linux, it only works for Unity launcher.
 **Note:** Unity launcher requires the existence of a `.desktop` file to work,
 for more information please read [Desktop Environment Integration][unity-requirement].
 
-### `app.getBadgeCount()` _Linux_ _macOS_
+### `app.getBadgeCount()`
```
zcbenz, line 1191: This change should be reverted.

zcbenz (approved):

PR #27094, opened 2020-12-21

sentialx:
#### Description of Change

Backport of #26802

#### Release Notes

Notes: none


zcbenz:
According to the last meeting of wg-releases, we don't think this backport needs approval from the working group since it is a patch bump, and the working group approved the backport anyway. https://github.com/electron/electron/pull/26802#issuecomment-747118328

zcbenz (approved):

PR #27095, opened 2020-12-21

sentialx:
#### Description of Change

Backport of #26802

I had to mark `process` as any, since the `mainModule` property has been deprecated and the pre-commit hook wouldn't let me commit changes.

#### Release Notes

Notes: none


zcbenz:
According to the last meeting of wg-releases, we don't think this backport needs approval from the working group since it is a patch bump, and the working group approved the backports anyway.

zcbenz (approved):

PR #27108, opened 2020-12-23
Update to include Arch Linux libraries

markwylde:
#### Description of Change
Add the libraries required to build on Arch Linux / Manjaro

#### Checklist

Notes: none.


codebytere:
@markwylde please correct your commit message per our guidelines - it must have a semantic prefix (ex. `docs:`)

PR #27113, opened 2020-12-23
build: enable the node code cache with GN builds

### LINKED ISSUE #19311: Build the node code cache through GN

MarshallOfSound:
As in title, closes #19311

On my super fast machine it doesn't make a noticeable different (1ms) but on slower low end machines this should result in a nice startup improvement.

Notes: no-notes


jkleinsc (dismissed):
Windows x64 failed to build

MarshallOfSound:
[Enable the node code cache with the Electron build](https://trello.com/c/VGiDdEGr/13-enable-the-node-code-cache-with-the-electron-build)

patches/node/build_add_gn_build_files.patch
```diff
@@ -70,15 +74,11 @@ index 0000000000000000000000000000000000000000..446119163d1f7bad577cb0b7b217ecf2
 +
 +copy("node_js2c_inputs") {
 +  sources = library_files
-+  outputs = [
-+    "$target_gen_dir/js2c_inputs/{{source_target_relative}}",
-+  ]
++  outputs = [ "$target_gen_dir/js2c_inputs/{{source_target_relative}}" ]
```
nornagon, line 77: would it be possible to do the `gn format` in a separate commit? it's hard to see what changed here.

MarshallOfSound: Not now, it's too late I think, I'd have to undo the format manually :rip:

nornagon: can you reformat it as a separate PR, commit that, then merge that in here?

MarshallOfSound: Ah hm, I suspect that will result in some nasty merge conflicts.  Let me try locally

MarshallOfSound: @nornagon The isolated patch file updates are in this commit --> https://github.com/electron/electron/pull/27113/commits/a795072feb8644de73281b8bb21d413273018cf2

nornagon (changes requested):
requesting changes for above comments

lastthyme (approved):

Issue #27131, opened 2020-12-26
Top right part of BrowserView in frameless BrowserWindow is not clickable


quanglam2807:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details
Top right part of BrowserView in frameless BrowserWindow is not clickable

* **Electron Version:** 11.1.1 (broken with >= 11.0.4)
* **Operating System:** Windows & Linux (tested on Windows 10 20H2, reported by a Linux Mint user)
* **Last Known Working Electron version:** 11.0.3

### Expected Behavior

Buttons in `BrowserView` should be clickable.

### Actual Behavior

Buttons in `BrowserView` are not clickable.

### To Reproduce

Replace `main.js` in `electron-quick-start` with the following code.
```js
// Modules to control application life and create native browser window
const {app, BrowserWindow, BrowserView} = require('electron')
const path = require('path')

function createWindow () {
  // Create the browser window.
  const mainWindow = new BrowserWindow({
    width: 800,
    height: 600,
    frame: false,
    webPreferences: {
      preload: path.join(__dirname, 'preload.js')
    }
  })

  const view = new BrowserView({})
  view.webContents.loadURL('https://discordapp.com/login')
  mainWindow.setBrowserView(view)
  view.setBounds({
    x: 0,
    y: 0,
    height: 600,
    width: 800,
  })
  view.setAutoResize({ width: true, height: true })


  // Open the DevTools.
  view.webContents.openDevTools({ mode: 'detach' })
}


// This method will be called when Electron has finished
// initialization and is ready to create browser windows.
// Some APIs can only be used after this event occurs.
app.whenReady().then(() => {
  createWindow()
  
  app.on('activate', function () {
    // On macOS it's common to re-create a window in the app when the
    // dock icon is clicked and there are no other windows open.
    if (BrowserWindow.getAllWindows().length === 0) createWindow()
  })
})

// Quit when all windows are closed, except on macOS. There, it's common
// for applications and their menu bar to stay active until the user quits
// explicitly with Cmd + Q.
app.on('window-all-closed', function () {
  if (process.platform !== 'darwin') app.quit()
})

[...]

PR #27143, opened 2020-12-29
fix: unclickable regions in some draggable BrowserViews

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/27131.
Closes https://github.com/electron/electron/issues/27149.

Fixes an issue where some BrowserViews might exhibit unclickable button behavior when loading some websites. What was happening was that #26738 made the mistaken assumption that snapped areas should be draggable, which meant that when users set `-webkit-app-region: no-drag` the areas would become draggable (and thus not clickable).

Tested with https://gist.github.com/f1426205af6799b5d208d1cd54136457 and confirmed not to regress initial fix made in https://github.com/electron/electron/pull/26738 with https://gist.github.com/7278c8c62ec976110afb47ef2d861a7a.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where some draggable regions were not clickable when loaded into BrowserViews on Windows.

### DIFF
--- a/shell/browser/native_browser_view_views.cc
+++ b/shell/browser/native_browser_view_views.cc
@@ -40,7 +40,6 @@ void NativeBrowserViewViews::UpdateDraggableRegions(
   auto snapped_regions = mojo::Clone(regions);
   for (auto& snapped_region : snapped_regions) {
     snapped_region->bounds.Offset(offset);
-    snapped_region->draggable = true;
   }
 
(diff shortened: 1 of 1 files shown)


jkleinsc (approved):

quanglam2807:
Thank you so much, @codebytere!

My app relied quite a lot on `BrowserView` so this bug is critical. Is it possible to cut a release with this patch this week?

Issue #27112, opened 2020-12-23
Why is the BrowserView being dragged?


ssoju:
### Preflight Checklist

* [v] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [v] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [v] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 11.1.1
* **Operating System:**
  * macOS 11.1
* **Last Known Working Electron version:**
  * 9.0.0

### Expected Behavior
It shouldn't be dragged, it should be a click.

### Actual Behavior
It is dragged in BrowserView and cannot be clicked.

### To Reproduce

```
// main.js
const findInpage = new BrowserView({
  webPreferences: {
    nodeIntegration: true,
  },
});

const [viewWidth] = mainWindow.getContentSize();
mainwindow.setBrowserView(findInPage);
findInPage.setBounds({
  x: viewWidth - 406,
  y: 2,
  width: 402,
  height: 42,
});
findInPage.webContents.loadURL('..find-in-page.html');


// find-in-page.html
<html>
<head><style>...</style></head>
<body style="-webkit-app-region: no-drag;">
<div>
  <input /><span>0/0</span>
  <button>Aa</button>
  <button>up</button>
  <button>down</button>
  <button>close</button>
</div>
</body>
</html>
```

### Screenshots

https://user-images.githubusercontent.com/11021464/103020822-7bd18a00-458c-11eb-99ca-a31b4d837c8f.mov

codebytere:
@taras-slapdash would you mind providing a standalone [Fiddle](https://github.com/electron/fiddle) example that I can verify with?

taras-slapdash:
@codebytere 

Sorry, I couldn't use Electron Fiddle because I didn't find a way to create another HTML file that I could use for `BrowserView`. So here I created a small gist for you – https://gist.github.com/taras-slapdash/9dd56c8dae033bcb77cd107259591121

As I was creating this example, I actually figured out what was causing the problem. It turns out that the presence of `-webkit-app-region: no-drag;` (notice it's `no-drag`) in the `BrowserView` causes it to become draggable and (what's even worse) some of the mouse events are blocked (e.g. CSS styles are ignored when hovering over a button).

[...]

PR #27147, opened 2020-12-29
fix: default offset when no BrowserView drag regions

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/27112.

Fixes an issue where non-draggable regions on `BrowserView`s would have an incorrectly assumed `(x,y)` origin of `(0,0)` which is an assumption that can be made for `BrowserWindow`s but _not_ for `BrowserView`s. In the latter case, we should calculate and pass the origin point of the BrowserView and pass it instead.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixes an issue where non-draggable regions on BrowserViews could have incorrectly calculated bounds.


MarshallOfSound (approved):

codebytere:
Failure:

is not germane, merging.

PR #27156, opened 2020-12-31
docs: coordinate system of badly named method getCursorScreenPoint

jameshfisher:
[Electron inherits this confusing name from Chromium](https://github.com/chromium/chromium/blob/99314be8152e688bafbbf9a615536bdbb289ea87/ui/display/win/screen_win.cc#L677-L681). We can also see there that the return value is a DIPPoint, due to `ScreenToDIPPoint` call:

    gfx::Point ScreenWin::GetCursorScreenPoint() {
      POINT pt;
      ::GetCursorPos(&pt);
      return gfx::ToFlooredPoint(ScreenToDIPPoint(gfx::PointF(gfx::Point(pt))));
    }

I lost over a day due to debugging this. I don't think we can change the method name due to backwards compatibility, but we can at least make amends in the documentation.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/screen.md
+++ b/docs/api/screen.md
@@ -92,8 +92,6 @@ Returns [`Point`](structures/point.md)
 
 The current absolute position of the mouse pointer.
 Despite the name of this method, the return value is a DIP point, not a screen point!
-To get the screen point (for example, to pass to [robotjs](https://github.com/octalmage/robotjs) for mouse control),
-use `dipToScreenPoint` on the return value.
 
 ### `screen.getPrimaryDisplay()`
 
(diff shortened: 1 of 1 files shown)


zcbenz (commented):
Simply noting that the returned point is DIP should be enough, most of Electron's APIs return DIP points and we don't mention `dipToScreenPoint` in every one of them.

jameshfisher:
@zcbenz I've removed the additional advice

zcbenz (approved):

PR #27157, opened 2020-12-31
docs: coordinate system of Display.bounds and Display.workArea

jameshfisher:
#### Description of Change
These properties are in DIP points (rather than screen points). We can use `screen.dipToScreen*` to convert to screen points.

#### Checklist

Notes: none.


codebytere (approved):

codebytere:
@jameshfisher please correct your commit message per our guidelines - it must have a semantic prefix (ex. `doc:`)

PR #27186, opened 2021-01-04
chore: duplicate symbols for mojom interfaces in component build

riccardopiola:
#### Description of Change

When enabling the component build it fails (at least on windows where I tested) due the fact that many mojom interfaces are included both by linking against `blink_common.dll` and by having `//third_party/blink/public/mojom:mojom_core` (source_set) as a dependency of `shell/common/api:mojo`. 
This change mirrors what happens in `//content/public/common/BUILD.gn`, which overrides the dependency on the source set with one on the blink_common dll, thus fixing the duplicate problem.
The override happens only if the component build is enabled so it shouldn't cause any issues for the builds already working.

#### Note
Note that this change does not make it so that you can build with is_component_build=true right after cloning, for that to happen there are a few more problems that I encountered and that need to be fixed as well, nonetheless the one in this PR was by far the hardest to track down and only requires a small change to a single file be fixed.

Other steps for the component build to succeed:
- Add the cflag `"-Wno-macro-redefined"` to `//third-party/zlib:zlib_warnings` (for some reason ZEXTERN is defined the same way both in `//third-party/zlib/chromeconf.h` and `third-party/zlib/zconf.h` and both headers are included in the build.
- Build electron-node as a static library even in component build. (Building it as a shared_library in component build adds some missing imports)

Notes: none

### DIFF
--- a/shell/common/api/BUILD.gn
+++ b/shell/common/api/BUILD.gn
@@ -9,6 +9,11 @@ mojom("mojo") {
     "//third_party/blink/public/mojom:mojom_core",
     "//ui/gfx/geometry/mojom",
   ]
+  
+  # Needed for component build or we'll get duplicate symbols for many mojom
+  # interfaces aready included in blink_common.dll
+  overridden_deps = [ "//third_party/blink/public/mojom:mojom_core" ]
+  component_deps = [ "//third_party/blink/public/common" ]
 
   enabled_features = []
   if (enable_remote_module) {
--- a/shell/common/api/BUILD.gn
+++ b/shell/common/api/BUILD.gn
@@ -9,7 +9,7 @@ mojom("mojo") {
     "//third_party/blink/public/mojom:mojom_core",
     "//ui/gfx/geometry/mojom",
   ]
-  
+
   # Needed for component build or we'll get duplicate symbols for many mojom
   # interfaces aready included in blink_common.dll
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
Electron specifically does not support building with `is_component_build` enabled.  Our CI and local builds all using the `testing.gni` config file which notable does not use component build.  May I ask why you're trying to build with `is_component_build`?

riccardopiola:
I use component build because it's much faster when making small changes. There may be something wrong with my pc but without it enabled i would have to wait 5 minutes for a single file change in //electron/shell, with component build it is much faster (around 30 seconds).

riccardopiola:
Should also add that `is_component_build=true` causes problems (at least on windows) if you also don't enable `is_debug=true` as  that combination is not even supported in chrome i think.

zcbenz (approved):
I'm fine having this in Electron's codebase since it does not make the build configuration more complicated.

I think it is good thing if we can eventually bring back component build, even with goma the final linking step is still taking too much time when building in a virtual machine.

ckerr:
@riccardopiola can you fix the procedural problems with the PR, e.g. fixing the lint warnings, adding a release-notes comment, and using a semantic commit message?

riccardopiola:
I fixed the lint and semantic commit problems, as for the release notes I added a line but I'm not really sure it should even be mentioned as this commit doesn't really enable component build.

Issue #27101, opened 2020-12-21
Make process type transformation optional on calls to win.setVisibleOnAllWorkspaces


CyrusRoshan:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description

(macOS specific)

Prior to https://github.com/electron/electron/pull/24956, the `setVisibleOnAllWorkspaces` only ran `SetCollectionBehavior(visible, NSWindowCollectionBehaviorCanJoinAllSpaces);`, and did not modify the process type. After PR 24956, the process type was changed as well. This makes sense--developers who used Electron would otherwise have to manually hide/show dock, which was unintuitive.

However, the change also caused some dock bugginess, and affected the visibility of other windows. Whereas before this change, you could quickly "pin" and "unpin" a window to a workspace, after this change, your dock might hide/show, your other windows would be hidden. The workarounds I found for this had a worse user experience than the behavior I was able to create prior to the change. 

This behavior makes PR 24956 a breaking change.

### Proposed Solution

Making the `TransformProcessType` call optional, defaulting to `true` on [setVisibleOnAllWorkspaces](https://github.com/electron/electron/blob/cdc94e4cefb1a1565557c77c5a3e186b430ad5ec/shell/browser/native_window_mac.mm#L1273) would solve this, while still allowing basic usage of this function to remain intuitive (no manual dock showing/hiding).

### Alternatives Considered

Alternatively, exposing `SetCollectionBehavior` would allow for reproducing the same behavior `setVisibleOnAllWorkspaces` had prior to the change. I'm not sure how much work this would be, but I think this option is fine as well. 

### Additional Information

An example of how the current behavior can cause issues (note dock behavior, one of the windows goes missing when using `setVisibleOnAllWorkspaces`, which wasn't the case prior to the change): https://github.com/CyrusRoshan/electron-example

CyrusRoshan:
[...]

PR #27200, opened 2021-01-05
feat: add option to not transform processes on win.SetVisibleOnAllWorkspaces

CyrusRoshan:
#### Description of Change

Allows calls to win.SetVisibleOnAllWorkspaces to skip transforming the process type. This allows us to use the same behavior found in Electron versions 9.2.1 and under, which is useful for "pinning" an app to a workspace/desktop, without causing other issues related to process transformation. 

This PR adds optional backwards-compatibility with 9.2.1 and under, while maintaining the same default behavior.

Fixes: (#27101)
Process transformation introduced in this PR: [PR #24956](https://github.com/electron/electron/pull/24956)

Stakeholders:
@codebytere 
@zcbenz 
@MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Allowed skipping process type transformation in win.SetVisibleOnAllWorkspaces on macOS.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -1624,7 +1624,10 @@ Returns `Boolean` - Whether the menu bar is visible.
 * `visible` Boolean
 * `options` Object (optional)
   * `visibleOnFullScreen` Boolean (optional) _macOS_ - Sets whether
-    the window should be visible above fullscreen windows
+    the window should be visible above fullscreen windows.
+  * `skipTransformProcessType` Boolean (optional) _macOS_ - Transforming
+    the process type must hide the window and dock for a short time. This
+    can be skipped if the process type has already been transformed.
 
 Sets whether the window should be visible on all workspaces.
 
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -1625,9 +1625,13 @@ Returns `Boolean` - Whether the menu bar is visible.
 * `options` Object (optional)
   * `visibleOnFullScreen` Boolean (optional) _macOS_ - Sets whether
     the window should be visible above fullscreen windows.
-  * `skipTransformProcessType` Boolean (optional) _macOS_ - Transforming
-    the process type must hide the window and dock for a short time. This
-    can be skipped if the process type has already been transformed.
+  * `skipTransformProcessType` Boolean (optional) _macOS_ - Calling
+    setVisibleOnAllWorkspaces will by default transform the process
+    type between UIElementApplication and ForegroundApplication to
+    ensure the correct behavior. However, this will hide the window
+    and dock for a short time every time it is called. If your window
+    is already of type UIElementApplication, you can bypass this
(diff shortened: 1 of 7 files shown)


zcbenz (approved):

zcbenz:
This PR requires approval from @electron/wg-api since it adds a new option.

CyrusRoshan:
Sweet, thank you! Let me know if there's anything I can do to help.

codebytere (approved):

docs/api/browser-window.md
```diff
@@ -1626,6 +1626,8 @@ Returns `Boolean` - Whether the menu bar is visible.
 * `options` Object (optional)
   * `visibleOnFullScreen` Boolean (optional) _macOS_ - Sets whether
     the window should be visible above fullscreen windows
+  * `skipTransformProcessType` Boolean (optional) _macOS_ - Sets whether
+    the process type should be transformed. Default is `false`
```
jkleinsc, line 1630: While functionally true, this description doesn't seem really helpful to developers.  Can you reword to provide better insight to developers as to why you would want to set this?

jkleinsc (changes requested):
Adding this option looks good to me, but the documentation needs to be clearer as to the functionality it provides

docs/api/browser-window.md
CyrusRoshan: Sure! How does the updated description sound?

docs/api/browser-window.md
```diff
@@ -1625,7 +1625,10 @@ Returns `Boolean` - Whether the menu bar is visible.
 * `visible` Boolean
 * `options` Object (optional)
   * `visibleOnFullScreen` Boolean (optional) _macOS_ - Sets whether
-    the window should be visible above fullscreen windows
+    the window should be visible above fullscreen windows.
+  * `skipTransformProcessType` Boolean (optional) _macOS_ - Transforming
+    the process type must hide the window and dock for a short time. This
+    can be skipped if the process type has already been transformed.
```
jkleinsc, line 1631:
```suggestion
  * `skipTransformProcessType` Boolean (optional) _macOS_ - By default on macOS, setVisibleOnAllWorkspaces transforms the process type in order to hide the window and dock for a short time.  If this is undesirable, setting this value to true will override that behavior.
```

jkleinsc (changes requested):
@CyrusRoshan the updated description is a little more helpful but it is still unclear.  I'm still not 100% clear why we are changing the process type and I think that's the part that needs clarifying.  Maybe including more context from #27101 would be helpful.  I'm trying to think of developers who don't have the context that you do if that makes sense.

jkleinsc (approved):
@CyrusRoshan thanks for updating the description.  It is much clearer now what this option does.

zcbenz:
This PR has got enough approvals from members of wg-api, it is a bug of bot that `api-review/approved ✅ ` label can not be applied.

PR #27217, opened 2021-01-07
refactor: remove path from nativeImage converter

codebytere:
Backport of #26546.

See that PR for details.

Notes: none


miniak (approved):

codebytere:
> session.serviceWorkers getFromVersionID() should report the correct script url an

test is not related.

PR #27225, opened 2021-01-07
refactor: use platform-specific TaskRunner to print

codebytere:
#### Description of Change

We should be using the upstream TaskRunner logic that Chromium does, as seen [here](https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/ui/webui/print_preview/local_printer_handler_default.cc;l=36-57;bpv=1;bpt=1). This fixes that. 

cc @deepak1556 @zcbenz @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -440,6 +440,29 @@ base::string16 GetDefaultPrinterAsync() {
   }
   return base::UTF8ToUTF16(printer_name);
 }
+
+// Copied from
+// chrome/browser/ui/webui/print_preview/local_printer_handler_default.cc:L36-L54
+scoped_refptr<base::TaskRunner> CreatePrinterHandlerTaskRunner() {
+  // USER_VISIBLE because the result is displayed in the print preview dialog.
+#if !defined(OS_WIN)
+  static constexpr base::TaskTraits kTraits = {
+      base::MayBlock(), base::TaskPriority::USER_VISIBLE};
+#endif
+
+#if defined(USE_CUPS)
+  // CUPS is thread safe.
+  return base::ThreadPool::CreateTaskRunner(kTraits);
+#elif defined(OS_WIN)
+  // Windows drivers are likely not thread-safe and need to be accessed on the
+  // UI thread.
+  return content::GetUIThreadTaskRunner(
+      {base::MayBlock(), base::TaskPriority::USER_VISIBLE});
+#else
+  // Be conservative on unsupported platforms.
+  return base::ThreadPool::CreateSingleThreadTaskRunner(kTraits);
+#endif
+}
 #endif
 
 struct UserDataLink : public base::SupportsUserData::Data {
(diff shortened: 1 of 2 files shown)


shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2514,9 +2537,9 @@ void WebContents::Print(gin::Arguments* args) {
     settings.SetIntKey(printing::kSettingDpiVertical, dpi);
   }
 
-  base::ThreadPool::PostTaskAndReplyWithResult(
-      FROM_HERE, {base::MayBlock(), base::TaskPriority::USER_BLOCKING},
-      base::BindOnce(&GetDefaultPrinterAsync),
+  scoped_refptr<base::TaskRunner> tr = CreatePrinterHandlerTaskRunner();
+  base::PostTaskAndReplyWithResult(
+      tr.get(), FROM_HERE, base::BindOnce(&GetDefaultPrinterAsync),
```
deepak1556, line 2542: * The task runner will not be available after this call, since the `base::PostTaskAndReplyWithResult` takes it in as raw pointer. We should store the task runner as a member.

* The call here can instead be

```
print_task_runner_->PostTaskAndReplyWithResult()
```

deepak1556 (approved):
Thanks!

zcbenz (dismissed):
Windows CI are failing and it seems to be related to the changes in this PR since it crashed at print related functions.

shell/browser/api/electron_api_web_contents.cc
codebytere: @deepak1556 i think we do need to use `base::PostTaskAndReplyWithResult` like [upstream](https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/ui/webui/print_preview/local_printer_handler_default.cc;l=136-140;drc=8e5afb8dfb07d9f6f6f7fa5ad65c57cb5927097e;bpv=1;bpt=1)

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -421,8 +424,11 @@ bool IsDeviceNameValid(const base::string16& device_name) {
 }
 
 base::string16 GetDefaultPrinterAsync() {
-  base::ScopedBlockingCall scoped_blocking_call(FROM_HERE,
-                                                base::BlockingType::MAY_BLOCK);
+#if defined(OS_WIN)
+  // Blocking is needed here because Windows printer drivers are oftentimes
+  // not thread-safe and have to be accessed on the UI thread.
+  PrintingScopedAllowBlocking allow_blocking;
```
deepak1556, line 430: Can we instead use `base::ScopedAllowBlocking` instead of inheriting testing class ?

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2514,8 +2546,8 @@ void WebContents::Print(gin::Arguments* args) {
     settings.SetIntKey(printing::kSettingDpiVertical, dpi);
   }
 
-  base::ThreadPool::PostTaskAndReplyWithResult(
-      FROM_HERE, {base::MayBlock(), base::TaskPriority::USER_BLOCKING},
+  base::PostTaskAndReplyWithResult(
```
deepak1556, line 2549: Although upstream uses this method, its actually deprecated and will soon be replaced https://source.chromium.org/chromium/chromium/src/+/master:base/task_runner_util.h;l=34-38

shell/browser/api/electron_api_web_contents.cc
codebytere: @deepak1556 no - see comment at top; we would need to patch the class in Chromium to declare this class a friend class

deepak1556: Missed that, in that case you can use `base::ThreadRestrictions::ScopedAllowIO` which offers same result and we have been using it in our code.

PR #27276, opened 2021-01-12
chore: only do semantic check on PR titles

zcbenz:
It is not necessary requiring commit messages to be semantic, since we always do "squash and merge" so as long as the PR title is semantic the final commit message will be too.

This would make it easier to merge PRs from beginners since lots of them do not understand why the "
Semantic Pull Requests" check keeps failing. 

Notes: none

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,0 +1,2 @@
+# Always validate the PR title, and ignore the commits
+titleOnly: true


ckerr (approved):
The two CI failures look unrelated

codebytere (approved):

dsanders11:
@zcbenz, unfortunately I think there's an edge case this misses. There's been a few recent commits which slipped in without semantic commit messages, like 86e746c36b20c54c8e77d3d65bc9470b247b232c and 12aa991df2acd9e62f5c17325e5e81fa6dfffd74.

According to [this thread](https://github.community/t/how-to-change-the-default-squash-merge-commit-message/1155/49), if there's a single commit, "squash and merge" will use the commit message, but if there's multiple commits, it will use the title of the PR. Since it's a GitHub shortcoming no ideal solution to the issue, but currently non-semantic commits are slipping in as a result. 🙁

zcbenz:
Yeah I noticed it too :(, it seems that we can probably solve it with https://github.com/squash-commit-app/squash-commit-app.

dsanders11:
Unfortunately I don't think that will work for forks, given https://github.com/squash-commit-app/squash-commit-app/issues/5. I'm not sure if GitHub apps can push commits to forks, even if maintainers can, that might be a limitation that makes it hard to use that approach. 😞

dsanders11:
@zcbenz, it looks like GitHub just released a feature which would fix this: [Default to PR titles for squash merge commit messages](https://github.blog/changelog/2022-05-11-default-to-pr-titles-for-squash-merge-commit-messages/)

Can we look at enabling this setting?

Issue #27252, opened 2021-01-09
v12 renderer process crash with `node-pty` native module


Eugeny:
### Issue Details

* **Electron Version:**
  * 12.0.0-beta.12
* **Operating System:**
  * macOS 11.2 x64
* **Last Known Working Electron version:**
  * 11.1.1

### To Reproduce

`require('node-pty').spawn('/bin/sh')`

### Additional Information

Tracebacks seem to vary

#### Traceback 1

```
Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       EXC_I386_GPFLT
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [72809]

Thread 0 Crashed:: CrRendererMain  Dispatch queue: com.apple.main-thread
0   com.github.Electron.framework 	0x0000000103c05e69 v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 65961
1   com.github.Electron.framework 	0x0000000103bfde8f v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 33231
2   com.github.Electron.framework 	0x0000000103bfaf1a v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 21082
3   com.github.Electron.framework 	0x0000000103cecebf v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 1012223
4   com.github.Electron.framework 	0x0000000103bfde8f v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 33231
5   com.github.Electron.framework 	0x0000000103bfde8f v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 33231
6   com.github.Electron.framework 	0x0000000103bfde8f v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 33231
7   com.github.Electron.framework 	0x0000000103bfde8f v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 33231
8   com.github.Electron.framework 	0x0000000103bfde8f v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 33231
[...]

PR #27296, opened 2021-01-13
fix: crash in renderer process with non-context-aware modules

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/27252. 

We need to return `false` when a user attempts to load a non-context-aware native module in the renderer process with `app.allowRendererProcessReuse` enabled. This became an issue in Electron v12 and on because the default of `app.allowRendererProcessReuse` switched to `true`.

Verified by adding `require('node-pty').spawn('/bin/sh')` to an arbitrary renderer process and observing the correct error:

```
"Uncaught Error: Loading non-context-aware native module in renderer: '/private/var/folders/w4/6ckkx75165l174lkz07vgx_m0000gn/T/tmp-64859-HboWDaDRZg4c/node_modules/node-pty/build/Release/pty.node',
```

in console.

cc @deepak1556 @zcbenz 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash in renderer process when loading non-context-aware modules with `app.allowRendererProcessReuse` enabled.


MarshallOfSound (approved):

deepak1556 (approved):

zcbenz (approved):

codebytere:
Failure is known flake:

PR #27348, opened 2021-01-18
feat: add ContextBridgeMutability feature

nikitakot:
#### Description of Change

This PR adds a new `base::Feature` - `ContextBridgeMutability`. When the feature is enabled values that are exposed via context bridge are not freezed and are not made readonly.

**ContextBridgeMutability off**
image

**ContextBridgeMutability on**
image

As discussed with @MarshallOfSound no need to document it.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Adding `ContextBridgeMutability` feature that skips context bridge `DeepFreeze` and `SetReadOnlyNonConfigurable` when exposing a value.

### DIFF
--- a/shell/renderer/api/electron_api_context_bridge.cc
+++ b/shell/renderer/api/electron_api_context_bridge.cc
@@ -11,6 +11,7 @@
 #include <utility>
 #include <vector>
 
+#include "base/feature_list.h"
 #include "base/no_destructor.h"
 #include "base/strings/string_number_conversions.h"
 #include "content/public/renderer/render_frame.h"
@@ -24,6 +25,12 @@
 #include "shell/common/world_ids.h"
 #include "third_party/blink/public/web/web_element.h"
 #include "third_party/blink/public/web/web_local_frame.h"
+
+namespace features {
+
+const base::Feature kContextBridgeMutability{"ContextBridgeMutability",
+                                             base::FEATURE_DISABLED_BY_DEFAULT};
+}
 
 namespace electron {
 
@@ -552,6 +559,12 @@ void ExposeAPIInMainWorld(v8::Isolate* isolate,
     if (maybe_proxy.IsEmpty())
       return;
     auto proxy = maybe_proxy.ToLocalChecked();
+
+    if (base::FeatureList::IsEnabled(features::kContextBridgeMutability)) {
+      global.Set(key, proxy);
+      return;
+    }
+
     if (proxy->IsObject() && !proxy->IsTypedArray() &&
         !DeepFreeze(v8::Local<v8::Object>::Cast(proxy), main_context))
       return;
--- a/spec-main/api-context-bridge-spec.ts
+++ b/spec-main/api-context-bridge-spec.ts
@@ -5,6 +5,7 @@ import * as fs from 'fs-extra';
 import * as http from 'http';
 import * as os from 'os';
 import * as path from 'path';
+import * as cp from 'child_process';
 
 import { closeWindow } from './window-helpers';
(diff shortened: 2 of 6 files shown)


nikitakot:
p.s. - haven't wrote any tests for this, but happy to do it if someone explains me how to do something like `app.commandLine.appendSwitch('enable-features', 'ContextBridgeMutability');` in `spec-main/api-context-bridge-spec.ts`.

MarshallOfSound:
> haven't wrote any tests for this, but happy to do it if someone explains me how to do something like app.commandLine.appendSwitch('enable-features', 'ContextBridgeMutability'); in spec-main/api-context-bridge-spec.ts.

You should be able to do _exactly_ that.

nikitakot:
> > haven't wrote any tests for this, but happy to do it if someone explains me how to do something like app.commandLine.appendSwitch('enable-features', 'ContextBridgeMutability'); in spec-main/api-context-bridge-spec.ts.
> 
> You should be able to do _exactly_ that.

Can you be more concrete please? Calling `app.commandLine.appendSwitch('enable-features', 'ContextBridgeMutability');` in `spec-main/api-context-bridge-spec.ts` seems like too late, the parameter doesn't setup. It works if I put it in `spec-main/index.js` though, but then that parameter is set for all other tests too.

zcbenz:
> Can you be more concrete please? Calling `app.commandLine.appendSwitch('enable-features', 'ContextBridgeMutability');` in `spec-main/api-context-bridge-spec.ts` seems like too late, the parameter doesn't setup. It works if I put it in `spec-main/index.js` though, but then that parameter is set for all other tests too.

You can start a new Electron app and read its output, for example:
https://github.com/electron/electron/blob/master/spec-main/api-menu-spec.ts#L908-L922

nikitakot:
> > Can you be more concrete please? Calling `app.commandLine.appendSwitch('enable-features', 'ContextBridgeMutability');` in `spec-main/api-context-bridge-spec.ts` seems like too late, the parameter doesn't setup. It works if I put it in `spec-main/index.js` though, but then that parameter is set for all other tests too.
> 
> You can start a new Electron app and read its output, for example:
> https://github.com/electron/electron/blob/master/spec-main/api-menu-spec.ts#L908-L922

done

miniak (approved):

nornagon:
What's the reasoning for this? It seems questionably useful to me.

nikitakot:
> What's the reasoning for this? It seems questionably useful to me.

For the case when you're modifying exposed values in a existing codebase and don't wanna refactor it when you enable context isolation.

nikitakot:
> What's the reasoning for this? It seems questionably useful to me.

@nornagon does the reason I posted makes sense to you? we've used this functionality in our electron fork, but I'm ok with abandoning this PR if you don't wanna merge it

nornagon:
Hm, I'm skeptical that this is broadly useful. It's possible to replicate this behavior with a Proxy, which I think is more flexible and clearer (it's surprising, i think, that any modifications aren't reflected back to the isolated world).

I'm OK to merge this behind the flag as it's quite a small change and well-tested, but I don't think this should ever be "sanctioned" behavior.

nikitakot:
> it's surprising, i think, that any modifications aren't reflected back to the isolated world)

why? object in isolated and main worlds are completely different objects, so this part is pretty obvious for me at least

let's merge it then

nornagon:
@nikitakot that follows if you're familiar with the implementation details of contextBridge, but for a naive user I think it's surprising, because in other ways contextBridge seems "seamless", and feels like a regular JS object.

PR #27351, opened 2021-01-18
chore: cherry-pick 9eeb09a73570 from chromium

tosmolka:
[x11] Fixed UAF in drag and drop.

To short-cut round trips to the X server when drag and drop happens
between two Chromium windows, XDragContext stored the raw pointer to
XDragDropClient of the source window in its source_client_ attribute.
This made possible to access the deleted object (use after free) if the
source window had been destroyed during the operation.  In short,
although the target context can call the source client directly via the
shortcut, the PropertyNotify event comes from the X server (not using
the shortcut), and apparently it can come to the target context after
the source window and its client had been destroyed but before the
target context is notified.  See the issue for full details.

Here the XDragContext::source_client_ is removed, and all its uses
are replaced with getting the client from the global map of clients
[1]. The client removes itself from the map upon destruction [2] so
this change eliminates the vulnerability.

For the record, there is the test in the interactive_ui_tests suite
(namely BookmarkBarViewTest22.CloseSourceBrowserDuringDrag) that should
emulate this situation but is has some flaws [3].

[1] https://source.chromium.org/chromium/chromium/src/+/master:ui/base/x/x11_drag_drop_client.cc;l=120
[2] https://source.chromium.org/chromium/chromium/src/+/master:ui/base/x/x11_drag_drop_client.cc;l=200
[3] https://crbug.com/1106379

Bug: 1153595
Change-Id: Ibb875cb4fa04ddfa8f99b39e4dab654048da86c7
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2567229
Reviewed-by: Thomas Anderson <thomasanderson@chromium.org>
Commit-Queue: Alexander Dunaev <adunaev@igalia.com>
Cr-Commit-Position: refs/heads/master@{#833577}

Notes: Security: backported fix for 1153595.

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -125,3 +125,4 @@ merge_m86_ensure_that_buffers_used_by_imagedecoder_haven_t_been.patch
 cherry-pick-2d18de63acf1.patch
 only_zero_out_cross-origin_audio_that_doesn_t_get_played_out.patch
 fix_setparentacessibile_crash_win.patch
+cherry-pick-9eeb09a73570.patch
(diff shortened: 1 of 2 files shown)


MarshallOfSound (approved):

tosmolka:
This issue was probably introduced in 86.0.4218.0, abandoning backport to 10-x-y.

PR #27368, opened 2021-01-19
fix: apply tzdata2020f to ICU

miniak:
#### Description of Change
We got reports of wrong times displaying for customers in Yukon, Canada.
Update the timezone database in ICU to the latest version to fix this.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Updated the ICU time zone database to the latest 2020f version.


deepak1556 (approved):
No version bump icu with update to only timezone data. LGTM for patch bump

zcbenz (approved):

PR #27374, opened 2021-01-19
feat: add Windows fallbackToUser option to OpenExternalOptions

rzhao271:
#### Description of Change

On Windows, if a file does not have a program associated with it, the open call fails with error SE_ERR_NOASSOC. this PR adds the option to make an openas call instead in that case, which opens an `Open With` dialog.

#### Checklist

- [ ] `npm test` passes
- [X] tests are changed (https://github.com/electron/electron/blob/master/docs/development/testing.md)

#### Release Notes

Notes: Added `fallbackToUser` option to OpenExternalOptions for Windows.

### DIFF
--- a/docs/api/shell.md
+++ b/docs/api/shell.md
@@ -40,6 +40,7 @@ Open the given file in the desktop's default manner.
 * `options` Object (optional)
   * `activate` Boolean (optional) _macOS_ - `true` to bring the opened application to the foreground. The default is `true`.
   * `workingDirectory` String (optional) _Windows_ - The working directory.
+  * `fallbackToUser` Boolean (optional) _Windows_ - `true` to open an Open With dialog if the file to be opened does not have a program associated with it. The default is `false`.
 
 Returns `Promise<void>`
 
--- a/docs/api/shell.md
+++ b/docs/api/shell.md
@@ -40,7 +40,7 @@ Open the given file in the desktop's default manner.
 * `options` Object (optional)
   * `activate` Boolean (optional) _macOS_ - `true` to bring the opened application to the foreground. The default is `true`.
   * `workingDirectory` String (optional) _Windows_ - The working directory.
-  * `fallbackToUser` Boolean (optional) _Windows_ - `true` to open an Open With dialog if the file to be opened does not have a program associated with it. The default is `false`.
+  * `fallbackToUser` Boolean (optional) _Windows_ - `true` to show an `Open With` dialog if the file to be opened does not have a program associated with it. The default is `false`.
 
 Returns `Promise<void>`
 
--- a/shell/common/platform_util.h
+++ b/shell/common/platform_util.h
@@ -31,6 +31,7 @@ void OpenPath(const base::FilePath& full_path, OpenCallback callback);
 
 struct OpenExternalOptions {
   bool activate = true;
+  bool fallbackToUser = false;
   base::FilePath working_dir;
 };
 
--- a/shell/common/platform_util.h
+++ b/shell/common/platform_util.h
@@ -31,7 +31,7 @@ void OpenPath(const base::FilePath& full_path, OpenCallback callback);
 
 struct OpenExternalOptions {
   bool activate = true;
-  bool fallbackToUser = false;
+  bool fallback_to_user = false;
   base::FilePath working_dir;
 };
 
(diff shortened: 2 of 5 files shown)


rzhao271:
Related downstream issue: 
https://github.com/microsoft/vscode/issues/84265

shell/common/platform_util.h
```diff
@@ -31,6 +31,7 @@ void OpenPath(const base::FilePath& full_path, OpenCallback callback);
 
 struct OpenExternalOptions {
   bool activate = true;
+  bool fallbackToUser = false;
```
miniak, line 34: should be `fallback_to_user`

shell/common/platform_util_win.cc
```diff
@@ -244,10 +244,17 @@ std::string OpenExternalOnWorkerThread(
   base::string16 escaped_url = L"\"" + base::UTF8ToUTF16(url.spec()) + L"\"";
   base::string16 working_dir = options.working_dir.value();
 
-  if (reinterpret_cast<ULONG_PTR>(
-          ShellExecuteW(nullptr, L"open", escaped_url.c_str(), nullptr,
-                        working_dir.empty() ? nullptr : working_dir.c_str(),
-                        SW_SHOWNORMAL)) <= 32) {
+  ULONG openCallRes = reinterpret_cast<ULONG>(ShellExecuteW(
```
miniak, line 247: should be `open_call_res`

shell/common/platform_util_win.cc
```diff
@@ -244,10 +244,17 @@ std::string OpenExternalOnWorkerThread(
...
-          ShellExecuteW(nullptr, L"open", escaped_url.c_str(), nullptr,
-                        working_dir.empty() ? nullptr : working_dir.c_str(),
-                        SW_SHOWNORMAL)) <= 32) {
+  LONGLONG open_call_res = reinterpret_cast<LONGLONG>(ShellExecuteW(
+      nullptr, L"open", escaped_url.c_str(), nullptr,
+      working_dir.empty() ? nullptr : working_dir.c_str(), SW_SHOWNORMAL));
+  if (options.fallback_to_user && open_call_res == SE_ERR_NOASSOC) {
+    const OPENASINFO info = {escaped_url.c_str(), nullptr,
+                             OAIF_ALLOW_REGISTRATION | OAIF_REGISTER_EXT |
+                                 OAIF_EXEC | OAIF_FILE_IS_URI};
```
deepak1556, line 253: why `OAIF_EXEC` ? we are only delegating to find applications that can open the url, there is no file to execute.

Can this just be `OAIF_ALLOW_REGISTRATION | OAIF_REGISTER_EXT | OAIF_FILE_IS_URI` ?

rzhao271: No. `OAIF_EXEC` is required for Windows to display the proper program select dialog, and then open the link with that selected program. Without this flag, the user only sees a message from Windows telling them they can manually change the file association through Settings > Apps > Default Apps.

spec/ts-smoke/electron/main.ts
```diff
@@ -1056,7 +1056,8 @@ shell.openPath('/home/user/Desktop/test.txt').then(err => {
 })
 
 shell.openExternal('https://github.com', {
-  activate: false
+  activate: false,
+  fallbackToUser: true
```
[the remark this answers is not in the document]
rzhao271: Since the default value of `fallbackToUser` was false, I thought that setting it to true would help with the smoke test more.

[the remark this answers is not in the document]
deepak1556: There is no way we are ensuring that the dialogs are responded here, I would suggest to remove this.

zcbenz (approved):
API LGTM

docs/api/shell.md
```diff
@@ -40,6 +40,7 @@ Open the given file in the desktop's default manner.
 * `options` Object (optional)
   * `activate` Boolean (optional) _macOS_ - `true` to bring the opened application to the foreground. The default is `true`.
   * `workingDirectory` String (optional) _Windows_ - The working directory.
+  * `fallbackToUser` Boolean (optional) _Windows_ - `true` to open an Open With dialog if the file to be opened does not have a program associated with it. The default is `false`.
```
deepak1556, line 43:
```suggestion
  * `fallbackToUser` Boolean (optional) _Windows_ - `true` to show an `Open With` dialog if the file to be opened does not have a program associated with it. The default is `false`.
```

deepak1556 (approved):

rzhao271:
Closing since the downstream issue has been closed (not reproducible).

PR #27376, opened 2021-01-19
fix: increase stack size on windows

deepak1556:
#### Description of Change

Refs https://github.com/microsoft/vscode/issues/112172

Upstream issue https://bugs.chromium.org/p/chromium/issues/detail?id=1164201

Fix is based on https://bugs.chromium.org/p/chromium/issues/detail?id=1164201#c6

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Increase stack size on windows x64 to 8MB

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -1154,6 +1154,19 @@ if (is_mac) {
         ldflags += [ "/guard:cf,nolongjmp" ]
       }
 
+      if (current_cpu == "x86") {
+        # Set the initial stack size to 0.5MiB, instead of the 1.5MiB needed by
+        # Chrome's main thread. This saves significant memory on threads (like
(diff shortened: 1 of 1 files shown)


codebytere (approved):

poiru:
Ooh! We've been getting layout related stack overflow crashes on Windows that I haven't been able to figure out. Hopefully this will take care of it, thanks @deepak1556!

BUILD.gn
```diff
@@ -1154,6 +1154,19 @@ if (is_mac) {
         ldflags += [ "/guard:cf,nolongjmp" ]
       }
 
+      if (current_cpu == "x86") {
```
nornagon, line 1157: Would be great to drop a link to the upstream bit this is copied from!

Issue #28, opened 2013-06-19
Add native method to get version number


probablycorey:
We are using an internal build system that automatically increments the version number. Could we add a native method to atom-shell to return the version number by reading the `Info.plist`?

I considered doing this in javascript, but since different code will be needed on different platforms, I think it makes more sense to put this in atom-shell.

PR #27412, opened 2021-01-20
chore: cherry-pick ffd6ff5a61b9 from v8

ppontes:
Merged: [wasm-simd] Fix loading fp pair registers

We were incorrectly clearing the high reg from the list of regs to load.
The intention was to prevent double (and incorrect) loading - loading
128 bits from the low fp and the loading 128 bits from the high fp.
But this violates the assumption that the two regs in a pair would be
set or unset at the same time.

The fix here is to introduce a new enum for register loads, a nop, which
does nothing. The high fp of the fp pair will be tied to this nop, so as
we iterate down the reglist, we load 128 bits using the low fp, then
don't load anything for the high fp.

Bug: chromium:1161654
(cherry picked from commit 8c698702ced0de085aa91370d8cb44deab3fcf54)

Change-Id: Ib8134574b24f74f24ca9efd34b3444173296d8f1
No-Try: true
No-Presubmit: true
No-Tree-Checks: true
Reviewed-on: https://chromium-review.googlesource.com/c/v8/v8/+/2619416
Commit-Queue: Zhi An Ng <zhin@chromium.org>
Reviewed-by: Clemens Backes <clemensb@chromium.org>
Cr-Commit-Position: refs/branch-heads/8.8@{#28}
Cr-Branched-From: 2dbcdc105b963ee2501c82139eef7e0603977ff0-refs/heads/8.8.278@{#1}
Cr-Branched-From: 366d30c99049b3f1c673f8a93deb9f879d0fa9f0-refs/heads/master@{#71094}

Notes: Security: backported fix for chromium:1161654.

### DIFF
--- a/patches/v8/.patches
+++ b/patches/v8/.patches
@@ -15,3 +15,4 @@ cherry-pick-8c725f7b5bbf.patch
 cherry-pick-146bd99e762b.patch
 cherry-pick-633f67caa6d0.patch
 cherry-pick-290fe9c6e245.patch
+cherry-pick-ffd6ff5a61b9.patch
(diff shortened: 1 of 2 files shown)


codebytere (approved):

zcbenz (approved):

codebytere:
> <webview> tag did-change-theme-color event emits when theme color changes

is not relevant.

Issue #27283, opened 2021-01-12
Fullscreen not working on MacOS


thegnuu:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 12.0.0-beta.14
* **Operating System:**
  * macOS 11.0.1
* **Last Known Working Electron version:**
  * 11.1.1

### Expected Behavior
The fullscreen button should be enabled and work, see this image from electron 11.1.1

![Screenshot 2021-01-12 at 17 02 51](https://user-images.githubusercontent.com/5129086/104339611-14a65600-54f8-11eb-82d9-a7e00aa28643.png)

### Actual Behavior
The fullscreen button is disabled and does not work, see this image from electron 12.0.0-beta.14.

![Screenshot 2021-01-12 at 17 10 10](https://user-images.githubusercontent.com/5129086/104340509-0c9ae600-54f9-11eb-8aa5-38d95b6f77cc.png)

### To Reproduce
Simply take any project and update to Electron 12.0.0-beta.14

codebytere:
I am unable to reproduce this with what you have so far specified.

<img width="894" alt="Screen Shot 2021-01-12 at 7 52 11 PM" src="https://user-images.githubusercontent.com/2036040/104404539-b7b69a00-550f-11eb-9503-b764d6bf53fa.png">

Please provide us with a more complete reproducible sample if you would like for us to attempt to triage further.

thegnuu:
Of course, I played around with a fiddle and just realized that the issue only happens with `titleBarStyle: 'hiddenInset'`. Sorry I missed that before.

Here is an Electron Fiddle: https://gist.github.com/8cb381dc37fcad24cd25dedcb3b118a9

With this you should be able to reproduce it by switching between v11 and v12 (I only tested beta.14)

jkleinsc:
I can verify this issue happens on 12.0.0-beta.14 with the gist above.  FWIW it works in the latest nightly

codebytere:
Probably this: https://github.com/electron/electron/commit/558bcc65bc5615866fdd6bdfe72be0ca42675b90

[...]

PR #27428, opened 2021-01-21
fix: hiddenInset missing maximize button

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/27283.
Refs https://github.com/electron/electron/pull/26901.

Previous we early-exited subsequent logic [here](https://github.com/electron/electron/pull/26901/files#diff-148ef9a1a6f049d9fef8c6272a4871d89315f8fc6cd80dc2d41a28b21b29184cL1714-L1715) for non-normal `titleBarStyle`s, but this change made it such that `SetMaximizable(false);` would be called for `hidden` and `hiddenInset` which would cause those styles to lose their zoom buttons. Fix this by moving that line into other associated logic exclusive to `normal` and `customButtonsOnHover` styles.

cc @jkleinsc @zcbenz  

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where the zoom button was missing for windows with `titleBarStyle: hiddenInset` on macOS.


zcbenz (approved):
Thanks for fixing this!

miniak (approved):

PR #27441, opened 2021-01-21
fix: don't throw on bad icons in BrowserWindow constructor

### LINKED ISSUE #27303: App crashes when icon is set after updating to electron 11.2
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * v11.2.0
* **Operating System:**
  * macOS 10.15.7
  * v11.1.1
  
### Expected Behavior
Application should open when icon is set, if failed to load icon file I expect to get an exception and not SIGSEGV

### Actual Behavior
Application is terminated on the BrowserWindow constructor when _icon_ is supplied

### To Reproduce

* Update to electron v11.2
* Make sure to have the _icon_ property set when creating new instance of _BrowserWindow_
* In the example below I used VSCode icon as an example, I've also tried to copy the file to the project dir, it has no affect.
   This happens with other icns files so the file itself is not important.

```javascript

function createWindow () {
    icon: "/Applications/Visual Studio Code.app/Contents/Resources/Code.icns", // <-- comment to avoid crash, or use electron 11.1.1
    /**
     * Jan 13 13:44:22  Electron Helper[26232] <Error>: read failed
       Jan 13 13:44:22  Electron Helper[26232] <Error>: SeatbeltExecServer: failed to read buffer length.
       Failed to initialize sandbox.
       /Users/idanwork/Projects/electron-icon/node_modules/electron/dist/Electron.app/Contents/MacOS/Electron exited with signal SIGSEGV
     */
   // icon: `${__dirname}/Code.icns`, // Tried to copy the icon locally, didn't help
  })

  win.loadURL('https://www.electronjs.org/docs/api/browser-window');
}

app.whenReady().then(createWindow)
app.on('window-all-closed', () => app.quit())
app.on('activate', () => BrowserWindow.getAllWindows().length === 0 && createWindow());
```

package json
```json
{
  "name": "electron-icon",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "run": "electron ."
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "electron": "^11.2.0"
  }
}
```

### Additional Information

The example assumes you have _Visual Studio Code_ in the _Applications_ folder, you can alter it to use any other icon file.
Reverting the same code to electron v11.1.x works
Error can't be caught and the application is terminated
I've tried using other icon files and put the files in the project dir, the issue persists.

_Edited text in comment_

ckerr:
#### Description of Change

Fixes #27303.

Don't throw exceptions in the BrowserWindow constructor when fed an invalid image.

This change adds a new parameter to TryConvert that specifies what to do on error -- to throw or to warn. The default is to throw.

This change also adds a new protected method SetIconImpl() that is basically the old SetIcon() + that new parameter. This was added instead of appending a switch to SetIcon() because the latter is directly callable as public API and it doesn't make sense to expose this switch. SetIcon() is now a one-liner that calls SetIconImpl() with 'throw' behavior. The BrowserWindow constructor calls SetIconImpl() with 'warn' behavior.

CC @codebytere 

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: fix regression that crashed Electron when processing an invalid icon.

### DIFF
--- a/shell/browser/api/electron_api_base_window.cc
+++ b/shell/browser/api/electron_api_base_window.cc
@@ -101,7 +101,7 @@ BaseWindow::BaseWindow(v8::Isolate* isolate,
 #if defined(TOOLKIT_VIEWS)
   v8::Local<v8::Value> icon;
   if (options.Get(options::kIcon, &icon)) {
-    SetIcon(isolate, icon);
+    SetIconImpl(isolate, icon, NativeImage::OnConvertError::kWarn);
   }
 #endif
 }
@@ -1009,8 +1009,15 @@ bool BaseWindow::SetThumbarButtons(gin_helper::Arguments* args) {
 
 #if defined(TOOLKIT_VIEWS)
 void BaseWindow::SetIcon(v8::Isolate* isolate, v8::Local<v8::Value> icon) {
+  SetIconImpl(isolate, icon, NativeImage::OnConvertError::kThrow);
+}
+
+void BaseWindow::SetIconImpl(v8::Isolate* isolate,
+                             v8::Local<v8::Value> icon,
+                             NativeImage::OnConvertError on_error) {
   NativeImage* native_image = nullptr;
-  if (!NativeImage::TryConvertNativeImage(isolate, icon, &native_image))
+  if (!NativeImage::TryConvertNativeImage(isolate, icon, &native_image,
+                                          on_error))
     return;
 
 #if defined(OS_WIN)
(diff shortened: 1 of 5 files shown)


codebytere (approved):

MarshallOfSound (approved):
Short term this works and makes sense, I've discussed the underlying issue here with a bunch of people though and we really should find a way to solve this generically for all constructors and all theoretical errors.

The problem is specific to super c++ constructors that are bound to JS constructors.  i.e. this case of `BaseWindow` being "supered" to `BrowserWindow`.  There is no way for the C++ `BrowserWindow` constructor to know that `BaseWindow` threw a JS exception and should early terminate, also early termination of a C++ constructor is weird and not easy to follow behavior.

My long term suggestion would be to refactor our gin'ed classes into having an empty constructor, an `IsInitialized()` check (which should be hard `CHECK`'ed in the gin converter) and move all the initialization logic to an `Initialize` method that can return a boolean for success.  This would all be hidden from end users (you'd still call `new BrowserWindow()` but under the hood that should create the new C++ class and call `Initialize` early terminating at each class layer if the parent class `Initialize` failed.  There are probably flaws in this plan but we should definitely figure out a way to do it.

zcbenz (approved):

PR #27446, opened 2021-01-22
fix: CSP with unsafe-eval detection with Trusted Types

### LINKED ISSUE #27211: [Feature Request]: Enable w3c trusted types usage with electron apps

VS Code is in the process of adopting trusted types, see https://github.com/microsoft/vscode/issues/103699 and https://github.com/w3c/webappsec-trusted-types. This is going well and electron plays along nicely. However, there is one place in  `web-view-impl.ts` which directly assign a string to the `innerHTML` property: https://github.com/electron/electron/blob/34156c424cb1964cdd203293cbed138d19ee9c9d/lib/renderer/web-view/web-view-impl.ts#L47

When using trusted types this isn't allows (due to the potential for XSS) and either needs to be avoided or explicitly allowed via a trusted types policy. 

This can reproduced with a renderer like this

```html
<!DOCTYPE html>
<html>
		<meta http-equiv="Content-Security-Policy" content="require-trusted-types-for 'script'; trusted-types *">

  <head>
    <meta charset="UTF-8">
    <title>Hello World!</title>
    <link rel="stylesheet" type="text/css" href="./styles.css">
  </head>
  <body>
    <h1>Hello World!</h1>
    
    <script>
      const webview = document.createElement('webview');
      document.body.append(webview);
    </script>
 
  </body>
</html>
```

There is also a fiddle for the whole thing but that seems to trigger more/other violations that we don't see in VS Code. Might be related to the way electron fiddles work: https://gist.github.com/edd3cf4923d96a613c13954f546dd7d0

miniak:
#### Description of Change
Fixes #27211
Need to use `eval('')` instead of `new Function('')` due to a bug in [Trusted Types for function constructor](https://github.com/w3c/webappsec-trusted-types/wiki/Trusted-Types-for-function-constructor).

Before:
Screen Shot 2021-01-22 at 6 05 07 AM
After:
Screen Shot 2021-01-22 at 6 05 23 AM

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Fixed CSP with `unsafe-eval` detection with Trusted Types.

### DIFF
--- a/lib/renderer/security-warnings.ts
+++ b/lib/renderer/security-warnings.ts
@@ -81,7 +81,7 @@ const isUnsafeEvalEnabled: () => Promise<boolean> = function () {
   // Call _executeJavaScript to bypass the world-safe deprecation warning
   return (webFrame as any)._executeJavaScript(`(${(() => {
     try {
-      new Function(''); // eslint-disable-line no-new,no-new-func
+      eval(window.trustedTypes.emptyScript); // eslint-disable-line no-eval
     } catch {
       return false;
     }
--- a/spec-main/security-warnings-spec.ts
+++ b/spec-main/security-warnings-spec.ts
@@ -22,6 +22,7 @@ describe('security warnings', () => {
   let server: http.Server;
   let w: BrowserWindow;
   let useCsp = true;
+  let useTrustedTypes = false;
   let serverUrl: string;
 
   before((done) => {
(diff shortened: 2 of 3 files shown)


deepak1556 (approved):

lib/renderer/security-warnings.ts
```diff
@@ -81,7 +81,7 @@ const isUnsafeEvalEnabled: () => Promise<boolean> = function () {
   // Call _executeJavaScript to bypass the world-safe deprecation warning
   return (webFrame as any)._executeJavaScript(`(${(() => {
     try {
-      new Function(''); // eslint-disable-line no-new,no-new-func
+      eval((window as any).trustedTypes.emptyScript); // eslint-disable-line no-eval
```
koto, line 84: Nit: To be future-proof (e.g. in case this code ever runs on a browser engine without Trusted Types) you may want to check if `window.trustedTypes` exists, and simply use an empty string instead, e.g. like so:

```
eval((window as any).trustedTypes?.emptyScript ?? '')
```

This relies on the fact that `emptyScript`, while wrapping over an empty string, is truthy.

miniak: Trusted types are supported since Chromium 83, which is Electron 9. Therefore there is no need for a check

zcbenz (approved):

Issue #24427, opened 2020-07-05
ipcRenderer.invoke can't handle custom errors


k3nt0w:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for a feature request that matches the one I want to file, without success.

### Problem Description

The product I'm developing now implements a custom error for gRPC to make the error handling. Then, by passing the additional information for the errors made by gRPC to the renderer, I am trying to do a flexible display of error messages with the renderer.

For example, I am trying to use custom errors in the following situations

- display the ID of data that failed to be received
- Handle custom error codes contracted to the gRPC server

However, [ipcRenderer.invoke](https://github.com/electron/electron/blob/659e79fc08c6ffc2f7506dd1358918d97d240147/lib/renderer/api/ipc-renderer.ts#L24-L30) does not allow to receive custom error generated on main process because this function make simple Error() instance with error message.

```ts
ipcRenderer.invoke = async function (channel, ...args) {
  const { error, result } = await ipc.invoke(internal, channel, args);
  if (error) {
    throw new Error(`Error invoking remote method '${channel}': ${error}`);
  }
  return result;
};
```

### Proposed Solution

I don't fully understand the [ipc.invoke process described in cpp](https://github.com/electron/electron/blob/72a089262e31054eabd342294ccdc4c414425c99/shell/renderer/api/electron_api_ipc_renderer.cc#L105-L129). If ipc.invoke can return a error thrown by the main process directly, I think the following changes will satisfy my wishes.

```ts
ipcRenderer.invoke = async function (channel, ...args) {
  const { error, result } = await ipc.invoke(internal, channel, args);
  if (error) {
    throw error;
  }
  return result;
};
```

However, the information on the channel drops off. I'd like to come up with a solution here when how to handle custom error policy is decieded.

### Additional Information

If it's the current implementation in relation to the cpp part of the implementation, please let me know. I'm glad to see it. I'm also glad that this custom error policy is being used as an anti If it's a pattern, I'd be happy to point that out as well.

tomm1e:
Major issue!

[...]

PR #27461, opened 2021-01-24
docs: Update ipcMain.handle() docs in ipc-main.md for error-handling details

ash0x0:
#### Description of Change
Include information in ipcMain documentation about ipcMain.handle() error handling and workaround provided by @nornagon included in issue #24427.

#### Checklist
- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/ipc-main.md
+++ b/docs/api/ipc-main.md
@@ -120,9 +120,9 @@ The `event` that is passed as the first argument to the handler is the same as
 that passed to a regular event listener. It includes information about which
 WebContents is the source of the invoke request.
 
-Errors thrown through `handle` in the main process are not transparent as they 
-are serialized and only the `message` property from the original error is 
-provided to the renderer process. Please refer to 
+Errors thrown through `handle` in the main process are not transparent as they
+are serialized and only the `message` property from the original error is
+provided to the renderer process. Please refer to
 [#24427](https://github.com/electron/electron/issues/24427) for details.
 
 ### `ipcMain.handleOnce(channel, listener)`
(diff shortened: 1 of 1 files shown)


MarshallOfSound (commented):
> If needed, an unsafe workaround uses a wrapper for `handle` to throw 
custom errors.

I'm not sure we want to explicitly document unsafe workarounds, them existing in issues and such is OK.  But something being in our docs is tantamount to recommending it and I don't think we should recommend this.

docs/api/ipc-main.md
```diff
@@ -120,6 +120,45 @@ The `event` that is passed as the first argument to the handler is the same as
 that passed to a regular event listener. It includes information about which
 WebContents is the source of the invoke request.
 
+Errors thrown through `handle` in the main process are not transparent as they 
+are serialized and only the error message from the is provided to the renderer
```
miniak, line 124: there is something missing here "from the"

ash0x0:
> 
> 
> > If needed, an unsafe workaround uses a wrapper for `handle` to throw
> > custom errors.
> 
> I'm not sure we want to explicitly document unsafe workarounds, them existing in issues and such is OK. But something being in our docs is tantamount to recommending it and I don't think we should recommend this.

Agreed. Just referring the issue should be enough. If it's ok to have a reference to an open issue in the docs that is.

docs/api/ipc-main.md
ash0x0: fixed

nornagon (approved):

zcbenz (approved):

PR #27475, opened 2021-01-25
chore: bump chromium to e836cbe709f3cd703f233de8eb6cc6ec99b72c9d

codebytere:
Updating Chromium to e836cbe709f3cd703f233de8eb6cc6ec99b72c9d (master).

See [all changes in 97adba472d4bf3f97675b5d9eade34526ebf730c..e836cbe709f3cd703f233de8eb6cc6ec99b72c9d](https://chromium.googlesource.com/chromium/src/+log/97adba472d4bf3f97675b5d9eade34526ebf730c..e836cbe709f3cd703f233de8eb6cc6ec99b72c9d?n=10000&pretty=fuller)

Notes: no-notes


deepak1556:
Failing subframe tests are related to refactor from https://chromium-review.googlesource.com/c/chromium/src/+/2601063 ,  haven't gotten to investigating the fix. Next upgrade master, please look into it.

shell/browser/electron_browser_context.cc
```diff
@@ -431,9 +434,33 @@ void ElectronBrowserContext::SetCorsOriginAccessListForOrigin(
...
+                              base::RetainedRef(bc_setter.get())));
+    }
+  }
+
+  // Keep the per-profile access list up to date so that we can use this to
+  // restore NetworkContext settings at anytime, e.g. on restarting the
+  // network service.
+  scoped_refptr<content::SharedCorsOriginAccessList>
+      shared_cors_origin_access_list =
+          content::SharedCorsOriginAccessList::Create();
```
deepak1556, line 460: Browser context should maintain a reference to the access list and it must be published for the content layer to pick it up, three changes are needed here:

1) `shared_cors_origin_access_list` should be a member of `ElectronBrowserContext` https://github.com/electron/electron/blob/e4bcb81f47abe2651399c4d1c7763079d92f1574/shell/browser/electron_browser_context.cc#L112

2) `ElectronBrowserContext` should implement `GetSharedCorsOriginAccessList` https://github.com/electron/electron/blob/e4bcb81f47abe2651399c4d1c7763079d92f1574/shell/browser/electron_browser_context.cc#L465

3) Network context must be initialized with this list https://github.com/electron/electron/blob/e4bcb81f47abe2651399c4d1c7763079d92f1574/shell/browser/net/network_context_service.cc#L85

shell/browser/lib/bluetooth_chooser.cc
```diff
@@ -50,6 +50,11 @@ void BluetoothChooser::SetAdapterPresence(AdapterPresence presence) {
   switch (presence) {
     case AdapterPresence::ABSENT:
     case AdapterPresence::POWERED_OFF:
+    // Chrome currently directs the user to system preferences
+    // to grant bluetooth permission for this case, should we
+    // do something similar ?
+    // https://chromium-review.googlesource.com/c/chromium/src/+/2617129
+    case AdapterPresence::UNAUTHORIZED:
```
jkleinsc, line 57: I think we should fire an event in this case, but it can be handled outside this PR.

jkleinsc (approved):

shell/renderer/electron_api_service_impl.cc
```diff
@@ -134,24 +143,20 @@ void ElectronApiServiceImpl::Message(bool internal,
...
-  if (!document_created_)
+  // See: https://chromium-review.googlesource.com/c/chromium/src/+/2601063.
+  if (!document_created_) {
+    PendingElectronApiServiceMessage message;
+    message.internal = internal;
+    message.channel = channel;
+    message.arguments = std::move(arguments);
+    message.sender_id = sender_id;
+
+    pending_messages_.push(std::move(message));
```
nornagon, line 157: Can we just delay binding the mojo pipe until the document has been created? Unbound mojo pipes queue messages internally, I believe.

shell/browser/electron_browser_context.cc
```diff
@@ -431,9 +436,32 @@ void ElectronBrowserContext::SetCorsOriginAccessListForOrigin(
...
     base::OnceClosure closure) {
-  // TODO(nornagon): actually set the CORS access lists. This is called from
-  // extensions/browser/renderer_startup_helper.cc.
-  base::ThreadTaskRunnerHandle::Get()->PostTask(FROM_HERE, std::move(closure));
+  auto& context_map = ElectronBrowserContext::browser_context_map();
+  auto barrier_closure =
+      BarrierClosure(1 + context_map.size(), std::move(closure));
+
+  for (auto& iter : context_map) {
+    if (iter.second) {
```
nornagon, line 444: i'm not sure this can ever be `false`? `context_map` is `std::map<PartitionKey, std::unique_ptr<ElectronBrowserContext>>`, so `iter.second` is `std::unique_ptr`. I don't think it's ever null in our usage?

nornagon (changes requested):
see comments

ckerr (commented):
NB: this is mooted by @nornagon's suggestion of buffering these messages in mojo, which IMO should be Plan A. I wrote the items below _after_ seeeing that :slightly_smiling_face: 

The new `service_` member field only remembers the last-opened RenderFrame's service and so it's possible that pre-create messages will never get pumped in the current implementation if more than one RenderFrame is created.

There's also a possibility of messages being processed out-of-order `ElectronApiServiceImpl::DidCreateDocumentElement()` is  called before `RendererClientBase::RunScriptsAtDocumentStart()` due to how `ElectronApiServiceImpl::document_created_` is used. Possibly `ProcessPendingMessages()` should be made private and called by `ElectronApiServiceImpl::DidCreateDocumentElement()` -- this would also remove the need for the new `service_` field, preventing the previous paragraph's potential pitfall

shell/browser/electron_browser_context.cc
ckerr: A DCHECK might be a better way of signalling intent then; "this shouldn't ever be nullptr" is less ambiguous than "if we have a pointer"

shell/renderer/electron_api_service_impl.h
```diff
@@ -19,11 +20,24 @@ namespace electron {
...
+struct PendingElectronApiServiceMessage {
+  bool internal;
+  std::string channel;
+  blink::CloneableMessage arguments;
+  int32_t sender_id;
+
+  PendingElectronApiServiceMessage() = default;
+  ~PendingElectronApiServiceMessage() = default;
+  PendingElectronApiServiceMessage(PendingElectronApiServiceMessage&&) =
+      default;
```
ckerr, line 32: Are these `= default` necessary?

nornagon (approved):
lgtm on behalf of @codebytere

PR #27477, opened 2021-01-25
refactor: use C++11 class member variable initialization

miniak:
#### Description of Change
Follow up to #13030

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/shell/browser/api/electron_api_base_window.h
+++ b/shell/browser/api/electron_api_base_window.h
@@ -274,7 +274,7 @@ class BaseWindow : public gin_helper::TrackableObject<BaseWindow>,
   // Reference to JS wrapper to prevent garbage collection.
   v8::Global<v8::Value> self_ref_;
 
-  base::WeakPtrFactory<BaseWindow> weak_factory_;
+  base::WeakPtrFactory<BaseWindow> weak_factory_{this};
 };
 
 }  // namespace api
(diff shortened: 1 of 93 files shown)


shell/renderer/api/electron_api_spell_check_client.cc
```diff
@@ -73,8 +73,7 @@ class SpellCheckClient::SpellcheckRequest {
 SpellCheckClient::SpellCheckClient(const std::string& language,
                                    v8::Isolate* isolate,
                                    v8::Local<v8::Object> provider)
-    : pending_request_param_(nullptr),
```
miniak, line 76: https://github.com/electron/electron/blob/ed126eced457fcaa3c998b32f97d68576e7f362c/shell/renderer/api/electron_api_spell_check_client.h#L100

shell/browser/api/electron_api_base_window.h
```diff
@@ -274,7 +274,7 @@ class BaseWindow : public gin_helper::TrackableObject<BaseWindow>,
   // Reference to JS wrapper to prevent garbage collection.
   v8::Global<v8::Value> self_ref_;
 
-  base::WeakPtrFactory<BaseWindow> weak_factory_;
+  base::WeakPtrFactory<BaseWindow> weak_factory_{this};
```
ckerr, line 277: NB for reviewers: since the parent class [is initialized before](https://stackoverflow.com/a/40633475) these members during construction, this should be safe

ckerr (approved):
:heart:

shell/browser/api/electron_api_base_window.h
miniak: also we've been using this construct in some classes already

codebytere (approved):

ckerr:
@miniak does it make sense to backport these to avoid code shear between supported branches?

PR #27481, opened 2021-01-25
fix: remove windowName set as title for native window.open()

mlaurencin:
#### Description of Change

This PR fixes the behavior where the Electron window.open() parameter _frameName_ is displayed as the title of the window. According to the [native window.open() documentation](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#parameters) (which has the corresponding parameter _windowName_) "this will NOT be used as the window's title", so the parameter is no longer set as the title.

Gist: https://gist.github.com/mlaurencin/c2445005bc1ae9c317ce219c260f1d6e

Initially discovered while fixing issue closed by https://github.com/electron/electron/pull/27521

BREAKING CHANGE

#### Checklist

- [x] `npm test` passes
- [x] tests are changed or added

#### Release Notes

Notes: Fixed so window.open() parameter _frameName_ is no longer set as window title


lib/browser/guest-window-manager.ts
```diff
@@ -211,12 +211,25 @@ function makeBrowserWindowOptions ({ embedder, features, frameName, overrideOpti
 
   const deprecatedInheritedOptions = getDeprecatedInheritedOptions(embedder);
 
+  console.log(frameName);
```
codebytere, line 214: Some extra console logs here :)

ckerr (approved):
Code looks fine. Fix works. :+1:

MarshallOfSound (dismissed):

codebytere (approved):

codebytere:
@mlaurencin you'll need to go through and remove title checks from affected tests - looks like there are ~10 of them or so

codebytere:
Some thoughts here - i went back and looked at loc's PR and saw [this line](https://github.com/electron/electron/pull/23130/files#diff-9d9b85d928f49155cfeb1ed13a3dc60d57d8e9de1a091dc2648055c557b2f3fbL62-L64) which was then refactored to [this line](https://github.com/electron/electron/pull/23130/files#diff-04a8a22dc52239d972ceaec63f4e50704158913f92e1e41dd9f37fc0deb5236dR228) and realized that they're not functionally equivalent. The former explicitly checks for `null` while the latter sets the title as `frameName` for any _falsy_ value - this would make it a default since it would be `undefined` unless set. So - I think there are two issues here, and we should decide whether to handle them in the same PR or split them into two. 

The first issue, which is against spec as noted above, is that the framename-as-title used to be optional and now isn't, which can be fixed by re-adding an explicit null check. That would be semver-minor. 

The second issue is that we should never at _any point_ likely be setting the `frameName` as the title - however, we've apparently been doing this in some form for a long time, and it's possible that users rely on this. If we remove this, it would be a breaking change and thus `semver-major`.

Personally, I think our best approach here may be to split the two above issues into two PRs. What do you all think?

mlaurencin:
> Some thoughts here - i went back and looked at loc's PR and saw [this line](https://github.com/electron/electron/pull/23130/files#diff-9d9b85d928f49155cfeb1ed13a3dc60d57d8e9de1a091dc2648055c557b2f3fbL62-L64) which was then refactored to [this line](https://github.com/electron/electron/pull/23130/files#diff-04a8a22dc52239d972ceaec63f4e50704158913f92e1e41dd9f37fc0deb5236dR228) and realized that they're not functionally equivalent. The former explicitly checks for `null` while the latter sets the title as `frameName` for any _falsy_ value - this would make it a default since it would be `undefined` unless set. So - I think there are two issues here, and we should decide whether to handle them in the same PR or split them into two.
> 
> The first issue, which is against spec as noted above, is that the framename-as-title used to be optional and now isn't, which can be fixed by re-adding an explicit null check. That would be a `semver-patch` change.
> 
> The second issue is that we should never at _any point_ likely be setting the `frameName` as the title - however, we've apparently been doing this in some form for a long time, and it's possible that users rely on this. If we remove this, it would be a breaking change and thus `semver-major` change.
> 
> Personally, I think our best approach here may be to split the two above issues into two PRs. What do you all think?

This sounds reasonable to me. If everyone is in agreement, I can open up a new PR with the null checking as the `semver-patch` version, and continue updating the tests on this PR as the `semver-major` version. And also then have the new PR be the one to close out the referred to issue instead of this one, since it would be merged before.

spec-main/chromium-spec.ts
```diff
@@ -744,20 +744,14 @@ describe('chromium features', () => {
...
-      expect(window.getTitle()).to.equal('hello');
-    });
-
     it('does not throw an exception when the frameName is a built-in object property', async () => {
       const w = new BrowserWindow({ show: false });
       w.loadURL('about:blank');
       w.webContents.executeJavaScript('{ b = window.open(\'\', \'__proto__\'); null }');
-      const [, window] = await emittedOnce(app, 'browser-window-created');
-      expect(window.getTitle()).to.equal('__proto__');
+      const [, , frameName, , test] = await emittedOnce(w.webContents, 'new-window');
```
codebytere, line 751: this issue is caused by https://github.com/eslint/eslint/issues/9760 - given that `test` is unused here and it looks like you're console logging it i think we should just disable the rule for this line and comment above:

```js
// https://github.com/eslint/eslint/issues/9760
// eslint-disable-next-line array-bracket-spacing
const [, , frameName, ,] = await emittedOnce(w.webContents, 'new-window');
```

MarshallOfSound: I don't think we shouldn't need to disable this lint rule? The `console.log` should be removed and then this line just becomes

```suggestion
      const [, , frameName] = await emittedOnce(w.webContents, 'new-window');
```

mlaurencin: Oh great, I wasn't aware that the return value could work in that way, but it looks to be functioning. Thank you

codebytere (approved):

MarshallOfSound (approved):

Issue #7073, opened 2016-09-02
persistent cookies should be encrypted


ccnokes:
This request may be unfeasible or out of scope, in which case, feel free to just close/delete. But no one seems to have brought this up before so I thought I'd mention it. 

My org's security team recently recommended that we encrypt the cookie value of our app's session cookie. They don't like that someone could open the sqlite cookies file and view the session token in plain text. Our web app uses a session cookie as it's authentication mechanism, so I'm stuck with storing that token in a cookie. Because Chrome encrypts cookie values by default, it would be nice to be able to encrypt cookie values in Electron too. If not the whole cookie store, then maybe having some middleware on `session` that gets run before `get`/`set` so that I can encrypt/decrypt it on get/set--that way it's persisted encrypted, but to the web app, it's unencrypted. 

If there's a way to do this currently that I'm overlooking, let me know 😸 .

zcbenz:
@deepak1556 After merging the fix OS X would emit a warning dialog every time when Electron runs:

<img width="546" alt="screen shot 2016-09-19 at 4 24 21 pm" src="https://cloud.githubusercontent.com/assets/639601/18624871/af751bc4-7e85-11e6-8ffb-512e2de44284.png">

After a deep looking it seems that the change would just read and store from the same security storage with Chromium browser (and all other Electron apps), which is going to cause problems. And it also makes existing apps lose cookies after updating Electron. 

So I'm reverting the change and reopening this issue.

deepak1556:
Should have asked for help to verify on OS X, there are two things we can do either override `CookieCryptoDelegate` with our version of encrypt and decrypt functions or provide an api in `session.cookies` to let users provide their version of the functions.

tmaher:
[Reading over the high-level docs for Keychain](https://developer.apple.com/library/content/documentation/Security/Conceptual/keychainServConcepts/02concepts/concepts.html#//apple_ref/doc/uid/TP30000897-CH204-SW11), specifically the "Sharing Keychain Items" section, it looks like code signing is required to retain keychain access across recompiles. Specifically, the `keychain-access-group` entitlement seems to be governing. Unfortunately, this is at about the edge of my xcode experience. I did find these, which might help:

1. Walkthrough of "Configuring Keychain Sharing" in xcode: https://developer.apple.com/library/content/documentation/IDEs/Conceptual/AppDistributionGuide/AddingCapabilities/AddingCapabilities.html#//apple_ref/doc/uid/TP40012582-CH26-SW15
[...]

PR #27524, opened 2021-01-27
feat: add experimental cookie encryption support

MarshallOfSound:
This is mostly exploratory but the intention is to provide an Electron Fuse to enable cookie encryption for Electron apps.  This has several technical challenges (mostly coming down to hard coded assumptions in Chromium) and the intention is for this WIP PR to remain open while I figure out how to make it work on all three platforms in a generic and sustainable way.

Refs #7073

Platform support as of the moment you're reading this:
* [x] macOS
* [x] Windows 7
* [x] Windows 10
* [x] Linux (many variants, only ticked when all chromium secret stores are supported)

Migration paths from **Un-Encrypted** to **Encrypted** is completely handled by Chromium, it will continue to read un-encrypted cookies but will encrypt-on-write so over time even existing cookies will either expire or be written-over with an encrypted value.

Going the other way (encryp -> unencrypt) is not supported and will be a lossy action (all cookies will be considered invalid or be garbled in HTTP requests).  This "revert" story should be documented before landing.

By default Chromium uses values like `Chromium Safe Storage` to store the encryption keys, Electron will use `${app.getName()} Safe Storage`.  This results in `Electron Safe Storage` being the default and packaged apps will have `My App Safe Storage`.

Notes: Added experimental cookie encryption support behind an [Electron Fuse](https://www.electronjs.org/docs/tutorial/fuses)


nornagon:
Cool stuff :)

MarshallOfSound:
[Enable cookie encryption in Electron](https://trello.com/c/pplrnuyU/5-enable-cookie-encryption-in-electron)

patches/chromium/make_keychainpassword_service_name_and_service_name_mutable_strings.patch
```diff
@@ -0,0 +1,193 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Samuel Attard <sattard@slack-corp.com>
+Date: Wed, 27 Jan 2021 15:43:35 -0800
+Subject: make KeychainPassword::service_name, kApplicationName and
+ service_name mutable strings
+
+This should be upstreamable, it's just a type change that makes it
```
nornagon, line 7: Can you try upstreaming this before we land please? :)

MarshallOfSound: This patch comment isn't completely accurate since I added the patches required for linux.  The macOS bit in isolation I can for sure try to upstream, the linux bits probably need a bit of thinking through before trying to upstream 🤔

shell/browser/browser_process_impl.cc
```diff
@@ -83,15 +89,32 @@ BuildState* BrowserProcessImpl::GetBuildState() {
...
+  OSCrypt::RegisterLocalPrefs(pref_registry.get());
+#endif
+
   auto pref_store = base::MakeRefCounted<ValueMapPrefStore>();
   ApplyProxyModeFromCommandLine(pref_store.get());
   prefs_factory.set_command_line_prefs(std::move(pref_store));
-  prefs_factory.set_user_prefs(new OverlayUserPrefStore(new InMemoryPrefStore));
+
+  // Only use a persistent prefs store when cookie encryption is enabled as that
+  // is the only key that needs it
```
nornagon, line 104: what does the cookie encryption code use this for?

build/fuses/fuses.json5
```diff
@@ -2,5 +2,7 @@
   "_comment": "Modifying the fuse schema in any breaking way should result in the _version prop being incremented.  NEVER remove a fuse or change its meaning, instead mark it as removed with 'r'",
   "_schema": "0 == off, 1 == on, r == removed fuse",
   "_version": 1,
-  "run_as_node": "1"
+  "run_as_node": "1",
+  // Cookie Encryption is currently only supported on macOS and Windows
+  "cookie_encryption": "1"
```
nornagon, line 7: does this absolutely have to be a fuse? could it be a feature instead?

shell/browser/net/system_network_context_manager.cc
```diff
@@ -217,6 +232,41 @@ void SystemNetworkContextManager::OnNetworkServiceCreated(
...
       network_context_.BindNewPipeAndPassReceiver(),
       CreateNetworkContextParams());
+
+  if (electron::fuses::IsCookieEncryptionEnabled()) {
+    std::string app_name = electron::Browser::Get()->GetName();
+#if defined(OS_MAC)
+    KeychainPassword::service_name = app_name + " Safe Storage";
+    KeychainPassword::account_name = app_name;
+#endif
+#if defined(OS_LINUX)
```
nornagon, line 242:
```suggestion
#elif defined(OS_LINUX)
```

build/fuses/fuses.json5
MarshallOfSound: @nornagon For codesigning reasons, yes IMO.  A feature flag is controlled via JS (and therefore not enforced via code signing), a fuse is enforced for code signing and for something security-related like cookie encryption I think it makes more sense to be a fuse

nornagon: I'm not sure this makes sense, if you're worried about an attack which modifies JS, then such an attack could easily steal cookies post-decryption. Protecting against JS modification needs a different approach (e.g. https://github.com/electron/electron/pull/24470#issuecomment-663166146)

shell/browser/browser_process_impl.cc
MarshallOfSound: On windows the encryption secret is protected with `DPAPI` and the "protected" encryption key is stored in plain text in `Local State` on disk.

build/fuses/fuses.json5
MarshallOfSound: > Protecting against JS modification needs a different approach

Correct, but we can't do that yet.  This is IMO a reasonable middle ground that dramatically raises the bar of someone else turning this off / messing with it.  We also still don't have a _nice_ way of toggling `base::Feature` flags from JS, `appendSwitch` is not super ideal 😆

patches/chromium/make_keychainpassword_service_name_and_service_name_mutable_strings.patch
MarshallOfSound: Both patches have been completely upstreamed and accepted

shell/browser/browser_process_impl.cc
```diff
@@ -83,15 +89,32 @@ BuildState* BrowserProcessImpl::GetBuildState() {
...
+  if (electron::fuses::IsCookieEncryptionEnabled()) {
+    base::FilePath prefs_path;
+    CHECK(base::PathService::Get(electron::DIR_USER_DATA, &prefs_path));
+    prefs_path = prefs_path.Append(FILE_PATH_LITERAL("Local State"));
+    base::ThreadRestrictions::ScopedAllowIO allow_io;
+    scoped_refptr<JsonPrefStore> user_pref_store =
+        base::MakeRefCounted<JsonPrefStore>(prefs_path);
+    user_pref_store->ReadPrefs();
+    prefs_factory.set_user_prefs(user_pref_store);
```
nornagon, line 113: Does this mean that when cookie encryption is enabled we might store other, non-cookie-encryption-related prefs on disk?

shell/browser/electron_browser_main_parts.cc
```diff
@@ -539,6 +541,16 @@ void ElectronBrowserMainParts::PreMainMessageLoopStartCommon() {
...
 #endif
   media::SetLocalizedStringProvider(MediaStringProvider);
+
+#if defined(OS_WIN)
+  if (electron::fuses::IsCookieEncryptionEnabled()) {
+    auto* local_state = g_browser_process->local_state();
+    DCHECK(local_state);
+
+    bool os_crypt_init = OSCrypt::Init(local_state);
+    DCHECK(os_crypt_init);
```
nornagon, line 551: I feel like this should maybe be higher than a `DCHECK`?

MarshallOfSound: c.f. https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/chrome_browser_main_win.cc;l=583-584?q=OSCrypt::Init&ss=chromium

shell/browser/browser_process_impl.cc
MarshallOfSound: I guess so yeah, I would have made it JSON prefs for all cases but that felt like a bit of an overreach for this PR.  It would align better with Chromium to do that though

nornagon: What preferences are in that group? What would we start storing on disk that previously wasn't persisted?

shell/browser/electron_browser_main_parts.cc
nornagon: hm, ok.

shell/browser/browser_process_impl.cc
MarshallOfSound: My understanding is the only other pref we register into this store is the `kProxy` pref which is overridden / ignored by the per-browser-context proxy settings later down the line.  That pref is never read / set in our code so being in memory vs stored to disk wouldn't appear to have any affected.

nornagon (approved):

shell/browser/browser_process_impl.cc
deepak1556: The reason we use an in-memory pref for the proxy config is because we reuse proxy config code from `//chrome` https://github.com/electron/electron/blob/master/patches/chromium/proxy_config_monitor.patch to interact with the proxy service and that relies on reading the configs from a pref store https://source.chromium.org/chromium/chromium/src/+/main:components/proxy_config/pref_proxy_config_tracker_impl.cc, I didn't find the need the flush these values to disk at that time as they are mainly required only to communicate the data between the browser process and the proxy service.

deepak1556: If the pref store is change to a persistent one when cookie encryption is enabled we should also make relevant changes here https://github.com/electron/electron/blob/master/shell/browser/api/electron_api_session.cc#L536

shell/browser/net/system_network_context_manager.cc
```diff
@@ -218,6 +228,39 @@ void SystemNetworkContextManager::OnNetworkServiceCreated(
...
+      const base::CommandLine& command_line =
+          *base::CommandLine::ForCurrentProcess();
+
+      network::mojom::CryptConfigPtr config =
+          network::mojom::CryptConfig::New();
+      config->application_name = app_name;
+      config->product_name = app_name;
+      // c.f.
+      // https://source.chromium.org/chromium/chromium/src/+/master:chrome/common/chrome_switches.cc;l=689;drc=9d82515060b9b75fa941986f5db7390299669ef1
+      config->store = command_line.GetSwitchValueASCII("password-store");
```
deepak1556, line 254: We build `chrome_switches` as part for `//chromium_src:chrome` target, so better to use the constant keys instead to avoid any breakage.

shell/browser/browser_process_impl.cc
MarshallOfSound: @deepak1556 The BrowserProcess prefs store and the BrowserContext prefs store are different and don't need to be aligned afaik.  They store things in different locations and having one in memory and one not seems fine to me

jkleinsc (commented):
Is there anywhere we can document this feature?  It doesn't seem that this feature is very discoverable as is.

shell/browser/browser_process_impl.cc
```diff
@@ -83,15 +89,32 @@ BuildState* BrowserProcessImpl::GetBuildState() {
...
 
 void BrowserProcessImpl::PostEarlyInitialization() {
-  // Mock user prefs, as we only need to track changes for a
-  // in memory pref store. There are no persistent preferences
   PrefServiceFactory prefs_factory;
   auto pref_registry = base::MakeRefCounted<PrefRegistrySimple>();
   PrefProxyConfigTrackerImpl::RegisterPrefs(pref_registry.get());
+#if defined(OS_WIN)
+  OSCrypt::RegisterLocalPrefs(pref_registry.get());
+#endif
```
samuelmaddock, line 97: Should this also be behind the added fuse?

samuelmaddock (approved):
API LGTM

shell/browser/browser_process_impl.cc
MarshallOfSound: In order to better support toggling the fuse I left this registered all the time (otherwise the secret pref is deleted on next launch and all existing cookies are lost)

shell/browser/browser_process_impl.cc
deepak1556: You are right, they don't need to be in sync.

Proxy config in BrowserProcess pref store are read by the SystemNetworkContextManager whereas the BrowserContext pref store are read by per-session network contexts.

deepak1556 (approved):

theogravity:
It's unclear to me what the behavior is when you have support enabled and the user denies the prompt for keychain access in MacOS.

Do the cookies get stored unencrypted then? Is there a way for us to know that the user has denied this access?

PR #27546, opened 2021-01-29
refactor: load preload script directly as a string

miniak:
#### Description of Change
Might improve performance slightly

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes


nornagon (approved):

ppontes (approved):

petef19:
Hello,

looking to load preload script directly inline (as a string) in main.js instead of from an external file.

Is this what this PR does ?

If so, how can I use it ? What is the syntax ?

I'm on Electron 12.0.5, Win 10 x64.

Thanks.

Issue #414, opened 2014-06-20
Remove Border Radius From Frameless Window On Mac


Subash:
Fameless window on mac contains border radius on both top and bottom, is there any option to remove that ?
![prepros_app](https://cloud.githubusercontent.com/assets/3478693/3336788/4a182816-f83b-11e3-82ad-e15a6ed3467c.png)

zcbenz:
The problem is it is actually quite hard to remove the bottom rounded corners, there is an upstream bug tracked at:
https://code.google.com/p/chromium/issues/detail?id=355412

captbaritone:
Looking to use Electron to build an OSx version of https://github.com/captbaritone/winamp2-js, but I would need to be able to turn off rounded corners for it to look good.

Seems like it might not be a solvable problem, but I wanted to add another example where non-rounded corners might be useful.

Cheapskate01:
I'm actually having the exact opposite problem. I would like the corners to round on windows and linux. Can anybody help me out?

mikaa123:
A hack that I used to remove the rounded border was to make the window transparent, then use a css `clip-path` property.

KevCJones:
A bit of a bump but wondering if anyone found a non-hack way of doing this?

gabrielstuff:
@mikaa123 how do you handle the click through the transparent window ? A real life example of border radius issue is when you want to display content on a video projector or a LED screen or any other projector.
Real life example, you want a frameless window in 500x500 and you have 10 of them.
You think you will simply start a script to place each window side by side and let the app start.
Well, no you can't, you are forced to put all of your app in one application and manage your tiled window inside one app. This sucks : /

mikaa123:
@gabrielstuff I don't it's possible to let a click go through a window, even if it's transparent

ivmelo:
Is there a way to remove only the top rounded corners? I am building a tray app and it kinda looks off with the top rounded corners.

brockb:
I really need square corners as well. I'm doing screen capture of a video in a chromeless window, let's say placed at X=0, Y=0, and the screen capture is picking up those rounded corners. I need to be able to capture every pixel of the video (including the missing pixels in the corners). There has to be a way to make it where border radius can be set. I really need a radius of zero.

MondoGao:
Leave the corners away!
You can get a radius-less window-like window by:
```javascript
  win = new BrowserWindow({
    frame: false, 
    transparent: true,
  });
```
```css
html {
[...]

PR #27572, opened 2021-02-01
feat: add roundedCorners option for BrowserWindow

zcbenz:
#### Description of Change

Add an option to remove rounded corners of frameless windows on macOS, close https://github.com/electron/electron/issues/414.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Add `roundedCorners` option for `BrowserWindow`.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -228,7 +228,7 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       experimental.
   * `trafficLightPosition` [Point](structures/point.md) (optional) - Set a
     custom position for the traffic light buttons in frameless windows.
-  * `roundedCorner` Boolean (optional) - Whether frameless window should have
+  * `roundedCorners` Boolean (optional) - Whether frameless window should have
     rounded corners on macOS. Default is `true`.
   * `fullscreenWindowTitle` Boolean (optional) _Deprecated_ - Shows the title in
     the title bar in full screen mode on macOS for `hiddenInset` titleBarStyle.
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -228,6 +228,8 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       experimental.
   * `trafficLightPosition` [Point](structures/point.md) (optional) - Set a
     custom position for the traffic light buttons in frameless windows.
+  * `roundedCorner` Boolean (optional) - Whether frameless window should have
+    rounded corners on macOS. Default is `true`.
   * `fullscreenWindowTitle` Boolean (optional) _Deprecated_ - Shows the title in
     the title bar in full screen mode on macOS for `hiddenInset` titleBarStyle.
     Default is `false`.
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -314,7 +314,7 @@ NativeWindowMac::NativeWindowMac(const gin_helper::Dictionary& options,
   // The NSWindowStyleMaskFullSizeContentView style removes rounded corners
   // for framless window.
   bool rounded_corner = true;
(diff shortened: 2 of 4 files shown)


samuelmaddock (commented):
BrowserWindows with vibrancy options seem to apply their own corner radius size. Would this new option be compatible with this behavior?
https://github.com/electron/electron/blob/95e5193c0297e728de1f532d8fde952262185e52/shell/browser/native_window_mac.mm#L1242-L1258

Would it be possible to instead provide a `cornerRadius` option where the amount could be specified? If it were `0`, rounded corners could be disabled. I'm not too familiar with macOS development to know whether this is supported or not.

docs/api/browser-window.md
```diff
@@ -228,6 +228,8 @@ It creates a new `BrowserWindow` with native properties as set by the `options`.
       experimental.
   * `trafficLightPosition` [Point](structures/point.md) (optional) - Set a
     custom position for the traffic light buttons in frameless windows.
+  * `roundedCorner` Boolean (optional) - Whether frameless window should have
```
miniak, line 231: wouldn't `roundedCorners` be a better name?

zcbenz:
> BrowserWindows with vibrancy options seem to apply their own corner radius size. Would this new option be compatible with this behavior?

Thanks for noticing it, I have fixed the code to work with vibrancy views.

> Would it be possible to instead provide a `cornerRadius` option where the amount could be specified? If it were `0`, rounded corners could be disabled. I'm not too familiar with macOS development to know whether this is supported or not.

No, the referenced code is used to set corner radius of the vibrancy view so it would not cover the window's rounded corners, macOS does not provide a way to change the corner radius of the window.

> wouldn't roundedCorners be a better name?

Yeah it sounds a better name, I have changed the option's name.

samuelmaddock (approved):

miniak (approved):

codebytere (approved):

ldenoue:
I noticed that when using this property `rounderCorners: true` when creating a BrowserWindow with a Tray, my app and windows never close.

giacomocerquone:
> I noticed that when using this property `roundedCorners: true` when creating a BrowserWindow with a Tray, my app and windows never close.

yes, a repro has been created and the issue is tracked here: https://github.com/electron/electron/issues/28686

I tried to have a look at the code, but honestly, I'm no obj-c macOS dev sadly, and couldn't figure out much.s
@zcbenz would be nice to have some hints on what to investigate upon.

Alecyrus:
@zcbenz  Is it possible that this option is available for windows 11?

zcbenz:
> @zcbenz Is it possible that this option is available for windows 11?

I can't find any Windows API for it. There is also an article on this topic: https://valinet.ro/2022/01/21/Disable-the-rounded-corners-in-Windows-11.html, but it does not imply any windows style to disable the rounded corner.

renegadevi:
> @zcbenz Is it possible that this option is available for windows 11?

I need a solution for this as well. Windowed mode is now "broken" in windows 11 because of rounded corners which is not always the intended use when you want a on purpose a dimensioned canvas. This issue is gone when pressing F11 to go into fullscreen mode or when using kioskmode, which is not ideal either for all use cases. At the same time, the full canvas is still recognised as a square with all the pixels even tho the pixels are like masked above.


Windows 10:

windows_10


Windows 11: (corners being cut off)
windows_11

renegadevi:
I found one workaround. By using `thickFrame: false` it fixes both the rounded corners and the position being 0,0 again. 

However, this in turn causes the F11 key only to fullscreen, I cannot longer back out from fullscreen. 

@Alecyrus @zcbenz

Pratiyankkumar:
I can't use roundedCorners: true in windows .(Is it just for macOS??)

wangrongding:
Nice! After searching for a while, I finally found it!👏👏👏

Issue #27457, opened 2021-01-24
Renderer entirely hangs with "uncaught illegal access" when loading page with 2 iframes, nodeIntegrationInSubFrames, and node-ipc


Patronics:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 11.2.1, also verfied on Electron 10.2.x and 9.0.0
* **Operating System:**
  * verified on MacOS 10.14.6, MacOS 11.1, Debian 4.19.160-2, and Debian 5.9.15
* **Last Known Working Electron version:**
Electron 8.5.5 works.

### Expected Behavior
Page sucessfully loads and js continues running normally.

### Actual Behavior
The render window completely hangs partway through loading, is unresponsive to all input including the developer console, and can only be killed from the main process, or by killing the main process. When loading for the first time, there's no output in the developer console, but when reloading an intermittent form of the bug (by wrapping the node-ipc connectTo call in an async function with a delay) following stack trace is sometimes displayed: 
![Screen Shot 2021-01-15 at 5 45 55 PM](https://user-images.githubusercontent.com/9359101/105620751-076d4f00-5db5-11eb-99bf-336a2e12f99f.png). Although the error is displayed as being on line 1 of my html file, line 1 is simply `<html>`, so that's obviously not correct. 

### To Reproduce

Install [this minimal example project. ](https://github.com/Patronics/example-electron-hang) by runing

```sh
$ git clone https://github.com/Patronics/example-electron-hang.git
$ npm install
$ npm start
```
Notice that when running the app, the big iframe's contents never load, and the app is unresponsive to the normal keyboard shortcuts to interact with it (Quit, open the, dev tools, etc.). Look at the source code (especially index.html) to see that the JS abruptly stops running, unlike it expected behavior. The comments describe which lines seem necessary to trigger the bug, and which are just to demonstrate the behavior.

### Additional Information

[...]

PR #27582, opened 2021-02-01
fix: libuv hang when nodeIntegrationInSubframes enabled

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/27457.

Refs https://github.com/electron/electron/commit/13f319eddfe2207558862aa5556da22fea983bd6

If we're loading Node.js into a subframe and we've already initialized the Node.js integration, that means we're in `DidCreateScriptContext` because a new subframe was created, not because we've reloaded the page. If that's the case, that means the last call to `node_bindings_->PrepareMessageLoop()` may still be active and calling it twice in a row can hang the page. In that case, don't re-prepare the message loop.

cc @deepak1556 @zcbenz @MarshallOfSound @ckerr 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where libuv might hang with multiple subframes when `nodeIntegrationInSubframes` is enabled.

### DIFF
--- a/shell/common/node_bindings.h
+++ b/shell/common/node_bindings.h
@@ -159,6 +159,8 @@ class NodeBindings {
   // Isolate data used in creating the environment
   node::IsolateData* isolate_data_ = nullptr;
 
+  int backend_fd_;
+
   base::WeakPtrFactory<NodeBindings> weak_factory_{this};
 
   DISALLOW_COPY_AND_ASSIGN(NodeBindings);
--- a/shell/common/node_bindings.h
+++ b/shell/common/node_bindings.h
@@ -159,7 +159,7 @@ class NodeBindings {
   // Isolate data used in creating the environment
   node::IsolateData* isolate_data_ = nullptr;
 
-  int backend_fd_;
+  int backend_fd_ = -1;
 
   base::WeakPtrFactory<NodeBindings> weak_factory_{this};
 
--- a/shell/common/node_bindings.h
+++ b/shell/common/node_bindings.h
@@ -159,7 +159,11 @@ class NodeBindings {
   // Isolate data used in creating the environment
   node::IsolateData* isolate_data_ = nullptr;
 
-  int backend_fd_ = -1;
+#if defined(OS_WIN)
+  HANDLE handle_;
+#else
+  int handle_ = -1;
+#endif
 
   base::WeakPtrFactory<NodeBindings> weak_factory_{this};
 
--- a/shell/renderer/electron_renderer_client.cc
+++ b/shell/renderer/electron_renderer_client.cc
@@ -113,7 +113,6 @@ void ElectronRendererClient::DidCreateScriptContext(
   // the integration, that means we're here because a new subframe was created,
   // not because we've reloaded the page. In that case, don't re-prepare the
   // message loop.
-  bool node_in_subframe = !is_main_frame && allow_node_in_subframes;
   bool should_load_node =
       (is_main_frame || is_devtools || allow_node_in_subframes) &&
       !IsWebViewFrame(renderer_context, render_frame);
@@ -126,7 +125,7 @@ void ElectronRendererClient::DidCreateScriptContext(
     node_integration_initialized_ = true;
     node_bindings_->Initialize();
     node_bindings_->PrepareMessageLoop();
-  } else if (reuse_renderer_processes_enabled && !node_in_subframe) {
+  } else if (reuse_renderer_processes_enabled) {
     node_bindings_->PrepareMessageLoop();
   }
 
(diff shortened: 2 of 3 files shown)


ckerr (approved):
This works & seems reasonable.

It's a little cumbersome though. If someone else has a better idea I'd be happy for an alternate :slightly_smiling_face:

shell/renderer/electron_renderer_client.cc
```diff
@@ -109,20 +109,24 @@ void ElectronRendererClient::DidCreateScriptContext(
...
-  }
 
   injected_frames_.insert(render_frame);
 
   if (!node_integration_initialized_) {
     node_integration_initialized_ = true;
     node_bindings_->Initialize();
     node_bindings_->PrepareMessageLoop();
-  } else if (reuse_renderer_processes_enabled) {
+  } else if (reuse_renderer_processes_enabled && !node_in_subframe) {
```
deepak1556, line 129: An edge case situation where one creates a window with only `nodeIntegrationInSubframes: true`, render process reuse enabled and the main frame doesn't have nodeIntegration. Now when they reload the page, the subframe node integration message loop will not work correctly https://github.com/electron/electron/pull/25869

`else if (reuse_renderer_processes_enabled` should not be based on frames but instead whether the page was loaded in the same process. Can be checked with browser navigation state or the message loop handles or renderer client state.

zcbenz (commented):
> If that's the case, that means the last call to node_bindings_->PrepareMessageLoop() may still be active and calling it twice in a row can hang the page.

If this is the root cause, may be we can fix it by checking whether backend fd has changed?

Also when calling `PrepareMessageLoop`, I think we should probably do a release first to avoid possible resources leaks.

```c++
void PrepareMessageLoop() {
  int backend_fd = uv_backend_fd(uv_loop_);
  if (backend_fd == old_backend_fd)
    return;

  if (IsInitialized(embed_sem_))
    FreeTheHandles();

  ...
}
```

Patronics:
Hey, is there any news about this issue? Is this pull request functional to fix the problem? If so, is there any chance I could get a compiled version of it (or instructions to compile it), so I could test and verify it resolves the issue?

shell/renderer/electron_renderer_client.cc
```diff
@@ -109,12 +109,15 @@ void ElectronRendererClient::DidCreateScriptContext(
                        (is_not_opened || reuse_renderer_processes_enabled);
   bool is_devtools = IsDevToolsExtension(render_frame);
   bool allow_node_in_subframes = prefs.node_integration_in_sub_frames;
+  // If we're loading Node.js into a subframe and we've already initialized
+  // the integration, that means we're here because a new subframe was created,
+  // not because we've reloaded the page. In that case, don't re-prepare the
+  // message loop.
```
deepak1556, line 115: Is this comment needed ?

deepak1556 (approved):
LGTM

shell/common/node_bindings.h
```diff
@@ -159,6 +159,8 @@ class NodeBindings {
   // Isolate data used in creating the environment
   node::IsolateData* isolate_data_ = nullptr;
 
+  int backend_fd_;
```
ckerr, line 162:
```suggestion
  int backend_fd_ = -1;
```

ckerr:
Not the end of the world, but aren't we leaking the semaphore on reload now?

zcbenz (approved):
I'm good with current change, but we need to fix leak of uv handles on reloads with another PR.

codebytere:
Looks like this is legitimately dying on Windows -

```
✗ Electron tests failed with code 0xc0000409.
ERROR Error: Command failed: C:\Program Files\nodejs\node.exe ./script/spec-runner.js
    at checkExecSyncError (child_process.js:630:11)
    at Object.execFileSync (child_process.js:648:15)
    at runSpecRunner (C:\Users\codebytere\.electron_build_tools\src\e-test.js:33:16)
    at Object.<anonymous> (C:\Users\codebytere\.electron_build_tools\src\e-test.js:63:3)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module.load (internal/modules/cjs/loader.js:985:32)
    at Function.Module._load (internal/modules/cjs/loader.js:878:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47
```

Investigating now

codebytere:
It looks like this bug is _only_ present on Linux and macOS - and `uv_backend_fd` doesn't work on iocp.

zcbenz:
> It looks like this bug is _only_ present on Linux and macOS - and `uv_backend_fd` doesn't work on iocp. So we'll gate it to those OSes

We need to prevent duplicate initializations even if it does not seem to cause problems for now. 

On Windows you can access IOCP handle via `uv_loop_->iocp`, the `uv_backend_fd` always returns -1 there.

Issue #27378, opened 2021-01-20
BrowserView: flickers when switching between view


kevinlinv:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** 10.1.7 and beyond (I tested on latest 11.2.0 too)
* **Operating System** macOS 10.15.7
* **Last Known Working Electron version:** 10.1.6

NOTE: tested in windows, the bug is not present.

### Expected Behavior

Switching browser view should **not flickers**.

### Actual Behavior

Switching browser view **flickers**.

### To Reproduce

Load electron fiddle friendly gist here https://gist.github.com/kevinlinv/eada1e5bdd1282db338971c4c42c671e

<details>
<summary>Collapse here for the code in case gist is not loading</summary>

`index.html`
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8">
    <title>Hello World!</title>
    <link rel="stylesheet" type="text/css" href="./styles.css">
  </head>
  <body>
    <div style="display:flex">
      <p style="margin:0px; padding: 5px">MyApp</p>
      <button id="tab1">tab1</button>
      <button id="tab2">tab2</button>
    </div>
    <script>
      // You can also require other files to run in this process
      require('./renderer.js')
    </script>
  </body>
</html>
```

`main.js`
```js
const {app, ipcMain, BrowserView, BrowserWindow} = require('electron')

function createWindow () {
  const mainWindow = new BrowserWindow({
    width: 1080,
    height: 720,
    webPreferences: { nodeIntegration: true }
  })
  

  const views = [
    createView({ backgroundColor: '#ddd', url: 'https://www.youtube.com/watch?v=RDfjXj5EGqI' }),
    createView({ backgroundColor: '#000', url: 'https://electronjs.org' }),
  ];
  const focusView = (id) => {
    const { width, height } = mainWindow.getContentBounds();
    const windowContentHeight = 40;
    mainWindow.setBrowserView(views[id]);
    views[id].setBounds({
      x: 0,
      y: windowContentHeight, 
      width,
      height: height - windowContentHeight,
    });
  }
[...]

PR #27585, opened 2021-02-02
fix: BrowserView rendering flicker

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/27378.

In an `NSView`, `drawRect` draws the view’s image within the specified rectangle. It's invoked _very_ frequently and as such performance is paramount. As calls to `drawRect` need to traverse the whole responder chain, we were adding a costly [`getenv`](https://pubs.opengroup.org/onlinepubs/009695399/functions/getenv.html) call to every invocation, which created a rendering flicker. This was debug-only, so we should not override it at all, but keep it commented out.

See https://developer.apple.com/documentation/appkit/nsview/1483686-drawrect for more details.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an occasional white flicker present when rendering BrowserViews in close succession.

### DIFF
--- a/shell/browser/native_browser_view_mac.mm
+++ b/shell/browser/native_browser_view_mac.mm
@@ -4,6 +4,7 @@
 
 #include "shell/browser/native_browser_view_mac.h"
 
+#import <objc/runtime.h>
 #include <vector>
 
 #include "shell/browser/ui/drag_util.h"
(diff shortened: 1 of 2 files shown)


ckerr (commented):
While I agree this is a feature that's only intended for debugging, IMO it's still a documented feature and so removing it could be considered a breaking change.

Hypothetical example: say for instance someone has an Electron app with a CI system that sets `ELECTRON_DEBUG_DRAG_REGIONS=1` and then does some integration tests and takes pictures while the test is being run. These tests are now going to start behaving differently because of this change.

If the flicker is coming from the cost of excessive # of calls to `getenv()`, let's just call it once and cache that value, and then keep querying that cached value instead.

IMO this is a `semver/major` change because we're changing documented features, even though it's not a big feature.

ckerr:
Could @electron/wg-api members give a 2nd opinion on the semver label this PR should get?

shell/browser/native_browser_view_mac.mm
```diff
@@ -133,13 +133,14 @@ - (void)mouseDragged:(NSEvent*)event {
   [self.window setFrameOrigin:newOrigin];
 }
 
-// For debugging purposes only.
-- (void)drawRect:(NSRect)aRect {
-  if (getenv("ELECTRON_DEBUG_DRAG_REGIONS")) {
```
zcbenz, line 138: Instead of calling `getenv` in every `drawRect`, could caching the result in the constructor fix the flicker?

```objective-c
- (id)initWithFlag:(BOOL)enableDebug {
  enableDebug_ = enableDebug;
  ...
}

- (void)drawRect:(NSRect)aRect {
  if (enableDebug_) {
    ...
  }
}
```

If overriding the method itself is causing problems, another option is to dynamically override the `drawRect` method with `class_replaceMethod`, you can find some examples in `native_window_mac.mm`.

codebytere: Ill try this!

zcbenz (approved):

jkleinsc (approved):

PR #27597, opened 2021-02-03
fix: [net] no errors on response stream

### LINKED ISSUE #24948: When proxy login fails, a dialog of "Error: net::ERR_TUNNEL_CONNECTION_FAILED" is displayed even if the error was catched
### Preflight Checklist

### Issue Details

* **Electron Version:**
  9.1.2
* **Operating System:**
  Windows 10 2004

### Expected Behavior
Error dialog is NOT shown if error captured

### Actual Behavior
Error dialog is shown if error captured

### To Reproduce

https://gist.github.com/617c8c6596c269df002996da38d39350
Just run the fiddle under a password protected proxy

### Screenshots

image

nornagon:
#### Description of Change
Fixes #24948. To better match Node, don't emit errors on the response stream
(only the request stream).

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed errors being emitted on the response stream of requests created with the 'net' module in some cases.

### DIFF
--- a/lib/browser/api/net.ts
+++ b/lib/browser/api/net.ts
@@ -432,9 +432,7 @@ export class ClientRequest extends Writable implements Electron.ClientRequest {
       if (this._response) { this._response._storeInternalData(null, null); }
     });
     this._urlLoader.on('error', (event, netErrorString) => {
-      const error = new Error(netErrorString);
-      if (this._response) this._response.destroy(error);
-      this._die(error);
+      this._die(new Error(netErrorString));
     });
 
     this._urlLoader.on('login', (event, authInfo, callback) => {
@@ -515,7 +513,7 @@ export class ClientRequest extends Writable implements Electron.ClientRequest {
     this.destroy(err);
     if (this._urlLoader) {
       this._urlLoader.cancel();
-      if (this._response) this._response.destroy(err);
+      if (this._response) this._response.destroy();
     }
   }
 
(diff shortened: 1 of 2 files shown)


nornagon:
Hm, actually I'm not 100% sure what the right thing to do here is... as is, the stream emits _both_ `response` (with a status code of 407) _and_ `error` (with `net::ERR_TUNNEL_CONNECTION_FAILED`). That seems surprising.

Thoughts? @electron/wg-api?

jkleinsc:
@nornagon I think having both the response and error message are helpful as they give more context as to what happened:
eg 407 status is "Proxy Authentication Required" but without the context of the error code `net::ERR_TUNNEL_CONNECTION_FAILED` it isn't clear why the request failed.

zcbenz (approved):
> Hm, actually I'm not 100% sure what the right thing to do here is... as is, the stream emits both response (with a status code of 407) and error (with net::ERR_TUNNEL_CONNECTION_FAILED). That seems surprising.

I don't have a good idea what is the best way to deal with this situation, but I think documenting this behavior should be enough.

jkleinsc (changes requested):
@nornagon can we add documentation for this change?

jkleinsc:
API LGTM

poiru:
@nornagon I'd like to see this merged! Friendly ping in case you lost track of this.

nornagon:
@poiru do you have thoughts about what to do when the tunnel connection fails? Should we emit _both_ a `response` _and_ an `error` event? Should we try to hide the response event?

poiru:
I think an `error` event alone would be more in line with my expectations.

poiru:
An `error` alone would be what I'd expect, but I also see the benefit of https://github.com/electron/electron/pull/27597#issuecomment-773603319. Lets emit both and document this clearly?

nornagon:
Discussed with API WG, let's for now change this to only emit the `error` event, and in future if we want to provide access to the response information as well, we can decide what to do then.

jkleinsc:
Closing due to inactivity

PR #27598, opened 2021-02-03
fix: cherry-pick 5c7ad5393f74 from chromium

erickzhao:
Mac a11y: Use the keyboard focusable element for
 NSAccessibilityTextChangeElement

When setting the NSAccessibilityTextChangeElement property for
NSAccessibilitySelectedTextChangedNotifications, use the keyboard
focusable element instead of the element that has the focus side of the
text selection. Using the latter, when the element is an empty group,
VoiceOver will focus the containing Web View (when using the VO
cursor follows keyboard focus setting). This makes it impossible to use
the down keyboard key to move past these empty nodes.

VoiceOver cursor" setting in contenteditable nodes.

AX-Relnotes: Fix a bug with the "Synchronize keyboard focus and
Bug: 952922
Change-Id: I3627936726f89b01132c32bd5d83758fc7c3dac4
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2642686
Auto-Submit: Martin Robinson <mrobinson@igalia.com>
Commit-Queue: Nektarios Paisios <nektar@chromium.org>
Reviewed-by: Nektarios Paisios <nektar@chromium.org>
Cr-Commit-Position: refs/heads/master@{#846707}

### Release Notes
Notes: Backported fix for https://crbug.com/952922.


zcbenz:
This PR requires review from @electron/wg-upgrades

jkleinsc (approved):

codebytere (approved):

Issue #21586, opened 2019-12-19
Memory leak when opening and closing windows


t57ser:
### Issue Details

* **Electron Version:**
  * 5, 6, 7, 8
* **Operating System:**
  * Windows 10
* **Last Known Working Electron version:**
  * ?

### Expected Behavior
Opening and closing window should not increase memory

### Actual Behavior
Memory will increase for each window even after the window has been destroyed

### To Reproduce
```
const { app, BrowserWindow, BrowserView } = require("electron");

const init = () => {
	let browserWindow2 = new BrowserWindow({
		show: true,
	});
	let j = 0;
	const interval = setInterval(() => {
		j++;
		if (j > 10) {
			clearInterval(interval);
		}
		for(let i = 0; i < 20; i++) {
			let browserWindow = new BrowserWindow({
				show: true,
			});
			setTimeout(() => {
				browserWindow.close();
			}, 5000)
		}
	}, 1000);

};
app.on("ready", () => {
	init();
});
```
The above code snipped should not result in more memory than just opening a single window.
I also tried to call the garbage collector manually and I also tried browserWindow.destroy(). No difference.

jackple:
Is there any solution?

pushkin-:
Can someone test if this repros in Electron 11?

t57ser:
Still happening in electron 11. Opening a single window results in 40mb used RAM while executing the code i posted results in about double the memory.

VladTabaka:
Same here (versions 9 and 11) is it somehow related to https://github.com/electron/electron/pull/23578 ?

nornagon:
Did some experimenting with LSan and this looks like a possible candidate:

```
=================================================================
==43956==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 32640 byte(s) in 40 object(s) allocated from:
    #0 0x100593330 in __asan_memmove (/Users/nornagon/work/electron-gn/src/out/Testing-asan/Electron.app/Contents/MacOS/libclang_rt.asan_osx_dynamic.dylib:x86_64+0x43330)
    #1 0x112b5d1e7 in operator new(unsigned long) (/Users/nornagon/work/electron-gn/src/out/Testing-asan/Electron.app/Contents/Frameworks/Electron Framework.framework/Versio
ns/A/Electron Framework:x86_64+0x1128d1e7)
[...]

PR #27621, opened 2021-02-04
fix: memory leak in TrackableObject

nornagon:
#### Description of Change
This fixes a memory leak in TrackableObject (of which BaseWindow and
BrowserWindow are the only subclasses). It was registering an exit-time
destructor for every TrackableObject, and never removing them. This is replaced
by the gin_helper::CleanedUpAtExit helper class, which removes the destructor
from the list when the object is destroyed.

Ref #21586.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a memory leak when creating BrowserWindows.

### DIFF
--- a/shell/browser/electron_browser_main_parts.cc
+++ b/shell/browser/electron_browser_main_parts.cc
@@ -231,14 +231,6 @@ int ElectronBrowserMainParts::GetExitCode() {
   return exit_code_ != nullptr ? *exit_code_ : 0;
 }
 
-void ElectronBrowserMainParts::RegisterDestructionCallback(
-    base::OnceClosure callback) {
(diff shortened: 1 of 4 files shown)


MarshallOfSound (approved):

deepak1556:
>  It was registering an exit-time destructor for every TrackableObject, and never removing them

@nornagon can you clarify what is not removed, are you referring to the entries in `destructors_` list ?

nornagon:
@deepak1556 correct

deepak1556 (approved):

codebytere (approved):

zcbenz (approved):

PR #27629, opened 2021-02-04
fix: use gin pageallocator for v8 allocations

deepak1556:
#### Description of Change

Based on https://github.com/electron/electron/pull/26331

With macOS 11.2 on mac arm64, loading wasm modules crashes node process, this has been narrowed down to be an issue with the default page allocator from v8, using the partition alloc page allocator works fine and given we have already switched the default allocator in 12-x-y and above. This change helps apps on 11-x-y running on mac arm64 to load wasm modules in node processes. Also given the Paritition-alloc everywhere move https://bugs.chromium.org/p/chromium/issues/detail?id=1121427 this should align with it too.

Refs https://github.com/microsoft/vscode/issues/115646

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: fix crash with loading wasm modules in `ELECTRON_RUN_AS_NODE` processes on mac arm64


MarshallOfSound (changes requested):
I initially didn't backport the original PR due to it's inherent risky nature (literally changing the entire memory allocator for the main process).  I'm not sure I can confidently say this is safe or it's `semver/patch` (it changes codesigning behavior for instance as outlined in the original PR)

I won't hold this back if everyone is on board, but I feel like I need to point out this is **risky** and not a simple PR behind the scenes, the impact is quite far reaching.

deepak1556:
> literally changing the entire memory allocator for the main process

The allocator has an impact only for allocations from v8 and not allocations from other components in the process (which are already using partition allocator in the main process). And v8 has been using this allocator in the renderer process alongside node. Also to add this change not only impacts main process but also any node process spawned with `ELECTRON_RUN_AS_NODE`.

> the impact is quite far reaching.

Can you expand ?

> I'm not sure I can confidently say this is safe or it's semver/patch (it changes codesigning behavior for instance as outlined in the original PR)

Agree that this is `semver/minor`

MarshallOfSound:
> the impact is quite far reaching.

It swaps out things like the arraybuffer allocator that we give to node, which transitively impacts memory allocations which native modules have to deal with (just as a single example).  How exactly this impacts certain advanced use cases has yet to be seen (because this change was only made in Electron 12).  Can we make this conditional and put it behind a fuse or flag or someething? Then it would be `semver/patch` and opt-in?

deepak1556:
> It swaps out things like the arraybuffer allocator that we give to node, which transitively impacts memory allocations which native modules have to deal with (just as a single example)

Don't we do that even in the renderer with nodeIntegration enabled ?

> Can we make this conditional and put it behind a fuse or flag or someething? Then it would be semver/patch and opt-in?

Yup, can make this change. Will add it behind fuse.

deepak1556:
Fuse is not available in `11-x-y`, so added it behind buildflag

MarshallOfSound (approved):

jkleinsc (approved):

deepak1556:
Closing in favor of https://github.com/electron/electron/pull/27684

PR #27630, opened 2021-02-04
fix: clean up base::LinkedList in context_bridge::ObjectCache

### LINKED ISSUE #27039: Memory leak when passing IPC events over contextBridge
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 11.1.0 and 9.3.5
* **Operating System:**
  * Windows 10 2004
  * NA

### Expected Behavior
use of webContents.send + ipcRenderer.on cleans up memory after use

### Actual Behavior

In a real-world case, I have a renderer for the UI and a separate browserWindow for an audio process. The audio process monitors audio devices and reports the current volume of all devices via ipcRenderer.send (through preload contextBridge to the main process). The main process then sends this volume data over to the main UI renderer via webContents.send whereas the UI renderer listens via ipcRenderer.on (via preload contextBridge). This happens about once every 100 milliseconds. Memory use of the electron app, especially the renderer, slowly climbs to theoretical Infinitum. The starting memory is about 350MB. Within about a day, it's up to 1,500MB.

It does not matter what kind of content is passed through as arguments, nor its size (garbage collector seems to work on arguments). What matters is how often / frequently webContents.send is called and if there is an ipcRenderer.on listener for it (memory leak still exists even if there is nothing in the listener function so long as the listener exists).

### To Reproduce

Electron fiddle with a very simple example calling webContents.send every millisecond (to make the test quick... but the issue exists if you do every 100 milliseconds too but is much slower) and sending a Math.random() random number to the renderer. Memory use will increase over the course of a few minutes. I tested with a 10,000-item map as an argument as well, which made memory spike very quickly and get garbage-collected repeatedly, resulting in an average memory growth of about the same as Math.random(). This leads me to believe the problem is not the data being sent but the IPC execution itself.

[https://gist.github.com/37cde4adc11b3aa3f435e5d2698d9e05](https://gist.github.com/37cde4adc11b3aa3f435e5d2698d9e05)

### Additional Information

Additional info: Using handle/invoke seems to bypass this memory leak. However this is not ideal because renderer UI has to query main every 100 milliseconds for audio volume, and main has to locally store volume information (via another invoke / handle between main and the audio process). This creates additional VU meter lag.

MarshallOfSound:
base::LinkedList does not delete its members on destruction. We need to manually ensure the linkedlist is empty when the ObjectCache is destroyed.

Fixes #27039

Notes: Fixed memory leak when sending non-primitives over the context bridge

### DIFF
--- a/shell/renderer/api/context_bridge/object_cache.cc
+++ b/shell/renderer/api/context_bridge/object_cache.cc
@@ -21,7 +21,15 @@ ObjectCachePairNode::ObjectCachePairNode(ObjectCachePair&& pair) {
 ObjectCachePairNode::~ObjectCachePairNode() = default;
 
 ObjectCache::ObjectCache() {}
-ObjectCache::~ObjectCache() = default;
+ObjectCache::~ObjectCache() {
+  for (const auto& pair : proxy_map_) {
+    while (!pair.second.empty()) {
+      ObjectCachePairNode* node = pair.second.head()->value();
+      node->RemoveFromList();
+      delete node;
+    }
+  }
+}
 
 void ObjectCache::CacheProxiedObject(v8::Local<v8::Value> from,
                                      v8::Local<v8::Value> proxy_value) {


nornagon (approved):

shell/renderer/api/context_bridge/object_cache.cc
```diff
@@ -21,7 +21,15 @@ ObjectCachePairNode::ObjectCachePairNode(ObjectCachePair&& pair) {
...
-ObjectCache::~ObjectCache() = default;
+ObjectCache::~ObjectCache() {
+  for (const auto& pair : proxy_map_) {
+    while (!pair.second.empty()) {
+      ObjectCachePairNode* node = pair.second.head()->value();
+      node->RemoveFromList();
+      delete node;
+    }
+  }
+}
```
ckerr, line 32: This is a little larger scope, but -- what is the benefit of using base::LinkedList here instead of std::list, which is simpler / doesn't require a ObjectCachePairNode / doesn't have these gotchas?

According to the docs, the  main reason to use base::LinkedList are for performance reasons of (1) removal is O(1) rather than O(n), and (2) insertion with base::LinkedList<T> never requires heap allocations. But removal never called except in the destructor, and the items we insert are heap-allocated ObjectCachePairNodes, so do either of those advantages apply here?

Jonna184 (approved):

Jonna184 (approved):

shell/renderer/api/context_bridge/object_cache.cc
MarshallOfSound: Agree in general, I want to land this fix then I'll follow it up to see if we can remove our usage of `LinkedList`.  I can't remember exactly why we used it originally, but this code has been through several refactors and it's possible our usage of LinkedList is no longer required.

PR #27635, opened 2021-02-05
fix: Backport SVG filter invalidation fix from Chromium

rzhao271:
#### Description of Change

This is a backport of an upstream fix that is already in Electron v12.0.0-beta.16 and newer.

Downstream issue:
https://github.com/microsoft/vscode/issues/112693
In a VS Code, an SVG produced by a notebook wasn't rendering on time.

Diff to backport from Chromium:
https://chromium.googlesource.com/chromium/src/+/15a8dbe6efe865eaaae63b48b6a3bacd3420a89b%5E%21/
There was an SVG rendering glitch in Chromium where invalidate wasn't getting called in some cases.

CC @deepak1556 

#### Checklist

- [ ] `npm test` passes
- [x] tests are added, **though in Chromium** (https://github.com/electron/electron/blob/master/docs/development/testing.md)

#### Release Notes

Notes: Fix svg with filter content not being rendered

### DIFF
--- a/patches/chromium/fix_svg_filter_data_invalidation_for_empty_containers.patch
+++ b/patches/chromium/fix_svg_filter_data_invalidation_for_empty_containers.patch
@@ -1,0 +1,116 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Fredrik=20S=C3=B6derqvist?= <fs@opera.com>
+Date: Tue, 1 Dec 2020 17:15:59 +0000
+Subject: Fix SVG filter data invalidation for empty containers
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+Empty containers will fail to create FilterData because they have an
+empty bounding box. When children were added to the container, making
+the bounding box non-empty we would fail to mark the paint properties as
+needing update because of the check for existing FilterData in
+SVGElementResourceClient::InvalidateFilterData().
+
+Instead always dispose of the filter data and mark paint properties for
+update, and use the |filter_data_dirty_| flag as an early-out check.
+
+Since we can now call InvalidateFilterData() directly in
+SVGElementResourceClient::ResourceElementChanged(), fold
+ClearFilterData() into InvalidateFilterData().
+
+Bug: 1154050
+Change-Id: I01c8ebac4f62532e2c17aabc3d998d7601dbb71b
+Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2566992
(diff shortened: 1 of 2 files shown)


patches/chromium/fix_svg_filter_data_invalidation_for_empty_containers.patch
```diff
@@ -0,0 +1,118 @@
...
+-    // we don't need to perform the associated invalidations.
+-    layout_object->SetNeedsPaintPropertyUpdate();
+-    MarkFilterDataDirty();
++  LayoutObject* layout_object = element_->GetLayoutObject();
++  if (!layout_object)
++    return;
++  // TODO(fs): If the resource element (for a filter) doesn't actually change
++  // we don't need to perform the associated invalidations.
++  InvalidateFilterData();
++  if (layout_object->Parent()) {
```
deepak1556, line 50: This codepath was only added as a refactor from https://chromium-review.googlesource.com/c/chromium/src/+/2489650

It is not available in this branch, the change should be

```
LayoutObject* layout_object = element_->GetLayoutObject();
if (!layout_object)
    return;
// TODO(fs): If the resource element (for a filter) doesn't actually change
// we don't need to perform the associated invalidations.
InvalidateFilterData();
SVGResourcesCache::ResourceReferenceChanged(*layout_object);
```

deepak1556 (changes requested):

patches/chromium/fix_svg_filter_data_invalidation_for_empty_containers.patch
rzhao271: Realized that I should be using `ResourceReferenceChanged` before seeing this comment 😅

patches/chromium/fix_svg_filter_data_invalidation_for_empty_containers.patch
```diff
@@ -0,0 +1,115 @@
...
+ void SVGElementResourceClient::ResourceElementChanged() {
+-  if (LayoutObject* layout_object = element_->GetLayoutObject()) {
+-    ClearFilterData();
++  LayoutObject* layout_object = element_->GetLayoutObject();
++  if (!layout_object)
++    return;
++  // TODO(fs): If the resource element (for a filter) doesn't actually change
++  // we don't need to perform the associated invalidations.
++  InvalidateFilterData();
++  if (layout_object->Parent()) {
```
deepak1556, line 45:
```suggestion
```

condition is not required, `ResourceReferenceChanged` already does that check.

deepak1556 (approved):

deepak1556 (approved):

jkleinsc (approved):

jkleinsc:
Merging as CI failure is flake unrelated to this PR.

Issue #27673, opened 2021-02-08
Electron v11 breaks crypto.createDiffieHellman with non-number generators


SchoofsKelvin:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details
* **Electron Version:**
  * Electron v11 (11.0.0-beta.1, 11.0.0, 11.2.1 and 11.2.3 are tested and affected)
* **Operating System:**
  * Personally tested on Windows 10 (1909), but possibly affects others _(a bunch of VS Code users on different OSes)_
* **Last Known Working Electron version:**
  * Electron v10 (tested 10.3.2)

### Bug
Renderer process crashes when using [crypto.createDiffieHellman](https://nodejs.org/docs/latest-v12.x/api/crypto.html#crypto_crypto_creatediffiehellman_prime_primeencoding_generator_generatorencoding) with a non-number generator as argument. The exact same code works fine in the NodeJS version I tested, namely v12.18.3, the same as electron v11.2.3 uses.

Originally noticed in VS Code extensions (SchoofsKelvin/vscode-sshfs#239) after VS Code 1.53.0 switched from electron v9 to v11.2.1.

### To Reproduce
Electron fiddle: https://gist.github.com/SchoofsKelvin/90f78f8aeffa04e344bf8dbd3ade4961
![image](https://user-images.githubusercontent.com/14597409/107291105-e91b7a80-6a67-11eb-8349-eecabb4b764f.png)
Clicking a button will either show e.g. "test1 done" or crash the renderer, resulting in the page becoming fully white.
Alternatively, simply run this in the renderer process of an affected version and it should crash the renderer process:
```js
require('crypto').createDiffieHellman("abc", "\02")
```

Issue #26045, opened 2020-10-19
did-change-theme-color tests are flaky on WOA


jkleinsc:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * master
* **Operating System:**
  * Windows on ARM
* **Last Known Working Electron version:**
  * ?

### Expected Behavior

`webContents module did-change-theme-color event is triggered with correct theme color` and `<webview> tag did-change-theme-color event emits when theme color changes` tests should pass on Windows on ARM.

### Actual Behavior

The tests timeout, eg:
```
not ok 397 <webview> tag did-change-theme-color event emits when theme color changes
  Error: Timeout of 180000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\vsts-agent-win-x86-2.166.4\_work\1\s\src\electron\spec\webview-spec.js)
```
and 
```
not ok 1344 webContents module did-change-theme-color event is triggered with correct theme color
  Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\vsts-agent-win-x86-2.166.4\_work\1\s\src\electron\spec-main\api-web-contents-spec.ts)
      at listOnTimeout (internal/timers.js:554:17)
      at processTimers (internal/timers.js:497:7)
```

### To Reproduce

Run the did-change-theme-color tests on WOA.

### Additional Information

See https://github.visualstudio.com/electron/_test/analytics?definitionId=41&contextType=build for statistics on the flaky failures

ckerr:
Data point: it doesn't seem to be exclusive to WOA. For example I got the same test failure in https://ci.appveyor.com/project/electron-bot/electron-ljo26/builds/35900922 for a docs change PR :slightly_smiling_face:

ckerr:
I did a little bit of testing on this. Disclaimer: I found it pretty easy to reproduce on an Ubuntu 20.04 box so this may or may not be the same issue that's triggering the WOA failures.

[...]

PR #27674, opened 2021-02-09
fix: crash in crypto.createDiffieHellman

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/27673.

We disabled the contents of one of the Diffie-Hellman constructors owing to BoringSSL compatibilities, but I've determined that the functionality itself is compatible; it's just the error throwers that aren't. As such, we can specifically remove those and restore `crypto.createDiffieHellman()` functionality. 

Tested with https://gist.github.com/SchoofsKelvin/90f78f8aeffa04e344bf8dbd3ade4961.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash when calling `crypto.createDiffieHellman()` with certain parameters.

### DIFF
--- a/patches/node/fix_comment_out_incompatible_crypto_modules.patch
+++ b/patches/node/fix_comment_out_incompatible_crypto_modules.patch
@@ -9,42 +9,46 @@ with what's exposed through BoringSSL. I plan to upstream parts of this or
 otherwise introduce shims to reduce friction.
 
 diff --git a/src/node_crypto.cc b/src/node_crypto.cc
-index 91cb94d8dbe9db0adbee5e005649188e1ccbcbf9..2429ebf6035de392c6bac18c0b924b995fb3daeb 100644
+index 91cb94d8dbe9db0adbee5e005649188e1ccbcbf9..c3d12dc4cc18888815ff5e2c30a21974322d1faf 100644
 --- a/src/node_crypto.cc
 +++ b/src/node_crypto.cc
-@@ -5191,6 +5191,7 @@ bool DiffieHellman::Init(int primeLength, int g) {
- 
+@@ -5192,11 +5192,11 @@ bool DiffieHellman::Init(int primeLength, int g) {
  bool DiffieHellman::Init(const char* p, int p_len, int g) {
    dh_.reset(DH_new());
-+#if 0
    if (p_len <= 0) {
-     BNerr(BN_F_BN_GENERATE_PRIME_EX, BN_R_BITS_TOO_SMALL);
-     return false;
-@@ -5199,6 +5200,7 @@ bool DiffieHellman::Init(const char* p, int p_len, int g) {
-     DHerr(DH_F_DH_BUILTIN_GENPARAMS, DH_R_BAD_GENERATOR);
+-    BNerr(BN_F_BN_GENERATE_PRIME_EX, BN_R_BITS_TOO_SMALL);
++    OPENSSL_PUT_ERROR(BN, BN_R_BITS_TOO_SMALL);
      return false;
    }
-+#endif
+   if (g <= 1) {
+-    DHerr(DH_F_DH_BUILTIN_GENPARAMS, DH_R_BAD_GENERATOR);
++    OPENSSL_PUT_ERROR(DH, DH_R_BAD_GENERATOR);
+     return false;
+   }
    BIGNUM* bn_p =
-       BN_bin2bn(reinterpret_cast<const unsigned char*>(p), p_len, nullptr);
-   BIGNUM* bn_g = BN_new();
-@@ -5214,6 +5216,7 @@ bool DiffieHellman::Init(const char* p, int p_len, int g) {
- 
+@@ -5215,18 +5215,18 @@ bool DiffieHellman::Init(const char* p, int p_len, int g) {
  bool DiffieHellman::Init(const char* p, int p_len, const char* g, int g_len) {
    dh_.reset(DH_new());
-+#if 0
    if (p_len <= 0) {
-     BNerr(BN_F_BN_GENERATE_PRIME_EX, BN_R_BITS_TOO_SMALL);
-     return false;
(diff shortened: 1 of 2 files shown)


patches/node/fix_comment_out_incompatible_crypto_modules.patch
```diff
@@ -28,23 +28,29 @@ index 91cb94d8dbe9db0adbee5e005649188e1ccbcbf9..2429ebf6035de392c6bac18c0b924b99
...
- 
+@@ -5215,18 +5217,18 @@ bool DiffieHellman::Init(const char* p, int p_len, int g) {
  bool DiffieHellman::Init(const char* p, int p_len, const char* g, int g_len) {
    dh_.reset(DH_new());
-+#if 0
    if (p_len <= 0) {
-     BNerr(BN_F_BN_GENERATE_PRIME_EX, BN_R_BITS_TOO_SMALL);
+-    BNerr(BN_F_BN_GENERATE_PRIME_EX, BN_R_BITS_TOO_SMALL);
++    // BNerr(BN_F_BN_GENERATE_PRIME_EX, BN_R_BITS_TOO_SMALL);
+     return false;
```
ckerr, line 37: If we're re-enabling these p_len <= 0 tests here (sans the `BNerr()` call) should we do the same above for other version of DiffieHellman::Init()?

codebytere: yeah good catch we should

patches/node/fix_comment_out_incompatible_crypto_modules.patch
```diff
@@ -9,42 +9,46 @@ with what's exposed through BoringSSL. I plan to upstream parts of this or
...
-@@ -5191,6 +5191,7 @@ bool DiffieHellman::Init(int primeLength, int g) {
- 
+@@ -5192,11 +5192,11 @@ bool DiffieHellman::Init(int primeLength, int g) {
  bool DiffieHellman::Init(const char* p, int p_len, int g) {
    dh_.reset(DH_new());
-+#if 0
    if (p_len <= 0) {
-     BNerr(BN_F_BN_GENERATE_PRIME_EX, BN_R_BITS_TOO_SMALL);
+-    BNerr(BN_F_BN_GENERATE_PRIME_EX, BN_R_BITS_TOO_SMALL);
++    // BNerr(BN_F_BN_GENERATE_PRIME_EX, BN_R_BITS_TOO_SMALL);
```
nornagon, line 20: Commenting out these errors seems bad! Won't that mean we don't get proper error messages in these error cases?

nornagon (changes requested):

patches/node/fix_comment_out_incompatible_crypto_modules.patch
codebytere: no - it means that we get slightly less clean fallback messages, which differ slightly.

if i for example run `crypto.createDiffieHellman(-1)` i see:

Screen Shot 2021-02-09 at 9 53 16 AM

nornagon: I think you've hit the wrong constructor in your test. There are several for DiffieHellman::Init, and this patch does not affect the one-argument version.

Here's an example of how this patch fails to produce an appropriate error:

```
> require('crypto').createDiffieHellman('', 'utf8')
Uncaught Error: Initialization failed
    at new DiffieHellman (internal/crypto/diffiehellman.js:85:19)
    at Object.createDiffieHellman (crypto.js:135:10)
```

nornagon: I've updated the PR to use error routines that work with boringssl.

patches/node/fix_comment_out_incompatible_crypto_modules.patch
```diff
@@ -9,42 +9,46 @@ with what's exposed through BoringSSL. I plan to upstream parts of this or
...
-@@ -5191,6 +5191,7 @@ bool DiffieHellman::Init(int primeLength, int g) {
- 
+@@ -5192,11 +5192,11 @@ bool DiffieHellman::Init(int primeLength, int g) {
  bool DiffieHellman::Init(const char* p, int p_len, int g) {
    dh_.reset(DH_new());
-+#if 0
    if (p_len <= 0) {
-     BNerr(BN_F_BN_GENERATE_PRIME_EX, BN_R_BITS_TOO_SMALL);
+-    BNerr(BN_F_BN_GENERATE_PRIME_EX, BN_R_BITS_TOO_SMALL);
++    OPENSSL_PUT_ERROR(BN, BN_R_BITS_TOO_SMALL);
```
ckerr, line 20: Question about the PR description:

> We disabled the contents of one of the Diffie-Hellman constructors owing to BoringSSL compatibilities, but I've determined that the functionality itself is compatible; it's just the error throwers that aren't.

Is this saying that the semantics of the error handlers are incompatible (i.e. errors are fundamentally handled differently), or just that the API is incompatible (i.e. `BNerr()` doesn't exist in boringssl)? @codebytere I'm pretty sure that you meant the latter, but want to confirm.

If so, then :+1: on this change being the correct one. Both BNerr and OPENSSL_PUT_ERROR are convenience macros that resolve to `ERR_put_error(..., __FILE__, __LINE__)` so this is a good way to reduce friction when swapping between openssl and boringssl. (And to the comment at the top of this patch file, this is the change that should be upstreamed.)

codebytere: > BNerr() doesn't exist in boringssl

correct

ckerr (approved):
:+1: on re-enabling this code but swapping out the error handler invocation to a portable macro

ckerr:
`win-x64-testing` failure appears to be the [did-change-theme-color](https://github.com/electron/electron/issues/26045) flake unrelated to this PR

nornagon (approved):

PR #27678, opened 2021-02-09
docs: MessagePorts guide

nornagon:
#### Description of Change
This adds a guide page explaining some details about how MessagePorts work in
Electron and some example use-cases.

Another example I'd like to add is that of "ports as capabilities", i.e. as a
security feature--if you have the MessagePort, you're allowed to use the
feature. That might be a bit abstract though, and I'm not aware of any apps
using that strategy in the wild.

#### Checklist

#### Release Notes

Notes: none


deepak1556 (approved):
👍

PR #27699, opened 2021-02-10
fix: Replace ClearFilterData with InvalidateFilterData

rzhao271:
#### Description of Change

This is a continuation of #27635. It turned out that the new `InvalidateFilterData` function was still not being called, so this PR replaces the `ClearFilterData` function with `InvalidateFilterData` completely.

/cc @deepak1556 

#### Checklist

- [X] ~~`npm test` passes~~ Manually verified by building a release build and running it against a fiddle.

#### Release Notes

Notes: Fix SVG content with filter sometimes not being rendered.


nornagon:
@rzhao271 since there's no test here, have you confirmed that this actually fixes the issue? It seems like #27635 didn't.

rzhao271:
@nornagon sorry about that. I failed to test that PR properly on my end before it got merged.

Here is a demo showing the fixed behaviour for this PR, where as soon as the app loads, the black circle shows up:

recording (1)

Here is the current faulty behaviour in 11.2.3 where one must invalidate the layout before the black circle shows up:

recording (2)

patches/chromium/replace_clearfilterdata_with_invalidatefilterdata.patch
```diff
@@ -0,0 +1,133 @@
...
+     if (resources->HasClipOrMaskOrFilter()) {
+       InvalidationModeMask invalidation_mask =
+           SVGResourceClient::kBoundariesInvalidation;
+-      bool filter_data_invalidated = false;
+       if (resources->Filter()) {
+-        filter_data_invalidated = client->ClearFilterData();
+-        invalidation_mask |=
+-            filter_data_invalidated ? SVGResourceClient::kPaintInvalidation : 0;
++        client->InvalidateFilterData();
++        invalidation_mask |= SVGResourceClient::kPaintInvalidation;
```
deepak1556, line 26: `InvalidateFilterData` calls into `layout_object->SetNeedsPaintPropertyUpdate` and I don't see `SVGResourceClient::kPaintInvalidation` mask set in https://chromium-review.googlesource.com/c/chromium/src/+/2454053 , is the mask really needed in combination with `InvalidateFilterData`  ?

patches/chromium/replace_clearfilterdata_with_invalidatefilterdata.patch
```diff
@@ -0,0 +1,133 @@
...
++++ b/third_party/blink/renderer/core/layout/svg/svg_resources_cache.cc
+@@ -99,9 +99,8 @@ void SVGResourcesCache::ClientLayoutChanged(LayoutObject& object) {
+     invalidation_flags = SVGResourceClient::kBoundariesInvalidation;
+   bool filter_data_invalidated = false;
+   if (resources->Filter()) {
+-    filter_data_invalidated = client->ClearFilterData();
+-    invalidation_flags |=
+-        filter_data_invalidated ? SVGResourceClient::kPaintInvalidation : 0;
++    client->InvalidateFilterData();
++    invalidation_flags |= SVGResourceClient::kPaintInvalidation;
```
deepak1556, line 130: Same question about `invalidation_flags`

deepak1556 (changes requested):

patches/chromium/replace_clearfilterdata_with_invalidatefilterdata.patch
```diff
@@ -0,0 +1,133 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Raymond Zhao <raymondzhao@microsoft.com>
+Date: Wed, 10 Feb 2021 12:03:34 -0800
+Subject: Replace ClearFilterData with InvalidateFilterData
+
+This is a hand-patch of https://chromium-review.googlesource.com/c/chromium/src/+/2566992.
```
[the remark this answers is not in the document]
rzhao271: I just took a look at https://chromium-review.googlesource.com/c/chromium/src/+/2454053 and it seems like it changes around too much for me to apply.
I made this patch by manually replacing instances of `ClearFilterData` with `InvalidateFilterData`, with the `InvalidateFilterData` from https://chromium-review.googlesource.com/c/chromium/src/+/2566992

patches/chromium/replace_clearfilterdata_with_invalidatefilterdata.patch
rzhao271: I'll leave the `MarkClientForInvalidation` code along with the masks/flags for now, as adding in the https://chromium-review.googlesource.com/c/chromium/src/+/2454053 patch requires a lot of changes, for example adding in an entire `SVGResourceInvalidator`.

patches/chromium/replace_clearfilterdata_with_invalidatefilterdata.patch
rzhao271: Just confirmed that `invalidation_flags` is indeed not needed here, and same with the previous comment above.

patches/chromium/replace_clearfilterdata_with_invalidatefilterdata.patch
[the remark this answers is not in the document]
rzhao271: Synced offline. Updated the formatting and included a link to https://chromium-review.googlesource.com/c/chromium/src/+/2454053.

deepak1556 (approved):

zcbenz (approved):

PR #12696, opened 2018-04-24
Refactor NativeWindow (Part 9): Use views::Widget on macOS

zcbenz:
This PR uses `views::Widget` to replace current manual creation of `NSWindow`, it would allow us to put arbitrary `views::View` inside `TopLevelWindow` on macOS.

To make the migration graceful, I'm still using our own logics for managing window's styles and states, so this change should have minimal affects on window's behaviors. But it can still cause quirks and regressions since Chromium is adding a lot of custom logics to `NSWindow` in `views::Widget`, please let me know if you noticed any quirk after this change.

PR #6068, opened 2016-06-15
Redispatch events to handle native OSX shortcuts

kevinsawicki:
This pull request pulls in code from [command_dispatcher.mm](https://cs.chromium.org/codesearch/f/chromium/src/ui/base/cocoa/command_dispatcher.mm) that redispatches events in an attempt to correctly handle native keyboard shortcuts like `Ctrl+F2` and `Cmd-~` on Mac OS X.

### Tested Scenarios

- [x] Menu-based shortcuts like `Cmd-R`, `Cmd-A`, `Cmd-Alt-I`.
- [x] Native shortcuts like `Cmd-~` and `Ctrl-F2`
- [x] Calling `preventDefault()` on a `keydown` event to disable something native like `Cmd-A` to select all
- [x] Typing and keybindings in `<input>` elements 
- [x] Docked devtools window
- [x] Undocked devtools window
- [ ] Others?

If anyone is able to test this branch for their app and verify menus and shortcuts still work as expected, that would be really helpful. /cc @electron/maintainers 

Closes #5246 
Closes #6016 
Depends on https://github.com/electron/brightray/issues/227

Related reading:

https://www.chromium.org/developers/os-x-keyboard-handling
https://github.com/nwjs/nw.js/commit/6658fbb839a9623f373b51fee37370f5bf500a6d

Issue #6016, opened 2016-06-12
[OS X] Ctrl+F2 keyboard shortcut to focus menu bar does not work


miniak:
* Electron version: 1.2.2
* Operating system: OS X 10.9.5

Pressing Ctrl+F2 should focus the menu bar. It works in all native Mac applications. It does nothing in Electron apps. This is a quite serious accessibility issue.

PR #27701, opened 2021-02-10
fix: restore window event redispatching on mac

### LINKED ISSUE #27268: [OS X] Ctrl+F2 keyboard shortcut to focus menu bar does not work (accessibility)
### Preflight Checklist

* [ x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [ x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [ x] I have searched the issue tracker for an issue that matches the one I want to file, without success. -> I can see one that was resolved in 2016, but it seems to have regressed.

### Issue Details

* **Electron Version:** 9.0.4
* **Operating System:** Mac OS 10.15.7
* **Last Known Working Electron version:**(unknown) 

### Expected Behavior
Pressing CTRL F2 should focus on the app menu bar (Apple Logo)

### Actual Behavior
Pressing CTRL F2 does nothing.
The only way to focus the app menu bar is to press CTRL F3 (which focuses the Dock), then press CTRL F2. I am not able to get around this in my project and it does not work in any production electron apps (e.g. VSCode/Slack etc...). Native apps do not have an issue with this.

### To Reproduce
CTRL F2 does not focus app menu bar.

This was reported and fixed in 2016 : https://github.com/electron/electron/issues/6016 , but it seems to have regressed.

clavin:
#### Description of Change
A past refactor of what is now `ElectronNSWindow` removed `EventDispatchingWindow` from its inheritance chain, without also adding `ElectronNSWindow` where `EventDispatchingWindow` was previously used. This small change restores that functionality to `ElectronNSWindow`.

Fixes #27268 which is a modern duplicate of #6016 which was fixed in #6068 but then became unfixed in #12696 and has been (presumably) broken ever since.

cc @MarshallOfSound

#### Checklist

#### Release Notes

Notes: Fixed OS-level shortcuts on macOS (e.g. Ctrl + F2, ⌘ + ~).

### DIFF
--- a/shell/browser/api/electron_api_web_contents_mac.mm
+++ b/shell/browser/api/electron_api_web_contents_mac.mm
@@ -6,6 +6,7 @@
 #include "shell/browser/api/electron_api_web_contents.h"
 #include "shell/browser/ui/cocoa/event_dispatching_window.h"
 #include "shell/browser/web_contents_preferences.h"
+#include "ui/base/cocoa/command_dispatcher.h"
 #include "ui/events/keycodes/keyboard_codes.h"
 
 #import <Cocoa/Cocoa.h>
@@ -58,12 +59,23 @@ bool WebContents::PlatformHandleKeyboardEvent(
       [[NSApp mainMenu] performKeyEquivalent:event.os_event])
     return true;
 
+  // Let the window redispatch the OS event
   if (event.os_event.window &&
       [event.os_event.window isKindOfClass:[EventDispatchingWindow class]]) {
     [event.os_event.window redispatchKeyEvent:event.os_event];
     // FIXME(nornagon): this isn't the right return value; we should implement
     // devtools windows as Widgets in order to take advantage of the
     // pre-existing redispatch code in bridged_native_widget.
+    return false;
+  } else if (event.os_event.window &&
+             [event.os_event.window
+                 conformsToProtocol:@protocol(CommandDispatchingWindow)]) {
+    NSObject<CommandDispatchingWindow>* window =
+        static_cast<NSObject<CommandDispatchingWindow>*>(event.os_event.window);
+    [[window commandDispatcher] redispatchKeyEvent:event.os_event];
+    // FIXME(clavin): Not exactly sure what to return here, likely the same
+    // situation as the branch above. If a future refactor removes
+    // |EventDispatchingWindow| then only this branch will need to remain.
     return false;
   }
 


clavin:
⚠️ Just realized I missed the one reference to `EventDispatchingWindow`: devtools detatched windows. Those would (and do) break with this current fix 😭. Working on seeing if I can somehow bring the `EventDispatchingWindow` functionality to `ElectronNSWindow` without duplicating anything, therefore keeping the working devtools windows while also getting working app windows.

clavin:
Alright, this new solution uses the `CommandDispatcher` in `NativeMacWidgetNSWindow` (of which `ElectronNSWindow` inherits from) to redispatch events for electron windows. Both app windows and detatched devtools windows (which still use the older `EventDispatchingWindow`, whose code was originally taken from `NativeMacWidgetNSWindow`) now handle all these keyboard shortcuts correctly, e.g. <kbd>⌘</kbd>+<kbd>\`</kbd> and <kbd>Ctrl</kbd>+<kbd>F2</kbd>.

In the future, a refactor could simplify the code here by changing detatched devtools windows to inherit from `NativeMacWidgetNSWindow`/`ElectronNSWindow` instead (or something of that nature), which would remove the `EventDispatchingWindow` class and its special handling here altogether.

I couldn't figure out a way to test this & couldn't find any tests for this functionality in the chromium source, so I haven't added any tests in this PR.

MarshallOfSound (approved):

clavin:
Just to 🧠 brain dump in a more digestible way:

This is what the inheritance looked like<sup>1</sup> before #6016:

Inheritance before issue 6016

In this version, certain key shortcuts<sup>2</sup> like <kbd>Ctrl</kbd> + <kbd>F2</kbd> were getting lost and not carrying out their expected functionality.

In Chromium, this issue is handled by a class called `CommandDispatcher` which, essentially, will send some incoming events back to `NSApp` ("redispatching") and ignore those events when they come back down the second time<sup>3</sup>.

So, this original issue was fixed in #6068 by introducing a new class called `EventDispatchingWindow` between the windows in electron and `NSWindow`:

Inheritance after PR 6068

This `EventDispatchingWindow` class is a thin wrapper around `NSWindow` that provides only that redispatching functionality that `CommandDispatcher` provides.

A few years later, a refactor in #12696 changed the inheritance of windows created by apps:

Inheritance after the refactor in PR 12696

This refactored `ElectronNSWindow` to inherit from Chromium's `NativeWidgetMacNSWindow` class instead of `EventDispatchingWindow`. However, the code where redispatching actually happens was not updated with this change, meaning windows created by electron apps no longer had this fix. (Undocked/detatched devtools windows, however, were left unchanged and thus still worked.)

It is important to note here that `NativeWidgetMacNSWindow` owns a `CommandDispatcher` of its own. This is relevant because the functionality of `EventDispatchingWindow` comes from `CommandDispatcher`.

This PR takes the simplest route to fix #27268:

Inheritance after this PR

The code where redispatching actually happens is extended to also recognize windows that have a `CommandDispatcher` (`ElectronNSWindow` fits in this category) and redispatch using that.

### An alternate route

This PR opts to take a simple path to fix a simple oversight in a refactor. However, this issue could also be fixed by having detatched devtools windows implemented as a window with a `CommandDispatcher` (this is represented by a protocol named `CommandDispatchingWindow`), e.g. `NativeWidgetMacNSWindow`. This would allow us to remove the slightly-redundant `EventDispatchingWindow` class.

---

### What codebase does X come from?

Electron:
- `EventDispatchingWindow`
- `ElectronNSWindow`

Chromium:
- `CommandDispatcher`
- `CommandDispatchingWindow`
- `NativeWidgetMacNSWindow`

### Notes

<sup>1</sup> Some parts of the chain are left out of each image, e.g. `EventDispatchingWindow` actually inherits directly from a class called `UnderlayOpenGLHostingWindow`, but that class has nearly zero functionality and isn't relevant here. Additionally, `ElectronNSWindow` had different names in the past when some of these changes took place. Basically, minor details have been cut in this brain dump.

<sup>2</sup> The only shortcut that I could find that failed like this was <kbd>Ctrl</kbd> + <kbd>F2</kbd>, which focuses the menu bar/apple logo on macOS. Other related shortcuts, like <kbd>Ctrl</kbd> + <kbd>F3</kbd> (focuses the dock) or <kbd>Cmd</kbd> + <kbd>\`</kbd> (switches between windows in the same app) mysteriously work without this event redispatching.

<sup>3</sup> I do not know why this fixes the issue, but it does.

zcbenz (approved):

zcbenz:
Very nice investigation 👍 !

mxdbx:
Is there any plan to backport this to 9-x-y?

clavin:
@mxdbx I'm not planning on it since that version of electron is no longer supported 😕

PR #27711, opened 2021-02-11

sentialx:
#### Description of Change

Backport of #27007 

#### Release Notes


zcbenz (changes requested):
@sentialx Build is failing:

```
../../electron/shell/browser/api/electron_api_browser_window.cc:378:3: error: use of undeclared identifier 'BaseWindow'
  BaseWindow::SetTopBrowserView(value, args);
  ^
1 error generated.
```

zcbenz (approved):

Issue #27188, opened 2021-01-05
Using window.history.back() and window.history.forward() does not update BrowserWindow titles.


eengoron:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  
  * 8.1.1
  * 11.1.1
* **Operating System:**
  * macOS 10.14, macOS 10.15
* **Last Known Working Electron version:**
  * N/A

### Expected Behavior

When using `window.history.back()` or `window.history.forward()`, the title of the BrowserWindow should update to the title of the page you're navigating to. 

### Actual Behavior

When using `window.history.back()` or `window.history.forward()`, the title of the BrowserWindow does not update to the title of the page you're navigating to. 

### To Reproduce

1) Download and start the `electron-quick-start` example
2) Open Devtools > Console
3) Enter `window.location = 'https://www.google.com'` to get to Google (The window title will update to Google)
4) Enter `window.history.back()` to go back to the page that was originally opened. 

The window title will still say `Google` instead of `Hello World!` as it's supposed to. 

### Screenshots

Here's a gif where I navigate to two websites and then use `window.history.back()` to show this issue:
![title_doesnot_update](https://user-images.githubusercontent.com/2817521/103596854-dc39c300-4ecc-11eb-88a3-18d7bb209018.gif)

### Additional Information:
This does work in the normal Chrome browser, so this looks to be an Electron specific issue.

ckerr:
Confirmed. I don't think this is platform-specific; FWIW I'm seeing it on Linux as well

gdavidkov:
It is still an issue. Please repeat the steps in the description. The title isn't being updated.

PR #27724, opened 2021-02-12
fix: add support for title change with window history navigation

mlaurencin:
#### Description of Change

I am still working on improving this PR, so any additional insights into better ways to accomplish this (or other general review) would be appreciated.

Closes https://github.com/electron/electron/issues/27188

This PR restores the functionality where the title of a browser window is properly updated when navigating through its session history. Before, when you tried to navigate to a previously loaded page in the window history (such as with commands like `window.history.back()`) the title of the page would not be updated to reflect this navigation. Now the window's title will be properly updated.

cc @MarshallOfSound , @nornagon , @zcbenz  

**Further Notes (for more context):**

The code path that these navigation commands were initially following was changed with [this PR](https://github.com/electron/electron/pull/22336), specifically with [this commit](https://github.com/electron/electron/pull/22336/commits/56ea879a33b581e7ac3cff4e21df68375281f8b4) to enable reusing renderer processes by default. 

Where this change matters is in [this function](https://source.chromium.org/chromium/chromium/src/+/master:content/browser/web_contents/web_contents_impl.cc;l=5996;drc=2ac64302ae161cd6b5e4b1254497bdf5fd6d3415?ss=chromium&start=21) `WebContentsImpl::UpdateTitleForEntry` in Chromium. What I have determined is that when the navigation takes place, the information of the pending page is passed into the function by `NavigationEntryImpl* entry`. Then when the function checks the title that is trying to be set, `final_title`, it sees that it is the same as the entry's title and does not set it. However, [the line at the end here](https://source.chromium.org/chromium/chromium/src/+/master:content/browser/web_contents/web_contents_impl.cc;l=6024;drc=2ac64302ae161cd6b5e4b1254497bdf5fd6d3415?ss=chromium&start=21) `observers_.NotifyObservers(&WebContentsObserver::TitleWasSet, entry);` informs the observers that a title was set, so by bailing early the window is not informed that it needs to update its title. Thus the title stays the same as the one from before.

This was not a problem previously as a new url was loaded rather than actually navigating in the history. In this way, the `entry` did not yet have its title set, thus making the title `""` and the function did not return early.

**Current Issues:**

1.  In the case that `final_title` and `entry->GetTitle()` are different, `TitleWasSet` will be called twice: once on the Electron side and once on the Chromium side.
2.  Since I added this change into `WebContents::DidStartNavigation` I noticed that it gets called quite often (even in some cases where it appears that you did not initiate navigation).

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:  None

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -1636,8 +1636,11 @@ void WebContents::DidStartNavigation(
     content::NavigationHandle* navigation_handle) {
   content::NavigationEntry* entry =
       web_contents()->GetController().GetPendingEntry();
-  if (entry)
+  // If a history entry has been made and the current window title is different,
+  // proceed with setting the new title
+  if (entry && (WebContents::GetTitle() != entry->GetTitle())) {
     WebContents::TitleWasSet(entry);
+  }
   EmitNavigationEvent("did-start-navigation", navigation_handle);
 }
 
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -1703,11 +1703,11 @@ void WebContents::DidFinishNavigation(
 
   // If a history entry has been made and the forward/back call has been made,
   // proceed with setting the new title
-  if (entry &&
-      ui::PageTransitionTypeIncludingQualifiersIs(
-          ui::PAGE_TRANSITION_FORWARD_BACK,
-          ui::PageTransitionFromInt(navigation_handle->GetPageTransition() &
-                                    ui::PAGE_TRANSITION_FORWARD_BACK)))
+  ui::PageTransition transition =
+      ui::PageTransitionFromInt(navigation_handle->GetPageTransition() &
+                                ui::PAGE_TRANSITION_FORWARD_BACK);
+  if (entry && ui::PageTransitionTypeIncludingQualifiersIs(
+                   ui::PAGE_TRANSITION_FORWARD_BACK, transition))
     WebContents::TitleWasSet(entry);
 }
 
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -1702,8 +1702,7 @@ void WebContents::DidFinishNavigation(
 
   // If a history entry has been made and the forward/back call has been made,
   // proceed with setting the new title
-  if (entry &&
-      (entry->GetTransitionType() & ui::PAGE_TRANSITION_FORWARD_BACK))
+  if (entry && (entry->GetTransitionType() & ui::PAGE_TRANSITION_FORWARD_BACK))
     WebContents::TitleWasSet(entry);
 }
 
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -1702,11 +1702,8 @@ void WebContents::DidFinishNavigation(
 
   // If a history entry has been made and the forward/back call has been made,
   // proceed with setting the new title
-  ui::PageTransition transition =
-      ui::PageTransitionFromInt(navigation_handle->GetPageTransition() &
-                                ui::PAGE_TRANSITION_FORWARD_BACK);
-  if (entry && ui::PageTransitionTypeIncludingQualifiersIs(
-                   ui::PAGE_TRANSITION_FORWARD_BACK, transition))
+  if (entry &&
+      (entry->GetTransitionType() & ui::PAGE_TRANSITION_FORWARD_BACK))
     WebContents::TitleWasSet(entry);
 }
 
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -1700,6 +1700,9 @@ void WebContents::DidFinishNavigation(
   }
   content::NavigationEntry* entry = navigation_handle->GetNavigationEntry();
 
+  // This check is needed due to an issue in Chromium
+  // Check the Chromium issue to keep updated:
+  // https://bugs.chromium.org/p/chromium/issues/detail?id=1178663
   // If a history entry has been made and the forward/back call has been made,
   // proceed with setting the new title
   if (entry && (entry->GetTransitionType() & ui::PAGE_TRANSITION_FORWARD_BACK))
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -1634,6 +1634,10 @@ void WebContents::RenderFrameDeleted(
 
 void WebContents::DidStartNavigation(
     content::NavigationHandle* navigation_handle) {
+  content::NavigationEntry* entry =
+      web_contents()->GetController().GetPendingEntry();
+  if (entry)
+    WebContents::TitleWasSet(entry);
   EmitNavigationEvent("did-start-navigation", navigation_handle);
 }
 
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -3,7 +3,6 @@
 // found in the LICENSE file.
 
 #include "shell/browser/api/electron_api_web_contents.h"
-#include "ui/base/page_transition_types.h"  //FIXME(@mlaurencin): Decide if I need this or to include something else for the enums
 
 #include <limits>
 #include <memory>
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 #include "shell/browser/api/electron_api_web_contents.h"
+#include "ui/base/page_transition_types.h"  //FIXME(@mlaurencin): Decide if I need this or to include something else for the enums
 
 #include <limits>
 #include <memory>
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
Test failure appears to be legitimate.  In order to avoid double-emitting the "title-updated" event you'll probably have to add logic to ensure you don't call `TitleWasSet` without the title actually changing

deepak1556:
Couple of pointers:

> This was not a problem previously because when the renderer process is not reused, a new url is loaded rather than actually navigating in the history.

Irrespective of whether `RenderProcessReuse` is set or not, renderer initiated navigations like history navigations tend to follow the same security policy of `SitePerProcess` model. So when you navigate back or forth, sites from different origins will not be loaded in the same process, it would break the model.

> In the case that final_title and entry->GetTitle() are different, TitleWasSet will be called twice: once on the Electron side and once on the Chromium side.

* Don't set the title for all navigations under `BeginNavigation`, title should be set only when the navigation has committed, check the state using the `NavigationHandle` under the `NavigationEntry`, also we are specifically interested in render side navigations, also check for that `NavigationHandle::IsRendererInitiated`

> Then when the function checks the title that is trying to be set, final_title, it sees that it is the same as the entry's title and does not set it

Were you able to narrow down why that condition was set, if a process doesn't navigate from `about:blank` which happens during this history navigation, then `final_title` will be same as the `entry` title. I believe this is a problem at the //content layer, I quickly tried to check what `//content/shell:content_shell` does (without any of our patches applied). It also doesn't get the correct observer events and the window title is not updated. Its best to solve this issue in upstream and backport the fix as a next step.

mlaurencin:
Some follow-ups:

> So when you navigate back or forth, sites from different origins will not be loaded in the same process, it would break the model.

Thank you for that explanation. I'm going to remove the "because when the renderer process is not reused" then to avoid the implied cause-effect relationship. I believe the loading new url vs navigating in the history is still relevant though.

> Don't set the title for all navigations under `BeginNavigation`, title should be set only when the navigation has committed

Just to confirm, by `BeginNavigation` are you referring to the `WebContents::DidStartNavigation` function? From my understanding of [this document](https://chromium.googlesource.com/chromium/src/+/master/docs/navigation.md#navigation) the function `WebContents::DidFinishNavigation` is called after the navigation has committed, so I see that as a good alternative as opposed to performing checks with the `NavigationHandle`.

> I quickly tried to check what `//content/shell:content_shell` does (without any of our patches applied). It also doesn't get the correct observer events and the window title is not updated. Its best to solve this issue in upstream and backport the fix as a next step.

I am not entirely clear on what you did to test with this description. In terms of narrowing down why the condition was set, from what I have observered the `entry` is what is being navigated to in the history. I do not follow your logic for why `final_title` will then necessarily be the same as the `entry` title, but it definitely is not set by the current title of the window. That is as much as I have been able to narrow down. Are others viewing this as an upstream issue as well?

deepak1556:
> I believe the loading new url vs navigating in the history is still relevant though.

To clarify, what matters here is whether the navigation was browser side navigation or render side navigation.

In electron, without RenderProcessReuse mode all navigations are browser side so that we can force the process restart and under this mode the page load will start from `about:blank` whose title is `""` and thats why the `final_title == entry->GetTitle()` check will not get hit and title updates have been working fine.

> I am not entirely clear on what you did to test with this description.

* Build `//content/shell:content_shell` without applying any of patches from `//electron/patches`.
* Launch content_shell , this will load google.com by default
* open devtools
* `window.location = "https://electronjs.org"` 
* window title will update to document title from `electronjs.org`
* `window.history.back()`
* See that the window title remains unchanged

> I do not follow your logic for why final_title will then necessarily be the same as the entry title, but it definitely is not set by the current title of the window.

When the process doesn't have any `NavigationEntry`, (i-e) there was no RFH swap inside the process then the only entry available is the currently committed one, and which is true in the example cases we are testing. Hence the `final_title` which Blink provides based on the document title is the same as title entry holds when it was a pending entry, unless the navigation started from `about:blank` which isn't the case as seen above.

deepak1556:
>  From my understanding of this document the function WebContents::DidFinishNavigation is called after the navigation has committed, so I see that as a good alternative as opposed to performing checks with the NavigationHandle

That is correct, I was suggesting to verify `NavigationHandle::IsRendererInitiated` additionally so that we can isolate render side navigations which is what we are interested in this issue.

mlaurencin:
> That is correct, I was suggesting to verify `NavigationHandle::IsRendererInitiated` additionally so that we can isolate render side navigations which is what we are interested in this issue.

While the navigation done by `window.location=` is registering as being initiated by the renderer, the `window.history.back()` navigation is not. I checked in both `WebContents::DidStartNavigation` and `WebContents::DidFinishNavigation` and the `navigation_handle` pointer returns `false` in both. On a similarly note, after moving the implementation to `WebContents::DidFinishNavigation` I noticed that the `WebContents` title has been updated and the `entry` is no longer accessible. The title of the browser window still does not get updated though as I believe that would happen [here](https://github.com/electron/electron/blob/47792d2ecb7b52085556eb7581128bd1973a8c7a/shell/browser/api/electron_api_web_contents.cc#L1717).

In terms of then getting this fix together for the short-time solution of having it backported, it seems to be fixing the issue with the title difference check and call to `WebContents::TitleWasSet` in `WebContents::DidStartNavigation`. I am not sure however why the current tests are failing.

deepak1556:
> While the navigation done by window.location= is registering as being initiated by the renderer, the window.history.back() navigation is not. I checked in both WebContents::DidStartNavigation and WebContents::DidFinishNavigation and the navigation_handle pointer returns false in both.

Thanks for checking on that, turns out history navigations are indeed browser initiated navigations. The following trace confirmed it,

```
{
    "result_file": "/tmp/trace_event.log",
    "trace_config": {
        "record_mode": "record-until-full",
        "included_categories": [
            "toplevel",
            "navigation",
            "browser",
            "mojom",
            "renderer_host",
            "renderer"
        ],
        "excluded_categories": [
            "*"
        ]
    }
}
```

The difference is that history navigations swap to the already committed entries rather than starting the navigation from the initial blank document. Check for `ClassifyNavigation` entry in the above trace result.

Additionally, from https://chromium.googlesource.com/chromium/src/+/master/docs/session_history.md#navigating-with-replacement

> Some types of navigations can replace the previously committed joint session history item for a frame, rather than creating a new item.
> The first non-blank URL after the initial empty document (unless the frame was explicitly created with about:blank as the URL).

Confirms my statement in https://github.com/electron/electron/pull/27724#issuecomment-778583171

> In terms of then getting this fix together for the short-time solution of having it backported, it seems to be fixing the issue with the title difference check and call to WebContents::TitleWasSet in WebContents::DidStartNavigation

Is there a reason this needs a temporary fix in Electron considering the issue has been present for a quite a while. Can we try getting this fixed in upstream ?

mlaurencin:
> Is there a reason this needs a temporary fix in Electron considering the issue has been present for a quite a while. Can we try getting this fixed in upstream ?

Given there is a simple and reasonable temporary fix possible, I think it would better to implement it. It also wouldn’t cause an issue when the upstream fix is put in. So all in all, I think it would be good to fix this problem now if users are experiencing it.

zcbenz (commented):
You can call `navigation_handle->GetPageTransition()` to test whether it is a back forward transition.

And you can use `navigation_handle->GetNavigationEntry()` to get the associated navigation entry instead of using the pending entry, so you can compare the entry itself instead of the title to fix duplicate `TitleWasSet` calls.

In this way you can probably avoid getting messy with browser/renderer initialized navigations.

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1698,6 +1698,16 @@ void WebContents::DidFinishNavigation(
...
   }
+  content::NavigationEntry* entry = navigation_handle->GetNavigationEntry();
+
+  // If a history entry has been made and the forward/back call has been made,
+  // proceed with setting the new title
+  ui::PageTransition transition =
+      ui::PageTransitionFromInt(navigation_handle->GetPageTransition() &
+                                ui::PAGE_TRANSITION_FORWARD_BACK);
+  if (entry && ui::PageTransitionTypeIncludingQualifiersIs(
+                   ui::PAGE_TRANSITION_FORWARD_BACK, transition))
```
deepak1556, line 1709: Why the int conversion ?

```suggestion
  if (entry &&
      (entry->GetTransitionType() & ui::PAGE_TRANSITION_FORWARD_BACK))
```

deepak1556:
FYI, upstream issue https://bugs.chromium.org/p/chromium/issues/detail?id=1178663 for further investigation on the root cause.

shell/browser/api/electron_api_web_contents.cc
mlaurencin: This stemmed from some logging I was doing before the if statement. This is more efficient though, so thank you.

codebytere (approved):

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1698,6 +1698,12 @@ void WebContents::DidFinishNavigation(
            frame_process_id, frame_routing_id);
     }
   }
+  content::NavigationEntry* entry = navigation_handle->GetNavigationEntry();
+
+  // If a history entry has been made and the forward/back call has been made,
+  // proceed with setting the new title
```
zcbenz, line 1704: Can you add a link to the Chromium bug?

mlaurencin: Yes, just added it to the latest commit.

zcbenz (approved):

PR #27729, opened 2021-02-15
test: begin next test after last app is quit

zcbenz:
#### Description of Change

The app used for testing crashReporter may crash unintentionally after intentionally crashing a WebContents, which would then send 2 crashes in one test, and making the next tests show wrong results.

```
crashReporter module should send minidump when sandboxed renderer crashes api-crash-reporter-spec.ts 643 ms
Error message:
ptype: expected 'browser' to equal 'renderer'
Error stack trace:
AssertionError: ptype: expected 'browser' to equal 'renderer'
    at checkCrash (electron\spec-main\api-crash-reporter-spec.ts:39:35)
    at Context.<anonymous> (electron\spec-main\api-crash-reporter-spec.ts:154:7)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

This PR makes sure the next test is only started after the test app used by previous exited.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/spec-main/api-crash-reporter-spec.ts
+++ b/spec-main/api-crash-reporter-spec.ts
@@ -141,16 +141,18 @@ ifdescribe(!isLinuxOnArm && !process.mas && !process.env.DISABLE_CRASH_REPORTER_
   describe('should send minidump', () => {
     it('when renderer crashes', async () => {
       const { port, waitForCrash } = await startServer();
-      runCrashApp('renderer', port);
+      const waitAppExit = runCrashApp('renderer', port);
       const crash = await waitForCrash();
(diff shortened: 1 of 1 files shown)


zcbenz:
Note that the root cause for the failing test -- the test app crashed unexpectedly, is not fixed by this PR, and I'll address it with another PR. This PR means to make the tests more robust.

zcbenz:
Hmm this does not fix the crash in CI, I'll try another approach.

mbp15:
[]()

PR #27730, opened 2021-02-15
fix: check WebContents before emitting render-process-gone event

zcbenz:
#### Description of Change

Fix a crash that caused by `render-process-gone` event accessing WebContents after user has destroyed it in the `crashed` event.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix crash when destroying WebContents in the `crashed` event.


zcbenz:
This should be merged after https://github.com/electron/electron/pull/27731, which would fix the CI failure.

miniak (approved):

codebytere (approved):

PR #27741, opened 2021-02-16
refactor: load preload script directly as a string

miniak:
Backport of #27546

See that PR for details.

Notes: no-notes


codebytere (approved):

petef19:
Hello, 

looking to load preload directly inline (as a string) in main.js instead of from an external file.

Is this what this PR does ?

If so, how can I use it ? What is the syntax ?

I'm on Electron 12.0.5, Win 10 x64.

Thanks.

Issue #27503, opened 2021-01-26
Not able to access contents of an IFRAME in OOPIFs


wayvad:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details
Electron version: 10.2
OS: Windows
Last known working Electron version: 8.2

### Expected Behavior
Be able to use DevTools or webdriver to view / automate contents in an out-of-process iframe.

### Actual Behavior
Starting with Electron 9, no longer to use DevTools to inspect elements inside OOPIF. Please see this Fiddle for repro: https://gist.github.com/81a58bffae7d48d2f1a4c4cb7300c86c

More info from Deepak:
      Definitely a bug in electron not seeing the same issue with chrome, used a local http server to load the above html file in chrome and electron, chrome displays the contents for the OOPIF. 

  I just started to narrow down the version this first regressed, it is somewhere between 9.0.0-beta.9 and 9.0.0-beta.12

### To Reproduce
Please use this Fiddle: https://gist.github.com/81a58bffae7d48d2f1a4c4cb7300c86c

### Screenshots

![image](https://user-images.githubusercontent.com/40037392/105921292-85b43600-5fed-11eb-919c-66006c447749.png)

### Additional Information

I contacted Milan Burda and Deepak Mohan to confirm that this is indeed an Electron bug.

ckerr:
@wayvad can you walk me through this? I think I might be looking at the wrong behavior in your test if you're seeing the breakage somewhere in [9.0.0-beta.9..9.0.0-beta.12].

When I test in the latest 8 version (8.5.5) I am able to click the inspector's "Select an element in the page to inspect it" button, then mouseover the items inside iframe, and the elements are highlighted and inspected. When I try this in 9.0.0-beta.1, I no longer see those items highlighted and the element tree stops at `<iframe></iframe>`. Which sounds like what you're describing, but appears to have been introduced sometime before `9.0.0-beta.1`.

So if I'm looking at the wrong thing, help me out so I can look at the right behavior.

Or if this is the right thing, can you double-check your versions to confirm the beta 9 & 12 numbers?

ckerr:
[...]

PR #26432, opened 2020-11-10
fix: disable swapping BrowsingInstance triggered by COOP in <webview>

sentialx:
#### Description of Change
Since Electron is using `SiteInstance::CreateForURL` instead of `SiteInstance::CreateForGuest`, Chromium tries to swap the browsing instance when COOP is triggered, but that's not supported in `<webview>`. So I made a patch disabling COOP only for guest WebContents.

Could this be backported to 9-x-y?

Fixes #25469, #25421

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Fixed loading twitter.com and other websites enforcing Cross-Origin-Opener-Policy in `<webview>`.

PR #27744, opened 2021-02-16
fix: create guest site instances for guest web contents.

ppontes:
Chromium does not swap processes for navigations in webview tag guests. Electron apps have come to rely on this for debugging and automated testing of iframes inside webviews.

With [this change](https://chromium.googlesource.com/chromium/src.git/+/bcaccaf3027c66c096d8bcf28dc2c67cf3a40287), Chromium moved from relying on a site instance's URL being a guest scheme to explicitly marking site instances as having been created for guest frames.

Electron must follow this change when creating site instances for web contents used for webviews, so that these site instances are properly marked as guests and the Chromium behaviour of reusing the same process inside webviews is kept.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed an issue that made iframes inside webviews inaccessible to DevTools and automated test tools (#27503).

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -719,8 +719,8 @@ WebContents::WebContents(v8::Isolate* isolate,
   std::unique_ptr<content::WebContents> web_contents;
   if (IsGuest()) {
     scoped_refptr<content::SiteInstance> site_instance =
-        content::SiteInstance::CreateForURL(session->browser_context(),
-                                            GURL("chrome-guest://fake-host"));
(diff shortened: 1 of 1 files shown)


deepak1556:
https://github.com/electron/electron/pull/26432 also attempts at this, I am in favor of swapping the callsite to use `CreateForGuest` but there was concern about not being able to load webUI pages. Thoughts on that ?

nornagon:
It would be preferable from a security perspective if guest views _did_ swap processes. Can we figure out a way to inform the relevant developer tools that the process swap happened?

Issue #23640, opened 2020-05-18
heap use-after-free in OSR


nornagon:
Abridged report:

```
==529==ERROR: AddressSanitizer: heap-use-after-free on address 0x6030000d5680 at pc 0x55d10f042120 bp 0x7f03d170ef30 sp 0x7f03d170ef28
READ of size 1 at 0x6030000d5680 thread T6 (VizCompositorTh)
    #0 0x55d10f04211f in mojo::SequenceLocalSyncEventWatcher::SequenceLocalState::OnEventSignaled() ./../../mojo/public/cpp/bindings/lib/sequence_local_sync_event_watcher.cc:214:47
[...]
    #9 0x55d104fcd6a7 in viz::mojom::DisplayClientProxy::IsOffscreen(bool*) ./gen/services/viz/privileged/mojom/compositing/display_private.mojom.cc:789:28
    #10 0x55d11adfe260 in viz::OutputSurfaceProviderImpl::CreateSoftwareOutputDeviceForPlatform(unsigned long, viz::mojom::DisplayClient*) ./../../components/viz/service/display_embedder/output_surface_provider_impl.cc:234:25
[...]

0x6030000d5680 is located 16 bytes inside of 24-byte region [0x6030000d5670,0x6030000d5688)
freed by thread T6 (VizCompositorTh) here:
    #0 0x55d1015515dd in operator delete(void*) /b/s/w/ir/cache/builder/src/third_party/llvm/compiler-rt/lib/asan/asan_new_delete.cpp:160:3
[...]
    #9 0x55d10f0136da in mojo::internal::InterfacePtrStateBase::~InterfacePtrStateBase() ./../../mojo/public/cpp/bindings/lib/interface_ptr_state.cc:15:20
[...]
    #26 0x55d11ae4c3a0 in viz::FrameSinkManagerImpl::DestroyCompositorFrameSink(viz::FrameSinkId const&, base::OnceCallback<void ()>) ./../../components/viz/service/frame_sinks/frame_sink_manager_impl.cc:192:18
[...]

previously allocated by thread T6 (VizCompositorTh) here:
[...]
    #6 0x55d104fcd6a7 in viz::mojom::DisplayClientProxy::IsOffscreen(bool*) ./gen/services/viz/privileged/mojom/compositing/display_private.mojom.cc:789:28
[...]

PR #27774, opened 2021-02-18
fix: pass IsOffscreen via parameter instead of sync IPC method

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/23640.

The offscreen patch was sending a sync IPC message to call `IsOffScreen`, which caused memory corruptions since the related classes might be destroyed while waiting for the message to reply.

This PR refactors the offscreen patch to pass `IsOffScreen` via async parameter to avoid doing the sync IPC call.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


deepak1556 (approved):

jkleinsc (approved):

nornagon (approved):
seems like a much better approach, thanks!

hopefully this unblocks https://github.com/electron/electron/pull/23570 too!

codebytere (approved):

PR #27786, opened 2021-02-18
fix: ensure owner window is valid in ResetBrowserViews

codebytere:
#### Description of Change

Ensure that the `owner_window` is valid before potentially calling any methods on it. 

In the initial PR we checked it in [`AddBrowserView`](https://github.com/electron/electron/pull/27000/files#diff-9e3935b7f7b31cb73a8e976e9725bab5c20c499dc5d1e33c7c7f62e3f90b72c9R765) but not in [`ResetBrowserViews`](https://github.com/electron/electron/pull/27000/files#diff-9e3935b7f7b31cb73a8e976e9725bab5c20c499dc5d1e33c7c7f62e3f90b72c9R1086) and we should check both.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a potential crash when resetting `BrowserView`s.

### DIFF
--- a/shell/browser/api/electron_api_base_window.cc
+++ b/shell/browser/api/electron_api_base_window.cc
@@ -1118,7 +1118,7 @@ void BaseWindow::ResetBrowserViews() {
       // reset if the owner window is *this* window.
       if (browser_view->web_contents()) {
         auto* owner_window = browser_view->web_contents()->owner_window();
-        if (owner_window == window_.get()) {
+        if (owner_window && owner_window == window_.get()) {
           browser_view->web_contents()->SetOwnerWindow(nullptr);
           owner_window->RemoveBrowserView(browser_view->view());
         }


MarshallOfSound (approved):

shell/browser/api/electron_api_base_window.cc
```diff
@@ -1118,7 +1118,7 @@ void BaseWindow::ResetBrowserViews() {
       // reset if the owner window is *this* window.
       if (browser_view->web_contents()) {
         auto* owner_window = browser_view->web_contents()->owner_window();
-        if (owner_window == window_.get()) {
+        if (owner_window && owner_window == window_.get()) {
```
nornagon, line 1121: could we instead check `if (!window_.get()) return` at the top of the function?

codebytere: @nornagon this wouldn't fix the issue - the owner window could still be nullptr in that case

PR #27815, opened 2021-02-19
fix: check web_contents() for destroyed WebContents

zcbenz:
#### Description of Change

When Electron is quitting, it may happen that the `content::WebContents` is destroyed while `electron::api::WebContents` is still alive.  This was causing a few CI flakes.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -1078,7 +1078,7 @@ content::WebContents* WebContents::OpenURLFromTab(
     return nullptr;
   }
 
-  if (!weak_this)
+  if (!weak_this || !web_contents())
     return nullptr;
 
   content::NavigationController::LoadURLParams load_url_params(params.url);
@@ -1411,7 +1411,7 @@ void WebContents::RenderProcessGone(base::TerminationStatus status) {
   Emit("crashed", status == base::TERMINATION_STATUS_PROCESS_WAS_KILLED);
 
   // User might destroy WebContents in the crashed event.
-  if (!weak_this)
+  if (!weak_this || !web_contents())
     return;
 
   v8::Isolate* isolate = JavascriptEnvironment::GetIsolate();
@@ -1482,7 +1482,7 @@ void WebContents::DidFinishLoad(content::RenderFrameHost* render_frame_host,
   // ⚠️WARNING!⚠️
   // Emit() triggers JS which can call destroy() on |this|. It's not safe to
   // assume that |this| points to valid memory at this point.
-  if (is_main_frame && weak_this)
+  if (is_main_frame && weak_this && web_contents())
     Emit("did-finish-load");
 }
 
@@ -1886,6 +1886,7 @@ void WebContents::WebContentsDestroyed() {
   // also do not want any method to be used, so just mark as destroyed here.
   MarkDestroyed();
 
+  Observe(nullptr);  // this->web_contents() will return nullptr
   Emit("destroyed");
 
   // For guest view based on OOPIF, the WebContents is released by the embedder
@@ -2017,7 +2018,7 @@ void WebContents::LoadURL(const GURL& url,
   // ⚠️WARNING!⚠️
   // LoadURLWithParams() triggers JS events which can call destroy() on |this|.
   // It's not safe to assume that |this| points to valid memory at this point.
-  if (!weak_this)
+  if (!weak_this || !web_contents())
     return;
 
   // Required to make beforeunload handler work.
--- a/spec-main/api-web-contents-spec.ts
+++ b/spec-main/api-web-contents-spec.ts
@@ -2016,6 +2016,13 @@ describe('webContents module', () => {
       });
       contents.loadURL('about:blank').then(() => contents.forcefullyCrashRenderer());
     });
+
+    it('does not crash main process when quiting in it', async () => {
+      const appPath = path.join(mainFixturesPath, 'apps', 'quit', 'main.js');
+      const appProcess = ChildProcess.spawn(process.execPath, [appPath]);
+      const [code] = await emittedOnce(appProcess, 'close');
+      expect(code).to.equal(0);
+    });
   });
 
   it('emits a cancelable event before creating a child webcontents', async () => {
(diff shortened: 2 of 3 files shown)


codebytere (approved):

nornagon (commented):
it seems like this _must_ be missing situations where we expect `web_contents()` to be non-null. There are 100+ references to `web_contents()` in this file alone.

Is there some way we could be more safe here?

zcbenz:
> it seems like this _must_ be missing situations where we expect `web_contents()` to be non-null. There are 100+ references to `web_contents()` in this file alone.
> 
> Is there some way we could be more safe here?

The `content::WebContents` is only destroyed when the `electron::api::WebContents` is being destroyed, so for most cases `web_contents()` can not be null.

The only exception is when user decides to destroy the WebContents during an event, and the `electron::api::WebContents` continues to use `web_contents()` immediately after emitting the event, only in this case do we need to check if the WebContents has been destroyed. Fortunately there are only a few cases and we are already using the `weak_this` hack to do the check, this PR adds another check for a corner case on exit.

Ideally we should refactor the `electron::api::WebContents` class to eliminate the corner cases of lifetime at all, but I don't have a clear plan for now.

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1411,7 +1411,7 @@ void WebContents::RenderProcessGone(base::TerminationStatus status) {
   Emit("crashed", status == base::TERMINATION_STATUS_PROCESS_WAS_KILLED);
 
   // User might destroy WebContents in the crashed event.
-  if (!weak_this)
+  if (!weak_this || !web_contents())
```
nornagon, line 1414: Can this ever be called when `web_contents()` is nullptr? RenderProcessGone is a WebContentsObserver method, so if we've called `Observe(nullptr)` we shouldn't ever get an observer method call. If `Emit("crashed")` destroyed the WebContents, then the weak_this check will be enough to guard this.

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1482,7 +1482,7 @@ void WebContents::DidFinishLoad(content::RenderFrameHost* render_frame_host,
   // ⚠️WARNING!⚠️
   // Emit() triggers JS which can call destroy() on |this|. It's not safe to
   // assume that |this| points to valid memory at this point.
-  if (is_main_frame && weak_this)
+  if (is_main_frame && weak_this && web_contents())
```
nornagon, line 1485: here too.

nornagon (commented):
Hm... this change makes `web_contents()` return nullptr _before_ the `api::WebContents` is deleted, and there's an unspecified amount of time in between when `web_contents()` starts returning nullptr and when the `api::WebContents` is properly deleted. However, we have already called `MarkDestroyed()` so any entry point from JS will hit the [IsDestroyed()](https://github.com/electron/electron/blob/462de5f97a302987dc5fa5c222781ceed040f390/shell/common/gin_helper/function_template.h#L178) check. Further, since we're no longer observing the WebContents, no `WebContentsObserver` methods will be called. That leaves only `WebContentsDelegate` methods which can be newly be called in this state when `web_contents()` can return null.

There are many many such methods and I suspect that we will introduce new bugs in future where we implement WebContentsDelegate methods expecting that `web_contents()` will be non-null, but it isn't always. To wit, here are a few that reference `web_contents()` that haven't been addressed in this PR:
- `WebContents::HandleContextMenu`
- `WebContents::CloseContents`
- `WebContents::OnGoToEntryOffset`

I think this PR improves things a bit, but also sheds light on a bunch of problems that are already here.

zcbenz:
> this change makes `web_contents()` return nullptr _before_ the `api::WebContents` is deleted

The `WebContentsDestroyed` is called in the `content::WebContents`'s destructor with lots of things already getting cleared, so by the time this method is called we should not be accessing `web_contents()` anymore. Making following `web_contents()` calls crash with null pointer error would be better leaving them in an undecided state that may or may not crash.

https://source.chromium.org/chromium/chromium/src/+/master:content/browser/web_contents/web_contents_impl.cc;l=1004;drc=5d4bf842d53c6c47f6c3656cea4a1d2964ac60d5;bpv=0;bpt=1

> Can this ever be called when `web_contents()` is nullptr? RenderProcessGone is a WebContentsObserver method, so if we've called `Observe(nullptr)` we shouldn't ever get an observer method call. If `Emit("crashed")` destroyed the WebContents, then the weak_this check will be enough to guard this.

For most cases no, but there is an edge case for `app.quit`, as you can see in the test in this PR.

The call stack is something like this:

```
RenderProcessGone()
  => Emit("crashed")
    => Browser::Quit()
      => NativeWindow::Close()
        => BrowserWindow::OnWindowClosed()
          => api_web_contents_->DestroyWebContents(!Browser::Get()->is_shutting_down())
  => content::WebContents destroyed now
```

The `content::WebContents` is cleared by the `DestroyWebContents` method, usually it is destroyed asynchronously to avoid having null `web_contents()` when `api::WebContents` is alive, but when app is quitting it is destroyed synchronously and there is a short time when `web_contents()` is null while `api::WebContents` is not destroyed.

zcbenz:
Hmmm after writing this down, I realized that the root cause for this crash, is that there is an edge case causing `content::WebContents` to be destroyed before `api::WebContents`, and we should fix the edge case instead of working around it.

I'll create another PR to cleanup the shutdown routine of WebContents to eliminate this kind of edge cases. But it will be a relatively large refactoring to core shutdown code, and I'm not confident backporting it would not cause other crashes in the stable branches. So I'm in favor of using this PR as a temporary fix to make CI happy, and I'll fix the root cause in the master branch.

zcbenz:
@nornagon I have created https://github.com/electron/electron/pull/27920 to fix the root cause of the crash, by eliminating the cases that `web_contents()` could be null before `api::WebContents` is destroyed.

My plan is:
1. Merge this PR and backport it to older branches. (This will fix the persistent crashReporter CI failure on 12).
2. Revert this change in #27920, which would fix the root cause instead of working around the crash in this PR.
3. Merge #27920 to master without backporting it, since it is a refactoring to shutdown code that may cause new crashes, or expose existing crashes that were hidden by the nondeterministic shutdown order.

Are you OK with it?

nornagon:
Hm, alright. Seems better than nothing probably. And any new crashes caused by this PR are ... probably rare?

nornagon (approved):

PR #27823, opened 2021-02-19
fix: isFullScreen typo

TonyWuu:
#### Description of Change
`TypeError: this.isFullscreen is not a function`
I believe this should fix `isDocumentEdited()` api

#### Checklist 
* [x]  PR description included and stakeholders cc'd

#### Release Notes
Notes: none

### DIFF
--- a/lib/browser/api/base-window.ts
+++ b/lib/browser/api/base-window.ts
@@ -43,7 +43,7 @@ Object.defineProperty(BaseWindow.prototype, 'kiosk', {
 });
 
 Object.defineProperty(BaseWindow.prototype, 'documentEdited', {
-  get: function () { return this.isFullscreen(); },
+  get: function () { return this.isFullScreen(); },
   set: function (edited) { this.setDocumentEdited(edited); }
 });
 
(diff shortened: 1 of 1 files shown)


codebytere (approved):

lib/browser/api/base-window.ts
```diff
@@ -43,7 +43,7 @@ Object.defineProperty(BaseWindow.prototype, 'kiosk', {
 });
 
 Object.defineProperty(BaseWindow.prototype, 'documentEdited', {
-  get: function () { return this.isFullscreen(); },
+  get: function () { return this.isFullScreen(); },
```
jkleinsc, line 46: This never should have been isFullScreen.  isDocumentEdited is the proper function to call
```suggestion
  get: function () { return this.isDocumentEdited(); },
```

jkleinsc (changes requested):
the proper function to call here is `isDocumentEdited` not `isFullScreen`

lib/browser/api/base-window.ts
TonyWuu: This makes way more sense :) I thought for some reason this was intentional

jkleinsc (approved):

PR #27860, opened 2021-02-22
docs: fix header level of some events in app.md

nornagon:
Makes the TOC look weird

image

Notes: none


Jkweeks (changes requested):
Why is it gone I need it to be put intoy bank account

MarshallOfSound (approved):

codebytere (approved):

zcbenz (approved):

Issue #6015, opened 2016-06-12
win.capturePage never calls back after calling `hide()` for a hidden window


brian-mann:
- Electron version: 1.2.2
- Operating system: Linux

This is a problem only in linux. It works in OSX.

``` javascript
win.hide()
win.capturePage(function(img){
 // never called
})
```

zcbenz:
I can't reproduce by simply capturing a page loaded from filesystem,  can you provide a minimal app that can reproduce this problem?

brian-mann:
Okay was able to reproduce in a new example app. There are two things which cause this to happen.

1. The window must currently be hidden
2. You must call `capturePage` asynchronously

```javascript
var app = require('electron').app;
var BrowserWindow = require('electron').BrowserWindow;

var mainWindow = null;

app.on('ready', function() {
  var opts = {
    width: 1280,
    height: 720,
    show: false,
    frame: false,
    devTools: false
  }

  mainWindow = new BrowserWindow(opts);

  // hiding the window is key, if it is shown it will work
  mainWindow.hide()

  mainWindow.loadURL('file://' + __dirname + '/index.html');

  setTimeout(function(){
    mainWindow.capturePage(function(data){
        // this function is never called
        console.log("captured data", data)
    })
  }, 1000) // calling this async is key. if you reduce the setTimeout value it will work

});
```

zcbenz:
I can reproduce with your example, calling `hide()` for a hidden window seems to cause problems.

codebytere:
closing due to abandonment.

jcj52436999:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  joe.jackson.two@gmail.com
    Domain joe-jackson.net has exceeded the max emails per hour (124/100 (124%)) allowed.  Message will be reattempted later

------- This is a copy of the message, including all the headers. ------
Received: from github-smtp2-ext3.iad.github.net ([192.30.252.194]:45330 helo=github-smtp2b-ext-cp1-prd.iad.github.net)
	by rs5.websitehostserver.net with esmtps (TLSv1.2:ECDHE-RSA-AES256-GCM-SHA384:256)
	(Exim 4.89)
	(envelope-from <noreply@github.com>)
	id 1dx1SQ-0018wQ-Np
	for jcjackson@joe-jackson.net; Tue, 26 Sep 2017 20:48:17 -0500
[...]

PR #27883, opened 2021-02-23
fix: capturePage not resolving with hidden windows

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/6015.
Closes https://github.com/electron/electron/issues/27891.

If the renderer is suspended or otherwise occluded, there's no guarantee that `view->CopyFromSurface` will ever be called and thus the promise ever resolved (see [here](https://source.chromium.org/chromium/chromium/src/+/master:content/public/browser/render_widget_host_view.h;l=218-219?q=CopyFromSurface&ss=chromium)). Handle this by checking for the current page visibility and returning an empty image if this is the case.

cc @ckerr @zcbenz @MarshallOfSound @deepak1556 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where `win.capturePage()` never called back after calling `hide()` for a hidden window on some platforms.

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -1371,7 +1371,7 @@ Returns `Boolean` - Whether the window's document has been edited.
 
 Returns `Promise<NativeImage>` - Resolves with a [NativeImage](native-image.md)
 
-Captures a snapshot of the page within `rect`. Omitting `rect` will capture the whole visible page.
+Captures a snapshot of the page within `rect`. Omitting `rect` will capture the whole visible page. If the page is not visible, `rect` may be empty.
 
 #### `win.loadURL(url[, options])`
 
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -2829,6 +2829,18 @@ v8::Local<v8::Promise> WebContents::CapturePage(gin::Arguments* args) {
     return handle;
   }
 
+#if !defined(OS_MAC)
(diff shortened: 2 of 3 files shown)


zcbenz (approved):

deepak1556:
@codebytere https://github.com/electron/electron/pull/27892 looks like a better solution to this issue as well, it has also narrowed down the root issue.

CezaryKulakowski:
@deepak1556 @codebytere I've made some testing and my fix (https://github.com/electron/electron/pull/27892) doesn't cover all cases: it may happen that callback isn't called at all. It can be reproduced by test added to this pr. I've checked that this fix covers both cases: with it both tests (mine and the one added by this pr) pass. I'd say: let's close my pr and merge this one.

deepak1556:
I am good with adding this fix back, but in your PR it is indicated that the responses sometime come back on a different thread is that only the case for hidden pages ?

CezaryKulakowski:
Yes, I've checked that this fix also covers my test case. It seems that without any fix callback is called on bad thread or it's not called at all for hidden pages so this fix should cover both cases.

deepak1556:
Thanks for clarifying, @codebytere wanna revive this PR ?

ztob:
hi guys,

this issue still occurs.
we run electron app with 30-50 concurrent `BrowserWindow` with hidden state and at some point on some (random) pages the capture method will hang forever in the promise.
I did implemented Page.captureScreenshot through devtools protocol but the same thing.

`electron v16.0.4`

Issue #27891, opened 2021-02-24
WebContents.capturePage can freeze browser process


CezaryKulakowski:
### Preflight Checklist

* [x ] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [ x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [ x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 13.0.0-nightly.20210223
* **Operating System:**
  * Windows 10
* **Last Known Working Electron version:**
  * It probably never worked

### Expected Behavior
Call to BrowserWindow.webContents.capturePage does not freeze the application.

### Actual Behavior
Call to BrowserWindow.webContents.capturePage freezes the application.

### To Reproduce
Run following app:
```
const { app, BrowserWindow } = require('electron')

function startTest() {
  let win = new BrowserWindow();

  win.on('hide', () => {
    // We have to wait a moment as 'hide' is called too soon
    // and windows isn't really hidden at this point.
    setTimeout( () => {
      win.webContents.capturePage();
      app.quit();
    }, 100);
  });

  win.webContents.loadURL('about:blank');
  win.hide();
}

app.on('ready', startTest)
```

pushkin-:
I think this can be closed as a dupe of [this](https://github.com/electron/electron/issues/6015) @codebytere assigned that issue to herself yesterday FYI

codebytere:
I think it is - i'm also not sure that we want to solve this recursively as in your PR.

PR #27892, opened 2021-02-24
fix: WebContents.capturePage can freeze browser process

CezaryKulakowski:
When BrowserWindow.webContents.capturePage is called on hidden
page we try to resolve its promise on the worker thread instead of
the main thread of the browser process. It happens because when page
is hidden compositor calls the callback OnCapturePageDone immediately
without posting it to the main thread). When it happens worker thread
dead locks as it tries to acquire lock kept by the main thread (the
lock which guards the promise returned by capturePage).

This change fixes the problem by assuring that promise is always
resolved on the main thread of the browser process.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed possible dead lock on call to WebContents.capturePage

### DIFF
--- a/spec-main/api-browser-window-spec.ts
+++ b/spec-main/api-browser-window-spec.ts
@@ -1353,6 +1353,22 @@ describe('BrowserWindow module', () => {
       // Values can be 0,2,3,4, or 6. We want 6, which is RGB + Alpha
       expect(imgBuffer[25]).to.equal(6);
     });
+
+    it('does not freeze browser process when called on hidden page', async() => {
+      await new Promise<void>((resolve) => {
+        const w = new BrowserWindow({ show: true });
+        w.on('hide', () => {
+          // We have to wait a moment as 'hide' is called too soon
+          // and windows isn't really hidden at this point.
+          setTimeout( async () => {
+            await w.webContents.capturePage();
+            resolve();
+          }, 100);
+        });
+        w.webContents.loadURL('about:blank');
+        w.hide();
+      });
+    });
   });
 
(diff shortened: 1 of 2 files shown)


CezaryKulakowski:
Fixes https://github.com/electron/electron/issues/27891.

deepak1556 (commented):
Nice find, LGTM

Failing tests on mac are related.

CezaryKulakowski:
Test fails on mac as it looks like `webContents.on('hide')` is not called after call to `win.hide()`. I'll look at it during the weekend.

spec-main/api-browser-window-spec.ts
```diff
@@ -1353,6 +1353,22 @@ describe('BrowserWindow module', () => {
       // Values can be 0,2,3,4, or 6. We want 6, which is RGB + Alpha
       expect(imgBuffer[25]).to.equal(6);
     });
+
+    it('does not freeze browser process when called on hidden page', async() => {
```
zcbenz, line 1357: Can you also fix the lint error?

```
/home/builduser/project/src/electron/spec-main/api-browser-window-spec.ts
  1357:75  error  Missing space before function parentheses  space-before-function-paren
  1363:22  error  There should be no space after this paren  space-in-parens
```

CezaryKulakowski: @zcbenz we should go with this fix: https://github.com/electron/electron/pull/27883 (see my comment in that pull request).

zcbenz:
Closing in favor of https://github.com/electron/electron/pull/27883.

Issue #27805, opened 2021-02-18
File downloads using electron.net requests piped to writable streams silently stops


andreasdj:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 12.0.0-beta.26
* **Operating System:**
  * Windows 10 (2004)
* **Last Known Working Electron version:**
  * 11.2.3

### Expected Behavior
When piping a file response stream initiated by an `electron.net.request` to a writable file stream, I expect it to successfully download and save the file using a stable internet connection.

### Actual Behavior
The download stops randomly and most often never completes.

### To Reproduce
The following gist includes an example of how the issue can be reproduced:
https://gist.github.com/c940aa29e51685fe0a7fe5b68ca247ad

### Additional Information
I've tried to find the root cause for this intermittent issue but without success so far. It seems to be a combination of the buffering inside the node Stream `pipe` method and the `electron.net.request`. Adding a couple of tests made that might give some more insight:

- By skipping buffering and using something like `response.on('data', (chunk: Buffer) => fileStream.write(chunk))` directly I cannot reproduce the issue.
- By increasing the highWaterMark (>512kb) for the writable target stream the issue happens more rarely.
- By adding an extra delay after the app ready event the issue happens more rarely.
- By using an external library (`got`) not using electron.net for making the request I'm able to pipe the response to the target stream without any issues.
- By using another session than the `defaultSession` the issue happens more rarely

khwaaj:
I debugged this today and found a race condition related to the resume callback in [IncomingMessage](https://github.com/electron/electron/blob/dfed04b71b91b92a6018d421d9d3319f6153e83e/lib/browser/api/net.ts#L38). Apparently the `resume` callback can trigger `_storeInternalData` before the stored value is [overwritten with `null`](https://github.com/electron/electron/blob/dfed04b71b91b92a6018d421d9d3319f6153e83e/lib/browser/api/net.ts#L125), causing the new value to be overwritten instead, and thus stalling the download.

This is the pattch I applied to our fork to fix the issue:

```diff
[...]

PR #27898, opened 2021-02-24
fix: prevent race-condition in electron.net

khwaaj:
#### Description of Change

Fixes a race-condition in electron.net reported in issue https://github.com/electron/electron/issues/27805.

The race exists in the implementation of [IncomingMessage](https://github.com/electron/electron/blob/master/lib/browser/api/net.ts#L38) in the [_pushInternalData](https://github.com/electron/electron/blob/master/lib/browser/api/net.ts#L122) function. Under some circumstances, e.g. when a pipe throttles the response data, the callback stored in `this._resume` can be updated by [_storeInternalData](https://github.com/electron/electron/blob/master/lib/browser/api/net.ts#L109), triggered by the callback itself, before it is [reset to `null`](https://github.com/electron/electron/blob/master/lib/browser/api/net.ts#L125). This results in it being overwritten with `null`, effectively stalling the download indefinitely waiting for permission to continue.

The suggested fix simply uses a local proxy variable to cache the callback value and reset it before actually executing the callback, preventing the race.

#### Checklist

#### Release Notes

Notes: Fixed race condition in electron.net implementation.

### DIFF
--- a/lib/browser/api/net.ts
+++ b/lib/browser/api/net.ts
@@ -119,10 +119,13 @@ class IncomingMessage extends Readable {
       this._shouldPush = this.push(chunk);
     }
     if (this._shouldPush && this._resume) {
-      this._resume();
       // Reset the callback, so that a new one is used for each
-      // batch of throttled data
+      // batch of throttled data. Do this before calling resume to avoid a
+      // potential race-condition
+      const resume = this._resume;
       this._resume = null;
+
+      resume();
     }
   }
 


VerteDinde (approved):
Great catch! This fix looks good to me - the test we introduced to ensure the throttled download continued seems to be passing. I'm not able to reproduce the race condition itself.

codebytere (approved):

nornagon (approved):
nice catch.

nornagon:
It looks like one of the net module tests is failing on the ASan build. It might not be ASan-specific, but the ASan build is a bit slower, so it could be exposing another timing issue. Would you mind taking a look?

khwaaj:
@nornagon, the same test didn't fail the first build, so it could very well be an artifact of exposing another intermittent race condition.

I have no quick way of testing it currently, but it appears to be triggered after an abortion test has concluded, in `afterEach`, and yields a mojo error [from the URL loader](https://github.com/electron/electron/blob/master/shell/browser/api/electron_api_url_loader.cc#L198). I get the feeling something that should not be is triggered after the abort, but it warrants further investigation. I'll see if I can make some time to take a closer look at it.

nornagon:
I'm OK merging this while investigation continues on the other possible race.

khwaaj:
@nornagon, thanks for merging. Is there anything I need to do to get this back-ported to `12-x-y`?

I spent some time this weekend trying to reproduce the asan-issue in my setup, but despite introducing various delays, trying to provoke it, I've been unsuccessful at doing so. I don't have more time to look into this right now so I'm unsure how to best proceed here.

VerteDinde:
@khwaaj I went ahead and backported your PR to 12 so it could get into the stable release tomorrow (sorry for jumping the gun if you wanted to backport it yourself, I just wanted to make sure it made the cut). Typically, our Trop bot will backport it for you, but if you want to backport in the future, all you need to do is pull the relevant branch (i.e., `12-x-y`) and either cherry-pick or add your code in.

If there's any questions you have, let me know, and thanks again for the fix! 🙂

PR #27920, opened 2021-02-26
refactor: cleanup how WebContents is destroyed

zcbenz:
#### Description of Change

This PR cleans up the shutdown routine of WebContents, so:
1. The `content::WebContents` is always destroyed synchronously in the destructor of `api::WebContents`.
2. The only way to destroy `api::WebContents` is to call `api::WebContents::Destroy`, both in C++ and JavaScript.

With this change, it becomes much clearer how WebContents is destroyed, and the WebContents will have a deterministic destruction order.

Note that this PR might introduce new crashes, we should not backport it to old branches.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/electron_api_base_window.cc
+++ b/shell/browser/api/electron_api_base_window.cc
@@ -114,8 +114,7 @@ BaseWindow::BaseWindow(gin_helper::Arguments* args,
 }
 
 BaseWindow::~BaseWindow() {
-  if (!window_->IsClosed())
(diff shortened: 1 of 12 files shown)


zcbenz:
This PR should not be merged before https://github.com/electron/electron/pull/27815.

deepak1556 (approved):
LGTM

PR #27928, opened 2021-02-27
fix: warning when worldSafeExecuteJavaScript is disabled

miniak:
#### Description of Change
Command-line switches are no longer used to pass `webPreferences` to the renderer, replace with `getWebPreference`.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Fixed warning when `worldSafeExecuteJavaScript` is disabled.

### DIFF
--- a/lib/renderer/api/web-frame.ts
+++ b/lib/renderer/api/web-frame.ts
@@ -48,8 +48,10 @@ class WebFrame extends EventEmitter {
   }
 }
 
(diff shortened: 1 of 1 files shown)


miniak:
@codebytere it needs to be backported to other versions as well, however only E12 uses `getWebPreference`, E10 and E11 need a manual backport, those are still using the commandline arguments

codebytere:
@miniak i added targets for bookkeeping, so for 11 you can just push the correct change to trop's PR

PR #27938, opened 2021-03-01
fix: prevent race-condition in electron.net

VerteDinde:
Backport of #27898. (cc @khwaaj)

See that PR for details.

Notes: Fixed race condition in electron.net implementation.


jkleinsc (approved):

jkleinsc:

PR #27940, opened 2021-03-01
fix: values return from the ctx bridge with dynamic property support should themselves support dynamic properties

VerteDinde:
Backport of #27899.

See that PR for details.

Notes: Fixed issue where window.open() would not return an object with a location.href setter when contextIsolation is enabled and nativeWindowOpen is disabled.


jkleinsc:
Merging as CI failures are flakes unrelated to PR change.

PR #27953, opened 2021-03-02
chore: Add a condition to crashReporter deprecate log

xland:
When developer set  submitURL to '' crash reports will be saved  at `...\AppData\Roaming\...\Crashpad\reports`, will not be uploaded to the server.
So  at this time `deprecate.log('Sending uncompressed crash reports....')`  is  unnecessary.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none.

### DIFF
--- a/lib/browser/api/crash-reporter.ts
+++ b/lib/browser/api/crash-reporter.ts
@@ -18,7 +18,7 @@ class CrashReporter {
 
     if (submitURL == null) throw new Error('submitURL is a required option to crashReporter.start');
 
-    if (!compress && submitURL != '') {
+    if (!compress && uploadToServer) {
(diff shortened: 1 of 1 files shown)


codebytere (approved):
Thanks!

codebytere:
@xland can you please fix linting?

lib/browser/api/crash-reporter.ts
```diff
@@ -18,7 +18,7 @@ class CrashReporter {
 
     if (submitURL == null) throw new Error('submitURL is a required option to crashReporter.start');
 
-    if (!compress) {
+    if (!compress && submitURL != '') {
```
nornagon, line 21:
```suggestion
    if (!compress && uploadToServer) {
```

I think it's better to check `uploadToServer` than `submitURL`

zcbenz (approved):

PR #27958, opened 2021-03-02
docs: suspend and resume are available on Linux

### LINKED ISSUE #27761: PowerMonitor incorrectly omits Linux for suspend, resume events
https://www.electronjs.org/docs/api/power-monitor#powermonitor
Previously used to be 'suspend, resume' events also for linux. 
Why you remove it? 
Are there alternative solutions?
I need to after sleep mode "resume" event returns sleep time.
And now if update electron, it will not work.

ssiatras:
#### Description of Change
Closes #27761. 

Updates accordingly PowerMonitor documentation for `suspend` and `resume`, which were added on Linux by #25149.

#### Checklist

#### Release Notes

Notes: None

### DIFF
--- a/docs/api/power-monitor.md
+++ b/docs/api/power-monitor.md
@@ -8,11 +8,11 @@ Process: [Main](../glossary.md#main-process)
 
 The `powerMonitor` module emits the following events:
 
-### Event: 'suspend' _macOS_ _Windows_
+### Event: 'suspend' _Linux_ _macOS_ _Windows_
 
 Emitted when the system is suspending.
 
-### Event: 'resume' _macOS_ _Windows_
+### Event: 'resume' _Linux_ _macOS_ _Windows_
 
 Emitted when system is resuming.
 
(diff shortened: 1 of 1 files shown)


codebytere (dismissed):
-

docs/api/power-monitor.md
```diff
@@ -8,11 +8,11 @@ Process: [Main](../glossary.md#main-process)
 
 The `powerMonitor` module emits the following events:
 
-### Event: 'suspend' _macOS_ _Windows_
+### Event: 'suspend' _Linux_ _macOS_ _Windows_
```
codebytere, line 11: We only add platforms when there's partial support - this can now just be `### Event: 'suspend'`

docs/api/power-monitor.md
```diff
@@ -8,11 +8,11 @@ Process: [Main](../glossary.md#main-process)
...
 
 Emitted when the system is suspending.
 
-### Event: 'resume' _macOS_ _Windows_
+### Event: 'resume' _Linux_ _macOS_ _Windows_
```
codebytere, line 15: Same as above

codebytere (changes requested):

codebytere (approved):

ckerr (approved):

zcbenz (approved):

PR #28007, opened 2021-03-04
feat: add process.contextId used by @electron/remote

miniak:
#### Description of Change
The built-in remote module is usable in sandboxed renderers. However the standalone https://github.com/electron/remote version is not due to `process._linkedBinding` not being available to the preload script. Expose public API for that.

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: Added `process.contextId` used by `@electron/remote`.

### DIFF
--- a/docs/api/process.md
+++ b/docs/api/process.md
@@ -35,6 +35,7 @@ In sandboxed renderers the `process` object contains only a subset of the APIs:
 * `versions`
 * `mas`
 * `windowsStore`
+* `contextId`
 
 ## Events
 
@@ -132,6 +133,11 @@ A `String` representing Electron's version string.
 
 A `Boolean`. If the app is running as a Windows Store app (appx), this property is `true`,
 for otherwise it is `undefined`.
+
+### `process.contextId` _Readonly_
+
+A `String` (optional) representing a globally unique ID of the current JavaScript context.
+This property is only available in the renderer process.
 
 ## Methods
 
--- a/docs/api/process.md
+++ b/docs/api/process.md
@@ -137,6 +137,8 @@ for otherwise it is `undefined`.
 ### `process.contextId` _Readonly_
 
 A `String` (optional) representing a globally unique ID of the current JavaScript context.
+Each frame has its own JavaScript context. When contextIsolation is enabled, the isolated
+world also has a separate JavaScript context.
 This property is only available in the renderer process.
 
 ## Methods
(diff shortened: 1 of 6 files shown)


docs/api/process.md
```diff
@@ -133,6 +134,11 @@ A `String` representing Electron's version string.
 A `Boolean`. If the app is running as a Windows Store app (appx), this property is `true`,
 for otherwise it is `undefined`.
 
+### `process.versions.contextId` _Readonly_
```
nornagon, line 137:
```suggestion
### `process.contextId` _Readonly_
```

docs/api/process.md
```diff
@@ -133,6 +134,11 @@ A `String` representing Electron's version string.
...
+
+A `String` (optional) representing unique ID of the current JavaScript context.
```
nornagon, line 139: it's only unique within the renderer i believe, we should note what scope it's unique in.

lib/sandboxed_renderer/init.ts
```diff
@@ -89,6 +89,9 @@ Object.defineProperty(preloadProcess, 'noDeprecation', {
   }
 });
 
+const contextId = v8Util.getHiddenValue<string>(global, 'contextId');
+Object.defineProperty(process, 'contextId', { get: () => contextId });
```
nornagon, line 93: does this need to be defined on `preloadProcess`?

nornagon: ah nevermind, i see this is handled below.

miniak: yes, this is exactly why whole change is necessary as the binding is only available on the internal process object, not to the preload script.

docs/api/process.md
miniak: isn't it globally unique?
https://github.com/electron/electron/blob/f114dcfd6e44a1b67cb83af40097fd49ea1f9239/shell/renderer/renderer_client_base.cc#L130-L138

nornagon: ah, you're right.

docs/api/process.md
```diff
@@ -133,6 +134,11 @@ A `String` representing Electron's version string.
 A `Boolean`. If the app is running as a Windows Store app (appx), this property is `true`,
 for otherwise it is `undefined`.
 
+### `process.contextId` _Readonly_
+
+A `String` (optional) representing a globally unique ID of the current JavaScript context.
```
nornagon, line 139:
```suggestion
A `String` (optional) representing a globally unique ID of the current JavaScript context.
Each frame has its own JavaScript context. When contextIsolation is enabled, the isolated
world also has a separate JavaScript context.
```

nornagon (approved):

zcbenz (approved):

codebytere (approved):

Issue #18139, opened 2019-05-03
Uncaught ReferenceError: process is not defined


msudgh:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** 5.0.0
* **Operating System:** Arch Linux

### Expected Behavior

I'd expect to access `process` object in my renderer process

### Actual Behavior

Chrome dev tools throw `ReferenceError` while accessing to `process` object.
```js
Uncaught ReferenceError: process is not defined
```

codebytere:
@msudgh `nodeIntegration` is turned off by default in `5`, so you'll need to enable it via:
```
const { BrowserWindow } = require('electron')
let win = new BrowserWindow({
  webPreferences: {
    nodeIntegration: false
  }
})
```
which we caution against, or enable it via preload script.

msudgh:
@codebytere Thanks. it'd be better to update docs. this functionality existed in previous versions.

codebytere:
@msudgh this information was presented both in release notes and in breaking change documentation leading up to and including the stable release :)

Raghav9o9:
Hello I am also getting this error and still it is not fixed for me pls help me.

Raghav9o9:
Thanks it worked for me, you don't that how happy now I am thank you very very much...

phamkimtho:
Thank you very much. Now I can solved my problem!

const { BrowserWindow } = require('electron')
let win = new BrowserWindow({
  webPreferences: {
    nodeIntegration: true
  }
})

swapnilvaidankar:
> @msudgh `nodeIntegration` is turned off by default in `5`, so you'll need to enable it via:
> 
> ```js
> const { BrowserWindow } = require('electron')
> let win = new BrowserWindow({
>   webPreferences: {
>     nodeIntegration: true
>   }
> })
> ```
> 
> which we caution against, or add it via preload script.

Where to add this above code?

schmorrison:
> > @msudgh `nodeIntegration` is turned off by default in `5`, so you'll need to enable it via:
> > ```js
> > const { BrowserWindow } = require('electron')
[...]

PR #28008, opened 2021-03-04
Update quick-start.md

gclotault:
#### Description of Change
Added contexIsolation: false to BrowserWindow options (see https://github.com/electron/electron/issues/18139#issuecomment-790087307)

### DIFF
--- a/docs/tutorial/quick-start.md
+++ b/docs/tutorial/quick-start.md
@@ -61,7 +61,8 @@ function createWindow () {
     width: 800,
     height: 600,
     webPreferences: {
-      nodeIntegration: true
+      nodeIntegration: true,
+      contextIsolation: false
     }
   })
 


zcbenz (commented):
Can you also update the line numbers in the "What is going on above?" section?

MarshallOfSound (changes requested):
We should update the example to function with the secure defaults, our docs should never explicitly or implicitly recommend insecure options

ckerr (changes requested):
:+1: on @MarshallOfSound's comment; this is why electron-quick-start [added a preload script](https://github.com/electron/electron-quick-start/pull/279)

ckerr:
However, I think @gclotault is right that the instructions in `docs/tutorial/quick-start.md` won't work as-is due to the contextIsolation change. We should update this file to use a preload script and to explain why that's preferred. Also we might want to include links to [Electron Fiddle](https://github.com/electron/fiddle) and [electron-quick-start](https://github.com/electron-quick-start).

erickzhao:
I personally think this is okay for a band-aid fix before rewriting the tutorials to use sensible defaults at some later point.

Right now, our tutorial just doesn't work, and that's a very confusing experience for newbies.

jkleinsc:
Thanks for the PR @gclotault!  While this solution technically works, the preferred way to handle this is with a preload script.  Since #28223 makes that change, I am going to close this PR in favor of #28223.

Issue #4985, opened 2016-04-01
Blinking / Flashing while loading


nicozimmermann94:
* Electron version: 0.37.3
* Operating system: Windows 10 x64

Hello, is there a way do turn off the white flash while loading a site?
![blink](https://cloud.githubusercontent.com/assets/9193934/14204320/c749f8b8-f803-11e5-9f3e-16a33da3993d.gif)

nicozimmermann94:
Thank you very much.

zcbenz:
@nicozimmermann94 I think I have probably misunderstood your issue, there is actually no way to remove the white flash when you do page navigations, because browser works by downloading and rendering page in the same time, so when navigation happens you will the progress of previous page getting erased and the new page getting rendered, which is the white flash you saw.

The only workaround is to avoid page navigations in your app completely, with techniques like ajax, otherwise your app will just behave like normal browsers.

nicozimmermann94:
@zcbenz But when I visit the website with the "normal" chrome/chromium webbrowser, I do not get those white flashes. It seems like chrome waits to render it before it refreshes the view.

diaryzhang:
@zcbenz I build a demo based on electron and nwjs separately,  I found electron still has the issue of visual flash when navigating, but nwjs does not has this issue. Does electron have the plan or date to solve this issue, or leave this issue to chromium? Nwjs is using libchromiumcontent, and electron is using Chromium, is that the reason?

zcbenz:
@diaryzhang Electron forces the renderer process to restart for every navigation, while Chrome and nwjs only restarts the renderer process for cross-site navigations, that should be why you can see the flashing.

nicozimmermann94:
@zcbenz is there a way to force electron to behave like chrome?

diaryzhang:
@zcbenz can we add a option to loadUrl API to control whether force to restart the renderer process or not. I believe many devs have the visual flash issue.

PR #28022, opened 2021-03-05
fix: delay setting background color for transparent windows

VerteDinde:
#### Description of Change

Addresses an issue with loading transparent renderviews introduced by #27593. While #27593 fixed a prolonged white flash while loading a new window has a set background color, it unfortunately broke opacity for certain transparent windows. 

Now, transparent child windows created by window.open without an explicit background color and no external URL load with the default white background. This PR delays setting the background color for transparent windows in LoadURL, similar to how they behaved prior to the fix.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where transparency was not being respected for child windows created by native window.open path.

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -2006,7 +2006,7 @@ void WebContents::LoadURL(const GURL& url,
   // For transparent windows, set the background color of RenderWidgetHostView.
   // We need to set this color after LoadURL, because the RenderViewHost is only
   // created after loading a page; a default white color is rendered otherwise.
-  auto* const view = weak_this->web_contents()->GetRenderWidgetHostView();
+  auto* const view = this->web_contents()->GetRenderWidgetHostView();
   if (view) {
     auto* web_preferences = WebContentsPreferences::From(web_contents());
     std::string color_name;
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -1357,7 +1357,8 @@ void WebContents::RenderViewCreated(content::RenderViewHost* render_view_host) {
   if (!background_throttling_)
     render_view_host->SetSchedulerThrottling(false);
 
-  // Set the background color of RenderWidgetHostView.
+  // If the RenderWidgetHostView contains a background color (is not
+  // transparent), set the background color earlier.
   auto* const view = web_contents()->GetRenderWidgetHostView();
   auto* web_preferences = WebContentsPreferences::From(web_contents());
   if (view && web_preferences) {
@@ -1365,8 +1366,6 @@ void WebContents::RenderViewCreated(content::RenderViewHost* render_view_host) {
     if (web_preferences->GetPreference(options::kBackgroundColor,
                                        &color_name)) {
       view->SetBackgroundColor(ParseHexColor(color_name));
-    } else {
-      view->SetBackgroundColor(SK_ColorTRANSPARENT);
     }
   }
 }
@@ -2003,6 +2002,21 @@ void WebContents::LoadURL(const GURL& url,
 
   // Required to make beforeunload handler work.
   NotifyUserActivation();
+
+  // For transparent windows, set the background color of RenderWidgetHostView.
+  // We need to set this color after LoadURL, because the RenderViewHost is only
+  // created after loading a page; a default white color is rendered otherwise.
+  auto* const view = weak_this->web_contents()->GetRenderWidgetHostView();
+  if (view) {
+    auto* web_preferences = WebContentsPreferences::From(web_contents());
+    std::string color_name;
+    if (web_preferences->GetPreference(options::kBackgroundColor,
+                                       &color_name)) {
+      view->SetBackgroundColor(ParseHexColor(color_name));
+    } else {
+      view->SetBackgroundColor(SK_ColorTRANSPARENT);
+    }
+  }
 }
 
 void WebContents::DownloadURL(const GURL& url) {
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -1055,6 +1055,16 @@ void WebContents::AddNewContents(
   v8::HandleScope handle_scope(isolate);
   auto api_web_contents =
       CreateAndTake(isolate, std::move(new_contents), Type::kBrowserWindow);
+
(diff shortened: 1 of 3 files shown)


clavin (approved):

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2003,6 +2002,21 @@ void WebContents::LoadURL(const GURL& url,
 
   // Required to make beforeunload handler work.
   NotifyUserActivation();
+
+  // For transparent windows, set the background color of RenderWidgetHostView.
+  // We need to set this color after LoadURL, because the RenderViewHost is only
+  // created after loading a page; a default white color is rendered otherwise.
+  auto* const view = weak_this->web_contents()->GetRenderWidgetHostView();
```
nornagon, line 2009: nit: `weak_this`'s purpose is to detect whether we got destroyed when `LoadURLWithParams` triggered some javascript. After the check for `!weak_this`, we know we didn't get deleted so we can use `this` like normal. It's functionally the same, but is a little bit conceptually neater.
```suggestion
  auto* const view = this->web_contents()->GetRenderWidgetHostView();
```

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2003,6 +2002,21 @@ void WebContents::LoadURL(const GURL& url,
 
   // Required to make beforeunload handler work.
   NotifyUserActivation();
+
+  // For transparent windows, set the background color of RenderWidgetHostView.
+  // We need to set this color after LoadURL, because the RenderViewHost is only
+  // created after loading a page; a default white color is rendered otherwise.
```
nornagon, line 2008: Hm, I'm not sure `LoadURL` is quite the right place to do this. If we need to do this for every render view in the window, perhaps `RenderViewHostChanged` is the right place? That would additionally account for cross-process navigations.

I'm not totally sure that this will exercise the code path I'm imagining as I haven't pulled down this PR to test locally, but here's a fiddle that I think might tickle some bad behavior with this PR applied: https://gist.github.com/5aef6fdac4379c36df0bc7b9aa8aecf7

1. It opens a child window pointing at a data: url (when you click the button)
2. the child window is transparent
3. clicking the "x-process nav" link will result in a change of RenderWidgetHostView, and on current 12.0.0 results in a flash of white.

If the above fiddle doesn't produce a flash of white with this PR applied, then we're probably safe doing this here in LoadURL, but if it **does** flash white, we should probably move this code to RenderViewHostChanged (or some other similar method that also gets called when a process swap happens).

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1357,16 +1357,15 @@ void WebContents::RenderViewCreated(content::RenderViewHost* render_view_host) {
...
+  // transparent), set the background color earlier.
   auto* const view = web_contents()->GetRenderWidgetHostView();
   auto* web_preferences = WebContentsPreferences::From(web_contents());
   if (view && web_preferences) {
     std::string color_name;
     if (web_preferences->GetPreference(options::kBackgroundColor,
                                        &color_name)) {
       view->SetBackgroundColor(ParseHexColor(color_name));
-    } else {
-      view->SetBackgroundColor(SK_ColorTRANSPARENT);
```
nornagon, line 1369: I'm not sure I understand why we don't want to set the background color to transparent here any more..?

shell/browser/api/electron_api_web_contents.cc
nornagon: also: just saw that in 13.x this was moved to RenderFrameCreated, which is probably not exactly correct as it will be called when child frames are created as well as the main frame, and we don't need to update the background color when a child frame is created. it _would_ get called for x-process navigations though, which explains why the symptom is fixed in 13.

shell/browser/api/electron_api_web_contents.cc
VerteDinde: That's fair; I removed it just because it doesn't seem to be respecting the transparent option at all, and is setting the background color to white regardless, so I thought keeping it would be confusing for future debugging. Happy to leave it in though!

shell/browser/api/electron_api_web_contents.cc
VerteDinde: Oh awesome, thanks so much for making a Fiddle too! 🙂 Will give this a try and report back - if it works, I'll go ahead and move the code over to RenderViewHostChanged 🙇‍♀️

shell/browser/api/electron_api_web_contents.cc
VerteDinde: Thanks, and thanks for explaining! Committed the change

shell/browser/api/electron_api_web_contents.cc
MarshallOfSound: I think this is required for the `transparent: true` option to work correctly

shell/browser/api/electron_api_web_contents.cc
VerteDinde: Update: it looks as though moving the setBackgroundColor logic to either `RenderViewHostChanged` or `RenderFrameHostChanged` doesn't solve the problem. Both are triggered in the code path (logging verified this), but neither seem to set the background color for transparency correctly. After double checking 13-x-y and master, it looks as though the issue is still present in both of those branches as well.

The bug seems to happen in a when a window.open method is triggered from an existing BrowserWindow, and creates a new, transparent BrowserWindow overlaying the previous one. When trying to load an external site, like Github or CNN, there is a brief white flash present regardless of #27593's fix. (Reading previous conversations, it [may not be possible to completely get rid of that flashing](https://github.com/electron/electron/issues/4985#issuecomment-204941449) on the webContents side.)

I'm a little puzzled on why `LoadURL` specifically is correctly setting the background color, but other methods (which are also firing!) are not. `LoadURL` _after_ `RenderViewCreated`, so this may be a race condition where our RenderViewHost hasn't been created yet? Both `RenderViewHostChanged` and `RenderFrameHostChanged` do fire after `LoadURL`, but don't seem to have an affect on the existing webContents 🤔

MarshallOfSound: @VerteDinde Do you have a consistent fiddle repro for this?

deepak1556:
@VerteDinde Should this be closed in favor of https://github.com/electron/electron/pull/28054 ?

VerteDinde:
@deepak1556 We can, sorry! Thanks 🙂

PR #28038, opened 2021-03-08
test: give some time for spellchecker to toggle

zcbenz:
#### Description of Change

Toggling spellchecker works asynchronously, the test should wait for a while before checking the result.

#### Release Notes

Notes: none


jkleinsc (changes requested):
Instead of using an arbitrary delay could we instead add something to periodically check if the value has been set.  For example, I did the following to wait for a badge count to be set:
```
  async function waitForBadgeCount (value: number) {
    let badgeCount = app.getBadgeCount();
    while (badgeCount !== value) {
      await new Promise(resolve => setTimeout(resolve, 10));
      badgeCount = app.getBadgeCount();
    }
    return badgeCount;
  }
```
(From https://github.com/electron/electron/blob/master/spec-main/chromium-spec.ts#L1591)

zcbenz:
That looks better 👍 , I'll try to refactor this test.

PR #28054, opened 2021-03-08
fix: ensure child window transparency works

MarshallOfSound:
Windows opened via window.open and intecepted via setWindowOpenHandler or the `new-window` event should (a) have the correct background color and (b) that background color should be transparent if specified.

The changes in api_web_contents fix (a) and the changes in web_contents_preferences fix (b).

Notes: Child windows with specified background colors or transpency now work as intended


VerteDinde (approved):
Looks awesome! 🙇‍♀️

patches/chromium/moves_background_color_setter_of_webview_to_blinks_webprefs_logic.patch
```diff
@@ -0,0 +1,43 @@
...
++  if (prefs.guest_instance_id) {  // webview.
++    web_view->SetBaseBackgroundColor(SK_ColorTRANSPARENT);
++  } else {  // normal window.
++    std::string color_str = prefs.background_color;
++    Color blink_color;
++    bool success = blink_color.SetFromString(WebString::FromASCII(color_str));
++    SkColor color =
++        !success || color_str.empty() ? SK_ColorTRANSPARENT : static_cast<SkColor>(blink_color);
++    web_view->SetBaseBackgroundColor(color);
++  }
```
nornagon, line 40: since both arms of this `if` result in the same function call, let's refactor to:

```
SkColor color = SK_ColorTRANSPARENT;
if (!prefs.guest_instance_id) {
  Color blink_color;
  if (blink_color.SetFromString(WebString::FromASCII(prefs.background_color)))
    color = blink_color;
}
web_view->SetBaseBackgroundColor(color);
```
I believe `SetFromString` already returns false in the empty-string case, so this simplifies the logic somewhat also.

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1057,6 +1057,16 @@ void WebContents::AddNewContents(
...
   auto api_web_contents =
       CreateAndTake(isolate, std::move(new_contents), Type::kBrowserWindow);
+
+  // We call RenderFrameCreated here as at this point the empty "about:blank"
+  // render frame has already been created.  If the window never navigates again
+  // RenderFrameCreated won't be called and certain prefs like
+  // "kBackgroundColor" will not be applied.
+  auto* frame = api_web_contents->MainFrame();
+  if (frame) {
+    api_web_contents->RenderFrameCreated(frame);
```
nornagon, line 1067: It's extremely confusing to have WebContentsObserver methods called from non-WebContents places. Extract the common logic and call it from both here and the observer method.

nornagon (changes requested):

zcbenz:
Should this also target 13-x-y?

nornagon (approved):

filiptrplan:
Any update on 11-x-y? The problem is quite annoying but I don't have the expertise to backport it myself.

nornagon:
Removing needs-manual-bp/11-x-y since 11.x is security fixes only now.

PR #28105, opened 2021-03-10
feat: allow omitting submitURL when uploadToServer is false

nornagon:
#### Description of Change
If there's no uploading, no submit URL is needed.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: The `submitURL` option for `crashReporter.start` is no longer a required argument when `uploadToServer` is false.


deepak1556 (approved):

zcbenz (approved):

zcbenz:
The test also needs to be updated.

PR #28118, opened 2021-03-11
test: reliably wait for spellchecker to load

zcbenz:
#### Description of Change

Make the spellchecker tests more reliable by using a busy loop to check spellchecker's state instead of waiting for fixed time.

#### Release Notes

Notes: none

### DIFF
--- a/shell/common/api/electron_api_v8_util.cc
+++ b/shell/common/api/electron_api_v8_util.cc
@@ -138,6 +138,10 @@ void TriggerFatalErrorForTesting(v8::Isolate* isolate) {
   v8::ExtensionConfiguration config(1, bDeps);
   v8::Context::New(isolate, &config);
 }
+
+void RunUntilIdle() {
+  base::RunLoop().RunUntilIdle();
+}
 #endif
 
 void Initialize(v8::Local<v8::Object> exports,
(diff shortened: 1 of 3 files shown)


nornagon (changes requested):
can we wait for a specific event instead of adding a new primitive?

zcbenz:
> can we wait for a specific event instead of adding a new primitive?

I had tried to find an event to wait for, but I found out that the spellchecker has actually already been loaded when page is loaded, it is blink that not ready to take right clicks.

For example with the bellow app on Linux, you can see that right clicking the textarea immediately after page is loaded will select no text.

I dug a little why text selecting is not working immediately, but then I determined that it was too much work to just fix a flaky test.

```js
const {app, BrowserWindow} = require('electron');

app.on('ready', async () => {
  const mainWindow = new BrowserWindow({ width: 400, height: 400, webPreferences: { nodeIntegration: true, contextIsolation: false } });
  await mainWindow.loadFile('./spec-main/fixtures/chromium/spellchecker.html')
  await mainWindow.webContents.executeJavaScript('document.body.querySelector("textarea").value = "Beautiful asd asd"');
  await mainWindow.webContents.executeJavaScript('document.body.querySelector("textarea").focus()');

  mainWindow.webContents.on('context-menu', (event, params) => { console.log(params); });
  mainWindow.webContents.sendInputEvent({ type: 'mouseDown', button: 'right', x: 43, y: 42 });
});
```

nornagon (approved):

PR #28136, opened 2021-03-11
docs: remove "Additional information" section from issue template

nornagon:
The template includes a comment section at the end anyway:

image

Notes: none


nornagon:
Ugh, it looks like the UI has changed and now the comment at the bottom is gone. Let's add this back.

Issue #15494, opened 2018-10-31
Simple fullscreen: is not properly sizing the window when display metrics change


bpasero:
* Output of `node_modules/.bin/electron --version`: 2.0.12
* Operating System (Platform and Version): macOS 10.14

**Expected Behavior**
When having a window in simple fullscreen and you change display metrics or change from 1 monitor to 2 the window should size according to the new dimensions.

**Actual behavior**
The window size does not change when display metrics change.

**To Reproduce**
* Run Fiddle: https://gist.github.com/bpasero/e4013ea83033a8c231f58f94c0df57c3
* Change display metrics or change monitor count

=> the window is wrongly sized now and needs a transition from normal back to fullscreen to recover.

sofianguy:
@bpasero is this bug still present in version 4.x or later?

bpasero:
@sofianguy reproduces with Electron 6.0.1

electron-triage:
The Electron version reported on this issue is no longer supported. See our [supported versions documentation](https://www.electronjs.org/docs/tutorial/support#supported-versions).

If this is still reproducible on a supported version, please open a new issue with any other new information that a maintainer should know.

Thank you for taking the time to report this issue and helping to make Electron better! Your help is appreciated.

bpasero:
@electron-triage @nornagon @codebytere @sofianguy @samuelmaddock this reproduces in Electron 11 still. Our workaround in VSCode is still to do this:

https://github.com/microsoft/vscode/blob/bd272ea6b7485f62bfbe7a1833e82fc9f8f84214/src/vs/platform/windows/electron-main/window.ts#L472-L484

bpasero:
This still reproduces for me with the provided Fiddle in Electron 11 and 12.

codebytere:
@bpasero do you have a screencap? I just tried locally and could not

bpasero:
@codebytere 

![recording (2)](https://user-images.githubusercontent.com/900690/116245055-d4967680-a768-11eb-99ec-27de2997d1a5.gif)

I think the change has just not been backported to Electron 12.

PR #28150, opened 2021-03-12
fix: recalibrate simpleFullscreen when display metrics change

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/15494.

Overrides `OnDisplayMetricsChanged` in `NativeWindowMac` so that when display metrics change and the `BrowserWindow` is in `simpleFullscreen` mode, the frame is recalibrated for the new bounds.

Tested with https://gist.github.com/bpasero/e4013ea83033a8c231f58f94c0df57c3.

cc @bpasero 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where windows in `simpleFullscreen` mode were not properly resizing when display metrics changed.

### DIFF
--- a/shell/browser/native_window.h
+++ b/shell/browser/native_window.h
@@ -147,6 +147,7 @@ class NativeWindow : public base::SupportsUserData,
   virtual bool IsExcludedFromShownWindowsMenu() = 0;
   virtual void SetSimpleFullScreen(bool simple_fullscreen) = 0;
   virtual bool IsSimpleFullScreen() = 0;
+  virtual void UpdateFrame() = 0;
   virtual void SetKiosk(bool kiosk) = 0;
   virtual bool IsKiosk() = 0;
   virtual bool IsTabletMode() const;
--- a/shell/browser/native_window.h
+++ b/shell/browser/native_window.h
@@ -147,7 +147,6 @@ class NativeWindow : public base::SupportsUserData,
   virtual bool IsExcludedFromShownWindowsMenu() = 0;
   virtual void SetSimpleFullScreen(bool simple_fullscreen) = 0;
   virtual bool IsSimpleFullScreen() = 0;
-  virtual void UpdateFrame() = 0;
   virtual void SetKiosk(bool kiosk) = 0;
   virtual bool IsKiosk() = 0;
   virtual bool IsTabletMode() const;
@@ -208,6 +207,7 @@ class NativeWindow : public base::SupportsUserData,
   virtual void SetTrafficLightPosition(base::Optional<gfx::Point> position) = 0;
   virtual base::Optional<gfx::Point> GetTrafficLightPosition() const = 0;
   virtual void RedrawTrafficLights() = 0;
+  virtual void UpdateFrame() = 0;
 #endif
 
   // Touchbar API
--- a/shell/browser/native_window_mac.h
+++ b/shell/browser/native_window_mac.h
@@ -14,6 +14,7 @@
 
 #include "base/mac/scoped_nsobject.h"
 #include "shell/browser/native_window.h"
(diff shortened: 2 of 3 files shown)


shell/browser/native_window_mac.mm
```diff
@@ -238,6 +239,12 @@ void SetFrameSize(NSView* self, SEL _cmd, NSSize size) {
   super_impl(self, _cmd, size);
 }
 
+// In simpleFullScreen mode, update the frame for new bounds.
+void UpdateFrame(NSWindow* window) {
+  NSRect fullscreenFrame = [window.screen frame];
```
deepak1556, line 244: Is it safe to access `NSWindow` in the callback, is there a guarantee that it will be alive when called ?

Can the post task call be changed to act on `weakptr` receiver, like

```
base::PostTask(
      FROM_HERE, {content::BrowserThread::UI},
      base::BindOnce(&NativeWindow::UpdateFrame, GetWeakPtr()));
```

jho293:
Bantu aku dengan $1 aku dalam kesusahan uang

Kalau berkenan transfer via rekening mandiri

No. 152-00-1748706-3

Terima kasih orang baik



JhoynSannMalakian

Pada tanggal Jum, 12 Mar 2021 16.23, Robo ***@***.***>
menulis:

shell/browser/native_window_mac.mm
codebytere: sure - sounds good to me! i think it should be safe to do as is but minds well be safer

deepak1556:
Compilation errors on Linux needs fix, LGTM otherwise. Thanks!

codebytere:
@deepak1556 should be set to go!

shell/browser/native_window.h
```diff
@@ -147,6 +147,7 @@ class NativeWindow : public base::SupportsUserData,
   virtual bool IsExcludedFromShownWindowsMenu() = 0;
   virtual void SetSimpleFullScreen(bool simple_fullscreen) = 0;
   virtual bool IsSimpleFullScreen() = 0;
+  virtual void UpdateFrame() = 0;
```
deepak1556, line 150: minor style nit: This can be behind [#if defined(OS_MAC)](https://github.com/electron/electron/pull/28150/files#diff-bcde70ba560ca4f196ebb71ba43f10d2551c112219bb3474647ee50312b74f17R205) below which avoids having to implement noop for other platforms.

deepak1556 (approved):
LGTM with style change.

shell/browser/native_window_mac.h
```diff
@@ -86,6 +89,7 @@ class NativeWindowMac : public NativeWindow, public ui::NativeThemeObserver {
   bool IsExcludedFromShownWindowsMenu() override;
   void SetSimpleFullScreen(bool simple_fullscreen) override;
   bool IsSimpleFullScreen() override;
+  void UpdateFrame() override;
```
zcbenz, line 92: Can you put the method in the same order with the base class? i.e. put it under the `RedrawTrafficLights` method, or move the declaration in base class to bellow `IsSimpleFullScreen`.

shell/browser/native_window_mac.mm
```diff
@@ -882,6 +886,24 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
   [window setExcludedFromWindowsMenu:excluded];
 }
 
+// In simpleFullScreen mode, update the frame for new bounds.
+void NativeWindowMac::UpdateFrame() {
```
zcbenz, line 890: Can you put the method implementations in the same order with the declarations?

zcbenz (approved):

PR #28156, opened 2021-03-12
test: disable some tests under ASan which might receive SIGKILL because of OOM

zcbenz:
#### Description of Change

Under ASan build, the tests that spawn child Electron apps may receive SIGKILL because of out of memory.

I believe Circle CI does not allow enabling swap space, so disabling those tests under ASan build seems to be the best choice.

This PR also moves some tests to `spec-main/fixtures/crash-cases`.

#### Release Notes

Notes: none


codebytere (approved):

jkleinsc (changes requested):
Looks like some of the newly enabled tests are failing

zcbenz:
@jkleinsc This PR only disables more tests, so it is not responsible for the new failing tests.

I have pushed another commit to also disable webview tests for ASan, which has the same problem that the renderer process of webview are being killed because of out of memory.

The asan tests are green now.

jkleinsc (approved):

Issue #28154, opened 2021-03-12
[Bug]: Renderer Promise API stops working after page reload when allowRendererProcessReuse is true


VeryCrazyDog:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/electron/electron/issues) for a feature request that matches the one I want to file, without success.

### Electron Version

12.0.0

### What operating system are you using?

Windows

### Operating System Version

10

### What arch are you using?

x64

### Expected Behavior

From the reproduce steps, expecting to see local files printed in the console:

...
preload.js:11 Reading file .gitignore
preload.js:11 Reading file index.html
preload.js:11 Reading file LICENSE.md
preload.js:11 Reading file main.js
preload.js:11 Reading file package-lock.json
...

### Actual Behavior

List is printed the first time the app loads. Hit Ctrl+R and observe only one line printed, or none at all.

### Testcase Gist URL

https://gist.github.com/e0d8e593f6f2d737a795d8e20ab5a37d

### Additional Information

It might be related to issue #22119

codebytere:
@VeryCrazyDog looks like this was re-introduced when i fixed another issue: https://github.com/electron/electron/pull/27582. I'll try to get a fix in soon.

lishid:
In the closed bug report, there was an additional issue that was not fixed in v11: (https://github.com/electron/electron/issues/22119#issuecomment-751515274)

> After many many sessions of debugging I found a repro (I'm on Windows 10):
> 
> - Perform a long-running async node call (in my case, the call was `fs.writeFile`)
> - Reload the page before the call finishes.
> - Observe that any subsequent node calls never triggers the callback.
> - Reload again fixes the issue.

Not sure if this helps, but hopefully it can be resolved together 🙏

aabuhijleh:
@lishid which v11 version are you using? This bug is supposed to be re-introduced in 11.4.0

However, this PR is still open https://github.com/electron/electron/pull/28336

The fix has not been ported to the 11.x.y series yet

lishid:
I have only tested up until 11.3.0.

PR #28175, opened 2021-03-13
fix: libuv hang on Windows

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/28154.

https://github.com/electron/electron/pull/25869 used `uv_loop_->iocp` to decide whether to re-prep the message loop, and it turns out not to be the case that it works and updates the same way that `uv_backend_fd` does, so what resulted was that when `allowRendererProcessReuse` was true the loop would not re-prep on reload and libuv would hang.

This fixes that by only handling this case on non-Windows systems.

Tested with https://gist.github.com/codebytere/2face1c26d09e38c06f49111c3fca13f.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where some Node.js modules would hang on page reload on Windows.

### DIFF
--- a/shell/common/node_bindings.cc
+++ b/shell/common/node_bindings.cc
@@ -534,15 +534,13 @@ void NodeBindings::LoadEnvironment(node::Environment* env) {
 void NodeBindings::PrepareMessageLoop() {
 #if !defined(OS_WIN)
   int handle = uv_backend_fd(uv_loop_);
-#else
-  HANDLE handle = uv_loop_->iocp;
-#endif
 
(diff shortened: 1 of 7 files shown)


MarshallOfSound (commented):
Is there _any_ way we can test for this?

codebytere:
@MarshallOfSound i'll see what i can do 🤔

jkleinsc (approved):

MarshallOfSound (approved):

codebytere:
Asan test failures are unrelated - merging.

Issue #27838, opened 2021-02-21
isMaximized does not reflect OS window state on windows


suedama1756:
Changes to isMaximized implementation recently have made reporting of isMaximized worse as they do not use standard operating system calls  to determine the window state. Maximizing a window using the windows system menu, or using WIN + Up Arrow will always return isMaximized() false.  

### Preflight Checklist

* [x ] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [ x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [ x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  Seen in 10.3.2 and 11.2.3
* **Operating System:**
  Windows 10
* **Last Known Working Electron version:**
  Not 100% sure as different versions have various issues with isMaximized.

### Expected Behavior
Execute 

```js
   const { app, BrowserWindow } = require('electron');
  app.on('ready', function() {
     const window = new BrowserWindow({
         frame: false,
         thickFrame: false,
         width: 200,
         height: 200  
     });
     window.show();
     setInterval(() => {
        if (!window.isDestroyed()) {
            console.log(window.isMaximized())
        }
     }, 5000);
  });
```
Should return true when the window has been maximized by the operating system

### Actual Behavior
Returns false

### To Reproduce

1. Enter the code above into a file named main.js and execute with 
```bash
electron main.js
```
2. maximize the window using  "alt+space" menu and select maximize, or use "WIN+Up Arrow" shortcut

### Additional Information

Replacing the implementations on BrowserWindow with calls to native operating system APIs via a NAPI extension fixes this issue for me. I'd love to understand why electron does not use these standard calls as the code is simpler and should be more reliable (perhaps there were legacy issues, if so do they still apply?)

```js
   isMaximized() { 
       return Native.getWindowPlacement(this.getNativeWindowHandle()).showCmd === 3;
   },
   maximize() {
       Native.showWindow(window.getNativeWindowHandle(), 3);
   }
```

The NAPI extension is simply a wrapper on the following two windows APIs https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowplacement and https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-showwindow

Issue #27264, opened 2021-01-11
Circular corners in the main window does not work correctly


jgomez109:
### Issue Details

* **Electron Version:**
  * electron@11.1.1
* **Operating System:**
  * Windows-10-10.0.19041-SP0

### Expected Behavior
 Window with circular corners

### Actual Behavior
When the window is maximized and then minimized, the square and black borders appear.

### To Reproduce
Just put in the main.js 
``trasnparent = true,  frame = false
``
And with html and css create  the circular corners (Below I add the code)

### Screenshots
![pb](https://user-images.githubusercontent.com/56138728/104198981-e429c380-53f4-11eb-940a-2dbe485a83c4.jpg)

#### Main.js:
```const { app, BrowserWindow } = require('electron')

let win;

function createWindow () {
  win = new BrowserWindow({
    width: 800,
    height: 600,
    minWidth:400,
    minHeight: 300,
    frame: false,
    transparent: true,
    webPreferences: {
      nodeIntegration: true
    }
  })

  win.loadFile('index.html')
}


app.whenReady().then(createWindow)

app.on('window-all-closed', () => {
  if (process.platform !== 'darwin') {
    app.quit()
  }
})

app.on('activate', () => {
  if (BrowserWindow.getAllWindows().length === 0) {
    createWindow()
  }
})
```
#### index.html:
```<!DOCTYPE html>
<html>

<head>
    <meta charset="UTF-8">
    <link rel="stylesheet" href="styles.css">
</head>

<body>
    <div class="Body">

        <div class="WindowBar">
            <div class="LogoTitleBar">
                <div class="LogoBar">
                    <img class="LogoBar"
                        src="https://i.pinimg.com/originals/46/0e/78/460e78cb4c61f55da3802ca5d1d68d15.jpg" alt="">
                </div>
                <div class="TitleBar">
                    My App
                </div>
            </div>
            <div class="MenuBar">
                <div class="btn btn1"></div>
                <div class="btn btn2"></div>
                <div class="btn btn3"></div>
            </div>
        </div>
        <div class="content">
            <!-- <img class="ImgContent" src="https://wallpaperaccess.com/full/3542858.png" alt=""> -->
        </div>


    </div>

</body>

</html>
```

#### styles.css:
```html,
body {
    width : 100%;
    height: 100vh;
    margin: 0;
}

:root {
    --radius_value: 15px;
    --radius_btn  : 15px;
}



.Body {
    width            : 100%;
    height           : 100vh;
[...]

PR #6417, opened 2016-07-09
Correctly handle maximize/fullscreen in window without thickFrame

zcbenz:
For window without thickFrame, we have to keep them always in normal window state, otherwise weird things will happen.

Close #1690.

PR #28207, opened 2021-03-16
fix: enable system maximization for frameless windows except if transparent

mlaurencin:
#### Description of Change

Closes https://github.com/electron/electron/issues/27838
Closes https://github.com/electron/electron/issues/27264

Previously windows without a frame (including transparent windows) used a workaround in order to maximize properly (See these two PRs: https://github.com/electron/electron/pull/6417 & https://github.com/electron/electron/pull/26586). With the latest change however, windows without frames could no longer be properly maximized using the Windows system menu (accessible through `alt + space`).

With this PR, the ability to maximize and unmaximize using the Windows system menu and double clicking the title bar, has been disabled for transparent windows. These windows will only be maximizable through Electron's API. Non-transparent frameless windows no longer use the workaround and can be maximized using Electron's API or the Windows system menu interchangeably.

[Fiddle for testing](https://gist.github.com/mlaurencin/bd78bc0378dc2060a3571797e3169e39)

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes:  Transparent windows cannot be maximized using the Windows system menu or by double clicking the title bar.

### DIFF
--- a/shell/browser/native_window_views_win.cc
+++ b/shell/browser/native_window_views_win.cc
@@ -150,7 +150,7 @@ HHOOK NativeWindowViews::mouse_hook_ = NULL;
 
 void NativeWindowViews::Maximize() {
   // Only use Maximize() when window has WS_THICKFRAME style
-  if (::GetWindowLong(GetAcceleratedWidget(), GWL_STYLE) & WS_THICKFRAME) {
+  if (!transparent()) {
     if (IsVisible())
       widget()->Maximize();
     else
@@ -323,6 +323,12 @@ bool NativeWindowViews::PreHandleMSG(UINT message,
       NotifyWindowSystemContextMenu(GET_X_LPARAM(l_param),
                                     GET_Y_LPARAM(l_param), &prevent_default);
       return prevent_default;
+    }
+    case WM_SYSCOMMAND: {
+      if (transparent() && w_param == SC_MAXIMIZE) {
+        return true;
+      }
+      return false;
     }
     default:
       return false;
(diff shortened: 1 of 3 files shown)


nornagon:
Does this break a valid use case where people might _want_ to be able to allow a user to maximize a transparent window with alt+space / titlebar double-click?

mlaurencin:
> Does this break a valid use case where people might _want_ to be able to allow a user to maximize a transparent window with alt+space / titlebar double-click?

No, this does not break the use case, but that's because it was not working correctly before either. Unfortunately, both using `alt+space` and double clicking the title bar cause the system maximization to be called, which breaks the workaround I mentioned in the PR body. I was able to narrow down its usage to only transparent windows (which are a subset), but the limitations still exist. By disabling these cases, these additional limitations are now being made explicit. (I am adding them to the list of transparent window limitations in the documentation with this next commit)

I agree that those are valid use cases though, so disabling these functionalities is just handling the current issues and a new workaround may be good to explore in the future.

shell/browser/native_window_views_win.cc
```diff
@@ -324,8 +324,31 @@ bool NativeWindowViews::PreHandleMSG(UINT message,
...
+      if (transparent() && ((w_param & max_mask) == SC_MAXIMIZE)) {
+        return true;
+      }
+      return false;
+    }
+    case WM_INITMENU: {
+      // This is handling the scenario where the menu might get triggered by the
+      // user doing "alt + space" resulting in system maximization and restore
+      // being used on transparent windows when that does not work.
+      HMENU menu = GetSystemMenu(GetAcceleratedWidget(), false);
```
zcbenz, line 339: This can be moved into the `if (transparent()) { ... }`.

zcbenz (approved):

codebytere (approved):

ckerr (approved):

PR #28221, opened 2021-03-16
docs: correct example main.js in quick start

tomuky2k:
#### Description of Change
Add contextIsolation: false to the quick start guide, so the example code works correctly.
Error shown on https://www.electronjs.org/docs/tutorial/quick-start

#### Checklist

- [x] PR description included

#### Release Notes

fix example code in manual for quick start guide.

### DIFF
--- a/docs/tutorial/quick-start.md
+++ b/docs/tutorial/quick-start.md
@@ -61,7 +61,8 @@ function createWindow () {
     width: 800,
     height: 600,
     webPreferences: {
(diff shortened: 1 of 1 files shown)


jkleinsc:
Thanks for the PR @tomuky2k!  While this solution technically works, the preferred way to handle this is with a preload script.  Since #28223 makes that change, I am going to close this PR in favor of #28223.

PR #28230, opened 2021-03-17
test: spellchecker may take several minutes to load under ASan

zcbenz:
#### Description of Change

The spellchecker may take several minutes to load under ASan, give it 10 minutes to load.

#### Release Notes

Notes: none

### DIFF
--- a/spec-main/spellchecker-spec.ts
+++ b/spec-main/spellchecker-spec.ts
@@ -10,7 +10,7 @@ const features = process._linkedBinding('electron_common_features');
 const v8Util = process._linkedBinding('electron_common_v8_util');
 
 ifdescribe(features.isBuiltinSpellCheckerEnabled())('spellchecker', function () {
-  this.timeout(200 * 1000);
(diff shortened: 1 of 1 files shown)


nornagon:
several minutes :O

is there some way we can reduce the load time?

zcbenz:
It only takes so long when running in CI, I don't know why but I think it is probably because the CI does not have much RAM left and the disk read is very slow, so I can't think of a way to make it faster.

The good news is the spellchecker is only slow for the first load, once it has been loaded for one time the following tests would run fast, so the whole spellchecker tests wouldn't take like half hour to run.

codebytere (approved):

nornagon:
@zcbenz would it be possible we could load a smaller "testing" dictionary?

zcbenz:
> would it be possible we could load a smaller "testing" dictionary?

Yeah it is possible with existing APIs and I think it would be a very good way to make the tests faster and more stable.

But it would take time to figure out how to generate a custom dictionary and rewrite the tests, so I prefer to increase timeout for now and fix other CI failures first, and I have added a note in the tests.

Issue #28197, opened 2021-03-15
[Bug]: Electron not compatible with latest selenium-webdriver


kr3l:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/electron/electron/issues) for a feature request that matches the one I want to file, without success.

### Electron Version

10.1.3

### What operating system are you using?

Windows

### Operating System Version

Windows 10.0.19042 Build 19042

### What arch are you using?

x64

### Expected Behavior

Should start electron.

This is the same issue as https://github.com/electron/electron/issues/22408, which was closed because it happened on an unsupported electron version. 

### Actual Behavior

Fails with (node:20272) UnhandledPromiseRejectionWarning: SessionNotCreatedError: session not created: No matching capabilities found

To reproduce, follow the instructions https://www.electronjs.org/docs/tutorial/using-selenium-and-webdriver under 'Setting up with WebDriverJS'. The instructions work for selenium-webdriver 3.4.0, but no longer work for more recent versions of selenium-webdriver, like selenium-webdriver 4.0.0-beta.1.

I used the following test file:

```
const webdriver = require('selenium-webdriver');
const ChildProcess = require('child_process');
const chromedriver = require('chromedriver');

const chromedriverProcess = ChildProcess.spawn(
    chromedriver.path, // process path goes here
    [], // arguments go here
    {
        stdio: 'inherit'
    }
);
const electronBinary = 'MyElectronBinary.exe';
const driver = new webdriver.Builder()
    // The "9515" is the port opened by chrome driver.
    .usingServer('http://localhost:9515')
    .withCapabilities({
        chromeOptions: {
            // Here is the path to your Electron binary.
            binary: electronBinary
        }
    })
    .forBrowser('electron')
    .build();

driver.get('http://www.google.com');
driver.findElement(webdriver.By.name('q')).sendKeys('webdriver');
driver.findElement(webdriver.By.name('btnG')).click();
driver.wait(() => {
    return driver.getTitle().then((title) => {
        return title === 'webdriver - Google Search';
    });
}, 1000);

driver.quit();
```

which fails to start electron from the test and results in this error:

```
"C:\Program Files\nodejs\node.exe" test.js
[...]

PR #28231, opened 2021-03-17
docs: updated example to work with latest versions of selenium-webdriver

kr3l:
#### Description of Change

See #28197 . The existing snippet works with selenium-webdriver <= 3.6.0, but any more recent version seems to require using 'goog:chromeOptions' and forBrowser('chrome').

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/docs/tutorial/using-selenium-and-webdriver.md
+++ b/docs/tutorial/using-selenium-and-webdriver.md
@@ -86,12 +86,12 @@ const driver = new webdriver.Builder()
   // The "9515" is the port opened by chrome driver.
   .usingServer('http://localhost:9515')
   .withCapabilities({
-    chromeOptions: {
+    'goog:chromeOptions': {
       // Here is the path to your Electron binary.
       binary: '/Path-to-Your-App.app/Contents/MacOS/Electron'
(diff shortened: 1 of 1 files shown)


VerteDinde (approved):
Nice, thanks for taking the time to update this! 😄

docs/tutorial/using-selenium-and-webdriver.md
```diff
@@ -86,12 +86,12 @@ const driver = new webdriver.Builder()
   // The "9515" is the port opened by chrome driver.
   .usingServer('http://localhost:9515')
   .withCapabilities({
-    chromeOptions: {
+    'goog:chromeOptions': {
```
nornagon, line 89: does `goog:chromeOptions` work with older versions too?

kr3l: Yes, I tested in the older selenium-webdriver 3.4.0 and 3.6.0 versions and `goog:chromeOptions` works.

nornagon (approved):

codebytere (approved):

PR #28245, opened 2021-03-17
docs: update Node global symbols example to use contextBridge

itsananderson:
#### Description of Change

This is another piece of example code that no longer works with `contextIsolation: false`. Since you can't expose values through `global` anymore, it no longer made sense to have the example live in the `process` `'loaded'` event docs. I moved it to the `contextBridge` docs since that's what you'd now use to expose Node APIs.

#### Checklist

#### Release Notes

Notes: no-notes

### DIFF
--- a/docs/api/context-bridge.md
+++ b/docs/api/context-bridge.md
@@ -122,7 +122,7 @@ Be very cautious about which globals and APIs you expose to untrusted remote con
 const { contextBridge } = require('electron')
 const crypto = require('crypto')
 contextBridge.exposeInMainWorld('nodeCrypto', {
-  sha256sum(data) {
+  sha256sum (data) {
     const hash = crypto.createHash('sha256')
     hash.update(data)
     return hash.digest('hex')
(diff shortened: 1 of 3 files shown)


docs/api/context-bridge.md
```diff
@@ -110,3 +110,16 @@ has been included below for completeness:
...
+
+The `contextBridge` can be used by the preload script to give your renderer access to Node global symbols.
+The table of supported types described above also applies to Node globals that you expose through `contextBridge`.
+Please note that many Node APIs grant access to local system resources.
+Be very cautious about which globals and APIs you expose to untrusted remote content.
+
+```javascript
+const { contextBridge } = require('electron')
+contextBridge.exposeInMainWorld('setImmediate', setImmediate)
+contextBridge.exposeInMainWorld('clearImmediate', clearImmediate)
```
nornagon, line 124: Perhaps we could have a more useful/relevant example than `setImmediate` here? And perhaps also show an example of not directly exposing Node stuff, but rather exposing a safe wrapper.

```suggestion
const crypto = require('crypto')
contextBridge.exposeInMainWorld('nodeCrypto', {
  sha256sum(data) {
    const hash = crypto.createHash('sha256')
    hash.update(data)
    return hash.digest('hex')
  }
})
```

jkleinsc (approved):

zcbenz (approved):

PR #28274, opened 2021-03-18
feat: add webFrameMain.findInFrame/stopFindInFrame

deepak1556:
#### Description of Change

Refs https://github.com/microsoft/vscode/issues/96307

An app switching from `<webview>` to use OOPIF cannot use the current find functionality exposed on `electron::api::WebContents::Find` as it will initiate search from the top-level WebContents. The new api allows users to target a particular frame where the search has to be initiated from.

- [ ] Upstream patch, pending discussion https://bugs.chromium.org/p/chromium/issues/detail?id=1189789
- [ ] API WG review

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added webFramMain.findInFrame/stopFindInFrame


shell/browser/api/electron_api_web_frame_main.cc
```diff
@@ -316,13 +348,39 @@ void WebFrameMain::Connect() {
   }
 }
 
+content::FindRequestManager* WebFrameMain::GetOrCreateFindRequestManager() {
```
samuelmaddock, line 351: Need to check that the RFH is still alive.

```suggestion
content::FindRequestManager* WebFrameMain::GetOrCreateFindRequestManager() {
  if (!CheckRenderFrame())
    return nullptr;
```

shell/browser/api/electron_api_web_frame_main.cc
```diff
@@ -122,6 +127,33 @@ bool WebFrameMain::Reload() {
...
+  gin_helper::Dictionary dict;
+  auto options = blink::mojom::FindOptions::New();
+  if (args->GetNext(&dict)) {
+    dict.Get("forward", &options->forward);
+    dict.Get("matchCase", &options->match_case);
+    dict.Get("findNext", &options->new_session);
+  }
+
+  GetOrCreateFindRequestManager()->Find(request_id, search_text,
+                                        std::move(options));
```
samuelmaddock, line 148:
```suggestion
  auto find_request_manager = GetOrCreateFindRequestManager();
  if (!find_request_manager)
    return 0;

  find_request_manager->Find(request_id, search_text, std::move(options));
```

shell/browser/api/electron_api_web_frame_main.cc
```diff
@@ -316,13 +348,39 @@ void WebFrameMain::Connect() {
...
+    // No existing FindRequestManager found, so one must be created.
+    auto* web_contents = static_cast<content::WebContentsImpl*>(
+        content::WebContents::FromRenderFrameHost(render_frame_));
+    find_request_manager_ = std::make_unique<content::FindRequestManager>(
+        web_contents,
+        static_cast<content::RenderFrameHostImpl*>(render_frame_));
+  }
+  // Concurrent find sessions must not overlap, so destroy any existing
+  // FindRequestManagers in any inner WebFrameMain.
+  for (content::RenderFrameHost* rfh : render_frame_->GetFramesInSubtree()) {
```
samuelmaddock, line 362: If a find session exists in a parent frame, would that be an issue? If so, this might need to be `render_frame_->GetMainFrame()->GetFramesInSubtree()`.

samuelmaddock (changes requested):
It would be great if the WebContents.findInPage/stopFindInPage could be replaced with this new implementation. ie. just calling `webContents.mainFrame.findInFrame` internally.

shell/browser/api/electron_api_web_frame_main.cc
deepak1556: When does this happen ?

samuelmaddock: In case a reference to the frame is held after its been destroyed.
```js
const { mainFrame } = browserWindow.webContents;
browserWindow.close();
mainFrame.findInFrame(...);
```

It can also occur upon navigation in some cases.

nornagon (changes requested):
marking request changes until we have more info on upstreaming.

shell/browser/api/electron_api_web_frame_main.cc
deepak1556: Good catch, yup it should be `render_frame_->GetMainFrame()->GetFramesInSubtree()`

deepak1556:
> It would be great if the WebContents.findInPage/stopFindInPage could be replaced with this new implementation. ie. just calling webContents.mainFrame.findInFrame internally.

They are different in behavior, hence I haven't replaced them. `FindInPage` will go over the frames in all `WebContents` of a page (i-e) if a top-level `WebContents` can have multiple child `WebContents`. This api is only for frames in the same `WebContents`

mjbvz:
@deepak1556 @ rzhao271 Thanks for looking into this.

I just tried it out for VS Code but once difference I noticed compared to search inside `<webviews>` is that find seems to focus the iframe. This causes problems when typing in our custom search widget, since the text input keeps losing focus

Any ideas on if that's something this API could change or if we can workaround this?

deepak1556:
> Can we do some subclassing instead? Can we apply the filtering at a different point? Any thoughts?

The filtering in this case has to be performed at `FindRequestManager`, basically this class owns a per frame client called [FindInPageClient](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/find_in_page_client.h) that makes the find request with blink and awaits for response. `FindRequestManager` then combines the results from all the `FindInPageClient` for a given tree including out of process frames and allows traversal of the active matches.

I will first update the branch against main since the current patch is outdated and will check if there is a possibility to minimize it.

ckerr:
@deepak1556 is this PR still live?

deepak1556:
There is need to upstream the frame find implementation but don't have cycles to visit this anytime soon. Closing for now.

PR #28285, opened 2021-03-18
perf: do not double-proxy methods being return over the contextBridge

MarshallOfSound:
Previously a function sent over the bridge would result in a proxy function existing in the other context, sending it back would create a proxy-of-a-proxy and so on.  With each successive pass over the bridge it would add another layer of proxy'ing which would get slower and slower.  This change ensures that when a method is cleanly passed back over the bridge to it's origin we use the stored "original function" as the return value thus ensuring a max proxy depth of 1.

This is safe to do for function because the lifetime of the proxy function is _already_ tied to the lifetime of the original function.  We can't do this for others objects / types without some smarts and last time we tried to implement those smarts it caused massive issues.

Notes: Improved performance characteristics of functions being sent back and forth over the contextBridge.

### DIFF
--- a/shell/renderer/api/electron_api_context_bridge.cc
+++ b/shell/renderer/api/electron_api_context_bridge.cc
@@ -34,6 +34,8 @@ namespace context_bridge {
 const char* const kProxyFunctionPrivateKey = "electron_contextBridge_proxy_fn";
 const char* const kSupportsDynamicPropertiesPrivateKey =
     "electron_contextBridge_supportsDynamicProperties";
+const char* const kOriginalFunctionPrivateKey =
+    "electron_contextBridge_original_fn";
 
 }  // namespace context_bridge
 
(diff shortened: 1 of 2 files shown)


codebytere (approved):

shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -172,9 +174,26 @@ v8::MaybeLocal<v8::Value> PassValueToOtherContext(
...
+      // If this function has already been sent over the bridge,
+      // then it is being sent _back_ over the bridge and we can
+      // simply return the original method here for performance reasons
+
+      // For safety reasons we check if the destination context is the
+      // creation context of the original method.  If it's not we proceed
+      // with the proxy logic
+      if (maybe_original_fn.ToLocal(&proxy_func) && proxy_func->IsFunction() &&
+          proxy_func.As<v8::Object>()->CreationContext() ==
+              destination_context) {
```
nornagon, line 193: feels like some of these should be `CHECK` or maybe even `DCHECK`
```suggestion
      if (maybe_original_fn.ToLocal(&proxy_func)) {
        DCHECK(proxy_func->IsFunction());
        DCHECK(proxy_func.As<v8::Object>()->CreationContext() ==
               destination_context);
```

since i don't think there's any way these could _not_ be true?

MarshallOfSound: I can make the `CreationContext` check fail by using native modules are even just the `vm` module in NodeJS.  Because it was possible for legit user code to trip it I made it a silent condition rather than a hard `CHECK`

jkleinsc (approved):

Issue #27704, opened 2021-02-11
python-dbusmock & libgnome-keyring-dev are not available on Ubuntu 20


RangerMauve:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 13.0.0-nightly.20210210 
* **Operating System:**
  * Pop!_OS 20.10 
* **Last Known Working Electron version:**
  * N/A 

### Expected Behavior

When following the [build instructions for linux](https://www.electronjs.org/docs/development/build-instructions-linux#prerequisites), I expected the command to install dependencies on Ubuntu to run.

### Actual Behavior

How ever, it errors out saying that `libgnome-keyring-dev` could not be located and that `python-dbusmock` has no installation candidate (seems to be available for Ubuntu 18).

```
sudo apt-get install build-essential clang libdbus-1-dev libgtk-3-dev \
>                        libnotify-dev libgnome-keyring-dev \
>                        libasound2-dev libcap-dev libcups2-dev libxtst-dev \
>                        libxss1 libnss3-dev gcc-multilib g++-multilib curl \
>                        gperf bison python-dbusmock openjdk-8-jre
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package python-dbusmock is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Unable to locate package libgnome-keyring-dev
E: Package 'python-dbusmock' has no installation candidate
```

### To Reproduce

Try to build electron on Ubuntu 20 with electron's build tools.

### Additional Information

Trying to install the python-dbusmock using pip lead to an error about syntax:

```
unning: Main process specs
Traceback (most recent call last):
  File "/home/mauve/programming/electron/src/electron/script/dbus_mock.py", line 7, in <module>
    from dbusmock import DBusTestCase
  File "/home/mauve/.local/lib/python2.7/site-packages/dbusmock/__init__.py", line 14, in <module>
    from dbusmock.mockobject import (DBusMockObject, MOCK_IFACE,
[...]

PR #28288, opened 2021-03-19
build: add boringssl headers to node_headers tarball

### LINKED ISSUE #27751: Header distribution tar.gz is missing BoringSSL files
### Preflight Checklist

### Issue Details

* **Electron Version:**
  *  v11.2.3
* **Operating System:**
  *  N/A
  *  Don't know

### Expected Behavior

Downloading https://www.electronjs.org/headers/v11.2.3/node-v11.2.3-headers.tar.gz should give me the boringssl header files just like the corresponding Node.js header distribution gives openssl header files.

### Actual Behavior

Headers are missing for BoringSSL

### To Reproduce

Download https://www.electronjs.org/headers/v11.2.3/node-v11.2.3-headers.tar.gz and look for openssl or boringssl headers

seesemichaelj:
#### Description of Change

NodeJS adds OpenSSL headers to the header tarball, but Electron
currently doesn't add the BoringSSL headers to the header tarball.
This patch adds the necessary build commands to copy the headers
into the `node_headers` directory prior to tarballing.

This PR resolves #27751

CC'ing others interested in this PR: @davidmurdoch @alexhultman

#### Checklist

- [x] PR description included and stakeholders cc'd (I'm not sure who the Electron maintainer stakeholders would be)
- [ ] `npm test` passes (have not been able to run them yet)
    - I'm not sure if these need to pass for this PR, but I'm having issues running the tests due to issues with dbusmock (same issue as #27704)
    - I will try to resolve these issues tomorrow and see if I get tests to run.

#### Release Notes

Notes: Added BoringSSL headers to node_headers tarball.

### DIFF
--- a/patches/node/build_add_gn_build_files.patch
+++ b/patches/node/build_add_gn_build_files.patch
@@ -7,10 +7,10 @@ This adds GN build files for Node, so we don't have to build with GYP.
 
 diff --git a/BUILD.gn b/BUILD.gn
 new file mode 100644
-index 0000000000000000000000000000000000000000..446119163d1f7bad577cb0b7b217ecf24b994526
+index 0000000000000000000000000000000000000000..864a8572e70bf71c39876b19a15397ab867cbf7d
 --- /dev/null
 +++ b/BUILD.gn
-@@ -0,0 +1,360 @@
+@@ -0,0 +1,377 @@
 +import("//electron/build/asar.gni")
 +import("//v8/gni/v8.gni")
 +
@@ -307,6 +307,22 @@ index 0000000000000000000000000000000000000000..446119163d1f7bad577cb0b7b217ecf2
 +  header_group_index += 1
 +}
 +
++import("//third_party/boringssl/BUILD.generated.gni")
++
++all_boringssl_headers = crypto_headers + ssl_headers
++boringssl_header_index = 0
++foreach(header, all_boringssl_headers) {
++  all_boringssl_headers[boringssl_header_index] = "../boringssl/$header"
++  boringssl_header_index += 1
++}
++
++copy("boringssl_headers") {
++  sources = all_boringssl_headers
++  outputs = [
++    "$node_headers_dir/include/node/boringssl/{{source_file_part}}",
++  ]
++}
++
 +copy("zlib_headers") {
 +  sources = [
 +    "deps/zlib/zconf.h",
@@ -363,6 +379,7 @@ index 0000000000000000000000000000000000000000..446119163d1f7bad577cb0b7b217ecf2
 +                  ":zlib_headers",
 +                  ":node_gypi_headers",
 +                  ":node_version_header",
++                  ":boringssl_headers",
 +                ]
 +}
 +


seesemichaelj:
@codebytere Do you know if the errors in the CI builds are applicable? Seems like the errors that are actually being reported are credential issues, is  this correct or is there something I can/should do to fix these errors?

```
GOMA:clang++ compiling ../../third_party/jsoncpp/source/src/lib_json/json_writer.cpp:*ERROR*: compiler_proxy [478.4521ms]: 
**************************************************************
Not logged in. Run this command and retry the build:
    pushd $GOMACLIENTDIR && vpython goma_auth.py login && popd
**************************************************************
```

Thanks!

seesemichaelj:
I've tried to run tests, but the issues at #27704 are preventing me from running the tests on my Ubuntu 20.04 setup. I can try getting a Windows environment set up to run the tests there if requested by the maintainers? I'm not sure how necessary this PR is to have tests, so don't want to put forward the effort unless explicitly requested.

Thanks for your time and consideration!

alexhultman:
Another question is whether the node.lib on Windows and node.exe exports the BoringSSL **symbols** or not. On regular Node.js it did not, so back in 2016 I PR:ed this.

Also I wonder how it is possible for Electron to have introduced a new ABI numbering system **only based on the fact BoringSSL is used**, while not even exporting said BoringSSL support to begin with - feels like a thought was dropped halfway through here - or am I missing something?

nornagon:
@alexhultman the main reason for the different ABI numbering is that we build a different version of V8 than Node.js, and apply patches as needed to accommodate.

nornagon:
and, also to @alexhultman's point, we do not export BoringSSL symbols, so this PR is somewhat useless without that.

Exporting BoringSSL symbols has been [discussed in the past](https://github.com/electron/electron/issues/13176#issuecomment-610501697) and decided against, so unless something significant has changed since then, I'm still 👎  on exporting these symbols.

nornagon (changes requested):
(as above)

alexhultman:
> so this PR is somewhat useless without that.

This PR is **entirely useless, dangerous and invalid** unless you also export the matching symbols.

nornagon:
@alexhultman thanks, I've already stated that, please keep things calm and professional here as this is a workplace :)

alexhultman:
From where did you get the impression I was not calm? I added more information. You said it was "somewhat useless" I said it is entirely useless and also dangerous (since people can end up linking against wrong version, which breaks everything in subtle ways). You do not need to tell me how to act in public, I am not an animal.

seesemichaelj:
Thanks @alexhultman for providing your input on the necessities of exporting the symbols for this PR to be useful; makes sense, and I agree that the symbols would need to be exported to make sense.

Thank you @nornagon for the swift response to looking at this PR! I really appreciate the maintainers here at Electron and the time you take to address outside contributions.

I do think that the conversation is getting heated, but we can still recover and have a productive discourse about this.

For what it's worth, I agree with sentiments in the comment https://github.com/electron/electron/issues/13176#issuecomment-610305707, but I disagree with Electron's desire to diverge away from NodeJS practices. I understand that the group that maintains Electron is not associated with the group that maintains NodeJS, and you're providing an awesome tool for the public to benefit from. So what you end up deciding is the way it goes, and that's totally understandable. However, application developers that use NodeJS expect (maybe they shouldn't, I'm not sure) a mostly fluid transition using Electron from using NodeJS. I think it's important for large-reaching tools like Electron and NodeJS to stay lock-step in terms of how things are delivered unless absolutely necessary (i.e. need to use BoringSSL due to using Chromium), etc.

There's a really high chance I don't understand the full picture, and I apologize for my ignorance, but if Electron maintainers have strong opinions to not exporting symbols, then why are they still exported in NodeJS? This divergence causes industry issues of dependencies and tools that rely on the technology stack of NodeJS or Electron, but then hit a wall when trying to move forward with migrating/adopting the other library. If there was consistency, then we wouldn't be in this position.

Again, y'all are providing a free tool for everyone to thrive on, but being consistent (either by adopting the practices of NodeJS or by getting NodeJS to adopt Electron's practices) to me is the most important takeaway here. The outcome may cause issues (i.e. if NodeJS stops exporting the OpenSSL symbols, then @alexhultman's super valuable, extremely fast websocket server library https://github.com/uNetworking/uWebSockets.js would need potentially major changes [I'm not sure of the impact]).

This PR's hope was to try to close the divide between NodeJS and Electron one step closer. I'm happy to try to amend this PR to export the BoringSSL symbols if it's decided that Electron maintainers would be open to that. However, we (@davidmurdoch and I) from Truffle Suite can understand if you have strong reservations against that and will try to make do with that decision.

Again, I appreciate the time and consideration by the Electron maintainers, and you particularly @nornagon!

seesemichaelj:
> Do you have any specific 3rd-party native modules in mind that you're trying to build?

Yup! @alexhultman maintains https://github.com/uNetworking/uWebSockets.js which leverages the native library https://github.com/uNetworking/uWebSockets

As far as porting to work with BoringSSL; @alexhultman made it seem like the support would be relatively trivial, but I don't know if we've actually gotten to that point of integration yet.

alexhultman:
We support BoringSSL I use it all the time.

miniak:
@seesemichaelj what's the plan with this PR? there are no updates for almost a year

seesemichaelj:
@miniak As I no longer have a need for https://github.com/uNetworking/uWebSockets.js in Electron, I no longer have an opinion on the state of this PR. I'm going to close it; if others still need this, I welcome them to open their own PR.

alexhultman:
We ship our own statically linked BoringSSL so this is no longer a problem for us.

PR #25103, opened 2020-08-24
feat: Implement electron.DesktopCapturer.SetSkipCursor

CapOM:
#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [ ] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [ ] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).
- [ ] This is **NOT A BREAKING CHANGE**. Breaking changes may not be merged to master until 11-x-y is branched.

#### Release Notes

Notes: Can disable or re-enable cursor capturer while the stream is running

PR #28302, opened 2021-03-19
feat: Implement desktopCapturer.setSkipCursor

adityak368:
#### Description of Change

feat: Implement electron.DesktopCapturer.setSkipCursor

Useful to dynamically stop cursor capture in the stream.
By default the cursor is captured.
First argument is a deviceId so one can stop capturing
the cursor for a specific streamed source.

electron#23923

This is essentially a cherrypick of the the previous implementation of @CapOM

https://github.com/electron/electron/pull/25103

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Implemented DesktopCapturer.setSkipCursor API. Useful to dynamically stop cursor capture in the stream.


adityak368:
Note: I am not familiar with chromium and electron development. But would like to help as much as I can and get this over the line and merged

nmilev-cursive:
@adityak368 If there's anything you need help with to make this PR successful, let me know!

adityak368:
@nmilev-cursive @CapOM  I have just recently started looking into chromium development. I had to change the patch a bit so that it would hopefully pass the builds and work properly. Would be nice if you could have a look at the changes made (especially in the patch) and let me know if i missed something or could do something better

ns-jisorce:
Try to run the tests locally on your machine, I believe you will be able to reproduce the crashes

lib/browser/rpc-server.ts
```diff
@@ -71,6 +71,18 @@ if (BUILDFLAG(ENABLE_DESKTOP_CAPTURER)) {
 
     return typeUtils.serialize(await desktopCapturer.getSourcesImpl(event, options));
   });
+
+  ipcMainInternal.handle(IPC_MESSAGES.DESKTOP_CAPTURER_SET_SKIP_CURSOR, function (event, sourceId, skipCursor, stack) {
+    logStack(event.sender, 'desktopCapturer.setSkipCursor()', stack);
+    const customEvent = emitCustomEvent(event.sender, 'desktop-capturer-set-skip-cursor');
```
zcbenz, line 77: Can you add documentation for this event?

zcbenz (commented):
Can you add some tests? We don't have to test its functionality, it would be helpful to know it does not crash after Chromium upgrades.

lib/browser/rpc-server.ts
adityak368: Sure. Added the documentation for the event. Added basic tests for the api.

adityak368:
@ns-jisorce I only have a Windows machine. I ran the tests locally there. only 1 test failed (minimize restore window) which seems to be unrelated to the change. Also the tests failing in linux and mac builds seem to be unrelated to the change

danielehrhardt:
Tested this on Mac. Looks good.

zcbenz (approved):

zcbenz:
This PR requires review from @electron/wg-upgrades, and requires API review from @electron/wg-api .

ns-jisorce:
The change looks good to me, if you can restore the author on the first commit that would be nice. (in principle it would have been preserved if you did git cherry-pick)
```
git commit --amend --no-edit --author="Julien Isorce <julien.isorce@gmail.com>"
```

Thanks!

nornagon (changes requested):
This is a significant sized patch with no champion in the maintainers group, and no functional tests. It is thus very likely to cause issues when upgrading in future.

There are a few possible paths forward for this to get merged. The first and best is to see if there is a way to achieve the same result without requiring a patch. I am not personally familiar with the code, but I think this is worth investigating. 

If there really is no way to achieve this functionality without a patch, the next step would be to see if we can drastically reduce the size of the patch. Smaller patches are easier to maintain, because touching less code means the patch is less likely to conflict when upgrading.

If the patch size cannot be significantly reduced, the next thing to try would be to submit these patches upstream. See the [chromium contributing guidelines](https://chromium.googlesource.com/chromium/src/+/master/docs/contributing.md) and [WebRTC contributing guidelines](https://webrtc.org/support/contributing/) for how to get started with that.

If you try all the above and do not meet success, the final option is to significantly expand documentation and test coverage for these patches. The goal here would be that if a maintainer has to handle fixing this patch, then the tests should be able to report whether or not the feature is still working as intended, and the maintainer should be able to clearly understand the intention of each part of the patch, in order to be able to correctly resolve conflicts.

Thanks for your patience with this, I know it's a lot of work. Our maintainer team is small and we have to carefully consider the trade-offs of adding patches to the codebase. Every patch adds to the ongoing maintenance burden of keeping Electron up-to-date, and we have to ensure that the project remains sustainable.

adityak368:
The topic of upstreaming this was already discussed before in the issue and we concluded that patching electron was better. Not sure if something has changed since then. The webrtc part of the patch is pretty straightforward in my opinion. The chromium part is maybe a bit more involved. Other changes are basically for the electron API which is also pretty straightforward. 

The docs for the API and its functionality are already present. Basic tests are written to know that it does not crash. Please let me know what other documentation you have in mind. Also I am not sure how we can write tests to check the functionality as the cursor is pretty much embedded in the video stream. Any suggestions and help to achieve that is welcome. 

@ns-jisorce What are your thoughts? @nornagon Please suggest how to move forward

nornagon:
@adityak368 surely, patching Electron is easier for _you_, but it is not easier for the maintainers. For context, we currently have 110 patches against Chromium, and if merged this patch would be the 6th-largest patch. Chromium constantly changes, and we regularly have to update our patches to keep pace. The more and larger patches we have, the larger our ongoing maintenance burden.

I've outlined several possible paths for moving forward already. The PR adds events which are never tested, nor is calling the method while a capture is occurring. Calling the method from the renderer is not tested either. The documentation this PR adds is pretty thin. There is no example code, it's not documented whether the events are emitted when calling `setSkipCursor` from the main process or only from the renderer, and there is only the barest explanation of what the feature does. We don't need an essay of docs for the feature, as it's fairly straightforward, but it could be better. If a maintainer with no context about what this feature is intended to do is expected to fix a 350-line patch in order to keep the feature working, they had better be able to understand what "working" means.

lib/browser/desktop-capturer.ts
```diff
@@ -80,3 +80,11 @@ export const getSourcesImpl = (event: Electron.IpcMainEvent | null, args: Electr
 
   return getSources;
 };
+
+export const setSkipCursorImpl = (event: Electron.IpcMainEvent | null, sourceId: string, skipCursor: boolean) => {
+  const capturer: ElectronInternal.DesktopCapturer | null = createDesktopCapturer();
```
nornagon, line 85: This is a memory leak. `createDesktopCapturer()` returns an object which is not released until capturing has finished.

shell/browser/api/electron_api_desktop_capturer.cc
```diff
@@ -122,6 +125,11 @@ void DesktopCapturer::OnSourceUnchanged(DesktopMediaList* list) {
   UpdateSourcesList(list);
 }
 
+void DesktopCapturer::SetSkipCursor(const std::string& device_id,
+                                    bool skip_cursor) {
+  content::MediaDeviceRequest::SetSkipCursor(device_id, skip_cursor);
```
nornagon, line 130: since this is a static method, which does not need a `DesktopCapturer` object for context, it should be exposed directly on the bindings object in `Initialize`.

lastthyme (approved):

lastthyme (approved):

shell/browser/api/electron_api_desktop_capturer.cc
adityak368: Done. Removed the method from DesktopCapturer and moved it to Initialize

lib/browser/desktop-capturer.ts
adityak368: Not creating a desktopcapturer anymore. calling the static method directly

adityak368:
@nornagon Addressed review comments. Removed the events. Added more documentation. Added sample code. Added test to call setskipcursor during the capture.

adityak368:
@nornagon Please let me know if there everything is fine or if there are any other changes required.

hackers365:
up

danielehrhardt (approved):

michalzaq12:
up

danielehrhardt:
Why did you closed the Pull Request?

PR #28303, opened 2021-03-19
chore: create script to add frontmatter to docs

molant:
This is part of the efforts to improve [Electron's documentation and reorganize the content](https://github.com/electron/electronjs.org/issues/52320).

**This is a draft PR and should not be merged.**

As part of the reorganization, @erickzhao and I come up with the idea of adding frontmatter to the documentation files. This will allow us to have better search and SEO results. On top of that, it will allow us to move things around more easily until we find the right content structure.
Because there are over 200 md files under `/docs`, I've created this script to make it easier to add it. It takes the first (sub)heading as the title, and the first paragraph as the description (after doing some cleanup and ignoring some patterns).

**How to test it?**

Run `node script/generate-frontmatter.js` and all md files under `/docs` will get updated.

**TODO:**

* [ ] find out how frontmatter affects the autogenerated API docs (@MarshallOfSound can you point me in the right direction?)
* [ ] identify other possible problems when adding frontmatter (downstream in `i18n` with crowdin and `electronjs.org`)
* [ ] add frontmatter for real and fix the empty descriptions of some files


MarshallOfSound (commented):
I don't _think_ this will have a negative effect on the docs parser, the easiest way to tell though is to run it and then compare the generated `electron.d.ts` and `electron-api.json` files.

I do think we shouldn't run this in-source though as it just results in duplicated information, is there any reason we can't run this as a pre-translation step in the `electron-i18n` or `electron-docs` modules so that the duplication doesn't ever actually live in source?

molant:
For a bit more context, my goal with this script is to help the initial step of adding frontmatter to the docs and I do think that the "add" part should only be run once. I haven't added the frontmatter to this PR (yet) to make it easier to review initially, but my plan is to review all titles and descriptions before committing.

> I do think we shouldn't run this in-source though as it just results in duplicated information, is there any reason we can't run this as a pre-translation step in the `electron-i18n` or `electron-docs` modules so that the duplication doesn't ever actually live in source?

When you talk about duplication, you mean in the API category, right?
While the title extraction is mostly accurate for all files, the description isn't. There are some guides and the structure API files that do not have one or the resulting text isn't great and can't be automated. I think that frontmatter should be added the moment a new markdown file is created and by the author of that file, wherever the source repo is, to catch any possible issues as soon as possible. While there might be duplication in some parts of the API, I think the pros outweigh the cons.

miniak:
@molant what's the plan with this PR. It seems abandoned, there are no updates for the last few months.

molant:
This is no longer needed after the changes done in the website and docs overall 😊

Issue #20248, opened 2019-09-16
Cannot use service worker in html document served from custom protocol: The document is in an invalid state.


mjbvz:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** 
  * 7.0.0-beta.4
* **Operating System:** 
  * macOS 10.14.6
* **Last Known Working Electron version:** 
  * NA

### Expected Behavior
Serve up a simple html page inside an iframe using a custom protocol (`test-scheme` in this example). The html page should be able to register a service worker to handle requests

Here's the example main process code:

```js
protocol.registerSchemesAsPrivileged([
  { scheme: 'test-scheme', privileges: { allowServiceWorkers: true, standard: true, secure: true } }
]);
```

```js
 protocol.registerStringProtocol('test-scheme, (request, cb) => {
    cb({
      mimeType: 'text/html',
      charset: 'utf-8',
      data: `<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8">
    <title>Heey World!</title>
  </head>
  <body>
    Hello iframe!
    <script>
      window.navigator.serviceWorker.register('service-worker.js', {scope: './'}).then(function(registration) {
        console.log('All good')
      })
    </script>
  </body>
</html>`
    })
  });
```

### Actual Behavior
The service worker cannot be registered. In the dev tools console, I see the error: 

```
DOMException: Failed to register a ServiceWorker: The document is in an invalid state.
```

### To Reproduce
Use the electron fiddle: https://gist.github.com/98ead2337cf17febffa64f731a8cc3e8

Look at the devtools output to see the errors

### Additional Information

/cc @deepak1556 - This is blocks migrating VS Code to use normal iframes for its webviews

cawa-93:
I encountered the same problem

* **Electron Version:**
  * 8.0.0-beta.5
* **Operating System:**
  * macOS 10.15.1
* **Last Known Working Electron version:**
  * N/A

I use custom protocol:
```js
protocol.registerSchemesAsPrivileged([
  {
    scheme: 'app',
    privileges: {
      secure: true,
      standard: true,
[...]

PR #28326, opened 2021-03-22
fix: make sure service worker scheme is registered with allowServiceWorkers

zcbenz:
#### Description of Change

Fix https://github.com/electron/electron/issues/20248.

The bug was caused by 2 reasons:
1. ServiceWorker loader does not have WebContents associated, so it did not get custom protocols registered.
2. `registerSchemesAsPrivileged` is called after service worker schemes get registered, so the custom scheme could not be recognized as service worker scheme.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix service worker not working with custom protocol.

### DIFF
--- a/shell/app/electron_content_client.cc
+++ b/shell/app/electron_content_client.cc
@@ -187,9 +187,13 @@ void ElectronContentClient::AddAdditionalSchemes(Schemes* schemes) {
   auto* command_line = base::CommandLine::ForCurrentProcess();
   std::string process_type =
       command_line->GetSwitchValueASCII(::switches::kProcessType);
-  // Register schemes for browser process and network utility process, the
-  // registration for renderer process happens in `RendererClientBase`.
-  if (process_type.empty() || process_type == ::switches::kUtilityProcess) {
+  // Browser Process registration happens in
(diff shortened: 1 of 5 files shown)


codebytere (approved):

jkleinsc (approved):

deepak1556:
@zcbenz https://github.com/electron/electron/blob/master/patches/chromium/delay_lock_the_protocol_scheme_registry.patch should let us register schemes on the browser process from `registerSchemesAsPrivileged` . But didn't realize there was no url_util call for registering service worker schemes and savable schemes so effectively we were only registering it on network and renderer process https://source.chromium.org/chromium/chromium/src/+/master:content/common/url_schemes.cc;l=112-118.

So we can add a method `url::AddServiceWorkerScheme` that would allow us to register the scheme from browser process via the `registerSchemesAsPrivileged` method and we can keep the condition in `AddAdditionalSchemes` to only handle network process, this way we don't have to use the testing api.

zcbenz:
@deepak1556 Thanks for the insights on this issue, I have changed the code to add service worker scheme directly instead of full re-registration.

deepak1556 (approved):
👍

PR #28328, opened 2021-03-22
fix: window.print() in pdf plugin

codebytere:
#### Description of Change

Fixes an issue where `window.print()` did not work properly when printing a pdf from the pdf plugin. Pulled from Chromium's [corresponding impl](https://source.chromium.org/chromium/chromium/src/+/master:chrome/renderer/printing/chrome_print_render_frame_helper_delegate.cc;l=64-81;drc=ce83ee3f482ef49409bf97c28dd28cd8da8072c7;bpv=1;bpt=1). 

Tested with https://gist.github.com/293b0a64d8564da954fc222b9aba9ce1.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where `window.print()` did not work properly when printing a pdf from the pdf plugin.

### DIFF
--- a/lib/browser/api/web-contents.ts
+++ b/lib/browser/api/web-contents.ts
@@ -94,6 +94,7 @@ const defaultPrintingSetting = {
   pagesPerSheet: 1,
   isFirstRequest: false,
   previewUIID: 0,
+  // True, if the document source is modifiable. e.g. HTML and not PDF.
   previewModifiable: true,
   printToPDF: true,
   deviceName: 'Save as PDF',
--- a/shell/renderer/printing/print_render_frame_helper_delegate.cc
+++ b/shell/renderer/printing/print_render_frame_helper_delegate.cc
@@ -11,6 +11,7 @@
 
 #if BUILDFLAG(ENABLE_EXTENSIONS)
 #include "extensions/common/constants.h"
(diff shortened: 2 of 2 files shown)


lib/browser/api/web-contents.ts
```diff
@@ -94,7 +94,8 @@ const defaultPrintingSetting = {
   pagesPerSheet: 1,
   isFirstRequest: false,
   previewUIID: 0,
-  previewModifiable: true,
+  // True, if the document source is modifiable. e.g. HTML and not PDF.
```
codebytere, line 97: tangential fix - not directly related to the problem at hand.

nornagon (changes requested):
can we have an automated test? :)

codebytere:
@nornagon what would you envision there? the issue with printing tests generally is that they create blocking dialogs we then can't get rid of - it's why we don't test `window.print()` in other contexts either

nornagon:
ah, i didn't realise there was a dialog involved, that's a shame. :okay:

nornagon (approved):

deepak1556 (approved):

MarshallOfSound (approved):

jkleinsc (approved):

PR #28332, opened 2021-03-22
fix: isolate Pepper plugins

### LINKED ISSUE #28200: [Bug]: Crash when Using mpv.js on Latest Version of Electron
### Preflight Checklist

### Electron Version

12.0.1

### What operating system are you using?

Windows

### Operating System Version

Windows 10 Pro 2020

### What arch are you using?

x64

### Last Known Working Electron version

11.3.0

### Expected Behavior

The app works perfectly fine with an embedded video player.

### Actual Behavior

The app crashes when creating an element with an embedded video player.

### Testcase Gist URL

https://github.com/pavladan/mpv.js-on-react

Change the version of electron to 12.0.1 to see the crash, 11.3.0 works perfectly fine.

Note: In order to see the crash on 12.0.1, adjustments to the settings must be made
```
    width: 1280,
    height: 574,
    autoHideMenuBar: true,
    useContentSize: process.platform !== "linux",
    title: "mpv.js example player",
    webPreferences: {
      nodeIntegration: true,
      plugins: true,
      enableRemoteModule: true,
      nodeIntegrationInSubFrames: true,
      contextIsolation: false,
      worldSafeExecuteJavaScript: false
    },
  });
```

clavin:
#### Description of Change

[A recent change in Chromium](https://source.chromium.org/chromium/chromium/src/+/37893bf803b256be100b1bd1805abcd069c33140) started isolating (basically) all Pepper plugins. It seems like the smart thing for us to do is to follow suit and mimic their new behavior in our own `IsOriginIsolatedPepperPlugin` method.

Fixes #28200. See [this deep dive comment](https://github.com/electron/electron/issues/28200#issuecomment-801496222) for more context.

CC @nornagon 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Pepper plugins are now isolated from origins.

### DIFF
--- a/shell/renderer/renderer_client_base.cc
+++ b/shell/renderer/renderer_client_base.cc
@@ -361,22 +361,7 @@ bool RendererClientBase::IsPluginHandledExternally(
 
 bool RendererClientBase::IsOriginIsolatedPepperPlugin(
     const base::FilePath& plugin_path) {
-  // Taken from Chrome's own implementation of this same method:
-  // https://source.chromium.org/chromium/chromium/src/+/37893bf803b256be100b1bd1805abcd069c33140:chrome/renderer/chrome_content_renderer_client.cc;l=1341-1363
-  //
-  // NOTE: the logic for NaCl is excluded here as we don't currently support it.
-
-  // Hosting plugins in-process is inherently incompatible with attempting to
-  // process-isolate plugins from different origins.
-  auto* cmdline = base::CommandLine::ForCurrentProcess();
-  if (cmdline->HasSwitch(::switches::kPpapiInProcess)) {
-    // The kPpapiInProcess switch should only be used by tests.
-    CHECK_NE(kPdfPluginPath, plugin_path.value());
-
-    return false;
-  }
-
-  // Isolate all the other plugins (including the PDF plugin + test plugins).
+  // Isolate all Pepper plugins, including the PDF plugin.
   return true;
 }
 
--- a/shell/renderer/renderer_client_base.cc
+++ b/shell/renderer/renderer_client_base.cc
@@ -361,7 +361,10 @@ bool RendererClientBase::IsPluginHandledExternally(
 
 bool RendererClientBase::IsOriginIsolatedPepperPlugin(
     const base::FilePath& plugin_path) {
-  // Taken from chrome's own implementation of this method.
+  // Taken from Chrome's own implementation of this same method:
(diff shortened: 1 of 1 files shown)


shell/renderer/renderer_client_base.cc
```diff
@@ -361,11 +361,20 @@ bool RendererClientBase::IsPluginHandledExternally(
 
 bool RendererClientBase::IsOriginIsolatedPepperPlugin(
     const base::FilePath& plugin_path) {
-#if BUILDFLAG(ENABLE_PDF_VIEWER)
-  return plugin_path.value() == kPdfPluginPath;
-#else
-  return false;
-#endif
+  // Taken from chrome's own implementation of this method.
```
codebytere, line 364: Can you link to it? Helpful if it changes in the future

clavin: Changed to include a link to the method in question in the next commit 👍

shell/renderer/renderer_client_base.cc
```diff
@@ -361,11 +361,23 @@ bool RendererClientBase::IsPluginHandledExternally(
...
+
+  // Hosting plugins in-process is inherently incompatible with attempting to
+  // process-isolate plugins from different origins.
+  auto* cmdline = base::CommandLine::ForCurrentProcess();
+  if (cmdline->HasSwitch(::switches::kPpapiInProcess)) {
+    // The kPpapiInProcess switch should only be used by tests.
+    CHECK_NE(kPdfPluginPath, plugin_path.value());
+
+    return false;
+  }
```
nornagon, line 377: Hm, given (a) we don't really support `--ppapi-in-process` ("The kPpapiInProcess switch should only be used by tests") and (b) this `CHECK` will trigger if you pass that flag with this code here, since we always load the pdf plugin, I think we can actually exclude this whole thing and just have this function `return true` 🤔

clavin: Oh my bad, for some reason I assumed that test behavior would also apply to our own tests, and I did not consider that the PDF plugin always loads therefore always triggering this `CHECK` 🤕

Thank you for the insight 🙇

jkleinsc (approved):

NADESHIKON:
Fantastic work, respect 🔥

zcbenz (approved):

NADESHIKON:
May I ask when will this pull request get merged? I am not able to migrate my project to latest version of electron without this.

codebytere (approved):

nornagon (approved):

Issue #28327, opened 2021-03-22
[Bug]: "Uncaught Error:" added to message of errors thrown in a function defined in preload script


raksooo:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/electron/electron/issues) for a feature request that matches the one I want to file, without success.

### Electron Version

12.0.1

### What operating system are you using?

macOS

### Operating System Version

11.2

### What arch are you using?

x64

### Expected Behavior

`error.message` shouldn't be prefixed with "Uncaught Error:"

### Actual Behavior

`error.message` is prefixed with "Uncaught Error:" when an error is thrown in a function defined in the preload script and caught in the renderer process. This only happens when `contextIsolation` is enabled.

### Testcase Gist URL

https://gist.github.com/f11c77f9bc2cbf98d42a9fcaf477c80a

codebytere:
This is probably related to the fact that Errors aren't serializable at present over the `contextBridge` - we pull the message out and re-construct it in the destination context. I'll see what i can do here.

MarshallOfSound:
@codebytere This was solved specifically for `webFrame.executeJavaScript`.  You could extract that logic into the context bridge methods --> https://github.com/electron/electron/pull/26889/files#diff-678990291ce65586b9e4e509f021d2d699ba5ca164b2a0569c5342c60ab548b4R276-R287

PR #28346, opened 2021-03-23
fix: errors thrown in functions over the `contextBridge`

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/28327.

Fixes an issue where errors thrown in functions passed over the `contextBridge` would be prefixed with `Uncaught Error` so that the error message then incorrectly displayed as `Error: Uncaught Error: test`. This is owed to the fact that errors aren't serializable as-is, so we should be pulling the message off the exception and re-constructing and throwing it in the destination context

Tested with https://gist.github.com/f11c77f9bc2cbf98d42a9fcaf477c80a.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where errors thrown in functions passed over the `contextBridge` could be displayed incorrectly.

### DIFF
--- a/shell/renderer/api/electron_api_context_bridge.cc
+++ b/shell/renderer/api/electron_api_context_bridge.cc
@@ -438,17 +438,22 @@ void ProxyFunctionWrapper(const v8::FunctionCallbackInfo<v8::Value>& info) {
         did_error = true;
         v8::Local<v8::Value> exception = try_catch.Exception();
 
-        v8::MaybeLocal<v8::Value> maybe_message =
-            exception.As<v8::Object>()->Get(
-                func_owning_context,
-                gin::ConvertToV8(args.isolate(), "message"));
+        const char* err_msg =
+            "An unknown exception occurred in the isolated context, an error "
+            "occurred but a valid exception was not thrown.";
 
-        if (!maybe_message.ToLocal(&error_message) ||
-            !error_message->IsString()) {
-          error_message = gin::StringToV8(
-              args.isolate(),
-              "An unknown exception occurred in the isolated context, an error "
-              "occurred but a valid exception was not thrown.");
+        if (!exception->IsNull() && exception->IsObject()) {
+          v8::MaybeLocal<v8::Value> maybe_message =
+              exception.As<v8::Object>()->Get(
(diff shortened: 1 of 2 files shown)


shell/renderer/api/electron_api_context_bridge.cc
```diff
@@ -408,29 +408,39 @@ void ProxyFunctionWrapper(const v8::FunctionCallbackInfo<v8::Value>& info) {
...
+
+        const char* err_msg =
+            "An unknown exception occurred in the isolated context, an error "
+            "occurred but a valid exception was not thrown.";
+
+        if (!exception->IsNull() && exception->IsObject()) {
+          v8::MaybeLocal<v8::Value> maybe_message =
+              exception.As<v8::Object>()->Get(
+                  func_owning_context,
+                  gin::ConvertToV8(args.isolate(), "message"));
```
nornagon, line 428: ehhhh, I do not love this, as `Object::Get` can call arbitrary JS and end up throwing again. Can we find a different way to trim the message?

nornagon (changes requested):

shell/renderer/api/electron_api_context_bridge.cc
nornagon: Yeah... I'm also confused by
> errors aren't serializable as-is

Error is definitely serializable 🤔

codebytere: @MarshallOfSound do you know why it was added that way initially?

codebytere: @nornagon this looks pretty similar to what V8 themselves do: https://source.chromium.org/chromium/chromium/src/+/master:v8/src/inspector/value-mirror.cc;l=245-253;drc=effdbe12c7d315055a0a4ddfea3770f2dc2cf84d?q=%22value-%3EIsSymbol()%22&ss=chromium

nornagon: huh, TIL. That's in inspector code, so not really prod stuff, but it's good enough for me.

nornagon (approved):

Issue #28319, opened 2021-03-21
[Bug]: Thumbar disappears after win.hide() is called - and can't be displayed again afterwards


Araxeus:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/electron/electron/issues) for a feature request that matches the one I want to file, without success.

### Electron Version

11.2.3

### What operating system are you using?

Windows

### Operating System Version

Windows 10 v19042

### What arch are you using?

x64

### Expected Behavior

After calling setThumbarButtons() and then `window.hide()`  and then `window.show() ` 
I should be able to set thumbar buttons back to their original state (or create new)

### Actual Behavior

Thumbar function seems to break completely after `window.hide()` is called
(It wont display anymore, calling `setThumbarButtons([...])` returns true but wont actually display anything
⚠️ The bug can also be replicated by just creating the thumbar while the window is already hidden

### Testcase Gist URL

https://gist.github.com/883fa102da7c81e47bb866aa20089292

 -----

#12240 #22081 seems to have the same issue on older Electron versions

I encountered this bug while working on [youtube-music](https://github.com/th-ch/youtube-music/pull/200/files)

[EDIT]: I made the whole thing work by using 
  `win.minimize(); win.setSkipTaskbar(true);`
  instead of
  `window.hide();`
  but still, the original bug exists

ckerr:
Thanks for reporting this and helping to make Electron better!

Would it be possible for you to make a standalone testcase with only the code necessary to reproduce the issue? For example, [Electron Fiddle](https://www.electronjs.org/fiddle) is a great tool for making small test cases and makes it easy to publish your test case to a [gist](https://gist.github.com) that Electron maintainers can use.

Stand-alone test cases make fixing issues go more smoothly: it ensure everyone's looking at the same issue, it removes all unnecessary variables from the equation, and it can also provide the basis for automated regression tests.

I'm adding the `blocked/need-repro` label for this reason. After you make a test case, please link to it in a followup comment.

Thanks in advance! Your help is appreciated.

Araxeus:
[...]

PR #28366, opened 2021-03-24
fix: disappearing thumbar after win.hide()

codebytere:
#### Description of Change

Fixes https://github.com/electron/electron/issues/28319.

When the window is removed from the taskbar via `win.hide()`, the thumbnail buttons need to be set up again. Ensure that when the window is hidden, the taskbar host is notified that it should re-add them.

Tested with https://gist.github.com/883fa102da7c81e47bb866aa20089292.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where the thumbar disappeared after `win.hide()` on Windows.

### DIFF
--- a/shell/browser/native_window_views.cc
+++ b/shell/browser/native_window_views.cc
@@ -444,6 +444,10 @@ void NativeWindowViews::Hide() {
 #endif
 
 #if defined(OS_WIN)
+  // When the window is removed from the taskbar via win.hide(),
+  // the thumbnail buttons need to be set up again.
+  // Ensure that when the window is hidden,
+  // the taskbar host is notified that it should re-add them.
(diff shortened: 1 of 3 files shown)


shell/browser/native_window_views.cc
```diff
@@ -442,6 +442,10 @@ void NativeWindowViews::Hide() {
   if (!features::IsUsingOzonePlatform() && global_menu_bar_)
     global_menu_bar_->OnWindowUnmapped();
 #endif
+
+#if defined(OS_WIN)
+  taskbar_host_.SetThumbarButtonsAdded(false);
```
zcbenz, line 447: Can you add comment for this line? Without context it is hard to understand why this call is needed.

zcbenz (approved):

PR #28367, opened 2021-03-24
Update file-filter for better understanding

AbraaoFarah:
Adding code example for file-filter for better understanding of functionality for those who only read the documentation.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:

### DIFF
--- a/docs/api/structures/file-filter.md
+++ b/docs/api/structures/file-filter.md
@@ -1,4 +1,14 @@
 # FileFilter Object
 
 * `name` String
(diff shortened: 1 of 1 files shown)


MarshallOfSound (changes requested):
This example isn't accurate (you don't need the `.`) and there are already multiple examples of this in the `dialog.md` docs file.  If there's missing information it should probably go there, this file is almost purely a type definition file

PR #28369, opened 2021-03-24
refactor: remove more uses of `v8::Isolate::GetCurrent()`

codebytere:
#### Description of Change

`v8::Isolate::GetCurrent()` is deprecated, and we should be using our helper method where possible.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/api/electron_api_crash_reporter.cc
+++ b/shell/browser/api/electron_api_crash_reporter.cc
@@ -190,8 +190,9 @@ namespace {
 
 #if defined(MAS_BUILD)
 void GetUploadedReports(
+    v8::Isolate* isolate,
     base::OnceCallback<void(v8::Local<v8::Value>)> callback) {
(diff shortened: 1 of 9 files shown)


nornagon (approved):

zcbenz (changes requested):
`JavascriptEnvironment` only lives in the browser process, we can't use it in `shell/common` or `shell/renderer` files. 

There is a similar `blink::MainThreadIsolate()` API available in renderer, but since the code in renderer may run in workers with different isolates, I think the only safe way to replace `v8::Isolate::GetCurrent()` is to pass isolate explicitly.

codebytere:
@zcbenz i think this should do it - all changes I made here are either invoked in the main process or I now pass in the isolate explicitly

zcbenz (approved):

codebytere:
Test failures not germane.

Issue #27722, opened 2021-02-12
will-prevent-unload doesn't fire for BrowserView


pushkin-:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** 11.2.1
* **Operating System:** Windows 10

### Expected Behavior
I expect that `will-prevent-unload` will fire when closing a BrowserView with a page that cancels the `beforeunload` event

### Actual Behavior
`will-prevent-unload` never fires. It only fires for BrowserWindows.

### To Reproduce
1. start main.js
2. in the devtools, execute: `onbeforeunload = (e) => { e.preventDefault(); e.returnValue = false; }`
3. Close the window.
4. The window does close, but the `will-prevent-unload` handler never executes

```
const { app, BrowserWindow, BrowserView } = require('electron')

async function createWindow () {
  const win = new BrowserWindow();
  const view = new BrowserView();

  view.setBounds({x: 0, y: 0, width: 600, height: 600});
  win.setBrowserView(view);

  await view.webContents.loadURL('https://google.com')
  view.webContents.openDevTools();

  view.webContents.on("will-prevent-unload", (e) => {
    e.preventDefault();
  });
}

app.whenReady().then(createWindow)

app.on('window-all-closed', () => {
  if (process.platform !== 'darwin') {
    app.quit()
  }
});
```

Related (if not a dupe) of [this issue](https://github.com/electron/electron/issues/19194)

pushkin-:
I'll add that it seems to fire when navigating the iframe, but not when closing the page with the iframe.

Issue #20746, opened 2019-10-25
beforeunload and unload do not fire when closing a window with a browser view


pushkin-:
### Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:** 5.0.7
* **Operating System:** Windows 10

### Expected Behavior
When I navigate a `BrowserView` to a page that has `beforeunload` and `unload` handlers, I expect that closing the window will cause them to fire.

### Actual Behavior
They do not fire when closing the window.

Notably, if I don't use `BrowserView` and just navigate the `BrowserWindow` to the page, those handlers will execute.

### To Reproduce

`npm start` the code below (in main.js). In the devtools that opens up, run `window.onbeforeunload = () => true;`, then close the window. In the working case, the window will not close. In the broken case that uses `BrowserView`, it will.

```
const { app, BrowserWindow, BrowserView } = require('electron')

function brokenCase() {
	const mainWindow = new BrowserWindow();
	const view = new BrowserView();
	mainWindow.setBrowserView(view);
	view.webContents.loadURL("https://google.com");
	view.webContents.openDevTools();
}

function workingCase() {
	const mainWindow = new BrowserWindow();
	mainWindow.webContents.loadURL("https://google.com");
	mainWindow.webContents.openDevTools();
}

app.on('ready', brokenCase)
```

Maybe related to [this comment](https://github.com/electron/electron/issues/19194#issuecomment-544407723)

pushkin-:
very similar to [this](https://github.com/electron/electron/issues/14020). I notice that browserview's have a destroy method, but not close method. Should we just implement one that triggers beforeunload and unload, or should we expect that closing the window that contains the browser view to trigger those events as well?

pushkin-:
This is also an issue in Electron 7.1.2

pushkin-:
Still a problem in Electron 10.1.2

Now, I also get a "An unhandled win32 exception occurred in electron.exe" when closing the window

pushkin-:
When fixing this, make sure that beforeunload fires for in iframes as well (see [this issue](https://github.com/electron/electron/issues/27716))

PR #28382, opened 2021-03-25
fix: beforeunload and unload firing in BrowserViews

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/20746.
Closes https://github.com/electron/electron/issues/27722.

Fixes an issue where `beforeunload` and `unload` weren't firing at all in `BrowserView`s. This was happening because we were only checking the BrowserWindow's `webContents` to see whether or not the events needed to be fired, and we also needed to check the `BrowserView`s WebContents.

However, this change makes it such that the event will be emitted for `BrowserViews` but will _not_ be respected - I've chosen not to tie the `BrowserView` lifecycle to its owning BrowserWindow should one exist per the [specification](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event). Either way this is undefined behavior, and the alternative would be to wait until _all_ associated `BrowserViews` for a `BrowserWindow` had their `webContents` destroyed, and given that a `BrowserView` doesn't _need_ to be associated with a `BrowserWindow` this would ultimately be more confusing for end users.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed an issue where `beforeunload` and `unload` weren't firing properly in `BrowserView`s.

### DIFF
--- a/docs/api/web-contents.md
+++ b/docs/api/web-contents.md
@@ -375,6 +375,8 @@ win.webContents.on('will-prevent-unload', (event) => {
 })
 ```
 
+**Note:** This will be emitted for `BrowserViews` but will _not_ be respected - this is because we have chosen not to tie the `BrowserView` lifecycle to its owning BrowserWindow should one exist per the [specification](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event).
+
 #### Event: 'crashed' _Deprecated_
 
 Returns:
--- a/shell/browser/api/electron_api_browser_window.cc
+++ b/shell/browser/api/electron_api_browser_window.cc
@@ -238,35 +238,35 @@ void BrowserWindow::OnCloseButtonClicked(bool* prevent_default) {
   if (window_unresponsive_closure_.IsCancelled())
     ScheduleUnresponsiveEvent(5000);
 
-  if (!web_contents()) {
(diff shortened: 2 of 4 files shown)


docs/api/web-contents.md
```diff
@@ -375,6 +375,8 @@ win.webContents.on('will-prevent-unload', (event) => {
 })
 ```
 
+**Note:** This will be emitted for `BrowserViews` but will _not_ be respected - this is because we have chosen not to tie the `BrowserView` lifecycle to its owning BrowserWindow should one exist per the [specification](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event).
```
nornagon, line 378: what about this event won't be respected? preventing the default?

codebytere: @nornagon yes - the function body will be run but the return value won't be respected

zcbenz (approved):

docs/api/web-contents.md
nornagon: Er, sorry for not getting back to this. I think this is unclear in the docs as written and should be clarified.

pushkin-:
@codebytere Hi - do you have an idea of when this could be backported to 11?

PR #28393, opened 2021-03-25
fix: add an accessible name and role to the app menu bar

mzajaczkowski:
#### Description of Change

The Electron application menu bar does not have an accessible name nor role, the change adds them.

#### Checklist

- [x] `npm test` passes (same failure after "navigator.serial does not return a port when permission denied" for a build with and w/o the change)

#### Release Notes

Notes: The electron menu bar has proper accessibility attributes


BUILD.gn
```diff
@@ -368,6 +369,8 @@ source_set("electron_lib") {
     "//third_party/webrtc_overrides:webrtc_component",
     "//third_party/widevine/cdm:headers",
     "//third_party/zlib/google:zip",
+    "//ui/accessibility",
+    "//ui/base",
```
codebytere, line 373: We probably don't want to take on these new dependencies wholesale, as we should just pull out what we need here for a change of this relatively small size

codebytere (changes requested):
Thanks - one issue but this is on the right track!

BUILD.gn
```diff
@@ -331,6 +331,7 @@ source_set("electron_lib") {
     "//components/network_session_configurator/common",
     "//components/pref_registry",
     "//components/prefs",
+    "//components/strings:components_strings_grit",
```
MarshallOfSound, line 334: I think these should be added to `electron_paks`, not directly here

mzajaczkowski: In Chromium this target is a dep pf targets that include "components/strings/grit/components_strings.h" (example: //components/printing/browser) and it seems corrects as the header is an output of that target:
`grit("components_strings") {
  source = "../components_strings.grd"
  defines = [
    "enable_arcore=$enable_arcore",
    "enable_plugins=$enable_plugins",
    "enable_print_preview=$enable_print_preview",
    "enable_vr=$enable_vr",
  ]

  outputs = [ "grit/components_strings.h" ]
  foreach(locale, locales_with_pseudolocales) {
    outputs += [ "components_strings_$locale.pak" ]
  }

  if (is_android) {
    outputs += android_components_strings_java_resources
  }
}`

At the same time "//components/strings:components_strings" is a dep of repack_locales("${target_name}_locales") in //electron/electron_packs.gni, so I believe we're on the safe side here.

BUILD.gn
mzajaczkowski: Builds fine for me, pushed a fixup.

zcbenz (approved):
Can you rebase this PR?

miniak:
@mzajaczkowski are you still planning to land this PR?

MarshallOfSound:
Closing this PR as abandoned, if someone wants to pick up this work please feel free to raise a new PR for review

PR #28396, opened 2021-03-25
docs: backport current tutorials to 12-x-y

molant:
#### Description of Change

This brings all the documentation changes to the guides that are missing in the `12-x-y` release in preparation to start using the documentation from the right branch.

I tried to bring the right commits but there were more than 100 and it was impossible to bring them effectively. Instead I've copy pasted the `tutorials` folder and verified that the contents are relevant to v12.

I guess that @erickzhao would be the right stakeholder but there might be others as well.

There will be another PR to do the same for v13.

#### Checklist

#### Release Notes

Notes: none


ckerr (approved):
All these changes look reasonable to me. None of them are version-specific and are reasonable for any supported branch.

docs/tutorial/using-selenium-and-webdriver.md
```diff
@@ -86,12 +86,12 @@ const driver = new webdriver.Builder()
   // The "9515" is the port opened by chrome driver.
   .usingServer('http://localhost:9515')
   .withCapabilities({
-    chromeOptions: {
+    'goog:chromeOptions': {
```
ckerr, line 89: Same question here as for 13: was `goog:chromeOptions` the right property name all along, or did it change? And if it changed, is `goog.chromeOptions` the correct key for Electron 12?

erickzhao: I just checked the history for this doc and https://github.com/electron/electron/pull/28231 specified that this was required for the latest versions of `selenium-webdriver`, and that the new property name still worked with older versions.

zcbenz (approved):

zcbenz:
The lint CI check is somehow not working, which blocks merging the PR. @molant Can you do a rebase or amend to force restarting the CI?

molant:
@zcbenz I've rebased, hopefully this time works.
Thanks 🙏

molant:
@zcbenz I believe the failing tests are unrelated.
Thanks!

PR #28431, opened 2021-03-29
Update native-file-drag-drop.md

AlGantori:
Missing information.  Incomplete example.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes:


codebytere (changes requested):
@AlGantori this looks like you'd like to ask a question - this does not add any new information to the tutorial. Please elaborate on your goals here, as this is not currently something we would be able to merge.

zcbenz:
I'm closing this PR since it intends to ask a question instead of making a change. Please use issues for this purpose, otherwise we won't be able to track it.

PR #28432, opened 2021-03-29
fix: handle an unparsable pdf manifest

MarshallOfSound:
This aligns our component extension loading logic with https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/extensions/component_loader.cc;l=215-216;drc=062c315a858a87f834e16a144c2c8e9591af2beb

Notes: Fixes rare crash when initializing the internal PDF extension


nornagon (approved):
better than a segfault i suppose. i wish i knew what caused it to be null though.

codebytere (approved):

PR #28454, opened 2021-03-30
build: fix unicode patch file comparison in git.py

MarshallOfSound:
This caused some patches to fail incorrectly as the patch file included non-ascii characters, we have to manually convert using the utf8 charset.

Appears to have been broken by #28360

Notes: no-notes

### DIFF
--- a/script/lib/git.py
+++ b/script/lib/git.py
@@ -307,17 +307,19 @@ def export_patches(repo, out_dir, patch_range=None, dry_run=False):
     # exported patch differs from what exists. Report number of mismatched
     # patches and fail if there's more than one.
     patch_count = 0
+    bad_patches = []
     for patch in patches:
       filename = get_file_name(patch)
       filepath = posixpath.join(out_dir, filename)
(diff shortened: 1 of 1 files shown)


codebytere (approved):

ckerr (approved):
Thank you. I am hitting the same issue and was just about to put up a PR :smile: 

Also, thanks for adding the log of which files failed

script/lib/git.py
```diff
@@ -307,17 +307,19 @@ def export_patches(repo, out_dir, patch_range=None, dry_run=False):
     # exported patch differs from what exists. Report number of mismatched
     # patches and fail if there's more than one.
     patch_count = 0
+    bad_patches = []
```
ckerr, line 310: _nit_ we don't need `patch_count` anymore; it can be replaced with `len(bad_patches)`

MarshallOfSound:
@ckerr Purely because I'm hitting this again locally and it's annoying me I'm going to merge this but I'll throw up a PR for your nit straight after because you're 100% right 😄

Issue #23315, opened 2020-04-28
Using worker_threads that keep running indefinitely fails assertion on SIGTERM


KishanBagaria:
### Issue Details

* **Electron Version:**
  * v9.0.0-beta.21 and 8.2.4
* **Operating System:**
  * macOS 10.15.4

### Actual Behavior

An electron app that uses a worker thread that runs indefinitely in the main process (because of a server listening on a port etc.) throws the following error on exit:

```
/Users/kishan/Downloads/electron-test/node_modules/electron/dist/Electron.app/Contents/MacOS/Electron[93383]: ../../third_party/electron_node/src/node_worker.cc:358:virtual node::worker::Worker::~Worker(): Assertion `stopped_' failed.
 1: 0x10f4ce9b5 node::Buffer::New(v8::Isolate*, char*, unsigned long, void (*)(char*, void*), void*) [/Users/kishan/Downloads/electron-test/node_modules/electron/dist/Electron.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
 2: 0x10f4ce73f node::Buffer::New(v8::Isolate*, char*, unsigned long, void (*)(char*, void*), void*) [/Users/kishan/Downloads/electron-test/node_modules/electron/dist/Electron.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
 3: 0x10f54eb01 node::MultiIsolatePlatform::CancelPendingDelayedTasks(v8::Isolate*) [/Users/kishan/Downloads/electron-test/node_modules/electron/dist/Electron.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
 4: 0x10f54eb3e node::MultiIsolatePlatform::CancelPendingDelayedTasks(v8::Isolate*) [/Users/kishan/Downloads/electron-test/node_modules/electron/dist/Electron.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
 5: 0x10f47f56f node::EmitAsyncDestroy(node::Environment*, node::async_context) [/Users/kishan/Downloads/electron-test/node_modules/electron/dist/Electron.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
 6: 0x10f44cc4e node::FreeEnvironment(node::Environment*) [/Users/kishan/Downloads/electron-test/node_modules/electron/dist/Electron.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
[...]

PR #28468, opened 2021-03-31
fix: use correct Node.js shutdown order

addaleax:
#### Description of Change

The Environment is created after the Isolate and its platform data,
so it also needs to be torn down before it. This fixes the crash
mentioned below.

(Thanks to @indutny for debugging this!)

Fixes: https://github.com/electron/electron/issues/23315

The crash itself affects versions based on Node.js 12.x, but the order is also wrong on master (diff applies cleanly everywhere I checked), so I opened a PR against master in the hope that it goes into at least Electron 11.x soon as well (or, rather, that’s what @indutny would like to see :))

#### Checklist

(I did none of these things because buildig Electron takes forever and I don’t plan on doing that anytime soon, sorry – still :blue_heart: the Electron team)

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/electron_browser_main_parts.cc
+++ b/shell/browser/electron_browser_main_parts.cc
@@ -523,9 +523,9 @@ void ElectronBrowserMainParts::PostMainMessageLoopRun() {
   // Destroy node platform after all destructors_ are executed, as they may
   // invoke Node/V8 APIs inside them.
   node_env_->env()->set_trace_sync_io(false);
-  js_env_->OnMessageLoopDestroying();
   node::Stop(node_env_->env());
   node_env_.reset();
+  js_env_->OnMessageLoopDestroying();
 
   auto default_context_key = ElectronBrowserContext::PartitionKey("", false);
   std::unique_ptr<ElectronBrowserContext> default_context = std::move(


addaleax:
… what’s wrong with the pull request title?

indutny-signal:
FWIW, the minimal test case for this issue appears to be:

```js
const { app } = require('electron');
const { Worker } = require('worker_threads');

const worker = new Worker('', {
  eval: true,
});
worker.terminate().then(() => {
  app.quit();
});
```

deepak1556:
Thanks for the PR! 

`JavscriptEnvironment::OnMessageLoopDestroying` unregisters isolate only in versions `<=11.x.y`, it also performs the job of running the destruction sequence of some `v8::Global` electron api classes that cannot be delayed after the node environment is destroyed. A better fix would be to leave the order unchanged and backport https://github.com/electron/electron/pull/25249/commits/5c1ccfd0994c422bdb0375eec7d5bd65fd6d596f along with relevant nodejs commit to `11.x.y` branch. Thoughts ?

indutny:
@deepak1556 I agree, and I can see that it is indeed not unregistering isolate in the latest master anymore! Electron 12 is not susceptible to this crash, and the change that you mentioned doesn't have UnregisterIsolate call either. So I'm :+1: for backporting it. Do you need any help with that, or with testing it?

indutny-signal:
...and I can also see the failures in the CI because the change that we've came up is too broad, so in any case it might be best to close this Pull Request once the backport is done.

deepak1556:
Cool, if you already have the backport ready then feel free to open a PR for a faster review cycle otherwise I can get one out shortly after I sync electron-11 branch in my local copy.

Closing this in the meantime :)

indutny-signal:
@deepak1556 it appears to apply cleanly on my checkout. Let me quickly verify that it gets rid of the worker_threads issue, and open a PR!

indutny:
@deepak1556 here you go https://github.com/electron/electron/pull/28471

addaleax:
> `JavscriptEnvironment::OnMessageLoopDestroying` unregisters isolate only in versions `<=11.x.y`,

@deepak1556 Okay, I missed that, but … where does it unregister it now? I can’t find anything in the source code, but not unregistering it generally leads to crashes (apart from leaking memory) when a later Isolate is registered with the platform that happens to be allocated at the same address as a previously registered one (this happens frequently enough to have made our CI flaky in the past).

indutny:
@addaleax good point. I should be able to send a pull request for that in a moment now that I understand the semantics properly.

indutny:
Well, surprise surprise, but it turns out that isolate is unregistered in master and 12-x-y, and the fix was different there: 2132fdfa28c0d72f8183e4aa46f86ff2ccb9da6c . @deepak1556 I'll add isolate unregistering to my PR just to make sure that we call it.

indutny:
@addaleax I'm thinking

```diff
diff --git a/shell/app/node_main.cc b/shell/app/node_main.cc
index 73db4f0e1..ff9c7c373 100644
--- a/shell/app/node_main.cc
+++ b/shell/app/node_main.cc
@@ -279,10 +279,6 @@ int NodeMain(int argc, char* argv[]) {
     node::Stop(env);
     node::FreeEnvironment(env);
     node::FreeIsolateData(isolate_data);
-
-    gin_env.platform()->DrainTasks(isolate);
-    gin_env.platform()->CancelPendingDelayedTasks(isolate);
-    gin_env.platform()->UnregisterIsolate(isolate);
   }
 
   // According to "src/gin/shell/gin_main.cc":
diff --git a/shell/browser/electron_browser_main_parts.cc b/shell/browser/electron_browser_main_parts.cc
index 607f4e0b8..06fb5272a 100644
--- a/shell/browser/electron_browser_main_parts.cc
+++ b/shell/browser/electron_browser_main_parts.cc
@@ -475,6 +475,7 @@ void ElectronBrowserMainParts::PreMainMessageLoopRun() {
 }
 
 bool ElectronBrowserMainParts::MainMessageLoopRun(int* result_code) {
+  DCHECK_NE(js_env_, nullptr);
   js_env_->OnMessageLoopCreated();
   exit_code_ = result_code;
   return content::BrowserMainParts::MainMessageLoopRun(result_code);
@@ -526,6 +527,7 @@ void ElectronBrowserMainParts::PostMainMessageLoopRun() {
   js_env_->OnMessageLoopDestroying();
   node::Stop(node_env_->env());
   node_env_.reset();
+  js_env_.reset();
 
   auto default_context_key = ElectronBrowserContext::PartitionKey("", false);
   std::unique_ptr<ElectronBrowserContext> default_context = std::move(
diff --git a/shell/browser/javascript_environment.cc b/shell/browser/javascript_environment.cc
index 343e40654..1ffd45ffb 100644
--- a/shell/browser/javascript_environment.cc
+++ b/shell/browser/javascript_environment.cc
@@ -167,6 +167,11 @@ JavascriptEnvironment::~JavascriptEnvironment() {
   }
   isolate_->Exit();
   g_isolate = nullptr;
+
+  DCHECK_NE(platform_, nullptr);
+  platform_->DrainTasks(isolate_);
+  platform_->CancelPendingDelayedTasks(isolate_);
+  platform_->UnregisterIsolate(isolate_);
 }
 
 class EnabledStateObserverImpl final
```

What is your opinion on this fix?

indutny:
@deepak1556 FWIW, the patch above together with [these two lines from my PR](https://github.com/electron/electron/commit/5c1ccfd0994c422bdb0375eec7d5bd65fd6d596f#diff-1c79615b27d8bc10f4bbc0349f1d35e3ead065a22f02016f95bbc42b37db05f5L284-L285) is sufficient to fix worker_threads crashes. Maybe we should be doing that instead?

deepak1556:
@addaleax good catch. @indutny here is what I am thinking as next steps,

1) Merge https://github.com/electron/electron/pull/28471 to `11-x-y` which will align the environment shutdown path on all branches

2) If you can create a PR with https://github.com/electron/electron/pull/28468#issuecomment-811989180 against master, it will be backported to all release lines.

> FWIW, the patch above together with these two lines from my PR is sufficient to fix worker_threads crashes. Maybe we should be doing that instead?

Yup, but since we already touching this code path aligning the implementation is better.

indutny:
@deepak1556 I agree. Here's the PR: https://github.com/electron/electron/pull/28486

PR #28469, opened 2021-03-31
fix: free IsolateData in ~NodeEnvironment

addaleax:
#### Description of Change

This seems to just have been missing here, leaking memory
(and breaking the API contract for Node.js embedding).

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/shell/browser/javascript_environment.cc
+++ b/shell/browser/javascript_environment.cc
@@ -363,7 +363,9 @@ void JavascriptEnvironment::OnMessageLoopDestroying() {
 NodeEnvironment::NodeEnvironment(node::Environment* env) : env_(env) {}
 
 NodeEnvironment::~NodeEnvironment() {
+  auto* isolate_data = env_->isolate_data();
   node::FreeEnvironment(env_);
+  node::FreeIsolateData(isolate_data);
 }
 
 }  // namespace electron


nornagon (approved):

shell/browser/javascript_environment.cc
```diff
@@ -360,10 +360,13 @@ void JavascriptEnvironment::OnMessageLoopDestroying() {
   base::CurrentThread::Get()->RemoveTaskObserver(microtasks_runner_.get());
 }
 
-NodeEnvironment::NodeEnvironment(node::Environment* env) : env_(env) {}
+NodeEnvironment::NodeEnvironment(node::Environment* env,
+                                 node::IsolateData* isolate_data)
+    : env_(env), isolate_data_(isolate_data) {}
 
 NodeEnvironment::~NodeEnvironment() {
   node::FreeEnvironment(env_);
```
deepak1556, line 368: Can this instead be

```
auto* isolate_data = env_->isolate_data();
node::FreeEnvironment(env_);
node::FreeIsolateData(isolate_data);
```

Avoids having to change the class definition.

deepak1556 (commented):
Question: Any reason why `FreeEnvironment` doesn't destroy the isolate data ? Embedder creates the isolate data as a raw pointer and passes it to node, expectation was node would handle the lifetime.

deepak1556:
@nornagon do we want to backport this ? I see no harm in doing it, thoughts ?

nornagon:
seems like a good idea to me

shell/browser/javascript_environment.cc
addaleax: Sure, done!

addaleax:
> Question: Any reason why `FreeEnvironment` doesn't destroy the isolate data ? Embedder creates the isolate data as a raw pointer and passes it to node, expectation was node would handle the lifetime.

The point of having IsolateData separate from Environment is that way, Node.js supports multiple Environments per Isolate. That means that Environments cannot own IsolateData.

deepak1556 (approved):

deepak1556:
Linux failures are unrelated, merging.

PR #28474, opened 2021-04-01
docs: Clarify proper net.request() error handling

poiru:
Attaching a `error` handler to the `ClientRequest` is not sufficient. Suggest
adding an `error` handler to the response `IncomingMessage` object as well.

Notes: no-notes

### DIFF
--- a/docs/api/client-request.md
+++ b/docs/api/client-request.md
@@ -135,6 +135,9 @@ Emitted when the `net` module fails to issue a network request. Typically when
 the `request` object emits an `error` event, a `close` event will subsequently
 follow and no response object will be provided.
 
+Note that the response object may also emit an `error`. To avoid uncaught
+exceptions, attach an `error` handler to both `request` and `response`.
+
 #### Event: 'close'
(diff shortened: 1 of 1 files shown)


poiru:
Note that due to

https://github.com/electron/electron/blob/2a55ae4b8504c9d0863ac38435bd4f246aa23711/lib/browser/api/net.ts#L437-L441

... the `response` will also emit an error when the `request` emits an error because `this._response.destroy(error);` ends up here in Node.js:

https://github.com/nodejs/node/blob/3ef95627cc911da8198f093960f4490facdfb8e2/lib/internal/streams/destroy.js#L92-L96

@nornagon Is this intentional? If not, we should change the problematic line to `this._response.destroy();` to (partly) avoid this footgun.

nornagon (changes requested):
I'm not sure this is the right resolution, see also https://github.com/electron/electron/pull/27597

poiru:
Oh! Hope that gets merged, will close this.

PR #28480, opened 2021-04-01
docs: the minimum supported version of macOS is now 10.11

MarshallOfSound:
Notes: no-notes

### DIFF
--- a/docs/tutorial/support.md
+++ b/docs/tutorial/support.md
@@ -94,7 +94,7 @@ Following platforms are supported by Electron:
 ### macOS
 
 Only 64bit binaries are provided for macOS, and the minimum macOS version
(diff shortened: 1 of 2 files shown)


nornagon (changes requested):
IMO this deserves release notes.

MarshallOfSound:
@nornagon Maybe it does, but not in this PR.  The change was made in Electron 12 so if anything those notes should be retroactively updated.  Adding notes in this PR is misleading and incorrect as the fundamental OS limitation was modified a while back

nornagon (approved):

nornagon:
we should probably backport this to 12 then.

codebytere (approved):

Aekewu (approved):

PR #28481, opened 2021-04-01
feat: add `process.getWindowsVersionType()` API

erickzhao:
#### Description of Change

Exposes a new API that returns Chromium's Windows `VersionType` enum.

https://source.chromium.org/chromium/chromium/src/+/master:base/win/windows_version.h;l=57-68;drc=76bb9de1de4e0e52d2d40c2c8d5641f7b20e4b50

cc @MarshallOfSound 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added new `process.getWindowsVersionType()` API.

### DIFF
--- a/docs/api/process.md
+++ b/docs/api/process.md
@@ -218,7 +218,7 @@ that all statistics are reported in Kilobytes.
 
 ### `process.getSystemProductType()`
 
-Returns `String` - Returns `server` if operating system is Windows Server. Otherwise, returns `desktop`.
+Returns `String` - Can be `server` or `desktop`. Returns `server` if operating system is Windows Server. Otherwise, returns `desktop`.
 
 ### `process.getSystemVersion()`
 
--- a/docs/api/process.md
+++ b/docs/api/process.md
@@ -218,7 +218,8 @@ that all statistics are reported in Kilobytes.
 
 ### `process.getSystemProductType()`
 
-Returns `String` - Can be `server` or `desktop`. Returns `server` if operating system is Windows Server. Otherwise, returns `desktop`.
+Returns `String` - Can be `server` or `desktop`.Returns `server` if operating
+system is Windows Server. Otherwise, returns `desktop`.
 
 ### `process.getSystemVersion()`
 
--- a/docs/api/process.md
+++ b/docs/api/process.md
@@ -19,6 +19,7 @@ In sandboxed renderers the `process` object contains only a subset of the APIs:
 * `getBlinkMemoryInfo()`
 * `getProcessMemoryInfo()`
 * `getSystemMemoryInfo()`
+* `getSystemProductType()`
 * `getSystemVersion()`
 * `getCPUUsage()`
 * `getIOCounters()`
@@ -214,6 +215,10 @@ Returns `Object`:
 
 Returns an object giving memory usage statistics about the entire system. Note
 that all statistics are reported in Kilobytes.
+
+### `process.getSystemProductType()`
+
+Returns `String` - Returns `server` if operating system is Windows Server. Otherwise, returns `desktop`.
 
 ### `process.getSystemVersion()`
 
--- a/docs/api/process.md
+++ b/docs/api/process.md
@@ -218,8 +218,8 @@ that all statistics are reported in Kilobytes.
 
 ### `process.getWindowsVersionType()`
 
-Returns `String` - Can be `server` or `desktop`. Returns `server` if operating
-system is Windows Server. Otherwise, returns `desktop`.
+Returns `String` - Returns the type of Windows version running on the host operating system.
+Can be `home`, `professional`, `server`, `enterprise`, `education`, or `educationPro`.
 
 ### `process.getSystemVersion()`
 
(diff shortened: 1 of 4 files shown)


docs/api/process.md
```diff
@@ -215,6 +216,10 @@ Returns `Object`:
 Returns an object giving memory usage statistics about the entire system. Note
 that all statistics are reported in Kilobytes.
 
+### `process.getSystemProductType()`
+
+Returns `String` - Returns `server` if operating system is Windows Server. Otherwise, returns `desktop`.
```
MarshallOfSound, line 221:
```suggestion
Returns `String` - Can be `server` or `desktop`. Returns `server` if operating system is Windows Server. Otherwise, returns `desktop`.
```

This should make the type definitions correct

erickzhao: Thank you! Wasn't sure about the semantics. 😅

nornagon (changes requested):
We should probably consider exposing the whole enum: https://source.chromium.org/chromium/chromium/src/+/master:base/win/windows_version.h;l=57-68;drc=76bb9de1de4e0e52d2d40c2c8d5641f7b20e4b50

I think this could also be named more specifically as it's only useful on Windows. Perhaps `getWindowsVersionType()`? I think it might also make sense to throw on non-Windows platforms.

nornagon:
Also, what's the intended use case for this? What do apps need to differently on Windows Server than other versions of Windows?

MarshallOfSound:
> Also, what's the intended use case for this? What do apps need to differently on Windows Server than other versions of Windows?

Sometimes when stopping support of older OS's you need to be able to distinguish between the same OS version running on different "product types".  i.e. Windows 7 Home vs Server 2008.  It's helpful information to send with autoUpdate requests.

erickzhao:
@MarshallOfSound Is there a usecase to eventually expand this `server` vs `desktop` difference to Linux? In that case, it'd make sense for the API to not be renamed to be Windows-specific.

nornagon:
> @MarshallOfSound Is there a usecase to eventually expand this `server` vs `desktop` difference to Linux? In that case, it'd make sense for the API to not be renamed to be Windows-specific.

I don't think it's a good idea to try to pretend that these things are comparable between linux and windows.

MarshallOfSound:
Agreed, if we need the distinction on Linux it will probably have a different usecase and a vastly more subtle implementation.  Let's make this Windows specific for now

docs/api/process.md
```diff
@@ -215,6 +216,11 @@ Returns `Object`:
 Returns an object giving memory usage statistics about the entire system. Note
 that all statistics are reported in Kilobytes.
 
+### `process.getWindowsVersionType()`
```
nornagon, line 219:
```suggestion
### `process.getWindowsVersionType()` _Windows_
```

shell/common/api/electron_bindings.cc
```diff
@@ -204,6 +207,32 @@ v8::Local<v8::Value> ElectronBindings::GetSystemMemoryInfo(
...
+    v8::Isolate* isolate) {
+  std::string type;
+  std::map<base::win::VersionType, std::string> versionMap{
+      {base::win::SUITE_HOME, "home"},
+      {base::win::SUITE_PROFESSIONAL, "professional"},
+      {base::win::SUITE_SERVER, "server"},
+      {base::win::SUITE_ENTERPRISE, "enterprise"},
+      {base::win::SUITE_EDUCATION, "education"},
+      {base::win::SUITE_EDUCATION_PRO, "educationPro"},
+      {base::win::SUITE_LAST, "invalid"}};
```
nornagon, line 221: I think this would be better expressed with a `switch` statement.

spec/api-process-spec.js
```diff
@@ -73,6 +74,16 @@ describe('process module', () => {
     });
   });
 
+  describe('process.getWindowsVersionType()', () => {
```
nornagon, line 77:
```suggestion
  ifdescribe(process.platform === 'win32')('process.getWindowsVersionType()', () => {
```

& then drop the `if`s on the individual tests

nornagon (commented):
API LGTM, couple of nits in the code.

nornagon (approved):
marking approved to speed things along, but please 👀  my comments before merging!

jkleinsc:
API LGTM

spec/api-process-spec.js
erickzhao: The second test checks if it throws on non-Windows platforms, though. I copied the `ifit` pattern from other tests that tested platform-specific functionality.🤔

jkleinsc (changes requested):
Blocking this until https://github.com/electron/electron/pull/28481/files#r696118830 is resolved.

miniak:
@erickzhao are you still planning to land this PR?

shell/common/api/electron_bindings.cc
```diff
@@ -5,6 +5,7 @@
 #include "shell/common/api/electron_bindings.h"
 
 #include <algorithm>
+#include <map>
```
miniak, line 8: I don't see an `std::map` used anywhere in your code

miniak:
Agreed with @erickzhao to close the PR for now

Issue #26862, opened 2020-12-07
Windows JumpList empty when a single entry has a long description


bpasero:
### Issue Details

* **Electron Version:**
  * 11.0.3
* **Operating System:**
  * Windows 10
* **Last Known Working Electron version:**
  * 

### Expected Behavior
An invalid entry to the `JumpList` does not cause other entries to disappear.

### Actual Behavior
A single invalid `JumpList` entry makes all other entries disappear.

### To Reproduce

```ts
// Modules to control application life and create native browser window
const { app, BrowserWindow } = require('electron')

// Keep a global reference of the window object, if you don't, the window will
// be closed automatically when the JavaScript object is garbage collected.
let mainWindow

function createWindow() {
  // Create the browser window.
  mainWindow = new BrowserWindow({
    width: 800,
    height: 600,
    webPreferences: {
      nodeIntegration: true
    }
  })

  // and load the index.html of the app.
  mainWindow.loadFile('index.html')

  // Open the DevTools.
  // mainWindow.webContents.openDevTools()

  // Emitted when the window is closed.
  mainWindow.on('closed', function () {
    // Dereference the window object, usually you would store windows
    // in an array if your app supports multi windows, this is the time
    // when you should delete the corresponding element.
    mainWindow = null
  })
}

// This method will be called when Electron has finished
// initialization and is ready to create browser windows.
// Some APIs can only be used after this event occurs.
app.on('ready', createWindow)

app.setJumpList([
  {
    type: "custom",
    name: "Recent Workspaces",
    items: [
      {
        type: "task",
        title: "Title",
        description: "Description",
        program: process.execPath
      },
      {
        type: "task",
        title: "Long Title",
        description: "DescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescriptionDescription",
        program: process.execPath
      }
    ]
  }
])

// Quit when all windows are closed.
app.on('window-all-closed', function () {
  // On OS X it is common for applications and their menu bar
  // to stay active until the user quits explicitly with Cmd + Q
  if (process.platform !== 'darwin') {
    app.quit()
  }
})

app.on('activate', function () {
[...]

PR #28485, opened 2021-04-01
fix: reject task append to JumpList when description exceeds 260 characters

SushiJackal:
#### Description of Change

Fixed: #26862

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Fixed a bug where, when a JumpList task description exceeded 260 characters, the JumpList was empty, despite valid entries.


shell/browser/ui/win/jump_list.cc
```diff
@@ -28,6 +28,9 @@ bool AppendTask(const JumpListItem& item, IObjectCollection* collection) {
       FAILED(link->SetDescription(item.description.c_str())))
     return false;
 
+  if (wcslen(item.description.c_str()) > 260)
```
nornagon, line 31: Why 260? is this a platform constant?

Also, could we use `.size()` instead of `wcslen`?
```suggestion
  if (item.description.size() > 260)
```

nornagon (changes requested):
Is a silent failure the right approach here? Should we perhaps throw an error or truncate the content instead?

Also, this limit should be documented somewhere.

shell/browser/ui/win/jump_list.cc
SushiJackal: 260 seems to be a Windows 10 constant. I had to test lengths to find this limit, since I couldn't find it listed anywhere, except for in the [SetDescription](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ishelllinka-setdescription) documentation. Here it is mentioned as "`INFOTIPSIZE`", though even if the description exceeds 260 characters, `SetDescription` returns `S_OK`, so a success.

I just ran a test with `size()` and it seems to work as well. Why is this preferred? Should I make a new commit with this change included?

SushiJackal:
@nornagon the error is not silenced, just not handled by this part of the script. Within the `AppendCategory` function in jump_list.cc, there is an error output on rejection of tast append, `[12720:0402/020411.590:ERROR:jump_list.cc(242)] Failed to append task 'Long Title' to Jump List.`

Perhaps it would be better to serve additional information within the rejection itself though, along the lines of "The description of <task name> exceeds the maximum character limit of 260!"

shell/browser/ui/win/jump_list.cc
nornagon: @SushiJackal it's preferred because `std::wstring` already has the size of the string stored, so `.size()` is O(1), whereas `wcslen()` is O(n). It's also easier to read :) Please do go ahead and update this PR to use `.size()`.

SushiJackal: Ooh, alright! Thanks for clearing that up for me. I'll commit right away

shell/browser/ui/win/jump_list.cc
```diff
@@ -28,6 +28,9 @@ bool AppendTask(const JumpListItem& item, IObjectCollection* collection) {
       FAILED(link->SetDescription(item.description.c_str())))
     return false;
 
+  if (item.description.size() > 260)
```
nornagon, line 31: Maybe add a comment here as well, something like
```suggestion
  // The documentation for SetDescription suggests that the maximum length
  // of the parameter is INFOTIPSIZE (1024), but experimentation determined
  // that descriptions longer than 260 characters cause silent failure, despite
  // SetDescription still returning S_OK in that situation.
  if (item.description.size() > 260)
```

SushiJackal: I'll add a comment!

SushiJackal: `// SetDescription limits the size of the parameter to INFOTIPSIZE (1024),`
`  // which suggests rejection when exceeding that limit, but experimentation`
`  // has shown that descriptions longer than 260 characters cause a silent`
`  // failure, despite SetDescription returning the success code S_OK.`

Would this be okay?

nornagon: seems good to me!

SushiJackal:
Where should this limit be documented? Should there be a change in the documentation? If yes, where?

shell/browser/ui/win/jump_list.cc
SushiJackal: I made a commit with the comment included!

nornagon (approved):

nornagon:
> Should there be a change in the documentation to include this limit? If yes, where?

Yeah, it should go in the docs for `app.setJumpList` under docs/api/app.md

SushiJackal:
@nornagon I added the documentation changes, do they work like this? Added a small remark to jump-list-item.md as well

nornagon (approved):

SushiJackal:
Is there anything left to do?

zcbenz (approved):

PR #28471, opened 2021-03-31
fix: crash when exiting app with active nodejs worker_threads

indutny:
#### Description of Change

Refs https://github.com/electron/electron/pull/28468 for details

Fixes: https://github.com/electron/electron/issues/23315

#### Checklist

- [ ] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [ ] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Fix crash when exiting app with active nodejs worker_threads

PR #28486, opened 2021-04-02
fix: call `UnregisterIsolate` consistently

indutny:
#### Description of Change

`JavascriptEnvironment` is the class that calls `RegisterIsolate()`
so it should be the one to call `UnregisterIsolate`, and this can happen
right before disposing the aforementioned `isolate`.

See: https://github.com/electron/electron/pull/28468

#### Checklist

cc @deepak1556 

- [x] `npm test` passes

#### Release Notes

Notes: no-notes

### DIFF
--- a/shell/browser/javascript_environment.cc
+++ b/shell/browser/javascript_environment.cc
@@ -160,6 +160,8 @@ JavascriptEnvironment::JavascriptEnvironment(uv_loop_t* event_loop)
 }
 
 JavascriptEnvironment::~JavascriptEnvironment() {
+  platform_->DrainTasks(isolate_);
+
   {
     v8::Locker locker(isolate_);
     v8::HandleScope scope(isolate_);
@@ -167,6 +169,10 @@ JavascriptEnvironment::~JavascriptEnvironment() {
   }
   isolate_->Exit();
   g_isolate = nullptr;
+
+  DCHECK_NE(platform_, nullptr);
+  platform_->CancelPendingDelayedTasks(isolate_);
+  platform_->UnregisterIsolate(isolate_);
 }
 
 class EnabledStateObserverImpl final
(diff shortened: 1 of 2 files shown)


shell/browser/javascript_environment.cc
```diff
@@ -167,6 +167,11 @@ JavascriptEnvironment::~JavascriptEnvironment() {
   }
   isolate_->Exit();
   g_isolate = nullptr;
+
+  DCHECK_NE(platform_, nullptr);
+  platform_->DrainTasks(isolate_);
```
deepak1556, line 172: Since we are running tasks, should this be moved before the context and isolate are exited.

indutny: Force pushed. PTAL

deepak1556 (approved):

codebytere (approved):

nornagon (changes requested):
Should this have some release notes? Is there any observable effect from this change? Does it fix any crash?

indutny:
@nornagon there is no observable effect AFAIK. I think it makes sense to add release notes to the [other PR](https://github.com/electron/electron/pull/28471) that got landed in 11-x-y since it actually fixes the worker_threads crash in v11.

shell/browser/javascript_environment.cc
```diff
@@ -160,13 +160,19 @@ JavascriptEnvironment::JavascriptEnvironment(uv_loop_t* event_loop)
...
+
   {
     v8::Locker locker(isolate_);
     v8::HandleScope scope(isolate_);
     context_.Get(isolate_)->Exit();
   }
   isolate_->Exit();
   g_isolate = nullptr;
+
+  DCHECK_NE(platform_, nullptr);
```
nornagon, line 173: is this `DCHECK` checking that `Context::Exit` or `Isolate::Exit` didn't null out `platform_`? If not, then should it perhaps be at the beginning of the function, before the first reference to `platform_`?

indutny: You're absolutely right. This check has to be at the top. I was addressing another review feedback and forgot to move it up with that statement. It is resolved now, though!

nornagon (approved):

indutny-signal:
Should this be merged cc @nornagon @deepak1556 @codebytere ?

indutny:
Awesome. Thanks a lot!

PR #28489, opened 2021-04-02
fix: allow accessing file:// when web security is disabled

zcbenz:
#### Description of Change

Subresource `file://` requests should be explicitly allowed when web security is disabled, closes https://github.com/electron/electron/issues/23757.

This PR also rewrites webview web security tests, since they were passing even though `file://` requests were actually failing.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix failing to request file:// resources when web security is disabled.

### DIFF
--- a/shell/browser/electron_browser_client.cc
+++ b/shell/browser/electron_browser_client.cc
@@ -1369,9 +1369,14 @@ void ElectronBrowserClient::RegisterNonNetworkSubresourceURLLoaderFactories(
   if (!render_process_host || !render_process_host->GetBrowserContext())
     return;
 
+  content::RenderFrameHost* frame_host =
+      content::RenderFrameHost::FromID(render_process_id, render_frame_id);
+  content::WebContents* web_contents =
+      content::WebContents::FromRenderFrameHost(frame_host);
+
   ProtocolRegistry::FromBrowserContext(render_process_host->GetBrowserContext())
       ->RegisterURLLoaderFactories(URLLoaderFactoryType::kDocumentSubResource,
-                                   factories);
+                                   factories, web_contents);
 
 #if BUILDFLAG(ENABLE_ELECTRON_EXTENSIONS)
   auto factory = extensions::CreateExtensionURLLoaderFactory(render_process_id,
@@ -1379,10 +1384,6 @@ void ElectronBrowserClient::RegisterNonNetworkSubresourceURLLoaderFactories(
   if (factory)
     factories->emplace(extensions::kExtensionScheme, std::move(factory));
 
-  content::RenderFrameHost* frame_host =
-      content::RenderFrameHost::FromID(render_process_id, render_frame_id);
-  content::WebContents* web_contents =
-      content::WebContents::FromRenderFrameHost(frame_host);
   if (!web_contents)
     return;
 
--- a/shell/browser/electron_browser_client.cc
+++ b/shell/browser/electron_browser_client.cc
@@ -1286,9 +1286,10 @@ void ElectronBrowserClient::RegisterNonNetworkNavigationURLLoaderFactories(
           context, ukm_source_id,
           false /* we don't support extensions::WebViewGuest */));
 #endif
+  // Always allow navigating to file:// URLs.
   auto* protocol_registry = ProtocolRegistry::FromBrowserContext(context);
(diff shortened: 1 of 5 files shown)


codebytere (approved):

shell/browser/protocol_registry.cc
```diff
@@ -72,11 +75,18 @@ void ProtocolRegistry::RegisterURLLoaderFactories(
     DCHECK(!base::Contains(*factories, url::kFileScheme));
     factories->emplace(url::kFileScheme, AsarURLLoaderFactory::Create());
   } else if (type == URLLoaderFactoryType::kDocumentSubResource) {
-    // Only support requesting file:// subresource URLs when Chromium does so,
-    // it is usually supported under file:// or about:blank documents.
+    // Only support requesting file:// subresource URLs when:
+    // 1. it is supported by Chromium browser, such as under file:// or
+    //    about:blank documents;
+    // 2. web security is turned off.
```
nornagon, line 81: this is (1) OR (2), right?
```suggestion
    // 1. it is supported by Chromium browser, such as under file:// or
    //    about:blank documents; or
    // 2. web security is turned off.
```

shell/browser/protocol_registry.h
```diff
@@ -27,7 +27,8 @@ class ProtocolRegistry {
 
   void RegisterURLLoaderFactories(
       URLLoaderFactoryType type,
-      content::ContentBrowserClient::NonNetworkURLLoaderFactoryMap* factories);
+      content::ContentBrowserClient::NonNetworkURLLoaderFactoryMap* factories,
+      content::WebContents* web_contents = nullptr);
```
nornagon, line 31: nit: I think we can [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) this a bit and just pass a bool for whether to allow file scheme access:

```suggestion
      bool allow_file_access = false);
```

& then move the WebPreferences logic into ElectronBrowserClient::RegisterNonNetworkSubresourceURLLoaderFactories

spec-main/chromium-spec.ts
```diff
@@ -287,6 +287,39 @@ describe('web security', () => {
...
+      return await w.webContents.executeJavaScript('loadFile()');
+    }
+
+    it('is forbidden when web security is enabled', async () => {
+      const w = new BrowserWindow({ show: false, webPreferences: { webSecurity: true } });
+      const result = await loadFile(w);
+      expect(result).to.equal('failed');
+    });
+
+    it('is forbidden when web security is disabled', async () => {
```
nornagon, line 316:
```suggestion
    it('is allowed when web security is disabled', async () => {
```

spec/webview-spec.js
```diff
@@ -321,27 +343,13 @@ describe('<webview> tag', function () {
 
   describe('disablewebsecurity attribute', () => {
     it('does not disable web security when not set', async () => {
-      const jqueryPath = path.join(__dirname, '/static/jquery-2.0.3.min.js');
```
nornagon, line 324: Can we delete jquery-2.0.3.min.js from our tests now?

zcbenz: There are still a few places using it, but I think we should be able to rework those places to remove this file.

nornagon (approved):

jtbandes:
It looks like this PR also fixed https://github.com/electron/electron/issues/28572! 🙌  Was that an intentional side-effect of this change? Would it make sense to add more tests to ensure the SharedWorker behavior specifically does not regress?

shell/browser/electron_browser_client.cc
```diff
@@ -1297,8 +1298,10 @@ void ElectronBrowserClient::
         NonNetworkURLLoaderFactoryMap* factories) {
   auto* protocol_registry =
       ProtocolRegistry::FromBrowserContext(browser_context);
-  protocol_registry->RegisterURLLoaderFactories(
-      URLLoaderFactoryType::kWorkerMainResource, factories);
+  // Workers are not allowed to request file:// URLs, there is no particular
+  // reason for it, and we could consider supporting it in future.
```
jtbandes, line 1302: This comment makes me wonder if the fix for SharedWorkers was unintentional 😛

PR #28496, opened 2021-04-02
feat: enable sandbox by default

### LINKED ISSUE #28466: [Feature Request]: Enable `sandbox: true` by default for `BrowserWindow`
### Preflight Checklist

### Problem Description

Since Electron v5 the default behavior of `BrowserWindow` is to create it with `nodeIntegration: false`. In v12 `contextIsolation` was turned on as well (see #23506). Another default that Electron should have is `sandbox: true`.

### Proposed Solution

I'd like the defaults of creating a new `BrowserWindow` to be equivalent to the following:

```js
  webPreferences: {
    nodeIntegration: false,
    contextIsolation: true,
    sandbox: true
  }
);
```

### Alternatives Considered

No alternatives considered. As @MarshallOfSound said in #23506:

> We're making this change to improve the **default** security of Electron apps so that your app is only insecure if you have deliberately opted in to the insecure behaviour.

### Additional Information

Happy to dig and try to find any data needed to make a decission.

nornagon:
#### Description of Change
Closes #28466.

This enables `sandbox: true` by default unless `nodeIntegration: true` is
specified. This is a non-breaking change because without `nodeIntegration:
true`, there's no documented way to perform any action that would cause
different behavior depending on whether the sandbox is enabled or not.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Renderer processes without node integration enabled are now sandboxed by default.


MarshallOfSound (changes requested):
This is most definitely a breaking change, `nodeIntegration: false` still has NodeJS in preload scripts.  Defaulting to `sandbox: true` will break preload script that utilize node.js

nornagon:
@MarshallOfSound do you have thoughts about how we can approach making this the default?

perhaps a phased deprecation:

1. If `preload` is empty and `nodeIntegration` is false or unspecified, enable sandbox by default.
2. If `preload` is non-null and `nodeIntegration` is false or unspecified, and `sandbox` is unspecified, warn that the default will be changing.
3. Enable sandbox by default when `nodeIntegration` is false or unspecified (i.e. this PR)

MarshallOfSound:
> If preload is empty and nodeIntegration is false or unspecified, enable sandbox by default.

I think you also have to care about `nodeIntegrationInSubFrames` and `nodeIntegrationInWorkers` too, in general I think those three steps if sufficiently documented are a viable path to enabling sandbox by default.

nornagon:
Is it the case that `nodeIntegrationIn(SubFrames|Workers): true` w/o `nodeIntegration: true` has an effect? that seems odd, that you'd be able to enable node integration in subframes/workers without it being enabled in the main frame, but uh, if so, then sure.

nornagon:
Closing in favor of #30197

PR #28456, opened 2021-03-31
chore: remove deprecated `worldSafeExecuteJavaScript` option

miniak:
#### Description of Change
Remove deprecated `worldSafeExecuteJavaScript` option from `webPreferences`.

This is already documented in Electron 14 breaking changes: 
https://github.com/electron/electron/blob/master/docs/breaking-changes.md#removed-worldsafeexecutejavascript

#### Checklist
- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes
Notes: Deprecated `worldSafeExecuteJavaScript` option was removed from `webPreferences`.

PR #28497, opened 2021-04-02
refactor: clean up webFrame implementation to use gin wrappers

MarshallOfSound:
The previous implementation of webFrame in the renderer process leaked sub-frame contexts and global objects across the context boundaries thus making it possible for apps to either maliciously or accidentally violate the contextIsolation boundary.

E.g. `electron.webFrame.top.context` leaked the `window` object from the other side of the isolated world by accident (not by API, purely by implementation issues)

This re-implementation binds all methods in native code directly to content::RenderFrame instances instead of relying on JS to provide a "window" with every method request.  This is much more consistent with the rest of the Electron codebase and is substantially safer.

Notes: no-notes


mesca216 (approved):

mesca216 (commented):
**Deleted**

lib/renderer/web-frame-init.ts
```diff
@@ -13,11 +13,6 @@ export const webFrameInit = () => {
   ipcRendererUtils.handle(IPC_MESSAGES.RENDERER_WEB_FRAME_METHOD, (
     event, method: keyof WebFrameMethod, ...args: any[]
   ) => {
-    // TODO(MarshallOfSound): Remove once the world-safe-execute-javascript deprecation warning is removed
-    if (method.startsWith('executeJavaScript')) {
-      return (webFrame as any)[`_${method}`](...args);
-    }
```
nornagon, line 19: do we want to keep this in for backportability?

MarshallOfSound: Removing this (the deprecation warning) makes the implementation easier.  Need to figure out how best to handle this in older branches you're right

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -349,555 +349,640 @@ class SpellCheckerHolder final : public content::RenderFrameObserver {
...
-  auto* render_frame = GetRenderFrame(window);
-  if (!render_frame) {
-    thrower.ThrowError(
-        "Render frame was torn down before webFrame.setVisualZoomLevelLimits "
-        "could be executed");
-    return;
+  bool MaybeGetRenderFrame(std::string* error_msg,
+                           const std::string& method_name,
+                           content::RenderFrame** render_frame_ptr) {
+    auto* frame = render_frame();
```
nornagon, line 432:
```suggestion
    content::RenderFrame* frame = render_frame();
```

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -349,555 +349,640 @@ class SpellCheckerHolder final : public content::RenderFrameObserver {
...
-  SetZoomLevel(thrower, window, blink::PageZoomFactorToZoomLevel(factor));
-}
+  const char* GetTypeName() override { return "WebFrameRenderer"; }
 
-double GetZoomFactor(gin_helper::ErrorThrower thrower,
-                     v8::Local<v8::Value> window) {
-  double zoom_level = GetZoomLevel(thrower, window);
-  return blink::PageZoomLevelToZoomFactor(zoom_level);
-}
+  void OnDestruct() override {}
```
nornagon, line 415:
```suggestion
  void OnDestruct() override {
    // The underlying RenderFrame can be destroyed while this class
    // still lives, so all calls to |render_frame()| must be null-checked.
    // The MaybeGetRenderFrame method handles this.
  }
```

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -349,555 +349,640 @@ class SpellCheckerHolder final : public content::RenderFrameObserver {
...
-        "could be executed");
-    return;
+  bool MaybeGetRenderFrame(std::string* error_msg,
+                           const std::string& method_name,
+                           content::RenderFrame** render_frame_ptr) {
+    auto* frame = render_frame();
+    if (!frame) {
+      *error_msg = "Render frame was torn down before webFrame." + method_name +
+                   " could be "
+                   "executed";
```
nornagon, line 436:
```suggestion
                   " could be executed";
```

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -349,555 +349,640 @@ class SpellCheckerHolder final : public content::RenderFrameObserver {
...
-    return gin::ConvertToV8(isolate, prefs.enable_plugins);
-  } else if (pref_name == options::kEnableWebSQL) {
-    return gin::ConvertToV8(isolate, prefs.enable_websql);
-  } else if (pref_name == options::kWebviewTag) {
-    return gin::ConvertToV8(isolate, prefs.webview_tag);
-  }
-  return v8::Null(isolate);
-}
+ private:
+  bool MaybeGetRenderFrame(const gin_helper::ErrorThrower& thrower,
```
nornagon, line 418: slight preference for this to receive `v8::Isolate*` instead of `gin_helper::ErrorThrower`. The latter is a zero-cost wrapper around an isolate pointer and can be "constructed" for free.

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -349,555 +349,640 @@ class SpellCheckerHolder final : public content::RenderFrameObserver {
...
 
-  blink::WebFrame* web_frame = render_frame->GetWebFrame();
-  if (web_frame->IsWebLocalFrame()) {
-    web_frame->ToWebLocalFrame()->GetDocument().RemoveInsertedStyleSheet(
-        blink::WebString::FromUTF16(key));
+  void AllowGuestViewElementDefinition(gin_helper::ErrorThrower thrower,
+                                       v8::Local<v8::Object> context,
+                                       v8::Local<v8::Function> register_cb) {
+    v8::HandleScope handle_scope(thrower.isolate());
+    v8::Context::Scope context_scope(context->CreationContext());
```
nornagon, line 563: Why does this method have a handle / context scope when none of the others do? It doesn't look like it's ever called from outside of V8.

Also, is `context` still needed here?

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -349,555 +349,640 @@ class SpellCheckerHolder final : public content::RenderFrameObserver {
...
+                                      v8::Isolate* isolate) {
+    content::RenderFrame* render_frame;
+    if (!MaybeGetRenderFrame(thrower, "parent", &render_frame))
+      return v8::Null(isolate);
+
+    blink::WebFrame* frame = render_frame->GetWebFrame()->Parent();
+    if (frame && frame->IsWebLocalFrame())
+      return WebFrameRenderer::Create(
+                 isolate,
+                 content::RenderFrame::FromWebFrame(frame->ToWebLocalFrame()))
```
nornagon, line 878: What does this change do..?

shell/renderer/api/electron_api_web_frame.cc
```diff
@@ -349,555 +349,640 @@ class SpellCheckerHolder final : public content::RenderFrameObserver {
...
+      return v8::Null(isolate);
   }
 
-  return render_frame->GetRoutingID();
-}
+  v8::Local<v8::Value> FindFrameByName(gin_helper::ErrorThrower thrower,
+                                       v8::Isolate* isolate,
+                                       const std::string& name) {
+    content::RenderFrame* render_frame;
+    if (!MaybeGetRenderFrame(thrower, "getFrameForSelector", &render_frame))
```
nornagon, line 959:
```suggestion
    if (!MaybeGetRenderFrame(thrower, "findFrameByName", &render_frame))
```

nornagon (changes requested):

shell/renderer/api/electron_api_web_frame.cc
MarshallOfSound: I don't know, this method was copy-pasted.  Happy to take a look at some point but I don't want to change actual implementations here.

shell/renderer/api/electron_api_web_frame.cc
MarshallOfSound: This specifically, changes `GetFrameParent` to return a new `WebFrameRenderer` instance for the given RenderFrame instead of just returning that frames global object.

nornagon: Ah, I see, we used to do that wrapping on the JS side. 👍

nornagon (approved):

lib/renderer/web-frame-init.ts
miniak: I am removing this in https://github.com/electron/electron/pull/28456

miniak (approved):

miniak:
@MarshallOfSound there are conflicts after my PR https://github.com/electron/electron/pull/28456 was merged

PR #28510, opened 2021-04-05
fix: update crash-reporter.md,crashReporter.start() is not supported in the renderer

bei-yang:
The latest version（12.0.2） does not support the renderer process

### DIFF
--- a/docs/api/crash-reporter.md
+++ b/docs/api/crash-reporter.md
@@ -2,7 +2,7 @@
 
 > Submit crash reports to a remote server.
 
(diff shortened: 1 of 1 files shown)


nornagon (changes requested):
This is not true, the `addExtraParameter`, `removeExtraParameter` and `getParameters` methods are available in the renderer.

zcbenz:
I'm closing this PR since crashReporter is available in renderer, and it is documented that start() is only available in the main process.

PR #28517, opened 2021-04-05
build: add safety net for mac_deployment_target

MarshallOfSound:
As of #28480 we now dynamically determine the `LSMinimumSystemVersion` value as part of the build process.  To avoid this changing and no one realizing we now have this assert which will trip during a Chromium upgrade if they bump the minimum supported macOS version so we can update our documentation appropriately.

Notes: no-notes

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -32,6 +32,10 @@ if (is_mac) {
   import("//ui/gl/features.gni")
   import("//v8/gni/v8.gni")
   import("build/rules.gni")
+
+  assert(
+      mac_deployment_target == "10.11.0",
+      "Chromium has updated the mac_deployment_target, please update this assert, update the supported versions documentation (docs/tutorial/support.md) and flag this as a breaking change")
 }
 
 if (is_linux) {


BUILD.gn
```diff
@@ -32,6 +32,10 @@ if (is_mac) {
   import("//ui/gl/features.gni")
   import("//v8/gni/v8.gni")
   import("build/rules.gni")
+
+  assert(
+      mac_deployment_target == "10.11.0",
+      "Chromium has updated the mac_deployment_target, please update this assert, update the supported versions documentation and flag this as a breaking change")
```
nornagon, line 38: put a path to the doc here?

nornagon (approved):

miniak (approved):

indutny-signal:
So looks like this didn't actually work for 20-x-y release which dropped macOS 10.11 and 10.12. Any idea why?

MarshallOfSound:
@indutny-signal it tripped, it just didn't result in the supported platforms / breaking changes doc being updates 🙃 

https://github.com/electron/electron/pull/34272/files

indutny-signal:
Welp :) Was that PR opened by a bot? If yes - could we modify the bot to trip more alarms and fail unless human intervenes?

PR #28513, opened 2021-04-05
fix: pass postData to new-window event

nornagon:
#### Description of Change
This adds the missing `postData` parameter to `new-window`, which was previously omitted. This is the only call site of `emitDeprecatedNewWindowEvent`, so I can only assume its omission was accidental.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).

#### Release Notes

Notes: Fixed `postData` parameter missing from `new-window` event.

Issue #28377, opened 2021-03-24
[Feature Request]: Expose @electron/internal/common/parse-features-string


Prinzhorn:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/electron/electron/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

`new-window` deprecated => `setWindowOpenHandler` only gives me a raw `details.features` string. I'd like to use `parse-features-string.ts` to parse it.

### Proposed Solution

Please let me use [`parse-features-string`](https://github.com/electron/electron/search?q=parse-features-string), maybe as static `BrowserWindow.parseFeaturesString`

### Alternatives Considered

Writing my own parser. Or maybe `setWindowOpenHandler` could give me a parsed thing? `new-window` had `options` with parsed `width` `height` etc.

### Additional Information

Maybe it's already possible to import it and I just can't find it

Prinzhorn:
Closing in favor of #28380

PR #28518, opened 2021-04-05
feat: add more info in setWindowOpenHandler details

### LINKED ISSUE #28380: [Feature Request]: Add {options, disposition, additionalFeatures, referrer, postData} to setWindowOpenHandler
### Preflight Checklist

### Problem Description

This supersedes #28377. `setWindowOpenHandler` is missing all the fancy arguments that `new-window` had. And `did-create-window` is too late because you need the arguments to decide what to do.

Some examples:

1. I only want `disposition === 'new-window'` and `deny` all others
2. In case of `disposition === '*-tab'` I want to `deny` and instead create a `BrowserView` for my custom tab implementation
3. I want to handle `postBody` without creating a new window

You get the gist.

### Proposed Solution

If `setWindowOpenHandler` is supposed to replace `new-window` it needs to offer the same flexibility

### Alternatives Considered

Use `did-create-window` and undo the window if needed. But that defeats the original purpose

> new-window is fired after the contents is already created by Chrome. This is a problem when the application wants to prevent that window from being opened, as killing the window during initialization causes many problems and leads to the inevitable crash.

### Additional Information

n/a

nornagon:
#### Description of Change
This adds `disposition`, `referrer` and `postBody` to the details object passed to the window open handler.

Builds on https://github.com/electron/electron/pull/28513, which fixes a number of PostBody-related issues.

Closes #28380.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `disposition`, `referrer` and `postBody` to the details object passed to the window open handler registered with `setWindowOpenHandler`.


nornagon:
This should perhaps be target/12-x-y too

Prinzhorn:
Would it be possible to add the parsed `options` as well? Originally I opened #28377 before I realized `setWindowOpenHandler` was lacking additional information too.

nornagon:
> Would it be possible to add the parsed `features` as well? Originally I opened #28377 before I realized `setWindowOpenHandler` was lacking additional information too.

parsed features is a little more complicated because:

1. there's already a property called `features` in the details object which contains the raw features string, and
2. parsing the features currently splits the results between `options`, `webPreferences` and `additionalFeatures`, which isn't really a great API and I'd rather not propagate that bad API to the new method.

If it helps, the code for parsing the features isn't much more complicated than `features.split(",")`: https://github.com/electron/electron/blob/77dcf1020a5736235838353ee72bdc75900dc417/lib/common/parse-features-string.ts#L59-L72

Prinzhorn:
@nornagon thanks for looking into this, I'll probably use JSON then if I can use any string. I'm in control of the features string anyway.

miniak (approved):

jkleinsc (approved):

SonjayaJetBrain (approved):

zcbenz (approved):

SonjayaJetBrain:
automation call pr with json to travis

Issue #23252, opened 2020-04-23
Cannot use ELECTRON_ENABLE_SECURITY_WARNINGS and load WASM


alexdima:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 8.2.3
* **Operating System:**
  * macOS 10.15.4 

### Expected Behavior

I would expect to be able to use `ELECTRON_ENABLE_SECURITY_WARNINGS` and be able to load WASM.

### Actual Behavior

From https://github.com/WebAssembly/content-security-policy/issues/7 , it looks like the only way to be able to use WASM is to add `'unsafe-eval'` to the CSP. But doing that triggers Electron's warning.

### To Reproduce
Self contained example at https://github.com/alexdima/electron-wasm-issue

### Screenshots

Without `'unsafe-eval'`:
![image](https://user-images.githubusercontent.com/5047891/80075250-b4771a80-854a-11ea-8d59-6947de1dc6ef.png)

With `'unsafe-eval'`:
![image](https://user-images.githubusercontent.com/5047891/80075295-c5c02700-854a-11ea-906a-b4485e29df00.png)

### Additional Information

We would like to be able to use `ELECTRON_ENABLE_SECURITY_WARNINGS` and load WASM at the same time.

cc @bpasero @deepak1556

electron-triage:
The Electron version reported on this issue is no longer supported. See our [supported versions documentation](https://www.electronjs.org/docs/tutorial/support#supported-versions).

**If you're still experiencing this issue on a supported version, please open a new issue with an updated repro - a [Fiddle](https://www.electronjs.org/fiddle) is very appreciated.**

Electron has a large issues backlog. To help our team prioritize, we're closing older issues and asking for new issues with updated repro steps if it affects a supported version. This helps sort what issues are still relevant and helps us fix them more quickly.

Thanks for your patience and understanding!

alexdima:
This still reproduces with Electron 11.x using the exact same steps (the same repo)

[...]

PR #28529, opened 2021-04-06
feat: add process.setWasmCodegenAllowed API

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/23252.

Currently according to WebAssembly/content-security-policy#7, the only way to be able to use wasm is to add 'unsafe-eval' to the CSP, which makes wasm unavailable to apps that want to have safe content security policies.

This PR adds a `process.setWasmCodegenAllowed` API that override the default check routine to force enabling/disabling wasm code generation.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Add `process.setWasmCodegenAllowed` API.

### DIFF
--- a/docs/api/process.md
+++ b/docs/api/process.md
@@ -14,6 +14,7 @@ In sandboxed renderers the `process` object contains only a subset of the APIs:
 
 * `crash()`
 * `hang()`
+* `setWasmCodegenAllowed()`
 * `getCreationTime()`
 * `getHeapStatistics()`
 * `getBlinkMemoryInfo()`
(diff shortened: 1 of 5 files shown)


deepak1556:
https://bugs.chromium.org/p/chromium/issues/detail?id=948834#c19 is opening up `wasm-eval` to all URLs behind a runtime flag, we can just backport the change instead of a new api. Thoughts ? Given the comment in https://chromium-review.googlesource.com/c/chromium/src/+/2724817/16/third_party/blink/web_tests/virtual/wasm-csp/README.md, looks like it will be default at some point.

zcbenz:
It looks a much better solution! I'll start a new PR for it.

PR #28535, opened 2021-04-06
fix: support wasm-eval csp behind WebAssemblyCSP flag

### LINKED ISSUE #23252: Cannot use ELECTRON_ENABLE_SECURITY_WARNINGS and load WASM

zcbenz:
#### Description of Change

This PR backports https://chromium.googlesource.com/chromium/src/+/83913676803db53648b6a47d159102a7cf1dac36 from upstream Chromium, 
which makes the `wasm-eval` CSP available behind the `WebAssemblyCSP` blink runtime feature.

This feature enables apps to use wasm without having to use the 'unsafe-eval' CSP, closes #23252.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Support `wasm-eval` csp behind `WebAssemblyCSP` flag.


codebytere:
Do we think this is `semver-minor`, as opposed to a fix and thus `semver-patch`?

zcbenz:
This is a change already in upstream, and we will get it in Electron after a few Chromium upgrades. The major purpose of this PR is to add tests and backport it to older branches.

And this is probably more `semver-patch` than `semver-minor`, since it is behind a blink flag.

deepak1556 (approved):
Nice!

nornagon (approved):

PR #28550, opened 2021-04-06
feat: remove BrowserWindow option inheritance

nornagon:
#### Description of Change
Removes the deprecated and not terribly useful 'feature' of inheriting
BrowserWindow options from parents on `window.open()`.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Child windows no longer inherit BrowserWindow construction options from their parents.


lib/browser/guest-window-manager.ts
```diff
@@ -246,7 +244,7 @@ export function makeWebPreferences ({ embedder, secureOverrideWebPreferences = {
   const openerId = parentWebPreferences.nativeWindowOpen ? null : embedder.id;
 
   return {
-    ...(useDeprecatedBehaviorForOptionInheritance && deprecatedInheritedOptions ? deprecatedInheritedOptions.webPreferences : null),
+    ...(embedder.getOwnerBrowserWindow() ? null : parentWebPreferences),
```
jkleinsc, line 247: @nornagon why are the parentWebPreferences used here?  Isn't this what is being removed?  I'm guessing I'm missing something here but since `getDeprecatedInheritedOptions` returned `embedder.getLastWebPreferences()`, what's the difference of this change?

nornagon: I was trying to fix some tests, this was bringing back _only_ `webPreferences` (not BrowserWindow options), and only for <webview>s. I might have fixed the tests a different way though, so I'll try again to remove this completely.

miniak (approved):

jkleinsc (approved):

zcbenz (approved):

PR #28552, opened 2021-04-06

### LINKED ISSUE #28511: Make nativeWindowOpen: true the default
The native window.open system has been better for quite a while now, and it resolves a lot of issues. We should make it the default and deprecate the old `BrowserWindowProxy`-based method.

nornagon:
#### Description of Change
Make `nativeWindowOpen: true` the default.

This option has been better than the alternative for some time (fewer bugs,
faster new windows, and more flexible). Let's make it the default.

Closes #28511.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: `nativeWindowOpen: true` is now the default.

### DIFF
--- a/shell/browser/web_contents_preferences.cc
+++ b/shell/browser/web_contents_preferences.cc
@@ -478,7 +478,7 @@ void WebContentsPreferences::OverrideWebkitPrefs(
   GetPreloadPath(&prefs->preload);
 
   // Check if nativeWindowOpen is enabled.
-  prefs->native_window_open = IsEnabled(options::kNativeWindowOpen);
+  prefs->native_window_open = IsEnabled(options::kNativeWindowOpen, true);
 
   // Check if we have node integration specified.
   prefs->node_integration = IsEnabled(options::kNodeIntegration);
--- a/shell/browser/web_contents_preferences.cc
+++ b/shell/browser/web_contents_preferences.cc
@@ -141,7 +141,7 @@ WebContentsPreferences::WebContentsPreferences(
   SetDefaultBoolIfUndefined(options::kDisableHtmlFullscreenWindowResize, false);
   SetDefaultBoolIfUndefined(options::kWebviewTag, false);
   SetDefaultBoolIfUndefined(options::kSandbox, false);
-  SetDefaultBoolIfUndefined(options::kNativeWindowOpen, false);
+  SetDefaultBoolIfUndefined(options::kNativeWindowOpen, true);
   SetDefaultBoolIfUndefined(options::kContextIsolation, true);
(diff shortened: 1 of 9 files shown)


shell/browser/web_contents_preferences.cc
```diff
@@ -141,7 +141,7 @@ WebContentsPreferences::WebContentsPreferences(
   SetDefaultBoolIfUndefined(options::kDisableHtmlFullscreenWindowResize, false);
   SetDefaultBoolIfUndefined(options::kWebviewTag, false);
   SetDefaultBoolIfUndefined(options::kSandbox, false);
-  SetDefaultBoolIfUndefined(options::kNativeWindowOpen, false);
+  SetDefaultBoolIfUndefined(options::kNativeWindowOpen, true);
```
miniak, line 144: you need to also change the default value in https://github.com/electron/electron/blob/master/shell/browser/web_contents_preferences.cc#L481

miniak, line 144: you need to also change the default value in https://github.com/electron/electron/blob/968b30c9b4459e4f7030042b595f26926481e260/shell/browser/web_contents_preferences.cc#L481 due to the problem described in https://github.com/electron/electron/pull/18369

miniak: please also cleanup https://github.com/electron/electron/blob/968b30c9b4459e4f7030042b595f26926481e260/shell/browser/web_contents_preferences.cc#L196-L202

miniak (changes requested):

shell/browser/web_contents_preferences.cc
nornagon: Can you expand on how `SetDefaults` should be cleaned up? We still need to enable `kNativeWindowOpen` when sandbox is true in case a caller explicitly sets it to false, this will override it.

nornagon: > you need to also change the default value in
> 
> https://github.com/electron/electron/blob/968b30c9b4459e4f7030042b595f26926481e260/shell/browser/web_contents_preferences.cc#L481
> 
> due to the problem described in #18369

also, wondering how we could write a test for this?

miniak: > Can you expand on how `SetDefaults` should be cleaned up? We still need to enable `kNativeWindowOpen` when sandbox is true in case a caller explicitly sets it to false, this will override it.
You're right, ignore this comment. I actually added that :)

miniak (approved):

jkleinsc (approved):

codebytere (approved):

zcbenz (approved):

noncash15:
nativewindowopen-default#28552

PR #28567, opened 2021-04-07
build: give ASAN tests more memory to avoid SIGKILL or disabling tests

MarshallOfSound:
Reverts #28156 and #28509 in favor of just giving the ASAN tests a bigger box with more memory, previously they had a 4GB box.  They now get a 16GB box (with potential for us to go all the way to 40GB if we wanted to, but we really shouldn't need to.

Notes: no-notes

### DIFF
--- a/spec-main/chromium-spec.ts
+++ b/spec-main/chromium-spec.ts
@@ -367,8 +367,7 @@ describe('web security', () => {
   });
 });
 
-// Running child app under ASan might receive SIGKILL because of OOM.
-ifdescribe(!process.env.IS_ASAN)('command line switches', () => {
+describe('command line switches', () => {
   let appProcess: ChildProcess.ChildProcessWithoutNullStreams | undefined;
   afterEach(() => {
     if (appProcess && !appProcess.killed) {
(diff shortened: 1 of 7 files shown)


nornagon (approved):
📈

spec-main/chromium-spec.ts
```diff
@@ -375,7 +376,8 @@ ifdescribe(!process.env.IS_ASAN)('command line switches', () => {
     ifit(process.platform === 'linux')('should not change LC_ALL when --lang is not set', async () => testLocale('', lcAll, true));
   });
 
-  describe('--remote-debugging-pipe switch', () => {
+  // TODO(nornagon): figure out why these tests fail under ASan.
+  ifdescribe(!isAsan)('--remote-debugging-pipe switch', () => {
```
zcbenz, line 380: I think it would be better manually removing all `ifdescribe(!process.env.IS_ASAN)` rather than reverting the commits, there are a few changes that should be kept.

For example tests like this can be enabled for asan, and the comment can be removed, and some tests were moved to `spec-main/fixtures/crash-cases` and it would be better keep them there.

jkleinsc (approved):

Issue #21699, opened 2020-01-08
dev tools: unable to load source maps over custom protocol


bendemboski:
### Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

* **Electron Version:**
  * 7.1.7
* **Operating System:**
  * macOS 10.14.6
* **Last Known Working Electron version:**
  * 6.1.7

### Expected Behavior
When loading a Javascript file using a custom protocol, opening the dev tools should load the source maps.

### Actual Behavior
It does not load the source maps, and produces no error message other than an `ERR_UNKNOWN_URL_SCHEME` in the net log if net logging is enabled.

### To Reproduce
To see it work properly on Electron 6.1.7:
1. Run [this fiddle](https://gist.github.com/d0cb93e9c7d6ad198fe38db77038495a) on Electron 6.1.7
2. Open the dev tools
3. Note that `source map requested!` was logged to the main process console, and the source map for `test.js` (including `test.ts`) was loaded into the dev tools

To see it fail on Electron 7.1.7:
1. Run [this fiddle](https://gist.github.com/d0cb93e9c7d6ad198fe38db77038495a) on Electron 7.1.7
2. Open the dev tools
3. Note that `source map requested!` was not logged to the main process console, and neither `test.js.map` nor `test.ts` were loaded

### Additional Information
This is probably more fallout from Chromium's network service. When I enable net logging (`app.commandLine.appendSwitch('log-net-log', 'netlog.json')`), I see the following when opening the dev tools:

```
{"params":{"priority":"IDLE","traffic_annotation":129652775,"url":"serve://dist/assets/test.js.map"},"phase":1,"source":{"id":27,"type":1},"time":"20404995","type":2},
{"phase":1,"source":{"id":27,"type":1},"time":"20404995","type":103},
{"phase":2,"source":{"id":27,"type":1},"time":"20404995","type":103},
{"params":{"load_flags":0,"method":"GET","privacy_mode":0,"url":"serve://dist/assets/test.js.map"},"phase":1,"source":{"id":27,"type":1},"time":"20404995","type":99},
[...]

Issue #28572, opened 2021-04-07
[Bug]: SharedWorker fails to load from asar


jtbandes:
### Preflight Checklist

I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
I have searched the [issue tracker](https://www.github.com/electron/electron/issues) for a feature request that matches the one I want to file, without success.

### Electron Version

12.0.0

### What operating system are you using?

macOS

### Operating System Version

macOS Big Sur 11.2.3

### What arch are you using?

arm64 (including Apple Silicon)

### Expected Behavior

A renderer should be able to `new SharedWorker()` or `new Worker()` with a URL pointing to a script inside the asar.

### Actual Behavior

A regular `Worker` script can be loaded from the asar, but a `SharedWorker` can not. A `SharedWorker` works properly when running via `electron .webpack`, but fails to load when packaged into an app with `.asar`.

Limited error information is available, because the worker doesn't appear in `webContents.getAllSharedWorkers()` so dev tools cannot be opened (possibly related: #25512). The worker's `onerror` event is triggered with an error, but the error contains little information. Here is an example screenshot where I reproduced the issue in the developer console. The Network tab doesn't include any info about the worker script we're trying to load.

![](https://user-images.githubusercontent.com/14237/113946383-076dd000-97a4-11eb-9d78-85c6d1802563.png)

On the other hand, I am able to `fetch` the URL directly. So if I run this in the dev console:
```js
const w = new SharedWorker(URL.createObjectURL(await (await fetch("...app.asar/...")).blob()));
w.onerror = (e) => console.log("worker error", e);
```
then the worker does load and I can open its dev tools (although it soon encounters an error when the webpack-bundled worker script tries to use `importScripts` with a relative URL that doesn't work on my `blob:` url.)

### Testcase Gist URL

https://gist.github.com/jtbandes/c339b490fb4914c54d61d6fda9134096

I found a similar issue discussed [on Stack Overflow](https://stackoverflow.com/q/59254016/23649), but it seems these people are trying to use a SharedWorker in node, but I'm trying to use it in the renderer.

jtbandes:
I created a sample project to reproduce the issue: https://gist.github.com/jtbandes/c339b490fb4914c54d61d6fda9134096

[...]

PR #28573, opened 2021-04-08
fix: load source maps from custom protocols and asar bundles

deepak1556:
#### Description of Change

Fixes https://github.com/electron/electron/issues/21699

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: Allow loading source maps from custom protocols and asar bundles

### DIFF
--- a/filenames.gni
+++ b/filenames.gni
@@ -404,6 +404,8 @@ filenames = {
     "shell/browser/native_window_observer.h",
     "shell/browser/net/asar/asar_url_loader.cc",
     "shell/browser/net/asar/asar_url_loader.h",
+    "shell/browser/net/asar/asar_url_loader_factory.cc",
+    "shell/browser/net/asar/asar_url_loader_factory.h",
     "shell/browser/net/cert_verifier_client.cc",
     "shell/browser/net/cert_verifier_client.h",
     "shell/browser/net/electron_url_loader_factory.cc",
--- a/shell/browser/net/asar/asar_url_loader_factory.cc
+++ b/shell/browser/net/asar/asar_url_loader_factory.cc
@@ -1,0 +1,40 @@
+// Copyright (c) 2021 GitHub, Inc.
+// Use of this source code is governed by the MIT license that can be
+// found in the LICENSE file.
+
+#include "shell/browser/net/asar/asar_url_loader_factory.h"
+
+#include "shell/browser/net/asar/asar_url_loader.h"
+
+namespace electron {
+
+// static
+mojo::PendingRemote<network::mojom::URLLoaderFactory>
+AsarURLLoaderFactory::Create() {
+  mojo::PendingRemote<network::mojom::URLLoaderFactory> pending_remote;
+
+  // The AsarURLLoaderFactory will delete itself when there are no more
+  // receivers - see the SelfDeletingURLLoaderFactory::OnDisconnect method.
+  new AsarURLLoaderFactory(pending_remote.InitWithNewPipeAndPassReceiver());
+
+  return pending_remote;
+}
+
+AsarURLLoaderFactory::AsarURLLoaderFactory(
+    mojo::PendingReceiver<network::mojom::URLLoaderFactory> factory_receiver)
+    : network::SelfDeletingURLLoaderFactory(std::move(factory_receiver)) {}
+AsarURLLoaderFactory::~AsarURLLoaderFactory() = default;
+
+void AsarURLLoaderFactory::CreateLoaderAndStart(
+    mojo::PendingReceiver<network::mojom::URLLoader> loader,
+    int32_t routing_id,
+    int32_t request_id,
+    uint32_t options,
+    const network::ResourceRequest& request,
+    mojo::PendingRemote<network::mojom::URLLoaderClient> client,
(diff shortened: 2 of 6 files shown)


sentialx:
Does it also fix handling source maps on `chrome-extension` protocol?

deepak1556:
Nope it does not, can you provide a example use case. I don't see chrome providing support for the extension protocol https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/devtools/devtools_ui_bindings.cc;l=912-963

sentialx:
I noticed there are lots of warnings like this
image

While on Chrome it doesn't occur

deepak1556:
Interesting, can you open an issue with minimal repro, it will help debug the issue. I could add the extension URLLoaderFactory to fix the issue, but would like to understand why the issue doesn't happen in chrome.

jkleinsc (commented):
@deepak1556 could we add test(s) for this?

nornagon (commented):
This changes every `file://` request to go through the ASAR loader instead. How sure are we that this doesn't break something / change behavior for non-asar files? Should we restrict this to only asar files in the app's bundle?

deepak1556:
@nornagon the asar URLLoader will default to file factory if its not an asar file https://github.com/electron/electron/blob/master/shell/browser/net/asar/asar_url_loader.cc#L110 , also we use the asar factory as the default for file scheme in the navigation factory list https://github.com/electron/electron/blob/master/shell/browser/electron_browser_client.cc#L1291 so it shouldn't break any existing behavior.

nornagon:
ah so this change is only for subresources?

nornagon (approved):

nornagon:
Does this also fix https://github.com/electron/electron/issues/28572 ?

deepak1556:
>  could we add test(s) for this?

This required opening a source file on the devtools source panel. I tried to emulate a unittest behavior by issuing the [Network.loadNetworkResource](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-loadNetworkResource) CDP to trigger the custom protocol response handler but that didn't work. Will check if Debugger or Page module commands help.

deepak1556:
> ah so this change is only for subresources?

Specifically only the resources loaded by devtools.

> Does this also fix #28572 ?

Nope, that requires change in https://github.com/electron/electron/blob/master/shell/browser/electron_browser_client.cc#L1296. We set the asar factory if chromium already sets a factory for the file scheme https://github.com/electron/electron/blob/master/shell/browser/protocol_registry.cc#L67, which for a worker depends on the requesting origin to be on file scheme as well https://source.chromium.org/chromium/chromium/src/+/master:content/browser/worker_host/worker_script_fetch_initiator.cc;l=227. Based on the users' test case the conditions match, so not sure why the issue is hit. Requires further investigation.

jtbandes:
Thanks for looking into it! I'm excited about this fix for source maps as well as shared workers.

I actually noticed that on latest master, the SharedWorker from my sample code in #28572 seems to work correctly. So it may already have been fixed. I'm curious if anyone knows which change it was that fixed it?

deepak1556:
> Edit: looks like it was fixed by #28489.

Yup that would be correct. Before the change asar factory was not registered for worker resources, was looking at the code in master while writing https://github.com/electron/electron/pull/28573#issuecomment-816890659

jtbandes:
Should I be concerned by this comment? It makes it sound like allowing workers to load `file://` urls was unintentional.
 https://github.com/electron/electron/blob/247c4c89b2634453ba6b3b92c19a83cc0853c8db/shell/browser/electron_browser_client.cc#L1301-L1304

deepak1556:
yes the change does seem unintentional. 

/cc @zcbenz the PR changes the registration of asar factory from

```
if (type == URLLoaderFactoryType::kNavigation) {
```

to

```
auto file_factory = factories->find(url::kFileScheme);
  if (file_factory != factories->end()) {
    // If Chromium already allows file access then replace the url factory to
    // also loading asar files.
    file_factory->second = AsarURLLoaderFactory::Create();
  } 
```

`RegisterNonNetworkWorkerMainResourceURLLoaderFactories` factory map will contain file factory as long as the embedding page origin is file https://source.chromium.org/chromium/chromium/src/+/master:content/browser/worker_host/worker_script_fetch_initiator.cc;l=99 . Should we remove the comment in that case ?

zcbenz:
> Should I be concerned by this comment? It makes it sound like allowing workers to load `file://` urls was unintentional.

There are 2 cases for web workers to load asar files:
1. when Chromium already allows loading file:// urls, for example from local pages;
2. when Chromium does not allow loading file:// urls, for example from remote pages.

Case 1 is allowed by #28489, i.e. if web worker can load local files before then it now can also recognize asar files.

Case 2 is not, i.e. if web worker can not load local files before then it still can not read neither local files or asar files. This is what the referenced comment means.

I agree the comment is not precise, it should be "Workers are not allowed to request file:// URLs __in remote pages__".

zcbenz (approved):

deepak1556:
Unable to create a reliable test with CDP, merging based on manual testing.

nornagon:
I'm removing needs-manual-bp/11-x-y since 11.x is security fixes only now.

PR #28621, opened 2021-04-12
build: better error handling for release builds

codebytere:
#### Description of Change

Mild improvement in release build error handling.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/script/release/ci-release-build.js
+++ b/script/release/ci-release-build.js
@@ -235,11 +235,14 @@ async function callAppVeyor (targetBranch, job, options) {
     method: 'POST'
   };
   jobRequestedCount++;
-  const appVeyorResponse = await makeRequest(requestOpts, true).catch(err => {
-    console.log('Error calling AppVeyor:', err);
-  });
-  const buildUrl = `https://ci.appveyor.com/project/electron-bot/${appVeyorJobs[job]}/build/${appVeyorResponse.version}`;
-  console.log(`AppVeyor release build request for ${job} successful.  Check build status at ${buildUrl}`);
+
+  try {
+    const { version } = await makeRequest(requestOpts, true);
+    const buildUrl = `https://ci.appveyor.com/project/electron-bot/${appVeyorJobs[job]}/build/${version}`;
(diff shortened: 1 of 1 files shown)


VerteDinde (approved):
Thanks! 🙇‍♀️

ckerr (approved):
LGTM

MarshallOfSound (changes requested):
This will silently absorb the errors here and make sudowoodo think the script succeeded right?

codebytere:
@MarshallOfSound the errors were already being caught if you look at the old code - the thing i'm fixing is that the lines after assume success

MarshallOfSound:
@codebytere I might be missing something but in the case where `makeRequest` throws an error the `console.log` for the success state wouldn't run.  So this would only catch errors in the actual console.log call which I don't understand how they would happen

clavin (approved):
The original code is a little deceptive in that the `.catch()` on the promises mean that the `await` will never throw:

```js
await Promise.reject(new Error('oops')); // throws here
console.log('after'); // NOT printed
```

```js
await Promise.reject(new Error('oops')).catch(err => console.log('caught', err));
console.log('after'); // always printed
```

The original code is structured as `await`ing some promise returned from `.catch()` and then doing something with the returned value (which is `undefined` when that promise is rejected and caught by `.catch()`):

```js
// this `await` never throws in the way you would expect from `await`:
const something = await makeRequest(...).catch(err => {
    console.log('oops', err);
});

// which means this code will always run, sometimes with `something = undefined`:
const buildUrl = `https://example.ci/${something}`; // uh oh
console.log(`Example release build request for ${job} successful.  Check build status at ${buildUrl}`);
```

The new code removes the `.catch()` calls to make the code more clear and actually sequence correctly when the `makeRequest` promise is rejected.

jkleinsc:
@codebytere @MarshallOfSound iirc, sudowoodo looks for the message `AppVeyor release build request for ${job} successful.`  to determine if a job call was successful so I think this change is ok.

MarshallOfSound (approved):

PR #28652, opened 2021-04-14
Update native-file-drag-drop.md

dengjie13520:
electronic 12 + needs to set contextisolation to false

### DIFF
--- a/docs/tutorial/native-file-drag-drop.md
+++ b/docs/tutorial/native-file-drag-drop.md
@@ -33,6 +33,7 @@ document.getElementById('drag').ondragstart = (event) => {
   ipcRenderer.send('ondragstart', '/absolute/path/to/the/item')
 }
 ```
+Note: electronic 12 + needs to set contextisolation to false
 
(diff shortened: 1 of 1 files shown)


MarshallOfSound (changes requested):
The example should be updated to work with `contextIsolation: true` rather than suggesting an insecure option be used

miniak:
@dengjie13520 also FYI it is Electron not "electronic"

codebytere:
@dengjie13520 would you like to move this PR forward?

docs/tutorial/native-file-drag-drop.md
```diff
@@ -33,6 +33,7 @@ document.getElementById('drag').ondragstart = (event) => {
   ipcRenderer.send('ondragstart', '/absolute/path/to/the/item')
 }
 ```
+Note: electronic 12 + needs to set contextisolation to false
```
GrahamSH-LLK, line 36: You mean Electron, afaik. Also, there should be a period at the end.

codebytere:
Closing as no response has been given. Happy to consider re-opening though.

PR #28706, opened 2021-04-16
feat: add WebFrameMain.visibilityState

samuelmaddock:
#### Description of Change

Adds `WebFrameMain.visibilityState` instance property which aligns with `document.visibilityState` in the renderer.

This can be useful for telling whether a frame is visible without needing to execute any JS in the renderer.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added `WebFrameMain.visibilityState` instance property.

### DIFF
--- a/spec-main/api-web-frame-main-spec.ts
+++ b/spec-main/api-web-frame-main-spec.ts
@@ -143,6 +143,11 @@ describe('webFrameMain module', () => {
 
       expect(webFrame.visibilityState).to.equal('visible');
       w.hide();
+
+      // Wait for visibility to propagate. If this ends up being flaky, we can
+      // look into binding WebContentsObserver::OnVisibilityChanged.
+      await new Promise(resolve => setTimeout(resolve, 10));
+
       expect(webFrame.visibilityState).to.equal('hidden');
     });
   });
--- a/spec-main/api-web-frame-main-spec.ts
+++ b/spec-main/api-web-frame-main-spec.ts
@@ -6,6 +6,7 @@ import { BrowserWindow, WebFrameMain, webFrameMain, ipcMain } from 'electron/mai
 import { closeAllWindows } from './window-helpers';
 import { emittedOnce, emittedNTimes } from './events-helpers';
 import { AddressInfo } from 'net';
+import { waitForTrue } from './spec-helpers';
 
 describe('webFrameMain module', () => {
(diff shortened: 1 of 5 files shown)


miniak (approved):

jkleinsc (approved):

jkleinsc (changes requested):
Looks like newly added test is failing on macOS: https://app.circleci.com/pipelines/github/electron/electron/38127/workflows/4af17c41-7d17-47c2-b050-a1c088106dec/jobs/840026

samuelmaddock:
> Looks like newly added test is failing on macOS: https://app.circleci.com/pipelines/github/electron/electron/38127/workflows/4af17c41-7d17-47c2-b050-a1c088106dec/jobs/840026

Thanks, I'll investigate. All of the Goma builds were failing previously.

nornagon (approved):

nornagon:
API LGTM

samuelmaddock:
I've added a delay in the test to fix the issue on macOS. ec9fe90371dc1303de56ead4fc002e3c8cea0b92

I'm not happy with this, but I don't think there's any other reliable way to test it without binding another event. Let me know whether this is okay.

spec-main/api-web-frame-main-spec.ts
```diff
@@ -143,6 +143,11 @@ describe('webFrameMain module', () => {
 
       expect(webFrame.visibilityState).to.equal('visible');
       w.hide();
+
```
jkleinsc, line 146: Instead of an arbitrary timeout, you could do something like the following:
```
  async function waitForVisibilityState (webFrame, expectedState) {
    let state = webFrame.visibilityState;
    while (webFrame.visibilityState !== expectedState) {
      await new Promise(resolve => setTimeout(resolve, 10));
      state = webFrame.visibilityState;
    }
    return state;
  }
  expect(waitForVisibilityState(webFrame, 'hidden')).to.eventually.equal('hidden');
```

jkleinsc (changes requested):
Relying on a timeout will produce flaky results but we can use something like I've suggested here to effectively wait for the change.

spec-main/api-web-frame-main-spec.ts
samuelmaddock: I've added a helper for this kind of behavior 9c5ae84

jkleinsc (approved):

jkleinsc:
API LGTM

PR #28712, opened 2021-04-18
fix: ensure the typescript definitions only export correct value types

### LINKED ISSUE #22167: TS definitions of NativeImage static methods are incorrect
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 8.0.0
* **Operating System:**
  * any
  * (not sure)

### Expected Behavior
```js
import { NativeImage } from 'electron'
const nativeImage = NativeImage.createFromPath(imagePath)
```
**Type checking should warn that NativeImage in undefined.**

### Discussion
In `electron.d.ts`, `createFromPath` and other methods are defined as static methods of `NativeImage`, but the exported module is `nativeImage`, not `NativeImage`. 

Typescript is probably confused with this:
image

The workaround is currently to:
```js
import { nativeImage as NativeImage } from 'electron'
```

If I'm understanding right, `electron.d.ts` is generated with https://github.com/electron/typescript-definitions from the json api documentation, so the issue might live there.

MarshallOfSound:
In typescript there are two main types of "types" you can export, value types (class, const) and definition types (type, interface).  The typescript compiler will let anything declared via const or class be used as a value.  Unfortunately we were exporting a bunch of things (see the diff) as class/const when they weren't actually exported values. This lead to typescript being happy but the runtime throwing errors (not something we want).

This change passes "exported-in" context through our docs, to the parser and then to the definitions generator to ensure we only mark things as exported in the ts defs that we actually export.

Fixes #22167

This PR is currently targeting `update-ts-parser` as it relies on some of those changes and this shows the actual diff.

I've marked this as `semver/patch` but out of an abundance of caution this shouldn't be backported

Notes: Fixed typescript export definitions such that not-exported values are not declared in the exports interface

### DIFF
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -126,7 +126,7 @@ state is `hidden` in order to minimize power consumption.
 
 > Create and control browser windows.
 
-Exported in: [Main](../glossary.md#main-process)
+Process: [Main](../glossary.md#main-process)
 
 `BrowserWindow` is an [EventEmitter][event-emitter].
 
--- a/docs/api/browser-window.md
+++ b/docs/api/browser-window.md
@@ -126,7 +126,7 @@ state is `hidden` in order to minimize power consumption.
 
 > Create and control browser windows.
 
-Process: [Main](../glossary.md#main-process)
+Exported in: [Main](../glossary.md#main-process)
 
 `BrowserWindow` is an [EventEmitter][event-emitter].
 
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -22,5 +22,8 @@
   "no-trailing-spaces": {
     "br_spaces": 0
   },
-  "single-h1": false
+  "single-h1": false,
+  "no-inline-html": {
+    "allowed_elements": ["br"]
+  }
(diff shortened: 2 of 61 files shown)


nornagon (changes requested):
I think "Exported in" is more confusing than "Process". Can we keep the previous verbiage?

MarshallOfSound:
@nornagon We have to distinguish between a module being "available" in a process and a module being "exported" in the `electron` module in that process.  If you have an alternative proposal for the words to use to distinguish that I'm open to suggestions, but we can't use the same verbiage in both cases.

docs/api/browser-window.md
```diff
@@ -2,7 +2,7 @@
 
 > Create and control browser windows.
 
-Process: [Main](../glossary.md#main-process)
+Export in: [Main](../glossary.md#main-process)
```
nornagon, line 5: Was this left by mistake?

MarshallOfSound: Oh lol, was actually a typo lol

nornagon (approved):

PR #28729, opened 2021-04-19
build: read node files as binary files

jkleinsc:
#### Description of Change

Recently our releases started failing while uploading the node checksums due to the following error:
 "UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8b in position 1: invalid start byte\n"

This error was happening because the python script was trying to read the node files as utf-8 instead of binary.  This PR updates that python script to instead read those files as binary.

#### Checklist

#### Release Notes

Notes: none


VerteDinde (approved):
Thanks so much for finding a fix for this, and so quickly!

MarshallOfSound (approved):

Issue #28718, opened 2021-04-19
[Bug]: option title for the showSaveDialog is not displayed on Ubuntu


Tabea4:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/electron/electron/issues) for a feature request that matches the one I want to file, without success.

### Electron Version

11.1.1

### What operating system are you using?

Ubuntu

### Operating System Version

20.04.

### What arch are you using?

x64

### Expected Behavior

Hi, 
I’m currently using Electron version 11.1.1 and I’m working on Ubuntu 20.04. I have a question concerning the method `showSaveDialog` (https://www.electronjs.org/docs/api/dialog#dialogshowsavedialogbrowserwindow-options):
On my OS, the title is not displayed in the dialog, even though it is set in the code.
For the `showOpenDialog` however, the title is displayed. (Also see attached screenshots)
Does `showSaveDialog` not support displaying the title, or is this only an issue on Ubuntu?

Thank you for your help!

ShowOpenDialog: title (`Select connection matrix to load`) is displayed as expected
![showOpenDialog](https://user-images.githubusercontent.com/34033202/115217328-b051ee80-a105-11eb-86a0-771402f8883d.jpg)
ShowSaveDialog: no title is displayed even though it is set in the code
![showSaveDialog](https://user-images.githubusercontent.com/34033202/115217333-b1831b80-a105-11eb-85b0-c40804b8f043.jpg)

### Actual Behavior

Title is not displayed (see above screenshot from `showSaveDialog`)

Prinzhorn:
I'm using Ubuntu myself and I'm curious if this is just a limitation of the GtkFileChooserDialog? Where would the title be displayed when the input for the file name takes up the place?

Tabea4:
@Prinzhorn Thanks for the hint. It seems like it is indeed a limitation of the GtkFileChooserDialog. Do you happen to know if the same “issue” exists on other desktop environments and/or other operating systems? 
[...]

PR #28747, opened 2021-04-20
docs: option title for showSaveDialog is not supported for some platforms

Tabea4:
see #28718

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/docs/api/dialog.md
+++ b/docs/api/dialog.md
@@ -154,7 +154,7 @@ dialog.showOpenDialog(mainWindow, {
 
 * `browserWindow` [BrowserWindow](browser-window.md) (optional)
 * `options` Object
-  * `title` String (optional)
+  * `title` String (optional) - Title of the dialog, some platforms will not show it.
   * `defaultPath` String (optional) - Absolute directory path, absolute file
     path, or file name to use by default.
   * `buttonLabel` String (optional) - Custom label for the confirmation button, when
@@ -185,7 +185,7 @@ The `filters` specifies an array of file types that can be displayed, see
 
 * `browserWindow` [BrowserWindow](browser-window.md) (optional)
 * `options` Object
-  * `title` String (optional)
+  * `title` String (optional) - Title of the dialog, some platforms will not show it.
   * `defaultPath` String (optional) - Absolute directory path, absolute file
     path, or file name to use by default.
   * `buttonLabel` String (optional) - Custom label for the confirmation button, when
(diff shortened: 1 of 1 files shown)


miniak:
@Tabea4 instead of "some platforms will not show it" we should say which ones

docs/api/dialog.md
```diff
@@ -185,7 +185,7 @@ The `filters` specifies an array of file types that can be displayed, see
 
 * `browserWindow` [BrowserWindow](browser-window.md) (optional)
 * `options` Object
-  * `title` String (optional)
+  * `title` String (optional) - Title of the dialog, some platforms will not show it.
```
codebytere, line 188: This should specify the platforms on which it works. 

See
https://github.com/electron/electron/blob/2865d161b720e209f29ca84858492e095dc3e161/docs/api/dialog.md#L96-L103 for prior art.

codebytere (changes requested):

Tabea4:
@miniak @codebytere 
I think that this is not necessarily only a matter of the platform, but of the desktop environment. I’m on Ubuntu using the Gnome desktop environment. I don’t have the possibility to check this on every desktop environment, that’s why I just wanted to add a note, so that developers know that this might not work on their machine. This is similar to the option `title` of the `showMessageBoxSync` : https://www.electronjs.org/docs/api/dialog#dialogshowmessageboxsyncbrowserwindow-options, where it also only says `some platforms will not show it.`

zcbenz (commented):
I think we could take this if it says "on Linux some desktop environments will not show it".

Tabea4:
> I think we could take this if it says "on Linux some desktop environments will not show it".

I changed it accordingly.

zcbenz (approved):

docs/api/dialog.md
```diff
@@ -154,7 +154,7 @@ dialog.showOpenDialog(mainWindow, {
 
 * `browserWindow` [BrowserWindow](browser-window.md) (optional)
 * `options` Object
-  * `title` String (optional)
+  * `title` String (optional) - Title of the dialog, on _Linux_ some desktop environments will not show it.
```
codebytere, line 157: Let's clean this phrasing up a bit? 

> `title` String (optional) - The dialog title. Cannot be displayed on some Linux desktop environments.

docs/api/dialog.md
```diff
@@ -185,7 +185,7 @@ The `filters` specifies an array of file types that can be displayed, see
...
```
codebytere, line 188: Same as above.

docs/api/dialog.md
Tabea4: Changed it as suggested

docs/api/dialog.md
Tabea4: Changed it as well

codebytere (approved):

PR #25114, opened 2020-08-24
feat: add shell.trashItem() to replace shell.moveItemToTrash()

nornagon:
#### Description of Change
Closes #11555.

#### Checklist

- [x] PR description included and stakeholders cc'd
- [x] `npm test` passes
- [x] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [x] relevant documentation is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)
- [x] [PR release notes](https://github.com/electron/clerk/blob/master/README.md) describe the change in a way relevant to app developers, and are [capitalized, punctuated, and past tense](https://github.com/electron/clerk/blob/master/README.md#examples).
- [x] This is **NOT A BREAKING CHANGE**. Breaking changes may not be merged to master until 11-x-y is branched.

#### Release Notes

Notes: Added a new asynchronous `shell.trashItem()` API, replacing the synchronous `shell.moveItemToTrash()`.

PR #28748, opened 2021-04-20
fix: shell.trashItem crash when called in renderer

### LINKED ISSUE #28029: shell.trashItem crashes the process
### Preflight Checklist

### Issue Details

* **Electron Version:**
  * 12.0.0
* **Operating System:**
  * macOS 10.14
  * N/A

### Expected Behavior

`shell.trashItem` should just work.

### Actual Behavior

It does its job, and then immediately crashes the process.

### To Reproduce

I don't have a fiddle, but this line makes the process crash for me:

```js
return shell.trashItem ( path ).then ( () => true, () => false );
```

While this works:

```js
return Promise.resolve ( shell.moveItemToTrash ( path, false ) );
```

In addition I'm getting as output from the main process:

```
{ reason: 'crashed', exitCode: 11 }
{ reason: 'crashed', exitCode: 5 }
{ reason: 'crashed', exitCode: 5 }
{ reason: 'crashed', exitCode: 5 }
{ reason: 'crashed', exitCode: 5 }
{ reason: 'crashed', exitCode: 5 }
{ reason: 'crashed', exitCode: 5 }
{ reason: 'crashed', exitCode: 5 }
and so on...
```

Basically it crashes the window with exit code 11, then my app opens gain the window after it crashed, and it crashes again immediately with exit code 5.

nornagon:
#### Description of Change
Fixes #28029. This has been broken in the renderer since it was introduced (by me 😳 ) in #25114.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed a crash when calling `shell.trashItem()` from the renderer process.

### DIFF
--- a/spec-main/api-shell-spec.ts
+++ b/spec-main/api-shell-spec.ts
@@ -80,7 +80,7 @@ describe('shell module', () => {
     it('works in the renderer process', async () => {
       const w = new BrowserWindow({ show: false, webPreferences: { nodeIntegration: true, contextIsolation: false } });
       w.loadURL('about:blank');
-      await expect(w.webContents.executeJavaScript('require(\'electron\').shell.trashItem(\'does-not-exist\')')).to.be.rejectedWith(/does-not-exist|Failed to move item/);
+      await expect(w.webContents.executeJavaScript('require(\'electron\').shell.trashItem(\'does-not-exist\')')).to.be.rejectedWith(/does-not-exist|Failed to move item|Failed to create FileOperation/);
     });
   });
 });
--- a/spec-main/api-shell-spec.ts
+++ b/spec-main/api-shell-spec.ts
@@ -62,6 +62,8 @@ describe('shell module', () => {
   });
 
   describe('shell.trashItem()', () => {
+    afterEach(closeAllWindows);
+
     it('moves an item to the trash', async () => {
       const dir = await fs.mkdtemp(path.resolve(app.getPath('temp'), 'electron-shell-spec-'));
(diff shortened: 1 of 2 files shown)


codebytere (approved):

jkleinsc (changes requested):
Looks like newly added test is failing on all Windows platforms.

spec-main/api-shell-spec.ts
```diff
@@ -74,5 +76,11 @@ describe('shell module', () => {
       const filename = path.join(app.getPath('temp'), 'does-not-exist');
       await expect(shell.trashItem(filename)).to.eventually.be.rejected();
     });
+
+    it('works in the renderer process', async () => {
+      const w = new BrowserWindow({ show: false, webPreferences: { nodeIntegration: true, contextIsolation: false } });
+      w.loadURL('about:blank');
+      await expect(w.webContents.executeJavaScript('require(\'electron\').shell.trashItem(\'does-not-exist\')')).to.be.rejectedWith(/does-not-exist|Failed to move item/);
```
nornagon, line 83:
```suggestion
      await expect(w.webContents.executeJavaScript('require(\'electron\').shell.trashItem(\'does-not-exist\')')).to.be.rejectedWith(/does-not-exist|Failed to move item|Failed to create FileOperation/);
```

zcbenz:
@nornagon The newly added test seems to fail consistently on ia32 Windows.

nornagon:
@zcbenz thanks for the ping. I have no idea why this test would fail on ia32 windows but not x64 windows, so for now I'm just going to disable the test to unbreak master.

PR #28766, opened 2021-04-21
build: disable pseudolocales in Electron

codebytere:
#### Description of Change

Remove pseudocode locales from Electron, which were added in https://chromium-review.googlesource.com/c/chromium/src/+/2774898 and rolled in https://github.com/electron/electron/pull/28462. We don't use these.  

See https://electronhq.slack.com/archives/CC8J5QC00/p1618961550011900?thread_ts=1618932845.005500&cid=CC8J5QC00.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none

### DIFF
--- a/build/args/all.gn
+++ b/build/args/all.gn
@@ -22,7 +22,8 @@ dawn_enable_vulkan_validation_layers = false
 # This breaks native node modules
 libcxx_abi_unstable = false
 
-# Electron does not use these.
+# These are disabled because they cause the zip manifest to differ between
+# testing and release builds.
 # See https://chromium-review.googlesource.com/c/chromium/src/+/2774898.
 enable_pseudolocales = false
 
--- a/build/args/all.gn
+++ b/build/args/all.gn
@@ -22,4 +22,8 @@ dawn_enable_vulkan_validation_layers = false
 # This breaks native node modules
 libcxx_abi_unstable = false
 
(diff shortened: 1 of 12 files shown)


shell/browser/native_window_mac.mm
```diff
@@ -330,8 +330,8 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
     styleMask |= NSWindowStyleMaskClosable;
   if (resizable_)
     styleMask |= NSResizableWindowMask;
-  if (!useStandardWindow || transparent() || !has_frame())
```
MarshallOfSound, line 333: Oops?

MarshallOfSound (approved):

shell/browser/native_window_mac.mm
codebytere: already removed lol

build/args/all.gn
```diff
@@ -22,4 +22,8 @@ dawn_enable_vulkan_validation_layers = false
 # This breaks native node modules
 libcxx_abi_unstable = false
 
+# Electron does not use these.
```
nornagon, line 25:
```suggestion
# These are disabled because they cause the zip manifest to differ between
# testing and release builds.
```

jkleinsc (approved):

PR #28769, opened 2021-04-21
refactor: clean up unused parts of the render_frame_host patch file

MarshallOfSound:
We can **finally** delete 95% of this awful patch file, the only thing left behind is `RegisterPendingSiteInstance` which is super minimal and is currently used for our process ID --> webContents mapping logic.

Notes: none


nornagon (approved):
omg HECK YEA

PR #28773, opened 2021-04-22
fix: handle async nature of [NSWindow -toggleFullScreen]

codebytere:
Backport of #25470.

See that PR for details.

Notes: Fixed an issue where multiple calls to `window.setFullScreen` could cause problems.


jkleinsc (changes requested):
Looks like there is a reference to the removed `exiting_fullscreen_`:
https://app.circleci.com/pipelines/github/electron/electron/38352/workflows/42ae2e12-8c0f-42f8-953d-97827e9dfe74/jobs/845355

jkleinsc (approved):

PR #28786, opened 2021-04-22
refactor: remove ProcessPreferences

nornagon:
#### Description of Change
This removes the ProcessPreferences class which used a flawed mapping from
process ID to WebContents to determine various facts about how that WebContents
was launched. A single process can be involved in multiple WebContents, so this
logic was incorrect. Instead, look up the WebContents / WebPreferences directly
associated with the requests in question.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: none


MarshallOfSound (approved):
😍

Issue #28338, opened 2021-03-22
[Bug]: clicking on a link before loadURL has resolved causes the promise to reject, even if the link navigation was prevented


pushkin-:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/electron/electron/issues) for a feature request that matches the one I want to file, without success.

### Electron Version

12.0.1, 11.2.1

### What operating system are you using?

Windows

### Operating System Version

Windows 10 v1909

### What arch are you using?

x64

### Expected Behavior

I expect that calling `preventDefault` in the `will-navigate` handler will prevent us from attempting the navigation

### Actual Behavior

This doesn't seem to happen if we're in the middle of a `loadURL` call

this issue may very well be a dupe of [this one](https://github.com/electron/electron/issues/17526) and related to [this one](https://github.com/electron/electron/issues/19847) though I'm not calling `loadURL` twice.

I'm experiencing an issue in my app where if there's network delay, I make a `loadURL` and click a link on the main page before `loadURL` has finished, we get an `ERR_ABORTED` error. From reading in other issues, this is expected when a navigation happens in the middle of a navigation.

So I attempted to solve this by canceling the navigation in the `will-navigate` handler when the main window's webcontents are still loading (`isLoadingMainFrame()`), however, this doesn't solve the crash. The navigation is prevented, but `WebContents::did-start-navigation` has already fired for whatever reason (I'd expect it to fire *after* `will-navigate`), and we get the crash

1. add network delay. What I did was start Fiddler, Rules -> Performance -> Simulate Modem speeds
2. start main.js
3. Click a link (like the "Sign Up" link)
4. observe crash:

![image](https://user-images.githubusercontent.com/6374473/112063138-f0d24280-8b2e-11eb-9789-291049488711.png)

main.js:
```
(async () => {
	const { app, BrowserWindow, dialog } = require("electron");
	app.on("window-all-closed", async () => {
		app.quit();
	});

	await app.whenReady();
	const mainWindow = new BrowserWindow();
	mainWindow.webContents.on("will-navigate", (e) => {
		if (mainWindow.webContents.isLoadingMainFrame()) {
			e.preventDefault();
		}
	});
	try {
[...]

PR #28839, opened 2021-04-26
refactor: remove internal navigation controller

MarshallOfSound:
There's an unanswered question here with a potential change in "reload" behavior.  It's a toss up between "the old behavior was a bug" and "the new behavior is potentially unsafe / also a bug".  See the implementation of `WebContents::Reload`

Notes: no-notes


shell/browser/api/electron_api_web_contents.h
```diff
@@ -154,17 +154,26 @@ class WebContents : public gin::Wrappable<WebContents>,
...
+  void Reload();
+  void ReloadIgnoringCache();
   void DownloadURL(const GURL& url);
   GURL GetURL() const;
   std::u16string GetTitle() const;
   bool IsLoading() const;
   bool IsLoadingMainFrame() const;
   bool IsWaitingForResponse() const;
   void Stop();
-  void ReloadIgnoringCache();
```
MarshallOfSound, line 164: So I found out while doing this that this header declaration has existed for 6 years and pointed at, well, nothing lol

pushkin-:
Not to take over your refactor PR, but [here's](https://github.com/electron/electron/issues/28338) an issue that could maybe be fixed up that relates to the navigation controller :)

MarshallOfSound:
@pushkin- That issue partially seems like expected behavior to me and is also unrelated to the navigation controller implementation.  Specifically the code that powers that is the implementation of `loadURL` which is Electron specific and independent of the chromium navigation controller.

deepak1556 (approved):

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2007,18 +2023,64 @@ void WebContents::Stop() {
...
 }
 
+bool WebContents::CanGoBack() const {
+  return web_contents()->GetController().CanGoBack();
+}
+
 void WebContents::GoBack() {
+  if (!CanGoBack())
+    return;
   web_contents()->GetController().GoBack();
```
nornagon, line 2033: nit: 
```suggestion
  if (CanGoBack())
    web_contents()->GetController().GoBack();
```

reads a bit easier. same for the other methods here.

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -2007,18 +2023,64 @@ void WebContents::Stop() {
...
 void WebContents::GoToOffset(int offset) {
+  if (!CanGoToOffset(offset))
+    return;
   web_contents()->GetController().GoToOffset(offset);
 }
 
+bool WebContents::CanGoToIndex(int index) const {
+  if (index < 0)
+    return false;
+  return index < GetHistoryLength();
```
nornagon, line 2059:
```suggestion
  return index >= 0 && index < GetHistoryLength();
```

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -1973,6 +1976,23 @@ void WebContents::LoadURL(const GURL& url,
   NotifyUserActivation();
 }
 
+// TODO(MarshallOfSound): Figure out what we need to do with post data here, I
+// believe the default behavior when we pass "true" is to phone out to the
+// delegate and then the controller expects this method to be called again with
+// "false" if the user approves the reload.  For now this would result in
+// ".reload()" calls on POST data domains failing silently.  Passing false would
+// result in them succeeding, but reposting which although more correct could be
+// considering a breaking change.
```
nornagon, line 1985: should we expose this as a param?

MarshallOfSound: I think we should, but that also (for correct implementations) would require us exposing the delegate event `did-get-post-data-reload-request` or something 🤔  I think for now defaulting to the safe value and exposing the new API later is OK.

Thoughts?

nornagon: sgtm

PR #28858, opened 2021-04-27
test: worker should be able to load asar files

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/28572.

Add test that Worker and SharedWorker should be able to load asar files passed via `file://` urls.

#### Release Notes

Notes: none


codebytere (approved):

jkleinsc (approved):

jkleinsc:
Merging as CI failure is known issue unrelated to this PR.

PR #28883, opened 2021-04-28
build: remove the vendor directory

MarshallOfSound:
The `vendor` directory is a remnant of times long since gone when we had brightray and libchromiumcontent (never forget).  It is currently the hope to a single header file that afaics we do not need anymore and randomly pyyaml gets cloned there but that is easily movable.

This commit removes all references to the vendor directory from scripts, docs, helpers, etc. and removes the directory itself.

Notes: none


codebytere (approved):

MarshallOfSound:
Not a clean backport and there's no benefit to backporting this so just gonna leave it in master

Issue #28436, opened 2021-03-30
[Bug]: Electron propagates GDK_BACKEND=x11 to subprocesses, breaking shell.openExternal with Firefox on Wayland


andersk:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/electron/electron/issues) for a feature request that matches the one I want to file, without success.

### Electron Version

12.0.9
13.0.1
14.1.0
15.1.1

### What operating system are you using?

Other Linux

### Operating System Version

NixOS 21.05

### What arch are you using?

x64

### Last Known Working Electron version

12.0.0-beta.20

### Expected Behavior

`shell.openExternal("https://electronjs.org/")` opens a browser tab at https://electronjs.org/. (My browser is Firefox on Wayland with `MOZ_ENABLE_WAYLAND=1`, and has other tabs already open.)

### Actual Behavior

Firefox shows this error and does not open a tab:

“Firefox is already running, but is not responding. To use Firefox, you must first close the existing Firefox process, restart your device, or use a different profile.”

### Testcase Gist URL

https://gist.github.com/31ec2165f4aa806536e11d7d99d1e26f

I bisected this to v12.0.0-beta.20...v12.0.0-beta.21 (9faf23509d054062def1332500c05c8ee7fb9bb7...5dbb6356effbfc0f6e0528583fae832290c14207).

The difference between v12.0.0-beta.20’s working invocation of `xdg-open` and v12.0.0-beta.21’s failing invocation is that the latter has `GDK_BACKEND=x11` in the environment. I confirmed that this is responsible for the failure with manual invocations of

```sh
xdg-open https://electronjs.org/  # works
env GDK_BACKEND=x11 xdg-open https://electronjs.org/  # fails
```

I don’t know if Electron or Chromium need to set `GDK_BACKEND=x11` for their own purposes, but it definitely should not be propagated to subprocesses such as `xdg-open`.

codebytere:
Looks like this was added by https://chromium-review.googlesource.com/c/chromium/src/+/2586184

deepak1556:
[...]

PR #28898, opened 2021-04-28
fix: don't propagate GDK_BACKEND to subprocs

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/28436.

`GDK_BACKEND` is set by Chromium as of [this CL](https://chromium-review.googlesource.com/c/chromium/src/+/2586184) and can detrimentally affect external app behaviors, so we want to unset it and replace with the user's setting if one exists.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where `GDK_BACKEND` was being propagated to subprocesses on Linux.

### DIFF
--- a/shell/browser/electron_browser_main_parts.cc
+++ b/shell/browser/electron_browser_main_parts.cc
@@ -156,7 +156,7 @@ std::u16string MediaStringProvider(media::MessageId id) {
 
 #if defined(OS_LINUX)
 
-std::string gdk_backend;
+const char* gdk_backend;
 
 void OverrideLinuxAppDataPath() {
   base::FilePath path;
@@ -380,7 +380,7 @@ void ElectronBrowserMainParts::PostDestroyThreads() {
 
 #if defined(OS_LINUX)
 // static
-std::string ElectronBrowserMainParts::GetGDKBackend() {
+const char* ElectronBrowserMainParts::GetGDKBackend() {
   return gdk_backend;
 }
 #endif
@@ -391,7 +391,9 @@ void ElectronBrowserMainParts::ToolkitInitialized() {
   // https://chromium-review.googlesource.com/c/chromium/src/+/2586184
   // and can detrimentally affect external app behaviors, so we want to
   // check if the user has set it so we can use it later.
-  base::Environment::Create()->GetVar("GDK_BACKEND", &gdk_backend);
+  std::string backend;
+  if (base::Environment::Create()->GetVar("GDK_BACKEND", &backend))
+    gdk_backend = backend.c_str();
 
   auto linux_ui = BuildGtkUi();
   linux_ui->Initialize();
--- a/shell/browser/electron_browser_main_parts.cc
+++ b/shell/browser/electron_browser_main_parts.cc
@@ -155,6 +155,9 @@ std::u16string MediaStringProvider(media::MessageId id) {
 }
 
 #if defined(OS_LINUX)
+
+std::string gdk_backend;
+
 void OverrideLinuxAppDataPath() {
   base::FilePath path;
   if (base::PathService::Get(DIR_APP_DATA, &path))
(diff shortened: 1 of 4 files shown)


MarshallOfSound (approved):

shell/common/platform_util_linux.cc
```diff
@@ -122,6 +122,19 @@ bool XDGUtil(const std::vector<std::string>& argv,
   // bring up a new terminal if necessary.  See "man mailcap".
   options.environment["MM_NOTTTY"] = "1";
 
+  // This is set by Chromium here:
+  // https://chromium-review.googlesource.com/c/chromium/src/+/2586184
+  // and can detrimentally affect external app behaviors, so we want to
+  // unset it and replace with the user's setting if one exists.
+  std::string gdk_backend;
+  if (base::Environment::Create()->GetVar("GDK_BACKEND", &gdk_backend)) {
```
deepak1556, line 130: This will only get the value set by chromium, as `base::Environment` uses `[get/set]env` api internally. We should cache the value of `GDK_BACKEND` before we call `BuildGtkUi` https://github.com/electron/electron/blob/c37c9adbe2fc457d02396f4c02ac536a60a080ca/shell/browser/electron_browser_main_parts.cc#L380, so that we get expected value set by the user.

shell/common/platform_util_linux.cc
```diff
@@ -122,6 +123,14 @@ bool XDGUtil(const std::vector<std::string>& argv,
   // bring up a new terminal if necessary.  See "man mailcap".
   options.environment["MM_NOTTTY"] = "1";
 
+  // If the user set a GDK_BACKEND value of their own, use that,
+  // otherwise unset it. Setting values in EnvironmentMap to an empty-string
+  // will make sure that they get unset from the environment via
+  // AlterEnvironment().
```
zcbenz, line 129: The comment should mention that we are doing this because Chromium is setting `GDK_BACKEND` during GTK initialization.

shell/browser/electron_browser_main_parts.cc
```diff
@@ -375,8 +378,23 @@ void ElectronBrowserMainParts::PostDestroyThreads() {
...
+
 void ElectronBrowserMainParts::ToolkitInitialized() {
 #if defined(OS_LINUX)
+  // This is set by Chromium here:
+  // https://chromium-review.googlesource.com/c/chromium/src/+/2586184
+  // and can detrimentally affect external app behaviors, so we want to
+  // check if the user has set it so we can use it later.
+  std::string backend;
+  if (base::Environment::Create()->GetVar("GDK_BACKEND", &backend))
+    gdk_backend = backend.c_str();
```
zcbenz, line 396: `backend` is a local variable and once `ToolkitInitialized ` finishes the `const char*` will point to freed memory. A common pattern for this kind of cache is:

```c++
base::Optional<std::string>& ElectronBrowserMainParts::GetGDKBackend() {
  static base::Optional<std::string> gdk_backend;
  return gdk_backend;
}

void ElectronBrowserMainParts::ToolkitInitialized() {
  ...
  GetGDKBackend().reset(backend);
  ...
}
```

zcbenz (changes requested):
Can you also add test to check if `GDK_BACKEND` has actually been cleared? It is possible that Chromium decides to move the code earlier and invalidates our workaround.

shell/browser/electron_browser_main_parts.cc
```diff
@@ -375,8 +376,24 @@ void ElectronBrowserMainParts::PostDestroyThreads() {
   fake_browser_process_->PostDestroyThreads();
 }
 
+#if defined(OS_LINUX)
+// static
+base::Optional<std::string> ElectronBrowserMainParts::GetGDKBackend() {
```
zcbenz, line 381: It should return `base::Optional<std::string>&` otherwise it could not be modified.

shell/common/platform_util_linux.cc
```diff
@@ -122,6 +123,18 @@ bool XDGUtil(const std::vector<std::string>& argv,
   // bring up a new terminal if necessary.  See "man mailcap".
   options.environment["MM_NOTTTY"] = "1";
 
+  // If the user set a GDK_BACKEND value of their own, use that,
+  // otherwise unset it becuase Chromium is setting GDK_BACKEND
+  // during GTK initialization and we want to respect user preference.
+  // Setting values in EnvironmentMap to an empty-string
+  // will make sure that they get unset from the environment via
+  // AlterEnvironment().
+  base::Optional<std::string> gdk_backend =
```
zcbenz, line 132: Using `const base::Optional<std::string>& gdk_backend = ` can avoid unnecessary copy.

zcbenz (commented):
The code looks good to me, I think we just need a test to check if `GDK_BACKEND` has actually been cleared.

spec-main/chromium-spec.ts
```diff
@@ -367,6 +367,13 @@ describe('web security', () => {
   });
 });
 
+ifdescribe(process.platform === 'linux')('subprocesses', () => {
+  it('does not propagate GDK_BACKEND', () => {
+    const backend = process.env.GDK_BACKEND;
```
zcbenz, line 372: I think we should start a child app and explicitly set its environment variable to do the test, we don't know if the test environment has the `GDK_BACKEND` set.

zcbenz (approved):

PR #28909, opened 2021-04-28
chore: Browser::SetAppUserModelID is Windows only

miniak:
#### Description of Change
After #28127 this is Windows only https://github.com/electron/electron/blob/56c3103e73ac71bd6845a31a2e37f5ffd0810be7/shell/browser/api/electron_api_app.cc#L1538-L1541

#### Checklist
- [x] `npm test` passes

#### Release Notes
Notes: no-notes

### DIFF
--- a/shell/browser/browser.h
+++ b/shell/browser/browser.h
@@ -83,8 +83,10 @@ class Browser : public WindowListObserver {
   // Clear the recent documents list.
   void ClearRecentDocuments();
 
+#if defined(OS_WIN)
   // Set the application user model ID.
   void SetAppUserModelID(const std::wstring& name);
+#endif
 
   // Remove the default protocol handler registry key
   bool RemoveAsDefaultProtocolClient(const std::string& protocol,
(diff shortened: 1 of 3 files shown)


codebytere (approved):

MarshallOfSound (changes requested):
@nornagon you originally landed this https://github.com/electron/electron/commit/640a1451126b404a6a9557c6c080cc83d51805de

It doesn't have a ref but this is a breaking change (previously there was a noop provided on other platforms).  Can we get insight into why the change was made?

nornagon:
@MarshallOfSound because wstrings are windows only, so it wasn't compiling.

MarshallOfSound:
Gotcha, we should restore the other platform noop implementations in master and 13-x-y 👍  Probably easiest to do in JS land 😄

MarshallOfSound:
Closing this out in favor of the above mentioned solution

miniak:
@MarshallOfSound it still makes sense to merge this to remove dead code in C++

MarshallOfSound (approved):
Fair

jkleinsc (approved):

jkleinsc:
Merging as CI failure is unrelated to PR change.

PR #28920, opened 2021-04-28
docs: uniformize module API doc format

erickzhao:
#### Description of Change
This PR ensures that all API modules are present in the README doc, as there were a couple missing. It also formats all modules to contain a level-1 heading and a blockquote description.

cc @MarshallOfSound just want to make sure I'm doing well by the expected doc format

#### Checklist

#### Release Notes

Notes: none


MarshallOfSound (approved):
Is there potentially a lint script we could write to automatically make sure that the `README` isn't missing any top level docs links?  Not required for this but worth thinking about to ensure this stuff stays in sync

erickzhao:
> Is there potentially a lint script we could write to automatically make sure that the README isn't missing any top level docs links? Not required for this but worth thinking about to ensure this stuff stays in sync

There is for the WIP statically generated website, but it's a bit difficult for the current website because there are a lot of non-module docs in the `/api/` folder that we don't actually show, and I'm not sure if this was intentional or not.

These are either miscellaneous docs about APIs or classes that are contained within other modules (e.g. `/api/experimental`, `api/client-request`, `api/command-line`, `api/dock`, etc.).

MarshallOfSound:
> These are either miscellaneous docs about APIs or classes that are contained within other modules (e.g. /api/experimental, api/client-request, api/command-line, api/dock, etc.).

https://github.com/electron/electron/pull/28712 might help with that, it makes it a bit more obvious which APIs are exported vs which are just "available somewhere in the API"

Issue #28428, opened 2021-03-29
[Bug]: Mas build application crashes on launch when codesigned with "com.apple.security.app-sandbox" enabled 


str4nge-4ttr4ctor:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/electron/electron/issues) for a feature request that matches the one I want to file, without success.

### Electron Version

10.3.2
10.4.1
11.3.0
12.0.1
9.4.4

### What operating system are you using?

macOS

### Operating System Version

macOS Catalina 10.15.7

### What arch are you using?

x64

### Related bug
[27319](https://github.com/electron/electron/issues/27319)

### Expected Behavior

Application to launch without crashing when code-signed with entitlement "com.apple.security.app-sandbox" enabled 

### Actual Behavior

Application crashes on launch when code-signed with entitlement "com.apple.security.app-sandbox" enabled. 

Tried the following electron releases:
https://github.com/electron/electron/releases/download/v10.3.2/electron-v10.3.2-mas-x64.zip
https://github.com/electron/electron/releases/download/v10.4.1/electron-v10.4.1-mas-x64.zip
https://github.com/electron/electron/releases/download/v11.3.0/electron-v11.3.0-mas-x64.zip
https://github.com/electron/electron/releases/download/v12.0.1/electron-v12.0.1-mas-x64.zip
https://github.com/electron/electron/releases/download/v9.4.4/electron-v9.4.4-mas-x64.zip

### To Reproduce

When sandbox permissions are turned off, application launches without any issues. The crash only occurs when sandbox permissions are turned on.

**Crash report when launching the app:**

```
Crashed Thread:        26  ThreadPoolSingleThreadForegroundBlocking0

Exception Type:        EXC_BREAKPOINT (SIGTRAP)
Exception Codes:       0x0000000000000002, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Trace/BPT trap: 5
Termination Reason:    Namespace SIGNAL, Code 0x5
Terminating Process:   exc handler [77030]
```

**Error when running code-signed electron binary:**

[...]

PR #28922, opened 2021-04-29
docs: revise Mac App Store Submission Guide

zcbenz:
#### Description of Change

Close https://github.com/electron/electron/issues/28428.

The Mac App Store Submission Guide has been quite outdated that, it is impossible to make a signed app that can actually run by following it.

I have revised the guide to add latest information about provisioning profile and different kinds of certificates, and the instructions are now using `electron-osx-sign` instead of a custom script that does not run.

#### Release Notes

Notes: none

### DIFF
--- a/docs/tutorial/mac-app-store-submission-guide.md
+++ b/docs/tutorial/mac-app-store-submission-guide.md
@@ -40,7 +40,8 @@ The simplest way to get signing certificates is to use Xcode:
 
 The "Apple Development" certificate is used to sign apps for development and
 testing, on machines that have been registered on Apple Developer website. The
-method of registration will be described later.
+method of registration will be described in
+[Install provisioning profile](#install-provisioning-profile).
 
 Apps signed with the "Apple Development" certificate cannot be submitted to Mac
 App Store. For that purpose, apps must be signed with the "Apple Distribution"
--- a/docs/tutorial/mac-app-store-submission-guide.md
+++ b/docs/tutorial/mac-app-store-submission-guide.md
@@ -3,12 +3,12 @@
 This guide provides information on:
 
 * How to sign Electron apps on macOS;
-* How to submit Electron apps to Mac App Store;
+* How to submit Electron apps to Mac App Store (MAS);
 * The limitations of the MAS build.
 
 ## Requirements
 
-To sign Electron apps, following tools must be installed first:
+To sign Electron apps, the following tools must be installed first:
 
 * Xcode 11 or above.
 * The [electron-osx-sign][electron-osx-sign] npm module.
@@ -19,14 +19,14 @@ fee.
 
 ## Sign Electron apps
 
-Electron apps can be distributed through Mac App Store or outside it, each way
+Electron apps can be distributed through Mac App Store or outside it. Each way
 requires different ways of signing and testing. This guide mainly focuses on
 distribution via Mac App Store, but will also mention other methods.
 
-Following steps describe how to get the certificates from Apple, how to sign
+The following steps describe how to get the certificates from Apple, how to sign
 Electron apps, and how to test them.
 
-### Get Certificates
+### Get certificates
 
 The simplest way to get signing certificates is to use Xcode:
 
@@ -42,27 +42,27 @@ The "Apple Development" certificate is used to sign apps for development and
 testing, on machines that have been registered on Apple Developer website. The
 method of registration will be described later.
 
-Apps signed with the "Apple Development" certificate can not be submitted to Mac
-App Store, for that purpose apps must be signed with the "Apple Distribution"
+Apps signed with the "Apple Development" certificate cannot be submitted to Mac
+App Store. For that purpose, apps must be signed with the "Apple Distribution"
 certificate instead. But note that apps signed with the "Apple Distribution"
-certicate can not run directly, it must be also be signed by Apple to be able to
-run, which will only be possible after downloading from Mac App Store.
+certificate cannot run directly. They must be also be signed by Apple to be able to
+run, which will only be possible after downloading the app from the Mac App Store.
 
 #### Other Certificates
 
 You may notice that there are also other kinds of certificates.
 
 The "Developer ID Application" certificate is used to sign apps before
-distributing it outside the Mac App Store.
+distributing them outside the Mac App Store.
 
 The "Developer ID Installer" and "Mac Installer Distribution" certificates are
 used to sign the Mac Installer Package instead of the app itself. Most Electron
 apps do not use Mac Installer Package so they are generally not needed.
 
 Apps signed with "Apple Development" and "Apple Distribution" certificates can
-only run under sandbox mode, so they must use the MAS build of Electron. But
-the "Developer ID Application" certificate does not have this restrictions, apps
-signed with it can use either the normal build or the MAS build of Electron.
+only run under sandbox mode, so they must use the MAS build of Electron.
+However, the "Developer ID Application" certificate does not have this restrictions, so
+apps signed with it can use either the normal build or the MAS build of Electron.
 
 #### Legacy Certificate Names
 
@@ -77,29 +77,29 @@ using one of the old names.
 
 ### Install Provisioning Profile
 
-If you are going to submit the app to Mac App Store, to test it on your local
-machine before submitting it, you have to install the app's provisioning profile
-on your machine, and sign the app with the "Apple Development" certificate and
-the provisioning profile.
+If you want to test your app on your local machine before submitting your app to
+the Mac App Store, you have to install the app's provisioning profile on your machine,
+and sign the app with the "Apple Development" certificate and the provisioning profile.
 
-To register the provisioning profile, you can follow below steps:
+To register the provisioning profile, you can follow the below steps:
 
 1. Open the "Certificates, Identifiers & Profiles" page on the
    [Apple Developer](https://developer.apple.com/account) website.
-2. Add a new App ID for you app in the "Identifiers" page.
-3. Register your local machine in the "Devices" page, you can find your
+2. Add a new App ID for your app in the "Identifiers" page.
+3. Register your local machine in the "Devices" page. You can find your
    machine's "Device ID" in the "Hardware" page of the "System Information" app.
 4. Register a new Provisioning Profile in the "Profiles" page, and download it
    to `/path/to/yourapp.provisionprofile` and install it.
 
-### Enable Sandbox in Apps
+### Enable Apple's App Sandbox
 
-Apps submitted to Mac App Store must run under [App Sandbox][app-sandboxing],
-and only the MAS build of Electron can run under sandbox mode.
+Apps submitted to the Mac App Store must run under Apple's
+[App Sandbox][app-sandboxing]. Only the MAS build of Electron
+can run in the App Sandbox.
 
-When signing app with `electron-osx-sign`, it will automatically add App Sandbox
-capacity to app's entitlements, but if you are using custom entitlements
-following capacities must be added:
+When signing your app with `electron-osx-sign`, it will automatically add the App
+Sandbox capacity to your app's entitlements, but if you are using custom entitlements,
+the following capacities must be added:
 
 ```xml
 <?xml version="1.0" encoding="UTF-8"?>
@@ -121,8 +121,8 @@ and the `your.bundle.id` should be replaced with the App ID of the app.
 
 #### Extra Steps Without `electron-osx-sign`
 
-If you are signing the app without using `electron-osx-sign`, you should also
-make sure following entitlements are added to the binaries and helpers in the
+If you are signing your app without using `electron-osx-sign`, you should also
+make sure the following entitlements are added to the binaries and helpers in your
 app's bundle:
 
 ```xml
@@ -152,8 +152,8 @@ Apple Developer account's Team ID as its value:
 ```
 
 When using `electron-osx-sign` the `ElectronTeamID` key will be added
-automatically by extracting the Team ID from the certificate's name, you should
-manually add this key if `electron-osx-sign` could not find out the correct Team
+automatically by extracting the Team ID from the certificate's name. You may
+need to manually add this key if `electron-osx-sign` could not find the correct Team
 ID.
 
 ### Sign Apps for Development
@@ -173,7 +173,7 @@ The signed app can only run on the machines that registered by the provisioning
 profile, and this is the only way to test the signed app before submitting to
 Mac App Store.
 
-### Sign Apps for Submitting to Mac App Store
+### Sign apps for submitting to the Mac App Store
 
 To sign an app that will be submitted to Mac App Store, you must sign it with
 the "Apple Distribution" certificate. Note that apps signed with this
@@ -183,7 +183,7 @@ certificate will not run anywhere, unless it is downloaded from Mac App Store.
 electron-osx-sign YourApp.app --identity='Apple Distribution'
 ```
 
-### Sign Apps for Distribution Outside Mac App Store
+### Sign apps for distribution outside the Mac App Store
 
 If you don't plan to submit the app to Mac App Store, you can sign it the
 "Developer ID Application" certificate. In this way there is no requirement to
@@ -203,7 +203,7 @@ This tutorial does not cover [App Notarization][app-notarization], but you might
 want to do it otherwise Apple may prevent users from using your app outside Mac
 App Store.
 
-## Submit Apps to Mac App Store
+## Submit Apps to the Mac App Store
 
 After signing the app with the "Apple Distribution" certificate, you can
 continue to submit it to Mac App Store.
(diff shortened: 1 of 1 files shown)


zcbenz:
Also @sethlu for review.

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -261,15 +323,12 @@ Electron uses following cryptographic algorithms:
 * RIPEMD - [ISO/IEC 10118-3](https://webstore.ansi.org/RecordDetail.aspx?sku=ISO%2FIEC%2010118-3:2004)
 
 [developer-program]: https://developer.apple.com/support/compare-memberships/
+[electron-osx-sign]: https://github.com/electron-userland/electron-osx-sign
```
malept, line 326:
```suggestion
[electron-osx-sign]: https://github.com/electron/electron-osx-sign
```

malept (commented):
I don't have much context with MAS submissions, but can Electron Forge / Electron Packager be used to make the submissions without needing to invoke `electron-osx-sign`, via their respective `osxSign` parameters? If so, I'd prefer that we prioritize that over the `electron-osx-sign` instructions and get rid of the non-`electron-osx-sign` instructions entirely.

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -1,68 +1,105 @@
...
 
-`parent.plist`:
+### Enable Sandbox in Apps
+
+Apps submitted to Mac App Store must run under [App Sandbox][app-sandboxing],
+and only the MAS build of Electron can run under sandbox mode.
+
+When signing app with `electron-osx-sign`, it will automatically add App Sandbox
+capacity to app's entitlements, but if you are using custom entitlements
+following capacities must be added:
```
erickzhao, line 102: Added a few more words here to make the "App Sandbox" branding explicit everywhere, to avoid any possibility of confusion with other uses of `sandbox` in Electron.

```suggestion
### Enable Apple's App Sandbox

Apps submitted to the Mac App Store must run under Apple's
[App Sandbox][app-sandboxing]. Only the MAS build of Electron
can run in the App Sandbox.

When signing your app with `electron-osx-sign`, it will automatically add the App
Sandbox capacity to your app's entitlements, but if you are using custom entitlements,
the following capacities must be added:
```

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -79,7 +116,14 @@ Then, you need to prepare three entitlements files.
...
 
-`loginhelper.plist`:
+The `TEAM_ID` should be replaced with your Apple Developer account's Team ID,
+and the `your.bundle.id` should be replaced with the App ID of the app.
+
+#### Extra Steps Without `electron-osx-sign`
+
+If you are signing the app without using `electron-osx-sign`, you should also
+make sure following entitlements are added to the binaries and helpers in the
+app's bundle:
```
erickzhao, line 126:
```suggestion
If you are signing your app without using `electron-osx-sign`, you should also
make sure the following entitlements are added to the binaries and helpers in your
app's bundle:
```

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -88,80 +132,98 @@ Then, you need to prepare three entitlements files.
...
+</plist>
 ```
 
-If you are new to app sandboxing under macOS, you should also read through
-Apple's [Enabling App Sandbox][enable-app-sandbox] to have a basic idea, then
-add keys for the permissions needed by your app to the entitlements files.
+When using `electron-osx-sign` the `ElectronTeamID` key will be added
+automatically by extracting the Team ID from the certificate's name, you should
+manually add this key if `electron-osx-sign` could not find out the correct Team
+ID.
```
erickzhao, line 157:
```suggestion
When using `electron-osx-sign` the `ElectronTeamID` key will be added
automatically by extracting the Team ID from the certificate's name. You may
need to manually add this key if `electron-osx-sign` could not find the correct Team
ID.
```

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -1,68 +1,105 @@
...
+3. Select a team and click "Manage Certificates";
+4. In the lower-left corner of the signing certificates sheet, click the Add
+   button (+), and add following certificates:
+   * "Apple Development"
+   * "Apple Distribution"
 
-Then, you need to prepare three entitlements files.
+The "Apple Development" certificate is used to sign apps for development and
+testing, on machines that have been registered on Apple Developer website. The
+method of registration will be described later.
```
erickzhao, line 43: Maybe link the exact heading for the below section

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -184,8 +246,8 @@ more information.
 ### Additional Entitlements
 
 Depending on which Electron APIs your app uses, you may need to add additional
-entitlements to your `parent.plist` file to be able to use these APIs from your
-app's Mac App Store build.
+entitlements to your app's entitlements file otherwise App Sandbox would prevent
+using them.
```
erickzhao, line 250:
```suggestion
entitlements to your app's entitlements file. Otherwise, the App Sandbox may prevent
you from using them.
```

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -88,80 +132,98 @@ Then, you need to prepare three entitlements files.
...
+with `--no-gatekeeper-assess`, the `electron-osx-sign` will skip the macOS
+GateKeeper check as your app usually has not been notarized yet by this step.
+
+<!-- TODO(zcbenz): Add a chapter about App Notarization -->
+This tutorial does not cover [App Notarization][app-notarization], but you might
+want to do it otherwise Apple may prevent users from using your app outside Mac
+App Store.
 
-### Upload Your App
+## Submit Apps to Mac App Store
```
erickzhao, line 206:
```suggestion
## Submit Apps to the Mac App Store
```

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -88,80 +132,98 @@ Then, you need to prepare three entitlements files.
...
+If you are signing without `electron-osx-sign`, you must place the provisioning
+profile to `YourApp.app/Contents/embedded.provisionprofile`.
 
-```sh
-electron-osx-sign YourApp.app YourApp.app/Contents/Resources/app/node_modules/nativemodule/build/release/nativemodule
+The signed app can only run on the machines that registered by the provisioning
+profile, and this is the only way to test the signed app before submitting to
+Mac App Store.
+
+### Sign Apps for Submitting to Mac App Store
```
erickzhao, line 176:
```suggestion
### Sign apps for submitting to the Mac App Store
```

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -88,80 +132,98 @@ Then, you need to prepare three entitlements files.
...
+
+To sign an app that will be submitted to Mac App Store, you must sign it with
+the "Apple Distribution" certificate. Note that apps signed with this
+certificate will not run anywhere, unless it is downloaded from Mac App Store.
+
+```bash
+electron-osx-sign YourApp.app --identity='Apple Distribution'
+```
+
+### Sign Apps for Distribution Outside Mac App Store
```
erickzhao, line 186:
```suggestion
### Sign apps for distribution outside the Mac App Store
```

erickzhao (commented):
Thanks so much for doing this!

zcbenz:
> can Electron Forge / Electron Packager be used to make the submissions without needing to invoke `electron-osx-sign`, via their respective `osxSign` parameters? If so, I'd prefer that we prioritize that over the `electron-osx-sign` instructions and get rid of the non-`electron-osx-sign` instructions entirely.

Since `electron-packager` directly calls `electron-osx-sign`, it is possible to replace every `electron-osx-sign` call with `electron-packager`. But this guide focuses on signing and mentions nothing about packaging so I don't think it will make the instructions simpler.

I have already removed most manual instructions, and the only command readers will run is one line of `electron-osx-sign`, which they can easily translate to options of different packagers. There are still some mentions of how to sign without `electron-osx-sign`, but they are not part of the instructions, and the information is necessary to understand how signing works under the hood.

malept:
> > can Electron Forge / Electron Packager be used to make the submissions without needing to invoke `electron-osx-sign`, via their respective `osxSign` parameters? If so, I'd prefer that we prioritize that over the `electron-osx-sign` instructions and get rid of the non-`electron-osx-sign` instructions entirely.
> 
> Since `electron-packager` directly calls `electron-osx-sign`, it is possible to replace every `electron-osx-sign` call with `electron-packager`. But this guide focuses on signing and mentions nothing about packaging so I don't think it will make the instructions simpler.

Since we've already started referencing Forge config in several guides, I'd like to be consistent. I'm pretty sure if we merge this as-is, folks who read these docs will assume that they will need to install `electron-osx-sign` *in addition* to Forge or Packager when that is not the case.

> There are still some mentions of how to sign without `electron-osx-sign`, but they are not part of the instructions, and the information is necessary to understand how signing works under the hood.

I wonder how much we actually need to mention how it works under the hood. In my experience of helping out in places such as the Discord server, folks don't differentiate and assume that the entire guide is all a part of the same set of instructions, rather than "here's how you do it *with* `electron-osx-sign`, and here's how you do it without".

zcbenz:
> Since we've already started referencing Forge config in several guides, I'd like to be consistent. I'm pretty sure if we merge this as-is, folks who read these docs will assume that they will need to install `electron-osx-sign` _in addition_ to Forge or Packager when that is not the case.

Some of the most popular apps do not use Electron Forge / Electron Packager (as far as I know at least Atom, VS Code, Teams), so the documentation should probably not assume using them.

And currently an annoying thing with signing for Mac App Store is, you have to sign with one certificate for  development, and another certificate for submitting, and if you are distributing outside Mac App Store, you have to use yet another certificate. I can not merge all use cases in one config, and there is no best workflow, every company seems to have their own best practice.

If there is a golden config that works for everyone I would be happy to include it here, but there does not seem to have one. The packagers all seem to assume that an app only needs to be signed with one certificate, and I think they need to figure out a better workflow to handle the case of Mac App Store.

> I wonder how much we actually need to mention how it works under the hood. In my experience of helping out in places such as the Discord server, folks don't differentiate and assume that the entire guide is all a part of the same set of instructions, rather than "here's how you do it _with_ `electron-osx-sign`, and here's how you do it without".

I agree we don't really need to mention it in this guide, and I'm fine putting the internal things in another document.

But this PR means to be improvement of the existing guide, the old guide uses a custom script and this PR changes to use `electron-osx-sign` instead, the old guide has lots of instructions on manual code signing and this PR reduces those to minimal and puts them in a small chapter. So this guide is only getting better instead of getting worse.

We can surely go further by using packagers as the main method, and move the internal things away, but I think they should be done as progressive efforts in separate PRs, this PR means to be a `fix` for docs instead of a `refactor` or ` feat`.

Toinane (approved):

codebytere (approved):

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -1,53 +1,107 @@
...
 
-### Get Certificate
+* Xcode 11 or above.
+* The [electron-osx-sign][electron-osx-sign] npm module.
 
-To submit your app to the Mac App Store, you first must get a certificate from
-Apple. You can follow these [existing guides][nwjs-guide] on web.
+You also have to register an Apple Developer account and join the
+[Apple Developer Program][developer-program], which has a yearly subscription
+fee.
```
nornagon, line 18: Minor nit, I'm not sure we need to call out the details of the Apple Developer Program here.

```suggestion
[Apple Developer Program][developer-program].
```

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -1,53 +1,107 @@
...
 
-### Get Team ID
+## Sign Electron apps
 
-Before signing your app, you need to know the Team ID of your account. To locate
-your Team ID, Sign in to [Apple Developer Center](https://developer.apple.com/account/),
-and click Membership in the sidebar. Your Team ID appears in the Membership
-Information section under the team name.
+Electron apps can be distributed through Mac App Store or outside it. Each way
+requires different ways of signing and testing. This guide mainly focuses on
```
nornagon, line 23:
```suggestion
requires different ways of signing and testing. This guide focuses on
```

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -1,53 +1,107 @@
...
+distribution via Mac App Store, but will also mention other methods.
```
nornagon, line 24: 🤔  if we're also describing other distribution methods in this guide, is "Mac App Store Submission Guide" the right name for this document? Should that other information be separated into a "Mac Code Signing Guide" or something?

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -1,53 +1,107 @@
...
-</dict>
-</plist>
-```
+1. Open Xcode and open "Accounts" preferences;
+2. Sign in with your Apple account;
+3. Select a team and click "Manage Certificates";
+4. In the lower-left corner of the signing certificates sheet, click the Add
+   button (+), and add following certificates:
+   * "Apple Development"
+   * "Apple Distribution"
```
nornagon, line 39: Is there any Apple documentation we could link to here?

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -1,53 +1,107 @@
...
+
+To get the provisioning profile, you can follow the below steps:
+
+1. Open the "Certificates, Identifiers & Profiles" page on the
+   [Apple Developer](https://developer.apple.com/account) website.
+2. Add a new App ID for your app in the "Identifiers" page.
+3. Register your local machine in the "Devices" page. You can find your
+   machine's "Device ID" in the "Hardware" page of the "System Information" app.
+4. Register a new Provisioning Profile in the "Profiles" page, and download it
+   to `/path/to/yourapp.provisionprofile`.
```
nornagon, line 94: Can we link to some Apple documentation here?

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -1,53 +1,107 @@
...
+
+### Enable Apple's App Sandbox
+
+Apps submitted to the Mac App Store must run under Apple's
+[App Sandbox][app-sandboxing], and only the MAS build of Electron can run with
+the App Sandbox.
```
nornagon, line 100: I think it's worth reiterating this here to make the point a bit stronger
```suggestion
the App Sandbox. The standard darwin build of Electron will fail to launch
when run under App Sandbox.
```

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -1,53 +1,107 @@
...
+
+When signing your app with `electron-osx-sign`, it will automatically add the
+necessary capacities to your app's entitlements, but if you are using custom
```
nornagon, line 103: I think each individual part of an app's entitlements is called an "entitlement" (not "capacity")

```suggestion
necessary entitlements to your app's entitlements, but if you are using custom
```

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -56,13 +110,14 @@ Then, you need to prepare three entitlements files.
...
-    <true/>
   </dict>
 </plist>
 ```
 
-`parent.plist`:
+#### Extra steps without `electron-osx-sign`
+
+If you are signing your app without using `electron-osx-sign`, you must ensure
+the app bundle's entitlements have at least following capacities:
```
nornagon, line 120:
```suggestion
the app bundle's entitlements have at least following keys:
```

Could write "the app bundle's entitlements have at least following entitlements" here but that sounds a bit awkward. I think "keys" is a bit more correct than "capacities" though.

docs/tutorial/mac-app-store-submission-guide.md
```diff
@@ -88,80 +147,97 @@ Then, you need to prepare three entitlements files.
...
-Also note that native modules may have intermediate files produced which should
-not be included (as they would also need to be signed). If you use
-[electron-packager][electron-packager] before version 8.1.0, add
-`--ignore=.+\.o$` to your build step to ignore these files. Versions 8.1.0 and
-later ignore those files by default.
+### Sign apps for distribution outside the Mac App Store
+
+If you don't plan to submit the app to Mac App Store, you can sign it the
+"Developer ID Application" certificate. In this way there is no requirement to
+use App Sandbox and the app can be built with either MAS or the normal build of
```
nornagon, line 205: I don't think we should suggest that apps that are run outside of App Sandbox be built with the MAS version, as it hamstrings sandboxing. Perhaps note here that if you're not using App Sandbox, you should not use the MAS version (or inversely, only use the MAS build if you're enabling App Sandbox?)

docs/tutorial/mac-app-store-submission-guide.md
zcbenz: Yeah this chapter should eventually be moved into a separate guide, but I would like to do it after adding more information about App Notarization, currently this chapter has too little information to form a new guide.

Issue #28620, opened 2021-04-12
[Feature Request]: Preload scripts for Web Worker


hamst:
### Preflight Checklist

I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
I have searched the [issue tracker](https://www.github.com/electron/electron/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Using the option `nodeIntegrationInWorker` to enable node integration in worker threads is not safe cause it makes a very powerful API accessible for any workers with no control over it.

### Proposed Solution

Adding a flag 'preloadInWorker' in webPreferences of BrowserWindow to specify a script to be loaded before any other scripts in Web Workers. The script will always have access to node APIs, and it's aimed to introduce the global symbols that will be visible in the worker scope and considered safe enough to use. That should work the same way as 'preload' option but be applicable for the web worker threads.

### Alternatives Considered

n/a

### Additional Information

n/a

Prinzhorn:
See also discussion in https://github.com/electron/electron/issues/22582

Jelmerro:
@codebytere Is there anything we can do to help solve this issue? I am using preloads to override and hide some sensitive APIs from browser contexts, only to find out they aren't actually loaded in workers. This means there is no way to fix issues like [Electron's broken atob/btoa implementation](https://github.com/electron/electron/issues/31018), or hiding APIs such as battery/keyboard/gpuinfo from being accessed. It would help with fixing these sensitive APIs in workers which would increase security for web contexts. This is still a relevant request for the latest electron 20 and 21, where there is currently no way to run preloads in workers.

threema-danilo:
This issue is a major security hardening roadblock when relying on workers that need to access certain (but not all) Node APIs.

What would be needed to bring preload script support to workers?

From what I can tell, something like the following code snippet in https://github.com/electron/electron/blob/db2bf1a0d1be6b81ca2d1a6b693b78753d3ba6ea/shell/renderer/web_worker_observer.cc#L48C25-L48C55 would load the current preload scripts for the worker, is that right?

```c++
auto* isolate = context->GetIsolate();
auto binding = v8::Object::New(isolate);
InitializeBindings(binding, context, render_frame);

std::vector<v8::Local<v8::String>> sandbox_preload_bundle_params = {
[...]

PR #28923, opened 2021-04-29
feat: add preloadInWorker API

hamst:
#### Description of Change
Add `preloadInWorker` for `webPreferences` in `BrowserWindow` to define a script to be loaded first before any other scripts in worker.

cc @codebytere 

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Added preloadInWorker to define a script to be loaded first before any other scripts in worker.

### DIFF
--- a/lib/worker/init.ts
+++ b/lib/worker/init.ts
@@ -12,24 +12,61 @@ require('../common/reset-search-paths');
 // Import common settings.
 require('@electron/internal/common/init');
 
-// Export node bindings to global.
-const { makeRequireFunction } = __non_webpack_require__('internal/modules/cjs/helpers') // eslint-disable-line
-global.module = new Module('electron/js2c/worker_init');
-global.require = makeRequireFunction(global.module);
+// Process command line arguments.
+const { hasSwitch, getSwitchValue } = process._linkedBinding('electron_common_command_line');
 
-// Set the __filename to the path of html file if it is file: protocol.
-if (self.location.protocol === 'file:') {
-  const pathname = process.platform === 'win32' && self.location.pathname[0] === '/' ? self.location.pathname.substr(1) : self.location.pathname;
-  global.__filename = path.normalize(decodeURIComponent(pathname));
-  global.__dirname = path.dirname(global.__filename);
+const parseOption = function (name: string) {
+  return hasSwitch(name) ? getSwitchValue(name) : null;
+};
 
-  // Set module's filename so relative require can work as expected.
-  global.module.filename = global.__filename;
+const nodeIntegration = hasSwitch('node-integration-in-worker');
+const preloadScript = parseOption('preload-in-worker');
 
-  // Also search for module under the html file.
-  global.module.paths = Module._nodeModulePaths(global.__dirname);
+if (nodeIntegration) {
+  // Export node bindings to global.
+  const { makeRequireFunction } = __non_webpack_require__('internal/modules/cjs/helpers') // eslint-disable-line
+  global.module = new Module('electron/js2c/worker_init');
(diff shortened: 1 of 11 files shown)


lib/worker/init.ts
```diff
@@ -12,24 +12,61 @@ require('../common/reset-search-paths');
...
+
+const parseOption = function (name: string) {
+  return hasSwitch(name) ? getSwitchValue(name) : null;
+};
+
+const nodeIntegration = hasSwitch('node-integration-in-worker');
+const preloadScript = parseOption('preload-in-worker');
+
+if (nodeIntegration) {
+  // Export node bindings to global.
```
zcbenz, line 26: This file shares lots of code with `lib/renderer/init.ts`, can you move the shared code into a utility function? Like what `windowSetup` and `webViewInit` do.

zcbenz (changes requested):
From the document it is unclear how this option works with `contextIsolation`, can you make it clear and add related tests?

And I think this option should either respect the `contextIsolation` option of the window, or just always run the preload script in isolated context. /cc @electron/wg-api

hamst:
@zcbenz It seems that contextIsolation is not supported in WebWorkers, so adding preloadInWorker doesn't affect it in any ways. I will add some details in documents, as far as having tests with contextIsolation, will they be useful?

zcbenz:
The `contextIsolation` controls whether preload scripts should run in the web page's context or in a separate context, currently it is not related to WebWorker because WebWorker does not have preload scripts.

But since this PR is adding `preloadInWorker`, we have to define how `contextIsolation` affects preload scripts in WebWorker. We can either:
1. Make `preloadInWorker` respect the `contextIsolation` option that, `contextIsolation` should be able to control whether preload scripts should run in the worker script's context or in a separate context ;
2. Or ignore the  `contextIsolation` option and always run the preload scripts in separate context in WebWorker.

From current implementation, the preload scripts specified by `preloadInWorker` are running in the same context of the worker script, which is against the "secure by default" principle that we have in Electron's APIs.

MarshallOfSound (changes requested):
Without solving the ctx isolation issue:

> This script will always have access to node APIs no matter whether node integration in worker is turned on or off.

is fundamentally a massive security vulnerability that we can't / won't land into this project.  Putting a hard block on this pending a proposal for how to deal with ctx isolation

hamst:
Right now we have an option `nodeIntegrationInWorker` but don't support `contextIsolation` in worker. Being enabled `nodeIntegrationInWorker` allows access to powerful API without any limitation for all web worker scripts loaded in webpage, and there is no way to limit this. Very unlikely that this option without `contextIsolation` support follows "secure by default" principle, and that's what we have today. So `contextIsolation` for worker must be supported no matter if we have `preloadInWorker` or not.
Option `preloadInWorker` makes things way better: you define the script with all access to API, and you can make some API visible for all workers, but it requires additional steps and unlikely could be done unintentionally. By default, nothing will be exposed. That's a big difference from security prospective.
So getting `contextIsolation` support for workers should be a separate task which must be done to be "secure by defaults" regardless `preloadInWorker`. As `contextIsolation` heavily relies on contextBridge module, some work needs to be done to relax the requirement to avoid using electron modules in workers https://www.electronjs.org/docs/tutorial/multithreading#available-apis and make the implementation thread-safe.

zcbenz:
@hamst Thanks for explaining, I see your point now. I'm good with this API, since it is an improvement over `nodeIntegrationInWorker` on security, before `contextIsolation` is supported in worker.

hamst:
Would be good to link the issue #28620 to this PR.

PR #28924, opened 2021-04-29
docs: update dark mode tutorial

SergioBerrio:
i have added an attribute which helped me to make the aplication run

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none.

### DIFF
--- a/docs/tutorial/dark-mode.md
+++ b/docs/tutorial/dark-mode.md
@@ -138,7 +138,8 @@ function createWindow () {
     width: 800,
     height: 600,
     webPreferences: {
-      nodeIntegration: true
+      nodeIntegration: true,
+      contextIsolation: false
     }
   })
 


MarshallOfSound (changes requested):
As with other examples we should update the example to support `contextIsolation: true` rather than disabling an important security feature

codebytere:
@SergioBerrio please address the above comments if you would like for this PR to move forward!

docs/tutorial/dark-mode.md
```diff
@@ -138,7 +138,8 @@ function createWindow () {
     width: 800,
     height: 600,
     webPreferences: {
-      nodeIntegration: true
+      nodeIntegration: true,
+      contextIsolation: false
     }
```
matthijsgroen, line 143: This was a thing that bothered me in the docs a bit.. why are the examples not using a 'preload' script and promoting the proper behavior? (That would be using `{ contextIsolation: true, nodeIntegration: false }` right?)

I wanted to fix all security related things and the consensus on stack overflow and guides would be: 'security', why bother. While the documentation on contextIsolation was actually really good and a way nicer way from the webapp to communicate with electron, since with the preload script the whole communication with electron is more like pure feature detection.

zcbenz:
I'm closing this PR since the example should be updated to use preload scripts instead, please let me know if you still plan to work on this PR and I'll reopen it.

PR #28930, opened 2021-04-29
chore: cherry-pick 668cf831e912 from chromium

ppontes:
Never fail in ReceiverSet::Add

Because of how UniqueReceiverSet is implemented and used, it is
dangerous to allow Add() to fail: callers reasonably assume that added
objects are still alive immediately after the Add() call.

This changes ReceiverId to a uint64 and simply CHECK-fails on
insert collision.

This fundamentally increases binary size of 32-bit builds, because
a widely used 32-bit data type is expanding to 64 bits for the sake
of security and stability. It is effectively unavoidable for now, and
also just barely above the tolerable threshold.

A follow-up (but less backwards-mergeable) change should be able to
reduce binary size beyond this increase by consolidating shared
code among ReceiverSet template instantiations.

Fixed: 1185732
Change-Id: I9acf6aaaa36e10fdce5aa49a890173caddc13c52
Binary-Size: Unavoidable (see above)
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/2778871
Commit-Queue: Ken Rockot <rockot@google.com>
Auto-Submit: Ken Rockot <rockot@google.com>
Reviewed-by: Robert Sesek <rsesek@chromium.org>
Cr-Commit-Position: refs/heads/master@{#865815}

Notes: Security: backported fix to CVE-2021-21207.

### DIFF
--- a/patches/chromium/.patches
+++ b/patches/chromium/.patches
@@ -156,3 +156,4 @@ cherry-pick-6b84dc72351b.patch
 cherry-pick-1028ffc9bd83.patch
 privacy_budget_remove_unnecessary_kcanvasreadback_metrics.patch
 cherry-pick-406ae3e8a9a8.patch
+cherry-pick-668cf831e912.patch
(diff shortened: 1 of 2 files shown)


alaraap (approved):

alaraap:
https://docs.github.com/en/graphql/reference/mutations#cloneproject

MarshallOfSound (approved):

PR #28938, opened 2021-04-29
refactor: replace --offscreen flag with ApplyWebPreferences

nornagon:
#### Description of Change
This moves --offscreen out of AppendCommandLineSwitches and applies it as part
of WebPreferences. It's still a process-wide flag, but this makes it
theoretically switchable at runtime, e.g. to in future support spare renderer
processes.

It's part of cleaning up WebContentsPreferences::AppendCommandLineSwitches to
only contain things that strictly must be command-line switches.

#### Checklist

- [ ] `npm test` passes

#### Release Notes

Notes: none


MarshallOfSound (commented):
> It's still a process-wide flag, but this makes it theoretically switchable at runtime

If it's a process wide flag should we have more complex logic for determining what to do if two WebContents exist in the same process with different settings.  i.e. I assume one setting should win over the other. E.g. `true + false === false` or something

nornagon:
@MarshallOfSound this matches existing logic in this method that sets runtime-enabled blink features process-wide

deepak1556 (approved):

MarshallOfSound (approved):

codebytere (approved):

PR #28945, opened 2021-04-30
docs: remove reference to global Electron install

erickzhao:
#### Description of Change

This is a pattern that we actively want to discourage.

#### Checklist

#### Release Notes

Notes: none

### DIFF
--- a/docs/tutorial/installation.md
+++ b/docs/tutorial/installation.md
@@ -11,14 +11,6 @@ npm install electron --save-dev
 See the [Electron versioning doc][versioning] for info on how to
 manage Electron versions in your apps.
 
-## Global Installation
-
-You can also install the `electron` command globally in your `$PATH`:
-
-```sh
-npm install electron -g
-```
-
 ## Customization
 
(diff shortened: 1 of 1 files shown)


VerteDinde (approved):

docs/tutorial/installation.md
```diff
@@ -11,14 +11,6 @@ npm install electron --save-dev
 See the [Electron versioning doc][versioning] for info on how to
 manage Electron versions in your apps.
 
-## Global Installation
```
Kilian, line 14: Rather than removing it all together we can consider changing it to:

## Running Electron without installing it

You can also run an electron without installing Electron:

```sh
npx electron .
```

This will let you run an electron app, but keep in mind any other dependencies will also not be installed.

codebytere (approved):

docs/tutorial/installation.md
erickzhao: Updated!

Kilian (approved):

PR #15071, opened 2018-10-10
fix: Promise resolution and unit test

nitsakh:
##### Description of Change
Address issue from ch69 upgrade
https://github.com/orgs/electron/projects/24#card-13689743

##### Checklist

- [ ] PR description included and stakeholders cc'd
- [ ] `npm test` passes
- [ ] tests are [changed or added](https://github.com/electron/electron/blob/master/docs/development/testing.md)
- [ ] relevant documentation is changed or added
- [ ] PR title follows semantic [commit guidelines](https://github.com/electron/electron/blob/master/docs/development/pull-requests.md#commit-message-guidelines)

##### Release Notes

Notes: no-notes

PR #28957, opened 2021-05-01
fix: allow Node.js to manage microtasks queue

indutny:
#### Description of Change

When `uv_run()` resulted in invocation of JS functions the microtask
queue checkpoint in Node's CallbackScope was a no-op because the
expected microtask queue policy was `kExplicit` and Electron ran under
`kScoped` policy. This change switches policy to `kExplicit` right
before `uv_run()` and reverts it back to original value after `uv_run()`
completes to provide better compatibility with Node.

cc @codebytere since you touched this code.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Allow Node.js to manage microtasks queue by using explicit microtasks policy before calling `uv_run()`

### DIFF
--- a/shell/common/node_bindings.cc
+++ b/shell/common/node_bindings.cc
@@ -575,8 +575,9 @@ void NodeBindings::UvRunOnce() {
   // Enter node context while dealing with uv events.
   v8::Context::Scope context_scope(env->context());
 
-  // Perform microtask checkpoint after running JavaScript.
-  gin_helper::MicrotasksScope microtasks_scope(env->isolate());
+  auto old_policy = env->isolate()->GetMicrotasksPolicy();
+  DCHECK_EQ(v8::MicrotasksScope::GetCurrentDepth(env->isolate()), 0);
+  env->isolate()->SetMicrotasksPolicy(v8::MicrotasksPolicy::kExplicit);
 
   if (browser_env_ != BrowserEnvironment::kBrowser)
     TRACE_EVENT_BEGIN0("devtools.timeline", "FunctionCall");
@@ -586,6 +587,8 @@ void NodeBindings::UvRunOnce() {
 
   if (browser_env_ != BrowserEnvironment::kBrowser)
     TRACE_EVENT_END0("devtools.timeline", "FunctionCall");
+
+  env->isolate()->SetMicrotasksPolicy(old_policy);
 
   if (r == 0)
     base::RunLoop().QuitWhenIdle();  // Quit from uv.
--- a/shell/common/node_bindings.cc
+++ b/shell/common/node_bindings.cc
@@ -575,6 +575,10 @@ void NodeBindings::UvRunOnce() {
   // Enter node context while dealing with uv events.
   v8::Context::Scope context_scope(env->context());
 
+  // Node.js expects `kExplicit` microtasks policy and will run microtasks
+  // checkpoints after every call into JavaScript. Since we use a different
+  // policy in the renderer - switch to `kExplicit` and then drop back to the
+  // previous policy value.
   auto old_policy = env->isolate()->GetMicrotasksPolicy();
   DCHECK_EQ(v8::MicrotasksScope::GetCurrentDepth(env->isolate()), 0);
   env->isolate()->SetMicrotasksPolicy(v8::MicrotasksPolicy::kExplicit);


indutny:
Here's the "test case" (benchmark really): https://github.com/indutny/electron-napi-churn-demo

Build instructions:
```sh
npm install
```

Test instructions (Node.js):
```sh
node app/preload.js
```

Test instructions (electron):
```sh
./node_modules/.bin/electron app/main.js
```

Numbers:
```sh
# Node.js
190.523221ms
# Electron (unpatched)
369.185357ms
# Electron (patched)
194.212882ms
```

indutny:
The downside of this approach is that malformed code like this:
```js
function hang() {
  setImmediate(hang);
}
hang();
```
causes starvation of the main thread (while UV loop continues to run fine).

codebytere:
@indutny is this something you plan to upstream? Also - could you please add some release notes?

deepak1556:
Can we check if `env->immediate_idle_handle()` is not next in queue of `uv_loop_->idle_handles` before running the loop to avoid the starvation issue ?

indutny:
> @indutny is this something you plan to upstream?

I'm not sure if it makes much sense for Node.js, since there is no extra cost for waiting for a next uv loop iteration there. @addaleax could you take a look a the patch? I'd appreciate to hear your opinion on this issue.

> Also - could you please add some release notes?

Done, sorry for not writing these immediately!

indutny:
> Can we check if `env->immediate_idle_handle()` is not next in queue of `uv_loop_->idle_handles` before running the loop to avoid the starvation issue ? I think it might also help with addressing the failing tests.

This is a great idea! Let me come up with a patch real quick!

indutny:
btw, as you can imagine I've ran into while investigating performance issues in a real app. In that case `napi_threadsafe_function` was called through `neon` since there it is easier to keep it around than to keep the `FunctionContext` (rust wrapper around `napi_env` essentially).

indutny:
@deepak1556 @codebytere all resolved except for submitting it as a PR to Node.js (see above).

patches/node/node-api_poll_immediately_in_threadsafe_function.patch
```diff
@@ -0,0 +1,33 @@
...
+--- a/src/node_api.cc
++++ b/src/node_api.cc
+@@ -316,6 +316,11 @@ class ThreadSafeFunction : public node::AsyncResource {
+         call_js_cb(env, js_callback, context, data);
+       });
+     }
++
++    if (popped_value) {
++      node::Mutex::ScopedLock lock(this->mutex);
++      CHECK_EQ(0, uv_idle_start(&idle, IdleCb));
```
addaleax, line 29: I don’t think you need to lock the mutex here, the idle handle can only meaningfully be accessed from the owning thread (which runs this code) anyway

shell/common/node_bindings.cc
```diff
@@ -575,17 +575,27 @@ void NodeBindings::UvRunOnce() {
...
+    r = uv_run(uv_loop_, UV_RUN_NOWAIT);
 
-  if (browser_env_ != BrowserEnvironment::kBrowser)
-    TRACE_EVENT_END0("devtools.timeline", "FunctionCall");
+    if (browser_env_ != BrowserEnvironment::kBrowser)
+      TRACE_EVENT_END0("devtools.timeline", "FunctionCall");
+
+    // Make sure that `setImmediate` yields to browser to avoid starvation
+    auto* immediate_idle_handle =
+        reinterpret_cast<uv_handle_t*>(env->immediate_idle_handle());
```
addaleax, line 594: Is there a public API that Node.js could reasonable introduce at some point here (without exposing the idle handle itself, which is really more of an implementation detail)?

addaleax (commented):
I think if you want to upstream this, it might be nicer to move the `if (size == 0) { ... }` block in this function to after the `if (popped_value) { ... }` block (and thus avoiding the call to `uv_idle_stop()` if it’s not necessary), instead of restarting the idle handle even if there are not actually any new items in the queue?

patches/node/node-api_poll_immediately_in_threadsafe_function.patch
indutny: Good point! Let me try it out and submit a PR to Node as well.

shell/common/node_bindings.cc
indutny: > I think if you want to upstream this, it might be nicer to move the `if (size == 0) { ... }` block in this function to after the `if (popped_value) { ... }` block (and thus avoiding the call to `uv_idle_stop()` if it’s not necessary), instead of restarting the idle handle even if there are not actually any new items in the queue?

Here's the reason why this change is actually needed. In half of the cases that I've observed in our app the `Push` was called from within the JS callback of the threadsafe function. In another half - from the microtasks queue (because there were tons of promises used in JS). The extra iteration is extra cost, but for Electron the cost of not doing the iteration is substantially higher. I'm going to add `else if (!popped_value) {` before `uv_idle_stop()` to simplify the logic, but I think I'd rather have the extra iteration. With this in mind, do you think I should submit it to Node.js?

indutny:
> I think if you want to upstream this, it might be nicer to move the `if (size == 0) { ... }` block in this function to after the `if (popped_value) { ... }` block (and thus avoiding the call to `uv_idle_stop()` if it’s not necessary), instead of restarting the idle handle even if there are not actually any new items in the queue?

(Argh, wanted to move my comment to top-level and copy-paste failed me)

I've change the patch to make it avoid calling `uv_idle_stop` when we popped the value (i.e. the queue is active), but for this optimization to have an effect - the extra iteration is still required. For our app, the `Push()` was called not only from the threadsafe function, but also from the microtasks queue (promises...). I don't believe that this optimization is necessary for Node.js, because the cost of extra full uv tick is substantially lower than in Electron.

With this in mind, would you say I should submit it upstream (Node.js)?

shell/common/node_bindings.cc
indutny: Sounds like a good idea to me. This PR just wants to know if `setImmediate` is scheduled for run in order to interleave `uv_run()` with browser's event loop and avoid starvation.

addaleax:
@indutny Right, but … my point is more like, after the `if (popped_value) { ... }` the JS callback and its microtask queue have run already, so you would know for sure whether there are new entries in the queue?

> I don't believe that this optimization is necessary for Node.js, because the cost of extra full uv tick is substantially lower than in Electron.

Yeah, I agree. Maybe it would also make sense to do something here that matches what we do for e.g. incoming messages on MessagePorts, i.e. loop over the queue synchronously up to a specific iteration number limit, and only return control back to the event loop and defer the other iterations if that limit has been exceeded? (Because the only reason I could imagine `DispatchOne()` only dispatching a single item is to avoid event loop starvation.)

indutny:
> @indutny Right, but … my point is more like, after the `if (popped_value) { ... }` the JS callback and its microtask queue have run already, so you would know for sure whether there are new entries in the queue?

The trouble is that microtasks queue is emptied after `uv_run()` completes so `DispatchOne()` is leaved and can't check the state anymore. I'm not sure if there are any implications of running the microtasks queue in `DispatchOne()`, but this sounds a bit heavy-handed to me. What do you think?

> 
> > I don't believe that this optimization is necessary for Node.js, because the cost of extra full uv tick is substantially lower than in Electron.
> 
> Yeah, I agree. Maybe it would also make sense to do something here that matches what we do for e.g. incoming messages on MessagePorts, i.e. loop over the queue synchronously up to a specific iteration number limit, and only return control back to the event loop and defer the other iterations if that limit has been exceeded? (Because the only reason I could imagine `DispatchOne()` only dispatching a single item is to avoid event loop starvation.)

I see. As an extra sanity check, I think this could be a great idea. Maybe we don't even need that idle handle check if there is a limit to the execution count anyway (cc @deepak1556 @codebytere )? Correct me if I'm wrong @addaleax, but Node.js uses 1000 as a limit for same tick message processing and maybe Electron should do here the same?

addaleax:
> > @indutny Right, but … my point is more like, after the `if (popped_value) { ... }` the JS callback and its microtask queue have run already, so you would know for sure whether there are new entries in the queue?
> 
> The trouble is that microtasks queue is emptied after `uv_run()` completes so `DispatchOne()` is leaved and can't check the state anymore. I'm not sure if there are any implications of running the microtasks queue in `DispatchOne()`, but this sounds a bit heavy-handed to me. What do you think?

`DispatchOne()` contains a `CallbackScope` around the relevant JS function call, which results in the microtask queue being run, so that’s already happening here (unless Electron suppresses that somehow? In which case I think that’s the real bug here). But yeah, the microtask queue should always already be empty when `uv_run()` returns.

> Correct me if I'm wrong @addaleax, but Node.js uses 1000 as a limit for same tick message processing and maybe Electron should do here the same?

Yes, or the patch to Node.js could also do the same thing here already if that makes sense (in which case upstreaming it also seems like a good idea).

indutny:
I actually had a patch in works that did that and many other things :joy:, but then microtasks queue killed whole idea. Let me give it few more tryouts to see why `CallbackScope` doesn't empty it. A totally uneducated question - but could microtasks queue in `CallbackScope` not be run because it is under this condition?

```C++
  if (!tick_info->has_tick_scheduled()) {
    env_->context()->GetMicrotaskQueue()->PerformCheckpoint(env_->isolate());

    perform_stopping_check();
  }
```

addaleax:
> A totally uneducated question - but could microtasks queue in `CallbackScope` not be run because it is under this condition?

Well… if `has_tick_scheduled()` is true, it should call the next tick queue, which then runs microtasks at its end, so I think this should be fine even in Electron?

If the `CallbackScope` doesn’t empty it, it might be because there is one on an outer scope (that contains `uv_run()`), which would break a bunch of assumptions that `CallbackScope` makes.

indutny:
@addaleax could `has_tick_scheduled` be false for me?

addaleax:
@indutny That really depends on the JS callback here :) But, sure, if the callback doesn’t schedule any `process.nextTick()` calls, then it should be false.

indutny:
@addaleax the callback is basically an empty `async function() {}` and the Rust code invokes `.then()` on its return value to schedule another threadsafe function execution.

indutny:
Here's the PR with the speed improvements for Node.js: https://github.com/nodejs/node/pull/38506

I've just tested the electron build without that extra `uv_idle_t` iteration, and the performance improvement for our app is gone with it. Sounds like at least for our use case it is absolutely necessary.

indutny:
> If the `CallbackScope` doesn’t empty it, it might be because there is one on an outer scope (that contains `uv_run()`), which would break a bunch of assumptions that `CallbackScope` makes.

FWIW, Electron only has `gin_helper::MicrotasksScope`, `v8::HandleScope`, and `v8::Context::Scope` around `uv_run()` call.

addaleax:
Maybe the `MicrotasksScope` is messing with the explicit `PerformCheckpoint()` call here?

indutny:
@addaleax I don't know, sounds like the call happens only on destructor and this in turn happens only after `uv_run()` leaves. So if the threadsafe function was queued from there - it wasn't called from `uv_run()`.

I can try adding some debug printing to the build to see why it is such. Let me see.

indutny:
Oh gosh, I see now that the condition is **not** has_tick_scheduled. My bad! Yeah, it is funky.

indutny:
Alright, now I have a better understanding of what's going on in Electron thanks to invaluable help of @addaleax . Electron uses scoped microtask policy (with some floating v8 patches so that node's explicit policy use doesn't crash it), and so the microtasks queue is not emptied until we left `uv_run()`. This explains why I was seeing `napi_threadsafe_function` calls happening after `uv_run()`!

Seems like there are two ways to go forward with this:

* Patch Node to use the `MicrotasksScope` in the `CallbackScope` instead of manual checkpoint
* Do an extra idle tick in the threadsafe function

My impression is that the preference (correct me if I'm wrong @addaleax ) should go for the first one. In this case, it looks like no changes to the electron's code base would be needed other than pulling up the patches from Node.js upstream!

addaleax:
> * Patch Node to use the `MicrotasksScope` in the `CallbackScope` instead of manual checkpoint

Yeah, I’m sorry, I thought that that would help, but ultimately, neither leaving a MicrotasksScope nor calling MicrotaskQueue::PerformCheckpoint() have any effect when inside another MicrotasksScope.

I’m wondering why Electron has the outer MicrotasksScope in the first place – can we avoid that somehow?

indutny:
I'm going to try out changing the microtasks policy right before entering uv_run() and then changing it back to what it was before that. Wish me luck!

indutny:
@codebytere @deepak1556 I've updated PR (title, text, and commit). Seems like only the microtask policy change is needed for Electron at the moment. I'll follow up with another PR when this Node.js change will land: https://github.com/nodejs/node/pull/38506 . With combined power of these two `napi_threadsafe_function` is going to be blazingly fast in Electron!

indutny:
Would appreciate another review. Thanks!

addaleax (approved):

deepak1556:
> I’m wondering why Electron has the outer MicrotasksScope in the first place – can we avoid that somehow?

For the browser process in Electron, we follow the same microtask policy as node in the explicit mode. There is a task observer that hooks into the chromium main thread task runner and allows us to perform checkpoint https://github.com/electron/electron/blob/e0fa327daf7dc925a09f81a941087c4d77aedb3f/shell/browser/microtasks_runner.h#L16-L21

Originally we were running our own checkpoints but that had issues and we later switched to always let node do it in the browser process https://github.com/electron/electron/pull/24131


For the renderer, we had to follow the scoped policy set by blink to avoid breaking the web apis. `gin::MicrotasksScope` was created to automatically adjust the electron apis to these different policies depending on the process they were used from https://github.com/electron/electron/blob/e0fa327daf7dc925a09f81a941087c4d77aedb3f/shell/common/gin_helper/microtasks_scope.cc#L13-L19



> I’m wondering why Electron has the outer MicrotasksScope in the first place – can we avoid that somehow?

This was originally added in https://github.com/electron/electron/pull/2447 , I think it was added when node did not perform any checkpoints as part of `uv_run`. The current approach LGTM /cc @zcbenz who originally added this.

deepak1556 (approved):
LGTM

@codebytere goma auth has expired in linux ci.

indutny-signal:
Getting some crashes locally:
```
  * frame #0: 0x00007fff2048f8b6 libsystem_malloc.dylib`malloc_error_break
    frame #1: 0x00007fff204833c4 libsystem_malloc.dylib`malloc_vreport + 440
    frame #2: 0x00007fff20497702 libsystem_malloc.dylib`malloc_zone_error + 183
    frame #3: 0x00007fff2047b182 libsystem_malloc.dylib`tiny_free_list_add_ptr + 1224
    frame #4: 0x00007fff2047a7aa libsystem_malloc.dylib`tiny_free_no_lock + 1116
    frame #5: 0x00007fff2047a1f9 libsystem_malloc.dylib`free_tiny + 442
    frame #6: 0x0000000107964960 Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 6114656
    frame #7: 0x0000000107938762 Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 5933922
    frame #8: 0x0000000107946d9e Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 5992862
    frame #9: 0x00000001079468be Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 5991614
    frame #10: 0x000000010797afb3 Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 6206387
    frame #11: 0x0000000107977a4a Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 6192714
    frame #12: 0x000000010797a92e Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 6204718
```

Investigating. Will probably need ASAN build.

deepak1556:
Also with this change, I am wondering if we need the `MicrotasksRunner` in the browser process, we have explicit scopes setup for the callsites in the electron internal which was not the case when this was originally added https://github.com/electron/electron/pull/15071. Can try removing it in a follow-up.

indutny-signal:
...and of course it doesn't crash with ASAN on. Could be my Node.js change that I'm floating on top of this patch. Going to verify it.

indutny-signal:
Better stack trace:
```
 * thread #1: tid = 0x29ab81, 0x00007fff2048f8b6 libsystem_malloc.dylib`malloc_error_break, name = 'CrRendererMain', queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x00007fff2048f8b6 libsystem_malloc.dylib`malloc_error_break
    frame #1: 0x00007fff204833c4 libsystem_malloc.dylib`malloc_vreport + 440
    frame #2: 0x00007fff20497702 libsystem_malloc.dylib`malloc_zone_error + 183
    frame #3: 0x00007fff2047b182 libsystem_malloc.dylib`tiny_free_list_add_ptr + 1224
    frame #4: 0x00007fff2047845d libsystem_malloc.dylib`tiny_malloc_from_free_list + 1640
    frame #5: 0x00007fff20477876 libsystem_malloc.dylib`tiny_malloc_should_clear + 233
    frame #6: 0x00007fff20476793 libsystem_malloc.dylib`szone_malloc_should_clear + 66
    frame #7: 0x000000010b83fb4f Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 6229839
    frame #8: 0x000000010b83fb4f Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 6229839
    frame #9: 0x00007fff2048fdfe libsystem_malloc.dylib`_malloc_zone_malloc + 118
    frame #10: 0x000000010b7b4906 Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 5659910
    frame #11: 0x000000010b0161af Electron Framework`v8_inspector::protocol::Binary::fromBase64(v8_inspector::String16 const&, bool*) + 1887
    frame #12: 0x000000010b027d46 Electron Framework`v8_inspector::V8StackTraceId::V8StackTraceId(unsigned long, std::__1::pair<long long, long long>, bool) + 134
    frame #13: 0x000000010b01d5dd Electron Framework`v8_inspector::DumpAsyncTaskStacksStateForTest(v8_inspector::V8Inspector*) + 28573
    frame #14: 0x000000010b0288e8 Electron Framework`v8_inspector::V8StackTraceId::V8StackTraceId(unsigned long, std::__1::pair<long long, long long>, bool) + 3112
    frame #15: 0x000000010b028639 Electron Framework`v8_inspector::V8StackTraceId::V8StackTraceId(unsigned long, std::__1::pair<long long, long long>, bool) + 2425
    frame #16: 0x000000010b028fc4 Electron Framework`v8_inspector::V8StackTraceId::V8StackTraceId(unsigned long, std::__1::pair<long long, long long>, bool) + 4868
    frame #17: 0x000000010b01c64d Electron Framework`v8_inspector::DumpAsyncTaskStacksStateForTest(v8_inspector::V8Inspector*) + 24589
    frame #18: 0x000000010b01a43f Electron Framework`v8_inspector::DumpAsyncTaskStacksStateForTest(v8_inspector::V8Inspector*) + 15871
    frame #19: 0x000000010b019e07 Electron Framework`v8_inspector::DumpAsyncTaskStacksStateForTest(v8_inspector::V8Inspector*) + 14279
    frame #20: 0x000000010b01a033 Electron Framework`v8_inspector::DumpAsyncTaskStacksStateForTest(v8_inspector::V8Inspector*) + 14835
    frame #21: 0x000000010ac6a734 Electron Framework`v8::internal::ClassScope::ResolvePrivateNamesPartially() + 43028
    frame #22: 0x000000010ac6a147 Electron Framework`v8::internal::ClassScope::ResolvePrivateNamesPartially() + 41511
    frame #23: 0x000000010b2bcf58 Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 451416
    frame #24: 0x000000010b256dcf Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 33231
    frame #25: 0x000000010b283fc1 Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 218049
    frame #26: 0x000000010b306dbb Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 754107
    frame #27: 0x000000010b276ef7 Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 164599
    frame #28: 0x000000010b254998 Electron Framework`v8::internal::SetupIsolateDelegate::SetupHeap(v8::internal::Heap*) + 23960
    frame #29: 0x000000010acbd501 Electron Framework`v8::internal::Execution::Call(v8::internal::Isolate*, v8::internal::Handle<v8::internal::Object>, v8::internal::Handle<v8::internal::Object>, int, v8::internal::Handle<v8::internal::Object>*) + 2705
    frame #30: 0x000000010acbdcf3 Electron Framework`v8::internal::Execution::TryCall(v8::internal::Isolate*, v8::internal::Handle<v8::internal::Object>, v8::internal::Handle<v8::internal::Object>, int, v8::internal::Handle<v8::internal::Object>*, v8::internal::Execution::MessageHandling, v8::internal::MaybeHandle<v8::internal::Object>*, bool) + 355
    frame #31: 0x000000010acbddd0 Electron Framework`v8::internal::Execution::TryCall(v8::internal::Isolate*, v8::internal::Handle<v8::internal::Object>, v8::internal::Handle<v8::internal::Object>, int, v8::inter
nal::Handle<v8::internal::Object>*, v8::internal::Execution::MessageHandling, v8::internal::MaybeHandle<v8::internal::Object>*, bool) + 576
    frame #32: 0x000000010acd5986 Electron Framework`v8::internal::MicrotaskQueue::RunMicrotasks(v8::internal::Isolate*) + 422
    frame #33: 0x000000010acd57b0 Electron Framework`v8::internal::MicrotaskQueue::PerformCheckpoint(v8::Isolate*) + 32
    frame #34: 0x000000010cc9f10a Electron Framework`node::CallbackScope::~CallbackScope() + 1146
    frame #35: 0x000000010cc9ec51 Electron Framework`node::CallbackScope::~CallbackScope() + 49
    frame #36: 0x000000010cd04a5c Electron Framework`napi_ref_threadsafe_function + 4860
```

indutny-signal:
Argh, nevermind. That was bad nodejs binding, not electron! This PR is good :grin:

codebytere (approved):

shell/common/node_bindings.cc
```diff
@@ -575,8 +575,9 @@ void NodeBindings::UvRunOnce() {
   // Enter node context while dealing with uv events.
   v8::Context::Scope context_scope(env->context());
 
-  // Perform microtask checkpoint after running JavaScript.
-  gin_helper::MicrotasksScope microtasks_scope(env->isolate());
+  auto old_policy = env->isolate()->GetMicrotasksPolicy();
+  DCHECK_EQ(v8::MicrotasksScope::GetCurrentDepth(env->isolate()), 0);
+  env->isolate()->SetMicrotasksPolicy(v8::MicrotasksPolicy::kExplicit);
```
zcbenz, line 580: Can you add some comments on why this is needed? It will help when changing the code in future.

zcbenz (approved):

jkleinsc (changes requested):
Following up on @zcbenz's request: https://github.com/electron/electron/pull/28957#discussion_r625005309, can you add some comments to the code?

indutny:
@jkleinsc @zcbenz fixed. thanks your for feedback!

shell/common/node_bindings.cc
indutny: Added a comment. Thanks!

jkleinsc (approved):

indutny:
> crashReporter module getLastCrashReport returns the last uploaded report

Seems like unrelated CI failure, right? If yes - should this be merged?

indutny-signal:
Yay. Sorry to be a bother, but could it be backported to 12.x?

indutny-signal:
Thanks!

Issue #28927, opened 2021-04-29
[Bug]: Quit does not work anymore when a window has a modal dialog open


bpasero:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/electron/electron/issues) for a feature request that matches the one I want to file, without success.

### Electron Version

12.0.5

### What operating system are you using?

macOS

### Operating System Version

11.3

### What arch are you using?

arm64 (including Apple Silicon)

### Expected Behavior

Quit should work always.

### Actual Behavior

Quit does not work if a window has a modal dialog open. After closing the modal dialog you cannot click quit anymore.

Steps:
* run the fiddle
* quit (e.g. `Cmd+Q`)
* close the modal dialog
* quit again 

=> 🐛 it does not work

### Testcase Gist URL

https://gist.github.com/5f332310d6c0ebdea3bee907cc1be2fc

bpasero:
👏

PR #28967, opened 2021-05-03
fix: close attached sheet on window close

codebytere:
#### Description of Change

Closes https://github.com/electron/electron/issues/28927.

If a sheet is attached to the window when we call `[window_ performClose:nil]`, the window won't close properly
even after the user has ended the sheet. Ensure it's closed before calling `[window_ performClose:nil]`. Also fix an issue where `endSheet` may have been called with a `nil` sheet, triggering the following:

```sh
2021-05-03 11:39:00.871 Electron[30923:22268077] endSheet:returnCode: requires a non-nil sheet
```

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where the window couldn't be closed if a user tried to quit with a message box showing.

### DIFF
--- a/shell/browser/native_window_mac.mm
+++ b/shell/browser/native_window_mac.mm
@@ -473,6 +473,12 @@ void NativeWindowMac::Close() {
     return;
   }
 
+  // If a sheet is attached to the window when we call
+  // [window_ performClose:nil], the window won't close properly
+  // even after the user has ended the sheet.
+  // Ensure it's closed before calling [window_ performClose:nil].
+  SetEnabled(true);
+
   [window_ performClose:nil];
 
   // Closing a sheet doesn't trigger windowShouldClose,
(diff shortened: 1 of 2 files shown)


jkleinsc (approved):

shell/browser/native_window_mac.mm
```diff
@@ -473,6 +473,12 @@ void ViewDidMoveToSuperview(NSView* self, SEL _cmd) {
     return;
   }
 
+  // If a sheet is attached to the window when we call
+  // [window_ performClose:nil], the window won't close properly
+  // even after the user has ended the sheet.
+  // Ensure it's closed before calling [window_ performClose:nil].
+  SetEnabled(true);
```
zcbenz, line 480: The `Close()` would then be able to close an explicitly disabled window, but it seems that `SetEnabled(false)` crashes on macOS, so it probably does not matter.

zcbenz (approved):

shell/browser/native_window_mac.mm
codebytere: that was essentially my thought process as well - happy to consider another approach but this seemed to be the best all-around solution 🤔

PR #28971, opened 2021-05-03
fix: prevent crash on web-contents creation when error is thrown

### LINKED ISSUE #28749: Crash when throwing an error in app.on('web-contents-created')
Repro gist: https://gist.github.com/2b0373a300af0931c0c29b0e8dff8b73

Stack trace:

```
[39005:0420/121556.035182:FATAL:electron_api_browser_window.cc(71)] Check failed: web_contents_view.get(). 
0   Electron Framework                  0x000000010ba38aa9 base::debug::CollectStackTrace(void**, unsigned long) + 9
1   Electron Framework                  0x000000010b955563 base::debug::StackTrace::StackTrace() + 19
2   Electron Framework                  0x000000010b96e07f logging::LogMessage::~LogMessage() + 175
3   Electron Framework                  0x000000010b96ef4e logging::LogMessage::~LogMessage() + 14
4   Electron Framework                  0x0000000106cdffdd electron::api::BrowserWindow::BrowserWindow(gin::Arguments*, gin_helper::Dictionary const&) + 765
5   Electron Framework                  0x0000000106ce24c0 electron::api::BrowserWindow::New(gin_helper::ErrorThrower, gin::Arguments*) + 368
6   Electron Framework                  0x0000000106ce2b42 base::internal::Invoker<base::internal::BindState<gin_helper::WrappableBase* (*)(gin_helper::ErrorThrower, gin::Arguments*)>, gin_helper::WrappableBase* (gin_helper::ErrorThrower, gin::Arguments*)>::Run(base::internal::BindStateBase*, gin_helper::ErrorThrower&&, gin::Arguments*) + 34
7   Electron Framework                  0x0000000106ce2af8 gin_helper::WrappableBase* gin_helper::internal::InvokeFactory<gin_helper::ErrorThrower, gin::Arguments*>(gin::Arguments*, base::RepeatingCallback<gin_helper::WrappableBase* (gin_helper::ErrorThrower, gin::Arguments*)> const&) + 152
8   Electron Framework                  0x0000000106ce29b1 void gin_helper::internal::InvokeNew<gin_helper::WrappableBase* (gin_helper::ErrorThrower, gin::Arguments*)>(base::RepeatingCallback<gin_helper::WrappableBase* (gin_helper::ErrorThrower, gin::Arguments*)> const&, v8::Isolate*, gin_helper::Arguments*) + 81
9   Electron Framework                  0x0000000106cd0624 gin_helper::Dispatcher<void (v8::Isolate*, gin_helper::Arguments*)>::DispatchToCallback(v8::FunctionCallbackInfo<v8::Value> const&) + 356
10  Electron Framework                  0x00000001088fc1a2 v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo) + 994
11  Electron Framework                  0x00000001088f9bd0 v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<true>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) + 928
12  Electron Framework                  0x00000001088f8d1f v8::internal::Builtin_Impl_HandleApiCall(v8::internal::BuiltinArguments, v8::internal::Isolate*) + 383
13  Electron Framework                  0x00000001088f890d v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) + 109
14  ???                                 0x0000004600318bdf 0x0 + 300650957791
```

it's a DCHECK, but I'm pretty sure it will crash later if this is null.

VerteDinde:
#### Description of Change

Fixes #28749 

Currently, if an unhandled error is thrown during web-contents creation on a new browser window within JavaScript's scope, the error is not being correctly bubbled up, causing a DCHECK and hard app crash. More specifically, here are the steps that could lead to a crash (thanks @MarshallOfSound for the breakdown!):

- User's JS code triggers some native code
- The native code emits an event
- The event is emitted in the same scope as the JS code that triggered the native code
- The event handler throws an error
- The native code then tries to run more JS (emit another event, call a getter, etc.), which cannot happen because there is a pending exception

This PR adds Node's exception error handler to the `_init` method for web-contents creation, bubbling the error into a different scope and allowing us to handle it more appropriately, rather than hard crashing.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fix crash when an exception occurs within the event emitter.

### DIFF
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -3708,7 +3708,13 @@ gin::Handle<WebContents> WebContents::New(
     const gin_helper::Dictionary& options) {
   gin::Handle<WebContents> handle =
       gin::CreateHandle(isolate, new WebContents(isolate, options));
+  // Send unhandled errors within init to a different scope,
+  // so they're handled within JavaScript and don't crash within C++
+  v8::TryCatch try_catch(isolate);
   gin_helper::CallMethod(isolate, handle.get(), "_init");
+  if (try_catch.HasCaught()) {
+    node::errors::TriggerUncaughtException(isolate, try_catch);
+  }
   return handle;
 }
 
@@ -3719,7 +3725,13 @@ gin::Handle<WebContents> WebContents::CreateAndTake(
     Type type) {
   gin::Handle<WebContents> handle = gin::CreateHandle(
       isolate, new WebContents(isolate, std::move(web_contents), type));
+  // Send unhandled errors within init to a different scope,
+  // so they're handled within JavaScript and don't crash within C++
+  v8::TryCatch try_catch(isolate);
   gin_helper::CallMethod(isolate, handle.get(), "_init");
+  if (try_catch.HasCaught()) {
+    node::errors::TriggerUncaughtException(isolate, try_catch);
+  }
   return handle;
 }
 
@@ -3739,7 +3751,11 @@ gin::Handle<WebContents> WebContents::FromOrCreate(
   WebContents* api_web_contents = From(web_contents);
   if (!api_web_contents) {
     api_web_contents = new WebContents(isolate, web_contents);
+    v8::TryCatch try_catch(isolate);
     gin_helper::CallMethod(isolate, api_web_contents, "_init");
+    if (try_catch.HasCaught()) {
+      node::errors::TriggerUncaughtException(isolate, try_catch);
+    }
   }
   return gin::CreateHandle(isolate, api_web_contents);
 }
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -3708,13 +3708,7 @@ gin::Handle<WebContents> WebContents::New(
     const gin_helper::Dictionary& options) {
   gin::Handle<WebContents> handle =
       gin::CreateHandle(isolate, new WebContents(isolate, options));
-  // Send unhandled errors within init to a different scope,
-  // so they're handled within JavaScript and don't crash within C++
-  v8::TryCatch try_catch(isolate);
-  gin_helper::CallMethod(isolate, handle.get(), "_init");
-  if (try_catch.HasCaught()) {
-    node::errors::TriggerUncaughtException(isolate, try_catch);
-  }
+  gin_helper::CallMethodCatchException(isolate, handle.get(), "_init");
   return handle;
 }
 
@@ -3725,13 +3719,7 @@ gin::Handle<WebContents> WebContents::CreateAndTake(
     Type type) {
   gin::Handle<WebContents> handle = gin::CreateHandle(
       isolate, new WebContents(isolate, std::move(web_contents), type));
-  // Send unhandled errors within init to a different scope,
-  // so they're handled within JavaScript and don't crash within C++
-  v8::TryCatch try_catch(isolate);
-  gin_helper::CallMethod(isolate, handle.get(), "_init");
-  if (try_catch.HasCaught()) {
-    node::errors::TriggerUncaughtException(isolate, try_catch);
-  }
+  gin_helper::CallMethodCatchException(isolate, handle.get(), "_init");
   return handle;
 }
 
@@ -3751,11 +3739,7 @@ gin::Handle<WebContents> WebContents::FromOrCreate(
   WebContents* api_web_contents = From(web_contents);
   if (!api_web_contents) {
     api_web_contents = new WebContents(isolate, web_contents);
-    v8::TryCatch try_catch(isolate);
-    gin_helper::CallMethod(isolate, api_web_contents, "_init");
-    if (try_catch.HasCaught()) {
-      node::errors::TriggerUncaughtException(isolate, try_catch);
-    }
+    gin_helper::CallMethodCatchException(isolate, api_web_contents, "_init");
   }
   return gin::CreateHandle(isolate, api_web_contents);
 }
--- a/shell/browser/api/electron_api_web_contents.cc
+++ b/shell/browser/api/electron_api_web_contents.cc
@@ -3708,7 +3708,7 @@ gin::Handle<WebContents> WebContents::New(
     const gin_helper::Dictionary& options) {
   gin::Handle<WebContents> handle =
       gin::CreateHandle(isolate, new WebContents(isolate, options));
-  gin_helper::CallMethodCatchException(isolate, handle.get(), "_init");
+  CatchException(isolate, handle);
   return handle;
 }
 
@@ -3719,7 +3719,7 @@ gin::Handle<WebContents> WebContents::CreateAndTake(
     Type type) {
   gin::Handle<WebContents> handle = gin::CreateHandle(
       isolate, new WebContents(isolate, std::move(web_contents), type));
-  gin_helper::CallMethodCatchException(isolate, handle.get(), "_init");
+  CatchException(isolate, handle);
   return handle;
 }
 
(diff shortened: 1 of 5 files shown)


shell/common/gin_helper/event_emitter_caller.h
```diff
@@ -34,7 +35,15 @@ v8::Local<v8::Value> EmitEvent(v8::Isolate* isolate,
...
-  return internal::CallMethodWithArgs(isolate, obj, "emit", &concatenated_args);
+
+  v8::TryCatch try_catch(isolate);
+  v8::Local<v8::Value> value =
+      internal::CallMethodWithArgs(isolate, obj, "emit", &concatenated_args);
+  if (try_catch.HasCaught()) {
+    LOG(ERROR) << "An unknown error occurred in the event emitter";
+  }
+
+  return value;
```
VerteDinde, line 46: Is returning the value after an error a good best practice here?

shell/common/gin_helper/event_emitter_caller.h
```diff
@@ -34,7 +35,15 @@ v8::Local<v8::Value> EmitEvent(v8::Isolate* isolate,
   internal::ValueVector concatenated_args = {gin::StringToV8(isolate, name)};
   concatenated_args.reserve(1 + args.size());
   concatenated_args.insert(concatenated_args.end(), args.begin(), args.end());
-  return internal::CallMethodWithArgs(isolate, obj, "emit", &concatenated_args);
+
+  v8::TryCatch try_catch(isolate);
+  v8::Local<v8::Value> value =
+      internal::CallMethodWithArgs(isolate, obj, "emit", &concatenated_args);
+  if (try_catch.HasCaught()) {
+    LOG(ERROR) << "An unknown error occurred in the event emitter";
```
MarshallOfSound, line 43: We need to rethrow the caught exception here, we should either send this to the global error handler `global.onerror` or do `Emit('error', exception)` which is more nodejs-esque. 🤔 The NodeJS `error` event handler will dispatch the error event to the global handler for us iirc.

shell/common/gin_helper/event_emitter_caller.h
MarshallOfSound: Yes, `CallMethodWithArgs` guarantees the value is non-empty even if an error is thrown.

shell/common/gin_helper/event_emitter_caller.h
VerteDinde: Thanks! Will make that change 🙂

zcbenz (approved):

zcbenz:
Can you add test for this? Putting https://gist.github.com/nornagon/2b0373a300af0931c0c29b0e8dff8b73 under `spec-main/fixtures/crash-cases/` should be enough.

codebytere (approved):

VerteDinde:
@zcbenz Will do! Thanks for reviewing! 🙂

ckerr (approved):

shell/browser/api/electron_api_web_contents.cc
```diff
@@ -3708,7 +3708,13 @@ gin::Handle<WebContents> WebContents::New(
     const gin_helper::Dictionary& options) {
   gin::Handle<WebContents> handle =
       gin::CreateHandle(isolate, new WebContents(isolate, options));
+  // Send unhandled errors within init to a different scope,
+  // so they're handled within JavaScript and don't crash within C++
+  v8::TryCatch try_catch(isolate);
   gin_helper::CallMethod(isolate, handle.get(), "_init");
+  if (try_catch.HasCaught()) {
+    node::errors::TriggerUncaughtException(isolate, try_catch);
+  }
```
nornagon, line 3717: It might be useful to have a helper method that does this logic? eg. `gin_helper::CallMethodCatchException` or something

nornagon (approved):

shell/browser/api/electron_api_web_contents.cc
VerteDinde: @nornagon Great idea, changed it add and use the helper 🙇‍♀️

spec-main/fixtures/crash-cases/webview-contents-error-on-creation/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+// Quit when all windows are closed, except on macOS. There, it's common
+// for applications and their menu bar to stay active until the user quits
+// explicitly with Cmd + Q.
+app.on('window-all-closed', function () {
+  if (process.platform !== 'darwin') app.quit();
+});
+
+// In this file you can include the rest of your app's specific main process
+// code. You can also put them in separate files and require them here.
```
nornagon, line 47: Let's simplify this to the bare minimum to make it obvious what's being tested. Also, let's exit the process with a 0 exit code when we're successful.

```suggestion
const { app, BrowserWindow } = require('electron');

app.on('web-contents-created', (wc) => {
  throw new Error();
});

app.whenReady().then(async () => {
  const mainWindow = new BrowserWindow({ show: false });
  await mainWindow.loadURL('about:blank');
  process.exit(0);
})
```

VerteDinde: Oof, yes, thank you for catching that 🙇‍♀️

shell/common/gin_helper/event_emitter_caller.h
```diff
@@ -64,6 +65,28 @@ v8::Local<v8::Value> CustomEmit(v8::Isolate* isolate,
...
+  v8::EscapableHandleScope scope(isolate);
+  v8::Local<v8::Object> v8_object;
+  if (object->GetWrapper(isolate).ToLocal(&v8_object)) {
+    // Send unhandled errors within init to a different scope,
+    // so they're handled within JavaScript and don't crash within C++
+    v8::TryCatch try_catch(isolate);
+    v8::Local<v8::Value> value = scope.Escape(CustomEmit(
+        isolate, v8_object, method_name, std::forward<Args>(args)...));
+    if (try_catch.HasCaught()) {
+      node::errors::TriggerUncaughtException(isolate, try_catch);
```
MarshallOfSound, line 82: This is unsafe, we can't call this method in a `common` file.  Can we put it back in the callsite for now until we figure out a more generic strategy that supports non-node renderer contexts too.

cc @nornagon because I think you requested this change, we can move things to a helper but only once we've thought through the behavior in other contexts.  Let's fix the exact problem this time and refactor into a more generic solution at other callsites in the future

MarshallOfSound (changes requested):

shell/common/gin_helper/event_emitter_caller.h
nornagon: just put it in a helper that's in `shell/browser`

VerteDinde: @nornagon @MarshallOfSound Converted the method into a static helper within web contents - that way, it's more DRY but stays close to the original call site. I added an overload since the two `_init` functions take slightly different params (one takes a gin handle, the other web_contents).

If we want to store the helper somewhere else or if there's another/better way to do the function overload, let me know, happy to change it! 🙂  Thank you both for reviewing too!

VerteDinde: Update: The helper was failing tests on Mac - I went ahead and moved the catches back to the original call sites.

MarshallOfSound (approved):

